PR #3260, opened 2026-01-12
Add upstream Identity Provider types and interfaces

jhrozek:
Introduce foundational types for communicating with upstream Identity Providers (e.g. Google) in the authorization server:

- Provider interface: abstracts IDP operations (authorization URL generation, code exchange, token refresh, userinfo)
- IDTokenNonceValidator: optional interface for OIDC nonce validation to prevent replay attacks
- UserInfoSubjectValidator: optional interface for OIDC Section 5.3.4 subject validation to prevent impersonation
- Tokens: represents access/refresh/ID tokens from upstream IDP
- IDTokenClaims: parsed OIDC ID token claims including azp for multi-audience token validation
- Config: upstream IDP connection configuration with redirect URI validation

These types are implemented by OIDCProvider in the full authorization server and used for token exchange flows where the auth server acts as an OAuth client to upstream IDPs.

### DIFF
--- a/pkg/authserver/upstream/types.go
+++ b/pkg/authserver/upstream/types.go
@@ -17,7 +17,6 @@ package upstream
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"net/url"
 	"strings"
 	"time"
@@ -131,12 +130,6 @@ type UserInfoSubjectValidator interface {
 // that does not match the expected subject from the ID Token.
 // This validation is required per OIDC Core Section 5.3.4 to prevent user impersonation.
 var ErrUserInfoSubjectMismatch = fmt.Errorf("userinfo subject does not match expected subject")
-
-// HTTPClient is an interface for HTTP client operations.
-// This allows for mocking in tests.
-type HTTPClient interface {
-	Do(req *http.Request) (*http.Response, error)
-}
 
 // Config contains configuration for connecting to an upstream
 // Identity Provider (e.g., Google, Okta, Auth0).


pkg/authserver/upstream/types.go
```diff
@@ -0,0 +1,249 @@
...
+
+// ErrUserInfoSubjectMismatch is returned when the UserInfo endpoint returns a subject
+// that does not match the expected subject from the ID Token.
+// This validation is required per OIDC Core Section 5.3.4 to prevent user impersonation.
+var ErrUserInfoSubjectMismatch = fmt.Errorf("userinfo subject does not match expected subject")
+
+// HTTPClient is an interface for HTTP client operations.
+// This allows for mocking in tests.
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
```
dmjb: Nitpick: I feel like we have this interface for testing somewhere else in the code... perhaps we should consider defining it in one place?

dmjb (dismissed):

pkg/authserver/upstream/types.go
jhrozek: thanks for the review, I ended up removing the interface for now, it will be used later in tests. You're right we do have an interface in one of the oauth packages, so my plan is to move the interface to a shared networking code and make it public when I need to use it in tests in a couple of PRs 

(there PRs are cherry-picked from a larger branch and sometimes code is commited that will be used only in a later commit)

pkg/authserver/server/session/session_test.go
```diff
@@ -240,11 +240,10 @@ func TestSession_UpstreamSessionID(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name      string
-		session   *Session
-		setID     string
-		wantGet   string
-		wantPanic bool
+		name    string
```
jhrozek, line 243: the wantPanic removal is a drive-by fix - it ended up being unused after I had refactored the tests

PR #3263, opened 2026-01-12
Add OAuth client factory with RFC 8252 loopback support to authserver

jhrozek:
Add client factory and LoopbackClient for native OAuth clients like VS Code and Claude Code. Currently the code is unused (but tested!), when integrated, the authorize handler will use LoopbackClient.

LoopbackClient wraps fosite.DefaultClient to provide RFC 8252 Section 7.3 dynamic port matching for "localhost" hostname. Fosite's built-in isMatchingAsLoopback uses net.ParseIP().IsLoopback() which only supports IP addresses (127.0.0.1, ::1), not the "localhost" hostname.

### DIFF
--- a/.github/workflows/pr-size-label-apply.yml
+++ b/.github/workflows/pr-size-label-apply.yml
@@ -123,7 +123,7 @@ jobs:
               hasXLLabel: hasXLLabel,
               hasJustification: hasJustification,
               needsEnforcement: hasXLLabel && !hasJustification,
-              shouldDismiss: hasXLLabel && hasJustification
+              shouldDismiss: (hasXLLabel && hasJustification) || !hasXLLabel
             };
 
       - name: Request changes if no justification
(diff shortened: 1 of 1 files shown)


pkg/authserver/server/registration/client.go
```diff
@@ -0,0 +1,257 @@
...
+
+// hostnamesMatch checks if two hostnames should be considered equivalent for
+// loopback matching purposes.
+//
+// Per RFC 8252, the hostname must match exactly. We normalize localhost to
+// be case-insensitive, but 127.0.0.1 and localhost are treated as different
+// hostnames (a client registered with 127.0.0.1 will not match localhost requests).
+func hostnamesMatch(requested, registered string) bool {
+	// Case-insensitive comparison for localhost
+	if strings.EqualFold(requested, "localhost") && strings.EqualFold(registered, "localhost") {
```
JAORMX, line 254: this looks brittle; wouldn't it be better to use [url.Parse](https://pkg.go.dev/net/url#Parse) and then compares [the hostnames](https://pkg.go.dev/net/url#example-URL.Hostname)?

jhrozek: Thanks for the review! Looking at the call site, `hostnamesMatch` is called from `matchesAsLoopback` (line 216) which already parses the URIs:

```go
requested, err := url.Parse(requestedURI)
// ...
registered, err := url.Parse(registeredURI)
// ...
if !hostnamesMatch(requested.Hostname(), registered.Hostname()) {
```

So `hostnamesMatch` receives the result of `url.Hostname()` calls, not raw URIs. These are already clean hostname strings (e.g., `localhost`, `127.0.0.1`, `::1`). Adding `url.Parse` inside `hostnamesMatch` would actually break it, since hostnames alone are not valid URIs.

That said, I agree the function signature and documentation could be clearer about the expected input format. I've pushed a commit (362dc5e) that improves the documentation to explicitly note that the parameters are pre-parsed hostname strings from `url.Hostname()`, not raw URIs.

Issue #2213, opened 2025-10-15
Limit the logs size on API


peppescg:
Currently we are returning the entire log file contents without size limits. This can introduce a bunch of issues, memory, response and latency issues.
We should limit the log size, using pagination probably for having the full logs

PR #3279, opened 2026-01-13
Implement logging pagination for the UI

dmjb:
Previously the number of lines returned by the logging endpoints was unbounded. This could lead to problems for clients, or even cause the server to crash.

Implement limit/offset pagination on the logging endpoints. Enforce a maximum limit of 1,000 log lines per request. Return the total number of log lines in the response headers so that the caller can figure out how many requests they need to make.

**NOTE**: Extra testing will be added in a follow-on PR where I will also do some additional refactoring. I have left it out in order to keep the PR readable.

# Details:

 ## New Response Headers

 Both /api/v1beta/workloads/{name}/logs and /api/v1beta/workloads/{name}/proxy-logs now include:

 - X-Total-Lines: Total number of log lines available
 - X-Offset: Number of lines skipped (the offset query parameter value)
 - X-Limit: Maximum lines returned (the lines query parameter value)

 ## Example

```
 GET /api/v1beta/workloads/my-server/logs?lines=100&offset=200

 HTTP/1.1 200 OK
 Content-Type: text/plain
 X-Total-Lines: 5432
 X-Offset: 200
 X-Limit: 100
```

Fixes: #2213

### DIFF
--- a/pkg/container/docker/client.go
+++ b/pkg/container/docker/client.go
@@ -481,17 +481,7 @@ func (c *Client) GetWorkloadLogs(
 	if maxLines > 1000 {
 		maxLines = 1000 // cap at max for API protection
 	}
-
-	// For pagination support (when maxLines > 0), we need to read all logs to get total count
-	// For CLI (maxLines <= 0), we can skip counting
-	needsTotalCount := maxLines > 0
 	tailLines := "all"
-
-	// If not paginating and not needing count, we could use Tail optimization
-	// But since we need total count for API responses, we always read all when maxLines > 0
-	if !needsTotalCount && maxLines > 0 {
-		tailLines = strconv.Itoa(maxLines)
-	}
 
 	options := container.LogsOptions{
 		ShowStdout: true,


pkg/api/v1/workloads.go
```diff
@@ -534,7 +541,33 @@ func (s *WorkloadRoutes) getLogsForWorkload(w http.ResponseWriter, r *http.Reque
...
-	logs, err := s.workloadManager.GetLogs(ctx, name, false)
+	// Parse and validate lines parameter
+	maxLines := 100 // default
+	if linesParam := r.URL.Query().Get("lines"); linesParam != "" {
+		parsed, err := strconv.Atoi(linesParam)
+		if err != nil || parsed <= 0 {
+			http.Error(w, "Invalid lines parameter: must be a positive integer", http.StatusBadRequest)
+			return
+		}
+		if parsed > 10000 {
```
jhrozek: Is it expected that workloads cap the number of lines at 10k but docker and k8s at 1k?

pkg/container/docker/client.go
```diff
@@ -468,22 +468,45 @@ func (c *Client) RemoveWorkload(ctx context.Context, workloadName string) error
...
+	// - If maxLines > 0: limit to that number (for API usage), cap at 1000
+	// - If offset > 0 or maxLines > 0: need total count for pagination, must read all logs
+	// - For API usage, always read all logs to get accurate total count
+
+	if maxLines > 1000 {
+		maxLines = 1000 // cap at max for API protection
+	}
+
+	// For pagination support (when maxLines > 0), we need to read all logs to get total count
+	// For CLI (maxLines <= 0), we can skip counting
```
jhrozek: should we have an upper limit anyway of say several MB to avoid having all the logs in memory using e.g. io.LimitReader and then send a header to the client telling it that the logs have been truncaed?

pkg/container/docker/client.go
```diff
@@ -1775,3 +1815,27 @@ func (c *Client) inspectContainerByName(ctx context.Context, workloadName string
 
 	return c.api.ContainerInspect(ctx, containerID)
 }
+
+// sliceLogLines applies offset and limit to log content for pagination.
+// offset: number of lines to skip from the beginning
+// maxLines: maximum number of lines to return (0 = unlimited)
+func sliceLogLines(content string, offset int, maxLines int) string {
```
jhrozek: this function...

pkg/container/kubernetes/client.go
```diff
@@ -284,10 +310,51 @@ func (c *Client) GetWorkloadLogs(ctx context.Context, workloadName string, follo
...
+		logContent = sliceLogLines(logContent, offset, maxLines)
+	}
+
+	return logContent, totalLines, nil
+}
+
+// sliceLogLines applies offset and limit to log content for pagination.
+// offset: number of lines to skip from the beginning
+// maxLines: maximum number of lines to return (0 = unlimited)
+func sliceLogLines(content string, offset int, maxLines int) string {
```
jhrozek: is quite similar to this function...

pkg/workloads/manager.go
```diff
@@ -1523,3 +1558,27 @@ func (d *DefaultManager) getRemoteWorkloadsFromState(
 
 	return remoteWorkloads, nil
 }
+
+// sliceProxyLogLines applies offset and limit to log content for pagination.
+// offset: number of lines to skip from the beginning
+// maxLines: maximum number of lines to return (0 = unlimited)
+func sliceProxyLogLines(content string, offset int, maxLines int) string {
```
jhrozek: .. and this function...

pkg/container/docker/client.go
```diff
@@ -468,22 +468,45 @@ func (c *Client) RemoveWorkload(ctx context.Context, workloadName string) error
...
+
+	// For pagination support (when maxLines > 0), we need to read all logs to get total count
+	// For CLI (maxLines <= 0), we can skip counting
+	needsTotalCount := maxLines > 0
+	tailLines := "all"
+
+	// If not paginating and not needing count, we could use Tail optimization
+	// But since we need total count for API responses, we always read all when maxLines > 0
+	if !needsTotalCount && maxLines > 0 {
+		tailLines = strconv.Itoa(maxLines)
```
jhrozek: needsTotalCount is maxLines > 0 so this line can be rewritten as:
```
!(maxLines > 0) && maxLines > 0
```
which is never true right?

pkg/api/v1/workloads.go
dmjb: I changed the limit and missed this spot.

pkg/container/docker/client.go
dmjb: This was a leftover from a previous iteration of the code, it is no longer relevant.

jhrozek (approved):

pkg/container/docker/monitor.go
```diff
@@ -125,7 +125,7 @@ func (m *ContainerMonitor) monitor(ctx context.Context) {
 				// Container has exited, get logs and info
 				// Create a short timeout context for these operations, derived from parent
 				infoCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
-				logs, _ := m.runtime.GetWorkloadLogs(infoCtx, m.containerName, false)
+				logs, _, _ := m.runtime.GetWorkloadLogs(infoCtx, m.containerName, false, 100, 0)
```
amirejaz: This returns the first 100 lines of the logs. Shouldn’t we retrieve the last 100 (or the last n) lines instead, since those are typically more useful? Or am I misunderstanding the use case?

pkg/container/kubernetes/client.go
```diff
@@ -241,7 +242,32 @@ func (c *Client) AttachToWorkload(ctx context.Context, workloadName string) (io.
...
+	}
+
+	// For pagination support (when maxLines > 0), we need to read all logs to get total count
+	// For CLI (maxLines <= 0), we can skip counting
+	needsTotalCount := maxLines > 0
+	var tailLines *int64
+
+	// If not paginating and not needing count, we could use TailLines optimization
+	// But since we need total count for API responses, we always read all when maxLines > 0
+	if !needsTotalCount && maxLines > 0 {
```
amirejaz, line 265: looks like condition `!needsTotalCount && maxLines > 0` can never be true.

pkg/container/kubernetes/client.go
```diff
@@ -241,7 +242,32 @@ func (c *Client) AttachToWorkload(ctx context.Context, workloadName string) (io.
...
+) (string, int, error) {
+	// Handle maxLines and offset:
+	// - If maxLines <= 0: unlimited (for CLI usage)
+	// - If maxLines > 0: limit to that number (for API usage), cap at 1000
+	// - If offset > 0 or maxLines > 0: need total count for pagination, must read all logs
+	// - For API usage, always read all logs to get accurate total count
+
+	if maxLines > 1000 {
+		maxLines = 1000 // cap at max for API protection
+	}
```
amirejaz, line 256: nit: I think we should also check `offset` should be a non-negative number

pkg/util/strings.go
```diff
@@ -0,0 +1,28 @@
...
+		return ""
+	}
+
+	lines := strings.Split(content, "\n")
+
+	// Apply offset
+	if offset >= len(lines) {
+		return "" // offset beyond available lines
+	}
+	lines = lines[offset:]
```
amirejaz, line 20: The API description says the endpoint returns the last (most recent) 100 lines by default, but the current slicing logic applies offset/maxLines from the start of the log content (oldest-first).

pkg/container/kubernetes/client.go
dmjb: This was leftover, I'll get rid of it here.

PR #3280, opened 2026-01-13
Add embedding engine

ptelang:
Add MCPEmbedding CRD for embedding model deployment in operator

Introduces a new MCPEmbedding custom resource to deploy HuggingFace
embedding models as MCP servers in Kubernetes. This enables semantic
search and similarity features for MCP tools and resources.

Key additions:
- MCPEmbedding CRD with model caching, resource management, and grouping
- Controller managing deployment lifecycle and model initialization
- Helm chart templates and CRD definitions
- Example configurations for basic and advanced use cases
- CRD API documentation updates

The embedding server uses huggingface-embedding-inference with persistent
volume caching for downloaded models and integrates with MCPGroups for
organizing backend workloads.

## Large PR Justification
Multiple related changes that would break if separated


cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
```diff
@@ -0,0 +1,274 @@
...
+	// ConditionVolumeReady indicates whether the PVC for model caching is ready
+	ConditionVolumeReady = "VolumeReady"
+)
+
+// Condition reasons for MCPEmbedding
+// Image validation, GroupRef, and PodTemplate reasons are shared with MCPServer
+
+const (
+	// ConditionReasonModelDownloading indicates the model is being downloaded
+	ConditionReasonModelDownloading = "ModelDownloading"
```
therealnb, line 26: Classically, this would be done in an init container, but if you have a controller, this is probably better.

cmd/thv-operator/controllers/embeddingserver_controller_test.go
```diff
@@ -0,0 +1,343 @@
...
+			assert.Equal(t, tt.expectedSize, size)
+			assert.Equal(t, tt.expectedAccess, string(accessMode))
+
+			// Verify storage class name if provided
+			if tt.modelCache.StorageClassName != nil {
+				assert.Equal(t, storageClassName, *tt.modelCache.StorageClassName)
+			}
+		})
+	}
+}
```
therealnb, line 329: We probably want some performance tests to check for things like OOM, throughput, latency, etc.

examples/operator/embedding-servers/basic-embedding.yaml
```diff
@@ -0,0 +1,20 @@
+# Basic MCPEmbedding example with minimal configuration
+# This creates an embedding server using the default text-embeddings-inference image
+apiVersion: toolhive.stacklok.dev/v1alpha1
+kind: MCPEmbedding
+metadata:
+  name: basic-embedding
+  namespace: toolhive-system
+spec:
+  # Required: HuggingFace model to use
+  model: "sentence-transformers/all-MiniLM-L6-v2"
```
jerm-dro, line 10: This doesn't require any huggingface secret, right?

examples/operator/embeddings/embedding-with-group.yaml
```diff
@@ -0,0 +1,40 @@
...
+---
+# Create an embedding server that belongs to the group
+apiVersion: toolhive.stacklok.dev/v1alpha1
+kind: MCPEmbedding
+metadata:
+  name: ml-embedding
+  namespace: toolhive-system
+spec:
+  # Reference the MCPGroup
+  groupRef: "ml-services"
```
jerm-dro: Why would the embedding service need to reference a group?

examples/operator/embedding-servers/basic-embedding.yaml
jerm-dro: How would users provided them if they needed to?

cmd/thv-operator/api/v1alpha1/mcpembedding_types.go
```diff
@@ -0,0 +1,274 @@
...
+	PodTemplateSpec *runtime.RawExtension `json:"podTemplateSpec,omitempty"`
+
+	// ResourceOverrides allows overriding annotations and labels for resources created by the operator
+	// +optional
+	ResourceOverrides *EmbeddingResourceOverrides `json:"resourceOverrides,omitempty"`
+
+	// GroupRef is the name of the MCPGroup this embedding server belongs to
+	// Must reference an existing MCPGroup in the same namespace
+	// +optional
+	GroupRef string `json:"groupRef,omitempty"`
```
jerm-dro: Is this necessary? We only use `GroupRef` if we need to do some discovery of objects in the group. 

In this case, the embedding server doesn't need to discover anything or be discovered by anything, right?

cmd/thv-operator/api/v1alpha1/mcpembedding_types.go
```diff
@@ -0,0 +1,274 @@
...
+	// ConditionReasonVolumeCreating indicates the PVC is being created
+	ConditionReasonVolumeCreating = "VolumeCreating"
+	// ConditionReasonVolumeReady indicates the PVC is ready
+	ConditionReasonVolumeReady = "VolumeReady"
+	// ConditionReasonVolumeFailed indicates the PVC creation failed
+	ConditionReasonVolumeFailed = "VolumeFailed"
+)
+
+// MCPEmbeddingSpec defines the desired state of MCPEmbedding
+type MCPEmbeddingSpec struct {
```
jerm-dro: nit: I'd suggest calling this and similar types `EmbeddingServer` for a few reasons:
- Other CRDs that run servers have the "Server" suffix.
- This server isn't directly involved in the MCP protocol and instead just provides embeddings which could be used outside the MCP context.

cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
```diff
@@ -0,0 +1,274 @@
...
+	// Note that to modify the specific container the embedding server runs in, you must specify
+	// the 'embedding' container name in the PodTemplateSpec.
+	// +optional
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Type=object
+	PodTemplateSpec *runtime.RawExtension `json:"podTemplateSpec,omitempty"`
+
+	// ResourceOverrides allows overriding annotations and labels for resources created by the operator
+	// +optional
+	ResourceOverrides *EmbeddingResourceOverrides `json:"resourceOverrides,omitempty"`
```
jerm-dro, line 96: Nit: Is this necessary? Can we delete it?

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,989 @@
...
+
+	// Validate PodTemplateSpec early
+	if !r.validateAndUpdatePodTemplateStatus(ctx, embedding) {
+		return ctrl.Result{}, nil
+	}
+
+	// Validate image
+	if err := r.validateImage(ctx, embedding); err != nil {
+		return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
+	}
```
jerm-dro, line 157: Why swallow the error here? At the very least it should be logged.

cmd/thv-operator/controllers/mcpembedding_controller.go
```diff
@@ -0,0 +1,989 @@
...
+// handleDeletion handles the deletion of MCPEmbedding resources
+//
+//nolint:unparam // ctrl.Result return kept for consistency with reconciler pattern
+func (r *MCPEmbeddingReconciler) handleDeletion(
+	ctx context.Context,
+	embedding *mcpv1alpha1.MCPEmbedding,
+) (ctrl.Result, bool, error) {
+	if embedding.GetDeletionTimestamp() == nil {
+		return ctrl.Result{}, false, nil
+	}
```
jerm-dro: nit: I'd prefer we only call `handleDeletion` if we know it's a deletion. 

That means the caller does the following:
```go
if embedding.GetDeletionTimestamp() != nil {
    return r.handleDeletion(ctx, embedding)
}
```
Then, we don't this `bool` return to indicate if we're done or not.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,989 @@
...
+	if controllerutil.ContainsFinalizer(embedding, embeddingFinalizerName) {
+		return ctrl.Result{}, false, nil
+	}
+
+	controllerutil.AddFinalizer(embedding, embeddingFinalizerName)
+	err := r.Update(ctx, embedding)
+	if err != nil {
+		return ctrl.Result{}, true, err
+	}
+	return ctrl.Result{}, false, nil
```
jerm-dro, line 201: nit: this only returns `true` if there's also an `err` and `ctrl.Result` is always `ctrl.Result{}`. I'd recommend removing the two and updating the caller:

```
err := r.ensureFinalizer(...)
if err != nil {
    return ctrl.Result{}
}
```

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,989 @@
...
+		ctxLogger.Info("Creating a new Deployment", "Deployment.Namespace", dep.Namespace, "Deployment.Name", dep.Name)
+		err = r.Create(ctx, dep)
+		if err != nil {
+			ctxLogger.Error(err, "Failed to create new Deployment", "Deployment.Namespace", dep.Namespace, "Deployment.Name", dep.Name)
+			return ctrl.Result{}, true, err
+		}
+		return ctrl.Result{Requeue: true}, true, nil
+	} else if err != nil {
+		ctxLogger.Error(err, "Failed to get Deployment")
+		return ctrl.Result{}, true, err
```
jerm-dro: In this case, I think we want to requeue so the service gets created.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,989 @@
...
+	}
+	return ctrl.Result{}, false, nil
+}
+
+// ensureDeployment ensures the deployment exists and is up to date
+func (r *MCPEmbeddingReconciler) ensureDeployment(
+	ctx context.Context,
+	embedding *mcpv1alpha1.MCPEmbedding,
+) (ctrl.Result, bool, error) {
```
jerm-dro: nit: I haven't looked too closely, but can you find a refactor that allows us to delete the bool return?

Maybe relying on `result.Requeue`?

jerm-dro (changes requested):
This seems like it's on the right track, but I think we want an e2e test demonstrating:
- defining embedding servers with the CRD results in the creation of a deployment, service, etc
- we can call that service and get the expected results
- when we delete the CRD, the other objects are also deleted.

Also, please modify the pull request so it's against main. With an e2e test, I'd prefer to merge this right into main. It'll be one fully functional and validated change, so no need to wait on merging into main.

jerm-dro:
> This seems like it's on the right track, but I think we want an e2e test demonstrating:
> ...

You can crib off existing tests here: https://github.com/stacklok/toolhive/tree/main/test/e2e/thv-operator/virtualmcp

Putting these changes in the `virtualmcp` directory is fine with me, since we'll eventually be testing this when it's wired up to the gateway.

examples/operator/embedding-servers/basic-embedding.yaml
ptelang: Huggingface token can be provided in an env var in the spec.

examples/operator/embeddings/embedding-with-group.yaml
ptelang: Earlier I thought a vMCP server will discover its embedding server using the group ref, and we would deploy an embedding server per vMCP server. But that is not necessary and adds unecessary complexity. For now, I removed group ref.

cmd/thv-operator/api/v1alpha1/mcpembedding_types.go
ptelang: GroupRef is now removed. It is not needed at this time

cmd/thv-operator/api/v1alpha1/mcpembedding_types.go
ptelang: Good point, I renamed MCPEmbedding to EmbeddingServer.

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Logging added

cmd/thv-operator/controllers/embeddingserver_controller_test.go
ptelang: I created performance tests but did not add them to the repo yet due to complexities with how they will run in CI.

cmd/thv-operator/controllers/mcpembedding_controller.go
ptelang: All function calls in the reconciler loop are following that pattern. So I am going to keep it as-is for now.

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Fixed.

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Several functions are using the pattern of returning a bool. We can refactor, but I don't see it as a big issue. If you agree, I would like to leave it unchanged.

cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
ptelang: This is following the pattern for other crds like MCPServer. I will keep these for now.

PR #3284, opened 2026-01-14
Disable healthchecks for remote workloads for now

dmjb:
We have received reports of the remote workload healthchecks being brittle and putting workloads into an unhealthy state when they are still working. Disable for now as we investigate the problem further.

### DIFF
--- a/pkg/transport/proxy/transparent/pinger.go
+++ b/pkg/transport/proxy/transparent/pinger.go
@@ -17,12 +17,25 @@ type MCPPinger struct {
 	client    *http.Client
 }
 
+const (
+	// DefaultPingerTimeout is the default timeout for health check pings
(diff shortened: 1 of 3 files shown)


pkg/transport/http.go
```diff
@@ -260,8 +262,8 @@ func (t *HTTPTransport) Start(ctx context.Context) error {
 		targetURI,
 		t.prometheusHandler,
 		t.authInfoHandler,
-		true,
-		t.remoteURL != "",
+		isRemote, // TODO: reinstate this universally once we figure out how to make the checks less brittle.
```
amirejaz: should not this be `!isRemote` for health checks?

amirejaz (changes requested):
added a comment

PR #3288, opened 2026-01-14
Create E2E suite for ToolHive REST API

dmjb:
Our REST API has no e2e coverage, and various bugs have slipped through the cracks which could have been caught if we had similar coverage to what we have for the CLI.

This PR lays the foundations for an e2e test suite. A single trivial test for the healthcheck endpoint is added in order to start this effort. Extra scenarios will be added in follow up PRs.

As with our other E2E suites, these tests are run as a separate parallel action in our builds.

### DIFF
--- a/test/e2e/api_helpers.go
+++ b/test/e2e/api_helpers.go
@@ -40,6 +40,7 @@ type Server struct {
 	ctx        context.Context
 	cancel     context.CancelFunc
 	serverErr  chan error
+	done       chan struct{}
 	httpClient *http.Client
 }
 
@@ -63,6 +64,7 @@ func NewServer(config *ServerConfig) (*Server, error) {
 		ctx:       ctx,
 		cancel:    cancel,
 		serverErr: make(chan error, 1),
+		done:      make(chan struct{}),
 		httpClient: &http.Client{
 			Timeout: config.RequestTimeout,
 		},
@@ -70,6 +72,7 @@ func NewServer(config *ServerConfig) (*Server, error) {
 
 	// Start the server in a goroutine
 	go func() {
+		defer close(server.done)
 		// Create container runtime for the API server
 		containerRuntime, err := container.NewFactory().Create(ctx)
 		if err != nil {
@@ -144,8 +147,8 @@ func (s *Server) WaitForReady() error {
 // Stop stops the API server
 func (s *Server) Stop() {
 	s.cancel()
-	// Give server time to shut down gracefully
-	time.Sleep(100 * time.Millisecond)
+	// Wait for server to shut down gracefully
+	<-s.done
 }
 
 // Get performs a GET request to the specified path


.github/workflows/e2e-tests.yml
```diff
@@ -107,3 +107,52 @@ jobs:
...
+        run: ./test/e2e/api/run_tests.sh
+
+      - name: Upload test results (${{ matrix.title }})
+        if: always()
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
+        with:
+          name: ${{ matrix.artifact }}
+          path: |
+            test/e2e/api/ginkgo-report.xml
+            test/e2e/api/junit-report.xml
```
jhrozek: is this prep for later? I don't see the ginkgo invocation in the shell wrapper set up to produce junit

test/e2e/api_helpers.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+		}
+	}
+}
+
+// Stop stops the API server
+func (s *Server) Stop() {
+	s.cancel()
+	// Give server time to shut down gracefully
+	time.Sleep(100 * time.Millisecond)
```
jhrozek: every time I use a sleep instead of channel-based synchronization (checking something like `<-s.done`) I regret it later :-)

test/e2e/api_helpers.go
```diff
@@ -0,0 +1,191 @@
...
+
+			resp, err := s.httpClient.Do(req)
+			if err != nil {
+				continue
+			}
+			_ = resp.Body.Close()
+
+			// Server is ready if we get any response
+			if resp.StatusCode == http.StatusNoContent || resp.StatusCode == http.StatusServiceUnavailable {
+				return nil
```
jhrozek, line 141: I might be misreading the code but is 503 considered ready here?

test/e2e/api_healthcheck_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"io"
+	"net/http"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/stacklok/toolhive/test/e2e/api"
+)
+
+var _ = Describe("Healthcheck API", Label("api", "healthcheck"), func() {
```
jhrozek: should we add an e2e label?

test/e2e/api/run_tests.sh
```diff
@@ -0,0 +1,73 @@
...
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m' # No Color
+
+echo -e "${GREEN}ToolHive API E2E Test Runner${NC}"
+echo "================================"
+
+# Set TOOLHIVE_DEV environment variable to true
+export TOOLHIVE_DEV=true
```
jhrozek: I couldn't find any reference to this variable in the codebase, did you mean TOOLHIVE_DEBUG ?

jhrozek (approved):
I hve a bunch of comments but they are non-blocking

test/e2e/api/run_tests.sh
dmjb: It used to be used for disabling telemetry, it is present in the other e2e suite. I can remove it.

PR #3297, opened 2026-01-14
Add base OAuth 2.0 provider implementation for authserver

jhrozek:
This commit adds BaseOAuth2Provider, which implements OAuth 2.0 authorization code flow with PKCE support. It uses the golang.org/x/oauth2 library internally for token exchange and refresh operations.

Why use golang.org/x/oauth2:
- Handles OAuth 2.0 token responses correctly per RFC 6749, including edge cases like case-insensitive token_type comparison
- Manages token refresh automatically through TokenSource
- Extracts id_token from token extras (standard OIDC extension point)
- Battle-tested against many real-world IDP quirks

Why this architecture:
BaseOAuth2Provider is designed to be embedded by OIDCProvider (see auth-proxy-pr-7-upstream-fixes branch). The base provider handles pure OAuth 2.0 flows while OIDC adds discovery, ID token validation, and UserInfo fetching. This separation allows supporting both:
- OIDC providers (Google, Okta) via discovery
- Pure OAuth 2.0 providers (GitHub) via explicit endpoint configuration

The authserver uses these providers to implement token proxying: when a downstream client authenticates, the authserver exchanges codes with the upstream IDP, stores the upstream tokens, and issues its own tokens to the client. This enables MCP servers to authenticate users through enterprise IDPs without direct IDP integration.

UserInfo extensibility:
Adds UserInfoConfig and UserInfoFetcher interface to support providers with non-standard userinfo endpoints. OIDC providers use the discovered userinfo_endpoint with Bearer auth, but providers like GitHub require custom endpoints (/user), different field names (id vs sub), and additional headers (X-GitHub-Api-Version). UserInfoFieldMapping enables this without provider-specific code.

Other changes:
- Tokens value object extracted to separate file (tokens.go). Over the several PRs I want to minimize tokens.go, I don't like having a grab-bag file with unrelated types.

## Large PR Justification

- Over half of the PR is tests
- I think having the oidc provider and the userinfo handling could in theory be split, but having them in one PR shows how they fit together


pkg/authserver/upstream/oauth2.go
```diff
@@ -0,0 +1,470 @@
...
+
+// Validate checks that OAuth2Config has all required fields.
+func (c *OAuth2Config) Validate() error {
+	if c.AuthorizationEndpoint == "" {
+		return errors.New("authorization_endpoint is required for OAuth2 providers")
+	}
+	if c.TokenEndpoint == "" {
+		return errors.New("token_endpoint is required for OAuth2 providers")
+	}
+	return c.CommonOAuthConfig.Validate()
```
JAORMX, line 136: These endpoints are only checked for non-empty values, but not for HTTPS (for non-localhost) or private IPs. While `newHTTPClientForHost()` applies SSRF protections at request time, validating here would catch misconfigurations earlier.

Consider using `networking.ValidateEndpointURLWithInsecure()` or adding validation similar to `validateRedirectURI()` above.

pkg/authserver/upstream/userinfo_config.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if parsed.Scheme == "" || parsed.Host == "" {
+		return errors.New("endpoint_url must be an absolute URL with scheme and host")
+	}
+
+	if parsed.Scheme != networking.HttpScheme && parsed.Scheme != networking.HttpsScheme {
+		return errors.New("endpoint_url must use http or https scheme")
+	}
+
+	return nil
```
JAORMX, line 79: Unlike `validateRedirectURI()` in oauth2.go, this allows HTTP to any host. Might want to add the same loopback restriction for consistency:

```go
if parsed.Scheme == networking.HttpScheme && !networking.IsLocalhost(parsed.Host) {
    return errors.New("endpoint_url with http scheme requires loopback address")
}
```

pkg/authserver/upstream/oauth2.go
```diff
@@ -0,0 +1,470 @@
...
+	TokenEndpoint string
+
+	// UserInfo contains configuration for fetching user information (optional).
+	// When nil, the provider does not support UserInfo fetching.
+	UserInfo *UserInfoConfig
+}
+
+// Validate checks that OAuth2Config has all required fields.
+func (c *OAuth2Config) Validate() error {
+	if c.AuthorizationEndpoint == "" {
```
JAORMX, line 124: `TokenEndpoint` gets parsed/validated in `NewOAuth2Provider()` (line 285), but `AuthorizationEndpoint` is only checked for non-empty here. A malformed URL would only fail at runtime when `AuthorizationURL()` is called.

pkg/authserver/upstream/types_test.go
```diff
@@ -14,362 +14,5 @@
 
 package upstream
```
JAORMX: This file only contains a cleanup comment - could be removed in this PR.

pkg/authserver/upstream/oauth2.go
```diff
@@ -0,0 +1,470 @@
...
+}
+
+// Validate checks that CommonOAuthConfig has all required fields.
+func (c *CommonOAuthConfig) Validate() error {
+	if c.ClientID == "" {
+		return errors.New("client_id is required")
+	}
+	if c.RedirectURI == "" {
+		return errors.New("redirect_uri is required")
+	}
```
tgrunnagle, line 103: question: is `ClientSecret` not required?

pkg/authserver/upstream/oauth2.go
```diff
@@ -0,0 +1,470 @@
...
+// The config must include AuthorizationEndpoint and TokenEndpoint.
+func NewOAuth2Provider(config *OAuth2Config, opts ...OAuth2ProviderOption) (*BaseOAuth2Provider, error) {
+	if config == nil {
+		return nil, errors.New("config is required")
+	}
+
+	logger.Infow("creating OAuth2 provider",
+		"authorization_endpoint", config.AuthorizationEndpoint,
+		"token_endpoint", config.TokenEndpoint,
+		"client_id", config.ClientID,
```
tgrunnagle: nit: In my experience, some organizations consider client ids to be pseudo-secret. Maybe don't want to log it here.

jhrozek:
thanks @tgrunnagle and @JAORMX for the review. I think I addressed all of the comments in a fixup commit per comment.

There was another comment by @JAORMX on Slack that I will address as a separate refactoring PR which was about reusing structures between pkg/auth and pkg/authserver - I would prefer to make a fixup commit if possible.

pkg/authserver/upstream/oauth2.go
jhrozek: yeah, good point, removed

pkg/authserver/upstream/oauth2.go
jhrozek: no, not always, I added a comment to the code

tgrunnagle (approved):

pkg/authserver/upstream/types_test.go
jhrozek: yeah, I've been cherry-picking and polishing the commits from a larger branch where types.go was kindof a grab-bag of different code and I hated it (think utils.go kind of a file). This was a leftover - fixed, file removed.

pkg/authserver/upstream/oauth2.go
jhrozek: good point, fixed

pkg/authserver/upstream/userinfo_config.go
jhrozek: good point fixed

pkg/authserver/upstream/oauth2.go
jhrozek: I reused the utility function from networking

PR #3298, opened 2026-01-14
Add in-memory authserver storage implementation

jhrozek:
This is a preparatory commit for the auth-proxy feature. The storage package provides the persistence layer for the OAuth 2.0 authorization server. In the complete branch, this storage is:

- Passed to authserver.New() and used by the fosite OAuth2 provider for authorization codes, access tokens, refresh tokens, and PKCE
- Used by handlers to validate clients and store pending authorizations while users authenticate with the upstream identity provider
- Exposed via Server.IDPTokenStorage() for the token swap middleware, which replaces client JWTs with upstream IDP tokens for backend requests

This commit adds:

- Storage interface embedding fosite's OAuth2 storage interfaces plus ToolHive-specific interfaces for upstream tokens, pending authorizations, and client registration
- MemoryStorage implementation with thread-safe maps and background cleanup
- Comprehensive documentation in doc.go explaining fosite's type system (Requester, Session, signature vs request ID) to help developers understand why the API is designed as it is

The in-memory implementation is suitable for single-instance deployments and development. Production deployments will use a distributed storage backend (Redis or database).

## Large PR Justification

We need to implement all of the required fosite interfaces, doing it in multiple commits would just muddy the waters.

### DIFF
--- a/pkg/authserver/storage/memory.go
+++ b/pkg/authserver/storage/memory.go
@@ -155,69 +155,121 @@ func (s *MemoryStorage) cleanupLoop() {
 }
 
 // cleanupExpired removes all expired entries from storage.
+// Uses collect-then-delete pattern: collects expired keys under read lock,
+// then deletes under write lock. This minimizes write lock hold time.
 //
 //nolint:gocyclo // Function is straightforward, just repetitive cleanup loops for each storage type
 func (s *MemoryStorage) cleanupExpired() {
+	now := time.Now()
+
+	// Phase 1: Collect expired keys under read lock
+	s.mu.RLock()
+
+	var expiredAuthCodes []string
+	for k, v := range s.authCodes {
+		if now.After(v.expiresAt) {
+			expiredAuthCodes = append(expiredAuthCodes, k)
+		}
+	}
+
+	var expiredInvalidatedCodes []string
+	for k, v := range s.invalidatedCodes {
+		if now.After(v.expiresAt) {
+			expiredInvalidatedCodes = append(expiredInvalidatedCodes, k)
+		}
+	}
+
+	var expiredAccessTokens []string
+	for k, v := range s.accessTokens {
+		if now.After(v.expiresAt) {
+			expiredAccessTokens = append(expiredAccessTokens, k)
+		}
+	}
+
+	var expiredRefreshTokens []string
+	for k, v := range s.refreshTokens {
+		if now.After(v.expiresAt) {
+			expiredRefreshTokens = append(expiredRefreshTokens, k)
+		}
+	}
+
+	var expiredPKCERequests []string
+	for k, v := range s.pkceRequests {
+		if now.After(v.expiresAt) {
+			expiredPKCERequests = append(expiredPKCERequests, k)
+		}
+	}
+
+	var expiredUpstreamTokens []string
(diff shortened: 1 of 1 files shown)


JAORMX:
@claude please do a thorough security review of this, leverage agents if possible.

pkg/authserver/storage/memory.go
```diff
@@ -0,0 +1,837 @@
...
+		case <-ticker.C:
+			s.cleanupExpired()
+		}
+	}
+}
+
+// cleanupExpired removes all expired entries from storage.
+//
+//nolint:gocyclo // Function is straightforward, just repetitive cleanup loops for each storage type
+func (s *MemoryStorage) cleanupExpired() {
```
JAORMX, line 162: How often is this called, and, is it called concurrently with other functions that acquire the lock? This seems super expensive and even dangerous from a concurrency perspective, since it holds the lock a long damn time. What do you think?

pkg/authserver/storage/memory.go
```diff
@@ -0,0 +1,837 @@
...
+}
+
+// InvalidateAuthorizeCodeSession marks an authorization code as used/invalid.
+// Subsequent calls to GetAuthorizeCodeSession will return ErrInvalidatedAuthorizeCode.
+func (s *MemoryStorage) InvalidateAuthorizeCodeSession(_ context.Context, code string) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if _, ok := s.authCodes[code]; !ok {
+		logger.Debugw("authorization code not found for invalidation", "code", code)
```
JAORMX: Is this intended?

pkg/authserver/storage/memory.go
```diff
@@ -0,0 +1,837 @@
...
+
+// LoadPendingAuthorization retrieves a pending authorization by internal state.
+// Returns a defensive copy to prevent aliasing issues.
+func (s *MemoryStorage) LoadPendingAuthorization(_ context.Context, state string) (*PendingAuthorization, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	entry, ok := s.pendingAuthorizations[state]
+	if !ok {
+		logger.Debugw("pending authorization not found", "state", state)
```
JAORMX: is it safe to print the state to the logs?

pkg/authserver/storage/memory.go
jhrozek: I thought having these in debug is not so bad, since it's opt-in and debug is only ever enabled during development, but past my development I also didn't see a reason to keep it, removed

pkg/authserver/storage/memory.go
jhrozek: state is, code (which you commented on separately) is worse :-) removed either way

pkg/authserver/storage/memory.go
jhrozek: every 5 minutes by default and yes, it would block the other functions.

I refactored the cleanup to collect the expired items separately holding only a read-only lock and then clean up the expired ones with a rw lock, that should speed things up.

Either way, given that all these operations are in-memory I'm not terribly concerned about lock congestion.

JAORMX (approved):

PR #3034, opened 2025-12-15
Fix workload restart API by using background context

eleftherias:
## Issue

  API restart endpoint (`/api/v1beta/workloads/{name}/restart`) was returning 202 Accepted but workloads were getting stuck in "Stopped" state instead of restarting. 

## Root Cause

The API was passing the HTTP request context `r.Context()` to restart operations. Since the API returned 202 immediately by design, the HTTP request completed and its context got cancelled, causing the background restart goroutines to be cancelled mid-operation.
This was not a problem before, because `RestartWorkloads` was not using the provided context. Once `RestartWorkloads` started using the context, the bug appeared.

## Fix

 Changed both restart endpoints to use `context.Background()` instead of `r.Context()`, allowing restart operations to complete asynchronously after the HTTP response is sent. This preserves the intended async behaviour while ensuring restarts complete successfully.

PR #3301, opened 2026-01-15
Fix workload deletion API by using background context

amirejaz:
## Problem
When deleting workloads through the API, we're getting errors:
```
 failed to acquire lock for workload <name>: context canceled 
```

This happens in `setWorkloadStatusInternal` when trying to acquire a file lock.

## Root Cause
The `deleteWorkload` and `deleteWorkloadsBulk` endpoints were using the HTTP request context (`r.Context()`) for async operations. When the HTTP handler returns (after sending `202 Accepted`), the request context gets cancelled. The async deletion operation then tries to acquire locks using this cancelled context, causing the error.

## Solution
Changed both delete endpoints to use `context.Background()` instead of the request context, matching the pattern used by `restartWorkload` and `restartWorkloadsBulk` (fixed in #3034).

## Why This is Safe
- The timeout is still preserved: `deleteWorkload` creates a child context with `AsyncOperationTimeout` (5 minutes) via `context.WithTimeout(ctx, AsyncOperationTimeout)`
- This matches the established pattern for async operations that should continue after HTTP request completion
- The bug was present from the start (June 2025) but manifests intermittently based on timing and lock contention

## Testing
- [ ] Tested delete operation works without errors
- [ ] Verified timeout is still enforced (5 minutes)
- [ ] Confirmed behavior matches restart operations
## Files Changed:
pkg/api/v1/workloads.go (2 functions: deleteWorkload, deleteWorkloadsBulk)
Related:
Fixes the same issue pattern as #3034 (restart operations)

### DIFF
--- a/pkg/api/v1/workloads.go
+++ b/pkg/api/v1/workloads.go
@@ -210,11 +210,11 @@ func (s *WorkloadRoutes) restartWorkload(w http.ResponseWriter, r *http.Request)
 //	@Failure		404		{string}	string	"Not Found"
 //	@Router			/api/v1beta/workloads/{name} [delete]
 func (s *WorkloadRoutes) deleteWorkload(w http.ResponseWriter, r *http.Request) error {
-	ctx := r.Context()
 	name := chi.URLParam(r, "name")
 
 	// Use the bulk method with a single workload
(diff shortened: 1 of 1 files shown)


amirejaz:
closing this PR as #3302 already covers this fix.

PR #3306, opened 2026-01-15
Add OAuth2 authorization server configuration and JWKS utilities

jhrozek:
Implement the client-facing OAuth2 authorization server configuration using Ory Fosite's composable handler architecture. The core integration centers on AuthorizationServerConfig, a wrapper around fosite.Config that extends it with JWK/JWKS support for JWT signing. A Factory pattern enables pluggable handler composition where each factory receives the full config, storage, and strategy, returning handlers that are auto- registered based on which fosite interfaces they implement (AuthorizeEndpointHandler, TokenEndpointHandler, TokenIntrospector, RevocationHandler, PushedAuthorizeEndpointHandler). Security settings include enforced PKCE with S256-only per MCP specification, token lifespan bounds validation, and HMAC secret rotation support via RotatedGlobalSecrets for zero-downtime key changes.

This module serves as the configuration layer in the authserver's federation model, where ToolHive acts as an intermediary between MCP clients and upstream identity providers. MCP clients authenticate to this authorization server, which then federates authentication to upstream IDPs. AuthorizationServerConfig bridges AuthorizationServerParams (cryptographic materials, token lifespans, issuer URL) with the fosite framework, and is consumed by both the storage layer (fosite's authorization/token/PKCE handlers) and the HTTP handlers serving OAuth2 endpoints. PublicJWKS() enables safe key exposure at /.well-known/jwks.json for client token verification.

### DIFF
--- a/pkg/authserver/server/provider.go
+++ b/pkg/authserver/server/provider.go
@@ -101,6 +101,22 @@ func validateIssuerURL(issuer string) error {
 	return nil
 }
 
+// validateHMACSecrets validates that all HMAC secrets meet the minimum length requirement.
+func validateHMACSecrets(secrets *servercrypto.HMACSecrets) error {
+	if secrets == nil {
+		return fmt.Errorf("HMAC secrets are required")
+	}
+	if len(secrets.Current) < servercrypto.MinSecretLength {
+		return fmt.Errorf("current HMAC secret must be at least %d bytes", servercrypto.MinSecretLength)
+	}
+	for i, rotated := range secrets.Rotated {
+		if len(rotated) < servercrypto.MinSecretLength {
+			return fmt.Errorf("rotated HMAC secret [%d] must be at least %d bytes", i, servercrypto.MinSecretLength)
+		}
+	}
+	return nil
+}
+
 // NewAuthorizationServerConfig creates an AuthorizationServerConfig from the provided configuration.
 func NewAuthorizationServerConfig(cfg *AuthorizationServerParams) (*AuthorizationServerConfig, error) {
 	if cfg == nil {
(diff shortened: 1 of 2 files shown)


pkg/authserver/server/provider.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	if cfg.SigningKey == nil {
+		return nil, fmt.Errorf("signing key is required")
+	}
+	if cfg.HMACSecrets == nil {
+		return nil, fmt.Errorf("HMAC secrets are required")
+	}
+	if len(cfg.HMACSecrets.Current) < servercrypto.MinSecretLength {
+		return nil, fmt.Errorf("current HMAC secret must be at least %d bytes", servercrypto.MinSecretLength)
+	}
```
JAORMX, line 139: We validate the current secret, but not the rotated ones in `cfg.HMACSecrets.Rotated`. If someone constructs `HMACSecrets` directly (bypassing `LoadHMACSecrets`), they could pass weak rotated secrets that still get used for verification.

Worth adding a loop to check those too?

pkg/authserver/server/provider.go
```diff
@@ -0,0 +1,259 @@
...
+	return provider
+}
+
+// GetSigningKey returns the config's signing key.
+func (c *AuthorizationServerConfig) GetSigningKey(_ context.Context) *jose.JSONWebKey {
+	return c.SigningKey
+}
+
+// GetSigningJWKS returns the config's signing JWKS. This includes private keys.
+func (c *AuthorizationServerConfig) GetSigningJWKS(_ context.Context) *jose.JSONWebKeySet {
```
JAORMX: nit: The name `GetSigningJWKS` doesn't make it obvious this returns private keys. Someone might accidentally use this instead of `PublicJWKS()` for an endpoint.

Maybe rename to `GetPrivateSigningJWKS()` or make the comment louder?

pkg/authserver/server/provider.go
jhrozek: thanks, I added the rotated secrets only after the current secrets was wired through (after I had looked at what Ory Hydra does..) and forgot to copy over the checks.

pkg/authserver/server/provider.go
jhrozek: not a nit, people would see JWKS and think public keys, thanks

PR #3307, opened 2026-01-15
Fix workload update API by using background context for async operation

amirejaz:
## Problem
The `updateWorkload` endpoint was using the HTTP request context (`r.Context()`) for async operations. When the handler returns, the request context gets cancelled, which can cause failures in async operations like lock acquisition.

## Solution
Changed `updateWorkload` to use `context.Background()` instead of the request context, matching the pattern used by `deleteWorkload`, `restartWorkload`, and `stopWorkload` endpoints.

### DIFF
--- a/pkg/api/v1/workload_service.go
+++ b/pkg/api/v1/workload_service.go
@@ -97,7 +97,8 @@ func (s *WorkloadService) UpdateWorkloadFromRequest(ctx context.Context, name st
 	}
 
 	// Use the manager's UpdateWorkload method to handle the lifecycle
-	if _, err := s.workloadManager.UpdateWorkload(ctx, name, runConfig); err != nil {
+	// Use background context since this is async operation
+	if _, err := s.workloadManager.UpdateWorkload(context.Background(), name, runConfig); err != nil {
 		return nil, fmt.Errorf("failed to update workload: %w", err)
 	}
 
--- a/pkg/api/v1/workloads.go
+++ b/pkg/api/v1/workloads.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"time"
 
 	"github.com/go-chi/chi/v5"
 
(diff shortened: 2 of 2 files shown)


pkg/api/v1/workloads.go
```diff
@@ -296,7 +296,7 @@ func (s *WorkloadRoutes) createWorkload(w http.ResponseWriter, r *http.Request)
 //	@Failure		404			{string}	string	"Not Found"
 //	@Router			/api/v1beta/workloads/{name}/edit [post]
 func (s *WorkloadRoutes) updateWorkload(w http.ResponseWriter, r *http.Request) error {
-	ctx := r.Context()
+	ctx := context.Background()
```
JAORMX: If this is going to have a dedicated context, do we want to have a specific timeout + cancellation for this?

dmjb: In addition to @JAORMX 's comment, I would add two things:

1. Consider using the request context for everything except the async call (i.e. s.workloadService.UpdateWorkloadFromRequest)
2. When you create the context for the async call, add a comment explain why you are creating the new context.

amirejaz: > If this is going to have a dedicated context, do we want to have a specific timeout + cancellation for this?

Good point. We have `AsyncOperationTimeout` in manager layer which is the timeout for async workload operations. But as Don pointed out, I will reuse the request context for sync operations and create another background context with timeout for async operation.

pkg/api/v1/workload_service.go
```diff
@@ -97,7 +97,8 @@ func (s *WorkloadService) UpdateWorkloadFromRequest(ctx context.Context, name st
 	}
 
 	// Use the manager's UpdateWorkload method to handle the lifecycle
-	if _, err := s.workloadManager.UpdateWorkload(ctx, name, runConfig); err != nil {
+	// Use background context since this is async operation
+	if _, err := s.workloadManager.UpdateWorkload(context.Background(), name, runConfig); err != nil {
```
amirejaz, line 101: Using a timeout context here causes issues because UpdateWorkload is asynchronous. Deferring `cancel()` cancels the context as soon as the caller returns, which prematurely cancels the async operation.

The workload manager already owns and enforces the async timeout, so we can safely skip creating a timeout context at this call site.

amirejaz: cc: @JAORMX

PR #3311, opened 2026-01-15
Fix swagger generation for ImageMetadata embedded fields

lujunsan:
Remove yaml:",inline" tag from BaseServerMetadata embedding in
  ImageMetadata struct. The yaml-only tag caused swaggo v2.0.0-rc5
  to skip inlining the embedded fields in the OpenAPI schema, resulting
  in missing fields like name, description, tier, status, etc.

  Anonymous struct embedding without tags allows both proper JSON
  marshaling and correct swagger schema generation.


eleftherias (approved):

JAORMX:
this broke toolhive-registry.

PR #3312, opened 2026-01-15
feat(vmcp): integrate a dummy optimizer

jerm-dro:
# Summary 

This PR integrates a `DummyOptimizer` into the `vMCP` server. In a future PR, the dummy optimizer will be replaced by a real optimizer that relies on embedding search.

Closes https://github.com/StacklokLabs/research/issues/134
Closes https://github.com/StacklokLabs/research/issues/139
Closes https://github.com/StacklokLabs/research/issues/140

# Changes

The commits largely tell the story in an easy to digest sequence, with a bit of debugging and polish at the end.

At a high level:

1. **extract AddOnRegisterSession hook to standalone method** - This is a noop. In a later commit, I modify the contained function. It was already complex, so moving it to a separate method is intended to make the code a bit more maintainable.
2. **add optimizer interfaces and types** - the purpose of this PR
3. **implement a dummy optimizer** - this is necessary for e2e testing this PR and ensuring everything is wired up correctly.
4. **operator: deep copy config fields** - modify the CRD config converter to deep copy the embedded config. This means new config fields like the optimizer settings don't require an update of the converter.
5. **introduce a reflection module for producing JSON schema objects from structs** - Rather than hardcoding the JSON schema, allow it to be derived from the structs definition.
6. **integration of the optimizer** - on each session creation, if the optimizer is enabled, we'll create an optimizer containing all of the tools & composite tools. The only tools exposed to the client will be `find_tool` and `call_tool`.

## Testing

Unit tests have been added for the new modules. An e2 test proves the integration is correct by demonstrating finding and calling both a tool and a composite tool. This is currently backed by the `DummyOptimizer` but should continue to work when replaced with a real optimizer.

## Large PR Justification

This PR is large for a few reasons:
- it generates docs and operator files
- it demonstrates a minimal-viable integration of an optimizer with an e2e test.

Making this PR smaller would require not fully integrating or testing, which I believe would cause confusion / code churn.

### DIFF
--- a/deploy/charts/operator-crds/Chart.yaml
+++ b/deploy/charts/operator-crds/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: toolhive-operator-crds
 description: A Helm chart for installing the ToolHive Operator CRDs into Kubernetes.
 type: application
-version: 0.0.97
+version: 0.0.98
 appVersion: "0.0.1"
--- a/deploy/charts/operator-crds/README.md
+++ b/deploy/charts/operator-crds/README.md
@@ -1,6 +1,6 @@
 # ToolHive Operator CRDs Helm Chart
 
-![Version: 0.0.97](https://img.shields.io/badge/Version-0.0.97-informational?style=flat-square)
+![Version: 0.0.98](https://img.shields.io/badge/Version-0.0.98-informational?style=flat-square)
 ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square)
 
 A Helm chart for installing the ToolHive Operator CRDs into Kubernetes.


pkg/vmcp/optimizer/dummy_optimizer.go
```diff
@@ -0,0 +1,122 @@
...
+//
+// Returns all matching tools with a score of 1.0 (exact match semantics).
+// TokenMetrics are returned as zero values (not implemented in dummy).
+func (d DummyOptimizer) FindTool(_ context.Context, input FindToolInput) (*FindToolOutput, error) {
+	if input.ToolDescription == "" {
+		return nil, fmt.Errorf("tool_description is required")
+	}
+
+	// Log all tools in the optimizer for debugging
+	fmt.Printf("[DummyOptimizer.FindTool] Searching for %q in %d tools:\n", input.ToolDescription, len(d.tools))
```
jhrozek: The use of fmt.Printf is fine for a dummy commit (the optimizer is only enabled through config), but let's not have prints going forward once we're out of the dummy phase.

pkg/vmcp/optimizer/optimizer.go
```diff
@@ -0,0 +1,88 @@
...
+	CallTool(ctx context.Context, input CallToolInput) (*mcp.CallToolResult, error)
+}
+
+// FindToolInput contains the parameters for finding tools.
+type FindToolInput struct {
+	// ToolDescription is a natural language description of the tool to find.
+	ToolDescription string `json:"tool_description" description:"Natural language description of the tool to find"`
+
+	// ToolKeywords is an optional list of keywords to narrow the search.
+	ToolKeywords []string `json:"tool_keywords,omitempty" description:"Optional keywords to narrow search"`
```
jhrozek, line 41: ToolKeywords is probably something to be used later (just trying to solidify my understanding of the PR)

pkg/vmcp/server/server.go
```diff
@@ -119,6 +120,10 @@ type Config struct {
 	// Only set when running in K8s with outgoingAuth.source: discovered.
 	// Used for /readyz endpoint to gate readiness on cache sync.
 	Watcher Watcher
+
+	// OptimizerFactory builds an optimizer from a list of tools.
+	// If not set, the optimizer is disabled.
+	OptimizerFactory func([]server.ServerTool) optimizer.Optimizer
```
jhrozek, line 126: should either the factory return (Optimizer, err) or should the caller check for nil? Right now we do neither

pkg/vmcp/server/adapter/optimizer_adapter.go
```diff
@@ -0,0 +1,100 @@
...
+func createCallToolHandler(opt optimizer.Optimizer) func(context.Context, mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+	return func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		input, err := schema.Translate[optimizer.CallToolInput](request.Params.Arguments)
+		if err != nil {
+			return mcp.NewToolResultError(fmt.Sprintf("invalid arguments: %v", err)), nil
+		}
+
+		result, err := opt.CallTool(ctx, input)
+		if err != nil {
+			return mcp.NewToolResultError(fmt.Sprintf("call_tool failed: %v", err)), nil
```
jhrozek, line 81: more of a philosophical question - I assume returning the internal errors might lead to the optimizer "leaking" what tools it was calling etc. Now I don't think there's a security risk, but if that's something that should be exposed to the client through errors or just logged.

(I'm mostly wondering what do you think here, I can't decide either way)

pkg/vmcp/server/adapter/optimizer_adapter.go
```diff
@@ -0,0 +1,100 @@
...
+		return result, nil
+	}
+}
+
+// mustMarshalSchema marshals a schema to JSON, panicking on error.
+// This is safe because schemas are generated from known types at startup.
+func mustGenerateSchema[T any]() json.RawMessage {
+	s, err := schema.GenerateSchema[T]()
+	if err != nil {
+		panic(fmt.Sprintf("failed to generate schema: %v", err))
```
jhrozek, line 94: note to self or other reviewers - at first I was going to push back on panic() but mustGenerateSchema is 1) properly named and 2) called on package init through global variables so any bug would be caught by tests

pkg/vmcp/optimizer/optimizer.go
```diff
@@ -0,0 +1,88 @@
...
+type ToolMatch struct {
+	// Name is the unique identifier of the tool.
+	Name string `json:"name"`
+
+	// Description is the human-readable description of the tool.
+	Description string `json:"description"`
+
+	// Parameters is the JSON schema for the tool's input parameters.
+	// Uses json.RawMessage to preserve the original schema format.
+	Parameters json.RawMessage `json:"parameters"`
```
jhrozek: Question: Should parameters be optional to allow calling tools that take no parameters (however rare they might be)?

pkg/vmcp/optimizer/optimizer.go
```diff
@@ -0,0 +1,88 @@
...
+type FindToolOutput struct {
+	// Tools contains the matching tools, ranked by relevance.
+	Tools []ToolMatch `json:"tools"`
+
+	// TokenMetrics provides information about token savings from using the optimizer.
+	TokenMetrics TokenMetrics `json:"token_metrics"`
+}
+
+// ToolMatch represents a tool that matched the search criteria.
+type ToolMatch struct {
```
jhrozek, line 54: Actually let's see if I understand the flow:
 - the client calls `find_tool`
 - FindTool returns FindToolOutput with a []ToolMatch
 - that ToolMatch gets serialized to the client through `mcp.NewToolResultStructuredOnly`

in that case, is it expected that we use different terminology than the MCP protocol (parameters vs inputSchema that MCP uses, see e.g. https://modelcontextprotocol.io/specification/2025-03-26/server/tools#listing-tools)

pkg/vmcp/optimizer/optimizer.go
```diff
@@ -0,0 +1,88 @@
...
+	ToolKeywords []string `json:"tool_keywords,omitempty" description:"Optional keywords to narrow search"`
+}
+
+// FindToolOutput contains the results of a tool search.
+type FindToolOutput struct {
+	// Tools contains the matching tools, ranked by relevance.
+	Tools []ToolMatch `json:"tools"`
+
+	// TokenMetrics provides information about token savings from using the optimizer.
+	TokenMetrics TokenMetrics `json:"token_metrics"`
```
jhrozek, line 50: why is this in the find tool output (just asking, unfamiliar with how optimizer works)

pkg/vmcp/schema/reflect.go
```diff
@@ -0,0 +1,184 @@
+package schema
```
jhrozek, line 1: have you considered e.g. https://github.com/invopop/jsonschema ?

pkg/vmcp/optimizer/dummy_optimizer.go
jerm-dro: I removed this anyways.

pkg/vmcp/optimizer/optimizer.go
jerm-dro: Yea, ToolKeywords can optionally be used by the MCP client if it thinks some keywords are important for refining the search.

pkg/vmcp/server/server.go
jerm-dro: Right now, the dummy optimizer construction always succeeds, so there's no need to return an error.

In the future, if the real optimizer may produce an error, we'll update this type.

pkg/vmcp/server/adapter/optimizer_adapter.go
jerm-dro: I think exposing the error to the MCP client is important if you want it to correct its behavior. Without information on the failure, the model is pretty much hopeless in figuring out the problem.

pkg/vmcp/server/adapter/optimizer_adapter.go
jerm-dro: Added a comment to clarify the contract.

pkg/vmcp/server/adapter/optimizer_adapter.go
jerm-dro: Added a comment to explain reasoning

pkg/vmcp/optimizer/optimizer.go
jerm-dro: As your comment below points out, this is actually better off called "InputSchema," which is actually required.

pkg/vmcp/optimizer/optimizer.go
jerm-dro: I've updated this to be called input schema.

pkg/vmcp/optimizer/optimizer.go
jerm-dro: The research folks wanted this here. I'll make a note to ask why.

I suspect we can delete it if we eventually push the o11y into the optimizer, but I'll leave it here for now.

pkg/vmcp/schema/reflect.go
jerm-dro: Given the simplicity of our use case, I opted for rolling our own. If we were to elaborate on our implementation more, I could see us using a more robust implementation.

jhrozek (approved):

Issue #3178, opened 2026-01-02
List workloads API endpoint shows incorrect transport type


aponcedeleonch:
## Bug description

The `/api/v1beta/workloads` endpoint incorrectly reports the transport type for workloads, while querying individual workloads via `/api/v1beta/workloads/{name}` returns the correct transport type.

## Steps to reproduce

1. Deploy a workload with stdio transport (e.g., the time MCP server)
2. List all workloads: `curl http://localhost:8080/api/v1beta/workloads | jq`
3. Query the specific workload: `curl http://localhost:8080/api/v1beta/workloads/time | jq`

## Expected behavior

Both endpoints should report the same transport type. Since the container has MCP_TRANSPORT: stdio environment variable, the transport type should be stdio in both responses.

## Actual behavior

The list endpoint shows `"transport_type": "streamable-http"`:
```json
{
  "name": "time",
  "package": "docker.io/mcp/time:latest",
  "url": "http://127.0.0.1:10618/mcp",
  "port": 10618,
  "transport_type": "streamable-http",
  "proxy_mode": "streamable-http",
  "status": "running",
  ...
}
```

The individual workload endpoint correctly shows `"transport": "stdio"`:
```json
{
  "image": "docker.io/mcp/time:latest",
  ...
  "env_vars": {
    "MCP_TRANSPORT": "stdio"
  },
  "transport": "stdio",
  "proxy_mode": "streamable-http",
  ...
}
```

## Environment (if relevant)

```bash
% thv version
ToolHive v0.6.15
Commit: d390486ff968e9a2c0b79d5aca9112144e228b69
Built: 2025-12-17 19:58:30 UTC
Go version: go1.25.5
Platform: darwin/arm64
```

## Additional context
Any additional information or logs you think might help.

aponcedeleonch:
@Sanskarzz I assume you're volunteering to solve this issue, thanks a lot for the help! Will assign the issue to you but please feel free to un-assign yourself

Sanskarzz:
@aponcedeleonch  Yes, your assumption is correct. I’m happy to contribute to this project. Thanks for assigning me the issue!

eleftherias:
Closed via #3313

PR #3313, opened 2026-01-15
fix: correct transport type reporting in workload list endpoint

Sanskarzz:
Issue: #3178 

Description

This PR fixes a bug where the /api/v1beta/workloads (list) endpoint incorrectly reported the transport_type as streamable-http for workloads that were actually using stdio transport.

Changes

- Updated the `minimalRunConfig` struct in `pkg/workloads/types/types.go` to include the Transport field.
- Refactored `WorkloadFromContainerInfo` to:
       Load the actual transport type from the stored `RunConfig`. 
       Use state-of-truth transport type for both the domain model and URL generation.
       Fall back to labels only if the run configuration is missing.

### DIFF
--- a/pkg/runner/config_test.go
+++ b/pkg/runner/config_test.go
@@ -674,7 +674,7 @@ func TestRunConfig_WithStandardLabels(t *testing.T) {
 			expected: map[string]string{
 				"toolhive":           "true",
 				"toolhive-name":      "test-server",
-				"toolhive-transport": "sse", // Should be "sse" not "stdio"
+				"toolhive-transport": "stdio", // Should be "stdio" even when proxied
 				"toolhive-port":      "60000",
 			},
 		},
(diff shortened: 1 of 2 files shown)


pkg/workloads/types/types.go
```diff
@@ -17,8 +17,9 @@ import (
 
 // minimalRunConfig represents just the fields we need from a run configuration
 type minimalRunConfig struct {
-	Group     string `json:"group,omitempty" yaml:"group,omitempty"`
-	ProxyMode string `json:"proxy_mode,omitempty" yaml:"proxy_mode,omitempty"`
+	Group     string              `json:"group,omitempty" yaml:"group,omitempty"`
+	ProxyMode string              `json:"proxy_mode,omitempty" yaml:"proxy_mode,omitempty"`
+	Transport types.TransportType `json:"transport" yaml:"transport"`
```
yrobla: you miss the omitempty here, or is that a mandatory field?

yrobla: added some comments + it misses some testing

Sanskarzz: Done

amirejaz:
@Sanskarzz this looks good. There are some CI test failures and a few lint errors to address.

PR #3315, opened 2026-01-15
Add HTTP-based PDP authorizer

ghaskins:
IIntroduce a general-purpose authorization backend using HTTP-based Policy
Decision Points (PDPs). This authorizer can work with any PDP server that
implements the PORC (Principal-Operation-Resource-Context) decision endpoint.

Key features:
- HTTP client for connecting to PDP servers via /decision endpoint
- PORC mapping for MCP requests (Principal, Operation, Resource, Context)
- Configurable context inclusion (args, operation metadata)
- JWT claim extraction for principal attributes (roles, groups, scopes)

The authorizer uses a simple API contract:
- POST /decision with PORC JSON body
- Response: {"allow": true/false}

Compatible with Manetu PolicyEngine (MPE) and any custom PDP implementing
the same API.

## Large PR Justification

Multiple related changes that would break if separated

### DIFF
--- a/test/e2e/http_pdp_authz_test.go
+++ b/test/e2e/http_pdp_authz_test.go
@@ -271,7 +271,7 @@ var _ = Describe("HTTP PDP Authorization", Label("middleware", "authz", "http-pd
 				// This should fail because query_vulnerabilities_batch is not authorized
 				_, err := mcpClient.CallTool(ctx, "query_vulnerabilities_batch", arguments)
 				Expect(err).To(HaveOccurred(), "Should fail to call unauthorized tool")
-				Expect(err.Error()).To(ContainSubstring("authorization"), "Error should mention authorization")
+				Expect(err.Error()).To(ContainSubstring("Unauthorized"), "Error should mention Unauthorized")
 
 				GinkgoWriter.Printf("Expected authorization failure for unauthorized tool: %v\n", err)
 			})


JAORMX (commented):
Do you mind if we generalize this a little bit more? I am really keen on this authorizer because it's so general purpose. Having an HTTP-based authorizer is something that's applicable quite generally and could potentially be used with other PDPs that would respect the same API signature.

So... Tell me what you think about this:

* Let's rename this authorizer to something more general like: `httpv1` or something of the sort.
* We shall keep the PORC mappings and... basically keep this same implementation.
* Let's remove the manetu MPE policy domain examples from this particular PR to keep it constrained and smaller

There are some particular pieces that are tied to MPE (e.g. probe mode), but that's fine, we can keep those in this PR and generalize later.

What do you think?

Regarding the MPE policy domain samples: I want to find a good place for folks to view an e2e sample of this, and there we could add the MPE policy samples. What do you think?

ghaskins:
@JAORMX I pushed an update with your suggested changes

docs/authz.md
```diff
@@ -368,6 +369,157 @@ This means that `forbid` policies take precedence over `permit` policies.
...
+### PORC mapping
+
+The HTTP PDP authorizer uses the PORC (Principal-Operation-Resource-Context)
+model for authorization decisions. ToolHive automatically maps MCP requests to
+PORC:
+
+| MCP Concept | PORC Field | Format |
+|-------------|------------|--------|
+| Client identity | `principal.sub` | From JWT `sub` claim |
+| Roles | `principal.mroles` | From JWT `roles` or `mroles` claim |
```
yrobla: why mroles or mgroups instead of roles/groups?

docs/authz.md
```diff
@@ -368,6 +369,157 @@ This means that `forbid` policies take precedence over `permit` policies.
...
+| Groups | `principal.mgroups` | From JWT `groups` or `mgroups` claim |
+| Scopes | `principal.scopes` | From JWT `scope` or `scopes` claim |
+| MCP operation | `operation` | `mcp:<feature>:<operation>` (e.g., `mcp:tool:call`) |
+| MCP resource | `resource` | `mrn:mcp:<server>:<feature>:<id>` (e.g., `mrn:mcp:myserver:tool:weather`) |
```
yrobla, line 519: can you add some comment explaining this format, some link to documentation, etc?

pkg/authz/authorizers/http/config.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// ContextConfig configures what context information is included in the PORC.
+// All options default to false, meaning no MCP context is included by default.
+type ContextConfig struct {
+	// IncludeArgs enables inclusion of tool/prompt arguments in context.mcp.args.
+	// Default is false.
+	IncludeArgs bool `json:"include_args,omitempty" yaml:"include_args,omitempty"`
+
+	// IncludeOperation enables inclusion of MCP operation metadata in context.mcp:
```
yrobla, line 47: what will happen if we add some context.mcp, but we do not include operation? will it break? is that covered?

docs/authz.md
ghaskins: MPE is derived from our platform product, and our platform product had a scheme where any Manetu-specific claims added to the JWT had an 'm' prefix (mroles, mclearance, mannotations, etc).  It probably made more sense when this was an MPE-specific patch, but now that it's moving in a "it's generic HTTP" direction, it stands out more.  Not sure what the best solution is, with the current MPE they need those prefixes to work.

docs/authz.md
ghaskins: Will do (we have an explainer here: https://manetu.github.io/policyengine/concepts/mrn).  Ill add this to the docs

pkg/authz/authorizers/http/config.go
ghaskins: operation is a "mandatory phase", so if its missing, the access control decision will be a DENY and the audit-trail will reflect that at least part of the reason for the DENY was the missing operation.  There's some more info available here:

https://manetu.github.io/policyengine/concepts/policy-conjunction#phase-requirements

docs/authz.md
ghaskins: I should add:  As you will see in the note here: https://manetu.github.io/policyengine/concepts/mrn#mrn-format

The MRN format is optional.  All we really care about is unique strings.  We do encourage good organization, and the MRN scheme is an example of how to accomplish that.  Perhaps what I can do is come up with a generic way to talk about the organizational properties without using "mrn" in the cited examples

docs/authz.md
yrobla: i believe that this needs to be more abstract, and we could add some mappers between the specific implementations and the generic settings

pkg/authz/authorizers/http/config.go
yrobla: can this be documented as well?

docs/authz.md
ghaskins: Here's what I am thinking: originally, this PR was written to add "mpev1" to the config.  This was generalized to 'httpv1' in recognition that much of it is applicable to many contexts.  However, maybe I took it too far in the generalization.

So, what I am thinking is:

- most of the current logic for http-based authorizers remains as a reusable substrate
- we have an 'mpev1' type that uses that substrate but does the MPE specific mapping and can have a natural organization to the documentation (e.g. m prefixes needed, operation mandatory, etc).
- other systems may come along and also share the http-based authorizers code, but put whatever mappings, etc, they need in place in a similar way.

thoughts?

yrobla: Something like that could works:

We could use a ClaimMapper interface with type-specific implementations:                       
  - MPEClaimMapper for type: mpev1 - handles m-prefixed claims (mroles, mgroups, mannotations)          
  - OIDCClaimMapper for future type: oidcv1 - uses standard OIDC claims (roles, groups)                 
                                                                                                        
The mapper would be injected into PORCBuilder based on the authorizer type. This keeps the shared HTTP    
infrastructure (client, PORC builder, context config) generic while allowing each PDP type to have its
 own claim mapping conventions.

yrobla:
Hi @ghaskins how are you doing? I triggered some review for you some days ago, and i also see that the PR is failing in lint. I am approaching you to see the status of the work, if you are able to continue the collaboration. We can also take the issues on your behalf and put this PR on shape to be merged. Please keep us informed, thanks!

PR #3321, opened 2026-01-16
Add OAuth handler infrastructure and discovery endpoints

jhrozek:
This patch introduces the HTTP handler layer for the OAuth 2.0 authorization server. The Handler struct coordinates all OAuth/OIDC endpoints and provides route registration via chi router (consistent with ToolHive's API patterns). The discovery endpoints (/.well-known/openid-configuration and jwks.json) are fully implemented and OIDC Discovery 1.0 compliant, including the REQUIRED fields subject_types_supported and id_token_signing_alg_values_supported. The signing algorithms are dynamically extracted from the JWKS keys. OAuth endpoints (authorize, token, callback, register) are stubbed for future implementation.

In the full authserver, the Handler is instantiated by server_impl.go with four dependencies: a fosite.OAuth2Provider (the OAuth protocol engine), AuthorizationServerConfig (issuer, token lifespans, signing keys), Storage (where OAuth state is persisted), and an upstream.Provider (for relaying tokens to/from the upstream IDP). The Server.Handler() method returns the chi router as http.Handler, which can be mounted into any HTTP server. This design keeps the handler layer focused on HTTP concerns while delegating OAuth logic to fosite and state management to the storage layer.

### DIFF
--- a/pkg/registry/data/registry.json
+++ b/pkg/registry/data/registry.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/stacklok/toolhive/main/pkg/registry/data/toolhive-legacy-registry.schema.json",
   "version": "1.0.0",
-  "last_updated": "2026-01-19T00:19:53Z",
+  "last_updated": "2026-01-20T00:18:37Z",
   "servers": {
     "adb-mysql-mcp-server": {
       "description": "Official MCP server for AnalyticDB for MySQL of Alibaba Cloud",
@@ -896,7 +896,7 @@
         "automation",
         "browser"
       ],
-      "image": "ghcr.io/stacklok/dockyard/npx/brightdata-mcp:2.7.1",
+      "image": "ghcr.io/stacklok/dockyard/npx/brightdata-mcp:2.8.3",
       "permissions": {
         "network": {
           "outbound": {
@@ -1766,7 +1766,7 @@
       "metadata": {
         "stars": 20,
         "pulls": 12390,
-        "last_updated": "2025-12-11T02:36:44Z"
+        "last_updated": "2026-01-19T02:44:55Z"
       },
       "repository_url": "https://github.com/stackloklabs/gofetch",
       "tags": [
@@ -2083,7 +2083,7 @@
         "update",
         "issues"
       ],
-      "image": "ghcr.io/github/github-mcp-server:v0.28.1",
+      "image": "ghcr.io/github/github-mcp-server:v0.29.0",
       "permissions": {
         "network": {
           "outbound": {
@@ -2647,9 +2647,9 @@
         "get_error_log"
       ],
       "metadata": {
-        "stars": 242,
+        "stars": 260,
         "pulls": 17082,
-        "last_updated": "2025-12-11T02:36:44Z"
+        "last_updated": "2026-01-19T02:44:55Z"
       },
       "repository_url": "https://github.com/voska/hass-mcp",
       "tags": [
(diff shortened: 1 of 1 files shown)


pkg/authserver/server/handlers/discovery.go
```diff
@@ -0,0 +1,150 @@
...
+	if err := json.NewEncoder(w).Encode(publicJWKS); err != nil {
+		logger.Errorw("failed to encode JWKS",
+			"error", err.Error(),
+		)
+		http.Error(w, "internal server error", http.StatusInternalServerError)
+		return
+	}
+}
+
+// OIDCDiscoveryHandler handles GET /.well-known/openid-configuration requests.
```
tgrunnagle: Question: Do we also want to support `/.well-known/oauth-authorization-server` in case the client is not using OIDC discovery?

tgrunnagle (approved):

pkg/authserver/server/handlers/discovery.go
```diff
@@ -0,0 +1,150 @@
...
+	DefaultJWKSCacheMaxAge = 3600
+
+	// DefaultDiscoveryCacheMaxAge is the Cache-Control max-age for the discovery endpoint (1 hour).
+	// Aligned with Google's OIDC discovery cache policy.
+	DefaultDiscoveryCacheMaxAge = 3600
+)
+
+// OIDCDiscoveryDocument represents the OIDC discovery document structure.
+// Implements OpenID Connect Discovery 1.0 specification.
+type OIDCDiscoveryDocument struct {
```
JAORMX: Don't we have this in another package?

yrobla: i just saw the same duplication, is that needed or can it be reconciled?

pkg/authserver/server/handlers/discovery.go
yrobla: we also managed in vmcp/server, right?

yrobla:
added some comments... there is duplication , also i do not see integration with the current flow, is that intentional?

pkg/authserver/server/handlers/discovery.go
jhrozek: I'll do it once the move to the shared structures is completed.

pkg/authserver/server/handlers/discovery.go
jhrozek: we do. We can't easily use the same code (the existing code is client-side, authserver adds server-side), but we can and should converge on a single set of structures. Thanks, good call, I'll open a new PR.

pkg/authserver/server/handlers/discovery.go
jhrozek: in vmcp we only delegate to handlers from pkg/auth

jhrozek: sorry, I think you meant something else the well-known endpoint in pkg/auth is a protected resource endpoint ("where do I get a token to access this MCP server") while authserver adds an endpoint that says "this is how you get a token" (e.g. is PKCE support enabled, the token endpoint is blah)

jhrozek: In general I think you have a point that the RFC-level structure should be moved to pkg/auth, let me know if I should do it in PR #3349

jhrozek:
> added some comments... there is duplication 

PR #3349 addresses the structure-level duplication

> also i do not see integration with the current flow, is that intentional?

yes, this is a new component, integration will come up gradually

pkg/authserver/server/handlers/discovery.go
jhrozek: done in a separate patch

dmjb (approved):

PR #3323, opened 2026-01-16
Add EmbeddingServer CRD for deploying HuggingFace embedding models

ptelang:
## Summary

This PR introduces the `EmbeddingServer` Custom Resource Definition (CRD) for the ToolHive Kubernetes Operator, enabling deployment and management of HuggingFace Text Embeddings Inference (TEI) servers.

### Key Features

- **New CRD**: `EmbeddingServer` for deploying HuggingFace Text Embeddings Inference servers in Kubernetes
- **Model Management**: Supports downloading and caching embedding models from HuggingFace
- **Persistent Storage**: Optional PVC support for model caching to speed up restarts after initial download
- **Resource Configuration**: Full control over deployment resources, environment variables, and pod templates
- **Group Integration**: Can be associated with MCPGroups for organization and management
- **Status Management**: Comprehensive condition-based status tracking:
  - `ModelReady`: Indicates whether the embedding model is downloaded and ready
  - `VolumeReady`: Indicates whether the PVC for model caching is ready
  - `ImageValidated`: Validates the container image
  - `PodTemplateValid`: Validates the pod template configuration

### Components Added

- **CRD Definition**: `cmd/thv-operator/api/v1alpha1/embeddingserver_types.go`
- **Controller**: `cmd/thv-operator/controllers/embeddingserver_controller.go`
- **Unit Tests**: Comprehensive test coverage for the controller
- **Example Configurations**:
  - `basic-embedding.yaml`: Minimal configuration with default settings
  - `embedding-with-cache.yaml`: Configuration with persistent model cache (10Gi PVC)
  - `embedding-advanced.yaml`: Advanced setup with custom resources and environment variables
- **Documentation**: README with detailed examples and usage instructions

## Test plan

- [x] Unit tests for EmbeddingServer controller
- [x] Example YAML configurations validated
- [x] E2E operator tests (Chainsaw) for EmbeddingServer CRD
- [x] Manual testing of embedding server deployment in Kubernetes cluster
- [ ] Verification of model caching with persistent volumes

## Large PR Justification

Co-Authored-By: Claude <noreply@anthropic.com>


cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
```diff
@@ -0,0 +1,269 @@
...
+	// +kubebuilder:default=8080
+	Port int32 `json:"port,omitempty"`
+
+	// Args are additional arguments to pass to the embedding inference server
+	// +optional
+	Args []string `json:"args,omitempty"`
+
+	// Env are environment variables to set in the container
+	// +optional
+	Env []EnvVar `json:"env,omitempty"`
```
jerm-dro, line 74: Blocking: I believe on the original PR you mentioned that Huggingface access tokens would be provided through the `env`. I'd suggest instead:

```go
// HFTokenSecretRef is a reference to a Kubernetes Secret containing the huggingface token.
// If provided, the secret value will be provided to the embedding server for authentication with huggingface.
// +optional
HFTokenSecretRef SecretKeyRef
```

This will be more secure than storing the tokens in plaintext within the env vars. This also follows our conventions where other CRDs need to reference secret values.

cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
```diff
@@ -0,0 +1,269 @@
...
+	// ResourceMetadataOverrides is embedded to inherit annotations and labels fields
+	ResourceMetadataOverrides `json:",inline"` // nolint:revive
+
+	// PodTemplateMetadataOverrides defines metadata overrides for the pod template
+	// +optional
+	PodTemplateMetadataOverrides *ResourceMetadataOverrides `json:"podTemplateMetadataOverrides,omitempty"`
+
+	// Env are environment variables to set in the embedding container
+	// +optional
+	Env []EnvVar `json:"env,omitempty"`
```
jerm-dro: Nit: this appears to be duplicated by the `Env` on `EmbeddingServerSpec`

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,985 @@
...
+	ctx context.Context,
+	embedding *mcpv1alpha1.EmbeddingServer,
+) (ctrl.Result, error) {
+	// Validate PodTemplateSpec early
+	if !r.validateAndUpdatePodTemplateStatus(ctx, embedding) {
+		return ctrl.Result{}, nil
+	}
+
+	// Validate image
+	if err := r.validateImage(ctx, embedding); err != nil {
```
jerm-dro, line 146: nit: log the ignored error so we know why it's failing

test/e2e/chainsaw/operator/single-tenancy/test-scenarios/embeddingserver/with-cache/embeddingserver.yaml
```diff
@@ -0,0 +1,27 @@
...
+  model: "sentence-transformers/all-MiniLM-L6-v2"
+  image: "ghcr.io/huggingface/text-embeddings-inference:cpu-1.5"
+  imagePullPolicy: IfNotPresent
+  port: 8080
+  replicas: 1
+  # Enable model caching
+  modelCache:
+    enabled: true
+    size: "5Gi"
+    accessMode: "ReadWriteOnce"
```
jerm-dro, line 17: praise: I like that you also tested the model caching functionality 😄

test/e2e/chainsaw/operator/single-tenancy/test-scenarios/embeddingserver/with-cache/chainsaw-test.yaml
```diff
@@ -0,0 +1,108 @@
...
+          if [ "$PVC_STATUS" != "Bound" ]; then
+            echo "PVC is not bound. Current status: $PVC_STATUS"
+            kubectl describe pvc $PVC_NAME -n toolhive-system
+            exit 1
+          fi
+
+          echo "✓ PVC is bound"
+
+          # Verify the volume is mounted in the deployment
+          VOLUME_MOUNTED=$(kubectl get deployment $DEPLOYMENT_NAME -n toolhive-system -o jsonpath='{.spec.template.spec.volumes[?(@.persistentVolumeClaim.claimName=="'$PVC_NAME'")].name}' 2>/dev/null || echo "")
```
jerm-dro: Question, non-blocking: could we also check that some model files are written to the volume?

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,1031 @@
...
+	ctxLogger := log.FromContext(ctx)
+
+	// Try to update the deployment
+	err := r.Update(ctx, deployment)
+	if err == nil {
+		return nil
+	}
+
+	// If it's a conflict error, fetch the latest version and try again
+	if errors.IsConflict(err) {
```
jerm-dro: nit: I'd recommend deleting the logic here and returning the error. Returning the error will cause the reconciler loop to run again, which will automatically get the latest version of everything.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,1031 @@
...
+		})
+	}
+
+	return nil
+}
+
+// pvcForEmbedding creates a PVC for the embedding model cache
+func (r *EmbeddingServerReconciler) pvcForEmbedding(embedding *mcpv1alpha1.EmbeddingServer) *corev1.PersistentVolumeClaim {
+	pvcName := fmt.Sprintf("%s-model-cache", embedding.Name)
+
```
jerm-dro: Blocking: I think we'll be better off treating the `EmbeddingServer` as a `StatefulSet`. If we treat it as a deployment and then start attaching PVCs to it, I can imagine we run into some problems when scaling. Multiple replicas will try to mount the same PVC [1]. I can expect that leads to problems where `ReadWriteMany` is not supported. Even if it is supported, it could also cause races during the startup of multiple embedding servers.

Yes, this means the standard way of scaling an `EmbeddingServer` will result in copies of the model, but that shouldn't be much space in the grand scheme and can be written to a cheap storage class, since it's ready only once at startup. 

[1] https://stackoverflow.com/questions/63695888/how-does-pvc-actually-work-when-its-attached-to-a-deployment-with-multiple-repl

jerm-dro (changes requested):
This mostly looks good, but I have a few blocking changes: 
- Let's use a Secret to plumb huggingface tokens into the EmbeddingServer
- Let's treat the server as a StatefulSet rather than attaching PVCs to a deployment

ChrisJBurns:
@ptelang  Thanks for this. Is there any way we can split this into multiple PRs? 4k diffs is quite a large change and I feel folks will miss stuff in the cracks due to trying to review the entire thing. I've also noticed that the controller has 1k lines of code but only 300 lines of unit tests. Is there a way we can increase the test coverage to ensure the code does what we want?

ChrisJBurns:
@claude  Can you please review this PR using the necessary subagents in this repository. Leverage your knowledge of software engineering best practices (such as SOLID / DRY etc) and Kubernetes controller in order to support your review.

cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
ptelang: Good call! HFTokenSecretRef is now added.

cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
ptelang: Good catch! It is a duplicate. I removed it.

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Done

test/e2e/chainsaw/operator/single-tenancy/test-scenarios/embeddingserver/with-cache/chainsaw-test.yaml
ptelang: Done

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Done

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Good point. EmbeddingServer is now a StatefulSet.

ptelang:
@ChrisJBurns 
> Is there any way we can split this into multiple PRs? 4k diffs is quite a large change and I feel folks will miss stuff in the cracks due to trying to review the entire thing. 

I will split the PRs tomorrow. 


> I've also noticed that the controller has 1k lines of code but only 300 lines of unit tests. Is there a way we can increase the test coverage to ensure the code does what we want?

Good point. I added more unit tests.

cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
ptelang: Good call! HFTokenSecretRef is added.

cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
```diff
@@ -0,0 +1,272 @@
...
+	// +kubebuilder:validation:Enum=ReadWriteOnce;ReadWriteMany;ReadOnlyMany
+	// +optional
+	AccessMode string `json:"accessMode,omitempty"`
+}
+
+// EmbeddingResourceOverrides defines overrides for annotations and labels on created resources
+type EmbeddingResourceOverrides struct {
+	// Deployment defines overrides for the Deployment resource
+	// +optional
+	Deployment *EmbeddingDeploymentOverrides `json:"deployment,omitempty"`
```
jerm-dro: Suggestion: Since we've changed the embedding server to a statefulset, I think this should be renamed to match.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,1162 @@
...
+		return ctrl.Result{}, nil
+	} else if err != nil {
+		ctxLogger.Error(err, "Failed to get StatefulSet")
+		return ctrl.Result{}, err
+	}
+
+	// Ensure the statefulset size matches the spec
+	desiredReplicas := embedding.GetReplicas()
+	if *statefulSet.Spec.Replicas != desiredReplicas {
+		statefulSet.Spec.Replicas = &desiredReplicas
```
jerm-dro, line 235: musings, non-blocking: it's quite unfortunate we are manually comparing objects. I looked into this more and what you're doing here is not different from `MCPServerReconciler.deploymentNeedsUpdate`. In general, I think this opens us up to all sorts of "I modified X but Y didn't update" bugs. However, you're following patterns, so I don't think you need to change this PR.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,1162 @@
...
+}
+
+// updateStatefulSetWithRetry updates the statefulset
+// The reconciler loop will automatically retry on conflicts
+func (r *EmbeddingServerReconciler) updateStatefulSetWithRetry(
+	ctx context.Context,
+	statefulSet *appsv1.StatefulSet,
+) error {
+	return r.Update(ctx, statefulSet)
+}
```
jerm-dro: nit, non-blocking: this function is unnecessary

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,1162 @@
...
+	r.mergePodTemplateSpec(embedding, &podTemplate)
+
+	return podTemplate
+}
+
+// mergePodTemplateSpec merges user-provided PodTemplateSpec customizations
+func (r *EmbeddingServerReconciler) mergePodTemplateSpec(
+	embedding *mcpv1alpha1.EmbeddingServer,
+	podTemplate *corev1.PodTemplateSpec,
+) {
```
jerm-dro, line 722: Non-blocking, musings: This merging code is complex, but I don't think this PR is doing anything terribly out of the usual. Some of our CRDs support `podTemplateSpec`, but they all have their own implementations of this merge logic:

**VMCP**: https://github.com/stacklok/toolhive/blob/main/cmd/thv-operator/controllers/virtualmcpserver_deployment.go#L697-L749
**Registry**: https://github.com/stacklok/toolhive/blob/main/cmd/thv-operator/pkg/registryapi/podtemplatespec.go#L471-L522


In future work, we should figure out how to consolidate on one implementation, so we can have well-tested, predictable behavior.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,1162 @@
...
+	// Check if the number of replicas changed
+	desiredReplicas := embedding.GetReplicas()
+	if *statefulSet.Spec.Replicas != desiredReplicas {
+		return true
+	}
+
+	// Compare containers by checking specific important fields
+	if len(statefulSet.Spec.Template.Spec.Containers) != 1 {
+		return true
+	}
```
jerm-dro: Blocking: This check seems incorrect. It's possible someone adds a sidecar in the pod template spec - meaning we have 2 containers. As written, this code would always require an update to such a stateful set.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,1162 @@
...
+			existingEnvMap[env.Name] = env.Value
+		}
+	}
+
+	if !reflect.DeepEqual(expectedEnvMap, existingEnvMap) {
+		return true
+	}
+
+	// Check HF_TOKEN secret reference
+	expectedHFTokenRef := embedding.Spec.HFTokenSecretRef
```
jerm-dro: Blocking: I think this code becomes much simpler if you use `statefulSetForEmbedding` first:
```
newStatefulSet := r.statefulSetForEmbedding(ctx, embedding)
if !reflect.DeepEqual(newStatefulSet.Annotations, currentSts.Annotations) {
   return true
}
// and so on...
```

Most of these comparisons above and below turn into equality/deep equality checks.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -0,0 +1,1162 @@
...
+			embedding.Status.Message = "Embedding server is running"
+		} else if statefulSet.Status.Replicas > 0 && statefulSet.Status.ReadyReplicas == 0 {
+			// Check if pods are downloading the model
+			embedding.Status.Phase = mcpv1alpha1.EmbeddingServerPhaseDownloading
+			embedding.Status.Message = "Downloading embedding model"
+		} else {
+			embedding.Status.Phase = mcpv1alpha1.EmbeddingServerPhasePending
+			embedding.Status.Message = "Waiting for statefulset"
+		}
+	}
```
jerm-dro, line 1033: non-blocking, for next time: In go, an idiomatic way of re-writing:
```go
var foo T
if someCondition {
   foo = something
} else {
   foo = something else
}
```

is to use a anonymous function:
```go
foo := func () T {
   if someCondition {
       return something
   } 
   return something else
}() // immediately executed to get the result
```

The benefits are: 1) foo is treated as immutable, so fewer possibilities of bugs and 2) we can avoid complex nesting/branching.

cmd/thv-operator/test-integration/embedding-server/embeddingserver_creation_test.go
```diff
@@ -0,0 +1,1179 @@
...
+						},
+					},
+					Spec: corev1.ServiceSpec{
+						Ports: []corev1.ServicePort{{Port: 8080}},
+					},
+				},
+			},
+		},
+		// TODO(embeddingserver): Update assertion when ResourceOverrides on pod template is implemented.
+		// Expected: Annotations: {"pod-annotation": "pod-value"}, Labels: {"pod-label": "pod-value"} on pod template
```
jerm-dro: Can't this be deleted? It looks like the FinalState is correct.

test/e2e/chainsaw/operator/multi-tenancy/test-scenarios/embeddingserver/README.md
```diff
@@ -0,0 +1,157 @@
+# EmbeddingServer Multi-Tenancy E2E Tests
```
jerm-dro: Blocking: It doesn't appear other chainsaw tests have associated READMEs. Mind deleting this?

test/e2e/chainsaw/operator/single-tenancy/test-scenarios/embeddingserver/README.md
```diff
@@ -0,0 +1,155 @@
+# EmbeddingServer E2E Tests
```
jerm-dro: Blocking: Similar to comment above. Please delete.

cmd/thv-operator/api/v1alpha1/embeddingserver_types.go
ptelang: Sorry, this is left over from the change we made. I fixed this and other places where embedding server related fields were using "deployment".

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Ack

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Good call, function removed!

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Good catch! Now, fixed. Changed the logic to search for the embedding container by name.

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Should we create a followup issue to track this?

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Yes, updated now.

cmd/thv-operator/test-integration/embedding-server/embeddingserver_creation_test.go
ptelang: Yes, done

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: I modified the code per your suggestion and also added this guideline in CLAUDE.md file.

test/e2e/chainsaw/operator/multi-tenancy/test-scenarios/embeddingserver/README.md
ptelang: File deleted.

test/e2e/chainsaw/operator/single-tenancy/test-scenarios/embeddingserver/README.md
ptelang: File deleted.

cmd/thv-operator/controllers/embeddingserver_controller.go
jerm-dro: There's an open RFC for improving some CRDs: https://github.com/stacklok/toolhive-rfcs/pull/23

It's on my plate to more closely review that RFC. I think we may also want to add an item about enabling reuse of implementations. If it doesn't fit in that PR, I will open up a separate/issue RFC.

CLAUDE.md
```diff
@@ -313,6 +313,40 @@ For the complete documentation structure and navigation, see `docs/arch/README.m
...
+  } else {
+      phase = PhaseDefault
+  }
+  ```
+
+  **Benefits**:
+  - The variable is immutable after assignment, reducing bugs from accidental modification
+  - All decision logic is in one place with explicit returns
+  - Clearer logic flow and easier to understand
+  - Reduces cognitive load from tracking which branch sets which value
```
jerm-dro, line 348: Praise: thanks for adding this 😄

cmd/thv-operator/controllers/embeddingserver_controller.go
ptelang: Ack

Issue #3267, opened 2026-01-12
Noisy warning when loading workloads with telemetry config


jerm-dro:
# Description

Users who upgrade ToolHive from an older version encounter noisy warnings when running commands like `thv run`, `thv restart`, `thv rm`, or `thv ls`:

```
WARN  failed to load state for fetch: json: cannot unmarshal number into Go struct field Config.telemetry_config.samplingRate of type string
```

This warning appears repeatedly and can cause workload operations to fail entirely.

#### Cause

The `SamplingRate` field in `telemetry.Config` was changed from `float64` to `string` to support proper serialization. However, existing run config files created with older versions still store `samplingRate` as a number:

**Old format (created by older ToolHive versions):**
```json
{
  "telemetry_config": {
    "samplingRate": 0.1
  }
}
```

**New format (expected by current ToolHive):**
```json
{
  "telemetry_config": {
    "samplingRate": "0.1"
  }
}
```

When the new code tries to unmarshal the old config, Go's JSON decoder fails because it cannot convert a JSON number to a Go string field.

#### Impact

- `thv` commands against some old workflows with telemetry enabled can fail.
- Noisy warning logs 

#### Steps to Reproduce

1. Run an older version of ToolHive
2. Create a workload with telemetry enabled:
   ```
   thv run --name fetch --otel-endpoint http://localhost:4318 --otel-sampling-rate 0.1 fetch
   ```
3. Upgrade to a newer version of ToolHive
4. Run any command that loads the workload config:
   ```
   thv restart fetch
   ```
5. Observe the warning/error

#### Workaround

Manually edit the run config file at `~/Library/Application Support/toolhive/runconfigs/<workload>.json` and change `"samplingRate": 0.1` to `"samplingRate": "0.1"`.

#### Suggested Fix

Add a migration that runs at startup to automatically convert old config files from the float format to the string format.

jerm-dro:
I've noticed this issue still reproduces because of another instance of the old telemetry config elsewhere in the config struct. 

PR incoming.

PR #3328, opened 2026-01-17
fix(config): migrate another telemetry config

jerm-dro:
# Summary 

Fixes https://github.com/stacklok/toolhive/issues/3267

This PR adds another migration to finish migrating the other instance of the `telemetry.Config` in the runtime config. 

# Details

The original fix for the bug only migrated the `telemetry_config` field in the runtime config. However, it's also possible for `telemetry.Config` to exist in the `runtimeConfig.middleware_configs[i].parameters.config` as seen below:

```json
{
  "middleware_configs": [
    {
      "type": "telemetry",
      "parameters": {
        "config": {
          "endpoint": "",
          "serviceName": "toolhive-mcp-proxy",
          "serviceVersion": "v0.6.13",
          "tracingEnabled": true,
          "metricsEnabled": true,
          "samplingRate": 1,
          "headers": {},
          "insecure": true,
          "enablePrometheusMetricsPath": true,
          "environmentVariables": [
            "USER",
            "HOST"
          ]
        },
        "server_name": "mermaid",
        "transport": "streamable-http"
      }
    }
  ]
}
```

**How was this missed?**
The `parameters` field is an untyped json blob. 

# Changes
- refactor the telemetry migration so it runs against both `telemetry_config` and the `middleware_configs`.
- introduce another migration to ensure the `middleware_configs` is migrated in the event users have already upgraded
- add some tests to verify the migration of `middleware_configs` according to the structure above.

### DIFF
--- a/pkg/migration/middleware_telemetry.go
+++ b/pkg/migration/middleware_telemetry.go
@@ -3,6 +3,7 @@ package migration
 import (
 	"sync"
 
+	"github.com/stacklok/toolhive/pkg/config"
 	"github.com/stacklok/toolhive/pkg/logger"
 )
 
@@ -15,9 +16,23 @@ var middlewareTelemetryMigrationOnce sync.Once
 // It repeats performTelemetryConfigMigration, because an earlier iteration did not migrate middleware telemetry configs.
 func CheckAndPerformMiddlewareTelemetryMigration() {
 	middlewareTelemetryMigrationOnce.Do(func() {
+		// Check if migration was already performed
+		appConfig := config.NewDefaultProvider().GetConfig()
+		if appConfig.MiddlewareTelemetryMigration {
+			logger.Debugf("Telemetry config migration already completed, skipping")
+			return
+		}
+
 		if err := performTelemetryConfigMigration(); err != nil {
 			logger.Errorf("Failed to perform middleware telemetry migration: %v", err)
 			return
+		}
+
+		// Mark migration as completed
+		if err := config.UpdateConfig(func(c *config.Config) {
+			c.MiddlewareTelemetryMigration = true
+		}); err != nil {
+			logger.Errorf("Failed to update config after middleware telemetry migration: %v", err)
 		}
 	})
 }
--- a/pkg/migration/telemetry_config.go
+++ b/pkg/migration/telemetry_config.go
@@ -21,9 +21,23 @@ var telemetryMigrationOnce sync.Once
 // It handles both deprecated top-level telemetry_config and middleware-based telemetry configs.
 func CheckAndPerformTelemetryConfigMigration() {
 	telemetryMigrationOnce.Do(func() {
+		// Check if migration was already performed
+		appConfig := config.NewDefaultProvider().GetConfig()
+		if appConfig.TelemetryConfigMigration {
+			logger.Debugf("Telemetry config migration already completed, skipping")
+			return
+		}
+
 		if err := performTelemetryConfigMigration(); err != nil {
 			logger.Errorf("Failed to perform telemetry config migration: %v", err)
 			return
+		}
+
+		// Mark migration as completed
+		if err := config.UpdateConfig(func(c *config.Config) {
+			c.TelemetryConfigMigration = true
+		}); err != nil {
+			logger.Errorf("Failed to update config after telemetry config migration: %v", err)
 		}
 	})
 }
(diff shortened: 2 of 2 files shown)


jhrozek (commented):
Thanks for fixing the middleware telemetry config migration! I have a few questions about the implementation to make sure I understand the intended behavior.

pkg/migration/middleware_telemetry.go
```diff
@@ -0,0 +1,23 @@
...
+// middlewareTelemetryMigrationOnce ensures the middleware telemetry migration only runs once
+var middlewareTelemetryMigrationOnce sync.Once
+
+// CheckAndPerformMiddlewareTelemetryMigration checks if middleware telemetry migration is needed and performs it.
+// This migration ensures middleware-based telemetry configs are properly migrated.
+// It handles any additional middleware telemetry configuration migrations beyond the samplingRate conversion.
+// It repeats performTelemetryConfigMigration, because an earlier iteration did not migrate middleware telemetry configs.
+func CheckAndPerformMiddlewareTelemetryMigration() {
+	middlewareTelemetryMigrationOnce.Do(func() {
+		if err := performTelemetryConfigMigration(); err != nil {
```
jhrozek, line 26: Question about migration behavior for existing users:

I want to make sure I understand the intended behavior here. This calls `performTelemetryConfigMigration()`, which checks `TelemetryConfigMigration` flag at the start:

```go
if appConfig.TelemetryConfigMigration {
    logger.Debugf("Telemetry config migration already completed, skipping")
    return nil
}
```

For users who already ran an older version of ToolHive, `TelemetryConfigMigration` would already be `true` (set by the earlier migration that only handled top-level config). This means the migration would return immediately without migrating middleware configs.

Is this the intended behavior? It seems like existing users who need their middleware configs migrated won't get them migrated because the flag check happens before the migration logic runs.

If this should work for existing users, would the fix be to check `MiddlewareTelemetryMigration` instead (or in addition to) `TelemetryConfigMigration`?

pkg/config/config.go
```diff
@@ -24,21 +24,22 @@ const lockTimeout = 1 * time.Second
...
-	BuildAuthFiles           map[string]string   `yaml:"build_auth_files,omitempty"`
+	Secrets                      Secrets             `yaml:"secrets"`
+	Clients                      Clients             `yaml:"clients"`
+	RegistryUrl                  string              `yaml:"registry_url"`
+	RegistryApiUrl               string              `yaml:"registry_api_url"`
+	LocalRegistryPath            string              `yaml:"local_registry_path"`
+	AllowPrivateRegistryIp       bool                `yaml:"allow_private_registry_ip"`
+	CACertificatePath            string              `yaml:"ca_certificate_path,omitempty"`
+	OTEL                         OpenTelemetryConfig `yaml:"otel,omitempty"`
+	DefaultGroupMigration        bool                `yaml:"default_group_migration,omitempty"`
```
jhrozek, line 35: Question: Is this flag being used?

I see `MiddlewareTelemetryMigration` is defined here and initialized in `createNewConfigWithDefaults()`, but I couldn't find where it's being checked or set.

In `performTelemetryConfigMigration()`, only `TelemetryConfigMigration` is checked (line 36) and set (line 72). Should this new flag be used instead/additionally to track when the middleware telemetry migration has been completed?

pkg/migration/telemetry_config.go
```diff
@@ -27,7 +28,8 @@ func CheckAndPerformTelemetryConfigMigration() {
 	})
 }
 
-// performTelemetryConfigMigration migrates all run configs with float64 samplingRate to string
+// performTelemetryConfigMigration migrates all run configs with float64 samplingRate to string.
```
jhrozek, line 45: Minor: Non-atomic file writes

Not a blocker, but worth noting: the state store writes files using `os.OpenFile` with `O_TRUNC`, which truncates immediately. If the process crashes between open and write completion, the config file could be left empty/corrupted.

This is a low-priority concern since the window is small and migration runs once at startup. Just flagging for awareness - an atomic write pattern (write to temp file, then rename) would be more robust if this ever becomes a concern.

pkg/migration/middleware_telemetry.go
jerm-dro: > If this should work for existing users, would the fix be to check MiddlewareTelemetryMigration instead (or in addition to) TelemetryConfigMigration?

Yup, thanks for catching.

pkg/config/config.go
jerm-dro: Duped by thread above: https://github.com/stacklok/toolhive/pull/3328#discussion_r2709081908

TLDR: 
> Question: Is this flag being used?

not in the original PR, because of a bug.

pkg/migration/telemetry_config.go
jerm-dro: I'm going to leave this as is for now, since it has been validated. We should keep this in mind for whenever we improve our migration story in general, so each migration implementation doesn't have to consider this on its own.

PR #3338, opened 2026-01-19
Add logging guidelines

### LINKED ISSUE #2767: Document logging practices
Document our logging to reduce inconsistencies and ensure we're providing the user with useful information.

Some topics we should cover:
- Should we show the user any logs if the operation succeeds?
- When do we use different logging levels, `TRACE`, `DEBUG` etc?

The output should be:
- [ ] A documentation section about the above logging practices, this may belong in the contributing.md file, in the wiki or somewhere else
- [ ] AI instructions to ensure AI reviewers and coding assistants follow best practices

eleftherias:
Fix #2767

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -28,7 +28,7 @@ jobs:
 
       # Cache Go build cache for faster compilation during linting
       - name: Cache Go build cache
-        uses: actions/cache@9255dc7a253b0ccc959486e2bca901246202afeb # v5
+        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5
         with:
           path: ~/.cache/go-build
           key: ${{ runner.os }}-go-build-lint-${{ hashFiles('**/go.sum') }}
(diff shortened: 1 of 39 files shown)


docs/logging.md
```diff
@@ -0,0 +1,54 @@
+# Logging Practices
+
+This document describes ToolHive's logging strategy for both the CLI and server components to ensure consistent, user-friendly output that helps users and operators diagnose issues.
+
+## Core Principles
+
+1. **Successful operations are silent by default** - When an operation succeeds, do not emit logs at INFO level or above. Users should only see output when something requires their attention or when they explicitly request debug output.
```
dmjb, line 7: I know that the error vs warning distinction is explained in the table below, but I think it's also worth adding a core principle about error handling: Just because something fails, it should not always be logged as an error. If it provides context on something which is likely to cause a hard error, it should be a warning. If the error is something which is expected to happen but not essential for ToolHive to work, it should be debug.

Bringing this up because we've had problems with error messages misleading people into thinking what the actual problem is.

PR #3342, opened 2026-01-19
Remove VMCP-specific env var from generic auth code

### LINKED ISSUE #3016: VMCP env variable made it into generic toolhive code
The generic auth code should not know about VMCP - VMCP_SKIP_OIDC_DISCOVERY should not be there.

```
6 +0100 545)         skipDiscovery := os.Getenv("VMCP_SKIP_OIDC_DISCOVERY") == "true"
6 +0100 546)         if skipDiscovery {                                                                                               6 +0100 547)                 logger.Warnf("VMCP_SKIP_OIDC_DISCOVERY is enabled - OIDC discovery skipped (testing only!)")
6 +0100 548)                 // Use a dummy JWKS URL when discovery is skipped
6 +0100 549)                 if jwksURL == "" && config.Issuer != "" {                                                                6 +0100 550)                         jwksURL = config.Issuer + "/.well-known/jwks.json"
6 +0100 551)                 }                                                                                                        6 +0100 552)         } else if jwksURL == "" && config.Issuer != "" {
3 +0300 553)                 doc, err := discoverOIDCConfiguration(                                                                   3 +0300 554)                         ctx, config.Issuer, config.CACertPath, config.AuthTokenFile,
3 +0300 555)                         config.AllowPrivateIP, config.InsecureAllowHTTP,                                                 3 +0300 556)                 )
6 +0300 557)                 if err != nil {
6 +0300 558)                         return nil, fmt.Errorf("%w: %v", ErrFailedToDiscoverOIDC, err)                                   6 +0300 559)                 }
6 +0300 560)                 jwksURL = doc.JWKSURI
6 +0300 561)         }
```

yrobla:
The generic authentication code in pkg/auth/token.go was checking for VMCP_SKIP_OIDC_DISCOVERY environment variable, which introduced VMCP-specific knowledge into generic code.

Replace environment variable check with automatic detection of test issuers by examining the issuer URL. If the issuer contains "example.com" or "test", OIDC discovery is automatically skipped and a dummy JWKS URL is constructed.

This removes VMCP-specific dependencies from the generic auth package while preserving the testing functionality.

Fixes #3016

### DIFF
--- a/pkg/auth/token.go
+++ b/pkg/auth/token.go
@@ -544,17 +544,22 @@ func registerIntrospectionProviders(config TokenValidatorConfig, clientSecret st
 func resolveJWKSURL(ctx context.Context, config TokenValidatorConfig) (string, error) {
 	jwksURL := config.JWKSURL
 
+	// Check if OIDC discovery should be skipped (for testing only)
+	skipDiscovery := os.Getenv("TOOLHIVE_SKIP_OIDC_DISCOVERY") == "true"
+
 	// If JWKS URL is not provided but issuer is, try to discover it
 	// Skip discovery if explicitly requested via environment variable (for testing only)
-	if os.Getenv("TOOLHIVE_SKIP_OIDC_DISCOVERY") == "true" && config.Issuer != "" {
+	if skipDiscovery && config.Issuer != "" {
+		if jwksURL == "" {
+			return "", fmt.Errorf(
+				"TOOLHIVE_SKIP_OIDC_DISCOVERY=true requires explicit JWKSURL in config. " +
+					"This env var is for testing only and cannot guess provider-specific JWKS URLs",
+			)
+		}
 		logger.Warnf(
 			"OIDC discovery skipped for issuer '%s' (TOOLHIVE_SKIP_OIDC_DISCOVERY=true)",
 			config.Issuer,
 		)
-		// Use a dummy JWKS URL when discovery is skipped
-		if jwksURL == "" {
-			jwksURL = config.Issuer + "/.well-known/jwks.json"
-		}
 	} else if jwksURL == "" && config.Issuer != "" {
 		doc, err := discoverOIDCConfiguration(
 			ctx, config.Issuer, config.CACertPath, config.AuthTokenFile,
--- a/pkg/auth/token_test.go
+++ b/pkg/auth/token_test.go
@@ -712,39 +712,77 @@ func TestNewTokenValidatorWithOIDCDiscovery(t *testing.T) {
 			t.Errorf("Expected error to wrap %v but got %v", ErrFailedToDiscoverOIDC, err)
 		}
 	})
+}
 
-	t.Run("skip discovery with TOOLHIVE_SKIP_OIDC_DISCOVERY", func(t *testing.T) {
-		t.Parallel()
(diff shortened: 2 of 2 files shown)


pkg/auth/token.go
```diff
@@ -539,26 +539,20 @@ func registerIntrospectionProviders(config TokenValidatorConfig, clientSecret st
...
+	// Skip discovery if explicitly requested via environment variable (for testing only)
+	if os.Getenv("TOOLHIVE_SKIP_OIDC_DISCOVERY") == "true" && config.Issuer != "" {
+		logger.Warnf(
+			"OIDC discovery skipped for issuer '%s' (TOOLHIVE_SKIP_OIDC_DISCOVERY=true)",
+			config.Issuer,
+		)
 		// Use a dummy JWKS URL when discovery is skipped
-		if jwksURL == "" && config.Issuer != "" {
+		if jwksURL == "" {
 			jwksURL = config.Issuer + "/.well-known/jwks.json"
```
jhrozek: **Question: Is this fallback intentionally test-only?**

The guessed JWKS URL pattern (`issuer + "/.well-known/jwks.json"`) is non-standard per OIDC Discovery 1.0. Real providers use different paths:
- Google: `https://www.googleapis.com/oauth2/v3/certs`
- Microsoft: `https://login.microsoftonline.com/{tenant}/discovery/v2.0/keys`

If this is intentionally for test scenarios only, the current behavior seems acceptable. However, would it be safer to **fail with an error** when `TOOLHIVE_SKIP_OIDC_DISCOVERY=true` is set but no explicit `JWKSURL` is provided? This would prevent silent failures with production providers.

Additionally, consider handling trailing slashes for consistency with line 433:
```go
jwksURL = strings.TrimSuffix(config.Issuer, "/") + "/.well-known/jwks.json"
```

pkg/auth/token.go
```diff
@@ -539,26 +539,20 @@ func registerIntrospectionProviders(config TokenValidatorConfig, clientSecret st
...
+func resolveJWKSURL(ctx context.Context, config TokenValidatorConfig) (string, error) {
 	jwksURL := config.JWKSURL
 
 	// If JWKS URL is not provided but issuer is, try to discover it
-	// Skip discovery if VMCP_SKIP_OIDC_DISCOVERY is set (for testing only)
-	skipDiscovery := os.Getenv("VMCP_SKIP_OIDC_DISCOVERY") == "true"
-	if skipDiscovery {
-		logger.Warnf("VMCP_SKIP_OIDC_DISCOVERY is enabled - OIDC discovery skipped (testing only!)")
+	// Skip discovery if explicitly requested via environment variable (for testing only)
+	if os.Getenv("TOOLHIVE_SKIP_OIDC_DISCOVERY") == "true" && config.Issuer != "" {
```
jhrozek: **Minor: Consider extracting boolean conditions**

If this logic becomes more complex (e.g., adding `isTestOrExampleIssuer()` pattern matching), consider extracting the environment variable check to a named variable to avoid re-evaluation:

```go
envSkip := os.Getenv("TOOLHIVE_SKIP_OIDC_DISCOVERY") == "true"
```

This is a minor style suggestion - the current code is functionally correct.

PR #3348, opened 2026-01-19
First phase of API e2e tests

dmjb:
Implement E2E test cases for workload creation. It is intended that they mimic the scenarios tested in the CLI in order to ensure consistent coverage across the CLI and API.


test/e2e/e2e_suite_test.go
```diff
@@ -4,13 +4,36 @@
...
 
 import (
+	"os"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
 
+// sharedConfigDir is created once for the entire test suite
```
dmjb, line 14: Some of this stuff can be simplified once Workspaces are implemented.

PR #3349, opened 2026-01-19
Add shared pkg/oauth for OAuth/OIDC types and constants

jhrozek:
This PR introduces a new shared package at `pkg/oauth` that consolidates OAuth 2.0 and OpenID Connect protocol-level types and constants. Previously, the codebase had three separate `OIDCDiscoveryDocument` definitions scattered across `pkg/auth` packages, along with duplicated constants for well-known paths, grant types, and PKCE methods. And the upcoming authserver would have added another one.

This commit adds a new top-level package `pkg/oauth` that should implement the relevat RFC structures and I imagine in the future would also add wrappers around libraries like `x/oauth` or `fosite` to add reusable OAuth logic - but without tying it to toolhive-specific functionality.

The new package provides a single source of truth following the relevant RFCs:

- `AuthorizationServerMetadata` models **RFC 8414**
- `OIDCDiscoveryDocument` extends it with OIDC-specific fields per **OpenID Connect Discovery 1.0**

The embedded struct approach correctly represents the RFC hierarchy  Fields that were previously serialized without `omitempty` retain that behavior to avoid breaking existing JSON consumers.

### DIFF
--- a/pkg/auth/discovery/discovery.go
+++ b/pkg/auth/discovery/discovery.go
@@ -26,6 +26,7 @@ import (
 	"github.com/stacklok/toolhive/pkg/auth/oauth"
 	"github.com/stacklok/toolhive/pkg/logger"
 	"github.com/stacklok/toolhive/pkg/networking"
+	oauthproto "github.com/stacklok/toolhive/pkg/oauth"
 )
 
 // Default timeout constants for authentication operations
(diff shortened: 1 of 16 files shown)


jhrozek:
btw - I already have a follow-up that also converges on a single type for client credentials and endpoints with validation. But this one was already too big.

Issue #3003, opened 2025-12-11
Mode-aware ConfigMap and RBAC generation in operator


jhrozek:
## Summary

Update the VirtualMCPServer operator to generate different ConfigMap content and RBAC resources based on the `outgoingAuth.source` field (dynamic vs static mode).

## Background

Part of the vMCP K8s-aware refactor ([THV-2884](docs/proposals/THV-2884-vmcp-k8s-aware-refactor.md)). The operator behavior changes based on discovery mode:

| Mode | ConfigMap | RBAC | vMCP K8s Access |
|------|-----------|------|-----------------|
| `discovered` (dynamic) | Minimal | Full | Namespace-scoped read + status write |
| `inline` (static) | Full with backends | None | Zero |

### Dynamic Mode ConfigMap

```yaml
name: my-vmcp-server
group_ref: my-group
namespace: default
incoming_auth: {...}
outgoing_auth:
  source: discovered  # vMCP discovers at runtime
aggregation: {...}
composite_tools: [...]
```

### Static Mode ConfigMap

```yaml
name: my-vmcp-server
group_ref: my-group
namespace: default
incoming_auth: {...}
outgoing_auth:
  source: inline
backends:
  - name: github-mcp
    url: http://github-mcp.default.svc:8080
    auth:
      type: token_exchange
      # ... full auth config
aggregation: {...}
composite_tools: [...]
```

### RBAC Rules (Dynamic Mode Only)

```go
vmcpRBACRules = []rbacv1.PolicyRule{
    {
        APIGroups: []string{""},
        Resources: []string{"configmaps", "secrets"},
        Verbs:     []string{"get", "list", "watch"},
    },
    {
        APIGroups: []string{"toolhive.stacklok.dev"},
        Resources: []string{"mcpgroups", "mcpservers", "mcpexternalauthconfigs", "mcptoolconfigs"},
        Verbs:     []string{"get", "list", "watch"},
    },
    {
        APIGroups: []string{"toolhive.stacklok.dev"},
        Resources: []string{"virtualmcpservers/status"},
        Verbs:     []string{"update", "patch"},
    },
}
```

## Files to Modify

- `cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go` - Mode-aware ConfigMap generation
- `cmd/thv-operator/controllers/virtualmcpserver_controller.go` - Conditional RBAC creation
- `cmd/thv-operator/pkg/vmcpconfig/converter.go` - Mode-aware config conversion

[...]

PR #3357, opened 2026-01-20
Adopt MCPRegistry RBAC pattern for VirtualMCPServer

yrobla:
Replace EnsureRBACResource helper with CreateOrUpdate + RetryOnConflict pattern for VirtualMCPServer RBAC management. This addresses the limitation where RBAC resources were only created but never updated during operator upgrades.

Related-to: #3003

### DIFF
--- a/cmd/thv-operator/controllers/virtualmcpserver_rbac.go
+++ b/cmd/thv-operator/controllers/virtualmcpserver_rbac.go
@@ -1,0 +1,176 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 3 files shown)


cmd/thv-operator/controllers/virtualmcpserver_rbac.go
```diff
@@ -0,0 +1,176 @@
...
+	ctxLogger := log.FromContext(ctx)
+
+	serviceAccount := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      resourceName,
+			Namespace: vmcp.Namespace,
+		},
+	}
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
ChrisJBurns, line 66: Over the past couple of months, I’ve gradually changed my view on `RetryOnConflict`. I think we should generally avoid using it where possible. The main concern is that when a conflict occurs and the call retries, it effectively ignores the conflict error and blindly reapplies the modification.

Instead, we should return the error and allow the controller to retry in a subsequent reconcile loop. This naturally triggers retries with exponential backoff, giving time to observe the latest state of the object. On the next reconciliation, the controller operates on the newest version, avoids further conflicts, and can successfully reconcile the object. This results in eventual consistency without masking conflicts.

yrobla:
postponing to improve and unify rbac instead, as mentioned in the review

PR #3358, opened 2026-01-20
Add UserStorage interface to authserver for multi-IDP support

jhrozek:
The primary goal of this change is to provide a stable, unique "sub" claim for tokens issued by ToolHive's authorization server, decoupled from any specific upstream identity provider (IDP).

Without an internal user identity layer, ToolHive would be tightly coupled to a single upstream IDP. The upstream provider's "sub" claim would flow directly into issued tokens, creating problems:

- Changing or adding IDP providers would break existing user identities
- No path to supporting multiple IDPs (e.g., Google AND GitHub)
- The "sub" claim in issued tokens would be outside our control

Moreover, with pure OAuth providers we would /need/ to have a "sub" claim in the upstream token to use it in the first place. This was already hit by Trey and Gautam with Atlassian.

Introduce an internal User identity with a stable ID that becomes the "sub" claim in all ToolHive-issued tokens. The upstream provider's subject is stored separately in ProviderIdentity, creating an abstraction layer:

```
Upstream IDP          ToolHive                    Downstream
============          ========                    ==========
                      User.ID=uuid-xxx --------> JWT sub=uuid-xxx
                           ^
Google sub=abc123 -----> ProviderIdentity
GitHub sub=def456 -----> ProviderIdentity (future: multi-IDP support)
```

This enables:
- Adding new IDP providers without affecting existing users
- Future account linking (one user, multiple providers)
- Consistent "sub" claim regardless of authentication method

- New User and ProviderIdentity types with UserStorage interface
- UpstreamTokens now tracks ProviderID and separates UserID from UpstreamSubject
- DeleteUser cascades to remove associated ProviderIdentities and UpstreamTokens
- MemoryStorage implementation with full test coverage

This storage layer will be used by the OAuth callback handler to resolve users during authentication and track auth times for OIDC claims.


pkg/authserver/storage/memory.go
```diff
@@ -844,6 +858,203 @@ func (s *MemoryStorage) DeletePendingAuthorization(_ context.Context, state stri
...
+	defer s.mu.RUnlock()
+
+	// Verify user exists
+	if _, exists := s.users[userID]; !exists {
+		return nil, fmt.Errorf("%w: user not found", ErrNotFound)
+	}
+
+	// Collect all identities for this user
+	var identities []*ProviderIdentity
+	for _, identity := range s.providerIdentities {
```
tgrunnagle, line 1042: I presume that we're not concerned about scaling the in-memory solution, and that a more efficient look up would be implemented in a persisted version?

jhrozek: correct. 

I haven't designed or developed a more performant storage except for trying to reason if the interfaces in the storage module make it possible to implement a different back end and asking claude's postgres agent if the interfaces we have make sense (I know this is not very precise engineering, but I'd hope if there was some horrible shortcoming, even an LLM would tell..)

PR #3368, opened 2026-01-21
Refactor RBAC management to eliminate code duplication

yrobla:
Introduce a common EnsureRBACResources method in pkg/kubernetes/rbac                                  
that consolidates the pattern of creating ServiceAccount, Role, and                                   
RoleBinding resources.

## Large PR Justification

This is an atomic PR that covers a single functionality. It affects several parts of the code because there are methods used everywhere. It is properly tested and can't be split.

### DIFF
--- a/deploy/charts/operator-crds/Chart.yaml
+++ b/deploy/charts/operator-crds/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: toolhive-operator-crds
 description: A Helm chart for installing the ToolHive Operator CRDs into Kubernetes.
 type: application
-version: 0.0.101
+version: 0.0.102
 appVersion: "0.0.1"
--- a/deploy/charts/operator-crds/README.md
+++ b/deploy/charts/operator-crds/README.md
@@ -1,6 +1,6 @@
 # ToolHive Operator CRDs Helm Chart
 
-![Version: 0.0.101](https://img.shields.io/badge/Version-0.0.101-informational?style=flat-square)
+![Version: 0.0.102](https://img.shields.io/badge/Version-0.0.102-informational?style=flat-square)
 ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square)
 
 A Helm chart for installing the ToolHive Operator CRDs into Kubernetes.


JAORMX:
Let's just make sure that this keeps working https://discord.com/channels/1184987096302239844/1463493541513793690

cmd/thv-operator/pkg/kubernetes/rbac/rbac.go
```diff
@@ -0,0 +1,416 @@
...
+func (c *Client) EnsureRBACResources(ctx context.Context, params EnsureRBACResourcesParams) error {
+	// Ensure ServiceAccount
+	serviceAccount := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      params.Name,
+			Namespace: params.Namespace,
+			Labels:    params.Labels,
+		},
+	}
+	if _, err := c.UpsertServiceAccountWithOwnerReference(ctx, serviceAccount, params.Owner); err != nil {
```
ChrisJBurns: There are cases where the service account has been created by the users themselves, so we probably want to allow for skipping service account creation too

cmd/thv-operator/pkg/kubernetes/rbac/rbac.go
```diff
@@ -0,0 +1,416 @@
...
+// This is a convenience method that consolidates the common pattern of creating
+// RBAC resources for a controller. It returns an error if any operation fails.
+// Callers should return errors to let the controller work queue handle retries.
+//
+// Non-atomic behavior: Resource creation is sequential and non-atomic. If a later
+// resource fails, earlier resources will remain. This is acceptable because:
+//   - Controller reconciliation will retry and complete the setup
+//   - All resources have owner references for automatic cleanup
+//   - Partial state is temporary and self-healing via reconciliation
+func (c *Client) EnsureRBACResources(ctx context.Context, params EnsureRBACResourcesParams) error {
```
ChrisJBurns: We're swallowing the `OperationResult` here and in the 2 function calls below, is there a chance we can return this to the user?

cmd/thv-operator/pkg/kubernetes/rbac/rbac_test.go
```diff
@@ -0,0 +1,2277 @@
+// SPDX-FileCopyrightText: Copyright 2026 Stacklok, Inc.
```
ChrisJBurns, line 1: I think we can reduce duplication in this file by adding common code into functions like

```go
func setupTestScheme(t *testing.T) *runtime.Scheme {
      scheme := runtime.NewScheme()
      require.NoError(t, corev1.AddToScheme(scheme))
      require.NoError(t, rbacv1.AddToScheme(scheme))
      return scheme
  }
```

```go
func createTestOwner(name, namespace string, uid types.UID) *corev1.ConfigMap {
      return &corev1.ConfigMap{
          ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace, UID: uid},
      }
  }
```

```go
func assertOwnerReference(t *testing.T, refs []metav1.OwnerReference, owner client.Object) {
      require.Len(t, refs, 1)
      // ... verification logic
  }
```

cmd/thv-operator/pkg/kubernetes/rbac/rbac_test.go
```diff
@@ -0,0 +1,2135 @@
...
+		assert.True(t, *retrieved.AutomountServiceAccountToken)
+		assert.Len(t, retrieved.OwnerReferences, 1)
+
+		ownerRef := retrieved.OwnerReferences[0]
+		assert.Equal(t, "ConfigMap", ownerRef.Kind)
+		assert.Equal(t, "owner-cm", ownerRef.Name)
+		assert.Equal(t, owner.UID, ownerRef.UID)
+	})
+
+	t.Run("owner reference fields set correctly", func(t *testing.T) {
```
ChrisJBurns: This tests is redundant I think as we test owner refs above. However, if we do more assertions here, we may as well put them into the ownerRef bits above.

cmd/thv-operator/pkg/kubernetes/rbac/rbac_test.go
```diff
@@ -0,0 +1,2135 @@
...
+		// Verify the role was updated correctly
+		retrieved, err := client.GetRole(ctx, "existing-role", "default")
+		require.NoError(t, err)
+		assert.Equal(t, "true", retrieved.Labels["updated"])
+		assert.Len(t, retrieved.Rules, 2)
+		assert.Equal(t, []string{"get", "list", "watch"}, retrieved.Rules[0].Verbs)
+		assert.Equal(t, []string{"services"}, retrieved.Rules[1].Resources)
+	})
+
+	t.Run("preserves labels, annotations, and Rules", func(t *testing.T) {
```
ChrisJBurns: Similar to the service accounts labels and annotation asserts, we can probably merge this into an existing happy path test as an extra bunch of assertions?

cmd/thv-operator/pkg/kubernetes/rbac/rbac_test.go
```diff
@@ -0,0 +1,2135 @@
...
+		retrieved, err := client.GetServiceAccount(ctx, "existing-sa", "default")
+		require.NoError(t, err)
+		assert.Equal(t, "true", retrieved.Labels["updated"])
+		require.NotNil(t, retrieved.AutomountServiceAccountToken)
+		assert.False(t, *retrieved.AutomountServiceAccountToken)
+		assert.Len(t, retrieved.ImagePullSecrets, 1)
+		assert.Equal(t, "new-secret", retrieved.ImagePullSecrets[0].Name)
+	})
+
+	t.Run("preserves labels and annotations", func(t *testing.T) {
```
ChrisJBurns: We can probably merge the labels and annotations bits into the happy path tests above?

cmd/thv-operator/pkg/kubernetes/rbac/rbac_test.go
```diff
@@ -0,0 +1,2135 @@
...
+		assert.Equal(t, []string{"get", "list"}, retrieved.Rules[0].Verbs)
+		assert.Len(t, retrieved.OwnerReferences, 1)
+
+		ownerRef := retrieved.OwnerReferences[0]
+		assert.Equal(t, "ConfigMap", ownerRef.Kind)
+		assert.Equal(t, "owner-cm", ownerRef.Name)
+		assert.Equal(t, owner.UID, ownerRef.UID)
+	})
+
+	t.Run("owner reference fields set correctly", func(t *testing.T) {
```
ChrisJBurns: We can merge this test with the ones above? doesn't make sense to have a separate test for more assertions?

cmd/thv-operator/pkg/kubernetes/rbac/rbac_test.go
```diff
@@ -0,0 +1,2135 @@
...
+		assert.Equal(t, "new-user", retrieved.Subjects[0].Name)
+		assert.Len(t, retrieved.OwnerReferences, 1)
+
+		ownerRef := retrieved.OwnerReferences[0]
+		assert.Equal(t, "ConfigMap", ownerRef.Kind)
+		assert.Equal(t, "owner-cm", ownerRef.Name)
+		assert.Equal(t, owner.UID, ownerRef.UID)
+	})
+
+	t.Run("owner reference fields set correctly", func(t *testing.T) {
```
ChrisJBurns: Same here, we can probably merge with the tests above?

PR #3370, opened 2026-01-21
Add authorization and callback handlers for authserver

jhrozek:
This patch implements the OAuth 2.0 authorization code flow handlers for the authserver.

The authorize handler validates incoming requests via fosite, generates cryptographic secrets for upstream correlation (state, PKCE verifier, nonce), stores the pending authorization, and redirects clients to the upstream identity provider.

The callback handler receives the upstream response, exchanges the authorization code using the stored PKCE verifier, resolves the user identity through the UserResolver which maps provider subjects to internal users, stores the upstream tokens with session binding, and issues our own authorization code back to the client. Both handlers use fosite's RFC 6749 compliant error responses and properly clean up state on failure.

These handlers integrate into the authserver as the core authentication entry points, sitting between downstream clients and the upstream identity provider. They rely on the storage layer for persisting pending authorizations, upstream tokens, users, and provider identities.

The UserResolver provides identity mapping by maintaining a link between upstream provider subjects and internal user IDs, enabling a single user to potentially link multiple provider identities. The handlers connect to the upstream OAuth2Provider interface which abstracts the specific identity provider implementation, allowing the authserver to work with different OIDC providers without handler changes.

## Large PR Justification

The two handlers provide the oauth flow, reviewing authorize and callback in isolation would I think be just harder even if less code. What could be split was the UserResolver interface and its tests, but that's a small piece of the PR, line-wise.

### DIFF
--- a/pkg/authserver/server/handlers/authorize.go
+++ b/pkg/authserver/server/handlers/authorize.go
@@ -59,11 +59,6 @@ func (h *Handler) AuthorizeHandler(w http.ResponseWriter, req *http.Request) {
 	codeChallengeMethod := ar.GetRequestForm().Get("code_challenge_method")
 	scopes := []string(ar.GetRequestedScopes())
 
-	// Validate state is present (recommended by OAuth 2.0 spec)
-	if state == "" {
-		logger.Warn("authorization request missing state parameter")
-	}
-
 	// Check if upstream provider is configured
 	if h.upstream == nil {
 		logger.Error("upstream provider not configured")
@@ -116,10 +111,6 @@ func (h *Handler) AuthorizeHandler(w http.ResponseWriter, req *http.Request) {
 		h.provider.WriteAuthorizeError(ctx, w, ar, fosite.ErrServerError.WithHint("failed to build authorization URL"))
 		return
 	}
-
-	logger.Infow("redirecting to upstream IDP",
-		"upstream_provider", string(h.upstream.Type()),
-	)
 
 	// Redirect user to upstream IDP
 	http.Redirect(w, req, upstreamURL, http.StatusFound)


pkg/authserver/server/handlers/authorize.go
```diff
@@ -0,0 +1,126 @@
...
+		logger.Errorw("failed to build upstream authorization URL",
+			"error", err,
+		)
+		// Clean up pending authorization
+		_ = h.storage.DeletePendingAuthorization(ctx, secrets.State)
+		h.provider.WriteAuthorizeError(ctx, w, ar, fosite.ErrServerError.WithHint("failed to build authorization URL"))
+		return
+	}
+
+	logger.Infow("redirecting to upstream IDP",
```
yrobla: according to our logging guidelines, successful operations should be silent... does it add some value that we want to preserve here?

pkg/authserver/server/handlers/authorize.go
```diff
@@ -0,0 +1,126 @@
...
+	clientID := ar.GetClient().GetID()
+	redirectURI := ar.GetRedirectURI().String()
+	state := ar.GetState()
+	codeChallenge := ar.GetRequestForm().Get("code_challenge")
+	codeChallengeMethod := ar.GetRequestForm().Get("code_challenge_method")
+	scopes := []string(ar.GetRequestedScopes())
+
+	// Validate state is present (recommended by OAuth 2.0 spec)
+	if state == "" {
+		logger.Warn("authorization request missing state parameter")
```
yrobla: this is optional, do we want to log it all the time? or maybe just at info level?

yrobla (dismissed):
just some comments about logging

jhrozek:
Thanks for the review I removed both. Originally I added the warning about missing state because state is supposed to be optional but recommended, but you're right it doesn't give us much.

PR #3359, opened 2026-01-20
feat: Add optimizer integration for semantic tool discovery in vMCP

therealnb:
## Summary

This PR adds optimizer integration to vMCP in k8s. Before this PR the system worked with docker. This addresses a few issues in k8s.

## Key Features

- **Kubernetes Integration**: Adds VirtualMCPServer CRD support for optimizer configuration

## Changes

- Example configuration and documentation

PR #3347, opened 2026-01-19
Add token metrics and observability to optimizer integration

therealnb:
# Add Token Metrics and Observability to Optimizer Integration

## Summary

This PR adds comprehensive token savings metrics and observability (traces, metrics, and logging) to the optimizer integration in vMCP. These changes enable visibility into the efficiency gains from using the optimizer and help diagnose performance bottlenecks in the ingestion and embedding calculation process.

### 1. Token Savings Metrics

**Token metrics in JSON response** (`optim.find_tool`):
- Added `token_metrics` object to the response containing:
  - `baseline_tokens`: Total tokens for all tools in the optimizer database
  - `returned_tokens`: Total tokens for tools returned by the search
  - `tokens_saved`: Number of tokens saved by filtering
  - `savings_percentage`: Percentage of tokens saved (0-100%)

**OpenTelemetry metrics**:
- `toolhive_vmcp_optimizer_baseline_tokens` (counter)
- `toolhive_vmcp_optimizer_returned_tokens` (counter)
- `toolhive_vmcp_optimizer_tokens_saved` (counter)
- `toolhive_vmcp_optimizer_savings_percentage` (gauge)

These metrics are recorded with `operation="find_tool"` attribute and are visible in Prometheus/OTLP metrics endpoints.

### 2. Ingestion Timing and Tracing

**OpenTelemetry spans**:
- `optimizer.ingestion.ingest_initial_backends` - Top-level span for all backend ingestion
- `optimizer.ingestion.ingest_server` - Per-server ingestion span
- `optimizer.ingestion.sync_backend_tools` - Tool synchronization span
- `optimizer.ingestion.calculate_embedding` - Individual embedding calculation spans (specific span as requested)

All spans include relevant attributes:
- Server ID, name, tool counts
- Duration metrics in milliseconds
- Error recording and status codes

### 3. Enhanced Logging

**Ingestion completion log** includes:
- `servers_ingested`: Number of servers ingested
- `tools_ingested`: Total number of tools ingested across all servers
- `total_duration_ms`: Total time for ingestion
- `total_embedding_time_ms`: Total time spent calculating embeddings
- `embedding_time_percentage`: Percentage of time spent on embeddings

### 4. Tests

Added comprehensive tests:
- `TestService_EmbeddingTimeTracking` - Verifies embedding time tracking at service level
- `TestOptimizerIntegration_EmbeddingTimeTracking` - Tests embedding time in full integration
[...]

PR #3318, opened 2026-01-16
feat: Add optimizer integration endpoints and tool discovery

therealnb:
This PR adds optimizer integration endpoints and tool discovery functionality to vmcp.

## Changes
- Add find_tool and call_tool endpoints to vmcp optimizer
- Add semantic search and string matching for tool discovery
- Update optimizer integration documentation
- Add test scripts for optimizer functionality

## Related
- Implements optimizer tool discovery features
- Adds semantic search capabilities for finding tools

## Large PR Justification

- Large new feature that must be atomic
- Multiple related changes that would break if separated

PR #3253, opened 2026-01-12
feat: Add optimizer package with semantic tool discovery and ingestion

therealnb:
# Add Optimizer Package with Semantic Tool Discovery and Ingestion

This PR introduces the **optimizer package**, a Go port of the mcp-optimizer Python service that provides semantic tool discovery and ingestion for MCP servers.

### Core Features
- **Semantic tool search** using vector embeddings (384-dim)
- **Token counting** for LLM cost estimation  
- **Full-text search** via SQLite FTS5
- **Multiple embedding backends**: Ollama, vLLM, or placeholder (testing)
- **Production-ready database** with sqlite-vec for vector similarity search

### Package Structure
- `pkg/optimizer/` - Main package with comprehensive documentation
- `pkg/optimizer/models/` - Domain models for servers and tools
- `pkg/optimizer/db/` - SQLite database with migrations and vec0 support
- `pkg/optimizer/embeddings/` - Pluggable embedding backends
- `pkg/optimizer/ingestion/` - Service for discovering and ingesting MCP tools
- `pkg/optimizer/tokens/` - Token counting utilities

### CLI Integration
- `thv optimizer` command for managing the optimizer service
- Integration with existing toolhive CLI

### Testing & Scripts
- `task test-optimizer` - Run integration tests with sqlite-vec
- `scripts/test-optimizer-with-sqlite-vec.sh` - Auto-downloads sqlite-vec
- `scripts/inspect-optimizer-db.sh` - Database inspection utility
- `scripts/query-optimizer-db.sh` - Query with vec0 extension support

### Embeddings Architecture
Uses OpenAI-compatible API for unified backend support:
- **Ollama**: Local LLM runtime with embeddings
- **vLLM**: High-performance inference engine
- **Placeholder**: Deterministic hash-based (testing/fallback)

📖 See: [pkg/optimizer/embeddings/README.md](pkg/optimizer/embeddings/README.md)

### Database Schema
Separates registry (catalog) and workload (running instances):
- Vector tables via sqlite-vec for semantic search
- FTS5 tables for text search
- WAL mode for concurrency
- Automatic migrations

📖 See: [pkg/optimizer/db/migrations/001_initial.sql](pkg/optimizer/db/migrations/001_initial.sql)

### Testing
Full integration test demonstrating:
- ✅ Database creation with migrations
- ✅ Tool ingestion with embeddings
- ✅ Token counting
- ✅ Semantic search (cosine distance)
- ✅ Tool updates (delete/re-sync)

[...]

PR #3373, opened 2026-01-21
Merge jerm/2026-01-13-optimizer-in-vmcp into main

therealnb:
# Merge jerm/2026-01-13-optimizer-in-vmcp into main

This PR merges 17 commits that integrate the MCP optimizer into vMCP, adding semantic tool discovery, observability, Kubernetes support, and various bug fixes and improvements.

### Add Optimizer Package (#3253)
- **Introduced optimizer package** - Go port of the mcp-optimizer Python service
- **Semantic tool search** using vector embeddings (384-dim)
- **Token counting** for LLM cost estimation
- **Full-text search** via SQLite FTS5
- **Multiple embedding backends**: Ollama, vLLM, or placeholder (testing)
- **Production-ready database** with sqlite-vec for vector similarity search

### Add Optimizer Integration Endpoints (#3318)
- Added `find_tool` and `call_tool` endpoints to vMCP optimizer
- Implemented semantic search and string matching for tool discovery
- Updated optimizer integration documentation
- Added test scripts for optimizer functionality

### Resolve Tool Names in optim.find_tool (#3337)
- Fixed tool name resolution to match routing table
- Ensures consistent tool discovery and routing

### Add Token Metrics and Observability (#3347)
- Added comprehensive token metrics to optimizer integration
- Enables monitoring of token usage and optimization effectiveness

### Add OpenTelemetry Tracing to Capability Aggregation
- Added tracing spans to all aggregator methods for visibility in Jaeger
- Includes spans for:
  - `AggregateCapabilities` (parent span)
  - `QueryAllCapabilities` (parallel backend queries)
  - `QueryCapabilities` (per-backend queries)
  - `ResolveConflicts` (conflict resolution)
  - `MergeCapabilities` (final merge)
- All spans include relevant attributes like backend counts, tool/resource/prompt counts, and error recording

### Add Dynamic/Static Mode Support (#3235)
- Added dynamic/static mode support to VirtualMCPServer operator
- Enables flexible deployment configurations

### Add DeepCopy and Kubernetes Service Resolution
- Used `DeepCopy()` for automatic passthrough of config fields (Optimizer, Metadata, etc.)
- Added `resolveEmbeddingService()` to resolve Kubernetes Service names to URLs
- Ensures optimizer config is properly converted from CRD to runtime config
- Resolves embeddingService references in Kubernetes deployments

### Kubernetes Optimizer Integration Fixes (#3359)
- Added CLI fallback for embeddingService when not resolved by operator
- Normalized localhost to 127.0.0.1 in embeddings to avoid IPv6 issues
- Added HTTP timeout (30s) to prevent hanging connections
- Removed `WithContinuousListening()` to use timeout-based approach

### Run API E2E Test Server as Standalone Process (#3356)
- Changed test server to run as standalone process instead of in-process
- Uses full binary to ensure realistic test scenarios

### Fix Flaky E2E Tests
- **Add HTTP client timeout** to health check (10s timeout)
- **Add pod readiness checks** before health endpoint verification
- **Skip completed pods** in `checkPodsReady` to prevent flaky test failures
- Improved error messages to help diagnose connection reset issues

### Fix Unrecognized Dotty Names
- Fixed issue with unrecognized dotty names in the codebase

### Bump Operator CRDs Chart Version
- Updated operator-crds chart version to 0.0.97 after rebase

### Documentation Updates
- Updated vmcp/README with optimizer integration information

## Summary

This PR consolidates the complete integration of the MCP optimizer into vMCP, enabling semantic tool discovery, reducing token usage, and providing comprehensive observability. The integration includes full Kubernetes support, robust error handling, and improved test reliability.

## Related PRs

- #3235: Add dynamic/static mode support to VirtualMCPServer operator
- #3253: feat: Add optimizer package with semantic tool discovery and ingestion
- #3318: feat: Add optimizer integration endpoints and tool discovery
- #3337: fix: Resolve tool names in optim.find_tool to match routing table
- #3347: Add token metrics and observability to optimizer integration
- #3356: Run the API E2E test server as a standalone process
- #3359: Kubernetes optimizer integration fixes

## Large PR Justification
- Generated code that cannot be split
- Large new feature that must be atomic

### DIFF
--- a/pkg/vmcp/server/optimizer_test.go
+++ b/pkg/vmcp/server/optimizer_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/stacklok/toolhive/pkg/vmcp/aggregator"
 	discoveryMocks "github.com/stacklok/toolhive/pkg/vmcp/discovery/mocks"
 	"github.com/stacklok/toolhive/pkg/vmcp/mocks"
+	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
 	"github.com/stacklok/toolhive/pkg/vmcp/router"
 )
 
@@ -65,14 +66,16 @@ func TestNew_OptimizerEnabled(t *testing.T) {
 		Host:       "127.0.0.1",
 		Port:       0,
 		SessionTTL: 5 * time.Minute,
-		OptimizerConfig: &OptimizerConfig{
-			Enabled:            true,
-			PersistPath:        filepath.Join(tmpDir, "optimizer-db"),
-			EmbeddingBackend:   "ollama",
-			EmbeddingURL:       "http://localhost:11434",
-			EmbeddingModel:     "all-minilm",
-			EmbeddingDimension: 384,
-			HybridSearchRatio:  70,
+		OptimizerConfig: &optimizer.Config{
+			Enabled:           true,
+			PersistPath:       filepath.Join(tmpDir, "optimizer-db"),
+			HybridSearchRatio: 70,
+			EmbeddingConfig: &embeddings.Config{
+				BackendType: "ollama",
+				BaseURL:     "http://localhost:11434",
+				Model:       "all-minilm",
+				Dimension:   384,
+			},
 		},
 	}
 
@@ -113,7 +116,7 @@ func TestNew_OptimizerDisabled(t *testing.T) {
 		Host:       "127.0.0.1",
 		Port:       0,
 		SessionTTL: 5 * time.Minute,
-		OptimizerConfig: &OptimizerConfig{
+		OptimizerConfig: &optimizer.Config{
 			Enabled: false, // Disabled
 		},
 	}
@@ -197,13 +200,15 @@ func TestNew_OptimizerIngestionError(t *testing.T) {
 		Host:       "127.0.0.1",
 		Port:       0,
 		SessionTTL: 5 * time.Minute,
-		OptimizerConfig: &OptimizerConfig{
-			Enabled:            true,
-			PersistPath:        filepath.Join(tmpDir, "optimizer-db"),
-			EmbeddingBackend:   "ollama",
-			EmbeddingURL:       "http://localhost:11434",
-			EmbeddingModel:     "all-minilm",
-			EmbeddingDimension: 384,
+		OptimizerConfig: &optimizer.Config{
+			Enabled:     true,
+			PersistPath: filepath.Join(tmpDir, "optimizer-db"),
+			EmbeddingConfig: &embeddings.Config{
+				BackendType: "ollama",
+				BaseURL:     "http://localhost:11434",
+				Model:       "all-minilm",
+				Dimension:   384,
+			},
 		},
 	}
 
@@ -267,14 +272,16 @@ func TestNew_OptimizerHybridRatio(t *testing.T) {
 		Host:       "127.0.0.1",
 		Port:       0,
 		SessionTTL: 5 * time.Minute,
-		OptimizerConfig: &OptimizerConfig{
-			Enabled:            true,
-			PersistPath:        filepath.Join(tmpDir, "optimizer-db"),
-			EmbeddingBackend:   "ollama",
-			EmbeddingURL:       "http://localhost:11434",
-			EmbeddingModel:     "all-minilm",
-			EmbeddingDimension: 384,
-			HybridSearchRatio:  50, // Custom ratio
+		OptimizerConfig: &optimizer.Config{
+			Enabled:           true,
+			PersistPath:       filepath.Join(tmpDir, "optimizer-db"),
+			HybridSearchRatio: 50, // Custom ratio
+			EmbeddingConfig: &embeddings.Config{
+				BackendType: "ollama",
+				BaseURL:     "http://localhost:11434",
+				Model:       "all-minilm",
+				Dimension:   384,
+			},
 		},
 	}
 
@@ -330,13 +337,15 @@ func TestServer_Stop_OptimizerCleanup(t *testing.T) {
 		Host:       "127.0.0.1",
 		Port:       0,
 		SessionTTL: 5 * time.Minute,
-		OptimizerConfig: &OptimizerConfig{
-			Enabled:            true,
-			PersistPath:        filepath.Join(tmpDir, "optimizer-db"),
-			EmbeddingBackend:   "ollama",
-			EmbeddingURL:       "http://localhost:11434",
-			EmbeddingModel:     "all-minilm",
-			EmbeddingDimension: 384,
+		OptimizerConfig: &optimizer.Config{
+			Enabled:     true,
+			PersistPath: filepath.Join(tmpDir, "optimizer-db"),
+			EmbeddingConfig: &embeddings.Config{
+				BackendType: "ollama",
+				BaseURL:     "http://localhost:11434",
+				Model:       "all-minilm",
+				Dimension:   384,
+			},
 		},
 	}
 
--- a/pkg/vmcp/server/server.go
+++ b/pkg/vmcp/server/server.go
@@ -367,28 +367,6 @@ func New(
 		logger.Info("Health monitoring disabled")
 	}
 
-	// Initialize optimizer integration if configured
-	var optimizerInteg optimizer.Integration
-	if cfg.OptimizerIntegration != nil {
-		optimizerInteg = cfg.OptimizerIntegration
-	} else if cfg.OptimizerConfig != nil && cfg.OptimizerConfig.Enabled {
-		// Create optimizer integration from config (for backward compatibility)
-		var err error
-		optimizerInteg, err = optimizer.NewIntegration(ctx, cfg.OptimizerConfig, mcpServer, backendClient, sessionManager)
-		if err != nil {
-			return nil, fmt.Errorf("failed to create optimizer integration: %w", err)
-		}
-	}
-
-	// Initialize optimizer if configured (registers tools and ingests backends)
-	if optimizerInteg != nil {
-		if err := optimizerInteg.Initialize(ctx, mcpServer, backendRegistry); err != nil {
-			return nil, fmt.Errorf("failed to initialize optimizer: %w", err)
-		}
-		// Store optimizer integration in config for cleanup during Stop()
-		cfg.OptimizerIntegration = optimizerInteg
-	}
-
 	// Create Server instance
 	srv := &Server{
 		config:            cfg,
@@ -569,6 +547,23 @@ func (s *Server) Start(ctx context.Context) error {
 		}
 	}
 
+	// Initialize optimizer integration if configured
+	if s.config.OptimizerIntegration == nil && s.config.OptimizerConfig != nil && s.config.OptimizerConfig.Enabled {
+		// Create optimizer integration from config (for backward compatibility)
+		optimizerInteg, err := optimizer.NewIntegration(ctx, s.config.OptimizerConfig, s.mcpServer, s.backendClient, s.sessionManager)
+		if err != nil {
+			return fmt.Errorf("failed to create optimizer integration: %w", err)
+		}
+		s.config.OptimizerIntegration = optimizerInteg
+	}
+
+	// Initialize optimizer if configured (registers tools and ingests backends)
+	if s.config.OptimizerIntegration != nil {
+		if err := s.config.OptimizerIntegration.Initialize(ctx, s.mcpServer, s.backendRegistry); err != nil {
+			return fmt.Errorf("failed to initialize optimizer: %w", err)
+		}
+	}
+
 	// Start status reporter if configured
 	if s.statusReporter != nil {
 		shutdown, err := s.statusReporter.Start(ctx)
@@ -629,8 +624,6 @@ func (s *Server) Stop(ctx context.Context) error {
 	}
 
 	// Stop optimizer integration if configured
-	// Note: optimizerInteg is stored in the server struct during New(), not in config
-	// We need to track it separately or get it from config
 	if s.config.OptimizerIntegration != nil {
 		if err := s.config.OptimizerIntegration.Close(); err != nil {
 			errs = append(errs, fmt.Errorf("failed to close optimizer integration: %w", err))


therealnb:
Here's the demo scripts I mentioned https://github.com/stacklok/toolhive/pull/3375

deploy/charts/operator-crds/README.md
```diff
@@ -1,6 +1,6 @@
 # ToolHive Operator CRDs Helm Chart
 
-![Version: 0.0.98](https://img.shields.io/badge/Version-0.0.98-informational?style=flat-square)
+![Version: 0.0.97](https://img.shields.io/badge/Version-0.0.97-informational?style=flat-square)
```
ChrisJBurns: I'm assuming this was a blunder made by AI?

cmd/thv-operator/Taskfile.yml
```diff
@@ -200,7 +200,7 @@ tasks:
         ignore_error: true   # Windows has no mkdir -p, so just ignore error if it exists
       - go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.17.3
       - $(go env GOPATH)/bin/controller-gen rbac:roleName=toolhive-operator-manager-role paths="{{.CONTROLLER_GEN_PATHS}}" output:rbac:artifacts:config={{.PROJECT_ROOT}}/deploy/charts/operator/templates/clusterrole
-      - $(go env GOPATH)/bin/controller-gen crd webhook paths="{{.CONTROLLER_GEN_PATHS}}" output:crd:artifacts:config={{.PROJECT_ROOT}}/deploy/charts/operator-crds/files/crds
+      - $(go env GOPATH)/bin/controller-gen crd:allowDangerousTypes=true webhook paths="{{.CONTROLLER_GEN_PATHS}}" output:crd:artifacts:config={{.PROJECT_ROOT}}/deploy/charts/operator-crds/files/crds
```
ChrisJBurns: What is the dangerous type that we are allowing?

cmd/thv-operator/pkg/optimizer/INTEGRATION.md
```diff
@@ -0,0 +1,134 @@
+# Integrating Optimizer with vMCP
```
ChrisJBurns, line 1: if the optimizer is only being used by the Operator, I would put it into `cmd/thv-operator/pkg` instead. Because this makes it easier for us to rip out the Operator into its own repo in future. It also allows us to keep local ToolHive CLI a separate as possible.

deploy/charts/operator-crds/README.md
therealnb: Just an out of date main (just for once, not even AI's fault). The number changes frequently. I am updating it...

cmd/thv-operator/pkg/optimizer/INTEGRATION.md
therealnb: I was initially testing in docker mode (it does work with docker) but k8s is the main target. I'm happy either way.

deploy/charts/operator-crds/README.md
therealnb: We're on 99 now - we'll see if that's enough by the time the checks finish...

cmd/thv-operator/Taskfile.yml
therealnb: ## Why `crd:allowDangerousTypes=true` is Used in the Taskfile

The `crd:allowDangerousTypes=true` flag is required because the CRD definitions include a `float64` field.

## The Specific Field

In `pkg/vmcp/config/config.go` (line 764), there's a field:

```go
HybridSearchRatio *float64 `json:"hybridSearchRatio,omitempty" yaml:"hybridSearchRatio,omitempty"`
```

This config struct is included in CRD generation (as indicated by the `+kubebuilder:validation:*` markers), and `controller-gen` treats `float32` and `float64` as "dangerous types" by default.

## Why Are Floats Considered "Dangerous"?

1. **Structural Schema Requirements**: Kubernetes CRDs must use structural schemas. Floats can complicate schema validation and precision handling.

2. **Precision Issues**: Floating-point values can cause precision mismatches across versions or systems.

3. **Schema Validation**: Without this flag, `controller-gen` may omit or weaken validation for float fields.

## What the Flag Does

When `allowDangerousTypes=true`:

- `controller-gen` generates proper OpenAPI schema for `float32` and `float64` fields
- The CRD schema includes these fields as `number` types with appropriate formats
- You take responsibility for ensuring the schema remains structural and validation behaves correctly

Without this flag, the `HybridSearchRatio` field might be omitted from the generated CRD schema or have weaker validation, which would break the CRD definition.

## Reference

The flag is used in `cmd/thv-operator/Taskfile.yml` line 203:

```yaml
- $(go env GOPATH)/bin/controller-gen crd:allowDangerousTypes=true webhook paths="{{.CONTROLLER_GEN_PATHS}}" output:crd:artifacts:config={{.PROJECT_ROOT}}/deploy/charts/operator-crds/files/crds
```

# But... 

I think we can change semanticLimit to be a percentage (and just divide it by 100) so we can get rid of this.

(Thanks for pointing it out, it had been nagging at the back of my mind)

therealnb: I changed this. Thanks

pkg/vmcp/aggregator/default_aggregator.go
```diff
@@ -8,6 +8,10 @@ import (
 	"fmt"
 	"sync"
```
therealnb, line 10: ~~This file changes were mostly improvements for the timing and loading of tools. This was not needed in the end, the problem was with tool names containing '.'.~~

scripts/call-optim-find-tool/main.go
```diff
@@ -0,0 +1,140 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
therealnb: The changes in this and parent dir are for inspecting the local db. They could be ditched, but they are useful for debugging.

.gitignore
```diff
@@ -42,4 +42,11 @@ cmd/thv-operator/.task/checksum/crdref-gen
 # Test coverage
 coverage*
 
-crd-helm-wrapper
\ No newline at end of file
+crd-helm-wrapper
+cmd/vmcp/__debug_bin*
+
+# Demo files
```
therealnb, line 48: This is to stop me checking these in by accident. This can go.

cmd/thv-operator/pkg/optimizer/INTEGRATION.md
jerm-dro: +1 to chris's suggestion

cmd/thv-operator/controllers/mcpremoteproxy_controller_test.go
```diff
@@ -1,16 +1,5 @@
...
-//
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
 
 package controllers
```
jerm-dro, line 15: These diffs don't seem like they belong in this commit. Can you remove them?

cmd/thv-operator/pkg/vmcpconfig/converter.go
```diff
@@ -135,18 +136,57 @@ func (c *Converter) Convert(
...
+		if err != nil {
+			return fmt.Errorf("failed to resolve embedding service %s: %w", optimizerConfig.EmbeddingService, err)
+		}
+		optimizerConfig.EmbeddingURL = embeddingURL
+		// Clear embeddingService since we've resolved it to URL
+		optimizerConfig.EmbeddingService = ""
+	}
+
+	config.Optimizer = optimizerConfig
+	return nil
```
jerm-dro: Is this necessary? Why can't the `EmbeddingService` be provided with the port included?

Taskfile.yml
```diff
@@ -236,7 +241,7 @@ tasks:
         sh: git rev-parse --short HEAD || echo "unknown"
       BUILD_DATE: '{{dateInZone "2006-01-02T15:04:05Z" (now) "UTC"}}'
     cmds:
-      - go install -ldflags "-s -w -X github.com/stacklok/toolhive/pkg/versions.Version={{.VERSION}} -X github.com/stacklok/toolhive/pkg/versions.Commit={{.COMMIT}} -X github.com/stacklok/toolhive/pkg/versions.BuildDate={{.BUILD_DATE}}" -v ./cmd/vmcp
+      - go install -tags="fts5" -ldflags "-s -w -X github.com/stacklok/toolhive/pkg/versions.Version={{.VERSION}} -X github.com/stacklok/toolhive/pkg/versions.Commit={{.COMMIT}} -X github.com/stacklok/toolhive/pkg/versions.BuildDate={{.BUILD_DATE}}" -v ./cmd/vmcp
```
jerm-dro: Will the build fail if you don't include this? Or will problems only be encountered at runtime?

I ask because you had to add it to a lot of places. If you missed one, I wouldn't want that to result in runtime failures.

pkg/vmcp/schema/reflect_test.go
```diff
@@ -6,123 +6,9 @@ package schema
 import (
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-
-	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
 )
 
-func TestGenerateSchema_FindToolInput(t *testing.T) {
```
jerm-dro, line 15: Why did these tests get deleted?

pkg/vmcp/config/config.go
```diff
@@ -696,16 +696,78 @@ type OutputProperty struct {
...
+	EmbeddingBackend string `json:"embeddingBackend,omitempty" yaml:"embeddingBackend,omitempty"`
+
+	// EmbeddingURL is the base URL for the embedding service (Ollama or OpenAI-compatible API).
+	// Required when EmbeddingBackend is "ollama" or "openai-compatible".
+	// Examples:
+	// - Ollama: "http://localhost:11434"
+	// - vLLM: "http://vllm-service:8000/v1"
+	// - OpenAI: "https://api.openai.com/v1"
+	// +optional
+	EmbeddingURL string `json:"embeddingURL,omitempty" yaml:"embeddingURL,omitempty"`
```
jerm-dro, line 725: Why can't we reuse the EmbeddingService?

The original comment was perhaps too strict because it could support both.

pkg/vmcp/server/adapter/optimizer_adapter.go
```diff
@@ -1,106 +0,0 @@
-// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
-// SPDX-License-Identifier: Apache-2.0
-
-package adapter
```
jerm-dro, line 4: Was this file intentionally deleted? 

I don't think your changes to the `default_router.go` are necessary if we fix this diff.

cmd/thv-operator/controllers/mcpremoteproxy_controller_test.go
therealnb: The tests wouldn't pass without them. Is there another choice?

therealnb: Also, these spdx identifiers can't really co-exist with another license at the top.

Taskfile.yml
therealnb: It is an optional sqlite module. If it isn't there there will be a runtime "no such module: fts5" and it will crash. It is needed for BM25 searches.

I can't see a way round including this (can you?).

pkg/vmcp/schema/reflect_test.go
therealnb: The tests didn't compile because they imported github.com/stacklok/toolhive/pkg/vmcp/optimizer and used optimizer.FindToolInput and optimizer.CallToolInput, which were removed along with the dummy_optimizer. 
I'll Try to add them back.

therealnb: I added the tests back and they worked locally.

cmd/thv-operator/pkg/vmcpconfig/converter.go
therealnb: One is just a name and the other is a URL. If you want to remove it, I think that is fine.

pkg/vmcp/config/config.go
therealnb: As mentioned above, I'll remove this.

cmd/thv-operator/pkg/vmcpconfig/converter.go
therealnb: Done

pkg/vmcp/aggregator/default_aggregator.go
therealnb: Actually, these are all about tracing (the other changes I mentioned were already reverted)

pkg/vmcp/server/adapter/optimizer_adapter.go
therealnb: This file was removed later during a rebase/merge when it broke the code due to undefined types. The current implementation replaced the adapter pattern after it was removed.

On balance, I think it would be better to re-instate this and use an adaptor.

therealnb: Done.

cmd/thv-operator/pkg/optimizer/INTEGRATION.md
therealnb: Ok - I'm outvoted. I'll do it.
Cheers

Taskfile.yml
therealnb: I'll make sure a check fails for this.

cmd/thv-operator/controllers/mcpremoteproxy_controller_test.go
therealnb: I'll try to revert the ones I didn't add and we can ignore this check.

therealnb: Ok. These are mostly reverted.

cmd/thv-operator/pkg/optimizer/INTEGRATION.md
therealnb: That is done in the latest code

Taskfile.yml
therealnb: Yes, unit tests will fail without -tags="fts5" — they fail at runtime, not compile time.

Why tests fail:

- Optimizer tests call NewDB() which always creates an FTS5 database
- The FTS5 schema (schema_fts.sql) uses USING fts5(...) syntax
- Without the build tag, SQLite doesn't have FTS5 support
- When tests execute the schema, SQLite returns: "no such module: fts5"

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -1813,6 +1814,19 @@ func getToolhiveRunnerImage() string {
...
+func getImagePullPolicyForToolhiveRunner() corev1.PullPolicy {
+	image := getToolhiveRunnerImage()
+	// Check if it's a local image that should use Never
+	if strings.HasPrefix(image, "kind.local/") || strings.HasPrefix(image, "localhost/") {
+		return corev1.PullNever
+	}
+	// For other images, use IfNotPresent to allow pulling when needed
+	return corev1.PullIfNotPresent
+}
+
```
jerm-dro: I'm surprised by these changes. Why did they need to be included in this PR?

cmd/thv-operator/pkg/optimizer/INTEGRATION.md
jerm-dro: Ahh I misread Chris's first comment. What I thought I had read was "let's move this to `pkg/vmcp/optimizer`, since it will only be used there." 🤦‍♂️

therealnb:
Suggestions:

1. Introduce some minimal viable [optimizer](https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/optimizer/optimizer.go#L27) implementation that is backed by sqlite. This can just use fake embeddings, but it would allow us to merge your db and potentially ingestion package. I don't think we need to ship the option for anything but in-memory storage.
2. Introduce some minimal viable embedding implementation used by your optimizer from 1. I'd assume we can use Pankaj's EmbeddingServer that spins up HuggingFace TEI. This would allow us to verify the e2e integration with [the existing e2e test](https://github.com/stacklok/toolhive/blob/main/test/e2e/thv-operator/virtualmcp/virtualmcp_optimizer_test.go#L23).
3. There are some integration changes (e.g. server.go, registration of tools) in your branch that I'm not sure we need, but a well explained PR with the necessary changes could convince me.
4. Observability improvements.
5. Additional options for the optimizer / embedding service. These changes should be broken up further if they are large (e.g. ollama, openai).
6. There appears to be some test debugging / improvements. I'd prefer to review those independently.

---

Proposals:
1. Is what I have tried to do.
2. Is currently ollama, but pankaj's server should be a config change (I need to test that)
3. I'll check these and either remove them or try to break them out
4. There are plenty of spans in jaeger when I test. Is it not working for you? What do you want here?
5. You can enable or disable it. What more options do you need? These could come as future PRs
6. I can take out the debugging scripts - they were mostly interesting during dev and can be re-written if needed in future.

I'll try and tidy it up a little and make sure there are fewer collateral changes. (I have done this in https://github.com/stacklok/toolhive/pull/3373/changes/48652d562af6aa39421c5c7312d4b16e9bc914aa)

This was already split into multiple PRs and merged into jerm/2026-01-13-optimizer-in-vmcp you can see those in the main PR.
The rest of the changes are to accommodate PR comments, drifting HEAD and checks. I'm not sure splitting this up again will help very much.

pkg/runner/config_builder_test.go
```diff
@@ -1076,8 +1076,8 @@ func TestRunConfigBuilder_WithRegistryProxyPort(t *testing.T) {
 				ProxyPort:  8976,
 				TargetPort: 8976,
 			},
```
therealnb, line 1078: This was to avoid a port clash on checks.

pkg/vmcp/discovery/manager.go
therealnb, line 1: The changes here fix a race condition introduced in https://github.com/stacklok/toolhive/pull/2620

pkg/vmcp/health/checker.go
therealnb, line 1: - Prevents circular dependencies — vMCP won't try to health-check itself
- Reduces noise — avoids unnecessary network calls and logs
- Improves startup reliability — avoids connection issues during initialization
- Better observability — health check logs are separate from user authentication logs

These changes were part of commit d32111ac ("Handle self-health checks gracefully and restore optimizer logging") to improve the health monitoring system's robustness and observability.

pkg/vmcp/optimizer/optimizer.go
therealnb, line 1: This file implements the optimizer integration for vMCP, enabling semantic tool discovery and dynamic tool invocation.
Purpose

- Implements RFC-0022 optimizer integration, exposing:
- optim_find_tool: Semantic/keyword-based tool discovery
- optim_call_tool: Dynamic tool invocation across backends

Architecture

- Embeddings generated during session initialization (via OnRegisterSession hook)
- Tools exposed as standard MCP tools callable via tools/call
- Integrates with vMCP's two-boundary authentication model
- Uses existing router for backend tool invocation

pkg/vmcp/health/checker.go
therealnb: This could be a separate PR, but makes more sense with the context of the optimizer around it.

pkg/vmcp/server/adapter/capability_adapter.go
therealnb, line 1: ## Changes to optimizer_adapter.go and capability_adapter.go

The optimizer integration introduces a different operational mode for vMCP:

- Normal mode: exposes all backend tools directly to clients
- Optimizer mode: exposes only two optimizer tools (optim_find_tool and optim_call_tool) instead of all backend tools

Why this pattern:

- Consistency: All tool types (backend, composite, optimizer) go through the adapter layer.
- Single responsibility: optimizer_adapter.go handles optimizer tool creation; capability_adapter.go coordinates all adapter operations.
- Dependency inversion: The server uses the interface, not concrete optimizer types.

Why These Changes Were Necessary

- Different tool exposure model: Optimizer mode replaces direct backend tool exposure with two semantic search tools.
- Anti-corruption layer: The adapter isolates the server from optimizer implementation details.
- Testability: The interface allows mocking optimizer handlers in tests.
- Maintainability: Centralizes tool creation logic and keeps the server code clean.

Usage in Server
The server uses these adapters in two places:

- Global registration (startup): Registers optimizer tools globally so they're available immediately.
- Per-session registration: Registers optimizer tools early in session initialization (before other capabilities).
- This ensures optim_find_tool and optim_call_tool are available when clients call list_tools, even before full capability discovery completes.

In summary, these adapters enable optimizer mode while preserving separation of concerns, testability, and consistency with the existing adapter pattern.

cmd/thv-operator/controllers/mcpserver_controller.go
therealnb, line 1: This is a simple change to allow the pull policy to be passed through to the MCP server from the operator. This is useful mostly for development (although, there could be other cases where shuffling images around) when you build a file and you load it into the local kind registry.

Without this, the pods will try to pull the images.

This could be a separate PR.

cmd/thv-operator/pkg/vmcpconfig/converter.go
therealnb, line 1: We could get rid of embeddingService altogether?

pkg/vmcp/client/client.go
therealnb, line 1: As part of the tidy up I took this code out again, but it is required
or in k8s we get

2026/01/23 13:43:04 INFO: listening to server forever
2026/01/23 13:43:04 ERROR: server does not support listening

So, I put it back.

This could be an isolated PR, but lacks context without the optimizer changes.

pkg/vmcp/discovery/manager.go
therealnb: Could be a separate PR

therealnb:
This has been moved to two component PRs https://github.com/stacklok/toolhive/pull/3439 and https://github.com/stacklok/toolhive/pull/3440.

Issue #3335, opened 2026-01-19
Persist dynamically registered OAuth clients across sessions


eleftherias:
Formalising this future enhancement into an issue
https://github.com/stacklok/toolhive/blob/1f6ecc0bf645d55f9f32bf5f0341ce17b58fb6bc/docs/remote-mcp-authentication.md?plain=1#L434

## Problem                                                                                                                                                                            
                                                                                                                                                                                        
When using Dynamic Client Registration (RFC 7591) for remote MCP servers, client credentials are not persisted across restarts. This causes:                                                          
                                                                                                                                                                                        
  1. New OAuth client registered on every `thv run` execution                                                                                                                           
  2. Orphaned client registrations accumulate in OAuth providers                                                                                                                        
  3. Risk of rate limiting from excessive registration requests                                                                                                                                                                                                                                            

## Acceptance Criteria                                                                                                                                                                   
                                                                                                                                                                                        
  - DCR credentials (client_id, client_secret) persisted after registration                                                                                                             
  - DCR metadata (client_secret_expires_at, registration_access_token) stored                                                                                                           
  - Credentials reused across thv run restarts                                                                                                                                          
  - Expired credentials trigger automatic re-registration                                                                                                                               
  - thv secret list shows persisted DCR secrets                                                                                                                                         
  - thv secret rm can remove persisted credentials                                                                                                                                      
  - Backward compatibility maintained for existing configs                                                                                                                              
  - Documentation updated                                                                                                                                                               
                                                                                                                                                                                        
## References                                                                                                                                                                                                                                     
  - https://datatracker.ietf.org/doc/html/rfc7591

carlos-gn:
@eleftherias actually this happened to me with the atlassian mcp. Assign it to me :)

eleftherias:
Hey @carlos-gn, another contributor has created a PR for this issue. They already had a partial solution based on a previous issue they had worked on.
Sorry about the coordination issues here, I hope you didn't spend too much time implementing this already.

If you're still up for contributing, https://github.com/stacklok/toolhive/issues/995 and https://github.com/stacklok/toolhive/issues/883 have been bugging us for a while and could use some eyes.

PR #3382, opened 2026-01-21
feat(auth): persist OAuth tokens across workload restarts

### LINKED ISSUE #3331: Oauth flow persistence and usability
## Bug description
I am trying this for the first time; I trying to connect to atlassian-remote just to get started.
I have the linux keyctl setup and can add/list/get secrets.

stating 
```
> toolhive start atlassian-remote
4:49PM  INFO    Loaded configuration from state for atlassian-remote
4:49PM  INFO    Starting tooling server atlassian-remote...
4:49PM  INFO    Logging to: /home/jpambrun/.local/share/toolhive/logs/atlassian-remote.log
4:49PM  INFO    MCP server is running in the background (PID: 4120044)
4:49PM  INFO    Use 'thv stop atlassian-remote' to stop the server
```
but it doesn't work, the logs indicate the issue
```
4:52PM  INFO    Starting OAuth authentication flow for issuer: https://cf.mcp.atlassian.com
4:52PM  INFO    Successfully registered OAuth client dynamically - client_id: nG58...
4:52PM  INFO    Using OAuth endpoints - authorize_url: https://mcp.atlassian.com/v1/authorize, token_url: https://cf.mcp.atlassian.com/v1/token
4:52PM  INFO    Opening browser to: https://mcp.atlassian.com/v1/authorize?client_id=nG58rBQ_SHP1FKbf&code_challenge=Ryo...&code_challenge_method=S256&redirect_uri=http%3A%2F%2Flocalhost%3A8666%2Fcallback&response_type=code&scope=openid+profile&state=3ELw..
```
I only have 30s to start, cat the logs and go through the flow. This isn't amazing, but it works. 

However, I expected to see some new secrets with `toolhive secret list`, but didn't see any. Restarting the Atlassian mcp I am presented with the same logs requesting another oauth flow.

Am I missing something? 

## Steps to reproduce
Provide steps or commands needed to reproduce the issue.
1. toolhive start atlassian-remote
2. cat logs, and go through the flow
3. toolhive stop atlassian-remote
4. toolhive start atlassian-remote
5. [not authenticated]

## Expected behavior
1. toolhive start atlassian-remote should present me with the oauth login url
2. I expect some sort of secret to be persisted and being able to restart the mcp server without having to go through the flow again.

## Actual behavior
No persistence 

## Environment (if relevant)
- OS/version: linux under WSL
- ToolHive version: v0.7.1 (73d4f262da88e290b9ce82683996cfdf838db432)

## Additional context
Any additional information or logs you think might help.

flfeurmou-indeed:
## Summary

This change enables remote MCP servers (like Datadog and Glean) to restore their OAuth sessions after workload restarts without requiring a new browser-based login.

## Problem

When using remote MCP servers with OAuth authentication (e.g., Datadog, Glean), users must re-authenticate via browser every time:
- The workload is restarted
- The laptop wakes from sleep
- The OAuth access token expires

This creates significant friction, especially for developers who frequently close their laptops (e.g., commuting).

## Solution

Persist OAuth tokens (access token, refresh token, expiry) to the run configuration, enabling automatic session restoration:

1. **Token caching in config**: Added `CachedAccessToken`, `CachedRefreshToken`, `CachedTokenExpiry` fields to `remote.Config`

2. **PersistingTokenSource wrapper**: Intercepts token refreshes and persists updated tokens to disk

3. **Session restoration**: On workload start, checks for cached tokens and restores the session without browser interaction

4. **Automatic refresh persistence**: When tokens are auto-refreshed by the oauth2 library, the new tokens are saved

## Changes

- `pkg/auth/remote/config.go`: Added token caching fields and helper methods
- `pkg/auth/remote/handler.go`: Restore from cached tokens, persist after OAuth flow  
- `pkg/auth/remote/persisting_token_source.go`: New wrapper to persist refreshed tokens
- `pkg/auth/discovery/discovery.go`: Expose token details in `OAuthFlowResult`
- `pkg/runner/runner.go`: Set up token persister callback

## Testing

- Added unit tests for `PersistingTokenSource`
- Added unit tests for `Config.HasValidCachedTokens()` and `Config.ClearCachedTokens()`
- All existing tests pass

## Fixes

Fixes #3331

### DIFF
--- a/pkg/auth/remote/handler.go
+++ b/pkg/auth/remote/handler.go
@@ -10,7 +10,6 @@ import (
 	"golang.org/x/oauth2"
 
 	"github.com/stacklok/toolhive/pkg/auth/discovery"
-	"github.com/stacklok/toolhive/pkg/auth/oauth"
 	"github.com/stacklok/toolhive/pkg/logger"
 	"github.com/stacklok/toolhive/pkg/secrets"
 )
@@ -185,41 +184,6 @@ func (h *Handler) wrapWithPersistence(result *discovery.OAuthFlowResult) oauth2.
 	return tokenSource
 }
 
-// getClientCredentials returns client credentials for token refresh.
-// If configured credentials exist, they are used. Otherwise, if a registration
-// endpoint is available, dynamic client registration is performed.
-// This allows token restoration to work even when the original client was
-// dynamically registered (client credentials are re-obtained on each run).
-func (h *Handler) getClientCredentials(
-	ctx context.Context,
-	scopes []string,
-	authServerInfo *discovery.AuthServerInfo,
-) (clientID, clientSecret string, err error) {
-	// If client credentials are configured, use them
-	if h.config.ClientID != "" {
-		return h.config.ClientID, h.config.ClientSecret, nil
-	}
-
-	// No configured credentials - try dynamic registration
-	if authServerInfo == nil || authServerInfo.RegistrationEndpoint == "" {
-		return "", "", fmt.Errorf("no client credentials configured and dynamic registration not available")
-	}
-
-	logger.Infof("No client credentials configured, performing dynamic client registration")
-
-	// Create registration request
-	registrationRequest := oauth.NewDynamicClientRegistrationRequest(scopes, h.config.CallbackPort)
-
-	// Perform dynamic registration
-	response, err := oauth.RegisterClientDynamically(ctx, authServerInfo.RegistrationEndpoint, registrationRequest)
-	if err != nil {
-		return "", "", fmt.Errorf("dynamic client registration failed: %w", err)
-	}
-
-	logger.Infof("Successfully obtained client credentials via dynamic registration")
-	return response.ClientID, response.ClientSecret, nil
-}
-
 // tryRestoreFromCachedTokens attempts to create a TokenSource from cached tokens
 func (h *Handler) tryRestoreFromCachedTokens(
 	ctx context.Context,
@@ -237,16 +201,10 @@ func (h *Handler) tryRestoreFromCachedTokens(
 		return nil, fmt.Errorf("failed to retrieve cached refresh token: %w", err)
 	}
 
-	// Get client credentials - may need dynamic registration
-	clientID, clientSecret, err := h.getClientCredentials(ctx, scopes, authServerInfo)
-	if err != nil {
-		return nil, fmt.Errorf("failed to obtain client credentials: %w", err)
-	}
-
 	// Build OAuth2 config for token refresh
 	oauth2Config := &oauth2.Config{
-		ClientID:     clientID,
-		ClientSecret: clientSecret,
+		ClientID:     h.config.ClientID,
+		ClientSecret: h.config.ClientSecret,
 		Scopes:       scopes,
 		Endpoint: oauth2.Endpoint{
 			AuthURL:  h.config.AuthorizeURL,


pkg/runner/runner.go
```diff
@@ -561,6 +561,19 @@ func (r *Runner) handleRemoteAuthentication(ctx context.Context) (oauth2.TokenSo
...
 	authHandler := remote.NewHandler(r.Config.RemoteAuthConfig)
 
+	// Set up token persister to save tokens across restarts
+	authHandler.SetTokenPersister(func(accessToken, refreshToken string, expiry time.Time) error {
+		r.Config.RemoteAuthConfig.CachedAccessToken = accessToken
+		r.Config.RemoteAuthConfig.CachedRefreshToken = refreshToken
+		r.Config.RemoteAuthConfig.CachedTokenExpiry = expiry
+
+		// Save the updated config to persist tokens
+		if err := r.Config.SaveState(ctx); err != nil {
```
amirejaz: This will store the access and refresh tokens in plain text in the run configuration which can be a security concern.

We can use the Toolhive secret manager to store the token reference as we are doing with Bearer Token and Client Secret.

pkg/runner/runner.go
```diff
@@ -561,6 +561,19 @@ func (r *Runner) handleRemoteAuthentication(ctx context.Context) (oauth2.TokenSo
 	// Create remote authentication handler
 	authHandler := remote.NewHandler(r.Config.RemoteAuthConfig)
 
+	// Set up token persister to save tokens across restarts
+	authHandler.SetTokenPersister(func(accessToken, refreshToken string, expiry time.Time) error {
+		r.Config.RemoteAuthConfig.CachedAccessToken = accessToken
```
amirejaz: I was wondering if we could avoid saving access_token, we can generate one using the refresh token?

pkg/auth/remote/handler.go
```diff
@@ -34,82 +41,210 @@ func (h *Handler) Authenticate(ctx context.Context, remoteURL string) (oauth2.To
...
+	}
+
+	// Priority 3: OAuth authentication (if configured or detected)
+	return h.authenticateWithOAuth(ctx, remoteURL)
+}
+
+// authenticateWithOAuth detects and handles OAuth authentication requirements
+func (h *Handler) authenticateWithOAuth(ctx context.Context, remoteURL string) (oauth2.TokenSource, error) {
+	// Try to detect if authentication is required
 	authInfo, err := discovery.DetectAuthenticationFromServer(ctx, remoteURL, nil)
```
amirejaz, line 52: nit: this is called in both Priority 2 and 3. We can call `DetectAuthenticationFromServer` and `discoverIssuerAndScopes` once (before Priority 2) and pass the `discoverIssuerAndScopes` result into both `tryRestoreFromCachedTokens` and `authenticateWithOAuth`.

pkg/runner/runner.go
flfeurmou-indeed: Good catch! Updated to use the secret manager. Refresh tokens now go in the OS keyring, and the config just stores the reference (like OAUTH_REFRESH_TOKEN_workload). Added graceful fallback if secrets aren't set up yet.

pkg/runner/runner.go
flfeurmou-indeed: Yeah that makes sense. Removed it. We only persist the refresh token now since the access token gets regenerated on restore anyway.

pkg/auth/remote/handler.go
flfeurmou-indeed: Done. Moved both calls up to Authenticate and pass the results down to tryRestoreFromCachedTokens and performOAuthFlow. Deleted the authenticateWithOAuth wrapper since it's no longer needed.

pkg/auth/remote/persisting_token_source.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Check if the token was refreshed (access token changed)
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.lastToken == nil || token.AccessToken != p.lastToken.AccessToken {
+		// Token was refreshed, persist the refresh token
```
amirejaz: Nit: Refresh tokens are long-lived and usually don’t change on every access-token refresh. We could avoid unnecessary writes by persisting only when a non-empty refresh token differs from the previously stored one.

amirejaz:
@flfeurmou Thanks for addressing the feedback. This looks solid now, I’ve added one additional comment. Sorry for the fragmented review.

pkg/auth/remote/persisting_token_source.go
flfeurmou-indeed: Good point. Updated to only persist when the refresh token is non-empty and actually different from what we stored. Pushed just now.

amirejaz:
@flfeurmou-indeed During further testing, I realized we need to either persist the client credentials (i.e., `client_id` and `client_secret`) or dynamically re-register the client. Otherwise, token reuse may fail for some mcp servers because they require client credentials to be passed. I tested against one such server and observed the following error:

```
5:00PM WARN Failed to restore from cached tokens, will perform fresh OAuth flow: cached tokens are invalid or expired: oauth2: "invalid_client" "Client ID is required"
```

We already have an issue open to track persisting client credentials: https://github.com/stacklok/toolhive/issues/3335.
For this PR, we can re-register the client on each run, and later enhance this to store and reuse the client credentials.

flfeurmou-indeed:
Added getClientCredentials() that does dynamic registration when no client_id is configured. So when restoring from cached tokens, we re-register the client first (if DCR is available), then use those credentials for the token refresh. This keeps this PR simple and defers the full client credentials persistence (#3335) to a separate PR.

amirejaz:
@flfeurmou-indeed I tested the flow with the latest code and noticed that some servers reject newly generated client credentials because they differ from the original ones. I got the following error:

```
7:51PM WARN Failed to restore from cached tokens, will perform fresh OAuth flow: cached tokens are invalid or expired: oauth2: "invalid_grant" "Client ID mismatch"
```


I think we can merge this PR based on the second-to-last commit, without re-registering the client, and prioritize #3335 separately. Please let me know if that works for you, or if you have another suggestion.

flfeurmou-indeed:
> @flfeurmou-indeed I tested the flow with the latest code and noticed that some servers reject newly generated client credentials because they differ from the original ones. I got the following error:
> 
> ```
> 7:51PM WARN Failed to restore from cached tokens, will perform fresh OAuth flow: cached tokens are invalid or expired: oauth2: "invalid_grant" "Client ID mismatch"
> ```
> 
> I think we can merge this PR based on the second-to-last commit, without re-registering the client, and prioritize #3335 separately. Please let me know if that works for you, or if you have another suggestion.

Works for me. Reverted the DCR commit, so this PR is back to just the token persistence with the refresh-token-change optimization. Happy to merge as-is and tackle #3335 as a quick follow-up since this PR lays most of the groundwork. Let me know if there's anything else needed.

amirejaz:
@flfeurmou-indeed Approved 👍 we can merge once CI is green.

amirejaz:
@flfeurmou-indeed I was going to assign #3335 to you, but it’s already assigned to someone else.

flfeurmou-indeed:
@amirejaz sounds good. I'd be happy to take #3335 if carlos-gn hasn't started. Just let me know!

flfeurmou-indeed:
> @amirejaz sounds good. I'd be happy to take #3335 if carlos-gn hasn't started. Just let me know!

For context, I already implemented and tested this locally while working on my setup. Handles PKCE flows (empty client_secret) and works with both Datadog and Glean. Ready to open a PR whenever.

amirejaz:
@flfeurmou-indeed awesome, please file the PR. I will review it tomorrow.

PR #3384, opened 2026-01-21
fix(3383): loss of pod causes stdio transport to stop

### LINKED ISSUE #3383: [BUG] stdio: transport enters zombie state when re-attachment fails
## Bug description

When the Docker daemon becomes unavailable while a ToolHive stdio transport proxy is running, the transport enters a "zombie" state where it appears to be running but cannot communicate with the container. The proxy continues accepting requests but is unable to forward them, and the automatic restart mechanism is not triggered.

This affects workloads using the **stdio transport** .

## Steps to reproduce

1. Start a ToolHive MCP server using stdio transport:
   ```bash
   thv run --name test-server ghcr.io/some/mcp-server
   ```

2. While the server is running, make the Docker daemon unavailable (e.g., stop the Docker service, or simulate network issues to the Docker socket)

3. The container's stdout pipe will close, triggering the stdio transport's re-attachment logic

4. Re-attachment fails because Docker is unavailable

5. Observe that:
   - The proxy runner continues running
   - `IsRunning()` returns `true`
   - The proxy accepts HTTP requests but cannot forward them to the container
   - No automatic restart is triggered

## Expected behavior

When stdio transport re-attachment fails due to Docker being unavailable, the transport should:
1. Call `Stop()` to properly shut down
2. Close the shutdown channel so `IsRunning()` returns `false`
3. Allow the monitoring loop in `Runner.Run()` to detect the stopped transport
4. Trigger the automatic restart flow with exponential backoff

## Actual behavior

When stdio transport re-attachment fails:
1. `processStdout` exits its read loop but does NOT call `Stop()`
2. `processMessages` continues running
3. The HTTP proxy continues accepting requests
4. `IsRunning()` returns `true` (shutdown channel is still open)
5. The monitoring loop never detects that the transport has stopped
6. The system remains in a zombie state indefinitely

## Workaround

The workload must be manually stopped and restarted.

## Environment (if relevant)
- OS/version: macOS / Linux
- ToolHive version: All versions with stdio transport re-attachment logic

## Additional context

The issue is in `pkg/transport/stdio.go` in the `processStdout` function. After `attemptReattachment` fails and returns `false`, the function logs "Container stdout closed - exiting read loop" and returns, but it does not call `t.Stop(ctx)` to properly shut down the transport.

**Relevant log output showing the time gap between stdout closing and transport shutdown:**
```
2025-05-29T00:50:09.464Z WARN Container stdout closed - checking if container is still running
2025-05-29T00:50:09.504Z WARN Docker socket unavailable (attempt 1/3), will retry: Cannot connect to the Docker daemon
...
2025-05-29T00:50:09.574Z WARN Failed to re-attach after all retry attempts
2025-05-29T00:50:09.574Z INFO Container stdout closed - exiting read loop
[... 13+ minutes of zombie state ...]
2025-05-29T01:03:55.447Z INFO Stopping stdio transport
```

**Fix:** Add a call to `t.Stop(ctx)` after the "exiting read loop" log message in `processStdout` so the stdio transport properly shuts down when re-attachment fails.

jerm-dro:
## Summary

Fixes #3383

When the Docker daemon becomes unavailable while a stdio transport proxy is running, the transport would enter a "zombie" state where `IsRunning()` returns `true` but it cannot communicate with the container. This prevented the automatic restart mechanism from triggering. The fix ensures the transport properly shuts down when re-attachment fails.

## Changes

The fix here is actually quite small, but I found the existing recovery code hard to understand. Consequently, I made slight refactors to backfill unit tests to document the actual relationship between the transport and the runner's monitoring behavior.

### Fix: Call `Stop()` after failed re-attachment (`pkg/transport/stdio.go`)

Added a call to `t.Stop(ctx)` in `processStdout` after re-attachment fails. Previously, the function would log "Container stdout closed - exiting read loop" and return without shutting down the transport, leaving it in a zombie state where:
- `processMessages` continued running
- The HTTP proxy continued accepting requests
- `IsRunning()` returned `true`
- The monitoring loop never detected the failure

### Refactor: Extract monitoring loop into `monitorTransport` (`pkg/runner/runner.go`)

Extracted the transport monitoring goroutine from `Run()` into a standalone `monitorTransport` function with a `transportChecker` function type alias. This improves testability and reduces code in the main `Run()` method.

### Refactor: Extract transport stopped handling into `handleTransportStopped` (`pkg/runner/runner.go`)

Extracted the `case <-doneCh` handling logic into a testable `handleTransportStopped` function with a `transportStoppedDeps` struct for dependency injection. This allows unit testing the restart decision logic without complex integration setup.

### Test updates (`pkg/transport/stdio_test.go`)

- Added `TestTransport_ZombieStateWhenDockerUnavailable` to verify the fix
- Updated `TestProcessStdout_EOFWithFailedReattachment`, `TestConcurrentReattachment`, and `TestStdinRaceCondition` with additional mock expectations for `Stop()` and `StopWorkload()` calls
- Updated test assertions to reflect the new behavior where stdin/stdout are nil after `Stop()` is called

## Testing

- [x] `TestTransport_ZombieStateWhenDockerUnavailable` - verifies transport shuts down properly when Docker is unavailable during re-attachment
- [x] `Test_monitorTransport` - verifies monitoring loop detects stopped transport, continues polling on errors, and respects context cancellation
- [x] `TestHandleTransportStopped` - verifies restart decision logic (restart when workload exists, no restart when removed, restart on check failure)
- [x] All existing transport and runner tests pass
- [ ] Manual e2e test with real docker outage

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -403,38 +403,9 @@ func (r *Runner) Run(ctx context.Context) error {
 		logger.Info("Press Ctrl+C to stop or wait for container to exit")
 	}
 
-	// Create a done channel to signal when the server has been stopped
-	doneCh := make(chan struct{})
-
-	// Start a goroutine to monitor the transport's running state
-	go func() {
-		for {
-			// Safely check if transportHandler is nil
-			if transportHandler == nil {
-				logger.Info("Transport handler is nil, exiting monitoring routine...")
-				close(doneCh)
-				return
-			}
-
-			// Check if the transport is still running
-			running, err := transportHandler.IsRunning(ctx)
-			if err != nil {
-				logger.Errorf("Error checking transport status: %v", err)
-				// Don't exit immediately on error, try again after pause
-				time.Sleep(1 * time.Second)
-				continue
-			}
-			if !running {
-				// Transport is no longer running (container exited or was stopped)
-				logger.Warn("Transport is no longer running, attempting automatic restart...")
-				close(doneCh)
-				return
-			}
-
-			// Sleep for a short time before checking again
-			time.Sleep(1 * time.Second)
-		}
-	}()
+	// Start monitoring the transport's running state
+	// monitorTransport returns a channel that will be closed when the transport stops
+	transportStoppedCh := monitorTransport(ctx, transportHandler.IsRunning, 1*time.Second)
 
 	// At this point, we can consider the workload started successfully.
 	// However, we should preserve unauthenticated status if it was already set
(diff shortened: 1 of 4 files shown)


jerm-dro:
Closing because I was not able to reproduce the underlying issue that motivates this change

PR #3386, opened 2026-01-21
Add bearer token support for remote auth proxy

jhrozek:
Previously, specifying --remote-auth-bearer-token without also setting --remote-auth or --remote-auth-client-id would silently ignore the bearer token. The authentication flow was never triggered.

Add shouldHandleOutgoingAuth() that also checks for bearer token configuration via flag, file, or environment variable. When a bearer token is configured, it is now used as Priority 1 in the authentication flow, before OAuth discovery.

This enables using a pre-obtained bearer token for token exchange without requiring a full OAuth configuration.

### DIFF
--- a/cmd/thv/app/proxy.go
+++ b/cmd/thv/app/proxy.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"os"
 	"os/signal"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


yrobla:
fix looks good, but could you add some testing to avoid regressions and ensure this doesn't fail again?

Derek2Tu:
@jhrozek what's the status of this issue?

jhrozek:
> @jhrozek what's the status of this issue?

this somehow fell off my radar @Derek2Tu but it's a really small issue that I just happened to run into in my testing. Are you aware of anyone needing the fix (because I assume if yes then they actually want something different fixed...)

PR #3389, opened 2026-01-21
Strip endpoint prefix from incoming request paths

amirejaz:
## Summary

Implements automatic stripping of `endpointPrefix` from incoming request paths in the transparent HTTP proxy. This enables ingress routing scenarios where requests arrive with a prefix (e.g., `/playwright/sse`) that needs to be removed before forwarding to the backend MCP server.

## Changes

- **Added `stripEndpointPrefix()` function**: Strips prefix from request paths with intelligent handling
- **Added `mcpServerBasePath` tracking**: Tracks MCP server base path (from `remoteURL` for remote servers, or transport type for local servers)
- **Added `normalizePath()` helper**: Normalizes paths for consistent comparison
- **Updated proxy director**: Strips prefix before forwarding requests to backend
- **Added `GetMCPServerBasePath()` helper**: Single source of truth for MCP server paths

### Normal Stripping
- `/playwright/sse` with `endpointPrefix=/playwright`, `mcpServerBasePath=/sse` → `/sse` ✓

### Prefix Matches Server Path
- `/playwright/sse` with `endpointPrefix=/playwright`, `mcpServerBasePath=/playwright` → `/playwright/sse` (preserved) ✓
- `/sse/sse/endpoint` with both `/sse` → `/sse/endpoint` (strips duplicate) ✓

### Edge Cases
- Prefix not at start → no change
- Exact match → `/`
- Empty prefix/path → no change

## Files Modified

- `pkg/transport/proxy/transparent/transparent_proxy.go`
- `pkg/transport/http.go`
- `pkg/transport/url.go`
- `pkg/transport/proxy/transparent/transparent_test.go`

## Testing

Comprehensive test coverage with 20+ test cases covering all scenarios. All tests pass.

### DIFF
--- a/pkg/transport/http.go
+++ b/pkg/transport/http.go
@@ -212,69 +212,16 @@ func (t *HTTPTransport) Start(ctx context.Context) error {
 		return fmt.Errorf("container deployer not set")
 	}
 
-	// Determine target URI
-	var targetURI string
-
-	if t.remoteURL != "" {
-		// For remote MCP servers, construct target URI from remote URL
-		remoteURL, err := url.Parse(t.remoteURL)
-		if err != nil {
-			return fmt.Errorf("failed to parse remote URL: %w", err)
-		}
-		targetURI = (&url.URL{
-			Scheme: remoteURL.Scheme,
-			Host:   remoteURL.Host,
-		}).String()
-		logger.Infof("Setting up transparent proxy to forward from host port %d to remote URL %s",
-			t.proxyPort, targetURI)
-	} else {
-		if t.containerName == "" {
-			return transporterrors.ErrContainerNameNotSet
-		}
-
-		// For local containers, use the configured target URI
-		if t.targetURI == "" {
-			return fmt.Errorf("target URI not set for HTTP transport")
-		}
-		targetURI = t.targetURI
-		logger.Infof("Setting up transparent proxy to forward from host port %d to %s",
-			t.proxyPort, targetURI)
-	}
-
-	// Create middlewares slice
-	var middlewares []types.NamedMiddleware
-
-	// Add the transport's existing middlewares
-	middlewares = append(middlewares, t.middlewares...)
-
 	isRemote := t.remoteURL != ""
 
-	// Add OAuth token injection middleware for remote authentication if we have a token source
-	if isRemote && t.tokenSource != nil {
-		tokenMiddleware := t.createTokenInjectionMiddleware()
-		middlewares = append(middlewares, types.NamedMiddleware{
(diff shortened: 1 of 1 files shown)


pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -371,12 +380,87 @@ func (t *tracingTransport) detectInitialize(body []byte) bool {
...
+		return "/"
+	}
+	// Ensure it starts with "/"
+	if !strings.HasPrefix(path, "/") {
+		path = "/" + path
+	}
+	// Remove trailing slash, but keep "/" as "/"
+	if path != "/" {
+		path = strings.TrimSuffix(path, "/")
+	}
```
blkt, line 396: **concern:** I'm not sure this is correct, we should probably avoid doing this.

Some routers are not very strict (or are they? 🤷) and allow the user to specify a trailing slash in their paths, mapping `/foo` and `/foo/` to different handlers. This is usually the case when`/foo/{param}` is accepted and an empty string is passed.

At the proxy level, knowing whether the trailing slash is desired or not requires intimate knowledge of the MCP server. We should let the Client shoot itself in the foot and let it figure out whether the trailing slash is required or not.

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -371,12 +380,87 @@ func (t *tracingTransport) detectInitialize(body []byte) bool {
...
 // modifyResponse modifies HTTP responses based on transport-specific requirements.
 // Delegates to the appropriate ResponseProcessor based on transport type.
 func (p *TransparentProxy) modifyResponse(resp *http.Response) error {
 	return p.responseProcessor.ProcessResponse(resp)
 }
 
+// stripEndpointPrefix removes the endpointPrefix from the beginning of the request path if present.
+// This handles cases where:
+// 1. Client makes direct request to proxy with prefix (e.g., /playwright/sse)
+// 2. Ingress doesn't strip the prefix before forwarding
```
blkt, line 409: **concern:** could you clarify which scenario is solved by this?

My concern is that we would be adding logic that belongs to the load balancer or HTTP gateway, so I would gather a list of occurrences of these issues in the wild before introducing potentially confusing logic.

All major ingress controllers from both cloud providers and independent vendors (e.g. kong, istio, traefik, haproxy) support some form of prefix stripping in more or less convoluted ways.

amirejaz:
@blkt I agree that the issue is more related to the infrastructure than the proxy itself. A better solution would be to address this at the infrastructure level. I’m going to close this PR.

PR #3393, opened 2026-01-22
Debugging some failures that only occur in CI

jerm-dro:


jerm-dro:
Will cherry-pick the debugging capabilities into a new PR.

PR #3407, opened 2026-01-22
Add KeyManager for signing key lifecycle to authserver

jhrozek:
This PR introduces signing key management for the OAuth authorization server, consisting of:

  - KeyStorage interface defining CRUD operations for signing key persistence. This enables is to add more key storage back ends in the future.
  - MemoryStorage implementation of KeyStorage supporting multiple keys for future rotation
  - KeyManager that handles key generation policy, algorithm selection, and RFC 7638 JWK Thumbprint derivation for deterministic key IDs.

When working on the original MVP of the authserver, having to generate signing keys manually was a complaint I received - production and hardened deployments would mount signing keys generated according to the organization requirements, but in order to just kick the tires in dev mode, it's been inconvenient.

Rather than requiring operators to pre-generate and configure keys, the KeyManager enables automatic key generation with secure defaults (ES256/P-256).

When no signing key is configured, the server generates an ephemeral key on startup—convenient for development while logging a warning about token invalidation on restart.

### DIFF
--- a/pkg/authserver/server/keys/config.go
+++ b/pkg/authserver/server/keys/config.go
@@ -27,10 +27,15 @@ type Config struct {
 	// These keys are included in the JWKS endpoint for token verification but are NOT
 	// used for signing new tokens.
 	//
-	// This enables key rotation:
-	//   - Deploy new signing key as SigningKeyFile
-	//   - Move old key filename to FallbackKeyFiles
-	//   - Tokens signed with old keys remain verifiable until they expire
+	// Key rotation (single replica): update SigningKeyFile to the new key and move
+	// the old filename here. Tokens signed with old keys remain verifiable until
+	// they expire.
+	//
+	// Key rotation (multiple replicas): to avoid a window where one replica signs
+	// with a key not yet advertised by another replica's JWKS endpoint:
+	//  1. Add the new key to FallbackKeyFiles and roll out to all replicas.
+	//  2. Promote it to SigningKeyFile, move the old key to FallbackKeyFiles, roll out.
+	//  3. Remove the old key from FallbackKeyFiles after its tokens have expired.
 	FallbackKeyFiles []string
 }
 


pkg/authserver/storage/types.go
```diff
@@ -268,6 +269,47 @@ type UserStorage interface {
...
+
+	// StoreSigningKey persists a signing key.
+	// This overwrites any existing key with the same KeyID.
+	StoreSigningKey(ctx context.Context, key *SigningKeyData) error
+
+	// GetPublicKeys returns all public keys for the JWKS endpoint.
+	// This includes the current key and any rotated keys still within
+	// their grace period, enabling clients to validate tokens signed
+	// with either old or new keys during rotation.
+	GetPublicKeys(ctx context.Context) ([]*SigningKeyData, error)
```
tgrunnagle: Nit: the function name seems a little misleading because the `SigningKeyData` contains private key functionality.

pkg/authserver/storage/types.go
```diff
@@ -268,6 +269,47 @@ type UserStorage interface {
...
+
+// KeyStorage provides signing key persistence.
+// Storage implementations handle only CRUD operations; key generation
+// and lifecycle management are handled by the keys.Manager.
+//
+// This interface supports future key rotation by allowing multiple keys
+// to be stored. The current signing key is identified by being the most
+// recently created, while older keys remain available for verification
+// during graceful rotation periods.
+type KeyStorage interface {
```
tgrunnagle: Would you help me understand what a prod-ready implementation of `KeyStorage` would look like? Should it support key rotation without restarting the AS? I can imagine a simple case where the AS manifest points to a k8s secret and loads the keys at start up, but then key rotation would require service restarts. Is that sufficient?

pkg/authserver/server/keys/manager.go
```diff
@@ -0,0 +1,148 @@
...
+
+// GetOrCreateSigningKey returns the current signing key, generating one if needed.
+// This is the primary method consumers should use to obtain a signing key.
+//
+// If no key exists in storage, a new key is generated using the configured
+// algorithm, and the Key ID is derived using RFC 7638 JWK Thumbprint.
+//
+// Thread-safety: Uses double-checked locking to ensure only one key is generated
+// even when multiple goroutines call this method concurrently on fresh storage.
+func (m *Manager) GetOrCreateSigningKey(ctx context.Context) (*storage.SigningKeyData, error) {
```
tgrunnagle: Question: Is there any reason to have a `GetSigningKey` that doesn't have the option to create one if it doesn't exist? I'm thinking for production deployments where they want to ensure that keys are read from a persisted storage (e.g. a k8s secret). That or have an ENV var that will cause the function to error out if no key is found.

pkg/authserver/storage/types.go
jhrozek: I pushed a new patch (the previous patches are unchanged) that changes the interface and removes this unsafe shortcut, thanks for pointing it out.

pkg/authserver/storage/types.go
jhrozek: I was actually not envisioning using this storage in prod..I think production would mount secrets from e.g. Vault to files through volumes. Point being I don't think the authserver should be in the business of key storage, I'd prefer to just make it kube-native and read from files.

Which leads me to questioning the design, I no longer think the interface belogs to the storage layer at all.

I'm thinking about moving the interface to the server package and separating two interfaces actually, one to read keys (will have both memory and file back ends) and one to generate keys (memory only).

WDYT?

pkg/authserver/server/keys/manager.go
jhrozek: the GetOrCreate would go away with the change I was thinking about above

pkg/authserver/storage/types.go
tgrunnagle: If not separate interfaces, maybe a `GenerateSigningKey` function (that caches in memory) in this one that can be selectively invoked at start up based on some config?

pkg/authserver/server/keys/config.go
```diff
@@ -0,0 +1,50 @@
...
+	// If empty with KeyDir set, NewProviderFromConfig returns an error.
+	// If both KeyDir and SigningKeyFile are empty, an ephemeral key is generated.
+	SigningKeyFile string
+
+	// FallbackKeyFiles are filenames of additional keys for verification (relative to KeyDir).
+	// These keys are included in the JWKS endpoint for token verification but are NOT
+	// used for signing new tokens.
+	//
+	// This enables key rotation:
+	//   - Deploy new signing key as SigningKeyFile
```
tgrunnagle: I think the key rotation flow would have to look like this to avoid ever signing with something that isn't showing up on the JWKS endpoint (when there are multiple AS instances):
1) deploy NEW key to "Fallback" on all servers
2) move NEW key to "Signing", OLD to "Fallback" on all servers
3) eventually remove OLD from "Fallback"

If you agree we might want to document that in the comment.

pkg/authserver/server/keys/provider.go
```diff
@@ -0,0 +1,233 @@
...
+		algorithm = DefaultAlgorithm
+	}
+	return &GeneratingProvider{algorithm: algorithm}
+}
+
+// SigningKey returns the signing key, generating one if needed.
+// Thread-safe: uses mutex to ensure only one key is generated.
+// Returns a copy to prevent external mutation of internal state.
+func (p *GeneratingProvider) SigningKey(_ context.Context) (*SigningKeyData, error) {
+	p.mu.Lock()
```
tgrunnagle, line 150: nit: can you acquire the lock only if `p.key == nil`?

pkg/authserver/server/keys/provider.go
```diff
@@ -0,0 +1,233 @@
...
+		return nil, err
+	}
+
+	logger.Warnw("generated ephemeral signing key - tokens will be invalid after restart",
+		"algorithm", key.Algorithm,
+		"keyID", key.KeyID,
+	)
+
+	p.key = key
+	return &SigningKeyData{
```
tgrunnagle, line 173: Question for my understanding: are you creating a copy of the key here to avoid accidental manipulation outside of this code?

pkg/authserver/server/keys/config.go
jhrozek: I added a comment to a separate commit for easier review.

I think the procedure you proposed is clever, especially for multi-replica deployments.

pkg/authserver/server/keys/provider.go
jhrozek: no, because then you would have multiple access to the same variable where at least one of them is a write and go flags that as a race.

pkg/authserver/server/keys/provider.go
jhrozek: exactly

PR #3417, opened 2026-01-23
fix: improve stability of remote MCP server connections

### LINKED ISSUE #3419: Remote MCP servers fail after laptop sleep/wake cycles
## Problem

Remote MCP servers (like Datadog and Glean) frequently fail to restart after laptop sleep/wake cycles with errors like:

```
failed to listen: listen tcp 127.0.0.1:52626: bind: address already in use
```

or panics like:

```
http: superfluous response.WriteHeader call from telemetry.(*responseWriter).WriteHeader
```

## Root Cause

1. **Port conflicts after unclean shutdown**: After laptop sleep, zombie toolhive processes may hold ports without releasing them. When the server tries to restart, it fails because the port is still in use.

2. **Panic from duplicate WriteHeader calls**: The telemetry middleware's `responseWriter` can call `WriteHeader` multiple times in certain error scenarios (particularly with reverse proxies), causing panics.

## Expected Behavior

Remote MCP servers should survive laptop sleep/wake cycles and restart cleanly without manual intervention.

## Environment

- macOS (observed on Apple Silicon)
- Toolhive v0.7.x
- Remote MCP servers using OAuth (Datadog, Glean)

## Proposed Solution

1. Add `SO_REUSEADDR` socket option to TCP listeners to allow port reuse after unclean shutdowns
2. Add a guard in the telemetry middleware to prevent duplicate `WriteHeader` calls

flfeurmou-indeed:
## Summary

This PR addresses two issues that cause remote MCP servers (like Datadog and Glean) to fail after laptop sleep/wake cycles.

## Problem

After laptop sleep, remote MCP servers often fail to restart with errors like:
- `failed to listen: listen tcp 127.0.0.1:52626: bind: address already in use`
- `http: superfluous response.WriteHeader call from telemetry.(*responseWriter).WriteHeader`

## Root Cause

1. **Port conflicts after unclean shutdown**: After laptop sleep, zombie toolhive processes may hold ports without releasing them. When the server tries to restart, it fails because the port is still in use.

2. **Panic from duplicate WriteHeader calls**: The telemetry middleware's `responseWriter` can call `WriteHeader` multiple times in certain error scenarios, causing panics in Go's reverse proxy.

### 1. SO_REUSEADDR socket option
Added `SO_REUSEADDR` to the TCP listeners in both `transparent` and `httpsse` proxies. This allows the new process to bind to the port even if a zombie process is still holding it.

```go
lc := net.ListenConfig{
    Control: func(network, address string, c syscall.RawConn) error {
        var opErr error
        if err := c.Control(func(fd uintptr) {
            opErr = unix.SetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEADDR, 1)
        }); err != nil {
            return err
        }
        return opErr
    },
}
```

### 2. WriteHeader guard
Added a `headerWritten` flag to the telemetry middleware's `responseWriter` to prevent duplicate `WriteHeader` calls.

## Testing

Tested with Datadog and Glean MCP servers through multiple sleep/wake cycles on macOS. Both servers now survive laptop sleep and restart cleanly without requiring manual intervention.

## Files Changed

- `pkg/transport/proxy/transparent/transparent_proxy.go` - Added SO_REUSEADDR
- `pkg/transport/proxy/httpsse/http_proxy.go` - Added SO_REUSEADDR  
- `pkg/telemetry/middleware.go` - Added WriteHeader guard

Fixes #3419

### DIFF
--- a/pkg/telemetry/middleware.go
+++ b/pkg/telemetry/middleware.go
@@ -438,7 +438,7 @@ func (rw *responseWriter) Write(data []byte) (int, error) {
 		rw.headerWritten = true
 		rw.statusCode = http.StatusOK // Write() implicitly uses 200 - this is what actually happened
 	}
-	
+
 	n, err := rw.ResponseWriter.Write(data)
 	rw.bytesWritten += int64(n)
 	return n, err
--- a/pkg/telemetry/middleware_test.go
+++ b/pkg/telemetry/middleware_test.go
@@ -671,7 +671,7 @@ func TestResponseWriter_WriteThenWriteHeader(t *testing.T) {
 	assert.Equal(t, len(data), n)
 	assert.Equal(t, int64(len(data)), rw.bytesWritten)
 	assert.Equal(t, string(data), rec.Body.String())
-	
+
 	// Verify that headers were marked as written
 	assert.True(t, rw.headerWritten, "headerWritten should be true after Write() call")
 	assert.Equal(t, http.StatusOK, rw.statusCode, "Status code should be 200 after Write()")


pkg/transport/proxy/httpsse/http_proxy.go
```diff
@@ -161,8 +163,20 @@ func (p *HTTPSSEProxy) Start(_ context.Context) error {
...
 
 	// Create a listener to get the actual port when using port 0
+	// Use ListenConfig with SO_REUSEADDR to allow port reuse after unclean shutdown
 	addr := fmt.Sprintf("%s:%d", p.host, p.port)
-	listener, err := net.Listen("tcp", addr)
+	lc := net.ListenConfig{
+		Control: func(_, _ string, c syscall.RawConn) error {
+			var opErr error
+			if err := c.Control(func(fd uintptr) {
+				opErr = unix.SetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEADDR, 1)
```
JAORMX: This will break Windows builds - `golang.org/x/sys/unix` doesn't compile there.

We use platform-specific files elsewhere for this (see `pkg/workloads/sysproc_*.go`). Consider creating `listener_unix.go` with `//go:build !windows` and a no-op `listener_windows.go`. That would also let you share the code with `transparent_proxy.go` instead of duplicating it.

pkg/telemetry/middleware.go
```diff
@@ -405,12 +405,18 @@ func (*HTTPMiddleware) finalizeSpan(span trace.Span, rw *responseWriter, duratio
...
+	headerWritten bool // Guard against double WriteHeader calls
 }
 
-// WriteHeader captures the status code.
+// WriteHeader captures the status code. Guards against duplicate calls which
+// can cause panics in Go's reverse proxy (http: superfluous response.WriteHeader call).
 func (rw *responseWriter) WriteHeader(statusCode int) {
+	if rw.headerWritten {
+		return // Silently ignore duplicate WriteHeader calls
+	}
```
JAORMX, line 418: Nice fix! Would be good to add a quick test for the duplicate `WriteHeader` scenario to prevent regressions - the existing `TestResponseWriter` only covers single calls.

flfeurmou-indeed:
Thanks for the review! I've addressed the Windows build issue by extracting the `ListenConfig` logic into a new `socket` package with platform-specific implementations:

*   `pkg/transport/proxy/socket/socket_unix.go`: Uses `SO_REUSEADDR` (builds on `!windows`)
*   `pkg/transport/proxy/socket/socket_windows.go`: Uses default configuration (builds on `windows`)

I also fixed the missing license headers and package comments to satisfy the linter.

*Note: The latest linting failure in CI appears to be a transient network timeout fetching the JSON schema (`context deadline exceeded`), unrelated to these changes.*

PR #3418, opened 2026-01-23
Persist DCR client credentials for OAuth session restoration

### LINKED ISSUE #3335: Persist dynamically registered OAuth clients across sessions

flfeurmou-indeed:
## Summary

Extend OAuth token persistence to also store Dynamic Client Registration (DCR) client credentials. This enables workloads that use DCR (like Datadog and Glean) to restore OAuth sessions across restarts without requiring a new browser-based authentication flow.

This PR builds on #3382 which added refresh token persistence, addressing the remaining issue where DCR servers still required re-authentication because the dynamically registered `client_id` (and `client_secret` where applicable) were lost on restart.

## Problem

Remote MCP servers that use Dynamic Client Registration (e.g., Datadog, Glean) would lose their OAuth sessions after workload restarts because:
1. The refresh token was being persisted (#3382 fix)
2. But the DCR-issued `client_id` was not persisted
3. Without the original `client_id`, the token refresh request fails with "Client ID is required" or similar errors

## Solution

Store DCR client credentials in the secret manager alongside refresh tokens:
- `CachedClientIDRef`: Reference to the stored client ID
- `CachedClientSecretRef`: Reference to the stored client secret (when issued)

For PKCE flows (like Datadog), only the `client_id` is persisted since no `client_secret` is issued.

## Changes

- Add `CachedClientIDRef` and `CachedClientSecretRef` fields to `remote.Config`
- Add `HasCachedClientCredentials()` and `ClearCachedClientCredentials()` helper methods
- Add `TokenTypeOAuthClientID` to secrets for secure storage
- Expose `ClientID` and `ClientSecret` in `OAuthFlowResult` from discovery
- Add `ClientCredentialsPersister` type for DCR credential persistence
- Update handler to restore and persist DCR credentials
- Update runner to wire up the client credentials persister
- Regenerate swagger documentation
- Add unit tests for new functionality

## Testing

- All existing tests pass
- New unit tests added for `HasCachedClientCredentials` and `ClearCachedClientCredentials`
- Manually tested with Datadog MCP (PKCE flow) - sessions survive restarts

## Related

Closes #3335

### DIFF
--- a/pkg/auth/remote/config_test.go
+++ b/pkg/auth/remote/config_test.go
@@ -175,14 +175,14 @@ func TestConfig_HasCachedClientCredentials(t *testing.T) {
 		{
 			name: "has cached client ID only",
 			config: Config{
-				CachedClientID: "OAUTH_CLIENT_ID_test",
+				CachedClientID: "test_client_id",
 			},
 			expected: true,
 		},
 		{
 			name: "has both cached credentials",
 			config: Config{
-				CachedClientID:        "OAUTH_CLIENT_ID_test",
+				CachedClientID:        "test_client_id",
 				CachedClientSecretRef: "OAUTH_CLIENT_SECRET_test",
 			},
 			expected: true,
@@ -211,7 +211,7 @@ func TestConfig_ClearCachedClientCredentials(t *testing.T) {
 	t.Parallel()
 
 	config := Config{
-		CachedClientID:        "OAUTH_CLIENT_ID_test",
+		CachedClientID:        "test_client_id",
 		CachedClientSecretRef: "OAUTH_CLIENT_SECRET_test",
 	}
 
--- a/pkg/auth/secrets/secrets.go
+++ b/pkg/auth/secrets/secrets.go
@@ -28,8 +28,6 @@ const (
 	// TokenTypeOAuthRefreshToken represents a cached OAuth refresh token
 	// #nosec G101 - this is a type identifier, not a credential
 	TokenTypeOAuthRefreshToken TokenType = "oauth_refresh_token"
-	// TokenTypeOAuthClientID represents a cached OAuth client ID from Dynamic Client Registration
-	TokenTypeOAuthClientID TokenType = "oauth_client_id" // #nosec G101
 )
 
 // tokenTypeConfig holds configuration for each token type
@@ -54,11 +52,6 @@ var tokenTypeConfigs = map[TokenType]tokenTypeConfig{
 		prefix:       "OAUTH_REFRESH_TOKEN_",
 		target:       "oauth_refresh",
 		errorContext: "OAuth refresh token",
-	},
-	TokenTypeOAuthClientID: {
-		prefix:       "OAUTH_CLIENT_ID_",
-		target:       "oauth_client_id",
-		errorContext: "OAuth client ID",
 	},
 }
 


pkg/auth/remote/config.go
```diff
@@ -54,6 +54,11 @@ type Config struct {
 	// This enables session restoration without requiring a new browser-based login.
 	CachedRefreshTokenRef string    `json:"cached_refresh_token_ref,omitempty" yaml:"cached_refresh_token_ref,omitempty"`
 	CachedTokenExpiry     time.Time `json:"cached_token_expiry,omitempty" yaml:"cached_token_expiry,omitempty"`
+
+	// Cached DCR client credentials for persistence across restarts.
+	// These are obtained during Dynamic Client Registration and needed to refresh tokens.
+	CachedClientIDRef     string `json:"cached_client_id_ref,omitempty" yaml:"cached_client_id_ref,omitempty"`
```
amirejaz: we can save clientID as plain text as this is public

pkg/auth/remote/config.go
```diff
@@ -54,6 +54,11 @@ type Config struct {
...
+	CachedClientSecretRef string `json:"cached_client_secret_ref,omitempty" yaml:"cached_client_secret_ref,omitempty"`
```
amirejaz, line 62: We need to store additional DCR metadata (such as client_secret_expires_at and registration_access_token) and implement handling for client secret expiration in DCR.

flfeurmou-indeed:
Addressed the review feedback:

✅ **ClientID as plain text**: Changed `CachedClientIDRef` to `CachedClientID` - now stored directly in config instead of as a secret reference (since it's public information)

✅ **Added DCR metadata fields**:
- `CachedClientSecretExpiresAt` - tracks when the client secret expires
- `CachedRegistrationAccessTokenRef` - for updating/deleting the client registration

⏸️ **Expiration handling logic**: The fields are in place, but I haven't implemented the actual expiration check and re-registration flow yet. Let me know if that's needed in this PR or if it can be a follow-up.

pkg/auth/secrets/secrets.go
```diff
@@ -28,6 +28,8 @@ const (
 	// TokenTypeOAuthRefreshToken represents a cached OAuth refresh token
 	// #nosec G101 - this is a type identifier, not a credential
 	TokenTypeOAuthRefreshToken TokenType = "oauth_refresh_token"
+	// TokenTypeOAuthClientID represents a cached OAuth client ID from Dynamic Client Registration
+	TokenTypeOAuthClientID TokenType = "oauth_client_id" // #nosec G101
```
amirejaz: this is not needed as we are not storing the secret ref for client id

pkg/auth/remote/config_test.go
```diff
@@ -158,3 +158,69 @@ func TestConfig_UnmarshalJSON_BearerTokenFields(t *testing.T) {
...
+	}{
+		{
+			name:     "no cached credentials",
+			config:   Config{},
+			expected: false,
+		},
+		{
+			name: "has cached client ID only",
+			config: Config{
+				CachedClientID: "OAUTH_CLIENT_ID_test",
```
amirejaz: Nit: This shows the client being saved as a ref; we can use lowercase to depict the actual plaintext.

amirejaz:
@flfeurmou-indeed that makes sense. I’ll create a separate issue to handle client secret expiry.

I’ve added a couple of comments; once those are addressed, we can merge this PR and handle client secret expiry in a follow-up PR.

amirejaz:
@flfeurmou-indeed Thanks for addressing the feedback 🚀 I’ve approved the PR.

Issue #3147, opened 2025-12-23
Issue #1: Add StatusReporter abstraction for vMCP runtime (foundation)


4t8dd:
## Summary

  Implement the core StatusReporter abstraction to enable vMCP runtime to report operational status. This is the foundation for eliminating duplicate backend discovery work between the operator and vMCP runtime.

  This issue implements the **interface and CLI-mode implementation only** (no Kubernetes integration). The K8s implementation will follow in a separate PR to keep changes focused and reviewable.

  ## Motivation

  Currently, the operator must infer vMCP runtime status through polling and discovery. This creates:
  - Duplicate backend discovery work (operator discovers, then vMCP discovers again)
  - Delayed status updates (operator polls periodically)
  - No visibility into vMCP runtime health

  By adding StatusReporter, the vMCP runtime can push status updates directly to the control plane.

  ## Scope of This PR

  ### What's Included ✅
  - Platform-agnostic `Reporter` interface
  - Status model (Phase, Conditions, DiscoveredBackends)
  - `NoOpReporter` for CLI mode
  - Tests for NoOpReporter
  - Basic documentation

  Dependencies

  None - this is a pure code addition with zero external dependencies.

  Acceptance Criteria

  - Reporter interface defined
  - Status model complete with all types
  - NoOpReporter implemented
  - README documents the abstraction
  - No Kubernetes dependencies in this PR
  - Linter passes

related to #2854

4t8dd:
@yrobla ah, you implement again? I already got a PR for this in Dec, waiting for review.

jhrozek:
@4t8dd I apologize, we had a large backlog of PRs during the holiday break and we're still catching up. Yours and Yolanda's are on my list, I'll see about merging them today/tomorrow.

Sorry about that, it was not that we ignore community contributions, it's an oversight because of the large number of PRs/issues we've been handling in the past couple of weeks.

4t8dd:
@jhrozek OK. I got you. 

what I can not understand is why people implements again even with my PR provided? Is this a competition? Are you goging to review both of them to pick better one to accept?

I created that PR before the Chriastmas, And no one can review it.  Now @yrobla  just came out to implement them again without any comments on my PR. why? what is your strategy to communicate with contributors? Let me know this. Thanks.

jhrozek:
> [@jhrozek](https://github.com/jhrozek) OK. I got you.
> 
[...]

PR #3421, opened 2026-01-23
Refactor health status to separate status from reason

yrobla:
Previously, BackendHealthStatus conflated WHAT the status is with WHY the backend is in that state. For example, BackendUnauthenticated was both a status value and a reason, making it unclear whether a backend was unavailable due to authentication issues or other causes.

This improves observability by providing clear separation between the backend's availability state and the underlying cause of that state.

Related-to: #3147

This change introduces BackendHealthReason to explicitly capture WHY a backend is in its current state, while BackendHealthStatus describes WHAT the state is (healthy, degraded, unhealthy, unknown).

### DIFF
--- a/pkg/vmcp/health/checker.go
+++ b/pkg/vmcp/health/checker.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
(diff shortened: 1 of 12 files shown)


yrobla:
i will implement later when we add more integration with k8s as well

PR #3422, opened 2026-01-23
Fix race condition when deleting workload just after creation

dmjb:
During creation of e2e tests, an existing bug was spotted. If a workload was deleted before the proxy-runner created the container, the deletion logic would decide that the workload did not exist and do nothing. Afterwards, the proxy-runner would create the container, leading to the workload not being deleted, or stuck in a half-deleted state.

Improve the deletion logic to make sure that this does not happen.

Tests which cover this will be added in a separate PR.

### DIFF
--- a/pkg/workloads/manager.go
+++ b/pkg/workloads/manager.go
@@ -841,21 +841,21 @@ func (d *DefaultManager) stopProcess(ctx context.Context, name string) {
 	// Try to read the PID and kill the process
 	pid, err := d.statuses.GetWorkloadPID(ctx, name)
 	if err != nil {
-		logger.Errorf("No PID file found for %s, proxy may not be running in detached mode", name)
+		logger.Debugf("No PID file found for %s, proxy may not be running in detached mode", name)
 		return
 	}
 
 	// PID file found, try to kill the process
 	logger.Infof("Stopping proxy process (PID: %d)...", pid)
 	if err := process.KillProcess(pid); err != nil {
-		logger.Warnf("Warning: Failed to kill proxy process: %v", err)
+		logger.Debugf("Warning: Failed to kill proxy process: %v", err)
 	} else {
-		logger.Info("Proxy process stopped")
+		logger.Debugf("Proxy process stopped")
 	}
 
 	// Clean up PID file after successful kill
 	if err := process.RemovePIDFile(name); err != nil {
-		logger.Warnf("Warning: Failed to remove PID file: %v", err)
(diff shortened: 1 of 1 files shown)


pkg/workloads/manager.go
```diff
@@ -867,7 +876,28 @@ func (d *DefaultManager) removeContainer(ctx context.Context, name string) error
...
+				// Container is gone, deletion complete
+				logger.Debugf("Container %s successfully removed from runtime", name)
+				return nil
+			}
+			// Some other error occurred
+			logger.Warnf("Error checking container status during removal: %v", err)
+			return fmt.Errorf("failed to verify container removal: %w", err)
+		}
+		// Container still exists, wait and retry
+		time.Sleep(retryDelay)
```
JAORMX: This polling loop doesn't respect context cancellation - it'll keep waiting up to 3 seconds even if the context is cancelled. Consider adding a check:

```go
for range maxRetries {
    select {
    case <-ctx.Done():
        return fmt.Errorf("context cancelled while waiting for container removal: %w", ctx.Err())
    default:
    }
    // ... rest of the loop
}
```

pkg/workloads/manager.go
```diff
@@ -752,29 +752,38 @@ func (d *DefaultManager) deleteContainerWorkload(ctx context.Context, name strin
...
+		baseName = name
+		isAuxiliary = false
+		logger.Debugf("Container not found for workload %s, proceeding with state cleanup", name)
+	}
 
-		// Clean up associated resources
-		d.cleanupWorkloadResources(ctx, name, baseName, labels.IsAuxiliaryWorkload(containerLabels))
+	// Stop proxy-runner process AFTER container removal to prevent recreation
+	// Skip for auxiliary workloads like inspector that don't use proxy processes
+	if !isAuxiliary {
```
JAORMX, line 778: The old code only called `stopProxyIfNeeded` when `container.IsRunning()` was true. Now it's always called regardless of running state. I assume this is intentional to handle the race where the proxy is still starting up? If so, might be worth a quick comment explaining why.

dmjb: Yeah, the cause of this bug is the assumption that the proxy is only running when the container is running, which is not true. I can add some extra comments to explain further.

dmjb: It seems there is no reason to avoid the call to `stopProxyIfNeeded` except to avoid some warnings from being logged. I changed these to debug logs since, per the new logging guidelines, none of these should be displayed anyway.

Issue #3316, opened 2026-01-15
[FEATURE] Support configurable header passthrough for MCP server requests


Derek2Tu:
**Description:**

Currently, ToolHive passes through certain headers to MCP servers as a side effect of not stripping them, rather than through intentional configuration. We should implement explicit support for specifying additional headers to forward to MCP servers.

**Requirements:**
- Add configuration option to define a list of headers that should be passed through to MCP servers
- Support this functionality in both local and Kubernetes deployment modes
- Make header passthrough behavior intentional and documented rather than incidental

jhrozek:
@tgrunnagle @glageju do we have more details on this?

tgrunnagle:
I think this refers to the cloud_id header for Atlassian - if so the existing header pass-through from the client supports it. @Derek2Tu can you confirm?

jhrozek:
Hey, working on this. Splitting it into a few PRs to keep things reviewable:

1. Core middleware + restricted headers blocklist (sending shortly)
2. RunConfig types and builder options
3. Runner wiring (secret resolution timing matters here)
4. CLI flags (`--remote-forward-headers`, `--remote-forward-headers-secret`) for both `thv proxy` and `thv run`
5. MCPRemoteProxy CRD support
6. Conflict detection with token exchange / OAuth middleware

Following RFC THV-0026 — additive mode, hop-by-hop/smuggling headers blocked, Authorization allowed with a warning unless it conflicts with existing auth middleware.

jhrozek:
FYI, the CLI part of this feature was released today in 0.8.2.

The CRD/k8s part is under review. I'll work on docs as well.

@Derek2Tu should this feature be configurable through the UI as well?

jhrozek:
The operator part was merged and the API extension. UI support is coming up in https://github.com/stacklok/toolhive-studio/issues/1530

jhrozek:
I am going to close this issue since this feature is now available across CLI, UI (toolhive studio) and k8s.

CLI docs: https://docs.stacklok.com/toolhive/guides-cli/run-mcp-servers#run-a-remote-mcp-server
k8s docs: https://docs.stacklok.com/toolhive/guides-k8s/remote-mcp-proxy#inject-custom-headers
In the UI the headers are available when running a remote MCP server.

PR #3423, opened 2026-01-23
Add header forward middleware for remote MCP servers

jhrozek:
Implement middleware that injects configured headers into requests before they are forwarded to remote MCP servers. This enables operators to configure headers server-side, removing the burden from clients.

This commit adds the core middleware; subsequent commits add RunConfig types, CLI flags, runner wiring, and CRD support.

Key design decisions:
- RestrictedHeaders blocklist prevents misconfiguration of hop-by-hop, request smuggling, and identity spoofing headers
- Authorization is allowed with a warning (valid for static tokens)
- Header names are pre-canonicalized at creation time
- Supports both factory pattern (thv run) and direct creation (thv proxy)
- Header values are never logged, only names at DEBUG level

Related: #3316


pkg/transport/middleware/header_forward.go
```diff
@@ -0,0 +1,134 @@
...
+	"Proxy-Authenticate":  true,
+	"Proxy-Connection":    true,
+	// Request smuggling vectors
+	"Transfer-Encoding": true,
+	"Content-Length":    true,
+	// Identity spoofing
+	"X-Forwarded-For":   true,
+	"X-Forwarded-Host":  true,
+	"X-Forwarded-Proto": true,
+	"X-Real-Ip":         true,
```
JAORMX, line 45: Let's add the `Forwarded` header to this list https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Forwarded

JAORMX: Ah, probably `Http2-Settings` should be included as well https://http.dev/http2-settings

jhrozek: thanks, I added a separate commit for easier review

PR #3424, opened 2026-01-23
Add client capacity enforcement to authserver storage layer

jhrozek:
Introduce a configurable maximum client limit (DefaultMaxClients=100) with ErrCapacityExceeded sentinel error. MemoryStorage enforces the limit in RegisterClient and exposes WithMaxClients option.


jhrozek:
Actually, without proper client management and a way to be able to delete clients this could cause more harm than good at this point, let me come back to it later.

PR #3427, opened 2026-01-23
Add header forward configuration and wire middleware in runner

jhrozek:
This PR builds on #3423 (which added the header-forward middleware itself) by adding the configuration model and integrating it into the runner lifecycle.

There are three commits in the PR, especially because one just moves code around and splitting this way makes the PR easier to review.

HeaderForwardConfig is added to RunConfig with two sources of headers:
 - AddPlaintextHeaders for literal values stored as-is, and
 - AddHeadersFromSecret for references to secrets in ToolHive's secret manager that are resolved at runtime. 

Secret values are held only in a non-serialized field (resolvedHeaders), ensuring they are never written to disk via SaveState(). The ResolvedHeaders() accessor provides the merged view for middleware consumption.

Header-forward middleware cannot be created at builder time (WithMiddlewareFromFlags) because secret-backed values aren't yet resolved. Instead, it is added in Runner.Run() after WithSecrets() completes. To support this, Runner.Run() is reordered so secrets are processed before middleware creation. The middleware is placed at the innermost position (just before recovery) so headers are injected closest to the backend.

Related: #3316

### DIFF
--- a/pkg/runner/config_test.go
+++ b/pkg/runner/config_test.go
@@ -1887,6 +1887,15 @@ func TestRunConfig_resolveHeaderForwardSecrets(t *testing.T) {
 			wantResolved:  map[string]string{"X-Existing": "existing-value", "X-New": "resolved-value"},
 			wantPlaintext: map[string]string{"X-Existing": "existing-value"},
 		},
+		{
+			name: "secret overrides plaintext for same header name",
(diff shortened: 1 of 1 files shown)


jhrozek:
> looks good, but a test validating what's the behaviour with same header names will be nice

thanks, good idea, I added one

PR #3428, opened 2026-01-23
Add POST /oauth/register handler for dynamic client registration

jhrozek:
Implement the registration endpoint already advertised in the OAuth discovery metadata. The handler validates requests via the registration package's allowlists, creates a public LoopbackClient, and persists it through the storage.ClientRegistry interface.

### DIFF
--- a/pkg/authserver/server/handlers/dcr.go
+++ b/pkg/authserver/server/handlers/dcr.go
@@ -6,6 +6,7 @@ package handlers
 import (
 	"encoding/json"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
@@ -27,6 +28,16 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
 
 	// Limit request body size to prevent DoS attacks
 	req.Body = http.MaxBytesReader(w, req.Body, maxDCRBodySize)
+
+	// Validate Content-Type header (RFC 7591 requires application/json)
+	contentType := req.Header.Get("Content-Type")
+	if !strings.HasPrefix(contentType, "application/json") {
+		writeDCRError(w, http.StatusBadRequest, &registration.DCRError{
+			Error:            registration.DCRErrorInvalidClientMetadata,
+			ErrorDescription: "Content-Type must be application/json",
+		})
+		return
+	}
 
 	// Parse request body
 	var dcrReq registration.DCRRequest
--- a/pkg/authserver/server/handlers/dcr_test.go
+++ b/pkg/authserver/server/handlers/dcr_test.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	"github.com/ory/fosite"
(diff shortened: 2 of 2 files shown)


pkg/authserver/server/handlers/dcr.go
```diff
@@ -0,0 +1,111 @@
...
+// RegisterClientHandler handles POST /oauth/register requests.
+// It implements RFC 7591 Dynamic Client Registration for public clients
+// with loopback redirect URIs only.
+func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+
+	// Limit request body size to prevent DoS attacks
+	req.Body = http.MaxBytesReader(w, req.Body, maxDCRBodySize)
+
+	// Parse request body
```
yrobla, line 42: should be nice to check if the content type is actually json? also, validating that actually we are receiving a post here?

pkg/authserver/server/handlers/dcr_test.go
```diff
@@ -0,0 +1,150 @@
...
+			expectedStatus: http.StatusCreated,
+		},
+		{
+			name:           "invalid JSON body",
+			requestBody:    "not-valid-json",
+			expectedStatus: http.StatusBadRequest,
+			expectedError:  registration.DCRErrorInvalidClientMetadata,
+		},
+		{
+			name: "validation error propagated",
```
yrobla, line 51: maybe more test, such as sending invalid method, sending oversized body, sending multiple urls

pkg/authserver/server/handlers/dcr.go
jhrozek: Thanks for the review, @yrobla! Good catch on the Content-Type validation.

**Content-Type check:** Added! RFC 7591 does require `application/json`, so I've implemented validation that accepts `application/json` and charset variations. This provides clearer error messages for misconfigured clients.

**HTTP method validation:** The handler is registered via `r.Post()` in chi, which automatically returns 405 Method Not Allowed for non-POST requests before the handler is ever invoked. Adding a manual check inside the handler would be redundant defensive coding, so I've left that to the router layer.

pkg/authserver/server/handlers/dcr_test.go
jhrozek: Thanks for the suggestions on test coverage!

**Oversized body test:** Added! This exercises the `MaxBytesReader` logic directly and documents the 64KB limit in the test suite.

**Invalid method test:** Since the route is registered with `r.Post()`, chi automatically returns 405 for non-POST requests before the handler runs. Testing this would verify chi router behavior rather than handler logic, so I've skipped it for the handler unit tests.

**Multiple redirect URIs:** This is already covered extensively in `registration/dcr_test.go` — see `TestValidateDCRRequest` which includes cases like "valid request with all fields specified" (multiple URIs), "too many redirect URIs" (11 URIs exceeds limit), and "invalid redirect URI in list". The handler test's "validation error propagated" case verifies that these validation errors are correctly returned to clients.

jhrozek:
@yrobla sorry for the delay in addressing the comments, I wrote the patches, replied to you but forgot to push the new code.

PR #3430, opened 2026-01-23
fix: properly detect and kill zombie supervisor processes

### LINKED ISSUE #3429: Remote MCP servers leave zombie processes after laptop sleep, blocking restart
## Problem

After laptop sleep/wake cycles, old `thv` processes sometimes remain running and continue holding their ports. When attempting to restart these servers, the new process fails with:

```
failed to listen: listen tcp 127.0.0.1:XXXXX: bind: address already in use
```

This happens even with #3417 (SO_REUSEADDR), because `SO_REUSEADDR` cannot override an actively running process, it only helps with sockets in `TIME_WAIT` state.

## Observed behavior

1. Laptop goes to sleep with Datadog/Glean MCP servers running
2. Laptop wakes up
3. Old `thv start <server>` processes are still running (visible in `ps aux | grep thv`)
4. `thv start <server>` or `thv restart <server>` fails because the old process holds the port
5. Manual `kill <pid>` is required before the server can start

## Environment

- macOS (Apple Silicon)
- Toolhive v0.7.x
- Remote MCP servers (Datadog, Glean) using OAuth

## Possible solutions

1. **PID file tracking**: Store PID on start, check/kill stale process on restart
2. **Better signal handling**: Ensure processes terminate cleanly on SIGTERM/SIGHUP during sleep
3. **Port-based cleanup**: Before binding, check if another `thv` process holds the port and terminate it
4. **Enhanced `thv restart`**: Make restart explicitly kill the old process before starting new one

## Workaround

Currently using a wrapper script that runs `pkill -f "thv start <server>"` before starting.

## Related

- #3417 (SO_REUSEADDR for TIME_WAIT scenarios)
- #3419 (parent issue for sleep/wake stability)

Happy to submit a PR if you'd like me to implement one of these solutions.

flfeurmou-indeed:
## Summary

This PR fixes the issue where MCP servers fail to restart after laptop sleep/wake cycles due to zombie supervisor processes.

## Problem

After laptop sleep, old supervisor processes may:
1. Still have PID files but not actually be running
2. Be running but unresponsive to SIGTERM signals
3. Hold ports and prevent new servers from starting

The current code only checks if a PID file exists (`GetWorkloadPID`), not if the process is actually alive. Additionally, `KillProcess` only sends SIGTERM which zombie processes may ignore.

### 1. Proper process liveness detection (`manager.go`)

`isSupervisorProcessAlive` now uses `process.FindProcess` to actually check if the process is running (via signal 0), not just if a PID file exists. If a process is dead but its PID file remains, the stale file is cleaned up automatically.

### 2. Forceful kill for zombie processes (`kill_unix.go`)

`KillProcess` now:
1. Sends SIGTERM for graceful shutdown
2. Waits 500ms
3. If process is still alive, sends SIGKILL to force termination

This handles zombie processes that survive laptop sleep and don't respond to SIGTERM.

## Testing

- Updated existing tests to mock the new process checking behavior
- All existing tests pass
- Tested manually with Datadog and Glean MCP servers through sleep/wake cycles

Fixes #3429

## Checklist

- [x] Code follows project style guidelines
- [x] Tests added/updated
- [x] Documentation updated (N/A - internal change)
- [x] Signed-off-by trailer included


amirejaz:
@flfeurmou-indeed Copilot added some comments

flfeurmou-indeed:
Addressed the Copilot review feedback:

✅ **Error handling**: Changed string comparison to `errors.Is(err, os.ErrProcessDone)` for robustness
✅ **Extra blank line**: Removed the extra blank line after imports
✅ **Duplicate comment**: Removed duplicate "DefaultManager" comment, ProcessFinder now has its own proper comment
✅ **Zombie test case**: Added test case for the scenario where PID file exists but process is dead
✅ **Personal file**: Removed `FLFEURMOU_LOCAL_FIXES.md` from the PR

⏸️ **500ms wait time**: Kept as-is for now. The value is documented in the comment. Happy to make it configurable if that's a blocker.

theJC:
❤️ ❤️ 
I am so happy to see this MR, this is a constant problem for me and has got me staying away from leveraging toolhive lately because its so disruptive!    Thank you @flfeurmou-indeed for digging into this issue!

amirejaz:
@flfeurmou-indeed Let’s keep this PR open for now, as it may no longer be needed since we’ve merged the DCR persistence PR. If the issue still persists, we can discuss a better solution for handling this in the file state manager.

dmjb:
I think the change made in  #3543 addresses the underlying problem, closing this one out.

PR #3433, opened 2026-01-24
feat: implement health-aware restart logic for remote workloads

flfeurmou-indeed:
## Summary

This PR implements health-aware restart logic for remote workloads (like Glean and Datadog). It enhances the supervisor check to verify not just that the process exists, but that it is responsive to HTTP health checks.

## Problem

After laptop sleep/wake cycles, remote MCP server proxies often end up in a "zombie" state where the process exists (PID is valid) but the network socket is broken or the application is hung. The current logic only checks for PID existence, so it assumes the workload is healthy and refuses to restart it, or fails to clean it up properly.

## Solution

Updated `isSupervisorHealthy` (formerly `isSupervisorProcessAlive`) to:
1. Check if the process exists (using `process.FindProcess`)
2. For remote workloads, perform an HTTP GET request to the health endpoint
3. Return `false` if either check fails, triggering the restart/cleanup logic

This allows ToolHive to automatically detect and recover from "alive but unresponsive" states.

## Changes

- Updated `DefaultManager` to support dependency injection for health checks
- Implemented `checkHTTPHealth` helper
- Updated `isSupervisorHealthy` to use the new check
- Updated unit tests to cover the new logic and mock the health checks

## Testing

- Added unit tests for healthy/unhealthy supervisor scenarios
- Verified that existing tests pass
- (Manual verification recommended on a real sleep/wake cycle)

### DIFF
--- a/pkg/workloads/manager.go
+++ b/pkg/workloads/manager.go
@@ -1,3 +1,6 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
 // Package workloads contains high-level logic for managing the lifecycle of
 // ToolHive-managed containers.
 package workloads
@@ -6,6 +9,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


pkg/workloads/manager.go
```diff
@@ -809,25 +775,125 @@ func (d *DefaultManager) getWorkloadContainer(ctx context.Context, name string)
 	return &container, nil
 }
 
+// checkHTTPHealth performs a simple HTTP GET request to check health
+func checkHTTPHealth(ctx context.Context, url string) bool {
```
dmjb, line 779: We have these sorts of health checks implemented further down the stack (in the proxy and transport layer).

flfeurmou-indeed:
Closing this PR in favor of the existing healthcheck mechanism. I'll test with the env var enabled once PR #3462 is included in a release, as suggested by @dmjb.

PR #3439, opened 2026-01-26
Infrastructure improvements and bugfixes for vMCP

therealnb:
# Infrastructure Improvements and Bugfixes for vMCP

This PR contains infrastructure improvements and bugfixes that enable better observability, reliability, and performance for vMCP. These changes are independent of the optimizer implementation and can be merged separately.

### Observability & Tracing

**OpenTelemetry Tracing in Capability Aggregation** (`pkg/vmcp/aggregator/default_aggregator.go`)
- Added comprehensive tracing spans to all aggregator methods
- Includes spans for `AggregateCapabilities`, `QueryAllCapabilities`, `QueryCapabilities`, `ResolveConflicts`, and `MergeCapabilities`
- All spans include relevant attributes (backend counts, tool/resource/prompt counts, error recording)
- **Why needed**: Provides visibility into capability aggregation performance and helps diagnose issues in distributed tracing systems like Jaeger

### Performance Improvements

**Singleflight Deduplication for Discovery** (`pkg/vmcp/discovery/manager.go`)
- Added singleflight group to prevent duplicate aggregation work when multiple requests arrive concurrently
- Ensures only one aggregation happens per cache key at a time
- Includes double-check cache pattern after acquiring singleflight lock
- **Why needed**: Prevents wasted CPU cycles and reduces backend load when multiple clients request capabilities simultaneously. Critical for high-traffic scenarios.

### Health Check Improvements

**Self-Check Prevention** (`pkg/vmcp/health/checker.go`)
- Added `selfURL` field to prevent server from health-checking itself
- Implements URL normalization for comparison (handles localhost/127.0.0.1 variations)
- Short-circuits health checks targeting the server's own URL
- **Why needed**: Prevents wasteful self-checks and potential connection issues during startup. Improves server efficiency.

**Health Check Context Marker** (`pkg/vmcp/health/checker.go`)
- Added `WithHealthCheckMarker` to bypass authentication logging for health checks
- Health checks verify backend availability and should not require user credentials
- **Why needed**: Reduces noise in authentication logs and clarifies that health checks are system-level operations, not user requests.

### Client Reliability Fixes

**HTTP Client Timeout** (`pkg/vmcp/client/client.go`)
- Added 30-second timeout to HTTP client to prevent hanging connections
- Removed `WithContinuousListening()` in favor of timeout-based approach
- Set timeout for streamable HTTP client as well
- **Why needed**: Prevents connections from hanging indefinitely, which can cause resource exhaustion and make the system unresponsive. Critical for production reliability.

### Test Reliability Improvements

**E2E Test Pod Readiness Checks** (`test/e2e/thv-operator/virtualmcp/helpers.go`)
- Skip completed pods (Succeeded, Failed) in readiness checks
- Only check pods in Running phase
- Added validation to ensure at least one running pod exists
- **Why needed**: Prevents flaky test failures when old pod deployments exist. Tests were failing because they checked all pods including completed ones from previous test runs.

**E2E Test HTTP Client Timeout** (`test/e2e/thv-operator/virtualmcp/virtualmcp_auth_discovery_test.go`)
- Added HTTP client with 10-second timeout for health checks
- Added pod readiness verification before health endpoint checks
- Improved error messages for connection reset issues
- **Why needed**: Prevents test timeouts and makes test failures more diagnosable. Ensures pods are actually ready before attempting health checks.

### Build & Infrastructure

**Gitignore Updates** (`.gitignore`)
- Added `cmd/vmcp/__debug_bin*` patterns for debug binaries
- **Why needed**: Prevents debug binaries from being accidentally committed

**Linter Configuration** (`.golangci.yml`)
- Excluded `scripts$` directory from linting
- **Why needed**: Scripts directory contains helper scripts that don't need to follow strict Go linting rules

**Code Coverage Configuration** (`codecov.yaml`)
- Excluded test coverage files (`*_test_coverage.go`) from coverage reports
- **Why needed**: These are generated coverage test files and shouldn't be included in coverage calculations

**Chart Version Bump** (`deploy/charts/operator-crds/Chart.yaml`)
- Updated version to 0.0.97
- **Why needed**: Standard version bump after changes

## Testing

All existing tests pass. The changes improve test reliability and don't break any existing functionality.

## Related

This PR is split from #3373. The optimizer implementation will follow in a separate PR that depends on these infrastructure improvements.

## Large PR Justification
- These changes were requested as one PR


pkg/vmcp/aggregator/default_aggregator.go
```diff
@@ -43,12 +48,20 @@ func NewDefaultAggregator(
 		backendClient:    backendClient,
 		conflictResolver: conflictResolver,
 		toolConfigMap:    toolConfigMap,
+		tracer:           otel.Tracer("github.com/stacklok/toolhive/pkg/vmcp/aggregator"),
```
jerm-dro: Blocking: to follow the conventions in the codebase, you should pass in a `TracerProvider` like here:

https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/server/server.go#L281

We don't register the global tracer until here: https://github.com/stacklok/toolhive/blob/main/pkg/telemetry/config.go#L226, so I think you _might_ get a tracer that does nothing if you directly get the tracer from otel like you've done here.

jerm-dro: FWIW, I think this tracer provider machinery is over-engineered, but I think it's best to follow the conventions we have.

pkg/vmcp/aggregator/default_aggregator.go
```diff
@@ -86,6 +107,13 @@ func (a *defaultAggregator) QueryAllCapabilities(
 	ctx context.Context,
 	backends []vmcp.Backend,
 ) (map[string]*BackendCapabilities, error) {
+	ctx, span := a.tracer.Start(ctx, "aggregator.QueryAllCapabilities",
+		trace.WithAttributes(
+			attribute.Int("backends.count", len(backends)),
+		),
+	)
+	defer span.End()
```
jerm-dro: Suggestion: capture the error & status in the defer like so:

```go
) (_ map[string]*BackendCapabilities, retErr error) {
   ctx, span := a.tracer.Start(ctx, "aggregator.QueryAllCapabilities",
		trace.WithAttributes(
			attribute.Int("backends.count", len(backends)),
		),
	)
	defer func () {
       if retErr != nil {
         span.RecordError(retErr)
   	     span.SetStatus(codes.Error, retErr.Error())
       }
       span.End()
    }()
```

This lets you avoid repeatedly capturing the error statuses.

pkg/vmcp/client/client.go
```diff
@@ -696,8 +696,6 @@ func (h *httpBackendClient) ReadResource(
 	}
 
 	// Extract _meta field from backend response
-	// Note: Due to MCP SDK limitations, the SDK's ReadResourceResult may not include Meta.
-	// This preserves it for future SDK improvements.
```
jerm-dro, line 700: Nit: I don't think this comment needs to be deleted. It explains a gap in functionality that is not immediately obvious in the code.

pkg/vmcp/client/client.go
```diff
@@ -127,8 +128,6 @@ func (a *authRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
 		return nil, fmt.Errorf("authentication failed for backend %s: %w", a.target.WorkloadID, err)
 	}
 
-	logger.Debugf("Applied authentication strategy %q to backend %s", a.authStrategy.Name(), a.target.WorkloadID)
```
jerm-dro, line 130: Nit: can we put this log back?

pkg/vmcp/discovery/manager.go
```diff
@@ -142,24 +150,41 @@ func (m *DefaultManager) Discover(ctx context.Context, backends []vmcp.Backend)
...
-		return nil, fmt.Errorf("%w: %w", ErrDiscoveryFailed, err)
+		return nil, err
 	}
 
-	// Cache the result (skips caching if at capacity and key doesn't exist)
-	m.cacheCapabilities(cacheKey, caps)
-
-	return caps, nil
+	return result.(*aggregator.AggregatedCapabilities), nil
 }
```
jerm-dro: Blocking: Do we have performance numbers motivating this change?  It seems like a bit of premature optimization. In the absence of data, I'd prefer to keep things simple.

pkg/vmcp/health/checker.go
```diff
@@ -62,16 +75,28 @@ func NewHealthChecker(client vmcp.BackendClient, timeout time.Duration, degraded
...
 
 	logger.Debugf("Performing health check for backend %s (%s)", target.WorkloadName, target.BaseURL)
 
+	// Short-circuit health check if targeting ourselves
+	// This prevents the server from trying to health check itself, which would work
+	// but is wasteful and can cause connection issues during startup
+	if h.selfURL != "" && h.isSelfCheck(target.BaseURL) {
+		logger.Debugf("Skipping health check for backend %s - this is the server itself", target.WorkloadName)
+		return vmcp.BackendHealthy, nil
+	}
```
jerm-dro: Blocking: Based on my reading of the code, vMCP doesn't attempt to check itself. The only caller is here: https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/server/server.go#L343 which uses a backend client. This seems like a case of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), right? If so, can we remove these changes.

test/e2e/thv-operator/virtualmcp/helpers.go
```diff
@@ -89,8 +89,9 @@ func checkPodsReady(ctx context.Context, c client.Client, namespace string, labe
 	}
 
 	for _, pod := range podList.Items {
+		// Skip pods that are not running (e.g., Succeeded, Failed from old deployments)
 		if pod.Status.Phase != corev1.PodRunning {
-			return fmt.Errorf("pod %s is in phase %s", pod.Name, pod.Status.Phase)
+			continue
```
jerm-dro, line 97: Nit: I think it's fine to change this function, because we're only ever using it to check for 1 expected pod. However, you should update the comment and the comments on the callers to make the contract change clear. Previously, we waited for all pods to be ready. Now, we're just waiting for some pod.

.gitignore
```diff
@@ -42,4 +42,11 @@ cmd/thv-operator/.task/checksum/crdref-gen
...
-crd-helm-wrapper
\ No newline at end of file
+crd-helm-wrapper
+cmd/vmcp/__debug_bin*
+
+# Demo files
+examples/operator/virtual-mcps/vmcp_optimizer.yaml
+scripts/k8s_vmcp_optimizer_demo.sh
+examples/ingress/mcp-servers-ingress.yaml
+/vmcp
```
jerm-dro: nit: I'd prefer to remove these lines from the .gitignore. These seem to unique to your demo and not something that others are likely to encounter.

therealnb: Yes - they were only temporary, to stop me checking in the actual files.

pkg/vmcp/health/checker.go
therealnb: This was actually happening (hence the code) as part of the race condition. It might not happen now https://github.com/stacklok/toolhive/pull/3450 is fixed. I can try this again.

pkg/vmcp/discovery/manager.go
therealnb: This was a fix for a different race condition from the one you fixed in https://github.com/stacklok/toolhive/pull/3450.

- Multiple concurrent requests arrive simultaneously at startup
- Each request triggers expensive capability aggregation
- Results in duplicate work and "recursive-looking duplicate 'Starting capability aggregation' logs"

This is not an optimisation.

pkg/vmcp/health/checker.go
therealnb: I don't think it does anymore. I'll remove it.

pkg/vmcp/client/client.go
therealnb: Ok

pkg/vmcp/client/client.go
therealnb: ok

pkg/vmcp/aggregator/default_aggregator.go
therealnb: ok - will do this.
Sometimes the tracer worked and sometimes it didn't, so thanks for pointing this out.

.gitignore
therealnb: I will remove them when you approve the rest.

pkg/vmcp/aggregator/default_aggregator.go
therealnb: Ok. This is cool. I can use this in QueryCapabilities too.
Thanks for the suggestion.

test/e2e/thv-operator/virtualmcp/helpers.go
therealnb: Will do.

pkg/vmcp/aggregator/default_aggregator.go
therealnb: done

pkg/vmcp/aggregator/default_aggregator.go
therealnb: done

pkg/vmcp/client/client.go
therealnb: done

pkg/vmcp/client/client.go
therealnb: done

pkg/vmcp/health/checker.go
therealnb: Done

test/e2e/thv-operator/virtualmcp/helpers.go
therealnb: done

.gitignore
therealnb: Done

pkg/vmcp/discovery/manager.go
therealnb: I took this out as a separate PR https://github.com/stacklok/toolhive/pull/3471 so I can view in isolation. It may not be required with the new code. I will test.

jerm-dro (approved):

PR #3440, opened 2026-01-26
Add MCP optimizer implementation for semantic tool discovery

therealnb:
# Add MCP Optimizer Implementation for Semantic Tool Discovery

This PR adds the complete MCP optimizer implementation to vMCP, enabling semantic tool discovery and reducing token usage for LLMs working with large toolsets.

## Overview

The optimizer allows vMCP to expose `optim.find_tool` and `optim.call_tool` operations instead of all backend tools directly. This reduces token usage by allowing LLMs to discover relevant tools on demand via semantic search rather than receiving all tool definitions upfront.

### Core Optimizer Package

**Semantic Tool Search** (`pkg/optimizer/`)
- Vector embeddings (384-dim) for semantic similarity search
- Full-text search via SQLite FTS5 for BM25 text matching
- Hybrid search combining semantic and BM25 results (configurable ratio)
- Multiple embedding backends:
  - Ollama (local HTTP API)
  - OpenAI-compatible (vLLM, OpenAI, etc.)
  - Placeholder (deterministic hash-based, for testing)

**Token Counting** (`pkg/optimizer/tokens/`)
- LLM cost estimation based on token counts
- Supports monitoring token usage and optimization effectiveness

**Database Layer** (`pkg/optimizer/db/`)
- SQLite-based storage with sqlite-vec for vector similarity search
- FTS5 for full-text search
- Hybrid search implementation combining both approaches
- Persistent and in-memory storage options

**Ingestion Service** (`pkg/optimizer/ingestion/`)
- Ingests tools from all backends in the group
- Generates embeddings for tool metadata
- Maintains searchable index of all available tools

### vMCP Integration

**Optimizer Endpoints** (`pkg/vmcp/optimizer/`)
- `optim.find_tool`: Semantic and string-based tool discovery
- `optim.call_tool`: Tool invocation with automatic routing
- Integration with vMCP server lifecycle
- Comprehensive test coverage (unit, integration, semantic search, string matching)

**Server Integration** (`pkg/vmcp/server/`)
- Optimizer initialization and lifecycle management
- Session-based optimizer instances
- Integration with discovery manager and backend registry

**Router Updates** (`pkg/vmcp/router/`)
- Special handling for `optim_*` prefixed tools
- Prevents routing optimizer tools to backends (handled by vMCP itself)

### Kubernetes Operator Support

**Service Resolution** (`cmd/thv-operator/pkg/vmcpconfig/converter.go`)
- Resolves Kubernetes Service names to URLs for embedding services
- Handles `embeddingService` → `embeddingURL` conversion
- Supports in-cluster deployments

**CRD Schema** (`deploy/charts/operator-crds/`)
- Complete optimizer configuration schema
- Supports all optimizer features (embeddings, persistence, hybrid search)
- Documentation updates

### Configuration

**OptimizerConfig** (`pkg/vmcp/config/config.go`)
- Comprehensive configuration options:
  - `enabled`: Enable/disable optimizer
  - `embeddingBackend`: Choose embedding provider
  - `embeddingURL`: Embedding service URL
  - `embeddingModel`: Model name for embeddings
  - `embeddingDimension`: Vector dimension
  - `persistPath`: Optional persistence path
  - `ftsDBPath`: FTS5 database path
  - `hybridSearchRatio`: Semantic vs BM25 mix (0-100%)
  - `embeddingService`: Kubernetes service name (K8s only)

**CLI Integration** (`cmd/vmcp/app/commands.go`)
- Optimizer configuration parsing from YAML
- Runtime configuration and initialization
- Logging and status reporting

### Build System

**Build Tags** (`Taskfile.yml`)
- Added `-tags="fts5"` build flag for SQLite FTS5 support
- Required for optimizer functionality
- Applied to all vmcp builds (build, install)

**Test Task** (`Taskfile.yml`)
- Added `test-optimizer` task for optimizer integration tests
- Uses sqlite-vec for vector search testing

### Examples & Scripts

**Example Configuration** (`examples/vmcp-config-optimizer.yaml`)
- Complete example showing optimizer configuration
- Demonstrates all configuration options

**Helper Scripts** (`scripts/`)
- `test-optimizer-with-sqlite-vec.sh`: Integration testing
- `inspect-optimizer-db.sh`: Database inspection
- `query-optimizer-db.sh`: Query testing
- Various chromem inspection tools

## Documentation

- Optimizer package documentation in `pkg/optimizer/README.md`
- Integration guide in `pkg/optimizer/INTEGRATION.md`
- CRD API documentation updates
- Example configurations

## Testing

- Comprehensive unit tests for all optimizer components
- Integration tests for optimizer endpoints
- Semantic search test suite
- String matching test suite
- E2E tests for Kubernetes deployments

## Dependencies

- `chromem-go`: Vector database for embeddings
- `sqlite-vec`: SQLite extension for vector similarity search
- `go.uber.org/mock`: Mock generation for tests

## Build Requirements

- Requires `-tags="fts5"` build flag for FTS5 support
- SQLite with FTS5 extension
- sqlite-vec extension for vector search

## Related

- Depends on infrastructure improvements from optimizer-enablers PR
- Split from #3373
- Implements semantic tool discovery as described in optimizer documentation

## Large PR Justification
- This is the second part of a two part PR.

### DIFF
--- a/pkg/telemetry/config.go
+++ b/pkg/telemetry/config.go
@@ -196,9 +196,16 @@ func NewProvider(ctx context.Context, config Config) (*Provider, error) {
 		return nil, err
 	}
 
+	// Apply default for ServiceVersion if not provided
+	// Documentation states: "When omitted, defaults to the ToolHive version"
+	serviceVersion := config.ServiceVersion
+	if serviceVersion == "" {
+		serviceVersion = versions.GetVersionInfo().Version
+	}
+
 	telemetryOptions := []providers.ProviderOption{
 		providers.WithServiceName(config.ServiceName),
-		providers.WithServiceVersion(config.ServiceVersion),
+		providers.WithServiceVersion(serviceVersion),
 		providers.WithOTLPEndpoint(config.Endpoint),
 		providers.WithHeaders(config.Headers),
 		providers.WithInsecure(config.Insecure),


deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_virtualmcpservers.yaml
```diff
@@ -677,17 +677,74 @@ spec:
...
-                          EmbeddingService is the name of a Kubernetes Service that provides the embedding service
-                          for semantic tool discovery. The service must implement the optimizer embedding API.
+                          EmbeddingBackend specifies the embedding provider: "ollama", "openai-compatible", or "placeholder".
+                          - "ollama": Uses local Ollama HTTP API for embeddings
+                          - "openai-compatible": Uses OpenAI-compatible API (vLLM, OpenAI, etc.)
+                          - "placeholder": Uses deterministic hash-based embeddings (for testing/development)
+                        enum:
+                        - ollama
+                        - openai-compatible
+                        - placeholder
```
ptelang: This enum should be modified to match BackendType values permitted in the Config struct in: toolhive/cmd/thv-operator/pkg/optimizer/embeddings/manager.go: 

ollama, vllm, unified, openai

pkg/vmcp/optimizer/internal/embeddings/manager.go
```diff
@@ -0,0 +1,219 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
jerm-dro, line 1: Blocking: Can we consolidate the optimizer code into one package?

I suggest:
- `pkg/vmcp/optimizer/` - contains only the bits that are necessary for integration.
- `pkg/vmcp/optimizer/internal/` contains most of other package here in subdirectories (e.g. embeddings, db, ingestion, db, models, tokens). These are implementation details of the optimizer and don't need to be available outside the optimizer package.

I find intentional package design helps in a few ways:
- motivates clear, well-defined APIs
- encapsulation of implementation details keeps code flexible
- reduces the cognitive load for people working within and outside the directory

pkg/vmcp/optimizer/internal/db/backend_server.go
```diff
@@ -0,0 +1,243 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
jerm-dro, line 1: Blocker: This db package has quite a few public structs, but no interfaces.

Can we refactor this so the db package has one public interface which has methods for reading and writing data? Then, I'd expect one public function that accepts a config (and the embedding function) and returns something that implements the database interface. Only exposing concrete structs and no interfaces means that everything that depends on this package has to rely on the real implementations, which undermines testability.

The public interface should be well tested in a behavioral way (e.g., write some data, read it back through the public API). Testing and factoring like this will allow us to encapsulate the complexity of this package and document/validate its behavior.

Also, I noticed some of the public methods here aren't used at all `BackendServerOps.Get`, `BackendServerOps.Search`, etc. It'd be good to remove these functions which aren't used at all.

pkg/vmcp/optimizer/internal/ingestion/service.go
```diff
@@ -0,0 +1,346 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
jerm-dro, line 1: Suggestion: This is a pretty small package. Once as the `Db` package has a more well defined API, it may make sense to fold this package into that.

pkg/vmcp/optimizer/config.go
```diff
@@ -0,0 +1,42 @@
...
+	if cfg.EmbeddingBackend != "" || cfg.EmbeddingURL != "" || cfg.EmbeddingModel != "" || cfg.EmbeddingDimension > 0 {
+		optimizerCfg.EmbeddingConfig = &embeddings.Config{
+			BackendType: cfg.EmbeddingBackend,
+			BaseURL:     cfg.EmbeddingURL,
+			Model:       cfg.EmbeddingModel,
+			Dimension:   cfg.EmbeddingDimension,
+		}
+	}
+
+	return optimizerCfg
```
jerm-dro: Blocker: We shouldn't introduce new internal config types that result in these translation layers. When changes are made to the internal config, it must also be made to actual config and translation layer. It's easy to miss plumbing layers like this and that results in bugs. 

Let's just rely on the actual server config types for the optimizer implementation.

pkg/vmcp/optimizer/integration.go
```diff
@@ -0,0 +1,42 @@
...
+	// HandleSessionRegistration handles session registration for optimizer mode.
+	// Returns true if optimizer mode is enabled and handled the registration,
+	// false if optimizer is disabled and normal registration should proceed.
+	// The resourceConverter function converts vmcp.Resource to server.ServerResource.
+	HandleSessionRegistration(
+		ctx context.Context,
+		sessionID string,
+		caps *aggregator.AggregatedCapabilities,
+		mcpServer *server.MCPServer,
+		resourceConverter func([]vmcp.Resource) []server.ServerResource,
```
jerm-dro: I'd prefer to leave the Session startup code within the server, at least for this PR.

The implementation of this `Integration` gizmo is very coupled to whatever else happens during session startup, so I think it makes if the registration responsibility remains in the server.

pkg/vmcp/optimizer/optimizer.go
```diff
@@ -1,91 +1,927 @@
...
+}
+
+// Initialize performs all optimizer initialization:
+//   - Registers optimizer tools globally with the MCP server
+//   - Ingests initial backends from the registry
 //
-// Implementations may use various strategies for tool matching:
-//   - DummyOptimizer: Exact string matching (for testing)
-//   - EmbeddingOptimizer: Semantic similarity via embeddings (production)
-type Optimizer interface {
```
jerm-dro, line 27: Why was this optimizer interface removed? 

I was expecting that your code would provide an implementation of this. Then, you would not need to create the `Integration` object. All the ingestion logic would be handled at startup time by the `OptimizerFactory func([]server.ServerTool) optimizer.Optimizer` in server.go, which was also deleted.

pkg/vmcp/optimizer/internal/embeddings/ollama_test.go
```diff
@@ -0,0 +1,69 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
jerm-dro, line 1: The OllamaBackend and OpenAICompatibleBackend both implement the Backend interface but have separate, non-shared tests.

Could we:

1. Add an httptest mock server for the Ollama API (similar to what exists for OpenAI)
2. Create shared table-driven tests that exercise both backends through the Backend interface

This would ensure both implementations have proper unit test coverage and behave consistently.

pkg/vmcp/server/adapter/capability_adapter.go
```diff
@@ -4,6 +4,7 @@
 package adapter
 
 import (
+	"context"
```
jerm-dro: Question, blocking: Going back to [this](https://github.com/stacklok/toolhive/pull/3440/changes#r2729944460), I'm surprised to see changes to this file. If we plugged into the pre-existing integration are are these changes necessary?

pkg/vmcp/optimizer/optimizer.go
therealnb: An interface here seemed like overkill, but I get the advantages of interfaces in general and forgot that in go you pretty much need them to do mocks (I've been doing python for years).

I'll put this back.

deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_virtualmcpservers.yaml
therealnb: ok. Will do.

therealnb: Done

pkg/vmcp/optimizer/internal/ingestion/service.go
therealnb: It is small, but has distinct responsibilities:

- Embedding generation and management
- Token counting
- Telemetry tracking (embedding time)
- Event-driven ingestion orchestration

If you insist, I can fold it in, but I'm not convinced.

pkg/vmcp/optimizer/internal/db/backend_server.go
therealnb: I had a stab at this - please take a look.

pkg/vmcp/optimizer/internal/embeddings/manager.go
therealnb: Makes sense. Will attempt this.

pkg/vmcp/optimizer/optimizer.go
therealnb: The interface and factory are back. Please review.

pkg/vmcp/optimizer/internal/embeddings/manager.go
therealnb: Should be done. Check it out.

pkg/vmcp/server/adapter/capability_adapter.go
therealnb: We need to change this a bit because this interface is how optimizer communicates back. The definition can probably go in pkg/vmcp/server/adapter/handler_factory.go.

I'll move that

pkg/vmcp/optimizer/config.go
therealnb: Ok, this was in the name of encapsulation, but I get your point. I will change this.

therealnb: Ok, another big change, but this is done now.

examples/vmcp-config-optimizer.yaml
```diff
@@ -0,0 +1,126 @@
...
+# discover relevant tools on demand via semantic search.
+#
+# The optimizer ingests tools from all backends in the group, generates
+# embeddings, and provides semantic search capabilities.
+
+optimizer:
+  # Enable the optimizer
+  enabled: true
+  
+  # Embedding backend: "ollama" (default), "openai-compatible", or "vllm"
```
ptelang: Update to match match BackendType values permitted in the Config struct in: toolhive/cmd/thv-operator/pkg/optimizer/embeddings/manager.go:

ollama, vllm, unified, openai

therealnb:
Will need to check https://github.com/stacklok/toolhive/pull/3471 too.

examples/vmcp-config-optimizer.yaml
therealnb: This file shouldn't have been checked in. Will remove it.

therealnb: removed

therealnb:
Retested. We definitely need https://github.com/stacklok/toolhive/pull/3471 to stop a race condidtion.

therealnb:
I've split PR #3440 into three smaller PRs:

## Created PRs

| PR | Branch | Title | URL |
|----|--------|-------|-----|
| **PR 1** | `optimizer-pr1-packages` | Add optimizer packages and infrastructure (non-functional) | https://github.com/stacklok/toolhive/pull/3516 |
| **PR 2** | `optimizer-pr2-integration` | Wire optimizer into server startup and CLI | https://github.com/stacklok/toolhive/pull/3517 |
| **PR 3** | `optimizer-pr3-docs` | Add optimizer documentation and README updates | https://github.com/stacklok/toolhive/pull/3518 |

## What's in Each PR

**PR 1 (Non-functional):** ~10,400 lines
- All new optimizer internal packages (`db/`, `embeddings/`, `ingestion/`, `models/`, `tokens/`)
- `EmbeddingOptimizer` implementation and updated `Optimizer` interface
- Updated `dummy_optimizer.go` for backward compatibility
- Dependencies (`go.mod`/`go.sum`)
- Config schema and CRD updates
- `OptimizerHandlerProvider` interface in adapter

**PR 2 (Integration):** ~630 lines added, ~770 removed
- Server wiring (`server.go`, `capability_adapter.go`)
- CLI integration (`commands.go`)
- Operator converter changes
- `Taskfile.yml` build flags
- Deletes `dummy_optimizer.go`
- Test updates for integration

**PR 3 (Documentation):** ~385 lines
- README updates (`cmd/vmcp/README.md`)
- New docs (`pkg/vmcp/optimizer/README.md`, `REFACTORING.md`)
- CRD API docs and Helm chart README updates

## Merge Order

1. Merge PR 1 first (standalone, no dependencies)
2. Merge PR 2 after PR 1 (depends on PR 1)
3. Merge PR 3 after PR 2 (depends on PR 1 & 2)

You will also need https://github.com/stacklok/toolhive/pull/3471 to handle a race condition if this has not been addressed elsewhere.


**Note:** There's a known merge conflict when combining PR1 and PR2 - PR1 modifies `dummy_optimizer.go` while PR2 deletes it. The resolution is to accept the deletion.

PR #3449, opened 2026-01-26
Implement tests for missing workload API endpoints

dmjb:
Bulk endpoints, and operations on named workloads.

### DIFF
--- a/test/e2e/api_workload_lifecycle_test.go
+++ b/test/e2e/api_workload_lifecycle_test.go
@@ -847,6 +847,12 @@ var _ = Describe("Workload Lifecycle API", Label("api", "workloads", "lifecycle"
 			}
 		})
 
+		AfterEach(func() {
+			for _, name := range workloadNames {
+				deleteWorkload(apiServer, name)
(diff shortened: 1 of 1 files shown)


test/e2e/api_workload_lifecycle_test.go
```diff
@@ -382,9 +383,549 @@ var _ = Describe("Workload Lifecycle API", Label("api", "workloads", "lifecycle"
...
+				resp := bulkRestartWorkloads(apiServer, bulkReq)
+				defer resp.Body.Close()
+
+				By("Verifying response status is 400 Bad Request")
+				Expect(resp.StatusCode).To(Equal(http.StatusBadRequest))
+			})
+		})
+	})
+
+	Describe("POST /api/v1beta/workloads/delete - Bulk delete workloads", func() {
```
jhrozek, line 840: I know this is a delete test, but would it make sense to add an AfterEach to delete the created workloads to not leave stuff hanging around if the test fails? All the other tests have an AfterEach.

Issue #3149, opened 2025-12-23
Implement K8SReporter for Kubernetes status updates


4t8dd:
Implement K8SReporter to update VirtualMCPServer.Status from the vMCP runtime.

Four step to implement:
1. K8SReporter Implementation
2. RBAC Permissions. Update vMCP runtime service account to edit VMCP CRD status.
3. Server Integration:
    1. update server config
    2. start/stop along with server.
4. update runtime main to initialize:
      Flow:
        1 . Check VMCP_NAME and VMCP_NAMESPACE env vars (set by operator in deployment)
        2. If present → Kubernetes mode → create K8SReporter
        3. If absent → CLI mode → create NoOpReporter
        4. Pass reporter to Server config
        5. Server lifecycle manages reporter (Start/Stop)

4. Test coverage
   we need Unit Test and a new e2e test case for this.
  

Refer to #2854 for more details  and depends #3148 #3147

PR #3457, opened 2026-01-27
Integrate status reporting into vMCP runtime

yrobla:
Wire up periodic status reporting to enable vMCP runtime to report operational status to the control plane (Kubernetes API or CLI logging).

Changes:
  - Add BuildStatus() method to health.Monitor that converts health state to vmcp.Status format with phase determination, backend details, and Kubernetes-style conditions
  - Create periodicStatusReporting() in server package that runs in background goroutine, pulling health data every 30 seconds and sending to configured reporter
  - Integrate status reporting lifecycle into server.Start() and Stop() with proper shutdown handling
  - Update commands.go to use factory pattern (vmcpstatus.NewReporter) instead of hardcoded LoggingReporter, enabling auto-detection of K8s vs CLI mode via environment variables
  - Add comprehensive tests for status building logic (phase determination, condition generation, message formatting)

Phase determination logic:
  - Ready: All backends healthy (or no backends yet - cold start)
  - Degraded: Some backends healthy, some degraded/unhealthy
  - Failed: No healthy backends
  - Pending: Waiting for backends to be discovered

The implementation respects the separation between health monitoring (data gathering) and status reporting (persistence), with K8sReporter handling Kubernetes-specific fields like ObservedGeneration.

Related-to: #3149

## Large PR Justification

This is an atomic PR, covering the integration of status into vmcp. It includes complete testing.
It cannot be split as it's an isolated functionality.

### DIFF
--- a/pkg/vmcp/health/monitor.go
+++ b/pkg/vmcp/health/monitor.go
@@ -299,7 +299,12 @@ func (m *Monitor) IsBackendHealthy(backendID string) bool {
 // Returns counts of healthy, degraded, unhealthy, and total backends.
 func (m *Monitor) GetHealthSummary() Summary {
 	allStates := m.statusTracker.GetAllStates()
+	return computeSummary(allStates)
+}
 
+// computeSummary computes a Summary from a snapshot of backend states.
+// This is a pure function that takes a states map and returns aggregated counts.
+func computeSummary(allStates map[string]*State) Summary {
 	summary := Summary{
 		Total:           len(allStates),
 		Healthy:         0,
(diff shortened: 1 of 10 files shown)


pkg/vmcp/health/monitor.go
```diff
@@ -340,3 +347,209 @@ func (s Summary) String() string {
...
+	// Use the actual AuthConfigRef populated during backend discovery.
+	// In K8s mode, this is the name of the MCPExternalAuthConfig resource.
+	// In CLI mode or when not discovered via K8s, this may be empty.
+	return backend.AuthConfigRef, backend.AuthConfig.Type
+}
+
+// formatBackendMessage creates a human-readable message for a backend's health state.
+func formatBackendMessage(state *State) string {
+	if state.LastError != nil {
+		return fmt.Sprintf("%s (failures: %d)", state.LastError.Error(), state.ConsecutiveFailures)
```
jhrozek, line 479: maybe we could log `LastError.Error()` and only return a more generic failure (maybe have a couplf of categories we could choose from using errors.Is)

pkg/vmcp/server/server.go
```diff
@@ -592,6 +612,11 @@ func (s *Server) Stop(ctx context.Context) error {
 		}
 	}
 
+	// Cancel status reporting goroutine if running
+	if s.statusReportingCancel != nil {
+		s.statusReportingCancel()
+	}
+
 	// Run shutdown functions (e.g., status reporter, future components)
 	for _, shutdown := range s.shutdownFuncs {
```
jerm-dro, line 621: Nit: could you have appended the `statusReportingCancel` to the `s.shutdownFuncs` below?

Then, we wouldn't need to keep track of the `statusReportingCtx` and `statusReportingCancel`.

pkg/vmcp/status/k8s_reporter.go
```diff
@@ -145,8 +148,28 @@ func (*K8sReporter) updateStatus(vmcpServer *mcpv1alpha1.VirtualMCPServer, statu
...
+	// when the condition Status hasn't changed. This is important for Kubernetes-style
+	// condition semantics - LastTransitionTime should only update on Status transitions.
+	//
+	// First, identify which condition types are present in the new status
+	newConditionTypes := make(map[string]bool)
+	for _, cond := range status.Conditions {
+		newConditionTypes[cond.Type] = true
+	}
+
+	// Remove known condition types that are no longer present (e.g., Degraded after recovery)
```
jerm-dro, line 161: Non-blocking, question: why do you need to remove known conditions? How can you assume they are no longer present?

yrobla:
ok implementing in follow ups

PR #3458, opened 2026-01-27
Add headerForward to MCPRemoteProxy CRD

jhrozek:
## Summary

Add support for forwarding custom headers to remote MCP servers via MCPRemoteProxy. Headers can be specified as plaintext values or resolved from Kubernetes Secrets at runtime.

### New CRD Fields

```yaml
spec:
  headerForward:
    addPlaintextHeaders:     # Static headers (non-sensitive)
      X-Tenant-ID: "tenant-123"
    addHeadersFromSecret:    # Headers from K8s Secrets
      - headerName: "X-API-Key"
        valueSecretRef:
          name: api-key-secret
          key: api-key
```

### 1. Allow TOOLHIVE_SECRETS_PROVIDER env var to bypass SetupCompleted check

Previously, `GetProviderType()` checked `SetupCompleted` before checking the environment variable, causing Kubernetes deployments to fail when using environment-injected secrets.

Reorders the checks so the environment variable is checked first, allowing Kubernetes deployments to specify `TOOLHIVE_SECRETS_PROVIDER=environment` without requiring local secrets setup.

### 2. Add headerForward to MCPRemoteProxy CRD

**Implementation:**
- Operator creates env vars with SecretKeyRef for secret-backed headers
- Operator sets `TOOLHIVE_SECRETS_PROVIDER=environment` to enable the EnvironmentProvider in the runner
- RunConfig stores secret identifiers (not values)
- Runner resolves secrets via EnvironmentProvider at startup
- Header forward middleware adds headers to outgoing requests

**Security:**
- Secret values never stored in ConfigMaps (only identifiers)
- `kubectl describe` shows `<set to key 'x' in secret 'y'>`
- `resolvedHeaders` field is not serialized to disk

### Secret-backed headers only

```yaml
kind: MCPRemoteProxy
metadata:
  name: secret-headers-proxy
spec:
  remoteURL: http://mcp-backend.default.svc.cluster.local:8080/mcp
  transport: streamable-http
  oidcConfig:
    type: kubernetes
    kubernetes:
      issuer: https://kubernetes.default.svc.cluster.local
      audience: https://kubernetes.default.svc.cluster.local
  headerForward:
    addHeadersFromSecret:
      - headerName: "X-API-Key"
        valueSecretRef:
          name: api-key-secret
          key: api-key
      - headerName: "Authorization"
        valueSecretRef:
          name: auth-token-secret
          key: token
```

### Mixed plaintext and secret headers

```yaml
kind: MCPRemoteProxy
metadata:
  name: mixed-headers-proxy
spec:
  remoteURL: http://mcp-backend.default.svc.cluster.local:8080/mcp
  transport: streamable-http
  oidcConfig:
    type: kubernetes
    kubernetes:
      issuer: https://kubernetes.default.svc.cluster.local
      audience: https://kubernetes.default.svc.cluster.local
  headerForward:
    addPlaintextHeaders:
      X-Tenant-ID: "tenant-456"
      X-Request-Source: "toolhive-proxy"
    addHeadersFromSecret:
      - headerName: "X-API-Key"
        valueSecretRef:
          name: api-key-secret
          key: api-key
```

## Test plan

Tested in kind cluster using yardstick MCP server (patched to log incoming headers):

- [x] Plaintext headers only
- [x] Secret-backed headers only  
- [x] Mixed plaintext and secret headers
- [x] Multiple secrets, multiple keys from same secret
- [x] Missing secret (graceful failure with clear error)

Verified headers appear in yardstick logs:
```
Header: X-Api-Key: sk-test-api-key-12345
Header: Authorization: Bearer super-secret-token-67890
Header: X-Tenant-Id: tenant-456
```


cmd/thv-operator/api/v1alpha1/mcpremoteproxy_types.go
```diff
@@ -7,6 +7,30 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// HeaderForwardConfig defines header forward configuration for remote servers.
+type HeaderForwardConfig struct {
+	// AddPlaintextHeaders is a map of header names to literal values to inject into requests.
+	// WARNING: Values are stored in plaintext. Use addHeadersFromSecret for sensitive data.
```
ChrisJBurns: We may want to just add a sentence about the headers being shown in the CR manifest on a `kubectl describe` just so its really clear to users.

cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go
```diff
@@ -224,3 +227,34 @@ func labelsForRunConfigRemoteProxy(proxyName string) map[string]string {
...
+		headerSecrets := make(map[string]string, len(proxy.Spec.HeaderForward.AddHeadersFromSecret))
+		for _, headerSecret := range proxy.Spec.HeaderForward.AddHeadersFromSecret {
+			if headerSecret.ValueSecretRef == nil {
+				continue
+			}
+			// Get the secret identifier (not the full env var name)
+			_, secretIdentifier := ctrlutil.GenerateHeaderForwardSecretEnvVarName(proxy.Name, headerSecret.HeaderName)
+			headerSecrets[headerSecret.HeaderName] = secretIdentifier
+		}
+		*options = append(*options, runner.WithHeaderForwardSecrets(headerSecrets))
```
ChrisJBurns, line 258: Is it possible that there could be two env vars with the same name? This would likely be user error, but not sure if it is an easy mistake that can happen via `addPlaintextHeaders` and `addHeadersFromSecret` ?

cmd/thv-operator/api/v1alpha1/mcpremoteproxy_types.go
```diff
@@ -7,6 +7,30 @@ import (
...
+	// AddHeadersFromSecret references Kubernetes Secrets for sensitive header values.
+	// +optional
+	AddHeadersFromSecret []HeaderFromSecret `json:"addHeadersFromSecret,omitempty"`
+}
+
+// HeaderFromSecret defines a header whose value comes from a Kubernetes Secret.
+type HeaderFromSecret struct {
+	// HeaderName is the HTTP header name (e.g., "X-API-Key")
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
ChrisJBurns, line 27: nit: Probably not, but should we add a max length or do you think that would cause more issues than it avoids?

ChrisJBurns (approved):
Looks good to me, a couple of comments, but are more potential for considerations than non blocking.

cmd/thv-operator/api/v1alpha1/mcpremoteproxy_types.go
jhrozek: added, good suggestion

cmd/thv-operator/api/v1alpha1/mcpremoteproxy_types.go
jhrozek: I don't think there is a limit to the size of http headers..on the other side, having a pragmatic and generous limit like 255 chars is probably a good choice.

cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go
jhrozek: right now secrets take precedence. This was a conscious decision - if someone mistypes and put the same header, we should treat it as a secret. I just tested with:
```yaml
# MCPRemoteProxy with duplicate header names in plaintext and secret
# Purpose: Test behavior when same header is defined in both places
# Expected: Secret value should override plaintext value (secrets-win)
# Prerequisites: Deploy 00-namespace.yaml, 01-secrets.yaml, 02-yardstick-backend.yaml first
---
kind: MCPRemoteProxy
metadata:
  name: duplicate-headers-proxy
  namespace: header-forward-test
spec:
  # Point to yardstick backend (must be deployed first)
  remoteURL: http://mcp-yardstick-proxy.header-forward-test.svc.cluster.local:8080/mcp
  transport: streamable-http

  # Kubernetes OIDC for service account token validation
  oidcConfig:
    type: kubernetes
    kubernetes:
      issuer: https://kubernetes.default.svc.cluster.local
      audience: https://kubernetes.default.svc.cluster.local

  headerForward:
    # Plaintext value for X-API-Key - should be OVERRIDDEN by secret
    addPlaintextHeaders:
      X-API-Key: "PLAINTEXT-VALUE-SHOULD-NOT-APPEAR"
      X-Unique-Plaintext: "this-should-appear"

    # Secret value for X-API-Key - should WIN over plaintext
    addHeadersFromSecret:
      - headerName: "X-API-Key"
        valueSecretRef:
          name: api-key-secret
          key: api-key
      - headerName: "X-Unique-Secret"
        valueSecretRef:
          name: auth-token-secret
          key: token
```

PR #3459, opened 2026-01-27
E2E test case for discovery API

dmjb:


test/e2e/api_discovery_test.go
```diff
@@ -0,0 +1,346 @@
...
+type clientStatusResponse struct {
+	Clients []client.MCPClientStatus `json:"clients"`
+}
+
+// -----------------------------------------------------------------------------
+// Helper functions
+// -----------------------------------------------------------------------------
+
+func discoverClients(server *e2e.Server) *http.Response {
+	resp, err := http.Get(server.BaseURL() + "/api/v1beta/discovery/clients")
```
jhrozek: should this use apiserver.Get? Just checking other tests it seems like that's what the helpers use..or is there a point is using raw http.Get?

test/e2e/api_discovery_test.go
```diff
@@ -0,0 +1,346 @@
...
+				successCount := 0
+				for i := 0; i < numRequests; i++ {
+					if <-done {
+						successCount++
+					}
+				}
+
+				// All requests should succeed
+				Expect(successCount).To(Equal(numRequests))
+				Expect(errors).To(BeEmpty())
```
jhrozek: does this also report what the errors are or would we have to collect them manually?

PR #3463, opened 2026-01-27
Improve registry API behaviour for unreachable registries

dmjb:
Previously a registry could be added without validating that it exists. When calling a registry API which referred to a non existent registry, it would time out for 10 seconds and return a 500.

Change the behaviour to:

1. Validate that the registry is contactable and matches the contract we expect.
2. Reduce timeout and return 504 when using an unreachable registry.


dmjb:
superseded

PR #3466, opened 2026-01-27
Add configurable UserInfo field mapping and UserInfo support for authserver's OAuth provider

jhrozek:
Enhance UserInfoFieldMapping to support ordered fallback field chains, enabling compatibility with non-standard OAuth providers like GitHub that use different claim names (e.g., "id" instead of "sub").

Changes:
- Replace single field names with ordered lists (SubjectFields, NameFields, EmailFields)
- Add ResolveField() helper with support for numeric ID conversion
- Add ResolveSubject/ResolveName/ResolveEmail methods with defaults
- Implement UserInfoProvider interface on BaseOAuth2Provider
- Support additional headers for provider-specific requirements (e.g., GitHub API versioning)
- Add JSON/YAML struct tags for file-based configuration support

Example YAML configuration for GitHub:
```
upstream:
  client_id: "my-app"
  client_secret: "secret"
  redirect_uri: "http://localhost:8080/callback"
  authorization_endpoint: "https://github.com/login/oauth/authorize"
  token_endpoint: "https://github.com/login/oauth/access_token"
  scopes:
    - user:email
  userinfo:
    endpoint_url: "https://api.github.com/user"
    additional_headers:
      X-GitHub-Api-Version: "2022-11-28"
    field_mapping:
      subject_fields: ["id", "login"]
      name_fields: ["name", "login"]
      email_fields: ["email"]
```

### DIFF
--- a/pkg/authserver/upstream/userinfo_config.go
+++ b/pkg/authserver/upstream/userinfo_config.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"strconv"
 
 	"github.com/stacklok/toolhive/pkg/networking"
 )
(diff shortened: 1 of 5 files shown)


pkg/authserver/upstream/userinfo_config.go
```diff
@@ -19,75 +19,140 @@ import (
...
+	return DefaultNameFields
+}
 
-// GetSubjectField returns the configured subject field or the default.
-func (m *UserInfoFieldMapping) GetSubjectField() string {
-	if m != nil && m.SubjectField != "" {
-		return m.SubjectField
+// GetEmailFields returns the configured email fields or the default.
+func (m *UserInfoFieldMapping) GetEmailFields() []string {
+	if m != nil && len(m.EmailFields) > 0 {
```
tgrunnagle, line 87: Do we want to allow an empty list for email and name fields (in case the field doesn't exist for the Idp for whatever reason?)

jhrozek: This method just returns the right field name (e.g. typically "email"). Then the method ResolveField iterates through the claims and returns the matching one or `""` as the default. 

The only field we require is `sub` (mapped or verbatim)

jhrozek:
I hope it's OK to merge with the explanation to your question @tgrunnagle - if not, just shout and I'll do a follow-up.

PR #3468, opened 2026-01-27
Simplify controller and add E2E tests for status reporting

yrobla:
Simplify VirtualMCPServer controller by delegating to runtime status reporter              
                                                                                             
Remove ~1200 lines by delegating backend discovery and health monitoring                   
from controller to vMCP runtime StatusReporter.                                            
                                                                                             
Key changes:                                                                               
  - Remove discoverBackends() from controller reconciliation (~130 lines)                    
  - Fix RBAC permissions to work in all modes (discovered and inline)                        
  - Remove obsolete controller-side discovery tests (807 lines)                              
  - Add comprehensive E2E status reporting tests (408 lines)                                 
  - Improve health monitoring with dynamic backend management                                
  - Remove duplicate slow test from yardstick suite (saves 133s per run)                     
                                                                                             
Controller now focuses on infrastructure (RBAC, Deployment, Service).                      
Runtime handles discovery, health monitoring, and status updates.  
  
Related-to: #3149

## Large PR Justification

This is a cleanup pr, that removes substantial lines of code to replace with newer ones. It also includes tests to validate the new behaviour.

### DIFF
--- a/pkg/vmcp/server/status_reporting.go
+++ b/pkg/vmcp/server/status_reporting.go
@@ -56,13 +56,12 @@ func (s *Server) periodicStatusReporting(ctx context.Context, config StatusRepor
 	// This ensures that the first status report has accurate health information
 	// rather than reporting with backendCount=0 before checks complete
 	s.healthMonitorMu.RLock()
-	if s.healthMonitor != nil {
+	healthMon := s.healthMonitor
+	s.healthMonitorMu.RUnlock()
+	if healthMon != nil {
 		logger.Debug("Waiting for initial health checks to complete before first status report")
-		s.healthMonitorMu.RUnlock()
-		s.healthMonitor.WaitForInitialHealthChecks()
+		healthMon.WaitForInitialHealthChecks()
 		logger.Debug("Initial health checks complete, proceeding with status reporting")
-	} else {
-		s.healthMonitorMu.RUnlock()
 	}
 
 	ticker := time.NewTicker(interval)
@@ -85,6 +84,18 @@ func (s *Server) periodicStatusReporting(ctx context.Context, config StatusRepor
 
 // reportStatus collects current runtime status and sends it to the reporter.
 func (s *Server) reportStatus(ctx context.Context, reporter vmcpstatus.Reporter) {
+	// Update health monitor with current backends from registry (for dynamic discovery)
+	if dynamicReg, ok := s.backendRegistry.(vmcp.DynamicRegistry); ok {
+		currentBackends := dynamicReg.List(ctx)
+		logger.Debugf("Refreshing backends from registry: %d backends found", len(currentBackends))
+		s.healthMonitorMu.RLock()
+		healthMon := s.healthMonitor
+		s.healthMonitorMu.RUnlock()
+		if healthMon != nil {
+			healthMon.UpdateBackends(currentBackends)
+		}
+	}
+
 	// Build status from health monitor if available
 	var status *vmcp.Status
 
--- a/cmd/vmcp/app/commands.go
+++ b/cmd/vmcp/app/commands.go
@@ -188,6 +188,17 @@ func getVersion() string {
 	return "dev"
 }
 
+// getStatusReportingInterval extracts the status reporting interval from config.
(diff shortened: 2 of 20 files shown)


pkg/vmcp/server/status_reporting.go
```diff
@@ -52,10 +52,23 @@ func (s *Server) periodicStatusReporting(ctx context.Context, config StatusRepor
...
 	logger.Infof("Starting periodic status reporting (interval: %v)", interval)
 
+	// Wait for initial health checks to complete before first status report
+	// This ensures that the first status report has accurate health information
+	// rather than reporting with backendCount=0 before checks complete
+	s.healthMonitorMu.RLock()
+	if s.healthMonitor != nil {
+		logger.Debug("Waiting for initial health checks to complete before first status report")
+		s.healthMonitorMu.RUnlock()
+		s.healthMonitor.WaitForInitialHealthChecks()
```
jhrozek: nit: There's a small race window here - we release the lock on line 61, then call the method on line 62. If something set `healthMonitor` to nil in between, we'd panic.

In practice this is fine because of lifecycle constraints, but for consistency with the pattern used in `reportStatus()` below (lines 91-102), we could capture the reference under the lock:

```suggestion
	s.healthMonitorMu.RLock()
	healthMon := s.healthMonitor
	s.healthMonitorMu.RUnlock()

	if healthMon != nil {
		logger.Debug("Waiting for initial health checks to complete before first status report")
		healthMon.WaitForInitialHealthChecks()
		logger.Debug("Initial health checks complete, proceeding with status reporting")
	}
```

Not urgent - happy to address in a follow-up if you prefer.

cmd/thv-operator/controllers/virtualmcpserver_deployment.go
```diff
@@ -65,6 +67,11 @@ var vmcpRBACRules = []rbacv1.PolicyRule{
 		Resources: []string{"mcpgroups", "mcpservers", "mcpremoteproxies", "mcpexternalauthconfigs", "mcptoolconfigs"},
 		Verbs:     []string{"get", "list", "watch"},
```
jhrozek, line 68: Good call adding this TODO! Just to confirm - this is intentional that inline mode gets the full RBAC permissions (including secrets access) even though it only needs status reporting permissions, right?

The simplicity argument makes sense. If this ever becomes a concern for security-conscious users, they can always provide their own ServiceAccount.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -1560,138 +1553,6 @@ func convertBackendsToStaticBackends(
 // a list of DiscoveredBackend objects with their current status.
 //
 //nolint:gocyclo
```
jhrozek, line 1543: the gocyclo directive could go away if you end up working some more on the PR..or in a next PR

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -499,32 +492,28 @@ func (r *VirtualMCPServerReconciler) ensureAllResources(
...
 	// If a service account is specified, we don't need to create one
 	if vmcp.Spec.ServiceAccount != nil {
 		return nil
 	}
 
-	// Dynamic mode (discovered): Ensure RBAC resources exist
+	// Always ensure RBAC resources exist for status reporting
+	// Status reporting (via K8sReporter) runs in ALL modes (discovered and inline),
+	// so the vMCP runtime needs permissions to update VirtualMCPServer status
+	// regardless of the outgoing auth source mode.
```
jhrozek, line 516: Isn't the problem with this that in the inline mode we'd still assign all the RBAC privileges including to read secrets?

jhrozek: we don't have to do it in this PR, but one of the points of the inline mode was that the vMCP binary will not get more privileges that it needs through RBAC - vMCP is already exposed to the outside world, so giving it powers to call the k8s API might make deployers nervous

jhrozek (approved):
I think tightening the RBAC should be a follow-up

PR #3472, opened 2026-01-27
Add authserver configuration with validation and auto-generation

jhrozek:
This PR introduces the `Config` type for the OAuth authorization server, providing a pure configuration interface with full validation. The config supports signing keys via a `KeyProvider` interface (with auto-generation for development), HMAC secrets for opaque tokens (authorization codes and refresh tokens), sensible token lifespan defaults (1h access, 7d refresh, 10m auth code), and pre-registered OAuth client definitions supporting both public and confidential client types. Issuer URL validation enforces HTTPS per OIDC Core Section 3.1.2.1 and RFC 8414, with a localhost exception for development workflows.

This configuration layer is consumed by the upcoming `newServer()` implementation which creates the fosite-based OAuth 2.0 authorization server. The server calls `applyDefaults()` to fill in missing values and `Validate()` to ensure correctness before constructing the `AuthorizationServerParams` that configure the OAuth2 provider, handlers, and endpoints. The "just works" philosophy enables quick development startup with ephemeral secrets while warning about their limitations, and the config cleanly separates concerns: this layer handles what the server needs, while the server implementation handles how it delivers OAuth/OIDC functionality.

### DIFF
--- a/pkg/authserver/config.go
+++ b/pkg/authserver/config.go
@@ -70,6 +70,23 @@ type Config struct {
 	// At least one upstream is required - the server delegates authentication to the upstream IDP.
 	// Currently only a single upstream is supported.
 	Upstreams []UpstreamConfig
+
+	// ScopesSupported lists the OAuth 2.0 scope values advertised in discovery documents.
+	// If nil or empty, defaults to ["openid", "offline_access"].
+	// This is advertised in /.well-known/openid-configuration and
+	// /.well-known/oauth-authorization-server discovery endpoints.
+	ScopesSupported []string
+
+	// AllowedAudiences is the list of valid resource URIs that tokens can be issued for.
+	// Per RFC 8707, the "resource" parameter in authorization and token requests is
+	// validated against this list. MCP clients are required to include the resource
+	// parameter, so this should be configured with the canonical URIs of all MCP servers
+	// this authorization server issues tokens for.
+	//
+	// Security: An empty list means NO audiences are permitted (secure default).
+	// When empty, any request with a "resource" parameter will be rejected with
+	// "invalid_target". Configure this for proper MCP specification compliance.
+	AllowedAudiences []string
 }
 
 // GetUpstream returns the primary upstream configuration.
@@ -101,6 +118,13 @@ func (c *Config) Validate() error {
 
 	if err := c.validateUpstreams(); err != nil {
 		return err
+	}
+
+	// AllowedAudiences is required for MCP compliance.
+	// Per MCP specification, clients MUST include the "resource" parameter (RFC 8707),
+	// which requires the server to have configured allowed audiences to validate against.
+	if len(c.AllowedAudiences) == 0 {
+		return fmt.Errorf("at least one allowed audience is required for MCP compliance (RFC 8707 resource parameter validation)")
 	}
 
 	logger.Debugw("authserver config validation passed",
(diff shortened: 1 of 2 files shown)


pkg/authserver/config.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/stacklok/toolhive/pkg/authserver/upstream"
+	"github.com/stacklok/toolhive/pkg/logger"
+	"github.com/stacklok/toolhive/pkg/networking"
+	"github.com/stacklok/toolhive/pkg/oauth"
+)
+
+// Config is the pure configuration for the OAuth authorization server.
+// All values must be fully resolved (no file paths, no env vars).
+// This is the interface that consumers should use to configure the server.
+type Config struct {
```
jhrozek, line 35: @tgrunnagle this is not user-facing API, but it /should/ be along the lines of the CRDs (the CRDs will generate this config in one way or the other)

pkg/authserver/config.go
```diff
@@ -0,0 +1,229 @@
...
+	// Issuer is the issuer identifier for this authorization server.
+	// This will be included in the "iss" claim of issued tokens.
+	Issuer string
+
+	// KeyProvider provides signing keys for JWT operations.
+	// Supports key rotation by returning multiple public keys for JWKS.
+	// If nil, an ephemeral key will be auto-generated (development only).
```
jhrozek, line 42: @tgrunnagle I was wondering about a `devmode: true` option to be added to the CRDs to enable someone to kick the tires without having to create the keys and mount them. It would have to come with a disclaimer that the keys are ephemeral, but might make the "time to demo" shorter.

pkg/authserver/config.go
```diff
@@ -0,0 +1,229 @@
...
+	// RefreshTokenLifespan is the duration that refresh tokens are valid.
+	// If zero, defaults to 7 days.
+	RefreshTokenLifespan time.Duration
+
+	// AuthCodeLifespan is the duration that authorization codes are valid.
+	// If zero, defaults to 10 minutes.
+	AuthCodeLifespan time.Duration
+
+	// Clients is the list of pre-registered OAuth clients.
+	Clients []ClientConfig
```
jhrozek: Currently unused, we could remove this tbh.

pkg/authserver/config.go
tgrunnagle: I think keeping the code simpler and providing clear instructions to create and mount the keys in a cluster would be better. If there is a discoverable example manifest it shouldn't take more than a minute to set up?

pkg/authserver/config.go
tgrunnagle: Let's remove it if not used - we can always add it later

pkg/authserver/config.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// Upstream contains configuration for connecting to an upstream IDP.
+	// This field is required - the server delegates authentication to the upstream IDP.
+	Upstream *upstream.OAuth2Config
```
tgrunnagle: Should we start this as a slice with validation limiting it to one element? Or since this isn't user facing we can change the code later

pkg/authserver/config.go
tgrunnagle: It looks like `applyDefaults` handles "dev mode" already. I don't think an additional config field is necessary

PR #3485, opened 2026-01-28
Add E2E tests for secrets management

dmjb:
Due to dependency issues (keychain for encrypted, or 1password), these tests only test the environment type. This does not cover all secrets management functionality, but verifies basic correctness of the secrets endpoints.

### DIFF
--- a/test/e2e/api_secrets_test.go
+++ b/test/e2e/api_secrets_test.go
@@ -389,11 +389,8 @@ var _ = Describe("Secrets API", Label("api", "secrets", "e2e"), func() {
 				resp := updateSecret(apiServer, "test-key", "")
 				defer resp.Body.Close()
 
-				By("Verifying response status is 400 or 405")
-				Expect(resp.StatusCode).To(SatisfyAny(
-					Equal(http.StatusBadRequest),
-					Equal(http.StatusMethodNotAllowed),
-				))
+				By("Verifying response status is 400")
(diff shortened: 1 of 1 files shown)


test/e2e/api_secrets_test.go
```diff
@@ -0,0 +1,420 @@
...
+				Expect(resp.StatusCode).To(Equal(http.StatusMethodNotAllowed))
+			})
+
+			It("should reject creating secret with empty key", func() {
+				By("Attempting to create secret with empty key")
+				resp := createSecret(apiServer, "", "some-value")
+				defer resp.Body.Close()
+
+				By("Verifying response status is 400 or 405")
+				Expect(resp.StatusCode).To(SatisfyAny(
```
eleftherias: Is it possible that it'll return 2 different status codes today, or is this for future-proofing?

test/e2e/api_secrets_test.go
```diff
@@ -0,0 +1,477 @@
...
+				Expect(resp.StatusCode).To(Equal(http.StatusMethodNotAllowed))
+			})
+
+			It("should reject updating with empty value", func() {
+				By("Attempting to update with empty value")
+				resp := updateSecret(apiServer, "test-key", "")
+				defer resp.Body.Close()
+
+				By("Verifying response status is 400 or 405")
+				Expect(resp.StatusCode).To(SatisfyAny(
```
eleftherias: Should this only check for `StatusMethodNotAllowed`?

PR #3488, opened 2026-01-28
Add embeddedAuthServer type to MCPExternalAuthConfig CRD

tgrunnagle:
## Summary

Add the `embeddedAuthServer` authentication type to the `MCPExternalAuthConfig` CRD, enabling MCP servers in Kubernetes to integrate with an embedded OAuth2/OIDC authorization server.

**Issue**: stacklok/stacklok-epics#226

## Changes

- Add `ExternalAuthTypeEmbeddedAuthServer` constant with value `"embeddedAuthServer"`
- Add `EmbeddedAuthServerConfig` struct with configuration for:
  - Issuer URL (HTTPS required, no trailing slash per RFC 8414)
  - Signing key secret references (1-5 keys for rotation)
  - HMAC secret references for opaque token signing
  - Token lifespan configuration (access, refresh, auth code)
  - Upstream provider configuration (OIDC or OAuth2)
  - Allowed audiences for RFC 8707 resource validation
- Add `TokenLifespanConfig`, `UpstreamProviderConfig`, `OIDCUpstreamConfig`, and `OAuth2UpstreamConfig` structs
- Add webhook validation ensuring:
  - Config exclusivity (embeddedAuthServer not set with other auth types)
  - Single upstream provider limit (runtime validation for future multi-IDP support)
  - Upstream provider type/config consistency
- Add placeholder handling in controller code for future integration
- Bump operator-crds chart version to 0.0.105

## Test Plan

- [x] New unit tests for embeddedAuthServer webhook validation (15+ test cases)
- [x] Existing webhook tests pass
- [x] CRD generation succeeds (`task operator-generate && task operator-manifests`)
- [x] Helm template renders correctly
- [x] Chart linting passes

## Large PR Justification

- Most lines are generated.
- Majority of remaining lines are unit tests with no logical redundancy.


cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
```diff
@@ -120,6 +129,243 @@ type BearerTokenConfig struct {
...
+	// Currently only a single upstream provider is supported (validated at runtime).
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	UpstreamProviders []UpstreamProviderConfig `json:"upstreamProviders"`
+
+	// AllowedAudiences is the list of valid resource URIs that tokens can be issued for.
+	// Per RFC 8707, the "resource" parameter in authorization and token requests is
+	// validated against this list.
+	// +optional
+	AllowedAudiences []string `json:"allowedAudiences,omitempty"`
```
jhrozek: I don't think the AllowedAudiences should be part of this type - sorry I missed that in the initial RFC. The fact that we even have an array in the Go package was me planning for the shared authserver deployment where we'd track the allowed audiences based on what servers are using this authserver instance.

I think an interesting dichotomy here is - each MCPServer will get its own authserver instance. This means that we can infer the allowed audience from this server's `oidcConfig.inline.audience`  (or equivalent for oidc.Config.configMap/kubernetes). We can also default the scopesSupported to the scopes from that section.

However, I think it would be nice UX if the externalauthconfig could be shared and used by multiple MCPServer instances - in that case audience must not be shared (and thus not part of this CRD).

It's a similar issue with the scopes parameter.

jhrozek: Umm re-reading my comment I realize that maybe it might be easier to understand with an example: we'd have a single MCPExternalAuthConfig pointing with an AuthServer definition with the same issuer (although the issuer could very well default to resourceUrl I guess). The MCPExternalAuthConfig would list a github IDP, pointing to a secret and client_id.

This ExternalAuthConfig could then be linked to by a MCP server that allows listing github issues (aud=mcp-github-issues-lister, scopes=[github:issues]) and another MCP server that allows reading github repos (aud=mcp-github-repos-reader, scopes=[github:repos])

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
```diff
@@ -120,6 +129,243 @@ type BearerTokenConfig struct {
...
+	// +optional
+	AuthCodeLifespan string `json:"authCodeLifespan,omitempty"`
+}
+
+// UpstreamProviderType identifies the type of upstream Identity Provider.
+type UpstreamProviderType string
+
+const (
+	// UpstreamProviderTypeOIDC is for OIDC providers with discovery support
+	UpstreamProviderTypeOIDC UpstreamProviderType = "oidc"
```
jhrozek, line 207: FYI I haven't merged the OIDC provider yet. Will do soon.

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
```diff
@@ -120,6 +129,243 @@ type BearerTokenConfig struct {
...
+	ClientID string `json:"clientId"`
+
+	// ClientSecretRef references a Kubernetes Secret containing the OAuth 2.0 client secret.
+	// Optional for public clients using PKCE instead of client secret.
+	// +optional
+	ClientSecretRef *SecretKeyRef `json:"clientSecretRef,omitempty"`
+
+	// RedirectURI is the callback URL where the upstream IDP will redirect after authentication.
+	// +kubebuilder:validation:Required
+	RedirectURI string `json:"redirectUri"`
```
jhrozek: This is another piece that could prevent from sharing the mcpexternalauthconfig CR - I wonder if we could default to resourceUrl/oauth/callback?

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_webhook.go
```diff
@@ -124,5 +135,78 @@ func (r *MCPExternalAuthConfig) validateUnauthenticated() error {
...
+}
+
+// validateUpstreamProviders validates the upstream provider configurations
+func (r *MCPExternalAuthConfig) validateUpstreamProviders() error {
+	cfg := r.Spec.EmbeddedAuthServer
+	if cfg == nil {
+		return nil
+	}
+
+	// Note: MinItems=1 and MaxItems=1 are enforced by kubebuilder markers,
```
jhrozek: nit: only minItems exists

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_webhook.go
```diff
@@ -82,6 +84,9 @@ func (r *MCPExternalAuthConfig) validateTokenExchange() error {
 	if r.Spec.BearerToken != nil {
 		return fmt.Errorf("bearerToken must not be set when type is 'tokenExchange'")
 	}
+	if r.Spec.EmbeddedAuthServer != nil {
```
yrobla: i see lots of code duplication here, maybe we can create a helper for validating embeddedauthserver, passing some param for different error logs?

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
tgrunnagle: I see why AllowedAudiences wouldn't be necessary in an embedded AS.

I see that the MCPServer type can specify `scopes` in `OIDCConfigRef` - is that how we would derive the AS's `.well-known/oauth-authorization-server` `scopes_supported` field?

IIUC `MCPServerSpec.OIDCConfigRef` is already used to populate the `scopes_supported` on the MCP Server's `.well-known/oauth-protected-resource`, please correct me if I'm mistaken.

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
tgrunnagle: Seems reasonable to default to that (though sharing MCPExternalAuthConfigs with embedded ASs seems to be getting strained). How do you determine `resourceUrl`?

tgrunnagle: Oh I see `OIDCConfigRef.ResourceURL`

PR #3487, opened 2026-01-28
Implement bearer token controller logic and environment variable management

amirejaz:
## Summary
Implements the controller logic and runtime integration for bearer token authentication in Kubernetes, building on the foundation laid in PR #3224. This completes the bearer token authentication flow by adding reconciliation, RunConfig generation, environment variable management, and secret watching capabilities.

## Context
This PR builds on PR #3224 which added:
- `BearerTokenConfig` CRD type in `MCPExternalAuthConfig`
- `ExternalAuthTypeBearerToken` enum value
- Webhook validation for bearer token configuration
- Basic type definitions and API structure

This PR implements the operational logic to make bearer tokens work end-to-end.

### Controller Implementation
- **MCPExternalAuthConfig Controller:**
  - Added reconciliation logic for `bearerToken` type in `mcpexternalauthconfig_controller.go`
  - Implements secret watch to trigger reconciliation when referenced secrets change
  - Calculates config hash including secret content (SHA256, truncated to 16 hex chars) for change detection
  - Refactored secret reference finding into dedicated functions:
    - `findMCPExternalAuthConfigsReferencingSecret()` - finds configs referencing a secret
    - `configReferencesSecret()` - checks if a config references a specific secret

- **MCPRemoteProxy Controller:**
  - Integrated bearer token configuration into RunConfig generation (`mcpremoteproxy_runconfig.go`)
  - Converts Kubernetes Secret references to CLI format (`"secret-name,target=bearer_token"`)
  - Validates secret existence and key presence before deployment
  - Generates environment variables for bearer token secrets

### Environment Variable Management
- **`GenerateBearerTokenEnvVar()`:** Creates `TOOLHIVE_SECRET_{secret-name}` env vars from Secret references
- **`EnsureRequiredEnvVars()`:** Auto-detects `TOOLHIVE_SECRET_*` env vars and sets `TOOLHIVE_SECRETS_PROVIDER=environment`
  - Checks all env vars (not just specific ones) for `TOOLHIVE_SECRET_*` prefix
  - Only sets provider when actual secrets are present (excludes provider env var itself)
  - Ensures secrets provider is correctly configured for CLI-format secret resolution

### Secret Resolution Flow
1. User creates `MCPExternalAuthConfig` with `bearerToken` type referencing a Kubernetes Secret
2. Controller validates secret exists and contains required key
3. Controller generates `TOOLHIVE_SECRET_{secret-name}` env var in pod spec
[...]

PR #3497, opened 2026-01-29
Add secret-aware hash calculation and secret watching for MCPExternalAuthConfig

amirejaz:
## Summary

Enhances `MCPExternalAuthConfig` controller to include secret values in hash calculation and adds secret watching to automatically trigger reconciliation when referenced secrets change.

**Depends on:** PR #3487

## Changes

- **Secret-aware hash calculation**: Includes bearer token and token exchange client secret values in `configHash` to detect secret rotations
- **Secret watching**: Automatically reconciles `MCPExternalAuthConfig` resources when their referenced secrets change
- **Comprehensive test coverage**: Added tests for secret-aware hashing and secret watching functionality

## Benefits

- Automatic reconciliation on secret rotation
- Accurate change detection (hash includes secret values)
- Consistent with existing watch patterns in the codebase

## Testing

All tests pass, including new tests for secret-aware hashing and secret watching.

### DIFF
--- a/cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
+++ b/cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
@@ -5,9 +5,12 @@ package controllers
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


jhrozek:
I don't love that this adds the ability to read secrets, but if we want to support updates when secrets change as a feature then we don't have a lot of options.

One thing that should be improved is that the PR seems to watch for /all/ secrets? Wouldn't it be better to implement predicates and filter by secrets based on e.g. labels or some naming conventions?

jhrozek:
Would it be acceptable to poll for secrets to be able to limit the RBAC permissions to get instead?

Ideally we'd split the secret watching into a separate process that only does that and it not exposed to the outside unlike mcpremote proxy

amirejaz:
@jhrozek Good points. This is more of a nice-to-have feature. The core functionality without Secret watching is implemented in a separate, currently open PR ([#3499](https://github.com/stacklok/toolhive/pull/3499)), and I agree that the security concerns outweigh the benefit with the current implementation, especially when the operator runs cluster-scoped.

I like the idea of splitting Secret watching into a separate internal process that has the required privileges but is not externally exposed.

As an interim step, would it make sense to avoid watching Secrets for now and rely on get-only access with periodic reconciliation to detect changes? This would let us keep least-privilege defaults while we design the separate secret-watcher component.

Happy to either defer this feature until then, or adjust the current PR in that direction if that’s preferable.

ChrisJBurns:
> Would it be acceptable to poll for secrets to be able to limit the RBAC permissions to get instead?

I prefer this over the over permissive process of being able to read secrets for sure

PR #3499, opened 2026-01-29
Implement ExcludeAll feature for VirtualMCPServer tool filtering

### LINKED ISSUE #2779: Add ExcludeAll option for tool filtering in VirtualMCPServer aggregation
## Problem

The current `WorkloadToolConfig.Filter` field is an allow-list of tool names:
- `Filter: ["tool1", "tool2"]` → allows only those tools
- `Filter: []` or `Filter: nil` → allows ALL tools (no filtering applied)

There's **no way to exclude all tools** from a backend in VirtualMCPServer aggregation. This limits flexibility in several scenarios:

1. **Composite-only VirtualMCPServer**: You want to expose only composite tools without any direct backend tools
2. **Temporary disabling**: Disable all tools from a backend without removing it from the group
3. **Selective enablement**: Keep group membership but control tool exposure at the VirtualMCPServer level

## Technical Context

Using Go's `nil` vs empty slice distinction (`nil` = allow all, `[]` = allow none) is elegant but problematic with Kubernetes CRDs due to JSON serialization with `omitempty` - the distinction is lost after API server round-trips.

### 1. Per-workload `excludeAll` boolean

Add an `ExcludeAll` field to `WorkloadToolConfig`:

```yaml
spec:
  aggregation:
    tools:
      - workload: backend-a
        filter: ["echo"]        # Allow only echo tool
      - workload: backend-b
        excludeAll: true        # Exclude all tools from this backend
```

```go
type WorkloadToolConfig struct {
    Workload   string            `json:"workload"`
    Filter     []string          `json:"filter,omitempty"`
    Overrides  map[string]...    `json:"overrides,omitempty"`
    ExcludeAll bool              `json:"excludeAll,omitempty"`  // NEW
}
```

### 2. Global `excludeAllTools` at aggregation level

Add a global flag to exclude all tools from all backends, useful for composite-only VirtualMCPServers:

```yaml
spec:
  aggregation:
    excludeAllTools: true       # Exclude all backend tools globally
    conflictResolution: prefix
  composite:
    tools:
      - name: my-workflow
        # ... composite tool definition
```

```go
type AggregationConfig struct {
    ConflictResolution string                `json:"conflictResolution,omitempty"`
    Tools              []WorkloadToolConfig  `json:"tools,omitempty"`
    ExcludeAllTools    bool                  `json:"excludeAllTools,omitempty"`  // NEW
}
```

This enables a clean pattern for **composite-only VirtualMCPServers** that orchestrate backend tools without exposing them directly.

## Behavior Matrix

| Config | Behavior |
|--------|----------|
| `excludeAllTools: true` | All backend tools excluded globally |
| `excludeAll: true` (per-workload) | All tools from that specific backend excluded |
| `filter: ["tool1"]` | Only listed tools allowed |
| `filter: []` or omitted | All tools allowed (current behavior) |

## Use Case Example

Expose only a composite workflow tool, hiding the underlying backend tools:

```yaml
apiVersion: mcp.toolhive.stacklok.com/v1alpha1
kind: VirtualMCPServer
metadata:
  name: workflow-only-vmcp
spec:
  groupRef:
    name: my-backends
  aggregation:
    excludeAllTools: true  # Hide all direct backend tools
  composite:
    tools:
      - name: analyze-and-report
        description: "Runs analysis workflow across multiple backends"
        steps:
          - name: fetch-data
            tool: backend-a/fetch
          - name: process
            tool: backend-b/process
            dependsOn: [fetch-data]
```

Clients see only `analyze-and-report`, not the underlying `fetch` or `process` tools.

JAORMX:
## Summary

Implements runtime support for the `excludeAll` configuration fields that were already defined in the CRD schema but not implemented:

- **Global exclusion**: `aggregation.excludeAllTools: true` excludes all tools from all backends
- **Per-workload exclusion**: `aggregation.tools[].excludeAll: true` excludes all tools from a specific backend

The `ExcludeAll` check takes precedence over `Filter` and `Overrides` settings, allowing users to easily exclude all tools from a backend without needing a workaround like `filter: ["nonexistent_tool"]`.

### Example Usage

```yaml
# Per-workload exclusion
aggregation:
  tools:
    - workload: backend-to-exclude
      excludeAll: true

# Global exclusion (resources-only vMCP)
aggregation:
  excludeAllTools: true
```

## Test plan

- [x] Unit tests pass (`task test`)
- [x] Linter passes (`task lint`)
- [ ] E2E tests pass (`task test-e2e`)

Fixes #2779

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pkg/vmcp/aggregator/default_aggregator.go
+++ b/pkg/vmcp/aggregator/default_aggregator.go
@@ -98,13 +98,11 @@ func (a *defaultAggregator) QueryCapabilities(ctx context.Context, backend vmcp.
 	}
 
 	// Apply per-backend tool filtering and overrides (before conflict resolution)
-	var processedTools []vmcp.Tool
-	if a.excludeAllTools {
-		logger.Debugf("ExcludeAllTools is true globally, returning empty tools for backend %s", backend.ID)
-		processedTools = []vmcp.Tool{}
-	} else {
-		processedTools = processBackendTools(ctx, backend.ID, capabilities.Tools, a.toolConfigMap[backend.ID])
-	}
+	// NOTE: ExcludeAll (both global and per-workload) is NOT applied here.
+	// This is intentional - we need all tools in the routing table so composite
+	// tools can call backend tools. ExcludeAll is applied in MergeCapabilities
+	// to control which tools are advertised to the LLM.
+	processedTools := processBackendTools(ctx, backend.ID, capabilities.Tools, a.toolConfigMap[backend.ID])
 
 	// Convert to BackendCapabilities
 	result := &BackendCapabilities{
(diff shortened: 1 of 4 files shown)


jerm-dro (approved):
Thanks for getting this over the finish line! Nitpicking: you could add one test that demonstrates ExcludeAll does not prevent the backend's tools from being used by composite tools.

PR #3504, opened 2026-01-29
Limit RBAC permissions for inline mode VirtualMCPServers

yrobla:
In inline mode, VirtualMCPServer pods receive all backend configuration through the VirtualMCPServer spec and don't need to discover backends from Kubernetes resources. However, they were still granted full RBAC permissions including access to secrets and configmaps.

Since vMCP is exposed to the outside world via HTTP, granting unnecessary Kubernetes API permissions increases the security risk. This change implements conditional RBAC based on the outgoing auth source mode:

  - Inline mode: Minimal permissions (read own spec + update status)
  - Discovered mode: Full permissions (read secrets, configmaps, MCP resources)

The implementation creates two separate RBAC rule sets and selects the appropriate one based on spec.outgoingAuth.source. Existing resources default to discovered mode for backward compatibility.

Added comprehensive tests to verify correct permissions are granted for each mode, including validation that inline mode has no secret or configmap access while still maintaining status update capabilities.

Also removed orphaned comment and nolint directive for deleted discoverBackends function.

Related-to: #3149

### DIFF
--- a/cmd/thv-operator/controllers/virtualmcpserver_controller.go
+++ b/cmd/thv-operator/controllers/virtualmcpserver_controller.go
@@ -16,6 +16,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -510,18 +511,26 @@ func (r *VirtualMCPServerReconciler) ensureRBACResources(
...
 	rbacClient := rbac.NewClient(r.Client, r.Scheme)
 	serviceAccountName := vmcpServiceAccountName(vmcp.Name)
 
-	// Ensure Role with permissions to discover backends and update status
+	// Select RBAC rules based on outgoing auth mode
+	// - inline mode: Minimal permissions (read own spec + update status)
+	// - discovered mode: Full permissions (read secrets, configmaps, MCP resources + update status)
+	rules := func() []rbacv1.PolicyRule {
+		// Default to discovered mode if not specified
+		if vmcp.Spec.OutgoingAuth == nil || vmcp.Spec.OutgoingAuth.Source == "" || vmcp.Spec.OutgoingAuth.Source == "discovered" {
```
jhrozek: let's use the OutgoingAuthSourceDiscovered constant here

cmd/thv-operator/controllers/virtualmcpserver_controller_test.go
```diff
@@ -498,6 +498,143 @@ func TestVirtualMCPServerEnsureRBACResources_StaticMode(t *testing.T) {
 	assert.NoError(t, err, "RoleBinding should be created in static mode for status reporting")
 }
 
+// TestVirtualMCPServerEnsureRBACResources_InlineMode tests that inline mode uses
+// minimal RBAC permissions (no secret/configmap access) for security
+func TestVirtualMCPServerEnsureRBACResources_InlineMode(t *testing.T) {
```
jhrozek, line 437: this seems to be the same test as StaticMode?

PR #3513, opened 2026-01-29
Add core OAuth authorization server interface and implementation

jhrozek:
This commit introduces a centralized OAuth 2.0 Authorization Server for ToolHive, enabling MCP clients to authenticate via external Identity Providers (like Google, Okta) and receive JWTs for accessing MCP servers.

Why this change:
- MCP clients need a way to authenticate users and obtain tokens
- ToolHive needs to issue its own JWTs rather than passing through IDP tokens
- Centralizing auth allows consistent token format across all MCP servers

Core changes:

Server interface (server.go):
- Handler() returns HTTP handler for all OAuth/OIDC endpoints
- IDPTokenStorage() provides access to upstream IDP tokens for middleware
- Close() releases resources

Endpoints served:
- /.well-known/openid-configuration (OIDC Discovery)
- /.well-known/oauth-authorization-server (RFC 8414)
- /.well-known/jwks.json (JSON Web Key Set)
- /oauth/authorize (Authorization endpoint)
- /oauth/token (Token endpoint)
- /oauth/callback (Upstream IDP callback)
- /oauth/register (Dynamic Client Registration, RFC 7591)

Drive-by fixes discovered during integration:
- Add ScopesSupported to OAuth metadata (RFC 8414 compliance)
- Remove duplicate ScopesSupported field from OIDCDiscoveryDocument that shadowed the embedded AuthorizationServerMetadata field
- Sanitize upstream IDP error messages to avoid exposing internal details to clients (security hardening)
- Add mockgen directive for upstream.OAuth2Provider interface - to be able to properly test the new code


pkg/authserver/server_impl.go
```diff
@@ -0,0 +1,216 @@
...
+	"github.com/stacklok/toolhive/pkg/authserver/storage"
+	"github.com/stacklok/toolhive/pkg/authserver/upstream"
+	"github.com/stacklok/toolhive/pkg/logger"
+)
+
+// server is the internal implementation of the Server interface.
+type server struct {
+	handler     http.Handler
+	storage     storage.Storage
+	upstreamIDP upstream.OAuth2Provider
```
tgrunnagle, line 26: should this be some wrapper around either an OAuth2 or OIDC provider? (for when OIDC is added)

jhrozek: yes, the way I'm thinking this would go would be that I'd have an upstream.Provider and then we'll have upstream.OAuth2Provider and upstream.OIDCProvider.

PR #3520, opened 2026-01-29
Add Embedded Auth Server Configuration Support to Operator

tgrunnagle:
Depends on #3488 

## Summary

This PR enables the Kubernetes operator to configure an embedded OAuth2/OIDC authorization server within MCP server proxy runners.

This implementation follows [RFC-0031](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0031-auth-server-integration.md), which specifies integrating the existing ToolHive auth server (`pkg/authserver/`) directly into proxy runner pods.

## Key Changes

**CRD Extension**: Added `embeddedAuthServer` type to `MCPExternalAuthConfig` with full configuration for:
- Signing keys (mounted as read-only volumes for JWT signing)
- HMAC secrets (for internal token encryption with rotation support)
- Token lifespans (access, refresh, authorization code)
- Upstream provider settings (OIDC discovery or explicit OAuth2 endpoints)

**Controller Utilities**: New `controllerutil/authserver.go` provides functions to:
- Generate secure volumes/mounts for cryptographic material (0400 permissions)
- Inject upstream client secrets via environment variables
- Build `EmbeddedAuthServerConfig` for the RunConfig

**RunConfig Extension**: Added `EmbeddedAuthServerConfig` struct and builder option to carry auth server configuration from operator to proxy runner.

## Closes

- stacklok/stacklok-epics#228 - Add EmbeddedAuthServerConfig to RunConfig
- stacklok/stacklok-epics#229 - Implement controller utility functions for embedded auth server

## Testing

- Unit tests for all new controller utility functions (>80% coverage)
- RunConfig serialization/deserialization tests
- Existing operator tests pass

## Large PR Justification

- Majority of lines are tests tightly coupled with new config types, or generated doc updates

### DIFF
--- a/cmd/thv-operator/pkg/controllerutil/authserver_test.go
+++ b/cmd/thv-operator/pkg/controllerutil/authserver_test.go
@@ -744,24 +744,27 @@ func TestAddEmbeddedAuthServerConfigOptions_Validation(t *testing.T) {
 	err := mcpv1alpha1.AddToScheme(scheme)
 	require.NoError(t, err)
 
-	// Valid embedded auth server config
-	validExternalAuthConfig := &mcpv1alpha1.MCPExternalAuthConfig{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "embedded-auth-config",
-			Namespace: "default",
-		},
-		Spec: mcpv1alpha1.MCPExternalAuthConfigSpec{
-			Type: mcpv1alpha1.ExternalAuthTypeEmbeddedAuthServer,
-			EmbeddedAuthServer: &mcpv1alpha1.EmbeddedAuthServerConfig{
-				Issuer: "https://auth.example.com",
-				SigningKeySecretRefs: []mcpv1alpha1.SecretKeyRef{
-					{Name: "signing-key", Key: "private.pem"},
-				},
-				HMACSecretRefs: []mcpv1alpha1.SecretKeyRef{
-					{Name: "hmac-secret", Key: "hmac"},
+	// Helper function to create a fresh external auth config for each test
+	// This avoids data races when running subtests in parallel
+	newExternalAuthConfig := func() *mcpv1alpha1.MCPExternalAuthConfig {
+		return &mcpv1alpha1.MCPExternalAuthConfig{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "embedded-auth-config",
+				Namespace: "default",
+			},
+			Spec: mcpv1alpha1.MCPExternalAuthConfigSpec{
+				Type: mcpv1alpha1.ExternalAuthTypeEmbeddedAuthServer,
+				EmbeddedAuthServer: &mcpv1alpha1.EmbeddedAuthServerConfig{
+					Issuer: "https://auth.example.com",
+					SigningKeySecretRefs: []mcpv1alpha1.SecretKeyRef{
+						{Name: "signing-key", Key: "private.pem"},
+					},
+					HMACSecretRefs: []mcpv1alpha1.SecretKeyRef{
+						{Name: "hmac-secret", Key: "hmac"},
+					},
 				},
 			},
-		},
+		}
 	}
 
 	tests := []struct {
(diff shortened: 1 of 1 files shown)


cmd/thv-operator/controllers/mcpremoteproxy_deployment.go
```diff
@@ -31,6 +31,21 @@ func (r *MCPRemoteProxyReconciler) deploymentForMCPRemoteProxy(
...
 	env := r.buildEnvVarsForProxy(ctx, proxy)
+
+	// Add embedded auth server volumes and env vars if configured (single call for efficiency)
+	if proxy.Spec.ExternalAuthConfigRef != nil {
+		authServerVolumes, authServerMounts, authServerEnvVars, err := ctrlutil.GenerateAuthServerConfig(
+			ctx, r.Client, proxy.Namespace, proxy.Spec.ExternalAuthConfigRef,
+		)
+		if err != nil {
+			ctxLogger := log.FromContext(ctx)
+			ctxLogger.Error(err, "Failed to generate embedded auth server configuration")
```
jhrozek, line 42: I think we should error after this line.

In general I prefer not to write:
```
if err != nil {
 // oops
} else {
 // hurray
} 
```

but rather:
```go
if err != nil {
 // oops
}

// hooray
```

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -1046,6 +1046,21 @@ func (r *MCPServerReconciler) deploymentForMCPServer(
...
 	}
 
+	// Add embedded auth server volumes and env vars if configured
+	if m.Spec.ExternalAuthConfigRef != nil {
+		authServerVolumes, authServerMounts, authServerEnvVars, err := ctrlutil.GenerateAuthServerConfig(
+			ctx, r.Client, m.Namespace, m.Spec.ExternalAuthConfigRef,
+		)
+		if err != nil {
+			ctxLogger := log.FromContext(ctx)
+			ctxLogger.Error(err, "Failed to generate embedded auth server configuration")
```
jhrozek, line 1056: same as above?

seeing the same pattern twice makes me wonder if the graceful error handling was maybe intended, but then I'd prefer to see a comment explaining why we only log the error.

Is it because the controller would retry and there's no point in retrying?

cmd/thv-operator/pkg/controllerutil/tokenexchange.go
```diff
@@ -125,9 +125,7 @@ func AddExternalAuthConfigOptions(
 		// No config to add for unauthenticated
 		return nil
 	case mcpv1alpha1.ExternalAuthTypeEmbeddedAuthServer:
-		// Embedded auth server config is handled separately (via volume mounts)
-		// Controller integration will be in a future task
-		return nil
+		return AddEmbeddedAuthServerConfigOptions(ctx, c, namespace, externalAuthConfigRef, options)
```
jhrozek: for sometimes later: I find it odd this logic is in tokenexchange.go ...

jhrozek: (not your fault obv)

cmd/thv-operator/pkg/controllerutil/authserver.go
```diff
@@ -0,0 +1,381 @@
...
+	AuthServerHMACVolumePrefix = "authserver-hmac-secret-"
+
+	// AuthServerKeysMountPath is the base path where signing keys are mounted
+	AuthServerKeysMountPath = "/etc/toolhive/authserver/keys"
+
+	// AuthServerHMACMountPath is the base path where HMAC secrets are mounted
+	AuthServerHMACMountPath = "/etc/toolhive/authserver/hmac"
+
+	// AuthServerKeyFilePattern is the pattern for signing key filenames
+	AuthServerKeyFilePattern = "key-%d.pem"
```
jhrozek, line 35: praise: %d for rotation support

cmd/thv-operator/controllers/mcpserver_controller.go
tgrunnagle: Looks like the pattern is return `nil` here if there is an error, then the controller check for nil and then does a `return ctrl.Result{}, fmt.Errorf("failed to create updated Deployment object")`

cmd/thv-operator/controllers/mcpremoteproxy_deployment.go
tgrunnagle: Updating to log and return `nil` if there is an error - the caller (controller) will then raise an error.

PR #3523, opened 2026-01-29
triggers docs at the end of the release workflow

ChrisJBurns:


.github/workflows/releaser.yml
```diff
@@ -189,7 +189,7 @@ jobs:
 
   update-docs-website-binaries:
     name: Trigger Docs Update for Binaries
```
danbarr: Nit, but "for Binaries" not needed anymore since everything's unified?

```suggestion
update-docs-website:
name: Trigger Docs Update
```

ChrisJBurns: Ahh yes good spot, i forgot about that!

Issue #3036, opened 2025-12-15
Add health monitoring and circuit breaker to vMCP server


yrobla:
## Description

Implement health monitoring and circuit breaker patterns to detect and handle backend failures gracefully.

**Scope:**
- Periodic health checks for backend MCP servers
- Backend health status tracking and reporting
- Circuit breaker implementation for failing backends
- Automatic backend removal/restoration based on health
- Health status reflected in vMCP status and capabilities

### 1. Backend Health Checks
- Periodic health check requests to each backend
- Configurable check interval (default: 30s)
- Track consecutive failures
- Mark backend unhealthy after threshold (default: 3 failures)
- Health check endpoint: MCP `ping` or `tools/list`

### 2. Health Status Tracking
- Per-backend health state: `healthy`, `unhealthy`, `unknown`
- Last successful health check timestamp
- Failure count and error messages
- Health status exposed in vMCP status/metrics

### 3. Circuit Breaker
- Three states: `closed` (normal), `open` (failing), `half-open` (testing recovery)
- Configurable failure threshold (default: 5 failures)
- Configurable timeout for open state (default: 60s)
- Automatic transition to half-open for recovery testing
- Track circuit breaker state per backend

### 4. Backend Availability Management
- Remove unhealthy backend tools from aggregated capabilities
- Return error when routing to unavailable backend
- Automatically restore backend when health recovers
- Log backend state transitions (healthy ↔ unhealthy)
- Emit metrics for monitoring systems

### 5. Failure Modes
- **fail mode** (default): Fail entire request if backend unavailable
- **best_effort mode**: Return partial results, include errors for failed backends
- Configurable per vMCP instance

### 6. Integration Points
- Update aggregated capabilities when backend health changes
- Routing layer checks health before forwarding
- Status reporting includes backend health summary
- Metrics exported for Prometheus/observability

## Implementation Notes

- Health checks run in background goroutines
- Use context for cancellation on shutdown
- Circuit breaker prevents cascading failures
- Health status should be cached to avoid repeated checks
- Transition events should be logged and emitted as metrics
- Follow existing ToolHive observability patterns in `pkg/telemetry/`
- Consider using `sony/gobreaker` or similar library

## Reference

- [vMCP Design Proposal](https://github.com/stacklok/toolhive/blob/main/docs/proposals/THV-2106-virtual-mcp-server.md) - Backend Unavailability, Partial Failures, Circuit Breaker sections

## Acceptance Criteria

[...]

PR #3528, opened 2026-01-30
Add circuit breaker pattern to vMCP health monitoring

yrobla:
Implement circuit breaker state machine for backend health monitoring in Virtual MCP Server. The circuit breaker prevents cascading failures by tracking backend failures and transitioning through three states: Closed (normal operation), Open (failing), and HalfOpen (recovery testing).

Key changes:
  - Add CircuitBreaker with thread-safe state machine
  - Integrate circuit breaker with status tracker
  - Integrate circuit breaker with health monitor
  - Add comprehensive test coverage

The circuit breaker is disabled by default for backwards compatibility. Configuration wiring from YAML/CRD to health monitor is intentionally left for a follow-up PR to keep this change focused on the core implementation.

## Large PR Justification

This is an atomic PR that exposes the circuit breaker capability. It includes comprehensive tests to validate it. Cannot be split as it's an isolated functionality.

Related-to: #3036

### DIFF
--- a/pkg/vmcp/health/circuit_breaker.go
+++ b/pkg/vmcp/health/circuit_breaker.go
@@ -22,16 +22,35 @@ const (
 	CircuitHalfOpen CircuitState = "half_open"
 )
 
+// CircuitBreaker defines the interface for circuit breaker implementations.
+type CircuitBreaker interface {
+	// RecordSuccess records a successful operation
+	RecordSuccess()
+	// RecordFailure records a failed operation
+	RecordFailure()
+	// CanAttempt checks if an operation should be allowed based on circuit state
+	CanAttempt() bool
+	// GetState returns the current state of the circuit breaker
+	GetState() CircuitState
+	// GetLastStateChange returns the time when the state last changed
+	GetLastStateChange() time.Time
+	// GetFailureCount returns the current failure count
+	GetFailureCount() int
+	// GetSnapshot returns an immutable snapshot of the circuit breaker state
+	GetSnapshot() circuitBreakerSnapshot
+}
+
 // circuitBreaker manages circuit breaker state for a single backend.
 // It implements the circuit breaker pattern to prevent cascading failures
 // by tracking failures and transitioning through states:
 // Closed → Open → HalfOpen → Closed
 type circuitBreaker struct {
-	mu sync.Mutex
-
 	// name is used for logging purposes to identify which backend this circuit breaker belongs to
+	// name is immutable after initialization and doesn't require mutex protection
 	name string
 
+	mu sync.Mutex
+	// Fields below are protected by mu
 	state            CircuitState
 	failureCount     int
 	failureThreshold int
@@ -67,8 +86,7 @@ func (cb *circuitBreaker) RecordSuccess() {
 	cb.halfOpenTestInProgress = false
 
 	if cb.state != CircuitClosed {
-		cb.state = CircuitClosed
-		cb.lastStateChange = time.Now()
+		cb.transitionTo(CircuitClosed)
 
 		// Log successful recovery
 		if previousState == CircuitHalfOpen {
@@ -83,23 +101,36 @@ func (cb *circuitBreaker) RecordFailure() {
 	cb.mu.Lock()
 	defer cb.mu.Unlock()
 
-	previousState := cb.state
 	cb.failureCount++
 	cb.lastFailureTime = time.Now()
 	cb.halfOpenTestInProgress = false
 
 	if cb.state == CircuitClosed && cb.failureCount >= cb.failureThreshold {
-		cb.state = CircuitOpen
-		cb.lastStateChange = time.Now()
+		cb.transitionTo(CircuitOpen)
 		logger.Warnf("Circuit breaker for backend %s OPENED (threshold exceeded)", cb.name)
 	} else if cb.state == CircuitHalfOpen {
 		// Failed in half-open state, go back to open
-		cb.state = CircuitOpen
-		cb.lastStateChange = time.Now()
-		if previousState == CircuitHalfOpen {
-			logger.Warnf("Circuit breaker for backend %s returned to OPEN from half-open (recovery failed)", cb.name)
-		}
+		cb.transitionTo(CircuitOpen)
+		logger.Warnf("Circuit breaker for backend %s returned to OPEN from half-open (recovery failed)", cb.name)
 	}
+}
+
+// transitionTo changes the circuit breaker state and updates the lastStateChange timestamp.
+// Must be called with lock held.
+func (cb *circuitBreaker) transitionTo(newState CircuitState) {
+	cb.state = newState
+	cb.lastStateChange = time.Now()
+}
+
+// tryTransitionOpenToHalfOpen checks if the circuit is OPEN and timeout has elapsed,
+// and transitions to HALF_OPEN if so. Returns true if transition occurred.
+// Must be called with lock held.
+func (cb *circuitBreaker) tryTransitionOpenToHalfOpen() bool {
+	if cb.state == CircuitOpen && time.Since(cb.lastStateChange) >= cb.timeout {
+		cb.transitionTo(CircuitHalfOpen)
+		return true
+	}
+	return false
 }
 
 // CanAttempt checks if an operation should be allowed based on circuit state.
@@ -114,9 +145,7 @@ func (cb *circuitBreaker) CanAttempt() bool {
 
 	case CircuitOpen:
 		// Check if timeout has elapsed to transition to half-open
-		if time.Since(cb.lastStateChange) >= cb.timeout {
-			cb.state = CircuitHalfOpen
-			cb.lastStateChange = time.Now()
+		if cb.tryTransitionOpenToHalfOpen() {
 			cb.halfOpenTestInProgress = true
 			return true
 		}
@@ -158,6 +187,10 @@ func (cb *circuitBreaker) GetFailureCount() int {
 }
 
 // GetSnapshot returns an immutable snapshot of the circuit breaker state.
+// This is a read-only operation that does not trigger state transitions.
+// The snapshot reflects the current state at the time of the call.
+// Note: If the circuit is OPEN and the timeout has elapsed, the snapshot will still
+// show OPEN until the next call to CanAttempt() triggers the transition to HALF_OPEN.
 func (cb *circuitBreaker) GetSnapshot() circuitBreakerSnapshot {
 	cb.mu.Lock()
 	defer cb.mu.Unlock()
@@ -176,4 +209,44 @@ type circuitBreakerSnapshot struct {
 	FailureCount    int
 	LastStateChange time.Time
 	LastFailureTime time.Time
+}
+
+// alwaysClosedCircuit is a no-op circuit breaker implementation that always allows operations.
+// Used when circuit breaker is disabled.
+type alwaysClosedCircuit struct{}
+
+// RecordSuccess is a no-op for the always-closed circuit.
+func (*alwaysClosedCircuit) RecordSuccess() {}
+
+// RecordFailure is a no-op for the always-closed circuit.
+func (*alwaysClosedCircuit) RecordFailure() {}
+
+// CanAttempt always returns true for the always-closed circuit.
+func (*alwaysClosedCircuit) CanAttempt() bool {
+	return true
+}
+
+// GetState always returns CircuitClosed.
+func (*alwaysClosedCircuit) GetState() CircuitState {
+	return CircuitClosed
+}
+
+// GetLastStateChange returns zero time since the circuit never changes state.
+func (*alwaysClosedCircuit) GetLastStateChange() time.Time {
+	return time.Time{}
+}
+
+// GetFailureCount always returns 0.
+func (*alwaysClosedCircuit) GetFailureCount() int {
+	return 0
+}
+
+// GetSnapshot returns a snapshot representing a closed circuit with no failures.
+func (*alwaysClosedCircuit) GetSnapshot() circuitBreakerSnapshot {
+	return circuitBreakerSnapshot{
+		State:           CircuitClosed,
+		FailureCount:    0,
+		LastStateChange: time.Time{},
+		LastFailureTime: time.Time{},
+	}
 }
--- a/pkg/vmcp/health/circuit_breaker_test.go
+++ b/pkg/vmcp/health/circuit_breaker_test.go
@@ -221,6 +221,43 @@ func TestCircuitBreaker_GetSnapshot(t *testing.T) {
 	assert.True(t, snapshot2.LastStateChange.After(snapshot.LastStateChange))
 }
 
+func TestCircuitBreaker_GetSnapshotIsReadOnly(t *testing.T) {
+	t.Parallel()
+
+	timeout := 50 * time.Millisecond
+	cb := newCircuitBreaker(2, timeout, "test-backend")
+
+	// Open the circuit
+	cb.RecordFailure()
+	cb.RecordFailure()
+	assert.Equal(t, CircuitOpen, cb.GetState())
+
+	// GetSnapshot before timeout - should be OPEN
+	snapshot1 := cb.GetSnapshot()
+	assert.Equal(t, CircuitOpen, snapshot1.State)
+
+	// Wait for timeout to elapse
+	time.Sleep(timeout + 20*time.Millisecond)
+
+	// GetSnapshot after timeout - should STILL be OPEN (GetSnapshot is read-only)
+	snapshot2 := cb.GetSnapshot()
+	assert.Equal(t, CircuitOpen, snapshot2.State)
+	// LastStateChange should not have changed since no transition occurred
+	assert.Equal(t, snapshot1.LastStateChange, snapshot2.LastStateChange)
+
+	// Verify GetState also shows OPEN (no transition until CanAttempt is called)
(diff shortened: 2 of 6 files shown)


pkg/vmcp/health/circuit_breaker_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// Open the circuit
+	for i := 0; i < 3; i++ {
+		cb.RecordFailure()
+	}
+	assert.Equal(t, CircuitOpen, cb.GetState())
+	assert.False(t, cb.CanAttempt())
+
+	// Wait for timeout
+	time.Sleep(timeout + 10*time.Millisecond)
```
jhrozek: The 10ms margin here is pretty tight. Under CI load this could flake - I'd bump it to at least 50ms.

```suggestion
	time.Sleep(timeout + 50*time.Millisecond)
```

Long term, a mock clock would make these tests deterministic, but the margin increase should be fine for now.

pkg/vmcp/health/monitor.go
```diff
@@ -339,6 +398,14 @@ func (m *Monitor) monitorBackend(ctx context.Context, backend *vmcp.Backend, isI
 func (m *Monitor) performHealthCheck(ctx context.Context, backend *vmcp.Backend) {
 	logger.Debugf("Performing health check for backend %s (%s)", backend.Name, backend.BaseURL)
 
+	// Check circuit breaker state before attempting health check
+	if m.circuitBreakerConfig != nil && m.circuitBreakerConfig.Enabled {
+		if !m.statusTracker.CanAttemptHealthCheck(backend.ID) {
+			logger.Debugf("Circuit breaker OPEN for backend %s, skipping health check", backend.Name)
```
jhrozek: This log says "OPEN" but we also skip health checks when in half-open with a test already in progress. Might be worth logging when we actually transition to half-open and start a recovery test - right now that transition happens silently inside CanAttempt() and there's no visibility at the monitor level.

Something like:
```go
if cbState, _ := m.statusTracker.GetCircuitBreakerState(backend.ID); cbState == CircuitHalfOpen {
    logger.Debugf("Circuit breaker testing recovery for backend %s", backend.Name)
}
```

pkg/vmcp/health/monitor_test.go
```diff
@@ -840,3 +896,279 @@ func TestMonitor_UpdateBackends(t *testing.T) {
...
+	mockClient.EXPECT().
+		ListCapabilities(gomock.Any(), gomock.Any()).
+		Return(nil, errors.New("connection failed")).
+		Times(10)
+
+	err = monitor.Start(ctx)
+	require.NoError(t, err)
+
+	// Wait for multiple health checks
+	time.Sleep(1 * time.Second)
```
jhrozek: Fixed sleeps like this tend to be flaky in CI and slower than necessary. Consider using require.Eventually instead:

```go
require.Eventually(t, func() bool {
    state, _ := monitor.GetBackendState("backend-1")
    return state != nil && state.ConsecutiveFailures > 5
}, 2*time.Second, 50*time.Millisecond, "should record multiple failures")
```

pkg/vmcp/health/monitor.go
```diff
@@ -104,6 +108,24 @@ type MonitorConfig struct {
...
+	// Enabled controls whether circuit breaker is active.
+	Enabled bool
+
+	// FailureThreshold is the number of failures before opening the circuit.
+	// Must be >= 1. Recommended: 5 failures.
+	FailureThreshold int
+
+	// Timeout is the duration to wait in open state before attempting recovery.
+	// Recommended: 60s.
+	Timeout time.Duration
```
jerm-dro, line 130: nit: add kubebuilder annotations here. For example:
```
// +kubebuilder:validation:Minimum=1
// +kubebuilder:default=5
// Must be >= 1. Recommended: 5 failures.
FailureThreshold int
```

pkg/vmcp/health/status.go
```diff
@@ -214,6 +285,21 @@ func (t *statusTracker) RecordFailure(backendID string, backendName string, stat
...
+		state.circuitBreaker.RecordFailure()
+		cbState := state.circuitBreaker.GetState()
+
+		// Log state transitions
+		switch {
+		case previousCBState == CircuitClosed && cbState == CircuitOpen:
+			logger.Warnf("Circuit breaker for backend %s OPENED (threshold exceeded)", backendName)
+		case previousCBState == CircuitHalfOpen && cbState == CircuitOpen:
+			logger.Warnf("Circuit breaker for backend %s returned to OPEN from half-open (recovery failed)", backendName)
+		}
```
jerm-dro: This logic seems like it should be entirely pushed down within `circuitBreaker.RecordFailure`. During `RecordFailure`, why not log if a transition has occurred?

pkg/vmcp/health/status.go
```diff
@@ -144,6 +200,21 @@ func (t *statusTracker) RecordSuccess(backendID string, backendName string, stat
...
+		state.circuitBreaker.RecordSuccess()
+		cbState := state.circuitBreaker.GetState()
+
+		// Log important state transitions
+		switch {
+		case previousCBState == CircuitHalfOpen && cbState == CircuitClosed:
+			logger.Infof("Circuit breaker for backend %s CLOSED after successful recovery from half-open", backendName)
+		case previousCBState != CircuitClosed && cbState == CircuitClosed:
+			logger.Infof("Circuit breaker for backend %s transitioned to CLOSED", backendName)
+		}
```
jerm-dro: Similar to below: let's push this transition logging responsibility into the circuit breaker.

pkg/vmcp/health/monitor.go
```diff
@@ -339,6 +398,31 @@ func (m *Monitor) monitorBackend(ctx context.Context, backend *vmcp.Backend, isI
...
+				// This should not happen - circuit is closed but CanAttemptHealthCheck returned false
+				logger.Debugf("Circuit breaker state inconsistency for backend %s, skipping health check", backend.Name)
+			}
+			return
+		}
+
+		// If we reach here with a half-open circuit, we're attempting the recovery test
+		if cbState, exists := m.statusTracker.GetCircuitBreakerState(backend.ID); exists && cbState == CircuitHalfOpen {
+			logger.Debugf("Circuit breaker testing recovery for backend %s", backend.Name)
+		}
```
jerm-dro: It seems like the circuit breaker / status tracker implementation is a leaky abstraction. Callers of the circuit breaker have to know a lot about the internal state / implementation of the circuit breaker to know what to do. Similar to other comments, this code here seems like it could be replaced with one `TryCheckHealth` method on the status tracker. 

Then, the `Monitor` won't have to reason about what exactly circuit breaker states mean.

pkg/vmcp/health/monitor.go
```diff
@@ -183,10 +216,26 @@ func (m *Monitor) Start(ctx context.Context) error {
...
+			m.circuitBreakerConfig.FailureThreshold, m.circuitBreakerConfig.Timeout)
+	}
+
 	// Start health check goroutine for each backend
 	m.backendsMu.Lock()
 	for i := range m.backends {
 		backend := &m.backends[i] // Capture backend pointer for this iteration
+
+		// Initialize circuit breaker if enabled
+		if m.circuitBreakerConfig != nil && m.circuitBreakerConfig.Enabled {
```
jerm-dro: In monitor.go, I notice the pattern `if m.circuitBreakerConfig != nil && m.circuitBreakerConfig.Enabled {`  then `InitializeCircuitBreaker`. This seems like the monitor knows too much about how the status tracker's state. 

Suggestion: Teach the `statusTracker` to lazily initialize the circuit breaker.

In `monitorBackend`, during the proposed call to `TryCheckHealth` we could initialize the circuit breaker. Then, this `Monitor` struct won't need to keep track of the config at all. It just passes the config to the `statusTracker` when it is constructed, and the `statusTracker` is responsible for constructing circuit breakers according to the config.


WDYT?

pkg/vmcp/health/circuit_breaker.go
```diff
@@ -0,0 +1,161 @@
...
+	CircuitOpen CircuitState = "open"
+	// CircuitHalfOpen indicates recovery testing - limited requests allowed
+	CircuitHalfOpen CircuitState = "half_open"
+)
+
+// CircuitBreaker manages circuit breaker state for a single backend.
+// It implements the circuit breaker pattern to prevent cascading failures
+// by tracking failures and transitioning through states:
+// Closed → Open → HalfOpen → Closed
+type CircuitBreaker struct {
```
jerm-dro: Does this `CircuitBreaker` need to be public? It's only referenced within this package.

pkg/vmcp/health/circuit_breaker.go
```diff
@@ -0,0 +1,161 @@
...
+	mu sync.RWMutex
```
jerm-dro: Suggestion: use a plain `sync.Mutex` unless you have a strong reason to use `RWMutex`. 

The performance characteristics of `sync.Mutex` vs `sync.RWMutex` are actually very nuanced [1]. I prefer to default to `sync.Mutex` because the performance is good enough and the simpler locking (i.e. not deciding between read / write) means fewer bugs.


[1] https://zephyrtronium.github.io/articles/rwmutex.html

jerm-dro (changes requested):
I'd like to simplify the interfaces between the Monitor, statusTracker, and circuitbreaker before merging this code.

pkg/vmcp/health/circuit_breaker.go
```diff
@@ -20,12 +22,15 @@ const (
...
 
-// CircuitBreaker manages circuit breaker state for a single backend.
+// circuitBreaker manages circuit breaker state for a single backend.
 // It implements the circuit breaker pattern to prevent cascading failures
 // by tracking failures and transitioning through states:
 // Closed → Open → HalfOpen → Closed
-type CircuitBreaker struct {
-	mu sync.RWMutex
+type circuitBreaker struct {
+	mu sync.Mutex
```
jerm-dro: nit, feel free to ignore: When adding mutexes to classes, I prefer the "mutex hat" pattern, where the mutex is directly above the fields it protects. Fields above the mutex, in this case, name don't require holding the mutex to access the state safely.

https://dmitri.shuralyov.com/idiomatic-go/entries/2

pkg/vmcp/health/monitor.go
```diff
@@ -216,25 +218,14 @@ func (m *Monitor) Start(ctx context.Context) error {
...
 
-		// Initialize circuit breaker if enabled
-		if m.circuitBreakerConfig != nil && m.circuitBreakerConfig.Enabled {
-			m.statusTracker.InitializeCircuitBreaker(
-				backend.ID,
-				m.circuitBreakerConfig.FailureThreshold,
-				m.circuitBreakerConfig.Timeout,
-			)
-		}
+		// Circuit breaker will be lazily initialized on first health check
```
jerm-dro: nit: these straggling comments reference code that has been deleted. Let's remove these comments.

pkg/vmcp/health/status.go
```diff
@@ -257,6 +260,12 @@ func (t *statusTracker) RecordFailure(backendID string, backendName string, stat
 			logger.Warnf("Backend %s initialized with failure (1/%d failures, status: %s): %v",
 				backendName, t.unhealthyThreshold, vmcp.BackendUnknown, err)
 		}
+
+		// Lazily initialize circuit breaker for new state
+		t.ensureCircuitBreaker(backendID, backendName)
+		if state.circuitBreaker != nil {
+			state.circuitBreaker.RecordFailure()
+		}
```
jerm-dro: I found this code a bit more confusing than it needs to be:
1. the `state` is inserted into `t.states` on line 252 without a circuit breaker.
2. the same `state` is mutated by `ensureCircuitBreaker` after pulling the data out of the map and adding the circuit breaker.


I think we can simplify this code and ensure the circuit breaker is constructed if it's needed when the `backendHealthState` is constructed:


```go
state = &backendHealthState{
	status:              vmcp.BackendUnknown,
	consecutiveFailures: 1,
	lastCheckTime:       time.Now(),
	lastError:           err,
	lastTransitionTime:  time.Now(),
    circuitBreaker: func() *circuitBreaker() {
        if t.circuitBreakerConfig == nil || !t.circuitBreakerConfig.Enabled {
        	return nil
    	}

      return newCircuitBreaker(
      	t.circuitBreakerConfig.FailureThreshold,
      	t.circuitBreakerConfig.Timeout,
      	backendName,
      )
  }()
}
```
This is nice, because it means `backendHealthState` is fully constructed when it's initialized. There's no additional steps for plugging in the circuitbreaker.

pkg/vmcp/health/status.go
```diff
@@ -373,33 +373,45 @@ func (t *statusTracker) ClearRemovedFlag(backendID string) {
...
 //
 // Parameters:
 //   - backendID: Unique identifier for the backend
-//   - failureThreshold: Number of failures before opening the circuit
-//   - timeout: Duration to wait in open state before attempting recovery
-func (t *statusTracker) InitializeCircuitBreaker(backendID string, failureThreshold int, timeout time.Duration) {
+//   - backendName: Human-readable name for logging purposes
+//   - failureThreshold: Number of failures before opening the circuit (ignored, uses config)
+//   - timeout: Duration to wait in open state before attempting recovery (ignored, uses config)
+func (t *statusTracker) InitializeCircuitBreaker(backendID, backendName string, _ int, _ time.Duration) {
```
jerm-dro: This is only called in tests. Can we delete it?

pkg/vmcp/health/status.go
```diff
@@ -201,19 +213,10 @@ func (t *statusTracker) RecordSuccess(backendID string, backendName string, stat
 		state.lastTransitionTime = time.Now()
 	}
 
-	// Update circuit breaker
+	// Lazily initialize and update circuit breaker (logging handled internally)
+	t.ensureCircuitBreaker(backendID, backendName)
```
jerm-dro: I don't think this is necessary. If `exists`, then `state` is guaranteed to have a circuitBreaker if it needs one.

pkg/vmcp/health/status.go
```diff
@@ -171,6 +177,12 @@ func (t *statusTracker) RecordSuccess(backendID string, backendName string, stat
 		}
 		t.states[backendID] = state
 		logger.Debugf("Backend %s initialized as %s", backendName, status)
+
+		// Lazily initialize circuit breaker for new state
+		t.ensureCircuitBreaker(backendID, backendName)
+		if state.circuitBreaker != nil {
+			state.circuitBreaker.RecordSuccess()
+		}
```
jerm-dro: Similar comment to below, I think we move the creation of the circuitBreaker up into the construction of the `backendHealthState`.

pkg/vmcp/health/status.go
jerm-dro: When you do this, can you deduplicate this piece of logic between the RecordSuccess and RecordFailure:

```go
state, exists := t.states[backendID]
	if !exists {
		// Initialize new state
		state = &backendHealthState{
			status:              vmcp.BackendUnknown,
			consecutiveFailures: 1,
			lastCheckTime:       time.Now(),
			lastError:           err,
			lastTransitionTime:  time.Now(),
		}
		t.states[backendID] = state
```

I imagine it would just become `state, existed := t.getState(backendID)`.

pkg/vmcp/health/status.go
```diff
@@ -286,19 +295,10 @@ func (t *statusTracker) RecordFailure(backendID string, backendName string, stat
 			backendName, state.consecutiveFailures, t.unhealthyThreshold, state.status, status, err)
 	}
 
-	// Update circuit breaker
+	// Lazily initialize and update circuit breaker (logging handled internally)
+	t.ensureCircuitBreaker(backendID, backendName)
```
jerm-dro: Similar to the RecordSuccess implementation, the circuitBreaker is guaranteed to exist by now.

pkg/vmcp/health/status.go
```diff
@@ -72,16 +83,67 @@ func (t *statusTracker) isRemoved(backendID string) bool {
...
 		LastError:           state.lastError,
 		LastTransitionTime:  state.lastTransitionTime,
 	}
+
+	// Include circuit breaker state if enabled
+	if state.circuitBreaker != nil {
+		snapshot := state.circuitBreaker.GetSnapshot()
+		result.CircuitState = snapshot.State
+		result.CircuitLastChanged = snapshot.LastStateChange
+	}
```
jerm-dro: Suggestion, feel free to ignore: A lot of code within `statusTracker` needs to check if the `circuitBreaker` is nil. I wonder if the circuitBreaker was an interface if it'd make this simpler. Then, we could have a `alwaysClosedCircuit`.

pkg/vmcp/health/status.go
```diff
@@ -72,16 +83,102 @@ func (t *statusTracker) isRemoved(backendID string) bool {
...
 		Status:              state.status,
 		ConsecutiveFailures: state.consecutiveFailures,
 		LastCheckTime:       state.lastCheckTime,
+		LastErrorCategory:   sanitizeError(state.lastError),
 		LastError:           state.lastError,
 		LastTransitionTime:  state.lastTransitionTime,
 	}
+
+	// Include circuit breaker state
+	snapshot := state.circuitBreaker.GetSnapshot()
```
amirejaz, line 177: nit: this state may not be the latest, as `CanAttempt` may change the state I am wondering if we should also add the same logic in `GetSnapshot`?

amirejaz (dismissed):

PR #3530, opened 2026-01-30
Integrate circuit breaker with vMCP routing and aggregation

yrobla:
Complete the circuit breaker integration by wiring it through the Virtual MCP Server stack for automatic backend failure handling.

Configuration wiring (cmd/vmcp/app/commands.go):
  - Wire circuit breaker config from YAML to health monitor
  - Pass configurable failure threshold and timeout to monitor
  - Log circuit breaker enablement on server startup

Capability filtering (pkg/vmcp/aggregator/default_aggregator.go):
  - Filter tools, resources, and prompts from unhealthy backends
  - Skip capabilities when backend status is Unhealthy, Unknown, or Unauthenticated
  - Keep capabilities from Healthy and Degraded backends (degraded backends are still operational)
  - Add isBackendHealthy helper for status evaluation

Routing protection (pkg/vmcp/router/default_router.go):
  - Check backend health before routing requests
  - Return ErrBackendUnavailable when backend is unhealthy
  - Add isTargetHealthy helper for status evaluation
  - Log warnings when routing fails due to backend unavailability

Testing (pkg/vmcp/router/default_router_test.go):
  - Add HealthStatus to all test fixtures
  - Add test cases for unhealthy backend routing failures
  - Add test cases for unauthenticated backend failures
  - Add test cases for degraded backend success (still operational)
  - Cover all three routing methods: tools, resources, prompts

Related-to: #3036

### DIFF
--- a/pkg/vmcp/aggregator/default_aggregator.go
+++ b/pkg/vmcp/aggregator/default_aggregator.go
@@ -157,7 +157,6 @@ func (a *defaultAggregator) QueryAllCapabilities(
 
 	// Query each backend in parallel
 	for _, backend := range backends {
-		backend := backend // Capture loop variable
 		g.Go(func() error {
 			caps, err := a.QueryCapabilities(ctx, backend)
 			if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/vmcp/aggregator/default_aggregator.go
```diff
@@ -302,65 +302,83 @@ func (a *defaultAggregator) MergeCapabilities(
...
-		} else {
-			target := vmcp.BackendToTarget(backend)
-			// Store the original resource URI for forwarding to backend
-			target.OriginalCapabilityName = resource.URI
-			routingTable.Resources[resource.URI] = target
+			continue
 		}
+
+		// Filter out resources from unhealthy backends
+		if !backend.HealthStatus.IsHealthyForRouting() {
```
jerm-dro, line 364: suggestion: In the event health changes, checking this multiple times for the same backend while aggregating can cause partial states where the tools for the backend are excluded but the resources are included. This is probably not a big deal, but you could prevent it by only checking a backend's health once per aggregation.

jerm-dro (commented):
I think the comment here: https://github.com/stacklok/toolhive/pull/3533#discussion_r2748136614 is also relevant to this PR, which adds more complexity to the `defaultAggregator`.

yrobla:
closing to start a new one, it diverged too much so better to start clean

PR #3531, opened 2026-01-30
Add OAuth authorization server integration tests

jhrozek:
Add integration test for the full OAuth 2.0 Authorization Code flow with PKCE using mockoidc as the upstream IDP.

TestIntegration_FullPKCEFlow validates the complete OAuth flow including authorization through the upstream IDP, client state preservation, token exchange with PKCE verification, and refresh token issuance via the offline_access scope. The test also verifies RFC 8707 resource parameter handling for audience binding and comprehensive JWT claims (iss, sub, aud, iat, exp, scp).

### DIFF
--- a/pkg/authserver/server_impl.go
+++ b/pkg/authserver/server_impl.go
@@ -205,7 +205,7 @@ func createProvider(authServerConfig *oauthserver.AuthorizationServerConfig, sto
 	// - Storage: where to persist authorization codes, tokens, and client data
 	// - Strategy: how to generate and validate tokens
 	// - Factories: which OAuth grant types to enable (each adds handlers for specific flows)
-	provider := compose.Compose(
+	return compose.Compose(
 		authServerConfig.Config,
 		stor,
 		&compose.CommonStrategy{CoreStrategy: jwtStrategy},
@@ -213,6 +213,4 @@ func createProvider(authServerConfig *oauthserver.AuthorizationServerConfig, sto
 		compose.OAuth2RefreshTokenGrantFactory, // Refresh token grant
 		compose.OAuth2PKCEFactory,              // PKCE for public clients
 	)
-
-	return provider
 }


pkg/authserver/server_impl.go
```diff
@@ -205,12 +205,14 @@ func createProvider(authServerConfig *oauthserver.AuthorizationServerConfig, sto
 	// - Storage: where to persist authorization codes, tokens, and client data
 	// - Strategy: how to generate and validate tokens
 	// - Factories: which OAuth grant types to enable (each adds handlers for specific flows)
-	return compose.Compose(
+	provider := compose.Compose(
```
yrobla: you can just return the composer.Compose() without adding a var here

pkg/authserver/integration_test.go
```diff
@@ -0,0 +1,487 @@
...
+	require.True(t, ok, "exp claim should be a number")
+	expectedExp := iat + testAccessTokenLifetime.Seconds()
+	assert.InDelta(t, expectedExp, exp, 2, "exp should be iat + configured token lifetime (within 2s tolerance)")
+
+	// Verify scope claim matches requested scopes
+	scope, ok := claims["scp"].([]interface{})
+	require.True(t, ok, "scp claim should be an array")
+	scopeStrings := make([]string, len(scope))
+	for i, s := range scope {
+		scopeStrings[i] = s.(string)
```
yrobla: you should validate the type here

pkg/authserver/integration_test.go
```diff
@@ -0,0 +1,487 @@
...
+		"redirect_uri":  {testRedirectURI},
+		"client_id":     {testClientID},
+		"code_verifier": {verifier},
+	}
+	if resource != "" {
+		params.Set("resource", resource)
+	}
+
+	tokenResp := makeTokenRequest(t, serverURL, params)
+	defer tokenResp.Body.Close()
```
yrobla, line 385: we are closing the body twice... here and on parsetokenresponse

PR #3533, opened 2026-01-30
Add partial failure mode support to vMCP aggregator

yrobla:
Implements configurable failure handling when backend MCP servers become unavailable. The aggregator now supports two modes:

  - fail (default): Returns error if all backends unavailable; logs warning if some backends are down but continues with healthy ones
  - best_effort: Continues with available backends regardless of failures; logs info about unavailable backends

Related-to: #3036

### DIFF
--- a/pkg/vmcp/aggregator/default_aggregator.go
+++ b/pkg/vmcp/aggregator/default_aggregator.go
@@ -26,17 +26,20 @@ type defaultAggregator struct {
 	conflictResolver ConflictResolver
 	toolConfigMap    map[string]*config.WorkloadToolConfig // Maps backend ID to tool config
 	tracer           trace.Tracer
+	failureMode      string // "fail" or "best_effort"
 }
 
 // NewDefaultAggregator creates a new default aggregator implementation.
 // conflictResolver handles tool name conflicts across backends.
 // workloadConfigs specifies per-backend tool filtering and overrides.
 // tracerProvider is used to create a tracer for distributed tracing (pass nil for no tracing).
+// failureMode determines behavior when backends are unavailable ("fail" or "best_effort").
 func NewDefaultAggregator(
 	backendClient vmcp.BackendClient,
 	conflictResolver ConflictResolver,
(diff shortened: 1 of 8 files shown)


pkg/vmcp/aggregator/default_aggregator.go
```diff
@@ -310,6 +323,13 @@ func (a *defaultAggregator) MergeCapabilities(
 			continue
 		}
 
+		// Track backend health
+		if backend.HealthStatus.IsHealthyForRouting() {
+			healthyBackends[backend.ID] = true
+		} else {
+			unhealthyBackends[backend.ID] = string(backend.HealthStatus)
+		}
```
jerm-dro, line 331: Suggestion: The factoring of the `defaultAggregator` leaves a bit to be desired and I think this change motivates improving it. 

### Background

The `defaultAggregator` implements the `Aggregator` interface, which has 4 unused public methods: `QueryCapabilities`, `QueryAllCapabilities`, `ResolveConflicts`, and `MergeCapabilities`.

### Suggestion

Rather than inlining more logic into `defaultAggregator`, lets:
1. Remove the unused methods from the interface.
2. Implement a `circuitBreakerAggregator` for your new logic. This implements the new and improved `Aggregator` interface and decorates any arbitrary aggregator:

```go 
type circuitBreakerAggregator struct {
   // this would really just be the defaultAggregator
    inner Aggregator
}


func (c circuitBreakerAggregator) AggregateCapabilities(...) {
    // check all the backends for health
    err := c.enforceFailureMode(...)
    if err != nil {
        return nil, err
    }
    return inner.AggregateCapabilities(...)
}
```

I think this makes your change cleaner, because it doesn't add any complexity to the existing `defaultAggregator`'s construction or runtime behavior. It would also enable pretty straightforward unit testing of the circuitBreakerAggregator, because you don't need the whole `defaultAggregator` to test it.

What do you think?

yrobla:
reimplementing as there were so many changes, makes sense to start from scratch

PR #3543, opened 2026-02-01
Fix zombie process when proxy stops due to health check failure

### LINKED ISSUE #3542: Zombie process when TransparentProxy stops due to health check failure
# Zombie process when TransparentProxy stops due to health check failure

## Description

When health checks fail for a remote MCP server using streamable-http transport, the proxy listener shuts down but the thv process does NOT exit. This leaves a "zombie" process that:

- Still shows as "running" in `thv list`
- Has no TCP listener (port is released)
- Continues running telemetry goroutines
- Requires manual intervention (`thv rm` + restart) to recover

This blocks the automatic restart/recovery mechanism from working.

**Note:** This bug primarily affects remote MCP servers with health checks enabled (`TOOLHIVE_REMOTE_HEALTHCHECKS=true`). For local containers, container exit is detected via a different path that doesn't have this issue. However, the fix is general and ensures `HTTPTransport.IsRunning()` correctly reflects proxy state in all scenarios.

## Root Cause

`HTTPTransport.IsRunning()` only checks its own `shutdownCh`, not the proxy's:

```go
// pkg/transport/http.go
func (t *HTTPTransport) IsRunning(_ context.Context) (bool, error) {
    select {
    case <-t.shutdownCh:
        return false, nil
    default:
        return true, nil  // BUG: Only checks its OWN shutdownCh
    }
}
```

When health checks fail:
- `TransparentProxy.handleHealthCheckFailure()` calls `p.Stop(ctx)`
- This closes the **proxy's** `shutdownCh`
- But `HTTPTransport` has a **separate** `shutdownCh` that remains open
- The runner's status monitor polls `transport.IsRunning()`, which keeps returning `true`
- The process never exits → zombie

## Steps to Reproduce

- Save the mock server code below as `mock-mcp-server.go` and build it:
```bash
go build -o mock-mcp-server mock-mcp-server.go
./mock-mcp-server --port 19999 --fail-after 15 &
```

<details>
<summary>mock-mcp-server.go (click to expand)</summary>

```go
package main

import (
	"flag"
	"fmt"
	"net/http"
	"os"
	"sync/atomic"
	"time"
)

var requestCount atomic.Int64

func main() {
	port := flag.Int("port", 19999, "Port to listen on")
	failAfter := flag.Int("fail-after", 0, "Start failing after N requests")
	flag.Parse()

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		count := requestCount.Add(1)

		if *failAfter > 0 && int(count) > *failAfter {
			w.WriteHeader(http.StatusInternalServerError)
			return
		}

		if r.URL.Path == "/sse" || r.URL.Path == "/mcp" {
			w.Header().Set("Content-Type", "text/event-stream")
			w.Header().Set("Cache-Control", "no-cache")
			w.WriteHeader(http.StatusOK)
			w.Write([]byte("event: endpoint\ndata: http://localhost/messages\n\n"))
			if f, ok := w.(http.Flusher); ok {
				f.Flush()
			}
			return
		}
		w.WriteHeader(http.StatusOK)
	})

	fmt.Fprintf(os.Stderr, "Mock server on port %d, failing after %d requests\n", *port, *failAfter)
	server := &http.Server{
		Addr:         fmt.Sprintf("127.0.0.1:%d", *port),
		ReadTimeout:  10 * time.Second,
		WriteTimeout: 10 * time.Second,
	}
	server.ListenAndServe()
}
```

</details>

- Start thv with health checks enabled:
```bash
export TOOLHIVE_REMOTE_HEALTHCHECKS=true
thv run "http://127.0.0.1:19999/mcp" --name health-test
```

- Wait ~2 minutes for initialization and health checks to fail (3 consecutive failures)

- Check the logs and process state:
```bash
grep -i "health" ~/Library/Application\ Support/toolhive/logs/health-test.log
# Should show: "Health check failed... (attempt 3/3)"
# Should show: "initiating proxy shutdown"

pgrep -f "thv start health-test --foreground"
# BUG: Returns a PID - process is still running!

nc -z 127.0.0.1 <PORT>
# Connection refused - port not listening, but process alive = ZOMBIE
```

## Evidence from Logs

**Before fix (zombie behavior):**
```
4:29PM  WARN  Health check failed for http://127.0.0.1:19999 (attempt 3/3): status=degraded
4:29PM  ERROR Health check failed for http://127.0.0.1:19999 after 3 consecutive attempts; initiating proxy shutdown
4:29PM  WARN  Health check failed for remote server health-test, marking as unhealthy
# ... no further output, process hangs as zombie
```

**After fix (self-healing behavior):**
```
4:34PM  WARN  Health check failed for http://127.0.0.1:19999 (attempt 3/3): status=degraded
4:34PM  ERROR Health check failed for http://127.0.0.1:19999 after 3 consecutive attempts; initiating proxy shutdown
4:34PM  WARN  Health check failed for remote server health-test, marking as unhealthy
4:34PM  WARN  Transport is no longer running, attempting automatic restart...
4:34PM  INFO  MCP server health-test stopped, restart needed
4:34PM  INFO  MCP server health-test started successfully
```

## Expected Behavior

Process exits after health check failure, triggering the restart mechanism.

## Proposed Fix

Update `HTTPTransport.IsRunning()` to also check `proxy.IsRunning()`:

```go
func (t *HTTPTransport) IsRunning(ctx context.Context) (bool, error) {
    t.mutex.Lock()
    defer t.mutex.Unlock()

    select {
    case <-t.shutdownCh:
        return false, nil
    default:
        if t.proxy != nil {
            proxyRunning, err := t.proxy.IsRunning(ctx)
            if err != nil {
                return false, err
            }
            if !proxyRunning {
                return false, nil
            }
        }
        return true, nil
    }
}
```

This requires adding `IsRunning()` to the `Proxy` interface. `TransparentProxy` already has this method; `HTTPSSEProxy` and `HTTPProxy` (streamable) need implementations added.

The fix includes unit tests and an e2e test (`test/e2e/health_check_zombie_test.go`) that verifies the runner detects when the proxy has stopped.

## Environment

- ToolHive version: v0.8.2
- OS: macOS (likely affects Linux as well, but untested)
- Affected scenarios: Remote MCP servers with `TOOLHIVE_REMOTE_HEALTHCHECKS=true`

## Related Issues

- #3429 - Remote MCP servers leave zombie processes after laptop sleep, blocking restart (same symptom, may be same root cause)
- #3383 - stdio: transport enters zombie state when re-attachment fails (similar pattern for stdio transport)

gkatz2:
## Summary

- Fix zombie process issue where the ToolHive proxy process continues running after health check failure triggers proxy shutdown
- Add `IsRunning()` method to the `Proxy` interface
- Update `HTTPTransport.IsRunning()` to check proxy state in addition to its own shutdown channel
- Add unit tests and e2e test for the fix

## Problem

When health checks fail 3 consecutive times, `TransparentProxy.Stop()` is called which closes the proxy's shutdown channel. However, `HTTPTransport.IsRunning()` only checks its own shutdown channel, not the proxy's. This causes the runner's status monitor to keep returning `true`, preventing the process from exiting and blocking the self-healing restart mechanism.

## Solution

`HTTPTransport.IsRunning()` now also calls `proxy.IsRunning()` to detect when the proxy has stopped itself due to health check failure. This enables the runner to detect the stopped state and trigger automatic restart.

## Test plan

- [x] Unit tests for `IsRunning()` behavior (`pkg/transport/http_test.go`)
- [x] E2e test verifying runner detects proxy shutdown (`test/e2e/health_check_zombie_test.go`)
- [x] Manual testing confirmed fix enables self-healing restart

Fixes #3542

### DIFF
--- a/pkg/transport/http.go
+++ b/pkg/transport/http.go
@@ -431,15 +431,14 @@ func (t *HTTPTransport) IsRunning() (bool, error) {
 		// Also check if the proxy is still running (handles health check failure case)
 		// When health checks fail, the proxy stops itself but the transport's
 		// shutdownCh may not be closed, causing the runner to hang as a zombie process.
+		proxyRunning := true
+		var err error
 		if t.proxy != nil {
-			proxyRunning, err := t.proxy.IsRunning()
+			proxyRunning, err = t.proxy.IsRunning()
 			if err != nil {
 				return false, err
 			}
-			if !proxyRunning {
-				return false, nil
-			}
 		}
-		return true, nil
+		return proxyRunning, nil
 	}
 }


pkg/transport/types/transport.go
```diff
@@ -171,6 +171,11 @@ type Proxy interface {
 	// Stop stops the proxy.
 	Stop(ctx context.Context) error
 
+	// IsRunning checks if the proxy is currently running.
+	// This is used by HTTPTransport to detect when the proxy has stopped
+	// (e.g., due to health check failure) even if the transport itself hasn't been stopped.
+	IsRunning(ctx context.Context) (bool, error)
```
dmjb: Does the context ever get used here? The http and streamable proxy do not use it, and the code in http.go just passes the context onto the proxy it calls.

pkg/transport/http.go
```diff
@@ -425,6 +428,18 @@ func (t *HTTPTransport) IsRunning(_ context.Context) (bool, error) {
...
+		if t.proxy != nil {
+			proxyRunning, err := t.proxy.IsRunning(ctx)
+			if err != nil {
+				return false, err
+			}
+			if !proxyRunning {
+				return false, nil
+			}
+		}
 		return true, nil
```
dmjb:
```suggestion
		proxyRunning := true
		if t.proxy != nil {
			proxyRunning, err := t.proxy.IsRunning(ctx)
			if err != nil {
				return false, err
			}
		}
		return proxyRunning, nil
```

dmjb:
@gkatz2 Thanks for this change. If you can address the pieces of feedback I have here, I'm happy to merge.

dmjb:
Looks like you may also need to run `task gen` locally to re-update some mocks.

gkatz2:
Thanks, @dmjb. Just pushed a commit for the mock generation issue. Looking at the code suggestion now.

gkatz2:
**Re: Context parameter in `IsRunning`**

You're right - the context isn't used by any of the implementations. I added it for consistency with `Start`/`Stop`, but since `IsRunning` is just checking a channel state (instant, non-blocking), context serves no purpose here. Will remove it.

**Re: Code restructure**

I think the suggestion has an issue - using `:=` inside the `if` block creates a new `proxyRunning` that shadows the outer one, meaning that although the code appears to be using the same instance of the `proxyRunning` variable throughout, there are actually two. The bug manifests when the proxy has stopped (e.g., due to health check failure) - `IsRunning()` returns `false, nil`, the inner `proxyRunning` is set to `false`, but when the `if` block exits, the outer `proxyRunning` (still `true`) is returned.

Here are two corrected versions:

**Option A** - declare at top, single return:
```go
proxyRunning := true
var err error
if t.proxy != nil {
    proxyRunning, err = t.proxy.IsRunning(ctx)
    if err != nil {
        return false, err
    }
}
return proxyRunning, nil
```

**Option B** - early return:
```go
if t.proxy != nil {
    proxyRunning, err := t.proxy.IsRunning(ctx)
    if err != nil {
        return false, err
    }
    return proxyRunning, nil
}
return true, nil
```

Let me know which you prefer and I'll update!

gkatz2:
@dmjb - I wanted to get your thoughts on one aspect of this PR.

The latest commit adds a `TOOLHIVE_HEALTH_CHECK_INTERVAL` environment variable to the transparent proxy. This allows the E2E test to use a 1-second health check interval instead of the default 10 seconds, which reduces test runtime from 30+ seconds to ~12 seconds.

I realize this expands the scope a bit - adding production code to support faster E2E testing. A few options:

1. **Keep it as-is**: Being able to control the health check interval could be valuable for production use cases (e.g., adjusting for slow/unreliable backends, or tightening intervals for faster failure detection)
2. **Remove it**: I can revert that part and let the E2E test run slower (~30-35s). The test would still be reliable with the other fixes (status-based detection instead of log parsing)

Happy to go either direction based on your preference. What do you think?

dmjb:
@gkatz2 I think the interval env var is a good idea, let's keep it 👍

dmjb:
@gkatz2 Thanks for addressing the feedback. I'll merge when the build goes green.

gkatz2:
@dmjb Implemented the code restructure per your suggestion. I added `var err error` since using `=` instead of `:=` requires the variable to be pre-declared.

```go
proxyRunning := true
var err error
if t.proxy != nil {
    proxyRunning, err = t.proxy.IsRunning()
    if err != nil {
        return false, err
    }
}
return proxyRunning, nil
```

CI is passing.

**Update**: Oops, didn't see you already merged it! Thanks.

PR #3550, opened 2026-02-02
Update the arch docs for the registry

rdimitrov:
The following PR bumps the architecture docs to reflect the latest state of the toolhive-registry-server.

### DIFF
--- a/docs/arch/00-overview.md
+++ b/docs/arch/00-overview.md
@@ -110,9 +110,9 @@ A specialized binary used by the Kubernetes operator. Located in `cmd/thv-proxyr
 Deployment (proxy-runner) -> StatefulSet (MCP server)
 ```
 
-### 4. Registry Server (External Project)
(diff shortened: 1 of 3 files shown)


docs/arch/00-overview.md
```diff
@@ -110,17 +110,17 @@ A specialized binary used by the Kubernetes operator. Located in `cmd/thv-proxyr
 Deployment (proxy-runner) -> StatefulSet (MCP server)
 ```
 
-### 4. Registry API Server (thv-registry-api)
+### 4. Registry Server (External Project)
```
JAORMX: while it's a separate repo. It's still part of the "toolhive platform" umbrella. Should we not add the "External project" tag here?

rdimitrov: That's a good point! 👍

JAORMX (approved):

Issue #2192, opened 2025-10-14
FR: Add support for Gemini CLI client


samuv:
Requested form a user in toolhive studio repo [here](https://github.com/stacklok/toolhive-studio/issues/1035)

Gemini CLI: https://github.com/google-gemini/gemini-cli

---

## Contributor Guidance

This is a good first issue! Here's some context to help you get started:

### Where to look
The client integration system is in `pkg/client/config.go`. Look at how other CLI clients are defined - there are 20+ existing examples to follow.

### What's involved
1. **Add a new constant** around line 70 in the `MCPClient` constants block:
   ```go
   GeminiCli MCPClient = "gemini-cli"
   ```

2. **Add a configuration entry** to the `supportedClientIntegrations` slice (starts around line 115). Use Claude Code CLI as a reference pattern (lines 222-235).

3. **Update tests** in `pkg/client/config_test.go` to include the new client.

### Research needed
Before implementing, you'll need to find out from the [Gemini CLI repository](https://github.com/google-gemini/gemini-cli):
- Where does Gemini CLI store its config file? (e.g., `~/.gemini/` or `~/.config/gemini/`)
- What is the config file name and format? (JSON or YAML)
- How are MCP servers configured in that file?

### Reference examples
Look at these existing client definitions for patterns to follow:
- **Claude Code CLI** (lines 222-235) - another CLI tool, good template
- **VS Code** (lines 237-258) - shows platform-specific paths
- **Cursor** (lines 260-280) - similar structure

### How to test
```bash
task build
task test
# After adding, verify with:
./bin/thv client list
```

### Useful commands
- `task build` - Build the binary
- `task test` - Run unit tests
- `task lint` - Run linter

PR #3552, opened 2026-02-02
Add Google Gemini CLI client support with dynamic URL field mapping

aponcedeleonch:
Closes: https://github.com/stacklok/toolhive/issues/2192

Add support for Google Gemini CLI as a registered MCP client. For Gemini CLI:
- SSE transport uses the "url" field
- Streamable HTTP transport uses the "httpUrl" field

The implementation introduces:
- New MCPServersUrlLabelMap (map[TransportType]string). Transport type to URL
- Update all client configurations to specify URL field per transport type
- `buildMCPServer` helper function to construct MCPServer structs with appropriate field names
- Support for "httpUrl" field in MCPServer struct
- Add Uri field support for clients like Goose that use "uri" instead of "url"
- Comprehensive tests covering URL field derivation logic and fallback behavior

### DIFF
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -779,7 +779,8 @@ func (cm *ClientManager) Upsert(cf ConfigFile, name string, url string, transpor
 }
 
 // buildMCPServer constructs an MCPServer struct with the appropriate URL field and optional type field.
-// The URL field name is always derived from the transport type using MCPServersUrlLabelMap.
+// The URL field name is determined by looking up the transport type in MCPServersUrlLabelMap.
+// If the map is nil or the transport type is not found, it falls back to "url" as the default.
 // For most clients, all transport types map to the same URL field (e.g., "url"), but some clients
 // like Gemini CLI use different URL fields per transport type (e.g., "url" for SSE, "httpUrl" for streamable HTTP).
 func buildMCPServer(url, transportType string, clientCfg *mcpClientConfig) MCPServer {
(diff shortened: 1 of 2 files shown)


pkg/client/config.go
```diff
@@ -497,6 +505,27 @@ var supportedClientIntegrations = []mcpClientConfig{
...
+		// The SupportedTransportTypesMap values represent URL field names when
+		// DeriveUrlFieldFromTransport is true.
+		SupportedTransportTypesMap: map[types.TransportType]string{
+			types.TransportTypeStdio:          "httpUrl",
+			types.TransportTypeSSE:            "url",
+			types.TransportTypeStreamableHTTP: "httpUrl",
+		},
+		IsTransportTypeFieldSupported: false,
+		DeriveUrlFieldFromTransport:   true,
+		MCPServersUrlLabel:            "httpUrl", // fallback if transport type not in map
```
eleftherias: Could we repurpose `MCPServersUrlLabel` into a `MCPServersUrlLabelMap`, like we do for `SupportedTransportTypesMap`?

Then we wouldn't need `DeriveUrlFieldFromTransport` and we would always derive the url field from the transport.

For Gemini CLI it would be:
```
MCPServersUrlLabelMap: map[types.TransportType]string{
			types.TransportTypeStdio:          "httpUrl",
			types.TransportTypeSSE:            "url",
			types.TransportTypeStreamableHTTP: "httpUrl",
		}
```

and for most other clients it would be:
```
MCPServersUrlLabelMap: map[types.TransportType]string{
			types.TransportTypeStdio:          "url",
			types.TransportTypeSSE:            "url",
			types.TransportTypeStreamableHTTP: "url",
		}
```

aponcedeleonch: Yes, sounds good. Will implement

aponcedeleonch: Implemented :)

aponcedeleonch:
@claude review this code

PR #3541, opened 2026-01-30
Integrate Embedded Auth Server with Runner

tgrunnagle:
Depends on #3540 

## What

This PR integrates the embedded authorization server with the `pkg/runner/` package, enabling MCP servers to have their own OAuth2/OIDC authorization server when `EmbeddedAuthServerConfig` is configured.

Key changes:
- Add `EmbeddedAuthServerConfig` field to `RunConfig` for serializable auth server configuration
- Initialize and start the embedded auth server during `Runner.Run()` when configured
- Mount auth server routes at dedicated prefixes to avoid conflicts with MCP endpoints
- Clean up the embedded auth server during `Runner.Cleanup()`

## Why

This completes the Runner side of the Embedded Authorization Server feature ([stacklok-epics#230](https://github.com/stacklok/stacklok-epics/issues/230)). The integration allows:

1. **Per-server authorization** - Each MCP server can optionally run its own embedded authorization server that delegates authentication to upstream identity providers (Okta, Auth0, GitHub, etc.)

2. **Clean endpoint separation** - Auth server routes are mounted at specific prefixes (`/oauth/`, `/.well-known/oauth-authorization-server`, etc.) to avoid conflicts with MCP protocol endpoints like `/.well-known/oauth-protected-resource`

3. **Proper lifecycle management** - The embedded auth server is initialized during runner startup and cleaned up during shutdown, ensuring resources are properly released

## Related

- Epic: [stacklok-epics#225](https://github.com/stacklok/stacklok-epics/issues/225) (Embedded Authorization Server)
- Implements `pkg/runner/` changes for: [stacklok-epics#230](https://github.com/stacklok/stacklok-epics/issues/230)

PR #3556, opened 2026-02-02
Integration Tests for Embedded Auth Server

tgrunnagle:
Depends on #3541 

## Summary

Adds integration tests for the embedded authorization server to ensure OAuth/OIDC endpoints work correctly when integrated with the proxy runner and operate alongside MCP endpoints without conflict.

**Related Issue:** stacklok/stacklok-epics#234

## Why This Change

The embedded auth server is a critical component that enables MCP clients to authenticate using standard OAuth 2.0 flows. Without integration tests, we risk:

- Breaking OAuth/OIDC discovery endpoints that clients depend on for configuration
- Introducing routing conflicts between auth endpoints and MCP endpoints
- Resource leaks if cleanup isn't handled properly
- Regressions in RFC 8707 (resource indicators) compliance required by MCP spec

### OAuth/OIDC Compliance
- Discovery endpoints return valid metadata (JWKS, OAuth AS Metadata, OIDC Discovery)
- Authorization flow correctly redirects to upstream IDP
- RFC 8707 resource parameter is enforced (required for MCP)
- Dynamic Client Registration (RFC 7591) works correctly
- Token endpoint properly rejects invalid requests

### Runner Integration
- Auth endpoints are mounted at correct URL prefixes
- OAuth endpoints don't interfere with MCP endpoints
- Auth server handles concurrent requests safely
- Resources are cleaned up properly when runner stops

### Configuration Validation
- Missing required config (audiences, upstreams) produces clear errors
- Invalid issuer URLs are rejected
- Both ephemeral (dev mode) and file-based signing keys work

## Test Approach

Tests use a mock upstream IDP rather than real OAuth providers to:
- Enable parallel test execution without external dependencies
- Provide deterministic, reproducible results
- Allow testing edge cases and error conditions

## Files Changed

```
test/integration/authserver/
├── authserver_integration_test.go    # Auth server functionality tests
├── runner_integration_test.go        # Runner integration tests
├── README.md                         # Test documentation
└── helpers/
    ├── authserver.go                 # Auth server test utilities
    ├── http_client.go                # OAuth client for testing flows
    └── mock_upstream.go              # Mock upstream IDP
```

**+1,406 lines** across 6 new files

## Large PR Justification
- All changes are part of one integration test suite


test/integration/authserver/README.md
```diff
@@ -0,0 +1,122 @@
+# Auth Server Integration Tests
+
```
jhrozek: do we need to submit the README? This will go stale quite fast

test/integration/authserver/helpers/http_client.go
```diff
@@ -0,0 +1,174 @@
...
+		tb:      tb,
+		baseURL: baseURL,
+		httpClient: &http.Client{
+			Timeout: 10 * time.Second,
+		},
+	}
+}
+
+// GetJWKS fetches the JWKS endpoint and returns the parsed response.
+func (c *OAuthClient) GetJWKS() (map[string]interface{}, int) {
```
jhrozek: helper functions should call c.tb.Helper() otherwise if something fails here, the traceback will point to the method, not the caller

test/integration/authserver/helpers/http_client.go
```diff
@@ -0,0 +1,174 @@
...
+			CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
+				// Don't follow redirects - we want to inspect them
+				return http.ErrUseLastResponse
+			},
+		},
+	}
+}
+
+// NewOAuthClientWithRedirects creates an HTTP client that follows redirects.
+func NewOAuthClientWithRedirects(tb testing.TB, baseURL string) *OAuthClient {
```
jhrozek: I don't see this function being used anywhere?

test/integration/authserver/helpers/http_client.go
```diff
@@ -0,0 +1,174 @@
...
+
+// RegisterClient performs dynamic client registration.
+func (c *OAuthClient) RegisterClient(clientMetadata map[string]interface{}) (map[string]interface{}, int) {
+	body, err := json.Marshal(clientMetadata)
+	require.NoError(c.tb, err)
+
+	resp, err := c.httpClient.Post(
+		c.baseURL+"/oauth/register",
+		"application/json",
+		strings.NewReader(string(body)),
```
jhrozek: NewReader already accepts a []byte slice..

test/integration/authserver/runner_integration_test.go
```diff
@@ -0,0 +1,266 @@
...
+		// Make concurrent requests to various endpoints
+		var wg sync.WaitGroup
+		errors := make(chan error, 10)
+
+		for i := 0; i < 10; i++ {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				client := helpers.NewOAuthClient(t, server.URL)
+				_, statusCode := client.GetJWKS()
```
jhrozek: this helper calls require which internally calls FailNow which is not safe to call from within a goroutine: https://pkg.go.dev/testing#B.FailNow - we'd have to switch to assert instead of require but then also abort the test I guess?

test/integration/authserver/helpers/http_client.go
tgrunnagle: Removed `require` from helper functions and returned errors, so shouldn't need `c.tb.Helper()` now

PR #3257, opened 2026-01-12
Standardize proxy port field in MCPRemoteProxy

slyt3:
### **Description:**

This PR aligns MCPRemoteProxy with MCPServer by standardizing the proxy port field naming.

### Summary

- Added proxyPort to MCPRemoteProxySpec
- Deprecated the legacy port field
- Preserved backward compatibility during the transition

**Details**

- GetProxyPort() now resolves ports in the following order:
                                                        1. proxyPort
                                                        2. port
                                                        3. default (8080)

- Updated controller and deployment tests
- Added test coverage to ensure proxyPort takes precedence over port
- Verified existing resources using port continue to work

This improves API consistency across MCP resources while maintaining a smooth migration path for existing users.

Fixes #3102

PR #3559, opened 2026-02-02
chore: cache go build/tools in helm charts workflow

slyt3:
PR, as requested in #3257, it will add go build + go tools cache to the current helm-chart-test.yml workflow (the old Heml lint/test workflows were removed upstream)

@dmjb

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -454,7 +454,7 @@ func (r *Runner) Run(ctx context.Context) error {
 			}
 
 			// Check if the transport is still running
-			running, err := transportHandler.IsRunning(ctx)
+			running, err := transportHandler.IsRunning()
 			if err != nil {
 				logger.Errorf("Error checking transport status: %v", err)
 				// Don't exit immediately on error, try again after pause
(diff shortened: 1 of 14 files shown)


ChrisJBurns:
@slyt3 Thanks for this PR. Are you sure we need this? Looking at a [CI run for this action without this change](https://github.com/stacklok/toolhive/actions/runs/21679532492/job/62509575444), the steps that take the most time is the setting up of the Kind cluster and the running of the tests themselves. There doesn't seem to be any need for caching any go modules? The CI run for [this PR](https://github.com/stacklok/toolhive/actions/runs/21680209045/job/62512150147?pr=3559) seems to be a bit longer too so I'm more convinced now that we may not need this caching.

What do you think?

slyt3:
@ChrisJBurns  I reran and compared step timings;  its like now its [few seconds faster the](https://github.com/stacklok/toolhive/actions/runs/21681669620/job/62517589643?pr=3559) long steps (~1m49s and ~2m) are unchanged, and the small steps vary by only a few seconds; yea you right I dont see a clear improvement from the cache, so I’m OK to drop this PR

ChrisJBurns:
@slyt3 Yep no problems, I'll close this PR. Thanks for the idea though, we could always use more of the "how to make this better" mindset! ❤️

PR #3562, opened 2026-02-03
Add OIDC provider upstream support to authserver

jhrozek:
Implements OIDC provider support for upstream identity providers like Google, Microsoft/Entra, Okta, and Auth0.

Benefits for users:

1. Simpler configuration - just specify the issuer URL instead of individual endpoints:

```
   # Before (OAuth2 - must look up each endpoint URL):
   upstream:
     type: oauth2
     client_id: "..."
     authorization_endpoint: "https://accounts.google.com/o/oauth2/v2/auth"
     token_endpoint: "https://oauth2.googleapis.com/token"
     userinfo_endpoint: "https://openidconnect.googleapis.com/v1/userinfo"

   # After (OIDC - endpoints auto-discovered):
   upstream:
     type: oidc
     client_id: "..."
     issuer: "https://accounts.google.com"
```

The OIDC provider enables cryptographic identity verification - ID tokens are signed JWTs validated locally, no extra network call to UserInfo endpoint.

The Implementation uses go-oidc library for discovery and delegates issuer validation to it. ID token validation is stubbed for a follow-up PR.


pkg/authserver/upstream/idtoken_claims.go
```diff
@@ -1,57 +0,0 @@
...
-// See the License for the specific language governing permissions and
-// limitations under the License.
-
-package upstream
-
-import "time"
-
-// IDTokenClaims contains the standard OIDC ID Token claims.
-// See OIDC Core Section 2 for claim definitions.
-type IDTokenClaims struct {
```
jhrozek, line 21: this file and struct is no longer needed, we just go-oidc's claims struct instead. Less code.

pkg/authserver/upstream/oidc.go
```diff
@@ -0,0 +1,336 @@
...
+// from the identity provider. When enabled, the "prompt=consent" parameter is added
+// to authorization requests, forcing the user to re-consent even if they have
+// previously authorized the application.
+//
+// This is useful for:
+//   - Testing OAuth flows to verify consent screen behavior
+//   - Obtaining a new refresh token when the original has been lost or revoked
+//   - Ensuring the user explicitly confirms permissions after scope changes
+//   - Applications that require explicit user consent on every authentication
+func WithForceConsentScreen(force bool) OIDCProviderOption {
```
jhrozek, line 82: this is just an option I used for my local testing. It's not needed to ship it on the other hand, it's not exactly code that could break...

pkg/authserver/upstream/oidc.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return "", fmt.Errorf("%w: ID token validation failed: %v", ErrIdentityResolutionFailed, err)
+	}
+	return claims.Subject, nil
+}
+
+// validateIDToken validates an ID token and returns the parsed token.
+// TODO: Implement full validation using p.verifier in a follow-up PR.
+func (*OIDCProviderImpl) validateIDToken(_ context.Context, _, _ string) (*oidc.IDToken, error) {
+	// Stub - full implementation in follow-up PR
```
jhrozek, line 232: I wanted to keep the PR size small, I'll send the validation code in a follow-up

pkg/authserver/upstream/oidc.go
```diff
@@ -0,0 +1,336 @@
...
+// NewProvider() before this function is called.
+func validateDiscoveryDocument(doc *oauthproto.OIDCDiscoveryDocument, expectedIssuer string) error {
+	// Validate required OIDC fields per spec
+	if err := doc.Validate(true); err != nil {
+		return err
+	}
+
+	// Security: validate that discovered endpoints use secure schemes.
+	// This prevents a malicious discovery document from redirecting requests to attacker-controlled servers.
+	if err := validateEndpointOrigin(doc.AuthorizationEndpoint, expectedIssuer); err != nil {
```
jhrozek, line 260: this adds some duplication between pkg/auth and pkg/authserver - with this PR we reuse the structures, but not all of the logic. I already have a local patch to reduce it, but because the patch also touches pkg/auth I'd like to send it separately for easier review.

pkg/authserver/upstream/types.go
```diff
@@ -36,26 +36,5 @@ type UserInfo struct {
 	Claims map[string]any `json:"-"`
 }
 
-// OIDCEndpoints contains the discovered endpoints for an OIDC provider.
-type OIDCEndpoints struct {
```
jhrozek, line 40: we use pkg/oauth now - this code was added before we had pkg/auth

PR #3571, opened 2026-02-03
Implement backend session persistence with client pooling

### LINKED ISSUE #3062: vMCP: Implement backend session persistence with client pooling
## Problem

When a client makes multiple tool calls through vMCP within the same MCP session, each call to a backend MCP server creates a **new backend client and new backend session**. This breaks stateful backends that rely on session persistence (e.g., Playwright browser contexts, database transactions, conversation state).

### Current Behavior (Bug)

```
Client Session (single)
    │
    ├─► Tool Call 1 ──► vMCP ──► NEW Backend Client ──► Backend (session A)
    │                              └─► Initialize + CallTool + Close
    │
    ├─► Tool Call 2 ──► vMCP ──► NEW Backend Client ──► Backend (session B)
    │                              └─► Initialize + CallTool + Close
    │
    └─► Tool Call 3 ──► vMCP ──► NEW Backend Client ──► Backend (session C)
                                   └─► Initialize + CallTool + Close
```

The backend sees 3 separate sessions (A, B, C) instead of one persistent session.

### Expected Behavior

```
Client Session (single)
    │
    ├─► Tool Call 1 ──► vMCP ──► Cached Backend Client ──► Backend (session X)
    ├─► Tool Call 2 ──► vMCP ──► Cached Backend Client ──► Backend (session X)
    └─► Tool Call 3 ──► vMCP ──► Cached Backend Client ──► Backend (session X)
```

## Solution

Implement a **Session-Scoped Client Pool** that caches initialized MCP clients per `(vmcpSessionID, backendID)` tuple.

### Architecture

Each `VMCPSession` owns its own pool. Users never share connections:

```
VMCPSession "user-A"                VMCPSession "user-B"
    └─► pool["backend1"] = clientA      └─► pool["backend1"] = clientB  (DIFFERENT)
```

### Key Interface

```go
type BackendClientPool interface {
    // GetOrCreate returns cached client for backendID, or creates new one.
    GetOrCreate(ctx context.Context, target *vmcp.BackendTarget) (*client.Client, error)

    // MarkUnhealthy marks a client for replacement on next GetOrCreate.
    // Called when connection errors occur (reset, EOF, timeout).
    MarkUnhealthy(backendID string)

    // Close shuts down all pooled clients. Called on session termination.
    Close() error
}
```

## Implementation Steps

1. **Create `BackendClientPool` interface and implementation** - `pkg/vmcp/client/pool.go`
2. **Add session ID context propagation** - `pkg/vmcp/discovery/middleware.go`
3. **Extend `VMCPSession` with pool** - `pkg/vmcp/session/vmcp_session.go`
4. **Initialize pool on session registration** - `pkg/vmcp/server/server.go`
5. **Cleanup pool on session termination** - `pkg/vmcp/server/session_adapter.go`
6. **Modify `httpBackendClient` to use pool** - `pkg/vmcp/client/client.go`
7. **Add connection error detection** - for MarkUnhealthy functionality

## Files to Modify/Create

| File | Action |
|------|--------|
| `pkg/vmcp/client/pool.go` | CREATE - Pool interface + implementation |
| `pkg/vmcp/client/pool_test.go` | CREATE - Pool unit tests |
| `pkg/vmcp/session/vmcp_session.go` | MODIFY - Add pool field, Close() |
| `pkg/vmcp/client/client.go` | MODIFY - Use pool, remove defer Close() |
| `pkg/vmcp/discovery/middleware.go` | MODIFY - Propagate session ID |
| `pkg/vmcp/server/server.go` | MODIFY - Initialize pool in hook |
| `pkg/vmcp/server/session_adapter.go` | MODIFY - Cleanup pool in Terminate() |

## Error Handling

| Error Type | Action |
|------------|--------|
| Connection refused/reset | MarkUnhealthy, return error |
| Timeout | Return error (keep client) |
| MCP tool error | Return error (keep client) |
| Pool closed | Return error |

## Testing

**Unit tests:**
- GetOrCreate creates client on first access
- GetOrCreate returns same client for same backend
- Concurrent access creates only one client (double-checked locking)
- MarkUnhealthy causes recreation
- Close() closes all clients

**Integration tests:**
- Multiple tool calls use same backend session
- Session termination cleans up pool
- Error recovery works

yrobla:
Add session-scoped MCP client pooling to preserve backend state across multiple tool calls within the same vMCP session. Previously, each tool call created a new backend client and session, breaking stateful backends that rely on session persistence (e.g., Playwright browser contexts, database transactions, conversation state).

Implementation:
  - BackendClientPool: Thread-safe pool that caches initialized MCP clients per (sessionID, backendID) tuple with GetOrCreate, MarkUnhealthy, and Close methods
  - pooledBackendClient: Decorator that wraps httpBackendClient to add session-scoped pooling without changing the BackendClient interface
  - Refactor client.go to extract callToolWithClient, readResourceWithClient, and getPromptWithClient helpers for code reuse
  - Add clientPool field to VMCPSession for storing pool per session
  - Propagate session ID through context via SessionIDContextKey
  - Automatic cleanup via session TTL (no manual lifecycle management)
  - Connection error detection with MarkUnhealthy for pool eviction

Testing:
  - Comprehensive test coverage (~700 lines across pool_test.go and pooled_client_test.go)
  - Unit tests for pool operations, concurrency, error handling
  - Integration tests for session-scoped pooling and lifecycle
  - All tests passing with zero lint issues

Fixes #3062

### DIFF
--- a/pkg/vmcp/client/pooled_client.go
+++ b/pkg/vmcp/client/pooled_client.go
@@ -75,10 +75,12 @@ func (p *pooledBackendClient) getOrCreatePool(ctx context.Context) (*BackendClie
 	}
 
 	// Check if pool already exists in session
-	if poolData := vmcpSess.GetClientPool(); poolData != nil {
-		if pool, ok := poolData.(*BackendClientPool); ok {
+	if poolInterface := vmcpSess.GetClientPool(); poolInterface != nil {
+		// Type assert to concrete type
+		if pool, ok := poolInterface.(*BackendClientPool); ok {
 			return pool, nil
 		}
+		return nil, fmt.Errorf("invalid client pool type: %T", poolInterface)
 	}
 
 	// Create new pool and store in session
@@ -126,7 +128,7 @@ func (p *pooledBackendClient) CallTool(
 	result, err := p.callToolWithClient(ctx, c, target, toolName, arguments, meta)
 	if err != nil {
 		// Check if it's a connection error that warrants marking unhealthy
-		if isConnectionError(err) {
+		if shouldMarkUnhealthy(err) {
 			pool.MarkUnhealthy(target.WorkloadID)
 		}
 		return nil, err
@@ -165,7 +167,7 @@ func (p *pooledBackendClient) ReadResource(
 
 	result, err := p.readResourceWithClient(ctx, c, target, uri)
 	if err != nil {
-		if isConnectionError(err) {
+		if shouldMarkUnhealthy(err) {
 			pool.MarkUnhealthy(target.WorkloadID)
 		}
 		return nil, err
@@ -205,7 +207,7 @@ func (p *pooledBackendClient) GetPrompt(
 
 	result, err := p.getPromptWithClient(ctx, c, target, name, arguments)
 	if err != nil {
-		if isConnectionError(err) {
+		if shouldMarkUnhealthy(err) {
 			pool.MarkUnhealthy(target.WorkloadID)
 		}
 		return nil, err
@@ -214,10 +216,11 @@ func (p *pooledBackendClient) GetPrompt(
 	return result, nil
 }
 
-// isConnectionError checks if an error indicates a broken connection.
+// shouldMarkUnhealthy determines if an error indicates a broken connection
+// that warrants removing the client from the pool.
 // Returns true for network errors, connection resets, EOFs, and similar issues
 // that suggest the client should be recreated.
-func isConnectionError(err error) bool {
+func shouldMarkUnhealthy(err error) bool {
 	if err == nil {
 		return false
 	}
--- a/pkg/vmcp/session/vmcp_session.go
+++ b/pkg/vmcp/session/vmcp_session.go
@@ -15,6 +15,13 @@ import (
 // Compile-time check that VMCPSession implements transportsession.Session interface.
 var _ transportsession.Session = (*VMCPSession)(nil)
 
+// ClientPool defines the interface for session-scoped client pooling.
+// This interface avoids import cycles by not depending on concrete client implementations.
+type ClientPool interface {
+	// Any methods that VMCPSession needs from the pool would go here.
+	// For now, this is a marker interface since VMCPSession only stores/retrieves the pool.
+}
+
 // VMCPSession extends StreamableSession with domain-specific routing data.
 // This keeps routing table state in the application layer (pkg/vmcp/server)
 // rather than polluting the transport layer (pkg/transport/session) with domain concerns.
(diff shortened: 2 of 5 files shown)


pkg/vmcp/client/pooled_client.go
```diff
@@ -0,0 +1,254 @@
...
+)
+
+// pooledBackendClient wraps httpBackendClient to add session-scoped client pooling.
+// Implements the BackendClient interface by delegating to the underlying client.
+//
+// This decorator reuses initialized MCP clients across multiple tool calls within
+// the same session, preserving backend state (e.g., Playwright browser contexts,
+// database transactions, conversation history).
+type pooledBackendClient struct {
+	*httpBackendClient // Embed to inherit ListCapabilities
```
jerm-dro, line 27: I think pooledBackendClient is too coupled to the `httpBackendClient`. The reason it needs to depend on it directly is because the current implementation opens/closes clients per-request. Now, those public methods on the original `httpBackendClient` are never called, because we always decorate with this pooling client? 

Stepping back, we need to move the responsibility for opening and closing connections to the backends from the `client.go` to something that has the same lifetime as a session. I imagine we can do something simpler:
  1. On session creation, create the clients for the lifetime of the session. We aggregate capabilities and such on session startup. It seems like a reasonable place to initialize client connections as well.
  2. Update the existing httpBackendClient so it doesn't close connections on each request.
  3. On session end, close all the clients.

What do you think? It will take some digging to flesh this out more (e.g. finding the appropriate construction/destruction spots, plumbing the clients everywhere they need to go).

I recognize this is easier said than done. Clients are constructed in quite a few places. However, I think that will make the code clearer clients are injected into a session and used throughout.

pkg/vmcp/session/vmcp_session.go
```diff
@@ -35,6 +35,7 @@ type VMCPSession struct {
 	*transportsession.StreamableSession
 	routingTable *vmcp.RoutingTable
 	tools        []vmcp.Tool // Stores tools with InputSchema for type coercion
+	clientPool   interface{} // Stores BackendClientPool for session-scoped client pooling
```
jerm-dro: Blocker: When does this pool get closed? Don't we have to close the connections?

pkg/vmcp/client/pooled_client.go
yrobla: i like your proposal but i see that's one with higher risk because of what you said, there are too many places where we construct clients. We can go with first option as i designed if we want to solve it now, or create some rfc with the architectural change. I can investigate it and we can discuss together

Issue #995, opened 2025-07-09
Inspector container not always removed when inspector command is stopped.


dmjb:
Steps to repro:

1. Start inspector command
2. Hit Ctrl-C while the CLI is waiting for the inspector to start
3. Run docker ps and observe inspector container is still there

JAORMX:
I just verified that this is still an issue. The command should clean up if the context is cancelled.

carlos-gn:
Hi @eleftherias @JAORMX assign it to me pls

JAORMX:
Sure @carlos-gn ! Thanks for volunteering

carlos-gn:
here's the PR @JAORMX @eleftherias https://github.com/stacklok/toolhive/pull/3573

PR #3573, opened 2026-02-03
Add inspector auto-cleanup on interrupt

carlos-gn:
## Summary

  - Add automatic cleanup of inspector container when process is interrupted with Ctrl+C
  - Handle SIGINT/SIGTERM signals gracefully in both ready and startup states
  - Add comprehensive E2E tests for auto-cleanup behavior

Closes: https://github.com/stacklok/toolhive/issues/995

## Problem

  Previously, interrupting the inspector command left orphaned containers that required manual cleanup with thv rm inspector.

## Solution

  The inspector now listens for interrupt signals and automatically removes its container before exiting. Cleanup is
  performed whether the interrupt occurs during startup or after the inspector is fully ready.

## Test plan

  - Run thv inspector <server> and press Ctrl+C after UI is ready
  - Verify container is removed with thv list --all
  - Run task test-e2e to verify new auto-cleanup tests pass

### DIFF
--- a/cmd/thv/app/inspector.go
+++ b/cmd/thv/app/inspector.go
@@ -155,6 +155,10 @@ func inspectorCmdFunc(cmd *cobra.Command, args []string) error {
 		false, // Do not isolate network
 	)
 	if err != nil {
+		// Clean up any partially created container if deployment was interrupted
+		if cleanupErr := cleanupInspectorContainer(context.Background(), "inspector"); cleanupErr != nil {
+			logger.Debugf("Failed to cleanup inspector container after deployment error: %v", cleanupErr)
+		}
 		return fmt.Errorf("failed to create inspector workload: %w", err)
 	}
 


test/e2e/inspector_autocleanup_test.go
```diff
@@ -0,0 +1,322 @@
...
+
+	BeforeEach(func() {
+		config = e2e.NewTestConfig()
+
+		// Check if thv binary is available
+		err := e2e.CheckTHVBinaryAvailable(config)
+		Expect(err).ToNot(HaveOccurred(), "thv binary should be available")
+	})
+
+	Context("Normal operation", func() {
```
eleftherias: I wonder if the normal operation is already sufficiently covered by `inspector_test.go`

test/e2e/inspector_autocleanup_test.go
```diff
@@ -0,0 +1,322 @@
...
+			By("Waiting for inspector process to exit")
+			err = helper.waitForInspectorExit(15 * time.Second)
+			Expect(err).ToNot(HaveOccurred(), "Inspector should exit after interrupt")
+
+			By("Verifying inspector container is cleaned up")
+			Expect(helper.verifyInspectorContainerGone()).To(BeTrue(), "Container should be cleaned up")
+		})
+	})
+
+	Context("Error handling and edge cases", func() {
```
eleftherias: For the sake of e2e test speed, I'd remove this edge case test. If you feel strongly that we should keep it let me know.

eleftherias (commented):
Thanks for the PR @carlos-gn! The code looks good. There's one failing test and I've left a couple of suggestions inline.

test/e2e/inspector_autocleanup_test.go
carlos-gn: Yes!

test/e2e/inspector_autocleanup_test.go
carlos-gn: Agree!! I've simplified it. I've gotten too excited creating tests hahah

eleftherias:
Thanks @carlos-gn!

PR #3579, opened 2026-02-03
Add lazy OIDC discovery with retry

jhrozek:
Implement deferred OIDC discovery for the token validator to handle the chicken-and-egg problem where the authserver handlers may not be ready when the validator is created. This occurs with embedded auth servers running in the same pod, where Kubernetes Service endpoints take time to propagate even after the container is listening.

Discovery is now deferred until the first ValidateToken call using a sync.Once pattern. If discovery fails, exponential backoff retries up to 3 times with a ~3.5 second total window, which handles the typical networking propagation delays. Each attempt has a 5-second timeout and respects context cancellation.

This enables the auth proxy to validate tokens from its own embedded auth server without requiring complex startup sequencing or readiness coordination.

### DIFF
--- a/pkg/auth/token.go
+++ b/pkg/auth/token.go
@@ -347,28 +347,20 @@ var (
 	ErrMissingIssuerAndJWKSURL = errors.New("either issuer or JWKS URL must be provided")
 )
 
-// oidcDiscoveryParams holds the parameters needed to perform lazy OIDC discovery.
-type oidcDiscoveryParams struct {
-	issuer            string
-	caCertPath        string
-	authTokenFile     string
-	allowPrivateIP    bool
-	insecureAllowHTTP bool
-}
-
 // TokenValidator validates JWT or opaque tokens using OIDC configuration.
 type TokenValidator struct {
 	// OIDC configuration
-	issuer        string
-	audience      string
-	jwksURL       string
-	clientID      string
-	clientSecret  string // Optional client secret for introspection
-	jwksClient    *jwk.Cache
-	introspectURL string       // Optional introspection endpoint
-	client        *http.Client // HTTP client for making requests
-	resourceURL   string       // (RFC 9728)
-	registry      *Registry    // Token introspection providers
+	issuer            string
+	audience          string
+	jwksURL           string
+	clientID          string
+	clientSecret      string // Optional client secret for introspection
+	jwksClient        *jwk.Cache
+	introspectURL     string       // Optional introspection endpoint
+	client            *http.Client // HTTP client for making requests
+	resourceURL       string       // (RFC 9728)
+	registry          *Registry    // Token introspection providers
+	insecureAllowHTTP bool         // Allow HTTP (non-HTTPS) OIDC issuers for development/testing
 
 	// Lazy JWKS registration
 	jwksRegistered      bool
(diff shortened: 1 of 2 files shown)


jhrozek:
drafting because the tests suck

pkg/auth/token.go
```diff
@@ -532,19 +552,52 @@ func registerIntrospectionProviders(config TokenValidatorConfig, clientSecret st
...
+	// Determine if we need lazy OIDC discovery.
+	// Discovery is deferred when JWKS URL is not provided but issuer is,
+	// allowing the validator to start before the OIDC provider is ready.
+	// When set, ensureOIDCDiscovered will populate jwksURL on first use.
+	var discoveryConfig *oidcDiscoveryParams
 
-	// If JWKS URL is not provided but issuer is, try to discover it
-	// Skip discovery if explicitly requested via environment variable (for testing only)
+	// Skip discovery if TOOLHIVE_SKIP_OIDC_DISCOVERY is set (for testing only)
+	skipDiscovery := o.envReader.Getenv("TOOLHIVE_SKIP_OIDC_DISCOVERY") == "true"
```
tgrunnagle, line 577: do we still need `skipDiscovery` if discovery is deferred to the first validate call?

pkg/auth/token.go
```diff
@@ -626,45 +662,152 @@ func NewTokenValidator(ctx context.Context, config TokenValidatorConfig) (*Token
...
+	}
+
+	return validator, nil
 }
 
 // ensureJWKSRegistered ensures that the JWKS URL is registered with the cache.
 // This is called lazily on first use to avoid blocking startup.
+// On failure, registration is retried on subsequent calls (transient failures
+// should not permanently disable the validator).
 func (v *TokenValidator) ensureJWKSRegistered(ctx context.Context) error {
```
tgrunnagle, line 656: nit: is it worth adding a check to ensure oidc discovery has happened? I see `getKeyFromJWKS` is called after `ensureOIDCDiscovered` in `ValidateToken`, but it might make debugging easier?

pkg/auth/token.go
jhrozek: yes, I think so. Currently the flag is used in tests that use a dummy issuer. If we rewrote the tests to use a real IDP then we could drop the flag I guess.

pkg/auth/token.go
jhrozek: good idea, see the newly pushed commit

pkg/auth/token.go
```diff
@@ -346,6 +347,15 @@ var (
...
 )
 
+// oidcDiscoveryParams holds the parameters needed to perform lazy OIDC discovery.
+type oidcDiscoveryParams struct {
+	issuer            string
+	caCertPath        string
+	authTokenFile     string
+	allowPrivateIP    bool
+	insecureAllowHTTP bool
+}
```
JAORMX: Isn't there some duplicate structure?

jhrozek: Are you thinking https://github.com/stacklok/toolhive/blob/788057bd7b4b043cc065e5bb9e35b9b574bc41ae/pkg/oauth/discovery.go#L65 ? That serves a different purpose, the discovery document is protocol description and this struct is more about options for the discovery operation

jhrozek: looking at the code again the structure you pointed to:
```go
type oidcDiscoveryParams struct {
	issuer            string
	caCertPath        string
	authTokenFile     string
	allowPrivateIP    bool
	insecureAllowHTTP bool
}
```
is a subset of TokenValidatorConfig with 14 fields, but it seems excessive to carry a large struct from unrelated concern to save a private struct with 7 members..

jhrozek: aactually because the only reason we need that struct in the first place is to construct a http client which the caller already had constructed earlier we can get away without the struct at all. I'll push a new patch, thanks for the review.

PR #3580, opened 2026-02-03
Add OIDC provider methods and ID token validation

jhrozek:
This implements the remaining OIDC provider functionality building on the OIDCProvider type as well as implements OIDC ID token validation.

This PR will be followed by integration tests that verify the end-to-end OIDC flow, and then config pipeline changes that preserve the OIDC provider type through the authserver configuration so that OIDC-specific features like ID token validation are available at runtime.

### DIFF
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -11,15 +11,16 @@ This document primarily covers the middleware system for `thv` and `thv-proxyrun
 The middleware chain consists of the following components:
 
 1. **Authentication Middleware**: Validates JWT tokens and extracts client identity
-2. **Token Exchange Middleware**: Exchanges JWT tokens for external service tokens (optional)
-3. **MCP Parsing Middleware**: Parses JSON-RPC MCP requests and extracts structured data
-4. **Tool Mapping Middleware**: Enables tool filtering and override capabilities through two complementary middleware components that process outgoing `tools/list` responses and incoming `tools/call` requests (optional)
-5. **Usage Metrics Middleware**: Collects anonymous usage metrics for ToolHive development (optional)
-6. **Telemetry Middleware**: Instruments requests with OpenTelemetry (optional)
-7. **Authorization Middleware**: Evaluates Cedar policies to authorize requests (optional)
-8. **Audit Middleware**: Logs request events for compliance and monitoring (optional)
-9. **Header Forward Middleware**: Injects custom headers into requests to remote MCP servers (optional)
-10. **Recovery Middleware**: Catches panics and returns HTTP 500 errors (always present)
+2. **Upstream Token Swap Middleware**: Exchanges ToolHive JWTs for upstream IdP tokens (automatic with embedded auth server)
+3. **Token Exchange Middleware**: Exchanges JWT tokens for external service tokens via OAuth 2.0 Token Exchange (optional)
+4. **MCP Parsing Middleware**: Parses JSON-RPC MCP requests and extracts structured data
+5. **Tool Mapping Middleware**: Enables tool filtering and override capabilities through two complementary middleware components that process outgoing `tools/list` responses and incoming `tools/call` requests (optional)
+6. **Usage Metrics Middleware**: Collects anonymous usage metrics for ToolHive development (optional)
+7. **Telemetry Middleware**: Instruments requests with OpenTelemetry (optional)
+8. **Authorization Middleware**: Evaluates Cedar policies to authorize requests (optional)
+9. **Audit Middleware**: Logs request events for compliance and monitoring (optional)
+10. **Header Forward Middleware**: Injects custom headers into requests to remote MCP servers (optional)
+11. **Recovery Middleware**: Catches panics and returns HTTP 500 errors (always present)
 
 ## Architecture Diagram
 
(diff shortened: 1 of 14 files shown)


pkg/authserver/upstream/oidc.go
```diff
@@ -219,18 +227,38 @@ func (p *OIDCProviderImpl) ResolveIdentity(ctx context.Context, tokens *Tokens,
...
-	// Stub - full implementation in follow-up PR
-	return nil, errors.New("ID token validation not yet implemented")
+func (p *OIDCProviderImpl) validateIDToken(ctx context.Context, idToken, nonce string) (*oidc.IDToken, error) {
+	if p.verifier == nil {
+		return nil, errors.New("ID token verifier not initialized")
+	}
+
+	token, err := p.verifier.Verify(ctx, idToken)
+	if err != nil {
+		return nil, err
```
JAORMX: Worth wrapping this error for debugging context:

```suggestion
		return nil, fmt.Errorf("failed to verify ID token: %w", err)
```

pkg/authserver/upstream/oidc.go
```diff
@@ -241,6 +269,128 @@ func (p *OIDCProviderImpl) supportsPKCE() bool {
...
+	}
+
+	// OIDC-specific: Validate ID token per OIDC Core Section 3.1.3.7.
+	// Per Section 3.1.3.3, ID token MUST be present in OIDC token responses.
+	if tokens.IDToken == "" {
+		return nil, errors.New("ID token required for OIDC provider")
+	}
+	if _, err := p.verifier.Verify(ctx, tokens.IDToken); err != nil {
```
JAORMX: `ExchangeCode` validates the token but skips nonce validation, while `ResolveIdentity` does both. Per OIDC Core 3.1.3.7, nonce must be validated if it was sent in the auth request.

If someone calls `ExchangeCode` without `ResolveIdentity`, they skip replay protection. Is that intentional? Might be worth either passing nonce here or documenting the expected call flow.

pkg/authserver/upstream/oidc.go
```diff
@@ -241,6 +269,128 @@ func (p *OIDCProviderImpl) supportsPKCE() bool {
...
+	tokens, err := p.BaseOAuth2Provider.RefreshTokens(ctx, refreshToken)
+	if err != nil {
+		return nil, err
+	}
+
+	// OIDC-specific: Validate ID token if present.
+	// Per OIDC Core Section 12.2, refresh responses MAY include a new ID token
+	// (unlike ExchangeCode where it's required per Section 3.1.3.3).
+	if tokens.IDToken != "" && p.verifier != nil {
+		if _, err := p.verifier.Verify(ctx, tokens.IDToken); err != nil {
```
JAORMX: Same nonce question as `ExchangeCode`. The spec is murky on refresh flows since there's no new auth request. If skipping nonce validation here is intentional, a brief comment explaining why would help.

pkg/authserver/upstream/oidc.go
jhrozek: Yes, it was intentional but maybe bad API design. It is not a security issue, because the caller of the upstream providers (the auth-proxy callback) requires the sub claim and the only way to get it is to call ResolveIdentity which verifies the nonce.

Longer explanation:

So I was trying to construct the upstream APIs to build atop the protocols and maybe that leaks the protocols too much into the code, requiring protocol knowledge.

The ExchangeCode is pure oauth - the oidc provider just uses the base OAuth provider implementation and there's no nonce there.

The separate ResolveIdentity step exists on the API layer because OAuth requires a separate step (calling UserInfo endpoint). With OIDC, you get the ID token for free and the only thing that the ResolveIdentity step does with OIDC is to verify the nonce and parse the claims.

It is not a security issue because 

What if I changed the provider interface to combine ResolveIdentity and ExchangeCode into ExchangeCodeForIdentity in the next PR? I think it would make sense because you always need to call them together.

pkg/authserver/upstream/oidc.go
jhrozek: suggestion taken

pkg/authserver/upstream/oidc.go
jhrozek: the spec says that the refresh should not contain nonce. But, while I was checking the spec I found out that I was missing another check - to make sure the sub matches the original. I added that as a separate patch.

pkg/authserver/upstream/oidc.go
```diff
@@ -219,18 +232,38 @@ func (p *OIDCProviderImpl) ResolveIdentity(ctx context.Context, tokens *Tokens,
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to verify ID token: %w", err)
+	}
+
+	// Validate nonce if expected (was sent in authorization request).
+	// This ensures that when a nonce is provided, the token MUST contain it
+	// and it MUST match, preventing replay attacks.
+	if nonce != "" {
+		if token.Nonce == "" {
+			return nil, errors.New("ID token missing nonce claim when nonce was expected")
```
tgrunnagle: nit: why `errors.New` here and `ErrNonceMismatch` below?

pkg/authserver/upstream/oidc.go
tgrunnagle, line 236: Question: Do we always request `openid` scope? Asking because `scopes` for `oidcConfig` is configurable.

pkg/authserver/upstream/oidc.go
jhrozek: thanks this was an oversight

pkg/authserver/upstream/oidc.go
jhrozek: so, per the spec the openid scope is mandatory, if omitted, the request should be treated as oauth - I think it should be an error to use an oidc provider and not include the openid scope (you should just use the plain oauth provider in that case)

Issue #3566, opened 2026-02-03
AWS STS: Add generic CEL evaluation engine


jhrozek:
Create a reusable CEL engine in `pkg/cel/` for expression compilation and evaluation.

The engine should support lazy initialization, compiled expression caching, and thread-safe concurrent use. Include detailed error types with line/column info for debugging CEL syntax errors.

This provides the foundation for AWS STS role selection and can be reused elsewhere (vMCP routing, authorization).

Depends on: `github.com/google/cel-go`

Ref: THV-0033

PR #3582, opened 2026-02-03
Add generic CEL expression engine package (Mediator strikes back!)

jhrozek:
Most of this PR is adapted code from https://github.com/mindersec/minder/blob/main/pkg/engine/selectors/selectors.go (which I also wrote so don't hold back during the review you won't be screaming at yourself)

Introduce a reusable CEL (Common Expression Language) engine in pkg/cel/ for compiling and evaluating expressions against arbitrary data contexts. The engine provides lazy-initialized, thread-safe environment caching, expression compilation with structured parse and type-check error reporting, boolean and generic value evaluation helpers, and built-in safeguards against denial-of-service via configurable expression length and runtime cost limits.

This is the first PR in the AWS STS integration series. The CEL engine will be used by the role mapper component to evaluate OIDC token claims against configurable expressions, enabling
flexible mapping from identity claims to AWS STS roles without hardcoding authorization logic.

Fixes: #3566

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/go-jose/go-jose/v4 v4.1.3
 	github.com/go-logr/zapr v1.3.0
 	github.com/gofrs/flock v0.13.0
+	github.com/google/cel-go v0.27.0
 	github.com/google/go-cmp v0.7.0
 	github.com/google/go-containerregistry v0.20.7
 	github.com/google/uuid v1.6.0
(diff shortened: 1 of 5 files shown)


JAORMX:
@jhrozek do you think it makes sense to add this to toolhive-core as an alpha package?

jhrozek:
> @jhrozek do you think it makes sense to add this to toolhive-core as an alpha package?

I think it makes sense if you're OK with short-cutting the usual inclusion criteria (https://github.com/stacklok/toolhive-core?tab=readme-ov-file#fast-track-simple-packages)

Maybe we can say that the package has been used for very long because it's been used in minder?

jhrozek:
Closed in favor of https://github.com/stacklok/toolhive-core/pull/11

PR #3588, opened 2026-02-04
Ensure API E2E tests clean up after themselves

dmjb:
The code previously relied on API calls to clean up local test state. However, since the APIs are async, and the servers get spun up and down during the tests, a lot of workloads would end up in a partially deleted state. This change uses the CLI for final cleanup, ensuring that the cleanup is complete. In tests where delete endpoints are tested, it introduces some logic to wait for the deletion to finish.

This mostly affects local runs of the E2E tests.

### DIFF
--- a/docs/server/docs.go
+++ b/docs/server/docs.go
@@ -3765,7 +3765,7 @@ const docTemplate = `{
         },
         "/api/v1beta/workloads/delete": {
             "post": {
-                "description": "Delete multiple workloads by name or by group",
+                "description": "Delete multiple workloads by name or by group asynchronously.\nReturns 202 Accepted immediately. Deletion happens in the background.",
                 "requestBody": {
                     "content": {
                         "application/json": {
@@ -3795,7 +3795,7 @@ const docTemplate = `{
                                 }
                             }
                         },
-                        "description": "Accepted"
+                        "description": "Accepted - deletion started"
                     },
                     "400": {
                         "content": {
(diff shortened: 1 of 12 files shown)


rdimitrov (dismissed):

jhrozek:
LGTM, but the linting should be fixed

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **2 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Synchronous implementation of intended asynchronous API</strong>
   - [Location](https://github.com/stacklok/toolhive/blob/ecaee3f5274b9c71d1ec4b2b9f001dde7fd5a034/pkg/api/v1/workloads.go#L248-L255): `pkg/api/v1/workloads.go:L248-L255`
   - Impact: The handler may block until the deletion is complete, which might lead to client timeouts and doesn't quite match the "background processing" mentioned in the documentation.
   - Suggestion: Consider offloading the deletion logic to a background goroutine or job queue so the handler can return the 202 response immediately.

2. <strong>Potential for orphaned resources during test cleanup</strong>
   - [Location](https://github.com/stacklok/toolhive/blob/ecaee3f5274b9c71d1ec4b2b9f001dde7fd5a034/test/e2e/e2e_suite_test.go#L91-L99): `test/e2e/e2e_suite_test.go:L91-L99`
   - Impact: If the environment is torn down while asynchronous deletions are still in progress, some resources or containers might not be fully cleaned up.
   - Suggestion: It might be safer to update the cleanup logic to verify or poll for the actual removal of resources before finalizing the suite teardown.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

PR #3596, opened 2026-02-04
Add TOML config file support with array-of-tables format

aponcedeleonch:
## Summary
- Add TOML configuration file format support infrastructure with the array-of-tables (`[[section]]`) format
- This format is used by clients like Mistral Vibe
- Foundation PR for TOML support - nested tables format will follow in a separate PR

## Changes
- Added `TOML` extension constant and `TOMLStorageType` type with `TOMLStorageTypeArray` constant
- Added shared TOML helper functions: `tomlWithFileLock`, `readTOMLConfig`, `writeTOMLConfig`, `extractURLFromMCPServer`, `convertToAnySlice`
- Added `TOMLConfigUpdater` struct and methods for array-of-tables format
- Added helper functions `extractServersKeyFromConfig` and `extractURLLabelFromConfig`
- Updated `validateConfigFileFormat` to support TOML validation
- Updated `CreateClientConfig` to support TOML file creation
- Updated `retrieveConfigFileMetadata` to create TOML updaters
- Added comprehensive tests for `TOMLConfigUpdater`

### DIFF
--- a/pkg/client/config_editor.go
+++ b/pkg/client/config_editor.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pelletier/go-toml/v2"
 	"github.com/tailscale/hujson"
 	"github.com/tidwall/gjson"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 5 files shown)


pkg/client/config_editor.go
```diff
@@ -292,6 +293,221 @@ func (ycu *YAMLConfigUpdater) Remove(serverName string) error {
...
+	var config map[string]any
+	if err := toml.Unmarshal(content, &config); err != nil {
+		return nil, fmt.Errorf("failed to parse existing TOML config: %w", err)
+	}
+	return config, nil
+}
+
+// writeTOMLConfig marshals and writes the config to the specified TOML file path.
+func writeTOMLConfig(path string, config map[string]any) error {
+	updatedContent, err := toml.Marshal(config)
```
dmjb, line 339: One issue based on related bug - make these file writes atomic. We've had issues where ToolHive died mid-write and corrupted the configs. See an example of how this is handled elsewhere: https://github.com/stacklok/toolhive/pull/3590/changes#diff-e41e855488e84a81dea6993910442e502c0e229bdcb61d841daa5adaf848581fR753

(We may want to make that atomic file write into a shared helper function for use in multiple places)

aponcedeleonch: I see that for other client config files we're using a lock file, I used the same for TOML files.

Do you think it makes sense to do atomic writes using a temp file even with a lock? If so, probably it's better to handle it as a separate PR since it should also impact JSON and YAML files implementations

aponcedeleonch: I am implementing it in a separate PR. Let's wait for that one as a base to have the common logic for atomic writes in place and then I can re-use it in this PR

aponcedeleonch: Handled in: https://github.com/stacklok/toolhive/pull/3613

dmjb (approved):

Issue #3567, opened 2026-02-03
AWS STS: Add role mapper with CEL-based selection


jhrozek:
Implement role selection that maps JWT claims to IAM roles using CEL expressions.

Support two syntaxes:
- Simple: `claim: admins` compiles internally to `"admins" in claims.groups`
- Full CEL: `matcher: '"admins" in claims.groups && !has(claims.act)'`

Priority-based selection: when multiple roles match, lowest priority wins.

Files: `pkg/auth/awssts/role_mapper.go`, `pkg/auth/awssts/config.go`

Depends on: #3566

Ref: THV-0033

PR #3609, opened 2026-02-04
Add CEL-based AWS STS role mapper for claim-based IAM role selection

jhrozek:
Implement a role mapper that selects IAM roles based on JWT claims using CEL expressions with priority-based selection. The mapper supports two configuration modes: a simple claim syntax where a value like "admins" is checked for membership in a configurable claim (defaulting to "groups"), and a full CEL matcher syntax for complex expressions such as agent delegation checks using the RFC 7519 "act" claim. All CEL expressions are compiled at configuration load time for fail-fast validation, claim values are validated against a safe-character regex to prevent CEL injection, and role ARNs are validated using the AWS SDK. When multiple mappings match, the one with the lowest priority number wins, with configuration order as a tie-breaker. A fallback role ARN can be configured for when no mapping matches.

This is the first in a series of PRs that add AWS STS authentication to ToolHive, enabling MCP servers to authenticate with AWS services by exchanging incoming OIDC tokens for temporary AWS credentials via STS AssumeRoleWithWebIdentity. Subsequent PRs then add token exchange with SigV4 request signing, credential caching, HTTP middleware integration for the CLI runner, operator CRD support via MCPExternalAuthConfig, and user-facing documentation. This PR provides the claim-to-role mapping foundation that all subsequent layers depend on.

Fixes: #3567

### DIFF
--- a/pkg/auth/awssts/config.go
+++ b/pkg/auth/awssts/config.go
@@ -1,0 +1,69 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package awssts provides AWS STS token exchange with SigV4 signing support.
+package awssts
+
+// MinSessionDuration is the minimum allowed session duration (AWS limit).
+const MinSessionDuration int32 = 900
+
+// MaxSessionDuration is the maximum allowed session duration (12 hours).
+const MaxSessionDuration int32 = 43200
+
+// defaultRoleClaim is the default JWT claim to use for role mapping.
+const defaultRoleClaim = "groups"
+
+// Config holds configuration for AWS STS token exchange.
+type Config struct {
+	// Region is the AWS region for STS and SigV4 signing.
+	Region string `json:"region" yaml:"region"`
+
+	// Service is the AWS service name for SigV4 signing (default: "aws-mcp").
+	Service string `json:"service" yaml:"service"`
+
+	// FallbackRoleArn is the IAM role ARN to assume when no role mapping matches.
+	FallbackRoleArn string `json:"fallback_role_arn,omitempty" yaml:"fallback_role_arn,omitempty"`
+
+	// RoleMappings maps JWT claim values to IAM roles with priority.
+	RoleMappings []RoleMapping `json:"role_mappings,omitempty" yaml:"role_mappings,omitempty"`
+
+	// RoleClaim is the JWT claim to use for role mapping (default: "groups").
+	RoleClaim string `json:"role_claim,omitempty" yaml:"role_claim,omitempty"`
+
+	// SessionDuration is the duration in seconds for assumed role credentials.
+	SessionDuration int32 `json:"session_duration,omitempty" yaml:"session_duration,omitempty"`
+
+	// SessionNameClaim is the JWT claim to use for role session name (default: "sub").
+	SessionNameClaim string `json:"session_name_claim,omitempty" yaml:"session_name_claim,omitempty"`
+}
+
+// GetRoleClaim returns the configured role claim or the default.
+func (c *Config) GetRoleClaim() string {
+	if c.RoleClaim != "" {
+		return c.RoleClaim
+	}
+	return defaultRoleClaim
+}
+
+// RoleMapping maps a JWT claim value or CEL expression to an IAM role with explicit priority.
+type RoleMapping struct {
+	// Claim is the simple claim value to match (e.g., group name).
+	// Internally compiles to a CEL expression: "<claim_value>" in claims["<role_claim>"]
+	// Mutually exclusive with Matcher.
+	Claim string `json:"claim,omitempty" yaml:"claim,omitempty"`
+
+	// Matcher is a CEL expression for complex matching against JWT claims.
+	// The expression has access to a "claims" variable containing all JWT claims.
+	// Examples:
+	//   - "admins" in claims["groups"]
+	//   - claims["sub"] == "user123" && !("act" in claims)
+	// Mutually exclusive with Claim.
+	Matcher string `json:"matcher,omitempty" yaml:"matcher,omitempty"`
+
+	// RoleArn is the IAM role ARN to assume when this mapping matches.
+	RoleArn string `json:"role_arn" yaml:"role_arn"`
+
+	// Priority determines selection order (lower number = higher priority).
+	// When multiple mappings match, the one with the lowest priority is selected.
+	Priority int `json:"priority" yaml:"priority"`
+}
--- a/pkg/auth/awssts/role_mapper.go
+++ b/pkg/auth/awssts/role_mapper.go
@@ -1,0 +1,284 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package awssts
+
+import (
+	"cmp"
+	"fmt"
+	"regexp"
+	"slices"
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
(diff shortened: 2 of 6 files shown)


jhrozek:
I know you wanted to take a stab at the review as well @ChrisJBurns - I'll be happy to get post-merge reviews as well! I'm going to merge since this is the a PR in a series and not a load-bearing code yet.

pkg/auth/awssts/role_mapper.go
```diff
@@ -0,0 +1,284 @@
...
+		return fmt.Errorf("%w: not an IAM ARN: %s", ErrInvalidRoleArn, roleArn)
+	}
+
+	// Resource should start with "role/"
+	if !strings.HasPrefix(parsed.Resource, "role/") {
+		return fmt.Errorf("%w: not a role ARN: %s", ErrInvalidRoleArn, roleArn)
+	}
+
+	// Verify account ID is present and valid (12 digits)
+	if len(parsed.AccountID) != 12 {
```
ChrisJBurns, line 61: nit: not sure if this validates if there's digits - only that there are 12 characters. The [`arn.Parse()`](https://github.com/aws/aws-sdk-go-v2/blob/v1.41.1/aws/arn/arn.go#L61) itself doesn't validate either, it just separates the segments using `:` delimiter. However, given a world where I've never seen an account ID include letters, I think we're safe here 😆

pkg/auth/awssts/role_mapper.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/stacklok/toolhive/pkg/logger"
+)
+
+// safeClaimValueRegex defines the whitelist of characters allowed in JWT claim values
+// used for CEL expression interpolation. This prevents CEL injection while covering
+// legitimate group/role claim values from major identity providers (Azure AD, Okta,
+// Auth0, Google Workspace, Keycloak, LDAP).
+//
+// Blocked characters that enable CEL injection: " \ ( ) | & ` $ { } [ ] < > ^ %
+var safeClaimValueRegex = regexp.MustCompile(`^[a-zA-Z0-9@.:,;/\-_=+*#!?'~ ]+$`)
```
ChrisJBurns, line 26: I asked Claude about this because regex always makes my head hurt, and it's a Friday, and I like my head 😆 
"CEL uses ' as an alternative string delimiter. While the current buildCELExpression interpolates into double-quoted CEL strings (so single quotes inside are treated as literals today), keeping ' in the allowlist violates defense-in-depth -- if the interpolation format ever changes, it becomes an injection vector.
  Real-world group names from IdPs (Azure AD, Okta, Auth0) don't use single quotes. Remove it.

  Note: The security agent flagged this as "CRITICAL exploitable now" -- that's overstated. With the current double-quote interpolation in buildCELExpression, a single quote inside "admin' || 'x" stays as a literal character within the double-quoted CEL string. It's not currently exploitable, but should still be
  removed as a precaution."

pkg/auth/awssts/role_mapper.go
```diff
@@ -0,0 +1,284 @@
...
+// ValidateConfig validates the AWS STS configuration structure.
+// It checks that required fields are present, ARNs are well-formed, claim values
+// are safe for CEL interpolation, and session duration is within bounds.
+//
+// This performs structural validation only — CEL expression compilation is handled
+// by NewRoleMapper. It is safe to call standalone for early validation at config
+// load time. NewRoleMapper calls this internally, so callers do not need to call both.
+func ValidateConfig(cfg *Config) error {
+	if cfg == nil {
+		return fmt.Errorf("config is nil")
```
ChrisJBurns, line 190: nit: we seem to be using `ErrMissingRegion` below which is `errors.New("AWS region is required")`. Wondering if we should do here for consistency? `return ErrNilConfig` `ErrNilConfig = errors.New("config is nil")`  etc etc.

pkg/auth/awssts/role_mapper.go
```diff
@@ -0,0 +1,284 @@
...
+	for i, mapping := range cfg.RoleMappings {
+		if err := validateRoleMapping(i, mapping); err != nil {
+			return err
+		}
+	}
+
+	// Validate session duration if specified
+	if cfg.SessionDuration != 0 {
+		if cfg.SessionDuration < MinSessionDuration {
+			return fmt.Errorf("session duration %d is below minimum %d seconds", cfg.SessionDuration, MinSessionDuration)
```
ChrisJBurns, line 227: nit: we could maybe use sentinel errors (`ErrInvalidSessionDuration`) here for use with `errors.Is()`?

pkg/auth/awssts/config.go
```diff
@@ -0,0 +1,69 @@
...
+	//   - claims["sub"] == "user123" && !("act" in claims)
+	// Mutually exclusive with Claim.
+	Matcher string `json:"matcher,omitempty" yaml:"matcher,omitempty"`
+
+	// RoleArn is the IAM role ARN to assume when this mapping matches.
+	RoleArn string `json:"role_arn" yaml:"role_arn"`
+
+	// Priority determines selection order (lower number = higher priority).
+	// When multiple mappings match, the one with the lowest priority is selected.
+	Priority int `json:"priority" yaml:"priority"`
```
ChrisJBurns, line 68: If this is omitted, this becomes `0` by default which leads to highest priority? Is that intended? What happens if multiple are declared but both are omitted and become `0` by default?

pkg/auth/awssts/role_mapper_test.go
```diff
@@ -0,0 +1,595 @@
...
+			}
+			require.Error(t, err)
+			if tt.wantErrIs != nil {
+				assert.ErrorIs(t, err, tt.wantErrIs)
+			}
+		})
+	}
+}
+
+func TestRoleMapper_Concurrency(t *testing.T) {
```
ChrisJBurns, line 526: nit: The test checks results are "one of the expected roles" but doesn't verify the right role for the right input. When `i % 3 == 0` (groups=admins), the role should always be `AdminRole`. As written, I _think_ the test would pass even if the mapper returned random valid roles?

pkg/auth/awssts/role_mapper_test.go
```diff
@@ -0,0 +1,595 @@
...
+			case 0:
+				groups = []any{"admins"}
+			case 1:
+				groups = []any{"developers"}
+			case 2:
+				groups = []any{"users"}
+			}
+
+			claims := map[string]any{
+				"sub":    "user" + string(rune('0'+i%10)),
```
ChrisJBurns, line 568: nit: forgot `rune` existed 😆 , would `fmt.Sprintf("user%d", i)` be better here, even works beyond single digits

pkg/auth/awssts/role_mapper.go
```diff
@@ -0,0 +1,284 @@
...
+// This performs structural validation only — CEL expression compilation is handled
+// by NewRoleMapper. It is safe to call standalone for early validation at config
+// load time. NewRoleMapper calls this internally, so callers do not need to call both.
+func ValidateConfig(cfg *Config) error {
+	if cfg == nil {
+		return fmt.Errorf("config is nil")
+	}
+
+	// Region is required
+	if cfg.Region == "" {
```
ChrisJBurns, line 194: nit: given AWS STS is available in all standard regions, `us-wesr-1` would fail at runtime during an actual STS call. Could we perhaps validate that the specified region exist ahead of time?

Something like (this was very much Clauded - so definitely verify this works 😄 )

```go
import "github.com/aws/aws-sdk-go/aws/endpoints"

func getAllRegions() []string {
    var regions []string
    resolver := endpoints.DefaultResolver()
    partitions := resolver.(endpoints.EnumPartitions).Partitions()
    
    for _, p := range partitions {
        for id := range p.Regions() {
            regions = append(regions, id)
        }
    }
    return regions
}
```

PR #3610, opened 2026-02-04
Align authserver DCR client scopes with discovery scopes_supported

jhrozek:
DCR was assigning hardcoded DefaultScopes ["openid", "profile", "email"] to registered clients, while the discovery document advertised whatever was configured in oidcConfig.inline.scopes. When those differed, clients would read scopes_supported from discovery, request those scopes, and get rejected by fosite because the client wasn't allowed them.

For example, with this MCPServer config:

```
    oidcConfig:
      inline:
        scopes: [user:email, read:user, repo]
```

the discovery document would advertise scopes_supported: ["user:email", "read:user", "repo"], but a DCR client would only be allowed ["openid", "profile", "email"]. Requesting "user:email" would fail with invalid_scope.

The fix: DCR now reads h.config.ScopesSupported so clients are allowed to request exactly the scopes the server advertises. The config default for ScopesSupported references registration.DefaultScopes directly, giving a single source of truth instead of two arrays to keep in sync.

### DIFF
--- a/pkg/authserver/config.go
+++ b/pkg/authserver/config.go
@@ -13,6 +13,7 @@ import (
 
 	servercrypto "github.com/stacklok/toolhive/pkg/authserver/server/crypto"
 	"github.com/stacklok/toolhive/pkg/authserver/server/keys"
+	"github.com/stacklok/toolhive/pkg/authserver/server/registration"
 	"github.com/stacklok/toolhive/pkg/authserver/upstream"
 	"github.com/stacklok/toolhive/pkg/logger"
 	"github.com/stacklok/toolhive/pkg/networking"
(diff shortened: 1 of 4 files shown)


jhrozek:
@tgrunnagle I saw one of your manifests using explicit scopes in the oidcConfig.inline.scopes attribute and realized (by testing and running into a bug) that we don't handle the situation well, all my testing was using implicit scopes for the auth server. This fixes the bug. Although in most cases explicit scopes should not be needed, the upstream scopes are configured separately using `MCPExternalAuthConfig.upstreamProviders[].scopes`

JAORMX (approved):

PR #3613, opened 2026-02-05
Extract atomic file write utility to shared fileutils package

aponcedeleonch:
The atomicWriteFile function was previously defined in pkg/workloads/statuses/file_status.go and used only for status file writes. This refactoring extracts it to a new pkg/fileutils package as AtomicWriteFile, making it reusable across the codebase.

Changes:
- Create new pkg/fileutils package with AtomicWriteFile function
- Move atomic write tests to pkg/fileutils/atomic_test.go
- Update pkg/client/config_editor.go to use fileutils.AtomicWriteFile for JSON/YAML config writes
- Update pkg/workloads/statuses/file_status.go to use fileutils.AtomicWriteFile for status writes
- Remove duplicate atomicWriteFile implementation and tests from file_status.go

This improves code reusability and ensures consistent atomic write behavior across all file operations (config files, status files) preventing partial writes during concurrent access or system failures.


dmjb (dismissed):

pkg/client/config_editor.go
```diff
@@ -92,8 +93,8 @@ func (jcu *JSONConfigUpdater) Upsert(serverName string, data MCPServer) error {
 		logger.Errorf("Failed to format the patched file: %v", err)
 	}
 
-	// Write back to the file
-	if err := os.WriteFile(jcu.Path, formatted, 0600); err != nil {
+	// Write back to the file atomically
+	if err := fileutils.AtomicWriteFile(jcu.Path, formatted, 0600); err != nil {
 		logger.Errorf("Failed to write file: %v", err)
```
jhrozek, line 98: out of curiosity why are the errors ignored? (The pattern is repeated through the file, so I assume it's not a mistake)

dmjb: I think the logic was to treat the failure to update the config as a non-fatal error when creating a workload. However, these logs should be logged as warnings, not errors.

aponcedeleonch: Yes, there's some inconsistency in how we manage failing writes to a file. I see that for JSON files the errors are just logged and for YAML and TOML are returned. Will make sure everything is consistent in a separate PR.

Issue #215, opened 2025-04-18
failed to validate config file for Roo code on Linux


angelogalanti:
thv config register-client roo-code

A new version of ToolHive is available: v0.0.15
Currently running: dev
Successfully registered client: roo-code
1:17PM WRN failed to validate config file: %!w(*errors.errorString=&{file does not exist: /home/x/.config/Code/User/globalStorage/rooveterinaryinc.roo-cline/settings/mcp_settings.json})

The path for the mcp_settings.json in Linux is:
/home/x/.vscode-server/data/User/globalStorage/rooveterinaryinc.roo-cline/settings/mcp_settings.json

JAORMX:
Hey! Thanks for reaching out! We'll get to fixing this.

How did you install? Are you building from source? Can you pull the latest if so? Seems to me like your missing some changes that might help. Also, I suggest using 'task build' since that will give us the actual version you're running.

dmjb:
@angelogalanti It appears you are using VScode server. We currently support the VScode desktop editor on Linux, and I can confirm that `$HOME/.config/Code/User/globalStorage/rooveterinaryinc.roo-cline/settings/mcp_settings.json` is the correct path for the desktop client on Linux. This was validated with the latest [VScode package from here](https://code.visualstudio.com/docs/setup/linux#_install-vs-code-on-linux) and installing on Debian.

I am going to rename this issue, because I think we should explore support VScode server as well.

jorgemoralespou:
We're also using coder/code-server and even though it follows VSCode file structure, the root is located in a different directory, and also this can be configurable per installation. It would be ideal if there was an option to make a configurable [mcpClientConfig](https://github.com/stacklok/toolhive/blob/main/pkg/client/config.go#L51-L61) that the user could provide in the toolhive configuration file that is being worked on in  https://github.com/stacklok/toolhive/pull/919.

PR #3618, opened 2026-02-05
Add code-server as a supported client

dmjb:
Fixes: #215


aponcedeleonch (changes requested):
I think you missed one spot: `cmd/thv/app/client.go`. We need to come up with a way to keep the clients centralized in `pkg/client/config.go` and propagate any client that we add there to all the needed places

Issue #3377, opened 2026-01-21
OAuth login URL is not shown when running remote server with auth


eleftherias:
## Bug description
When running 
```
thv run --remote-auth-skip-browser
```
I expect to see the log
```
Please open this URL in your browser:
```

However, the foreground process exits too quickly and it's not possible to see the log unless I open the `.log` file.
This makes it difficult to log in to remote servers when my shell doesn't have permissions to open a browser.

## Steps to reproduce
```
thv run context7-remote --remote-auth-skip-browser
```

## Expected behavior
I expect to see the login URL printed:
```
Please open this URL in your browser: https://context7.com/api/oauth/authorize?client_id=...
```
Or another mechanism that allows me to continue the OAuth flow.

## Actual behavior
`thv run` return immediately. The log with the URL is only available when searching through the log file.

## Additional context
This problem becomes more painful given the 30s timeout when connecting to remote servers.

PR #3621, opened 2026-02-05
Change workload manager to follow logging best practices

dmjb:
This changes the logging in the workload manager to follow best practices. This also creates a custom error for a place where the error was being checked through string comparison.

### DIFF
--- a/pkg/workloads/manager.go
+++ b/pkg/workloads/manager.go
@@ -549,7 +549,8 @@ func (d *DefaultManager) RunWorkloadDetached(ctx context.Context, runConfig *run
 				logger.Warnf("Failed to close log file: %v", err)
 			}
 		}()
-		logger.Debugf("Logging to: %s", logFilePath)
+		// Keeping this log at INFO level until https://github.com/stacklok/toolhive/issues/3377 is fixed
+		logger.Infof("Logging to: %s", logFilePath)
 	}
 
 	// Use the start command to start the detached process


pkg/workloads/manager.go
```diff
@@ -422,7 +422,7 @@ func (d *DefaultManager) RunWorkload(ctx context.Context, runConfig *runner.RunC
 
 	for attempt := 1; attempt <= maxRetries; attempt++ {
 		if attempt > 1 {
-			logger.Infof("Restart attempt %d/%d for %s after %v delay", attempt, maxRetries, runConfig.BaseName, retryDelay)
```
eleftherias, line 421: This is one that I opted to keep at INFO level because it could be a long running operation.

pkg/workloads/manager.go
```diff
@@ -546,15 +546,14 @@ func (d *DefaultManager) RunWorkloadDetached(ctx context.Context, runConfig *run
...
 	if err != nil {
-		logger.Warnf("Warning: Failed to create log file: %v", err)
+		logger.Warnf("Failed to create log file: %v", err)
 	} else {
 		defer func() {
 			if err := logFile.Close(); err != nil {
 				logger.Warnf("Failed to close log file: %v", err)
 			}
 		}()
-		// Keeping this log at INFO level until https://github.com/stacklok/toolhive/issues/3377 is fixed
```
eleftherias, line 552: This is another one I'd like to keep at INFO level until https://github.com/stacklok/toolhive/issues/3377 is fixed

pkg/workloads/manager.go
```diff
@@ -997,19 +996,19 @@ func (d *DefaultManager) updateSingleWorkload(ctx context.Context, workloadName
 	childCtx, cancel := context.WithTimeout(ctx, AsyncOperationTimeout)
 	defer cancel()
 
-	logger.Infof("Starting update for workload %s", workloadName)
```
eleftherias, line 1001: This is another one that could be a long running operation, so it could be good to keep the feedback

pkg/workloads/manager.go
```diff
@@ -542,15 +542,14 @@ func (d *DefaultManager) RunWorkloadDetached(ctx context.Context, runConfig *run
...
+		logger.Infof("Failed to create log file: %v", err)
 	} else {
 		defer func() {
 			if err := logFile.Close(); err != nil {
 				logger.Warnf("Failed to close log file: %v", err)
 			}
 		}()
-		// Keeping this log at INFO level until https://github.com/stacklok/toolhive/issues/3377 is fixed
-		logger.Infof("Logging to: %s", logFilePath)
+		logger.Debugf("Logging to: %s", logFilePath)
```
eleftherias: I think this change was reverted but it came back

eleftherias (approved):

Issue #3623, opened 2026-02-05
[Bug] vMCP does not follow standard MCP Streamable HTTP behavior


juancarlosm:
## Bug description                                                                                                                                                                                                                         
                                                                                                                                                                                                                                             
The vMCP server does not follow standard MCP Streamable HTTP behavior. When a client makes a GET request to `/mcp` without the `Accept: text/event-stream` header, the server should reject it — as other MCP servers do 

Instead, vMCP silently opens a streaming connection, which can break health checks and confuse clients that didn't intend to open a stream.

Additionally, the server does not send keepalive pings on streaming connections, making it appear hung and causing proxies and load balancers to close idle connections.

## Steps to reproduce

```bash
# GET without Accept header — should be rejected
curl http://localhost:4483/mcp

# GET with Accept header — should show periodic pings
curl http://localhost:4483/mcp -H "Accept: text/event-stream"
```
Expected behavior

- GET without Accept: text/event-stream returns HTTP 406 with a JSON-RPC error
- GET with Accept: text/event-stream opens a streaming connection with periodic keepalive pings

Actual behavior

- GET without the header hangs ~15s then silently opens a stream
- GET with the header opens a stream but sends no keepalive pings

Additional context

- Spec reference: https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#streamable-http
- Fix: https://github.com/juancarlosm/toolhive/commit/2d5698d8 (~30 lines, single file)

jerm-dro:
Assigning @juancarlosm since they have a fix up and myself so it's clear they have a reviewer.

PR #3624, opened 2026-02-05
vMCP does not follow standard MCP Streamable HTTP behavior

juancarlosm:
Added:

  - GET without Accept: text/event-stream returns HTTP 406 with a JSON-RPC error
  - GET with Accept: text/event-stream opens a streaming connection with periodic keepalive pings

fixes: #3623

Tested:

```bash
curl http://vmcp-search.mcp.svc.cluster.local:4483
{"jsonrpc":"2.0","id":"server-error","error":{"code":-32600,"message":"Not Acceptable: Client must accept text/event-stream"}}

curl http://vmcp-search.mcp.svc.cluster.local:4483 -H "Accept: text/event-stream"
event: message
data: {"jsonrpc":"2.0","id":1,"method":"ping"}
```


pkg/vmcp/server/server.go
```diff
@@ -429,9 +432,21 @@ func (s *Server) Start(ctx context.Context) error {
...
+	// Execution order: telemetry → backend enrichment → discovery → audit → auth → accept-check → handler
+	//
+	// Accept header validation is the innermost wrapper so invalid GETs are rejected
+	// immediately without triggering discovery timeout or middleware overhead.
+	var mcpHandler http.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodGet && !acceptsEventStream(r) {
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusNotAcceptable)
+			//nolint:errcheck // Best-effort write of error response body
+			_, _ = w.Write([]byte(`{"jsonrpc":"2.0","id":"server-error","error":{"code":-32600,"message":"Not Acceptable: Client must accept text/event-stream"}}`))
```
jerm-dro: nit: let's log the error instead of ignoring it entirely.

pkg/vmcp/server/server.go
```diff
@@ -429,9 +432,21 @@ func (s *Server) Start(ctx context.Context) error {
...
 
 	// MCP endpoint - apply middleware chain (wrapping order, execution happens in reverse):
-	// Code wraps: auth → audit → discovery → backend enrichment → telemetry
-	// Execution order: telemetry → backend enrichment → discovery → audit → auth → handler
-	var mcpHandler http.Handler = streamableServer
+	// Code wraps: accept-check → auth → audit → discovery → backend enrichment → telemetry
+	// Execution order: telemetry → backend enrichment → discovery → audit → auth → accept-check → handler
+	//
+	// Accept header validation is the innermost wrapper so invalid GETs are rejected
+	// immediately without triggering discovery timeout or middleware overhead.
```
jerm-dro: Blocker: The comment contradicts the code. As the code is written, this validation will actually execute last, because the middlewares below wrap it. 

The comment is correct this is the innermost, but we actually want it to be the outermost, so it can fail requests immediately.

jerm-dro: The comment above re: code wrapping and execution order also appears incorrect. The actual execution order is:
```
// Execution order: recovery → auth → audit → discovery → backend enrichment → telemetry → handler
```
Let's remove the statement about "Code wraps," since the ordering of wrapping is clearly documented below. The statement about execution order is good to keep, since it documents our intention.

pkg/vmcp/server/server.go
```diff
@@ -429,9 +432,21 @@ func (s *Server) Start(ctx context.Context) error {
...
+	var mcpHandler http.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
jerm-dro: nit: can you pull this out into a `headerValidatingMiddleware` function? 

This `server.Start` code already has a lot going on. It's fine to keep the new function in this file.

pkg/vmcp/server/server.go
```diff
@@ -1150,3 +1167,8 @@ func (s *Server) handleBackendHealth(w http.ResponseWriter, _ *http.Request) {
 		logger.Errorf("Failed to write backend health response: %v", err)
 	}
 }
+
+// acceptsEventStream checks whether the request's Accept header includes text/event-stream.
+func acceptsEventStream(r *http.Request) bool {
```
jerm-dro: suggestion: delete this thin function and inline it in the proposed `headerValidatingMiddleware`.

pkg/vmcp/server/server.go
```diff
@@ -487,9 +502,11 @@ func (s *Server) Start(ctx context.Context) error {
 		Handler:           mux,
 		ReadHeaderTimeout: defaultReadHeaderTimeout,
 		ReadTimeout:       defaultReadTimeout,
-		WriteTimeout:      defaultWriteTimeout,
-		IdleTimeout:       defaultIdleTimeout,
-		MaxHeaderBytes:    defaultMaxHeaderBytes,
+		// WriteTimeout is intentionally omitted. SSE connections are long-lived
```
jerm-dro: Question, Blocking: can you write a test to confirm this needs to be removed? From my reading of the `WriteTimeout`, this is actually how long it takes to respond to _each_ request on the stream. 

> // WriteTimeout is the maximum duration before timing out
	// writes of the response. It is reset whenever a new
	// request's header is read. Like ReadTimeout, it does not
	// let Handlers make decisions on a per-request basis.
	// A zero or negative value means there will be no timeout.


In other words, I expect a session can live for more than `30s` if each request on the stream is responded to in less than `30s`.

juancarlosm:
Thanks for the review! Here's what I changed:

- Extracted `headerValidatingMiddleware` as a standalone function
- Moved it to outermost position (executes before recovery and all other middleware)
- Fixed the execution order comment
- Logging the `w.Write` error instead of ignoring it
- Inlined `acceptsEventStream` into the middleware
- Added integration tests (5 cases: GET without Accept, GET with wrong Accept, GET with correct Accept, GET with multiple Accept types, POST without Accept)

On **WriteTimeout**: Go's `WriteTimeout` sets an absolute deadline on the TCP connection via `SetWriteDeadline(time.Now().Add(d))` when request headers are read. It's only cleared after `ServeHTTP` returns. For SSE, the handler never returns — so the stream dies after exactly 30s regardless of write activity. The "reset whenever a new request's header is read" refers to HTTP keep-alive pipelining (separate requests on the same connection), not per-write within a response. This is confirmed by Go maintainers in [golang/go#16100](https://github.com/golang/go/issues/16100). I've added a reference in the commit message.

pkg/vmcp/server/server.go
```diff
@@ -474,10 +476,14 @@ func (s *Server) Start(ctx context.Context) error {
 		logger.Info("Authentication middleware enabled for MCP endpoints")
 	}
 
-	// Apply recovery middleware last (so it executes first and catches panics from all other middleware)
+	// Apply recovery middleware (catches panics from all inner middleware)
```
jerm-dro: blocker: let's put the recovery middleware as the outermost, so it always runs. Even though your code is correct and doesn't contain a panic, I'd prefer not to make exceptions to having this in place.

juancarlosm:
Thanks @jerm-dro for the second round of review!

Changes in this push:

- **Recovery middleware moved to outermost position** — it now wraps everything including header validation, so panics are always caught.
- **WriteTimeout restored** — I've split that change out to keep this PR focused on the header validation fix. I'll open a separate issue/PR to address WriteTimeout and explore `ResponseController` for per-request timeouts.

juancarlosm:
Hi @jerm-dro, any update on this?

jerm-dro (approved):
Sorting out the merge conflicts may clear the approval. Just `@` me and I can approve again.

juancarlosm:
@jerm-dro fixed!

PR #3639, opened 2026-02-05
Add Redis Storage Backend for Auth Server

### LINKED ISSUE #3628: Auth Server: Implement Redis Storage: Core Implementation (Phase 1)
## Description

Implement a Redis Sentinel-backed storage backend for the embedded authorization server that satisfies the existing `Storage` interface. This enables horizontal scaling of ToolHive auth servers by allowing multiple instances to share authentication state via Redis, supporting high-availability deployments with automatic failover.

This is Phase 1 of the Redis Storage feature, focusing on core implementation with unit tests using miniredis for mocking.

## Context

The current `MemoryStorage` implementation stores all OAuth2 tokens, authorization codes, user data, and pending authorizations in memory. This prevents horizontal scaling and causes state loss on restarts. This task adds `RedisStorage` as an alternative backend that implements the same `Storage` interface, enabling distributed deployments.

**Parent Issue**: https://github.com/stacklok/stacklok-epics/issues/197
**Dependencies**: None (this is the first task in the epic)
**Blocks**: TASK-002 (Integration Testing), TASK-003 (Documentation & Operator CRD Updates)

## Acceptance Criteria

- [ ] `RedisStorage` struct implements all methods of the `Storage` interface (verified by compile-time check)
- [ ] Configuration types support Sentinel-only deployment with ACL user authentication
- [ ] Key generation uses hash tags for multi-tenancy (`thv:auth:{ns}:{name}:<type>:<id>`)
- [ ] Unit tests achieve >80% coverage using miniredis
- [ ] Connection health checking implemented via Redis PING
- [ ] `task lint` passes without errors
- [ ] `task test` passes without errors
- [ ] New dependencies added to `go.mod`: `github.com/redis/go-redis/v9` and `github.com/alicebob/miniredis/v2`

Follow the patterns established by `MemoryStorage` in `pkg/authserver/storage/memory.go`. The Redis implementation should:

1. Use `redis.UniversalClient` from go-redis to support Sentinel failover
2. Serialize `fosite.Requester` and other types to JSON for storage
3. Use Redis TTL (SETEX/PSETEX) for automatic expiration instead of background cleanup
4. Implement secondary indexes for request ID lookups required by RFC 7009 token revocation
5. Use Lua scripts for atomic operations that span multiple keys

### Patterns & Frameworks

- **go-redis/v9**: Official Redis client library with built-in Sentinel support and connection pooling
- **miniredis/v2**: In-memory Redis mock for unit testing (no external Redis required)
- **JSON serialization**: Marshal/unmarshal all stored types to JSON (consistent with existing patterns)
- **Interface compliance**: Use compile-time check pattern from `memory.go:1099-1104`
- **Defensive copies**: Follow pattern from `memory.go:700-712` when returning data
- **Error wrapping**: Wrap errors with fosite errors as shown in `memory.go:332-333`

### Code Pointers

- `pkg/authserver/storage/types.go` - `Storage` interface definition (all methods to implement)
- `pkg/authserver/storage/memory.go` - Reference implementation to follow
- `pkg/authserver/storage/config.go` - Add `TypeRedis` constant and Redis configuration types
- `pkg/authserver/storage/memory_test.go` - Test patterns to follow (mock types, table-driven tests)
- `docs/proposals/THV-0035-auth-server-redis-storage.md` - Full RFC with detailed design

### Component Interfaces

**RedisConfig (Runtime Configuration)**

```go
// pkg/authserver/storage/redis.go

// RedisConfig holds Redis connection configuration for runtime use.
type RedisConfig struct {
    // SentinelConfig is required - Sentinel-only deployment
    SentinelConfig *SentinelConfig

    // ACLUserConfig is required - ACL user authentication only
    ACLUserConfig *ACLUserConfig

    // KeyPrefix for multi-tenancy: "thv:auth:{ns}:{name}:"
    KeyPrefix string

    // Timeouts (defaults: Dial=5s, Read=3s, Write=3s)
    DialTimeout  time.Duration
    ReadTimeout  time.Duration
    WriteTimeout time.Duration
}

type SentinelConfig struct {
    MasterName    string
    SentinelAddrs []string
    DB            int
}

type ACLUserConfig struct {
    Username string
    Password string
}
```

**RedisRunConfig (Serializable Configuration)**

```go
// pkg/authserver/storage/config.go

type RedisRunConfig struct {
    DeploymentMode string               `json:"deploymentMode" yaml:"deploymentMode"` // Must be "sentinel"
    SentinelConfig *SentinelRunConfig   `json:"sentinelConfig,omitempty" yaml:"sentinelConfig,omitempty"`
    AuthType       string               `json:"authType" yaml:"authType"` // Must be "aclUser"
    ACLUserConfig  *ACLUserRunConfig    `json:"aclUserConfig,omitempty" yaml:"aclUserConfig,omitempty"`
    KeyPrefix      string               `json:"keyPrefix" yaml:"keyPrefix"`
    DialTimeout    string               `json:"dialTimeout,omitempty" yaml:"dialTimeout,omitempty"`
    ReadTimeout    string               `json:"readTimeout,omitempty" yaml:"readTimeout,omitempty"`
    WriteTimeout   string               `json:"writeTimeout,omitempty" yaml:"writeTimeout,omitempty"`
}

type SentinelRunConfig struct {
    MasterName    string   `json:"masterName" yaml:"masterName"`
    SentinelAddrs []string `json:"sentinelAddrs" yaml:"sentinelAddrs"`
    DB            int      `json:"db,omitempty" yaml:"db,omitempty"`
}

type ACLUserRunConfig struct {
    UsernameEnvVar string `json:"usernameEnvVar" yaml:"usernameEnvVar"`
    PasswordEnvVar string `json:"passwordEnvVar" yaml:"passwordEnvVar"`
}
```

**RedisStorage Factory**

```go
// pkg/authserver/storage/redis.go

type RedisStorage struct {
    client redis.UniversalClient
    config RedisConfig
}

// NewRedisStorage creates Redis-backed storage with Sentinel failover support.
// Returns error if configuration validation fails or connection cannot be established.
func NewRedisStorage(ctx context.Context, cfg RedisConfig) (*RedisStorage, error)

// Close closes the Redis client connection.
func (s *RedisStorage) Close() error

// Ping checks Redis connectivity (health check).
func (s *RedisStorage) Ping(ctx context.Context) error
```

**Key Generation Utilities**

```go
// pkg/authserver/storage/redis_keys.go

// DeriveKeyPrefix creates the key prefix from server namespace and name.
func DeriveKeyPrefix(namespace, name string) string {
    return fmt.Sprintf("thv:auth:{%s}:{%s}:", namespace, name)
}

// Key types as constants for type safety
const (
    KeyTypeAccess       = "access"
    KeyTypeRefresh      = "refresh"
    KeyTypeAuthCode     = "authcode"
    KeyTypePKCE         = "pkce"
    KeyTypeClient       = "client"
    KeyTypeUser         = "user"
    KeyTypeProvider     = "provider"
    KeyTypeUpstream     = "upstream"
    KeyTypePending      = "pending"
    KeyTypeInvalidated  = "invalidated"
    KeyTypeJWT          = "jwt"
    KeyTypeReqIDAccess  = "reqid:access"
    KeyTypeReqIDRefresh = "reqid:refresh"
)
```

**Serializable Requester Wrapper**

```go
// pkg/authserver/storage/redis.go

// storedRequest is a serializable wrapper for fosite.Requester
type storedRequest struct {
    ClientID      string          `json:"client_id"`
    RequestedAt   time.Time       `json:"requested_at"`
    Scopes        []string        `json:"scopes"`
    GrantedScopes []string        `json:"granted_scopes"`
    Form          url.Values      `json:"form"`
    Session       json.RawMessage `json:"session"`
    RequestID     string          `json:"request_id"`
    ExpiresAt     time.Time       `json:"expires_at"`
}
```

### Testing Strategy

**Unit Tests** (using miniredis)

- [ ] Test `NewRedisStorage` with valid and invalid configurations
- [ ] Test connection health check (`Ping`)
- [ ] Test all `Storage` interface methods:
  - [ ] `RegisterClient` / `GetClient` / `ClientAssertionJWTValid` / `SetClientAssertionJWT`
  - [ ] `CreateAuthorizeCodeSession` / `GetAuthorizeCodeSession` / `InvalidateAuthorizeCodeSession`
  - [ ] `CreateAccessTokenSession` / `GetAccessTokenSession` / `DeleteAccessTokenSession`
  - [ ] `CreateRefreshTokenSession` / `GetRefreshTokenSession` / `DeleteRefreshTokenSession`
  - [ ] `RotateRefreshToken`
  - [ ] `RevokeAccessToken` / `RevokeRefreshToken` / `RevokeRefreshTokenMaybeGracePeriod`
  - [ ] `CreatePKCERequestSession` / `GetPKCERequestSession` / `DeletePKCERequestSession`
  - [ ] `StoreUpstreamTokens` / `GetUpstreamTokens` / `DeleteUpstreamTokens`
  - [ ] `StorePendingAuthorization` / `LoadPendingAuthorization` / `DeletePendingAuthorization`
  - [ ] `CreateUser` / `GetUser` / `DeleteUser`
  - [ ] `CreateProviderIdentity` / `GetProviderIdentity` / `UpdateProviderIdentityLastUsed`
  - [ ] `GetUserProviderIdentities`
- [ ] Test key generation utilities
- [ ] Test JSON serialization/deserialization of `storedRequest`
- [ ] Test TTL handling (tokens expire correctly)
- [ ] Test error handling (Redis connection errors, not found, etc.)

**Edge Cases**

- [ ] Empty or nil configuration values
- [ ] Invalid Sentinel addresses
- [ ] Connection failures during operations
- [ ] Expired tokens return appropriate errors
- [ ] Concurrent access patterns (similar to `memory_test.go` concurrent tests)

## Out of Scope

- Integration tests with real Redis (covered in TASK-002)
- Operator CRD types (covered in TASK-003)
- Runner integration to create storage from config (covered in TASK-003)
- TLS/mTLS configuration for Redis connections (future enhancement)
- Custom connection pool configuration (use go-redis defaults)
- Redis persistence (RDB/AOF) - this implementation is memory-only

## References

- Epic: https://github.com/stacklok/stacklok-epics/issues/197
- RFC Design Document: `docs/proposals/THV-0035-auth-server-redis-storage.md`
- Existing Storage Interface: `pkg/authserver/storage/types.go`
- Reference Implementation: `pkg/authserver/storage/memory.go`
- go-redis Documentation: https://redis.uptrace.dev/
- miniredis Documentation: https://github.com/alicebob/miniredis

tgrunnagle:
Closes #3628

## Summary

Implements a Redis Sentinel-backed storage backend for the authorization server's `Storage` interface, enabling horizontal scaling of ToolHive auth servers. Multiple instances can now share authentication state via Redis with automatic failover support. This is Phase 1 of the Redis Storage feature, providing the core implementation with comprehensive unit tests.

### Storage Backend (`pkg/authserver/storage/`)

- Added `RedisStorage` struct implementing all 30+ methods of the `Storage` interface
- Implemented serialization wrappers (`storedSession`, `storedClient`, `storedProviderIdentity`, etc.) for JSON storage
- Used Redis TTL for automatic token expiration instead of background cleanup
- Created secondary indexes for request ID lookups required by RFC 7009 token revocation
- Added Lua script for atomic `UpdateProviderIdentityLastUsed` operation to prevent race conditions
- Implemented compile-time interface compliance check

### Key Generation (`pkg/authserver/storage/redis_keys.go`)

- Added `DeriveKeyPrefix` function using hash tag format `thv:auth:{ns:name}:` for Redis Cluster slot co-location
- Defined key type constants for all stored data types (access tokens, refresh tokens, auth codes, PKCE, clients, users, etc.)
- Added helper functions for consistent key generation

### Configuration (`pkg/authserver/storage/config.go`)

- Added `TypeRedis` storage type constant
- Added `RedisRunConfig` and related types for serializable configuration (Sentinel addresses, ACL credentials, timeouts)
- Credentials read from environment variables for security

### Dependencies (`go.mod`)

- Added `github.com/redis/go-redis/v9` for Redis client with Sentinel support
- Added `github.com/alicebob/miniredis/v2` for unit testing

## Implementation Details

- Sentinel-only deployment mode enforced (standalone Redis not supported for HA requirements)
- ACL user authentication required (legacy AUTH not supported for security)
- Hash tag format `{ns:name}` combines namespace and name to ensure all keys for a server land in the same Redis Cluster slot
- TTL set on secondary index sets to prevent memory growth from orphaned indexes
- Form field serialization preserves multi-value support (`url.Values` stored as `map[string][]string`)

## Testing

- Comprehensive unit tests using miniredis (no external Redis required)
- 85.2% code coverage (exceeds 80% requirement)
- Tests cover all Storage interface methods, error handling, TTL behavior, and concurrent access
- Edge cases tested: expired tokens, non-existent resources, duplicate creation, input validation

## Additional Notes

- Future work: Integration tests with real Redis (Phase 2), Operator CRD updates (Phase 3)
- Parent epic: stacklok/stacklok-epics#197

## Large PR Justification

- Implements one interface and corresponding unit tests. No integration with other components in this PR.

### DIFF
--- a/pkg/authserver/storage/config.go
+++ b/pkg/authserver/storage/config.go
@@ -65,9 +65,6 @@ type RunConfig struct {
 // RedisRunConfig is the serializable Redis configuration for RunConfig.
 // This is designed for Sentinel-only deployments with ACL user authentication.
 type RedisRunConfig struct {
-	// DeploymentMode must be "sentinel" - only Sentinel deployments are supported.
-	DeploymentMode string `json:"deploymentMode" yaml:"deploymentMode"`
-
 	// SentinelConfig contains Sentinel-specific configuration.
 	SentinelConfig *SentinelRunConfig `json:"sentinelConfig,omitempty" yaml:"sentinelConfig,omitempty"`
 
--- a/pkg/authserver/storage/redis.go
+++ b/pkg/authserver/storage/redis.go
@@ -81,9 +81,8 @@ type ACLUserConfig struct {
 // It provides distributed storage for OAuth2 tokens, authorization codes,
 // user data, and pending authorizations, enabling horizontal scaling.
 type RedisStorage struct {
-	client         redis.UniversalClient
-	keyPrefix      string
-	sessionFactory session.Factory
+	client    redis.UniversalClient
+	keyPrefix string
 }
 
 // storedSession is a serializable wrapper for fosite.Requester.
@@ -137,9 +136,8 @@ func NewRedisStorage(ctx context.Context, cfg RedisConfig) (*RedisStorage, error
 	}
 
 	return &RedisStorage{
-		client:         client,
-		keyPrefix:      cfg.KeyPrefix,
-		sessionFactory: defaultSessionFactory,
+		client:    client,
+		keyPrefix: cfg.KeyPrefix,
 	}, nil
 }
 
@@ -147,9 +145,8 @@ func NewRedisStorage(ctx context.Context, cfg RedisConfig) (*RedisStorage, error
 // This is useful for testing with miniredis.
 func NewRedisStorageWithClient(client redis.UniversalClient, keyPrefix string) *RedisStorage {
 	return &RedisStorage{
-		client:         client,
-		keyPrefix:      keyPrefix,
-		sessionFactory: defaultSessionFactory,
+		client:    client,
+		keyPrefix: keyPrefix,
 	}
 }
 
@@ -183,17 +180,17 @@ func validateConfig(cfg *RedisConfig) error {
 	return nil
 }
 
-// Close closes the Redis client connection.
-func (s *RedisStorage) Close() error {
-	return s.client.Close()
-}
-
-// Ping checks Redis connectivity (health check).
-func (s *RedisStorage) Ping(ctx context.Context) error {
+// Health checks Redis connectivity.
+func (s *RedisStorage) Health(ctx context.Context) error {
 	if err := s.client.Ping(ctx).Err(); err != nil {
 		return fmt.Errorf("redis health check failed: %w", err)
 	}
 	return nil
+}
+
+// Close closes the Redis client connection.
+func (s *RedisStorage) Close() error {
+	return s.client.Close()
 }
 
 // -----------------------
@@ -1335,7 +1332,7 @@ func unmarshalRequester(ctx context.Context, data []byte, s *RedisStorage) (fosi
 
 	// Create a session prototype via factory, then deserialize the full session
 	// blob into it. This preserves JWT claims, headers, and upstream session IDs.
-	sess := s.sessionFactory("", "", "")
+	sess := defaultSessionFactory("", "", "")
 	if len(stored.Session) > 0 {
 		if err := json.Unmarshal(stored.Session, sess); err != nil {
 			return nil, fmt.Errorf("failed to unmarshal session data: %w", err)
--- a/pkg/authserver/storage/memory.go
+++ b/pkg/authserver/storage/memory.go
@@ -139,6 +139,11 @@ func NewMemoryStorage(opts ...MemoryStorageOption) *MemoryStorage {
 	return s
 }
 
+// Health is a no-op for in-memory storage since it is always available.
+func (*MemoryStorage) Health(_ context.Context) error {
+	return nil
+}
+
 // Close stops the background cleanup goroutine and waits for it to finish.
 // This should be called when the storage is no longer needed.
 func (s *MemoryStorage) Close() error {
--- a/pkg/authserver/storage/mocks/mock_storage.go
+++ b/pkg/authserver/storage/mocks/mock_storage.go
@@ -714,6 +714,20 @@ func (mr *MockStorageMockRecorder) GetUserProviderIdentities(ctx, userID any) *g
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetUserProviderIdentities", reflect.TypeOf((*MockStorage)(nil).GetUserProviderIdentities), ctx, userID)
 }
 
+// Health mocks base method.
+func (m *MockStorage) Health(ctx context.Context) error {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "Health", ctx)
+	ret0, _ := ret[0].(error)
+	return ret0
+}
+
+// Health indicates an expected call of Health.
+func (mr *MockStorageMockRecorder) Health(ctx any) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Health", reflect.TypeOf((*MockStorage)(nil).Health), ctx)
+}
+
 // InvalidateAuthorizeCodeSession mocks base method.
 func (m *MockStorage) InvalidateAuthorizeCodeSession(ctx context.Context, code string) error {
 	m.ctrl.T.Helper()
--- a/pkg/authserver/storage/redis_test.go
+++ b/pkg/authserver/storage/redis_test.go
@@ -1273,14 +1273,14 @@ func TestRedisKeyGeneration(t *testing.T) {
 
 // --- Health Check Tests ---
 
-func TestRedisStorage_Ping(t *testing.T) {
+func TestRedisStorage_Health(t *testing.T) {
 	withRedisStorage(t, func(ctx context.Context, s *RedisStorage, _ *miniredis.Miniredis) {
-		err := s.Ping(ctx)
+		err := s.Health(ctx)
 		require.NoError(t, err)
 	})
 }
 
-func TestRedisStorage_Ping_ConnectionFailure(t *testing.T) {
+func TestRedisStorage_Health_ConnectionFailure(t *testing.T) {
 	t.Parallel()
 
 	mr := miniredis.RunT(t)
@@ -1292,6 +1292,6 @@ func TestRedisStorage_Ping_ConnectionFailure(t *testing.T) {
 	// Close the miniredis server to simulate connection failure
 	mr.Close()
 
-	err := storage.Ping(context.Background())
+	err := storage.Health(context.Background())
 	require.Error(t, err)
 }
--- a/pkg/authserver/storage/types.go
+++ b/pkg/authserver/storage/types.go
@@ -329,6 +329,10 @@ type Storage interface {
 	// token exchange. Keyed by the same code/signature as the authorization code.
 	pkce.PKCERequestStorage
 
+	// Health checks connectivity to the storage backend.
+	// Returns nil if the storage is healthy and reachable.
+	Health(ctx context.Context) error
+
 	// Close releases any resources held by the storage implementation.
 	// This should be called when the storage is no longer needed.
 	Close() error


pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+	}
+
+	return identities, nil
+}
+
+// -----------------------
+// Serialization Helpers
+// -----------------------
+
+// marshalRequester serializes a fosite.Requester to JSON.
```
jhrozek, line 1292: We verified that fosite's handlers call `Sanitize()` on the request before passing it to storage — same approach as Ory Hydra. The form data here is already stripped of sensitive fields (client_secret, password, code_verifier). Checked all code paths in the auth server: no direct calls to Create* methods bypass fosite's handler chain.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+		cfg.DialTimeout = DefaultDialTimeout
+	}
+	if cfg.ReadTimeout == 0 {
+		cfg.ReadTimeout = DefaultReadTimeout
+	}
+	if cfg.WriteTimeout == 0 {
+		cfg.WriteTimeout = DefaultWriteTimeout
+	}
+
+	client := redis.NewFailoverClient(&redis.FailoverOptions{
```
jhrozek, line 120: There's no TLS configuration here. Are we planning to add TLS support as a follow-up? For production Sentinel deployments this would be needed.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+	if string(data) == nullMarker {
+		return nil, nil
+	}
+
+	var stored storedUpstreamTokens
+	if err := json.Unmarshal(data, &stored); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal upstream tokens: %w", err)
+	}
+
+	expiresAt := time.Unix(stored.ExpiresAt, 0)
```
jhrozek: Bug: when `ExpiresAt` is zero (upstream tokens with no expiry), `time.Unix(0, 0)` produces 1970-01-01, so the check on line 792 always returns `ErrExpired`. The write path in `marshalUpstreamTokensWithTTL` treats zero as "use default TTL", but the read path here rejects it. Need to either skip the expiry check when `stored.ExpiresAt == 0` or store a sentinel value that means "no expiry".

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+	data, err := marshalRequester(request)
+	if err != nil {
+		return fmt.Errorf("failed to marshal request: %w", err)
+	}
+
+	return s.client.Set(ctx, key, data, ttl).Err()
+}
+
+// GetAuthorizeCodeSession retrieves the authorization request for a given code.
+func (s *RedisStorage) GetAuthorizeCodeSession(ctx context.Context, code string, _ fosite.Session) (fosite.Requester, error) {
```
jhrozek, line 335: There's a TTL gap between the auth code (10 min) and the invalidation marker (30 min). If the auth code key expires but the invalidation marker is still alive, a replayed code gets `ErrNotFound` instead of `ErrInvalidatedAuthorizeCode`. Fosite treats these differently — `ErrInvalidatedAuthorizeCode` triggers token revocation (the replay attack protection from RFC 6819), while `ErrNotFound` just fails the request, leaving the previously issued tokens alive.

One fix: check the invalidation marker first (you already do), and if it exists, return `ErrInvalidatedAuthorizeCode` immediately without trying to GET the auth code key. That way the marker works correctly even after the code itself expires.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+
+	key := redisKey(s.keyPrefix, KeyTypeAccess, signature)
+	ttl := getTTLFromRequester(request, fosite.AccessToken, DefaultAccessTokenTTL)
+
+	data, err := marshalRequester(request)
+	if err != nil {
+		return fmt.Errorf("failed to marshal request: %w", err)
+	}
+
+	// Store the token
```
jhrozek: The three Redis operations here (SET, SADD, EXPIRE) are not atomic. If SADD or EXPIRE fails, the compensating deletes try to clean up, but those can also fail leaving the store in an inconsistent state. Consider using a Redis pipeline (`TxPipeline`) to execute all three as a single round-trip. Same pattern applies to `CreateRefreshTokenSession`.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+	client, err := s.GetClient(ctx, stored.ClientID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get client for session: %w", err)
+	}
+
+	// Convert form values back (stored.Form is map[string][]string, same as url.Values)
+	form := url.Values(stored.Form)
+
+	// Create session with expiration times
+	session := &redisSession{
```
jhrozek: `marshalRequester` captures all the requester fields correctly, but from the session it only extracts `Subject` and `ExpiresAt` (the only data accessible through the `fosite.Session` interface). On the read side, `unmarshalRequester` reconstructs a `redisSession` that implements `fosite.Session` but not `oauth2.JWTSessionContainer`.

This breaks token refresh. The flow is: `GetRefreshTokenSession` loads the stored session -> fosite clones it onto the new request (`flow_refresh.go:87`) -> `PopulateTokenEndpointResponse` generates new tokens -> `DefaultJWTStrategy.generate()` does `requester.GetSession().(JWTSessionContainer)` (`strategy_jwt.go:112`). Since `redisSession` doesn't implement `JWTSessionContainer`, this type assertion fails and refresh returns an error.

Beyond the type assertion, the JWT claims Extra map (`tsid`, `client_id`) and `UpstreamSessionID` are lost, so even if you worked around the assertion, refreshed tokens would be missing claims and upstream token exchange would break.

The standard fosite pattern (used by Hydra's SQL storage) is to serialize the entire session as a JSON blob: `json.Marshal(request.GetSession())` on write, `json.Unmarshal(blob, sessionPrototype)` on read. The session prototype comes from the `fosite.Session` parameter that all `Get*Session` methods receive — currently discarded as `_ fosite.Session` on lines 297, 392, 484, 624. `oauth2.JWTSession` round-trips cleanly with Go's default JSON encoding.

The `session.UpstreamSession` interface (which embeds `JWTSessionContainer`) and the compile-time check `var _ UpstreamSession = (*Session)(nil)` already exist in the session package to catch this kind of mismatch. With this fix, `redisSession` and `redisRequester` can be removed entirely — you return `fosite.Request` directly with a real `*session.Session` that satisfies all the required interfaces.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+		if ttl < 0 {
+			ttl = DefaultAccessTokenTTL
+		}
+	}
+
+	return data, ttl, nil
+}
+
+// StoreUpstreamTokens stores the upstream IDP tokens for a session.
+func (s *RedisStorage) StoreUpstreamTokens(ctx context.Context, sessionID string, tokens *UpstreamTokens) error {
```
jhrozek, line 780: Upstream IDP tokens are stored as plaintext JSON. For a first implementation this is fine, but worth noting that Hydra has an optional `EncryptSessionData` feature that encrypts session blobs before persisting them. Could be worth considering as a follow-up for tokens at rest.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+	if cfg.SentinelConfig == nil {
+		return errors.New("sentinel configuration is required")
+	}
+	if cfg.SentinelConfig.MasterName == "" {
+		return errors.New("sentinel master name is required")
+	}
+	if len(cfg.SentinelConfig.SentinelAddrs) == 0 {
+		return errors.New("at least one sentinel address is required")
+	}
+	if cfg.ACLUserConfig == nil {
```
jhrozek, line 168: This checks that `ACLUserConfig` is non-nil but doesn't validate that `Username` and `Password` are non-empty. Passing `&ACLUserRunConfig{}` with blank strings would silently connect with empty credentials.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+	RequestedAt        time.Time           `json:"requested_at"`
+	RequestedScopes    []string            `json:"requested_scopes"`
+	GrantedScopes      []string            `json:"granted_scopes"`
+	RequestedAudience  []string            `json:"requested_audience"`
+	GrantedAudience    []string            `json:"granted_audience"`
+	Form               map[string][]string `json:"form"`
+	RequestID          string              `json:"request_id"`
+	Subject            string              `json:"subject"`
+	ExpiresAt          map[string]int64    `json:"expires_at"`
+	AccessTokenExpiry  int64               `json:"access_token_expiry,omitempty"`
```
jhrozek: `AccessTokenExpiry`, `RefreshTokenExpiry`, and `AuthCodeExpiry` are never written or read anywhere. Looks like leftovers from before the `ExpiresAt` map approach was adopted. Should be removed to avoid confusion.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1407 @@
...
+// updateUpstreamUserSets handles adding/removing upstream token references in user sets.
+func (s *RedisStorage) updateUpstreamUserSets(ctx context.Context, key, oldUserID string, tokens *UpstreamTokens) error {
+	newUserID := ""
+	if tokens != nil {
+		newUserID = tokens.UserID
+	}
+
+	// Clean up old user's set if UserID changed
+	if oldUserID != "" && oldUserID != newUserID {
+		oldUserUpstreamSetKey := redisSetKey(s.keyPrefix, "user:upstream", oldUserID)
```
jhrozek: `"user:upstream"` and `"user:providers"` are used as hardcoded strings in 7 places across this file, but `redis_keys.go` already defines constants for other key types. Should add `KeyTypeUserUpstream` and `KeyTypeUserProviders` constants for consistency.

pkg/authserver/storage/redis.go
tgrunnagle: Tracked by https://github.com/stacklok/toolhive/issues/3681

pkg/authserver/storage/redis.go
tgrunnagle: Tokens are not persisted to disk by design

pkg/authserver/storage/config.go
```diff
@@ -61,4 +57,57 @@ func DefaultConfig() *Config {
...
+
+	// RedisConfig is the Redis-specific configuration when Type is "redis".
+	RedisConfig *RedisRunConfig `json:"redisConfig,omitempty" yaml:"redisConfig,omitempty"`
+}
+
+// RedisRunConfig is the serializable Redis configuration for RunConfig.
+// This is designed for Sentinel-only deployments with ACL user authentication.
+type RedisRunConfig struct {
+	// DeploymentMode must be "sentinel" - only Sentinel deployments are supported.
+	DeploymentMode string `json:"deploymentMode" yaml:"deploymentMode"`
```
jhrozek: This is for future extensibility?

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1389 @@
...
+	return nil
+}
+
+// Close closes the Redis client connection.
+func (s *RedisStorage) Close() error {
+	return s.client.Close()
+}
+
+// Ping checks Redis connectivity (health check).
+func (s *RedisStorage) Ping(ctx context.Context) error {
```
jhrozek: Ha, interesting, do you think we should have a Ping (or maybe a more generic Health()) method on the generic storage interface? In the memory case it could just return nil/noop.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1389 @@
...
+	Password string
+}
+
+// RedisStorage implements the Storage interface with Redis Sentinel backend.
+// It provides distributed storage for OAuth2 tokens, authorization codes,
+// user data, and pending authorizations, enabling horizontal scaling.
+type RedisStorage struct {
+	client         redis.UniversalClient
+	keyPrefix      string
+	sessionFactory session.Factory
```
jhrozek: initially I thought the session factory was to be used in tests, but it doesn't seem to be the case - it's fine to leave as it, not a big deal, mostly wondering if you have a future use in mind.

pkg/authserver/storage/redis.go
```diff
@@ -0,0 +1,1389 @@
...
+	data, err := json.Marshal(stored)
+	if err != nil {
+		return fmt.Errorf("failed to marshal client: %w", err)
+	}
+
+	// Public clients (from DCR) expire to prevent unbounded growth.
+	// Confidential clients don't expire.
+	ttl := time.Duration(0)
+	if client.IsPublic() {
+		ttl = DefaultPublicClientTTL
```
jhrozek, line 250: I think this is fine, I'd like to make a separate pass (with an RFC and a discussion) on how do we handle DCR clients - I suspect we'll want the number configurable at least.

pkg/authserver/storage/redis.go
tgrunnagle: I was thinking we could call it during initialization for an early check. Will add it to the interface.

pkg/authserver/storage/redis.go
tgrunnagle: It was originally meant for tests, but you're right tests are just using defaultSessionFactory. Will remove.

pkg/authserver/storage/config.go
tgrunnagle: Yes, but I'll remove it for simplicity

PR #3642, opened 2026-02-06
Exclude unhealthy backends from vMCP capability aggregation

yrobla:
Implement backend health filtering to prevent exposing tools from unavailable backends when circuit breakers are open, addressing issue #3036 acceptance criterion.

When backends become unhealthy (circuit breaker open, authentication failures, unknown status), their tools are automatically excluded from the aggregated capabilities returned to clients. When backends recover, tools are automatically restored on the next session initialization.

Changes:
  - Add filterHealthyBackends() in discovery middleware to exclude unhealthy/unknown/unauthenticated backends from aggregation
  - Include healthy and degraded backends (operational but slow)
  - Move circuit breaker and health check validation from commands.go to pkg/vmcp/config/validator.go for better separation of concerns
  - Add comprehensive unit tests for backend filtering (8 test cases)
  - Add validation tests for failure handling config (11 test cases)
  - Update E2E circuit breaker test with capability filtering notes

Related-to: #3036

### DIFF
--- a/pkg/vmcp/discovery/middleware_test.go
+++ b/pkg/vmcp/discovery/middleware_test.go
@@ -134,7 +134,7 @@ func TestMiddleware_InitializeRequest(t *testing.T) {
 
 	// Wrap handler with middleware
 	backendRegistry := vmcp.NewImmutableRegistry(backends)
-	middleware := Middleware(mockMgr, backendRegistry, createTestSessionManager(t))
+	middleware := Middleware(mockMgr, backendRegistry, createTestSessionManager(t), nil)
 	wrappedHandler := middleware(testHandler)
 
 	// Create initialize request (no session ID header)
@@ -204,7 +204,7 @@ func TestMiddleware_SubsequentRequest_SkipsDiscovery(t *testing.T) {
 
 	// Wrap handler with middleware
 	backendRegistry := vmcp.NewImmutableRegistry(backends)
-	middleware := Middleware(mockMgr, backendRegistry, sessionMgr)
+	middleware := Middleware(mockMgr, backendRegistry, sessionMgr, nil)
 	wrappedHandler := middleware(testHandler)
 
 	// Create subsequent request (with session ID header)
@@ -245,7 +245,7 @@ func TestMiddleware_DiscoveryTimeout(t *testing.T) {
 	})
 
 	backendRegistry := vmcp.NewImmutableRegistry(backends)
-	middleware := Middleware(mockMgr, backendRegistry, createTestSessionManager(t))
+	middleware := Middleware(mockMgr, backendRegistry, createTestSessionManager(t), nil)
 	wrappedHandler := middleware(testHandler)
 
 	// Initialize request (no session ID) - discovery should happen
@@ -286,7 +286,7 @@ func TestMiddleware_DiscoveryFailure(t *testing.T) {
 	})
 
 	backendRegistry := vmcp.NewImmutableRegistry(backends)
-	middleware := Middleware(mockMgr, backendRegistry, createTestSessionManager(t))
+	middleware := Middleware(mockMgr, backendRegistry, createTestSessionManager(t), nil)
 	wrappedHandler := middleware(testHandler)
 
 	// Initialize request (no session ID) - discovery should happen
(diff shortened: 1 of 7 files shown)


JAORMX:
Is aggregation done periodically? What if there's a temporary downtime in one of the servers? will they be excluded?

yrobla:
> Is aggregation done periodically? What if there's a temporary downtime in one of the servers? will they be excluded?

No, aggregation happens at MCP initialization, not periodically. Capability aggregation happens during MCP protocol initialization (initialize request), then results are cached

Once capabilities are discovered at initialization, they're cached. If a backend
goes down after initialization, its tools remain in the capability list but will fail when
called. The health filtering only prevents unhealthy backends from being included at initialization time.

I can change to periodical in a follow-up

jerm-dro:
> > excluded?
> 
> No, aggregation happens at MCP initialization, not periodically. Capability aggregation happens during MCP protocol initialization (initialize request), then results are cached
> 
> Once capabilities are discovered at initialization, they're cached. If a backend goes down after initialization, its tools remain in the capability list but will fail when called. The health filtering only prevents unhealthy backends from being included at initialization time.
> 
> I can change to periodical in a follow-up

@yrobla This is correct. However, I don't think we should to do this immediately in a follow-up. This has always been a gap. For example, if the underlying MCPServer capabilities change, the vMCP will not update the capabilities within a session. 

Fixing this will require mutating session-scoped state and notifying the MCP client. We can add this functionality today, but I think we'd be better off improving the existing session code first (i.e. this WIP RFC: https://github.com/stacklok/toolhive-rfcs/issues/39). What do you think?

yrobla:
> > > excluded?
> > 
> > 
> > No, aggregation happens at MCP initialization, not periodically. Capability aggregation happens during MCP protocol initialization (initialize request), then results are cached
> > Once capabilities are discovered at initialization, they're cached. If a backend goes down after initialization, its tools remain in the capability list but will fail when called. The health filtering only prevents unhealthy backends from being included at initialization time.
> > I can change to periodical in a follow-up
> 
> @yrobla This is correct. However, I don't think we should to do this immediately in a follow-up. This has always been a gap. For example, if the underlying MCPServer capabilities change, the vMCP will not update the capabilities within a session.
> 
> Fixing this will require mutating session-scoped state and notifying the MCP client. We can add this functionality today, but I think we'd be better off improving the existing session code first (i.e. this WIP RFC: [stacklok/toolhive-rfcs#39](https://github.com/stacklok/toolhive-rfcs/issues/39)). What do you think?

I can see some relationship if we check aggregation at session level, instead of doing that at initialize level. This will help. But also, this will not be periodical, is that what you meant relating that to sesions?

jerm-dro:
> 
> I can see some relationship if we check aggregation at session level, instead of doing that at initialize level. This will help. But also, this will not be periodical, is that what you meant relating that to sesions?

I think so, but to add a bit more detail I'll revise the domain model I had suggested:

```go
// Session represents an active MCP session with all its capabilities and resources.
// This is a pure domain object - no protocol concerns.
type Session interface {
    ID() string

    // Capabilities - returns discovered tools/resources for this session
    // PREVIOIUSLY, capabilities were static.
    // Tools() []Tool
    // Resources() []Resource
    
    // Capabilities() returns a channel containing the up to date capabilities within the session.
    // The channel is guaranteed to hold the initial set of capabilities on construction.
    // The channel will receive updated capabilities if the underlying MCP capabilities change
    // (e.g. circuit breaker, backend revised capabilities). 
    // The channel capacity is 1 to guarantee updates are consumed in order.
    Capabilities() <-chan Capabilities

    // MCP operations - routing logic is encapsulated here
    CallTool(ctx context.Context, name string, arguments map[string]any) (*ToolResult, error)
    ReadResource(ctx context.Context, uri string) (*ResourceResult, error)
    GetPrompt(ctx context.Context, name string, arguments map[string]any) (*PromptResult, error)

    // Lifecycle
    Close() error
}
```

The `Capabilities` method would encapsulate the "how do we recompute capabilities." Periodically re-running the same function in a goroutine would be fine. The caller handles the "how do we notify the MCP client of updated capabilities."

pkg/vmcp/discovery/middleware_test.go
```diff
@@ -517,3 +517,176 @@ func TestMiddleware_ContextTimeoutHandling(t *testing.T) {
...
 	assert.Equal(t, http.StatusGatewayTimeout, rec.Code)
 }
+
+func TestFilterHealthyBackends(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name             string
+		backends         []vmcp.Backend
+		expectedCount    int
```
jerm-dro, line 527: nit: `expectedCount` is redundant with `expectedBackends`. Delete `expectedCount`

PR #3658, opened 2026-02-06
Add SigV4 request signing for AWS API authentication

jhrozek:
This implements the requestSigner component that signs HTTP requests using AWS Signature Version 4. The signer hashes the request body, adds required headers (Authorization, X-Amz-Date, X-Amz-Security-Token), and ensures the request body remains readable after signing. A 10MB payload limit prevents memory exhaustion from oversized requests.

Related: #3567

### DIFF
--- a/pkg/auth/awssts/signer.go
+++ b/pkg/auth/awssts/signer.go
@@ -21,6 +21,10 @@ import (
 // maxPayloadSize is the maximum request body size (10 MB) for SigV4 signing.
 const maxPayloadSize = 10 * 1024 * 1024
 
+// emptySHA256 is the well-known SHA-256 hash of an empty string, used for
+// SigV4 signing of requests with no body.
+const emptySHA256 = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+
 // defaultService is the AWS service name used in SigV4 signing for AWS MCP Server.
 // This value appears in the credential scope of the Authorization header:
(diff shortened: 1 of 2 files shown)


pkg/auth/awssts/signer.go
```diff
@@ -0,0 +1,152 @@
...
+//
+// Returns:
+//   - payloadHash: Hex-encoded SHA-256 hash of the body
+//   - bodyBytes: The body content (for replacing the consumed reader)
+//   - error: Any error reading the body
+func (*requestSigner) hashPayload(req *http.Request) (string, []byte, error) {
+	// Handle empty body
+	if req.Body == nil || req.Body == http.NoBody {
+		// SHA-256 of empty string
+		return "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855", nil, nil
```
JAORMX: @jhrozek I trust you but I am afraid of hallucinations. Can we move this to a constant and have a test that ensures it's indeed said hash on an empty string?

jhrozek: good idea. btw I also went wtf when claude wrote this code, so I checked it with [prometheus' signing code ](https://github.com/prometheus/sigv4/blob/6b687b3b8374b75ee42777e9f200abe62c8b52d7/sigv4.go#L137) (which sadly can't be reused because they use a round-tripper which seemed like it's an overkill for us).

PR #3659, opened 2026-02-06
Add AWS STS token exchange for OIDC-to-AWS credential conversion

jhrozek:
This implements the Exchanger component that converts OIDC identity tokens into temporary AWS credentials using STS AssumeRoleWithWebIdentity. The exchanger uses regional STS endpoints for lower latency and maps STS API errors to package-level sentinel errors for consistent error handling.

Related: #3567

### DIFF
--- a/pkg/auth/awssts/exchange.go
+++ b/pkg/auth/awssts/exchange.go
@@ -6,10 +6,13 @@ package awssts
 import (
 	"context"
 	"fmt"
+	"regexp"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
+
+	"github.com/stacklok/toolhive/pkg/logger"
 )
 
 // STSClient defines the interface for STS operations, enabling mock injection for testing.
@@ -27,8 +30,6 @@ type Exchanger struct {
 }
 
 // NewExchanger creates a new Exchanger with a regional STS client.
-//
-// It uses regional endpoints (https://sts.{region}.amazonaws.com) for lower latency.
 func NewExchanger(ctx context.Context, region string) (*Exchanger, error) {
 	if region == "" {
 		return nil, ErrMissingRegion
@@ -42,19 +43,18 @@ func NewExchanger(ctx context.Context, region string) (*Exchanger, error) {
 	return &Exchanger{client: client}, nil
 }
 
-// newRegionalSTSClient creates an STS client configured for the regional endpoint.
-// Using regional endpoints (https://sts.{region}.amazonaws.com) provides lower latency
-// compared to the global endpoint.
+// newRegionalSTSClient creates an STS client configured for the specified region.
+// The SDK automatically resolves regional STS endpoints for lower latency.
 func newRegionalSTSClient(ctx context.Context, region string) (STSClient, error) {
-	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(region))
+	cfg, err := config.LoadDefaultConfig(ctx,
+		config.WithRegion(region),
+		config.WithCredentialsProvider(aws.AnonymousCredentials{}),
+	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load AWS config: %w", err)
 	}
 
-	regionalEndpoint := fmt.Sprintf("https://sts.%s.amazonaws.com", region)
-	return sts.NewFromConfig(cfg, func(o *sts.Options) {
-		o.BaseEndpoint = aws.String(regionalEndpoint)
-	}), nil
+	return sts.NewFromConfig(cfg), nil
 }
 
 // ExchangeToken performs AssumeRoleWithWebIdentity to exchange an identity token
@@ -64,7 +64,7 @@ func (e *Exchanger) ExchangeToken(
 	token, roleArn, sessionName string,
 	durationSeconds int32,
 ) (*aws.Credentials, error) {
-	if err := validateInputs(token, roleArn, durationSeconds); err != nil {
+	if err := validateInputs(token, roleArn, sessionName, durationSeconds); err != nil {
 		return nil, err
 	}
 
@@ -77,11 +77,12 @@ func (e *Exchanger) ExchangeToken(
 
 	output, err := e.client.AssumeRoleWithWebIdentity(ctx, input)
 	if err != nil {
-		return nil, fmt.Errorf("STS AssumeRoleWithWebIdentity failed: %w", err)
+		logger.Debugf("STS AssumeRoleWithWebIdentity failed: %v", err)
+		return nil, ErrSTSExchangeFailed
 	}
 
-	if output.Credentials == nil {
-		return nil, fmt.Errorf("STS returned nil credentials")
+	if output == nil || output.Credentials == nil {
+		return nil, ErrSTSNilCredentials
 	}
 
 	return &aws.Credentials{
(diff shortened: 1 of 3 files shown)


pkg/auth/awssts/exchange.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return &Exchanger{client: client}, nil
+}
+
+// newRegionalSTSClient creates an STS client configured for the regional endpoint.
+// Using regional endpoints (https://sts.{region}.amazonaws.com) provides lower latency
+// compared to the global endpoint.
+func newRegionalSTSClient(ctx context.Context, region string) (STSClient, error) {
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(region))
```
ChrisJBurns: I don't think you need to call `config.LoadDefaultConfig` here for `AssumeRoleWithWebIdentity`?

[`AssumeRoleWithWebIdentity`](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html) is designed to work out pre-existing AWS creds, the web-id token is the auth. Calling `LoadDefaultConfig` does a call to the internal metadata server (IMDS/IMDSv2) reads .aws/credentials, then reads env vars that aren't relevant. You should be able to use an anonymous creds here.

Something like

```go
cfg, err := config.LoadDefaultConfig(ctx,
      config.WithRegion(region),
      config.WithCredentialsProvider(aws.AnonymousCredentials{}),
  )
```

pkg/auth/awssts/exchange.go
```diff
@@ -0,0 +1,115 @@
...
+	token, roleArn, sessionName string,
+	durationSeconds int32,
+) (*aws.Credentials, error) {
+	if err := validateInputs(token, roleArn, durationSeconds); err != nil {
+		return nil, err
+	}
+
+	input := &sts.AssumeRoleWithWebIdentityInput{
+		RoleArn:          aws.String(roleArn),
+		RoleSessionName:  aws.String(sessionName),
```
ChrisJBurns, line 73: nit: could we validate the session names? I believe they have to be a [certain length](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html#API_AssumeRoleWithWebIdentity_RequestParameters) and characters

pkg/auth/awssts/exchange.go
```diff
@@ -0,0 +1,115 @@
...
+// newRegionalSTSClient creates an STS client configured for the regional endpoint.
+// Using regional endpoints (https://sts.{region}.amazonaws.com) provides lower latency
+// compared to the global endpoint.
+func newRegionalSTSClient(ctx context.Context, region string) (STSClient, error) {
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(region))
+	if err != nil {
+		return nil, fmt.Errorf("failed to load AWS config: %w", err)
+	}
+
+	regionalEndpoint := fmt.Sprintf("https://sts.%s.amazonaws.com", region)
```
ChrisJBurns: Probs a nit: is there a SSRF risk here? an unvalidated region value like `evil.com/foo?x=` as region could produce `https://sts.evil.com/foo?x=.amazonaws.com` which the HTTP client could follow. There will likely be TLS cert validation so this is probably a nothing burger. I'm not too deep on the details of the caller, so i doubt a user is specifying this directly which makes it even more of a nothing burger. Alternatively, we could use the SDK's built-in endpoint resolution by removing the BaseEndpoint override entirely.  the SDK will use regional endpoints when `config.WithRegion` is set.

pkg/auth/awssts/exchange.go
```diff
@@ -0,0 +1,115 @@
...
+		SessionToken:    aws.ToString(output.Credentials.SessionToken),
+		Expires:         aws.ToTime(output.Credentials.Expiration),
+		CanExpire:       true,
+	}, nil
+}
+
+// validateInputs validates the exchange inputs.
+func validateInputs(token, roleArn string, durationSeconds int32) error {
+	if token == "" {
+		return fmt.Errorf("token is required")
```
ChrisJBurns: nit: a couple of bare errors here, and they can't be matched with `errors.Is()`. the rest of the package uses sentinel errors, wondering if its worth doing so in this function too?

pkg/auth/awssts/exchange_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/aws/aws-sdk-go-v2/service/sts/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// mockSTSClient implements STSClient for testing.
+type mockSTSClient struct {
```
ChrisJBurns, line 20: nit: i assume this was intended, but any reason we didn't use mockgen here? i assume simplicity and overkill

pkg/auth/awssts/exchange.go
```diff
@@ -0,0 +1,115 @@
...
+	input := &sts.AssumeRoleWithWebIdentityInput{
+		RoleArn:          aws.String(roleArn),
+		RoleSessionName:  aws.String(sessionName),
+		WebIdentityToken: aws.String(token),
+		DurationSeconds:  aws.Int32(durationSeconds),
+	}
+
+	output, err := e.client.AssumeRoleWithWebIdentity(ctx, input)
+	if err != nil {
+		return nil, fmt.Errorf("STS AssumeRoleWithWebIdentity failed: %w", err)
```
ChrisJBurns: Yeah so you'd probably not want to wrap errors here. STS errors can contain token fragments and role ARN details. I'd consider mapping to generic errors and logging the error itself at DEBUG.

pkg/auth/awssts/exchange.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	if output.Credentials == nil {
```
ChrisJBurns: nit: although typically the SDK doesn't reutnr `nil` output on success, should we check `output` for `nil`  here just in case?

pkg/auth/awssts/exchange.go
jhrozek: Oh, I thought it was the other way around and that LoadDefaultConfig uses anonymous creds by default - because I tested the whole branch in k8s where .aws/credentials is not present at all and the home dir is IIRC not even writeable.

Anyway, suggestion taken, thanks!

pkg/auth/awssts/exchange.go
jhrozek: I did the latter. I don't think a SSRF would be a risk, but it's simpler logic, too

pkg/auth/awssts/exchange.go
jhrozek: I did that for consistency sake, but I don't think you'll be realistically matching the sentinel errors, they are not recoverable anyway

pkg/auth/awssts/exchange_test.go
jhrozek: exactly. I didn't generate mocks when I merged the interfaces and when it was time to use the mocks it was easier to roll a very simple one than to muddy the subsequent PRs with generated mocks

pkg/auth/awssts/exchange.go
jhrozek: good call done

pkg/auth/awssts/exchange.go
jhrozek: yep, it costs nothing, added

pkg/auth/awssts/exchange.go
jhrozek: sure done

pkg/auth/awssts/exchange.go
ChrisJBurns: Ahh I think it may have an order of precedence then. If it can't do the env vars, it may default to the next in order. That makes more sense.

PR #3665, opened 2026-02-06
Fix flakey unit tests

dmjb:
These would fail intermittently. Make them more robust.

### DIFF
--- a/docs/server/docs.go
+++ b/docs/server/docs.go
@@ -1440,6 +1440,14 @@ const docTemplate = `{
             "skills.InstalledSkill": {
                 "description": "InstalledSkill is set if the skill is installed.",
                 "properties": {
+                    "clients": {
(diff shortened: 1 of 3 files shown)


pkg/transport/proxy/transparent/transparent_test.go
```diff
@@ -1060,12 +1060,25 @@ func TestTransparentProxy_RemoteServerFailure_ConnectionRefused(t *testing.T) {
...
+			callbackInvoked = true
+		}
+		running, _ := proxy.IsRunning()
+		if !running {
+			proxyStopped = true
+		}
+		if callbackInvoked && proxyStopped {
+			break
+		}
+	}
```
JAORMX: while this probably works. Wouldn't we want to use channels to wait for the shutdown instead?

PR #3687, opened 2026-02-06
chore(vMCP): Add tool call labels to vMCP telemetry metrics

jerm-dro:
# Summary

- **Add `mcp.ParsingMiddleware` to the vMCP server middleware chain** so the telemetry middleware can label incoming request metrics with the actual `mcp_method` (e.g. `tools/call`, `initialize`) instead of `"unknown"`.
- **Add `mcp_resource_id` label to common request metrics** (`toolhive_mcp_requests`, `toolhive_mcp_request_duration`) in the telemetry HTTP middleware, surfacing the tool name for `tools/call`, resource URI for `resources/read`, or prompt name for `prompts/get`.
- **Add `tool_name` and `prompt_name` attributes to backend telemetry** spans and metrics (`toolhive_vmcp_backend_requests`, `toolhive_vmcp_backend_requests_duration`) for finer-grained observability on outgoing backend calls.
- **Add `TestIntegration_TelemetryMiddleware`** covering incoming request counters, backend call metrics, custom resource attributes, `mcp_method` label propagation, and `mcp_resource_id` label propagation.

# Changed files

| File | Change |
|------|--------|
| `pkg/vmcp/server/server.go` | Wire `mcp.ParsingMiddleware` into the middleware chain before telemetry |
| `pkg/vmcp/server/telemetry.go` | Add `tool_name`/`prompt_name` attrs to backend metric `record()` |
| `pkg/telemetry/middleware.go` | Add `mcp_resource_id` to common metric attributes in `recordMetrics` |
| `pkg/vmcp/server/integration_test.go` | New `TestIntegration_TelemetryMiddleware` + `mcp_resource_id` assertion |
| `pkg/telemetry/integration_test.go` | Assert `mcp_resource_id` on request counter; extract goconst constants |
| `pkg/telemetry/middleware_test.go` | Use `metricRequestCounter` constant to fix goconst lint |

# Test plan

- [x] `go test ./pkg/telemetry/...` — all telemetry unit tests pass
- [x] `go test ./pkg/vmcp/server/...` — all vMCP server tests pass (including new `TestIntegration_TelemetryMiddleware`)

### DIFF
--- a/pkg/telemetry/integration_test.go
+++ b/pkg/telemetry/integration_test.go
@@ -23,6 +23,11 @@ import (
 	"github.com/stacklok/toolhive/pkg/mcp"
 )
 
+const (
+	testToolName         = "github_search"
+	metricRequestCounter = "toolhive_mcp_requests"
+)
+
 func TestTelemetryIntegration_EndToEnd(t *testing.T) {
 	t.Parallel()
 
@@ -200,7 +205,7 @@ func TestTelemetryIntegration_WithRealProviders(t *testing.T) {
 	mcpRequest := &mcp.ParsedMCPRequest{
 		Method:     "tools/call",
 		ID:         "test-123",
-		ResourceID: "github_search",
+		ResourceID: testToolName,
 		Arguments: map[string]interface{}{
 			"query":   "test query",
 			"api_key": "secret123", // This should be redacted
@@ -237,7 +242,7 @@ func TestTelemetryIntegration_WithRealProviders(t *testing.T) {
 	}
 
 	assert.Equal(t, "tools/call", attrMap["mcp.method"])
-	assert.Equal(t, "github_search", attrMap["mcp.tool.name"])
+	assert.Equal(t, testToolName, attrMap["mcp.tool.name"])
 	assert.Equal(t, "test-123", attrMap["mcp.request.id"])
 	assert.Equal(t, "POST", attrMap["http.method"])
 	assert.Equal(t, int64(200), attrMap["http.status_code"])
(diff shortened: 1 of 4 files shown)


pkg/vmcp/server/server.go
```diff
@@ -429,15 +430,20 @@ func (s *Server) Start(ctx context.Context) error {
...
 
 	if s.config.TelemetryProvider != nil {
 		mcpHandler = s.config.TelemetryProvider.Middleware(s.config.Name, "streamable-http")(mcpHandler)
 		logger.Info("Telemetry middleware enabled for MCP endpoints")
 	}
 
+	// Apply MCP parsing middleware to extract JSON-RPC method from request body.
+	// This runs before telemetry so that recordMetrics can label metrics with the
+	// actual mcp_method (e.g. "tools/call", "initialize") instead of "unknown".
+	mcpHandler = mcpparser.ParsingMiddleware(mcpHandler)
```
JAORMX, line 445: It's odd we didn't have this as the audit middleware would have needed this. This should be ine of the first middleware to kick in. How is this placed in the pipeline compared to the proxyrunner?

jerm-dro: The audit middleware (`Auditor.Middleware()` actually doesn't depend on GetParsedMCPRequest — it uses a different `backendEnrichmentMiddleware` which populates other context variables which are finally plucked here: https://github.com/stacklok/toolhive/blob/main/pkg/audit/auditor.go#L156-L213.           


As for the proxyrunner, its middleware runs the parser before telemetry.

jerm-dro: Both of these middleware systems are quite confusing:
- proxy: relies on plumbing context variables. defines middleware in order, applies it in reverse, effectively executing it in the order it was originally defined.
- vmcp: relies on context variables. defines middleware more explicitly: wrap the handler meaning the first wrappers execute later in the middleware stack.

😵 

There's an opportunity to clean this up. I noted this as something in need of an RFC.

Issue #3647, opened 2026-02-06
Client metadata extension for skills


JAORMX:
## Description

Extend ToolHive's `pkg/client/` system with skill-specific metadata fields, adding skill directory paths for supported clients (Claude Code, Codex, Copilot, OpenCode) and skill path resolution logic.

## Context

ToolHive's client system already supports 26+ MCP clients. This task extends `mcpClientConfig` with skill-specific fields so the installer knows where to place skill files for each client.

**Dependencies**: stacklok/toolhive#3645
**Blocks**: Skill installer and extractor

## Acceptance Criteria

- [ ] `mcpClientConfig` extended with: `SupportsSkills bool`, `SkillsRelPath []string`, `SkillsPlatformPrefix map[string][]string`
- [ ] Skill paths configured for:
  - Claude Code: global `~/.claude/skills/`, local `.claude/skills/`
  - Codex: global `~/.codex/skills/`, local `.codex/skills/`
  - Copilot: global `~/.copilot/skills/`, local `.github/skills/`
  - OpenCode: local only `.opencode/skill/`
- [ ] `GetSkillPath(client MCPClient, skillName string, scope Scope) (string, error)` function
- [ ] `ClientSupportsSkills(client MCPClient) bool` function
- [ ] `ListSkillSupportingClients() []MCPClient` function
- [ ] Tests for path resolution, SPDX headers, no regressions

## Technical Approach

Extend existing `pkg/client/config.go`. Add Codex/Copilot entries if missing.

## References

- stacklok/toolhive#3645

JAORMX:
Heads up: PR https://github.com/stacklok/toolhive/pull/3700 (now merged) renamed `MCPClient` → `ClientApp` and `MCPClientStatus` → `ClientAppStatus` across the codebase. The motivation was exactly this issue — with skills support coming, the client config struct will hold both MCP server settings and skill paths, so `MCPClient` was misleading.

The acceptance criteria here should be updated to reflect the new naming:
- `mcpClientConfig` is now the config struct to extend (the internal config wasn't renamed, but the public type was)
- Function signatures should use `ClientApp` instead of `MCPClient`, e.g.:
  - `GetSkillPath(client ClientApp, skillName string, scope Scope) (string, error)`
  - `ClientSupportsSkills(client ClientApp) bool`
  - `ListSkillSupportingClients() []ClientApp`

JAORMX:
[...]

PR #3700, opened 2026-02-09
Rename MCPClient type to ClientApp

JAORMX:
## Summary

- Renames the `MCPClient` type to `ClientApp` across the codebase — these represent AI coding tools (Claude Code, Cursor, Codex, etc.), not MCP-specific clients
- `MCPClientStatus` becomes `ClientAppStatus` as a consequence
- MCP-specific types and fields (`MCPServer`, `MCPServersPathPrefix`, etc.) are intentionally unchanged since they genuinely describe MCP concepts
- Regenerated mocks and swagger docs

## Motivation

With skills support coming (#3647), the client config struct will hold both MCP server settings and skill paths. Having the type called `MCPClient` would be misleading since skills have nothing to do with MCP. This rename prepares clean naming for that work.

## Test plan

- [x] `go build ./...` passes
- [x] `task test` (pkg/client, pkg/api, pkg/skills) — all pass
- [x] `task docs` regenerated
- [x] `task gen` regenerated mocks
- [x] Verified vmcp test helper `MCPClient` (different type) is untouched

### DIFF
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -30,60 +30,62 @@ const lockTimeout = 1 * time.Second
 // defaultURLFieldName is the default URL field name used when no specific mapping exists
 const defaultURLFieldName = "url"
 
-// MCPClient is an enum of supported MCP clients.
-type MCPClient string
+// ClientApp is an enum of supported AI clients (IDEs, editors, and coding tools).
+//
+//nolint:revive // ClientApp is intentionally named for clarity across packages
+type ClientApp string
 
 const (
 	// RooCode represents the Roo Code extension for VS Code.
-	RooCode MCPClient = "roo-code"
+	RooCode ClientApp = "roo-code"
 	// Cline represents the Cline extension for VS Code.
-	Cline MCPClient = "cline"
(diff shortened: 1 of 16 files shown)


pkg/client/config.go
```diff
@@ -122,7 +124,7 @@ const (
 
 // mcpClientConfig represents a configuration path for a supported MCP client.
 type mcpClientConfig struct {
```
aponcedeleonch, line 126: nit: missed this `struct`

pkg/client/config.go
aponcedeleonch, line 174: nit: missed this `slice`, probably should be something like `clientsAppConfig`

aponcedeleonch (approved):
looks good just 2 minor nits that can be addressed in a future PR

pkg/client/config.go
JAORMX: I have a queued PR for this :smile: wanted to keep this small.

pkg/client/config.go
JAORMX: See the comment above.

Issue #3645, opened 2026-02-06
Skill types, interfaces, and command/API stubs


JAORMX:
## Description

Define the foundational types, interfaces, and stub implementations for the skill lifecycle management feature in ToolHive. This task establishes the `SkillService` interface, domain types, CLI command stubs for `thv skill {list,install,uninstall,info,validate,build,push}`, REST API route stubs at `/api/v1beta/skills/*` returning 501 Not Implemented, and wires everything into the root command and `ServerBuilder`. This is the root task that unblocks all parallel work on subsequent skill lifecycle tasks.

## Context

This is the first task in the [Skills Lifecycle Management epic](https://github.com/stacklok/stacklok-epics/issues/239), which ports skill management from the standalone Skillet CLI into the ToolHive ecosystem. The stubs-first approach enables parallel development across Wave 2 tasks (parser/validator, client metadata, state management) and the toolhive-core OCI library.

A key architectural decision for skills is that CLI commands call the REST API via an HTTP client (unlike existing commands that call `pkg/` directly). This requires the API server (`thv serve`) to be running. The stubs created in this task establish that pattern.

**Dependencies**: None (root task)
**Blocks**: TASK-005, TASK-006, TASK-007, TASK-010

## Acceptance Criteria

- [ ] `pkg/skills/service.go` defines the `SkillService` interface with methods: `List`, `Install`, `Uninstall`, `Info`, `Validate`, `Build`, `Push`
- [ ] `pkg/skills/types.go` defines domain types: `InstalledSkill`, `SkillMetadata`, `SkillIndex`, `SkillIndexEntry`, option/result structs, and enums (`Scope`, `InstallStatus`)
- [ ] `pkg/skills/client/client.go` defines the HTTP client struct with `NewClient(baseURL string) *Client` constructor (stub)
- [ ] `pkg/api/v1/skills.go` defines `SkillsRouter` with chi subrouter and route stubs for all 7 endpoints returning `501 Not Implemented`
- [ ] `pkg/api/v1/skills_types.go` defines request/response types for the skills REST API
- [ ] `cmd/thv/app/skill.go` defines `skillCmd` parent command and 7 subcommands as stubs printing "not yet implemented"
- [ ] `skillCmd` is registered in `NewRootCmd()` in `cmd/thv/app/commands.go`
- [ ] `pkg/api/server.go` `ServerBuilder` has a `skillManager` field and `WithSkillManager()` builder method
[...]

PR #3702, opened 2026-02-09
Add SKILL.md parser and skill directory validator

### LINKED ISSUE #3646: SKILL.md parser and validator
## Description

Implement SKILL.md parsing (YAML frontmatter extraction) and skill validation in ToolHive's `pkg/skills/` package. Provides the ability to parse skill metadata from SKILL.md files and validate skills against naming conventions and structural requirements.

## Context

Part of the Skills Lifecycle Management epic. Used by the skill service for install, build, and validate operations.

**Dependencies**: stacklok/toolhive#3645
**Blocks**: Skill service implementation

## Acceptance Criteria

- [ ] `pkg/skills/parser.go` parses SKILL.md YAML frontmatter (name, description, version, allowed-tools, license, requires)
- [ ] Supports `allowed-tools` as space/comma-delimited string or YAML array
- [ ] Parses dependency references from metadata
- [ ] Max 64KB frontmatter, max 100 dependencies
- [ ] `pkg/skills/validator.go` validates skill directories:
  - SKILL.md exists and has valid frontmatter
  - Name matches `^[a-z0-9][a-z0-9-]{0,62}[a-z0-9]$`
  - Required fields present (name, description)
  - No path traversal patterns
  - No symlinks in skill directory
- [ ] Returns structured validation results with errors and warnings
- [ ] All tests pass, SPDX headers present

## References

- stacklok/toolhive#3645
- [Agent Skills Specification](https://agentskills.io/specification)

JAORMX:
## Summary

- Implement SKILL.md frontmatter parsing with YAML extraction, `StringOrSlice` unmarshaler for `allowed-tools`, and dependency parsing from `metadata.skillet.requires`
- Implement skill directory validation: name regex, required fields, field length constraints, name-directory match, symlink detection, path traversal checks
- Add comprehensive table-driven tests (40+ test cases) covering parser and validator edge cases, boundary conditions, and security limits

## Details

Part of the Skills Lifecycle Management epic. Depends on #3645 (stubs/types/interfaces, already merged).

### Parser (`pkg/skills/parser.go` + types in `types.go`)
- `ParseSkillMD()` extracts YAML frontmatter and body from SKILL.md content
- `StringOrSlice` custom YAML unmarshaler supports space-delimited (per Agent Skills spec), comma-delimited, and YAML array formats
- 64KB frontmatter size limit (prevents YAML billion laughs attacks)
- 100 dependency limit for `skillet.requires` metadata

### Validator (`pkg/skills/validator.go`)
- `ValidateSkillDir()` validates a skill directory against the [Agent Skills specification](https://agentskills.io/specification)
- Name validation: `^[a-z0-9][a-z0-9-]{0,62}[a-z0-9]$`, no consecutive hyphens
- Name must match directory name
- Required fields: name, description
- Field constraints: description <= 1024 chars, compatibility <= 500 chars
- Security: symlink detection, path traversal detection
- Warning: SKILL.md > 500 lines

### Types added to `types.go`
- `SkillFrontmatter` - raw YAML frontmatter representation
- `ParseResult` - parsed SKILL.md contents
- `Dependency` - OCI reference dependency
- `StringOrSlice` - flexible YAML unmarshaler

### `ValidationResult` updated in `options.go`
- Added `Warnings []string` field for non-blocking validation warnings

Closes #3646

## Large PR Justification

- Tests take a lot of space here

## Test plan

- [x] `task test` — all unit tests pass (40+ test cases)
- [x] `task lint` — 0 lint issues
- [x] `task license-check` — all SPDX headers present
- [x] Verified against Skillet reference implementation
- [x] Verified against Agent Skills specification

### DIFF
--- a/pkg/skills/parser.go
+++ b/pkg/skills/parser.go
@@ -7,16 +7,12 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"strings"
 
 	"gopkg.in/yaml.v3"
 )
 
 // frontmatterDelimiter is the YAML frontmatter delimiter.
 var frontmatterDelimiter = []byte("---")
-
-// skilletRequiresKey is the metadata key for external OCI dependencies.
-const skilletRequiresKey = "skillet.requires"
 
 // MaxFrontmatterSize is the maximum size of frontmatter content in bytes.
 // This prevents YAML parsing attacks (e.g., billion laughs).
@@ -36,9 +32,9 @@ func ParseSkillMD(content []byte) (*ParseResult, error) {
 		return nil, err
 	}
 
-	requires, err := parseRequires(fm.Metadata)
+	requires, err := toDependencies(fm.Requires)
 	if err != nil {
-		return nil, fmt.Errorf("parsing skillet.requires: %w", err)
+		return nil, fmt.Errorf("parsing requires: %w", err)
 	}
 
 	return &ParseResult{
@@ -97,27 +93,20 @@ func extractFrontmatter(content []byte) (*SkillFrontmatter, []byte, error) {
 	return &fm, body, nil
 }
 
-// parseRequires extracts the skillet.requires field from metadata and converts
-// the newline-separated OCI references to Dependencies.
-func parseRequires(metadata map[string]string) ([]Dependency, error) {
-	requiresStr, ok := metadata[skilletRequiresKey]
-	if !ok || requiresStr == "" {
+// toDependencies converts a list of OCI reference strings to Dependencies.
+func toDependencies(refs []string) ([]Dependency, error) {
+	if len(refs) == 0 {
 		return nil, nil
 	}
 
-	lines := strings.Split(requiresStr, "\n")
+	if len(refs) > MaxDependencies {
+		return nil, fmt.Errorf("too many dependencies: %d (max %d)", len(refs), MaxDependencies)
+	}
 
-	deps := make([]Dependency, 0, len(lines))
-	for _, line := range lines {
-		line = strings.TrimSpace(line)
-		if line == "" {
-			continue
-		}
-
-		deps = append(deps, Dependency{Reference: line})
-
-		if len(deps) > MaxDependencies {
-			return nil, fmt.Errorf("too many dependencies: more than %d", MaxDependencies)
+	deps := make([]Dependency, 0, len(refs))
+	for _, ref := range refs {
+		if ref != "" {
+			deps = append(deps, Dependency{Reference: ref})
 		}
 	}
 
--- a/pkg/skills/types.go
+++ b/pkg/skills/types.go
@@ -86,18 +86,14 @@ func (s *StringOrSlice) UnmarshalYAML(value *yaml.Node) error {
 
 // SkillFrontmatter represents the raw YAML frontmatter from a SKILL.md file.
 type SkillFrontmatter struct {
-	Name          string        `yaml:"name"`
-	Description   string        `yaml:"description"`
-	Version       string        `yaml:"version,omitempty"`
-	AllowedTools  StringOrSlice `yaml:"allowed-tools,omitempty"`
-	License       string        `yaml:"license,omitempty"`
-	Compatibility string        `yaml:"compatibility,omitempty"`
-	// Metadata is an arbitrary key-value map per the Agent Skills spec.
-	// Note: the spec defines metadata as map[string]string, so extension
-	// fields like "skillet.requires" must be encoded as scalar strings
-	// (e.g. newline-delimited). YAML array values will silently fail
-	// to unmarshal into this map.
-	Metadata map[string]string `yaml:"metadata,omitempty"`
+	Name          string            `yaml:"name"`
+	Description   string            `yaml:"description"`
+	Version       string            `yaml:"version,omitempty"`
+	AllowedTools  StringOrSlice     `yaml:"allowed-tools,omitempty"`
+	Requires      StringOrSlice     `yaml:"toolhive.requires,omitempty"`
+	License       string            `yaml:"license,omitempty"`
+	Compatibility string            `yaml:"compatibility,omitempty"`
+	Metadata      map[string]string `yaml:"metadata,omitempty"`
 }
 
 // Dependency represents an external skill dependency (OCI reference).
--- a/pkg/skills/validator.go
+++ b/pkg/skills/validator.go
@@ -61,27 +61,36 @@ func ValidateSkillDir(path string) (*ValidationResult, error) {
 		}, nil
 	}
 
-	// Validate name
+	// Validate parsed fields
+	errs = append(errs, validateFields(result, filepath.Base(filepath.Clean(path)))...)
+
+	// Collect warnings
+	warnings = append(warnings, collectWarnings(result, content)...)
+
+	return &ValidationResult{
+		Valid:    len(errs) == 0,
+		Errors:   errs,
+		Warnings: warnings,
+	}, nil
+}
+
+// validateFields checks parsed frontmatter fields against spec constraints.
+func validateFields(result *ParseResult, dirName string) []string {
+	var errs []string
+
 	if err := validateName(result.Name); err != nil {
 		errs = append(errs, err.Error())
 	}
-
-	// Validate name matches directory
-	dirName := filepath.Base(filepath.Clean(path))
 	if result.Name != "" && result.Name != dirName {
 		errs = append(errs,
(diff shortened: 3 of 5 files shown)


pkg/skills/types.go
```diff
@@ -28,6 +34,81 @@ const (
...
+
+// SkillFrontmatter represents the raw YAML frontmatter from a SKILL.md file.
+type SkillFrontmatter struct {
+	Name          string            `yaml:"name"`
+	Description   string            `yaml:"description"`
+	Version       string            `yaml:"version,omitempty"`
+	AllowedTools  StringOrSlice     `yaml:"allowed-tools,omitempty"`
+	License       string            `yaml:"license,omitempty"`
+	Compatibility string            `yaml:"compatibility,omitempty"`
+	Metadata      map[string]string `yaml:"metadata,omitempty"`
```
aponcedeleonch: ## Inconsistency: `skillet.requires` vs `allowed-tools` list handling

Both `allowed-tools` and `skillet.requires` are conceptually lists of strings, but they're handled very differently:

- `allowed-tools` gets the flexible `StringOrSlice` custom unmarshaler (string or array, multiple delimiters) — great!
- `skillet.requires` is trapped in `map[string]string`, forcing it to be a newline-delimited scalar only. Then `parseRequires()` manually splits on `\n`.

This means if someone writes the natural YAML array form:
```yaml
metadata:
  skillet.requires:
    - ghcr.io/org/skill-a:v1.0.0
    - ghcr.io/org/skill-b:latest
```
it will **silently fail** because `map[string]string` can't hold a `[]string` value.

**Suggestion**: Promote `requires` to a top-level `SkillFrontmatter` field with its own `StringOrSlice` (or a `DependencyList` variant using newline as delimiter). This would:
1. Make it consistent with `allowed-tools`
2. Eliminate the `parseRequires` function and `skilletRequiresKey` constant
3. Support both string and array YAML formats
4. Keep `Metadata` for truly unstructured key-value pairs

Something like:
```go
type SkillFrontmatter struct {
    Name          string            `yaml:"name"`
    Description   string            `yaml:"description"`
    Version       string            `yaml:"version,omitempty"`
    AllowedTools  StringOrSlice     `yaml:"allowed-tools,omitempty"`
    Requires      StringOrSlice     `yaml:"requires,omitempty"`       // promoted
    License       string            `yaml:"license,omitempty"`
    Compatibility string            `yaml:"compatibility,omitempty"`
    Metadata      map[string]string `yaml:"metadata,omitempty"`
}
```

pkg/skills/types.go
```diff
@@ -28,6 +34,81 @@ const (
...
+	case yaml.ScalarNode:
+		str := value.Value
+		if str == "" {
+			*s = nil
+			return nil
+		}
+
+		// Determine delimiter: use comma if present, otherwise space (per spec)
+		var parts []string
+		if strings.Contains(str, ",") {
```
aponcedeleonch, line 59: Nit: clean implementation overall. One edge case worth noting: a mixed-delimiter string like `"Read,Glob Grep"` would split on comma (since comma is present), yielding `["Read", "Glob Grep"]` — i.e. `"Glob Grep"` stays as one token. This seems like an acceptable tradeoff given the spec says space-delimited is the canonical format, but might be worth a brief comment documenting the precedence.

pkg/skills/parser.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if !bytes.HasPrefix(content, frontmatterDelimiter) {
+		return nil, nil, ErrInvalidFrontmatter
+	}
+
+	// Skip opening delimiter and optional newline
+	rest := content[len(frontmatterDelimiter):]
+	rest = bytes.TrimPrefix(rest, []byte("\n"))
+
+	endIdx := bytes.Index(rest, frontmatterDelimiter)
```
aponcedeleonch: Minor: the size check on line 79 happens *after* `bytes.Index` has already scanned the full `rest` content. This means a 100MB file with no closing `---` would be fully scanned before returning `ErrInvalidFrontmatter` (and never even reach the size check). Consider adding a fast-fail length check before the `bytes.Index` call:

```suggestion
	if len(rest) > MaxFrontmatterSize {
		return nil, nil, fmt.Errorf("%w: content too large for frontmatter parsing",
			ErrInvalidFrontmatter)
	}

	endIdx := bytes.Index(rest, frontmatterDelimiter)
```

pkg/skills/validator.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if strings.Contains(name, "--") {
+		return fmt.Errorf("invalid skill name %q: must not contain consecutive hyphens", name)
+	}
+	return nil
+}
+
+// checkSymlinks walks the directory and checks for symbolic links.
+func checkSymlinks(path string) error {
+	return filepath.Walk(path, func(p string, _ os.FileInfo, err error) error {
```
aponcedeleonch: Minor: consider using `filepath.WalkDir` instead of `filepath.Walk` (same for `checkPathTraversal` below). `WalkDir` (Go 1.16+) is more efficient — it doesn't call `Stat` on every entry. Since you're calling `os.Lstat` manually anyway, `WalkDir` would avoid the redundant `Stat` that `Walk` does internally.

Also, both `checkSymlinks` and `checkPathTraversal` walk the same directory tree. They could be combined into a single walk to avoid traversing the tree twice.

pkg/skills/types.go
JAORMX: Great catch, and yeah, the silent failure is a real footgun. The reason `skillet.requires` lives in `metadata` is that it's not part of the [Agent Skills spec](https://agentskills.io/specification) — it's an extension. The spec defines `metadata` as "a map from string keys to string values", so `map[string]string` is by design. Skillet's POC implemented it this way too, but the spec is the source of truth here.

That said, you're right that someone writing the natural YAML array form would get bitten with no warning. I'll add a comment documenting this limitation. For a follow-up, we could add validation that warns if a metadata value looks like it was intended to be an array. WDYT?

pkg/skills/types.go
JAORMX: Good call. Will add a comment documenting the precedence: comma wins over space when both are present. The spec says space-delimited is the canonical format; comma and array support are our compatibility extensions.

pkg/skills/parser.go
JAORMX: Great catch on the ordering. Will add the early size check before `bytes.Index`. Note that we're also somewhat protected since the caller reads via `os.ReadFile` which is bounded by available memory, but the early check is good defense-in-depth. No reason to scan 100MB looking for `---` when we know we'll reject it anyway.

pkg/skills/validator.go
JAORMX: Good suggestions on both counts. `WalkDir` avoids the redundant `Stat` that `Walk` does internally, and since we're calling `os.Lstat` manually anyway it's a no-brainer. Will also combine both checks into a single walk function. Cleaner and more efficient.

pkg/skills/types.go
aponcedeleonch: if space-delimited is the canonical format, wouldn't we want that to have precedence?

JAORMX: Good question! So, basically, comma-first is correct *because* the canonical space-delimited format never contains commas. It naturally falls through to the space path:

- `Read Glob Grep` (canonical) -> no commas found -> splits on space -> `["Read", "Glob", "Grep"]` (correct)
- `Read, Glob, Grep` (compat) -> commas found -> splits on comma -> `["Read", "Glob", "Grep"]` (correct)

If we flipped it and gave space precedence:

- `Read, Glob, Grep` -> splits on space -> `["Read,", "Glob,", "Grep"]` (broken)

The commas would stay stuck to the tool names. Note that tool names themselves never contain commas, so comma presence is a reliable signal that the author intended comma-delimited format.

pkg/skills/types.go
aponcedeleonch: I would add support for writing natural YAML TBH since we're augmenting what the spec supports already. I would also add the warning you're suggesting. That way users know that their skill might not work outside of ToolHive

pkg/skills/types.go
aponcedeleonch: ok, sounds good. Maybe we should also at least log a warning that we're doing delimited not supported originally by the spec? Will leave it up to you :)

pkg/skills/types.go
JAORMX: Quick update: we're promoting requires to a top-level frontmatter field (with StringOrSlice, so both YAML arrays and newline-delimited strings work). Also renaming from skillet.requires to toolhive.requires since skillet was the PoC name.

PR #3712, opened 2026-02-09
Add client skill metadata for skill path resolution

### LINKED ISSUE #3647: Client metadata extension for skills

JAORMX:
## Summary

- Extend `clientAppConfig` with skill-specific fields (`SupportsSkills`, `SkillsGlobalPath`, `SkillsProjectPath`, `SkillsPlatformPrefix`) so the skill installer knows where to place skill files for each AI client
- Populate skill paths for ClaudeCode, Codex, and OpenCode matching the canonical paths from each client's official documentation
- Replace unused `ScopeSystem` with `ScopeProject` for project-local skill installations
- Add `ClientManager` methods: `SupportsSkills()`, `ListSkillSupportingClients()`, `GetSkillPath()`, and standalone `DetectProjectRoot()` utility
- Export `ValidateSkillName()` from `pkg/skills/validator.go` (reuses the existing regex from #3702 rather than duplicating it)

### Skill paths (verified against each client's official docs)

| Client | Global (user) | Project | Source |
|--------|--------------|---------|--------|
| ClaudeCode | `~/.claude/skills/` | `.claude/skills/` | [Claude Code docs](https://code.claude.com/docs/en/skills) |
| Codex | `~/.agents/skills/` | `.agents/skills/` | [Codex docs](https://developers.openai.com/codex/skills) |
| OpenCode | `~/.config/opencode/skills/` | `.opencode/skills/` | [OpenCode docs](https://opencode.ai/docs/skills/) |

Copilot is deferred — no `ClientApp` entry exists in ToolHive yet (see [comment on issue](https://github.com/stacklok/toolhive/issues/3647#issuecomment-3872015267)).

### Design decisions

- All skill functions are **methods on `ClientManager`** using dependency injection (`homeDir`, `clientIntegrations`), not global state or direct OS calls
- `DetectProjectRoot` is a **standalone utility**, not auto-called by `GetSkillPath` — callers compose as needed
- Skill name validation **delegates to `skills.ValidateSkillName()`** (the canonical allowlist regex from #3702) instead of reimplementing

Closes #3647

## Test plan

- [x] `task test` — all pkg/client/ and pkg/skills/ tests pass (30 test cases)
- [x] `task lint` — zero issues
- [x] `task docs` — swagger regenerated for `ScopeSystem` → `ScopeProject`
- [x] Tests use `NewTestClientManager` with injected deps (no filesystem/CWD dependencies)
- [x] All sentinel errors verified with `errors.Is` in tests
- [x] Path traversal, null bytes, uppercase, consecutive hyphens all rejected by allowlist regex

### DIFF
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -122,13 +122,25 @@ const (
 	TOMLStorageTypeArray TOMLStorageType = "array"
 )
 
+// Platform represents a runtime.GOOS value used as a key in platform-specific path maps.
+type Platform string
+
+const (
+	// PlatformLinux is the Linux platform.
+	PlatformLinux Platform = "linux"
+	// PlatformDarwin is the macOS platform.
+	PlatformDarwin Platform = "darwin"
+	// PlatformWindows is the Windows platform.
+	PlatformWindows Platform = "windows"
+)
+
 // clientAppConfig represents a configuration path for a supported MCP client.
 type clientAppConfig struct {
 	ClientType                    ClientApp
 	Description                   string
 	RelPath                       []string
 	SettingsFile                  string
-	PlatformPrefix                map[string][]string
+	PlatformPrefix                map[Platform][]string
 	MCPServersPathPrefix          string
 	Extension                     Extension
 	SupportedTransportTypesMap    map[types.TransportType]string // stdio mapped to streamable-http (SSE deprecated)
@@ -145,11 +157,11 @@ type clientAppConfig struct {
 	SupportsSkills    bool     // Whether this client supports skills
 	SkillsGlobalPath  []string // Path segments for global skills dir (from home dir)
 	SkillsProjectPath []string // Path segments for project-local skills dir (from project root)
-	// SkillsPlatformPrefix maps runtime.GOOS values (e.g., "linux", "darwin") to path
-	// segments inserted between home dir and SkillsGlobalPath. Needed for clients
-	// following platform conventions (e.g., XDG ~/.config/ on Linux/macOS).
+	// SkillsPlatformPrefix maps Platform values to path segments inserted between
+	// home dir and SkillsGlobalPath. Needed for clients following platform conventions
+	// (e.g., XDG ~/.config/ on Linux/macOS).
 	// If nil or missing an entry for the current OS, no prefix is added.
-	SkillsPlatformPrefix map[string][]string
+	SkillsPlatformPrefix map[Platform][]string
 }
 
 // extractServersKeyFromConfig extracts the servers key from MCPServersPathPrefix
@@ -188,10 +200,10 @@ var supportedClientIntegrations = []clientAppConfig{
 		RelPath: []string{
 			"Code", "User", "globalStorage", "rooveterinaryinc.roo-cline", "settings",
 		},
-		PlatformPrefix: map[string][]string{
-			"linux":   {".config"},
-			"darwin":  {"Library", "Application Support"},
-			"windows": {"AppData", "Roaming"},
+		PlatformPrefix: map[Platform][]string{
+			PlatformLinux:   {".config"},
+			PlatformDarwin:  {"Library", "Application Support"},
+			PlatformWindows: {"AppData", "Roaming"},
 		},
 		MCPServersPathPrefix: "/mcpServers",
 		Extension:            JSON,
@@ -214,10 +226,10 @@ var supportedClientIntegrations = []clientAppConfig{
 		RelPath: []string{
 			"Code", "User", "globalStorage", "saoudrizwan.claude-dev", "settings",
 		},
-		PlatformPrefix: map[string][]string{
-			"linux":   {".config"},
-			"darwin":  {"Library", "Application Support"},
-			"windows": {"AppData", "Roaming"},
+		PlatformPrefix: map[Platform][]string{
+			PlatformLinux:   {".config"},
+			PlatformDarwin:  {"Library", "Application Support"},
+			PlatformWindows: {"AppData", "Roaming"},
 		},
 		MCPServersPathPrefix: "/mcpServers",
 		Extension:            JSON,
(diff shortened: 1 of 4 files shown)


pkg/skills/validator.go
```diff
@@ -137,6 +140,15 @@ func validateName(name string) error {
...
 }
 
+// validateName checks that a skill name matches the required pattern.
+// Empty names are allowed here since they are caught by the required fields check in ValidateSkillDir.
+func validateName(name string) error {
+	if name == "" {
+		return nil // Caught by required fields check
+	}
+	return ValidateSkillName(name)
+}
```
aponcedeleonch: **Concern: `ValidateSkillName` and `validateName` have duplicated logic with a subtle semantic difference.**

The new `validateName` wrapper (lines 143-150) exists only to handle the empty-string case differently — it returns `nil` for empty names because the caller (`validateFields` at line 81) has its own "name is required" check downstream.

However, this creates two functions with near-identical logic and a subtle difference that's easy to miss. Consider eliminating `validateName` entirely and having `validateFields` handle the empty-name guard itself, then call `ValidateSkillName` directly. Update `validateFields` (lines 78-103) to:
```go
if result.Name == "" {
    errs = append(errs, "name is required")
} else {
    if err := ValidateSkillName(result.Name); err != nil {
        errs = append(errs, err.Error())
    }
    if result.Name != dirName {
        errs = append(errs, fmt.Sprintf("skill name %q must match directory name %q", result.Name, dirName))
    }
}
```

This makes the empty-string handling explicit at the call site and avoids potential confusion where `ValidateSkillName("")` returns an error but `validateName("")` returns nil.

pkg/client/skills.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func (cm *ClientManager) lookupClientAppConfig(clientType ClientApp) *clientAppConfig {
+	for i := range cm.clientIntegrations {
+		if cm.clientIntegrations[i].ClientType == clientType {
+			return &cm.clientIntegrations[i]
+		}
+	}
+	return nil
+}
```
aponcedeleonch, line 121: **Concern: Duplicated patterns with existing code in `config.go` and `discovery.go`.**

**a) Client lookup loop:** `lookupClientAppConfig` (lines 118-125) iterates `clientIntegrations` to find a config by `ClientType`. The exact same loop exists in `retrieveConfigFileMetadata` (config.go:984-991) and `CreateClientConfig` (config.go:871-876). A shared `lookupClientAppConfig` on `ClientManager` could be reused by all three.

**b) Path building:** `buildSkillsGlobalPath` (lines 127-138) assembles `homeDir + platformPrefix[runtime.GOOS] + pathSegments + suffix` via `filepath.Join`. This is structurally identical to:
- `buildConfigFilePath` (config.go:1052-1058)
- `buildConfigDirectoryPath` (discovery.go:140-146)

All three do the same thing with different final segments. A shared path-builder helper could reduce this.

Not a blocker for this PR, but worth addressing to avoid the pattern of each new feature adding its own copy of these utility loops.

pkg/client/config.go
```diff
@@ -141,6 +141,11 @@ type clientAppConfig struct {
 	YAMLDefaults        map[string]any  // Default values to add to entries
 	// TOML-specific configuration (only used when Extension == TOML)
 	TOMLStorageType TOMLStorageType // How servers are stored in TOML (map or array)
+	// Skill-specific configuration
+	SupportsSkills       bool                // Whether this client supports skills
+	SkillsGlobalPath     []string            // Path segments for global skills dir (from home dir)
+	SkillsProjectPath    []string            // Path segments for project-local skills dir (from project root)
+	SkillsPlatformPrefix map[string][]string // Platform-specific prefixes for global skills path
```
aponcedeleonch: **Concern: `SkillsPlatformPrefix` is stringly-typed and lacks documentation on why it's needed.**

**1. Stringly-typed map keys:** The keys are bare strings (`"linux"`, `"darwin"`) matched against `runtime.GOOS` at runtime. A typo like `"Darwin"` would silently fail. The existing `PlatformPrefix` field (line 131) has the same pre-existing issue, but this is a good opportunity to at least add a comment documenting that keys must be valid `runtime.GOOS` values.

**2. Missing documentation on purpose:** The field exists because some clients store global skills in platform-specific base directories (e.g., OpenCode follows XDG conventions, so skills live under `~/.config/opencode/skills/` on Linux/macOS). This isn't explained anywhere. Suggested doc comment:

```go
// SkillsPlatformPrefix maps runtime.GOOS values (e.g., "linux", "darwin") to path
// segments inserted between home dir and SkillsGlobalPath. Needed for clients
// following platform conventions (e.g., XDG ~/.config/ on Linux/macOS).
// If nil or missing an entry for the current OS, no prefix is added.
SkillsPlatformPrefix map[string][]string
```

PR #3713, opened 2026-02-09
Add rfc8707 resource parameter refresh token

rodrigo1208:
Add RFC 8707 resource parameter support for OAuth token refresh                                                                       
                                                                                                                                        
  Summary                                                                                                                               
                                                                                                                                        
  Fixes token refresh to include the RFC 8707 resource parameter when connecting to remote OAuth-protected MCP servers. Initial         
  authentication (authorization + token exchange) already includes the resource parameter, but token refresh was missing it because the 
  standard golang.org/x/oauth2 library doesn't support custom parameters during refresh.                                                
                                                                                                                                        
  Problem                                                                                                                               
                                                                                                                                        
  Before this PR:                                                                                                                       
  ✅ Authorization request: includes resource parameter                                                                                 
  ✅ Token exchange:        includes resource parameter                                                                                 
  ❌ Token refresh:         missing resource parameter                                                                                  
                                                                                                                                        
  When access tokens expire, the refresh request didn't include the resource parameter, breaking RFC 8707 compliance and causing issues 
  with authorization servers that require it.                                                                                           
                                                                                                                                        
  Solution                                                                                                                              
                                                                                                                                        
  Implemented resourceTokenSource - a custom oauth2.TokenSource that wraps the standard token source and adds the resource parameter to 
  refresh requests.                                                                                                                     
                                                                                                                                        
  Changes                                                                                                                               
                                                                                                                                        
  New: pkg/auth/oauth/resource_token_source.go (118 lines)                                                                              
  - Custom token source that includes resource parameter in refresh requests                                                            
  - Maintains standard behavior (caching, automatic refresh)                                                                            
                                                                                                                                        
  Modified: pkg/auth/oauth/flow.go (+7 lines)                                                                                           
  - Uses resourceTokenSource when resource is configured                                                                                
  - Falls back to standard token source when no resource specified                                                                      
                                                                                                                                        
  Tests: 23 new tests (600 lines)                                                                                                       
  - Token refresh includes resource parameter (core functionality)                                                                      
  - RFC 8707 compliance verification                                                                                                    
  - Error handling and backward compatibility                                                                                           
                                                                                                                                        
  Example                                                                                                                               
                                                                                                                                        
  Before:                                                                                                                               
  POST /oauth/token                                                                                                                     
  grant_type=refresh_token&refresh_token=xyz                                                                                            
                                                                                                                                        
  After:                                                                                                                                
  POST /oauth/token                                                                                                                     
  grant_type=refresh_token&refresh_token=xyz&resource=https://api.example.com                                                           
                                                                                                                                        
  Testing                                                                                                                               
                                                                                                                                        
  ✅ All tests pass: task test-all                                                                                                      
  ✅ Linting clean: task lint-fix                                                                                                       
  ✅ Backward compatible: works with or without resource parameter                                                                      
                                                                                                                                        
  Files Changed                                                                                                                         
                                                                                                                                        
  pkg/auth/oauth/resource_token_source.go      | 118 lines (new)                                                                        
  pkg/auth/oauth/resource_token_source_test.go | 600 lines (new)                                                                        
  pkg/auth/oauth/flow.go                       |   7 lines (modified)                                                                   
  pkg/auth/oauth/flow_test.go                  | 218 lines (modified)

### DIFF
--- a/pkg/auth/oauth/resource_token_source.go
+++ b/pkg/auth/oauth/resource_token_source.go
@@ -6,7 +6,6 @@ package oauth
 import (
 	"context"
 	"fmt"
-	"io"
 	"net/http"
 	"time"
 
@@ -25,52 +24,16 @@ type resourceTokenSource struct {
 	httpClient *http.Client
 }
 
-type limitedReadCloser struct {
-	io.Reader
-	c io.Closer
-}
-
-func (l *limitedReadCloser) Close() error {
-	return l.c.Close()
-}
-
-type limitResponseBodyRoundTripper struct {
-	next  http.RoundTripper
-	limit int64
-}
-
-func (l *limitResponseBodyRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
-	resp, err := l.next.RoundTrip(req)
-	if err != nil {
-		return nil, err
-	}
-
-	if resp != nil && resp.Body != nil && l.limit > 0 {
-		resp.Body = &limitedReadCloser{
-			Reader: io.LimitReader(resp.Body, l.limit),
-			c:      resp.Body,
-		}
-	}
-
-	return resp, nil
-}
-
 // newResourceTokenSource creates a token source that includes the resource parameter
 // in all token requests, including refresh requests.
 // The resource parameter must be non-empty (caller should check before calling).
 func newResourceTokenSource(config *oauth2.Config, token *oauth2.Token, resource string) oauth2.TokenSource {
-	baseTransport := http.DefaultTransport
-	if baseTransport == nil {
-		baseTransport = &http.Transport{}
-	}
-
 	return &resourceTokenSource{
 		config:   config,
 		resource: resource,
 		token:    token,
 		httpClient: &http.Client{
-			Transport: &limitResponseBodyRoundTripper{next: baseTransport, limit: 4096},
-			Timeout:   30 * time.Second,
+			Timeout: 30 * time.Second,
 		},
 	}
 }


pkg/auth/oauth/resource_token_source.go
```diff
@@ -0,0 +1,116 @@
...
+		_ = resp.Body.Close() // Ignore error on close during cleanup
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("token refresh failed with status %d", resp.StatusCode)
+	}
+
+	// Parse the token response
+	var token oauth2.Token
+	if err := json.NewDecoder(resp.Body).Decode(&token); err != nil {
```
jhrozek: It looks like the oauth2.Token structure contains:
```
	// Expiry is the optional expiration time of the access token.
	//
	// If zero, [TokenSource] implementations will reuse the same
	// token forever and RefreshToken or equivalent
	// mechanisms for that TokenSource will not be used.
	Expiry time.Time `json:"expiry,omitempty"`

	// ExpiresIn is the OAuth2 wire format "expires_in" field,
	// which specifies how many seconds later the token expires,
	// relative to an unknown time base approximately around "now".
	// It is the application's responsibility to populate
	// `Expiry` from `ExpiresIn` when required.
	ExpiresIn int64 `json:"expires_in,omitempty"
````

but the response from the refresh only contains per https://datatracker.ietf.org/doc/html/rfc6749#section-5.1 
```
   expires_in
         RECOMMENDED.  The lifetime in seconds of the access token.  For
         example, the value "3600" denotes that the access token will
         expire in one hour from the time the response was generated.
         If omitted, the authorization server SHOULD provide the
         expiration time via other means or document the default value.
```
this is an irritating limitation of x/oauth2  - ExpiresIn is unmarshalled by the Decode call, but not Expiry. But calls like Token.Valid() would only check the Expiry value. This is not visible if you just use the public API of the library because they do the following internally:
```
  func (e *tokenJSON) expiry() time.Time {
      if v := e.ExpiresIn; v != 0 {
          return time.Now().Add(time.Duration(v) * time.Second)
      }
      return time.Time{}
  }
```
I think we need to copy a similar block to this code, what do you think?

Comment (jhrozek) [inline pkg/auth/oauth/resource_token_source.go]: additionally, could we use an `io.LimitReader` call here as well to protect against malicious token endpoints?

File: pkg/auth/oauth/resource_token_source.go [hunk 2]
```diff
@@ -0,0 +1,116 @@
...
+	// Add resource parameter per RFC 8707
+	v.Set("resource", r.resource)
+
+	// Add client credentials
+	if r.config.ClientID != "" {
+		v.Set("client_id", r.config.ClientID)
+	}
+	if r.config.ClientSecret != "" {
+		v.Set("client_secret", r.config.ClientSecret)
+	}
```
jhrozek, line 78: I wonder if it would be good to default to, per https://www.rfc-editor.org/rfc/rfc6749#section-2.3.1 HTTP Basic Auth and only fall back to POST body?

Although I think most IDPs support both, I wonder if basic auth would be a more reasonable default?

pkg/auth/oauth/resource_token_source.go
```diff
@@ -0,0 +1,116 @@
...
+	resp, err := r.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("token refresh request failed: %w", err)
+	}
+	defer func() {
+		_ = resp.Body.Close() // Ignore error on close during cleanup
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("token refresh failed with status %d", resp.StatusCode)
```
jhrozek: It would be nice to dump the error logs per https://www.rfc-editor.org/rfc/rfc6749#section-5.2

claude suggested (untested):
```go
  if resp.StatusCode != http.StatusOK {
      body, _ := io.ReadAll(io.LimitReader(resp.Body, 4096))
      rErr := &oauth2.RetrieveError{
          Response: resp,
          Body:     body,
      }
      // Try to parse RFC 6749 error fields
      var errJSON struct {
          Error       string `json:"error"`
          Description string `json:"error_description"`
          URI         string `json:"error_uri"`
      }
      if json.Unmarshal(body, &errJSON) == nil {
          rErr.ErrorCode = errJSON.Error
          rErr.ErrorDescription = errJSON.Description
          rErr.ErrorURI = errJSON.URI
      }
      return nil, fmt.Errorf("token refresh with resource parameter failed: %w", rErr)
  }
```

pkg/auth/oauth/resource_token_source.go
jhrozek: we could then also have a test that includes Token.Valid:
```
  assert.False(t, token.Expiry.IsZero(), "refreshed token should have expiry set")
  assert.True(t, token.Expiry.After(time.Now()), "refreshed token should expire in the future")
  assert.True(t, token.Valid(), "refreshed token should be valid")
```

jhrozek (commented):
I think this is a nice addition, thank you for the PR! I have some comments, I mostly think the one about ExpiresIn/Expiry is important.

I also wonder if ResourceTokenSource should rather be a reusable component in pkg/oauth rather than pkg/auth/oauth? But that's a nit and the code can be moved whenever

pkg/auth/oauth/resource_token_source.go
JAORMX: Good catch on the `ExpiresIn`/`Expiry` issue, Jakub. That's definitely the most important one here.

So, basically, instead of manually parsing the token response and having to deal with all these edge cases ourselves (the `ExpiresIn` conversion, auth style, refresh token preservation, error parsing...), there's actually a neat workaround: we can use [`oauth2.Config.Exchange`](https://pkg.go.dev/golang.org/x/oauth2#Config.Exchange) with [`oauth2.SetAuthURLParam`](https://pkg.go.dev/golang.org/x/oauth2#SetAuthURLParam) overrides.

Note that despite the name, `SetAuthURLParam` applies to the **token endpoint POST body**, not just the auth URL. So you can override `grant_type` to `refresh_token` and tack on the `resource` parameter. Something like:

```go
func (r *resourceTokenSource) Token() (*oauth2.Token, error) {
    if r.refreshToken == "" {
        return nil, fmt.Errorf("oauth2: token expired and refresh token is not set")
    }

    opts := []oauth2.AuthCodeOption{
        oauth2.SetAuthURLParam("grant_type", "refresh_token"),
        oauth2.SetAuthURLParam("refresh_token", r.refreshToken),
        oauth2.SetAuthURLParam("resource", r.resource),
    }

    token, err := r.config.Exchange(context.Background(), "", opts...)
    if err != nil {
        return nil, fmt.Errorf("token refresh with resource parameter failed: %w", err)
    }

    // Preserve refresh token if the server rotated it
    if token.RefreshToken != "" {
        r.refreshToken = token.RefreshToken
    }

    return token, nil
}
```

This gets you [`expires_in` handling](https://pkg.go.dev/golang.org/x/oauth2#Token.ExpiresIn), [auth style auto-detection](https://pkg.go.dev/golang.org/x/oauth2#AuthStyle), refresh token preservation, extra fields via [`Token.Extra()`](https://pkg.go.dev/golang.org/x/oauth2#Token.Extra), and structured error parsing via [`oauth2.RetrieveError`](https://pkg.go.dev/golang.org/x/oauth2#RetrieveError) for free from the standard library's battle-tested internals. The only downside is a harmless empty `code=` in the POST body, which compliant servers will just ignore.

This is a known workaround since `golang.org/x/oauth2` doesn't support custom refresh parameters (open issue https://github.com/golang/oauth2/issues/521).

With this in mind, it would simplify this whole file quite a bit and address all the review comments in one shot.

pkg/auth/oauth/resource_token_source.go
```diff
@@ -0,0 +1,134 @@
...
+	if baseTransport == nil {
+		baseTransport = &http.Transport{}
+	}
+
+	return &resourceTokenSource{
+		config:   config,
+		resource: resource,
+		token:    token,
+		httpClient: &http.Client{
+			Transport: &limitResponseBodyRoundTripper{next: baseTransport, limit: 4096},
```
jhrozek: sorry, I think mine and @JAORMX 's comments were conflicting in the case of limiting of the responses.

I think since you implemented @JAORMX 's suggestion (which I think makes sense!) we no longer need the limit reader because one of the nice things we get from using the stdlib is its round-tripper:
https://cs.opensource.google/go/x/oauth2/+/master:internal/token.go;l=258

So my suggestion would be to remove limitResponseBodyRoundTripper and limitedReadCloser and probably also baseTransport and just use:
```go
  httpClient: &http.Client{
      Timeout: 30 * time.Second,
  },
```

PR #3725, opened 2026-02-09
adds claude skill for setting up otel in kind

ChrisJBurns:
Using a `task` to setup the OTEL stack in Kind can work, but is error prone over time when charts and versions change. 

This PR:
-  adds a claude skill that deploys the OTEL stack and Grafana/Promtheus/Tempo for when we are devving the telemetry capability
- removes jaeger in favour of tempo due to compatibility with grafana
- enables claude agent teams

### DIFF
--- a/pkg/telemetry/middleware.go
+++ b/pkg/telemetry/middleware.go
@@ -38,6 +38,11 @@ const (
 	networkProtocolHTTP = "http"
 )
 
+// MCPHistogramBuckets are the bucket boundaries defined by the MCP OTEL semantic conventions
+// for MCP server histograms (e.g. mcp.server.operation.duration).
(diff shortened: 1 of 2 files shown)


.claude/settings.json
```diff
@@ -31,5 +31,8 @@
       "Bash(pre-commit run:*)"
     ],
     "deny": []
+  },
+  "env": {
+    "CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS": "1"
```
jhrozek, line 36: should we commit this?

ChrisJBurns: Yeah I thought I'd commit only to enable it across the board. Happy to remove it though if folks don't want it enabled by default

jhrozek: it's in my CC anyway, I can't stand not having the latest stuff

jhrozek (approved):

Issue #3569, opened 2026-02-03
AWS STS: Integrate into CLI runner middleware


jhrozek:
Connect AWS STS authentication to ToolHive's middleware system.

The middleware handles the complete flow: extract JWT claims from auth context, select IAM role, exchange token via STS, sign request with SigV4, forward to AWS MCP Server.

Depends on: #3568

Ref: THV-0033

PR #3726, opened 2026-02-09
Integrate AWS STS token exchange with SigV4 signing middleware

jhrozek:
Add an HTTP middleware that converts incoming OIDC bearer tokens into SigV4-signed AWS requests. The pipeline extracts JWT claims from the auth context, selects an IAM role via the RoleMapper, calls STS AssumeRoleWithWebIdentity for temporary credentials, and signs the outbound request with SigV4.

When a target URL is configured, the middleware clones the request with the target host for signing purposes but copies only the SigV4 headers (Authorization, X-Amz-Date, X-Amz-Security-Token) back to the original request, leaving host rewriting to the reverse proxy. Request bodies are buffered with a 10 MB size limit and restored for downstream handlers.

Wire the middleware into the runner: add AWSStsConfig to RunConfig, a WithAWSStsConfig builder option, factory registration in GetSupportedMiddlewareFactories, and placement in
PopulateMiddlewareConfigs after audit/authz but before header forwarding.

Also add GetService() and GetSessionDuration() config helpers, an ErrAccessDenied sentinel, and clean up the signer's body-close handling to use defer.

Fixes: #3569

### DIFF
--- a/pkg/auth/awssts/middleware.go
+++ b/pkg/auth/awssts/middleware.go
@@ -1,0 +1,302 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package awssts provides AWS STS token exchange with SigV4 signing support.
+package awssts
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+
+	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/logger"
+	"github.com/stacklok/toolhive/pkg/transport/types"
+)
+
+// Middleware type constant
+const (
+	MiddlewareType = "awssts"
+)
+
+// Default session name claim when not specified in config.
+const defaultSessionNameClaim = "sub"
+
+// MiddlewareParams represents the parameters for AWS STS middleware.
+type MiddlewareParams struct {
+	AWSStsConfig *Config `json:"aws_sts_config,omitempty"`
+	// TargetURL is the remote MCP server URL for SigV4 signing.
+	// The request must be signed with the target host, not the proxy host.
+	TargetURL string `json:"target_url,omitempty"`
+}
+
+// Middleware wraps AWS STS middleware functionality.
+type Middleware struct {
+	middleware types.MiddlewareFunction
+	exchanger  *Exchanger
+}
+
+// Handler returns the middleware function used by the proxy.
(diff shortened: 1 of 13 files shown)


pkg/auth/awssts/middleware.go
```diff
@@ -0,0 +1,302 @@
...
+	logger.Debugf("Signing request for target host: %s", targetURL.Host)
+
+	if err := signer.SignRequest(r.Context(), signingReq, creds); err != nil {
+		return err
+	}
+
+	// Copy only the SigV4 headers back — these are the only headers the
+	// AWS SDK v4 signer sets during SignHTTP.
+	r.Header.Set("Authorization", signingReq.Header.Get("Authorization"))
+	r.Header.Set("X-Amz-Date", signingReq.Header.Get("X-Amz-Date"))
```
jhrozek, line 268: signing a copy and then copying the headers seemed like a decent compromise between the simplest option (just rewrite the r.Host field here) and the cleanest option (separate STS and signing middlewares) which seemed excessivelly complex at the same time.

pkg/auth/awssts/signer.go
```diff
@@ -135,6 +135,8 @@ func (*requestSigner) hashPayload(req *http.Request) (string, []byte, error) {
 		return emptySHA256, nil, nil
 	}
 
+	defer func() { _ = req.Body.Close() }()
```
jhrozek, line 138: drive-by fix

pkg/auth/awssts/signer_test.go
```diff
@@ -103,19 +109,22 @@ func TestSigner_SignRequest(t *testing.T) {
...
 		req, _ := http.NewRequestWithContext(ctx, "POST", "https://aws-mcp.us-east-1.api.aws/mcp", nil)
 
 		err := signer.SignRequest(ctx, req, nil)
 		require.Error(t, err)
 	})
 }
 
-//nolint:paralleltest // Tests share signer state
 func TestSigner_HashPayload(t *testing.T) {
+	t.Parallel()
```
jhrozek, line 122: drive-by fix

pkg/auth/awssts/middleware.go
```diff
@@ -0,0 +1,302 @@
...
+	return m.middleware
+}
+
+// Close cleans up any resources used by the middleware.
+func (*Middleware) Close() error {
+	return nil
+}
+
+// CreateMiddleware is the factory function for AWS STS middleware.
+func CreateMiddleware(config *types.MiddlewareConfig, runner types.MiddlewareRunner) error {
```
jhrozek, line 56: FYI this feature is k8s-only, the middleware will not be exposed through k8s flags.

jhrozek:
the PR got too big, I'll split and open smaller ones

PR #3729, opened 2026-02-09
Add UseLegacyAttributes flag to telemetry config

ChrisJBurns:
## Summary
- Adds a new `UseLegacyAttributes` boolean field to the telemetry `Config` struct, defaulting to `true` for backward compatibility with existing dashboards and alerts
- Plumbs the flag through CLI (`--otel-use-legacy-attributes`), config file (`use-legacy-attributes`), operator CRD types (`OpenTelemetryConfig.useLegacyAttributes`), operator spectoconfig (reads from CR spec), runner config builder
- CRDs updated: VirtualMCPServer, MCPServer, MCPRemoteProxy — all include `useLegacyAttributes` with `default: true`
- The flag is **not yet consumed** by the telemetry middleware; a follow-up PR will add the conditional attribute emission logic

## Context

This is the first PR in a series to support the MCP OTEL semantic conventions migration. The flag provides the plumbing so that a subsequent PR can conditionally emit legacy attribute names alongside the new standard names. The default will change to `false` in a future release, and the flag will eventually be removed.

## Test plan
- [x] Unit tests for `getTelemetryFromFlags` updated to cover `UseLegacyAttributes` plumbing (CLI precedence, config fallback, default behavior)
- [x] Runner config builder tests updated with new parameter
- [x] Spectoconfig tests updated with `UseLegacyAttributes` assertion and new test case for reading from CR spec
- [x] CRD manifests regenerated for VirtualMCPServer, MCPServer, and MCPRemoteProxy
- [x] `go build ./...` compiles cleanly
- [x] `task lint` passes with 0 issues
- [x] All affected package tests pass

### DIFF
--- a/pkg/registry/data/registry.json
+++ b/pkg/registry/data/registry.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/stacklok/toolhive/main/pkg/registry/data/toolhive-legacy-registry.schema.json",
   "version": "1.0.0",
-  "last_updated": "2026-02-09T00:24:33Z",
+  "last_updated": "2026-02-10T00:28:53Z",
   "servers": {
     "adb-mysql-mcp-server": {
       "description": "Official MCP server for AnalyticDB for MySQL of Alibaba Cloud",
@@ -877,9 +877,9 @@
         "search_engine_batch"
       ],
       "metadata": {
-        "stars": 1947,
+        "stars": 2007,
         "pulls": 65,
-        "last_updated": "2026-01-25T13:38:14Z"
+        "last_updated": "2026-02-09T03:01:53Z"
       },
       "repository_url": "https://github.com/brightdata/brightdata-mcp",
       "tags": [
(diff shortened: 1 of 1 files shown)


cmd/thv/app/run_flags.go
```diff
@@ -912,8 +918,9 @@ func getOidcFromFlags(cmd *cobra.Command) (string, string, string, string, strin
 
 // getTelemetryFromFlags extracts telemetry configuration from command flags
 func getTelemetryFromFlags(cmd *cobra.Command, config *cfg.Config, otelEndpoint string, otelSamplingRate float64,
-	otelEnvironmentVariables []string, otelInsecure bool, otelEnablePrometheusMetricsPath bool) (
-	string, float64, []string, bool, bool) {
+	otelEnvironmentVariables []string, otelInsecure bool, otelEnablePrometheusMetricsPath bool,
+	otelUseLegacyAttributes bool) (
+	string, float64, []string, bool, bool, bool) {
```
jhrozek, line 923: seems to be ripe for refactoring at some point

ChrisJBurns: Yeah I agree! I recon when we add the `Telemetry` CRD, it would be a good time to do it

ChrisJBurns: Yeah I agree! I recon when we add the `Telemetry` CRD, it would be a good time to do it

PR #3730, opened 2026-02-09
Add SigV4 signing middleware for AWS STS token exchange

jhrozek:
Add an HTTP middleware that converts incoming OIDC bearer tokens into SigV4-signed AWS requests. The pipeline extracts JWT claims from the auth context, selects an IAM role via the RoleMapper, calls STS AssumeRoleWithWebIdentity for temporary credentials, and signs the outbound request with SigV4.

When a target URL is configured, the middleware clones the request with the target host for signing purposes but copies only the SigV4 headers (Authorization, X-Amz-Date, X-Amz-Security-Token) back to the original request, leaving host rewriting to the reverse proxy. Request bodies are buffered with a 10 MB size limit and restored for downstream handlers.

Also add GetService() and GetSessionDuration() config helpers, an ErrAccessDenied sentinel, and clean up the signer's body-close handling to use defer.

Fixes: #3569

### DIFF
--- a/pkg/auth/awssts/middleware.go
+++ b/pkg/auth/awssts/middleware.go
@@ -184,11 +184,16 @@ func createAWSStsMiddlewareFunc(
 				return
 			}
 
-			// Extract session name from claims
+			// Extract and validate session name from claims
 			sessionName, err := extractSessionName(claims, sessionNameClaim)
 			if err != nil {
 				logger.Warnf("Failed to extract session name: %v", err)
 				http.Error(w, "Missing session name claim", http.StatusUnauthorized)
+				return
+			}
+			if err := ValidateSessionName(sessionName); err != nil {
+				logger.Warnf("Invalid session name %q: %v", sessionName, err)
+				http.Error(w, "Invalid session name", http.StatusUnauthorized)
 				return
 			}
 


pkg/auth/awssts/errors.go
```diff
@@ -39,4 +39,7 @@ var (
 
 	// ErrSTSNilCredentials is returned when STS returns a response without credentials.
 	ErrSTSNilCredentials = errors.New("STS returned nil credentials")
+
+	// ErrAccessDenied is returned when AWS STS denies access.
+	ErrAccessDenied = errors.New("access denied")
```
ChrisJBurns: This is only used in the test mock, is that right?

pkg/auth/awssts/middleware.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	// Create the SigV4 signer
+	signer, err := newRequestSigner(cfg.Region, withService(cfg.GetService()))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create SigV4 signer: %w", err)
+	}
+
+	// Determine session name claim
+	sessionNameClaim := cfg.SessionNameClaim
```
ChrisJBurns, line 121: Could we reuse the session name validation from previous PRs?

pkg/auth/awssts/errors.go
jhrozek: yeah, good catch, I moved the error to the test and made it unexported.

pkg/auth/awssts/middleware.go
jhrozek: that's what I get for cherry-picking commits from a single branch with stack commits :-) I forgot to reuse that, added!

ChrisJBurns (approved):

PR #3748, opened 2026-02-10
Rename span attributes to MCP OTEL semantic conventions

ChrisJBurns:
## Summary

- Adds new standard OTEL semantic convention attribute names alongside the gated legacy names
- New attributes are always emitted; legacy names are conditionally emitted behind `UseLegacyAttributes` (from #3747)
- Adds `mapTransport` helper to convert MCP transport types to OTEL `network.transport` values
- Updates span naming: `mcp.tools/call` → `tools/call` (method only, no resource target to avoid cardinality issues)
- Renames `rpc.system` to `rpc.system.name` (per OTEL deprecation), moves `rpc.system` to legacy block
- Emits `network.protocol.name` and `network.protocol.version` in SSE connection spans

Key attribute renames (new names always emitted, old names gated):
| Old (legacy) | New (standard) |
|-----|-----|
| `http.method` | `http.request.method` |
| `http.url` | `url.full` |
| `http.status_code` | `http.response.status_code` |
| `mcp.method` | `mcp.method.name` |
| `mcp.request.id` | `jsonrpc.request.id` |
| `mcp.tool.name` | `gen_ai.tool.name` |
| `mcp.tool.arguments` | `gen_ai.tool.call.arguments` |
| `mcp.transport` | `network.transport` |
| `mcp.resource.id` | `mcp.resource.uri` |
| `rpc.system` | `rpc.system.name` |

New attributes with no legacy equivalent:
- `jsonrpc.protocol.version`
- `gen_ai.operation.name`, `gen_ai.prompt.name`
- `network.protocol.name`, `url.scheme`, `url.path`, `url.query`
- `server.address`, `user_agent.original`
- `http.request.body.size`, `http.response.body.size`

Transport mapping:
| Transport | `network.transport` |
|-----------|-----|
| `stdio` | `pipe` |
| `sse` | `tcp` |
| `streamable-http` | `tcp` |

Stacked on #3747 (gates legacy attributes behind the flag).

## Test plan

- [x] Unit tests for dual emission (legacy enabled: both old and new present; legacy disabled: only new present)
- [x] Unit tests for `mapTransport` helper
- [x] Unit tests for `createSpanName` format (method only, no resource ID)
- [x] Unit tests for `mcp.resource.uri` conditional logic (present for `resources/read`, absent for `tools/call`)
- [x] Unit tests for span status verification (`codes.Ok` for 2xx, `codes.Error` for 5xx)
- [x] Integration test verifies both old and new attribute names present
- [x] `task lint` passes
- [x] `task test` passes

### DIFF
--- a/pkg/telemetry/middleware.go
+++ b/pkg/telemetry/middleware.go
@@ -29,9 +29,6 @@ import (
 const (
 	// instrumentationName is the name of this instrumentation package
 	instrumentationName = "github.com/stacklok/toolhive/pkg/telemetry"
-	// mcpProtocolVersion is the MCP protocol version reported in spans.
-	// TODO: read from MCP-Protocol-Version header when available.
-	mcpProtocolVersion = "2025-06-18"
 	// methodPromptsGet is the MCP method name for prompts/get
 	methodPromptsGet = "prompts/get"
 	// networkTransportTCP is the OTEL value for TCP transport
@@ -249,7 +246,6 @@ func (m *HTTPMiddleware) addMCPAttributes(ctx context.Context, span trace.Span, 
 	span.SetAttributes(
 		attribute.String("mcp.method.name", parsedMCP.Method),
 		attribute.String("rpc.system.name", "jsonrpc"),
-		attribute.String("mcp.protocol.version", mcpProtocolVersion),
 		attribute.String("jsonrpc.protocol.version", "2.0"),
 	)
 
--- a/pkg/telemetry/middleware_test.go
+++ b/pkg/telemetry/middleware_test.go
@@ -1613,7 +1613,6 @@ func TestHTTPMiddleware_LegacyAttributes_Disabled(t *testing.T) {
 				assert.Contains(t, span.attributes, "mcp.method.name")
 				assert.Contains(t, span.attributes, "rpc.system.name")
 				assert.Contains(t, span.attributes, "jsonrpc.request.id")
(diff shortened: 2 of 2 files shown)


pkg/telemetry/middleware.go
```diff
@@ -219,20 +242,37 @@ func (m *HTTPMiddleware) addMCPAttributes(ctx context.Context, span trace.Span,
 		return
 	}
 
+	// New OTEL MCP semantic convention attributes (always emitted)
+	span.SetAttributes(
+		attribute.String("mcp.method.name", parsedMCP.Method),
```
jhrozek, line 247: If someone like me has no idea where these come from: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/

pkg/telemetry/middleware.go
```diff
@@ -331,6 +385,19 @@ func (m *HTTPMiddleware) extractBackendTransport(r *http.Request) string {
...
 }
 
+func mapTransport(mcpTransport string) (networkTransport, protocolName, protocolVersion string) {
+	switch mcpTransport {
+	case "stdio":
+		return "pipe", "", ""
+	case "sse":
+		return networkTransportTCP, networkProtocolHTTP, "1.1"
+	case "streamable-http":
+		return networkTransportTCP, networkProtocolHTTP, "2"
```
jhrozek, line 395: is it always http/2 for streamable http or "usually" ?

Issue #3731, opened 2026-02-09
Back the dummyOptimizer with a shared SQLite instance and basic similarity search


jerm-dro:
### Context

The current `dummyOptimizer` (`pkg/vmcp/optimizer/dummy_optimizer.go`) uses an in-memory map with simple case-insensitive substring matching. This doesn't scale to production workloads and duplicates tool state across every optimizer instance (one per MCP session).

Prototype reference: https://github.com/stacklok/toolhive/pull/3516 (see `pkg/vmcp/optimizer/internal/db/` for prior art on SQLite + FTS5)

### Requirements

1. Introduce a **storage interface** that abstracts how tools are stored and searched. The `dummyOptimizer` should depend on this interface rather than directly on SQLite. This enables:
   - Unit testing the optimizer with a mock/in-memory store
   - Integration testing the SQLite implementation independently
2. Implement the storage interface with a **shared SQLite backend** backed by FTS5 full-text search. The SQLite instance must be shared across all optimizer instances created by the `OptimizerFactory` (`pkg/vmcp/server/server.go:134`), deduplicating tool storage.
3. Despite the shared database, each optimizer instance must **only return tools for the capabilities passed at factory creation time** (i.e., the `[]server.ServerTool` slice passed to `NewDummyOptimizer`). This preserves per-session scoping.
4. Implement a basic BM25/FTS5-based similarity search in `FindTool`, replacing the current substring matching.
5. The `Optimizer` interface (`pkg/vmcp/optimizer/optimizer.go`) should gain a `Close()` method so the database can be cleaned up. Update the server integration in `pkg/vmcp/server/server.go` to call `Close()` on session teardown.
6. The factory function signature in `server.Config.OptimizerFactory` may need to change to accept a shared storage reference. Consider a two-phase pattern: a top-level "optimizer manager" (holds the storage instance) and per-session optimizer instances.

### High-Level Implementation

- Define a storage interface (e.g., `ToolStore`) in `pkg/vmcp/optimizer/` with methods for tool ingestion, search, and cleanup
- Create an internal SQLite implementation under `pkg/vmcp/optimizer/internal/db/` that satisfies the interface:
  - A single shared `*sql.DB` connection with FTS5 virtual table for tool search
  - Tool ingestion (upsert tools with name, description, input schema)
  - FTS5-based search that returns ranked results
- Update `dummyOptimizer` to accept a `ToolStore` and a scoping set of tool names
[...]

PR #3749, opened 2026-02-10
Add ToolStore interface and refactor DummyOptimizer to use it

aponcedeleonch:
Related to: https://github.com/stacklok/toolhive/issues/3731

PR 1/2

Introduce a ToolStore abstraction that separates tool storage and search from the Optimizer interface. This lays the foundation for replacing the in-memory substring matching with a SQLite FTS5 backend in a follow-up PR.

Changes:
- Add ToolStore interface (UpsertTools, Search, Close) in store.go
- Add InMemoryToolStore implementation preserving existing behavior
- Refactor DummyOptimizer to delegate search to a shared ToolStore with per-instance scope filtering
- Update OptimizerFactory signature to return (Optimizer, error)
- Update commands.go wiring to create shared InMemoryToolStore
- Add comprehensive tests for InMemoryToolStore

### DIFF
--- a/pkg/vmcp/optimizer/dummy_optimizer.go
+++ b/pkg/vmcp/optimizer/dummy_optimizer.go
@@ -32,13 +32,13 @@ type DummyOptimizer struct {
 //
 // The tools slice should contain all backend tools (as ServerTool with handlers).
 // Tools are upserted into the shared store and scoped for this optimizer instance.
-func NewDummyOptimizer(store ToolStore, tools []server.ServerTool) (Optimizer, error) {
+func NewDummyOptimizer(ctx context.Context, store ToolStore, tools []server.ServerTool) (Optimizer, error) {
 	toolMap := make(map[string]server.ServerTool, len(tools))
 	for _, tool := range tools {
 		toolMap[tool.Tool.Name] = tool
 	}
 
-	if err := store.UpsertTools(context.Background(), tools); err != nil {
+	if err := store.UpsertTools(ctx, tools); err != nil {
 		return nil, fmt.Errorf("failed to upsert tools into store: %w", err)
 	}
 
@@ -104,9 +104,9 @@ func (d *DummyOptimizer) toolNames() []string {
 // NewDummyOptimizerFactory returns an OptimizerFactory that creates DummyOptimizer
 // instances backed by a shared InMemoryToolStore. All optimizers created by the
 // returned factory share the same underlying storage, enabling cross-session search.
-func NewDummyOptimizerFactory() func([]server.ServerTool) (Optimizer, error) {
+func NewDummyOptimizerFactory() func(context.Context, []server.ServerTool) (Optimizer, error) {
 	store := NewInMemoryToolStore()
-	return func(tools []server.ServerTool) (Optimizer, error) {
-		return NewDummyOptimizer(store, tools)
+	return func(ctx context.Context, tools []server.ServerTool) (Optimizer, error) {
+		return NewDummyOptimizer(ctx, store, tools)
 	}
 }
--- a/pkg/vmcp/server/server.go
+++ b/pkg/vmcp/server/server.go
@@ -138,7 +138,7 @@ type Config struct {
 
 	// OptimizerFactory builds an optimizer from a list of tools.
 	// If not set, the optimizer is disabled.
-	OptimizerFactory func([]server.ServerTool) (optimizer.Optimizer, error)
+	OptimizerFactory func(context.Context, []server.ServerTool) (optimizer.Optimizer, error)
 
 	// StatusReporter enables vMCP runtime to report operational status.
 	// In Kubernetes mode: Updates VirtualMCPServer.Status (requires RBAC)
@@ -885,6 +885,7 @@ func (s *Server) injectCapabilities(
 // 1. Converts all tools (backend + composite) to SDK format with handlers
 // 2. Injects the optimizer capabilities into the session
 func (s *Server) injectOptimizerCapabilities(
+	ctx context.Context,
 	sessionID string,
 	caps *aggregator.AggregatedCapabilities,
 ) error {
@@ -898,7 +899,7 @@ func (s *Server) injectOptimizerCapabilities(
 	}
 
 	// Create optimizer instance from factory
-	opt, err := s.config.OptimizerFactory(sdkTools)
+	opt, err := s.config.OptimizerFactory(ctx, sdkTools)
 	if err != nil {
 		return fmt.Errorf("failed to create optimizer: %w", err)
 	}
@@ -1015,7 +1016,7 @@ func (s *Server) handleSessionRegistration(
 		"prompt_count", len(caps.RoutingTable.Prompts))
 
 	if s.config.OptimizerFactory != nil {
-		err = s.injectOptimizerCapabilities(sessionID, caps)
+		err = s.injectOptimizerCapabilities(ctx, sessionID, caps)
 		if err != nil {
 			logger.Errorw("failed to create optimizer tools",
 				"error", err,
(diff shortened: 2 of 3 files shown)


pkg/vmcp/optimizer/dummy_optimizer.go
```diff
@@ -5,72 +5,67 @@ package optimizer
...
-		toolMap[tool.Tool.Name] = tool
+		scope = append(scope, tool.Tool.Name)
+		handlers[tool.Tool.Name] = tool
 	}
 
-	return DummyOptimizer{
-		tools: toolMap,
+	// Upsert tools into the shared store (best-effort; errors logged at call site if needed)
+	//nolint:errcheck // UpsertTools for InMemoryToolStore never returns an error
+	_ = store.UpsertTools(context.Background(), tools)
```
jerm-dro: blocker: let's return the errors. Swallowing errors could later cause confusing / misbehaving similarity search.

pkg/vmcp/optimizer/dummy_optimizer.go
```diff
@@ -5,72 +5,67 @@ package optimizer
...
+
+	return &DummyOptimizer{
+		store:    store,
+		scope:    scope,
```
jerm-dro: nit: "scope" is an unintuitive word in this context. I'd recommend calling this something simple like "allowedTools"

pkg/vmcp/optimizer/store.go
```diff
@@ -0,0 +1,123 @@
...
+	UpsertTools(ctx context.Context, tools []server.ServerTool) error
+
+	// Search finds tools matching the query string.
+	// The scope parameter limits results to only tools with names in the given set.
+	// If scope is empty, all tools are searched.
+	// Returns matches ranked by relevance.
+	Search(ctx context.Context, query string, scope []string) ([]ToolMatch, error)
+
+	// Close releases any resources held by the store.
+	Close() error
```
jerm-dro: When should close be called? When the vMCP server shutsdown or the client (e.g. a single optimizer) is done using it?

cmd/vmcp/app/commands.go
```diff
@@ -484,7 +485,10 @@ func runServe(cmd *cobra.Command, _ []string) error {
 
 	if cfg.Optimizer != nil {
 		// TODO: update this with the real optimizer.
-		serverCfg.OptimizerFactory = optimizer.NewDummyOptimizer
+		store := optimizer.NewInMemoryToolStore()
+		serverCfg.OptimizerFactory = func(tools []mcpserver.ServerTool) (optimizer.Optimizer, error) {
+			return optimizer.NewDummyOptimizer(store, tools), nil
+		}
```
jerm-dro: Let's move these lines into `optimizer` so we can write a test that optimizers created from the same factory share the same underlying storage.

pkg/vmcp/optimizer/optimizer.go
```diff
@@ -33,6 +33,9 @@ type Optimizer interface {
 	// Returns the tool's result or an error if the tool is not found or execution fails.
 	// Returns the MCP CallToolResult directly from the underlying tool handler.
 	CallTool(ctx context.Context, input CallToolInput) (*mcp.CallToolResult, error)
+
+	// Close releases resources held by this optimizer instance.
+	Close() error
```
jerm-dro: Blocker: Given the current implementation does nothing on close, let's remove this. Similarly, let's remove this from `ToolStore`.  This will also allow you to delete the changes to `server.go`.

A little background here: the current code for managing sessions and things with session lifetimes is complex. We have a WIP RFC here: https://github.com/stacklok/toolhive-rfcs/pull/38. If we don't need any session cleanup in this PR, let's not add it. If we eventually do, we can add it. Ideally, when it's much easier to do so.

pkg/vmcp/optimizer/store.go
```diff
@@ -0,0 +1,123 @@
...
+	"strings"
+	"sync"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+// ToolStore defines the interface for storing and searching tools.
+// Implementations may use in-memory maps, SQLite FTS5, or other backends.
+type ToolStore interface {
```
jerm-dro, line 22: Nit: you should call out this ToolStore is accessed in parallel and implementations must be thread-safe.

aponcedeleonch: Added some lines to the comment

pkg/vmcp/optimizer/dummy_optimizer.go
aponcedeleonch: you're right, fixed it

pkg/vmcp/optimizer/store.go
aponcedeleonch: yes, removed it for the moment to avoid confussions

pkg/vmcp/optimizer/optimizer.go
aponcedeleonch: sounds good! removed it

cmd/vmcp/app/commands.go
aponcedeleonch: sounds good, done!

pkg/vmcp/optimizer/dummy_optimizer.go
aponcedeleonch: changed to allowedTools

pkg/vmcp/optimizer/store.go
```diff
@@ -0,0 +1,118 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+)
+
+// ToolStore defines the interface for storing and searching tools.
```
yrobla, line 17: shouldn't we have a "Close" method here?

pkg/vmcp/server/server.go
```diff
@@ -897,8 +897,14 @@ func (s *Server) injectOptimizerCapabilities(
 		return fmt.Errorf("failed to convert tools to SDK format: %w", err)
 	}
 
+	// Create optimizer instance from factory
```
yrobla, line 901: pr says:
Track session optimizers in Server for cleanup on shutdown

but i do not see where is it tracked ?

pkg/vmcp/optimizer/store.go
aponcedeleonch: yes, there was a `Close` but in this PR was a no-op so was causing a little bit of confusion in review. Will introduce the method it in the follow up PR

pkg/vmcp/server/server.go
aponcedeleonch: you're right, right now we're not cleaning up anything but it's by design since there's nothing to cleanup yet. Let me fix the description

pkg/vmcp/optimizer/dummy_optimizer.go
```diff
@@ -5,72 +5,61 @@ package optimizer
...
+// Tools are upserted into the shared store and scoped for this optimizer instance.
+func NewDummyOptimizer(store ToolStore, tools []server.ServerTool) (Optimizer, error) {
 	toolMap := make(map[string]server.ServerTool, len(tools))
 	for _, tool := range tools {
 		toolMap[tool.Tool.Name] = tool
 	}
 
-	return DummyOptimizer{
-		tools: toolMap,
+	if err := store.UpsertTools(context.Background(), tools); err != nil {
```
yrobla: i think it would be better if we pass the context as parameter on newdummyoptimizer(ctx...) then we do not have to create the context here

aponcedeleonch: you're right, updated the code

pkg/vmcp/server/adapter/optimizer_adapter_test.go
```diff
@@ -33,6 +33,10 @@ func (m *mockOptimizer) CallTool(ctx context.Context, input optimizer.CallToolIn
 	return mcp.NewToolResultText("ok"), nil
 }
 
+func (*mockOptimizer) Close() error {
+	return nil
+}
```
jerm-dro, line 38: nit: this is unused and not needed

PR #38, opened 2025-03-27
Disable SLSA for now and rename formula release to vt

rdimitrov:
The following PR:
* Renames the formula as vt (so one can do `brew install stacklok/tap/vt` instead of `vibetool`)
* Temporarily disables the SLSA provenance generation (for some reason it fails, did not debug this but might be because the repo is still private since the same workflow works for `frizbee` for example)

PR #3752, opened 2026-02-10
Add `mcp.server.operation.duration` metric per OTEL MCP semantic conventions

ChrisJBurns:
## Summary

- Adds the `mcp.server.operation.duration` histogram metric as defined by the [OTEL MCP semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/mcp.md)
- Records operation latency with spec-defined attributes: `mcp.method.name`, `gen_ai.tool.name`, `gen_ai.operation.name`, `gen_ai.prompt.name`, `jsonrpc.protocol.version`, `network.transport`, `network.protocol.name`, `network.protocol.version`, `error.type`
- Exports `MCPHistogramBuckets` for reuse by future metrics (e.g. `mcp.server.session.duration`)
- Adds warning log when MCP method cannot be determined (indicates middleware misconfiguration)
- Existing custom `toolhive_mcp_*` metrics are preserved for backward compatibility

**Note:** `mcp.server.session.duration` is intentionally deferred — it will be trivial to add after the session management architecture refactor (#38).

## Test plan

- [x] `TestHTTPMiddleware_OperationDuration/tools/call` — verifies metric recorded with tool attributes
- [x] `TestHTTPMiddleware_OperationDuration/prompts/get` — verifies metric recorded with prompt attributes
- [x] `TestHTTPMiddleware_OperationDuration/non-MCP` — verifies metric is NOT recorded for non-MCP requests
- [x] `task lint-fix` passes with 0 issues
- [x] `task test` passes

### DIFF
--- a/pkg/telemetry/middleware.go
+++ b/pkg/telemetry/middleware.go
@@ -38,6 +38,11 @@ const (
 	networkProtocolHTTP = "http"
 )
 
+// MCPHistogramBuckets are the bucket boundaries defined by the MCP OTEL semantic conventions
+// for MCP server histograms (e.g. mcp.server.operation.duration).
+// See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/mcp.md
+var MCPHistogramBuckets = []float64{0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300}
+
 // HTTPMiddleware provides OpenTelemetry instrumentation for HTTP requests.
 type HTTPMiddleware struct {
 	config         Config
@@ -51,6 +56,7 @@ type HTTPMiddleware struct {
 	// Metrics
 	requestCounter    metric.Int64Counter
 	requestDuration   metric.Float64Histogram
+	operationDuration metric.Float64Histogram
 	activeConnections metric.Int64UpDownCounter
 }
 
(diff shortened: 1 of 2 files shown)


pkg/telemetry/middleware.go
```diff
@@ -38,6 +38,10 @@ const (
 	networkProtocolHTTP = "http"
 )
 
+// MCPHistogramBuckets are the bucket boundaries defined by the MCP OTEL semantic conventions
+// for mcp.server.operation.duration and mcp.server.session.duration histograms.
+var MCPHistogramBuckets = []float64{0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5, 10, 30, 60, 120, 300}
```
jerm-dro, line 44: Nit: link to documentation

pkg/telemetry/middleware.go
```diff
@@ -631,6 +644,11 @@ func (m *HTTPMiddleware) recordMetrics(ctx context.Context, r *http.Request, rw
 	// Record request duration
 	m.requestDuration.Record(ctx, duration.Seconds(), attrs)
 
+	// Record OTEL MCP spec mcp.server.operation.duration for actual MCP requests
+	if mcpMethod != "unknown" {
```
jerm-dro, line 651: Question, non-blocking: why would this be unknown? Can we remove this if?

jerm-dro: If this is an unknown, it means the middleware is messed up:

https://github.com/stacklok/toolhive/pull/3687

We should log a warning in this event.

pkg/transport/session/manager.go
```diff
@@ -27,12 +27,50 @@ type Session interface {
 	SetMetadata(key, value string)
 }
 
+// EndReason indicates why a session ended.
+type EndReason int
```
jerm-dro: How would you feel about not merging these bits right now?

@yrobla is working on improving the session management architecture here https://github.com/stacklok/toolhive-rfcs/pull/38

Once as this is implemented, it would be trivial to implement this change and you won't have to touch anything in `vmcp/server.go`.

ChrisJBurns: yep, removed

PR #3770, opened 2026-02-11
Wire AWS STS middleware into the runner middleware chain

jhrozek:
Register the AWS STS middleware in the runner so it can be activated via RunConfig.

Add AWSStsConfig to RunConfig, a WithAWSStsConfig builder option, and factory registration in GetSupportedMiddlewareFactories. Place the middleware in PopulateMiddlewareConfigs after audit/authz but before header forwarding — only authorized requests trigger credential exchange and SigV4 signing happens as late as possible.

Regenerate Swagger docs to include the new awssts.Config and awssts.RoleMapping schemas.

Consolidate the three GetSupportedMiddlewareFactories tests into one table-driven test and extract the duplicated createMinimalAuthServerConfig helper to package level.

Related: #3569

### DIFF
--- a/pkg/runner/config.go
+++ b/pkg/runner/config.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/stacklok/toolhive/pkg/audit"
 	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/auth/awssts"
 	"github.com/stacklok/toolhive/pkg/auth/remote"
 	authsecrets "github.com/stacklok/toolhive/pkg/auth/secrets"
 	"github.com/stacklok/toolhive/pkg/auth/tokenexchange"
@@ -116,6 +117,9 @@ type RunConfig struct {
 	// When set along with EmbeddedAuthServerConfig, this middleware exchanges ToolHive JWTs
 	// for upstream IdP tokens before forwarding requests to the MCP server.
 	UpstreamSwapConfig *upstreamswap.Config `json:"upstream_swap_config,omitempty" yaml:"upstream_swap_config,omitempty"`
+
+	// AWSStsConfig contains AWS STS token exchange configuration for accessing AWS services
+	AWSStsConfig *awssts.Config `json:"aws_sts_config,omitempty" yaml:"aws_sts_config,omitempty"`
 
 	// DEPRECATED: Middleware configuration.
 	// AuthzConfig contains the authorization configuration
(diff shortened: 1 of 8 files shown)


pkg/runner/config.go
```diff
@@ -117,6 +118,9 @@ type RunConfig struct {
 	// for upstream IdP tokens before forwarding requests to the MCP server.
 	UpstreamSwapConfig *upstreamswap.Config `json:"upstream_swap_config,omitempty" yaml:"upstream_swap_config,omitempty"`
 
+	// AWSStsConfig contains AWS STS token exchange configuration for accessing AWS services
+	AWSStsConfig *awssts.Config `json:"aws_sts_config,omitempty" yaml:"aws_sts_config,omitempty"`
```
JAORMX, line 122: is this meant to be middleware configuration? should it go with other middleware? cc @blkt

jhrozek: fwiw I modeled this after the token exchange and header injection code. If there's other pattern I should be using, I'll be happy to do that.

pkg/runner/middleware.go
```diff
@@ -269,3 +280,20 @@ func addUpstreamSwapMiddleware(
 	}
 	return append(middlewares, *upstreamSwapMwConfig), nil
 }
+
+// addAWSStsMiddleware adds AWS STS middleware if configured
+func addAWSStsMiddleware(middlewares []types.MiddlewareConfig, config *RunConfig) ([]types.MiddlewareConfig, error) {
+	if config.AWSStsConfig == nil {
```
ChrisJBurns, line 288: Do we want to check if the `config.RemoteURL` is empty here too? just to avoid the middleware signing against the wrong thing?

pkg/runner/config.go
blkt: I don't know how valuable is it to enforce that distinction anymore. The main problem is that `RunConfig` represents both the MCP server's config and the Proxy's config, and doubts regarding where to put any bit of configuration stem out of that.

I think it's sensible to have each configuration parameter as a top level field of a struct, it's much easier to know where to find things, and have two separate structs for MCP server and Proxy configurations.
Finally, I would leave to the code under `pkg/runner/runner.go` to determine the correct order of middlewares.

This is easier said than done because this requires a data migration. This is one of the reasons why I believe having run configs on database would be a boon, because data migrations are an established practice.

I would not bang the head to hard on this and leave it where it's easy to understand. Having middleware configurations in a list rather than in top level fields gives little advantage.

pkg/runner/middleware.go
jhrozek: Good call, done

pkg/runner/config.go
jhrozek: thank you, as we confirmed on Slack, this can be read as "the current code is good enough for now".

Issue #3648, opened 2026-02-06
Skill state management


JAORMX:
## Description

Implement the hybrid skill state management system in ToolHive's `pkg/skills/` package: per-skill `.thv-skill.json` metadata files and a global `skill-index.json` for fast lookups.

## Context

Part of the [Skills Lifecycle Management epic](https://github.com/stacklok/stacklok-epics/issues/239). Implements state tracking that distinguishes managed (installed via thv) from unmanaged (manual) skills. Uses a hybrid approach: per-skill files for resilience, global index for performance.

**Dependencies**: stacklok/toolhive#3645
**Blocks**: Skill installer, Skill service implementation

## Acceptance Criteria

- [ ] `pkg/skills/state.go` manages per-skill `.thv-skill.json` metadata files
- [ ] `pkg/skills/index.go` manages `skill-index.json` at `$XDG_STATE_HOME/toolhive/skills/`
- [ ] Atomic writes: write to `.tmp` file, rename (per-skill files)
- [ ] File locking: `flock` for index file (uses `gofrs/flock`)
- [ ] Index rebuild: can reconstruct index from per-skill metadata files
- [ ] CRUD operations: `WriteMetadata`, `ReadMetadata`, `DeleteMetadata`, `ListManaged`
- [ ] Index operations: `AddToIndex`, `RemoveFromIndex`, `GetFromIndex`, `ListFromIndex`, `RebuildIndex`
- [ ] Composite key format: `{name}:{client}:{scope}` for index entries
- [ ] Tests with isolated temp directories, SPDX headers

## Technical Approach

Use ToolHive's existing `pkg/state/` patterns with skill-specific storage.

### Code Pointers
- `pkg/state/local.go` — LocalStore for XDG path patterns
- Skillet `internal/infrastructure/filesystem/metadata.go` — `.skillet.json` management
- TASK-001 `pkg/skills/types.go` — `SkillMetadata`, `SkillIndex` types

## References

- Epic: stacklok/stacklok-epics#239
- stacklok/toolhive#3645

JAORMX:
**Sub-tasks and next steps:**

This story now has 3 sub-issues for sequential implementation:
1. #3738 — Storage interfaces, errors, noop, factory
2. #3739 — SQLite foundation (db, PRAGMAs, migrations)  
3. #3740 — SQLite SkillStore implementation

Once complete, work continues in #3741 (Phase 1: basic SkillService + API handlers), which depends on this story.

[...]

PR #3771, opened 2026-02-11
Add SQLite foundation with connection, PRAGMAs, and goose migrations

### LINKED ISSUE #3739: SQLite database foundation — connection, PRAGMAs, and migrations
## Description

Create the `pkg/storage/sqlite/` package with database connection management, PRAGMA configuration, and the embedded SQL migration runner. This is the SQLite infrastructure layer — no business logic yet.

Adds `modernc.org/sqlite` (pure Go, no CGo) as a dependency. Migrations auto-run inside `Open()` so every process (CLI one-shot or `thv serve`) gets a consistent schema.

**Parent**: #3648
**RFC**: https://github.com/stacklok/toolhive-rfcs/pull/41

## Acceptance Criteria

- [ ] `modernc.org/sqlite` added to `go.mod`
- [ ] `pkg/storage/sqlite/db.go` — `Open()` (singleton at `$XDG_STATE_HOME/toolhive/toolhive.db`) and `OpenWithPath(path)` (for tests)
- [ ] PRAGMA configuration on every connection: `journal_mode=WAL`, `busy_timeout=5000`, `synchronous=NORMAL`, `foreign_keys=ON`, `cache_size=-2000`
- [ ] DSN includes `_txlock=immediate`
- [ ] `MaxOpenConns(1)` per process
- [ ] `pkg/storage/sqlite/migrations.go` — embedded SQL migration runner with `schema_version` table, runs inside `Open()`
- [ ] `pkg/storage/sqlite/migrations/001_create_skills.sql` — `installed_skills` table with `(name, scope)` PK, CHECK constraints, index on status
- [ ] Migrations are idempotent (skip already-applied versions)
- [ ] Concurrent migration safety (second process blocks, then skips)
- [ ] `Close()` function for cleanup
- [ ] SPDX headers on all files
- [ ] Tests: db open/close, PRAGMA verification, migration idempotency, `OpenWithPath(t.TempDir())`

## Schema

```sql
CREATE TABLE installed_skills (
    name         TEXT NOT NULL,
    scope        TEXT NOT NULL DEFAULT 'user' CHECK (scope IN ('user', 'system')),
    version      TEXT NOT NULL DEFAULT '',
    description  TEXT NOT NULL DEFAULT '',
    author       TEXT NOT NULL DEFAULT '',
    tags         TEXT NOT NULL DEFAULT '[]',
    status       TEXT NOT NULL DEFAULT 'pending' CHECK (status IN ('installed', 'pending', 'failed')),
    installed_at TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ', 'now')),
    clients      TEXT NOT NULL DEFAULT '[]',
    PRIMARY KEY (name, scope)
);
```

## Code Pointers

- `pkg/config/singleton.go` — singleton pattern reference
- `pkg/state/local.go` — XDG directory path pattern
- `.goreleaser.yaml` — confirms `CGO_ENABLED=0` (must use pure Go driver)

## Size Estimate

~195 LOC code, 3 files + go.mod/go.sum (excluding tests)

JAORMX:
## Summary

- Introduce `pkg/storage/sqlite/` package with SQLite database connection management, PRAGMA configuration, and embedded migration runner
- Add `modernc.org/sqlite` (pure Go, no CGo) and `github.com/pressly/goose/v3` dependencies
- Create full RFC-0041 schema: `entries`, `installed_skills`, `skill_dependencies`, `oci_tags` tables

This is PR 2 of 3 for SQLite-backed skill state management ([RFC-0041](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0041-sqlite-state-management.md)).

Resolves #3739
Part of #3648

## Details

**db.go**: `Open(dbPath)` constructor with:
- Parent directory creation (0750)
- DSN: `file:<path>?_txlock=immediate`
- `SetMaxOpenConns(1)` for single-writer serialization
- PRAGMAs: `journal_mode=WAL`, `busy_timeout=5000`, `synchronous=NORMAL`, `foreign_keys=ON`, `cache_size=-2000`
- `DefaultDBPath()` helper using XDG (only xdg touchpoint)

**migrations.go**: Goose v3 Provider API with `//go:embed migrations/*.sql`

**001_create_entries_and_skills.sql**: 4 tables with CHECK constraints, CASCADE deletes, JSONB BLOBs, RFC 3339 timestamps

## Test plan

- [x] `pkg/storage/sqlite/db_test.go` — open/close, PRAGMA verification, directory creation, MaxOpenConns, DefaultDBPath
- [x] `pkg/storage/sqlite/migrations_test.go` — table existence, idempotent re-run, CHECK constraints
- [x] `task test` — full suite passes
- [x] `task docs` — no changes needed
- [x] `task build` — compiles cleanly


pkg/storage/sqlite/db.go
```diff
@@ -0,0 +1,103 @@
...
+
+// DB wraps a *sql.DB connection to a SQLite database.
+type DB struct {
+	db *sql.DB
+}
+
+// Open opens (or creates) a SQLite database at the given path. It ensures the
+// parent directory exists, configures recommended PRAGMAs for WAL mode, runs
+// any pending migrations, and verifies the connection before returning.
+func Open(dbPath string) (*DB, error) {
```
ChrisJBurns: Should this accept `context.Context`  to be able to control timeouts etc?

pkg/storage/sqlite/db.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, fmt.Errorf("opening database: %w", err)
+	}
+
+	// If setup fails after opening, close the connection before returning.
+	success := false
+	defer func() {
+		if !success {
+			// Best-effort close; the original error is more important.
+			if closeErr := sqlDB.Close(); closeErr != nil {
+				err = errors.Join(err, fmt.Errorf("closing database after setup failure: %w", closeErr))
```
ChrisJBurns, line 48: This cleanup silently discards the close error losing the `errr` assignment ?

pkg/storage/sqlite/migrations/001_create_entries_and_skills.sql
```diff
@@ -0,0 +1,55 @@
+-- +goose Up
+
+CREATE TABLE entries (
+    id          INTEGER PRIMARY KEY AUTOINCREMENT,
```
ChrisJBurns: Do we need `AUTOINCREMENT`? As it guarantees that IDs are never reused after deletion at the cost of maintaining an internal sequence table that prevents id reuse? For app internal db where IDs aren't exposed externally, plain `INTEGER PRIMARY KEY` should be sufficient and more performant

pkg/storage/sqlite/migrations/001_create_entries_and_skills.sql
```diff
@@ -0,0 +1,55 @@
...
+    scope        TEXT NOT NULL DEFAULT 'user'
+                      CHECK (scope IN ('user', 'project')),
+    project_root TEXT NOT NULL DEFAULT '',
+    reference    TEXT NOT NULL DEFAULT '',
+    tag          TEXT NOT NULL DEFAULT '',
+    digest       TEXT NOT NULL DEFAULT '',
+    version      TEXT NOT NULL DEFAULT '',
+    description  TEXT NOT NULL DEFAULT '',
+    author       TEXT NOT NULL DEFAULT '',
+    tags         BLOB DEFAULT NULL,
```
ChrisJBurns: Is ti worth adding a comment here to describe what will be stored here? 

```sql
tags         BLOB DEFAULT NULL, --JSON-encoded []string
client_apps  BLOB DEFAULT NULL,   -- JSON-encoded []string
```

pkg/storage/sqlite/db_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	db, err := Open(dbPath)
+	require.NoError(t, err)
+	defer db.Close()
+
+	tests := []struct {
+		pragma   string
+		expected string
+	}{
+		{"PRAGMA journal_mode", "wal"},
```
ChrisJBurns, line 59: nit: we're skipping `busy_timeout` and `cache_size`? is it worth adding them for completeness?

pkg/storage/sqlite/migrations/001_create_entries_and_skills.sql
```diff
@@ -0,0 +1,55 @@
+-- +goose Up
+
+CREATE TABLE entries (
+    id          INTEGER PRIMARY KEY AUTOINCREMENT,
+    entry_type  TEXT NOT NULL,
+    name        TEXT NOT NULL,
+    created_at  TEXT NOT NULL
+                    DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ', 'now')),
+    updated_at  TEXT NOT NULL
```
ChrisJBurns, line 11: `DEFAULT` only applies on `INSERT` which means SQLlite will never auto-update this value without a trigger. So on insert it will have the correct time but on update the `updated_at` will remain the same as it was on insert.

pkg/storage/sqlite/db.go
JAORMX: Good call — addressed. `Open` now accepts `context.Context` as the first parameter, which flows through to `runMigrations(ctx, db)` and `sqlDB.PingContext(ctx)`. No more hardcoded `context.Background()`.

pkg/storage/sqlite/db.go
JAORMX: Good catch — this was actually a real bug. `err` was not a named return, so the `errors.Join` mutation in the defer was silently lost. Fixed by changing the signature to `(_ *DB, err error)` so the named return propagates correctly. Also added a clarifying comment explaining the mechanism.

pkg/storage/sqlite/migrations/001_create_entries_and_skills.sql
JAORMX: Agreed — removed `AUTOINCREMENT` from both `entries` and `installed_skills`. IDs are purely internal FK references, never exposed to APIs. Plain `INTEGER PRIMARY KEY` is sufficient and avoids the `sqlite_sequence` overhead.

pkg/storage/sqlite/migrations/001_create_entries_and_skills.sql
JAORMX: Added `-- JSONB-encoded []string` comments to both BLOB columns.

pkg/storage/sqlite/db_test.go
JAORMX: Added — all 5 PRAGMAs now have test assertions: `journal_mode`, `busy_timeout`, `synchronous`, `foreign_keys`, and `cache_size`.

pkg/storage/sqlite/migrations/001_create_entries_and_skills.sql
JAORMX: Good observation. This is by design — the application layer explicitly sets `updated_at` in UPDATE queries. Added a SQL comment documenting this: `-- updated_at is set on INSERT via DEFAULT; the application layer is responsible for setting it explicitly in UPDATE statements.` We opted against a trigger to avoid hidden behavior and keep the contract explicit.

Issue #3676, opened 2026-02-06
Allow runtime image to be specified through UI/API


dmjb:
Following on from #3662 - change the API to allow the runtime image to be passed when creating a workload.

slyt3:
I will take this one, I will check the changes and update that APi accordingly, please assign it to me

jerm-dro:
Thanks @slyt3 for volunteering!

PR #3772, opened 2026-02-11
Allow runtime image in workload API

slyt3:
I added `runtime_config` to workload to create and update API requests for #3676; It now passes the runtime image override into the build/retriever flow and stores it in `RunConfig`, updated API docs and tests

### DIFF
--- a/pkg/telemetry/integration_test.go
+++ b/pkg/telemetry/integration_test.go
@@ -14,10 +14,13 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/metric/metricdata"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+	"go.opentelemetry.io/otel/trace"
 	tracenoop "go.opentelemetry.io/otel/trace/noop"
 
 	"github.com/stacklok/toolhive/pkg/mcp"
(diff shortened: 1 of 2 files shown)


slyt3:
I fixed the failing verify check separately: docs/server/swagger.json had an EOF newline mismatch; this is docs formatting only, not part of the #3676 runtime_config logic.

pkg/runner/protocol.go
```diff
@@ -157,20 +157,44 @@ func loadRuntimeConfig(
...
+			AdditionalPackages: append([]string{}, userConfig.AdditionalPackages...),
+		}
 	}
 
 	// Fall back to defaults
 	defaultConfig := templates.GetDefaultRuntimeConfig(transportType)
-	return &defaultConfig
+	return &templates.RuntimeConfig{
+		BuilderImage:       defaultConfig.BuilderImage,
+		AdditionalPackages: append([]string{}, defaultConfig.AdditionalPackages...),
```
JAORMX, line 196: This changes the merge semantic from full-replacement to append-with-base for **all callers** (CLI included), not just the API path. Before this PR, passing an override replaced everything; now `AdditionalPackages` are appended to base defaults and `BuilderImage` falls back to base if empty.

This is a breaking behavioral change for CLI users of `--runtime-add-package` -- they previously got exactly what they specified, now they get base packages plus their additions with no way to subtract.

A couple of options:
1. Keep full-replacement semantics here and let the API layer handle merging if that's desired.
2. If merge is the intended behavior for all paths, this should be a separate, clearly documented PR since it changes existing CLI behavior.

Also: there's double-trimming of `BuilderImage` -- `runtimeConfigFromRequest` already trims whitespace at the API boundary, and then `loadRuntimeConfig` trims again at line 173. Pick one place to trim (prefer the API boundary) and trust internal code receives clean input.

pkg/api/v1/workload_service.go
```diff
@@ -272,6 +275,11 @@ func (s *WorkloadService) BuildFullRunConfig(
 		runner.WithTelemetryConfigFromFlags("", false, false, false, "", 0.0, nil, false, nil, false),
 	}
 
+	// Runtime overrides only apply to protocol-scheme image builds.
+	if runtimeConfigOverride != nil && req.URL == "" {
+		options = append(options, runner.WithRuntimeConfig(runtimeConfigOverride))
+	}
+
```
JAORMX, line 282: The comment says "Runtime overrides only apply to protocol-scheme image builds" but if a user passes `runtime_config` with a regular Docker image (not `go://`, `npx://`, `uvx://`), no warning or error is returned -- it's silently ignored. This will confuse users who expect it to have an effect.

Consider returning a validation error (or at least a warning in the response) when `runtime_config` is provided alongside a non-protocol-scheme image.

pkg/runner/protocol_test.go
```diff
@@ -442,3 +443,59 @@ func TestCreateTemplateData(t *testing.T) {
...
+		BuilderImage: customImage,
+	})
+	if got == nil {
+		t.Fatal("loadRuntimeConfig returned nil config")
+		return
+	}
+	if got.BuilderImage != customImage {
+		t.Fatalf("BuilderImage = %q, want %q", got.BuilderImage, customImage)
+	}
+}
```
JAORMX, line 501: These tests use `t.Fatal`/`t.Fatalf` and `reflect.DeepEqual` instead of testify assertions (`assert.Equal`, `require.NotNil`, etc.). The other test files in this same PR (`workload_service_test.go`, `workloads_types_test.go`) correctly use testify, so let's keep things consistent here.

For example:
```go
// Instead of:
if got.BuilderImage != base.BuilderImage {
    t.Fatalf("BuilderImage = %q, want base %q", got.BuilderImage, base.BuilderImage)
}

// Use:
assert.Equal(t, base.BuilderImage, got.BuilderImage)
```

And `reflect.DeepEqual` can be replaced with `assert.Equal`.

pkg/api/v1/workloads_test.go
```diff
@@ -137,6 +139,28 @@ func TestCreateWorkload(t *testing.T) {
...
+			expectedRuntimeConfig: &templates.RuntimeConfig{
+				BuilderImage:       "golang:1.24-alpine",
+				AdditionalPackages: []string{"ca-certificates"},
+			},
+			expectedServerOrImage: "go://github.com/example/server",
+			expectedStatus:        http.StatusCreated,
+			expectedBody:          "test-workload",
+		},
 		{
 			name:        "with tool filters",
```
JAORMX, line 165: Good test for the happy path! A few edge cases are missing that would strengthen coverage:

- **Empty `runtime_config` (`{}`)**: Verify it doesn't override defaults (i.e., `runtimeConfigFromRequest` returns nil).
- **`runtime_config` with a non-protocol-scheme image**: e.g., `"image": "nginx:latest"` + `runtime_config` set. Verify the behavior is well-defined (currently silently ignored).
- **Update path round-trip**: Update a workload without specifying `runtime_config` and verify whether a previously stored config is preserved or cleared. This matters for UX since users may not expect an update to drop their runtime overrides.

slyt3:
@JAORMX thankyouuuu for taking time to look into this, gonna fix those issues!

JAORMX:
Hey @slyt3, thanks for working on this! The runtime image override for the workload API is a useful feature.

I'm going to close this for now since it's been a bit since the last update. The review feedback around the `loadRuntimeConfig` merge semantic and input validation still needs addressing, and I don't want this to go stale sitting in the queue.

If you want to pick this back up, feel free to reopen or open a fresh PR. The approach itself is solid... just needs those fixes. Happy hacking!

Issue #3399, opened 2026-01-22
Align telemetry with OpenTelemetry MCP semantic conventions


JAORMX:
## Summary

OpenTelemetry officially merged MCP semantic conventions on January 12, 2026 ([PR #2083](https://github.com/open-telemetry/semantic-conventions/pull/2083)). ToolHive should align its telemetry implementation with these standards for better observability tool compatibility and ecosystem alignment.

## Standard References

- **Main Documentation**: [docs/gen-ai/mcp.md](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/mcp.md)
- **Attribute Registry**: [model/mcp/registry.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/model/mcp/registry.yaml)
- **Metrics Definitions**: [model/mcp/metrics.yaml](https://github.com/open-telemetry/semantic-conventions/blob/main/model/mcp/metrics.yaml)
- **W3C Trace Context**: [https://www.w3.org/TR/trace-context/](https://www.w3.org/TR/trace-context/)

## Current State

ToolHive has solid telemetry foundation but predates the official conventions:
- **Middleware**: `pkg/telemetry/middleware.go` - spans, attributes, metrics for MCP proxy
- **vMCP**: `pkg/vmcp/server/telemetry.go` - backend and workflow telemetry  
- **Parser**: `pkg/mcp/parser.go` - already extracts `_meta` field (lines 228-233)

### 1. Update Attributes and Span Naming

**File**: `pkg/telemetry/middleware.go`

**Attribute Renames** (for standard compliance):
- `mcp.method` → `mcp.method.name` (line 222)
- `mcp.request.id` → `jsonrpc.request.id` (line 229)
- `mcp.tool.name` → `gen_ai.tool.name` (line 263)
- `mcp.tool.arguments` → `gen_ai.tool.call.arguments` (line 267, opt-in)
- `mcp.prompt.name` → `gen_ai.prompt.name` (line 279)
- `mcp.transport` → `network.transport` with value mapping:
  - `stdio` → `pipe`
  - `sse`, `streamable-http` → `tcp`

**Add Missing Required Attributes**:
- `mcp.protocol.version` - MCP spec version (e.g., "2025-11-25")
- `mcp.session.id` - Session identifier
- `jsonrpc.protocol.version` - When not "2.0"
- `error.type` - On failures (JSON-RPC error code or "tool_error")
- `rpc.response.status_code` - When response contains error
[...]

PR #3775, opened 2026-02-11
Align vMCP client telemetry with OTEL MCP semantic conventions

ChrisJBurns:
## Summary

- Aligns vMCP backend client spans and metrics with the OTEL MCP semantic conventions
- Span names changed from internal format (`telemetryBackendClient.call_tool`) to spec format (`tools/call my-tool`)
- Adds `mcp.method.name` (Required) to all client spans
- Adds spec-required attributes to `mcp.client.operation.duration` metric: `mcp.method.name`, `network.transport`, and `error.type` on error
- Adds `gen_ai.tool.name`, `gen_ai.prompt.name`, `mcp.resource.uri` to appropriate method spans
- Keeps all existing ToolHive-specific attributes for backward compatibility
- Avoids unbounded cardinality in ReadResource span names by omitting the URI from the span name (captured in attributes instead)
- Adds `resource_uri` backward-compat attribute to ReadResource, matching the pattern used by CallTool (`tool_name`) and GetPrompt (`prompt_name`)
- Adds unit tests for `mapActionToMCPMethod` and `mapTransportTypeToNetworkTransport` mapping functions

Parent issue: #3399

## Test plan

- [x] `go build ./pkg/vmcp/server/...` compiles cleanly
- [x] `go test ./pkg/vmcp/server/...` passes (including new unit tests)
- [x] `task lint-fix` passes with 0 issues

### DIFF
--- a/pkg/vmcp/server/telemetry.go
+++ b/pkg/vmcp/server/telemetry.go
@@ -14,6 +14,7 @@ import (
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/stacklok/toolhive/pkg/telemetry"
+	transporttypes "github.com/stacklok/toolhive/pkg/transport/types"
 	"github.com/stacklok/toolhive/pkg/vmcp"
 	"github.com/stacklok/toolhive/pkg/vmcp/server/adapter"
 )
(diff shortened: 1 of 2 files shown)


jerm-dro (approved):

pkg/vmcp/server/telemetry.go
```diff
@@ -94,40 +94,97 @@ type telemetryBackendClient struct {
...
+		return action
+	}
+}
+
+// mapTransportTypeToNetworkTransport maps MCP transport types to OTEL network.transport values.
+func mapTransportTypeToNetworkTransport(transportType string) string {
+	switch transportType {
+	case "stdio":
+		return "pipe"
+	case "sse", "streamable-http":
```
JAORMX: we have some already defined constants for these under the `transport` package IIRC. should we use those instead?

pkg/vmcp/server/telemetry.go
```diff
@@ -94,40 +94,97 @@ type telemetryBackendClient struct {
...
 
+// mapActionToMCPMethod maps internal action names to MCP method names per the OTEL MCP spec.
+func mapActionToMCPMethod(action string) string {
+	switch action {
+	case "call_tool":
+		return "tools/call"
+	case "read_resource":
+		return "resources/read"
+	case "get_prompt":
+		return "prompts/get"
```
JAORMX, line 106: nit: At some point we should probably create constants for these. No need to do it here though.

ChrisJBurns: Yeah I agree, I even asked the [go-sdk](https://github.com/modelcontextprotocol/go-sdk/issues/565) team to do this and their answer was essentially no.

pkg/vmcp/server/telemetry.go
ChrisJBurns: Cool, shall reuse, I've put something similar in the telemetry package too, I'll make sure to reuse them too.

JAORMX (approved):

PR #3786, opened 2026-02-11
Add SQLite FTS5-backed ToolStore for optimizer search

aponcedeleonch:
Closes: https://github.com/stacklok/toolhive/issues/3731

Add an SQLite FTS5 ToolStore implementation as an alternative to the existing InMemoryToolStore, and wire it into the vMCP server when optimizer config is present.

This adds:
- pkg/vmcp/optimizer/fts5store: SQLiteToolStore implementation using modernc.org/sqlite (pure Go, no CGO) with FTS5 virtual tables
- BM25 ranking for search results with LIKE-based fallback
- FTS5 query sanitization for safe handling of special characters
- Thread-safe concurrent access with sync.RWMutex
- Close() method on the ToolStore interface for resource cleanup
- OptimizerStoreCloser on server.Config for store lifecycle management
- OptimizerConfig.FTSDBPath for configurable database location (defaults to in-memory, use emptyDir in Kubernetes)
- Wire FTS5 store into vMCP serve command when optimizer is configured
- Comprehensive tests for search, upsert, and concurrency scenarios

Part 2 of the optimizer FTS5 migration (issue #3731).

## Large PR Justification

- Multiple related changes that would break if separated. It just introduces 1 new concept

### DIFF
--- a/pkg/vmcp/optimizer/internal/sqlite_store/schema.sql
+++ b/pkg/vmcp/optimizer/internal/sqlite_store/schema.sql
@@ -7,7 +7,10 @@ CREATE TABLE IF NOT EXISTS llm_capabilities (
     description TEXT NOT NULL DEFAULT ''
 );
 
--- FTS5 virtual table for full-text search with BM25 ranking
+-- FTS5 virtual table for full-text search with BM25 ranking.
+-- tokenize='porter' uses the Porter stemming algorithm so that morphological
+-- variants of a word (e.g. "running", "runs", "ran") match the root form "run".
+-- This improves recall for natural-language tool descriptions.
 CREATE VIRTUAL TABLE IF NOT EXISTS llm_capabilities_fts USING fts5(
     name,
     description,
--- a/pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
+++ b/pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
@@ -13,7 +13,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
-	"sync/atomic"
 
 	"github.com/mark3labs/mcp-go/server"
 	_ "modernc.org/sqlite" // registers the "sqlite" database/sql driver
@@ -21,37 +20,42 @@ import (
 	"github.com/stacklok/toolhive/pkg/vmcp/optimizer/internal/types"
 )
 
-// dbCounter ensures each in-memory SQLite database gets a unique name,
-// so multiple stores (e.g., in tests) don't share the same database.
-var dbCounter atomic.Int64
-
 //go:embed schema.sql
 var schemaSQL string
 
-// SQLiteToolStore implements a tool store using SQLite with FTS5 for full-text search.
-type SQLiteToolStore struct {
+// sqliteToolStore implements a tool store using SQLite with FTS5 for full-text search.
+// It satisfies the optimizer.ToolStore interface.
+type sqliteToolStore struct {
 	db *sql.DB
 }
 
-// NewSQLiteToolStore creates a new SQLiteToolStore backed by an in-memory SQLite database.
-func NewSQLiteToolStore() (*SQLiteToolStore, error) {
-	id := dbCounter.Add(1)
-	db, err := sql.Open("sqlite", fmt.Sprintf("file:memdb_%d?mode=memory&cache=shared", id))
+// NewSQLiteToolStore creates a new sqliteToolStore backed by a shared in-memory
+// SQLite database. All callers of this constructor share the same database,
+// which is the intended production behavior (one shared store per server).
+func NewSQLiteToolStore() (sqliteToolStore, error) {
+	return newSQLiteToolStore("file:memdb?mode=memory&cache=shared")
+}
+
+// newSQLiteToolStore creates a tool store backed by a database described
+// in the connectionString. It is useful for tests, where we want multiple
+// isolated (non-shared) databases.
+func newSQLiteToolStore(connectionString string) (sqliteToolStore, error) {
+	db, err := sql.Open("sqlite", connectionString)
 	if err != nil {
-		return nil, fmt.Errorf("failed to open sqlite database: %w", err)
+		return sqliteToolStore{}, fmt.Errorf("failed to open sqlite database: %w", err)
 	}
 
 	// Execute schema
 	if _, err := db.Exec(schemaSQL); err != nil {
 		_ = db.Close()
-		return nil, fmt.Errorf("failed to initialize schema: %w", err)
+		return sqliteToolStore{}, fmt.Errorf("failed to initialize schema: %w", err)
 	}
 
-	return &SQLiteToolStore{db: db}, nil
+	return sqliteToolStore{db: db}, nil
 }
 
 // UpsertTools adds or updates tools in the store.
-func (s *SQLiteToolStore) UpsertTools(ctx context.Context, tools []server.ServerTool) (retErr error) {
+func (s sqliteToolStore) UpsertTools(ctx context.Context, tools []server.ServerTool) (retErr error) {
 	tx, err := s.db.BeginTx(ctx, nil)
 	if err != nil {
 		return fmt.Errorf("failed to begin transaction: %w", err)
@@ -81,28 +85,33 @@ func (s *SQLiteToolStore) UpsertTools(ctx context.Context, tools []server.Server
 // The allowedTools parameter limits results to only tools with names in the given set.
 // If allowedTools is empty, no results are returned (empty = no access).
 // Returns matches ranked by relevance.
-func (s *SQLiteToolStore) Search(ctx context.Context, query string, allowedTools []string) ([]types.ToolMatch, error) {
+func (s sqliteToolStore) Search(ctx context.Context, query string, allowedTools []string) ([]types.ToolMatch, error) {
 	if len(allowedTools) == 0 {
 		return nil, nil
 	}
 
-	sanitized := sanitizeFTS5Query(query)
-	if sanitized == "" {
+	ftsExpr := sanitizeFTS5Query(query)
+	if ftsExpr == "" {
 		return nil, nil
 	}
 
-	return s.searchFTS5(ctx, sanitized, allowedTools)
+	return s.searchFTS5(ctx, ftsExpr, allowedTools)
 }
 
 // Close releases the underlying database connection.
-func (s *SQLiteToolStore) Close() error {
+func (s sqliteToolStore) Close() error {
 	return s.db.Close()
 }
 
 // searchFTS5 performs a full-text search using FTS5 MATCH with BM25 ranking.
 // It uses json_each() to pass the allowed tool names as a single JSON array
 // parameter, avoiding manual placeholder construction.
-func (s *SQLiteToolStore) searchFTS5(ctx context.Context, ftsQuery string, allowedTools []string) ([]types.ToolMatch, error) {
+//
+// The ftsExpr is produced by sanitizeFTS5Query and is always passed as a
+// parameterized ? value, never interpolated into SQL.
+func (s sqliteToolStore) searchFTS5(
+	ctx context.Context, ftsExpr string, allowedTools []string,
+) ([]types.ToolMatch, error) {
 	allowedJSON, err := json.Marshal(allowedTools)
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal allowed tools: %w", err)
@@ -115,7 +124,7 @@ func (s *SQLiteToolStore) searchFTS5(ctx context.Context, ftsQuery string, allow
 		  AND t.name IN (SELECT value FROM json_each(?))
 		ORDER BY rank`
 
-	rows, err := s.db.QueryContext(ctx, queryStr, ftsQuery, string(allowedJSON))
+	rows, err := s.db.QueryContext(ctx, queryStr, ftsExpr, string(allowedJSON))
 	if err != nil {
 		return nil, fmt.Errorf("FTS5 query failed: %w", err)
 	}
@@ -141,20 +150,30 @@ func (s *SQLiteToolStore) searchFTS5(ctx context.Context, ftsQuery string, allow
 // problematicWords contains words that FTS5 interprets as operators or that
 // are too common in tool metadata to be useful search terms. This set aligns
 // with Python mcp_optimizer's DEFAULT_FTS_PROBLEMATIC_WORDS.
-var problematicWords = map[string]bool{
-	"name": true, "description": true, "schema": true, "input": true,
-	"output": true, "type": true, "properties": true, "required": true,
-	"title": true, "id": true, "tool": true, "server": true,
-	"meta": true, "data": true, "content": true, "text": true,
-	"value": true, "field": true, "column": true, "table": true,
-	"index": true, "key": true, "primary": true,
+var problematicWords = map[string]struct{}{
+	"name": {}, "description": {}, "schema": {}, "input": {},
+	"output": {}, "type": {}, "properties": {}, "required": {},
+	"title": {}, "id": {}, "tool": {}, "server": {},
+	"meta": {}, "data": {}, "content": {}, "text": {},
+	"value": {}, "field": {}, "column": {}, "table": {},
+	"index": {}, "key": {}, "primary": {},
 }
 
-// sanitizeFTS5Query prepares a query string for FTS5 MATCH.
-// It quotes each word to prevent FTS5 column-name interpretation,
-// uses OR joining for multi-word queries, and handles words that
-// are problematic for FTS5 search. This aligns with the Python
-// mcp_optimizer's _sanitize_fts_query.
+// sanitizeFTS5Query prepares a user query string for use with FTS5 MATCH.
+//
+// The returned string is designed to be passed as a single ? parameter to
+// QueryContext. It cannot cause SQL injection because it is always bound via ?.
+//
+// FTS5 MATCH requires a single string operand containing the full query
+// expression (e.g., "read" OR "write"). Individual terms cannot be separate
+// ? SQL parameters because the OR/AND operators are part of the FTS5 query
+// language, not SQL.
+// See: https://sqlite.org/fts5.html#full_text_query_syntax
+//
+// Safety:
+//   - SQL injection is prevented because the expression is always bound via ?.
+//   - FTS5 operator injection is prevented by double-quoting each term and
+//     escaping embedded double-quotes (standard FTS5 escaping).
 func sanitizeFTS5Query(query string) string {
 	words := strings.Fields(strings.TrimSpace(query))
 	if len(words) == 0 {
@@ -163,7 +182,7 @@ func sanitizeFTS5Query(query string) string {
 
 	hasProblematic := false
 	for _, word := range words {
-		if problematicWords[strings.ToLower(word)] {
+		if _, ok := problematicWords[strings.ToLower(word)]; ok {
 			hasProblematic = true
 			break
 		}
--- a/pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store_test.go
+++ b/pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"testing"
 
 	"github.com/mark3labs/mcp-go/mcp"
@@ -14,9 +15,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func newTestStore(t *testing.T) *SQLiteToolStore {
+// testDBCounter ensures each test gets a unique in-memory database.
+var testDBCounter atomic.Int64
+
+func newTestStore(t *testing.T) sqliteToolStore {
 	t.Helper()
-	store, err := NewSQLiteToolStore()
+	id := testDBCounter.Add(1)
+	store, err := newSQLiteToolStore(fmt.Sprintf("file:testdb_%d?mode=memory&cache=shared", id))
 	require.NoError(t, err)
 	t.Cleanup(func() {
 		_ = store.Close()
@@ -28,14 +33,6 @@ func makeTools(tools ...mcp.Tool) []server.ServerTool {
 	result := make([]server.ServerTool, len(tools))
 	for i, tool := range tools {
 		result[i] = server.ServerTool{Tool: tool}
-	}
-	return result
-}
-
-func toolNames(tools []server.ServerTool) []string {
-	result := make([]string, len(tools))
-	for i, tool := range tools {
-		result[i] = tool.Tool.Name
 	}
 	return result
 }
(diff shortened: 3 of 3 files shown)


cmd/vmcp/app/commands.go
```diff
@@ -483,8 +484,14 @@ func runServe(cmd *cobra.Command, _ []string) error {
 	}
 
 	if cfg.Optimizer != nil {
-		// TODO: update this with the real optimizer.
-		serverCfg.OptimizerFactory = optimizer.NewDummyOptimizerFactory()
+		dbPath := cfg.Optimizer.FTSDBPath
```
jerm-dro: Blocker: We shouldn't need to store a path in the config, since the database is in memory?

cmd/vmcp/app/commands.go
```diff
@@ -483,8 +484,14 @@ func runServe(cmd *cobra.Command, _ []string) error {
...
+		store, err := fts5store.NewSQLiteToolStore(dbPath)
+		if err != nil {
+			return fmt.Errorf("failed to create optimizer store: %w", err)
+		}
+
+		serverCfg.OptimizerFactory = optimizer.NewDummyOptimizerFactoryWithStore(store)
+		serverCfg.OptimizerStoreCloser = store
```
jerm-dro: Suggestion: this code is creating and starting resources and producing shutdown functions that need to be called. Move this to server.Start.

Then, you don't need a special field to track the closer. You can append it to the shutdownFuncs array: https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/server/server.go#L224

pkg/vmcp/config/config.go
```diff
@@ -746,6 +746,12 @@ type OptimizerConfig struct {
 	// for semantic tool discovery. The service must implement the optimizer embedding API.
 	// +kubebuilder:validation:Required
 	EmbeddingService string `json:"embeddingService" yaml:"embeddingService"`
+
+	// FTSDBPath is the file path for the SQLite FTS5 database used for tool search.
+	// If empty, a temporary file is used. In Kubernetes, this should point to an
```
jerm-dro: Blocker: we don't want to use a temporary file, we want to use an in-memory db. Specifically, this connection string `file::memory:?cache=shared` according to the docs: https://www.sqlite.org/inmemorydb.html

pkg/vmcp/optimizer/fts5store/schema.sql
```diff
@@ -0,0 +1,31 @@
+-- SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Tools table stores tool metadata
+CREATE TABLE IF NOT EXISTS tools (
```
jerm-dro: Suggestion: let's not call this tools. What if we end up adding resources, prompts, or skills? We shouldn't have to create a brand new table. I'll leave the naming up to you, because I can't think of anything generic that I like.

pkg/vmcp/optimizer/fts5store/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+    name TEXT PRIMARY KEY,
+    description TEXT NOT NULL DEFAULT '',
+    input_schema TEXT NOT NULL DEFAULT '{}'
```
jerm-dro: Question, non-blocking: Is the schema used in the search at all? If not, I'd suggest deleting it. The callers of the optimizer will have remembered the input_schema.

pkg/vmcp/optimizer/fts5store/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+);
+
+-- FTS5 virtual table for full-text search with BM25 ranking
+CREATE VIRTUAL TABLE IF NOT EXISTS tools_fts USING fts5(
```
jerm-dro: nit: again, can we not call this tools?

pkg/vmcp/optimizer/fts5store/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+    name,
+    description,
+    content=tools,
+    content_rowid=rowid
+);
+
+-- Triggers to keep FTS index in sync with tools table
+CREATE TRIGGER IF NOT EXISTS tools_ai AFTER INSERT ON tools BEGIN
```
jerm-dro: nit: `tools_ai` would be clearer if it were `tools_after_insert`.

pkg/vmcp/optimizer/fts5store/sqlite_store.go
```diff
@@ -0,0 +1,250 @@
...
+	db, err := sql.Open("sqlite", dbPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open sqlite database: %w", err)
+	}
+
+	// Enable WAL mode for better concurrent read performance
+	if _, err := db.Exec("PRAGMA journal_mode=WAL"); err != nil {
+		_ = db.Close()
+		return nil, fmt.Errorf("failed to enable WAL mode: %w", err)
+	}
```
jerm-dro: Blocker: delete this. Let's use the assumption that sqlite will always be strictly in-memory to our advantage to keep the code simple.

pkg/vmcp/optimizer/fts5store/sqlite_store.go
```diff
@@ -0,0 +1,250 @@
...
+func (s *SQLiteToolStore) UpsertTools(ctx context.Context, tools []server.ServerTool) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	tx, err := s.db.BeginTx(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	defer func() {
+		// Rollback is a no-op if the transaction was already committed.
```
jerm-dro: Question, blocking: Is this true? I couldn't find any comments or source code stating this.

Rather than digging, you could just write the code so the tx is only rolled back on errors:
```go
func () UpsertTools(ctx context.Context, tools []server.ServerTool) (retErr error) {
	tx, err := s.db.BeginTx(ctx, nil)
	if err != nil {
		return fmt.Errorf("failed to begin transaction: %w", err)
	}
    defer func () {
    if retErr != nil {
      tx.Rollback()
    }
    }()
    // ... do the stuff
}
```

pkg/vmcp/optimizer/fts5store/sqlite_store.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
+)
+
+//go:embed schema.sql
+var schemaSQL string
+
+// SQLiteToolStore implements optimizer.ToolStore using SQLite with FTS5 for full-text search.
+type SQLiteToolStore struct {
+	db     *sql.DB
+	mu     sync.RWMutex
```
jerm-dro: sqlite implements the synchronization for us, so this is not necessary to protect the db. 

If this is just for the `closed` bool, I say delete the bool.

pkg/vmcp/optimizer/fts5store/sqlite_store.go
```diff
@@ -0,0 +1,250 @@
...
+		JOIN tools t ON t.rowid = fts.rowid
+		WHERE tools_fts MATCH ?`
+
+	if len(allowedTools) > 0 {
+		placeholders := make([]string, len(allowedTools))
+		for i, name := range allowedTools {
+			placeholders[i] = "?"
+			args = append(args, name)
+		}
+		queryStr += " AND t.name IN (" + strings.Join(placeholders, ", ") + ")"
```
jerm-dro: Suggestion: you can provide an array argument as an arg to `QueryContext`. That would allow you to write code like `... AND t.name in ?`. It should make this function much simpler.

pkg/vmcp/optimizer/fts5store/sqlite_store.go
```diff
@@ -0,0 +1,250 @@
...
+	queryStr := `SELECT name, description, input_schema FROM tools
+		WHERE (name LIKE ? OR description LIKE ?)`
+
+	if len(allowedTools) > 0 {
+		placeholders := make([]string, len(allowedTools))
+		for i, name := range allowedTools {
+			placeholders[i] = "?"
+			args = append(args, name)
+		}
+		queryStr += " AND name IN (" + strings.Join(placeholders, ", ") + ")"
```
jerm-dro: Similar to comment above re: using arrays in args.

pkg/vmcp/optimizer/dummy_optimizer_test.go
```diff
@@ -5,13 +5,135 @@ package optimizer
...
+	}
+	return nil
+}
+
+func (m *mockToolStore) Search(ctx context.Context, query string, allowedTools []string) ([]ToolMatch, error) {
+	if m.searchFunc != nil {
+		return m.searchFunc(ctx, query, allowedTools)
+	}
+	return nil, nil
+}
```
jerm-dro, line 35: Suggestion: these methods should panic if called but not configured. For example, `Search` will return empty results which may be misinterpreted as a passing test.

pkg/vmcp/optimizer/fts5store/sqlite_store.go
```diff
@@ -0,0 +1,250 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package fts5store implements a SQLite FTS5-based ToolStore for full-text
+// search over MCP tool metadata.
+package fts5store
```
jerm-dro: can you put this package into an internal/fts5store directory?

It should never be used outside the optimizer, and internal will enforce that.

pkg/vmcp/optimizer/fts5store/sqlite_store_test.go
```diff
@@ -0,0 +1,470 @@
...
+
+	// Empty allowedTools should return all matching tools
+	results, err := store.Search(ctx, "tool", nil)
+	if err != nil {
+		t.Fatalf("failed to search: %v", err)
+	}
+	if len(results) != 2 {
+		t.Fatalf("expected 2 results with empty allowedTools, got %d", len(results))
+	}
+}
```
jerm-dro: The tests in this file could be made more dense by making them tabular:
``` go
type testCase struct {
   rows []Tools
    query string
    expectedResults []ToolMatch
}
```

pkg/vmcp/optimizer/fts5store/sqlite_store_test.go
```diff
@@ -0,0 +1,470 @@
...
+func TestSQLiteToolStore_Search_NoResults(t *testing.T) {
+	t.Parallel()
+	store := newTestStore(t)
+	ctx := context.Background()
+
+	tools := makeTools(
+		mcp.NewTool("read_file", mcp.WithDescription("Read a file")),
+	)
+	if err := store.UpsertTools(ctx, tools); err != nil {
+		t.Fatalf("failed to upsert: %v", err)
```
jerm-dro: Suggestion, non-blocking: IDK if it's just me, but I'm accustomed write: `require.NoError(err)` instead of using `t.Fatal`. This allows you to avoid ifs / nesting, which makes tests easy to read.

pkg/vmcp/optimizer/store.go
```diff
@@ -25,10 +25,15 @@ type ToolStore interface {
 	UpsertTools(ctx context.Context, tools []server.ServerTool) error
 
 	// Search finds tools matching the query string.
-	// The scope parameter limits results to only tools with names in the given set.
-	// If scope is empty, all tools are searched.
+	// The allowedTools parameter limits results to only tools with names in the given set.
+	// If allowedTools is empty, all tools are searched.
```
jerm-dro: Blocker: On second read, I think this behavior "If allowedTools is empty, all tools are searched." is dangerous and should be changed. If the caller has no allowed tools, no results should be returned.

pkg/vmcp/server/server_test.go
```diff
@@ -288,6 +288,72 @@ func TestServer_Stop(t *testing.T) {
...
+	closed bool
+	err    error
+}
+
+func (s *stubCloser) Close() error {
+	s.closed = true
+	return s.err
+}
+
+func TestServer_Stop_ClosesOptimizerStore(t *testing.T) {
```
jerm-dro: I made a comment on `commands.go` around reusing the existing `shutdownFuncs` pattern. It'd be great to implement that comment, but keep this test to verify shutdown functions are called correctly.

jerm-dro (changes requested):
I'll take a closer look at the sqlite_store_test when they're in a more condensed form. Everything else looks mostly good aside for some blocking comments which should be easy to address.

cmd/vmcp/app/commands.go
aponcedeleonch: you're completely right, deleted it

cmd/vmcp/app/commands.go
aponcedeleonch: went with the suggestion, I wasn't aware of the shutdown functions. Now the store is also closed there

pkg/vmcp/optimizer/fts5store/schema.sql
aponcedeleonch: I went with `llm_capabilities`, I guess is as generic as it gets

pkg/vmcp/optimizer/fts5store/schema.sql
aponcedeleonch: you're right, in my mind this was needed because we used it in the python implementation but here is not needed at all. Removed it

pkg/vmcp/config/config.go
aponcedeleonch: yes! sorry for that, now is completely in memory

pkg/vmcp/optimizer/fts5store/schema.sql
aponcedeleonch: renamed

pkg/vmcp/optimizer/fts5store/schema.sql
aponcedeleonch: yes, you're right. renamed

pkg/vmcp/optimizer/fts5store/sqlite_store.go
aponcedeleonch: good catch, deleted it :)

pkg/vmcp/optimizer/fts5store/sqlite_store.go
aponcedeleonch: right!

pkg/vmcp/optimizer/fts5store/sqlite_store.go
aponcedeleonch: apparently not in SQLite natively, however LLM help me found a function called `json_each`. The query is now simplified

pkg/vmcp/optimizer/fts5store/sqlite_store.go
aponcedeleonch: changed now

pkg/vmcp/optimizer/fts5store/sqlite_store.go
aponcedeleonch: done!

pkg/vmcp/optimizer/fts5store/sqlite_store_test.go
aponcedeleonch: replaced :)

pkg/vmcp/server/server_test.go
aponcedeleonch: yes, went with your suggestion. Thanks!

pkg/vmcp/optimizer/fts5store/sqlite_store_test.go
aponcedeleonch: done! now they're more tabular

pkg/vmcp/optimizer/store.go
aponcedeleonch: yes, you're right. Also `allowedTools` should never be empty so it makes sense

pkg/vmcp/optimizer/dummy_optimizer_test.go
aponcedeleonch: fixed

pkg/vmcp/optimizer/fts5store/sqlite_store.go
aponcedeleonch: did it as you suggested. What I found is a comment in an example code in the documentation, although not strictly pointed out as part of any function. [Link](https://pkg.go.dev/database/sql#example-Tx.Prepare)

Example code:
```go
func main() {
	projects := []struct {
		mascot  string
		release int
	}{
		{"tux", 1991},
		{"duke", 1996},
		{"gopher", 2009},
		{"moby dock", 2013},
	}

	tx, err := db.Begin()
	if err != nil {
		log.Fatal(err)
	}
	defer tx.Rollback() // The rollback will be ignored if the tx has been committed later in the function.

	stmt, err := tx.Prepare("INSERT INTO projects(id, mascot, release, category) VALUES( ?, ?, ?, ? )")
	if err != nil {
		log.Fatal(err)
	}
	defer stmt.Close() // Prepared statements take up server resources and should be closed after use.

	for id, project := range projects {
		if _, err := stmt.Exec(id+1, project.mascot, project.release, "open source"); err != nil {
			log.Fatal(err)
		}
	}
	if err := tx.Commit(); err != nil {
		log.Fatal(err)
	}
}
```

pkg/vmcp/optimizer/internal/sqlite_store/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+    description TEXT NOT NULL DEFAULT ''
+);
+
+-- FTS5 virtual table for full-text search with BM25 ranking
+CREATE VIRTUAL TABLE IF NOT EXISTS llm_capabilities_fts USING fts5(
+    name,
+    description,
+    content=llm_capabilities,
+    content_rowid=rowid,
+    tokenize='porter'
```
jerm-dro, line 19: nit: can you add a comment to document what this is for and why it was selected?

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/mark3labs/mcp-go/server"
+	_ "modernc.org/sqlite" // registers the "sqlite" database/sql driver
+
+	"github.com/stacklok/toolhive/pkg/vmcp/optimizer/internal/types"
+)
+
+// dbCounter ensures each in-memory SQLite database gets a unique name,
+// so multiple stores (e.g., in tests) don't share the same database.
+var dbCounter atomic.Int64
```
jerm-dro: Suggestion: I think this is more complex than it needs to be.

I would recommend splitting up the constructor:


```go 

// NewSQLiteToolStore returns a SQLiteToolStore which uses a shared in-memory database.
func NewSQLiteToolStore() {
   return newSQLiteToolStore("file:memdb?mode=memory&cache=shared")
} 


// newSQLiteToolStore creates a tool store backed by a database described
// in the connectionString. It is useful for tests, where we want multiple not shared databases.
func newSQLiteToolStore(connectionString) {
 // the existing constructor code
}
```

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
```diff
@@ -0,0 +1,191 @@
...
+
+//go:embed schema.sql
+var schemaSQL string
+
+// SQLiteToolStore implements a tool store using SQLite with FTS5 for full-text search.
+type SQLiteToolStore struct {
```
jerm-dro: nit: SQLiteToolStore implements an interface, right?

If so, I'd suggest making `SQLiteToolStore` private and have `NewSQLiteToolStore` return the interface type (e.g. `ToolStore`) instead. The callers of `NewSQLiteToolStore` shouldn't need any methods or data from the implementation type. Everything they need should be on the interface.

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
```diff
@@ -0,0 +1,191 @@
...
+	if _, err := db.Exec(schemaSQL); err != nil {
+		_ = db.Close()
+		return nil, fmt.Errorf("failed to initialize schema: %w", err)
+	}
+
+	return &SQLiteToolStore{db: db}, nil
+}
+
+// UpsertTools adds or updates tools in the store.
+func (s *SQLiteToolStore) UpsertTools(ctx context.Context, tools []server.ServerTool) (retErr error) {
```
jerm-dro: nit: I don't think we need a pointer receiver, since the contained `*sql.DB` is a pointer. Nothing else is mutated on the struct. `func (s SQLiteToolStore)` should work just fine.

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	defer func() {
+		if retErr != nil {
+			_ = tx.Rollback()
+		}
+	}()
+
+	stmt, err := tx.PrepareContext(ctx, "INSERT OR REPLACE INTO llm_capabilities (name, description) VALUES (?, ?)")
```
jerm-dro, line 69: Praise: I didn't know about this but I like how tidy it makes the code 😄

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
```diff
@@ -0,0 +1,191 @@
...
+		})
+	}
+
+	return matches, rows.Err()
+}
+
+// problematicWords contains words that FTS5 interprets as operators or that
+// are too common in tool metadata to be useful search terms. This set aligns
+// with Python mcp_optimizer's DEFAULT_FTS_PROBLEMATIC_WORDS.
+var problematicWords = map[string]bool{
```
jerm-dro: Suggestion: in golang it is typical to represent a set as `map[string]struct{}`. 

Then, you check existence by: `_, ok := set[key]`

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
```diff
@@ -0,0 +1,191 @@
...
+		return `"` + escaped + `"`
+	}
+
+	// Multi-word with no problematic words: join with OR
+	quoted := make([]string, len(words))
+	for i, word := range words {
+		escaped := strings.ReplaceAll(word, `"`, `""`)
+		quoted[i] = `"` + escaped + `"`
+	}
+	return strings.Join(quoted, " OR ")
```
jerm-dro, line 203: Blocker: Mixing the query construction with parameters creates an opening for SQL injection, it's not a big deal for this use case. However, it could be easily fixed. One suggestion: return a `[]string` for all the query parameters to use. Then, it's the callers responsibility to construct the sql, but these args are inserted only via `?`.

jerm-dro (commented):
One minor blocker, but otherwise LGTM

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
aponcedeleonch: nice, yes changed!

pkg/vmcp/optimizer/internal/sqlite_store/schema.sql
aponcedeleonch: sure, comment added

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
aponcedeleonch: you're right! changed

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
aponcedeleonch: yes, you're right, changed!

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
aponcedeleonch: thanks :)

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
aponcedeleonch: nice, that's good to know. Thanks for this, changed

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
aponcedeleonch: the `OR` is part of the FTS query. We cannot have several strings in the `MATCH` parameter, in other words
```sql
MATCH ? OR ? OR ? ..
```sql
everything has to be a single string
```
MATCH ?
```

Hence we construct the single FTS query with the `OR` statements. Added some documentation about this

File: pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go [hunk 26]
```diff
@@ -13,45 +13,49 @@ import (
...
 }
 
-// NewSQLiteToolStore creates a new SQLiteToolStore backed by an in-memory SQLite database.
-func NewSQLiteToolStore() (*SQLiteToolStore, error) {
-	id := dbCounter.Add(1)
-	db, err := sql.Open("sqlite", fmt.Sprintf("file:memdb_%d?mode=memory&cache=shared", id))
+// NewSQLiteToolStore creates a new sqliteToolStore backed by a shared in-memory
+// SQLite database. All callers of this constructor share the same database,
+// which is the intended production behavior (one shared store per server).
+func NewSQLiteToolStore() (sqliteToolStore, error) {
```
jerm-dro, line 35: nit: can you return the interface rather than the private type?

aponcedeleonch: Wiil do that! 🙂

Issue #3650, opened 2026-02-06
Skill service implementation


JAORMX:
## Description

Extend the basic `SkillService` (from #3742) with full OCI operations, skill installation, parsing, validation, and upgrade support. This is the central orchestration layer that ties together all skill components.

**Note:** #3742 delivers the initial `service_impl.go` with basic storage CRUD. This task extends that same file with OCI pull/push, installer integration, unmanaged skill detection, and upgrade logic.

**Dependencies**: #3742 (basic service + server wiring), #3649 (installer), #3646 (parser/validator), toolhive-core#15 (registry client), toolhive-core#16 (packager)
**Blocks**: #3651 (full REST API), #3654 (groups integration)

## Acceptance Criteria

- [ ] `pkg/skills/service_impl.go` extended with full operations (builds on #3742's basic CRUD)
- [ ] `List()`: merges SQLite results with filesystem scan for unmanaged skills
- [ ] `Install()`: pulls from registry (via toolhive-core), extracts to client directory, persists to SQLite via `SkillStore.Create()` (extends the basic metadata-only install from #3742)
- [ ] `Uninstall()`: removes managed skill files, deletes from SQLite, refuses to remove unmanaged
- [ ] `Info()`: returns detailed skill information from SQLite metadata + SKILL.md frontmatter
- [ ] `Validate()`: runs parser + validator on skill directory, returns structured results
- [ ] `Build()`: uses toolhive-core packager to create OCI artifact, stores locally (replaces "not implemented" from #3742)
- [ ] `Push()`: uses toolhive-core registry client to push artifact (replaces "not implemented" from #3742)
- [ ] Upgrade support via `-u` flag: resolves original reference, pulls latest
- [ ] Unit tests with mocked `SkillStore` (via mockgen), SPDX headers

## Code Pointers

- `pkg/skills/service_impl.go` — basic service created by #3742 (to be extended)
- `pkg/skills/service.go` — `SkillService` interface (from #3645)
- `pkg/storage/interfaces.go` — `SkillStore` interface (from #3648)

## References

- [RFC-0041: SQLite-Based State Management](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0041-sqlite-state-management.md)
- #3645, #3742

JAORMX:
**Update: State management approach changed to SQLite**

[...]

PR #3804, opened 2026-02-12
Add basic SkillService implementation and API handlers

### LINKED ISSUE #3741: SkillService implementation and integration
## Description

Implement the `SkillService` interface using the SQLite-backed `SkillStore`, wire it into the HTTP server and CLI commands, and replace the 501 API stubs with working handlers.

This story is blocked by #3648 (skill state management / storage layer).

**RFC**: https://github.com/stacklok/toolhive-rfcs/pull/41
**Dependencies**: #3648

## Scope

Three PRs covering the full integration stack:

1. **SkillService implementation + server wiring** — Implement `SkillService` in `pkg/skills/` using `storage.SkillStore`, wire into `ServerBuilder.createDefaultManagers()`, add `sqlite.Close()` to shutdown path
2. **Skills CLI commands** — Flesh out `cmd/thv/app/skill_*.go` stubs, remove `Hidden: true`, add output formatting (text + JSON)
3. **Skills API handlers** — Replace 501 stubs in `pkg/api/v1/skills.go` with real implementations

### Service Implementation (PR 1)
- [ ] `pkg/skills/service_impl.go` implements `SkillService` with `SkillStore` dependency injection
- [ ] Constructor: `NewService(store storage.SkillStore) SkillService`
- [ ] `ServerBuilder.createDefaultManagers()` creates SQLite store → service on startup
- [ ] `sqlite.Close()` called on server shutdown path
- [ ] Unit tests with `MockSkillStore`

### CLI Commands (PR 2)
- [ ] `thv skill list` — lists installed skills with scope filter
- [ ] `thv skill install <name>` — installs a skill
- [ ] `thv skill uninstall <name>` — removes an installed skill
- [ ] `thv skill info <name>` — shows skill details
- [ ] `Hidden: true` removed from skill parent command
- [ ] `--format json` and `--format text` support
- [ ] CLI commands are thin wrappers delegating to `pkg/skills`

### API Handlers (PR 3)
- [ ] `GET /api/v1beta/skills` — list skills
- [ ] `POST /api/v1beta/skills/install` — install skill
- [ ] `POST /api/v1beta/skills/uninstall` — uninstall skill
- [ ] `GET /api/v1beta/skills/{name}` — get skill info
- [ ] Request validation and error handling
- [ ] API tests with mock service

## Code Pointers

- `pkg/skills/service.go` — existing `SkillService` interface
- `pkg/api/v1/skills.go` — existing 501 stubs and route definitions
- `cmd/thv/app/skill*.go` — existing CLI command stubs
- `pkg/api/server.go` — `ServerBuilder`, `createDefaultManagers()`, shutdown path
- `pkg/groups/` — reference for CLI→service→store wiring pattern

JAORMX:
## Summary

This PR delivers the service layer and API wiring needed to make the skills
endpoints functional, building on the SQLite SkillStore merged in #3801.

**Why a sub-package?** `pkg/skills` and `pkg/storage` have a circular import
(storage uses skills types, skills needs the store interface). The service
implementation lives in `pkg/skills/skillsvc/` to break this cycle cleanly.

### What each endpoint does now

| Endpoint | Before | After |
|----------|--------|-------|
| `GET /api/v1beta/skills` | 501 | Lists installed skills, optional `?scope=` filter |
| `POST /api/v1beta/skills/install` | 501 | Creates a "pending" skill record (no OCI pull yet) |
| `POST /api/v1beta/skills/uninstall` | 501 | Deletes a skill by name+scope |
| `GET /api/v1beta/skills/{name}` | 501 | Returns skill info or `{installed: false}` |
| `POST /api/v1beta/skills/validate` | 501 | Still 501 (deferred) |
| `POST /api/v1beta/skills/build` | 501 | Still 501 (deferred) |
| `POST /api/v1beta/skills/push` | 501 | Still 501 (deferred) |

### Server lifecycle changes

`createDefaultManagers()` now opens a `SkillStore` via `sqlite.NewDefaultSkillStore()`
and passes it to `skillsvc.New()`. The store reference is tracked as an `io.Closer`
on the `Server` struct and closed in `cleanup()` — before unix socket removal, even
on crash paths. When callers inject their own service via `WithSkillManager()`, they
own the store lifecycle (documented in the method comment).

### Service design decisions

- **Install creates "pending" only** — no OCI pull in this phase, that's #3650
- **Info returns 200 always** — `{installed: false}` for missing skills, not 404
- **Info hardcodes ScopeUser** — documented in code comment; project-scoped lookup
  comes when `InfoOptions` gains a `Scope` field
- **All name-accepting methods validate via `ValidateSkillName`** before touching the store
- **Scope defaults to `ScopeUser`** via extracted `defaultScope()` helper

### Handler patterns

Handlers follow the same pattern as `groups.go`: decode JSON → validate required
fields → call service → encode response. Error mapping uses `httperr.WithCode()`
for 4xx and lets `ErrorHandler` sanitize 5xx to generic messages.

Closes #3741

## Large PR Justification

The PR is ~980 lines total but only ~200 lines are production code (across 3 modified
files and 1 new file). The remaining ~780 lines are test files:
- `pkg/skills/skillsvc/skillsvc_test.go` (367 lines) — service unit tests
- `pkg/api/v1/skills_test.go` (380 lines) — handler tests
- `docs/server/` — auto-generated swagger docs

The production code cannot be split further — the service implementation, API handlers,
and server wiring are tightly coupled (handlers call the service, server creates the
service). Shipping them separately would leave broken 501 stubs between PRs.

## Test plan

- [x] 24 service unit tests — table-driven with MockSkillStore covering happy paths,
  validation errors, error propagation, scope defaulting, and version passthrough
- [x] 24 handler tests — table-driven with MockSkillService and httptest round-trips
  covering JSON decoding, empty/malformed input, status code mapping (400/404/409/501),
  scope+version passthrough, and response format validation
- [x] `task lint-fix` clean, `task build` compiles, race detector passes

### DIFF
--- a/pkg/api/v1/skills.go
+++ b/pkg/api/v1/skills.go
@@ -56,14 +56,11 @@ func (s *SkillsRoutes) listSkills(w http.ResponseWriter, r *http.Request) error 
 
 	result, err := s.skillService.List(r.Context(), skills.ListOptions{Scope: scope})
 	if err != nil {
-		return fmt.Errorf("failed to list skills: %w", err)
+		return err
 	}
 
 	w.Header().Set("Content-Type", "application/json")
-	if err := json.NewEncoder(w).Encode(skillListResponse{Skills: result}); err != nil {
-		return fmt.Errorf("failed to encode skills list: %w", err)
-	}
-	return nil
+	return json.NewEncoder(w).Encode(skillListResponse{Skills: result})
 }
 
 // installSkill installs a skill from a remote source.
@@ -89,13 +86,6 @@ func (s *SkillsRoutes) installSkill(w http.ResponseWriter, r *http.Request) erro
 		)
 	}
 
-	if req.Name == "" {
-		return httperr.WithCode(
-			fmt.Errorf("name is required"),
-			http.StatusBadRequest,
-		)
-	}
-
 	result, err := s.skillService.Install(r.Context(), skills.InstallOptions{
 		Name:    req.Name,
 		Version: req.Version,
@@ -108,10 +98,7 @@ func (s *SkillsRoutes) installSkill(w http.ResponseWriter, r *http.Request) erro
 	w.Header().Set("Content-Type", "application/json")
 	w.Header().Set("Location", fmt.Sprintf("/api/v1beta/skills/%s", result.Skill.Metadata.Name))
 	w.WriteHeader(http.StatusCreated)
-	if err := json.NewEncoder(w).Encode(installSkillResponse{Skill: result.Skill}); err != nil {
-		return fmt.Errorf("failed to encode install response: %w", err)
-	}
-	return nil
+	return json.NewEncoder(w).Encode(installSkillResponse{Skill: result.Skill})
 }
 
 // uninstallSkill removes an installed skill.
@@ -176,14 +163,11 @@ func (s *SkillsRoutes) getSkillInfo(w http.ResponseWriter, r *http.Request) erro
 
 	info, err := s.skillService.Info(r.Context(), skills.InfoOptions{Name: name, Scope: scope})
 	if err != nil {
-		return fmt.Errorf("failed to get skill info: %w", err)
+		return err
 	}
 
 	w.Header().Set("Content-Type", "application/json")
-	if err := json.NewEncoder(w).Encode(info); err != nil {
-		return fmt.Errorf("failed to encode skill info: %w", err)
-	}
-	return nil
+	return json.NewEncoder(w).Encode(info)
 }
 
 // validateSkill checks whether a skill definition is valid.
--- a/pkg/skills/skillsvc/skillsvc.go
+++ b/pkg/skills/skillsvc/skillsvc.go
@@ -78,7 +78,6 @@ func (s *service) Info(ctx context.Context, opts skills.InfoOptions) (*skills.Sk
 
 	return &skills.SkillInfo{
 		Metadata:       skill.Metadata,
-		Installed:      true,
 		InstalledSkill: &skill,
 	}, nil
 }
--- a/docs/server/docs.go
+++ b/docs/server/docs.go
@@ -1553,7 +1553,7 @@ const docTemplate = `{
                 ]
             },
             "skills.InstalledSkill": {
-                "description": "InstalledSkill is set if the skill is installed.",
+                "description": "InstalledSkill contains the full installation record.",
                 "properties": {
                     "clients": {
                         "description": "Clients is the list of client identifiers the skill is installed for.\nTODO: Refactor client.ClientApp to a shared package so it can be used here instead of []string.",
@@ -1617,10 +1617,6 @@ const docTemplate = `{
             },
             "skills.SkillInfo": {
                 "properties": {
-                    "installed": {
-                        "description": "Installed indicates whether the skill is currently installed.",
-                        "type": "boolean"
-                    },
                     "installed_skill": {
                         "$ref": "#/components/schemas/skills.InstalledSkill"
                     },
--- a/docs/server/swagger.json
+++ b/docs/server/swagger.json
@@ -1546,7 +1546,7 @@
                 ]
             },
             "skills.InstalledSkill": {
-                "description": "InstalledSkill is set if the skill is installed.",
+                "description": "InstalledSkill contains the full installation record.",
                 "properties": {
                     "clients": {
                         "description": "Clients is the list of client identifiers the skill is installed for.\nTODO: Refactor client.ClientApp to a shared package so it can be used here instead of []string.",
@@ -1610,10 +1610,6 @@
             },
             "skills.SkillInfo": {
                 "properties": {
-                    "installed": {
-                        "description": "Installed indicates whether the skill is currently installed.",
-                        "type": "boolean"
-                    },
                     "installed_skill": {
                         "$ref": "#/components/schemas/skills.InstalledSkill"
                     },
--- a/docs/server/swagger.yaml
+++ b/docs/server/swagger.yaml
@@ -1427,7 +1427,7 @@ components:
       - InstallStatusPending
       - InstallStatusFailed
     skills.InstalledSkill:
-      description: InstalledSkill is set if the skill is installed.
+      description: InstalledSkill contains the full installation record.
       properties:
         clients:
           description: |-
@@ -1477,9 +1477,6 @@ components:
       - ScopeProject
     skills.SkillInfo:
       properties:
-        installed:
-          description: Installed indicates whether the skill is currently installed.
-          type: boolean
         installed_skill:
           $ref: '#/components/schemas/skills.InstalledSkill'
         metadata:
--- a/pkg/api/v1/skills_test.go
+++ b/pkg/api/v1/skills_test.go
@@ -110,22 +110,28 @@ func TestSkillsRouter(t *testing.T) {
 			expectedBody:   `"my-skill"`,
 		},
 		{
-			name:           "install skill empty name",
-			method:         "POST",
-			path:           "/",
-			body:           `{"name":""}`,
-			setupMock:      func(_ *skillsmocks.MockSkillService) {},
+			name:   "install skill empty name",
+			method: "POST",
+			path:   "/",
+			body:   `{"name":""}`,
+			setupMock: func(svc *skillsmocks.MockSkillService) {
+				svc.EXPECT().Install(gomock.Any(), skills.InstallOptions{Name: ""}).
+					Return(nil, httperr.WithCode(fmt.Errorf("invalid skill name: must not be empty"), http.StatusBadRequest))
+			},
 			expectedStatus: http.StatusBadRequest,
-			expectedBody:   "name is required",
+			expectedBody:   "invalid skill name",
 		},
 		{
-			name:           "install skill missing name field",
-			method:         "POST",
-			path:           "/",
-			body:           `{}`,
-			setupMock:      func(_ *skillsmocks.MockSkillService) {},
+			name:   "install skill missing name field",
+			method: "POST",
+			path:   "/",
+			body:   `{}`,
+			setupMock: func(svc *skillsmocks.MockSkillService) {
+				svc.EXPECT().Install(gomock.Any(), skills.InstallOptions{Name: ""}).
+					Return(nil, httperr.WithCode(fmt.Errorf("invalid skill name: must not be empty"), http.StatusBadRequest))
+			},
 			expectedStatus: http.StatusBadRequest,
-			expectedBody:   "name is required",
+			expectedBody:   "invalid skill name",
 		},
 		{
 			name:           "install skill malformed json",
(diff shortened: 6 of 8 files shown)


pkg/skills/skillsvc/skillsvc.go
```diff
@@ -0,0 +1,113 @@
...
+// supported and will be added when InfoOptions gains a Scope field.
+func (s *service) Info(ctx context.Context, opts skills.InfoOptions) (*skills.SkillInfo, error) {
+	if err := skills.ValidateSkillName(opts.Name); err != nil {
+		return nil, httperr.WithCode(err, http.StatusBadRequest)
+	}
+
+	skill, err := s.store.Get(ctx, opts.Name, skills.ScopeUser, "")
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			return &skills.SkillInfo{Installed: false}, nil
```
ChrisJBurns: Is this returning `installed: false` in the case of a not found? Shouldn't we respond with `404` instead?

This is inconsistent with:
  - REST semantics (GET a resource that doesn't exist = 404)
  - The uninstall endpoint in this same PR, which correctly returns 404 for missing skills
  - ToolHive's own workloads/groups APIs, which return 404 for missing resources

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Info returns detailed information about a skill.
+// Info always queries user-scoped skills; project-scoped lookup is not yet
+// supported and will be added when InfoOptions gains a Scope field.
+func (s *service) Info(ctx context.Context, opts skills.InfoOptions) (*skills.SkillInfo, error) {
+	if err := skills.ValidateSkillName(opts.Name); err != nil {
+		return nil, httperr.WithCode(err, http.StatusBadRequest)
+	}
+
+	skill, err := s.store.Get(ctx, opts.Name, skills.ScopeUser, "")
```
ChrisJBurns: We hardcode to `ScopeUser`? but on other endpoints we respect the scope from the request (with a default fallback)? A skill installed with `scope: "project"` would show up in `GET /skills?scope=project` but not in `GET /skills/{name}`? Is that right?

ChrisJBurns: Additionally, we don't seem to do scope validation? So someone could send `?scope=foobar` and its silently accepted because the `defaultScope()` function just checks if the `scope` string is empty?

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if err := s.store.Create(ctx, skill); err != nil {
+		return nil, err
+	}
+
+	return &skills.InstallResult{Skill: skill}, nil
+}
+
+// Uninstall removes an installed skill.
+func (s *service) Uninstall(ctx context.Context, opts skills.UninstallOptions) error {
```
ChrisJBurns, line 60: I think `Uninstall` should use `DELETE /skills/{name}` or something similar. Otherwise we break HTTP method semantics (`POST` is not idempotent, but uninstall should be) and API caches can't distinguish between install and uninstall (both are POST to different sub-paths)

pkg/api/v1/skills.go
```diff
@@ -59,10 +74,41 @@ func (*SkillsRoutes) listSkills(_ http.ResponseWriter, _ *http.Request) error {
...
+		Name:    req.Name,
+		Version: req.Version,
+		Scope:   req.Scope,
+	})
+	if err != nil {
+		return err
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusCreated)
```
ChrisJBurns, line 100: nit: `201` is good, but typically you'd also respond with a `Location` header too to let clients know where to find the skill thats been created without having to parse the response body.

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return s.store.Delete(ctx, opts.Name, defaultScope(opts.Scope), "")
+}
+
+// Info returns detailed information about a skill.
+// Info always queries user-scoped skills; project-scoped lookup is not yet
+// supported and will be added when InfoOptions gains a Scope field.
+func (s *service) Info(ctx context.Context, opts skills.InfoOptions) (*skills.SkillInfo, error) {
+	if err := skills.ValidateSkillName(opts.Name); err != nil {
+		return nil, httperr.WithCode(err, http.StatusBadRequest)
```
ChrisJBurns, line 71: nit: we're leaking http concerns into svc layer. In other parts of toolhive we define typed errors and the middleware/handler maps them to http codes. Also the handler that calls this function seems to already do its own validation for empty names, so this line maybe redundant?

pkg/api/v1/skills.go
```diff
@@ -73,10 +119,35 @@ func (*SkillsRoutes) installSkill(_ http.ResponseWriter, _ *http.Request) error
...
-//	@Failure		501		{string}	string	"Not Implemented"
+//	@Failure		400		{string}	string	"Bad Request"
+//	@Failure		404		{string}	string	"Not Found"
+//	@Failure		500		{string}	string	"Internal Server Error"
 //	@Router			/api/v1beta/skills/uninstall [post]
-func (*SkillsRoutes) uninstallSkill(_ http.ResponseWriter, _ *http.Request) error {
-	return httperr.WithCode(fmt.Errorf("not implemented"), http.StatusNotImplemented)
+func (s *SkillsRoutes) uninstallSkill(w http.ResponseWriter, r *http.Request) error {
+	var req uninstallSkillRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
```
ChrisJBurns: nit: there's no check for empty body. so if there's an empty body we give a `400 invalid request body`. this is ok, but slightly misleading, a better check for empty body with an error of `400 request body is required` is more helpful

pkg/skills/skillsvc/skillsvc.go
JAORMX: Fixed. Info() now returns storage.ErrNotFound directly, which the error handler middleware maps to 404. Updated the tests to verify this behavior.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Fixed. Info() now uses defaultScope(opts.Scope) instead of hardcoding ScopeUser. Added tests to verify it respects both project scope and defaults to user scope when empty.

pkg/skills/skillsvc/skillsvc.go
JAORMX: This is actually the established ToolHive pattern. We define typed errors with HTTP codes at the package level, then service methods return them:

- pkg/groups/errors.go: ErrGroupNotFound = httperr.WithCode(..., http.StatusNotFound)
- pkg/storage/errors.go: ErrNotFound = httperr.WithCode(..., http.StatusNotFound)  
- pkg/container/runtime/types.go: ErrWorkloadNotFound = httperr.WithCode(..., http.StatusNotFound)

The handlers just return these errors, and the ErrorHandler middleware extracts the status code. It keeps the layering clean since the errors are defined once per package, not created ad-hoc in every service method.

The validation here does use httperr.WithCode inline, which I agree is less clean. We could extract that to a package-level ErrInvalidSkillName if you think that would be better.

pkg/api/v1/skills.go
```diff
@@ -63,10 +78,40 @@ func (*SkillsRoutes) listSkills(_ http.ResponseWriter, _ *http.Request) error {
...
+	var req installSkillRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return httperr.WithCode(
+			fmt.Errorf("invalid request body: %w", err),
+			http.StatusBadRequest,
+		)
+	}
+
+	if req.Name == "" {
+		return httperr.WithCode(
```
aponcedeleonch: nit: duplicate validation — this `req.Name == ""` check is redundant since `Install()` in the service layer already calls `ValidateSkillName(opts.Name)`, which rejects empty names. The other handlers (uninstall, info) don't do this double-check; they validate in the handler via `ValidateSkillName` and let the service re-validate.

Pick one layer. Given that the service already validates, you could drop this block and let the service's `ValidateSkillName` handle it (its error message is slightly different — "invalid skill name: must not be empty" vs "name is required" — but that's arguably more consistent with the rest of the API).

pkg/api/v1/skills.go
```diff
@@ -45,10 +46,24 @@ func SkillsRouter(skillService skills.SkillService) http.Handler {
...
 //	@Router			/api/v1beta/skills [get]
-func (*SkillsRoutes) listSkills(_ http.ResponseWriter, _ *http.Request) error {
-	return httperr.WithCode(fmt.Errorf("not implemented"), http.StatusNotImplemented)
+func (s *SkillsRoutes) listSkills(w http.ResponseWriter, r *http.Request) error {
+	scope := skills.Scope(r.URL.Query().Get("scope"))
+	if err := skills.ValidateScope(scope); err != nil {
+		return httperr.WithCode(err, http.StatusBadRequest)
+	}
+
+	result, err := s.skillService.List(r.Context(), skills.ListOptions{Scope: scope})
```
aponcedeleonch, line 57: nit: inconsistent error wrapping — `listSkills` and `getSkillInfo` wrap service errors with `fmt.Errorf("failed to X: %w", err)`, while `installSkill` and `uninstallSkill` return `err` bare. Since the `ErrorHandler` middleware sanitizes 5xx messages to generic "Internal Server Error" anyway, the wrapping doesn't add value for the client. Either wrap consistently across all four handlers, or (simpler) return bare errors everywhere and let the middleware handle it.

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	skill, err := s.store.Get(ctx, opts.Name, defaultScope(opts.Scope), "")
+	if err != nil {
+		return nil, err
+	}
+
```
aponcedeleonch, line 78: nit: now that `Info()` returns `storage.ErrNotFound` (→ 404) for missing skills, the `Installed: true` field in the response is always `true` — you'd never get a 200 with `Installed: false` since that case is now a 404. The `Installed` field on `SkillInfo` is effectively vestigial.

Not necessarily something to fix in this PR, but worth a follow-up to either remove the field or document why it's kept (e.g., if a future code path could return a 200 with partial info for uninstalled but registry-known skills).

pkg/api/v1/skills.go
JAORMX: Good catch\! Dropped the handler-level `req.Name == ""` check entirely. The service's `ValidateSkillName` handles it now, which also makes the error message consistent with how uninstall and info do it (`"invalid skill name: must not be empty"`).

pkg/api/v1/skills.go
JAORMX: Yeah, that was sloppy on my part. Went with bare errors everywhere now — all four handlers just `return err` from service calls, and the encode paths are simplified to `return json.NewEncoder(w).Encode(...)`. The middleware handles the rest.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Went ahead and removed it in this same commit since it was a one-liner. If we ever need a 200 with partial info for registry-known-but-not-installed skills, we can reintroduce it then.

Issue #3732, opened 2026-02-09
Add vector embedding support and hybrid similarity search to the optimizer ToolStore


jerm-dro:
**Depends on:** #3731

### Context

Issue #3731 introduces a `ToolStore` interface backed by SQLite + FTS5 for basic text similarity search. This issue adds vector embedding support to enable semantic search — finding tools by meaning rather than keyword overlap. A hybrid approach (FTS5 + vector similarity) improves recall for natural-language queries.

In this iteration, we use a **fake/stub embedding client** to keep the change small and focused on the storage and search mechanics. A real embedding provider is wired in a subsequent issue.

Prototype reference: https://github.com/stacklok/toolhive/pull/3516 (see `pkg/vmcp/optimizer/internal/embeddings/` and the hybrid search implementation in `internal/db/`)

### Requirements

1. Extend the `ToolStore` interface (from #3731) to support storing and querying vector embeddings alongside tool metadata.
2. Introduce an **embedding client interface** that the store uses to generate embeddings for tool descriptions at ingestion time and for queries at search time. This iteration uses a fake/stub implementation that returns deterministic vectors.
3. Implement **hybrid search**: combine FTS5 BM25 scores with vector cosine similarity to produce a blended ranking. The weighting between the two should be configurable.
4. Include **Go benchmark tests** (`go test -bench`) demonstrating embedding search latency over ~1000 tools with a production-grade number of embedding dimensions (e.g., 768 or 1536). These benchmarks establish a baseline for performance regression detection.

### High-Level Implementation

- Define an embedding client interface (e.g., `EmbeddingClient`) in `pkg/vmcp/optimizer/` with a method to generate embeddings from text
- Create a fake/stub implementation that returns deterministic fixed-dimension vectors (useful for testing and benchmarking)
- Extend the SQLite `ToolStore` to store embedding vectors alongside tool records
- Implement vector cosine similarity search (consider using an in-process vector library or pure-Go implementation)
- Add a hybrid search method that blends FTS5 rank and cosine similarity with a configurable ratio
- Update the `dummyOptimizer` to use hybrid search when an `EmbeddingClient` is available, falling back to FTS5-only when it is not

### Acceptance Tests

- [ ] **Embedding client interface tests**: Fake client returns vectors of the expected dimensionality and is deterministic
- [ ] **Vector storage**: Tools ingested with embeddings can be retrieved; embeddings are persisted correctly in SQLite
[...]

PR #3808, opened 2026-02-12
Add vector embedding support and hybrid search to ToolStore

aponcedeleonch:
Closes: https://github.com/stacklok/toolhive/issues/3732

Extend the SQLite-backed ToolStore with optional embedding-based semantic search. When an EmbeddingClient is provided, Search runs FTS5 and cosine similarity in parallel via errgroup, merges and deduplicates results by keeping the lower distance score, and caps output at 4 results.

Key additions:
- EmbeddingClient interface in internal/types (Embed, EmbedBatch, Dimension, Close)
- FakeEmbeddingClient using SHA-256 seeded RNG with L2-normalized vectors
- Pure Go cosine similarity/distance in internal/similarity package
- Binary encode/decode helpers for embedding BLOB storage
- normalizeBM25 rescaled to [0, 2) range to align with cosine distance
- Comprehensive unit tests, concurrency tests, and benchmarks

## Large PR Justification

- Multiple related changes that would break if separated. This PR introduces semantic search and a wait to combine results from full text search and semantic


pkg/vmcp/optimizer/internal/similarity/cosine_bench_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func BenchmarkCosineDistance_768(b *testing.B) {
+	a := randomVector(768)
+	v := randomVector(768)
+	b.ResetTimer()
+	b.ReportAllocs()
+	for b.Loop() {
+		CosineDistance(a, v)
+	}
+}
```
jerm-dro, line 38: Can you add some comments noting the results at the time of this PR?

aponcedeleonch: Good idea! Added a file in the folder called BENCHMARKS.md

pkg/vmcp/optimizer/internal/similarity/cosine_bench_test.go
```diff
@@ -0,0 +1,38 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
yrobla, line 1: why do we need to have benchmarks as part of the code? isn't that more on the research side?

pkg/vmcp/optimizer/internal/sqlite_store/BENCHMARKS.md
```diff
@@ -0,0 +1,33 @@
+# SQLite ToolStore Benchmarks
```
yrobla: same, why do we need it here?

pkg/vmcp/optimizer/internal/similarity/cosine_bench_test.go
aponcedeleonch: true, they're not actually testing any functionality of the code. Since there's no benchmarking dedicated place for the moment we decided to keep them here. I added a todo to both files stating that benchmarking should be moved away from this repo

pkg/vmcp/optimizer/internal/sqlite_store/BENCHMARKS.md
aponcedeleonch: copy paste from the other comment:
> true, they're not actually testing any functionality of the code. Since there's no benchmarking dedicated place for the moment we decided to keep them here. I added a todo to both files stating that benchmarking should be moved away from this repo

pkg/vmcp/optimizer/internal/similarity/cosine_bench_test.go
jerm-dro: Let's keep the test per our discussion on slack, but we don't need to commit the file.

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
```diff
@@ -122,9 +207,10 @@ func (s sqliteToolStore) searchFTS5(
 		JOIN llm_capabilities t ON t.rowid = fts.rowid
 		WHERE llm_capabilities_fts MATCH ?
 		  AND t.name IN (SELECT value FROM json_each(?))
-		ORDER BY rank`
+		ORDER BY rank
+		LIMIT ?`
```
jerm-dro, line 217: Why limit? What happens if you have a hybrid search and a row's rank is low and below the limit, but the cosine similarity is high?

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
```diff
@@ -147,6 +233,100 @@ func (s sqliteToolStore) searchFTS5(
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal allowed tools: %w", err)
+	}
+
+	queryStr := `SELECT name, description, embedding
+		FROM llm_capabilities
+		WHERE embedding IS NOT NULL
+		  AND name IN (SELECT value FROM json_each(?))`
+
+	rows, err := s.db.QueryContext(ctx, queryStr, string(allowedJSON))
```
jerm-dro, line 272: Blocker, question: Do we really need to query the DB twice?

Why can't we:
1. Load the allowed tools, their rank, and embeddings.
2. Compute the cosine similarity for each returned row
3. Compute the merged score
4. Return the top K results based on score.

pkg/vmcp/optimizer/internal/types/types.go
```diff
@@ -12,6 +40,23 @@ type ToolMatch struct {
...
+// Implementations may use local models, remote APIs, or deterministic fakes.
+type EmbeddingClient interface {
+	// Embed returns a vector embedding for the given text.
+	Embed(ctx context.Context, text string) ([]float32, error)
+
+	// EmbedBatch returns vector embeddings for multiple texts.
+	EmbedBatch(ctx context.Context, texts []string) ([][]float32, error)
+
+	// Dimension returns the dimensionality of the embeddings produced.
+	Dimension() int
```
jerm-dro: suggestion: delete - you can just gather this from the returned embedding?

pkg/vmcp/optimizer/embedding.go
```diff
@@ -0,0 +1,11 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package optimizer
+
+import "github.com/stacklok/toolhive/pkg/vmcp/optimizer/internal/types"
+
+// EmbeddingClient generates vector embeddings from text.
+// It is defined in the internal/types package and aliased here so that
+// external consumers continue to use optimizer.EmbeddingClient.
+type EmbeddingClient = types.EmbeddingClient
```
jerm-dro: Question, blocking: why does this need to be re-exported?

pkg/vmcp/optimizer/fake_embedding.go
```diff
@@ -0,0 +1,75 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package optimizer
```
jerm-dro: suggestion: move this embedding related code to the `similarity` package it seems very related, and the actual `CosineSimilarity` function is an implementation detail.

pkg/vmcp/optimizer/embedding.go
jerm-dro: Suggestion: You may be able to avoid re-exporting things by making the entrypoint to this package, `NewSQLiteToolStore`, accept the config object. When that function is called, it can freely reference things in the internal packages to construct objects.

pkg/vmcp/optimizer/internal/similarity/cosine_bench_test.go
aponcedeleonch: Done. Deleted `BENCHMARKS.md`, kept the bench test files.

pkg/vmcp/optimizer/internal/sqlite_store/BENCHMARKS.md
aponcedeleonch: Deleted the file. Bench tests are kept with a TODO note about moving them to a dedicated benchmarking location.

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
aponcedeleonch: Good question. The LIMIT is per-method, not global. In hybrid mode, FTS5 and semantic search each independently return their top-k results (split by `hybridSemanticToolsRatio`). So if a tool has a low BM25 rank but high cosine similarity, the semantic query will still surface it independently. The two result sets are then merged and deduplicated in `mergeResults`, keeping the best score per tool.

Added a comment to `searchFTS5` explaining this.

pkg/vmcp/optimizer/internal/sqlite_store/sqlite_store.go
aponcedeleonch: Yes, we need two separate queries because BM25 rank and cosine similarity are fundamentally different metrics that can't be meaningfully combined in a single SQL query:

- **BM25 rank** is a hidden FTS5 column computed on-the-fly from term frequency during a `MATCH` query. It only exists in the context of an FTS5 search.
- **Cosine similarity** requires loading raw embedding blobs and computing vector distance in Go.

These scores aren't comparable (not apples-to-apples), so we don't try to merge them into a single composite score. Instead, each method returns its own top-k, and `mergeResults` deduplicates by tool name, keeping the lower (better) score when both methods find the same tool.

Added a comment to `searchSemantic` explaining this rationale.

pkg/vmcp/optimizer/internal/types/types.go
aponcedeleonch: Done. Removed `Dimension()` from the `EmbeddingClient` interface. Callers can infer dimensionality from `len(embedding)` when needed.

pkg/vmcp/optimizer/embedding.go
aponcedeleonch: Done. Followed the config object suggestion: `NewSQLiteToolStore` now accepts a `*SQLiteStoreConfig` instead of `EmbeddingClient`. The config has an `EmbeddingDimension` field — when set, the function internally constructs the embedding client from the `internal` packages. External consumers never see `EmbeddingClient`.

Deleted `embedding.go` (the re-export file).

pkg/vmcp/optimizer/fake_embedding.go
aponcedeleonch: Done. Moved `FakeEmbeddingClient` and its tests to `internal/similarity/fake_embedding.go`.

Issue #3360, opened 2026-01-20
[BUG] vmcp: Enable Admission Webhooks for Kubernetes Operator


jerm-dro:
# Summary

The ToolHive Kubernetes operator has admission webhook code for validating `VirtualMCPServer`, `VirtualMCPCompositeToolDefinition`, and `MCPExternalAuthConfig` resources, but these webhooks have never been functional. The controller-runtime v0.23.0 upgrade exposed this issue.

### What Happened

During the upgrade to controller-runtime v0.23.0, the operator began failing at startup with:

```
"error":"open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory"
```

### Root Cause Analysis

Investigation revealed that **the webhooks were never actually working**:

1. **In controller-runtime v0.22.x**: The old webhook API (`ctrl.NewWebhookManagedBy(mgr).For(r).Complete()`) silently failed to register webhooks. The webhook server never started because no webhooks were registered with it.

2. **In controller-runtime v0.23.0**: The new generic webhook API (`builder.WebhookManagedBy[T](mgr, r).WithValidator(r).Complete()`) properly registers webhooks, which triggers the webhook server to start, which then fails because TLS certificates are not available.

### Missing Infrastructure

Even if the webhook server started, the webhooks would not function because:

| Component | Required | Status |
|-----------|----------|--------|
| ValidatingWebhookConfiguration | ✓ | Not deployed by helm chart |
| Webhook Service | ✓ | Not deployed by helm chart |
| Port 9443 exposed | ✓ | Not in deployment spec |
| TLS certificates | ✓ | No cert-manager integration |

The `config/webhook/manifests.yaml` file exists (kubebuilder-generated) but is never deployed.

## Impact of Missing Webhooks

The webhooks perform validation-only (no mutation). Without them:

| Resource | Webhook Validation | Controller Validation | Risk |
|----------|-------------------|----------------------|------|
| VirtualMCPServer | Disabled | Partial (during reconcile) | Low - caught at reconcile |
| MCPExternalAuthConfig | Disabled | None | **High - invalid configs silently accepted** |
| VirtualMCPCompositeToolDefinition | Disabled | None | **High - invalid configs silently accepted** |

### Example Validations Not Enforced

**MCPExternalAuthConfig:**
- Can create `tokenExchange` type without required `tokenExchange` config
- Can set conflicting configs (both `tokenExchange` and `headerInjection`)
- Unsupported auth types are accepted

[...]

PR #3819, opened 2026-02-13
Replace non-functional webhooks with hybrid validation approach

yrobla:
Remove validation webhooks that were never deployed due to missing infrastructure (no cert-manager, webhook service, or TLS setup). Replace with a hybrid validation strategy using CEL for simple rules and controller-side validation for complex business logic.

Closes: #3360

## Large PR Justification

This is a code simplification that involves removing large pieces of code, so that's why the big diff.


cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
```diff
@@ -38,6 +40,15 @@ type ExternalAuthType string
...
+// +kubebuilder:validation:XValidation:rule="self.type == 'headerInjection' ? has(self.headerInjection) : !has(self.headerInjection)",message="headerInjection configuration must be set if and only if type is 'headerInjection'"
+// +kubebuilder:validation:XValidation:rule="self.type == 'bearerToken' ? has(self.bearerToken) : !has(self.bearerToken)",message="bearerToken configuration must be set if and only if type is 'bearerToken'"
+// +kubebuilder:validation:XValidation:rule="self.type == 'embeddedAuthServer' ? has(self.embeddedAuthServer) : !has(self.embeddedAuthServer)",message="embeddedAuthServer configuration must be set if and only if type is 'embeddedAuthServer'"
+// +kubebuilder:validation:XValidation:rule="self.type == 'awsSts' ? has(self.awsSts) : !has(self.awsSts)",message="awsSts configuration must be set if and only if type is 'awsSts'"
+// +kubebuilder:validation:XValidation:rule="self.type == 'unauthenticated' ? (!has(self.tokenExchange) && !has(self.headerInjection) && !has(self.bearerToken) && !has(self.embeddedAuthServer) && !has(self.awsSts)) : true",message="no configuration must be set when type is 'unauthenticated'"
```
jerm-dro, line 49: Questions, blocking: These are quite complex kubebuilder statements and I'm worried about maintenance:
- What happens if one of the fields referenced here are changed? Will something fail so we update this validation?
- Do we have tests that validate each of these? I'm assuming there's no code coverage, but it's important these are covered.

Issue #3820, opened 2026-02-13
Implement TokenMetrics calculation in Optimizer


aponcedeleonch:
Implement actual token counting logic for `TokenMetrics` in the Optimizer. Currently, the `TokenMetrics` struct exists in `pkg/vmcp/optimizer/optimizer.go` but is always returned with zero values. We need to calculate token usage metrics every time tools are searched, so clients can understand the token savings achieved by the optimizer.

## Background

The `TokenMetrics` struct is defined as:

```go
type TokenMetrics struct {
    BaselineTokens int     `json:"baseline_tokens"`
    ReturnedTokens int     `json:"returned_tokens"`
    SavingsPercent float64 `json:"savings_percent"`
}
```

When the optimizer is enabled, clients interact with tools through `find_tool` and `call_tool` instead of receiving all tools directly. The token metrics should quantify the benefit of this optimization by measuring how many tokens are saved by returning only relevant tools instead of the full tool set.

## Requirements

- **BaselineTokens**: Calculate the estimated token count if all tools in the Optimizer's store were sent to the client. This should be computed by summing the token count of every tool's definition (name, description, input schema).
- **ReturnedTokens**: Calculate the actual token count for only the tools returned by a search query (`FindTool`).
- **SavingsPercent**: Compute `((BaselineTokens - ReturnedTokens) / BaselineTokens) * 100` to express the percentage of tokens saved.

## Implementation Notes

- Token counting should happen in the `FindTool` flow, where tools are searched and returned to the client.
- The token count for a tool should be based on its serialized representation (name + description + JSON schema), since that is what gets sent to the LLM.
- Consider using a lightweight token estimation approach (e.g., character count / 4 as a rough approximation, or a proper tokenizer library) — discuss tradeoffs in the PR.
- The `DummyOptimizer` in `pkg/vmcp/optimizer/dummy_optimizer.go` currently returns `TokenMetrics{}`. This should be updated to return real values.
- The `FindToolOutput` already includes a `TokenMetrics` field, so the return path is already wired up.
- The `ToolStore` interface (`pkg/vmcp/optimizer/internal/types/types.go`) and its implementations (in-memory and SQLite) may need to expose the total tool count or total token baseline for metrics calculation.

## Acceptance Criteria

- [ ] `FindToolOutput.TokenMetrics` returns accurate, non-zero values when the optimizer is used
- [ ] `BaselineTokens` reflects the total token cost of all tools in the store
[...]

PR #3821, opened 2026-02-13
Add token metrics to DummyOptimizer find_tool responses

aponcedeleonch:
Closes: https://github.com/stacklok/toolhive/issues/3820

Add TokenCounter abstraction and wire it into DummyOptimizer so that find_tool responses include TokenMetrics (baseline, returned, savings%).

Token counts are precomputed per-tool at optimizer construction time and the baseline total is cached as an immutable field, avoiding redundant summation on every FindTool call. The computeTokenMetrics helper operates entirely on optimizer-local state—no store involvement needed.

The TokenCounter interface lives in the optimizer package (not the internal types package) since only the optimizer layer uses it. CharDivTokenCounter provides the default heuristic (JSON byte length / 4).

Factory functions (NewDummyOptimizerFactory, NewDummyOptimizerFactoryWithStore) accept or create a TokenCounter and pass it through to NewDummyOptimizer.

### DIFF
--- a/pkg/vmcp/optimizer/dummy_optimizer.go
+++ b/pkg/vmcp/optimizer/dummy_optimizer.go
@@ -28,6 +28,9 @@ type DummyOptimizer struct {
 	tools map[string]server.ServerTool
 
 	// tokenCounts holds precomputed per-tool token estimates, indexed by tool name.
+	// Immutable after construction: token counts are computed once in NewDummyOptimizer
+	// and never modified. The tools are fixed per session (one optimizer per session),
+	// and the TokenCounter is set at configuration time, so counts cannot change at runtime.
 	tokenCounts map[string]int
 
 	// baselineTokens is the precomputed sum of all per-tool token counts.
--- a/pkg/vmcp/optimizer/dummy_optimizer_test.go
+++ b/pkg/vmcp/optimizer/dummy_optimizer_test.go
@@ -38,111 +38,80 @@ func (*mockToolStore) Close() error {
 	return nil
 }
 
-// TestDummyOptimizer_MockStore tests the optimizer against a mock ToolStore,
-// verifying search delegation, scoping, and error handling without any database.
-func TestDummyOptimizer_MockStore(t *testing.T) {
+// TestDummyOptimizer_SearchDelegation verifies that FindTool delegates to the
+// store with the correct query and allowedTools, and computes token metrics.
+func TestDummyOptimizer_SearchDelegation(t *testing.T) {
 	t.Parallel()
 
-	tests := []struct {
-		name           string
-		tools          []server.ServerTool
-		searchFunc     func(ctx context.Context, query string, allowedTools []string) ([]ToolMatch, error)
-		upsertFunc     func(ctx context.Context, tools []server.ServerTool) error
-		input          FindToolInput
-		expectedNames  []string
-		expectErr      bool
-		errContains    string
-		expectCreate   bool // if false, expect NewDummyOptimizer to fail
-		createErr      string
-		checkMetrics   bool
-		expectBaseline int
-		expectReturned int
-	}{
-		{
-			name: "delegates search to store with allowedTools and computes metrics",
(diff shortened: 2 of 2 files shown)


pkg/vmcp/optimizer/dummy_optimizer.go
```diff
@@ -26,32 +26,48 @@ type DummyOptimizer struct {
 
 	// tools contains all available tools indexed by name.
 	tools map[string]server.ServerTool
+
+	// tokenCounts holds precomputed per-tool token estimates, indexed by tool name.
+	tokenCounts map[string]int
```
jerm-dro, line 34: question, non-blocking: is this also immutable? Add a comment to explain.

pkg/vmcp/optimizer/dummy_optimizer_test.go
```diff
@@ -44,19 +44,22 @@ func TestDummyOptimizer_MockStore(t *testing.T) {
...
+		input          FindToolInput
+		expectedNames  []string
+		expectErr      bool
+		errContains    string
+		expectCreate   bool // if false, expect NewDummyOptimizer to fail
+		createErr      string
+		checkMetrics   bool
+		expectBaseline int
+		expectReturned int
 	}{
```
jerm-dro: Blocker: This table is getting quite complex 😵 

Can we break it up into more focused tests so it's easier to read?

jerm-dro (dismissed):
LGTM but let's simplify the tests

pkg/vmcp/optimizer/dummy_optimizer.go
aponcedeleonch: `tokenCounts` is immutable after construction — counts are computed once in `NewDummyOptimizer` and never modified. Tools are fixed per session (one optimizer per session), and the `TokenCounter` is set at configuration time so it cannot change at runtime. Added a comment to the field.

pkg/vmcp/optimizer/dummy_optimizer_test.go
aponcedeleonch: Good call — split the table-driven test into three focused tests:
- `TestDummyOptimizer_SearchDelegation`: verifies query/allowedTools delegation and metrics
- `TestDummyOptimizer_SearchError`: verifies store search errors propagate  
- `TestDummyOptimizer_UpsertError`: verifies creation fails on upsert errors

jhrozek (approved):

pkg/vmcp/optimizer/dummy_optimizer.go
```diff
@@ -62,9 +81,11 @@ func (d *DummyOptimizer) FindTool(ctx context.Context, input FindToolInput) (*Fi
 		return nil, fmt.Errorf("tool search failed: %w", err)
 	}
 
+	metrics := computeTokenMetrics(d.baselineTokens, d.tokenCounts, matches)
+
 	return &FindToolOutput{
 		Tools:        matches,
-		TokenMetrics: TokenMetrics{}, // Zero values for dummy
+		TokenMetrics: metrics,
```
jerm-dro, line 88: Question, non-blocking: who's going to use these metrics? for what?

If they're just going to be used to emit a metric, I'd suggest emitting the metric from `DummyOptimizer` and removing this field from the `FindToolOutput`.

jerm-dro: No need to respond to this comment now, but just something to think about as you iterate here

aponcedeleonch: the `TokenMetrics` are an easy way to let know AI clients about the benefits from Optimizer. So when Optimizer gets used by Claude Code / Copilot / etc, the clients could said something like:
> I found the tool X and in the meantime saved 85% of tokens

This is not meant to collect metrics in the sense of OTEL or prometheus or similar. However, we could make them part of that. Probably better to keep separate of this effort (?). I don't have a strong opinion

PR #3831, opened 2026-02-16
Harden scope handling in DCR and callback handler

### LINKED ISSUE #3745: Embedded auth server: DCR grants all supported scopes and callback lacks scope validation
## Summary

The embedded OAuth authorization server has two scope-handling gaps that weaken least-privilege enforcement. These are **not** critical privilege escalation vulnerabilities — fosite's default `WildcardScopeStrategy` already prevents clients from obtaining truly arbitrary scopes — but they reduce defense-in-depth and create risk if `ScopesSupported` is ever expanded with sensitive scopes.

### 1. DCR registers every client with ALL `ScopesSupported` scopes

In `pkg/authserver/server/handlers/dcr.go`, the DCR handler ignores what scopes a client actually needs and assigns every server-supported scope:

```go
fositeClient, err := registration.New(registration.Config{
    // ...
    Scopes: h.config.ScopesSupported, // all of them, unconditionally
})
```

There is no `scope` field on `DCRRequest`, so clients cannot request a subset. Every DCR client gets registered with the full set (default: `["openid", "offline_access"]`).

**Impact today:** Minimal — the default `ScopesSupported` is `["openid", "offline_access"]`, both of which are benign. But if an operator adds sensitive scopes (e.g., `admin`, `write`), every DCR client automatically receives them.

### 2. Callback handler grants all stored scopes unconditionally

In `pkg/authserver/server/handlers/callback.go`, the `writeAuthorizationResponse` method copies every scope from `PendingAuthorization` directly into both `RequestedScope` and `GrantedScope` with no filtering:

```go
for _, scope := range pending.Scopes {
    authorizeRequest.RequestedScope = append(authorizeRequest.RequestedScope, scope)
    authorizeRequest.GrantedScope = append(authorizeRequest.GrantedScope, scope)
}
```

This bypasses fosite's normal scope-granting flow. Under normal operation this is safe because fosite validates scopes at `/authorize` before they reach storage. But the callback constructs a fresh `fosite.NewAuthorizeRequest()` (the struct, not `provider.NewAuthorizeRequest()`) and never runs fosite's `validateAuthorizeScope()`.

**Impact today:** Not directly exploitable — invalid scopes cannot reach `PendingAuthorization.Scopes` because fosite rejects them at `/authorize`. However, it creates a latent vulnerability surface: any future code path that writes to `PendingAuthorization` without going through fosite's authorize validation would inherit blind scope granting.

### 3. No explicit `ScopeStrategy` configured (relies on fosite default)

`pkg/authserver/server/provider.go` does not set `ScopeStrategy` in the fosite config. Fosite defaults to `WildcardScopeStrategy` (`config_default.go:344`), which **does** validate scopes — the finding document's claim of "zero scope validation" is inaccurate. For simple scope names without `.` separators (like `openid`, `offline_access`), `WildcardScopeStrategy` behaves identically to `ExactScopeStrategy`.

**Impact today:** None for current scope names. A theoretical concern exists if dot-separated hierarchical scopes are introduced (e.g., a client with `read.*` registered could match `read.secret`), but this is not a current scenario. Still, explicitly setting `ExactScopeStrategy` is better practice than relying on a default.

## What Should Change

1. **Set `ScopeStrategy: fosite.ExactScopeStrategy` explicitly** in the fosite config — don't rely on the default
2. **Add `scope` field to DCR** — let clients request specific scopes; validate against `ScopesSupported`; default to `["openid", "offline_access"]` when omitted
3. **Validate scopes in the callback handler** — filter `pending.Scopes` against the client's registered scopes before granting (the `fositeClient` is already loaded at that point)
4. **Add tests** — scope rejection at `/authorize`, scope filtering in callback, DCR scope validation

## Severity Assessment

**Low-Medium.** The current default configuration is safe. The gaps become meaningful only under expanded `ScopesSupported` or future code changes to the authorize flow. The fixes are straightforward and improve defense-in-depth.

## Code Pointers

- `pkg/authserver/server/provider.go` — fosite config, missing `ScopeStrategy`
- `pkg/authserver/server/handlers/dcr.go` — DCR handler, unconditional scope assignment
- `pkg/authserver/server/handlers/callback.go` — `writeAuthorizationResponse`, blind scope granting
- `pkg/authserver/server/registration/client.go` — `DefaultScopes` definition
- `pkg/authserver/config.go` — `ScopesSupported` defaults

jhrozek:
The embedded auth server had three scope-handling gaps that weakened least-privilege enforcement. These are not directly exploitable under the current default configuration, but reduce defense-in-depth:

1. DCR registered every client with all ScopesSupported scopes unconditionally, ignoring what the client actually needs.

2. The callback handler copied all stored scopes from PendingAuthorization into GrantedScope without filtering against the client's registration.

3. No explicit ScopeStrategy was configured — fosite's default WildcardScopeStrategy works for current scope names but is less strict than necessary.

Changes:
- Set fosite ScopeStrategy to ExactScopeStrategy explicitly
- Add scope field to DCR requests; validate against ScopesSupported; default to DefaultScopes when omitted
- Filter callback scopes against client registration and log when scopes are dropped
- Deduplicate scope values
- Add integration test for scope elevation rejection
- Add prefix/substring scope validation tests
- Replace hardcoded scope slices with registration.DefaultScopes
- Fix stale ScopesSupported comments in config.go

Fixes #3745

### DIFF
--- a/pkg/authserver/server/registration/dcr.go
+++ b/pkg/authserver/server/registration/dcr.go
@@ -19,6 +19,7 @@ package registration
 
 import (
 	"slices"
+	"strings"
 
 	"github.com/stacklok/toolhive/pkg/oauth"
 )
@@ -64,6 +65,11 @@ type DCRRequest struct {
 	// ResponseTypes is an array of OAuth 2.0 response types the client may use.
 	// Defaults to ["code"] if not specified.
 	ResponseTypes []string `json:"response_types,omitempty"`
+
+	// Scope is a space-separated list of OAuth 2.0 scope values the client may use.
+	// If not specified, defaults to the server's default scopes.
+	// All requested scopes must be supported by the server.
+	Scope string `json:"scope,omitempty"`
 }
 
 // DCRResponse represents a successful OAuth 2.0 Dynamic Client Registration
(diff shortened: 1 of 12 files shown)


pkg/authserver/server/registration/dcr.go
amirejaz, line 102: nit: (non-blocking)We observed some servers return scopes in list format and the unmarshalling failed so we used a custom type: https://github.com/stacklok/toolhive/blob/main/pkg/auth/oauth/dynamic_registration.go#L152.

jhrozek: I'll address this in a separate PR, note that this is /us being a DCR server/ and emitting the scope. But there's room for deduplication.

PR #3832, opened 2026-02-16
Add extensible container runtime registry

JAORMX:
## Summary

- Introduces a global `runtime.Registry` struct so built-in runtimes self-register via `init()` and external runtimes can register by blank-importing their package
- Moves `RuntimeInfo`/`RuntimeInitializer` types from `container/factory.go` to `runtime/types.go` (renamed to `Info`/`Initializer` to avoid stutter)
- Factory seeds from the registry instead of hardcoding Docker and Kubernetes; auto-detection uses configurable priority ordering (lower = tried first)
- `NewFactoryFromRegistry()` enables fully isolated, parallel testing — no shared global state in tests

## Design

Follows the existing `pkg/authz/authorizers/registry.go` self-registration pattern:

```go
// External runtime in a separate repo:
func init() {
    runtime.RegisterRuntime(&runtime.Info{
        Name:     "myruntime",
        Priority: 150,
        Initializer: func(ctx context.Context) (runtime.Runtime, error) {
            return NewClient(ctx)
        },
    })
}
```

Priority convention: Docker=100, Kubernetes=200. External runtimes pick relative values.

## Test plan

- [x] `task build` — binary builds, no import cycles
- [x] `task test` — full unit test suite passes (with race detector)
- [x] `task lint-fix` — 0 lint issues
- [x] Registry tests: registration, dedup panic, priority ordering, same-priority name sort, isolation between instances
- [x] Factory tests: seeding from registry, auto-detection priority, skip unavailable, nil detector, none available
- [x] All tests run fully in parallel with isolated `NewRegistry()` instances
- [ ] Manual: `TOOLHIVE_RUNTIME=docker thv list` still works

### DIFF
--- a/pkg/container/runtime/registry.go
+++ b/pkg/container/runtime/registry.go
@@ -1,0 +1,132 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package runtime
+
+import (
+	"fmt"
+	"sort"
+	"sync"
+)
+
+// Registry holds a set of registered runtimes. It is safe for concurrent use.
+// Tests should create isolated instances via NewRegistry() rather than using
+// the DefaultRegistry, so they can run fully in parallel.
+type Registry struct {
+	mu       sync.RWMutex
+	runtimes map[string]*Info
+}
+
+// NewRegistry creates a new, empty Registry.
+func NewRegistry() *Registry {
+	return &Registry{
+		runtimes: make(map[string]*Info),
+	}
+}
+
+// Register adds a runtime to the registry.
+// It panics if info is nil, has an empty name, has a nil initializer,
+// or if a runtime with the same name is already registered.
+func (r *Registry) Register(info *Info) {
+	if info == nil {
+		panic("runtime info cannot be nil")
+	}
+	if info.Name == "" {
+		panic("runtime name cannot be empty")
+	}
+	if info.Initializer == nil {
+		panic("runtime initializer cannot be nil")
+	}
+	if info.Priority < 0 {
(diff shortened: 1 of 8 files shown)


pkg/container/runtime/types.go
```diff
@@ -316,6 +316,30 @@ func IsKubernetesRuntimeWithEnv(envReader env.Reader) bool {
...
+	// Name is the runtime name (e.g., "docker", "kubernetes")
+	Name string
+	// Priority determines the order in which runtimes are tried during auto-detection.
+	// Lower values are tried first. Convention:
+	//   100 = Docker (default local runtime)
+	//   200 = Kubernetes (detected via env vars)
+	//   50-99 = External runtimes preferred before Docker
+	//   101-199 = External runtimes between Docker and Kubernetes
+	//   201+ = External fallback runtimes
+	Priority int
```
jhrozek, line 335: are negative values permitted at all?

pkg/container/runtime/registry.go
```diff
@@ -0,0 +1,123 @@
...
+	defer r.mu.Unlock()
+
+	if _, exists := r.runtimes[info.Name]; exists {
+		panic(fmt.Sprintf("runtime already registered: %s", info.Name))
+	}
+	r.runtimes[info.Name] = info
+}
+
+// Get returns the Info for the given name, or nil if not found.
+func (r *Registry) Get(name string) *Info {
```
jhrozek, line 54: is it expected that we care about locks in the registry but then return pointers to the internal struct that the caller can modify at will?

JAORMX:
@claude please review this following the project's guidelines.

jhrozek (approved):

PR #3833, opened 2026-02-16
Add shared skill types

blkt:
This change adds types related to RFC `THV-0029`. They will be moved under `stacklok/toolhive-core` in the near future.

References stacklok/toolhive-registry-server#441

### DIFF
--- a/pkg/registry/registry/skills_types.go
+++ b/pkg/registry/registry/skills_types.go
@@ -1,0 +1,85 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package registry
+
+// SkillPackage represents a distribution package (OCI or Git) in request/response payloads.
+type SkillPackage struct {
+	// RegistryType is the type of registry the package is from.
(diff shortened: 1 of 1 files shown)


pkg/registry/registry/skills_types.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// SkillRepository represents source repository metadata.
+type SkillRepository struct {
+	URL  string `json:"url,omitempty"`
+	Type string `json:"type,omitempty"`
+}
+
+// PublishSkillRequest is the request body for POST /skills (publish).
+type PublishSkillRequest struct {
```
rdimitrov: Let's have PublishSkillRequest and SkillResponse be a single Skill struct, on the registry server side we can then have something like 

```go
  type SkillListResponse struct {
      Skills   []Skill           `json:"skills"`
      Metadata SkillListMetadata `json:"_meta"`
  }
```

blkt: Nice catch, that's a leftover.

blkt:
Test failure is not related to code changes, I'm merging this.

PR #3835, opened 2026-02-16
Add skill installer and extractor for OCI artifact extraction

### LINKED ISSUE #3649: Skill installer and extractor
## Description

Implement the skill installer that extracts OCI artifact layers to the appropriate client skill directory. Handles managed vs unmanaged detection, overwrite protection, and security validation (path traversal, symlinks).

## Context

Part of the [Skills Lifecycle Management epic](https://github.com/stacklok/stacklok-epics/issues/239). Ports from Skillet's `internal/infrastructure/filesystem/installer.go` and `extractor.go`. Uses the OCI store from toolhive-core (TASK-002), client paths from TASK-006, and state management from TASK-007.

**Dependencies**: stacklok/toolhive-core#14, stacklok/toolhive#3647, stacklok/toolhive#3648
**Blocks**: Skill service implementation

## Acceptance Criteria

- [ ] `pkg/skills/installer.go` extracts OCI layer tar.gz to client skill directory
- [ ] Resolves target directory using client metadata skill paths (from TASK-006)
- [ ] Detects existing skills: managed (has `.thv-skill.json`) vs unmanaged
- [ ] Overwrite protection: refuses to overwrite unmanaged skills unless `--force`
- [ ] Upgrade detection: same digest = unchanged, different digest = upgraded
- [ ] Creates `.thv-skill.json` metadata file on install
- [ ] Updates skill index on install/uninstall
- [ ] Clean removal: removes skill directory + metadata on uninstall
- [ ] Security: validates tar entries for path traversal (`../`), symlinks, absolute paths, device entries
- [ ] Size limits: max 100MB per file, 500MB total extraction
- [ ] All tests pass, SPDX headers present

## Technical Approach

Port from Skillet `internal/infrastructure/filesystem/`, adapting to use toolhive-core OCI store and ToolHive client/state systems.

### Code Pointers
- Skillet `internal/infrastructure/filesystem/installer.go` — install/uninstall logic
- Skillet `internal/infrastructure/filesystem/extractor.go` — tar extraction with security
- Skillet `internal/infrastructure/filesystem/reader.go` — skill directory reading
- `pkg/skills/types.go` — domain types (from TASK-001)

## References

- Epic: stacklok/stacklok-epics#239
- stacklok/toolhive-core#14, stacklok/toolhive#3647, stacklok/toolhive#3648

JAORMX:
Closes #3649

## Why

This is the critical-path task for the Skills Lifecycle Management epic — all remaining skill tasks are blocked on having an installer that can extract OCI artifact content to disk and clean up on uninstall.

Today, `Install()` only creates a pending DB record and `Uninstall()` only deletes it. Neither touches the filesystem.

## What

**Filesystem operations** (`pkg/skills/installer.go`):
- `Extract()` decompresses tar.gz OCI layers to client skill directories with size limits (500MB total, 100MB/file, 1000 files max) and permission sanitization (setuid/setgid stripped, capped at 0644)
- `Remove()` safely deletes skill directories with guards against dangerous paths
- Security follows skillet's defense-in-depth pattern: pre-extraction symlink validation of target path, pre-write containment check per file, post-extraction filesystem walk

**Service layer** (`pkg/skills/skillsvc/`):
- `Install()` now handles managed/unmanaged detection (same digest = no-op, different digest = upgrade, unmanaged dir on disk = refuse unless `--force`)
- `Uninstall()` removes files for all clients before deleting the DB record
- Functional options pattern (`WithPathResolver`) keeps the service extensible for #3650 (OCI store/registry)

**Wiring**:
- `PathResolver` interface in `pkg/skills/` avoids the `pkg/skills` ↔ `pkg/client` import cycle
- `clientPathAdapter` in `pkg/api/server.go` bridges `*client.ClientManager` to `PathResolver`
- `LayerData`/`Reference`/`Digest` are internal-only fields (`json:"-"`) — not exposed via HTTP API. The API path still creates pending records; extraction is only reachable from internal code (OCI pull in #3650)

## Large PR Justification

This touched the API so the swagger occupies some space.


pkg/skills/installer.go
```diff
@@ -0,0 +1,197 @@
...
+	if homeErr == nil && absPath == homeDir {
+		return fmt.Errorf("refusing to remove dangerous path %q", absPath)
+	}
+	// If we couldn't determine the home directory, refuse shallow paths as a safety net.
+	// Count path depth by splitting on separator (e.g., "/var/home/user" → 4 components).
+	if homeErr != nil && pathDepth(absPath) < 4 {
+		return fmt.Errorf("refusing to remove shallow path %q (could not determine home directory)", absPath)
+	}
+
+	if _, err := os.Stat(absPath); os.IsNotExist(err) {
```
jhrozek: elsewhere we try to protect against symlinks, maybe Remove should also use Lstat, EvalSymlinks etc?

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -97,6 +130,167 @@ func (*service) Push(_ context.Context, _ skills.PushOptions) error {
...
+	clientType, targetDir string,
+	existing skills.InstalledSkill,
+) (*skills.InstallResult, error) {
+	if _, err := skills.Extract(opts.LayerData, targetDir, true); err != nil {
+		return nil, fmt.Errorf("extracting skill upgrade: %w", err)
+	}
+
+	sk := buildInstalledSkill(opts, scope, clientType, existing.Clients)
+	if err := s.store.Update(ctx, sk); err != nil {
+		return nil, err
```
jhrozek, line 228: should we revert the side effect of Extract? (same in freshInstall)

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -33,36 +50,52 @@ func (s *service) List(ctx context.Context, opts skills.ListOptions) ([]skills.I
...
+
+	// Remove files for each client that was installed.
+	if s.pathResolver != nil {
+		for _, clientType := range existing.Clients {
+			skillPath, pathErr := s.pathResolver.GetSkillPath(clientType, opts.Name, scope, opts.ProjectRoot)
+			if pathErr != nil {
+				continue // best-effort cleanup
+			}
+			if rmErr := skills.Remove(skillPath); rmErr != nil {
+				return fmt.Errorf("removing skill files for client %q: %w", clientType, rmErr)
```
jhrozek: GetSkillPath tries to recover, but Uninstall errors hard, I think we should choose one pattern and stick with it

pkg/skills/installer.go
```diff
@@ -0,0 +1,197 @@
...
+
+	// Resolve to absolute path for safety checks
+	absPath, err := filepath.Abs(skillDir)
+	if err != nil {
+		return fmt.Errorf("resolving absolute path: %w", err)
+	}
+
+	// Guard against removing dangerous paths
+	homeDir, homeErr := os.UserHomeDir()
+	if absPath == "/" {
```
jhrozek: what's the state of Windows support in toolhive in general?

pkg/skills/installer.go
```diff
@@ -0,0 +1,197 @@
...
+	SkillDir string
+	// Files is the number of files written.
+	Files int
+}
+
+// Extract decompresses a tar.gz OCI layer and writes files to targetDir.
+// If targetDir exists and force is false, an error is returned.
+// If force is true, the existing directory is removed before extraction.
+func Extract(layerData []byte, targetDir string, force bool) (*ExtractResult, error) {
+	// Decompress gzip with total size limit
```
jhrozek, line 59: Do we need (later, in follow ups) think about serializing the methods since they are invoked from REST API?

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -97,6 +130,167 @@ func (*service) Push(_ context.Context, _ skills.PushOptions) error {
...
+		)
+	}
+
+	if _, err := skills.Extract(opts.LayerData, targetDir, opts.Force); err != nil {
+		return nil, fmt.Errorf("extracting skill: %w", err)
+	}
+
+	sk := buildInstalledSkill(opts, scope, clientType, nil)
+	if err := s.store.Create(ctx, sk); err != nil {
+		return nil, err
```
jhrozek, line 256: should we try to revert the extraction above if store operation fails?

pkg/skills/skillsvc/skillsvc.go
jhrozek, line 228: same question about reverting the extraction

pkg/skills/skillsvc/skillsvc.go
jhrozek: why do we do best effort cleanup here but abort on first remove error?

pkg/skills/installer.go
```diff
@@ -0,0 +1,197 @@
...
+	// SkillDir is the absolute path where the skill was extracted.
+	SkillDir string
+	// Files is the number of files written.
+	Files int
+}
+
+// Extract decompresses a tar.gz OCI layer and writes files to targetDir.
+// If targetDir exists and force is false, an error is returned.
+// If force is true, the existing directory is removed before extraction.
+func Extract(layerData []byte, targetDir string, force bool) (*ExtractResult, error) {
```
jhrozek, line 58: general question because I haven't been following the skills work - we seem to call these as REST API handlers, do we need to serialize the handlers to avoid races when writing?

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -97,6 +130,167 @@ func (*service) Push(_ context.Context, _ skills.PushOptions) error {
...
+	opts skills.InstallOptions,
+	scope skills.Scope,
+	clientType, targetDir string,
+	existing skills.InstalledSkill,
+) (*skills.InstallResult, error) {
+	if _, err := skills.Extract(opts.LayerData, targetDir, true); err != nil {
+		return nil, fmt.Errorf("extracting skill upgrade: %w", err)
+	}
+
+	sk := buildInstalledSkill(opts, scope, clientType, existing.Clients)
```
jhrozek, line 224: I'm not sure I understand how the client scoping works here, skills.Extract seems to extract to one client dir but we pass all existing.Clients here?

pkg/skills/installer.go
jhrozek: We seem to be trying to protect against symlinks in some codepaths but not in Remove, is it intentional?

pkg/skills/installer.go
```diff
@@ -0,0 +1,197 @@
...
+// path component for symlinks. This prevents symlink attacks where an attacker
+// places a symlink at a parent directory before extraction.
+func validatePathNoSymlinks(targetDir string) error {
+	absTarget, err := filepath.Abs(targetDir)
+	if err != nil {
+		return fmt.Errorf("resolving absolute path: %w", err)
+	}
+
+	// Walk each component from the root down, checking existing segments.
+	current := "/"
```
jhrozek: is it expected that the check is Unix only?

pkg/skills/installer.go
JAORMX: Fixed! `Remove` now uses `Lstat` to detect if the target itself is a symlink (refuses to remove it), then `EvalSymlinks` to resolve parent components before the dangerous-path checks. Added a test verifying symlinks are refused.

JAORMX: Fixed — same approach as in the other symlink thread. `Remove` now has full symlink protection via `Lstat` + `EvalSymlinks`.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Good catch — both `upgradeSkill` and `freshInstall` now call `installer.Remove(targetDir)` to roll back the extracted files if the store Create/Update fails. Also introduced an `Installer` interface for DI so the rollback behavior is properly tested with mocks.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Fixed — `freshInstall` now rolls back the extraction if `store.Create` fails, same pattern as `upgradeSkill`.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Fixed — see the other thread; both methods now roll back on store failure.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Agreed. `Uninstall` now uses best-effort throughout — both `GetSkillPath` and `Remove` errors are collected via `errors.Join` and returned after all clients have been attempted. The store `Delete` still proceeds regardless, and cleanup errors are surfaced to the caller.

JAORMX: Fixed — both are now best-effort. See the other thread for details on the `errors.Join` approach.

pkg/skills/installer.go
JAORMX: Good question. Currently the REST API doesn't serialize concurrent skill operations, so theoretically two concurrent installs of the same skill could race. In practice the risk is low (single-user CLI tool), but it's worth a follow-up. I'll file an issue for adding a mutex or file lock in the service layer.

pkg/skills/installer.go
JAORMX: See my answer in the other thread — good call, filing a follow-up issue for serialization.

pkg/skills/skillsvc/skillsvc.go
JAORMX: The `existing.Clients` list tracks *all* client types the skill has been installed for (e.g. `["claude-code", "cursor"]`). When upgrading, we extract to one `targetDir` (for the current `clientType`) but pass the full `existing.Clients` to `buildInstalledSkill` so the DB record preserves the complete list. `buildInstalledSkill` then merges `clientType` into `existingClients` if not already present. The extraction itself only happens for the single client being operated on.

pkg/skills/installer.go
JAORMX: Fixed! `Remove` now uses `filepath.Dir(p) == p` to detect filesystem roots cross-platform, and `validatePathNoSymlinks` uses `filepath.VolumeName` to determine the walk root (so it handles `C:\` and friends correctly).

pkg/skills/installer.go
JAORMX: Nope, not intentional. Fixed now. The root is derived from `filepath.VolumeName(absTarget)`, which returns `"C:"` on Windows and `""` on Unix, so we fall back to `"/"` only when there's no volume name.

pkg/skills/installer.go
JAORMX: Filed #3845 to track this.

Issue #3733, opened 2026-02-09
Wire a real embedding client for the EmbeddingServer into the optimizer


jerm-dro:
**Depends on:** #3732

### Context

Issue #3732 introduces the `EmbeddingClient` interface and hybrid search backed by a fake/stub implementation. This issue replaces the stub with a real embedding client that calls the `EmbeddingServer` custom resource, enabling true semantic tool discovery in production.

Prototype reference: https://github.com/stacklok/toolhive/pull/3517 (see server integration and CLI wiring)

### Requirements

1. Implement the `EmbeddingClient` interface with a real HTTP client that calls the `EmbeddingServer` custom resource endpoint.
2. Push the use of the `EmbeddingClient` **into the database layer** (the `ToolStore` SQLite implementation). The database is responsible for calling the embedding client at ingestion time and storing the resulting vectors. If an item already exists in the database with embeddings, no re-embedding is needed — the database acts as the cache.
3. Wire the real client into the optimizer factory in `cmd/vmcp/app/commands.go`, configured via `OptimizerConfig` fields (`pkg/vmcp/config/config.go`).
4. Add configuration fields for the EmbeddingServer URL and embedding dimensions to `OptimizerConfig`.
5. If the EmbeddingServer is unavailable, the optimizer should **fail** — no graceful degradation or fallback.
6. Update the existing e2e test that enables the optimizer (`test/e2e/thv-operator/virtualmcp/virtualmcp_optimizer_test.go`) so it passes with the real embedding client wired in.

### High-Level Implementation

- Create a real `EmbeddingClient` implementation under `pkg/vmcp/optimizer/internal/embeddings/` that calls the EmbeddingServer endpoint
- Update the SQLite `ToolStore` implementation to accept an `EmbeddingClient` dependency. At tool ingestion time, the store calls the client to generate embeddings and persists them alongside the tool record. On lookup, embeddings are already present in the database.
- Extend `OptimizerConfig` with fields for the EmbeddingServer URL and embedding dimensions
- Update the factory setup in `cmd/vmcp/app/commands.go` to instantiate the real client from config and pass it to the `ToolStore`
- Update the existing e2e optimizer test to exercise the full path with a real EmbeddingServer

### Acceptance Tests

- [ ] **Real embedding client**: Successfully calls the EmbeddingServer endpoint and returns vectors of the configured dimensionality
[...]

PR #3839, opened 2026-02-16
Add EmbeddingServerRef to VirtualMCPServer for optimizer integration

aponcedeleonch:
Related to: https://github.com/stacklok/toolhive/issues/3733

Add `embeddingServerRef` field to VirtualMCPServer, allowing it to reference an existing EmbeddingServer resource for optimizer integration. The operator validates the reference, waits for EmbeddingServer readiness, and auto-populates the optimizer config with the resolved embedding service URL.

Key changes:
- Add `EmbeddingServerRef` type and spec field on VirtualMCPServerSpec
- Add webhook validation: ref name must be non-empty, optimizer requires either `embeddingServerRef` or manual `embeddingService`
- Auto-populate optimizer with defaults when `embeddingServerRef` is set without explicit optimizer config
- Add controller-level validation for referenced EmbeddingServer existence and readiness
- Add `EmbeddingServerReady` condition type and status reporting
- Add EmbeddingServer watch so status changes trigger VirtualMCPServer reconciliation
- Resolve embedding service URL from EmbeddingServer `Status.URL` into optimizer config
- Extract `runValidations` helper to reduce Reconcile complexity
- Make EmbeddingServer `model` and `image` fields optional with sensible defaults (`BAAI/bge-small-en-v1.5`, `cpu-latest`)
- Make OptimizerConfig `embeddingService` optional (auto-populated by operator)
- Update E2E tests to use a real EmbeddingServer instead of dummy values
- Update optimizer example YAML to demonstrate the `embeddingServerRef` flow

## Large PR Justification

- Multiple tightly coupled changes that would break if separated: CRD types, validation, controller logic, config wiring, and status reporting all depend on each other
- ~400 lines of auto-generated code (deepcopy, CRD YAMLs, mocks), ~400 of tests, and ~400 of logic

### DIFF
--- a/cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
+++ b/cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
@@ -389,28 +389,36 @@ func (r *VirtualMCPServer) Validate() error {
 	return r.validateEmbeddingServer()
 }
 
-// validateEmbeddingServer validates EmbeddingServerRef configuration.
+// validateEmbeddingServer validates EmbeddingServerRef and Optimizer configuration.
 // Rules:
-// - If config.optimizer is set, embeddingServerRef must be set
-// - If embeddingServerRef is set, config.optimizer must be set
 // - embeddingServerRef.name must be non-empty when ref is provided
+// - optimizer requires either embeddingServerRef or a manually set embeddingService
+// - if embeddingServerRef is set without optimizer, auto-populate optimizer with defaults
+//
+// The controller handles the remaining cases at runtime (event emission, URL population).
 func (r *VirtualMCPServer) validateEmbeddingServer() error {
-	hasRef := r.Spec.EmbeddingServerRef != nil
-	hasOptimizer := r.Spec.Config.Optimizer != nil
-
-	// If optimizer is set, embeddingServerRef must be set
-	if hasOptimizer && !hasRef {
-		return fmt.Errorf("spec.config.optimizer requires spec.embeddingServerRef to be set")
-	}
-
-	// If embeddingServerRef is set, optimizer must be set
-	if hasRef && !hasOptimizer {
-		return fmt.Errorf("spec.embeddingServerRef requires spec.config.optimizer to be set")
-	}
-
 	// Validate ref name is non-empty
-	if hasRef && r.Spec.EmbeddingServerRef.Name == "" {
+	if r.Spec.EmbeddingServerRef != nil && r.Spec.EmbeddingServerRef.Name == "" {
 		return fmt.Errorf("spec.embeddingServerRef.name is required")
+	}
+
+	hasOptimizer := r.Spec.Config.Optimizer != nil
+	hasRef := r.Spec.EmbeddingServerRef != nil
+	hasManualService := hasOptimizer && r.Spec.Config.Optimizer.EmbeddingService != ""
+
+	// Optimizer configured without any embedding source is an error.
+	// The user must either set embeddingServerRef or manually set optimizer.embeddingService.
+	if hasOptimizer && !hasRef && !hasManualService {
+		return fmt.Errorf(
+			"spec.config.optimizer requires an embedding service: " +
+				"set spec.embeddingServerRef (recommended) or spec.config.optimizer.embeddingService")
+	}
+
+	// EmbeddingServerRef is set but optimizer is not configured: auto-populate
+	// optimizer with default values so the embedding server is actually used.
+	// The controller emits a Kubernetes event for this case.
+	if hasRef && !hasOptimizer {
+		r.Spec.Config.Optimizer = &config.OptimizerConfig{}
 	}
 
 	return nil
--- a/cmd/thv-operator/controllers/virtualmcpserver_controller.go
+++ b/cmd/thv-operator/controllers/virtualmcpserver_controller.go
@@ -128,9 +128,13 @@ func (r *VirtualMCPServerReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	statusManager := virtualmcpserverstatus.NewStatusManager(vmcp)
 
 	// Run all pre-reconciliation validations.
-	// Returns an error when validation failed. Returns nil to continue.
-	if err := r.runValidations(ctx, vmcp, statusManager); err != nil {
+	// Returns (true, nil) to continue, (false, nil) when validation failed but
+	// should not requeue (user must fix spec), or (false, err) for transient errors
+	// that should trigger requeue.
+	if cont, err := r.runValidations(ctx, vmcp, statusManager); err != nil {
 		return ctrl.Result{}, err
+	} else if !cont {
+		return ctrl.Result{}, nil
 	}
 
 	// Ensure all resources
@@ -262,26 +266,31 @@ func (r *VirtualMCPServerReconciler) applyStatusUpdates(
 }
 
 // runValidations runs all pre-reconciliation validations (PodTemplateSpec, GroupRef,
-// CompositeToolRefs, EmbeddingServerRef). Returns an error when validation failed.
-// Returns nil to continue.
+// CompositeToolRefs, EmbeddingServerRef).
+// Returns (true, nil) to continue reconciliation.
+// Returns (false, nil) for spec validation errors that should NOT trigger requeue
+// (user must fix the spec; next reconciliation is triggered by spec changes).
+// Returns (false, error) for transient errors that should trigger requeue.
 func (r *VirtualMCPServerReconciler) runValidations(
 	ctx context.Context,
 	vmcp *mcpv1alpha1.VirtualMCPServer,
 	statusManager virtualmcpserverstatus.StatusManager,
-) error {
+) (bool, error) {
 	ctxLogger := log.FromContext(ctx)
 
-	// Validate spec configuration early (schema-level validation from types.go)
+	// Validate spec configuration early (schema-level validation from types.go).
+	// Don't requeue on validation errors — user must fix spec.
 	if err := r.validateSpec(ctx, vmcp, statusManager); err != nil {
-		return err
+		return false, nil
 	}
 
-	// Validate PodTemplateSpec early - before other validations
+	// Validate PodTemplateSpec early - before other validations.
+	// Don't requeue — user must fix the PodTemplateSpec.
 	if !r.validateAndUpdatePodTemplateStatus(ctx, vmcp, statusManager) {
 		if err := r.applyStatusUpdates(ctx, vmcp, statusManager); err != nil {
 			ctxLogger.Error(err, "Failed to apply status updates after PodTemplateSpec validation error")
 		}
-		return fmt.Errorf("PodTemplateSpec validation failed")
+		return false, nil
 	}
 
 	// Validate GroupRef
@@ -289,7 +298,7 @@ func (r *VirtualMCPServerReconciler) runValidations(
 		if applyErr := r.applyStatusUpdates(ctx, vmcp, statusManager); applyErr != nil {
 			ctxLogger.Error(applyErr, "Failed to apply status updates after GroupRef validation error")
 		}
-		return err
+		return false, err
 	}
 
 	// Validate CompositeToolRefs
@@ -297,7 +306,7 @@ func (r *VirtualMCPServerReconciler) runValidations(
 		if applyErr := r.applyStatusUpdates(ctx, vmcp, statusManager); applyErr != nil {
 			ctxLogger.Error(applyErr, "Failed to apply status updates after CompositeToolRefs validation error")
 		}
-		return err
+		return false, err
 	}
 
 	// Validate EmbeddingServerRef (when using reference mode)
@@ -306,11 +315,11 @@ func (r *VirtualMCPServerReconciler) runValidations(
 			if applyErr := r.applyStatusUpdates(ctx, vmcp, statusManager); applyErr != nil {
 				ctxLogger.Error(applyErr, "Failed to apply status updates after EmbeddingServerRef validation error")
 			}
-			return err
+			return false, err
 		}
 	}
 
-	return nil
+	return true, nil
 }
 
 // validateGroupRef validates that the referenced MCPGroup exists and is ready
@@ -543,9 +552,12 @@ func (r *VirtualMCPServerReconciler) ensureAllResources(
 	// Check EmbeddingServer readiness before proceeding to Deployment.
 	// RequeueAfter provides a safety net in case the Watches() events
 	// are missed (e.g., EmbeddingServer controller not running).
-	if ready, _, err := r.isEmbeddingServerReady(ctx, vmcp); err != nil {
+	esURL, err := r.isEmbeddingServerReady(ctx, vmcp)
+	if err != nil {
 		return ctrl.Result{}, err
-	} else if !ready {
+	}
+	// EmbeddingServer is configured but not yet ready — requeue
+	if esURL == nil && vmcp.Spec.EmbeddingServerRef != nil {
 		statusManager.SetPhase(mcpv1alpha1.VirtualMCPServerPhasePending)
 		statusManager.SetMessage("Waiting for EmbeddingServer to become ready")
 		statusManager.SetEmbeddingServerReadyCondition(
@@ -557,7 +569,7 @@ func (r *VirtualMCPServerReconciler) ensureAllResources(
 	}
 
 	// If an embedding server is configured and ready, set the condition
-	if embeddingServerNameForVMCP(vmcp) != "" {
+	if esURL != nil {
 		statusManager.SetEmbeddingServerReadyCondition(
 			mcpv1alpha1.ConditionReasonEmbeddingServerReady,
 			"EmbeddingServer is ready",
--- a/cmd/thv-operator/controllers/virtualmcpserver_embedding.go
+++ b/cmd/thv-operator/controllers/virtualmcpserver_embedding.go
@@ -14,35 +14,39 @@ import (
 )
 
 // isEmbeddingServerReady checks whether the referenced EmbeddingServer
-// is running and ready. If no embedding server is configured, returns true (no gate).
+// is running and ready. Returns a non-nil *string with the URL when ready.
+// Returns nil if no embedding server is configured (no gate).
+// The caller should check if vmcp.Spec.EmbeddingServerRef != nil && result == nil
+// to detect the "configured but not ready" case that requires requeue.
 func (r *VirtualMCPServerReconciler) isEmbeddingServerReady(
 	ctx context.Context,
 	vmcp *mcpv1alpha1.VirtualMCPServer,
-) (bool, string, error) {
+) (*string, error) {
 	name := embeddingServerNameForVMCP(vmcp)
 	if name == "" {
-		return true, "", nil // No embedding server configured, skip check
+		return nil, nil // No embedding server configured, skip check
 	}
 
 	es := &mcpv1alpha1.EmbeddingServer{}
 	err := r.Get(ctx, types.NamespacedName{Name: name, Namespace: vmcp.Namespace}, es)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			return false, "", nil // Informer cache may not have caught up yet
+			return nil, nil // Informer cache may not have caught up yet
 		}
-		return false, "", fmt.Errorf("failed to get EmbeddingServer %s: %w", name, err)
+		return nil, fmt.Errorf("failed to get EmbeddingServer %s: %w", name, err)
 	}
 
 	if es.Status.Phase == mcpv1alpha1.EmbeddingServerPhaseRunning && es.Status.ReadyReplicas > 0 {
-		return true, es.Status.URL, nil
+		url := es.Status.URL
+		return &url, nil
 	}
 
 	// Propagate failure so the VirtualMCPServer surfaces it instead of staying Pending
 	if es.Status.Phase == mcpv1alpha1.EmbeddingServerPhaseFailed {
-		return false, "", fmt.Errorf("EmbeddingServer %s has failed", name)
+		return nil, fmt.Errorf("EmbeddingServer %s has failed", name)
 	}
 
-	return false, "", nil
+	return nil, nil // Not ready yet
 }
 
 // resolveEmbeddingServiceURL looks up the referenced EmbeddingServer CR
--- a/cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
+++ b/cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
@@ -10,6 +10,7 @@ import (
 	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	mcpv1alpha1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1alpha1"
 	"github.com/stacklok/toolhive/cmd/thv-operator/pkg/kubernetes/configmaps"
@@ -50,17 +51,9 @@ func (r *VirtualMCPServerReconciler) ensureVmcpConfigConfigMap(
 		return err
 	}
 
-	// Auto-populate embedding service URL from the referenced EmbeddingServer.
-	// The URL is the full base URL (e.g., http://name.namespace.svc.cluster.local:port)
-	// so the optimizer can use it directly as an HTTP client endpoint.
-	if config.Optimizer != nil {
-		esURL, err := r.resolveEmbeddingServiceURL(ctx, vmcp)
-		if err != nil {
-			return fmt.Errorf("failed to resolve embedding service URL: %w", err)
-		}
-		if esURL != "" {
-			config.Optimizer.EmbeddingService = esURL
-		}
+	// Auto-populate optimizer config from EmbeddingServerRef or emit warnings.
+	if err := r.populateOptimizerEmbeddingService(ctx, vmcp, config); err != nil {
+		return err
 	}
 
 	// Validate the vmcp Config before creating the ConfigMap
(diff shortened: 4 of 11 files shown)


cmd/thv-operator/controllers/virtualmcpserver_embedding.go
```diff
@@ -0,0 +1,162 @@
...
+func embeddingServerName(vmcpName string) string {
+	return fmt.Sprintf("%s-embedding", vmcpName)
+}
+
+// labelsForEmbeddingServer returns labels for an auto-deployed EmbeddingServer.
+func labelsForEmbeddingServer(vmcpName string) map[string]string {
+	return map[string]string{
+		"toolhive.stacklok.io/component":          "embedding-server",
+		"toolhive.stacklok.io/virtual-mcp-server": vmcpName,
+		"toolhive.stacklok.io/managed-by":         "toolhive-operator",
```
ChrisJBurns: **Critical: Label domain inconsistency**

The project's API group and all authoritative labels use `toolhive.stacklok.dev`, but these labels use `toolhive.stacklok.io`. This will cause label selector mismatches and debugging confusion.

```suggestion
		"toolhive.stacklok.dev/component":          "embedding-server",
		"toolhive.stacklok.dev/virtual-mcp-server": vmcpName,
		"toolhive.stacklok.dev/managed-by":         "toolhive-operator",
```

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -1546,13 +1613,112 @@ func convertBackendsToStaticBackends(
...
+		)
+		statusManager.SetObservedGeneration(vmcp.Generation)
+		return err
+	} else if err != nil {
+		ctxLogger.Error(err, "Failed to get referenced EmbeddingServer", "name", refName)
+		return err
+	}
+
+	// Check if EmbeddingServer is ready
+	if es.Status.Phase != mcpv1alpha1.EmbeddingServerPhaseRunning || es.Status.ReadyReplicas == 0 {
```
ChrisJBurns: **Critical: Duplicate readiness check creates conflicting retry strategies for ref mode**

When `EmbeddingServerRef` is set, the EmbeddingServer readiness is checked in **two** places during the same reconciliation:

1. Here in `validateEmbeddingServerRef()` — returns an **error** if not ready, which triggers controller-runtime's **exponential backoff**
2. In `ensureAllResources()` → `isEmbeddingServerReady()` (line ~510) — returns `RequeueAfter: 30s` for a **fixed interval** requeue

For the ref case, this error fires first, so the intended 30-second fixed requeue is never reached. Exponential backoff is likely not the desired behavior for a transient "not ready yet" condition.

**Recommendation**: Make `validateEmbeddingServerRef` only validate **existence** (not readiness). Let `isEmbeddingServerReady` handle all readiness gating uniformly for both inline and ref modes. This simplifies the logic and ensures consistent retry behavior.

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
```diff
@@ -0,0 +1,162 @@
...
+//
+// When EmbeddingServerRef is used instead, this function is a no-op because
+// the referenced EmbeddingServer is not owned by this VirtualMCPServer.
+func (r *VirtualMCPServerReconciler) ensureEmbeddingServer(
+	ctx context.Context,
+	vmcp *mcpv1alpha1.VirtualMCPServer,
+) error {
+	// Skip if no inline embedding server spec is configured.
+	// When EmbeddingServerRef is used, the referenced resource is externally managed.
+	if vmcp.Spec.EmbeddingServer == nil {
```
ChrisJBurns: **Important: Switching from inline to ref mode leaves an orphaned EmbeddingServer**

If a user changes from `spec.embeddingServer` (inline) to `spec.embeddingServerRef`, the previously-created owned EmbeddingServer persists. This early return means no cleanup happens.

While owner references handle GC on VirtualMCPServer deletion, during the transition the zombie EmbeddingServer consumes resources (potentially GPU/memory).

**Recommendation**: When `EmbeddingServerRef` is set (ref mode), check for and delete any existing owned EmbeddingServer with the generated name pattern. Something like:

```go
if vmcp.Spec.EmbeddingServer == nil {
    // Clean up any previously-owned inline EmbeddingServer when switching to ref mode
    if vmcp.Spec.EmbeddingServerRef != nil {
        if err := r.cleanupOwnedEmbeddingServer(ctx, vmcp); err != nil {
            return err
        }
    }
    return nil
}
```

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return ""
+}
+
+// embeddingServerSpecChanged returns true if the desired spec differs from
+// the existing EmbeddingServer spec. Uses DeepEqual to ensure all fields
+// (Model, Image, Port, HFTokenSecretRef, Args, Env, Resources, Replicas, etc.)
+// are compared so that spec changes are never silently ignored.
+func embeddingServerSpecChanged(existing *mcpv1alpha1.EmbeddingServer, desired *mcpv1alpha1.EmbeddingServerSpec) bool {
+	return !reflect.DeepEqual(existing.Spec, *desired)
```
ChrisJBurns: **Important: `reflect.DeepEqual` is fragile with Kubernetes defaulting**

Kubernetes defaulting webhooks or admission controllers may set default values on the existing spec (e.g., `imagePullPolicy`, `port`, `replicas`) that aren't present in the desired spec if the user omitted them. This causes **spurious updates every reconciliation loop**.

`reflect.DeepEqual` with `*runtime.RawExtension` fields (PodTemplateSpec) can also behave unexpectedly when one side has `nil` and the other has an empty byte slice.

The existing `deploymentNeedsUpdate` in this controller does field-by-field comparison specifically to avoid this problem. Consider the same approach here, or at minimum add a unit test that verifies defaulted specs don't trigger false positives.

cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
```diff
@@ -80,6 +80,16 @@ func (r *VirtualMCPServerReconciler) ensureVmcpConfigConfigMap(
 		}
 	}
 
+	// Auto-populate embedding service name for embedding servers (inline or referenced).
+	// When the VirtualMCPServer has an embeddingServer spec (inline), the operator creates
+	// an EmbeddingServer CR and wires its service name into the optimizer config.
+	// When embeddingServerRef is used, the referenced EmbeddingServer's name is used directly.
+	if config.Optimizer != nil {
+		if esName := embeddingServerNameForVMCP(vmcp); esName != "" {
+			config.Optimizer.EmbeddingService = esName
```
ChrisJBurns: **Important: Verify Service name wiring is correct**

`embeddingServerNameForVMCP` returns the EmbeddingServer CR name (e.g., `my-vmcp-embedding`), but `EmbeddingService` is described as "the name of a Kubernetes Service." 

Please verify that the EmbeddingServer controller creates a Service with the **same name** as the CR. If it uses a different naming convention (e.g., `{name}-svc`), this wiring will silently break and the optimizer won't find the embedding endpoint.

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
```diff
@@ -0,0 +1,162 @@
...
+	name := embeddingServerNameForVMCP(vmcp)
+	if name == "" {
+		return true, "", nil // No embedding server configured, skip check
+	}
+
+	es := &mcpv1alpha1.EmbeddingServer{}
+	err := r.Get(ctx, types.NamespacedName{Name: name, Namespace: vmcp.Namespace}, es)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, "", nil // Not yet created
```
ChrisJBurns: **Important: Misleading comment after `ensureEmbeddingServer` just ran**

`ensureEmbeddingServer` is called immediately before `isEmbeddingServerReady`. If creation succeeded, a NotFound here means an informer cache delay, not "not yet created." Consider adding a debug log so operators can diagnose timing issues:

```suggestion
			return false, "", nil // Informer cache may not have caught up yet
```

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -1546,13 +1613,112 @@ func convertBackendsToStaticBackends(
 	return static
 }
 
+// validateEmbeddingServerRef validates that the referenced EmbeddingServer exists and is ready.
+// This follows the same pattern as validateGroupRef.
+func (r *VirtualMCPServerReconciler) validateEmbeddingServerRef(
```
ChrisJBurns, line 1723: **Suggestion: Missing Event recording**

Other validation functions (e.g., `validateAndUpdatePodTemplateStatus`, `validateGroupRef`) record events via `r.Recorder.Eventf`. This function does not record any events when the EmbeddingServer is not found or not ready.

This makes `kubectl describe virtualmcpserver` less useful for debugging. Consider adding events for not-found and not-ready cases, consistent with the existing pattern.

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
```diff
@@ -0,0 +1,162 @@
...
+			return fmt.Errorf("failed to set controller reference on EmbeddingServer: %w", err)
+		}
+
+		ctxLogger.Info("Creating EmbeddingServer", "name", embeddingName, "namespace", vmcp.Namespace)
+		if err := r.Create(ctx, es); err != nil {
+			return fmt.Errorf("failed to create EmbeddingServer: %w", err)
+		}
+
+		if r.Recorder != nil {
+			r.Recorder.Eventf(vmcp, "Normal", "EmbeddingServerCreated",
```
ChrisJBurns: **Nit: Use `corev1.EventTypeNormal` instead of string literal**

Other places in the controller use the `corev1.EventTypeNormal` constant. Use it here for consistency.

```suggestion
			r.Recorder.Eventf(vmcp, corev1.EventTypeNormal, "EmbeddingServerCreated",
```

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// EmbeddingServer exists - check if spec needs updating
+	if embeddingServerSpecChanged(existing, vmcp.Spec.EmbeddingServer) {
+		existing.Spec = *vmcp.Spec.EmbeddingServer
+		ctxLogger.Info("Updating EmbeddingServer spec", "name", embeddingName)
+		if err := r.Update(ctx, existing); err != nil {
+			return fmt.Errorf("failed to update EmbeddingServer: %w", err)
+		}
+		if r.Recorder != nil {
+			r.Recorder.Eventf(vmcp, "Normal", "EmbeddingServerUpdated",
```
ChrisJBurns: **Nit: Same as above — use `corev1.EventTypeNormal`**

```suggestion
			r.Recorder.Eventf(vmcp, corev1.EventTypeNormal, "EmbeddingServerUpdated",
```

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -452,45 +489,75 @@ func (r *VirtualMCPServerReconciler) ensureAllResources(
...
+		return ctrl.Result{}, err
+	} else if !ready {
+		statusManager.SetPhase(mcpv1alpha1.VirtualMCPServerPhasePending)
+		statusManager.SetMessage("Waiting for EmbeddingServer to become ready")
+		statusManager.SetEmbeddingServerReadyCondition(
+			mcpv1alpha1.ConditionReasonEmbeddingServerNotReady,
+			"EmbeddingServer is not yet ready",
+			metav1.ConditionFalse,
+		)
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
ChrisJBurns, line 568: **Suggestion: Consider shorter RequeueAfter for inline case**

For inline EmbeddingServers, the `Owns()` watch triggers reconciliation almost immediately on status changes. The 30-second requeue is described as a "safety net" — consider a shorter interval (e.g., 10s) for inline since the `Owns()` watch handles the primary path. The 30s may be more appropriate for the ref case only.

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
aponcedeleonch: Fixed — updated all three labels to use `toolhive.stacklok.dev` domain. Good catch\!

cmd/thv-operator/controllers/virtualmcpserver_controller.go
aponcedeleonch: Great observation. Refactored `validateEmbeddingServerRef` to only validate **existence** (not readiness). Readiness gating is now handled uniformly by `isEmbeddingServerReady` in `ensureAllResources` for both inline and ref modes, ensuring consistent fixed-interval requeue behavior. Updated tests accordingly.

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
aponcedeleonch: Added `cleanupOwnedEmbeddingServer()` that checks for and deletes the previously-owned inline EmbeddingServer when switching to ref mode. It verifies ownership via OwnerReferences before deleting, so it won't accidentally delete an unrelated CR with the same name pattern.

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
aponcedeleonch: Replaced `reflect.DeepEqual` with field-by-field comparison in the renamed `embeddingServerSpecNeedsUpdate()`. Key improvements:
- Simple scalar fields compared directly
- `*runtime.RawExtension` (PodTemplateSpec) compared via `bytes.Equal` on raw bytes, treating nil and empty as equal
- `*SecretKeyRef`, `*ModelCacheConfig`, `*int32` get nil-safe pointer comparisons
- `EmbeddingResourceOverrides` still uses `reflect.DeepEqual` since it only contains maps (no RawExtension/defaulting concerns)
- Added a doc comment noting that new `EmbeddingServerSpec` fields need a corresponding comparison line

cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
aponcedeleonch: Verified — the EmbeddingServer controller creates the Service with the **same name** as the CR (`embeddingserver_controller.go:273`: `Name: embedding.Name`), so the wiring is correct. No code change needed.

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
aponcedeleonch: Fixed — updated the comment to `// Informer cache may not have caught up yet` as suggested.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
aponcedeleonch: Added event recording for the not-found case in `validateEmbeddingServerRef` (`EmbeddingServerRefNotFound` warning event). Since readiness checking was moved out of this function (per [comment](https://github.com/stacklok/toolhive/pull/3839#discussion_r2813993940)), the not-ready event is now emitted in `ensureAllResources` where the readiness gate lives.

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
aponcedeleonch: Fixed — now using `corev1.EventTypeNormal`.

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
aponcedeleonch: Fixed — same as above.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
aponcedeleonch: Good idea. Added `embeddingServerRequeueInterval()` that returns **10s** for inline (where `Owns()` handles the primary path) and **30s** for ref (where `Watches()` may be slower).

yrobla (approved):

cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
```diff
@@ -58,6 +58,29 @@ type VirtualMCPServerSpec struct {
...
+	// +optional
+	EmbeddingServer *EmbeddingServerSpec `json:"embeddingServer,omitempty"`
+
+	// EmbeddingServerRef references an existing EmbeddingServer resource by name.
+	// Use this instead of EmbeddingServer when multiple VirtualMCPServers should share
+	// a single EmbeddingServer (e.g., when using the same embedding model).
+	// The referenced EmbeddingServer must exist in the same namespace and be ready.
+	// Mutually exclusive with EmbeddingServer.
+	// +optional
+	EmbeddingServerRef *EmbeddingServerRef `json:"embeddingServerRef,omitempty"`
```
jerm-dro: praise: nice clear comments explaining the relationships between the related config fields.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -258,6 +241,58 @@ func (r *VirtualMCPServerReconciler) applyStatusUpdates(
 	return nil
 }
 
+// runValidations runs all pre-reconciliation validations (PodTemplateSpec, GroupRef,
+// CompositeToolRefs, EmbeddingServerRef). Returns (true, err) when a validation failed
+// and the caller should return immediately. Returns (false, nil) to continue.
```
jerm-dro: suggestion: just return `error`. If the `done` value can be inferred from the nil-ness of the error, then the returned bool is redundant.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -1577,13 +1646,95 @@ func convertBackendsToStaticBackends(
...
+	vmcpList := &mcpv1alpha1.VirtualMCPServerList{}
+	if err := r.List(ctx, vmcpList, client.InNamespace(es.Namespace)); err != nil {
+		log.FromContext(ctx).Error(err, "Failed to list VirtualMCPServers for EmbeddingServer watch")
+		return nil
+	}
+
+	var requests []reconcile.Request
+	for _, vmcp := range vmcpList.Items {
+		// Only match VirtualMCPServers that reference this EmbeddingServer by name
+		if vmcp.Spec.EmbeddingServerRef != nil && vmcp.Spec.EmbeddingServerRef.Name == es.Name {
```
jerm-dro, line 1786: Blocker: what if the EmbeddingServer is created by the `EmbeddingServer` field (the 1:1 / auto-provisioning case)? Should changes to the underlying server (e.g. a user manually modified the server) cause reconciliation?

cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
```diff
@@ -58,6 +58,29 @@ type VirtualMCPServerSpec struct {
 	// The telemetry and audit config from here are also supported, but not required.
 	// +optional
 	Config config.Config `json:"config,omitempty"`
+
+	// EmbeddingServer optionally deploys an owned EmbeddingServer when the optimizer is enabled.
+	// If set, the controller creates an EmbeddingServer CR and auto-populates
+	// the optimizer's embeddingService field with the full service URL.
+	// Mutually exclusive with EmbeddingServerRef.
+	// +optional
+	EmbeddingServer *EmbeddingServerSpec `json:"embeddingServer,omitempty"`
```
jerm-dro: Question, blocking: What do you think about removing this? This field creates a bit of cognitive load / foot gun potential for operators. The benefit seems small; the operators still have define a full `EmbeddingServerSpec`. Not to mention it adds complexity to the code, because we now have 3 ways to configure the embedding server.

jerm-dro (changes requested):

cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
aponcedeleonch: Thank you!

cmd/thv-operator/controllers/virtualmcpserver_controller.go
aponcedeleonch: Good suggestion! Will change the return to just `error` — the boolean was unnecessary complexity.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
aponcedeleonch: Good question! Since we're removing inline mode (see below), this simplifies to the ref case only.

For `EmbeddingServerRef`, changes to the referenced EmbeddingServer **already trigger reconciliation** via `Watches(&mcpv1alpha1.EmbeddingServer{}, mapEmbeddingServerToVirtualMCPServer)`. The mapper lists VirtualMCPServers and matches by ref name, so readiness/status changes in the referenced EmbeddingServer correctly trigger VirtualMCPServer reconciliation.

I'll also update the comments to reflect the ref-only mode.

cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
aponcedeleonch: Agreed — I originally started with inline only, then added ref for reusability. Having both adds complexity with marginal benefit. Removing inline, keeping only `EmbeddingServerRef`.

cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
```diff
@@ -358,6 +385,34 @@ func (r *VirtualMCPServer) Validate() error {
...
+
+	// If optimizer is set, embeddingServerRef must be set
+	if hasOptimizer && !hasRef {
+		return fmt.Errorf("spec.config.optimizer requires spec.embeddingServerRef to be set")
+	}
+
+	// If embeddingServerRef is set, optimizer must be set
+	if hasRef && !hasOptimizer {
+		return fmt.Errorf("spec.embeddingServerRef requires spec.config.optimizer to be set")
+	}
```
jerm-dro: Blocker: I think this validation is too strict. It forces the user to define the optimizer config in two spots. It would make more sense to only expect the `EmbeddingServerRef` or `Config.Optimizer.Service`. If the ref is populated, then it overwrites the service.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -103,46 +105,25 @@ func (r *VirtualMCPServerReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-	}
-
-	// Validate PodTemplateSpec early - before other validations
-	if !r.validateAndUpdatePodTemplateStatus(ctx, vmcp, statusManager) {
-		// Invalid PodTemplateSpec - apply status updates and return without error to avoid infinite retries
-		// The user must fix the spec and the next reconciliation will retry
-		if err := r.applyStatusUpdates(ctx, vmcp, statusManager); err != nil {
-			ctxLogger.Error(err, "Failed to apply status updates after PodTemplateSpec validation error")
-		}
-		return ctrl.Result{}, nil
```
jerm-dro, line 140: note, blocking: with the refactor, the equivalent lines now return an error. The comment above suggests this was done intentionally, so let's preserve the behavior of not returning an error in this case.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -483,45 +518,70 @@ func (r *VirtualMCPServerReconciler) ensureAllResources(
...
+		return ctrl.Result{}, err
+	} else if !ready {
+		statusManager.SetPhase(mcpv1alpha1.VirtualMCPServerPhasePending)
+		statusManager.SetMessage("Waiting for EmbeddingServer to become ready")
+		statusManager.SetEmbeddingServerReadyCondition(
+			mcpv1alpha1.ConditionReasonEmbeddingServerNotReady,
+			"EmbeddingServer is not yet ready",
+			metav1.ConditionFalse,
+		)
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
jerm-dro, line 568: praise: nice use of RequeueAfter

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
```diff
@@ -0,0 +1,76 @@
...
+
+	mcpv1alpha1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1alpha1"
+)
+
+// isEmbeddingServerReady checks whether the referenced EmbeddingServer
+// is running and ready. If no embedding server is configured, returns true (no gate).
+func (r *VirtualMCPServerReconciler) isEmbeddingServerReady(
+	ctx context.Context,
+	vmcp *mcpv1alpha1.VirtualMCPServer,
+) (bool, string, error) {
```
jerm-dro: suggestion: return `(*string, error)` where the first result may be nil if the no embedding server is required.

cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
```diff
@@ -46,37 +46,21 @@ func (r *VirtualMCPServerReconciler) ensureVmcpConfigConfigMap(
...
-		}
-		if discoveredAuthConfig != nil {
-			config.OutgoingAuth = discoveredAuthConfig
-		}
-
-		// Discover backends with metadata
-		backends, err := r.discoverBackendsWithMetadata(ctx, vmcp)
-		if err != nil {
-			return fmt.Errorf("failed to discover backends for static mode: %w", err)
+		if err := r.populateStaticModeBackends(ctx, vmcp, config, typedWorkloads); err != nil {
```
jerm-dro: Thanks for making tiny refactors like this!

cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
```diff
@@ -46,37 +46,21 @@ func (r *VirtualMCPServerReconciler) ensureVmcpConfigConfigMap(
...
+			return err
 		}
+	}
 
-		// Get transport types from workload specs
-		transportMap, err := r.buildTransportMap(ctx, vmcp.Namespace, typedWorkloads)
+	// Auto-populate embedding service URL from the referenced EmbeddingServer.
+	// The URL is the full base URL (e.g., http://name.namespace.svc.cluster.local:port)
+	// so the optimizer can use it directly as an HTTP client endpoint.
+	if config.Optimizer != nil {
```
jerm-dro: suggestion: Going back to my earlier comment, I think it should be okay to not configure `config.Optimizer` and let the ref populate the optimizer config with all its defaults + embedding service.

jerm-dro (changes requested):
some minor changes requested, but this mostly LGTM

cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
aponcedeleonch: Addressed! The validation was reworked to be much more flexible:

- **`EmbeddingServerRef` alone** → auto-populates optimizer with default values (no need to set optimizer explicitly)
- **`optimizer.embeddingService` alone** → works with a warning event recommending `embeddingServerRef`
- **Both ref + optimizer** → ref overrides `embeddingService` (with a warning event)
- **Only errors** when optimizer is set without *any* embedding source (no ref, no manual service)

The controller's `populateOptimizerFromRef()` handles the ref-overrides-service logic at runtime with appropriate Kubernetes events for each case. Tests updated to cover all combinations.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
aponcedeleonch: Fixed! `runValidations` now returns `(bool, error)`:

- `(false, nil)` for spec validation failures → caller returns `ctrl.Result{}, nil` (no requeue; user must fix the spec, next reconciliation triggers on spec change)
- `(false, err)` for transient errors → caller returns the error for requeue
- `(true, nil)` → continue reconciliation

This preserves the original behavior of not returning an error for validation failures.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
aponcedeleonch: Thank you\!

cmd/thv-operator/controllers/virtualmcpserver_embedding.go
aponcedeleonch: Done! `isEmbeddingServerReady` now returns `(*string, error)` exactly as suggested:

- `nil, nil` → no embedding server configured (no gate) or not ready yet
- `&url, nil` → ready, URL available
- `nil, err` → transient error (triggers requeue)

The caller checks `esURL == nil && vmcp.Spec.EmbeddingServerRef != nil` to detect the "configured but not ready" case that needs requeue.

cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
aponcedeleonch: Thank you\!

cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
aponcedeleonch: Implemented! This is now handled in two stages:

1. **Validation** (`validateEmbeddingServer`): When `EmbeddingServerRef` is set but optimizer is nil, auto-populates `Optimizer = &OptimizerConfig{}` with defaults
2. **Controller** (`populateOptimizerEmbeddingService`): Resolves the EmbeddingServer URL and writes it into `config.Optimizer.EmbeddingService`

An `OptimizerAutoConfigured` Kubernetes event is emitted so the user is aware that defaults were applied. Tests cover this flow end-to-end.

jerm-dro (approved):
LGTM! 

Suggestion: add some integration tests to validate the integration of all this code. Examples are here: https://github.com/stacklok/toolhive/blob/issue-3733/pr2-real-tei-client/cmd/thv-operator/test-integration/embedding-server/embeddingserver_update_test.go

You can do that in a separate PR, if at all.

Issue #3396, opened 2026-01-22
Webhook Middleware Phase 1: Core webhook package (types, HTTP client, HMAC signing)


JAORMX:
## Overview

Implement the foundational webhook package for the Dynamic Webhook Middleware feature. This phase creates the core types, HTTP client, and HMAC signing functionality that will be used by both validating and mutating webhook middleware.

**RFC**: https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0017-dynamic-webhook-middleware.md

## Files to Create

| File | Purpose |
|------|---------|
| `pkg/webhook/types.go` | Core types: `WebhookType`, `FailurePolicy`, `WebhookConfig`, request/response structs |
| `pkg/webhook/client.go` | HTTP client for calling webhooks with TLS/timeout support |
| `pkg/webhook/signing.go` | HMAC-SHA256 payload signing (`X-ToolHive-Signature` header) |
| `pkg/webhook/errors.go` | Error types: timeout, network, validation errors |

## Key Types

```go
// pkg/webhook/types.go
type WebhookType string
const (
    WebhookTypeValidating WebhookType = "validating"
    WebhookTypeMutating   WebhookType = "mutating"
)

type FailurePolicy string
const (
    FailurePolicyFail   FailurePolicy = "fail"   // fail-closed
    FailurePolicyIgnore FailurePolicy = "ignore" // fail-open
)

type WebhookConfig struct {
    Name          string        `json:"name"`
    URL           string        `json:"url"`
    Timeout       time.Duration `json:"timeout"`
    FailurePolicy FailurePolicy `json:"failure_policy"`
    TLSConfig     *TLSConfig    `json:"tls_config,omitempty"`
    HMACSecretRef string        `json:"hmac_secret_ref,omitempty"`
}

type WebhookRequest struct {
    Version    string           `json:"version"` // "v0.1.0"
    UID        string           `json:"uid"`
    Timestamp  time.Time        `json:"timestamp"`
    Principal  *Principal       `json:"principal"`
    MCPRequest json.RawMessage  `json:"mcp_request"`
    Context    *RequestContext  `json:"context"`
}

type Principal struct {
    Sub    string            `json:"sub"`
    Email  string            `json:"email,omitempty"`
    Name   string            `json:"name,omitempty"`
    Groups []string          `json:"groups,omitempty"`
    Claims map[string]string `json:"claims,omitempty"`
}

type RequestContext struct {
[...]

PR #3840, opened 2026-02-16
Add foundational webhook package (Phase 1 of dynamic webhook middleware)

Sanskarzz:
Fix: #3396 

**Summary** 

Implements Phase 1 of [RFC THV-0017](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0017-dynamic-webhook-middleware.md), creating the core pkg/webhook/ package with types, HTTP client, HMAC signing, and error handling.

ToolHive currently has no mechanism for external policy enforcement or request transformation. This PR lays the foundation for the dynamic webhook middleware system that will allow operators to configure validating and mutating webhooks for MCP request interception.

**Changes**

File | Description
-- | --
types.go | Core types: Type, FailurePolicy, Config (with validation), Request/Response/MutatingResponse, TLSConfig, Principal, RequestContext
errors.go | Error hierarchy: WebhookError base type, TimeoutError, NetworkError, InvalidResponseError — all with Unwrap() for errors.Is/errors.As
signing.go | HMAC-SHA256 payload signing (SignPayload) and constant-time verification (VerifySignature)
client.go | HTTP client with TLS/mTLS support, configurable timeouts, HMAC header injection, response size limiting (1MB), and error classification
types_test.go | Config validation tests (12 cases)
errors_test.go | Error type, message, and unwrap chain tests
signing_test.go | Signing round-trip, tamper detection, malformed signature, and determinism tests
client_test.go | HTTP client tests using httptest: validating/mutating responses, HMAC headers, timeouts, response size limits, content type

**Design decisions**

- No stutter naming: Types are webhook.Config, webhook.Type, etc. per project lint rules (not webhook.WebhookConfig)
- Follows existing patterns: Error types follow pkg/config/errors.go (struct with Unwrap()), HTTP client follows pkg/networking/http_client.go (transport setup, TLS config, timeouts)
- Uses errors.As for error type inspection (not type assertions), matching project convention
- Signature format: sha256=hex(HMAC-SHA256(secret, "timestamp.payload")) with X-ToolHive-Signature and X-ToolHive-Timestamp headers per RFC spec

**Testing**

- 22 tests pass with -race flag
- 0 lint issues from golangci-lint

## Large PR Justification

New foundational package (`pkg/webhook/`) — the 4 source files are tightly coupled (client depends on types, errors, signing) and form the minimum viable unit for Phase 1. Over 65% of lines are tests.


pkg/webhook/client.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// buildTransport creates an http.RoundTripper with the specified TLS configuration,
+// wrapped in a ValidatingTransport for security.
+func buildTransport(tlsCfg *TLSConfig) (http.RoundTripper, error) {
+	transport := &http.Transport{
+		TLSHandshakeTimeout:   10 * time.Second,
+		ResponseHeaderTimeout: 10 * time.Second,
+		MaxIdleConns:          100,
+		MaxIdleConnsPerHost:   10,
```
JAORMX, line 165: When `tlsCfg == nil`, this returns a bare `http.Transport` without the `ValidatingTransport` wrapper. That means the SSRF protections (private IP blocking, etc.) only kick in when TLS is explicitly configured.

The `#nosec G704` on line 114 claims coverage via `ValidatingTransport`, but that's only true in the TLS branch.

The fix is to always wrap in `ValidatingTransport`:

```go
func buildTransport(tlsCfg *TLSConfig) (http.RoundTripper, error) {
	transport := &http.Transport{
		TLSHandshakeTimeout:   10 * time.Second,
		ResponseHeaderTimeout: 10 * time.Second,
		MaxIdleConns:          100,
		MaxIdleConnsPerHost:   10,
		IdleConnTimeout:       90 * time.Second,
	}

	if tlsCfg != nil {
		// ... apply TLS config to transport ...
	}

	allowHTTP := tlsCfg != nil && tlsCfg.InsecureSkipVerify
	return &networking.ValidatingTransport{
		Transport:         transport,
		InsecureAllowHTTP: allowHTTP,
	}, nil
}
```

Note that the RFC says "TLS/HTTPS required," but in practice folks will want to hit `http://my-service.my-namespace.svc.cluster.local` for in-cluster webhooks where TLS isn't configured. So we probably want `InsecureSkipVerify` (or a new `AllowHTTP` field) to gate HTTP access explicitly, rather than a hard HTTPS requirement. That way the default is secure, but operators running in-cluster can opt in intentionally.

pkg/webhook/types.go
```diff
@@ -0,0 +1,183 @@
...
+	if c.Name == "" {
+		return fmt.Errorf("webhook name is required")
+	}
+	if c.URL == "" {
+		return fmt.Errorf("webhook URL is required")
+	}
+	if _, err := url.ParseRequestURI(c.URL); err != nil {
+		return fmt.Errorf("webhook URL is invalid: %w", err)
+	}
+	if c.FailurePolicy != FailurePolicyFail && c.FailurePolicy != FailurePolicyIgnore {
```
JAORMX, line 93: Related to the transport comment: `url.ParseRequestURI` happily accepts `http://` URLs. It'd be nice to validate the scheme here so operators get a clear error at config time rather than a confusing runtime rejection from the transport.

That said, we need to account for in-cluster use cases where HTTP to a cluster-local service is reasonable. So the check could be gated on whether `InsecureSkipVerify` (or a future `AllowHTTP` flag) is set:

```go
parsed, err := url.ParseRequestURI(c.URL)
if err != nil {
    return fmt.Errorf("webhook URL is invalid: %w", err)
}
if parsed.Scheme != "https" && (c.TLSConfig == nil || !c.TLSConfig.InsecureSkipVerify) {
    return fmt.Errorf("webhook URL must use HTTPS (set insecure_skip_verify for development/in-cluster HTTP)")
}
```

pkg/webhook/types.go
```diff
@@ -0,0 +1,183 @@
...
+	URL string `json:"url"`
+	// Timeout is the maximum time to wait for a webhook response.
+	Timeout time.Duration `json:"timeout"`
+	// FailurePolicy determines behavior when the webhook call fails.
+	FailurePolicy FailurePolicy `json:"failure_policy"`
+	// TLSConfig holds optional TLS configuration (CA bundles, client certs).
+	TLSConfig *TLSConfig `json:"tls_config,omitempty"`
+	// HMACSecretRef is an optional reference to an HMAC secret for payload signing.
+	HMACSecretRef string `json:"hmac_secret_ref,omitempty"`
+}
```
JAORMX, line 74: Note that `time.Duration` serializes to JSON as nanoseconds (`5000000000` for 5s). The RFC shows `timeout: 5s` in YAML config, so a human-written duration string won't unmarshal into this field directly.

Not a blocker for this PR, but worth deciding on before the config integration lands. Options: custom type with `UnmarshalJSON`/`UnmarshalYAML`, a `string` field parsed with `time.ParseDuration`, or documenting that the config layer handles the conversion.

pkg/webhook/client.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/stacklok/toolhive/pkg/networking"
+)
+
+// Client is an HTTP client for calling webhook endpoints.
+type Client struct {
+	httpClient  *http.Client
+	config      Config
+	hmacSecret  []byte
+	webhookType Type
+}
```
JAORMX, line 32: `webhookType` is stored but never read. There's no dispatch logic based on type -- callers just pick `Call` vs `CallMutating` directly. Either remove the field or, if the plan is to add a `Send` method that dispatches based on type, leave a TODO.

pkg/webhook/client.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	transport.TLSClientConfig = tlsConfig
+	return &networking.ValidatingTransport{
+		Transport:         transport,
+		InsecureAllowHTTP: false,
+	}, nil
+}
+
+// classifyError examines an HTTP client error and returns an appropriately
```
JAORMX: If a parent context is cancelled or its deadline exceeds, the error may be `context.DeadlineExceeded` or `context.Canceled` without wrapping a `net.Error`. Those would fall through to `NetworkError` when they're really timeouts/cancellations.

```go
if errors.Is(err, context.DeadlineExceeded) {
    return NewTimeoutError(webhookName, err)
}
```

pkg/webhook/client.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return nil, NewNetworkError(c.config.Name, fmt.Errorf("failed to read response body: %w", err))
+	}
+	if int64(len(respBody)) > MaxResponseSize {
+		return nil, NewInvalidResponseError(c.config.Name,
+			fmt.Errorf("response body exceeds maximum size of %d bytes", MaxResponseSize))
+	}
+
+	// 5xx errors indicate webhook operational failures.
+	if resp.StatusCode >= http.StatusInternalServerError {
```
JAORMX, line 141: The RFC's failure mode table treats HTTP 422 specially -- it's "Deny (422)" regardless of failure policy. With the current code, 422 gets lumped with all other non-200/non-5xx as `InvalidResponseError`, so the middleware layer won't be able to distinguish "422 from webhook = always deny" from "400 bad request = follow failure policy."

Consider adding the HTTP status code to the error (e.g., a `StatusCode` field on `InvalidResponseError`), or a separate error path for status codes that should bypass failure policy.

pkg/webhook/types.go
```diff
@@ -0,0 +1,183 @@
...
+	Context *RequestContext `json:"context"`
+}
+
+// Principal contains the authenticated user's identity information.
+type Principal struct {
+	// Sub is the subject identifier (user ID).
+	Sub string `json:"sub"`
+	// Email is the user's email address.
+	Email string `json:"email,omitempty"`
+	// Name is the user's display name.
```
JAORMX, line 133: JWT claims can carry non-string values (arrays, numbers, booleans). `map[string]string` works for the RFC examples but might be too restrictive for real-world JWTs. Consider `map[string]any` to avoid lossy conversion when wiring this up to actual JWT claims downstream.

pkg/webhook/signing.go
```diff
@@ -0,0 +1,56 @@
...
+	msg := fmt.Sprintf("%d.", timestamp)
+	mac.Write([]byte(msg))
+	mac.Write(payload)
+	return signaturePrefix + hex.EncodeToString(mac.Sum(nil))
+}
+
+// VerifySignature verifies an HMAC-SHA256 signature against the given timestamp
+// and payload. The signature should be in the format "sha256=<hex-encoded-signature>".
+// Comparison is done in constant time to prevent timing attacks.
+func VerifySignature(secret []byte, timestamp int64, payload []byte, signature string) bool {
```
JAORMX, line 44: Since this is exported, webhook server authors will likely reach for it. Worth adding a doc note that callers are responsible for checking timestamp freshness (e.g., reject if older than 5 minutes) to prevent replay attacks.

```go
// Callers should independently verify that the timestamp is recent
// (e.g., within 5 minutes) to prevent replay attacks.
```

pkg/webhook/client_test.go
```diff
@@ -0,0 +1,741 @@
...
+				w.Write([]byte("not json"))
+			},
+			expectError: true,
+			errorType:   &InvalidResponseError{},
+		},
+		{
+			name: "non-200 non-5xx response",
+			serverHandler: func(w http.ResponseWriter, _ *http.Request) {
+				w.WriteHeader(http.StatusBadRequest)
+				w.Write([]byte("bad request"))
```
JAORMX, line 160: Project convention: tests should use `t.Context()` instead of `context.Background()`. This applies to all the `client.Call(context.Background(), req)` calls throughout this file (~10 instances). `t.Context()` gets automatically cancelled when the test ends, which helps catch goroutine leaks.

pkg/webhook/client.go
Sanskarzz: Done

pkg/webhook/types.go
Sanskarzz: Done

pkg/webhook/client.go
Sanskarzz: Added todo.

pkg/webhook/client.go
Sanskarzz: Added

pkg/webhook/client.go
Sanskarzz: Added

pkg/webhook/types.go
Sanskarzz: Added

pkg/webhook/signing.go
Sanskarzz: Done

pkg/webhook/client_test.go
Sanskarzz: Resolved

PR #3843, opened 2026-02-17
Relocate monitor and errors to runtime package

JAORMX:
## Summary

- Break the coupling between `factory.go`/`transport/` and the `docker` package — they imported it only for generic error types and the container monitor, neither of which is Docker-specific
- With the extensible runtime registry (#3832), new runtimes would otherwise need to depend on `docker/` for shared monitoring or duplicate the code; placing the monitor in `runtime/` makes it available to any runtime implementation
- Add test coverage for the monitor polling loop, which was previously untested (container exit, removal, restart detection, context cancellation, transient error recovery)

Deprecated re-export aliases in `docker/errors.go` keep `docker/client.go` compiling unchanged — migration of internal Docker references can happen incrementally.

## Test plan

- [x] `task build` — no import cycles
- [x] `task lint` — 0 issues
- [x] All affected package tests pass with `-race`
- [x] `factory.go`, `transport/http.go`, `transport/stdio.go` no longer import `docker`

### DIFF
--- a/pkg/container/runtime/errors.go
+++ b/pkg/container/runtime/errors.go
@@ -68,5 +68,5 @@ func NewContainerError(err error, containerID, message string) *ContainerError {
 
 // IsContainerNotFound checks if the error is a container not found error
 func IsContainerNotFound(err error) bool {
-	return errors.Is(err, ErrContainerNotFound) || (err != nil && err.Error() == ErrContainerNotFound.Error())
+	return errors.Is(err, ErrContainerNotFound)
 }


pkg/container/runtime/errors.go
```diff
@@ -0,0 +1,72 @@
...
+	return &ContainerError{
+		Err:         err,
+		ContainerID: containerID,
+		Message:     message,
+	}
+}
+
+// IsContainerNotFound checks if the error is a container not found error
+func IsContainerNotFound(err error) bool {
+	return errors.Is(err, ErrContainerNotFound) || (err != nil && err.Error() == ErrContainerNotFound.Error())
```
jhrozek: so we check both the sentinel and the string? Why?

JAORMX: Good point — the string comparison fallback was unnecessary since `errors.Is()` already handles unwrapping via `ContainerError.Unwrap()`. Removed in 6915cc2e.

jhrozek (approved):

PR #3844, opened 2026-02-17
Add real TEI embedding client for semantic tool discovery

aponcedeleonch:
Closes: https://github.com/stacklok/toolhive/issues/3733
Depends on: https://github.com/stacklok/toolhive/pull/3839

Introduce a TEIClient that calls the HuggingFace Text Embeddings Inference HTTP API for vector embeddings, replacing the need for fake/deterministic embeddings in production. The store factory now selects between a real TEI client (when a service URL is configured), a fake client (for testing), or FTS5-only search.

Wire the full OptimizerConfig through the server instead of a boolean flag, so the embedding service URL and timeout are available when creating the store and client.

### DIFF
--- a/pkg/vmcp/optimizer/internal/similarity/tei_client.go
+++ b/pkg/vmcp/optimizer/internal/similarity/tei_client.go
@@ -11,6 +11,9 @@ import (
 	"io"
 	"net/http"
 	"time"
+
+	vmcpconfig "github.com/stacklok/toolhive/pkg/vmcp/config"
+	"github.com/stacklok/toolhive/pkg/vmcp/optimizer/internal/types"
 )
 
 const (
@@ -21,37 +24,35 @@ const (
 	embedPath = "/embed"
 )
 
-// TEIClientConfig configures the TEI embedding client.
-type TEIClientConfig struct {
-	// BaseURL is the full URL of the TEI embedding service
-	// (e.g. "http://my-embedding.ml.svc.cluster.local:8080").
-	// Required.
-	BaseURL string
-
-	// Timeout is the HTTP request timeout. Defaults to 30s.
-	Timeout time.Duration
+// NewEmbeddingClient creates an EmbeddingClient from the given optimizer
+// configuration. It returns (nil, nil) if cfg is nil or no embedding service
+// URL is configured, meaning semantic search will be disabled.
+func NewEmbeddingClient(cfg *vmcpconfig.OptimizerConfig) (types.EmbeddingClient, error) {
+	if cfg == nil || cfg.EmbeddingService == "" {
+		return nil, nil
+	}
+	return newTEIClient(cfg.EmbeddingService, time.Duration(cfg.EmbeddingServiceTimeout))
 }
 
-// TEIClient implements types.EmbeddingClient by calling the HuggingFace
+// teiClient implements types.EmbeddingClient by calling the HuggingFace
 // Text Embeddings Inference (TEI) HTTP API.
-type TEIClient struct {
+type teiClient struct {
 	baseURL    string
 	httpClient *http.Client
 }
 
-// NewTEIClient creates a new TEI embedding client that calls the specified endpoint.
-func NewTEIClient(cfg TEIClientConfig) (*TEIClient, error) {
-	if cfg.BaseURL == "" {
+// newTEIClient creates a new TEI embedding client that calls the specified endpoint.
+func newTEIClient(baseURL string, timeout time.Duration) (*teiClient, error) {
+	if baseURL == "" {
 		return nil, fmt.Errorf("TEI BaseURL is required")
 	}
 
-	timeout := cfg.Timeout
 	if timeout == 0 {
 		timeout = defaultTimeout
 	}
 
-	return &TEIClient{
-		baseURL: cfg.BaseURL,
+	return &teiClient{
+		baseURL: baseURL,
 		httpClient: &http.Client{
 			Timeout: timeout,
 		},
@@ -60,12 +61,16 @@ func NewTEIClient(cfg TEIClientConfig) (*TEIClient, error) {
 
 // embedRequest is the JSON body sent to the TEI /embed endpoint.
 type embedRequest struct {
-	Inputs   []string `json:"inputs"`
-	Truncate bool     `json:"truncate"`
+	Inputs []string `json:"inputs"`
+	// Truncate tells the TEI server to silently truncate input texts that
+	// exceed the model's maximum token length instead of returning an error.
+	// We always set this to true because tool descriptions may exceed model
+	// limits and we prefer embedding a truncated description over a request failure.
+	Truncate bool `json:"truncate"`
 }
 
 // Embed returns a vector embedding for the given text.
-func (c *TEIClient) Embed(ctx context.Context, text string) ([]float32, error) {
+func (c *teiClient) Embed(ctx context.Context, text string) ([]float32, error) {
 	results, err := c.EmbedBatch(ctx, []string{text})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


aponcedeleonch:
The PR expects the `EmbeddingServer` to be ready and provided. Such work is introduced in #3839 . It's better to wait for that to get merged and then we can proceed with this one. Although the code on this shouldn't change a great deal. Will mark this PR as draft for the moment to not merge it accidentally

pkg/vmcp/optimizer/internal/similarity/tei_client.go
```diff
@@ -0,0 +1,127 @@
...
+// TEIClientConfig configures the TEI embedding client.
+type TEIClientConfig struct {
+	// BaseURL is the full URL of the TEI embedding service
+	// (e.g. "http://my-embedding.ml.svc.cluster.local:8080").
+	// Required.
+	BaseURL string
+
+	// Timeout is the HTTP request timeout. Defaults to 30s.
+	Timeout time.Duration
+}
```
jerm-dro, line 35: Suggestion: delete. This contains the same information as the `config.OptimizerConfig`, so either pass that through to `NewTEIClient` or pass the two args, url and timeout, without a struct.

pkg/vmcp/optimizer/internal/similarity/tei_client.go
```diff
@@ -0,0 +1,127 @@
...
+
+// TEIClient implements types.EmbeddingClient by calling the HuggingFace
+// Text Embeddings Inference (TEI) HTTP API.
+type TEIClient struct {
+	baseURL    string
+	httpClient *http.Client
+}
+
+// NewTEIClient creates a new TEI embedding client that calls the specified endpoint.
+func NewTEIClient(cfg TEIClientConfig) (*TEIClient, error) {
```
jerm-dro: Can we make `TEIClient` private? 

I think we can expose fewer public details from `internal/similarity` by having one `NewEmbeddingClient(optimizerConfig) (types.EmbeddingClient, error)` function. Then, `store.go` doesn't have to be aware about what exact client is being constructed, it's just an implementation detail of the package.

pkg/vmcp/optimizer/internal/similarity/tei_client.go
```diff
@@ -0,0 +1,127 @@
...
+		httpClient: &http.Client{
+			Timeout: timeout,
+		},
+	}, nil
+}
+
+// embedRequest is the JSON body sent to the TEI /embed endpoint.
+type embedRequest struct {
+	Inputs   []string `json:"inputs"`
+	Truncate bool     `json:"truncate"`
```
jerm-dro: nit: can you comment what truncate does?

pkg/vmcp/optimizer/store.go
```diff
@@ -98,17 +100,43 @@ func (s *InMemoryToolStore) Search(_ context.Context, query string, allowedTools
...
+			embClient, err = similarity.NewTEIClient(similarity.TEIClientConfig{
+				BaseURL: cfg.EmbeddingServiceURL,
+				Timeout: cfg.EmbeddingServiceTimeout,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to create TEI embedding client: %w", err)
+			}
+		case cfg.EmbeddingDimension > 0:
+			embClient = similarity.NewFakeEmbeddingClient(cfg.EmbeddingDimension)
+		}
```
jerm-dro: mentioned above, but I think we should push the client construction logic into `internal/similarity`

jerm-dro (changes requested):

pkg/vmcp/optimizer/internal/similarity/tei_client.go
aponcedeleonch: Done. Deleted `TEIClientConfig` and now pass `*config.OptimizerConfig` directly to a new `NewEmbeddingClient(cfg)` factory in the similarity package. The internal `newTEIClient` takes `(baseURL string, timeout time.Duration)` extracted from the config.

pkg/vmcp/optimizer/internal/similarity/tei_client.go
aponcedeleonch: Done. Renamed `TEIClient` → `teiClient` (private). The package now exposes a single `NewEmbeddingClient(cfg *config.OptimizerConfig) (types.EmbeddingClient, error)` factory. Callers only interact with the `EmbeddingClient` interface — the concrete type is an implementation detail.

pkg/vmcp/optimizer/internal/similarity/tei_client.go
aponcedeleonch: Done. Added a comment explaining that `truncate` tells the TEI server to silently truncate inputs exceeding the model's max token length instead of returning an error. We always set it to `true` because tool descriptions may exceed model limits and we prefer embedding a truncated description over a request failure.

pkg/vmcp/optimizer/store.go
aponcedeleonch: Done. Client construction is now fully inside `internal/similarity` via `NewEmbeddingClient(cfg)`. The `store.go` layer just re-exports it and passes the pre-built `EmbeddingClient` through. Also removed `SQLiteStoreConfig` entirely — `NewSQLiteToolStore` now takes the `EmbeddingClient` directly since it was the only field left.

pkg/vmcp/config/config.go
```diff
@@ -745,9 +745,15 @@ type OptimizerConfig struct {
 	// EmbeddingService is the full base URL of the embedding service endpoint
 	// (e.g., http://my-embedding.default.svc.cluster.local:8080) for semantic
 	// tool discovery. Auto-populated by the operator from the EmbeddingServer
-	// Status.URL (inline or referenced). Do not set manually.
+	// Status.URL. Do not set manually.
```
jerm-dro, line 748: Suggestion: elaborate. "In a k8s environment, it is more convenient to use the EmbeddingServerRef ..."

pkg/vmcp/optimizer/internal/similarity/tei_client.go
```diff
@@ -0,0 +1,132 @@
...
+	}, nil
+}
+
+// embedRequest is the JSON body sent to the TEI /embed endpoint.
+type embedRequest struct {
+	Inputs []string `json:"inputs"`
+	// Truncate tells the TEI server to silently truncate input texts that
+	// exceed the model's maximum token length instead of returning an error.
+	// We always set this to true because tool descriptions may exceed model
+	// limits and we prefer embedding a truncated description over a request failure.
```
jerm-dro, line 68: nice explanation

pkg/vmcp/optimizer/store.go
```diff
@@ -20,6 +21,18 @@ import (
...
+// It is defined in the internal/types package and aliased here so that
+// external consumers can reference the type.
+type EmbeddingClient = types.EmbeddingClient
+
+// NewEmbeddingClient creates an EmbeddingClient from the given optimizer
+// configuration. Returns (nil, nil) if cfg is nil or no embedding service
+// is configured, meaning only FTS5 full-text search will be used.
+func NewEmbeddingClient(cfg *vmcpconfig.OptimizerConfig) (EmbeddingClient, error) {
+	return similarity.NewEmbeddingClient(cfg)
+}
```
jerm-dro, line 34: Suggestion: package up all these constructors and interfaces into a single: `func NewOptimizerFactory(cfg) OptimizerFactory`. The only need to expose these types and constructors is so the server.go can construct the objects, but the server.go really only does this to produce an `OptimizerFactory`.

aponcedeleonch: yes, was planning to do just that! Will take care of it in #3896

pkg/vmcp/optimizer/internal/similarity/tei_client.go
aponcedeleonch: thanks!

pkg/vmcp/config/config.go
aponcedeleonch: Sure, good call, will elaborate in a follow-up PR

PR #3851, opened 2026-02-17
Add per-skill concurrency protection to install/uninstall

### LINKED ISSUE #3845: Add concurrency protection to skill install/uninstall operations
## Context

Raised during review of #3835 — the skill service methods (`Install`, `Uninstall`) are called from REST API handlers but have no serialization. Two concurrent requests operating on the same skill could race on both filesystem operations and store writes.

## What needs to happen

Add a mutex or file lock in the `skillsvc.service` layer to serialize operations on the same skill name. This would prevent:

- Two concurrent installs of the same skill racing on extraction and store creation
- An install and uninstall of the same skill running simultaneously
- Two upgrades with different digests overwriting each other

## Notes

- In practice the risk is low today (single-user CLI tool), but the API surface is exposed via REST so it's worth hardening.
- A per-skill lock (keyed by name+scope) would be more granular than a global mutex, but a global mutex is simpler and likely sufficient given the expected concurrency level.

JAORMX:
## Summary

- Adds a per-skill mutex (keyed by `scope/name/projectRoot`) to serialize `Install` and `Uninstall` operations on the same skill, preventing races from concurrent REST API requests
- Different skills proceed concurrently without blocking each other
- Includes a concurrency test that uses atomic counters to verify mutual exclusion (not just crash-freedom)

Closes #3845

## Test plan

- [x] `task test` passes with `-race` enabled
- [x] New `TestConcurrentInstallAndUninstall` verifies mutual exclusion per skill using atomic counters inside mock `DoAndReturn`
- [ ] Review that the two-phase locking pattern in `skillLock.lock()` is sound (global mutex for map access, per-key mutex for operations)

### DIFF
--- a/pkg/skills/skillsvc/skillsvc.go
+++ b/pkg/skills/skillsvc/skillsvc.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
@@ -102,6 +103,13 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
 
 	scope := defaultScope(opts.Scope)
 
+	// Canonicalize the project root so that equivalent paths
+	// (e.g. trailing slash, ".." segments) produce the same lock key
+	// and DB record.
+	if opts.ProjectRoot != "" {
+		opts.ProjectRoot = filepath.Clean(opts.ProjectRoot)
+	}
+
 	unlock := s.locks.lock(opts.Name, scope, opts.ProjectRoot)
 	defer unlock()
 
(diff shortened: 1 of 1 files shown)


pkg/skills/skillsvc/skillsvc.go
```diff
@@ -34,16 +35,46 @@ func WithInstaller(inst skills.Installer) Option {
...
+	// locks holds per-key mutexes. INVARIANT: entries must never be deleted
+	// from this map. The two-phase lock() method depends on pointers remaining
+	// valid after the global mutex is released. See lock() for details.
+	locks map[string]*sync.Mutex
+}
+
+// lock acquires a per-skill mutex and returns a function that releases it.
+func (sl *skillLock) lock(name string, scope skills.Scope, projectRoot string) func() {
+	sl.mu.Lock()
+	key := string(scope) + "/" + name + "/" + projectRoot
```
jhrozek, line 53: did we call filepath.Clean on the project root earlier? If not, I'd suggest doing it otherwise we risk creating multiple keys for the same path

JAORMX: Good catch\! It wasn't cleaned anywhere upstream. Added `filepath.Clean` on `projectRoot` in both `Install` and `Uninstall` before the lock is acquired, so equivalent paths (trailing slash, "..", etc.) produce the same lock key and DB record.

Issue #3852, opened 2026-02-17
empty jwks_uri breaks cursor


jhrozek:
This was a cascading issue...

- our OAuth PRM was including jwks_uri which is not needed in the first place because this set of keys is meant to  sign the /resource server's/ messages. Note that this is NOT the IDP JWKS URI that's advertised by the IDP itself!
- we did include the IDP value there originally, which was fine, since we don't sign anything by thv
- however, when we switched to lazy OIDC discovery and fetching the JWKS IDP keys lazily, an unintended side-effect was that jwks_uri was set to an empty string on startup
- an empty string is not evaluated by some clients (e.g. cursor).

The proper fix is to just remove jwks_uri from our PRM.

PR #3853, opened 2026-02-17
Remove jwks_uri from Protected Resource Metadata response

jhrozek:
RFC 9728 defines jwks_uri in PRM as the resource server's own signing keys (e.g. FAPI message signing), not the authorization server's keys.

ToolHive's MCP servers don't sign responses, so the field should be absent. Since lazy OIDC discovery (5fd6adb), the JWKS URL was empty at handler construction time, producing "jwks_uri": "" which fails the MCP TypeScript SDK's Zod validation and causes PRM to be silently discarded by clients like Cursor.

Fixes: #3852

### DIFF
--- a/pkg/auth/token_test.go
+++ b/pkg/auth/token_test.go
@@ -1217,7 +1217,8 @@ func TestNewAuthInfoHandler(t *testing.T) {
 
 			// Check response body if expected
 			if tc.expectBody {
-				// Read raw body to verify jwks_uri is absent from JSON
+				// Regression test: verify jwks_uri is absent from the JSON response.
+				// See https://github.com/stacklok/toolhive/issues/3852
 				bodyBytes := rec.Body.Bytes()
 				var rawMap map[string]any
 				if err := json.Unmarshal(bodyBytes, &rawMap); err != nil {


pkg/auth/token_test.go
```diff
@@ -1223,8 +1217,18 @@ func TestNewAuthInfoHandler(t *testing.T) {
...
 			if tc.expectBody {
+				// Read raw body to verify jwks_uri is absent from JSON
+				bodyBytes := rec.Body.Bytes()
+				var rawMap map[string]any
+				if err := json.Unmarshal(bodyBytes, &rawMap); err != nil {
+					t.Fatalf("Failed to decode raw response body: %v", err)
+				}
+				if _, exists := rawMap["jwks_uri"]; exists {
+					t.Errorf("jwks_uri must not appear in the PRM response (RFC 9728 §3.2)")
+				}
```
jerm-dro, line 1229: suggestion, non-blocking: add a link to the issue explaining this is to detect a regression.

rdimitrov (approved):

JAORMX (approved):

Issue #3865, opened 2026-02-18
[vMCP] Define Session and SessionFactory interfaces (Phase 1)


yrobla:
Introduce the core interfaces and default implementations for the new session-scoped architecture described in THV-0038. This is purely additive — no changes to existing flows or existing tests.

**New files**:
- `pkg/vmcp/session/session.go` — `Session` interface (domain object: `CallTool`, `ReadResource`, `GetPrompt`, `Tools`, `Resources`, `Prompts`, `BackendSessions`, `Close`)
- `pkg/vmcp/session/factory.go` — `SessionFactory` interface and `defaultSessionFactory`
- `pkg/vmcp/session/default_session.go` — `defaultSession` embedding `transportsession.Session` for metadata, owning backend clients in an internal map, tracking backend session IDs

**Key requirements**:
- Factory initializes clients per backend in parallel (`errgroup`, bounded concurrency configurable via `max_backend_init_concurrency`, default 10, per-session not global)
- Per-backend timeout so a slow backend does not block session creation
- Partial initialization: log warnings for failed backends, continue with successful ones
- Thread safety: `sync.RWMutex` protecting client map; `RLock` released before network I/O; `sync.WaitGroup` in-flight counter so `Close()` waits for in-flight calls before tearing down clients
- `Close()` closes all owned backend clients

### Acceptance Criteria

- [ ] `Session` interface is defined in `pkg/vmcp/session/session.go` with all required methods
- [ ] `SessionFactory` interface and `defaultSessionFactory` are defined and implemented
- [ ] `defaultSession` implements both `transportsession.Session` and the vMCP `Session` interface
- [ ] Backend clients are initialized in parallel with bounded concurrency
- [ ] A slow or failing backend does not prevent the session from being created with the remaining backends
- [ ] All reads from the client map are protected by `RLock` and locks are released before network I/O
- [ ] `Close()` waits for all in-flight operations to complete before closing backend clients
- [ ] No changes to existing code or tests
- [ ] Unit tests cover: `CallTool`, `ReadResource`, `Close`, parallel init with timeouts, partial failure, interface composition

PR #3882, opened 2026-02-18
Add Session and SessionFactory interfaces for vMCP (Phase 1)

yrobla:
Implements issue #3865 and RFC THV-0038 (session-scoped client lifecycle).
  Phase 1 introduces pkg/vmcp/session/, which owns the lifetime of persistent                                                              
  MCP connections to backend servers for the duration of a client session.
                                                                                                                                           
  session.go: Session interface extending transportsession.Session, with the                                                               
  dual-layer state model (serialisable metadata vs. non-serialisable runtime
  state) and sticky-session constraint documented.

  factory.go: SessionFactory and defaultSessionFactory. Backends are
  initialised in parallel (errgroup, capped at 10 concurrent) with partial
  failure tolerance — failed backends are skipped so the session proceeds
  with the remaining healthy ones.

  default_session.go: Thread-safe dispatch using RWMutex + WaitGroup. The
  critical invariant is wg.Add(1) under RLock, which prevents a race between
  in-flight operations and Close() draining the group.

  connector.go: mcpConnectedBackend wrapping a persistent mcp-go client per
  backend, with a layered HTTP transport chain (auth → identity propagation)
  and an MCP Initialize handshake on connect.

  Also extracted ConvertMCPContent, ConvertMCPContents, and
  ConcatenateResourceContents into pkg/vmcp/conversion to eliminate
  duplication between the client and session packages. Two bugs fixed in
  pkg/vmcp/client in the process: io.LimitReader was incorrectly applied to
  SSE (a single long-lived response body, not per-event), and tool input
  schemas were being copied field-by-field rather than via a JSON round-trip,
  silently dropping additionalProperties and future SDK fields.
  
## Large PR Justification

This is a complete PR with isolated functionality an related tests. Cannot be split.
  
  Closes: #3865

### DIFF
--- a/pkg/vmcp/session/default_session.go
+++ b/pkg/vmcp/session/default_session.go
@@ -12,13 +12,14 @@ import (
 
 	transportsession "github.com/stacklok/toolhive/pkg/transport/session"
 	"github.com/stacklok/toolhive/pkg/vmcp"
+	"github.com/stacklok/toolhive/pkg/vmcp/session/internal/backend"
 )
 
-// Compile-time assertions: defaultSession must implement both interfaces.
-var _ Session = (*defaultSession)(nil)
-var _ transportsession.Session = (*defaultSession)(nil)
+// Compile-time assertions: defaultMultiSession must implement both interfaces.
+var _ MultiSession = (*defaultMultiSession)(nil)
+var _ transportsession.Session = (*defaultMultiSession)(nil)
 
-// Sentinel errors returned by defaultSession methods.
+// Sentinel errors returned by defaultMultiSession methods.
 var (
 	// ErrSessionClosed is returned when an operation is attempted on a closed session.
 	ErrSessionClosed = errors.New("session is closed")
@@ -32,52 +33,22 @@ var (
 	// ErrPromptNotFound is returned when the requested prompt is not in the routing table.
 	ErrPromptNotFound = errors.New("prompt not found in session routing table")
 
-	// ErrNoBackendClient is returned when there is no client available for a backend.
+	// ErrNoBackendClient is returned when the routing table references a backend
+	// that has no entry in the connections map. This indicates an internal
+	// invariant violation: under normal operation MakeSession always populates
+	// both maps together, so this error should never be seen at runtime.
 	ErrNoBackendClient = errors.New("no client available for backend")
 )
 
-// connectedBackend abstracts a persistent, initialised MCP connection to a
-// single backend server. It is created once per backend during session creation
-// (MakeSession) and reused for the lifetime of the session.
-//
-// Implementations must be safe for concurrent use.
-type connectedBackend interface {
-	// callTool invokes a named tool on this backend.
-	// target is used for capability name translation via GetBackendCapabilityName.
-	callTool(
-		ctx context.Context,
-		target *vmcp.BackendTarget,
-		toolName string,
-		arguments map[string]any,
-		meta map[string]any,
-	) (*vmcp.ToolCallResult, error)
-
-	// readResource reads a resource from this backend.
-	readResource(ctx context.Context, target *vmcp.BackendTarget, uri string) (*vmcp.ResourceReadResult, error)
-
-	// getPrompt retrieves a prompt from this backend.
-	getPrompt(
-		ctx context.Context,
-		target *vmcp.BackendTarget,
-		name string,
-		arguments map[string]any,
-	) (*vmcp.PromptGetResult, error)
-
-	// sessionID returns the backend-assigned session ID (if any).
-	// Returns "" if the backend did not assign a session ID.
-	sessionID() string
-
-	// close closes the underlying transport connection.
-	close() error
-}
-
-// defaultSession is the production Session implementation.
+// defaultMultiSession is the production MultiSession implementation.
 //
 // # Thread-safety model
 //
-// mu guards connections, routingTable, closed, and the wg.Add call. RLock is
-// held only long enough to retrieve state and atomically increment the in-flight
-// counter (wg.Add); it is released before network I/O begins.
+// mu guards connections, closed, and the wg.Add call. RLock is held only
+// long enough to retrieve state and atomically increment the in-flight counter
+// (wg.Add); it is released before network I/O begins.
+// routingTable, tools, resources, and prompts are written once during
+// MakeSession and are read-only thereafter — they do not require lock protection.
 //
 // wg tracks in-flight operations. Close() sets closed=true under write lock,
 // then waits for wg to reach zero before tearing down backend connections.
@@ -87,13 +58,20 @@ type connectedBackend interface {
 //
 // # Lifecycle
 //
-//  1. Created by defaultSessionFactory.MakeSession (Phase 1: purely additive).
+//  1. Created by defaultMultiSessionFactory.MakeSession (Phase 1: purely additive).
 //  2. CallTool / ReadResource / GetPrompt increment wg, perform I/O, decrement wg.
-//  3. Close() sets closed=true, waits for wg, then closes all backend clients.
-type defaultSession struct {
+//  3. Close() sets closed=true, waits for wg, then closes all backend sessions.
+//
+// # Composite tools
+//
+// Composite tools (VirtualMCPCompositeToolDefinition) are out of scope for
+// Phase 1. When they are introduced they will be resolved at a higher layer
+// (e.g. the vMCP router or handler) and injected alongside the backend tool
+// list, rather than being routed through the backend connections held here.
+type defaultMultiSession struct {
 	transportsession.Session // embedded interface — provides ID, Type, timestamps, etc.
 
-	connections     map[string]connectedBackend // backend workload ID → persistent client
+	connections     map[string]backend.Session // backend workload ID → persistent backend session
 	routingTable    *vmcp.RoutingTable
 	tools           []vmcp.Tool
 	resources       []vmcp.Resource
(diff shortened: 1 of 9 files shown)


pkg/vmcp/session/factory.go
```diff
@@ -0,0 +1,299 @@
...
+// must be closed when the session ends. The returned CapabilityList is used
+// to populate the session's routing table and capability lists.
+//
+// On error the factory treats the failure as a partial failure: a warning is
+// logged and the backend is excluded from the session.
+type BackendConnector func(
+	ctx context.Context,
+	target *vmcp.BackendTarget,
+	identity *auth.Identity,
+) (connectedBackend, *vmcp.CapabilityList, error)
```
jerm-dro: Blocker: This type and its usage is complex:
- `BackendConnector` is public
- `BackendConnector` returns a private `connectedBackend`
- `NewSessionFactory` takes a `BackendConnector`
- `NewHTTPBackendConnector` appears to be the only thing that returns a `BackendConnector` and it's defined in this module.

How can we simplify this?

pkg/vmcp/session/factory.go
```diff
@@ -0,0 +1,299 @@
...
+
+	for _, b := range backends {
+		backend := b // capture loop variable
+		g.Go(func() error {
+			if r := f.initOneBackend(gCtx, backend, identity); r != nil {
+				resultsCh <- *r
+			}
+			return nil // always nil: partial init, failures already logged
+		})
+	}
```
jerm-dro: Suggestion: this can be simplified to avoid using a channel and err group:


```go
results := make([]initResults, len(backends))
var wg sync.WaitGroup
wg.Add(len(backends))

for i := range backends {
   go func () {
    defer wg.Done()
    results[i] = f.initOneBackend(...)
   }()
}

wg.Wait()
```


In other golang projects, I've seen things like this implemented generically and added to a `foreach` package. Adding such a capability to the common repo would be a welcome contribution.

pkg/vmcp/session/default_session_test.go
```diff
@@ -0,0 +1,1075 @@
...
+
+// mockConnectedBackend is an in-process connectedBackend for testing.
+type mockConnectedBackend struct {
+	callToolFunc     func(ctx context.Context, target *vmcp.BackendTarget, toolName string, arguments, meta map[string]any) (*vmcp.ToolCallResult, error)
+	readResourceFunc func(ctx context.Context, target *vmcp.BackendTarget, uri string) (*vmcp.ResourceReadResult, error)
+	getPromptFunc    func(ctx context.Context, target *vmcp.BackendTarget, name string, arguments map[string]any) (*vmcp.PromptGetResult, error)
+	sessID           string
+	closeCalled      atomic.Bool
+	closeErr         error
+}
```
jerm-dro: Suggestion: could you setup a real backend MCP server for tests? 

I think there's existing tests for composite tools that spin up in-process servers for testing. The benefit of writing tests with the a real server is that it would exercise the connector bits, which are complex and an important part of the integration.

pkg/vmcp/session/default_session.go
```diff
@@ -0,0 +1,244 @@
...
+		arguments map[string]any,
+	) (*vmcp.PromptGetResult, error)
+
+	// sessionID returns the backend-assigned session ID (if any).
+	// Returns "" if the backend did not assign a session ID.
+	sessionID() string
+
+	// close closes the underlying transport connection.
+	close() error
+}
```
jerm-dro: Suggestion: this looks exactly the same as the `Session` object minus a few methods, which makes sense. It's also a session! Should we call this thing a session and then `Session` becomes:

```go
// MultiSession contains multiple client connection with their own sessions
type MultiSession interface {
Session
BackendSessions() map[string]string
}
```

This would allow us to delete the `connectedBackend` concept, which makes the package conceptually a lot simpler. A vmcp session is just a session of sessions.

jerm-dro: If you go that direction, I'd recommend moving what is now `connectedBackend` to its own package:
`vmcp/session/internal/session` or something to keep the connection code separate from the routing code.

pkg/vmcp/session/default_session.go
```diff
@@ -0,0 +1,244 @@
...
+// and the live connection for the backend that owns it.
+//
+// On success, wg.Add(1) has been called before the lock is released. The
+// caller MUST call wg.Done() (typically via defer) when the I/O completes.
+// On error, wg.Add was never called.
+func (s *defaultSession) lookupBackend(
+	capName string,
+	table map[string]*vmcp.BackendTarget,
+	notFoundErr error,
+) (*vmcp.BackendTarget, connectedBackend, error) {
```
jerm-dro: Is it really necessary to return the backend just to pass it as an argument to `connectedBackend`? This seems like something that can be baked into the `connectedBackend`.

pkg/vmcp/session/default_session.go
```diff
@@ -0,0 +1,244 @@
...
+	ctx context.Context,
+	toolName string,
+	arguments map[string]any,
+	meta map[string]any,
+) (*vmcp.ToolCallResult, error) {
+	target, conn, err := s.lookupBackend(toolName, s.routingTable.Tools, ErrToolNotFound)
+	if err != nil {
+		return nil, err
+	}
+	defer s.wg.Done()
```
jerm-dro, line 167: nit: as currently factored, it's not clear why the caller needs to call this.

I'd suggest returning a `done` function from `lookupBackend` that should be called when the backend is no longer in use. Under the covers, that calls `s.wg.done()`, but the caller doesn't need to know that.

pkg/vmcp/session/default_session.go
```diff
@@ -0,0 +1,244 @@
...
+//  3. Close() sets closed=true, waits for wg, then closes all backend clients.
+type defaultSession struct {
+	transportsession.Session // embedded interface — provides ID, Type, timestamps, etc.
+
+	connections     map[string]connectedBackend // backend workload ID → persistent client
+	routingTable    *vmcp.RoutingTable
+	tools           []vmcp.Tool
+	resources       []vmcp.Resource
+	prompts         []vmcp.Prompt
+	backendSessions map[string]string // backend workload ID → backend-assigned session ID
```
jerm-dro: How do the composite tools fit into this? Will they be added in a later PR?

pkg/vmcp/session/default_session.go
```diff
@@ -0,0 +1,244 @@
...
+
+	mu     sync.RWMutex
+	wg     sync.WaitGroup
```
jerm-dro, line 82: suggestion: rename `wg` to `inFlightRequests`. It's a bit unusual to see `wg` without seeing `go` routines created, but that's happening way up the stack. In flight requests better describes what the count represents.

pkg/vmcp/session/session.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// GetPrompt retrieves the named prompt from the appropriate backend for
+	// this session.
+	//
+	// arguments contains the prompt input parameters.
+	GetPrompt(
+		ctx context.Context,
+		name string,
+		arguments map[string]any,
+	) (*vmcp.PromptGetResult, error)
```
jerm-dro, line 87: Can we move the `Session` interface definition to some where more common so it can be used in `internal/backend` and embedded here? Perhaps `session/types/`?

The motivation: in the short term, we can avoid repeating ourselves. In the long term, I suspect we can consolidate on one type. IIUC, right now, the interfaces are different because we expose different methods so the session may be serialized. On the [RFC](https://github.com/stacklok/toolhive-rfcs/pull/38#issuecomment-3894452220), I had some ideas behind a more generic `Marshal` interface method.

Regardless, let's avoid repeating ourselves if we can.

pkg/vmcp/session/default_session.go
```diff
@@ -0,0 +1,222 @@
...
+	connections     map[string]backend.Session // backend workload ID → persistent backend session
+	routingTable    *vmcp.RoutingTable
+	tools           []vmcp.Tool
+	resources       []vmcp.Resource
+	prompts         []vmcp.Prompt
+	backendSessions map[string]string // backend workload ID → backend-assigned session ID
+
+	mu     sync.RWMutex
+	wg     sync.WaitGroup
+	closed bool
```
jerm-dro, line 83: Blocker, but can be addressed in subsequent PR: The use of mutexes, wait group, and a closed bool is kind of complex. I like the solution of using the `sync.WaitGroup`, but I think it adds complexity to the `defaultMultiSession` object.

One idea:

factor this out into:
```go
type AdmissionQueue interface {

  // TryAdmit returns true if the request is admitted.
  // If true is returned, the returned func() should be invoked
  // when the request is complete.
  TryAdmit() (bool, func())
  
  // CloseAndDrain immediately causes all subsequent calls to `TryAdmit` to return false.
  // CloseAndDrain blocks until all previously admitted requests are complete.
  // Idempotent - once closed, the queue remains closed.
  CloseAndDrain()
}
```

Your implementation can rely on `sync.WaitGroup` & mutex and bool for closure. That implementation can be thoroughly unit-tested in a way that's decoupled from the `Session` concept.

Then, the `defaultMultiSession` won't have to be complicated by this and doesn't need the mutex to protect admission.

pkg/vmcp/session/internal/backend/mcp_session.go
```diff
@@ -0,0 +1,428 @@
...
+	identity *auth.Identity
+}
+
+func (i *identityRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if i.identity != nil {
+		ctx := auth.WithIdentity(req.Context(), i.identity)
+		req = req.Clone(ctx)
+	}
+	return i.base.RoundTrip(req)
+}
```
jerm-dro, line 71: Blocker, but can be addressed in later PR: These round trippers are very important, but it doesn't appear we have tests that validate their implementation.

Can you find a way to test these? The in process server may help here.

jerm-dro (approved):
Approving because I know there are many more PRs to come and don't want to prevent progress, but please address the blockers.

pkg/vmcp/session/internal/backend/mcp_session.go
yrobla: will do in a follow up

PR #3885, opened 2026-02-18
Suppress update message if is trigger by desktop

samuv:
When the thv is bundled with the Desktop app, we don’t need to show the upgrade message because it’s handled by the app’s release cycle.

### DIFF
--- a/cmd/thv/app/commands.go
+++ b/cmd/thv/app/commands.go
@@ -116,10 +116,6 @@ func checkForUpdates() {
 	if updates.ShouldSkipUpdateChecks() {
 		return
 	}
-	// Desktop app manages its own updates
-	if desktop.IsDesktopManagedCLI() {
-		return
-	}
 
 	versionClient := updates.NewVersionClient()
 	updateChecker, err := updates.NewUpdateChecker(versionClient)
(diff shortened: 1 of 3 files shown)


eleftherias:
Because we bundle the update check with metrics, I wonder if this would cause us to lose CLI usage metrics. @lujunsan Could you take a look?

cmd/thv/app/commands.go
```diff
@@ -116,6 +116,10 @@ func checkForUpdates() {
 	if updates.ShouldSkipUpdateChecks() {
 		return
 	}
+	// Desktop app manages its own updates
+	if desktop.IsDesktopManagedCLI() {
```
lujunsan: I agree with the intent of the change, but currently this is potentially an issue; we bundle the update check with the gathering of some metrics, so stopping this here will also stop the metrics from being recorded.

In my opinion, if the intent is to suppress the update message on the CLI, we can add this same logic to decide if we should print the message or not, or to print something else.

samuv:
thanks @eleftherias and @lujunsan. I’m moving the check to where the message-printing logic lives, as suggested, so we can leave the rest untouched

cmd/thv/app/commands.go
samuv: sure, it makes sense, moved!

samuv:
@claude review this PR

PR #3894, opened 2026-02-19
Implement OCI Build and Push for skill service

JAORMX:
## Summary

This is **Part 1 of #3650** (Complete Skill Service OCI Operations), which adds the ability to package skills into OCI artifacts and push them to registries via the `thv serve` REST API.

### Background

The skill service (`pkg/skills/skillsvc/`) already handles listing, installing (extraction-based), uninstalling, and querying skills. However, the `Build`, `Push`, and `Validate` operations were stubbed out as 501 Not Implemented. The OCI primitives needed to implement them — `Store`, `SkillPackager`, and `RegistryClient` — are already available in `toolhive-core` (v0.0.5), but were not wired into the service.

### What this PR does

**Service layer (`skillsvc.go`):** Adds three new functional options (`WithOCIStore`, `WithPackager`, `WithRegistryClient`) to inject OCI dependencies into the skill service. Implements `Build()` which validates the skill path, calls the packager to create a local OCI artifact, tags it in the local store, and returns a reference. Implements `Push()` which resolves a local reference and pushes the artifact to a remote registry. Updates `Validate()` to delegate to `ValidateSkillDir` after path validation.

**Server wiring (`server.go`):** In the default manager creation path, initializes the OCI store (at `DefaultStoreRoot()`), registry client, and packager, and passes them to the skill service constructor.

**HTTP handlers (`skills.go`):** Replaces the three 501 stub handlers with real implementations that decode JSON requests, delegate to the service, and encode responses. Updates Swagger annotations accordingly.

### Security hardening

A code review identified several issues that are addressed in this PR:

- **Path traversal (HIGH):** `Build` and `Validate` accept a filesystem path from the API. Added `validateLocalPath()` which rejects empty paths, relative paths, and paths containing `..` traversal segments. The check operates on the raw path before `filepath.Clean` to catch traversal attempts that Clean would silently resolve.
- **OCI tag injection (MEDIUM):** Added `validateOCITag()` which enforces the OCI distribution spec tag format (`^[\w][\w.-]{0,127}$`) before writing to the local store.
- **Error information leakage (MEDIUM):** The `Push` 404 path previously wrapped the internal store error, which could expose filesystem paths. Changed to return a generic "reference not found" message without wrapping the underlying error.

### What comes next (Part 2)

Part 2 will add OCI pull-based Install — when `Install()` receives a registry reference but no layer data, it will pull the artifact from the registry, extract the layer, and hand off to the existing extraction path. It builds on the `ociStore` and `registry` fields added here.

## Test plan

- [x] `TestBuild` — 10 table-driven cases covering nil packager, empty/relative/traversal paths, invalid tag, packager error, successful build with explicit tag, build falling back to config name, and build returning digest reference
- [x] `TestPush` — 5 table-driven cases covering nil registry, empty reference, resolve-not-found, registry push error, and successful push
- [x] `TestValidate` — extended with path validation cases (empty, relative, traversal)
- [x] Handler tests for validate/build/push — success, bad request, and service error for each
- [x] `task test` — all unit tests pass
- [x] `task lint` — 0 issues
- [x] `task docs` — swagger docs regenerated

### DIFF
--- a/pkg/api/v1/skills.go
+++ b/pkg/api/v1/skills.go
@@ -181,10 +181,25 @@ func (s *SkillsRoutes) getSkillInfo(w http.ResponseWriter, r *http.Request) erro
 //	@Produce		json
 //	@Param			request	body		validateSkillRequest	true	"Validate request"
 //	@Success		200		{object}	skills.ValidationResult
-//	@Failure		501		{string}	string	"Not Implemented"
+//	@Failure		400		{string}	string	"Bad Request"
+//	@Failure		500		{string}	string	"Internal Server Error"
 //	@Router			/api/v1beta/skills/validate [post]
-func (*SkillsRoutes) validateSkill(_ http.ResponseWriter, _ *http.Request) error {
-	return httperr.WithCode(fmt.Errorf("not implemented"), http.StatusNotImplemented)
+func (s *SkillsRoutes) validateSkill(w http.ResponseWriter, r *http.Request) error {
+	var req validateSkillRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return httperr.WithCode(
+			fmt.Errorf("invalid request body: %w", err),
+			http.StatusBadRequest,
+		)
+	}
+
+	result, err := s.skillService.Validate(r.Context(), req.Path)
+	if err != nil {
+		return err
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	return json.NewEncoder(w).Encode(result)
 }
 
 // buildSkill builds a skill from a local directory into an OCI artifact.
@@ -196,10 +211,28 @@ func (*SkillsRoutes) validateSkill(_ http.ResponseWriter, _ *http.Request) error
 //	@Produce		json
 //	@Param			request	body		buildSkillRequest	true	"Build request"
 //	@Success		200		{object}	skills.BuildResult
-//	@Failure		501		{string}	string	"Not Implemented"
+//	@Failure		400		{string}	string	"Bad Request"
+//	@Failure		500		{string}	string	"Internal Server Error"
 //	@Router			/api/v1beta/skills/build [post]
-func (*SkillsRoutes) buildSkill(_ http.ResponseWriter, _ *http.Request) error {
-	return httperr.WithCode(fmt.Errorf("not implemented"), http.StatusNotImplemented)
+func (s *SkillsRoutes) buildSkill(w http.ResponseWriter, r *http.Request) error {
+	var req buildSkillRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		return httperr.WithCode(
+			fmt.Errorf("invalid request body: %w", err),
+			http.StatusBadRequest,
+		)
+	}
+
+	result, err := s.skillService.Build(r.Context(), skills.BuildOptions{
+		Path: req.Path,
+		Tag:  req.Tag,
+	})
+	if err != nil {
+		return err
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	return json.NewEncoder(w).Encode(result)
 }
 
 // pushSkill pushes a built skill artifact to a remote registry.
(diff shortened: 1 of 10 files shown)


pkg/api/v1/skills.go
```diff
@@ -210,8 +243,24 @@ func (*SkillsRoutes) buildSkill(_ http.ResponseWriter, _ *http.Request) error {
 //	@Accept			json
 //	@Param			request	body	pushSkillRequest	true	"Push request"
 //	@Success		204		{string}	string	"No Content"
-//	@Failure		501		{string}	string	"Not Implemented"
+//	@Failure		400		{string}	string	"Bad Request"
+//	@Failure		500		{string}	string	"Internal Server Error"
```
aponcedeleonch, line 248: The `Push` service method returns `http.StatusNotFound` when `ociStore.Resolve` fails (see `skillsvc.go:289-292`), but the Swagger annotations only declare `400` and `500`.

Note: I checked `validateSkill` and `buildSkill` too — neither can return 404. `ValidateSkillDir` returns a `ValidationResult{Valid: false}` (not an HTTP error) when the path is missing, and `Build` wraps packager errors as 500s. So only this handler needs the 404 annotation.

```suggestion
//	@Failure		400		{string}	string	"Bad Request"
//	@Failure		404		{string}	string	"Not Found"
//	@Failure		500		{string}	string	"Internal Server Error"
```

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -184,17 +215,88 @@ func (s *service) Info(ctx context.Context, opts skills.InfoOptions) (*skills.Sk
...
+			return nil, err
+		}
+	}
+
+	result, err := s.packager.Package(ctx, opts.Path, ociskills.DefaultPackageOptions())
+	if err != nil {
+		return nil, fmt.Errorf("packaging skill: %w", err)
+	}
+
+	tag := func() string {
```
aponcedeleonch, line 250: The tag fallback chain is non-obvious. A short comment documenting the precedence would help future readers:

```suggestion
	// Tag resolution precedence:
	// 1. Explicit tag from BuildOptions.Tag
	// 2. Skill name from the parsed config (SKILL.md frontmatter)
	// 3. No tag — use raw digest as the reference
	tag := func() string {
```

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -362,6 +464,40 @@ func buildInstalledSkill(
 	}
 }
 
+// validateLocalPath checks that a path is non-empty, absolute, and does not
+// contain ".." path traversal segments. This prevents API clients from
+// accessing arbitrary directories on the host filesystem via traversal.
+func validateLocalPath(path string) error {
```
aponcedeleonch, line 474: Other path validators in the codebase (`pkg/permissions/profile.go:validatePath`, `pkg/workloads/types/validate.go:analyzeWorkloadName`) also check for null bytes. Consider adding a null-byte check here for defense-in-depth:

```go
if strings.Contains(path, "\x00") {
    return httperr.WithCode(errors.New("path contains null bytes"), http.StatusBadRequest)
}
```

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -362,6 +464,40 @@ func buildInstalledSkill(
 	}
 }
 
+// validateLocalPath checks that a path is non-empty, absolute, and does not
```
aponcedeleonch, line 471: Minor/low-priority: this function is general-purpose "validate an API-supplied filesystem path" logic. Consider whether it could live in a shared location (`pkg/fileutils/` or `pkg/skills/`) so other API handlers that accept paths can reuse it. Not a blocker for this PR.

aponcedeleonch (dismissed):
LGTM! Clean implementation with solid security hardening and good test coverage. A few minor suggestions inline.

pkg/api/v1/skills.go
```diff
@@ -210,8 +243,26 @@ func (*SkillsRoutes) buildSkill(_ http.ResponseWriter, _ *http.Request) error {
 //	@Accept			json
 //	@Param			request	body	pushSkillRequest	true	"Push request"
 //	@Success		204		{string}	string	"No Content"
-//	@Failure		501		{string}	string	"Not Implemented"
+//	@Failure		400		{string}	string	"Bad Request"
+//	@Failure		400		{string}	string	"Bad Request"
```
aponcedeleonch: This line is replicated
```
//	@Failure		400		{string}	string	"Bad Request"
```

aponcedeleonch (approved):

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -184,17 +215,92 @@ func (s *service) Info(ctx context.Context, opts skills.InfoOptions) (*skills.Sk
...
+		return nil, fmt.Errorf("packaging skill: %w", err)
+	}
+
+	// Tag resolution precedence:
+	// 1. Explicit tag from BuildOptions.Tag
+	// 2. Skill name from the parsed config (SKILL.md frontmatter)
+	// 3. No tag — use raw digest as the reference
+	tag := func() string {
+		if opts.Tag != "" {
+			return opts.Tag
```
jhrozek, line 252: should this tag be validated, too?

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -184,17 +215,92 @@ func (s *service) Info(ctx context.Context, opts skills.InfoOptions) (*skills.Sk
...
+	}
+	if opts.Reference == "" {
+		return httperr.WithCode(
+			errors.New("reference is required"),
+			http.StatusBadRequest,
+		)
+	}
+
+	d, err := s.ociStore.Resolve(ctx, opts.Reference)
+	if err != nil {
```
jhrozek, line 292: nit, we don't really log the real error message and always return a 404 regardless of the real issue

Issue #3762, opened 2026-02-10
Add similarity search configuration parameters to the optimizer


jerm-dro:
**Depends on:** #3733

### Context

The mcp-optimizer supports several configurable parameters that control similarity search behavior. As the optimizer is migrated into vMCP, these parameters need to be exposed through the `OptimizerConfig` (`pkg/vmcp/config/config.go`) so operators can tune search quality and result limits.

The mcp-optimizer currently supports:

| Parameter | Type | Default | Range | Description |
|-----------|------|---------|-------|-------------|
| `tool_distance_threshold` | float | 1.0 | 0.0–2.0 | Max cosine distance for filtering results. 0 = identical, 2 = opposite. |
| `hybrid_search_semantic_ratio` | float | 0.5 | 0.0–1.0 | Blend ratio of semantic vs BM25 search. 0.0 = all BM25, 1.0 = all semantic. |
| `max_tools_to_return` | int | 8 | 1–50 | Maximum number of tools returned from a search (top-k). |

Parameters from the mcp-optimizer that are **not applicable** to vMCP:
- `embedding_model_name`, `embedding_threads`, `enable_embedding_cache` — vMCP uses an external EmbeddingServer, not a local model
- `max_servers_to_return` — vMCP operates at the tool level, not the server level
- `registry_ingestion_batch_size`, `workload_ingestion_batch_size` — different ingestion architecture
- `encoding` — token counting, not relevant to vMCP's optimizer

### Requirements

1. Add the following fields to `OptimizerConfig` in `pkg/vmcp/config/config.go`:
   - `ToolDistanceThreshold` (float, default 1.0) — max cosine distance for filtering search results
   - `HybridSearchSemanticRatio` (float, default 0.5) — blend ratio between semantic and BM25 search
   - `MaxToolsToReturn` (int, default 8) — maximum number of tools returned per search
2. Wire these config values through to the `ToolStore` and optimizer so they are respected at search time.
3. Expose these as CRD fields on the VirtualMCPServer resource for Kubernetes deployments.
4. Validate ranges at config load time (distance: 0.0–2.0, ratio: 0.0–1.0, max tools: 1–50).

### High-Level Implementation

- Add fields to `OptimizerConfig` with appropriate defaults and validation
- Pass the config values to the SQLite `ToolStore` hybrid search implementation
- Add corresponding CRD fields to the VirtualMCPServer spec
- Update the factory wiring in `cmd/vmcp/app/commands.go`

[...]

PR #3896, opened 2026-02-19
Add configurable search parameters to optimizer

aponcedeleonch:
Closes: https://github.com/stacklok/toolhive/issues/3762

## Summary

Expose three search tuning parameters through `OptimizerConfig` so operators can control search quality and result limits without code changes:

- **`maxToolsToReturn`** — Maximum number of tool results per search query (1–50, default: 8)
- **`hybridSearchSemanticRatio`** — Balance between semantic (meaning-based) and keyword search (0.0 = all keyword, 1.0 = all semantic, default: 0.5)
- **`semanticDistanceThreshold`** — Maximum distance for semantic search results; results beyond this threshold are filtered out (0 = identical, 2 = completely unrelated, default: 1.0). Does not apply to keyword search.

Unset values use sensible defaults. Float parameters are serialized as strings in the CRD for portability, and converted to native Go types by `optimizer.GetAndValidateConfig()`.

## Changes

- Add three new fields to `config.OptimizerConfig` (CRD/YAML) and corresponding parsed fields in `optimizer.Config` (internal runtime type)
- Add `optimizer.GetAndValidateConfig()` to validate ranges and parse string→float conversion in a single place
- Update `NewSQLiteToolStore` to accept `optimizer.Config` and plumb the search parameters into the store
- Consolidate optimizer validation: removed duplicate validation from `config.Validator`, now handled solely by `optimizer.GetAndValidateConfig()`
- Refactor `server.go` to accept the parsed `optimizer.Config` directly instead of the raw CRD config
- Table-driven unit tests for `GetAndValidateConfig` covering valid values, boundaries, and error cases
- Updated CRD manifests and generated docs


cmd/vmcp/app/commands.go
```diff
@@ -502,6 +506,59 @@ func runServe(cmd *cobra.Command, _ []string) error {
 	return srv.Start(ctx)
 }
 
+func getOptimizerConfig(cfg *config.Config) (*optimizer.Config, error) {
```
jerm-dro: nits:
- [ ] rename this `GetAndValidateConfig`
- [ ] move this to where the `optimizer.Config` is defined
- [ ] add unit tests

pkg/vmcp/config/config.go
```diff
@@ -748,6 +748,31 @@ type OptimizerConfig struct {
...
+
+	// SemanticDistanceThreshold is the maximum cosine distance for semantic search results.
+	// Results with distance > threshold are filtered out in semantic search only.
+	// This does not apply to FTS5: BM25 ranks are normalized scores, not true distances.
+	// Cosine distance: 0 = identical, 2 = opposite.
+	// Defaults to "1.0" if not specified or empty.
+	// Serialized as a string because CRDs do not support float types portably.
+	// +kubebuilder:validation:Pattern=`^([0-9]*[.])?[0-9]+$`
+	// +optional
+	SemanticDistanceThreshold string `json:"semanticDistanceThreshold,omitempty" yaml:"semanticDistanceThreshold,omitempty"`
```
jerm-dro: Suggestion: this documentation leaks implementation details (e.g. FTS5, BM25). Can we put this in language that's higher level for a user that doesn't know how tool search is implemented?

pkg/vmcp/config/validator.go
```diff
@@ -430,6 +436,38 @@ func (*DefaultValidator) validateCompositeToolRefs(refs []CompositeToolRef) erro
 	return nil
 }
 
+func (*DefaultValidator) validateOptimizer(opt *OptimizerConfig) error {
```
jerm-dro: Blocker: this appears to duplicate some code `getOptimizerConfig`. Let's consolidate.

pkg/vmcp/optimizer/internal/types/types.go
```diff
@@ -58,3 +58,18 @@ type EmbeddingClient interface {
...
+
+	// MaxToolsToReturn limits the number of tools returned by FindTool.
+	MaxToolsToReturn *int `json:"max_tools_to_return"`
+
+	// HybridSemanticRatio controls the balance between semantic and exact search in hybrid mode.
+	HybridSemanticRatio *float64 `json:"hybrid_semantic_ratio"`
+
+	// SemanticDistanceThreshold sets the maximum cosine distance for semantic search results (0.0 = identical, 2.0 = opposite).
+	SemanticDistanceThreshold *float64 `json:"semantic_distance_threshold"`
+}
```
jerm-dro: Blocker: does this differ from the `config.OptimizerConfig`? If not, delete it. Historically, creating duplicate type definitions causes problems:
- developers don't realize there's two types and forget to update both
- developers must remember to convert one type to the other. If that doesn't happen, configuration is silently dropped.

jerm-dro (changes requested):

cmd/vmcp/app/commands.go
aponcedeleonch: Addressed:
- **Renamed**: Function is now `GetAndValidateConfig`
- **Moved**: Function lives in `pkg/vmcp/optimizer/optimizer.go`, next to the `optimizer.Config` type alias it returns
- **Unit tests**: Table-driven tests added in `pkg/vmcp/optimizer/optimizer_test.go` covering valid values, boundary values, and all error cases

pkg/vmcp/config/config.go
aponcedeleonch: Addressed. The field comments now use user-friendly language:
- `HybridSearchSemanticRatio`: "controls the balance between semantic (meaning-based) and keyword search results"
- `SemanticDistanceThreshold`: "maximum distance for semantic search results" with "Range: 0 = identical, 2 = completely unrelated"
- No references to FTS5, BM25, or other implementation internals

The only implementation-adjacent note is "Serialized as a string because CRDs do not support float types portably" — this explains the design choice for CRD authors rather than search internals. Let me know if you'd like that removed too.

pkg/vmcp/config/validator.go
aponcedeleonch: Addressed. The duplicate `validateOptimizer` method was removed from the validator. Validation is now consolidated into a single function: `optimizer.GetAndValidateConfig()` in `pkg/vmcp/optimizer/optimizer.go`.

The validator has a comment pointing to the optimizer package:
```go
// Note: Optimizer validation is handled by optimizer.GetAndValidateConfig
// in pkg/vmcp/optimizer/optimizer.go when the optimizer is constructed.
```

The call site in `commands.go` now calls `optimizer.GetAndValidateConfig(cfg.Optimizer)` before constructing the server config.

pkg/vmcp/optimizer/internal/types/types.go
aponcedeleonch: This type is intentionally different from `config.OptimizerConfig` and serves a distinct purpose:

- **`config.OptimizerConfig`** is CRD/YAML-serializable. Kubernetes CRDs don't support float types portably, so `HybridSearchSemanticRatio` and `SemanticDistanceThreshold` are `string` fields, and `MaxToolsToReturn` is a plain `int` (where zero means "unset").
- **`types.OptimizerConfig`** (aliased as `optimizer.Config`) holds the parsed, validated Go values: `*float64` for the float fields and `*int` for max tools (pointers to distinguish "not set" from "zero value").

The conversion between them happens in `optimizer.GetAndValidateConfig()`, which parses strings to floats, validates ranges, and returns the typed config.

Added a doc comment on `types.OptimizerConfig` explaining this rationale:
```go
// This struct intentionally duplicates some fields from config.OptimizerConfig
// (pkg/vmcp/config) because the two serve different purposes:
//   - config.OptimizerConfig is the CRD/YAML-serializable type. Kubernetes CRDs
//     do not support float types portably, so float parameters are encoded as strings.
//   - This struct holds the parsed, validated, native Go values (float64, *int)
//     consumed by the optimizer internals.
```

PR #3898, opened 2026-02-19
Add OCI pull-based install flow for skills

### LINKED ISSUE #3650: Skill service implementation

JAORMX:
## Summary

- When the `Install()` Name field contains an OCI reference (e.g., `ghcr.io/org/my-skill:v1`), the service pulls the artifact from the remote registry, extracts metadata and layer data, then feeds it through the existing extraction pipeline
- Two-step OCI reference detection: structural pre-check (`strings.ContainsAny` for `/:@`) + `go-containerregistry` `ParseReference` validation — provably disjoint from skill name character set
- No API schema changes — the existing `installSkillRequest.Name` field accepts both plain names and OCI references

### Security hardening

- **Context timeout**: 5-minute timeout on registry pulls prevents hangs from slow/malicious registries
- **Name mismatch = hard error**: Declared skill name must match the OCI reference repository name (supply chain defense per [Agent Skills spec](https://agentskills.io/specification))
- **Compressed layer size limit**: 50MB pre-check before loading into memory (prevents OOM before decompression limits kick in)
- **Layer media type validation**: Rejects unexpected content (must be `tar+gzip`)
- **Skill name validation**: `ValidateSkillName` on the config-extracted name
- **Existing protections apply**: Decompression bomb limit (500MB), per-file limit (100MB), file count (1000), path traversal rejection, symlink rejection, permission sanitization

Closes #3650

## Test plan

- [x] Table-driven `TestInstallFromOCI` with 10 cases: registry/store/pathResolver not configured, pull error, invalid skill name in artifact, name mismatch rejected (422), successful pull+install, version hydration from config, caller version preserved, invalid OCI reference (400)
- [x] Tests use real `ociskills.Store` + `Packager` for high-fidelity artifact creation
- [x] All existing tests pass (`task test`)
- [x] Lint clean (`task lint-fix` — 0 issues)

### DIFF
--- a/pkg/skills/skillsvc/skillsvc.go
+++ b/pkg/skills/skillsvc/skillsvc.go
@@ -144,16 +144,15 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
 		opts.ProjectRoot = filepath.Clean(opts.ProjectRoot)
 	}
 
-	// Structural check: skill names never contain '/', ':', or '@'.
-	// OCI references always require at least one of these.
-	if strings.ContainsAny(opts.Name, "/:@") {
-		if _, err := nameref.ParseReference(opts.Name); err != nil {
-			return nil, httperr.WithCode(
-				fmt.Errorf("invalid OCI reference %q: %w", opts.Name, err),
-				http.StatusBadRequest,
-			)
-		}
-		return s.installFromOCI(ctx, opts, scope)
+	ref, isOCI, err := parseOCIReference(opts.Name)
+	if err != nil {
+		return nil, httperr.WithCode(
+			fmt.Errorf("invalid OCI reference %q: %w", opts.Name, err),
+			http.StatusBadRequest,
+		)
+	}
+	if isOCI {
+		return s.installFromOCI(ctx, opts, scope, ref)
 	}
 
 	// Plain skill name — validate and proceed with existing flow.
@@ -331,10 +330,27 @@ const ociPullTimeout = 5 * time.Minute
 // malicious artifact from causing OOM before the decompression limits kick in.
 const maxCompressedLayerSize int64 = 50 * 1024 * 1024 // 50 MB
 
+func parseOCIReference(name string) (nameref.Reference, bool, error) {
+	// Structural check: skill names never contain '/', ':', or '@'.
+	// OCI references always require at least one of these.
+	if !strings.ContainsAny(name, "/:@") {
+		return nil, false, nil
+	}
+
+	ref, err := nameref.ParseReference(name)
+	if err != nil {
+		return nil, true, err
+	}
+	return ref, true, nil
+}
+
 // installFromOCI pulls a skill artifact from a remote registry, extracts
 // metadata and layer data, then delegates to the standard extraction flow.
 func (s *service) installFromOCI(
-	ctx context.Context, opts skills.InstallOptions, scope skills.Scope,
+	ctx context.Context,
+	opts skills.InstallOptions,
+	scope skills.Scope,
+	ref nameref.Reference,
 ) (*skills.InstallResult, error) {
 	if s.registry == nil || s.ociStore == nil {
 		return nil, httperr.WithCode(
@@ -356,7 +372,10 @@ func (s *service) installFromOCI(
 
 	pulledDigest, err := s.registry.Pull(pullCtx, s.ociStore, ociRef)
 	if err != nil {
-		return nil, fmt.Errorf("pulling OCI artifact %q: %w", ociRef, err)
+		return nil, httperr.WithCode(
+			fmt.Errorf("pulling OCI artifact %q: %w", ociRef, err),
+			http.StatusBadGateway,
+		)
 	}
 
 	layerData, skillConfig, err := s.extractOCIContent(ctx, pulledDigest)
@@ -377,20 +396,18 @@ func (s *service) installFromOCI(
 	// match the repository name in the OCI reference. A mismatch could
 	// indicate a supply chain attack (e.g., a trusted reference pointing to
 	// an artifact that overwrites a different skill).
-	if ref, parseErr := nameref.ParseReference(ociRef); parseErr == nil {
-		repo := ref.Context().RepositoryStr()
-		if idx := strings.LastIndex(repo, "/"); idx >= 0 {
-			repo = repo[idx+1:]
-		}
-		if repo != skillConfig.Name {
-			return nil, httperr.WithCode(
-				fmt.Errorf(
-					"skill name %q in artifact does not match OCI reference repository %q",
-					skillConfig.Name, repo,
-				),
-				http.StatusUnprocessableEntity,
-			)
-		}
+	repo := ref.Context().RepositoryStr()
+	if idx := strings.LastIndex(repo, "/"); idx >= 0 {
+		repo = repo[idx+1:]
+	}
+	if repo != skillConfig.Name {
+		return nil, httperr.WithCode(
+			fmt.Errorf(
+				"skill name %q in artifact does not match OCI reference repository %q",
+				skillConfig.Name, repo,
+			),
+			http.StatusUnprocessableEntity,
+		)
 	}
 
 	// Hydrate install options from the pulled artifact.
@@ -401,6 +418,7 @@ func (s *service) installFromOCI(
 	if opts.Version == "" && skillConfig.Version != "" {
 		opts.Version = skillConfig.Version
 	}
+	// Note: version is optional; if both are empty, install without a version.
 
 	unlock := s.locks.lock(opts.Name, scope, opts.ProjectRoot)
 	defer unlock()
@@ -450,6 +468,8 @@ func (s *service) extractOCIContent(ctx context.Context, d digest.Digest) ([]byt
 		)
 	}
 
+	// Skills use a single-layer format (one tar.gz). Validate the first
+	// (and only expected) layer.
 	if manifest.Layers[0].MediaType != ocispec.MediaTypeImageLayerGzip {
 		return nil, nil, httperr.WithCode(
 			fmt.Errorf("unexpected layer media type %q, expected %q",
--- a/pkg/skills/skillsvc/skillsvc_test.go
+++ b/pkg/skills/skillsvc/skillsvc_test.go
@@ -5,6 +5,7 @@ package skillsvc
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"os"
@@ -15,6 +16,8 @@ import (
 	"time"
 
 	godigest "github.com/opencontainers/go-digest"
+	specs "github.com/opencontainers/image-spec/specs-go"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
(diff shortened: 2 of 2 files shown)


pkg/skills/skillsvc/skillsvc.go
```diff
@@ -303,6 +323,175 @@ func (s *service) Push(ctx context.Context, opts skills.PushOptions) error {
...
+	// name field matches the parent directory name; by extension, it should
+	// match the repository name in the OCI reference. A mismatch could
+	// indicate a supply chain attack (e.g., a trusted reference pointing to
+	// an artifact that overwrites a different skill).
+	if ref, parseErr := nameref.ParseReference(ociRef); parseErr == nil {
+		repo := ref.Context().RepositoryStr()
+		if idx := strings.LastIndex(repo, "/"); idx >= 0 {
+			repo = repo[idx+1:]
+		}
+		if repo != skillConfig.Name {
```
reyortiz3: does case sensitive matter here?

JAORMX: Per the Agent Skills spec, `name` must be lowercase and match the parent directory name, so the comparison is intended to be case-sensitive. We already validate the skill name (lowercase only), and OCI repo names are typically lowercase, so a mismatch here should indicate an invalid artifact or supply-chain issue rather than a casing quirk.

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -141,6 +144,23 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
...
+	// OCI references always require at least one of these.
+	if strings.ContainsAny(opts.Name, "/:@") {
+		if _, err := nameref.ParseReference(opts.Name); err != nil {
+			return nil, httperr.WithCode(
+				fmt.Errorf("invalid OCI reference %q: %w", opts.Name, err),
+				http.StatusBadRequest,
+			)
+		}
+		return s.installFromOCI(ctx, opts, scope)
+	}
```
aponcedeleonch, line 156: **Suggestion: Extract OCI reference detection into a reusable function**

The structural check (`strings.ContainsAny`) + `ParseReference` validation block could be extracted into a small helper like `classifyInstallName(name string) (isOCI bool, err error)`. This would:
- Improve readability of `Install()` (currently interleaving two concerns: routing and validation)
- Be independently testable
- Be reusable if other entry points (e.g., future "upgrade" or "info by OCI ref") need the same logic

Not a blocker — the current inline version is clear enough — but would improve maintainability.

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -303,6 +323,175 @@ func (s *service) Push(ctx context.Context, opts skills.PushOptions) error {
...
+			)
+		}
+	}
+
+	// Hydrate install options from the pulled artifact.
+	opts.Name = skillConfig.Name
+	opts.LayerData = layerData
+	opts.Reference = ociRef
+	opts.Digest = pulledDigest.String()
+	if opts.Version == "" && skillConfig.Version != "" {
```
aponcedeleonch, line 418: **Clarify behavior when both version sources are empty**

If `opts.Version == ""` *and* `skillConfig.Version == ""` (which is valid — `SkillConfig.Version` is `omitempty`), then `opts.Version` stays empty and the skill is installed with no version.

Is this intentional? If so, a brief comment here would help future readers understand the expected fallthrough. If a version is considered required for OCI-sourced installs, this should be validated.

```suggestion
	if opts.Version == "" && skillConfig.Version != "" {
		opts.Version = skillConfig.Version
	}
	// Note: if both are empty, the skill is installed without a version.
	// This is valid — version is optional in the Agent Skills spec.
```

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -303,6 +323,175 @@ func (s *service) Push(ctx context.Context, opts skills.PushOptions) error {
...
+		)
+	}
+
+	// Supply chain defense: the declared skill name must match the last path
+	// component of the OCI reference. The Agent Skills spec requires that the
+	// name field matches the parent directory name; by extension, it should
+	// match the repository name in the OCI reference. A mismatch could
+	// indicate a supply chain attack (e.g., a trusted reference pointing to
+	// an artifact that overwrites a different skill).
+	if ref, parseErr := nameref.ParseReference(ociRef); parseErr == nil {
```
aponcedeleonch: **Avoid redundant `ParseReference` call — or make the error path explicit**

`ParseReference(ociRef)` was already validated successfully in `Install()` (line 150). This second call silently skips the entire name-mismatch check if parsing somehow fails (`if parseErr == nil`), which:
1. Is a dead code path (can't actually fail here)
2. Could mask a real bug if the code is restructured later

Consider either:
- Passing the already-parsed `nameref.Reference` as a parameter to `installFromOCI`
- Changing `if parseErr == nil` to a hard error (`return nil, fmt.Errorf("BUG: ...)`) since failure here indicates a programming error

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -303,6 +323,175 @@ func (s *service) Push(ctx context.Context, opts skills.PushOptions) error {
...
+	}
+
+	ociRef := opts.Name
+
+	pullCtx, cancel := context.WithTimeout(ctx, ociPullTimeout)
+	defer cancel()
+
+	pulledDigest, err := s.registry.Pull(pullCtx, s.ociStore, ociRef)
+	if err != nil {
+		return nil, fmt.Errorf("pulling OCI artifact %q: %w", ociRef, err)
```
aponcedeleonch: **Minor: Inconsistent error wrapping**

This error doesn't include an `httperr.WithCode()`, so it will bubble up to the API handler without an explicit HTTP status code. Compare with the similar nil-check errors above (lines 341, 346) which use `http.StatusInternalServerError`.

Depending on how the handler defaults, this might result in a 500 anyway, but wrapping it explicitly would be more consistent.

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -303,6 +323,175 @@ func (s *service) Push(ctx context.Context, opts skills.PushOptions) error {
...
+	}
+
+	if len(manifest.Layers) == 0 {
+		return nil, nil, httperr.WithCode(
+			errors.New("OCI manifest contains no layers"),
+			http.StatusUnprocessableEntity,
+		)
+	}
+
+	if manifest.Layers[0].MediaType != ocispec.MediaTypeImageLayerGzip {
```
aponcedeleonch, line 473: **Nit: Document single-layer assumption**

The code consistently uses `manifest.Layers[0]` for media type, size, and blob retrieval. This is correct for the current skill format (single tar.gz layer), but a brief comment documenting this assumption would help future readers — especially if multi-layer artifacts are ever introduced.

Something like:
```go
// Skills use a single-layer format (one tar.gz). Validate the first
// (and only expected) layer.
```

pkg/skills/skillsvc/skillsvc_test.go
```diff
@@ -422,6 +422,280 @@ func TestInstallWithExtraction(t *testing.T) {
...
+			}
+			if tt.wantRefSaved != "" {
+				assert.Equal(t, tt.wantRefSaved, result.Skill.Reference)
+			}
+		})
+	}
+}
+
 func TestUninstall(t *testing.T) {
 	t.Parallel()
```
aponcedeleonch, line 768: **Missing test: oversized layer rejection**

The `maxCompressedLayerSize` check (50MB limit) in `extractOCIContent` doesn't have a corresponding test case. Consider adding a test that creates an artifact with `manifest.Layers[0].Size > maxCompressedLayerSize` to verify the 422 rejection path.

Also minor: the `"registry not configured"` and `"ociStore not configured"` cases exercise very similar code paths (both hit the nil guard at the top of `installFromOCI`). Not a big deal, but they're near-duplicates.

aponcedeleonch (dismissed):

pkg/skills/skillsvc/skillsvc.go
JAORMX: Extracted OCI ref detection into parseOCIReference and pass the parsed ref into installFromOCI for reuse and readability. Fixed in 174357ec.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Added a note clarifying the optional-version behavior; if both are empty we install without a version per spec. Fixed in 174357ec.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Removed the redundant ParseReference by passing the parsed ref into installFromOCI. Fixed in 174357ec.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Wrapped the pull error with httperr.WithCode (502) for explicit status consistency. Fixed in 174357ec.

pkg/skills/skillsvc/skillsvc.go
JAORMX: Documented the single-layer assumption before the layer validation. Fixed in 174357ec.

pkg/skills/skillsvc/skillsvc_test.go
JAORMX: Added an oversized layer rejection test (layer size > maxCompressedLayerSize) to assert the 422 path. Fixed in 174357ec.

Issue #3866, opened 2026-02-18
[vMCP] Implement SessionManager and wire up behind feature flag (Phase 2)


yrobla:
Depends on #3865

Implement `SessionManager` and wire it into the vMCP server behind a `sessionManagementV2` feature flag (default `false`). The old code path must remain fully functional when the flag is disabled.

`SessionManager` bridges domain logic (`Session`/`SessionFactory`) to the MCP SDK protocol by implementing the SDK's `SessionIdManager` interface. It delegates all session storage to `transportsession.Manager`, avoiding a parallel storage path and automatically inheriting pluggable backend support (LocalStorage today, RedisStorage in future).

**New files**:
- `pkg/vmcp/server/session_manager.go` — `SessionManager` implementation
- `pkg/vmcp/config/config.go` — add `sessionManagementV2` flag (default `false`)

**Key methods** (implements SDK's `SessionIdManager`):
- `Generate() string` — phase 1 of the two-phase creation pattern: creates a unique session ID, stores an empty placeholder via `sessionStorage.AddSession()`, returns the ID to the SDK. No context available at this point.
- `CreateSession(ctx, identity, backends) (string, error)` — phase 2: called from the `OnRegisterSession` hook once context is available; calls `SessionFactory.MakeSession()` to build the fully-formed session and replaces the placeholder in storage.
- `Validate(sessionID string) (isTerminated bool, err error)` — checks whether a session exists and is active via `sessionStorage.Get()`.
- `Terminate(sessionID string) (isTerminated bool, err error)` — loads the session, calls `Close()` to release all backend clients, then removes it from storage via `sessionStorage.Delete()`.
- `GetAdaptedTools(sessionID string) []mcp.Tool` — loads the session, converts its domain `Tool` types to SDK format, and attaches handlers that delegate to `session.CallTool()`.

**Files to modify**: `pkg/vmcp/server/server.go`, `pkg/vmcp/server/hooks.go`

**Dual-layer storage model**: `SessionManager` must preserve the separation between the serializable metadata layer (stored via `transportsession.Manager`) and the non-serializable runtime layer (MCP clients, routing table, held in-process only). When `RedisStorage` is plugged in, session metadata must persist automatically without any changes to `SessionManager`. Runtime state is always recreated on demand.

### Acceptance Criteria

- [ ] `SessionManager` is implemented and delegates all storage to `transportsession.Manager` (no separate `sync.Map` or parallel storage)
[...]

PR #3906, opened 2026-02-20
Add SessionManager behind sessionManagementV2 flag (Phase 2)

yrobla:
Implement RFC THV-0038 Phase 2: a vmcpSessionManager that bridges the MultiSession/MultiSessionFactory domain logic to the MCP SDK's SessionIdManager interface. The feature is gated behind a sessionManagementV2 config flag (default false), leaving the existing code path completely unchanged when disabled.

Key changes:
  - pkg/vmcp/server/session_manager.go: new vmcpSessionManager with two-phase creation — Generate() stores a placeholder, CreateSession() (called from OnRegisterSession hook) replaces it with a fully-formed MultiSession via MakeSessionWithID(). Terminate() calls Close() before deleting to release backend connections. GetAdaptedTools() returns SDK tools with session-scoped handlers delegating to CallTool().
  - pkg/vmcp/session/factory.go: add MakeSessionWithID() to the MultiSessionFactory interface; refactor MakeSession to share a private makeSession() implementation.
  - pkg/transport/session/manager.go: add ReplaceSession() upsert to swap a placeholder with a fully-formed MultiSession.
  - pkg/vmcp/config/config.go: add SessionManagementV2 bool field.
  - pkg/vmcp/server/server.go: wire vmcpSessionManager when flag+factory are set; add handleSessionRegistrationV2() hook handler.
  - pkg/vmcp/discovery/middleware.go: skip Phase 1 routing-table reconstruction for Phase 2 MultiSession sessions.
  - Tests: unit tests for all new methods, ReplaceSession coverage in transport/session, and an end-to-end integration test covering initialize, tool call routing, and the old-path-unaffected case.

Closes: #3866

## Large PR Justification

This is an atomic PR that covers an isolated functionality, and includes complete tests. Cannot be split.

### DIFF
--- a/pkg/transport/session/manager.go
+++ b/pkg/transport/session/manager.go
@@ -179,10 +179,10 @@ func (m *Manager) Get(id string) (Session, bool) {
 	return sess, true
 }
 
-// ReplaceSession upserts a session into storage, replacing any existing session
-// with the same ID. Used by SessionManager to replace placeholder sessions
+// UpsertSession inserts or updates a session in storage, replacing any existing
+// session with the same ID. Used by SessionManager to replace placeholder sessions
 // with fully-formed MultiSession objects after phase-2 construction.
-func (m *Manager) ReplaceSession(session Session) error {
+func (m *Manager) UpsertSession(session Session) error {
 	if session == nil {
 		return fmt.Errorf("session cannot be nil")
 	}
--- a/pkg/vmcp/discovery/middleware.go
+++ b/pkg/vmcp/discovery/middleware.go
@@ -12,6 +12,13 @@
 // that periodically rediscovers capabilities, detects changes via hash comparison, and
 // pushes updates to active sessions via MCP tools/list_changed notifications. Middleware
 // flow remains unchanged - still just retrieves from session cache on subsequent requests.
+//
+// TODO(sessionManagementV2): This entire middleware package can be deleted after the
+// sessionManagementV2 migration is complete. For MultiSession (the new path), the
+// middleware does nothing except validate that the session exists, which is already
+// handled by the SDK via SessionIdManager.Validate(). Tool routing is handled by
+// session-scoped handlers registered with AddSessionTools. This middleware only exists
+// for VMCPSession backward compatibility (routing table reconstruction from session data).
 package discovery
 
 import (
@@ -267,6 +274,10 @@ func handleSubsequentRequest(
 	// If the session is a MultiSession, tools are already registered with the SDK via
 	// AddSessionTools and routed by session-scoped handlers. No routing-table
 	// reconstruction is needed — pass through without modifying the context.
+	//
+	// TODO(sessionManagementV2): Remove this type coercion check once the sessionManagementV2
+	// migration is complete and all sessions are MultiSession instances. At that point, this
+	// early return becomes the only code path and the VMCPSession fallback below can be deleted.
 	if _, isMulti := rawSess.(vmcpsession.MultiSession); isMulti {
 		//nolint:gosec // G706: session ID is not an injection vector
 		slog.Debug("session uses session-scoped tool routing; skipping routing-table reconstruction",
--- a/pkg/vmcp/server/server.go
+++ b/pkg/vmcp/server/server.go
@@ -36,6 +36,7 @@ import (
 	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
 	"github.com/stacklok/toolhive/pkg/vmcp/router"
 	"github.com/stacklok/toolhive/pkg/vmcp/server/adapter"
+	"github.com/stacklok/toolhive/pkg/vmcp/server/sessionmanager"
 	vmcpsession "github.com/stacklok/toolhive/pkg/vmcp/session"
 	vmcpstatus "github.com/stacklok/toolhive/pkg/vmcp/status"
 )
@@ -207,7 +208,7 @@ type Server struct {
 
 	// vmcpSessionMgr is the Phase 2 session manager. Non-nil only when
 	// Config.SessionManagementV2 is true and Config.SessionFactory is non-nil.
-	vmcpSessionMgr *vmcpSessionManager
+	vmcpSessionMgr SessionManager
 
 	// Composite tool workflow definitions keyed by tool name.
 	// Initialized during construction and read-only thereafter.
@@ -383,12 +384,12 @@ func New(
 	}
 
 	// Create session manager if the feature flag is enabled.
-	var vmcpSessMgr *vmcpSessionManager
+	var vmcpSessMgr SessionManager
 	if cfg.SessionManagementV2 {
 		if cfg.SessionFactory == nil {
 			return nil, fmt.Errorf("SessionManagementV2 is enabled but no SessionFactory was provided")
 		}
-		vmcpSessMgr = newVMCPSessionManager(sessionManager, cfg.SessionFactory, backendRegistry)
+		vmcpSessMgr = sessionmanager.New(sessionManager, cfg.SessionFactory, backendRegistry)
 		slog.Info("session-scoped backend lifecycle enabled")
 	}
 
@@ -1039,7 +1040,9 @@ func (s *Server) handleSessionRegistration(
 
 	// Delegate to session-scoped backend lifecycle when vmcpSessionMgr is active.
 	if s.vmcpSessionMgr != nil {
-		s.handleSessionRegistrationV2(ctx, session)
+		// Error is logged and handled within handleSessionRegistrationV2.
+		// The session is terminated on failure; no further action needed here.
+		_ = s.handleSessionRegistrationV2(ctx, session)
 		return
 	}
 
@@ -1153,9 +1156,21 @@ func (s *Server) handleSessionRegistration(
 //
 //   - Prompts: not supported in either path until the SDK adds
 //     AddSessionPrompts.
-func (s *Server) handleSessionRegistrationV2(ctx context.Context, session server.ClientSession) {
+func (s *Server) handleSessionRegistrationV2(ctx context.Context, session server.ClientSession) (retErr error) {
 	sessionID := session.SessionID()
 	slog.Debug("creating session-scoped backends", "session_id", sessionID)
+
+	// Defer cleanup: if any error occurs, terminate the session and log failures.
+	defer func() {
+		if retErr != nil {
+			if _, termErr := s.vmcpSessionMgr.Terminate(sessionID); termErr != nil {
+				slog.Warn("failed to clean up session after error",
+					"session_id", sessionID,
+					"error", termErr,
+					"original_error", retErr)
+			}
+		}
+	}()
 
 	// NOTE: the initialize response (including the Mcp-Session-Id header) has
 	// already been sent to the client before this hook fires. Any error below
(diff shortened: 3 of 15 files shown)


pkg/transport/session/manager.go
```diff
@@ -179,6 +179,21 @@ func (m *Manager) Get(id string) (Session, bool) {
 	return sess, true
 }
 
+// ReplaceSession upserts a session into storage, replacing any existing session
+// with the same ID. Used by SessionManager to replace placeholder sessions
+// with fully-formed MultiSession objects after phase-2 construction.
+func (m *Manager) ReplaceSession(session Session) error {
```
jerm-dro: nit, feel free to ignore: why not call this method `UpsertSession`?

pkg/vmcp/discovery/middleware.go
```diff
@@ -224,16 +253,37 @@ func handleSubsequentRequest(
...
+			"session_id", sessionID,
+			"method", r.Method,
+			"path", r.URL.Path)
+		return ctx, fmt.Errorf("session not found: %s", sessionID)
+	}
+
+	// If the session is a MultiSession, tools are already registered with the SDK via
+	// AddSessionTools and routed by session-scoped handlers. No routing-table
+	// reconstruction is needed — pass through without modifying the context.
+	if _, isMulti := rawSess.(vmcpsession.MultiSession); isMulti {
```
jerm-dro, line 281: nit: leave a TODO to delete the type coercion to `MultiSession` once the migration is complete. In the absence of a TODO, it may be easy to forget this step.

pkg/vmcp/server/adapter/conversions_test.go
```diff
@@ -1,144 +0,0 @@
-// SPDX-FileCopyrightText: Copyright 2026 Stacklok, Inc.
```
jerm-dro, line 1: Question, non-blocking: Why did this get moved?

pkg/vmcp/server/server.go
```diff
@@ -197,6 +205,10 @@ type Server struct {
 	// Capability adapter for converting aggregator types to SDK types
 	capabilityAdapter *adapter.CapabilityAdapter
 
+	// vmcpSessionMgr is the Phase 2 session manager. Non-nil only when
+	// Config.SessionManagementV2 is true and Config.SessionFactory is non-nil.
+	vmcpSessionMgr *vmcpSessionManager
```
jerm-dro: Does this have to be the implementation type? or could this be an interface type `server.SessionIdManager`?

pkg/vmcp/server/server.go
```diff
@@ -1084,6 +1125,94 @@ func (s *Server) handleSessionRegistration(
...
+		slog.Error("failed to get session-scoped tools",
+			"session_id", sessionID,
+			"error", err)
+		if _, termErr := s.vmcpSessionMgr.Terminate(sessionID); termErr != nil {
+			slog.Warn("failed to clean up session after tool retrieval failure",
+				"session_id", sessionID,
+				"error", termErr)
+		}
+		return
+	}
```
jerm-dro: Suggestion: this error handling repeats itself. It seems like this function could be refactored so it either:
- does the logging and termination in a `defer func ()` which looks at some `retErr` to decide if it's necessary.
- call a helper function which returns an `err` and then implement the conditional termination outside the helper function.

pkg/vmcp/server/server.go
```diff
@@ -1084,6 +1125,94 @@ func (s *Server) handleSessionRegistration(
...
+	// terminates the session internally, but the client holds a session ID that
+	// will appear to have no tools (tool calls will return "not found" from the
+	// SDK). This is an architectural constraint of the two-phase pattern — there
+	// is no way to retract the session ID after it has been sent.
+	//
+	// NOTE: there is a brief race window between the client receiving the session
+	// ID and this hook completing. A client that pipelines a tools/call immediately
+	// after initialize may receive a "tool not found" error before AddSessionTools
+	// completes. Conforming MCP clients call tools/list before tools/call, so this
+	// window is expected to be harmless in practice.
```
jerm-dro, line 1186: praise: nice comments that explain the trade-offs and hidden complexity here 😄

pkg/vmcp/server/session_management_v2_realbackend_integration_test.go
```diff
@@ -0,0 +1,322 @@
...
+	)
+
+	streamableSrv := mcpserver.NewStreamableHTTPServer(mcpSrv)
+	mux := http.NewServeMux()
+	mux.Handle("/mcp", streamableSrv)
+
+	ts := httptest.NewServer(mux)
+	t.Cleanup(ts.Close)
+	return ts.URL + "/mcp"
+}
```
jerm-dro: Suggestion, non-blocking: This seems like a common test utility that would be useful to factor out into some testcommon/ or something.

pkg/vmcp/server/session_management_v2_realbackend_integration_test.go
```diff
@@ -0,0 +1,322 @@
...
+	initResp := postMCP(t, ts.URL, map[string]any{
+		"jsonrpc": "2.0",
+		"id":      1,
+		"method":  "initialize",
+		"params": map[string]any{
+			"protocolVersion": "2025-06-18",
+			"capabilities":    map[string]any{},
+			"clientInfo":      map[string]any{"name": "test", "version": "1.0"},
+		},
+	}, "")
```
jerm-dro: suggestion, non-blocking: similar to above, it would be nice to factor this out into a testutils directory. Over the long term, tests that rely on real servers would use similar higher-level semantics (e.g. `client.initialize_session`), which reduces cognitive load when reading / writing tests.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -0,0 +1,340 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
jerm-dro, line 1: Suggestion, non-blocking: move this to a separate `sessionmanager/` package would allow us to keep `server/` more focused. Moving this code would also be a good forcing function for clarifying the interface to this code. As currently written, the server relies on a real implementation, which makes unit testing harder.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -0,0 +1,340 @@
...
+	// Resolve the caller identity (may be nil for anonymous access).
+	identity, _ := auth.IdentityFromContext(ctx)
+
+	// List all available backends from the registry.
+	rawBackends := sm.backendRegistry.List(ctx)
+	backends := make([]*vmcp.Backend, len(rawBackends))
+	for i := range rawBackends {
+		b := rawBackends[i] // avoid loop-variable capture
+		backends[i] = &b
+	}
```
jerm-dro, line 157: nit: is loop variable capture still a thing?

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -0,0 +1,340 @@
...
+		// the flag is visible to any storage backend (including distributed ones).
+		// TTL cleanup will remove the record later; the terminated flag lets
+		// Validate() return isTerminated=true during the intervening window.
+		sess.SetMetadata(metadataKeyTerminated, metadataValTrue)
+		if replaceErr := sm.storage.ReplaceSession(sess); replaceErr != nil {
+			slog.Warn("vmcpSessionManager.Terminate: failed to persist terminated flag for placeholder",
+				"session_id", sessionID, "error", replaceErr)
+			// Non-fatal: the in-memory flag is set; TTL will clean up anyway.
+		}
+	}
```
jerm-dro, line 295: Why is this else branch needed? Why would we have a non-`MultiSession`?

pkg/vmcp/server/session_manager.go
```diff
@@ -0,0 +1,340 @@
...
+	multiSess, ok := sm.GetMultiSession(sessionID)
+	if !ok {
+		return nil, fmt.Errorf("vmcpSessionManager.GetAdaptedTools: session %q not found or not a multi-session", sessionID)
+	}
+
+	domainTools := multiSess.Tools()
+	sdkTools := make([]mcpserver.ServerTool, 0, len(domainTools))
+
+	for _, domainTool := range domainTools {
+		t := domainTool // capture loop variable
```
jerm-dro: nit: same as above - not needed

pkg/vmcp/session/factory.go
```diff
@@ -49,6 +50,17 @@ type MultiSessionFactory interface {
...
+	// MakeSessionWithID creates a new MultiSession with a specific session ID.
+	// This is used by SessionManager to create sessions using the SDK-assigned ID
+	// rather than generating a new UUID internally.
+	//
+	// The id parameter must be non-empty and should be a valid MCP session ID
+	// (visible ASCII characters, 0x21 to 0x7E per the MCP specification).
+	//
+	// All other behaviour (partial initialisation, bounded concurrency, etc.)
+	// is identical to MakeSession.
+	MakeSessionWithID(ctx context.Context, id string, identity *auth.Identity, backends []*vmcp.Backend) (MultiSession, error)
```
jerm-dro, line 68: Is `MakeSession` needed on this interface then?

pkg/vmcp/server/server.go
```diff
@@ -484,7 +511,15 @@ func (s *Server) Handler(_ context.Context) (http.Handler, error) {
 	if healthMon != nil {
 		healthStatusProvider = healthMon
 	}
-	mcpHandler = discovery.Middleware(s.discoveryMgr, s.backendRegistry, s.sessionManager, healthStatusProvider)(mcpHandler)
+	discoveryOpts := func() []discovery.MiddlewareOption {
+		if s.vmcpSessionMgr != nil {
+			return []discovery.MiddlewareOption{discovery.WithSessionScopedRouting()}
+		}
+		return nil
+	}()
```
jerm-dro, line 520: question, non-blocking: when the migration is complete, will we delete the `discovery.Middleware` entirely?

IIUC, the `handleSubsequentRequest` functionality for v2 doesn't do anything except validate the session exists, so it would be safe to delete.

pkg/transport/session/manager.go
```diff
@@ -179,6 +179,21 @@ func (m *Manager) Get(id string) (Session, bool) {
...
+// session with the same ID. Used by SessionManager to replace placeholder sessions
+// with fully-formed MultiSession objects after phase-2 construction.
+func (m *Manager) UpsertSession(session Session) error {
+	if session == nil {
+		return fmt.Errorf("session cannot be nil")
+	}
+	if session.ID() == "" {
+		return fmt.Errorf("session ID cannot be empty")
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
aponcedeleonch, line 192: Nit: The same hardcoded `5*time.Second` appears in `AddWithID`, `AddSession`, `Get`, `UpsertSession`, and `Delete`. Consider extracting it to a package-level constant for readability and easier adjustment:

```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), defaultStorageTimeout)
```

With a constant at the top of the file:
```go
const defaultStorageTimeout = 5 * time.Second
```

Can be done in a follow-up.

pkg/vmcp/server/server.go
```diff
@@ -370,6 +383,16 @@ func New(
...
 	}
 
+	// Create session manager if the feature flag is enabled.
+	var vmcpSessMgr SessionManager
+	if cfg.SessionManagementV2 {
+		if cfg.SessionFactory == nil {
+			return nil, fmt.Errorf("SessionManagementV2 is enabled but no SessionFactory was provided")
+		}
+		vmcpSessMgr = sessionmanager.New(sessionManager, cfg.SessionFactory, backendRegistry)
+		slog.Info("session-scoped backend lifecycle enabled")
```
aponcedeleonch, line 393: Suggestion: Since `handleSessionRegistrationV2` documents several unsupported scenarios (Optimizer, Composite tools, Resources, Prompts), it would be helpful to log warnings at startup when incompatible features are configured alongside V2. This way operators get early feedback instead of silent feature loss.

After this line, consider adding:
```go
		if cfg.OptimizerConfig != nil {
			slog.Warn("SessionManagementV2 does not support optimizer mode; optimizer configuration will be ignored")
		}
		if len(workflowDefs) > 0 {
			slog.Warn("SessionManagementV2 does not support composite tools; workflow definitions will be ignored",
				"workflow_count", len(workflowDefs))
		}
```

Can be done in a follow-up.

aponcedeleonch (approved):
Great work on this PR! The two-phase session creation pattern is clean, race conditions are thoughtfully handled with the double-check pattern, and the test coverage is thorough. Approving with two minor suggestions that can be addressed in a follow-up.

pkg/transport/session/manager.go
yrobla: ok will do in a follow up

pkg/vmcp/server/server.go
yrobla: ok will do in a follow up

Issue #3758, opened 2026-02-10
Rename dummyOptimizer to production optimizer


jerm-dro:
### Parent Epic
stacklok/stacklok-epics#201

**Depends on:** #3733

### Context

The optimizer started as a `dummyOptimizer` (`pkg/vmcp/optimizer/dummy_optimizer.go`) with stub substring matching. After issues #3731, #3732, and #3733, it is backed by a shared SQLite store with hybrid FTS5 + vector embedding search and a real EmbeddingServer client. The "dummy" name no longer reflects its capabilities and is misleading.

### Requirements

1. Rename `dummyOptimizer` to a production-appropriate name (e.g., `embeddingOptimizer` or simply `optimizer`) across all files, types, constructors, and references.
2. Rename `dummy_optimizer.go` and `dummy_optimizer_test.go` to match the new name.
3. Update the factory function `NewDummyOptimizer` to a new name (e.g., `NewOptimizer` or `NewEmbeddingOptimizer`).
4. Update all references in `cmd/vmcp/app/commands.go`, `pkg/vmcp/server/server.go`, adapter code, and tests.
5. Remove the TODO comment in `cmd/vmcp/app/commands.go` that says `// TODO: update this with the real optimizer.`

### High-Level Implementation

- Rename the struct, constructor, file names, and test files
- Find and update all references across the codebase
- Verify all tests pass after the rename
- This is a pure refactoring change with no behavioral modifications

### Acceptance Tests

- [ ] **No behavioral changes**: All existing unit tests and e2e tests pass without modification (beyond import/name updates)
- [ ] **No "dummy" references remain**: No references to `dummy_optimizer`, `DummyOptimizer`, or `NewDummyOptimizer` exist in the codebase
- [ ] **TODO removed**: The TODO comment in `cmd/vmcp/app/commands.go` is removed

PR #3909, opened 2026-02-20
Prepare optimizer for production by refactoring and renaming

aponcedeleonch:
Closes: https://github.com/stacklok/toolhive/issues/3758

Prepare the optimizer package for production use by reorganizing code into focused internal sub-packages and replacing placeholder implementations with real ones.

## Changes

- **Remove `DummyOptimizer`**: Replace the fake/placeholder optimizer with a single production `toolOptimizer` that delegates search to `ToolStore` and scopes results per session
- **Remove fake embedding helper** (`internal/similarity/fake_embedding.go`): Inline necessary test helpers directly in test files, removing the placeholder package
- **Extract token counting** into `internal/tokencounter` with a `Counter` interface and `JSONByteCounter` implementation (previously lived in the top-level optimizer package)
- **Rename `internal/sqlite_store` to `internal/toolstore`**: Better reflects the package's role as the tool storage layer
- **Move `store.go`** from the top-level optimizer package into `internal/toolstore`, consolidating store logic in one place
- **Add mockgen-generated mocks** for `ToolStore` and `EmbeddingClient` interfaces
- **Add comprehensive unit tests** for the production `toolOptimizer` using mock store
- **Remove unused exported aliases** (`EmbeddingClient`, `NewEmbeddingClient`)

## Large PR Justification

This is a refactoring PR that reorganizes internal packages. The high line count is driven by file moves/renames and replacing placeholder implementations with production code, along with their corresponding tests. Splitting further would leave the optimizer in a broken intermediate state.


pkg/vmcp/optimizer/internal/toolstore/sqlite_store.go
```diff
@@ -267,11 +275,14 @@ func (s sqliteToolStore) searchFTS5(
 		matches = append(matches, types.ToolMatch{
 			Name:        name,
 			Description: description,
-			Score:       normalizeBM25(rank),
```
yrobla, line 270: mm i see Score removed, but i see a leftover:

```
yolanda@MacBook-Pro-de-Yolanda toolhive % cat pkg/vmcp/server/adapter/optimizer_adapter_test.go|grep Score
                                                Score:       1.0,
```

pkg/vmcp/optimizer/optimizer.go
```diff
@@ -23,6 +23,8 @@ import (
 
 	vmcpconfig "github.com/stacklok/toolhive/pkg/vmcp/config"
 	"github.com/stacklok/toolhive/pkg/vmcp/optimizer/internal/similarity"
+	"github.com/stacklok/toolhive/pkg/vmcp/optimizer/internal/tokencounter"
```
yrobla, line 26: commit message says "adding logs to optimizer", but i don't see logs added?

pkg/vmcp/optimizer/internal/toolstore/sqlite_store.go
aponcedeleonch: Good catch\! Score removal is actually not part of this PR — I've updated the description to remove that bullet. The Score field will be removed in a separate follow-up PR to avoid overcomplicating this one even further.

pkg/vmcp/optimizer/optimizer.go
aponcedeleonch: The initial idea was to introduce logging in this PR, but I decided to separate it into its own PR to keep this one focused on the refactoring. I've updated the description to remove that bullet.

PR #3932, opened 2026-02-21
fix: validate volume mount source paths before spawning detached process

### LINKED ISSUE #2485: Volume mount errors not reported when source path doesn't exist
## Problem

When running an MCP server with a volume mount using `thv run --volume <source>:<target>`, if the source file or directory does not exist on the host, the workload fails to start but **no error is shown to the user** in the terminal. The command appears to succeed, but the workload immediately enters `error` state.

The actual error is only visible in the proxy log file (`~/Library/Application Support/toolhive/logs/<workload-name>.log`):

```
Error: failed to restart workload playwright: failed to set up workload: failed to create container: failed to create mcp container: failed to create container: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /host_mnt/Users/dan/fakedir
```

## Reproduction Steps

```bash
# Attempt to run with a non-existent source path
thv run --volume /nonexistent/path:/app playwright

# Command appears to succeed with output like:
# MCP server is running in the background (PID: 12345)
# Use 'thv stop playwright' to stop the server

# But checking status shows:
thv list
# NAME        STATUS   ... 
# playwright  error    ...

# Error only visible in logs:
cat ~/Library/Application\ Support/toolhive/logs/playwright.log
```

## Root Cause

The issue occurs due to the detached process architecture of `thv run`:

### 1. Detached Process Returns Before Container Creation

When `thv run` is executed without `--foreground`, `RunWorkloadDetached` spawns a detached process and returns immediately:

```go
// pkg/workloads/manager.go:435
if err := detachedCmd.Start(); err != nil {
    // Only checks if the detached process started, not if container succeeded
    return fmt.Errorf("failed to start detached process: %v", err)
}

// Line 452 - Logs success before container is actually created
logger.Infof("MCP server is running in the background (PID: %d)", detachedCmd.Process.Pid)
```

The detached process executes `thv restart <name> --foreground` (`manager.go:393`), and container creation happens inside that spawned process. Any errors during container creation are written to the log file, not returned to the original CLI process.

### 2. No Pre-flight Volume Path Validation

The volume mount processing in `pkg/runner/config_builder.go:909-976` (`processVolumeMounts`):
- Validates mount format (colon-separated syntax) via `permissions.MountDeclaration.Parse()` (`config_builder.go:948`)
- Checks for duplicate mount targets (`config_builder.go:954`)
- **Does NOT validate that source paths exist** on the host filesystem

The `Parse()` method in `pkg/permissions/profile.go:410` only parses the mount string format, not filesystem existence.

### 3. Docker Validates Too Late

Docker validates mount paths during container creation (`pkg/container/docker/client.go:1287-1341`), but this happens in the detached process after the CLI has already reported success to the user.

## Impact

This creates a poor user experience:
1. Users see a success message but the server never starts
2. No immediate indication of what went wrong
3. Users must manually check logs to discover the issue
4. Debugging requires knowledge of ToolHive's log file locations

## Proposed Solution

Add pre-flight validation of volume mount source paths before spawning the detached process. The validation should:

### Option 1: Validate in `processVolumeMounts()` (Recommended)

Add filesystem validation in `pkg/runner/config_builder.go:processVolumeMounts()` after parsing the mount:

```go
// After line 951 (mount.Parse())
source, target, err := mount.Parse()
if err != nil {
    return fmt.Errorf("invalid volume format: %s (%v)", volume, err)
}

// NEW: Validate source path exists (only for non-resource URIs)
if !strings.HasPrefix(source, "resource://") {
    // Convert to absolute path if relative
    absSource := source
    if !filepath.IsAbs(source) {
        cwd, err := os.Getwd()
        if err != nil {
            return fmt.Errorf("failed to resolve relative path %s: %w", source, err)
        }
        absSource = filepath.Join(cwd, source)
    }
    
    // Check if path exists
    if _, err := os.Stat(absSource); err != nil {
        if os.IsNotExist(err) {
            return fmt.Errorf("volume source path does not exist: %s", absSource)
        }
        return fmt.Errorf("failed to access volume source path %s: %w", absSource, err)
    }
}
```

**Pros:**
- Fails fast during config building
- Error is returned synchronously to CLI
- Works for both foreground and detached modes
- Prevents invalid config from being saved

**Cons:**
- May need special handling for Kubernetes operator (where paths are container-relative)
- Need to ensure validation only runs for CLI/Docker contexts

### Option 2: Validate in `RunWorkloadDetached()` Before Spawning Process

Add validation in `pkg/workloads/manager.go:RunWorkloadDetached()` before line 393:

```go
// Before spawning detached process, validate volume mounts
if err := validateVolumeMounts(runConfig.PermissionProfile); err != nil {
    return fmt.Errorf("invalid volume configuration: %w", err)
}
```

**Pros:**
- Doesn't affect config builder logic
- Isolated to workload execution path

**Cons:**
- More duplicated code
- Error occurs later in the process

### Recommendation

**Option 1** is preferred because:
1. It fails earlier in the process (during config building)
2. It prevents invalid configs from being created
3. The validation can be context-aware (skip for Kubernetes operator)
4. Aligns with the existing validation pattern in `config_builder.go`

### Additional Considerations

1. **Context-aware validation**: Use `buildContext` field to skip validation for Kubernetes operator:
   ```go
   if b.buildContext == BuildContextOperator {
       // Skip filesystem validation for K8s - paths are container-relative
       continue
   }
   ```

2. **Symbolic links**: Use `os.Stat()` (follows symlinks) rather than `os.Lstat()` to allow symlinked volumes

3. **Permissions**: Consider also checking read/write permissions on the source path to catch permission errors early

4. **Error messages**: Provide clear, actionable error messages that help users understand and fix the issue

## Implementation Notes

The fix should be implemented in these files:
- `pkg/runner/config_builder.go` - Add validation to `processVolumeMounts()` method
- Consider extracting validation into a separate method for testability
- Add unit tests covering:
  - Non-existent paths
  - Relative vs absolute paths
  - Symbolic links
  - Permission issues
  - Resource URIs (should skip validation)
  - Kubernetes context (should skip validation)

## Related Code References

- Volume flag parsing: `cmd/thv/app/run_flags.go:152-158`
- Volume processing: `pkg/runner/config_builder.go:909-976`
- Mount declaration parsing: `pkg/permissions/profile.go:410`
- Detached execution: `pkg/workloads/manager.go:364-456`
- Container creation: `pkg/container/docker/client.go:1287-1341`
- Mount configuration: `pkg/container/docker/client.go:621-709`

---

## Contributor Guidance

This is a good first issue! Here's some context to help you get started:

### Where to look
The volume mount processing happens in `pkg/runner/config_builder.go` in the `processVolumeMounts()` function (around line 900).

### What's happening
When a user specifies `--volume /nonexistent/path:/container/path`, the path is parsed and added to the permission profile without checking if the source path actually exists. The error only surfaces later when Docker tries to create the container, resulting in a confusing error message.

### Suggested approach
- Add a path existence check using standard Go `os.Stat()` after parsing the mount declaration
- Return a clear, user-friendly error message like "volume mount source path does not exist: /path"
- Consider handling relative paths (convert to absolute first)
- Add a test case in `pkg/runner/config_builder_test.go`

### How to test
```bash
task build
./bin/thv run --volume /nonexistent/path:/container some-mcp-server
# Should show a clear error about the path not existing
```

### Useful commands
- `task build` - Build the binary
- `task test` - Run unit tests
- `task lint` - Run linter

stakeswky:
## Summary

Fixes #2485

When running `thv run --volume <source>:<target>` with a non-existent source path, the command previously appeared to succeed but the workload would silently enter an error state. The actual error was only visible in the proxy log file.

## Changes

**`pkg/runner/config_builder.go`** — Added pre-flight validation in `processVolumeMounts()`:
- Checks that source paths exist on the host filesystem using `os.Stat()` (follows symlinks)
- Resolves relative paths to absolute before checking
- Skips validation for Kubernetes operator context (paths are container-relative)
- Skips validation for `resource://` URIs
- Returns a clear, actionable error message: `volume mount source path does not exist: /path`

**`pkg/runner/config_builder_test.go`** — Updated existing tests and added coverage:
- Existing volume mount tests now use `t.TempDir()` for real source paths instead of hardcoded `/host` paths
- Added test case for non-existent source path (expects error)

## Approach

Implemented **Option 1** from the issue (validate in `processVolumeMounts()`) because:
1. It fails at the earliest possible point (during config building)
2. It prevents invalid configs from being created
3. It's context-aware (skips for K8s operator)
4. It aligns with the existing validation pattern in `config_builder.go`

## Before / After

**Before:**
```bash
$ thv run --volume /nonexistent/path:/app playwright
MCP server is running in the background (PID: 12345)
# Silently fails — error only in log file
```

**After:**
```bash
$ thv run --volume /nonexistent/path:/app playwright
Error: volume mount source path does not exist: /nonexistent/path
```

### DIFF
--- a/pkg/runner/config_test.go
+++ b/pkg/runner/config_test.go
@@ -752,7 +752,8 @@ func TestRunConfigBuilder(t *testing.T) {
 	}
 	host := localhostStr
 	debug := true
-	volumes := []string{"/host:/container"}
+	tmpVolumeDir := t.TempDir()
+	volumes := []string{tmpVolumeDir + ":/container"}
 	secretsList := []string{"secret1,target=ENV_VAR1"}
 	authzConfigPath := "" // Empty to skip loading the authorization configuration
 	permissionProfile := permissions.ProfileNone
@@ -1304,9 +1305,11 @@ func TestRunConfigBuilder_VolumeProcessing(t *testing.T) {
 	runtime := &runtimemocks.MockRuntime{}
 	validator := &mockEnvVarValidator{}
 
+	tmpReadDir := t.TempDir()
+	tmpWriteDir := t.TempDir()
 	volumes := []string{
-		"/host/read:/container/read:ro",
-		"/host/write:/container/write",
+		tmpReadDir + ":/container/read:ro",
+		tmpWriteDir + ":/container/write",
 	}
 
 	config, err := NewRunConfigBuilder(context.Background(), nil, nil, validator,


pkg/runner/config_builder.go
```diff
@@ -1042,6 +1044,22 @@ func (b *runConfigBuilder) processVolumeMounts() error {
 			return fmt.Errorf("invalid volume format: %s (%w)", volume, err)
 		}
 
+		// Validate source path exists on the host filesystem (CLI context only).
```
JAORMX, line 1047: **Must-fix**: The `resource://` prefix check is incorrect. `MountDeclaration.Parse()` returns `source` as `scheme + "://" + resource` — where `scheme` is the actual URI scheme (e.g., `volume`, `tmpfs`), not the literal string `resource`. So a mount like `volume://mydata:/app` returns `source = "volume://mydata"`, which does **not** start with `resource://` and would incorrectly be stat'd on the filesystem.

`MountDeclaration` already has an `IsResourceURI()` method that handles this correctly. Suggested fix:

```suggestion
		if b.buildContext != BuildContextOperator && !mount.IsResourceURI() {
```

This also makes the code self-documenting — it reads as "skip if this is a resource URI" rather than checking a raw string prefix.

pkg/runner/config_builder.go
```diff
@@ -1042,6 +1044,22 @@ func (b *runConfigBuilder) processVolumeMounts() error {
...
+		// Skip for Kubernetes operator context where paths are container-relative,
+		// and for resource URIs which are not filesystem paths.
+		if b.buildContext != BuildContextOperator && !strings.HasPrefix(source, "resource://") {
+			absSource := source
+			if !filepath.IsAbs(absSource) {
+				absSource, err = filepath.Abs(absSource)
+				if err != nil {
+					return fmt.Errorf("failed to resolve volume mount source path %q: %w", source, err)
+				}
+			}
```
JAORMX, line 1057: **Must-fix**: All `os.Stat` errors are reported as "does not exist", but the path could exist and be inaccessible (e.g., permission denied). This would send users debugging in the wrong direction.

The issue itself suggested the correct pattern. Distinguish `os.IsNotExist` from other errors:

```suggestion
			if _, err := os.Stat(absSource); err != nil {
				if os.IsNotExist(err) {
					return fmt.Errorf("volume mount source path does not exist: %s", absSource)
				}
				return fmt.Errorf("failed to access volume mount source path %s: %w", absSource, err)
			}
```

Note: this also switches to `absSource` in the "does not exist" message (see next comment).

pkg/runner/config_builder.go
```diff
@@ -1042,6 +1044,22 @@ func (b *runConfigBuilder) processVolumeMounts() error {
...
+		// Validate source path exists on the host filesystem (CLI context only).
+		// Skip for Kubernetes operator context where paths are container-relative,
+		// and for resource URIs which are not filesystem paths.
+		if b.buildContext != BuildContextOperator && !strings.HasPrefix(source, "resource://") {
+			absSource := source
+			if !filepath.IsAbs(absSource) {
+				absSource, err = filepath.Abs(absSource)
+				if err != nil {
+					return fmt.Errorf("failed to resolve volume mount source path %q: %w", source, err)
+				}
```
JAORMX, line 1056: **Should-fix**: When a relative path is used (e.g., `./data:/app`), the error shows the original relative path (`./data`) rather than the resolved absolute path. If the user's working directory isn't what they expect, this makes debugging harder.

Consider showing `absSource` instead of `source` in the error message so the user can see exactly which path was checked.

pkg/runner/config_builder_test.go
```diff
@@ -312,6 +317,14 @@ func TestRunConfigBuilder_Build_WithVolumeMounts(t *testing.T) {
 			},
 			expectError: true,
 		},
+		{
+			name: "Non-existent source path",
+			builderOptions: []RunConfigBuilderOption{
+				WithVolumes([]string{"/nonexistent/path/that/does/not/exist:/container"}),
+				WithPermissionProfile(permissions.BuiltinNoneProfile()),
```
JAORMX, line 324: **Should-fix**: A couple of test cases are missing that would strengthen confidence in the guard conditions:

1. **Operator context with non-existent path**: A test using `NewOperatorRunConfigBuilder` with `WithVolumes([]string{"/nonexistent/path:/container"})` that asserts *no error*. This proves the `BuildContextOperator` skip actually works. Without this, a regression that removes the operator guard would go undetected.

2. **Resource URI skips validation**: A test with a resource URI volume (e.g., `volume://mydata:/app`) that asserts *no error*. This would have caught the `resource://` prefix bug noted in the other comment.

JAORMX:
Hey @stakeswky, thanks for the contribution! Validating volume mount source paths is a good catch and something we should have.

I'm closing this for now since there hasn't been activity since the review. The main things that still need fixing are using `IsResourceURI()` instead of the `strings.HasPrefix` check, and properly distinguishing `ENOENT` from other `os.Stat` failures. Not a ton of work, but it's been sitting a while.

If you want to pick this up again, feel free to reopen or start a new PR. The core idea is good!

PR #3953, opened 2026-02-25
Handle embedded ResourceContents in vmcp bridge

### LINKED ISSUE #3952: vmcp bridge silently drops embedded ResourceContents from tool results
## Summary

The vmcp bridge in `pkg/vmcp/server/adapter/handler_factory.go` silently discards embedded `ResourceContents` from MCP tool results, replacing them with an empty string. This causes tools that return file content via `NewToolResultResource` (notably `github_get_file_contents` from the GitHub MCP server) to appear to succeed but return no usable content to the model.

## Root cause

In `convertToMCPContent()`:

```go
case "resource":
    // Handle embedded resources if needed
    // For now, convert to text
    slog.Warn("converting resource content to empty text - embedded resources not yet supported")
    return mcp.NewTextContent("")
```

When a tool returns a `mcp.ResourceContents` embedded in a `CallToolResult` (via `utils.NewToolResultResource`), the bridge converts it to an empty `TextContent`. The actual file content — present in `ResourceContents.Text` — is thrown away.

## Observed behaviour

`github_get_file_contents` returns two content items:
1. A `TextContent` with the human-readable description: `"successfully downloaded text file (SHA: <sha>)"`
2. A `ResourceContents` with the actual file text in `ResourceContents.Text` and MIME type

The model receives item 1 (the description) but item 2 is replaced with `""`. The model therefore sees only `"successfully downloaded text file"` with no actual file content — it looks like success but delivers nothing.

## Expected behaviour

For `ResourceContents` with a non-empty `Text` field, the bridge should convert to `mcp.NewTextContent(content.Text)` (or preserve it as a proper embedded resource if the downstream client supports it). For binary `Blob` content, base64 encoding or a resource link would be appropriate.

## Suggested fix

```go
case "resource":
    if content.Text != "" {
        return mcp.NewTextContent(content.Text)
    }
    // Binary blob — could encode as base64 or return a resource link
    slog.Warn("binary embedded resource content not yet supported", "mime_type", content.MimeType)
    return mcp.NewTextContent("")
```

## Impact

Any MCP tool that uses `NewToolResultResource` to return text file content (the standard pattern in the GitHub MCP server for `get_file_contents`) is silently broken when proxied through the vmcp bridge. This affects reading source files from private repositories via the GitHub MCP tool.

JAORMX:
## Summary

- Fix vmcp bridge silently discarding embedded `ResourceContents` from MCP tool results, replacing them with empty strings
- Add bidirectional `EmbeddedResource` handling in content conversion (both `TextResourceContents` and `BlobResourceContents`)
- Change empty-resource fallback to return `EmbeddedResource` instead of `TextContent` to preserve `type: "resource"` semantics

This broke tools like `github_get_file_contents` that return file content via `NewToolResultResource` — the model would see "successfully downloaded text file" but receive no actual file content.

Closes #3952

## Test plan

- [x] Unit tests for `ConvertMCPContent` (MCP SDK → vmcp): text resource, blob resource, empty URI/MimeType
- [x] Unit tests for `convertToMCPContent` (vmcp → MCP SDK): text resource, blob resource, text+data precedence, empty URI/MimeType, empty fallback
- [x] Full `pkg/vmcp/...` test suite passes
- [x] Lint clean

### DIFF
--- a/pkg/vmcp/conversion/content.go
+++ b/pkg/vmcp/conversion/content.go
@@ -29,6 +29,16 @@ func ConvertMCPContent(content mcp.Content) vmcp.Content {
 	if audio, ok := mcp.AsAudioContent(content); ok {
 		return vmcp.Content{Type: "audio", Data: audio.Data, MimeType: audio.MIMEType}
 	}
+	if res, ok := mcp.AsEmbeddedResource(content); ok {
+		if textRes, ok := mcp.AsTextResourceContents(res.Resource); ok {
+			return vmcp.Content{Type: "resource", Text: textRes.Text, URI: textRes.URI, MimeType: textRes.MIMEType}
+		}
+		if blobRes, ok := mcp.AsBlobResourceContents(res.Resource); ok {
+			return vmcp.Content{Type: "resource", Data: blobRes.Blob, URI: blobRes.URI, MimeType: blobRes.MIMEType}
+		}
+		slog.Debug("Embedded resource has unknown resource contents type", "type", fmt.Sprintf("%T", res.Resource))
+		return vmcp.Content{Type: "resource"}
+	}
 	slog.Debug("Encountered unknown MCP content type", "type", fmt.Sprintf("%T", content))
 	return vmcp.Content{Type: "unknown"}
 }
--- a/pkg/vmcp/conversion/conversion_test.go
+++ b/pkg/vmcp/conversion/conversion_test.go
@@ -185,6 +185,31 @@ func TestConvertMCPContent(t *testing.T) {
 			input: mcp.NewAudioContent("base64audiodata", "audio/mpeg"),
 			want:  vmcp.Content{Type: "audio", Data: "base64audiodata", MimeType: "audio/mpeg"},
 		},
+		{
+			name: "embedded resource with text content",
+			input: mcp.NewEmbeddedResource(mcp.TextResourceContents{
+				URI:      "file://readme.md",
+				MIMEType: "text/markdown",
+				Text:     "# Hello World",
+			}),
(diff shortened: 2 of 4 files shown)


pkg/vmcp/conversion/content.go
```diff
@@ -29,6 +29,16 @@ func ConvertMCPContent(content mcp.Content) vmcp.Content {
...
 	}
+	if res, ok := mcp.AsEmbeddedResource(content); ok {
+		if textRes, ok := mcp.AsTextResourceContents(res.Resource); ok {
+			return vmcp.Content{Type: "resource", Text: textRes.Text, URI: textRes.URI, MimeType: textRes.MIMEType}
+		}
+		if blobRes, ok := mcp.AsBlobResourceContents(res.Resource); ok {
+			return vmcp.Content{Type: "resource", Data: blobRes.Blob, URI: blobRes.URI, MimeType: blobRes.MIMEType}
+		}
+		slog.Debug("Embedded resource has unknown resource contents type", "type", fmt.Sprintf("%T", res.Resource))
+		return vmcp.Content{Type: "resource"}
```
jhrozek, line 40: This returns `{Type: "resource"}` with empty URI/Text/Data, which when it hits `convertToMCPContent` in handler_factory.go ends up creating a `TextResourceContents` with an empty `uri`. The MCP spec marks `uri` as required on resource contents (both spec versions).

Probably not a real issue since the SDK only has `TextResourceContents` and `BlobResourceContents` today, but we could avoid the edge case by returning `"unknown"` here instead — that way it goes through the same fallback as other unrecognized types:

```suggestion
		return vmcp.Content{Type: "unknown"}
```

PR #3961, opened 2026-03-02
Fix operator e2e CI failures and spurious deployment updates

aponcedeleonch:
## Summary

- Fix operator e2e chainsaw tests failing across all K8s versions due to yardstick image pull timeout
- Disable containerd image store before Kind cluster creation (Docker 29+ workaround matching `test-e2e-lifecycle.yml`), with an inline comment referencing the upstream KIND issue
- Pre-pull and load `ghcr.io/stackloklabs/yardstick/yardstick-server:1.1.1` into the Kind cluster before running chainsaw tests
- Replace `reflect.DeepEqual` with `equality.Semantic.DeepEqual` in the MCPServer controller when comparing Env vars and resource requirements, preventing spurious deployment updates caused by nil-vs-empty differences
- Fix infinite deployment update loop caused by annotation fight between the controller and K8s deployment controller: use subset check for annotation comparison and merge annotations instead of replacing them, across MCPServer, MCPRemoteProxy, and VirtualMCPServer controllers

## Test plan

- [ ] All three `E2E Tests Operator` matrix jobs pass in the `operator-ci.yml` workflow
- [ ] Specifically confirm the `apps/v1/Deployment @ test-namespace/yardstick` assertion succeeds within the timeout
- [ ] Verify no unnecessary deployment rollouts occur during reconciliation
- [ ] Confirm deployments reach `availableReplicas` without getting stuck in a reconcile loop

### DIFF
--- a/cmd/thv-operator/controllers/mcpserver_controller.go
+++ b/cmd/thv-operator/controllers/mcpserver_controller.go
@@ -1634,7 +1634,7 @@ func (r *MCPServerReconciler) deploymentNeedsUpdate(
 		return true
 	}
 
-	if !mapIsSubset(expectedAnnotations, deployment.Annotations) {
+	if !ctrlutil.MapIsSubset(expectedAnnotations, deployment.Annotations) {
 		return true
 	}
 
@@ -1657,17 +1657,6 @@ func (r *MCPServerReconciler) deploymentNeedsUpdate(
 	}
 
 	return false
-}
-
-// mapIsSubset returns true if every key-value pair in subset exists in superset.
-// Extra keys in superset (e.g. K8s-managed annotations) are ignored.
-func mapIsSubset(subset, superset map[string]string) bool {
-	for k, v := range subset {
-		if sv, ok := superset[k]; !ok || sv != v {
-			return false
-		}
-	}
-	return true
 }
 
 // serviceNeedsUpdate checks if the service needs to be updated
--- a/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
+++ b/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
@@ -768,7 +768,7 @@ func (*MCPRemoteProxyReconciler) deploymentMetadataNeedsUpdate(
 		return true
 	}
 
-	if !mapIsSubset(expectedAnnotations, deployment.Annotations) {
+	if !ctrlutil.MapIsSubset(expectedAnnotations, deployment.Annotations) {
 		return true
 	}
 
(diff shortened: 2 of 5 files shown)


cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -1659,6 +1659,17 @@ func (r *MCPServerReconciler) deploymentNeedsUpdate(
 	return false
 }
 
+// mapIsSubset returns true if every key-value pair in subset exists in superset.
+// Extra keys in superset (e.g. K8s-managed annotations) are ignored.
+func mapIsSubset(subset, superset map[string]string) bool {
```
JAORMX: Minor: `mapIsSubset` is now used from multiple controllers. Consider moving it to a shared helper (e.g., `cmd/thv-operator/pkg/controllerutil`) so it’s easier to discover/reuse.

JAORMX (commented):
Inline note about helper placement.

cmd/thv-operator/controllers/mcpserver_controller.go
reyortiz3: nit: `if len(subset) > len(superset) { return false }`

.github/workflows/operator-ci.yml
```diff
@@ -195,6 +208,11 @@ jobs:
         cloud_provider: true
         node_image: ${{ matrix.version }}
 
+    - name: Pre-load test images
```
ChrisJBurns, line 211: Curious, do we get much of a decrease in test times with this? As the image itself is pretty small, so I can't imagine it stuck pulling the image for too long?

cmd/thv-operator/controllers/mcpserver_controller.go
aponcedeleonch: Moved mapIsSubset to a dedicated controllers/helpers.go file so it's easier to discover and reuse across controllers. Thanks for the suggestion!

.github/workflows/operator-ci.yml
aponcedeleonch: yeah probably not although it also doesn't hurt. Actually this was part of the initial fix since I thought that pulling the image was actually the culprit of the timeout problem. I would vote for keeping the step unless you feel very strongly about it

JAORMX (dismissed):

cmd/thv-operator/controllers/mcpserver_controller.go
aponcedeleonch: Scratch that. I moved the function to `cmd/thv-operator/pkg/controllerutil/maps.go`

aponcedeleonch: true that! Added the condition along with some unit tests for the function in `cmd/thv-operator/pkg/controllerutil/maps_test.go`

PR #3962, opened 2026-03-02
feat(authserver): add OAuth 2.0 client_credentials grant support for M2M clients

RobertWi:
## Summary

Add `client_credentials` grant type support to ToolHive's built-in OAuth authorization server, enabling machine-to-machine (M2M) authentication as recommended by the [MCP specification (2025-03-26)](https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization#oauth-grant-types).

## Motivation

The MCP spec explicitly recommends `client_credentials` for non-human clients:

> MCP servers **SHOULD** support the OAuth grant types that best align with the intended audience. For instance:
> 1. Authorization Code: useful when the client is acting on behalf of a (human) end user.
> 2. Client Credentials: the client is another application (not a human)

PR #3425 correctly blocked `client_credentials` in DCR because the server couldn't fulfill it at the time. This PR implements the capability, then updates the allowlist.

### Production code (5 files, ~100 lines)

- **`server_impl.go`**: Wire `OAuth2ClientCredentialsGrantFactory` into fosite composition (1 line)
- **`registration/dcr.go`**: Support confidential client registration — `redirect_uris` optional for `client_credentials`-only clients, `client_secret_basic`/`client_secret_post` auth methods accepted, secret generation
- **`handlers/dcr.go`**: Generate and return `client_secret` for confidential clients in DCR response
- **`handlers/token.go`**: Populate session subject with client ID for `client_credentials` grants so JWT has meaningful `sub` claim
- **`handlers/discovery.go`**: Advertise `client_credentials` grant and `client_secret_basic`/`client_secret_post` in OAuth AS + OIDC discovery metadata

### Storage / Authorization

No changes needed. The existing storage interface already satisfies fosite's requirements for `client_credentials`. Cedar authorization naturally handles `Client::` principals derived from the `sub` claim.

### Tests (5 files, ~340 lines)

- Updated DCR validation tests: flipped `client_credentials` rejection cases to success, added 6 new confidential client test cases
- Updated token handler tests: fixed unsupported grant type test, added 3 new `client_credentials` tests (success, wrong secret, resource parameter)
- Updated discovery tests: added `client_credentials` + auth method assertions
- Added 3 integration tests: basic flow with JWT claim validation, RFC 8707 audience binding, wrong secret rejection

## Test Evidence

Full test suite with race detector — all 10 authserver packages pass:

```
$ go test ./pkg/authserver/... -count=1 -race

ok  	github.com/stacklok/toolhive/pkg/authserver	5.844s
ok  	github.com/stacklok/toolhive/pkg/authserver/runner	1.350s
ok  	github.com/stacklok/toolhive/pkg/authserver/server	1.666s
ok  	github.com/stacklok/toolhive/pkg/authserver/server/crypto	2.011s
ok  	github.com/stacklok/toolhive/pkg/authserver/server/handlers	7.307s
ok  	github.com/stacklok/toolhive/pkg/authserver/server/keys	1.360s
ok  	github.com/stacklok/toolhive/pkg/authserver/server/registration	3.573s
ok  	github.com/stacklok/toolhive/pkg/authserver/server/session	1.178s
ok  	github.com/stacklok/toolhive/pkg/authserver/storage	3.477s
ok  	github.com/stacklok/toolhive/pkg/authserver/upstream	2.098s
```

New tests added by this PR — all pass:

```
--- PASS: TestIntegration_ClientCredentials_BasicFlow (0.20s)
--- PASS: TestIntegration_ClientCredentials_WithAudience (0.20s)
--- PASS: TestIntegration_ClientCredentials_WrongSecret (0.34s)
--- PASS: TestTokenHandler_ClientCredentials_Success (0.17s)
--- PASS: TestTokenHandler_ClientCredentials_WrongSecret (0.15s)
--- PASS: TestTokenHandler_ClientCredentials_WithResource (0.15s)
```

`go vet ./pkg/authserver/...` — clean, no findings.

## Backward Compatibility

Fully backward compatible. Existing public client registrations and `authorization_code` flows are unaffected:
- `defaultGrantTypes` for empty DCR requests remain `["authorization_code", "refresh_token"]`
- Public client validation logic is unchanged (just moved into an `else` branch)
- All existing tests pass without modification (only the two tests that explicitly tested `client_credentials` rejection were updated)

## Diff Stats

```
 10 files changed, 436 insertions(+), 65 deletions(-)
```

### DIFF
--- a/pkg/authserver/integration_test.go
+++ b/pkg/authserver/integration_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/ory/fosite"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/crypto/bcrypt"
 
 	servercrypto "github.com/stacklok/toolhive/pkg/authserver/server/crypto"
 	"github.com/stacklok/toolhive/pkg/authserver/server/keys"
@@ -35,6 +36,9 @@ const (
 	testRedirectURI = "http://localhost:8080/callback"
 	testIssuer      = "http://localhost"
 	testAudience    = "https://mcp.example.com"
+
+	testConfidentialClientID     = "test-confidential-client"
+	testConfidentialClientSecret = "test-confidential-secret"
 
 	// testAccessTokenLifetime is the configured access token lifetime in setupTestServer.
 	testAccessTokenLifetime = time.Hour
@@ -51,6 +55,7 @@ type testServerOptions struct {
 	upstream            upstream.OAuth2Provider
 	scopes              []string
 	accessTokenLifespan time.Duration
+	confidentialClient  bool
 }
 
 // testServerOption is a functional option for test server setup.
@@ -74,6 +79,13 @@ func withScopes(scopes []string) testServerOption {
 func withAccessTokenLifespan(d time.Duration) testServerOption {
 	return func(opts *testServerOptions) {
 		opts.accessTokenLifespan = d
+	}
+}
+
+// withConfidentialClient registers a confidential client for client_credentials testing.
+func withConfidentialClient() testServerOption {
+	return func(opts *testServerOptions) {
+		opts.confidentialClient = true
 	}
 }
 
@@ -137,6 +149,22 @@ func setupTestServer(t *testing.T, opts ...testServerOption) *testServer {
 		Public:        true,
 	})
 	require.NoError(t, err)
+
+	// Optionally register a confidential client for client_credentials testing
+	if options.confidentialClient {
+		hashedSecret, err := bcrypt.GenerateFromPassword([]byte(testConfidentialClientSecret), bcrypt.DefaultCost)
+		require.NoError(t, err)
+		err = stor.RegisterClient(ctx, &fosite.DefaultClient{
+			ID:            testConfidentialClientID,
+			Secret:        hashedSecret,
+			RedirectURIs:  nil,
+			ResponseTypes: nil,
+			GrantTypes:    []string{"client_credentials"},
+			Scopes:        []string{"openid"},
+			Public:        false,
+		})
+		require.NoError(t, err)
+	}
 
 	// 5. Build upstream config for newServer
 	// When no upstream is provided, use a dummy config that satisfies validation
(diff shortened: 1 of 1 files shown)


JAORMX:
Thanks for putting this together, @RobertWi. M2M auth is a real gap and the code is well-structured. However, after reviewing against the current MCP spec and OAuth security best practices, I have concerns about the overall approach.

The spec reference is outdated. The PR cites the 2025-03-26 MCP spec, but the latest (2025-11-25) moved client_credentials to an optional draft extension in a separate repo. See SEP-1046 for background.

DCR + client_credentials contradicts the extension spec. The extension explicitly states: "Dynamic Client Registration is not used in this flow. All credentials must be pre-registered through out-of-band administrative channels." This PR does the opposite: anyone who can reach the unauthenticated DCR endpoint can register a confidential client and immediately get tokens with zero human involvement. That's a fundamental change in security posture.

No principal differentiation. M2M tokens get sub=<client_id> (UUID), same format as user tokens. Cedar policies can't distinguish machine from human principals, so existing permit(principal, ...) rules would authorize both identically.

Mixed grant types edge case. Registering with ["authorization_code", "client_credentials"] creates a public client (no secret) that can never actually use client_credentials. Fosite blocks this at token time, but it creates dead-letter configs. RFC 7591 Section 5 explicitly warns against combining these.

If we want this (and I think we eventually should!), the approach should align with the extension spec:
1. Pre-registered confidential clients only (admin-created, not through unauthenticated DCR)
2. Reference the actual extension spec, not the outdated core spec
3. Differentiate M2M principals in Cedar
4. Consider waiting for the extension to move past draft status

Happy to discuss the right path forward!

RobertWi:
Thanks for the detailed review @JAORMX — sorry for the slow reply, needed to step back and do more homework on this.

You're right that my approach was off. I was trying to solve a practical problem — getting AI agents authenticated to MCP servers — and jumped to implementation without fully understanding where the spec landed. The unauthenticated DCR path was the wrong call, especially for the K8s case.

I've since dug into SEP-1046 and see it was accepted and landed as a draft extension in the `ext-auth` repo. So it seems like an agreed direction, though I'm not plugged into when (or if) it'll move past draft. That's part of my uncertainty here.

From what I can tell, there's currently no M2M auth path at all in ToolHive — agents simply can't authenticate programmatically. For my use case (fleet of agents in K8s), that's a blocker.

Given your feedback, is it feasible to move forward with something that aligns with the extension spec? I'm thinking along these lines:

- **Operator-provisioned credentials** — instead of agents registering themselves via DCR, the operator creates and injects client credentials when deploying a workload. Similar to how K8s handles ServiceAccount tokens — the workload doesn't request its own identity, the platform provisions it. The operator already manages the lifecycle of MCP servers, so it's a natural place to also manage which agents get credentials to talk to them.
- **Separate identity model for machine principals** — so Cedar can structurally tell human and machine apart, not just a flag on the same token format.
- **Support JWT assertions and client secrets** as the extension spec describes.

Happy to rework this or close it — just want to understand if there's a viable path forward.

jhrozek:
You're referring to is https://github.com/modelcontextprotocol/ext-auth/blob/main/specification/draft/oauth-client-credentials.mdx right?

I'm curious to know how does your deployment mental model look like? In general when it comes to the MCP side of things, toolhive only mandates that there is a bearer token that authenticates the caller and that it can be verified against a known issuer. The token doesn't have to come from the toolhive AS.

That extension doesn't change that, it's just about how the client retrieves the credentials and that's really up to the client and the authorization server. Depending on the authorization server you use it might already be supported (Keycloak, Entra and Okta already do). So I wonder if any changes on the toolhive side are needed at all or if the setup is more about wiring an existing external AS and provisioning clients?

About toolhive's AS - I'm not sure adding the client_credentials flow there is a good idea or maybe I'm not understanding how it would work. One change would be to run the AS standalone without proxying to upstreams (right now the AS is meant to function as a proxy). The other change would be to create some kind of a management channel to pre-create the clients. All in all I don't see the value in extending the toolhive AS this way, honestly.

jhrozek:
>  One change would be to run the AS standalone without proxying to upstreams (right now the AS is meant to function as a proxy). The other change would be to create some kind of a management channel to pre-create the clients. All in all I don't see the value in extending the toolhive AS this way, honestly.

the standalone AS use-case is not as problematic of a change if we see any use for it but the client management API is a big pull. I'd rather just skip ahead to using a SPIFFE bridge where you just trust the SPIFFE trust domain instead of having to manage clients yourself.

JAORMX:
Hey @RobertWi, thanks for the thoughtful follow-up and for digging into SEP-1046. The conversation here with @jhrozek has been really useful.

So... I think at this point we're past what a single PR can answer. There are real design questions here: whether we extend the toolhive AS at all, whether an external AS (Keycloak, Entra, Okta) is the right path, whether a SPIFFE bridge makes more sense, and how we differentiate machine principals in Cedar. These are architectural decisions that affect the whole auth story.

I'd suggest we move this to an RFC. That way we can lay out the options, get input from more folks, and land on an approach before writing code. Would you be up for drafting one? You clearly have a good handle on the problem space. We can use the discussion here as a starting point.

Note that I'm not closing this PR... just suggesting we figure out the "what" before the "how." Happy to help shape the RFC if useful!

RobertWi:
After sitting with all your feedback, I think this should be closed — wrong solution to a real problem. Let me walk through how I ended up here.

1. Deployed `VirtualMCPServer` with `outgoingAuth: source: discovered` against OIDC-protected backends. Everything came back 401 — vMCP had no way to get tokens for its own backends.

2. Dug into `MCPExternalAuthConfig` looking for a way to wire client_credentials. None of the existing strategies fit — `tokenExchange` needs an incoming token, `bearerToken` is static, etc.

3. Found the TODO in `validator.go:192` — `"client_credentials"` listed as a future strategy.

4. That's where I made the wrong call. Instead of requesting the missing *outgoing auth strategy*, I jumped to adding `client_credentials` to the *authserver* — a different component that wouldn't have actually solved my problem.

@JAORMX @jhrozek — you were right to push back.

Thanks for the time you put into reviewing this. Closing.

PR #3963, opened 2026-03-02
DO NOT MERGE: add stdio e2e debug dumps

JAORMX:
## Summary
- replace stdio deployment readiness assertion with a wait script that dumps deployment/pod/rs logs on failure
- add the same debug dump for the stdio+streamable-http scenario

## Testing
- not run (debug-only change for CI diagnostics)


aponcedeleonch:
@JAORMX do we still need this? I guess it was needed to debug the CI but now we should be ok

Issue #3867, opened 2026-02-18
[vMCP] Implement session token hash binding (blocks Phase 2 production rollout)


yrobla:
Bind each session to the authentication token that created it to prevent a stolen session ID from being reused with different credentials. This is a **blocking requirement** — the `sessionManagementV2` flag must not be enabled in any production environment until this is implemented.

**Implementation**:
- During `CreateSession()`: compute `SHA256(bearerToken)` and store it in the session
- On each subsequent request to that session: recompute the hash of the current auth token and compare with the stored value
- On mismatch: reject with HTTP 401, error message "session authentication mismatch", and terminate the session

### Acceptance Criteria

- [ ] `CreateSession()` stores a `SHA256` hash of the bearer token in the session at creation time
- [ ] Every subsequent request validates the current token hash against the stored value
- [ ] A request with a mismatched token receives HTTP 401 with a clear error message
- [ ] The session is terminated immediately on a mismatch
- [ ] Anonymous sessions (no bearer token) are handled gracefully without errors
- [ ] The raw token is never stored or logged — only the hash
- [ ] Unit tests cover: valid token passes, mismatched token rejected, session terminated on mismatch, anonymous session flow

PR #3964, opened 2026-03-02
Implement HMAC-SHA256 token binding for session security

yrobla:
Implements session-level token binding using HMAC-SHA256 with per-session
salts to prevent session hijacking attacks. Each session is cryptographically
bound to the authentication token used to create it.

Security model:
- Token hashes use HMAC-SHA256 with a server-managed secret
- Each session has a unique random salt (16 bytes)
- Validation uses constant-time comparison to prevent timing attacks
- Each request independently validates the caller token against the session owner

On token mismatch, the request is rejected with an unauthorized error.
The session remains active but subsequent requests will also be rejected
unless the correct token is provided, effectively making the session unusable
to attackers.

Configuration:
- Set VMCP_SESSION_HMAC_SECRET environment variable for production
- Falls back to insecure default for development/testing (with warning)

Closes: #3867

## Large PR Justification

This is an atomic PR that covers single functionality + tests. It cannot be splitted

### DIFF
--- a/cmd/vmcp/README.md
+++ b/cmd/vmcp/README.md
@@ -237,7 +237,7 @@ spec:
               key: hmac-secret
 ```
 
-**Note**: If `VMCP_SESSION_HMAC_SECRET` is not set, a default insecure secret is used. This is acceptable for development/testing but **NOT recommended for production**.
+**Note**: Kubernetes deployments **require** `VMCP_SESSION_HMAC_SECRET` to be set (the server will fail to start without it). For non-Kubernetes environments (local development/testing), a default insecure secret is used as a fallback, but this is **NOT recommended for production**.
 
 ## Tool Aggregation & Conflict Resolution
 
--- a/cmd/vmcp/app/commands.go
+++ b/cmd/vmcp/app/commands.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/stacklok/toolhive-core/env"
 	"github.com/stacklok/toolhive/pkg/audit"
+	"github.com/stacklok/toolhive/pkg/container/runtime"
 	"github.com/stacklok/toolhive/pkg/groups"
 	"github.com/stacklok/toolhive/pkg/telemetry"
 	"github.com/stacklok/toolhive/pkg/vmcp"
@@ -287,6 +288,53 @@ func discoverBackends(
 	return backends, backendClient, outgoingRegistry, nil
 }
 
+// createSessionFactory creates a MultiSessionFactory with HMAC-SHA256 token binding.
+// The HMAC secret is read from the VMCP_SESSION_HMAC_SECRET environment variable.
+//
+// Behavior:
+//   - If VMCP_SESSION_HMAC_SECRET is set: validates length and creates factory with the secret
+//   - If running in Kubernetes without secret: returns error (production safety requirement)
+//   - Otherwise: logs warning and creates factory with default insecure secret
+//
+// Returns an error only when running in Kubernetes without a secret (fail-fast for production).
+func createSessionFactory(outgoingRegistry vmcpauth.OutgoingAuthRegistry) (vmcpsession.MultiSessionFactory, error) {
+	const (
+		envKey                  = "VMCP_SESSION_HMAC_SECRET"
+		minRecommendedSecretLen = 32
+	)
+
+	hmacSecret := os.Getenv(envKey)
+
+	if hmacSecret != "" {
+		// Validate secret length (warn if too short, but still use it)
+		if secretLen := len(hmacSecret); secretLen < minRecommendedSecretLen {
+			// G706: Safe - only logging integer length, not the secret itself
+			slog.Warn( //nolint:gosec
+				"HMAC secret is shorter than recommended length - consider using a longer secret",
+				"actual_length", secretLen,
+				"recommended_length", minRecommendedSecretLen,
+			)
+		}
+		slog.Info("using HMAC secret from VMCP_SESSION_HMAC_SECRET environment variable for session token binding")
+		return vmcpsession.NewSessionFactory(
+			outgoingRegistry,
+			vmcpsession.WithHMACSecret([]byte(hmacSecret)),
+		), nil
+	}
+
+	// No secret provided - check if we're in Kubernetes (production environment)
+	if runtime.IsKubernetesRuntime() {
+		return nil, fmt.Errorf(
+			"VMCP_SESSION_HMAC_SECRET environment variable is required when running in Kubernetes. " +
+				"Generate a secure secret with: openssl rand -base64 32",
+		)
+	}
+
+	// Development mode: use default insecure secret with warning
+	slog.Warn("VMCP_SESSION_HMAC_SECRET not set - using default insecure secret (NOT recommended for production)")
+	return vmcpsession.NewSessionFactory(outgoingRegistry), nil
+}
+
 // runServe implements the serve command logic
 //
 //nolint:gocyclo // Complexity from server initialization and configuration is acceptable
@@ -474,31 +522,13 @@ func runServe(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("failed to validate optimizer config: %w", err)
 	}
 
-	// Create session factory for SessionManagementV2 with HMAC-SHA256 token binding
-	// The HMAC secret should be provided via VMCP_SESSION_HMAC_SECRET environment variable.
-	// If not set, a default insecure secret is used (NOT recommended for production).
-	const minRecommendedSecretLen = 32
+	// Create session factory only when SessionManagementV2 is enabled
 	var sessionFactory vmcpsession.MultiSessionFactory
-	hmacSecretEnv := os.Getenv("VMCP_SESSION_HMAC_SECRET")
-	if hmacSecretEnv != "" {
-		// Validate secret length
-		secretLen := len(hmacSecretEnv)
-		if secretLen < minRecommendedSecretLen {
-			// G706: Safe - only logging integer length, not the secret itself
-			slog.Warn( //nolint:gosec
-				"HMAC secret is shorter than recommended length - consider using a longer secret",
-				"actual_length", secretLen,
-				"recommended_length", minRecommendedSecretLen,
-			)
+	if cfg.Operational.SessionManagementV2 {
+		sessionFactory, err = createSessionFactory(outgoingRegistry)
+		if err != nil {
+			return err
 		}
-		slog.Info("using HMAC secret from VMCP_SESSION_HMAC_SECRET environment variable for session token binding")
-		sessionFactory = vmcpsession.NewSessionFactory(
-			outgoingRegistry,
-			vmcpsession.WithHMACSecret([]byte(hmacSecretEnv)),
-		)
-	} else {
-		slog.Warn("VMCP_SESSION_HMAC_SECRET not set - using default insecure secret (NOT recommended for production)")
-		sessionFactory = vmcpsession.NewSessionFactory(outgoingRegistry)
 	}
 
 	serverCfg := &vmcpserver.Config{
@@ -516,6 +546,7 @@ func runServe(cmd *cobra.Command, _ []string) error {
 		Watcher:                 backendWatcher,
 		StatusReporter:          statusReporter,
 		OptimizerConfig:         optCfg,
+		SessionManagementV2:     cfg.Operational.SessionManagementV2,
 		SessionFactory:          sessionFactory,
 	}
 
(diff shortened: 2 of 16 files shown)


pkg/vmcp/server/token_binding.go
```diff
@@ -0,0 +1,191 @@
...
+//
+//   - Sessions whose metadata contains no MetadataKeyTokenHash entry pass
+//     through unchanged; this preserves backward compatibility with sessions
+//     created before token binding was introduced.
+func tokenBindingMiddleware(
+	sessionStorage *transportsession.Manager,
+	terminator sessionTerminator,
+) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
jerm-dro: Blocker: something seems off here - do we need two nested function declarations? I would expect just one.

pkg/vmcp/server/token_binding.go
```diff
@@ -0,0 +1,191 @@
...
+				tokenErrorCode = ""
+			case auth.ErrAuthHeaderMissing:
+				tokenErrorCode = "missing"
+			case auth.ErrInvalidAuthHeaderFormat:
+				tokenErrorCode = "invalid_format"
+			case auth.ErrEmptyBearerToken:
+				tokenErrorCode = "empty_token"
+			default:
+				tokenErrorCode = "unknown"
+			}
```
jerm-dro: Suggestion: this seems like a good candidate to factor out to reduce complexity.

pkg/vmcp/server/token_binding.go
```diff
@@ -0,0 +1,191 @@
...
+			// Use stack-allocated arrays to avoid per-request heap allocations.
+			const normalizedLen = 64 // hex-encoded SHA256 length
+			var normalizedStored, normalizedCurrent [normalizedLen]byte
+			copy(normalizedStored[:], storedHash)
+			copy(normalizedCurrent[:], currentHash)
+			cmp := subtle.ConstantTimeCompare(normalizedStored[:], normalizedCurrent[:])
+			// G115: Safe conversion - storedHash/currentHash are either "" (0 bytes)
+			// or hex-encoded SHA256 (64 bytes), both well within int32 range.
+			lengthEq := subtle.ConstantTimeEq(int32(len(storedHash)), int32(len(currentHash))) //nolint:gosec
+			hashesMatch := (cmp & lengthEq) == 1
```
jerm-dro: Suggestion: this also could be factored out into a common security or hashcmp package.

pkg/vmcp/server/token_binding.go
```diff
@@ -0,0 +1,191 @@
...
+//     e.g. initialize) pass through — token binding only applies to established
+//     sessions.
+//
+//   - Sessions whose metadata contains no MetadataKeyTokenHash entry pass
+//     through unchanged; this preserves backward compatibility with sessions
+//     created before token binding was introduced.
+func tokenBindingMiddleware(
+	sessionStorage *transportsession.Manager,
+	terminator sessionTerminator,
+) func(http.Handler) http.Handler {
```
jerm-dro: Blocker: can you explore alternatives to implementing this as middleware? I'd like to avoid adding more middleware layers, since:
- It's cumbersome to test and ensure it's integrated properly.
- This implementation relies on reading/writing things to the session metadata store. This reproduces the same problem we had previously: data needs to be read/written to the session storage in order to implement crucial logic. Ideally, the metadata store is only used for persistent storage. 

One thought I had was moving the responsibility for enforcing the caller identity into the session object. For example, that would require saving the hashed bearer token on the MultiSession and then enforcing it at tool call time with the current identity:
```go
CallTool(ctx context.Context, caller *auth.Identity, toolName string, arguments map[string]any, meta map[string]any) (*vmcp.ToolCallResult, error)                                                                                                 
ReadResource(ctx context.Context, caller *auth.Identity, uri string) (*vmcp.ResourceReadResult, error)
GetPrompt(ctx context.Context, caller *auth.Identity, name string, arguments map[string]any) (*vmcp.PromptGetResult, error)
```

An approach like this is nice, because:
- Testing becomes much easier. You can implement this session-hijack prevention as a decorator with only 1 mock.
- Explicit requirement on the session object: "if you want to invoke a method on the session, you must say who you are." 


What do you think?

pkg/vmcp/client/client.go
```diff
@@ -628,7 +629,10 @@ func (h *httpBackendClient) CallTool(
 // ReadResource retrieves a resource from the backend MCP server.
 // Returns the complete resource result including _meta field.
 func (h *httpBackendClient) ReadResource(
-	ctx context.Context, target *vmcp.BackendTarget, uri string,
+	ctx context.Context,
+	_ *auth.Identity, // Caller validation happens at session level, not HTTP transport level
```
jerm-dro: nit: this comment is unnecessarily repeated

pkg/vmcp/session/default_session.go
```diff
@@ -66,6 +70,14 @@ type defaultMultiSession struct {
...
 
 	queue AdmissionQueue
+
+	// Token binding fields: enforce that subsequent requests come from the same
+	// identity that created the session.
+	// These fields are immutable after session creation (no mutex needed).
+	boundTokenHash string // HMAC-SHA256 hash of creator's token (empty for anonymous)
+	tokenSalt      []byte // Random salt used for HMAC (empty for anonymous)
+	hmacSecret     []byte // Server-managed secret for HMAC-SHA256
+	allowAnonymous bool   // Whether to allow nil caller
```
jerm-dro: Blocker: Rather than adding more complexity here, can we decorate:

```go

type HijackPreventionDecorator struct {
  MultiSession 
   boundTokenHash string // HMAC-SHA256 hash of creator's token (empty for anonymous)
	tokenSalt      []byte // Random salt used for HMAC (empty for anonymous)
	hmacSecret     []byte // Server-managed secret for HMAC-SHA256
	allowAnonymous bool 
}

// Then, overwrite the relevant interface methods:

func (s HijackPreventionDecorator) CallTool(
	ctx context.Context,
	caller *auth.Identity,
	toolName string,
	arguments map[string]any,
	meta map[string]any,
) (*vmcp.ToolCallResult, error) {
	// Validate caller identity
	if err := s.validateCaller(caller); err != nil {
		return nil, err
	}
    return s.MultiSession.CallTool(...)
}

```

I'd prefer to not add more responsibilities to the existing session object.

pkg/vmcp/session/factory.go
```diff
@@ -258,13 +388,45 @@ func validateSessionID(id string) error {
...
+		if err != nil {
+			return "", nil, fmt.Errorf("failed to generate token salt: %w", err)
+		}
+		// Compute HMAC-SHA256 hash with server secret and per-session salt
+		boundTokenHash = HashToken(identity.Token, f.hmacSecret, tokenSalt)
+	}
+	return boundTokenHash, tokenSalt, nil
+}
+
+// populateBackendMetadata adds backend IDs to session metadata.
```
jerm-dro, line 367: Similar to the comment above about adding a decorator. We should move most of the functionality into a `session/security` package. There's complexity here that I would like to insulate the core session implementation from as much as possible.

cmd/vmcp/app/commands.go
```diff
@@ -468,6 +474,33 @@ func runServe(cmd *cobra.Command, _ []string) error {
...
+				"recommended_length", minRecommendedSecretLen,
+			)
+		}
+		slog.Info("using HMAC secret from VMCP_SESSION_HMAC_SECRET environment variable for session token binding")
+		sessionFactory = vmcpsession.NewSessionFactory(
+			outgoingRegistry,
+			vmcpsession.WithHMACSecret([]byte(hmacSecretEnv)),
+		)
+	} else {
+		slog.Warn("VMCP_SESSION_HMAC_SECRET not set - using default insecure secret (NOT recommended for production)")
```
jerm-dro: Suggestion: can we check if the k8s API is available or something and fail rather than warn?

cmd/vmcp/README.md
```diff
@@ -168,6 +168,77 @@ outgoingAuth:
...
+**Security Features:**
+
+- **HMAC-SHA256 Hashing**: Token hashes use HMAC with a server-managed secret
+- **Per-Session Salt**: Each session has a unique random salt
+- **Constant-Time Comparison**: Prevents timing attacks
+- **Request-Level Validation**: Each request independently validates the caller token; failed validation terminates the session immediately to prevent session hijacking attacks
+
+**Configuration:**
+
+Set the HMAC secret via environment variable (required for production):
```
jerm-dro, line 186: Not in this PR, but can we automate this in the operator?

yrobla: ok will do in a follow up

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -157,7 +163,10 @@ func (sm *Manager) CreateSession(
 	}
 
 	// Build the fully-formed MultiSession using the SDK-assigned session ID.
-	sess, err := sm.factory.MakeSessionWithID(ctx, sessionID, identity, backends)
+	// Sessions created with an identity are bound to that identity (allowAnonymous=false).
+	// Sessions created without an identity allow anonymous access (allowAnonymous=true).
+	allowAnonymous := vmcpsession.ShouldAllowAnonymous(identity)
```
jerm-dro, line 168: Suggestion: move the calculation of `ShouldAllowAnonymous` into `session/security`. This an implementation detail of constructing a session that doesn't concern the session manager.

pkg/vmcp/session/factory.go
jerm-dro: Rather than adding more complexity to the defaultMultiSessionFactory, let's move:

computeTokenBinding
- lines 314-328
- lines 452-466
- MetadataKeyTokenSalt, MetadataKeyTokenHash
- HijackPreventionDecorator

into session/security as well.

We've got concerns spread in two locations which makes the code harder to follow:

defaultMultiSessionFactory - handles all of the initialization of the NewHijackPreventionDecorator and its dependencies.
defaultMultiSessionFactory - handles all writing HijackPreventionDecorator's metadata to session storage
HijackPreventionDecorator - actually uses the data

We could encapsulate the logic better if the responsibilities for initialization, persistence were moved entirely to session/security. That would look like:
package security
```go
func PreventSessionHijacking(
   session MultiSession, // the session to protect
   hmacSecret string,
   identity *auth.Identity,
   sessionStorage // whatever the interface is that lets you write metadata.
) MultiSession // the decorated MultiSession that prevents hijacking.
```

That should allow you to make everything else in that session/security package private. That would also make testing more declarative.

pkg/vmcp/server/token_binding_integration_test.go
```diff
@@ -0,0 +1,347 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
jerm-dro, line 1: Suggestion: delete these tests. 

They're so complex to setup, and we've factored the code so it is easily unit tested. I think these tests provide little marginal value.

pkg/vmcp/server/session_management_v2_integration_test.go
```diff
@@ -105,24 +107,54 @@ func newV2FakeFactory(tools []vmcp.Tool) *v2FakeMultiSessionFactory {
...
+		testSecret := []byte("integration-test-secret")
+		testSalt := []byte("test-salt-123456")
+		tokenHash := security.HashToken(identity.Token, testSecret, testSalt)
+		baseSession.SetMetadata(vmcpsession.MetadataKeyTokenHash, tokenHash)
+		baseSession.SetMetadata(vmcpsession.MetadataKeyTokenSalt, hex.EncodeToString(testSalt))
+	} else {
+		baseSession.SetMetadata(vmcpsession.MetadataKeyTokenHash, "")
+	}
+
+	sess := newV2FakeMultiSession(baseSession, f.tools)
```
jerm-dro, line 129: Per my comment below: I think these changes are unnecessary and we can delete this level of integration testing.

pkg/vmcp/session/factory.go
jerm-dro: You may also find it useful to move some of the `token_binding_test.go` into the session/security package as well.

It'd be good to leave some unit tests at this level, just to assert the hijack prevention is integrated properly..

pkg/vmcp/session/security/security.go
```diff
@@ -0,0 +1,58 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
jerm-dro, line 1: really, this package should be `session/internal/security`. There's no reason for it to be used outside of session.

PR #3966, opened 2026-03-02
Fix GoReleaser CLI docs tarball upload path

ChrisJBurns:
## Summary
- GoReleaser v2 doesn't support absolute paths in `extra_files` glob patterns. The `/tmp/thv-cli-docs.tar.gz` path was internally prefixed with `./`, producing `.//tmp/thv-cli-docs.tar.gz` which fails with `stat: invalid argument`
- Moves the CLI docs tarball to a relative `build/` directory, which survives GoReleaser's `--clean` flag (only wipes `dist/`)
- Adds `build/` to `.gitignore`

Fixes the failure in: https://github.com/stacklok/toolhive/actions/runs/22589401671/job/65445484908

## Test plan
- [ ] Trigger a release and verify the CLI docs tarball is included in the GitHub release artifacts

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -141,7 +141,9 @@ jobs:
           rm ./thv-test
 
       - name: Bundle CLI docs
-        run: tar -czf /tmp/thv-cli-docs.tar.gz -C docs/cli .
+        run: |
+          mkdir -p build
+          tar -czf build/thv-cli-docs.tar.gz -C docs/cli .
 
       - name: Run GoReleaser
         id: run-goreleaser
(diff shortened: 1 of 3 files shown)


.github/workflows/releaser.yml
```diff
@@ -141,7 +141,9 @@ jobs:
           rm ./thv-test
 
       - name: Bundle CLI docs
-        run: tar -czf /tmp/thv-cli-docs.tar.gz -C docs/cli .
+        run: |
+          mkdir -p build
+          tar -czf build/thv-cli-docs.tar.gz -C docs/cli .
```
aponcedeleonch, line 146: I had a very similar fix but putting the tar in root, i.e. `tar -czf thv-cli-docs.tar.gz -C docs/cli .`

aponcedeleonch (approved):
nice!

PR #3974, opened 2026-03-03
Refactor session timeouts and add V2 feature warnings

yrobla:
Extract hardcoded timeout values in session manager into package-level constants for better maintainability. Add startup warnings when SessionManagementV2 is configured with incompatible features.

Changes:
- pkg/transport/session: Extract 5s and 30s timeouts to named constants (defaultOperationTimeout and cleanupOperationTimeout) to improve readability and make adjustments easier
- pkg/vmcp/server: Warn operators at startup when SessionManagementV2 is enabled alongside unsupported features (optimizer mode, composite tools) to provide early feedback instead of silent feature loss

Addresses PR review feedback from @aponcedeleonch.

Closes: #3866

### DIFF
--- a/pkg/transport/session/manager.go
+++ b/pkg/transport/session/manager.go
@@ -11,6 +11,13 @@ import (
 	"time"
 )
 
+const (
+	// defaultOperationTimeout is the timeout for standard storage operations
+	defaultOperationTimeout = 5 * time.Second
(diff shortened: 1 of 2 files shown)


aponcedeleonch (dismissed):
nice, thanks!

aponcedeleonch (approved):
nice catch!

PR #3975, opened 2026-03-03
Fix remote workload restart port bind error when unauthenticated

amirejaz:
## Problem

When a remote workload becomes unauthenticated (e.g., token refresh fails when the laptop disconnects from the internet), the proxy process continues running and holding the port. If the user runs `thv restart`, it fails with "port already in use" because we never stopped the existing proxy before attempting to start a new one.

## Solution

1. **Stop proxy before restart**: Always call `stopProxyIfNeeded` before loading config and starting a new runner for any restart path.

2. **Unified restart flow**: Consolidate the restart logic for all remote workload states (Running+dead, Unauthenticated, Unhealthy, Error). The flow now:
   - Stops the proxy first (so "Stopped" status matches actual state)
   - Sets Stopping → removeClientConfigurations → Stopped
   - Loads config and starts the new runner

   This removes special handling for Running+dead and keeps status consistent with reality.

## Changes

- **pkg/workloads/manager.go**: Refactor `maybeSetupRemoteWorkload` to use the unified restart flow.
- **pkg/workloads/manager_test.go**: Update tests for the new flow; add a test for unauthenticated restart.

## Testing

Unit tests updated and passing. Covers:
- Remote workload with healthy supervisor (no-op)
- Remote workload with dead supervisor (restart with unified flow)
- Remote workload unauthenticated (restart with unified flow)

### DIFF
--- a/cmd/thv-operator/test-integration/mcp-remote-proxy/mcpremoteproxy_controller_integration_test.go
+++ b/cmd/thv-operator/test-integration/mcp-remote-proxy/mcpremoteproxy_controller_integration_test.go
@@ -264,10 +264,8 @@ var _ = Describe("MCPRemoteProxy Controller", Label("k8s", "remoteproxy"), func(
 			}, MediumTimeout)
 
 			By("verifying the phase is Pending (since deployment is not ready in envtest)")
-			phase, err := proxyHelper.GetRemoteProxyPhase(proxy.Name)
-			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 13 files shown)


eleftherias:
The changes make sense. I'll wait for the PR checks to be green before approving. It looks like there's a flaky tests causing a failure now.

amirejaz:
@eleftherias could you please review again? I’ve updated the PR to fix the e2e tests. The issue was that cleanup was running on the first MCP server start as well; now it only runs on restart (non-Starting) states.

PR #3978, opened 2026-03-03
Add telemetry instrumentation for optimizer

aponcedeleonch:
## Summary
- Wrap the optimizer factory with a telemetry decorator (`telemetryOptimizer`) that records OTel metrics and traces for `FindTool` and `CallTool` operations
- Metrics include request/error counters, duration histograms, result-set size, and token-savings percentages
- The decorator is applied in `Server.Start` when a `TelemetryProvider` is configured
- Add comprehensive table-driven unit tests covering success, error, and not-found scenarios

## Test plan
- [x] Unit tests pass (`go test ./pkg/vmcp/server/... -count=1`)
- [x] `go vet` clean
- [x] `gofmt` clean
- [ ] Deploy with OTel stack and verify metrics appear in Prometheus/Grafana

### DIFF
--- a/pkg/vmcp/server/telemetry.go
+++ b/pkg/vmcp/server/telemetry.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/mark3labs/mcp-go/mcp"
+	mcpserver "github.com/mark3labs/mcp-go/server"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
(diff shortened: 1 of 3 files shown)


pkg/vmcp/server/telemetry.go
```diff
@@ -332,3 +335,198 @@ func (t *telemetryWorkflowExecutor) ExecuteWorkflow(ctx context.Context, params
 
 	return result, err
 }
+
+// monitorOptimizer wraps an optimizer factory so that every Optimizer instance
+// produced by the factory is decorated with telemetry (metrics + traces).
+func monitorOptimizer(
```
aponcedeleonch, line 341: not sure if this code should live here or inside `optimizer` package. I don't have strong opinions about either, feel free to chime in and I can make the change

yrobla: i think that adding in telemetry makes sense

aponcedeleonch:
@jerm-dro will merge this. If you have any comment feel free to leave it and I will address it in a future PR

Issue #3985, opened 2026-03-04
Flaky test: TestServer_HealthMonitoring_Enabled


JAORMX:
`TestServer_HealthMonitoring_Enabled` in `pkg/vmcp/server/health_monitoring_test.go` is flaky due to tight timing thresholds.

The test uses:
- `check_interval=50ms`
- `unhealthy_threshold=1`
- `timeout=10ms`
- `degraded_threshold=5ms`

Under load or with the race detector enabled, the health check can exceed the 5ms degraded threshold, causing the test to see `"degraded"` instead of the expected `"healthy"` status.

```
health_monitoring_test.go:152:
    Error:      Not equal:
                expected: "healthy"
                actual  : "degraded"
```

The fix should increase the timing thresholds to be more tolerant of CI/local machine variance.

PR #3986, opened 2026-03-04
Enable sticky sessions on operator-created Services

JAORMX:
## Summary

- Set `SessionAffinity: ClientIP` on all three operator-created Service types (MCPServer, MCPRemoteProxy, VirtualMCPServer) so stateful MCP session protocols (SSE, streamable-http) work correctly when replicas > 1
- Add drift detection in `serviceNeedsUpdate()` for each controller so existing Services without `ClientIP` affinity get reconciled
- Copy `SessionAffinity` in the `ensureService` update paths alongside existing `Ports`/`Type` copies

## Test plan

- [x] `go test -race ./cmd/thv-operator/controllers/...` — all pass
- [x] `task test` — all pass (one pre-existing flaky test unrelated: #3985)
- [ ] Deploy operator to cluster, create MCPServer/MCPRemoteProxy/VirtualMCPServer, verify `kubectl get svc <name> -o jsonpath='{.spec.sessionAffinity}'` returns `ClientIP`

### DIFF
--- a/docs/arch/09-operator-architecture.md
+++ b/docs/arch/09-operator-architecture.md
@@ -324,6 +324,7 @@ For complete examples of all CRDs, see the [`examples/operator/mcp-servers/`](..
 3. **Service**
    - Exposes proxy deployment
    - Type: ClusterIP, LoadBalancer, or NodePort
+   - SessionAffinity: ClientIP (ensures stateful MCP sessions reach the same pod)
    - Routes traffic to proxy
 
 4. **ConfigMap** (RunConfig)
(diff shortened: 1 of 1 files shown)


eleftherias:
Should we make this configurable by the user?

JAORMX:
@eleftherias we could, though I don't know in what case you wouldn't want this.

eleftherias:
If the server is stateless then the stickiness isn't needed and isn't the best way to distribute the load. Also if they use an external load balancer this could cause conflicts with the policy they've configured.

JAORMX:
@eleftherias gotcha. In most cases the MCP server is stateful, though. And this is the case for our proxy as well. Should we then default to havin this enabled and have a flag to optionally disable this?

eleftherias:
Yes that makes sense

JAORMX:
@eleftherias should I add the configurability to a separate PR or to this one?

eleftherias (approved):

eleftherias:
Separate PR works

Issue #3653, opened 2026-02-06
CLI skills commands implementation


JAORMX:
## Description

Replace the CLI command stubs (from TASK-001) with real implementations that call the Skills HTTP client. Handle flags, output formatting (table/JSON), and shell completion.

## Context

Part of the Skills Lifecycle Management epic. These are thin wrappers — they parse flags, call the HTTP client, and format output. No business logic in CLI code per ToolHive conventions.

**Dependencies**: Skills HTTP client
**Blocks**: ServerBuilder wiring and E2E tests

## Acceptance Criteria

- [ ] `thv skill list` — lists installed skills in table/JSON format, supports `--client`, `--global`, `--local`, `--group`, `--format`
- [ ] `thv skill install <ref>` — installs skill, supports `--client` (required), `--global`/`--local`, `--group`, `--force`, `-u`/`--upgrade`
- [ ] `thv skill uninstall <name>` — removes managed skill, supports `--client` (required), `--global`/`--local`
- [ ] `thv skill info <name>` — shows skill details, supports `--client`, `--format`
- [ ] `thv skill validate <path>` — validates skill directory, supports `--format`
- [ ] `thv skill build <path>` — builds OCI artifact, supports `--tag`, `--platform`
- [ ] `thv skill push <ref>` — pushes to registry
- [ ] Silent success pattern (no output on success unless `--debug`)
- [ ] Actionable error messages with hints
- [ ] Shell completion via `ValidArgsFunction`
- [ ] SPDX headers

## Technical Approach

Follow `cmd/thv/app/group.go` pattern. Use flag helpers (`AddFormatFlag`, etc.).

### Code Pointers
- `cmd/thv/app/group.go` — command pattern
- `cmd/thv/app/common.go` — flag helpers
- `docs/cli-best-practices.md` — CLI guidelines

PR #3988, opened 2026-03-04
Implement skill install and uninstall CLI commands

JAORMX:
## Summary
- Replace empty `skill install` stub with real implementation supporting `--client`, `--scope` (default "user"), `--force`, and `--project-root` flags
- Replace empty `skill uninstall` stub with real implementation supporting `--scope` (default "user"), `--project-root` flags with shell completion and `validateProjectRootForScope` validation
- Replace empty `skill validate` stub with real implementation supporting `--format` flag, absolute path resolution, error/warning output, and non-zero exit on invalid (silent on success per convention)
- Replace empty `skill build` stub with real implementation supporting `--tag`/`-t` flag, printing the built OCI reference on success (documented exception to silent-success)
- Replace empty `skill push` stub with real implementation following silent-success convention
- Move `validateProjectRootForScope` to `skill_helpers.go` for reuse across install/uninstall

Supersedes #3989 (now included here). Closes stacklok/toolhive#3653.

## Test plan
- [ ] `task lint` passes with no new warnings
- [ ] `task test` passes (existing unit tests)
- [ ] `thv skill install --help` shows correct flags (--client, --scope, --force, --project-root)
- [ ] `thv skill uninstall --help` shows correct flags (--scope defaults to "user", --project-root)
- [ ] `thv skill validate --help` shows correct flags (--format)
- [ ] `thv skill build --help` shows correct flags (--tag/-t)
- [ ] `thv skill push --help` shows correct usage

### DIFF
--- a/cmd/thv/app/skill_build.go
+++ b/cmd/thv/app/skill_build.go
@@ -3,18 +3,52 @@
 
 package app
 
-import "github.com/spf13/cobra"
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"github.com/stacklok/toolhive/pkg/skills"
+)
+
+var skillBuildTag string
 
 var skillBuildCmd = &cobra.Command{
 	Use:   "build [path]",
 	Short: "Build a skill",
-	Long:  `Build a skill from a local directory into an OCI artifact that can be pushed to a registry.`,
-	Args:  cobra.ExactArgs(1),
-	RunE: func(_ *cobra.Command, _ []string) error {
-		return nil
+	Long: `Build a skill from a local directory into an OCI artifact that can be pushed to a registry.
+
+On success, prints the OCI reference of the built artifact to stdout.`,
+	Args: cobra.ExactArgs(1),
+	ValidArgsFunction: func(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+		return nil, cobra.ShellCompDirectiveFilterDirs
 	},
+	RunE: skillBuildCmdFunc,
 }
 
 func init() {
 	skillCmd.AddCommand(skillBuildCmd)
+
+	skillBuildCmd.Flags().StringVarP(&skillBuildTag, "tag", "t", "", "OCI tag for the built artifact")
+}
+
+func skillBuildCmdFunc(cmd *cobra.Command, args []string) error {
+	absPath, err := filepath.Abs(args[0])
+	if err != nil {
+		return fmt.Errorf("failed to resolve path: %w", err)
+	}
+
+	c := newSkillClient()
+
+	result, err := c.Build(cmd.Context(), skills.BuildOptions{
+		Path: absPath,
+		Tag:  skillBuildTag,
+	})
+	if err != nil {
+		return formatSkillError("build skill", err)
+	}
+
+	fmt.Println(result.Reference)
+	return nil
 }
--- a/cmd/thv/app/skill_install.go
+++ b/cmd/thv/app/skill_install.go
@@ -4,8 +4,6 @@
 package app
 
 import (
-	"fmt"
-
 	"github.com/spf13/cobra"
 
 	"github.com/stacklok/toolhive/pkg/skills"
@@ -55,15 +53,4 @@ func skillInstallCmdFunc(cmd *cobra.Command, args []string) error {
 	}
 
 	return nil
-}
-
-// validateProjectRootForScope returns a PreRunE that ensures --project-root is
-// provided when --scope is "project".
-func validateProjectRootForScope(scopeVar, projectRootVar *string) func(*cobra.Command, []string) error {
-	return func(_ *cobra.Command, _ []string) error {
-		if skills.Scope(*scopeVar) == skills.ScopeProject && *projectRootVar == "" {
-			return fmt.Errorf("--project-root is required when --scope is %q", skills.ScopeProject)
-		}
-		return nil
-	}
 }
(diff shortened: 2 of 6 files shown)


cmd/thv/app/skill_uninstall.go
```diff
@@ -3,18 +3,47 @@
...
+	Use:               "uninstall [skill-name]",
+	Short:             "Uninstall a skill",
+	Long:              `Remove a previously installed skill by name.`,
+	Args:              cobra.ExactArgs(1),
+	ValidArgsFunction: completeSkillNames,
+	PreRunE:           validateSkillScope(&skillUninstallScope),
+	RunE:              skillUninstallCmdFunc,
 }
 
 func init() {
```
aponcedeleonch, line 30: nit: `install` defaults `--scope` to `"user"` explicitly, but `uninstall` defaults to `""`. The server-side `defaultScope()` maps empty to `"user"` anyway, so the runtime behavior is the same — but the `--help` output will look inconsistent between the two commands. Consider making it explicit here too:

```suggestion
	skillUninstallCmd.Flags().StringVar(&skillUninstallScope, "scope", string(skills.ScopeUser), "Scope to uninstall from (user, project)")
```

cmd/thv/app/skill_uninstall.go
```diff
@@ -3,18 +3,47 @@
...
 
 var skillUninstallCmd = &cobra.Command{
-	Use:   "uninstall [skill-name]",
-	Short: "Uninstall a skill",
-	Long:  `Remove a previously installed skill by name.`,
-	Args:  cobra.ExactArgs(1),
-	RunE: func(_ *cobra.Command, _ []string) error {
-		return nil
-	},
+	Use:               "uninstall [skill-name]",
```
aponcedeleonch, line 18: nit: `install` chains `validateProjectRootForScope` in its `PreRunE` to catch `--scope project` without `--project-root` early. `uninstall` has the same `--project-root` flag but doesn't validate it, so the error would surface deeper in the stack instead of at the CLI layer. Consider:

```suggestion
	PreRunE: chainPreRunE(
		validateSkillScope(&skillUninstallScope),
		validateProjectRootForScope(&skillUninstallScope, &skillUninstallProjectRoot),
	),
```

cmd/thv/app/skill_install.go
```diff
@@ -3,19 +3,67 @@
...
+		Name:        args[0],
+		Scope:       skills.Scope(skillInstallScope),
+		Client:      skillInstallClient,
+		Force:       skillInstallForce,
+		ProjectRoot: skillInstallProjectRoot,
+	})
+	if err != nil {
+		return formatSkillError("install skill", err)
+	}
+
```
aponcedeleonch, line 54: nit: `validateProjectRootForScope` is a general-purpose validator (like `validateSkillScope`), but it lives in `skill_install.go`. If uninstall (and potentially future commands) reuse it, it'd be more discoverable in `skill_helpers.go` alongside the other shared validators.

cmd/thv/app/skill_build.go
```diff
@@ -3,18 +3,50 @@
...
+
+	result, err := c.Build(cmd.Context(), skills.BuildOptions{
+		Path: absPath,
+		Tag:  skillBuildTag,
+	})
+	if err != nil {
+		return formatSkillError("build skill", err)
+	}
+
+	fmt.Println(result.Reference)
```
aponcedeleonch, line 52: nit: The project convention is silent success — `install`, `uninstall`, and `push` in this PR all return nothing on success. Printing the reference here is arguably useful for scripting (`ref=$(thv skill build .)`), but it breaks the pattern. If it's intentional, maybe document it in the `Long` description so users know to expect it.

cmd/thv/app/skill_validate.go
```diff
@@ -3,18 +3,69 @@
...
+	if err != nil {
+		return formatSkillError("validate skill", err)
+	}
+
+	switch skillValidateFormat {
+	case FormatJSON:
+		data, err := json.MarshalIndent(result, "", "  ")
+		if err != nil {
+			return fmt.Errorf("failed to marshal JSON: %w", err)
+		}
```
aponcedeleonch, line 52: nit: Per silent-success convention, printing `"Valid"` in text mode is a bit chatty. The exit code already signals success. Errors and warnings are useful output, but the happy path could just be silent (JSON mode already communicates validity through the structured output).

cmd/thv/app/skill_uninstall.go
JAORMX: Good catch, fixed. Now defaults to `string(skills.ScopeUser)` so `--help` is consistent across both commands.

cmd/thv/app/skill_uninstall.go
JAORMX: Yep, makes total sense to catch it early at the CLI layer. Done, uninstall now chains `validateProjectRootForScope` in its `PreRunE` too.

cmd/thv/app/skill_install.go
JAORMX: Agreed. Moved it to `skill_helpers.go` alongside `validateSkillScope` and the other shared validators.

cmd/thv/app/skill_build.go
JAORMX: Yeah, the output is intentional for the scripting use case you mentioned (`ref=$(thv skill build .)`). Added it to the `Long` description so it's documented and not surprising.

cmd/thv/app/skill_validate.go
JAORMX: Fair point. Removed the "Valid" print, happy path is now silent. Errors and warnings still print since those are actionable.

PR #3992, opened 2026-03-04
Add ClusterIP service with SessionAffinity for MCP server backend

JAORMX:
## Summary

- Adds a new regular ClusterIP service (`mcp-<name>`) with `SessionAffinity: ClientIP` for MCP server StatefulSets, alongside the existing headless service (`mcp-<name>-headless`)
- The proxy-runner now targets this new service instead of the headless one, providing proper load balancing with session stickiness when the StatefulSet scales beyond 1 replica
- Adds finalizer cleanup for the new service and Chainsaw e2e test assertions for all SSE and streamable-http scenarios

## Context

The headless service bypasses kube-proxy, so DNS returns raw pod IPs with no load balancing or session affinity. If the MCP server StatefulSet scales beyond 1 replica, sessions break because requests may hit different pods. The new ClusterIP service solves this with `SessionAffinity: ClientIP`.

## Changes

| File | Change |
|------|--------|
| `pkg/container/kubernetes/client.go` | New `createMCPService()` function, called from `DeployWorkload()` after `createHeadlessService()`. Moves `SSEHeadlessServiceName` assignment to the new service. Extracts `serviceFieldManager` constant. |
| `cmd/thv-operator/controllers/mcpserver_controller.go` | Adds cleanup of `mcp-<name>` service in `finalizeMCPServer()` |
| `test/e2e/chainsaw/operator/**/assert-mcpserver-svc.yaml` | 4 new assertion files verifying the service exists with `sessionAffinity: ClientIP` |
| `test/e2e/chainsaw/operator/**/chainsaw-test.yaml` | 4 updated test files adding the new assertion step |

## Test plan

- [x] `task lint` passes (0 issues)
- [x] `task test` passes (unit tests)
- [ ] Deploy operator, create MCPServer with SSE transport, verify both services exist:
  - `kubectl get svc mcp-<name>-headless` (headless, unchanged)
  - `kubectl get svc mcp-<name>` (new service with `sessionAffinity: ClientIP`)
- [ ] Chainsaw e2e tests pass for single-tenancy and multi-tenancy SSE/streamable-http scenarios

### DIFF
--- a/pkg/container/kubernetes/client_test.go
+++ b/pkg/container/kubernetes/client_test.go
@@ -15,6 +15,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	corev1apply "k8s.io/client-go/applyconfigurations/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
@@ -856,8 +857,14 @@ func TestDeployWorkloadCreatesBackendServices(t *testing.T) {
 
 			// Verify owner reference on headless service
 			require.Len(t, headlessSvc.OwnerReferences, 1)
+			assert.Equal(t, "apps/v1", headlessSvc.OwnerReferences[0].APIVersion)
 			assert.Equal(t, "StatefulSet", headlessSvc.OwnerReferences[0].Kind)
 			assert.Equal(t, containerName, headlessSvc.OwnerReferences[0].Name)
+			assert.Equal(t, k8stypes.UID("test-uid-123"), headlessSvc.OwnerReferences[0].UID)
+			require.NotNil(t, headlessSvc.OwnerReferences[0].Controller)
+			assert.True(t, *headlessSvc.OwnerReferences[0].Controller)
+			require.NotNil(t, headlessSvc.OwnerReferences[0].BlockOwnerDeletion)
+			assert.True(t, *headlessSvc.OwnerReferences[0].BlockOwnerDeletion)
 
 			// Verify the MCP ClusterIP service was created
 			mcpSvc, err := clientset.CoreV1().Services("default").Get(
@@ -870,8 +877,14 @@ func TestDeployWorkloadCreatesBackendServices(t *testing.T) {
 
 			// Verify owner reference on MCP service
 			require.Len(t, mcpSvc.OwnerReferences, 1)
+			assert.Equal(t, "apps/v1", mcpSvc.OwnerReferences[0].APIVersion)
 			assert.Equal(t, "StatefulSet", mcpSvc.OwnerReferences[0].Kind)
 			assert.Equal(t, containerName, mcpSvc.OwnerReferences[0].Name)
+			assert.Equal(t, k8stypes.UID("test-uid-123"), mcpSvc.OwnerReferences[0].UID)
+			require.NotNil(t, mcpSvc.OwnerReferences[0].Controller)
+			assert.True(t, *mcpSvc.OwnerReferences[0].Controller)
+			require.NotNil(t, mcpSvc.OwnerReferences[0].BlockOwnerDeletion)
+			assert.True(t, *mcpSvc.OwnerReferences[0].BlockOwnerDeletion)
 
 			// Verify MCPServiceName was set on options
 			assert.Equal(t, "mcp-"+containerName, options.MCPServiceName)


pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
...
+		slog.Info("no ports configured for MCP transport, skipping service creation")
+		return nil
+	}
+
+	svcName := fmt.Sprintf("mcp-%s", containerName)
+
+	// Create the service apply configuration with SessionAffinity
+	serviceApply := corev1apply.Service(svcName, namespace).
+		WithLabels(labels).
+		WithSpec(corev1apply.ServiceSpec().
```
ChrisJBurns: **`SessionAffinity: ClientIP` provides proxy-runner-level, not MCP-session-level stickiness.**

Since the proxy-runner pod is the client connecting to this service, ALL requests from a single proxy-runner pod come from the same IP. This means:
- 1 proxy-runner pod → all MCP sessions pinned to 1 backend (no load balancing)
- N proxy-runner pods → each pinned to 1 backend (coarse distribution)

MCP sessions are identified by the `Mcp-Session-Id` HTTP header (L7), but `ClientIP` affinity operates at L4 with no header awareness.

This is still an improvement over the headless service (stable VIP, graceful pod restart handling), but shall we document the limitation? True per-session routing will require `Mcp-Session-Id`-based routing at the proxy layer.

pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
...
+			WithSelector(map[string]string{
```
ChrisJBurns: **Missing `SessionAffinityConfig.ClientIP.TimeoutSeconds`.**

Kubernetes defaults to 10800s (3 hours) when no timeout is specified. This should be set explicitly to match expected MCP session duration.

```suggestion
			WithSessionAffinity(corev1.ServiceAffinityClientIP).
			WithSessionAffinityConfig(corev1apply.SessionAffinityConfig().
				WithClientIP(corev1apply.ClientIPConfig().
					WithTimeoutSeconds(1800))))
```

pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
...
+			WithType(corev1.ServiceTypeClusterIP).
+			WithSessionAffinity(corev1.ServiceAffinityClientIP))
+
+	// Apply the service using server-side apply
+	fieldManager := serviceFieldManager
+	_, err = c.client.CoreV1().Services(namespace).
+		Apply(ctx, serviceApply, metav1.ApplyOptions{
+			FieldManager: fieldManager,
+			Force:        true,
+		})
```
ChrisJBurns: **Use `slog.Debug` instead of `slog.Info`.**

Per ToolHive logging guidelines: "Do not log at INFO or above for successful operations." Service creation during workload deployment is a normal success path.

```suggestion
	slog.Debug("created MCP service with session affinity", "name", svcName)
```

pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
...
+
+	if err != nil {
+		return fmt.Errorf("failed to apply MCP service: %w", err)
```
ChrisJBurns: **Consider doing this rename in this PR rather than leaving a TODO.**

The field now holds a ClusterIP service name, not a headless service name. It's a mechanical 3-file rename (`runtime/types.go`, `runtime/setup.go`, `client.go`) — shipping with the misleading name creates a window where the name actively contradicts its content.

pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
 
 	slog.Info("created headless service for HTTP transport", "name", containerName)
 
+	return nil
+}
```
ChrisJBurns, line 1022: **Consider extracting shared service creation logic.**

`createHeadlessService()` and `createMCPService()` share ~80% of their logic (port creation, label/selector setup, server-side apply). Consider extracting a common `applyService()` helper that both functions call with different config (ClusterIP vs None, with/without SessionAffinity). This reduces duplication and makes adding future service types straightforward.

pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
...
+
+// createMCPService creates a regular ClusterIP service with SessionAffinity for the MCP server StatefulSet.
+// This service provides load balancing with client-IP-based session stickiness, which the proxy-runner
+// uses as its target host. The headless service is retained for DNS discovery purposes.
+func (c *Client) createMCPService(
+	ctx context.Context,
+	containerName string,
+	namespace string,
+	labels map[string]string,
+	options *runtime.DeployWorkloadOptions,
```
ChrisJBurns, line 1038: **Missing owner references — risk of resource leak.**

This service is created via server-side apply with no owner references. If the StatefulSet or MCPServer CR is deleted outside the operator finalizer path, this service will be orphaned.

Consider setting `OwnerReferences` in the service's `ObjectMeta` (pointing to the StatefulSet), or at minimum adding a `toolhive.stacklok.io/managed-by` label so orphaned services can be discovered and cleaned up. Note: the existing `createHeadlessService()` has the same gap.

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -1409,6 +1409,8 @@ func (r *MCPServerReconciler) updateMCPServerStatus(ctx context.Context, m *mcpv
 }
 
 // finalizeMCPServer performs the finalizer logic for the MCPServer
+//
```
ChrisJBurns: **Consider extracting a helper instead of adding `nolint:gocyclo`.**

The get-then-delete pattern now repeats 3 times in this function (StatefulSet, headless service, MCP service). A small helper would eliminate the complexity and the need for the lint suppression:

```go
func (r *MCPServerReconciler) deleteIfExists(ctx context.Context, obj client.Object, name, namespace, kind string) error {
	err := r.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, obj)
	if err == nil {
		if delErr := r.Delete(ctx, obj); delErr != nil && !errors.IsNotFound(delErr) {
			return fmt.Errorf("failed to delete %s %s: %w", kind, name, delErr)
		}
		return nil
	}
	if !errors.IsNotFound(err) {
		return fmt.Errorf("failed to check %s %s: %w", kind, name, err)
	}
	return nil
}
```

test/e2e/chainsaw/operator/single-tenancy/test-scenarios/sse/assert-mcpserver-svc.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: mcp-st-sse
+  namespace: toolhive-system
+spec:
+  sessionAffinity: ClientIP
```
ChrisJBurns, line 8: Consider also asserting `spec.type: ClusterIP` and `spec.sessionAffinityConfig.clientIP.timeoutSeconds` (if the timeout suggestion is adopted) to make the assertion more complete.

pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
...
+		})
+
+	if err != nil {
+		return fmt.Errorf("failed to apply MCP service: %w", err)
+	}
+
+	slog.Info("created MCP service with session affinity", "name", svcName)
+
+	// TODO: rename SSEHeadlessServiceName to MCPServiceName
 	options.SSEHeadlessServiceName = svcName
```
jhrozek: nit: The `SSEHeadlessServiceName` field name is now misleading — it stores the ClusterIP service name, not the headless one. The comment in `types.go` also says "Only applicable when using SSE transport" but it's now used for streamable-http too. Consider renaming to `MCPServiceName` (small surface area: `types.go`, `setup.go`, `client.go`).

pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
...
+
+	// Create the service apply configuration with SessionAffinity
+	serviceApply := corev1apply.Service(svcName, namespace).
+		WithLabels(labels).
+		WithSpec(corev1apply.ServiceSpec().
+			WithSelector(map[string]string{
+				"app": containerName,
+			}).
+			WithPorts(servicePorts...).
+			WithType(corev1.ServiceTypeClusterIP).
```
jhrozek: question: `createServicePorts` can produce ports with `NodePort` set (e.g. host-port bindings in 30000-32767 range). `createHeadlessService` handles this by upgrading the service type to `NodePort` (lines 921-928), but `createMCPService` hard-codes `ClusterIP`. Kubernetes will reject a ClusterIP service with `nodePort` set on any port. Is this an intentional omission (NodePort bindings not expected with session-affinity services), or should we add the same type detection logic here?

pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
...
+			WithSessionAffinity(corev1.ServiceAffinityClientIP))
```
jhrozek: question: `SessionAffinity=ClientIP` without `SessionAffinityConfig` defaults to 10800s (3h). Was this an intentional decision? If so, a brief comment documenting why the default is appropriate for MCP sessions would be helpful for future readers.

pkg/container/kubernetes/client.go
```diff
@@ -946,6 +956,59 @@ func (c *Client) createHeadlessService(
 
 	slog.Info("created headless service for HTTP transport", "name", containerName)
 
+	return nil
+}
+
+// createMCPService creates a regular ClusterIP service with SessionAffinity for the MCP server StatefulSet.
```
jhrozek, line 1030: nit: `createMCPService` and `createHeadlessService` share nearly identical structure (call `createServicePorts`, guard on empty ports, format name, build spec, SSA apply). The only differences are name suffix, `ClusterIP("None")` vs `SessionAffinity`, and NodePort handling. Consider extracting a shared `applyService` helper parameterized by a config struct to reduce duplication.

pkg/container/kubernetes/client.go
```diff
@@ -421,11 +425,17 @@ func (c *Client) DeployWorkload(ctx context.Context,
 	slog.Info("applied statefulset", "name", createdStatefulSet.Name)
 
 	if transportTypeRequiresHeadlessService(transportType) && options != nil {
-		// Create a headless service for SSE transport
+		// Create a headless service for DNS discovery
```
jhrozek, line 438: nit: `transportTypeRequiresHeadlessService` now gates creation of both the headless service *and* the new ClusterIP service. The name only reflects the headless part. Consider renaming to `transportTypeRequiresBackendServices` to match its expanded scope.

pkg/container/kubernetes/client.go
JAORMX: Good call on documenting this. I've added a comment on the `mcpServiceSessionAffinityTimeout` constant that explicitly calls out the L4 vs L7 limitation:

> This provides proxy-runner-level stickiness (L4), not per-MCP-session stickiness (L7). True per-session routing would require Mcp-Session-Id-based routing at the proxy layer.

For the current single-replica default this is fine, and it's still a solid improvement over the headless service (stable VIP, graceful restarts). We can track the L7 routing as a follow-up when we actually need multi-replica scaling.

pkg/container/kubernetes/client.go
JAORMX: Done! Set to 1800s (30 min) via a named constant `mcpServiceSessionAffinityTimeout`. 3h felt way too long for our use case.

pkg/container/kubernetes/client.go
JAORMX: Good catch, fixed. I cargo-culted the existing `slog.Info` from `createHeadlessService` instead of checking the logging guidelines. Both now go through the shared `applyService` helper using `slog.Debug`.

pkg/container/kubernetes/client.go
JAORMX: Agreed, done. Renamed to `MCPServiceName` across all three files. No reason to ship a name that actively contradicts what it holds.

pkg/container/kubernetes/client.go
JAORMX: Done! Extracted `applyService` with a `serviceConfig` struct. Both `createHeadlessService` and `createMCPService` are now thin wrappers. This also helped us catch a bug: the headless path could produce an invalid `clusterIP=None + type=NodePort` combo if ports had NodePort values. The shared helper now skips NodePort promotion for headless services.

pkg/container/kubernetes/client.go
JAORMX: Yeah, this is a pre-existing gap that affects the headless service too. I'd rather not scope-creep this PR with the owner reference plumbing. Let's track it separately -- it's worth doing properly for both services at once.

cmd/thv-operator/controllers/mcpserver_controller.go
JAORMX: Done, basically your exact suggestion. The `nolint:gocyclo` is gone and the finalizer reads much cleaner now.

test/e2e/chainsaw/operator/single-tenancy/test-scenarios/sse/assert-mcpserver-svc.yaml
JAORMX: Done! All four assertion files now also check `spec.type: ClusterIP` and `sessionAffinityConfig.clientIP.timeoutSeconds: 1800`.

pkg/container/kubernetes/client.go
JAORMX: Yep, done. Renamed to `MCPServiceName` across all three files and updated the comment to mention both SSE and streamable-http.

pkg/container/kubernetes/client.go
JAORMX: Great catch! This was indeed a bug. I've unified both service creation paths into a shared \`applyService\` helper, and the NodePort type detection now runs for both. Note that I also had to guard against the *reverse* problem: headless services (\`clusterIP=None\`) cannot be \`type: NodePort\` -- Kubernetes rejects that combo. So the helper skips NodePort promotion when \`cfg.headless\` is true.

pkg/container/kubernetes/client.go
JAORMX: Not intentional -- 3h was just the K8s default sneaking in. Now set explicitly to 1800s (30 min) via a named constant with a comment explaining the choice.

pkg/container/kubernetes/client.go
JAORMX: Done! Both you and Chris flagged this one. Extracted \`applyService\` with a \`serviceConfig\` struct -- both callers are now thin wrappers passing their config.

pkg/container/kubernetes/client.go
JAORMX: Done, renamed to \`transportTypeRequiresBackendServices\`.

pkg/container/kubernetes/client.go
JAORMX: Actually, went ahead and fixed this in the latest push. Both services now have the StatefulSet set as their owner reference via `metav1.OwnerReference` with `BlockOwnerDeletion` and `Controller` set. Kubernetes GC will clean them up if the StatefulSet is deleted outside the operator path.

Also added a table-driven unit test covering both SSE and streamable-http transports that verifies service creation, session affinity config, and all owner reference fields (UID, APIVersion, Kind, Name, Controller, BlockOwnerDeletion).

pkg/container/kubernetes/client.go
```diff
@@ -328,6 +331,8 @@ func (c *Client) GetWorkloadLogs(ctx context.Context, workloadName string, follo
 }
 
 // DeployWorkload implements runtime.Runtime.
+//
+//nolint:gocyclo
```
yrobla, line 335: i would have prefered to refactor and avoid the nolint...

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -1455,57 +1455,46 @@ func (r *MCPServerReconciler) updateMCPServerStatus(ctx context.Context, m *mcpv
 	return r.Status().Update(ctx, m)
 }
 
+// deleteIfExists fetches a Kubernetes object by name and namespace, and deletes it if it exists.
+// Returns nil if the object was not found or was successfully deleted.
+func (r *MCPServerReconciler) deleteIfExists(ctx context.Context, obj client.Object, name, namespace, kind string) error {
+	err := r.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, obj)
+	if err == nil {
+		if delErr := r.Delete(ctx, obj); delErr != nil && !errors.IsNotFound(delErr) {
```
yrobla, line 1463: we have lost the logging here, is that intentional?

yrobla (approved):
approving with a pair of comments

pkg/container/kubernetes/client.go
JAORMX: Good call. Yeah, the nolint is a bit lazy here. I'll extract the service creation block into its own method (something like `ensureBackendServices`) in a follow-up so we can drop the nolint cleanly. Tracking it with the other follow-up items from this review.

cmd/thv-operator/controllers/mcpserver_controller.go
JAORMX: Not intentional, it got lost during the refactoring into `deleteIfExists`. I'll add debug-level logging back in the follow-up so we still have visibility into finalization for troubleshooting.

PR #3994, opened 2026-03-04
Make SessionAffinity configurable via CRD field

JAORMX:
## Summary

- Add `sessionAffinity` field (enum: `ClientIP`/`None`, default: `ClientIP`) to `MCPServerSpec`, `MCPRemoteProxySpec`, and `VirtualMCPServerSpec`
- Wire the field through service creation and drift detection in all three controllers
- Fix pre-existing bug in MCPServer controller where service update path did not copy Labels/Annotations (could cause infinite reconciliation)

Follow-up to #3986 which hardcoded `SessionAffinity: ClientIP`. Review feedback identified this should be configurable: stateless MCP servers don't need stickiness and external load balancers may conflict with client-side affinity.

EmbeddingServer is intentionally excluded — embedding servers are stateless inference endpoints, not MCP protocol services.

## Test plan

- [x] `task lint` passes (0 issues)
- [x] `task test` passes
- [x] Generated CRDs contain new field with default and enum validation
- [ ] Manual: deploy operator, create MCPServer without `sessionAffinity` → Service has `ClientIP`
- [ ] Manual: create MCPServer with `sessionAffinity: None` → Service has `None`
- [ ] Manual: change `sessionAffinity` on existing resource → Service updates via drift detection


JAORMX:
@claude be a home and use a skill to fix the helm chart linter issue.

PR #3999, opened 2026-03-04
Add actionable logging when secrets file decryption fails

ChrisJBurns:
## Summary

- When AES-GCM decryption fails with "message authentication failed", log a warning explaining the likely causes (incorrect password or file corruption) and provide recovery guidance pointing users to delete the secrets file and run `thv secret setup`
- Wrap the encrypted provider creation error in `factory.go` for consistent error context

This addresses user-reported confusion around the opaque `cipher: message authentication failed` error, which can occur when:
1. The keyring password was reset/cleared and a different password was entered
2. The secrets file was corrupted by a non-atomic write (addressed by #3998)

## Test plan

- [x] `go test ./pkg/secrets/...` passes
- [x] `go vet ./pkg/secrets/...` passes
- [ ] Manual verification: delete keyring entry, enter wrong password, confirm warning message is logged

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
       fail-fast: false
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
--- a/docs/server/docs.go
+++ b/docs/server/docs.go
@@ -1826,6 +1826,10 @@ const docTemplate = `{
                         "description": "Force allows overwriting unmanaged skill directories",
                         "type": "boolean"
                     },
+                    "group": {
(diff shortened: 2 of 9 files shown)


JAORMX (approved):

pkg/secrets/encrypted.go
```diff
@@ -174,6 +175,13 @@ func NewEncryptedManager(filePath string, key []byte) (Provider, error) {
 		}
 		decryptedContents, err := aes.Decrypt(encryptedContents, key)
 		if err != nil {
+			if strings.Contains(err.Error(), "message authentication failed") {
```
reyortiz3, line 178: nit: there is no way to use error.Is(), right?

ChrisJBurns: Unfortunately no, Go's crypto/cipher package defines the GCM authentication error as a plain unexported `errors.New("cipher: message authentication failed")`. There's no exported error variable to compare against with `errors.Is()`.

I could add a sentinel error in the aes package and wrap the GCM error with it in Decrypt. Then `encrypted.go` can use `errors.Is()`. But thats probs overkill

PR #4001, opened 2026-03-04
Remove invalid 'apply' RBAC verb from operator roles

### LINKED ISSUE #3995: Remove invalid 'apply' RBAC verb from operator roles
## Summary

The `apply` string appears in RBAC verb lists in the operator controller code, but `apply` is not a valid Kubernetes RBAC verb. Server-side apply uses the `patch` verb (HTTP PATCH with `application/apply-patch+yaml`), which is already present in all affected rules.

The `apply` verb is silently accepted by the Kubernetes API server but never matches any real API request — it is a dead letter stored in etcd.

### 1. `defaultRBACRules` (compiled into proxy-runner Role)

- `cmd/thv-operator/controllers/mcpserver_controller.go` (lines 62 and 67)

### 2. Kubebuilder markers (compiled into operator ClusterRole)

- `cmd/thv-operator/controllers/mcpserver_controller.go` (lines 143, 151)
- `cmd/thv-operator/controllers/mcpremoteproxy_controller.go` (line 46)
- `cmd/thv-operator/controllers/virtualmcpserver_controller.go` (line 100)

### 3. Generated manifests

- `deploy/charts/operator/templates/clusterrole/role.yaml`

### 4. Chainsaw test assertions

- `test/e2e/chainsaw/operator/multi-tenancy/setup/assert-rbac-clusterrole.yaml`
- `test/e2e/chainsaw/operator/single-tenancy/setup/assert-rbac-clusterrole.yaml` (if exists)

## Fix

Remove `apply` from all verb lists. The `patch` verb already covers server-side apply operations. After fixing markers, run `task operator-manifests` to regenerate the ClusterRole manifest. Update Chainsaw assert files to match.

## Origin

Introduced in PR #620 (Jun 2025, multi-namespace support).

## Impact

None — the verb is harmless but misleading. Cleanup only.

ChrisJBurns:
## Summary

- Remove the invalid `apply` RBAC verb from all operator controller kubebuilder markers, `defaultRBACRules`, generated ClusterRole manifest, and Chainsaw test assertions
- `apply` is not a valid Kubernetes RBAC verb — server-side apply uses the `patch` verb (already present in all rules), so the `apply` entry was a silent no-op stored in etcd

Closes #3995

## Test plan

- [x] Operator controller unit tests pass (`go test ./cmd/thv-operator/controllers/...`)
- [ ] Chainsaw E2E tests pass in CI (single-tenancy and multi-tenancy)
- [ ] Verify no `apply` verb remains in RBAC-related files

### DIFF
--- a/deploy/charts/operator-crds/README.md
+++ b/deploy/charts/operator-crds/README.md
@@ -48,6 +48,35 @@ To ensure CRDs are upgraded alongside the chart, this chart places CRDs in `temp
 
 However, placing CRDs in `templates/` means they would be deleted when the Helm release is uninstalled, which could result in data loss. To prevent this, CRDs are annotated with `helm.sh/resource-policy: keep` by default (controlled by `crds.keep`). This ensures CRDs persist even after uninstalling the chart.
 
+## Important: Namespace Consistency
+
+When installing this chart, Helm stamps all CRDs with a `meta.helm.sh/release-namespace` annotation set to the namespace used at install time. This annotation **cannot be changed** by subsequent `helm upgrade` commands targeting a different namespace.
+
+You are free to install this chart in any namespace, but you **must use the same namespace consistently** for all future upgrades. If you plan to install the operator chart in `toolhive-system`, install the CRD chart there too:
+
+```shell
+helm upgrade -i toolhive-operator-crds oci://ghcr.io/stacklok/toolhive/toolhive-operator-crds -n toolhive-system --create-namespace
+```
+
+### Migrating from a Different Namespace
+
+If you previously installed the CRD chart without specifying a namespace (defaulting to `default`) and now want to upgrade using a different namespace, you will see an error like:
+
+```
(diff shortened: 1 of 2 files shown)


deploy/charts/operator/templates/clusterrole/role.yaml
```diff
@@ -47,37 +48,12 @@ rules:
...
-  - services
-  verbs:
-  - apply
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
```
aponcedeleonch, line 62: The chainsaw tests are failing because these got deleted. Not sure how that happened since this change should be unrelated to those

ChrisJBurns: Yeah I think its because removing apply caused the generator to merge services into an existing rule (since verbs now matched). However i think the chainsaw tests are failing because the assert role hasn't got the same permissions being removed. I'll check with claude to be sure

PR #4008, opened 2026-03-05
Automate HMAC secret management

yrobla:
Automatically generate and manage HMAC secrets when SessionManagementV2 is enabled in VirtualMCPServer resources, eliminating manual secret creation.

## Large PR Justification

This is an atomic PR that implements a single functionality and it's adding comprehensive tests. Cannot be split.

Related-to: #3867

### DIFF
--- a/pkg/vmcp/session/sessionv2_integration_test.go
+++ b/pkg/vmcp/session/sessionv2_integration_test.go
@@ -1,0 +1,498 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package session
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/hex"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/vmcp"
+	sessiontypes "github.com/stacklok/toolhive/pkg/vmcp/session/types"
+)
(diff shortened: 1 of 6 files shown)


pkg/vmcp/session/sessionv2_integration_test.go
```diff
@@ -0,0 +1,498 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
jerm-dro, line 1: suggestion: delete? Don't we already have tests like this?

test/e2e/thv-operator/virtualmcp/virtualmcp_hmac_secret_test.go
```diff
@@ -0,0 +1,318 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	mcpv1alpha1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1alpha1"
+	vmcpconfig "github.com/stacklok/toolhive/pkg/vmcp/config"
+)
+
+var _ = ginkgo.Describe("VirtualMCPServer HMAC Secret E2E Tests", func() {
```
jerm-dro, line 24: Suggestion: can you make these tests more behavioral? For example, you could create two clients with different identities and demonstrate they cannot use each others' sessions. That would also let you delete  more of the integration tests within vmcp/ that are coupled to implementation details (e.g. metadata headers).

PR #4009, opened 2026-03-05
fix: oauth issues and add tokenResponseMapping for non-standard providers

aron-muon:
## Summary

Fixes for OAuth-related issues discovered while deploying MCPRemoteProxy with embeddedAuthServer against Asana, Atlassian, and GovSlack, plus a new feature for non-standard token response formats.

### Bug Fixes

**1. Fix TOOLHIVE_DEBUG env var not enabling debug logging** (`cmd/thv-proxyrunner/main.go`, `cmd/thv/main.go`)

The logger was initialized in `main.go` before `viper.BindEnv("debug", "TOOLHIVE_DEBUG")` was called in `commands.go`, so the env var had no effect on log level. Moved the env var binding before logger initialization. Applied the same fix to the main `thv` binary which was also missing the env var binding.

**2. Propagate upstream user name and email into JWT claims** (`pkg/authserver/`)

The embedded auth server resolved user identity (name, email) from the upstream IDP via the userInfo endpoint but only stored the subject claim in the JWT. This caused audit logs to show `"user":"anonymous"` despite successful authentication. Now propagates name and email as standard OIDC claims (per OIDC Core Section 5.1).

Introduced a `UserClaims` struct to avoid parameter-ordering mistakes in `session.New()`, and added `NameClaimKey`/`EmailClaimKey` constants for consistency with existing claim key constants. Claims extraction from ID tokens now logs a warning on failure instead of silently ignoring the error.

**3. Fix remote URL path not forwarded to backend server** (`pkg/transport/`)

When the remote URL has a path (e.g., `https://mcp.asana.com/v2/mcp`), the proxy stripped it and only used the scheme+host as the target. Client requests to `/mcp` were forwarded to `https://mcp.asana.com/mcp` instead of `https://mcp.asana.com/v2/mcp`, causing 401 errors from Asana and Atlassian because the endpoint doesn't exist at the wrong path.

Added `WithRemoteBasePath` option to the transparent proxy that rewrites incoming request paths to the remote server's configured path.

### Feature

**4. Add tokenResponseMapping for non-standard OAuth token responses** (`pkg/authserver/upstream/`)

Some OAuth providers (e.g., GovSlack) nest token fields under non-standard paths. GovSlack returns `access_token` under `authed_user.access_token`, causing Go's oauth2 library to fail with `"server response missing access_token"`.

Added a `tokenResponseMapping` field to `OAuth2UpstreamConfig` that configures dot-notation paths (via gjson, already a dependency) for extracting token fields from non-standard response formats. Implemented as a response-rewriting `http.RoundTripper` that normalizes the response JSON before the oauth2 library parses it — keeping all of Go's oauth2 RFC 6749 compliance (error handling, token type validation, refresh via `TokenSource`). The `token_type` is always normalized to `"Bearer"` since non-standard providers may use different values (e.g., GovSlack uses `"user"`).

Example CRD usage:
```yaml
oauth2Config:
  tokenEndpoint: "https://slack-gov.com/api/oauth.v2.access"
  tokenResponseMapping:
    accessTokenPath: "authed_user.access_token"
    scopePath: "authed_user.scope"
```

## Review feedback addressed

- Introduced `session.UserClaims` struct to avoid swappable string parameters in `session.New()`
- Added `NameClaimKey`/`EmailClaimKey` constants for consistency with `TokenSessionIDClaimKey` and `ClientIDClaimKey`
- Log warning when optional ID token claims extraction fails (instead of silently ignoring)
- Removed unused `TokenTypePath` field — `token_type` is always normalized to `"Bearer"`, so the field was dead config
- Fixed `pathOrDefault` calls to use standard field names as defaults (`"refresh_token"`, `"expires_in"`, `"scope"`) instead of empty strings
- Used `io.LimitReader` for token response body reads to prevent unbounded memory allocation
- Fixed `Content-Length` header handling: use `Header.Del` instead of setting to empty string
- Ran `task operator-generate`, `task operator-manifests`, and `task crdref-gen` to regenerate deepcopy, CRD schemas, and API docs
- Applied the same `TOOLHIVE_DEBUG` fix to the main `thv` binary

## Test plan

- [x] All existing tests pass
- [x] New tests for remote path forwarding (5 test cases: Asana, Atlassian, GitHub, base path, full path scenarios)
- [x] New tests for token response mapping (rewrite unit tests, RoundTripper integration tests, error passthrough, nil mapping, validation)
- [x] Session tests updated for name/email JWT claims with UserClaims struct
- [x] Verified in production: Asana, Atlassian, and GovSlack all authenticate and forward requests successfully
- [x] Full build clean, golangci-lint clean

## Large PR Justification

- This is a large feature


cmd/thv-proxyrunner/main.go
```diff
@@ -15,6 +15,13 @@ import (
 )
 
 func main() {
+	// Bind TOOLHIVE_DEBUG env var early, before logger initialization.
+	// This must happen before viper.GetBool("debug") so the env var
+	// is available when configuring the log level.
+	if err := viper.BindEnv("debug", "TOOLHIVE_DEBUG"); err != nil {
+		slog.Error("failed to bind TOOLHIVE_DEBUG env var", "error", err)
```
jhrozek, line 22: I think we have the same bug in the main thv binary if you wanted to fix it there, too (although it's much more important in the proxy runner)

pkg/authserver/server/session/session.go
```diff
@@ -92,7 +97,7 @@ type Session struct {
 // Note: The remaining RFC 9068 required claims (iss, aud, exp, iat, jti) are
 // populated by Fosite during token generation. This session only initializes
 // custom claims that are not part of Fosite's standard JWT handling.
-func New(subject, idpSessionID, clientID string) *Session {
+func New(subject, idpSessionID, clientID, name, email string) *Session {
```
jhrozek: not a blocker for this patch, perhaps we could do it in a follow up but now the function has 5 string parameters, it might be too easy to swap them by accident.

Claude suggested this pattern:
```go
  type UserClaims struct {
      Name  string
      Email string
  }

  func New(subject, idpSessionID, clientID string, claims UserClaims) *Session {

  Call sites become:
  // Callback handler (real data):
  sess := session.New(subject, sessionID, pending.ClientID, session.UserClaims{
      Name:  name,
      Email: email,
  })

  // Deserialization template (zero value):
  sess := session.New("", "", "", session.UserClaims{})
```

pkg/authserver/upstream/oidc.go
```diff
@@ -265,9 +265,19 @@ func (p *OIDCProviderImpl) ExchangeCodeForIdentity(
 		"expires_at", tokens.ExpiresAt.Format(time.RFC3339),
 	)
 
+	// Extract optional standard claims (name, email) from ID token
+	var idClaims struct {
+		Name  string `json:"name"`
+		Email string `json:"email"`
+	}
+	// Best-effort: if claims extraction fails, we still have the subject
+	_ = validatedToken.Claims(&idClaims)
```
jhrozek: I think it's fine that the extraction is best-effort but do you think we should at least log an error when/if it happens?

pkg/authserver/server/session/session.go
```diff
@@ -108,6 +113,14 @@ func New(subject, idpSessionID, clientID string) *Session {
 		claimsExtra[ClientIDClaimKey] = clientID
 	}
 
+	// Add user profile claims per OIDC Core Section 5.1
+	if name != "" {
+		claimsExtra["name"] = name
```
jhrozek: again a nitpick, but I wonder if we should have for consistency also constants, it seems that we have `TokenSessionIDClaimKey` and `ClientIDClaimKey` but we use name and email verbatim

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
```diff
@@ -331,6 +331,45 @@ type OAuth2UpstreamConfig struct {
...
 	// +optional
 	Scopes []string `json:"scopes,omitempty"`
+
+	// TokenResponseMapping configures custom field extraction from non-standard token responses.
+	// Some OAuth providers (e.g., GovSlack) nest token fields under non-standard paths
+	// instead of returning them at the top level. When set, ToolHive performs the token
+	// exchange HTTP call directly and extracts fields using the configured dot-notation paths.
+	// If nil, standard OAuth 2.0 token response parsing is used.
+	// +optional
+	TokenResponseMapping *TokenResponseMapping `json:"tokenResponseMapping,omitempty"`
```
jhrozek, line 341: I think since this patch adds a pointer, you should run `task operator-generate` to regenerate the deepcopy files.

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
```diff
@@ -331,6 +331,45 @@ type OAuth2UpstreamConfig struct {
...
+	// AccessTokenPath is the dot-notation path to the access token in the response.
+	// Example: "authed_user.access_token"
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	AccessTokenPath string `json:"accessTokenPath"`
+
+	// TokenTypePath is the dot-notation path to the token type in the response.
+	// If not specified, defaults to "token_type".
+	// +optional
+	TokenTypePath string `json:"tokenTypePath,omitempty"`
```
jhrozek: I see the `AccessTokenPath` attribute being used in the round-tripper but I don't see the `TokenTypePath` attribute being used anywhere. Could we remove it for now or do you see it being useful later?

(Or did I misread and it is actually used? Always a possibility :-))

pkg/authserver/upstream/token_exchange.go
```diff
@@ -0,0 +1,139 @@
...
+		original["access_token"] = v.String()
+	}
+
+	// Always set token_type to "Bearer" for the oauth2 library.
+	// Non-standard providers may use different values (e.g., GovSlack uses "user")
+	// that the oauth2 library rejects. Since we're already using a custom mapping,
+	// the original token_type value is not meaningful for standard validation.
+	original["token_type"] = "Bearer"
+
+	if path := pathOrDefault(mapping.RefreshTokenPath, ""); path != "" {
```
jhrozek: This confused me, why is the code passing `""` as the default? Shouldn't it say e.g:
```go
path := pathOrDefault(mapping.RefreshTokenPath, "refresh_token")
```
?

pkg/authserver/upstream/token_exchange.go
```diff
@@ -0,0 +1,139 @@
...
+	if req.URL.String() != t.tokenURL {
+		return resp, nil
+	}
+
+	// Only rewrite successful responses (errors should pass through for proper error handling)
+	if resp.StatusCode != http.StatusOK {
+		return resp, nil
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
jhrozek: let's be paranoid and use a LimitReader

pkg/authserver/upstream/token_exchange.go
```diff
@@ -0,0 +1,139 @@
...
+	_ = resp.Body.Close()
+	if err != nil {
+		return nil, err
+	}
+
+	rewritten := rewriteTokenResponse(body, t.mapping)
+
+	resp.Body = io.NopCloser(bytes.NewReader(rewritten))
+	resp.ContentLength = int64(len(rewritten))
+	resp.Header.Set("Content-Length", "")
```
jhrozek: hmm, why are we setting the header to an empty value?

jhrozek (commented):
This is great work. I left some comments inline, the only that really need to be done is to re-run the generate task target to fix the deepcopy files.

the rest of the comments are mostly nits and questions.

Thanks a lot of for the PR!

pkg/authserver/server/session/session.go
aron-muon: Good point - added UserClaims struct with Name and Email fields

pkg/authserver/upstream/oidc.go
aron-muon: Changed to log a warning via slog.Warn when extraction fails

pkg/authserver/server/session/session.go
aron-muon: Added constants NameClaimKey = "name" and EmailClaimKey = "email"

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
aron-muon: Ran task operator-generate, task operator-manifests, and task crdref-gen to regenerate deepcopy, CRD schemas, and API docs

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
aron-muon: Actually yes it wasn't being used - it is normalized to "Bearer" token type anyway. Removed TokenTypePath in latest commit

pkg/authserver/upstream/token_exchange.go
aron-muon: Changed to use "refresh_token" default, same for ExpiresInPath → "expires_in" and ScopePath → "scope"

pkg/authserver/upstream/token_exchange.go
aron-muon: Wrapped io.ReadAll(resp.Body) with io.LimitReader(resp.Body, maxResponseSize)

pkg/authserver/upstream/token_exchange.go
aron-muon: Changed to `resp.Header.Del("Content-Length")` originally the code set it to an empty string which is incorrect, deleting the header lets Go use `resp.ContentLength` (already set correctly)

cmd/thv-proxyrunner/main.go
aron-muon: Added there as well!

aron-muon:
> This is great work. I left some comments inline, the only that really need to be done is to re-run the generate task target to fix the deepcopy files.
> 
> the rest of the comments are mostly nits and questions.
> 
> Thanks a lot of for the PR!

Thanks for the review! I appreciate the work ya'll have put into this - it works pretty well in our PoC!

jhrozek:
> ## Large PR Detected
> This PR exceeds 1000 lines of changes and requires justification before it can be reviewed.

eh, go away bot

jhrozek:
@aron-muon I should have ran CI earlier, sorry would you mind running `task docs`, submitting the changes it produces and pushing again?

aron-muon:
> @aron-muon I should have ran CI earlier, sorry would you mind running `task docs`, submitting the changes it produces and pushing again?

All set!

PR #4011, opened 2026-03-05
Free port holder before spawning remote workload restart

amirejaz:
## Problem

On restart, the PID in the status file becomes stale because:

1. The parent (`RunWorkloadDetached`) writes the new child's PID to the status file right after spawn, **before** the child successfully binds.
2. If the child fails to bind (e.g. "address already in use" because the old proxy still holds the port, or the port was not yet released after `stopProxyIfNeeded`), the child exits.
3. The status file then contains the dead child's PID, while the process actually holding the port is no longer tracked. That PID is stale and we can no longer reliably kill the port holder from the status file.

PR #3975 fixed the case where `stopProxyIfNeeded` was not called for some restart paths (e.g. unauthenticated). This change adds a safeguard for the remaining case where the port is still in use when we spawn the new child.

## Solution

Before spawning in `maybeSetupRemoteWorkload`, detect if the proxy port is in use, resolve the PID of the process holding it with `GetProcessOnPort`, kill it, and wait for it to exit so the port is free when the child binds.

## Changes

- **pkg/networking**: `GetProcessOnPort(port int) (int, error)` — Unix: `lsof -nP -t -iTCP:PORT -sTCP:LISTEN`; Windows: parse `netstat -ano` for LISTENING.
- **pkg/process**: `WaitForExit(ctx, pid, timeout)` — polls until the process exits or context/timeout.
- **pkg/workloads**: `freePortHolderIfNeeded` — called from `maybeSetupRemoteWorkload` before returning the runner; if the port is in use, kills the holder and waits for exit.

## Testing

- Unit tests for `GetProcessOnPort` (invalid port, free port, port in use)
- Unit tests for `WaitForExit` (already exited, context cancelled)
- Workload manager tests pass

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -26,6 +26,7 @@ jobs:
       - name: Run Trivy vulnerability scanner in repo mode
         uses: aquasecurity/trivy-action@e368e328979b113139d6f9068e03accaed98a518 # 0.34.1
         with:
+          version: 'v0.69.3'
           scan-type: 'fs'
           scan-ref: '.'
           format: 'sarif'
@@ -47,6 +48,7 @@ jobs:
       - name: Run Trivy configuration scanner
         uses: aquasecurity/trivy-action@e368e328979b113139d6f9068e03accaed98a518 # 0.34.1
         with:
+          version: 'v0.69.3'
           scan-type: 'config'
           scan-ref: './deploy/charts'
           format: 'sarif'


pkg/networking/port_unix.go
```diff
@@ -0,0 +1,56 @@
...
+	"os/exec"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// GetProcessOnPort returns the PID of the process listening on the given TCP port.
+// Returns 0 if the port is free or if the holder cannot be determined.
+// Uses lsof which is available on Linux and macOS.
+func GetProcessOnPort(port int) (int, error) {
```
aponcedeleonch: Consider using `gopsutil` instead of shelling out to `lsof`/`netstat`. It's already an indirect dependency (`github.com/shirou/gopsutil/v4`) and [`net.Connections("tcp")`](https://pkg.go.dev/github.com/shirou/gopsutil/v4/net#Connections) returns connections with PIDs, states, and local addresses — no output parsing needed.

This would eliminate both `port_unix.go` and `port_windows.go` (~125 lines) and replace them with a single cross-platform file:

```suggestion
func GetProcessOnPort(port int) (int, error) {
	if port <= 0 || port > 65535 {
		return 0, fmt.Errorf("invalid port %d", port)
	}

	conns, err := psnet.Connections("tcp")
	if err != nil {
		return 0, fmt.Errorf("failed to list connections: %%w", err)
	}

	for _, c := range conns {
		if c.Laddr.Port == uint32(port) && c.Status == "LISTEN" && c.Pid > 0 {
			return int(c.Pid), nil
		}
	}
	return 0, nil
}
```

With `psnet "github.com/shirou/gopsutil/v4/net"` — this would go in a single `port_process.go` file with no build tags.

pkg/process/wait.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"context"
+	"time"
+)
+
+// WaitForExit waits for the process with the given PID to exit.
+// It polls FindProcess every 50ms until the process is no longer running
+// or the context is cancelled or the timeout is reached.
+// Returns nil when the process has exited, or an error on timeout/context cancellation.
+func WaitForExit(ctx context.Context, pid int, timeout time.Duration) error {
```
aponcedeleonch: This function accepts both a `context.Context` and a `timeout` parameter, but the caller (`freePortHolderIfNeeded`) already wraps the context with `context.WithTimeout(ctx, 5*time.Second)`. This creates a redundant double-timeout and a subtle bug: the internal deadline check on line 33 happens *after* the ticker fires, so the first iteration always runs even if the deadline has already passed.

Drop the `timeout` param and rely solely on the context:

```suggestion
func WaitForExit(ctx context.Context, pid int) error {
	ticker := time.NewTicker(50 * time.Millisecond)
	defer ticker.Stop()

	for {
		alive, err := FindProcess(pid)
		if err != nil {
			return err
		}
		if !alive {
			return nil
		}

		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-ticker.C:
		}
	}
}
```

The caller already handles the timeout via `context.WithTimeout`, and the test would update to pass a context with deadline instead of a duration.

pkg/workloads/manager.go
```diff
@@ -868,6 +869,40 @@ func (d *DefaultManager) stopProxyIfNeeded(ctx context.Context, name, baseName s
...
+func (*DefaultManager) freePortHolderIfNeeded(ctx context.Context, runConfig *runner.RunConfig) {
+	if runConfig == nil || runConfig.Port <= 0 {
+		return
+	}
+
+	if networking.IsAvailable(runConfig.Port) {
+		return
+	}
+
+	portPID, err := networking.GetProcessOnPort(runConfig.Port)
```
aponcedeleonch, line 884: This kills whatever process is holding the port, but it could be a legitimate non-ToolHive process that happens to be on the same port (e.g., if the port was reassigned after the old proxy exited). Consider verifying the PID belongs to a ToolHive proxy before killing it — for example, by checking the process command line for the `thv` binary name or the `TOOLHIVE_DETACHED` env var. This would prevent accidentally killing unrelated processes.

pkg/networking/port.go
```diff
@@ -177,3 +180,39 @@ func ValidateCallbackPort(callbackPort int, clientID string) error {
 func IsPreRegisteredClient(clientID string) bool {
 	return clientID != ""
 }
+
+// GetProcessOnPort returns the PID of the process listening on the given TCP port.
+// Returns 0 if the port is free or if the holder cannot be determined.
+// Uses gopsutil which provides cross-platform support (Linux: /proc, Windows: GetExtendedTcpTable,
+// Darwin/FreeBSD: lsof).
+func GetProcessOnPort(port int) (int, error) {
+	if port <= 0 || port > 65535 {
```
reyortiz3: nit: `port > MaxPort`

pkg/networking/port.go
```diff
@@ -177,3 +180,39 @@ func ValidateCallbackPort(callbackPort int, clientID string) error {
...
+			pids = append(pids, int(c.Pid))
+		}
+	}
+
+	if len(pids) == 0 {
+		return 0, nil
+	}
+
+	sort.Ints(pids)
+	return pids[0], nil
```
reyortiz3: simpler?

```
for _, c := range conns {
		if c.Laddr.Port == uint32(port) && c.Status == "LISTEN" {
			return int(c.Pid), nil
		}
	}

	return 0, nil
```

pkg/process/toolhive_proxy.go
```diff
@@ -0,0 +1,123 @@
...
+		strings.Contains(lower, `\`+toolHiveBinaryName) ||
+		strings.HasPrefix(lower, toolHiveBinaryName+" ") ||
+		strings.Contains(lower, " "+toolHiveBinaryName+" ") ||
+		strings.HasSuffix(lower, " "+toolHiveBinaryName)
+}
+
+// IsToolHiveProxy returns true if the given PID belongs to a ToolHive proxy process.
+// Prefer IsToolHiveProxyForWorkload when the workload name is known to avoid
+// accidentally killing another workload's proxy.
+func IsToolHiveProxy(pid int) (bool, error) {
```
reyortiz3: this method is used for test only? if so, then remove it?

pkg/networking/port.go
amirejaz: done

pkg/networking/port.go
```diff
@@ -177,3 +179,27 @@ func ValidateCallbackPort(callbackPort int, clientID string) error {
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to get TCP connections: %w", err)
+	}
+
+	port32 := uint32(port) //nolint:gosec // G115 - port validated in [1, 65535]
+	for _, c := range conns {
+		if c.Laddr.Port != port32 || c.Status != "LISTEN" || c.Pid <= 0 {
+			continue
+		}
+		return int(c.Pid), nil
```
reyortiz3: nit: claude still suggests this simpler version with an inverted condition

```
if c.Laddr.Port == uint32(port) && c.Status == "LISTEN" { //nolint:gosec // G115 - port validated in [1, 65535]
			return int(c.Pid), nil
		}
```

approving as is anyways

reyortiz3 (dismissed):

PR #4012, opened 2026-03-05
Fix infinite reconciliation caused by PodTemplateSpec JSON comparison

aponcedeleonch:
## Summary

- **Fix infinite reconciliation loop** when a VirtualMCPServer has a `podTemplateSpec`: The previous approach compared full JSON of the freshly-generated deployment template against the cluster template. These always differed because the Kubernetes API server adds defaulted fields (`terminationGracePeriodSeconds`, `dnsPolicy`, `restartPolicy`, `schedulerName`, container `terminationMessagePath`, etc.) to the stored Deployment that are absent from the controller-generated template. Every reconcile detected a false "needs update", updated the deployment, which triggered another reconcile — an infinite loop.
- **Replace JSON comparison with SHA256 hash annotation**: Store a hash of the raw `PodTemplateSpec` input as a deployment annotation (`toolhive.stacklok.io/podtemplatespec-hash`). On subsequent reconciles, compare only the hash — if the user's input hasn't changed, no update is needed, regardless of K8s-defaulted fields on the cluster side.
- **Handle edge cases**: PodTemplateSpec added (no previous annotation), removed (annotation exists but spec is nil), and nil deployment annotations.

### How the previous flow was broken

1. User applies a VirtualMCPServer with `podTemplateSpec` (e.g., resource overrides)
2. Controller creates the Deployment — K8s API server stores it with added defaults
3. Next reconcile: `podTemplateSpecNeedsUpdate` generates a fresh template (no K8s defaults) and compares JSON against the cluster template (has K8s defaults) → **always differs**
4. Controller updates the Deployment → triggers another reconcile → back to step 3
5. Result: `DeploymentUpdated` events firing continuously (observed as x10+ over 30s)

## Test plan

- [x] Rewrote `TestVirtualMCPServerPodTemplateSpecNeedsUpdate` with 8 table-driven test cases covering all scenarios (matching hash, changed spec, added/removed spec, nil annotations, K8s defaults)
- [x] All existing PodTemplateSpec tests pass (`TestVirtualMCPServerPodTemplateSpecDeterministic`, `TestVirtualMCPServerPodTemplateSpecPreservesContainer`, `TestVirtualMCPServerPodTemplateSpecResourceOverride`)
- [x] Full controller test suite passes

### DIFF
--- a/cmd/thv-operator/controllers/virtualmcpserver_controller.go
+++ b/cmd/thv-operator/controllers/virtualmcpserver_controller.go
@@ -1026,7 +1026,7 @@ func (r *VirtualMCPServerReconciler) podTemplateMetadataNeedsUpdate(
 // fields like terminationGracePeriodSeconds, dnsPolicy, etc.), this compares a SHA256 hash of
 // the raw PodTemplateSpec input stored as a deployment annotation.
 func (*VirtualMCPServerReconciler) podTemplateSpecNeedsUpdate(
-	_ context.Context,
+	ctx context.Context,
 	deployment *appsv1.Deployment,
 	vmcp *mcpv1alpha1.VirtualMCPServer,
 	_ []workloads.TypedWorkload,
@@ -1048,6 +1048,7 @@ func (*VirtualMCPServerReconciler) podTemplateSpecNeedsUpdate(
 	expectedHash, err := checksum.HashRawJSON(vmcp.Spec.PodTemplateSpec.Raw)
 	if err != nil {
 		// If we can't hash, assume update is needed
+		log.FromContext(ctx).Error(err, "Failed to hash PodTemplateSpec, assuming update needed")
 		return true
 	}
 	return deployment.Annotations[podTemplateSpecHashAnnotation] != expectedHash


aponcedeleonch:
@JAORMX Good call! I checked and the operator already has a `checksum` package at `cmd/thv-operator/pkg/runconfig/configmap/checksum/` — however its existing functions (`ComputeConfigMapChecksum`, etc.) are specific to ConfigMap hashing: they iterate over Go maps (Data, Labels, Annotations) with sorted keys. That pattern doesn't apply here since we're hashing raw JSON bytes from `runtime.RawExtension`, not parsed Go structs.

So I added a new reusable `HashRawJSON(raw []byte) (string, error)` function to that same checksum package. It unmarshals the raw JSON into `any`, then re-marshals it — Go's `json.Marshal` sorts map keys alphabetically, which gives us deterministic output regardless of the original field ordering. The SHA256 hash is then computed on this canonical form.

Both call sites in the controller (`podTemplateSpecNeedsUpdate`) and deployment (`buildDeploymentMetadataForVmcp`) now use this shared function instead of inline `sha256.Sum256` on raw bytes. I also confirmed there are no other direct `sha256` usages left in the operator controllers — it's all consolidated in the checksum package now.

Tests added for `HashRawJSON` covering: same fields in different order → same hash, nested objects, different values → different hash, and invalid JSON error handling.

ChrisJBurns:
I would have thought that we'd be able to reuse the `podTemplateSpec` stuff for the `MCPServer`?

jerm-dro:
@ChrisJBurns Good question! It's a natural thing to wonder about, but the two controllers handle `PodTemplateSpec` quite differently under the hood.

**MCPServer** never directly applies the patch to the deployment's `spec.template`. Instead, it [serializes the user's `PodTemplateSpec` into JSON and passes it as a `--k8s-pod-patch=` CLI argument](https://github.com/stacklok/toolhive/blob/main/cmd/thv-operator/controllers/mcpserver_controller.go#L969) to the toolhive proxy runner — change detection then [compares that string against what's already in `container.Args`](https://github.com/stacklok/toolhive/blob/main/cmd/thv-operator/controllers/mcpserver_controller.go#L1623). Since both sides are derived from user input, K8s-defaulted fields never enter the picture. (`equality.Semantic.DeepEqual` is used there too, but only for things like [env vars](https://github.com/stacklok/toolhive/blob/main/cmd/thv-operator/controllers/mcpserver_controller.go#L1597) and [resource requirements](https://github.com/stacklok/toolhive/blob/main/cmd/thv-operator/controllers/mcpserver_controller.go#L1648) where both sides are freshly generated.)

**VirtualMCPServer** takes a different path: [`applyPodTemplateSpecToDeployment`](https://github.com/stacklok/toolhive/blob/main/cmd/thv-operator/controllers/virtualmcpserver_deployment.go#L763) applies the patch directly to the deployment's `spec.template` via strategic merge patch. Once that rendered template is stored in the cluster, the API server injects defaults (`terminationGracePeriodSeconds`, `dnsPolicy`, etc.). A freshly generated template (no defaults) compared against the cluster version (with defaults) always looks different — hence the infinite loop. The hash-based approach in [`podTemplateSpecNeedsUpdate`](https://github.com/stacklok/toolhive/blob/main/cmd/thv-operator/controllers/virtualmcpserver_controller.go#L1028) avoids this by comparing the raw user input rather than the rendered deployment.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -1022,51 +1021,36 @@ func (r *VirtualMCPServerReconciler) podTemplateMetadataNeedsUpdate(
...
-		return true
-	}
-
-	expectedJSON, err := json.Marshal(expectedDeployment.Spec.Template)
+	// Compare hash of the raw PodTemplateSpec input against the stored annotation.
+	// Avoids comparing full rendered templates which always differ due to
+	// Kubernetes-defaulted fields (terminationGracePeriodSeconds, dnsPolicy, etc.).
+	// Uses HashRawJSON to ensure deterministic hashing regardless of JSON field ordering.
+	expectedHash, err := checksum.HashRawJSON(vmcp.Spec.PodTemplateSpec.Raw)
 	if err != nil {
```
jerm-dro, line 1049: Nit: let's also log the error. An error here could result in unexplained reconcile loops.

jerm-dro (approved):

cmd/thv-operator/controllers/virtualmcpserver_controller.go
aponcedeleonch: good idea, done!

JAORMX (approved):

PR #4019, opened 2026-03-05
feat(vmcp): log active sessions on creation, expiry, and periodically

### LINKED ISSUE #3876: [vMCP] Expose active backend sessions on health endpoint
Depends on #3866

Session lifecycle events and periodic session counts should be logged for operational debugging and compliance auditing when sessionManagementV2 is enabled.

## Required logging

**On session creation** (`CreateSession`):
- Log at `Info` level with: session ID, backend count, backend names

**On session termination/expiry** (`Terminate`):
- Log at `Info` level with: session ID, backend count

**Periodic active session summary** (every 1 minute):
- Log at `Info` level with: total active session count, per-backend session counts

## Implementation

- `Manager.CreateSession()`: log at `Info` when a session is fully created
- `Manager.Terminate()`: log at `Info` with backend count for MultiSessions; log placeholder termination separately
- `Manager.StartPeriodicLogging(ctx, interval)`: background goroutine that calls internal `logActiveSessions()` on a ticker
- `logActiveSessions()`: iterates sessions via `Range()`, aggregates per-backend counts, logs summary
- Called from `Server.Start()` when `SessionManagementV2` is enabled, with a 1-minute default interval

## Acceptance Criteria

- Session creation is logged at `Info` with session ID, backend count, and backend names
- Session termination is logged at `Info` with session ID and backend count
- Active sessions are periodically logged at `Info` with total count and per-backend counts
- No session IDs or backend session IDs are exposed over the network (logging only, to operators)
- The periodic logger stops when the server context is cancelled
- The implementation degrades gracefully with distributed storage backends (logs a one-time warning)

---

yrobla:
Log session lifecycle events and periodic active session counts for operational debugging and compliance auditing when sessionManagementV2 is enabled.

## Changes

- **`Manager.CreateSession()`**: upgraded log to `Info` with session ID, backend count, and backend names
- **`Manager.Terminate()`**: log `Info` with session ID and backend count when a MultiSession expires; separate log for placeholder termination
- **`Manager.StartPeriodicLogging(ctx, interval)`**: new method that starts a background goroutine logging active session counts every minute
- **`Manager.logActiveSessions()`**: internal helper that iterates sessions via `Range()`, aggregates per-backend session counts, and logs a summary; logs a one-time warning when distributed storage makes enumeration unavailable
- **`Server.Start()`**: calls `StartPeriodicLogging` when `SessionManagementV2` is enabled

## Removed

- `/status` endpoint session fields (`Sessions`, `SessionsStatus`, `BackendUsage`) — session data is now surfaced through logs only
- `ListActiveSessions()` from `SessionManager` interface

Closes: #3876


jerm-dro (commented):
Is this change really necessary? I get that it'd be nice to have visibility into active sessions, but exposing an endpoint seems heavy-handed. Alternative approaches:
- periodically log all the active sessions in memory
- log sessions as they are created and expired

What do you think?

yrobla:
> Is this change really necessary? I get that it'd be nice to have visibility into active sessions, but exposing an endpoint seems heavy-handed. Alternative approaches:
> 
> * periodically log all the active sessions in memory
> * log sessions as they are created and expired
> 
> What do you think?

ok this task came generated by claude, based on the rfc. A bit overkill, i agree. So i modified the issue to log content, as you were saying, and i modified the pr.

pkg/transport/session/manager.go
```diff
@@ -219,6 +219,13 @@ func (m *Manager) Stop() error {
 	return nil
 }
 
+// SupportsRange reports whether the current storage backend supports Range().
+// Returns true only for LocalStorage; distributed backends (Redis, Valkey) return false.
+func (m *Manager) SupportsRange() bool {
+	_, ok := m.storage.(*LocalStorage)
+	return ok
+}
+
```
jerm-dro, line 228: Blocker: How about we wait to implement this until we can support all session storage types?

We're imminently going to introduce Redis storage. That will come along with a LRU cache for in-memory sessions. That will be easier to instrument in a way that works for single and multi replica use cases.

jerm-dro (changes requested):

pkg/transport/session/manager.go
yrobla: ok leaving on draft for now then

amirejaz:
@yrobla  could you take a look at this draft PR and confirm if it’s still relevant?

yrobla:
> @yrobla could you take a look at this draft PR and confirm if it’s still relevant?

well, we agreed to implement it at some point, but it was not critical. If you don't mind, i prefer to keep them as draft, to resume the work when needed.

PR #4028, opened 2026-03-06
changes from review

yrobla:
Simplify code, remove tests and move related packages

Related-to: #3867

## Large PR Justification

This is a code refactor, that means moving and deleting files, so that's the reason for the big diff


pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -165,7 +165,7 @@ func (sm *Manager) CreateSession(
 	// Build the fully-formed MultiSession using the SDK-assigned session ID.
 	// Sessions created with an identity are bound to that identity (allowAnonymous=false).
 	// Sessions created without an identity allow anonymous access (allowAnonymous=true).
-	allowAnonymous := vmcpsession.ShouldAllowAnonymous(identity)
+	allowAnonymous := identity == nil || identity.Token == ""
```
JAORMX: is this what we want in this case? what about handling anonymous? that still seems to be what the comment says.

pkg/vmcp/session/session.go
```diff
@@ -58,3 +59,36 @@ type MultiSession interface {
...
+	// Re-exported from types package for convenience.
+	MetadataKeyTokenHash = sessiontypes.MetadataKeyTokenHash
+
+	// MetadataKeyTokenSalt is the session metadata key that holds the hex-encoded
+	// random salt used for HMAC-SHA256 token hashing. Each session has a unique salt
+	// to prevent attacks across multiple sessions.
+	//
+	// Re-exported from types package for convenience.
+	MetadataKeyTokenSalt = sessiontypes.MetadataKeyTokenSalt
+)
```
jerm-dro, line 23: Blocker: These are only needed for integration tests. I would delete these and refactor the integration tests or delete them so they don't rely on this implementation detail.

pkg/vmcp/session/internal/security/security.go
```diff
@@ -0,0 +1,434 @@
...
+	SHA256HexLen = 64
+
+	// MetadataKeyTokenHash is the session metadata key for the token hash.
+	// Imported from types package to ensure consistency across all packages.
+	MetadataKeyTokenHash = sessiontypes.MetadataKeyTokenHash
+
+	// MetadataKeyTokenSalt is the session metadata key for the token salt.
+	// Imported from types package to ensure consistency across all packages.
+	MetadataKeyTokenSalt = sessiontypes.MetadataKeyTokenSalt
+)
```
jerm-dro: nit: these can be private once we delete the re-export

pkg/vmcp/session/internal/security/security.go
```diff
@@ -0,0 +1,434 @@
...
+
+// GenerateSalt generates a cryptographically secure random salt for token hashing.
+// Returns 16 bytes of random data from crypto/rand.
+//
+// Each session should have a unique salt to provide additional entropy and prevent
+// attacks that work across multiple sessions.
+func GenerateSalt() ([]byte, error) {
```
jerm-dro: nit: this can be private

pkg/vmcp/session/internal/security/security.go
```diff
@@ -0,0 +1,434 @@
...
+// The raw token is never stored — only the hash.
+//
+// Parameters:
+//   - token: The bearer token to hash
+//   - secret: Server-managed HMAC secret (should be 32+ bytes)
+//   - salt: Per-session random salt (typically 16 bytes)
+//
+// Security: Uses HMAC-SHA256 instead of plain SHA256 to prevent rainbow table
+// attacks and offline brute force if session state leaks from Redis/Valkey.
+func HashToken(token string, secret, salt []byte) string {
```
jerm-dro: nit: this can/should be private as well. The only caller that needs it public is token_binding_test. It just seems like those tests are in the wrong package.

pkg/vmcp/session/internal/security/security.go
```diff
@@ -0,0 +1,434 @@
...
+		arguments map[string]any,
+		meta map[string]any,
+	) (*vmcp.ToolCallResult, error)
+
+	// ReadResource reads the resource at the given URI.
+	ReadResource(ctx context.Context, caller *auth.Identity, uri string) (*vmcp.ResourceReadResult, error)
+
+	// GetPrompt retrieves the named prompt with the given arguments.
+	GetPrompt(ctx context.Context, caller *auth.Identity, name string, arguments map[string]any) (*vmcp.PromptGetResult, error)
+}
```
jerm-dro: Blocker: Instead of defining a new interface, can we reuse `MultiSession`? 

That would require moving `MultiSession` to the `types` package, but it will make this code easier to understand. As a reader, it's easier to see "oh, session hijack prevention is just a layer on top of any MultiSession for added protection. It does not modify the interface."

Introducing a new interface makes it more confusing, because readers have to understand how the decorator and internal type relate.

pkg/vmcp/session/internal/security/security.go
```diff
@@ -0,0 +1,434 @@
...
+// The decorator is applied by PreventSessionHijacking to ALL sessions (both authenticated
+// and anonymous). For authenticated sessions, it validates the caller's token matches
+// the creator's token. For anonymous sessions (allowAnonymous=true), it allows nil
+// callers and prevents session upgrade attacks by rejecting any token presentation.
+//
+// The decorator holds the underlying session as an interface{} to avoid circular
+// dependencies, and implements pass-through methods by type-asserting to the
+// necessary interfaces.
+type HijackPreventionDecorator struct {
+	session interface{} // The underlying session (stored as interface{} to avoid circular deps)
```
jerm-dro: Blocker: this should be `MultiSession` and we should embed the interface, so we don't have to redefine every method. We'll only define the ones that require hijack prevention.

pkg/vmcp/session/internal/security/security.go
```diff
@@ -0,0 +1,434 @@
...
+//
+// The decorator is applied by PreventSessionHijacking to ALL sessions (both authenticated
+// and anonymous). For authenticated sessions, it validates the caller's token matches
+// the creator's token. For anonymous sessions (allowAnonymous=true), it allows nil
+// callers and prevents session upgrade attacks by rejecting any token presentation.
+//
+// The decorator holds the underlying session as an interface{} to avoid circular
+// dependencies, and implements pass-through methods by type-asserting to the
+// necessary interfaces.
+type HijackPreventionDecorator struct {
```
jerm-dro: nit: this can be private

pkg/vmcp/session/internal/security/security.go
```diff
@@ -0,0 +1,434 @@
...
+	// Validate caller identity
+	if err := d.validateCaller(caller); err != nil {
+		return nil, err
+	}
+
+	return d.session.(HijackableSession).ReadResource(ctx, caller, uri)
+}
+
+// GetPrompt validates the caller identity before delegating to the underlying session.
+func (d *HijackPreventionDecorator) GetPrompt(
```
jerm-dro: nit: For all these methods, let's use a value receiver, because we don't need to mutate state:
`func (d HijackPreventionDecorator)`.

pkg/vmcp/session/internal/security/security.go
```diff
@@ -0,0 +1,434 @@
...
+// The return type is concrete to eliminate the need for runtime casts at call sites.
+//
+// Returns an error if:
+//   - session doesn't implement SessionMetadataWriter interface
+//   - salt generation fails
+func PreventSessionHijacking(
+	session interface{},
+	hmacSecret []byte,
+	identity *auth.Identity,
+	allowAnonymous bool,
```
jerm-dro: Blocker: let's make `allowAnonymous` calculated within `PreventSessionHijacking`. It appears to be a implementation detail and we can derive whether `allowAnonymous` is true/false from the `identity`.

If there are tests that depend on providing this value, then I think they should either be deleted / refactored to not depend on explicitly providing this value. I would imagine tests can implicitly enable / disable `allowAnonymous` by providing an `identity` with a token.

jerm-dro (changes requested):

pkg/vmcp/session/internal/security/security.go
```diff
@@ -0,0 +1,286 @@
...
+//   - Makes defensive copies of secret and salt to prevent external mutation
+//   - Uses constant-time comparison to prevent timing attacks
+//   - Prevents session upgrade attacks (anonymous → authenticated)
+//   - Raw tokens are never stored, only HMAC-SHA256 hashes
+//
+// Returns an error if:
+//   - session doesn't implement sessiontypes.MultiSession interface
+//   - salt generation fails
+func PreventSessionHijacking(
+	session interface{},
```
jerm-dro, line 228: blocker: Session should be a `MultiSession`

pkg/vmcp/session/internal/security/hijack_prevention_test.go
```diff
@@ -0,0 +1,232 @@
...
+			decorator := &hijackPreventionDecorator{
+				MultiSession:   baseSession,
+				allowAnonymous: tt.allowAnonymous,
+				boundTokenHash: tt.boundTokenHash,
+				tokenSalt:      testTokenSalt,
+				hmacSecret:     testSecret,
+			}
+
+			// Test validateCaller directly on the decorator
+			err := decorator.validateCaller(tt.caller)
```
jerm-dro, line 152: blocker: rather than calling `validateCaller` can you call each of the decorated methods (e.g. callTool, prompt, and so on)?

That would verify the code is implemented and integrated correctly.

pkg/vmcp/session/internal/security/hijack_prevention_test.go
```diff
@@ -0,0 +1,232 @@
...
+			errChan <- err
+		}()
+	}
+
+	// Wait for all goroutines and assert in main goroutine (thread-safe)
+	for i := 0; i < numGoroutines; i++ {
+		err := <-errChan
+		assert.NoError(t, err, "concurrent validation should succeed")
+	}
+}
```
jerm-dro, line 196: Suggestion: delete - this is unnecessary, since the decorator does not modify its state.

PR #4030, opened 2026-03-06
Add disableWorkloadRBAC flag to skip per-workload RBAC creation

JAORMX:
## Summary

- Adds `DISABLE_WORKLOAD_RBAC` env var and `operator.rbac.disableWorkloadRBAC` Helm value (default: `false`)
- When enabled, the operator skips creating per-workload ServiceAccount, Role, and RoleBinding resources
- Conditionally removes `roles`/`rolebindings` permissions and ServiceAccount write verbs from the operator's ClusterRole
- Conditionally excludes the registry-api ClusterRole and ClusterRoleBinding from the chart

## Motivation

Some Kubernetes platform teams enforce strict policies on which cluster-scoped resources an operator's ClusterRole may reference. In environments managed by GitOps tools like ArgoCD, app-projects must whitelist every cluster-scoped resource the operator needs — including `roles` and `rolebindings` permissions used for dynamic RBAC creation at runtime. This creates friction for adoption in security-conscious environments.

By allowing the operator to opt out of per-workload RBAC management, platform teams can:

- **Pre-provision RBAC via GitOps**: Ship ServiceAccount, Role, and RoleBinding alongside each workload CR in dedicated per-workload Helm charts or Kustomize overlays, keeping RBAC under full version control and audit
- **Satisfy ArgoCD app-project constraints**: The operator's ClusterRole no longer needs `roles`/`rolebindings` permissions, eliminating the need to whitelist those resources in app-project definitions
- **Enforce least-privilege at the operator level**: The operator only retains read access to ServiceAccounts (`get/list/watch`) and loses all write permissions for RBAC resources
- **Support policy engines**: Works cleanly with OPA/Gatekeeper or Kyverno policies that restrict which controllers may create RBAC resources

This is PR1 of a two-part effort. A follow-up PR will add per-workload Helm charts that bundle SA + Role + RoleBinding + CR for each workload type, providing a turnkey solution for externally-managed RBAC.

## Changes

**Go code:**
- New `DISABLE_WORKLOAD_RBAC` constant and flag reading in `main.go`, passed to all controller setup functions
- `DisableWorkloadRBAC bool` field added to `MCPServerReconciler`, `MCPRemoteProxyReconciler`, and `VirtualMCPServerReconciler` — each guards `ensureRBACResources()` with an early return
- `registryapi.manager` accepts the flag and guards its own `ensureRBACResources()`
- `NewMCPRegistryReconciler` forwards the flag to the registry API manager

**Helm charts:**
- `operator.rbac.disableWorkloadRBAC` value in `values.yaml` and `values-openshift.yaml`
- `DISABLE_WORKLOAD_RBAC` env var in `deployment.yaml`
- ClusterRole: `serviceaccounts` split into its own rule block with conditional write verbs; `roles`/`rolebindings` block wrapped in conditional
- `registry-api-clusterrole.yaml` and `registry-api-clusterrolebinding.yaml` wrapped in conditional

**Tests:**
- Unit test per controller verifying `ensureRBACResources` returns nil and creates no resources when disabled
- CI values file (`externalRBAC-values.yaml`) for Helm template linting

## How to use

```yaml
# values.yaml
operator:
  rbac:
    disableWorkloadRBAC: true
```

When this flag is set, users must pre-create the following resources for each workload:
- **MCPServer**: `<name>-proxy-runner` ServiceAccount/Role/RoleBinding + `<name>-mcp-server` ServiceAccount
- **MCPRemoteProxy**: `<name>-remote-proxy-runner` ServiceAccount/Role/RoleBinding
- **VirtualMCPServer**: `<name>-vmcp` ServiceAccount/Role/RoleBinding
- **MCPRegistry**: `<name>-registry-api` ServiceAccount/Role/RoleBinding

Deployments will fail to schedule if the required ServiceAccounts are not present — this is a safe fail-closed behavior.

## Test plan

- [x] `task test` — all unit tests pass (exit 0)
- [x] `helm template` with `externalRBAC-values.yaml` — ClusterRole has no roles/rolebindings, SA is read-only, registry-api resources absent
- [x] `helm template` with defaults — output unchanged from main
- [x] `task helm-docs` — README regenerated

### DIFF
--- a/deploy/charts/operator/templates/clusterrole/workload-rbac-role.yaml
+++ b/deploy/charts/operator/templates/clusterrole/workload-rbac-role.yaml
@@ -1,0 +1,35 @@
+{{- if not .Values.operator.rbac.disableWorkloadRBAC }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: toolhive-operator-workload-rbac-role
+  labels:
+    {{- include "toolhive.labels" . | nindent 4 }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  - rolebindings
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+{{- end }}
--- a/cmd/thv-operator/controllers/mcpregistry_controller.go
+++ b/cmd/thv-operator/controllers/mcpregistry_controller.go
@@ -67,10 +67,6 @@ func NewMCPRegistryReconciler(k8sClient client.Client, scheme *runtime.Scheme, d
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 //
-// For creating registry-api RBAC resources
-// +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings,verbs=get;list;watch;create;update;patch;delete
-//
 // For granting registry-api permissions (operator must have these to grant them via Role)
 // +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpservers;mcpremoteproxies;virtualmcpservers,verbs=get;list;watch
 // +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=httproutes;gateways,verbs=get;list;watch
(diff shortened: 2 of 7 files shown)


cmd/thv-operator/main.go
```diff
@@ -46,6 +46,10 @@ const (
 	featureServer   = "ENABLE_SERVER"
 	featureRegistry = "ENABLE_REGISTRY"
 	featureVMCP     = "ENABLE_VMCP"
+	// disableWorkloadRBAC disables per-workload RBAC management (ServiceAccount, Role, RoleBinding).
+	// When enabled, the operator will not create RBAC resources for workloads,
+	// allowing them to be managed externally (e.g., via per-workload Helm charts).
```
ChrisJBurns, line 51: I'm not sure if we want to create Helm Charts for this - Seems a bit overkill? It will only be 3 resources?

deploy/charts/operator/templates/registry-api-clusterrole.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.operator.rbac.disableWorkloadRBAC }}
```
ChrisJBurns, line 1: I don't actually think this is used anyways. This was added previously by RedHat but I don't think it's used.

deploy/charts/operator/templates/registry-api-clusterrolebinding.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.operator.rbac.disableWorkloadRBAC }}
```
ChrisJBurns, line 1: Same here

deploy/charts/operator/templates/clusterrole/workload-rbac-role.yaml
```diff
@@ -0,0 +1,35 @@
+{{- if not .Values.operator.rbac.disableWorkloadRBAC }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
ChrisJBurns, line 3: How is this auto generated? I'm not able to see the kubebuilder annotations

JAORMX: this isn't. I had to do this manually because kubebuilder kept removing my templating.

ChrisJBurns: Yeah, I had a feeling haha! We should remove the kubebuilder anyways, its better to just do it manually and not have the mix of automatic/manual maintenance of them.

PR #4036, opened 2026-03-06
fix: refresh upstream tokens transparently instead of forcing re-auth

aron-muon:
## Summary

Users are forced to fully re-authenticate with upstream OAuth providers every time the upstream access token expires (controlled by `accessTokenLifespan`), even though valid refresh tokens exist in storage.

The root cause is that upstream access tokens and refresh tokens are stored together in a single storage entry, with the entry's TTL/expiry set to the access token's expiry. When the access token expires, the entry is deleted (Redis) or marked expired (memory) — losing the refresh token, which is typically valid for 30-90 days or longer depending on the provider. The `upstreamswap` middleware has no refresh path and returns 401, forcing full re-authentication.

This affects both Redis and in-memory storage backends — Redis deletes the key via TTL, and memory storage's cleanup goroutine removes the expired entry.

## Type of change

- [x] Bug fix

## Root cause

The storage model bundles access + refresh tokens in a single entry (`UpstreamTokens` struct). The entry TTL is derived from the access token's `ExpiresAt`, so when the access token expires:

1. **Storage deletes or expires the entry** — losing the still-valid refresh token
2. **`GetUpstreamTokens` returns `nil, ErrExpired`** — discarding the token data
3. **Middleware returns 401** — no refresh path exists

Ideally, access and refresh tokens would be stored separately with independent TTLs matching their actual lifetimes. This fix extends the bundled entry's TTL as a pragmatic solution; a future refactor could separate them for cleaner lifecycle management.

### Storage layer (`pkg/authserver/storage/`)
- Extended upstream token entry TTL by `DefaultRefreshTokenTTL` (30 days) in both Redis and memory storage, so refresh tokens survive past access token expiry
- Changed `GetUpstreamTokens` to return token data alongside `ErrExpired` (instead of `nil`) so callers can use the refresh token
- Memory storage now checks the token's own `ExpiresAt` (access token expiry) rather than the entry's `expiresAt` (storage TTL) for the expired check

### Token refresher (`pkg/authserver/refresher.go`)
- New `UpstreamTokenRefresher` interface in `storage/types.go`
- Implementation wraps `upstream.OAuth2Provider.RefreshTokens()` + `UpstreamTokenStorage.StoreUpstreamTokens()`
- Preserves binding fields (ProviderID, UserID, UpstreamSubject, ClientID) across refresh
- Handles refresh token rotation (keeps old refresh token if provider doesn't issue a new one)

### Plumbing
- Exposed refresher through `Server` → `EmbeddedAuthServer` → `Runner` → `MiddlewareRunner` using the same lazy accessor pattern as `GetUpstreamTokenStorage`

### Middleware (`pkg/auth/upstreamswap/`)
- Middleware now attempts transparent refresh before returning 401
- Extracted `getOrRefreshUpstreamTokens` helper to keep cyclomatic complexity under lint threshold
- Only requires re-auth when the refresh token itself is invalid/revoked

## Production validation

Deployed to a production cluster with Redis (AWS Valkey) storage (we use a sentinel emulator which basically just returns the Valkey URL in each case. One of the little clever ways we could use Valkey). All four upstream providers successfully refreshed tokens transparently — no user re-authentication required:

| Provider | Token Endpoint | Access Token Lifetime | Refresh Token Rotated |
|----------|---------------|----------------------|----------------------|
| Atlassian | `cf.mcp.atlassian.com/v1/token` | 1 hour | Yes |
| Asana | `app.asana.com/-/oauth_token` | 1 hour | Yes |
| Slack (GovSlack) | `slack-gov.com/api/oauth.v2.access` | 12 hours | Yes |
| Google | `oauth2.googleapis.com/token` | 1 hour | Yes |

Redis TTLs confirmed updated to ~30 days (previously ~1 hour). GitHub has not yet expired its 8-hour access token but uses the same code path.

## Test plan

- [x] Updated storage tests to verify tokens returned alongside `ErrExpired`
- [x] Updated cleanup tests for extended TTL
- [x] Updated middleware tests with refresher parameter
- [x] All existing unit tests pass
- [x] Build clean, golangci-lint clean
- [x] Deployed to production — verified transparent refresh for Atlassian, Asana, Slack, and Google

Yes — upstream OAuth sessions now persist beyond the access token lifetime. Users will no longer be forced to re-authenticate as long as their upstream refresh token is valid (typically 30 days to indefinite depending on the provider).

## Large PR Justification


jhrozek:
hey @aron-muon thanks a lot for the patches this was actually something we had on our roadmap so the feature makes complete sense 🙏🏻 

I'll make sure to review the PR as soon as possible.

pkg/authserver/refresher.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/stacklok/toolhive/pkg/authserver/upstream"
+)
+
+// upstreamTokenRefresher implements storage.UpstreamTokenRefresher by wrapping
+// an upstream OAuth2Provider (for token refresh) and UpstreamTokenStorage (for
+// persisting the refreshed tokens).
+type upstreamTokenRefresher struct {
+	provider upstream.OAuth2Provider
+	storage  storage.UpstreamTokenStorage
+}
```
jhrozek, line 22: Nit: ToolHive convention is to add a compile-time interface check for unexported types that implement exported interfaces. This catches accidental drift early.

```suggestion
}

// Compile-time check that upstreamTokenRefresher implements storage.UpstreamTokenRefresher.
var _ storage.UpstreamTokenRefresher = (*upstreamTokenRefresher)(nil)
```

pkg/auth/upstreamswap/middleware.go
```diff
@@ -233,3 +229,71 @@ func createMiddlewareFunc(cfg *Config, storageGetter StorageGetter) types.Middle
...
 }
+
+// getOrRefreshUpstreamTokens retrieves upstream tokens from storage, automatically
+// refreshing them if expired and a refresh token is available.
+func getOrRefreshUpstreamTokens(
+	ctx context.Context,
+	stor storage.UpstreamTokenStorage,
+	sessionID string,
+	refresherGetter RefresherGetter,
+) (*storage.UpstreamTokens, error) {
```
jhrozek, line 248: Not a blocker for this PR, but worth flagging: when multiple concurrent requests hit `getOrRefreshUpstreamTokens` with the same expired session, each one will independently call `RefreshAndStore`. With providers that rotate refresh tokens (issuing a single-use replacement), all but the first caller will use a stale refresh token and fail.

A `singleflight.Group` keyed on `sessionID` would collapse concurrent refreshes into one. I've actually been working on something similar in a parallel branch, so I'll address this in a follow-up rather than piling onto this PR.

pkg/auth/upstreamswap/middleware.go
```diff
@@ -207,16 +213,6 @@ func createMiddlewareFunc(cfg *Config, storageGetter StorageGetter) types.Middle
...
-			// Defense in depth: some storage implementations may return tokens
-			// without checking expiry (the interface does not require it).
-			if tokens.IsExpired(time.Now()) {
-				slog.Warn("Upstream tokens expired",
-					"middleware", "upstreamswap")
-				writeUpstreamAuthRequired(w)
-				return
-			}
-
 			// 6. Inject access token
```
jhrozek: Nit: step numbering jumps from 4 to 6 — step 5 (the old explicit expiry check) was removed but the subsequent comment wasn't renumbered.

jhrozek (commented):
Nice work on transparent token refresh! A few minor comments below.

pkg/authserver/storage/redis_test.go
```diff
@@ -622,19 +622,24 @@ func TestRedisStorage_UpstreamTokens(t *testing.T) {
...
 			retrieved, err := s.GetUpstreamTokens(ctx, "expired")
 			require.Error(t, err)
 			assert.ErrorIs(t, err, ErrExpired)
-			assert.Nil(t, retrieved)
+			// Tokens should be returned alongside ErrExpired for refresh purposes
+			require.NotNil(t, retrieved)
+			assert.Equal(t, "expired-token", retrieved.AccessToken)
+			assert.Equal(t, "refresh-token", retrieved.RefreshToken)
 		})
 	})
```
jhrozek, line 645: Nit: the unit test here was properly updated for the new `ErrExpired`-returns-data contract, but the corresponding redis **integration** test in `redis_integration_test.go` (around line 667) still discards the return value with `_, err := s.GetUpstreamTokens(...)`. It should be updated to assert that the returned token data is non-nil, consistent with this change.

Without that, the integration test won't catch a regression back to `return nil, ErrExpired` on a real Redis instance.

pkg/authserver/refresher_test.go
```diff
@@ -0,0 +1,219 @@
...
+					Return(&upstream.Tokens{
+						AccessToken:  "new-access",
+						RefreshToken: "new-refresh",
+						IDToken:      "new-id-token",
+						ExpiresAt:    newExpiry,
+					}, nil)
+			},
+			setupStorage: func(s *storagemocks.MockUpstreamTokenStorage) {
+				s.EXPECT().StoreUpstreamTokens(gomock.Any(), "session-1", gomock.Any()).
+					DoAndReturn(func(_ context.Context, _ string, tokens *storage.UpstreamTokens) error {
```
jhrozek, line 63: Minor test correctness issue: the `DoAndReturn` closures here (and at line ~105) capture the **parent** `t` from `TestUpstreamTokenRefresher_RefreshAndStore`, not the subtest's `t`. Since subtests run with `t.Parallel()`, assertion failures inside these closures will be attributed to the parent test rather than the specific failing subtest.

To fix, the `setupStorage`/`setupProvider` function signatures could take the subtest `t` as a parameter, e.g.:

```go
setupStorage func(t *testing.T, s *storagemocks.MockUpstreamTokenStorage)
```

Then call `tt.setupStorage(t, mockStorage)` from within the subtest.

pkg/auth/upstreamswap/middleware.go
```diff
@@ -233,3 +229,71 @@ func createMiddlewareFunc(cfg *Config, storageGetter StorageGetter) types.Middle
...
+			if refreshed := tryRefreshUpstreamTokens(ctx, sessionID, tokens, refresherGetter); refreshed != nil {
+				return refreshed, nil
+			}
+		}
+		return nil, err
+	}
+
+	// Defense in depth: some storage implementations may return tokens
+	// without checking expiry (the interface does not require it).
+	if tokens.IsExpired(time.Now()) {
```
jhrozek: Bug: `tokens.IsExpired(time.Now())` returns `true` when `ExpiresAt` is zero (`time.Time{}`), because `time.Now().After(time.Time{})` is always true. This means tokens from providers that don't set `expires_in` (which is RECOMMENDED but not REQUIRED per RFC 6749 Section 5.1) will always be treated as expired here, triggering unnecessary refresh attempts.

Both storage implementations guard against this (`!result.ExpiresAt.IsZero()` in memory, `stored.ExpiresAt != 0` in Redis), so storage returns these tokens as valid — but then this defense-in-depth check immediately contradicts that. If the refresher is unavailable or fails, it returns 401 for a perfectly valid token.

```suggestion
	if !tokens.ExpiresAt.IsZero() && tokens.IsExpired(time.Now()) {
```

jhrozek (commented):
Second pass: a few more findings from an adversarial review.

jhrozek:
This is great work. I submitted a bunch of nits but take them really as nits, none of htem are really blocking, if there's one I'd love to have fixed before merging, it's the tokens.IsExpired, feel free to just dismiss the rest.

pkg/auth/upstreamswap/middleware.go
aron-muon: Nice call, added a `singleflight.Group` keyed on sessionID to deduplicate concurrent refreshes. Scoped per middleware instance to avoid test interference and also added `TestSingleFlightRefresh_ConcurrentRequests` to verify only one `RefreshAndStore` fires with 10 concurrent expired-session requests

pkg/auth/upstreamswap/middleware.go
aron-muon: Fixed!

pkg/authserver/storage/redis_test.go
aron-muon: Updated, the integration test now asserts the returned token data is non-nil and checks field values (consistent with unit tests)

pkg/authserver/refresher_test.go
aron-muon: Fixed, `setupStorage` and `setupProvider` now take `*testing.T` as a parameter so closures capture the subtest's `t`

aron-muon:
> This is great work. I submitted a bunch of nits but take them really as nits, none of htem are really blocking, if there's one I'd love to have fixed before merging, it's the tokens.IsExpired, feel free to just dismiss the rest.

Thanks for the review! I included a fix for each nit and also added a fix for the multiple concurrent requests issue.

pkg/auth/upstreamswap/middleware.go
```diff
@@ -233,3 +236,95 @@ func createMiddlewareFunc(cfg *Config, storageGetter StorageGetter) types.Middle
...
+// to deduplicate concurrent refresh attempts for the same session.
+func doSingleFlightRefresh(
+	ctx context.Context,
+	sfGroup *singleflight.Group,
+	sessionID string,
+	expired *storage.UpstreamTokens,
+	refresherGetter RefresherGetter,
+) *storage.UpstreamTokens {
+	result, err, _ := sfGroup.Do(sessionID, func() (any, error) {
+		refreshed := tryRefreshUpstreamTokens(ctx, sessionID, expired, refresherGetter)
```
jhrozek: The `ctx` captured here is the first caller's request context. If that client disconnects, `ctx` is cancelled, the upstream HTTP call aborts, and **all** singleflight waiters get a failed refresh → 401.

This can happen naturally (flaky connections, LB timeouts) or be triggered intentionally (fire-and-cancel requests to block legitimate refreshes).

Since ToolHive is on Go 1.26, you can use `context.WithoutCancel` paired with a server-side timeout to detach from the client's lifecycle while still bounding the operation:

```suggestion
	result, err, _ := sfGroup.Do(sessionID, func() (any, error) {
		// Detach from the triggering request's context so that if the first
		// caller disconnects, the refresh still completes for waiting callers.
		// The 30s timeout bounds the operation independently from client lifecycle.
		refreshCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
		defer cancel()
		refreshed := tryRefreshUpstreamTokens(refreshCtx, sessionID, expired, refresherGetter)
```

pkg/auth/upstreamswap/middleware_test.go
```diff
@@ -764,3 +769,361 @@ func TestMiddleware_TsidClaimWrongType(t *testing.T) {
...
+
+func (f *fakeTokenStorage) GetUpstreamTokens(_ context.Context, _ string) (*storage.UpstreamTokens, error) {
+	return f.tokens, f.err
+}
+
+func (*fakeTokenStorage) StoreUpstreamTokens(_ context.Context, _ string, _ *storage.UpstreamTokens) error {
+	return nil
+}
+
+func (*fakeTokenStorage) DeleteUpstreamTokens(_ context.Context, _ string) error {
```
jhrozek, line 1148: The 50ms sleep creates an overlap window, but the test could be flaky on single-core CI runners (`GOMAXPROCS=1`). If goroutine 1 completes `sfGroup.Do` before goroutine 2 enters it, the singleflight key is already cleaned up and each goroutine refreshes independently.

A more deterministic approach: use channels in the fake so the test controls when the refresh completes, guaranteeing all goroutines are inside `sfGroup.Do` before letting the first one finish:

```go
type fakeRefresher struct {
	result    *storage.UpstreamTokens
	callCount *atomic.Int32
	entered   chan struct{} // closed once inside RefreshAndStore
	proceed   chan struct{} // test closes this to let refresh complete
}

func (f *fakeRefresher) RefreshAndStore(_ context.Context, _ string, _ *storage.UpstreamTokens) (*storage.UpstreamTokens, error) {
	f.callCount.Add(1)
	select {
	case f.entered <- struct{}{}:
	default:
	}
	<-f.proceed
	return f.result, nil
}
```

Then in the test, after `close(ready)`, read from `entered` to confirm the refresh is in-flight, then `close(proceed)` to let it complete. This makes the test deterministic regardless of `GOMAXPROCS`.

jhrozek (commented):
Third pass: reviewing the singleflight addition. Nice implementation overall — two findings from the new code.

jhrozek:
two more nits and then we'll be good to go 🚀

pkg/auth/upstreamswap/middleware.go
aron-muon: Nice!!

pkg/auth/upstreamswap/middleware_test.go
aron-muon: Great point, replaced the `time.Sleep` overlap with the deterministic barrier: all goroutines sync at `GetUpstreamTokens` via a `sync.WaitGroup`, then release simultaneously through a gate channel. The refresher blocks on a `proceed` channel so the test controls exactly when it completes. No timing dependency, works regardless of GOMAXPROCS

aron-muon:
> two more nits and then we'll be good to go 🚀

Nice comments! Thank you!

jhrozek (approved):
let's ship this 🚀

jhrozek:
some of the tests are flaky, I don't think the CI failures are tied to your PR, I'll shepherd the PR through

Issue #2289, opened 2025-10-22
MCPRemoteProxy: Improve error reporting via Kubernetes Events instead of crash loops


JAORMX:
## Problem
When MCPRemoteProxy encounters configuration errors (e.g., HTTPS requirement for OIDC issuer), the pod enters a CrashLoopBackOff state. This makes it difficult for users to understand what went wrong without checking pod logs.

### Current Behavior
1. User creates MCPRemoteProxy with invalid configuration
2. Pod starts and immediately crashes
3. Status shows `Phase: Ready` briefly, then pod crashes
4. User must run `kubectl logs` to discover the actual error
5. Pod continues to restart in CrashLoopBackOff

**Example error (only visible in logs):**
```
Error: failed to create middleware of type auth: failed to create authentication middleware:
failed to discover OIDC configuration: failed to fetch OIDC configuration:
Get "http://keycloak.example.com/realms/toolhive/.well-known/openid-configuration":
the supplied URL http://keycloak.example.com/realms/toolhive/.well-known/openid-configuration is not HTTPS scheme
```

### Proposed Solution
**Option 1: Kubernetes Events (Recommended)**
The operator should emit Kubernetes Events when validation fails:
```bash
kubectl describe mcpremoteproxy my-proxy
# Shows:
# Events:
#   Type     Reason                   Message
#   ----     ------                   -------
#   Warning  OIDCConfigurationError   OIDC issuer must use HTTPS scheme. HTTP is not supported. Set insecureAllowHTTP: true for development only.
```

**Option 2: Status Conditions**
Update the MCPRemoteProxy status with detailed error information:
```yaml
status:
  phase: Failed
  conditions:
    - type: Ready
      status: "False"
      reason: OIDCConfigurationError
      message: "OIDC issuer URL must use HTTPS. Found: http://keycloak.example.com/realms/toolhive"
```

**Option 3: Both**
Combine Events for immediate feedback and Status Conditions for persistent state.

### Benefits
1. **Better UX**: Users get immediate feedback via `kubectl get` and `kubectl describe`
2. **Reduced noise**: No crash loops for configuration errors
3. **Actionable errors**: Clear messages about what's wrong and how to fix it
4. **K8s native**: Follows Kubernetes patterns for error reporting
5. **Audit trail**: Events provide a history of configuration issues

### Implementation Suggestions
**Validation in Operator:**
- Validate configuration before creating the deployment
- Emit Events for validation failures
- Set Status conditions appropriately
[...]

PR #4037, opened 2026-03-06
MCPRemoteProxy: Add remaining configuration validations

ChrisJBurns:
## Summary

Extends #4024 with the four remaining MCPRemoteProxy configuration validations, each surfacing errors via Kubernetes Events and Status Conditions:

- **Remote URL format validation** — rejects malformed URLs or unsupported schemes (`ftp://`, empty host)
- **JWKS URL scheme validation** — JWKS endpoints must use HTTPS (key material transport)
- **Cedar authorization policy syntax** — validates inline Cedar policies parse correctly before deployment
- **ConfigMap/Secret reference existence** — verifies referenced authz ConfigMaps and header Secrets exist in the namespace

All validations follow the foundation pattern from #4024: fail-fast in `validateSpec()`, set `ConfigurationValid=False` condition with a specific reason, emit a Warning event, and move the proxy to `Failed` phase.

No network calls are made — URL validations are format/scheme checks only; ConfigMap/Secret checks are in-cluster reads.

Closes https://github.com/stacklok/toolhive/issues/2289

## Test plan

- [x] Unit tests for `ValidateCedarPolicies` (6 cases)
- [x] Unit tests for `ValidateRemoteURL` (6 cases) and `ValidateJWKSURL` (5 cases)
- [x] Unit tests for reconciler condition-setting (5 new cases in `TestValidateSpecConfigurationConditions`)
- [x] Integration tests for status conditions (remote URL, JWKS URL, Cedar syntax, missing ConfigMap, missing Secret)
- [x] Integration tests for event emission (Cedar syntax, missing ConfigMap, missing Secret)
- [x] `go build ./cmd/thv-operator/...` passes
- [x] `golangci-lint` passes on changed packages

### DIFF
--- a/cmd/thv-operator/test-integration/mcp-remote-proxy/mcpremoteproxy_validation_integration_test.go
+++ b/cmd/thv-operator/test-integration/mcp-remote-proxy/mcpremoteproxy_validation_integration_test.go
@@ -84,22 +84,16 @@ var _ = Describe("MCPRemoteProxy Configuration Validation", Label("k8s", "remote
 	})
 
 	Context("Remote URL Format Validation", func() {
-		It("should set ConfigurationValid=False when remote URL has invalid scheme", func() {
-			By("creating an MCPRemoteProxy with ftp:// remote URL")
+		It("should reject creation when remote URL has invalid scheme via CRD validation", func() {
+			By("attempting to create an MCPRemoteProxy with ftp:// remote URL")
 			proxy := proxyHelper.NewRemoteProxyBuilder("test-bad-url").
 				WithRemoteURL("ftp://bad-scheme.example.com").
-				Create(proxyHelper)
+				Build()
 
-			By("waiting for the proxy to reach Failed phase")
-			statusHelper.WaitForPhase(proxy.Name, mcpv1alpha1.MCPRemoteProxyPhaseFailed, MediumTimeout)
-
-			By("verifying the ConfigurationValid condition")
-			statusHelper.WaitForConditionReason(
-				proxy.Name,
-				mcpv1alpha1.ConditionTypeConfigurationValid,
-				mcpv1alpha1.ConditionReasonRemoteURLInvalid,
-				MediumTimeout,
-			)
+			By("verifying the API server rejects the resource")
+			err := k8sClient.Create(testCtx, proxy)
+			Expect(err).To(HaveOccurred(), "expected CRD validation to reject ftp:// URL")
+			Expect(err.Error()).To(ContainSubstring("remoteURL"))
 		})
 	})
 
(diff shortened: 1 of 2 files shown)


cmd/thv-operator/controllers/mcpremoteproxy_controller.go
```diff
@@ -350,6 +350,13 @@ func (r *MCPRemoteProxyReconciler) validateSpec(ctx context.Context, proxy *mcpv
 		}
 	}
 
+	// Validate remote URL format
+	if err := validation.ValidateRemoteURL(proxy.Spec.RemoteURL); err != nil {
```
jhrozek, line 357: Two small things around this area:

1. The early empty-URL check above (line 339) doesn't set the `ConfigurationValid` condition or emit an event, unlike all the validation paths here. Could just drop that check and let `ValidateRemoteURL` handle the empty case — it already does the condition + event dance. Same gap exists for the external auth config block at 343-351.

2. This validate → `recordValidationEvent` → `setConfigurationInvalidCondition` → return sequence shows up 4 times. A small helper like `r.failValidation(proxy, reason, err) error` would clean it up and make it harder to forget a step (the external auth config block already misses both calls).

cmd/thv-operator/controllers/mcpremoteproxy_controller.go
```diff
@@ -409,6 +435,107 @@ func (*MCPRemoteProxyReconciler) validateOIDCIssuerURL(proxy *mcpv1alpha1.MCPRem
...
+					proxy,
+					mcpv1alpha1.ConditionReasonAuthzConfigMapNotFound,
+					msg,
+				)
+				setConfigurationInvalidCondition(
+					proxy,
+					mcpv1alpha1.ConditionReasonAuthzConfigMapNotFound,
+					msg,
+				)
+				return fmt.Errorf("%s", msg)
```
jhrozek: `fmt.Errorf("%s", msg)` → `errors.New(msg)` — no wrapping happening here so `fmt.Errorf` is unnecessary. Same on line 529.

cmd/thv-operator/controllers/mcpremoteproxy_controller.go
```diff
@@ -409,6 +435,107 @@ func (*MCPRemoteProxyReconciler) validateOIDCIssuerURL(proxy *mcpv1alpha1.MCPRem
...
+			}
+			return fmt.Errorf("failed to fetch authorization ConfigMap %q: %w", cmName, err)
```
jhrozek: If `r.Get` fails with something other than NotFound (RBAC error, timeout, etc.), the raw K8s API error gets wrapped and ends up in `proxy.Status.Message` via line 102. Those errors can contain internal stuff like API server URLs or RBAC details. Might want to log the full error and return something generic for status, the way the NotFound path already does with its sanitized `msg`.

ChrisJBurns:
@jhrozek  Give it another read on the latest commits, should have addressed your feedback

PR #4043, opened 2026-03-09
Fix authz response filtering for SSE transport

### LINKED ISSUE #4044: Authorization response filtering fails with Content-Length mismatch for SSE transport
## Description

When using authorization middleware (Cedar policies) with `streamable-http` transport, the response filtering for list operations (`tools/list`, `prompts/list`, `resources/list`) fails with errors.

## Environment

- Transport: `streamable-http`
- Authorization: Cedar policies configured via MCPRemoteProxy
- Client: Claude Desktop

## Error Messages

```
{"level":"ERROR","msg":"error writing filtered response","error":"http: wrote more than the declared Content-Length"}

{"level":"DEBUG","msg":"processSSEResponseToBuffer: failed to decode JSON-RPC message","error":"unmarshaling jsonrpc message: unexpected end of JSON input"}
```

## Root Cause

The `ResponseFilteringWriter` in `pkg/authz/response_filter.go` has two issues:

1. **Content-Length mismatch**: The original `Content-Length` header is forwarded, but the filtered response has a different size, causing HTTP layer errors.

2. **Chunked SSE data**: SSE responses arrive in multiple network chunks. The middleware attempts to parse incomplete JSON before the full SSE event is received.

## Expected Behavior

- Tools list should be filtered according to Cedar policies and displayed correctly in Claude Desktop
- No Content-Length mismatch errors

## Actual Behavior

- Empty tool list shown in Claude Desktop
- Errors logged in proxy pod

## Reproduction Steps

1. Deploy MCPRemoteProxy with Cedar authorization policies
2. Use `streamable-http` transport
3. Connect Claude Desktop to the proxy
4. Observe empty tool list and errors in logs

## Proposed Solution

- Delete `Content-Length` header before writing filtered responses
- Buffer SSE data until complete event (ending with `\n\n`) before processing
- Forward `WriteHeader` immediately for SSE streaming support

ChandraMohan0316:
## Summary

When using authorization (Cedar policies) with streamable-http transport, the response filtering middleware failed with two errors:
1. `http: wrote more than the declared Content-Length` - because filtered response size differs from original
2. `unexpected end of JSON input` - because SSE data arrives in chunks and was parsed before complete

This PR fixes the `ResponseFilteringWriter` to properly handle SSE streaming responses.

**Changes:**
- Delete `Content-Length` header before writing filtered responses to prevent size mismatch
- Buffer SSE data until complete event (ending with `\n\n`) before processing
- Forward `WriteHeader` immediately for SSE streaming support
- Handle chunked SSE data arriving in multiple `Flush()` calls

Fixes #4044

## Type of change

- [x] Bug fix

## Test plan

**Manual testing performed:**
- Deployed MCPRemoteProxy with Cedar authorization policies using streamable-http transport
- Verified tools list now appears correctly in Claude Desktop
- Confirmed no Content-Length mismatch errors in logs

Yes. Users with authorization policies configured on MCPRemoteProxy using streamable-http transport will now see their MCP server tools correctly in Claude Desktop, instead of an empty tool list.

The fix follows the same pattern used in `pkg/mcp/tool_filter.go` for handling SSE responses. Key changes:
- Added `headerWritten` field to track header state and prevent duplicate writes
- `Flush()` now checks for complete SSE events before processing
- `writeHeaderOnce()` helper ensures Content-Length is deleted before headers are sent


jhrozek:
Hey @Chandra-mohan31 thanks a lot of the contribution. After you posted your issue on Discord, I checked the code and came to the same conclusion about content-length but I wonder if the changes can be a bit simpler:

https://github.com/stacklok/toolhive/commit/c7c69183b98c2bd5ed6dd673e52fbcd6dab46fdf

the reason I'd prefer a simpler approach is that with the approach in this PR content-length is deleted in all paths, which forces chunked transfer but most importantly the simpler approach doesn't force you to write all the complex logic to Flush and avoids the parallel logic in Flush.

Would you be able to test that patch out, too?

jhrozek:
It seems like my original patch missed a spot in Flush(), can you try https://github.com/stacklok/toolhive/commit/eab265d5dc33da8fd5f6a9abeac47093bd8e7d05

@Chandra-mohan31 ?

ChandraMohan0316:
Hi @jhrozek ,Thank you, I tested out the code, It is working as expected. The tools are properly filtered out based on the rules defined and the AI assistant is able to see the capabilities.

jhrozek:
> Hi @jhrozek ,Thank you, I tested out the code, It is working as expected. The tools are properly filtered out based on the rules defined and the AI assistant is able to see the capabilities.

Yes! Great, thank you for testing!

 I will open a draft PR with it and let you see if you want to adjust your to match the simpler behaviour (so that you still get the credit you've earned) or if you prefer we use the second PR (and I'll be happy to add you as a co-author).

I can add inline comments to your PR to better explain why I prefer a simpler approach, but the short version is that this is security-critical and a little delicate code, so fewer changes usually means fewer chances of a regression in some unrelated paths. 

It doesn't mean that we don't value the contribution, on the contrary, as I said above I'll be happy to either let you redo the first approach or be included as a co-author. You've earned the credit :-)

ChandraMohan0316:
Thanks @jhrozek , I prefer you use the second PR and have me as the co-author.

jhrozek:
Superseded by #4092

PR #4042, opened 2026-03-09
Add git reference support for skill install

JAORMX:
## Summary

- Plain skill names (`thv skill install my-skill`) created "pending" records that never resolved — a dead-end UX problem. This PR adds git repository support as an installation source and replaces the pending path with an actionable 404 error.
- Users can now install skills directly from git repos: `thv skill install git://github.com/org/repo#path/to/skill`
- The operator's internal git package (`cmd/thv-operator/pkg/git/`) is moved to `pkg/git/` as a shared package, replacing `controller-runtime/pkg/log` with `log/slog` for CLI compatibility.

Relates to #4015

## Type of change

- [x] New feature

## Large PR Justification

This PR exceeds 1000 lines because ~550 lines are a **code move** (`cmd/thv-operator/pkg/git/` → `pkg/git/`) — the git client, types, LimitedFs, and tests are relocated with minor modifications (slog instead of controller-runtime logger). The net new production code is ~300 lines across the gitresolver package and skillsvc integration. Splitting the move into a separate PR would create a transient state where `pkg/git/` exists but nothing uses it, and the feature PR would depend on it — adding review overhead without reducing risk. The move and the feature that motivates it are logically atomic.

## Test plan

- [x] Unit tests (`task test`)
- [x] Linting (`task lint-fix`)

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/pkg/git/*` | **Deleted** — moved to `pkg/git/` |
| `pkg/git/*` | **New** — shared git client (from operator), adds `HeadCommitHash()`, `WithAuth` option, replaces controller-runtime logger with `log/slog` |
| `pkg/skills/gitresolver/reference.go` | **New** — `git://` URL parsing with SSRF prevention (private IP/localhost rejection, absolute path/backslash rejection) |
| `pkg/skills/gitresolver/resolver.go` | **New** — `Resolver` interface: clones repo, extracts skill files, enforces 2-min clone timeout |
| `pkg/skills/gitresolver/auth.go` | **New** — host-scoped auth resolution (`GITHUB_TOKEN` only sent to github.com, etc.) |
| `pkg/skills/gitresolver/writer.go` | **New** — writes files to disk with symlink checks, containment validation, permission cap (0644) |
| `pkg/skills/skillsvc/git_install.go` | **New** — `installFromGit`, `resolveGitReference`, `upsertGitSkill` methods with supply chain defense (SKILL.md name must match reference) |
[...]

Issue #4015, opened 2026-03-05
Skill install needs reference resolution (matching thv run pattern)


JAORMX:
## Problem

`thv skill install <name>` with a plain name creates a "pending" record that never resolves to "installed". The only working install path today is with a fully-qualified OCI reference:

```bash
# This works
thv skill install ghcr.io/org/my-skill:v1

# These don't — skill stays "pending" forever
thv skill build ./my-skill && thv skill install my-skill
thv skill install some-published-skill
```

This breaks both the local development workflow (build → install) and the registry-based discovery workflow (install by name).

## How `thv run` solves this for MCP servers

`thv run` has a cascading reference resolution model that skills should follow:

| Reference type | `thv run` example | Resolution |
|---|---|---|
| **Registry name** | `thv run filesystem` | Looks up in registry → gets OCI image ref → pulls |
| **Direct OCI ref** | `thv run ghcr.io/org/server:v1` | Pulls directly |
| **Protocol scheme** | `thv run uvx://mcp-server-git` | Builds from source template |
| **Remote URL** | `thv run https://api.example.com/mcp` | Proxies to remote |

The resolution cascade in `pkg/runner/retriever/retriever.go`:
1. Protocol scheme check (`uvx://`, `npx://`, `go://`)
2. Registry lookup (`provider.GetServer(name)`)
3. Fallthrough to direct image reference

## What `thv skill install` needs

Skills should support analogous reference types:

| Reference type | Example | Resolution |
|---|---|---|
| **Registry/index name** | `thv skill install my-skill` | Look up in skill index/registry → get OCI ref → pull & extract |
| **Direct OCI ref** | `thv skill install ghcr.io/org/my-skill:v1` | Pull & extract (already works) |
| **Local build tag** | `thv skill install my-skill` (after `thv skill build`) | Resolve from local OCI store → extract |
| **Git reference** | `thv skill install git://github.com/org/repo#path/to/skill` | Clone & extract (future) |

## Root cause

The `Install` method in `pkg/skills/skillsvc/skillsvc.go` has three paths:

1. **OCI reference** (contains `/`, `:`, `@`) → `installFromOCI()` → pulls from remote → extracts → **works**
2. **Plain name + LayerData** → `installWithExtraction()` → extracts → **works** (but unreachable from CLI/API)
3. **Plain name, no LayerData** → `installPending()` → creates DB record with `status: "pending"` → **dead end**

Path 3 is always hit for plain names because:
[...]

PR #4045, opened 2026-03-09
Move shared git package from operator to pkg/git

JAORMX:
## Summary

- The operator's internal git client (`cmd/thv-operator/pkg/git/`) is needed by the CLI for upcoming git-based skill installation. Moving it to `pkg/git/` makes it a shared package available to both the operator and CLI.
- Replaces `controller-runtime/pkg/log` with `log/slog` so the package works outside the operator's controller-runtime context.
- Rewrites tests to use testify assertions and `t.TempDir()` for isolation, dropping trivial field-accessibility tests and slow network-dependent tests in favor of deterministic local-repo integration tests.
- Adds path traversal validation, CloneConfig mutual-exclusion validation, and sentinel errors for improved security and error handling.

Relates to #4015

## Type of change

## Test plan

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/pkg/git/*` | **Deleted** — moved to `pkg/git/` |
| `pkg/git/client.go` | **Moved** from operator, swapped `controller-runtime/pkg/log` for `log/slog`, added path validation, sentinel errors, cleanup logging |
| `pkg/git/fs.go` | **Moved** from operator, switched sentinel errors to `errors.New` |
| `pkg/git/types.go` | **Moved** from operator, added `validate()` for CloneConfig mutual exclusion |
| `pkg/git/doc.go` | **Updated** package documentation |
| `pkg/git/client_test.go` | **Rewritten** with testify, `t.TempDir()`, added path traversal and validation tests |
| `pkg/git/integration_test.go` | **Rewritten** with testify, local-repo tests only, added tag-based clone test |

No

## Large PR Justification

- Large refactoring that must be atomic: the package move deletes 991 lines from `cmd/thv-operator/pkg/git/` and recreates 331 lines in `pkg/git/`. The deletions account for the majority of the diff size. Splitting the delete and create into separate PRs would leave the codebase in a broken state.
- Test rewrite is part of the move: the old tests depended on `controller-runtime/pkg/log` which is no longer available in the new location, so they had to be rewritten in the same PR.

### Operator test files intentionally dropped during move

- **`commit_test.go`**: Cloned from real `github.com/stacklok/toolhive` over the network — slow, fragile, skipped in `-short` mode. Same scenario covered by local-repo integration tests.
- **`types_test.go`**: Trivial field-accessibility tests ("set URL, check URL is set") that add no meaningful coverage per the project's test quality guidelines.

### Security hardening added during move

- Path traversal validation in `GetFileContent` (rejects `..`, absolute paths, null bytes)
- `CloneConfig.validate()` enforces mutual exclusion of Branch/Tag/Commit
- Sentinel errors (`ErrNilRepository`, `ErrInvalidFilePath`, `ErrInvalidCloneConfig`) for proper `errors.Is()` checking

This is PR 1 of 2 — split from #4042. PR 2 (git-based skill install feature) builds on this.

### DIFF
--- a/pkg/git/client.go
+++ b/pkg/git/client.go
@@ -5,8 +5,11 @@ package git
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
+	"path/filepath"
+	"strings"
 
 	"github.com/go-git/go-billy/v5/memfs"
 	"github.com/go-git/go-billy/v5/util"
@@ -16,6 +19,15 @@ import (
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/go-git/go-git/v5/storage/filesystem"
 )
+
+// ErrNilRepository is returned when a nil repository is passed to an operation that requires one.
+var ErrNilRepository = errors.New("repository is nil")
+
+// ErrInvalidFilePath is returned when a file path contains traversal or absolute components.
+var ErrInvalidFilePath = errors.New("invalid file path")
+
+// ErrInvalidCloneConfig is returned when CloneConfig has conflicting ref specifications.
+var ErrInvalidCloneConfig = errors.New("invalid clone config: at most one of Branch, Tag, or Commit may be specified")
 
 // Client defines the interface for Git operations
 type Client interface {
@@ -56,6 +68,10 @@ func NewDefaultGitClient(opts ...ClientOption) *DefaultGitClient {
 
 // Clone clones a repository with the given configuration
 func (c *DefaultGitClient) Clone(ctx context.Context, config *CloneConfig) (*RepositoryInfo, error) {
+	if err := config.validate(); err != nil {
+		return nil, err
+	}
+
 	// Prepare clone options
 	cloneOptions := &git.CloneOptions{
 		URL:  config.URL,
@@ -135,7 +151,12 @@ func (c *DefaultGitClient) Clone(ctx context.Context, config *CloneConfig) (*Rep
 // GetFileContent retrieves the content of a file from the repository
 func (*DefaultGitClient) GetFileContent(repoInfo *RepositoryInfo, path string) ([]byte, error) {
 	if repoInfo == nil || repoInfo.Repository == nil {
-		return nil, fmt.Errorf("repository is nil")
+		return nil, ErrNilRepository
+	}
+
+	// Reject absolute paths, traversal, and null bytes
+	if filepath.IsAbs(path) || strings.Contains(path, "..") || strings.ContainsRune(path, 0) {
+		return nil, fmt.Errorf("%w: %s", ErrInvalidFilePath, path)
 	}
 
 	// Get the HEAD reference
@@ -174,7 +195,7 @@ func (*DefaultGitClient) GetFileContent(repoInfo *RepositoryInfo, path string) (
 // Cleanup removes local repository directory
 func (*DefaultGitClient) Cleanup(_ context.Context, repoInfo *RepositoryInfo) error {
 	if repoInfo == nil || repoInfo.Repository == nil {
-		return fmt.Errorf("repository is nil")
+		return ErrNilRepository
 	}
 
 	// 1. Clear object cache explicitly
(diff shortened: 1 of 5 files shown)


pkg/git/client.go
```diff
@@ -154,28 +172,27 @@ func (*DefaultGitClient) GetFileContent(repoInfo *RepositoryInfo, path string) (
...
+		slog.Debug("Clearing object cache")
 		repoInfo.objectCache.Clear()
 	}
 
 	// 2. Clear worktree filesystem
 	worktree, err := repoInfo.Repository.Worktree()
 	if err == nil && worktree.Filesystem != nil {
-		logger.V(1).Info("Clearing worktree filesystem")
+		slog.Debug("Clearing worktree filesystem")
 		_ = util.RemoveAll(worktree.Filesystem, "/")
```
aponcedeleonch: Nit: Both `util.RemoveAll` errors (here and line 196) are silently discarded, yet `Cleanup` returns `error`. A partial cleanup will return `nil`, hiding the fact that resources may not have been fully released.

Consider at least logging the errors, or collecting them with `errors.Join` and returning:

```go
if err := util.RemoveAll(worktree.Filesystem, "/"); err != nil {
    slog.Warn("Failed to clear worktree filesystem", "error", err)
}
```

pkg/git/client.go
```diff
@@ -29,19 +29,37 @@ type Client interface {
...
 func (c *DefaultGitClient) Clone(ctx context.Context, config *CloneConfig) (*RepositoryInfo, error) {
-	// Prepare clone options (no authentication for initial version)
+	// Prepare clone options
 	cloneOptions := &git.CloneOptions{
-		URL: config.URL,
+		URL:  config.URL,
+		Auth: c.auth,
 	}
 
 	// Set reference if specified (but not for commit-based clones)
```
aponcedeleonch, line 81: Nit: `CloneConfig` allows setting `Branch`, `Tag`, and `Commit` simultaneously, with precedence resolved implicitly here (`Commit` wins, then `Branch` before `Tag`). This makes it easy for callers to accidentally set conflicting fields.

Consider either:
1. Adding a `Validate()` method on `CloneConfig` that rejects conflicting combinations, or
2. Adding a short comment on the struct in `types.go` documenting the precedence, e.g.: `// At most one of Branch, Tag, or Commit should be set. If Commit is set, Branch and Tag are ignored.`

pkg/git/integration_test.go
```diff
@@ -0,0 +1,190 @@
...
+	require.NoError(t, err)
+
+	// Verify we're on the feature branch and have the feature file
+	content, err := client.GetFileContent(repoInfo, "feature.txt")
+	require.NoError(t, err)
+	assert.Equal(t, "feature branch content", string(content))
+
+	// Clean up
+	require.NoError(t, client.Cleanup(t.Context(), repoInfo))
+}
```
aponcedeleonch, line 119: Nit: There are integration tests for branch-based and commit-based cloning, but no test for **tag-based cloning** (`CloneConfig{Tag: "v1.0.0"}`). Since tag support via `NewTagReferenceName` is new code in this PR, it would be good to add a `TestDefaultGitClient_CloneWithTag` following the same pattern as this test.

pkg/git/client.go
JAORMX: Done — switched from silently discarding to logging with `slog.Warn` on failure. See cba20b1a.

pkg/git/client.go
JAORMX: Done — added `CloneConfig.validate()` that rejects conflicting combinations (returns `ErrInvalidCloneConfig` if more than one of Branch/Tag/Commit is set). Called at the start of `Clone`. See cba20b1a.

pkg/git/integration_test.go
JAORMX: Done — added `TestDefaultGitClient_CloneWithTag` that creates a tag, adds a post-tag commit, then verifies cloning at the tag only sees the tagged tree. See cba20b1a.

PR #4047, opened 2026-03-09
Add DoS protection and idle timeout to vMCP session manager

### LINKED ISSUE #3874: [vMCP] Implement resource exhaustion and DoS protection for session creation
Depends on #3866

Session-scoped clients multiply backend connections (N sessions × M backends). Without limits, a burst of session creation requests can exhaust file descriptors or memory. Implement configurable limits to protect the system.

**Global session limit**:
- Configurable via `TOOLHIVE_MAX_SESSIONS` (or equivalent config key)
- When reached: reject new `InitializeRequest` with HTTP 503 and a `Retry-After` header
- Error body: `{"error": {"code": -32000, "message": "Maximum concurrent sessions exceeded. Please try again later or contact administrator."}}`
- No request queueing — reject immediately; `Retry-After` gives clients the retry signal
- Do not expose current session count in the error response

**Per-client session limit**:
- Configurable maximum sessions per client identity/IP
- Enforced independently of the global limit

**Idle session timeout**:
- Sessions inactive for a configurable period (e.g. 5 minutes with no incoming requests) should be expired proactively, independently of the global session TTL
- This reclaims backend connections faster than waiting for the full TTL to elapse, reducing the steady-state connection count under low-traffic conditions
- Idle timeout must reset on every request touching the session (i.e. any `CallTool`, `ReadResource`, or `GetPrompt` resets the idle clock)
- Configurable separately from the session TTL; idle timeout must be shorter than or equal to TTL

### Acceptance Criteria

- [ ] A configurable global session limit exists and defaults to a documented value
- [ ] Requests exceeding the limit receive HTTP 503 with a `Retry-After` header
- [ ] The error response body does not include the current session count or any internal metrics
- [ ] A configurable per-client session limit is enforced independently of the global limit
- [ ] A configurable idle timeout exists and defaults to a documented value shorter than the session TTL
- [ ] Sessions with no activity for longer than the idle timeout are expired and their backend clients closed
- [ ] Any request to a session resets the idle clock
- [ ] Idle timeout is enforced independently of — and cannot exceed — the session TTL
- [ ] Requests within the limit are unaffected
- [ ] Tests cover: 503 + `Retry-After` on global limit hit, per-client limit enforced, idle session expired after inactivity, active session not expired while in use, error response contains no internal data

---

yrobla:
## Summary

Implements resource-exhaustion protections for the session-scoped backend lifecycle (SessionManagementV2), resolving issue #3874.

Global session limit: new session requests (no Mcp-Session-Id header) receive HTTP 503 with a Retry-After header when the server-wide cap is reached. Default: 100 sessions.

Per-client session limit: CreateSession enforces a maximum number of concurrent sessions per auth.Identity.Subject. Anonymous clients are exempt. The counter is rolled back on all failure paths. Default: 10 sessions per identity.

Idle session timeout: a background reaper goroutine terminates sessions that have had no CallTool activity for longer than the configured threshold. The idle clock resets on every tool call and is initialised when a session is fully established. The reaper is wired into shutdownFuncs so it stops cleanly on server shutdown. Default: 5 min.

All three limits are configurable via Config fields; zero values select the defaults. The Limits struct is passed to sessionmanager.New() and all existing call sites are updated.

Fixes #3874 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

No

### DIFF
--- a/pkg/vmcp/server/server.go
+++ b/pkg/vmcp/server/server.go
@@ -16,6 +16,7 @@ import (
 	"log/slog"
 	"net"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -66,6 +67,13 @@ const (
 	// defaultSessionTTL is the default session time-to-live duration.
 	// Sessions that are inactive for this duration will be automatically cleaned up.
 	defaultSessionTTL = 30 * time.Minute
+
+	// defaultIdleCheckInterval is how often the idle reaper scans for inactive sessions.
+	defaultIdleCheckInterval = time.Minute
+
+	// defaultRetryAfterSeconds is the Retry-After value returned with HTTP 503
+	// when the global session limit is reached.
+	defaultRetryAfterSeconds = 30
 )
 
 //go:generate mockgen -destination=mocks/mock_watcher.go -package=mocks -source=server.go Watcher
@@ -161,6 +169,21 @@ type Config struct {
 	// SessionFactory creates MultiSessions for session management.
 	// Required; must not be nil.
 	SessionFactory vmcpsession.MultiSessionFactory
+
+	// MaxSessions is the global concurrent session limit when SessionManagementV2 is enabled.
+	// Requests that would exceed this limit receive HTTP 503 with a Retry-After header.
+	// 0 uses the default (100). Requires SessionManagementV2 = true.
+	MaxSessions int
+
+	// MaxSessionsPerClient is the per-identity session limit when SessionManagementV2 is enabled.
+	// Keyed by auth.Identity.Subject; anonymous clients are not limited.
+	// 0 uses the default (10). Requires SessionManagementV2 = true.
+	MaxSessionsPerClient int
+
+	// IdleSessionTimeout is the duration after which inactive sessions are proactively
+	// expired when SessionManagementV2 is enabled. Must be ≤ SessionTTL.
+	// 0 uses the default (5 minutes). Requires SessionManagementV2 = true.
+	IdleSessionTimeout time.Duration
 }
 
 // Server is the Virtual MCP Server that aggregates multiple backends.
@@ -274,6 +297,24 @@ func New(
 	}
 	if cfg.SessionTTL == 0 {
 		cfg.SessionTTL = defaultSessionTTL
+	}
+	if cfg.MaxSessions == 0 {
+		cfg.MaxSessions = sessionmanager.DefaultMaxSessions
+	}
+	if cfg.MaxSessionsPerClient == 0 {
+		cfg.MaxSessionsPerClient = sessionmanager.DefaultMaxSessionsPerClient
+	}
+	if cfg.IdleSessionTimeout == 0 {
+		cfg.IdleSessionTimeout = sessionmanager.DefaultIdleSessionTimeout
+	}
+	// IdleSessionTimeout must not exceed SessionTTL: if it did, the transport
+	// TTL reaper could evict sessions before the idle reaper fires, leaving
+	// per-client counters and idle-tracking maps stale.
+	if cfg.IdleSessionTimeout > cfg.SessionTTL {
+		slog.Warn("IdleSessionTimeout exceeds SessionTTL; clamping to SessionTTL",
+			"idle_session_timeout", cfg.IdleSessionTimeout,
+			"session_ttl", cfg.SessionTTL)
+		cfg.IdleSessionTimeout = cfg.SessionTTL
 	}
 
 	// Create hooks for SDK integration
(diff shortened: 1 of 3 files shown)


pkg/vmcp/server/server.go
```diff
@@ -575,6 +628,37 @@ func (s *Server) Handler(_ context.Context) (http.Handler, error) {
...
+				_, _ = w.Write([]byte(
+					`{"error":{"code":-32000,"message":"Maximum concurrent sessions exceeded. ` +
+						`Please try again later or contact administrator."}}`,
+				))
+				return
+			}
+		}
+		next.ServeHTTP(w, r)
+	})
+}
```
jerm-dro, line 651: Blocker: can we remove the rate limiting behavior from this PR?

It's not really needed today, so I'd prefer to avoid adding it.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -68,19 +110,40 @@ type Manager struct {
...
+
+	// idleActivityMu guards idleActivity.
+	idleActivityMu sync.RWMutex
+	idleActivity   map[string]time.Time // sessionID → last active time
+
+	// activeSessionCount tracks sessions that have been generated but not yet
+	// terminated, excluding terminated placeholders left for TTL cleanup.
+	// This gives an accurate count for global limit enforcement, unlike
+	// storage.Count() which includes those terminated placeholders.
+	activeSessionCount atomic.Int64
```
jerm-dro, line 128: Blocker: this is adding a lot of complexity to `Manager`. We should refactor the `Manager` struct first so we don't have to mix all these concerns together in one class. That would also make unit testing much simpler. I would expect a change like this doesn't require the integration_tests at all.

jerm-dro (changes requested):
I don't think these changes are required for the session v2 work. Would you be okay setting this work aside for now? 

I think we'll tackle a related problem as part of the upcoming scale work, where we'll need to implement an LRU for the session storage so not all sessions are held in memory.

yrobla:
ok setting as draft for now. I know that is not essential but i was speeding up with pending tasks while the main code was reviewed...

amirejaz:
@yrobla  could you take a look at this draft PR and confirm if it’s still relevant?

yrobla:
> @yrobla could you take a look at this draft PR and confirm if it’s still relevant?

well, we agreed to implement it at some point, but it was not critical. If you don't mind, i prefer to keep them as draft, to resume the work when needed.

JAORMX:
So, I went through this pretty thoroughly and I have some thoughts. First off, the intent here is good... DoS protection for session management is something we'll want eventually. But there are some issues that I think need addressing before this can move forward.

I agree with Jeremy's two blockers, and I found a few more things on top of those.

**The Manager is already doing too much**

Jeremy flagged this and he's right. The `Manager` struct today has 4 fields and ~15 methods covering session lifecycle, caching, SDK adaptation, backend notification, and session decoration. That's already borderline. This PR would add three separate synchronization primitives (`atomic.Int64`, `sync.Mutex`, `sync.RWMutex`) each protecting different state subsets, plus a background goroutine. That's our own anti-pattern #3 in `.claude/rules/vmcp-anti-patterns.md` staring right at us.

The decomposition path is actually pretty clear:
- Session admission (global + per-client limits) should be its own type with its own lock and tests
- Idle reaping... we already have a TODO in `RestorableCache` (cache.go line 56) for exactly this. The cache already has `onEvict` callbacks. That's the natural home for it, not a new goroutine on Manager

**Counter-drift bug in Generate()**

OK this one is subtle but real. When storage fails in `Generate()`, the `activeSessionCount` decrement is conditional:

```go
if sm.limits.MaxSessions > 0 {
    sm.activeSessionCount.Add(-1)
}
```

If you incremented the counter, you need to decrement it on failure. Period. Making the rollback conditional on a config value that could theoretically change between the increment and the error path means the counter can drift permanently. Once it drifts, the global limit rejects new sessions forever until restart. The fix is simple... make the decrement unconditional.

Also, the limited path increments *before* storage (reservation pattern) while the unlimited path increments *after*. That asymmetry is confusing and easy to break.

**Per-client map entries stick around at zero**

`decrementPerClientCount` decrements the count but never deletes the map entry when it hits 0. So every unique subject that ever connected leaves a permanent zero-valued entry in `perClientCounts`. Not a huge deal, but it's unbounded growth with no eviction. Easy fix.

**The middleware splits admission logic across two layers**

The `sessionLimitMiddleware` does an optimistic (non-atomic) check, then `Generate()` does the real atomic enforcement. This means:
- If the middleware rejects, fine, early exit
- If the middleware passes but `Generate()` rejects, `Generate()` returns `""` and the SDK gets a cryptic failure

Pick one place for the check. Either the middleware owns it entirely (and returns a proper 503), or `Generate()` owns it entirely. Splitting it is anti-pattern #4 from our own rules.

**Missing test coverage**

The tests that are here use good patterns (timestamp back-dating instead of `time.Sleep`, nice). But there's no:
- Concurrent admission test (N goroutines racing against MaxSessions=M... that's the whole point of the atomic reservation pattern and it's untested)
- `sessionLimitMiddleware` test (HTTP 503, Retry-After header, existing sessions pass through)
- `IdleSessionTimeout > SessionTTL` clamping test

Codecov shows 68.7% patch coverage with 46 lines missing.

**Small thing: wrong change type**

The PR checks "Refactoring (no behavior change)" but adding session limits and idle timeout is a new feature. Not a big deal but worth fixing.

**Bottom line**

I think we should close this one out and reopen it as smaller, focused PRs once we've done the Manager decomposition. Something like:

1. PR to extract session admission into its own type
2. PR to implement the `RestorableCache` idle sweep (the TODO is already there waiting for us)
3. PR to wire the limits into the server config

That way each piece is testable and reviewable on its own. Thoughts?

PR #4051, opened 2026-03-09
Add gitresolver package for git:// skill references

JAORMX:
## Summary

- The upcoming git-based skill install feature (`thv skill install git://...`) needs a package to parse git references, clone repos, resolve auth, and securely write extracted files. This PR adds that foundation as a self-contained package without wiring it into the skill service yet.
- New `pkg/skills/gitresolver/` package with four components: URL parsing with SSRF prevention, clone resolution with timeout, host-scoped credential resolution, and secure file writing with symlink/containment checks.
- Adds `HeadCommitHash` helper to `pkg/git/` for retrieving the HEAD commit hash after clone (used by the resolver to record provenance).

Relates to #4015

## Type of change

- [x] New feature

## Large PR Justification

This PR is 1268 lines but ~655 lines (over half) are unit tests. The remaining ~613 lines are a single self-contained package (`pkg/skills/gitresolver/`) with four tightly coupled components (URL parsing, clone resolution, auth, file writing) that share types and validation logic. Splitting these into separate PRs would create packages that don't compile independently and add review overhead without reducing risk. The 14-line `HeadCommitHash` addition to `pkg/git/` is the only change outside the new package.

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/skills/gitresolver/reference.go` | **New** — `git://` URL parsing with SSRF prevention (private IP/localhost rejection, absolute path/backslash rejection) |
| `pkg/skills/gitresolver/resolver.go` | **New** — `Resolver` interface: clones repo, extracts skill files, enforces 2-min clone timeout |
| `pkg/skills/gitresolver/auth.go` | **New** — host-scoped auth resolution (`GITHUB_TOKEN` only sent to github.com, etc.) |
| `pkg/skills/gitresolver/writer.go` | **New** — writes files to disk with symlink checks, containment validation, permission cap (0644) |
| `pkg/git/client.go` | **Modified** — added `HeadCommitHash` helper function |
| `*_test.go` | Comprehensive tests for all four components |

No — this is a foundation package with no integration yet. The next PR will wire it into skillsvc.

### Security design

- **Credential exfiltration prevented**: Tokens are host-scoped per-clone — `GITHUB_TOKEN` is only sent to `github.com`, `GITLAB_TOKEN` only to `gitlab.com`. `GIT_TOKEN` is an opt-in fallback for other hosts.
- **Clone timeout**: 2-minute `context.WithTimeout` in `Resolve()` prevents slowloris-style hangs.
- **Path validation hardened**: `validateSkillPath` rejects absolute paths and backslashes in addition to `..` traversal.
- **SSRF**: Static host validation rejects localhost/private IPs.
- **Resource limits**: `LimitedFs` caps clones at 10k files / 100MB.

### Known limitations for follow-up

- DNS rebinding not mitigated (needs custom `http.Transport` with `DialContext` hook)
- No host allowlist (denylist only)
- `collectFiles` reads only top-level files in the skill directory (flat structure)

This is PR 2 of 3 — split from #4042. PR 1 (#4045, merged) moved the git package. PR 3 will integrate gitresolver into skillsvc.

### DIFF
--- a/pkg/skills/gitresolver/writer.go
+++ b/pkg/skills/gitresolver/writer.go
@@ -7,8 +7,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 
+	"github.com/stacklok/toolhive/pkg/fileutils"
 	"github.com/stacklok/toolhive/pkg/skills"
 )
 
@@ -22,49 +22,34 @@ func WriteFiles(files []FileEntry, targetDir string, force bool) error {
 	// Sanitize targetDir early so all downstream os calls use the clean path.
 	targetDir = filepath.Clean(targetDir)
 
-	// Handle existing directory
-	if _, statErr := os.Stat(targetDir); statErr == nil { // lgtm[go/path-injection] #nosec G304
-		if !force {
-			return fmt.Errorf("target directory %q already exists; use force to overwrite", targetDir)
-		}
-		if err := os.RemoveAll(targetDir); err != nil { // lgtm[go/path-injection] #nosec G304 -- targetDir is cleaned above
-			return fmt.Errorf("removing existing directory: %w", err)
-		}
-	}
-
-	// Pre-extraction: validate that no existing path components are symlinks.
-	// Reuses the same check as the OCI installer (pkg/skills/installer.go).
-	if err := skills.ValidatePathNoSymlinks(targetDir); err != nil {
-		return fmt.Errorf("target path validation: %w", err)
-	}
-
-	if err := os.MkdirAll(targetDir, skills.DirPermissions); err != nil { // lgtm[go/path-injection] #nosec G304
-		return fmt.Errorf("creating target directory: %w", err)
-	}
-
-	// Containment and write logic mirrors pkg/skills/installer.go:writeFiles.
-	cleanTarget := targetDir + string(os.PathSeparator)
-
-	for _, f := range files {
-		destPath := filepath.Clean(filepath.Join(targetDir, filepath.FromSlash(f.Path)))
-
-		// Containment check: ensure destPath is beneath targetDir.
-		if !strings.HasPrefix(destPath, cleanTarget) {
-			return fmt.Errorf("path traversal detected: file %q escapes target directory", f.Path)
+	return fileutils.WithFileLock(targetDir, func() error {
+		// Handle existing directory
+		if _, statErr := os.Stat(targetDir); statErr == nil { // lgtm[go/path-injection] #nosec G304
+			if !force {
+				return fmt.Errorf("target directory %q already exists; use force to overwrite", targetDir)
+			}
+			if err := os.RemoveAll(targetDir); err != nil { // lgtm[go/path-injection] #nosec G304 -- targetDir is cleaned above
+				return fmt.Errorf("removing existing directory: %w", err)
+			}
 		}
 
-		parentDir := filepath.Dir(destPath)
-		if err := os.MkdirAll(parentDir, skills.DirPermissions); err != nil {
-			return fmt.Errorf("creating directory %q: %w", parentDir, err)
+		// Pre-extraction: validate that no existing path components are symlinks.
+		// Reuses the same check as the OCI installer (pkg/skills/installer.go).
+		if err := skills.ValidatePathNoSymlinks(targetDir); err != nil {
+			return fmt.Errorf("target path validation: %w", err)
 		}
 
-		// Sanitize file permissions: strip setuid/setgid/sticky, cap at 0644
-		mode := (f.Mode & 0o777) & skills.FilePermissionMask
-
-		if err := os.WriteFile(destPath, f.Content, mode); err != nil {
-			return fmt.Errorf("writing file %q: %w", f.Path, err)
+		if err := os.MkdirAll(targetDir, skills.DirPermissions); err != nil { // lgtm[go/path-injection] #nosec G304
+			return fmt.Errorf("creating target directory: %w", err)
 		}
-	}
 
-	return nil
+		for _, f := range files {
+			mode := (f.Mode & 0o777) & skills.FilePermissionMask
+			if err := fileutils.WriteContainedFile(targetDir, f.Path, f.Content, skills.DirPermissions, mode); err != nil {
+				return err
+			}
+		}
+
+		return nil
+	})
 }
--- a/pkg/fileutils/contained.go
+++ b/pkg/fileutils/contained.go
@@ -1,0 +1,36 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package fileutils
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// WriteContainedFile writes content to filePath (relative) inside targetDir,
+// ensuring the resulting path does not escape targetDir. Parent directories
+// are created with dirPerm, and the file is written atomically with filePerm.
+//
+// targetDir must already be filepath.Clean'd by the caller.
+func WriteContainedFile(targetDir, filePath string, content []byte, dirPerm, filePerm os.FileMode) error {
+	cleanTarget := targetDir + string(os.PathSeparator)
+	destPath := filepath.Clean(filepath.Join(targetDir, filepath.FromSlash(filePath)))
+
+	if !strings.HasPrefix(destPath, cleanTarget) {
+		return fmt.Errorf("path traversal detected: file %q escapes target directory", filePath)
+	}
+
+	parentDir := filepath.Dir(destPath)
+	if err := os.MkdirAll(parentDir, dirPerm); err != nil {
+		return fmt.Errorf("creating directory %q: %w", parentDir, err)
+	}
+
+	if err := AtomicWriteFile(destPath, content, filePerm); err != nil {
+		return fmt.Errorf("writing file %q: %w", filePath, err)
+	}
+
+	return nil
+}
--- a/pkg/skills/installer.go
+++ b/pkg/skills/installer.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	ociskills "github.com/stacklok/toolhive-core/oci/skills"
+	"github.com/stacklok/toolhive/pkg/fileutils"
 )
 
 const (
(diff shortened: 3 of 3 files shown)


pkg/skills/gitresolver/resolver.go
```diff
@@ -0,0 +1,226 @@
...
+	cloneConfig := &git.CloneConfig{
+		URL: ref.URL,
+	}
+	if ref.Ref != "" {
+		// Try as branch/tag first (go-git will determine the correct one).
+		// If the ref looks like a full commit hash, use commit checkout instead.
+		if len(ref.Ref) == 40 && isHex(ref.Ref) {
+			cloneConfig.Commit = ref.Ref
+		} else {
+			cloneConfig.Branch = ref.Ref
```
aponcedeleonch, line 112: **[BUG-1] Tag refs won't clone.**

Tag names like `v1.0.0` get set as `cloneConfig.Branch`, which results in `plumbing.NewBranchReferenceName` creating `refs/heads/v1.0.0`. go-git does **not** auto-detect branch vs tag — so tags will fail to clone.

The `CloneConfig` struct already has a `Tag` field, and `client.go` already handles it with `plumbing.NewTagReferenceName`. The resolver just never uses it.

This is a user-visible failure for the most common use case (`git://...@v1.0.0`). Consider detecting tag vs branch (e.g. semver-like patterns → `Tag`) or trying both reference types with a fallback.

pkg/skills/gitresolver/writer.go
```diff
@@ -0,0 +1,105 @@
...
+			return fmt.Errorf("writing file %q: %w", f.Path, err)
+		}
+	}
+
+	return nil
+}
+
+// validatePathNoSymlinks walks up from the target path checking each existing
+// path component for symlinks.
+func validatePathNoSymlinks(targetDir string) error {
```
aponcedeleonch: **[DUP-1] `validatePathNoSymlinks` is a near-verbatim copy of `pkg/skills/installer.go:199-230`.**

The only difference is the error message string (`"refusing to extract"` vs `"refusing to write"`). This should be extracted to a shared location — `pkg/fileutils/` already exists and has `AtomicWriteFile`, so it would be a natural home. Both `installer.go` and `writer.go` should call the shared version.

pkg/skills/gitresolver/writer.go
```diff
@@ -0,0 +1,105 @@
...
+	filePermissionMask os.FileMode = 0644
+)
+
+// WriteFiles writes resolved skill files to the target directory.
+// If force is true, any existing directory is removed before writing.
+//
+// Security: targetDir is produced by PathResolver.GetSkillPath (a trusted
+// internal source that builds paths from known base directories). File paths
+// within the archive are validated via containment check against targetDir.
+func WriteFiles(files []FileEntry, targetDir string, force bool) error {
```
aponcedeleonch, line 21: **[DUP-2] `WriteFiles` containment logic is duplicated from `pkg/skills/installer.go:114-140`.**

The following are identical between both files:
- Constants `dirPermissions = 0750` and `filePermissionMask = 0644`
- `cleanTarget := filepath.Clean(targetDir) + string(os.PathSeparator)`
- Containment check: `!strings.HasPrefix(destPath, cleanTarget)`
- Permission sanitization: `(f.Mode & 0o777) & filePermissionMask`
- The directory creation + error format strings

The only material difference is the input type (`gitresolver.FileEntry` vs OCI's `FileEntry`, which have the same shape). Consider extracting the shared write-with-containment logic and constants to a common helper.

pkg/skills/gitresolver/reference.go
```diff
@@ -0,0 +1,186 @@
...
+	if r.Path != "" {
+		return path.Base(r.Path)
+	}
+	// Extract from URL: "https://github.com/org/repo" -> "repo"
+	trimmed := strings.TrimSuffix(r.URL, ".git")
+	return path.Base(trimmed)
+}
+
+// validateHost checks the host is not localhost, a private IP, or empty.
+func validateHost(host string) error {
```
aponcedeleonch, line 124: **[DUP-3] `validateHost` reimplements SSRF validation that already exists in `pkg/networking/utilities.go`.**

The existing `isPrivateIP` (unexported) and `IsLocalhost` in `pkg/networking/` cover the same ground and are already tested. The existing code uses explicit CIDR blocks including `fc00::/7` (IPv6 ULA) which this reimplementation may not fully cover via `net.IP.IsPrivate()`.

Consider exporting `IsPrivateIP` from `pkg/networking/` and reusing it here, or creating a combined `IsPrivateOrLocalhost(hostname string) error` utility. This gives a single source of truth for SSRF prevention across the codebase.

pkg/skills/gitresolver/resolver_test.go
```diff
@@ -0,0 +1,205 @@
...
+	resolver := NewResolver(WithGitClient(git.NewDefaultGitClient()))
+	ref := &GitReference{URL: dir}
+
+	result, err := resolver.Resolve(t.Context(), ref)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "reading SKILL.md")
+	assert.Nil(t, result)
+}
+
+func TestResolver_Resolve_ContextCancellation(t *testing.T) {
```
aponcedeleonch, line 250: **[NEW-6] This test hits real GitHub and is flaky in offline/CI environments.**

It connects to `https://github.com/nonexistent/nonexistent-repo-12345` with a cancelled context. The comment says "Should fail from context cancellation (or network error)" — the nondeterminism means different error paths are exercised depending on network availability.

Consider using the `WithGitClient` option with a mock client that blocks on clone, so the test deterministically exercises the context cancellation path without network access.

pkg/skills/gitresolver/resolver.go
```diff
@@ -0,0 +1,226 @@
...
+	auth := ResolveAuth(cloneURL)
+	var opts []git.ClientOption
+	if auth != nil {
+		opts = append(opts, git.WithAuth(auth))
+	}
+	return git.NewDefaultGitClient(opts...)
+}
+
+// Resolve clones a git repository and extracts skill files from it.
+func (r *defaultResolver) Resolve(ctx context.Context, ref *GitReference) (*ResolveResult, error) {
```
aponcedeleonch, line 93: **[Coverage] `resolver.go` is at 60.86% patch coverage (26 missing, 10 partials).**

Critical untested paths:
- **Commit-based checkout** (line 99-100): No test provides a 40-char hex `Ref`, so the `cloneConfig.Commit` branch is never exercised.
- **Production auth resolution** (line 74-84): `clientForURL` when `fixedClient` is nil. All tests inject a fixed client.
- **`collectFiles` error paths** (lines 158-211): Only indirectly tested through `Resolve`. No test covers `Head()` failure, `CommitObject()` failure, or file read errors.

Similarly, `writer.go` is at 60.97% — `validatePathNoSymlinks` (symlink rejection), `MkdirAll` failures, and nested file path creation are all untested.

pkg/skills/gitresolver/resolver.go
```diff
@@ -0,0 +1,226 @@
...
+			Content: []byte(content),
+			Mode:    mode,
+		})
+	}
+
+	return files, nil
+}
+
+// isHex checks if a string is a valid hexadecimal string.
+func isHex(s string) bool {
```
aponcedeleonch, line 225: **[MINOR-1] `isHex("")` returns `true`.**

The `for _, c := range s` loop iterates zero times on an empty string and falls through to `return true`. Currently guarded by the `len(ref.Ref) == 40` check at the call site, but the function is semantically incorrect on its own.

```suggestion
func isHex(s string) bool {
	if s == "" {
		return false
	}
	for _, c := range s {
```

Alternatively, consider using `encoding/hex.DecodeString()` from stdlib.

pkg/git/client.go
```diff
@@ -247,3 +247,17 @@ func (*DefaultGitClient) updateRepositoryInfo(repoInfo *RepositoryInfo) error {
 
 	return nil
 }
+
+// HeadCommitHash returns the commit hash of the HEAD reference.
+func HeadCommitHash(repoInfo *RepositoryInfo) (string, error) {
```
aponcedeleonch: **[NEW-1] `HeadCommitHash` is a package-level function, not on the `Client` interface.**

Every other operation on `RepositoryInfo` goes through the `Client` interface or is a method on `DefaultGitClient`. As a free function, this cannot be mocked by callers, which limits testability of the resolver's commit-hash retrieval. Consider making it a method on `Client` for consistency.

pkg/skills/gitresolver/writer.go
```diff
@@ -0,0 +1,105 @@
...
+	// Pre-extraction: validate that no existing path components are symlinks.
+	if err := validatePathNoSymlinks(targetDir); err != nil {
+		return fmt.Errorf("target path validation: %w", err)
+	}
+
+	if err := os.MkdirAll(targetDir, dirPermissions); err != nil { // lgtm[go/path-injection] #nosec G304
+		return fmt.Errorf("creating target directory: %w", err)
+	}
+
+	cleanTarget := filepath.Clean(targetDir) + string(os.PathSeparator)
```
aponcedeleonch: **[NEW-3] Redundant `filepath.Clean`.**

`targetDir` is already cleaned at line 28 (`targetDir = filepath.Clean(targetDir)`). The `filepath.Clean` here is a no-op.

```suggestion
	cleanTarget := targetDir + string(os.PathSeparator)
```

pkg/skills/gitresolver/auth.go
```diff
@@ -0,0 +1,82 @@
...
+
+// tokenMapping maps environment variable names to the git hosts they are scoped to.
+// Tokens are only sent to their matching host to prevent credential exfiltration.
+var tokenMapping = []struct {
+	envVar string
+	hosts  []string // empty means the token is sent to any host (user opt-in)
+}{
+	{envVar: "GITHUB_TOKEN", hosts: []string{"github.com"}},
+	{envVar: "GITLAB_TOKEN", hosts: []string{"gitlab.com"}},
+	{envVar: "GIT_TOKEN", hosts: nil}, // fallback: sent to any host
```
aponcedeleonch, line 24: **[NEW-5] `GIT_TOKEN` is silently sent to any host.**

With `hosts: nil`, this token is sent to any arbitrary host. If a user sets `GIT_TOKEN` and encounters a malicious `git://evil.com/...` reference, their token is sent without any indication. Consider at minimum adding a debug-level log when the fallback token is used for a non-standard host (i.e., not github.com/gitlab.com), so users can audit credential usage.

pkg/skills/gitresolver/writer.go
```diff
@@ -0,0 +1,70 @@
...
+		if err := os.MkdirAll(parentDir, skills.DirPermissions); err != nil {
+			return fmt.Errorf("creating directory %q: %w", parentDir, err)
+		}
+
+		// Sanitize file permissions: strip setuid/setgid/sticky, cap at 0644
+		mode := (f.Mode & 0o777) & skills.FilePermissionMask
+
+		if err := os.WriteFile(destPath, f.Content, mode); err != nil {
+			return fmt.Errorf("writing file %q: %w", f.Path, err)
+		}
```
aponcedeleonch, line 51: **Consolidate file-writing logic and use atomic writes + locking.**

The containment-check + write loop here (lines 46-66) is near-identical to `installer.go:114-139`. More importantly, both use raw `os.WriteFile`, while the rest of the codebase uses `fileutils.AtomicWriteFile` and `fileutils.WithFileLock` for safe file operations (see `config_editor.go`, `encrypted.go`, `file_status.go`).

Suggest extracting a shared helper and wrapping the whole operation in a lock:

```suggestion
// writeContainedFile writes a single file with containment check, atomic write, and sanitized permissions.
func writeContainedFile(cleanTarget, targetDir string, filePath string, content []byte, mode fs.FileMode) error {
	destPath := filepath.Clean(filepath.Join(targetDir, filepath.FromSlash(filePath)))

	if !strings.HasPrefix(destPath, cleanTarget) {
		return fmt.Errorf("path traversal detected: file %q escapes target directory", filePath)
	}

	parentDir := filepath.Dir(destPath)
	if err := os.MkdirAll(parentDir, skills.DirPermissions); err != nil {
		return fmt.Errorf("creating directory %q: %w", parentDir, err)
	}

	sanitizedMode := (mode & 0o777) & skills.FilePermissionMask
	if err := fileutils.AtomicWriteFile(destPath, content, sanitizedMode); err != nil {
		return fmt.Errorf("writing file %q: %w", filePath, err)
	}
	return nil
}
```

Then both `WriteFiles` here and `writeFiles` in `installer.go` call it, and the top-level `WriteFiles` wraps everything in `fileutils.WithFileLock(targetDir, func() error { ... })` to prevent concurrent installs from corrupting each other.

PR #4053, opened 2026-03-09
Handle mid-session backend crashes gracefully

### LINKED ISSUE #3875: [vMCP] Handle mid-session backend crashes gracefully
When a backend crashes, times out, or terminates its connection mid-session, the vMCP session should continue operating with the
  remaining healthy backends rather than failing entirely. This is distinct from Issue 4 (backend session expiration), which triggers
  re-initialization. A crash or connection loss should propagate a clear per-backend error and leave the session alive.

  Behavior:

  Failure Type: Single backend crashes
  Session Behaviour: Session continues with other backends
  Client Action: Returns error for that backend; client can retry or use other tools
  ────────────────────────────────────────
  Failure Type: Multiple backends fail
  Session Behaviour: Session continues with remaining healthy backends
  Client Action: Returns errors for failed backends
  ────────────────────────────────────────
  Failure Type: All backends fail
  Session Behaviour: Session remains active but all tool calls fail
  Client Action: Client should delete session and reinitialize
  ────────────────────────────────────────
  Failure Type: Backend times out
  Session Behaviour: Return timeout error
  Client Action: Client can retry or switch to a different tool
  ────────────────────────────────────────
  Failure Type: Backend terminates gracefully
  Session Behaviour: Detect via connection close, mark unavailable
  Client Action: Client receives error, can retry later

  Error messages: Return "backend <id> unavailable: <reason>" — never a generic "tool not found".

  Acceptance Criteria

  - A single backend crash does not terminate the vMCP session
  - Tool calls to the failed backend return a clear error identifying the backend and reason
  - Tool calls to healthy backends in the same session continue to work normally
  - If all backends fail, the session remains active but all tool calls return errors
  - Error messages identify the backend by ID and include the failure reason
  - Unit tests cover: single backend failure, multiple backend failure, all backends failure, healthy backends unaffected

yrobla:
## Summary

Why: When a backend crashed mid-session, errors propagated with no indication of which backend failed, making it impossible for clients to distinguish a single-backend outage from a total session failure. This also violated the issue requirement that error messages identify the backend.

Fixes #3875

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

What changed:

lookupBackend now returns *vmcp.BackendTarget alongside the connection
CallTool, ReadResource, and GetPrompt wrap backend errors as backend "<id>" unavailable: <cause>, keeping the original error unwrappable via errors.Is
Session resilience was already correct (one backend crash never terminates the session); this change makes the error surface match the spec
Affected components: pkg/vmcp/session/default_session.go, default_session_test.go

No

### DIFF
--- a/pkg/vmcp/session/default_session.go
+++ b/pkg/vmcp/session/default_session.go
@@ -15,16 +15,6 @@ import (
 	"github.com/stacklok/toolhive/pkg/vmcp/session/internal/backend"
 )
 
-// isCallerDrivenError reports whether err reflects the caller cancelling or
-// timing out the request rather than a backend failure. These errors should
-// not be labelled "unavailable" because the backend may be healthy.
-func isCallerDrivenError(err error) bool {
-	return errors.Is(err, context.Canceled) ||
-		errors.Is(err, context.DeadlineExceeded) ||
-		errors.Is(err, vmcp.ErrCancelled) ||
-		errors.Is(err, vmcp.ErrTimeout)
-}
-
 // Compile-time assertions: defaultMultiSession must implement both interfaces.
 var _ MultiSession = (*defaultMultiSession)(nil)
 var _ transportsession.Session = (*defaultMultiSession)(nil)
@@ -153,10 +143,7 @@ func (s *defaultMultiSession) CallTool(
 	defer done()
 	result, err := conn.CallTool(ctx, toolName, arguments, meta)
 	if err != nil {
-		if isCallerDrivenError(err) {
-			return nil, err
-		}
-		return nil, fmt.Errorf("backend %s unavailable: %w", target.WorkloadID, err)
+		return nil, fmt.Errorf("backend %q request failure: %w", target.WorkloadID, err)
 	}
 	return result, nil
 }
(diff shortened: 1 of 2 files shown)


pkg/vmcp/session/default_session.go
```diff
@@ -136,12 +146,19 @@ func (s *defaultMultiSession) CallTool(
...
 		return nil, err
 	}
 	defer done()
-	return conn.CallTool(ctx, toolName, arguments, meta)
+	result, err := conn.CallTool(ctx, toolName, arguments, meta)
+	if err != nil {
+		if isCallerDrivenError(err) {
+			return nil, err
+		}
+		return nil, fmt.Errorf("backend %s unavailable: %w", target.WorkloadID, err)
```
jerm-dro: nit: `unavailable` isn't actually true. The backend may be available and returned a failure for some other reason.

Let's change this to a more generic prefix, so debuggers aren't confused. Suggestion: "backend request failure: %w"

pkg/vmcp/session/default_session.go
```diff
@@ -15,6 +15,16 @@ import (
...
 
+// isCallerDrivenError reports whether err reflects the caller cancelling or
+// timing out the request rather than a backend failure. These errors should
+// not be labelled "unavailable" because the backend may be healthy.
+func isCallerDrivenError(err error) bool {
+	return errors.Is(err, context.Canceled) ||
+		errors.Is(err, context.DeadlineExceeded) ||
+		errors.Is(err, vmcp.ErrCancelled) ||
+		errors.Is(err, vmcp.ErrTimeout)
+}
```
jerm-dro: suggestion: delete this. Let's just wrap all errors regardless. Just because something `isCallerDrivenError` doesn't mean we don't want it wrapped with the backend. For example, if the error is `context.DeadlineExceeded` it'd be great to know which backend we timed out waiting on.

PR #4057, opened 2026-03-09
Return RFC 6750 compliant JSON error responses from auth middleware

### LINKED ISSUE #4055: Auth middleware returns plain text 401 instead of RFC 6750 compliant JSON error response
## Description

When a request arrives at the vMCP (or MCPServer) without an `Authorization` header, the auth middleware returns a `401` response with a plain text body:

```
authorization header required
```

RFC 6750 Section 3 requires the error response body to be JSON:

```json
{
  "error": "invalid_token",
  "error_description": "authorization header required"
}
```

The `Content-Type` header is also incorrectly set to `text/plain; charset=utf-8` instead of `application/json`.

## Impact

Any RFC 6750-compliant OAuth client that tries to parse the 401 response body as JSON will fail. This is observed with Claude Code, which surfaces the following error:

```
Error: HTTP 401: Invalid OAuth error response: SyntaxError: JSON Parse error: Unexpected identifier "authorization". Raw body: authorization header required
```

## Root Cause

In `pkg/auth/token.go`, the `Middleware` function calls `http.Error(w, err.Error(), http.StatusUnauthorized)` when the Authorization header is missing. `http.Error` sets `Content-Type: text/plain` and writes the raw error string as the body.

```go
tokenString, err := ExtractBearerToken(r)
if err != nil {
    w.Header().Set("WWW-Authenticate", v.buildWWWAuthenticate(false, ""))
    http.Error(w, err.Error(), http.StatusUnauthorized) // plain text, not JSON
    return
}
```

## Expected Behaviour

The 401 response should:
- Set `Content-Type: application/json`
- Return a JSON body conforming to RFC 6750 Section 3

## Steps to Reproduce

1. Configure a VirtualMCPServer with `incomingAuth.type: oidc`
2. Connect with an MCP client (e.g. Claude Code) without a valid Bearer token
3. Observe the `401` response with plain text body

## References

- [RFC 6750 Section 3 — The WWW-Authenticate Response Header Field](https://www.rfc-editor.org/rfc/rfc6750#section-3)
- Relevant code: `pkg/auth/token.go` — `TokenValidator.Middleware`

amirejaz:
## Summary

The token validator middleware was returning plain text error bodies via `http.Error()` on `401` responses. RFC 6750-compliant OAuth clients (e.g. Claude Code) expect a JSON error body and fail to parse the plain text response with:

```
JSON Parse error: Unexpected identifier "authorization". Raw body: authorization header required
```

This PR adds a `writeOAuthError` helper that returns a proper RFC 6750 Section 3 JSON error response with `Content-Type: application/json` and a body of the form:

```json
{
  "error": "invalid_request",
  "error_description": "authorization header required"
}
```

All three error paths in `TokenValidator.Middleware` are updated:
- Missing/invalid Authorization header → `invalid_request`
- Invalid token → `invalid_token`
- Invalid claims → `invalid_token`

Fixes #4055

## Type of change

- [x] Bug fix

## Test plan

- [x] Added `TestMiddleware_RFC6750JSONErrorResponse` covering missing Authorization header, wrong scheme, and malformed bearer token — verifying `Content-Type: application/json`, valid JSON body, correct `error` code, and non-empty `error_description`
- [x] Full `pkg/auth/...` test suite passes

### DIFF
--- a/pkg/auth/token.go
+++ b/pkg/auth/token.go
@@ -1038,8 +1038,8 @@ func (v *TokenValidator) buildWWWAuthenticate(errorCode string, errDescription s
 
 // RFC 6750 error code constants for Bearer token authentication.
 const (
-	OAuthErrInvalidRequest  = "invalid_request"
-	OAuthErrInvalidToken    = "invalid_token"
+	OAuthErrInvalidRequest    = "invalid_request"
+	OAuthErrInvalidToken      = "invalid_token"
 	OAuthErrInsufficientScope = "insufficient_scope"
 )
 


pkg/auth/token_test.go
```diff
@@ -2140,3 +2140,90 @@ func TestTokenValidator_GoogleTokeninfoIntegration(t *testing.T) {
...
+			}
+			if err := json.NewDecoder(res.Body).Decode(&body); err != nil {
+				t.Fatalf("response body is not valid JSON: %v", err)
+			}
+			if body.Error != tt.wantErrorCode {
+				t.Errorf("expected error code %q, got %q", tt.wantErrorCode, body.Error)
+			}
+			if !strings.Contains(body.ErrorDescription, tt.wantDescSubstring) {
+				t.Errorf("expected error_description to contain %q, got %q", tt.wantDescSubstring, body.ErrorDescription)
+			}
```
jerm-dro: nit: can you use `require.NoError` and such?

I prefer using required since it makes the tests less nested and easier to read.

amirejaz: addressed

jerm-dro (dismissed):

pkg/auth/token.go
```diff
@@ -1035,22 +1035,40 @@ func (v *TokenValidator) buildWWWAuthenticate(includeError bool, errDescription
...
+	_, _ = w.Write(body)
+}
+
 // Middleware creates an HTTP middleware that validates JWT tokens and creates Identity.
 func (v *TokenValidator) Middleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// Extract the bearer token from the Authorization header
 		tokenString, err := ExtractBearerToken(r)
 		if err != nil {
 			w.Header().Set("WWW-Authenticate", v.buildWWWAuthenticate(false, ""))
```
reyortiz3: question the RFC shows this as an example:
```
     HTTP/1.1 401 Unauthorized
     WWW-Authenticate: Bearer realm="example",
                       error="invalid_token",
                       error_description="The access token expired"
```

which `buildWWWAuthenticate(includeError bool, errDescription string)` does when passing `includeError=true`, but it seems it doesn't support the `invalid_request` or `insufficient_scope`.

pkg/auth/token.go
```diff
@@ -1035,22 +1035,40 @@ func (v *TokenValidator) buildWWWAuthenticate(includeError bool, errDescription
...
+		Error:            errorCode,
+		ErrorDescription: description,
+	})
+	if err != nil {
+		http.Error(w, "internal server error", http.StatusInternalServerError)
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(status)
+	_, _ = w.Write(body)
```
reyortiz3, line 1064: question to understand the issue, what was reported was that the response was in plain-text instead Json, right? or also that the had to return this payload:

```
{
  "error": "invalid_request",
  "error_description": "authorization header required"
}
```
?

pkg/auth/token.go
```diff
@@ -1035,22 +1035,40 @@ func (v *TokenValidator) buildWWWAuthenticate(includeError bool, errDescription
 	return "Bearer " + strings.Join(parts, ", ")
 }
 
+// writeOAuthError writes an RFC 6750 compliant JSON error response.
+func writeOAuthError(w http.ResponseWriter, errorCode, description string, status int) {
+	body, err := json.Marshal(struct {
+		Error            string `json:"error"`
+		ErrorDescription string `json:"error_description"`
```
reyortiz3: if it has to return the payload in this format, can we move this to a proper/separate DTO struct?

pkg/auth/token.go
```diff
@@ -1035,22 +1035,40 @@ func (v *TokenValidator) buildWWWAuthenticate(includeError bool, errDescription
...
+
 // Middleware creates an HTTP middleware that validates JWT tokens and creates Identity.
 func (v *TokenValidator) Middleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// Extract the bearer token from the Authorization header
 		tokenString, err := ExtractBearerToken(r)
 		if err != nil {
 			w.Header().Set("WWW-Authenticate", v.buildWWWAuthenticate(false, ""))
-			http.Error(w, err.Error(), http.StatusUnauthorized)
+			writeOAuthError(w, "invalid_request", err.Error(), http.StatusUnauthorized)
```
reyortiz3: nit: just consider using consts

pkg/auth/token_test.go
```diff
@@ -2140,3 +2141,79 @@ func TestTokenValidator_GoogleTokeninfoIntegration(t *testing.T) {
...
+			tt.setupRequest(req)
+			rr := httptest.NewRecorder()
+
+			handler.ServeHTTP(rr, req)
+
+			res := rr.Result()
+			defer res.Body.Close()
+
+			require.Equal(t, tt.wantStatus, res.StatusCode)
+			require.True(t, strings.HasPrefix(res.Header.Get("Content-Type"), "application/json"),
```
reyortiz3, line 2211: nit: `require.Equal`?

pkg/auth/token.go
amirejaz: This is the error when the response is returned in plain text format:
`JSON Parse error: Unexpected identifier "authorization". Raw body: authorization header required`

pkg/auth/token.go
amirejaz: addressed

pkg/auth/token.go
amirejaz: good catch, addressed

pkg/auth/token.go
amirejaz: addressed

pkg/auth/token.go
reyortiz3: I see, the question was mainly the response format:
```
{
  "error": "invalid_request",
  "error_description": "authorization header required"
}
```
for I saw, the RFC requires those in the `WWW-Authenticate` header, not necessarily in the body, just any Json it seems. But I think it's fine in the response body to also return them as json

PR #4063, opened 2026-03-10
Make proxy response timeout configurable

### LINKED ISSUE #4061: Streamable HTTP proxy has hardcoded 30s response timeout causing 504 on long-running MCP tools
### Description

The streamable HTTP proxy (`pkg/transport/proxy/streamable/streamable_proxy.go`) has a hardcoded 30-second response timeout (`defaultResponseTimeout`). Any MCP tool call that takes longer than 30 seconds returns HTTP 504 Gateway Timeout, even though the backend MCP server is still processing the request.

This only affects MCP servers running with **stdio transport** — the default for most container-based servers (npx://, uvx://, and Docker images without a native HTTP server). Servers that natively use `--transport sse` or `--transport streamable-http` (like `fetch`) go through the transparent proxy (`pkg/transport/proxy/transparent/`), which has no such timeout.

### Steps to Reproduce

1. Run a stdio transport MCP server with `thv run` (e.g., `time`, or any npx/uvx-based server)
2. Call an MCP tool that takes longer than 30 seconds to respond
3. The proxy returns HTTP 504 after exactly 30 seconds

**Note**: Servers with native HTTP transport (like `fetch`, which uses `streamable-http`) are not affected — they use a different proxy code path.

### Expected Behavior

Long-running MCP tool calls should complete successfully. A 30-second timeout is too aggressive for many legitimate use cases (e.g., tools that query large datasets, perform complex searches, or interact with slow external APIs).

### Actual Behavior

The proxy returns HTTP 504 Gateway Timeout after 30 seconds. The three affected code paths are:
- `handleSingleRequest` — JSON response path
- `handleSingleRequestSSE` — SSE streaming path
- `processSingleMessage` — batch processing path

All three use the same `defaultResponseTimeout` constant at line 33.

### Proposed Fix

1. Bump the default from 30s to 5m (sufficient for most use cases without configuration)
2. Add `TOOLHIVE_PROXY_RESPONSE_TIMEOUT` environment variable for users who need a different value (follows the existing `TOOLHIVE_REMOTE_HEALTHCHECKS` pattern for env var overrides)

### Environment

- ToolHive version: v0.11.1 (and current main)
- OS: macOS (the code path is platform-independent, so likely affects Linux too)

gkatz2:
## Summary

- The streamable HTTP proxy (used by stdio transport) had a hardcoded 30-second response timeout. Any MCP tool call exceeding 30s returned a timeout error, even though the backend server was still processing. This made legitimate long-running tools unusable.
- Bump the default from 30s to 5m and add `TOOLHIVE_PROXY_RESPONSE_TIMEOUT` environment variable for users who need a different value. Follows the existing `TOOLHIVE_REMOTE_HEALTHCHECKS` pattern for env var overrides.

Fixes #4061

## Type of change

- [x] Bug fix

## Test plan

Built a custom stdio MCP server with a 35-second delay on tool calls.
Verified the fix in both directions:

| Scenario | Timeout | Result |
|----------|---------|--------|
| `TOOLHIVE_PROXY_RESPONSE_TIMEOUT=30s` + 35s tool | 30s | Timeout at 30.00s (expected) |
| Default (no env var) + 35s tool | 5m | Success at 35.00s |
| `TOOLHIVE_PROXY_RESPONSE_TIMEOUT=1us` + any tool | 1us | Timeout (even initialize fails) |

## Changes

| File | Change |
|------|--------|
| `pkg/transport/proxy/streamable/streamable_proxy.go` | Bump default to 5m, add `responseTimeout` field, add `resolveResponseTimeout()`, update 3 usage sites |
| `pkg/transport/proxy/streamable/streamable_proxy_test.go` | Table-driven tests for timeout resolution + constructor wiring tests |
| `docs/arch/03-transport-architecture.md` | Document `TOOLHIVE_PROXY_RESPONSE_TIMEOUT` env var |

The default proxy response timeout increases from 30 seconds to 5 minutes. Users who were hitting 504 timeouts on slow MCP tools will no longer see them without any configuration change. Users who need a different timeout can set `TOOLHIVE_PROXY_RESPONSE_TIMEOUT` (e.g., `10m`, `120s`).

- Only the streamable HTTP proxy for stdio transport is affected (`pkg/transport/proxy/streamable/`). The transparent proxy used by `--transport sse` and `--transport streamable-http` has no such timeout.
- The env var is read once at proxy creation time (in `NewHTTPProxy`) and stored as a struct field, avoiding repeated `os.Getenv` calls on every request.
- Invalid or non-positive values log a warning and fall back to the default.

### DIFF
--- a/pkg/transport/proxy/streamable/streamable_proxy.go
+++ b/pkg/transport/proxy/streamable/streamable_proxy.go
@@ -719,6 +719,7 @@ func resolveRequestTimeout() time.Duration {
 		"env_var", proxyRequestTimeoutEnv, "value", v, "default", defaultRequestTimeout)
 	return defaultRequestTimeout
 }
+
 // createWaiter registers a waiter channel for the given request ID and returns cleanup fn.
 func (p *HTTPProxy) createWaiter(sessID string, id jsonrpc2.ID) (chan jsonrpc2.Message, func()) {
 	key := idKeyFromID(id)


pkg/transport/proxy/streamable/streamable_proxy.go
```diff
@@ -29,14 +30,20 @@ const (
 	// StreamableHTTPEndpoint is the endpoint for streamable HTTP.
 	StreamableHTTPEndpoint = "/mcp"
 
-	// Default timeouts and buffer sizes
-	defaultResponseTimeout = 30 * time.Second
+	// defaultResponseTimeout is the maximum time to wait for a response from the
+	// MCP server. Override with TOOLHIVE_PROXY_RESPONSE_TIMEOUT (e.g. "30s", "2m").
+	defaultResponseTimeout = 5 * time.Minute
```
reyortiz3: - would suggest to `60s` default timeout to match common timeouts, e.g. MCP TS SDK
- also suggest to rename "response" --> "request" in the names, e.g. `defaultRequestTimeout`, `TOOLHIVE_PROXY_REQUEST_TIMEOUT`

cc: @rdimitrov @JAORMX

gkatz2: Thanks for the review!

- **Timeout value**: Happy to adjust the default if that's what you all prefer. My argument for a higher value: for some agent tasks, it's worth waiting for a slow backend rather than timing out, and it's difficult for a middleware layer to know when these situations arise. For example, Slack's `conversations_unreads` call consistently exceeds 30 seconds for my org's workspace. With a 30s default, that tool call was failing every time for me. I picked 5 minutes to give headroom for even slower MCP tools. A generous default means ToolHive works out of the box for more use cases without requiring an env var override — which, for large organizations means a hassle either for lots of individuals or a hassle for a central dev platform team. Do you see a downside to a higher default timeout value than 60s?
- **Naming**: The constant `defaultResponseTimeout` was the pre-existing name on main, so I kept it and used "response" consistently in `TOOLHIVE_PROXY_RESPONSE_TIMEOUT`. Happy to rename everything to "request" if that is your preference.

reyortiz3: Hi @gkatz2,

- **Timeout**: I would still recommend the 60s mainly these reasons:
   1. It's the most common default timeout in official [MCP SDKs](https://github.com/modelcontextprotocol/typescript-sdk/blob/b0ef89ffaf6db8b3c52cd8919e8949b0f1da9ca4/packages/core/src/shared/protocol.ts#L110). There is a proposal open to make it default to 60s https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1539 
   2. It can make it not obvious to debug/troubleshoot slow MCPs
   3. A slow MCP tool might in general cause a degraded experience, could be sign to review the design of the agentic system and see if it's possible to use techniques to reduce the latency
   4. it's still overridable with the env var
- **Naming**: renaming `response` -> `request` should be fine as it's the _request_ what times out

Still, a valid concern on how to make it easier to override the env var since other uses cases might even require different timeouts

gkatz2: Okay, default timeout adjusted to 60s and "response" naming switched to "request". Update has been pushed.

reyortiz3 (approved):

pkg/transport/proxy/streamable/streamable_proxy.go
reyortiz3: Thanks for making this change.

aponcedeleonch (approved):
Looks good, thanks for the contribution!

## Suggestion: Per-MCP-Server Request Timeout (follow-up)

The global env var solves the immediate problem well. As a natural next step, it would be valuable to support a per-server `--proxy-request-timeout` flag on `thv run`, since different MCP servers can have very different response characteristics. The env var would remain as a global fallback.

**This can be done as a separate follow-up** — just leaving breadcrumbs on how it fits into the existing architecture.

### Precedence

```
--proxy-request-timeout flag  >  TOOLHIVE_PROXY_REQUEST_TIMEOUT env var  >  60s default
```

### Implementation sketch

The pattern already exists for other per-server settings (`--trust-proxy-headers`, `--endpoint-prefix`). It touches ~7 files with small additions each:

**`pkg/runner/config.go`** — add field to RunConfig:
```go
ProxyRequestTimeout time.Duration `json:"proxy_request_timeout,omitempty" yaml:"proxy_request_timeout,omitempty"`
```

**`pkg/runner/config_builder.go`** — add builder option:
```go
func WithProxyRequestTimeout(d time.Duration) RunConfigBuilderOption {
    return func(c *RunConfig) { c.ProxyRequestTimeout = d }
}
```

**`cmd/thv/app/run_flags.go`** — add CLI flag:
```go
cmd.Flags().DurationVar(&f.ProxyRequestTimeout, "proxy-request-timeout", 0,
    "Request timeout for the streamable HTTP proxy used by stdio servers (e.g. 2m, 120s). 0 uses global default.")
```

**`pkg/transport/proxy/streamable/streamable_proxy.go`** — accept per-server value, fall back to env then default:
```go
func resolveRequestTimeout(perServer time.Duration) time.Duration {
    if perServer > 0 {
        return perServer
    }
    if v := os.Getenv(proxyRequestTimeoutEnv); v != "" {
        if d, err := time.ParseDuration(v); err == nil && d > 0 {
            return d
        }
    }
    return defaultRequestTimeout
}
```

The remaining files (`types/transport.go`, `runner.go`, `http.go`) just propagate the value through the existing config plumbing.

aponcedeleonch:
I rebased to check if the errors in the e2e tests were fixed by some recent merged PRs on `main`

gkatz2:
@reyortiz3 @aponcedeleonch Sorry for the approval reset — I had to push a fix to a test. The only change since your reviews:

`TestNewHTTPProxyDefaultTimeout` was reading `TOOLHIVE_PROXY_REQUEST_TIMEOUT` from the ambient environment, so it would fail for anyone who has that env var set. Fixed by clearing it with `t.Setenv` (which requires dropping `t.Parallel()`):

```diff
-func TestNewHTTPProxyDefaultTimeout(t *testing.T) {
-	t.Parallel()
+// Not parallel because t.Setenv is needed to clear any ambient
+// TOOLHIVE_PROXY_REQUEST_TIMEOUT from the test runner's environment.
+func TestNewHTTPProxyDefaultTimeout(t *testing.T) { //nolint:paralleltest
+	t.Setenv(proxyRequestTimeoutEnv, "")
 	proxy := NewHTTPProxy("localhost", 0, nil)
 	assert.Equal(t, defaultRequestTimeout, proxy.requestTimeout)
 }
```

No production code changes — just the one test. CI is running now. Would appreciate a re-approval when you get a chance.

pkg/transport/proxy/streamable/streamable_proxy.go
```diff
@@ -694,6 +702,23 @@ func (p *HTTPProxy) handleNotificationOrClientResponse(w http.ResponseWriter, ms
...
+		d, err := time.ParseDuration(v)
+		if err != nil || d <= 0 {
+			slog.Warn("invalid proxy request timeout, using default",
+				"env_var", proxyRequestTimeoutEnv, "value", v, "default", defaultRequestTimeout)
+			return defaultRequestTimeout
+		}
+		slog.Debug("using custom proxy request timeout", "timeout", d)
+		return d
+	}
+	return defaultRequestTimeout
```
reyortiz3: (minor): took a second look here with minor improvement to lower code/cognitive complexity

```
if v := os.Getenv(proxyRequestTimeoutEnv); v != "" {
		d, _ := time.ParseDuration(v)
		if d > 0 {
            return d
		}
        slog.Warn("invalid proxy request timeout, using default",
				"env_var", proxyRequestTimeoutEnv,
                 "value", v,
                 "default", defaultRequestTimeout)
	}
    slog.Debug("using custom proxy request timeout", "timeout", d)
	return defaultRequestTimeout
}
```

gkatz2: How about this version? It keeps a flat, linear flow — each check either returns or falls through, no nesting:

```go
func resolveRequestTimeout() time.Duration {
	v := os.Getenv(proxyRequestTimeoutEnv)
	if v == "" {
		return defaultRequestTimeout
	}
	d, _ := time.ParseDuration(v)
	if d > 0 {
		slog.Debug("using custom proxy request timeout", "timeout", d)
		return d
	}
	slog.Warn("invalid proxy request timeout, using default",
		"env_var", proxyRequestTimeoutEnv, "value", v, "default", defaultRequestTimeout)
	return defaultRequestTimeout
}
```

It's a little longer, but I find it easier to understand at a glance. I'm happy to go with whatever approach fits your conventions.

reyortiz3: yes, this version with less nesting works too

amirejaz: @gkatz2 I committed your suggestion.

PR #4068, opened 2026-03-10
Add TLS support for Redis connections

### LINKED ISSUE #3681: Auth server: Add TLS support for Redis token storage
The initial implementation does not enable TLS with Redis token storage for the Authorization Server. For many internal-only, single-tenant production workloads, NetworkPolicies + Redis AUTH is a pragmatic and widely-used approach. If you're subject to compliance audits or operating in a multi-tenant/zero-trust environment, add TLS — either natively in Redis or via a service mesh sidecar.

TODO design details

aron-muon:
## Summary

Adds structured TLS configuration to the Redis storage config in `MCPExternalAuthConfig`, supporting separate TLS settings for Redis master and Sentinel connections. This is needed because:

- go-redis `FailoverClient` applies a single `TLSConfig` to both sentinel and master connections
- Sentinel instances (or emulators) may use self-signed certificates while the master (e.g., ElastiCache) uses a proper CA
- Operators need per-connection control over TLS verification and CA certificates

Implemented via a custom `Dialer` on `FailoverOptions` that applies the correct TLS config based on whether the target address is a sentinel or master.

Closes https://github.com/stacklok/toolhive/issues/3681

## Changes

| Layer | File | Change |
|-------|------|--------|
| CRD | `mcpexternalauthconfig_types.go` | Added `RedisTLSConfig` struct with `enabled`, `insecureSkipVerify`, `caCertSecretRef`; added `tls` and `sentinelTls` fields on `RedisStorageConfig` |
| Operator | `controllerutil/authserver.go` | Added `convertRedisTLSConfig` to propagate CRD → RunConfig; volume generation gated on `TLS.Enabled`; CA cert volumes mounted as secrets with 0400 permissions |
| RunConfig | `storage/config.go` | Added `RedisTLSRunConfig` struct with `TLS` and `SentinelTLS` on `RedisRunConfig` |
| Runtime | `storage/redis.go` | Added `RedisTLSConfig` struct, `buildTLSConfig` (returns error on invalid PEM), `newTLSDialer` (named function for testability), `configureTLSDialer` (handles all master/sentinel TLS combinations including mixed TLS/plaintext) |
| Runner | `embeddedauthserver.go` | Added `convertRedisTLSRunConfig` — reads CA cert from file, returns hard error on failure |
| Tests | `redis_tls_test.go` (storage), `redis_tls_test.go` (runner), `authserver_test.go` | `TestConfigureTLSDialer` (9 cases), `TestNewTLSDialer` (3 cases), `TestBuildTLSConfig` (6 cases incl. error paths), `TestConvertRedisTLSRunConfig` (5 cases incl. error), `TestGenerateAuthServerVolumes_RedisTLS` (6 cases) |
| Generated | CRD manifests, swagger docs, CRD API docs | Regenerated |

## CRD Usage

```yaml
kind: MCPExternalAuthConfig
spec:
  embeddedAuthServer:
    storage:
      type: redis
      redis:
        tls:
          enabled: true
          insecureSkipVerify: false          # optional, default false
          caCertSecretRef:                   # optional — custom CA bundle
            name: redis-ca-cert
            key: ca.crt
        sentinelTls:
          enabled: true
          insecureSkipVerify: true           # for self-signed sentinel certs
          caCertSecretRef:                   # optional — alternative to insecureSkipVerify
            name: sentinel-emulator-tls
            key: tls.crt
        sentinelConfig:
          masterName: mymaster
          db: 0
          sentinelAddrs:
            - sentinel-emulator:26379
        aclUserConfig:
          usernameSecretRef:
            name: valkey-auth
            key: username
          passwordSecretRef:
            name: valkey-auth
            key: password
        dialTimeout: 5s
        readTimeout: 3s
        writeTimeout: 3s
```

### Configuration Options

| Field | Type | Description |
|-------|------|-------------|
| `tls.enabled` | bool | Enable TLS for master connections |
| `tls.insecureSkipVerify` | bool | Skip cert verification for master |
| `tls.caCertSecretRef` | SecretKeyRef | Custom CA cert for master verification |
| `sentinelTls.enabled` | bool | Enable TLS for sentinel connections (falls back to `tls` if unset) |
| `sentinelTls.insecureSkipVerify` | bool | Skip cert verification for sentinel |
| `sentinelTls.caCertSecretRef` | SecretKeyRef | Custom CA cert for sentinel verification |

## Design Decisions

- **Separate configs rather than single `InsecureSkipVerify`**: Standard Redis Sentinel deployments use different TLS configurations for sentinel and data nodes. This isn't just a workaround for sentinel emulators — it's the correct general design.
- **Custom `Dialer` handles all TLS combinations**: The dialer is installed whenever master and sentinel need *different* TLS treatment — including using plaintext (`net.Dialer.Dial`) for the side with TLS disabled. This correctly handles master-only, sentinel-only, both-enabled, and mixed configurations. Extracted into named `newTLSDialer` and `configureTLSDialer` functions for testability.
- **Hard errors for misconfiguration**: `buildTLSConfig` returns an error if CA PEM data is malformed (instead of silently creating an empty cert pool that replaces system CAs). `convertRedisTLSRunConfig` returns an error if a configured CA cert file can't be read (instead of silently falling back to system CAs and causing confusing handshake failures).
- **`SecretKeyRef` for CA certs**: Follows the existing CRD pattern used by `aclUserConfig` for credential references.
- **Volume mounting gated on `Enabled`**: CA cert volumes are only created when `TLS.Enabled` is true, preventing unnecessary volume mounts when TLS is disabled.
- **Fallback behavior**: When `sentinelTls` is nil, it falls back to the `tls` config. This means setting only `tls.enabled: true` applies TLS uniformly to both (backward compatible with the simpler model). An explicit `sentinelTls: {enabled: false}` is respected and results in plaintext sentinel connections.

## Test plan

- [x] All existing tests pass
- [x] Full build clean
- [x] Lint clean
- [x] New unit tests for `configureTLSDialer` (all TLS combos), `newTLSDialer` (address matching, plaintext fallback), `buildTLSConfig` (error on invalid PEM), `convertRedisTLSRunConfig` (hard error on missing file), and `GenerateAuthServerVolumes` (volume gated on `Enabled`)
- [x] Deployed with `tls.enabled: true` + `sentinelTls.insecureSkipVerify: true` against ElastiCache Valkey with transit encryption — verifying in staging

Yes — operators can now configure TLS for Redis connections with granular control over master and sentinel TLS settings. Default behavior is unchanged (no TLS when fields are omitted).

## Large PR Justification

- Generated code that cannot be split
- Large refactoring that must be atomic


pkg/authserver/storage/redis.go
```diff
@@ -126,7 +170,39 @@ func NewRedisStorage(ctx context.Context, cfg RedisConfig) (*RedisStorage, error
...
+	}
+
+	// Configure TLS for sentinel connections via custom dialer.
+	// go-redis applies TLSConfig to both sentinel and master connections,
+	// so we use a custom dialer to apply different TLS configs per target.
+	sentinelTLS := cfg.SentinelTLS
+	if sentinelTLS == nil {
+		sentinelTLS = cfg.TLS // fall back to master TLS config
+	}
+	if sentinelTLS != nil && sentinelTLS.Enabled && opts.TLSConfig != nil {
```
jhrozek: The condition requires both master and sentinel TLS to be enabled. Two scenarios break:

Master TLS on + Sentinel TLS explicitly off (sentinelTls: {enabled: false}): sentinelTLS.Enabled is false → no dialer → opts.TLSConfig stays set → go-redis applies master TLS to sentinel connections too. 

and:
Sentinel TLS on + Master TLS off: opts.TLSConfig is nil → condition fails → no dialer → no TLS anywhere, despite sentinel TLS being requested?

Should the code instead try to install the custom dialer whenever master and sentinel need different TLS treatment — including using net.Dialer.DialContext (plaintext) for the side that has TLS disabled.

pkg/authserver/storage/redis.go
```diff
@@ -99,6 +126,23 @@ type storedSession struct {
...
+	if cfg == nil {
+		return nil
+	}
+	tc := &tls.Config{
+		MinVersion: tls.VersionTLS12,
+		InsecureSkipVerify: cfg.InsecureSkipVerify, //nolint:gosec // G402: configurable per-deployment
+	}
+	if len(cfg.CACert) > 0 {
+		pool := x509.NewCertPool()
+		pool.AppendCertsFromPEM(cfg.CACert)
```
jhrozek: AppendCertsFromPEM return value is discarded

If the PEM data is malformed (corrupt, DER-encoded, double-base64), AppendCertsFromPEM returns false but the empty pool is still assigned to RootCAs. This replaces system root CAs with an empty pool, causing every TLS handshake to fail with an opaque "certificate signed by unknown authority" error 

Consider checking the return value and either returning an error (changing signature to (*tls.Config, error)) or at minimum logging a warning and skipping the empty pool assignment so system CAs are used as fallback.

pkg/authserver/runner/embeddedauthserver.go
```diff
@@ -475,9 +475,34 @@ func convertRedisRunConfig(rc *storage.RedisRunConfig) (*storage.RedisConfig, er
...
+	}
+	cfg := &storage.RedisTLSConfig{
+		Enabled:            rc.Enabled,
+		InsecureSkipVerify: rc.InsecureSkipVerify,
+	}
+	if rc.CACertFile != "" {
+		// #nosec G304 - file path is from configuration, not user input
+		data, err := os.ReadFile(rc.CACertFile)
+		if err != nil {
+			slog.Warn("Failed to read Redis CA cert file, using system CAs",
```
jhrozek: should this be a hard error?

pkg/authserver/storage/redis.go
```diff
@@ -126,7 +170,39 @@ func NewRedisStorage(ctx context.Context, cfg RedisConfig) (*RedisStorage, error
...
+	sentinelTLS := cfg.SentinelTLS
+	if sentinelTLS == nil {
+		sentinelTLS = cfg.TLS // fall back to master TLS config
+	}
+	if sentinelTLS != nil && sentinelTLS.Enabled && opts.TLSConfig != nil {
+		masterTLS := opts.TLSConfig
+		sentinelTLSCfg := buildTLSConfig(sentinelTLS)
+		// Clear the global TLSConfig — we handle TLS per-connection in the dialer
+		opts.TLSConfig = nil
+		opts.Dialer = func(ctx context.Context, network, addr string) (net.Conn, error) {
```
jhrozek: could this be a named function so that the code could be tested?

cmd/thv-operator/pkg/controllerutil/authserver.go
```diff
@@ -161,6 +173,57 @@ func GenerateAuthServerVolumes(
 		})
 	}
 
+	// Generate volumes for Redis TLS CA certificates
+	if authConfig.Storage != nil && authConfig.Storage.Redis != nil {
```
jhrozek, line 177: I keep forgetting what the test coverage is for this code, but it would be nice to add a test

cmd/thv-operator/pkg/controllerutil/authserver.go
```diff
@@ -161,6 +173,57 @@ func GenerateAuthServerVolumes(
...
+							Key:  ref.Key,
+							Path: RedisTLSCACertFileName,
+						}},
+						DefaultMode: k8sptr.To(int32(0400)),
+					},
+				},
+			})
+			volumeMounts = append(volumeMounts, corev1.VolumeMount{
+				Name:      volumeName,
+				MountPath: fmt.Sprintf("%s/%s", RedisTLSCACertMountPath, RedisTLSCACertFileName),
```
jhrozek, line 197: the thing that prompted me to ask for tests was that the volume seems to be mounted even with redis.TLS.Enabled=False?

jhrozek (commented):
great work, CI seems to be unhappy (unused ctx..) plus I've added some nits inline.

Thanks for the ongoing contributions to toolhive!

pkg/authserver/storage/redis.go
aron-muon: Good catch, both scenarios were broken. Refactored the TLS setup into a `configureTLSDialer()` helper that installs the custom dialer whenever *either* side needs TLS, not just when both do. The dialer uses `net.Dialer.Dial()` (plaintext) for whichever side has TLS disabled. Sentinel falls back to master config when nil, but an explicit `enabled: false` is respected. Added `TestConfigureTLSDialer` covering all combinations (both on, master only, sentinel only, neither, master on + sentinel explicitly off, invalid certs).

pkg/authserver/storage/redis.go
aron-muon: Fixed, `buildTLSConfig` now returns `(*tls.Config, error)`. If `AppendCertsFromPEM` returns false, we return `fmt.Errorf("failed to parse CA certificate PEM data")` instead of silently assigning an empty pool. Updated the test to expect an error for invalid PEM.

pkg/authserver/runner/embeddedauthserver.go
aron-muon: Yes - changed `convertRedisTLSRunConfig` to return `(*storage.RedisTLSConfig, error)`

pkg/authserver/storage/redis.go
aron-muon: Done, extracted the closure into `newTLSDialer(masterTLS, sentinelTLS, sentinelAddrs, timeout)` which returns the dialer function. Also extracted the orchestration logic into `configureTLSDialer()`. Both are tested independently

cmd/thv-operator/pkg/controllerutil/authserver.go
aron-muon: Added `TestGenerateAuthServerVolumes_RedisTLS` with 6 cases: TLS enabled with CA cert (creates volume), TLS disabled with CA cert (no volume), TLS enabled without CA cert (no volume), both master+sentinel with CA certs (2 volumes), sentinel only (1 volume), nil storage (no volumes). Also verifies 0400 permissions and read-only mounts

cmd/thv-operator/pkg/controllerutil/authserver.go
aron-muon: Nice catch, added `redis.TLS.Enabled` and `redis.SentinelTLS.Enabled` checks to the volume generation conditions. The "TLS disabled with CA cert does NOT create volume" test case explicitly covers this

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
```diff
@@ -481,6 +481,16 @@ type RedisStorageConfig struct {
 	// +kubebuilder:default="3s"
 	// +optional
 	WriteTimeout string `json:"writeTimeout,omitempty"`
+
+	// TLS configures TLS for connections to the Redis/Valkey master.
+	// Required when the cluster has transit encryption enabled.
+	// +optional
+	TLS *RedisTLSConfig `json:"tls,omitempty"`
```
jhrozek, line 488: sorry, should have pointed this out earlier AND our CI should have caught that (I don't know why it didn't that must be a CI bug), but I think you need to run `task operator-generate` to regenerate the deepcopy files.

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
```diff
@@ -521,6 +531,23 @@ type SentinelServiceRef struct {
 	Port int32 `json:"port,omitempty"`
 }
 
+// RedisTLSConfig configures TLS for Redis connections.
+type RedisTLSConfig struct {
+	// Enabled activates TLS for this connection type.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
```
jhrozek: Design question: why did you add the Enabled field explicitly, could we just enable TLS based on the presence of this struct?

jhrozek (commented):
looks like we need to regenerate the deepcopies before merging, sorry, I should have checked this earlier and our CI should have found it on its own.

On the bright side, I did test this end-to-end on a kind cluster. Set up a Redis master + 3-node Sentinel cluster with TLS enabled  (self-signed CA, server certs with proper SANs). I verified the full storage round-trip by registering an OAuth dynamic client through the proxy and confirming the key (thv:auth:{...}:client:<id>) was persisted in Redis.

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
aron-muon: No worries, reran `task operator-generate`, `task operator-manifests`, and `task crdref-gen`. The missing `DeepCopyInto`/`DeepCopy` methods for `RedisTLSConfig` are now generated.

cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go
aron-muon: Good point, I removed `Enabled` entirely. Presence of the struct now activates TLS, `nil` means plaintext. To support all combinations cleanly (including "master TLS + sentinel plaintext"), I also removed the fallback from `sentinelTls` to `tls`. This means users who want the same TLS config for both must specify it on both fields, but the semantics are explicit and unsurprising. The trade-off felt worth it for the simpler API.

jhrozek (approved):
thanks for addressing the review feedback so quickly!

PR #4071, opened 2026-03-10
Fix CI build failures on main

JAORMX:
## Summary

- The main build broke after e783bf54 ("Automate HMAC secret management") due to a missing package prefix on `MetadataKeyTokenSalt` in `sessionv2_integration_test.go`, causing a typecheck lint failure that cascaded into test compilation failures for dependent packages.
- `TestMiddleware_ContextTimeoutHandling` hard-coded a 15-second wait for the `discoveryTimeout` constant, making it slow and fragile. Made the timeout injectable via `WithDiscoveryTimeout` middleware option so the test uses 100ms instead.

## Type of change

- [x] Bug fix

## Test plan

No

### DIFF
--- a/pkg/vmcp/discovery/middleware.go
+++ b/pkg/vmcp/discovery/middleware.go
@@ -45,6 +45,7 @@ const (
 // middlewareConfig holds optional configuration for Middleware.
 type middlewareConfig struct {
 	sessionScopedRouting bool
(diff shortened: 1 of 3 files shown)


amirejaz:
@copilot review this PR if test time can further be reduced

PR #4073, opened 2026-03-10
Fix CI failures from typecheck error and flaky keyctl test

aponcedeleonch:
## Summary

CI on main is broken after the HMAC secret management merge (e783bf54). This PR fixes two issues causing the lint and test jobs to fail:

1. **Typecheck error**: `sessionv2_integration_test.go` referenced `MetadataKeyTokenSalt` without the `sessiontypes.` package qualifier, failing golangci-lint's typecheck and preventing the session package tests from compiling. The compilation failure also cascaded into a test reporter panic (`BUG: Empty package name encountered`), causing false-positive test failures.

2. **Flaky keyctl test**: After `DeleteAll` unlinks and revokes a kernel keyring key, `KeyctlSearch` may still briefly return the key. The `Get` method now returns `ErrNotFound` for `EKEYREVOKED`/`EKEYEXPIRED` read errors instead of a generic failure, making the delete-then-get pattern reliable.

## Type of change

## Test plan

- [x] Verified session package compiles with `go vet ./pkg/vmcp/session/...`
- [x] Verified session tests compile with `go test -run=^$ -count=0 ./pkg/vmcp/session/...`
- [ ] CI lint job passes
- [ ] CI test job passes

### DIFF
--- a/pkg/secrets/keyring/keyctl_linux.go
+++ b/pkg/secrets/keyring/keyctl_linux.go
@@ -6,6 +6,7 @@
 package keyring
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


aponcedeleonch:
Closing in favor of: https://github.com/stacklok/toolhive/pull/4071

PR #4078, opened 2026-03-10
Fix flaky E2E tests caused by port allocation race condition

aponcedeleonch:
## Summary

- E2E Tests Core jobs are flaky on main, failing across consecutive CI runs with HTTP 500 on workload creation during bulk operations, followed by suite timeout cascades
- Root cause: port allocation in `pkg/networking/port.go` has a TOCTOU race — `IsAvailable()` opens/closes listeners, creating a window where concurrent goroutines allocate the same port
- Fix adds process-level mutex and recently-allocated port tracking to prevent concurrent goroutines from receiving the same port
- Port allocation errors now surface as HTTP 503 (Service Unavailable) instead of generic 500, and bulk E2E tests use retry + NodeTimeout for resilience

## Type of change

- [x] Bug fix

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/networking/port.go` | Add `sync.Mutex` + `recentlyAllocated` map with 30s TTL to prevent TOCTOU races in `FindAvailable()` and `FindOrUsePort()` |
| `pkg/runner/config.go` | Add `ErrPortUnavailable` sentinel error; wrap all port allocation failures with it |
| `pkg/api/v1/workload_service.go` | Map `ErrPortUnavailable` to HTTP 503; wrap `SaveState`/`RunWorkloadDetached` errors with explicit `httperr` codes |
| `test/e2e/api_workloads_test.go` | Add response body logging on unexpected status; add `createWorkloadWithRetry` helper (3 attempts, 2s backoff on 5xx) |
| `test/e2e/api_workload_lifecycle_test.go` | Use `createWorkloadWithRetry` in bulk stop/restart/delete tests; add `NodeTimeout(5m)` to prevent cascade failures |

Port allocation errors during workload creation now return HTTP 503 (Service Unavailable) instead of a generic 500 (Internal Server Error), giving clients a clearer signal to retry.

- The `recentlyAllocated` map uses a 30s TTL with lazy purging — entries are cleaned up on the next `FindAvailable()`/`FindOrUsePort()` call, not via a background goroutine
- `FindOrUsePort` inlines the fallback port search rather than calling `FindAvailable()` to avoid double-locking the mutex
- The retry helper is intentionally only used in bulk operation tests where the race was observed; single-workload tests keep strict assertions

### DIFF
--- a/pkg/networking/port.go
+++ b/pkg/networking/port.go
@@ -11,8 +11,20 @@ import (
 	"log/slog"
 	"math/big"
 	"net"
+	"sync"
+	"time"
 
 	gopsutilnet "github.com/shirou/gopsutil/v4/net"
+)
+
+var (
+	// portMu protects recentlyAllocated from concurrent access
+	portMu sync.Mutex
+	// recentlyAllocated tracks ports returned by FindAvailable/FindOrUsePort
+	// to prevent TOCTOU races when multiple goroutines allocate ports concurrently.
+	recentlyAllocated = map[int]time.Time{}
+	// portReservationTTL is how long a port stays reserved in recentlyAllocated
+	portReservationTTL = 30 * time.Second
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/networking/port.go
```diff
@@ -11,10 +11,22 @@ import (
...
+	"time"
 
 	gopsutilnet "github.com/shirou/gopsutil/v4/net"
 )
 
+var (
+	// portMu protects recentlyAllocated from concurrent access
+	portMu sync.Mutex
+	// recentlyAllocated tracks ports returned by FindAvailable/FindOrUsePort
+	// to prevent TOCTOU races when multiple goroutines allocate ports concurrently.
```
JAORMX, line 24: Why not use  sync map?

aponcedeleonch:
I thought port allocation was the issue for the flaky tests but turns out it wasn't or at least now the e2e tests seem to be more stable without getting this merged. Closing this PR for now

PR #4086, opened 2026-03-11
fix(auth): use inactivity TTL for upstream refresh-token entries

### LINKED ISSUE #3914: Fix upstream token storage TTL for refresh-token-bearing entries
## Summary

`StoreUpstreamTokens` sets the storage entry's TTL to the upstream access token's `ExpiresAt` (~1h). This garbage-collects the refresh token along with the expired access token, even though the refresh token may be valid for 30+ days.

## What

Change `StoreUpstreamTokens` TTL logic:
- If the entry has a refresh token: TTL = `now + 2h` (sliding inactivity window)
- If no refresh token: TTL = `ExpiresAt` (unchanged)

Add `DefaultUpstreamInactivityTimeout` constant.

## Why

This is a prerequisite for upstream token refresh — without it, there's nothing left to refresh by the time anyone tries.

SergioChan:
## Summary

- Fixes upstream token storage TTL selection so refresh-token-bearing entries use a **2h sliding inactivity window** instead of being tied to access token expiry semantics.
- Preserves existing no-refresh behavior by using the access token `ExpiresAt` when present (or default access-token TTL fallback when absent).
- Unifies TTL calculation in a shared helper used by both memory and Redis storage paths.
- Adds focused unit tests for both storage backends to verify refresh vs non-refresh TTL behavior and updates cleanup expectations.

Fixes #3914

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- Validation performed:
  - `git diff --check`
  - Attempted: `go test ./pkg/authserver/storage -run "Test(MemoryStorage|RedisStorage)_UpstreamTokens" -count=1`
- Why full automated tests were not run here:
  - The environment cannot provide the required Go toolchain (`go1.26`), so `go test` fails before package compilation.

## Changes

| File | Change |
|------|--------|
| `pkg/authserver/storage/config.go` | Added `DefaultUpstreamInactivityTimeout` (2h). |
| `pkg/authserver/storage/upstream_tokens_ttl.go` | Added shared TTL/expiry calculation helpers for upstream token entries. |
| `pkg/authserver/storage/memory.go` | Switched upstream token expiry assignment to shared helper; updated expiry semantics comment. |
| `pkg/authserver/storage/redis.go` | Switched Redis TTL derivation to shared helper. |
| `pkg/authserver/storage/memory_test.go` | Added TTL behavior tests and updated cleanup expectation for no-refresh entries. |
| `pkg/authserver/storage/redis_test.go` | Added TTL behavior tests and adjusted expired-token comment for refresh-token inactivity behavior. |

Yes. Upstream auth sessions that include a refresh token now expire based on inactivity (~2h sliding window) instead of long-lived storage tied to refresh TTL extension. Entries without a refresh token continue to follow access-token expiry behavior.

### DIFF
--- a/pkg/authserver/storage/config.go
+++ b/pkg/authserver/storage/config.go
@@ -28,6 +28,10 @@ const (
 	// DefaultRefreshTokenTTL is the default TTL for refresh tokens when not extractable from session.
 	DefaultRefreshTokenTTL = 30 * 24 * time.Hour // 30 days
 
+	// DefaultUpstreamInactivityTimeout is the sliding inactivity window for upstream
+	// token entries that include a refresh token.
+	DefaultUpstreamInactivityTimeout = 2 * time.Hour
+
 	// DefaultAuthCodeTTL is the default TTL for authorization codes (RFC 6749 recommendation).
 	DefaultAuthCodeTTL = 10 * time.Minute
 
--- a/pkg/authserver/storage/memory.go
+++ b/pkg/authserver/storage/memory.go
@@ -693,14 +693,7 @@ func (s *MemoryStorage) StoreUpstreamTokens(_ context.Context, sessionID string,
 	defer s.mu.Unlock()
 
 	now := time.Now()
-	// Add DefaultRefreshTokenTTL beyond access token expiry so the refresh token
-	// survives in storage for transparent token refresh by the middleware.
-	var expiresAt time.Time
-	if tokens != nil && !tokens.ExpiresAt.IsZero() {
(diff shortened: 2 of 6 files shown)


jhrozek:
@SergioChan why is this approach better? This approach means that if the token is unused during the sliding window activity then it's lost and the user must re-auth. 

Could you explain what issues did you face with the current refresh token storage?

PR #4090, opened 2026-03-11
Preserve tool annotations and output schema through vmcp

JAORMX:
## Summary

- MCP tool annotations (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`, `Title`) were silently dropped during vmcp aggregation, blocking downstream authorization profiles. This adds the `ToolAnnotations` vmcp domain type and propagates it through the full pipeline.
- `OutputSchema` existed on `vmcp.Tool` but was never populated from backend MCP servers. This populates it using the same JSON round-trip pattern as `InputSchema`.
- Both fields now flow end-to-end: backend query → tool adapter → all three conflict resolvers (prefix/priority/manual) → merge → SDK adapter output.

## Type of change

- [x] New feature

## Test plan

22 new test cases across 4 test files:
- `pkg/vmcp/conversion/content_test.go` — conversion functions (15 cases including round-trip)
- `pkg/vmcp/aggregator/tool_adapter_annotations_test.go` — preservation through overrides (3 cases)
- `pkg/vmcp/server/adapter/capability_adapter_annotations_test.go` — SDK output (4 cases)

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/types.go` | Add `ToolAnnotations` struct and `Annotations` field to `Tool` |
| `pkg/vmcp/conversion/content.go` | Add `ConvertToolAnnotations`, `ConvertToolOutputSchema`, `ToMCPToolAnnotations` |
| `pkg/vmcp/client/client.go` | Populate `Annotations` + `OutputSchema` from backend |
| `pkg/vmcp/session/internal/backend/mcp_session.go` | Same for session-based client path |
| `pkg/vmcp/aggregator/aggregator.go` | Add fields to `ResolvedTool` |
| `pkg/vmcp/aggregator/prefix_resolver.go` | Propagate through prefix resolver |
| `pkg/vmcp/aggregator/priority_resolver.go` | Propagate through all 3 code paths |
| `pkg/vmcp/aggregator/manual_resolver.go` | Propagate through manual resolver |
| `pkg/vmcp/aggregator/default_aggregator.go` | Propagate through fallback + merge |
| `pkg/vmcp/aggregator/tool_adapter.go` | Preserve through filtering/renaming |
| `pkg/vmcp/server/adapter/capability_adapter.go` | Emit `Annotations` and `RawOutputSchema` to SDK types |

No direct user-facing change. Backend MCP tool annotations and output schemas are now preserved through vmcp aggregation instead of being silently dropped. This enables downstream consumers (like authorization profiles) to use these fields.

- **ACL pattern preserved**: `vmcp.ToolAnnotations` is a vmcp-owned type (not `mcp.ToolAnnotation` directly) to maintain the Anti-Corruption Layer — mcp-go is only used at ingestion and SDK adapter boundaries.
- **PR 2 follow-up**: `Icons`, `Execution`, Resource `Annotations`, and Prompt `Icons` are deferred to a follow-up PR per the split strategy.
- **`ConvertToolOutputSchema` nil check**: Uses `schema.Type == ""` rather than byte-length check because a zero-valued `ToolOutputSchema` marshals to `{"type":"","properties":{},"required":[]}` (non-empty JSON but semantically empty).

### DIFF
--- a/pkg/vmcp/aggregator/aggregator.go
+++ b/pkg/vmcp/aggregator/aggregator.go
@@ -118,6 +118,12 @@ type ResolvedTool struct {
 	// InputSchema is the JSON Schema for parameters.
 	InputSchema map[string]any
 
+	// OutputSchema is the JSON Schema for tool output (optional).
+	OutputSchema map[string]any
+
+	// Annotations describes behavioral hints for the tool (optional).
+	Annotations *vmcp.ToolAnnotations
+
 	// BackendID identifies the backend providing this tool.
 	BackendID string
 
--- a/pkg/vmcp/aggregator/default_aggregator.go
+++ b/pkg/vmcp/aggregator/default_aggregator.go
@@ -244,6 +244,8 @@ func (a *defaultAggregator) ResolveConflicts(
 					OriginalName: tool.Name,
 					Description:  tool.Description,
 					InputSchema:  tool.InputSchema,
+					OutputSchema: tool.OutputSchema,
+					Annotations:  tool.Annotations,
 					BackendID:    backendID,
 				}
 			}
@@ -321,10 +323,12 @@ func (a *defaultAggregator) MergeCapabilities(
 
 		if shouldAdvertise {
 			tools = append(tools, vmcp.Tool{
-				Name:        resolvedTool.ResolvedName,
-				Description: resolvedTool.Description,
-				InputSchema: resolvedTool.InputSchema,
-				BackendID:   resolvedTool.BackendID,
+				Name:         resolvedTool.ResolvedName,
+				Description:  resolvedTool.Description,
+				InputSchema:  resolvedTool.InputSchema,
+				OutputSchema: resolvedTool.OutputSchema,
+				Annotations:  resolvedTool.Annotations,
+				BackendID:    resolvedTool.BackendID,
 			})
 		}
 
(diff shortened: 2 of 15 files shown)


pkg/vmcp/aggregator/aggregator.go
```diff
@@ -118,6 +118,12 @@ type ResolvedTool struct {
...
 	InputSchema map[string]any
 
+	// OutputSchema is the JSON Schema for tool output (optional).
+	OutputSchema map[string]any
+
+	// Annotations describes behavioral hints for the tool (optional).
+	Annotations *vmcp.ToolAnnotations
+
 	// BackendID identifies the backend providing this tool.
 	BackendID string
```
aponcedeleonch, line 128: Instead of keep adding fields to this struct to mimic the original tool can we somehow leverage an untampered version of the `mcp.Tool` in here. For example, this struct could be
```go
import (
	"github.com/mark3labs/mcp-go/mcp"
)

type ResolvedTool struct {
    // ResolvedName is the final name exposed to clients (after conflict resolution).
	ResolvedName string 
    // BackendID identifies the backend providing this tool.
	BackendID string
    // Original tool
	Tool mcp.Tool
}
```

and then we could have some logic to replace the name in `Tool` with `ResolvedName` before returning it

JAORMX: Right! the idea was to not rely on mcp-go for this since it's a DTO.

PR #4091, opened 2026-03-11
Enable session management v2 by default

### LINKED ISSUE #3872: [vMCP] Remove old code path after Phase 2 validation (Phase 3)
Depends on #3866, #3867, #3871

After `sessionManagementV2` has been validated in CI and run in production without rollbacks, flip the default and delete the old implementation to eliminate the maintenance burden of two code paths.

**Steps**:
1. Enable `sessionManagementV2` in CI and validate all tests pass
2. Change default from `false` to `true`, release, and monitor the bake period (session creation latency, tool call latency, connection counts, error rates)
3. After a successful bake period with no rollbacks, delete all old code

**Files to delete/modify**:
- `pkg/vmcp/discovery/middleware.go` — delete
- `pkg/vmcp/client/client.go` — remove `httpBackendClient`
- `pkg/vmcp/server/server.go`, `hooks.go` — remove feature flag conditionals and old code path
- Old `VMCPSession` implementation files — delete

### Acceptance Criteria

- [ ] `sessionManagementV2` defaults to `true`
- [ ] The feature flag and all conditional logic are removed
- [ ] `httpBackendClient` is deleted
- [ ] `pkg/vmcp/discovery/middleware.go` is deleted
- [ ] Old `VMCPSession` implementation files are deleted
- [ ] All existing tests pass with the new code path as the only path
- [ ] E2E tests verify: full vMCP workflow with multiple backends, backend state preservation across tool calls, session expiration closes all backend clients
- [ ] High-throughput tests confirm connection count and heap allocation remain stable under sustained load (no unbounded growth)

---

yrobla:
## Summary

Session management v2 is now the default for all vMCP deployments.

**Why:** Session management v2 provides per-session backend lifecycle with HMAC token binding and session-scoped routing. It is now stable enough to be on by default.

**What changed:**
- `SessionManagementV2` in `OperationalConfig` changed from `bool` to `*bool` so that an unset field (`nil`) can be distinguished from an explicit opt-out (`false`)
- Default value is now `true` — all new and existing vMCP deployments that omit the flag will use session management v2
- Explicit opt-out remains possible by setting `sessionManagementV2: false` in the config YAML or CRD spec
- HMAC secret is created by default by the operator; skipped only when `sessionManagementV2` is explicitly `false`
- CRD schema is unchanged (`type: boolean`); only the description text was updated

Fixes #3872

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/config/config.go` | `SessionManagementV2 bool` → `*bool` |
| `pkg/vmcp/config/defaults.go` | Default `*true`; save/restore around mergo to preserve explicit `*false` |
| `pkg/vmcp/config/zz_generated.deepcopy.go` | Regenerated to deep-copy `*bool` field |
| `cmd/vmcp/app/commands.go` | Dereference `*bool` with nil-means-true semantics |
| `cmd/thv-operator/controllers/virtualmcpserver_controller.go` | Skip HMAC creation only when `SessionManagementV2 == *false` |
| `cmd/thv-operator/controllers/virtualmcpserver_deployment.go` | Skip HMAC env var only when `SessionManagementV2 == *false` |
| `deploy/charts/operator-crds/` | CRD description text updated (schema type unchanged) |
| `test/e2e/.../virtualmcp_session_management_v2_test.go` | Added Context 4: opt-out via explicit `false`; updated Context 2 to test default-true behaviour |

Yes. vMCP deployments that previously omitted `sessionManagementV2` will now use session management v2 automatically. Users who need the old (v1) behaviour must explicitly set `sessionManagementV2: false` in their config.

- `*bool` is required because mergo treats a `bool` zero-value (`false`) as "unset" and would overwrite an explicit opt-out with the default `true`. `EnsureOperationalDefaults` saves the original pointer and value before merging and restores both afterwards to preserve pointer identity.
- The operator reads the raw CRD spec before `EnsureOperationalDefaults` runs, so the controller and deployment code implement nil-means-true directly.
- The CRD YAML schema (`type: boolean`) is unchanged; no API version bump is needed.

## Large PR Justification

This is an isolated functionality, enabling a new session management flow. It comes with complete tests and fixes some problems dected when enabling it . Cannot be split.

### DIFF
--- a/cmd/vmcp/app/commands.go
+++ b/cmd/vmcp/app/commands.go
@@ -297,7 +297,10 @@ func discoverBackends(
 //   - Otherwise: logs warning and creates factory with default insecure secret
 //
 // Returns an error only when running in Kubernetes without a secret (fail-fast for production).
-func createSessionFactory(outgoingRegistry vmcpauth.OutgoingAuthRegistry) (vmcpsession.MultiSessionFactory, error) {
+func createSessionFactory(
+	outgoingRegistry vmcpauth.OutgoingAuthRegistry,
+	agg aggregator.Aggregator,
+) (vmcpsession.MultiSessionFactory, error) {
 	const (
 		envKey                  = "VMCP_SESSION_HMAC_SECRET"
 		minRecommendedSecretLen = 32
@@ -319,6 +322,7 @@ func createSessionFactory(outgoingRegistry vmcpauth.OutgoingAuthRegistry) (vmcps
 		return vmcpsession.NewSessionFactory(
 			outgoingRegistry,
 			vmcpsession.WithHMACSecret([]byte(hmacSecret)),
+			vmcpsession.WithAggregator(agg),
 		), nil
 	}
 
@@ -332,7 +336,7 @@ func createSessionFactory(outgoingRegistry vmcpauth.OutgoingAuthRegistry) (vmcps
 
 	// Development mode: use default insecure secret with warning
 	slog.Warn("VMCP_SESSION_HMAC_SECRET not set - using default insecure secret (NOT recommended for production)")
-	return vmcpsession.NewSessionFactory(outgoingRegistry), nil
+	return vmcpsession.NewSessionFactory(outgoingRegistry, vmcpsession.WithAggregator(agg)), nil
 }
 
 // runServe implements the serve command logic
@@ -527,7 +531,7 @@ func runServe(cmd *cobra.Command, _ []string) error {
 	sessionManagementV2 := cfg.Operational.SessionManagementV2 == nil || *cfg.Operational.SessionManagementV2
 	var sessionFactory vmcpsession.MultiSessionFactory
 	if sessionManagementV2 {
-		sessionFactory, err = createSessionFactory(outgoingRegistry)
+		sessionFactory, err = createSessionFactory(outgoingRegistry, agg)
 		if err != nil {
 			return err
 		}
--- a/deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_virtualmcpservers.yaml
+++ b/deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_virtualmcpservers.yaml
@@ -692,6 +692,7 @@ spec:
                         - debug
                         type: string
                       sessionManagementV2:
+                        default: true
                         description: |-
                           SessionManagementV2 enables session-scoped backend client lifecycle.
                           When true, vMCP creates real backend connections per session via MultiSessionFactory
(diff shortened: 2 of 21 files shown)


cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -681,8 +681,10 @@ func (r *VirtualMCPServerReconciler) ensureHMACSecret(
 ) error {
 	ctxLogger := log.FromContext(ctx)
 
-	// Only ensure HMAC secret if Session Management V2 is enabled
-	if vmcp.Spec.Config.Operational == nil || !vmcp.Spec.Config.Operational.SessionManagementV2 {
+	// Skip HMAC secret creation only when SessionManagementV2 is explicitly set to false.
+	// nil means "unset" which defaults to true (v2 enabled).
+	op := vmcp.Spec.Config.Operational
+	if op != nil && op.SessionManagementV2 != nil && !*op.SessionManagementV2 {
```
jerm-dro, line 687: Could you avoid doing this by using kubebuilder annotations to make this default to true:

https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/config/config.go#L449


That would make the diff much smaller.

jerm-dro (dismissed):

cmd/thv-operator/controllers/virtualmcpserver_controller.go
yrobla: it will help to make it clear , but does not help in the case of operational config not being populated, it should also default to true then, so we need the pointer...

cmd/vmcp/app/commands.go
```diff
@@ -332,7 +336,7 @@ func createSessionFactory(outgoingRegistry vmcpauth.OutgoingAuthRegistry) (vmcps
 
 	// Development mode: use default insecure secret with warning
 	slog.Warn("VMCP_SESSION_HMAC_SECRET not set - using default insecure secret (NOT recommended for production)")
-	return vmcpsession.NewSessionFactory(outgoingRegistry), nil
+	return vmcpsession.NewSessionFactory(outgoingRegistry, vmcpsession.WithAggregator(agg)), nil
```
jerm-dro, line 339: Non-blocking: is aggregator a required parameter? Should `NewSessionFactory` not pass it as an option then?

jerm-dro (approved):

cmd/vmcp/app/commands.go
yrobla: will do in a. follow up

PR #4097, opened 2026-03-11
test(authserver): deflake cleanup loop periodic assertion

### LINKED ISSUE #4096: Flaky test: TestMemoryStorage_CleanupLoop/cleanup_runs_periodically
## Summary

`TestMemoryStorage_CleanupLoop/cleanup_runs_periodically` in `pkg/authserver/storage/memory_test.go` is flaky due to timing sensitivity.

## Details

The test creates a storage with a 50ms cleanup interval, inserts an expired auth code, then sleeps 100ms and asserts the code was cleaned up:

```go
storage := NewMemoryStorage(WithCleanupInterval(50 * time.Millisecond))
// ... insert expired session ...
time.Sleep(100 * time.Millisecond)
assert.Equal(t, 0, storage.Stats().AuthCodes) // fails intermittently
```

Under load (e.g., `-race` with many parallel tests), the cleanup goroutine may not have fired within the 100ms sleep window, causing the assertion to fail with `expected: 0, actual: 1`.

**Location**: `pkg/authserver/storage/memory_test.go:716-729`

## Suggested fix

Replace the fixed sleep with a polling loop using `require.Eventually` or `assert.Eventually`:

```go
require.Eventually(t, func() bool {
    return storage.Stats().AuthCodes == 0
}, 2*time.Second, 25*time.Millisecond, "expired auth code should be cleaned up")
```

This is deterministic regardless of system load while still testing the same cleanup behavior.

SergioChan:
## Summary

- `TestMemoryStorage_CleanupLoop/cleanup runs periodically` was flaky because it relied on a fixed `time.Sleep(100ms)` and could miss cleanup execution under load.
- Replaced the fixed sleep with `require.Eventually(..., 2*time.Second, 25*time.Millisecond)` so the assertion waits deterministically until the cleanup loop removes expired auth codes.

Fixes #4096

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Manual testing:
- Verified the updated test logic now polls for cleanup completion instead of assuming cleanup has run at a fixed time boundary.
- Ran `git diff --check` to confirm no whitespace or patch formatting issues.
- Attempted to run targeted unit tests, but local Go attempted to download toolchain `go1.26` and failed with `toolchain not available` in this environment.

No.

### DIFF
--- a/pkg/authserver/storage/memory_test.go
+++ b/pkg/authserver/storage/memory_test.go
@@ -724,8 +724,9 @@ func TestMemoryStorage_CleanupLoop(t *testing.T) {
 		require.NoError(t, storage.CreateAuthorizeCodeSession(ctx, "expired", expiredRequest))
 		assert.Equal(t, 1, storage.Stats().AuthCodes)
 
-		time.Sleep(100 * time.Millisecond)
-		assert.Equal(t, 0, storage.Stats().AuthCodes)
+		require.Eventually(t, func() bool {
+			return storage.Stats().AuthCodes == 0
+		}, 2*time.Second, 25*time.Millisecond, "expired auth code should be cleaned up")
 	})
 
 	t.Run("close stops cleanup goroutine", func(t *testing.T) {


JAORMX:
This was already assigned.

PR #4108, opened 2026-03-11
Make transparent proxy health check parameters configurable

### LINKED ISSUE #4084: Make transparent proxy health check parameters configurable
## Problem

When a transient network disruption occurs between the transparent proxy and a container, the proxy's health check fails consecutively and triggers proxy shutdown. This invalidates the client's MCP session ("Session not found"), requiring a full client restart to recover.

The total window from first failure to shutdown is ~40 seconds with current defaults (3 ticks at 10s intervals + 5s retry delays). Any network blip lasting longer than this kills the proxy, even though the MCP server process inside the container is perfectly healthy.

## Root cause

The transparent proxy health check parameters are mostly hardcoded:

| Parameter | Default | Configurable? |
|---|---|---|
| Health check interval | 10s | Yes — `TOOLHIVE_HEALTH_CHECK_INTERVAL` |
| Ping timeout | 5s | No — `DefaultPingerTimeout` in `pinger.go` |
| Retry delay between failures | 5s | No — `DefaultHealthCheckRetryDelay` |
| Consecutive failures before shutdown | 3 | No — `healthCheckRetryCount` constant |

The `with*` option functions (`withHealthCheckPingTimeout()`, `withHealthCheckRetryDelay()`) already exist but are unexported and only used in tests.

## Proposed fix

1. Expose the hardcoded parameters as environment variables, following the existing pattern of `TOOLHIVE_HEALTH_CHECK_INTERVAL`:

| Env Var | Controls | Default |
|---|---|---|
| `TOOLHIVE_HEALTH_CHECK_PING_TIMEOUT` | Per-attempt ping timeout | 5s |
| `TOOLHIVE_HEALTH_CHECK_RETRY_DELAY` | Delay between consecutive failures | 5s |
| `TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD` | Consecutive failures before shutdown | 5 |

2. Raise the default failure threshold from 3 to 5 to improve resilience out of the box. This extends the tolerance window from ~40s to ~60s, reducing false positives from transient network disruptions without significantly delaying detection of genuinely dead servers.

### Why not just increase the interval?

Increasing `TOOLHIVE_HEALTH_CHECK_INTERVAL` delays detection but doesn't increase resilience — the failure count is still fixed. A user who wants to tolerate longer network blips would need a very large interval, which means legitimate failures also take a long time to detect. Configuring the failure threshold separately allows both quick detection (short interval) and network tolerance (higher threshold).

### Scope

This proposal is only for the transparent proxy (SSE and streamable HTTP transports). The vMCP server has a separate circuit breaker pattern.

## Related

- ToolHive issues: #3783, #1357
- ToolHive PRs: #4063

gkatz2:
## Summary

- The transparent proxy health check has mostly hardcoded parameters (ping timeout, retry delay, failure threshold), making it impossible to tune resilience for environments with transient network disruptions. When the proxy kills itself due to false positives, the client's MCP session is invalidated, requiring a full client restart to recover.
- Add three new environment variables (`TOOLHIVE_HEALTH_CHECK_PING_TIMEOUT`, `TOOLHIVE_HEALTH_CHECK_RETRY_DELAY`, `TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD`) following the existing `TOOLHIVE_HEALTH_CHECK_INTERVAL` pattern.
- Raise the default failure threshold from 3 to 5, extending the tolerance window from ~40s to ~60s out of the box.
- Add warning logs when invalid env var values are provided, and debug logs when custom values are successfully applied.

Fixes #4084

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Built a test binary and started an MCP server with invalid env vars (`TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD=banana`, `TOOLHIVE_HEALTH_CHECK_PING_TIMEOUT=not-a-duration`). Verified warning logs appeared in the server log file with the env var name, invalid value, and default fallback. Confirmed `default=5` in the threshold warning, showing the new default is active.

## Changes

| File | Change |
|------|--------|
| `pkg/transport/proxy/transparent/transparent_proxy.go` | Add env var constants, getter functions, struct field for failure threshold, wire getters in constructor, replace package-level `healthCheckRetryCount` constant with struct field |
| `pkg/transport/proxy/transparent/transparent_test.go` | Table-driven tests for all 4 getters, constructor wiring tests, option function tests, test helper `newMinimalProxy()` |
| `docs/arch/03-transport-architecture.md` | Add "Health Check Tuning Parameters" section documenting all 4 env vars |

Yes. Three new environment variables for tuning health check behavior:

| Environment Variable | Description | Default | Type |
|---|---|---|---|
| `TOOLHIVE_HEALTH_CHECK_PING_TIMEOUT` | Timeout for each health check ping | `5s` | duration |
| `TOOLHIVE_HEALTH_CHECK_RETRY_DELAY` | Delay between retry attempts after a failure | `5s` | duration |
| `TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD` | Consecutive failures before proxy shutdown | `5` | integer |

**Behavioral change:** The default failure threshold increases from 3 to 5. This means the proxy tolerates approximately 60 seconds of consecutive health check failures before shutting down (previously ~40 seconds). This improves resilience to transient network disruptions without significantly delaying detection of genuinely dead servers.

- The existing `getHealthCheckInterval()` was also updated to add `slog.Warn` on invalid values and `slog.Debug` on custom values, for consistency with the new getter functions.
- The `healthCheckRetryCount` package-level constant was replaced with a `healthCheckFailureThreshold` struct field to allow per-proxy configuration via the options pattern.
- Per-server flags (e.g., `--health-check-failure-threshold` on `thv run`) would be a natural follow-up, following the precedence pattern suggested in #4063. Happy to include that in this PR if you'd prefer it bundled, or as a separate follow-up.

### DIFF
--- a/docs/arch/03-transport-architecture.md
+++ b/docs/arch/03-transport-architecture.md
@@ -370,7 +370,9 @@ The transparent proxy health check behavior can be tuned via environment variabl
 
 Duration values use Go's `time.ParseDuration` format (e.g., `10s`, `500ms`, `1m30s`). Invalid values are ignored with a warning log, and the default is used instead.
 
-**Failure window**: With the defaults, the proxy tolerates roughly `(threshold-1) × (interval + retryDelay)` plus the initial failure tick before shutting down — approximately 70 seconds with default values. This is designed to survive transient network disruptions without prematurely killing healthy backends.
+**Threshold of 1**: Setting `TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD=1` means the proxy shuts down on the first health check failure with no retries.
+
+**Failure window**: With the defaults, the proxy tolerates roughly `(threshold-1) × (interval + retryDelay)` before shutting down — approximately 60 seconds with default values. This is designed to survive transient network disruptions without prematurely killing healthy backends. If `TOOLHIVE_HEALTH_CHECK_PING_TIMEOUT` exceeds `TOOLHIVE_HEALTH_CHECK_INTERVAL`, each health check cycle takes longer than one interval tick, extending the failure window beyond what the formula predicts.
 
 **Usage example** (increase tolerance for a flaky network):
 ```bash
--- a/pkg/transport/proxy/transparent/transparent_proxy.go
+++ b/pkg/transport/proxy/transparent/transparent_proxy.go
@@ -1072,12 +1072,7 @@ func (p *TransparentProxy) CloseListener() error {
 // performHealthCheckRetry performs a retry health check after a delay
 // Returns true if the retry was successful (health check recovered), false otherwise
 func (p *TransparentProxy) performHealthCheckRetry(ctx context.Context) bool {
-	retryDelay := p.healthCheckRetryDelay
-	if retryDelay == 0 {
-		retryDelay = DefaultHealthCheckRetryDelay
-	}
-
-	retryTimer := time.NewTimer(retryDelay)
+	retryTimer := time.NewTimer(p.healthCheckRetryDelay)
 	defer retryTimer.Stop()
 
 	select {
--- a/pkg/transport/proxy/transparent/transparent_test.go
+++ b/pkg/transport/proxy/transparent/transparent_test.go
@@ -706,7 +706,7 @@ func TestWithHealthCheckFailureThresholdOption(t *testing.T) {
 }
 
 func TestWithHealthCheckFailureThresholdOption_IgnoresNonPositive(t *testing.T) {
-	t.Parallel()
+	t.Setenv(HealthCheckFailureThresholdEnvVar, "")
 
 	proxy := newMinimalProxy(withHealthCheckFailureThreshold(0))
 


amirejaz:
@claude could you review this PR?

amirejaz:
@claude Review this PR focusing on the new TOOLHIVE_HEALTH_CHECK_* env vars and their interactions, edge cases, and documentation.

docs/arch/03-transport-architecture.md
```diff
@@ -336,6 +336,31 @@ export TOOLHIVE_REMOTE_HEALTHCHECKS=true
...
+| Environment Variable | Description | Default | Type |
+|---|---|---|---|
+| `TOOLHIVE_HEALTH_CHECK_INTERVAL` | How often to run health checks | `10s` | duration |
+| `TOOLHIVE_HEALTH_CHECK_PING_TIMEOUT` | Timeout for each health check ping | `5s` | duration |
+| `TOOLHIVE_HEALTH_CHECK_RETRY_DELAY` | Delay between retry attempts after a failure | `5s` | duration |
+| `TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD` | Consecutive failures before proxy shutdown | `5` | integer |
+
+Duration values use Go's `time.ParseDuration` format (e.g., `10s`, `500ms`, `1m30s`). Invalid values are ignored with a warning log, and the default is used instead.
+
+**Failure window**: With the defaults, the proxy tolerates roughly `(threshold-1) × (interval + retryDelay)` plus the initial failure tick before shutting down — approximately 70 seconds with default values. This is designed to survive transient network disruptions without prematurely killing healthy backends.
```
amirejaz: **Inaccurate estimate**: The formula in this sentence gives 60s with defaults, not ~70s.

`(threshold-1) × (interval + retryDelay) = (5-1) × (10s + 5s) = 60s`

The "plus the initial failure tick" phrase is ambiguous — it's not a separate additive term, it's already captured by the (threshold-1) cycles in the formula. The 70s figure doesn't follow from the math.

```suggestion
**Failure window**: With the defaults, the proxy tolerates roughly `(threshold-1) × (interval + retryDelay)` before shutting down — approximately 60 seconds with default values. This is designed to survive transient network disruptions without prematurely killing healthy backends.
```

docs/arch/03-transport-architecture.md
```diff
@@ -336,6 +336,31 @@ export TOOLHIVE_REMOTE_HEALTHCHECKS=true
...
+
```
amirejaz, line 376: **Undocumented interaction worth calling out**: when `TOOLHIVE_HEALTH_CHECK_PING_TIMEOUT > TOOLHIVE_HEALTH_CHECK_INTERVAL`, each retry blocks the monitor goroutine longer than one interval tick. The next tick fires into a buffered channel, so the effective cycle time becomes `pingTimeout + retryDelay + pingTimeout` rather than `interval + retryDelay`. This silently extends the failure window beyond what the formula predicts.

Consider adding a note here:
> **Note**: If `TOOLHIVE_HEALTH_CHECK_PING_TIMEOUT` is larger than `TOOLHIVE_HEALTH_CHECK_INTERVAL`, the effective failure window will exceed the formula above, because each retry blocks the health monitor goroutine for the full ping timeout duration.

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -646,10 +714,10 @@ func (p *TransparentProxy) handleHealthCheckFailure(
 	slog.Warn("health check failed",
 		"target", p.targetURI,
 		"attempt", consecutiveFailures,
-		"max_attempts", healthCheckRetryCount,
+		"max_attempts", p.healthCheckFailureThreshold,
 		"status", status)
 
-	if consecutiveFailures < healthCheckRetryCount {
+	if consecutiveFailures < p.healthCheckFailureThreshold {
```
amirejaz, line 1109: **Edge case: `threshold=1` silently disables retries.** After `consecutiveFailures++` the count is 1. `1 < 1` is false, so execution falls straight through to shutdown — `performHealthCheckRetry` is never called. A user setting `TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD=1` would get "instant shutdown on first failure" behaviour with no retry, which is almost certainly not what they expect given the name.

The getter already accepts 1 as valid (the "valid minimum" test case). Options:
1. Raise the minimum to 2 in `getHealthCheckFailureThreshold` (validate `n >= 2`), or
2. Document this clearly in the env var table in the arch doc (e.g. "Note: a value of 1 disables retries").

At minimum this should be noted in the docs.

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -598,15 +675,6 @@ func (p *TransparentProxy) CloseListener() error {
...
-// These values are designed to handle transient network issues like
-// VPN/firewall idle connection timeouts (commonly 5-10 minutes).
-const (
-	// healthCheckRetryCount is the number of consecutive failures before marking unhealthy.
-	// This prevents immediate shutdown on transient network issues.
-	healthCheckRetryCount = 3
-)
-
 // performHealthCheckRetry performs a retry health check after a delay
 // Returns true if the retry was successful (health check recovered), false otherwise
```
amirejaz, line 1073: **Redundant guard**: The `retryDelay == 0` branch below (line 682) can never be true anymore. `healthCheckRetryDelay` is now always initialised to a positive value by `getHealthCheckRetryDelay()` in the constructor — either from the env var or `DefaultHealthCheckRetryDelay`. The `withHealthCheckRetryDelay` option also guards against non-positive values.

This was meaningful when the struct field could be left at the zero-value, but it's now dead code. Consider removing it to avoid confusion for future readers.

pkg/transport/proxy/transparent/transparent_test.go
```diff
@@ -570,15 +570,148 @@ func TestTransparentProxy_NilUnauthorizedCallback(t *testing.T) {
...
+
+func TestWithHealthCheckFailureThresholdOption(t *testing.T) {
+	t.Parallel()
+
+	proxy := newMinimalProxy(withHealthCheckFailureThreshold(8))
+
+	assert.Equal(t, 8, proxy.healthCheckFailureThreshold)
+}
+
+func TestWithHealthCheckFailureThresholdOption_IgnoresNonPositive(t *testing.T) {
```
amirejaz, line 708: **Test is env-var sensitive and can produce a spurious failure.** This test calls `t.Parallel()` but doesn't isolate itself from `TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD`. If that env var is set in the test environment (e.g. to `10`), `newMinimalProxy()` picks it up. The option `withHealthCheckFailureThreshold(0)` is then a no-op (0 is not > 0), so `proxy.healthCheckFailureThreshold` becomes `10` — and the assertion against `DefaultHealthCheckFailureThreshold` (5) fails.

Fix: drop `t.Parallel()` and add `t.Setenv` for isolation, matching the pattern used in `TestNewTransparentProxyDefaultValues`:

```suggestion
func TestWithHealthCheckFailureThresholdOption_IgnoresNonPositive(t *testing.T) {
	t.Setenv(HealthCheckFailureThresholdEnvVar, "")

	proxy := newMinimalProxy(withHealthCheckFailureThreshold(0))

	assert.Equal(t, DefaultHealthCheckFailureThreshold, proxy.healthCheckFailureThreshold)
}
```

gkatz2:
Hi @amirejaz — thanks for the thorough review. Your 5 inline findings plus the Copilot bot's suggestions have all been addressed in the last two commits (`9e89dc30` and `a41f8a87`):

- Test hermeticity: all getter tests and the default-values test now explicitly set env vars (including empty string for default cases)
- Failure window formula corrected to ~60s
- `pingTimeout > interval` interaction documented
- `threshold=1` behavior (immediate shutdown, no retries) documented
- Dead `retryDelay == 0` guard removed

Would you have time for a re-review? Thanks!

docs/arch/03-transport-architecture.md
```diff
@@ -336,6 +336,31 @@ export TOOLHIVE_REMOTE_HEALTHCHECKS=true
...
+| Environment Variable | Description | Default | Type |
+|---|---|---|---|
+| `TOOLHIVE_HEALTH_CHECK_INTERVAL` | How often to run health checks | `10s` | duration |
+| `TOOLHIVE_HEALTH_CHECK_PING_TIMEOUT` | Timeout for each health check ping | `5s` | duration |
+| `TOOLHIVE_HEALTH_CHECK_RETRY_DELAY` | Delay between retry attempts after a failure | `5s` | duration |
+| `TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD` | Consecutive failures before proxy shutdown | `5` | integer |
+
+Duration values use Go's `time.ParseDuration` format (e.g., `10s`, `500ms`, `1m30s`). Invalid values are ignored with a warning log, and the default is used instead.
+
+**Failure window**: With the defaults, the proxy tolerates approximately 60 seconds of consecutive health check failures before shutting down (5 intervals × 10s + retry delays). This is designed to survive transient network disruptions without prematurely killing healthy backends.
```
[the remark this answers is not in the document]
gkatz2: Addressed in 9e89dc30 — fixed the formula and removed the ambiguous phrasing.

pkg/transport/proxy/transparent/transparent_test.go
```diff
@@ -570,15 +570,153 @@ func TestTransparentProxy_NilUnauthorizedCallback(t *testing.T) {
...
+		{name: "negative duration", envValue: "-5s", expected: DefaultHealthCheckInterval},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.envValue != "" {
+				t.Setenv(HealthCheckIntervalEnvVar, tt.envValue)
+			}
+			assert.Equal(t, tt.expected, getHealthCheckInterval())
+		})
```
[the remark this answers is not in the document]
gkatz2: Addressed in 9e89dc30 — all getter tests now explicitly set the env var for every subtest (including empty string for defaults), and TestNewTransparentProxyDefaultValues clears all health check env vars.

pkg/transport/proxy/transparent/transparent_test.go
```diff
@@ -570,15 +570,153 @@ func TestTransparentProxy_NilUnauthorizedCallback(t *testing.T) {
...
+		{name: "negative duration", envValue: "-5s", expected: DefaultPingerTimeout},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.envValue != "" {
+				t.Setenv(HealthCheckPingTimeoutEnvVar, tt.envValue)
+			}
+			assert.Equal(t, tt.expected, getHealthCheckPingTimeout())
+		})
```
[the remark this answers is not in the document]
gkatz2: Addressed in 9e89dc30 — all getter tests now explicitly set the env var for every subtest (including empty string for defaults), and TestNewTransparentProxyDefaultValues clears all health check env vars.

pkg/transport/proxy/transparent/transparent_test.go
```diff
@@ -570,15 +570,153 @@ func TestTransparentProxy_NilUnauthorizedCallback(t *testing.T) {
...
+		{name: "negative duration", envValue: "-5s", expected: DefaultHealthCheckRetryDelay},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.envValue != "" {
+				t.Setenv(HealthCheckRetryDelayEnvVar, tt.envValue)
+			}
+			assert.Equal(t, tt.expected, getHealthCheckRetryDelay())
+		})
```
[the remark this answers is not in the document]
gkatz2: Addressed in 9e89dc30 — all getter tests now explicitly set the env var for every subtest (including empty string for defaults), and TestNewTransparentProxyDefaultValues clears all health check env vars.

pkg/transport/proxy/transparent/transparent_test.go
```diff
@@ -570,15 +570,153 @@ func TestTransparentProxy_NilUnauthorizedCallback(t *testing.T) {
...
+		{name: "float value", envValue: "2.5", expected: DefaultHealthCheckFailureThreshold},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.envValue != "" {
+				t.Setenv(HealthCheckFailureThresholdEnvVar, tt.envValue)
+			}
+			assert.Equal(t, tt.expected, getHealthCheckFailureThreshold())
+		})
```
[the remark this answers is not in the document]
gkatz2: Addressed in 9e89dc30 — all getter tests now explicitly set the env var for every subtest (including empty string for defaults), and TestNewTransparentProxyDefaultValues clears all health check env vars.

pkg/transport/proxy/transparent/transparent_test.go
```diff
@@ -570,15 +570,153 @@ func TestTransparentProxy_NilUnauthorizedCallback(t *testing.T) {
...
+
+func TestNewTransparentProxyDefaultValues(t *testing.T) {
+	t.Parallel()
+
+	proxy := newMinimalProxy()
+
+	assert.Equal(t, DefaultPingerTimeout, proxy.healthCheckPingTimeout)
+	assert.Equal(t, DefaultHealthCheckRetryDelay, proxy.healthCheckRetryDelay)
+	assert.Equal(t, DefaultHealthCheckFailureThreshold, proxy.healthCheckFailureThreshold)
+	assert.Equal(t, DefaultHealthCheckInterval, proxy.healthCheckInterval)
```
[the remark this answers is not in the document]
gkatz2: Addressed in 9e89dc30 — all getter tests now explicitly set the env var for every subtest (including empty string for defaults), and TestNewTransparentProxyDefaultValues clears all health check env vars.

docs/arch/03-transport-architecture.md
```diff
@@ -336,6 +336,31 @@ export TOOLHIVE_REMOTE_HEALTHCHECKS=true
...
+
+**Usage example** (increase tolerance for a flaky network):
+```bash
+export TOOLHIVE_HEALTH_CHECK_FAILURE_THRESHOLD=10
+export TOOLHIVE_HEALTH_CHECK_RETRY_DELAY=10s
+```
+
+> **Note**: These parameters only affect the transparent proxy (used by SSE and streamable HTTP transports). The stdio transport's streamable HTTP proxy has a separate `TOOLHIVE_PROXY_REQUEST_TIMEOUT` setting. The vMCP server uses its own circuit breaker pattern.
```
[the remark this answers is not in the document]
gkatz2: Addressed in 9e89dc30 — removed the reference to the not-yet-merged env var.

docs/arch/03-transport-architecture.md
gkatz2: Fixed in a41f8a87 — corrected to 60s with the accurate formula.

docs/arch/03-transport-architecture.md
gkatz2: Added a note about this in a41f8a87 — the docs now call out that pingTimeout > interval extends the effective failure window.

pkg/transport/proxy/transparent/transparent_proxy.go
gkatz2: Documented this behavior in a41f8a87 — the docs now note that threshold=1 means immediate shutdown with no retries.

pkg/transport/proxy/transparent/transparent_proxy.go
gkatz2: Removed in a41f8a87 — the dead code guard was pre-existing but is now gone.

pkg/transport/proxy/transparent/transparent_test.go
gkatz2: Fixed in a41f8a87 — the option function test now isolates env vars.

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -598,24 +675,10 @@ func (p *TransparentProxy) CloseListener() error {
...
 func (p *TransparentProxy) performHealthCheckRetry(ctx context.Context) bool {
-	retryDelay := p.healthCheckRetryDelay
-	if retryDelay == 0 {
-		retryDelay = DefaultHealthCheckRetryDelay
-	}
-
-	retryTimer := time.NewTimer(retryDelay)
+	retryTimer := time.NewTimer(p.healthCheckRetryDelay)
```
jhrozek, line 1075: Nice cleanup removing the zero-value fallback here — since `getHealthCheckRetryDelay()` always returns a positive value, the guard was dead code.

The same pattern still exists in two other places in this file though:
- Line ~387: `if pingTimeout == 0 { pingTimeout = DefaultPingerTimeout }`
- Line ~738: `if interval == 0 { interval = DefaultHealthCheckInterval }`

Both are equally unreachable now. Might be worth removing them too for consistency, or adding a brief comment that they're kept as defense-in-depth.

jhrozek:
@gkatz2 again I'm sorry for letting you wait so long for a review, this is bad contributor experience. We need to make sure work in-flight is handed over better.

I think the PR is in good shape overall and normally I would have acked but seeing there are conflicts, I added some nits inline - don't feel forced to resolve them, they are mostly observations.

gkatz2:
> @gkatz2 again I'm sorry for letting you wait so long for a review, this is bad contributor experience. We need to make sure work in-flight is handed over better.
> 
> I think the PR is in good shape overall and normally I would have acked but seeing there are conflicts, I added some nits inline - don't feel forced to resolve them, they are mostly observations.

Whoops, I missed this comment. @jhrozek, how do you typically handle addressing missed nits like these? Do you want a separate issue for them? Do you want to let them go for now?

jhrozek:
> > @gkatz2 again I'm sorry for letting you wait so long for a review, this is bad contributor experience. We need to make sure work in-flight is handed over better.
> > I think the PR is in good shape overall and normally I would have acked but seeing there are conflicts, I added some nits inline - don't feel forced to resolve them, they are mostly observations.
> 
> Whoops, I missed this comment. @jhrozek, how do you typically handle addressing missed nits like these? Do you want a separate issue for them? Do you want to let them go for now?

No problem @gkatz2 the main decision point is the approved/changes requested. The nits  are something you can either fix in the same PR if you feel like it or send a follow-up.

gkatz2:
> @gkatz2 again I'm sorry for letting you wait so long for a review, this is bad contributor experience. We need to make sure work in-flight is handed over better.
> I think the PR is in good shape overall and normally I would have acked but seeing there are conflicts, I added some nits inline - don't feel forced to resolve them, they are mostly observations.

> Whoops, I missed this comment. @jhrozek, how do you typically handle addressing missed nits like these? Do you want a separate issue for them? Do you want to let them go for now?

> No problem @gkatz2 the main decision point is the approved/changes requested. The nits are something you can either fix in the same PR if you feel like it or send a follow-up.

Sounds good. Follow-up PR: #4791

PR #4110, opened 2026-03-11
Remove discovery cache to fix flaky vMCP tests

ChrisJBurns:
## Summary

- The discovery manager's per-user capability cache was causing flaky E2E test failures in `virtualmcp_discovered_mode_test.go`. When a new backend wasn't ready during the first `AggregateCapabilities` call, the partial result (e.g., 4 tools from 2 backends instead of 6 from 3) was cached and served for all subsequent requests for up to 5 minutes (TTL).
- Since new MCP sessions are infrequent and aggregation latency is negligible compared to LLM round-trips, the cache is premature optimization. This PR removes it entirely, computing discovery results fresh on each request.
- Reverts the `QueriedBackendCount` field added to `AggregationMetadata` in the initial fix approach, since it's no longer needed without caching.

## Type of change

- [x] Bug fix

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/discovery/manager.go` | Remove all cache machinery (map, TTL, size limit, cleanup goroutine, version tracking). `Discover` now calls `AggregateCapabilities` directly. `NewManagerWithRegistry` accepts registry param for API compat but ignores it. `Stop()` is now a no-op. |
| `pkg/vmcp/discovery/manager_test.go` | Remove all cache-specific tests (caching, expiration, size limit, version invalidation, partial result). Keep core tests: `TestNewManager`, `TestNewManagerWithRegistry`, `TestDefaultManager_Discover`, `TestDefaultManager_Stop`. |
| `pkg/vmcp/discovery/manager_test_coverage.go` | Deleted — all 3 tests were cache-specific. |
| `pkg/vmcp/aggregator/aggregator.go` | Remove `QueriedBackendCount` field from `AggregationMetadata`. |
| `pkg/vmcp/aggregator/default_aggregator.go` | Remove `QueriedBackendCount` assignment and related span/log attributes. |

No

Per @jerm-dro's [review comment](https://github.com/stacklok/toolhive/pull/4110#discussion_r2921527185), the cache was premature optimization. This approach is simpler and eliminates the root cause rather than working around it.

## Large PR Justification
- removing a bunch of code around caching


pkg/vmcp/aggregator/default_aggregator.go
```diff
@@ -475,19 +475,23 @@ func (a *defaultAggregator) AggregateCapabilities(
 		return nil, fmt.Errorf("failed to merge capabilities: %w", err)
 	}
 
-	// Update metadata with backend count
+	// Update metadata with backend counts
 	aggregated.Metadata.BackendCount = len(backends)
+	aggregated.Metadata.QueriedBackendCount = len(capabilities)
```
reyortiz3: are backends and capabilities the same thing? e.g. `QueriedBackendCount = len(capabilities)`

ChrisJBurns: @reyortiz3 

Good question! I don't think they they're the same thing. Capabilities is a `map[backendID]*BackendCapabilities` returned by `QueryAllCapabilities`. When a backend fails to respond, it's silently skipped (logged as a warning, not added to the map), so `len(capabilities)` equals the number of backends that were successfully queried, which can be less than `len(backends)`. I've added a clarifying comment in the code. Happy to be corrected by @yrobla  @amirejaz  @jerm-dro here though

reyortiz3: ok, thanks for clarifying, I see now.  I was thrown off by the naming mismatch, `aggregated.Metadata.QueriedBackendCount = len(capabilities)`, would have expected something like `aggregated.Metadata.QueriedCapabilitiesCount = len(capabilities)`

but this is not a blocking comment, I defer to vmcp team

pkg/vmcp/discovery/manager.go
```diff
@@ -157,8 +157,17 @@ func (m *DefaultManager) Discover(ctx context.Context, backends []vmcp.Backend)
...
+	// the result is partial and should not be cached — otherwise subsequent requests
+	// would keep hitting the stale cache entry until it expires (5 min TTL),
+	// preventing newly added backends from being discovered.
+	if caps.Metadata != nil && caps.Metadata.QueriedBackendCount < caps.Metadata.BackendCount {
+		slog.Debug("skipping cache for partial aggregation result",
+			"queried_backends", caps.Metadata.QueriedBackendCount,
+			"total_backends", caps.Metadata.BackendCount)
+	} else {
+		m.cacheCapabilities(cacheKey, caps)
+	}
```
jerm-dro: Honestly, I wonder if we should just delete the cache? It seems like a premature optimization. New sessions are rare and latency is going to be dominated by whatever the LLM is doing.

ChrisJBurns: Yep sounds good, removed!

pkg/vmcp/discovery/manager.go
```diff
@@ -52,82 +38,34 @@ type Manager interface {
...
-	// Start background cleanup goroutine
-	m.wg.Add(1)
-	go m.cleanupExpiredEntries()
-
-	return m, nil
+// NewManagerWithRegistry creates a new discovery manager.
+// The registry parameter is accepted for API compatibility but is no longer used
+// for caching. Discovery results are computed fresh on each request.
+func NewManagerWithRegistry(agg aggregator.Aggregator, _ vmcp.DynamicRegistry) (Manager, error) {
+	return NewManager(agg)
```
jerm-dro, line 61: Let's just delete this function and use `NewManager` instead?

jerm-dro (approved):

PR #4111, opened 2026-03-11
Wire non-interactive registry auth for serve mode

ChrisJBurns:
## Summary

- **Why:** When `thv serve` runs with OAuth-authenticated registries, it must never open a browser for OAuth flows — it's a headless API server. Desktop clients (ToolHive Studio) need: (1) machine-readable auth status to show the correct UI state, (2) structured error responses when auth is missing so they can prompt the user to authenticate, and (3) API endpoints to trigger login/logout without needing a terminal.

- **What:** Adds serve-mode registry auth support to the `thv serve` API:
  - **Non-interactive provider**: `WithInteractive(false)` suppresses browser OAuth flows in serve mode
  - **Auth status in responses**: `auth_status` (`none` / `configured` / `authenticated`) and `auth_type` fields in all registry GET responses
  - **Structured 503 errors**: JSON `{"code": "registry_auth_required", "message": "..."}` when tokens are missing, instead of generic 500s
  - **Typed HTTP errors**: `RegistryHTTPError` with `Unwrap()` for 401/403 detection from upstream registries, with `LimitReader` on error bodies
  - **Auth error propagation**: 401s from the upstream registry are wrapped as `ErrRegistryAuthRequired` and surfaced as 503s to clients
  - **API auth endpoints** (serve mode only):
    - `POST /api/v1beta/registry/auth/login` — triggers interactive OAuth flow (desktop-only, opens browser)
    - `POST /api/v1beta/registry/auth/logout` — clears cached OAuth tokens and registry cache
  - **OAuth config via API**: `PUT /api/v1beta/registry/default` now accepts an optional `auth` object with `issuer`, `client_id`, `audience`, and `scopes` fields for configuring registry OAuth
  - **AuthManager extension**: `GetAuthStatus()` method consolidates auth state inspection for API responses

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/registry/api/client.go` | Add `ErrRegistryUnauthorized`, `RegistryHTTPError` type with `Unwrap()` for 401/403. Replace `fmt.Errorf` with typed errors and `LimitReader` in `GetServer`, `fetchServersPage`, `SearchServers`. |
| `pkg/registry/factory.go` | Add `ProviderOption` / `WithInteractive()`, thread opts through `GetDefaultProviderWithConfig` and `NewRegistryProvider`. |
| `pkg/registry/provider_api.go` | Wrap 401 from validation probe and `GetRegistry()` with `ErrRegistryAuthRequired` for structured 503 responses. |
| `pkg/registry/auth_manager.go` | Add `GetAuthStatus()` method and `AuthStatusNone` / `AuthStatusConfigured` / `AuthStatusAuthenticated` constants. |
| `pkg/registry/auth_manager_test.go` | Table-driven tests for `GetAuthStatus()`. |
| `pkg/api/v1/registry.go` | Serve-mode support: `RegistryRoutes.serveMode`, `NewRegistryRoutesForServe()`, `RegistryRouter(serveMode bool)`. Structured 503 errors: `writeRegistryAuthRequiredError()`, `isRegistryAuthError()`. Auth status: `resolveAuthStatus()`, `auth_status`/`auth_type` fields in `registryInfo` and `getRegistryResponse`. API endpoints: `POST /auth/login`, `POST /auth/logout`. OAuth config: `processAuthUpdate()`, `UpdateRegistryAuthRequest`. Auth error handling in `listRegistries`, `getRegistry`, `listServers`, `updateRegistry`. |
| `pkg/api/server.go` | Pass `serveMode: true` to `RegistryRouter()`. |
| `docs/server/*` | Regenerated swagger docs. |

Yes.

- Registry API responses (`GET /api/v1beta/registry` and `GET /api/v1beta/registry/default`) now include `auth_status` and `auth_type` fields.
- When `thv serve` cannot authenticate to a registry (no cached tokens), registry endpoints return HTTP 503 with `{"code": "registry_auth_required"}` instead of a generic 500.
- New `POST /api/v1beta/registry/auth/login` and `POST /api/v1beta/registry/auth/logout` endpoints (serve mode only) allow desktop clients to trigger OAuth login/logout via the API.
- `PUT /api/v1beta/registry/default` now accepts an `auth` object to configure OAuth settings via the API.

- **503 not 401**: HTTP 503 is intentional per RFC-0043. Studio is authenticated to the `thv serve` API — the problem is that `thv serve` itself lacks a registry credential. 503 correctly signals a server-side dependency issue, not a client auth failure.
- **`auth_status` limitations**: Cannot detect "expired" from config alone (would require calling the token source). We return "authenticated" when a cached token ref exists and rely on Studio handling subsequent 401s as a signal to re-login.
- **Browser-open from login endpoint**: The `/auth/login` endpoint opens the user's default browser for the OAuth flow. This is intentional for desktop-only serve mode (Studio running `thv serve` locally). A follow-up could return a redirect URL instead for remote/container environments.
- **No `omitempty` on auth fields**: `auth_status` and `auth_type` intentionally omit `omitempty` so clients always receive the fields, even when the value is `"none"` or empty string.

### DIFF
--- a/docs/server/docs.go
+++ b/docs/server/docs.go
@@ -857,6 +857,28 @@ const docTemplate = `{
                 },
                 "type": "object"
             },
+            "github_com_stacklok_toolhive_pkg_registry.OAuthPublicConfig": {
+                "description": "AuthConfig contains the non-secret OAuth configuration when auth is configured.\nNil when auth_status is \"none\".",
+                "properties": {
+                    "audience": {
+                        "type": "string"
+                    },
+                    "client_id": {
+                        "type": "string"
+                    },
+                    "issuer": {
+                        "type": "string"
+                    },
+                    "scopes": {
+                        "items": {
+                            "type": "string"
+                        },
+                        "type": "array",
+                        "uniqueItems": false
+                    }
+                },
+                "type": "object"
+            },
             "github_com_stacklok_toolhive_pkg_runner.HeaderForwardConfig": {
                 "description": "HeaderForward contains configuration for injecting headers into requests to remote servers.",
                 "properties": {
@@ -1801,6 +1823,9 @@ const docTemplate = `{
             "pkg_api_v1.getRegistryResponse": {
                 "description": "Response containing registry details",
                 "properties": {
+                    "auth_config": {
+                        "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_registry.OAuthPublicConfig"
+                    },
                     "auth_status": {
                         "description": "AuthStatus is one of: \"none\", \"configured\", \"authenticated\".\nIntentionally omits omitempty — see registryInfo for rationale.",
                         "type": "string"
@@ -2059,6 +2084,9 @@ const docTemplate = `{
             "pkg_api_v1.registryInfo": {
                 "description": "Basic information about a registry",
                 "properties": {
+                    "auth_config": {
+                        "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_registry.OAuthPublicConfig"
+                    },
                     "auth_status": {
                         "description": "AuthStatus is one of: \"none\", \"configured\", \"authenticated\".\nIntentionally omits omitempty so clients always receive the field,\neven when the value is \"none\" (the zero-value equivalent).",
                         "type": "string"
--- a/docs/server/swagger.json
+++ b/docs/server/swagger.json
@@ -850,6 +850,28 @@
                 },
                 "type": "object"
             },
+            "github_com_stacklok_toolhive_pkg_registry.OAuthPublicConfig": {
+                "description": "AuthConfig contains the non-secret OAuth configuration when auth is configured.\nNil when auth_status is \"none\".",
+                "properties": {
+                    "audience": {
+                        "type": "string"
+                    },
+                    "client_id": {
+                        "type": "string"
+                    },
+                    "issuer": {
+                        "type": "string"
+                    },
+                    "scopes": {
+                        "items": {
+                            "type": "string"
+                        },
+                        "type": "array",
+                        "uniqueItems": false
+                    }
+                },
+                "type": "object"
+            },
             "github_com_stacklok_toolhive_pkg_runner.HeaderForwardConfig": {
                 "description": "HeaderForward contains configuration for injecting headers into requests to remote servers.",
                 "properties": {
@@ -1794,6 +1816,9 @@
             "pkg_api_v1.getRegistryResponse": {
                 "description": "Response containing registry details",
                 "properties": {
+                    "auth_config": {
+                        "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_registry.OAuthPublicConfig"
+                    },
                     "auth_status": {
                         "description": "AuthStatus is one of: \"none\", \"configured\", \"authenticated\".\nIntentionally omits omitempty — see registryInfo for rationale.",
                         "type": "string"
@@ -2052,6 +2077,9 @@
             "pkg_api_v1.registryInfo": {
                 "description": "Basic information about a registry",
                 "properties": {
+                    "auth_config": {
+                        "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_registry.OAuthPublicConfig"
+                    },
                     "auth_status": {
                         "description": "AuthStatus is one of: \"none\", \"configured\", \"authenticated\".\nIntentionally omits omitempty so clients always receive the field,\neven when the value is \"none\" (the zero-value equivalent).",
                         "type": "string"
--- a/docs/server/swagger.yaml
+++ b/docs/server/swagger.yaml
@@ -821,6 +821,23 @@ components:
           type: array
           uniqueItems: false
       type: object
+    github_com_stacklok_toolhive_pkg_registry.OAuthPublicConfig:
+      description: |-
+        AuthConfig contains the non-secret OAuth configuration when auth is configured.
+        Nil when auth_status is "none".
+      properties:
+        audience:
+          type: string
+        client_id:
+          type: string
+        issuer:
+          type: string
+        scopes:
+          items:
+            type: string
+          type: array
+          uniqueItems: false
+      type: object
     github_com_stacklok_toolhive_pkg_runner.HeaderForwardConfig:
       description: HeaderForward contains configuration for injecting headers into
         requests to remote servers.
@@ -1609,6 +1626,8 @@ components:
     pkg_api_v1.getRegistryResponse:
       description: Response containing registry details
       properties:
+        auth_config:
+          $ref: '#/components/schemas/github_com_stacklok_toolhive_pkg_registry.OAuthPublicConfig'
         auth_status:
           description: |-
             AuthStatus is one of: "none", "configured", "authenticated".
@@ -1806,6 +1825,8 @@ components:
     pkg_api_v1.registryInfo:
       description: Basic information about a registry
       properties:
+        auth_config:
+          $ref: '#/components/schemas/github_com_stacklok_toolhive_pkg_registry.OAuthPublicConfig'
         auth_status:
           description: |-
             AuthStatus is one of: "none", "configured", "authenticated".


pkg/registry/auth/login.go
```diff
@@ -0,0 +1,120 @@
...
+
+// clearRegistryCache removes the persistent cache file for the given registry URL.
+// Uses the same path derivation as CachedAPIRegistryProvider.
+func clearRegistryCache(registryURL string) error {
+	if registryURL == "" {
+		return nil
+	}
+	hash := sha256.Sum256([]byte(registryURL))
+	cacheFile, err := xdg.CacheFile(fmt.Sprintf("toolhive/cache/registry-%x.json", hash[:4]))
+	if err != nil {
```
jhrozek: **[ARCHITECTURE] Duplicated cache path derivation** (HIGH)

This `sha256 → hash[:4] → "toolhive/cache/registry-%x.json"` formula is duplicated from `pkg/registry/provider_cached.go:74-75`, where it uses the `persistentCacheSubdir` constant. Here it's hardcoded as `"cache"`.

If the cache path formula ever changes in `provider_cached.go` (different hash length, different subdir), this logout code will silently fail to clear the correct cache file.

**Suggestion**: Extract a shared `RegistryCachePath(registryURL string) (string, error)` function in `pkg/registry/` and call it from both places. Or add a `ClearCache(url string) error` function to the registry package.

pkg/api/v1/registry.go
```diff
@@ -17,8 +17,127 @@ import (
...
+	_ = json.NewEncoder(w).Encode(body)
+}
+
+// resolveAuthStatus reads the config and returns the auth_status and auth_type
+// strings for API responses.
+// Status logic:
+//   - RegistryAuth.Type == "" → ("none", "") — no auth configured
+//   - Type == "oauth", no CachedRefreshTokenRef → ("configured", "oauth")
+//   - Type == "oauth", CachedRefreshTokenRef present → ("authenticated", "oauth")
+func resolveAuthStatus(cfg *config.Config) (authStatus, authType string) {
```
jhrozek: **[ARCHITECTURE] `resolveAuthStatus` overlaps with `AuthManager.GetAuthInfo()`** (MEDIUM)

This function reimplements the same config inspection as `AuthManager.GetAuthInfo()` in `pkg/registry/auth_manager.go:81-91` (checks `CachedRefreshTokenRef`), but returns three-state strings vs boolean. `GetAuthInfo()` currently has zero production callers.

Per ToolHive's thin-wrapper principle, consider extending `AuthManager` with a `GetAuthStatus() (status, authType string)` method so the API layer just calls through.

pkg/api/v1/registry.go
```diff
@@ -17,8 +17,127 @@ import (
...
+	if cfg == nil || cfg.RegistryAuth.Type == "" {
+		return authStatusNone, ""
+	}
+	switch cfg.RegistryAuth.Type {
+	case config.RegistryAuthTypeOAuth:
+		if cfg.RegistryAuth.OAuth == nil ||
+			cfg.RegistryAuth.OAuth.CachedRefreshTokenRef == "" {
+			return authStatusConfigured, "oauth"
+		}
```
jhrozek: **[DESIGN QUESTION] Browser-open from server process** (MEDIUM)

This endpoint calls `auth.Login(...)` with `interactive=true`, triggering `performOAuthFlow` which opens the user's default browser and starts a local callback listener.

Is this intentional for desktop-only serve mode (Studio running `thv serve` locally)? In headless/remote/container environments this will fail or hang.

Would it be better to return a redirect URL in the JSON response that the client (Studio) opens itself, rather than having the server process open the browser? That would also work behind reverse proxies where `localhost:PORT/callback` isn't reachable from the browser.

test-registry-auth-serve-mode.md
```diff
@@ -0,0 +1,144 @@
+# Testing Registry Auth Serve Mode (PR 2A)
```
jhrozek: **[CONVENTIONS] Test plan at repository root** (LOW)

This manual test plan references a different branch (`feat/registry-auth-serve-mode`), mentions "PR 2B, not yet implemented" for login (which is now implemented), and includes an internal registry URL.

Consider removing from committed code and including in the PR description instead, or moving to `docs/testing/` if it should persist.

pkg/api/v1/registry.go
```diff
@@ -17,8 +17,127 @@ import (
...
+// It triggers an interactive OAuth flow that opens the user's browser.
+// This endpoint is only available in serve mode.
+func (rr *RegistryRoutes) registryAuthLogin(w http.ResponseWriter, r *http.Request) {
+	secretsProvider, err := auth.NewSecretsProvider(rr.configProvider)
+	if err != nil {
+		slog.Error("failed to create secrets provider", "error", err)
+		http.Error(w, "Failed to create secrets provider", http.StatusInternalServerError)
+		return
+	}
+
```
jhrozek, line 94: **[CONVENTIONS] Missing `omitempty` on `auth_status`/`auth_type`** (INFO)

The `AuthStatus`/`AuthType` fields on `registryInfo` and `getRegistryResponse` (lines 768-771, 798-801) serialize `auth_type` as `""` when no auth is configured. This is inconsistent with other optional fields like `Audience` in `UpdateRegistryAuthRequest` which use `omitempty`.

Is this intentional for easier client parsing (always present)? If so, a brief comment would clarify.

pkg/registry/auth_manager.go
```diff
@@ -50,6 +50,12 @@ func (c *DefaultAuthManager) SetOAuthAuth(issuer, clientID, audience string, sco
 		return fmt.Errorf("OIDC discovery failed for issuer %s: %w", issuer, err)
 	}
 
+	// Default to openid + offline_access if no scopes provided.
+	// offline_access is required to receive a refresh token from the provider.
+	if len(scopes) == 0 {
+		scopes = []string{"openid", "offline_access"}
```
jhrozek: **[OAUTH] Default scopes defined in three locations** (LOW)

The default `["openid", "offline_access"]` scopes are defined here, in `cmd/thv/app/config_registryauth.go` (flag default), and a different set `["openid", "profile", "email"]` in `pkg/auth/oauth/oidc.go:227`.

Consider extracting a single `DefaultRegistryScopes` constant to avoid divergence.

test-registry-auth-serve-mode.md
ChrisJBurns: Yeah this was just me asking Claude to output an .md file so i can run through things manually locally. This was only committed so I didn't lose it across branches. This won't stay for long after I've proved everything locally.

jhrozek:
looks like swagger is unhappy but the code lgtm

pkg/registry/auth/login.go
ChrisJBurns: Fixed. Extracted shared `RegistryCacheFilePath()` and `registryCachePath()` into `pkg/registry/auth/cache.go` with a single `registryCacheRelPath()` as the source of truth. Both `provider_cached.go` and `clearRegistryCache()` now delegate to it. Landed in PR #4121.

pkg/api/v1/registry.go
ChrisJBurns: Good call. Added `GetAuthStatus()` to `AuthManager` that builds on `GetAuthInfo()`, and `resolveAuthStatus()` now just delegates to it. Fixed in cf7770ef.

pkg/api/v1/registry.go
ChrisJBurns: Yes, this is intentional for desktop-only serve mode where Studio runs `thv serve` locally and the user has a browser. The PR description calls this out explicitly. Returning a redirect URL is a good improvement for remote/container environments — I'll track that as a follow-up.

pkg/api/v1/registry.go
ChrisJBurns: Intentional. Added comments to the struct fields explaining the rationale: clients should always receive these fields so they can distinguish "no auth configured" (empty string) from "field missing" without extra parsing logic.

pkg/registry/auth_manager.go
ChrisJBurns: `config_registryauth.go` now uses `auth.DefaultOAuthScopes()` instead of a hardcoded slice. The different set in `pkg/authserver/server/registration/` (`["openid", "profile", "email", "offline_access"]`) is intentionally different — that's for the auth server's dynamic client registration, which needs `profile` and `email` scopes that registry auth doesn't require.

ChrisJBurns:
@jhrozek  I might need another approval

PR #4102, opened 2026-03-11
Add tool annotation context for Cedar/HTTP authz

JAORMX:
## Summary

- Cedar policies can reference entity attributes like `resource.readOnlyHint == true`, but MCP tool annotations (readOnlyHint, destructiveHint, idempotentHint, openWorldHint) never reached the Cedar or HTTP PDP authorizers. This adds context-based annotation injection so authorization policies can make decisions based on tool properties.
- Adds a `ToolAnnotations` struct and context helpers (`WithToolAnnotations`/`ToolAnnotationsFromContext`) in the `authorizers` package — the shared interface package that both Cedar and HTTP PDP already import, avoiding import cycles.
- Cedar authorizer merges annotations into resource entity attributes, enabling policies like `resource.readOnlyHint == true`.
- HTTP PDP authorizer includes annotations under `context.mcp.annotations` in the PORC structure.
- This is the foundation for PR 2 (annotation cache for regular ToolHive) and PR 3 (vmcp annotation flow).

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`)
- [x] Linting (`task lint-fix`)

## Changes

| File | Change |
|------|--------|
| `pkg/authz/authorizers/annotations.go` | New: `ToolAnnotations` struct, context helpers, `AnnotationsToMap` converter |
| `pkg/authz/authorizers/annotations_test.go` | New: Round-trip context tests, nil handling, map conversion tests |
| `pkg/authz/authorizers/cedar/core.go` | Modified: `authorizeToolCall` accepts context, reads annotations and merges into resource entity attributes; moved `AuthorizeWithJWTClaims` to public methods section |
| `pkg/authz/authorizers/http/core.go` | Modified: `AuthorizeWithJWTClaims` enriches PORC context with annotations; new `enrichPORCWithAnnotations` helper |
| `pkg/authz/authorizers/cedar/annotations_integration_test.go` | New: Integration tests verifying Cedar policies can use `resource.readOnlyHint`, `resource.destructiveHint`, and graceful degradation without annotations |

## Does this introduce a user-facing change?

No — this adds the internal plumbing. Users will benefit once annotations are populated (via annotation cache in PR 2 or vmcp discovery in PR 3), enabling Cedar policies that reference tool annotations.

## Special notes for reviewers

- **Import cycle avoidance**: Annotations live in `pkg/authz/authorizers/` (not `pkg/authz/`) because Cedar and HTTP PDP packages import `authorizers` but not the parent `authz` package.
[...]

PR #4114, opened 2026-03-12
Add annotation cache for tools/list to tools/call flow

JAORMX:
## Summary

- In regular ToolHive, tool annotations (readOnlyHint, destructiveHint, etc.) are visible in `tools/list` responses but lost by the time `tools/call` arrives as a separate HTTP request. This adds an in-memory annotation cache that bridges the two operations, enabling Cedar/HTTP PDP policies to reference annotations during `tools/call` authorization.
- Builds on #4102 which added the annotation context plumbing and authorizer integration.

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/authz/annotation_cache.go` | New: `AnnotationCache` with concurrent-safe `Get`/`Set`/`SetFromToolsList`, nil-safe on nil receiver |
| `pkg/authz/annotation_cache_test.go` | New: Round-trip, bulk populate, nil safety, concurrent access tests |
| `pkg/authz/response_filter.go` | Modified: `ResponseFilteringWriter` gains `annotationCache` field; `filterToolsResponse` populates cache from parsed tool list |
| `pkg/authz/middleware.go` | Modified: `Middleware` closure creates per-proxy cache; `tools/call` looks up annotations and injects via `WithToolAnnotations` |
| `pkg/authz/response_filter_test.go` | Modified: Updated `NewResponseFilteringWriter` call sites for new parameter |

No — internal plumbing. Users benefit when Cedar policies reference tool annotations (e.g. `resource.readOnlyHint == true`), which now work for `tools/call` after a `tools/list` has populated the cache.

- **One cache per proxy**: Created in the `Middleware` closure, so each MCP server proxy has its own cache. Tool names are scoped to a single server, so this is correct.
- **Graceful degradation**: If `tools/call` arrives before any `tools/list`, `Get` returns nil and no annotations are injected — same behavior as before PR 1.
- **Nil-safe receiver**: All `AnnotationCache` methods handle nil receiver, so existing code passing `nil` to `NewResponseFilteringWriter` works unchanged.
- **Cache populated before filtering**: `SetFromToolsList` runs before the authorization loop, so annotations for all tools (including ones that will be filtered out) are cached.

### DIFF
--- a/pkg/authz/annotation_cache.go
+++ b/pkg/authz/annotation_cache.go
@@ -54,24 +54,30 @@ func (c *AnnotationCache) Set(toolName string, annotations *authorizers.ToolAnno
 }
 
 // SetFromToolsList extracts annotations from a tools/list response and
-// populates the cache. Only tools that have at least one non-nil annotation
-// hint are cached; tools with all-nil hints (the zero value) are skipped
-// to avoid unnecessary memory consumption.
+// replaces the entire cache contents. The full replacement ensures that
+// tools whose annotations were removed in a subsequent tools/list response
+// do not retain stale cached entries.
+//
+// Only tools that have at least one non-nil annotation hint are cached;
+// tools with all-nil hints (the zero value) are skipped to avoid
+// unnecessary memory consumption.
 //
 // This method is a no-op if the cache is nil.
 func (c *AnnotationCache) SetFromToolsList(tools []mcp.Tool) {
 	if c == nil {
 		return
 	}
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	newTools := make(map[string]*authorizers.ToolAnnotations, len(tools))
 	for i := range tools {
 		ann := &tools[i].Annotations
 		if !hasAnyHint(ann) {
 			continue
 		}
-		c.tools[tools[i].Name] = convertMCPAnnotation(ann)
+		newTools[tools[i].Name] = convertMCPAnnotation(ann)
 	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.tools = newTools
 }
 
 // hasAnyHint reports whether the MCP tool annotation has at least one
@@ -84,7 +90,9 @@ func hasAnyHint(ann *mcp.ToolAnnotation) bool {
 }
 
 // convertMCPAnnotation converts an mcp-go ToolAnnotation to the authz
-// ToolAnnotations type used by authorizers.
+// ToolAnnotations type used by authorizers. Only hint fields are copied;
+// the Title field is intentionally omitted because authorizers only use
+// hints for policy decisions.
 func convertMCPAnnotation(ann *mcp.ToolAnnotation) *authorizers.ToolAnnotations {
 	return &authorizers.ToolAnnotations{
 		ReadOnlyHint:    ann.ReadOnlyHint,
--- a/pkg/authz/annotation_cache_test.go
+++ b/pkg/authz/annotation_cache_test.go
@@ -238,6 +238,29 @@ func TestAnnotationCache_ConcurrentAccess(t *testing.T) {
 	assert.NotNil(t, got.ReadOnlyHint, "ReadOnlyHint should be set")
 }
 
+func TestAnnotationCache_SetFromToolsListEvictsStaleEntries(t *testing.T) {
+	t.Parallel()
+
+	cache := NewAnnotationCache()
+
+	// First tools/list: two tools with annotations.
+	cache.SetFromToolsList([]mcp.Tool{
+		{Name: "weather", Annotations: mcp.ToolAnnotation{ReadOnlyHint: boolPtr(true)}},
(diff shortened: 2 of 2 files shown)


pkg/authz/annotation_cache.go
```diff
@@ -0,0 +1,103 @@
...
+// Set stores annotations for a single tool. It is a no-op if the cache
+// is nil.
+func (c *AnnotationCache) Set(toolName string, annotations *authorizers.ToolAnnotations) {
+	if c == nil {
+		return
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.tools[toolName] = annotations
+}
```
aponcedeleonch, line 54: small suggestion: make the key of the map some combination of `<server_name><separator><tool_name>` to avoid key clashes between tools with the same name that belong to a different MCP server

JAORMX, line 54: Good thinking! Note that the cache is actually scoped per proxy instance already. Each call to `Middleware()` in `middleware.go` creates its own `NewAnnotationCache()` inside the closure, so every MCP server gets an independent cache. Tool names within a single MCP server are unique per the protocol spec, so there shouldn't be key collisions here.

So, basically, cross-server clashes can't happen because different servers never share the same cache instance.

JAORMX (commented):
Thanks for the review! Replied inline.

pkg/authz/annotation_cache.go
```diff
@@ -0,0 +1,103 @@
...
+	for i := range tools {
+		ann := &tools[i].Annotations
+		if !hasAnyHint(ann) {
+			continue
+		}
+		newTools[tools[i].Name] = convertMCPAnnotation(ann)
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.tools = newTools
```
jhrozek, line 80: this was flagged by an automated review. I think it's acceptable and I honestly don't know how realistic the issue even is, but apparently tools/list can be paginated and if it were, then each page would overwrite the previous page's annotations.

I don't know how much of a real-world impact this has and don't know if there exist servers that paginate.

pkg/authz/middleware.go
```diff
@@ -165,10 +165,19 @@ func handleUnauthorized(w http.ResponseWriter, msgID interface{}, err error) {
...
+// subsequent tools/call request arrives, the cached annotations are injected into
+// the request context so that authorizers can use them for policy decisions.
+//
 // The authorizer parameter should implement the authorizers.Authorizer interface,
 // which can be created using authz.CreateMiddlewareFromConfig() or directly
 // from an authorizer package (e.g., cedar.NewCedarAuthorizer()).
 func Middleware(a authorizers.Authorizer, next http.Handler) http.Handler {
+	// Cache is shared across requests for the same proxy.
+	// Populated from tools/list responses, read during tools/call.
+	annotationCache := NewAnnotationCache()
```
jhrozek, line 179: it would be nice to track a follow-up enhancement to invalidate the cache on list_changed - I don't think it is a big deal because currently the client will/should just call tools/list after the changed notification.

pkg/authz/annotation_cache.go
aponcedeleonch: ah got it! I didn't notice this was going to be per proxy. All good then

pkg/authz/annotation_cache.go
JAORMX: Good catch. So, I dug into the MCP spec on this one and yes, `tools/list` does support pagination via `nextCursor`. Each page is a separate JSON-RPC request/response pair, and our proxy forwards them transparently (it doesn't aggregate pages). That means with the current full-replacement approach, page 2 would evict page 1's annotations.

That said, the real-world impact is pretty low right now:

1. I'm not aware of any MCP servers that actually paginate `tools/list` today
2. The failure mode is graceful. A cache miss just means the authorizer doesn't get the hint annotations (readOnly, destructive, etc.). Name-based authorization still works fine, so there's no security bypass here.
3. Switching to merge semantics has its own trade-off: we'd lose the stale-entry cleanup that the full replacement gives us

I'll add a comment in the code documenting this limitation and track a follow-up to switch to merge semantics if pagination becomes a real thing in practice.

pkg/authz/middleware.go
JAORMX: Agreed, worth tracking as a follow-up. Note that the current behavior is actually spec-compliant and, interestingly, *better* than the naive "clear cache on `list_changed`" approach.

Per the MCP spec, the expected flow is that clients re-call `tools/list` after receiving the notification. When they do, the cache gets refreshed naturally. If we proactively *cleared* the cache on `list_changed`, we'd create a window where `tools/call` requests have *no* annotations at all instead of stale ones. Stale annotations are strictly better for policy decisions than no annotations.

The proper enhancement would be for the proxy to issue its own `tools/list` to the backend when it sees the notification, but that's a bigger architectural change (the middleware would need to initiate backend requests). I'll file a follow-up issue to track this.

PR #4115, opened 2026-03-12
Split vmcp auth/authz middleware for annotation flow

JAORMX:
## Summary

- The vmcp middleware chain previously composed auth+parser+authz into a single middleware, which meant authz ran before discovery could provide tool annotations. This splits auth and authz so authz runs after discovery, enabling annotation-aware authorization policies.
- Builds on #4102 which added the annotation context plumbing and authorizer integration.

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/auth/factory/incoming.go` | Modified: `NewIncomingAuthMiddleware` returns 4 values (auth, authz, handler, error) instead of 3. Auth+parser composed; authz returned separately. |
| `pkg/vmcp/auth/factory/incoming_test.go` | Modified: Updated for 4-return signature, added test for separate authz return |
| `pkg/vmcp/auth/factory/authz_not_wired_test.go` | Modified: Updated to compose auth and authz separately |
| `pkg/vmcp/server/annotation_enrichment.go` | New: Middleware that reads tool annotations from discovery context and injects via `WithToolAnnotations` for authz |
| `pkg/vmcp/server/annotation_enrichment_test.go` | New: Table-driven tests for backend tools, composite tools, passthrough cases |
| `pkg/vmcp/server/server.go` | Modified: Added `AuthzMiddleware` field to Config; inserted authz and annotation-enrichment in middleware chain |
| `cmd/vmcp/app/commands.go` | Modified: Updated call site for 4-return, passes `AuthzMiddleware` to server config |

No — internal middleware reordering. Users benefit when Cedar policies reference tool annotations in vmcp, which now work because authz runs after discovery has populated annotations.

- **New middleware ordering**: `auth+parser → audit → discovery → annotation-enrichment → authz → handler`. The key change is authz moved from before discovery to after.
- **Annotation enrichment is conditional**: Only applied when `AuthzMiddleware` is configured. No overhead when authz is disabled.
- **Only enriches tools/call**: The enrichment middleware is a no-op for all non-`tools/call` methods.
- **Backwards compatible**: When no authz is configured, behavior is identical to before — `authzMiddleware` is nil and skipped.

### DIFF
--- a/pkg/mcp/parser.go
+++ b/pkg/mcp/parser.go
@@ -68,6 +68,13 @@ type ParsedMCPRequest struct {
 //	}
 func ParsingMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Skip if already parsed by an outer middleware (e.g. auth composes
+		// ParsingMiddleware and server.go applies it again for the no-auth case).
+		if GetParsedMCPRequest(r.Context()) != nil {
+			next.ServeHTTP(w, r)
+			return
+		}
+
 		// Check if we should parse this request
 		if !shouldParseMCPRequest(r) {
 			next.ServeHTTP(w, r)
--- a/pkg/vmcp/server/annotation_enrichment.go
+++ b/pkg/vmcp/server/annotation_enrichment.go
@@ -7,6 +7,8 @@ import (
 	"log/slog"
 	"net/http"
 
+	"github.com/mark3labs/mcp-go/mcp"
+
 	"github.com/stacklok/toolhive/pkg/authz/authorizers"
 	mcpparser "github.com/stacklok/toolhive/pkg/mcp"
 	"github.com/stacklok/toolhive/pkg/vmcp"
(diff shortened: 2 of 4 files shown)


pkg/vmcp/auth/factory/incoming.go
```diff
@@ -25,27 +25,33 @@ import (
...
+//   - authInfoHandler: Handler for /.well-known/oauth-protected-resource endpoint (may be nil)
+//   - err: Error if middleware creation fails
 func NewIncomingAuthMiddleware(
 	ctx context.Context,
 	cfg *config.IncomingAuthConfig,
-) (func(http.Handler) http.Handler, http.Handler, error) {
+) (
+	authMw func(http.Handler) http.Handler,
+	authzMw func(http.Handler) http.Handler,
+	authInfoHandler http.Handler,
```
jhrozek, line 47: are the named returns just for clarity? I think the function then returns local variables anyway?

JAORMX: yes, just readability. Returning the handlers looked quite off.

Issue #3917, opened 2026-02-20
Wire UpstreamTokenService and rewrite swap middleware


jhrozek:
## Summary

Connect the UpstreamTokenService to the middleware and simplify the swap logic.

## What

- `EmbeddedAuthServer` constructs the service and exposes it via `UpstreamTokenService()` accessor
- `MiddlewareRunner` interface changes from `GetUpstreamTokenStorage()` to `GetUpstreamTokenService()`
- Middleware rewritten: replace direct storage access + manual expiry check with a single `GetValidTokens()` call + error-to-HTTP mapping

## Why

Completes the upstream token refresh feature. The middleware becomes a thin layer that gets a valid credential or maps an error to the appropriate HTTP status.

Issue #3916, opened 2026-02-20
Implement UpstreamTokenService


jhrozek:
## Summary

New service package that owns upstream token lifecycle: read, refresh, and error handling.

## What

New package `pkg/authserver/upstreamtoken/` with:
- `Service` interface — single method `GetValidTokens(ctx, sessionID)` returning an opaque `UpstreamCredential`
- `inProcessService` implementation — on-read refresh using the existing `upstream.OAuth2Provider.RefreshTokens()`, singleflight deduplication, error classification (permanent vs transient), session revocation on permanent upstream failure
- Sentinel errors for session not found, refresh failed, no refresh token

## Why

The middleware should call a service-level interface, not raw storage. This abstraction represents the business operation and can be exposed over the network when the auth server moves out-of-process.

Issue #3915, opened 2026-02-20
Add storage methods for upstream token refresh


jhrozek:
## Summary

Add storage-layer methods that the upstream token refresh service will depend on.

## What

- `UpdateUpstreamTokens` on `UpstreamTokenStorage` — atomically updates token values for an existing session and resets the sliding-window TTL.
- `SessionRevoker` interface with `RevokeSessionByTSID` — revokes all internal access and refresh tokens where the session's upstream session ID matches the given tsid. Implemented on `MemoryStorage` using the existing O(n) scan pattern.

## Why

Foundation for the UpstreamTokenService. Separated from the service implementation to keep PRs focused.

PR #4117, opened 2026-03-12
Make the token refresh code reusable for vMCP and testable

jhrozek:
## Summary

This PR refactors the in-middleware refresh into an `UpstreamTokenService` that transparently refreshes expired upstream access tokens on behalf of the user, so MCP sessions survive beyond the upstream IDP's token lifespan.

### Commits

1. **Add upstreamtoken package for upstream token lifecycle** — Introduces a reusable `upstreamtoken.Service` interface that encapsulates read → validate expiry → refresh → dedup behind a single `GetValidTokens` call. Designed for reuse by vMCP.

2. **Wire UpstreamTokenService into runner and middleware** — Simplifies the `MiddlewareRunner` interface from two storage-level accessors to a single `GetUpstreamTokenService`. The upstreamswap middleware now consumes the `Service` interface directly instead of composing storage and refresher itself.

3. **Simplify runner startup with eager service initialization** — Reorders `Run()` to initialize the embedded auth server before middleware creation, eliminating the lazy accessor with mutex and caching. The service is now a simple field read.

4. **Fail closed when upstream token service is unavailable** — When a request carries a `tsid` claim (meaning it expects upstream token injection), the previous behavior passed it through with the original JWT, leaking the auth server token to the backend. Now returns 503 instead.

Tested end-to-end against a live Okta instance in a kind cluster. The ToolHive proxy was exposed via ngrok to avoid port-forward instability, and VS Code was used as the MCP client to exercise the real-world OAuth flow (dynamic client registration, PKCE, browser login).

After authenticating through Okta, the proxy successfully obtained upstream tokens including a refresh token. The Okta authorization server was configured with a 5-minute access token lifespan. Once the upstream AT expired, the next MCP tool call from VS Code triggered a transparent refresh — the `UpstreamTokenService` detected the expiry, used singleflight-deduplicated refresh via Okta's token endpoint, stored the new tokens, and injected the fresh AT into the proxied request. The tool call succeeded with no 401 and no re-authentication prompt. Subsequent requests used the refreshed token with a full 5-minute TTL.

No. This is transparent to MCP clients — sessions that previously broke after the upstream AT expired now continue working silently.

- The `upstreamtoken` package is intentionally designed as a standalone service (not embedded in middleware) so vMCP can reuse it for the same refresh logic without depending on the middleware layer.

Fixes: #3914
Fixes: #3915 
Fixes: #3916 
Fixes: #3917 

## Large PR Justification

- Large refactoring that must be atomic

### DIFF
--- a/pkg/auth/upstreamswap/middleware.go
+++ b/pkg/auth/upstreamswap/middleware.go
@@ -175,12 +175,14 @@ func createMiddlewareFunc(cfg *Config, serviceGetter ServiceGetter) types.Middle
 				return
 			}
 
-			// 3. Get token service
+			// 3. Get token service — fail closed if unavailable.
+			// The tsid claim confirms this request expects upstream token injection;
+			// passing through with the original JWT would leak it to the backend.
 			svc := serviceGetter()
 			if svc == nil {
-				slog.Warn("Token service unavailable, proceeding without swap",
+				slog.Warn("Token service unavailable, cannot perform required upstream swap",
 					"middleware", "upstreamswap")
-				next.ServeHTTP(w, r)
+				http.Error(w, "authentication service temporarily unavailable", http.StatusServiceUnavailable)
 				return
 			}
 
--- a/pkg/auth/upstreamswap/middleware_test.go
+++ b/pkg/auth/upstreamswap/middleware_test.go
@@ -19,9 +19,9 @@ import (
 	"go.uber.org/mock/gomock"
 
 	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/auth/upstreamtoken"
 	"github.com/stacklok/toolhive/pkg/authserver/storage"
 	storagemocks "github.com/stacklok/toolhive/pkg/authserver/storage/mocks"
-	"github.com/stacklok/toolhive/pkg/auth/upstreamtoken"
 	"github.com/stacklok/toolhive/pkg/transport/types"
 	"github.com/stacklok/toolhive/pkg/transport/types/mocks"
 )
@@ -47,7 +47,7 @@ func requestWithIdentity(tsid string) *http.Request {
 	identity := &auth.Identity{
 		Subject: "user123",
 		Claims: map[string]any{
-			"sub":                          "user123",
+			"sub":                                "user123",
 			upstreamtoken.TokenSessionIDClaimKey: tsid,
 		},
 	}
@@ -169,7 +169,7 @@ func TestMiddleware_NoTsidClaim(t *testing.T) {
 	assert.True(t, nextCalled, "next handler should be called")
 }
 
-func TestMiddleware_ServiceUnavailable(t *testing.T) {
+func TestMiddleware_ServiceUnavailable_FailsClosed(t *testing.T) {
 	t.Parallel()
 
 	cfg := &Config{}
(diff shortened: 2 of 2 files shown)


pkg/auth/upstreamswap/middleware.go
```diff
@@ -184,152 +167,57 @@ func createMiddlewareFunc(cfg *Config, storageGetter StorageGetter, refresherGet
...
-				// validation (e.g., subject/client mismatch). All three are client-attributable
-				// errors that require the caller to re-authenticate with the upstream IdP.
-				if errors.Is(err, storage.ErrExpired) ||
-					errors.Is(err, storage.ErrNotFound) ||
-					errors.Is(err, storage.ErrInvalidBinding) {
+
+				// Client-attributable errors require re-authentication.
+				if errors.Is(err, upstreamtoken.ErrSessionNotFound) ||
+					errors.Is(err, upstreamtoken.ErrNoRefreshToken) ||
+					errors.Is(err, upstreamtoken.ErrRefreshFailed) ||
```
tgrunnagle, line 198: Question: is it worth handling transient failures (e.g. upstream Idp temporarily unavailable) differently? Or should we keep it simple for now and just force re-auth with 401 for everything?

pkg/auth/upstreamtoken/service.go
```diff
@@ -0,0 +1,122 @@
...
+	if expired.RefreshToken == "" {
+		return nil, ErrNoRefreshToken
+	}
+
+	if s.refresher == nil {
+		slog.Debug("token refresher not configured, cannot refresh upstream tokens")
+		return nil, ErrNoRefreshToken
+	}
+
+	result, err, _ := s.sfGroup.Do(sessionID, func() (any, error) {
```
tgrunnagle, line 96: Claude found a race condition that's probably worth documenting/following up on:

The `singleflight.Group` is per-`InProcessService` (i.e., per-process). In a multi-pod deployment with Redis storage and refresh token rotation (single-use refresh tokens), two pods can independently attempt to refresh the same session simultaneously:

1. Pod A reads expired `AT`, `RT1` from Redis
2. Pod B reads expired `AT`, `RT1` from Redis (same session, same moment)
3. Pod A's singleflight fires → `RefreshAndStore(RT1)` → IDP rotates to `RT2`, stored in Redis
4. Pod B's singleflight fires independently → `RefreshAndStore(RT1)` → IDP **rejects** (RT1 already consumed), refresh fails → `ErrRefreshFailed` → 401 sent to the user's MCP client

The result is a spurious session termination caused by a pod race, not by the user's credentials being invalid.

This is a known architectural limitation: fixing it requires distributed coordination that `storage.UpstreamTokenRefresher` doesn't currently support. Options include:
- A distributed lock (e.g., Redis SETNX / Redlock) around the per-session refresh path
- Compare-and-swap semantics on `RefreshAndStore` — only write if the stored refresh token still matches (another pod hasn't already rotated it); return the current stored tokens on CAS failure instead of an error

This is not introduced by this PR — it exists in the prior code too. However, given that this PR advertises vMCP reuse (where multi-instance is the default), the limitation should be documented either in the package doc or as a follow-up issue.

pkg/auth/upstreamswap/middleware.go
jhrozek: I don't think so after having checked vscode (the only "serious" open-source client). vscode treats 401 as "reauth but don't mark as failed" while anything else (or anything >500? I don't remember) as "just mark this server as failed, the user needs to restart it". They also have an upper bound (three I think?) of auth retries after which the server is permanently failed.

I don't know how other clients behave, but for vscode 401 seems like the best option, because the user would just get re-prompted thrice at most, other http codes would just mark the server as failed.

pkg/auth/upstreamtoken/service.go
jhrozek: If it's OK, I would do a follow-up

jhrozek: #4122

PR #4121, opened 2026-03-12
Add registry login & logout CLI commands

ChrisJBurns:
## Summary

- RFC-0043 requires users to authenticate against private registries. This PR adds `thv registry login` and `thv registry logout` CLI commands so users can interactively authenticate before fetching from a protected registry.
- `thv registry login` triggers a browser-based OAuth PKCE flow, persists the refresh token via the secrets provider, and updates the config with the token reference. It also accepts `--registry`, `--issuer`, `--client-id`, `--audience`, and `--scopes` flags to configure and persist OAuth settings in a single step if not already configured.
- `thv registry logout` clears the cached refresh token from secrets (both stored ref and derived key fallback), removes the persistent registry cache file, and resets token references in the config.
- Login validates all missing configuration upfront and presents a single actionable error listing every required flag.
- Login rejects local file registries early with a clear message.
- Extracts shared `ConfigureOAuth` function used by both `Login` and `AuthManager.SetOAuthAuth` to eliminate duplicated OIDC validation and config persistence logic.
- Extracts `DefaultOAuthScopes()` so both login and `set-registry-auth` use the same defaults.
- Adds issuer URL validation (HTTPS required, HTTP allowed only for localhost per OIDC spec).
- Extracts registry cache path helpers into `cache.go` for reuse between cached provider and login/logout.
- Moves `NewSecretsProvider` out of `pkg/registry/auth` into the CLI layer as `newSecretsProvider` to keep the auth package focused on business logic.

## Type of change

- [x] New feature

## Test plan

Manually tested on macOS against Okta-backed Stacklok registry:
1. Configured registry auth with `thv config set-registry-auth`
2. Ran `thv registry login` — browser opened, completed OAuth flow, refresh token persisted to keychain
3. Verified `thv registry list` returned authenticated results
4. Ran `thv registry logout` — confirmed token removed from keychain, cache file deleted, config cleared
5. Verified `thv registry list` returns 503 (auth required) after logout

## Changes

| File | Change |
|------|--------|
| `cmd/thv/app/registry_login.go` | New `thv registry login` command with OAuth flags and `newSecretsProvider` helper |
| `cmd/thv/app/registry_logout.go` | New `thv registry logout` command reusing `newSecretsProvider` |
| `cmd/thv/app/config_registryauth.go` | Use `auth.DefaultOAuthScopes()` for consistent scope defaults |
| `pkg/registry/auth/login.go` | Core `Login`/`Logout` business logic, `ConfigureOAuth` shared helper, config validation |
| `pkg/registry/auth/login_test.go` | 45 unit tests covering all exported and unexported functions |
| `pkg/registry/auth/issuer_validation.go` | HTTPS issuer URL validation per OIDC spec |
| `pkg/registry/auth/cache.go` | Extracted registry cache path helpers for reuse |
| `pkg/registry/auth/oauth_token_source.go` | Use three-index slice in `ensureOfflineAccess` to avoid mutating caller's slice |
| `pkg/registry/auth_manager.go` | `SetOAuthAuth` delegates to shared `auth.ConfigureOAuth` |
| `pkg/registry/provider_cached.go` | Use `auth.RegistryCacheFilePath` instead of inline cache path derivation |

Yes. Two new CLI commands:
- `thv registry login` — Authenticate with the configured registry via browser-based OAuth. Supports inline configuration of registry URL and OAuth settings via flags.
- `thv registry logout` — Clear cached registry credentials and registry cache.

- This is part of the RFC-0043 series. The serve-mode API endpoints (login/logout via HTTP) will follow in a separate PR once this merges.
- `ensureOfflineAccess()` is intentionally additive — it appends `offline_access` only if not already present, preserving any user-configured scopes.
- Cache path derivation is now shared via `auth.RegistryCacheFilePath` / `auth.registryCachePath` in `cache.go`, eliminating the previous duplication between `CachedAPIRegistryProvider` and `clearRegistryCache`.

## Large PR Justification
- Contains all code for login/logout functionality. I can split into smaller PRs if people prefer to review logout first and then login


pkg/registry/auth/login.go
```diff
@@ -0,0 +1,268 @@
...
+func Logout(ctx context.Context, configProvider config.Provider, secretsProvider secrets.Provider) error {
+	cfg, err := configProvider.LoadOrCreateConfig()
+	if err != nil {
+		return fmt.Errorf("loading config: %w", err)
+	}
+	if err := validateOAuthConfig(cfg); err != nil {
+		return err
+	}
+
+	if ref := cfg.RegistryAuth.OAuth.CachedRefreshTokenRef; ref != "" {
```
jhrozek, line 106: **Suggestion (defense in depth)**: If `CachedRefreshTokenRef` is empty but a secret was stored under the derived key — possible when `updateConfigTokenRef` fails (it only logs a warning, does not return an error) — the orphaned secret won't be cleaned up here.

Consider also attempting `DeleteSecret` with `DeriveSecretKey(registryURL, issuer)` as a fallback, regardless of whether `CachedRefreshTokenRef` is set.

jhrozek (commented):
Review from Claude Code — 5 inline comments on code quality and correctness.

pkg/registry/auth/oauth_token_source.go
```diff
@@ -151,18 +151,31 @@ func (o *oauthTokenSource) buildOAuthFlowConfig(ctx context.Context) (*oauth.Con
 		callbackPort = remote.DefaultCallbackPort
 	}
 
+	scopes := ensureOfflineAccess(o.oauthCfg.Scopes)
```
jhrozek, line 154: **Pre-existing issue** (not introduced by this PR): `updateConfigTokenRef` (line 213) calls the global `config.UpdateConfig(...)` instead of an injected `config.Provider`. The token reference is always written to the default config file location, even when a non-default path is in use.

Filed as #4123.

pkg/registry/auth/login.go
```diff
@@ -0,0 +1,268 @@
...
+	default:
+		return fmt.Errorf("unsupported registry type %q for %q", registryType, opts.RegistryURL)
+	}
+	return nil
+}
+
+// ensureOAuthConfig checks whether OAuth auth is already configured and,
+// if not, attempts to configure it from the supplied opts (issuer + clientID).
+// Returns an actionable error when auth cannot be determined.
+func ensureOAuthConfig(
```
jhrozek, line 222: **Nit (duplication)**: `ensureOAuthConfig` largely duplicates `DefaultAuthManager.SetOAuthAuth` in `auth_manager.go` — same OIDC discovery with 10s timeout, same default scopes, same `UpdateConfig` with identical `RegistryAuth` struct. The only difference is the `context.Context` parameter.

Consider extracting a shared helper or having this delegate to `SetOAuthAuth` (enhanced to accept a context) to avoid the two implementations drifting.

pkg/registry/auth/login.go
```diff
@@ -0,0 +1,268 @@
...
+			ErrRegistryAuthRequired,
+		)
+	}
+	return nil
+}
+
+// checkMissingLoginConfig inspects the current config and opts, and returns a
+// single formatted error listing every flag the user still needs to provide.
+func checkMissingLoginConfig(cfg *config.Config, opts LoginOptions) error {
+	hasRegistry := cfg.RegistryApiUrl != "" || cfg.RegistryUrl != "" || cfg.LocalRegistryPath != ""
```
jhrozek: **Nit (duplication)**: This `hasRegistry` expression is duplicated verbatim at line 173 in `ensureRegistryURL`. Consider extracting a `hasRegistryConfig(cfg *config.Config) bool` helper to keep the check in one place.

pkg/registry/auth/login.go
```diff
@@ -0,0 +1,268 @@
...
+
+	if _, err := oauth.DiscoverOIDCEndpoints(discoveryCtx, opts.Issuer); err != nil {
+		return fmt.Errorf("OIDC discovery failed for issuer %s: %w", opts.Issuer, err)
+	}
+
+	scopes := func() []string {
+		if len(opts.Scopes) > 0 {
+			return opts.Scopes
+		}
+		return []string{"openid", "offline_access"}
```
jhrozek: **Nit (duplication)**: The default scopes `["openid", "offline_access"]` are hardcoded here, in `auth_manager.go:56`, and in `cmd/thv/app/config_registryauth.go:50`. Consider defining a shared `DefaultRegistryOAuthScopes` constant to avoid silent drift between the three code paths.

pkg/registry/auth/login.go
ChrisJBurns: Yep, good spot. This actually happened locally in testing too.

PR #4150, opened 2026-03-15
Set AuthStyle to InParams for public PKCE OAuth clients

### LINKED ISSUE #4149: OAuth PKCE token exchange fails with invalid_grant against strict OAuth 2.1 servers
## Bug description

Connecting ToolHive to a remote MCP server that uses a strict OAuth 2.1 implementation (such as Datadog's `mcp.datadoghq.com`) fails during the PKCE token exchange. The user completes the browser-based authorization successfully, but the token exchange silently fails.

When constructing `oauth2.Endpoint` structs, ToolHive does not set `AuthStyle`, which defaults to `oauth2.AuthStyleAutoDetect` (the zero value). Go's `golang.org/x/oauth2` library then uses the following auto-detect strategy:

1. **First attempt**: HTTP Basic Auth — sends `Authorization: Basic base64(client_id:)` (empty password for public clients)
2. **If that fails**: Retry with `client_id` in the POST body

For public PKCE clients registered with `token_endpoint_auth_method=none`, strict OAuth 2.1 servers reject the Basic Auth attempt — but **consume the single-use authorization code** in the process. The library's retry with `client_id` in the POST body then fails with `invalid_grant` because the code was already burned by the first attempt.

This affects three code paths where `oauth2.Endpoint` is constructed without `AuthStyle`:
- `pkg/auth/oauth/flow.go` — authorization code exchange (the primary failure path)
- `pkg/auth/remote/handler.go` — token refresh from cached tokens
- `pkg/registry/auth/oauth_token_source.go` — registry auth token refresh

## Steps to reproduce

1. Run a remote MCP server that uses a strict OAuth 2.1 implementation (e.g., Datadog):
   ```
   thv run --name datadog-mcp https://mcp.datadoghq.com/api/unstable/mcp-server/mcp
   ```
2. Complete the browser-based OAuth authorization when prompted
3. Observe the result of the token exchange

## Expected behavior

Token exchange succeeds on the first attempt by sending `client_id` in the POST body, matching the `token_endpoint_auth_method=none` registration.

## Actual behavior

Token exchange fails with:
```
invalid_grant: authorization code has already been used
```

The auto-detect logic sends HTTP Basic Auth first, which the strict server rejects while consuming the authorization code. The retry with the correct auth style fails because the code is single-use.

## Environment

- **ToolHive version**: Reproduced on v0.12.1
- **OS**: macOS (likely all platforms — the issue is in Go library behavior, not OS-specific)

gkatz2:
## Summary

ToolHive's OAuth PKCE token exchange fails against strict OAuth 2.1 servers (e.g., Datadog's `mcp.datadoghq.com`) because `oauth2.Endpoint` structs default to `AuthStyleAutoDetect`, which probes with HTTP Basic Auth first and burns the single-use authorization code. See the linked issue for the full mechanism.

- Set `AuthStyle: oauth2.AuthStyleInParams` in the three locations where `oauth2.Endpoint` is constructed without it
- Add a regression test with a strict mock server that rejects Basic Auth for public clients

Fixes #4149

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Manual testing: Verified OAuth PKCE flow completes successfully against Datadog's MCP server (`mcp.datadoghq.com`)

## Changes

| File | Change |
|------|--------|
| `pkg/auth/oauth/flow.go` | Set `AuthStyleInParams` on endpoint in `NewFlow` |
| `pkg/auth/remote/handler.go` | Set `AuthStyleInParams` on endpoint in `tryRestoreFromCachedTokens` |
| `pkg/registry/auth/oauth_token_source.go` | Set `AuthStyleInParams` on endpoint in `buildOAuth2Config` |
| `pkg/auth/oauth/flow_test.go` | Add `TestAuthStyleInParams_StrictPublicClientServer` regression test |

Yes. Remote MCP servers with strict OAuth 2.1 implementations that previously failed with `invalid_grant` during PKCE token exchange will now connect successfully. No configuration changes needed.

**Why `AuthStyleInParams` is safe**: `AuthStyleAutoDetect` tries HTTP Basic Auth first. For public PKCE clients (the common case — ToolHive's DCR registers with `token_endpoint_auth_method=none`), this sends an `Authorization: Basic base64(client_id:)` header with an empty password. Spec-compliant servers reject this and consume the single-use authorization code, making the retry fail with `invalid_grant`. `AuthStyleInParams` sends `client_id` in the POST body instead, avoiding the problem entirely.

For the less common case where users provide their own client credentials via `--remote-auth-client-id`/`--remote-auth-client-secret`, `AuthStyleInParams` sends the secret in the POST body (`client_secret_post` style). This is accepted by the vast majority of OAuth servers. The only servers that would break are those that exclusively require `client_secret_basic` and reject POST body credentials — an uncommon configuration, and a much smaller risk than the auth-code-burning bug this fixes.


pkg/auth/remote/handler.go
```diff
@@ -256,8 +256,9 @@ func (h *Handler) tryRestoreFromCachedTokens(
 		ClientSecret: clientSecret,
 		Scopes:       scopes,
 		Endpoint: oauth2.Endpoint{
-			AuthURL:  h.config.AuthorizeURL,
-			TokenURL: h.config.TokenURL,
+			AuthURL:   h.config.AuthorizeURL,
+			TokenURL:  h.config.TokenURL,
+			AuthStyle: oauth2.AuthStyleInParams,
```
amirejaz: This hardcodes `AuthStyleInParams` for all clients, but `resolveClientCredentials` (called on line 251) can return a non-empty `clientSecret` from DCR-registered confidential clients — the comment in that function explicitly says "may be empty for PKCE flows", implying it can also be non-empty.

Forcing `InParams` on a confidential client will break any OAuth server that mandates `client_secret_basic` (`AuthStyleInHeader`). RFC 6749 §2.3.1 allows both methods, but some enterprise IdPs enforce one exclusively.

Consider making the style conditional:

```go
authStyle := oauth2.AuthStyleInParams // safe for public PKCE clients
if clientSecret != "" {
    authStyle = oauth2.AuthStyleAutoDetect // let the library negotiate for confidential clients
}
```

For confidential clients, `AutoDetect` is safe from the auth-code-burning problem: the first Basic Auth attempt succeeds (valid credentials), so the code is never consumed by a failed probe.

pkg/auth/oauth/flow.go
```diff
@@ -124,8 +124,9 @@ func NewFlow(config *Config) (*Flow, error) {
 		RedirectURL:  redirectURL,
 		Scopes:       config.Scopes,
 		Endpoint: oauth2.Endpoint{
-			AuthURL:  config.AuthURL,
-			TokenURL: config.TokenURL,
+			AuthURL:   config.AuthURL,
+			TokenURL:  config.TokenURL,
+			AuthStyle: oauth2.AuthStyleInParams,
```
amirejaz: Same concern as `handler.go`: `config.ClientSecret` is documented as "optional for PKCE flow" (line 34), meaning it can also be non-empty for confidential clients. Hardcoding `InParams` here could break servers that require `client_secret_basic`.

Suggestion: apply the same conditional logic — use `InParams` only when `config.ClientSecret == ""`.

pkg/auth/oauth/flow_test.go
```diff
@@ -1204,3 +1204,117 @@ func TestProcessToken_ResourceTokenSourceSelection(t *testing.T) {
...
+//  3. The authorization code is single-use — consumed by the rejected first attempt
+//  4. The retry with client_id in POST body gets "invalid_grant" because the code
+//     was already burned
+func TestAuthStyleInParams_StrictPublicClientServer(t *testing.T) {
+	t.Parallel()
+
+	// Simulate a strict OAuth 2.1 server that rejects Basic Auth for public clients.
+	// This mirrors Datadog's behavior: public clients (token_endpoint_auth_method=none)
+	// MUST send client_id in the POST body, not via Authorization header.
+	requestCount := 0
```
amirejaz: `requestCount` is a plain `int` incremented inside the `httptest.Server` goroutine and read in the test goroutine without synchronisation. This is a data race that `go test -race` will flag.

```suggestion
	var requestCount atomic.Int32
```

Also update the increment on the next line to `requestCount.Add(1)` and the final assertion to `assert.Equal(t, int32(1), requestCount.Load(), ...)`. Don't forget to add `"sync/atomic"` to the imports.

amirejaz:
Hey @gkatz2, great fix! I pushed a small follow-up commit to your branch to address the conditional AuthStyle for confidential clients and a data race in the test.

PR #4154, opened 2026-03-15
Preserve query params in remote MCP server URLs

### LINKED ISSUE #4153: Remote MCP server URL query parameters are silently dropped, ignoring server configuration
## Bug Description

When registering a remote MCP server whose URL includes query parameters (e.g., `?toolsets=core,alerting,...`), those query parameters are not forwarded to the upstream server. Requests arrive at the remote server with only the base path — the query string is silently discarded. The server falls back to its default behavior as if no configuration was provided.

There is no error or warning — the server registers and starts successfully, making the data loss invisible to the user.

## Steps to Reproduce

1. Register a remote MCP server with query parameters that configure its behavior:
   ```bash
   thv run --name datadog \
     "https://mcp.datadoghq.com/api/unstable/mcp-server/mcp?toolsets=core,alerting,apm,cases,dbm,error-tracking,feature-flags,llmobs,product-analytics,networks,onboarding,security,software-delivery,synthetics" \
     --remote-auth
   ```
2. Connect an MCP client (e.g., Claude Code) to the registered server.
3. Inspect the available tools.

## Expected Behavior

All query parameters present in the registration URL are forwarded to the remote server on every request. For the Datadog example above, all explicitly requested toolsets are available — approximately 76 tools across the configured toolsets.

See the [Datadog MCP server documentation](https://docs.datadoghq.com/bits_ai/mcp_server) for how toolsets are selected via query parameters.

## Actual Behavior

The remote server receives requests with no query string and falls back to its default configuration. For the Datadog example, only the default toolset is returned — approximately 20 tools — regardless of which `toolsets` were specified in the registration URL.

## Environment

- ToolHive version: v0.12.1 (and current `main`)
- OS: macOS (code path is platform-independent; likely affects all platforms)
- Affects all remote MCP servers that rely on URL query parameters for configuration

## Additional Context

This affects any remote MCP server that uses query parameters to select capabilities, namespaces, or configuration at the URL level. The registration appears to succeed, so there is no indication to the user that their configuration is being ignored.

I plan to submit a PR to resolve this issue.

theJC:
## Summary

- **Why**: When a remote MCP server is registered with query parameters in the URL (e.g., `?toolsets=core,alerting,...`), ToolHive was silently dropping them before forwarding requests to the upstream server. The server fell back to its default behavior, returning fewer capabilities than configured — with no error or warning to indicate anything was wrong. For example, registering the [Datadog MCP server](https://docs.datadoghq.com/bits_ai/mcp_server) with explicit toolsets returned ~20 tools (default) instead of ~76 (configured).
- The transparent proxy was built to use only `scheme+host` as its `targetURI`, with the path extracted separately as `remoteBasePath`. Query parameters were never accounted for in this design and were discarded at parse time.
- Extract `RawQuery` from the registration URL alongside the path and forward it on every outbound request via a new `WithRemoteRawQuery` proxy option. Remote params are prepended before any client-supplied params (first-value-wins semantics). Raw string concatenation is used intentionally — `url.Values.Encode()` would percent-encode characters like commas that some APIs (e.g., Datadog's `toolsets` parameter) expect as literals.
- Also preserve query parameters in the URL returned to clients by `GenerateMCPServerURL` for SSE and streamable transports.
- Add `TransparentProxy.ListenerAddr()` to expose the bound address after start, enabling cross-package tests to use OS-assigned ports (port 0) without fixed port assignments.

Fixes #4153

## Type of change

- [x] Bug fix

## Test plan

Registered the Datadog MCP server with full toolsets query parameter and verified ~76 tools were available via Claude Code (vs ~20 before the fix).

## Changes

| File | Change |
|------|--------|
| `pkg/transport/http.go` | Extract `remoteRawQuery` from `remoteURL`; pass via `WithRemoteRawQuery` option |
| `pkg/transport/proxy/transparent/transparent_proxy.go` | Add `remoteRawQuery` field, `WithRemoteRawQuery` option, director merge logic, `ListenerAddr()` method |
| `pkg/transport/url.go` | Extract `generateRemoteMCPServerURL` helper; preserve `RawQuery` in generated client URLs |
| `pkg/transport/url_test.go` | Add 3 cases covering query params in streamable and SSE generated URLs |
| `pkg/transport/proxy/transparent/remote_path_test.go` | Add `TestRemoteQueryForwarding` (4 cases); migrate to `ListenerAddr()` |
| `pkg/transport/http_remote_query_test.go` | New: `TestHTTPTransport_Start_RemoteURLQueryParams` — covers the `Start()` extraction path |
| `test/e2e/remote_mcp_query_params_test.go` | New: E2E test using a local mock upstream to verify full CLI→proxy→upstream query forwarding |

Remote MCP servers registered with query parameters in the URL will now have those parameters forwarded to the upstream server on every request, honoring the configuration as specified.

The merge strategy in the proxy director prepends remote params before any client-supplied params (`remoteRawQuery + "&" + existingQuery`). This means if a client happens to send the same key, the server receives it twice — but most HTTP servers adopt first-value-wins semantics, so the operator-configured value takes precedence. The comment in the director explains this explicitly.

### DIFF
--- a/pkg/transport/url.go
+++ b/pkg/transport/url.go
@@ -66,7 +66,14 @@ func GenerateMCPServerURL(transportType string, proxyMode string, host string, p
 }
 
 // generateRemoteMCPServerURL builds the proxy URL for a remote MCP server,
-// preserving both the path and query parameters from the remote URL.
+// using only the path from the remote URL.
+//
+// Query parameters are intentionally excluded from the generated client URL.
+// The transparent proxy forwards them on every outbound request via
+// WithRemoteRawQuery, so including them here would cause duplication —
+// the upstream would receive the same parameter twice (e.g.
+// "toolsets=core&toolsets=core"). Clients connect to the clean proxy
+// URL; the proxy transparently appends the configured query string.
 func generateRemoteMCPServerURL(base, containerName, remoteURL string, isSSE, isStreamable bool) string {
 	targetURL, err := url.Parse(remoteURL)
 	if err != nil {
@@ -79,25 +86,15 @@ func generateRemoteMCPServerURL(base, containerName, remoteURL string, isSSE, is
 	if path == "/" {
 		path = ""
 	}
-	rawQuery := targetURL.RawQuery
 
 	if isSSE {
 		if path == "" {
 			path = ssecommon.HTTPSSEEndpoint
 		}
-		sseURL := appendQuery(fmt.Sprintf("%s%s", base, path), rawQuery)
-		return fmt.Sprintf("%s#%s", sseURL, url.PathEscape(containerName))
+		return fmt.Sprintf("%s%s#%s", base, path, url.PathEscape(containerName))
 	}
 	if isStreamable {
-		return appendQuery(fmt.Sprintf("%s%s", base, path), rawQuery)
+		return fmt.Sprintf("%s%s", base, path)
 	}
 	return ""
-}
-
-// appendQuery appends a raw query string to a URL if non-empty.
-func appendQuery(u, rawQuery string) string {
-	if rawQuery == "" {
-		return u
-	}
-	return u + "?" + rawQuery
 }
--- a/pkg/transport/proxy/transparent/transparent_proxy.go
+++ b/pkg/transport/proxy/transparent/transparent_proxy.go
@@ -525,9 +525,9 @@ func (p *TransparentProxy) Start(ctx context.Context) error {
 
 			// Merge query parameters from the remote URL into the outbound request.
 			// Remote params are prepended so they appear first; most HTTP servers
-			// adopt first-value-wins semantics for duplicate keys, so operator-
+			// adopt first-value-wins semantics for duplicate keys, ensuring operator
 			// configured values (e.g., toolsets=core,alerting) take precedence over
-			// any same-named params a client might send.
+			// any client-supplied params with the same key.
 			// Raw string concatenation is intentional: url.Values.Encode() would
 			// percent-encode characters like commas that some APIs expect as literals.
 			if p.remoteRawQuery != "" {
--- a/pkg/transport/url_test.go
+++ b/pkg/transport/url_test.go
@@ -166,34 +166,40 @@ func TestGenerateMCPServerURL(t *testing.T) {
 			expected:      "http://localhost:12345/api#test-container",
 		},
 		{
-			name:          "Streamable HTTP with query parameters in targetURI",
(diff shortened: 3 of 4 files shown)


pkg/transport/url.go
```diff
@@ -85,3 +64,40 @@ func GenerateMCPServerURL(transportType string, proxyMode string, host string, p
 
 	return ""
 }
+
+// generateRemoteMCPServerURL builds the proxy URL for a remote MCP server,
+// preserving both the path and query parameters from the remote URL.
+func generateRemoteMCPServerURL(base, containerName, remoteURL string, isSSE, isStreamable bool) string {
```
amirejaz, line 77: One thing to consider: if GenerateMCPServerURL includes query params in the client URL (e.g. http://localhost:12345/mcp?toolsets=core,alerting), the client will send those params on each request. The proxy then merges remoteRawQuery with the client’s query string, so the upstream receives:
`toolsets=core,alerting&toolsets=core,alerting`

Most servers use first-value-wins, so behavior is likely fine, but it’s redundant. Since the proxy already adds remoteRawQuery to every outbound request, the upstream gets the right params even when the client URL has no query string. Omitting query params from GenerateMCPServerURL would avoid this duplication while still forwarding them correctly. 

Alternatively, a short comment in the merge logic or in GenerateMCPServerURL documenting this behavior would help future readers.

theJC: Excellent feedback, thank you for your time reviewing.  

The latest commit I've added should hopefully address your comment.  You are right, this is much cleaner without this redundancy.

theJC: I believe the two failing tests are pre-existing flaky or environment failures and nothing to do with this code change... I dont believe I can can manually get github to re-run those tests.

theJC:
@amirejaz -

Can I please put in a plug for a review of https://github.com/stacklok/toolhive/pull/4150 ?

With this one that is now merged, and that one, once they are both released, that unblocks us from successfully using Toolhive with the Datadog MCP server.

Once both are delivered and we are able to use Toolhive and Datadog's MCP server together, I will do my best to push DD to add Toolhive to their documented client list w/directions as well.

amirejaz:
@theJC A new THV release has been deployed, including #4154 and #4150.

amirejaz:
Hey @theJC, just wanted to check in and see how things are going with Toolhive + the Datadog MCP server.

Really appreciate you pushing on the Datadog side. If there’s anything you need from us to help move things along.

Would be great to see this land 🙂

PR #4157, opened 2026-03-16
Add notifications/elicitation/complete to MCP parser

JAORMX:
## Summary

- The MCP 2025-11-25 spec defines `notifications/elicitation/complete` as a server-to-client notification sent when an out-of-band URL-mode elicitation completes. This was the only spec-defined method not yet handled by the parser (30 of 31 covered).
- Adds a handler that extracts `elicitationId` as the resource identifier, plus tests for the middleware round-trip and extraction logic.

## Type of change

- [x] New feature

## Test plan

### DIFF
--- a/pkg/mcp/parser.go
+++ b/pkg/mcp/parser.go
@@ -184,30 +184,31 @@ type methodHandler func(map[string]interface{}) (string, map[string]interface{})
 
 // methodHandlers maps MCP methods to their respective handlers
 var methodHandlers = map[string]methodHandler{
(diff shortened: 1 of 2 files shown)


jhrozek (approved):
LGTM — spec-compliant, well-tested, completes coverage of all 31 MCP 2025-11-25 methods.

While reviewing, noticed a pre-existing issue: `progress/update` in the `methodHandlers` map is not a valid MCP spec method (only `notifications/progress` is). Filed #4170 to track cleanup.

PR #4165, opened 2026-03-16
Bring composite tools into session abstraction

### LINKED ISSUE #3872: [vMCP] Remove old code path after Phase 2 validation (Phase 3)

yrobla:
## Summary

Composite tool workflow engines were previously relying on the discovery middleware to inject DiscoveredCapabilities into the request context so that the shared stateless router could route backend tool calls within workflows. This created an implicit coupling between the middleware and composite tool execution that made unit-testing harder and was a source of integration bugs.

Fixes #3872 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

Affected components: pkg/vmcp/router, pkg/vmcp/composer, pkg/vmcp/server, pkg/vmcp/discovery

| File | Change |
|------|--------|
|      |        |

## Large PR Justification

This is covering an isolated functionality with complete testing. Cannot be split.


pkg/vmcp/composer/workflow_engine.go
```diff
@@ -91,6 +97,23 @@ func NewWorkflowEngine(
...
+
+// EngineOption configures a workflowEngine after construction.
+// Use the provided With* helpers rather than implementing this directly.
+type EngineOption func(*workflowEngine)
+
+// WithTools binds a resolved tool list to the engine for argument type coercion.
+// Use this when creating per-session engines via router.NewSessionRouter.
+func WithTools(tools []vmcp.Tool) EngineOption {
+	return func(e *workflowEngine) { e.tools = tools }
 }
```
jerm-dro: blocker: Are tools are always provided? If so, they should use an arg rather than option pattern.

pkg/vmcp/server/server.go
```diff
@@ -336,23 +342,34 @@ func New(
...
+		return composer.NewWorkflowEngine(
+			router.NewSessionRouter(sessionRT), backendClient, elicitationHandler, stateStore, workflowAuditor,
+			composer.WithTools(sessionTools),
+		)
+	}
+
+	// Validate workflows (fail fast on invalid definitions)
 	var err error
-	workflowDefs, workflowExecutors, err = validateAndCreateExecutors(workflowComposer, workflowDefs)
+	workflowDefs, err = validateWorkflows(workflowComposer, workflowDefs)
```
jerm-dro, line 365: Suggestion: add composite tools to the session concept. On the surface, they seem unrelated. However, the composite tools that are available should be determined on a session-by-session basis. For example, depending on the identity / claims of a user, they may not have authorization to use all underlying tools in a composite tool, and thus shouldn't have access to a composite tool either.

pkg/vmcp/session/compositetools/decorator.go
```diff
@@ -0,0 +1,113 @@
...
+// CallTool dispatches composite tool names to their workflow executors.
+// Unknown names are delegated to the embedded session.
+func (d *compositeToolsDecorator) CallTool(
+	ctx context.Context,
+	caller *auth.Identity,
+	toolName string,
+	arguments map[string]any,
+	meta map[string]any,
+) (*vmcp.ToolCallResult, error) {
+	if exec, ok := d.executors[toolName]; ok {
```
jerm-dro, line 83: suggestion: to avoid nesting do the else branch first.

```go

exec, ok := exec, ok := d.executors[toolName]
if !ok {
    return d.MultiSession.CallTool(ctx, caller, toolName, arguments, meta)
}
// else do the composite tools path.
```

pkg/vmcp/server/server.go
```diff
@@ -1020,134 +1050,131 @@ func (s *Server) handleSessionRegistrationImpl(ctx context.Context, session serv
...
+// calls DecorateSession. The optimizer replaces the tool list with find_tool + call_tool.
+func (s *Server) applyOptimizerDecorator(ctx context.Context, sessionID string) error {
+	// Snapshot the pre-optimizer SDK tools for the optimizer to index.
+	// This includes composite tools if the composite decorator was already applied.
+	sdkTools, err := s.vmcpSessionMgr.GetAdaptedTools(sessionID)
+	if err != nil {
+		return fmt.Errorf("failed to get tools for optimizer: %w", err)
 	}
 
-	if len(adaptedTools) > 0 {
```
jerm-dro, line 1095: Two suggestions that would be nice to address in a later PR, or at least note in a comment: 
1. If we pulled the `optimizerFactory` into the session factory, then we could completely remove the optimizer code from `server.go`. The optimizer is baked into the session construction and is not a server concern.
2. Once the above is done and you can construct the session all at once, which would allow you delete the `DecorateSession` method.

At a high level, I'm imagining:
1. The session factory function is built at startup time based on config. If there's an optimizer configured, the factory will return a MultiSession that's decorated with an optimizer.
2. At runtime, the factory is unconditionally called and you always get a `MultiSession`.

pkg/vmcp/server/server.go
```diff
@@ -1020,134 +1050,131 @@ func (s *Server) handleSessionRegistrationImpl(ctx context.Context, session serv
...
+	compositeToolsMeta := convertWorkflowDefsToTools(sessionDefs)
+	if err := validateNoToolConflicts(multiSess.Tools(), compositeToolsMeta); err != nil {
+		slog.Warn("composite tool name conflict detected; skipping composite tools",
 			"session_id", sessionID,
 			"error", err)
-		return nil, nil
+		return nil
 	}
 
-	sdkTools, err := s.capabilityAdapter.ToCompositeToolSDKTools(compositeTools, s.workflowExecutors)
```
jerm-dro, line 1061: Similar to below, let's pull this into the session factory.

pkg/vmcp/server/composite_tool_converter.go
```diff
@@ -7,12 +7,95 @@ package server
...
 import (
 	"fmt"
+	"strings"
 
 	"github.com/stacklok/toolhive/pkg/vmcp"
 	"github.com/stacklok/toolhive/pkg/vmcp/composer"
 	"github.com/stacklok/toolhive/pkg/vmcp/config"
 )
 
+// filterWorkflowDefsForSession returns only the workflow definitions whose every
```
jerm-dro, line 17: non-blocking: let's move this and the related tests out of server/ into a lower level package.

This could wait until the session factory refactor mentioned below.

pkg/vmcp/server/adapter/optimizer_adapter.go
```diff
@@ -3,124 +3,10 @@
...
-	"github.com/mark3labs/mcp-go/server"
-
-	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
-	"github.com/stacklok/toolhive/pkg/vmcp/schema"
-)
-
-// OptimizerToolNames defines the tool names exposed when optimizer is enabled.
+// OptimizerToolNames defines the tool names exposed when optimizer mode is enabled.
+// These constants are kept here for backwards compatibility with existing tests and
+// callers. The actual tool implementation lives in the optimizerdec session decorator.
```
jerm-dro, line 8: Can this file be entirely deleted?

PR #4167, opened 2026-03-16
fix(vmcp): exempt SSE connections from WriteTimeout to prevent drops

### LINKED ISSUE #3691: Remove WriteTimeout from vMCP server to prevent SSE connection drops
## Summary

Go's `http.Server.WriteTimeout` sets an absolute deadline on the TCP connection when request headers are read. For long-lived SSE connections used by the MCP Streamable HTTP transport, this causes the stream to be killed after the timeout (currently 30s) regardless of write activity.

This was originally part of #3624 but was split out per reviewer feedback to keep that PR focused on header validation.

## Problem

`WriteTimeout` applies to the entire response duration, not per-write. Since MCP Streamable HTTP uses SSE streams that remain open indefinitely, the connection is terminated after 30s even if data is actively flowing. See [golang/go#16100](https://github.com/golang/go/issues/16100) for confirmation from Go maintainers.

## Proposed solution

- Remove `WriteTimeout` from the `http.Server` config
- Explore using `http.ResponseController` for per-request write deadlines, so non-SSE requests still have timeout protection

Related PR: #3624

yrobla:
## Summary

Go's `http.Server.WriteTimeout` sets an absolute deadline on the entire response duration. For long-lived SSE connections used by the MCP Streamable HTTP transport, this caused streams to be killed after 30s regardless of write activity (see golang/go#16100).

**Approach:** Keep `http.Server.WriteTimeout` (it protects all routes from slow/hung non-SSE clients), and add `transportmiddleware.WriteTimeout` in `pkg/transport/middleware` that uses `http.ResponseController.SetWriteDeadline(time.Time{})` to clear the write deadline for qualifying SSE connections only. A connection qualifies when all three hold: method is GET, `Accept` header contains `text/event-stream`, and path matches the MCP endpoint exactly. Non-qualifying requests are left completely untouched — `http.Server.WriteTimeout` remains in effect for them.

Tests added:
- Unit tests verifying `SetWriteDeadline` is called (qualifying SSE) / not called (POST, DELETE, GET without SSE headers, GET on wrong path)
- TCP-level integration test confirming an SSE stream survives past `WriteTimeout` (asserts both elapsed time ≥ stream duration and event count > what's possible before timeout fires)
- vMCP server integration tests verifying the middleware is correctly wired end-to-end

Fixes #3691

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/transport/middleware/write_timeout.go` | New: `WriteTimeout` middleware — clears write deadline for qualifying SSE connections only |
| `pkg/transport/middleware/write_timeout_test.go` | New: unit + TCP-level integration tests |
| `pkg/vmcp/server/server.go` | Apply `transportmiddleware.WriteTimeout` to MCP handler; keep `http.Server.WriteTimeout` for all other routes |
| `pkg/vmcp/server/write_timeout_integration_test.go` | New: vMCP server integration tests for SSE write timeout behaviour |
| `pkg/vmcp/server/session_management_realbackend_integration_test.go` | Refactor: extract `newRealTestHandler` helper to allow `httptest.Server` configuration in tests |

No — this is an internal fix. SSE streams that were silently dropped after 30s will now remain open.

The middleware only touches qualifying SSE requests (GET + `Accept: text/event-stream` + matching endpoint path). Non-qualifying requests are intentionally left untouched — resetting the deadline to `time.Now().Add(defaultTimeout)` would extend it (since `time.Now()` is later than connection-accept time), weakening the protection for health, metrics, and JSON-RPC POST routes. If `SetWriteDeadline` fails on a qualifying SSE connection (e.g., a wrapping `ResponseWriter` that doesn't implement the interface), a `Warn` log is emitted so the gap is visible in observability tooling.

### DIFF
--- a/pkg/vmcp/server/server.go
+++ b/pkg/vmcp/server/server.go
@@ -27,6 +27,7 @@ import (
 	mcpparser "github.com/stacklok/toolhive/pkg/mcp"
 	"github.com/stacklok/toolhive/pkg/recovery"
 	"github.com/stacklok/toolhive/pkg/telemetry"
+	transportmiddleware "github.com/stacklok/toolhive/pkg/transport/middleware"
 	transportsession "github.com/stacklok/toolhive/pkg/transport/session"
 	"github.com/stacklok/toolhive/pkg/vmcp"
 	"github.com/stacklok/toolhive/pkg/vmcp/composer"
@@ -47,7 +48,10 @@ const (
 	// defaultReadTimeout is the maximum duration for reading the entire request, including body.
 	defaultReadTimeout = 30 * time.Second
 
-	// defaultWriteTimeout is the maximum duration before timing out writes of the response.
+	// defaultWriteTimeout is the server-level write deadline set on http.Server.WriteTimeout.
+	// It protects all routes (health, metrics, well-known, etc.) from slow-write clients.
+	// For qualifying SSE (GET) connections, transportmiddleware.WriteTimeout clears this
+	// per-request via http.ResponseController.SetWriteDeadline(time.Time{}) (golang/go#16100).
 	defaultWriteTimeout = 30 * time.Second
 
 	// defaultIdleTimeout is the maximum amount of time to wait for the next request when keep-alive's are enabled.
@@ -556,6 +560,13 @@ func (s *Server) Handler(_ context.Context) (http.Handler, error) {
 
 	// Apply Accept header validation (rejects GET requests without Accept: text/event-stream)
 	mcpHandler = headerValidatingMiddleware(mcpHandler)
+
+	// Clear the write deadline for qualifying SSE connections (GET +
+	// Accept: text/event-stream + MCP endpoint path) so the server-level
+	// WriteTimeout does not kill long-lived SSE streams (see golang/go#16100).
+	// Non-qualifying requests are left untouched; http.Server.WriteTimeout
+	// (defaultWriteTimeout) remains in effect for them.
+	mcpHandler = transportmiddleware.WriteTimeout(s.config.EndpointPath)(mcpHandler)
 
 	// Apply recovery middleware as outermost (catches panics from all inner middleware)
 	mcpHandler = recovery.Middleware(mcpHandler)
--- a/pkg/vmcp/server/write_timeout_integration_test.go
+++ b/pkg/vmcp/server/write_timeout_integration_test.go
@@ -1,0 +1,112 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package server_test
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net/http"
+	"net/http/httptest"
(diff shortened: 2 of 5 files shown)


pkg/vmcp/server/server.go
```diff
@@ -1247,6 +1256,35 @@ var notAcceptableBody = []byte(
...
+// However, it also kills long-lived SSE (GET) streams after 30s regardless of write
+// activity (see golang/go#16100). For MCP GET connections we clear the write deadline
+// by setting it to zero (no deadline), overriding the server-level timeout. All other
+// methods (POST, DELETE, etc.) keep the server's write timeout unchanged.
+func writeTimeoutMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodGet {
+			rc := http.NewResponseController(w)
+			// time.Time{} (zero value) means no deadline, overriding the server WriteTimeout.
+			if err := rc.SetWriteDeadline(time.Time{}); err != nil {
```
jerm-dro: Blocker: this unconditionally erases the timeout for all GET requests, regardless of if they're SSE or not?

Can you make an implementation of this `writeTimeoutMiddleware` that's a bit more complete. For example, you'd test:
- for all non-qualifying requests, a preconfigured default timeout is used.
- for all qualifying requests, there is no timeout.
- what qualifies for exemption? It should be some combination of headers, path, etc.

pkg/vmcp/server/server.go
```diff
@@ -1247,6 +1256,35 @@ var notAcceptableBody = []byte(
...
 )
 
+// writeTimeoutMiddleware exempts SSE connections from the server-level write timeout.
+//
+// http.Server.WriteTimeout protects all routes including health and metrics endpoints.
+// However, it also kills long-lived SSE (GET) streams after 30s regardless of write
+// activity (see golang/go#16100). For MCP GET connections we clear the write deadline
+// by setting it to zero (no deadline), overriding the server-level timeout. All other
+// methods (POST, DELETE, etc.) keep the server's write timeout unchanged.
+func writeTimeoutMiddleware(next http.Handler) http.Handler {
```
jerm-dro: Suggestion: pull this out into a separate middleware package. We don't need to bloat `server.go` with the implementation details.

pkg/vmcp/server/write_timeout_integration_test.go
```diff
@@ -0,0 +1,112 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package server_test
```
jerm-dro, line 4: Suggestion: delete this test. It is redundant with the new `write_timeout_test.go`.

yrobla: ok will do in a follow up

PR #4168, opened 2026-03-16
Allow MCPRemoteProxy to work without upstream or client auth

aron-muon:
## Summary

MCPRemoteProxy always injected upstream IdP tokens when the embedded auth server was configured, and the transparent proxy always emitted `X-Forwarded-Host` to remote upstreams. Two consequences:

1. **Public/anonymous upstreams couldn't be proxied with client OAuth.** Operators who want to authenticate clients (Okta, GitHub, etc.) but proxy a public MCP server (e.g. documentation APIs) had no way to skip the upstream token injection.
2. **Cross-host redirect loops.** Upstreams that build redirect URLs from `X-Forwarded-Host` would echo the proxy's hostname back, creating loops the transparent proxy then refused to follow ("refusing cross-host redirect from remote MCP server").

This PR addresses both with two focused changes per @ChrisJBurns's review (#4168 (review)).

**What changed (2 logical changes):**

1. **`disableUpstreamTokenInjection` flag on `EmbeddedAuthServerConfig`.** When `true`, the embedded auth server still runs the OAuth flow for clients, but after the client JWT is validated a `strip-auth` middleware (`pkg/transport/middleware/strip_auth.go`) removes the client's credential headers — `Authorization`, `Cookie`, and `Proxy-Authorization`, mirroring the set net/http refuses to copy across cross-host redirects — so the backend receives an unauthenticated request. Combining the flag with token exchange or AWS STS is rejected at config-population time, since those would re-add credentials after the strip; `headerForward` remains available for attaching static credentials. Ordering (auth validates before strip-auth removes) is pinned by a dedicated test.

2. **Drop `X-Forwarded-Host` for remote upstreams in the transparent proxy.** `SetXForwarded()` still runs, so remote backends keep receiving `X-Forwarded-For`/`-Proto` for IP logging and rate limiting; only `X-Forwarded-Host` — the header implicated in the redirect loop — is removed. Covered by a unit test for both the local and remote paths.

**Why the two changes ship together:** they are the two halves of the same capability — proxying a public third-party MCP server with client-side OAuth. The flag removes the auth blocker (injected tokens → 401s from a public upstream); the `X-Forwarded-Host` fix removes the transport blocker (redirect loops) hit immediately against the same class of upstreams. Each was verified against the same deployment; the transport fix is now scoped to a single header deletion.

**Dropped from earlier revisions of this PR:**

- ~~Optional `oidcConfig`~~ — @ChrisJBurns confirmed this is already resolved on main: both `OIDCConfig` and `OIDCConfigRef` are pointer types and `resolveAndAddOIDCConfig` returns `(nil, nil)` cleanly when both are absent.
- ~~Custom `forwardFollowingRedirects`~~ — main now has a more secure redirect-following implementation with cross-host and HTTPS-downgrade guards.

**Up to date with main** — latest merge resolved adjacent-addition conflicts against the CIMD and `baselineClientScopes` work (both sides kept); generated artifacts differ from main only by the new field.

**Related issues:** the multi-agent review also surfaced a pre-existing token-passthrough in the upstreamswap `custom` header strategy, tracked separately as #5504.

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests — full suite via `task test`; new coverage: `TestStripAuthMiddleware` (factory/handler/Close, all three headers), `TestPopulateMiddlewareConfigs_StripAuthOrdering` (auth-before-strip invariant), `TestPopulateMiddlewareConfigs_StripAuthConflicts` (token exchange / AWS STS rejection), `TestSetXForwardedHeaders` (local vs remote X-Forwarded-Host)
- [x] Build verification — `task build` and `task build-operator`
- [x] Generated artifacts — `task operator-generate`, `task operator-manifests`, `task crdref-gen`, `task docs` (controller-gen v0.17.3); committed

## API Compatibility

- [x] This PR does not break the `v1beta1` API. The added field (`disableUpstreamTokenInjection`) is optional with a `false` default, matching the existing behavior. The new config-population error applies only to configurations combining the new flag with token exchange or AWS STS, which previously misbehaved silently.

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go` | Add `DisableUpstreamTokenInjection` to `EmbeddedAuthServerConfig`, documenting the stripped header set |
| `cmd/thv-operator/pkg/controllerutil/authserver.go` | Wire CRD field through to runtime `RunConfig` |
| `cmd/thv-operator/pkg/controllerutil/authserver_test.go` | Tests for wiring + default value |
| `pkg/authserver/config.go` | Add `DisableUpstreamTokenInjection` to runtime `RunConfig` |
| `pkg/transport/middleware/strip_auth.go` (+ test) | New strip-auth middleware: removes `Authorization`, `Cookie`, `Proxy-Authorization` |
| `pkg/runner/middleware.go` | Select strip-auth instead of upstream-swap when injection disabled; reject token-exchange/AWS-STS combinations |
| `pkg/runner/middleware_test.go` | Branch-selection, ordering-invariant, and conflict-rejection tests |
| `pkg/transport/proxy/transparent/transparent_proxy.go` (+ `xforwarded_test.go`) | Keep `SetXForwarded()`, delete only `X-Forwarded-Host` for remote upstreams; debug logging |
| `deploy/charts/operator-crds/{files,templates}/...` | Regenerated CRD YAMLs (controller-gen v0.17.3) |
| `docs/operator/crd-api.md`, `docs/server/{docs.go,swagger.{json,yaml}}` | Regenerated docs |

Yes. One new capability for `MCPExternalAuthConfig` (type `embeddedAuthServer`):

- **`spec.embeddedAuthServer.disableUpstreamTokenInjection: true`** — authenticates MCP clients via the embedded OAuth server, then strips the client's credential headers (`Authorization`, `Cookie`, `Proxy-Authorization`) so the backend receives an unauthenticated request. Useful for proxying public MCP servers (documentation, search, etc.) where you want client auth but no upstream auth. Use `headerForward` for static upstream credentials; combining with token exchange or AWS STS is rejected at startup.

The X-Forwarded-Host change is a bug fix; no API surface change.


ChrisJBurns:
Hey @aron-muon  thanks for the PR. Apologies for the delay on response, we were actually discussing internally the direction of the `MCPRemoteProxy` as we've noticed that there are new user cases coming up that it doesn't the purpose to solve. I'll be submitting an RFC soon detail this change. I'll leave this PR open for now just so we can come back to it when we do the refactoring 🚀

ChrisJBurns:
@aron-muon  hey bud, just got back from leave, for reference this [RFC](https://github.com/stacklok/toolhive-rfcs/pull/55) has been raised that aims to provide a new CRD that will eventually lead to the deprecation and removal of `MCPRemoteProxy`

aron-muon:
> @aron-muon hey bud, just got back from leave, for reference this [RFC](https://github.com/stacklok/toolhive-rfcs/pull/55) has been raised that aims to provide a new CRD that will eventually lead to the deprecation and removal of `MCPRemoteProxy`

Very exciting!

lujunsan:
@claude review the changes and check if this overlaps with the MCPOIDCConfig CRD changes that have been added recently.

aron-muon:
> @claude review the changes and check if this overlaps with the MCPOIDCConfig CRD changes that have been added recently.

Cleanly rebased now! @lujunsan

ChrisJBurns:
Hey @aron-muon, thanks for keeping this rebased! I did a deep dive into the current state of `main` to see which parts of this PR are still needed.

**Goal 2 (optional `oidcConfig` / anonymous proxy) is already resolved on main.** Both `OIDCConfig` (which is being removed soon in favour of the dedicated CRD) and `OIDCConfigRef` are optional pointers, and `resolveAndAddOIDCConfig()` gracefully returns `(nil, nil)` when both are absent. The OIDC resolver, `AddOIDCConfigOptions`, and controller validation all handle the nil case without error. So that piece can be dropped.

**Goals 1 and 3 are still unresolved:**
- **`DisableUpstreamTokenInjection`** — no equivalent exists. There's currently no way to run the embedded auth server for client-facing OAuth without also injecting tokens upstream.
- **`SetXForwarded()` skipping for remote upstreams** — still called unconditionally in `transparent_proxy.go`. The codebase has workarounds for specific symptoms (AWS SigV4, SSE URL rewriting), but the root cause remains.

It's also worth noting that for users who don't need the proxy pod at all — they just want vMCP to call a remote server with optional auth — `MCPServerEntry` + `VirtualMCPServer` already avoids all of these issues since there's no intermediate proxy setting X-Forwarded headers or injecting tokens. Goals 1 and 3 are specifically relevant when you *need* the proxy pod (e.g., for client-facing embedded auth server OAuth flows).

Would you be open to splitting this into a focused PR covering just goals 1 and 3? That would make it smaller and easier to review. Happy to help coordinate.

reyortiz3:
hi @aron-muon! 👋 just checking on this PR if there is any other help needed. Thanks!

aron-muon:
> Hey @aron-muon, thanks for keeping this rebased! I did a deep dive into the current state of `main` to see which parts of this PR are still needed.
> 
> **Goal 2 (optional `oidcConfig` / anonymous proxy) is already resolved on main.** Both `OIDCConfig` (which is being removed soon in favour of the dedicated CRD) and `OIDCConfigRef` are optional pointers, and `resolveAndAddOIDCConfig()` gracefully returns `(nil, nil)` when both are absent. The OIDC resolver, `AddOIDCConfigOptions`, and controller validation all handle the nil case without error. So that piece can be dropped.
> 
> **Goals 1 and 3 are still unresolved:**
> 
> * **`DisableUpstreamTokenInjection`** — no equivalent exists. There's currently no way to run the embedded auth server for client-facing OAuth without also injecting tokens upstream.
> * **`SetXForwarded()` skipping for remote upstreams** — still called unconditionally in `transparent_proxy.go`. The codebase has workarounds for specific symptoms (AWS SigV4, SSE URL rewriting), but the root cause remains.
> 
> It's also worth noting that for users who don't need the proxy pod at all — they just want vMCP to call a remote server with optional auth — `MCPServerEntry` + `VirtualMCPServer` already avoids all of these issues since there's no intermediate proxy setting X-Forwarded headers or injecting tokens. Goals 1 and 3 are specifically relevant when you _need_ the proxy pod (e.g., for client-facing embedded auth server OAuth flows).
> 
> Would you be open to splitting this into a focused PR covering just goals 1 and 3? That would make it smaller and easier to review. Happy to help coordinate.

Updated with just goal 1 and 3!

aron-muon:
> hi @aron-muon! 👋 just checking on this PR if there is any other help needed. Thanks!

Ready for re-review, sorry for the delay!

aron-muon:
Merged latest main — resolved the adjacent-addition conflicts with the CIMD / `baselineClientScopes` work (both kept), and added direct test coverage for the strip-auth middleware factory and handler to close the codecov patch-coverage gap on `pkg/runner/middleware.go`. Ready for re-review.

pkg/runner/middleware.go
```diff
@@ -398,6 +407,45 @@ func injectUpstreamProviderIfNeeded(
...
+}
+
+// stripAuthMiddleware removes the Authorization header from requests.
+type stripAuthMiddleware struct{}
+
+// Handler returns the middleware function.
+func (*stripAuthMiddleware) Handler() types.MiddlewareFunction {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			r.Header.Del("Authorization")
```
ChrisJBurns: **strip-auth removes only `Authorization` — other client credential headers still reach the public upstream.**

`ExtractBearerToken` confirms the ToolHive JWT only ever rides in `Authorization`, so the PR's *stated* goal (don't leak the ToolHive JWT) is met. But this feature targets *untrusted public* upstreams, and a client can also carry credentials in `Cookie` or `Proxy-Authorization`, which the transparent proxy forwards verbatim. Either strip those too, or document explicitly that only the bearer token is stripped so operators don't over-trust the isolation. The existing test asserting `X-Custom` survives encodes the narrow contract.

*Security + OAuth reviewers — consensus 8/10.*

```suggestion
			r.Header.Del("Authorization")
			r.Header.Del("Cookie")
			r.Header.Del("Proxy-Authorization")
```

pkg/runner/middleware.go
```diff
@@ -398,6 +407,45 @@ func injectUpstreamProviderIfNeeded(
 	return cedar.InjectUpstreamProvider(authzCfg, providerName)
 }
 
+// stripAuthMiddlewareType is the type identifier for the auth header stripping middleware.
+const stripAuthMiddlewareType = "strip-auth"
```
ChrisJBurns: **Define this middleware in its own package, matching every other middleware.**

Each entry in `GetSupportedMiddlewareFactories` lives in its own package exposing an exported `MiddlewareType` + `CreateMiddleware` (`upstreamswap.MiddlewareType`, `headerfwd` in `header_forward.go`, `awssts`, `validating`, `mutating`). `pkg/runner` is the wiring/bridge layer, not an implementation home — this is the only inline middleware. `header_forward.go` is the precedent: an equally trivial header middleware that still got its own file/package. Suggest extracting to e.g. `pkg/transport/middleware/stripauth` with exported `MiddlewareType = "strip-auth"` and `CreateMiddleware`, then register it like the others; keep only the `addAuthHeaderStripMiddleware` wiring helper here.

*Go/architecture reviewer — consensus 8/10.*

pkg/runner/middleware.go
```diff
@@ -343,6 +346,12 @@ func addUpstreamSwapMiddleware(
 		return middlewares, nil
 	}
 
+	// When upstream token injection is disabled, strip the Authorization header
+	// so the client's ToolHive JWT doesn't leak to the upstream server.
+	if config.EmbeddedAuthServerConfig.DisableUpstreamTokenInjection {
+		return addAuthHeaderStripMiddleware(middlewares)
```
ChrisJBurns, line 400: **Pin the strip-auth placement & ordering invariants.**

Two related points on where strip-auth sits in the chain:

1. It must run *after* the auth middleware validates the JWT. It does today (auth is appended first, applied outermost), but nothing asserts it — a future reorder would silently break client auth. Add an ordering-assertion test (strip-auth index > auth index) and a comment stating the invariant.
2. strip-auth is added in the `upstream-swap` *slot* (outer), while token-injecting middleware (token-exchange / AWS-STS / OBO) run *later*, closer to the backend. If `DisableUpstreamTokenInjection` is ever combined with one of those, they could re-add an `Authorization` header *after* the strip, defeating it. Consider stripping closest to the backend, or rejecting that config combination at build time.

*OAuth + security + general reviewers — consensus 7/10.*

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -1026,7 +1041,15 @@ func (p *TransparentProxy) Start(ctx context.Context) error {
...
 		Rewrite: func(pr *httputil.ProxyRequest) {
 			pr.SetURL(targetURL)
-			pr.SetXForwarded()
+
+			// Only set X-Forwarded-* headers for local backends.
+			// For remote upstreams, these headers leak the proxy's hostname
+			// (X-Forwarded-Host) to third-party servers, which can cause
+			// 307 redirect loops when the upstream uses that header to
+			// construct redirect URLs pointing back to the proxy.
+			if !p.isRemote {
```
ChrisJBurns: **Scrub `X-Forwarded` surgically, and add a test.**

Skipping `SetXForwarded()` wholesale fixes the `X-Forwarded-Host` leak / 307 loop, but it also (a) stops `httputil` from scrubbing *client-supplied* `X-Forwarded-*` headers — a spoofed host can still pass through to the upstream — and (b) drops `X-Forwarded-For`/`-Proto`, which a remote backend may legitimately use for IP logging/rate limiting. Only `X-Forwarded-Host` is implicated in the redirect loop. Prefer:

```go
pr.SetXForwarded()
if p.isRemote {
    pr.Out.Header.Del("X-Forwarded-Host")
}
```

This behavioral change also has no test — add one asserting a remote upstream omits `X-Forwarded-Host` while a local backend still sets it.

*Security + general reviewers — consensus 8/10.*

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -279,6 +279,16 @@ type EmbeddedAuthServerConfig struct {
...
 
+	// DisableUpstreamTokenInjection prevents the embedded auth server from injecting
+	// upstream IdP tokens into requests forwarded to the backend MCP server.
+	// When true, the embedded auth server still handles OAuth flows for clients
+	// but does not swap ToolHive JWTs for upstream tokens on outgoing requests.
+	// This is useful when the backend MCP server does not require authentication
+	// (e.g., public documentation servers) but you still want client authentication.
+	// +kubebuilder:default=false
+	// +optional
+	DisableUpstreamTokenInjection bool `json:"disableUpstreamTokenInjection,omitempty"`
```
ChrisJBurns, line 295: **Document that enabling this flag strips the `Authorization` header.**

This comment (and the `RunConfig` comment at `pkg/authserver/config.go:98`, plus the generated CRD/swagger text) describe only *"does not swap ToolHive JWTs for upstream tokens."* The load-bearing, security-relevant behavior is that the client's `Authorization` header is **removed** before forwarding — the backend receives an unauthenticated request. A reader could reasonably assume the bearer token passes through untouched. Add a sentence to both comments stating the header is stripped, and regenerate the CRD/swagger docs.

*All 5 reviewers touched this — consensus 9/10.*

ChrisJBurns (commented):
## Multi-agent review — PR #4168

Reviewed by 5 specialist agents (security, OAuth/token-flow, Go & architecture, operator/CRD, general/tests). **No HIGH-severity findings.** The design is sound, opt-in (`disableUpstreamTokenInjection` defaults `false`), backward-compatible, well-tested, and the validate-JWT-before-strip ordering is correct. The items below are about completeness of credential stripping, middleware placement, and docs.

### Consensus findings

| # | Finding | Severity | Score | Where |
|---|---------|----------|-------|-------|
| F4 | Docs omit that enabling the flag *strips* the `Authorization` header | MEDIUM | 9/10 | `mcpexternalauthconfig_types.go:290`, `config.go:98` |
| F1 | strip-auth removes only `Authorization`; `Cookie`/`Proxy-Authorization` still leak to public upstream | MEDIUM | 8/10 | `middleware.go:440` |
| F2 | `SetXForwarded()` skipped wholesale for remote — leaves client `X-Forwarded-*` unscrubbed, drops `X-Forwarded-For` | MEDIUM | 8/10 | `transparent_proxy.go:1050` |
| F3 | strip-auth middleware defined inline in the wiring layer, not its own package | MEDIUM | 8/10 | `middleware.go:411` |
| F5/F6 | strip-auth placement/ordering invariants not pinned (after auth; before downstream token injectors) | MEDIUM | 7/10 | `middleware.go:352` |

Each is detailed in an inline comment.

### Holistic assessment

The PR cleanly substitutes the `upstream-swap` middleware with a `strip-auth` middleware when `DisableUpstreamTokenInjection` is set, and gates `X-Forwarded` for remote upstreams. End-to-end wiring (CRD → `BuildAuthServerRunConfig` → `authserver.RunConfig` → `addUpstreamSwapMiddleware`) is complete and nil-safe. The operator/CRD surface is fully convention-compliant: plain `bool` + `+kubebuilder:default=false` is correct (not `*bool`), v1beta1-compatible, and exactly the two CRDs embedding `EmbeddedAuthServerConfig` were regenerated. Auth ordering is correct — the client JWT is fully validated before the header is stripped, and authz/audit read identity from context, so stripping doesn't break them.

The recurring theme across reviewers: the feature is framed as "no client credentials reach the public upstream," but the implementation only guarantees that for the *bearer token*. Closing that gap (or tightening the docs to match the narrower guarantee) is the most valuable change.

**Pros**
- Opt-in, secure-by-default; existing deployments unaffected.
- Strong test coverage: factory + handler + Close, branch selection, both wiring paths; negative assertion that unrelated headers pass through.
- Correct middleware ordering (validate-before-strip).

**Cons / risks to weigh**
- "strip-auth" implies broader credential isolation than it delivers (F1).
- Wholesale `X-Forwarded` skip is broader than the bug requires (F2).
- New middleware diverges from the package-per-middleware convention (F3).

### PR scope (F8, LOW, 7/10)

The `X-Forwarded` redirect-loop fix and the per-request debug logging in `transparent_proxy.go` are arguably a separate concern from the auth-injection feature. Per `.claude/rules/pr-creation.md` ("each PR = one logical change"), consider splitting the `X-Forwarded` fix into its own PR (it's independently consequential and revertable), or state in the description why they're coupled.

### Pre-existing, out of scope (F11, track separately)

The OAuth reviewer noted that the *enabled* `upstreamswap` `custom` header strategy (`pkg/auth/upstreamswap/middleware.go`) forwards the client's `Authorization` JWT upstream alongside the injected upstream token — a pre-existing token-passthrough behavior, not introduced here. Worth a separate issue; the new strip-auth path is the model the custom path could adopt.

### Dropped findings (audit trail, no action)

- **nolint:gosec G706 on new debug logs** — flagged for consistency, but the Lint CI check passed, so the lint-break premise is disproven. *(Verify req.URL carries no query credentials on this path — it does not.)*
- **`createStripAuthMiddleware` registers with the const instead of `config.Type`** — cosmetic symmetry only.
- **`slog.Debug` vs `slog.DebugContext`** — loses trace correlation; INFO.
- **`bool` vs `*bool` / `default=false`+`omitempty` redundancy** — confirmed correct and conventional for this codebase.

---
*Recommendation: **COMMENT** — no blockers; address the MEDIUM items (esp. F4 docs + F1 credential scope) before merge.*

ChrisJBurns:
Thanks @aron-muon  I wanted to push up the fixes to the review I submitted above, but I don't think I have permissions on your fork (im assuming). If I do, I'm more than happy to push the fixes up myself to save you having to wait longer for the PR to go in.

pkg/runner/middleware.go
aron-muon: Done in 9071bc80c99ac6933a0da41be78c136970202e12 — applied the suggestion; the stripped set (Authorization, Cookie, Proxy-Authorization) now mirrors what net/http refuses to copy across cross-host redirects, and the test asserts all three are removed.

pkg/runner/middleware.go
aron-muon: Done in 9071bc80c99ac6933a0da41be78c136970202e12 — extracted to `pkg/transport/middleware/strip_auth.go` with exported `StripAuthMiddlewareName` + `CreateStripAuthMiddleware`. I followed the `header_forward.go` precedent literally (a sibling file in `pkg/transport/middleware`, which is a flat one-file-per-middleware package) rather than a new subpackage; only the `addAuthHeaderStripMiddleware` wiring helper remains in `pkg/runner`.

pkg/runner/middleware.go
aron-muon: Done in 9071bc80c99ac6933a0da41be78c136970202e12 — (1) `TestPopulateMiddlewareConfigs_StripAuthOrdering` pins auth-before-strip, with the invariant documented at the wiring site; (2) went with build-time rejection: `DisableUpstreamTokenInjection` + token exchange or AWS STS now fails `PopulateMiddlewareConfigs` with a descriptive error (covered by `TestPopulateMiddlewareConfigs_StripAuthConflicts`). header-forward remains allowed deliberately — injecting a static API key for the backend while stripping client credentials is a coherent combination.

pkg/transport/proxy/transparent/transparent_proxy.go
aron-muon: Done in 84464ec98517b2159509ea810d8138b358df3e36 — applied the suggestion: `SetXForwarded()` always runs and only `X-Forwarded-Host` is deleted for remote upstreams, via a `setXForwardedHeaders` helper. New `TestSetXForwardedHeaders` covers both paths (local keeps Host; remote omits it; For/Proto present in both).

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
aron-muon: Done in c641a503709e66298b4b3db57563a99a135e58d2 — both comments now state that the client credential headers (Authorization, Cookie, Proxy-Authorization) are removed after JWT validation and the backend receives an unauthenticated request, plus the headerForward escape hatch and the token-exchange/AWS-STS incompatibility. CRD schemas, crd-api.md, and swagger regenerated.

aron-muon:
> Thanks @aron-muon I wanted to push up the fixes to the review I submitted above, but I don't think I have permissions on your fork (im assuming). If I do, I'm more than happy to push the fixes up myself to save you having to wait longer for the PR to go in.

All set, implemented requested changes

ChrisJBurns (approved):

ChrisJBurns:
@aron-muon  Awesome, cheers bud!! 🚀  Approved and merged

PR #4194, opened 2026-03-17
feat: Add client secret expiry tracking and automatic renew for DCR

### LINKED ISSUE #3631: Implement Client Secret Expiry and Renewal for Dynamic Client Registration
# Implement Client Secret Expiry and Renewal for Dynamic Client Registration

## Summary

ToolHive currently implements token persistence and Dynamic Client Registration (DCR) persistence, but lacks support for tracking and renewing expired client secrets. When an OAuth provider issues a client secret with an expiration time (per RFC 7591), ToolHive should:

1. Store the expiration time from the DCR response
2. Monitor for secret expiration
3. Automatically renew the client secret before it expires using the registration access token

### ✅ Implemented

1. **Token Persistence**: Refresh tokens are persisted with expiry tracking (`CachedTokenExpiry`)
2. **DCR Client Credentials Persistence**: Client ID and client secret are stored
   - Client ID stored as plain text in `CachedClientID`
   - Client secret stored securely via secret manager reference in `CachedClientSecretRef`
3. **Config Structure**: The `CachedSecretExpiry` field exists in `pkg/auth/remote/config.go` to store expiry time

### ❌ Missing Implementation

1. **DCR Response Metadata Not Stored**:
   - `ClientSecretExpiresAt` (int64 Unix timestamp) from DCR response is not persisted
   - `RegistrationAccessToken` is not stored (needed for secret renewal)
   - `RegistrationClientURI` is not stored (needed for secret renewal)

2. **Data Flow Gaps**:
   - `OAuthFlowResult` struct (`pkg/auth/discovery/discovery.go`) doesn't include DCR metadata
   - `ClientCredentialsPersister` callback signature only accepts `(clientID, clientSecret string)`, missing expiry and registration metadata
   - DCR response metadata is lost after `handleDynamicRegistration()` completes

3. **Expiry Checking Logic**:
   - No logic to check if `CachedSecretExpiry` indicates an expired or soon-to-expire secret
   - No validation when restoring cached credentials

4. **Renewal Logic**:
   - No implementation to renew client secret using RFC 7591 Update Client Registration
   - No use of `RegistrationAccessToken` and `RegistrationClientURI` for renewal

### Relevant Code Locations

1. **DCR Response Structure**: `pkg/auth/oauth/dynamic_registration.go`
   - `DynamicClientRegistrationResponse` includes `ClientSecretExpiresAt`, `RegistrationAccessToken`, `RegistrationClientURI`
   
2. **OAuth Flow Result**: `pkg/auth/discovery/discovery.go`
   - `OAuthFlowResult` struct needs to include DCR metadata fields
   
3. **Persistence Callback**: `pkg/auth/remote/handler.go`
   - `ClientCredentialsPersister` type signature needs to accept additional parameters
   - `wrapWithPersistence()` calls the persister but doesn't pass expiry/metadata
   
4. **Runner Implementation**: `pkg/runner/runner.go`
   - `SetClientCredentialsPersister()` callback implementation needs to store expiry and registration metadata
   
5. **Config Storage**: `pkg/auth/remote/config.go`
   - `CachedSecretExpiry` field exists but is never populated
   - `CachedRegTokenRef` field exists but is never populated

### RFC 7591 Compliance

Per RFC 7591 (OAuth 2.0 Dynamic Client Registration Protocol):
- Section 3.2.1: Client secret expiration is optional but should be honored
- Section 3.2.2: Registration access token can be used to update client registration
- Section 2.1: Client registration can be updated via PUT to `registration_client_uri`

### Phase 1: Store DCR Metadata

1. **Extend `OAuthFlowResult`**:
   ```go
   type OAuthFlowResult struct {
       // ... existing fields ...
       
       // DCR metadata for secret renewal
       ClientSecretExpiresAt   time.Time // Zero means no expiry
       RegistrationAccessToken string    // For updating registration
       RegistrationClientURI   string    // Endpoint for registration updates
   }
   ```

2. **Update `handleDynamicRegistration()`**:
   - Store DCR response metadata in `OAuthFlowConfig` or pass it through to `OAuthFlowResult`
   - Convert `ClientSecretExpiresAt` (int64 Unix timestamp) to `time.Time`

3. **Extend `ClientCredentialsPersister` signature**:
   ```go
   type ClientCredentialsPersister func(
       clientID string,
       clientSecret string,
       secretExpiry time.Time,
       registrationAccessToken string,
       registrationClientURI string,
   ) error
   ```

4. **Update persistence in `runner.go`**:
   - Store `secretExpiry` in `CachedSecretExpiry`
   - Store `registrationAccessToken` securely via secret manager in `CachedRegTokenRef`
   - Store `registrationClientURI` (can be plain text, or derive from issuer)

### Phase 2: Expiry Checking

1. **Add expiry check in `resolveClientCredentials()`** (`pkg/auth/remote/handler.go`):
   - Check if `CachedSecretExpiry` indicates secret is expired or expiring soon (e.g., within 24 hours)
   - If expired/expiring, trigger renewal before using credentials

2. **Add validation in `tryRestoreFromCachedTokens()`**:
   - Verify secret hasn't expired before attempting token refresh
   - Return appropriate error if secret is expired

### Phase 3: Secret Renewal

1. **Implement RFC 7591 Update Client Registration**:
   - Create function to update client registration using `RegistrationAccessToken`
   - Use PUT request to `RegistrationClientURI` per RFC 7591 Section 2.1
   - Handle response with new `client_secret` and updated `client_secret_expires_at`

2. **Automatic Renewal Logic**:
   - When secret is expiring soon, automatically renew it
   - Update stored secret and expiry in config
   - Handle renewal failures gracefully (log warning, allow manual intervention)

3. **Error Handling**:
   - If renewal fails, log warning and continue with existing secret (may still work)
   - If secret is already expired and renewal fails, return error requiring re-authentication

## Testing Considerations

1. **Unit Tests**:
   - Test expiry checking logic with various expiry times
   - Test renewal flow with mock DCR server
   - Test error handling for expired secrets

2. **Integration Tests**:
   - Test with OAuth provider that issues expiring secrets
   - Test automatic renewal before expiry
   - Test behavior when renewal fails

3. **Edge Cases**:
   - Zero expiry (secret never expires)
   - Missing registration access token (some providers don't provide it)
   - Clock skew handling
   - Concurrent renewal attempts

## Related Code References

- DCR Response: `pkg/auth/oauth/dynamic_registration.go:135-153`
- OAuth Flow Result: `pkg/auth/discovery/discovery.go:505-517`
- Persistence Callback: `pkg/auth/remote/handler.go:22-47`
- Runner Persistence: `pkg/runner/runner.go:666-694`
- Config Structure: `pkg/auth/remote/config.go:58-68`

## Acceptance Criteria

- [ ] DCR response metadata (`ClientSecretExpiresAt`, `RegistrationAccessToken`, `RegistrationClientURI`) is stored
- [ ] Expiry checking logic validates secret expiry before use
- [ ] Automatic renewal of expiring secrets using RFC 7591 Update Client Registration
- [ ] Graceful handling of renewal failures
- [ ] Comprehensive test coverage for expiry and renewal scenarios
- [ ] Documentation updated to explain secret expiry behavior

## Additional Notes

- This feature is important for long-running workloads that rely on DCR
- Some OAuth providers (e.g., Keycloak) issue client secrets with expiration times
- The implementation should be backward compatible with providers that don't expire secrets (zero expiry value)

Sanskarzz:
## Summary

Implements client secret expiry tracking and automatic renewal for Dynamic Client Registration (DCR)

Fixes: #3631

ToolHive already stored the `client_id` and `client_secret` from DCR responses, but discarded the `registration_access_token`, `registration_client_uri`, and `client_secret_expires_at` fields. Without these, there was no way to detect or renew an expired client secret, causing silent authentication failures for long-running workloads on providers like Keycloak that issue expiring secrets.

This PR implements the full RFC 7591 / RFC 7592 secret lifecycle in three phases:

1. **Store** all DCR response metadata through the persistence pipeline
2. **Detect** expiry proactively (24 h buffer) and on session restore
3. **Renew** the secret via RFC 7592 §2.2 before it becomes a problem

Note
All the phases are added in the same PR.

---

### `pkg/auth/remote/persisting_token_source.go`

Extended `ClientCredentialsPersister` function type signature:

```go
// Before
type ClientCredentialsPersister func(clientID, clientSecret string) error

// After
type ClientCredentialsPersister func(
    clientID string,
    clientSecret string,
    secretExpiry time.Time,           // zero = never expires (RFC 7591 §3.2.1)
    registrationAccessToken string,   // for RFC 7592 management operations
    registrationClientURI string,     // endpoint for RFC 7592 updates
    tokenEndpointAuthMethod string,   // "client_secret_post" or "client_secret_basic" (RFC 7591 §2)
) error
```

### `pkg/auth/remote/config.go`

- Added `CachedRegClientURI` field to store `registration_client_uri` as plain text
- Added `CachedTokenEndpointAuthMethod` field to store the authentication method used during DCR
- Updated `ClearCachedClientCredentials()` to also clear the new fields

### `pkg/auth/discovery/discovery.go`

- Added DCR renewal metadata fields to `OAuthFlowConfig` (used as the threading vehicle from `handleDynamicRegistration` through to the result)
- Extended `OAuthFlowResult` with `SecretExpiry`, `RegistrationAccessToken`, `RegistrationClientURI`, and `TokenEndpointAuthMethod`
- Updated `handleDynamicRegistration()` to capture all four from `DynamicClientRegistrationResponse`:
  - `ClientSecretExpiresAt > 0` → `time.Unix(...)` (zero if the field is `0`, meaning never expires)
  - `RegistrationAccessToken` and `RegistrationClientURI` copied as-is
  - `TokenEndpointAuthMethod` captured from response (defaults to `client_secret_post` in request)
- Updated `newOAuthFlow()` to populate the new fields in `OAuthFlowResult`

### `pkg/auth/remote/handler.go`

- Updated `wrapWithPersistence()` to pass all 6 arguments to `clientCredentialsPersister`
- Added `"time"` import
- `resolveClientCredentials()` now proactively calls `renewClientSecret()` when the secret is expiring within 24 h; renewal failures are soft-logged and execution continues
- `tryRestoreFromCachedTokens()` now checks expiry before attempting token refresh:
  - Within 24 h buffer + renewal fails → warning, continue with existing secret
  - Past expiry + renewal fails → hard error, forces a fresh OAuth flow

### `pkg/runner/runner.go`

Updated `SetClientCredentialsPersister` callback to match the new 6-argument signature and persist:
- `CachedSecretExpiry` — stored directly in config
- `registrationAccessToken` — stored securely in the secret manager, reference saved to `CachedRegTokenRef`
- `registrationClientURI` — stored as plain text in `CachedRegClientURI`
- `tokenEndpointAuthMethod` — stored directly in config as `CachedTokenEndpointAuthMethod`

### `pkg/auth/remote/secret_renewal.go` *(new file)*

Implements RFC 7592 §2.2 client secret renewal:

- `isSecretExpiredOrExpiringSoon()` — returns true when the secret is within `secretExpiryBuffer` (24 h) of expiry or already past it; false for zero expiry (never expires)
- `renewClientSecret(ctx)` — sends an authenticated HTTP `PUT` to `registration_client_uri` per RFC 7592 §2.2:
  - Retrieves `registrationAccessToken` from the secret manager
  - Validates `registration_client_uri` (must be HTTPS or localhost)
  - Sends current client metadata as the request body (required by RFC 7592)
  - Parses the response (RFC 7592 §2.1), extracts new `client_secret`, `client_secret_expires_at`, and optionally a rotated `registration_access_token`
  - Persists everything via `clientCredentialsPersister`
- `validateRegistrationClientURI(uri)` — validates the URI is HTTPS (or localhost for development)

### `pkg/auth/remote/secret_renewal_test.go` *(new file)*

16 new tests covering:

| Test | Coverage |
|---|---|
| `TestIsSecretExpiredOrExpiringSoon` (5 cases) | Zero, far-future, within-buffer, past-expiry, at-boundary |
| `TestValidateRegistrationClientURI` (6 cases) | Empty, HTTPS, HTTP non-localhost rejected, localhost/127.0.0.1 allowed, invalid URL |
| `TestRenewClientSecret_MissingConfig` (3 cases) | No URI, no token ref, no secret provider |
| `TestRenewClientSecret_Success` | Happy path with mock RFC 7592 server; token rotation |
| `TestRenewClientSecret_ServerError` | HTTP 401 from server |
| `TestRenewClientSecret_NoPersister` | Server OK but no persister configured |
| `TestRenewClientSecret_ZeroExpiryInResponse` | `client_secret_expires_at: 0` → `time.Time{}` (never expires) |

---

## Breaking Changes

> [!WARNING]
> **`ClientCredentialsPersister` function type signature changed.**
>
> Any code outside this repository that implements or calls `ClientCredentialsPersister` must be updated to use the new 6-parameter signature. Within this repository, the only call site is `pkg/runner/runner.go`, which is updated in this PR.

---

## Backward Compatibility

| Scenario | Behaviour |
|---|---|
| Provider does **not** issue expiring secrets (`client_secret_expires_at = 0`) | `CachedSecretExpiry` is `time.Time{}`. `isSecretExpiredOrExpiringSoon()` returns `false`. No renewal is attempted. |
| Provider does **not** support RFC 7592 (no `registration_access_token` / URI) | `renewClientSecret()` returns an immediate error. Caller logs a warning and continues with the existing secret.  |
| Provider **does** issue expiring secrets and supports RFC 7592 | Secret renewed automatically within 24 h of expiry.  |
| Renewal **fails** but secret is still within its validity window | Soft warning logged, existing secret used.  |
| Renewal **fails** and secret is past its expiry | Hard error returned; caller performs a fresh OAuth flow. |

---

## RFC References

- [RFC 7591 3.2.1](https://www.rfc-editor.org/rfc/rfc7591#section-3.2.1) — `client_secret_expires_at`, `registration_access_token`, `registration_client_uri` in registration response
- [RFC 7592 2.2](https://www.rfc-editor.org/rfc/rfc7592#section-2.2) — Client Update Request (PUT with Bearer token)
- [RFC 7592 2.1](https://www.rfc-editor.org/rfc/rfc7592#section-2.1) — Client Information Response (response format for read/update)

---

## Test Results

```
ok  github.com/stacklok/toolhive/pkg/auth               PASS
ok  github.com/stacklok/toolhive/pkg/auth/awssts        PASS
ok  github.com/stacklok/toolhive/pkg/auth/discovery     PASS
ok  github.com/stacklok/toolhive/pkg/auth/oauth         PASS
ok  github.com/stacklok/toolhive/pkg/auth/remote        PASS  (16 new tests)
ok  github.com/stacklok/toolhive/pkg/auth/secrets       PASS
ok  github.com/stacklok/toolhive/pkg/auth/tokenexchange PASS
ok  github.com/stacklok/toolhive/pkg/auth/upstreamswap  PASS
ok  github.com/stacklok/toolhive/pkg/auth/upstreamtoken PASS
```

## Type of change

- [ ] Bug fix
- [*] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

I have done E2E testing with Keycloak.
```
sanskar@sanskar-HP-Laptop-15s-du1xxx:~/opensource/stacklok/toolhive$ ./bin/thv logs keycloak-e2e-detached --proxy | grep -E "renew|RFC 7592|expired"
{"time":"2026-03-22T03:19:13+05:30","level":"DEBUG","msg":"DCR response includes registration access token for RFC 7592 operations"}
{"time":"2026-03-22T03:20:32+05:30","level":"DEBUG","msg":"Stored DCR registration access token for RFC 7592 operations"}
{"time":"2026-03-22T03:24:40+05:30","level":"DEBUG","msg":"DCR response includes registration access token for RFC 7592 operations"}
sanskar@sanskar-HP-Laptop-15s-du1xxx:~/opensource/stacklok/toolhive$ 
```
```
{"time":"2026-03-22T03:46:24+05:30","level":"INFO","msg":"Cached client secret is expiring or expired; attempting renewal before token restore","expiry":"2024-01-01T00:00:00Z"}
{"time":"2026-03-22T03:46:24+05:30","level":"DEBUG","msg":"Attempting RFC 7592 client secret renewal","registration_client_uri":"http://localhost:8080/realms/mcp-test/clients-registrations/openid-connect/8327e637-eef9-4f2e-93fb-768e5ed6b4a1"}
{"time":"2026-03-22T03:46:24+05:30","level":"INFO","msg":"Successfully renewed client secret via RFC 7592","client_id":"8327e637-eef9-4f2e-93fb-768e5ed6b4a1","new_expiry_zero":true,"reg_token_rotated":true}
```
```

- [*] Unit tests (`task test`)
- [*] E2E tests (`task test-e2e`)
- [*] Linting (`task lint-fix`)
- [*] Manual testing (describe below) Pending

### CODE REVIEW

File: pkg/auth/remote/secret_renewal.go [hunk 1]
```diff
@@ -0,0 +1,231 @@
...
+		return fmt.Errorf("client credentials persister not configured; cannot save renewed secret")
+	}
+
+	var newExpiry time.Time
+	if updateResp.ClientSecretExpiresAt > 0 {
+		newExpiry = time.Unix(updateResp.ClientSecretExpiresAt, 0)
+	}
+
+	// Use the rotated registration_access_token if provided; fall back to existing.
+	newRegToken := updateResp.RegistrationAccessToken
```
amirejaz, line 189: **[Spec / Secret leak]** When the server rotates the `registration_access_token` (`newRegToken != ""`), `persistClientCredentials` calls `GenerateUniqueSecretNameWithPrefix` which mints a **new** `OAUTH_REG_TOKEN_*` key. The old key is never deleted, so each renewal accumulates a stale secret in the store.

Same issue applies to `OAUTH_CLIENT_SECRET_*` keys — every call to `persistClientCredentials` with a non-empty `clientSecret` generates a new key.

Fix: before writing the new secret, delete the old key if `CachedRegTokenRef` / `CachedClientSecretRef` is already set (or reuse the existing key name rather than generating a fresh one).

pkg/auth/remote/handler.go
```diff
@@ -217,6 +227,18 @@ func (h *Handler) resolveClientCredentials(ctx context.Context) (clientID, clien
 		clientID = h.config.CachedClientID
 		slog.Debug("Using cached DCR client credentials", "client_id", clientID)
 
+		// Proactively renew the client secret if it is expiring soon (RFC 7592)
+		if h.isSecretExpiredOrExpiringSoon() {
```
amirejaz, line 231: **[Spec / Double renewal]** `tryRestoreFromCachedTokens` (below, ~line 270) and `resolveClientCredentials` (here) both call `renewClientSecret` independently. When `tryRestoreFromCachedTokens` is on the call path it calls `renewClientSecret` first, then calls `resolveClientCredentials`, which evaluates `isSecretExpiredOrExpiringSoon()` again — but `h.config.CachedSecretExpiry` is **never updated in memory** by `persistClientCredentials` (only the persisted config is updated). So the second check still fires and a second PUT is sent to the RFC 7592 endpoint.

Fix: after a successful renewal, update `h.config.CachedSecretExpiry` in memory so the second `isSecretExpiredOrExpiringSoon()` call returns `false`.

pkg/auth/remote/handler.go
```diff
@@ -217,6 +227,18 @@ func (h *Handler) resolveClientCredentials(ctx context.Context) (clientID, clien
...
+			slog.Info("Cached client secret is expiring soon, attempting renewal",
```
amirejaz, line 232: **[Convention]** `slog.Info` fires on every token resolution when the secret is expiring. Per `.claude/rules/go-style.md` "silent success — no output at INFO or above for successful operations". This should be `slog.Debug` (or `slog.Warn` only when the secret is **already past** expiry, not just approaching it).

```suggestion
			slog.Debug("Cached client secret is expiring soon, attempting renewal",
```

pkg/auth/remote/handler.go
```diff
@@ -243,6 +265,27 @@ func (h *Handler) tryRestoreFromCachedTokens(
 		return nil, fmt.Errorf("secret provider not configured, cannot restore cached tokens")
 	}
 
+	// Check if the cached client secret is expired before attempting token refresh.
+	// If it has fully expired and renewal also fails we must force a fresh OAuth flow.
+	if h.isSecretExpiredOrExpiringSoon() {
+		slog.Info("Cached client secret is expiring or expired; attempting renewal before token restore",
```
amirejaz, line 271: **[Convention]** Same `slog.Info` → `slog.Debug`/`slog.Warn` issue as above. Diagnostic messages that fire on every `tryRestoreFromCachedTokens` call should not be at INFO level.

```suggestion
		slog.Debug("Cached client secret is expiring or expired; attempting renewal before token restore",
```

pkg/auth/remote/secret_renewal.go
```diff
@@ -0,0 +1,231 @@
...
+	if err := validateRegistrationClientURI(h.config.CachedRegClientURI); err != nil {
+		return fmt.Errorf("invalid registration_client_uri: %w", err)
+	}
+
+	// Build the update request body with the current client metadata.
+	// Per RFC 7592 §2.2, the request MUST include all client metadata fields
+	// that were provided during the initial registration.
+	updateReq := clientUpdateRequest{
+		ClientID:                h.config.CachedClientID,
+		ClientName:              "ToolHive MCP Client",
```
amirejaz, line 97: **[Spec compliance]** Two problems with this request body:

1. `ClientName` is a hardcoded string literal instead of `oauth.ToolHiveMCPClientName`. If the constant changes, the value sent during registration and the value sent during renewal will diverge — violating RFC 7592's requirement for consistent metadata.

2. `h.config.CallbackPort` is the **configured default** port, not the port that was actually registered during DCR (which may have been reassigned by `networking.FindOrUsePort`). If the port differs, strict RFC 7592 servers will reject the request because the `redirect_uri` won't match.

Fix: use `oauth.ToolHiveMCPClientName` for the name, and persist the registered callback port (similar to how `CachedRegClientURI` is now persisted) so it can be echoed back accurately.

pkg/auth/remote/secret_renewal.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal client update request: %w", err)
+	}
+
+	// Create PUT request per RFC 7592 §2.2
+	req, err := http.NewRequestWithContext(
+		ctx,
+		http.MethodPut,
+		h.config.CachedRegClientURI,
+		strings.NewReader(string(reqBody)),
```
amirejaz, line 114: **[Convention]** `reqBody` is already `[]byte` from `json.Marshal`. Converting it to `string` then wrapping in `strings.NewReader` creates an unnecessary allocation and copy.

```suggestion
		bytes.NewReader(reqBody),
```

Add `"bytes"` to the import block and remove `"strings"` if it is no longer used elsewhere.

pkg/auth/remote/secret_renewal.go
```diff
@@ -0,0 +1,231 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create client update request: %w", err)
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Authorization", "Bearer "+regAccessToken) //nolint:gosec // G117
```
amirejaz, line 122: **[Convention]** The `//nolint:gosec // G117` comment is wrong. `G117` is the gosec rule for `unsafe.Pointer`. The suppression here has no valid gosec rule to suppress — building an `Authorization` header from a retrieved secret value is not a hardcoded credential (which would be `G101`). Either remove the directive entirely or replace with the correct rule number and justification.

pkg/auth/remote/secret_renewal.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Execute the request
+	httpClient := &http.Client{
```
amirejaz, line 125: **[Suggestion]** A new `http.Client` and `http.Transport` are created on every renewal call, which prevents TCP connection reuse. It also makes this function impossible to test with an injected transport.

Consider accepting an injectable `networking.HTTPClient` (the same interface used by `registerClientDynamicallyWithClient`) so tests can provide a custom transport and production code can reuse a shared client.

pkg/auth/remote/secret_renewal.go
```diff
@@ -0,0 +1,231 @@
...
+		return fmt.Errorf("client update request failed: %w", err)
+	}
+	defer func() {
+		if closeErr := resp.Body.Close(); closeErr != nil {
+			slog.Debug("Failed to close renewal response body", "error", closeErr)
+		}
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		errorBody, _ := io.ReadAll(io.LimitReader(resp.Body, 4096))
```
amirejaz, line 144: **[Suggestion]** `io.LimitReader(resp.Body, 4096)` reads at most 4096 bytes. When the error response body is larger, the remaining bytes are not drained before `resp.Body.Close()` fires (the deferred close runs next), which prevents the underlying TCP connection from being reused.

Add a full drain after reading the error body:
```go
errorBody, _ := io.ReadAll(io.LimitReader(resp.Body, 4096))
_, _ = io.Copy(io.Discard, resp.Body)
```

pkg/auth/remote/secret_renewal.go
```diff
@@ -0,0 +1,231 @@
...
+		updateResp.ClientSecret,
+		newExpiry,
+		newRegToken,
+		newRegURI,
+		h.config.CachedTokenEndpointAuthMethod,
+	); err != nil {
+		return fmt.Errorf("failed to persist renewed client secret: %w", err)
+	}
+
+	slog.Info("Successfully renewed client secret via RFC 7592",
```
amirejaz, line 206: **[Convention]** `slog.Info` for a successful operation. Per `.claude/rules/go-style.md` "silent success" rule, this should be `slog.Debug`.

```suggestion
	slog.Debug("Successfully renewed client secret via RFC 7592",
```

pkg/auth/remote/secret_renewal.go
```diff
@@ -0,0 +1,231 @@
...
+		"client_id", updateResp.ClientID,
+		"new_expiry_zero", newExpiry.IsZero(),
+		"reg_token_rotated", newRegToken != "")
+
+	return nil
+}
+
+// validateRegistrationClientURI validates that the registration_client_uri is
+// a valid HTTPS URL (or localhost for development).
+func validateRegistrationClientURI(registrationClientURI string) error {
```
amirejaz, line 216: **[Edge case]** `validateRegistrationClientURI` validates scheme and host but permits any path, including `/`. A `registration_client_uri` of `https://example.com/` passes validation but would send a PUT to the server root — almost certainly a misconfiguration.

RFC 7592 §2.2 implies the URI must be the unique management endpoint for a specific client registration. Add a check:
```go
if parsedURL.Path == "" || parsedURL.Path == "/" {
    return fmt.Errorf("registration_client_uri must include a non-root path: %s", registrationClientURI)
}
```

pkg/auth/remote/secret_renewal_test.go
```diff
@@ -0,0 +1,604 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/stacklok/toolhive/pkg/secrets"
+)
+
+// mockSecretProvider is a simple in-memory secret store for tests.
+// It implements the full secrets.Provider interface.
+type mockSecretProvider struct {
```
amirejaz, line 24: **[Convention]** Per `.claude/rules/testing.md`: "Use `go.uber.org/mock` (gomock) framework — never hand-write mocks."

A generated mock for `secrets.Provider` already exists at `pkg/secrets/mocks/mock_provider.go` (run `task gen` to regenerate). Please replace this hand-written `mockSecretProvider` with the generated mock to stay consistent with project conventions and avoid silent interface drift if `secrets.Provider` gains new methods.

pkg/auth/remote/secret_renewal_test.go
```diff
@@ -0,0 +1,604 @@
...
+			CachedRegTokenRef:  "rat-ref",
+		},
+		secretProvider: &mockSecretProvider{
+			secrets: map[string]string{"rat-ref": "rat-token"},
+		},
+	}
+
+	err := h.renewClientSecret(context.Background())
+	if err == nil {
+		t.Fatal("expected error, got nil")
```
amirejaz, line 408: **[Convention]** This test uses `t.Fatal` / `t.Errorf` while the rest of the test file uses testify `require` / `assert`. Per `.claude/rules/testing.md`, prefer `require.NoError` / `assert.Contains` throughout for consistency.

```suggestion
		require.Error(t, err)
```

pkg/runner/runner.go
```diff
@@ -782,6 +738,91 @@ func (r *Runner) handleRemoteAuthentication(ctx context.Context) (oauth2.TokenSo
...
+
+func (r *Runner) persistClientCredentials(
+	ctx context.Context,
+	secretManager secrets.Provider,
+	clientID, clientSecret string,
+	secretExpiry time.Time,
+	regAccessToken, regClientURI string,
+	tokenEndpointAuthMethod string,
+) error {
+	r.Config.RemoteAuthConfig.CachedClientID = clientID
```
amirejaz, line 779: **[Convention / Correctness]** `r.Config.RemoteAuthConfig.CachedClientID` is updated in memory **before** the subsequent `StoreSecretInManagerWithProvider` and `SaveState` calls. If either of those fails, the in-memory config reflects the new `clientID` but the durable state is inconsistent with it.

Per `.claude/rules/go-style.md`: write to durable storage before updating in-memory state. Move all `r.Config.RemoteAuthConfig.*` assignments to after `SaveState` returns successfully.

amirejaz (changes requested):
## Review: PR #4194 — Client Secret Expiry and Renewal (RFC 7592)

The overall design is sound and addresses a real DCR lifecycle gap. Found a few spec compliance issues, a double-renewal race, and some convention violations that should be addressed before merge.

### Summary of findings
- **Must fix**: secret accumulation in the secrets store on renewal, double PUT to RFC 7592 endpoint, hardcoded `ClientName` literal, wrong `CallbackPort` in renewal request body
- **Should fix**: hand-written mock (use generated one), `slog.Info` for success/diagnostic paths, `strings.NewReader(string([]byte))` allocation, wrong `//nolint:gosec` rule number, in-memory state updated before `SaveState`
- **Nice to have**: injectable HTTP client for testability, full response body drain, testify consistency

amirejaz:
Hi @Sanskarzz — thank you for this contribution! Client secret expiry tracking is a real gap in the DCR lifecycle and this PR addresses it thoughtfully.

It has been about 3.5 weeks since review feedback was left with no follow-up, so we're closing for now to keep the queue tidy. The feedback is all addressable — please feel free to open a new PR when you have time to work through it, and don't hesitate to reach out on [Discord](https://discord.gg/stacklok) if anything in the review needs clarification.

Thanks for your effort on this!

PR #4198, opened 2026-03-18
Key authserver's upstream token storage on (sessionID, providerName)

### LINKED ISSUE #4136: Implement multi-provider upstream token storage layer (RFC-0052 Phase 1)
## Description

Add a `providerName` dimension to the `UpstreamTokenStorage` interface so that multiple upstream IDPs can have their tokens stored per-session without overwriting each other. This is the foundational storage layer change for RFC-0052 — all subsequent phases (handler chain, config/operator, identity enrichment) break at compile time without these new interface signatures landing first.

## Context

RFC-0052 extends the embedded OAuth authorization server to support multiple upstream IDPs in a single instance. Today, `UpstreamTokenStorage` stores exactly one set of tokens per session (`map[sessionID]*timedEntry`), which makes it impossible to accumulate tokens from a second provider during a sequential authorization chain. This task restructures the storage layer to accommodate `N` providers per session by keying on `(sessionID, providerName)`.

This is the root task of the RFC-0052 implementation plan. TASK-002 (Handler and Server Layer), TASK-003 (Config and Operator), and TASK-004 (Identity Enrichment) all depend on the new interface signatures produced here.

**Dependencies**: None (root task)
**Blocks**: TASK-002, TASK-003, TASK-004

> **Breaking Change — Redis key pattern**: Upstream tokens stored under the old Redis key pattern `{prefix}upstream:{sessionID}` (a single key per session) will be unreachable after upgrade. There is no automatic migration. Existing sessions will require re-authentication after deploying this change. This must be documented in the PR description.

## Acceptance Criteria

- [ ] `UpstreamTokenStorage` interface in `pkg/authserver/storage/types.go` has updated signatures: `StoreUpstreamTokens(ctx, sessionID, providerName string, tokens *UpstreamTokens)`, `GetUpstreamTokens(ctx, sessionID, providerName string)`, `GetAllUpstreamTokens(ctx, sessionID string) (map[string]*UpstreamTokens, error)`, and `DeleteUpstreamTokens(ctx, sessionID string)` (session-scoped delete, signature unchanged)
- [ ] `PendingAuthorization` struct in `pkg/authserver/storage/types.go` has two new fields: `UpstreamProviderName string` and `SessionID string` with doc comments matching the architecture spec
- [ ] `MemoryStorage.upstreamTokens` field is restructured from `map[string]*timedEntry[*UpstreamTokens]` to `map[string]map[string]*timedEntry[*UpstreamTokens]` (outer key = sessionID, inner key = providerName)
- [ ] `MemoryStorage.StoreUpstreamTokens` and `GetUpstreamTokens` accept and use `providerName`; both validate that `providerName` is non-empty and return a clear error if it is
- [ ] `MemoryStorage.GetAllUpstreamTokens` is implemented: returns a `map[string]*UpstreamTokens` (defensive copies) for all providers under the given sessionID; returns an empty map (not an error) for an unknown sessionID
- [ ] `MemoryStorage` eviction TTL for upstream tokens uses `max(accessToken.ExpiresAt, refreshToken.ExpiresAt)` instead of `tokens.ExpiresAt.Add(DefaultRefreshTokenTTL)`
- [ ] Redis storage uses new key pattern `{prefix}upstream:{sessionID}:{providerName}` for each provider key (via new `redisUpstreamKey` helper in `redis_keys.go`)
- [ ] Redis storage maintains a session-scoped index SET at `{prefix}upstream:idx:{sessionID}` (using SADD) for efficient `GetAllUpstreamTokens` and session-level delete
- [ ] `RedisStorage.StoreUpstreamTokens` uses a Lua script for atomic write of the provider key + SADD to the session index SET
- [ ] `RedisStorage.DeleteUpstreamTokens` uses a Lua script for atomic SMEMBERS + DEL of all provider keys + DEL of the index SET
- [ ] `RedisStorage.GetAllUpstreamTokens` is implemented using SMEMBERS + MGET (two round-trips); returns empty map for unknown session
- [ ] `RedisStorage.StoreUpstreamTokens` and `GetUpstreamTokens` validate that `providerName` is non-empty and return a clear error if it is
- [ ] All call sites of `StoreUpstreamTokens` and `GetUpstreamTokens` within `pkg/authserver/` are updated to pass a `providerName` argument: `pkg/authserver/server/handlers/callback.go` and `pkg/authserver/refresher.go`
- [ ] Storage mocks in `pkg/authserver/storage/mocks/mock_storage.go` are regenerated via `mockgen` to reflect the new interface signatures
- [ ] `memory_test.go` is updated: existing upstream token tests pass `providerName`; new test cases cover multi-provider store/get, session delete wipes all providers, `GetAllUpstreamTokens` returns all providers (including those with expired tokens), `GetAllUpstreamTokens` returns empty map for unknown session, and eviction TTL uses `max(access.ExpiresAt, refresh.ExpiresAt)`
- [ ] `redis_test.go` and `redis_integration_test.go` are updated with equivalent coverage for the Redis backend (miniredis and real Redis respectively)
- [ ] `task license-check` passes (all new/modified `.go` files have SPDX headers)
- [ ] `task lint-fix` passes with no remaining lint errors
- [ ] `task test` passes (unit tests)
- [ ] PR description documents the Redis breaking change (old key pattern `{prefix}upstream:{sessionID}` becomes unreachable; sessions require re-authentication)
- [ ] `upstreamtoken.Service` interface in `pkg/auth/upstreamtoken/types.go` has updated `GetValidTokens` signature: `GetValidTokens(ctx context.Context, sessionID, providerName string) (*UpstreamCredential, error)`
- [ ] `InProcessService.GetValidTokens` in `pkg/auth/upstreamtoken/service.go` accepts and uses `providerName` when calling `storage.GetUpstreamTokens`; `refreshOrFail` is updated to thread `providerName` through its call chain
- [ ] `upstreamswap` middleware `Config` struct in `pkg/auth/upstreamswap/middleware.go` gains a `ProviderName string` field; `validateConfig` returns an error if `ProviderName` is empty
- [ ] `upstreamswap` middleware passes `cfg.ProviderName` as the `providerName` argument to `svc.GetValidTokens` in `createMiddlewareFunc`
- [ ] The proxy runner derives `ProviderName` from the single upstream provider name in the embedded auth server config and stores it in the `MiddlewareConfig` parameters for the `upstreamswap` middleware during wiring (the proxy runner only ever has one upstream provider, so a single `ProviderName` in `MiddlewareConfig` is sufficient for Phase 1)
- [ ] Mocks for `upstreamtoken.Service` (if generated) are regenerated to reflect the new `GetValidTokens` signature
- [ ] `pkg/auth/upstreamswap/middleware_test.go` is updated: existing tests pass `providerName` to `GetValidTokens`; new test cases cover empty `ProviderName` config validation and correct forwarding of `ProviderName` to the service

Start with `pkg/authserver/storage/types.go` — update the `UpstreamTokenStorage` interface and `PendingAuthorization` struct first so the compiler immediately surfaces every call site that needs updating. Then update `memory.go` (restructure the nested map and TTL logic), then `redis_keys.go` (add `redisUpstreamKey`), then `redis.go` (new key pattern, session index, Lua scripts, `GetAllUpstreamTokens`). Update the two call sites in `callback.go` and `refresher.go` last (they will need a `providerName` argument — derive this from `pending.UpstreamProviderName` in the callback handler, and from `expired.ProviderID` in the refresher). Regenerate mocks after all interface changes are complete.

Next, propagate `providerName` up through the token service layer: update `upstreamtoken.Service.GetValidTokens` in `pkg/auth/upstreamtoken/types.go` and `InProcessService.GetValidTokens` in `pkg/auth/upstreamtoken/service.go`. Finally, wire the provider name into the `upstreamswap` middleware: add `ProviderName` to `Config`, enforce it in `validateConfig`, pass it to `GetValidTokens` in `createMiddlewareFunc`, and update the proxy runner to derive the provider name from the embedded auth server config when constructing the middleware config.

### Patterns and Frameworks

- **Defensive copies on read and write**: Follow the existing pattern in `MemoryStorage.StoreUpstreamTokens` and `GetUpstreamTokens` — make a full struct copy before storing and before returning, so external mutations do not affect stored state. Preserve this when restructuring to the nested map.
- **Lua scripts for atomicity in Redis**: Follow the existing `storeUpstreamTokensScript` pattern in `redis.go`. The new `StoreUpstreamTokens` Lua script must atomically write the provider key and SADD to the session index SET in one round-trip. The new `DeleteUpstreamTokens` Lua script must atomically SMEMBERS, DEL all member keys, and DEL the index SET.
- **`warnOnCleanupErr` for best-effort secondary index cleanup**: Use this existing helper when non-critical index cleanup operations fail (see `redis.go`).
- **`redisKey` / `redisProviderKey` / `redisSetKey` key generation conventions**: Follow the same `fmt.Sprintf` format in `redis_keys.go` for the new `redisUpstreamKey` helper.
- **`errors.Is()` / `fmt.Errorf` with `%w`**: Use for error wrapping and inspection throughout; never swallow errors silently.
- **Table-driven tests with `require.NoError`**: Follow patterns in `memory_test.go` and `redis_test.go`; use `require.NoError(t, err)` rather than `t.Fatal`.
- **`go.uber.org/mock` (mockgen)**: Regenerate mocks with `mockgen -destination=mocks/mock_storage.go -package=mocks -source=types.go Storage,PendingAuthorizationStorage,ClientRegistry,UpstreamTokenStorage,UpstreamTokenRefresher,UserStorage`.
- **SPDX license headers**: All new or modified `.go` files must have SPDX headers; run `task license-fix` to add them automatically.

### Code Pointers

- `pkg/authserver/storage/types.go` — Primary file to modify: `UpstreamTokenStorage` interface signatures (add `providerName` param to `StoreUpstreamTokens`/`GetUpstreamTokens`, add `GetAllUpstreamTokens`), and `PendingAuthorization` struct (add `UpstreamProviderName` and `SessionID` fields). The `//go:generate mockgen` directive at the top of this file controls mock regeneration.
- `pkg/authserver/storage/memory.go` lines 48–80 — `MemoryStorage` struct definition: change `upstreamTokens map[string]*timedEntry[*UpstreamTokens]` to `map[string]map[string]*timedEntry[*UpstreamTokens]`.
- `pkg/authserver/storage/memory.go` lines 685–777 — Current `StoreUpstreamTokens`, `GetUpstreamTokens`, `DeleteUpstreamTokens`: shows the defensive copy pattern, `timedEntry` wrapper, eviction TTL calculation, and error wrapping to replicate when restructuring. TTL calculation on line 699 uses `tokens.ExpiresAt.Add(DefaultRefreshTokenTTL)` — this must change to `max(accessToken.ExpiresAt, refreshToken.ExpiresAt)` (note: `UpstreamTokens` currently has only one `ExpiresAt` field; confirm whether the RFC intent is to track `RefreshToken` expiry separately or to use existing `ExpiresAt` as the access token expiry).
- `pkg/authserver/storage/redis_keys.go` — Add `redisUpstreamKey(prefix, sessionID, providerName string) string` returning `"{prefix}upstream:{sessionID}:{providerName}"`. The session index SET key should follow the `redisSetKey` pattern: `redisSetKey(prefix, "upstream:idx", sessionID)` → `"{prefix}upstream:idx:{sessionID}"`.
- `pkg/authserver/storage/redis.go` lines 889–1004 — Current Redis `StoreUpstreamTokens`, `GetUpstreamTokens`, `DeleteUpstreamTokens` and `storeUpstreamTokensScript` Lua script. The new implementation must replace the single-key pattern with the two-key pattern (provider key + session index SET). The `marshalUpstreamTokensWithTTL` helper (line ~885) is reusable.
- `pkg/authserver/storage/mocks/mock_storage.go` — Generated file; do not edit by hand. Regenerate after all interface changes.
- `pkg/authserver/server/handlers/callback.go` line 127 — Call site for `StoreUpstreamTokens`; must add `providerName` argument. At this call site, `providerName` should come from `pending.UpstreamProviderName` (set by the authorize handler when it creates the `PendingAuthorization`). Note: in the current single-upstream handler, `pending.UpstreamProviderName` does not exist yet — adding it to `PendingAuthorization` here is sufficient for Phase 1 correctness; Phase 2 will drive multi-upstream chain logic that populates it per-leg.
- `pkg/authserver/server/handlers/callback.go` line 141 — `DeleteUpstreamTokens` call: signature is unchanged (session-scoped).
- `pkg/authserver/refresher.go` line 70 — Call site for `StoreUpstreamTokens`; must add `providerName` argument. Use `expired.ProviderID` as the `providerName` (the refresher already has access to the expired token struct which carries the provider ID).
- `pkg/authserver/storage/memory_test.go` line 429 — `TestMemoryStorage_UpstreamTokens` test suite: update all test cases with the new `providerName` parameter and add new multi-provider and `GetAllUpstreamTokens` cases.
- `pkg/authserver/storage/redis_test.go` and `pkg/authserver/storage/redis_integration_test.go` — Redis equivalents; follow the same test pattern updates.
- `pkg/auth/upstreamtoken/types.go` — `Service` interface: update `GetValidTokens` signature to add `providerName string` as the second parameter (after `sessionID`).
- `pkg/auth/upstreamtoken/service.go` line 50 — `InProcessService.GetValidTokens`: add `providerName string` parameter; pass it to `storage.GetUpstreamTokens` and thread it through `refreshOrFail` (which calls `RefreshAndStore` — the refresher signature itself is not changing in Phase 1, but `providerName` must be available so `StoreUpstreamTokens` inside the refresher can use it).
- `pkg/auth/upstreamswap/middleware.go` line 32 — `Config` struct: add `ProviderName string \`json:"provider_name" yaml:"provider_name"\`` field representing the single upstream provider this proxy runner instance serves.
- `pkg/auth/upstreamswap/middleware.go` line 96 — `validateConfig`: add a check that `cfg.ProviderName != ""`; return a descriptive error if empty.
- `pkg/auth/upstreamswap/middleware.go` line 190 — `svc.GetValidTokens(r.Context(), tsid)` call site: update to `svc.GetValidTokens(r.Context(), tsid, cfg.ProviderName)`.
- Proxy runner wiring code (search for `upstreamswap` or `MiddlewareConfig` construction in `pkg/runner/` or `cmd/thv/`) — locate where the `upstreamswap` middleware config is built and add `ProviderName` derived from the single upstream IDP provider name in the embedded auth server config.

### Component Interfaces

```go
// pkg/authserver/storage/types.go

// UpstreamTokenStorage — updated interface
type UpstreamTokenStorage interface {
    // StoreUpstreamTokens stores upstream IDP tokens for (sessionID, providerName).
    // Returns an error if sessionID or providerName is empty.
    StoreUpstreamTokens(ctx context.Context, sessionID, providerName string, tokens *UpstreamTokens) error

    // GetUpstreamTokens retrieves upstream IDP tokens for (sessionID, providerName).
    // Returns ErrNotFound if the entry does not exist.
    // Returns ErrExpired (with token data) if the access token has expired — caller
    // may use the refresh token for transparent renewal.
    // Returns ErrInvalidBinding if binding validation fails.
    GetUpstreamTokens(ctx context.Context, sessionID, providerName string) (*UpstreamTokens, error)

    // GetAllUpstreamTokens returns all upstream tokens for the session, keyed by
    // provider name. Returns an empty map (not an error) for an unknown sessionID.
    // Expired tokens are included; callers that need freshness checks should use
    // GetUpstreamTokens per-provider.
    GetAllUpstreamTokens(ctx context.Context, sessionID string) (map[string]*UpstreamTokens, error)

    // DeleteUpstreamTokens removes all upstream tokens for the session (all providers).
    // Returns ErrNotFound if the session does not exist.
    DeleteUpstreamTokens(ctx context.Context, sessionID string) error
}

// PendingAuthorization — new fields added to existing struct
type PendingAuthorization struct {
    // ... existing fields unchanged ...

    // UpstreamProviderName is the provider being authenticated in this chain leg.
    // Set by the authorize handler when creating the PendingAuthorization;
    // read by the callback handler to key token storage.
    UpstreamProviderName string

    // SessionID is the TSID for the session being accumulated across all chain legs.
    // Generated server-side only using rand.Text(); never accepted from client input.
    SessionID string
}
```

```go
// pkg/authserver/storage/redis_keys.go — new helper

// redisUpstreamKey generates a Redis key for a per-provider upstream token entry.
// Format: "{prefix}upstream:{sessionID}:{providerName}"
// Note: sessionID and providerName must not contain colons in practice
// (sessionID is rand.Text() output; providerName is a config-supplied name).
func redisUpstreamKey(prefix, sessionID, providerName string) string {
    return fmt.Sprintf("%s%s:%s:%s", prefix, KeyTypeUpstream, sessionID, providerName)
}

// Session index SET key (use existing redisSetKey):
// redisSetKey(prefix, "upstream:idx", sessionID) → "{prefix}upstream:idx:{sessionID}"
```

```go
// pkg/authserver/storage/memory.go — MemoryStorage struct field change

// upstreamTokens maps sessionID -> providerName -> timedEntry.
// Outer key is the session TSID; inner key is the provider name.
upstreamTokens map[string]map[string]*timedEntry[*UpstreamTokens]
```

```go
// pkg/auth/upstreamtoken/types.go — updated Service interface

// Service owns the upstream token lifecycle: read, refresh, error handling.
type Service interface {
    // GetValidTokens returns a valid upstream credential for (sessionID, providerName).
    // It transparently refreshes expired access tokens using the refresh token.
    //
    // Returns:
    //   - *UpstreamCredential on success
    //   - ErrSessionNotFound if no upstream tokens exist for the session
    //   - ErrNoRefreshToken if the access token is expired and no refresh token is available
    //   - ErrRefreshFailed if the refresh attempt fails (e.g., revoked refresh token)
    GetValidTokens(ctx context.Context, sessionID, providerName string) (*UpstreamCredential, error)
}
```

```go
// pkg/auth/upstreamswap/middleware.go — Config struct update

// Config holds configuration for upstream swap middleware.
type Config struct {
    // ProviderName is the upstream IDP provider name used to key token storage.
    // For the proxy runner this is derived from the single upstream provider in
    // the embedded auth server config. Required — validation returns an error if empty.
    ProviderName string `json:"provider_name" yaml:"provider_name"`

    // HeaderStrategy determines how to inject the token: "replace" (default) or "custom".
    HeaderStrategy string `json:"header_strategy,omitempty" yaml:"header_strategy,omitempty"`

    // CustomHeaderName is the header name when HeaderStrategy is "custom".
    CustomHeaderName string `json:"custom_header_name,omitempty" yaml:"custom_header_name,omitempty"`
}
```

### Testing Strategy

**Unit Tests — `pkg/authserver/storage/memory_test.go`**
- [ ] `TestMemoryStorage_UpstreamTokens`: update all existing subtests to pass `providerName` (e.g., `"provider-a"`)
- [ ] Multi-provider store/get: store tokens for two providers under the same sessionID; verify each `GetUpstreamTokens` call returns the correct tokens for each provider independently
- [ ] Cross-provider isolation: storing tokens for provider A does not overwrite or affect tokens for provider B
- [ ] `GetAllUpstreamTokens` — two providers: store two providers, call `GetAllUpstreamTokens`, verify map has exactly two entries with correct defensive copies
- [ ] `GetAllUpstreamTokens` — unknown session: call `GetAllUpstreamTokens` for a sessionID that has never been stored; verify returns empty map and no error
- [ ] `GetAllUpstreamTokens` — includes expired tokens: store a token with an expired `ExpiresAt`; verify `GetAllUpstreamTokens` includes it (no expiry filtering at bulk-read level)
- [ ] Session delete wipes all providers: store two providers, call `DeleteUpstreamTokens`, verify both providers are gone (subsequent `GetUpstreamTokens` for each returns `ErrNotFound`)
- [ ] Eviction TTL uses `max(access.ExpiresAt, refresh.ExpiresAt)`: confirm the cleanup loop evicts entries at the correct deadline (test using manipulated `time.Now` or very short TTLs)
- [ ] `StoreUpstreamTokens` with empty `providerName` returns a non-nil error
- [ ] `GetUpstreamTokens` with empty `providerName` returns a non-nil error

**Unit Tests — `pkg/authserver/storage/redis_test.go` (miniredis)**
- [ ] Same multi-provider store/get coverage as memory tests above
- [ ] `GetAllUpstreamTokens` returns empty map for unknown session (SMEMBERS on non-existent key)
- [ ] Session index SET (`{prefix}upstream:idx:{sessionID}`) is populated on store and cleaned up atomically on delete
- [ ] `StoreUpstreamTokens` with empty `providerName` returns a non-nil error
- [ ] `GetUpstreamTokens` with empty `providerName` returns a non-nil error

**Integration Tests — `pkg/authserver/storage/redis_integration_test.go`**
- [ ] Multi-provider round-trip: store two providers, retrieve individually and via `GetAllUpstreamTokens`, verify correct values (requires `//go:build integration` tag and real Redis via testcontainers)
- [ ] Session delete wipes all providers atomically in Redis
- [ ] TTL expiry: keys expire from Redis after the configured TTL (verify via `PTTL` or waiting)

**Unit Tests — `pkg/auth/upstreamswap/middleware_test.go`**
- [ ] Update existing tests: mock `svc.GetValidTokens` now expects `(ctx, sessionID, providerName)` — update expectations to pass the configured `ProviderName`
- [ ] Empty `ProviderName` in config: `CreateMiddleware` (or `validateConfig`) returns a non-nil error
- [ ] `ProviderName` forwarding: verify the middleware passes `cfg.ProviderName` to `GetValidTokens` (not a hardcoded value)

**Edge Cases**
- [ ] `GetAllUpstreamTokens` with a sessionID that has one expired and one valid provider: verify both are returned (expiry is per-provider at the `GetUpstreamTokens` level, not filtered at bulk-read)
- [ ] Concurrent `StoreUpstreamTokens` for different providers under the same sessionID (memory backend): verify no data corruption under the `sync.RWMutex`
- [ ] Redis Lua script atomicity: verify that a `StoreUpstreamTokens` failure partway through does not leave the session index SET in an inconsistent state (e.g., SADD succeeds but key write fails — Lua rollback)

## Out of Scope

- Handler layer changes (sequential chain logic, `nextMissingUpstream`, multi-upstream authorize/callback): covered in TASK-002
- Config and operator validation changes (removing/moving `len > 1` guard): covered in TASK-003
- `Identity.UpstreamTokens` enrichment: covered in TASK-004
- Full Redis migration tooling for tokens stored under the old `{prefix}upstream:{sessionID}` key pattern
- `UpstreamTokenRefresher.RefreshAndStore` multi-provider support (the refresher in `pkg/authserver/refresher.go` currently takes a single `sessionID`; Phase 1 only updates the `StoreUpstreamTokens` call within it to pass `providerName` — no signature change to `RefreshAndStore` itself)
- Transparent token refresh or distributed locking for concurrent refresh races
- Dynamic upstream IDP registration at runtime
- Multi-provider `upstreamswap` middleware support (TASK-004 will handle the case where the proxy runner needs to serve multiple upstream providers; Phase 1 configures a single `ProviderName` per middleware instance)

## References

- RFC: `docs/proposals/THV-0052-multi-upstream-idp-authserver.md`
- Parent epic: #3924
- Architecture doc for auth server storage: `docs/arch/11-auth-server-storage.md`
- Related Stacklok epic: https://github.com/stacklok/stacklok-epics/issues/251

jhrozek:
## Summary

Multi-upstream IDP support requires each provider's tokens to be stored and retrieved independently per session. The existing storage keyed upstream tokens on sessionID alone, making it impossible to hold tokens from more than one provider concurrently. This PR re-keys the storage to (sessionID, providerName) and adds backwards-compatible migration for existing deployments.

- Restructure `UpstreamTokenStorage` interface so tokens are keyed by (sessionID, providerName), with a new `GetAllUpstreamTokens` bulk-read method for session enumeration
- Memory backend uses nested map; Redis backend uses per-provider keys with a session index SET for atomic enumeration and deletion
- Add `PendingAuthorization.UpstreamProviderName` and `SessionID` fields for future multi-leg authorization chains
- Add backwards-compatible migration paths for both Redis token keys and user provider identities, so upgrading deployments preserve sessions and internal user IDs
- Harden migration against cross-provider identity merge: scope legacy identity lookup to the single protocol-type ID matching the upstream's type, not a global scan
- Propagate transient storage errors during legacy lookup (fail closed) instead of silently creating duplicate users
- Add `isLegacyUpstreamProviderID` guard in Redis migration to prevent logical-name tokens from being re-claimed under a different provider namespace
- Enforce `pending.UpstreamProviderName` in `CallbackHandler` to reject misrouted callbacks that would associate a session with the wrong provider
- Inject `UserResolver` into `NewHandler` as a dependency so the handler carries no migration-specific knowledge
- Derive `ProviderName` for the upstream swap middleware from the first upstream config entry

Fixes #4136

## Type of change

- [x] New feature

## Test plan

- [x] Extended the integration tests
- [x] Manual testing: end-to-end verified in a Kubernetes cluster. This branch is cherry-picked from a complete branch implementing the full multi-upstream feature.

No. This is an internal storage restructuring with backwards-compatible migration. Existing single-upstream deployments continue to work without configuration changes.

## Large PR Justification

Mostly the data migration makes things difficult to split. I was unsure if the migration was even needed given that all existing deployments are running with 1 upstream only, but I thought that the case where all users must re-log after an upgrade would be irritating enough.

### DIFF
--- a/pkg/authserver/storage/redis_integration_test.go
+++ b/pkg/authserver/storage/redis_integration_test.go
@@ -1220,3 +1220,196 @@ func TestIntegration_UnicodeInIdentifiers(t *testing.T) {
 		assert.Equal(t, "sub-données-中文", pi.ProviderSubject)
 	})
 }
+
+// --- Legacy Data Migration (Real Redis) ---
+//
+// These tests verify the one-shot bulk migration against real Redis,
+// catching SCAN behavior, pipeline atomicity, and TTL handling that
+// miniredis may not reproduce faithfully.
+
+func TestIntegration_MigrateLegacyUpstreamData(t *testing.T) {
+	t.Parallel()
+
+	t.Run("full migration lifecycle", func(t *testing.T) {
+		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
+			// Seed a user (needed for DeleteUser cascade test later).
+			userID := "migrate-user"
+			now := time.Now()
+			require.NoError(t, s.CreateUser(ctx, &User{ID: userID, CreatedAt: now, UpdatedAt: now}))
+
+			// Seed a legacy upstream token key: upstream:{sessionID} (no provider suffix).
+			sessionID := "legacy-sess-1"
+			legacyTokenKey := redisKey(s.keyPrefix, KeyTypeUpstream, sessionID)
+			legacyTokenJSON := fmt.Sprintf(
+				`{"provider_id":"oidc","access_token":"legacy-at","refresh_token":"legacy-rt","id_token":"legacy-idt","expires_at":0,"user_id":"%s","upstream_subject":"upstream-sub","client_id":"test-client"}`,
+				userID,
+			)
+			require.NoError(t, s.client.Set(ctx, legacyTokenKey, legacyTokenJSON, time.Hour).Err())
+
+			// Seed a legacy provider identity: provider:4:oidc:{subject}
+			legacyIdentityKey := redisProviderKey(s.keyPrefix, "oidc", "upstream-sub")
+			legacyIdentityJSON := fmt.Sprintf(
+				`{"user_id":"%s","provider_id":"oidc","provider_subject":"upstream-sub","linked_at":%d,"last_used_at":%d}`,
+				userID, now.Unix(), now.Unix(),
+			)
+			require.NoError(t, s.client.Set(ctx, legacyIdentityKey, legacyIdentityJSON, 0).Err())
+
+			// Also add the legacy identity to the user's provider set (as origin/main would).
+			userProviderSetKey := redisSetKey(s.keyPrefix, KeyTypeUserProviders, userID)
+			require.NoError(t, s.client.SAdd(ctx, userProviderSetKey, legacyIdentityKey).Err())
+
+			// --- Run migration ---
+			require.NoError(t, s.MigrateLegacyUpstreamData(ctx, "default", "oidc"))
+
+			// --- Verify token migration ---
+
+			// Legacy key should be gone.
+			exists, err := s.client.Exists(ctx, legacyTokenKey).Result()
+			require.NoError(t, err)
+			assert.Equal(t, int64(0), exists, "legacy token key should be deleted after migration")
+
+			// Token should be readable under the new key format.
+			tokens, err := s.GetUpstreamTokens(ctx, sessionID, "default")
+			require.NoError(t, err)
+			require.NotNil(t, tokens)
+			assert.Equal(t, "legacy-at", tokens.AccessToken)
+			assert.Equal(t, "legacy-rt", tokens.RefreshToken)
+			assert.Equal(t, "default", tokens.ProviderID, "ProviderID should be patched to logical name")
+			assert.Equal(t, userID, tokens.UserID)
+			assert.Equal(t, "upstream-sub", tokens.UpstreamSubject)
+			assert.Equal(t, "test-client", tokens.ClientID)
+
+			// Session index set should contain the new key.
+			idxKey := redisSetKey(s.keyPrefix, KeyTypeUpstreamIdx, sessionID)
+			newTokenKey := redisUpstreamKey(s.keyPrefix, sessionID, "default")
+			isMember, err := s.client.SIsMember(ctx, idxKey, newTokenKey).Result()
+			require.NoError(t, err)
+			assert.True(t, isMember, "session index should contain the migrated token key")
+
+			// User:upstream reverse index should contain the new key (for DeleteUser cascade).
+			userUpstreamKey := redisSetKey(s.keyPrefix, KeyTypeUserUpstream, userID)
+			isMember, err = s.client.SIsMember(ctx, userUpstreamKey, newTokenKey).Result()
+			require.NoError(t, err)
+			assert.True(t, isMember, "user:upstream set should contain the migrated token key")
+
+			// --- Verify identity migration ---
+
+			// New identity should be readable under the logical provider name.
+			identity, err := s.GetProviderIdentity(ctx, "default", "upstream-sub")
+			require.NoError(t, err)
+			assert.Equal(t, userID, identity.UserID)
+			assert.Equal(t, "default", identity.ProviderID)
+
+			// Legacy identity should still exist (not deleted for safe rollback).
+			legacyIdentity, err := s.GetProviderIdentity(ctx, "oidc", "upstream-sub")
+			require.NoError(t, err)
+			assert.Equal(t, userID, legacyIdentity.UserID, "legacy identity should be preserved")
+
+			// --- Verify DeleteUser cascade includes migrated token ---
+			require.NoError(t, s.DeleteUser(ctx, userID))
+
+			_, err = s.GetUpstreamTokens(ctx, sessionID, "default")
+			assert.ErrorIs(t, err, ErrNotFound, "migrated token should be removed by DeleteUser cascade")
+
+			_, err = s.GetUser(ctx, userID)
+			assert.ErrorIs(t, err, ErrNotFound)
+		})
+	})
+
+	t.Run("idempotent: second run is a no-op", func(t *testing.T) {
+		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
+			// Seed and migrate.
+			legacyKey := redisKey(s.keyPrefix, KeyTypeUpstream, "idem-sess")
+			legacyJSON := `{"provider_id":"oidc","access_token":"idem-at","expires_at":0,"user_id":"u1","upstream_subject":"s1","client_id":"c1"}`
+			require.NoError(t, s.client.Set(ctx, legacyKey, legacyJSON, time.Hour).Err())
+
+			require.NoError(t, s.MigrateLegacyUpstreamData(ctx, "default", "oidc"))
+
+			// Verify migrated.
+			tokens, err := s.GetUpstreamTokens(ctx, "idem-sess", "default")
+			require.NoError(t, err)
+			assert.Equal(t, "idem-at", tokens.AccessToken)
+
+			// Run migration again — should be a no-op.
+			require.NoError(t, s.MigrateLegacyUpstreamData(ctx, "default", "oidc"))
+
+			// Token should still be there, unchanged.
+			tokens, err = s.GetUpstreamTokens(ctx, "idem-sess", "default")
+			require.NoError(t, err)
+			assert.Equal(t, "idem-at", tokens.AccessToken)
+		})
+	})
+
+	t.Run("no legacy data is a clean no-op", func(t *testing.T) {
+		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
+			// Run migration on an empty store — should succeed silently.
+			require.NoError(t, s.MigrateLegacyUpstreamData(ctx, "default", "oidc"))
+		})
+	})
+
+	t.Run("TTL preserved during migration", func(t *testing.T) {
+		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
+			legacyKey := redisKey(s.keyPrefix, KeyTypeUpstream, "ttl-sess")
+			legacyJSON := `{"provider_id":"oidc","access_token":"ttl-at","expires_at":0,"user_id":"u1","upstream_subject":"s1","client_id":"c1"}`
+			require.NoError(t, s.client.Set(ctx, legacyKey, legacyJSON, 30*time.Second).Err())
+
+			require.NoError(t, s.MigrateLegacyUpstreamData(ctx, "default", "oidc"))
+
+			// Verify the new key has a TTL close to the original.
+			newKey := redisUpstreamKey(s.keyPrefix, "ttl-sess", "default")
+			ttl := s.client.TTL(ctx, newKey).Val()
+			assert.InDelta(t, 30, ttl.Seconds(), 5, "migrated key TTL should be close to original")
+		})
+	})
+
+	t.Run("SCAN pagination with >100 legacy keys", func(t *testing.T) {
+		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
+			// Seed 150 legacy keys to force at least 2 SCAN iterations (batch size = 100).
+			const keyCount = 150
+			for i := 0; i < keyCount; i++ {
+				legacyKey := redisKey(s.keyPrefix, KeyTypeUpstream, fmt.Sprintf("page-sess-%d", i))
+				data := fmt.Sprintf(
+					`{"provider_id":"oidc","access_token":"at-%d","expires_at":0,"user_id":"u-%d","upstream_subject":"s-%d","client_id":"c-%d"}`,
+					i, i, i, i,
+				)
+				require.NoError(t, s.client.Set(ctx, legacyKey, data, time.Hour).Err())
+			}
+
+			require.NoError(t, s.MigrateLegacyUpstreamData(ctx, "default", "oidc"))
+
+			// Every legacy key should have been migrated.
+			for i := 0; i < keyCount; i++ {
+				tokens, err := s.GetUpstreamTokens(ctx, fmt.Sprintf("page-sess-%d", i), "default")
+				require.NoError(t, err, "key %d should be migrated", i)
+				assert.Equal(t, fmt.Sprintf("at-%d", i), tokens.AccessToken)
+				assert.Equal(t, "default", tokens.ProviderID)
+			}
+
+			// No legacy keys should remain.
+			for i := 0; i < keyCount; i++ {
+				legacyKey := redisKey(s.keyPrefix, KeyTypeUpstream, fmt.Sprintf("page-sess-%d", i))
+				exists, err := s.client.Exists(ctx, legacyKey).Result()
+				require.NoError(t, err)
+				assert.Equal(t, int64(0), exists, "legacy key %d should be deleted", i)
+			}
+		})
+	})
+
+	t.Run("new-format keys not touched by migration", func(t *testing.T) {
+		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
+			// Store a token via the normal write path (new format).
+			require.NoError(t, s.StoreUpstreamTokens(ctx, "new-sess", "github", &UpstreamTokens{
+				ProviderID: "github", AccessToken: "new-at",
+				UserID: "u1", ExpiresAt: time.Now().Add(time.Hour),
+			}))
+
+			require.NoError(t, s.MigrateLegacyUpstreamData(ctx, "default", "oidc"))
+
+			// The new-format token should be unchanged.
+			tokens, err := s.GetUpstreamTokens(ctx, "new-sess", "github")
+			require.NoError(t, err)
+			assert.Equal(t, "new-at", tokens.AccessToken)
+			assert.Equal(t, "github", tokens.ProviderID, "new-format token ProviderID should be untouched")
+		})
+	})
+}


pkg/authserver/server/handlers/handler.go
```diff
@@ -31,22 +31,36 @@ type Handler struct {
...
 func NewHandler(
 	provider fosite.OAuth2Provider,
 	config *server.AuthorizationServerConfig,
 	stor storage.Storage,
 	upstreamIDP upstream.OAuth2Provider,
+	upstreamName string,
+	userResolver *UserResolver,
 ) *Handler {
+	if userResolver == nil {
+		panic("handlers.NewHandler: userResolver must not be nil")
```
jerm-dro: suggestion, non-blocking: should this return an error instead?

pkg/authserver/storage/memory.go
```diff
@@ -214,10 +214,16 @@ func (s *MemoryStorage) cleanupExpired() {
...
 
-	var expiredUpstreamTokens []string
-	for k, v := range s.upstreamTokens {
-		if now.After(v.expiresAt) {
-			expiredUpstreamTokens = append(expiredUpstreamTokens, k)
+	type upstreamKey struct {
+		sessionID    string
+		providerName string
+	}
+	var expiredUpstreamTokens []upstreamKey
```
jerm-dro, line 223: suggestion: use this as the first key in `upstreamTokens`: `map[upstreamKey]*timedEntry[*UpstreamTokens`

That would allow you to avoid nesting maps.

pkg/authserver/storage/redis.go
```diff
@@ -886,15 +898,19 @@ func marshalUpstreamTokensWithTTL(tokens *UpstreamTokens) ([]byte, time.Duration
 	return data, ttl, nil
 }
 
-// StoreUpstreamTokens stores the upstream IDP tokens for a session.
-// Uses a Lua script to atomically read the old UserID, write new token data,
+// StoreUpstreamTokens stores the upstream IDP tokens for a session and provider.
+// Uses a Lua script to atomically write token data, update the session index set,
```
jerm-dro, line 902: Suggestion: define "session index set" and why it's needed. I've seen it used a few times but not described.

pkg/authserver/storage/redis.go
```diff
@@ -805,10 +805,12 @@ type storedUpstreamTokens struct {
...
 
 // storeUpstreamTokensScript atomically reads the existing UserID, writes new token
-// data, and updates user reverse-index sets. This prevents a race condition where
-// concurrent writes for the same session could leave orphaned entries in user sets.
+// data, updates the session index set, and updates user reverse-index sets.
+// This prevents a race condition where concurrent writes for the same session
+// could leave orphaned entries in user sets.
 //
-// KEYS[1] = token key
+// KEYS[1] = per-provider token key (e.g. "thv:auth:{ns:name}:upstream:{sessionID}:{providerName}")
```
jerm-dro, line 812: What is `ns:name`? Is it the provider?

pkg/authserver/storage/redis.go
```diff
@@ -805,10 +805,12 @@ type storedUpstreamTokens struct {
...
+// KEYS[2] = session index set key (e.g. "thv:auth:{ns:name}:upstream:idx:{sessionID}")
```
jerm-dro, line 813: Do I understand correctly?

The session index set is created so that we can use Redis' TTL mechanism to delete expired tokens.

pkg/authserver/storage/redis.go
```diff
@@ -922,12 +938,192 @@ func (s *RedisStorage) StoreUpstreamTokens(ctx context.Context, sessionID string
...
+		return s.migrateLegacyUpstreamTokens(ctx, sessionID, providerName, legacyKey, legacyTokens)
+	}
+	return tokens, err
+}
+
+// migrateLegacyUpstreamTokens moves a legacy upstream token entry from the old
+// key format (upstream:{sessionID}) to the new format (upstream:{sessionID}:{providerName}).
+// It patches ProviderID to the logical provider name so that subsequent refreshes
+// write to the correct key. The legacy key is deleted to prevent stale reads.
+// TODO: Remove once all deployments have migrated.
```
jerm-dro: How will you know that all deployments have been migrated?

pkg/authserver/storage/redis.go
```diff
@@ -922,12 +938,192 @@ func (s *RedisStorage) StoreUpstreamTokens(ctx context.Context, sessionID string
...
+		if (legacyErr != nil && !errors.Is(legacyErr, ErrExpired)) || legacyTokens == nil {
+			return tokens, err
+		}
+		if !isLegacyUpstreamProviderID(legacyTokens.ProviderID) {
+			// Legacy key exists but belongs to a logical provider name already;
+			// do not re-label it under a different namespace.
+			return nil, fmt.Errorf("%w: upstream tokens not found for provider %q", ErrNotFound, providerName)
+		}
+		return s.migrateLegacyUpstreamTokens(ctx, sessionID, providerName, legacyKey, legacyTokens)
+	}
```
jerm-dro: Question, non-blocking: We don't need to change this PR, but how could we perform this migration once in bulk (e.g. at startup time of the authserver)? is this something you considered but decided against?

pkg/authserver/server/handlers/handler.go
jhrozek: yeah, I kept finding these CC gems and removing them but I missed this one. Good find, should have been a nack.

pkg/authserver/storage/memory.go
jhrozek: done, good suggestion

pkg/authserver/storage/redis.go
jhrozek: done

pkg/authserver/storage/redis.go
jhrozek: I updated the comment, but basically it is the k8s namespace and the server name e.g. `{toolhive-system:github-mcp}`.

pkg/authserver/storage/redis.go
jhrozek: I don't think so, this index set exists for finding all session's keys without a full scan during GetAllUpstreamTokens and GetAllUpstreamTokens

pkg/authserver/storage/redis.go
jhrozek: yeah you wouldn't which is also why your next comment makes sense

pkg/authserver/storage/redis.go
jhrozek: I ended up prototyping it to weigh pros and cons and I actually like it. Initially I just thought it would be more work and complexity but it turns out it's the other way around - we trade one-time testable complexity for runtime complexity during the one-by-one lookups which seems like the right call.

jhrozek:
Thanks for the review @jerm-dro ! I took all your suggestions, they all made sense. 

I haven't tested the bulk migration yet outside the tests in the PR. I won't merge the PR until I've done so even if you ack in the meantime.

jhrozek:
OK, I managed to test the migration end-to-end in a kind cluster and it seemed to work.

I deployed thv from main, set up a github MCP server and authenticated to seed the legacy data. Upgraded to this branch, checked the migration log:
```
  migrated legacy upstream token    session_id=LNV44... provider_name=github
  migrated legacy provider identity legacy_provider_id=oauth2 provider_name=github provider_subject=715522
  legacy data migration complete    tokens_migrated=1 tokens_skipped=0 identities_migrated=1 identities_skipped=0
```
then checked the data in redis directly, they looked correct:
    - Legacy token key deleted, new key upstream:{sessionID}:github created
    - Session index set (upstream:idx:{sessionID}) populated
    - User reverse index (user:upstream:{userID}) updated (for DeleteUser cascade)

finally I called an MCP tool from the same VSCode session to see if I'm going to be reprompted for auth, I was not.

pkg/authserver/storage/redis.go
jerm-dro: Suggestion: Have you explored treating the key as just the session ID and the values as a `map[providerName]token`? I believe that could be done with a redis hash: https://redis.io/docs/latest/develop/data-types/hashes/

Arguably, that schema will be easier to reason about:
- You'd be able to delete the session index set key.
- `GetAllUpstreamTokens` becomes trivial.
- Expiration is implemented on a single key and it works for all tokens.

pkg/authserver/storage/redis_test.go
```diff
@@ -663,12 +663,309 @@ func TestRedisStorage_UpstreamTokens(t *testing.T) {
...
+	})
+
+}
+
+// --- Bulk Migration Tests ---
+
+func TestRedisStorage_MigrateLegacyUpstreamData(t *testing.T) {
+	t.Parallel()
+
+	t.Run("migrates legacy token key to new format", func(t *testing.T) {
```
jerm-dro, line 776: nice 😄

pkg/authserver/storage/redis.go
jhrozek: I think this is a very nice idea and I was prototyping the implementation, but during it I found out that the AWS managed valkey service doesn't support HEXPIRE which we would need to enable per-key TTL expiration (e.g. github token expires in 8 hours but atlassian token in 6 hours). And we already have users in pre-production running authserver using the AWS valkey service, so we'd effectively break them.

Too bad, the code would really be nicer and now would be a good time to migrate.

Since I think we don't want to completely forgot the idea I filed: https://github.com/stacklok/toolhive/issues/4279

jerm-dro: Thanks for investigating further. I agree with the decision to not do this now.

PR #4200, opened 2026-03-18
Resolve plain skill names from registry during install

JAORMX:
## Summary

- `thv skill install <name>` previously created a dead-end "pending" DB record when a plain name wasn't found in the local OCI store. This made the command silently fail for users who expected registry-based resolution.
- Now the install command queries the configured skill registry/index for an exact name match and, if found, extracts the OCI reference and installs from it. If nothing resolves, a 404 with an actionable error message is returned instead of a silent pending record.

This is Phase 2 of #4015. The resolution cascade is now:
1. Local OCI store (from `thv skill build`)
2. Registry/index lookup (new)
3. Actionable 404 error (replaces silent pending record)

Relates to #4015

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/skills/skillsvc/skillsvc.go` | Add `SkillLookup` interface and `WithSkillLookup` option; implement `resolveFromRegistry` method; replace `installPending` fallback with registry lookup cascade; fix lock management to prevent deadlock |
| `pkg/api/server.go` | Wire registry provider as `WithSkillLookup` via `buildSkillLookupOption` helper; rename local variable to avoid import shadow |
| `pkg/skills/skillsvc/skillsvc_test.go` | Update tests that expected pending fallback to expect 404; fix `TestConcurrentInstallAndUninstall` to exercise install critical section; add `TestInstallFromRegistry` with 9 subtests covering happy path, ambiguity, no OCI package, error degradation, case-insensitive matching, and supply-chain validation |

Yes. `thv skill install <name>` now resolves plain skill names from the configured registry instead of creating a non-functional pending record. If the skill is not found anywhere, users get an actionable error message suggesting the OCI reference syntax.

- **Lock management**: The per-skill lock is released before entering the registry lookup path because `installFromOCI` acquires its own lock on the artifact's skill name (which could be the same key). Go's `sync.Mutex` is not re-entrant, so holding both would deadlock. The TOCTOU window is safe because `installWithExtraction` performs idempotent checks (same digest = no-op, different digest = upgrade).
- **`installPending` removed**: The function and its callers are removed. The `InstallStatusPending` constant is retained for existing DB records.
- **Security hardening**: Registry-returned data in error messages is truncated (identifiers capped at 256 chars, ambiguity candidates capped at 5). An audit log at INFO level records registry-resolved OCI references.
- **Graceful degradation**: Registry lookup errors are logged at WARN and fall through to the not-found error, so a registry outage never blocks installation by explicit OCI reference.

### DIFF
--- a/test/e2e/api_skills_test.go
+++ b/test/e2e/api_skills_test.go
@@ -269,7 +269,7 @@ func createUpstreamRegistryWithSkill(skillName, ociReference string) string {
 
 // Test suite
 
-var _ = Describe("Skills API", Label("api", "api-clients", "skills", "e2e"), func() {
+var _ = Describe("Skills API", Label("api", "api-registry", "skills", "e2e"), func() {
 	var (
 		config    *e2e.ServerConfig
 		apiServer *e2e.Server
@@ -1021,6 +1021,13 @@ var _ = Describe("Skills API", Label("api", "api-clients", "skills", "e2e"), fun
 			})
 			defer updateResp.Body.Close()
 			Expect(updateResp.StatusCode).To(Equal(http.StatusOK))
+
+			// Reset registry to default after this test to avoid polluting
+			// the shared config directory used by other E2E tests.
+			DeferCleanup(func() {
+				resetResp := updateRegistry(apiServer, "default", map[string]interface{}{})
+				resetResp.Body.Close()
+			})
 
 			By("Installing by plain skill name — should resolve from registry")
 			installResp := installSkill(apiServer, installSkillRequest{Name: skillName})


pkg/api/server.go
```diff
@@ -595,6 +603,25 @@ func createListener(address string, isUnixSocket bool) (net.Listener, string, er
...
+func buildSkillLookupOption() skillsvc.Option {
+	cfg, err := config.NewDefaultProvider().LoadOrCreateConfig()
+	if err != nil {
+		slog.Debug("skill lookup not available: could not load config", "error", err)
+		return nil
+	}
+	if cfg == nil {
+		return nil
+	}
+	rp, err := registry.NewRegistryProvider(cfg, registry.WithInteractive(false))
```
amirejaz: `buildSkillLookupOption` takes no `context.Context`, but `NewRegistryProvider` can make a blocking network call here — specifically when `cfg.RegistryApiUrl` or `cfg.RegistryUrl` is set, the constructor probes the endpoint to verify reachability (`"custom registry at X is not reachable"`). That I/O runs uncancellable during server startup. The default path (embedded/local registry) is fine.

The parent `createDefaultManagers` has a `ctx` that could be threaded down if `NewRegistryProvider` grows context support. Not urgent for the default case, but worth tracking.

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -562,6 +601,76 @@ func (s *service) resolveFromLocalStore(ctx context.Context, opts *skills.Instal
...
+		}
+		return nil, httperr.WithCode(
+			fmt.Errorf("ambiguous skill name %q matches multiple registry entries: %s%s; install by full OCI reference instead",
+				name, strings.Join(candidates, ", "), suffix),
+			http.StatusConflict,
+		)
+	}
+
+	// Exactly one match -- find an OCI package.
+	for _, pkg := range matches[0].Packages {
```
amirejaz, line 652: nit: the first one is silently chosen. This is probably fine for now, but worth a comment explaining the selection policy.

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -568,6 +601,78 @@ func (s *service) resolveFromLocalStore(ctx context.Context, opts *skills.Instal
...
+// configured skill registry/index. Returns (ref, nil) on success, (nil, nil) when
+// no match is found or no lookup is configured, or (nil, err) on ambiguity.
+func (s *service) resolveFromRegistry(name string) (nameref.Reference, error) {
+	if s.skillLookup == nil {
+		return nil, nil
+	}
+
+	results, err := s.skillLookup.SearchSkills(name)
+	if err != nil {
+		slog.Warn("registry skill lookup failed, falling back to not-found", "name", name, "error", err)
```
jhrozek, line 614: Nit: when `SearchSkills` returns an error (network failure, auth error, server 500), the error is logged at WARN but converted to `(nil, nil)`, making it indistinguishable from "no match found." The user sees "skill not found in local store or registry" with no indication that the registry was actually unreachable.

The `s.skillLookup == nil` check at line 608 already handles the "no registry configured" case, so a non-nil lookup returning an error is a genuine problem worth surfacing. Consider either returning the error to the caller or including a hint in the not-found message that the registry lookup failed.

PR #4201, opened 2026-03-18
Default skills to the "default" group on install

JAORMX:
## Summary

Workloads already default to the "default" group when no group is specified
(`pkg/api/v1/workload_service.go`), but skills did not. When
`thv skill install <name>` was called without `--group`, the group field stayed
empty and `registerSkillInGroup` was a no-op, meaning skills were never added to
any group unless the user explicitly passed `--group`.

- Add the same defaulting logic in the skill service `Install` method so skills
  are automatically registered in the "default" group, matching workload behavior
- Update the existing test case to verify default group registration occurs

## Type of change

- [x] Bug fix

## Test plan

Skills installed without `--group` are now automatically added to the "default"
group, matching the existing behavior for workloads. Previously they were not
added to any group.

### DIFF
--- a/pkg/skills/skillsvc/skillsvc.go
+++ b/pkg/skills/skillsvc/skillsvc.go
@@ -188,6 +188,12 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
 	// the same lock key and DB record.
 	opts.ProjectRoot = projectRoot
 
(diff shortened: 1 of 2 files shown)


JAORMX:
@amirejaz skills are not yet exposed. All commands are hidden and we haven't been promoting them yet.

Issue #4228, opened 2026-03-19
Scoped secret store: admin escape hatch for system key inspection


amirejaz:
## Summary

Phase 5 of the scoped secret store implementation (#4192).

Provides a privileged escape hatch for operators who need to inspect or delete system-managed secrets directly, without going through the scoped wrappers.

## Work

- `--system` flag on `thv secret list` to include system-reserved keys in output
- `--system` flag on `thv secret delete` to allow deletion of a system-reserved key
- Both flags require explicit opt-in and display a warning
- E2E tests covering flag behaviour

## Dependencies

Depends on Phase 4. This is the last phase — the feature is complete once this ships.

Issue #4227, opened 2026-03-19
Scoped secret store: wire callers to use ScopedProvider and UserProvider


amirejaz:
## Summary

Phase 4 of the scoped secret store implementation (#4192).

Update all callers to use the correct provider wrapper so system secrets are isolated in practice, not just in theory.

## Work

- Registry auth (`pkg/registry/auth/`): use `CreateScopedSecretProvider(ScopeRegistry)`
- Workload auth (`pkg/auth/`): use `CreateScopedSecretProvider(ScopeWorkloads)`
- All user-facing entry points (CLI secret commands, API secrets routes, MCP tool server): use `CreateUserSecretProvider`
- Integration tests verifying isolation end-to-end

## Dependencies

Depends on Phase 3. Must ship in the same PR as Phase 3 (migration infrastructure) — callers must not be updated before migration runs, and migration is useless without updated callers.

Issue #4226, opened 2026-03-19
Scoped secret store: migration infrastructure for existing keys


amirejaz:
## Summary

Phase 3 of the scoped secret store implementation (#4192).

Existing users may have registry tokens and workload auth secrets stored under bare keys (no `__thv_` prefix). This phase adds crash-safe migration infrastructure to rename them into the correct scope on first run.

## Work

- `MigrateSystemKeys(ctx, provider, migrations []KeyMigration) error` in `migration.go`
- Uses `BulkDeleteSecrets` (store-before-delete ordering) so a crash mid-migration does not leave secrets unreachable
- Guarded by a config flag so migration only runs once
- Unit tests covering full migration, partial failure, and idempotency

## Dependencies

Depends on Phase 2. Must ship together with Phase 4 (wire callers) — shipping migration without updated callers, or updated callers without migration, would break existing users.

Issue #4225, opened 2026-03-19
Scoped secret store: factory helpers for ScopedProvider and UserProvider


amirejaz:
## Summary

Phase 2 of the scoped secret store implementation (#4192).

Add factory convenience helpers so callers construct wrappers consistently without needing to know the layering order.

## Work

- `CreateUserSecretProvider(providerType) (secrets.Provider, error)` in `factory.go` — wraps `CreateSecretProvider` result in `UserProvider`
- `CreateScopedSecretProvider(providerType, scope string) (secrets.Provider, error)` in `factory.go` — wraps `CreateSecretProvider` result in `ScopedProvider`

## Dependencies

Depends on Phase 1 (#4192).

PR #4229, opened 2026-03-19
Add scoped and user secret providers with system key isolation

### LINKED ISSUE #4224: Scoped secret store: core providers and system key isolation
## Summary

Phase 1 of the scoped secret store implementation (#4192).

Introduces the foundational types that isolate system-managed secrets from user-managed secrets using a reserved `__thv_<scope>_` key prefix.

## Work

- `ScopedProvider`: wraps any `Provider` and namespaces all operations under `__thv_<scope>_`. Used by internal callers (registry auth, workload auth, enterprise login).
- `UserProvider`: wraps any `Provider` and blocks access to system-reserved keys. Used at all user-facing boundaries (CLI, API, MCP tool server).
- `SystemKeyPrefix`, `ScopeRegistry`, `ScopeWorkloads`, `ScopeAuth` constants.
- `ErrReservedKeyName`: returned when a user command attempts to manage a system key.
- `BulkDeleteSecrets` added to the `Provider` interface so `Cleanup` on both wrappers is handled in a single write on `EncryptedManager` (no-op on read-only providers).

## Status

Covered by PR linked to this issue.

amirejaz:
## Summary

ToolHive's secret store currently uses a flat keyspace shared between system-managed secrets (registry OAuth tokens, workload auth credentials) and user-managed secrets. This means system tokens are visible and modifiable via `thv secret` commands, and a `Cleanup()` call from one context could inadvertently wipe secrets owned by another. This PR introduces the foundational isolation layer to fix that.

- Adds `ScopedProvider`: wraps any `Provider` and namespaces all operations under `__thv_<scope>_<name>`. Internal callers (registry auth, workload auth, enterprise login) will use this to keep their secrets isolated.
- Adds `UserProvider`: wraps any `Provider` and blocks access to system-reserved keys (`__thv_*`). All user-facing entry points (CLI, API, MCP tool server) will use this once callers are wired up (Phase 4, #4227).
- Adds `SystemKeyPrefix = "__thv_"`, named scopes (`ScopeRegistry`, `ScopeWorkloads`, `ScopeAuth`), and `ErrReservedKeyName` for consistent rejection of reserved key access.
- Adds `BulkDeleteSecrets` to the `Provider` interface so both wrappers' `Cleanup` methods operate atomically — no-op on read-only providers (1Password, env), single write on `EncryptedManager`.

Closes #4224

## Implementation plan

This is Phase 1 of a 5-phase rollout tracked in #4192:

| Phase | Issue | Status |
|-------|-------|--------|
| 1 — Core providers + `BulkDeleteSecrets` | #4224 | This PR |
| 2 — Factory helpers | #4225 | Up next |
| 3 — Migration infrastructure | #4226 | Depends on Phase 2 |
| 4 — Wire callers (ships with Phase 3) | #4227 | Depends on Phase 3 |
| 5 — Admin escape hatch (`--system` flag) | #4228 | Depends on Phase 4 |

Phases 3 and 4 must ship together — wiring callers before migration runs (or vice versa) would break existing users with secrets stored under bare keys.

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/secrets/scoped.go` | New — `ScopedProvider`, `UserProvider`, constants, `ErrReservedKeyName` |
| `pkg/secrets/scoped_test.go` | New — full table-driven tests for both providers |
| `pkg/secrets/types.go` | Add `BulkDeleteSecrets` to `Provider` interface |
| `pkg/secrets/encrypted.go` | Implement `BulkDeleteSecrets` (single lock, single write) |
| `pkg/secrets/environment.go` | `BulkDeleteSecrets` no-op (read-only) |
| `pkg/secrets/1password.go` | `BulkDeleteSecrets` no-op (read-only) |
| `pkg/secrets/fallback.go` | `BulkDeleteSecrets` delegates to primary |
| `pkg/secrets/mocks/mock_provider.go` | Regenerated |
| `pkg/environment/environment_test.go` | Updated hand-written mock to satisfy interface |

No. The wrappers exist but no callers use them yet — that happens in Phase 4 (#4227).

**Migration note (Phase 4):** When `UserProvider` is wired into user-facing code paths, any user whose secrets happen to be named with the `__thv_` prefix will receive `ErrReservedKeyName` instead of accessing them. This prefix was previously undocumented and unprotected, so real-world conflicts are unlikely, but Phase 4 release notes should explicitly call this out.

The key prefix `__thv_<scope>_<name>` (e.g. `__thv_registry_mytoken`) uses double-underscore and no slashes deliberately — 1Password treats `/` as a path separator which caused conflicts with an earlier `thv_sys/<scope>/` design. See the RFC (THV-0056) for the full discussion.

`BulkDeleteSecrets` is included here rather than Phase 2 because both `Cleanup()` implementations on `ScopedProvider` and `UserProvider` depend on it. Adding it to the interface now avoids a follow-up that would touch every provider file again.

## Large PR Justification

This PR is flagged as large due to tests, but the production code change is well within limits (~265 lines across 6 files excluding tests and generated code). The size cannot be reduced further because:
- Adding `BulkDeleteSecrets` to the `Provider` interface requires updating all implementations atomically — splitting across PRs would break the build mid-way.
- This is already Phase 1 of a 5-phase rollout. Phases 2–5 are in separate issues (#4225, #4226, #4227, #4228) and will be separate PRs.
- The bulk of the line count is comprehensive test coverage for all new methods and previously untested cases.

### DIFF
--- a/pkg/secrets/scoped.go
+++ b/pkg/secrets/scoped.go
@@ -11,6 +11,17 @@ import (
 )
 
 // SecretScope is the type for system-managed secret scope identifiers.
+//
+// Invariants that every SecretScope value MUST satisfy:
+//   - Non-empty: an empty scope would produce the prefix "__thv__", which is
+//     ambiguous and cannot be reliably stripped.
+//   - No underscores: the key format is "__thv_<scope>_<name>"; an underscore
+//     inside the scope would make it impossible to determine where the scope
+//     ends and the name begins.
+//
+// All constants declared in this package (ScopeRegistry, ScopeWorkloads,
+// ScopeAuth) satisfy these invariants. Custom scopes introduced in the future
+// must be validated against them.
 type SecretScope string
 
 const (
@@ -196,7 +207,10 @@ func (u *UserProvider) ListSecrets(ctx context.Context) ([]SecretDescription, er
 	return result, nil
 }
 
-// DeleteSecrets removes all named keys, rejecting any that are system-reserved.
+// DeleteSecrets removes all named keys with all-or-nothing semantics: it
+// validates every name in the list before issuing any delete to the underlying
+// store. If any name is system-reserved the entire operation is aborted and
+// ErrReservedKeyName is returned without deleting anything.
 func (u *UserProvider) DeleteSecrets(ctx context.Context, names []string) error {
 	for _, name := range names {
 		if isSystemKey(name) {
--- a/pkg/secrets/scoped_test.go
+++ b/pkg/secrets/scoped_test.go
@@ -6,6 +6,7 @@ package secrets_test
 import (
 	"context"
 	"errors"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -740,6 +741,13 @@ func TestUserProvider_DeleteSecrets(t *testing.T) {
 			wantReserve: true,
 		},
 		{
+			name:        "mixed input: aborts without deleting when any key is reserved",
+			inputNames:  []string{"valid-key", "__thv_registry_reserved"},
+			wantErr:     true,
+			wantReserve: true,
+			// expectCall is false: the inner provider must NOT be called at all
+		},
+		{
 			name:       "propagates error from inner",
 			inputNames: []string{"valid-key"},
 			wantErr:    true,
@@ -775,6 +783,30 @@ func TestUserProvider_DeleteSecrets(t *testing.T) {
 				require.NoError(t, err)
 			}
 		})
+	}
+}
+
+// ---------------------------------------------------------------------------
+// SecretScope invariant tests
+// ---------------------------------------------------------------------------
+
+// TestSecretScopeInvariants verifies that every declared SecretScope constant
+// satisfies the invariants documented on the SecretScope type:
+//   - non-empty
+//   - contains no underscores (underscore is the delimiter in "__thv_<scope>_<name>")
+func TestSecretScopeInvariants(t *testing.T) {
+	t.Parallel()
+
+	scopes := []secrets.SecretScope{
+		secrets.ScopeRegistry,
+		secrets.ScopeWorkloads,
+		secrets.ScopeAuth,
+	}
+
+	for _, scope := range scopes {
+		s := string(scope)
+		assert.NotEmpty(t, s, "scope %q must not be empty", s)
+		assert.False(t, strings.Contains(s, "_"), "scope %q must not contain underscores", s)
 	}
 }
 


pkg/secrets/scoped.go
```diff
@@ -0,0 +1,233 @@
...
+const (
+	// SystemKeyPrefix is the prefix used for all system-managed secret keys.
+	// Any key starting with this prefix is reserved for internal use.
+	// The double-underscore and trailing underscore make it visually distinct
+	// and avoid conflicts with backends (e.g. 1Password) that treat "/" as a
+	// path separator.
+	SystemKeyPrefix = "__thv_"
+
+	// ScopeRegistry is the scope for registry OAuth refresh tokens.
+	ScopeRegistry = "registry"
```
reyortiz3: nit: consider adding a scope type, e.g. `type SecretScope string`

pkg/secrets/scoped.go
```diff
@@ -0,0 +1,233 @@
...
+// ErrReservedKeyName is returned when a user command attempts to manage a
+// secret whose name is reserved for system use.
+var ErrReservedKeyName = errors.New("secret name is reserved for system use and cannot be managed via user commands")
+
+// ScopedProvider wraps a Provider and namespaces all operations under a
+// system-managed scope prefix ("__thv_<scope>_"). It is intended for
+// internal callers such as registry auth and workload auth that need isolated
+// key spaces inside the shared secrets store.
+type ScopedProvider struct {
+	inner Provider
```
reyortiz3: wouldn't `provider Provider` be more clear?

pkg/secrets/types.go
```diff
@@ -54,6 +54,8 @@ type Provider interface {
 	SetSecret(ctx context.Context, name, value string) error
 	DeleteSecret(ctx context.Context, name string) error
 	ListSecrets(ctx context.Context) ([]SecretDescription, error)
+	// BulkDeleteSecrets removes all named keys. Read-only providers treat this as a no-op.
+	BulkDeleteSecrets(ctx context.Context, keys []string) error
```
reyortiz3: nit: `DeleteSecrets`?

pkg/secrets/scoped.go
```diff
@@ -0,0 +1,233 @@
...
+	return u.inner.BulkDeleteSecrets(ctx, toDelete)
+}
+
+// Capabilities delegates to the underlying provider.
+func (u *UserProvider) Capabilities() ProviderCapabilities {
+	return u.inner.Capabilities()
+}
+
+// scopedKey builds the internal storage key in the form "__thv_<scope>_<name>".
+func scopedKey(scope, name string) string {
```
reyortiz3: nit: why not
```
func (s *ScopedProvider) getScopedKey(name string) string {
   return SystemKeyPrefix + s.scope + "_" + name
}
```

pkg/secrets/scoped.go
```diff
@@ -0,0 +1,233 @@
...
+// Capabilities delegates to the underlying provider.
+func (s *ScopedProvider) Capabilities() ProviderCapabilities {
+	return s.inner.Capabilities()
+}
+
+// UserProvider wraps a Provider and hides all system-reserved keys from
+// user-facing callers (CLI, API, MCP tool server). Any attempt to read or
+// modify a key that starts with the system prefix is rejected with
+// ErrReservedKeyName.
+type UserProvider struct {
```
reyortiz3, line 155: now reading the PR description if I understand correctly, `ScopedProvider` is basically a System-secrets Provider vs User-secrets provider, if so, wouldn't be more appropriate to name them as such? e.g. `SystemProvider` vs `UserProvider`. Also suggest to have system/scoped and user in different files.

pkg/secrets/scoped.go
```diff
@@ -0,0 +1,233 @@
...
+// other secrets untouched.
+func (s *ScopedProvider) Cleanup() error {
+	ctx := context.Background()
+
+	all, err := s.inner.ListSecrets(ctx)
+	if err != nil {
+		return err
+	}
+
+	prefix := scopedKey(s.scope, "")
```
reyortiz3: also, to avoid this `""`,  can defined a separate small helper, e.g. `s.getScopePrefix()`

pkg/secrets/encrypted_test.go
```diff
@@ -370,7 +370,61 @@ func TestEncryptedManager_Concurrency(t *testing.T) {
...
+
+	_, err = manager.GetSecret(ctx, "key2")
+	assert.Error(t, err, "key2 should have been deleted")
+	assert.Contains(t, err.Error(), "not found", "Error should indicate key2 was not found")
+
+	value3, err := manager.GetSecret(ctx, "key3")
+	assert.NoError(t, err, "key3 should still exist")
+	assert.Equal(t, "value3", value3, "key3 value should be unchanged")
+
+	// Case 2: Persists to disk — reload from the same file and verify only key3 remains.
```
reyortiz3: shouldn't these cases be separate unit tests?

pkg/secrets/scoped.go
```diff
@@ -0,0 +1,233 @@
...
+// secrets managed by ScopedProvider.
+func NewUserProvider(inner Provider) Provider {
+	return &UserProvider{inner: inner}
+}
+
+// GetSecret returns the secret for name, or ErrReservedKeyName if the name is
+// a system-reserved key.
+func (u *UserProvider) GetSecret(ctx context.Context, name string) (string, error) {
+	if isSystemKey(name) {
+		return "", fmt.Errorf("%w: %q", ErrReservedKeyName, name)
```
reyortiz3: it would be useful to have the actual operation (get/set/delete) in the error message somehow so it's easier to see which operation is failing

pkg/secrets/scoped.go
```diff
@@ -0,0 +1,233 @@
...
+func (s *ScopedProvider) ListSecrets(ctx context.Context) ([]SecretDescription, error) {
+	all, err := s.inner.ListSecrets(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	prefix := scopedKey(s.scope, "")
+	var result []SecretDescription
+	for _, desc := range all {
+		if strings.HasPrefix(desc.Key, prefix) {
```
reyortiz3, line 94: do we have to worry about case sensitivity? i.e. could a key be like `__THV_REGISTRY_*`?

pkg/secrets/scoped.go
amirejaz: Good point on the naming\! I've kept `ScopedProvider` instead of `SystemProvider` for two reasons: (1) it accurately describes the *mechanism* (namespacing by scope) rather than the audience, making it clear that different scopes (registry, workloads, auth) are isolated from each other; (2) callers already refer to it via `NewScopedProvider(inner, ScopeWorkloads)` which reads naturally. Splitting into separate files is also a good idea — I'll track that as a follow-up refactor once the migration stack lands.

pkg/secrets/scoped.go
amirejaz: No, case sensitivity isn't a concern here. The system key prefix `__thv_` is only ever written by our own code (in `getScopedKey`) and never constructed from user input. User-facing commands (CLI/API) go through `UserProvider` which checks `strings.HasPrefix(name, "__thv_")` — a case-sensitive check — so a user cannot accidentally slip past the guard with `__THV_` either, since we'd never write such a key in the first place.

pkg/secrets/scoped.go
```diff
@@ -0,0 +1,241 @@
...
+	// ScopeRegistry is the scope for registry OAuth refresh tokens.
+	ScopeRegistry SecretScope = "registry"
+
+	// ScopeWorkloads is the scope for remote workload authentication tokens
+	// (OAuth client secrets, bearer tokens, OAuth refresh tokens).
+	ScopeWorkloads SecretScope = "workloads"
+
+	// ScopeAuth is reserved for enterprise CLI/Desktop login tokens.
+	ScopeAuth SecretScope = "auth"
+)
```
aponcedeleonch, line 44: The `SecretScope` constants look good, but we should add a doc comment to the type (or a package-level note) explicitly stating two invariants: (1) a scope must be non-empty, and (2) a scope must not contain `_`, because the key format `__thv_<scope>_<name>` uses `_` as its only delimiter — a scope containing `_` would make the boundary ambiguous and could cause silent key collisions.

Since these constraints can't be enforced by the type system alone, it would be worth adding a unit test that validates all declared `SecretScope` constants satisfy these invariants. That way, if someone adds a new scope in the future the test fails immediately rather than causing a subtle bug at runtime. Something like:

```go
func TestSecretScopeInvariants(t *testing.T) {
    scopes := []SecretScope{ScopeRegistry, ScopeWorkloads, ScopeAuth}
    for _, s := range scopes {
        require.NotEmpty(t, string(s), "SecretScope must not be empty")
        require.NotContains(t, string(s), "_", "SecretScope must not contain '_' (used as delimiter in key format)")
    }
}
```

This makes the constraint self-documenting and regression-proof.

pkg/secrets/scoped.go
```diff
@@ -0,0 +1,241 @@
...
+
+// ErrReservedKeyName is returned when a user command attempts to manage a
+// secret whose name is reserved for system use.
+var ErrReservedKeyName = errors.New("secret name is reserved for system use and cannot be managed via user commands")
```
aponcedeleonch, line 48: We need to document the migration story for users who already have secrets whose names start with `__thv_` before this change ships. After `UserProvider` is wired in (Phase 4), those users will hit a silent cliff:
- `GetSecret` → `ErrReservedKeyName` (can't read)
- `ListSecrets` → key silently omitted (can't discover)
- `DeleteSecret` → `ErrReservedKeyName` (can't clean up)

The secret becomes permanently inaccessible through normal CLI/API flows with no recovery path.

Two options worth considering:
1. Add a preflight check during startup (or on the first `UserProvider` operation) that scans existing keys and fails fast with an actionable error message if any user-owned keys match the `__thv_*` pattern, prompting the user to rename or delete them before upgrading.
2. Document in the PR / release notes that secrets with names starting with `__thv_` are unsupported and must be renamed before upgrading to this version.

Either way, please make this explicit — silently breaking access to a user's secret is a bad UX cliff.

pkg/secrets/scoped.go
```diff
@@ -0,0 +1,241 @@
...
+	var result []SecretDescription
+	for _, desc := range all {
+		if !isSystemKey(desc.Key) {
+			result = append(result, desc)
+		}
+	}
+	return result, nil
+}
+
+// DeleteSecrets removes all named keys, rejecting any that are system-reserved.
```
aponcedeleonch: Please document the failure semantics for `DeleteSecrets` when the input slice contains a mix of valid keys and reserved (`__thv_*`) keys. The current implementation does a pre-validation pass (checks all keys first, then delegates), which means it's fail-fast / all-or-nothing: if any key is reserved, no keys are deleted.

That's the right choice for a cleanup operation — partial deletes leave the system in an inconsistent state — but please add a doc comment making this contract explicit, and a unit test covering the mixed-input case so the behaviour is locked in and visible to future readers.

pkg/secrets/scoped.go
reyortiz3: Fair point. One more comment on the naming, just before making it final since would be harder to change later. My reasoning is on generally naming based on "what" it does (ie. protecting/isolating system-defined secrets) vs the "how" internal mechanics (ie. by using scopes). And the other is the inconsistency of "user" (is the audience) vs "scope" (is the mechanism).

Just fed CC and this is the output:

> The asymmetry is the strongest argument in your favor. The file already has both types:
> 
> UserProvider — named by who it serves (users)
> ScopedProvider — named by how it works (key prefixing)
> That inconsistency is a real design smell. A reader encountering the pair ScopedProvider / UserProvider has to mentally map two different naming axes. SystemProvider / UserProvider is immediately coherent — both names answer "who is this for?"
> 
> Where the reviewer has a point: ScopedProvider with a SecretScope parameter hints that there are multiple system scopes (registry, workloads, auth), not just one monolithic system bucket. SystemProvider could mislead a reader into thinking it's a singleton concern. That said, this is fully addressed by the constructor signature NewSystemProvider(inner, ScopeWorkloads) and the docstring — the type name doesn't need to carry that weight.
> 
> My assessment: Your "what vs how" framing is the right principle. Names should reflect the concept in the domain model, not the implementation mechanism. The mechanism (key namespacing) belongs in the doc comment, which it already has. Rename ScopedProvider → SystemProvider for symmetry and clarity — the reviewer's counter-arguments are valid but not strong enough to override the consistency benefit.

But feel free to make the final call.

PR #4236, opened 2026-03-19
feat(session): add NewManagerWithRedis constructor (RC-7)

### LINKED ISSUE #4208: Wire Redis Backend Selection into session.Manager (RC-7)
## Description

Add a factory constructor to `pkg/transport/session/manager.go` that selects `RedisStorage` when a `RedisConfig` is provided and falls back to `LocalStorage` otherwise. This wires the Redis storage backend implemented in RC-6 into the session manager lifecycle, enabling the manager to be constructed with an externalized Redis backend without changing the signatures of any existing constructors (`NewManager`, `NewTypedManager`).

This task is the second layer of the THV-0047 horizontal scaling foundation. Once complete, the vMCP server can construct a Redis-backed session manager by passing a `RedisConfig`, and all existing callers that do not supply a config continue to use in-memory `LocalStorage` without modification.

## Context

`pkg/transport/session/manager.go` currently exposes three constructors: `NewManager`, `NewTypedManager`, and `NewManagerWithStorage`. The first two always use `NewLocalStorage()` as the storage backend. `NewManagerWithStorage` is a lower-level extension point that accepts any `Storage` implementation. RC-7 adds a higher-level convenience constructor — `NewManagerWithRedis` or similar — that internally constructs a `RedisStorage` from a `RedisConfig` and calls `NewManagerWithStorage`, keeping existing callers untouched.

Parent epic: stacklok/stacklok-epics#262

**Dependencies**: #266 (RC-6: Redis Storage Backend for transport/session)
**Blocks**: TASK-003 (RC-8: Persist vMCP Session Metadata to Redis on Creation)

## Acceptance Criteria

- [ ] A new exported constructor (e.g. `NewManagerWithRedis`) exists in `pkg/transport/session/manager.go` that accepts a `RedisConfig` (or a pointer) and a `Factory` and returns `(*Manager, error)`
- [ ] When `RedisConfig` is non-nil (or the pointer is non-nil), the constructor calls `NewRedisStorage` (from RC-6) and passes the resulting `RedisStorage` to `NewManagerWithStorage`
- [ ] The TTL passed to `NewManagerWithStorage` is also used as the Redis key TTL so that session expiry in Redis matches the manager's cleanup interval
- [ ] `NewManager` signature is unchanged and continues to use `NewLocalStorage` internally
- [ ] `NewTypedManager` signature is unchanged and continues to use `NewLocalStorage` internally
- [ ] `NewManagerWithStorage` signature is unchanged (still the general extension point for custom storage)
- [ ] All existing callers of `NewManager`, `NewTypedManager`, and `NewManagerWithStorage` compile and pass tests without modification
- [ ] Unit tests cover: Redis config present → `RedisStorage` is used; `nil` or absent config → `LocalStorage` is used
- [ ] Unit tests verify the constructor returns an error when `NewRedisStorage` fails (e.g. invalid config)
- [ ] All tests pass (`go test ./pkg/transport/session/...`)

Add one new function to `pkg/transport/session/manager.go`:

```
NewManagerWithRedis(ttl time.Duration, factory Factory, cfg RedisConfig) (*Manager, error)
```

Inside, call `NewRedisStorage(cfg, ttl)` — the constructor defined in TASK-001. If `NewRedisStorage` returns an error (bad config, TLS cert load failure), propagate it. Otherwise pass the `RedisStorage` to `NewManagerWithStorage` and return the manager.

Because `NewManagerWithStorage` already starts the cleanup goroutine, no additional lifecycle code is needed. For Redis-backed storage the cleanup goroutine calls `DeleteExpired` which is a no-op (Redis TTL handles expiry), so there is no performance concern.

The call site that needs to opt in to Redis (`pkg/vmcp/server/server.go`, line 382) will be updated in a later task. This task only adds the constructor; the server wiring is deferred.

### Patterns & Frameworks

- Follow the existing constructor pattern in `pkg/transport/session/manager.go`: construct a `Manager` struct literal, start `go m.cleanupRoutine()`, return
- Use `NewManagerWithStorage` as the sole internal extension point — do not duplicate the manager struct initialization
- Return `(*Manager, error)` from the new constructor to propagate `NewRedisStorage` errors; all existing constructors return `*Manager` (no error) because `NewLocalStorage` is infallible
- Test helper pattern: follow `pkg/authserver/storage/redis_test.go`'s `newTestRedisStorage` / `withRedisStorage` pattern using `github.com/alicebob/miniredis/v2`

### Code Pointers

- `pkg/transport/session/manager.go` — file to modify; add `NewManagerWithRedis` after `NewManagerWithStorage` (line 104); the `ttl` field is the value to forward to `NewRedisStorage` as the key TTL
- `pkg/transport/session/storage_redis.go` — new file from TASK-001; provides `NewRedisStorage(cfg RedisConfig, ttl time.Duration) (*RedisStorage, error)` which this constructor calls
- `pkg/transport/session/redis_config.go` — new file from TASK-001; provides the `RedisConfig` type accepted by the new constructor
- `pkg/transport/session/storage_local.go` — shows what `NewLocalStorage` returns for comparison; the fallback path that existing callers rely on
- `pkg/transport/session/storage_test.go` — existing test patterns for storage and manager; new tests for `NewManagerWithRedis` should follow the same table-driven, `t.Parallel()` style
- `pkg/authserver/storage/redis_test.go` — `newTestRedisStorage` / `withRedisStorage` helper pattern using `miniredis.RunT(t)`; adapt for session-layer tests
- `pkg/vmcp/server/server.go` line 382 — current `NewManager` call that will eventually be replaced; this task does NOT change this call site

### Component Interfaces

```go
// NewManagerWithRedis creates a session manager backed by Redis.
// cfg supplies the Redis connection configuration; ttl is applied as both the
// manager's cleanup interval and the Redis key TTL.
// Returns an error if the Redis client cannot be constructed (e.g. invalid TLS config).
// Callers that do not require Redis should continue to use NewManager or NewTypedManager.
func NewManagerWithRedis(ttl time.Duration, factory Factory, cfg RedisConfig) (*Manager, error) {
    storage, err := NewRedisStorage(cfg, ttl)
    if err != nil {
        return nil, fmt.Errorf("creating redis storage: %w", err)
    }
    return NewManagerWithStorage(ttl, factory, storage), nil
}
```

No changes to existing signatures:

```go
// Unchanged — existing callers continue to compile without modification.
func NewManager(ttl time.Duration, factory interface{}) *Manager
func NewTypedManager(ttl time.Duration, sessionType SessionType) *Manager
func NewManagerWithStorage(ttl time.Duration, factory Factory, storage Storage) *Manager
```

### Testing Strategy

Use `github.com/alicebob/miniredis/v2` (`miniredis.RunT(t)`) for all Redis-backed tests. No live Redis connection is required.

**Unit Tests**
- [ ] `NewManagerWithRedis` — valid config pointing at a miniredis instance → returns non-nil `*Manager` and nil error
- [ ] `NewManagerWithRedis` — `RedisConfig` with an unreachable address → construction should succeed (go-redis defers connection); verify that the returned manager is non-nil (lazy connection is the go-redis default)
- [ ] `NewManagerWithRedis` — `RedisConfig` with invalid TLS PEM in `CACert` → returns a non-nil error
- [ ] `NewManagerWithRedis` — round-trip: store a session via `manager.AddWithID`, retrieve it via `manager.Get`; verify the session can be found (exercises the Redis path end-to-end with miniredis)
- [ ] `NewManager` — verify it still returns a manager backed by `LocalStorage` (type assertion `m.storage.(*LocalStorage)` succeeds)
- [ ] `NewTypedManager` — verify it still returns a manager backed by `LocalStorage`

**Integration Tests**
- [ ] None required at this layer; unit tests with miniredis provide sufficient coverage

**Edge Cases**
- [ ] Calling `Stop()` on a Redis-backed manager closes the underlying Redis client (go-redis `client.Close()` is called via `RedisStorage.Close()`)
- [ ] `DeleteExpired` called by the cleanup goroutine on a Redis-backed manager is a no-op and does not return an error

## Out of Scope

- Implementing `RedisStorage` itself (TASK-001 / RC-6)
- Changing the call site in `pkg/vmcp/server/server.go` to use `NewManagerWithRedis` (deferred to a later wiring task, or done as part of TASK-003+)
- Any changes to `pkg/vmcp` packages
- Redis Cluster support
- Operational concerns: Redis provisioning, Kubernetes configuration

## References

- RFC THV-0047: https://github.com/stacklok/toolhive-rfcs/pull/47
- Parent epic: stacklok/stacklok-epics#262
- Upstream dependency (RC-6): stacklok/stacklok-epics#266
- `Storage` interface: `pkg/transport/session/storage.go`
- `NewManagerWithStorage` extension point: `pkg/transport/session/manager.go` (line 104)
- Authserver Redis test helper pattern: `pkg/authserver/storage/redis_test.go`

yrobla:
## Summary

Wires the Redis storage backend into the session manager lifecycle (RC-7), and establishes UUID as the canonical session ID format across all storage backends and proxy transports.

**NewManagerWithRedis constructor (RC-7)**
- `NewManagerWithRedis(ctx, ttl, factory, cfg)` accepts a `RedisConfig` and constructs a Redis-backed `*Manager` via `NewRedisStorage` + `NewManagerWithStorage`
- `ctx` is used for the initial Ping during construction (improvement over the original issue spec, which did not thread context through)
- Existing `NewManager`, `NewTypedManager`, and `NewManagerWithStorage` signatures are unchanged

**UUID enforcement (breaking API change — see compatibility notes below)**
- `Manager` now validates that all session IDs are UUID-format on `AddWithID`, `AddSession`, `UpsertSession`, and `Delete`
- `validateSessionID` is an unexported helper that enforces this consistently across all storage backends (previously `RedisStorage` alone had per-operation UUID checks, creating a store-then-can't-load inconsistency)
- `RedisStorage` no longer does its own UUID checks; validation lives exclusively in `Manager`

**Non-UUID Mcp-Session-Id normalization in transparent proxy**
- The MCP spec does not mandate UUID format for `Mcp-Session-Id`; upstream servers may issue arbitrary strings
- `normalizeSessionID` (new, in `pkg/transport/proxy/transparent`) maps any non-UUID external session ID to a deterministic UUID v5 (RFC 4122 URL namespace), so the transparent proxy can store upstream-issued IDs in the UUID-only session manager without a reverse-mapping table
- Applied at all three callsites where the transparent proxy reads `Mcp-Session-Id` from upstream responses or SSE stream data

Fixes #4208

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/transport/session/manager.go` | Add `NewManagerWithRedis`, `validateSessionID`; wire UUID validation into all manager methods |
| `pkg/transport/session/storage_redis.go` | Remove per-operation UUID checks (moved to Manager); add `Username` ACL field wiring; add `KeyPrefix` trailing-colon validation; sliding-window TTL godoc |
| `pkg/transport/session/redis_config.go` | Add `Username` field for Redis 6+ ACL auth |
| `pkg/transport/proxy/transparent/session_id.go` | New: `normalizeSessionID` — UUID pass-through or deterministic UUID v5 derivation |
| `pkg/transport/proxy/transparent/transparent_proxy.go` | Apply `normalizeSessionID` at all `Mcp-Session-Id` callsites |
| `pkg/transport/proxy/transparent/sse_response_processor.go` | Apply `normalizeSessionID` before `AddWithID` |
| Tests | Updated throughout to use UUID-format session IDs; new tests for Redis round-trips, ACL auth, UUID validation, and `normalizeSessionID` |

No direct user-facing change. Internal plumbing for Redis-backed horizontal scaling (THV-0047). The Redis constructor is not yet wired into the vMCP server (deferred to RC-8).

**Compatibility**: `Manager.AddWithID`, `AddSession`, `UpsertSession`, and `Delete` now reject non-UUID IDs with `"invalid session ID format"`. All existing toolhive call sites generate UUIDs internally, so there is no runtime regression. The transparent proxy is the only place that ingests externally-sourced session IDs; `normalizeSessionID` handles that path.

**Sliding-window TTL**: `RedisStorage` uses `GETEX` on every `Load` to refresh the TTL, meaning sessions remain alive indefinitely while actively used. Revocation requires an explicit `Delete`. No absolute max lifetime is enforced; a `MaxLifetime`/`CreatedAt` check is the noted path forward if a hard cap is needed.

### DIFF
--- a/pkg/transport/session/manager.go
+++ b/pkg/transport/session/manager.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"log/slog"
 	"time"
+
+	"github.com/google/uuid"
 )
 
 const (
@@ -114,19 +116,14 @@ func NewManagerWithStorage(ttl time.Duration, factory Factory, storage Storage) 
 }
 
 // NewManagerWithRedis creates a session manager backed by Redis.
+// ctx is used for the initial Ping during construction and should carry any
+// deadline appropriate for the connection attempt (e.g. a startup timeout).
 // cfg supplies the Redis connection configuration; ttl is applied as both the
 // manager's cleanup interval and the Redis key TTL.
 // Returns an error if the Redis client cannot be constructed (e.g. invalid config or TLS error).
-// Callers that do not require Redis should continue to use NewManager or NewTypedManager.
-//
-// cfg is intentionally passed by value (not *RedisConfig): NewManagerWithRedis is a dedicated
-// Redis constructor. Callers that want LocalStorage should use NewManager or NewTypedManager
-// directly; there is no "nil config falls back to local" path by design.
-//
-// context.Background() is passed to NewRedisStorage because the Ping timeout is already bounded
-// by cfg.DialTimeout (default 5 s). A redundant context deadline at this layer would add no value.
-func NewManagerWithRedis(ttl time.Duration, factory Factory, cfg RedisConfig) (*Manager, error) {
-	storage, err := NewRedisStorage(context.Background(), cfg, ttl)
+// Callers that do not require Redis should use NewManager or NewTypedManager instead.
+func NewManagerWithRedis(ctx context.Context, ttl time.Duration, factory Factory, cfg RedisConfig) (*Manager, error) {
+	storage, err := NewRedisStorage(ctx, cfg, ttl)
 	if err != nil {
 		return nil, fmt.Errorf("creating redis storage: %w", err)
 	}
@@ -151,11 +148,23 @@ func (m *Manager) cleanupRoutine() {
 	}
 }
 
-// AddWithID creates (and adds) a new session with the provided ID.
-// Returns error if ID is empty or already exists.
-func (m *Manager) AddWithID(id string) error {
+// validateSessionID returns an error if id is empty or not a valid UUID.
+// UUID format is enforced across all storage backends to keep ID semantics consistent.
+func validateSessionID(id string) error {
 	if id == "" {
 		return fmt.Errorf("session ID cannot be empty")
+	}
+	if _, err := uuid.Parse(id); err != nil {
+		return fmt.Errorf("invalid session ID format: %w", err)
+	}
+	return nil
+}
+
+// AddWithID creates (and adds) a new session with the provided ID.
+// Returns error if ID is empty, not a valid UUID, or already exists.
+func (m *Manager) AddWithID(id string) error {
+	if err := validateSessionID(id); err != nil {
+		return err
 	}
 	// Check if session already exists
 	ctx, cancel := context.WithTimeout(context.Background(), defaultOperationTimeout)
@@ -176,8 +185,8 @@ func (m *Manager) AddSession(session Session) error {
 	if session == nil {
 		return fmt.Errorf("session cannot be nil")
 	}
-	if session.ID() == "" {
-		return fmt.Errorf("session ID cannot be empty")
+	if err := validateSessionID(session.ID()); err != nil {
+		return err
 	}
 
 	// Check if session already exists
@@ -213,8 +222,8 @@ func (m *Manager) UpsertSession(session Session) error {
 	if session == nil {
 		return fmt.Errorf("session cannot be nil")
 	}
-	if session.ID() == "" {
-		return fmt.Errorf("session ID cannot be empty")
+	if err := validateSessionID(session.ID()); err != nil {
+		return err
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), defaultOperationTimeout)
 	defer cancel()
@@ -222,8 +231,11 @@ func (m *Manager) UpsertSession(session Session) error {
 }
 
 // Delete removes a session by ID.
-// Returns an error if the deletion fails.
+// Returns an error if the ID is invalid or the deletion fails.
 func (m *Manager) Delete(id string) error {
+	if err := validateSessionID(id); err != nil {
+		return err
+	}
 	ctx, cancel := context.WithTimeout(context.Background(), defaultOperationTimeout)
 	defer cancel()
 	return m.storage.Delete(ctx, id)
--- a/pkg/transport/session/redis_config.go
+++ b/pkg/transport/session/redis_config.go
@@ -21,8 +21,14 @@ type RedisConfig struct {
 	// SentinelConfig, when non-nil, activates Sentinel mode. Mutually exclusive with Addr.
 	SentinelConfig *SentinelConfig
 
-	// Password is the Redis AUTH password.
-	Password string
+	// Username is the Redis ACL username (Redis 6.0+). When non-empty, both
+	// Username and Password are sent as ACL credentials (AUTH username password).
+	// Leave empty to authenticate as the default user (legacy AUTH password).
+	Username string
+
+	// Password is the Redis AUTH password. Used with Username for ACL auth,
+	// or alone for legacy AUTH with the default user.
+	Password string //nolint:gosec // G101: not a hardcoded credential
 
 	// DB is the Redis database index.
 	DB int
--- a/pkg/transport/session/storage_redis.go
+++ b/pkg/transport/session/storage_redis.go
@@ -39,11 +39,16 @@ func validateRedisConfig(cfg *RedisConfig) error {
 	if cfg.KeyPrefix == "" {
 		return errors.New("KeyPrefix is required")
 	}
+	if cfg.KeyPrefix[len(cfg.KeyPrefix)-1] != ':' {
+		return errors.New("KeyPrefix must end with ':' to avoid key collisions (e.g. \"thv:vmcp:session:\")")
+	}
 	return nil
 }
 
 // NewRedisStorage constructs a RedisStorage from a RedisConfig.
-// ttl is the expiry applied to every key on Store.
+// ttl is the expiry applied to every key on Store and refreshed on every Load (sliding window).
+// Because TTL is sliding, sessions remain valid indefinitely while actively used; revocation
+// requires an explicit Delete call. There is no absolute maximum session lifetime.
 func NewRedisStorage(ctx context.Context, cfg RedisConfig, ttl time.Duration) (*RedisStorage, error) {
 	if err := validateRedisConfig(&cfg); err != nil {
 		return nil, fmt.Errorf("invalid redis configuration: %w", err)
@@ -73,6 +78,7 @@ func NewRedisStorage(ctx context.Context, cfg RedisConfig, ttl time.Duration) (*
 		opts := &redis.FailoverOptions{
 			MasterName:    cfg.SentinelConfig.MasterName,
 			SentinelAddrs: cfg.SentinelConfig.SentinelAddrs,
+			Username:      cfg.Username,
 			Password:      cfg.Password,
 			DB:            cfg.DB,
 			DialTimeout:   cfg.DialTimeout,
@@ -84,6 +90,7 @@ func NewRedisStorage(ctx context.Context, cfg RedisConfig, ttl time.Duration) (*
 	} else {
 		opts := &redis.Options{
 			Addr:         cfg.Addr,
+			Username:     cfg.Username,
 			Password:     cfg.Password,
 			DB:           cfg.DB,
 			DialTimeout:  cfg.DialTimeout,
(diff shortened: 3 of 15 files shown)


pkg/transport/session/storage_redis.go
```diff
@@ -0,0 +1,205 @@
...
+		if !pool.AppendCertsFromPEM(cfg.CACert) {
+			return nil, fmt.Errorf("failed to parse CA certificate PEM data")
+		}
+		tc.RootCAs = pool
+	}
+	return tc, nil
+}
+
+func (s *RedisStorage) key(id string) string {
+	return s.keyPrefix + id
```
jhrozek, line 147: **nit(security):** The session ID used here comes from the client-controlled `Mcp-Session-Id` HTTP request header and flows through `Manager.Get`/`Delete` → `Storage.Load`/`Delete` → `key()` with no format validation anywhere in the stack (I checked — the vMCP `sessionmanager.Manager.Validate()` only does an empty-string check, and the SDK's built-in format validation is replaced by the custom manager).

Since session IDs are always generated as UUID v4, consider validating the format at the storage or manager layer to prevent a crafted header value from targeting arbitrary Redis keys. This could be a simple UUID regex check in `Store`/`Load`/`Delete`, or in `Manager.AddWithID`/`AddSession` at the creation boundary.

pkg/transport/session/redis_config.go
```diff
@@ -0,0 +1,61 @@
...
+
+// RedisConfig configures the Redis storage backend for session storage.
+// Addr is used for standalone; SentinelConfig activates Sentinel mode (mutually exclusive).
+type RedisConfig struct {
+	// Addr is the Redis server address for standalone mode (e.g., "host:port").
+	Addr string
+
+	// SentinelConfig, when non-nil, activates Sentinel mode. Mutually exclusive with Addr.
+	SentinelConfig *SentinelConfig
+
```
jhrozek, line 23: **suggestion:** Redis 6.0+ ACL auth uses username + password. This config only has `Password`, which means deployments using ACL users (or AWS ElastiCache/MemoryDB IAM auth) can't specify a non-`default` username.

For reference, the authserver's Redis implementation (`pkg/authserver/storage/redis.go`) handles this with a separate `ACLUserConfig` struct containing both `Username` and `Password` fields, wired into `FailoverOptions.Username`/`Password`. Consider adding a `Username` field here for parity.

pkg/transport/session/storage_redis.go
```diff
@@ -0,0 +1,205 @@
...
+
+// Load retrieves a session by ID. Returns ErrSessionNotFound when the key does not exist.
+// The Redis eviction TTL is refreshed atomically via GETEX on every read so that active
+// sessions are not evicted between accesses. The session's UpdatedAt timestamp is not modified.
+func (s *RedisStorage) Load(ctx context.Context, id string) (Session, error) {
+	if id == "" {
+		return nil, fmt.Errorf("cannot load session with empty ID")
+	}
+
+	data, err := s.client.GetEx(ctx, s.key(id), s.ttl).Bytes()
```
jhrozek, line 182: **suggestion(doc):** The `GETEX` sliding-window TTL means a session accessed at least once per TTL window will never expire — it can live indefinitely as long as the client keeps making requests. This is a reasonable design for keeping active sessions alive, but the security implication (a compromised token remains valid indefinitely while in use) should be documented explicitly.

Consider adding a note to the `Load` godoc and/or `NewRedisStorage` that session revocation depends entirely on explicit `Delete` calls, and that there is no absolute maximum session lifetime. If that's needed later, a `MaxLifetime` field checked in `Load` would be the path forward.

pkg/transport/session/manager.go
```diff
@@ -113,6 +113,26 @@ func NewManagerWithStorage(ttl time.Duration, factory Factory, storage Storage)
...
+// Returns an error if the Redis client cannot be constructed (e.g. invalid config or TLS error).
+// Callers that do not require Redis should continue to use NewManager or NewTypedManager.
+//
+// cfg is intentionally passed by value (not *RedisConfig): NewManagerWithRedis is a dedicated
+// Redis constructor. Callers that want LocalStorage should use NewManager or NewTypedManager
+// directly; there is no "nil config falls back to local" path by design.
+//
+// context.Background() is passed to NewRedisStorage because the Ping timeout is already bounded
+// by cfg.DialTimeout (default 5 s). A redundant context deadline at this layer would add no value.
+func NewManagerWithRedis(ttl time.Duration, factory Factory, cfg RedisConfig) (*Manager, error) {
```
jhrozek: **suggestion:** `NewManagerWithRedis` passes `context.Background()` to `NewRedisStorage`. The comment argues `DialTimeout` already bounds the Ping, but `DialTimeout` only governs TCP connection establishment — not the full Ping round-trip (e.g., a stalled TLS negotiation or auth exchange can hang longer).

Since this is a new public API, consider accepting `context.Context` as the first parameter before it's published — it's the standard Go pattern for constructors that do I/O, and it's impossible to add without a breaking change later. `NewRedisStorage` already accepts a context; the PR discards it at this layer.

pkg/transport/session/storage_redis.go
```diff
@@ -0,0 +1,205 @@
...
+		pool := x509.NewCertPool()
+		if !pool.AppendCertsFromPEM(cfg.CACert) {
+			return nil, fmt.Errorf("failed to parse CA certificate PEM data")
+		}
+		tc.RootCAs = pool
+	}
+	return tc, nil
+}
+
+func (s *RedisStorage) key(id string) string {
```
jhrozek, line 146: **nit:** `key()` concatenates `keyPrefix + id` with no enforcement that the prefix ends with a delimiter. A caller using `KeyPrefix: "thvsession"` with ID `"abc"` produces `"thvsessionabc"`, which could collide with prefix `"thv"` + ID `"sessionabc"`.

The godoc says `(e.g., "thv:vmcp:session:")` but this is advisory. Consider validating in `validateRedisConfig` that `KeyPrefix` ends with `:` to make the requirement explicit.

PR #4238, opened 2026-03-19
Preserve tool annotations and output schema in GetAdaptedTools

### LINKED ISSUE #4235: SessionManager.GetAdaptedTools drops tool annotations
## Summary

`SessionManager.GetAdaptedTools()` in `pkg/vmcp/server/sessionmanager/session_manager.go` creates `mcp.Tool` structs but **omits the `Annotations` field**. This causes Cedar authorization policies that rely on tool annotations (e.g. `readOnlyHint`, `destructiveHint`, `openWorldHint`) to fail — every tool is denied because the resource entity has no annotation attributes.

## Root Cause

In `GetAdaptedTools()` (line ~351), the `mcp.Tool` is constructed with only `Name`, `Description`, and `RawInputSchema`:

```go
tool := mcp.Tool{
    Name:           domainTool.Name,
    Description:    domainTool.Description,
    RawInputSchema: schemaJSON,
}
```

The `domainTool` (`vmcp.Tool`) **does** have `Annotations *ToolAnnotations` populated from backend discovery, but it is never copied to the `mcp.Tool`. The existing `capability_adapter.go` correctly includes annotations:

```go
// pkg/vmcp/server/adapter/capability_adapter.go:73
Annotations: conversion.ToMCPToolAnnotations(tool.Annotations),
```

But the session manager bypasses the capability adapter and has its own inline tool adaptation that is incomplete.

## Impact

- **All Cedar `call_tool` policies with annotation `when` clauses fail** — `resource has readOnlyHint` is always false because the attribute is never set on the entity.
- The `safe-tools` authorization profile (which permits `readOnlyHint=true` tools) filters out **every tool**, resulting in an empty `tools/list` response.
- The `observe` profile (list-only, no `call_tool`) is unaffected since it doesn't evaluate `call_tool` policies.
- `full-access` is unaffected since it has no authz middleware.

## Reproduction

1. Run a vmcp proxy with Cedar authz policies that use annotation `when` clauses (e.g. brood-box's `safe-tools` profile)
2. Query `tools/list` — returns empty list
3. Check debug logs — every `cedar decision` for `call_tool` is `deny` with empty diagnostics (no policy matched)

Debug log excerpt:
```json
{"level":"DEBUG","msg":"cedar authorization check","principal":{"__entity":{"type":"Client","id":"anonymous"}},"action":{"__entity":{"type":"Action","id":"call_tool"}},"resource":{"__entity":{"type":"Tool","id":"context7_query-docs"}}}
{"level":"DEBUG","msg":"cedar decision","decision":"deny","diagnostic":{}}
```

## Fix

Add the missing `Annotations` (and likely `OutputSchema`) field in `GetAdaptedTools()`:

```go
tool := mcp.Tool{
    Name:           domainTool.Name,
    Description:    domainTool.Description,
    RawInputSchema: schemaJSON,
    Annotations:    conversion.ToMCPToolAnnotations(domainTool.Annotations),
}
```

`OutputSchema` should also be checked — the capability adapter includes it but the session manager likely drops it too.

Ideally, `GetAdaptedTools` should delegate to the existing `capability_adapter.ToSDKTools()` rather than duplicating the conversion inline, to prevent future field omissions.

## Affected Versions

- v0.12.2 (confirmed)
- Likely introduced in `a6ccdf1d` (Add SessionManager) or `e719f471` (Remove session v1)

JAORMX:
## Summary

- Cedar authorization policies relying on tool annotations (e.g. `readOnlyHint`, `destructiveHint`) were failing because `SessionManager.GetAdaptedTools` constructed `mcp.Tool` structs without copying `Annotations` or `OutputSchema` from the domain tool — every tool appeared annotation-less, so policies denied all requests.
- Added the missing `Annotations` field (via `conversion.ToMCPToolAnnotations`) and `RawOutputSchema` marshalling to match what `capability_adapter.go` already does.
- Added a test case covering both annotated and plain tools to prevent regression.

Fixes #4235

## Type of change

- [x] Bug fix

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/server/sessionmanager/session_manager.go` | Copy `Annotations` and `OutputSchema` when building `mcp.Tool` in `GetAdaptedTools` |
| `pkg/vmcp/server/sessionmanager/session_manager_test.go` | Add "preserves annotations and output schema" test case |

Tools exposed through vMCP sessions now correctly carry their annotations and output schema, restoring Cedar policy evaluation and ensuring MCP clients see the full tool metadata.

### DIFF
--- a/pkg/vmcp/server/sessionmanager/session_manager_test.go
+++ b/pkg/vmcp/server/sessionmanager/session_manager_test.go
@@ -861,6 +861,78 @@ func TestSessionManager_GetAdaptedTools(t *testing.T) {
 		assert.Contains(t, string(byName["alpha"].RawInputSchema), `"type"`)
 	})
 
+	t.Run("preserves annotations and output schema", func(t *testing.T) {
+		t.Parallel()
+
+		boolPtr := func(b bool) *bool { return &b }
+		tools := []vmcp.Tool{
+			{
+				Name:        "annotated",
+				Description: "tool with annotations",
+				InputSchema: map[string]any{"type": "object"},
+				OutputSchema: map[string]any{
+					"type": "object",
+					"properties": map[string]any{
+						"result": map[string]any{"type": "string"},
+					},
+				},
+				Annotations: &vmcp.ToolAnnotations{
+					Title:           "Annotated Tool",
+					ReadOnlyHint:    boolPtr(true),
+					DestructiveHint: boolPtr(false),
+				},
+			},
+			{
+				Name:        "plain",
+				Description: "tool without annotations or output schema",
+				InputSchema: map[string]any{"type": "object"},
+			},
+		}
+		ctrl := gomock.NewController(t)
+		factory := sessionfactorymocks.NewMockMultiSessionFactory(ctrl)
+		factory.EXPECT().
+			MakeSessionWithID(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
+			DoAndReturn(func(_ context.Context, id string, _ *auth.Identity, _ bool, _ []*vmcp.Backend) (vmcpsession.MultiSession, error) {
+				return newMockSession(t, ctrl, id, tools), nil
+			}).Times(1)
+
+		registry := newFakeRegistry()
+		sm, _ := newTestSessionManager(t, factory, registry)
+
+		sessionID := sm.Generate()
+		_, err := sm.CreateSession(context.Background(), sessionID)
+		require.NoError(t, err)
+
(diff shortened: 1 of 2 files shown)


pkg/vmcp/server/sessionmanager/session_manager_test.go
```diff
@@ -861,6 +861,78 @@ func TestSessionManager_GetAdaptedTools(t *testing.T) {
 		assert.Contains(t, string(byName["alpha"].RawInputSchema), `"type"`)
 	})
 
+	t.Run("preserves annotations and output schema", func(t *testing.T) {
```
jerm-dro, line 864: nice regression test 😄

Issue #4209, opened 2026-03-18
Implement configurable StatefulSet replica count (RC-11)


yrobla:
## Description

Make the StatefulSet replica count configurable by wiring `RunConfig.BackendReplicas` from `pkg/runner/config.go` through to the Kubernetes client in `pkg/container/kubernetes/client.go`, where `WithReplicas` is currently hardcoded to `1`. When `BackendReplicas` is nil, the field is omitted entirely from the server-side apply spec so that HPA or kubectl retains control of replica scaling; when it is set, the specified value is applied. This change is RC-11 from the proxyrunner horizontal scaling epic.

## Context

Part of epic [#263](https://github.com/stacklok/stacklok-epics/issues/263) — Horizontal Scaling for Proxyrunner (THV-0047). Today all Kubernetes StatefulSets are deployed with `WithReplicas(1)` hardcoded in `pkg/container/kubernetes/client.go:404`. To allow the proxyrunner to run as multiple replicas, or to leave replica management to an HPA, this hardcoded call must be made conditional on whether `RunConfig.BackendReplicas` is set. TASK-001 (#265) added the `BackendReplicas *int32` field to `RunConfig`; this task threads that value through the call chain: `RunConfig` → `runtime.Setup` → `DeployWorkloadOptions` → `client.DeployWorkload` → `StatefulSetSpec`.

**Dependencies**: #265 (Extend RunConfig with replica and cache-size fields)
**Blocks**: None (TASK-003, TASK-004 are independent; TASK-005, TASK-006 depend on TASK-004)

## Acceptance Criteria

- [ ] `DeployWorkloadOptions` in `pkg/container/runtime/types.go` has a new `BackendReplicas *int32` field
- [ ] `pkg/runtime/setup.go`'s `Setup` function accepts a `backendReplicas *int32` parameter and assigns it to `containerOptions.BackendReplicas`
- [ ] `pkg/runner/runner.go` passes `r.Config.BackendReplicas` to `runtime.Setup`
- [ ] `pkg/container/kubernetes/client.go`'s `DeployWorkload` uses `options.BackendReplicas`: when non-nil, calls `appsv1apply.StatefulSetSpec().WithReplicas(*options.BackendReplicas)`; when nil, omits `WithReplicas` entirely
- [ ] When `BackendReplicas` is nil, the applied StatefulSet spec does not include a `replicas` field (verified via server-side apply field inspection or assertion that the spec's `.Replicas` remains controlled by the previous field manager)
[...]

PR #4239, opened 2026-03-19
Add BackendReplicas and SessionCacheSize to RunConfig

### LINKED ISSUE #4203: Extend RunConfig with replica and cache-size fields
## Description

Add two new fields to `RunConfig` in `pkg/runner/config.go`: `BackendReplicas *int32` for configuring the desired StatefulSet replica count, and `SessionCacheSize int` for bounding the local LRU session cache. These fields are the foundation for all horizontal scaling work in the proxyrunner epic — downstream tasks (RC-11, RC-12, RC-13) read these values to control StatefulSet replica management and session cache sizing.

## Context

Part of epic [stacklok/stacklok-epics#263](https://github.com/stacklok/stacklok-epics/issues/263) — Horizontal Scaling for Proxyrunner (THV-0047). `RunConfig` is ToolHive's serializable workload configuration and is part of the public API contract. This task extends it with two optional fields that are zero-value-safe: nil `BackendReplicas` means "don't touch replicas" (preserving HPA/kubectl control), and zero `SessionCacheSize` will cause consuming code to apply a sensible default of 1000.

**Dependencies**: None
**Blocks**: TASK-002 (Configurable StatefulSet replica count), TASK-004 (Implement RoutingStorage)

## Acceptance Criteria

- [ ] `RunConfig` struct in `pkg/runner/config.go` has a new `BackendReplicas *int32` field with JSON tag `json:"backend_replicas,omitempty"` and YAML tag `yaml:"backend_replicas,omitempty"`
- [ ] `RunConfig` struct has a new `SessionCacheSize int` field with JSON tag `json:"session_cache_size,omitempty"` and YAML tag `yaml:"session_cache_size,omitempty"`
- [ ] Both fields are optional and zero-value-safe: existing RunConfig JSON/YAML files that do not include the new fields continue to deserialize without error and without changing existing behavior
- [ ] `BackendReplicas` is a pointer type (`*int32`) so that `nil` can be distinguished from an explicit value of `0`
- [ ] Doc comments follow the existing field comment style in `RunConfig` and explain the nil/zero-value semantics for each field
- [ ] Round-trip serialization test: a `RunConfig` with both fields set serializes to JSON/YAML and deserializes back with values preserved
- [ ] Round-trip serialization test: a `RunConfig` without the new fields (omitted) deserializes with `BackendReplicas == nil` and `SessionCacheSize == 0`
- [ ] All existing tests pass without modification

Add the two fields to the `RunConfig` struct in `pkg/runner/config.go`, following the existing field ordering and comment conventions. No changes to constructor functions, serialization helpers, or validation logic are needed — the `omitempty` tags handle backward-compatible serialization, and the zero-value semantics are defined by the consuming code in TASK-002 and TASK-004.

### Patterns & Frameworks

- Follow the existing field style in `RunConfig`: each field has a Go doc comment, JSON tag, and YAML tag. Fields that are optional use `omitempty` on both tags.
- Use `*int32` (pointer) for `BackendReplicas` to allow nil-vs-zero distinction, consistent with how Kubernetes API types represent optional integer fields (e.g., `appsv1.StatefulSetSpec.Replicas`).
- Use plain `int` (not a pointer) for `SessionCacheSize` since zero is a meaningful sentinel that consuming code maps to a default — consistent with how other integer config fields in the repo are handled.
- `RunConfig` is documented as part of ToolHive's API contract (see the struct comment at line 46); do not change the schema version `CurrentSchemaVersion` in this PR — these additions are backward-compatible and additive.

### Code Pointers

- `pkg/runner/config.go` — The `RunConfig` struct starting at line 46; add the new fields here, following the comment and tag style of existing fields (e.g., `K8sPodTemplatePatch` at line 150 is a good nearby example of a Kubernetes-specific optional field with both JSON and YAML tags)
- `pkg/transport/session/storage.go` — The `Storage` interface; understanding this helps clarify why `SessionCacheSize` is needed (it bounds the LRU cache used by `RoutingStorage` in TASK-004)
- `pkg/container/kubernetes/client.go` lines 401–404 — The `WithReplicas(1)` call that TASK-002 will make conditional on `BackendReplicas != nil`

### Component Interfaces

The two new fields to add to `RunConfig`:

```go
// BackendReplicas is the desired StatefulSet replica count for the MCP backend.
// When nil, WithReplicas is omitted from the StatefulSet apply spec so that
// HPA or kubectl retains control of the replica count.
BackendReplicas *int32 `json:"backend_replicas,omitempty" yaml:"backend_replicas,omitempty"`

// SessionCacheSize is the maximum number of sessions held in the proxyrunner's
// local LRU cache. When 0, a default of 1000 is used by the consuming code.
SessionCacheSize int `json:"session_cache_size,omitempty" yaml:"session_cache_size,omitempty"`
```

### Testing Strategy

**Unit Tests**

- [ ] Serialize a `RunConfig` with `BackendReplicas` set to a non-nil value (e.g., `int32(3)`) and `SessionCacheSize` set to `500`; unmarshal the JSON output and assert the fields round-trip correctly
- [ ] Serialize a `RunConfig` that omits both new fields (i.e., zero-valued `RunConfig{}`); assert the JSON output does not contain `backend_replicas` or `session_cache_size` keys (omitempty behavior)
- [ ] Deserialize a JSON payload that does not include `backend_replicas` or `session_cache_size`; assert `BackendReplicas == nil` and `SessionCacheSize == 0`
- [ ] Deserialize a JSON payload that sets `backend_replicas: 2`; assert `BackendReplicas != nil && *BackendReplicas == 2`

**Integration Tests**

- [ ] No integration tests required for this task — the field additions are structural only and exercised by downstream tasks

**Edge Cases**

- [ ] `backend_replicas: 0` in JSON — assert this deserializes to `BackendReplicas` pointing to `int32(0)`, not nil (important: `0` is an explicit value distinct from omitted)
- [ ] YAML round-trip with both fields set — assert values are preserved

## Out of Scope

- Validation logic for `BackendReplicas` or `SessionCacheSize` (e.g., range checks) — deferred to consuming code in TASK-002 and TASK-004
- Wiring `BackendReplicas` into the Kubernetes client (TASK-002)
- Wiring `SessionCacheSize` into `RoutingStorage` construction (TASK-004, TASK-006)
- Changes to `CurrentSchemaVersion` — these fields are backward-compatible additions
- CLI flags or operator CRD changes to expose these fields — operator/CLI concerns outside this task

## References

- [Epic stacklok/stacklok-epics#263: Horizontal Scaling for Proxyrunner (THV-0047)](https://github.com/stacklok/stacklok-epics/issues/263)
- `pkg/runner/config.go` — `RunConfig` struct (lines 46–213)
- `pkg/container/kubernetes/client.go` lines 401–404 — downstream consumer of `BackendReplicas`
- `pkg/transport/session/storage.go` — `Storage` interface; context for `SessionCacheSize`

yrobla:
## Summary

Extends RunConfig with two optional, zero-value-safe fields needed by the proxyrunner horizontal scaling epic (THV-0047). BackendReplicas (*int32, nil = unmanaged) controls StatefulSet replica count; SessionCacheSize (*int32, 0 = default 1000) bounds the local LRU session cache. Both fields use omitempty so existing configs deserialize without modification.

Fixes #4203 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
|      |        |

### DIFF
--- a/pkg/runner/config.go
+++ b/pkg/runner/config.go
@@ -211,14 +211,25 @@ type RunConfig struct {
 	// This is the serializable RunConfig; secrets are referenced by file paths or env var names.
 	EmbeddedAuthServerConfig *authserver.RunConfig `json:"embedded_auth_server_config,omitempty" yaml:"embedded_auth_server_config,omitempty"` //nolint:lll
 
+	// ScalingConfig contains configuration for horizontal scaling of the proxy runner.
+	// Only applicable when running in Kubernetes with the ToolHive operator.
+	// When nil, no scaling configuration is applied (single-replica default behavior).
+	ScalingConfig *ScalingConfig `json:"scaling_config,omitempty" yaml:"scaling_config,omitempty"`
+}
+
+// ScalingConfig contains configuration for horizontal scaling of the proxy runner backend.
+// It is intentionally kept as a separate struct so future scaling knobs (MinReplicas,
+// MaxReplicas, ScaleDownWindow, etc.) can be added here without growing the top-level
+// RunConfig shape.
+type ScalingConfig struct {
 	// BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.
-	// When omitted or null, replicas are unmanaged (preserving HPA or manual kubectl control).
-	// When set (including 0), the value is an explicit replica count passed to consuming code.
+	// When nil, replicas are unmanaged (preserving HPA or manual kubectl control).
+	// When set (including 0), the value is an explicit replica count.
 	BackendReplicas *int32 `json:"backend_replicas,omitempty" yaml:"backend_replicas,omitempty"`
 
 	// SessionCacheSize is the maximum number of sessions held in the local LRU cache.
-	// A value of 0 causes consuming code to apply a sensible default (1000).
-	SessionCacheSize int `json:"session_cache_size,omitempty" yaml:"session_cache_size,omitempty"`
+	// When nil, consuming code applies a sensible default (e.g. 1000).
+	SessionCacheSize *int32 `json:"session_cache_size,omitempty" yaml:"session_cache_size,omitempty"`
 }
 
 // WriteJSON serializes the RunConfig to JSON and writes it to the provided writer
--- a/docs/server/docs.go
+++ b/docs/server/docs.go
@@ -896,10 +896,6 @@ const docTemplate = `{
                     "aws_sts_config": {
                         "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_auth_awssts.Config"
                     },
-                    "backend_replicas": {
-                        "description": "BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.\nWhen omitted or null, replicas are unmanaged (preserving HPA or manual kubectl control).\nWhen set (including 0), the value is an explicit replica count passed to consuming code.",
-                        "type": "integer"
-                    },
                     "base_name": {
                         "description": "BaseName is the base name used for the container (without prefixes)",
                         "type": "string"
@@ -1016,6 +1012,9 @@ const docTemplate = `{
                     "runtime_config": {
                         "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_container_templates.RuntimeConfig"
                     },
+                    "scaling_config": {
+                        "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_runner.ScalingConfig"
+                    },
                     "schema_version": {
                         "description": "SchemaVersion is the version of the RunConfig schema",
                         "type": "string"
@@ -1027,10 +1026,6 @@ const docTemplate = `{
                         },
                         "type": "array",
                         "uniqueItems": false
-                    },
-                    "session_cache_size": {
-                        "description": "SessionCacheSize is the maximum number of sessions held in the local LRU cache.\nA value of 0 causes consuming code to apply a sensible default (1000).",
-                        "type": "integer"
                     },
                     "target_host": {
                         "description": "TargetHost is the host to forward traffic to (only applicable to SSE transport)",
@@ -1089,6 +1084,20 @@ const docTemplate = `{
                         },
                         "type": "array",
                         "uniqueItems": false
+                    }
+                },
+                "type": "object"
+            },
+            "github_com_stacklok_toolhive_pkg_runner.ScalingConfig": {
+                "description": "ScalingConfig contains configuration for horizontal scaling of the proxy runner.\nOnly applicable when running in Kubernetes with the ToolHive operator.\nWhen nil, no scaling configuration is applied (single-replica default behavior).",
+                "properties": {
+                    "backend_replicas": {
+                        "description": "BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.\nWhen nil, replicas are unmanaged (preserving HPA or manual kubectl control).\nWhen set (including 0), the value is an explicit replica count.",
+                        "type": "integer"
+                    },
+                    "session_cache_size": {
+                        "description": "SessionCacheSize is the maximum number of sessions held in the local LRU cache.\nWhen nil, consuming code applies a sensible default (e.g. 1000).",
+                        "type": "integer"
                     }
                 },
                 "type": "object"
--- a/docs/server/swagger.json
+++ b/docs/server/swagger.json
@@ -889,10 +889,6 @@
                     "aws_sts_config": {
                         "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_auth_awssts.Config"
                     },
-                    "backend_replicas": {
-                        "description": "BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.\nWhen omitted or null, replicas are unmanaged (preserving HPA or manual kubectl control).\nWhen set (including 0), the value is an explicit replica count passed to consuming code.",
-                        "type": "integer"
-                    },
                     "base_name": {
                         "description": "BaseName is the base name used for the container (without prefixes)",
                         "type": "string"
@@ -1009,6 +1005,9 @@
                     "runtime_config": {
                         "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_container_templates.RuntimeConfig"
                     },
+                    "scaling_config": {
+                        "$ref": "#/components/schemas/github_com_stacklok_toolhive_pkg_runner.ScalingConfig"
+                    },
                     "schema_version": {
                         "description": "SchemaVersion is the version of the RunConfig schema",
                         "type": "string"
@@ -1020,10 +1019,6 @@
                         },
                         "type": "array",
                         "uniqueItems": false
-                    },
-                    "session_cache_size": {
-                        "description": "SessionCacheSize is the maximum number of sessions held in the local LRU cache.\nA value of 0 causes consuming code to apply a sensible default (1000).",
-                        "type": "integer"
                     },
                     "target_host": {
                         "description": "TargetHost is the host to forward traffic to (only applicable to SSE transport)",
@@ -1082,6 +1077,20 @@
                         },
                         "type": "array",
                         "uniqueItems": false
+                    }
+                },
+                "type": "object"
+            },
+            "github_com_stacklok_toolhive_pkg_runner.ScalingConfig": {
+                "description": "ScalingConfig contains configuration for horizontal scaling of the proxy runner.\nOnly applicable when running in Kubernetes with the ToolHive operator.\nWhen nil, no scaling configuration is applied (single-replica default behavior).",
+                "properties": {
+                    "backend_replicas": {
+                        "description": "BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.\nWhen nil, replicas are unmanaged (preserving HPA or manual kubectl control).\nWhen set (including 0), the value is an explicit replica count.",
+                        "type": "integer"
+                    },
+                    "session_cache_size": {
+                        "description": "SessionCacheSize is the maximum number of sessions held in the local LRU cache.\nWhen nil, consuming code applies a sensible default (e.g. 1000).",
+                        "type": "integer"
                     }
                 },
                 "type": "object"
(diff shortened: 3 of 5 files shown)


pkg/runner/config.go
```diff
@@ -210,6 +210,15 @@ type RunConfig struct {
...
+
+	// BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.
+	// When omitted or null, replicas are unmanaged (preserving HPA or manual kubectl control).
+	// When set (including 0), the value is an explicit replica count passed to consuming code.
+	BackendReplicas *int32 `json:"backend_replicas,omitempty" yaml:"backend_replicas,omitempty"`
+
+	// SessionCacheSize is the maximum number of sessions held in the local LRU cache.
+	// A value of 0 causes consuming code to apply a sensible default (1000).
+	SessionCacheSize int `json:"session_cache_size,omitempty" yaml:"session_cache_size,omitempty"`
 }
```
JAORMX, line 233: So, `RunConfig` already has ~48 fields (11 of which are deprecated), and it's explicitly documented as part of ToolHive's API contract (lines 44-45). These two fields are the start of a new concern... horizontal scaling tuning. Adding them flat means every future scaling knob (`MinReplicas`, `MaxReplicas`, `ScaleDownWindow`, etc.) also goes top-level, and the struct keeps growing.\n\nWe already have a well-established pattern for optional feature-specific config: `EmbeddedAuthServerConfig`, `RemoteAuthConfig`, `RuntimeConfig`, `HeaderForwardConfig`. All pointer-to-struct with `omitempty`.\n\nI think we should follow the same pattern here:\n\n```go\n// ScalingConfig contains configuration for horizontal scaling of the proxy runner.\n// Only applicable when running in Kubernetes with the ToolHive operator.\ntype ScalingConfig struct {\n\t// BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.\n\t// When nil, replicas are unmanaged (preserving HPA or manual kubectl control).\n\t// When set (including 0), the value is an explicit replica count.\n\tBackendReplicas *int32 `json:\"backend_replicas,omitempty\" yaml:\"backend_replicas,omitempty\"`\n\n\t// SessionCacheSize is the maximum number of sessions held in the local LRU cache.\n\t// When nil, consuming code applies a sensible default (e.g. 1000).\n\tSessionCacheSize *int32 `json:\"session_cache_size,omitempty\" yaml:\"session_cache_size,omitempty\"`\n}\n```\n\nThen on `RunConfig`:\n\n```go\n// ScalingConfig contains configuration for horizontal scaling of the proxy runner.\n// When nil, no scaling configuration is applied (single-replica default behavior).\nScalingConfig *ScalingConfig `json:\"scaling_config,omitempty\" yaml:\"scaling_config,omitempty\"`\n```\n\nThis gives us:\n- `nil` = not applicable. CLI, MCPRemoteProxy, and non-scaling MCPServer workloads get no serialized noise in their ConfigMaps\n- Namespacing. Scaling knobs are grouped together, easy to find in a 50-field struct\n- Evolvability. Future scaling fields go here without touching the top-level API shape\n- Consistency with `EmbeddedAuthServerConfig`, `HeaderForward`, `RuntimeConfig`, etc.">

pkg/runner/config.go
```diff
@@ -210,6 +210,15 @@ type RunConfig struct {
...
 	EmbeddedAuthServerConfig *authserver.RunConfig `json:"embedded_auth_server_config,omitempty" yaml:"embedded_auth_server_config,omitempty"` //nolint:lll
+
+	// BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.
+	// When omitted or null, replicas are unmanaged (preserving HPA or manual kubectl control).
+	// When set (including 0), the value is an explicit replica count passed to consuming code.
+	BackendReplicas *int32 `json:"backend_replicas,omitempty" yaml:"backend_replicas,omitempty"`
+
+	// SessionCacheSize is the maximum number of sessions held in the local LRU cache.
+	// A value of 0 causes consuming code to apply a sensible default (1000).
+	SessionCacheSize int `json:"session_cache_size,omitempty" yaml:"session_cache_size,omitempty"`
```
JAORMX: `SessionCacheSize int` with `omitempty`... this is a bug in the making.\n\nWith `encoding/json`, `omitempty` on a bare `int` omits the zero value from serialized output. So \"user explicitly set 0\" and \"field was not set\" become indistinguishable after a JSON round-trip through `WriteJSON`/`ReadJSON`.\n\nThe doc says \"A value of 0 causes consuming code to apply a sensible default (1000)\". That works for a single in-memory pass, but if the semantics ever need to distinguish \"unset\" from \"explicitly 0\" (e.g. to disable caching), this breaks silently. And since `RunConfig` is an API contract, changing the type later is a breaking change.\n\nThis should be `*int32` like `BackendReplicas`. That also fixes:\n- The type inconsistency (two fields in the same PR, same category, one `*int32` and one bare `int`)\n- Platform-dependent width (`int` is 32 or 64 bit depending on platform, `int32` is explicit for a serialized contract)">

pkg/runner/config.go
```diff
@@ -210,6 +210,15 @@ type RunConfig struct {
 	// When set, the proxy runner will start an embedded auth server that delegates to upstream IDPs.
 	// This is the serializable RunConfig; secrets are referenced by file paths or env var names.
 	EmbeddedAuthServerConfig *authserver.RunConfig `json:"embedded_auth_server_config,omitempty" yaml:"embedded_auth_server_config,omitempty"` //nolint:lll
+
+	// BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.
+	// When omitted or null, replicas are unmanaged (preserving HPA or manual kubectl control).
+	// When set (including 0), the value is an explicit replica count passed to consuming code.
+	BackendReplicas *int32 `json:"backend_replicas,omitempty" yaml:"backend_replicas,omitempty"`
```
JAORMX, line 228: Question: does the proxy runner binary actually _read_ `BackendReplicas` at runtime, or is it only consumed by the operator controller to set StatefulSet replicas?\n\nIf only the controller uses it, it might belong on the CRD spec directly (like `ServiceType` and `SessionAffinity` today) rather than being serialized into the RunConfig ConfigMap. Threading a deployment-orchestration concern through a runtime config struct crosses a boundary.">

yrobla: Good question — the proxyrunner binary **does** read `BackendReplicas` at runtime, by design in this epic.

The key distinction is ownership: the **operator** creates the proxyrunner **Deployment** directly in `mcpserver_controller.go` (that path never goes through `RunConfig`). But the **proxyrunner itself** creates the backend **StatefulSet** by calling `runtime.Setup()` → `client.DeployWorkload()` on startup, reading its configuration from `RunConfig`.

TASK-002 ([#4209](https://github.com/stacklok/toolhive/issues/4209)) explicitly specifies this chain:
> `RunConfig → runtime.Setup → DeployWorkloadOptions → client.DeployWorkload → StatefulSetSpec`

So `BackendReplicas` controls something the proxyrunner binary itself deploys — it's not just a deployment-orchestration concern owned by the operator.

Your comparison to `ServiceType`/`SessionAffinity` is a fair one architecturally, but those fields configure Kubernetes objects the **operator owns**. The backend StatefulSet is different: the proxyrunner is its deployer, not the operator. If ownership ever moves to the operator (like `EmbeddingServer` does with its StatefulSet), that would be the natural time to revisit where this field lives. The overall architecture tracking is in the epic [stacklok-epics#263](https://github.com/stacklok/stacklok-epics/issues/263).

pkg/runner/config.go
```diff
@@ -210,6 +210,26 @@ type RunConfig struct {
...
+// RunConfig shape.
+type ScalingConfig struct {
+	// BackendReplicas is the desired StatefulSet replica count for the proxy runner backend.
+	// When nil, replicas are unmanaged (preserving HPA or manual kubectl control).
+	// When set (including 0), the value is an explicit replica count.
+	BackendReplicas *int32 `json:"backend_replicas,omitempty" yaml:"backend_replicas,omitempty"`
+
+	// SessionCacheSize is the maximum number of sessions held in the local LRU cache.
+	// When nil, consuming code applies a sensible default (e.g. 1000).
+	SessionCacheSize *int32 `json:"session_cache_size,omitempty" yaml:"session_cache_size,omitempty"`
```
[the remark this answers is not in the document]
yrobla: This was changed to `*int32` per JAORMX's explicit request in the earlier inline comment on the same file — see the thread at line 221. The original issue spec used `int`, but JAORMX correctly identified that `int` with `omitempty` silently drops the zero value on JSON round-trip, making 'unset' and 'explicitly 0' indistinguishable after serialization. Since `RunConfig` is an API contract and we can't change the type later without a breaking change, `*int32` is the right call here. The issue's acceptance criteria were superseded by the review.

pkg/runner/config.go
```diff
@@ -210,6 +210,26 @@ type RunConfig struct {
 	// When set, the proxy runner will start an embedded auth server that delegates to upstream IDPs.
 	// This is the serializable RunConfig; secrets are referenced by file paths or env var names.
 	EmbeddedAuthServerConfig *authserver.RunConfig `json:"embedded_auth_server_config,omitempty" yaml:"embedded_auth_server_config,omitempty"` //nolint:lll
+
+	// ScalingConfig contains configuration for horizontal scaling of the proxy runner.
+	// Only applicable when running in Kubernetes with the ToolHive operator.
+	// When nil, no scaling configuration is applied (single-replica default behaviour).
```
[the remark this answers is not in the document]
yrobla: Fixed.

pkg/runner/config.go
```diff
@@ -210,6 +210,26 @@ type RunConfig struct {
...
+	ScalingConfig *ScalingConfig `json:"scaling_config,omitempty" yaml:"scaling_config,omitempty"`
```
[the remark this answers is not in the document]
yrobla: Added both: (1) `assert.Nil(t, got.ScalingConfig.SessionCacheSize)` to the `backend_replicas 2` test to assert the field remains nil when omitted, and (2) a new `session_cache_size 0 in JSON deserializes to pointer-to-zero, not nil` subtest that mirrors the existing `backend_replicas: 0` test and also asserts `BackendReplicas` is nil when absent. All 7 subtests pass.

PR #4241, opened 2026-03-19
feat(proxyrunner): wire SIGTERM-aware graceful shutdown (RC-14)

### LINKED ISSUE #4207: Wire SIGTERM-aware graceful shutdown in thv-proxyrunner entrypoint
## Description

Wire `signal.NotifyContext` for `SIGTERM`, `SIGQUIT`, and `os.Interrupt` in `cmd/thv-proxyrunner/main.go` and propagate the resulting context through `cobra.ExecuteContext` so that Kubernetes pod termination triggers graceful connection drain rather than abrupt process exit. Currently, `main.go` calls `app.NewRootCmd().Execute()` with no signal-aware context, meaning SIGTERM from the pod lifecycle will terminate the process immediately without giving in-flight MCP proxy connections a chance to drain.

## Context

This task implements RC-14 from RFC THV-0047 (Manual Horizontal Scaling for vMCP and Proxy Runner). When the proxyrunner is deployed as multiple StatefulSet replicas, Kubernetes sends SIGTERM to pods during rolling updates, scale-down events, and node drains. Without a signal-aware context, in-flight MCP sessions are dropped at the transport layer without going through the graceful drain path in `transportHandler.Stop`. The fix is a small, self-contained change to `main.go` that mirrors the existing pattern in `cmd/vmcp/main.go`. The operator is responsible for setting an appropriate `terminationGracePeriodSeconds`; the proxyrunner simply stops accepting new connections and waits for `transportHandler.Stop` to complete.

**Dependencies**: None
**Blocks**: None (TASK-003 is independent of all other tasks in this epic)

## Acceptance Criteria

- [ ] `cmd/thv-proxyrunner/main.go` creates a `signal.NotifyContext` cancellable on `os.Interrupt`, `syscall.SIGTERM`, and `syscall.SIGQUIT`
- [ ] `defer cancel()` is called immediately after `signal.NotifyContext` to ensure the OS signal listener is released on exit
- [ ] `app.NewRootCmd().ExecuteContext(ctx)` is called instead of `app.NewRootCmd().Execute()`, passing the signal-aware context
- [ ] `runCmdFunc` in `cmd/thv-proxyrunner/app/run.go` already reads `cmd.Context()` — no changes needed there; the context flows through without modification
- [ ] On SIGTERM, the proxyrunner stops accepting new requests and waits for `transportHandler.Stop` to complete before exiting (exercised by the existing context-cancellation path in `runner.Run`)
- [ ] Error handling on `ExecuteContext` matches the vmcp pattern: log the error via `slog.Error` and call `os.Exit(1)`
- [ ] No signal channels or additional goroutines are created in `main.go` — `signal.NotifyContext` is the single mechanism
- [ ] Existing single-replica proxyrunner behaviour is unchanged (context cancellation on CTRL-C already works; this extends the same mechanism to SIGTERM/SIGQUIT)
- [ ] All existing tests pass

Replace the current `app.NewRootCmd().Execute()` call in `cmd/thv-proxyrunner/main.go` with the `signal.NotifyContext` + `ExecuteContext` pattern taken verbatim from `cmd/vmcp/main.go`. The context flows through the existing chain: `ExecuteContext` → `runCmdFunc` (which reads `cmd.Context()`) → `mcpRunner.Run(ctx)` → the `select { case <-ctx.Done() }` block that calls `stopMCPServer`, which calls `transportHandler.Stop`. No changes to the `app` package are required.

### Patterns & Frameworks

- `signal.NotifyContext` from `os/signal` (stdlib) — creates a context that is cancelled when the specified OS signals arrive; this is the idiomatic Go pattern for signal-driven graceful shutdown since Go 1.16
- `syscall.SIGTERM` and `syscall.SIGQUIT` from `syscall` (stdlib) — standard Kubernetes termination signals; SIGQUIT additionally triggers a goroutine dump useful for debugging stuck drain scenarios
- `cobra.Command.ExecuteContext` — passes the context to all subcommand `RunE` functions via `cmd.Context()`; `runCmdFunc` already calls `cmd.Context()` on line 77 of `run.go`, so no changes to the command handler are needed
- Operator controls drain duration via `terminationGracePeriodSeconds` on the proxyrunner Deployment/Pod spec — the proxyrunner itself does not set an internal exit deadline

### Code Pointers

- `cmd/thv-proxyrunner/main.go` — the only file that needs to change; replace `Execute()` with `signal.NotifyContext` + `ExecuteContext`; add imports for `context`, `fmt`, `os/signal`, and `syscall`
- `cmd/vmcp/main.go` — reference implementation to follow exactly; lines 31–38 show the complete pattern
- `cmd/thv-proxyrunner/app/run.go:77` — `ctx := cmd.Context()` already reads the context from Cobra; no change needed here
- `pkg/runner/runner.go:529–531` — `select { case <-ctx.Done(): stopMCPServer("Context cancelled") }` is the existing drain path that triggers `transportHandler.Stop`; it fires automatically once the signal-aware context is cancelled

### Component Interfaces

The change is entirely in `cmd/thv-proxyrunner/main.go`. The final `main` function should match this structure:

```go
func main() {
	// Bind TOOLHIVE_DEBUG env var early, before logger initialization.
	if err := viper.BindEnv("debug", "TOOLHIVE_DEBUG"); err != nil {
		slog.Error("failed to bind TOOLHIVE_DEBUG env var", "error", err)
	}

	// Initialize the logger
	var opts []logging.Option
	if viper.GetBool("debug") {
		opts = append(opts, logging.WithLevel(slog.LevelDebug))
	}
	l := logging.New(opts...)
	slog.SetDefault(l)

	// Create a context that will be canceled on signal
	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM, syscall.SIGQUIT)
	defer cancel()

	// Execute the root command with context
	if err := app.NewRootCmd().ExecuteContext(ctx); err != nil {
		slog.Error(fmt.Sprintf("Error executing command: %v", err))
		os.Exit(1)
	}
}
```

New imports required: `"context"`, `"fmt"`, `"os/signal"`, `"syscall"` (in addition to the existing imports already present in the file).

### Testing Strategy

**Unit Tests**
- [ ] No unit tests are required for `main.go` directly (it is an entrypoint with no testable logic beyond the signal wiring); the existing runner tests cover the drain path
- [ ] Verify that `runCmdFunc` receives a non-nil context via `cmd.Context()` — this is already exercised by existing runner unit tests in `pkg/runner/runner_test.go`

**Integration Tests**
- [ ] Manual / e2e: send SIGTERM to a running proxyrunner pod; verify the process does not exit until in-flight proxy connections close or the `terminationGracePeriodSeconds` elapses
- [ ] Manual: send `os.Interrupt` (CTRL-C) to a locally running proxyrunner; verify clean exit (existing behaviour should be preserved)

**Edge Cases**
- [ ] Double-signal: sending a second SIGTERM after the first should be handled by `signal.NotifyContext` natively (the context is already cancelled; no additional action needed)
- [ ] Signal received before `transportHandler.Start`: `ctx.Done()` fires before the transport is running; `stopMCPServer` is never called because the `select` is reached only after `Start` returns — verify the process exits cleanly in this path

## Out of Scope

- Operator-side `terminationGracePeriodSeconds` configuration — this is an operator concern, not a proxyrunner code change
- vMCP-side graceful shutdown — covered by the vMCP epic
- Any changes to `cmd/thv-proxyrunner/app/run.go`, `cmd/thv-proxyrunner/app/commands.go`, or `pkg/runner/runner.go` — the context chain and drain logic already exist
- Custom exit-deadline timeout inside the proxyrunner process — the operator controls the drain window via `terminationGracePeriodSeconds`

## References

- RFC THV-0047: https://github.com/stacklok/toolhive-rfcs/pull/47 — source RFC; RC-14 defines the SIGTERM drain requirement
- Parent epic: https://github.com/stacklok/stacklok-epics/issues/263
- Reference implementation: `cmd/vmcp/main.go` in `stacklok/toolhive` — identical signal-context pattern to apply
- `signal.NotifyContext` Go documentation: https://pkg.go.dev/os/signal#NotifyContext

yrobla:
## Summary

Replace Execute() with signal.NotifyContext + ExecuteContext() in cmd/thv-proxyrunner/main.go, mirroring cmd/vmcp/main.go. SIGTERM, SIGQUIT, and os.Interrupt now cancel the root context so in-flight MCP proxy connections drain via transportHandler.Stop instead of being dropped abruptly during Kubernetes pod termination.

Fixes #4207 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
|      |        |

### DIFF
--- a/test/e2e/proxyrunner_graceful_shutdown_test.go
+++ b/test/e2e/proxyrunner_graceful_shutdown_test.go
@@ -1,0 +1,146 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"net"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"syscall"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/stacklok/toolhive/pkg/runner"
+	"github.com/stacklok/toolhive/test/e2e"
+)
+
+var _ = Describe("ProxyRunner Graceful Shutdown", Label("proxyrunner", "graceful-shutdown", "e2e"), Serial, func() {
+	var (
+		config         *e2e.TestConfig
+		serverName     string
+		tempDir        string
+		proxyRunnerCmd *exec.Cmd
+		exportedConfig *runner.RunConfig
+	)
+
+	BeforeEach(func() {
+		config = e2e.NewTestConfig()
+
+		err := e2e.CheckTHVBinaryAvailable(config)
+		Expect(err).ToNot(HaveOccurred(), "thv binary should be available for testing")
+
+		_, err = exec.LookPath(proxyRunnerBinaryPath())
+		Expect(err).ToNot(HaveOccurred(),
+			"thv-proxyrunner binary not found; set THV_PROXYRUNNER_BINARY or add it to PATH")
+
+		serverName = fmt.Sprintf("proxyrunner-shutdown-%d", GinkgoRandomSeed())
+		tempDir = GinkgoT().TempDir()
+
+		By("Starting an OSV MCP server via thv to obtain a valid runconfig.json")
+		e2e.NewTHVCommand(config, "run", "--name", serverName, "osv").ExpectSuccess()
+
+		err = e2e.WaitForMCPServer(config, serverName, 60*time.Second)
+		Expect(err).ToNot(HaveOccurred(), "OSV server should start within 60s")
+
+		By("Exporting the run configuration to tempDir/runconfig.json")
+		configPath := filepath.Join(tempDir, "runconfig.json")
+		e2e.NewTHVCommand(config, "export", serverName, configPath).ExpectSuccess()
+
+		configData, err := os.ReadFile(configPath)
+		Expect(err).ToNot(HaveOccurred(), "exported runconfig.json should be readable")
+
+		exportedConfig = &runner.RunConfig{}
+		Expect(json.Unmarshal(configData, exportedConfig)).To(Succeed(), "exported config should be valid JSON")
+		Expect(exportedConfig.Image).ToNot(BeEmpty(), "exported config should have a non-empty image")
+		Expect(exportedConfig.Port).To(BeNumerically(">", 0), "exported config should have a valid port")
+
+		By("Stopping the thv-managed server to free the container name and port")
+		Expect(e2e.StopAndRemoveMCPServer(config, serverName)).To(Succeed())
+
+		By("Waiting for the port to be released")
+		Eventually(func() bool {
+			conn, err := net.DialTimeout("tcp", fmt.Sprintf("localhost:%d", exportedConfig.Port), 1*time.Second)
+			if err != nil {
+				return true // port is free
+			}
+			conn.Close()
+			return false
+		}, 15*time.Second, 500*time.Millisecond).Should(BeTrue(), "port should be released after server removal")
+	})
+
+	AfterEach(func() {
+		if proxyRunnerCmd != nil && proxyRunnerCmd.Process != nil {
+			_ = proxyRunnerCmd.Process.Kill()
+			_, _ = proxyRunnerCmd.Process.Wait()
+			proxyRunnerCmd = nil
+		}
+		// Best-effort: remove any container left behind by the proxyrunner
+		_ = e2e.StopAndRemoveMCPServer(config, serverName)
+	})
+
+	It("exits cleanly when SIGTERM is received after the proxy is ready", func() {
+		By("Starting thv-proxyrunner with the exported runconfig.json")
+		proxyRunnerCmd = exec.Command( //nolint:gosec // Intentional for e2e testing
+			proxyRunnerBinaryPath(), "run", exportedConfig.Image,
+		)
+		proxyRunnerCmd.Dir = tempDir // runconfig.json is picked up from the working directory
+		proxyRunnerCmd.Stdout = GinkgoWriter
+		proxyRunnerCmd.Stderr = GinkgoWriter
+		Expect(proxyRunnerCmd.Start()).To(Succeed(), "thv-proxyrunner should start")
+
+		proxyAddr := fmt.Sprintf("localhost:%d", exportedConfig.Port)
+
+		By(fmt.Sprintf("Waiting for the proxy to accept connections on %s", proxyAddr))
+		Eventually(func() error {
+			conn, err := net.DialTimeout("tcp", proxyAddr, 1*time.Second)
+			if err != nil {
+				return err
+			}
+			conn.Close()
+			return nil
+		}, 90*time.Second, 2*time.Second).Should(Succeed(), "proxy port should become reachable within 90s")
+
+		By("Sending SIGTERM to thv-proxyrunner (simulating Kubernetes pod termination)")
+		Expect(proxyRunnerCmd.Process.Signal(syscall.SIGTERM)).To(Succeed())
+
+		By("Asserting thv-proxyrunner exits within 30 seconds of SIGTERM")
+		done := make(chan error, 1)
+		go func() { done <- proxyRunnerCmd.Wait() }()
+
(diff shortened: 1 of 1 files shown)


cmd/thv-proxyrunner/main.go
```diff
@@ -30,8 +33,14 @@ func main() {
...
-	// Skip update check for completion command or if we are running in kubernetes
-	if err := app.NewRootCmd().Execute(); err != nil {
+	// Create a signal-aware context so SIGTERM from Kubernetes pod lifecycle,
+	// SIGQUIT, and os.Interrupt all trigger graceful connection drain via
+	// transportHandler.Stop rather than abrupt process exit.
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM, syscall.SIGQUIT)
+	defer cancel()
+
+	if err := app.NewRootCmd().ExecuteContext(ctx); err != nil {
+		slog.Error("error executing command", "error", err)
```
jerm-dro, line 43: Blocker: can we integration test this?

I ask because I made a similar change to `thv` a while back and it was quite painful, because the underlying code had not actually been designed for graceful shutdown.

jerm-dro (dismissed):

cmd/thv-proxyrunner/main.go
yrobla: Yes — and there's already a pattern for it in the e2e suite we can follow.

**Is the underlying code actually designed for graceful shutdown?**

Yes, unlike the previous `thv` change. `runner.Run()` already selects on `ctx.Done()` and calls `stopMCPServer()` → `transportHandler.Stop(cleanupCtx)` ([runner.go:529-531](https://github.com/stacklok/toolhive/blob/main/pkg/runner/runner.go#L529)). That cleanup path was already there — this change just ensures OS signals actually cancel the context rather than killing the process abruptly, which would bypass `transportHandler.Stop()` entirely.

**How to integration test it**

`proxy_tunnel_e2e_test.go` already does exactly this pattern:
```go
_ = proxyTunnelCmd.Process.Signal(os.Interrupt)
done := make(chan error, 1)
go func() { done <- proxyTunnelCmd.Wait() }()
select {
case <-done:
case <-time.After(10 * time.Second):
    Fail("proxy did not exit after SIGINT within 10s")
}
```
([proxy_tunnel_e2e_test.go:158-167](https://github.com/stacklok/toolhive/blob/main/test/e2e/proxy_tunnel_e2e_test.go#L158))

I'll add an e2e test for `thv-proxyrunner` that starts the binary, sends SIGTERM, and asserts it exits within a timeout with the container cleaned up — same structure as the tunnel test.

jerm-dro, line 43: Does the proxyrunner's `run` command actually use `cmd.Context()` and propagate cancellation to the HTTP server / transport handler? If not, the signal cancels the context but nothing drains gracefully.

yrobla: Yes, cancellation is fully propagated through the call chain:

1. `runCmdFunc` calls `cmd.Context()` ([run.go:77](https://github.com/stacklok/toolhive/blob/main/cmd/thv-proxyrunner/app/run.go#L77)) to get the signal-aware context
2. That context flows into `mcpRunner.Run(ctx)` ([run.go:211](https://github.com/stacklok/toolhive/blob/main/cmd/thv-proxyrunner/app/run.go#L211))
3. Inside `Run`, it's passed to `transportHandler.Start(ctx)` ([runner.go:388](https://github.com/stacklok/toolhive/blob/main/pkg/runner/runner.go#L388))
4. The main select loop explicitly handles `<-ctx.Done()` and calls `stopMCPServer("Context cancelled")` ([runner.go:529-531](https://github.com/stacklok/toolhive/blob/main/pkg/runner/runner.go#L529))
5. `stopMCPServer` calls `transportHandler.Stop()` with a fresh 1-minute timeout context for graceful drain ([runner.go:446](https://github.com/stacklok/toolhive/blob/main/pkg/runner/runner.go#L446))

So SIGTERM → context cancelled → `<-ctx.Done()` unblocks → `transportHandler.Stop()` drains connections gracefully.

PR #4188, opened 2026-03-17
Remove non-spec progress/update method from MCP parser

majiayu000:
## Summary

The MCP parser included a `progress/update` method handler that is not part of the MCP specification. This conflicts with the spec-compliant `notifications/progress` handler that already exists and works correctly. Removing the non-spec method to align with the protocol specification.

- Remove `progress/update` map entry and dead `handleProgressMethod` function from parser
- Remove two related test cases from parser tests
- Remove `progress/update` from authorization middleware (method map and skip-auth condition)

Fixes #4170

## Type of change

- [x] Bug fix

## Test plan

- [x] Unit tests (`task test`)
- [x] Linting (`task lint-fix`)

Ran `go test ./pkg/mcp/... ./pkg/authz/...` — all tests pass. Verified `go build ./...` compiles cleanly.

PR #4242, opened 2026-03-19
Add factory helpers for user and scoped secret providers

amirejaz:
## Summary

- Callers currently have to manually wrap the result of `CreateSecretProvider` in `UserProvider` or `ScopedProvider`; this PR adds dedicated constructors so the wrapping is consistent and boilerplate-free.
- Adds `CreateUserSecretProvider(providerType)` — creates the base provider and wraps it in `UserProvider`, blocking all `__thv_*` keys. Intended for user-facing callers (CLI, API, MCP tool server).
- Adds `CreateScopedSecretProvider(providerType, scope)` — creates the base provider and wraps it in `ScopedProvider`, namespacing all key operations under `__thv_<scope>_`. Intended for internal callers such as the registry or workloads subsystem.

This is **Phase 2** of the scoped secret store implementation (part of #4188). Phase 1 (#4229) introduced `ScopedProvider` and `UserProvider`; this PR exposes them through the factory layer.

## Type of change

## Test plan

- [x] Unit tests added to `pkg/secrets/factory_test.go` covering both helpers for the environment provider type (the only provider that works without external dependencies in CI), system-key blocking, and the unknown-provider error path.
- [x] `go test ./pkg/secrets/...` — all tests pass.
- [x] `golangci-lint run ./pkg/secrets/...` — 0 issues.

This PR is stacked on top of `scoped-secret-providers` (Phase 1 #4229). The diff against that branch is a single commit touching only `factory.go` and `factory_test.go`.

### DIFF
--- a/pkg/secrets/factory.go
+++ b/pkg/secrets/factory.go
@@ -271,24 +271,67 @@ func CreateSecretProviderWithPassword(managerType ProviderType, password string)
 	return primary, nil
 }
 
-// CreateUserSecretProvider creates a Provider that filters out system-reserved keys,
-// suitable for user-facing callers (CLI, API, MCP tool server).
-func CreateUserSecretProvider(managerType ProviderType) (Provider, error) {
-	inner, err := CreateSecretProvider(managerType)
-	if err != nil {
-		return nil, err
-	}
-	return NewUserProvider(inner), nil
+// ProviderOption configures how CreateProvider wraps the underlying provider.
+type ProviderOption func(*providerOptions) error
+
+// providerOptions holds the accumulated state of all ProviderOptions passed to CreateProvider.
+type providerOptions struct {
+	scope      *SecretScope // nil = no scoping
+	userFacing bool
 }
 
-// CreateScopedSecretProvider creates a Provider that namespaces all operations
-// under the given scope, suitable for internal callers.
-func CreateScopedSecretProvider(managerType ProviderType, scope SecretScope) (Provider, error) {
+// WithScope returns a ProviderOption that wraps the provider with a ScopedProvider
+// for the given scope, keeping system secrets isolated under "__thv_<scope>_".
+// Mutually exclusive with WithUserFacing.
+func WithScope(scope SecretScope) ProviderOption {
+	return func(o *providerOptions) error {
+		if o.userFacing {
+			return errors.New("WithScope and WithUserFacing are mutually exclusive")
+		}
+		o.scope = &scope
+		return nil
+	}
+}
+
+// WithUserFacing returns a ProviderOption that wraps the provider with a UserProvider,
+// blocking access to any key that starts with the system prefix "__thv_".
+// Suitable for CLI, API, and MCP tool server callers.
+// Mutually exclusive with WithScope.
+func WithUserFacing() ProviderOption {
+	return func(o *providerOptions) error {
+		if o.scope != nil {
+			return errors.New("WithUserFacing and WithScope are mutually exclusive")
+		}
+		o.userFacing = true
+		return nil
+	}
+}
+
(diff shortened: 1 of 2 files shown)


pkg/secrets/factory_test.go
```diff
@@ -112,3 +112,71 @@ func TestEnvVarPrefix(t *testing.T) { //nolint:paralleltest
 		assert.Equal(t, "TOOLHIVE_SECRET_", secrets.EnvVarPrefix)
 	})
 }
+
+func TestCreateUserSecretProvider(t *testing.T) { //nolint:paralleltest
+	ctx := context.Background()
```
JAORMX: why not use `t.Context()`?

pkg/secrets/factory_test.go
```diff
@@ -112,3 +112,71 @@ func TestEnvVarPrefix(t *testing.T) { //nolint:paralleltest
...
+
+	t.Run("unknown provider returns error", func(t *testing.T) { //nolint:paralleltest
+		provider, err := secrets.CreateUserSecretProvider(secrets.ProviderType("unknown"))
+		assert.Error(t, err)
+		assert.Nil(t, provider)
+	})
+}
+
+func TestCreateScopedSecretProvider(t *testing.T) { //nolint:paralleltest
+	ctx := context.Background()
```
JAORMX: `t.Context()` would be better.

pkg/secrets/factory.go
```diff
@@ -271,6 +271,26 @@ func CreateSecretProviderWithPassword(managerType ProviderType, password string)
 	return primary, nil
 }
 
+// CreateUserSecretProvider creates a Provider that filters out system-reserved keys,
+// suitable for user-facing callers (CLI, API, MCP tool server).
+func CreateUserSecretProvider(managerType ProviderType) (Provider, error) {
```
reyortiz3: would suggest using the functional options pattern instead of adding individual `Create*` helper methods

amirejaz: Thanks for the suggestion! After thinking it through, I'd like to keep the two named functions instead. Here's the reasoning:

1. **The scope argument is required, not optional** — `CreateScopedSecretProvider` enforces at compile time that a scope must be provided. With `CreateProvider(managerType, WithScope(...))`, calling `CreateProvider(managerType)` with no options is valid Go but silently returns a raw unwrapped provider, which is likely a silent bug at the call site.

2. **The two modes are mutually exclusive** — you'd never combine `WithUserFilter()` and `WithScope()` on the same provider. Functional options pay off when configurations genuinely compose; here they'd just add indirection.

3. **Clarity at call sites** — `CreateScopedSecretProvider(providerType, ScopeWorkloads)` is self-documenting without needing to know the options API exists. Happy to revisit if we end up needing more wrapping modes in the future.

reyortiz3: I see the intent. But I think having the functional options with a kind off a builder pattern is more extensible than creating different Factory methods for each combination (it seems we already have a similar pattern e.g. `config_builder.go`):

1. The same argument could be made if by mistake it calls `CreateUserProvier(managerType)`, is valid Go and silently returns a raw unwrapped provider and effectively a bug, only a unit test would ensure this

2. I've seen validating mutually exclusive options when building the object, e.g. https://github.com/grpc/grpc-go/blob/master/clientconn.go#L484, and https://github.com/temporalio/sdk-go/blob/master/internal/client.go#L1255

3. I would think they're equally clear:  `CreateScopedSecretProvider(providerType, ScopeWorkloads)` and `CreateProvider(managerType, WithScope(...))`

amirejaz: Addressed. Mind taking another look?

reyortiz3:
lgtm

PR #4280, opened 2026-03-19
Fix composite tool type coercion for non-advertised tools

jerm-dro:
## Summary

- Composite tool workflows using backend tools excluded from advertising (via `excludeAll` or `filter`) fail type coercion because the workflow engine only searches the advertised tools list for InputSchema. Template expansion always produces strings, and without schema-based coercion backends reject mistyped arguments (e.g. `pullNumber` sent as `"42"` instead of `float64(42)`).
- Add `AllRoutableTools()` to the `MultiSession` interface, returning all tools in the routing table including non-advertised ones with their schemas. The server now passes this to the composer factory so the workflow engine can coerce arguments for any routable tool.

## Type of change

- [x] Bug fix

## Test plan

Ran `go test ./pkg/vmcp/...` — all existing and new tests pass. New tests:
- `TestWorkflowEngine_DotConvention_CoercesTypesWithSessionRouter` — verifies dot-convention tool name resolution + coercion with a real session router
- `TestWorkflowEngine_DotConvention_ExcludedToolCoercesWithAllRoutableTools` — reproduces the exact bug (non-advertised tool) and verifies the fix

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/session/types/session.go` | Add `AllRoutableTools()` to `MultiSession` interface |
| `pkg/vmcp/session/default_session.go` | Implement `AllRoutableTools()` with new `allRoutableTools` field |
| `pkg/vmcp/session/factory.go` | Build `allRoutableTools` from routing table entries; return from `buildRoutingTableWithAggregator` |
| `pkg/vmcp/session/compositetools/decorator.go` | Delegate `AllRoutableTools()` to underlying session |
| `pkg/vmcp/session/optimizerdec/decorator.go` | Delegate `AllRoutableTools()` to underlying session |
| `pkg/vmcp/server/server.go` | Pass `AllRoutableTools()` instead of `Tools()` to composer factory |
| `pkg/vmcp/session/types/mocks/mock_session.go` | Add mock for `AllRoutableTools()` |
| `pkg/vmcp/composer/workflow_engine_test.go` | Two new reproduction/verification tests |
| `pkg/vmcp/server/telemetry_integration_test.go` | Add `AllRoutableTools()` stub |
| `pkg/vmcp/server/session_management_integration_test.go` | Add `AllRoutableTools()` mock expectation |

The root cause: `getToolInputSchema()` in `workflow_engine.go` searches `e.tools` for a tool's `InputSchema` to drive type coercion after template expansion. When aggregator advertising filters (`excludeAll` / `filter`) exclude a tool from `Tools()`, the schema lookup returns nil → passthrough schema → no coercion → backend rejects string-typed arguments.

The fix is intentionally minimal — `AllRoutableTools()` provides schema access for all routing-table tools without changing the advertised tools list or the workflow engine itself.


jerm-dro:
Don't like this implementation. Drafting a better version.

pkg/vmcp/session/factory.go
```diff
@@ -300,7 +312,51 @@ func buildRoutingTableWithAggregator(
...
+	return rt, advertisedTools, allRoutableTools, allResources, allPrompts, nil
+}
+
+// buildAllRoutableTools constructs a tool list for every entry in the routing
+// table by looking up the original backend tool schema. The returned slice uses
+// the resolved (routing-table) name as the tool Name, preserving the InputSchema
+// from the backend so that workflow type coercion works for non-advertised tools.
+func buildAllRoutableTools(
+	toolsRouting map[string]*vmcp.BackendTarget,
+	toolsByBackend map[string][]vmcp.Tool,
```
jhrozek, line 324: `targets` doesn't seem to be used anywhere in this function — the loop already gets targets from `toolsRouting`. Can we drop this parameter?

pkg/vmcp/session/factory.go
```diff
@@ -300,7 +312,51 @@ func buildRoutingTableWithAggregator(
 		}
```
jhrozek, line 312: This function has some interesting logic (the index build, `GetBackendCapabilityName` reverse lookup, sorting) but no direct unit tests. Would be good to have a few table-driven cases — at minimum empty input, a routing entry with no matching backend tool, and multiple backends.

pkg/vmcp/server/session_management_integration_test.go
```diff
@@ -112,6 +112,9 @@ func newMockFactory(t *testing.T, ctrl *gomock.Controller, tools []vmcp.Tool) (*
 			toolsCopy := make([]vmcp.Tool, len(tools))
 			copy(toolsCopy, tools)
 			mock.EXPECT().Tools().Return(toolsCopy).AnyTimes()
+			allRoutableCopy := make([]vmcp.Tool, len(tools))
+			copy(allRoutableCopy, tools)
```
jhrozek, line 116: The mock returns the same list for both `Tools()` and `AllRoutableTools()` here. Was that intentional, or should we have a test case where they diverge (the scenario this PR is fixing)?

pkg/vmcp/session/factory.go
```diff
@@ -259,11 +259,17 @@ func buildRoutingTable(results []initResult) (*vmcp.RoutingTable, []vmcp.Tool, [
 // pipeline (overrides, conflict resolution, advertising filter) to the raw
 // backend capabilities in results, producing resolved tool names identical to
 // the standard aggregation path. Resources and prompts pass through unchanged.
+//
+// Returns:
+//   - rt: the routing table
+//   - advertisedTools: tools passing the advertising filter (for tools/list)
+//   - allRoutableTools: ALL tools in the routing table with schemas (for workflow coercion)
+//   - allResources, allPrompts: resources and prompts
```
jhrozek, line 267: This is up to 6 return values now, two of which are `[]vmcp.Tool` with different semantics. Might be worth bundling into a result struct at some point — makes it harder to accidentally swap the two tool slices at the call site.

PR #4281, opened 2026-03-19
Retry transient network errors in background token monitor

reyortiz3:
## Summary

- OAuth token refresh fails when WARP VPN disconnects overnight — the refresh endpoint becomes temporarily unreachable from the home ISP IP, causing the workload to be immediately marked as unauthenticated and requiring manual intervention.
- Adds `isTransientNetworkError` to distinguish DNS failures, TCP-level errors, and timeouts (transient) from OAuth 4xx / `invalid_grant` responses (non-transient). The background monitor now retries transient failures with exponential backoff (10 s initial → 2 min max, bounded by 5 attempts and 5 min elapsed time). Context cancellation exits the loop cleanly without marking the workload as unauthenticated. Non-transient errors still fail immediately as before.

## Type of change

- [x] Bug fix

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/auth/monitored_token_source.go` | Add `isTransientNetworkError`, exponential backoff retry loop in `Token`, injectable `newRetryBackOff` for tests |
| `pkg/auth/monitored_token_source_test.go` | Tests for `isTransientNetworkError` and the three retry scenarios: recovery, context cancellation, transient→non-transient escalation |
| `go.mod` / `go.sum` | Add `github.com/cenkalti/backoff/v5` (direct dep), bump otel exporters and grpc-gateway (incidental `go mod tidy` updates) |

No — the workload stays authenticated through VPN reconnects instead of going unauthenticated overnight. No CLI or config changes.

- Retries are bounded by both attempt count (`WithMaxTries`, default 5) and wall-clock time (`WithMaxElapsedTime`, default 5 min). Both limits are configurable via environment variables (`TOOLHIVE_TOKEN_REFRESH_MAX_TRIES`, `TOOLHIVE_TOKEN_REFRESH_MAX_ELAPSED_TIME`).
- The `newRetryBackOff` field is a test seam; it is `nil` in production and replaced with a fast backoff in tests to avoid multi-second sleeps.
- The code scanning alert on `resolveTokenRefreshMaxTries` is a false positive — `ParseUint` with `strconv.IntSize` already constrains the value to fit in `uint`.

### DIFF
--- a/pkg/auth/monitored_token_source.go
+++ b/pkg/auth/monitored_token_source.go
@@ -291,7 +291,8 @@ func (mts *MonitoredTokenSource) resetTimer(d time.Duration) {
 
 // getRetryBackOff returns the backoff to use for transient-error retries.
 // Uses mts.newRetryBackOff if set (e.g. in tests); otherwise returns a default
-// exponential backoff with no MaxElapsedTime (context cancellation is the stop signal).
+// exponential backoff. The caller (retryTransientRefresh) applies WithMaxTries
+// and WithMaxElapsedTime to bound the overall retry loop.
 func (mts *MonitoredTokenSource) getRetryBackOff() backoff.BackOff {
 	if mts.newRetryBackOff != nil {
 		return mts.newRetryBackOff()
@@ -299,7 +300,6 @@ func (mts *MonitoredTokenSource) getRetryBackOff() backoff.BackOff {
 	eb := backoff.NewExponentialBackOff()
 	eb.InitialInterval = resolveTokenRefreshInitialRetryInterval()
 	eb.MaxInterval = resolveTokenRefreshMaxRetryInterval()
-	// No MaxElapsedTime — context cancellation is the stop signal.
 	eb.Reset()
 	return eb
 }


pkg/auth/monitored_token_source.go
```diff
@@ -5,15 +5,27 @@ package auth
...
 
 	"github.com/stacklok/toolhive/pkg/container/runtime"
 )
 
+const (
+	// tokenRefreshInitialRetryInterval is the starting interval for exponential
+	// backoff when a token refresh fails during background monitoring.
+	tokenRefreshInitialRetryInterval = 30 * time.Second
+	// tokenRefreshMaxRetryInterval caps the exponential growth of the retry interval.
+	tokenRefreshMaxRetryInterval = 5 * time.Minute
```
reyortiz3: what should the appropriate values be?

pkg/auth/monitored_token_source.go
```diff
@@ -106,28 +184,123 @@ func (mts *MonitoredTokenSource) resetTimer(d time.Duration) {
...
+func isTransientNetworkError(err error) bool {
+	if err == nil ||
+		errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) ||
+		errors.As(err, new(*oauth2.RetrieveError)) {
+		return false
+	}
+
+	// Transient: DNS lookup failure, TCP-level (connection refused, unreachable), or timeout.
+	_, isDNS := errors.AsType[*net.DNSError](err)
+	_, isOp := errors.AsType[*net.OpError](err)
```
jhrozek: net.OpError is the error wrapping a tls.CertificateVerificationError. I don't think this is a huge deal, we wouldn't let the error through just retrying, just might be worth tigtening to only transport-level errors in the future (ECONNREFUSED etc).

Perhaps a follow-up

pkg/auth/monitored_token_source.go
```diff
@@ -106,28 +184,123 @@ func (mts *MonitoredTokenSource) resetTimer(d time.Duration) {
...
+			return true, 0
+		}
+		wait := time.Until(tok.Expiry)
+		if wait < time.Second {
+			wait = time.Second
+		}
+		return false, wait
+	}
+
+	if !isTransientNetworkError(err) {
```
jhrozek: I'm not that familiar with this code, but I wonder why the errors are retried in onTick only and not e.g. in Token() ?

pkg/auth/monitored_token_source.go
```diff
@@ -106,28 +184,123 @@ func (mts *MonitoredTokenSource) resetTimer(d time.Duration) {
...
+	slog.Warn("token refresh failed due to transient network error, retrying with backoff",
+		"workload", mts.workloadName,
+		"error", err,
+	)
+
+	b := mts.getRetryBackOff()
+
+	var refreshed *oauth2.Token
+
+	_, retryErr := backoff.Retry(mts.monitoringCtx, func() (*oauth2.Token, error) {
```
jhrozek: I think you can assign to refresh here dierctly, at least that's how I've been using the library in k8s

pkg/auth/monitored_token_source.go
reyortiz3: good question, by looking at the code and comment on `MonitoredTokenSource`: _"MonitoredTokenSource is a wrapper around an oauth2.TokenSource that monitors authentication failures"_ sounds like it's an appropriate place, also `.Token()` is from the golang oauth2 `TokenSource` interface

pkg/auth/monitored_token_source.go
reyortiz3: agreed on these default values, cc: @jhrozek

pkg/auth/monitored_token_source.go
reyortiz3: updated

pkg/auth/monitored_token_source.go
amirejaz, line 190: Gap: the per-request path has no transient error protection.

This `Token()` method is the per-request path — called when an MCP client request needs a fresh token. It calls `markAsUnauthenticated` immediately on any error, including transient network errors (DNS timeout, connection refused). If a request arrives while the VPN is disconnected and the cached token has expired,  this path will mark the workload unauthenticated instantly, bypassing all the retry logic added to `onTick()`.

reyortiz3: good catch, and related to @jhrozek comment. I see now, `onTick()` had the `.Token()` code duplicated instead of just calling `mts.Token()`, it makes sense to refactor that so it cover the original issue pls this gap

reyortiz3: updated

pkg/auth/monitored_token_source.go
reyortiz3: updated with @amirejaz's comment

pkg/auth/monitored_token_source.go
```diff
@@ -49,17 +120,69 @@ func NewMonitoredTokenSource(
...
+	// Transient network error (e.g. VPN disconnected, DNS unavailable).
+	// Retry with exponential backoff until the network recovers or monitoring stops.
+	slog.Warn("token refresh failed due to transient network error, retrying with backoff",
+		"workload", mts.workloadName,
+		"error", err,
+	)
+
+	b := mts.getRetryBackOff()
+
+	tok, retryErr := backoff.Retry(mts.monitoringCtx, func() (*oauth2.Token, error) {
```
amirejaz, line 215: This retry logic is **fully synchronous and per-request**, which has important implications during transient failures:

- `backoff.Retry(...)` does **not spawn a goroutine**; it blocks the caller.  
- As written, **every incoming request during a transient outage will enter its own retry loop**, each with its own backoff instance.

#### Implications

- Multiple concurrent callers will repeatedly call `mts.tokenSource.Token()` independently.
- There is **no coordination or deduplication** of token refresh attempts.
- When the network recovers, this can lead to a **burst of simultaneous retries** (“thundering herd”) rather than a single shared recovery.
- All callers remain **blocked until retry succeeds, context is canceled, or backoff stops**.

aponcedeleonch: Addressed the thundering herd concern in 170f3271. The retry path in `Token()` is now wrapped in `singleflight.Do`, so concurrent callers share a single retry loop instead of each running their own. When the one active retry resolves, all waiters get the result immediately.

Added a test (`TestMonitoredTokenSource_SingleflightDeduplicatesConcurrentRetries`) that launches 10 concurrent callers during a transient error and asserts the underlying token source is called at most `numCallers + 3` times, not `numCallers * retries`.

JAORMX:
Code looks good! One thing before we merge though...

The "Special notes for reviewers" section says `MaxElapsedTime` is intentionally left at zero with no deadline, and that the retry loop runs until the context is cancelled or the network recovers. But `retryTransientRefresh()` actually passes `backoff.WithMaxTries(5)` and `backoff.WithMaxElapsedTime(5*time.Minute)`, so retries *are* bounded. The comment in `getRetryBackOff()` ("No MaxElapsedTime — context cancellation is the stop signal") is stale for the same reason.

Can you update the PR description and that code comment to match what the code actually does? The bounded retry is the right call, the docs just need to catch up :)

Note that the code scanning alert on `resolveTokenRefreshMaxTries` is a false positive... `ParseUint` with `strconv.IntSize` already constrains the value to fit in `uint`.

PR #4290, opened 2026-03-20
Prevent permanent backend unhealthy marking after startup race

### LINKED ISSUE #4278: vMCP health monitor permanently marks backends unhealthy after startup race
## Summary

When a VirtualMCPServer pod starts at the same time as its backend MCPServer proxies (e.g., during a simultaneous Flux reconciliation), the health monitor's initial health checks fail because the proxies aren't ready yet. Once marked unhealthy, the backends **never recover** despite becoming fully healthy — the vMCP stays in a `Degraded` state indefinitely until the pod is manually restarted.

## Reproduction

1. Deploy a VirtualMCPServer and its backend MCPServers simultaneously (e.g., via a single Flux Kustomization)
2. The vMCP pod starts and runs its initial health checks before the backend proxy pods are ready
3. Initial health checks fail with: `transport error: server returned 4xx for initialize POST, likely a legacy SSE server`
4. Subsequent health checks (every 30s) continue to fail with the same error, even though the backends are now fully healthy
5. The vMCP reports `Degraded` with increasing `consecutiveFailures`

## Observed behavior

- The vMCP `status.phase` stays `Degraded` with `message: "Some backends are unhealthy"`
- `consecutiveFailures` increments every health check interval (30s) and never resets
- Port-forwarding to the backend proxy services and sending the same initialize POST returns HTTP 200 with a valid MCP response
- A second VirtualMCPServer (anonymous auth, already running before the backends were deployed) shows all the same backends as `ready` using the exact same proxy URLs
- Deleting the vMCP pod (forcing a fresh start) immediately resolves the issue — all backends become healthy

## Expected behavior

Each health check cycle should create a fully independent client connection to the backend. If a backend was unavailable at startup but later becomes healthy, the health monitor should detect the recovery and transition the backend from `unhealthy` → `healthy`.

## Environment

- toolhive v0.12.4 (`ghcr.io/stacklok/toolhive/vmcp:v0.12.4`)
- mcp-go v0.45.0
- Kubernetes (EKS), backends managed via Flux CD
- VirtualMCPServer with `manual` conflict resolution and OIDC + Cedar auth

## Analysis

The health check path (`health/monitor.go` → `health/checker.go` → `client/client.go` → `ListCapabilities` → `defaultClientFactory`) appears to create a new `*client.Client` and `http.Client` on each cycle, but all share `http.DefaultTransport` as the base round tripper. The mcp-go `StreamableHTTP` transport returns `ErrLegacySSEServer` for any 4xx during an initialize POST — this may be getting cached or the connection pool in `http.DefaultTransport` may be holding onto stale connection state.

## Secondary issue

A previously deleted `MCPServer` (`admin-tools`) remained in the vMCP's `discoveredBackends` list after the resource was removed from the cluster. The vMCP controller did not reconcile the removal, resulting in perpetual DNS lookup failures for `mcp-admin-tools-proxy`.

yrobla:
## Summary

Health checks could permanently mark backends as unhealthy due to a race condition: `http.DefaultTransport` was shared across all backend clients, causing stale keep-alive connections to replaced K8s pods to persist and return 4xx indefinitely.

- Replace the shared `http.DefaultTransport` reference with `newBackendTransport()`, which clones `DefaultTransport` (or constructs an equivalent) per `defaultClientFactory` call. Each client gets its own isolated connection pool, so a replaced pod's stale connections cannot affect requests to other backends or future calls.

Fixes #4278

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

### DIFF
--- a/pkg/vmcp/client/client.go
+++ b/pkg/vmcp/client/client.go
@@ -82,6 +82,32 @@ func NewHTTPBackendClient(registry vmcpauth.OutgoingAuthRegistry) (vmcp.BackendC
 	return c, nil
 }
 
+// newBackendTransport creates a *http.Transport with the same defaults as http.DefaultTransport.
+// If http.DefaultTransport is a *http.Transport, it is cloned directly (preserving any
+// environment-specific settings like TLS config or proxy overrides). Otherwise a transport
+// with the standard Go defaults is constructed, preserving proxy, dial timeout, HTTP/2, and
+// idle-connection settings that a zero-value &http.Transport{} would drop.
+func newBackendTransport() *http.Transport {
+	if dt, ok := http.DefaultTransport.(*http.Transport); ok {
+		return dt.Clone()
+	}
+	// http.DefaultTransport has been replaced (e.g. in tests or by a third-party library).
+	// Construct a transport with the same defaults as the Go standard library uses for
+	// http.DefaultTransport so we don't silently drop proxy, timeout, or HTTP/2 settings.
+	return &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   30 * time.Second,
+			KeepAlive: 30 * time.Second,
+		}).DialContext,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
+}
+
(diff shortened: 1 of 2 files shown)


yrobla:
Thanks for the detailed review\!

Agreed on both points:

1. **Transport fix** — I'll strip out the cache machinery and use newBackendTransport() directly That's the actual fix and it's sufficient.

2. **Error mapping** — I'll move the mcp-go sentinel handling and the \ pattern additions into a follow-up PR so they can be reviewed independently.

Working on the simplification now.

PR #4292, opened 2026-03-20
Fix missing resource parameter on cached token restore

rodrigo1208:
When restoring from cached tokens, token refreshes were using the plain oauth2.Config.TokenSource, which drops the RFC 8707 resource parameter. Use NewResourceTokenSource (exported) on the cached path, matching the existing fresh OAuth flow behaviour.

## Summary
When a session is restored from cached tokens (i.e. after a process restart), token refreshes were issued without the RFC 8707 resource parameter. This caused the authorization server to return access
  tokens with a missing or incorrect aud claim, breaking authentication on any endpoint that strictly validates it. The fresh OAuth flow was already correct — NewResourceTokenSource was introduced in #3713
  to handle this — but the cached restore path bypassed it entirely, always falling back to the plain oauth2.Config.TokenSource.                                                                               
                                                                                                                                
  - Export newResourceTokenSource → NewResourceTokenSource so it can be used outside the oauth package                                                                                                         
  - Apply the same resource-aware token source on the cached restore path in tryRestoreFromCachedTokens, mirroring the pattern already used in flow.go                                                         
                                                                                                                                                                                                               
  Type of change                                                                                                                                                                                               
                                                                                                                                                                                                               
  - Bug fix                                                 

  Test plan

  - Unit tests (task test)                                                                                                                                                                                     
  - Manual testing (describe below)
                                                                                                                                                                                                               
  Stop + start the server with a cached session configured with a resource parameter. Confirmed the restored token source now sends resource on refresh and the server returns tokens with the correct aud     
  claim.                                                                                                                                                                                                       
                                                                                                                                                                                                               
  Special notes for reviewers                                                                                                                                                                                  
   
  The root cause was subtle: CreateTokenSourceFromCached wraps oauth2.Config.TokenSource, which has no knowledge of RFC 8707. The fix mirrors exactly what flow.go:processToken already does — the only missing
   piece was exposing NewResourceTokenSource for cross-package use.

### DIFF
--- a/pkg/auth/remote/handler.go
+++ b/pkg/auth/remote/handler.go
@@ -11,7 +11,6 @@ import (
 	"golang.org/x/oauth2"
 
 	"github.com/stacklok/toolhive/pkg/auth/discovery"
-	"github.com/stacklok/toolhive/pkg/auth/oauth"
 	"github.com/stacklok/toolhive/pkg/secrets"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/auth/remote/handler.go
```diff
@@ -282,12 +283,21 @@ func (h *Handler) tryRestoreFromCachedTokens(
 		}
 	}
 
-	// Create token source from cached refresh token
-	baseSource := CreateTokenSourceFromCached(
-		oauth2Config,
-		refreshToken,
-		h.config.CachedTokenExpiry,
-	)
+	// Create token source from cached refresh token.
```
amirejaz, line 285: nit: the handler now inlines logic similar to CreateTokenSourceFromCached. You could instead extend that helper with an optional resource parameter and call it from the handler to avoid duplication. There are not many references to `CreateTokenSourceFromCached`

amirejaz:
Approved, thanks for addressing the feedback.

PR #4300, opened 2026-03-20
Fix inspector URL using wrong transport for proxied stdio servers

### LINKED ISSUE #4298: `thv inspector` builds wrong URL for stdio servers proxied as streamable-http
## Bug

`thv inspector` generates an incorrect MCP Inspector URL for servers that use `stdio` transport but are proxied as `streamable-http`. The URL uses `transport=sse` with a `/sse` path instead of `transport=streamable-http` with `/mcp`.

### Steps to reproduce

1. Run a server that uses stdio transport with streamable-http proxy mode (e.g., `thv run github`)
2. Run `thv inspector github`
3. Observe the generated URL

### Expected

```
http://localhost:6274?transport=streamable-http&serverUrl=http://host.docker.internal:56677/mcp&MCP_PROXY_AUTH_TOKEN=...
```

### Actual

```
http://localhost:6274?transport=sse&serverUrl=http://host.docker.internal:56677/sse&MCP_PROXY_AUTH_TOKEN=...
```

Opening the inspector requires manually changing the protocol from SSE to HTTP and fixing the URL path from `/sse` to `/mcp`.

### Root cause

`getServerPortAndTransport` in `cmd/thv/app/inspector.go` reads `c.TransportType` (the server's native transport, e.g. `stdio`) instead of `c.ProxyMode` (how the proxy actually exposes the server, e.g. `streamable-http`).

Then the URL-building logic at line 176 treats `stdio` the same as `sse`:

```go
if transportType == types.TransportTypeSSE || transportType == types.TransportTypeStdio {
    suffix = sseSuffix           // "sse"
    transportTypeStr = sseSuffix // "sse"
}
```

The fix should use `ProxyMode` (or the effective proxy mode) to determine the inspector URL transport and path, similar to how `pkg/workloads/manager.go:204-209` already uses `ProxyMode` instead of `TransportType` for client-facing transport decisions.

danbarr:
## Summary

- `thv inspector` was building the MCP Inspector URL using the server's native `TransportType` (e.g. `stdio`) instead of `ProxyMode` to determine the transport and path suffix. This caused stdio servers proxied as streamable-http to get a URL with `transport=sse` and `/sse` instead of `transport=streamable-http` and `/mcp`.
- Extracted URL-building logic into a `buildInspectorURL` helper and changed `getServerPortAndTransport` to return `types.ProxyMode` from the workload's `ProxyMode` field.

Fixes #4298

## Type of change

- [x] Bug fix

## Test plan

Ran `thv inspector github` against a running github server (stdio proxied as streamable-http) and verified the URL now uses `transport=streamable-http` with `/mcp` path.

Yes — `thv inspector` now generates the correct MCP Inspector URL for servers proxied as streamable-http. Previously, users had to manually change the transport and path in the inspector UI.

### DIFF
--- a/cmd/thv/app/inspector.go
+++ b/cmd/thv/app/inspector.go
@@ -112,7 +112,7 @@ func inspectorCmdFunc(cmd *cobra.Command, args []string) error {
 	authToken := hex.EncodeToString(tokenBytes)
 
 	// find the port of the server if it is running / exists
-	serverPort, proxyMode, err := getServerPortAndTransport(ctx, serverName)
+	serverPort, proxyMode, err := getServerPortAndProxyMode(ctx, serverName)
 	if err != nil {
 		return fmt.Errorf("failed to find server: %w", err)
 	}
@@ -184,7 +184,7 @@ func inspectorCmdFunc(cmd *cobra.Command, args []string) error {
 	}
 }
 
-func getServerPortAndTransport(ctx context.Context, serverName string) (int, types.ProxyMode, error) {
+func getServerPortAndProxyMode(ctx context.Context, serverName string) (int, types.ProxyMode, error) {
 	manager, err := workloads.NewManager(ctx)
 	if err != nil {
 		return 0, types.ProxyModeStreamableHTTP, fmt.Errorf("failed to create status manager: %w", err)
@@ -241,12 +241,10 @@ func cleanupInspectorContainer(ctx context.Context, name string) error {
 // buildInspectorURL constructs the URL for the MCP Inspector UI, encoding the
 // transport mode, server address, and authentication token as query parameters.
 func buildInspectorURL(uiPort int, proxyMode types.ProxyMode, serverPort int, authToken string) string {
-	suffix := func() string {
-		if proxyMode == types.ProxyModeSSE {
-			return sseSuffix
-		}
-		return "mcp"
-	}()
+	suffix := "mcp"
+	if proxyMode == types.ProxyModeSSE {
+		suffix = sseSuffix
+	}
 	return fmt.Sprintf(
 		"http://localhost:%d?transport=%s&serverUrl=http://host.docker.internal:%d/%s&MCP_PROXY_AUTH_TOKEN=%s",
 		uiPort, proxyMode, serverPort, suffix, authToken)


cmd/thv/app/inspector.go
```diff
@@ -195,35 +184,30 @@ func inspectorCmdFunc(cmd *cobra.Command, args []string) error {
 	}
 }
 
-func getServerPortAndTransport(ctx context.Context, serverName string) (int, types.TransportType, error) {
-	// Instantiate the status manager and list all workloads.
+func getServerPortAndTransport(ctx context.Context, serverName string) (int, types.ProxyMode, error) {
```
reyortiz3: nit: `getServerPortAndProxyMode`

cmd/thv/app/inspector.go
```diff
@@ -253,3 +237,17 @@ func cleanupInspectorContainer(ctx context.Context, name string) error {
...
+
+// buildInspectorURL constructs the URL for the MCP Inspector UI, encoding the
+// transport mode, server address, and authentication token as query parameters.
+func buildInspectorURL(uiPort int, proxyMode types.ProxyMode, serverPort int, authToken string) string {
+	suffix := func() string {
+		if proxyMode == types.ProxyModeSSE {
+			return sseSuffix
+		}
+		return "mcp"
+	}()
```
reyortiz3: nit:
```
suffix := "mcp"
if proxyMode == types.ProxyModeSSE {
	suffix = sseSuffix
}
```

cmd/thv/app/inspector.go
```diff
@@ -171,18 +171,7 @@ func inspectorCmdFunc(cmd *cobra.Command, args []string) error {
...
-			suffix = sseSuffix
-			transportTypeStr = sseSuffix
-		} else {
-			suffix = "mcp"
-			transportTypeStr = "streamable-http"
-		}
-		inspectorURL := fmt.Sprintf(
-			"http://localhost:%d?transport=%s&serverUrl=http://host.docker.internal:%d/%s&MCP_PROXY_AUTH_TOKEN=%s",
-			inspectorUIPort, transportTypeStr, serverPort, suffix, authToken)
+		inspectorURL := buildInspectorURL(inspectorUIPort, proxyMode, serverPort, authToken)
```
reyortiz3, line 174: nit (just to consider): probably small enough that didn't need to be a separate helper method

```
suffix := "mcp"
if proxyMode == types.ProxyModeSSE {
	suffix = sseSuffix
}
inspectorURL := fmt.Sprintf(
		"http://localhost:%d?transport=%s&serverUrl=http://host.docker.internal:%d/%s&MCP_PROXY_AUTH_TOKEN=%s",
		uiPort, proxyMode, serverPort, suffix, authToken)
```

cmd/thv/app/inspector.go
danbarr: Done 👍

cmd/thv/app/inspector.go
danbarr: Done 👍

cmd/thv/app/inspector.go
danbarr: Fair point — keeping it extracted for now since it lets us unit test the URL construction directly (which is how we caught the original bug pattern). But agreed it's borderline.

reyortiz3: not blocking for this comment since so far I've seen it's a current pattern -- and it's the fact that as a general consensus, from testing techniques and methodologies perspective, to not unit test private/helper methods directly but test indirectly through the public APIs (aka, public methods, obvservable behaviors).

danbarr: I'll go ahead and merge this to get the fix in, but happy to re-address it in a way that follows the team's consensus approach. I did see an existing e2e for inspector but it's a pretty shallow check; adding additional tests for each combination of modes felt heavy.

Issue #4301, opened 2026-03-20
awesome-agent-skills-mcp fails in ToolHive: list_skills unusable due to missing git + response format mismatch


jsprof26:
## Summary
When [`awesome-agent-skills-mcp](https://github.com/shadowrootdev/awesome-agent-skills-mcp)` is launched by ToolHive as a npm converted to Docker container, `list_skills` fails to provide usable results. The container cannot sync the skills repo because `git` is not installed (`spawn git ENOENT`), and the tool response format is JSON-in-text rather than structured output matching the declared `outputSchema`.

## Environment
* **OS:** Windows 11 (10.0.26200)
* **Docker Desktop:** 4.65.0
* **Docker Engine:** 29.2.1

## Steps to Reproduce
1. Start ToolHive with `awesome-agent-skills-mcp` enabled (containerized).
2. Call `refresh_skills` or `list_skills` through ToolHive.
3. Inspect container logs (`docker logs <container-id>`).

## Actual Behavior
* Container logs show git sync failure: `Failed to initialize git repository: GitError: Error: spawn git ENOENT`
* Attempted command includes clone of: `https://github.com/VoltAgent/awesome-agent-skills.git`
* `list_skills` returns empty data (`{"skills":[],"total":0}`) when sync failed.
* ToolHive reports list failure / structured output mismatch (tool has `outputSchema` but response is not treated as structured).

## Expected Behavior
* `refresh_skills` should successfully clone/pull the skills repository.
* `list_skills` should return a populated, schema-valid structured object (not JSON serialized into a text field).
* ToolHive should be able to parse the response without structured-content errors.

## Evidence (from container)
**Git missing in container:**
`docker exec <container-id> git --version` -> executable not found

**Relevant log lines:**
* `Failed to initialize git repository: GitError: Error: spawn git ENOENT`
* `Cloning repository from https://github.com/VoltAgent/awesome-agent-skills.git...`
* `Repository sync failed: Failed to initialize: Error: spawn git ENOENT`

**Tool response observed:**
`{"skills":[],"total":0}` via `content[0].text`

## Suspected Root Causes
1. **Container image missing git**
   * Repo sync cannot run, so skills registry stays empty.
2. **Tool output format mismatch**
   * `list_skills` declares an `outputSchema` object but returns JSON string inside `content: [{ type: "text", text: JSON.stringify(...) }]`, which ToolHive may not accept as structured output.

## Proposed Fix
[...]

PR #4304, opened 2026-03-21
Merge runtime config override with defaults

lorr1:
## Summary

`--runtime-add-package` without `--runtime-image` created a `RuntimeConfig` override with an empty `BuilderImage`, which `loadRuntimeConfig` returned as-is — bypassing defaults and producing an invalid Dockerfile (`FROM  AS builder`). Even with `--runtime-image`, the override's `AdditionalPackages` replaced the defaults instead of merging with them.

Separately, the npx, go, and uvx Dockerfile templates only installed `AdditionalPackages` in the builder stage. The runtime stage hardcoded `ca-certificates`, so runtime dependencies like `git` added via `--runtime-add-package` were missing from the final container.

Fixes first half of #4301

## Type of change

- [x] Bug fix

## Test plan

Built `thv` from the branch and ran `thv run --transport stdio --runtime-add-package git --name awesome npx://awesome-agent-skills-mcp`. Verified:
1. Dockerfile renders `FROM node:22-alpine AS builder` (not empty)
2. Both builder and runtime stages run `apk add --no-cache git ca-certificates`
3. Container starts, clones the skills repo with git, and loads 547 skills

## Changes

| File | Change |
|------|--------|
| `pkg/runner/protocol.go` | Add `mergeRuntimeConfig` that fills empty `BuilderImage` from defaults and dedup-merges `AdditionalPackages`; update `loadRuntimeConfig` to call it |
| `pkg/runner/protocol_test.go` | Add `TestMergeRuntimeConfig` (5 table cases) and `TestLoadRuntimeConfigMergesOverrideWithDefaults` |
| `pkg/container/templates/npx.tmpl` | Runtime stage: use `AdditionalPackages` instead of hardcoded `ca-certificates` |
| `pkg/container/templates/go.tmpl` | Same |
| `pkg/container/templates/uvx.tmpl` | Same (debian, alpine, and default branches) |
| `pkg/container/templates/templates_test.go` | Add `TestRuntimeStageInstallsAdditionalPackages` covering all three transports |

`--runtime-add-package` now works correctly without requiring `--runtime-image`, and packages are available at runtime inside the container.

### DIFF
--- a/docs/cli/thv_run.md
+++ b/docs/cli/thv_run.md
@@ -172,7 +172,7 @@ thv run [flags] SERVER_OR_IMAGE_OR_PROTOCOL [-- ARGS...]
       --remote-forward-headers stringArray          Headers to inject into requests to remote MCP server (format: Name=Value, can be repeated)
       --remote-forward-headers-secret stringArray   Headers with secret values from ToolHive secrets manager (format: Name=secret-name, can be repeated)
       --resource-url string                         Explicit resource URL for OAuth discovery endpoint (RFC 9728)
-      --runtime-add-package stringArray             Add additional packages to install in the builder stage (can be repeated)
+      --runtime-add-package stringArray             Add additional packages to install in the builder and runtime stages (can be repeated)
       --runtime-image string                        Override the default base image for protocol schemes (e.g., golang:1.24-alpine, node:20-alpine, python:3.11-slim)
       --secret stringArray                          Specify a secret to be fetched from the secrets manager and set as an environment variable (format: NAME,target=TARGET)
       --target-host string                          Host to forward traffic to (only applicable to SSE or Streamable HTTP transport) (default "127.0.0.1")
--- a/docs/server/docs.go
+++ b/docs/server/docs.go
@@ -740,7 +740,7 @@ const docTemplate = `{
                 "description": "RuntimeConfig allows overriding the default runtime configuration\nfor this specific workload (base images and packages)",
                 "properties": {
                     "additional_packages": {
-                        "description": "AdditionalPackages lists extra packages to install in builder stage\nExamples for Alpine: [\"git\", \"make\", \"gcc\"]\nExamples for Debian: [\"git\", \"build-essential\"]",
+                        "description": "AdditionalPackages lists extra packages to install in the builder and\nruntime stages.\nExamples for Alpine: [\"git\", \"make\", \"gcc\"]\nExamples for Debian: [\"git\", \"build-essential\"]",
                         "items": {
                             "type": "string"
                         },
@@ -748,7 +748,7 @@ const docTemplate = `{
                         "uniqueItems": false
                     },
                     "builder_image": {
-                        "description": "BuilderImage is the full image reference for the builder stage\nExamples: \"golang:1.25-alpine\", \"node:22-alpine\", \"python:3.13-slim\"",
+                        "description": "BuilderImage is the full image reference for the builder stage.\nAn empty string signals \"use the default for this transport type\" during config merging.\nExamples: \"golang:1.25-alpine\", \"node:22-alpine\", \"python:3.13-slim\"",
                         "type": "string"
                     }
                 },
--- a/docs/server/swagger.json
+++ b/docs/server/swagger.json
@@ -733,7 +733,7 @@
                 "description": "RuntimeConfig allows overriding the default runtime configuration\nfor this specific workload (base images and packages)",
                 "properties": {
                     "additional_packages": {
-                        "description": "AdditionalPackages lists extra packages to install in builder stage\nExamples for Alpine: [\"git\", \"make\", \"gcc\"]\nExamples for Debian: [\"git\", \"build-essential\"]",
+                        "description": "AdditionalPackages lists extra packages to install in the builder and\nruntime stages.\nExamples for Alpine: [\"git\", \"make\", \"gcc\"]\nExamples for Debian: [\"git\", \"build-essential\"]",
                         "items": {
                             "type": "string"
                         },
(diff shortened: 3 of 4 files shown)


pkg/container/templates/npx.tmpl
```diff
@@ -69,7 +69,7 @@ RUN cat /tmp/custom-ca.crt >> /etc/ssl/certs/ca-certificates.crt && \
 {{end}}
 
 # Install runtime dependencies
-RUN apk add --no-cache ca-certificates
+RUN apk add --no-cache {{range .RuntimeConfig.AdditionalPackages}}{{.}} {{end}}
```
aponcedeleonch, line 75: If `AdditionalPackages` is ever empty (e.g. user config specifies `additional_packages: []`, or `GetDefaultRuntimeConfig` is called with an unknown transport type returning an empty `RuntimeConfig{}`), this renders as:
```
RUN apk add --no-cache 
```
…which fails the Docker build. Same issue in `go.tmpl` and `uvx.tmpl` (all three branches).

Consider wrapping with an `{{if}}` guard:
```
{{if .RuntimeConfig.AdditionalPackages}}
RUN apk add --no-cache {{range .RuntimeConfig.AdditionalPackages}}{{.}} {{end}}
{{end}}
```

pkg/runner/protocol.go
```diff
@@ -153,19 +153,23 @@ func createTemplateData(
...
-	// If override is provided, validate and use it
+	// If override is provided, merge with defaults before validating
 	if override != nil {
-		if err := override.Validate(); err != nil {
+		merged := mergeRuntimeConfig(transportType, override)
+		if err := merged.Validate(); err != nil {
 			return nil, fmt.Errorf("invalid runtime config override: %w", err)
 		}
-		return override, nil
+		return merged, nil
```
aponcedeleonch, line 172: The override path now correctly merges with defaults — nice. However, the user config path just below (via `provider.GetRuntimeConfig()`) still returns the raw config without merging. If a user writes a `config.yaml` with `additional_packages: [git]` (omitting `ca-certificates`), the runtime stage will no longer install it since the templates no longer hardcode `ca-certificates`.

On `main` this was safe because the runtime stage always hardcoded `ca-certificates` regardless of config. With this PR that safety net is gone.

Suggest applying the same merge logic to the user config path as well, e.g.:
```go
if rc := provider.GetRuntimeConfig(string(transportType)); rc != nil {
    merged := mergeRuntimeConfig(transportType, rc)
    if err := merged.Validate(); err != nil {
        return nil, fmt.Errorf("invalid user runtime config: %%w", err)
    }
    return merged, nil
}
```

pkg/container/templates/npx.tmpl
aponcedeleonch: Addressed — all templates now have `{{if .RuntimeConfig.AdditionalPackages}}` guards, and the PR also includes `TestEmptyAdditionalPackagesDoesNotBreakBuild` verifying this. Thanks!

pkg/runner/protocol.go
aponcedeleonch: Addressed — both the override path and the user config path now call `mergeRuntimeConfig()`. Looks good.

pkg/container/templates/runtime_config.go
```diff
@@ -26,7 +26,8 @@ type RuntimeConfig struct {
 	// Examples: "golang:1.25-alpine", "node:22-alpine", "python:3.13-slim"
 	BuilderImage string `json:"builder_image" yaml:"builder_image"`
```
aponcedeleonch, line 28: Nit: the `BuilderImage` field comment says "full image reference" but doesn't mention that an empty value is valid and means "inherit from defaults during merge." The `Validate()` godoc already mentions this, but callers reading the struct definition might miss it.

```suggestion
	// BuilderImage is the full image reference for the builder stage.
	// An empty string signals "use the default for this transport type" during config merging.
	// Examples: "golang:1.25-alpine", "node:22-alpine", "python:3.13-slim"
	BuilderImage string `json:"builder_image" yaml:"builder_image"`
```

pkg/container/templates/templates_test.go
```diff
@@ -392,6 +392,133 @@ func TestGetDockerfileTemplate(t *testing.T) {
...
+			parts := strings.SplitN(result, "\nFROM ", 2)
+			if len(parts) < 2 {
+				t.Fatal("Dockerfile does not contain a second FROM (runtime stage)")
+			}
+			runtimeStage := parts[1]
+
+			if !strings.Contains(runtimeStage, tt.wantInRuntime) {
+				t.Errorf("runtime stage does not install %q.\nRuntime stage:\n%s", tt.wantInRuntime, runtimeStage)
+			}
+		})
```
aponcedeleonch, line 458: Nit: nice that this test exists — it's exactly the scenario I was worried about. Consider also adding a `nil` `AdditionalPackages` case (as opposed to empty slice) since `{{if}}` treats both as falsy but they behave differently in Go. Not blocking.

aponcedeleonch (dismissed):
Both previous comments addressed — nice work. Two minor nits below, neither blocking. You might need to regenerate the swagger docs with `task docs` for the CI to pass though

The empty-packages nit I had in mind is already covered by `TestEmptyAdditionalPackagesDoesNotBreakBuild` — great to see that included.

pkg/container/templates/runtime_config.go
lorr1: Changed docstring.

pkg/container/templates/templates_test.go
lorr1: Done!

PR #4309, opened 2026-03-21
Add multi-violation CEL integration tests for OIDCConfigRef and AuthzConfigRef

ChrisJBurns:
## Summary

Adds some extra test cases to the CEL scenarios

### DIFF
--- a/cmd/thv-operator/test-integration/mcp-server/mcpserver_cel_validation_integration_test.go
+++ b/cmd/thv-operator/test-integration/mcp-server/mcpserver_cel_validation_integration_test.go
@@ -164,6 +164,38 @@ var _ = Describe("CEL Validation for OIDCConfigRef and AuthzConfigRef", Label("k
 		})
 	})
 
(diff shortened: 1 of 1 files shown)


ChrisJBurns:
@claude please update the pr description

jhrozek (approved):
Reviewed the full changeset (49 lines, single test file). Tests are technically correct — all assertions match actual CEL rule behavior and will pass. The `And(ContainSubstring(...))` pattern is the right Gomega idiom for order-independent multi-message checks, resource names are unique, and the file follows existing conventions.

Minor observations (non-blocking):
- The second OIDC test (`oidc-inline-with-k8s`, line 183) only triggers one CEL rule (the kubernetes rule), so it's a single-violation test in the multi-violation context — not incorrect, just mislabeled organizationally.
- The reverse AuthzConfigRef multi-violation direction (`type=inline` with only `configMap` set) isn't covered, but this is additive.

Neither is worth blocking on. LGTM.

Issue #4312, opened 2026-03-21
Composite tools: elicitation message doesn't expand template expressions


danbarr:
## Bug

Template expressions in elicitation step `message` fields (e.g. `{{.params.owner}}`) are not expanded — the raw placeholders are passed through to the client.

## Example

```yaml
- id: approval
  type: elicitation
  message: >
    PR #{{.params.pullNumber}} context gathered. Create a pending
    review on {{.params.owner}}/{{.params.repo}}?
```

Displays the literal `{{.params.pullNumber}}` instead of the resolved value.

## Root cause

In `pkg/vmcp/composer/workflow_engine.go`, tool step arguments are explicitly expanded via `templateExpander.Expand()` (~line 404), but `executeElicitationStep()` (~line 639) passes `step.Elicitation` directly to `RequestElicitation()` without any template expansion.

The elicitation handler in `pkg/vmcp/composer/elicitation_handler.go` (~line 168) then passes the raw `config.Message` straight into the MCP elicitation request.

## Suggested fix

Expand `step.Elicitation.Message` through the template expander before passing it to `RequestElicitation()` in `executeElicitationStep()`. Something like:

```go
expandedMessage, err := e.templateExpander.Expand(ctx, step.Elicitation.Message, workflowCtx)
if err != nil {
    return fmt.Errorf("failed to expand elicitation message: %w", err)
}
step.Elicitation.Message = expandedMessage.(string)
```

## Related

Documentation gaps for composite tools tracked in stacklok/docs-website#627.

danbarr:
Also related: the code comments in `pkg/vmcp/composer/composer.go` (lines 87-92) document the elicitation template paths as `{{.steps.stepid.action}}` and `{{.steps.stepid.content}}`, but the actual working paths are `{{.steps.stepid.output.action}}` and `{{.steps.stepid.output.content}}`. These comments should be updated to match the actual behavior. Tracked on the docs side in stacklok/docs-website#629.

PR #4313, opened 2026-03-22
fix(composer): expand template expressions in elicitation messages

saschabuehrle:
## Bug
https://github.com/stacklok/toolhive/issues/4312 — Template expressions like `{{.params.owner}}` in elicitation step `message` fields are passed through as raw placeholders instead of being expanded.

## Fix
Adds template expansion for `step.Elicitation.Message` in `executeElicitationStep()` before passing it to `RequestElicitation()`. This mirrors how `executeToolStep()` already expands `step.Arguments` through the template expander (line ~404).

The expansion wraps the message string in a temporary map to reuse the existing public `TemplateExpander.Expand()` interface, keeping the change minimal and consistent with the established expansion pattern.

## Testing
- Added `TestWorkflowEngine_ElicitationMessageTemplateExpansion` integration test that verifies `{{.params.repo}}` and `{{.params.env}}` placeholders in an elicitation message are expanded to their resolved values before reaching the SDK elicitation requester.

Happy to address any feedback.

Greetings, saschabuehrle


pkg/vmcp/composer/workflow_engine.go
```diff
@@ -634,6 +634,21 @@ func (e *workflowEngine) executeElicitationStep(
...
+			return err
+		}
+		if msg, ok := expanded["message"].(string); ok {
+			step.Elicitation.Message = msg
+		}
+	}
+
 	// Request elicitation (synchronous - blocks until response or timeout)
 	// Per MCP 2025-06-18: SDK handles JSON-RPC ID correlation internally
 	response, err := e.elicitationHandler.RequestElicitation(ctx, workflowCtx.WorkflowID, step.ID, step.Elicitation)
```
jerm-dro: This mutates `step.Elicitation.Message` in place. Since `step` is a pointer to the workflow definition, the original template could be lost after expansion — e.g., if workflows are re-executed or steps retried, the template wouldn't be available for re-expansion.

Compare with `executeToolStep` (line 404) which stores the expanded result in a local `expandedArgs` variable and never modifies `step.Arguments`.

Suggest constructing the config immutably via an immediately-invoked function, and passing the result to `RequestElicitation`:

```go
// Expand template expressions in elicitation message (e.g. {{.params.owner}})
elicitCfg, err := func() (*ElicitationConfig, error) {
    if step.Elicitation.Message == "" {
        return step.Elicitation, nil
    }
    wrapper := map[string]any{"message": step.Elicitation.Message}
    expanded, expandErr := e.templateExpander.Expand(ctx, wrapper, workflowCtx)
    if expandErr != nil {
        return nil, fmt.Errorf("%w: failed to expand elicitation message for step %s: %v",
            ErrTemplateExpansion, step.ID, expandErr)
    }
    if msg, ok := expanded["message"].(string); ok {
        cfgCopy := *step.Elicitation
        cfgCopy.Message = msg
        return &cfgCopy, nil
    }
    return step.Elicitation, nil
}()
if err != nil {
    workflowCtx.RecordStepFailure(step.ID, err)
    return err
}

response, err := e.elicitationHandler.RequestElicitation(ctx, workflowCtx.WorkflowID, step.ID, elicitCfg)
```

pkg/vmcp/composer/elicitation_integration_test.go
```diff
@@ -527,3 +527,61 @@ func TestDefaultElicitationHandler_SDKErrorHandling(t *testing.T) {
...
+	}
+
+	result, err := engine.ExecuteWorkflow(context.Background(), workflow, params)
+	require.NoError(t, err)
+	assert.Equal(t, WorkflowStatusCompleted, result.Status)
+
+	// Verify that template placeholders were expanded in the message
+	assert.Equal(t, "Deploy acme/widget to production?", capturedReq.Params.Message,
+		"elicitation message should have template expressions expanded")
+}
```
jerm-dro, line 605: Could you restructure this as a table-driven test? The happy path is well covered, but there are a few more cases worth exercising. Something like:

```go
func TestWorkflowEngine_ElicitationMessageTemplateExpansion(t *testing.T) {
    t.Parallel()

    tests := []struct {
        name            string
        message         string
        params          map[string]any
        expectedMessage string
        expectError     bool
    }{
        {
            name:            "template expressions expanded",
            message:         "Deploy {{.params.repo}} to {{.params.env}}?",
            params:          map[string]any{"repo": "acme/widget", "env": "production"},
            expectedMessage: "Deploy acme/widget to production?",
        },
        {
            name:            "empty message skips expansion",
            message:         "",
            params:          map[string]any{"repo": "acme/widget"},
            expectedMessage: "",
        },
        {
            name:            "no template expressions passes through unchanged",
            message:         "Are you sure?",
            params:          map[string]any{},
            expectedMessage: "Are you sure?",
        },
        {
            name:        "malformed template returns error",
            message:     "Deploy {{.params.repo to env",
            params:      map[string]any{"repo": "acme/widget"},
            expectError: true,
        },
    }
    // ...
}
```

This matches the table-driven style used elsewhere in this file (e.g., `TestDefaultElicitationHandler_SDKErrorHandling`).

saschabuehrle:
Thanks for the review — I’ve gone through the requested changes and queued follow-up updates. I’ll push the revision in a dedicated follow-up pass.

saschabuehrle:
Pushed an update for both points.

The step config stays immutable now (message expansion happens on a copied elicitation config), and the integration coverage is table-driven with both templated and plain message cases.

Greetings, saschabuehrle

PR #4314, opened 2026-03-22
Webhook Middleware Phase 2: Validating webhook middleware

### LINKED ISSUE #3397: Webhook Middleware Phase 2: Validating webhook middleware
## Overview

Implement validating webhook middleware that calls external HTTP services to approve or deny MCP requests. This middleware allows organizations to plug in external policy engines, approval workflows, or rate limiters.

**Depends on**: Phase 1 (Core webhook package)

## Files to Create

| File | Purpose |
|------|---------|
| `pkg/webhook/validating/middleware.go` | Validating webhook middleware implementation |
| `pkg/webhook/validating/config.go` | Configuration types and validation |

## Files to Modify

| File | Changes |
|------|---------|
| `pkg/runner/middleware.go` | Register `validating-webhook` in `GetSupportedMiddlewareFactories()` |
| `pkg/runner/config.go` | Add `ValidatingWebhooks []webhook.WebhookConfig` field to RunConfig |

## Middleware Implementation

```go
// pkg/webhook/validating/middleware.go
const MiddlewareType = "validating-webhook"

type MiddlewareParams struct {
    Webhooks []webhook.WebhookConfig `json:"webhooks"`
}

type Middleware struct {
    client     *webhook.Client
    webhooks   []webhook.WebhookConfig
    middleware types.MiddlewareFunction
}

func (m *Middleware) Handler() types.MiddlewareFunction {
    return m.middleware
}

func (m *Middleware) Close() error {
    return nil
}

func CreateMiddleware(config *types.MiddlewareConfig, runner types.MiddlewareRunner) error {
    // Implementation
}
```

## Request/Response Format

**Request to webhook** (POST):
```json
{
  "version": "v0.1.0",
  "uid": "unique-request-id",
  "timestamp": "2025-01-22T10:30:00Z",
  "principal": {
    "sub": "user123",
    "email": "user@example.com",
    "groups": ["engineering"]
  },
  "mcp_request": {
    "method": "tools/call",
    "params": { "name": "database_query", "arguments": {...} }
  },
  "context": {
    "server_name": "my-mcp-server",
    "transport": "sse",
    "source_ip": "192.0.2.1"
  }
}
```

**Response (allowed)**:
```json
{
  "version": "v0.1.0",
  "uid": "unique-request-id",
  "allowed": true
}
```

**Response (denied)**:
```json
{
  "version": "v0.1.0",
  "uid": "unique-request-id",
  "allowed": false,
  "code": 403,
  "message": "Production writes require approval",
  "reason": "RequiresApproval",
  "details": {
    "ticket_url": "https://tickets.example.com/PROD-1234"
  }
}
```

## Middleware Chain Position

Validating webhooks should be placed after MCP Parser and before Authorization:

```
Auth -> Token Exchange -> Tool Filter -> MCP Parser ->
[Mutating Webhooks] -> [Validating Webhooks] -> Telemetry -> Authorization -> Audit -> Recovery
```

## Failure Policies

| Policy | Behavior on webhook error |
|--------|---------------------------|
| `fail` (fail-closed) | Deny request with 403 |
| `ignore` (fail-open) | Allow request to continue |

Error conditions:
- Network errors
- Timeout
- HTTP 5xx from webhook
- Invalid JSON response
- Missing required fields

## Multiple Webhooks

When multiple validating webhooks are configured:
1. Execute in configuration order
2. If ANY webhook returns `allowed: false`, deny the request
3. If ALL webhooks return `allowed: true`, allow the request
4. Failure policy applies per-webhook

## Tests

- Unit tests with mock webhook servers (use `httptest`)
- Tests for `allowed=true` and `allowed=false` responses
- Tests for failure policies (fail-closed, fail-open)
- Tests for multiple webhooks in chain
- Tests for timeout handling
- Tests for invalid response handling
- Integration tests with middleware chain

## Acceptance Criteria

- [ ] Middleware registered in `GetSupportedMiddlewareFactories()`
- [ ] Correctly calls webhooks with RFC-compliant request format
- [ ] Handles `allowed: true/false` responses correctly
- [ ] Implements both failure policies
- [ ] Supports multiple webhooks in chain
- [ ] Comprehensive unit tests with >80% coverage
- [ ] Integration test with full middleware chain

Sanskarzz:
## Overview
This PR implements **Phase 2** of the Dynamic Webhook Middleware feature by introducing the **Validating Webhook Middleware**. Validating webhooks allow ToolHive to call external HTTP services (such as policy engines, bespoke approval workflows, or rate limiters) to strictly evaluate, approve, or deny MCP requests before they reach backend tools.

Fixes #3397 

### 1. `pkg/webhook/validating` Package
- **Configuration (`config.go`)**: Added `MiddlewareParams` struct supporting a chain of `webhook.Config` elements. Includes setup validation requiring >0 webhooks to be explicitly declared.
- **Middleware Handler (`middleware.go`)**: 
  - Implementation of the ToolHive `types.Middleware` interface factory.
  - Automatically intercepts MCP POST requests (post-parsing).
  - Composes the HTTP evaluation payload, embedding the original raw JSON-RPC `MCPRequest`, extracting User Principal attributes directly from the `auth.Identity` context, and recording the request Origin Context (`SourceIP`, `Transport`, `ServerName`).
  - **Evaluation Engine**: Invokes all configured webhooks sequentially. It eagerly denies the entire request (HTTP 403) providing an optional custom error message as soon as any webhook responds with `allowed: false`.
  - **Failure Policies**: Accurately respects `FailurePolicyFail` (fail-closed, blocks request on network/server errors) and `FailurePolicyIgnore` (fail-open, logs a warning on exception but continues pipeline).
- **Test Suite (`middleware_test.go`)**: Complete parallelized test-suite covering `Allowed=true` paths, denial paths, both failure policies, connection errors, and safe bypass for non-MCP calls. (Test Coverage sits above **75%**).

### 2. Runner Integration (`pkg/runner`)
- **`middleware.go`**: 
  - Registered `validating.CreateMiddleware` inside `GetSupportedMiddlewareFactories`.
  - Added dedicated configuration wiring (`addValidatingWebhookMiddleware`) securely positioning the validating evaluation block sequentially **after** `mcp-parser` but precisely **before** auditing (`telemetry`, `authz`). Thus blocking unverified telemetry pollution or unauthorized execution.
- **`config.go`**:
  - Expanded the global `RunConfig` exposing the `ValidatingWebhooks []webhook.Config` slice.

## Testing Performed
- Run `go test ./pkg/webhook/validating/... ./pkg/runner/...` (All unit tests passing).
- Run `task lint` / `task lint-fix` against the overall project (clean).

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

### DIFF
--- a/pkg/webhook/types.go
+++ b/pkg/webhook/types.go
@@ -62,17 +62,17 @@ type TLSConfig struct {
 // Config holds the configuration for a single webhook.
 type Config struct {
 	// Name is a unique identifier for this webhook.
-	Name string `json:"name"`
+	Name string `json:"name" yaml:"name"`
 	// URL is the HTTPS endpoint to call.
-	URL string `json:"url"`
+	URL string `json:"url" yaml:"url"`
 	// Timeout is the maximum time to wait for a webhook response.
-	Timeout time.Duration `json:"timeout"`
+	Timeout time.Duration `json:"timeout" yaml:"timeout" swaggertype:"primitive,integer"`
 	// FailurePolicy determines behavior when the webhook call fails.
-	FailurePolicy FailurePolicy `json:"failure_policy"`
+	FailurePolicy FailurePolicy `json:"failure_policy" yaml:"failure_policy"`
 	// TLSConfig holds optional TLS configuration (CA bundles, client certs).
-	TLSConfig *TLSConfig `json:"tls_config,omitempty"`
+	TLSConfig *TLSConfig `json:"tls_config,omitempty" yaml:"tls_config,omitempty"`
 	// HMACSecretRef is an optional reference to an HMAC secret for payload signing.
-	HMACSecretRef string `json:"hmac_secret_ref,omitempty"`
+	HMACSecretRef string `json:"hmac_secret_ref,omitempty" yaml:"hmac_secret_ref,omitempty"`
 }
 
 // Validate checks that the WebhookConfig has valid required fields.
--- a/pkg/webhook/validating/middleware.go
+++ b/pkg/webhook/validating/middleware.go
@@ -1,0 +1,202 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package validating
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"time"
+
+	"github.com/google/uuid"
+	"golang.org/x/exp/jsonrpc2"
+
+	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/mcp"
+	"github.com/stacklok/toolhive/pkg/transport/types"
+	"github.com/stacklok/toolhive/pkg/webhook"
+)
+
+// MiddlewareType is the type constant for the validating webhook middleware.
+const MiddlewareType = "validating-webhook"
+
+// FactoryMiddlewareParams extends MiddlewareParams with context for the factory.
+type FactoryMiddlewareParams struct {
+	MiddlewareParams
+	// ServerName is the name of the ToolHive instance.
+	ServerName string `json:"server_name"`
+	// Transport is the transport type (e.g., sse, stdio).
+	Transport string `json:"transport"`
+}
+
+// Middleware wraps validating webhook functionality for the factory pattern.
+type Middleware struct {
+	handler types.MiddlewareFunction
+}
+
+// Handler returns the middleware function used by the proxy.
+func (m *Middleware) Handler() types.MiddlewareFunction {
+	return m.handler
+}
+
+// Close cleans up any resources used by the middleware.
+func (*Middleware) Close() error {
+	return nil
+}
+
+type clientExecutor struct {
+	client *webhook.Client
+	config webhook.Config
+}
+
+// CreateMiddleware is the factory function for validating webhook middleware.
+func CreateMiddleware(config *types.MiddlewareConfig, runner types.MiddlewareRunner) error {
+	var params FactoryMiddlewareParams
+	if err := json.Unmarshal(config.Parameters, &params); err != nil {
+		return fmt.Errorf("failed to unmarshal validating webhook middleware parameters: %w", err)
+	}
+
+	if err := params.Validate(); err != nil {
+		return fmt.Errorf("invalid validating webhook configuration: %w", err)
+	}
+
+	// Create clients for each webhook
+	var executors []clientExecutor
+	for i, whCfg := range params.Webhooks {
+		client, err := webhook.NewClient(whCfg, webhook.TypeValidating, nil) // HMAC secret not yet plumbed
+		if err != nil {
+			return fmt.Errorf("failed to create client for webhook[%d] (%q): %w", i, whCfg.Name, err)
+		}
+		executors = append(executors, clientExecutor{client: client, config: whCfg})
+	}
+
+	mw := &Middleware{
+		handler: createValidatingHandler(executors, params.ServerName, params.Transport),
+	}
+	runner.AddMiddleware(MiddlewareType, mw)
+	return nil
+}
+
+func createValidatingHandler(executors []clientExecutor, serverName, transport string) types.MiddlewareFunction {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Skip if it's not a parsed MCP request (middleware runs after mcp parser)
+			parsedMCP := mcp.GetParsedMCPRequest(r.Context())
+			if parsedMCP == nil {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			// Read the request body to get the raw MCP request
+			bodyBytes, err := io.ReadAll(r.Body)
+			if err != nil {
+				sendErrorResponse(w, http.StatusInternalServerError, "Failed to read request body", parsedMCP.ID)
+				return
+			}
+			// Restore the request body for downstream handlers
+			r.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+
+			// Build the webhook request payload
+			reqUID := uuid.New().String()
+			whReq := &webhook.Request{
+				Version:    webhook.APIVersion,
+				UID:        reqUID,
+				Timestamp:  time.Now().UTC(),
+				MCPRequest: json.RawMessage(bodyBytes),
+				Context: &webhook.RequestContext{
+					ServerName: serverName,
+					SourceIP:   readSourceIP(r),
+					Transport:  transport,
+				},
+			}
+
+			// Add Principal if authenticated
+			if identity, ok := auth.IdentityFromContext(r.Context()); ok {
+				whReq.Principal = identity.GetPrincipalInfo()
+			}
+
+			// Call each webhook in order
+			for _, exec := range executors {
+				whName := exec.config.Name
+
+				resp, err := exec.client.Call(r.Context(), whReq)
+				if err != nil {
+					// Handle error based on failure policy
+					if exec.config.FailurePolicy == webhook.FailurePolicyIgnore {
+						slog.Warn("Validating webhook error ignored due to fail-open policy",
+							"webhook", whName, "error", err)
+						continue
+					}
+
+					slog.Error("Validating webhook error caused request denial",
+						"webhook", whName, "error", err)
+					sendErrorResponse(w, http.StatusForbidden, "Request denied by policy", parsedMCP.ID)
+					return
+				}
+
+				if !resp.Allowed {
+					slog.Info("Validating webhook denied request", "webhook", whName, "reason", resp.Reason, "message", resp.Message)
+
+					// Prevent information leaks by ignoring the webhook's message
+					msg := "Request denied by policy"
+
+					code := http.StatusForbidden
+
+					sendErrorResponse(w, code, msg, parsedMCP.ID)
+					return
+				}
+			}
+
+			// All webhooks allowed or ignored errors
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func readSourceIP(r *http.Request) string {
+	// Let runner handle X-Forwarded-For if TrustProxyHeaders is set.
+	// For now, simple RemoteAddr.
+	return r.RemoteAddr
+}
+
+func convertToJSONRPC2ID(id interface{}) (jsonrpc2.ID, error) {
+	if id == nil {
+		return jsonrpc2.ID{}, nil
+	}
+
+	switch v := id.(type) {
+	case string:
+		return jsonrpc2.StringID(v), nil
+	case int:
+		return jsonrpc2.Int64ID(int64(v)), nil
+	case int64:
+		return jsonrpc2.Int64ID(v), nil
+	case float64:
+		// JSON numbers are often unmarshaled as float64
+		return jsonrpc2.Int64ID(int64(v)), nil
+	default:
+		return jsonrpc2.ID{}, fmt.Errorf("unsupported ID type: %T", id)
+	}
+}
+
+func sendErrorResponse(w http.ResponseWriter, statusCode int, message string, msgID interface{}) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(statusCode)
+
+	id, err := convertToJSONRPC2ID(msgID)
+	if err != nil {
+		id = jsonrpc2.ID{} // Use empty ID if conversion fails
+	}
+
+	// Return a JSON-RPC 2.0 error so MCP clients can parse the denial.
+	// The HTTP status code signals the error at the transport level; the JSON-RPC body carries the detail.
+	errResp := &jsonrpc2.Response{
+		ID:    id,
+		Error: jsonrpc2.NewError(int64(statusCode), message),
+	}
+	_ = json.NewEncoder(w).Encode(errResp)
+}
--- a/pkg/webhook/validating/middleware_test.go
+++ b/pkg/webhook/validating/middleware_test.go
@@ -1,0 +1,510 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package validating
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/mcp"
+	"github.com/stacklok/toolhive/pkg/transport/types"
+	"github.com/stacklok/toolhive/pkg/webhook"
+)
+
+// closedServerURL is a URL that will always fail to connect (port 0 is reserved/closed).
+const closedServerURL = "http://127.0.0.1:0"
+
+//nolint:paralleltest // Shares a mock HTTP server and lastRequest state
+func TestValidatingMiddleware(t *testing.T) {
+	// Setup a mock webhook server
+	var lastRequest webhook.Request
+	mockResponse := webhook.Response{
+		Version: webhook.APIVersion,
+		UID:     "resp-uid",
+		Allowed: true,
+	}
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		require.Equal(t, http.MethodPost, r.Method)
+		require.Equal(t, "application/json", r.Header.Get("Content-Type"))
+
+		err := json.NewDecoder(r.Body).Decode(&lastRequest)
+		require.NoError(t, err)
+
+		w.Header().Set("Content-Type", "application/json")
+		err = json.NewEncoder(w).Encode(mockResponse)
+		require.NoError(t, err)
+	}))
+	defer server.Close()
+
+	// Create middleware handler
+	config := []webhook.Config{
+		{
+			Name:          "test-webhook",
+			URL:           server.URL,
+			Timeout:       webhook.DefaultTimeout,
+			FailurePolicy: webhook.FailurePolicyFail,
+			TLSConfig: &webhook.TLSConfig{
+				InsecureSkipVerify: true, // Need this for httptest server
+			},
+		},
+	}
+
+	var executors []clientExecutor
+	for _, cfg := range config {
+		client, err := webhook.NewClient(cfg, webhook.TypeValidating, nil)
+		require.NoError(t, err)
+		executors = append(executors, clientExecutor{client: client, config: cfg})
+	}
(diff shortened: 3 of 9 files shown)


Sanskarzz:
@JAORMX This PR is ready for review.

JAORMX:
@Sanskarzz you might need to rebase, there is something off with the PR as it's showing it's 39 commits.

pkg/webhook/validating/middleware.go
```diff
@@ -0,0 +1,187 @@
...
+					// Handle error based on failure policy
+					if exec.config.FailurePolicy == webhook.FailurePolicyIgnore {
+						slog.Warn("Validating webhook error ignored due to fail-open policy",
+							"webhook", whName, "error", err)
+						continue
+					}
+
+					slog.Error("Validating webhook error caused request denial",
+						"webhook", whName, "error", err)
+					sendErrorResponse(w, http.StatusForbidden, "Forbidden", fmt.Sprintf("Webhook %q error: %v", whName, err))
```
JAORMX: blocker: So... this is sending the webhook name *and* the raw error back to the client. That error can contain internal URLs, DNS names, connection details like `dial tcp 10.0.0.5:8443: connection refused`. Even though the user is authenticated at this point, we really shouldn't be leaking our internal topology to clients.

The good news is you're already logging the details on lines 134-135, which is the right place for them. The client response should just be generic.

```suggestion
				sendErrorResponse(w, http.StatusForbidden, \"Forbidden\", \"Request denied by policy\")
```

pkg/webhook/validating/middleware.go
```diff
@@ -0,0 +1,187 @@
...
+					return
+				}
+
+				if !resp.Allowed {
+					slog.Info("Validating webhook denied request", "webhook", whName, "reason", resp.Reason, "message", resp.Message)
+
+					msg := resp.Message
+					if msg == "" {
+						msg = fmt.Sprintf("Webhook %q denied the request", whName)
```
JAORMX: blocker: Same thing here. The default deny message exposes the internal webhook name to the client. The name is already in the log on line 141, so the client doesn't need it.

```suggestion
						msg = \"Request denied by policy\"
```

pkg/webhook/validating/middleware.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// Since we are intercepting an MCP request, we should really be returning a JSON-RPC error.
+	// However, if the error happens before actual execution, a standard HTTP error or a basic JSON
+	// with error details is typical. Here we'll follow standard HTTP error structure or JSON-RPC format.
+	// We'll return a JSON format that could be interpreted as a JSON-RPC error.
+	errResp := map[string]any{
+		"jsonrpc": "2.0",
+		"id":      nil,
+		"error": map[string]any{
+			"code":    statusCode,
```
JAORMX: suggestion: So, I went down a rabbit hole checking whether using HTTP status codes as JSON-RPC error codes is correct. The [JSON-RPC 2.0 spec](https://www.jsonrpc.org/specification#error_object) reserves `-32768` to `-32000` for pre-defined errors (with `-32000` to `-32099` for implementation-defined server errors), and says "the remainder of the space is available for application defined errors." So positive integers like `403` are technically in the "application defined" space and not *forbidden* by the spec.

The [MCP spec (2025-11-25)](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports) also explicitly allows HTTP 403 with a JSON-RPC error response body with no `id`, which is what we're doing here.

And the existing authz middleware (`pkg/authz/middleware.go:144`) already does `jsonrpc2.NewError(403, errorMsg)`, so this is an established pattern in the codebase.

One thing that *is* worth fixing: you're building the JSON-RPC response by hand with `map[string]any` instead of using the `jsonrpc2` library like the authz middleware does. Using the library would be more consistent and less error-prone.

pkg/webhook/validating/middleware.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+}
+
+func readSourceIP(r *http.Request) string {
+	// Let runner handle X-Forwarded-For if TrustProxyHeaders is set.
+	// For now, simple RemoteAddr.
+	return r.RemoteAddr
+}
+
+func sendErrorResponse(w http.ResponseWriter, statusCode int, _, message string) {
```
JAORMX: nitpick: The third parameter is explicitly discarded (`_`). If you don't need it, just drop it from the signature. Dead parameters are confusing for the next person reading this.

```suggestion
func sendErrorResponse(w http.ResponseWriter, statusCode int, message string) {
```

(This means updating the three call sites to drop their third argument too.)

pkg/webhook/validating/middleware.go
```diff
@@ -0,0 +1,187 @@
...
+
+				if !resp.Allowed {
+					slog.Info("Validating webhook denied request", "webhook", whName, "reason", resp.Reason, "message", resp.Message)
+
+					msg := resp.Message
+					if msg == "" {
+						msg = fmt.Sprintf("Webhook %q denied the request", whName)
+					}
+
+					code := resp.Code
```
JAORMX: question: This lets the webhook service control the HTTP status code we return to the client (any 4xx-5xx). Is that intentional? A misconfigured webhook could return `code: 503` and make the proxy look like it's having service issues, or `code: 401` and confuse auth flows.

The webhook's job is allow/deny... not to pick our HTTP semantics. Worth considering whether we should just always return 403 here and ignore the webhook's code preference.

pkg/webhook/validating/middleware.go
```diff
@@ -0,0 +1,187 @@
...
+	// Let runner handle X-Forwarded-For if TrustProxyHeaders is set.
+	// For now, simple RemoteAddr.
+	return r.RemoteAddr
+}
+
+func sendErrorResponse(w http.ResponseWriter, statusCode int, _, message string) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(statusCode)
+
+	// Since we are intercepting an MCP request, we should really be returning a JSON-RPC error.
```
JAORMX: nitpick: This comment reads like you're still deciding what to do ("we should really be returning...", "here we'll follow..."). Since this is a deliberate choice, the comment should state the decision, not the deliberation. Something like:

> Return a JSON-RPC 2.0 error so MCP clients can parse the denial. The HTTP status code signals the error at the transport level; the JSON-RPC body carries the detail.

pkg/webhook/types.go
```diff
@@ -66,7 +66,7 @@ type Config struct {
 	// URL is the HTTPS endpoint to call.
 	URL string `json:"url"`
 	// Timeout is the maximum time to wait for a webhook response.
-	Timeout time.Duration `json:"timeout"`
+	Timeout time.Duration `json:"timeout" yaml:"timeout" swaggertype:"primitive,integer"`
 	// FailurePolicy determines behavior when the webhook call fails.
 	FailurePolicy FailurePolicy `json:"failure_policy"`
```
JAORMX: blocker: So, `Timeout` got a `yaml` tag in this PR (line 69), but the other fields here (`Name`, `URL`, `FailurePolicy`, `TLSConfig`, `HMACSecretRef`) are still missing `yaml` tags. Since `RunConfig.ValidatingWebhooks` uses `yaml:"validating_webhooks"` for YAML deserialization, the nested `Config` fields need explicit `yaml` tags too.

Go's YAML library uses lowercased struct field names by default (e.g., `failurepolicy`), *not* the `json` tag values. So a YAML config with `failure_policy: fail` would silently fail to deserialize into `FailurePolicy`. That's a runtime bug that won't show up until someone actually tries to configure this via YAML.

All the fields need `yaml` tags matching the `json` ones. For example:

```go
FailurePolicy FailurePolicy `json:"failure_policy" yaml:"failure_policy"`
```

pkg/auth/remote/config.go
```diff
@@ -23,7 +23,7 @@ type Config struct {
 	ClientSecretFile string        `json:"client_secret_file,omitempty" yaml:"client_secret_file,omitempty"`
 	Scopes           []string      `json:"scopes,omitempty" yaml:"scopes,omitempty"`
 	SkipBrowser      bool          `json:"skip_browser,omitempty" yaml:"skip_browser,omitempty"`
-	Timeout          time.Duration `json:"timeout,omitempty" yaml:"timeout,omitempty" swaggertype:"string" example:"5m"`
+	Timeout          time.Duration `json:"timeout,omitempty" yaml:"timeout,omitempty" swaggertype:"primitive,integer"`
```
JAORMX: suggestion: This swagger type fix (`string` -> `primitive,integer`) is unrelated to the validating webhook feature and changes the OpenAPI spec contract for the auth timeout field. Could you split this into its own PR? Keeps this one focused on the webhook middleware.

pkg/webhook/validating/middleware_test.go
```diff
@@ -0,0 +1,313 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/mcp"
+	"github.com/stacklok/toolhive/pkg/transport/types"
+	"github.com/stacklok/toolhive/pkg/webhook"
+)
+
+//nolint:paralleltest // Shares a mock HTTP server and lastRequest state
+func TestValidatingMiddleware(t *testing.T) {
```
JAORMX, line 27: suggestion: The issue (#3397) explicitly calls out testing multiple webhooks in a chain:

> "When multiple validating webhooks are configured: Execute in configuration order. If ANY webhook returns `allowed: false`, deny the request."

The current tests only exercise a single webhook. We need at least:

- **First allows, second denies** (verifies sequential evaluation)
- **First denies** (verifies short-circuit... the second webhook should never be called)
- **Mixed failure policies** (first with `fail`, second with `ignore`, or vice versa)

These are important because the sequential loop in `createValidatingHandler` is the core logic that differentiates this from a single-webhook setup. Without these tests, we can't be confident the chain actually works as specified.

Also a few more branching paths worth covering:
- When `resp.Message` is empty (exercises the default message on line 144-145)
- When the webhook returns an out-of-range code like `0` or `200` (exercises the `code < 400 || code > 599` guard on line 149)
- When `auth.IdentityFromContext` returns `ok=false` (verifies `Principal` is nil in the payload)

Sanskarzz:
> @Sanskarzz you might need to rebase, there is something off with the PR as it's showing it's 39 commits.

@JAORMX , I have fixed the commit issue. The extra commits were caused by rebasing against my fork's stale `main` instead of `upstream/main`. I've since cleaned that up — the PR should now reflect only the relevant Phase 2 commits. I'll address all the review comments shortly.

pkg/webhook/validating/middleware.go
Sanskarzz: Now I have used the library "golang.org/x/exp/jsonrpc2"

pkg/webhook/validating/middleware.go
Sanskarzz: Done

pkg/webhook/validating/middleware.go
Sanskarzz: Removed internal webhook names and raw errors from the client-facing responses. It now returns a clean JSON-RPC error structure with a generic message (`"Request denied by policy"`).

pkg/webhook/validating/middleware.go
Sanskarzz: Done

pkg/webhook/validating/middleware.go
Sanskarzz: On denial, we now explicitly return an HTTP `403` status regardless of the webhook's preference, avoiding client-side confusion.

pkg/webhook/validating/middleware.go
Sanskarzz: Done

pkg/webhook/types.go
Sanskarzz: Added the missing explicit `yaml` tags correctly for the `webhook.Config` fields so no runtime bugs happen with YAML definitions.

pkg/auth/remote/config.go
Sanskarzz: Reverted the `pkg/auth/remote/config.go` swagger tag changes; I'll submit those separately if needed.

pkg/webhook/validating/middleware_test.go
Sanskarzz: *   Added `TestMultiWebhookChain` to verify sequential evaluation, short-circuiting when the first webhook denies, and mixed failure policies (e.g., fail-open on connection error).
 *   Added subtests for branching paths: out-of-range codes (e.g., `0`, `200` defaulting to `403`), empty `resp.Message`, and requests without an authenticated identity.

pkg/auth/remote/config.go
Sanskarzz: I have created [Issue](https://github.com/stacklok/toolhive/issues/4353) and [PR](https://github.com/stacklok/toolhive/pull/4354) fix for this integration.

Now the current PR is dependent

Sanskarzz: I'll fix the CI docs error after merging the [PR](https://github.com/stacklok/toolhive/pull/4354).

JAORMX:
@Sanskarzz I think you need to run `task docs` given the `swagger docs` failure.

Sanskarzz:
> @Sanskarzz I think you need to run `task docs` given the `swagger docs` failure.

@JAORMX  The CI is failing because swagger type fix (string -> primitive,integer). I mentioned this in the [review comment discussion](https://github.com/stacklok/toolhive/pull/4314#discussion_r2984470785) above. 

The Fix depends on this [PR](https://github.com/stacklok/toolhive/pull/4354) merge

JAORMX:
@Sanskarzz I missed that, let me check it out

pkg/webhook/validating/middleware.go
```diff
@@ -0,0 +1,177 @@
...
+	// Let runner handle X-Forwarded-For if TrustProxyHeaders is set.
+	// For now, simple RemoteAddr.
+	return r.RemoteAddr
+}
+
+func sendErrorResponse(w http.ResponseWriter, statusCode int, message string) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(statusCode)
+
+	// Return a JSON-RPC 2.0 error so MCP clients can parse the denial.
```
JAORMX, line 195: suggestion: The JSON-RPC error response uses `jsonrpc2.ID{}` which serializes as `"id": 0` — a fabricated ID that could collide with a real request. Since this middleware runs after the MCP parser, the actual request ID is available in context via `parsedMCP.ID`.

The authz middleware (`pkg/authz/middleware.go:128-156`) already handles this correctly by extracting the request ID with `convertToJSONRPC2ID()`. This middleware should follow the same pattern for JSON-RPC 2.0 / MCP spec compliance and consistency.

Per the MCP spec (2025-11-25), error responses MUST include the same ID as the request they correspond to. The Streamable HTTP transport does allow id-less error bodies for HTTP 403, but that exception is for transport-level rejections (bad Origin, unparseable request), not application-level policy denials where the request was already successfully parsed.

Concretely, `sendErrorResponse` could accept the parsed ID and convert it, falling back to empty on conversion failure — same as authz does. The `convertToJSONRPC2ID` helper in `pkg/authz` could be extracted to a shared location, or duplicated here if you'd prefer to keep the packages decoupled for now.

Sanskarzz: I copied the convertToJSONRPC2ID function over from pkg/authz to pkg/webhook/validating/middleware.go and used it.

Sanskarzz: Also updated the signature to receive the request ID directly via an interface{} (msgID interface{}). Using convertToJSONRPC2ID and also updated all invocations of sendErrorResponse(...) in the createValidatingHandler middleware so that they pass parsedMCP.ID

Sanskarzz:
> @Sanskarzz I missed that, let me check it out

@JAORMX  I have closed the [PR](https://github.com/stacklok/toolhive/pull/4354), and the CI swag doc error has been fixed.

pkg/webhook/validating/middleware.go
```diff
@@ -0,0 +1,202 @@
...
+				}
+
+				if !resp.Allowed {
+					slog.Info("Validating webhook denied request", "webhook", whName, "reason", resp.Reason, "message", resp.Message)
+
+					// Prevent information leaks by ignoring the webhook's message
+					msg := "Request denied by policy"
+
+					code := http.StatusForbidden
+
```
JAORMX, line 148: nitpick: Now that the webhook's code/message are intentionally ignored (right call), the `msg` and `code` variables are just unnecessary indirection. Could inline directly:

```suggestion
					sendErrorResponse(w, http.StatusForbidden, "Request denied by policy", parsedMCP.ID)
```

Four fewer lines and it's immediately obvious that 403 + generic message is the deliberate policy.

pkg/webhook/validating/middleware.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+func readSourceIP(r *http.Request) string {
+	// Let runner handle X-Forwarded-For if TrustProxyHeaders is set.
+	// For now, simple RemoteAddr.
+	return r.RemoteAddr
+}
+
+func convertToJSONRPC2ID(id interface{}) (jsonrpc2.ID, error) {
```
JAORMX, line 166: suggestion (follow-up): This is a copy of `convertToJSONRPC2ID` from `pkg/authz/middleware.go`. Two copies means they can drift. Worth extracting to a shared location in a follow-up so both authz and validating use the same implementation.

Not blocking this PR... just worth tracking.

pkg/webhook/validating/middleware_test.go
```diff
@@ -0,0 +1,510 @@
...
+		err := json.Unmarshal(rr.Body.Bytes(), &errResp)
+		require.NoError(t, err)
+
+		errObj, ok := errResp["Error"].(map[string]interface{})
+		require.True(t, ok)
+		assert.Equal(t, float64(http.StatusForbidden), errObj["code"])
+		assert.Equal(t, "Request denied by policy", errObj["message"])
+	})
+
+	t.Run("Denied Request - Out-of-Range Code Defaults to 403", func(t *testing.T) {
```
JAORMX, line 178: nitpick: The test name says "Out-of-Range Code Defaults to 403" which implies branching logic that normalizes the webhook's code. But the implementation always returns 403 regardless of `resp.Code` (the field is never read). The test validates the right behavior, but the name is misleading.

Something like `"Denied Request - Ignores Webhook Code Field"` would more accurately describe the invariant.

JAORMX (approved):
This looks good to merge. The inline comments are all minor (nitpicks + a follow-up suggestion). Up to you whether you want to address them in this PR or as a follow-up... either way works for me.

Sanskarzz:
> This looks good to merge. The inline comments are all minor (nitpicks + a follow-up suggestion). Up to you whether you want to address them in this PR or as a follow-up... either way works for me.


@JAORMX  I have squashed the commits into one, and I've already created the draft [PR](https://github.com/stacklok/toolhive/pull/4372) for phase 3 changes. I will address the above review comments and suggestions in Phase 3.
If this works for you, I think the PR is ready to merge.

PR #4316, opened 2026-03-23
Replace webhook.Principal with auth.Identity

### LINKED ISSUE #4315: Replace webhook.Principal with auth.Identity
## Problem

PR #3840 introduced `webhook.Principal` in `pkg/webhook/types.go`, but `auth.Identity` in `pkg/auth/identity.go` already represents the same concept. `webhook.Principal` is a strict subset:

| Field | `auth.Identity` | `webhook.Principal` |
|---|---|---|
| Subject | `Subject string` | `Sub string` |
| Name | `Name string` | `Name string` |
| Email | `Email string` | `Email string` |
| Groups | `Groups []string` | `Groups []string` |
| Claims | `Claims map[string]any` | `Claims map[string]any` |
| Token | `Token string` | — |
| TokenType | `TokenType string` | — |
| Metadata | `Metadata map[string]string` | — |

Having two types means:
- Identity changes must be kept in sync across both
- The middleware layer will need a mapping function to convert between them
- Consumers need to know which type to use in which context

## Proposed fix

Replace `webhook.Principal` with `auth.Identity` in the `webhook.Request` struct. The extra fields on `Identity` (Token, TokenType, Metadata) can be left empty when constructing the webhook request — the JSON `omitempty` tags on `Identity.MarshalJSON` already handle redaction/omission.

## Files affected

- `pkg/webhook/types.go` — remove `Principal` struct, import `auth.Identity`
- `pkg/webhook/client_test.go` — update test fixtures

JAORMX:
## Summary

- The webhook package (merged in #3840) defined its own `Principal` struct, but `auth.Identity` already represents the same concept with a superset of fields. Maintaining two parallel identity types creates sync burden and will require conversion logic when the middleware layer wires them together.
- Remove `webhook.Principal` and change `Request.Principal` to use `*auth.Identity` directly. Update all test fixtures accordingly.

Closes #4315

## Type of change

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/webhook/types.go` | Remove `Principal` struct, change `Request.Principal` type to `*auth.Identity`, add `pkg/auth` import |
| `pkg/webhook/client_test.go` | Replace `&Principal{Sub: "user1"}` with `&auth.Identity{Subject: "user1"}` (5 instances), add `pkg/auth` import |

- No import cycle: `pkg/auth` imports only stdlib.
- The JSON field name changes from `"sub"` to `"subject"` in the webhook payload. This is acceptable because the webhook API is pre-release (`v0.1.0`) with no external consumers yet.
- `auth.Identity` has a custom `MarshalJSON` that redacts `Token` — this is a feature, not a problem, since tokens should never leak into webhook payloads.
- `auth.Identity.MarshalJSON` emits all fields (including empty ones) rather than using `omitempty`. This is slightly more verbose but consistent with how identity is serialized elsewhere.

### DIFF
--- a/pkg/webhook/types.go
+++ b/pkg/webhook/types.go
@@ -119,7 +119,8 @@ type Request struct {
 	// Timestamp is when the request was created.
 	Timestamp time.Time `json:"timestamp"`
 	// Principal contains the authenticated user's identity information.
-	Principal *auth.Identity `json:"principal"`
+	// Uses PrincipalInfo (not Identity) so credentials never enter the webhook payload.
+	Principal *auth.PrincipalInfo `json:"principal"`
 	// MCPRequest is the raw MCP JSON-RPC request.
 	MCPRequest json.RawMessage `json:"mcp_request"`
 	// Context provides additional metadata about the request origin.
--- a/pkg/audit/auditor_test.go
+++ b/pkg/audit/auditor_test.go
@@ -482,10 +482,12 @@ func TestExtractSubjects(t *testing.T) {
 
 		req := httptest.NewRequest("GET", "/test", nil)
 		identity := &auth.Identity{
-			Subject: claims["sub"].(string),
-			Name:    claims["name"].(string),
-			Email:   claims["email"].(string),
-			Claims:  claims,
+			PrincipalInfo: auth.PrincipalInfo{
+				Subject: claims["sub"].(string),
+				Name:    claims["name"].(string),
+				Email:   claims["email"].(string),
+				Claims:  claims,
+			},
 		}
 		ctx := auth.WithIdentity(req.Context(), identity)
 		req = req.WithContext(ctx)
@@ -507,8 +509,10 @@ func TestExtractSubjects(t *testing.T) {
 
 		req := httptest.NewRequest("GET", "/test", nil)
 		identity := &auth.Identity{
-			Subject: claims["sub"].(string),
-			Claims:  claims,
+			PrincipalInfo: auth.PrincipalInfo{
+				Subject: claims["sub"].(string),
+				Claims:  claims,
+			},
 		}
 		ctx := auth.WithIdentity(req.Context(), identity)
 		req = req.WithContext(ctx)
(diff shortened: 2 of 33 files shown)


JAORMX:
@Sanskarzz did this one as a follow up of your previous PR introducing the webhook fundamentals.

Sanskarzz:
> @Sanskarzz did this one as a follow up of your previous PR introducing the webhook fundamentals.

@JAORMX, thanks for the ping! Great catch on the duplicate identity types. Unifying them into `auth.Identity` definitely simplifies things and avoids the sync burden you mentioned in the issue. I've taken a quick look at the PR, and the approach looks solid. Thank you for the cleanup!

pkg/webhook/types.go
```diff
@@ -117,28 +119,13 @@ type Request struct {
 	// Timestamp is when the request was created.
 	Timestamp time.Time `json:"timestamp"`
 	// Principal contains the authenticated user's identity information.
-	Principal *Principal `json:"principal"`
+	Principal *auth.Identity `json:"principal"`
```
jhrozek: `auth.Identity` carries a `Token` field with the raw JWT. `MarshalJSON` redacts it to `"REDACTED"`, but the old `Principal` had no token field at all — the credential never entered the struct.

Does it matter that we're now sending `token` (even redacted), `tokenType`, and `metadata` to external webhook endpoints? The consumer doesn't use any of these. Might be worth stripping the token before building the webhook request, or using a projection that only includes the fields webhook receivers actually need.

pkg/webhook/types.go
```diff
@@ -117,28 +119,13 @@ type Request struct {
...
-	// Groups is a list of groups the user belongs to.
-	Groups []string `json:"groups,omitempty"`
-	// Claims contains additional identity claims. Using map[string]any to support
-	// real-world JWT claims which may carry non-string values (arrays, numbers, booleans).
-	Claims map[string]any `json:"claims,omitempty"`
-}
-
 // RequestContext provides metadata about the request origin and environment.
 type RequestContext struct {
 	// ServerName is the ToolHive/vMCP instance name handling the request.
```
jhrozek, line 132: This changes the wire format of `principal` — `"sub"` becomes `"subject"`, empty fields are no longer omitted, and there are new fields (`token`, `tokenType`, `metadata`). No consumers yet so probably fine, but is this intentional?

JAORMX:
@jhrozek I changed the approach to add a subset of the principal to the auth package itself and make the Identity struct take that into use. I believe this is more re-usable.

Issue #263, opened 2025-04-23
Add a method to rotate encryption password


danbarr:
ToolHive currently lacks any method to rotate/change the encryption password. Users must delete the keyring entry (with `thv secret reset-keyring`) and secrets_encrypted file and start over.

It would be good to have a method to do this.

We might also consider letting ToolHive generate a random password, assuming keyring access is verified on the user's system? (This could also imply an additional `thv secret init --generate` command to do this initially, and/or a question when running the first `secret` command?)

### Potential workflow

Interactive version:
```bash
$ thv secret rotate-password
Enter the new encryption password: 

# User enters new password, ToolHive re-encrypts the file, then updates the keyring entry
```

Non-interactive version:
```bash
$ thv secret rotate-password --generate

# ToolHive generates a strong random password, re-encrypts the file, then updates the keyring entry
# IF keyring update fails for any reason, display the generated password to the user so they're not locked out of their file?
```

JAORMX:
Still valid! Having to nuke everything with `reset-keyring` just to change your password isn't ideal.

We've got the infrastructure in place to support this (the `EncryptedManager` can already re-encrypt, and we have `GenerateSecurePassword()` for the `--generate` flag). It's mostly a matter of wiring up the command.

We'll get to this soon.

PR #4318, opened 2026-03-23
Wire session-aware backend routing in proxy transports (RC-12)

### LINKED ISSUE #4212: Wire session-aware backend routing and LRU cache-size in proxy transports (RC-12, RC-13)
## Description

Switch all three proxy transport constructors (streamable, HTTP-SSE, and transparent) to use `session.NewManagerWithRedis`, enabling session-sticky routing across multiple proxyrunner replicas (RC-12).

> **Note**: RC-13 (LRU cache-size wiring) has been dropped. `RedisStorage.Load` uses `GETEX`, which atomically returns session data and refreshes the TTL in a single round-trip, making a local LRU cache layer unnecessary. See [#4294](https://github.com/stacklok/toolhive/pull/4294) for the rationale.

Parent epic: stacklok/stacklok-epics#263

**Dependencies**: #4236 (RC-7 — NewManagerWithRedis constructor), #4239 (TASK-001 — RunConfig fields)
**Blocks**: none (leaf node in the DAG)

### RC-12 — Session-aware routing
- [ ] `NewHTTPProxy` (streamable): constructs session manager via `session.NewManagerWithRedis`; accepts a Redis storage override via `WithSessionStorage` option (defaults to `session.NewLocalStorage()` when not provided)
- [ ] `NewHTTPSSEProxy`: same
- [ ] `NewTransparentProxyWithOptions`: same
- [ ] Transparent proxy `Rewrite` reads `"backend_url"` metadata key from session and overrides `pr.Out.URL`; falls back to static `targetURL` when absent or unparseable
- [ ] Non-initialize requests with unknown session ID return HTTP 400 in transparent proxy handler
- [ ] Existing single-replica behavior preserved when no `WithSessionStorage` is provided (defaults to `LocalStorage`)

### General
- [ ] All existing tests in `pkg/transport/proxy/streamable/`, `pkg/transport/proxy/httpsse/`, and `pkg/transport/proxy/transparent/` pass
- [ ] `go vet ./pkg/transport/proxy/...` reports no issues
- [ ] Tests for transparent proxy `Rewrite`: (a) `backend_url` in metadata overrides target; (b) no `backend_url` falls back to static target; (c) unknown session ID on non-initialize returns HTTP 400

### RC-12 — Storage wiring

Add `WithSessionStorage(storage session.Storage) Option` to each proxy constructor. When not provided, default to `session.NewLocalStorage()`. Pass the storage to `session.NewManagerWithRedis`:

```go
// At call sites in pkg/transport/http.go, when Redis is configured:
mgr := session.NewManagerWithRedis(redisClient)
// Otherwise fall back to local storage (single-replica default)
```

Extend the transparent proxy `Rewrite`:

```go
if sid := pr.In.Header.Get("Mcp-Session-Id"); sid != "" {
    if sess, ok := p.sessionManager.Get(sid); ok {
        if backendURL, _ := sess.GetMetadataValue("backend_url"); backendURL != "" {
            if parsed, err := url.Parse(backendURL); err == nil {
                pr.Out.URL = parsed
            }
        }
    }
}
```

### Code Pointers

- `pkg/transport/proxy/streamable/streamable_proxy.go:72–98` — `NewHTTPProxy`; `session.NewManager` → `session.NewManagerWithRedis`
- `pkg/transport/proxy/streamable/streamable_proxy.go:617–621` — `resolveSessionForBatch`; pattern for HTTP 400 on unknown session
- `pkg/transport/proxy/httpsse/http_proxy.go:93–120` — `NewHTTPSSEProxy`; same replacement
- `pkg/transport/proxy/transparent/transparent_proxy.go:253–310` — `NewTransparentProxyWithOptions`; session manager wiring
- `pkg/transport/proxy/transparent/transparent_proxy.go:487–513` — `Rewrite` closure; `backend_url` override injected here
- `pkg/transport/http.go` — call sites for both proxy constructors

## Out of Scope

- RoutingStorage / LRU cache layer (dropped — see [#4294](https://github.com/stacklok/toolhive/pull/4294))
- `RunConfig.SessionCacheSize` wiring (no longer needed)
- Backend pod selection / round-robin on `initialize`
- StatefulSet replica count wiring (TASK-002) or graceful shutdown (TASK-003)

## References

- [Epic #263](https://github.com/stacklok/stacklok-epics/issues/263)
- [RC-7 #4236: NewManagerWithRedis constructor](https://github.com/stacklok/toolhive/pull/4236)

yrobla:
## Summary

Implements RC-12 from the horizontal scaling RFC (THV-0047, epic #263):

- Adds `WithSessionStorage(session.Storage) Option` to all three proxy transports (streamable, HTTP-SSE, transparent), allowing a Redis-backed session store to be injected for multi-replica deployments
- Transparent proxy now stores `backend_url = targetURI` in session metadata when an `initialize` response returns `Mcp-Session-Id`, and routes subsequent requests to the originating pod via the `Rewrite` closure
- Non-initialize requests with an unknown session ID are rejected with HTTP 400 rather than silently forwarded

Fixes #4212 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
|      |        |

## Large PR Justification

This is a complete PR with an isolated functionality and complete tests. Cannot be split.


pkg/transport/proxy/httpsse/http_proxy.go
```diff
@@ -89,13 +89,32 @@ type HTTPSSEProxy struct {
...
+	return func(p *HTTPSSEProxy) {
+		if storage == nil {
+			return
+		}
+		if p.sessionManager != nil {
+			_ = p.sessionManager.Stop()
+		}
+		sseFactory := func(id string) session.Session { return session.NewSSESession(id) }
+		p.sessionManager = session.NewManagerWithStorage(session.DefaultSessionTTL, sseFactory, storage)
+	}
```
jerm-dro, line 129: **blocker:** `Manager.Range()` is a no-op for non-`LocalStorage` backends (`manager.go:264-268`). HTTPSSEProxy relies on `Range()` in three critical paths: disconnecting all clients on `Stop` (line 223), detecting connected clients for message forwarding (line 278), and broadcasting SSE events to all sessions (line 443). If Redis storage is injected here, all three silently stop working — messages are never delivered and sessions are never disconnected.

This should be addressed before Redis storage is integrated. Ideally we'd remove `Range()` from `Manager` altogether — it's a `LocalStorage`-specific concept that doesn't translate to distributed backends. The SSE proxy needs a different mechanism for tracking live connections that doesn't depend on iterating the session store.

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -428,7 +467,9 @@ func (t *tracingTransport) RoundTrip(req *http.Request) (*http.Response, error)
 			slog.Debug("detected Mcp-Session-Id header", "session_id", ct)
 			internalID := normalizeSessionID(ct)
 			if _, ok := t.p.sessionManager.Get(internalID); !ok {
```
jerm-dro, line 484: **nitpick:** `"backend_url"` is used as a metadata key in three places: stored here, read in the `Rewrite` closure (line 555), and asserted in tests. Consider extracting this to a `const` to avoid silent breakage if one occurrence is changed without the others.

jerm-dro, line 484: **suggestion:** Rather than storing the proxy's own `targetURI` here and overriding scheme/host in the `Rewrite` closure on every request, could we resolve the backend pod's stable DNS name via the K8s API (e.g., `http://mcp-server-0.mcp-server.default.svc:8080`) before creating the session? If the session's `backend_url` is already the pod's full address, the `Rewrite` closure becomes unnecessary — the reverse proxy can target the session's URL directly.

jerm-dro (dismissed):
`Manager.Range()` is a no-op for non-local storage backends, and HTTPSSEProxy depends on it for core functionality. This needs to be resolved before Redis storage can be safely injected. See inline comments for details and two additional suggestions.

pkg/transport/proxy/transparent/transparent_proxy.go
yrobla: The transparent proxy intentionally doesn't resolve pod DNS — it has no K8s API access and only knows its configured targetURI. The backend_url stored on initialize is the proxy's own targetURI as a default; the actual pod-specific URL is meant to be injected by the vMCP/operator layer before the request reaches the proxy (e.g. by setting backend_url directly on the session after resolving mcp-server-0.mcp-server.default.svc:8080 via K8s). When that happens, the Rewrite closure picks it up automatically. In single-replica deployments targetURI == the pod address so the behavior is identical.

pkg/transport/proxy/streamable/streamable_proxy.go
```diff
@@ -68,12 +68,31 @@ type HTTPProxy struct {
...
+		sFactory := func(id string) session.Session { return session.NewStreamableSession(id) }
+		p.sessionManager = session.NewManagerWithStorage(session.DefaultSessionTTL, sFactory, storage)
+	}
+}
+
 // NewHTTPProxy creates a new HTTPProxy for streamable HTTP transport.
 func NewHTTPProxy(
 	host string,
 	port int,
 	prometheusHandler http.Handler,
```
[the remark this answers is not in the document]
yrobla: i do not care about breaking changes

pkg/transport/proxy/httpsse/http_proxy.go
```diff
@@ -89,13 +96,38 @@ type HTTPSSEProxy struct {
...
 // NewHTTPSSEProxy creates a new HTTP SSE proxy for transports.
 func NewHTTPSSEProxy(
 	host string,
 	port int,
 	trustProxyHeaders bool,
 	prometheusHandler http.Handler,
-	middlewares ...types.NamedMiddleware,
+	middlewares []types.NamedMiddleware,
+	opts ...Option,
 ) *HTTPSSEProxy {
```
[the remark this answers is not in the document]
yrobla: i do not care about breaking changes

pkg/transport/proxy/transparent/transparent_proxy.go
jerm-dro: Makes sense. Thanks for explaining.

pkg/transport/proxy/httpsse/http_proxy.go
```diff
@@ -493,6 +527,9 @@ func (p *HTTPSSEProxy) removeClient(clientID string) {
 		sseSession.Disconnect()
```
jerm-dro: **question (blocking):** `removeClient` fetches the session from `sessionManager.Get(clientID)` and calls `Disconnect()` on it. With a distributed storage backend, `Get` deserializes a *new* `SSESession` with a different `MessageCh` — `Disconnect()` closes the wrong channel, leaving the actual live connection in `liveSSESessions` undrained.

Should this disconnect from `liveSSESessions` instead? Something like:
```go
if val, ok := p.liveSSESessions.LoadAndDelete(clientID); ok {
    if sess, ok := val.(*session.SSESession); ok {
        sess.Disconnect()
    }
}
```
Then delete from `sessionManager` after.

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -507,6 +560,22 @@ func (p *TransparentProxy) Start(ctx context.Context) error {
...
 			pr.SetXForwarded()
 
+			// Route to the originating backend pod when session metadata contains backend_url.
+			// Falls back to static targetURL when the session doesn't exist or has no backend_url.
+			if sid := pr.In.Header.Get("Mcp-Session-Id"); sid != "" {
+				if sess, ok := p.sessionManager.Get(normalizeSessionID(sid)); ok {
+					if backendURLStr, exists := sess.GetMetadataValue(sessionMetadataBackendURL); exists && backendURLStr != "" {
+						if parsed, err := url.Parse(backendURLStr); err == nil {
+							pr.Out.URL.Scheme = parsed.Scheme
+							pr.Out.URL.Host = parsed.Host
```
jerm-dro, line 604: **suggestion:** `url.Parse("mcp-server-0:8080")` succeeds but returns empty `Scheme`/`Host`, which would overwrite `pr.Out.URL.Scheme` and `pr.Out.URL.Host` with empty strings and silently break routing. Guard against non-absolute URLs:
```go
if parsed, err := url.Parse(backendURLStr); err == nil && parsed.Scheme != "" && parsed.Host != "" {
    pr.Out.URL.Scheme = parsed.Scheme
    pr.Out.URL.Host = parsed.Host
} else {
    slog.Debug("backend_url from session metadata is not a valid absolute URL; using static target",
        sessionMetadataBackendURL, backendURLStr, "error", err)
}
```
This also warrants a unit test — e.g., a session with `backend_url` set to `"mcp-server-0:8080"` should fall back to the static target, not corrupt the outbound URL.

pkg/transport/proxy/httpsse/http_proxy.go
```diff
@@ -89,13 +96,38 @@ type HTTPSSEProxy struct {
 	closedClientsMutex sync.Mutex
 }
 
+// Option configures an HTTPSSEProxy.
+type Option func(*HTTPSSEProxy)
+
+// WithSessionStorage injects a custom storage backend into the session manager.
+// When not provided, the proxy uses in-memory LocalStorage (single-replica default).
+// Provide a Redis-backed storage for multi-replica deployments so all replicas
```
jerm-dro, line 103: **suggestion:** The godoc on `WithSessionStorage` should document the architectural context: `HTTPSSEProxy` is currently only used by `StdioTransport` for stdio-backed MCP servers. SSE fan-out and POST handling are both local to the instance holding the live connection, so Redis storage enables cross-replica session metadata sharing but does not solve cross-replica message delivery — a POST accepted on replica B won't reach a client whose SSE connection is on replica A. Callers must ensure an external load balancer provides session affinity when using distributed storage with this proxy.

Clients should prefer Streamable HTTP (`ProxyModeStreamableHTTP`), which is also supported on `StdioTransport` and doesn't have this affinity constraint.

PR #4319, opened 2026-03-23
Wire server discovery protocol into thv serve

JAORMX:
## Summary

RFC-0034 proposes a long-running local server architecture where `thv serve` advertises itself via a discovery file so clients (CLI, Studio) can find it without hardcoded ports. The discovery package (`pkg/server/discovery/`) was already fully implemented and tested but had zero imports anywhere in the codebase. This PR wires it in, giving us a working discovery protocol that immediately benefits Studio and the skills CLI client.

- The server now writes a discovery file (`$XDG_STATE_HOME/toolhive/server/server.json`) on startup with the actual listen URL, PID, and a cryptographic nonce, and removes it on shutdown
- The `/health` endpoint returns the nonce via `X-Toolhive-Nonce` so clients can verify server instance identity (prevents PID-reuse confusion)
- The skills client now auto-discovers the server via the discovery file before falling back to `TOOLHIVE_API_URL` or `localhost:8080`, with loopback and socket-path validation on discovered URLs
- Fixes: SIGTERM handling (was only SIGINT), 30-second shutdown timeout (was unbounded), symlink rejection on the discovery file read path, directory permission tightening, constant-time nonce comparison
- A startup guard refuses to start if another healthy server is already running

## Type of change

- [x] New feature

## Test plan

Manual testing: run `thv serve`, verify `server.json` appears with correct URL/PID/nonce, `curl -v /health` shows `X-Toolhive-Nonce` header, stop server and verify file is removed, start a second server while first is running and verify it refuses.

## Changes

| File | Change |
|------|--------|
| `pkg/server/discovery/*.go` | Discovery package (was implemented but unwired) with security hardening: symlink check on read, dir permission tightening, constant-time nonce comparison, exported URL validators |
| `pkg/api/server.go` | Nonce generation, `WithNonce()` builder, `ListenURL()`, discovery write/remove in lifecycle, startup guard, shutdown timeout |
| `pkg/api/v1/healthcheck.go` | Nonce parameter, `X-Toolhive-Nonce` header on 204 responses |
| `cmd/thv/app/server.go` | Add `SIGTERM` to signal handling |
| `pkg/skills/client/client.go` | Discovery-first URL resolution with loopback/socket-path validation |
| `pkg/api/server_test.go` | Tests for `generateNonce`, `ListenURL` (TCP + Unix) |
| `pkg/api/v1/healtcheck_test.go` | Tests for nonce header present/absent/unhealthy |
| `pkg/server/discovery/*_test.go` | Tests for symlink read rejection, dir permission tightening, paralleltest fixes |

`thv serve` now writes a discovery file that allows other tools (Studio, skills CLI) to automatically find the running server without requiring `TOOLHIVE_API_URL` or knowing the port. This is transparent to users -- existing behavior is preserved as a fallback.

- The `pkg/server/discovery/` package was previously implemented and tested in a separate effort but never wired in. This PR includes it as new files since it had no prior commits.
- The startup guard prevents silent orphaning: if a healthy server is already running, the new server refuses to start. Stale discovery files (from crashes/SIGKILL) are automatically cleaned up.
- The nonce is explicitly **not a secret** -- it is a server instance identifier returned in a plaintext HTTP header and stored in a 0600 file. It prevents PID-reuse confusion, not unauthorized access.
- Security review found no high/critical issues. TOCTOU gaps on symlink checks are mitigated by the 0700 directory permissions (single-user trust model).
- Future work: `thv server start/stop/status` subcommands, CLI auto-start, lock file singleton, SSE events, Studio integration.

## Large PR Justification

- This contains both the setup of the server discovery and the usage.

### DIFF
--- a/pkg/server/discovery/health.go
+++ b/pkg/server/discovery/health.go
@@ -62,6 +62,24 @@ func CheckHealth(ctx context.Context, serverURL string, expectedNonce string) er
 // buildHealthClient returns an HTTP client and request URL appropriate for
 // the given server URL scheme.
 func buildHealthClient(serverURL string) (*http.Client, string, error) {
+	client, baseURL, err := HTTPClientForURL(serverURL)
+	if err != nil {
+		return nil, "", err
+	}
+	healthURL, err := url.JoinPath(baseURL, "health")
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to build health URL: %w", err)
+	}
+	return client, healthURL, nil
+}
+
+// HTTPClientForURL returns an HTTP client configured for the given server URL
+// and the base URL to use for requests. For unix:// URLs it creates a client
+// with a Unix socket transport and returns "http://localhost" as the base URL.
+// For http:// URLs it validates the host is a loopback address and returns a
+// default client. The returned client has no timeout set; callers should apply
+// their own timeout via context or client.Timeout.
+func HTTPClientForURL(serverURL string) (*http.Client, string, error) {
 	switch {
 	case strings.HasPrefix(serverURL, "unix://"):
 		socketPath, err := ParseUnixSocketPath(serverURL)
@@ -75,13 +93,13 @@ func buildHealthClient(serverURL string) (*http.Client, string, error) {
 				},
 			},
 		}
-		return client, "http://localhost/health", nil
+		return client, "http://localhost", nil
 
 	case strings.HasPrefix(serverURL, "http://"):
 		if err := ValidateLoopbackURL(serverURL); err != nil {
 			return nil, "", err
 		}
-		return &http.Client{}, serverURL + "/health", nil
+		return &http.Client{}, serverURL, nil
 
 	default:
 		return nil, "", fmt.Errorf("unsupported URL scheme: %s", serverURL)
--- a/pkg/skills/client/client.go
+++ b/pkg/skills/client/client.go
@@ -12,7 +12,6 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
-	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -81,19 +80,23 @@ func NewClient(baseURL string, opts ...Option) *Client {
 //  1. The TOOLHIVE_API_URL environment variable (explicit override)
 //  2. The server discovery file (auto-detected running server)
 //  3. The default URL http://127.0.0.1:8080
-func NewDefaultClient(opts ...Option) *Client {
-	return newDefaultClientWithEnv(&env.OSReader{}, opts...)
+//
+// The context is used for the server discovery health check; it is not stored.
+func NewDefaultClient(ctx context.Context, opts ...Option) *Client {
+	return newDefaultClientWithEnv(ctx, &env.OSReader{}, opts...)
 }
 
 // newDefaultClientWithEnv is the testable core of NewDefaultClient.
-func newDefaultClientWithEnv(envReader env.Reader, opts ...Option) *Client {
+func newDefaultClientWithEnv(ctx context.Context, envReader env.Reader, opts ...Option) *Client {
 	// 1. Explicit env var override always wins.
 	if base := envReader.Getenv(envAPIURL); base != "" {
 		return NewClient(base, opts...)
 	}
 
 	// 2. Try server discovery.
-	if base, httpOpts := resolveViaDiscovery(); base != "" {
+	if base, httpOpts := resolveViaDiscovery(ctx); base != "" {
+		// Discovery opts go first so caller-supplied opts can override them
+		// (e.g. a caller-provided WithTimeout replaces the discovery default).
 		merged := make([]Option, 0, len(httpOpts)+len(opts))
 		merged = append(merged, httpOpts...)
 		merged = append(merged, opts...)
@@ -107,8 +110,8 @@ func newDefaultClientWithEnv(envReader env.Reader, opts ...Option) *Client {
 // resolveViaDiscovery attempts to find a running server via the discovery file.
 // It returns the base URL and any additional options (e.g. a Unix socket transport).
 // On failure it returns empty values and the caller falls back to the default.
-func resolveViaDiscovery() (string, []Option) {
-	result, err := discovery.Discover(context.Background())
+func resolveViaDiscovery(ctx context.Context) (string, []Option) {
+	result, err := discovery.Discover(ctx)
 	if err != nil {
 		slog.Debug("server discovery failed", "error", err)
 		return "", nil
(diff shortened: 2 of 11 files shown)


pkg/api/server.go
```diff
@@ -557,6 +586,50 @@ func (s *Server) Start(ctx context.Context) error {
 	}
 }
 
+// writeDiscoveryFile writes the server discovery file if a nonce is configured.
+// It checks for an existing healthy server first to prevent silent orphaning.
+func (s *Server) writeDiscoveryFile(ctx context.Context) error {
```
aponcedeleonch, line 599: Non-blocking: `writeDiscoveryFile` is not multi-process safe. There's a TOCTOU window between `Discover()` and `WriteServerInfo()` — two processes starting simultaneously could both see `StateNotFound` and overwrite each other's discovery file.

There's an existing `fileutils.WithFileLock` helper that provides OS-level advisory file locking with timeout and retry. The fix would be to wrap the entire check-then-write sequence inside it:

```go
err := fileutils.WithFileLock(discovery.FilePath(), func() error {
    // Discover + switch + WriteServerInfo all inside the lock
})
```

This way the discover-check and write happen atomically from the perspective of other processes.

pkg/skills/client/client.go
```diff
@@ -74,19 +77,80 @@ func NewClient(baseURL string, opts ...Option) *Client {
...
+
+	// 3. Fall back to the default URL.
+	return NewClient(defaultBaseURL, opts...)
+}
+
+// resolveViaDiscovery attempts to find a running server via the discovery file.
+// It returns the base URL and any additional options (e.g. a Unix socket transport).
+// On failure it returns empty values and the caller falls back to the default.
+func resolveViaDiscovery() (string, []Option) {
+	result, err := discovery.Discover(context.Background())
```
aponcedeleonch: Non-blocking / follow-up: `resolveViaDiscovery()` uses `context.Background()`, so the 5s health timeout runs independently of any caller deadline.

The fix would be to propagate context from the call sites: `newSkillClient(ctx)` → `NewDefaultClient(ctx, opts...)` → `resolveViaDiscovery(ctx)`. Every caller in `cmd/thv/app/skill_*.go` is inside a Cobra `RunE` with `cmd.Context()` available, so context is accessible at all 8 call sites. It's mechanical but touches many files.

Given the hard 5s timeout and that health checks to localhost resolve in milliseconds (or fail fast on connection refused), this is low-risk. Fine as a follow-up.

pkg/api/server_test.go
```diff
@@ -0,0 +1,71 @@
...
+		require.NoError(t, err)
+
+		nonce2, err := generateNonce()
+		require.NoError(t, err)
+
+		assert.NotEqual(t, nonce1, nonce2)
+	})
+}
+
+func TestListenURL_TCP(t *testing.T) {
```
aponcedeleonch: Nit: `TestListenURL_TCP` and `TestListenURL_UnixSocket` could be consolidated into a single table-driven test per project conventions.

pkg/api/v1/healthcheck_test.go
```diff
@@ -83,3 +84,101 @@ func TestGetHealthcheck(t *testing.T) {
 		assert.Equal(t, expectedError.Error()+"\n", resp.Body.String())
 	})
 }
+
+func TestGetHealthcheck_ReturnsNonceHeader(t *testing.T) {
```
aponcedeleonch, line 88: Nit: The filename `healtcheck_test.go` has a pre-existing typo (missing 'h'). Since this PR adds new tests to the file, it would be a good time to rename it to `healthcheck_test.go` for consistency with `healthcheck.go`.

pkg/server/discovery/discover.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil, err
+	}
+
+	// Try health check with nonce verification
+	if err := CheckHealth(ctx, info.URL, info.Nonce); err == nil {
+		return &DiscoverResult{State: StateRunning, Info: info}, nil
+	}
+
+	// Health check failed — check if the process is still alive
+	alive, err := process.FindProcess(info.PID)
```
aponcedeleonch, line 76: Nit: When `FindProcess` returns an error (the line below), the error is silently discarded and the result defaults to `StateStale`. Consider logging it at Debug level for diagnostics, consistent with the project's logging conventions.

pkg/api/v1/healthcheck.go
```diff
@@ -9,18 +9,22 @@ import (
...
+	"github.com/stacklok/toolhive/pkg/server/discovery"
 )
 
 // HealthcheckRouter sets up healthcheck route.
-func HealthcheckRouter(containerRuntime rt.Runtime) http.Handler {
-	routes := &healthcheckRoutes{containerRuntime: containerRuntime}
+// The nonce parameter, when non-empty, is returned via the X-Toolhive-Nonce
+// header so clients can verify they are talking to the expected server instance.
+func HealthcheckRouter(containerRuntime rt.Runtime, nonce string) http.Handler {
+	routes := &healthcheckRoutes{containerRuntime: containerRuntime, nonce: nonce}
```
blkt, line 19: **question:** I might be wrong, but I assume the `nonce` is checked for security reasons, could you clarify what's the scenario we're protecting from?

pkg/skills/client/client.go
```diff
@@ -74,19 +77,80 @@ func NewClient(baseURL string, opts ...Option) *Client {
...
+	// 1. Explicit env var override always wins.
+	if base := envReader.Getenv(envAPIURL); base != "" {
+		return NewClient(base, opts...)
+	}
+
+	// 2. Try server discovery.
+	if base, httpOpts := resolveViaDiscovery(); base != "" {
+		merged := make([]Option, 0, len(httpOpts)+len(opts))
+		merged = append(merged, httpOpts...)
+		merged = append(merged, opts...)
```
blkt, line 102: **nit:** this operation shadows `httpOpts` with `opts` when they both provide the same `WithXXX` function with different parameters (assuming said `WithXXX` function always sets the value and does not perform more complex operations). It's probably worth a comment clarifying this is working as intended.

pkg/skills/client/client.go
```diff
@@ -74,19 +77,80 @@ func NewClient(baseURL string, opts ...Option) *Client {
...
+		return "", nil
+	}
+
+	serverURL := result.Info.URL
+
+	// Validate and configure transport based on URL scheme.
+	switch {
+	case strings.HasPrefix(serverURL, "unix://"):
+		socketPath, err := discovery.ParseUnixSocketPath(serverURL)
+		if err != nil {
```
blkt: **nit:** this validation and the one below could happen as part of a `WithBaseURL` (or pair of `WithUnixSocket` and `WithHTTPSocket`) function, improving encapsulation.

pkg/server/discovery/health.go
```diff
@@ -0,0 +1,128 @@
...
+	case strings.HasPrefix(serverURL, "http://"):
+		if err := ValidateLoopbackURL(serverURL); err != nil {
+			return nil, "", err
+		}
+		return &http.Client{}, serverURL + "/health", nil
+
+	default:
+		return nil, "", fmt.Errorf("unsupported URL scheme: %s", serverURL)
+	}
+}
```
blkt, line 107: **nit:** this logic seems duplicated with the one below.

pkg/api/v1/healthcheck.go
JAORMX: Good question! So, the scenario here is a TOCTOU race between the discovery file and the actual server. The discovery file says "hey, there's a server at port 8080"... but by the time the client connects, the original server might have died and something else could be listening on that port. The nonce lets the client verify it's actually talking to the server that wrote the discovery file, not some random process that happened to bind the same address. It's basically instance identity verification.

pkg/skills/client/client.go
JAORMX: Yeah, this is intentional. Discovery opts go first so caller opts can override them if needed. I'll add a comment clarifying that.

pkg/skills/client/client.go
JAORMX: That's a good call. I'd rather do that as a follow-up PR though, since it touches the option constructor pattern and I'd want to think through the API a bit. Keeping this PR focused on the discovery wiring itself.

pkg/server/discovery/health.go
JAORMX: Yeah, both `buildHealthClient` and `resolveViaDiscovery` switch on the URL scheme with similar transport setup. The validation helpers (`ParseUnixSocketPath`, `ValidateLoopbackURL`) are already shared, but the HTTP client construction is duplicated. I'll clean this up in a follow-up PR... probably by extracting a shared transport builder that both can use.

pkg/skills/client/client.go
JAORMX: Done! added a comment in 3de135a0 explaining the intentional ordering (discovery opts first so caller opts can override).

pkg/skills/client/client.go
JAORMX: Addressed partially: the validation now lives inside `discovery.HTTPClientForURL()` which both the health checker and the skills client call. Moving it into `WithBaseURL`/`WithUnixSocket` option constructors is a deeper API redesign that I'll tackle separately.

pkg/server/discovery/health.go
JAORMX: Fixed! extracted `discovery.HTTPClientForURL()` that both `buildHealthClient` and the skills client's `resolveViaDiscovery` now use. Also switched `buildHealthClient` to use `url.JoinPath` instead of string concatenation.

PR #4326, opened 2026-03-24
Respect global config for OTel tracing and metrics enabled

### LINKED ISSUE #4323: Global OTel tracing-enabled and metrics-enabled config ignored by thv run
## Bug description

`thv config otel set-tracing-enabled false` and `thv config otel set-metrics-enabled false` report success, but `thv run` ignores these values. New servers always get `tracingEnabled: true` and `metricsEnabled: true` in their runconfig regardless of global config.

Two bugs contribute to this:

**1. `getTelemetryFromFlags` skips TracingEnabled/MetricsEnabled fallback (functional)**

In `cmd/thv/app/run_flags.go`, `getTelemetryFromFlags` applies global config fallbacks for 6 of 8 OTel fields but omits `TracingEnabled` and `MetricsEnabled`. Additionally, `buildRunConfig` passes raw `runFlags.OtelTracingEnabled`/`runFlags.OtelMetricsEnabled` to `WithTelemetryConfigFromFlags`, bypassing the fallback entirely for the proxy runner config.

Since both CLI flags default to `true`, global config is never consulted and servers always start with telemetry enabled.

**2. `omitempty` on bool fields hides `false` in config.yaml (cosmetic)**

In `pkg/config/config.go`, `OpenTelemetryConfig` uses `omitempty` on `MetricsEnabled` and `TracingEnabled`. Since `false` is Go's zero value, `yaml.Marshal` drops these fields from the YAML output. PR #3761 fixed this in `pkg/telemetry/config.go` (the runtime config struct) but not in `pkg/config/config.go` (the global application config struct).

## Steps to reproduce

```bash
thv config otel set-tracing-enabled false
thv config otel get-tracing-enabled
# Output: Current OpenTelemetry tracing enabled: false

cat ~/Library/Application\ Support/toolhive/config.yaml | grep tracing
# (no output — omitempty drops the field)

thv run <any-server>
jq '.telemetry_config.tracingEnabled' \
  ~/Library/Application\ Support/toolhive/runconfigs/<server>.json
# Output: true (should be false)
```

## Expected behavior

When `tracing-enabled` or `metrics-enabled` is set to `false` in global config, `thv run` should respect that value unless overridden by an explicit CLI flag. The `false` value should also be visible in `config.yaml`.

## Actual behavior

Global config values for `tracing-enabled` and `metrics-enabled` are silently ignored. Servers always start with both enabled (the CLI flag default), regardless of global config.

## Additional context

- Related: #2295 (global OTel settings not applied via API/UI)
- Related: #1355 (prior fix in same function left these two fields out of the fallback)
- Related: #1899 (same symptom for the `insecure` field, since fixed)
- PR #3761 fixed `omitempty` in `pkg/telemetry/config.go` but not `pkg/config/config.go`

gkatz2:
## Summary

- `thv config otel set-tracing-enabled false` was silently ignored by `thv run` because `getTelemetryFromFlags` had no fallback for `TracingEnabled` or `MetricsEnabled`, and `buildRunConfig` bypassed the fallback entirely for the proxy runner config
- Users could not globally disable telemetry without passing CLI flags on every invocation

Fixes #4323

## Type of change

- [x] Bug fix

## Test plan

Deployed a local build and verified:
1. With global config `tracing-enabled: false` and `metrics-enabled: false`, `thv run` produces a runconfig with `null` telemetry (no telemetry initialized)
2. With explicit `--otel-tracing-enabled=true`, CLI flag correctly overrides global config
3. `config.yaml` now shows `tracing-enabled: false` instead of silently dropping the field

## Changes

| File | Change |
|------|--------|
| `pkg/config/config.go` | Change `TracingEnabled`/`MetricsEnabled` from `bool` with `omitempty` to `*bool` without `omitempty`, matching the existing `UseLegacyAttributes` pattern |
| `cmd/thv/app/run_flags.go` | Add config fallback for both fields in `getTelemetryFromFlags`; use resolved values in `buildRunConfig`; return nil from `createTelemetryConfig` when both signals are disabled |
| `cmd/thv/app/otel.go` | Update setters/getters/unsetters for `*bool` |
| `cmd/thv/app/run_flags_test.go` | Update existing tests and add 4 new cases for tracing/metrics fallback |

`thv config otel set-tracing-enabled` and `set-metrics-enabled` now take effect when starting servers with `thv run`. Previously these settings were silently ignored.

The `*bool` pattern (nil = never set, false = explicitly disabled) is necessary because the CLI defaults for these flags are `true`, unlike `Insecure` and `EnablePrometheusMetricsPath` which default to `false`. A struct-level comment on `OpenTelemetryConfig` explains this distinction.


cmd/thv/app/otel.go
```diff
@@ -413,22 +413,23 @@ func getOtelMetricsEnabledCmdFunc(_ *cobra.Command, _ []string) error {
...
+	fmt.Printf("Current OpenTelemetry metrics enabled: %t\n", metricsEnabled)
 	return nil
 }
 
 func unsetOtelMetricsEnabledCmdFunc(_ *cobra.Command, _ []string) error {
 	configProvider := config.NewDefaultProvider()
 	cfg := configProvider.GetConfig()
 
-	if !cfg.OTEL.MetricsEnabled {
+	if cfg.OTEL.MetricsEnabled == nil || !*cfg.OTEL.MetricsEnabled {
```
jhrozek: I think there's a subtle issue with the `unset` guard here. It treats "not configured" (nil) and "explicitly set to false" the same way — so if someone runs `set-metrics-enabled false` and later tries to `unset` it, they get "already disabled" and the explicit `false` stays in the config file, still overriding the CLI default.

Should we check just `== nil` instead? That way `unset` always removes the entry. Same thing applies to `unsetOtelTracingEnabledCmdFunc` at line 473.

cmd/thv/app/run_flags.go
```diff
@@ -958,8 +970,8 @@ func getOidcFromFlags(cmd *cobra.Command) (string, string, string, string, strin
 // getTelemetryFromFlags extracts telemetry configuration from command flags
 func getTelemetryFromFlags(cmd *cobra.Command, config *cfg.Config, otelEndpoint string, otelSamplingRate float64,
 	otelEnvironmentVariables []string, otelInsecure bool, otelEnablePrometheusMetricsPath bool,
-	otelUseLegacyAttributes bool) (
-	string, float64, []string, bool, bool, bool) {
+	otelUseLegacyAttributes bool, otelTracingEnabled bool, otelMetricsEnabled bool) (
+	string, float64, []string, bool, bool, bool, bool, bool) {
```
jhrozek: This is now returning 8 positional values with 5 consecutive `bool`s — it's getting pretty easy to accidentally swap them at call sites. Would you be up for pulling these into a small struct in this PR, or would you rather do that as a follow-up?

cmd/thv/app/run_flags.go
```diff
@@ -708,6 +710,16 @@ func configureMiddlewareAndOptions(
 		oidcScopes := extractOIDCValues(oidcConfig)
 	finalOtelEndpoint, finalOtelSamplingRate, finalOtelEnvironmentVariables := extractTelemetryValues(telemetryConfig)
 
+	// Extract resolved tracing/metrics values from the middleware telemetry config.
+	// These must match what setupTelemetryConfiguration resolved (with global config
+	// fallbacks) rather than the raw runFlags values, which ignore global config.
+	finalTracingEnabled := runFlags.OtelTracingEnabled
```
jhrozek: Heads up — when both signals get disabled via global config, `createTelemetryConfig` returns nil, and the fallback here ends up using the raw `runFlags` values (which default to `true`). So the resolved "disabled" state gets lost. Right now it doesn't cause problems because `extractTelemetryValues(nil)` returns an empty endpoint and `MaybeMakeConfig` bails out too, but it feels fragile.

Maybe we could pass the resolved values from `getTelemetryFromFlags` to both consumers, or just default to `false` here when `telemetryConfig` is nil?

cmd/thv/app/run_flags.go
gkatz2: Let's do the struct refactor now. It's a tiny bit of scope creep, I suppose, but the resulting code will be clearer.

cmd/thv/app/otel.go
gkatz2: Good catch — that's a real bug from the `*bool` migration. Fixed: the guard now checks `== nil` only, so `unset` always clears the entry regardless of the current value. Also updated the messages to match: "is not configured" for the no-op case, "Successfully unset" for the success case.

cmd/thv/app/run_flags.go
gkatz2: Agreed, the implicit dependency on downstream nil checks was fragile. I went with your second suggestion — default to `false` when `telemetryConfig` is nil. Makes the intent obvious: no telemetry config means tracing and metrics are both disabled.

PR #4334, opened 2026-03-24
Preserve prompt message structure through vMCP relay

JAORMX:
## Summary

The vMCP relay was flattening prompt messages into a single `"[role] text\n"` string, losing individual message roles, content types (images, audio), and multi-message structure. AI clients receiving prompts through the relay got a degraded single-string representation instead of the structured messages the backend MCP server returned.

This changes `PromptGetResult.Messages` from `string` to `[]PromptMessage`, preserving each message's role and full content structure through the entire relay path (backend → session → client → handler factory → MCP response).

- Add `PromptMessage` type with `Role` and `Content` fields to `vmcp.types`
- Replace lossy `ConvertPromptMessages` (string concatenation) with `ConvertMCPPromptMessages` / `ToMCPPromptMessages` round-trip converters
- Update handler factory to relay structured messages instead of wrapping a flat string in a synthetic "assistant" message
- Update all tests to assert on structured message content

## Type of change

- [x] Bug fix

## Test plan

Prompt responses from virtual MCP servers now preserve the original message structure (roles, content types) from backend servers, instead of flattening them into a single text string.

- **This PR**: prompt message structure
- fix-resource-relay: resource text/blob distinction and per-item metadata
- fix-content-annotations: per-content annotations (audience, priority, lastModified)

### DIFF
--- a/pkg/vmcp/types.go
+++ b/pkg/vmcp/types.go
@@ -450,11 +450,22 @@ type ResourceReadResult struct {
 	Meta map[string]any
 }
 
+// PromptMessage represents a single message in a prompt response,
+// preserving the role and content structure from the backend.
+type PromptMessage struct {
+	// Role is the message role. The MCP spec defines "user" and "assistant";
+	// backends may also send other values which are relayed as-is.
+	Role string
+	// Content is the message content, supporting all MCP content types.
+	Content Content
+}
+
 // PromptGetResult wraps a prompt response with metadata.
 // This preserves both the prompt messages AND the _meta field from the backend MCP server.
 type PromptGetResult struct {
-	// Messages is the concatenated prompt text from all messages.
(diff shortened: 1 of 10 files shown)


pkg/vmcp/types.go
```diff
@@ -450,11 +450,22 @@ type ResourceReadResult struct {
 	Meta map[string]any
 }
 
+// PromptMessage represents a single message in a prompt response,
+// preserving the role and content structure from the backend.
+type PromptMessage struct {
+	// Role is the message role. The MCP spec defines "user" and "assistant";
+	// backends may also send other values which are relayed as-is.
+	Role string
```
yrobla, line 458: i was thinking about making role a type, but as you said "backends may also send other values which are relayed as-is"

PR #4335, opened 2026-03-24
Fix resource relay to preserve text/blob distinction

JAORMX:
## Summary

The vMCP relay was concatenating all resource contents into a single `[]byte` with one `MimeType`, losing the text/blob distinction, per-item URIs, and per-item MIME types when a resource returned multiple content items. This meant multi-content resources and blob resources were degraded through the relay.

This changes `ResourceReadResult` from `Contents []byte` + `MimeType string` to `Contents []ResourceContent`, preserving each item's type (text vs blob), URI, MIME type, and data independently.

- Add `ResourceContent` type with Text, Data, URI, MimeType fields to `vmcp.types`
- Replace lossy `ConcatenateResourceContents` with `ConvertMCPResourceContents` / `ToMCPResourceContents` round-trip converters
- Update handler factory to relay structured resource contents instead of reconstructing from concatenated bytes
- Remove the now-unused concatenation helper and its tests
- Update all tests to assert on structured resource content

## Type of change

- [x] Bug fix

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/types.go` | Add `ResourceContent` type, update `ResourceReadResult` |
| `pkg/vmcp/conversion/content.go` | Add `ConvertMCPResourceContents` / `ToMCPResourceContents`, remove `ConcatenateResourceContents` |
| `pkg/vmcp/conversion/conversion_test.go` | Add round-trip tests for resource contents |
| `pkg/vmcp/client/client.go` | Use structured conversion instead of concatenation |
| `pkg/vmcp/server/adapter/handler_factory.go` | Relay structured contents directly |
| `pkg/vmcp/server/adapter/handler_factory_test.go` | Update to use structured resource content |
| `pkg/vmcp/server/sessionmanager/session_manager.go` | Update resource content handling |
| `pkg/vmcp/session/internal/backend/mcp_session.go` | Use structured conversion |

Resource responses from virtual MCP servers now preserve individual content items with their text/blob types, URIs, and MIME types, instead of concatenating everything into a single byte slice.

- fix-prompt-messages: prompt message structure
- **This PR**: resource text/blob distinction and per-item metadata
- fix-content-annotations: per-content annotations (audience, priority, lastModified)

### DIFF
--- a/pkg/vmcp/conversion/content.go
+++ b/pkg/vmcp/conversion/content.go
@@ -130,7 +130,10 @@ func ConvertMCPResourceContents(contents []mcp.ResourceContents) []vmcp.Resource
 				Blob:     blobRes.Blob,
 			})
 		} else {
-			slog.Debug("Skipping unknown resource contents type", "type", fmt.Sprintf("%T", c))
+			// Warn rather than debug: an unrecognized resource type likely
+			// indicates a protocol change or bug, and silently dropping data
+			// should be visible to operators.
+			slog.Warn("Skipping unknown resource contents type", "type", fmt.Sprintf("%T", c))
 		}
 	}
 	return result
@@ -141,6 +144,8 @@ func ConvertMCPResourceContents(contents []mcp.ResourceContents) []vmcp.Resource
 func ToMCPResourceContents(contents []vmcp.ResourceContent) []mcp.ResourceContents {
 	result := make([]mcp.ResourceContents, 0, len(contents))
 	for _, c := range contents {
+		// Blob takes precedence: a non-empty Blob field means this is a blob resource.
+		// If both Text and Blob are set the Text field is ignored.
 		if c.Blob != "" {
 			result = append(result, mcp.BlobResourceContents{
 				URI:      c.URI,
--- a/pkg/vmcp/types.go
+++ b/pkg/vmcp/types.go
@@ -441,6 +441,7 @@ type ResourceContent struct {
 	// Text is the text content (non-empty for text resources).
 	Text string
 	// Blob is the base64-encoded binary content (non-empty for blob resources).
+	// Exactly one of Text or Blob should be set; Blob takes precedence in ToMCPResourceContents.
 	Blob string
 }
 


pkg/vmcp/types.go
```diff
@@ -431,17 +431,25 @@ type ToolCallResult struct {
...
+// preserving the text vs blob distinction from the MCP protocol.
+type ResourceContent struct {
+	// URI is the resource URI.
+	URI string
+	// MimeType is the content type of this resource item.
+	MimeType string
+	// Text is the text content (non-empty for text resources).
+	Text string
+	// Blob is the base64-encoded binary content (non-empty for blob resources).
+	Blob string
```
yrobla, line 445: The `ToMCPResourceContents` function uses `Blob != ""` as the discriminant to choose between text and blob types, but this contract isn't documented on the struct. If someone accidentally sets both `Text` and `Blob`, the content silently becomes a blob (text is dropped). A struct-level comment spelling out that exactly one of `Text`/`Blob` should be non-empty — and that `Blob` takes precedence — would make this invariant explicit and save future confusion.

```suggestion
	// Blob is the base64-encoded binary content (non-empty for blob resources).
	// Exactly one of Text or Blob should be set; Blob takes precedence in ToMCPResourceContents.
	Blob string
```

pkg/vmcp/conversion/content.go
```diff
@@ -114,36 +113,50 @@ func ToMCPContents(contents []vmcp.Content) []mcp.Content {
...
+	}
+	return result
+}
+
+// ToMCPResourceContents converts []vmcp.ResourceContent to []mcp.ResourceContents,
+// reconstructing the text vs blob distinction.
+func ToMCPResourceContents(contents []vmcp.ResourceContent) []mcp.ResourceContents {
+	result := make([]mcp.ResourceContents, 0, len(contents))
+	for _, c := range contents {
+		if c.Blob != "" {
```
yrobla, line 149: The discriminant here is implicit — the function uses `c.Blob != ""` to decide the content type, but there's no comment explaining the invariant or what happens when both `Text` and `Blob` are non-empty (Blob wins, Text is silently dropped). A short comment would help readers understand the intention:

```suggestion
		// Blob takes precedence: a non-empty Blob field means this is a blob resource.
		// If both Text and Blob are set the Text field is ignored.
		if c.Blob != "" {
```

pkg/vmcp/conversion/content.go
```diff
@@ -114,36 +113,50 @@ func ToMCPContents(contents []vmcp.Content) []mcp.Content {
...
+				Text:     textRes.Text,
+			})
+		} else if blobRes, ok := mcp.AsBlobResourceContents(c); ok {
+			result = append(result, vmcp.ResourceContent{
+				URI:      blobRes.URI,
+				MimeType: blobRes.MIMEType,
+				Blob:     blobRes.Blob,
+			})
+		} else {
+			slog.Debug("Skipping unknown resource contents type", "type", fmt.Sprintf("%T", c))
```
yrobla: why dropping warn to debug?

pkg/vmcp/types.go
JAORMX: Fixed in da9b5eff — added the invariant documentation to the `Blob` field.

pkg/vmcp/conversion/content.go
JAORMX: Fixed in da9b5eff — added the discriminant comment before the `if c.Blob != ""` check.

pkg/vmcp/conversion/content.go
JAORMX: Good catch — restored to `slog.Warn`. Dropping an unrecognized resource type silently at debug level would hide protocol changes or bugs from operators. Fixed in da9b5eff.

PR #4336, opened 2026-03-24
Preserve content-level annotations through vMCP relay

JAORMX:
## Summary

The vMCP relay was silently dropping per-content `Annotations` (audience, priority, lastModified) during the mcp→vmcp→mcp conversion. The MCP spec defines these annotations on every content type (text, image, audio, embedded resource, resource link), but the conversion layer was not carrying them through.

This adds `ContentAnnotations` to the vmcp domain model and threads annotation preservation through all content type conversions, ensuring annotations survive the full relay round-trip.

- Add `ContentAnnotations` type (Audience, Priority, LastModified) to `vmcp.types`
- Add `Annotations *ContentAnnotations` field to `vmcp.Content`
- Add `ConvertMCPAnnotations` / `ToMCPAnnotations` converters for content-level annotations
- Update `ConvertMCPContent` and `ToMCPContent` to preserve annotations on all content types (text, image, audio, embedded resource, resource link)
- Add comprehensive tests including round-trip verification for all content types

## Type of change

- [x] Bug fix

## Test plan

Content annotations (audience hints, priority, timestamps) from backend MCP servers are now preserved through virtual MCP server responses, enabling AI clients to respect content targeting and ordering hints.

- fix-prompt-messages: prompt message structure
- fix-resource-relay: resource text/blob distinction and per-item metadata
- **This PR**: per-content annotations (audience, priority, lastModified)

The `ContentAnnotations` type follows the existing Anti-Corruption Layer pattern — vmcp types use `[]string` for audience rather than `[]mcp.Role`, keeping the domain model decoupled from mcp-go.

### DIFF
--- a/pkg/vmcp/conversion/content.go
+++ b/pkg/vmcp/conversion/content.go
@@ -80,18 +80,30 @@ func ConvertMCPContent(content mcp.Content) vmcp.Content {
 		return vmcp.Content{Type: vmcp.ContentTypeText, Text: text.Text, Annotations: ConvertMCPAnnotations(text.Annotations)}
 	}
 	if img, ok := mcp.AsImageContent(content); ok {
-		return vmcp.Content{Type: vmcp.ContentTypeImage, Data: img.Data, MimeType: img.MIMEType, Annotations: ConvertMCPAnnotations(img.Annotations)}
+		return vmcp.Content{
+			Type: vmcp.ContentTypeImage, Data: img.Data,
+			MimeType: img.MIMEType, Annotations: ConvertMCPAnnotations(img.Annotations),
+		}
 	}
 	if audio, ok := mcp.AsAudioContent(content); ok {
-		return vmcp.Content{Type: vmcp.ContentTypeAudio, Data: audio.Data, MimeType: audio.MIMEType, Annotations: ConvertMCPAnnotations(audio.Annotations)}
+		return vmcp.Content{
+			Type: vmcp.ContentTypeAudio, Data: audio.Data,
+			MimeType: audio.MIMEType, Annotations: ConvertMCPAnnotations(audio.Annotations),
+		}
 	}
 	if res, ok := mcp.AsEmbeddedResource(content); ok {
 		ann := ConvertMCPAnnotations(res.Annotations)
 		if textRes, ok := mcp.AsTextResourceContents(res.Resource); ok {
-			return vmcp.Content{Type: vmcp.ContentTypeResource, Text: textRes.Text, URI: textRes.URI, MimeType: textRes.MIMEType, Annotations: ann}
+			return vmcp.Content{
+				Type: vmcp.ContentTypeResource, Text: textRes.Text,
+				URI: textRes.URI, MimeType: textRes.MIMEType, Annotations: ann,
+			}
 		}
 		if blobRes, ok := mcp.AsBlobResourceContents(res.Resource); ok {
-			return vmcp.Content{Type: vmcp.ContentTypeResource, Data: blobRes.Blob, URI: blobRes.URI, MimeType: blobRes.MIMEType, Annotations: ann}
+			return vmcp.Content{
+				Type: vmcp.ContentTypeResource, Data: blobRes.Blob,
+				URI: blobRes.URI, MimeType: blobRes.MIMEType, Annotations: ann,
+			}
 		}
 		slog.Debug("Embedded resource has unknown resource contents type", "type", fmt.Sprintf("%T", res.Resource))
 		return vmcp.Content{Type: vmcp.ContentTypeResource, Annotations: ann}
--- a/pkg/vmcp/conversion/content_test.go
+++ b/pkg/vmcp/conversion/content_test.go
@@ -13,7 +13,7 @@ import (
 	"github.com/stacklok/toolhive/pkg/vmcp"
 )
 
-func boolPtr(b bool) *bool     { return &b }
+func boolPtr(b bool) *bool          { return &b }
 func float64Ptr(f float64) *float64 { return &f }
 
 func TestConvertToolAnnotations(t *testing.T) {


pkg/vmcp/conversion/content.go
```diff
@@ -17,27 +17,94 @@ import (
...
+		for i, r := range ann.Audience {
+			audience[i] = string(r)
+		}
+	}
+	if len(audience) == 0 && ann.Priority == nil && ann.LastModified == "" {
+		return nil
+	}
+	return &vmcp.ContentAnnotations{
+		Audience:     audience,
+		Priority:     ann.Priority,
```
yrobla: **Pointer aliasing across the ACL boundary**

Both `ConvertMCPAnnotations` (here) and `ToMCPAnnotations` (line 59) copy `Priority` by pointer reference rather than by value. After conversion, `mcp.Annotations.Priority` and `vmcp.ContentAnnotations.Priority` point to the same `float64`, so a mutation through one leaks silently into the other. Consistent with the existing `*bool` pattern for `ToolAnnotations`, but `float64` is a more likely mutation target across a relay boundary.

Consider deep-copying:
```go
var priority *float64
if ann.Priority != nil {
    p := *ann.Priority
    priority = &p
}
```
Then use `priority` instead of `ann.Priority` in the return struct. Same fix applies to `ToMCPAnnotations` line 59.

pkg/vmcp/conversion/content.go
```diff
@@ -17,27 +17,94 @@ import (
...
+		return nil
+	}
+	var audience []mcp.Role
+	if len(ann.Audience) > 0 {
+		audience = make([]mcp.Role, len(ann.Audience))
+		for i, a := range ann.Audience {
+			audience[i] = mcp.Role(a)
+		}
+	}
+	return &mcp.Annotations{
```
yrobla, line 69: **`ToMCPAnnotations` is asymmetric with `ConvertMCPAnnotations`**

`ConvertMCPAnnotations` returns `nil` when all fields are zero-valued (lines 34–36). `ToMCPAnnotations` has no equivalent guard — an empty-but-non-nil `&vmcp.ContentAnnotations{}` produces a non-nil `&mcp.Annotations{}`, which serializes as `"annotations":{}` on the wire and creates a round-trip inconsistency.

Add the same zero-value guard before this `return`:
```go
if len(audience) == 0 && ann.Priority == nil && ann.LastModified == "" {
    return nil
}
```

pkg/vmcp/types.go
```diff
@@ -380,6 +380,19 @@ const (
 	ContentTypeLink ContentType = "resource_link"
 )
 
+// ContentAnnotations describes per-content metadata annotations.
+// These are the vmcp-domain equivalents of mcp.Annotations, following the
+// Anti-Corruption Layer pattern (vmcp types are decoupled from mcp-go).
+type ContentAnnotations struct {
+	// Audience describes who the content is intended for (e.g., "user", "assistant").
+	Audience []string `json:"audience,omitempty"`
+	// Priority is a hint for display ordering (0.0 = least important, 1.0 = most important).
```
yrobla: **Nit: Priority range constraint is undocumented**

The MCP spec defines `priority` as a closed interval [0.0, 1.0]. The current comment describes the semantics but doesn't state the constraint. Worth adding: "Values outside [0.0, 1.0] are invalid per the MCP specification."

Similarly, `Audience` gives `"user", "assistant"` as examples — worth noting these are the only valid values per the spec (the full set of `mcp.Role` constants), not just examples.

pkg/vmcp/conversion/content_test.go
```diff
@@ -283,3 +284,228 @@ func TestAnnotationsRoundTrip(t *testing.T) {
...
+				Priority:     float64Ptr(0.8),
+				LastModified: "2025-01-12T15:00:58Z",
+			},
+			want: &mcp.Annotations{
+				Audience:     []mcp.Role{mcp.RoleUser, mcp.RoleAssistant},
+				Priority:     float64Ptr(0.8),
+				LastModified: "2025-01-12T15:00:58Z",
+			},
+		},
+	}
```
yrobla, line 397: **`TestToMCPAnnotations` missing the empty non-nil case**

Only `nil` and fully populated are covered. There's no case for `&vmcp.ContentAnnotations{}` (all fields zero-valued), which is exactly where the asymmetry with `ConvertMCPAnnotations` surfaces. Adding this case documents current behavior and will catch any future nil-guard changes:

```go
{
    name:  "empty (non-nil) input",
    input: &vmcp.ContentAnnotations{},
    want:  &mcp.Annotations{}, // contrast: ConvertMCPAnnotations returns nil for empty input
},
```

pkg/vmcp/conversion/content_test.go
```diff
@@ -283,3 +284,228 @@ func TestAnnotationsRoundTrip(t *testing.T) {
...
+			}(),
+		},
+		{
+			name: "embedded resource",
+			content: func() mcp.Content {
+				er := mcp.NewEmbeddedResource(mcp.TextResourceContents{URI: "file://x", Text: "txt"})
+				er.Annotated = mcp.Annotated{Annotations: ann}
+				return er
+			}(),
+		},
```
yrobla, line 473: **Blob embedded resource path is untested**

`TestContentAnnotationsRoundTrip_AllTypes` covers `TextResourceContents` but not `BlobResourceContents`, which is a separate `if` branch in `ConvertMCPContent` (lines 97–104 of content.go). A regression there wouldn't be caught. Worth adding:

```go
{
    name: "blob embedded resource",
    content: func() mcp.Content {
        er := mcp.NewEmbeddedResource(mcp.BlobResourceContents{
            URI: "file://blob", Blob: "base64data", MIMEType: "application/octet-stream",
        })
        er.Annotated = mcp.Annotated{Annotations: ann}
        return er
    }(),
},
```

yrobla:
**Review priority summary** (following up on the inline comments above):

**Fix in this PR:**
- `ToMCPAnnotations` asymmetric nil-elision ([comment](https://github.com/stacklok/toolhive/pull/4336#pullrequestreview-3998598761)) — 3-line guard, prevents `"annotations":{}` on the wire
- Blob embedded resource test gap ([comment](https://github.com/stacklok/toolhive/pull/4336#pullrequestreview-3998598761)) — a few lines, covers an existing untested branch

**Fine as follow-ups:**
- `Priority *float64` pointer aliasing — consistent with the existing `*bool` pattern for `ToolAnnotations` in the same file; unlikely to cause issues in a relay context
- `TestToMCPAnnotations` empty non-nil case — documents the asymmetry above; can follow the nil-elision fix
- Doc comment nits on `Priority` range and `Audience` valid values

pkg/vmcp/conversion/content.go
JAORMX: Fixed in 0fa964ac — both `ConvertMCPAnnotations` and `ToMCPAnnotations` now deep-copy the `Priority` pointer.

pkg/vmcp/conversion/content.go
JAORMX: Fixed in 0fa964ac — added the zero-value guard in `ToMCPAnnotations` for symmetry. Empty non-nil input now returns nil.

pkg/vmcp/types.go
JAORMX: Fixed in 0fa964ac — documented Priority as `[0.0, 1.0]` per spec and clarified Audience values are the `mcp.Role` constants ("user", "assistant"), not just examples.

pkg/vmcp/conversion/content_test.go
JAORMX: Fixed in 0fa964ac — added the empty non-nil test case. With the nil guard fix, it now correctly returns nil.

pkg/vmcp/conversion/content_test.go
JAORMX: Fixed in 0fa964ac — added blob embedded resource test case to `TestContentAnnotationsRoundTrip_AllTypes`.

PR #4338, opened 2026-03-24
Wire git resolver into skill install flow

JAORMX:
## Summary

The gitresolver package was fully implemented but had no connection to the skill install path. Users couldn't actually install skills from git repositories — the resolver sat unused.

This change integrates the git resolver so `thv skill install` supports git-based skill sources, both via explicit `git://` prefix and through registry entries that specify a git package type.

- Add `WithGitResolver` option and `installFromGit` method to skillsvc to handle git-based installs
- Detect `git://` prefix in `Install()` before falling through to OCI reference check
- Extend `resolveFromRegistry` to handle `"git"` package types alongside OCI
- Add dev-mode SSRF bypass in `validateHost` for E2E testability
- Generate mock for `gitresolver.Resolver` interface
- Add unit tests for all git install paths (fresh, upgrade, no-op, errors)
- Add E2E tests with a local dumb-HTTP git server infrastructure

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/api/server.go` | Wire `gitresolver.NewResolver()` into server builder |
| `pkg/skills/gitresolver/reference.go` | Add `IsGitReference` helper, export `ParseReference` |
| `pkg/skills/gitresolver/resolver.go` | Add `Resolver` interface declaration for mock generation |
| `pkg/skills/gitresolver/mocks/mock_resolver.go` | Generated mock for `Resolver` interface |
| `pkg/skills/skillsvc/skillsvc.go` | Add `WithGitResolver`, `installFromGit`, git-aware registry resolution |
| `pkg/skills/skillsvc/skillsvc_test.go` | Unit tests for git install paths |
| `test/e2e/api_skills_git_test.go` | E2E tests with local dumb-HTTP git server |

Yes — users can now install skills from git repositories using `thv skill install git://` or via registry entries with a git package type.

## Large PR Justification

- Generated code that cannot be split: mock for `gitresolver.Resolver` (57 lines)
- Test code that depends on the feature: unit tests (350 lines) and E2E tests (471 lines) exercise the new install paths and cannot be merged independently
- Actual production code is ~270 lines across 4 files, well within the 400-line limit when excluding tests and generated code

- The `validateHost` function has a dev-mode bypass (`THV_DEV_SKIP_SSRF_CHECK`) to allow E2E tests to use localhost git servers. This is gated behind a build/env check and never active in production.
- The E2E test spins up a temporary bare git repo served over dumb HTTP — no external network access needed.

### DIFF
--- a/pkg/skills/skillsvc/skillsvc_test.go
+++ b/pkg/skills/skillsvc/skillsvc_test.go
@@ -38,6 +38,8 @@ import (
 	storemocks "github.com/stacklok/toolhive/pkg/storage/mocks"
 )
 
+const testCommitHash = "abcdef1234567890abcdef1234567890abcdef12"
+
 func makeLayerData(t *testing.T) []byte {
 	t.Helper()
 	files := []ociskills.FileEntry{
@@ -2129,7 +2131,7 @@ func TestInstallFromRegistry(t *testing.T) {
 func TestInstallFromGit(t *testing.T) {
 	t.Parallel()
 
-	commitHash := "abcdef1234567890abcdef1234567890abcdef12"
+	commitHash := testCommitHash
 
 	tests := []struct {
 		name        string
@@ -2360,7 +2362,7 @@ func TestInstallFromGit(t *testing.T) {
 func TestInstallFromGitGroupRegistrationRollback(t *testing.T) {
 	t.Parallel()
 
-	commitHash := "abcdef1234567890abcdef1234567890abcdef12"
+	commitHash := testCommitHash
 
 	ctrl := gomock.NewController(t)
 
@@ -2403,7 +2405,7 @@ func TestInstallFromGitGroupRegistrationRollback(t *testing.T) {
 func TestInstallFromRegistryGitFallback(t *testing.T) {
 	t.Parallel()
 
-	commitHash := "abcdef1234567890abcdef1234567890abcdef12"
+	commitHash := testCommitHash
 
 	tests := []struct {
 		name     string


pkg/skills/skillsvc/skillsvc.go
```diff
@@ -203,6 +215,16 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
...
 	opts.ProjectRoot = projectRoot
 
+	// Git references (git://host/owner/repo[@ref][#path]) are dispatched first;
+	// the prefix is unambiguous and cannot collide with OCI references.
+	if gitresolver.IsGitReference(opts.Name) {
+		result, err := s.installFromGit(ctx, opts, scope)
+		if err != nil {
+			return nil, err
+		}
+		return result, s.registerSkillInGroup(ctx, opts.Group, result.Skill.Metadata.Name)
```
aponcedeleonch: If `installFromGit` succeeds (files written + DB record created) but `registerSkillInGroup` fails, the skill is installed on disk and in the DB but not added to the group. This leaves the system in an inconsistent state — the skill exists but isn't in the expected group.

Note: this is the same pattern used by the OCI install path, so it's not a regression introduced by this PR — but worth tracking as a follow-up improvement for both paths.

pkg/skills/gitresolver/reference.go
```diff
@@ -191,3 +200,10 @@ func validateSkillPath(p string) error {
 	}
 	return nil
 }
+
+// isDevMode returns true when TOOLHIVE_DEV=true. In dev mode, SSRF checks
+// for localhost and private IPs are relaxed to enable E2E testing with local
+// git HTTP servers.
+func isDevMode() bool {
+	return strings.EqualFold(os.Getenv("TOOLHIVE_DEV"), "true")
+}
```
aponcedeleonch, line 209: Missing unit tests for `validateHost` covering both branches of `isDevMode()`. Since this is security-critical SSRF protection, it would be good to have tests using `t.Setenv("TOOLHIVE_DEV", "true")` and without it to verify the protection works correctly in production configuration.

Also missing: unit tests for `buildGitReferenceFromRegistryURL` edge cases (e.g., `http://` URL silent promotion to `https://`, already prefixed with `git://`).

pkg/skills/skillsvc/skillsvc.go
JAORMX: Good catch — addressed this directly rather than deferring. Added `installAndRegister` helper that wraps `registerSkillInGroup` with a best-effort DB rollback (`store.Delete`) on failure. This covers all 5 call sites (git direct, OCI direct, installByName, and both registry lookup paths).

If group registration fails, the DB record is removed so a retry starts fresh rather than leaving an orphaned skill record. Files on disk are left in place (a retry with the same content will detect them).

Also added `TestInstallFromGitGroupRegistrationRollback` to verify the rollback behavior on the git install path, and updated the existing "group registration error" test to assert the `Delete` call.

pkg/skills/gitresolver/reference.go
JAORMX: Added both:

**`TestParseGitReferenceDevMode`** in `reference_test.go` — covers `validateHost` with `TOOLHIVE_DEV=true`:
- 5 success cases: localhost, 127.0.0.1, 10.x, 192.168.x, and localhost-with-port all allowed in dev mode (returning `http://` URLs)
- 3 error cases: empty host, no repo path, single path component still rejected in dev mode
- Uses `t.Setenv` in a separate non-parallel test function (`//nolint:paralleltest`)

**`TestBuildGitReferenceFromRegistryURL`** in `skillsvc_test.go` — covers edge cases:
- `https://` → `git://` conversion (happy path)
- `http://` silent promotion to `git://`
- `git://` pass-through
- `https://` with nested path/ref/fragment
- Error cases: empty git ref, unsupported `ftp://` scheme, bare string without scheme, missing repo path

JAORMX:
Rebased onto main and addressed all review feedback in 31069856:

**Inconsistent state fix** — Rather than deferring, fixed it directly. Added `installAndRegister` helper that wraps all 5 `registerSkillInGroup` call sites with a best-effort `store.Delete` rollback. If group registration fails after a successful install, the DB record is removed so a retry starts fresh instead of leaving an orphaned skill.

**New tests:**
- `TestParseGitReferenceDevMode` — covers `validateHost` with `TOOLHIVE_DEV=true` (5 success cases for localhost/private IPs allowed, 3 error cases for non-SSRF validation still enforced)
- `TestBuildGitReferenceFromRegistryURL` — covers URL scheme conversion edge cases (https→git, http silent promotion, git passthrough, error cases)
- `TestInstallFromGitGroupRegistrationRollback` — verifies the DB rollback on group registration failure for the git install path
- Updated existing "group registration error" test to assert the `Delete` rollback call

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -984,3 +1239,25 @@ func (s *service) registerSkillInGroup(ctx context.Context, groupName string, sk
...
+	groupName string,
+	skillName string,
+	scope skills.Scope,
+	projectRoot string,
+) (*skills.InstallResult, error) {
+	if err := s.registerSkillInGroup(ctx, groupName, skillName); err != nil {
+		// Best-effort rollback: remove the DB record so retries start fresh.
+		// Files on disk are left in place; a fresh install will detect them
+		// and either overwrite (force) or return a conflict.
+		_ = s.store.Delete(ctx, skillName, scope, projectRoot)
```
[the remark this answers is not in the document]
JAORMX: Keeping the current approach (delete DB record, leave files on disk). Here's the reasoning:

The alternative — keeping the DB record without group membership — leaves the skill in a **stuck state**: it's "installed" according to the API but not in any group. A retry would hit an "already exists" conflict, forcing the user to manually uninstall before trying again. That's worse UX than the current behavior.

With the current approach:
- **Retries work cleanly** — no DB record means the next `install` starts fresh
- **Stale files are harmless** — a fresh install will overwrite them (or detect them with `--force`)
- The failure mode is transient, not sticky

I'll add a comment in the code noting this tradeoff for future readers.

PR #4342, opened 2026-03-24
Inject run config store into WorkloadFromContainerInfo

### LINKED ISSUE #4341: Flaky test: WorkloadFromContainerInfo hits real XDG filesystem
`TestRuntimeStatusManager_GetWorkload/successful_get_workload` intermittently fails with an `EOF` error in CI.

## Root cause

`WorkloadFromContainerInfo` calls `loadRunConfigFields`, which internally creates a real `state.LocalStore` via `state.LoadRunConfig` → `NewRunConfigStore(DefaultAppName)`. This hits the real XDG state directory (`~/.local/state/toolhive/runconfigs/`) instead of using dependency injection.

When tests run in parallel (especially with `-race` and `-coverpkg=./...`), a race condition can occur:
1. Another test or process creates/truncates the runconfig file
2. `store.Exists()` returns `true` (file exists)
3. `json.Decode` fails with `EOF` because the file is empty or partially written

This also affects `fileStatusManager` tests where mock `GetReader` returns a single-use `io.NopCloser(strings.NewReader(...))` that gets consumed on the first read — now that `WorkloadFromContainerInfo` also reads from the store, subsequent reads return EOF.

## Fix

- Add `state.Store` parameter to `WorkloadFromContainerInfo` and `loadRunConfigFields` for proper dependency injection
- Add `runConfigStore` field to `runtimeStatusManager` (matching the existing pattern in `fileStatusManager`)
- Update all callers to pass the injected store
- Fix test mocks to use `DoAndReturn` for fresh readers instead of single-use `Return(mockReader, ...)`

## Evidence

CI failure: https://github.com/stacklok/toolhive/actions/runs/23490251202/job/68356267110?pr=4335

```
status_test.go:122:
    Error: Received unexpected error: EOF
    Test: TestRuntimeStatusManager_GetWorkload/successful_get_workload
status_test.go:123:
    Error: Not equal:
        expected: "test-workload"
        actual  : ""
```

JAORMX:
## Summary

- `WorkloadFromContainerInfo` called `loadRunConfigFields` which internally created a real `state.LocalStore` hitting the XDG filesystem on every call. In CI, parallel tests could create or truncate runconfig files, causing intermittent `EOF` errors from `json.Decode` — making `TestRuntimeStatusManager_GetWorkload` flaky.
- Add a `state.Store` parameter to `WorkloadFromContainerInfo` and `loadRunConfigFields` so callers inject the store, matching the existing DI pattern in `fileStatusManager`. Add a `runConfigStore` field to `runtimeStatusManager` for parity.
- Fix `file_status_test.go` mock readers to use `DoAndReturn` for fresh readers on each call, preventing EOF when the store is read more than once per test.

Fixes #4341

## Type of change

- [x] Bug fix

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/workloads/types/types.go` | Add `state.Store` param to `WorkloadFromContainerInfo` and `loadRunConfigFields`; rewrite `loadRunConfigFields` to use injected store directly |
| `pkg/workloads/types/workload_test.go` | Pass existing `store` to updated `WorkloadFromContainerInfo` |
| `pkg/workloads/statuses/status.go` | Add `runConfigStore` field to `runtimeStatusManager`; update `NewStatusManagerFromRuntime` to return error; pass store at 2 call sites |
| `pkg/workloads/statuses/file_status.go` | Pass `f.runConfigStore` at 6 call sites; name receiver on `mergeHealthyWorkloadData` |
| `pkg/workloads/statuses/status_test.go` | Inject mock store into `runtimeStatusManager`; set up `Exists` expectations |
| `pkg/workloads/statuses/file_status_test.go` | Change `Return(mockReader, nil).AnyTimes()` to `DoAndReturn(...)` for fresh readers |

No

The `loadRunConfigFields` rewrite switches from `state.LoadRunConfig` (which creates a new store internally) to direct `store.Exists` + `store.GetReader` calls. The semantics are identical: returns empty config when not found, propagates errors otherwise — but no longer creates a store on every invocation.

### DIFF
--- a/pkg/workloads/statuses/status_test.go
+++ b/pkg/workloads/statuses/status_test.go
@@ -24,21 +24,22 @@ import (
 
 const testWorkloadName = "test-workload"
 
-//nolint:paralleltest // Cannot use t.Parallel() with t.Setenv() in Go 1.24+
 func TestNewStatusManagerFromRuntime(t *testing.T) {
+	t.Parallel()
+
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
 	mockRuntime := rtmocks.NewMockRuntime(ctrl)
-	manager, err := NewStatusManagerFromRuntime(mockRuntime)
+	mockStore := stateMocks.NewMockStore(ctrl)
+	manager := NewStatusManagerFromRuntime(mockRuntime, mockStore)
 
-	assert.NoError(t, err)
 	assert.NotNil(t, manager)
 	assert.IsType(t, &runtimeStatusManager{}, manager)
 
 	rsm := manager.(*runtimeStatusManager)
 	assert.Equal(t, mockRuntime, rsm.runtime)
-	assert.NotNil(t, rsm.runConfigStore)
+	assert.Equal(t, mockStore, rsm.runConfigStore)
 }
 
 func TestRuntimeStatusManager_CreateWorkloadStatus(t *testing.T) {
--- a/pkg/workloads/statuses/status.go
+++ b/pkg/workloads/statuses/status.go
@@ -48,15 +48,11 @@ type StatusManager interface {
 }
 
 // NewStatusManagerFromRuntime creates a new instance of StatusManager from an existing runtime.
-func NewStatusManagerFromRuntime(runtime rt.Runtime) (StatusManager, error) {
-	runConfigStore, err := state.NewRunConfigStore(state.DefaultAppName)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create run config store: %w", err)
-	}
+func NewStatusManagerFromRuntime(runtime rt.Runtime, runConfigStore state.Store) StatusManager {
 	return &runtimeStatusManager{
 		runtime:        runtime,
 		runConfigStore: runConfigStore,
-	}, nil
+	}
 }
 
 // NewStatusManager creates a new status manager instance using the appropriate implementation
@@ -70,7 +66,11 @@ func NewStatusManager(runtime rt.Runtime) (StatusManager, error) {
 // This allows for dependency injection of environment variable access for testing.
 func NewStatusManagerWithEnv(runtime rt.Runtime, envReader env.Reader) (StatusManager, error) {
 	if rt.IsKubernetesRuntimeWithEnv(envReader) {
-		return NewStatusManagerFromRuntime(runtime)
+		runConfigStore, err := state.NewRunConfigStore(state.DefaultAppName)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create run config store: %w", err)
+		}
+		return NewStatusManagerFromRuntime(runtime, runConfigStore), nil
 	}
 	return NewFileStatusManager(runtime)
 }


pkg/workloads/types/types.go
```diff
@@ -25,30 +22,33 @@ type minimalRunConfig struct {
...
+		return &minimalRunConfig{}, nil
+	}
 
-	// Return the runconfig
-	return runConfig, nil
+	reader, err := store.GetReader(ctx, name)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = reader.Close() }()
```
yrobla: **Blocker**: `reader.Close()` error is silently discarded with `_ =`, but the established project pattern (see `state/runconfig.go:93-95`, `statuses/file_status.go:99-101`) is to log it. A file-backed reader can return flush/OS errors on `Close()` that are worth observing.

```suggestion
	defer func() {
		if err := reader.Close(); err != nil {
			slog.Warn("failed to close run config reader", "workload", name, "error", err)
		}
	}()
```

pkg/workloads/types/types.go
```diff
@@ -25,30 +22,33 @@ type minimalRunConfig struct {
...
 	}
+	if !exists {
+		return &minimalRunConfig{}, nil
+	}
 
-	// Return the runconfig
-	return runConfig, nil
+	reader, err := store.GetReader(ctx, name)
+	if err != nil {
+		return nil, err
```
yrobla: **Blocker**: TOCTOU between `store.Exists` (line 27) and `store.GetReader` here. If a workload is deleted between the two calls (e.g. a concurrent `thv delete`), `GetReader` returns an error — but the old code via `state.LoadRunConfig` handled this case gracefully by returning an empty config.

The PR description says "semantics are identical", but that only holds in the absence of concurrent writes. The fix is to treat a not-found error from `GetReader` the same as `!exists`:

```suggestion
		if state.IsNotFound(err) {
			return &minimalRunConfig{}, nil
		}
		return nil, err
```
(adjust the not-found check to match the actual sentinel/error type the store returns)

pkg/workloads/types/types.go
```diff
@@ -25,30 +22,33 @@ type minimalRunConfig struct {
...
-		return &config, nil
-	})
+// using the provided store. Returns empty struct if runconfig doesn't exist.
+func loadRunConfigFields(ctx context.Context, store state.Store, name string) (*minimalRunConfig, error) {
+	exists, err := store.Exists(ctx, name)
 	if err != nil {
-		if errors.Is(err, werr.ErrRunConfigNotFound) {
-			return &minimalRunConfig{}, nil
-		}
 		return nil, err
```
yrobla: **Nit** (follow-up OK): both `store.Exists` and `store.GetReader` errors are returned without wrapping the workload name, making it hard to tell which workload triggered a `ListWorkloads` failure. Consider:

```suggestion
		return nil, fmt.Errorf("failed to check run config for workload %q: %w", name, err)
```
Same pattern applies to the `GetReader` error a few lines below.

pkg/workloads/statuses/status_test.go
```diff
@@ -27,13 +28,15 @@ func TestNewStatusManagerFromRuntime(t *testing.T) {
 	defer ctrl.Finish()
 
 	mockRuntime := rtmocks.NewMockRuntime(ctrl)
-	manager := NewStatusManagerFromRuntime(mockRuntime)
+	manager, err := NewStatusManagerFromRuntime(mockRuntime)
```
yrobla: **Nit** (follow-up OK): `NewStatusManagerFromRuntime` internally calls `state.NewRunConfigStore` → `os.MkdirAll` on the real XDG state directory. Every other test in this file injects a `stateMocks.NewMockStore`. This one is inconsistent and could flake in sandboxed CI environments.

The simplest fix is `t.Setenv("XDG_STATE_HOME", t.TempDir())` before the call, or restructure `NewStatusManagerFromRuntime` to accept the store as a parameter (following the DI pattern this PR establishes).

JAORMX: Good catch — fixed in e7ec9e1b. Added `t.Setenv("XDG_STATE_HOME", t.TempDir())` so the test doesn't touch the real host state directory.

yrobla (approved):

Issue #4192, opened 2026-03-17
Add scoped secret store to isolate system-managed tokens from user secrets


amirejaz:
## Background

ToolHive's secret management uses a flat keyspace — all secrets share the same namespace in the backing store. Users can list, read, and delete any secret via \`thv secret list/get/delete\`.

Several system-managed tokens are currently stored in this flat namespace and are visible to users:

| Key pattern | Source |
|---|---|
| \`REGISTRY_OAUTH_<hex>\` | \`thv registry login\` refresh tokens |
| \`OAUTH_CLIENT_SECRET_<workload>\` | Remote workload OAuth client secrets |
| \`BEARER_TOKEN_<workload>\` | Remote workload bearer tokens |
| \`OAUTH_REFRESH_TOKEN_<workload>\` | Remote workload OAuth refresh tokens |

These are internal system credentials. Exposing them via \`thv secret list\` is confusing and risky — users can accidentally overwrite or delete tokens that workloads depend on.

## Proposed Solution

Introduce a **\`ScopedProvider\`** and **\`UserProvider\`** wrapper layer in \`pkg/secrets/\`. System callers get a scoped provider that prefixes every key with \`__thv_<scope>_\`. User-facing callers get a \`UserProvider\` wrapper that hides and blocks access to any key starting with \`__thv_\`.

Full design details: [THV-0056 RFC](https://github.com/stacklok/toolhive-rfcs/pull/56)

### Key prefix format

```
__thv_registry_REGISTRY_OAUTH_a1b2c3d4      ← registry OAuth refresh tokens
__thv_workloads_BEARER_TOKEN_github-app     ← remote workload auth tokens
__thv_workloads_OAUTH_CLIENT_SECRET_foo     ← remote workload OAuth client secrets
__thv_auth_access_token                     ← reserved for future enterprise login
<user-key>                                  ← user-managed secrets (unchanged)
```

### Scope constants

```go
const (
    SystemKeyPrefix = "__thv_"
    ScopeRegistry   = "registry"   // pkg/registry/auth/
    ScopeWorkloads  = "workloads"  // pkg/auth/secrets/
    ScopeAuth       = "auth"       // future: enterprise login tokens
)
```

### New types in \`pkg/secrets/scoped.go\`

**\`ScopedProvider\`** — for system callers:
- \`GetSecret/SetSecret/DeleteSecret\` transparently prefix the key with \`__thv_<scope>_\`
[...]

PR #4343, opened 2026-03-24
Wire all secret callers to scoped and user providers

amirejaz:
## Summary

- All secret provider call sites were previously using the bare `CreateSecretProvider`, giving every caller access to the full flat keyspace. This PR wires each caller to the appropriate wrapper introduced in Phase 1 (#4229), so secrets are isolated by design rather than by convention.
- System callers (workload auth tokens, registry OAuth credentials, build auth files) now use `CreateScopedSecretProvider`, writing and reading under the `__thv_<scope>_` prefix. These keys are invisible to user-facing secret commands.
- User-facing callers (`thv secret` commands, REST API, MCP tool server, header secrets, build-env-from-secrets) now use `CreateUserSecretProvider`, which blocks access to any `__thv_*` reserved key.
- `RunConfig.WithSecrets` and `ValidateSecrets` now take separate `systemProvider` and `userProvider` arguments so auth-token resolution and `--secret` flag resolution use the correct scope independently.

This is **Phase 3** of the scoped secret store (part of #4192), tracking issue #4227. It must be released together with Phase 4 (#4244 — migration infrastructure).

> **Release note**: Both this PR and the migration PR (#4244) must be merged before releasing. Neither is a breaking change on its own when no release occurs between the two merges.

## Type of change

## Test plan

- [x] `go test ./pkg/runner/... ./pkg/auth/secrets/... ./pkg/workloads/... ./pkg/mcp/server/...` — all pass
- [x] `golangci-lint run` — 0 issues

The `WithSecrets` signature change (one provider → two providers) is the most structurally significant change. The split is:
- `systemProvider` — used for `RemoteAuthConfig.ClientSecret`, `RemoteAuthConfig.BearerToken` (both written by the workload auth subsystem under `ScopeWorkloads`)
- `userProvider` — used for `c.Secrets` (`--secret` flags) and `HeaderForward.AddHeadersFromSecret` (both user-specified key names)

`config_buildauthfile.go` gets its own `getScopedWorkloadsSecretsManager()` helper rather than reusing the user-facing `getSecretsManager()`, since build auth files are system-owned (stored by `thv config buildauthfile set`, not by `thv secret set`).


cmd/thv/app/config_buildauthfile.go
```diff
@@ -298,3 +299,21 @@ func unsetBuildAuthFileCmdFunc(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+// getScopedWorkloadsSecretsManager creates a scoped secrets provider for build auth file operations.
+// Build auth file secrets are system-owned and stored under the workloads scope namespace.
+func getScopedWorkloadsSecretsManager() (secrets.Provider, error) {
```
reyortiz3: why not reuse the `auth/secretrs/secrets.go::GetSecretsManager()` ?

cmd/thv/app/config_buildenv.go
```diff
@@ -124,7 +124,7 @@ func validateSecretExists(ctx context.Context, secretName string) error {
 		return fmt.Errorf("failed to get secrets provider type: %w", err)
 	}
 
-	manager, err := secrets.CreateSecretProvider(providerType)
+	manager, err := secrets.CreateUserSecretProvider(providerType)
```
reyortiz3: (naming): why not call these `userSecretProvider`? I found `manager` a bit confusing / not clear

cmd/thv/app/config_buildenv.go
```diff
@@ -124,7 +124,7 @@ func validateSecretExists(ctx context.Context, secretName string) error {
 		return fmt.Errorf("failed to get secrets provider type: %w", err)
 	}
```
reyortiz3: I see this block duplicated in several places, can this be something that could be re-used from the `pkg/secrets`?

pkg/secrets/scoped.go
```diff
@@ -52,8 +53,26 @@ func NewScopedProvider(inner Provider, scope string) Provider {
...
+// If the scoped key is not found, it falls back to the bare (pre-migration) key.
+// This makes the provider safe to use before or during secret scope migration:
+// once migration completes and bare keys are deleted, the fallback is a no-op.
 func (s *ScopedProvider) GetSecret(ctx context.Context, name string) (string, error) {
-	return s.inner.GetSecret(ctx, scopedKey(s.scope, name))
+	val, err := s.inner.GetSecret(ctx, scopedKey(s.scope, name))
+	if err == nil {
+		return val, nil
+	}
+	if IsNotFoundError(err) {
```
reyortiz3, line 78: (outside the scope of this PR): I see `IsNotFoundError` is doing string .contains() on the error message, something to consider in the future to change to do something like `error.Is() `

amirejaz: Good catch — agreed this is fragile. Tracking it as a follow-up: replace the string-contains check in `IsNotFoundError` with a proper sentinel error (e.g. `var ErrSecretNotFound = errors.New("secret not found")`) so callers can use `errors.Is()`. Out of scope for this stack but worth a dedicated clean-up PR.

amirejaz:
Closing in favor of a new PR from the correct head branch (phase3-wire-callers). The new PR will show only wire callers changes without including migration commits.

PR #4345, opened 2026-03-24
Fix concurrent SetSecret calls silently clobbering each other

### LINKED ISSUE #4339: Concurrent SetSecret calls silently clobber each other (TOCTOU)
## Bug description
*This bug was investigated and validated with the assistance of Opus 4.6/Gpt 5.4. All reproduction scripts, patches, and test results were executed and verified on a real environment.*

`thv secret set` silently loses secrets when another process writes concurrently. Once persisted, a secret is safe (all future snapshots include it), but if OAuth-enabled servers are already running, their token refreshes prevent new secrets from being added reliably. A contributing factor is high frequency oauth token refreshes.

`SetSecret`, `DeleteSecret`, and `Cleanup` in `EncryptedManager` contain a TOCTOU race condition. `NewEncryptedManager` reads and decrypts the `secrets_encrypted` file into an in-memory `syncmap.Map` at construction time, outside any lock. When `SetSecret` later acquires `WithFileLock`, it updates the **stale** in-memory map and writes the entire map back to disk, silently overwriting any changes written by another process between the initial read and the locked write.

## Steps to reproduce
```bash
(while true; do
# Background writer simulating OAuth token refresh
  for k in 1 2 3 4 5; do
    echo "v$k" | thv secret set "_churn_$k" &>/dev/null
  done
  sleep 0.05
done) &
BG=$!

lost=0; total=50
for i in $(seq 1 $total); do
  thv secret delete my-secret &>/dev/null || true
  echo "value" | thv secret set my-secret &>/dev/null
  sleep 0.1
  thv secret get my-secret &>/dev/null || lost=$((lost + 1))
done

kill $BG 2>/dev/null; wait $BG 2>/dev/null
echo "Lost: $lost / $total"
```

**8% loss** on macOS with the background writer. 0% without it. Anecdotally it's closer to 50% after resuming from sleep / restarting toolhive.

## Expected behavior
`thv secret set` should permanently retain a secret.

## Actual behavior
- `thv run <mcp server>` fails with:
  ```
  Error: failed to validate workload parameters: error processing secrets:
  failed to get secrets: secret not found: <secret name>
  ```
- `thv secret get <name>` returns `secret not found` immediately after a successful
`thv secret set`.
- `thv secret list` shows hundreds of OAuth refresh tokens but no `<secret name>` entries.
- Secrets work right after setting, then disappear after the next OAuth token refresh
overwrites the file from its stale snapshot.

## Environment (if relevant)
- **OS/version:** macOS 26.3.1
- **ToolHive version:** v0.11.3 (v0.12.5 tested less extensively)
- **ToolHive Desktop:** v0.23.0 (v0.25.0 tested less extensively)
- **Secret provider:** `encrypted`
- **Config:** `provider_type: encrypted` in `config.yaml`

## Additional context

The bug is in `pkg/secrets/encrypted.go`. Three mutating methods are affected:

| Method         | Bug                                                                                     |
| -------------- | --------------------------------------------------------------------------------------- |
| `SetSecret`    | Reads file at construction, writes under lock — stale map overwrites concurrent changes |
| `DeleteSecret` | Same pattern; existence check uses stale map                                            |
| `Cleanup`      | Clears the in-memory map and writes under lock without reloading — wipes unrelated keys |

The file lock (`pkg/fileutils/lock.go`, using `gofrs/flock`) only serializes the
write. Each CLI invocation creates a fresh `EncryptedManager` that snapshots the file
at construction time. By the time the lock is acquired, the snapshot may be stale.

### Workaround

Our current workaround for internally developed MCP server instructions is to write-then-verify with retries to reduce the probability (0/200 losses vs 17/200 without the workaround):

```bash
#!/bin/bash
# Usage: ./thv-secret-set.sh my-secret
set -euo pipefail
NAME="$1"
read -rsp "Enter secret value: " VALUE; echo

for attempt in $(seq 1 10); do
  printf '%s' "$VALUE" | thv secret set "$NAME" 2>/dev/null
  sleep 1
  if thv secret get "$NAME" &>/dev/null; then
    echo "✓ $NAME persisted (attempt $attempt)"
    exit 0
  fi
  echo "  attempt $attempt: $NAME not found, retrying..." >&2
done

echo "ERROR: $NAME failed to persist after 10 attempts" >&2
exit 1
```

JAORMX:
## Summary

- `EncryptedManager` had a TOCTOU race: it loaded the secrets file at construction time and wrote that stale snapshot back under the file lock, silently overwriting changes from other processes. In practice, OAuth token refreshes from long-running proxy processes would clobber secrets set by concurrent `thv secret set` CLI invocations (and vice versa), causing ~8% secret loss under contention.
- Mutations (`SetSecret`, `DeleteSecret`, `Cleanup`) now re-read the file from disk inside the critical section before applying changes, eliminating the stale-snapshot problem.
- Added a per-path in-process `sync.Mutex` to `WithFileLock` because `flock(2)` does not provide mutual exclusion between different file descriptors within the same process.

Fixes #4339

## Type of change

- [x] Bug fix

## Test plan

- [x] Unit tests (`task test`) — existing `TestEncryptedManager_Concurrency` now passes reliably (previously flaky); ran 50 iterations with zero failures
- [x] Manual testing — verified `go vet` passes on both changed packages

## Changes

| File | Change |
|------|--------|
| `pkg/secrets/encrypted.go` | `SetSecret`/`DeleteSecret`/`Cleanup` now read-modify-write inside the lock instead of using a stale in-memory snapshot; extracted `readFileSecrets`/`writeFileSecrets` helpers; simplified `NewEncryptedManager` to reuse `readFileSecrets` |
| `pkg/fileutils/lock.go` | Added per-path `sync.Mutex` registry (`processLocks`) so `WithFileLock` serializes both across processes (flock) and within the same process (mutex) |

Yes — `thv secret set` will no longer silently lose secrets when other processes (e.g. OAuth token refreshes) write to the secrets file concurrently. Users who previously needed retry-based workarounds (#4339) should no longer experience secret loss.

- The in-memory `syncmap.Map` cache is now updated with targeted `Store`/`Delete` calls (not a full cache replacement) to avoid a window where concurrent `GetSecret` reads see an empty cache. This means the cache may not reflect keys added by other processes, but that's acceptable: CLI one-shots create a fresh manager per invocation, and long-running proxies only need their own tokens.
- `readFileSecrets` handles empty files and missing files gracefully, returning an empty map rather than erroring.
- The `DeleteSecret` existence check moved inside the lock and now checks the on-disk state, not the potentially-stale in-memory cache.

### DIFF
--- a/pkg/fileutils/lock.go
+++ b/pkg/fileutils/lock.go
@@ -26,11 +26,18 @@ const (
 // WithFileLock opens a new file descriptor, concurrent goroutines can all acquire
 // the flock simultaneously. This in-process mutex ensures serialization within a
 // single process, while the flock continues to protect cross-process access.
+//
+// This map is never pruned; callers should ensure the number of distinct
+// paths remains bounded (e.g. one secrets file per workload).
 var processLocks sync.Map
 
 // getProcessLock returns the in-process mutex for the given path,
 // creating one if it does not already exist.
 func getProcessLock(path string) *sync.Mutex {
+	// Fast path: return the existing mutex without allocating.
+	if val, ok := processLocks.Load(path); ok {
+		return val.(*sync.Mutex)
+	}
 	val, _ := processLocks.LoadOrStore(path, &sync.Mutex{})
 	return val.(*sync.Mutex)
 }
--- a/pkg/secrets/encrypted.go
+++ b/pkg/secrets/encrypted.go
@@ -33,7 +33,11 @@ type fileStructure struct {
 	Secrets map[string]string `json:"secrets"`
 }
 
-// GetSecret retrieves a secret from the secret store.
+// GetSecret retrieves a secret from the in-memory cache.
+// It does not re-read the file; secrets written by other processes after
+// construction may not be visible. This is intentional: CLI invocations
+// create a fresh manager per call, and long-running proxies only need
+// their own tokens.
 func (e *EncryptedManager) GetSecret(_ context.Context, name string) (string, error) {
 	if name == "" {
 		return "", errors.New("secret name cannot be empty")
@@ -63,6 +67,9 @@ func (e *EncryptedManager) SetSecret(_ context.Context, name, value string) erro
 		if err := e.writeFileSecrets(secrets); err != nil {
 			return err
 		}
+		// Update the in-memory cache after the disk write. There is a brief
+		// window where a concurrent GetSecret may return a stale value; this
+		// is acceptable because the file is the authoritative source of truth.
 		e.secrets.Store(name, value)
 		return nil
 	})
@@ -82,12 +89,18 @@ func (e *EncryptedManager) DeleteSecret(_ context.Context, name string) error {
 			return err
 		}
 		if _, ok := secrets[name]; !ok {
+			// Evict stale cache entry: another process may have already
+			// deleted this key from disk while it remained in our cache.
+			e.secrets.Delete(name)
 			return fmt.Errorf("cannot delete non-existent secret: %s", name)
 		}
 		delete(secrets, name)
 		if err := e.writeFileSecrets(secrets); err != nil {
 			return err
 		}
+		// Update the in-memory cache after the disk write. There is a brief
+		// window where a concurrent GetSecret may return a stale value; this
+		// is acceptable because the file is the authoritative source of truth.
 		e.secrets.Delete(name)
 		return nil
 	})


pkg/secrets/encrypted.go
```diff
@@ -65,15 +74,22 @@ func (e *EncryptedManager) DeleteSecret(_ context.Context, name string) error {
...
-
 	return fileutils.WithFileLock(e.filePath, func() error {
+		// Re-read the file inside the lock so the existence check
+		// reflects the current on-disk state.
+		secrets, err := e.readFileSecrets()
+		if err != nil {
+			return err
+		}
+		if _, ok := secrets[name]; !ok {
+			return fmt.Errorf("cannot delete non-existent secret: %s", name)
```
amirejaz, line 95: **Stale cache entry is not evicted on the not-found path.**

Because `GetSecret` reads only from `e.secrets` (in-memory cache) and `DeleteSecret` now reads from disk, the two can diverge:

1. Manager loads — `e.secrets` has `"mykey"`
2. Another process deletes `"mykey"` from disk
3. Caller calls `GetSecret("mykey")` → **succeeds** (stale cache)
4. Caller calls `DeleteSecret("mykey")` → `readFileSecrets` finds it gone → **"cannot delete non-existent secret"**

The caller just saw the key but cannot delete it. Worse, the stale entry is never cleaned up — every subsequent `GetSecret` keeps returning a value that no longer exists on disk.

The fix is a one-liner: evict the stale cache entry before returning the error:

```suggestion
		if _, ok := secrets[name]; !ok {
			e.secrets.Delete(name) // evict stale cache entry discovered to be gone on disk
			return fmt.Errorf("cannot delete non-existent secret: %s", name)
		}
```

pkg/secrets/encrypted.go
```diff
@@ -65,15 +74,22 @@ func (e *EncryptedManager) DeleteSecret(_ context.Context, name string) error {
...
+		}
+		delete(secrets, name)
+		if err := e.writeFileSecrets(secrets); err != nil {
+			return err
+		}
 		e.secrets.Delete(name)
```
amirejaz, line 104: **Cache update lags behind the disk write.** Between `writeFileSecrets` succeeding (line 88–90) and `e.secrets.Delete(name)` executing here, a concurrent `GetSecret` in the same process can still return the deleted value. The same window exists in `SetSecret` in the opposite direction (key written to disk but not yet in cache).

Both orderings are a deliberate trade-off (disk is the source of truth; the cache is a best-effort read optimisation). Worth a brief comment here and in `SetSecret` so future contributors don't try to "fix" the ordering:

```go
// Update the in-memory cache after the disk write. There is a brief window
// where a concurrent GetSecret may return a stale value; this is acceptable
// because the file is the authoritative source of truth.
e.secrets.Delete(name)
```

pkg/secrets/encrypted.go
```diff
@@ -111,15 +132,41 @@ func (*EncryptedManager) Capabilities() ProviderCapabilities {
...
-	// Convert syncmap.Map to map[string]string for JSON marshaling
-	secretsMap := make(map[string]string)
-	e.secrets.Range(func(key, value interface{}) bool {
-		secretsMap[key.(string)] = value.(string)
-		return true
-	})
+// readFileSecrets reads and decrypts the secrets file, returning the current
+// on-disk secrets. Returns an empty map for an empty or non-existent file.
+// Must be called while holding the file lock.
+func (e *EncryptedManager) readFileSecrets() (map[string]string, error) {
```
amirejaz, line 151: **`GetSecret` and `ListSecrets` still read from the stale in-memory cache** (neither method is changed by this PR). A long-running proxy whose peer process wrote a new secret after construction will never see it via `GetSecret`.

This is acknowledged in the PR description as acceptable, but there is no code-level comment capturing the design decision. Without one, a future contributor might "fix" this by re-reading from disk on every `GetSecret` call — which would bypass the cache and add per-call decrypt overhead.

Suggestion: add a comment on the struct or `GetSecret` documenting the intentional split:
```go
// GetSecret and ListSecrets read from the in-memory cache only and do not
// re-read the file. They may not reflect secrets written by other processes
// after construction. This is intentional: CLI invocations create a fresh
// manager per call, and long-running proxies only need their own tokens.
```

pkg/secrets/encrypted.go
```diff
@@ -142,59 +189,38 @@ func NewEncryptedManager(filePath string, key []byte) (Provider, error) {
...
-			return nil, fmt.Errorf("failed to decode secrets file: %w", err)
-		}
-
-		// Store each secret in the syncmap.Map
-		for key, value := range contents.Secrets {
-			manager.secrets.Store(key, value)
+	// Load the initial snapshot into the in-memory cache.
+	secrets, err := manager.readFileSecrets()
+	if err != nil {
+		if strings.Contains(err.Error(), "unable to decrypt") {
```
amirejaz, line 226: **`strings.Contains` on error messages is fragile.** If the wrapper message in `readFileSecrets` ever changes (e.g. `"unable to decrypt"` → `"failed to decrypt"`), the user-facing guidance is silently lost with no compile-time or test-time signal.

The original code matched `"message authentication failed"` from the underlying crypto library, which was even more brittle. This PR improves it by matching our own wrapper, but it's still string-matching.

The right fix is a sentinel error in the `aes` package:
```go
// pkg/secrets/aes
var ErrDecryptionFailed = errors.New("decryption failed")
```
Then the check becomes `errors.Is(err, aes.ErrDecryptionFailed)`, which is refactor-safe. Tracking this as a follow-up would be worthwhile.

pkg/fileutils/lock.go
```diff
@@ -19,9 +20,31 @@ const (
...
 
-// WithFileLock executes fn while holding an OS-level advisory file lock on path + ".lock".
-// It uses a 1-second timeout with 100ms retry interval.
+// processLocks provides per-path in-process mutual exclusion.
+// flock(2) does NOT provide mutual exclusion between different file descriptors
+// within the same process — only between different processes. Since each call to
+// WithFileLock opens a new file descriptor, concurrent goroutines can all acquire
+// the flock simultaneously. This in-process mutex ensures serialization within a
+// single process, while the flock continues to protect cross-process access.
+var processLocks sync.Map
```
amirejaz, line 32: **`processLocks` grows unbounded.** Entries are stored forever — one `*sync.Mutex` per distinct path, never evicted. For CLI one-shots and a fixed secrets file path this is harmless, but for any long-running process that calls `WithFileLock` over many distinct paths it is a minor memory leak.

At minimum document the trade-off in the comment:
> This map is never purged; callers should ensure the number of distinct paths remains bounded.

pkg/fileutils/lock.go
```diff
@@ -19,9 +20,31 @@ const (
...
+
+// getProcessLock returns the in-process mutex for the given path,
+// creating one if it does not already exist.
+func getProcessLock(path string) *sync.Mutex {
+	val, _ := processLocks.LoadOrStore(path, &sync.Mutex{})
```
amirejaz, line 41: **`LoadOrStore` allocates a new `*sync.Mutex{}` on every call**, even when the key already exists (the surplus mutex is immediately discarded). In the steady state (paths already registered) this is a GC allocation per `WithFileLock` call. A `Load`-first pattern avoids it:

```suggestion
func getProcessLock(path string) *sync.Mutex {
	if val, ok := processLocks.Load(path); ok {
		return val.(*sync.Mutex)
	}
	val, _ := processLocks.LoadOrStore(path, &sync.Mutex{})
	return val.(*sync.Mutex)
}
```

pkg/secrets/encrypted.go
JAORMX: Good catch — fixed in beb11980. `DeleteSecret` now evicts the stale cache entry before returning the error.

pkg/secrets/encrypted.go
JAORMX: Added the comment in beb11980 — both `SetSecret` and `DeleteSecret` now document the brief cache-lag window and why it's acceptable.

pkg/secrets/encrypted.go
JAORMX: Added the design comment on `GetSecret` in beb11980 — documents the intentional cache-only read and the reasoning (CLI one-shots vs long-running proxies).

pkg/secrets/encrypted.go
JAORMX: Agreed — a sentinel error in the `aes` package would be the right fix. Tracking as a follow-up to keep this PR focused on the TOCTOU race.

pkg/fileutils/lock.go
JAORMX: Documented the trade-off in beb11980 — the comment now notes the map is never pruned and callers should keep the set of distinct paths bounded.

pkg/fileutils/lock.go
JAORMX: Good call — fixed in beb11980. `getProcessLock` now uses a `Load`-first fast path to avoid allocating a new mutex on every call in steady state.

PR #4346, opened 2026-03-24
Add crash-safe migration for legacy system secret keys

amirejaz:
## Summary

- Existing users may have registry tokens and workload auth secrets stored under bare keys (`BEARER_TOKEN_`, `OAUTH_CLIENT_SECRET_`, `REGISTRY_OAUTH_`, etc.) that pre-date the scoped provider wrappers. Without migration, those keys would be unreachable after Phase 3 wires callers to use `ScopeWorkloads`/`ScopeRegistry` providers.
- Adds `MigrateSystemKeys` and `DiscoverMigrations` in `pkg/secrets/migration.go`: discovers bare system keys by known prefix and renames them into the `__thv_<scope>_` namespace using write-before-delete ordering for crash safety.
- Adds `CheckAndPerformSecretScopeMigration` in `pkg/migration/secret_scope.go`: guards migration behind a `SecretScopeMigration` config flag (same pattern as existing migrations) and is triggered from `cmd/thv/main.go` alongside the other startup migrations.

This is **Phase 4** of the scoped secret store (part of #4192), tracking issue #4226. It is stacked on Phase 3 (#4343) and must be released at the same time.

> **Release note**: This PR and the caller-wiring PR (#4343) must be merged before releasing. Neither is a breaking change on its own when no release occurs between the two merges.

## Type of change

## Test plan

- [x] `go test ./pkg/secrets/... ./pkg/migration/...` — all pass, including `TestMigrateSystemKeys` and `TestDiscoverMigrations`
- [x] `golangci-lint run` — 0 issues

Key design properties of the migration:

- **Write-before-delete**: new key is written before the old is deleted, so a crash mid-migration leaves the secret reachable under the new name on retry.
- **Idempotent**: if the old key is absent (already migrated or never existed), `GetSecret` returns a not-found error and the entry is silently skipped.
- **Discovery-based**: `DiscoverMigrations` lists all secrets and matches against `SystemKeyPrefixMappings`; no static workload registry is required.
- **One-shot**: the `SecretScopeMigration` config flag prevents re-running on every startup after migration completes.
- **Raw provider**: migration uses `CreateSecretProvider` (bare, no wrapper) so it can enumerate and rename both old bare keys and already-scoped `__thv_*` keys without restriction.

### DIFF
--- a/pkg/secrets/migration.go
+++ b/pkg/secrets/migration.go
@@ -38,8 +38,8 @@ var SystemKeyPrefixMappings = []struct {
 // Write-before-delete ordering ensures that a crash between the two operations
 // leaves the secret reachable under the new key. Keys that do not exist in
 // provider are silently skipped, making the function safe to retry.
-func MigrateSystemKeys(ctx context.Context, provider Provider, migrations []KeyMigration) error {
-	for _, m := range migrations {
+func MigrateSystemKeys(ctx context.Context, provider Provider, keyMigrations []KeyMigration) error {
+	for _, m := range keyMigrations {
 		// If the scoped key already exists (e.g. from a partial prior run),
 		// skip the write and just clean up the bare key.
 		_, err := provider.GetSecret(ctx, m.NewKey)


pkg/config/config.go
```diff
@@ -39,6 +39,7 @@ type Config struct {
 	DefaultGroupMigration        bool                                `yaml:"default_group_migration,omitempty"`
 	TelemetryConfigMigration     bool                                `yaml:"telemetry_config_migration,omitempty"`
 	MiddlewareTelemetryMigration bool                                `yaml:"middleware_telemetry_migration,omitempty"`
+	SecretScopeMigration         bool                                `yaml:"secret_scope_migration,omitempty"`
```
reyortiz3, line 42: maybe consider in the future something like `Migrations map[string]bool`?

pkg/secrets/migration.go
```diff
@@ -0,0 +1,85 @@
...
+	all, err := provider.ListSecrets(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("migration discovery: listing secrets: %w", err)
+	}
+
+	var migrations []KeyMigration
+	for _, desc := range all {
+		key := desc.Key
+		// Skip already-migrated keys.
+		if strings.HasPrefix(key, SystemKeyPrefix) {
```
reyortiz3: nit: `isSystemKey(key)`?

pkg/secrets/migration.go
```diff
@@ -0,0 +1,85 @@
...
+			continue
+		}
+		for _, mapping := range SystemKeyPrefixMappings {
+			if strings.HasPrefix(key, mapping.Prefix) {
```
reyortiz3, line 94: same as before, do we have worry about case sensitivity in all places it does `strings.HasPrefix(key, ... `?

pkg/secrets/migration.go
```diff
@@ -0,0 +1,85 @@
...
+	{"BUILD_AUTH_FILE_", ScopeWorkloads},
+	{"REGISTRY_OAUTH_", ScopeRegistry},
+}
+
+// MigrateSystemKeys renames keys from OldKey to NewKey in provider.
+// Write-before-delete ordering ensures that a crash between the two operations
+// leaves the secret reachable under the new key. Keys that do not exist in
+// provider are silently skipped, making the function safe to retry.
+func MigrateSystemKeys(ctx context.Context, provider Provider, migrations []KeyMigration) error {
+	for _, m := range migrations {
```
reyortiz3: nit: `keyMigrations` instead of `migrations`?

pkg/config/config.go
amirejaz: Good idea — a `Migrations map[string]bool` would let us add future one-shot migrations without touching the struct shape each time. Flagging as a follow-up refactor once this stack lands; don't want to mix a config-schema change into this PR.

pkg/secrets/migration.go
amirejaz: Same answer as the earlier comment on `UserProvider`: case sensitivity is not a concern here. The `__thv_` prefix is only ever written by our own code (via `getScopedKey`), never constructed from user input, so a key like `__THV_*` can never appear in the store.

pkg/secrets/migration.go
```diff
@@ -0,0 +1,100 @@
...
+		// If the scoped key already exists (e.g. from a partial prior run),
+		// skip the write and just clean up the bare key.
+		if _, err := provider.GetSecret(ctx, m.NewKey); err == nil {
+			slog.Debug("migration: scoped key already exists, skipping write",
+				"old_key", m.OldKey, "new_key", m.NewKey)
+			if delErr := provider.DeleteSecret(ctx, m.OldKey); delErr != nil && !IsNotFoundError(delErr) {
+				return fmt.Errorf("migration: deleting stale bare key %q: %w", m.OldKey, delErr)
+			}
+			continue
+		}
```
aponcedeleonch, line 53: **Bug: transient backend errors bypass the idempotency check and can overwrite already-migrated values.**

The current code only checks for `err == nil` (key exists). If `GetSecret` for the new key fails with a **non-not-found error** (e.g., backend temporarily unavailable, permission error), the code falls through to the write path — reads the old value, and overwrites the new key via `SetSecret`.

This defeats the write-before-delete idempotency guarantee: if the new key was already written in a prior partial run but the check fails transiently, the old (potentially stale) value clobbers the already-migrated one.

```suggestion
		_, err := provider.GetSecret(ctx, m.NewKey)
		if err == nil {
			slog.Debug("migration: scoped key already exists, skipping write",
				"old_key", m.OldKey, "new_key", m.NewKey)
			if delErr := provider.DeleteSecret(ctx, m.OldKey); delErr != nil && !IsNotFoundError(delErr) {
				return fmt.Errorf("migration: deleting stale bare key %q: %w", m.OldKey, delErr)
			}
			continue
		}
		if !IsNotFoundError(err) {
			return fmt.Errorf("migration: checking scoped key %q: %w", m.NewKey, err)
		}
```

This way, only a confirmed not-found proceeds to the write path. Any other error aborts safely — the config flag won't be set, so migration retries on next startup.

PR #4348, opened 2026-03-24
Add RegisterHandlers and wire embedded AS routes on vMCP mux

jhrozek:
## Summary

- The vMCP server needs to host both RFC 9728 protected resource metadata and embedded authorization server (AS) OAuth/OIDC endpoints on the same HTTP mux. The previous `/.well-known/` catch-all registration would conflict with AS discovery routes.
- Replace `/.well-known/` catch-all with explicit `/.well-known/oauth-protected-resource` path registrations so the AS can own its own `/.well-known/` endpoints without conflicts.
- Add `RegisterHandlers` method on `EmbeddedAuthServer` that registers `/.well-known/openid-configuration`, `/.well-known/oauth-authorization-server`, `/.well-known/jwks.json`, and `/oauth/` routes.
- Add `AuthServer` field to `vmcpserver.Config` with nil-guarded route registration in the mux setup.

Fixes #4141

## Type of change

- [x] New feature

## Test plan

- [x] Linting (`task lint-fix`)
- [x] Unit tests (`task test`)

## Changes

| File | Change |
|------|--------|
| `pkg/authserver/runner/embeddedauthserver.go` | Add `RegisterHandlers(mux)` method |
| `pkg/vmcp/server/server.go` | Add `AuthServer` field, explicit `.well-known/` paths, Mode B route registration |

## Special notes for reviewers

This is part of a stacked PR series for the embedded auth server feature. This PR adds only the server-side wiring points. The construction of the `EmbeddedAuthServer` and config loading come in later PRs:
- scaffolding-3: upstream_inject strategy types + cross-cutting validation
- scaffolding-4: operator reconciler condition for AuthServerConfig
- scaffolding-5: converter + CLI auth server config loading

Generated with [Claude Code](https://claude.com/claude-code)

PR #4349, opened 2026-03-24
Add upstream_inject strategy type and auth server validation

### LINKED ISSUE #4142: Phase 3: Startup validation (V-01..V-07) — upstream_inject types and validateAuthServerIntegration
## Description

Add the `upstream_inject` auth strategy types to `pkg/vmcp/auth/types/types.go` and implement the `validateAuthServerIntegration` function in `pkg/vmcp/config/validator.go`, covering all seven validation rules (V-01 through V-07) that guard the embedded auth server configuration. Write comprehensive table-driven unit tests for all rules and extend the existing `validateAuthServerConfig` unit tests in the operator CRD types package.

## Context

RFC-0053 adds an optional embedded OAuth/OIDC authorization server to vMCP. Phase 1 (#4140) established the foundational config model types — `AuthServerConfig`, the `AuthServer` field on `Config`, and the `AuthServerConfigRef` CRD field. Phase 3 builds on that skeleton to add the validation layer that prevents misconfigured auth wiring from reaching runtime. Without these startup checks, a user who mistakenly configures `upstream_inject` without an auth server, or who sets a mismatched issuer, would only discover the problem at token validation time rather than at startup.

The seven validation rules (V-01 through V-07) enforce the contracts between the embedded AS, the incoming OIDC config, and the outgoing backend auth strategies. V-01/V-02/V-06 guard the `upstream_inject` strategy against invalid provider references; V-03 warns about a common misconfiguration of `token_exchange` when the AS is the incoming auth provider; V-04/V-05/V-07 guard issuer and audience consistency between the AS and the incoming OIDC config.

**Parent epic**: #4120 — vMCP: add embedded authorization server
**RFC document**: `docs/proposals/THV-0053-vmcp-embedded-authserver.md`
**Dependencies**: #4140 (Phase 1 — foundation types must exist before validation can reference them)
**Blocks**: Phase 4 (operator reconciler), which requires the validation rules to be available for the Kubernetes reconciliation loop

## Acceptance Criteria

- [ ] `pkg/vmcp/auth/types/types.go` exports `StrategyTypeUpstreamInject = "upstream_inject"` constant alongside the existing strategy type constants
- [ ] `pkg/vmcp/auth/types/types.go` exports `UpstreamInjectConfig` struct with a `ProviderName string` field and `+kubebuilder:object:generate=true` annotation, with both `json:"providerName"` and `yaml:"providerName"` tags
- [ ] `pkg/vmcp/auth/types/types.go` `BackendAuthStrategy` struct has a new `UpstreamInject *UpstreamInjectConfig` field with `json:"upstreamInject,omitempty" yaml:"upstreamInject,omitempty"` tags
- [ ] `pkg/vmcp/config/validator.go` calls `validateAuthServerIntegration(cfg)` as the last step in `DefaultValidator.Validate()` (before the final error join)
- [ ] V-01: `validateAuthServerIntegration` returns an error when any backend strategy has `Type == "upstream_inject"` and `cfg.AuthServer == nil`
- [ ] V-02: `validateAuthServerIntegration` returns an error when `upstream_inject` `ProviderName` is non-empty but does not match any `Name` in `cfg.AuthServer.RunConfig.Upstreams`
- [ ] V-03: `validateAuthServerIntegration` emits `slog.Warn` (not an error) when any backend strategy has `Type == "token_exchange"` and the incoming auth type is `"oidc"` with the issuer equal to the AS issuer
- [ ] V-04: `validateAuthServerIntegration` returns an error when `cfg.AuthServer.RunConfig.Issuer` does not exactly match `cfg.IncomingAuth.OIDC.Issuer` (when both are non-nil/non-empty)
- [ ] V-05: `validateAuthServerIntegration` returns an error when `runner.NewEmbeddedAuthServer` internal config validation fails (i.e., the AS `RunConfig` itself is invalid)
- [ ] V-06: `validateAuthServerIntegration` returns an error when any backend strategy has `Type == "upstream_inject"` and `UpstreamInject.ProviderName` is empty
- [ ] V-07: `validateAuthServerIntegration` returns an error when `cfg.IncomingAuth.OIDC.Audience` is non-empty and is not present in `cfg.AuthServer.RunConfig.AllowedAudiences` (YAML path check only)
- [ ] `validateBackendAuthStrategy` in `validator.go` accepts `"upstream_inject"` as a valid strategy type (added to `validTypes`)
- [ ] `validateBackendAuthStrategy` returns an error when `Type == "upstream_inject"` and `UpstreamInject` config is nil (V-06 pre-check)
- [ ] Helper functions `collectAllBackendStrategies`, `hasUpstreamProvider`, and `containsString` are added to `validator.go`
- [ ] `pkg/vmcp/config/validator_test.go` contains a table-driven test function `TestValidateAuthServerIntegration` with one test case per rule: V-01 through V-07, plus a Mode A pass-through case (nil `AuthServer`) and a valid Mode B config case
- [ ] `cmd/thv-operator/api/v1alpha1/virtualmcpserver_types_test.go` contains unit tests for `validateAuthServerConfig` covering empty `Name` rejection and nil-pointer safety (nil `AuthServerConfigRef` must not panic or error)
- [ ] All existing tests continue to pass (Mode A configs with nil `AuthServer` must not be affected)
- [ ] All new Go files and new code additions include the SPDX license header
- [ ] `task test` passes
- [ ] `task lint` passes (or `task lint-fix` resolves all issues)

Work in three discrete steps: (1) add the `upstream_inject` types to `pkg/vmcp/auth/types/types.go`; (2) extend `validateBackendAuthStrategy` and add `validateAuthServerIntegration` with helpers to `validator.go`; (3) write the unit tests. Step 1 is a prerequisite for Step 2 because the validator references `authtypes.StrategyTypeUpstreamInject` and `authtypes.UpstreamInjectConfig`.

For V-05, use `runner.NewEmbeddedAuthServer` to validate the AS config internally. Because this function requires a `context.Context`, pass `context.Background()`. `NewEmbeddedAuthServer` validates the `RunConfig` synchronously during construction; the returned server can be discarded (or closed immediately with `defer as.Close()`) — the validation side-effect is what matters. This approach avoids duplicating the AS's own validation logic.

For V-03, use `slog.Warn` with a structured message; do not return an error. The warning should include the backend name and the matching issuer so operators can identify the misconfiguration.

The `containsString` helper is functionally identical to the existing private `contains(slice []string, item string) bool` in `validator.go`. Either alias it or add a new export — the architecture note in `research.md` says the existing helper "can be aliased as `containsString`". Since Go does not support type aliases for functions, add `containsString` as a new named function calling the existing `contains` helper, or simply inline the logic.

For the Ginkgo-vs-`testing` question: the architecture.md constraint says "if Phase 3 adds Ginkgo `DescribeTable` tests, a Ginkgo suite bootstrap must also be added to the `config` package. Alternatively, use the existing `testing` table-driven style to avoid the bootstrap." Use the existing `testing` package table-driven style to stay consistent with the current `validator_test.go` and avoid a Ginkgo bootstrap dependency. The table structure `[]struct{name, cfg, wantErr, wantWarn, errContains string}` cleanly covers all nine cases.

### Patterns and Frameworks

- **Table-driven tests (`testing` package)**: Match the pattern in `pkg/vmcp/config/validator_test.go` — `[]struct{name, cfg, wantErr bool, errMsg string}` with `t.Run` + `t.Parallel()`. Do not introduce Ginkgo to this package.
- **`slog.Warn` for non-fatal validation**: Follow the logging guideline in CLAUDE.md — `slog.Warn` for non-fatal issues (deprecations, fallback behavior). Use structured fields for context.
- **Helper composition**: `collectAllBackendStrategies` walks `cfg.OutgoingAuth.Default` and all entries in `cfg.OutgoingAuth.Backends`, returning `[]*authtypes.BackendAuthStrategy`. This is a simple slice-building function — keep it pure and testable.
- **SPDX headers**: Every Go file must start with `// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.` and `// SPDX-License-Identifier: Apache-2.0`. Use `task license-fix` to add missing headers automatically.
- **`require.NoError(t, err)` for test assertions**: Use `require` from `github.com/stretchr/testify/require` rather than `t.Fatal` for error checking in tests.
- **`+kubebuilder:object:generate=true` on `UpstreamInjectConfig`**: Required so `controller-gen` generates `DeepCopyInto`/`DeepCopy` for the new struct. After adding this annotation, run `task gen` from repo root to regenerate `pkg/vmcp/auth/types/zz_generated.deepcopy.go` (if it exists) or verify the generator picks up the new struct.

### Code Pointers

- `pkg/vmcp/auth/types/types.go` — Add `StrategyTypeUpstreamInject` constant here (alongside `StrategyTypeUnauthenticated`, `StrategyTypeHeaderInjection`, `StrategyTypeTokenExchange`). Add `UpstreamInjectConfig` struct and `UpstreamInject *UpstreamInjectConfig` field to `BackendAuthStrategy`. This is the primary type file for Phase 3 additions.
- `pkg/vmcp/config/validator.go` (lines 182–223) — `validateBackendAuthStrategy`: add `authtypes.StrategyTypeUpstreamInject` to `validTypes` slice (line 187–193) and add a `case authtypes.StrategyTypeUpstreamInject:` block in the switch statement handling V-01 (nil AS check) and V-06 (empty provider name check).
- `pkg/vmcp/config/validator.go` (lines 31–81) — `DefaultValidator.Validate()`: add `validateAuthServerIntegration(cfg)` call at the end, before the final error aggregation.
- `pkg/vmcp/config/validator.go` (lines 441–457) — Existing `contains` and `containsStrategy` helpers at the bottom of the file. Add the new helper functions `collectAllBackendStrategies`, `hasUpstreamProvider`, and `containsString` in this section.
- `pkg/vmcp/config/validator_test.go` — Add `TestValidateAuthServerIntegration` using the existing table-driven `testing` style. Import `authserver "github.com/stacklok/toolhive/pkg/authserver"` to construct test `RunConfig` values.
- `pkg/authserver/config.go` (lines 33–74) — `authserver.RunConfig`: the type wrapped by `AuthServerConfig.RunConfig`. Key validation-relevant fields: `Issuer string`, `AllowedAudiences []string`, `Upstreams []UpstreamRunConfig` (each with `Name string`). Use these to construct test fixtures.
- `pkg/authserver/runner/embeddedauthserver.go` — `NewEmbeddedAuthServer(ctx, *authserver.RunConfig) (*EmbeddedAuthServer, error)` — called by V-05 to perform internal AS config validation.
- `cmd/thv-operator/api/v1alpha1/virtualmcpserver_types_test.go` — Add `TestValidateAuthServerConfig` function here, following the existing `TestValidateEmbeddingServer` pattern (table-driven, parallel). Tests should cover: nil `AuthServerConfigRef` (no error), non-nil ref with non-empty `Name` (no error), non-nil ref with empty `Name` (error containing `"spec.authServerConfigRef.name is required"`).
- `cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go` (lines 362–398) — `VirtualMCPServer.Validate()`: the Phase 1 task (#4140) adds a call to `r.validateAuthServerConfig()` here. The Phase 3 tests exercise this path; verify the call is present before writing tests.

### Component Interfaces

New constant and types in `pkg/vmcp/auth/types/types.go`:

```go
// StrategyTypeUpstreamInject identifies the upstream inject strategy.
// This strategy injects an upstream provider token (acquired by the embedded AS)
// as a Bearer token when authenticating to the backend.
// The actual outgoing strategy implementation is deferred to a follow-up RFC;
// only the constant, config struct, and validation are added here.
const StrategyTypeUpstreamInject = "upstream_inject"

// UpstreamInjectConfig configures the upstream inject auth strategy.
// Used when Type = "upstream_inject".
// +kubebuilder:object:generate=true
type UpstreamInjectConfig struct {
    // ProviderName is the upstream provider whose token is injected as Bearer.
    // Must match a provider name in authServer.runConfig.upstreams.
    ProviderName string `json:"providerName" yaml:"providerName"`
}
```

Addition to `BackendAuthStrategy` in `pkg/vmcp/auth/types/types.go`:

```go
// UpstreamInject contains configuration for upstream inject auth strategy.
// Used when Type = "upstream_inject".
// +optional
UpstreamInject *UpstreamInjectConfig `json:"upstreamInject,omitempty" yaml:"upstreamInject,omitempty"`
```

New function signatures in `pkg/vmcp/config/validator.go`:

```go
// validateAuthServerIntegration validates the relationship between AuthServer config,
// IncomingAuth config, and backend auth strategies.
// Called from DefaultValidator.Validate() when cfg.AuthServer is non-nil (Mode B checks)
// or when upstream_inject strategies are present (V-01 requires AS).
// Rules V-01 through V-07 are checked here.
func (v *DefaultValidator) validateAuthServerIntegration(cfg *Config) error

// collectAllBackendStrategies returns a flat slice of all configured backend auth strategies,
// including the default strategy and all per-backend strategies.
func collectAllBackendStrategies(auth *OutgoingAuthConfig) []*authtypes.BackendAuthStrategy

// hasUpstreamProvider reports whether any upstream in cfg.AuthServer.RunConfig.Upstreams
// has Name equal to providerName.
func hasUpstreamProvider(cfg *Config, providerName string) bool

// containsString reports whether slice contains item.
func containsString(slice []string, item string) bool
```

Validation rule summary (for implementation reference):

| Rule | Condition | Severity | Short description |
|------|-----------|----------|-------------------|
| V-01 | `upstream_inject` used + `AuthServer == nil` | error | AS required for upstream_inject |
| V-02 | `upstream_inject` `ProviderName` not in AS upstreams | error | Unknown provider name |
| V-03 | `token_exchange` + AS is incoming auth issuer | `slog.Warn` | Likely misconfiguration |
| V-04 | `AuthServer.RunConfig.Issuer` != `IncomingAuth.OIDC.Issuer` | error | Issuer mismatch |
| V-05 | `NewEmbeddedAuthServer` internal validation fails | error | AS config invalid |
| V-06 | `upstream_inject` `ProviderName` is empty | error | Missing provider name |
| V-07 | `IncomingAuth.OIDC.Audience` not in `AllowedAudiences` | error | Audience not permitted |

### Testing Strategy

**Unit Tests — `TestValidateAuthServerIntegration` in `pkg/vmcp/config/validator_test.go`**

Use the existing table-driven `testing` style. Each test case sets up a `*Config` and checks `v.Validate(cfg) != nil` (or `== nil` for passing cases). Construct test `authserver.RunConfig` values inline in the table — they do not need to be fully valid for all rules (e.g., V-01 just needs `AuthServer == nil`).

- [ ] V-01 case: `OutgoingAuth.Backends["test"]` has `Type = "upstream_inject"`, `AuthServer = nil` → error containing `"upstream_inject"` and `"authServer"`
- [ ] V-02 case: `AuthServer` non-nil with upstream named `"github"`, backend has `upstream_inject` with `ProviderName = "unknown"` → error containing `"unknown"` and `"not found"`
- [ ] V-03 case: `AuthServer.RunConfig.Issuer = "https://as.example.com"`, `IncomingAuth.OIDC.Issuer = "https://as.example.com"`, backend has `token_exchange` → no error (warning only, captured via `slog` test hook or verified by absence of error)
- [ ] V-04 case: `AuthServer.RunConfig.Issuer = "https://as.example.com"`, `IncomingAuth.OIDC.Issuer = "https://external-idp.example.com"` → error containing `"issuer"` and `"mismatch"`
- [ ] V-05 case: `AuthServer.RunConfig` is structurally invalid (e.g., empty `Issuer`) → error surfacing the AS internal validation failure
- [ ] V-06 case: backend has `Type = "upstream_inject"`, `UpstreamInject.ProviderName = ""` → error containing `"providerName"` and `"required"`
- [ ] V-07 case: `AuthServer.RunConfig.AllowedAudiences = ["https://vmcp.example.com"]`, `IncomingAuth.OIDC.Audience = "https://other.example.com"` → error containing `"audience"` and `"allowedAudiences"`
- [ ] Mode A pass-through: `AuthServer = nil`, no `upstream_inject` backends → no error
- [ ] Valid Mode B config: all fields consistent (matching issuer, audience in `AllowedAudiences`, valid upstream name) → no error

**Unit Tests — `TestValidateAuthServerConfig` in `cmd/thv-operator/api/v1alpha1/virtualmcpserver_types_test.go`**

Follow the `TestValidateEmbeddingServer` pattern (table-driven, `t.Parallel()`, `require.NoError`/`require.Error` assertions):

- [ ] Nil `AuthServerConfigRef`: `VirtualMCPServer.Validate()` returns nil (no regression, Mode A)
- [ ] Non-nil `AuthServerConfigRef` with non-empty `Name`: `Validate()` returns nil
- [ ] Non-nil `AuthServerConfigRef` with empty `Name`: `Validate()` returns error containing `"spec.authServerConfigRef.name is required"`

**Edge Cases**

- [ ] `OutgoingAuth.Default` strategy is `upstream_inject` (not just per-backend) — `collectAllBackendStrategies` must include the default strategy in its output
- [ ] `OutgoingAuth` is nil — `collectAllBackendStrategies` must return empty slice without panicking
- [ ] `cfg.IncomingAuth` is not OIDC type (e.g., anonymous) — V-04 and V-07 must skip gracefully when `cfg.IncomingAuth.OIDC == nil`
- [ ] Multiple backends with mixed strategies — only `upstream_inject` backends trigger V-01/V-02/V-06
- [ ] Exact string match for V-04 and V-07 (no URL normalization — trailing slash differences cause failures)

## Out of Scope

- `upstream_inject` outgoing strategy implementation (deferred to a follow-up RFC — the constant and config struct are added, but the actual middleware that injects the token is not implemented here)
- Operator reconciler cross-resource validation (`AuthServerConfigValid` condition surfacing) — Phase 4
- CRD-to-config converter changes in `converter.go` — Phase 4
- E2E tests for Mode A/Mode B behavior — Phase 4
- HTTP handler registration and AS route mounting — Phase 2
- Documentation updates to `docs/arch/` — Phase 4
- Hot-reload of AS configuration
- Multi-upstream IDP support beyond what `UpstreamRunConfig.Name` already provides in `authserver.RunConfig`

## References

- RFC-0053 design document: `docs/proposals/THV-0053-vmcp-embedded-authserver.md`
- Parent epic: https://github.com/stacklok/toolhive/issues/4120
- Phase 1 (foundation, upstream dependency): https://github.com/stacklok/toolhive/issues/4140
- RFC-0052 (multi-upstream IDP, required for full E2E): https://github.com/stacklok/toolhive/issues/3924
- Auth strategy types (primary edit target): `pkg/vmcp/auth/types/types.go`
- Config validator (primary edit target): `pkg/vmcp/config/validator.go`
- Auth server `RunConfig` (test fixture type): `pkg/authserver/config.go`
- `EmbeddedAuthServer` (used for V-05 validation): `pkg/authserver/runner/embeddedauthserver.go`
- Validator tests (pattern to follow): `pkg/vmcp/config/validator_test.go`
- Operator CRD types tests (pattern to follow): `cmd/thv-operator/api/v1alpha1/virtualmcpserver_types_test.go`
- `validateEmbeddingServer` (pattern for `validateAuthServerConfig` tests): `cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go`

### LINKED ISSUE #4144: Phase 1: Add core types and sentinel for upstream_inject strategy (RFC-0054)
## Description

Add the foundational type definitions needed by all subsequent phases of RFC-0054. This phase introduces the `StrategyTypeUpstreamInject` constant, the `UpstreamInjectConfig` struct, the `ErrUpstreamTokenNotFound` sentinel error, and two new fields on existing structs (`UpstreamInject` on `BackendAuthStrategy`, `SubjectProviderName` on `TokenExchangeConfig`). The generated deep-copy file is also regenerated to reflect the new struct. No runtime behavior changes — all new fields are optional and zero-valued by default, so existing configurations are unaffected.

## Context

This is Phase 1 of the RFC-0054 epic (#3925), which implements the `upstream_inject` outgoing auth strategy for vMCP. The strategy reads upstream IDP access tokens from `identity.UpstreamTokens` (populated by RFC-0052's auth middleware) and injects them as `Authorization: Bearer` headers on outgoing backend requests.

Phase 1 is the root task: it adds the shared type definitions that Phases 2, 3, and 4 all depend on. Phases 2 (strategy implementations), 3 (startup validation), and 4 (CRD and converter) cannot start until these types are merged. Phase 4 in particular can be parallelized with Phases 2 and 3 immediately after Phase 1 lands, since it only depends on the types package — not on RFC-0052 or RFC-0053.

**Dependencies**: None (root task)
**Blocks**: TASK-002 (Phase 2: Strategy implementations), TASK-003 (Phase 3: Startup validation), TASK-004 (Phase 4: CRD and converter)

## Acceptance Criteria

- [ ] `StrategyTypeUpstreamInject = "upstream_inject"` constant is present in `pkg/vmcp/auth/types/types.go` alongside the other `StrategyType*` constants
- [ ] `UpstreamInjectConfig` struct is present with a single `ProviderName string` field, tagged `json:"providerName" yaml:"providerName"`, and carries both `// +kubebuilder:object:generate=true` and `// +gendoc` markers
- [ ] `ErrUpstreamTokenNotFound` sentinel variable is present: `var ErrUpstreamTokenNotFound = errors.New("upstream token not found")`
- [ ] `BackendAuthStrategy` struct has a new field: `UpstreamInject *UpstreamInjectConfig \`json:"upstreamInject,omitempty" yaml:"upstreamInject,omitempty"\``
- [ ] `TokenExchangeConfig` struct has a new field: `SubjectProviderName string \`json:"subjectProviderName,omitempty" yaml:"subjectProviderName,omitempty"\``
- [ ] `pkg/vmcp/auth/types/zz_generated.deepcopy.go` is regenerated (via `task gen`) and includes `DeepCopyInto`/`DeepCopy` for `UpstreamInjectConfig` and the nil-check pointer copy for the new `UpstreamInject` field on `BackendAuthStrategy`
- [ ] All existing unit tests continue to pass (`task test`)
- [ ] SPDX license header is present on all modified Go files (`task license-check`)

All changes in this phase are confined to `pkg/vmcp/auth/types/types.go`. The file is a leaf package with no dependencies on other `pkg/vmcp/*` packages — keep it that way; no new imports are needed beyond the standard `errors` package for the sentinel.

Add items in the following order to maintain the file's current organization (constants, then structs):

1. Append `StrategyTypeUpstreamInject` to the existing `const` block
2. Add the `UpstreamInjectConfig` struct after `TokenExchangeConfig` with both codegen markers
3. Add `ErrUpstreamTokenNotFound` as a package-level `var` after the `const` block
4. Add `UpstreamInject *UpstreamInjectConfig` to `BackendAuthStrategy` after the existing `TokenExchange` field
5. Add `SubjectProviderName string` to `TokenExchangeConfig` after the existing `SubjectTokenType` field (last field, `omitempty`)

After editing, run `task gen` to regenerate `zz_generated.deepcopy.go`. The generated output will include:
- A new `DeepCopyInto`/`DeepCopy` pair for `UpstreamInjectConfig` (trivial `*out = *in` since it has only scalar fields)
- A nil-check pointer copy block for `BackendAuthStrategy.UpstreamInject` in `BackendAuthStrategy.DeepCopyInto`

### Patterns & Frameworks

- Follow the existing pattern for strategy constants in the `const` block (see `StrategyTypeUnauthenticated`, `StrategyTypeHeaderInjection`, `StrategyTypeTokenExchange`)
- Follow the existing struct documentation style — Go doc comment explaining the type, then field-level comments explaining each field
- `UpstreamInjectConfig` follows the same shape as `HeaderInjectionConfig` (a typed config struct for a named strategy)
- Sentinel errors follow `errors.New(...)` with a descriptive lowercase message; they are checked via `errors.Is()` by callers — always wrap with `%w` at the call site so `errors.Is()` works at any depth
- Use `// +optional` annotation on the new `SubjectProviderName` field comment to signal the field is optional (consistent with operator API conventions)

### Code Pointers

- `pkg/vmcp/auth/types/types.go` — the only file with source code changes; all five additions land here
- `pkg/vmcp/auth/types/zz_generated.deepcopy.go` — regenerated artifact; do not edit manually; run `task gen` after editing `types.go`
- `pkg/vmcp/auth/strategies/unauthenticated.go` — reference for how strategy constants and types are used; note how `Name()` returns the constant directly
- `pkg/vmcp/auth/strategies/tokenexchange.go` — reference for how `TokenExchangeConfig` fields are used in `parseTokenExchangeConfig`; the new `SubjectProviderName` field will be consumed here in Phase 2
- `pkg/vmcp/auth/factory/outgoing.go` — reference for how strategy type constants are passed to `RegisterStrategy`; Phase 2 will add a new registration here using the constant defined in this phase

### Component Interfaces

The following types and values must be present after this phase — downstream phases depend on them exactly as specified:

```go
// New constant — added alongside existing StrategyType* constants
const StrategyTypeUpstreamInject = "upstream_inject"

// New sentinel error — checked with errors.Is() by callers
var ErrUpstreamTokenNotFound = errors.New("upstream token not found")

// New config struct — needs +kubebuilder:object:generate=true and +gendoc markers
// +kubebuilder:object:generate=true
// +gendoc
type UpstreamInjectConfig struct {
    // ProviderName is the name of the upstream IDP provider whose access token
    // should be injected as the Authorization: Bearer header on backend requests.
    ProviderName string `json:"providerName" yaml:"providerName"`
}

// BackendAuthStrategy gains one field (all existing fields unchanged)
type BackendAuthStrategy struct {
    Type            string                 `json:"type" yaml:"type"`
    HeaderInjection *HeaderInjectionConfig `json:"headerInjection,omitempty" yaml:"headerInjection,omitempty"`
    TokenExchange   *TokenExchangeConfig   `json:"tokenExchange,omitempty" yaml:"tokenExchange,omitempty"`
    // UpstreamInject contains configuration for the upstream inject auth strategy.
    // Used when Type = "upstream_inject".
    UpstreamInject  *UpstreamInjectConfig  `json:"upstreamInject,omitempty" yaml:"upstreamInject,omitempty"`
}

// TokenExchangeConfig gains one optional field (all existing fields unchanged)
type TokenExchangeConfig struct {
    // ... existing fields unchanged ...

    // SubjectProviderName is the upstream provider name whose token is used as the
    // RFC 8693 subject token instead of identity.Token when performing token exchange.
    // +optional
    SubjectProviderName string `json:"subjectProviderName,omitempty" yaml:"subjectProviderName,omitempty"`
}
```

### Testing Strategy

No new test files are required for this phase. The acceptance criteria for testing are:

**Compilation check**
- [ ] `go build ./pkg/vmcp/auth/types/...` succeeds with the new types present
- [ ] `go vet ./pkg/vmcp/auth/types/...` reports no issues

**Existing regression tests**
- [ ] `task test` passes without modification — no existing test should need updating since all new fields are additive and optional

**Codegen verification**
- [ ] `task gen` completes without errors
- [ ] `zz_generated.deepcopy.go` contains `DeepCopyInto` and `DeepCopy` for `UpstreamInjectConfig`
- [ ] `zz_generated.deepcopy.go` contains the nil-check pointer copy block for `UpstreamInject` in `BackendAuthStrategy.DeepCopyInto`
- [ ] The generated file is committed alongside the type changes

**License check**
- [ ] `task license-check` passes (SPDX headers on all modified files)

## Out of Scope

- Any runtime behavior changes — new fields are optional and must not alter behavior for configurations that omit them
- Implementation of `UpstreamInjectStrategy` (Phase 2)
- Startup validation rules V-01, V-02, V-06 (Phase 3)
- CRD type additions (`ExternalAuthTypeUpstreamInject`, `UpstreamInjectSpec`) and converter (Phase 4)
- Architecture documentation updates (`docs/arch/02-core-concepts.md`, `docs/vmcp-auth.md`)
- Step-up auth signaling (UC-06) — `ErrUpstreamTokenNotFound` is defined here but the intercept/redirect flow is a separate RFC
- Actor token (`ActorProviderName`) or any parallel field beyond `SubjectProviderName`

## References

- RFC-0054 (primary): `docs/proposals/THV-0054-vmcp-upstream-inject-strategy.md`
- Parent epic: #3925
- RFC-0052 (multi-upstream IDP, defines `identity.UpstreamTokens`): #3924
- RFC-0053 (embedded AS in vMCP, prerequisite for Phase 3): #4120

jhrozek:
## Summary

- Backend auth strategies need a way to inject upstream IDP tokens obtained by the embedded auth server into backend requests. This requires a new `upstream_inject` strategy type.
- Add `StrategyTypeUpstreamInject` and `UpstreamInjectConfig` to backend auth strategy types.
- Add `ValidateAuthServerIntegration(cfg *Config, rc *authserver.RunConfig)` with cross-cutting validation rules V-01 through V-07 (issuer consistency, audience matching, provider existence, etc.).
- The validator takes `*Config` and `*authserver.RunConfig` as separate arguments — no `RuntimeConfig` wrapper type needed.

Fixes #4142
Fixes #4144

## Type of change

- [x] New feature

## Test plan

Draft — stacked on #4348. Review the incremental diff only.

The validation functions accept `(*Config, *RunConfig)` as separate parameters rather than a combined `RuntimeConfig` type. This keeps the config package free of auth server dependencies and makes the function signatures explicit about what they inspect.

### DIFF
--- a/pkg/vmcp/config/validator.go
+++ b/pkg/vmcp/config/validator.go
@@ -484,11 +484,7 @@ func ValidateAuthServerIntegration(cfg *Config, rc *authserver.RunConfig) error 
 	}
 
 	// Issuer and audience consistency between auth server and incoming auth.
-	if err := validateAuthServerIncomingAuthConsistency(cfg, rc); err != nil {
-		return err
-	}
-
-	return nil
+	return validateAuthServerIncomingAuthConsistency(cfg, rc)
 }
 
 // validateAuthServerRunConfig performs lightweight structural validation of the


pkg/vmcp/config/validator.go
```diff
@@ -436,6 +445,185 @@ func (*DefaultValidator) validateCompositeToolRefs(refs []CompositeToolRef) erro
...
+	strategies map[string]*authtypes.BackendAuthStrategy,
+) {
+	if !hasAuthServerWithOIDCIncoming(cfg, rc) {
+		return
+	}
+	for name, strategy := range strategies {
+		if strategy.Type != authtypes.StrategyTypeTokenExchange {
+			continue
+		}
+		if rc.Issuer == cfg.IncomingAuth.OIDC.Issuer {
```
tgrunnagle: I'm a little confused about this check - `rc.Issuer == cfg.IncomingAuth.OIDC.Issuer` when vMCP is pointing to the embedded AS as its authn/z source, right? What's that have to do with `token_exchange`? Would you be able to add a comment about why `upstream_inject` is preferred here?

jhrozek:
I apologize @tgrunnagle I shouldn't have opened the PR before I went through it carefully myself - that was unfiltered CC output that I just gave a cursory look to. I reviewed the patch manually this time and re-did it. I'm marking it ready for review.

pkg/vmcp/config/validator.go
```diff
@@ -436,20 +445,200 @@ func (*DefaultValidator) validateCompositeToolRefs(refs []CompositeToolRef) erro
...
+// backends reference the same provider name.
+func warnDuplicateUpstreamInjectProviders(strategies map[string]*authtypes.BackendAuthStrategy) {
+	seen := make(map[string]string) // providerName -> first backend name
+	for name, strategy := range strategies {
+		if strategy.Type != authtypes.StrategyTypeUpstreamInject || strategy.UpstreamInject == nil {
+			continue
+		}
+		pn := authserver.ResolveUpstreamName(strategy.UpstreamInject.ProviderName)
+		if first, ok := seen[pn]; ok {
+			slog.Warn("multiple upstream_inject backends reference the same provider; likely a copy-paste error",
```
tgrunnagle: Question: why is this a warning case? Two backends requiring the same upstream token provider seems like a legit scenario.

pkg/vmcp/config/validator.go
```diff
@@ -436,20 +445,200 @@ func (*DefaultValidator) validateCompositeToolRefs(refs []CompositeToolRef) erro
...
+		)
+	}
+
+	return nil
+}
+
+// validateAuthServerRequiresOIDC checks that when the auth server is configured,
+// incomingAuth is OIDC. The AS issues tokens that the OIDC middleware
+// validates; without OIDC incoming auth the entire OAuth flow is pointless.
+func validateAuthServerRequiresOIDC(cfg *Config) error {
```
tgrunnagle, line 584: nit: this code is pretty similar to `hasAuthServerWithOIDCIncoming`, maybe could be refactored a bit

pkg/vmcp/config/validator.go
```diff
@@ -436,20 +445,200 @@ func (*DefaultValidator) validateCompositeToolRefs(refs []CompositeToolRef) erro
...
+}
+
+// collectAllBackendStrategies returns all backend auth strategies from the config.
+func collectAllBackendStrategies(cfg *Config) map[string]*authtypes.BackendAuthStrategy {
+	result := make(map[string]*authtypes.BackendAuthStrategy)
+	if cfg.OutgoingAuth == nil {
+		return result
+	}
+	if cfg.OutgoingAuth.Default != nil {
+		result["(default)"] = cfg.OutgoingAuth.Default
```
tgrunnagle: nit: does it matter if this is `"default"` vs with `()`? I'm not sure name is used except for logging, but maybe worth being consistent.

tgrunnagle (dismissed):
Just a couple non-blocking comments

jerm-dro (dismissed):
Clean design — keeping `(Config, RunConfig)` as separate parameters is the right call. Validation rules are thorough and well-tested. The `slices.Contains` cleanup is a nice bonus.

pkg/vmcp/config/validator.go
jhrozek: That function was in an earlier revision but got dropped during the rebase, sorry for the confusing patches. The idea behind it was: if the embedded AS is already managing the upstream OAuth flow, using token_exchange against the same issuer is redundant.

I did expand the doc comment on `validateAuthServerIncomingAuthConsistency` (the issuer/audience consistency check nearby) to clarify that it's strategy-agnostic.

jhrozek: (You might have reviewed the push before the rehoming from scaffolding-2 to main perhaps)

pkg/vmcp/config/validator.go
jhrozek: hmm actually you're right, I was thinking it would more probably be copy-paste but we'd actually be confusing in a legitimate scenario

pkg/vmcp/config/validator.go
jhrozek: The purpose was to avoid a key collision - this is a map key in a map that also stores user-defined backend names, so using just "default" could be problematic if someone named their backend default as well.. But "(default)" does look confusingly similar to authserver.DefaultUpstreamName.

I extracted the string to `defaultStrategyKey = "<default-strategy>"`, hope this is clearer.

pkg/vmcp/config/validator.go
jhrozek: not a nit, good call, added

jhrozek:
the additional commits address each of @tgrunnagle comments, patch per comment for easier review. the original patch is unchanged.

Issue #4143, opened 2026-03-13
Phase 4: Operator reconciler, HTTP handler unit tests, and E2E test coverage for embedded AS


tgrunnagle:
## Description

Wire the embedded authorization server into the Kubernetes operator and add comprehensive test coverage — unit tests for the vMCP HTTP handler, CLI E2E tests for Mode A/Mode B routing and negative validation, and operator E2E tests verifying the `AuthServerConfigValid` status condition lifecycle. This is the final phase of RFC-0053 that closes the loop between the CRD API, the operator reconciler, the config converter, and observable runtime behavior.

## Context

RFC-0053 adds an optional embedded OAuth/OIDC authorization server to vMCP. Phase 1 (#4140) established structural types and CRD fields. Phase 2 (#4141) mounted the AS HTTP routes on the vMCP mux. Phase 3 (#4142) added startup validation rules V-01 through V-07. Phase 4 (this task) completes the Kubernetes operator path: the reconciler resolves `spec.authServerConfigRef`, performs cross-resource validation, surfaces the `AuthServerConfigValid` status condition, and the config converter converts the referenced `MCPExternalAuthConfig` into the `AuthServerConfig` that vMCP reads at startup. This phase also delivers all deferred test coverage and documentation updates.

**Parent epic**: #4120 — vMCP: add embedded authorization server
**RFC document**: `docs/proposals/THV-0053-vmcp-embedded-authserver.md`
**Dependencies**: #4141 (Phase 2: Server wiring), #4142 (Phase 3: Startup validation)
**Blocks**: nothing — this is the final phase. Full end-to-end token flow testing depends on RFC-0052 (#3924) for `identity.UpstreamTokens` population.

### Operator Reconciler

- [ ] `cmd/thv-operator/controllers/virtualmcpserver_controller.go` `runValidations()` includes a new `validateAuthServerConfigRef` step that, when `spec.authServerConfigRef` is non-nil: fetches the referenced `MCPExternalAuthConfig` from the same namespace; verifies its `spec.type` is `"embeddedAuthServer"` (surfacing a `Failed` phase with descriptive message if not); verifies `spec.embeddedAuthServer.issuer` matches `spec.incomingAuth.oidcConfig.inline.issuer` (V-04 check, surfacing `Failed` phase with message containing `"issuer"` and `"mismatch"`); and verifies that the audience derived from `spec.incomingAuth.oidcConfig.inline.audience` is non-empty (defense-in-depth empty-audience guard)
[...]

PR #4350, opened 2026-03-24
Add AuthServerConfig validation condition to VirtualMCPServer reconciler

jhrozek:
## Summary

- When a VirtualMCPServer has an inline `AuthServerConfig`, the operator should validate it during reconciliation and surface the result as a status condition.
- Add `AuthServerConfigValidated` condition to the VirtualMCPServer reconciler that is set when inline AuthServerConfig is present on the spec.
- Add `SetAuthServerConfigValidatedCondition` to the status collector interface.
- Add E2E test for the condition lifecycle.

Fixes: #4143 

## Type of change

- [x] New feature

## Test plan

- [x] E2E tests (new `virtualmcp_authserver_config_test.go`)

Draft — stacked on #4349. Review the incremental diff only.


cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
tgrunnagle: I think this can be removed - looks like it's only used by tests?

jhrozek: thanks, this was a leftover from a previous CC-generated iteration

Issue #4353, opened 2026-03-24
Align OpenAPI/Swagger Type for time.Duration (pkg/auth/remote/config.go)


Sanskarzz:
## Bug description
Currently, the `Timeout` field in `pkg/auth/remote/config.go` uses `swaggertype:"string" example:"5m"`. This representation causes discrepancies in the generated OpenAPI spec compared to other `time.Duration` fields in the project that are represented as `primitive,integer`.

This issue aims to align the `swaggertype` tag for better consistency across the codebase.

### Proposed Changes
- Update the `Timeout` field in `pkg/auth/remote/config.go` to use `swaggertype:"primitive,integer"`.
- Regenerate the swagger documentation using `task docs`.

Sanskarzz:
/assign
Please assign me @JAORMX

JAORMX:
Done! Thanks for the great work!

JAORMX:
@Sanskarzz digging more into this, I'm a little confused. There is no other time.Duration that actually uses `primitive,integer`.

Sanskarzz:
The CI error in [PR](https://github.com/stacklok/toolhive/pull/4314) got fixed after the upstream pull. 
No need for this change.

PR #4354, opened 2026-03-24
Fix: Align OpenAPI Type for Auth Timeout

Sanskarzz:
This PR updates the `swaggertype` tag for the `Timeout` field in `pkg/auth/remote/config.go`. 

This change was originally part of PR #4314 but was requested to be split into its own PR to keep the main feature PR focused on the validating webhook middleware.

Fix: #4353 

### Changes Included
- Modified `pkg/auth/remote/config.go` to use `swaggertype:"primitive,integer"` for the `Timeout` field.
- Updated `docs/server/docs.go`, `docs/server/swagger.json`, and `docs/server/swagger.yaml` accordingly.

### Verification
- Ran `task docs` to confirm no unwanted schema bloat for `time.Duration`.
- Verified documentation synchronization with `./cmd/help/verify.sh`.

### DIFF
--- a/pkg/auth/remote/config.go
+++ b/pkg/auth/remote/config.go
@@ -11,7 +11,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/stacklok/toolhive-core/registry/types"
+	registry "github.com/stacklok/toolhive-core/registry/types"
 	httpval "github.com/stacklok/toolhive-core/validation/http"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/auth/remote/config.go
```diff
@@ -23,7 +23,7 @@ type Config struct {
 	ClientSecretFile string        `json:"client_secret_file,omitempty" yaml:"client_secret_file,omitempty"`
 	Scopes           []string      `json:"scopes,omitempty" yaml:"scopes,omitempty"`
 	SkipBrowser      bool          `json:"skip_browser,omitempty" yaml:"skip_browser,omitempty"`
-	Timeout          time.Duration `json:"timeout,omitempty" yaml:"timeout,omitempty" swaggertype:"string" example:"5m"`
+	Timeout          time.Duration `json:"timeout,omitempty" yaml:"timeout,omitempty" swaggertype:"primitive,integer"`
```
JAORMX, line 26: @Sanskarzz digging more into this, I'm a little confused. There is no other time.Duration that actually uses `primitive,integer`.

Sanskarzz: I was also confused, but after making this change here, the doc's error got resolved locally. And also removing swaggertype from here also resolved the error locally.
Do I need to make changes to all timeout fields?

amirejaz:
@Sanskarzz could you add the reason/motivation for this change?

Sanskarzz:
> @Sanskarzz could you add the reason/motivation for this change?

I have created this PR because this change was fixing CI docs error for this [PR](https://github.com/stacklok/toolhive/pull/4314), discussed [here](https://github.com/stacklok/toolhive/pull/4314#discussion_r2979515658).

BTW, somehow now the CI swag docs error got fixed automatically after upstream/main pull, I will close this PR and the associated Issue.

Thanks

PR #4359, opened 2026-03-25
Validate volume mount source paths before detached process

### LINKED ISSUE #2485: Volume mount errors not reported when source path doesn't exist

mvanhorn:
## Summary

Volume mount errors were silently buried in proxy log files when the source path didn't exist. Users saw a success message but the server immediately entered error state. This adds pre-flight validation in `processVolumeMounts()` so path errors surface immediately in the terminal.

This picks up where #3932 left off, addressing the two must-fix items from @JAORMX's review:
1. Uses `mount.IsResourceURI()` instead of `strings.HasPrefix` for resource URI detection
2. Uses `os.IsNotExist(err)` to distinguish "not exists" from "permission denied"

Fixes #2485

## Type of change

- [x] Bug fix

## Test plan

- [x] Ran `go build ./...` - passes
- [x] Ran `go test ./pkg/runner/... -count=1` - all tests pass
- [x] Added test case for non-existent source path error
- [x] Updated existing volume mount tests to use real temp directories
- [x] Verified resource URIs skip validation (via `mount.IsResourceURI()`)
- [x] Verified operator context skips validation (via `BuildContextCLI` check)

## Changes

| File | Change |
|------|--------|
| `pkg/runner/config_builder.go` | Add source path validation after `mount.Parse()`, before duplicate check |
| `pkg/runner/config_builder_test.go` | Use real temp dirs, add non-existent path test case |
| `pkg/runner/config_test.go` | Use real temp dirs for volume processing test |

### DIFF
--- a/pkg/runner/config_builder.go
+++ b/pkg/runner/config_builder.go
@@ -1063,13 +1063,9 @@ func (b *runConfigBuilder) processVolumeMounts() error {
 
 		// Validate source path exists on host filesystem (CLI context only)
 		if b.buildContext == BuildContextCLI && !mount.IsResourceURI() {
-			absSource := source
-			if !filepath.IsAbs(source) {
-				cwd, err := os.Getwd()
-				if err != nil {
-					return fmt.Errorf("failed to resolve relative path %s: %w", source, err)
-				}
-				absSource = filepath.Join(cwd, source)
+			absSource, err := filepath.Abs(source)
+			if err != nil {
+				return fmt.Errorf("failed to resolve path %s: %w", source, err)
 			}
 			if _, err := os.Stat(absSource); err != nil {
 				if os.IsNotExist(err) {
--- a/pkg/runner/config_builder.go
+++ b/pkg/runner/config_builder.go
@@ -9,6 +9,8 @@ import (
 	"log/slog"
 	"maps"
 	"net/url"
+	"os"
+	"path/filepath"
 	"slices"
 	"strings"
 
@@ -1057,6 +1059,24 @@ func (b *runConfigBuilder) processVolumeMounts() error {
 		source, target, err := mount.Parse()
 		if err != nil {
 			return fmt.Errorf("invalid volume format: %s (%w)", volume, err)
+		}
+
+		// Validate source path exists on host filesystem (CLI context only)
+		if b.buildContext == BuildContextCLI && !mount.IsResourceURI() {
+			absSource := source
+			if !filepath.IsAbs(source) {
+				cwd, err := os.Getwd()
+				if err != nil {
+					return fmt.Errorf("failed to resolve relative path %s: %w", source, err)
+				}
+				absSource = filepath.Join(cwd, source)
+			}
+			if _, err := os.Stat(absSource); err != nil {
+				if os.IsNotExist(err) {
+					return fmt.Errorf("volume source path does not exist: %s", absSource)
+				}
+				return fmt.Errorf("cannot access volume source path %s: %w", absSource, err)
+			}
 		}
 
 		// Check for duplicate mount target
--- a/pkg/runner/config_builder_test.go
+++ b/pkg/runner/config_builder_test.go
@@ -713,6 +713,32 @@ func TestNewOperatorRunConfigBuilder(t *testing.T) {
 	assert.NotNil(t, config.ContainerLabels, "ContainerLabels should be initialized")
 }
 
+// TestOperatorContextSkipsVolumeValidation verifies that operator context does not
+// validate volume source paths on the host filesystem, since paths reference
+// Kubernetes node paths rather than the operator pod's filesystem.
+func TestOperatorContextSkipsVolumeValidation(t *testing.T) {
+	t.Parallel()
+
+	mockValidator := &mockEnvVarValidator{}
+	imageMetadata := &regtypes.ImageMetadata{
+		BaseServerMetadata: regtypes.BaseServerMetadata{
+			Name:  "test-image",
+			Tools: []string{"tool1"},
+		},
+	}
+
+	config, err := NewOperatorRunConfigBuilder(
+		context.Background(),
+		imageMetadata,
+		nil,
+		mockValidator,
+		WithVolumes([]string{"/nonexistent/node/path:/container"}),
+		WithPermissionProfile(permissions.BuiltinNoneProfile()),
+	)
+	require.NoError(t, err, "Operator context should not validate source paths")
+	assert.NotNil(t, config)
+}
+
 // TestWithEnvVars tests the WithEnvVars method
 func TestWithEnvVars(t *testing.T) {
 	t.Parallel()
--- a/pkg/runner/config_builder_test.go
+++ b/pkg/runner/config_builder_test.go
@@ -242,6 +242,12 @@ func TestRunConfigBuilder_Build_WithVolumeMounts(t *testing.T) {
 	// Create a mock environment variable validator
 	mockValidator := &mockEnvVarValidator{}
 
+	// Create real temp directories so volume mount source path validation passes
+	hostDir := t.TempDir()
+	hostDir1 := t.TempDir()
+	hostDir2 := t.TempDir()
+	hostDir3 := t.TempDir()
+
 	testCases := []struct {
 		name                string
 		builderOptions      []RunConfigBuilderOption
@@ -261,7 +267,7 @@ func TestRunConfigBuilder_Build_WithVolumeMounts(t *testing.T) {
 		{
 			name: "Volumes without permission profile but with profile name",
 			builderOptions: []RunConfigBuilderOption{
-				WithVolumes([]string{"/host:/container"}),
+				WithVolumes([]string{hostDir + ":/container"}),
 				WithPermissionProfileNameOrPath(permissions.ProfileNone),
 			},
 			expectError:         false,
@@ -271,7 +277,7 @@ func TestRunConfigBuilder_Build_WithVolumeMounts(t *testing.T) {
 		{
 			name: "Read-only volume with existing profile",
 			builderOptions: []RunConfigBuilderOption{
-				WithVolumes([]string{"/host:/container:ro"}),
+				WithVolumes([]string{hostDir + ":/container:ro"}),
 				WithPermissionProfile(permissions.BuiltinNoneProfile()),
 			},
 			expectError:         false,
@@ -281,7 +287,7 @@ func TestRunConfigBuilder_Build_WithVolumeMounts(t *testing.T) {
 		{
 			name: "Read-write volume with existing profile",
 			builderOptions: []RunConfigBuilderOption{
-				WithVolumes([]string{"/host:/container"}),
+				WithVolumes([]string{hostDir + ":/container"}),
 				WithPermissionProfile(permissions.BuiltinNoneProfile()),
 			},
 			expectError:         false,
@@ -292,9 +298,9 @@ func TestRunConfigBuilder_Build_WithVolumeMounts(t *testing.T) {
 			name: "Multiple volumes with existing profile",
 			builderOptions: []RunConfigBuilderOption{
 				WithVolumes([]string{
-					"/host1:/container1:ro",
-					"/host2:/container2",
-					"/host3:/container3:ro",
+					hostDir1 + ":/container1:ro",
+					hostDir2 + ":/container2",
+					hostDir3 + ":/container3:ro",
 				}),
 				WithPermissionProfile(permissions.BuiltinNoneProfile()),
 			},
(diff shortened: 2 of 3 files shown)


pkg/runner/config_builder.go
```diff
@@ -1059,6 +1061,24 @@ func (b *runConfigBuilder) processVolumeMounts() error {
...
 
+		// Validate source path exists on host filesystem (CLI context only)
+		if b.buildContext == BuildContextCLI && !mount.IsResourceURI() {
+			absSource := source
+			if !filepath.IsAbs(source) {
+				cwd, err := os.Getwd()
+				if err != nil {
+					return fmt.Errorf("failed to resolve relative path %s: %w", source, err)
+				}
+				absSource = filepath.Join(cwd, source)
```
jhrozek: **Suggestion**: Consider using `filepath.Abs(source)` instead of the manual `os.Getwd()` + `filepath.Join()` approach. `filepath.Abs` does the same thing internally but also calls `filepath.Clean`, normalizing `..` segments in the resolved path. This would also be consistent with the existing `convertRelativePathToAbsolute` in `pkg/container/docker/client.go:807` which already uses `filepath.Abs`.

```suggestion
			absSource, err := filepath.Abs(source)
			if err != nil {
				return fmt.Errorf("failed to resolve volume source path %s: %w", source, err)
			}
```

jhrozek (commented):
Review from Claude Code — 2 inline comments on the new validation logic and test coverage.

pkg/runner/config_builder_test.go
```diff
@@ -310,6 +316,14 @@ func TestRunConfigBuilder_Build_WithVolumeMounts(t *testing.T) {
 			},
 			expectError: true,
 		},
+		{
+			name: "Non-existent source path",
+			builderOptions: []RunConfigBuilderOption{
+				WithVolumes([]string{"/nonexistent/path:/container"}),
+				WithPermissionProfile(permissions.BuiltinNoneProfile()),
+			},
```
jhrozek, line 324: **Suggestion**: Consider adding test cases for the two skip conditions in the new validation:

1. **Operator context**: `NewOperatorRunConfigBuilder` with `WithVolumes([]string{"/nonexistent/path:/container"})` should succeed — operator paths reference K8s node paths, not the operator pod's filesystem.
2. **Resource URI**: A CLI-context build with `WithVolumes([]string{"volume://my-volume:/container/data"})` should succeed — resource URIs have no host path to stat.

Both guards are new code and currently have no test coverage.

mvanhorn:
Addressed both suggestions in 53298a0:
- Switched to `filepath.Abs(source)` which handles the Getwd+Join+Clean in one call
- Added `TestOperatorContextSkipsVolumeValidation` confirming non-existent K8s node paths pass in operator context

pkg/runner/config_builder.go
```diff
@@ -1059,6 +1061,20 @@ func (b *runConfigBuilder) processVolumeMounts() error {
...
 		}
 
+		// Validate source path exists on host filesystem (CLI context only)
+		if b.buildContext == BuildContextCLI && !mount.IsResourceURI() {
+			absSource, err := filepath.Abs(source)
+			if err != nil {
+				return fmt.Errorf("failed to resolve path %s: %w", source, err)
+			}
+			if _, err := os.Stat(absSource); err != nil {
+				if os.IsNotExist(err) {
```
amirejaz, line 1071: **Nit — prefer `errors.Is` over `os.IsNotExist`**

`os.IsNotExist` predates Go 1.13's error-unwrapping model and will miss errors wrapped with `%w`. The idiomatic replacement is `errors.Is(err, os.ErrNotExist)`, which correctly traverses the error chain.

You'll also need to add `"errors"` to the import block.

```suggestion
				if errors.Is(err, os.ErrNotExist) {
```

See also: [Go blog — Working with Errors](https://go.dev/blog/go1.13-errors#checking-whether-an-error-is-a-specific-value).

amirejaz (commented):
Thanks for picking this back up and addressing the previous round of feedback — the switch to `filepath.Abs`, the real temp-dir usage in tests, and the new `TestOperatorContextSkipsVolumeValidation` test are all improvements over the original.

Two items remain before this is ready to merge:

1. **`os.IsNotExist` → `errors.Is`** — the idiomatic form since Go 1.13; `os.IsNotExist` does not unwrap error chains and is considered deprecated for new code.
2. **Missing test for the resource-URI skip** — the previous review specifically called this out as lacking coverage; the operator-context test was added but the resource-URI branch still has no test.

Both are small, targeted changes. Happy to help if anything is unclear.

pkg/runner/config_builder_test.go
```diff
@@ -310,6 +316,14 @@ func TestRunConfigBuilder_Build_WithVolumeMounts(t *testing.T) {
...
+			expectError: true,
+		},
```
amirejaz, line 326: **Missing test for the resource-URI skip path**

The previous review asked for coverage of both skip conditions. The operator-context case was added (`TestOperatorContextSkipsVolumeValidation`), but the resource-URI branch (`!mount.IsResourceURI()`) still has no test.

Without it, a future refactor could drop the guard and no test would catch the regression. Consider adding a case to this table:

```go
{
    name: "Resource URI source skips filesystem validation",
    builderOptions: []RunConfigBuilderOption{
        WithVolumes([]string{"volume://my-data:/container/data"}),
        WithPermissionProfile(permissions.BuiltinNoneProfile()),
    },
    expectError: false,
},
```

This passes a path that would fail `os.Stat` if the guard were absent, so it directly exercises the branch.

ChrisJBurns:
Thanks for picking this up, @mvanhorn — the approach here is solid, and the iterations (switching to `filepath.Abs`, using real temp dirs in the tests, and adding the operator-context skip test) were all good improvements.

Unfortunately this is now superseded: issue #2485 was fixed and closed by #4568, which merged on 2026-04-07. That PR landed the same pre-flight `os.Stat` validation in `processVolumeMounts()` — CLI-context only, operator path skipped — across the same three files.

The merged version also already covers the outstanding review feedback here:
- It uses the idiomatic `errors.Is(err, os.ErrNotExist)` rather than `os.IsNotExist`.
- The `mount.IsResourceURI()` guard is dropped entirely — that method no longer lives in this package, so the resource-URI skip test no longer applies. The `BuildContextCLI` gate is sufficient.
- Equivalent tests exist in `main` for both the CLI nonexistent-path error and the operator-context skip.

Since the fix is already in `main`, I'm closing this as superseded by #4568. Really appreciate the contribution and the responsiveness to review — sorry the timing meant it got beaten to the merge.

Issue #4360, opened 2026-03-25
Reset registry auth when registry changes


eleftherias:
Couple the registry server url and the registry server auth, so that when the URL changes, the auth is reset.

PR #4361, opened 2026-03-25
Couple registry server URL and registry auth

eleftherias:
Issue #4360

## Summary
When the registry URL changes, we should also reset the registry auth configuration.

## Changes
- Add registry auth flags to `thv config set-registry`
- Override existing registry auth whenever `thv config set-registry` is called
- Deprecate auth specific command `thv config set-registry-auth` and `thv config unset-registry-auth`
- Override existing registry auth when API receives empty auth
- Update docs to state authorization server needs to be OIDC-compliant, vanilla OAuth isn't supported at this time


pkg/api/v1/registry.go
```diff
@@ -518,44 +520,47 @@ func (rr *RegistryRoutes) updateRegistry(w http.ResponseWriter, r *http.Request)
...
+		// Check if it's a validation error - return 502 Bad Gateway
+		if isValidationError(err) {
+			http.Error(w, err.Error(), http.StatusBadGateway)
 			return
 		}
+		// Other errors - return 400 Bad Request
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
 	}
+	responseType = registryType
```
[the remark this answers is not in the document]
eleftherias: This behavioural change is intentional

pkg/registry/auth/login.go
```diff
@@ -18,24 +18,24 @@ import (
 )
 
 // DefaultOAuthScopes returns the default OAuth scopes for registry authentication.
-// openid is required for OIDC, offline_access is required for refresh tokens.
+// offline_access is required for the provider to return a refresh token.
 func DefaultOAuthScopes() []string {
-	return []string{"openid", "offline_access"}
+	return []string{"offline_access"}
```
jhrozek: **Question:** The default scopes changed from `["openid", "offline_access"]` to just `["offline_access"]`. Was the removal of `openid` intentional? The previous comment said "openid is required for OIDC." Some IdPs may not issue refresh tokens or user claims without it. If intentional, could you add a brief rationale to the PR description?

cmd/thv/app/config.go
```diff
@@ -163,13 +180,30 @@ func unsetCACertCmdFunc(_ *cobra.Command, _ []string) error {
 func setRegistryCmdFunc(_ *cobra.Command, args []string) error {
 	input := args[0]
 
+	provider := config.NewDefaultProvider()
+
 	service := registry.NewConfigurator()
 	registryType, err := service.SetRegistryFromInput(input, allowPrivateRegistryIp)
```
jhrozek, line 192: **Nit (ordering):** The registry URL is persisted to disk here *before* auth is cleared at line 195. If `UnsetAuth()` fails (e.g. disk full), the config ends up with the new URL but old tokens — the exact scenario this PR prevents. Consider clearing auth first (matching the pattern in `ensureRegistryURL` in `login.go:201` which already does this correctly), or combining both into a single `UpdateConfig` call.

cmd/thv/app/config_registryauth.go
```diff
@@ -21,9 +21,10 @@ var (
 )
 
 var setRegistryAuthCmd = &cobra.Command{
-	Use:   "set-registry-auth",
-	Short: "Configure OAuth/OIDC authentication for the registry",
-	Long: `Configure OAuth/OIDC authentication for the remote MCP server registry.
+	Use:        "set-registry-auth",
+	Short:      "Configure OIDC authentication for the registry",
+	Deprecated: "use 'thv config set-registry' with --issuer and --client-id flags instead",
```
jhrozek, line 26: **Question:** This deprecated command still sets auth without checking whether a registry URL is configured, bypassing the coupling invariant this PR establishes. Since it's deprecated, would it be simpler to just remove `set-registry-auth` and `unset-registry-auth` entirely rather than keeping them as deprecated?

cmd/thv/app/config_registryauth.go
```diff
@@ -32,15 +33,16 @@ Examples:
...
 }
 
 var unsetRegistryAuthCmd = &cobra.Command{
-	Use:   "unset-registry-auth",
-	Short: "Remove registry authentication configuration",
-	Long:  "Remove the OAuth/OIDC authentication configuration for the registry.",
-	RunE:  unsetRegistryAuthCmdFunc,
+	Use:        "unset-registry-auth",
+	Short:      "Remove registry authentication configuration",
+	Deprecated: "use 'thv config unset-registry' instead, which clears both registry and auth",
```
jhrozek: **Nit:** The deprecation message suggests `thv config unset-registry` as a replacement, but that also removes the registry URL — functionally different from the old `unset-registry-auth` which only cleared auth. Consider suggesting `thv config set-registry <current-url>` (without auth flags) or `thv registry login` to re-authenticate instead.

pkg/registry/auth_manager.go
```diff
@@ -64,7 +64,7 @@ func NewAuthManager(provider config.Provider) AuthManager {
 	}
 }
 
-// SetOAuthAuth configures OAuth/OIDC authentication for the registry.
+// SetOAuthAuth configures OIDC authentication for the registry.
 // PKCE (S256) is always enforced and not configurable.
 func (c *DefaultAuthManager) SetOAuthAuth(issuer, clientID, audience string, scopes []string) error {
 	updateFn, err := auth.ConfigureOAuth(context.Background(), issuer, clientID, audience, scopes)
```
jhrozek: **Nit:** `SetOAuthAuth` performs network I/O (OIDC discovery with a 10s timeout) but hardcodes `context.Background()`. The API handler at `registry.go` has a request context available that can't be propagated. Consider adding `ctx context.Context` to the `AuthManager` interface — all three callers have contexts available.

pkg/api/v1/registry.go
```diff
@@ -120,7 +120,8 @@ func (rr *RegistryRoutes) registryAuthLogin(w http.ResponseWriter, r *http.Reque
 
 	if err := auth.Login(r.Context(), rr.configProvider, secretsProvider, auth.LoginOptions{}); err != nil {
 		if isRegistryAuthError(err) {
-			http.Error(w, "Registry OAuth not configured; use 'thv config set-registry-auth' first", http.StatusBadRequest)
+			http.Error(w, "Registry OAuth not configured; call PUT /registry/default with a client ID and "+
+				"issuer URL first", http.StatusBadRequest)
```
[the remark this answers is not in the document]
eleftherias: Updated in new commit

cmd/thv/app/config_registryauth.go
eleftherias: It would be simpler to remove them, but since they've already been released I assumed we should first deprecate and then remove them in another release. I can update the `set-registry-auth` command to first check if a registry URL is configured?

pkg/registry/auth/login.go
eleftherias: This was intentional, I added a brief rational in the PR description
> Remove openid from default scopes because we don't use the ID token, only the access token

The long description is that even though we need an OIDC compliant authorization server, to use the issuer URL, we're not actually doing an OpenID Connect request, because we don't use the ID token.
We're just doing a OAuth 2.0 authorization flow to get the access token.

However, this is all theoretic. If in practice the IdPs expect the `openid` scope, I can add it back, it doesn't harm anything.

eleftherias: I added the `openid` scope back. Better to be safe.

cmd/thv/app/config.go
eleftherias: Updated

cmd/thv/app/config_registryauth.go
eleftherias: Updated `set-registry-auth` to check that a registry url is configured first

cmd/thv/app/config_registryauth.go
eleftherias: Updated

pkg/registry/auth_manager.go
eleftherias: Updated

PR #4365, opened 2026-03-25
Add OTEL distributed tracing and optional Sentry integration to the API server

samuv:
## Summary

Adds optional distributed tracing and error reporting to \`thv serve\` using an OTEL-first approach with Sentry as an optional backend.

- **\`otelhttp\` middleware** extracts W3C \`traceparent\` headers and creates per-request spans. A post-routing \`chiRouteSpanNamer\` middleware renames spans after chi resolves the route, producing parameterised names like \`GET /api/v1beta/workloads/{name}\` instead of raw URL paths
- **\`thv config otel set-endpoint\`** configures the OTLP backend — no new CLI flags for tracing; \`thv serve\` reuses the same OTEL config as \`thv run\`
- **Span processor registry** (\`telemetry.RegisterSpanProcessor\`) allows integrations to self-register during their own \`Init\` — \`pkg/telemetry/NewServeProvider\` has no provider-specific knowledge
- **Sentry bridge** (\`sentry-go/otel\`): when a DSN is provided, the Sentry span processor self-registers so spans export to both the OTLP backend and Sentry simultaneously
- **Sentry-only mode**: when no OTLP endpoint is configured but a DSN is provided, a minimal OTEL SDK is created — no external collector required
- **Panic and 5xx error recording**: recovery middleware and error handler call both \`span.RecordError\` + \`span.SetStatus(codes.Error)\` (vendor-agnostic OTEL) and \`sentrypkg.RecoverPanic\` / \`sentrypkg.CaptureException\` (explicit Sentry Issues). The Sentry span processor only creates transactions; explicit hub calls are required for errors to appear in the Sentry Issues tab
- **\`custom.instance_id\` tag**: every Sentry event is tagged with the anonymous instance ID for correlation with ToolHive Studio
- **ToolHive Studio compatibility**: enable \`propagateTraceparent: true\` in Studio's Sentry JS SDK to send W3C \`traceparent\` headers, giving end-to-end distributed traces across frontend and backend

When no DSN and no OTEL endpoint are configured, all operations are no-ops with zero overhead.

## Type of change

- New feature

Yes. \`thv serve\` now accepts optional \`--sentry-dsn\`, \`--sentry-environment\` and \`--sentry-traces-sample-rate\` flags (also readable from \`SENTRY_DSN\` / \`SENTRY_ENVIRONMENT\` env vars). Distributed tracing is configured via the existing \`thv config otel set-endpoint\` command.

The \`telemetry.RegisterSpanProcessor\` registry makes the OTEL provider setup fully provider-agnostic. Sentry self-registers during \`sentrypkg.Init\` — adding a future provider (e.g. Datadog) requires no changes to \`server.go\`.

Shutdown ordering is deterministic via LIFO defer: the OTEL provider (which flushes the Sentry span processor) shuts down before \`sentrypkg.Close()\`.

## Large PR Justification

This PR introduces a new cross-cutting observability capability that spans multiple layers of the stack and cannot be meaningfully split without leaving the codebase in a broken or misleading intermediate state:

- **Dependency chain is atomic** — the OTEL provider layer (`pkg/telemetry`), the span processor registry, the Sentry bridge (`pkg/sentry`), and the API server wiring (`pkg/api`, `cmd/thv/app`) must all land together. Merging any subset would either leave dead code, broken imports, or a non-functional tracing pipeline
- **Middleware ordering is a correctness constraint** — `otelhttp` must be outermost, `chiRouteSpanNamer` must be inner, and `recovery` must be innermost. Any partial merge that changes this order would silently break panic capture on active spans
- **Review feedback addressed in-place** — roughly half the diff is direct responses to review comments (security fixes, design improvements, test additions). These are inseparable from the original feature commits since they correct the same code paths

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -778,8 +778,6 @@ func (a *clientPathAdapter) ListSkillSupportingClients() []string {
 	return result
 }
 
-// generateNonce creates a cryptographically random nonce for server instance
-// identification. It returns a 32-character hex string (16 random bytes).
 // chiRouteSpanNamer is a middleware that renames the active OTEL span to reflect
 // the matched chi route pattern (e.g. "GET /api/v1beta/workloads/{name}") and
 // records each URL path parameter as a span attribute for drill-down visibility.


pkg/sentry/sentry.go
```diff
@@ -0,0 +1,127 @@
...
+	err := sentry.Init(sentry.ClientOptions{
+		Dsn:              cfg.DSN,
+		Environment:      cfg.Environment,
+		Release:          fmt.Sprintf("toolhive@%s", vi.Version),
+		TracesSampleRate: cfg.TracesSampleRate,
+		Debug:            cfg.Debug,
+		EnableTracing:    true,
+		AttachStacktrace: true,
+		SendDefaultPII:   true,
+	})
```
ChrisJBurns, line 59: **[Security - High]** `SendDefaultPII: true` instructs the Sentry SDK to attach `Authorization` headers, cookies, client IPs, and request bodies to every event. The API server handles OIDC bearer tokens, so any panic or captured exception will ship those credentials to Sentry's SaaS infrastructure.

This directly conflicts with the project's own rule in `.claude/rules/go-style.md`:
> "Never log credentials, tokens, API keys, or passwords"

**Suggestion:** Set `SendDefaultPII: false`. If specific non-sensitive context is needed, use a `BeforeSend` callback to explicitly allowlist safe fields and scrub `Authorization` headers, cookies, and other sensitive request data.

pkg/sentry/sentry.go
```diff
@@ -0,0 +1,127 @@
...
+	slog.Debug("sentry initialized", "environment", cfg.Environment)
+
+	// Tag every event and transaction with the anonymous instance ID so that
+	// Sentry events from the API server can be correlated with those from
+	// toolhive-studio, which sets the same value as custom.user_id.
+	if id, err := updates.TryGetAnonymousID(); err == nil && id != "" {
+		sentry.ConfigureScope(func(scope *sentry.Scope) {
+			scope.SetTag("custom.user_id", id)
+		})
+		slog.Debug("sentry anonymous instance ID tagged", "id", id)
```
ChrisJBurns, line 76: **[Nit]** The tag name `custom.user_id` is misleading — this is an anonymous *instance* ID, not a *user* ID. Anyone with Sentry project access could misinterpret it as PII. Combined with `SendDefaultPII: true`, this tag alongside IP addresses and request headers could de-anonymize the installation.

Also, Sentry's built-in PII detection heuristics may flag fields named `user_id`.

**Suggestion:** Rename to `custom.instance_id`.

pkg/api/server.go
```diff
@@ -176,7 +187,28 @@ func (b *ServerBuilder) WithSkillManager(manager skills.SkillService) *ServerBui
...
+	// and making span.RecordError a no-op. With otelhttp outer:
+	//   1. otelhttp starts span, calls next
+	//   2. recovery catches panic, calls span.RecordError, returns 500 normally
+	//   3. otelhttp's defer fires: span has error recorded + 500 status, then ends
+	// The span name uses chi's matched route pattern (e.g. "GET /api/v1beta/workloads/{name}")
+	// for clean grouping in Sentry and OTEL backends.
+	if b.otelEnabled {
+		r.Use(otelhttp.NewMiddleware("thv-api",
+			otelhttp.WithSpanNameFormatter(func(_ string, r *http.Request) string {
+				if routeCtx := chi.RouteContext(r.Context()); routeCtx != nil && routeCtx.RoutePattern() != "" {
```
ChrisJBurns: **[Bug - High]** The span name formatter is called at **span-start time** by `otelhttp`, which is *before* chi's routing has populated `RouteContext.RoutePattern()`. At this point in the middleware chain, chi has created a `RouteContext` (so the nil check passes), but `RoutePattern()` returns `""` because route matching hasn't happened yet.

`otelhttp` v0.65.0+ has a post-handler rename that checks `r.Pattern`, but that's a Go 1.22+ `net/http.ServeMux` feature — chi does not set `r.Pattern`.

**Result:** Every span will be named with the raw URL path (e.g., `GET /api/v1beta/workloads/my-server`) instead of the parameterized route (`GET /api/v1beta/workloads/{name}`). This causes **unbounded cardinality** in OTEL/Sentry backends, defeating route-based grouping.

**Fix:** Add an inner middleware (registered after chi routing) that renames the span:
```go
func chiRouteTagMiddleware(next http.Handler) http.Handler {
    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        next.ServeHTTP(w, r)
        if rctx := chi.RouteContext(r.Context()); rctx != nil && rctx.RoutePattern() != "" {
            trace.SpanFromContext(r.Context()).SetName(r.Method + " " + rctx.RoutePattern())
        }
    })
}
```

pkg/api/server.go
```diff
@@ -176,7 +187,28 @@ func (b *ServerBuilder) WithSkillManager(manager skills.SkillService) *ServerBui
...
+	// middleware catches a panic. If recovery were outer, otelhttp's defer span.End()
+	// would fire during panic unwinding — before recover() — leaving the span ended
+	// and making span.RecordError a no-op. With otelhttp outer:
+	//   1. otelhttp starts span, calls next
+	//   2. recovery catches panic, calls span.RecordError, returns 500 normally
+	//   3. otelhttp's defer fires: span has error recorded + 500 status, then ends
+	// The span name uses chi's matched route pattern (e.g. "GET /api/v1beta/workloads/{name}")
+	// for clean grouping in Sentry and OTEL backends.
+	if b.otelEnabled {
+		r.Use(otelhttp.NewMiddleware("thv-api",
```
ChrisJBurns: **[Security - Low]** With `otelhttp` as the outermost middleware, it reads W3C `traceparent` and `tracestate` headers from incoming requests *before* authentication. An untrusted client can:
1. Inject arbitrary trace IDs to correlate their requests across your telemetry backend
2. Set `sampled=1` in the traceparent to force 100% sampling of their requests, increasing telemetry costs
3. Pollute trace context with large `tracestate` headers

This is partially mitigated if you switch to a `ParentBased` sampler that ignores remote parent sampling decisions (see the comment on `otlp/tracing.go`).

pkg/api/server.go
```diff
@@ -757,13 +789,15 @@ func generateNonce() (string, error) {
...
+// If otelEnabled is true, OTEL HTTP middleware is added for W3C traceparent extraction and distributed tracing.
 func Serve(
 	ctx context.Context,
 	address string,
 	isUnixSocket bool,
 	debugMode bool,
 	enableDocs bool,
 	oidcConfig *auth.TokenValidatorConfig,
+	otelEnabled bool,
 	middlewares ...func(http.Handler) http.Handler,
```
ChrisJBurns, line 838: **[Design - Must Fix]** Adding `otelEnabled bool` as a positional parameter before the variadic `middlewares` breaks the public API of this exported function. The `ServerBuilder` already has a fluent `With*` pattern (`WithDebugMode`, `WithDocs`, `WithOIDCConfig`, etc.).

This approach also makes the call site progressively harder to read with 7+ positional booleans/strings.

**Fix:** The `otelEnabled` config should only be added as a field on `ServerBuilder` via the existing `WithOtelEnabled(bool)` method, and the `Serve` convenience function signature should remain unchanged. The caller can pass the OTEL middleware through the existing `middlewares` variadic, or the `Serve` function can accept a config struct.

cmd/thv/app/server.go
```diff
@@ -101,10 +127,75 @@ var serveCmd = &cobra.Command{
 			}()
 		}
 
-		return s.Serve(ctx, address, isUnixSocket, debugMode, enableDocs, oidcConfig)
+		return s.Serve(ctx, address, isUnixSocket, debugMode, enableDocs, oidcConfig, otelEnabled)
 	},
 }
 
+// initServeOTEL initialises the OTEL provider for thv serve using the global config
+// (set via `thv config otel set-endpoint`). No new CLI flags are introduced; serve reuses
```
ChrisJBurns: **[Convention - Must Fix]** This 50+ line function reads config, builds telemetry config, creates an OTEL provider, and spawns a shutdown goroutine. Per `.claude/rules/cli-commands.md`:

> **CRITICAL**: CLI commands must be thin wrappers that delegate to business logic in `pkg/`.

This business logic should be extracted to `pkg/telemetry/` (e.g., a `NewServeProvider` function) and invoked from here. The existing OTEL config construction already lives in `pkg/telemetry/config.go` (`MaybeMakeConfig`, `NewProvider`). The new function should follow the same pattern.

cmd/thv/app/server.go
```diff
@@ -41,6 +47,26 @@ var serveCmd = &cobra.Command{
...
+			DSN:              sentryDSN,
+			Environment:      sentryEnvironment,
+			TracesSampleRate: sentryTracesSampleRate,
+			Debug:            debugMode,
+		}
+		if err := sentrypkg.Init(sentryCfg); err != nil {
+			return fmt.Errorf("failed to initialize sentry: %w", err)
+		}
+		defer sentrypkg.Close()
+
```
ChrisJBurns, line 72: **[Bug - Medium]** There's a shutdown race between this `defer sentrypkg.Close()` and the goroutine in `initServeOTEL` that calls `provider.Shutdown()` on `ctx.Done()`. Both are triggered by context cancellation but execute non-deterministically:

1. `sentrypkg.Close()` calls `sentry.Flush(2s)` and tears down the Sentry client
2. The goroutine's `provider.Shutdown()` calls the Sentry span processor's `Flush`, which calls `hub.Flush(2s)`

If `sentrypkg.Close()` runs first, the span processor operates on an already-flushed client.

**Fix:** Make shutdown deterministic. Either:
- (a) Remove `defer sentrypkg.Close()` and rely on the OTEL provider shutdown to flush Sentry via the span processor, or
- (b) Make the OTEL provider shutdown synchronous (via defer, not goroutine), ordered before `sentrypkg.Close()` via LIFO:
```go
defer sentrypkg.Close()        // runs second (LIFO)
defer func() {                 // runs first
    shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
    defer cancel()
    provider.Shutdown(shutdownCtx)
}()
```

cmd/thv/app/server.go
```diff
@@ -101,10 +127,75 @@ var serveCmd = &cobra.Command{
...
+// initServeOTEL initialises the OTEL provider for thv serve using the global config
+// (set via `thv config otel set-endpoint`). No new CLI flags are introduced; serve reuses
+// the same OTEL config as thv run. Any span processors registered via
+// telemetry.RegisterSpanProcessor (e.g. by sentrypkg.Init) are automatically included.
+// Returns true when OTEL HTTP middleware should be enabled on the API server.
+func initServeOTEL(ctx context.Context, _ bool) (bool, error) {
+	configProvider := cfg.NewDefaultProvider()
+	appConfig := configProvider.GetConfig()
+
+	otelCfg := appConfig.OTEL
```
ChrisJBurns: **[Bug + Security - Medium]** Two issues with the Sentry flags:

1. **Misleading env var hints:** The description says `(env: SENTRY_DSN)` but there's no code to read from environment variables. This will confuse users in production deployments where env vars are the standard configuration mechanism. Either add env var binding or remove the hints.

2. **DSN exposed in process listings:** The DSN contains a key that allows event injection into the Sentry project. Passing it as a CLI flag makes it visible via `ps aux` and `/proc/<pid>/cmdline`. Prefer env var or file-based configuration for the DSN.

cmd/thv/app/server.go
```diff
@@ -101,10 +127,75 @@ var serveCmd = &cobra.Command{
...
+		telemetryCfg.SamplingRate = "0.05"
+	}
+
+	// No OTLP endpoint but registered processors are active (e.g. a Sentry bridge).
+	// Force tracing on with 100% OTEL sampling so every span reaches the processors.
+	// Each processor's own sampling config is then the sole gate.
+	if otelCfg.Endpoint == "" && hasRegisteredProcessors {
+		telemetryCfg.TracingEnabled = true
+		telemetryCfg.SamplingRate = "1.0"
+	}
```
ChrisJBurns: **[Performance - Low]** Setting 100% OTEL sampling in Sentry-only mode means the OTEL SDK fully constructs every span (allocating `ReadWriteSpan`, running through the sampler, calling `OnStart` on all processors) even though Sentry's own `TracesSampleRate` may drop most of them. At high RPS this overhead is non-trivial.

The architecture is functionally correct — the Sentry SDK's own `TracesSampleRate` is the effective gate (confirmed by tracing through `sentry-go` source). But consider documenting this overhead or mirroring the Sentry sample rate for the OTEL sampler.

pkg/recovery/recovery.go
```diff
@@ -33,6 +36,9 @@ func Middleware(next http.Handler) http.Handler {
 				}
 				stack := debug.Stack()
 				slog.Error(fmt.Sprintf("Panic recovered: %v\nStack trace:\n%s", rec, stack))
+				span := trace.SpanFromContext(r.Context())
+				span.RecordError(fmt.Errorf("%v", rec), trace.WithStackTrace(true))
+				span.SetStatus(codes.Error, fmt.Sprintf("panic: %v", rec))
```
ChrisJBurns: **[Security - Medium]** `fmt.Errorf("%v", rec)` with `trace.WithStackTrace(true)` sends the raw panic value and Go stack trace to external telemetry backends. The `rec` value can contain arbitrary data — including error messages that embed credentials, file paths, or internal state. Stack traces may include function arguments with secret values.

The detailed panic is already logged locally via `slog.Error` on the line above.

**Suggestion:** Use a generic message on the span to avoid leaking sensitive data:
```go
span.RecordError(fmt.Errorf("panic recovered"))
span.SetStatus(codes.Error, "panic recovered")
```

**Also:** The `CaptureException`/`RecoverPanic` functions in `pkg/sentry/sentry.go` are never called here. If those are intended for Sentry error tracking (separate from performance monitoring), they need to be wired in explicitly — the OTEL span processor alone only handles transactions, not standalone Sentry error events.

pkg/api/errors/handler.go
```diff
@@ -42,6 +45,9 @@ func ErrorHandler(fn HandlerWithError) http.HandlerFunc {
 		// For 5xx errors, log the full error but return a generic message
 		if code >= http.StatusInternalServerError {
 			slog.Error("internal server error", "error", err)
+			span := trace.SpanFromContext(r.Context())
+			span.RecordError(err)
```
ChrisJBurns: **[Security - Medium]** `span.RecordError(err)` and `span.SetStatus(codes.Error, err.Error())` send the full unwrapped error chain to the OTEL backend. While the project convention avoids sensitive data in errors, defense in depth matters — errors from underlying libraries (database drivers, HTTP clients, container runtimes) can include connection strings, hostnames, or partial credentials.

The existing code correctly withholds 5xx details from the HTTP client (line 51: `http.StatusText(code)`), but this PR sends the unfiltered error to external telemetry.

**Also:** The `sentrypkg.CaptureException(r, err)` function exists but is never called here. Same as with recovery middleware — if standalone Sentry error events are desired, they need explicit calls.

pkg/telemetry/providers/otlp/tracing.go
```diff
@@ -34,26 +34,39 @@ func createTraceExporter(ctx context.Context, config Config) (sdktrace.SpanExpor
...
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to create trace provider: %w", err)
-	}
-
-	provider := sdktrace.NewTracerProvider(
-		sdktrace.WithBatcher(exporter),
+	opts := []sdktrace.TracerProviderOption{
 		sdktrace.WithResource(res),
 		sdktrace.WithSampler(sdktrace.TraceIDRatioBased(config.SamplingRate)),
-	)
```
ChrisJBurns, line 56: **[Bug - Medium]** Bare `TraceIDRatioBased` is a root-only sampler. When a remote parent exists with `sampled=true` (via the W3C `traceparent` header from ToolHive Studio), this sampler may still drop the span if the trace ID doesn't fall within the ratio. This **breaks distributed tracing correlation**.

The correct pattern is:
```go
sdktrace.WithSampler(sdktrace.ParentBased(sdktrace.TraceIDRatioBased(config.SamplingRate)))
```

This is pre-existing, but now materially impactful since this PR adds `otelhttp` traceparent extraction. Without `ParentBased`, the advertised Studio end-to-end tracing won't work reliably at sampling rates below 1.0.

pkg/telemetry/registry.go
```diff
@@ -0,0 +1,57 @@
...
+	globalProcessors   []sdktrace.SpanProcessor
+	globalProcessorsMu sync.Mutex
+)
+
+// RegisterSpanProcessor registers an extra OTEL span processor to be included
+// in any provider created via NewProvider. This allows optional integrations
+// (e.g. a Sentry bridge, Datadog exporter) to self-register during their own
+// Init without coupling to the caller that creates the OTEL provider.
+func RegisterSpanProcessor(p sdktrace.SpanProcessor) {
+	if p == nil {
```
ChrisJBurns, line 29: **[Design - Low]** Two concerns with the global registry:

1. **No dedup guard:** If `sentry.Init()` is called twice (e.g., config reload, tests), `RegisterSpanProcessor` appends the same `sentryotel.NewSentrySpanProcessor()` singleton pointer twice. This causes `OnStart`/`OnEnd` to fire twice per span.

2. **Init ordering dependency:** `sentry.Init()` must be called *before* `telemetry.NewProvider()` for the processor to be picked up. This ordering is correct in the current code but is implicit and fragile. Consider documenting the ordering requirement, or adding a dedup check:
```go
func RegisterSpanProcessor(p sdktrace.SpanProcessor) {
    if p == nil { return }
    globalProcessorsMu.Lock()
    defer globalProcessorsMu.Unlock()
    for _, existing := range globalProcessors {
        if existing == p { return } // already registered
    }
    globalProcessors = append(globalProcessors, p)
}
```

pkg/telemetry/config.go
```diff
@@ -236,7 +237,8 @@ type Provider struct {
 }
 
 // NewProvider creates a new OpenTelemetry provider with the given configuration.
-func NewProvider(ctx context.Context, config Config) (*Provider, error) {
+// Optional extra span processors (e.g. a Sentry bridge) can be registered via extraProcessors.
+func NewProvider(ctx context.Context, config Config, extraProcessors ...sdktrace.SpanProcessor) (*Provider, error) {
 	// Validate configuration
```
ChrisJBurns, line 242: **[Testing]** The new `extraProcessors ...sdktrace.SpanProcessor` parameter and the `registeredSpanProcessors()` merge on the lines below are never tested end-to-end. No test calls `RegisterSpanProcessor(p)` followed by `NewProvider(ctx, config)` and then verifies the processor ends up in the SDK tracer.

The strategy tests set `ExtraSpanProcessors` directly on the `Config` struct, and `TestInit_RegistersSpanProcessor` only checks `HasRegisteredSpanProcessors()` — the complete pipeline is untested.

pkg/sentry/sentry_test.go
```diff
@@ -0,0 +1,154 @@
...
+			TracesSampleRate: 1.0,
+		})
+		require.NoError(t, err)
+		defer func() {
+			initialized.Store(false)
+			telemetry.ResetSpanProcessorsForTesting()
+		}()
+
+		assert.True(t, telemetry.HasRegisteredSpanProcessors())
+	})
```
ChrisJBurns, line 112: **[Testing - Medium]** Only the guard clauses are tested here (not-initialized, nil-error). The actual `hub.CaptureException(err)` call — the primary error-reporting path and the whole point of the Sentry integration — is never reached in any test.

**Suggestion:** Add a happy-path test using `sentry-go`'s test transport:
```go
t.Run("captures exception when initialized", func(t *testing.T) {
    initialized.Store(false)
    transport := &sentry.TransportMock{}
    err := sentry.Init(sentry.ClientOptions{
        Dsn:       "https://key@o0.ingest.sentry.io/0",
        Transport: transport,
    })
    require.NoError(t, err)
    initialized.Store(true)
    defer initialized.Store(false)
    
    req := httptest.NewRequest(http.MethodGet, "/", nil)
    CaptureException(req, errors.New("test error"))
    
    require.Equal(t, 1, len(transport.Events()))
})
```

pkg/sentry/sentry_test.go
```diff
@@ -0,0 +1,154 @@
...
+		initialized.Store(false)
+		req := httptest.NewRequest(http.MethodGet, "/", nil)
+		CaptureException(req, errors.New("test error"))
+	})
+
+	t.Run("no-op with nil error", func(_ *testing.T) {
+		initialized.Store(true)
+		defer initialized.Store(false)
+		req := httptest.NewRequest(http.MethodGet, "/", nil)
+		CaptureException(req, nil)
```
ChrisJBurns, line 127: **[Testing - Medium]** Same gap as `CaptureException` above — `RecoverPanic` happy path (initialized=true, non-nil recovered value) is never tested. The `hub.RecoverWithContext()` and `hub.Flush()` calls are completely unreached.

Also note: the hub fallback path (`GetHubFromContext` returns nil, clone `CurrentHub()`) — which is the *only* production path since the PR doesn't use `sentryhttp` middleware — is never exercised.

pkg/sentry/sentry.go
samuv: Fixed, set sendDefaultPii: false. No beforeSend callback needed for now, since we’re using generic messages on spans anyway (see the error handler / recovery comments below).

I believe Sentry already strips out sensitive data, but we don’t need that level of detail. We already collect some PII on the Studio side (mainly geolocation and OS information), which are the most relevant for us.

pkg/sentry/sentry.go
samuv: Fixed, renamed to `custom.instance_id`. Added a code comment noting that ToolHive Studio currently uses custom.user_id for the same value and should be aligned in a follow-up.

It’s not a problem right now, but you’re right, it should represent that consistently. The only downside is that I’d need to refactor all the dashboards. I’ll take care of that in the future; for now, it’s okay to have this drift.

pkg/api/server.go
samuv: Fixed, removed WithSpanNameFormatter entirely. Added chiRouteSpanNamer, a post-routing middleware that calls next.ServeHTTP first and then renames the span once RouteContext.RoutePattern() is populated. Spans now correctly show as `GET /api/v1beta/workloads/{name}.`

pkg/api/server.go
samuv: Acknowledged. Moving OTEL after auth would break the span lifetime guarantee for the recovery middleware (the span must be active when a panic is caught). Partially mitigated by switching to the ParentBased sampler,  see the sampler comment below. Added a code comment documenting the trade-off.

pkg/api/server.go
samuv: Fixed — removed otelEnabled from Serve(). The cmd layer now uses ServerBuilder directly and calls WithOtelEnabled(otelEnabled), keeping Serve()'s public signature clean.

cmd/thv/app/server.go
samuv: Fixed, extracted to pkg/telemetry/NewServeProvider(ctx) following the same pattern as NewProvider. The cmd layer is now a thin wrapper.

cmd/thv/app/server.go
samuv: Fixed. replaced the goroutine-based shutdown with deterministic LIFO defer ordering: the OTEL provider (which flushes the Sentry span processor) defers first, sentrypkg.Close() defers second. This guarantees the span processor flushes before the Sentry client tears down.

cmd/thv/app/server.go
samuv: Fixed, the code now reads SENTRY_DSN and SENTRY_ENVIRONMENT from environment variables as fallback when the CLI flags are empty. Flag descriptions updated to say "falls back to SENTRY_DSN env var" instead of the misleading (env: SENTRY_DSN).

cmd/thv/app/server.go
samuv: Documented in a code comment in NewServeProvider. The Sentry SDK's own TracesSampleRate remains the effective gate, this is an acceptable trade-off for Sentry-only mode where no external collector is running.

pkg/recovery/recovery.go
samuv: Fixed in both files, replaced fmt.Errorf("%v", rec) and span.RecordError(err) with generic messages ("panic recovered" / "internal server error"). Full details remain in the local slog.Error output only.

Also re-added explicit sentrypkg.RecoverPanic / sentrypkg.CaptureException calls alongside the OTEL span calls, as you noted, the Sentry span processor only creates transactions, not Issues. Both calls are needed for errors to appear in the Sentry Issues tab.

cc @reyortiz3

pkg/telemetry/providers/otlp/tracing.go
samuv: Fixed, wrapped with sdktrace.ParentBased(sdktrace.TraceIDRatioBased(...)). Incoming traceparent headers with sampled=true from ToolHive Studio are now always honoured regardless of the local sampling ratio.

pkg/telemetry/registry.go
samuv: Fixed,added a pointer-equality check in RegisterSpanProcessor so duplicate registrations are silently ignored. Added a doc comment documenting the ordering requirement (processors must register before NewProvider is called).

pkg/telemetry/config.go
samuv: Added TestNewProvider_PicksUpRegisteredProcessor in pkg/telemetry/registry_test.go, registers a tracetest.SpanRecorder, calls NewProvider, creates a span, and asserts the recorder received the OnEnd callback. Also added TestRegisterSpanProcessor_Dedup to cover the new dedup guard.

pkg/sentry/sentry_test.go
samuv: Added captures exception when initialized sub-test using gosentry.MockTransport, verifies a call to CaptureException results in exactly one event on the transport after Flush.

pkg/sentry/sentry_test.go
samuv: Added recovers panic and creates Sentry event sub-test using gosentry.MockTransport, verifies the hub.RecoverWithContext + hub.Flush path creates an event, including the hub fallback path (clone CurrentHub()) which is the only production path without sentryhttp

pkg/api/errors/handler.go
samuv: Fixed both points.

For the unfiltered error chain: span.RecordError now uses a generic fmt.Errorf("internal server error") instead of the raw err, so no error details from underlying libraries reach external backends. The full error is still logged locally via slog.Error.

For the missing CaptureException call: re-added sentrypkg.CaptureException(r, err) alongside the OTEL span call. As you correctly noted, the Sentry span processor only creates transactions — explicit hub calls are required for errors to appear in the Sentry Issues tab. Same fix applied to the recovery middleware with sentrypkg.RecoverPanic.

pkg/api/server.go
samuv: Thinking more about this, I also added URL path parameters as span attributes in the same middleware. This way, the concrete MCP name (and any other route params) are still visible at the trace level without increasing cardinality:

- Span name → GET /api/v1beta/workloads/{name} (low cardinality, used for grouping)
- Span attribute → url.path_param.name = "my-server" (high cardinality, for per-trace drill-down)

This is implemented generically in chiRouteSpanNamer by iterating over rctx.URLParams.Keys/Values, so all parameterized routes automatically record their path params, no per-route code needed.

Addressed [here](https://github.com/stacklok/toolhive/pull/4365/commits/d8b3bccd520d01a0ea4609acf14a5a4c9e0ede3f)

samuv:
@claude review this PR

ChrisJBurns (approved):

PR #4372, opened 2026-03-26
Webhook Middleware Phase 3: Mutating webhook middleware with JSONPatch

### LINKED ISSUE #3398: Webhook Middleware Phase 3: Mutating webhook middleware with JSONPatch
## Overview

Implement mutating webhook middleware that calls external HTTP services to transform MCP requests using JSONPatch (RFC 6902). This enables organizations to enrich requests with data from external sources (HR systems, CMDB, project databases) without modifying ToolHive code.

**Depends on**: Phase 1 (Core webhook package)

## Files to Create

| File | Purpose |
|------|---------|
| `pkg/webhook/mutating/middleware.go` | Mutating webhook middleware implementation |
| `pkg/webhook/mutating/config.go` | Configuration types and validation |
| `pkg/webhook/mutating/patch.go` | JSONPatch application and validation |

## Files to Modify

| File | Changes |
|------|---------|
| `pkg/runner/middleware.go` | Register `mutating-webhook` in `GetSupportedMiddlewareFactories()` |
| `pkg/runner/config.go` | Add `MutatingWebhooks []webhook.WebhookConfig` field to RunConfig |

## Middleware Implementation

```go
// pkg/webhook/mutating/middleware.go
const MiddlewareType = "mutating-webhook"

type MiddlewareParams struct {
    Webhooks []webhook.WebhookConfig `json:"webhooks"`
}

func CreateMiddleware(config *types.MiddlewareConfig, runner types.MiddlewareRunner) error {
    // 1. Parse params
    // 2. Create HTTP client
    // 3. Build middleware function that:
    //    - Gets parsed MCP request from context
    //    - Builds webhook request
    //    - Calls each webhook in order
    //    - Applies JSONPatch operations to mcp_request
    //    - Updates request context/body with mutated request
    //    - Handle failures per FailurePolicy
    // 4. Register with runner.AddMiddleware()
}
```

## Request/Response Format

**Request to webhook** (POST):
```json
{
  "version": "v0.1.0",
  "uid": "unique-request-id",
  "timestamp": "2025-01-22T10:30:00Z",
  "principal": {
    "sub": "user123",
    "email": "user@example.com"
  },
  "mcp_request": {
    "jsonrpc": "2.0",
    "id": 1,
    "method": "tools/call",
    "params": {
      "name": "database_query",
      "arguments": {
        "query": "SELECT * FROM users"
      }
    }
  },
  "context": {
    "server_name": "my-mcp-server",
    "transport": "sse"
  }
}
```

**Response (with JSONPatch)**:
```json
{
  "version": "v0.1.0",
  "uid": "unique-request-id",
  "allowed": true,
  "patch_type": "json_patch",
  "patch": [
    {
      "op": "add",
      "path": "/mcp_request/params/arguments/audit_user",
      "value": "user@example.com"
    },
    {
      "op": "add",
      "path": "/mcp_request/params/arguments/department",
      "value": "engineering"
    }
  ]
}
```

## JSONPatch Implementation

Use `github.com/evanphx/json-patch/v5` library for RFC 6902 compliance.

**Supported operations**:
- `add` - Add value at path
- `remove` - Remove value at path
- `replace` - Replace value at path
- `copy` - Copy value from one path to another
- `move` - Move value from one path to another
- `test` - Test value at path (for conditional patches)

**Security constraint**: Patches are scoped to the `mcp_request` container only. Attempts to modify `principal`, `context`, or other fields should be rejected.

```go
// pkg/webhook/mutating/patch.go
func ApplyPatch(original []byte, patch []JSONPatchOp) ([]byte, error)
func ValidatePatch(patch []JSONPatchOp) error
func IsPatchScopedToMCPRequest(patch []JSONPatchOp) bool
```

## Middleware Chain Position

Mutating webhooks should be placed after MCP Parser and before Validating webhooks:

```
Auth -> Token Exchange -> Tool Filter -> MCP Parser ->
[Mutating Webhooks] -> [Validating Webhooks] -> Telemetry -> Authorization -> Audit -> Recovery
```

## Failure Policies

| Policy | Behavior on webhook error |
|--------|---------------------------|
| `fail` (fail-closed) | Deny request with 500 |
| `ignore` (fail-open) | Use original (unmutated) request |

## Multiple Webhooks

When multiple mutating webhooks are configured:
1. Execute in configuration order
2. Each webhook receives the output of the previous mutation
3. Patches are applied sequentially
4. Failure policy applies per-webhook

## Tests

- JSONPatch application tests for all operations
- Tests for invalid/malformed patches
- Tests for patch scope validation (reject modifications outside mcp_request)
- Tests for multiple webhooks with chained mutations
- Tests for failure policies
- Integration tests with middleware chain

## Acceptance Criteria

- [ ] Middleware registered in `GetSupportedMiddlewareFactories()`
- [ ] Correctly calls webhooks with RFC-compliant request format
- [ ] Applies JSONPatch operations correctly
- [ ] Validates patch scope (mcp_request only)
- [ ] Implements both failure policies
- [ ] Supports multiple webhooks with chained mutations
- [ ] Comprehensive unit tests with >80% coverage

Sanskarzz:
## Summary
Implements **Phase 3** of the Dynamic Webhook Middleware system: **Mutating Webhook Middleware**, as outlined in RFC THV-0017.

This PR adds support for mutating webhooks that allow external HTTP services to transform incoming MCP requests dynamically using RFC 6902 JSON Patch operations. Mutating webhooks are executed after the MCP parser but before validating webhooks, allowing organizations to cleanly intercept and rewrite requests (e.g., adding default parameters, applying transformations, or injecting audit trails) before validation or authorization logic runs.

---

## Large PR Justification

This is a new feature package with a large test suite, and it needs to land as one coherent phase so the middleware wiring, patch handling, and coverage stay consistent.

---

Fixes #3398 

### 1. `pkg/webhook/mutating/` Package
Introduces the core logic for mutating webhooks:
- **`config.go`**: Defines configuration parameter types (`MiddlewareParams`, `FactoryMiddlewareParams`) mirroring the validating package.
- **`patch.go`**: Implements RFC 6902 JSONPatch parsing and application using `github.com/evanphx/json-patch/v5`.
  - Includes `ValidatePatch()` to ensure only supported operations are allowed.
  - Implements **Scope Validation (`IsPatchScopedToMCPRequest()`)** to ensure webhooks can only modify the `mcp_request` container inside the webhook payload, preventing unauthorized modification of the request principal, context, or envelope metadata.
- **`middleware.go`**: The HTTP middleware factory and handler.
  - Skips non-MCP requests gracefully.
  - Executes chained mutating webhooks in configuration order, passing the output of one as the input to the next.
  - Extracts the raw MCP request, evaluates it against the webhook endpoints, and applies the returned JSONPatches.
  - Supports RFC-defined failure policies (`fail` vs `ignore`) for connection errors or malformed patches.

### 2. Runner Wiring (`pkg/runner/`)
- **`config.go`**: Added `MutatingWebhooks []webhook.Config` to `RunConfig` to allow configuring mutating webhooks natively via state persistence or configuration YAML.
- **`middleware.go`**: Registered the mutating webhook middleware factory and inserted it into the middleware execution chain. It executes proactively **before** the validating webhook middleware, matching the RFC specifications.

### 3. Dependencies
- Added `github.com/evanphx/json-patch/v5` for RFC 6902-compliant patch application.

### 4. Addressed Phase 2 leftover review comments
- Inlined the variables in `sendErrorResponse` in the validating webhook middleware.
- Renamed the misleading test in `pkg/webhook/validating/middleware_test.go`.
- Extracted duplicate `convertToJSONRPC2ID` implementations from all 3 middlewares (authz, validating, mutating) to a central location (pkg/mcp/utils.go).

### 5. Full Middleware Chain Integration Test
- Added `TestWebhookMiddlewareChainIntegration` to `pkg/runner/webhook_integration_test.go` to test the execution flow of both mutating and validating webhooks as initialized by the runner configuration parsing.

### 6. Test Coverage Improvements:
- Added new tests checking failure states of patch extraction and application to increase `pkg/webhook/mutating/patch.go` and `pkg/webhook/mutating/middleware.go` coverage.
- Addressed low coverage for `PopulateMiddlewareConfigs` in `pkg/runner/middleware.go` by adding `TestPopulateMiddlewareConfigs_FullCoverage` which spans across all configuration branch setups.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

### DIFF
--- a/pkg/webhook/mutating/middleware.go
+++ b/pkg/webhook/mutating/middleware.go
@@ -1,0 +1,314 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package mutating
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"time"
+
+	"github.com/google/uuid"
+	"golang.org/x/exp/jsonrpc2"
+
+	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/mcp"
+	"github.com/stacklok/toolhive/pkg/transport/types"
+	"github.com/stacklok/toolhive/pkg/webhook"
+)
+
+// MiddlewareType is the type constant for the mutating webhook middleware.
+const MiddlewareType = "mutating-webhook"
+
+// Middleware wraps mutating webhook functionality for the factory pattern.
+type Middleware struct {
+	handler types.MiddlewareFunction
+}
+
+// Handler returns the middleware function used by the proxy.
+func (m *Middleware) Handler() types.MiddlewareFunction {
+	return m.handler
+}
+
+// Close cleans up any resources used by the middleware.
+func (*Middleware) Close() error {
+	return nil
+}
+
+type clientExecutor struct {
+	client *webhook.Client
+	config webhook.Config
+}
+
+// CreateMiddleware is the factory function for mutating webhook middleware.
+func CreateMiddleware(config *types.MiddlewareConfig, runner types.MiddlewareRunner) error {
+	var params FactoryMiddlewareParams
+	if err := json.Unmarshal(config.Parameters, &params); err != nil {
+		return fmt.Errorf("failed to unmarshal mutating webhook middleware parameters: %w", err)
+	}
+
+	if err := params.Validate(); err != nil {
+		return fmt.Errorf("invalid mutating webhook configuration: %w", err)
+	}
+
+	// Create clients for each webhook.
+	var executors []clientExecutor
+	for i, whCfg := range params.Webhooks {
+		client, err := webhook.NewClient(whCfg, webhook.TypeMutating, nil) // HMAC secret not yet plumbed
+		if err != nil {
+			return fmt.Errorf("failed to create client for webhook[%d] (%q): %w", i, whCfg.Name, err)
+		}
+		executors = append(executors, clientExecutor{client: client, config: whCfg})
+	}
+
+	mw := &Middleware{
+		handler: createMutatingHandler(executors, params.ServerName, params.Transport),
+	}
+	runner.AddMiddleware(MiddlewareType, mw)
+	return nil
+}
+
+func createMutatingHandler(executors []clientExecutor, serverName, transport string) types.MiddlewareFunction {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Skip if it's not a parsed MCP request (middleware runs after mcp parser).
+			parsedMCP := mcp.GetParsedMCPRequest(r.Context())
+			if parsedMCP == nil {
+				next.ServeHTTP(w, r)
+				return
+			}
+
+			// Read the request body to get the raw MCP request.
+			bodyBytes, err := io.ReadAll(r.Body)
+			if err != nil {
+				sendErrorResponse(w, http.StatusInternalServerError, "Failed to read request body", parsedMCP.ID)
+				return
+			}
+			// Restore the request body immediately; we will replace it after mutations.
+			r.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+
+			// currentMCPBody is the MCP JSON-RPC body we thread through the webhook chain.
+			// Each successful mutation replaces this with the patched version.
+			currentMCPBody := bodyBytes
+
+			// Build the base webhook request context (reused across all webhooks).
+			reqContext := &webhook.RequestContext{
+				ServerName: serverName,
+				SourceIP:   readSourceIP(r),
+				Transport:  transport,
+			}
+
+			// Resolve principal once (same for all webhooks in this chain).
+			var principal *auth.PrincipalInfo
+			if identity, ok := auth.IdentityFromContext(r.Context()); ok {
+				principal = identity.GetPrincipalInfo()
+			}
+
+			// Execute the webhook chain to apply mutations.
+			mutatedBody, err := executeMutations(r.Context(), executors, currentMCPBody, reqContext, principal, parsedMCP.ID, w)
+			if err != nil {
+				// executeMutations handles writing the error to the response implicitly when it returns an error.
+				return
+			}
+
+			// Replace the request body with the (potentially mutated) MCP body for downstream handlers.
+			r.Body = io.NopCloser(bytes.NewBuffer(mutatedBody))
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+// executeMutations runs the chain of mutating webhooks sequentially.
+// It returns the final mutated body, or an error if the chain was aborted.
+// If an error occurs that should abort the request, this function writes the error response.
+func executeMutations(
+	ctx context.Context,
+	executors []clientExecutor,
+	initialBody []byte,
+	reqContext *webhook.RequestContext,
+	principal *auth.PrincipalInfo,
+	msgID interface{},
+	w http.ResponseWriter,
+) ([]byte, error) {
+	currentBody := initialBody
+
+	for _, exec := range executors {
+		mutatedBody, err := executeSingleMutation(ctx, exec, currentBody, reqContext, principal, msgID, w)
+		if err != nil {
+			return nil, err
+		}
+		currentBody = mutatedBody
+	}
+
+	return currentBody, nil
+}
+
+// executeSingleMutation applies a single mutating webhook.
+func executeSingleMutation(
+	ctx context.Context,
+	exec clientExecutor,
+	currentBody []byte,
+	reqContext *webhook.RequestContext,
+	principal *auth.PrincipalInfo,
+	msgID interface{},
+	w http.ResponseWriter,
+) ([]byte, error) {
+	whName := exec.config.Name
+
+	whReq := &webhook.Request{
+		Version:    webhook.APIVersion,
+		UID:        uuid.New().String(),
+		Timestamp:  time.Now().UTC(),
+		MCPRequest: json.RawMessage(currentBody),
+		Context:    reqContext,
+		Principal:  principal,
+	}
+
+	resp, err := exec.client.CallMutating(ctx, whReq)
+	if err != nil {
+		if exec.config.FailurePolicy == webhook.FailurePolicyIgnore {
+			slog.Warn("Mutating webhook error ignored due to fail-open policy", "webhook", whName, "error", err)
+			return currentBody, nil
+		}
+		slog.Error("Mutating webhook error caused request denial", "webhook", whName, "error", err)
+		sendErrorResponse(w, http.StatusInternalServerError, "Webhook error", msgID)
+		return nil, err
+	}
+
+	// Explicit denial from a mutating webhook is always honored, regardless of failure policy.
+	// This differs from operational errors (network, timeout) where the failure policy applies.
+	if !resp.Allowed {
+		slog.Info("Mutating webhook denied request", "webhook", whName, "reason", resp.Reason)
+		sendErrorResponse(w, http.StatusForbidden, "Request denied by webhook policy", msgID)
+		return nil, fmt.Errorf("webhook denied request")
+	}
+
+	if resp.PatchType == "" || len(resp.Patch) == 0 {
+		return currentBody, nil
+	}
+
+	if resp.PatchType != patchTypeJSONPatch {
+		slog.Error("Mutating webhook returned unsupported patch type", "webhook", whName, "patch_type", resp.PatchType)
+		if exec.config.FailurePolicy == webhook.FailurePolicyIgnore {
+			return currentBody, nil
+		}
+		sendErrorResponse(w, http.StatusInternalServerError, "Unsupported patch type from webhook", msgID)
+		return nil, fmt.Errorf("unsupported patch type")
+	}
+
+	return applyMutationPatch(resp, whName, exec.config.FailurePolicy, currentBody, msgID, w)
+}
+
+func applyMutationPatch(
+	resp *webhook.MutatingResponse,
+	whName string,
+	failurePolicy webhook.FailurePolicy,
+	currentBody []byte,
+	msgID interface{},
+	w http.ResponseWriter,
+) ([]byte, error) {
+	var patchOps []JSONPatchOp
+	if err := json.Unmarshal(resp.Patch, &patchOps); err != nil {
+		slog.Error("Mutating webhook returned malformed patch", "webhook", whName, "error", err)
+		if failurePolicy == webhook.FailurePolicyIgnore {
+			return currentBody, nil
+		}
+		sendErrorResponse(w, http.StatusInternalServerError, "Malformed patch from webhook", msgID)
+		return nil, err
+	}
+
+	if err := ValidatePatch(patchOps); err != nil {
+		slog.Error("Mutating webhook patch failed validation", "webhook", whName, "error", err)
+		if failurePolicy == webhook.FailurePolicyIgnore {
+			return currentBody, nil
+		}
+		sendErrorResponse(w, http.StatusInternalServerError, "Invalid patch from webhook", msgID)
+		return nil, err
+	}
+
+	if !IsPatchScopedToMCPRequest(patchOps) {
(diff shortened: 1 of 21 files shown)


Sanskarzz:
@JAORMX  The PR is Ready for review.

JAORMX:
Amazing! I'll check it out on Monday. Had a day off on Friday, that's why I hadn't given feedback.

pkg/webhook/mutating/middleware.go
```diff
@@ -0,0 +1,310 @@
...
+			return currentBody, nil
+		}
+		slog.Error("Mutating webhook error caused request denial", "webhook", whName, "error", err)
+		sendErrorResponse(w, http.StatusInternalServerError, "Webhook error", msgID)
+		return nil, err
+	}
+
+	if !resp.Allowed {
+		slog.Info("Mutating webhook denied request", "webhook", whName, "reason", resp.Reason)
+		sendErrorResponse(w, http.StatusInternalServerError, "Request mutation denied by webhook", msgID)
```
JAORMX: blocker: So... when a webhook explicitly says `allowed: false`, that's a policy decision, not a server error. Right now this returns HTTP 500, but the validating middleware uses 403 for the same scenario (and rightly so). Using 500 here would confuse anyone triaging alerts... \"is the webhook broken, or did it deliberately deny the request?\"\n\nThe RFC failure modes table doesn't have a specific row for `allowed: false` on mutating webhooks, but semantically 403 is the right call here.\n\n```suggestion\n\t\tsendErrorResponse(w, http.StatusForbidden, \"Request denied by webhook policy\", msgID)\n```\n\nNote that if we change this, the `//nolint:unparam` on `sendErrorResponse` becomes justified since `statusCode` will no longer always be 500.

pkg/webhook/mutating/middleware.go
```diff
@@ -0,0 +1,310 @@
...
+	if !IsPatchScopedToMCPRequest(patchOps) {
+		slog.Error("Mutating webhook patch targets fields outside mcp_request — rejected", "webhook", whName)
+		if failurePolicy == webhook.FailurePolicyIgnore {
+			return currentBody, nil
+		}
+		sendErrorResponse(w, http.StatusInternalServerError, "Patch must be scoped to mcp_request", msgID)
+		return nil, fmt.Errorf("patch scope violation")
+	}
+
+	envelopeJSON, err := json.Marshal(whReq)
```
JAORMX: suggestion: This marshals the *entire* webhook request (including `principal` with email/subject/groups and `context` with source IP) into the envelope just so the patch paths like `/mcp_request/...` resolve correctly.\n\nWhile `IsPatchScopedToMCPRequest` prevents patches from *targeting* those fields, the sensitive data is still sitting in the document being patched. If the json-patch library ever surfaces document content in error messages, or if there's a bug in scope validation, the principal data could leak or be modified.\n\nA safer approach... construct a minimal envelope with only what the patches need:\n\n```go\nenvelope := map[string]json.RawMessage{\"mcp_request\": json.RawMessage(currentBody)}\nenvelopeJSON, err := json.Marshal(envelope)\n```\n\nThis way, even if scope validation has a bug, there's nothing sensitive to modify. Defense in depth!

pkg/webhook/mutating/middleware.go
```diff
@@ -0,0 +1,310 @@
...
+		UID:        uuid.New().String(),
+		Timestamp:  time.Now().UTC(),
+		MCPRequest: json.RawMessage(currentBody),
+		Context:    reqContext,
+		Principal:  principal,
+	}
+
+	resp, err := exec.client.CallMutating(ctx, whReq)
+	if err != nil {
+		if exec.config.FailurePolicy == webhook.FailurePolicyIgnore {
```
JAORMX, line 174: suggestion: Per RFC THV-0017's failure modes table, HTTP 422 from a webhook should **always deny** regardless of failure policy (`Deny (422)` for both `fail` and `ignore` columns). The client already surfaces the status code in `InvalidResponseError` (there's even a comment about it in `client.go:147`), but the middleware treats all errors uniformly via the failure policy.\n\nWith `ignore` policy, a 422 would currently result in using the original body instead of denying... which the RFC says shouldn't happen.\n\nI know this is also a pre-existing gap in the validating middleware from Phase 2, so it doesn't have to be fixed in this PR. But could you file a follow-up issue tracking the 422 handling for both middleware types?

pkg/webhook/mutating/middleware.go
```diff
@@ -0,0 +1,310 @@
...
+			slog.Warn("Mutating webhook error ignored due to fail-open policy", "webhook", whName, "error", err)
+			return currentBody, nil
+		}
+		slog.Error("Mutating webhook error caused request denial", "webhook", whName, "error", err)
+		sendErrorResponse(w, http.StatusInternalServerError, "Webhook error", msgID)
+		return nil, err
+	}
+
+	if !resp.Allowed {
```
JAORMX, line 185: suggestion: One more thing here... when `Allowed` is `false`, we always deny regardless of failure policy. That seems intentional (an explicit deny is different from an operational error), and I think it's the right call. But it's worth adding a short comment explaining the design decision, since the RFC failure modes table doesn't address this case for mutating webhooks. Something like:\n\n```suggestion\n\t// Explicit denial from a mutating webhook is always honored, regardless of failure policy.\n\t// This differs from operational errors (network, timeout) where the failure policy applies.\n\tif !resp.Allowed {\n```

pkg/webhook/mutating/patch.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	jsonpatch "github.com/evanphx/json-patch/v5"
+)
+
+// patchTypJSONPatch is the patch_type value for RFC 6902 JSON Patch.
```
JAORMX: nitpick: Tiny typo in the comment... `patchTypJSONPatch` is missing the 'e' in `Type`.\n\n```suggestion\n// patchTypeJSONPatch is the patch_type value for RFC 6902 JSON Patch.\n```

pkg/webhook/mutating/patch.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	return nil
+}
+
+// IsPatchScopedToMCPRequest returns true if all patch operations target paths
+// within the mcp_request container. This prevents webhooks from accidentally
+// or maliciously modifying principal, context, or other immutable envelope fields.
+func IsPatchScopedToMCPRequest(patch []JSONPatchOp) bool {
+	for _, op := range patch {
+		if !strings.HasPrefix(op.Path, mcpRequestPathPrefix) {
```
JAORMX, line 68: question: The prefix check requires `/mcp_request/` (with trailing slash). So a patch targeting `/mcp_request` exactly (no trailing slash, e.g. a `replace` on the entire object) would be rejected.\n\nIs that intentional? It means a webhook can't do a full replacement of the MCP request body via a single `replace` op on `/mcp_request`. I think rejecting it is actually the safer default (forces granular patches), but wanted to confirm that's the intent and not an accidental off-by-one in the prefix.

pkg/webhook/mutating/middleware.go
```diff
@@ -0,0 +1,310 @@
...
+	"github.com/stacklok/toolhive/pkg/mcp"
+	"github.com/stacklok/toolhive/pkg/transport/types"
+	"github.com/stacklok/toolhive/pkg/webhook"
+)
+
+// MiddlewareType is the type constant for the mutating webhook middleware.
+const MiddlewareType = "mutating-webhook"
+
+// Middleware wraps mutating webhook functionality for the factory pattern.
+type Middleware struct {
```
JAORMX, line 29: question: Both the mutating and validating middlewares have nearly identical `Middleware` struct, `clientExecutor` type, `CreateMiddleware` factory, `sendErrorResponse`, and `readSourceIP`. I totally get keeping the packages independent (easier to reason about, no coupling), but I'm curious... was this a deliberate choice?\n\nIf more webhook types come in later phases, we might end up with N copies. Might be worth extracting the common scaffold to `pkg/webhook/` at some point. Not asking for it in this PR though... just want to know if there was a design rationale here.

pkg/webhook/mutating/middleware.go
Sanskarzz: Done

pkg/webhook/mutating/middleware.go
Sanskarzz: Done

pkg/webhook/mutating/middleware.go
Sanskarzz: You're right that this is a gap relative to the RFC failure-modes table, and it applies to both mutating and validating middleware.

I’m not fixing it in this PR since it is broader than Phase 3 and the same behavior already exists in the Phase 2 validating path, but I’ll track it in a follow-up issue so we can handle webhook HTTP 422 as an unconditional deny regardless of failure policy in both middleware types.

pkg/webhook/mutating/middleware.go
Sanskarzz: Added

pkg/webhook/mutating/patch.go
Sanskarzz: Fixed

pkg/webhook/mutating/patch.go
Sanskarzz: Yes, that is intentional. I want to reject patching `/mcp_request` itself and only allow mutations beneath `/mcp_request/...`.

The reasoning is the same as you called out: forcing granular edits is the safer default, and it avoids a webhook replacing the entire MCP request object in one shot. I added a clarifying comment in `IsPatchScopedToMCPRequest()` and a test case to make that intent explicit rather than leaving it as an accidental-looking prefix check.

pkg/webhook/mutating/middleware.go
Sanskarzz: Yes, that was a deliberate choice for this phase.

I kept the mutating and validating packages structurally similar but independent so each middleware stays easy to reason about while the feature set is still settling. Since Phase 2 and Phase 3 have different execution semantics and failure handling, I preferred a little duplication over prematurely extracting a shared abstraction that might become awkward once more webhook-specific behavior lands.

That said, I agree there is a good future refactor opportunity here. If we add more webhook middleware types or the shared scaffolding grows further, extracting the common pieces into `pkg/webhook/` would make sense.

I think we should implement the refactoring later, after completing Phase 5.

pkg/webhook/mutating/middleware.go
Sanskarzz: I will create a follow-up issue after this PR is submitted. will add fix PR too.

Sanskarzz:
@JAORMX The PR is Ready for re-review.

pkg/webhook/mutating/middleware.go
```diff
@@ -0,0 +1,315 @@
...
+		return nil, fmt.Errorf("mcp_request field missing or empty in patched envelope")
+	}
+	return env.MCPRequest, nil
+}
+
+func readSourceIP(r *http.Request) string {
+	return r.RemoteAddr
+}
+
+//nolint:unparam // statusCode is currently always 500, but kept for API flexibility
```
JAORMX: So... this comment says `statusCode is currently always 500`, but line 187 now passes `http.StatusForbidden` (the fix from last round). The `nolint:unparam` directive is no longer needed and the comment is actively misleading.\n\nJust drop the nolint and the comment:\n\n```suggestion\n\nfunc sendErrorResponse(w http.ResponseWriter, statusCode int, message string, msgID interface{}) {\n```

Sanskarzz:
> Oh, and on the 422 follow-up issue you mentioned you'd file... could you go ahead and create that before this merges? That way we don't lose track of it. Thanks!

I have created the [issue](https://github.com/stacklok/toolhive/issues/4651)

pkg/webhook/mutating/middleware.go
Sanskarzz: done and also squash the commits

JAORMX:
Filed #4663 to track the HTTP 422 unconditional-deny handling for both validating and mutating middleware.

Sanskarzz:
> Filed #4663 to track the HTTP 422 unconditional-deny handling for both validating and mutating middleware.

@JAORMX  Btw, I already created issue #4651. I have closed this.

Sanskarzz:
@JAORMX  Now CI is green. The PR needs approval and merging.

PR #4375, opened 2026-03-26
Add forEach step type for VMCP composite tool workflows

JAORMX:
## Summary

- Workflows today require statically defined steps -- you cannot dynamically spawn steps based on a previous step's output. This makes patterns like "for each repository in an org, check for security advisories" impossible without hardcoding every item.
- Add a third step type `forEach` that iterates over a collection produced by a previous step and executes an inner tool step for each item, with configurable parallelism and error handling.

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/config/config.go` | Add 5 forEach fields to `WorkflowStepConfig`, update kubebuilder enum |
| `pkg/vmcp/config/composite_validation.go` | Add `ValidateForEachStep` with inner step, collection, limits validation |
| `pkg/vmcp/config/foreach_validation_test.go` | 12 validation test cases |
| `pkg/vmcp/composer/composer.go` | Add `StepTypeForEach`, forEach fields on `WorkflowStep`, 2 new `TemplateExpander` methods |
| `pkg/vmcp/composer/template_expander.go` | Unify expansion into single path with optional `extraCtx` for forEach |
| `pkg/vmcp/composer/workflow_engine.go` | Add `executeForEachStep` with collection resolution, bounded parallel execution, result aggregation |
| `pkg/vmcp/composer/foreach_test.go` | 10 execution test cases (basic, empty, abort, continue, max iterations, downstream, parallel, template context, string collection, invalid) |
| `pkg/vmcp/server/workflow_converter.go` | Handle forEach type conversion, default empty Type to "tool" |
| `pkg/vmcp/internal/compositetools/workflow_converter.go` | Check inner step tool accessibility for session filtering |
| `docs/arch/10-virtual-mcp-architecture.md` | Add step type enumeration |
| `docs/operator/composite-tools-quick-reference.md` | Add forEach to type enum, pattern example, validation rules |
| `docs/operator/virtualmcpcompositetooldefinition-guide.md` | Add forEach step type reference section |
| `docs/operator/advanced-workflow-patterns.md` | Add ForEach Iteration Patterns section |

Yes. Users can now define `forEach` steps in composite tool workflows to iterate over collections from previous steps. Example:

```yaml
steps:
  - id: list_repos
    type: tool
    tool: github.list_repos
    arguments:
      org: "{{.params.org}}"

  - id: check_advisories
    type: forEach
    collection: "{{json .steps.list_repos.output.repositories}}"
    itemVar: repo
    maxParallel: 5
    step:
      type: tool
      tool: github.list_security_advisories
      arguments:
        repo: "{{.forEach.repo.name}}"
    dependsOn: [list_repos]
    onError:
      action: continue
```

**Key design decisions:**
- `forEach` is a DAG-opaque macro -- the DAG sees one node; internal parallelism is self-managed via errgroup + semaphore. This avoids rearchitecting the DAG executor.
- Single inner step (tool type only). Multi-step iteration bodies and elicitation inner steps can be added later.
- Nested forEach is explicitly rejected at validation time.
- Template expander was refactored to eliminate code duplication: `expandValueWithDepth` and `expandStringInternal` now accept an optional `extraCtx` parameter, used by both normal and forEach expansion paths.

**Safety limits:**
- `maxIterations`: default 100, hard cap 1000
- `maxParallel`: default 10 (DAG default), hard cap 50
- `itemVar` cannot be `"index"` (reserved key)

**Interface change:** Two methods added to `TemplateExpander` (`ExpandString`, `ExpandWithForEach`). `ExpandString` is a natural complement to `Expand` (string vs map). `ExpandWithForEach` cleanly separates the forEach concern without mutating shared `WorkflowContext` state. Both are backed by the same unified internal implementation.

## Large PR Justification

- The forEach step type spans validation, execution, template expansion, CRD schema, and documentation -- these are tightly coupled and would break if separated (e.g., validation without the execution path, or CRD changes without the converter).
- ~400 lines are tests (`foreach_test.go`, `foreach_validation_test.go`).
- ~200 lines are generated CRD YAML and documentation updates.
- The core implementation logic is ~350 lines across 3 files.

### DIFF
--- a/pkg/vmcp/composer/workflow_engine.go
+++ b/pkg/vmcp/composer/workflow_engine.go
@@ -726,7 +726,7 @@ func (e *workflowEngine) executeForEachStep(
 	}
 	maxPar := step.MaxParallel
 	if maxPar <= 0 {
-		maxPar = e.dagExecutor.maxParallel
+		maxPar = e.dagExecutor.MaxParallel()
 	}
 	// Runtime cap to prevent goroutine/connection exhaustion even if validation is bypassed
 	const runtimeMaxParallel = 50
@@ -746,6 +746,7 @@ func (e *workflowEngine) executeForEachStep(
 			case sem <- struct{}{}:
 				defer func() { <-sem }()
 			case <-gCtx.Done():
+				results[i] = iterationResult{Index: i, Item: item, Status: "cancelled", Error: gCtx.Err().Error()}
 				return gCtx.Err()
 			}
 
--- a/pkg/vmcp/composer/dag_executor.go
+++ b/pkg/vmcp/composer/dag_executor.go
@@ -41,6 +41,11 @@ func newDAGExecutor(maxParallel int) *dagExecutor {
 	}
 }
 
+// MaxParallel returns the configured maximum parallelism for the DAG executor.
+func (d *dagExecutor) MaxParallel() int {
+	return d.maxParallel
+}
+
 // executionLevel represents a group of steps that can be executed in parallel.
 type executionLevel struct {
 	steps []*WorkflowStep


pkg/vmcp/composer/workflow_engine.go
```diff
@@ -680,6 +685,261 @@ func (e *workflowEngine) executeElicitationStep(
...
+	results := make([]iterationResult, len(items))
+	sem := make(chan struct{}, maxPar)
+	g, gCtx := errgroup.WithContext(ctx)
+
+	for i, item := range items {
+		g.Go(func() error {
+			select {
+			case sem <- struct{}{}:
+				defer func() { <-sem }()
+			case <-gCtx.Done():
```
yrobla, line 748: **[Should Fix] Cancelled iterations leave zero-value entries in `results`, producing misleading aggregated output.**

When `errgroup` cancels `gCtx` after the first failure (abort mode), goroutines still waiting on `sem` return early via this branch without writing to `results[i]`. Those slots remain as zero-value `iterationResult{}` — `Index: 0`, `Item: nil`, `Status: ""`. When `buildForEachOutput` processes them:
- each cancelled entry shows `index: 0` regardless of actual position
- `item: null`
- empty `status` falls silently into `failedCount`, inflating the failure count

This makes the diagnostic output misleading when debugging a failed forEach step. Not a correctness issue for the step result itself, but it could confuse operators reading the aggregated output.

Suggested fix — write a sentinel before returning:
```suggestion
			case <-gCtx.Done():
				results[i] = iterationResult{Index: i, Item: item, Status: "cancelled", Error: gCtx.Err().Error()}
				return gCtx.Err()
```

pkg/vmcp/composer/workflow_engine.go
```diff
@@ -680,6 +685,261 @@ func (e *workflowEngine) executeElicitationStep(
...
+	}
+
+	// Resolve configuration defaults
+	itemVar := step.ItemVar
+	if itemVar == "" {
+		itemVar = "item"
+	}
+	maxPar := step.MaxParallel
+	if maxPar <= 0 {
+		maxPar = e.dagExecutor.maxParallel
```
yrobla: **[Nit / Follow-up] Direct access to `dagExecutor.maxParallel` couples forEach to DAG internals.**

`e.dagExecutor.maxParallel` is a private field of a separate component. Both are in the same package so it compiles fine, but it means any refactoring of `dagExecutor` (e.g. making `maxParallel` dynamic or per-workflow) silently affects forEach defaults.

Not blocking — happy for this to be addressed in a follow-up. Consider either a `dagExecutor.MaxParallel()` accessor or storing a `defaultForEachParallel int` on `workflowEngine` at construction time.

pkg/vmcp/composer/workflow_engine.go
JAORMX: Good catch -- fixed in e581b465. Cancelled iterations now write `iterationResult{Index: i, Item: item, Status: "cancelled", Error: gCtx.Err().Error()}` before returning, so the aggregated output shows the correct index, item, and a clear "cancelled" status instead of zero values.

pkg/vmcp/composer/workflow_engine.go
JAORMX: Agreed -- addressed in the same commit (e581b465). Added a `MaxParallel()` accessor on `dagExecutor` and switched `executeForEachStep` to use it instead of reaching into the private field.

PR #4383, opened 2026-03-26
Wire embedded auth server into VirtualMCPServer converter and deployment

### LINKED ISSUE #4284: Wire GenerateAuthServerConfig into VirtualMCPServer deployment controller
## Problem

The VirtualMCPServer deployment controller does not mount embedded auth server secrets (signing keys, HMAC secrets, upstream client secrets) onto the vMCP pod. Without these, the embedded auth server cannot start even though:

1. The CRD field `authServerConfigRef` is accepted and validated (#4140)
2. The reconciler validates the reference and sets conditions (#4143)
3. The converter produces a `RuntimeConfig.AuthServer` from the referenced `MCPExternalAuthConfig` (#4143)
4. The vMCP binary has `RegisterHandlers` to mount AS routes on the mux (#4141)

The deployment controller has an explicit placeholder at `virtualmcpserver_deployment.go:533`:
```go
case mcpv1alpha1.ExternalAuthTypeEmbeddedAuthServer:
    // Embedded auth server secrets are handled separately (via volume mounts, not env vars)
    // Controller integration will be in a future task
    return nil, nil
```

The MCPServer controller already does this correctly at `mcpserver_controller.go:1101` by calling `ctrlutil.GenerateAuthServerConfig()`.

## What needs to happen

Call `ctrlutil.GenerateAuthServerConfig()` in the VirtualMCPServer deployment reconciler when `authServerConfigRef` is set, and add the resulting volumes, volume mounts, and env vars to the vMCP Deployment spec.

## Acceptance Criteria

- [ ] vMCP Deployment includes signing key volume mounts when `authServerConfigRef` is set
- [ ] vMCP Deployment includes auth server env vars (issuer, upstream config, etc.)
- [ ] OIDC discovery (`/.well-known/openid-configuration`) returns 200 on a vMCP with embedded AS
- [ ] JWKS endpoint returns 200
- [ ] Unauthenticated MCP request returns 401 (not 406)

## Key files

- `cmd/thv-operator/controllers/virtualmcpserver_deployment.go` — needs the wiring
- `cmd/thv-operator/pkg/controllerutil/authserver.go` — `GenerateAuthServerConfig()` already exists
- `cmd/thv-operator/controllers/mcpserver_controller.go:1099-1112` — reference implementation

## Parent issues

Part of #4143 (Phase 4: Operator reconciler for embedded AS)
Part of #4120 (vMCP: add embedded authorization server)

jhrozek:
## Summary

- The VirtualMCPServer operator had no way to convert an inline `AuthServerConfig` into an `authserver.RunConfig` or mount auth server secrets onto the vMCP pod. This blocks end-to-end OAuth flows for vMCP.
- Export `BuildAuthServerRunConfig` from `controllerutil` (previously unexported, MCPServer-only) and call it from the vMCP converter, reusing ~450 lines of conversion logic (signing keys, HMAC secrets, upstream providers, Redis/Sentinel storage).
- Operator: serialize RunConfig as a separate ConfigMap key (`authserver-config.yaml`), mount auth server volumes/env vars on the deployment, cross-validate auth server config against backend strategies. Spec validation errors now set a status condition instead of requeueing indefinitely.
- Binary: load sibling `authserver-config.yaml` if present, construct `EmbeddedAuthServer`, fail closed on errors.

Fixes #4284

## Type of change

- [x] New feature

## Test plan

Tested end-to-end as part of the full `vmcp-add-as-scaffolding-5` branch with a deployed VirtualMCPServer running the embedded auth server.

## Changes

| File | Change |
|------|--------|
| `controllerutil/authserver.go` | Export `BuildAuthServerRunConfig`, take explicit `allowedAudiences`/`scopesSupported` params instead of `*oidc.OIDCConfig` |
| `vmcpconfig/converter.go` | `Convert()` returns `(*Config, *RunConfig, error)`; add `convertAuthServerConfig`, `deriveAllowedAudiences`, `deriveScopesSupported` |
| `virtualmcpserver_vmcpconfig.go` | Serialize RunConfig as second ConfigMap key; cross-validate via `ValidateAuthServerIntegration` with status condition on failure |
| `virtualmcpserver_controller.go` | Add `SpecValidationError` type and `handleSpecValidationError` helper for no-requeue on spec errors |
| `virtualmcpserver_deployment.go` | Mount auth server volumes and env vars when `AuthServerConfig` is set |
| `cmd/vmcp/app/commands.go` | `loadAuthServerConfig` loads sibling file; construct `EmbeddedAuthServer` in `runServe` |

- `deriveAllowedAudiences` uses only `Resource` (RFC 8707), no fallback to `Audience` — consistent with the MCPServer path which hard-requires `ResourceURL`.
- `deriveScopesSupported` normalizes empty slices to nil so the auth server always applies its default scopes when none are configured.
- The `SpecValidationError` type distinguishes spec errors (no requeue, set status condition) from transient errors (requeue with backoff), matching the existing `validateAuthServerConfig` pattern.

### DIFF
--- a/cmd/thv-operator/pkg/vmcpconfig/converter.go
+++ b/cmd/thv-operator/pkg/vmcpconfig/converter.go
@@ -148,22 +148,7 @@ func (c *Converter) Convert(
 		config.Audit.Component = vmcp.Name
 	}
 
-	// Populate SessionStorage from the VirtualMCPServer spec.
-	// spec.sessionStorage is the authoritative source; always overwrite whatever
-	// the DeepCopy brought in from spec.config.sessionStorage.
-	// PasswordRef is K8s-specific and is resolved separately; the password is injected
-	// as the THV_SESSION_REDIS_PASSWORD environment variable by the deployment builder.
-	if vmcp.Spec.SessionStorage != nil &&
-		vmcp.Spec.SessionStorage.Provider == mcpv1alpha1.SessionStorageProviderRedis {
-		config.SessionStorage = &vmcpconfig.SessionStorageConfig{
-			Provider:  vmcp.Spec.SessionStorage.Provider,
-			Address:   vmcp.Spec.SessionStorage.Address,
-			DB:        vmcp.Spec.SessionStorage.DB,
-			KeyPrefix: vmcp.Spec.SessionStorage.KeyPrefix,
-		}
-	} else {
-		config.SessionStorage = nil
-	}
+	config.SessionStorage = convertSessionStorage(vmcp)
 
 	// Apply operational defaults (fills missing values)
 	config.EnsureOperationalDefaults()
@@ -279,6 +264,24 @@ func mapResolvedOIDCToVmcpConfig(
 	}
 
 	return config
+}
+
+// convertSessionStorage populates SessionStorage from the VirtualMCPServer spec.
+// spec.sessionStorage is the authoritative source; always overwrite whatever
+// the DeepCopy brought in from spec.config.sessionStorage.
+// PasswordRef is K8s-specific and is resolved separately; the password is injected
+// as the THV_SESSION_REDIS_PASSWORD environment variable by the deployment builder.
+func convertSessionStorage(vmcp *mcpv1alpha1.VirtualMCPServer) *vmcpconfig.SessionStorageConfig {
+	if vmcp.Spec.SessionStorage != nil &&
+		vmcp.Spec.SessionStorage.Provider == mcpv1alpha1.SessionStorageProviderRedis {
+		return &vmcpconfig.SessionStorageConfig{
+			Provider:  vmcp.Spec.SessionStorage.Provider,
+			Address:   vmcp.Spec.SessionStorage.Address,
+			DB:        vmcp.Spec.SessionStorage.DB,
+			KeyPrefix: vmcp.Spec.SessionStorage.KeyPrefix,
+		}
+	}
+	return nil
 }
 
 // convertAuthServerConfig converts the inline EmbeddedAuthServerConfig from the


tgrunnagle:
Regarding the failing test, this is what claude suggests

"E2E test bug in test/e2e/thv-operator/virtualmcp/virtualmcp_authserver_config_test.go: the test creates a VirtualMCPServer with IncomingAuth.Type = "anonymous" alongside a non-nil AuthServerConfig. During reconciliation, runValidations → validateAuthServerConfig will pass (issuer + upstreams are set) and momentarily set AuthServerConfigValidated = True. But then ensureVmcpConfigConfigMap → ValidateAuthServerIntegration → validateAuthServerRequiresOIDC will fail ("embedded auth server requires OIDC incoming auth"), overwriting the condition to False. The test waits for "True" and will always time out. The test needs either OIDC incoming auth configured, or the test should assert that the condition ends up False (if anonymous auth with an embedded AS is intentionally unsupported).E2E test bug in test/e2e/thv-operator/virtualmcp/virtualmcp_authserver_config_test.go: the test creates a VirtualMCPServer with IncomingAuth.Type = "anonymous" alongside a non-nil AuthServerConfig. During reconciliation, runValidations → validateAuthServerConfig will pass (issuer + upstreams are set) and momentarily set AuthServerConfigValidated = True. But then ensureVmcpConfigConfigMap → ValidateAuthServerIntegration → validateAuthServerRequiresOIDC will fail ("embedded auth server requires OIDC incoming auth"), overwriting the condition to False. The test waits for "True" and will always time out. The test needs either OIDC incoming auth configured, or the test should assert that the condition ends up False (if anonymous auth with an embedded AS is intentionally unsupported)."

jhrozek:
@jerm-dro would you mind re-stamping? There were conflicts plus rebasing the PR atop origin/main caused one function to go over the cyclomatic limit. No changes to the PR logic.

tgrunnagle:
@jhrozek I can stamp after merge conflicts are resolved

Issue #4374, opened 2026-03-26
Optimizer find_tool returns tools the user is not authorized to use


jerm-dro:
## Problem

When the optimizer is enabled with Cedar authorization, `find_tool` can return backend tool names and descriptions that the user is not authorized to call.

Cedar authz gates `tools/list` via response filtering (checking each tool against the user's policies) and gates `tools/call` via request authorization. But when a user calls `find_tool`, Cedar only checks whether the user is authorized to call `Tool::"find_tool"` itself — it does not filter the search results against the user's per-tool policies. The optimizer searches its full embedding store and returns matching backend tools regardless of the caller's authorization.

This is an information disclosure issue: a user who is permitted to use `find_tool` can discover the names and descriptions of tools they are explicitly denied access to.

## Steps to reproduce

1. Configure a vMCP server with the optimizer enabled and Cedar authorization
2. Add a Cedar policy that permits `find_tool` but denies a specific backend tool (e.g., `Tool::"admin_tool"`)
3. Call `find_tool` with a query that matches `admin_tool`
4. `find_tool` returns `admin_tool` in its results despite the user lacking authorization

## Expected behavior

`find_tool` results should be filtered against the caller's Cedar policies so that unauthorized tools are not disclosed.

Related: #4373

Generated with [Claude Code](https://claude.com/claude-code)

Issue #4373, opened 2026-03-26
Optimizer should be transparent to Cedar authorization policies


jerm-dro:
As a VirtualMCP server operator, I don't want to rewrite all my cedar policies when I enable/disable the optimizer.

## Problem

When both the optimizer and Cedar authorization are enabled on a vMCP server, enabling the optimizer silently breaks existing Cedar policies.

The optimizer replaces all backend tools with `find_tool` and `call_tool` meta-tools. Cedar's response filter evaluates each tool in `tools/list` against the user's policies, but those policies reference real backend tool names (e.g., `Tool::"weather"`). Since neither `Tool::"find_tool"` nor `Tool::"call_tool"` matches any permit policy, default-deny filters both out — `tools/list` returns zero tools with no error.

**There is a workaround today**: operators can write Cedar policies that explicitly permit the meta-tools and use argument inspection (`context.arg_tool_name`) to restrict which backend tools can be invoked through `call_tool`. But this requires operators to know about the optimizer's internal tool names, which defeats the purpose of the optimizer being a transparent performance layer.

## Expected behavior

Enabling the optimizer should not require rewriting Cedar authorization policies. An operator's existing policy like:

```cedar
permit(principal, action == Action::"call_tool", resource == Tool::"weather");
```

should continue to work regardless of whether the optimizer is enabled.

## Possible approaches

1. **Authz filter evaluates against original backend tool names** — During `tools/list` response filtering, if the optimizer is active, evaluate Cedar policies against the original backend tools rather than the meta-tools.

2. **Optimizer auto-injects permit policies** — The optimizer could automatically add Cedar permits for `find_tool` and `call_tool`, and translate the existing backend-tool permits into `arg_tool_name`-based `forbid`/`permit` rules.

3. **Authz bypass for optimizer meta-tools** — Mark `find_tool`/`call_tool` as system-level tools that skip Cedar evaluation entirely, relying on the optimizer's own `call_tool` to enforce authorization against the real backend tool name before proxying.

## Test coverage

Integration tests demonstrating the incompatibility and the workaround are in:
- `pkg/vmcp/server/session_management_integration_test.go` — `TestIntegration_OptimizerWithCedarAuthz`

Generated with [Claude Code](https://claude.com/claude-code)

PR #4385, opened 2026-03-26
Enforce Cedar policies on optimizer find_tool and call_tool

aponcedeleonch:
## Summary

When the optimizer is enabled alongside Cedar authorization, the optimizer's meta-tools (\`find_tool\` and \`call_tool\`) replace real backend tools in \`tools/list\` responses. Since operator-written Cedar policies reference real tool names (e.g., \`Tool::\"weather\"\`), the default-deny policy would filter out the meta-tools — leaving clients with zero tools. Additionally, \`call_tool\` invocations and \`find_tool\` results were not checked against Cedar policies.

All enforcement lives in the authz middleware layer. The optimizer decorator has no knowledge of the authorizer.

- **\`tools/list\`**: meta-tools (\`find_tool\`, \`call_tool\`) bypass Cedar response filtering via a pass-through set; real backend tools are filtered as usual
- **\`tools/call\` → \`call_tool\`**: middleware extracts the inner \`tool_name\` argument and authorizes that backend tool against Cedar before serving
- **\`tools/call\` → \`find_tool\`**: middleware intercepts the response and filters the returned tools list through Cedar, so callers cannot discover tools they are not authorized to call
- When the optimizer is not enabled, behavior is identical to main

Fixes https://github.com/stacklok/toolhive/issues/4374
Fixes https://github.com/stacklok/toolhive/issues/4373

## Type of change

- [x] Bug fix

## Test plan

## Large PR Justification

- ~800 of the ~1100 changed lines are test code which cannot be meaningfully separated from the production code they cover
- The remaining production code spans the authz middleware, response filter, and composition root — these form a single atomic change; splitting would leave Cedar enforcement partially wired between PRs

## Changes

| File | Change |
|------|--------|
| `pkg/authz/tool_filter.go` | New file: unexported `filterToolsByPolicy` and `authorizeToolCall` helpers |
| `pkg/authz/tool_filter_test.go` | Tests for the new helpers (moved to `package authz` to match unexported scope) |
| `pkg/authz/config.go` | `CreateMiddlewareFromConfig` / `GetMiddlewareFromFile` accept optional `passThroughTools` |
| `pkg/authz/config_test.go` | Updated tests |
| `pkg/authz/middleware.go` | `Middleware` accepts `passThroughTools`; `handleToolsCall` enforces Cedar on `call_tool` inner tool and wraps `find_tool` with response filtering |
| `pkg/authz/middleware_test.go` | Tests for pass-through meta-tool handling and `call_tool` authorization |
| `pkg/authz/response_filter.go` | Pass-through set in `filterToolsResponse`; `requiresResponseFiltering` covers `find_tool`; new `filterFindToolResponse` method |
| `pkg/authz/response_filter_test.go` | Tests for pass-through behavior and `find_tool` Cedar filtering |
| `cmd/vmcp/app/commands.go` | Builds `passThroughTools` when optimizer is enabled; passes to `NewIncomingAuthMiddleware` |
| `pkg/vmcp/auth/factory/incoming.go` | Accepts and forwards `passThroughTools` to `CreateMiddlewareFromConfig` |
| `pkg/vmcp/server/server.go` | Passes `passThroughTools` to the auth factory |
| `pkg/vmcp/session/optimizerdec/decorator.go` | Exports `CallToolArgToolName` / `CallToolArgParameters` constants used by the middleware |

The key design decision is **where** to enforce Cedar for optimizer meta-tools. Authorization is entirely contained in the `pkg/authz` middleware layer: `call_tool` requests are intercepted and the inner `tool_name` is authorized against the operator's real-tool Cedar policies; `find_tool` responses are filtered before reaching the client. The optimizer decorator is transparent to the auth system.

### DIFF
--- a/pkg/authz/middleware.go
+++ b/pkg/authz/middleware.go
@@ -303,8 +303,17 @@ func handleToolsCall(
 				parsedRequest.ID, toolName, innerArgs, next)
 			return
 		}
-		// find_tool or other pass-through without a wrapped toolName:
-		// allow through as a discovery operation — no authorization check.
+		// find_tool: allow through but filter the tools list in the response so
+		// callers cannot discover tools they are not authorized to call.
+		if parsedRequest.ResourceID == optimizerdec.FindToolName {
+			filteringWriter := NewResponseFilteringWriter(w, a, r, optimizerdec.FindToolName, annotationCache, passThroughTools)
+			next.ServeHTTP(filteringWriter, r)
+			if err := filteringWriter.FlushAndFilter(); err != nil {
+				slog.Warn("error filtering find_tool response", "error", err)
+			}
+			return
+		}
+		// Other pass-through tools without a wrapped toolName: allow through.
 		next.ServeHTTP(w, r)
 		return
 	}
--- a/pkg/authz/middleware_test.go
+++ b/pkg/authz/middleware_test.go
@@ -1108,7 +1108,7 @@ func TestMiddlewareOptimizerMetaTools(t *testing.T) {
 			expectHandlerHit: false,
 		},
 		{
-			name:     "find_tool passes through without authorization check",
+			name:     "find_tool request reaches handler (response filtering applied separately)",
 			toolName: "find_tool",
 			arguments: map[string]interface{}{
 				"tool_description": "search for web tools",
@@ -1426,14 +1426,14 @@ func TestHandleToolsCall(t *testing.T) {
 			expectStatus:     http.StatusOK,
 		},
 		{
-			// find_tool has no tool_name argument — it is a discovery operation
-			// and passes through without an authorization check.
-			name:     "find_tool — passes through without auth check",
+			// find_tool has no tool_name argument — the request reaches the handler
+			// and the response is filtered by Cedar before being returned.
+			name:     "find_tool — request reaches handler, response filtering applied",
 			toolName: "find_tool",
 			arguments: map[string]interface{}{
 				"tool_description": "search for web tools",
 			},
-			allowed:          false, // auth would deny, but it should never be reached
+			allowed:          false, // auth is not checked on the request itself
 			expectHandlerHit: true,
 			expectStatus:     http.StatusOK,
 		},
--- a/pkg/authz/response_filter.go
+++ b/pkg/authz/response_filter.go
@@ -17,6 +17,8 @@ import (
 	"golang.org/x/exp/jsonrpc2"
 
 	"github.com/stacklok/toolhive/pkg/authz/authorizers"
+	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
+	"github.com/stacklok/toolhive/pkg/vmcp/session/optimizerdec"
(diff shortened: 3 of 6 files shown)


jerm-dro (changes requested):
**`blocker:`** Thanks for picking this up and exploring the solution to both problems.

Threading the authorizer into the optimizer decorator creates coupling that will be hard to walk back. Once it's available deep in the session stack, future changes will naturally take dependencies on it, spreading authorization awareness across components that should be transparent to it.

I'd prefer to keep authorization as a single concern in the middleware layer, which already owns Cedar policy enforcement. The middleware can be extended to understand optimizer meta-tool semantics:

- For `tools/call` targeting `find_tool`: filter the response's tool list through Cedar (same logic as `tools/list` filtering)
- For `tools/call` targeting `call_tool`: extract the real `tool_name` from arguments and authorize against that

This isn't ideal either — it couples the middleware to optimizer argument shapes — but it's a less virulent form of tech debt. The middleware coupling is contained to one layer and doesn't invite further spread, whereas an authorizer available throughout the session stack tends to get used in more places over time. The longer-term goal is to untangle the authz middleware debt so other components only ever reason about tools the caller is already authorized to use, rather than each layer needing to know about authorization. That's a bigger design conversation we should have separately.

aponcedeleonch:
@jerm-dro Makes sense, I understand the reasoning.

I moved the enforcement entirely into `pkg/authz`:
- `tools/call` for `call_tool`: the authz middleware now extracts the inner toolName from the arguments and runs Cedar authorization against that backend tool directly.
- `tools/call` for `find_tool`: treated as a discovery operation and allowed through.
- commands.go builds the `passThroughTools` list (find_tool, call_tool) when the optimizer is enabled and passes it to NewIncomingAuthMiddleware.
- server.go, sessionmanager, and the optimizer decorator revert to their main-branch signatures — no knowledge of the authorizer anywhere below the middleware.

The optimizer decorator is back to being transparent to the auth system, all policy enforcement lives in pkg/authz.

aponcedeleonch:
@jerm-dro Thanks for the thorough review.

- **`find_tool` response filtering**: the middleware now intercepts the `find_tool` `tools/call` response and filters the embedded tools list through Cedar before it reaches the caller. `filterFindToolResponse` is a new method on `ResponseFilteringWriter`, dispatched via `requiresResponseFiltering` (renamed from `isListOperation`) — so it reuses all the existing buffering, flush, and SSE logic without any new writer type. The annotation cache is populated from the unfiltered list before filtering, mirroring `filterToolsResponse`.
- **Unexported `FilterToolsByPolicy` / `AuthorizeToolCall`**: both are now unexported (`filterToolsByPolicy`, `authorizeToolCall`); `tool_filter_test.go` moved to `package authz` to match.
- The optimizer decorator remains transparent to the auth system — the only change there is exporting two argument-key constants (`CallToolArgToolName`, `CallToolArgParameters`) that the middleware needs to extract the inner tool name from `call_tool` arguments.

PR #4386, opened 2026-03-26
Add migration window fallback to ScopedProvider

amirejaz:
## Summary

- When secret scope migration fails or hasn't yet completed, `ScopedProvider.GetSecret` would return not-found for any system secret that still exists under its bare pre-migration key (e.g. `BEARER_TOKEN_foo`). This would break workload auth silently on upgrade.
- Add a transparent bare-key fallback in `ScopedProvider.GetSecret`: if the scoped key (`__thv_<scope>_<name>`) is not found, try the bare key (`<name>`). Once migration completes and bare keys are deleted, the fallback finds nothing and becomes a natural no-op.
- No config check or injection is required — the deletion of bare keys at migration completion is what naturally disables the fallback.

## Type of change

## Test plan

- [x] New table-driven test `TestScopedProvider_GetSecret_MigrationFallback` covering: bare key fallback on scoped not-found, scoped key found (no fallback), both keys missing (returns original error), non-not-found error (skips bare lookup)
- [x] `go test ./pkg/secrets/...` passes

This PR sits between `phase2-factory-helpers` and `phase3-wire-callers` in the stack. It must be merged before `phase3-wire-callers` so that callers wired to scoped providers are safe during the migration window on user upgrade.

### DIFF
--- a/pkg/secrets/encrypted.go
+++ b/pkg/secrets/encrypted.go
@@ -45,7 +45,7 @@ func (e *EncryptedManager) GetSecret(_ context.Context, name string) (string, er
 
 	value, ok := e.secrets.Load(name)
 	if !ok {
-		return "", fmt.Errorf("secret not found: %s", name)
+		return "", fmt.Errorf("%w: %s", ErrSecretNotFound, name)
 	}
 	return value.(string), nil
 }
--- a/pkg/secrets/environment.go
+++ b/pkg/secrets/environment.go
@@ -31,7 +31,7 @@ func (e *EnvironmentProvider) GetSecret(_ context.Context, name string) (string,
 	envVar := e.prefix + name
 	value := os.Getenv(envVar)
 	if value == "" {
-		return "", fmt.Errorf("secret not found: %s", name)
+		return "", fmt.Errorf("%w: %s", ErrSecretNotFound, name)
 	}
 
 	return value, nil
--- a/pkg/secrets/factory.go
+++ b/pkg/secrets/factory.go
@@ -13,7 +13,6 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
-	"strings"
 	"sync"
 
 	"github.com/adrg/xdg"
@@ -128,8 +127,8 @@ func ValidateEnvironmentProvider(ctx context.Context, provider Provider, result 
 		return result
 	}
 
-	// Check that we get the expected error message
-	if !strings.Contains(err.Error(), "secret not found") {
+	// Check that we get the expected not-found error
+	if !IsNotFoundError(err) {
 		result.Error = fmt.Errorf("unexpected error format: %w", err)
 		result.Message = "Environment provider validation failed"
 		return result
--- a/pkg/secrets/fallback.go
+++ b/pkg/secrets/fallback.go
@@ -5,6 +5,7 @@ package secrets
 
 import (
 	"context"
+	"errors"
 	"log/slog"
 	"strings"
 )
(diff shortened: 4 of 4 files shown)


reyortiz3:
do users get some notification about migrations and future breaking changes in the cli or the ui?

pkg/secrets/scoped.go
```diff
@@ -55,8 +56,26 @@ func NewScopedProvider(inner Provider, scope SecretScope) Provider {
...
+	val, err := s.provider.GetSecret(ctx, s.getScopedKey(name))
+	if err == nil {
+		return val, nil
+	}
+	if IsNotFoundError(err) {
+		// Migration window: the scoped key does not exist yet. Try the bare key
+		// that was used before secret scope migration ran. After migration
+		// completes and the bare key is deleted, this lookup returns not-found
+		// and we fall through to return the original scoped-key error.
+		if bareVal, bareErr := s.provider.GetSecret(ctx, name); bareErr == nil {
```
aponcedeleonch: **Issue: bare-key backend error is silently discarded**

If the bare key lookup fails with a non-not-found error (e.g. a backend connection failure, auth error, or timeout), `bareErr` is silently dropped and the caller receives the original scoped-key "not found" error. This makes a backend failure invisible — the user/operator sees "secret not found" when the real problem is "backend unreachable".

A missing test case: `bareErr = errors.New("backend connection failed")`. In the current code this returns the scoped not-found error with no indication the bare-key lookup itself failed differently.

Consider distinguishing the two cases:
```go
if bareVal, bareErr := s.provider.GetSecret(ctx, name); bareErr == nil {
	slog.Debug("secret scope migration fallback: returning bare key",
		"scope", s.scope, "name", name)
	return bareVal, nil
} else if !IsNotFoundError(bareErr) {
	// Bare-key lookup hit a real backend error — surface it so the caller
	// doesn't misdiagnose a connection failure as "secret not found".
	return "", fmt.Errorf("scoped key not found and bare-key fallback failed: %w", bareErr)
}
```

pkg/secrets/scoped.go
```diff
@@ -55,8 +56,26 @@ func NewScopedProvider(inner Provider, scope SecretScope) Provider {
...
+// If the scoped key is not found, it falls back to the bare (pre-migration) key.
+// This makes the provider safe to use before or during secret scope migration:
+// once migration completes and bare keys are deleted, the fallback is a no-op.
 func (s *ScopedProvider) GetSecret(ctx context.Context, name string) (string, error) {
-	return s.provider.GetSecret(ctx, s.getScopedKey(name))
+	val, err := s.provider.GetSecret(ctx, s.getScopedKey(name))
+	if err == nil {
+		return val, nil
+	}
+	if IsNotFoundError(err) {
```
aponcedeleonch, line 78: **Issue: `IsNotFoundError` string-matching is now load-bearing for the migration fallback**

The entire migration fallback is gated on `IsNotFoundError` (defined in `fallback.go`). That function uses `strings.Contains` to match `"not found"` or `"does not exist"`. If any backend (Keychain, 1Password, Vault, …) returns a differently-worded response (e.g. `"key not present"`, `"entry missing"`, `"no such secret"`), `IsNotFoundError` returns `false`, the fallback is **silently skipped**, and users migrating on that backend will hit the scoped-key error with no fallback attempted — with no indication why.

This was pre-existing in the env-var fallback path, but this PR adds a second critical code path that depends on it. It may be worth auditing the "not found" error messages of each supported backend to confirm they're all covered, or introducing a sentinel error type (e.g. `ErrSecretNotFound`) that backends return so the check doesn't rely on string matching.

amirejaz: Good call. The doc comment on `IsNotFoundError` now explicitly documents which error messages each built-in backend returns and states the contract new backends must meet. The string-matching is load-bearing, but it's at least visible and auditable now.

Agreed that a sentinel error (e.g. `ErrSecretNotFound`) would be the robust long-term fix — no more reliance on message text, and the compiler enforces the contract for new backends. I'll open a follow-up issue to track that refactor so it doesn't get lost.

aponcedeleonch: **Concrete proposal for the sentinel error refactor**

I audited every `Provider` implementation — most backends *can* expose a typed error today:

| Provider | Current not-found error | Typed error possible? |
|---|---|---|
| `EnvironmentProvider` | `fmt.Errorf("secret not found: %s")` | Yes — we own it |
| `EncryptedManager` | `fmt.Errorf("secret not found: %s")` | Yes — we own it |
| `ZalandoKeyring` | `keyring.ErrNotFound` sentinel | **Already typed** |
| `KeyctlProvider` | `keyring.ErrNotFound` sentinel | **Already typed** |
| `OnePasswordManager` | Wrapped SDK error (no sentinel) | Needs adapter-level mapping |

The keyring layer already does the right thing with `errors.Is()`. The fix is to define a package-level sentinel and have each owned provider wrap it:

```go
// pkg/secrets/types.go
var ErrSecretNotFound = errors.New("secret not found")
```

Update providers:
```go
// EnvironmentProvider, EncryptedManager
return "", fmt.Errorf("%w: %s", ErrSecretNotFound, name)

// keyring composite boundary — map keyring.ErrNotFound at the edge
return "", fmt.Errorf("%w: %s", ErrSecretNotFound, key)
```

Update `IsNotFoundError` to prefer `errors.Is`, keeping string matching only as a legacy fallback for third-party providers (e.g. 1Password SDK which doesn't expose typed errors):

```go
func IsNotFoundError(err error) bool {
	if err == nil {
		return false
	}
	if errors.Is(err, ErrSecretNotFound) {
		return true
	}
	// Legacy fallback for providers that don't yet wrap ErrSecretNotFound.
	errStr := err.Error()
	return strings.Contains(errStr, "not found") ||
		strings.Contains(errStr, "does not exist")
}
```

This makes the string matching no longer load-bearing for core paths. New backends get a compiler-visible contract via the sentinel. The string fallback can be removed once the 1Password adapter maps SDK errors at its boundary.

PR #4393, opened 2026-03-26
fix: make experimental MCP server initialization non-fatal to thv serve

### LINKED ISSUE #4392: Thv serve crashes when experimental MCP server fails to initialize
## Description

When running `thv serve --experimental-mcp`, the entire HTTP server process crashes if the embedded MCP server fails to initialize (e.g., due to a registry authentication error).

This happens because `mcpserver.New()` is called synchronously in the `serve` command's `RunE` function, and any error from it is returned directly — killing the whole process, including the main HTTP API server.

## Steps to reproduce

1. Configure a custom registry API that requires OAuth authentication
2. Let the OAuth token expire (or misconfigure the credentials)
3. Run `thv serve --experimental-mcp`
4. The process exits immediately with: `failed to create MCP server: failed to create ToolHive handler: failed to get registry provider: registry authentication required`

## Expected behavior

The main HTTP API server (`s.Serve`) should continue running even if the experimental MCP server fails to initialize. The MCP failure should be logged as an error but not be fatal to the process.

## Actual behavior

The entire `thv serve` process crashes, making the HTTP API unavailable. This is particularly problematic for ToolHive Studio (Electron), which relies on the HTTP API being available regardless of the MCP server state.

## Proposed fix

Move `mcpserver.New()` inside the goroutine alongside `mcpServer.Start()`, so that a failure in MCP initialization only terminates the goroutine, not the main process. The error is logged with `slog.Error` and the HTTP server continues running normally.

peppescg:
## Summary

- Move `mcpserver.New()` inside the goroutine so that a failure in MCP server creation (e.g., registry authentication error) only terminates the goroutine instead of crashing the entire `thv serve` process
- The main HTTP API server continues running normally, enabling graceful degradation when the experimental MCP feature encounters issues

Fixes #4392

## Test plan

- [ ] Run `thv serve --experimental-mcp` with valid registry credentials → both HTTP API and MCP server start normally
- [ ] Run `thv serve --experimental-mcp` with expired/invalid registry credentials → HTTP API starts normally, MCP failure is logged but process stays alive
- [ ] Run `thv serve` without `--experimental-mcp` → no change in behavior

### DIFF
--- a/cmd/thv/app/server.go
+++ b/cmd/thv/app/server.go
@@ -74,34 +74,29 @@ var serveCmd = &cobra.Command{
 		if enableMCPServer {
 			fmt.Println("EXPERIMENTAL: Starting embedded MCP server")
 
-			// Create MCP server configuration
 			mcpConfig := &mcpserver.Config{
 				Host: mcpServerHost,
 				Port: mcpServerPort,
 			}
 
-			// Create and start the MCP server in a goroutine
-			mcpServer, err := mcpserver.New(ctx, mcpConfig)
-			if err != nil {
-				return fmt.Errorf("failed to create MCP server: %w", err)
-			}
-
 			go func() {
-				if err := mcpServer.Start(); err != nil {
-					slog.Error(fmt.Sprintf("MCP server error: %v", err))
+				mcpServer, err := mcpserver.New(ctx, mcpConfig)
+				if err != nil {
+					slog.Error("Failed to create MCP server, continuing without it", "error", err)
+					return
 				}
-			}()
 
-			// Ensure MCP server is shut down on context cancellation
-			go func() {
-				<-ctx.Done()
-				// Use Background context for MCP server shutdown. The parent context is already
-				// cancelled at this point, so we need a fresh context with its own timeout to
-				// ensure the shutdown operation completes successfully.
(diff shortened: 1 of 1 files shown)


jerm-dro (commented):
**`question:`** What's the use case for `thv serve` continuing to run when the server fails to start? Continuing without it means the process looks healthy but isn't doing what was asked.

The original crash-on-failure behavior gives immediate, unmissable feedback. Is there a scenario where the rest of the process has independent value?

peppescg:
> What's the use case for `thv serve` continuing to run when the server fails to start?

This issue kills the thv process, including the main HTTP API server ( thv serve). The desktop app cannot do anything cause the process was killed.

This happens when a registry is not correctly authenticated, and when you restart the http server and the thv process ( for instance restarting the desktop app), the process cannot be spawned and the http server as well.
This is caused by the toolhive mcp ( we are spawning the thv http server using experimental mcp) that is doing a registry request under the hood, but if the registry is not authenticated this kills the process.

> Is there a scenario where the rest of the process has independent value?

Thv serve cannot works if the thv process is not up and running, so I cannot call registry api for authenticating the registry or reset it

jerm-dro (approved):
Thanks for explaining

PR #4402, opened 2026-03-27
Wire Redis session storage and fix MultiSession lookup

### LINKED ISSUE #4220: Write unit tests for horizontal scaling operator behaviors (THV-0047)
## Description

Write unit tests covering all new operator behaviors introduced by TASK-004, TASK-005, TASK-006, and TASK-007: spec-driven replica setting (nil-passthrough and explicit values), stdio transport hard cap, session storage warning conditions, Redis env var injection via `SecretKeyRef`, `backendReplicas` passthrough in RunConfig, and `terminationGracePeriodSeconds` presence on generated Deployments. The existing `mcpserver_replicas_test.go` covers externally-driven (HPA) replica preservation; new test cases here cover the distinct spec-driven code paths added by the upstream tasks.

## Context

Epic THV-0047 adds horizontal scaling support to the ToolHive operator by extending MCPServer and VirtualMCPServer CRDs with replica counts, session storage configuration, and Redis injection. The upstream tasks (TASK-004 through TASK-007) implement the operator-side changes but delegate formal test coverage to this task. Without these tests, the new code paths — nil-passthrough replicas, stdio cap enforcement, session storage warning conditions, Redis env var injection, and RunConfig population — have no automated verification. This task closes that gap.

**Dependencies**: #280 (TASK-004 — MCPServer reconciler/deployment builder), #281 (TASK-005 — MCPServer RunConfig Redis/BackendReplicas), #278 (TASK-006 — VirtualMCPServer deployment builder), #277 (TASK-007 — VirtualMCPServer vMCP ConfigMap Redis injection)
**Blocks**: None (this is the final task in the DAG)

## Acceptance Criteria

- [ ] `mcpserver_replicas_test.go` is extended with spec-driven replica test cases: `spec.replicas == nil` yields `Deployment.Spec.Replicas == nil`; `spec.replicas = 1` yields `*Deployment.Spec.Replicas == 1`; `spec.replicas = 3` with transport `sse` yields `*Deployment.Spec.Replicas == 3`
- [ ] A test for the stdio transport hard cap verifies that the reconciler produces a Deployment with `*Spec.Replicas == 1` and sets a `StdioReplicaCapped` status condition when `spec.transport == "stdio"` and `spec.replicas = 3`
- [ ] A test verifies the session storage warning condition (`SessionStorageWarning`) is set with status `True` when `spec.replicas = 2` and `spec.sessionStorage` is nil
- [ ] A test verifies the session storage warning condition is set with status `True` when `spec.replicas = 2` and `spec.sessionStorage.provider == "memory"`
- [ ] A test verifies the session storage warning condition is set with status `False` when `spec.replicas = 2` and `spec.sessionStorage.provider == "redis"` (condition cleared)
- [ ] Tests for `createRunConfigFromMCPServer` verify: `backendReplicas = nil` yields nil `RunConfig.BackendReplicas`; `backendReplicas = 3` yields `RunConfig.BackendReplicas` pointing to `int32(3)`; `sessionStorage.provider == "redis"` populates `RunConfig.SessionRedis` with correct address, db, and keyPrefix; `sessionStorage.provider == "memory"` yields nil `RunConfig.SessionRedis`; `passwordRef` is NOT present anywhere in the serialized RunConfig
- [ ] Tests for `deploymentForMCPServer` (proxyrunner pod) verify: `THV_SESSION_REDIS_PASSWORD` env var is injected with correct `ValueFrom.SecretKeyRef` fields when `sessionStorage.provider == "redis"` and `passwordRef` is set; no `THV_SESSION_REDIS_PASSWORD` env var appears when `sessionStorage` is nil
- [ ] Tests for `deploymentForVirtualMCPServer` verify: `THV_SESSION_REDIS_PASSWORD` env var is injected correctly when `sessionStorage.provider == "redis"` and `passwordRef` is set; no `THV_SESSION_REDIS_PASSWORD` env var when `sessionStorage` is nil or `provider != "redis"` or `passwordRef` is nil
- [ ] Tests for `deploymentForMCPServer` and `deploymentForVirtualMCPServer` verify that `Spec.Template.Spec.TerminationGracePeriodSeconds` is non-nil and set to the expected constant value (e.g., 30 seconds) in all cases
- [ ] Tests for `deploymentForVirtualMCPServer` verify the nil-passthrough replica behavior: `spec.replicas == nil` yields `Spec.Replicas == nil`; `spec.replicas = 1` yields `*Spec.Replicas == 1`; `spec.replicas = 3` yields `*Spec.Replicas == 3`
- [ ] Tests for `ensureVmcpConfigConfigMap` (or the converter) verify: Redis `address`, `db`, and `keyPrefix` appear in the marshaled `config.yaml` when `provider == "redis"`; the `config.yaml` omits `sessionStorage` entirely when `provider == "memory"` or `spec.sessionStorage == nil`; no password field appears in `config.yaml`
- [ ] All pre-existing tests in `mcpserver_replicas_test.go` continue to pass — the new spec-driven tests do not interfere with the HPA-scenario tests
- [ ] `go test ./cmd/thv-operator/... ./pkg/runner/... ./pkg/vmcp/config/...` passes without failures

The new tests span four existing test files (and may introduce one or two new files if the scope warrants it). The primary organization is:

**1. Extend `mcpserver_replicas_test.go` with spec-driven replica cases**

The existing `TestReplicaBehavior` table-driven test covers externally-driven scaling (HPA scenario: the deployment already has N replicas, the reconciler runs). The new tests cover the complementary path: the MCPServer spec itself sets `Replicas`, and the deployment builder/reconciler must respect it. Since these are fundamentally different code paths (not just different table rows), add them as new top-level test functions rather than extending the existing table:

```go
func TestSpecDrivenReplicasNil(t *testing.T) { ... }
func TestSpecDrivenReplicasExplicit(t *testing.T) { ... }   // table with 1 and 3
func TestSpecDrivenStdioCapAt1(t *testing.T) { ... }
func TestSessionStorageWarningCondition(t *testing.T) { ... }  // table: nil, memory, redis
func TestTerminationGracePeriodMCPServer(t *testing.T) { ... }
```

The setup pattern mirrors `TestDefaultCreationHasOneReplica`: create an MCPServer with the desired spec (including new `Replicas` and `SessionStorage` fields), create a fake client with `fake.NewClientBuilder()`, instantiate `newTestMCPServerReconciler`, call `reconciler.Reconcile`, fetch the updated Deployment, and assert on `Spec.Replicas`, `Spec.Template.Spec.TerminationGracePeriodSeconds`, and the MCPServer's `Status.Conditions`.

For condition assertions, fetch the updated MCPServer after reconcile and use `apimeta.FindStatusCondition` to check the condition type, status, and reason.

**2. Extend `mcpserver_runconfig_test.go` with BackendReplicas and Redis cases**

Add new test cases to the existing `TestCreateRunConfigFromMCPServer` table (or add a new table function `TestRunConfigScalingFields`). The `createTestMCPServerWithConfig` helper at line 45 may need to be augmented with a new builder helper that accepts `BackendReplicas *int32` and `SessionStorage *mcpv1alpha1.SessionStorageConfig`. Test serialization by calling `json.Marshal` on `RunConfig` and asserting the presence/absence of `backend_replicas` and `session_redis` keys.

**3. Extend `virtualmcpserver_deployment_test.go` with replica, termination grace, and Redis env var cases**

The existing `TestDeploymentForVirtualMCPServer` at line 37 already asserts `*Deployment.Spec.Replicas == 1`. After TASK-006, this assertion will need updating (nil-passthrough means nil when `spec.replicas == nil`). Add additional test cases covering:
- `spec.replicas = nil` → `Spec.Replicas == nil`
- `spec.replicas = 3` → `*Spec.Replicas == 3`
- `TerminationGracePeriodSeconds` is set in all cases
- `buildRedisPasswordEnvVar` returns the correct `corev1.EnvVar` when `provider == "redis"` and `passwordRef` is set, and returns empty/nil for all other combinations

For the `buildRedisPasswordEnvVar` helper tests, call the function directly (it is a method on `*VirtualMCPServerReconciler`) with a minimal VirtualMCPServer struct containing only the relevant `SessionStorage` field.

**4. Extend `virtualmcpserver_vmcpconfig_test.go` with Redis session storage cases**

Add test cases to `TestCreateVmcpConfigFromVirtualMCPServer` (or a new `TestVmcpConfigRedisInjection` table) that:
- Set `vmcp.Spec.SessionStorage` with `provider == "redis"`, `address`, `db`, and `keyPrefix`
- Marshal the resulting config to YAML and assert on the presence of `sessionStorage.address`, `sessionStorage.db`, `sessionStorage.keyPrefix`
- Confirm no password field appears under `sessionStorage`
- Set `provider == "memory"` and confirm `sessionStorage` is absent from the YAML

The existing `gopkg.in/yaml.v3` import in `virtualmcpserver_vmcpconfig_test.go` is already available for YAML assertions.

### Patterns & Frameworks

- **Table-driven tests with `t.Parallel()`**: All new tests must use `t.Parallel()` and `t.Run(tt.name, func(...) { t.Parallel() ... })` following the existing `TestReplicaBehavior` pattern. Use `github.com/stretchr/testify/assert` and `github.com/stretchr/testify/require` for assertions.
- **`fake.NewClientBuilder()` pattern**: Construct a fake Kubernetes client with `fake.NewClientBuilder().WithScheme(testScheme).WithObjects(...).WithStatusSubresource(...).Build()`. The scheme must include both `corev1` and `mcpv1alpha1`.
- **`newTestMCPServerReconciler` helper**: Use this to construct a properly initialized `MCPServerReconciler` with a mock platform detector and `ImageValidationAlwaysAllow`.
- **`apimeta.FindStatusCondition`**: Use `k8s.io/apimachinery/pkg/api/meta.FindStatusCondition(&status.Conditions, conditionType)` to retrieve and assert on status conditions after reconcile.
- **Direct function call for builder tests**: For `deploymentForMCPServer`, `deploymentForVirtualMCPServer`, and `buildRedisPasswordEnvVar`, call the functions directly rather than going through a full reconcile loop when testing only the deployment shape. This reduces setup overhead and makes assertions more targeted.
- **JSON marshaling for RunConfig assertions**: Use `json.Marshal(runConfig)` and `json.Unmarshal` into a `map[string]interface{}` to assert on the presence or absence of fields without depending on exact struct shapes.

### Code Pointers

- `cmd/thv-operator/controllers/mcpserver_replicas_test.go` — Primary file to extend; contains `TestReplicaBehavior`, `TestConfigUpdatePreservesReplicas`, and `TestDefaultCreationHasOneReplica` as reference patterns. Note: `TestDefaultCreationHasOneReplica` at line 459 asserts `*deployment.Spec.Replicas == 1` for a server without `spec.replicas` set; after TASK-004, this may need updating to `nil` (nil-passthrough). Verify with the TASK-004 implementer.
- `cmd/thv-operator/controllers/mcpserver_runconfig_test.go` — Contains `createTestMCPServerWithConfig` helper at line 45 and `TestCreateRunConfigFromMCPServer` table at line 61; extend with Redis and BackendReplicas cases.
- `cmd/thv-operator/controllers/virtualmcpserver_deployment_test.go` — Contains `TestDeploymentForVirtualMCPServer` at line 37 with existing replica assertion at line 64 (`*deployment.Spec.Replicas == 1`); this assertion will need updating after TASK-006 changes the default to nil when `spec.replicas` is nil.
- `cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig_test.go` — Contains `TestCreateVmcpConfigFromVirtualMCPServer` table and `newTestConverter` helper; extend for Redis session storage fields.
- `cmd/thv-operator/controllers/mcpserver_test_helpers_test.go` — `newTestMCPServerReconciler` helper at line 30; use this to construct the reconciler in all MCPServer tests.
- `cmd/thv-operator/api/v1alpha1/mcpserver_types.go` — Source of `ConditionStdioReplicaCapped`, `ConditionSessionStorageWarning`, and associated reason constants added by TASK-004; import these rather than using string literals in tests.
- `cmd/thv-operator/controllers/virtualmcpserver_deployment.go` lines 289–329 — `buildOIDCEnvVars` — reference for the `SecretKeyRef` env var pattern that `buildRedisPasswordEnvVar` (added by TASK-006) follows; read to understand the expected `corev1.EnvVar` shape to assert against.
- `pkg/runner/config.go` — `RunConfig` struct with `BackendReplicas` and `SessionRedis` fields added by TASK-005; import `runner.SessionRedisConfig` for test assertions.
- `pkg/vmcp/config/config.go` — `Config` struct with `SessionStorage *SessionStorageConfig` field added by TASK-007; import `vmcpconfig.SessionStorageConfig` for converter test assertions.

### Component Interfaces

Key types needed for test assertions (all defined by upstream tasks):

```go
// From cmd/thv-operator/api/v1alpha1/mcpserver_types.go (added by TASK-004)
const ConditionStdioReplicaCapped  = "StdioReplicaCapped"
const ConditionSessionStorageWarning = "SessionStorageWarning"
const ConditionReasonStdioReplicaCapped = "StdioTransportCapAt1"
const ConditionReasonSessionStorageMissing = "SessionStorageMissingForReplicas"
const ConditionReasonSessionStorageConfigured = "SessionStorageConfigured"

// From pkg/runner/config.go (added by TASK-005)
type SessionRedisConfig struct {
    Address   string `json:"address,omitempty" yaml:"address,omitempty"`
    DB        int32  `json:"db,omitempty" yaml:"db,omitempty"`
    KeyPrefix string `json:"key_prefix,omitempty" yaml:"key_prefix,omitempty"`
}
// runner.RunConfig also gains:
//   BackendReplicas *int32
//   SessionRedis *SessionRedisConfig

// From pkg/vmcp/config/config.go (added by TASK-007)
type SessionStorageConfig struct {
    Address   string `json:"address,omitempty" yaml:"address,omitempty"`
    DB        int32  `json:"db,omitempty" yaml:"db,omitempty"`
    KeyPrefix string `json:"keyPrefix,omitempty" yaml:"keyPrefix,omitempty"`
}
// vmcpconfig.Config also gains:
//   SessionStorage *SessionStorageConfig
```

Illustrative test structure for spec-driven replica nil-passthrough (extend `mcpserver_replicas_test.go`):

```go
func TestSpecDrivenReplicasNil(t *testing.T) {
    t.Parallel()
    name, namespace := "spec-nil-replica", testNamespaceDefault
    mcpServer := &mcpv1alpha1.MCPServer{
        ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace},
        Spec: mcpv1alpha1.MCPServerSpec{
            Image: "test-image:latest", Transport: "sse", ProxyPort: 8080,
            Replicas: nil, // explicit nil — hands-off mode
        },
    }
    // ... fake client setup, reconcile, fetch deployment
    assert.Nil(t, deployment.Spec.Replicas, "spec.replicas == nil must yield Deployment.Spec.Replicas == nil")
}
```

Illustrative test for stdio cap with condition assertion:

```go
func TestSpecDrivenStdioCapAt1(t *testing.T) {
    t.Parallel()
    replicas := int32(3)
    mcpServer := &mcpv1alpha1.MCPServer{
        // ...
        Spec: mcpv1alpha1.MCPServerSpec{
            Transport: "stdio", Replicas: &replicas, // ...
        },
    }
    // ... reconcile
    // Assert deployment has 1 replica
    assert.Equal(t, int32(1), *deployment.Spec.Replicas)
    // Assert condition is set
    condition := apimeta.FindStatusCondition(updatedMCPServer.Status.Conditions,
        mcpv1alpha1.ConditionStdioReplicaCapped)
    require.NotNil(t, condition)
    assert.Equal(t, metav1.ConditionTrue, condition.Status)
    assert.Equal(t, mcpv1alpha1.ConditionReasonStdioReplicaCapped, condition.Reason)
}
```

Illustrative test for Redis env var injection in VirtualMCPServer deployment:

```go
func TestBuildRedisPasswordEnvVar(t *testing.T) {
    t.Parallel()
    tests := []struct {
        name        string
        sessionStorage *mcpv1alpha1.SessionStorageConfig
        wantEnvVars int
        wantEnvName string
        wantSecretName string
        wantSecretKey  string
    }{
        {
            name: "redis with passwordRef injects env var",
            sessionStorage: &mcpv1alpha1.SessionStorageConfig{
                Provider: "redis",
                PasswordRef: &mcpv1alpha1.SecretKeyRef{Name: "redis-secret", Key: "password"},
            },
            wantEnvVars: 1, wantEnvName: "THV_SESSION_REDIS_PASSWORD",
            wantSecretName: "redis-secret", wantSecretKey: "password",
        },
        {name: "nil sessionStorage returns no env vars", sessionStorage: nil, wantEnvVars: 0},
        {name: "memory provider returns no env vars",
            sessionStorage: &mcpv1alpha1.SessionStorageConfig{Provider: "memory"}, wantEnvVars: 0},
        {name: "redis without passwordRef returns no env vars",
            sessionStorage: &mcpv1alpha1.SessionStorageConfig{Provider: "redis"}, wantEnvVars: 0},
    }
    // ...
}
```

### Testing Strategy

**Unit Tests**

MCPServer reconciler / deployment builder (extend `mcpserver_replicas_test.go`):
- [ ] `TestSpecDrivenReplicasNil` — `spec.replicas == nil` produces a Deployment with `Spec.Replicas == nil` (nil-passthrough)
- [ ] `TestSpecDrivenReplicas1` — `spec.replicas = 1` produces `*Spec.Replicas == 1`
- [ ] `TestSpecDrivenReplicas3SseTransport` — `spec.replicas = 3`, transport `sse` produces `*Spec.Replicas == 3`
- [ ] `TestSpecDrivenStdioCapAt1` — `spec.transport = "stdio"`, `spec.replicas = 3` produces a Deployment with `*Spec.Replicas == 1` and sets `StdioReplicaCapped` condition with status `True` and reason `StdioTransportCapAt1`
- [ ] `TestSessionStorageWarningConditionNilStorage` — `spec.replicas = 2`, `spec.sessionStorage = nil` sets `SessionStorageWarning` condition with status `True` and reason `SessionStorageMissingForReplicas`
- [ ] `TestSessionStorageWarningConditionMemory` — `spec.replicas = 2`, `spec.sessionStorage.provider = "memory"` sets `SessionStorageWarning` with status `True`
- [ ] `TestSessionStorageWarningConditionRedis` — `spec.replicas = 2`, `spec.sessionStorage.provider = "redis"` sets `SessionStorageWarning` with status `False` and reason `SessionStorageConfigured`
- [ ] `TestTerminationGracePeriodMCPServer` — `deploymentForMCPServer` output always has `Spec.Template.Spec.TerminationGracePeriodSeconds` equal to the expected constant

MCPServer RunConfig (extend `mcpserver_runconfig_test.go`):
- [ ] `backendReplicas = nil` → `RunConfig.BackendReplicas == nil`
- [ ] `backendReplicas = 3` → `*RunConfig.BackendReplicas == 3`
- [ ] `sessionStorage.provider = "memory"` → `RunConfig.SessionRedis == nil`
- [ ] `sessionStorage = nil` → `RunConfig.SessionRedis == nil`
- [ ] `sessionStorage.provider = "redis"`, address `"redis.svc:6379"`, db `2`, keyPrefix `"thv:"` → `RunConfig.SessionRedis` equals `{Address: "redis.svc:6379", DB: 2, KeyPrefix: "thv:"}`
- [ ] Serialized RunConfig JSON omits `backend_replicas` when nil; includes it when set
- [ ] Serialized RunConfig JSON does not contain any password or credential when `passwordRef` is set in the CRD spec

VirtualMCPServer deployment builder (extend `virtualmcpserver_deployment_test.go`):
- [ ] `spec.replicas = nil` → `Spec.Replicas == nil`
- [ ] `spec.replicas = 1` → `*Spec.Replicas == 1`
- [ ] `spec.replicas = 3` → `*Spec.Replicas == 3`
- [ ] `TerminationGracePeriodSeconds` is set to the expected constant in all replica variants
- [ ] `buildRedisPasswordEnvVar` with `provider = "redis"` and `passwordRef.name = "redis-secret"`, `passwordRef.key = "password"` returns one `corev1.EnvVar` named `THV_SESSION_REDIS_PASSWORD` with correct `ValueFrom.SecretKeyRef`
- [ ] `buildRedisPasswordEnvVar` with nil `sessionStorage` returns no env vars
- [ ] `buildRedisPasswordEnvVar` with `provider = "memory"` returns no env vars
- [ ] `buildRedisPasswordEnvVar` with `provider = "redis"` and nil `passwordRef` returns no env vars

VirtualMCPServer vMCP ConfigMap (extend `virtualmcpserver_vmcpconfig_test.go`):
- [ ] `provider = "redis"` with all fields set → marshaled `config.yaml` contains `sessionStorage.address`, `sessionStorage.db`, `sessionStorage.keyPrefix`
- [ ] `provider = "redis"` → marshaled `config.yaml` does NOT contain any password field
- [ ] `provider = "memory"` → `config.yaml` omits `sessionStorage` entirely
- [ ] `spec.sessionStorage = nil` → `config.yaml` omits `sessionStorage` entirely

**Integration Tests**
- [ ] Not in scope for this task; these unit tests provide sufficient automated coverage for the new operator behaviors

**Edge Cases**
- [ ] `spec.replicas = 0` — Deployment has `*Spec.Replicas == 0` (scale-to-zero is valid); verify the stdio cap logic does not interfere with zero-replica requests for stdio transport
- [ ] `backendReplicas = 0` — `RunConfig.BackendReplicas` is a pointer to `int32(0)`, not nil; verify `json.Marshal` includes the field (omitempty only applies to nil pointers, not zero-value pointed-to integers)
- [ ] `sessionStorage.db = 0` — Verify the zero-value DB field is correctly serialized in both RunConfig and vMCP config YAML (omitempty on `int32` will omit zero; if vMCP requires explicit `db: 0`, the YAML tag may need to be `yaml:"db"` without omitempty — coordinate with TASK-007 implementer on the chosen behavior)
- [ ] Nil-passthrough after initial set: updating MCPServer from `spec.replicas = 2` to `spec.replicas = nil` does not set `Deployment.Spec.Replicas` to any value (verify `deploymentNeedsUpdate` preserves this invariant)
- [ ] Pre-existing `TestDefaultCreationHasOneReplica` may need updating if TASK-004 changes the default behavior for `spec.replicas = nil` from `1` to `nil`; coordinate with the TASK-004 implementer before marking that test as passing

## Out of Scope

- Application-level session storage implementation inside proxyrunner or vMCP — separate epics
- Integration tests with a real Kubernetes API server or live Redis instance
- HPA integration tests
- Redis deployment, provisioning, or lifecycle
- CLI, UI, or any changes outside `cmd/thv-operator/` and `pkg/runner/` and `pkg/vmcp/config/`
- New feature implementation — this task writes tests only; all implementation is in TASK-004 through TASK-007

## References

- Epic THV-0047: Horizontal Scaling - CRD and Operator Changes — https://github.com/stacklok/stacklok-epics/issues/264
- Upstream TASK-004: MCPServer reconciler and deployment builder updates — https://github.com/stacklok/stacklok-epics/issues/280
- Upstream TASK-005: MCPServer RunConfig Redis and BackendReplicas injection — https://github.com/stacklok/stacklok-epics/issues/281
- Upstream TASK-006: VirtualMCPServer deployment builder updates — https://github.com/stacklok/stacklok-epics/issues/278
- Upstream TASK-007: VirtualMCPServer vMCP ConfigMap Redis injection — https://github.com/stacklok/stacklok-epics/issues/277
- RFC THV-0047: Manual Horizontal Scaling for vMCP and Proxy Runner — https://github.com/stacklok/toolhive-rfcs/pull/47
- Existing HPA-scenario replica tests: `cmd/thv-operator/controllers/mcpserver_replicas_test.go`
- Existing RunConfig tests: `cmd/thv-operator/controllers/mcpserver_runconfig_test.go`
- Existing VirtualMCPServer deployment tests: `cmd/thv-operator/controllers/virtualmcpserver_deployment_test.go`
- Existing vMCP ConfigMap tests: `cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig_test.go`

yrobla:
## Summary

Two bugs prevented Redis-backed VirtualMCPServer from working correctly across pods:

1. The vmcp process never read the SessionStorage field from its config, always using in-memory (local) storage regardless of operator config. Fixed by adding SessionStorageConfig to server.Config and a buildRedisConfig helper in commands.go that reads the provider, address, DB, and password (via THV_SESSION_REDIS_PASSWORD) from the vmcp config at startup.

2. With Redis-backed storage, sessionmanager.Manager.GetMultiSession() always returned (nil, false) for fully-formed sessions. The transportsession.Manager.Get() deserialises from Redis into a plain *StreamableSession, losing the defaultMultiSession type. The type assertion sess.(vmcpsession.MultiSession) therefore always failed, causing GetAdaptedTools to error, which triggered the cleanup defer to call Terminate(), marking the session terminated=true in Redis. The client's next request got a 404 "session terminated".

   Fixed by adding a node-local sync.Map (multiSessions) to sessionmanager.Manager as the authoritative store for live MultiSession objects. CreateSession stores the live session there after UpsertSession succeeds; GetMultiSession and Terminate consult the local map first before falling back to the pluggable backend.

Also adds:
- E2E test: VirtualMCPServer Redis Session Continuity — verifies that MCP sessions survive complete pod replacement when Redis is configured (ServiceAffinityNone, 2 replicas, all pods deleted and replaced).
- Redis image constant and DeployRedis/CleanupRedis helpers in the e2e test package.

Fixes #4220

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

**Breaking behavior change: HTTP 401 → HTTP 404 for unknown/expired sessions**

Previously, the discovery middleware returned HTTP 401 for subsequent requests (requests carrying an `Mcp-Session-Id` header) when the session was not found or not yet fully initialized. With this change, the middleware passes those requests through to the MCP SDK, which returns HTTP 404 ("session not found").

The new behavior is more semantically accurate — an unknown session ID is a "not found" condition, not an authorization failure — but any monitoring, alerting, or client code that matches on HTTP 401 specifically for session-expiry detection will need to be updated to also handle HTTP 404.

- **`multiSessions` local map**: intentionally node-local (not shared via Redis). Redis holds serialisable metadata only; live `MultiSession` objects (which hold open HTTP connections to backends) cannot be serialised. Cross-pod recovery is handled by `RestoreSession`, which reopens backend connections from stored metadata on cache miss.
- **Two-phase session creation**: `Generate()` creates a placeholder; `CreateSession()` (called from `OnRegisterSession`) upgrades it. This window means a session can exist as a placeholder with no `MultiSession` in the local map — `handleSubsequentRequest` accounts for this by skipping capability injection rather than erroring.
- **Race window in `Terminate()`**: noted inline with a TODO for a "terminating" tombstone if needed.
- **`context.Background()` in `GetMultiSession`**: `MultiSessionGetter` carries no context (SDK interface constraint); noted inline with a TODO to add context propagation.

## Large PR Justification

This is a complete PR that covers a single functionality. It includes comprehensive tests. Cannot be split.

### DIFF
--- a/pkg/transport/session/session_data_storage_local.go
+++ b/pkg/transport/session/session_data_storage_local.go
@@ -1,0 +1,178 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package session
+
+import (
+	"context"
+	"fmt"
+	"maps"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// localDataEntry wraps session metadata with a storage-owned last-access
+// timestamp used for TTL-based eviction.
+type localDataEntry struct {
+	metadata       map[string]string
+	lastAccessNano atomic.Int64
+}
+
+func newLocalDataEntry(metadata map[string]string) *localDataEntry {
+	e := &localDataEntry{metadata: metadata}
+	e.lastAccessNano.Store(time.Now().UnixNano())
+	return e
+}
+
+func (e *localDataEntry) lastAccess() time.Time {
+	return time.Unix(0, e.lastAccessNano.Load())
+}
+
+// LocalSessionDataStorage implements SessionDataStorage using an in-memory
+// sync.Map with TTL-based eviction.
+//
+// Sessions are evicted if they have not been accessed within the configured TTL.
+// A background goroutine runs until Close is called.
+type LocalSessionDataStorage struct {
+	sessions sync.Map // map[string]*localDataEntry
+	ttl      time.Duration
+	stopCh   chan struct{}
+	stopOnce sync.Once
+}
+
+// Store creates or updates session metadata.
+func (s *LocalSessionDataStorage) Store(_ context.Context, id string, metadata map[string]string) error {
+	if id == "" {
+		return fmt.Errorf("cannot store session data with empty ID")
+	}
+	if metadata == nil {
+		metadata = make(map[string]string)
+	}
+	// Store a defensive copy so callers cannot mutate stored data.
+	copied := maps.Clone(metadata)
+	s.sessions.Store(id, newLocalDataEntry(copied))
+	return nil
+}
+
+// Load retrieves session metadata and refreshes its last-access timestamp.
+// Returns ErrSessionNotFound if the session does not exist.
+func (s *LocalSessionDataStorage) Load(_ context.Context, id string) (map[string]string, error) {
+	if id == "" {
+		return nil, fmt.Errorf("cannot load session data with empty ID")
+	}
+
+	val, ok := s.sessions.Load(id)
+	if !ok {
+		return nil, ErrSessionNotFound
+	}
+	entry, ok := val.(*localDataEntry)
+	if !ok {
+		return nil, fmt.Errorf("invalid entry type in local session data storage")
+	}
+
+	// Refresh last-access by swapping in a new entry pointer. This prevents
+	// a concurrent DeleteExpired from evicting a session that is actively in use.
+	newEntry := newLocalDataEntry(entry.metadata)
+	s.sessions.CompareAndSwap(id, entry, newEntry)
+	// If CAS fails, another goroutine already replaced the entry; the map
+	// always holds a fresh pointer, so eviction of this session is safe.
+
+	return maps.Clone(entry.metadata), nil
+}
+
+// Exists reports whether a session is present without refreshing its last-access
+// timestamp. This is safe for eviction probes: it will not extend the lifetime
+// of an idle session.
+func (s *LocalSessionDataStorage) Exists(_ context.Context, id string) (bool, error) {
+	if id == "" {
+		return false, fmt.Errorf("cannot check existence of session data with empty ID")
+	}
+	_, ok := s.sessions.Load(id)
+	return ok, nil
+}
+
+// StoreIfAbsent atomically creates session metadata only if the session ID
+// does not already exist. Uses sync.Map.LoadOrStore for atomicity.
+func (s *LocalSessionDataStorage) StoreIfAbsent(_ context.Context, id string, metadata map[string]string) (bool, error) {
+	if id == "" {
+		return false, fmt.Errorf("cannot store session data with empty ID")
+	}
+	if metadata == nil {
+		metadata = make(map[string]string)
+	}
+	copied := maps.Clone(metadata)
+	_, loaded := s.sessions.LoadOrStore(id, newLocalDataEntry(copied))
+	return !loaded, nil
+}
+
+// Delete removes session metadata. Not an error if absent.
+func (s *LocalSessionDataStorage) Delete(_ context.Context, id string) error {
+	if id == "" {
+		return fmt.Errorf("cannot delete session data with empty ID")
+	}
+	s.sessions.Delete(id)
+	return nil
+}
+
+// Close stops the background cleanup goroutine and clears all stored metadata.
+func (s *LocalSessionDataStorage) Close() error {
+	s.stopOnce.Do(func() { close(s.stopCh) })
+	s.sessions.Range(func(key, _ any) bool {
+		s.sessions.Delete(key)
+		return true
+	})
+	return nil
+}
+
+// minCleanupInterval is the floor applied to the cleanup ticker interval.
+// time.NewTicker panics when given a duration ≤ 0, so any TTL smaller than
+// 2ns would produce ttl/2 == 0 without this guard. 1ms is a practical floor
+// that avoids the panic without restricting legitimate short TTLs in tests.
+const minCleanupInterval = time.Millisecond
+
+func (s *LocalSessionDataStorage) cleanupRoutine() {
+	if s.ttl <= 0 {
+		return
+	}
+	interval := s.ttl / 2
+	if interval < minCleanupInterval {
+		interval = minCleanupInterval
+	}
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			s.deleteExpired()
+		case <-s.stopCh:
+			return
+		}
+	}
+}
+
+func (s *LocalSessionDataStorage) deleteExpired() {
+	cutoff := time.Now().Add(-s.ttl)
+	var toDelete []struct {
+		id    string
+		entry *localDataEntry
+	}
+	s.sessions.Range(func(key, val any) bool {
+		entry, ok := val.(*localDataEntry)
+		if ok && entry.lastAccess().Before(cutoff) {
+			id, ok := key.(string)
+			if ok {
+				toDelete = append(toDelete, struct {
+					id    string
+					entry *localDataEntry
+				}{id, entry})
+			}
+		}
+		return true
+	})
+	for _, item := range toDelete {
+		if item.entry.lastAccess().Before(cutoff) {
+			s.sessions.CompareAndDelete(item.id, item.entry)
+		}
+	}
+}
--- a/pkg/transport/session/storage.go
+++ b/pkg/transport/session/storage.go
@@ -26,12 +26,12 @@ type Storage interface {
 	// that DeleteExpired does not evict sessions that are actively being accessed.
 	Load(ctx context.Context, id string) (Session, error)
 
-	// Peek reports whether a session exists without refreshing its eviction TTL.
+	// Exists reports whether a session exists without refreshing its eviction TTL.
 	// Returns (true, nil) if the session exists, (false, nil) if it definitively
 	// does not exist, and (false, err) if the check could not be completed (e.g.
 	// storage unavailable). Callers must treat non-nil errors as "unknown"
 	// rather than "not found".
-	Peek(ctx context.Context, id string) (bool, error)
+	Exists(ctx context.Context, id string) (bool, error)
 
 	// Delete removes a session from the storage backend.
 	// It is not an error if the session doesn't exist.
--- a/pkg/vmcp/discovery/middleware.go
+++ b/pkg/vmcp/discovery/middleware.go
@@ -25,15 +25,21 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
-	"strings"
 	"time"
 
-	transportsession "github.com/stacklok/toolhive/pkg/transport/session"
 	"github.com/stacklok/toolhive/pkg/vmcp"
 	"github.com/stacklok/toolhive/pkg/vmcp/aggregator"
 	"github.com/stacklok/toolhive/pkg/vmcp/health"
 	vmcpsession "github.com/stacklok/toolhive/pkg/vmcp/session"
 )
+
+// MultiSessionGetter retrieves live MultiSession objects by session ID.
+// Implemented by sessionmanager.Manager to allow the discovery middleware to
+// obtain routing tables for subsequent requests without depending on the
+// transport-layer session storage.
+type MultiSessionGetter interface {
+	GetMultiSession(sessionID string) (vmcpsession.MultiSession, bool)
+}
 
 const (
 	// discoveryTimeout is the maximum time for capability discovery.
@@ -87,7 +93,7 @@ func WithDiscoveryTimeout(timeout time.Duration) MiddlewareOption {
 func Middleware(
 	manager Manager,
 	registry vmcp.BackendRegistry,
-	sessionManager *transportsession.Manager,
+	sessionGetter MultiSessionGetter,
 	healthStatusProvider health.StatusProvider,
 	opts ...MiddlewareOption,
 ) func(http.Handler) http.Handler {
@@ -114,7 +120,7 @@ func Middleware(
 				ctx, err = handleInitializeRequest(ctx, r, manager, registry, healthStatusProvider, cfg.timeout)
 			} else {
 				// Subsequent request: retrieve cached capabilities from session.
-				ctx, err = handleSubsequentRequest(ctx, r, sessionID, sessionManager)
+				ctx = handleSubsequentRequest(ctx, r, sessionID, sessionGetter)
 			}
 
 			if err != nil {
@@ -257,48 +263,29 @@ func handleInitializeRequest(
 }
 
 // handleSubsequentRequest retrieves cached capabilities from the session.
-// Returns updated context with capabilities or an error.
+// Returns updated context with capabilities injected when available.
 func handleSubsequentRequest(
 	ctx context.Context,
 	r *http.Request,
 	sessionID string,
-	sessionManager *transportsession.Manager,
-) (context.Context, error) {
+	getter MultiSessionGetter,
+) context.Context {
 	//nolint:gosec // G706: session ID and request fields are not injection vectors
 	slog.Debug("retrieving capabilities from session for subsequent request",
 		"session_id", sessionID,
 		"method", r.Method,
 		"path", r.URL.Path)
 
-	// First, validate the session exists at all.
-	rawSess, exists := sessionManager.Get(sessionID)
-	if !exists {
+	// Retrieve the live MultiSession. Returns (nil, false) when the session
+	// doesn't exist or is still a placeholder (Phase 2 not yet complete).
+	// In either case, skip capability injection — the SDK's own session
+	// validation will reject truly unknown session IDs downstream.
+	multiSess, ok := getter.GetMultiSession(sessionID)
+	if !ok {
 		//nolint:gosec // G706: session ID is not an injection vector
-		slog.Error("session not found",
-			"session_id", sessionID,
-			"method", r.Method,
-			"path", r.URL.Path)
-		return ctx, fmt.Errorf("session not found: %s", sessionID)
-	}
-
-	// Backend tool handlers (created by DefaultHandlerFactory) resolve their backend
-	// target by calling router.RouteTool(ctx, name), which reads DiscoveredCapabilities
-	// from the request context. Inject capabilities built from the session's routing
-	// table so these handlers can route correctly on subsequent requests.
-	// Note: composite tool workflow engines are created per-session and route via
-	// SessionRouter directly, so they no longer depend on this context value.
-	multiSess, isMulti := rawSess.(vmcpsession.MultiSession)
-	if !isMulti {
-		// The session is still a StreamableSession placeholder — Phase 2
-		// (OnRegisterSession / CreateSession) has not yet replaced it with a
-		// MultiSession. This can happen if the client sends a request in the
-		// brief window between receiving the session ID and the hook completing.
-		// Skip capability injection and let the SDK respond (tools list will be
-		// temporarily empty, but no 500 is returned to the client).
-		//nolint:gosec // G706: session ID is not an injection vector
-		slog.Debug("session initialisation in progress, skipping capability injection",
+		slog.Debug("multi-session not available, skipping capability injection",
 			"session_id", sessionID)
-		return ctx, nil
+		return ctx
 	}
 
 	routingTable := multiSess.GetRoutingTable()
@@ -309,7 +296,7 @@ func handleSubsequentRequest(
 		//nolint:gosec // G706: session ID is not an injection vector
 		slog.Debug("multi-session routing table not yet initialised; skipping capability injection",
 			"session_id", sessionID)
-		return ctx, nil
+		return ctx
 	}
 	//nolint:gosec // G706: session ID is not an injection vector
 	slog.Debug("injecting capabilities from multi-session routing table for composite tool routing",
@@ -319,7 +306,7 @@ func handleSubsequentRequest(
 		RoutingTable: routingTable,
 		Tools:        multiSess.Tools(),
 	}
-	return WithDiscoveredCapabilities(ctx, capabilities), nil
+	return WithDiscoveredCapabilities(ctx, capabilities)
 }
 
 // handleDiscoveryError writes appropriate HTTP error responses based on the error type.
@@ -328,14 +315,5 @@ func handleDiscoveryError(w http.ResponseWriter, _ *http.Request, err error) {
 		http.Error(w, http.StatusText(http.StatusGatewayTimeout), http.StatusGatewayTimeout)
 		return
 	}
-
-	// Check for session-related errors
-	errMsg := err.Error()
-	if strings.Contains(errMsg, "session not found") {
-		http.Error(w, "Session not found", http.StatusUnauthorized)
-		return
-	}
-
-	// Default to service unavailable for other errors
 	http.Error(w, http.StatusText(http.StatusServiceUnavailable), http.StatusServiceUnavailable)
 }
--- a/pkg/vmcp/server/sessionmanager/session_manager.go
+++ b/pkg/vmcp/server/sessionmanager/session_manager.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/mark3labs/mcp-go/mcp"
 	mcpserver "github.com/mark3labs/mcp-go/server"
+	"golang.org/x/sync/singleflight"
 
 	"github.com/stacklok/toolhive/pkg/auth"
 	transportsession "github.com/stacklok/toolhive/pkg/transport/session"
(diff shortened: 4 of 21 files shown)


pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -114,38 +136,67 @@ func New(
...
 		}
+		if exists {
+			slog.Warn("Manager: UUID collision detected; retrying", "session_id", sessionID)
+			continue
+		}
+
+		if err := sm.storage.Store(ctx, sessionID, map[string]string{}); err != nil {
+			slog.Error("Manager: failed to store placeholder session",
+				"session_id", sessionID, "attempt", attempt+1, "error", err)
+			continue
```
JAORMX: So, the `Exists` + `Store` pair here isn't atomic. With Redis across multiple pods, two pods could both see `Exists=false` for the same UUID and both call `Store`, with the second silently overwriting the first.

UUID collisions are astronomically unlikely, sure. But the code *explicitly* adds this guard for that scenario, so it should actually work. Redis has `SET ... NX` (set-if-not-exists) which would make this truly atomic. As-is, the guard gives a false sense of protection while leaving the exact race it's trying to prevent. Consider adding a `StoreIfAbsent` to `DataStorage` that maps to `SetNX` on Redis.

Also, `context.Background()` here with no timeout... if Redis is unreachable, `Exists` and `Store` will hang for however long go-redis's default timeout is. The old code path used `context.WithTimeout`. Worth adding that back.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -259,13 +320,16 @@ func (sm *Manager) Validate(sessionID string) (isTerminated bool, err error) {
 		return false, fmt.Errorf("Manager.Validate: empty session ID")
 	}
 
-	sess, exists := sm.storage.Get(sessionID)
-	if !exists {
+	metadata, err := sm.storage.Load(context.Background(), sessionID)
```
JAORMX: `sm.storage.Load(context.Background(), sessionID)` with no timeout.

The SDK calls `Validate()` on every incoming HTTP request. If Redis is slow or unreachable, this blocks the request goroutine indefinitely (or until go-redis's default timeout, which can be long). The old code used `context.WithTimeout`. Worth wrapping this the same way.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -288,60 +352,78 @@ func (sm *Manager) Validate(sessionID string) (isTerminated bool, err error) {
...
+			// transient storage error (e.g. Redis unavailable) leaves both
+			// storage and the local map in a consistent state and the caller can
+			// safely retry. Removing from the local map first would lose the
+			// MultiSession reference on retry, preventing cleanup of the stale
+			// storage entry.
+			if deleteErr := sm.storage.Delete(ctx, sessionID); deleteErr != nil {
+				return false, fmt.Errorf("Manager.Terminate: failed to delete session from storage: %w", deleteErr)
+			}
+			sm.multiSessions.Delete(sessionID)
+			if closeErr := multiSess.Close(); closeErr != nil {
```
JAORMX, line 412: I see the comment explaining why we delete from storage *before* the local map (retry safety). That reasoning makes sense.

But there's a window between `storage.Delete` succeeding and `multiSessions.Delete` where a concurrent goroutine could have *already* loaded the session from the local map (line 370-371), be holding a live reference, and then find the storage entry gone when it later hits `Validate()`. The window is narrow and probably acceptable in practice, but worth acknowledging with a comment.

Could also consider a "terminating" tombstone in the local map to reject concurrent reads during teardown. Not blocking, but worth thinking about for a follow-up.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -352,24 +434,74 @@ func (sm *Manager) Terminate(sessionID string) (isNotAllowed bool, err error) {
...
+	// uncached session would otherwise each open independent backend
+	// connections before any had a chance to populate the local cache.
+	type restoreResult struct {
+		sess vmcpsession.MultiSession
+	}
+	v, err, _ := sm.restoreGroup.Do(sessionID, func() (any, error) {
+		ctx := context.Background()
+		metadata, loadErr := sm.storage.Load(ctx, sessionID)
+		if loadErr != nil {
+			if !errors.Is(loadErr, transportsession.ErrSessionNotFound) {
```
JAORMX, line 500: `context.Background()` in the singleflight callback here too. `RestoreSession` opens backend HTTP connections, so if a backend is down this blocks with no cancellation from the caller's HTTP request context.

I know `GetMultiSession`'s signature (from the interface) doesn't accept a context, so this is a design constraint. Worth documenting as a known limitation. Maybe a follow-up to add context propagation through `MultiSessionGetter`?

pkg/transport/session/storage.go
```diff
@@ -26,6 +26,13 @@ type Storage interface {
 	// that DeleteExpired does not evict sessions that are actively being accessed.
 	Load(ctx context.Context, id string) (Session, error)
 
+	// Peek reports whether a session exists without refreshing its eviction TTL.
+	// Returns (true, nil) if the session exists, (false, nil) if it definitively
+	// does not exist, and (false, err) if the check could not be completed (e.g.
+	// storage unavailable). Callers must treat non-nil errors as "unknown"
+	// rather than "not found".
+	Peek(ctx context.Context, id string) (bool, error)
```
JAORMX: Is `Peek` called anywhere in this PR? I couldn't find a caller. Adding methods to a stable interface is a breaking change for any external/test implementations. If it's for future use, I'd rather it wait until there's an actual consumer.

Also, naming nit: the new `DataStorage` interface uses `Exists()` for the same concept (check presence without refreshing TTL). Having both `Peek` and `Exists` for the same thing on different interfaces is confusing. If we do keep both, let's align the naming.

pkg/vmcp/discovery/middleware.go
JAORMX, line 1: Heads up: the behavior for terminated/unknown sessions changed from HTTP 401 to HTTP 404. The middleware no longer returns an error for unknown sessions in `handleSubsequentRequest`... it lets the request through to the SDK, which returns 404.

This is arguably more accurate ("session not found" vs "unauthorized"), but it's a user-visible change. Any monitoring or clients matching on 401 for session expiry will break. Worth calling this out in the PR description.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -537,3 +673,76 @@ func (sm *Manager) GetAdaptedResources(sessionID string) ([]mcpserver.ServerReso
...
+		// Definitive not-found: the storage backend has expired this session.
+		sm.multiSessions.Delete(sessionID)
+		if multiSess, ok := value.(vmcpsession.MultiSession); ok {
+			if closeErr := multiSess.Close(); closeErr != nil {
+				slog.Warn("eviction: error closing expired multi-session",
+					"session_id", sessionID, "error", closeErr)
+			}
+		}
+		slog.Debug("eviction: removed expired multi-session from node-local map",
+			"session_id", sessionID)
```
JAORMX, line 792: The eviction loop calls `storage.Exists()` for every entry in the `multiSessions` map on every tick (TTL/2 interval). With Redis, that's one `EXISTS` call per live session per tick. Fine at current scale, but worth keeping an eye on if session counts grow... batching with a Redis pipeline would help there.

More immediately: if a session is concurrently in use (mid `CallTool` RPC) when eviction fires `Close()`, that yanks HTTP connections mid-flight. A grace period or ref-counting would be the proper fix, but might be overkill right now. At minimum, worth a `Warn`-level log when evicting so it's debuggable.

pkg/transport/session/session_data_storage_local.go
JAORMX, line 1: Both `LocalSessionDataStorage` and `RedisSessionDataStorage` have non-trivial logic (TTL cleanup, CAS-based eviction, background goroutine lifecycle, serialization/deserialization) but no dedicated unit tests. The session manager tests exercise them indirectly, but dedicated tests covering Store/Load round-trip, Load miss, Exists for present/absent keys, Delete idempotency, and TTL-based eviction would really help here.

pkg/vmcp/discovery/middleware.go
yrobla: Good catch. Updated the PR description to call this out explicitly under "Does this introduce a user-facing change?" — notes the 401→404 shift, why it happened, and what callers matching on 401 for session expiry need to update.

PR #4404, opened 2026-03-27
Update Redis metadata when backend sessions expire (RC-16)

### LINKED ISSUE #4213: Update Redis Metadata When Backend Sessions Expire (RC-16)
## Description

Update `pkg/vmcp/session/default_session.go` and `pkg/vmcp/server/sessionmanager/session_manager.go` so that when a backend session expires or is lost within a vMCP session, the corresponding Redis metadata keys (`vmcp.backend.session.{workloadID}` and `vmcp.backend.ids`) are removed or updated atomically. This keeps the Redis metadata consistent with the actual state of connected backends, preventing `RestoreSession` (RC-9) from attempting reconnection to backend sessions that are no longer valid.

## Context

The vMCP horizontal scaling epic (THV-0047) stores per-backend session IDs in Redis metadata so that any pod can reconstruct a session on an LRU cache miss. RC-8 (TASK-003, #273) established the convention of writing `MetadataKeyBackendSessionPrefix+workloadID` into transport-session metadata at session creation time. However, backends are not static — they can disconnect, expire, or be explicitly removed during the lifetime of a vMCP session.

Without RC-16, Redis metadata may describe backends that no longer exist at the time `RestoreSession` is invoked. A reconstructing pod would attempt to connect to those stale `backend_session_id` hints, producing errors or silently connecting to the wrong backend state. RC-16 closes this gap by ensuring that any backend-session loss event (expiry, explicit close, removal) propagates to the Redis metadata layer before the in-memory `MultiSession` is evicted.

Parent epic: stacklok/stacklok-epics#262

**Dependencies**: #273 (RC-8: Persist vMCP Session Metadata to Redis on Creation)
**Blocks**: None (leaf task in DAG)

## Acceptance Criteria

- [ ] When a backend session within a `defaultMultiSession` expires or is detected as lost, `vmcp.backend.session.{workloadID}` is removed from the transport-session metadata (via `transportsession.Session.SetMetadata` with an empty value or by calling `storage.UpsertSession` after mutation)
- [ ] After the above removal, `vmcp.backend.ids` is refreshed to reflect only the workload IDs of backends that still have valid sessions (i.e., remaining connected backends only)
- [ ] The metadata update is persisted to the `transportsession.Manager` (which writes through to Redis when `RedisStorage` is configured) before or at the point of connection teardown
- [ ] When the last backend session is lost, `vmcp.backend.ids` is set to an empty string or the key is cleared (consistent with the existing behavior in `populateBackendMetadata` when no backends connect)
- [ ] The expiry/loss path in `sessionmanager.Manager` (`pkg/vmcp/server/sessionmanager/session_manager.go`) calls `storage.UpsertSession` with the updated `MultiSession` to propagate the metadata mutation to the storage backend
- [ ] `RestoreSession` (TASK-005/RC-9) receiving metadata updated by this task does not attempt to reconnect to cleared/removed backend IDs
- [ ] Existing session creation and termination flows are unaffected (no regression in `makeSession`, `CreateSession`, or `Terminate`)
- [ ] Unit tests cover: (a) single-backend expiry removes the correct metadata keys, (b) partial expiry (some backends remain) refreshes `vmcp.backend.ids` with only surviving IDs, (c) full expiry clears both keys
- [ ] All tests pass

The implementation has two distinct change sites:

**1. `pkg/vmcp/session/default_session.go`** — Add a method (or extend `Close`) that, before or alongside closing a backend connection, calls `SetMetadata` on the embedded `transportsession.Session` to clear `MetadataKeyBackendSessionPrefix+workloadID` and rebuild `MetadataKeyBackendIDs` from remaining entries. Because `defaultMultiSession.backendSessions` is a snapshot map populated at creation, it can be used to enumerate which workload IDs are being removed. The metadata mutation itself is synchronous and does not require a lock (the `backendSessions` map is write-once after session creation).

**2. `pkg/vmcp/server/sessionmanager/session_manager.go`** — At the point where a backend-session loss is detected (e.g., during a `CallTool` / `ReadResource` / `GetPrompt` error that indicates the backend is gone, or during session cleanup), call `storage.UpsertSession(multiSess)` to flush the updated metadata from the in-memory `MultiSession` into the `transportsession.Manager` (and therefore Redis). The existing `Terminate` path already calls `storage.Delete` for full session termination; the new path covers the partial case where only a subset of backends are lost.

**Important**: the `defaultMultiSession.backendSessions` map is a `map[string]string` that maps workload IDs to backend session IDs. After a backend is disconnected, this map entry should be cleared (`delete(s.backendSessions, workloadID)`) and the metadata updated accordingly. Since `backendSessions` is written once after session creation and the map fields are documented as read-only after that, a small targeted update here is safe as long as it is bounded to the teardown path where no concurrent I/O operations can be in-flight (admission queue is drained or the specific backend is already unreachable).

### Patterns & Frameworks

- Follow the metadata mutation pattern already established in `populateBackendMetadata` in `pkg/vmcp/session/factory.go`: iterate the surviving workload IDs, join them with commas, and write to `MetadataKeyBackendIDs`
- Use `strings.Join` for rebuilding `MetadataKeyBackendIDs` (consistent with `populateBackendMetadata`)
- Use `transportsession.Session.SetMetadata(key, "")` to clear individual `MetadataKeyBackendSessionPrefix+workloadID` keys — empty string is the conventional "cleared" sentinel since the `Session` metadata map is `map[string]string`
- Call `storage.UpsertSession` in `sessionmanager.Manager` to flush to Redis; this is the same call used in `CreateSession` (line 200 of `session_manager.go`) to replace placeholder sessions
- Follow the `slog`-based structured logging pattern used throughout both files for observability (`slog.Info` or `slog.Debug` for normal expiry events, `slog.Warn` for unexpected failures)

### Code Pointers

- `pkg/vmcp/session/default_session.go` — Primary change site. The `defaultMultiSession` struct holds `backendSessions map[string]string` (workload ID → backend session ID) and embeds `transportsession.Session` for metadata. The `Close()` method at line 201 iterates `connections` and calls `conn.Close()` for each — the metadata update should occur here or in a new `removeBackend(workloadID string)` helper called from the relevant teardown path
- `pkg/vmcp/session/factory.go:354-362` — `populateBackendMetadata`: reference implementation for how `MetadataKeyBackendIDs` and `MetadataKeyBackendSessionPrefix+workloadID` should be written; the removal logic is the inverse of this function
- `pkg/vmcp/session/factory.go:26-38` — `MetadataKeyBackendIDs` and the new `MetadataKeyBackendSessionPrefix` constant (added by RC-8/TASK-003) — use these constants; do not hard-code string literals
- `pkg/vmcp/server/sessionmanager/session_manager.go:196-210` — `CreateSession`: shows the `storage.UpsertSession(sess)` call pattern that persists a mutated session to the transport manager and through to Redis
- `pkg/vmcp/server/sessionmanager/session_manager.go:258-311` — `Terminate`: reference for how full session teardown is handled; RC-16 handles the subset (partial backend expiry) case
- `pkg/vmcp/session/internal/security/security.go` — `hijackPreventionDecorator` wraps `defaultMultiSession`; any new method added to `defaultMultiSession` that must be accessible externally will need to be exposed through the decorator or added to the `MultiSession` interface in `pkg/vmcp/session/types/session.go`

### Component Interfaces

The key design decision is where to place the metadata-update logic. Two approaches are viable:

**Option A — Method on `defaultMultiSession`** (preferred): Add a `removeBackendFromMetadata(workloadID string)` unexported helper that clears `MetadataKeyBackendSessionPrefix+workloadID` and rebuilds `MetadataKeyBackendIDs`. Call this from `Close()` before or alongside `conn.Close()`. The `sessionmanager.Manager` then calls `storage.UpsertSession(multiSess)` immediately after the backend connection teardown triggers.

```go
// removeBackendFromMetadata clears per-backend session metadata for workloadID
// and refreshes MetadataKeyBackendIDs to reflect only surviving backends.
// Must be called while the session is still accessible (before storage.Delete).
func (s *defaultMultiSession) removeBackendFromMetadata(workloadID string) {
    // Clear the per-backend session ID entry
    s.SetMetadata(MetadataKeyBackendSessionPrefix+workloadID, "")
    // Remove from the runtime map
    delete(s.backendSessions, workloadID)
    // Rebuild the comma-separated list of remaining backend IDs
    remaining := make([]string, 0, len(s.backendSessions))
    for id := range s.backendSessions {
        remaining = append(remaining, id)
    }
    sort.Strings(remaining) // deterministic ordering
    s.SetMetadata(MetadataKeyBackendIDs, strings.Join(remaining, ","))
}
```

**Option B — In `sessionmanager.Manager` only**: Mutate the metadata map directly via `multiSess.SetMetadata(...)` from the session manager. This avoids adding methods to `defaultMultiSession` but mixes concerns.

Option A is preferred because it keeps metadata management co-located with the data it describes.

The `sessionmanager.Manager` must then call `storage.UpsertSession` to flush the mutation:

```go
// In sessionmanager.Manager, after detecting backend loss for workloadID:
multiSess.removeBackendFromMetadata(workloadID) // via exposed helper or interface method
if err := sm.storage.UpsertSession(multiSess); err != nil {
    slog.Warn("failed to persist backend expiry to storage",
        "session_id", multiSess.ID(),
        "workload_id", workloadID,
        "error", err)
}
```

### Testing Strategy

**Unit Tests**

- [ ] Test: single backend expiry — after calling `removeBackendFromMetadata("backend-a")`, `GetMetadata()[MetadataKeyBackendSessionPrefix+"backend-a"]` is empty and `GetMetadata()[MetadataKeyBackendIDs]` does not contain `"backend-a"`
- [ ] Test: partial expiry — session created with two backends `{"a", "b"}`; after `removeBackendFromMetadata("a")`, `MetadataKeyBackendIDs` equals `"b"` and `MetadataKeyBackendSessionPrefix+"b"` is still present
- [ ] Test: full expiry — after all backends are removed via `removeBackendFromMetadata`, `MetadataKeyBackendIDs` is empty string and no `MetadataKeyBackendSessionPrefix+*` keys contain non-empty values
- [ ] Test: `storage.UpsertSession` is called in `sessionmanager.Manager` after backend loss, so the mutation is persisted (use a mock `transportsession.Manager` or spy `Storage`)
- [ ] Test: `Close()` on `defaultMultiSession` still works correctly after one backend has already been removed (idempotency)

**Integration Tests**

- [ ] Test: backend session expiry propagates to Redis metadata; a subsequent `RestoreSession` call does not attempt to reconnect to the expired backend and does not return an error for that specific backend (tolerates missing backends per RC-9 partial-initialization behavior)

**Edge Cases**

- [ ] Remove a workload ID that is not present in `backendSessions` — must be a no-op (no panic, no spurious metadata writes)
- [ ] Concurrent calls to `removeBackendFromMetadata` for different workload IDs — consider whether a mutex is needed given that `Close()` currently holds no lock (the `CloseAndDrain` pattern serializes in-flight ops but not concurrent `Close()` calls)
- [ ] Empty `backendSessions` map when `removeBackendFromMetadata` is called — `MetadataKeyBackendIDs` must be set to empty string, not omitted

## Out of Scope

- Implementing `RestoreSession` itself (TASK-005/RC-9)
- Adding the LRU cache to `sessionmanager.Manager` (TASK-006/RC-10)
- Any changes to `pkg/transport/session/storage_redis.go` or `pkg/transport/session/manager.go`
- Proxy Runner horizontal scaling
- Kubernetes / Helm deployment configuration

## References

- RFC THV-0047: https://github.com/stacklok/toolhive-rfcs/pull/47
- Parent epic: stacklok/stacklok-epics#262
- RC-8 (upstream dependency): stacklok/stacklok-epics#273
- Architecture document: `/Users/yolanda/development/toolhive/epic-planning-20260317-103930/vmcp/architecture.md`

yrobla:
## Summary

Add RemoveBackendFromMetadata to the MultiSession interface and implement it on defaultMultiSession to clear the per-backend session ID key and rebuild MetadataKeyBackendIDs when a backend disconnects or expires.

Add NotifyBackendExpired to sessionmanager.Manager to call RemoveBackendFromMetadata and flush the updated metadata to storage (Redis when configured), keeping Redis consistent with the actual set of connected backends.

Fixes #4213 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
|      |        |

### DIFF
--- a/pkg/vmcp/server/sessionmanager/session_manager.go
+++ b/pkg/vmcp/server/sessionmanager/session_manager.go
@@ -197,6 +197,13 @@ const terminateTimeout = 5 * time.Second
 // performs a single Redis SET. 5 s is consistent with terminateTimeout.
 const decorateTimeout = 5 * time.Second
 
+// notifyBackendExpiredTimeout bounds each individual storage operation inside
+// NotifyBackendExpired() — one Load and one Upsert, each capped independently.
+// Each is a single-key Redis operation, so 5 s per call is consistent with
+// terminateTimeout and decorateTimeout. Worst-case wall-clock for the function
+// is 2 × 5 s = 10 s.
+const notifyBackendExpiredTimeout = 5 * time.Second
+
 // Generate implements the SDK's SessionIdManager.Generate().
 //
 // Phase 1 of the two-phase creation pattern: creates a unique session ID,
@@ -520,17 +527,17 @@ func (sm *Manager) Terminate(sessionID string) (isNotAllowed bool, err error) {
 // cross-pod RestoreSession call does not attempt to reconnect to the expired
 // backend session.
 //
-// If the session is live in the node-local cache, RemoveBackendFromMetadata is
-// called on it to keep the in-memory state consistent. After a successful
-// storage update the session is evicted; the next GetMultiSession call will
-// trigger RestoreSession with the updated metadata.
+// After a successful storage update the session is evicted from the node-local
+// cache; the next GetMultiSession call triggers RestoreSession with the updated
+// metadata, discarding the stale in-memory copy.
 //
 // This is a best-effort operation. If the session is absent from storage (not
 // found or terminated) the call is a silent no-op. Storage errors are logged
 // but not returned; on error the cache is not evicted.
 func (sm *Manager) NotifyBackendExpired(sessionID, workloadID string) {
-	ctx := context.Background()
-	metadata, err := sm.storage.Load(ctx, sessionID)
+	loadCtx, loadCancel := context.WithTimeout(context.Background(), notifyBackendExpiredTimeout)
+	defer loadCancel()
+	metadata, err := sm.storage.Load(loadCtx, sessionID)
 	if err != nil {
 		if !errors.Is(err, transportsession.ErrSessionNotFound) {
 			slog.Warn("NotifyBackendExpired: failed to load session from storage",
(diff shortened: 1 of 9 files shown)


jerm-dro:
This PR overlaps significantly with #4464 (which has been approved). Once #4464 merges, this will need a rebase — expect substantial conflicts in `session_manager.go`, `factory.go`, `server.go`, `middleware.go`, and others. Happy to re-review after rebase.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -513,6 +521,116 @@ func (sm *Manager) Terminate(sessionID string) (isNotAllowed bool, err error) {
...
+// cross-pod RestoreSession call does not attempt to reconnect to the expired
+// backend session.
+//
+// After a successful storage update the session is evicted from the node-local
+// cache; the next GetMultiSession call triggers RestoreSession with the updated
+// metadata, discarding the stale in-memory copy.
+//
+// This is a best-effort operation. If the session is absent from storage (not
+// found or terminated) the call is a silent no-op. Storage errors are logged
+// but not returned; on error the cache is not evicted.
```
jerm-dro, line 536: `question:` This does a `storage.Load` as its first step, but there are no callers yet. Any caller that detects a backend expiry will almost certainly already have the session metadata or `MultiSession` in hand — otherwise it wouldn't know the backend expired. Consider accepting the metadata as a parameter to skip the redundant Load.

Also, I'd prefer backend expiry to be handled lazily (updating metadata when the session is next accessed or used) rather than via a health-checking mechanism that proactively scans for expired backends. Lazy updates are simpler and avoid the synchronization complexity of background health checkers.

Not blocking — this can be addressed in a follow-up when the caller is introduced.

yrobla: Will solve in issue #4667

PR #4410, opened 2026-03-27
Webhook Middleware Phase 4: CLI configuration support

### LINKED ISSUE #3400: Webhook Middleware Phase 4: CLI configuration support
## Overview

Add CLI support for configuring webhook middleware via command-line flags and configuration files. This enables users to easily configure validating and mutating webhooks when running MCP servers with `thv run`.

**Depends on**: Phase 2 (Validating webhook), Phase 3 (Mutating webhook)

## Files to Create

| File | Purpose |
|------|---------|
| `pkg/webhook/config.go` | YAML/JSON config file parsing and validation |

## Files to Modify

| File | Changes |
|------|---------|
| `cmd/thv/app/run_flags.go` | Add `--webhook-config` flag to `RunFlags` struct |
| `cmd/thv/app/run.go` | Parse webhook config and populate RunConfig |

## CLI Usage

```bash
# Single webhook config file
thv run postgres-mcp --webhook-config webhooks.yaml

# Multiple webhook config files (combined)
thv run postgres-mcp \
  --webhook-config policy-webhook.yaml \
  --webhook-config enricher-webhook.yaml
```

## Configuration File Format

```yaml
# webhooks.yaml
validating:
  - name: policy-check
    url: https://policy.example.com/validate
    timeout: 5s
    failure_policy: fail
    tls_config:
      ca_bundle: /path/to/ca.crt
      # skip_verify: false  # for development only
    hmac_secret_ref: webhook-secret,target=WEBHOOK_HMAC

  - name: rate-limiter
    url: https://ratelimit.example.com/check
    timeout: 2s
    failure_policy: ignore

mutating:
  - name: hr-enrichment
    url: https://hr-api.example.com/enrich
    timeout: 3s
    failure_policy: ignore

  - name: cmdb-enrichment
    url: https://cmdb-api.example.com/enrich
    timeout: 3s
    failure_policy: ignore
    tls_config:
      client_cert: /path/to/cert.pem
      client_key: /path/to/key.pem
```

## Config Parsing Implementation

```go
// pkg/webhook/config.go

type FileConfig struct {
    Validating []WebhookConfig `yaml:"validating" json:"validating"`
    Mutating   []WebhookConfig `yaml:"mutating" json:"mutating"`
}

func LoadConfig(path string) (*FileConfig, error)
func MergeConfigs(configs ...*FileConfig) *FileConfig
func ValidateConfig(config *FileConfig) error
```

## Flag Integration

```go
// cmd/thv/app/run_flags.go
type RunFlags struct {
    // ... existing flags ...
    
    // WebhookConfigs is a list of paths to webhook configuration files
    WebhookConfigs []string
}

// In flag registration
cmd.Flags().StringArrayVar(&flags.WebhookConfigs, "webhook-config", nil,
    "Path to webhook configuration file (can be specified multiple times)")
```

## RunConfig Population

In `cmd/thv/app/run.go`, after parsing flags:

1. Load each webhook config file
2. Merge configs (later files take precedence for same-named webhooks)
3. Validate merged config
4. Populate `RunConfig.ValidatingWebhooks` and `RunConfig.MutatingWebhooks`

## Validation Rules

- URL must be valid HTTPS URL (HTTP only allowed with explicit flag for dev)
- Timeout must be between 1s and 30s
- Failure policy must be "fail" or "ignore"
- TLS config paths must exist if specified
- HMAC secret ref must be valid secret reference format

## Error Messages

Clear error messages for common issues:
- "webhook config file not found: %s"
- "invalid webhook URL: must use HTTPS"
- "webhook timeout %s exceeds maximum of 30s"
- "TLS CA bundle file not found: %s"
- "invalid HMAC secret reference format"

## Tests

- Flag parsing tests
- Config file loading tests (YAML and JSON)
- Config merging tests
- Validation tests (invalid URLs, missing required fields)
- Integration test: `thv run` with webhook config

## Documentation

Update CLI documentation:
- Add `--webhook-config` to `thv run --help`
- Add example webhook config files to docs/examples/
- Update docs/middleware.md with webhook middleware section

## Acceptance Criteria

- [ ] `--webhook-config` flag added to `thv run`
- [ ] Supports YAML and JSON config formats
- [ ] Multiple config files can be specified and merged
- [ ] Clear validation error messages
- [ ] Config correctly populates RunConfig
- [ ] Unit tests with >80% coverage
- [ ] CLI documentation updated

Sanskarzz:
## Summary

This PR implements **Phase 4** of the Dynamic Webhook Middleware RFC (THV-0017), introducing CLI configuration support for passing validating and mutating webhooks to `thv run`.

Users can now define their webhook pipelines in YAML or JSON files and inject them into the ToolHive runner at startup using the `--webhook-config` flag.

Fixes #3400 

## Large PR Justification

### `pkg/webhook/config.go`
* **Parsing**: Introduced the `FileConfig` struct to unmarshal both YAML and JSON webhook configuration files (auto-detected via file extension).
* **Merging**: Implemented `MergeConfigs` to support combining multiple configuration files. Files passed later in the arguments take precedence for webhooks with the same `name` (last-writer-wins deduplication).
* **Validation**: Implemented `ValidateConfig` to aggregate `wh.Validate()` errors across all provided mutating and validating webhook definitions to fail fast at CLI startup.

### `cmd/thv/app/run_flags.go`
* Added `WebhookConfigs []string` property to `RunFlags`.
* Registered the `--webhook-config` flag which can be specified multiple times.
* Added the `loadAndMergeWebhookConfigs` helper to parse and process the files into a single `webhook.FileConfig`.

### `pkg/runner/config_builder.go`
* Added `WithValidatingWebhooks` and `WithMutatingWebhooks` builder functions to inject parsed webhooks into `RunConfig`.
* Fixed a gap in CLI runner initialization by updating `WithMiddlewareFromFlags` to actively instantiate and append `addMutatingWebhookMiddleware` and `addValidatingWebhookMiddleware` directly to the `thv` runner chain (matching the existing Operator behavior). 

### `pkg/webhook/types.go`
* **JSON timeout support**: Added custom JSON unmarshalling for webhook configs so JSON files accept human-readable duration strings like `"5s"` for `timeout`, matching the YAML experience.
* **Stricter validation**: Tightened validation to enforce the Phase 4 timeout contract (`1s` to `30s`) and fail fast when referenced TLS files do not exist.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Testing

* **Unit Tests**: Expanded coverage for file loading (`yaml` and `json`), JSON duration decoding, parsing errors, file aggregation, deduplication sorting, timeout bounds, and TLS path validation inside `pkg/webhook/config_test.go` and `pkg/webhook/types_test.go`.
* **Linting & Coverage Checks**: Clean `task lint` and passing `task test-coverage` requirements.
* **Manual Verification**: Successfully mapped a valid local configuration file configuring using thv CLI binary `./thv run fetch --webhook-config webhook-config/webhooks.yaml --debug` targets, confirming `auth -> mcp-parser -> mutating-webhook -> validating-webhook` proxy application inside `.thv logs`.

`webhook-config/webhooks.yaml`
```yaml
validating:
  - name: test-validator
    url: https://example.com/validate
    timeout: 5s
    failure_policy: ignore
    tls_config:
      insecure_skip_verify: true
mutating:
  - name: test-mutator
    url: https://example.com/mutate
    timeout: 3s
    failure_policy: ignore
    tls_config:
      insecure_skip_verify: true
```

image

### DIFF
--- a/docs/examples/webhooks.json
+++ b/docs/examples/webhooks.json
@@ -1,0 +1,23 @@
+{
+  "validating": [
+    {
+      "name": "policy-check",
+      "url": "https://policy.example.com/validate",
+      "failure_policy": "fail",
+      "timeout": "5s",
+      "tls_config": {
+        "ca_bundle_path": "/etc/toolhive/pki/webhook-ca.crt"
+      }
+    }
+  ],
+  "mutating": [
+    {
+      "name": "request-enricher",
+      "url": "https://enrichment.example.com/mutate",
+      "failure_policy": "ignore",
+      "tls_config": {
+        "insecure_skip_verify": true
+      }
+    }
+  ]
+}
--- a/docs/examples/webhooks.yaml
+++ b/docs/examples/webhooks.yaml
@@ -1,0 +1,15 @@
+validating:
+  - name: policy-check
+    url: https://policy.example.com/validate
+    failure_policy: fail
+    timeout: 5s
+    tls_config:
+      ca_bundle_path: /etc/toolhive/pki/webhook-ca.crt
+
+mutating:
+  - name: request-enricher
+    url: https://enrichment.example.com/mutate
+    failure_policy: ignore
+    # Omitting timeout uses the default of 10s.
+    tls_config:
+      insecure_skip_verify: true
--- a/docs/middleware.md
+++ b/docs/middleware.md
@@ -22,6 +22,39 @@ The middleware chain consists of the following components:
 10. **Header Forward Middleware**: Injects custom headers into requests to remote MCP servers (optional)
 11. **Recovery Middleware**: Catches panics and returns HTTP 500 errors (always present)
 
+## Dynamic webhook middleware
+
+ToolHive supports dynamic webhook middleware for request mutation and validation. Webhooks are configured externally and loaded at runtime with `thv run --webhook-config <file>`.
+
+Two webhook types are supported:
+
+1. **Mutating webhooks**: Transform the parsed MCP request before later policy evaluation.
+2. **Validating webhooks**: Approve or deny the request after mutation has completed.
+
+When configured together, the effective order is:
+
+1. Authentication
+2. Token exchange and related auth middleware, when configured
+3. MCP parsing
+4. Mutating webhooks
+5. Validating webhooks
+6. Telemetry, authorization, and audit middleware
+
+Multiple webhook definitions of the same type run in configuration order. When multiple `--webhook-config` files are provided, later files override earlier webhook definitions with the same `name`.
+
+Configuration files may be written in YAML or JSON. Duration values such as `timeout` accept strings like `5s`, and omitted timeouts default to `10s`.
+
+Example:
+
+```bash
+thv run postgres-mcp --webhook-config docs/examples/webhooks.yaml
+```
+
+Example config files:
+
+- [`docs/examples/webhooks.yaml`](examples/webhooks.yaml)
+- [`docs/examples/webhooks.json`](examples/webhooks.json)
+
 ## Architecture Diagram
 
 ```mermaid


Sanskarzz:
@JAORMX  This PR is ready for review.

JAORMX:
Amazing! I'll check it out on Monday

pkg/webhook/types.go
```diff
@@ -96,8 +100,8 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("webhook failure_policy must be %q or %q, got %q",
 			FailurePolicyFail, FailurePolicyIgnore, c.FailurePolicy)
 	}
-	if c.Timeout < 0 {
-		return fmt.Errorf("webhook timeout must be non-negative")
+	if c.Timeout < MinTimeout {
```
JAORMX: blocker: This change will break any config file that omits the `timeout` field.

On `main`, `Timeout == 0` is a valid sentinel meaning "use the default (10s)". `NewClient` in `client.go:42-44` handles this. But now that `Validate()` rejects anything below `MinTimeout` (1s), the zero value from an omitted field fails validation.

So a config like this:
```yaml
validating:
  - name: policy-check
    url: https://policy.example.com/validate
    failure_policy: fail
```
...gets rejected. That's the most natural way to write a config (relying on the 10s default).

Both YAML (yaml.v3 leaves Duration at zero) and the new `UnmarshalJSON` (returns early with Timeout=0 when the field is absent) hit this.

A couple of options to fix:
1. Keep `Timeout == 0` as the "use default" sentinel and only check `MinTimeout` for non-zero values:
```go
if c.Timeout != 0 && c.Timeout < MinTimeout {
```
2. Apply `DefaultTimeout` during config loading when the field is missing/zero, before validation runs.

I'd lean towards option 2 since it makes the intent explicit and means `Validate()` always sees a real value.

pkg/webhook/types.go
```diff
@@ -110,6 +114,54 @@ func (c *Config) Validate() error {
...
+		return err
+	}
+
+	*c = Config{
+		Name:          raw.Name,
+		URL:           raw.URL,
+		FailurePolicy: raw.FailurePolicy,
+		TLSConfig:     raw.TLSConfig,
+		HMACSecretRef: raw.HMACSecretRef,
+	}
```
JAORMX, line 138: suggestion: This is the other half of the timeout zero-value problem. When the JSON field is missing or `null`, the function returns early with `Timeout` still at the Go zero value (0). That zero then hits the new `MinTimeout` check in `Validate()` and fails.

If you go with the "apply default before validation" approach, this is where you'd set it:

```go
if len(raw.Timeout) == 0 || string(raw.Timeout) == "null" {
    c.Timeout = DefaultTimeout
    return nil
}
```

You'd want the same treatment in the YAML path too... either in `LoadConfig` after unmarshalling or as a post-processing step in `loadAndMergeWebhookConfigs`. Basically, anywhere a zero-value timeout survives into a `Config` struct, it should be normalized to `DefaultTimeout` before `ValidateConfig` runs.

pkg/webhook/config.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// ValidateConfig validates all webhook configurations in a FileConfig,
+// collecting all validation errors before returning.
+func ValidateConfig(cfg *FileConfig) error {
+	if cfg == nil {
+		return nil
+	}
+
+	var errs []error
```
JAORMX, line 124: nitpick: The `wh := wh` capture is unnecessary here. `wh` is only used within the same loop iteration (no closure escaping, no goroutine). And since we're on Go 1.22+, loop variables are per-iteration scoped anyway. Same on line 113. Safe to remove both.

pkg/runner/config_builder.go
```diff
@@ -592,6 +611,19 @@ func WithMiddlewareFromFlags(
...
+		middlewareConfigs, err = addMutatingWebhookMiddleware(middlewareConfigs, b.config)
+		if err != nil {
+			return err
+		}
+
+		// Add Validating webhooks
+		middlewareConfigs, err = addValidatingWebhookMiddleware(middlewareConfigs, b.config)
+		if err != nil {
+			return err
+		}
```
JAORMX, line 630: suggestion: The placement here is correct (mutating before validating, both before telemetry/authz/audit). Nice.

One minor thing... in `PopulateMiddlewareConfigs` (the operator path), usage metrics comes *after* webhooks, but in this CLI path, `addCoreMiddlewares` bundles usage metrics *before* webhooks. It won't cause bugs since usage metrics doesn't depend on webhook state, but it's a subtle ordering divergence between the two paths. Worth a comment here noting the difference, or moving the webhook insertion to before `addCoreMiddlewares` to match the operator ordering more closely.

Given that we already have a TODO about unifying these two paths, a comment is probably fine for now.

pkg/webhook/config_test.go
```diff
@@ -0,0 +1,312 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
JAORMX, line 1: suggestion: The test coverage here is solid for the happy paths and basic error cases. A few scenarios that would really help (and would have caught the blocker):

1. **Omitted timeout in YAML** - a config with no `timeout` field. Right now this produces `Timeout=0` which fails the new `MinTimeout` validation.
2. **Omitted timeout in JSON** - same issue, different parser path.
3. **JSON with numeric nanosecond timeout** - the `timeoutNanos int64` branch in `UnmarshalJSON` is currently uncovered.
4. **JSON with `"timeout": null`** - the null branch.

Also, Codecov is reporting 72% patch coverage against the >80% target from the issue. The builder functions (`WithValidatingWebhooks`, `WithMutatingWebhooks`) and the `loadAndMergeWebhookConfigs` helper in `run_flags.go` don't have test coverage.

pkg/webhook/types.go
Sanskarzz: Fixed. I normalized omitted file-config timeouts to DefaultTimeout during config loading and kept `Timeout==0` valid in Validate() so existing sentinel behavior is preserved outside the file-loading path as well.

pkg/webhook/types.go
Sanskarzz: Fixed. UnmarshalJSON now defaults missing/null timeout to 10s, and the YAML path gets the same normalization via post-load defaulting.

pkg/webhook/config.go
Sanskarzz: Removed both captures.

pkg/runner/config_builder.go
Sanskarzz: Added a note in WithMiddlewareFromFlags documenting the usage-metrics ordering difference vs PopulateMiddlewareConfigs until the two paths are unified.

pkg/webhook/config_test.go
Sanskarzz: Added tests for omitted timeout in YAML/JSON, JSON null timeout, JSON numeric nanoseconds, the CLI loadAndMerge helper, and the webhook builder options.

Sanskarzz:
> * The issue mentions adding example configs to `docs/examples/` and updating `docs/middleware.md`. Those aren't in this PR. If you want to do those as a follow-up that's fine, but let's track it.

Included in this PR. I added example webhook config files under docs/examples/ and added a dynamic webhook middleware section to docs/middleware.md.

PR #4430, opened 2026-03-30
Default to :latest tag for OCI skill install references

JAORMX:
## Summary

- Running `thv skill install ghcr.io/org/skill` (without a tag) produced a confusing `Error: failed to install skill: Bad Gateway` because the raw user input was passed to oras-go which requires an explicit tag or digest, while go-containerregistry had already defaulted to `latest` internally but that parsed reference was ignored.
- Build the qualified OCI reference from the parsed `nameref.Reference`, appending `:latest` only when the user omitted a tag. Explicit tags and digests pass through unchanged.
- Change the OCI pull error status from 502 (Bad Gateway) to 400 (Bad Request) since a pull failure due to a bad reference is a client input error, not an upstream gateway failure.

## Type of change

- [x] Bug fix

## Test plan

`thv skill install ghcr.io/org/skill` now defaults to the `:latest` tag instead of failing with "Bad Gateway". When OCI pull errors do occur, the error message is now surfaced to the user instead of being hidden behind a generic status text.

### DIFF
--- a/pkg/skills/skillsvc/skillsvc.go
+++ b/pkg/skills/skillsvc/skillsvc.go
@@ -509,6 +509,22 @@ const ociPullTimeout = 5 * time.Minute
 // malicious artifact from causing OOM before the decompression limits kick in.
 const maxCompressedLayerSize int64 = 50 * 1024 * 1024 // 50 MB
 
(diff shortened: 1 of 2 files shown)


JAORMX:
I actually tried using the String method and it seems the docs were off. They did kot contain the tag.

PR #4434, opened 2026-03-30
Track runtime ownership to prevent cross-runtime status corruption

### LINKED ISSUE #4432: Single-runtime assumption corrupts status files of workloads managed by other runtimes
## Summary

When multiple container runtimes coexist (e.g., podman and go-microvm), running any CLI command (`thv list`, `thv status`, `thv stop`, `thv restart`) with one runtime active **permanently corrupts the status files** of workloads managed by the other runtime, marking them as `unhealthy`.

This happens because the architecture assumes a single runtime owns all workloads. There is no concept of runtime ownership -- status files, labels, RunConfig, and `core.Workload` carry no information about which runtime created a workload.

## Reproduction

1. Start MCP servers using the default podman runtime:
   ```bash
   thv run some-mcp-server
   thv list  # shows running
   ```
2. Switch to the go-microvm runtime:
   ```bash
   export TOOLHIVE_RUNTIME=go-microvm
   thv list
   ```
3. Switch back to podman:
   ```bash
   unset TOOLHIVE_RUNTIME
   thv list  # podman servers now show as "unhealthy"
   ```

The podman containers are still running fine, but their status files have been overwritten.

## Root Cause

The bug lives in the file-based status manager's reconciliation logic during `ListWorkloads` and `GetWorkload`.

### The chain of events

1. **`Factory.Create()` creates a single runtime** (`pkg/container/factory.go:109-110`). If `TOOLHIVE_RUNTIME=go-microvm`, only the go-microvm client is initialized.

2. **`fileStatusManager.ListWorkloads()` reads status files for ALL workloads** (`pkg/workloads/statuses/file_status.go:288-289`) regardless of which runtime created them.

3. **It queries only the active runtime** (`file_status.go:283`): `f.runtime.ListWorkloads(ctx)` -- the go-microvm runtime returns zero podman containers.

4. **`mergeRuntimeAndFileWorkloads()` calls `handleRuntimeMissing()`** (`file_status.go:1128-1135`) for every workload that exists in status files but not in the active runtime's container list.

5. **`handleRuntimeMissing()` writes `WorkloadStatusUnhealthy` to the status file on disk** (`file_status.go:967-974`):
   ```go
   if fileWorkload.Status == rt.WorkloadStatusRunning || fileWorkload.Status == rt.WorkloadStatusStopped {
       contextMsg := fmt.Sprintf("workload %s not found in runtime, marking as unhealthy", workloadName)
       if err := f.SetWorkloadStatus(ctx, workloadName, rt.WorkloadStatusUnhealthy, contextMsg); err != nil {
           return core.Workload{}, err
       }
       fileWorkload.Status = rt.WorkloadStatusUnhealthy
   }
   ```

### Why runtime ownership is not tracked

None of the data structures carry runtime type information:

| Data Structure | Runtime field? | Notes |
|---|---|---|
| `workloadStatusFile` (`file_status.go:186-192`) | No | Only status, context, timestamps, PID |
| `core.Workload` (`pkg/core/workload.go:15-52`) | No | Only a `Remote` bool for remote vs container |
| `RunConfig` (`pkg/runner/config.go`) | No | `Deployer` field has `json:"-"` -- not persisted |
| Container labels (`pkg/labels/labels.go:46-53`) | No | No runtime-type label |
| `runtime.ContainerInfo` (`pkg/container/runtime/types.go:49-69`) | No | No runtime identifier |

## Affected Commands

All commands that call `ListWorkloads` or `GetWorkload` have write-on-read side effects:

| Command | Entry point | Side effect |
|---|---|---|
| `thv list` | `file_status.go:275` via `mergeRuntimeAndFileWorkloads` | Marks all workloads not in active runtime as UNHEALTHY |
| `thv status <name>` | `file_status.go:195` via `validateRunningWorkload` | Marks specific workload as UNHEALTHY if not in active runtime |
| `thv stop --all` / `--group` | Lists workloads first | Marks non-active-runtime workloads UNHEALTHY before stopping |
| `thv restart --all` / `--group` | Lists workloads first | Same as stop |

## What is NOT affected

- **go-microvm `recoverState()` does NOT harm podman processes.** The go-microvm state directory (`~/.config/toolhive/gomicrovm/vms/`) is completely separate from podman state, and `cleanupOrphanedRunner()` only reads PIDs from go-microvm-specific state files.
- **`thv serve` does not do periodic reconciliation** that would trigger this -- the damage only happens on explicit CLI commands.
- **Marking unhealthy does not kill containers or proxy processes** -- it only corrupts the status file. The actual workloads keep running.

## Possible Fix Directions

1. **Track runtime ownership**: Add a runtime-type field to status files (and/or container labels / RunConfig). During reconciliation, skip workloads that belong to a different runtime.

2. **Read-only list path**: Separate the read-only listing from the write-on-read validation. `thv list` should not mutate state as a side effect.

3. **Multi-runtime aware reconciliation**: Query all available runtimes during reconciliation, not just the active one. Only mark a workload unhealthy if it's missing from the runtime it was created with.

JAORMX:
## Summary

When multiple runtimes coexist (e.g., podman and an experimental runtime like go-microvm), running any CLI command (`thv list`, `thv status`, `thv stop`, `thv restart`) with one runtime active permanently corrupts the status files of workloads managed by the other runtime, marking them as "unhealthy." The root cause is that no data structure tracks which runtime owns a workload, and the reconciliation logic in `fileStatusManager` assumes all workloads belong to the single active runtime.

This PR adds runtime ownership tracking so that reconciliation skips workloads belonging to a different runtime:

- Add `Name() string` to the `Runtime` interface so the active runtime can identify itself
- Add `RuntimeName` field to `RunConfig` (persisted to disk) to record which runtime created a workload
- Guard `handleRuntimeMissing()` and `validateRunningWorkload()` in the file status manager to skip workloads owned by a different runtime
- Stamp `RuntimeName` at all workload creation entry points (CLI, API, MCP server)
- Opportunistically migrate legacy workloads (those created before this change) by stamping them with the active runtime's name when reconciliation confirms they are healthy

Fixes #4432

## Type of change

- [x] Bug fix

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/container/runtime/types.go` | Add `Name() string` to `Runtime` interface |
| `pkg/container/docker/client.go` | Implement `Name()` returning `"docker"` |
| `pkg/container/kubernetes/client.go` | Implement `Name()` returning `"kubernetes"` |
| `pkg/runner/config.go` | Add `RuntimeName` field to `RunConfig` |
| `cmd/thv/app/run.go` | Set `RuntimeName` at both CLI save sites |
| `pkg/api/v1/workload_service.go` | Set `RuntimeName` in API create path |
| `pkg/mcp/server/run_server.go` | Set `RuntimeName` in MCP server path |
| `pkg/workloads/statuses/file_status.go` | Add `isOwnedByActiveRuntime` helper; guard `validateRunningWorkload` and `handleRuntimeMissing`; add `migrateRuntimeName` for legacy workloads |

Workloads managed by one runtime are no longer incorrectly marked as "unhealthy" when a different runtime is active. Existing workloads without the new `runtime_name` field get migrated automatically... the first time reconciliation confirms they are healthy, they get stamped with the active runtime's name.

- **Backwards compatibility**: `RuntimeName` uses `omitempty`, so old RunConfigs deserialize with `""`. The `isOwnedByActiveRuntime` helper treats empty as "assume ours" to preserve existing health-check behavior for single-runtime users. The `migrateRuntimeName` function then progressively stamps legacy workloads, so the empty-string fallback is a transient state rather than permanent.
- **Migration safety**: `migrateRuntimeName` only stamps a workload after the runtime confirms it is actually healthy (container running, proxy OK). We never blindly claim ownership of a workload we haven't verified.
- **go-microvm runtime** is not yet on main, so it does not implement `Name()` here. It will need to add that method when it merges.
- There is an inefficiency where `GetReader` is called multiple times per workload during reconciliation (`isRemoteWorkload`, `isOwnedByActiveRuntime`, `migrateRuntimeName`). Tracked in #4456.

### DIFF
--- a/pkg/runner/config.go
+++ b/pkg/runner/config.go
@@ -55,6 +55,11 @@ type RunConfig struct {
 	// means unversioned (backward-compat with older operators, or non-operator callers).
 	MCPServerGeneration int64 `json:"mcpserver_generation,omitempty" yaml:"mcpserver_generation,omitempty"`
 
+	// RuntimeName is the registered name of the container runtime that owns this
+	// workload (e.g., "docker", "kubernetes"). Used during reconciliation to avoid
+	// corrupting status files of workloads managed by a different runtime.
+	RuntimeName string `json:"runtime_name,omitempty" yaml:"runtime_name,omitempty"`
+
 	// Image is the Docker image to run
 	Image string `json:"image" yaml:"image"`
 
--- a/pkg/workloads/statuses/file_status.go
+++ b/pkg/workloads/statuses/file_status.go
@@ -74,6 +74,38 @@ type fileStatusManager struct {
 	runConfigStore state.Store
 }
 
+// isOwnedByActiveRuntime checks whether the named workload was created by the
+// currently active runtime. It loads the runtime_name field from the persisted
+// RunConfig and compares it against f.runtime.Name(). Legacy workloads that
+// predate the runtime_name field (empty string) are conservatively treated as
+// owned by the active runtime so that existing validation behaviour is preserved.
+func (f *fileStatusManager) isOwnedByActiveRuntime(ctx context.Context, workloadName string) bool {
+	reader, err := f.runConfigStore.GetReader(ctx, workloadName)
+	if err != nil {
+		// RunConfig missing or unreadable -- assume ours so we don't silently
+		// skip validation for workloads we should be checking.
+		return true
+	}
+	defer func() {
+		if err := reader.Close(); err != nil {
+			slog.Warn("failed to close reader", "error", err)
+		}
+	}()
+
+	var config struct {
+		RuntimeName string `json:"runtime_name"`
+	}
+	if err := json.NewDecoder(reader).Decode(&config); err != nil {
+		return true // can't determine ownership -- assume ours
+	}
+
+	// Empty means legacy workload; treat as owned by active runtime.
+	if config.RuntimeName == "" {
+		return true
+	}
+	return config.RuntimeName == f.runtime.Name()
+}
+
 // isRemoteWorkload checks if a workload is remote by attempting to load its run configuration
 // and checking if it has a RemoteURL field set.
 // TODO: This is a temporary solution to check if a workload is remote
@@ -910,6 +942,12 @@ func (f *fileStatusManager) validateRunningWorkload(
 		return result, nil
 	}
 
+	// Skip validation for workloads owned by a different runtime to avoid
+	// corrupting their status files (see #4432).
+	if !f.isOwnedByActiveRuntime(ctx, workloadName) {
+		return result, nil
+	}
+
 	// Get raw container info from runtime (before label filtering)
 	containerInfo, err := f.runtime.GetWorkloadInfo(ctx, workloadName)
 	if err != nil {
@@ -961,6 +999,12 @@ func (f *fileStatusManager) handleRuntimeMissing(
 	if fileWorkload.Remote {
 		// Remote workloads don't exist in the container runtime, so it's normal for them to be missing
 		// Don't mark them as unhealthy
+		return fileWorkload, nil
+	}
+
+	// Skip reconciliation for workloads owned by a different runtime to avoid
+	// corrupting their status files (see #4432).
+	if !f.isOwnedByActiveRuntime(ctx, workloadName) {
 		return fileWorkload, nil
 	}
 
--- a/pkg/workloads/statuses/file_status.go
+++ b/pkg/workloads/statuses/file_status.go
@@ -106,6 +106,61 @@ func (f *fileStatusManager) isOwnedByActiveRuntime(ctx context.Context, workload
 	return config.RuntimeName == f.runtime.Name()
 }
 
+// migrateRuntimeName stamps a legacy workload (RuntimeName == "") with the
+// active runtime's name. This is called only after the runtime has confirmed the
+// workload is healthy, so we know with certainty it belongs to the active
+// runtime. Save failures are non-fatal — the migration will be retried on the
+// next reconciliation cycle.
+func (f *fileStatusManager) migrateRuntimeName(ctx context.Context, workloadName string) {
+	reader, err := f.runConfigStore.GetReader(ctx, workloadName)
+	if err != nil {
+		slog.Debug("skipping runtime name migration: cannot read config", "workload", workloadName, "error", err)
+		return
+	}
+
+	var raw map[string]json.RawMessage
+	decodeErr := json.NewDecoder(reader).Decode(&raw)
+	// Always close the reader before acting on the decode result.
+	if closeErr := reader.Close(); closeErr != nil {
+		slog.Warn("failed to close reader for runtime name migration", "workload", workloadName, "error", closeErr)
+	}
+	if decodeErr != nil {
+		return
+	}
+
+	// Check if runtime_name is already set
+	if rn, ok := raw["runtime_name"]; ok {
+		var name string
+		if err := json.Unmarshal(rn, &name); err == nil && name != "" {
+			return // Already migrated
+		}
+	}
+
+	// Stamp with the active runtime name
+	runtimeBytes, err := json.Marshal(f.runtime.Name())
+	if err != nil {
+		return
+	}
+	raw["runtime_name"] = runtimeBytes
+
+	writer, err := f.runConfigStore.GetWriter(ctx, workloadName)
+	if err != nil {
+		slog.Warn("failed to open writer for runtime name migration", "workload", workloadName, "error", err)
+		return
+	}
+	defer func() {
+		if err := writer.Close(); err != nil {
+			slog.Warn("failed to close writer for runtime name migration", "workload", workloadName, "error", err)
+		}
+	}()
+
+	encoder := json.NewEncoder(writer)
+	encoder.SetIndent("", "  ")
+	if err := encoder.Encode(raw); err != nil {
+		slog.Warn("failed to write migrated RunConfig", "workload", workloadName, "error", err)
+	}
+}
+
 // isRemoteWorkload checks if a workload is remote by attempting to load its run configuration
 // and checking if it has a RemoteURL field set.
 // TODO: This is a temporary solution to check if a workload is remote
@@ -964,7 +1019,12 @@ func (f *fileStatusManager) validateRunningWorkload(
 		return unhealthyWorkload, nil
 	}
 
-	// Runtime and proxy confirm workload is healthy - merge runtime data with file status
+	// Runtime and proxy confirm workload is healthy — opportunistically migrate
+	// legacy workloads that predate the runtime_name field so they are stamped
+	// with the owning runtime for future reconciliation cycles.
+	f.migrateRuntimeName(ctx, workloadName)
+
+	// Merge runtime data with file status
 	return f.mergeHealthyWorkloadData(containerInfo, result)
 }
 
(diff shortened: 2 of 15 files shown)


pkg/workloads/statuses/file_status.go
```diff
@@ -74,6 +74,38 @@ type fileStatusManager struct {
...
+
+	var config struct {
+		RuntimeName string `json:"runtime_name"`
+	}
+	if err := json.NewDecoder(reader).Decode(&config); err != nil {
+		return true // can't determine ownership -- assume ours
+	}
+
+	// Empty means legacy workload; treat as owned by active runtime.
+	if config.RuntimeName == "" {
```
amirejaz: **Correctness concern: the legacy-workload default preserves the exact bug this PR is fixing.**

Any workload created before this PR has `RuntimeName = ""`. The `empty → assume ours` default means: if a user has a go-microvm workload (no `RuntimeName`) and switches to Docker, `isOwnedByActiveRuntime` returns `true`, `handleRuntimeMissing` is reached, Docker doesn't know the workload, and it gets marked unhealthy — exactly the bug in #4432.

The PR description calls this "preserving current behavior until they are restarted", but the current behavior IS the bug. The conservative default is only safe for single-runtime users, which is precisely not the scenario this PR targets.

A safer default for legacy workloads would be to return `false` (skip reconciliation) rather than `true` (assume ours). The trade-off: a genuinely broken legacy workload whose container is missing would no longer be auto-marked unhealthy until it is restarted with a newer ToolHive version. That seems like an acceptable temporary regression compared to continuing to corrupt cross-runtime status files for all existing workloads.

pkg/workloads/statuses/file_status.go
```diff
@@ -964,6 +1002,12 @@ func (f *fileStatusManager) handleRuntimeMissing(
 		return fileWorkload, nil
 	}
 
+	// Skip reconciliation for workloads owned by a different runtime to avoid
+	// corrupting their status files (see #4432).
+	if !f.isOwnedByActiveRuntime(ctx, workloadName) {
```
amirejaz, line 1085: **Nit: the `fileWorkload.Remote` guard above (line 999) makes this `isOwnedByActiveRuntime` check unreachable for remote workloads — they already exit before reaching here.**

This is correct behavior (remote workloads belong to no runtime), but the ordering makes the intent implicit. A short comment on the Remote guard like `// Remote workloads bypass runtime-ownership checks — they have no owning runtime` would make it clear this is intentional and not a gap in the cross-runtime protection.

pkg/workloads/statuses/file_status.go
```diff
@@ -74,6 +74,38 @@ type fileStatusManager struct {
 	runConfigStore state.Store
 }
 
+// isOwnedByActiveRuntime checks whether the named workload was created by the
+// currently active runtime. It loads the runtime_name field from the persisted
+// RunConfig and compares it against f.runtime.Name(). Legacy workloads that
+// predate the runtime_name field (empty string) are conservatively treated as
+// owned by the active runtime so that existing validation behaviour is preserved.
+func (f *fileStatusManager) isOwnedByActiveRuntime(ctx context.Context, workloadName string) bool {
```
amirejaz, line 89: **Follow-up: double `GetReader` per workload during list.**

For each workload, `GetReader` is called once in `isRemoteWorkload` and again here in `isOwnedByActiveRuntime`. Both involve file I/O and happen sequentially per workload. For users with many workloads this adds up. Already noted in the PR special notes — worth a tracked follow-up issue so it doesn't get lost.

pkg/runner/config.go
```diff
@@ -48,6 +48,11 @@ type RunConfig struct {
 	// SchemaVersion is the version of the RunConfig schema
 	SchemaVersion string `json:"schema_version" yaml:"schema_version"`
 
+	// RuntimeName is the registered name of the container runtime that owns this
+	// workload (e.g., "docker", "kubernetes"). Used during reconciliation to avoid
+	// corrupting status files of workloads managed by a different runtime.
+	RuntimeName string `json:"runtime_name,omitempty" yaml:"runtime_name,omitempty"`
```
amirejaz, line 61: **Note: existing workloads are never auto-migrated.**

`RuntimeName` is stamped only at creation time (`SaveState`). Existing workloads on disk have `RuntimeName = ""` and will never be updated unless the workload is deleted and recreated. This is fine operationally, but means the cross-runtime protection only applies to workloads created after this PR is deployed — which ties back to the correctness concern in `isOwnedByActiveRuntime`.

amirejaz (changes requested):
## Review: Track runtime ownership to prevent cross-runtime status corruption

The approach is correct and well-structured — `RuntimeName` in `RunConfig` is the right place to record ownership, the `isOwnedByActiveRuntime` guard is clean, and the backwards-compatibility intent is understood. The comments below are a mix of one correctness concern and follow-up UX issues that the new `RuntimeName` field now makes tractable.

**Overall: the correctness concern in `pkg/workloads/statuses/file_status.go` (`isOwnedByActiveRuntime`, line 103) should be addressed before merging; the rest are follow-up issues.**

### Follow-up issues worth tracking

**Stale status in `thv list` for cross-runtime workloads**
When Docker is active and a go-microvm workload is in file state as `running`, `handleRuntimeMissing` now correctly skips reconciliation — but the workload still appears as `running` in `thv list` with no indication it belongs to a different runtime. `RuntimeName` is now persisted and could be surfaced as a `RUNTIME` column in `thv list`.

**`thv stop`/`restart` on a cross-runtime workload silently succeeds**
`stopContainerWorkload` in `pkg/workloads/manager.go` treats `ErrWorkloadNotFound` as a non-error — logs a warning, returns `nil`, and never updates the status file. So the workload stays `running` from the user's perspective with no error surfaced. Combined with the stale-status issue above, a user can: (1) see `foo` as `running`, (2) run `thv stop foo` — no error, (3) see `foo` still as `running`. Now that `RuntimeName` is available in `RunConfig`, `stopSingleWorkload` could detect the mismatch early and return a clear error like: _"workload 'foo' was created with the 'go-microvm' runtime; switch to that runtime to manage it."_

pkg/workloads/statuses/file_status.go
JAORMX: Good catch, and I agree this is the most important thing to get right here.

So, the concern about flipping to `false` is that it breaks health checks for the 100% of current users who are on a single runtime. Nobody is using the experimental go-microvm runtime yet... it's not even on main. Defaulting to "skip reconciliation" means a genuinely broken single-runtime workload just silently stays marked as running when it shouldn't be. That's a real regression for everyone today vs. a theoretical fix for a scenario that doesn't exist yet.

That said, I did add a migration in the follow-up commit that addresses the root of your concern. `migrateRuntimeName` now stamps legacy workloads with the active runtime's name, but only after reconciliation confirms the workload is actually healthy (container running, proxy OK). So the empty-string case gets progressively eliminated as workloads are validated. By the time anyone actually uses multi-runtime, their workloads will already be stamped.

pkg/workloads/statuses/file_status.go
JAORMX: Done! Added a comment that explains both why the early return exists and that it intentionally bypasses the runtime-ownership check below.

pkg/workloads/statuses/file_status.go
JAORMX: Filed #4456 to track this. Note that with the migration commit it's actually 3 reads per legacy workload now (`isRemoteWorkload` + `isOwnedByActiveRuntime` + `migrateRuntimeName`), though that drops to 2 once migration is done. Worth cleaning up properly.

pkg/runner/config.go
JAORMX: Addressed this in the follow-up commit. Legacy workloads now get their `RuntimeName` stamped opportunistically during reconciliation, but only when the runtime confirms the workload is healthy. So we're not blindly claiming ownership... we only stamp when we have strong evidence the workload actually belongs to the active runtime. The empty-string fallback in `isOwnedByActiveRuntime` stays as a safety net, but it should be a transient state now rather than permanent.

JAORMX:
Closing this for now — deprioritizing this work. Thanks @amirejaz for the thorough review; the runtime-ownership fix and the follow-ups we identified (tracked in #4456) are still valid and can be picked back up later if this becomes a priority again.

PR #4447, opened 2026-03-30
Refactor OAuth token persistence and fix Resource/Audience conflation

aponcedeleonch:
## Summary

- **Extract `TokenPersistenceManager`**: Three callers shared an identical pattern — nil-check a secrets provider, fetch a cached refresh token, create an `oauth2.TokenSource` from it. Extract this into `pkg/auth/remote.TokenPersistenceManager` to eliminate the duplication. `FetchRefreshToken` is provided separately from `RestoreFromCache` so callers can do the cheap secret lookup before paying the cost of OIDC network discovery.
- **Generalize `RegistryOAuthConfig` → `OAuthConfig`**: Rename and move to `pkg/config` with a new `Resource` field (RFC 8707) and an injectable `configUpdater` callback, decoupling the token-source from `pkg/config` internals so enterprise callers can supply their own persistence logic.
- **Fix Resource vs Audience conflation**: `Audience` (a provider-specific authorization URL parameter, e.g. Auth0) was being passed where `Resource` (RFC 8707 resource indicator, sent to the token endpoint) was expected. `Resource` now flows to `CreateOAuthConfigFromOIDC` and `Audience` is routed into `OAuthParams["audience"]`.

## Type of change

- [x] Refactoring (no functional changes, no API changes)

## Large PR Justification

Of the 990 lines changed, **813 are test code** (82%) and only **177 are production code** across 7 files. The production changes cannot be split because:

- **Atomic refactoring**: The `RegistryOAuthConfig` → `OAuthConfig` rename touches every caller site simultaneously — leaving half the callers on the old type would not compile. The new `Resource` field, `configUpdater` callback, and `TokenPersistenceManager` extraction are consumed by the same call chains, so they must land together.
- **Bug fix requires the refactoring**: The Resource/Audience conflation fix depends on the new `OAuthConfig.Resource` field introduced by the rename. Shipping the fix without the generalization would require a throwaway intermediate step.
- **Test-only bulk**: The 813 lines of new tests are regression guards for the bug fix and previously-untested code paths (`buildOAuthFlowConfig`, `wrapWithPersistence`, `resolveClientCredentials`, `configUpdater` callback). Stripping them into a follow-up PR would leave the bug fix unguarded during review.

## Test plan

- [x] New unit tests added for `TokenPersistenceManager.FetchRefreshToken` (all paths)
- [x] New unit tests added for `buildOAuthFlowConfig` Resource-vs-Audience split — critical regression guard ensuring `Audience` lands in `OAuthParams["audience"]` and `Resource` does NOT appear there
- [x] New unit tests added for injectable `configUpdater` callback invocation
- [x] New unit tests added for `handler.buildOAuthFlowConfig` endpoint-override logic
- [x] New unit tests added for `handler.wrapWithPersistence` persistence callbacks
- [x] New unit tests added for `handler.resolveClientCredentials` priority logic
- [x] All existing tests pass: `go test ./pkg/auth/remote/... ./pkg/registry/auth/... ./pkg/config/...`

No. This is an internal refactoring and bug fix in the OAuth token persistence layer. The Resource/Audience fix corrects incorrect token requests for RFC 8707-compliant servers and Auth0-style providers, but the CLI interface and configuration schema are unchanged.

### DIFF
--- a/pkg/auth/remote/doc.go
+++ b/pkg/auth/remote/doc.go
@@ -1,18 +1,28 @@
 // SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-// Package remote provides authentication handling for remote MCP servers.
+// Package remote provides authentication handling for remote MCP servers,
+// as well as general-purpose OAuth token source utilities used across the codebase.
 //
-// This package implements OAuth/OIDC-based authentication with automatic
-// discovery support for remote MCP servers. It handles:
+// # Remote MCP server authentication
+//
+// Handler.Authenticate() is the main entry point: it takes a remote URL
+// and performs all necessary discovery and authentication steps, including:
 //   - OAuth issuer discovery (RFC 8414)
 //   - Protected resource metadata (RFC 9728)
 //   - OAuth flow execution (PKCE-based)
 //   - Token source creation for HTTP transports
 //
-// The main entry point is Handler.Authenticate() which takes a remote URL
-// and performs all necessary discovery and authentication steps.
-//
 // Configuration is defined in pkg/runner.RemoteAuthConfig as part of the
 // runner's RunConfig structure.
+//
+// # General-purpose token source utilities
+//
+// These types and functions are also used outside of remote MCP auth (e.g. registry auth):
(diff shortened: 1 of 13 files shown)


aponcedeleonch:
closing since this PR is old and the refactoring done by this PR was already taken care of on a separate PR

PR #4454, opened 2026-03-31
Follow HTTP redirects in transparent proxy

### LINKED ISSUE #4453: Transparent proxy should follow HTTP redirects for remote MCP servers
## Bug description

When a remote MCP server returns an HTTP redirect (301, 302, 307, or 308), the transparent proxy passes the redirect response verbatim to the MCP client. MCP clients expect JSON-RPC responses and cannot handle redirects, so the server appears "running" in `thv list` but returns no tools. MCP calls receive an HTML redirect page instead of JSON-RPC, with no error indication from ToolHive.

## Steps to reproduce

1. Set up any HTTP server that returns a 3xx redirect (e.g., nginx returning `308 Permanent Redirect` from `http://` to `https://` in front of an MCP server)
2. Register it in ToolHive with the pre-redirect URL:
   ```
   thv run http://example.com/mcp --name test-server --transport streamable-http
   ```
3. Attempt to call the server:
   ```
   curl -X POST http://127.0.0.1:<proxy-port>/mcp \
     -H "Content-Type: application/json" \
     -H "Accept: application/json, text/event-stream" \
     -d '{"jsonrpc":"2.0","method":"tools/list","id":1}'
   ```

## Expected behavior

The proxy follows the redirect and returns the MCP server's JSON-RPC response to the client.

## Actual behavior

The proxy returns the raw 3xx redirect response (with an HTML body) to the client. The MCP call fails with no error logged by ToolHive.

## Additional context

The transparent proxy uses `httputil.ReverseProxy` with a custom `tracingTransport` as the `http.RoundTripper`. Since `ReverseProxy` uses `RoundTripper` (not `http.Client`), Go's automatic redirect-following logic never runs. After forwarding a request, `tracingTransport.RoundTrip` inspects the response for specific status codes — 401 (unauthorized callback), 2xx/404 (DELETE session cleanup), and 200 (session tracking) — but 3xx responses pass through unmodified.

gkatz2:
## Summary

- MCP clients cannot handle HTTP redirect responses (they expect JSON-RPC), so when a remote server redirects, the client silently fails. The transparent proxy now follows redirects in its `forward` method before returning the response to the client.
- Adds an `isRedirectStatus` helper that handles 301, 302, 307, and 308. The HTTP method and body are preserved across redirects via `req.Clone()`. A `maxRedirects` limit of 10 prevents infinite loops.
- Logs at WARN level on each redirect to nudge operators toward fixing the server URL.

Fixes #4453

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Tested with a local HTTP server returning 308 redirects to a running MCP server (design-system-mcp). Verified:
- `initialize` call succeeds through the redirect
- Proxy logs show the WARN message with from/to URLs and redirect count

Remote MCP servers behind HTTP redirects now work. Previously, the proxy passed redirect responses to the client, causing silent failures. No configuration needed — redirect following is automatic.

- The redirect logic lives in `forward` rather than `RoundTrip` so that session tracking, 401 handling, and DELETE cleanup in `RoundTrip` only see the final response after all redirects are resolved.
- 303 (See Other) is excluded from `isRedirectStatus` because it explicitly requires changing the method to GET, which would break JSON-RPC.

### DIFF
--- a/pkg/transport/proxy/transparent/transparent_proxy.go
+++ b/pkg/transport/proxy/transparent/transparent_proxy.go
@@ -736,7 +736,7 @@ func (r *backendRecovery) podBackendURL(capturedAddr string) string {
 	if err != nil {
 		return r.targetURI
 	}
-	if parsed.Scheme == "https" {
+	if parsed.Scheme == "https" { //nolint:goconst // protocol name, not a magic string
 		return r.targetURI
 	}
 	parsed.Host = capturedAddr
@@ -788,6 +788,7 @@ func followRedirects(
 		}
 
 		// Block HTTPS-to-HTTP downgrades to prevent silent loss of transport security.
+		//nolint:goconst // "https" is a protocol name, not a magic string worth extracting
 		if req.URL.Scheme == "https" && redirectURL.Scheme == "http" {
 			slog.Warn("refusing redirect that downgrades from HTTPS to HTTP",
 				"from", req.URL.String(), "to", redirectURL.String())


pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -377,12 +383,87 @@ func (p *TransparentProxy) serverInitialized() bool {
...
+	// changing the method would break every request.
+	//
+	// Note: request headers (including Authorization) are preserved across
+	// redirects because the proxy controls the target URL and redirects
+	// are expected to stay within the same trust domain.
+	for redirectsFollowed := 0; redirectsFollowed < maxRedirects &&
+		isRedirectStatus(resp.StatusCode); redirectsFollowed++ {
+		location := resp.Header.Get("Location")
+		if location == "" {
+			break
```
amirejaz: **SSRF via open redirect** — `req.URL.Parse(location)` accepts any absolute URL, including `http://kubernetes.default.svc/...` or `http://169.254.169.254/`. A compromised MCP server can redirect the proxy to any internal cluster service, and the proxy (with its typically permissive network policy) will dutifully forward the request.

Consider rejecting cross-host redirects:
```suggestion
		redirectURL, parseErr := req.URL.Parse(location)
		if parseErr != nil {
			slog.Warn("failed to parse redirect Location header",
				"location", location, "error", parseErr)
			break
		}
		// Only follow same-host redirects to prevent SSRF. A redirect to a
		// different host indicates the operator should update the target URL.
		if redirectURL.Host != req.URL.Host {
			slog.Warn("refusing cross-host redirect from remote MCP server; update the configured target URL",
				"from_host", req.URL.Host, "to_host", redirectURL.Host)
			break
		}
```

Note: the `if parseErr != nil` block that follows this line in the original is being absorbed into the suggestion above — remove the duplicate below.

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -377,12 +383,87 @@ func (p *TransparentProxy) serverInitialized() bool {
...
+			"from", req.URL.String(),
+			"to", redirectURL.String(),
+			"redirect_number", redirectsFollowed+1)
+
+		// Drain and close the redirect response body to release the
+		// underlying connection back to the transport's connection pool.
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+
+		// Clone preserves Method and all headers. We intentionally do not
```
amirejaz: **Authorization header forwarded to a different host** — `req.Clone()` copies all headers verbatim, including `Authorization`. If a cross-host redirect is ever permitted, the bearer token is delivered to the new host. The Go stdlib explicitly strips `Authorization` on cross-host redirects for this reason (`net/http/client.go:shouldCopyHeaderOnRedirect`).

If the same-host check suggested above is added, this becomes a defense-in-depth note; if cross-host redirects are intentionally supported in the future, add:
```go
if redirectURL.Host != originalHost {
    req.Header.Del("Authorization")
    req.Header.Del("Cookie")
    req.Header.Del("Proxy-Authorization")
}
```

pkg/transport/proxy/transparent/redirect_test.go
```diff
@@ -0,0 +1,280 @@
...
+				w.WriteHeader(tt.redirectStatus)
+			}))
+			defer redirector.Close()
+
+			p := NewTransparentProxy("127.0.0.1", 0, "", nil, nil, nil, false, false, "streamable-http", nil, nil, "", false)
+
+			targetURL, err := url.Parse(redirector.URL)
+			require.NoError(t, err)
+			proxy := createBasicProxy(p, targetURL)
+
```
amirejaz, line 102: No test covers a cross-host redirect scenario. It would be valuable to have a test that (a) sets up two separate servers, (b) has server A redirect to server B at a different host, and (c) asserts the expected behavior — either that the 3xx is returned as-is (if the same-host guard is added) or that the `Authorization` header is stripped (if cross-host is permitted). This makes the security decision explicit and prevents future regressions.

gkatz2:
Thanks for the detailed review! Both security issues have been addressed in the follow-up commit [3e034e44](https://github.com/stacklok/toolhive/commit/3e034e4452a91a18ca8d039ff2e12e4267c1c06c).

**SSRF / cross-host redirect (both issues addressed together):** The `forward()` loop now extracts the original host before following redirects and compares each redirect destination's host against it. If the redirect target is a different host, we break and return the 3xx response as-is — the client decides what to do. This prevents both SSRF and auth header leakage to a different host in one check.

**E2E failure:** The failing test (`should successfully delete running workload`) is in `api_workloads_test.go` and exercises the workload deletion API — it has no interaction with the redirect-following code path. The failure appears to be a timing issue where Docker took longer than 60s to clean up the container on the CI runner. Happy to re-trigger if that would help.

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -377,12 +383,99 @@ func (p *TransparentProxy) serverInitialized() bool {
...
+				"location", location, "error", parseErr)
+			break
+		}
+
+		// Block cross-host redirects to prevent SSRF and credential leakage.
+		if redirectURL.Host != originalHost {
+			slog.Warn("refusing cross-host redirect from remote MCP server; update the configured target URL",
+				"from_host", originalHost, "to_host", redirectURL.Host)
+			break
+		}
```
jhrozek: We can do this as a follow up but I wonder if we should also think about checking if the redirectURL is downgrading the transport scheme from https to http:
```go
if req.URL.Scheme == "https" && redirectURL.Scheme == "http" {
   break;
}
```

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -377,12 +383,99 @@ func (p *TransparentProxy) serverInitialized() bool {
...
+
+		slog.Warn("following HTTP redirect from remote MCP server; consider updating the server URL",
```
jhrozek: nit, but should this be Warn or Info? Since the message is fired on every redirect

pkg/transport/proxy/transparent/transparent_proxy.go
gkatz2: Added in the latest push — checks for HTTPS-to-HTTP downgrade and breaks with a warning.

pkg/transport/proxy/transparent/transparent_proxy.go
gkatz2: Changed to slog.Info.

jhrozek:
@gkatz2 I'm sorry there's another conflict (I think this time with your own PR to add the timeout). If you rebase I'll re-ack right away.

gkatz2:
@jhrozek, I resolved the conflict. Please take another look.

PR #4464, opened 2026-03-31
Migrate session manager to DataStorage and add RestoreSession

### LINKED ISSUE #4220: Write unit tests for horizontal scaling operator behaviors (THV-0047)

yrobla:
## Summary

Replace the `*transportsession.Manager` dependency in `sessionmanager.Manager` with the `DataStorage` interface (introduced in split PR 1), and simultaneously migrate `discovery.Middleware` from `*transportsession.Manager` to a `MultiSessionGetter` interface.

These two changes are intentionally landed together. Phase 2 (DataStorage migration) and phase 3 (middleware decoupling) cannot be split cleanly:

- Phase 2 alone is broken: sessions generated by the new `sessionmanager.Manager` are stored in `DataStorage` + `RestorableCache`, but never written into `*transportsession.Manager`. `discovery.Middleware` would call `sessionManager.Get(sessionID)` and find nothing, returning HTTP 401 for every subsequent request. A temporary bridge shim could paper over this, but it would only exist to be deleted in phase 3 — net negative for reviewers.
- Phase 3 alone is impossible: `MultiSessionGetter.GetMultiSession()` is backed by `vmcpSessionMgr`, which only exists after the phase 2 rewrite.
- The 401→404 behaviour change is a direct consequence of removing the stale `sessionManager.Get()` check in the middleware, not an independent feature. It belongs with the PR that removes that check.

Key changes:
- Add `RestoreSession()` to `MultiSessionFactory` (and decorating/mock impls); refactor `makeSession` → `makeBaseSession` so `RestoreSession` can reconstruct a live session from stored metadata without a bearer token
- Add `RestoreHijackPrevention()` in `pkg/vmcp/session/internal/security` to recreate the hijack-prevention decorator from stored hash/salt
- Rewrite `sessionmanager.Manager` to use `DataStorage`: node-local `RestorableCache[string, MultiSession]` for hot-path lookups with singleflight-deduplicated `RestoreSession` on cache miss and lazy liveness validation via `storage.Load()` on cache hits
- Extract `RestorableCache[K,V]` generic type (sync.Map + singleflight + check/evict callbacks) into `cache.go`
- Replace `*transportsession.Manager` in `discovery.Middleware` with a `MultiSessionGetter` interface backed by `vmcpSessionMgr`
- **Behavior change**: unknown or expired sessions in the middleware no longer return HTTP 401 — the middleware passes through and the SDK responds with 404 via `Validate()`
- Wire `server.go` to create `LocalSessionDataStorage`, pass it to `sessionmanager.New`, and route the discovery middleware through `vmcpSessionMgr`

Fixes #4220

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Unit tests (task test)
- [ ] E2E tests (task test-e2e)
- [ ] Linting (task lint-fix)

Yes. HTTP clients relying on 401 Unauthorized for unknown/expired session IDs on subsequent requests will now receive 404 Not Found (from the MCP SDK's session `Validate()`). This is semantically more correct — "session not found" is not an authorization failure.

## Large PR Justification

Phases 2 and 3 of the DataStorage migration are combined here because they are not independently deployable: phase 2 alone leaves discovery broken without a throwaway bridge shim, and phase 3 depends on the session manager that phase 2 introduces. See the Summary above for the full reasoning.

### DIFF
--- a/pkg/vmcp/server/sessionmanager/cache.go
+++ b/pkg/vmcp/server/sessionmanager/cache.go
@@ -1,0 +1,162 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package sessionmanager
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+
+	"golang.org/x/sync/singleflight"
+)
+
+// ErrExpired is returned by the check function passed to newRestorableCache to
+// signal that a cached entry has definitively expired and should be evicted.
+var ErrExpired = errors.New("cache entry expired")
+
+// errSentinelFound is returned inside the singleflight load function when a
+// non-V value (e.g. terminatedSentinel) is present in the map. Returning an
+// error aborts the load and causes Get to return (zero, false), consistent
+// with the behaviour of the initial-hit path that also returns (zero, false)
+// for non-V values.
+var errSentinelFound = errors.New("sentinel stored in cache")
+
+// RestorableCache is a node-local write-through cache backed by a sync.Map,
+// with singleflight-deduplicated restore on cache miss and lazy liveness
+// validation on cache hit.
+//
+// Type parameter K is the key type (must be comparable).
+// Type parameter V is the cached value type.
+//
+// Values are stored internally as any, which allows callers to place sentinel
+// markers alongside V entries (e.g. a tombstone during teardown). Get performs
+// a type assertion to V and treats non-V entries as "not found". Peek and
+// Store expose raw any access for sentinel use.
+type RestorableCache[K comparable, V any] struct {
+	m      sync.Map
+	flight singleflight.Group
+
+	// load is called on a cache miss. Return (value, nil) on success.
+	// A successful result is stored in the cache before being returned.
+	load func(key K) (V, error)
+
+	// check is called on every cache hit to confirm liveness. Returning nil
+	// means the entry is alive. Returning ErrExpired means it has definitively
+	// expired (the entry is evicted). Any other error is treated as a transient
+	// failure and the cached value is returned unchanged.
+	check func(key K) error
+
+	// onEvict is called after a confirmed-expired entry has been removed. The
+	// evicted value is passed to allow resource cleanup (e.g. closing
+	// connections). May be nil.
+	onEvict func(key K, v V)
+}
+
+// TODO: add an age-based sweep to bound the lifetime of entries that are
+// never accessed again after their storage TTL expires. The sweep would range
+// over m, compare each entry's insertion time against a caller-supplied maxAge,
+// and call onEvict for entries that are too old — all without touching storage.
+// Until then, entries for idle sessions leak backend connections until the
+// process restarts or the session ID is queried again.
+
+func newRestorableCache[K comparable, V any](
+	load func(K) (V, error),
+	check func(K) error,
+	onEvict func(K, V),
+) *RestorableCache[K, V] {
+	return &RestorableCache[K, V]{
+		load:    load,
+		check:   check,
+		onEvict: onEvict,
+	}
+}
+
+// Get returns the cached V value for key.
+//
+// On a cache hit, check is run first: ErrExpired evicts the entry and returns
+// (zero, false); transient errors return the cached value unchanged. Non-V
+// values stored via Store (e.g. sentinels) return (zero, false) without
+// triggering a restore.
+//
+// On a cache miss, load is called under a singleflight group so at most one
+// restore runs concurrently per key.
+func (c *RestorableCache[K, V]) Get(key K) (V, bool) {
+	if raw, ok := c.m.Load(key); ok {
+		v, isV := raw.(V)
+		if !isV {
+			var zero V
+			return zero, false
+		}
+		if err := c.check(key); err != nil {
+			if errors.Is(err, ErrExpired) {
+				c.m.Delete(key)
+				if c.onEvict != nil {
+					c.onEvict(key, v)
+				}
+				var zero V
+				return zero, false
+			}
+			// Transient error — keep the cached value.
+		}
+		return v, true
+	}
+
+	// Cache miss: use singleflight to prevent concurrent restores for the same key.
+	type result struct{ v V }
+	raw, err, _ := c.flight.Do(fmt.Sprint(key), func() (any, error) {
+		// Re-check the cache: a concurrent singleflight group may have stored
+		// the value between our miss check above and acquiring this group.
+		if stored, ok := c.m.Load(key); ok {
+			if v, isV := stored.(V); isV {
+				return result{v: v}, nil
+			}
+			// Non-V sentinel present (e.g. terminatedSentinel). Treat as a
+			// hard stop: do not call load() and do not overwrite the sentinel.
+			return nil, errSentinelFound
+		}
+		v, loadErr := c.load(key)
+		if loadErr != nil {
+			return nil, loadErr
+		}
+		// Guard against a sentinel being stored between load() completing and
+		// this Store call (Terminate() running concurrently). LoadOrStore is
+		// atomic: if a sentinel got in, we discard the freshly loaded value
+		// via onEvict rather than silently overwriting the sentinel.
+		if _, loaded := c.m.LoadOrStore(key, v); loaded {
+			if c.onEvict != nil {
+				c.onEvict(key, v)
+			}
+			return nil, errSentinelFound
+		}
+		return result{v: v}, nil
+	})
+	if err != nil {
+		var zero V
+		return zero, false
+	}
+	r, ok := raw.(result)
+	return r.v, ok
+}
+
+// Store sets key to value. value may be any type, including sentinel markers.
+func (c *RestorableCache[K, V]) Store(key K, value any) {
+	c.m.Store(key, value)
+}
+
+// Delete removes key from the cache.
+func (c *RestorableCache[K, V]) Delete(key K) {
+	c.m.Delete(key)
+}
+
+// Peek returns the raw value stored under key without type assertion, liveness
+// check, or restore. Used for sentinel inspection.
+func (c *RestorableCache[K, V]) Peek(key K) (any, bool) {
+	return c.m.Load(key)
+}
+
+// CompareAndSwap atomically replaces the value stored under key from old to
+// new. Both old and new may be any type, including sentinels.
+func (c *RestorableCache[K, V]) CompareAndSwap(key K, old, replacement any) bool {
+	return c.m.CompareAndSwap(key, old, replacement)
+}
--- a/pkg/vmcp/server/sessionmanager/session_manager.go
+++ b/pkg/vmcp/server/sessionmanager/session_manager.go
@@ -18,13 +18,11 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"sync"
 	"time"
 
 	"github.com/google/uuid"
 	"github.com/mark3labs/mcp-go/mcp"
 	mcpserver "github.com/mark3labs/mcp-go/server"
-	"golang.org/x/sync/singleflight"
 
 	"github.com/stacklok/toolhive/pkg/auth"
 	transportsession "github.com/stacklok/toolhive/pkg/transport/session"
@@ -44,9 +42,9 @@ const (
 	MetadataValTrue = "true"
 )
 
-// terminatedSentinel is stored in multiSessions when Terminate() begins
-// tearing down a MultiSession. GetMultiSession returns (nil, false) for
-// sentinel entries, and DecorateSession's CAS-based re-check will fail,
+// terminatedSentinel is stored in sessions when Terminate() begins tearing
+// down a MultiSession. sessions.Get returns (nil, false) for sentinel entries
+// (non-V values), and DecorateSession's CAS-based re-check will fail,
 // preventing concurrent writers from resurrecting a storage record that
 // Terminate() has already deleted.
 type terminatedSentinel struct{}
@@ -80,22 +78,12 @@ type Manager struct {
 	factory    vmcpsession.MultiSessionFactory
 	backendReg vmcp.BackendRegistry
 
-	// multiSessions is a node-local performance cache of live MultiSession
-	// objects. It is separate from storage because MultiSession contains
-	// un-serialisable runtime state (HTTP connections, routing tables).
-	// On a cache miss, GetMultiSession falls back to storage.Load +
-	// factory.RestoreSession to reconstruct the session.
-	multiSessions sync.Map // map[string]vmcpsession.MultiSession
-
-	// restoreGroup deduplicates concurrent factory.RestoreSession calls for the
-	// same session ID. Without this, a burst of requests for an uncached session
-	// would each open independent backend connections before any of them had a
-	// chance to populate the local cache.
-	restoreGroup singleflight.Group
-
-	sessionTTL time.Duration
-	stopOnce   sync.Once
-	stopCh     chan struct{}
+	// sessions is a node-local cache of live MultiSession objects, separate
+	// from storage because MultiSession contains un-serialisable runtime state
+	// (HTTP connections, routing tables). On a cache miss it restores the
+	// session from stored metadata; on a cache hit it confirms liveness via
+	// storage.Load, which also refreshes the Redis TTL.
+	sessions *RestorableCache[string, vmcpsession.MultiSession]
 }
 
 // New creates a Manager backed by the given SessionDataStorage and backend
@@ -107,7 +95,6 @@ type Manager struct {
 // on shutdown. If no cleanup is needed, a no-op function is returned.
 func New(
 	storage transportsession.DataStorage,
-	ttl time.Duration,
 	cfg *FactoryConfig,
 	backendRegistry vmcp.BackendRegistry,
 ) (*Manager, func(context.Context) error, error) {
@@ -140,24 +127,29 @@ func New(
 		}
 	}
 
-	// Build the Manager first so we can reference sm.Terminate directly in the
-	// factory closures, eliminating the forward-reference variable pattern.
+	// Build the Manager first so we can reference sm.Terminate and sm.sessions
+	// directly in closures, eliminating the forward-reference variable pattern.
 	sm := &Manager{
 		storage:    storage,
 		backendReg: backendRegistry,
-		sessionTTL: ttl,
-		stopCh:     make(chan struct{}),
 	}
+
+	sm.sessions = newRestorableCache(
+		sm.loadSession,
+		sm.checkSession,
+		func(id string, sess vmcpsession.MultiSession) {
+			if closeErr := sess.Close(); closeErr != nil {
+				slog.Warn("session cache: error closing evicted session",
+					"session_id", id, "error", closeErr)
+			}
+			slog.Warn("session cache: evicted expired session from node-local cache",
+				"session_id", id)
+		},
+	)
 
 	sm.factory = buildDecoratingFactory(cfg, optimizerFactory, instruments, sm.Terminate)
 
-	// Start the eviction loop at half the session TTL so node-local MultiSession
-	// entries are cleaned up promptly when the storage backend (e.g. Redis TTL)
-	// expires them.
-	sm.startEvictionLoop(ttl / 2)
-
 	cleanup := func(ctx context.Context) error {
-		sm.stop()
 		return optimizerCleanup(ctx)
 	}
 	return sm, cleanup, nil
@@ -213,17 +205,18 @@ const decorateTimeout = 5 * time.Second
 // The placeholder is replaced by CreateSession() in Phase 2 once context
 // is available via the OnRegisterSession hook.
 func (sm *Manager) Generate() string {
-	ctx, cancel := context.WithTimeout(context.Background(), generateTimeout)
-	defer cancel()
-
 	// Two attempts: the second handles both storage transients and the
 	// astronomically unlikely (but now correctly detected) UUID collision.
+	// Each attempt gets its own context so an expired deadline on attempt 0
+	// does not immediately abort attempt 1.
 	for attempt := range 2 {
+		ctx, cancel := context.WithTimeout(context.Background(), generateTimeout)
 		sessionID := uuid.New().String()
 
-		// StoreIfAbsent is an atomic SET NX on Redis, eliminating the TOCTOU
-		// race that Exists+Store would have in a multi-pod deployment.
-		stored, err := sm.storage.StoreIfAbsent(ctx, sessionID, map[string]string{})
+		// Create is an atomic SET NX on Redis, eliminating the TOCTOU
+		// race that a Load+Upsert would have in a multi-pod deployment.
+		stored, err := sm.storage.Create(ctx, sessionID, map[string]string{})
+		cancel()
 		if err != nil {
 			slog.Error("Manager: failed to store placeholder session",
 				"session_id", sessionID, "attempt", attempt+1, "error", err)
@@ -303,6 +296,7 @@ func (sm *Manager) CreateSession(
 	allowAnonymous := sessiontypes.ShouldAllowAnonymous(identity)
 	sess, err := sm.factory.MakeSessionWithID(ctx, sessionID, identity, allowAnonymous, backends)
 	if err != nil {
+		sm.cleanupFailedPlaceholder(sessionID, placeholder)
 		return nil, fmt.Errorf("Manager.CreateSession: failed to create multi-session: %w", err)
 	}
 
@@ -324,6 +318,7 @@ func (sm *Manager) CreateSession(
 	}
 	if err != nil {
 		_ = sess.Close()
+		sm.cleanupFailedPlaceholder(sessionID, placeholder)
 		return nil, fmt.Errorf(
 			"Manager.CreateSession: failed to re-check placeholder for session %q after backend init: %w",
 			sessionID, err,
@@ -342,20 +337,37 @@ func (sm *Manager) CreateSession(
 	// MultiSession itself is cached in the node-local multiSessions map below.
 	storeCtx, storeCancel := context.WithTimeout(ctx, createSessionStorageTimeout)
 	defer storeCancel()
-	if err := sm.storage.Store(storeCtx, sessionID, sess.GetMetadata()); err != nil {
-		// Best-effort close of the newly created session to release backend connections.
+	if err := sm.storage.Upsert(storeCtx, sessionID, sess.GetMetadata()); err != nil {
 		_ = sess.Close()
+		sm.cleanupFailedPlaceholder(sessionID, placeholder2)
 		return nil, fmt.Errorf("Manager.CreateSession: failed to store session metadata: %w", err)
 	}
 
-	// Cache the live MultiSession in the node-local map so that GetMultiSession
-	// can retrieve it without a storage round-trip on the hot path.
-	sm.multiSessions.Store(sessionID, sess)
+	// Cache the live MultiSession so that GetMultiSession can retrieve it.
+	sm.sessions.Store(sessionID, sess)
 
 	slog.Debug("Manager: created multi-session",
 		"session_id", sessionID,
 		"backend_count", len(backends))
 	return sess, nil
+}
+
+// cleanupFailedPlaceholder marks a placeholder session as terminated in storage
+// after a CreateSession failure. This prevents Validate() from returning
+// (false, nil) for an orphaned placeholder (which would make the SDK treat it
+// as a valid session), and prevents repeated Validate() calls from refreshing
+// the Redis TTL and keeping the placeholder alive indefinitely.
+//
+// Cleanup is best-effort: errors are logged but not returned, since the caller
+// already has an error to report.
+func (sm *Manager) cleanupFailedPlaceholder(sessionID string, metadata map[string]string) {
+	metadata[MetadataKeyTerminated] = MetadataValTrue
+	cleanupCtx, cancel := context.WithTimeout(context.Background(), createSessionStorageTimeout)
+	defer cancel()
+	if err := sm.storage.Upsert(cleanupCtx, sessionID, metadata); err != nil {
+		slog.Warn("Manager.CreateSession: failed to mark failed placeholder as terminated; it will linger until TTL expires",
+			"session_id", sessionID, "error", err)
+	}
 }
 
 // Validate implements the SDK's SessionIdManager.Validate().
@@ -418,9 +430,9 @@ func (sm *Manager) Terminate(sessionID string) (isNotAllowed bool, err error) {
 	ctx, cancel := context.WithTimeout(context.Background(), terminateTimeout)
 	defer cancel()
 
-	// Check the node-local map first: a fully-formed MultiSession is cached here
-	// while this pod owns it.
-	if v, ok := sm.multiSessions.Load(sessionID); ok {
+	// Check the node-local cache first: a fully-formed MultiSession is stored
+	// here while this pod owns it.
+	if v, ok := sm.sessions.Peek(sessionID); ok {
 		// A terminatedSentinel means another goroutine is already tearing down
 		// this session. Do not fall through to the placeholder path — that would
 		// race with the concurrent Terminate's storage.Delete and potentially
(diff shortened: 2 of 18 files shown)


pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -537,3 +761,161 @@ func (sm *Manager) GetAdaptedResources(sessionID string) ([]mcpserver.ServerReso
...
+
+		sdkPrompts = append(sdkPrompts, mcpserver.ServerPrompt{
+			Prompt:  prompt,
+			Handler: handler,
+		})
+		slog.Debug("Manager.GetAdaptedPrompts: adapted prompt", "session_id", sessionID, "prompt", domainPrompt.Name)
+	}
+
+	return sdkPrompts, nil
+}
```
jerm-dro, line 861: **blocker:** This function polls `storage.Exists()` for every cached session, but we're removing `Exists` from `DataStorage` (see feedback on #4458). The two are at odds — `DataStorage` is the source of truth for session liveness, and callers should discover expiry via `Load`, not a separate `Exists` probe.

The race this loop tries to prevent — a `MultiSession` lingering in the local cache after its storage entry has expired — is something we're going to have to contend with regardless. Even with this loop, there's always a window where the cache holds an entry whose underlying data has already expired. Proactive polling narrows that window but doesn't close it.

A simpler model that aligns with #4458 and accepts this reality:

1. **Drop the proactive eviction loop entirely.** The node-local `sync.Map` is a performance cache, not a source of truth. It doesn't need to be eagerly synchronized with storage.
2. **Invalidate lazily.** When a request arrives, `GetMultiSession` checks the local cache. On a hit, call `storage.Load()` to confirm the session is still alive (this also refreshes the Redis TTL). If `Load` returns `ErrSessionNotFound`, evict the cache entry, close backend connections, and return `(nil, false)`.
3. **Bound the cache independently.** If connection leaks for sessions that are never requested again are a concern, add a simple capacity limit or age-based sweep on the local `sync.Map` itself — no storage interaction needed. This is a cache policy, not a storage contract.

This eliminates `Exists` from both `DataStorage` and the old `Storage` interface, removes the eviction loop and its associated timeouts/goroutine, and keeps the storage interface to `Create` / `Upsert` / `Load` / `Delete` / `Close`.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -352,24 +500,90 @@ func (sm *Manager) Terminate(sessionID string) (isNotAllowed bool, err error) {
...
 // race with Terminate(): if the session is deleted between GetMultiSession and
-// UpsertSession, the upsert would silently resurrect a terminated session. The
-// re-check catches that window. A narrow TOCTOU gap remains between the
-// re-check and the upsert, but its consequence is bounded: Terminate() already
+// the store, the store would silently resurrect a terminated session.
+// The re-check catches that window. A narrow TOCTOU gap remains between the
+// re-check and the store, but its consequence is bounded: Terminate() already
 // called Close() on the underlying MultiSession before deleting it, so any
 // resurrected decorator wraps an already-closed session and will fail on first
 // use rather than leaking backend connections.
```
jerm-dro, line 616: **suggestion:** Consider extracting the `sync.Map` + `singleflight` cache-miss pattern into a generic, independently testable component (e.g., `RestorableCache[K, V]`).

The pattern here — local cache check, singleflight-deduplicated restore on miss, lazy invalidation — is generic and not session-specific. Extracting it would:

- Make the cache/restore/invalidation mechanics unit-testable without mock factories, mock storage, or session domain setup
- Keep `sessionmanager.Manager` focused on domain logic (two-phase creation, termination, decoration) rather than cache mechanics
- Provide a natural home for future cache policy (capacity limits, age-based sweep) without further bloating the session manager

The session manager would supply a `load` function that does `storage.Load` + `factory.RestoreSession`, and domain-specific coordination like `terminatedSentinel` would stay in the session manager.

pkg/vmcp/session/factory.go
```diff
@@ -480,16 +479,115 @@ func (f *defaultMultiSessionFactory) makeSession(
...
+	}
+
+	// Filter allBackends to the subset originally connected in this session.
+	filteredBackends := filterBackendsByStoredIDs(allBackends, storedMetadata[MetadataKeyBackendIDs])
+
+	// Reconstruct a minimal identity from stored metadata. The original bearer
+	// token is never persisted (only its HMAC-SHA256 hash is), so Token is empty.
+	// The security decorator is restored from the stored hash/salt below.
+	var identity *auth.Identity
+	if subject := storedMetadata[MetadataKeyIdentitySubject]; subject != "" {
```
jerm-dro, line 532: **suggestion:** `RestoreSession` has intimate knowledge of individual metadata keys (`MetadataKeyBackendIDs`, `MetadataKeyBackendSessionPrefix`, `MetadataKeyIdentitySubject`, `MetadataKeyTokenHash`, `MetadataKeyTokenSalt`) scattered across packages. Each layer (factory, security, session) reads and writes its own keys but they're all flattened into a single `map[string]string`, making it hard to reason about ownership and easy to accidentally overwrite or miss keys.

A more data-agnostic pattern would give each session layer its own opaque metadata entry — e.g., `map[metadataKey][]byte` where each layer owns exactly one key and serializes/deserializes its own state. Restoring a session becomes: unmarshal each layer's blob and pass it to that layer's constructor. No cross-layer key knowledge needed.

Not blocking and doesn't need to be addressed now — this can wait until the horizontal scaling work is complete and the full set of metadata requirements is clearer.

pkg/vmcp/server/sessionmanager/cache.go
```diff
@@ -0,0 +1,162 @@
...
+// and call onEvict for entries that are too old — all without touching storage.
+// Until then, entries for idle sessions leak backend connections until the
+// process restarts or the session ID is queried again.
+
+func newRestorableCache[K comparable, V any](
+	load func(K) (V, error),
+	check func(K) error,
+	onEvict func(K, V),
+) *RestorableCache[K, V] {
+	return &RestorableCache[K, V]{
```
jerm-dro, line 68: **blocker:** `RestorableCache` has no capacity limit and no background eviction. The only eviction path is the lazy `check()` on cache hits — if a session's Redis TTL expires while the session is idle (user closes browser, network drop, etc.), the `MultiSession` and its backend HTTP connections stay in the `sync.Map` indefinitely.

In a multi-tenant deployment with session churn, this means unbounded memory and connection growth until the process restarts.

This can be a fast follow, but it needs to be tracked before this merges. My preference: add a capacity limit so the cache evicts oldest entries when full, bounding the worst case. An age-based sweep (as the TODO here suggests) would be complementary but is more work. I'd suggest using some open source cache rather than rolling our own LRU implementation. Some options:
- https://github.com/hashicorp/golang-lru
- https://github.com/dgraph-io/ristretto

Whichever cache you use becomes an implementation detail. The unique behavior we wrap on top is `check()`.

pkg/vmcp/server/sessionmanager/cache.go
```diff
@@ -0,0 +1,162 @@
...
+//
+// Type parameter K is the key type (must be comparable).
+// Type parameter V is the cached value type.
+//
+// Values are stored internally as any, which allows callers to place sentinel
+// markers alongside V entries (e.g. a tombstone during teardown). Get performs
+// a type assertion to V and treats non-V entries as "not found". Peek and
+// Store expose raw any access for sentinel use.
+type RestorableCache[K comparable, V any] struct {
+	m      sync.Map
```
jerm-dro, line 37: **blocker:** The sentinel pattern (`terminatedSentinel`, `Peek`, untyped `Store(any)`, untyped `CompareAndSwap(any, any)`) adds significant complexity to prevent a specific race: `DecorateSession` resurrecting a storage record after `Terminate` deletes it. This works, but it pushes a storage-layer concern into the cache, punches through the cache's type safety, and makes the concurrency model hard to reason about.

A simpler design that separates concerns cleanly:

**1. Cache invariant — "no stale reads":** `check` must pass on every insertion and retrieval. The cache is fully typed (`Store(K, V)`, no `any`). `Terminate` just deletes from storage then evicts from the cache. No sentinels needed — any concurrent `Get()` or `Store()` runs `check`, which calls `storage.Load()`. If the record is gone, the operation is rejected.

**2. Storage invariant — "no resurrection after delete":** Replace `Upsert` with `Update` (Redis `SET XX`), which succeeds only if the key exists and returns `ErrSessionNotFound` if it doesn't. All current `Upsert` callers (`CreateSession`, `DecorateSession`, `cleanupFailedPlaceholder`, Terminate's placeholder path) expect the record to already exist — none of them need unconditional write semantics. With this contract, the DecorateSession race resolves naturally:

1. `Terminate`: `Delete(sessionID)` → succeeds
2. `DecorateSession`: `Update(sessionID, metadata)` → returns `ErrSessionNotFound`
3. `DecorateSession` rolls back the cache entry, returns error

This eliminates `Peek`, `terminatedSentinel`, untyped `Store`/`CompareAndSwap`, and the rollback logic in `Terminate`. Each layer owns one invariant: the cache guarantees liveness, the storage guarantees no resurrection.

This can be a fast follow — the current sentinel approach is correct and the race it prevents is narrow (client DELETE during initialization). But the complexity cost is real and should be addressed before horizontal scaling ships.

pkg/vmcp/server/sessionmanager/cache.go
jerm-dro, line 1: **suggestion:** `RestorableCache` is a general-purpose generic type (sync.Map + singleflight + liveness check) with no session-specific logic. Burying it inside `sessionmanager` makes it harder to discover and reuse. Consider moving it to `pkg/cache/` (new top-level package) — that's where general infrastructure lives, separate from the domain-specific `pkg/vmcp/cache/` (token caching).

If this is addressed in a follow-up, please use `git mv` in a separate commit from any functional changes so the move and the behavioral changes can be reviewed independently.

pkg/vmcp/server/sessionmanager/cache.go
jerm-dro, line 37: **suggestion:** Consider renaming `RestorableCache` to `ValidatingCache`. The distinguishing feature of this type is the invariant — every entry is validated against an external source of truth on every access — not the load-on-miss backfill, which is standard cache behavior. `ValidatingCache` communicates the contract directly and stays generic if this moves to `pkg/cache/`.

pkg/vmcp/server/sessionmanager/cache.go
yrobla: https://github.com/stacklok/toolhive/issues/4496

pkg/vmcp/server/sessionmanager/cache.go
yrobla: https://github.com/stacklok/toolhive/issues/4494

pkg/vmcp/server/sessionmanager/cache.go
yrobla: https://github.com/stacklok/toolhive/issues/4495

PR #4487, opened 2026-04-01
Restructure MCPTelemetryConfig CRD to nested spec per RFC THV-0023

### LINKED ISSUE #4262: Define MCPTelemetryConfig API Types
## Description

Create the `MCPTelemetryConfig` CRD type definitions using a nested structure with `openTelemetry` and `prometheus` sub-objects, as specified in [RFC THV-0023](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0023-crd-v1beta1-optimization.md).

### Why nested instead of flat?

The original implementation used inline embedding of `telemetry.Config` which exposed CLI-only fields (`environmentVariables`) and per-server fields (`serviceName`) in the CRD spec. The RFC specifies a nested structure that:
- Groups OTLP settings under `spec.openTelemetry` (endpoint, headers, tracing, metrics)
- Groups Prometheus settings under `spec.prometheus`
- Excludes `serviceName` (per-server, set via `MCPTelemetryConfigReference`)
- Excludes `environmentVariables` and `customAttributes` (CLI-only)
- Uses `resourceAttributes` instead of `customAttributes` for OTel resource attributes

### Target Structure

```yaml
kind: MCPTelemetryConfig
metadata:
  name: production-telemetry
  namespace: mcp-platform
spec:
  openTelemetry:
    enabled: true
    endpoint: otel-collector.monitoring.svc.cluster.local:4318
    insecure: false
    sensitiveHeaders:
      - name: Authorization
        secretKeyRef:
          name: otel-auth-secret
          key: bearer-token
      - name: X-API-Key
        secretKeyRef:
          name: otel-secret
          key: api-key
    resourceAttributes:
      service.namespace: toolhive
      deployment.environment: production
    metrics:
      enabled: true
    tracing:
      enabled: true
      samplingRate: "0.1"
    useLegacyAttributes: true
  prometheus:
    enabled: true
status:
  observedGeneration: 1
  configHash: "sha256:..."
  referencingWorkloads:
    - kind: MCPServer
      name: server-a
  conditions:
    - type: Valid
      status: "True"
```

### Go Types

```go
type MCPTelemetryConfigSpec struct {
    OpenTelemetry *MCPTelemetryOTelConfig `json:"openTelemetry,omitempty"`
    Prometheus    *PrometheusConfig        `json:"prometheus,omitempty"`
}

type MCPTelemetryOTelConfig struct {
    Enabled            bool                        `json:"enabled,omitempty"`
    Endpoint           string                      `json:"endpoint,omitempty"`
    Insecure           bool                        `json:"insecure,omitempty"`
    Headers            map[string]string            `json:"headers,omitempty"`
    SensitiveHeaders   []SensitiveHeader            `json:"sensitiveHeaders,omitempty"`
    ResourceAttributes map[string]string            `json:"resourceAttributes,omitempty"`
    Metrics            *OpenTelemetryMetricsConfig  `json:"metrics,omitempty"`
    Tracing            *OpenTelemetryTracingConfig  `json:"tracing,omitempty"`
    UseLegacyAttributes bool                        `json:"useLegacyAttributes"`
}

type MCPTelemetryConfigStatus struct {
    Conditions           []metav1.Condition  `json:"conditions,omitempty"`
    ObservedGeneration   int64               `json:"observedGeneration,omitempty"`
    ConfigHash           string              `json:"configHash,omitempty"`
    ReferencingWorkloads []WorkloadReference  `json:"referencingWorkloads,omitempty"`
}

// WorkloadReference is shared across CRDs (defined in mcpoidcconfig_types.go)
type WorkloadReference struct {
    Kind string `json:"kind"` // MCPServer, VirtualMCPServer, MCPRemoteProxy
    Name string `json:"name"`
}
```

### Key Design Decisions

- `Headers` is `map[string]string` for non-sensitive headers (e.g., `X-Custom-Tag: value`). Credential headers (bearer tokens, API keys) should use `sensitiveHeaders` with `SecretKeyRef` instead.
- `SensitiveHeaders` uses `SecretKeyRef` for credential headers (no inline secrets)
- `ResourceAttributes` replaces `customAttributes` — maps to `telemetry.Config.CustomAttributes` at runtime
- `ServiceName` excluded from shared config — it is per-server via `MCPTelemetryConfigReference`
- `WorkloadReference` is a shared type (defined in `mcpoidcconfig_types.go`) with `Kind` + `Name` — namespace is implicit since cross-namespace refs are not supported
- Conversion from CRD nested types to flat `telemetry.Config` is handled by `spectoconfig.NormalizeMCPTelemetryConfig`

### Phase 2 Fields (not yet implemented)

These RFC fields require runtime support and will be added in follow-up PRs:
- `spec.openTelemetry.protocol` (http/grpc) — needs gRPC exporter support
- `spec.openTelemetry.metrics.exportInterval` — needs `sdkmetric.WithInterval()`
- `spec.openTelemetry.tracing.propagators` — needs dynamic propagator factory
- `spec.prometheus.path` — hardcoded to `/metrics`
- `spec.prometheus.port` — shares transport port

### Acceptance Criteria

- [x] `MCPTelemetryConfigSpec` uses nested `openTelemetry`/`prometheus` structure matching RFC THV-0023
- [x] `Headers` is `map[string]string`
- [x] `SensitiveHeaders` uses `SecretKeyRef` for credential headers
- [x] `ResourceAttributes` field for shared OTel resource attributes
- [x] `ServiceName` excluded from shared config (per-server field)
- [x] `ReferencingWorkloads` uses shared `WorkloadReference` type (Kind + Name)
- [x] Conversion layer (`spectoconfig.NormalizeMCPTelemetryConfig`) maps nested CRD to flat `telemetry.Config`
- [x] `task operator-generate` and `task operator-manifests` succeed
- [x] Unit tests verify type structure and conversion correctness
- [x] Tests pass (`task test`)

ChrisJBurns:
## Summary

- The MCPTelemetryConfig CRD spec was using a flat inline embedding of `telemetry.Config` which exposed CLI-only fields (`environmentVariables`) and didn't match the nested `openTelemetry`/`prometheus` structure specified in [RFC THV-0023](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0023-crd-v1beta1-optimization.md)
- Restructure to use `spec.openTelemetry` and `spec.prometheus` sub-objects, add `resourceAttributes` (replacing `customAttributes`), move `sensitiveHeaders` under `openTelemetry`, and replace `referencingServers []string` with structured `referencingWorkloads []WorkloadReference`
- Update the conversion layer to map nested CRD types to flat `telemetry.Config` at runtime
- Add operator example YAML files for MCPTelemetryConfig and MCPServer with telemetryConfigRef

Closes #4262

## Type of change

## Test plan

## Changes

| File | Change |
|------|--------|
| `api/v1alpha1/mcptelemetryconfig_types.go` | Replace flat `telemetry.Config` inline embed with nested `MCPTelemetryOTelConfig`/`PrometheusConfig`. Add `WorkloadReference` type. Update validation, CEL rules, printcolumns |
| `api/v1alpha1/zz_generated.deepcopy.go` | Regenerated for new types |
| `controllers/mcptelemetryconfig_controller.go` | `findReferencingServers` → `findReferencingWorkloads` returning `[]WorkloadReference`. Extract `workloadRefsEqual` helper to fix cyclomatic complexity |
| `pkg/spectoconfig/telemetry.go` | Rewrite `NormalizeMCPTelemetryConfig` from one-liner `config := spec.Config` to field-by-field mapping from nested CRD to flat `telemetry.Config` |
| `pkg/controllerutil/telemetry.go` | Access sensitive headers via `Spec.OpenTelemetry.SensitiveHeaders` |
| `deploy/charts/operator-crds/` | CRD manifests regenerated |
| `examples/operator/telemetry-configs/` | New example YAMLs: basic config, production config with sensitiveHeaders, MCPServer with telemetryConfigRef |

Yes — this is a **breaking change** to the `MCPTelemetryConfig` CRD spec structure. Since the API is `v1alpha1` (explicitly unstable), breaking changes are expected. Users must update their MCPTelemetryConfig manifests from the flat structure to the nested structure. See the examples in `examples/operator/telemetry-configs/` for the new format.

## Large PR Justification
- is just a api type change with a lot of auto generation

### DIFF
--- a/cmd/thv-operator/api/v1alpha1/mcptelemetryconfig_types.go
+++ b/cmd/thv-operator/api/v1alpha1/mcptelemetryconfig_types.go
@@ -95,18 +95,6 @@ type MCPTelemetryConfigSpec struct {
 	Prometheus *PrometheusConfig `json:"prometheus,omitempty"`
 }
 
-// WorkloadReference identifies a Kubernetes workload that references a shared config resource.
-type WorkloadReference struct {
-	// Kind is the resource kind (e.g., "MCPServer")
-	Kind string `json:"kind"`
-
-	// Namespace is the resource namespace
-	Namespace string `json:"namespace"`
-
-	// Name is the resource name
-	Name string `json:"name"`
-}
-
 // MCPTelemetryConfigStatus defines the observed state of MCPTelemetryConfig
 type MCPTelemetryConfigStatus struct {
 	// Conditions represent the latest available observations of the MCPTelemetryConfig's state
--- a/cmd/thv-operator/controllers/mcptelemetryconfig_controller.go
+++ b/cmd/thv-operator/controllers/mcptelemetryconfig_controller.go
@@ -196,7 +196,7 @@ func (r *MCPTelemetryConfigReconciler) handleDeletion(
 	if len(referencingWorkloads) > 0 {
 		names := make([]string, 0, len(referencingWorkloads))
 		for _, ref := range referencingWorkloads {
-			names = append(names, fmt.Sprintf("%s/%s", ref.Namespace, ref.Name))
+			names = append(names, fmt.Sprintf("%s/%s", ref.Kind, ref.Name))
 		}
 		msg := fmt.Sprintf("cannot delete: still referenced by MCPServer(s): %v", names)
 		logger.Info(msg, "telemetryConfig", telemetryConfig.Name)
@@ -239,19 +239,12 @@ func (r *MCPTelemetryConfigReconciler) findReferencingWorkloads(
 		if server.Spec.TelemetryConfigRef != nil &&
 			server.Spec.TelemetryConfigRef.Name == telemetryConfig.Name {
 			refs = append(refs, mcpv1alpha1.WorkloadReference{
-				Kind:      "MCPServer",
-				Namespace: server.Namespace,
-				Name:      server.Name,
+				Kind: "MCPServer",
+				Name: server.Name,
 			})
 		}
 	}
 	slices.SortFunc(refs, func(a, b mcpv1alpha1.WorkloadReference) int {
-		if a.Namespace != b.Namespace {
-			if a.Namespace < b.Namespace {
-				return -1
-			}
-			return 1
-		}
 		if a.Name < b.Name {
 			return -1
 		}
@@ -266,6 +259,6 @@ func (r *MCPTelemetryConfigReconciler) findReferencingWorkloads(
 // workloadRefsEqual compares two WorkloadReference slices for equality.
 func workloadRefsEqual(a, b []mcpv1alpha1.WorkloadReference) bool {
 	return slices.EqualFunc(a, b, func(x, y mcpv1alpha1.WorkloadReference) bool {
-		return x.Kind == y.Kind && x.Namespace == y.Namespace && x.Name == y.Name
+		return x.Kind == y.Kind && x.Name == y.Name
 	})
 }
--- a/cmd/thv-operator/controllers/mcptelemetryconfig_controller_test.go
+++ b/cmd/thv-operator/controllers/mcptelemetryconfig_controller_test.go
@@ -704,8 +704,8 @@ func TestMCPTelemetryConfigReconciler_ReferenceTracking(t *testing.T) {
 
 	// ReferencingWorkloads should list server-a and server-b (sorted), but not server-c
 	assert.Equal(t, []mcpv1alpha1.WorkloadReference{
-		{Kind: "MCPServer", Namespace: "default", Name: "server-a"},
-		{Kind: "MCPServer", Namespace: "default", Name: "server-b"},
+		{Kind: "MCPServer", Name: "server-a"},
+		{Kind: "MCPServer", Name: "server-b"},
 	}, updated.Status.ReferencingWorkloads)
 }
 
--- a/deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcptelemetryconfigs.yaml
+++ b/deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcptelemetryconfigs.yaml
@@ -250,22 +250,24 @@ spec:
                 description: ReferencingWorkloads lists workloads that reference this
                   MCPTelemetryConfig
                 items:
-                  description: WorkloadReference identifies a Kubernetes workload
-                    that references a shared config resource.
+                  description: |-
+                    WorkloadReference identifies a workload that references a shared configuration resource.
+                    Namespace is implicit — cross-namespace references are not supported.
                   properties:
                     kind:
-                      description: Kind is the resource kind (e.g., "MCPServer")
+                      description: Kind is the type of workload resource
+                      enum:
+                      - MCPServer
+                      - VirtualMCPServer
+                      - MCPRemoteProxy
                       type: string
                     name:
-                      description: Name is the resource name
-                      type: string
-                    namespace:
-                      description: Namespace is the resource namespace
+                      description: Name is the name of the workload resource
+                      minLength: 1
                       type: string
                   required:
                   - kind
                   - name
-                  - namespace
                   type: object
                 type: array
             type: object
--- a/deploy/charts/operator-crds/templates/toolhive.stacklok.dev_mcptelemetryconfigs.yaml
+++ b/deploy/charts/operator-crds/templates/toolhive.stacklok.dev_mcptelemetryconfigs.yaml
@@ -253,22 +253,24 @@ spec:
                 description: ReferencingWorkloads lists workloads that reference this
                   MCPTelemetryConfig
                 items:
-                  description: WorkloadReference identifies a Kubernetes workload
-                    that references a shared config resource.
+                  description: |-
+                    WorkloadReference identifies a workload that references a shared configuration resource.
+                    Namespace is implicit — cross-namespace references are not supported.
                   properties:
                     kind:
-                      description: Kind is the resource kind (e.g., "MCPServer")
+                      description: Kind is the type of workload resource
+                      enum:
+                      - MCPServer
+                      - VirtualMCPServer
+                      - MCPRemoteProxy
                       type: string
                     name:
-                      description: Name is the resource name
-                      type: string
-                    namespace:
-                      description: Namespace is the resource namespace
+                      description: Name is the name of the workload resource
+                      minLength: 1
                       type: string
                   required:
                   - kind
                   - name
-                  - namespace
                   type: object
                 type: array
             type: object
--- a/docs/operator/crd-api.md
+++ b/docs/operator/crd-api.md
@@ -698,7 +698,6 @@ Config holds the configuration for OpenTelemetry instrumentation.
 
 _Appears in:_
 - [vmcp.config.Config](#vmcpconfigconfig)
-- [api.v1alpha1.MCPTelemetryConfigSpec](#apiv1alpha1mcptelemetryconfigspec)
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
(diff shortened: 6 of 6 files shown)


ChrisJBurns:
@claude review this PR focusing only on critical and high severity issues. Ignore medium/low severity, style preferences, and controversial opinions. Focus on:

- Correctness bugs (wrong logic, missing nil checks, data loss)
- Security issues (credential exposure, injection)
- Breaking changes that aren't documented
- Test coverage gaps for critical paths

ChrisJBurns:
@claude re-review this PR. The two high-severity issues from your previous review have been addressed:

1. **Enabled flag ignored** — `NormalizeMCPTelemetryConfig` now gates on `otel.Enabled` (commit fa4e4faa)
2. **Endpoint without tracing/metrics** — Added CEL validation rule rejecting endpoint when neither is enabled (commit fa4e4faa)

Both have new test cases. Please verify the fixes and flag any remaining critical/high issues only.

cmd/thv-operator/api/v1alpha1/mcptelemetryconfig_types.go
```diff
@@ -25,24 +23,89 @@ type SensitiveHeader struct {
...
+//   - Adds SensitiveHeaders for Kubernetes Secret-backed credentials
+//   - Adds ResourceAttributes for shared OTel resource attributes
 //
-// +kubebuilder:validation:XValidation:rule="!has(self.environmentVariables)",message="environmentVariables is a CLI-only field and cannot be set in MCPTelemetryConfig; use customAttributes for resource attributes"
 // +kubebuilder:validation:XValidation:rule="!has(self.headers) || !has(self.sensitiveHeaders) || self.sensitiveHeaders.all(sh, !(sh.name in self.headers))",message="a header name cannot appear in both headers and sensitiveHeaders"
+// +kubebuilder:validation:XValidation:rule=”!has(self.endpoint) || size(self.endpoint) == 0 || (has(self.tracing) && self.tracing.enabled) || (has(self.metrics) && self.metrics.enabled)”,message=”endpoint requires at least one of tracing or metrics to be enabled”
 //
 //nolint:lll // CEL validation rules exceed line length limit
-type MCPTelemetryConfigSpec struct {
-	telemetry.Config `json:",inline"` // nolint:revive
```
jerm-dro, line 39: Blocker: I actually strongly prefer this version. I wrote about this at this point in the RFC here: https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0023-crd-v1beta1-optimization.md#crd-types-and-application-config-relationship

Is there any reason we can't keep it this way?

ChrisJBurns: Good question — the type unification principle from the RFC is sound and I agree with the motivation (eliminating conversion bugs, single source of truth).

The problem is that **Go doesn't allow overriding JSON struct tags on embedded fields**. When we inline-embed `telemetry.Config`, the CRD spec exposes _every_ field from the canonical type, including:

- `environmentVariables` — CLI-only, not applicable to CRD-managed telemetry
- `serviceName` — must NOT be in the shared config (it's per-server via `MCPTelemetryConfigReference`)  
- `customAttributes` — CLI-only (the CRD equivalent is `resourceAttributes`)

The original implementation used CEL to block `environmentVariables` at admission, but `serviceName` and `customAttributes` leaked through silently. There's no way to use `json:"-"` on another struct's fields from the embedding site.

Issue #4262 itself noted this:
> **Why not embed `telemetry.Config` directly?** Go does not allow overriding JSON struct tags on fields of an embedded struct.

The RFC also shows the target YAML as a nested `openTelemetry`/`prometheus` structure (not a flat embedding), and the existing deprecated inline `MCPServer.Spec.Telemetry` already uses this nested pattern successfully.

The conversion layer is intentionally thin — `NormalizeMCPTelemetryConfig` is ~20 lines of field mapping, well-tested, and follows the exact same pattern as the existing `ConvertTelemetryConfig` for the inline path.

**Could we reconcile both goals?** If we restructured `telemetry.Config` itself to use the nested `openTelemetry`/`prometheus` shape and excluded CLI-only fields, we could potentially re-embed it. But that would be a much larger change touching the CLI, VirtualMCPServer config, and all telemetry consumers. Happy to discuss if that's the preferred long-term direction.

jerm-dro: I'm fine with not reusing all of `telemetry.Config` if there's some reason the whole struct shouldn't be included. Could we modify the `telemetry.Config` so the safe to embed parts are reusable without actually requiring a data migration:

```go
// TelemetryConfig
type Config struct {
     EmbeddedConfig `json:",inline"`
     // EnvVars, etc are down below
}
```

Then `EmbeddedConfig` would contain the safe to use fields and go directly in the CRD. The JSON schema would remain the same. You won't even have to touch the other consumers of the telemetry config, since `config.SomeEmbeddedField` should continue to work.

ChrisJBurns: Yeah there's a couple reasons around lack of overrides and we're also coupling our pkg `telemetry.Config` to the CRD itself. If we ever need to change the struct for CLI reasons for example, this would mean additional fields get put into the CRD for no reason.

There's a couple of ways of doing this from what I can see:
1. Use inline embedded structs that reduce DTO/converter code but they don't allow for overrides and they couple the CRD with the pkg config structs which could cause issues if we ever need to add/remove fields. It also causes leaks. They also result in a flat structure on the CRD itself which isn't very intuitive from a UX perspective
2. Nested types: This allows for us to control the structure of them on the CRD (and avoid flat structure), we decouple from the `telemetry.Config` so any changes to the underlying `telemetry.Config` doesn't result in CRD changes. There are no leaks and it supports overrides, The only cost is addition converter code, but I think that's necessary/justified for the decoupling
3. We restructure the underlying `telemetry.Config` itself, but that only solves the flat structure problem, not the coupling, leaking and override issues. We could do a bit more of an advanced restructure for Shared/CLIOnly/Operator but I think that's probably overkill.

jerm-dro: My suggestion above doesn't address:
> The RFC also shows the target YAML as a nested openTelemetry/prometheus structure

but to me this is a nice to have. I'd prioritize avoiding taking on the debt of creating new conversion layers and types. We can always come back and make the embedded type nested too, if that was important enough.

ChrisJBurns:
**Summary of offline discussion**

Jeremy and I discussed this concern in depth offline. Here's the conclusion we arrived at:

**Why we're keeping separate types (CRD types vs. runtime config types):**

1. **Override safety** — Go's `json:",inline"` embedding exposes every exported field from the embedded struct in the CRD schema. There's no way to hide or re-tag individual fields from the embedding site. For example, if `telemetry.Config` has `serviceName` (which the CLI needs), the CRD would expose it too — even though it must not be user-configurable per-server. The only options would be CEL admission rules (field is still visible in `kubectl explain` and the schema) or not embedding. Separate types give us explicit control over the CRD surface area.

2. **CRD stability** — The cost of a missed conversion layer is a feature that doesn't work, which gets caught before or shortly after a PR is submitted. The cost of an unintended field leaking into the CRD schema is potentially a breaking API change. We'd rather fail loudly on a missing mapping than silently expand the CRD contract.

**What we agreed to improve going forward:**

We acknowledged that the duplication and manual conversion layers are a real maintenance burden and error-prone. We agreed on this direction:

- **Keep types split** — there are genuine reasons for the structural differences.
- **Automate drift detection** — use reflection-based tests that enumerate fields on both the CRD type and the runtime config type, and fail if a field isn't either mapped in the conversion function or explicitly listed in an exclusion set. This catches missed conversions at test time without needing expensive e2e tests.
- **Standardize conversion patterns** — make all type conversions follow a consistent structure to reduce cognitive load.
- **RFC in progress** — I've opened a draft RFC at [stacklok/toolhive-rfcs#65](https://github.com/stacklok/toolhive-rfcs/pull/65) to formalize the approach. Jeremy will review later this week.

**TL;DR:** Merging this PR as-is. The separate types are intentional for safety, and we'll address the maintainability concerns through automated drift detection and a broader RFC.

PR #4493, opened 2026-04-01
Wire MCPOIDCConfig into VirtualMCPServer controller

### LINKED ISSUE #4248: MCPOIDCConfig CRD — Types, Controller, and Tests
**As a** platform engineer managing multiple MCP servers with the same identity provider,
**I want** to define OIDC configuration once in an `MCPOIDCConfig` resource and reference it from multiple workloads,
**so that** I can update issuer settings in one place instead of duplicating config across every MCPServer.

**Size**: L
**Dependencies**: None
**Labels**: `operator`, `api`, `oidc`

## Context

OIDC configurations are defined inline in multiple CRDs (MCPServer, MCPRemoteProxy, VirtualMCPServer). Platform teams managing 10+ MCPServer resources with the same OIDC provider must duplicate the full config in each resource. A single issuer URL change requires updating every resource.

This story extracts shared OIDC configuration into a dedicated reusable CRD following the pattern established by `MCPExternalAuthConfig`.

## Acceptance Criteria

- [x] `MCPOIDCConfig` CRD is defined with `inline`, `configMapRef`, and `kubernetesServiceAccount` variants
- [x] CEL validation enforces exactly one variant is set (no `type` discriminator field)
- [x] Status tracks `observedGeneration`, `configHash`, `referencingServers`, and standard `conditions`
- [x] Controller adds finalizer on first reconciliation
- [x] Controller tracks referencing workloads in `status.referencingServers`
- [x] Controller blocks deletion while references exist (finalizer pattern)
- [x] Controller cascades config changes to referencing workloads via annotation (`toolhive.stacklok.dev/oidcconfig-hash`)
- [x] Controller sets `Ready` condition based on config validation
- [x] Controller emits Warning event when duplicate audiences are detected across referencing workloads
- [x] CRD manifest is generated and included in `operator-crds` Helm chart
- [x] Unit tests cover type validation, controller reconciliation, finalizer logic, cascade, and audience uniqueness
- [x] Integration tests in `cmd/thv-operator/test-integration/` cover cross-CRD reference lifecycle, cascade reconciliation, and deletion protection

### LINKED ISSUE #4253: Workload CRD Config Reference Updates
**As a** platform engineer deploying MCP servers,
**I want** MCPServer, MCPRemoteProxy, and VirtualMCPServer to reference shared `MCPOIDCConfig` and `MCPTelemetryConfig` resources,
**so that** workloads use centralized configuration without duplication.

**Size**: XL
**Dependencies**: STORY-02 (MCPOIDCConfig), STORY-03 (MCPTelemetryConfig)
**Labels**: `operator`, `api`, `controller`

## Context

With MCPOIDCConfig and MCPTelemetryConfig CRDs available (STORY-02, STORY-03), workload CRDs need to be updated to reference them. Each workload can either reference a shared config CRD or specify inline configuration. Workload controllers must resolve references, merge per-server overrides, and fail closed if a referenced config is missing or not ready.

## Acceptance Criteria

- [x] MCPServer spec accepts `oidcConfigRef` pointing to an MCPOIDCConfig with per-server `audience` and `scopes` overrides
- [x] MCPServer spec accepts `telemetryConfigRef` pointing to an MCPTelemetryConfig with per-server `serviceName`
- [x] Inline config remains supported — CEL enforces `ref` and `inline` are mutually exclusive
- [x] MCPRemoteProxy spec accepts `oidcConfigRef` with same pattern
- [x] VirtualMCPServer spec accepts `oidcConfigRef` for incoming auth
- [x] MCPServer controller resolves config refs, merges per-server overrides, and passes resolved config to workload
- [x] MCPRemoteProxy controller resolves OIDC config refs
- [x] VirtualMCPServer controller resolves OIDC config ref
- [x] All workload controllers fail closed: if a referenced config CRD does not exist or is not `Ready`, the workload enters `Failed` phase with a descriptive condition
- [x] Config hash tracked in workload status for change detection
- [x] Unit tests cover ref resolution, inline fallback, fail-closed behavior, and override merging
- [x] Integration tests in `cmd/thv-operator/test-integration/` cover cross-CRD reference resolution for all workload types
- [x] Architecture documentation (`docs/arch/09-operator-architecture.md`) updated with new reference patterns
- [x] Helm chart RBAC examples updated for platform team / app team separation
- [x] Example manifests and migration guide provided

### Deferred / tracked separately

- **MCPRemoteProxy + VirtualMCPServer `telemetryConfigRef`**: Deferred — only MCPServer has telemetry config ref support currently
- **Default audience generation (`{kind}/{namespace}/{name}`)**: Deferred for later discussion
- **Status condition persistence bug**: Tracked in #4511

ChrisJBurns:
## Summary

- VirtualMCPServer previously only supported inline OIDC configuration (oidcConfig), requiring each server to duplicate shared provider settings. This wires up the shared MCPOIDCConfig resource pattern already established for MCPServer, allowing VirtualMCPServers to reference a shared MCPOIDCConfig via oidcConfigRef with per-server audience and scopes overrides.
- The MCPOIDCConfig controller now also tracks VirtualMCPServer references in ReferencingWorkloads and blocks deletion while referenced.
- This addresses the loose ends from #4248 (readiness probing via vMCP) and #4253 (token exchange via vMCP) by having VirtualMCPServer use the MCPOIDCConfig resource to configure OIDC through its existing vMCP config pipeline.

Closes #4253
Closes #4248

**Stacked on #4492** — merge that first.

## Type of change

- [x] New feature

## Test plan

- [x] Integration tests (envtest-based tests in `test-integration/mcp-oidc-config/`)

## Changes

| File | Change |
|------|--------|
| `api/v1alpha1/virtualmcpserver_types.go` | Add `oidcConfigRef` field to `IncomingAuthConfig` with CEL mutual-exclusivity validation; deprecate `oidcConfig`; add `OIDCConfigHash` to status |
| `controllers/virtualmcpserver_controller.go` | Add `handleOIDCConfig` for reference validation, hash tracking, and ReferencingWorkloads updates; add MCPOIDCConfig watch handler and RBAC |
| `controllers/virtualmcpserver_deployment.go` | Update `buildOIDCEnvVars`, `buildVolumesForVmcp`, and `validateSecretReferences` to handle MCPOIDCConfig inline client secrets and CA bundles; propagate errors |
| `controllers/mcpoidcconfig_controller.go` | Extend `findReferencingWorkloads` to include VirtualMCPServers; add VirtualMCPServer watch for reference tracking |
| `pkg/vmcpconfig/converter.go` | Add `OIDCConfigRef` resolution branch in `convertIncomingAuth` using `ResolveFromConfigRef`; add `mapResolvedOIDCToVmcpConfigFromRef` |
| `test-integration/mcp-oidc-config/suite_test.go` | Register VirtualMCPServer and MCPGroup controllers; add field indexers |
| `test-integration/mcp-oidc-config/mcpoidcconfig_virtualmcpserver_integration_test.go` | 8 integration test specs covering OIDCConfigRef validation, hash tracking, ReferencingWorkloads, deletion protection, and cross-server-type references |

Yes. VirtualMCPServer now supports `spec.incomingAuth.oidcConfigRef` to reference a shared MCPOIDCConfig resource for OIDC authentication. The inline `spec.incomingAuth.oidcConfig` field is deprecated and will be removed in v1beta1. The two fields are mutually exclusive.

- The converter's `mapResolvedOIDCToVmcpConfigFromRef` function is intentionally separate from `mapResolvedOIDCToVmcpConfig` because client secret detection differs: the MCPOIDCConfig path checks `MCPOIDCConfig.Spec.Inline.ClientSecretRef` rather than `OIDCConfigRef.Inline.ClientSecretRef`.
- The `handleOIDCConfig` method follows the same pattern as `MCPServerReconciler.handleOIDCConfig` but uses the `StatusManager` interface for batched status updates.
- Deployment builder functions (`buildVolumesForVmcp`, `buildOIDCEnvVars`, `buildEnvVarsForVmcp`) return errors instead of silently dropping MCPOIDCConfig fetch failures, unlike the MCPServer controller which uses the silent-swallow pattern.

### DIFF
--- a/cmd/thv-operator/controllers/virtualmcpserver_controller.go
+++ b/cmd/thv-operator/controllers/virtualmcpserver_controller.go
@@ -20,6 +20,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -118,6 +119,8 @@ type VirtualMCPServerReconciler struct {
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=create;get;list;watch
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=create;delete;get;list;patch;update;watch
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpoidcconfigs,verbs=get;list;watch
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpoidcconfigs/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=embeddingservers,verbs=get;list;watch
 // +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=embeddingservers/status,verbs=get
 
@@ -149,6 +152,15 @@ func (r *VirtualMCPServerReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, err
 	} else if !cont {
 		return ctrl.Result{}, nil
+	}
+
+	// Validate MCPOIDCConfigRef if referenced (before resource creation).
+	// handleOIDCConfig is a no-op when OIDCConfigRef is nil.
+	if err := r.handleOIDCConfig(ctx, vmcp, statusManager); err != nil {
+		if applyErr := r.applyStatusUpdates(ctx, vmcp, statusManager); applyErr != nil {
+			ctxLogger.Error(applyErr, "Failed to apply status updates after MCPOIDCConfig validation error")
+		}
+		return ctrl.Result{}, err
 	}
 
 	// Ensure all resources
@@ -1268,7 +1280,10 @@ func (r *VirtualMCPServerReconciler) containerNeedsUpdate(
 	}
 
 	// Check if environment variables have changed
-	expectedEnv := r.buildEnvVarsForVmcp(ctx, vmcp, typedWorkloads)
+	expectedEnv, err := r.buildEnvVarsForVmcp(ctx, vmcp, typedWorkloads)
+	if err != nil {
+		return true // Trigger update to surface the error
+	}
 	if !reflect.DeepEqual(container.Env, expectedEnv) {
 		return true
 	}
@@ -2131,6 +2146,12 @@ func (r *VirtualMCPServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&mcpv1alpha1.EmbeddingServer{},
 			handler.EnqueueRequestsFromMapFunc(r.mapEmbeddingServerToVirtualMCPServer),
 		).
+		// Watch referenced MCPOIDCConfigs so that validity/hash changes
+		// trigger VirtualMCPServer reconciliation.
+		Watches(
+			&mcpv1alpha1.MCPOIDCConfig{},
+			handler.EnqueueRequestsFromMapFunc(r.mapOIDCConfigToVirtualMCPServer),
+		).
 		Complete(r)
 }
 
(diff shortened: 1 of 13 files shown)


cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -2703,3 +2724,151 @@ func generateHMACSecret() (string, error) {
...
+	vmcp *mcpv1alpha1.VirtualMCPServer,
+	statusManager virtualmcpserverstatus.StatusManager,
+) error {
+	ctxLogger := log.FromContext(ctx)
+
+	if vmcp.Spec.IncomingAuth == nil || vmcp.Spec.IncomingAuth.OIDCConfigRef == nil {
+		// No MCPOIDCConfig referenced, clear any stored hash
+		if vmcp.Status.OIDCConfigHash != "" {
+			vmcp.Status.OIDCConfigHash = ""
+			if err := r.Status().Update(ctx, vmcp); err != nil {
```
jerm-dro, line 2742: **question:** `handleOIDCConfig` updates `OIDCConfigHash` via direct `r.Status().Update(ctx, vmcp)` calls (here and at the hash-update site below), but the rest of the reconciler buffers status changes through `statusManager` and flushes them in `applyStatusUpdates`, which fetches a fresh copy before writing. If the direct update succeeds and then `applyStatusUpdates` fetches the updated object, this works fine. But if there's a version conflict between the two writes, the hash update could be lost (the requeue would recover, but it's an extra round-trip). Was this intentional to keep things simple, or would it be cleaner to track `OIDCConfigHash` through the statusManager like the other status fields?

cmd/thv-operator/pkg/vmcpconfig/converter.go
```diff
@@ -177,8 +177,31 @@ func (c *Converter) convertIncomingAuth(
...
+		if err != nil {
+			ctxLogger.Error(err, "failed to resolve OIDC config from MCPOIDCConfig",
+				"vmcp", vmcp.Name,
+				"namespace", vmcp.Namespace,
+				"oidcConfigRef", vmcp.Spec.IncomingAuth.OIDCConfigRef.Name)
+			return nil, fmt.Errorf("OIDC resolution failed from MCPOIDCConfig %q: %w",
+				vmcp.Spec.IncomingAuth.OIDCConfigRef.Name, err)
+		}
+		if resolvedConfig != nil {
+			incoming.OIDC = mapResolvedOIDCToVmcpConfigFromRef(resolvedConfig, oidcCfg)
```
jerm-dro, line 201: **suggestion:** These two branches (OIDCConfigRef path and legacy OIDCConfig path) both resolve OIDC config and then mutate `incoming.OIDC`. This would be simpler as an immutable assignment with a helper — e.g.:

```go
oidc, err := c.resolveOIDCConfig(ctx, vmcp)
if err != nil {
    return nil, err
}

incoming := &vmcpconfig.IncomingAuthConfig{
    Type: vmcp.Spec.IncomingAuth.Type,
    OIDC: oidc,
}
```

That collapses the two branches into a single function that returns `(*vmcpconfig.OIDCConfig, error)` and makes `incoming` immutable after construction.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
ChrisJBurns: This is a valid observation. The `MCPServer` controller has the same pattern (direct `r.Status().Update()` for the hash), so we followed it for consistency. But the reviewer is right that the `VirtualMCPServer` reconciler is different - it uses `statusManager` for batched updates, which the `MCPServer` reconciler doesn't. So the mixed approach is a real mismatch specific to `VirtualMCPServer`.

  The clean fix would be to track `OIDCConfigHash` through the `statusManager`. However, the `StatusManager` interface currently has typed setters (`SetPhase`, `SetURL`, etc.) but no generic "set status field" method for arbitrary fields like `OIDCConfigHash`. We'd need to either add a `SetOIDCConfigHash`(hash string) method to the interface, or move the hash tracking to a different point in the flow.

 Perhaps a pragmatic middle ground is the direct update happens early in `Reconcile` (before `ensureAllResources`), so the `statusManager` flush happens after on a potentially stale `ResourceVersion`. But `applyStatusUpdates` re-fetches the object before writing, so it would pick up the new `ResourceVersion`. So in practice the conflict scenario is unlikely. Still, it's worth cleaning up.

I can raise a ticket on this if you want @jerm-dro

cmd/thv-operator/pkg/vmcpconfig/converter.go
ChrisJBurns: Yep. I'm happy to get this merged into a subsequent PR to keep changes small and frequent

cmd/thv-operator/controllers/virtualmcpserver_controller.go
ChrisJBurns: Ticket raised https://github.com/stacklok/toolhive/issues/4504

PR #4497, opened 2026-04-02
refactor(auth): extract transientRefresher and fix flaky singleflight test

yrobla:
## Summary

`TestMonitoredTokenSource_SingleflightDeduplicatesConcurrentRetries` was non-deterministic: after a barrier released all goroutines toward `refreshGroup.Do`, the singleflight leader could complete its retry and return before follower goroutines reached `Do` — causing each follower to start an independent retry loop and fail the call-count assertion.

The fix also addresses a design smell: the singleflight/retry logic and its test hooks were embedded in `MonitoredTokenSource`, making that struct own too many concerns.

**Changes:**

- Extract `transientRefresher` — a small, focused type that owns the `singleflight.Group`, exponential-backoff retry, and the two synchronisation hooks (`beforeEntry`, `afterEntry`). Nil in production; set in tests only.
- `MonitoredTokenSource` now holds a `*transientRefresher` and delegates transient-error handling to `refresher.Refresh()`. The three test hooks (`beforeEntry`, `afterEntry`, `newRetryBackOff`) are removed from the production struct.
- Rewrite the flaky test as `TestTransientRefresher_SingleflightDeduplicatesConcurrentRetries`, targeting `transientRefresher` directly with the two-phase barrier. No `MonitoredTokenSource` or status-updater mock needed. Key assertion simplifies to exactly 1 `tokenSource.Token()` call (only the singleflight leader retries).
- Fix a misleading comment that implied reading `callCount` inside `tokenFn` would require re-acquiring the mock's mutex and deadlock (it would not — the mutex is already held).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/auth/monitored_token_source.go` | Extract `transientRefresher` with `Refresh`, `retry`, `getBackOff`; `MonitoredTokenSource` delegates to it, test hooks removed |
| `pkg/auth/monitored_token_source_test.go` | Rewrite singleflight test targeting `transientRefresher` directly; update remaining tests to use `ats.refresher.newBackOff`; fix misleading comment; remove unused `sync/atomic` import |

No.

### DIFF
--- a/pkg/auth/monitored_token_source.go
+++ b/pkg/auth/monitored_token_source.go
@@ -120,6 +120,86 @@ type StatusUpdater interface {
 	SetWorkloadStatus(ctx context.Context, workloadName string, status runtime.WorkloadStatus, reason string) error
 }
 
+// transientRefresher deduplicates concurrent token fetches during transient
+// network failures and retries with exponential backoff. It is owned by
+// MonitoredTokenSource and can be tested in isolation.
+type transientRefresher struct {
+	group    singleflight.Group
+	source   oauth2.TokenSource
+	workload string
+
+	// newBackOff is a factory for the backoff used during retries.
+	// Nil in production; overridable in tests for fast execution.
+	newBackOff func() backoff.BackOff
+
+	// beforeEntry and afterEntry are nil in production. Tests set them to
+	// synchronise goroutines so that the singleflight group is fully formed
+	// before the leader's retry returns.
+	beforeEntry func()
+	afterEntry  func()
+}
+
+// Refresh deduplicates concurrent callers via singleflight and retries the
+// underlying token source with exponential backoff until the context is
+// cancelled or a non-transient error is returned.
+func (r *transientRefresher) Refresh(ctx context.Context, origErr error) (*oauth2.Token, error) {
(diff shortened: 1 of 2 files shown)


pkg/auth/monitored_token_source.go
```diff
@@ -120,6 +120,86 @@ type StatusUpdater interface {
...
+	newBackOff func() backoff.BackOff
+
+	// beforeEntry and afterEntry are nil in production. Tests set them to
+	// synchronise goroutines so that the singleflight group is fully formed
+	// before the leader's retry returns.
+	beforeEntry func()
+	afterEntry  func()
+}
+
+// Refresh deduplicates concurrent callers via singleflight and retries the
```
jerm-dro, line 142: **nitpick:** The `beforeEntry`/`afterEntry` hooks are still test-only fields on a production struct. The extraction improves things significantly — `transientRefresher` is much smaller and more focused than `MonitoredTokenSource` was — so this is fine for now. Just noting the pattern persists.

PR #4499, opened 2026-04-02
Add thv skill builds command to list locally-built OCI artifacts

samuv:
## Summary

`thv skill build` packages a skill directory into a local OCI artifact, but there was no way to see what had been built. This adds a `thv skill builds` command (and backing API endpoint) to list locally-built OCI skill artifacts stored in the local OCI Image Layout.

* New `LocalBuild` type carrying `tag`, `digest`, `name`, `description`, and `version`
* New `ListBuilds` method on `SkillService` reading tags from the local OCI store, filtering to skill artifacts by `ArtifactType == "dev.toolhive.skills.v1"`, and best-effort enriching them with skill config metadata from OCI image labels
* New `GET /api/v1beta/skills/builds` API endpoint returning `{"builds": [...]}`
* New `thv skill builds` CLI command with `--format json|text` support
* Fix `task docs` to exclude `pkg/mod` and `pkg/sumdb` from swag's file scan

## Type of change

* New feature

## Changes

| File | Change |
| ---- | ------ |
| pkg/skills/options.go | Add LocalBuild type |
| pkg/skills/service.go | Add ListBuilds to SkillService interface |
| pkg/skills/skillsvc/skillsvc.go | Implement ListBuilds with ArtifactType filter and best-effort metadata |
| pkg/skills/skillsvc/skillsvc_test.go | 7 unit tests for ListBuilds including ArtifactType filtering |
| pkg/skills/mocks/mock_service.go | Regenerated mock |
| pkg/api/v1/skills.go + skills_types.go | GET /api/v1beta/skills/builds handler and response type |
| pkg/api/v1/skills_test.go | 3 router tests for the new endpoint |
| pkg/skills/client/client.go + dto.go | ListBuilds HTTP client method |
| cmd/thv/app/skill_builds.go | thv skill builds CLI command |
| docs/server/ | Regenerated OpenAPI spec |
| Taskfile.yml | Add --exclude pkg/mod,pkg/sumdb to swag init |

Yes. Users can now run `thv skill builds` to see all locally-built skill artifacts with their tag, digest, name, and version. The same data is available via `GET /api/v1beta/skills/builds`.

`ListBuilds` filters the tag list by `ArtifactType == "dev.toolhive.skills.v1"` (the same value the packager sets at build time) so that remotely-pulled or non-skill artifacts sharing the local OCI store are not surfaced. Metadata extraction (`name`, `description`, `version`) is still best-effort: if the OCI config labels can't be parsed the artifact is included with those fields empty.

### DIFF
--- a/pkg/skills/skillsvc/skillsvc.go
+++ b/pkg/skills/skillsvc/skillsvc.go
@@ -523,6 +523,15 @@ func (s *service) ListBuilds(ctx context.Context) ([]skills.LocalBuild, error) {
 			continue
 		}
 
+		isSkill, typeErr := s.isSkillArtifact(ctx, d)
+		if typeErr != nil {
+			slog.Debug("failed to check artifact type in local OCI store", "tag", tag, "error", typeErr)
+			continue
+		}
+		if !isSkill {
+			continue
+		}
+
 		build := skills.LocalBuild{
 			Tag:    tag,
 			Digest: d.String(),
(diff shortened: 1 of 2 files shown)


pkg/skills/skillsvc/skillsvc.go
```diff
@@ -501,6 +501,48 @@ func (s *service) Push(ctx context.Context, opts skills.PushOptions) error {
...
+	}
+
+	builds := make([]skills.LocalBuild, 0, len(tags))
+	for _, tag := range tags {
+		d, resolveErr := s.ociStore.Resolve(ctx, tag)
+		if resolveErr != nil {
+			slog.Debug("failed to resolve tag in local OCI store", "tag", tag, "error", resolveErr)
+			continue
+		}
+
```
JAORMX, line 525: `ListTags` returns *every* tag in the store's `index.json`. It doesn't filter by content type at all. The same store holds artifacts from both `thv skill build` (local builds) and `thv skill install` (remote pulls via `registry.Pull`, which tags with the short ref + the full OCI reference). So this method will also return installed/pulled skills, not just locally-built ones.

We already set `ArtifactType: "dev.toolhive.skills.v1"` on both manifests and indexes at build time (see `packager.go:529` and `packager.go:576`). After resolving each tag, we should fetch the manifest/index and check that `ArtifactType == ArtifactTypeSkill` before including it in the results. That way we're using the OCI-native mechanism we already have, and this stays correct even if the store's scope changes later.

samuv: Fixed in the latest commit.

ListBuilds now calls `isSkillArtifact` after resolving each tag, which fetches the top-level index or manifest and checks ArtifactType == "dev.toolhive.skills.v1" before including the entry in results. Tags without the skill artifact type (e.g. remotely-pulled skills or anything else sharing the store) are silently skipped.

Added a test case "non-skill artifact is excluded" that stores one real skill artifact and one index with a different artifactType in the same store, and asserts only the skill artifact is returned.

Issue #4466, opened 2026-03-31
Eliminate self-referential HTTP fetch for JWKS when using embedded auth server


jerm-dro:
## User Story

As an embedded auth server operator, I shouldn't have to configure `jwksUrl` pointing my MCPServer back at itself, or enable scary-sounding flags like `insecureAllowHTTP` or `jwksAllowPrivateIP`, just to get token validation working. The embedded auth server and the token validator are running in the same process — why is there an HTTP round-trip at all?

## The Pain

Today, when you enable the embedded auth server on an MCPServer, OAuth flows work — tokens are issued successfully. But token *validation* silently fails because the token validator tries to fetch JWKS over HTTP from the proxy's own endpoint.

To make it work, you must set either `jwksAllowPrivateIP: true` (if the issuer resolves to a private IP) or `insecureAllowHTTP: true` (if the internal service URL is `http://`), or both. Neither is recommended for production, and the failures are silent — requiring deep debugging to diagnose.

The fundamental issue is that the embedded auth server's `KeyProvider` already holds the public keys in memory. The token validator doesn't need to fetch them over HTTP. The two components just don't talk to each other directly.

## What Should Happen

When embedded auth is enabled, token validation should use the in-process `KeyProvider` directly. No `jwksUrl`, no HTTP fetch, no flags. The config should be as simple as:

```yaml
oidcConfig:
  type: inline
  inline:
    issuer: https://example.com
    jwksSource: embedded
```

Or better yet, this should be the automatic default when the embedded auth server is active.

## Affected Files

- `pkg/networking/http_client.go` — `ValidatingTransport.RoundTrip`
- `pkg/auth/token.go` — `ensureJWKSRegistered`, `getKeyFromJWKS`
- `pkg/runner/runner.go` — runner wiring (pass `KeyProvider` to token validator)
- `cmd/thv-operator/api/v1alpha1/mcpserver_types.go` — `InlineOIDCConfig` struct

## Current Workarounds

**Using localhost (only `insecureAllowHTTP` needed):**
```yaml
oidcConfig:
  type: inline
  inline:
    issuer: https://example.com
    jwksUrl: http://localhost:8080/.well-known/jwks.json
    insecureAllowHTTP: true
```

**Using the Kubernetes service name (`insecureAllowHTTP` + `jwksAllowPrivateIP` needed):**
```yaml
oidcConfig:
  type: inline
  inline:
    issuer: https://example.com
[...]

PR #4502, opened 2026-04-02
Resolve JWKS keys in-process for embedded auth server (MCP server)

tgrunnagle:
## Summary

When the embedded auth server is enabled, token validation currently fails silently because the token validator fetches JWKS keys over HTTP from the proxy's own endpoint. This self-referential HTTP call requires operators to set `insecureAllowHTTP` and/or `jwksAllowPrivateIP` flags — insecure workarounds that are difficult to debug when missing.

This PR eliminates the self-referential HTTP fetch by wiring the embedded auth server's `KeyProvider` directly into the token validator. When both components run in the same process, JWKS keys are resolved in-memory with a graceful fallback to HTTP for cases where the local provider cannot satisfy the request.

Note: this only addresses the issue for the runner and proxy runner - vMCP wiring will come in a separate PR.

Related to #4466

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/auth/token.go` | Add `keyProvider` field to `TokenValidator`, `WithKeyProvider` option, `getKeyFromLocalProvider` method that resolves keys in-process before falling back to HTTP, and relax the JWKS URL requirement when a local provider is present |
| `pkg/auth/middleware.go` | Pass `KeyProvider` from runner into token validator options during middleware creation |
| `pkg/authserver/runner/embeddedauthserver.go` | Store and expose `keyProvider` from the auth server's key material |
| `pkg/runner/runner.go` | Capture `KeyProvider` from the embedded auth server during startup and expose it via `GetKeyProvider()` |
| `pkg/transport/types/transport.go` | Extend `MiddlewareRunner` interface with `GetKeyProvider()` |
| `pkg/transport/types/mocks/mock_transport.go` | Regenerated mock for updated `MiddlewareRunner` interface |
| `pkg/auth/token_test.go` | Tests for `getKeyFromLocalProvider`: kid match, kid miss (fallback), provider error (fallback), unsupported signing method, missing kid |
| `pkg/authserver/runner/embeddedauthserver_test.go` | Test that `KeyProvider()` returns a functional provider after construction |
| `pkg/auth/middleware_test.go` | Updated existing middleware tests for new `GetKeyProvider()` expectation |

Yes. Operators using the embedded auth server no longer need to configure `jwksUrl`, `insecureAllowHTTP`, or `jwksAllowPrivateIP` for token validation to work. The embedded auth server's signing keys are resolved in-process automatically.

- `getKeyFromLocalProvider` returns `(nil, nil)` to signal "fall back to HTTP" — this keeps the existing HTTP-based JWKS path untouched as a fallback and avoids breaking external OIDC providers.
- The `resolveClientSecret` helper was extracted as a standalone function during the refactor; the logic is unchanged.
- The `ErrMissingIssuerAndJWKSURL` validation is now relaxed when a `keyProvider` is present, since keys can be resolved without a JWKS URL.


pkg/auth/token.go
```diff
@@ -372,6 +373,12 @@ type TokenValidator struct {
 	// nil means no enrichment (no embedded auth server).
 	upstreamTokenReader upstreamtoken.TokenReader
 
+	// keyProvider provides in-process JWKS key lookups from the embedded auth
+	// server's KeyProvider. When set, getKeyFromJWKS resolves keys locally
+	// before falling back to HTTP. Eliminates self-referential HTTP calls.
+	// nil when no embedded auth server is configured.
+	keyProvider keys.KeyProvider
```
jerm-dro: **blocker:** `KeyProvider` exposes both `SigningKey()` (private key) and `PublicKeys()`, but the token validator only ever calls `PublicKeys()`. Accepting the full `KeyProvider` interface leaks private key access to a component that should only verify, never sign. We want to remain intentional about which components have access to private key material.

Please introduce a narrower interface:

```go
type PublicKeyProvider interface {
    PublicKeys(ctx context.Context) ([]*keys.PublicKeyData, error)
}
```

`KeyProvider` already satisfies this implicitly, so the auth server needs no changes — just the field type here and the `MiddlewareRunner.GetKeyProvider()` return type.

Issue #4491, opened 2026-04-01
Migrate ReferencingServers to ReferencingWorkloads on remaining CRDs


ChrisJBurns:
## Summary

RFC-0023 specifies a structured `referencingWorkloads` list with `kind` and `name` fields on shared configuration CRD status, rather than plain `[]string`. PR #4483 migrated MCPOIDCConfig to the new `WorkloadReference` type. The same migration is needed for the remaining three CRDs.

## CRDs to migrate

| CRD | File | Current field |
|-----|------|---------------|
| MCPToolConfig | `cmd/thv-operator/api/v1alpha1/toolconfig_types.go:80-83` | `ReferencingServers []string` |
| MCPExternalAuthConfig | `cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go:715-718` | `ReferencingServers []string` |
| MCPTelemetryConfig | `cmd/thv-operator/api/v1alpha1/mcptelemetryconfig_types.go:62-64` | `ReferencingServers []string` |

## Target state

Each CRD should use the `WorkloadReference` type (already defined in `mcpoidcconfig_types.go`):

```go
type WorkloadReference struct {
    Kind string `json:"kind"` // MCPServer, VirtualMCPServer, MCPRemoteProxy
    Name string `json:"name"`
}
```

Field rename: `ReferencingServers []string` → `ReferencingWorkloads []WorkloadReference`

## Controllers to update

For each CRD, the corresponding controller builds and compares the referencing list:

- **ToolConfigReconciler** (`toolconfig_controller.go`) — `updateReferencingServers()` method
- **MCPExternalAuthConfigReconciler** (`mcpexternalauthconfig_controller.go`) — `updateReferencingServers()` method  
- **MCPTelemetryConfigReconciler** — equivalent method

Each needs to build `WorkloadReference` structs with explicit `Kind` instead of plain server name strings.

## Pattern to follow

See the MCPOIDCConfig migration in PR #4483 for the exact pattern:
- `findReferencingServers` → `findReferencingWorkloads` returning `[]WorkloadReference`
- Watch handlers compare `ref.Kind == "MCPServer" && ref.Name == server.Name` instead of `ref == server.Name`
- Deletion protection message uses `"ReferencedByWorkloads"` reason

## Notes

- This is a breaking change to the CRD status schema (status fields are not part of the API contract for clients, but tooling that reads status will need updates)
[...]

PR #4506, opened 2026-04-02
Replace ReferencingServers with ReferencingWorkloads on MCPToolConfig

ChrisJBurns:
## Summary

Migrates MCPToolConfig from `ReferencingServers []string` to `ReferencingWorkloads []WorkloadReference`, aligning it with the structured workload reference pattern established on MCPOIDCConfig in #4492. This enables the status field to distinguish between different workload kinds (MCPServer, VirtualMCPServer, MCPRemoteProxy) rather than assuming all referencing workloads are MCPServers.

Partial fix for #4491

## Type of change

- [x] Enhancement to an existing feature

## Changes

| File | Changes |
|------|---------|
| `toolconfig_types.go` | Replace `ReferencingServers []string` with `ReferencingWorkloads []WorkloadReference`, update printer column |
| `toolconfig_controller.go` | Add `findReferencingWorkloads()`, update reconcile/deletion logic to use structured refs with `DeletionBlocked` condition |
| `toolconfig_controller_test.go` | Update assertions to use `WorkloadReference{Kind, Name}` structs |
| `toolconfig_controller_edge_cases_test.go` | Update assertions to use `WorkloadReference{Kind, Name}` structs |
| `mcptoolconfig_controller_integration_test.go` | Update assertions to use `WorkloadReference{Kind, Name}` structs |
| Generated files | Regenerated deepcopy, CRD YAML manifests |

## Test plan

- [x] Unit tests updated and passing (`go test ./cmd/thv-operator/controllers/... -run ToolConfig`)
- [x] Integration tests updated
- [x] CRD manifests regenerated via `task gen`

Follows the exact pattern from #4492 (MCPOIDCConfig migration). MCPExternalAuthConfig migration is in a separate PR. MCPTelemetryConfig will follow separately.

### DIFF
--- a/cmd/thv-operator/controllers/toolconfig_controller.go
+++ b/cmd/thv-operator/controllers/toolconfig_controller.go
@@ -98,6 +98,15 @@ func (r *ToolConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return r.handleConfigHashChange(ctx, toolConfig, configHash)
 	}
 
+	// Refresh ReferencingWorkloads list
+	referencingWorkloads, err := r.findReferencingWorkloads(ctx, toolConfig)
+	if err != nil {
+		logger.Error(err, "Failed to find referencing workloads")
+	} else if !slices.Equal(toolConfig.Status.ReferencingWorkloads, referencingWorkloads) {
+		toolConfig.Status.ReferencingWorkloads = referencingWorkloads
+		conditionChanged = true
+	}
+
 	// Update condition if it changed (even without hash change)
 	if conditionChanged {
 		if err := r.Status().Update(ctx, toolConfig); err != nil {
@@ -106,9 +115,7 @@ func (r *ToolConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		}
 	}
 
-	// Even when hash hasn't changed, update referencing servers list.
-	// This ensures ReferencingServers is updated when MCPServers are created or deleted.
-	return r.updateReferencingServers(ctx, toolConfig)
+	return ctrl.Result{}, nil
 }
 
 // handleConfigHashChange handles the logic when the config hash changes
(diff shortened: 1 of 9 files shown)


cmd/thv-operator/controllers/toolconfig_controller.go
```diff
@@ -134,13 +141,12 @@ func (r *ToolConfigReconciler) handleConfigHashChange(
...
-	serverNames := make([]string, 0, len(referencingServers))
+	// Update the status with the list of referencing workloads
+	refs := make([]mcpv1alpha1.WorkloadReference, 0, len(referencingServers))
 	for _, server := range referencingServers {
-		serverNames = append(serverNames, server.Name)
+		refs = append(refs, mcpv1alpha1.WorkloadReference{Kind: "MCPServer", Name: server.Name})
 	}
-	slices.Sort(serverNames)
-	toolConfig.Status.ReferencingServers = serverNames
+	toolConfig.Status.ReferencingWorkloads = refs
```
jerm-dro, line 150: **suggestion:** The old code sorted `serverNames` before comparison, but the new `refs` slice built in `handleConfigHashChange` has no ordering guarantee — `findReferencingMCPServers` returns servers in list order, which can vary between API calls. Meanwhile, the reconcile path uses `slices.Equal` (order-sensitive) to decide whether a status update is needed.

This means two reconcile runs that discover the same servers in different orders will trigger an unnecessary status write. Not a correctness issue, but it adds avoidable API server churn.

Consider sorting `refs` by name after construction (matching what the MCPOIDCConfig controller should probably also do), or switching to an order-insensitive comparison like building a set.

This is a pre-existing pattern-level issue (MCPOIDCConfig has the same gap), so fine to address separately if you prefer.

jerm-dro (approved):

cmd/thv-operator/controllers/toolconfig_controller.go
ChrisJBurns: Fixed in https://github.com/stacklok/toolhive/pull/4520/changes

PR #4507, opened 2026-04-02
Replace ReferencingServers with ReferencingWorkloads on MCPExternalAuthConfig

ChrisJBurns:
## Summary

Migrates MCPExternalAuthConfig from `ReferencingServers []string` to `ReferencingWorkloads []WorkloadReference`, aligning it with the structured workload reference pattern established on MCPOIDCConfig in #4492. This enables the status field to distinguish between different workload kinds (MCPServer, VirtualMCPServer, MCPRemoteProxy) rather than assuming all referencing workloads are MCPServers.

Partial fix for #4491

## Type of change

- [x] Enhancement to an existing feature

## Changes

| File | Changes |
|------|---------|
| `mcpexternalauthconfig_types.go` | Replace `ReferencingServers []string` with `ReferencingWorkloads []WorkloadReference`, update printer column |
| `mcpexternalauthconfig_controller.go` | Add `findReferencingWorkloads()`, update reconcile/deletion logic to use structured refs with `DeletionBlocked` condition |
| `mcpexternalauthconfig_controller_test.go` | Update assertions to use `WorkloadReference{Kind, Name}` structs |
| `mcpexternalauthconfig_controller_integration_test.go` | Update assertions to use `WorkloadReference{Kind, Name}` structs |
| Generated files | Regenerated deepcopy, CRD YAML manifests |

## Test plan

- [x] Unit tests updated and passing (`go test ./cmd/thv-operator/controllers/... -run MCPExternalAuthConfig`)
- [x] Integration tests updated
- [x] CRD manifests regenerated via `task gen`

Follows the exact pattern from #4492 (MCPOIDCConfig migration). MCPToolConfig migration is in a separate PR. MCPTelemetryConfig will follow separately.

### DIFF
--- a/cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
+++ b/cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
@@ -123,9 +123,9 @@ func (r *MCPExternalAuthConfigReconciler) Reconcile(ctx context.Context, req ctr
 		}
 	}
 
-	// Even when hash hasn't changed, update referencing servers list.
-	// This ensures ReferencingServers is updated when MCPServers are created or deleted.
-	return r.updateReferencingServers(ctx, externalAuthConfig)
+	// Even when hash hasn't changed, update referencing workloads list.
+	// This ensures ReferencingWorkloads is updated when MCPServers are created or deleted.
+	return r.updateReferencingWorkloads(ctx, externalAuthConfig)
 }
 
 // calculateConfigHash calculates a hash of the MCPExternalAuthConfig spec using Kubernetes utilities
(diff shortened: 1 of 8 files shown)


cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
```diff
@@ -250,28 +265,75 @@ func (r *MCPExternalAuthConfigReconciler) findReferencingMCPServers(
...
+		if server.Spec.ExternalAuthConfigRef != nil && server.Spec.ExternalAuthConfigRef.Name == externalAuthConfig.Name {
+			refs = append(refs, mcpv1alpha1.WorkloadReference{Kind: "MCPServer", Name: server.Name})
+		}
+	}
+	return refs, nil
+}
+
 // SetupWithManager sets up the controller with the Manager.
+// Watches MCPServer changes to maintain accurate ReferencingWorkloads status.
 func (r *MCPExternalAuthConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
```
jerm-dro, line 290: **suggestion:** Now that we have four controllers (MCPOIDCConfig, MCPTelemetryConfig, MCPToolConfig, MCPExternalAuthConfig) with nearly identical implementations of `findReferencingWorkloads`, `handleDeletion` (DeletionBlocked + requeue), stale-ref watch handlers, and reconcile-time refresh logic — and a fifth coming with MCPRemoteProxy in #4509 — it's worth extracting shared helpers for this pattern.

We're already seeing inconsistencies creep in: this PR correctly sorts refs with `SortFunc` in `handleConfigHashChange`, but #4506 (MCPToolConfig) doesn't. This PR keeps `updateReferencingWorkloads` as a separate method, while #4506 inlines it. These divergences will compound as more workload kinds are added.

A generic helper or callback-based approach (e.g., `FindReferencingWorkloads(ctx, client, namespace, ...WorkloadScanner)` + a shared `BuildStaleRefWatchHandler`) would reduce each controller to a few lines of config-specific wiring. The deletion and refresh patterns are ~30 lines each that differ only by field names.

This may make sense to address as part of the shared config resource lifecycle work in [stacklok/toolhive-rfcs#65](https://github.com/stacklok/toolhive-rfcs/pull/65/changes) rather than as a standalone cleanup — but flagging it now since the cognitive load of keeping these controllers in sync is already showing cracks.

Not blocking this PR on it.

jerm-dro (approved):

cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
ChrisJBurns: Yeah, when I saw the similar code I made a note straight away to refactor this to avoid duplication

ChrisJBurns: Fixed in https://github.com/stacklok/toolhive/pull/4520/changes

PR #4515, opened 2026-04-03
Support stateless streamable-HTTP remote servers

### LINKED ISSUE #4514: Zero tools from remote servers implementing stateless streamable-HTTP
## Bug description

Remote MCP servers that implement the MCP streamable-HTTP transport in stateless mode (POST-only, no SSE) are unusable with ToolHive. The transparent proxy forwards all HTTP methods to the remote server, including GET. Stateless servers don't serve an SSE stream on GET, so the response is not a valid SSE stream and the client never discovers tools.

## Steps to reproduce

1. Run a remote MCP server that implements stateless streamable-HTTP. One concrete example: a Spring AI server configured with `WebMvcStatelessServerTransport`. You can verify a server is stateless by confirming that a POST to the MCP endpoint returns a JSON-RPC response, but a GET does not return a valid SSE stream.
2. Register it with ToolHive: `thv run --name myserver https://my-stateless-server.example.com/mcp`
3. Connect a client (e.g., Claude Code) and attempt to list tools

## Expected behavior

Tools are discoverable. The client can send JSON-RPC requests via POST and receive inline responses.

## Actual behavior

Zero tools appear. The client sends GET to initiate an SSE stream; the proxy forwards it to the stateless server. Because stateless servers don't serve an SSE stream on GET, the response is not a valid SSE stream and the client never discovers tools.

## Environment (if relevant)

- Reproducible on current main

## Additional context

The MCP streamable-HTTP spec defines two server modes: stateful (supports GET for SSE stream initiation) and stateless (POST-only, inline responses, no sessions). The transparent proxy currently assumes stateful behavior for all remote servers. A server operating in stateless mode needs the proxy to restrict incoming requests to POST and to use a POST-based health check instead of the default GET probe.

gkatz2:
## Summary

- ToolHive had no way to proxy stateless streamable-HTTP MCP servers — servers that accept JSON-RPC only via POST and never open an SSE stream. When proxied, these servers showed zero tools because the proxy forwarded GET requests that stateless servers don't handle, preventing clients from completing tool discovery.
- Adds a `--stateless` flag to `thv run` that: (1) returns 405 for GET, HEAD, and DELETE requests at the proxy instead of forwarding them, and (2) switches the health check to a POST-based JSON-RPC ping appropriate for servers without an SSE endpoint.

Fixes #4514

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Verified against a real stateless streamable-HTTP server: GET and HEAD requests to the proxy return 405 with `method not allowed: server is stateless (POST only)`; POST requests are forwarded to the remote server. E2E tests are not included — no stateless MCP server fixture exists in the test infrastructure.

## Changes

| File | Change |
|------|--------|
| `cmd/thv/app/run_flags.go` | Add `--stateless` CLI flag |
| `pkg/runner/config.go` | Add `Stateless bool` to `RunConfig` |
| `pkg/runner/config_builder.go` | Add `WithStateless()` builder option |
| `pkg/runner/runner.go` | Wire stateless config to HTTP transport; warn when flag used without `--url` |
| `pkg/transport/http.go` | Add `SetStateless()` setter and `buildProxyOptions()` helper |
| `pkg/transport/proxy/transparent/transparent_proxy.go` | Add `WithStateless()` option, method gate middleware, stateless pinger selection |
| `pkg/transport/proxy/transparent/pinger.go` | Add `StatelessMCPPinger` for POST-based health checks |
| `pkg/transport/proxy/transparent/pinger_test.go` | Unit tests for `StatelessMCPPinger` |
| `pkg/transport/proxy/transparent/method_gate_test.go` | Unit tests for `statelessMethodGate` |

Yes. Adds `--stateless` flag to `thv run`. When set, the proxy returns 405 for GET, HEAD, and DELETE requests instead of forwarding them, enabling use with MCP servers that implement streamable-HTTP stateless mode.

### DIFF
--- a/test/e2e/stateless_proxy_test.go
+++ b/test/e2e/stateless_proxy_test.go
@@ -1,0 +1,263 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e_test
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"strings"
+	"sync/atomic"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/stacklok/toolhive/test/e2e"
+)
+
+var _ = Describe("Stateless Proxy Mode", Label("proxy", "stateless", "streamable-http", "e2e"), Serial, func() {
+	var (
+		config     *e2e.TestConfig
+		serverName string
+		mockServer *statelessMockMCPServer
+	)
+
+	BeforeEach(func() {
+		config = e2e.NewTestConfig()
+		serverName = e2e.GenerateUniqueServerName("stateless")
+
+		err := e2e.CheckTHVBinaryAvailable(config)
+		Expect(err).ToNot(HaveOccurred(), "thv binary should be available")
+	})
+
+	AfterEach(func() {
+		if mockServer != nil {
+			mockServer.Stop()
+			mockServer = nil
+		}
+
+		if config.CleanupAfter {
+			err := e2e.StopAndRemoveMCPServer(config, serverName)
+			Expect(err).ToNot(HaveOccurred(), "Should be able to stop and remove server")
+		}
+	})
+
+	Describe("Method gating for stateless servers", func() {
+		Context("when --stateless flag is set on a remote server", func() {
+			It("should reject GET requests and forward POST requests", func() {
+				By("Starting a stateless mock MCP server")
+				var err error
+				mockServer, err = newStatelessMockMCPServer()
+				Expect(err).ToNot(HaveOccurred(), "Should be able to start mock server")
+
+				mockServerURL := mockServer.URL()
+				GinkgoWriter.Printf("Mock server started at: %s\n", mockServerURL)
+
+				By("Starting thv with --stateless flag")
+				thvCmd := exec.Command(config.THVBinary, "run",
+					"--name", serverName,
+					"--stateless",
+					mockServerURL+"/mcp")
+				thvCmd.Env = append(os.Environ(),
+					"TOOLHIVE_REMOTE_HEALTHCHECKS=true",
+				)
+				thvCmd.Stdout = GinkgoWriter
+				thvCmd.Stderr = GinkgoWriter
+
+				err = thvCmd.Start()
+				Expect(err).ToNot(HaveOccurred(), "Should be able to start thv")
+
+				thvPID := thvCmd.Process.Pid
+				GinkgoWriter.Printf("thv process started with PID: %d\n", thvPID)
+
+				defer func() {
+					if proc, err := os.FindProcess(thvPID); err == nil {
+						_ = proc.Kill()
+					}
+				}()
+
+				By("Waiting for thv to register as running")
+				err = e2e.WaitForMCPServer(config, serverName, 60*time.Second)
+				Expect(err).ToNot(HaveOccurred(), "Server should be running within 60 seconds")
+
+				By("Getting the proxy URL")
+				proxyURL, err := e2e.GetMCPServerURL(config, serverName)
+				Expect(err).ToNot(HaveOccurred(), "Should be able to get proxy URL")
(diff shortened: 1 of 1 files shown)


pkg/runner/runner.go
```diff
@@ -394,6 +400,16 @@ func (r *Runner) Run(ctx context.Context) error {
...
 
+		// Configure stateless mode if requested. Remote workloads always use
+		// HTTPTransport, so a failed cast here is a programming error.
+		if r.Config.Stateless {
+			httpT, ok := transportHandler.(*transport.HTTPTransport)
+			if !ok {
+				return fmt.Errorf("internal error: remote transport is not an HTTPTransport")
+			}
+			httpT.SetStateless(true)
+		}
```
jerm-dro: suggestion: `SetStateless` is called on the transport after construction but before `Start`. This works, but it's a mutation window — if calls get reordered, stateless mode silently doesn't apply. Consider passing it through as a transport constructor option (like `WithRemoteBasePath`) to make it impossible to miss. This follows the same setter pattern as `SetOnHealthCheckFailed`, so it's at least consistent with the current codebase — just noting the opportunity.

pkg/runner/runner.go
```diff
@@ -304,6 +304,12 @@ func (r *Runner) Run(ctx context.Context) error {
 	// Set up the transport
 	slog.Debug("setting up transport", "transport", r.Config.Transport)
 
+	// Warn if stateless mode is set for a non-remote workload — it has no effect.
+	if r.Config.Stateless && r.Config.RemoteURL == "" {
+		slog.Warn("--stateless has no effect for local container workloads; use it only with remote URLs",
+			"server", r.Config.BaseName)
+	}
```
jerm-dro: blocker: This is gated on `RemoteURL != ""`, but a local container exposing stateless streamable-HTTP (not stdio) would have the same problem with GET-based health checks and GET forwarding. If a user runs `thv run --transport streamable-http --stateless some-image`, the proxy would still forward GETs to the container. The flag should work for any stateless streamable-HTTP server, not just remote ones.

pkg/transport/proxy/transparent/pinger_test.go
jerm-dro, line 1: suggestion: The unit tests cover the pinger and method gate well, but an e2e test would give confidence this doesn't regress end-to-end. The existing e2e infrastructure already invokes `thv run` directly (e.g., `health_check_zombie_test.go`), so adding a test that stands up a minimal stateless HTTP server (POST-only, returns 405 on GET) and verifies the proxy returns 405 on GET and forwards POST correctly should be straightforward. Separately, adding a stateless mode to [yardstick](https://github.com/StacklokLabs/yardstick) would make it easier to validate this locally — not a requirement for this PR, but it would help since there's no stateless MCP server fixture in the test infrastructure today.

pkg/runner/runner.go
gkatz2: Fixed in 91094600. The stateless configuration now runs after the remote-URL block, so it applies to any `HTTPTransport` regardless of whether `RemoteURL` is set. Also updated the comments, flag text, and generated docs to remove the remote-only language.

pkg/runner/runner.go
gkatz2: Agreed about the mutation window. The setter pattern here matches `SetOnHealthCheckFailed`, `SetOnUnauthorizedResponse`, `SetTokenSource`, and `SetRemoteURL` — all called between construction and `Start()` in the same block. Moving one to a constructor option while leaving the others as setters would be inconsistent. I think the right approach is a broader refactor that converts all of these to constructor options at once, rather than doing it piecemeal.

If you like, I can do that refactor as part of this PR. Or we can leave that for a follow-up. Which would you prefer?

pkg/transport/proxy/transparent/pinger_test.go
gkatz2: Added in 8a7267c9. The test starts an in-process mock MCP server (POST-only, returns valid JSON-RPC), runs `thv run --stateless <mockURL>/mcp`, and verifies GET→405, POST→forwarded, DELETE→405.

jerm-dro:
Thanks for addressing the feedback — the remote-only fix, the docs cleanup, and the e2e test all look great.

PR #4524, opened 2026-04-03
Fix flaky VirtualMCPServer scaling e2e test

ChrisJBurns:
## Summary

The `VirtualMCPServer Horizontal Scaling` e2e test intermittently timed out (300s) waiting for the VirtualMCPServer to become Ready, failing with "All backends are unhealthy". The health monitor permanently reported `server returned 4xx for initialize POST`.

**Root cause:** two bugs working together:

1. **Wrong localhost fallback in K8s backend discovery** — when `MCPServer.Status.URL` hadn't been set yet by the operator (race during startup), `mcpServerToBackend` fell back to `http://127.0.0.1:8080/mcp`. Inside the vMCP pod, port 8080 is the controller-runtime metrics server, which returns 4xx to MCP POST requests.

2. **Stale backend in health monitor** — `UpdateBackends()` only handled adding/removing backends by ID. When the reconciler later updated the registry with the correct service URL, the health check goroutine kept using the captured stale localhost pointer forever.

**Fix:**
- Remove the localhost fallback for both MCPServer and MCPRemoteProxy — if `Status.URL` is empty the backend is skipped and the reconciler adds it once the URL is set
- Teach `UpdateBackends` to restart goroutines when a backend's URL or transport type changes (defense-in-depth)

## Type of change

## Changes

| File | Change |
|---|---|
| `pkg/vmcp/workloads/k8s.go` | Remove localhost URL fallback for MCPServer and MCPRemoteProxy when `Status.URL` is empty |
| `pkg/vmcp/health/monitor.go` | `UpdateBackends()` detects property changes (URL, transport) and restarts monitoring goroutines; add `backendChanged()` helper |
| `pkg/vmcp/workloads/k8s_test.go` | Tests for empty `Status.URL` → backend skipped |
| `pkg/vmcp/health/monitor_test.go` | Tests for `backendChanged()` and `UpdateBackends` property-change restart |

## Test plan

- [x] Existing unit tests pass (`go test ./pkg/vmcp/...`)
- [x] New unit tests cover the empty URL skip and property-change restart scenarios
- [x] Lint passes (only pre-existing goconst issue remains)

The flakiness depends on whether the MCPServer controller sets `Status.URL` before or after the vMCP discovers backends. Fast operator reconcile → correct URL → test passes. Slow reconcile → localhost fallback → permanent health check failure.

The localhost fallback was fundamentally wrong for K8s mode where vMCP runs in a separate pod from the backend. Removing it is safe because the backend reconciler (watch controller) will add backends to the registry once `Status.URL` is populated.

### DIFF
--- a/pkg/vmcp/health/monitor.go
+++ b/pkg/vmcp/health/monitor.go
@@ -323,22 +323,37 @@ func (m *Monitor) UpdateBackends(newBackends []vmcp.Backend) {
 	// This ensures GetHealthSummary sees new backends before their health checks complete
 	m.backends = newBackends
 
-	// Start monitoring for new backends
+	// Start monitoring for new or changed backends
 	for id, backend := range newBackendsMap {
-		if _, exists := oldBackends[id]; !exists {
-			slog.Info("starting health monitoring for new backend", "backend", backend.Name)
-			backendCopy := backend
-
-			// Circuit breaker will be lazily initialized on first health check
-
-			backendCtx, cancel := context.WithCancel(m.ctx) //nolint:gosec // G118 - cancel stored in m.activeChecks, called during Stop
-			m.activeChecks[id] = cancel
-			m.wg.Add(1)
-			// Clear the "removed" flag if this backend was previously removed
-			// This allows health check results to be recorded again
-			m.statusTracker.ClearRemovedFlag(id)
-			go m.monitorBackend(backendCtx, &backendCopy, false) // false = dynamically added backend
+		oldBackend, exists := oldBackends[id]
+		if exists && !backendChanged(oldBackend, backend) {
+			continue // Existing backend with no relevant changes
 		}
+
+		if exists {
+			// Backend properties changed (e.g., URL updated after operator reconcile).
+			// Cancel the old goroutine so a new one starts with the updated properties.
+			slog.Info("restarting health monitoring for changed backend",
+				"backend", backend.Name, "old_url", oldBackend.BaseURL, "new_url", backend.BaseURL)
+			if cancel, ok := m.activeChecks[id]; ok {
+				cancel()
+				delete(m.activeChecks, id)
+			}
+		} else {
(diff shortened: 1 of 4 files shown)


ChrisJBurns:
@claude Can you use the agents to review this PR. Please raise any critical/high issues, or medium issues that are non controversial. Ignore nit picks  or controversial style preference points. I want to know if this change is backwards compatible and doesn't regress anything and just makes the implementation more solid and reliable. Eventually conclude with a recommendation for approval/request for changes.

pkg/vmcp/health/monitor.go
```diff
@@ -323,22 +323,37 @@ func (m *Monitor) UpdateBackends(newBackends []vmcp.Backend) {
...
+		backendCtx, cancel := context.WithCancel(m.ctx) //nolint:gosec // G118 - cancel stored in m.activeChecks, called during Stop
+		m.activeChecks[id] = cancel
+		m.wg.Add(1)
+		// Clear the "removed" flag if this backend was previously removed
+		// This allows health check results to be recorded again
+		m.statusTracker.ClearRemovedFlag(id)
+		go m.monitorBackend(backendCtx, &backendCopy, false) // false = dynamically added backend
 	}
 
 	// Stop monitoring for removed backends and clean up their state
```
jerm-dro, line 359: suggestion: The `activeChecks` map currently holds raw `context.CancelFunc` values, and `UpdateBackends` directly manages goroutine lifecycle (cancel, delete, create context, add to waitgroup, launch goroutine). This couples the monitor to the goroutine mechanics. Consider introducing a `backendCheck` interface (or similar) that owns `Cancel()` and `Start()` for a single backend — `activeChecks` would hold these instead of bare cancel funcs. That would decouple the two systems, make the restart-on-property-change logic easier to test in isolation (you could verify cancel/start calls without needing a full monitor with mock clients and timing), and give a natural place to hang per-backend state like the circuit breaker.

PR #4539, opened 2026-04-03
Route OIDCConfigHash through StatusManager in VirtualMCPServer

### LINKED ISSUE #4504: VirtualMCPServer: track OIDCConfigHash through statusManager instead of direct status update
## Summary

In `VirtualMCPServerReconciler.handleOIDCConfig`, the `OIDCConfigHash` field is written directly via `r.Status().Update(ctx, vmcp)`, while all other status fields in the VirtualMCPServer reconciler are buffered through the `statusManager` and flushed in a single batched `applyStatusUpdates` call.

This creates a mixed-write pattern: the direct update succeeds with one ResourceVersion, then the batched flush re-fetches and writes again. In practice this works because `applyStatusUpdates` re-fetches before writing, but a version conflict between the two writes would cause the hash update to be lost (recovered on requeue, but adds an unnecessary round-trip).

Ref: https://github.com/stacklok/toolhive/pull/4493#discussion_r2670261879

## Current behavior

```go
// In handleOIDCConfig — direct write
vmcp.Status.OIDCConfigHash = oidcConfig.Status.ConfigHash
if err := r.Status().Update(ctx, vmcp); err != nil {
    return fmt.Errorf("failed to update MCPOIDCConfig hash in status: %w", err)
}
```

## Desired behavior

Track `OIDCConfigHash` through the `StatusManager` like all other status fields. This requires adding a `SetOIDCConfigHash(hash string)` method to the `StatusManager` interface and its implementation in the `StatusCollector`.

## Context

The MCPServer controller uses the same direct-update pattern for `OIDCConfigHash`, but MCPServer doesn't use a `statusManager` — it writes status directly throughout. The inconsistency is specific to VirtualMCPServer which adopted `statusManager` for batched updates.

ChrisJBurns:
## Summary

`handleOIDCConfig` wrote `OIDCConfigHash` directly via `r.Status().Update(ctx, vmcp)` while every other VirtualMCPServer status field used the batched `StatusManager`. This mixed-write pattern could produce spurious 409 Conflict errors when the original `vmcp` object's `resourceVersion` became stale after earlier `applyStatusUpdates` calls in the same reconciliation.

Adds `SetOIDCConfigHash` to the `StatusManager` interface and routes both the set and clear paths through it, eliminating all direct VirtualMCPServer status writes from `handleOIDCConfig`.

Fixes #4504

## Type of change

- [x] Bug fix

## Changes

| File | Change |
|------|--------|
| `types.go` | Add `SetOIDCConfigHash(hash string)` to `StatusManager` interface |
| `collector.go` | Add `oidcConfigHash *string` field, setter, `UpdateStatus` application logic, debug log |
| `virtualmcpserver_controller.go` | Replace 2 direct `r.Status().Update()` calls with `statusManager.SetOIDCConfigHash(...)` |
| `collector_test.go` | Add set + clear tests, update batched test |
| `mocks/mock_collector.go` | Regenerated via `go generate` |

## Test plan

- [x] Unit tests (`task test`) — all pass, including 2 new collector tests
- [x] Linting (`task lint`) — 0 issues
- [x] Existing integration tests unaffected (no VirtualMCPServer unit test exercises `handleOIDCConfig`)

- The `r.Status().Update(ctx, oidcConfig)` in `updateOIDCConfigReferencingWorkloads` is intentionally unchanged — it writes to the MCPOIDCConfig resource (different resource), not VirtualMCPServer.
- Empty string clears the hash (matching `SetURL`/`SetMessage` semantics). The `*string` pointer in the collector distinguishes "never called" from "called with empty string".
- After this fix, `handleOIDCConfig` makes zero direct VirtualMCPServer status writes on any path.

### DIFF
--- a/cmd/thv-operator/pkg/virtualmcpserverstatus/collector.go
+++ b/cmd/thv-operator/pkg/virtualmcpserverstatus/collector.go
@@ -25,6 +25,7 @@ type StatusCollector struct {
 	message            *string
 	url                *string
 	observedGeneration *int64
+	oidcConfigHash     *string
 	conditions         map[string]metav1.Condition
 	discoveredBackends []mcpv1alpha1.DiscoveredBackend
 }
@@ -69,6 +70,12 @@ func (s *StatusCollector) SetURL(url string) {
 // SetObservedGeneration sets the observed generation to be updated.
 func (s *StatusCollector) SetObservedGeneration(generation int64) {
 	s.observedGeneration = &generation
+	s.hasChanges = true
+}
+
+// SetOIDCConfigHash sets the OIDC config hash to be updated.
+func (s *StatusCollector) SetOIDCConfigHash(hash string) {
+	s.oidcConfigHash = &hash
 	s.hasChanges = true
 }
 
@@ -169,6 +176,11 @@ func (s *StatusCollector) UpdateStatus(ctx context.Context, vmcpStatus *mcpv1alp
 			vmcpStatus.ObservedGeneration = *s.observedGeneration
 		}
 
+		// Apply OIDC config hash change
+		if s.oidcConfigHash != nil {
+			vmcpStatus.OIDCConfigHash = *s.oidcConfigHash
+		}
+
 		// Apply condition changes
 		for _, condition := range s.conditions {
 			if condition.Status == "" {
@@ -195,6 +207,7 @@ func (s *StatusCollector) UpdateStatus(ctx context.Context, vmcpStatus *mcpv1alp
 		ctxLogger.V(1).Info("Batched status update applied",
 			"phase", s.phase,
 			"message", s.message,
+			"oidcConfigHash", s.oidcConfigHash,
 			"conditionsCount", len(s.conditions),
 			"discoveredBackendsCount", len(s.discoveredBackends))
 		return true
--- a/cmd/thv-operator/controllers/virtualmcpserver_controller.go
+++ b/cmd/thv-operator/controllers/virtualmcpserver_controller.go
@@ -2772,10 +2772,7 @@ func (r *VirtualMCPServerReconciler) handleOIDCConfig(
 	if vmcp.Spec.IncomingAuth == nil || vmcp.Spec.IncomingAuth.OIDCConfigRef == nil {
 		// No MCPOIDCConfig referenced, clear any stored hash
 		if vmcp.Status.OIDCConfigHash != "" {
-			vmcp.Status.OIDCConfigHash = ""
-			if err := r.Status().Update(ctx, vmcp); err != nil {
-				return fmt.Errorf("failed to clear MCPOIDCConfig hash from status: %w", err)
-			}
+			statusManager.SetOIDCConfigHash("")
 		}
 		return nil
 	}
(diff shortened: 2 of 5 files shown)


ChrisJBurns:
@claude Please review this PR with the agents. Only mention critical/high issues and non controversial medium issues. ignore issues around style and preference. give an approve or reject conclusion

ChrisJBurns:
## Agent Review of Findings

Three specialist agents (Kubernetes operator expert, Go code writer, code reviewer) independently traced the full reconcile flow to evaluate the two medium findings. All three reached the same conclusion: **both are non-issues that require no follow-up.**

---

### Finding 1: `UpdateStatus` has no reset after successful application

**Verdict: Non-issue.**

- `StatusCollector` is created fresh per reconciliation (line 146) as a local variable — no cross-reconciliation reuse.
- Every call to `applyStatusUpdates` is immediately followed by a `return`. The agents exhaustively traced all call sites (lines 245, 317, 325, 333, 342, 352, 161, 170, 176, 218) and confirmed **exactly one invocation executes per reconciliation** — there is no fall-through from one call site to another.
- Even hypothetically, double-application is fully idempotent — `applyStatusUpdates` re-fetches a fresh `latest` from the API server before applying, so overwriting the same values is a no-op at the data level.
- Adding a reset would actually **introduce risk**: a future `defer`-based refactoring could silently lose updates collected after a partial apply. The current monotonic-accumulate pattern is the safer default for collect-then-flush.

### Finding 2: Hash comparison reads stale `vmcp.Status.OIDCConfigHash`

**Verdict: Non-issue. The premise is incorrect.**

- The agents traced the exact control flow: `runValidations` (line 152) only calls `applyStatusUpdates` on error paths that **return early from `Reconcile`** before reaching `handleOIDCConfig` (line 160). The happy path (which is the only path that continues to `handleOIDCConfig`) makes **zero** calls to `applyStatusUpdates`.
- So when `handleOIDCConfig` reads `vmcp.Status.OIDCConfigHash` at line 2835, the value is exactly what was fetched from the API server at reconcile-start (line 135). No intervening write has occurred.
- Additionally, `applyStatusUpdates` writes to a freshly fetched `latest` object (line 268), never mutating the caller's `vmcp` — so even if someone added a new `applyStatusUpdates` call before `handleOIDCConfig`, the in-memory value would still be correct.
- Only `handleOIDCConfig` ever calls `SetOIDCConfigHash`, so no other step could have persisted a different hash value.

---

Both findings are artifacts of structural code reading (seeing a missing reset, seeing a potentially stale field) without tracing the actual control flow. The control flow makes both scenarios unreachable.

cmd/thv-operator/pkg/virtualmcpserverstatus/collector.go
```diff
@@ -72,6 +73,12 @@ func (s *StatusCollector) SetObservedGeneration(generation int64) {
 	s.hasChanges = true
 }
 
+// SetOIDCConfigHash sets the OIDC config hash to be updated.
+func (s *StatusCollector) SetOIDCConfigHash(hash string) {
+	s.oidcConfigHash = &hash
+	s.hasChanges = true
```
blkt, line 79: **question:** is there the need to set this back to `false` anywhere or is this simply done by recreating the struct?

ChrisJBurns: Yeah so this is the same concern the agents already evaluated. The answer:  it's handled by recreating the struct. `NewStatusManager(vmcp)` creates a fresh `StatusCollector` at the start of every Reconcile call (line 146 of the controller), so `hasChanges` starts as `false` each time. The struct is scoped to a single reconciliation and garbage-collected when Reconcile returns

PR #4563, opened 2026-04-04
fix(operator): rename BackendAuthType enum value to camelCase

### LINKED ISSUE #4542: Rename BackendAuthType external_auth_config_ref to camelCase
## Summary

The enum value `"external_auth_config_ref"` is the only snake_case enum value in the entire CRD codebase. All other type discriminator enums use camelCase (`tokenExchange`, `headerInjection`, `bearerToken`, `embeddedAuthServer`, `awsSts`, `upstreamInject`, `kubernetesServiceAccount`, etc.). This should be standardized to `"externalAuthConfigRef"` before the API stabilizes.

### 1. Constant and enum markers (`cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go`)

**Constant** (~line 340): change the string value from `"external_auth_config_ref"` to `"externalAuthConfigRef"`:

```go
// Before
BackendAuthTypeExternalAuthConfigRef = "external_auth_config_ref"

// After
BackendAuthTypeExternalAuthConfigRef = "externalAuthConfigRef"
```

**Kubebuilder Enum markers** (~lines 165 and 170): update the validation enum values:

```go
// Before
// +kubebuilder:validation:Enum=discovered;external_auth_config_ref

// After
// +kubebuilder:validation:Enum=discovered;externalAuthConfigRef
```

There are two Enum markers in this file that reference this value — update both.

**CEL validation rules** (~lines 494, 497, 509): search for `'external_auth_config_ref'` in CEL expressions and update to `'externalAuthConfigRef'`. These are in `XValidation` rules on the `BackendAuthConfig` struct.

### 2. Converter logic (`cmd/thv-operator/pkg/vmcpconfig/converter.go`)

~Lines 455-458: update any string comparisons or references to the old value. The code uses the Go constant `BackendAuthTypeExternalAuthConfigRef` so this should just work after step 1, but verify no hardcoded strings exist.

### 3. Unit tests

These files have string literals `"external_auth_config_ref"` in test fixtures that need updating to `"externalAuthConfigRef"`:

- `cmd/thv-operator/api/v1alpha1/virtualmcpserver_types_test.go` (~lines 279, 281)
- `cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig_test.go` (~lines 193, 198, 221, 223, 261)
- `cmd/thv-operator/controllers/virtualmcpserver_externalauth_test.go` (~lines 309, 419, 482, 525, 640, 678, 767, 769, 802)
- `cmd/thv-operator/controllers/virtualmcpserver_watch_test.go` (~lines 754, 784, 834, 1409, 1427, 1463, 1481, 1487)

### 4. E2E tests

- `test/e2e/thv-operator/virtualmcp/virtualmcp_external_auth_test.go` (~lines 284, 331, 683, 733)

### 5. Regenerate

After all source changes:

```bash
task gen          # Regenerates CRD manifests (YAML files update automatically)
task crdref-gen   # Regenerates CRD API docs
task lint-fix     # Fix any lint issues
task lint         # Verify clean
task test         # Unit tests pass
```

The CRD YAML manifests in `deploy/charts/operator-crds/` and docs in `docs/operator/` are auto-generated — do not edit them manually.

## Notes

- The Go constant name `BackendAuthTypeExternalAuthConfigRef` stays the same — only its string value changes.
- Existing VirtualMCPServer resources with `type: external_auth_config_ref` will be rejected after the CRD update. This is acceptable for a `v1alpha1` API.
- The `"discovered"` value is lowercase (not camelCase) but that's fine — it's a single word, consistent with other single-word values like `"inline"` and `"anonymous"`.

tmchow:
The enum value \`external_auth_config_ref\` was the only snake_case type discriminator in the CRD codebase. Every other value uses camelCase (\`tokenExchange\`, \`headerInjection\`, \`bearerToken\`, \`embeddedAuthServer\`, \`awsSts\`, etc.). Renamed to \`externalAuthConfigRef\` for consistency before the API moves past v1alpha1.

**Changes:**
- Renamed the constant string value in \`virtualmcpserver_types.go\`
- Updated kubebuilder validation Enum markers and CEL validation rules
- Updated all test fixtures across 5 test files
- Regenerated CRD manifests and API docs

Build, lint (0 issues), and compilation all pass.

Fixes #4542


cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go
```diff
@@ -337,7 +337,7 @@ const (
 	BackendAuthTypeDiscovered = "discovered"
 
 	// BackendAuthTypeExternalAuthConfigRef references an MCPExternalAuthConfig resource
-	BackendAuthTypeExternalAuthConfigRef = "external_auth_config_ref"
+	BackendAuthTypeExternalAuthConfigRef = "externalAuthConfigRef"
```
jerm-dro, line 344: **Blocker:** it looks like this is used as an enum value. Changing the value will cause existing CRDs to break. Please introduce a `DeprecatedBackendAuthTypeExternalAuthConfigRef = external_auth_config_ref` so that we can still handle the old CRDs correctly. We should log.Warn that this old value is deprecated whenever it's used.

ChrisJBurns: Was this something you were ok to release as a breaking change now, or something you wanted a gradual rollout of? @jerm-dro

jerm-dro: Is the cost of the deprecation approach low? If so, I'd prefer to deprecate instead of break.

tmchow: Added `DeprecatedBackendAuthTypeExternalAuthConfigRef` constant, updated `kubebuilder` enum to accept both values, updated validation switch. lmk if i misunderstood

jerm-dro: This looks like what I was imagining. I think you'll need to regenerate some stuff. We'll also want to log a warning about the deprecation when the deprecated value is used.

tmchow:
Ran `controller-gen` for both `operator-generate` and `operator-manifests` -- no changes produced, so the generated files are already in sync.

The deprecation warning is in `converter.go` and fires when the old `external_auth_config_ref` value is used. Using `log.Info` since controller-runtime's logr doesn't have a Warn level, but happy to switch to a different approach if you prefer.

jerm-dro (approved):

PR #4564, opened 2026-04-04
Phase 5: Dynamic Webhook Middleware Kubernetes Controller

### LINKED ISSUE #3401: Webhook Middleware Phase 5: Kubernetes CRD and controller integration
## Overview

Implement Kubernetes CRD and controller support for webhook middleware configuration. This enables declarative webhook configuration for MCP servers running in Kubernetes clusters.

**Depends on**: Phase 2 (Validating webhook), Phase 3 (Mutating webhook)

## Files to Create

| File | Purpose |
|------|---------|
| `cmd/thv-operator/api/v1alpha1/mcpwebhookconfig_types.go` | CRD type definitions |
| `cmd/thv-operator/controllers/mcpwebhookconfig_controller.go` | Controller implementation |
| `cmd/thv-operator/pkg/controllerutil/webhook.go` | Webhook config resolution helpers |
| `config/crd/bases/toolhive.stacklok.dev_mcpwebhookconfigs.yaml` | CRD manifest (generated) |

## Files to Modify

| File | Changes |
|------|---------|
| `cmd/thv-operator/api/v1alpha1/mcpserver_types.go` | Add `WebhookConfigRef` field |
| `cmd/thv-operator/controllers/mcpserver_controller.go` | Handle webhook config resolution, add watcher |
| `cmd/thv-operator/controllers/mcpserver_runconfig.go` | Add webhook config to RunConfig builder |

## CRD Definition

```go
// cmd/thv-operator/api/v1alpha1/mcpwebhookconfig_types.go

type MCPWebhookConfigSpec struct {
    // Validating webhooks called to approve/deny requests
    Validating []WebhookSpec `json:"validating,omitempty"`
    
    // Mutating webhooks called to transform requests
    Mutating []WebhookSpec `json:"mutating,omitempty"`
}

type WebhookSpec struct {
    // Name is a unique identifier for this webhook
    Name string `json:"name"`
    
    // URL is the webhook endpoint (must be HTTPS)
    URL string `json:"url"`
    
    // Timeout for webhook calls (default: 10s, max: 30s)
    // +optional
    Timeout *metav1.Duration `json:"timeout,omitempty"`
    
    // FailurePolicy defines behavior on webhook errors
    // +kubebuilder:validation:Enum=Fail;Ignore
    // +kubebuilder:default=Fail
    FailurePolicy FailurePolicy `json:"failurePolicy,omitempty"`
    
    // TLSConfig for webhook connection
    // +optional
    TLSConfig *WebhookTLSConfig `json:"tlsConfig,omitempty"`
    
    // HMACSecretRef references a secret containing HMAC signing key
    // +optional
    HMACSecretRef *SecretKeyRef `json:"hmacSecretRef,omitempty"`
}

type WebhookTLSConfig struct {
    // CASecretRef references a secret containing CA certificate
    // +optional
    CASecretRef *SecretKeyRef `json:"caSecretRef,omitempty"`
    
    // ClientCertSecretRef references a secret containing client cert for mTLS
    // +optional
    ClientCertSecretRef *SecretKeyRef `json:"clientCertSecretRef,omitempty"`
    
    // InsecureSkipVerify disables certificate verification (NOT for production)
    // +optional
    InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
}

type SecretKeyRef struct {
    // Name of the secret
    Name string `json:"name"`
    // Key within the secret
    Key string `json:"key"`
}

type MCPWebhookConfigStatus struct {
    // ConfigHash is a hash of the spec for change detection
    ConfigHash string `json:"configHash,omitempty"`
    
    // ReferencingServers lists MCPServers using this config
    ReferencingServers []string `json:"referencingServers,omitempty"`
    
    // ObservedGeneration is the last observed generation
    ObservedGeneration int64 `json:"observedGeneration,omitempty"`
    
    // Conditions represent the latest available observations
    Conditions []metav1.Condition `json:"conditions,omitempty"`
}
```

## Example CRD Instance

```yaml
kind: MCPWebhookConfig
metadata:
  name: company-webhooks
  namespace: mcp-servers
spec:
  validating:
    - name: policy-check
      url: https://policy.company.com/validate
      timeout: 5s
      failurePolicy: Fail
      tlsConfig:
        caSecretRef:
          name: webhook-ca
          key: ca.crt
      hmacSecretRef:
        name: webhook-secrets
        key: policy-hmac

  mutating:
    - name: request-enricher
      url: https://enricher.company.com/mutate
      timeout: 3s
      failurePolicy: Ignore
```

## MCPServer Reference

```yaml
kind: MCPServer
metadata:
  name: my-mcp-server
spec:
  # ... other fields ...
  webhookConfigRef:
    name: company-webhooks
```

```go
// In mcpserver_types.go
type MCPServerSpec struct {
    // ... existing fields ...
    
    // WebhookConfigRef references an MCPWebhookConfig for webhook middleware
    // +optional
    WebhookConfigRef *WebhookConfigRef `json:"webhookConfigRef,omitempty"`
}

type WebhookConfigRef struct {
    // Name of the MCPWebhookConfig resource
    Name string `json:"name"`
}
```

## Controller Implementation

Follow the pattern from `MCPExternalAuthConfig`:

1. **Finalizer**: Prevent deletion while referenced by MCPServers
2. **Hash calculation**: Detect config changes
3. **Status updates**: Track referencing servers
4. **MCPServer reconciliation trigger**: On config changes

```go
// cmd/thv-operator/controllers/mcpwebhookconfig_controller.go

const webhookConfigFinalizer = "toolhive.stacklok.dev/webhookconfig-finalizer"

func (r *MCPWebhookConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
    // 1. Fetch MCPWebhookConfig
    // 2. Handle deletion (finalizer logic)
    // 3. Calculate config hash
    // 4. Update status if changed
    // 5. Find referencing MCPServers
    // 6. Trigger reconciliation for affected MCPServers
}
```

## Watch Setup

```go
// In mcpserver_controller.go SetupWithManager
webhookConfigHandler := handler.EnqueueRequestsFromMapFunc(
    func(ctx context.Context, obj client.Object) []reconcile.Request {
        // Find MCPServers that reference this MCPWebhookConfig
        // Return reconcile requests for each
    },
)

return ctrl.NewControllerManagedBy(mgr).
    For(&mcpv1alpha1.MCPServer{}).
    Watches(&mcpv1alpha1.MCPWebhookConfig{}, webhookConfigHandler).
    Complete(r)
```

## Config Resolution

```go
// cmd/thv-operator/pkg/controllerutil/webhook.go

func AddWebhookConfigOptions(
    ctx context.Context,
    c client.Client,
    namespace string,
    webhookConfigRef *mcpv1alpha1.WebhookConfigRef,
    options *[]runner.RunConfigBuilderOption,
) error {
    // 1. Fetch MCPWebhookConfig
    // 2. Resolve secret references (HMAC, TLS certs)
    // 3. Convert to runner.WebhookConfig
    // 4. Add to options
}
```

## Tests

- Controller unit tests with fake client
- Integration tests with envtest
- E2E tests with Chainsaw:
  - Create MCPWebhookConfig
  - Create MCPServer referencing it
  - Verify webhook config applied
  - Update MCPWebhookConfig, verify reconciliation
  - Delete MCPWebhookConfig (should fail while referenced)

## Acceptance Criteria

- [ ] CRD defined and generated (`make manifests`)
- [ ] Controller implements reconciliation loop
- [ ] Hash-based change detection working
- [ ] Finalizer prevents deletion while referenced
- [ ] MCPServer reconciliation triggered on config changes
- [ ] Secret references resolved correctly
- [ ] Unit tests with >80% coverage
- [ ] Integration tests with envtest
- [ ] E2E tests with Chainsaw

Sanskarzz:
### Summary

This PR implements the fifth phase of the dynamic webhook middleware configuration system (RFC THV-0017), introducing Kubernetes custom resource definitions (CRDs), their respective controller reconciling mechanisms, and integration into the core `MCPServer` lifecycle.

Fixes #3401 

## Large PR Justification

### Key Changes

1. **Introduced `MCPWebhookConfig` as an operator-managed API**
   - Added the webhook config CRD and controller support in the operator API surface.
   - Added `webhookConfigRef` on `MCPServer` so webhook middleware can be configured declaratively from Kubernetes.
   - Tracks `ConfigHash`, `ObservedGeneration`, and `ReferencingWorkloads` on the config resource.

2. **Aligned the controller with existing config-controller patterns**
   - Followed the `MCPExternalAuthConfig` reconciliation pattern for watches, finalizers, and reference tracking.
   - Requeues deletion checks with a `DeletionBlocked` condition instead of returning an error/backoff when the config is still referenced.
   - Avoids conflicting sequential status writes in the hash-update path.
   - Returns update errors instead of swallowing them.

3. **Integrated webhook config resolution into `MCPServer` reconciliation**
   - Resolves referenced `MCPWebhookConfig` objects during `MCPServer` reconciliation.
   - Surfaces webhook-config failures through the `Ready` condition, not only the phase field.
   - Requeues referencing `MCPServer`s when the referenced webhook config changes.

4. **Fixed the runtime path for TLS and HMAC-backed webhooks**
   - Validates `MCPWebhookConfig` specs against runtime webhook semantics before marking them `Valid=True`.
   - Mounts TLS secret material as files and passes actual filesystem paths to the webhook runtime.
   - Resolves `hmacSecretRef` in the validating and mutating middleware factories so request signing is active when configured.
   - Keeps webhook secret env vars deterministic and sanitized to avoid invalid names and spurious pod rollouts.

5. **Updated tests to match the real API and controller behavior**
   - Fixed Chainsaw cleanup ordering so `MCPServer` is removed before `MCPWebhookConfig`.
   - Fixed validation fixtures to use lowercase `fail` / `ignore`.
   - Fixed the CEL rule so mutating-only and validating-only configs are accepted, while empty specs are still rejected.
   - Added unit coverage for runtime validation, TLS secret mounting, HMAC secret resolution, and controller invalid-spec handling.

### Notable implementation details

- `MCPWebhookConfig` is served as a new `v1alpha1` API only. The v1beta1 package keeps the shared spec/status helper types, but it no longer registers a `MCPWebhookConfig` root resource.
- `MCPWebhookConfig` status uses `ReferencingWorkloads`, matching the pattern used by the other config CRDs.
- The webhook API type defines its own failure-policy type instead of importing the runtime implementation package.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

### Manual Verification

Manual testing was performed using a local Kind cluster and the `fetch` MCPServer.

1. **Setup**:
   - Deployed the operator using `task operator-deploy-local`.
   - Deployed an echo webhook server: `kubectl apply -f manual-testing-phase5/echo-server.yaml`.
   ```
    spec:
      containers:
      - name: echo
        image: ealen/echo-server:latest
   ```
2. **Configuration**:
   - Created an `MCPWebhookConfig` pointing to the echo server with `insecureSkipVerify: true`.
   - Created a `fetch` `MCPServer` referencing the config.
3. **Execution**:
   - Verified that the operator successfully reconciled the `MCPWebhookConfig` and generated a `configHash`.
   - Verified that the `fetch` server picked up the configuration and started the `thv-proxyrunner`.
   - **Result**: Inspected the `fetch` pod logs and confirmed that the mutating webhook middleware was active and correctly invoking the echo server (resulting in "denied request" logs as expected since the echo server doesn't return a valid `allowed: true` response).
4. **Dynamic Updates**:
   - Updated the `MCPWebhookConfig` (e.g., changed the failure policy or URL).
   - Verified that the operator detected the change and restarted the `fetch` pod automatically to load the new settings.


Sanskarzz:
Hey @JAORMX I need some guidance on fixing the E2E tests for this PR.

JAORMX:
So, I dug into the CI failures on this PR and there are three issues causing the operator e2e tests to fail. All three are in the test code, not the controller logic itself.

**1. Finalizer deadlock during chainsaw cleanup**

This is the big one. The `mcpwebhookconfig-reconciliation` test creates an MCPServer that references an MCPWebhookConfig. The controller correctly adds a finalizer to prevent deletion while referenced... but chainsaw cleans up in reverse creation order. So it tries to delete the MCPWebhookConfig *first*, the finalizer blocks it because the MCPServer still exists, and boom... `context deadline exceeded` after ~48 seconds.

The fix: add explicit cleanup ordering so the MCPServer gets deleted before the MCPWebhookConfig. The finalizer logic is actually working correctly here. It's just the test that doesn't account for it.

**2. `failurePolicy` case mismatch in the validation test**

The CRD enum is `fail`/`ignore` (lowercase), but the validation test uses `Fail` and `Ignore` (capitalized). The API server will reject those values outright since they're not in the enum. The test step is called `accept-valid-webhooks` and expects success... which it won't get.

**3. Empty spec vs CEL rule**

The CRD has a CEL validation rule: `size(self.validating) + size(self.mutating) > 0`. But the validation test has a step called `accept-empty-webhook` that applies `spec: {}` and expects it to succeed. That's `0 + 0 > 0` which is `false`, so the API server rejects it. Either the CEL rule needs to go (if empty specs should be valid) or the test needs to expect rejection.

Note that the `api-workloads` failure is a pre-existing flaky test unrelated to this PR.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *MCPWebhookConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&mcpv1alpha1.MCPWebhookConfig{}).
+		Complete(r)
```
JAORMX, line 251: So, this is the big one. The `MCPExternalAuthConfig` controller watches MCPServer (and MCPRemoteProxy) changes via `handler.EnqueueRequestsFromMapFunc` in its `SetupWithManager`. That's how `ReferencingWorkloads` stays up to date when servers come and go.

This controller only watches its own type. That means:

1. `ReferencingServers` won't update when an MCPServer is created or deleted
2. When a *new* MCPWebhookConfig is created that an existing (failed) MCPServer references, there's no trigger to re-reconcile that MCPServer
3. The chainsaw reconciliation test will hang waiting for the MCPServer to enter Running... because nothing wakes it up

Take a look at how the auth config controller does it (`mapMCPServerToExternalAuthConfig` + the Watches call). That's the pattern we need here.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
```diff
@@ -0,0 +1,252 @@
...
+			logger.Info("Cannot delete MCPWebhookConfig - still referenced by MCPServers",
+				"webhookConfig", webhookConfig.Name, "referencingServers", serverNames)
+
+			webhookConfig.Status.ReferencingServers = serverNames
+			if err := r.Status().Update(ctx, webhookConfig); err != nil {
+				logger.Error(err, "Failed to update MCPWebhookConfig status during deletion")
+			}
+
+			return ctrl.Result{}, fmt.Errorf("MCPWebhookConfig %s is still referenced by MCPServers: %v",
+				webhookConfig.Name, serverNames)
```
JAORMX, line 189: Returning an error here triggers controller-runtime's exponential backoff. So you'll get retries at 1s, 2s, 4s, 8s... eventually backing off to 16+ minutes between checks. That's also why the chainsaw cleanup times out after ~48 seconds.

The auth config controller handles this differently: it sets a `DeletionBlocked` condition (so users can see *why* deletion is stuck via `kubectl describe`), then returns `RequeueAfter: 30 * time.Second`. No error, no backoff, just a steady re-check.

Note that this is an expected state, not an error. The controller is doing its job by blocking deletion while references exist. It should communicate that calmly, not panic about it :)

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
```diff
@@ -0,0 +1,252 @@
...
+			"mcpserver", server.Name, "webhookConfig", webhookConfig.Name)
+
+		if server.Annotations == nil {
+			server.Annotations = make(map[string]string)
+		}
+		server.Annotations["toolhive.stacklok.dev/webhookconfig-hash"] = configHash
+
+		if err := r.Update(ctx, &server); err != nil {
+			logger.Error(err, "Failed to update MCPServer annotation", "mcpserver", server.Name)
+		}
```
JAORMX, line 155: This error is logged but not returned. If the annotation update fails, the reconciler thinks everything went fine, the MCPServer never gets the hash annotation, and nothing triggers a retry.

Per our Go style rules: return errors by default, never silently swallow them. Either return the error or collect failures and return them after the loop.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
```diff
@@ -0,0 +1,252 @@
...
+	hashChanged := webhookConfig.Status.ConfigHash != configHash
+	if hashChanged {
+		if err := r.handleConfigHashChange(ctx, webhookConfig, configHash); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Update condition if it changed
+	if conditionChanged {
+		if err := r.Status().Update(ctx, webhookConfig); err != nil {
```
JAORMX, line 97: When `hashChanged` is true, `handleConfigHashChange` already does a `Status().Update()` (which bumps the resourceVersion on the API server). Then control falls through here and this second `Status().Update()` operates on a stale resourceVersion... so you'll get `409 Conflict` errors.

The auth config controller avoids this by returning immediately from the hash change path. Consider doing the same, or consolidate all status mutations into a single write at the end of the reconcile loop.

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -268,6 +268,17 @@ func (r *MCPServerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, err
 	}
 
+	// Check if MCPWebhookConfig is referenced and handle it
+	if err := r.handleWebhookConfig(ctx, mcpServer); err != nil {
+		ctxLogger.Error(err, "Failed to handle MCPWebhookConfig")
+		// Update status to reflect the error
+		mcpServer.Status.Phase = mcpv1alpha1.MCPServerPhaseFailed
```
JAORMX, line 275: Every other handler in the reconcile loop (`handleExternalAuthConfig`, `handleTelemetryConfig`, `handleToolConfig`, `handleOIDCConfig`) calls `setReadyCondition` when failing. This one only sets Phase. So a webhook config failure won't show up in the Ready condition that users and tooling check via `kubectl get`.

Add a `setReadyCondition(mcpServer, metav1.ConditionFalse, mcpv1alpha1.ConditionReasonNotReady, err.Error())` call here to keep it consistent with the rest.

cmd/thv-operator/api/v1alpha1/mcpwebhookconfig_types.go
```diff
@@ -0,0 +1,136 @@
...
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// MCPWebhookConfig is the Schema for the mcpwebhookconfigs API.
+//
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=mwc
```
JAORMX, line 112: Missing `categories=toolhive` here. Every other CRD in the project has it, and without it `kubectl get toolhive` won't list MCPWebhookConfig resources. Should be:
```go
// +kubebuilder:resource:shortName=mwc,categories=toolhive
```

cmd/thv-operator/api/v1alpha1/mcpwebhookconfig_types.go
```diff
@@ -0,0 +1,136 @@
...
+
+// MCPWebhookConfigStatus defines the observed state of MCPWebhookConfig
+type MCPWebhookConfigStatus struct {
+	// ConfigHash is a hash of the spec, used for detecting changes
+	// +optional
+	ConfigHash string `json:"configHash,omitempty"`
+
+	// ReferencingServers lists the names of MCPServers currently using this configuration
+	// +optional
+	ReferencingServers []string `json:"referencingServers,omitempty"`
```
JAORMX, line 93: All the other config CRDs (auth, OIDC, telemetry, tool) use `[]WorkloadReference` with `+listType=map` and `+listMapKey=name` for this field. Using `[]string` here is inconsistent and would need a breaking API change if MCPRemoteProxy or VirtualMCPServer ever gains webhook support.

Would be good to match the established pattern here.

cmd/thv-operator/pkg/controllerutil/webhook.go
```diff
@@ -0,0 +1,179 @@
...
+				}
+			}
+		}
+	}
+
+	addSecrets(config.Spec.Validating)
+	addSecrets(config.Spec.Mutating)
+
+	for name, ref := range secretsToExpose {
+		envVarName := fmt.Sprintf("TOOLHIVE_SECRET_%s", name)
```
JAORMX, line 95: Two things here:

1. The map iteration is non-deterministic, so the env var order will be random on each reconcile. `deploymentNeedsUpdate` uses `DeepEqual` on the env var slice, so this will trigger spurious pod restarts. Sort the output by env var name before returning (or use `slices.SortFunc`).

2. The env var name has no sanitization. The existing `externalauth.go` uses `strings.ToUpper` and `envVarSanitizer.ReplaceAllString` to produce valid POSIX env var names. A secret named `my-hmac.key` would produce `TOOLHIVE_SECRET_my-hmac.key` here... which is not a valid env var name. Also worth adding a distinguishing prefix (like `WEBHOOK_HMAC_` instead of just `SECRET_`) to avoid collisions with other `TOOLHIVE_SECRET_*` vars from the auth side.

cmd/thv-operator/api/v1alpha1/mcpwebhookconfig_types.go
```diff
@@ -0,0 +1,136 @@
...
+	// +kubebuilder:validation:Format=duration
+	// +optional
+	Timeout *metav1.Duration `json:"timeout,omitempty"`
+
+	// FailurePolicy defines how to handle errors when communicating with the webhook.
+	// Supported values: "fail", "ignore". Defaults to "fail".
+	// +kubebuilder:validation:Enum=fail;ignore
+	// +kubebuilder:default=fail
+	// +optional
+	FailurePolicy webhook.FailurePolicy `json:"failurePolicy,omitempty"`
```
JAORMX, line 59: Nit: API types generally shouldn't import runtime implementation packages. The existing pattern (e.g., `ExternalAuthType` in the auth CRD) is to define the type locally in the API package. Importing `pkg/webhook` here creates a dependency from the API layer into the implementation layer, which can cause import cycle headaches down the road.

Consider defining `FailurePolicy` as a local type in `api/v1alpha1/`.

JAORMX (changes requested):
Thanks for the work here! The overall structure is solid and the CRD design makes sense. The main theme across most of my comments is: the issue says to follow the `MCPExternalAuthConfig` pattern, and there are a few spots where this controller diverges from it in ways that cause real problems.

The biggest ones:

1. **Missing MCPServer watch in `SetupWithManager`** -- this is what causes the chainsaw e2e test to fail. Without it, creating an MCPWebhookConfig won't trigger re-reconciliation of a failed MCPServer that references it. The `ReferencingServers` status also goes stale.

2. **`handleDeletion` returns an error instead of `RequeueAfter` + condition** -- this triggers exponential backoff and is the other half of why the chainsaw cleanup times out at ~48 seconds.

3. **Sequential `Status().Update()` calls** in the Reconcile loop will produce 409 Conflict errors in practice. The auth config controller avoids this by returning immediately from the hash-change path.

4. **`handleWebhookConfig` doesn't call `setReadyCondition`** -- breaks the pattern every other handler follows.

On the security side, `GenerateWebhookEnvVars` needs the same sanitization that `externalauth.go` applies to env var names, and the non-deterministic map iteration will cause spurious pod restarts.

The chainsaw tests also have a couple of issues I mentioned in a separate comment (failurePolicy casing, empty spec vs CEL rule).

Looking forward to the next iteration!

Sanskarzz:
Hey @JAORMX 
I wanted to let you know that I’m currently dealing with a family medical emergency, so I haven’t been able to start working review comments for this PR. I’ll do my best to push review commits for this in the next 4–5 days.

JAORMX:
@Sanskarzz completely understandable. I'm sorry you're dealing with such a thing. I wish you and your family the best and hope everything turns out OK. Don't worry about coming back to this, it can wait til you have more time, or I can pick it up. You'd, of course, keep credit for this work since you started it.

Thanks for the heads up.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
Sanskarzz: Added an `MCPServer` watch in `MCPWebhookConfigReconciler.SetupWithManager`, including the stale `ReferencingWorkloads` lookup so create/delete/ref-removal cases refresh status and requeue
  the relevant config.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
Sanskarzz: Changed deletion handling to set `DeletionBlocked`, update `ReferencingWorkloads`, and return `RequeueAfter: 30s` with no error.

cmd/thv-operator/pkg/controllerutil/webhook.go
Sanskarzz: Made webhook secret env vars deterministic and sanitized, with `WEBHOOK` identifiers to avoid collisions.

cmd/thv-operator/api/v1alpha1/mcpwebhookconfig_types.go
Sanskarzz: Replaced `ReferencingServers []string` with `ReferencingWorkloads []WorkloadReference`.

cmd/thv-operator/api/v1alpha1/mcpwebhookconfig_types.go
Sanskarzz: Added `categories=toolhive`.

cmd/thv-operator/controllers/mcpserver_controller.go
Sanskarzz: The MCPServer webhook error path now calls `setReadyCondition`.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
Sanskarzz: Updated the hash-change path to return immediately after `handleConfigHashChange`, avoiding the second stale `Status().Update()`.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
Sanskarzz: Annotation update failures are now collected and returned instead of only logged.

cmd/thv-operator/api/v1alpha1/mcpwebhookconfig_types.go
Sanskarzz: Updated Chainsaw coverage for lower-case `failurePolicy` and empty-spec CEL validation.

Sanskarzz:
@JAORMX I've addressed all the review comments and PR is ready for re-review.

JAORMX:
@Sanskarzz I got unexpectedly very busy but now I'm back on track. I'll check this out today.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	for _, server := range referencingServers {
+		logger.Info("Triggering reconciliation of MCPServer due to MCPWebhookConfig change",
+			"mcpserver", server.Name, "webhookConfig", webhookConfig.Name)
+
+		if server.Annotations == nil {
+			server.Annotations = make(map[string]string)
+		}
+		server.Annotations["toolhive.stacklok.dev/webhookconfig-hash"] = configHash
+
+		if err := r.Update(ctx, &server); err != nil {
```
JAORMX: **BLOCKER**

So, this is the highest-risk `r.Update` site in the PR. We're inside a loop over `referencingServers`, mutating each `MCPServer`'s annotations, then calling `r.Update(ctx, &server)`.

Two problems with this.

First, `.claude/rules/operator.md` is explicit on this: "Never use `r.Update` on a CR spec or metadata. Use `controllerutil.MutateAndPatchSpec` instead." The reason matters... `r.Update` is a full PUT that zeros any field the local cache doesn't track. We have multiple controllers writing `MCPServer` (auth, OIDC, telemetry, tool, this one). A stale snapshot from `findReferencingMCPServers` getting PUT back can clobber `spec.authzConfig` (auth controller's territory), or any new field added by another controller between our List and our Update.

Second, the canonical pattern at `cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go:181-189` already shows what this should look like:

```go
if err := ctrlutil.MutateAndPatchSpec(ctx, r.Client, &server, func(m *mcpv1beta1.MCPServer) {
    if m.Annotations == nil {
        m.Annotations = make(map[string]string)
    }
    m.Annotations["toolhive.stacklok.dev/externalauthconfig-hash"] = configHash
}); err != nil { ... }
```

The patch body only carries the diff, optimistic-locked.

Suggested fix:
- Switch this site to `MutateAndPatchSpec`.
- Same change at `:69` (finalizer add) and `:204` (finalizer remove). Operator rules apply equally to both.
- For the loop specifically, also re-`Get` each server inside the loop body so the diff is computed against a fresh snapshot... the List result is already racy by the time we get here.

While you're in there, the same rule applies to the status writes in this file (`:85`, `:105`, `:141`, `:196`, `:244`). They should all be `controllerutil.MutateAndPatchStatus`. The auth controller pre-dates that rule and hasn't been migrated yet, but greenfield code should land clean.

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -1034,6 +1046,26 @@ func (r *MCPServerReconciler) deploymentForMCPServer(
...
+
+		webhookVolumes, webhookMounts, err := ctrlutil.GenerateWebhookVolumes(ctx, r.Client, m.Namespace, m.Spec.WebhookConfigRef)
+		if err != nil {
+			ctxLogger := log.FromContext(ctx)
+			ctxLogger.Error(err, "Failed to generate webhook TLS volumes")
+			return nil
+		}
+		volumes = append(volumes, webhookVolumes...)
+		volumeMounts = append(volumeMounts, webhookMounts...)
+	}
```
JAORMX: **BLOCKER**

So, the error handling here is asymmetric in ways that hurt.

Two problems:

1. The env-var failure path logs and **continues**. The pod gets built without the HMAC/TLS env vars. `webhook.NewClient` at runtime then receives an empty `hmacSecret` and (per `pkg/webhook/client.go:113`) silently drops the signature. So a webhook that should be HMAC-signed sends unsigned requests, and unless the receiving server enforces signatures hard, this is a silent security control bypass.

2. The volume failure path returns `nil`. But `deploymentForMCPServer` returns `*appsv1.Deployment` with no error type. The caller can't distinguish "no deployment was built because of an error" from "no deployment is needed". Look at where it's called... the caller will treat `nil` as "everything's fine, move on", which silently kills the reconcile.

Per `.claude/rules/go-style.md`: "Return errors by default... never silently swallow."

Suggested:
- Add `error` to the function signature so failures propagate to the reconcile loop. The reconciler can then surface `Ready=False` with a real reason.
- If you want to keep the signature, at minimum: don't continue on env-var failure, fail-fast and log loudly.

Both passes there don't need both fixes... pick one. The "add error return" version is cleaner.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
```diff
@@ -0,0 +1,309 @@
...
+
+			return ctrl.Result{RequeueAfter: webhookConfigDeletionRequeueDelay}, nil
+		}
+
+		controllerutil.RemoveFinalizer(webhookConfig, WebhookConfigFinalizerName)
+		if err := r.Update(ctx, webhookConfig); err != nil {
+			logger.Error(err, "Failed to remove finalizer")
+			return ctrl.Result{}, err
+		}
+		logger.Info("Removed finalizer from MCPWebhookConfig", "webhookConfig", webhookConfig.Name)
```
JAORMX, line 257: **BLOCKER**

So, the deletion path correctly sets `DeletionBlocked=True` when references exist (good fix from round 1). But when the last reference goes away, we jump straight to finalizer removal without clearing the condition.

Two issues here:

1. Between "all refs removed" and "object actually deleted", the condition lingers as `True`. `kubectl describe mwc <name>` shows `DeletionBlocked: True, Reason: ReferencedByWorkloads` while we're in the middle of cleanly removing the resource... misleading for whoever's debugging.

2. There's an even worse edge case. If references get re-added after a previous deletion attempt was blocked but before deletion is re-attempted (workload pattern: delete server, re-create it, then try delete config), the condition stays `True` falsely between transitions.

Quick fix:

```go
meta.RemoveStatusCondition(&webhookConfig.Status.Conditions, mcpv1beta1.ConditionTypeDeletionBlocked)
if err := r.Status().Update(ctx, webhookConfig); err != nil {
    return ctrl.Result{}, err
}
controllerutil.RemoveFinalizer(...)
```

While you're in there, `webhookConfig.Status.ReferencingWorkloads` should also be set to `nil` in this branch. Same logic.

Note: `mcpexternalauthconfig_controller.go:230-237` has the same bug. Worth fixing there too, but that can be a separate PR.

cmd/thv-operator/controllers/mcpwebhookconfig_controller_test.go
JAORMX, line 1: **BLOCKER (file-level)**

So, the watch handler at `mcpwebhookconfig_controller.go:264-303` is exactly the 40-line `EnqueueRequestsFromMapFunc` you added in response to my last review. It handles two paths (currently-referenced via `WebhookConfigRef`, and stale-reference via scanning `Status.ReferencingWorkloads`) plus dedup, plus a `List` failure branch.

There's not a single test for it. Searching for `mcpServerHandler`, `MapFunc`, or "watch" in this file (or anywhere else) turns up nothing.

This is the regression vector that lands us back in the original chainsaw timeout state if anyone touches this code six months from now... and the code is non-trivial. Worth covering before merge.

Mirror the layout of `mapWebhookConfigToServers` in `mcpserver_webhook_test.go` (which is correctly tested). Cases to hit:

- `MCPServer` with `WebhookConfigRef` set... returns one request
- `MCPServer` in another config's `ReferencingWorkloads` (stale ref)... returns request for that config
- Both at once... both configs returned, no dupes (the `seen` dedup)
- Wrong object type... returns nil
- `List` failure... returns the partial requests already collected

Use `interceptor.Funcs{List: ...}` from `sigs.k8s.io/controller-runtime/pkg/client/interceptor` for the failure case.

cmd/thv-operator/pkg/controllerutil/webhook.go
```diff
@@ -0,0 +1,356 @@
...
+			ValueFrom: &corev1.EnvVarSource{
+				SecretKeyRef: &corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: ref.Name,
+					},
+					Key: ref.Key,
+				},
+			},
+		})
+	}
```
JAORMX: **MAJOR**

So, HMAC keys are mounted only as env vars while TLS material (CA, client cert, client key) gets file-mounted volumes (`GenerateWebhookVolumes` below). That asymmetry doesn't reflect the threat model... HMAC keys are equally sensitive.

Two consequences:

1. **K8s Secret name and key leak into pod metadata.** `webhookSecretIdentifier` builds names like `TOOLHIVE_SECRET_WEBHOOK_VALIDATING_<WEBHOOK>_HMAC_<SECRET_NAME>_<KEY>`. Values are masked when sourced from `valueFrom`, but the env var **names** are visible in `kubectl describe pod`, in CrashLoopBackOff debug pages, in any tooling that dumps pod specs. If a security team named the Secret `payment-webhook-prod-hmac-v3`, that name shows up unredacted. Doesn't leak the secret value, but does leak the secret topology.

2. **Subprocess inheritance.** Env vars are inherited by every child the proxy spawns. Mounted files only show up where you mount them.

Suggested fix: file-mount HMAC keys under `/etc/toolhive/webhooks/<identifier>/hmac` alongside TLS material, and have `webhook.NewClient` (or a new `webhook.LoadHMACSecret`) read from disk. This also resolves the empty-key issue I flagged on the main review body for `client.go`... files are easier to reason about.

Even if we keep the env-var approach for now, consider opaque IDs (a short hash of the identifier) so the K8s Secret name doesn't appear verbatim in pod metadata.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
```diff
@@ -0,0 +1,309 @@
...
+			}
+
+			return requests
+		},
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&mcpv1beta1.MCPWebhookConfig{}).
+		Watches(&mcpv1beta1.MCPServer{}, mcpServerHandler).
+		Complete(r)
```
JAORMX: **MAJOR**

So, the `Watches(&mcpv1beta1.MCPServer{}, mcpServerHandler)` has no predicates. Every `MCPServer` status update in the namespace fires the map func, which lists all webhook configs and scans `Status.ReferencingWorkloads`.

There's a self-cascade in here that's worth pointing out specifically. `handleConfigHashChange` at `:147-160` updates each referencing `MCPServer`'s annotations. That update fires this watch. The watch enqueues this `MCPWebhookConfig` back. The next reconcile re-runs the hash check (idempotent, fine), but we just paid a list-and-scan over the whole namespace for nothing.

In a busy namespace with lots of `MCPServer` status churn (pod ready transitions, replica changes, condition updates), this is a constant tax.

Suggested:

```go
return ctrl.NewControllerManagedBy(mgr).
    For(&mcpv1beta1.MCPWebhookConfig{}).
    Watches(&mcpv1beta1.MCPServer{}, mcpServerHandler,
        builder.WithPredicates(predicate.GenerationChangedPredicate{})).
    Complete(r)
```

`GenerationChangedPredicate` filters out status-only updates. If you want to also pick up deletions and label changes, write a custom predicate that looks at `Spec.WebhookConfigRef` plus deletion events.

Note: the auth controller has the same gap. Worth fixing in both, separately or together.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
```diff
@@ -0,0 +1,309 @@
...
+
+	conditionChanged := meta.SetStatusCondition(&webhookConfig.Status.Conditions, metav1.Condition{
+		Type:               mcpv1beta1.ConditionTypeValid,
+		Status:             metav1.ConditionTrue,
+		Reason:             "ValidationSucceeded",
+		Message:            "Spec validation passed",
+		ObservedGeneration: webhookConfig.Generation,
+	})
+
+	configHash := r.calculateConfigHash(webhookConfig.Spec)
```
JAORMX, line 107: **MAJOR (operational footgun)**

So, `CalculateConfigHash` hashes `webhookConfig.Spec`, which includes `HMACSecretRef.{Name,Key}` (the *reference*) but not the resolved secret value. That has a real-world consequence: rotating the HMAC value in the underlying K8s Secret without changing the Secret's name or key leaves the hash unchanged. The annotation on the `MCPServer` doesn't change. The deployment isn't rolled. Pods continue using the old HMAC key from their already-loaded env var.

Customers will hit this. Rotating an HMAC secret silently breaks the webhook handshake until someone manually rolls every pod that uses the config.

Three options here:

1. Hash the **resolved** Secret data alongside the spec (operator already has Secret read access). This forces a roll on rotation.
2. Move HMAC keys to file mounts (see the HMAC env-var disclosure comment on `pkg/controllerutil/webhook.go`). The kubelet projects Secret updates onto mounted volumes within seconds, and the runtime can re-read on each request or use `fsnotify`.
3. Document the gap loudly: tell users to rotate via new SecretKeyRef names, not in-place.

Option 2 is cleanest because it solves the env-var disclosure issue too.

pkg/webhook/secrets.go
JAORMX, line 1: **MAJOR (file-level)**

So, this is 56 lines of secret-resolution logic with multiple branches:

- empty `ref` returns `(nil, nil)` (line 18-20)
- `ParseSecretParameter` happy path vs raw-name fallback (line 22-25)
- `resolveSecretProvider`: env-var override, `SetupCompleted=false`, `GetProviderType` failure, `CreateProvider` failure (line 40-56)
- `provider.GetSecret` failure path

Codecov reports 43.47% on this file. Reading it, the only path that's exercised today is the env-var-provider happy path via `TestCreateMiddleware_ResolvesHMACSecret`. The error wrapping (`"failed to resolve webhook secret %q: %w"`) is untested... a refactor that drops the wrapping won't be caught.

This file is small and self-contained. A `secrets_test.go` with table-driven cases is cheap. At minimum:

- empty ref short-circuit (no provider call)
- env-var override path (`t.Setenv` for `secrets.ProviderEnvVar`)
- `SetupCompleted=false` returns `errors.Is(err, secrets.ErrSecretsNotSetup)` (sentinel check, not string match)
- missing-secret returns wrapped error, name preserved

Also, the error fallthrough on line 23-25 silently discards the `ParseSecretParameter` error:

```go
if secretParam, err := secrets.ParseSecretParameter(ref); err == nil {
    secretName = secretParam.Name
}
```

The fallback ("treat ref as raw name") is intentional but the err drop without a comment will confuse readers. Per `go-style.md` "Comment ignored errors", drop a one-liner explaining the dual-format intent.

Sanskarzz:
@JAORMX  No worries, thanks for the update!

cmd/thv-operator/controllers/mcpwebhookconfig_controller_test.go
```diff
@@ -0,0 +1,272 @@
...
+			expectFinalizer: true,
+			expectHash:      true,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			ctx := context.Background()
```
JAORMX: **MINOR (spread across multiple files)**

So, project convention (per `MEMORY.md` and the patterns in `mcpserver_oidcconfig_test.go`, `mcpserver_telemetry_cabundle_test.go`, etc.) is `t.Context()`, not `context.Background()`. Six call sites in this PR's test files miss it:

- `mcpwebhookconfig_controller_test.go:90, 162, 169, 253, 259`
- `mcpserver_webhook_test.go:162`
- `pkg/controllerutil/webhook_test.go:70`

`mcpserver_webhook_test.go:120` correctly uses `t.Context()` already... same file, just inconsistent.

`t.Context()` cancels automatically when the test finishes. `context.Background()` doesn't, so a hanging fake-client call can outlive the test goroutine. Quick mechanical fix.

cmd/thv-operator/controllers/mcpwebhookconfig_controller_test.go
```diff
@@ -0,0 +1,272 @@
...
+	ctx := context.Background()
+	result, err := r.handleDeletion(ctx, webhookConfig)
+	require.NoError(t, err)
+	assert.Equal(t, webhookConfigDeletionRequeueDelay, result.RequeueAfter)
+
+	// Delete server and try again
+	require.NoError(t, fakeClient.Delete(ctx, mcpServer))
+
+	_, err = r.handleDeletion(ctx, webhookConfig)
+	assert.NoError(t, err, "Should delete successfully after reference removed")
```
JAORMX: **MAJOR**

So, `TestMCPWebhookConfigReconciler_handleDeletion` asserts:

```go
result, err := r.handleDeletion(ctx, webhookConfig)
require.NoError(t, err)
assert.Equal(t, webhookConfigDeletionRequeueDelay, result.RequeueAfter)
```

That checks the `RequeueAfter` value and that no error came back. It does **not** check that the `DeletionBlocked` condition was set, that its `Reason` is `ReferencedByWorkloads`, that `Status.ReferencingWorkloads` was populated, or any of the actual behavior the round-1 fix added.

A regression that flips `Status` to `False`, drops the reason, or forgets to populate `ReferencingWorkloads` would not be caught.

Add post-call assertions:

```go
require.NoError(t, fakeClient.Get(ctx, client.ObjectKeyFromObject(webhookConfig), webhookConfig))
cond := meta.FindStatusCondition(webhookConfig.Status.Conditions, mcpv1beta1.ConditionTypeDeletionBlocked)
require.NotNil(t, cond)
assert.Equal(t, metav1.ConditionTrue, cond.Status)
assert.Equal(t, "ReferencedByWorkloads", cond.Reason)
assert.NotEmpty(t, webhookConfig.Status.ReferencingWorkloads)
```

Same gap on the "all refs gone" branch (line 259-262)... after the second `handleDeletion` call, re-fetch and assert finalizers are clear and (per the `DeletionBlocked` comment on the controller) the condition is removed.

JAORMX (changes requested):
Thanks for the rework, Sanskar. All nine items from the previous round landed properly, I verified each one against the code (the watch in `SetupWithManager`, the `DeletionBlocked` condition path, the sorted and sanitized env vars, the `[]WorkloadReference` switch, all of it). Solid follow-through.

I dug deeper this round and surfaced a handful of new issues. Detailed inline comments with code excerpts and suggested fixes are coming alongside this. Quick rollup:

## Blockers (in this PR's diff)

1. **`r.Update` violates `.claude/rules/operator.md`** at `cmd/thv-operator/controllers/mcpwebhookconfig_controller.go:69, 156, 204`. The annotation fan-out at `:156` is the riskiest one... it writes a stale `MCPServer` snapshot back and can zero fields owned by the auth controller.
2. **Silent error swallow in `deploymentForMCPServer`** at `cmd/thv-operator/controllers/mcpserver_controller.go:1049-1067`. On env-var failure we log and continue (pod starts without HMAC/TLS), on volume failure we return `nil`. The function has no error return so the caller can't distinguish "no deployment needed" from "build failed".
3. **`DeletionBlocked` condition never gets cleared** at `cmd/thv-operator/controllers/mcpwebhookconfig_controller.go:188-208`. The no-references branch jumps straight to `RemoveFinalizer` without `meta.RemoveStatusCondition` first.
4. **The watch fix from round 1 has zero test coverage**. `mcpServerHandler` at `cmd/thv-operator/controllers/mcpwebhookconfig_controller.go:264-303` is the exact 40-line `EnqueueRequestsFromMapFunc` you added in response to my last review and there's no unit test for either the `WebhookConfigRef` path or the stale-`ReferencingWorkloads` path. Easy regression vector.

## Pre-existing webhook code that this PR makes live

These are issues in `pkg/webhook/` (Phase 1 / Phase 2 code) that this PR doesn't touch directly, but that this PR's runtime activation makes load-bearing. Not on you to fix here, but they should land before we tell anyone the webhook integration is production-ready. Happy to file a separate hardening issue, or you can if you'd prefer.

5. **SSRF in the webhook HTTP client** at `pkg/webhook/client.go:160-212`. The custom `http.Transport` is built without `protectedDialerControl` wired in. `ValidatingTransport` only checks the URL scheme, it doesn't block private/loopback/link-local IPs (the `#nosec G704` comment at `:120` claims otherwise, but it's wrong). With `MCPWebhookConfig` create rights, a tenant can point `url:` at `169.254.169.254` (cloud metadata, IAM creds plus every JWT principal we send) or `kubernetes.default.svc.cluster.local`. Suggested fix: route through `networking.NewHttpClientBuilder()` so the dialer rejects private IPs, or set `transport.DialContext` directly. If we want in-cluster webhooks as a real use case, an explicit `AllowPrivateIPs` flag on `WebhookTLSConfig` (default false) is the way.

6. **`InsecureSkipVerify` also unlocks plaintext HTTP** at `pkg/webhook/types.go:95-99`. The flag conflates "skip cert validation" with "allow plaintext HTTP". And `client.go:169-170` propagates it into `ValidatingTransport.InsecureAllowHTTP=true`, which (per `pkg/networking/http_client.go:57-59`) **skips ALL request validation** when set... cert check, scheme check, and the SSRF guard from #5 if/when it's added. Flip one knob, lose three things. RFC THV-0017 says HTTPS-only. Suggested: split into two fields, or just remove the plaintext escape hatch entirely. Add a CEL rule on the CRD to enforce `self.url.startsWith('https://')` at admission time as well.

7. **`webhook.NewClient` silently signs with empty key** at `pkg/webhook/client.go:34-60`, with `:113` skipping signing when `len(c.hmacSecret) == 0`. If `ResolveSecret` returns `(nil, nil)` (the empty-ref branch at `pkg/webhook/secrets.go:18-20`, or a Secret whose key exists but has an empty value), we silently send unsigned requests. Per `.claude/rules/go-style.md` "Constructor Validation: Fail Loudly", this should error if `cfg.HMACSecretRef != "" && len(hmacSecret) == 0`.

8. **Webhook response body echoed into client errors** at `pkg/webhook/client.go:140-153, 229-236`. We echo the first 256 bytes of any non-2xx response into the returned error, which then gets logged at `slog.Error` in the middleware. Combined with #5, this is an exfiltration path... an attacker pointing at internal services reads the first 256 bytes of any error response straight from the cluster API into our logs. Suggested: log the body at DEBUG inside the webhook package, return only `(name, status_code, classification)` in the error.

9. **Unbounded `io.ReadAll` on the request body** in both `pkg/webhook/validating/middleware.go` and `pkg/webhook/mutating/middleware.go`. Client side correctly enforces `MaxResponseSize` via `io.LimitReader`, server side missed. A 10 GB body gets fully buffered before forwarding. Wrap with `http.MaxBytesReader(w, r.Body, MaxRequestSize)` and define `MaxRequestSize` next to `MaxResponseSize` in `types.go`.

## Majors (inline)

I've also flagged a handful of MAJORs inline (HMAC env-var leakage of Secret names into pod metadata, `ConfigHash` blind to Secret rotation, missing watch predicates, `pkg/webhook/secrets.go` having no test file, `DeletionBlocked` not asserted in tests). Those don't all need to block this PR... happy to scope which land here vs follow-up issues.

No rush on this. Take your time.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
Sanskarzz: Updated. Replaced the r.Update sites with MutateAndPatchSpec and the status writes with MutateAndPatchStatus. For MCPServer annotation fan-out, the reconciler now re-fetches each server before patching so the diff is built from a fresh object and avoids clobbering fields owned by other controllers.

cmd/thv-operator/controllers/mcpserver_controller.go
Sanskarzz: Updated. deploymentForMCPServer now returns (*appsv1.Deployment, error), and webhook env/config/volume generation failures are propagated instead of logged and ignored. The reconcile path now surfaces these as Ready=False / Failed instead of silently continuing.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
Sanskarzz: Updated. The deletion path now clears DeletionBlocked and Status.ReferencingWorkloads before removing the finalizer when references are gone.

cmd/thv-operator/controllers/mcpwebhookconfig_controller_test.go
Sanskarzz: Updated. Added unit coverage for mapMCPServerToWebhookConfig covering current WebhookConfigRef, stale ReferencingWorkloads, dedupe, wrong object type, and List failure returning partial requests.

cmd/thv-operator/pkg/controllerutil/webhook.go
Sanskarzz: Updated. Webhook HMAC secrets are no longer exposed as TOOLHIVE_SECRET_* env vars. They are mounted as files under /etc/toolhive/webhooks/<opaque-id>/hmac, and generated identifiers no longer embed the Kubernetes Secret name/key.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
Sanskarzz: Updated. Added GenerationChangedPredicate to the MCPServer watch so status-only updates do not enqueue MCPWebhookConfig reconciles or trigger the list-and-scan path unnecessarily.

cmd/thv-operator/controllers/mcpwebhookconfig_controller.go
Sanskarzz: Updated. Since HMAC secrets are now file-mounted, the webhook client re-reads absolute HMAC file refs per request. That lets projected Kubernetes Secret updates take effect without requiring the MCPWebhookConfig hash or pod template annotation to change.

pkg/webhook/secrets.go
Sanskarzz: Updated. Added tests for empty refs, env provider override, CLI-style refs, ErrSecretsNotSetup, missing-secret wrapping with the secret name preserved, and absolute file refs. Also added a comment explaining the intentional raw-name fallback when ParseSecretParameter fails.

cmd/thv-operator/controllers/mcpwebhookconfig_controller_test.go
Sanskarzz: Updated. Replaced the remaining context.Background() usages in these test files with t.Context(). Also strengthened the deletion tests to assert DeletionBlocked=True, Reason=ReferencedByWorkloads, populated ReferencingWorkloads, and cleanup of stale deletion-blocked status when references are gone.

JAORMX:
By the way, I just noticed that the webhook CRDs are `v1beta1`, can we kick them off as `v1alpha1` instead?

JAORMX (approved):
All round-2 items land cleanly. Verified each against the code at `d65e0d2d`:

**BLOCKERs (4/4)**
- `r.Update` → `MutateAndPatchSpec` / `MutateAndPatchStatus` everywhere, with a fresh `Get` inside the annotation fan-out loop so each patch is computed against current state.
- `deploymentForMCPServer` returns `(*appsv1.Deployment, error)`; both call sites surface failures as `Phase=Failed` / `Ready=False`.
- `handleDeletion` clears `DeletionBlocked` and zeroes `Status.ReferencingWorkloads` before `RemoveFinalizer`.
- `mapMCPServerToWebhookConfig` test covers all five cases including the `interceptor.Funcs` List-failure path.

**MAJORs (5/5)**
- HMAC keys file-mounted under `/etc/toolhive/webhooks/<opaque-id>/hmac`, identifiers no longer embed the K8s Secret name; runtime re-reads on each request via the absolute-path branch in `pkg/webhook/secrets.go` (so projected Secret rotation takes effect without a hash bump).
- `GenerationChangedPredicate` on the MCPServer watch.
- `pkg/webhook/secrets_test.go` covers empty refs, env-var override, CLI-style refs, `ErrSecretsNotSetup`, missing-secret wrapping, and absolute file refs; raw-name fallback comment added.
- Deletion test asserts condition status + reason, populated `ReferencingWorkloads`, and the cleanup branch.
- `t.Context()` adopted (the two `context.Context` references that remain are inside `interceptor.Funcs` mock callback signatures, which is the documented exception).

**v1alpha1 rename**: clean. Root `MCPWebhookConfig` moved to `v1alpha1`, spec/status helper types stayed in `v1beta1` consistent with the rest of the operator. CRD chart, chainsaw fixtures, and all controller imports aligned.

**Pre-existing `pkg/webhook/` items**: #7 (silent unsigned send when HMAC ref configured but empty) is fixed in this PR's `client.go:42-44`. #5 (SSRF — `buildTransport` doesn't route through `networking.NewHttpClientBuilder` and `ValidatingTransport` only checks scheme), #6 (`InsecureSkipVerify` still toggles both cert-skip and plaintext-HTTP, no CRD CEL rule enforcing `https://`), #8 (256-byte response-body preview still echoed in error strings), and #9 (unbounded `io.ReadAll` in both middleware files) are still present. Per round 2 those weren't on this PR to fix; I'll open a follow-up hardening issue so they don't get lost before this code is enabled in earnest.

Solid follow-through across both rounds.

Sanskarzz:
@JAORMX  Thanks for the detailed re-review and approval. Glad the round-2 fixes landed cleanly.
Agreed on tracking the remaining `pkg/webhook` hardening items separately before this is considered production-ready. I’ll keep an eye out for the follow-up issue, If possible please tag me and I can help with those fixes as well.

PR #4574, opened 2026-04-06
Add E2E acceptance test for backend routing after proxy restart

jerm-dro:
## Summary

- When `backendReplicas > 1` and the proxy runner restarts, the recovered session's `backend_url` points to the ClusterIP service. Kube-proxy may route to a backend pod that never handled the session's `initialize`, causing HTTP 404 / JSON-RPC `-32001` "session not found" errors.
- This PR adds an E2E acceptance test that demonstrates the bug: it creates an MCPServer with `backendReplicas=2`, establishes a session, deletes the proxy runner pod, and sends 5 requests with the same session ID. With 2 backends and random routing, P(all 5 hit correct pod) ≈ 3%, making the failure reliably detectable.
- **This test is expected to fail** until pod-specific headless DNS routing is implemented.

## Type of change

- [x] New feature

## Test plan

- [x] E2E tests (`task test-e2e`) — this PR *adds* the test; it is expected to fail on CI, proving the bug exists
- [x] Unit tests (`task test`) — no production code changed, existing tests unaffected

No

This test is intentionally expected to fail on CI. It demonstrates the backend routing bug that will be fixed in a follow-up PR implementing pod-specific headless DNS routing. The test context `"backendReplicas=2 and proxy runner restarts"` will pass once the fix lands.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,3 +1,3 @@
 [codespell]
-ignore-words-list = NotIn,notin,AfterAll,ND,aks,deriver,te
+ignore-words-list = NotIn,notin,AfterAll,ND,aks,deriver,te,clientA
 skip = *.svg,*.mod,*.sum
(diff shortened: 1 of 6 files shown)


jerm-dro:
folded into another PR

PR #4587, opened 2026-04-07
Add MinLength=1 validation to MCPServer Image field

### LINKED ISSUE #4585: Add MinLength=1 validation to Image field in MCPServer CRD
## Problem

The `Image` field in the MCPServer CRD spec (`cmd/thv-operator/api/v1alpha1/mcpserver_types.go`) has `+kubebuilder:validation:Required` but lacks a `MinLength=1` constraint. This means users can create an MCPServer with `image: ""` — the empty string passes CRD admission validation but fails at container runtime when the controller attempts to create a Deployment with an empty image.

## Location

- `cmd/thv-operator/api/v1alpha1/mcpserver_types.go:160-162`

```go
// Image is the container image for the MCP server
Image string `json:"image"`
```

## Impact

- Users can submit `image: ""` and only discover the error asynchronously via status conditions
- The controller attempts to create a Deployment with an empty image, which fails
- Other string fields in the same CRD (e.g., `sessionStorage.address`, `oidcConfigRef.name`) already correctly use `MinLength=1`

## Recommended Fix

1. Add `// +kubebuilder:validation:MinLength=1` marker to the `Image` field
2. Run `task gen` to regenerate CRD manifests

## Context

Identified during CRD stabilization audit (MF-5). This is a validation tightening — it is technically a breaking change since `image: ""` was previously accepted, but such input was never valid at runtime.

## Related

- MF-6 (ResourceList string validation — similar admission-time validation gap)
- MF-7 (SamplingRate string validation — similar pattern)

ChrisJBurns:
## Summary

- The `Image` field in `MCPServerSpec` only had a `Required` kubebuilder marker, which prevents omission but still accepts empty strings (`image: ""`). An empty image passes CRD admission validation but fails at container runtime when the controller creates a Deployment. Adding `MinLength=1` catches this invalid input at admission time, matching the pattern already used by other validated string fields in the same CRD.

Fixes #4585

## Type of change

- [x] Bug fix

## Test plan

Verified `minLength: 1` appears in the regenerated CRD manifest for the `image` field.

Yes — `image: ""` in MCPServer manifests will now be rejected at admission time with a validation error, rather than being accepted and failing asynchronously at container runtime. This is a validation tightening; empty image strings were never valid at runtime.

### DIFF
--- a/cmd/thv-operator/api/v1alpha1/mcpserver_types.go
+++ b/cmd/thv-operator/api/v1alpha1/mcpserver_types.go
@@ -159,6 +159,7 @@ const SessionStorageProviderRedis = "redis"
 type MCPServerSpec struct {
 	// Image is the container image for the MCP server
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
 	Image string `json:"image"`
 
 	// Transport is the transport method for the MCP server (stdio, streamable-http or sse)
--- a/deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpservers.yaml
+++ b/deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpservers.yaml
@@ -183,6 +183,7 @@ spec:
                 type: string
               image:
                 description: Image is the container image for the MCP server
+                minLength: 1
                 type: string
               mcpPort:
                 description: McpPort is the port that MCP server listens to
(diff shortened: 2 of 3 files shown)


ChrisJBurns:
@claude  Does this PR make much sense, is there a better validation we can add?

ChrisJBurns:
Closing because minLength doesn't really add any validation as a single character image isn't valid anyways.

PR #4605, opened 2026-04-07
Use template helpers for RBAC resource names in Helm chart

### LINKED ISSUE #4604: Helm chart RBAC templates use hardcoded names instead of template helpers
## Problem

The ClusterRole and ClusterRoleBinding templates in the operator Helm chart use hardcoded names instead of the `operator.fullname` and `operator.serviceAccountName` template helpers defined in `_helpers.tpl`.

**Affected files:**
- `deploy/charts/operator/templates/clusterrole/role.yaml` — hardcodes `toolhive-operator-manager-role`
- `deploy/charts/operator/templates/clusterrole/rolebinding.yaml` — hardcodes `toolhive-operator-manager-rolebinding`, `toolhive-operator-manager-role`, and `toolhive-operator` (ServiceAccount name)

**Other templates use helpers correctly:**
- `deployment.yaml` uses `{{ include "operator.fullname" . }}` and `{{ include "operator.serviceAccountName" . }}`
- `serviceaccount.yaml` uses `{{ include "operator.fullname" . }}`

## Impact

1. **`fullnameOverride` / `nameOverride` breaks RBAC** — the ServiceAccount gets a new name but the ClusterRoleBinding still references the old hardcoded name, so the operator loses permissions
2. **Multiple Helm releases collide** — deploying two releases in the same cluster produces duplicate ClusterRole/ClusterRoleBinding names, causing one to overwrite the other
3. **Inconsistency** — RBAC templates are the only templates not using the standard helpers

## Recommended Fix

Replace hardcoded strings with template helpers:
- ClusterRole name: `{{ include "operator.fullname" . }}-manager-role`
- ClusterRoleBinding name: `{{ include "operator.fullname" . }}-manager-rolebinding`
- ServiceAccount reference: `{{ include "operator.serviceAccountName" . }}`
- ClusterRole reference in binding: `{{ include "operator.fullname" . }}-manager-role`

## Breaking Change Note

This changes Helm resource names from hardcoded to templated. On upgrade, existing Helm releases will get new ClusterRole/ClusterRoleBinding names, orphaning the old ones. Users may need to manually delete the old resources after upgrading.

ChrisJBurns:
## Summary

- The ClusterRole, ClusterRoleBinding, and namespace-scoped RoleBinding templates hardcoded resource names (`toolhive-operator-manager-role`, `toolhive-operator-manager-rolebinding`) and the ServiceAccount reference (`toolhive-operator`) instead of using the `operator.fullname` and `operator.serviceAccountName` helpers from `_helpers.tpl`. This caused RBAC breakage when using `nameOverride`/`fullnameOverride` and resource name collisions when deploying multiple Helm releases in the same cluster.
- Replace all hardcoded names with the existing template helpers and upgrade labels from `toolhive.labels` to `operator.labels` for consistency with `deployment.yaml` and other templates.

Fixes #4604

## Type of change

- [x] Bug fix

## Test plan

Validated with `helm template` and `helm lint`:
- Default values: ClusterRole/ClusterRoleBinding names resolve to `toolhive-operator-manager-role` / `toolhive-operator-manager-rolebinding` (unchanged default behavior)
- With `--set fullnameOverride=my-custom-name`: names correctly resolve to `my-custom-name-manager-role` / `my-custom-name-manager-rolebinding`
- With `--set operator.rbac.scope=namespace --set 'operator.rbac.allowedNamespaces={ns1,ns2}'`: namespace-scoped RoleBindings use templated names and correct labels
- `helm lint` passes cleanly

## Changes

| File | Change |
|------|--------|
| `deploy/charts/operator/templates/clusterrole/role.yaml` | Replace hardcoded name with `operator.fullname` helper; add `operator.labels` |
| `deploy/charts/operator/templates/clusterrole/rolebinding.yaml` | Replace hardcoded names/SA refs with `operator.fullname`, `operator.serviceAccountName`; upgrade to `operator.labels` |

Yes — on upgrade, existing Helm releases will get new ClusterRole/ClusterRoleBinding resource names if using non-default `nameOverride`/`fullnameOverride`. Users with default values will see no change. Users with custom name overrides may need to manually delete orphaned old ClusterRole/ClusterRoleBinding resources after upgrading.

The `values.yaml` default `serviceAccount.name: "toolhive-operator"` means the `operator.serviceAccountName` helper always returns `toolhive-operator` even with `fullnameOverride`, since the explicit name takes precedence. This is a pre-existing inconsistency between `serviceaccount.yaml` (which uses `operator.fullname`) and the helper, and is out of scope for this fix.


deploy/charts/operator/templates/clusterrole/rolebinding.yaml
```diff
@@ -3,16 +3,16 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: toolhive-operator-manager-rolebinding
+  name: {{ include "operator.fullname" . }}-manager-rolebinding
   labels:
-    {{- include "toolhive.labels" . | nindent 4 }}
+    {{- include "operator.labels" . | nindent 4 }}
```
JAORMX, line 8: wouldn't we also want `toolhive.labels` here?

ChrisJBurns: Yep good point

JAORMX: responding to my own comment. This would effectively change the labels from:
```
app: toolhive
app.kubernetes.io/name: toolhive
```

To:

```
app.kubernetes.io/name: {{ include "operator.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
app.kubernetes.io/part-of: {{ include "operator.name" . }}
```

ChrisJBurns: Good catch — updated in 8687c008 to include both `operator.labels` and `toolhive.labels` on both the ClusterRoleBinding and namespace-scoped RoleBindings.

deploy/charts/operator/templates/clusterrole/role.yaml
```diff
@@ -2,7 +2,9 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: toolhive-operator-manager-role
+  name: {{ include "operator.fullname" . }}-manager-role
+  labels:
+    {{- include "operator.labels" . | nindent 4 }}
```
jhrozek, line 7: Nit: The bindings include both `operator.labels` and `toolhive.labels`, but the ClusterRole only has `operator.labels`. Consider adding `toolhive.labels` here too for consistency.

ChrisJBurns: Added in 1c27bcc3 — ClusterRole now includes both `operator.labels` and `toolhive.labels`, consistent with the bindings.

ChrisJBurns:
Closing this as it seems the `ClusterRole` is auto generated by kubebuilder. So it isn't something we can add helm templating into as easily

PR #4636, opened 2026-04-07
Add coding conventions derived from recent PR review feedback

yrobla:
## Summary

Captures patterns flagged in reviews across interface design, testing, operator development, and PR hygiene — so they apply automatically going forward rather than being caught in review.

Fixes #

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [x] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
|      |        |

### DIFF
--- a/.claude/rules/go-style.md
+++ b/.claude/rules/go-style.md
@@ -19,7 +19,7 @@ Check these whenever adding a method to an interface or defining a new type:
 
 - **Minimal surface**: Don't add interface methods that duplicate the semantics of existing ones. If an existing method already answers the question (possibly with a side effect), don't add a separate method for the same check.
 - **No silent no-ops**: A no-op that silently breaks callers who depend on the method working is a sign the interface is too broad. Narrow the interface or use a separate capability interface. Benign no-ops (e.g., `Close()` on an in-memory store) are fine.
-- **Option pattern must be compile-time safe**: Options should not have runtime failures. Never define a local anonymous interface inside an option and type-assert against it — a silent no-op results if the target doesn't implement it. Two typesafe approaches:
+- **Option pattern must be compile-time safe**: Never define a local anonymous interface inside an option and type-assert against it to check capability — a silent no-op results if the target doesn't implement it. (Returning an explicit error from an option for input validation is fine.) Two typesafe approaches:
   - *Config struct field*: put the setting on the config struct (e.g., `types.Config.SessionStorage`) so all consumers see it at compile time.
   - *Typed functional option*: use `func(*ConcreteType)` so the option only compiles against the correct receiver.
   If you need to cast inside an option to check whether the target supports it, the option is on the wrong abstraction. See #4638.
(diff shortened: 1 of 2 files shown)


.claude/rules/go-style.md
```diff
@@ -13,6 +13,33 @@ Applies to all Go files in the project.
 - Separate business logic from transport/protocol concerns
 - Keep packages focused on single responsibilities
```
jerm-dro, line 15: `suggestion:` The "No no-op implementations" rule is good but may be too absolute. A no-op `Close()` on an in-memory store is perfectly fine — the problem is when a no-op *silently breaks callers' expectations*. Consider softening to: "A no-op that silently breaks callers who depend on the method working is a sign the interface is too broad."

.claude/rules/go-style.md
```diff
@@ -13,6 +13,33 @@ Applies to all Go files in the project.
...
 
+## Interface Design
+
+Check these whenever adding a method to an interface or defining a new type:
```
jerm-dro, line 18: `suggestion:` This section should include guidance on the **option pattern**, which is a common source of silent failures. We encountered a case in #4638 where a transport option defines a local anonymous interface, type-asserts against an unexported method, and silently no-ops if the target doesn't implement it.

Options should not have runtime failures — they should be validated at compile time. Two typesafe approaches:
- **Config struct field**: put the setting on the config struct (like `types.Config.SessionStorage`), so all consumers see it at compile time
- **Typed functional option**: use `func(*ConcreteType)` so the option only compiles against the correct receiver

There should be no interface casting in option application. If you need to cast to check whether the target supports the option, the option is on the wrong abstraction.

.claude/rules/vmcp-anti-patterns.md
```diff
@@ -79,3 +79,11 @@ Adding caches, connection pools, or other performance optimizations without evid
...
 
 **Instead**: Start with the straightforward implementation. Measure under realistic load. Add optimization only when measurements show it's needed, and document the evidence in the commit or PR description.
+
+## 10. Mutable Domain Objects with Mutex Protection
+
+Adding a mutex to a domain object and mutating it in place when state changes. This grows in complexity with every new mutation and makes objects harder to reason about under concurrency.
+
+**Detect**: Mutex fields on domain structs; mutation methods on types that were previously read-only; in-place writes guarded by an object-level lock.
+
+**Instead**: Prefer immutable objects. When state changes, replace the object rather than mutating it — rebuild from the source of truth and let readers always work with a consistent snapshot.
```
jerm-dro: `suggestion:` The anti-pattern description is on the right track but undersells the core principle. The key question isn't "mutex or no mutex" — it's **can the object be reconstructed instead of mutated?** If state changes, prefer rebuilding from the source of truth and replacing the reference. This eliminates an entire class of concurrency bugs.

Beyond that, the guidance should address **synchronization layering**: multiple layers each holding their own mutex is a smell. Prefer centralizing synchronization at a single layer — this creates a clear contract that everything below is single-threaded, which is much easier to reason about. The only exception is sharding mutexes for performance, and that should be introduced after measurement shows contention, not preemptively (aligns with anti-pattern #9).

Suggested rewrite for the "Instead" section:
> **Instead**: Ask whether the object can be reconstructed rather than mutated — rebuild from the source of truth and replace the reference. If mutation is truly necessary, centralize synchronization at one layer rather than distributing mutexes across multiple layers. Sharded locks for performance should only be introduced after profiling shows contention.

PR #4638, opened 2026-04-07
Route MCP sessions to specific StatefulSet pods via headless DNS

### LINKED ISSUE #4575: Backend routing breaks after proxy runner restart with backendReplicas > 1
## Problem

When an MCPServer has `backendReplicas > 1`, the proxy runner stores the **ClusterIP service URL** (e.g. `http://mcp-<name>:8080`) as `backend_url` in session metadata. After a proxy runner restart, it recovers the session from Redis but routes via ClusterIP — kube-proxy may send the request to a backend pod that never handled the `initialize` for that session.

The backend pod returns HTTP 404 with JSON-RPC `-32001` ("session not found") because it has no record of that session. The mcp-go client surfaces this as:

```
transport error: failed to send request: session terminated (404). need to re-initialize
```

The session is not actually terminated — it simply doesn't exist on the pod that received the request.

Parent issue: #4484

## Failing acceptance test

PR #4574 adds an E2E test that demonstrates this bug. It fails on all 3 Kubernetes versions in CI:

- [v1.33.7](https://github.com/stacklok/toolhive/actions/runs/24040941288/job/70112255691), [v1.34.3](https://github.com/stacklok/toolhive/actions/runs/24040941288/job/70112255684), [v1.35.1](https://github.com/stacklok/toolhive/actions/runs/24040941288/job/70112255710)

Error from CI (v1.35.1):
```
[FAILED] Request 1/5 should succeed — session should route to the correct backend
Unexpected error:
    transport error: failed to send request: session terminated (404). need to re-initialize
In [It] at: mcpserver_scaling_test.go:382
```

### Test flow

1. Deploy MCPServer with `replicas=1`, `backendReplicas=2`, Redis session storage, `sessionAffinity=None`
2. Initialize MCP session, call `tools/list` — succeeds
3. Delete the proxy runner pod (Deployment recreates it)
4. Send 5 `tools/list` requests with the same session ID — fails on request 1

With 2 backends and random routing, P(all 5 hit correct pod) ≈ 3%.

## Root cause

1. **StatefulSet `serviceName` mismatch**: `buildStatefulSetSpec` sets `serviceName` to `containerName`, but the headless service is named `mcp-<containerName>-headless`. These must match for Kubernetes to create pod DNS records.

2. **ClusterIP stored as `backend_url`**: `RoundTrip` stores `targetURI` (ClusterIP) as `backend_url`. After restart, `Rewrite` routes via ClusterIP with `sessionAffinity=None`, hitting a random backend.

## Proposed fix

Store a **pod-specific headless DNS URL** (e.g. `myserver-0.mcp-myserver-headless.default.svc.cluster.local:8080`) as `backend_url` instead of the ClusterIP:

- Fix StatefulSet `serviceName` to match headless service name
- Add `HeadlessServiceConfig` to `ScalingConfig`
- Pre-select a random StatefulSet pod on initialize, store its headless DNS as `backend_url`
- Operator populates config when `backendReplicas > 1`
- Backward compatible: when config is nil (single backend), behavior is unchanged

yrobla:
## Summary

When backendReplicas > 1, the proxy was storing the ClusterIP service URL as backend_url on session initialize. After a proxy-runner restart, Redis-recovered sessions were routed through the ClusterIP, which sent requests to a random pod — one that had never seen the initialize and returned -32001 ("session not found").

Two root causes fixed:

1. StatefulSet serviceName was set to containerName instead of mcp-<name>-headless, so Kubernetes never created pod DNS records.

2. On initialize, the proxy now picks a random pod ordinal and stores its headless DNS URL (e.g. myserver-0.mcp-myserver-headless.default .svc.cluster.local) as backend_url instead of the ClusterIP. The Rewrite closure already reads backend_url to route follow-up requests, so sessions now survive proxy-runner restarts.

The operator populates a new HeadlessServiceConfig (StatefulSet name, headless service name, namespace, replica count) in ScalingConfig when backendReplicas > 1. Single-replica and non-Kubernetes deployments are unaffected.

Fixes #4575 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
|      |        |

## Large PR Justification

This is a complete PR that solves a bug, and is an isolated functionality. Cannot be split.

### DIFF
--- a/pkg/transport/proxy/transparent/transparent_proxy.go
+++ b/pkg/transport/proxy/transparent/transparent_proxy.go
@@ -141,9 +141,8 @@ type TransparentProxy struct {
 	shutdownTimeout time.Duration
 
 	// headlessService holds Kubernetes headless service info for pod-specific routing.
-	// When set (backendReplicas > 1), new sessions are pinned to a randomly chosen pod
-	// via its headless DNS name so routing survives proxy-runner restarts.
-	// When nil (single-replica or non-Kubernetes), the static targetURI is used.
+	// When set, new sessions are pinned to a specific pod via headless DNS so routing
+	// survives proxy-runner restarts. Nil for non-Kubernetes deployments.
 	headlessService *podHeadlessService
 }
 
@@ -240,14 +239,14 @@ func withShutdownTimeout(timeout time.Duration) Option {
 	}
 }
 
-// WithPodHeadlessService configures pod-specific routing for multi-replica StatefulSet deployments.
-// When set, each new MCP session is pinned to a randomly selected pod via its headless DNS name
+// WithPodHeadlessService configures pod-specific routing for Kubernetes StatefulSet deployments.
+// When set, each new MCP session is pinned to a specific pod via its headless DNS name
 // (e.g. myserver-0.mcp-myserver-headless.default.svc.cluster.local) so that session routing
-// survives proxy-runner restarts without being sent to the wrong pod.
-// The option is a no-op when replicas <= 1 or when any required field is empty.
+// survives proxy-runner restarts. For single-replica StatefulSets, ordinal 0 is always selected.
+// The option is a no-op when any required field is empty.
 func WithPodHeadlessService(statefulSetName, serviceName, namespace string, replicas int32) Option {
 	return func(p *TransparentProxy) {
-		if statefulSetName == "" || serviceName == "" || namespace == "" || replicas <= 1 {
+		if statefulSetName == "" || serviceName == "" || namespace == "" {
 			return
 		}
 		p.headlessService = &podHeadlessService{
--- a/cmd/thv-operator/controllers/mcpserver_runconfig.go
+++ b/cmd/thv-operator/controllers/mcpserver_runconfig.go
@@ -306,17 +306,16 @@ func populateScalingConfig(runConfig *runner.RunConfig, m *mcpv1alpha1.MCPServer
 		val := *m.Spec.BackendReplicas
 		runConfig.ScalingConfig.BackendReplicas = &val
 
-		// Populate headless service config when there are multiple backend replicas.
+		// Always populate headless service config when BackendReplicas is set.
 		// This enables the proxy runner to route each session to a specific pod via
 		// headless DNS (e.g. myserver-0.mcp-myserver-headless.default.svc.cluster.local)
-		// so sessions survive proxy-runner restarts without hitting the wrong backend pod.
-		if val > 1 {
-			runConfig.ScalingConfig.HeadlessService = &runner.HeadlessServiceConfig{
-				StatefulSetName: m.Name,
-				ServiceName:     fmt.Sprintf("mcp-%s-headless", m.Name),
-				Namespace:       m.Namespace,
-				Replicas:        val,
-			}
+		// so sessions survive proxy-runner restarts. For single-replica StatefulSets,
+		// ordinal 0 is always selected deterministically.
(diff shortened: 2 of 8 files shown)


pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -223,6 +240,25 @@ func withShutdownTimeout(timeout time.Duration) Option {
...
+			return
+		}
+		p.headlessService = &podHeadlessService{
+			statefulSetName: statefulSetName,
+			serviceName:     serviceName,
+			namespace:       namespace,
+			replicas:        replicas,
+		}
+	}
+}
```
jerm-dro, line 259: `blocker:` Make headless DNS routing the default for all Kubernetes deployments, and remove the `headlessSetter` type assertion pattern.

Two related issues:

**1. Headless routing should be unconditional for Kubernetes.** For a single-replica StatefulSet, `myserver-0.mcp-myserver-headless.ns.svc.cluster.local` is deterministic (always ordinal 0) and the headless service already exists because the operator creates it. This would eliminate the `if val > 1` guard in `populateScalingConfig`, remove the `replicas <= 1` no-op here, and give you one routing code path instead of two. `pickPodBackendURL` degenerates correctly — `rand.Int(rand.Reader, big.NewInt(1))` always returns 0. Non-Kubernetes deployments are unaffected since `HeadlessServiceConfig` stays nil.

**2. The `headlessSetter` anonymous interface in `factory.go` is unsafe.** If a transport doesn't implement the unexported method, the option silently no-ops — no compile-time check, no runtime warning, no error returned. This means a misconfiguration (e.g. applying the option to a transport type that doesn't support it) is invisible. Put this on `types.Config` like `SessionStorage`, so it flows through the same plumbing and doesn't rely on type assertions against unexported methods.

yrobla:
closed because the approach of the original issue is under review

JAORMX (changes requested):
So, there's a fundamental issue with the headless service that this PR builds on top of.

The headless service (`mcp-<name>-headless`) selects the MCP server backend pods directly via `app: containerName`. That means it points to the StatefulSet pods... which are the raw MCP backends, not the proxy runner. The proxy runner is a Deployment with different labels.

This matters because:

1. **For stdio transports, backend pods don't listen on the network at all.** The `transportTypeRequiresBackendServices` guard skips service creation for stdio, but the StatefulSet's `spec.serviceName` still references a service that doesn't exist in that case.

2. **The headless service exposes backend pods directly to the cluster**, bypassing the proxy's auth, session management, and routing logic. Anyone in the namespace can resolve `pod-0.mcp-myserver-headless.ns.svc.cluster.local` and hit the backend without going through the proxy.

3. **The random pod selection happens AFTER the initialize request already went through the ClusterIP.** So the initialize lands on pod X (via ClusterIP load balancing), but `pickPodBackendURL()` stores pod Y (random headless DNS) as `backend_url`. Follow-up requests go to pod Y, which never saw the initialize... and you get `-32001 session not found`. The copilot reviewer flagged this too.

Before building more on top of the headless service, I think we need to step back and reconsider whether this is the right mechanism for session pinning. The headless service was created before this PR but was never actually wired up (the StatefulSet `serviceName` didn't even match it). We should figure out the right approach before adding more plumbing on top of a broken foundation.

I'm going to open a separate issue to discuss removing or rethinking the headless service.

JAORMX:
Opened #4668 to track the headless service cleanup/rethink separately.

PR #4645, opened 2026-04-07
Remove in-memory cache from EncryptedManager

amirejaz:
## Summary

The `EncryptedManager` loaded secrets from disk once at construction time and served all `GetSecret` reads from an in-memory `syncmap.Map`. This meant any write or delete performed by a separate process after construction was invisible to a long-running process — the in-memory snapshot never updated.

- Remove the `syncmap.Map` cache entirely; `GetSecret` now reads and decrypts the file on every call, consistent with all other `Provider` implementations (`environment`, `keyring`) which already read live on every access
- `SetSecret` and `DeleteSecret` already re-read the file inside the write lock; remove the now-unnecessary cache update calls after each write
- `ListSecrets` and `Cleanup` updated to go through `readFileSecrets`/`writeFileSecrets` instead of the removed map
- `NewEncryptedManager` validates the file is readable at startup but no longer populates a cache

## Type of change

- [x] Bug fix

## Test plan

No.

The `syncmap.Map` was introduced for performance (nanosecond in-memory reads vs. a file decrypt per call). For a secrets file that is a few KB, AES-GCM decryption is fast enough that the per-call overhead is negligible compared to the correctness benefit. All other `Provider` implementations already pay this cost on every read.

### DIFF
--- a/pkg/secrets/encrypted.go
+++ b/pkg/secrets/encrypted.go
@@ -13,8 +13,6 @@ import (
 	"path"
 	"strings"
 
-	"golang.org/x/sync/syncmap"
-
 	"github.com/stacklok/toolhive/pkg/fileutils"
 	"github.com/stacklok/toolhive/pkg/secrets/aes"
 )
@@ -24,8 +22,7 @@ import (
 type EncryptedManager struct {
 	filePath string
 	// Key used to re-encrypt the secrets file if changes are needed.
-	key     []byte
-	secrets syncmap.Map // Thread-safe map for storing secrets
+	key []byte
 }
 
 // fileStructure is the structure of the secrets file.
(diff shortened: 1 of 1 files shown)


reyortiz3:
are there any concerns with increased I/O now? 

> This meant any write or delete performed by a separate process after construction was invisible to a long-running process 

So, it wasn't possible to update the cache on write/delete then?

amirejaz:
Good question! Same-process writes and deletes **did** update the cache — `SetSecret` and `DeleteSecret` called `e.secrets.Store`/`e.secrets.Delete` after writing to disk, so within a single process the cache stayed consistent. The problem was strictly cross-process: when the CLI writes `keySessionAccess` to disk, the running server has no way to know — the `syncmap` only reflects writes that went through that same process instance.

On the I/O concern: the increase is real, but the impact is minimal in practice. The secrets file is tiny (a handful of keys, a few KB), so the OS page cache keeps it in memory after the first access — actual disk reads are rare. The cost is dominated by syscall overhead, roughly microseconds on macOS. For the session auth use case this is one file read per HTTP request to the local API server, which is acceptable.

Alternatives like file watching (inotify/kqueue) or mtime-based cache invalidation would preserve caching while fixing cross-process visibility, but add platform-specific complexity. Given the file is tiny and stays in the page cache, reading on every `GetSecret` is the simpler and correct choice — the correctness win for cross-process logout signals clearly outweighs the cost.

reyortiz3:
Thanks for the reply! It makes sense, if any real issue is observed we can implement mtime-based cache invalidation later, but agree with this. Thanks

PR #4648, opened 2026-04-07
fix(telemetry): warn instead of crashing when OTLP endpoint has no enabled exporters

### LINKED ISSUE #4647: thv serve crashes on startup when OTLP endpoint is configured but tracing/metrics are disabled
## Summary

`thv serve` fails to start with exit code 1 when an OTLP endpoint is configured via `thv config otel set-endpoint` but both tracing and metrics are disabled. This causes ToolHive Studio to show a "Health check failed" error since the server never comes up.

## Error

```
Error: failed to initialize telemetry: OTLP endpoint is configured but both tracing and metrics are disabled; either enable tracing or metrics, or remove the endpoint
```

## Steps to Reproduce

1. Configure an OTLP endpoint: `thv config otel set-endpoint https://some-collector.example.com`
2. Do **not** enable tracing or metrics
3. Run `thv serve`
4. Process exits immediately with code 1

## Expected Behavior

`thv serve` should start successfully, logging a warning that the OTLP endpoint is configured but unused (since tracing and metrics are both disabled). The endpoint configuration should be silently ignored rather than treated as a fatal error.

## Workaround

```bash
thv config otel unset-endpoint
```

## Impact

This is a fatal error that prevents the server from starting entirely. Users of [ToolHive Studio hit](https://github.com/stacklok/toolhive-studio/issues/1932) this as a blank "Health check failed" screen with no clear indication of the root cause — the error is only visible in the main process logs.

peppescg:
## Summary
- When an OTLP endpoint is configured (`thv config otel set-endpoint`) but both tracing and metrics are disabled, `thv serve` crashes with a fatal validation error (exit code 1)
- This causes ToolHive Studio to show a "Health check failed" error with no clear indication of the root cause
- Instead of crashing, we now **enable tracing by default** when an endpoint is configured but no exporters are enabled, so the endpoint is actually used rather than silently ignored

## Test plan
- [x] Configured an OTLP endpoint (`thv config otel set-endpoint localhost:4317`) with tracing/metrics not enabled
- [x] Verified `thv serve` starts with a warning and tracing enabled by default, instead of crashing
- [x] Verified existing telemetry tests pass (`go test ./pkg/telemetry/...`)
- [ ] Verify normal OTLP flow still works when tracing or metrics are explicitly enabled

Fixes #4647

### DIFF
--- a/pkg/telemetry/serve.go
+++ b/pkg/telemetry/serve.go
@@ -78,9 +78,9 @@ func NewServeProvider(ctx context.Context) (provider *Provider, otelEnabled bool
 	return p, true, nil
 }
 
-// handleUnusedEndpoint clears the OTLP endpoint when both tracing and metrics
-// are disabled, so the server can start normally instead of crashing with a
-// fatal validation error.
+// handleUnusedEndpoint enables tracing by default when an OTLP endpoint is
+// configured but both tracing and metrics are disabled, so the server can start
+// normally instead of crashing with a fatal validation error.
 func handleUnusedEndpoint(otelCfg *config.OpenTelemetryConfig) {
 	if otelCfg.Endpoint == "" {
 		return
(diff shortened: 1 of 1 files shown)


pkg/telemetry/serve.go
```diff
@@ -74,3 +77,19 @@ func NewServeProvider(ctx context.Context) (provider *Provider, otelEnabled bool
...
+func handleUnusedEndpoint(otelCfg *config.OpenTelemetryConfig) {
+	if otelCfg.Endpoint == "" {
+		return
+	}
+	tracingOff := otelCfg.TracingEnabled == nil || !*otelCfg.TracingEnabled
+	metricsOff := otelCfg.MetricsEnabled == nil || !*otelCfg.MetricsEnabled
+	if tracingOff && metricsOff {
+		slog.Warn("OTLP endpoint is configured but tracing and metrics are both disabled; ignoring endpoint",
+			"endpoint", otelCfg.Endpoint)
+		otelCfg.Endpoint = ""
```
mauriciocoder: Why is it necessary to clear the endpoint in this situation?

peppescg: Changed the logic in the latest commit, instead of clearing the endpoint, we now enable tracing by default when an endpoint is provided. This way the endpoint is actually used and users don't need to explicitly run set-tracing-enabled

PR #4662, opened 2026-04-08
Add MCPServerEntry CRD types and MCPGroup status fields

JAORMX:
## Summary

VirtualMCPServer currently requires MCPRemoteProxy (which spawns proxy pods) to reach remote MCP servers. This forces OIDC auth on public remotes, creates dual auth boundary confusion, and wastes resources. RFC-55 introduces MCPServerEntry as a zero-infrastructure catalog entry — pure configuration telling vMCP where a remote server exists and how to authenticate.

This PR introduces the CRD types for MCPServerEntry and adds Entries/EntryCount tracking to MCPGroupStatus. It is the first of three PRs implementing #4656 (the controller and MCPGroup integration follow in subsequent PRs).

- Define `MCPServerEntrySpec` with remoteURL, transport, groupRef, externalAuthConfigRef, headerForward, and caBundleRef fields
- Define `MCPServerEntryStatus` with Valid/Pending/Failed phase model and GroupRefValid, AuthConfigValid, CABundleValid conditions (using "Valid" semantics since this is a configuration resource, not a workload)
- Add `Entries` and `EntryCount` fields to `MCPGroupStatus` for tracking MCPServerEntry membership
- Register MCPServerEntry in the CRD Helm wrapper feature flags map
- Generate CRD manifests, Helm templates, DeepCopy, and API reference docs

Refs #4656

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — operator unit tests pass
- [x] Linting (`task lint-fix`) — 0 issues
- [x] Code generation verified (`task operator-generate`, `task operator-manifests`, `task crdref-gen`)

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go` | New CRD types: spec, status, phase enum, condition constants |
| `cmd/thv-operator/api/v1alpha1/mcpgroup_types.go` | Add Entries/EntryCount status fields + printer column |
| `deploy/charts/operator-crds/crd-helm-wrapper/main.go` | Add `mcpserverentries` to feature flags map |
| `cmd/thv-operator/api/v1alpha1/zz_generated.deepcopy.go` | Generated DeepCopy methods |
| `deploy/charts/operator-crds/files/crds/` | Generated CRD manifests |
| `deploy/charts/operator-crds/templates/` | Generated Helm-wrapped CRD templates |
| `docs/operator/crd-api.md` | Regenerated API reference docs |

## Special notes for reviewers

[...]

Issue #4656, opened 2026-04-08
RFC-55: MCPServerEntry CRD — Types, Validation Controller, and MCPGroup Integration


JAORMX:
**As a** platform engineer managing remote MCP server backends,
**I want** a lightweight `MCPServerEntry` CRD that declares remote MCP endpoints without deploying infrastructure,
**so that** VirtualMCPServer can discover and route to remote backends via groupRef without unnecessary proxy pods.

**Size**: L
**Dependencies**: None
**Labels**: `operator`, `api`, `kubernetes`
**RFC**: [RFC-0055](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0055-mcpserverentry-direct-remote-backends.md)
**Epic**: stacklok/stacklok-epics#291

## Context

VirtualMCPServer currently requires MCPRemoteProxy (which spawns `thv-proxyrunner` pods) to reach remote MCP servers. This forces OIDC auth on public remotes, creates dual auth boundary confusion, and wastes resources deploying pods just to make HTTP calls.

RFC-0055 introduces `MCPServerEntry` as a zero-infrastructure catalog entry — pure configuration telling vMCP where a remote server exists and how to authenticate to it. MCPServerEntry is an intermediate step that will eventually be superseded by `MCPRemoteEndpoint` (RFC-0057, #4254).

This story covers the CRD types, validation-only controller, and MCPGroup integration.

### CRD Types (`cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go`)

- `MCPServerEntrySpec` with fields:
  - `remoteURL` (string, required) — HTTPS enforced by default, annotation override for development
  - `transport` (enum: `streamable-http` | `sse`, required)
  - `groupRef` (string, required — unlike MCPServer where it's optional)
  - `externalAuthConfigRef` (*ExternalAuthConfigRef, optional) — reuse existing type
  - `headerForward` (*HeaderForwardConfig, optional) — reuse existing type from MCPRemoteProxy
  - `caBundleRef` (*CABundleSource, optional) — reuse existing `CABundleSource` with `ConfigMapKeyRef`
- `MCPServerEntryStatus` with:
  - `Phase` (Pending/Ready/Failed)
  - `Conditions` (Ready, GroupRefValid, AuthConfigValid, CABundleValid)
  - `ObservedGeneration`
- CEL validation rules:
[...]

PR #4664, opened 2026-04-08
Add MCPServerEntry validation controller and MCPGroup integration

JAORMX:
## Summary

MCPServerEntry needs a controller to validate that referenced resources exist and to integrate with MCPGroup for discovery. This is the second PR in the RFC-55 implementation series, building on #4662 (CRD types).

- Introduce a validation-only controller that checks MCPGroup, MCPExternalAuthConfig, and CA bundle ConfigMap references exist, setting GroupRefValidated, ExternalAuthConfigValidated, and CABundleRefValidated conditions
- The controller never creates infrastructure (no pods, services, deployments) and never probes remote URLs
- Integrate MCPServerEntry into the MCPGroup controller: track membership in Entries/EntryCount, handle deletion notifications, watch for entry changes
- Wire field indexing for MCPServerEntry.Spec.GroupRef and register the controller in the operator main setup
- Watch MCPExternalAuthConfig, MCPGroup, and ConfigMap changes to re-validate entries when referenced resources are created/updated/deleted
- Use `Failed` phase (not `Pending`) when validations fail, matching the semantic distinction documented in the types

Refs #4656

## Type of change

- [x] New feature

## Large PR Justification

- The diff is inflated by ~710 lines because #4676 accidentally deleted `mcpserverentry_types.go` and both CRD YAML files that #4662 had already merged — those appear as new additions here but are restorations, not new code
- Excluding tests (538 lines), generated RBAC/CRD manifests (544 lines), and docs (104 lines), the actual new production code is ~427 lines (the controller) + ~127 lines (MCPGroup integration) + ~38 lines (main.go wiring)
- The controller + MCPGroup integration is a single logical change that cannot be split without shipping a controller with no tests or an MCPGroup integration with no controller

## Test plan

- [x] Unit tests (`task test`) — operator unit tests pass, including 8 table-driven subtests for MCPServerEntryReconciler and MCPGroup entry integration/deletion tests
- [x] RBAC regenerated via `task operator-manifests`

## Changes

| File | Change |
|------|--------|
| `mcpserverentry_types.go` | Remove unused AllowInsecureAnnotation, align condition type names with shared constants |
| `mcpserverentry_controller.go` | New validation-only controller: validates GroupRef (existence + Ready phase), AuthConfig, CABundle; watches MCPGroup, ConfigMap, and AuthConfig; field-indexed lookups; transient errors requeue instead of persisting misleading conditions |
| `mcpgroup_controller.go` | Add MCPServerEntry integration: finder, populator, deletion handler, watch |
| `main.go` | Extract `setupGroupRefFieldIndexes`, add MCPServerEntry controller setup |
| `mcpserverentry_controller_test.go` | 8 table-driven subtests + MCPGroup entry integration and deletion handler tests |
| `mcpgroup_controller_test.go` | Add MCPServerEntry field index to all fake client builders |
| `test-integration/*/suite_test.go` | Fix field indexer nesting bug (was inside MCPRemoteProxy error block) in 5 suites |
| `role.yaml`, CRD YAMLs, `crd-api.md` | Generated RBAC and CRD updates |
| `chainsaw/*/assert-rbac-clusterrole.yaml` | Add MCPServerEntry to RBAC assertions |

- The controller has no finalizer since it owns no external resources. MCPGroup handles the "member disappeared" case via its existing watch pattern.
- Transient errors (non-NotFound) from `Get` calls are returned to controller-runtime for requeue rather than persisting a potentially misleading condition. Detailed errors are logged server-side only.
- SSRF-style URL validation for `RemoteURL` is tracked separately in #4695 — the existing `^https?://` CEL pattern matches what MCPRemoteProxy already uses today.

### DIFF
--- a/test/e2e/chainsaw/operator/multi-tenancy/setup/assert-rbac-clusterrole.yaml
+++ b/test/e2e/chainsaw/operator/multi-tenancy/setup/assert-rbac-clusterrole.yaml
@@ -149,18 +149,19 @@ rules:
 - apiGroups:
   - toolhive.stacklok.dev
   resources:
+  - mcpserverentries
+  - virtualmcpcompositetooldefinitions
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - toolhive.stacklok.dev
+  resources:
   - mcptelemetryconfigs
   verbs:
   - get
   - list
   - patch
   - update
-  - watch
-- apiGroups:
-  - toolhive.stacklok.dev
-  resources:
-  - virtualmcpcompositetooldefinitions
-  verbs:
-  - get
-  - list
   - watch
--- a/test/e2e/chainsaw/operator/single-tenancy/setup/assert-rbac-clusterrole.yaml
+++ b/test/e2e/chainsaw/operator/single-tenancy/setup/assert-rbac-clusterrole.yaml
@@ -104,7 +104,6 @@ rules:
   - mcpoidcconfigs
   - mcpregistries
   - mcpremoteproxies
-  - mcpserverentries
   - mcpservers
   - mcptoolconfigs
   - virtualmcpservers
@@ -150,18 +149,19 @@ rules:
 - apiGroups:
   - toolhive.stacklok.dev
   resources:
+  - mcpserverentries
+  - virtualmcpcompositetooldefinitions
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - toolhive.stacklok.dev
+  resources:
   - mcptelemetryconfigs
   verbs:
   - get
   - list
   - patch
   - update
-  - watch
-- apiGroups:
-  - toolhive.stacklok.dev
-  resources:
-  - virtualmcpcompositetooldefinitions
-  verbs:
-  - get
-  - list
   - watch


cmd/thv-operator/controllers/mcpserverentry_controller.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+		meta.SetStatusCondition(&entry.Status.Conditions, metav1.Condition{
+			Type:               mcpv1alpha1.ConditionTypeMCPServerEntryGroupRefValid,
+			Status:             metav1.ConditionFalse,
+			Reason:             reason,
+			Message:            message,
+			ObservedGeneration: entry.Generation,
+		})
+		return false
+	}
```
jhrozek, line 189: This only checks that the MCPGroup exists, but not that it's in `Ready` phase. Both the MCPServer and MCPRemoteProxy controllers check `group.Status.Phase != MCPGroupPhaseReady` here and set a `GroupRefNotReady` reason — and the types file already defines `ConditionReasonMCPServerEntryGroupRefNotReady` for this, so it looks like the intent was there.

Should be a straightforward addition after the successful `Get`.

cmd/thv-operator/controllers/mcpserverentry_controller.go
```diff
@@ -0,0 +1,283 @@
...
+// SetupWithManager sets up the controller with the Manager.
+func (r *MCPServerEntryReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&mcpv1alpha1.MCPServerEntry{}).
+		Watches(
+			&mcpv1alpha1.MCPExternalAuthConfig{},
+			handler.EnqueueRequestsFromMapFunc(r.findEntriesForAuthConfig),
+		).
+		Complete(r)
+}
```
jhrozek, line 152: This watches `MCPExternalAuthConfig` changes but not `MCPGroup` or `ConfigMap` (CA bundle). Since this is a validation-only controller, stale conditions are its main failure mode.

If an MCPGroup or CA bundle ConfigMap is created after the MCPServerEntry, the entry stays `Pending` forever — unlike MCPServer/MCPRemoteProxy which have secondary triggers from Deployment changes, there's nothing else to re-trigger reconciliation here.

Worth adding watches for both, or at minimum a periodic requeue as fallback.

cmd/thv-operator/controllers/mcpserverentry_controller.go
```diff
@@ -0,0 +1,283 @@
...
+)
+
+// MCPServerEntryReconciler reconciles a MCPServerEntry object.
+// This is a validation-only controller — it never creates infrastructure
+// (no Deployment, Service, or Pod) and never probes remote URLs.
+type MCPServerEntryReconciler struct {
+	client.Client
+}
+
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpserverentries,verbs=get;list;watch;create;update;patch;delete
```
jhrozek: The controller only does `Get`, `List`, and `Status().Update()` — it never creates, updates, patches, or deletes MCPServerEntry objects. The RBAC marker here grants full CRUD though. Compare with the MCPGroup controller on line 41 which correctly uses `get;list;watch` for the same resource.

Should be:
```
// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpserverentries,verbs=get;list;watch
```
Then re-run `task operator-manifests`.

cmd/thv-operator/controllers/mcpserverentry_controller.go
```diff
@@ -0,0 +1,283 @@
...
+// Returns true if the validation passed.
+func (r *MCPServerEntryReconciler) validateGroupRef(
+	ctx context.Context,
+	entry *mcpv1alpha1.MCPServerEntry,
+) bool {
+	group := &mcpv1alpha1.MCPGroup{}
+	groupKey := types.NamespacedName{Namespace: entry.Namespace, Name: entry.Spec.GroupRef}
+
+	if err := r.Get(ctx, groupKey, group); err != nil {
+		reason := mcpv1alpha1.ConditionReasonMCPServerEntryGroupRefNotFound
```
jhrozek: When `r.Get` fails with a transient error (timeout, RBAC issue), the reason is still set to `GroupRefNotFound` even though it's not actually a 404. The message differentiates but the machine-readable reason doesn't. Same thing in the other two validate methods.

Might be worth either adding a separate reason for transient failures, or returning the error to force a requeue instead of persisting a potentially misleading condition.

cmd/thv-operator/controllers/mcpserverentry_controller.go
```diff
@@ -0,0 +1,283 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
jhrozek, line 1: No unit tests for this controller. It's 284 lines with non-trivial validation logic (GroupRef, ExternalAuthConfigRef, CABundleRef, phase transitions, watch mapping). The analogous MCPRemoteProxy controller has 14 test functions.

At minimum I'd expect table-driven tests covering: happy path (all valid), each ref missing, optional refs being nil, not-found entry, and status update conflict.

cmd/thv-operator/controllers/mcpgroup_controller.go
```diff
@@ -292,6 +328,22 @@ func (r *MCPGroupReconciler) findReferencingMCPRemoteProxies(
 	return mcpRemoteProxyList.Items, nil
 }
 
+// findReferencingMCPServerEntries finds all MCPServerEntries that reference the given MCPGroup
+func (r *MCPGroupReconciler) findReferencingMCPServerEntries(
```
jhrozek, line 332: The four new MCPServerEntry methods (`findReferencingMCPServerEntries`, `populateEntryStatus`, `findMCPGroupForMCPServerEntry`, `updateReferencingEntriesOnDeletion`) don't have tests. The analogous MCPRemoteProxy methods all do — e.g., `TestMCPGroupReconciler_findReferencingMCPRemoteProxies`, `TestMCPGroupReconciler_findMCPGroupForMCPRemoteProxy`, etc.

Also none of the existing reconciliation tests create MCPServerEntry objects, so `Status.EntryCount`/`Status.Entries` are never asserted nonzero, and the deletion cleanup path for entries is exercised only with an empty list.

jhrozek (commented):
Review of the controller and MCPGroup integration. Types-level comments on #4662.

cmd/thv-operator/test-integration/mcp-oidc-config/suite_test.go
```diff
@@ -121,6 +121,21 @@ var _ = BeforeSuite(func() {
 		return []string{mcpRemoteProxy.Spec.GroupRef}
 	}); err != nil {
 		Expect(err).ToNot(HaveOccurred())
+
+		// Set up field indexing for MCPServerEntry.Spec.GroupRef
+		err = k8sManager.GetFieldIndexer().IndexField(
+			context.Background(),
```
ChrisJBurns: **[CRITICAL]** *Found by: kubernetes-expert*

**Bug: MCPServerEntry field indexer is nested inside the MCPRemoteProxy error handler — it will never execute on the happy path.**

The MCPServerEntry field indexer registration (lines 125-140) is placed inside the `if err != nil` block for the MCPRemoteProxy indexer (line 122). This means the MCPServerEntry indexer only runs when the MCPRemoteProxy indexer *fails*. In the normal (success) path, it is skipped entirely.

The same bug exists in all four integration test suites:
- `mcp-oidc-config/suite_test.go` (here)
- `mcp-remote-proxy/suite_test.go`
- `mcp-server/suite_test.go`
- `virtualmcp/suite_test.go`

Note: `mcp-group/suite_test.go` has it correct — the indexer is a separate statement outside the `if err` block.

**Fix:** Close the `if err` block (line 123) *before* the MCPServerEntry indexer, so both run unconditionally:

```go
}); err != nil {
    Expect(err).ToNot(HaveOccurred())
}

// Set up field indexing for MCPServerEntry.Spec.GroupRef
err = k8sManager.GetFieldIndexer().IndexField(...)
```

cmd/thv-operator/test-integration/mcp-remote-proxy/suite_test.go
```diff
@@ -123,6 +123,21 @@ var _ = BeforeSuite(func() {
 		return []string{mcpRemoteProxy.Spec.GroupRef}
 	}); err != nil {
 		Expect(err).ToNot(HaveOccurred())
+
+		// Set up field indexing for MCPServerEntry.Spec.GroupRef
```
ChrisJBurns: **[CRITICAL]** *Found by: kubernetes-expert*

Same nesting bug as `mcp-oidc-config/suite_test.go` — the MCPServerEntry field indexer is inside the `if err != nil` block for the MCPRemoteProxy indexer and will never run on the happy path. See the comment on that file for the fix.

cmd/thv-operator/test-integration/mcp-server/suite_test.go
```diff
@@ -123,6 +123,21 @@ var _ = BeforeSuite(func() {
 		return []string{mcpRemoteProxy.Spec.GroupRef}
 	}); err != nil {
 		Expect(err).ToNot(HaveOccurred())
+
+		// Set up field indexing for MCPServerEntry.Spec.GroupRef
```
ChrisJBurns: **[CRITICAL]** *Found by: kubernetes-expert*

Same nesting bug — MCPServerEntry field indexer unreachable on happy path. See `mcp-oidc-config/suite_test.go` comment for details.

cmd/thv-operator/test-integration/virtualmcp/suite_test.go
```diff
@@ -123,6 +123,21 @@ var _ = BeforeSuite(func() {
 		return []string{mcpRemoteProxy.Spec.GroupRef}
 	}); err != nil {
 		Expect(err).ToNot(HaveOccurred())
+
+		// Set up field indexing for MCPServerEntry.Spec.GroupRef
```
ChrisJBurns: **[CRITICAL]** *Found by: kubernetes-expert*

Same nesting bug — MCPServerEntry field indexer unreachable on happy path. See `mcp-oidc-config/suite_test.go` comment for details.

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
```diff
@@ -0,0 +1,173 @@
...
+	ConditionReasonMCPServerEntryCABundleValid = "CABundleValid"
+
+	// ConditionReasonMCPServerEntryCABundleNotFound indicates the referenced CA bundle ConfigMap was not found.
+	ConditionReasonMCPServerEntryCABundleNotFound = "CABundleNotFound"
+
+	// ConditionReasonMCPServerEntryCABundleNotConfigured indicates no CA bundle ref is set.
+	ConditionReasonMCPServerEntryCABundleNotConfigured = "CABundleNotConfigured"
+)
+
+// AllowInsecureAnnotation is the annotation key to allow HTTP URLs for development.
```
ChrisJBurns: **[CRITICAL]** *Found by: kubernetes-expert, code-reviewer, security-advisor*

All three agents flagged this: `AllowInsecureAnnotation` is defined and documented in the `RemoteURL` field comment (lines 17-18: "HTTPS is enforced by default; set the annotation...to allow HTTP for development"), but:

1. The kubebuilder pattern `^https?://` (line 20) already allows both HTTP and HTTPS at admission time
2. The controller never reads or checks this annotation

This creates a **false sense of security** — users who read the docs will believe HTTPS is enforced when it is not. HTTP URLs are silently accepted regardless of the annotation.

**Options:**
- **(a)** Implement the check in the controller: reject HTTP URLs unless the annotation is `"true"`, and set a warning condition when the annotation is present
- **(b)** Remove the annotation constant and update the RemoteURL doc comment to accurately reflect that URL scheme enforcement is not in scope for this PR
- **(c)** Document it as a known TODO for a follow-up PR, but update the doc comment to not mislead users in the interim

cmd/thv-operator/controllers/mcpserverentry_controller.go
ChrisJBurns: **[MUST FIX]** *Found by: kubernetes-expert, security-advisor*

This RBAC marker grants `create;update;patch;delete` on the main `mcpserverentries` resource, but the controller is validation-only — it only reads entries and updates their status subresource. It never creates, updates, patches, or deletes MCPServerEntry resources.

Compare with the MCPGroup controller's RBAC for `mcpserverentries` which correctly uses only `get;list;watch`.

```suggestion
// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpserverentries,verbs=get;list;watch
```

After changing this, regenerate RBAC with `task operator-manifests`.

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
```diff
@@ -0,0 +1,173 @@
...
+type MCPServerEntryPhase string
+
+const (
+	// MCPServerEntryPhaseValid indicates all validations passed and the entry is usable.
+	MCPServerEntryPhaseValid MCPServerEntryPhase = "Valid"
+
+	// MCPServerEntryPhasePending indicates the entry is waiting for referenced resources.
+	MCPServerEntryPhasePending MCPServerEntryPhase = "Pending"
+
+	// MCPServerEntryPhaseFailed indicates a validation error occurred.
```
ChrisJBurns: **[SHOULD FIX]** *Found by: kubernetes-expert*

`MCPServerEntryPhaseFailed` is defined here but never set by the controller. The controller's `updateOverallStatus` only sets `Valid` or `Pending`. Using `Pending` when validations fail contradicts this comment ("a validation error occurred").

`Pending` implies eventual automatic resolution, but a missing GroupRef may be a permanent user error.

Either:
- Use `Failed` in the controller when validations fail (and `Pending` only for initial state before first reconcile)
- Remove `Failed` if it's not needed
- Add a doc comment clarifying the distinction (e.g., "Pending = refs missing, may self-heal; Failed = reserved for permanent spec-level errors")

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
```diff
@@ -0,0 +1,173 @@
...
+	ConditionReasonMCPServerEntryInvalid = "ConfigInvalid"
+
+	// ConditionReasonMCPServerEntryGroupRefValid indicates the referenced MCPGroup exists.
+	ConditionReasonMCPServerEntryGroupRefValid = "GroupRefIsValid"
+
+	// ConditionReasonMCPServerEntryGroupRefNotFound indicates the referenced MCPGroup was not found.
+	ConditionReasonMCPServerEntryGroupRefNotFound = "GroupRefNotFound"
+
+	// ConditionReasonMCPServerEntryGroupRefNotReady indicates the referenced MCPGroup is not Ready.
+	ConditionReasonMCPServerEntryGroupRefNotReady = "GroupRefNotReady"
```
ChrisJBurns: **[SHOULD FIX]** *Found by: kubernetes-expert*

`ConditionReasonMCPServerEntryGroupRefNotReady` is defined but never used — the controller only checks existence (found/not found), not the group's Ready state.

Either remove the unused constant, or add a TODO comment if checking group readiness is planned for a follow-up.

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+// Condition types for MCPServerEntry.
+const (
+	// ConditionTypeMCPServerEntryValid indicates overall validation status of the MCPServerEntry.
+	// Uses the shared "Valid" condition type since this is a configuration resource, not a workload.
+	ConditionTypeMCPServerEntryValid = ConditionTypeValid
+
+	// ConditionTypeMCPServerEntryGroupRefValid indicates whether the referenced MCPGroup exists.
+	ConditionTypeMCPServerEntryGroupRefValid = "GroupRefValid"
```
ChrisJBurns: **[SHOULD FIX]** *Found by: kubernetes-expert*

Condition type naming is inconsistent with existing CRDs:

| Concept | MCPServerEntry | MCPServer / MCPRemoteProxy |
|---------|---------------|---------------------------|
| Group ref valid | `"GroupRefValid"` | `"GroupRefValidated"` |
| Auth config valid | `"AuthConfigValid"` | `"ExternalAuthConfigValidated"` |
| CA bundle valid | `"CABundleValid"` | `"CABundleRefValidated"` |

This means the same semantic concept ("is the group reference valid?") has different condition type strings across resource types, complicating kubectl queries, dashboards, and the MCPGroup controller's deletion handler (which writes `GroupRefValid` on entries vs `GroupRefValidated` on servers).

Consider reusing the shared constants from `mcpserver_types.go` for consistency, or document why MCPServerEntry intentionally diverges.

cmd/thv-operator/controllers/mcpserverentry_controller.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *MCPServerEntryReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&mcpv1alpha1.MCPServerEntry{}).
+		Watches(
+			&mcpv1alpha1.MCPExternalAuthConfig{},
+			handler.EnqueueRequestsFromMapFunc(r.findEntriesForAuthConfig),
+		).
```
ChrisJBurns, line 142: **[SHOULD FIX]** *Found by: kubernetes-expert*

The controller watches `MCPExternalAuthConfig` changes to re-validate entries, but does **not** watch `ConfigMap` changes for CA bundle re-validation. This means:

- If a CA bundle ConfigMap is deleted after the entry is validated → entry stays `Valid` with stale conditions
- If a CA bundle ConfigMap is created after the entry was `Pending` → entry won't automatically recover

Since MCPServerEntry's entire purpose is validation, a stale CA bundle status is a functional gap. Consider adding a `Watches` for `corev1.ConfigMap` with a map function similar to `findEntriesForAuthConfig`, or document this as a known limitation with a TODO.

cmd/thv-operator/controllers/mcpserverentry_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	entry *mcpv1alpha1.MCPServerEntry,
+) bool {
+	group := &mcpv1alpha1.MCPGroup{}
+	groupKey := types.NamespacedName{Namespace: entry.Namespace, Name: entry.Spec.GroupRef}
+
+	if err := r.Get(ctx, groupKey, group); err != nil {
+		reason := mcpv1alpha1.ConditionReasonMCPServerEntryGroupRefNotFound
+		message := "Referenced MCPGroup not found"
+		if !errors.IsNotFound(err) {
+			message = "Failed to get referenced MCPGroup: " + err.Error()
```
ChrisJBurns: **[MEDIUM]** *Found by: security-advisor*

For non-NotFound errors, the raw `err.Error()` is included in the condition message: `"Failed to get referenced MCPGroup: " + err.Error()`. API server errors could contain internal details (network topology, internal service names, etc.) that are then visible to anyone who can read the MCPServerEntry status.

The same pattern appears in `validateExternalAuthConfigRef` (line 151) and `validateCABundleRef` (line 200).

Consider using a generic message for non-NotFound errors and logging the detailed error server-side only:
```go
if !errors.IsNotFound(err) {
    ctxLogger.Error(err, "Failed to get referenced MCPGroup")
    message = "Failed to get referenced MCPGroup"
}
```

cmd/thv-operator/controllers/mcpserverentry_controller.go
```diff
@@ -0,0 +1,283 @@
...
+
+	authConfig, ok := obj.(*mcpv1alpha1.MCPExternalAuthConfig)
+	if !ok {
+		ctxLogger.Error(nil, "Object is not an MCPExternalAuthConfig", "object", obj.GetName())
+		return nil
+	}
+
+	// List all MCPServerEntries in the same namespace
+	entryList := &mcpv1alpha1.MCPServerEntryList{}
+	if err := r.List(ctx, entryList, client.InNamespace(authConfig.Namespace)); err != nil {
```
ChrisJBurns: **[MEDIUM]** *Found by: kubernetes-expert*

`findEntriesForAuthConfig` lists **all** MCPServerEntries in the namespace and filters client-side. By contrast, `findReferencingMCPServerEntries` in the MCPGroup controller uses `client.MatchingFields` with a field index for efficient server-side filtering.

For consistency and scalability, consider adding a field indexer on `spec.externalAuthConfigRef.name` and using `client.MatchingFields` here instead.

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
```diff
@@ -0,0 +1,173 @@
...
+// MCPServerEntry is a zero-infrastructure catalog entry that declares a remote MCP
+// server endpoint. Unlike MCPRemoteProxy, it creates no pods, services, or deployments.
+//
+//nolint:lll // kubebuilder marker URLs exceed line length limit
+type MCPServerEntrySpec struct {
+	// RemoteURL is the URL of the remote MCP server.
+	// HTTPS is enforced by default; set the annotation
+	// toolhive.stacklok.dev/allow-insecure: "true" to allow HTTP for development.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^https?://`
```
ChrisJBurns, line 17: **[MEDIUM]** *Found by: security-advisor*

The URL pattern `^https?://` only validates the scheme prefix. It allows potentially dangerous URLs:
- Cloud metadata: `http://169.254.169.254/latest/meta-data/`
- Localhost: `http://127.0.0.1:8080/`
- Internal K8s services: `http://kubernetes.default.svc/`

Even though MCPServerEntry creates no infrastructure, the URL is consumed by downstream vMCP components. If any component fetches the URL in-cluster, these become SSRF vectors.

Consider adding validation (either CEL or in the controller) to reject well-known internal/metadata IP ranges, or at minimum document that the URL must point to an externally accessible endpoint.

cmd/thv-operator/controllers/mcpgroup_controller.go
```diff
@@ -161,11 +165,13 @@ func (r *MCPGroupReconciler) handleListFailure(
 		ObservedGeneration: mcpGroup.Generation,
 	})
 
-	// Clear both resource types' status fields to avoid stale data when entering Failed state
+	// Clear all resource types' status fields to avoid stale data when entering Failed state
```
ChrisJBurns, line 168: **[MEDIUM]** *Found by: kubernetes-expert*

If listing MCPServerEntries fails but listing MCPServers and MCPRemoteProxies succeeded, all three counts/lists are zeroed (lines 169-174). The already-retrieved server and proxy data is discarded.

This means a transient failure listing entries causes the MCPGroup to report zero servers too, which could confuse operators and monitoring. Consider only clearing the status for the resource type that failed, or document why the all-or-nothing approach was chosen.

cmd/thv-operator/controllers/mcpgroup_controller_test.go
```diff
@@ -205,6 +205,13 @@ func TestMCPGroupReconciler_Reconcile_BasicLogic(t *testing.T) {
 					}
 					return []string{mcpRemoteProxy.Spec.GroupRef}
 				}).
+				WithIndex(&mcpv1alpha1.MCPServerEntry{}, "spec.groupRef", func(obj client.Object) []string {
+					mcpServerEntry := obj.(*mcpv1alpha1.MCPServerEntry)
+					if mcpServerEntry.Spec.GroupRef == "" {
+						return nil
```
ChrisJBurns, line 211: **[MINOR]** *Found by: kubernetes-expert*

The same 7-line MCPServerEntry field indexer block is copy-pasted into 11 test function builders in this file. Combined with the 5 integration test suites, that's 16 copies of the same code. A shared test helper like `addGroupRefFieldIndexes(builder)` would significantly reduce duplication and maintenance burden.

cmd/thv-operator/controllers/mcpserverentry_controller.go
JAORMX: Good catch, yeah the intent was definitely there... I just forgot to wire it up. Fixed now. The controller checks `group.Status.Phase != MCPGroupPhaseReady` after the successful `Get` and sets `GroupRefNotReady` with a message showing the current phase. Matches what MCPServer and MCPRemoteProxy do.

cmd/thv-operator/controllers/mcpserverentry_controller.go
JAORMX: Yeah, you're right... stale conditions are basically the one thing this controller can get wrong since it does nothing else. Added watches for both `MCPGroup` and `corev1.ConfigMap` now. The ConfigMap watch uses a field index on `spec.caBundleRef.configMapRef.name` so it only re-queues entries that actually reference the changed ConfigMap (same pattern as the auth config watch, but with server-side filtering this time).

cmd/thv-operator/controllers/mcpserverentry_controller.go
JAORMX: Yep, fixed. Reduced to `get;list;watch` and re-ran `task operator-manifests`. The generated ClusterRole now groups `mcpserverentries` with the other read-only resources.

cmd/thv-operator/controllers/mcpserverentry_controller.go
JAORMX: Went with the "return the error" approach. All three validate methods now return `(bool, error)`. If the `Get` fails with anything other than NotFound, we return the error directly so the controller-runtime requeue logic kicks in. No misleading condition gets persisted, and the detailed error is logged server-side only (no more raw `err.Error()` in status conditions either, which addresses the security concern about leaking internal details).

cmd/thv-operator/controllers/mcpserverentry_controller.go
JAORMX: Fair point. Added `TestMCPServerEntryReconciler_Reconcile` with 8 table-driven subtests covering: happy path (all refs valid), optional refs nil, group not found, group not ready, auth config not found, CA bundle not found, entry not found (no-op), and CA bundle with nil configMapRef. All passing.

cmd/thv-operator/controllers/mcpgroup_controller.go
JAORMX: Added two new test functions:

- `TestMCPGroupReconciler_MCPServerEntryIntegration` that creates a group with 2 entries and asserts `EntryCount=2` and `Entries=["entry1","entry2"]` after reconciliation.
- `TestMCPGroupReconciler_EntryDeletionHandler` that calls `updateReferencingEntriesOnDeletion` and verifies both entries get `GroupRefValidated=False`.

So the non-zero EntryCount path and the deletion cleanup path are both covered now.

cmd/thv-operator/test-integration/mcp-oidc-config/suite_test.go
JAORMX: Nice catch! Yeah that was a sneaky nesting bug. Fixed in all four suites. The MCPServerEntry field indexer now lives outside the `if err` block so it runs unconditionally, matching how `mcp-group/suite_test.go` already had it.

cmd/thv-operator/test-integration/mcp-remote-proxy/suite_test.go
JAORMX: Fixed, same as above.

cmd/thv-operator/test-integration/mcp-server/suite_test.go
JAORMX: Fixed.

cmd/thv-operator/test-integration/virtualmcp/suite_test.go
JAORMX: Fixed.

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
JAORMX: Went with option (b). Removed the `AllowInsecureAnnotation` constant entirely and updated the RemoteURL doc comment to just say both HTTP and HTTPS are accepted at admission time. No point having a documented annotation that nothing checks... that's worse than having no annotation at all. If we want actual HTTPS enforcement later, we can add it properly in a follow-up with the controller actually checking.

cmd/thv-operator/controllers/mcpserverentry_controller.go
JAORMX: Fixed. Reduced to `get;list;watch` and regenerated with `task operator-manifests`.

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
JAORMX: Good call. The controller now sets `Failed` when validations fail, and `Pending` is just the initial kubebuilder default before the first reconcile. Updated the doc comments on the phase constants to make that distinction clear.

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
JAORMX: It's actually used now! The controller checks the group's Ready phase after the successful `Get`, so `ConditionReasonMCPServerEntryGroupRefNotReady` is wired up.

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
JAORMX: Fixed. The condition types now reuse the shared constants from `mcpserver_types.go` (`GroupRefValidated`, `ExternalAuthConfigValidated`, `CABundleRefValidated`). This was actually already done in the types file from #4662 but the controller was using the old names. The rebase + fix aligns everything now.

cmd/thv-operator/controllers/mcpserverentry_controller.go
JAORMX: Added. The controller now watches `corev1.ConfigMap` with a `findEntriesForConfigMap` map function backed by a field index on `spec.caBundleRef.configMapRef.name`. So if a CA bundle ConfigMap gets created, deleted, or updated, the referencing entries get re-validated automatically.

cmd/thv-operator/controllers/mcpserverentry_controller.go
JAORMX: Fixed by changing the approach entirely. Transient errors (non-NotFound) now return the error to the controller-runtime so it handles the requeue. No condition gets persisted at all for transient failures, which means no raw `err.Error()` ends up in status. The detailed error is only logged server-side via `ctxLogger.Error()`.

cmd/thv-operator/controllers/mcpserverentry_controller.go
JAORMX: Done. Added a field index on `spec.externalAuthConfigRef.name` in `SetupWithManager` and switched `findEntriesForAuthConfig` to use `client.MatchingFields`. Same pattern as the MCPGroup controller's `findReferencingMCPServerEntries`. Also added a similar field index for `spec.caBundleRef.configMapRef.name` for the new ConfigMap watch.

cmd/thv-operator/api/v1alpha1/mcpserverentry_types.go
JAORMX: Valid concern. I'd rather tackle URL validation properly in a follow-up though, since it touches how vMCP consumes these URLs downstream and I don't want to half-do it here. The MCPRemoteProxy CRD has the same `^https?://` pattern today too, so this isn't a regression specific to MCPServerEntry. Worth tracking as a separate issue for both resource types.

cmd/thv-operator/controllers/mcpgroup_controller.go
JAORMX: The all-or-nothing approach is intentional here. If any list operation fails, we don't know the actual membership state, so reporting partial data could be more misleading than reporting zero. Imagine the entry list fails but we report "2 servers, 0 entries" when there should be 5 entries... that partial view could cause downstream consumers to make bad routing decisions.

The `Failed` phase + the condition message saying which resource type failed to list should give operators enough signal to know what went wrong. But I'm open to changing this if you feel strongly about it.

cmd/thv-operator/controllers/mcpgroup_controller_test.go
JAORMX: Agreed, there's a lot of copy-paste here. I'll track this as a follow-up to extract a shared `addGroupRefFieldIndexes(builder)` helper. Didn't want to refactor the existing test infrastructure in this PR since it's already touching a bunch of files.

Issue #4495, opened 2026-04-02
vmcp: replace sentinel pattern with typed cache + conditional storage write (follow-up to #4464)


yrobla:
## Context

Follow-up to #4464 / #4494. The `terminatedSentinel` / `Peek` / untyped `Store(any)` / untyped `CompareAndSwap(any, any)` pattern in `RestorableCache` was introduced to prevent a specific race: `DecorateSession` resurrecting a storage record after `Terminate` deletes it. The approach is correct and the race window is narrow (client DELETE during session initialization), but it pushes a storage-layer concern into the cache, punches through the cache's type safety, and makes the concurrency model hard to reason about.

## Proposed design

Two clean invariants, one per layer:

### 1. Cache invariant — "no stale reads"

`check` runs on every insertion and retrieval. The cache is fully typed (`Store(K, V)`, no `any`). `Terminate` deletes from storage then evicts from the cache. No sentinels needed — any concurrent `Get()` or `Store()` runs `check`, which calls `storage.Load()`. If the record is gone, the operation is rejected naturally.

### 2. Storage invariant — "no resurrection after delete"

Replace `Upsert` with `Update` (Redis `SET XX`), which succeeds only if the key already exists and returns `ErrSessionNotFound` if it does not.

All current `Upsert` callers already expect the record to exist:
- `CreateSession` — called after `Generate` wrote the placeholder
- `DecorateSession` — called after session initialization
- `cleanupFailedPlaceholder` — cleanup path, record must exist
- `Terminate`'s placeholder path — same

None of them need unconditional write semantics.

### How the race resolves

With `Update` (SET XX) in place, the `Terminate` / `DecorateSession` race resolves at the storage layer without any cache-level coordination:

```
Terminate:        Delete(sessionID)              → succeeds
DecorateSession:  Update(sessionID, metadata)    → returns ErrSessionNotFound
DecorateSession:  rolls back cache entry, returns error
```

## What gets removed

- `terminatedSentinel` type and all uses
- `Peek(K) (any, bool)`
- `Store(K, any)` (replaced by typed `Store(K, V)` or removed entirely)
- `CompareAndSwap(K, any, any)`
- The rollback / sentinel-placement logic in `Terminate`
- Sentinel-specific test cases in `cache_test.go`

## What gets added / changed

- `DataStorage.Update(ctx, sessionID, metadata)` — conditional write, returns `ErrSessionNotFound` if key absent
- Redis implementation: `SET XX` with TTL refresh
[...]

PR #4669, opened 2026-04-08
Add LRU capacity to ValidatingCache, remove sentinel pattern, add storage Update

### LINKED ISSUE #4494: vmcp: address RestorableCache capacity limit and sentinel pattern complexity (follow-up to #4464)
## Context

Follow-up to #4464 (session manager DataStorage migration). The architecture is solid but two issues should be addressed before the cache is considered production-ready.

## Blocker 1: Cache capacity limit

`RestorableCache` has no size bound or background eviction sweep. Idle sessions (e.g. clients that disconnect without a clean MCP shutdown) will remain in the node-local cache until their Redis TTL expires — holding references to backend connections, routing tables, and token sources indefinitely.

**Required**: add at minimum a capacity limit (LRU or similar) so the cache cannot grow without bound. A background sweep keyed on entry age would also bound the leak window for truly idle sessions.

Related: capacity/LRU eviction was intentionally deferred in the initial implementation (see `cache.go` TODO comment).

## Blocker 2: Sentinel pattern complexity

The current `terminatedSentinel` / `Peek` / untyped `Store(any)` / `CompareAndSwap(any, any)` pattern was introduced to prevent session resurrection after `DeleteSession` races with a concurrent `RestoreSession`. It works correctly but adds surface area that reviewers flagged:

- `Store` and `CompareAndSwap` accept `any`, weakening type safety
- `Peek` exposes raw `any` values, requiring callers to type-assert
- The sentinel type leaks into test utilities (`testSentinel`, `Peek` assertions)

**Preferred fix**: eliminate the sentinel entirely by pushing the no-resurrection invariant into the storage layer:

1. Replace `DataStorage.Upsert` (or add an `Update`) with a **conditional write** — `SET XX` in Redis semantics — that returns `ErrSessionNotFound` if the key no longer exists. `RestoreSession` would call `Update` rather than `Upsert` after reconstructing the session, so a deleted key causes the restore to fail cleanly rather than re-inserting.
2. Make `RestorableCache` fully typed: remove `Store(any)`, `Peek`, and `CompareAndSwap(any, any)`; all cache writes go through the typed `load` path or an explicit typed `Set(K, V)`.

With this design, a `DeleteSession` → `RestoreSession` race is resolved by the storage layer (the conditional write fails), and the cache never needs to distinguish sentinel from live session.

## Additional clean-up

- **Rename** `RestorableCache` → `ValidatingCache` (the defining characteristic is that every cache hit is validated for liveness, not that entries are restorable).
- **Move** to `pkg/cache/` so it can be reused by other subsystems without importing a vmcp-internal package.

## Acceptance criteria

- [ ] `RestorableCache` (or `ValidatingCache`) has a configurable capacity limit; entries beyond the limit are evicted (LRU or equivalent)
- [ ] Optional: background sweep evicts entries older than a configurable max age without touching storage
- [ ] `terminatedSentinel`, `Peek`, untyped `Store`/`CompareAndSwap` are removed
- [ ] `DataStorage` exposes a conditional-write operation (`Update` / `SET XX`) and `RestoreSession` uses it
- [ ] Cache is renamed `ValidatingCache` and lives in `pkg/cache/`
- [ ] All existing `RestorableCache` tests pass under the new design (sentinel-specific tests replaced by storage-layer tests)

yrobla:
## Summary

- Introduce pkg/cache.ValidatingCache[K,V]: a fully-typed, capacity-bounded LRU cache using mutex+map+container/list. Replaces the sessionmanager-internal RestorableCache and its sync.Map backing store.
- Add cache.ErrExpired, cache.New(capacity, load, check, onEvict). A capacity of 0 keeps the existing unbounded behaviour; non-zero evicts the LRU entry (calling onEvict) when the limit is reached.
- Eliminate terminatedSentinel / Peek(any) / Store(any) / CompareAndSwap(any): all cache operations are now fully typed. The no-resurrection invariant is pushed into the storage layer instead.
- Add DataStorage.Update (SET XX semantics): overwrites metadata only if the key already exists, returning (false, nil) when absent. Implemented via Redis SetArgs{Mode:"XX"} and a mutex-protected Load+Store in the local in-memory backend.
- Replace Terminate's sentinel+rollback dance with a simpler storage.Delete followed by cache.Delete. Concurrent RestoreSession calls fail naturally via storage.Load returning ErrSessionNotFound.
- Replace updateMetadata's Load+Upsert+sentinel-check with a single storage.Update call, closing the TOCTOU resurrection window entirely.
- Replace DecorateSession's storage.Upsert with storage.Update; a (false, nil) result rolls back the in-cache CAS instead of silently resurrecting a deleted session.
- Add FactoryConfig.CacheCapacity to wire the limit through sessionmanager.New.
- Remove sessionmanager/cache.go and cache_test.go; cache tests live in pkg/cache/validating_cache_test.go, covering LRU eviction order, capacity=1, unlimited capacity, singleflight deduplication, and liveness expiry.

Fixes #4494 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## Changes

| File | Change |
|------|--------|
|      |        |

## Large PR Justification

This is a complete pr covering a single functionality. Cannot be split.

### DIFF
--- a/pkg/cache/validating_cache.go
+++ b/pkg/cache/validating_cache.go
@@ -121,13 +121,23 @@ func (c *ValidatingCache[K, V]) Get(key K) (V, bool) {
 			return nil, loadErr
 		}
 
-		// Guard against a concurrent Set/Delete that occurred while load() was
+		// Guard against a concurrent Set or Remove that occurred while load() was
 		// running. ContainsOrAdd stores only if absent; if another writer got
 		// in first, their value wins and we discard ours via onEvict.
 		ok, _ := c.lruCache.ContainsOrAdd(key, v)
 		if ok {
-			// Another writer stored a value first; retrieve the winner's value.
-			winner, _ := c.lruCache.Get(key)
+			// Another writer stored a value first; discard our loaded value and
+			// return the winner's. ContainsOrAdd and Get are separate lock
+			// acquisitions, so the winner may itself have been evicted by LRU
+			// pressure between the two calls — fall back to our freshly loaded
+			// value in that case rather than returning a zero value.
+			winner, found := c.lruCache.Get(key)
+			if !found {
+				// Winner was evicted between ContainsOrAdd and Get; keep our
+				// freshly loaded value rather than returning a zero value.
+				return result{v: v}, nil
+			}
+			// Discard our loaded value in favour of the winner.
 			if c.onEvict != nil {
 				c.onEvict(key, v)
 			}
@@ -149,6 +159,12 @@ func (c *ValidatingCache[K, V]) Get(key K) (V, bool) {
 // onEvict is called for it.
 func (c *ValidatingCache[K, V]) Set(key K, value V) {
 	c.lruCache.Add(key, value)
+}
+
+// Remove evicts the entry for key, calling onEvict if the key was present.
+// It is a no-op if the key is not in the cache.
+func (c *ValidatingCache[K, V]) Remove(key K) {
+	c.lruCache.Remove(key)
 }
 
 // Len returns the number of entries currently in the cache.
--- a/pkg/transport/session/session_data_storage_test.go
+++ b/pkg/transport/session/session_data_storage_test.go
@@ -162,48 +162,6 @@ func runDataStorageTests(t *testing.T, newStorage func(t *testing.T) DataStorage
 		assert.Error(t, err)
 	})
 
-	t.Run("Update on missing key returns false without creating it", func(t *testing.T) {
-		t.Parallel()
-		s := newStorage(t)
-		ctx := context.Background()
-
-		updated, err := s.Update(ctx, "nonexistent", map[string]string{"k": "v"})
-		require.NoError(t, err)
-		assert.False(t, updated, "Update on absent key must return false")
-
-		_, loadErr := s.Load(ctx, "nonexistent")
-		assert.ErrorIs(t, loadErr, ErrSessionNotFound, "Update must not create the key")
-	})
-
-	t.Run("Update on existing key overwrites metadata", func(t *testing.T) {
-		t.Parallel()
-		s := newStorage(t)
-		ctx := context.Background()
-
-		require.NoError(t, s.Upsert(ctx, "sess-update", map[string]string{"v": "original"}))
-
-		updated, err := s.Update(ctx, "sess-update", map[string]string{"v": "updated"})
-		require.NoError(t, err)
-		assert.True(t, updated, "Update on present key must return true")
-
-		loaded, loadErr := s.Load(ctx, "sess-update")
-		require.NoError(t, loadErr)
-		assert.Equal(t, "updated", loaded["v"])
-	})
-
-	t.Run("Update after Delete returns false", func(t *testing.T) {
-		t.Parallel()
-		s := newStorage(t)
-		ctx := context.Background()
-
-		require.NoError(t, s.Upsert(ctx, "sess-del", map[string]string{"v": "1"}))
-		require.NoError(t, s.Delete(ctx, "sess-del"))
-
-		updated, err := s.Update(ctx, "sess-del", map[string]string{"v": "2"})
-		require.NoError(t, err)
-		assert.False(t, updated, "Update after Delete must return false")
-	})
-
 	t.Run("Delete removes entry; subsequent Load returns ErrSessionNotFound", func(t *testing.T) {
 		t.Parallel()
 		s := newStorage(t)
--- a/pkg/vmcp/server/sessionmanager/session_manager.go
+++ b/pkg/vmcp/server/sessionmanager/session_manager.go
@@ -102,8 +102,12 @@ func New(
 	if cfg == nil || cfg.Base == nil {
 		return nil, nil, fmt.Errorf("sessionmanager.New: FactoryConfig.Base (SessionFactory) is required")
 	}
-	if cfg.CacheCapacity < 1 {
-		return nil, nil, fmt.Errorf("sessionmanager.New: CacheCapacity must be >= 1 (got %d)", cfg.CacheCapacity)
+	if cfg.CacheCapacity < 0 {
+		return nil, nil, fmt.Errorf("sessionmanager.New: CacheCapacity must be >= 0 (got %d)", cfg.CacheCapacity)
+	}
+	capacity := cfg.CacheCapacity
+	if capacity == 0 {
+		capacity = defaultCacheCapacity
 	}
 	if len(cfg.WorkflowDefs) > 0 && cfg.ComposerFactory == nil {
 		return nil, nil, fmt.Errorf("sessionmanager.New: ComposerFactory is required when WorkflowDefs are provided")
@@ -139,7 +143,7 @@ func New(
 	}
 
 	sm.sessions = cache.New(
-		cfg.CacheCapacity,
+		capacity,
 		sm.loadSession,
 		sm.checkSession,
 		func(id string, sess vmcpsession.MultiSession) {
@@ -389,10 +393,15 @@ func (sm *Manager) CreateSession(
 // Cleanup is best-effort: errors are logged but not returned, since the caller
 // already has an error to report.
 func (sm *Manager) cleanupFailedPlaceholder(sessionID string, metadata map[string]string) {
-	metadata[MetadataKeyTerminated] = MetadataValTrue
+	// Copy before mutating so the caller's map is not modified.
+	terminated := make(map[string]string, len(metadata)+1)
+	for k, v := range metadata {
+		terminated[k] = v
+	}
+	terminated[MetadataKeyTerminated] = MetadataValTrue
 	cleanupCtx, cancel := context.WithTimeout(context.Background(), createSessionStorageTimeout)
 	defer cancel()
-	if _, err := sm.storage.Update(cleanupCtx, sessionID, metadata); err != nil {
+	if _, err := sm.storage.Update(cleanupCtx, sessionID, terminated); err != nil {
 		slog.Warn("Manager.CreateSession: failed to mark failed placeholder as terminated; it will linger until TTL expires",
 			"session_id", sessionID, "error", err)
 	}
@@ -469,20 +478,25 @@ func (sm *Manager) Terminate(sessionID string) (isNotAllowed bool, err error) {
 	}
 
 	if _, isFullSession := metadata[sessiontypes.MetadataKeyTokenHash]; isFullSession {
-		// Phase 2 (full MultiSession): delete from storage. The node-local
-		// cache self-heals on the next Get: checkSession detects ErrSessionNotFound,
-		// evicts the entry, and onEvict closes backend connections.
+		// Phase 2 (full MultiSession): delete from storage, then evict from the
+		// node-local cache so onEvict closes backend connections immediately rather
+		// than waiting for the next Get or an LRU eviction.
 		if deleteErr := sm.storage.Delete(ctx, sessionID); deleteErr != nil {
 			return false, fmt.Errorf("Manager.Terminate: failed to delete session from storage: %w", deleteErr)
 		}
+		sm.sessions.Remove(sessionID)
 		slog.Info("Manager.Terminate: session terminated", "session_id", sessionID)
 		return false, nil
 	}
 
 	// Phase 1 (placeholder): mark terminated so CreateSession fast-fails and
 	// Validate returns isTerminated=true during the TTL window.
+	// Use Update (SET XX) rather than Upsert so we never resurrect a key that
+	// was concurrently deleted or expired between the Load above and this write.
+	// (false, nil) means already gone — treat as success.
 	metadata[MetadataKeyTerminated] = MetadataValTrue
-	if storeErr := sm.storage.Upsert(ctx, sessionID, metadata); storeErr != nil {
+	updated, storeErr := sm.storage.Update(ctx, sessionID, metadata)
+	if storeErr != nil {
 		slog.Warn("Manager.Terminate: failed to persist terminated flag for placeholder; attempting delete fallback",
 			"session_id", sessionID, "error", storeErr)
 		deleteCtx, deleteCancel := context.WithTimeout(context.Background(), terminateTimeout)
@@ -493,6 +507,9 @@ func (sm *Manager) Terminate(sessionID string) (isNotAllowed bool, err error) {
 				storeErr, deleteErr)
 		}
 		deleteCancel()
+	} else if !updated {
+		// Session expired or was concurrently deleted between Load and Update — already gone.
+		slog.Debug("Manager.Terminate: placeholder already gone before terminated flag could be set", "session_id", sessionID)
 	}
 
 	slog.Info("Manager.Terminate: session terminated", "session_id", sessionID)
@@ -714,14 +731,10 @@ func (sm *Manager) DecorateSession(sessionID string, fn func(sessiontypes.MultiS
 		return fmt.Errorf("DecorateSession: decorator changed session ID from %q to %q", sessionID, decorated.ID())
 	}
 
-	// Store the decorated session in the node-local cache.
-	// If storage.Update below fails or returns (false, nil) because Terminate
-	// deleted the storage key, the decorated entry remains in cache briefly;
-	// the next Gets checkSession detects the absent key, evicts the entry,
-	// and onEvict closes connections.
-	sm.sessions.Set(sessionID, decorated)
-
-	// Persist metadata to storage via conditional Update (SET XX).
+	// Persist metadata to storage first via conditional Update (SET XX).
+	// Only update the node-local cache after a successful write so that a
+	// storage error or a concurrent delete never leaves a decorated (but
+	// unpersisted) value in the cache where retries could stack decorations.
 	decorateCtx, decorateCancel := context.WithTimeout(context.Background(), decorateTimeout)
 	defer decorateCancel()
 	updated, err := sm.storage.Update(decorateCtx, sessionID, decorated.GetMetadata())
@@ -729,8 +742,12 @@ func (sm *Manager) DecorateSession(sessionID string, fn func(sessiontypes.MultiS
 		return fmt.Errorf("DecorateSession: failed to store decorated session metadata: %w", err)
 	}
 	if !updated {
+		// Session was deleted (by Terminate or TTL) between Get and Update.
+		// Evict the stale cache entry so onEvict closes backend connections.
+		sm.sessions.Remove(sessionID)
 		return fmt.Errorf("DecorateSession: session %q was deleted during decoration", sessionID)
 	}
+	sm.sessions.Set(sessionID, decorated)
 	return nil
 }
 
--- a/pkg/vmcp/server/sessionmanager/factory.go
+++ b/pkg/vmcp/server/sessionmanager/factory.go
@@ -32,6 +32,11 @@ import (
 
 const instrumentationName = "github.com/stacklok/toolhive/pkg/vmcp"
 
+// defaultCacheCapacity is the fallback used when FactoryConfig.CacheCapacity is
+// zero (the Go zero value). This ensures the cache is always bounded; omitting
+// CacheCapacity from a config does not silently enable unbounded growth.
+const defaultCacheCapacity = 1000
+
 // FactoryConfig holds the session factory construction parameters that the
 // session manager needs to build its decorating factory. It is separate from
 // server.Config to avoid a circular import between the server and sessionmanager
@@ -65,8 +70,9 @@ type FactoryConfig struct {
 
 	// CacheCapacity is the maximum number of live MultiSession entries held in
 	// the node-local ValidatingCache. When the cache is full the least-recently-used
-	// session is evicted (its backend connections are closed via onEvict).
-	// Must be >= 1; sessionmanager.New returns an error if this is zero or negative.
+	// session is evicted (its backend connections are closed via onEvict). A value of
+	// 0 uses defaultCacheCapacity (1000). Negative values are rejected by
+	// sessionmanager.New.
 	CacheCapacity int
 }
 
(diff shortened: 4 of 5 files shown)


pkg/cache/validating_cache.go
```diff
@@ -0,0 +1,300 @@
...
+// load is called on a cache miss to restore the value; it must not be nil.
+// check is called on every cache hit to confirm liveness; it must not be nil.
+// onEvict is called after any eviction (LRU or expiry); it may be nil.
+func New[K comparable, V any](
+	capacity int,
+	load func(K) (V, error),
+	check func(K) error,
+	onEvict func(K, V),
+) *ValidatingCache[K, V] {
+	if capacity < 0 {
```
jerm-dro: **question (blocking):** Is there a reason we're hand-rolling the LRU implementation rather than building on an established library like [`hashicorp/golang-lru/v2`](https://github.com/hashicorp/golang-lru) or [`dgraph-io/ristretto`](https://github.com/dgraph-io/ristretto)?

The custom value on top — singleflight-deduplicated load on miss, per-hit liveness validation, typed CAS — is real and worth keeping. But the underlying LRU mechanics (mutex + `container/list` + map, capacity enforcement, eviction ordering) are ~100 lines of code that these libraries already provide, battle-tested under concurrent load.

`hashicorp/golang-lru/v2` is generic (`lru.Cache[K, V]`), supports `Add`/`Get`/`Remove`/`Peek`/`Len` with an eviction callback, and is a single lightweight dependency. `ValidatingCache` could wrap it and layer singleflight + liveness on top. If there's a requirement I'm missing that prevents using a third-party cache here, please document it — otherwise I'd prefer we don't maintain our own LRU bookkeeping.

pkg/cache/validating_cache.go
```diff
@@ -0,0 +1,300 @@
...
+// capacity is the maximum number of entries. When the cache is full and a new
+// entry must be stored, the least-recently-used entry is evicted first.
+// A capacity of 0 disables the limit (the cache grows without bound).
+// Negative values panic.
+//
+// load is called on a cache miss to restore the value; it must not be nil.
+// check is called on every cache hit to confirm liveness; it must not be nil.
+// onEvict is called after any eviction (LRU or expiry); it may be nil.
+func New[K comparable, V any](
+	capacity int,
```
jerm-dro, line 55: **blocker:** A capacity of 0 meaning "unbounded" defeats the purpose of this change. The whole motivation (#4494) is that idle sessions leak backend connections indefinitely — an unbounded mode leaves that door open via misconfiguration or a missing config value (zero is Go's default for `int`).

Capacity should be required and positive. If a caller omits `CacheCapacity` from `FactoryConfig`, that should fail loudly at startup rather than silently falling back to unbounded growth. Remove the `capacity == 0` unlimited path and require `capacity >= 1`.

pkg/cache/validating_cache.go
```diff
@@ -0,0 +1,300 @@
...
+		var zero V
+		return zero, false
+	}
+	return e.val, true
+}
+
+// CompareAndSwap atomically replaces the value stored under key from old to
+// replacement. The comparison uses == semantics: for interface types this
+// means pointer equality when the dynamic value is a pointer. Returns true if
+// the swap was performed. Returns false (without panicking) if V's dynamic
```
jerm-dro: **question (blocking):** Why does `Peek` still exist on `ValidatingCache`? #4495 called for removing it — it was only needed to inspect sentinel markers, which are now gone.

The remaining call sites in `session_manager.go` use `Peek` in `Terminate` (to check if a session is cached before deleting storage) and `checkSession` (to compare cached backend IDs against storage). Both could use `Get` instead — `Terminate` is about to delete anyway so a liveness check is harmless, and `checkSession` *is* the liveness check so re-entering through `Get` needs thought but `Peek` bypassing validation is arguably worse.

If there's a reason to keep a no-side-effect read on the generic cache, I'd like to understand it. Otherwise, removing it shrinks the API surface and avoids callers silently skipping liveness validation.

pkg/cache/validating_cache.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	e.val = replacement
+	c.lru.MoveToFront(e.elem)
+	return true
+}
+
+// Len returns the number of entries currently in the cache.
+func (c *ValidatingCache[K, V]) Len() int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
jerm-dro: **question (blocking):** Same question for `CompareAndSwap` — #4495 called for removing it along with the sentinel pattern. The only remaining caller is `DecorateSession`, which uses CAS to guard against concurrent modification during decoration.

Now that the no-resurrection invariant is enforced at the storage layer via `Update` (SET XX), is the cache-level CAS still load-bearing? `DecorateSession` already calls `storage.Update` after the CAS and handles the `(false, nil)` case. Could this be simplified to `Delete` + `Set` (or just `Set`) on the cache, with `storage.Update` as the authoritative concurrency guard? If CAS is still needed for a reason beyond sentinel detection, please document what race it prevents that `storage.Update` doesn't.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -739,6 +718,28 @@ func (sm *Manager) loadSession(sessionID string) (vmcpsession.MultiSession, erro
...
+	if confirmErr != nil {
+		_ = restored.Close()
+		return nil, fmt.Errorf("Manager.loadSession: failed to confirm session %q still exists: %w", sessionID, confirmErr)
+	}
+	if !stillExists {
+		// Terminate() deleted the storage key between our Load and now; the
+		// session must not be cached.
+		slog.Debug("Manager.loadSession: session deleted during restore, discarding", "session_id", sessionID)
+		_ = restored.Close()
+		return nil, transportsession.ErrSessionNotFound
```
jerm-dro: **blocker:** The post-restore existence check adds complexity for a race that shouldn't happen in practice — a caller wouldn't terminate a session they're actively using. If `storage.Load` succeeded and `RestoreSession` completed, the session is valid. A concurrent `Terminate` during restore is an edge case that doesn't justify an extra storage round-trip on every cache miss.

Remove the `storage.Update` existence check and the surrounding error handling. If the extremely unlikely Terminate-during-restore race does occur, the next `checkSession` liveness validation will catch it and evict the entry naturally — that's what the liveness check is for.

pkg/transport/session/session_data_storage_local.go
```diff
@@ -35,7 +35,8 @@ func (e *localDataEntry) lastAccess() time.Time {
 // Sessions are evicted if they have not been accessed within the configured TTL.
 // A background goroutine runs until Close is called.
 type LocalSessionDataStorage struct {
-	sessions sync.Map // map[string]*localDataEntry
+	sessions sync.Map   // map[string]*localDataEntry
```
jerm-dro: **suggestion:** `LocalSessionDataStorage` now has both `sync.Mutex` and `sync.Map` — `Update` and `Delete` go through `mu`, while `Upsert` and `Create` use `sync.Map` alone. Two synchronization primitives on the same data is a correctness hazard for future contributors.

Since `mu` is now required for `Update`/`Delete` atomicity, the lock-free benefit of `sync.Map` is gone. Replace `sync.Map` with a plain `map[string]*localDataEntry` guarded by `mu` for all operations — simpler, one primitive, no mixed-mode concurrency to reason about. The performance difference is negligible at session-count scale.

pkg/cache/validating_cache.go
```diff
@@ -0,0 +1,300 @@
...
+		c.evictEntryLocked(e)
+	}
+	c.mu.Unlock()
+}
+
+// Peek returns the cached V value for key without updating the LRU order,
+// running a liveness check, or triggering a restore on a cache miss.
+// Used by callers that need to read the current in-memory value without
+// side effects (e.g. reading metadata for comparison inside a liveness check).
+func (c *ValidatingCache[K, V]) Peek(key K) (V, bool) {
```
jerm-dro: **blocker:** `Delete` should be removed. Every removal path should go through `onEvict` to maintain a single lifecycle invariant: the cache owns cleanup of its entries.

The two current callers of `Delete` don't need it:
- **`Terminate`**: just needs `storage.Delete`. The next `Get` fails the liveness check and triggers eviction + `onEvict` naturally.
- **`updateMetadata`**: writes new metadata to storage, then deletes from cache to force a re-restore. Instead, extend the liveness check (`checkSession`) to detect metadata drift — it already does this for backend IDs. The next `Get` would see the stale cache entry, evict it, and re-restore with fresh metadata. No explicit `Delete` needed.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -438,83 +457,42 @@ func (sm *Manager) Terminate(sessionID string) (isNotAllowed bool, err error) {
...
+			return false, fmt.Errorf("Manager.Terminate: failed to delete session from storage: %w", deleteErr)
+		}
+		slog.Info("Manager.Terminate: session terminated", "session_id", sessionID)
+		return false, nil
+	}
+
+	// Phase 1 (placeholder): mark terminated so CreateSession fast-fails and
+	// Validate returns isTerminated=true during the TTL window.
 	metadata[MetadataKeyTerminated] = MetadataValTrue
 	if storeErr := sm.storage.Upsert(ctx, sessionID, metadata); storeErr != nil {
```
JAORMX: This one made me do a double-take. The whole point of this PR is closing the TOCTOU resurrection window, and `CreateSession`, `DecorateSession`, and `updateMetadata` all correctly use `storage.Update` (SET XX) for that. But here in the placeholder termination path, we're still using `Upsert`, which can recreate a deleted key.

If between `Load` (line 461) and this `Upsert`, another goroutine or pod deletes the key, `Upsert` silently recreates it... which is exactly the race we're trying to eliminate.

`storage.Update` with `(false, nil)` handled as a no-op (key already gone, nothing to mark) would close this consistently with the rest of the PR.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -438,83 +457,42 @@ func (sm *Manager) Terminate(sessionID string) (isNotAllowed bool, err error) {
...
-	// TTL cleanup will remove the placeholder later.
+	if _, isFullSession := metadata[sessiontypes.MetadataKeyTokenHash]; isFullSession {
+		// Phase 2 (full MultiSession): delete from storage. The node-local
+		// cache self-heals on the next Get: checkSession detects ErrSessionNotFound,
+		// evicts the entry, and onEvict closes backend connections.
+		if deleteErr := sm.storage.Delete(ctx, sessionID); deleteErr != nil {
+			return false, fmt.Errorf("Manager.Terminate: failed to delete session from storage: %w", deleteErr)
+		}
+		slog.Info("Manager.Terminate: session terminated", "session_id", sessionID)
+		return false, nil
```
JAORMX, line 489: The PR description says "Replace Terminate's sentinel+rollback dance with a simpler storage.Delete followed by cache.Delete" but there's no `cache.Delete` call here. And `ValidatingCache` doesn't expose a `Delete` method at all.

The lazy self-healing via `checkSession` on the next `Get` works in theory, but if nobody calls `GetMultiSession` for this session again (client disconnected, done, etc.), the cached `MultiSession` holds open backend HTTP connections indefinitely... until LRU pressure evicts it. That could be a long time if the cache is large.

Would it make sense to add a `Delete(key K)` method to `ValidatingCache` (just wrapping `lruCache.Remove`, which fires `onEvict` automatically) and call it here after `storage.Delete`? That way backend connections get cleaned up promptly.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -767,24 +713,24 @@ func (sm *Manager) DecorateSession(sessionID string, fn func(sessiontypes.MultiS
...
-	// the local-map entry so the caller can retry. If Terminate() has since
-	// replaced the decorated entry with a sentinel, the rollback CAS returns
-	// false and we leave the sentinel in place.
+
+	// Store the decorated session in the node-local cache.
+	// If storage.Update below fails or returns (false, nil) because Terminate
+	// deleted the storage key, the decorated entry remains in cache briefly;
+	// the next Gets checkSession detects the absent key, evicts the entry,
+	// and onEvict closes connections.
+	sm.sessions.Set(sessionID, decorated)
```
JAORMX: The cache write happens before the storage write here. If `storage.Update` fails with a transient error (not a key deletion), the decorated session stays in cache while storage has the old metadata. The comment on lines 718-721 acknowledges this for the deletion case, but a transient storage error is different... `checkSession` will find the key alive and the stale decorated session persists in cache.

Flipping the order (persist to storage first, then update cache on success) would be safer. If `storage.Update` succeeds but the process crashes before `sessions.Set`, the next `Get` just re-loads from storage via `loadSession`. No inconsistency.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -366,13 +383,16 @@ func (sm *Manager) CreateSession(
 // as a valid session), and prevents repeated Validate() calls from refreshing
 // the Redis TTL and keeping the placeholder alive indefinitely.
 //
+// Uses Update (SET XX) so that a Terminate() that already deleted the key is
+// not inadvertently resurrected as a terminated entry.
+//
 // Cleanup is best-effort: errors are logged but not returned, since the caller
 // already has an error to report.
 func (sm *Manager) cleanupFailedPlaceholder(sessionID string, metadata map[string]string) {
 	metadata[MetadataKeyTerminated] = MetadataValTrue
```
JAORMX: Small thing... this mutates the caller's map in place. In `CreateSession`, `metadata` came from `storage.Load`. The local implementation clones on `Load` so it's safe today, but if someone changes the storage backend or the clone behavior, this becomes a subtle corruption bug.

Cloning the map before mutating would make this robust regardless of the storage implementation:

```go
cleanupMeta := maps.Clone(metadata)
cleanupMeta[MetadataKeyTerminated] = MetadataValTrue
```

pkg/cache/validating_cache.go
```diff
@@ -0,0 +1,157 @@
...
+			return nil, loadErr
+		}
+
+		// Guard against a concurrent Set/Delete that occurred while load() was
+		// running. ContainsOrAdd stores only if absent; if another writer got
+		// in first, their value wins and we discard ours via onEvict.
+		ok, _ := c.lruCache.ContainsOrAdd(key, v)
+		if ok {
+			// Another writer stored a value first; retrieve the winner's value.
+			winner, _ := c.lruCache.Get(key)
```
JAORMX: The second return value from `Get` is discarded here. `ContainsOrAdd` and `Get` are two separate lock acquisitions on the hashicorp LRU. Between them, the "winner" entry could theoretically be evicted by LRU pressure from another goroutine (especially at low capacity like `capacity=1`). If that happens, `winner` is the zero value and we return it as valid.

Narrow window? Yes. Impossible under concurrent load? Not quite. Something like:

```go
winner, found := c.lruCache.Get(key)
if c.onEvict != nil {
    c.onEvict(key, v)
}
if !found {
    // Winner evicted between ContainsOrAdd and Get.
    // Return our loaded value instead.
    return result{v: v}, nil
}
return result{v: winner}, nil
```

pkg/cache/validating_cache.go
```diff
@@ -0,0 +1,157 @@
...
+		if existing, ok := c.lruCache.Get(key); ok {
+			return result{v: existing}, nil
+		}
+
+		v, loadErr := c.load(key)
+		if loadErr != nil {
+			return nil, loadErr
+		}
+
+		// Guard against a concurrent Set/Delete that occurred while load() was
```
JAORMX: Nit: comment says "concurrent Set/Delete" but `ValidatingCache` has no `Delete` method. If you do add one (see my comment on Terminate), this becomes accurate. Otherwise just "concurrent Set" is more precise.

pkg/transport/session/session_data_storage_test.go
```diff
@@ -191,6 +233,74 @@ func runDataStorageTests(t *testing.T, newStorage func(t *testing.T) DataStorage
 		err := s.Delete(ctx, "")
 		assert.Error(t, err)
 	})
+
+	t.Run("Update overwrites existing entry and returns true", func(t *testing.T) {
```
JAORMX, line 195: The tests from line 237 through ~303 duplicate the ones at lines 165-205. For example, "Update overwrites existing entry and returns true" (237) does the same thing as "Update on existing key overwrites metadata" (178), and "Update on missing key returns (false, nil) without creating it" (253) duplicates "Update on missing key returns false without creating it" (165). One of each pair should be removed.

pkg/vmcp/server/sessionmanager/session_manager.go
jerm-dro: > Would it make sense to add a Delete(key K) method to ValidatingCache (just wrapping lruCache.Remove, which fires onEvict automatically) and call it here after storage.Delete? That way backend connections get cleaned up promptly.

I'd prefer to avoid cluttering the cache interface with `Delete` methods for a few reasons:
1. the cache invariants guarantee this old connection won't be used
2. the existence of a `Delete` method means it will start getting used in other places too. This adds complexity. Is deleting from the cache functionally necessary or a perf optimization?

To mitigate your concern, I'd rather add later some background job internal to the cache to clean up invalid entries.

pkg/transport/session/session_data_storage.go
```diff
@@ -39,6 +42,13 @@ type DataStorage interface {
 	// Upsert creates or updates session metadata with a sliding TTL.
 	Upsert(ctx context.Context, id string, metadata map[string]string) error
```
jerm-dro, line 44: **suggestion:** `Upsert` has zero production callers in the session manager after this PR — all five call sites were converted to `Update`. The remaining references are test fixtures seeding storage state, which can use `Create` or direct setup helpers instead. Consider removing `Upsert` from the `DataStorage` interface entirely to prevent future contributors from accidentally reintroducing the unconditional-write pattern this PR eliminates. Fine as a follow-up.

pkg/cache/validating_cache.go
```diff
@@ -0,0 +1,173 @@
...
+// cache is at capacity, the least-recently-used entry is evicted first and
+// onEvict is called for it.
+func (c *ValidatingCache[K, V]) Set(key K, value V) {
+	c.lruCache.Add(key, value)
+}
+
+// Remove evicts the entry for key, calling onEvict if the key was present.
+// It is a no-op if the key is not in the cache.
+func (c *ValidatingCache[K, V]) Remove(key K) {
+	c.lruCache.Remove(key)
```
jerm-dro, line 167: **suggestion:** Per the [earlier discussion with @JAORMX](https://github.com/stacklok/toolhive/pull/4669#discussion_r3058743070), I'd prefer to keep `Remove` off the cache API. Both call sites (Terminate Phase 2 and DecorateSession's `!updated` path) are prompt-cleanup optimizations — the cache self-heals on the next `Get` via `checkSession` in both cases. Exposing `Remove` invites callers to manage cache state imperatively rather than relying on the liveness invariant, which is the whole point of this design. A background sweep that proactively runs the cache's built-in `check` function to evict invalid entries would address the eager-cleanup concern without widening the API surface — small change, fine as a follow-up.

pkg/vmcp/server/sessionmanager/session_manager.go
```diff
@@ -664,32 +638,26 @@ func (sm *Manager) GetMultiSession(sessionID string) (vmcpsession.MultiSession,
...
+	}
+
+	// Compare backend IDs to detect cross-pod metadata drift.
+	// Only compare when the cached session carries MetadataKeyBackendIDs to
+	// avoid spurious evictions for sessions that don't track backend IDs.
+	if cachedIDs, present := sess.GetMetadata()[vmcpsession.MetadataKeyBackendIDs]; present {
+		if cachedIDs != metadata[vmcpsession.MetadataKeyBackendIDs] {
+			return cache.ErrExpired
 		}
 	}
```
jerm-dro, line 662: **suggestion:** This could compare the full metadata maps with `maps.Equal` instead of just `MetadataKeyBackendIDs`. The session manager doesn't need to know which specific key drifted — if in-memory state doesn't match storage, evict and let the next restore pick up current state. This also means new metadata keys get staleness detection for free without updating `checkSession`. Both sides are `map[string]string` so `maps.Equal` works directly with no reflection overhead.

pkg/cache/validating_cache.go
```diff
@@ -0,0 +1,173 @@
...
+		return zero, false
+	}
+	r, ok := raw.(result)
+	return r.v, ok
+}
+
+// Set stores value under key, moving the entry to the MRU position. If the
+// cache is at capacity, the least-recently-used entry is evicted first and
+// onEvict is called for it.
+func (c *ValidatingCache[K, V]) Set(key K, value V) {
```
jerm-dro, line 160: **suggestion:** The singleflight currently only deduplicates cache misses (the `load` path), but moving the boundary to wrap the entire `Get` — both hit+check and miss+load — simplifies the design. With the full `Get` in singleflight, only one goroutine runs per key at a time, so the miss path can just `Add` directly — the `ContainsOrAdd` + `Get` race handling goes away entirely. It also coalesces concurrent liveness checks into a single `storage.Load` round-trip per key under concurrent access.

pkg/transport/session/session_data_storage.go
yrobla: will solve on https://github.com/stacklok/toolhive/issues/4726

pkg/cache/validating_cache.go
yrobla: https://github.com/stacklok/toolhive/issues/4727

pkg/vmcp/server/sessionmanager/session_manager.go
yrobla: https://github.com/stacklok/toolhive/issues/4728

pkg/cache/validating_cache.go
yrobla: https://github.com/stacklok/toolhive/issues/4729

PR #4680, opened 2026-04-08
Add interactive TUI dashboard for managing MCP servers

### LINKED ISSUE #4370: feat(cli): add thv tui command for an interactive terminal UI experience
## Summary

Add a `thv tui` command that launches an interactive terminal user interface (TUI) for ToolHive, improving the overall user experience for users who prefer to manage their MCP servers directly from the terminal without relying on a separate web UI.

## Background

ToolHive already provides a solid CLI experience, but some workflows — such as inspecting running servers, browsing logs, or exploring the registry — benefit from a richer, interactive interface. A TUI improves the current user experience and integrates UI-style functionality directly in the terminal, without requiring users to switch to a browser or remember individual subcommands.

## Proposed Solution

Introduce a `thv tui` subcommand that opens an interactive terminal UI (e.g. built with Bubble Tea / Lip Gloss). The TUI should include the following views:

- **Running MCP list** — overview of all currently running MCP servers
- **MCP detail** — info panel for a selected running MCP (metadata, status, configuration)
- **Logs viewer** — MCP server logs and proxy logs, with in-view search/filtering
- **Tools list** — list of tools exposed by a running MCP, each with its description
- **Registry list** — browse servers available in the configured registry, with descriptions
- **MCP Inspector** — interactive panel to craft MCP request payloads, send them to a running MCP, read the JSON response, and copy it to the clipboard

## Acceptance Criteria

- [ ] \`thv tui\` launches without errors and renders a functional TUI
- [ ] All views listed above are implemented and navigable from within the TUI
- [ ] Logs viewer supports search/filtering within the log output
- [ ] MCP Inspector allows configuring the request payload, displays the JSON response, and supports copying to clipboard
- [ ] The command is documented in \`--help\` output and in the project docs
- [ ] Unit/integration tests cover the core TUI logic

peppescg:
## Summary

Managing MCP servers through individual CLI commands requires remembering flags and switching between multiple terminal windows. An interactive dashboard lets users monitor, inspect, and operate servers from a single view.

- Add `thv tui` command with a real-time terminal dashboard built on Bubble Tea
- Workload list with status indicators, filtering, and auto-refresh
- Tabbed panels: logs, server info, tools, proxy logs, and tool inspector
- Tool inspector with form-based calling and JSON tree response viewer
- Registry browser with search, detail view, and run-from-registry support
- Log search with case-insensitive highlighting and match navigation
- Unit subtests covering pure functions across test files
- Only existing code touched: `commands.go` (register subcommand) and `mcp.go` (shared client extraction)

Closes #4370

## Type of change

- [x] New feature

## Test plan

Manually verified: launching `thv tui`, browsing workloads, switching tabs, inspecting/calling tools (with typed integer/boolean/array args), searching logs, browsing the registry, running a server from registry, quitting with `q` and Ctrl+C, and keyboard navigation including help overlay.

## Changes

| File | Change |
|------|--------|
| `cmd/thv/app/tui.go` | New `thv tui` command entry point with signal handler for terminal reset |
| `cmd/thv/app/commands.go` | Register TUI subcommand (single line addition) |
| `cmd/thv/app/mcp.go` | Refactored to use shared `pkg/mcp/client` package |
| `cmd/thv/app/ui/` | Shared UI primitives: styles, spinner, help overlay, log handler |
| `pkg/mcp/client/client.go` | New shared MCP client package (transport selection, auto-detect, init handshake) |
| `pkg/tui/model.go` | Core Bubble Tea model and state |
| `pkg/tui/init.go` | Model initialization and initial commands |
| `pkg/tui/update.go` | Main update loop and message dispatch |
| `pkg/tui/update_navigation.go` | Keyboard navigation, tab switching, clean quit with context cancellation |
| `pkg/tui/update_inspector.go` | Tool inspector form and call logic |
| `pkg/tui/update_registry.go` | Registry browsing and run-from-registry |
| `pkg/tui/update_search.go` | Log search with highlighting |
| `pkg/tui/view*.go` | View rendering (main, info, inspector, registry, status bar, helpers) |
| `pkg/tui/actions.go` | Async commands (refresh, logs, tool calls, in-process registry run) |
| `pkg/tui/inspector.go` | Inspector form field management with JSON Schema type coercion |
| `pkg/tui/registry.go` | Registry data helpers (delegates to `provider.SearchServers`) |
| `pkg/tui/json_tree.go` | Collapsible JSON tree for tool responses |
| `pkg/tui/logformat.go` | Structured log line parsing |
| `pkg/tui/logs.go` | Log streaming with error surfacing and backoff |
| `pkg/tui/*_test.go` | Test files with unit subtests |

Yes. Users can now run `thv tui` to open an interactive terminal dashboard for monitoring and managing MCP servers. The dashboard provides real-time status, log viewing, tool inspection and calling, and registry browsing — all from a single terminal window. Press `q` or `Ctrl+C` to exit the TUI at any time; the terminal background is always restored cleanly, even on signals.

- This is a large PR because it introduces an entirely new feature surface area. All files except `commands.go` and `mcp.go` are new — no existing command logic or packages were modified beyond those two.
- The TUI is built on the [Charmbracelet](https://github.com/charmbracelet) ecosystem:
  - **Bubble Tea** (`bubbletea`) — main TUI framework (already a dependency before this branch)
  - **Bubbles** (`bubbles`) — companion component library (text inputs, viewports, spinners, etc.)
  - **Lip Gloss** (`lipgloss`) — styling and layout (already a dependency before this branch)
  - **charmbracelet/x/ansi** — ANSI string width and truncation utilities
  - **atotto/clipboard** — clipboard support (the only non-Charmbracelet addition)
- Unit tests cover pure functions (parsing, formatting, filtering, tree operations, type coercion) but not the interactive Bubble Tea model itself.
- Registry "run" uses the in-process `workloads.Manager.RunWorkloadDetached` with a `RunConfig` built from registry metadata — no subprocess shelling.
- MCP client creation logic is extracted into `pkg/mcp/client` and shared between the CLI and TUI, with proper version reporting and auto-detection.
- Inspector form fields coerce values to JSON Schema types (integer, number, boolean, array, object) before sending tool calls.
- Terminal background (OSC 11) is reset on normal quit, panic, and SIGTERM/SIGINT via a signal handler, preventing "stuck background" in terminal multiplexers.

## Large PR Justification

This PR is large because it introduces an entirely self-contained new feature with minimal modifications to existing packages. The bulk of the lines are new files in `pkg/tui/` and `cmd/thv/app/ui/`. The only changes to existing code are: a single line in `commands.go` to register the subcommand, and `mcp.go` refactored to use the new shared `pkg/mcp/client` package. Splitting this into smaller PRs would either ship a non-functional intermediate state or require artificial scaffolding that would be immediately removed.


pkg/tui/inspector.go
```diff
@@ -0,0 +1,179 @@
...
+	result := make(map[string]any)
+	for _, f := range fields {
+		v := strings.TrimSpace(f.input.Value())
+		if v == "" {
+			continue
+		}
+		result[f.name] = v
+	}
+	return result
+}
```
JAORMX: Blocker: every value is going out as a string here regardless of `f.typeName`. So a tool that declares `count: integer` in its schema gets `{"count":"42"}` and rejects the call. That's the inspector's whole job, so we need to switch on the type before sending: `strconv.ParseFloat`/`ParseBool` for scalars, `json.Unmarshal` for arrays/objects, and bubble validation errors back to the form rather than silently coercing. Worth a table-driven test per schema type while you're in there.

Related: required fields aren't checked before the call goes out either, since this function silently drops empties at line 152. Pre-flight `f.required && v == ""` and refuse the call (focus the first missing field, show a notif). Otherwise users get a confusing error from the server instead of from the form.

pkg/tui/update_navigation.go
```diff
@@ -0,0 +1,531 @@
...
+
+	// Create and connect the MCP client if not already present.
+	if m.mcpClient == nil {
+		c, err := createMCPClient(sel)
+		if err != nil {
+			m.toolsLoading = false
+			m.toolsErr = err
+			return nil
+		}
+		if err := connectMCPClient(m.ctx, c); err != nil {
```
JAORMX: Blocker: `createMCPClient` and `connectMCPClient(m.ctx, c)` run synchronously inside an Update handler. So the entire TUI (input, ticks, log streaming) freezes until the MCP server responds or times out. Bubble Tea's whole contract is that Update never blocks... wrap the connect inside the returned `tea.Cmd` and post a `mcpClientReadyMsg` (or similar) on success/failure.

pkg/tui/tools.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	if _, err := c.Initialize(ctx, initReq); err != nil {
+		return fmt.Errorf("initialize MCP client: %w", err)
+	}
+	return nil
+}
+
+// fetchTools queries the MCP server for its tool list via an already-connected client.
+func fetchTools(ctx context.Context, c *mcpclient.Client) ([]mcp.Tool, error) {
+	result, err := c.ListTools(ctx, mcp.ListToolsRequest{})
```
JAORMX: So we already have `createMCPClient`, `tryInitializeClient`, `determineTransportType`, and `createMCPClientWithAutoDetect` over in `cmd/thv/app/mcp.go:309-383`. This file reimplements the SSE / streamable-http switch, hardcodes `Version: "1.0.0"` (line 45) instead of `versions.GetVersionInfo().Version`, and skips the `auto` detection altogether.

Suggestion: extract the existing helpers into `pkg/mcp/client` (or similar) and let both the CLI and the TUI consume them. At minimum, reuse `tryInitializeClient` so the version doesn't lie about itself, and pick up autodetect for free.

pkg/tui/registry.go
```diff
@@ -0,0 +1,96 @@
...
+		return items
+	}
+	q := strings.ToLower(query)
+	var out []regtypes.ServerMetadata
+	for _, item := range items {
+		if strings.Contains(strings.ToLower(item.GetName()), q) ||
+			strings.Contains(strings.ToLower(item.GetDescription()), q) {
+			out = append(out, item)
+		}
+	}
```
JAORMX: `provider.SearchServers(query)` already does this and is the same code path `cmd/thv/app/search.go:42` uses. Drop `filterRegistryItems` and call `SearchServers` so the API-backed provider can push filtering server-side instead of pulling everything down for a substring match. Keeps the TUI and `thv search` aligned on the same semantics for free.

cmd/thv/app/ui/log_handler.go
```diff
@@ -0,0 +1,114 @@
...
+	default: // drop if channel is full
+	}
+	return nil
+}
+
+// WithAttrs returns the same handler (attrs are merged into Handle output).
+func (h *TUILogHandler) WithAttrs(_ []slog.Attr) slog.Handler { return h }
+
+// WithGroup returns the same handler (groups are omitted).
+func (h *TUILogHandler) WithGroup(_ string) slog.Handler { return h }
```
JAORMX: A few things in this file:

1. `CLIHandler` is dead code in this PR. I grepped and it's only referenced inside its own file. If it's a leftover from a removed feature, delete it.

2. `WithAttrs` and `WithGroup` (lines 67, 70, 111, 114) return the same handler unchanged on both `CLIHandler` and `TUILogHandler`. That violates the slog contract: anyone doing `log.With("workload", name)` will silently lose `workload` on every subsequent record. Either accumulate attrs into a returned wrapper or document explicitly that these handlers are end-of-pipeline only and must not be `.With`-wrapped.

3. `TUILogHandler.Handle` (line 91) builds a string with `prefix + ": " + r.Message + attrs` but never includes `r.Time`. Then `formatLogLine` over in `pkg/tui/logformat.go:31` only sees this string, fails the `raw[0] != '{'` check, and renders without a timestamp. So workload-tailed logs (JSON) get timestamps in the TUI but in-app logs don't... inconsistent.

4. Bigger picture: `TUILogHandler` plus `formatLogLine` form a round-trip where we emit a string and then re-decode it as JSON to color it. If the emitted format ever changes shape, this regresses silently. Push the `slog.Record` through the channel and format once at render time.

pkg/tui/view_helpers.go
```diff
@@ -0,0 +1,185 @@
...
+	return lines
+}
+
+// runesTruncate truncates s to at most n runes, appending "..." if truncated.
+func runesTruncate(s string, n int) string {
+	r := []rune(s)
+	if len(r) <= n {
+		return s
+	}
+	return string(r[:n-1]) + "…"
```
JAORMX, line 161: `runesTruncate` panics when `n <= 0` because `r[:n-1]` becomes `r[:-1]`. `truncateSidebar` (line 162) guards this case but `runesTruncate` doesn't, and a small terminal width can drive `n` to zero. `if n <= 1 { return "…" }` at the top fixes it.

pkg/tui/update_navigation.go
```diff
@@ -0,0 +1,531 @@
...
+//nolint:gocyclo // key-handler switch; complexity is inherent to dispatching over all normal-mode key bindings
+func (m *Model) handleNormalKey(msg tea.KeyMsg) tea.Cmd {
+	switch {
+	case key.Matches(msg, keys.Quit):
+		if m.mcpClient != nil {
+			_ = m.mcpClient.Close()
+			m.mcpClient = nil
+		}
+		m.quitting = true
+		return tea.Quit
```
JAORMX, line 72: Quit closes `m.mcpClient` here but leaves `logCtxCancel`, `proxyLogCancel`, and any in-flight `exec.CommandContext` from `runFromRegistry` running until process exit. Cancel them before `tea.Quit`.

pkg/tui/logs.go
```diff
@@ -0,0 +1,124 @@
...
+			lastLines = lines
+
+			for _, l := range newLines {
+				select {
+				case ch <- l:
+				case <-ctx.Done():
+					return
+				}
+			}
+		}
```
JAORMX, line 68: Polling every 500 ms with a fixed 50-line window will silently drop lines on chatty servers or error storms. So on no overlap, `diffLines` returns `next` entirely (line 104) and anything older than the 50-line tail is gone. `manager.GetLogs(ctx, name, follow=true, lines)` already supports tail-follow... worth using for stdout/stderr at least. Proxy logs may legitimately need polling, but they should reuse the same primitive.

Also: the `if err != nil { continue }` at line 35 means a removed workload, a permission error, or a runtime restart produces a silently frozen log pane forever. At minimum, surface the first error into `ch` (or a backoff with a visible error line) so users aren't staring at an empty pane.

pkg/tui/inspector.go
```diff
@@ -0,0 +1,179 @@
...
+
+		ti := textinput.New()
+		ti.Placeholder = fieldType
+		if reqSet[name] {
+			ti.Placeholder = fieldType + " (required)"
+		}
+		ti.Width = 40
+
+		fields = append(fields, formField{
+			input:    ti,
```
JAORMX, line 71: Required-field marker only lives in the placeholder text, which vanishes the moment the user types. So once they start filling in fields, there's no visual cue left for which ones are required. Render a sticky `*` (or a colored `(required)` suffix) next to the field name in `renderFormFieldFromStruct` when `f.required`, so the marker survives input.

Also worth handling JSON-Schema union types here: `def["type"].(string)` (around line 59 of `buildInspFields` upstream) ignores the `type:["string","null"]` form which is a `[]any` in JSON. Falls back to `"string"` silently and mistypes the value. Pick the first non-null entry, or handle `oneOf`/`anyOf` defensively.

pkg/tui/inspector_test.go
```diff
@@ -0,0 +1,233 @@
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			got := formatInspResult(tc.result)
+			if tc.expected != "" {
+				assert.Equal(t, tc.expected, got)
+			} else if tc.result != nil {
+				// For non-text and empty-content cases, just verify it returns non-empty valid output
+				require.NotEmpty(t, got)
+			}
```
JAORMX, line 297: The "non-text content" and "empty content" branches only assert `require.NotEmpty(t, got)`. So a stub returning `"x"` would pass. Tighten with `assert.Contains(t, got, "base64data")` and `assert.Contains(t, got, "isError")` (or unmarshal and assert structure).

Same shape of issue in two other places worth fixing in this PR:

- `logformat_test.go:70-76` `TestLevelStyle` only asserts `NotEmpty` for every level, so it doesn't verify ERROR vs INFO produce different output or that any ANSI escape is present.
- `search_test.go` `TestHighlightSubstring` only checks the result `Contains` the original substrings... but the input line already contains them. A no-op implementation passes. Assert `result != tc.line` when a match is expected and that an ESC byte (`\x1b`) is in the output.

While you're in the test files: `buildInspFields` (`inspector.go:37`) and `buildRunFormFields` (`registry.go:42`) are pure helpers right alongside the things you already cover, but neither has tests. Both are worth a few table-driven cases (required vs optional, secret vs env, non-scalar types).

pkg/tui/actions.go
```diff
@@ -0,0 +1,128 @@
...
+		if err != nil {
+			return runFormResultMsg{
+				name:   workloadName,
+				server: serverName,
+				err:    fmt.Errorf("%w: %s", err, strings.TrimSpace(string(out))),
+			}
+		}
+		return runFormResultMsg{name: workloadName, server: serverName}
+	}
+}
```
JAORMX, line 216: This is the biggest reuse gap in the PR. So... we already hold a `workloads.Manager` (init.go:19) and we use it for stop/restart/delete. Why shell out for run?

`pkg/workloads/manager.go:RunWorkloadDetached(ctx, *runner.RunConfig)` plus `pkg/runner.NewRunConfigBuilder` is the in-process equivalent. Building a `RunConfig` from `regtypes.ServerMetadata` and calling the manager would:

- drop `os/exec` entirely (no argv injection surface, no `--` separator concern, no half-init on TUI quit)
- avoid double-binary-resolution edge cases (snap/brew wrappers, custom builds, re-exec shims where `os.Executable()` returns something surprising)
- keep TUI behaviour in lockstep with `cmd/thv/app/run.go` (today they will drift)
- save secret values from showing up in `/proc/<pid>/cmdline`

I get that the subprocess is a defensible v1, the comment explaining `--env` vs `--secret` makes that clear and is actually really helpful context. But this is the only flow in the TUI that shells out, and it's the security-sensitive one. Worth doing now if cheap, otherwise a tracked follow-up before this lands in a release.

Two narrower nits if the subprocess stays:

1. Line 80: there's no `--` separator before the registry-derived `serverName` positional. Today this isn't exploitable (Cobra would consume `--name` as the flag value or reject), but it's brittle. Either insert `"--"` ahead of the positional or reject names matching `^-`.
2. Line 117-118: `cmd.CombinedOutput()` is bound to `m.ctx`. If the user quits the TUI mid-launch, `exec.CommandContext` SIGKILLs `thv run` and you can end up with a half-initialised workload (container started, runconfig not persisted). Detach with `context.WithoutCancel(m.ctx)` so the spawned process outlives the TUI session.

cmd/thv/app/ui/log_handler.go
JAORMX: A few things in this file:

1. `CLIHandler` is dead code in this PR. I grepped and it's only referenced inside its own file. If it's a leftover from a removed feature, delete it.

2. `WithAttrs` and `WithGroup` (lines 67, 70, 111, 114) return the same handler unchanged on both `CLIHandler` and `TUILogHandler`. That violates the slog contract: anyone doing `log.With("workload", name)` will silently lose `workload` on every subsequent record. Either accumulate attrs into a returned wrapper or document explicitly that these handlers are end-of-pipeline only and must not be `.With`-wrapped.

3. `TUILogHandler.Handle` (line 91) builds a string with `prefix + ": " + r.Message + attrs` but never includes `r.Time`. Then `formatLogLine` over in `pkg/tui/logformat.go:31` only sees this string, fails the `raw[0] != '{'` check, and renders without a timestamp. So workload-tailed logs (JSON) get timestamps in the TUI but in-app logs don't... inconsistent.

4. Bigger picture: `TUILogHandler` plus `formatLogLine` form a round-trip where we emit a string and then re-decode it as JSON to color it. If the emitted format ever changes shape, this regresses silently. Push the `slog.Record` through the channel and format once at render time.

JAORMX (commented):
Big PR but very self-contained, almost entirely new files in `pkg/tui/` and `cmd/thv/app/ui/` plus a one-line registration. So the size justification is sound and reviewing it as one chunk is the right call.

Conventions look clean across the board: SPDX headers everywhere, `%w` wrapping, no `logger.Fatal`, no leftover TODOs, and tests follow the project pattern (stdlib testing + testify, `t.Parallel`, table-driven, SPDX). The bones of the design are solid too: bounded log buffers, consistent stream cancellation on selection change, stable iteration via `slices.Sort`, `os.Executable()` rather than `LookPath`, single-quote escaping in the copy-as-curl path with a malicious-input test backing it. Nice work overall.

That said, there's a meaningful punch list before this is mergeable. Inline comments have the specifics. Quick summary:

### Blockers

- Inspector type coercion: every argument goes out as a string regardless of schema type. A tool declaring `count: integer` rejects every call. (`pkg/tui/inspector.go:147-157`)
- Required field validation isn't run before the tool call. (`pkg/tui/update_inspector.go:323-343`)
- Blocking MCP I/O on the Update goroutine freezes the TUI. (`pkg/tui/update_navigation.go:391-405`)

### Reuse worth taking on now

The biggest one by a wide margin: `runFromRegistry` shells out to `thv run` from inside `thv` itself when the TUI already holds a `workloads.Manager` and uses it for every other action. Calling `manager.RunWorkloadDetached` instead removes around 150 lines, eliminates the entire argv-injection surface, and stops the TUI and CLI from drifting from each other. That alone is worth doing here.

The other reuse gaps are all in the same spirit and called out inline:

- MCP client init duplicates `cmd/thv/app/mcp.go:createMCPClient` / `tryInitializeClient` / `determineTransportType` / `createMCPClientWithAutoDetect`.
- `filterRegistryItems` reimplements `provider.SearchServers`.
- `cmd/thv/app/ui/log_handler.go` has a `CLIHandler` that's dead code, and both handlers' `WithAttrs`/`WithGroup` violate the slog contract.

### Other notable items

- `runesTruncate` panics on `n <= 0` (`view_helpers.go:158`).
- Search prompt swallows `n` and `N` (`update_search.go:43-52`)... real UX bug.
- Quit doesn't cancel log/proxy contexts.
- Log polling silently drops lines on chatty streams (`logs.go:31-50`).
- A few tautological tests in `inspector_test.go`, `logformat_test.go`, and `search_test.go`.

### Suggested path

1. Fix the three blockers and the `runesTruncate` panic in this PR.
2. Take the `runFromRegistry` to `RunWorkloadDetached` path now if the lift is small, otherwise open a tracked follow-up. Same for the shared MCP client extraction.
3. Tighten the tautological tests and add coverage for `buildInspFields` / `buildRunFormFields` / `runFormSubmit` validation while you're already in the test files.
4. Log polling to tail-follow, large-array elision in the JSON tree, and a `pkg/tui.Run(...)` extraction (so `tuiCmdFunc` stays a thin wrapper) are reasonable follow-ups for separate PRs.

Thanks for keeping this self-contained, by the way. Reviewing 7k lines is a lot, but having the only existing-code touch be a single line in `commands.go` makes it tractable.

peppescg:
Fixed the registry name filtering on https://github.com/stacklok/toolhive/pull/4680/changes/3150020bbe9d11aa142c2eb2476ae6ebf1a251bc

Screenshot 2026-04-27 at 18 48 00

peppescg:
@claude succinct review

JAORMX (approved):

PR #4719, opened 2026-04-09
Add TelemetryConfigRef support to MCPRemoteProxy

### LINKED ISSUE #4620: Add TelemetryConfigRef support to MCPRemoteProxy
## Summary

MCPRemoteProxy only supports inline telemetry configuration, unlike MCPServer which supports both inline and reference-based telemetry. This means MCPRemoteProxy users cannot reference a shared MCPTelemetryConfig resource and must duplicate inline telemetry configuration across every instance.

**Severity**: SHOULD FIX
**Area**: CRD API Consistency
**Breaking**: No

## Location

- `cmd/thv-operator/api/v1alpha1/mcpremoteproxy_types.go:97-99`

## Problem

MCPRemoteProxy:
```go
Telemetry *TelemetryConfig `json:"telemetry,omitempty"`
```

MCPServer (supports both patterns):
```go
Telemetry          *TelemetryConfig    `json:"telemetry,omitempty"`
TelemetryConfigRef *TelemetryConfigRef `json:"telemetryConfigRef,omitempty"`
```

## Impact

- Configuration duplication across MCPRemoteProxy instances
- Cannot use centralized MCPTelemetryConfig resources for remote proxies
- Inconsistency with MCPServer API, which supports both patterns
- If the inline `Telemetry` field is deprecated (as the MCPServer pattern suggests), MCPRemoteProxy has no migration path

## Recommended Fix

1. Add `TelemetryConfigRef` to MCPRemoteProxySpec:
   ```go
   // TelemetryConfigRef references a shared MCPTelemetryConfig.
   // +optional
   TelemetryConfigRef *TelemetryConfigRef `json:"telemetryConfigRef,omitempty"`
   ```

2. Add CEL validation for mutual exclusivity (matching MCPServer pattern):
   ```go
   // +kubebuilder:validation:XValidation:rule=
   //   "!(has(self.telemetry) && has(self.telemetryConfigRef))",
   //   message="telemetry and telemetryConfigRef are mutually exclusive"
   ```

3. Deprecate the inline `Telemetry` field to match the MCPServer migration direction.

## Related Issues

- MCPRemoteProxy missing Replicas field — another MCPRemoteProxy API gap

ChrisJBurns:
## Summary

MCPRemoteProxy lacked the ability to reference shared `MCPTelemetryConfig` resources, forcing users to duplicate inline telemetry configuration across every proxy instance. This brings MCPRemoteProxy to parity with MCPServer's telemetry API by adding `TelemetryConfigRef` support.

Closes #4620

## Type of change

- [x] New feature

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/api/v1alpha1/mcpremoteproxy_types.go` | Add `TelemetryConfigRef` field with CEL mutual exclusivity validation, deprecate inline `Telemetry`, add `TelemetryConfigHash` status field and condition constants |
| `cmd/thv-operator/pkg/controllerutil/config.go` | Add `GetTelemetryConfigForMCPRemoteProxy()` helper — namespace-scoped fetch returning (nil, nil) for not-found |
| `cmd/thv-operator/controllers/mcpremoteproxy_controller.go` | Add `handleTelemetryConfig()` handler, call in `validateAndHandleConfigs()`, add `mapTelemetryConfigToMCPRemoteProxy()` watch mapper, register `MCPTelemetryConfig` watch in `SetupWithManager()` |
| `cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go` | Prefer `TelemetryConfigRef` over inline `Telemetry` in RunConfig; extract `addTelemetryOptions` and `resolveToolConfig` helpers to fix cyclomatic complexity |

## Test plan

- [x] `task lint-fix` passes (0 issues)
- [x] `task test` passes (all operator unit tests)
- [x] `go build ./...` passes
- [x] New unit tests: `TestGetTelemetryConfigForMCPRemoteProxy` (4 cases), `TestHandleTelemetryConfig_MCPRemoteProxy` (5 cases), `TestMapTelemetryConfigToMCPRemoteProxy`
- [x] Existing `TestCreateRunConfigFromMCPRemoteProxy` tests still pass with updated `ctx` parameter

- The `resolveToolConfig` extraction in `mcpremoteproxy_runconfig.go` is a refactor needed to keep cyclomatic complexity under the lint threshold (was already at 15 pre-change). No behavioral change.
- CEL validation is tested via the generated CRD manifest; integration test coverage for MCPRemoteProxy CEL rules can follow in a separate PR.
- VirtualMCPServer telemetry config ref is out of scope (separate issue).

### DIFF
--- a/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
+++ b/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
@@ -48,6 +48,7 @@ type MCPRemoteProxyReconciler struct {
 // +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcptoolconfigs,verbs=get;list;watch
 // +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpexternalauthconfigs,verbs=get;list;watch
 // +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpoidcconfigs,verbs=get;list;watch
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcptelemetryconfigs,verbs=get;list;watch
 // +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpoidcconfigs/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=create;delete;get;list;patch;update;watch
 // +kubebuilder:rbac:groups="",resources=services,verbs=create;delete;get;list;patch;update;watch
@@ -657,9 +658,11 @@ func (r *MCPRemoteProxyReconciler) handleTelemetryConfig(ctx context.Context, pr
 	ctxLogger := log.FromContext(ctx)
 
 	if proxy.Spec.TelemetryConfigRef == nil {
-		// No MCPTelemetryConfig referenced, clear any stored hash
-		meta.RemoveStatusCondition(&proxy.Status.Conditions, mcpv1alpha1.ConditionTypeMCPRemoteProxyTelemetryConfigRefValidated)
-		if proxy.Status.TelemetryConfigHash != "" {
+		// No MCPTelemetryConfig referenced, clear any stored hash and condition.
+		condType := mcpv1alpha1.ConditionTypeMCPRemoteProxyTelemetryConfigRefValidated
+		condRemoved := meta.FindStatusCondition(proxy.Status.Conditions, condType) != nil
+		meta.RemoveStatusCondition(&proxy.Status.Conditions, condType)
+		if condRemoved || proxy.Status.TelemetryConfigHash != "" {
 			proxy.Status.TelemetryConfigHash = ""
 			if err := r.Status().Update(ctx, proxy); err != nil {
 				return fmt.Errorf("failed to clear MCPTelemetryConfig hash from status: %w", err)
--- a/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
+++ b/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
@@ -709,7 +709,8 @@ func (r *MCPRemoteProxyReconciler) handleTelemetryConfig(ctx context.Context, pr
 	// Detect whether the condition is transitioning to True (e.g. recovering from
 	// a transient error). Without this check the status update is skipped when the
 	// hash is unchanged, leaving a stale False condition.
-	prevCondition := meta.FindStatusCondition(proxy.Status.Conditions, mcpv1alpha1.ConditionTypeMCPRemoteProxyTelemetryConfigRefValidated)
+	condType := mcpv1alpha1.ConditionTypeMCPRemoteProxyTelemetryConfigRefValidated
+	prevCondition := meta.FindStatusCondition(proxy.Status.Conditions, condType)
 	needsUpdate := prevCondition == nil || prevCondition.Status != metav1.ConditionTrue
 
 	meta.SetStatusCondition(&proxy.Status.Conditions, metav1.Condition{
--- a/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
+++ b/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
@@ -128,6 +128,16 @@ func (r *MCPRemoteProxyReconciler) validateAndHandleConfigs(ctx context.Context,
 		return err
 	}
 
+	// Handle MCPTelemetryConfig
+	if err := r.handleTelemetryConfig(ctx, proxy); err != nil {
+		ctxLogger.Error(err, "Failed to handle MCPTelemetryConfig")
+		proxy.Status.Phase = mcpv1alpha1.MCPRemoteProxyPhaseFailed
+		if statusErr := r.Status().Update(ctx, proxy); statusErr != nil {
+			ctxLogger.Error(statusErr, "Failed to update MCPRemoteProxy status after MCPTelemetryConfig error")
+		}
+		return err
+	}
+
 	// Handle MCPExternalAuthConfig
 	if err := r.handleExternalAuthConfig(ctx, proxy); err != nil {
 		ctxLogger.Error(err, "Failed to handle MCPExternalAuthConfig")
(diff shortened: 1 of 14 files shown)


cmd/thv-operator/controllers/mcpremoteproxy_telemetryconfig_test.go
```diff
@@ -0,0 +1,235 @@
...
+			err := reconciler.handleTelemetryConfig(ctx, tt.proxy)
+
+			if tt.expectError {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+
+			if tt.expectNoCondition {
+				for _, c := range tt.proxy.Status.Conditions {
```
jhrozek: **[HIGH] Test assertions read in-memory state, not persisted state**

All assertions in this loop read from `tt.proxy.Status.*` — the same pointer that `meta.SetStatusCondition` / `meta.RemoveStatusCondition` already mutated *before* `r.Status().Update()` was called. These assertions would pass even if the `Status().Update()` call were deleted entirely. `WithStatusSubresource` is correctly set up but has no practical effect because `fakeClient.Get` is never called.

The pattern from `TestMCPServerReconciler_handleOIDCConfig_ConditionPersistedOnRecovery` is the right model: after calling the handler, re-fetch and assert on the persisted object:
```go
persisted := &mcpv1alpha1.MCPRemoteProxy{}
require.NoError(t, fakeClient.Get(ctx, types.NamespacedName{
    Name: tt.proxy.Name, Namespace: tt.proxy.Namespace,
}, persisted))
// assert on persisted.Status.* instead of tt.proxy.Status.*
```

While fixing this, also add a recovery-from-False scenario: proxy has a hash matching the config's hash but a stale `False` condition — assert the `True` condition is persisted. This covers the `needsUpdate` logic at line 714 which is currently untested.

cmd/thv-operator/controllers/mcpremoteproxy_controller.go
```diff
@@ -641,6 +651,95 @@ func (r *MCPRemoteProxyReconciler) handleToolConfig(ctx context.Context, proxy *
...
 }
 
+// handleTelemetryConfig validates and tracks the hash of the referenced MCPTelemetryConfig.
+// It updates the MCPRemoteProxy status when the telemetry configuration changes.
+func (r *MCPRemoteProxyReconciler) handleTelemetryConfig(ctx context.Context, proxy *mcpv1alpha1.MCPRemoteProxy) error {
+	ctxLogger := log.FromContext(ctx)
+
+	if proxy.Spec.TelemetryConfigRef == nil {
+		// No MCPTelemetryConfig referenced, clear any stored hash
+		meta.RemoveStatusCondition(&proxy.Status.Conditions, mcpv1alpha1.ConditionTypeMCPRemoteProxyTelemetryConfigRefValidated)
```
jhrozek: **[MEDIUM] Condition removal not persisted when `TelemetryConfigHash` is already empty**

When `TelemetryConfigRef` is removed from a proxy whose hash is already `""` (e.g., a stale `False` condition remains from a prior error), `meta.RemoveStatusCondition` modifies the in-memory slice but `r.Status().Update` is skipped. The controller re-fetches on the next reconcile, discarding the change — the stale condition stays in the cluster.

```suggestion
	if proxy.Spec.TelemetryConfigRef == nil {
		// No MCPTelemetryConfig referenced, clear any stored hash and condition.
		condRemoved := meta.FindStatusCondition(proxy.Status.Conditions, mcpv1alpha1.ConditionTypeMCPRemoteProxyTelemetryConfigRefValidated) != nil
		meta.RemoveStatusCondition(&proxy.Status.Conditions, mcpv1alpha1.ConditionTypeMCPRemoteProxyTelemetryConfigRefValidated)
		if condRemoved || proxy.Status.TelemetryConfigHash != "" {
			proxy.Status.TelemetryConfigHash = ""
			if err := r.Status().Update(ctx, proxy); err != nil {
				return fmt.Errorf("failed to clear MCPTelemetryConfig hash from status: %w", err)
			}
		}
		return nil
	}
```

jhrozek (commented):
Code review from automated analysis. Five findings — three require changes in files not touched by this PR (prose below), two have inline comments on the diff.

---

### [HIGH] Missing `+kubebuilder:rbac` annotation for `mcptelemetryconfigs`

**File**: `cmd/thv-operator/controllers/mcpremoteproxy_controller.go` (around line 48–51)

The `MCPRemoteProxyReconciler` now lists and gets `MCPTelemetryConfig` objects (in `mapTelemetryConfigToMCPRemoteProxy` and via `GetTelemetryConfigForMCPRemoteProxy`), but there is no `+kubebuilder:rbac` marker for `mcptelemetryconfigs`. `task operator-manifests` generates the ClusterRole from these markers — without this one, the controller gets permission-denied errors at runtime the first time any proxy uses `TelemetryConfigRef`.

`MCPServerReconciler` already has:
```go
// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcptelemetryconfigs,verbs=get;list;watch
```
Add the same line to `MCPRemoteProxyReconciler` alongside the existing `mcptoolconfigs`/`mcpoidcconfigs` markers.

---

### [HIGH] `MCPTelemetryConfigReconciler.findReferencingWorkloads` does not scan `MCPRemoteProxy`

**File**: `cmd/thv-operator/controllers/mcptelemetryconfig_controller.go` line 259 (**not changed in this PR**)

`findReferencingWorkloads` only calls `FindWorkloadRefsFromMCPServers`. It never looks at `MCPRemoteProxy`, so:
1. **Deletion protection is broken**: an operator can delete an `MCPTelemetryConfig` referenced only by a proxy — the finalizer check passes.
2. `status.referencingWorkloads` will never include proxy names.
3. The reverse-direction watch (proxy changes → re-evaluate `MCPTelemetryConfig`) is also missing from `SetupWithManager`.

`MCPOIDCConfigReconciler` and `MCPExternalAuthConfigReconciler` were both extended to include `MCPRemoteProxy` when those ref fields were added. The same pattern needs to be applied here.

---

### [HIGH] Missing CA bundle volume mount for `TelemetryConfigRef` in proxy deployment

**File**: `cmd/thv-operator/controllers/mcpremoteproxy_deployment.go` (`buildVolumesForProxy`, **not changed in this PR**)

`addTelemetryOptions` correctly calls `ctrlutil.TelemetryCABundleFilePath(telCfg)` and embeds the path in the RunConfig. At pod startup the proxy runner opens that path via `os.ReadFile` to build a custom TLS config for the OTLP collector.

However, `buildVolumesForProxy` never calls `ctrlutil.AddTelemetryCABundleVolumes`. The ConfigMap is never mounted into the pod. Any user who sets `spec.openTelemetry.caBundleRef` in their `MCPTelemetryConfig` will get a file-not-found error at proxy startup.

`MCPServerReconciler.deploymentForMCPServer` already handles this correctly — the same block needs to be added to `deploymentForMCPRemoteProxy`.

cmd/thv-operator/controllers/mcpremoteproxy_telemetryconfig_test.go
jhrozek: **[HIGH] Test assertions read in-memory state, not persisted state**

All assertions here read from `tt.proxy.Status.*` — the same pointer that `meta.SetStatusCondition` / `meta.RemoveStatusCondition` already mutated *before* `r.Status().Update()` was called. These assertions would pass even if the `Status().Update()` call were deleted entirely. `WithStatusSubresource` is correctly set up but has no practical effect because `fakeClient.Get` is never called to verify what was actually persisted.

The pattern from `TestMCPServerReconciler_handleOIDCConfig_ConditionPersistedOnRecovery` is the right model — re-fetch from `fakeClient` after the handler call and assert on the result:
```go
persisted := &mcpv1alpha1.MCPRemoteProxy{}
require.NoError(t, fakeClient.Get(ctx, types.NamespacedName{
    Name: tt.proxy.Name, Namespace: tt.proxy.Namespace,
}, persisted))
// assert on persisted.Status.* instead of tt.proxy.Status.*
```

While fixing this, also add a recovery-from-False scenario: proxy has a hash matching the config's hash but a stale `False` condition — assert the `True` condition is persisted after the handler runs. This covers the `needsUpdate` logic at line 714 which is currently untested.

cmd/thv-operator/controllers/mcpremoteproxy_controller.go
jhrozek: **[MEDIUM] Condition removal not persisted when `TelemetryConfigHash` is already empty**

When `TelemetryConfigRef` is removed from a proxy whose hash is already `""` (e.g., a stale `False` condition remains from a prior error cycle), `meta.RemoveStatusCondition` modifies the in-memory slice but `r.Status().Update` is skipped. The controller re-fetches from the API server on the next reconcile, discarding the in-memory change — the stale condition stays in the cluster indefinitely.

```suggestion
	if proxy.Spec.TelemetryConfigRef == nil {
		// No MCPTelemetryConfig referenced, clear any stored hash and condition.
		condRemoved := meta.FindStatusCondition(proxy.Status.Conditions, mcpv1alpha1.ConditionTypeMCPRemoteProxyTelemetryConfigRefValidated) != nil
		meta.RemoveStatusCondition(&proxy.Status.Conditions, mcpv1alpha1.ConditionTypeMCPRemoteProxyTelemetryConfigRefValidated)
		if condRemoved || proxy.Status.TelemetryConfigHash != "" {
			proxy.Status.TelemetryConfigHash = ""
			if err := r.Status().Update(ctx, proxy); err != nil {
				return fmt.Errorf("failed to clear MCPTelemetryConfig hash from status: %w", err)
			}
		}
		return nil
	}
```

JAORMX (dismissed):

ChrisJBurns:
Thanks for the thorough review @jhrozek — all five findings are addressed in 09a948ec.

### [HIGH] Missing `+kubebuilder:rbac` annotation

Added `// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcptelemetryconfigs,verbs=get;list;watch` to `MCPRemoteProxyReconciler` alongside the existing `mcptoolconfigs`/`mcpoidcconfigs` markers. The generated ClusterRole already had this permission (from the `MCPTelemetryConfigReconciler`), but the marker is needed for correctness and documentation.

### [HIGH] `findReferencingWorkloads` does not scan MCPRemoteProxy

Extended `findReferencingWorkloads` to call `FindReferencingMCPRemoteProxies` and merge the results with `MCPServer` refs. Also added `mapMCPRemoteProxyToTelemetryConfig` and registered it as a watch in `SetupWithManager`, following the same pattern as `MCPExternalAuthConfigReconciler` and `MCPOIDCConfigReconciler`. This fixes deletion protection and `status.referencingWorkloads` for proxy-only references.

### [HIGH] Missing CA bundle volume mount

Added `addTelemetryCABundleVolumes` helper to `mcpremoteproxy_deployment.go` and called it from `deploymentForMCPRemoteProxy` right after `buildVolumesForProxy`. Follows the exact pattern from `MCPServerReconciler.deploymentForMCPServer` (line 1220).

cmd/thv-operator/controllers/mcpremoteproxy_telemetryconfig_test.go
ChrisJBurns: Fixed in 09a948ec. Tests now re-fetch persisted state via `fakeClient.Get()` and assert on the result for success paths. For error paths, assertions use in-memory state since the handler sets conditions but relies on the caller (`validateAndHandleConfigs`) to persist.

Also added two new test cases as suggested:
- **"recovery from False condition persists True"** — proxy has matching hash but a stale `False` condition from a prior fetch error; asserts the `True` condition is persisted after recovery. This covers the `needsUpdate` logic.
- **"nil ref with stale condition persists removal"** — proxy has no hash but a stale `False` condition; asserts the condition is removed in persisted state. This covers the fix for the condition-removal persistence bug.

cmd/thv-operator/controllers/mcpremoteproxy_controller.go
ChrisJBurns: Fixed in 09a948ec. Applied the suggested code exactly — now checks whether a condition exists before deciding to skip the status update. This ensures the stale condition is always persisted as removed, even when the hash was already empty.

Added a dedicated test case "nil ref with stale condition persists removal" that verifies this path by setting up a proxy with no hash but a stale False condition, then asserting the condition is gone from persisted state after the handler runs.

cmd/thv-operator/controllers/mcpremoteproxy_telemetryconfig_test.go
ChrisJBurns: Addressed — see reply above on the other thread.

cmd/thv-operator/controllers/mcpremoteproxy_controller.go
ChrisJBurns: Addressed — see reply above on the other thread.

jhrozek (approved):
All five review findings have been addressed in the latest push:

1. **RBAC annotation** — `+kubebuilder:rbac` for `mcptelemetryconfigs` added to `MCPRemoteProxyReconciler`
2. **findReferencingWorkloads** — now scans `MCPRemoteProxy` via `FindReferencingMCPRemoteProxies`, plus `Watches` and mapper added to `MCPTelemetryConfigReconciler.SetupWithManager`
3. **CA bundle volume mount** — new `addTelemetryCABundleVolumes` method wires up the ConfigMap volume in the proxy deployment
4. **Test assertions** — tests now re-fetch persisted state via `fakeClient.Get` instead of asserting on in-memory pointers
5. **nil-ref condition persistence** — `condRemoved || hash != ""` guard ensures the status update fires even when the hash was already empty

New recovery-from-False and stale-condition-removal test cases also added. Integration tests extended for the telemetry config controller. Looks good.

PR #4722, opened 2026-04-10
Return 503 for expired proxy tokens

### LINKED ISSUE #4721: Expired proxy tokens block MCP client reconnection
## Bug description

When the token injection middleware cannot retrieve an OAuth token (e.g., during a token refresh failure), it returns HTTP 401 Unauthorized to the MCP client. MCP SDKs interpret 401 as a signal to begin OAuth authentication with the server. The proxy's well-known endpoint correctly reports that no client-side OAuth is configured (HTTP 404), so the SDK's discovery attempt fails. In observed behavior, the client then marks the server as requiring authentication it cannot provide, preventing reconnection even after the proxy's token refreshes successfully.

The proxy manages OAuth internally — MCP clients should never need to authenticate with the proxy itself. Returning 401 for an internal credential failure is semantically incorrect and causes a harmful interaction with SDK auth logic.

## Steps to reproduce

Any HTTP server returning 401 triggers this. A minimal repro:

1. Start an HTTP server that returns 401 on POST to `/mcp`
2. Configure it as a streamable-HTTP MCP server in an MCP client
3. The client attempts OAuth discovery (`GET /.well-known/oauth-protected-resource`) instead of treating it as a connection error

In production, this occurs when a remote server's OAuth access token expires and the token refresh fails (transient server error, HTML error page instead of JSON, etc.).

## Expected behavior

The proxy should return 503 Service Unavailable with a `Retry-After` header when its internal token source fails. This tells MCP clients the server is temporarily unavailable without triggering OAuth discovery.

## Actual behavior

The proxy returns 401 Unauthorized. The MCP client enters an OAuth discovery flow against the proxy, which fails. The client caches this failure, preventing reconnection even after the token refreshes successfully.

## Additional context

In the bearer-token authentication flow, the 401 originates from the token injection middleware at `pkg/transport/middleware/token_injection.go`.

gkatz2:
## Summary

- The token injection middleware returns 401 when the OAuth token source fails, which MCP SDKs interpret as "begin OAuth authentication." Since the proxy manages OAuth internally and has no client-facing OAuth metadata, the SDK's discovery attempt fails and the client caches the server as unreachable.
- Change the error response to 503 Service Unavailable with `Retry-After: 10`, which clients treat as a transient connection error.
- Add unit tests for `CreateTokenInjectionMiddleware` (previously untested).

Fixes #4721

## Type of change

- [x] Bug fix

## Test plan

- [x] Manual testing: started a test HTTP server returning 503 on the token error path, confirmed MCP SDK (Claude Code) does not enter OAuth discovery. Compared against a 401 server where the SDK does enter OAuth discovery and caches "needs authentication."

MCP clients connecting through the proxy will see 503 instead of 401 when the proxy's OAuth tokens are expired. This is a behavioral change, but 503 is the correct status for this situation and prevents clients from entering a broken OAuth flow.

### DIFF
--- a/pkg/transport/middleware/token_injection.go
+++ b/pkg/transport/middleware/token_injection.go
@@ -15,7 +15,9 @@ import (
 )
 
 // CreateTokenInjectionMiddleware returns a middleware that injects a Bearer token
-// from the provided oauth2.TokenSource. It returns 401 when the workload is unauthenticated.
+// from the provided oauth2.TokenSource. It returns 503 Service Unavailable with a
+// Retry-After header when the token cannot be retrieved, so that MCP clients treat
+// the failure as transient rather than initiating an OAuth discovery flow.
 func CreateTokenInjectionMiddleware(tokenSource oauth2.TokenSource) types.MiddlewareFunction {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -24,8 +26,11 @@ func CreateTokenInjectionMiddleware(tokenSource oauth2.TokenSource) types.Middle
 				if err != nil {
 					slog.Warn("unable to retrieve OAuth token", "error", err)
 					// The token source (AuthenticatedTokenSource) handles marking
-					// the workload as unauthenticated in its Token() method
-					http.Error(w, "Authentication required", http.StatusUnauthorized)
+					// the workload as unauthenticated in its Token() method.
+					// Return 503 instead of 401 so MCP clients do not mistake this
+					// for a server that requires client-side OAuth authentication.
+					w.Header().Set("Retry-After", "10")
+					http.Error(w, "Token temporarily unavailable", http.StatusServiceUnavailable)
 					return
 				}
 
--- a/pkg/transport/middleware/token_injection.go
+++ b/pkg/transport/middleware/token_injection.go
@@ -14,6 +14,11 @@ import (
 	"github.com/stacklok/toolhive/pkg/transport/types"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/transport/middleware/token_injection.go
```diff
@@ -24,8 +26,11 @@ func CreateTokenInjectionMiddleware(tokenSource oauth2.TokenSource) types.Middle
 				if err != nil {
 					slog.Warn("unable to retrieve OAuth token", "error", err)
 					// The token source (AuthenticatedTokenSource) handles marking
-					// the workload as unauthenticated in its Token() method
-					http.Error(w, "Authentication required", http.StatusUnauthorized)
+					// the workload as unauthenticated in its Token() method.
+					// Return 503 instead of 401 so MCP clients do not mistake this
+					// for a server that requires client-side OAuth authentication.
+					w.Header().Set("Retry-After", "10")
```
jhrozek: Nit (feel free to take or leave): consider extracting `"10"` into a named constant with a short comment explaining it matches the initial `MonitoredTokenSource` backoff interval. Makes it easier to find if the backoff config ever changes.

```go
// retryAfterSecs tells MCP clients how long to wait before retrying.
// Matches the initial MonitoredTokenSource backoff interval.
const retryAfterSecs = "10"
```

jhrozek (dismissed):
Looks good to me! One nit below.

pkg/transport/middleware/token_injection.go
gkatz2: Good call. Fixed.

gkatz2:
@jhrozek, addressed the nit in the latest push. Ready for another look when someone has a chance.

jhrozek (approved):

PR #4723, opened 2026-04-10
Replace flaky proxy-restart test with cross-proxy routing

jerm-dro:
## Summary

The backend routing E2E test (`mcpserver_scaling_test.go`) deleted the proxy runner pod and waited for Kubernetes to recreate it, which was timing-sensitive — the pod often failed to come back up within the timeout window, leaving the VirtualMCPServer stuck in Pending and producing empty tool lists.

- Replace the pod-restart approach with a 2-proxy-replica deployment that port-forwards to each proxy pod directly
- Initialize a session on proxy A, then send a request to proxy B with the same session ID
- This verifies Redis-backed session sharing across proxy replicas without any pod lifecycle operations

The test now follows the same proven pattern as the existing "replicas=2 with Redis session storage" context in the same file.

## Type of change

- [x] Bug fix

## Test plan

- [x] E2E tests — this PR *is* the test fix

Net -39 lines. The new test is structurally simpler (no pod deletion, no wait-for-new-pod, no NodePort service) and deterministic (port-forward guarantees which proxy receives each request).

### DIFF
--- a/test/e2e/thv-operator/virtualmcp/mcpserver_scaling_test.go
+++ b/test/e2e/thv-operator/virtualmcp/mcpserver_scaling_test.go
@@ -26,7 +26,6 @@ import (
 
 	mcpv1alpha1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1alpha1"
 	"github.com/stacklok/toolhive/test/e2e/images"
-	"github.com/stacklok/toolhive/test/e2e/thv-operator/testutil"
 )
 
 // deployRedis creates a single-replica Redis Deployment and ClusterIP Service.
@@ -147,6 +146,8 @@ func waitForMCPServerRunning(name, namespace string, timeout, pollInterval time.
 
 // portForwardToPod starts a kubectl port-forward to a specific pod and returns the
 // local port and a cleanup function. The caller must call cleanup to stop the port-forward.
+//
+//nolint:unparam // namespace kept as parameter for reusability across test contexts
 func portForwardToPod(podName, namespace string, targetPort int32) (int, func(), error) {
 	// Find a free local port
 	listener, err := net.Listen("tcp", ":0")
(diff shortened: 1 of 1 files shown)


test/e2e/thv-operator/virtualmcp/mcpserver_scaling_test.go
```diff
@@ -280,84 +258,71 @@ var _ = ginkgo.Describe("MCPServer Cross-Replica Session Routing with Redis", fu
...
-			ginkgo.By("Sending 5 requests with the recovered session to verify backend routing")
+			gomega.Expect(clientB.Start(startCtx)).To(gomega.Succeed())
+
+			// Proxy B must route the session's backend requests to the correct
+			// backend pod (the one that handled initialize on proxy A). With 2
+			// backends and random ClusterIP routing, P(all 5 hit the right pod
+			// by chance) ≈ 3%, so 5 consecutive successes give high confidence
+			// that Redis-backed session routing is working.
+			ginkgo.By("Sending 5 requests on proxy B to verify consistent backend routing")
 			for i := range 5 {
```
jhrozek, line 318: Nit: the comment above correctly states P(false pass) ≈ 3% with 5 requests and 2 backends. That's just above the 1% threshold that's generally comfortable for CI. Bumping to 7 iterations would bring it to ~0.78%, or 10 for ~0.1%, if you ever wanted tighter statistical confidence here.

jhrozek (approved):
Nit: two small suggestions in case you wanted to make this even better — both are optional and not blocking.

test/e2e/thv-operator/virtualmcp/mcpserver_scaling_test.go
jerm-dro: In theory, yes, but I know the implementation actually caches the connection state after the first use. This is good enough for now. If it flakes, it indicates a problem.

PR #4731, opened 2026-04-10
Add LRU capacity to ValidatingCache, remove sentinel pattern, add storage Update

### LINKED ISSUE #4494: vmcp: address RestorableCache capacity limit and sentinel pattern complexity (follow-up to #4464)

yrobla:
## Summary

Re-applies the changes from #4669 which were reverted in #4730.

- Reintroduces `pkg/cache.ValidatingCache[K,V]`: a fully-typed, capacity-bounded LRU cache using mutex+map+container/list
- Removes the `terminatedSentinel` / sentinel+rollback pattern
- Adds `DataStorage.Update` (SET XX semantics) to close the TOCTOU resurrection window
- Adds `FactoryConfig.CacheCapacity` to wire the LRU limit through `sessionmanager.New`
- Cache tests live in `pkg/cache/validating_cache_test.go`

Fixes #4494

## Type of change

- [x] New feature

## Test plan

## Large PR Justification

This is the revert of an existing PR. Cannot be split.


jerm-dro (dismissed):
Some failing tests that seem minor.

I'm assuming there's no major changes here compared to the original, but LMK if that's not true

JAORMX (approved):

PR #4740, opened 2026-04-10
Fix GetDefaultProvider bypassing registered config factory

reyortiz3:
## Summary

`GetDefaultProvider()` in `pkg/registry/factory.go` called `config.NewDefaultProvider()`, which bypasses the `RegisterProviderFactory` hook entirely and always reads the local XDG config file. This means any enterprise build that registers a factory (e.g. to supply an `EnterpriseProvider` that fetches config from a remote server) was silently ignored — `thv registry` commands always showed the embedded registry instead of the configured one.

- Change the call from `config.NewDefaultProvider()` to `config.NewProvider()`, which checks the registered factory first and falls back to the default only when no factory is registered
- Add unit tests covering: factory respected, factory returns nil falls through, caching semantics (`sync.Once`), and reset behaviour

## Type of change

- [x] Bug fix

## Test plan

The regression test `TestGetDefaultProvider_RespectsRegisteredFactory` registers a factory pointing to a custom registry with a sentinel server name and asserts that server appears in `ListServers()` — this test fails on the old code and passes with the fix.

No change in open-source builds. In enterprise builds that register a `ProviderFactory`, `thv registry list` and related commands now correctly use the factory-backed config (e.g. a remote registry URL from the enterprise config server) instead of falling back to the embedded registry.

### DIFF
--- a/pkg/api/v1/registry.go
+++ b/pkg/api/v1/registry.go
@@ -14,7 +14,7 @@ import (
 
 	"github.com/go-chi/chi/v5"
 
-	"github.com/stacklok/toolhive-core/registry/types"
+	registry "github.com/stacklok/toolhive-core/registry/types"
 	"github.com/stacklok/toolhive/pkg/config"
 	regpkg "github.com/stacklok/toolhive/pkg/registry"
 	"github.com/stacklok/toolhive/pkg/registry/auth"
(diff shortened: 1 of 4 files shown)


peppescg (dismissed):

JAORMX (dismissed):

reyortiz3:
@claude can you review the comments are addressed?

reyortiz3:
@claude can you review if the comments are finally addressed?

peppescg (approved):

Issue #4742, opened 2026-04-10
Ship opt-in code mode for vMCP


jerm-dro:
## User Story

**As a** platform engineer,
**I want** my agents to be able to execute scripts on tools without shell access
**so that** they can safely reduce context bloat and inference cycles.

## Background

Agents today make sequential tool calls with model inference between each one. For multi-service workflows (e.g., incident triage across PagerDuty, Datadog, Slack, Jira, GitHub, Confluence), this means 10+ round-trips and significant token spend. The Starlark script middleware lets agents submit a single script that calls multiple tools server-side, with loops, conditionals, and `parallel()` fan-out, returning an aggregated result in one tool call.

A working prototype exists in draft PR #4714 (branch `jerm-dro/script-middleware-prototype`). This story hardens that prototype into a shippable, opt-in feature by addressing all known limitations.

### What this story delivers

1. **Config toggle** -- Add an opt-in boolean flag to the VirtualMCPServer spec (or vMCP server config) that enables/disables the script middleware. Defaults to disabled. When disabled, `execute_tool_script` must not appear in `tools/list` responses.

2. **Inner tool call mechanism** -- Replace `httptest.NewRecorder` usage in `innerToolCaller.CallTool` and `fetchToolList` with a proper internal dispatch mechanism. The current approach creates synthetic HTTP round-trips through `httptest.NewRecorder`; the replacement should invoke the middleware chain without constructing real HTTP request/response pairs (e.g., via a direct function call interface or an internal dispatcher).

3. **Step limit configuration** -- Make the per-script Starlark step limit configurable rather than hardcoded at `DefaultStepLimit` (100,000). Expose this via the same config surface as the opt-in toggle. Provide a sensible default.

4. **Concurrency cap for `parallel()`** -- Add a configurable maximum number of goroutines that `parallel()` can spawn concurrently. Currently `parallelBuiltin` launches one goroutine per callable with no bound. The cap should be configurable via the script config and have a sensible default (e.g., 10).

5. **Per-tool-call timeout** -- Add a configurable timeout for individual tool calls made from within a script. If a tool call exceeds the timeout, it should be cancelled and return a clear error to the script. Expose via config with a sensible default.

[...]

PR #4748, opened 2026-04-10
Introduce script engine package for vMCP code mode

jerm-dro:
## Summary

- Agents today batch multi-tool workflows through sequential tool calls with model inference between each one, wasting round-trips and tokens. vMCP code mode (#4742) lets agents submit a Starlark script that calls tools server-side and returns an aggregated result in one call.
- This PR introduces the core `pkg/script/` package — the Starlark execution engine, tool bridge, and builtins — with a clean public interface. The middleware (PR 2), optimizer integration (PR 3), and config/CRD wiring (PR 4) build on this foundation.

Part of #4742

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/script/script.go` | Public API: `Executor` interface, `Tool`, `Config`, `New()` |
| `pkg/script/executor.go` | Unexported executor wiring internals together |
| `pkg/script/description.go` | `GenerateToolDescription()` for `tools/list` injection |
| `pkg/script/internal/core/execute.go` | Starlark thread creation, script wrapping, step limit |
| `pkg/script/internal/conversions/starlark.go` | Bidirectional Go ↔ Starlark type conversion |
| `pkg/script/internal/conversions/result.go` | `ParseToolResult` for `mcp.CallToolResult` → Go values |
| `pkg/script/internal/conversions/toolname.go` | MCP tool name → valid Starlark identifier |
| `pkg/script/internal/builtins/builtins.go` | Single `Build()` entry point returning globals + reserved names |
| `pkg/script/internal/builtins/tools.go` | Tool callable factory with positional + kwargs support |
| `pkg/script/internal/builtins/calltool.go` | `call_tool("name", ...)` generic dispatch builtin |
| `pkg/script/internal/builtins/parallel.go` | `parallel()` concurrent fan-out with semaphore, context cancellation, child step limits |

No. This PR introduces a library package with no wiring to the server yet. Code mode becomes user-facing when the middleware is added in PR 2.

**This is a new, self-contained package with no integration into the existing server.** The XL size label is expected — the diff is ~60% tests and the rest is a new internal package tree. No existing files are modified.

**Review by commit** — each commit introduces one layer, building bottom-up:
1. Public API types (`Executor`, `Tool`, `Config`)
2. `internal/core` — Starlark execution engine
3. `internal/conversions` — type conversion + MCP result parsing
4. `internal/builtins` — `parallel()`, `call_tool()`, tool callable factory
5. Executor wiring + integration tests
6. Test style fix (require + table-driven)
7. Self-review fixes (data race, step limit on child threads, context cancellation, `load()` panic, data shadowing, builtins simplification)
8. Address review comments (remove reserved set, table-driven tests, JSON→structured data test)
9. Address automated review (remove `ToolCallTimeout` from Config — caller responsibility via `Tool.Call` closure, UTF-8 safe truncation, `go mod tidy`)

Key design decisions:
- **`Tool.Call` is a closure** — the middleware (PR 2) constructs these with callbacks that route through the middleware chain
- **Per-tool timeouts are a caller responsibility** — the engine passes `ctx` through to `Tool.Call` without adding deadlines. Callers enforce timeouts by wrapping `ctx` with `context.WithTimeout` in their closure.
- **Returns `*mcp.CallToolResult`** directly — no custom result types, the middleware serializes it as-is
- **Positional args** become `{"arg0": val, "arg1": val, ...}` — same convention for both direct tool calls and `call_tool()`
- **`parallel()` supports optional concurrency cap** via semaphore, propagates step limit to child threads, respects context cancellation
- **Data argument keys validated** against globals to prevent shadowing builtins

## Large PR Justification

- New self-contained package (`pkg/script/`) with no modifications to existing files
- ~60% of the diff is unit tests — splitting tests from implementation would make review harder, not easier
- The internal package tree (`core/`, `builtins/`, `conversions/`) must ship together for the public API to compile


jerm-dro:
<details><summary>Implementation Plan (approved before coding)</summary>

# PR 1: Introduce script engine package

## Context

Story #4742 — Ship opt-in code mode for vMCP. Greenfield implementation (prototype discarded). PR 1 introduces the core `pkg/script/` package with a clean public interface and internal Starlark implementation. Middleware (PR 2), optimizer integration (PR 3), and CRD wiring (PR 4) build on this.

## Public API (`pkg/script/`)

```go
// Executor runs Starlark scripts and describes the virtual tool.
type Executor interface {
    Execute(ctx context.Context, script string, data map[string]interface{}) (*mcp.CallToolResult, error)
    ToolDescription() string
}

// Tool bundles an MCP tool's metadata with a callback for invoking it.
type Tool struct {
    Name        string
    Description string
    Call        func(ctx context.Context, arguments map[string]interface{}) (*mcp.CallToolResult, error)
}

// Config holds script execution parameters.
type Config struct {
    StepLimit          uint64        // Max Starlark execution steps (default: 100,000)
    ParallelMax        int           // Max concurrent goroutines for parallel() (0 = unlimited)
    ToolCallTimeout    time.Duration // Per-tool-call timeout (0 = no timeout)
}

// New creates an Executor bound to the given tools and config.
// A nil config uses defaults.
func New(tools []Tool, cfg *Config) Executor
```

### Tool calling conventions

Both direct tool invocation (`my_tool(...)`) and `call_tool("name", ...)` use identical argument handling:
- `my_tool(key=val)` → `{"key": val}` (kwargs)
- `my_tool(val1, val2)` → `{"arg0": val1, "arg1": val2}` (positional)
- `my_tool(val1, key=val2)` → `{"arg0": val1, "key": val2}` (mixed)

`call_tool` is just an alternative for tools with names that aren't valid Starlark identifiers. Same argument handling, same `Tool.Call` dispatch.

## Internal packages

```
pkg/script/internal/
├── core/           # Starlark thread, script wrapping, execution
├── builtins/       # parallel(), call_tool() Starlark builtins
└── conversions/    # Go ↔ Starlark type conversion, MCP result parsing
```

### `internal/core/`

- `execute.go` — `Execute(script, globals, stepLimit) (*ExecuteResult, error)`. Thread creation, `def __main__()` wrapping, print capture.
- `execute_test.go` — step limit enforcement, basic execution, error propagation

### `internal/builtins/`

- `parallel.go` — `NewParallel(maxConcurrency int) *starlark.Builtin`. WaitGroup + optional semaphore. 0 = unlimited.
- `calltool.go` — `NewCallTool(ctx, toolMap) *starlark.Builtin`. Dispatches by name from first positional arg.
- `tools.go` — `MakeToolCallable(ctx, name, callFn) *starlark.Builtin`. Shared arg handling (positional → `argN`, kwargs, mixed). Used by both direct tool callables and `call_tool`.
- Tests for each

### `internal/conversions/`

- `starlark.go` — `GoToStarlark` / `StarlarkToGo` bidirectional conversion
- `result.go` — `ParseToolResult(*mcp.CallToolResult) (interface{}, error)` using `mcp.AsTextContent()`
- `toolname.go` — `SanitizeName(name) string`
- Tests for each

## How Execute() works

1. Builds `starlark.StringDict` globals from bound tools:
   - Each `Tool` → Starlark callable via `builtins.MakeToolCallable` (shared arg handling)
   - `call_tool()` via `builtins.NewCallTool` (same arg handling, name-based dispatch)
   - `parallel()` via `builtins.NewParallel(cfg.ParallelMax)`
   - `data` entries → top-level Starlark variables
2. Calls `core.Execute(script, globals, cfg.StepLimit)`
3. Converts result value → Go via `conversions.StarlarkToGo()`
4. Serializes to JSON, wraps in `*mcp.CallToolResult` (text content + optional logs item)

## Commits

1. **Add public API types** — `script.go` with `Executor`, `Tool`, `Config`, `New()` stub. `description.go` with `ToolDescription()`.
2. **Add internal/core** — Starlark execution engine with step limit + tests.
3. **Add internal/conversions** — Go ↔ Starlark type conversion, MCP result parsing, tool name sanitization + tests.
4. **Add internal/builtins** — `parallel()`, `call_tool()`, tool callable factory with positional+kwargs support + tests.
5. **Wire executor** — `executor.go` implementation connecting public API → internals. Integration tests in `script_test.go`.

## Tests

| AC | Test |
|----|------|
| unit: script can call multiple tools, loops/conditionals, return aggregated result | `script_test.go` |
| unit: `parallel()` fans out concurrently, returns in order | `builtins/parallel_test.go` |
| unit: step limit exceeded returns clear error | `core/execute_test.go` |
| unit: result unwrapping handles multiple formats | `conversions/result_test.go` |

Additional: positional/kwargs arg handling (`builtins/tools_test.go`), type conversion round-trips (`conversions/starlark_test.go`), name sanitization (`conversions/toolname_test.go`).

## Verification

```bash
task test          # unit tests pass
task lint-fix      # linting passes
task license-fix   # SPDX headers
```

</details>

pkg/script/internal/builtins/builtins.go
```diff
@@ -0,0 +1,67 @@
...
+		sanitized := conversions.SanitizeName(t.Name)
+		if existing, ok := seen[sanitized]; ok {
+			slog.Warn("tool name collision after sanitization",
+				"tool1", existing, "tool2", t.Name, "sanitized", sanitized)
+			continue
+		}
+		seen[sanitized] = t.Name
+
+		globals[sanitized] = makeToolCallable(ctx, sanitized, t.Call)
+		reserved[sanitized] = true
```
jerm-dro: This reserved set is unnecessary. The caller can just block prevent shadowing of anything in globals?

pkg/script/script_test.go
```diff
@@ -0,0 +1,271 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExecutor_MultiToolScript(t *testing.T) {
```
jerm-dro: These tests should be tabular.

pkg/script/script_test.go
```diff
@@ -0,0 +1,271 @@
...
+	require.NoError(t, err)
+	require.False(t, result.IsError)
+
+	text := extractText(t, result)
+	var parsed map[string]interface{}
+	require.NoError(t, json.Unmarshal([]byte(text), &parsed))
+	require.Contains(t, parsed, "user")
+	require.Contains(t, parsed, "posts")
+}
+
```
jerm-dro, line 40: There should also be a test here demonstrating the automatic conversion of JSON strings into structured data.

pkg/script/internal/builtins/parallel.go
```diff
@@ -0,0 +1,143 @@
...
+		for i, err := range state.errs {
+			if err != nil {
+				return nil, fmt.Errorf("parallel: task %d failed: %w", i, err)
+			}
+		}
+
+		// Merge child logs into parent thread
+		for _, logs := range state.childLogs {
+			for _, msg := range logs {
+				thread.Print(thread, msg)
```
JAORMX: **suggestion:** So, all N goroutines get launched immediately here, and the semaphore only gates *execution*, not goroutine creation. With something like `parallel([...1000 tasks...])` and `ParallelMax=2`, you'd end up with 1000 goroutines, 998 of which are just sitting there blocked on the semaphore.

Not a blocker for PR 1 since there's no server wiring yet, but worth considering a worker-pool pattern for PR 2 where the goroutine count matches the semaphore size and tasks are pulled from a channel. That way you're not paying for goroutines that are just waiting around.

pkg/script/internal/builtins/parallel.go
```diff
@@ -0,0 +1,143 @@
...
+		s.errs[idx] = fmt.Errorf("parallel: element %d is not callable (got %s)",
+			idx, s.fns.Index(idx).Type())
+		return
+	}
+
+	var logs []string
+	childThread := &starlark.Thread{
+		Name: fmt.Sprintf("%s/parallel-%d", parent.Name, idx),
+		Print: func(_ *starlark.Thread, msg string) {
+			logs = append(logs, msg)
```
JAORMX: **suggestion:** Each child thread gets the full `stepLimit`, so `parallel([f1, f2, f3])` effectively triples the step budget. And since `parallel()` is itself callable from within a parallel task, you can get exponential amplification with nesting: `parallel([lambda: parallel([lambda: parallel(...)])])`.

For untrusted scripts this undermines the sandboxing guarantee. A couple of options:
- Divide the remaining budget across children (`stepLimit / n` per child)
- Track a shared atomic step counter
- Add a nesting depth limit

Again, not blocking PR 1... but this should be addressed before code mode is exposed to external input.

pkg/script/internal/core/execute.go
```diff
@@ -0,0 +1,86 @@
...
+	var b strings.Builder
+	b.WriteString("def __main__():\n")
+	for _, line := range strings.Split(script, "\n") {
+		b.WriteString("    ")
+		b.WriteString(line)
+		b.WriteString("\n")
+	}
+	b.WriteString("__result__ = __main__()\n")
+	return b.String()
+}
```
JAORMX, line 90: **suggestion:** The wrapping approach indents every line by 4 spaces, which silently changes the content of triple-quoted strings:

```python
# Input script:
msg = """
first line
second line
"""
# After wrapping becomes:
def __main__():
    msg = """
    first line    # <-- extra spaces baked into the string
    second line
    """
```

If this is an accepted limitation (probably fine for tool orchestration scripts), a quick comment here would save future readers the head-scratching. If multi-line strings should work, an alternative would be using `exec()` on the raw script with a result capture variable instead of indenting.

pkg/script/internal/builtins/builtins_test.go
```diff
@@ -0,0 +1,258 @@
...
+			return mcp.NewToolResultText("ok"), nil
+		}},
+	}
+
+	globals := Build(context.Background(), tools, 100_000, 0)
+
+	require.Contains(t, globals, "my_tool", "sanitized tool name should be in globals")
+	require.Contains(t, globals, "call_tool", "call_tool should be in globals")
+	require.Contains(t, globals, "parallel", "parallel should be in globals")
+}
```
JAORMX, line 234: **nitpick:** The assertion here checks `require.NotNil(t, result)` for the `call_tool("my.tool")` path, but that's a pretty weak check. It'd be more convincing to assert on the actual content (e.g., that the result is `"second"` and not `"first"`) to prove the dispatch actually went to the right tool.

pkg/script/script.go
```diff
@@ -0,0 +1,90 @@
...
+func resolveConfig(cfg *Config) Config {
+	if cfg == nil {
+		return Config{StepLimit: DefaultStepLimit}
+	}
+	c := *cfg
+	if c.StepLimit == 0 {
+		c.StepLimit = DefaultStepLimit
+	}
+	return c
+}
```
JAORMX, line 93: **nitpick:** A negative `ParallelMax` silently becomes "unlimited" here since the check downstream is `> 0`. Probably fine in practice, but a quick validation (or just using `uint`) would make the contract clearer.

JAORMX (commented):
Clean package, well-structured. The internal/ layering is solid, the closure-based tool bridge is a nice design that keeps the engine decoupled from the middleware chain, and the test coverage is thorough (~55% of the diff is tests).

Main thing to track: the step budget amplification via nested `parallel()`. Not a blocker for this PR since there's no server wiring, but it should be resolved before code mode goes live.

pkg/script/internal/builtins/parallel.go
jerm-dro: Good call. Introduced `pkg/foreach` with a generic `Concurrent[T]` worker pool — goroutine count now matches `maxWorkers`, not the task count. `parallel()` delegates to it. The package is reusable elsewhere.

pkg/script/internal/builtins/parallel.go
jerm-dro: Fixed — step budget is now divided evenly across children (`stepLimit / N` per child). Nesting no longer amplifies the budget.

pkg/script/internal/core/execute.go
jerm-dro: Added a comment documenting this as an accepted limitation. Triple-quoted strings are uncommon in tool orchestration scripts, and the alternative (parsing + AST rewrite) is disproportionate complexity for this use case.

pkg/script/internal/builtins/builtins_test.go
jerm-dro: Fixed — now asserts `result.Value == "second"` to prove dispatch reaches `my.tool` and not `my-tool`.

pkg/script/script.go
jerm-dro: Fixed — `resolveConfig` now clamps negative values to 0 (unlimited).

JAORMX (dismissed):

JAORMX (approved):

PR #4751, opened 2026-04-11
Fix stale config in registry routes and WorkloadService

reyortiz3:
## Summary

Two independent config caches caused inconsistent state after a registry configuration change.

- **Bug 1 (registry.go)**: `NewRegistryRoutes` and `NewRegistryRoutesForServe` created independent `config.Provider` instances for `configProvider` and `configService`. The `getRegistry`/`listRegistries` responses mixed live metadata from `configService.GetRegistryInfo()` (type/source) with the server list from `getCurrentProvider()` — two separate providers that could reflect different config states. Fix: allocate one `p := config.NewProvider()` and pass it to both, so the response is always internally consistent.

- **Bug 2 (workload_service.go)**: `WorkloadService.appConfig` was a one-time startup snapshot (`*config.Config`). After a registry configuration change, workloads created in the same process lifetime got stale registry URLs baked into their persisted `RunConfig`. Fix: store a `config.Provider` and call `GetConfig()` at the point of use inside `BuildFullRunConfig`.

## Type of change

- [x] Bug fix

## Test plan

New tests added:
- `TestNewRegistryRoutes_ConfigServiceAndProviderAreConsistent`: calls `listRegistries` with a factory-backed local registry and asserts that both `Type` (from `configService`) and `ServerCount > 0` (from `getCurrentProvider`) reflect the same config source.
- `TestNewRegistryRoutesForServe_ConfigServiceAndProviderAreConsistent`: same for serve mode.
- `TestNewWorkloadService` extended: asserts `configProvider != nil` so the live-read contract is enforced.

## Changes

| File | Change |
|------|--------|
| `pkg/api/v1/registry.go` | Share single `config.Provider` between `configProvider` and `configService` in both constructors |
| `pkg/api/v1/workload_service.go` | Replace `appConfig *config.Config` snapshot with `configProvider config.Provider`; call `GetConfig()` at point of use |
| `pkg/api/v1/registry_factory_test.go` | Add consistency tests verifying shared provider across `configService` and `getCurrentProvider` |
| `pkg/api/v1/workload_service_test.go` | Update field name; assert `configProvider != nil` in `TestNewWorkloadService` |
| `pkg/api/v1/workloads_test.go` | Update field name from `appConfig` to `configProvider` |

No. The fixes are internal — they eliminate stale state that caused the `/registry` response to be internally inconsistent and caused newly-created workloads to store outdated registry URLs in their persisted `RunConfig`.

The pre-existing failure in `TestUpdateWorkload_PortReuse/Edit_with_explicit_port_should_use_that_port` is unrelated to this PR and reproduces on `main` without these changes.

### DIFF
--- a/pkg/api/v1/registry.go
+++ b/pkg/api/v1/registry.go
@@ -285,9 +285,10 @@ type RegistryRoutes struct {
 
 // NewRegistryRoutes creates a new RegistryRoutes with the default config provider
 func NewRegistryRoutes() *RegistryRoutes {
+	p := config.NewProvider()
 	return &RegistryRoutes{
-		configProvider: config.NewProvider(),
-		configService:  regpkg.NewConfigurator(),
+		configProvider: p,
+		configService:  regpkg.NewConfiguratorWithProvider(p),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


JAORMX (dismissed):

reyortiz3:
@claude can you check if the comments were addressed?

JAORMX (approved):

Issue #3157, opened 2025-12-24
CRDs need updates to accommodate authorizer-plugin support


ghaskins:
## Problem Statement

Pull request #3110 adds support for pluggable authorizers.  One remaining gap is the presence of Cedar-isms in the CRDs.  For example:

https://github.com/stacklok/toolhive/blob/c973cdc2a88dc7b2c0c3e33f65bb1eade39345e5/deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpservers.yaml#L93

There are a few ways to handle this, but they all likely require a new version of the affected CRDs:

## Possible Solutions

1.  New CRD schema drops "inline" authz support, relying purely on a configmap-reference.  Validation would remain at the code level but would be removed from the schema level.
2.  New CRD schema modifies "inline" authz support to relax schema checking (see https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#controlling-pruning) to accommodate various types of authz plugins.  Validation would remain at the code level, but is removed at the schema level
3.  We drop "configmap-reference" support and create CRD(s) for Authz configuration.  The inline configuration would be generalized to accept an inline representation of the Authz configuration using [RawExtensions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#rawextension). Likewise, the previous use cases for configmap-based Authz configuration would now use the new Authz CRD.  Validation would remain at the code and inline schema levels, and is now also introduced at the "external reference" level, since we've replaced ConfigMaps with a CRD.

ghaskins:
| Option | Inline Validation | External Reference Validation | Plugin Support |
|--------|-----------------|------------------------------|----------------|
| Current CRDs | Yes | No | No |
| Proposal 1 | n/a | No | Yes |
| Proposal 2 | No | No | Yes |
| Proposal 3 (*) | Yes | Yes | Yes |

(*): While Proposal 3 looks appealing feature-wise, it should be pointed out that it is also redundant with the concept of a ConfigMap and goes against the grain a bit, so this fact should be considered as well.

ChrisJBurns:
Apologies for the long silence here, @ghaskins. This deserved an answer at the time and did not get one.

[...]

PR #4777, opened 2026-04-13
Add MCPAuthzConfig CRD for backend-agnostic authorization

JAORMX:
## Summary

- The current inline authz configuration (`authzConfig`) is Cedar-specific and doesn't support alternative authorization backends like HTTP PDP. This adds a new `MCPAuthzConfig` CRD that decouples authorization configuration from the workload CRDs, enabling any registered authorizer backend to be used.
- Adds `ConfigKey() string` to the `AuthorizerFactory` interface so the MCPAuthzConfig controller can reconstruct full config JSON without hardcoding backend-specific keys. **This is a breaking change for out-of-tree implementers** — see the "Does this introduce a user-facing change?" section below.
- Creates the `MCPAuthzConfig` CRD with `type` + `config` (RawExtension) spec, validation via factory registry, hash-based change detection, referencing workload tracking, and deletion protection.
- Adds `AuthzConfigRef` field to MCPServer, MCPRemoteProxy, and VirtualMCPServer specs for referencing shared MCPAuthzConfig resources.

Relates to #3157

## Type of change

- [x] New feature

## Test plan

## Changes

> Note: this PR was re-implemented against the **`v1beta1`** API after `main` graduated v1beta1 to the storage version. The new CRD types live in `cmd/thv-operator/api/v1beta1/`, with a deprecated compat shell in `v1alpha1`, mirroring the established `MCPOIDCConfig` pattern. Generated CRD YAML, deepcopy, helm templates, RBAC, and CRD reference docs are regenerated accordingly.

| Area | Change |
|------|--------|
| `pkg/authz/authorizers/registry.go` | Add `ConfigKey()` to `AuthorizerFactory` interface; reject reserved key values (`""`, `"version"`, `"type"`) at `Register()` time |
| `pkg/authz/authorizers/cedar/core.go` | Implement `ConfigKey()` returning `"cedar"` |
| `pkg/authz/authorizers/http/core.go` | Implement `ConfigKey()` returning `"pdp"` |
| `cmd/thv-operator/api/v1beta1/mcpauthzconfig_types.go` | New CRD types + `Validate()` (consults `authorizers.IsRegistered` for defense-in-depth): `MCPAuthzConfig`, `MCPAuthzConfigSpec`, `MCPAuthzConfigStatus`, `MCPAuthzConfigReference` |
| `cmd/thv-operator/api/v1beta1/{mcpserver,mcpremoteproxy,virtualmcpserver}_types.go` | Add `AuthzConfigRef` + CEL mutual-exclusion XValidation |
| `cmd/thv-operator/api/v1alpha1/*` | Deprecated compat shims embedding v1beta1 |
| `cmd/thv-operator/controllers/mcpauthzconfig_controller.go` | MCPAuthzConfig reconciler: validation, canonical-JSON-hashed change detection, finalizer (via `MutateAndPatchSpec`), watches; success path collapses hash/refs/conditions into a single `MutateAndPatchStatus` patch |
| `cmd/thv-operator/app/app.go` | Register MCPAuthzConfig controller; backend blank-imports |
| `deploy/charts/operator*` | Regenerated CRD YAML, RBAC clusterrole, helm templates, CRD docs |

- This is part 1 of 2: CRD + controller only. Part 2 (#4778, wiring references into workload controllers) follows on top. The `AuthzConfigRef` field godoc on each workload type carries a `TODO(#4778)` marker that controller-gen strips from the rendered CRD description so the user-facing schema stays clean.
- `buildFullAuthzConfigJSON()` reconstructs the full config envelope (`{"version","type","<configKey>": ...}`) using the factory's `ConfigKey()` — this avoids hardcoding Cedar-specific structure. It also returns the resolved factory alongside the JSON so callers don't double-look-up the registry. Reserved envelope keys (`""`, `"version"`, `"type"`) are rejected at `Register()` time to prevent a misconfigured factory from shadowing envelope metadata.
- The controller watches MCPServer, VirtualMCPServer, and MCPRemoteProxy to maintain accurate `ReferencingWorkloads` status and enforce deletion protection. RBAC markers for the watched resources are present and rendered into the generated clusterrole. `ReferencingWorkloads` uses a compound `+listMapKey=kind,name` so cross-kind name reuse is distinct.
- CEL `XValidation` enforces `authzConfig`/`authzConfigRef` mutual exclusivity on all three workload specs, mirroring the `oidcConfig`/`oidcConfigRef` rules, and renders into both served CRD versions. New envtest specs in `cmd/thv-operator/test-integration/{mcp-server,mcp-remote-proxy,virtualmcp}/` exercise these admission rules.
- Status and finalizer writes go through `controllerutil.MutateAndPatchStatus` / `MutateAndPatchSpec` per `.claude/rules/operator.md`. A regression test (`TestMCPAuthzConfigReconciler_PreservesForeignConditions`) seeds a foreign condition the controller doesn't own and asserts it survives across reconciles.
- Config hashing canonicalises `spec.config` (re-marshals through `json.Unmarshal → json.Marshal` to sort keys / drop whitespace) before passing through `ctrlutil.CalculateConfigHash`, so noop kubectl-apply round-trips that re-serialise the same logical config don't flip the hash.

**Yes — a breaking change to an exported Go interface.** This affects downstream Go consumers that maintain their own `AuthorizerFactory` implementation; it does **not** affect Kubernetes users of the operator or anyone who consumes CRDs / the CLI.

- **Interface change**: `pkg/authz/authorizers.AuthorizerFactory` gains a new required method, `ConfigKey() string`. Any out-of-tree implementer (private fork, enterprise/closed-source backend) won't compile against this version until it adds the method. All in-tree implementers — `cedar`, `http`, and the test mocks — are migrated in this PR.
- **Why compile-time, not runtime**: returning the key at compile time is the structurally-correct shape per `.claude/rules/go-style.md` ("Option pattern must be compile-time safe"). The alternative — a type-assert on an optional capability interface — would silently no-op for missing implementations and is explicitly flagged in the rules.
- **Reserved values**: `Register()` panics at startup if a factory's `ConfigKey()` returns one of the reserved envelope keys (`""`, `"version"`, `"type"`). Out-of-tree backends must avoid these.
- **Known consumers**: a sweep of internal Go consumers of `pkg/authz/authorizers` confirms none implement `AuthorizerFactory` directly — they consume the `Authorizer` runtime interface and the `cedar.NewCedarAuthorizer` constructor, so the new method is transparent to them. Any other closed-source consumer maintaining its own factory should add the method when bumping the `toolhive` Go dependency.
- **Wire format unchanged**: the JSON envelope shape `{"version", "type", "<configKey>": ...}` is unchanged. Code that hand-emits this format (rather than implementing `AuthorizerFactory`) is unaffected.

## Large PR Justification

This PR is a single cohesive, atomic addition: a new `MCPAuthzConfig` CRD plus its controller. The pieces cannot be split without leaving the tree in a non-functional intermediate state:

- The API types, deepcopy, CRD YAML, helm templates, RBAC, and CRD reference docs are all machine-generated from the same type definitions — splitting them produces failing `verify-gen` runs.
- The controller, its `SetupWithManager` registration in `app.go`, and the `ConfigKey()` interface method are mutually dependent and won't compile apart.
- Of the additions, the large majority is generated YAML/docs/deepcopy and tests. Hand-written production Go is a small fraction.

PR #4778 is stacked on top of this and wires these refs into the workload controllers, so the CRD + controller must land first as a unit.


cmd/thv-operator/api/v1alpha1/mcpserver_types.go
```diff
@@ -290,10 +290,18 @@ type MCPServerSpec struct {
...
+	// Deprecated: Use AuthzConfigRef to reference a shared MCPAuthzConfig resource instead.
+	// AuthzConfig and AuthzConfigRef are mutually exclusive.
 	// +optional
 	AuthzConfig *AuthzConfigRef `json:"authzConfig,omitempty"`
 
+	// AuthzConfigRef references a shared MCPAuthzConfig resource for authorization.
+	// The referenced MCPAuthzConfig must exist in the same namespace as this MCPServer.
+	// Mutually exclusive with authzConfig.
+	// +optional
+	AuthzConfigRef *MCPAuthzConfigReference `json:"authzConfigRef,omitempty"`
```
ChrisJBurns, line 290: **[HIGH] Missing CEL XValidation for authzConfig/authzConfigRef mutual exclusivity** (Consensus: 10/10)

The `AuthzConfig` and `AuthzConfigRef` fields are documented as "mutually exclusive" but no CEL validation rule enforces this at the API server level. The existing OIDC pattern has:

```go
// +kubebuilder:validation:XValidation:rule="!(has(self.oidcConfig) && has(self.oidcConfigRef))",message="oidcConfig and oidcConfigRef are mutually exclusive..."
```

The same rule is needed here on `MCPServerSpec`, `MCPRemoteProxySpec`, and `IncomingAuthConfig`:

```go
// +kubebuilder:validation:XValidation:rule="!(has(self.authzConfig) && has(self.authzConfigRef))",message="authzConfig and authzConfigRef are mutually exclusive; use authzConfigRef to reference a shared MCPAuthzConfig"
```

Without this, users can set both fields simultaneously and the controller must decide which takes precedence.

*Raised by: kubernetes-expert, code-reviewer, toolhive-expert*

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,432 @@
...
+// via the authorizer factory registry, config hash computation, finalizer management,
+// reference tracking, and deletion protection when workloads reference this config.
+type MCPAuthzConfigReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpauthzconfigs,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpauthzconfigs/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpauthzconfigs/finalizers,verbs=update
```
ChrisJBurns, line 52: **[MEDIUM] Missing RBAC markers for watched workload resources** (Consensus: 10/10)

The controller watches MCPServer, VirtualMCPServer, and MCPRemoteProxy (via `SetupWithManager`) and lists them (in `findReferencingWorkloads`), but no `+kubebuilder:rbac` markers declare these permissions. The MCPOIDCConfig controller declares them:

```suggestion
// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpauthzconfigs/finalizers,verbs=update
// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpservers,verbs=get;list;watch
// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=virtualmcpservers,verbs=get;list;watch
// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpremoteproxies,verbs=get;list;watch
```

While these permissions are currently aggregated from other controllers, they should be self-declared for completeness and to prevent breakage if another controller is removed.

*Raised by: kubernetes-expert, code-reviewer, toolhive-expert*

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,432 @@
...
+		return nil, fmt.Errorf("failed to marshal full authz config: %w", err)
+	}
+	return result, nil
+}
+
+// mustMarshalJSON marshals a value to JSON or panics. Only for simple string values.
+func mustMarshalJSON(v string) json.RawMessage {
+	b, err := json.Marshal(v)
+	if err != nil {
+		panic(fmt.Sprintf("failed to marshal %q: %v", v, err))
```
ChrisJBurns: **[MEDIUM] mustMarshalJSON uses panic in controller reconciliation path** (Consensus: 10/10)

While `json.Marshal` on a string cannot realistically fail, panicking in a controller crashes the entire operator. The Go style rules reserve panics for top-level boundaries only. Since `BuildFullAuthzConfigJSON` already returns `([]byte, error)`, this should propagate errors instead.

Consider replacing this function with one that returns `(json.RawMessage, error)` and updating `BuildFullAuthzConfigJSON` to handle the error.

*Raised by: kubernetes-expert, go-security-reviewer, code-reviewer, toolhive-expert*

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,432 @@
...
+		return nil, fmt.Errorf("unsupported authorizer type: %s (registered types: %v)",
+			spec.Type, authorizers.RegisteredTypes())
+	}
+
+	configKey := factory.ConfigKey()
+
+	fullConfig := map[string]json.RawMessage{
+		"version": mustMarshalJSON(authzConfigVersion),
+		"type":    mustMarshalJSON(spec.Type),
+		configKey: spec.Config.Raw,
```
ChrisJBurns, line 212: **[MEDIUM] No nil-check on spec.Config.Raw before use** (Consensus: 8/10)

If `spec.Config.Raw` is nil (bypassed validation, manual testing, or future code changes), this produces `"cedar": null` in the JSON output, leading to confusing downstream errors. Add a defensive check before use:

```go
if len(spec.Config.Raw) == 0 {
    return nil, fmt.Errorf("config field is empty")
}
```

The `config` field is required in the CRD schema, but defense-in-depth is warranted for `RawExtension` fields.

*Raised by: code-reviewer, kubernetes-expert*

cmd/thv-operator/controllers/mcpauthzconfig_controller_test.go
```diff
@@ -0,0 +1,616 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
```
ChrisJBurns, line 1: **[MEDIUM] Test coverage gaps in watch handlers and deletion paths** (Consensus: 7/10)

The following methods have zero test coverage (Codecov reports 43.5% patch / 122 uncovered lines):
- `findReferencingWorkloads` (covers MCPServer, VirtualMCPServer, MCPRemoteProxy listing)
- `findStaleRefs` (stale reference cleanup)
- All three `map*ToAuthzConfig` watch handlers
- `SetupWithManager`

Additionally, the `handleDeletion` test only covers the MCPServer reference path -- VirtualMCPServer and MCPRemoteProxy deletion-blocking cases are missing.

At minimum, consider adding tests for `findReferencingWorkloads` covering all three workload types, and extending the deletion test with VirtualMCPServer and MCPRemoteProxy cases.

*Raised by: code-reviewer*

cmd/thv-operator/api/v1alpha1/mcpauthzconfig_types.go
```diff
@@ -0,0 +1,105 @@
...
+	ConfigHash string `json:"configHash,omitempty"`
+
+	// ReferencingWorkloads is a list of workload resources that reference this MCPAuthzConfig.
+	// Each entry identifies the workload by kind and name.
+	// +listType=map
+	// +listMapKey=name
+	// +optional
+	ReferencingWorkloads []WorkloadReference `json:"referencingWorkloads,omitempty"`
+}
+
```
ChrisJBurns, line 66: **[LOW] ReferencingWorkloads listMapKey=name may be insufficient** (Consensus: 7/10)

Using only `name` as the list map key could collide if an MCPServer and a VirtualMCPServer share the same name in the same namespace. The composite key should be `kind` + `name`:

```go
// +listMapKey=kind
// +listMapKey=name
```

This is a pre-existing pattern copied from MCPOIDCConfig, MCPTelemetryConfig, etc. -- consider fixing across all config CRDs in a separate PR.

*Raised by: kubernetes-expert*

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,432 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	mcpv1alpha1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1alpha1"
+	ctrlutil "github.com/stacklok/toolhive/cmd/thv-operator/pkg/controllerutil"
+	"github.com/stacklok/toolhive/pkg/authz/authorizers"
+	// Import authorizer backends so they register with the factory registry.
+	_ "github.com/stacklok/toolhive/pkg/authz/authorizers/cedar"
```
ChrisJBurns: **[LOW] Controller directly imports authorizer backend packages** (Consensus: 7/10)

The blank imports of `cedar` and `http` authorizer packages create a direct compile-time dependency from the controller to specific backends. No other controller in this package has this pattern. Consider moving these blank imports to `cmd/thv-operator/main.go` (where scheme registrations already happen), keeping the controller package backend-agnostic.

*Raised by: toolhive-expert, kubernetes-expert*

cmd/thv-operator/api/v1alpha1/mcpauthzconfig_types.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+// MCPAuthzConfigSpec defines the desired state of MCPAuthzConfig.
+// MCPAuthzConfig resources are namespace-scoped and can only be referenced by
+// MCPServer, MCPRemoteProxy, or VirtualMCPServer resources in the same namespace.
+type MCPAuthzConfigSpec struct {
+	// Type identifies the authorizer backend (e.g., "cedarv1", "httpv1").
+	// Must match a registered authorizer type in the factory registry.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
ChrisJBurns, line 30: **[LOW] No Validate() method on MCPAuthzConfig type** (Consensus: 7/10)

Other shared config CRDs (MCPOIDCConfig, MCPTelemetryConfig, MCPExternalAuthConfig) define a `Validate()` method on the type itself, called from the controller as `config.Validate()`. This CRD uses a standalone `validateAuthzConfigSpec()` function in the controller package instead.

The deviation is understandable (validation requires the factory registry, an external dependency), but a comment in the controller explaining why would help maintainability.

*Raised by: toolhive-expert, code-reviewer*

ChrisJBurns (changes requested):
**Agents consulted**: kubernetes-expert, go-security-reviewer, code-reviewer, toolhive-expert

### Consensus Summary
| 1 | Missing CEL XValidation for authzConfig/authzConfigRef mutual exclusivity | 10/10 | HIGH | Fix |
| 2 | Missing RBAC markers for watched workload resources | 10/10 | MEDIUM | Fix |
| 3 | mustMarshalJSON uses panic in controller | 10/10 | MEDIUM | Fix |
| 4 | No nil-check on spec.Config.Raw before use | 8/10 | MEDIUM | Fix |
| 5 | Test coverage gaps in watch handlers and deletion paths | 7/10 | MEDIUM | Fix |
| 6 | No Validate() method on MCPAuthzConfig type (pattern deviation) | 7/10 | LOW | Discuss |
| 7 | ReferencingWorkloads listMapKey=name insufficient for uniqueness | 7/10 | LOW | Discuss (pre-existing) |
| 8 | Controller blank-imports authorizer backend packages | 7/10 | LOW | Discuss |

### Overall
This is a well-structured PR that closely follows the established MCPOIDCConfig controller pattern. The CRD design is clean, the `ConfigKey()` interface addition is minimal, and the `findStaleRefs` helper is a nice DRY improvement over the inline stale-ref scanning in existing controllers. The part 1/part 2 split is a sensible approach for incremental delivery.

The main blocker is the missing CEL XValidation rules (Finding \#1). The existing OIDC pattern enforces mutual exclusivity between `oidcConfig` and `oidcConfigRef` via CEL rules, but no equivalent rules exist for `authzConfig`/`authzConfigRef`. Without these, the API server silently accepts resources with both fields set, contradicting the documented contract. The remaining findings are medium-severity improvements: RBAC marker completeness, replacing panic with error return, and adding a nil-check on `Config.Raw`.

---

JAORMX:
Thanks for the thorough multi-agent review. Addressed all findings. Note: since this was last touched, `main` graduated the operator API to **`v1beta1` as the storage version**, so the PR has been re-implemented against `v1beta1` (with a deprecated `v1alpha1` compat shell), mirroring the established `MCPOIDCConfig` pattern. Per-finding:

1. **HIGH — CEL XValidation mutual exclusivity:** Done. `!(has(self.authzConfig) && has(self.authzConfigRef))` is applied to `MCPServerSpec`, `MCPRemoteProxySpec`, and `IncomingAuthConfig` (vMCP), matching the `oidcConfig`/`oidcConfigRef` pattern, and renders into both served CRD versions.
2. **MEDIUM — RBAC markers:** Added; the regenerated clusterrole grants get/list/watch on the watched workload types plus the `mcpauthzconfigs` status/finalizers subresources. `mcpservers` RBAC is inherited from the MCPServer controller, consistent with MCPOIDCConfig.
3. **MEDIUM — panic in `mustMarshalJSON`:** Replaced with `marshalJSONString` returning an error, propagated through `BuildFullAuthzConfigJSON`.
4. **MEDIUM — nil-check on `spec.Config.Raw`:** Guarded (`len(spec.Config.Raw) == 0` → explicit error), plus a `Validate()` method on the type as defense-in-depth.
5. **MEDIUM — test coverage:** Added watch-handler tests (current + stale-ref enqueue for all three workload kinds, plus wrong-type and no-ref cases) and deletion-path tests (finalizer add, deletion-without-finalizer, reference-blocking per workload type).
6. **LOW — `Validate()` method:** Added, mirroring MCPOIDCConfig. Structural checks only; backend-specific validation stays in the controller because it needs the authorizer registry the API package must not import (documented inline).
7. **LOW — `ReferencingWorkloads` listMapKey:** Aligned to `+listMapKey=name`, matching the shared `WorkloadReference` convention used by every other config CRD. The "name not unique across kinds" concern is pre-existing and repo-wide — suggest addressing it uniformly in a separate change.
8. **LOW — controller blank-imports backends:** Kept out of the controller; they live in the operator entrypoint (`app.go`) so the controller stays backend-agnostic, with a comment explaining why.

Relates to #3157 (still open; this implements its "Solution option 3"). Part 2 (#4778) is stacked on top and will be rebased once this lands.

cmd/thv-operator/api/v1alpha1/types.go
```diff
@@ -126,6 +126,35 @@ type MCPOIDCConfigList struct {
 	Items           []MCPOIDCConfig `json:"items"`
 }
 
+// ─── MCPAuthzConfig ──────────────────────────────────────────────────────────
+
+//+kubebuilder:object:root=true
+//+kubebuilder:deprecatedversion:warning="toolhive.stacklok.dev/v1alpha1 is deprecated; use v1beta1"
```
ChrisJBurns, line 132: We don't need this? As `MCPAuthzConfig` is a new CRD?

cmd/thv-operator/api/v1alpha1/types.go
```diff
@@ -126,6 +126,35 @@ type MCPOIDCConfigList struct {
...
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:shortName=authzcfg,categories=toolhive
+//+kubebuilder:printcolumn:name="Type",type=string,JSONPath=`.spec.type`
+//+kubebuilder:printcolumn:name="Valid",type=string,JSONPath=`.status.conditions[?(@.type=='Valid')].status`
+//+kubebuilder:printcolumn:name="References",type=integer,JSONPath=`.status.referenceCount`
+//+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// MCPAuthzConfig is the deprecated v1alpha1 version of the MCPAuthzConfig resource.
```
ChrisJBurns, line 140: This isn't true?

cmd/thv-operator/api/v1beta1/mcpserver_types.go
```diff
@@ -293,10 +294,18 @@ type MCPServerSpec struct {
...
+	// Deprecated: Use AuthzConfigRef to reference a shared MCPAuthzConfig resource instead.
+	// AuthzConfig and AuthzConfigRef are mutually exclusive.
 	// +optional
 	AuthzConfig *AuthzConfigRef `json:"authzConfig,omitempty"`
 
+	// AuthzConfigRef references a shared MCPAuthzConfig resource for authorization.
+	// The referenced MCPAuthzConfig must exist in the same namespace as this MCPServer.
+	// Mutually exclusive with authzConfig.
+	// +optional
+	AuthzConfigRef *MCPAuthzConfigReference `json:"authzConfigRef,omitempty"`
```
ChrisJBurns, line 320: **F1 — MEDIUM:** `AuthzConfigRef` is added here (and to MCPRemoteProxy/VirtualMCPServer) but no workload controller resolves it into a runtime authz config in this PR. Grepping the repo shows `Spec.AuthzConfigRef` only read by the new MCPAuthzConfigReconciler for bookkeeping (finalizer / ReferenceCount). A user who follows the `Deprecated: Use AuthzConfigRef ...` guidance on the sibling inline field will get a successfully reconciled MCPServer with **no authorization enforced**. The PR description discloses this is part 1/2 with #4778 stacked, but the field-level docs and the `Deprecated:` annotation get ahead of the actual capability.

Consider one of:
1. Hold the `Deprecated:` annotation on the inline `AuthzConfig` field (and the equivalent on MCPRemoteProxy/VirtualMCPServer) until #4778 lands.
2. Add a controller-side validation that emits a Warning event when `AuthzConfigRef` is set on a workload until the consumer ships, so users aren't silently mis-configured.
3. Note the staging gap in the field's godoc (e.g. *"Consumed by workload controllers in #4778; until then this field is reference-tracking only and does not apply authorization."*).

cmd/thv-operator/api/v1beta1/mcpserver_types.go
```diff
@@ -200,6 +200,7 @@ const SessionStorageProviderRedis = "redis"
 
 // MCPServerSpec defines the desired state of MCPServer
 //
+// +kubebuilder:validation:XValidation:rule="!(has(self.authzConfig) && has(self.authzConfigRef))",message="authzConfig and authzConfigRef are mutually exclusive; use authzConfigRef to reference a shared MCPAuthzConfig"
```
ChrisJBurns, line 203: **F7 — MEDIUM:** This XValidation rule (and the parallel rules on `MCPRemoteProxySpec` and `IncomingAuthConfig`) is admission-time only, so unit tests with the fake client can't exercise it. There are no envtest cases that apply each CR with (a) only `authzConfig`, (b) only `authzConfigRef`, (c) both set — only (c) should be rejected.

Please add envtest coverage for all three new mutual-exclusion rules. This is the exact scenario the `.claude/rules/operator.md` testing guidance points at envtest for (over chainsaw).

cmd/thv-operator/api/v1beta1/mcpserver_types.go
```diff
@@ -293,10 +294,18 @@ type MCPServerSpec struct {
 	// +optional
 	OIDCConfigRef *MCPOIDCConfigReference `json:"oidcConfigRef,omitempty"`
 
-	// AuthzConfig defines authorization policy configuration for the MCP server
+	// AuthzConfig defines authorization policy configuration for the MCP server.
+	// Deprecated: Use AuthzConfigRef to reference a shared MCPAuthzConfig resource instead.
```
ChrisJBurns: **F16 — LOW:** Marking the inline field `Deprecated:` is correct in intent, but there is no migration guide for users today: no conversion table (`inline.policies` → `MCPAuthzConfig.spec.config.policies`), no operator-side auto-conversion, and no docs link from this comment. Combined with F1 (no workload controller consumes `AuthzConfigRef` yet), a user attempting migration in the wrong order ends up with no enforcement.

Either defer the `Deprecated:` annotation until #4778 lands, or add a short pointer to a migration doc (e.g. `docs/operator/migration-authzconfig.md`).

cmd/thv-operator/api/v1beta1/mcpauthzconfig_types.go
```diff
@@ -0,0 +1,134 @@
...
+// This method is called by the controller during reconciliation.
+//
+// Note: This provides defense-in-depth alongside CEL validation rules. CEL catches
+// issues at API admission time, but this method also validates stored objects to
+// catch any that bypassed CEL or were stored before CEL rules were added.
+//
+// Backend-specific validation (delegating to the authorizer factory registry) lives
+// in the controller rather than here, because the API types package must not import
+// the authorizer backends.
+func (r *MCPAuthzConfig) Validate() error {
```
ChrisJBurns, line 129: **F13 — LOW:** `Validate()` only checks `Type != ""` and `len(Config.Raw) != 0`. Whether the type is actually registered is checked only inside the controller via `authorizers.GetFactory(cfg.Type)`. Other callers of `Validate()` (a future webhook, CLI tooling, an admission-style helper) would happily accept an unknown type as "valid".

Consider consulting the authorizers registry from `Validate()` directly so all callers fail-closed:

```suggestion
func (r *MCPAuthzConfig) Validate() error {
	if r.Spec.Type == "" {
		return fmt.Errorf("spec.type is required")
	}
	if !authorizers.IsRegistered(r.Spec.Type) {
		return fmt.Errorf("spec.type %q is not a registered authorizer backend (registered: %v)", r.Spec.Type, authorizers.RegisteredTypes())
	}
	if len(r.Spec.Config.Raw) == 0 {
```
(Adjust the helper name to whatever the registry actually exposes.)

ChrisJBurns (commented):
## Multi-agent review summary

Five specialist reviewers (operator, Go, authz/security, tests, general) examined this PR; codex cross-review was skipped (CLI not installed). The design — a backend-agnostic MCPAuthzConfig CRD with a `RawExtension` payload, factory-driven validation, and a stacked `AuthzConfigRef` on workload specs — is the right shape, and the multi-version CRD plumbing (v1alpha1 compat shell embedding v1beta1, generated YAML / deepcopy / helm / RBAC / docs / chainsaw asserts) is internally consistent.

The implementation correctness is mostly good but inherits the same pre-existing anti-patterns the codebase rules flag — direct `r.Status().Update` and `r.Update` rather than `MutateAndPatchStatus` / `MutateAndPatchSpec`. There are also a handful of defense-in-depth gaps worth tightening before #4778 wires `AuthzConfigRef` into the workload controllers, since those gaps become user-visible at that point.

After consensus scoring there are **0 HIGH, 6 MEDIUM, 10 LOW** findings. No blocking issues, but several MEDIUMs cluster around operator-rule compliance, hash determinism, and CEL test coverage.

### Consensus summary

| # | Finding | Severity | Action |
|---|---------|----------|--------|
| F1 | AuthzConfigRef field added but no workload controller consumes it (intentional per PR description, but `Deprecated:` annotation pre-announces capability) | MEDIUM | Fix/confirm |
| F2 | Status writes use `r.Status().Update` instead of `MutateAndPatchStatus` | MEDIUM | Fix |
| F3 | Finalizer writes use `r.Update` instead of `MutateAndPatchSpec` | MEDIUM | Fix |
| F4 | `ReferencingWorkloads` refresh skipped on hash-change branch | LOW | Fix |
| F5 | Hash over `RawExtension` flaps on whitespace / key-order | MEDIUM | Fix |
| F6 | `BuildFullAuthzConfigJSON` lets factory key shadow reserved `version`/`type` keys | MEDIUM | Fix |
| F7 | CEL mutual-exclusion rules on three CRD types have no envtest coverage | MEDIUM | Fix |
| F8 | `DeletionBlocked` condition never cleared on success path | LOW | Fix |
| F9 | `findReferencingWorkloads` error swallowed; reconcile returns success | LOW | Fix |
| F10 | `BuildFullAuthzConfigJSON` exported but only used inside `controllers` package | LOW | Fix |
| F11 | No regression test for `Status.Conditions` wholesale-overwrite race | MEDIUM | Fix |
| F12 | No test for finalizer-removed-after-last-reference transition | LOW | Fix |
| F13 | `MCPAuthzConfig.Validate()` doesn't check Type against registry | LOW | Fix |
| F14 | kubebuilder RBAC marker omits `mcpservers` which the controller watches | LOW | Fix |
| F15 | No K8s events emitted on validation failure / deletion-block | LOW | Optional |
| F16 | Migration path from inline `authzConfig` to `authzConfigRef` undocumented | LOW | Optional |

Detailed comments are inline. Pros, tradeoffs and a suggested test plan are below.

### Pros

- Backend-agnostic envelope (`ConfigKey()`) is a clean way to support multiple authorizers without hardcoding Cedar specifics.
- `findStaleRefs` extraction is a notable improvement over the duplicated loops in `MCPOIDCConfig` — worth back-porting.
- Replacing `mustMarshalJSON`'s panic with an error-returning helper is the right direction (even if the helper is over-engineered, see review comments).
- Multi-version CRD plumbing is correctly applied; deepcopy, CRD YAML, helm templates, RBAC and chainsaw asserts all stay in sync.
- Stale-ref enqueue + sorting/dedup keep `Status.ReferencingWorkloads` stable across reconciles.
- Large-PR justification is accurate — the majority of additions are generated.

### Risks worth weighing

- The PR adds `AuthzConfigRef` to the workload specs and annotates the existing inline field as `Deprecated:` to point at it, but **no workload controller resolves the ref into a runtime authz config in this PR**. Until #4778 lands, a user who follows the deprecation guidance gets a workload running with no authorization enforced. The PR description discloses this; the field-level docs do not.
- Status / finalizer writes use raw `Update` calls, which the codebase's own operator rules (`.claude/rules/operator.md`) explicitly call out as the anti-pattern they exist to prevent. Sibling `MCPOIDCConfig` controller has the same issue, so this isn't a regression, but new code is the natural place to stop the bleed.

### Suggested manual verification

```bash
# Generated-artifact drift check
task operator-generate
task operator-manifests
task -d cmd/thv-operator crdref-gen
git diff --exit-code   # must be clean

# Smoke: valid + invalid MCPAuthzConfig
kubectl apply -f - <<EOF
kind: MCPAuthzConfig
metadata: { name: cedar-test, namespace: default }
spec:
  type: cedarv1
  config:
    version: "1.0"
    policies: ["permit(principal, action, resource);"]
    entitiesJson: "[]"
EOF
kubectl get mcpauthzconfig cedar-test -o yaml   # expect Valid=True, referenceCount=0

kubectl apply -f - <<EOF
kind: MCPAuthzConfig
metadata: { name: bad, namespace: default }
spec: { type: nope, config: {} }
EOF
kubectl get mcpauthzconfig bad -o jsonpath='{.status.conditions}'   # expect Valid=False
```

To confirm the silent-no-op claim under F1: apply only the `authzConfigRef` form (no inline `authzConfig`) on an MCPServer, then exec into the resulting pod and look for the authorization middleware in the runner config — at the current PR head it will be absent.

---
*Automated multi-agent review (5 specialists, no consensus disputes). The skill flagged 16 findings; 13 below the consensus threshold were dropped (see local review log).*

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+	// authzConfigVersion is the default version for reconstructed authz configs
+	authzConfigVersion = "1.0"
+)
+
+// MCPAuthzConfigReconciler reconciles a MCPAuthzConfig object.
+//
+// This controller manages the lifecycle of MCPAuthzConfig resources: validation
+// via the authorizer factory registry, config hash computation, finalizer management,
+// reference tracking, and deletion protection when workloads reference this config.
+type MCPAuthzConfigReconciler struct {
```
ChrisJBurns, line 45: **F15 — LOW (optional):** The reconciler doesn't take an `EventRecorder`. For a policy resource this is meaningful: cluster operators investigating a missing/invalid authz policy will not see `kubectl describe mcpauthzconfig` events explaining "Cedar policy rejected" or "deletion blocked by N workloads". Sibling controllers in this repo wire a recorder; adopting the same pattern here would improve the audit trail and discoverability.

Consider adding a `Recorder record.EventRecorder` field and wiring it in `SetupWithManager` via `mgr.GetEventRecorderFor("mcpauthzconfig-controller")`. Emit Warning events on `ConfigInvalid` / `DeletionBlocked`, Normal events on recovery. Avoid putting raw policy contents in event messages (the HTTP PDP URL in particular is topology-sensitive).

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpauthzconfigs,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpauthzconfigs/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpauthzconfigs/finalizers,verbs=update
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=virtualmcpservers,verbs=get;list;watch
+// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpremoteproxies,verbs=get;list;watch
```
ChrisJBurns, line 55: **F14 — LOW:** The kubebuilder markers above declare watch perms on `virtualmcpservers` and `mcpremoteproxies`, but `findReferencingWorkloads` and `SetupWithManager` also list/watch `mcpservers` (see line ~135 and the `Watches(&mcpv1beta1.MCPServer{}, ...)` registration). It works at runtime because other controllers already declare those permissions in the same clusterrole, but the marker mismatch will silently strip required permissions on a future regeneration if those other declarations move.

Add `// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpservers,verbs=get;list;watch` here and rerun `task operator-manifests` (the rendered clusterrole should be unchanged).

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+
+	// Check if the MCPAuthzConfig is being deleted
+	if !authzConfig.DeletionTimestamp.IsZero() {
+		return r.handleDeletion(ctx, authzConfig)
+	}
+
+	// Add finalizer if it doesn't exist
+	if !controllerutil.ContainsFinalizer(authzConfig, AuthzConfigFinalizerName) {
+		controllerutil.AddFinalizer(authzConfig, AuthzConfigFinalizerName)
+		if err := r.Update(ctx, authzConfig); err != nil {
```
ChrisJBurns: **F3 — MEDIUM:** `r.Update(ctx, authzConfig)` here (and the matching `RemoveFinalizer` `r.Update` later in `handleDeletion`) violates the operator rule "Spec / metadata patching" in `.claude/rules/operator.md`: spec/metadata writes must go through `controllerutil.MutateAndPatchSpec` so optimistic-lock merge-patch protects against zeroing fields written by other controllers, and finalizer-array merge semantics don't erase a concurrently-added finalizer from a different writer.

```go
// Recommended
if err := ctrlutil.MutateAndPatchSpec(ctx, r.Client, authzConfig, func(c *mcpv1beta1.MCPAuthzConfig) {
    controllerutil.AddFinalizer(c, AuthzConfigFinalizerName)
}); err != nil { ... }
```
Mirror for the remove path in `handleDeletion`. The sibling `MCPOIDCConfig` controller has the same issue — worth a follow-up to apply the same fix there.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+	if err := r.validateAuthzConfig(authzConfig); err != nil {
+		logger.Error(err, "MCPAuthzConfig spec validation failed")
+		meta.SetStatusCondition(&authzConfig.Status.Conditions, metav1.Condition{
+			Type:               mcpv1beta1.ConditionTypeAuthzConfigValid,
+			Status:             metav1.ConditionFalse,
+			Reason:             mcpv1beta1.ConditionReasonAuthzConfigInvalid,
+			Message:            err.Error(),
+			ObservedGeneration: authzConfig.Generation,
+		})
+		if updateErr := r.Status().Update(ctx, authzConfig); updateErr != nil {
```
ChrisJBurns: **F2 — MEDIUM:** This and the three other `r.Status().Update(ctx, authzConfig)` calls in this file (lines 127, 147, 276) violate `.claude/rules/operator.md` §"Status Writes". The rule mandates `controllerutil.MutateAndPatchStatus` because JSON merge-patch replaces the entire `Status.Conditions` array wholesale (the `+listType=map` marker is honored only by strategic-merge-patch). With `Status().Update` here, any concurrent writer that updates conditions on this CRD between this controller's Get and Update silently overwrites the entire conditions array — including condition types this controller never touches.

```go
// Recommended
if err := ctrlutil.MutateAndPatchStatus(ctx, r.Client, fresh, func(c *mcpv1beta1.MCPAuthzConfig) {
    meta.SetStatusCondition(&c.Status.Conditions, metav1.Condition{...})
}); err != nil { ... }
```
Apply the same fix to the other three call sites. The sibling `MCPOIDCConfig` controller has the same pattern; consider a follow-up there as well.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+	conditionChanged := meta.SetStatusCondition(&authzConfig.Status.Conditions, metav1.Condition{
+		Type:               mcpv1beta1.ConditionTypeAuthzConfigValid,
+		Status:             metav1.ConditionTrue,
+		Reason:             mcpv1beta1.ConditionReasonAuthzConfigValid,
+		Message:            "Spec validation passed",
+		ObservedGeneration: authzConfig.Generation,
+	})
+
+	// Calculate the hash of the current configuration
+	configHash := ctrlutil.CalculateConfigHash(authzConfig.Spec)
```
ChrisJBurns: **F5 — MEDIUM:** `ctrlutil.CalculateConfigHash(authzConfig.Spec)` hashes the spec struct including `Spec.Config` (a `runtime.RawExtension`). `RawExtension` stores the user's raw bytes verbatim, so two semantically-equivalent configs that differ only in whitespace, key ordering, or duplicate keys will hash differently and cause spurious status flips (e.g. on controller restart, on admission re-serialization, or after a noop re-apply). Conversely, two byte-different RawExtensions that round-trip to the same JSON object will hash differently even though the runtime behavior is identical.

Either (a) hash a canonical form (`unmarshal → re-marshal` to normalize key order and whitespace) before passing to `CalculateConfigHash`, or (b) document explicitly that the hash is over raw bytes and accept the flap. Either way, a unit test that proves the same logical config with different whitespace/key-order produces the same `ConfigHash` would lock the behavior in.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+
+	// Check if the hash has changed
+	hashChanged := authzConfig.Status.ConfigHash != configHash
+	if hashChanged {
```
ChrisJBurns: **F4 — LOW:** When `hashChanged` is true, the reconciler updates the hash and returns early at line 132, skipping the `findReferencingWorkloads` refresh further down. That means immediately after a config-spec change, the `ReferenceCount` / `ReferencingWorkloads` shown by `kubectl get` can be stale until the next workload watch event. Eventually consistent (the watches converge), but worth fixing because:

1. The comment on the refresh block implies it runs every reconcile.
2. Collapsing the two `Status().Update` calls into one would also halve API-server status writes on the spec-change path.

Consider moving the references refresh above the `if hashChanged` branch and emitting a single `Status().Update`.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+
+		if err := r.Status().Update(ctx, authzConfig); err != nil {
+			logger.Error(err, "Failed to update MCPAuthzConfig status")
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Refresh ReferencingWorkloads list
+	referencingWorkloads, err := r.findReferencingWorkloads(ctx, authzConfig)
```
ChrisJBurns, line 116: **F9 — LOW:** When `findReferencingWorkloads` returns an error here, it's logged and the reconcile continues with whatever cached refs were on the object (potentially writing stale status), then returns `ctrl.Result{}, nil` — silently swallowing a transient API-server failure. Sibling controllers return the error so controller-runtime requeues with backoff.

```suggestion
	referencingWorkloads, err := r.findReferencingWorkloads(ctx, authzConfig)
	if err != nil {
		logger.Error(err, "failed to find referencing workloads")
		return ctrl.Result{}, err
	}
```

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+// authzConfigEnvelope is a minimal struct for extracting version and type from reconstructed JSON.
+type authzConfigEnvelope struct {
+	Version string `json:"version"`
+	Type    string `json:"type"`
+}
+
+// BuildFullAuthzConfigJSON reconstructs the full authorizer config JSON from a
+// MCPAuthzConfig spec. The result is the same format accepted by authorizers.Config
+// and used in ConfigMaps: {"version": "1.0", "type": "<type>", "<configKey>": {<config>}}.
+func BuildFullAuthzConfigJSON(spec mcpv1beta1.MCPAuthzConfigSpec) ([]byte, error) {
```
ChrisJBurns: **F10 — LOW:** `BuildFullAuthzConfigJSON` is exported but its only production caller lives in the same `controllers` package. Per `.claude/rules/go-style.md` ("Package API surface"), exporting locks the signature for callers that don't yet exist. The doc comment hints this will be used by workload controllers in the future — if so, it'll likely want to live in `cmd/thv-operator/pkg/controllerutil/` next to `BuildInlineCedarAuthzConfig` to avoid downstream packages importing the `controllers` package.

Suggestion: unexport (`buildFullAuthzConfigJSON`) for now; promote and relocate when #4778 needs an actual cross-package caller, at which point the consumer's needs may also reshape the signature.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+	versionJSON, err := marshalJSONString(authzConfigVersion)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal version: %w", err)
+	}
+	typeJSON, err := marshalJSONString(spec.Type)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal type: %w", err)
+	}
+
+	fullConfig := map[string]json.RawMessage{
```
ChrisJBurns, line 217: **F6 — MEDIUM:** This map literal lets the factory-supplied `configKey` collide with the reserved envelope keys. Go map literals with duplicate keys silently take the last assignment, so a factory whose `ConfigKey()` returns `"version"` or `"type"` (or `""`) silently overwrites the envelope metadata with attacker-controlled JSON. Today only Cedar (`"cedar"`) and HTTP (`"pdp"`) register, but the `AuthorizerFactory` interface is exported and the registration path is open to test stubs and future / out-of-tree backends, so this is a latent defense-in-depth gap.

Add a guard in `Register()` (`pkg/authz/authorizers/registry.go`) that rejects `ConfigKey()` values in a reserved set (`"version"`, `"type"`, `""`) — panicking at process startup makes a misconfigured factory visible immediately. A short unit test that registers a factory with `ConfigKey() == "version"` and expects a panic locks the contract in.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,457 @@
...
+			return ctrl.Result{}, err
+		}
+
+		if len(referencingWorkloads) > 0 {
+			logger.Info("MCPAuthzConfig is still referenced by workloads, blocking deletion",
+				"authzConfig", authzConfig.Name,
+				"referencingWorkloads", referencingWorkloads)
+
+			meta.SetStatusCondition(&authzConfig.Status.Conditions, metav1.Condition{
+				Type:               mcpv1beta1.ConditionTypeDeletionBlocked,
```
ChrisJBurns: **F8 — LOW:** `handleDeletion` sets `ConditionTypeDeletionBlocked = True` here, but nothing on the non-deletion reconcile path clears it back to `False` (or removes the condition). Outside the deletion flow, the main `Reconcile` only touches `ConditionTypeAuthzConfigValid`. If a user attempts deletion, then cancels (e.g. removes the deletionTimestamp via finalizer removal followed by re-apply), the stale `DeletionBlocked: True` condition can persist and mislead operators.

On the success path of `Reconcile` (after validation passes), call `meta.SetStatusCondition` to set `DeletionBlocked = False` (reason e.g. `NoReferences`) or `meta.RemoveStatusCondition` if `DeletionTimestamp.IsZero()`. Same pattern is used by sibling config controllers.

cmd/thv-operator/controllers/mcpauthzconfig_controller_test.go
```diff
@@ -0,0 +1,757 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, time.Duration(0), result.RequeueAfter)
+
+	var afterSteady mcpv1beta1.MCPAuthzConfig
+	require.NoError(t, fakeClient.Get(ctx, req.NamespacedName, &afterSteady))
+	assert.Equal(t, initialHash, afterSteady.Status.ConfigHash, "Hash should not change")
+	assert.Equal(t, initialRV, afterSteady.ResourceVersion, "ResourceVersion should not change (no writes)")
+}
+
+func TestMCPAuthzConfigReconciler_ValidationFailureSetsCondition(t *testing.T) {
```
ChrisJBurns, line 317: **F11 — MEDIUM:** `.claude/rules/operator.md` §"Status Writes" warns that direct `r.Status().Update` (used by this controller — see F2) replaces `Status.Conditions` wholesale on CRDs via JSON merge-patch. There is no regression test that interleaves two condition writers (e.g. a validation-failure reconcile concurrent with a hash-change reconcile, or an externally-injected condition + a reconcile) and asserts that all set conditions survive.

Without such a test, a future change that re-introduces the race (or that converts the controller to `MutateAndPatchStatus` and gets it wrong) will go undetected. Suggested approach: pre-populate a fake `MCPAuthzConfig` with an externally-injected condition type the controller doesn't manage, reconcile once, then assert that condition is still present alongside the controller-managed `Valid` condition. This is the cheapest possible canary against the wholesale-overwrite class of bug.

cmd/thv-operator/controllers/mcpauthzconfig_controller_test.go
```diff
@@ -0,0 +1,757 @@
...
+			foundCondition = true
+			assert.Equal(t, metav1.ConditionFalse, cond.Status, "Valid condition should be False")
+			assert.Equal(t, mcpv1beta1.ConditionReasonAuthzConfigInvalid, cond.Reason)
+			break
+		}
+	}
+	assert.True(t, foundCondition, "Should have a Valid condition")
+}
+
+func TestMCPAuthzConfigReconciler_handleDeletion(t *testing.T) {
```
ChrisJBurns, line 357: **F12 — LOW:** `TestMCPAuthzConfigReconciler_handleDeletion` covers two static starting states (no workloads → finalizer removed; one workload → blocked/requeue) but never exercises the *transition* that matters in practice: a config blocked by N references, then the last referencing workload is deleted, then `handleDeletion` is invoked again and the finalizer is removed.

Add a sub-test that (1) creates a deleting config plus one referencing workload, invokes `handleDeletion` (assert requeue + finalizer still present), (2) deletes the referencing workload, invokes `handleDeletion` again (assert finalizer removed). This is the only path through the deletion-protection logic that actually flips state, and it is currently uncovered.

cmd/thv-operator/api/v1beta1/mcpserver_types.go
ChrisJBurns: Resolved in 4b647922 — kept the field, dropped the premature `Deprecated:` annotation on inline `AuthzConfig`, and added an explicit "consumed by workload controllers in a follow-up PR" note on `AuthzConfigRef` (mirrored across MCPServer, MCPRemoteProxy, and VirtualMCPServer.IncomingAuth).

cmd/thv-operator/api/v1beta1/mcpserver_types.go
ChrisJBurns: Resolved in a16d018a — added envtest cases for the three new mutex rules: mcp-server CEL test extended, plus new `mcpremoteproxy_authzconfig_cel_test.go` and `virtualmcpserver_authzconfig_cel_test.go` files. Each suite covers only-inline / only-ref / both-set; the both-set case asserts the apiserver rejects with the expected message. Runs via `task operator-test-integration`.

cmd/thv-operator/api/v1beta1/mcpserver_types.go
ChrisJBurns: Partially resolved in 4b647922: the `Deprecated:` annotation is gone (it was directing users at a non-functional field). The migration doc itself is deferred to #4778 — once `AuthzConfigRef` actually wires through, the migration becomes meaningful and documenting it before then would describe a non-functional path.

cmd/thv-operator/api/v1beta1/mcpauthzconfig_types.go
ChrisJBurns: Resolved in 586e212a — `Validate()` now calls `authorizers.IsRegistered(r.Spec.Type)` directly. Pulled the registry import into the API package (verified no cycle via `pkg/authz/authorizers`); the unit tests blank-import the cedar backend so the existing `cedarv1` cases continue to pass, and a new `"unregistered type fails"` case asserts the new branch.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Resolved in 3449b123 — added `// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=mcpservers,verbs=get;list;watch` to the markers block. Confirmed via `task operator-manifests` that the rendered clusterrole and chainsaw asserts are unchanged (other controllers already declared the permission).

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Resolved in 62aeb65c — both finalizer writes (add at line 81, remove in handleDeletion) now go through `ctrlutil.MutateAndPatchSpec` with the mutation inside the closure, so finalizer-array merge semantics are guarded by the optimistic-lock precondition.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Resolved in 62aeb65c — all four status writes now go through `ctrlutil.MutateAndPatchStatus` with mutations inside the closures (a small `setValidTrueCondition` helper factors out the Valid=True transition so the success path doesn't duplicate the literal). Followed up in 89e282ad by collapsing the hash + refs branches into a single status patch — see F4.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Resolved in 4afba123 — `canonicalizeSpecForHash` re-marshals `Spec.Config.Raw` (Go's encoding/json sorts map keys) before `CalculateConfigHash` walks the spec. `TestCanonicalizeSpecForHash` proves three variants (compact / indented / reordered keys) produce the same hash. Malformed JSON is returned unchanged so `Validate`/`validateAuthzConfig` remain the source of the user-facing error.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Resolved in 89e282ad — `findReferencingWorkloads` now runs once before the hash check, and the hash + refs updates collapse into a single `MutateAndPatchStatus` call. `TestMCPAuthzConfigReconciler_HashAndRefsLandInOneReconcile` locks in the property: after a single success-path reconcile, both `ConfigHash` and `ReferenceCount` reflect the live state.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Resolved in 89e282ad — `findReferencingWorkloads` errors now return up to controller-runtime so the reconcile is retried with backoff, rather than being logged and swallowed.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Resolved in f59d84cf — renamed to `buildFullAuthzConfigJSON` and updated the in-package callers and tests. Confirmed via grep that no out-of-package consumer existed. When #4778 needs to call it, the natural promotion target is `cmd/thv-operator/pkg/controllerutil/authz.go` next to `BuildInlineCedarAuthzConfig`.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Resolved in 586e212a — `Register()` now panics if a factory's `ConfigKey()` returns one of the reserved envelope keys (`""`, `"version"`, `"type"`). New `TestRegisterPanicsOnReservedConfigKey` table-test covers all three reserved values. Process-startup misconfiguration so the failure is loud.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Resolved in 89e282ad — the steady-state success path now `meta.RemoveStatusCondition(&c.Status.Conditions, ConditionTypeDeletionBlocked)` inside the same `MutateAndPatchStatus` closure that lands Valid=True. New `TestMCPAuthzConfigReconciler_ClearsDeletionBlockedOnSuccessPath` regression test seeds a stale `DeletionBlocked=True` and asserts it's gone after one reconcile.

cmd/thv-operator/controllers/mcpauthzconfig_controller_test.go
ChrisJBurns: Resolved in 52c7dc61 — new `TestMCPAuthzConfigReconciler_PreservesForeignConditions` is the canary. It seeds the resource with a foreign condition the controller doesn't own, reconciles, then asserts the foreign condition survives alongside the controller-owned `Valid=True`. Fails (intentionally) if anyone ever re-introduces `r.Status().Update` or pre-mutates conditions outside a `MutateAndPatchStatus` closure.

cmd/thv-operator/controllers/mcpauthzconfig_controller_test.go
ChrisJBurns: Resolved in 52c7dc61 — new `TestMCPAuthzConfigReconciler_FinalizerRemovedAfterLastRefDropped` exercises the transition: deleting config + workload referencing it, `handleDeletion` (requeue + finalizer present), delete the workload via the fake client, `handleDeletion` again (finalizer removed).

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Deferred to a parity follow-up rather than bundled into this PR — `.claude/rules/operator.md` §"Status Condition Parity" calls for adding observability features consistently across sibling CRDs, so the right shape is one PR that adds EventRecorder to MCPOIDCConfig, MCPExternalAuthConfig, and MCPAuthzConfig together.

Tracked in #5496.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
ChrisJBurns: Follow-up filed for the two sibling controllers (`MCPOIDCConfig`, `MCPExternalAuthConfig`) that still use raw `r.Status().Update` / `r.Update`: #5497.

ChrisJBurns:
## Pushed 12 commits addressing the multi-agent review

Hi @JAORMX — taking over this PR briefly to address the [consensus review feedback I posted earlier](https://github.com/stacklok/toolhive/pull/4777#pullrequestreview-4402584109). All 16 findings now either landed in commits or moved to follow-up issues, plus a merge of `main` and one extra fix from the older 2026-04-13 review.

### Commits (in order)

| # | Commit | Finding |
|---|--------|---------|
| 1 | [586e212a](https://github.com/stacklok/toolhive/commit/586e212a) — Guard reserved `ConfigKey` values and tighten `MCPAuthzConfig.Validate` | F6, F13 |
| 2 | [4afba123](https://github.com/stacklok/toolhive/commit/4afba123) — Canonicalize MCPAuthzConfig spec before hashing | F5 |
| 3 | [62aeb65c](https://github.com/stacklok/toolhive/commit/62aeb65c) — Use `MutateAndPatchStatus`/`Spec` in MCPAuthzConfig controller | F2, F3 |
| 4 | [89e282ad](https://github.com/stacklok/toolhive/commit/89e282ad) — Tighten MCPAuthzConfig reconcile flow | F4, F8, F9 |
| 5 | [f59d84cf](https://github.com/stacklok/toolhive/commit/f59d84cf) — Unexport `buildFullAuthzConfigJSON` | F10 |
| 6 | [3449b123](https://github.com/stacklok/toolhive/commit/3449b123) — Declare `mcpservers` RBAC marker | F14 |
| 7 | [52c7dc61](https://github.com/stacklok/toolhive/commit/52c7dc61) — Test condition preservation and last-ref finalizer transition | F11, F12 |
| 8 | [a16d018a](https://github.com/stacklok/toolhive/commit/a16d018a) — Add envtest for `authzConfig`/`authzConfigRef` mutual exclusion | F7 |
| 9 | [4b647922](https://github.com/stacklok/toolhive/commit/4b647922) — Reframe `AuthzConfigRef` as part 1 of 2 in godocs | F1, F16 |
| 10 | [1e2d1f94](https://github.com/stacklok/toolhive/commit/1e2d1f94) — Regenerate CRD manifests and reference docs | (generated) |
| 11 | [ee444498](https://github.com/stacklok/toolhive/commit/ee444498) — Merge `main` into the branch | conflict resolution |
| 12 | [193ff756](https://github.com/stacklok/toolhive/commit/193ff756) — Compound `(kind, name)` listMapKey on `ReferencingWorkloads` | original review finding #7 |

### Deferred — filed as follow-up issues

- **F15** — EventRecorder on MCPAuthzConfig: deferred to **#5496** so all three config controllers (`MCPOIDCConfig`, `MCPExternalAuthConfig`, `MCPAuthzConfig`) get one consistently (parity rule).
- **F16** — Migration doc for inline `authzConfig` → `authzConfigRef`: deferred to the workload-controller wiring PR (#4778), where the migration becomes user-meaningful. The premature `Deprecated:` annotation on `AuthzConfig` was dropped in commit 9.
- **`MutateAndPatchStatus`/`Spec` migration for the two sibling config controllers**: filed as **#5497**.
- **Compound listMapKey parity on `MCPOIDCConfig` / `MCPExternalAuthConfig`**: filed as **#5498**.

### Verification

- `task lint-fix` — 0 issues
- `task test` — all packages green (excluding pre-existing flake in `pkg/skills/client` on `main` and envtest-binary-required suites that run in CI)
- `task build` — clean
- `task operator-generate`, `task operator-manifests`, `task crdref-gen` — all run, working tree clean after regeneration

### Note on review state

@ChrisJBurns the [2026-04-13 review](https://github.com/stacklok/toolhive/pull/4777#pullrequestreview-4099809907) is still the active `CHANGES_REQUESTED` decision. All eight findings from that earlier review have now been addressed (CEL XValidation rules were already in PR; RBAC markers, mustMarshalJSON→error, nil-check on `Config.Raw`, watch-handler test coverage, `Validate()` method, listMapKey compound key, and the blank-import design discussion). When you get a chance, please re-review or dismiss so the PR can move forward.

cmd/thv-operator/controllers/mcpauthzconfig_controller.go
```diff
@@ -0,0 +1,475 @@
...
+
+	// strconv.Quote produces the same JSON-encoded form as json.Marshal on a
+	// Go string, without an error return: encoding/json's stringEncoder has
+	// no failure path for a valid Go string, which authzConfigVersion (a
+	// constant) and spec.Type (validated upstream) both are.
+	fullConfig := map[string]json.RawMessage{
+		"version":           json.RawMessage(strconv.Quote(authzConfigVersion)),
+		"type":              json.RawMessage(strconv.Quote(spec.Type)),
+		factory.ConfigKey(): spec.Config.Raw,
+	}
```
jhrozek: The comment's equivalence claim isn't quite accurate: `strconv.Quote` and `json.Marshal` do **not** produce the same bytes for all strings. `encoding/json` HTML-escapes `<`, `>`, and `&` (to `\u003c`/`\u003e`/`\u0026`) by default; `strconv.Quote` uses Go string-literal escaping and leaves those characters as-is.

No runtime impact today — `authzConfigVersion` is the constant `"1.0"` and `spec.Type` is validated to a registered backend (`cedarv1`/`httpv1`), none of which contain those characters, and both encodings decode to the same Go string. The risk is latent: a future type name containing `&`/`<`/`>` would diverge at the byte level from a `json.Marshal`-produced envelope, silently breaking any raw-byte comparison (cache key, golden file). The misleading comment is the bigger issue.

Using `json.Marshal` (which never errors on a `string`) removes the divergence and makes the comment unnecessary:

```suggestion
	versionJSON, _ := json.Marshal(authzConfigVersion)
	typeJSON, _ := json.Marshal(spec.Type)
	fullConfig := map[string]json.RawMessage{
		"version":           versionJSON,
		"type":              typeJSON,
		factory.ConfigKey(): spec.Config.Raw,
	}
```

Note: `strconv` becomes an unused import after this change (it's used nowhere else in the file) and will need to be dropped from the import block.

jhrozek (commented):
One inline note from a follow-up review pass on the authz-config envelope construction.

pkg/authz/authorizers/registry.go
```diff
@@ -14,6 +14,11 @@ import (
 // (e.g., Cedar, OPA) implements this interface to provide validation and
 // instantiation of authorizers from their specific configuration format.
 type AuthorizerFactory interface {
+	// ConfigKey returns the JSON key under which the backend-specific
+	// configuration is nested in the full authorizer config blob.
+	// For example, Cedar returns "cedar" and HTTP PDP returns "pdp".
+	ConfigKey() string
```
jhrozek, line 20: Adding `ConfigKey()` to the exported `AuthorizerFactory` interface is a **breaking change for out-of-tree implementers** — any backend that lives outside this repo won't compile until it adds the method. All in-tree implementers (`cedar`, `http`, and the test mocks) are updated in this PR, so the in-repo build is fine.

The compile-time break is the right trade-off here (preferable to a runtime nil/empty-key panic via the `reservedConfigKeys` guard), so no code change needed. Two coordination call-outs:

- Update any closed-source/enterprise authorizer backend that implements this interface in lockstep so a dependency bump doesn't break its build.
- Worth noting under the PR's breaking-change / user-facing-change section for downstream Go consumers.

jhrozek (commented):
Process note on the `AuthorizerFactory` interface change.

pkg/authz/authorizers/registry.go
ChrisJBurns: Agree on all counts — keeping the compile-time break.

Two things I followed up on:

**1. PR body updated.** Added a `Does this introduce a user-facing change?` section that calls out the breaking `AuthorizerFactory.ConfigKey()` addition, why we chose compile-time over runtime, and the reserved-key panic at `Register()` time.

**2. Swept the known internal Go consumers of `pkg/authz/authorizers`** — none implement `AuthorizerFactory` directly. They consume the `Authorizer` runtime interface and the `cedar.NewCedarAuthorizer` constructor, so the new method is transparent to them. Their `toolhive` dependency bump containing this PR will compile cleanly with no additional work. Any other closed-source consumer outside our visibility still needs the coordination heads-up you flagged.

> One related observation worth noting: at least one internal consumer hand-emits the envelope shape `{"version","type","cedar"}` with a hardcoded `"cedar"` json tag rather than going through the factory. The on-wire shape didn't change in this PR — but if anyone ever drifts upstream `Factory.ConfigKey()` and the inner `Options` json tag apart, that emitter is the silent failure point. The new `TestFactory_ConfigKeyMatchesStructTag` per backend (commit bc37dc15) is exactly the canary for that drift.

ChrisJBurns (approved):

jhrozek (approved):

Issue #4250, opened 2026-03-19
Remove Deprecated Fields


ChrisJBurns:
**As a** CRD API consumer,
**I want** deprecated fields removed in the v1alpha2 API,
**so that** the API surface is clean and there is no ambiguity about which field to use.

**Size**: M
**Dependencies**: None
**Labels**: `operator`, `api`, `breaking-change`

## Context

MCPServer and MCPRemoteProxy include deprecated fields with fallback helper methods. These were retained during v1alpha1 for backward compatibility, but v1alpha2 is a breaking change and the right time to remove them.

| Deprecated Field | Replacement | Location |
|---|---|---|
| `MCPServer.spec.port` | `proxyPort` | `mcpserver_types.go:96` |
| `MCPServer.spec.targetPort` | `mcpPort` | `mcpserver_types.go:103` |
| `MCPServer.spec.toolsFilter` | `toolConfigRef` | `mcpserver_types.go:175` |
| `InlineOIDCConfig.clientSecret` | `clientSecretRef` | `mcpserver_types.go:541` |
| `InlineOIDCConfig.thvCABundlePath` | `caBundleRef` | `mcpserver_types.go:553` |
| `MCPRemoteProxy.spec.port` | `proxyPort` | `mcpremoteproxy_types.go:48` |

## Acceptance Criteria

- [ ] `MCPServer.spec.port` removed (use `proxyPort`)
- [ ] `MCPServer.spec.targetPort` removed (use `mcpPort`)
- [ ] `MCPServer.spec.toolsFilter` removed (use `toolConfigRef`)
- [ ] `InlineOIDCConfig.clientSecret` removed (use `clientSecretRef`)
- [ ] `InlineOIDCConfig.thvCABundlePath` removed (use `caBundleRef`)
- [ ] `MCPRemoteProxy.spec.port` removed (use `proxyPort`)
- [ ] All fallback helper methods (`GetProxyPort()`, `GetMcpPort()`, etc.) simplified to use only the canonical field
- [ ] All controller code referencing deprecated fields is updated
- [ ] Unit tests updated to use only canonical fields
- [ ] Migration notes document the exact field mappings

Issue #3142, opened 2025-12-23
OIDC thvCABundlePath configuration not applied to token validator


danbarr:
## Problem

The `spec.oidcConfig.inline.thvCABundlePath` field in VirtualMCPServer CRDs is defined and passed through multiple layers of configuration, but is never actually used when creating the OIDC token validator. This prevents VirtualMCPServer from connecting to OIDC providers that use self-signed certificates or certificates issued by corporate/internal CAs.

## Impact

When using a VirtualMCPServer with OIDC authentication against a provider with a non-publicly-trusted certificate (e.g., Keycloak with cert-manager internal cluster issuer), TLS verification fails even when:
- The CA certificate is mounted into the pod
- The `thvCABundlePath` is correctly configured in the CRD

This affects:
- JWKS fetching
- OIDC discovery
- Token introspection

All OIDC HTTP connections will fail with TLS verification errors.

## Root Cause

The bug is in `pkg/runner/config_builder.go` in the `WithOIDCConfig` function (lines 331-341). When creating the `TokenValidatorConfig`, it does NOT populate the `CACertPath` and `AuthTokenFile` fields:

```go
b.config.OIDCConfig = &auth.TokenValidatorConfig{
    Issuer:            oidcIssuer,
    Audience:          oidcAudience,
    JWKSURL:           oidcJwksURL,
    IntrospectionURL:  oidcIntrospectionURL,
    ClientID:          oidcClientID,
    ClientSecret:      oidcClientSecret,
    AllowPrivateIP:    jwksAllowPrivateIP,
    InsecureAllowHTTP: insecureAllowHTTP,
    Scopes:            scopes,
    // MISSING: CACertPath and AuthTokenFile are not set here!
}
```

The values are stored on `RunConfig.ThvCABundle` and `RunConfig.JWKSAuthTokenFile` but never passed to the `OIDCConfig` structure that's actually used by the token validator.

## Evidence

The test files explicitly acknowledge this bug with comments in `cmd/thv-operator/controllers/mcpserver_runconfig_test.go` (lines 407-408, 810-811):

```go
// NOTE: CACertPath and AuthTokenFile are not currently mapped in WithOIDCConfig function
// This is likely a bug that should be fixed separately
assert.Equal(t, "", runConfig.OIDCConfig.CACertPath)
assert.Equal(t, "", runConfig.OIDCConfig.AuthTokenFile)
```

## Configuration Flow

The field IS correctly propagated through most of the stack:

[...]

PR #4784, opened 2026-04-13
Propagate protectedResourceAllowPrivateIP through OIDC resolver and converter

ChrisJBurns:
## Summary

The `protectedResourceAllowPrivateIP` CRD field was defined on both MCPServer and VirtualMCPServer but never actually propagated through the OIDC resolver, making the field a no-op. Additionally, the vmcpconfig converter incorrectly mapped it from `jwksAllowPrivateIP`, meaning the two fields could not be controlled independently.

Fixes bugs 2 and 3 from #3142 (Bug 1 was fixed separately via #4250).
The MCPServer/MCPRemoteProxy runner path remains out of scope — tracked in #4787.

<details>

- Add `ProtectedResourceAllowPrivateIP` field to the internal `oidc.OIDCConfig` struct so it can carry the value from CRD types through to consumers
- Propagate the field through all three user-facing resolve paths: inline, shared inline (MCPOIDCConfig), and ConfigMap
- Fix the vmcpconfig converter to map `ProtectedResourceAllowPrivateIP` from the correct source field (was incorrectly using `JWKSAllowPrivateIP`)
- Add the missing `JwksAllowPrivateIP` mapping in the converter so the auth factory's OR logic (`ProtectedResourceAllowPrivateIP || JwksAllowPrivateIP`) works correctly
- Add missing `JWKSURL` and `IntrospectionURL` mappings in `mapResolvedOIDCToVmcpConfigFromRef` (pre-existing gap fixed opportunistically)
- Document runtime OR behavior on both CRD fields

</details>

<details>

| File | Change |
|------|--------|
| `cmd/thv-operator/pkg/oidc/resolver.go` | Add `ProtectedResourceAllowPrivateIP` to `OIDCConfig` struct; propagate in `resolveInlineConfig`, `resolveFromInlineSharedConfig`, `resolveConfigMapConfig`; extract `configMapBoolTrue` constant |
| `cmd/thv-operator/pkg/vmcpconfig/converter.go` | Fix `mapResolvedOIDCToVmcpConfig` and `mapResolvedOIDCToVmcpConfigFromRef` to use correct source field; add missing `JwksAllowPrivateIP`, `JWKSURL`, `IntrospectionURL` mappings |
| `cmd/thv-operator/api/v1alpha1/mcpoidcconfig_types.go` | Document runtime OR behavior on private IP fields |
| `cmd/thv-operator/api/v1alpha1/mcpserver_types.go` | Document runtime OR behavior on private IP fields |
| `cmd/thv-operator/pkg/oidc/resolver_test.go` | Update "all fields" tests; add independent propagation test cases for inline and ConfigMap |
| `cmd/thv-operator/pkg/oidc/resolver_configref_test.go` | Add shared inline config propagation test |
| `cmd/thv-operator/pkg/vmcpconfig/converter_test.go` | Update "maps all fields" test; add independent field mapping test |

</details>

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

Setting `protectedResourceAllowPrivateIP: true` on VirtualMCPServer OIDC config now works as documented. Previously the field was silently ignored.

**Note**: The field remains a no-op for MCPServer and MCPRemoteProxy — those resource types use a different runner path that only supports `jwksAllowPrivateIP`. This is tracked in #4787.

**Behavioral note**: Setting `jwksAllowPrivateIP: true` previously had the unintended side effect of also enabling private IPs for the protected resource endpoint. After this fix, users who need both must set both fields explicitly.

## Test plan

- [x] `task lint-fix` passes (0 issues on affected packages)
- [x] `task test` passes on `cmd/thv-operator/pkg/oidc/...` and `cmd/thv-operator/pkg/vmcpconfig/...`
- [x] `task build` passes for `cmd/thv-operator/...`
- [x] New unit tests: independent propagation (resolver inline, ConfigMap, shared inline), independent field mapping (converter)
- [x] Existing tests updated to cover both fields

- The Kubernetes and ServiceAccount resolve paths intentionally do NOT propagate `ProtectedResourceAllowPrivateIP` — those paths auto-configure for cluster-internal OIDC where `JWKSAllowPrivateIP` is the relevant setting.
- MCPServer/MCPRemoteProxy runner path gap tracked in #4787.


ChrisJBurns (commented):
**Agents consulted**: kubernetes-expert, go-security-reviewer, code-reviewer, toolhive-expert
**Recommendation**: ✅ APPROVE

### Consensus Summary
| 1 | MCPServer/MCPRemoteProxy paths drop ProtectedResourceAllowPrivateIP | 8/10 | HIGH | Comment (pre-existing) |
| 2 | Downstream factory ORs both fields into single AllowPrivateIP | 7/10 | MEDIUM | Comment (pre-existing) |
| 3 | Missing ConfigMap independence test | 7/10 | LOW | Suggest |

### Overall
This is a clean, well-scoped bug fix. The core change -- adding `ProtectedResourceAllowPrivateIP` to the resolver's `OIDCConfig` struct, propagating it through all three resolution paths (inline, shared inline, ConfigMap), and fixing the converter mapping -- is correct and addresses both bugs from \#3142.

All three findings that survived consensus scoring are either pre-existing issues (F1, F2) or a minor test gap (F3). No agent identified any issues introduced by this PR. The test suite includes independence assertions that verify the two fields can be controlled separately, which directly guards against regression of the original bug. The behavioral change (separating `jwksAllowPrivateIP` from `protectedResourceAllowPrivateIP`) is clearly documented in the PR description.

The main pre-existing gaps to track: (1) MCPServer and MCPRemoteProxy CRD paths still don't propagate `ProtectedResourceAllowPrivateIP` through `runner.WithOIDCConfig`, and (2) the downstream auth factory ORs both fields into a single `AllowPrivateIP`, limiting the practical granularity of the separation.

---
*Generated with Claude Code*

cmd/thv-operator/pkg/vmcpconfig/converter.go
```diff
@@ -276,7 +276,8 @@ func mapResolvedOIDCToVmcpConfig(
 		Resource:                        resolved.ResourceURL,
 		JWKSURL:                         resolved.JWKSURL,
 		IntrospectionURL:                resolved.IntrospectionURL,
-		ProtectedResourceAllowPrivateIP: resolved.JWKSAllowPrivateIP,
+		ProtectedResourceAllowPrivateIP: resolved.ProtectedResourceAllowPrivateIP,
+		JwksAllowPrivateIP:              resolved.JWKSAllowPrivateIP,
```
ChrisJBurns, line 280: **\[F1\] MCPServer/MCPRemoteProxy paths drop ProtectedResourceAllowPrivateIP** (8/10 consensus, HIGH, pre-existing)

The fix here is correct for the VirtualMCPServer path. However, all three specialist agents noted that the MCPServer and MCPRemoteProxy paths through `runner.WithOIDCConfig` still don't propagate `ProtectedResourceAllowPrivateIP` -- that function's signature doesn't accept it, and the downstream `TokenValidatorConfig` has a single `AllowPrivateIP` field.

This is pre-existing and out of scope for this PR, but worth tracking as a follow-up issue so the field isn't silently ignored for those resource types.

*Raised by: kubernetes-expert, code-reviewer, toolhive-expert*

cmd/thv-operator/pkg/vmcpconfig/converter.go
```diff
@@ -276,7 +276,8 @@ func mapResolvedOIDCToVmcpConfig(
...
 		InsecureAllowHTTP:               resolved.InsecureAllowHTTP,
```
ChrisJBurns, line 281: **\[F2\] Downstream factory ORs both fields into single AllowPrivateIP** (7/10 consensus, MEDIUM, pre-existing)

Nice that `JwksAllowPrivateIP` is now correctly mapped here (was previously dropped entirely). Worth noting that downstream in `pkg/vmcp/auth/factory/incoming.go:176`, the auth factory computes `AllowPrivateIP: oidcCfg.ProtectedResourceAllowPrivateIP || oidcCfg.JwksAllowPrivateIP`, so setting either to `true` enables private IPs for all OIDC endpoints.

The CRD-level separation is semantically correct, but the runtime enforcement is coarser than users might expect. Consider documenting this behavior in the CRD field comments.

*Raised by: kubernetes-expert, toolhive-expert*

cmd/thv-operator/pkg/oidc/resolver_test.go
```diff
@@ -168,28 +168,30 @@ func TestResolve_ConfigMapType(t *testing.T) {
...
+					"issuer":                          "https://auth.example.com",
+					"audience":                        "test-audience",
+					"jwksUrl":                         "https://auth.example.com/.well-known/jwks.json",
+					"introspectionUrl":                "https://auth.example.com/introspect",
+					"clientId":                        "test-client",
+					"clientSecret":                    "test-secret",
+					"thvCABundlePath":                 "/etc/ssl/ca.pem",
+					"jwksAuthTokenPath":               "/etc/auth/token",
+					"jwksAllowPrivateIP":              "true",
+					"protectedResourceAllowPrivateIP": "true",
```
ChrisJBurns, line 180: **\[F3\] Missing ConfigMap independence test** (7/10 consensus, LOW)

The inline path (line 590) and shared inline path (`resolver_configref_test.go:157`) both have independence tests where `protectedResourceAllowPrivateIP: true` and `jwksAllowPrivateIP: false`. The ConfigMap path here sets both to `"true"` but has no test verifying they resolve independently.

Consider adding a ConfigMap test case where `jwksAllowPrivateIP: "true"` and `protectedResourceAllowPrivateIP` is absent (or `"false"`), asserting `JWKSAllowPrivateIP == true` and `ProtectedResourceAllowPrivateIP == false`.

*Raised by: kubernetes-expert, code-reviewer*

cmd/thv-operator/pkg/vmcpconfig/converter.go
ChrisJBurns: Agreed — this is pre-existing and intentionally out of scope. The design doc explicitly decided not to change the MCPServer/MCPRemoteProxy runner path because those resource types don't expose a protected resource metadata endpoint (the `AllowPrivateIP` on `TokenValidatorConfig` controls JWKS fetching only). Worth a follow-up issue to track.

cmd/thv-operator/pkg/vmcpconfig/converter.go
ChrisJBurns: Correct — the auth factory's OR logic at `incoming.go:176` means the runtime enforcement is coarser than the CRD-level separation. Adding a doc comment to the CRD fields noting this behavior is a reasonable follow-up.

cmd/thv-operator/pkg/oidc/resolver_test.go
ChrisJBurns: Good catch — added in a884b28f. The new test sets `jwksAllowPrivateIP: "true"` with `protectedResourceAllowPrivateIP` absent and asserts they resolve independently.

cmd/thv-operator/pkg/oidc/resolver.go
```diff
@@ -25,22 +25,24 @@ const (
...
+	ClientID                        string
+	ClientSecret                    string // #nosec G117 -- not a hardcoded credential, populated at runtime from config
+	ThvCABundlePath                 string
+	JWKSAuthTokenPath               string
+	ResourceURL                     string
+	JWKSAllowPrivateIP              bool
+	ProtectedResourceAllowPrivateIP bool
+	InsecureAllowHTTP               bool
+	Scopes                          []string
 }
```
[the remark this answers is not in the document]
ChrisJBurns: Valid — the PR description overclaimed. Updated it to clarify the fix is VirtualMCPServer-only. The MCPServer/MCPRemoteProxy runner path gap is now tracked in #4787. The design consideration there is whether `protectedResourceAllowPrivateIP` is even semantically meaningful for MCPServer (which doesn't expose protected resource metadata).

cmd/thv-operator/pkg/vmcpconfig/converter.go
```diff
@@ -319,7 +320,8 @@ func mapResolvedOIDCToVmcpConfigFromRef(
 		ClientID:                        resolved.ClientID,
 		Audience:                        resolved.Audience,
 		Resource:                        resolved.ResourceURL,
-		ProtectedResourceAllowPrivateIP: resolved.JWKSAllowPrivateIP,
+		ProtectedResourceAllowPrivateIP: resolved.ProtectedResourceAllowPrivateIP,
+		JwksAllowPrivateIP:              resolved.JWKSAllowPrivateIP,
 		InsecureAllowHTTP:               resolved.InsecureAllowHTTP,
 		Scopes:                          resolved.Scopes,
 	}
```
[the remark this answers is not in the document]
ChrisJBurns: Valid and agreed — since we were already modifying this function, fixed in ebaaa364. Added the missing `JWKSURL` and `IntrospectionURL` mappings to `mapResolvedOIDCToVmcpConfigFromRef` to match what `mapResolvedOIDCToVmcpConfig` already does.

jhrozek (approved):
LGTM -- clean bug fix with good test coverage

Issue #4741, opened 2026-04-10
vMCP Code Mode


jerm-dro:
## Overview

Add opt-in Starlark script execution ("code mode") to vMCP, allowing agents to batch multi-tool workflows into a single server-side call — reducing round-trips, context bloat, and inference cycles. Based on the prototype in PR #4714.

## Background

Agents today make sequential tool calls with model inference between each one. For multi-service workflows (e.g., incident triage across PagerDuty, Datadog, Slack, Jira), this means 10+ round-trips and significant token spend. Scripting is possible today only through CLI/shell access, which is less safe and unavailable for knowledge worker use cases where agents lack a shell environment. The prototype demonstrates ~15s improvement by executing a Starlark script server-side that calls tools directly, with `parallel()` fan-out — without requiring shell access.

This epic production-hardens the prototype and ships it as an opt-in feature. It does NOT implement the full RFC THV-0060 session initialization model.

## User Story Breakdown

| Story ID | Title | Description | Persona | Goal | Acceptance Criteria | GitHub Issue # |
|----------|-------|-------------|---------|------|---------------------|----------------|
| STORY-001 | Ship opt-in code mode for vMCP | As a platform engineer, I want my agents to be able to execute scripts on tools without shell access so that they can safely reduce context bloat and inference cycles | Platform engineer | Enable server-side script execution for agents | Config toggle, proper dispatch, configurable limits, timeouts, optimizer compat | |
| STORY-002 | Add observability for script execution | As a cluster operator, I want logging and metrics for script execution so that I can monitor and diagnose issues | Cluster operator | Operational visibility | Structured logs, execution metrics, telemetry integration | |
| STORY-003 | Track code mode adoption and usage | As a ToolHive developer, I want to know how many people are opting in to code mode and how many tool calls come through it so that I can gauge adoption and prioritize investment | ToolHive developer | Understand adoption | Opt-in counts, call volume comparison, telemetry pipeline | |

## Acceptance Criteria

- [ ] Code mode can be enabled per-VirtualMCPServer via config flag
- [ ] Agents can execute Starlark scripts with tool calls, loops, conditionals, and `parallel()`
- [ ] Script execution is bounded (step limits, concurrency caps, timeouts)
- [ ] Works correctly with optimizer enabled
- [ ] Observability: logs and metrics for script execution
- [ ] Adoption tracking: code mode opt-in and call volume metrics
- [ ] All linked user stories completed

## References

[...]

PR #4821, opened 2026-04-14
Add script middleware with config toggle for code mode

jerm-dro:
## Summary

- vMCP code mode (#4742) lets agents execute Starlark scripts server-side to batch multi-tool workflows into a single call. PR 1 introduced the script engine package; this PR adds the HTTP middleware that integrates it into the vMCP request pipeline.
- Adds a `ScriptEngine` config toggle (disabled by default) that controls whether the `execute_tool_script` virtual tool is available. When enabled, the middleware intercepts `tools/call` for script execution and injects the virtual tool into `tools/list` responses.

Part of #4742

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/config/config.go` | Add `ScriptEngineConfig` type and `ScriptEngine` field to Config |
| `pkg/script/response_capture.go` | Minimal `http.ResponseWriter` for capturing inner tool call responses |
| `pkg/script/middleware.go` | HTTP middleware: `tools/list` injection, `tools/call` interception, inner tool dispatch |
| `pkg/script/middleware_test.go` | Table-driven tests for config toggle, script execution, error paths |
| `pkg/vmcp/server/server.go` | Add `ScriptMiddleware` to Config, insert in Handler() chain after authz |
| `cmd/vmcp/app/commands.go` | Wire `createScriptMiddleware` from config |

No. The config toggle defaults to disabled. This becomes user-facing when CRD types are added in PR 4.

**Review by commit** — each commit introduces one concern:
1. Config type (`ScriptEngineConfig`)
2. Response capture writer (no `httptest.NewRecorder` in production)
3. Middleware implementation (tools/list injection, tools/call interception, inner dispatch)
4. Server wiring (Config field, Handler() chain, commands.go)
5. Unit tests

**Inner tool dispatch**: Scripts call tools through `next.ServeHTTP(capture, innerReq)` where `capture` is a purpose-built `responseCapture` writer. The parsed MCP request is cleared from context so the parser re-parses the synthetic request. This ensures inner calls flow through the full middleware chain (authz, discovery, etc.).

**Middleware chain placement**: Script middleware wraps after authz, so scripts only see authorized tools and inner tool calls are subject to authorization.

## Large PR Justification

- New self-contained middleware + config with no modifications to existing business logic
- ~50% of the diff is unit tests — splitting tests from implementation would make review harder
- Generated code (CRD manifests, deepcopy, crd-api.md) accounts for a significant portion of the diff


pkg/script/middleware_test.go
```diff
@@ -0,0 +1,319 @@
...
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	mcpparser "github.com/stacklok/toolhive/pkg/mcp"
+)
+
+func TestMiddleware(t *testing.T) {
```
jerm-dro, line 21: Blocker: these tests should assert that the synthetic ParsedMCPRequest matches the synthetic tool call.

pkg/script/middleware_test.go
```diff
@@ -0,0 +1,319 @@
...
+	t.Parallel()
```
jerm-dro, line 22: Blocker: these tests should assert that the synthetic requests pass through HTTP headers necessary for auth.

pkg/script/middleware_test.go
jerm-dro: Added `TestMiddleware_InnerToolCallParsedRequest` — asserts that the re-parsed `ParsedMCPRequest` has the correct `Method`, `ResourceID`, and `Arguments` for the inner tool call.

pkg/script/middleware_test.go
jerm-dro: Added `TestMiddleware_InnerToolCallPreservesAuthHeaders` — sends a request with `Authorization: Bearer test-token-123` and asserts the inner tool call carries the same header. The middleware already does `origReq.Header.Clone()` on line 217.

pkg/script/middleware.go
```diff
@@ -0,0 +1,422 @@
...
+		writeJSONRPCError(w, parsed.ID, -32000, "failed to fetch available tools")
+		return
+	}
+
+	// Build script.Tool wrappers that dispatch through the middleware chain
+	scriptTools := buildScriptTools(r, parsingNext, tools)
+
+	// Create and run executor
+	exec := New(scriptTools, cfg)
+	result, err := exec.Execute(r.Context(), scriptStr, data)
```
jhrozek, line 107: this passes the server-level context with its timeout, should we also have a separate timeout per execution of the starlark script?

pkg/vmcp/server/server.go
```diff
@@ -554,10 +559,11 @@ func (s *Server) Handler(_ context.Context) (http.Handler, error) {
 
 	// MCP endpoint - apply middleware chain (wrapping order, execution happens in reverse):
 	// Code wraps: auth+parser → audit → discovery → annotation-enrichment →
-	//   authz → backend-enrichment → MCP-parsing → telemetry
+	//   authz → script(+parser) → backend-enrichment → MCP-parsing → telemetry
```
jhrozek: script is before auth right? Or rather only script body is authed?

pkg/script/middleware.go
jerm-dro: Good catch — added a default 30s per-script execution timeout (`ScriptTimeout`) that bounds total wall-clock time including all tool calls. Applied via `context.WithTimeout` in `executor.Execute`, so tool `Call` closures that respect context cancellation will abort when the deadline is exceeded. Configurable via `scriptEngine.scriptTimeout`.

pkg/vmcp/server/server.go
jerm-dro: Fixed the comment — script runs before authz in the execution path. Auth (identity) runs before script, so the request IS authenticated. But authz does not see the `execute_tool_script` call itself. Inner tool calls dispatched from within scripts DO flow through authz, so individual tools are authorized.

jhrozek (dismissed):
The timeout and chain-order fixes address the two things I flagged.

jhrozek:
@jerm-dro apologizes for the delay in reviews, I missed you updated the PR. I'm approving to mostly signal I've re-reviewed the code but unfortunately there is a conflict.

jhrozek (dismissed):

jhrozek:
@jerm-dro apologies I missed the latest push during the public holidays, acked now

jerm-dro:
Closing this in favor of a re-implementation on top of Trey's vMCP interface refactor (#5419).

This PR added code mode as HTTP middleware that intercepts `tools/call` and injects `execute_tool_script` into `tools/list`. That's the same class of "meta" tool as the optimizer's `call_tool`/`find_tool` — a virtual tool that sits in front of the real tools and dispatches to them — so post-refactor it belongs at/near the optimizer's tool-call layer (as a `VMCP` decorator), not as a separate middleware that re-reads the body and clears the parsed request from context to re-dispatch. Rebuilding it there drops that machinery and keeps all the meta-tool logic in one place.

Parking the branch rather than reworking in place, since the refactor changes the seam this plugs into. Tracked under #4742 (see the [design note](https://github.com/stacklok/toolhive/issues/4741#issuecomment-4712451628)); I'll reopen against the new layer once #5419 lands.

PR #4841, opened 2026-04-15
Fix skill install for qualified namespace/name references and missing parent directory

samuv:
## Summary

- Fix `Install` to fall back to registry catalog lookup when a `namespace/name` reference (e.g. `io.github.stacklok/skill-creator`) fails OCI pull because the namespace is parsed as a registry host. Names with an explicit tag or digest (e.g. `ghcr.io/org/skill:v1`) or more than one `/` (e.g. `ghcr.io/org/skill`) are unambiguously OCI and do not fall back
- Extract `installFromResolvedRegistry` helper to share dispatch logic between `installFromRegistryLookup` and the new fallback path
- Fix `WriteFiles` failing to acquire the lock when the parent skills directory does not yet exist (e.g. first install to `~/.cline/skills/`)
- Fix pre-existing bug in `installFromResolvedRegistry` OCI branch: pass `result.Skill.Metadata.Name` instead of `opts.Name` (the OCI ref string) to `installAndRegister`, so group registration and rollback use the correct skill name

## Behavioral change note

Previously, a registry entry that existed but had no installable OCI or git package fell through and returned **404 Not Found**. It now returns **422 Unprocessable Entity** ("resolved but no installable package"), which is a more accurate response.

## Type of change

- [x] Bug fix

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/skills/skillsvc/skillsvc.go` | Add OCI→registry fallback in `Install` and `GetContent`; tighten guard to also block multi-segment refs; add debug log before fallback; fix `opts.Name` bug in OCI branch of `installFromResolvedRegistry` |
| `pkg/skills/skillsvc/skillsvc_test.go` | Extend `TestInstallQualifiedNameOCIFallback` with digest, multi-segment, and ambiguity cases; pin `Pull` mock arguments |
| `pkg/skills/gitresolver/writer.go` | `MkdirAll` parent before acquiring file lock |
| `pkg/skills/gitresolver/writer_test.go` | Add test for missing parent directory |

Yes — `thv skill install io.github.stacklok/<skill-name>` now works correctly instead of failing with "no such host".

### DIFF
--- a/pkg/skills/gitresolver/writer.go
+++ b/pkg/skills/gitresolver/writer.go
@@ -22,6 +22,12 @@ func WriteFiles(files []FileEntry, targetDir string, force bool) error {
 	// Sanitize targetDir early so all downstream os calls use the clean path.
 	targetDir = filepath.Clean(targetDir)
 
+	// Ensure the parent directory exists before acquiring the lock.
+	// WithFileLock opens targetDir+".lock", which requires the parent to exist.
+	if err := os.MkdirAll(filepath.Dir(targetDir), skills.DirPermissions); err != nil {
+		return fmt.Errorf("creating parent directory: %w", err)
+	}
+
 	return fileutils.WithFileLock(targetDir, func() error {
 		// Handle existing directory
 		if _, statErr := os.Stat(targetDir); statErr == nil { // lgtm[go/path-injection] #nosec G304
--- a/pkg/skills/skillsvc/skillsvc.go
+++ b/pkg/skills/skillsvc/skillsvc.go
@@ -240,12 +240,14 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
 			return s.installAndRegister(ctx, result, opts.Group, opts.Name, scope, opts.ProjectRoot)
 		}
 		// OCI pull failed — fall back to registry lookup for names that look
-		// like a qualified "namespace/name" (no ':' or '@'). Names with an
-		// explicit tag or digest are unambiguously OCI and should not trigger
-		// a registry search (e.g. "ghcr.io/org/skill:v1" should not fall back).
-		if strings.ContainsAny(opts.Name, ":@") {
+		// like a qualified "namespace/name" (no ':' or '@', exactly one '/').
+		// Names with an explicit tag or digest, or with more than one '/'
+		// (e.g. "ghcr.io/org/skill" or "ghcr.io/org/skill:v1"), are
+		// unambiguously OCI refs and must not trigger a registry search.
+		if strings.ContainsAny(opts.Name, ":@") || strings.Count(opts.Name, "/") > 1 {
 			return nil, ociErr
 		}
+		slog.Debug("OCI pull failed, attempting registry fallback", "name", opts.Name, "error", ociErr)
 		resolved, regErr := s.resolveFromRegistry(opts.Name)
 		if regErr != nil {
 			return nil, regErr
@@ -350,7 +352,10 @@ func (s *service) installFromResolvedRegistry(
 		if ociErr != nil {
 			return nil, ociErr
 		}
-		return s.installAndRegister(ctx, result, opts.Group, opts.Name, scope, opts.ProjectRoot)
+		// Use the skill name extracted from the artifact, not opts.Name which
+		// holds the OCI ref string. installFromOCI mutates its own copy of opts
+		// (Go pass-by-value), so the caller never sees the updated name.
+		return s.installAndRegister(ctx, result, opts.Group, result.Skill.Metadata.Name, scope, opts.ProjectRoot)
 	case resolved.GitURL != "":
 		slog.Info("resolved skill from registry (git)", "name", opts.Name, "git_url", resolved.GitURL)
 		opts.Name = resolved.GitURL
@@ -495,9 +500,9 @@ func (s *service) GetContent(ctx context.Context, opts skills.ContentOptions) (*
 
 	// OCI failed — try resolving via registry name lookup (e.g. "skill-creator"
 	// or "io.github.stacklok/skill-creator" from the catalog index).
-	// Skip for refs that are clearly OCI references (contain : or @) to avoid
-	// a wasted network round-trip searching for e.g. "skill:v1".
-	if strings.ContainsAny(ref, ":@") {
+	// Skip for refs that are clearly OCI references (contain ':', '@', or more
+	// than one '/') to avoid a wasted network round-trip.
+	if strings.ContainsAny(ref, ":@") || strings.Count(ref, "/") > 1 {
 		return nil, ociErr
 	}
 	resolved, regErr := s.resolveFromRegistry(ref)
--- a/pkg/skills/skillsvc/skillsvc.go
+++ b/pkg/skills/skillsvc/skillsvc.go
@@ -235,11 +235,25 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
 		)
 	}
 	if isOCI {
-		result, err := s.installFromOCI(ctx, opts, scope, ref)
-		if err != nil {
-			return nil, err
+		result, ociErr := s.installFromOCI(ctx, opts, scope, ref)
+		if ociErr == nil {
+			return s.installAndRegister(ctx, result, opts.Group, opts.Name, scope, opts.ProjectRoot)
 		}
-		return s.installAndRegister(ctx, result, opts.Group, opts.Name, scope, opts.ProjectRoot)
+		// OCI pull failed — fall back to registry lookup for names that look
+		// like a qualified "namespace/name" (no ':' or '@'). Names with an
+		// explicit tag or digest are unambiguously OCI and should not trigger
+		// a registry search (e.g. "ghcr.io/org/skill:v1" should not fall back).
+		if strings.ContainsAny(opts.Name, ":@") {
+			return nil, ociErr
+		}
+		resolved, regErr := s.resolveFromRegistry(opts.Name)
+		if regErr != nil {
+			return nil, regErr
+		}
+		if resolved != nil {
+			return s.installFromResolvedRegistry(ctx, opts, scope, resolved)
+		}
+		return nil, ociErr
 	}
 
 	// Plain skill name — validate and proceed with existing flow.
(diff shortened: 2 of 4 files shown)


pkg/skills/skillsvc/skillsvc.go
```diff
@@ -235,11 +235,25 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
...
+			return s.installAndRegister(ctx, result, opts.Group, opts.Name, scope, opts.ProjectRoot)
 		}
-		return s.installAndRegister(ctx, result, opts.Group, opts.Name, scope, opts.ProjectRoot)
+		// OCI pull failed — fall back to registry lookup for names that look
+		// like a qualified "namespace/name" (no ':' or '@'). Names with an
+		// explicit tag or digest are unambiguously OCI and should not trigger
+		// a registry search (e.g. "ghcr.io/org/skill:v1" should not fall back).
+		if strings.ContainsAny(opts.Name, ":@") {
+			return nil, ociErr
+		}
```
JAORMX: So, the `:@` check works well for the intended case (`io.github.stacklok/my-skill`), and I like the simplicity. But there's an edge case that bugs me...

Consider `myregistry.com/org/skill` (a legit 3-segment OCI ref, no tag). OCI pull fails (say, network blip), there's no `:` or `@`, so the fallback fires. `splitQualifiedName` splits on the last `/` giving namespace=`myregistry.com/org`, name=`skill`. If a registry entry happens to match that... we silently install from a completely different source than what the user typed.

The probability is low (namespace has to match exactly), but this is a supply-chain confusion vector in principle. A qualified `namespace/name` always has exactly one `/`, while multi-segment OCI refs have more. Something like:

```go
if strings.ContainsAny(opts.Name, ":@") || strings.Count(opts.Name, "/") > 1 {
    return nil, ociErr
}
```

...would close that gap without adding much complexity. What do you think?

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -235,11 +235,25 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
...
+		resolved, regErr := s.resolveFromRegistry(opts.Name)
```
JAORMX, line 250: Nit: before we try the fallback, it'd be nice to log the original OCI error at debug level. If someone's debugging why their install resolved to something unexpected, this breadcrumb would save them a lot of head-scratching.

Something like:
```go
slog.Debug("OCI pull failed, attempting registry fallback", "name", opts.Name, "error", ociErr)
```

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -337,6 +334,38 @@ func (s *service) installFromRegistryLookup(
...
+		if gitErr != nil {
+			return nil, gitErr
+		}
+		return s.installAndRegister(ctx, result, opts.Group, result.Skill.Metadata.Name, scope, opts.ProjectRoot)
+	}
+	return nil, httperr.WithCode(
+		fmt.Errorf("skill %q resolved from registry but has no installable package", opts.Name),
+		http.StatusUnprocessableEntity,
+	)
+}
```
JAORMX, line 371: Note that this is a subtle behavioral change for the existing `installFromRegistryLookup` path too (not just the new fallback). Previously, if a registry entry existed but had neither OCI nor git packages, the switch fell through and `installFromRegistryLookup` returned a 404 ("not found in local store or registry"). Now it returns 422 ("resolved but no installable package").

The new error is actually more correct and more helpful... but it IS a behavior change for existing callers. Worth mentioning in the PR description so reviewers are aware.

pkg/skills/gitresolver/writer.go
```diff
@@ -22,6 +22,12 @@ func WriteFiles(files []FileEntry, targetDir string, force bool) error {
 	// Sanitize targetDir early so all downstream os calls use the clean path.
 	targetDir = filepath.Clean(targetDir)
 
+	// Ensure the parent directory exists before acquiring the lock.
+	// WithFileLock opens targetDir+".lock", which requires the parent to exist.
+	if err := os.MkdirAll(filepath.Dir(targetDir), skills.DirPermissions); err != nil {
+		return fmt.Errorf("creating parent directory: %w", err)
+	}
```
JAORMX, line 29: The fix itself is correct and follows the same pattern we already use in `pkg/api/server.go:640-645` (which even has the comment "Ensure the discovery directory exists before acquiring the lock, since the lock file is created in the same directory.").

Two things though:

1. CodeQL flagged this line ("Uncontrolled data used in path expression"). The `#nosec G304` and `lgtm[go/path-injection]` annotations we use elsewhere in this file won't help here... those suppress Gosec and Semmle/LGTM respectively, not CodeQL. CodeQL uses its own suppression mechanism. This alert needs to be dismissed in GitHub's Security tab as a false positive (the path comes from `PathResolver.GetSkillPath`, a trusted internal source, as documented in the function's doc comment).

2. This bugfix has zero dedicated test coverage. All existing `WriteFiles` tests in `writer_test.go` create the parent directory before calling the function, so no test would actually fail if this fix got reverted. A test case where `targetDir` points into a non-existent parent directory would directly exercise the fix and is pretty simple to add.

pkg/skills/skillsvc/skillsvc_test.go
```diff
@@ -2931,6 +2931,208 @@ func TestInstallFromRegistryGitFallback(t *testing.T) {
 	}
 }
 
+// TestInstallQualifiedNameOCIFallback covers the scenario where Install
+// receives a "namespace/name" string, which is parsed as an OCI reference but
+// fails to pull because the namespace looks like a registry host. The service
+// must fall back to a registry catalogue lookup and complete the install from
```
JAORMX, line 2937: Good test structure overall. Four well-chosen cases, table-driven, parallel. The trick of relying on gomock's strict mode to assert `SearchSkills` is NOT called in the tag case is nice.

A couple of gaps I noticed:

- There's no test for the `@sha256:...` (digest) variant of the guard. Only `:` (tag) is tested. The `ContainsAny(":@")` check covers both, but a test would make the intent explicit.

- The `resolveFromRegistry` error path (line 250-252 in skillsvc.go) where `regErr != nil` through the new fallback isn't covered. A test with a qualified name that triggers an ambiguous registry match would exercise that.

- The first `Pull` call in tests 1, 2, and 4 uses `gomock.Any()` for all arguments. Pinning the third arg to the expected reference string (e.g., the qualified name with `:latest` appended by `nameref`) would catch regressions in how the reference is constructed before the first pull attempt.

JAORMX (changes requested):
Hey! Nice fix. The core approach is solid... the `:@` heuristic to distinguish "ambiguous namespace/name" from "unambiguous OCI ref with tag/digest" is clean and easy to reason about. The `installFromResolvedRegistry` extraction is a good refactor, and the writer.go fix addresses a real first-install bug.

I verified the locking story and it's fine. The fallback path in `Install()` never holds a lock when calling downstream installers, so no deadlock risk. `InstallOptions` is passed by value and slice fields are replaced wholesale, not mutated in place. All good there.

A few things I'd like addressed before merging:

**The multi-segment OCI edge case.** A ref like `myregistry.com/org/skill` (3 segments, no tag) passes the guard and triggers the fallback. The namespace filtering makes a false match unlikely, but it's a supply-chain confusion vector in principle. Adding `|| strings.Count(opts.Name, "/") > 1` to the guard closes this cleanly. See inline comment.

**The writer.go fix needs a test.** All existing `WriteFiles` tests pre-create the parent directory, so no test actually proves this fix works. A test with a non-existent parent would take ~10 lines and directly exercise the bugfix.

**CodeQL alert.** The `#nosec` / `lgtm` annotations won't suppress CodeQL (different tool). This needs to be dismissed in GitHub's Security tab.

The rest are nits and nice-to-haves (logging the original OCI error before fallback, noting the 404->422 behavioral change in the PR description, tightening `gomock.Any()` in tests). See inline comments for details.

pkg/skills/skillsvc/skillsvc.go
samuv: Good catch, agreed. Added the / count guard in both Install and GetContent:

```
if strings.ContainsAny(opts.Name, ":@") || strings.Count(opts.Name, "/") > 1 {
    return nil, ociErr
}
```
Also added a multi-segment OCI ref does not fall back to registry on pull failure test case to TestInstallQualifiedNameOCIFallback to pin the behaviour

pkg/skills/skillsvc/skillsvc.go
samuv: Done, added a `slog.Debug` before the fallback with the original error attached

pkg/skills/skillsvc/skillsvc.go
samuv: Good point, added a note to the PR description:

> Previously, a registry entry that existed but had no installable OCI or git package fell through and returned 404 Not Found. It now returns 422 Unprocessable Entity ("resolved but no installable package"), which is a more accurate response.

pkg/skills/gitresolver/writer.go
samuv: CodeQL: The alert needs to be dismissed manually in GitHub's Security tab as a false positive — the path comes from PathResolver.GetSkillPath, a trusted internal source. I've left a note to do that separately since it's outside the code changes.

Test coverage: Added TestWriteFiles/parent_directory_does_not_exist in writer_test.go — it points targetDir into a non-existent subdirectory and asserts that WriteFiles succeeds and creates the parent.

pkg/skills/skillsvc/skillsvc_test.go
samuv: All three gaps fixed:

- Digest variant: added digest ref (@sha256:...) does not fall back to registry on pull failure — uses a full 64-char hex digest so nameref.ParseReference accepts it, and asserts no SearchSkills call is made.

- Ambiguity error path: added registry ambiguity error surfaced to caller — SearchSkills returns two entries with the same namespace/name, causing resolveFromRegistry to return a 409 conflict, which the caller gets back directly.

- Pinned Pull args: replaced gomock.Any() on the third arg in tests 1, 2, and 4 with the expected reference string (e.g. "io.github.stacklok/my-skill:latest").

pkg/skills/skillsvc/skillsvc.go
```diff
@@ -235,11 +235,27 @@ func (s *service) Install(ctx context.Context, opts skills.InstallOptions) (*ski
...
+		result, ociErr := s.installFromOCI(ctx, opts, scope, ref)
+		if ociErr == nil {
+			return s.installAndRegister(ctx, result, opts.Group, opts.Name, scope, opts.ProjectRoot)
+		}
+		// OCI pull failed — fall back to registry lookup for names that look
+		// like a qualified "namespace/name" (no ':' or '@', exactly one '/').
+		// Names with an explicit tag or digest, or with more than one '/'
+		// (e.g. "ghcr.io/org/skill" or "ghcr.io/org/skill:v1"), are
+		// unambiguously OCI refs and must not trigger a registry search.
+		if strings.ContainsAny(opts.Name, ":@") || strings.Count(opts.Name, "/") > 1 {
```
JAORMX: There are functions to parse OCI references. Why not use those?

samuv: Partial reuse is possible. From the parsed name.Reference we can cleanly detect digests (type-assert to name.Digest) and multi-segment repos (ref.Context().RepositoryStr() slash count). What we can't recover from the parsed form is whether :latest was user-supplied or nameref-defaulted, foo/bar and foo/bar:latest both normalize to the same name.Tag. So we'd still need a string check for the tag case.

Happy to refactor to a hybrid (type-assert digest + slash-count on the parsed repo + string check for if you prefer. I kept it string-only originally because the three checks are equivalent in effect and keeping them together felt simpler to read. WDYT?

samuv: ok @JAORMX I went for the hybrid approach [here](https://github.com/stacklok/toolhive/pull/4841/changes/f528c88510881a357d89397892c403074df00344)

JAORMX (approved):

PR #4844, opened 2026-04-15
fix remote MCP servers stuck in starting when upstream returns errors

### LINKED ISSUE #4459: Remote MCP servers stuck in "starting" state when upstream returns errors
## Description

When a remote MCP server returns errors (e.g., HTTP 500), ToolHive keeps retrying the health check indefinitely without surfacing the failure to the user. The server remains stuck in `starting` status and the user gets no indication that something is wrong.

## Steps to reproduce

1. Run a remote MCP server that is returning errors (e.g., HTTP 500):
   ```
   thv run --name google-drive https://mcp.stacklok.dev/google-drive/mcp
   ```
2. Check server status:
   ```
   thv list -a
   ```
3. Observe the server stays in `starting` state indefinitely.

4. With debug logging enabled, the proxy logs show repeated failed attempts:
   ```
   level:DEBUG,msg:Server returned status,status_code:500,attempt:43
   ```

## Expected behavior

- ToolHive should detect that the remote server is consistently returning errors and transition the server to an **unhealthy** state.
- An error message should be surfaced to the user so they can act on it (e.g., check the remote server, retry later, etc.).

## Actual behavior

- The server stays in `starting` status forever.
- No error is shown to the user unless they manually inspect debug logs.
- `thv list` (without `-a`) doesn't even show the server since it never reaches `running`.

## Additional context

- This is a pre-existing issue, reproduced as far back as v0.12.0.
- The root cause in the initial investigation was the upstream MCP server itself being down (returning 500s), but ToolHive should handle this gracefully regardless.
- Consensus is to use the existing `unhealthy` state and surface a clear error message to the user.

slyt3:
When a remote server keeps returning HTTP 500, toolhive just leaves the workload stuck in
`starting` forever and never tells the user anything is wrong.

Turns out there were two bugs causing this:

1. Health checks for remote workloads were off by default — you had to set
   `TOOLHIVE_REMOTE_HEALTHCHECKS=true` manually, which most people would never do.
2. Even with that set, `monitorHealth` was skipping checks until the server finished
   initializing — but a server returning 500 never initializes, so it just skips forever.

So the workload sits in `starting` indefinitely with no way to know something is broken
unless you dig into debug logs.

What I changed:
- Removed the initialization guard so health checks always run
- Enabled health checks for remote workloads unconditionally, removed the env var gate
- After enough consecutive failures the workload goes to `unhealthy` and shows the reason in `thv list`
- Unhealthy remote workloads now show in `thv list` without needing `-a`

Fixes #4459

## Type of change

- [x] Bug fix

## Test plan

Added `TestTransparentProxy_RemoteServerFailure_HTTP500` — runs a proxy against a server
returning 500 without ever initializing it, confirms health check fires and proxy shuts down.
Passes.

## Changes

| File | Change |
|------|--------|
| `pkg/transport/proxy/transparent/transparent_proxy.go` | Remove `serverInitialized()` guard in `monitorHealth` |
| `pkg/transport/http.go` | Always enable health checks; remove dead `shouldEnableHealthCheck` helper |
| `pkg/workloads/manager.go` | Propagate `StatusContext`; show unhealthy without `-a` |
| `cmd/thv/app/list.go` | Print failure context in `thv list` status column |
| `pkg/transport/proxy/transparent/transparent_test.go` | Add regression test for HTTP 500 before initialization |
| `pkg/transport/http_test.go` | Update test to reflect always-enabled behavior |
| `test/e2e/health_check_zombie_test.go` | Remove no-op `TOOLHIVE_REMOTE_HEALTHCHECKS=true` |
| `test/e2e/stateless_proxy_test.go` | Remove no-op `TOOLHIVE_REMOTE_HEALTHCHECKS=true` |
| `docs/arch/03-transport-architecture.md` | Remove outdated env var docs |

Yes. Remote servers that keep failing now show as `unhealthy` in `thv list` with the reason,
instead of staying stuck in `starting` with nothing shown to the user.

## Large PR Justification
The line count is inflated by line-ending normalization in [transparent_test.go](vscode-file://vscode-app/c:/Users/manta.GITANA/AppData/Local/Programs/Microsoft%20VS%20Code/560a9dba96/resources/app/out/vs/code/electron-browser/workbench/workbench.html). Running gci on Windows wrote LF endings to a file that previously had CRLF endings, causing git to report ~1800 changed lines when the actual code changes are fewer than 50 lines.

The real content changes in this PR are:

Proxy stays alive after health check failures on remote servers (instead of shutting down)
Auto-recovery: status resets to running when the upstream server comes back
TOOLHIVE_REMOTE_HEALTHCHECKS env var removed (health checks always enabled)
Tests updated to assert stay-alive and recovery behavior
These changes cannot be split further without breaking the transport interface, the mock, and the tests simultaneously


pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -1258,13 +1258,6 @@ func (p *TransparentProxy) monitorHealth(parentCtx context.Context) {
...
-			if !p.serverInitialized() {
-				//nolint:gosec // G706: logging target URI from config
-				slog.Debug("mcp server not initialized yet, skipping health check",
-					"target", p.targetURI)
-				continue
-			}
-
 			alive := p.healthChecker.CheckHealth(parentCtx)
 			if alive.Status != healthcheck.StatusHealthy {
 				var shouldContinue bool
```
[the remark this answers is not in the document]
slyt3: the guard is now scoped to `!p.isRemote` local containers still skip health checks until initialized (no regression), remote workloads always run them since a server returning 500 never triggers setServerInitialized() anyway.

amirejaz:
## Review: Regression risk + proposed approach

The fix for the stuck-in-\`starting\` bug is correct and the removal of \`TOOLHIVE_REMOTE_HEALTHCHECKS\` is the right call. However, the current implementation introduces a regression for remote MCP servers with a **sleep/awake lifecycle** (e.g. scale-to-zero serverless backends): after ~45s of downtime (3 failures × 10s interval + 5s retry), the proxy self-terminates.

### Root cause of the regression

When \`handleHealthCheckFailure\` exhausts retries, it calls both \`onHealthCheckFailed()\` (sets status → \`unhealthy\`) **and** \`p.Stop()\` (kills the listener, exits \`monitorHealth\`). For a temporarily unavailable remote server there is nothing left to observe recovery — the proxy is dead.

### Proposed fix: soft-unhealthy with auto-recovery for remote workloads

The key insight: ToolHive does not control the lifecycle of a remote server the way it controls a container. Stopping the proxy because a remote server is temporarily unavailable is too aggressive.

**\`handleHealthCheckFailure\`** — branch on \`isRemote\`:

\`\`\`go
if p.onHealthCheckFailed != nil {
    p.onHealthCheckFailed()
}
if p.isRemote {
    // Stay in degraded mode, keep monitoring for recovery
    return consecutiveFailures, true  // shouldContinue = true, no Stop()
}
// Local container: we own the lifecycle, stop it
if err := p.Stop(ctx); err != nil { ... }
return consecutiveFailures, false
\`\`\`

**\`monitorHealth\`** — fire a recovery callback when health returns:

\`\`\`go
if consecutiveFailures >= healthCheckRetryCount && p.onHealthCheckRecovered != nil {
    p.onHealthCheckRecovered()
}
consecutiveFailures = 0
\`\`\`

**\`runner.go\`** — wire both callbacks:

\`\`\`go
transportHandler.SetOnHealthCheckFailed(func() {
    r.statusManager.SetWorkloadStatus(ctx, r.Config.BaseName, rt.WorkloadStatusUnhealthy, "Server stopped responding")
})
transportHandler.SetOnHealthCheckRecovered(func() {
    r.statusManager.SetWorkloadStatus(ctx, r.Config.BaseName, rt.WorkloadStatusRunning, "")
})
\`\`\`

**\`runner.go\`** — extend the existing unauthenticated status guard to also preserve \`unhealthy\`, preventing \`waitForInitializeSuccess\` from racing with the health check loop:

\`\`\`go
if currentWorkload.Status == rt.WorkloadStatusUnauthenticated ||
   currentWorkload.Status == rt.WorkloadStatusUnhealthy {
    // health checks drive this state, don't overwrite
} else {
    SetWorkloadStatus(running)
}
\`\`\`

### Resulting state machine for remote workloads

| Scenario | Behaviour |
|---|---|
| Server returns 500 from start (original bug) | Marked \`unhealthy\` in ~45s, stays there, visible in \`thv list\` |
| Server temporarily unavailable (e.g. scale-to-zero cold start) | Marked \`unhealthy\` during downtime, **auto-recovers** when it comes back — no manual restart |
| Permanently broken server | Stays \`unhealthy\` indefinitely; user sees it in \`thv list\` and runs \`thv stop\` manually |
| Local container workload | Unchanged — proxy stops on failure (ToolHive controls the container lifecycle) |

### Why \`TOOLHIVE_REMOTE_HEALTHCHECKS\` stays removed

The env var existed solely to prevent health checks from killing the proxy for temporarily unavailable servers. With auto-recovery, that reason is gone. Health checks are now safe to enable unconditionally for remote workloads: they catch broken servers fast and survive temporary downtime transparently.

amirejaz (changes requested):
added a comment

slyt3:
@amirejaz Yeah you right, stopping the proy was too aggresive, fixed it so remote proxies stay alive after fail and auto-recover when he serer come back, instead of dying and requiring a manual restrt, also wired recovery callback in runner.go and extended the atus guard to cover `unhealhty checks and removed `TOOLHIVE_REMOTE_HEALTHCEKS` the only reason it existed was to avoid the proxy ding on temporary outages, which now hanlded properly

aponcedeleonch (changes requested):
Thanks for digging into this one. I confirmed both root causes on main: monitorHealth skips the check forever while the server hasn't initialized (and a server stuck returning 500 never does), and remote health checks were gated behind an opt-in env var that defaults off. The fix targets the actual root cause, and the new regression tests (the HTTP 500 case and the auto-recovery cycle) pass for me. On the merits I'd take it.

What's blocking is diff hygiene, not correctness:
- A few test files have em dashes in existing comments corrupted into mojibake (looks like a Windows editor encoding thing), for example in backend_routing_test.go. Those need reverting.
- transparent_test.go has around 1800 lines of churn that's almost all mechanical blank-line insertions plus the same mojibake, with only about 5 real test changes buried in it. It's gofmt-clean so CI won't catch it, but it makes the diff really hard to review. Could you strip the churn so the diff shows just the real change? Re-running the formatter on a machine with UTF-8/LF settings should help.

Two non-blocking notes:
- During a sustained outage the failure callback fires and logs an error on every tick and re-writes the unhealthy status each time. Firing only on the transition into unhealthy would be quieter.
- Heads up for a maintainer: removing TOOLHIVE_REMOTE_HEALTHCHECKS makes remote health checks always-on, which reverses the earlier deliberate opt-in (the rationale there was avoiding traffic to remote servers). It's needed for the fix and matches the issue, but worth an explicit sign-off on the tradeoff.

Also, the PR body says the proxy shuts down on failure, but the actual behavior is stay-alive plus auto-recovery, and the Changes table is missing a few files. Worth tidying for the next reviewer. Once the whitespace and encoding are cleaned up, this is a clean merge.

slyt3:
@aponcedeleonch Hey, thanks for the review, fixed all three things:

Merged main in and resolved the conflict in manager.go
Fixed the mojibake em-dashes in backend_routing_test.go
Stripped the blank-line churn out of transparent_test.go so it only shows the real changes now 
Diff's a lot smaller now, should be easy to review. Let me know if I missed anything.

slyt3:
Also re: your two notes. The repeated logging/status-write on every failed tick during a sustained outage, didn't touch that yet, happy to fix it in a follow-up if you want it now instead.
Always-on remote health checks reversing the old opt-in, yeah, that's intentional, it's the whole point of the fix (silent starting forever otherwise). Flagging for whoever signs off on the tradeoff.

pkg/transport/proxy/transparent/transparent_proxy.go
```diff
@@ -1331,13 +1337,25 @@ func (p *TransparentProxy) handleHealthCheckFailure(
...
 	}
+
+	if p.isRemote {
+		// For remote workloads ToolHive does not own the server lifecycle.
+		// Stay in degraded mode and keep monitoring so auto-recovery is possible
+		// (e.g. scale-to-zero backends that come back after a cold start).
+		// Keep the counter at threshold so that monitorHealth can detect recovery
+		// (it fires onHealthCheckRecovered when consecutiveFailures >= threshold on
+		// the next successful check), then resets to 0 for the next outage window.
+		return p.healthCheckFailureThreshold, true
```
aponcedeleonch, line 1355: This return pins `consecutiveFailures` at the threshold while a remote server stays down. So every subsequent failing tick re-enters `handleHealthCheckFailure`, increments to threshold+1, skips the retry branch, and fires `onHealthCheckFailed()` again, along with the WARN at line 1327 and the ERROR at line 1342. For a server that's down for hours that means an ERROR log line plus a `SetWorkloadStatus(Unhealthy)` write every `healthCheckInterval`, indefinitely.

This is the repeated logging/status-write you flagged for a follow-up. I'd fix it here since it's small: make the failed notification edge-triggered, so `onHealthCheckFailed` fires only on the transition into the degraded state, not on every tick while degraded. Heads up that this is coupled with the runner race I flagged separately, they should be fixed together (edge-triggering this without closing that race turns a transient wrong-status window into a permanent one).

Minor: once pinned, the WARN at 1327 logs `attempt=6 max_attempts=5` on every tick, which reads oddly. Worth clamping or rewording if you keep any per-tick logging.

pkg/runner/runner.go
```diff
@@ -649,10 +664,13 @@ func (r *Runner) Run(ctx context.Context) error {
...
 
-	// Only set status to running if it's not already unauthenticated
-	// This preserves the unauthenticated state when bearer token authentication fails
-	if err == nil && currentWorkload.Status == rt.WorkloadStatusUnauthenticated {
-		slog.Debug("preserving unauthenticated status for workload", "workload", r.Config.BaseName)
+	// Only set status to running if it's not already unauthenticated or unhealthy.
+	// - unauthenticated: bearer token auth failed during init, preserve that state
+	// - unhealthy: health check fired before waitForInitializeSuccess returned,
+	//   the health check loop owns this state transition
+	if err == nil && (currentWorkload.Status == rt.WorkloadStatusUnauthenticated ||
```
aponcedeleonch, line 671: New concern, separate from the two notes last round. This read-then-write isn't atomic: `GetWorkload` at line 662 reads the status, then `SetWorkloadStatus(Running)` at line 675 writes it, while the health-monitor goroutine can fire `onHealthCheckFailed` -> `SetWorkloadStatus(Unhealthy)` in the window between them. When that happens, this branch clobbers `Unhealthy` back to `Running`.

Today it self-heals within one interval only because the failed callback re-fires every tick (see the `transparent_proxy.go` note). That's the coupling: if you make the callback edge-triggered without closing this race, the workload can get stuck at `running` for a server that's actually down.

Fix them together. Either re-read and only set `running` if the status is still startable, or gate this so it can't override a status the health-check loop owns.

pkg/workloads/manager.go
```diff
@@ -1824,12 +1824,16 @@ func (d *DefaultManager) getRemoteWorkloadsFromState(
...
+		// Unauthenticated are also surfaced so users see remote workloads
+		// that need attention without passing -a. Every other non-Running
+		// status is terminal until external intervention, and is hidden.
+		// Mirrors the filter in file_status.go:ListWorkloads.
 		if !listAll &&
 			workloadStatus.Status != rt.WorkloadStatusRunning &&
-			workloadStatus.Status != rt.WorkloadStatusAuthRetrying {
+			workloadStatus.Status != rt.WorkloadStatusAuthRetrying &&
+			workloadStatus.Status != rt.WorkloadStatusUnhealthy &&
+			workloadStatus.Status != rt.WorkloadStatusUnauthenticated {
```
aponcedeleonch, line 1836: This is what's failing CI (confirmed it's this PR, not a flake): `TestGetRemoteWorkloadsFromState_AuthRetryingVisibleWithoutAll` at `manager_test.go:2262` asserts an `Unauthenticated` remote workload stays hidden without `-a`, and this line now surfaces it.

Two things to sort out:

1. Was surfacing `Unauthenticated` intentional? The fix is about `Unhealthy`. If `Unauthenticated` is meant to show too, update the test and its doc comment (2185-2189). If it was scope creep, drop it and keep just `Unhealthy`.
2. The comment at line 1831 still says this "Mirrors the filter in file_status.go:ListWorkloads", but that filter (local path) still only surfaces `Running` + `AuthRetrying`, so the two have diverged. Either update `file_status.go` to match so local and remote behave the same, or reword the comment to say the remote path intentionally surfaces more statuses, and why.

Whichever way you go, the existing test needs updating so CI goes green.

aponcedeleonch (changes requested):
Thanks for turning this around. The stay-alive + auto-recovery design reads well and the diff is a lot tighter now.

Requesting changes mainly for the red CI, plus the two items worth locking down here rather than deferring. Details inline.

CI: `Tests / Test Go Code` is failing on `TestGetRemoteWorkloadsFromState_AuthRetryingVisibleWithoutAll`. I checked the job log, it's this PR and not a flake. The `manager.go` filter change surfaces `Unauthenticated` without `-a`, which that existing test asserts should stay hidden. See the inline note on `manager.go` for the two ways to resolve it.

On your two notes from last round:
- Repeated logging/status-write per failed tick: I'd rather fix it here than in a follow-up. It's a small change and the current behavior emits an ERROR log line and a status-write every interval for as long as a remote server is down. Inline on `transparent_proxy.go`.
- Always-on remote health checks reversing the old opt-in: agreed, that's the point of the fix. No concern from me there.

One new thing that wasn't in the earlier round: there's a TOCTOU race in `runner.go` between the health-check callback and the final `SetWorkloadStatus(Running)`. It's coupled to the logging fix, so worth reading the two inline notes together.

PR #4845, opened 2026-04-15
Return error from `UpdateConfig` callbacks

blkt:
Change the `UpdateConfig` callback signature from `func(*Config)` to `func(*Config) error` so that future callers can propagate validation or transformation errors from inside the closure instead of relying on side-channel error handling.

All existing closures return nil. `UpdateConfigAtPath` now short-circuits before writing to disk when the callback returns an error.

### DIFF
--- a/cmd/thv/app/client.go
+++ b/cmd/thv/app/client.go
@@ -292,15 +292,16 @@ func registerClientsGlobally(
 ) error {
 	for _, clientToRegister := range clients {
 		// Update the global config to register the client
-		err := config.UpdateConfig(func(c *config.Config) {
+		err := config.UpdateConfig(func(c *config.Config) error {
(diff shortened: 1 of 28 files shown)


blkt:
This change allows more complex scenarios where edit operations might be required to fail under some circumstances.

jhrozek (approved):
Looks good. Clean mechanical refactoring — the callback signature change is well-motivated, the short-circuit before disk write is correct, and all callers are consistently updated.

Two minor observations (non-blocking):
- No test exercises the new error path (callback returns error → config not written to disk). Worth adding to protect the invariant.
- The `"update function failed: %w"` wrapping may double-nest with caller wrapping when callbacks start returning real errors. Could simplify to a bare return.

PR #4849, opened 2026-04-15
Graduate CRDs from v1alpha1 to v1beta1

### LINKED ISSUE #2556: Graduate CRDs from v1alpha1 to v1beta1
## Summary

Promote all ToolHive CRDs from `v1alpha1` to `v1beta1` to signal API stability. This is a **clean break** — no conversion webhook, no multi-version serving. `v1beta1` becomes the sole storage and served version immediately. Users must recreate resources with the new `apiVersion`.

## Strategy

This is a purely mechanical rename — no logic changes, no field removals, no behavioral changes. Every file change is one of:

1. **Directory rename**: `api/v1alpha1/` → `api/v1beta1/`, package name update
2. **Import path update**: `cmd/thv-operator/api/v1alpha1` → `cmd/thv-operator/api/v1beta1`, alias `mcpv1alpha1` → `mcpv1beta1`
3. **String literal update**: `toolhive.stacklok.dev/v1alpha1` → `toolhive.stacklok.dev/v1beta1` in YAML, Markdown, and Go

## Blast Radius

| Category | Files | Change Type |
|----------|------:|-------------|
| API type definitions (`api/v1alpha1/`) | 20 | Directory rename + package name |
| Controllers (`controllers/`) | 59 | Import path + alias |
| Support packages (`cmd/thv-operator/pkg/`) | 47 | Import path + alias |
| Operator main (`cmd/thv-operator/main.go`) | 1 | Import path + alias |
| Integration tests (`test-integration/`) | 43 | Import path + alias |
| Shared packages (`pkg/`) | ~28 | Import path + alias |
| E2E Go tests (`test/e2e/`) | 33 | Import path + alias |
| Chainsaw YAML fixtures (`test/e2e/chainsaw/`) | 43 | `apiVersion` string |
| Example YAML files (`examples/operator/`) | ~47 | `apiVersion` string |
| Other YAML (`deploy/keycloak/`, root) | 6 | `apiVersion` string |
| CRD manifests (Helm charts) | 25 | Regenerated |
| Webhook manifests (`config/webhook/`) | 1 | Regenerated |
| Documentation (`.md` files) | 26 | String references |
| Swagger/API docs (`docs/server/`) | 3 | Regenerated |
| **Total** | **~382** | |

### Stage 1: Directory Rename + Package Update

```bash
mv cmd/thv-operator/api/v1alpha1 cmd/thv-operator/api/v1beta1
# Update package v1alpha1 → package v1beta1 in all files under api/v1beta1/
# Update GroupVersion in groupversion_info.go: Version: "v1alpha1" → "v1beta1"
```

### Stage 2: Global Find-Replace (scripted)

```bash
# Go import paths
find . -name '*.go' -not -path './.git/*' -not -path './.claude/worktrees/*' \
  -exec sed -i '' 's|cmd/thv-operator/api/v1alpha1|cmd/thv-operator/api/v1beta1|g' {} +

# Go import alias
find . -name '*.go' -not -path './.git/*' -not -path './.claude/worktrees/*' \
  -exec sed -i '' 's|mcpv1alpha1|mcpv1beta1|g' {} +

# YAML apiVersion
find . \( -name '*.yaml' -o -name '*.yml' \) -not -path './.git/*' -not -path './.claude/worktrees/*' \
  -exec sed -i '' 's|toolhive.stacklok.dev/v1alpha1|toolhive.stacklok.dev/v1beta1|g' {} +

# Markdown and docs
find . -name '*.md' -not -path './.git/*' -not -path './.claude/worktrees/*' \
  -exec sed -i '' 's|toolhive.stacklok.dev/v1alpha1|toolhive.stacklok.dev/v1beta1|g' {} +
find . -name '*.md' -not -path './.git/*' -not -path './.claude/worktrees/*' \
  -exec sed -i '' 's|api/v1alpha1|api/v1beta1|g' {} +

# Hardcoded string in pkg/export/k8s.go
sed -i '' 's|"toolhive.stacklok.dev/v1alpha1"|"toolhive.stacklok.dev/v1beta1"|g' pkg/export/k8s.go
```

### Stage 3: Manual Fixups

- Review `groupversion_info.go` package doc comment
- Review `pkg/export/k8s.go` and its test for hardcoded strings
- Fix any compilation errors

### Stage 4: Regenerate Generated Code

```bash
task operator-generate          # deepcopy
task operator-manifests         # CRD YAML, RBAC, webhook manifests
task docs                       # CLI and API docs (swagger)
task -d cmd/thv-operator crdref-gen  # CRD API reference docs
# Copy/verify Helm chart CRDs in deploy/charts/operator-crds/
```

### Stage 5: Verification

```bash
# Check for straggler references
grep -r "v1alpha1" . --include="*.go" --include="*.yaml" --include="*.yml" --include="*.md" \
  | grep -v ".git/" | grep -v ".claude/worktrees/" | grep -v "vendor/" \
  | grep "toolhive"
# Should return zero results

task lint-fix
task build
task test
task operator-test
```

## Key Files Requiring Manual Attention

| File | Why |
|------|-----|
| `cmd/thv-operator/api/v1beta1/groupversion_info.go` | GroupVersion constant + package doc |
| `cmd/thv-operator/main.go` | Import alias + all `mcpv1alpha1.` type references |
| `pkg/export/k8s.go:65` | Hardcoded `"toolhive.stacklok.dev/v1alpha1"` string |
| `pkg/export/k8s_test.go` | Test assertion for API version |
| `config/webhook/manifests.yaml` | Webhook paths include version (regenerated) |
| `docs/server/` | Generated swagger docs (regenerated) |

## PR Strategy

**Single PR** — This is a mechanical rename touching ~382 files with zero logic changes. Splitting would create merge conflicts (overlapping import blocks) and partial renames are harder to review. The PR will be large by line count but trivially reviewable since every change follows the same pattern.

Exception to the 400-line / 10-file PR rule: mechanical rename, same category as dependency updates and generated code.

## Pitfalls

- **Worktrees**: `.claude/worktrees/` contains stale repo copies — exclude from find-replace
- **Generated files**: Don't manually edit `zz_generated.deepcopy.go` or CRD YAML — use `task operator-generate` and `task operator-manifests`
- **Helm chart CRDs**: Files in `deploy/charts/operator-crds/` must be updated (verify Taskfile copies them or do manually)
- **Webhook paths**: Paths like `/validate-toolhive-stacklok-dev-v1alpha1-*` are regenerated from kubebuilder markers
- **No partial state**: This is all-or-nothing — half v1alpha1 / half v1beta1 won't compile
- **In-flight PRs**: Coordinate timing — land when the operator PR queue is clear to minimize conflicts

## Out of Scope (separate work)

- ~~CI job to warn on backwards-incompatible CRD changes~~ (deferred to follow-up)
- Removing deprecated fields (`OIDCConfig`, `Telemetry`, `UseLegacyAttributes`, `DeprecatedBackendAuthType`)
- Adding `EnvVar.ValueFrom` (#4547)
- CLI-side `UseLegacyAttributes` removal
- Conversion webhooks or multi-version serving infrastructure

ChrisJBurns:
## Summary

Signal API stability by promoting all ToolHive CRDs from `v1alpha1` to `v1beta1` — without breaking existing users. Originally scoped as a clean break (delete and recreate all resources), the PR now serves **both versions simultaneously** so operators can upgrade with zero downtime and migrate manifests at their own pace.

The core insight is that `v1alpha1` and `v1beta1` are **schema-identical** — only the version string differs. Rather than duplicate every field across both packages, the new `cmd/thv-operator/api/v1alpha1/` package defines only the root resource types (e.g. `MCPServer`, `MCPGroup`) as thin wrappers whose `Spec` and `Status` fields reference the canonical types from `v1beta1` (e.g. `Spec v1beta1.MCPServerSpec`). Because controller-gen walks field types when building the OpenAPI schema, both versions resolve to the exact same generated schema without any duplication of the actual model. The "migration" is then orchestrated entirely through two flags on each CRD version entry: `served` and `storage`. The new CRD manifest lists **both** versions as `served: true` — so existing `v1alpha1` clients keep working — but marks only `v1beta1` as `storage: true`, meaning new writes go into etcd at `v1beta1`. `v1alpha1` additionally carries `deprecated: true` with a `deprecationWarning`, which is what makes `kubectl` print the migration nudge on every access. Combined with `conversion.strategy: None` (which is safe precisely because the schemas are identical — Kubernetes just swaps the `apiVersion` string when serving an object at a different version than it was stored at), users can upgrade the CRD chart without deleting anything, keep reading and writing at `v1alpha1` indefinitely, and migrate to `v1beta1` at their own pace by simply re-applying their manifests. Once all objects have been re-stored at `v1beta1` (tracked via the CRD's `status.storedVersions`), a future release can drop the `v1alpha1` entry from the `versions` list and remove the wrapper package — completing the graduation with zero downtime at any point in the process.

Closes #2556

<details>
<summary><strong>What changed across commits</strong></summary>

1. **Mechanical rename** (earlier commits on this branch): `cmd/thv-operator/api/v1alpha1/` → `v1beta1/`, all Go import paths and aliases, YAML `apiVersion` strings in examples/chainsaw fixtures/deploy manifests, hardcoded version strings in `pkg/export/k8s.go`, and every generated artifact (deepcopy, CRD manifests, webhook manifests, Helm chart CRDs, swagger docs, CRD API reference docs).
2. **Regenerate stale audit deepcopy** — drive-by: `pkg/audit.Config.DetectApplicationErrors` was added without re-running `controller-gen`; this commit catches it up.
3. **Add `v1alpha1` wrapper types** — new minimal package: 12 root types + 12 list types (~380 hand-written lines), each reusing `v1beta1` `Spec`/`Status`. Each root type carries `+kubebuilder:deprecatedversion:warning` so `kubectl` prints a migration hint. `main.go` registers both schemes.
4. **Regenerate CRDs to serve both versions** — output of `task operator-manifests`. Every CRD now lists `v1alpha1` (served, non-storage, deprecated) and `v1beta1` (served, storage). No conversion webhook (defaults to `strategy: None`), which is safe because the schemas are identical.
5. **Remove obsolete pre-upgrade check** — the Helm hook that blocked upgrades when `v1alpha1` resources were present is no longer needed, since those resources now survive the upgrade untouched.

</details>

## Type of change

- [x] Other (describe): CRD graduation from `v1alpha1` to `v1beta1` via multi-version serving

## Test plan

- [x] Build verification (`task build` — operator binary + all Go packages compile cleanly after the rebase on `main`)
- [x] `task operator-generate` and `task operator-manifests` re-run cleanly and produce the multi-version CRDs
- [x] **End-to-end upgrade test** (local): installed v0.21.0 CRDs + operator from the OCI registry, created one CR of each of the 12 CRD kinds, upgraded the CRDs chart to this branch, rolled out the new operator via `task operator-deploy-local`, verified every Deployment UID was unchanged (zero-downtime), then re-applied the resources at `v1beta1` and confirmed still no restarts and `status.storedVersions` advanced to include `v1beta1`. 16 of 16 assertions passed.

Yes, but **non-breaking**. Existing `v1alpha1` resources continue to work after the upgrade; no deletion or recreation is required. Users will see a deprecation warning from `kubectl` on every access to a `v1alpha1` resource, and should migrate their manifests to `apiVersion: toolhive.stacklok.dev/v1beta1` at their convenience. `v1alpha1` will be removed in a future release once all stored objects have migrated.

This PR is large by line count but most of the volume is either generated output (`controller-gen` CRD manifests, `zz_generated.deepcopy.go`) or the original mechanical rename. Worth reviewing carefully:

- `cmd/thv-operator/api/v1alpha1/types.go` — the only non-trivial hand-written file. Verify the `Spec`/`Status` field references point to `v1beta1` and that the `+kubebuilder:deprecatedversion` and other markers match `v1beta1`'s (minus `+kubebuilder:storageversion`, which must be on exactly one version).
- `cmd/thv-operator/main.go` — confirms both schemes are registered.
- One sample CRD (e.g. `deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpgroups.yaml`) — confirm the structure: both versions listed, `v1beta1` marked `storage: true`, `v1alpha1` marked `deprecated: true` with a warning and `storage: false`, and the `openAPIV3Schema` sections under both versions are identical.

The remaining 23 CRD files follow the same pattern.

## Large PR Justification
- no way to split this PR out
- 


design-crd-graduation-v1beta1.md
```diff
@@ -0,0 +1,212 @@
...
+- Date: 2026-04-15
+
+## Goal
+
+Promote all ToolHive CRDs from `v1alpha1` to `v1beta1` to signal API stability and introduce rigor around future breaking changes. This is a **clean break** — no conversion webhook, no multi-version serving. `v1beta1` becomes the sole storage and served version. Users must recreate resources with the new `apiVersion`.
+
+## Scope
+
+- **In**:
+  - Rename `cmd/thv-operator/api/v1beta1/` → `cmd/thv-operator/api/v1beta1/`
```
ChrisJBurns: The global sed replaced `v1alpha1` with `v1beta1` inside this design doc too, making the "from → to" descriptions nonsensical. Multiple lines are affected:

- Line 14: `v1beta1/` → `v1beta1/` (should be `v1alpha1/` → `v1beta1/`)
- Line 17: `toolhive.stacklok.dev/v1beta1` → `toolhive.stacklok.dev/v1beta1`
- Lines 34-36: Strategy section repeats the same issue
- Lines 90, 105, 109, 113, 117, 119, 122: The sed commands reference `v1beta1` → `v1beta1`

All the "source" sides of the arrows should still say `v1alpha1` so readers can understand what was renamed to what.

design-crd-graduation-v1beta1.md
```diff
@@ -0,0 +1,212 @@
+# Design: Graduate CRDs from v1alpha1 to v1beta1
```
ChrisJBurns: **Question:** Should this design doc be committed to the repo root? It reads as a planning artifact (implementation steps, sed commands, etc.). Options:
1. Remove from the PR entirely (it served its purpose as a planning doc)
2. Move to `docs/arch/` if it's meant as reference documentation for the graduation decision
3. Keep as-is if this is intentional

I'd lean toward option 1 or 2 — repo root isn't the right home for design docs.

ChrisJBurns (commented):
## Review: Graduate CRDs from v1alpha1 to v1beta1

**Verified:**
- Core `GroupVersion` constant correctly changed to `v1beta1`
- All Go imports and aliases consistently updated (`mcpv1alpha1` → `mcpv1beta1`)
- Hardcoded `apiVersion` string in `pkg/export/k8s.go` updated
- Webhook manifests (both copies) correctly reference `v1beta1` paths and apiVersions
- Helm chart CRDs, swagger docs, and skill docs updated
- Old `v1alpha1/` directory fully removed — zero straggling `v1alpha1` references in production code
- All 377 files show symmetric add/delete counts, confirming purely mechanical changes

**Two items to address** — both in the design doc, not in production code.

blkt:
I checked the code with Claude and it reported that there are two more files to fix, `PROJECT` and `.claude/agents/kubernetes-expert.md`.
The `PROJECT` file seems to be the more important one since Kubebuilder tooling reads it, while the `.claude/agents/kubernetes-expert.md` guidance is lower priority but should be corrected to avoid confusing the agent.

JAORMX:
So, graduating CRDs from alpha to beta is one of those tasks where there isn't really a nice way to do it. Either you carry a conversion webhook and multi-version serving around forever (complex, but operationally smooth), or you do a clean break and ask folks to recreate their resources (simpler, but you better make sure the upgrade path is airtight). This PR goes with the clean break, which is a totally valid call for an alpha API... but the upgrade path isn't airtight yet.

First, the good news: the rename itself is clean. Went through a detailed review and couldn't find meaningful gaps in the mechanical side. All 18 CEL `XValidation` rules are byte-identical between `main` and this PR, every `+kubebuilder:default=` marker survived, RBAC is unchanged, `cmd/thv-operator/config/webhook/manifests.yaml` is consistently `v1beta1`, and there are no leftover `mcpv1alpha1` aliases or orphan `api/v1alpha1/` directory. Import rename was done surgically. Nice work!

So... what's missing?

### 1. The `PROJECT` file is stale

`PROJECT:14-15` still declares `path: .../api/v1alpha1` and `version: v1alpha1`. Next time someone runs `kubebuilder create api`, it scaffolds into a ghost directory.

### 2. Doc generation is still scoped to v1alpha1

`task docs` will produce broken output after merge:

- `docs/operator/crd-ref-config.yaml:14-15` points at `api/v1alpha1`
- `docs/operator/templates/markdown/type.tpl:47,76` filters on `hasSuffix "/api/v1alpha1"`
- `docs/operator/virtualmcpserver-api.md:16` still reads `**Version**: v1alpha1` (stale generated output, fix the template and regenerate)

### 3. No `+kubebuilder:storageversion` markers

None of the `cmd/thv-operator/api/v1beta1/*_types.go` files declare a storage version. Works today because there's only one version, but the moment v1beta2 or v1 lands we have a subtle apiserver error waiting. Cheap to add now.

### 4. Orphan-resource blocker on upgrade

This is the one that concerns me the most. The CRD drops `v1alpha1` entirely (no served, no stored, no conversion strategy). Any existing `v1alpha1` CR in etcd becomes undecodable after the chart upgrade, the operator can't list or finalize them, and their finalizers block namespace deletion. Folks with `MCPExternalAuthConfig` or `MCPOIDCConfig` objects holding credential references will be especially unhappy.

Two reasonable mitigations, pick one:

- Keep `v1alpha1` in the CRD as `served: false` with a `None` conversion strategy for a release, paired with a deprecation warning. Gives users one cycle to drain.
- Ship a Helm `pre-upgrade` hook Job that lists any pre-existing `v1alpha1` CRs and fails fast with a "drain these first, here's how" message. Can be opt-out via a values flag for folks who really want to push through.

---

None of this blocks the spirit of the graduation, and the rename itself is in good shape. It's really about making the upgrade safe for folks running ToolHive today. Happy to help wire up the pre-upgrade hook or the migration doc if it's useful!

ChrisJBurns:
Yeah it looks like rebasing with main has caused new files to be added with `v1alpha1` as they weren't there before and CC didn't seem to remember the task at hand and change them to `v1beta1`. Shall fix

deploy/charts/operator-crds/templates/pre-upgrade-check.yaml
```diff
@@ -0,0 +1,166 @@
...
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/component: pre-upgrade
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ .Release.Name }}-pre-upgrade
+      containers:
+        - name: check
+          image: bitnami/kubectl:latest
```
jhrozek: Is `bitnami/kubectl:latest` still a safe choice here? Bitnami moved its free public image catalog to `bitnamilegacy` in August 2025, and `docker.io/bitnami/kubectl:latest` now serves a deprecation placeholder for many users — the Job would fail to pull and Helm would report a generic hook-failed error with no remediation guidance. Using `:latest` on a pre-upgrade hook that can block cluster upgrades also means a future registry change can silently alter hook semantics.

Could we pin to a stable source and expose it via `.Values.upgradeCheck.image.{repository,tag,digest}`? Options: `registry.k8s.io/kubectl:v1.31.x` (official), `cgr.dev/chainguard/kubectl`, or `bitnamilegacy/kubectl:<digest>`.

deploy/charts/operator-crds/templates/pre-upgrade-check.yaml
```diff
@@ -0,0 +1,166 @@
...
+              if [ -z "$CRDS" ]; then
+                echo "No existing ToolHive CRDs found. Fresh install — nothing to check."
+                exit 0
+              fi
+
+              BLOCKED=""
+              for CRD in $CRDS; do
+                # Check if storedVersions contains v1alpha1
+                STORED=$(kubectl get crd "$CRD" -o jsonpath='{.status.storedVersions}' 2>/dev/null || true)
+                if echo "$STORED" | grep -q "v1alpha1"; then
```
jhrozek: nit (from an agent review — no human would have spotted this): `grep -q "v1alpha1"` on the serialized `{.status.storedVersions}` array does a bare substring match, so a future version like `v1alpha10` or `v1alpha1beta` would also match and cause the hook to block upgrades for no real reason. Same pattern at line 139. Want to harden this with exact membership, e.g. `kubectl get crd "$CRD" -o json | jq -e '.status.storedVersions | index("v1alpha1")'`, or at minimum `jsonpath='{.status.storedVersions[*]}' | grep -qw v1alpha1`?

deploy/charts/operator-crds/templates/pre-upgrade-check.yaml
```diff
@@ -0,0 +1,166 @@
...
+        app.kubernetes.io/component: pre-upgrade
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ .Release.Name }}-pre-upgrade
+      containers:
+        - name: check
+          image: bitnami/kubectl:latest
+          command:
+            - /bin/sh
+            - -ec
```
jhrozek: Should this Pod have a `securityContext` and `resources` set? On clusters enforcing PodSecurity admission at `restricted` level, a Pod without `runAsNonRoot: true`, dropped `ALL` capabilities, a `seccompProfile`, and `allowPrivilegeEscalation: false` is rejected at admission — which would make the hook fail to create the Job and block every Helm upgrade on those clusters. Same concern for clusters with a LimitRange that requires resource requests/limits. Would a restricted-baseline securityContext + minimal `resources.limits` (e.g. `100m` CPU / `128Mi` memory) make sense here?

test/e2e/upgrade/test-pre-upgrade-hook.sh
```diff
@@ -0,0 +1,417 @@
...
+  pass "Pre-upgrade hook was skipped (no Job created)"
+else
+  fail "Pre-upgrade hook ran despite upgradeCheck.enabled=false (${JOB_COUNT} jobs found)"
+fi
+
+if [ "$UPGRADE_RC" -ne 0 ]; then
+  log "Helm upgrade failed as expected (storedVersions validation) — the hook was bypassed but Kubernetes rejected the CRD update"
+  pass "Opt-out correctly skips only the hook, not the Kubernetes validation"
+else
+  warn "Helm upgrade succeeded with opt-out — this may happen if storedVersions was already patched"
```
jhrozek: Should this `warn` be a `fail`? Scenario 4's stated purpose is to prove that opt-out skips only the hook (not the Kubernetes storedVersions validation). If the upgrade succeeds on this branch, that contract is falsified — but `warn` doesn't increment `FAIL`, so the scenario reports PASS either way.

docs/operator/virtualmcpserver-observability.md
```diff
@@ -204,7 +204,7 @@ shared MCPTelemetryConfig pattern.
 
 ```yaml
 # Deprecated — use telemetryConfigRef instead
-apiVersion: toolhive.stacklok.dev/v1alpha1
+apiVersion: toolhive.stacklok.dev/v1beta1
```
jhrozek, line 207: Referencing the sentence a few lines above (around line 201): *"The inline `spec.config.telemetry` field still works but is deprecated and will be removed in v1beta1."* Now that v1beta1 is the storage version (per this PR) and the example immediately below still uses `v1beta1` with the inline field, the sentence is self-contradictory. Could we retarget the removal to "a future API version" (or "v1", if that's the plan)?

jhrozek (commented):
Review from an agent-assisted pass over the v1alpha1 → v1beta1 graduation. The mechanical rename looks clean and the storage-version markers are in place on all 12 types. A few questions on the new pre-upgrade hook and test harness, plus one stale doc line — inline. Nothing blocking from my side.

docs/operator/virtualmcpserver-observability.md
ChrisJBurns: Good catch — thanks. This was a broader docs staleness: the "removed in v1beta1" phrase appeared in 7 places (2 docs + 5 example YAML files) and all of them were contradicted by this PR now that `v1beta1` is the storage version and the deprecated inline fields are being kept. Fixed by retargeting all occurrences to "removed in a future API version" — that matches the multi-version strategy (a future release can drop `v1alpha1` and simultaneously remove these inline fields from `v1beta1` in a follow-up graduation).

Files updated:
- `docs/observability.md`
- `docs/operator/virtualmcpserver-observability.md`
- `examples/operator/virtual-mcps/vmcp_with_oidcconfig_ref.yaml`
- `examples/operator/virtual-mcps/vmcp_with_telemetry_ref.yaml`
- `examples/operator/virtual-mcps/vmcp_inline_incoming_auth.yaml`
- `examples/operator/mcp-servers/mcpremoteproxy_with_oidcconfig_ref.yaml`
- `examples/operator/mcp-servers/mcpserver_with_oidcconfig_ref.yaml`

Amended into the squashed commit (f17269c1).

JAORMX (dismissed):

cmd/thv-operator/api/v1alpha1/types.go
```diff
@@ -0,0 +1,385 @@
...
+//+kubebuilder:printcolumn:name="References",type=string,JSONPath=`.status.referencingWorkloads`
+//+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// MCPExternalAuthConfig is the deprecated v1alpha1 version of the MCPExternalAuthConfig resource.
+type MCPExternalAuthConfig struct {
+	metav1.TypeMeta   `json:",inline"` // nolint:revive
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   v1beta1.MCPExternalAuthConfigSpec   `json:"spec,omitempty"`
+	Status v1beta1.MCPExternalAuthConfigStatus `json:"status,omitempty"`
```
jhrozek, line 59: If anyone will read the PR later, this is the trick - the types are imported from beta into alpha so when controller-gen generates the schema, it walks the Go structures recursively, but the types are from beta, so the actual shape of the API ends up being identical.

cmd/thv-operator/api/v1beta1/mcpserver_types.go
```diff
@@ -867,6 +867,7 @@ const (
 )
 
 //+kubebuilder:object:root=true
+//+kubebuilder:storageversion
```
jhrozek, line 870: The diff shows this nicely, but for book-keeping and future (human or AI) readers: the only ways in which the `kubebuilder` annotations differ seem to be storageversion that's only present on beta APIs and deprecation on the alpha APIs. That makes sense.

jhrozek (dismissed):
I haven't actually tested the migration (I know you asked for testers), but I did spend a fair bit of time reviewing the PR in an interactive session with CC - mostly because I wasn't familiar with this serve-old-store-new approach. It seems clever and CC called it "a textbook implementation". I really could poke any holes it in, I like how the diff, if you exclude the mechanical replaces, is actually minimal and you can reason about the important things (I did review the annotations myself, old skool). The

I'm going to approve the patch, I guess the longer we hold off the harder it gets for you to merge. Let's chat if you want/need more testing.

jhrozek (dismissed):

rdimitrov (dismissed):

jhrozek (approved):

PR #4857, opened 2026-04-15
Add policy_stopped workload status and enforce policy gate on restart

reyortiz3:
## Summary

- Policy enforcement currently only gates workload _creation_ — running workloads are unaffected when a policy activates, and restart paths have no policy check at all. This PR adds the upstream groundwork to enforce the `non_registry_servers` policy on existing workloads.
- Adds `WorkloadStatusPolicyStopped` as a first-class status (following the precedent of `WorkloadStatusUnauthenticated`) so the policy layer can set it when stopping a non-compliant workload, giving the UI and CLI a programmatic signal rather than freeform text.
- Calls `EagerCheckCreateServer` in `restartSingleWorkload` so the policy gate blocks any restart attempt for workloads that violate the active policy.

## Changes

| File | Change |
|------|--------|
| `pkg/container/runtime/types.go` | Add `WorkloadStatusPolicyStopped` constant |
| `pkg/core/workload.go` | Add `policy_stopped` to `enums` tag for swagger |
| `pkg/workloads/manager.go` | Map `policy_stopped` → `BackendUnhealthy`; call `EagerCheckCreateServer` in restart path |
| `cmd/thv/app/list.go`, `status.go` | Add 🚫 indicator for `policy_stopped` in CLI output |
| `docs/server/swagger.*`, `docs.go` | Regenerated to include new enum value |

## Type of change

## Test plan

- [x] `task build` passes
- [x] `task test` passes (pre-existing failures in `pkg/client` and `pkg/telemetry` are unrelated to this change and reproduce on `main`)
- [x] Swagger enum updated and regenerated via `task docs`

CLI `list` and `status` commands will show `🚫 policy_stopped` for workloads stopped by policy, making the reason immediately visible rather than showing a plain `stopped` or `error`.


reyortiz3:
@claude can you verify if the comments were addressed?

reyortiz3:
@claude can you review if the comments have been addressed?

rdimitrov (dismissed):

pkg/workloads/manager.go
```diff
@@ -1086,6 +1088,12 @@ func (d *DefaultManager) restartSingleWorkload(ctx context.Context, name string,
 		return d.restartContainerWorkload(ctx, name, foreground)
 	}
 
+	// Check policy gates before restarting — the loaded RunConfig carries the same
+	// fields (RegistryAPIURL, RegistryURL, RemoteURL) that the gate evaluates on create.
+	if err := runner.EagerCheckCreateServer(ctx, runConfig); err != nil {
+		return fmt.Errorf("server restart blocked by policy: %w", err)
```
jhrozek, line 1097: Both policy gate check sites return an error here but never call `SetWorkloadStatus(ctx, name, rt.WorkloadStatusPolicyStopped, ...)`. The constant is defined, the CLI renders the indicator, and the swagger docs list it, but nothing ever transitions a workload into this status. A blocked restart just leaves the workload as `stopped`.

Is this intentional -- e.g. the operator or some downstream consumer will set this? If so it might be worth a comment saying so. Otherwise the display code and health mapping for `policy_stopped` are unreachable.

pkg/workloads/manager.go
```diff
@@ -243,6 +243,8 @@ func mapWorkloadStatusToVMCPHealth(status rt.WorkloadStatus) vmcp.BackendHealthS
 		return vmcp.BackendUnknown
 	case rt.WorkloadStatusUnauthenticated:
 		return vmcp.BackendUnauthenticated
+	case rt.WorkloadStatusPolicyStopped:
+		return vmcp.BackendUnhealthy
 	default:
```
jhrozek, line 248: Mapping `PolicyStopped` to `BackendUnhealthy` means vMCP can't distinguish "crashed" from "blocked by policy". Compare with `Unauthenticated` which gets its own `BackendUnauthenticated` variant. Is the generic mapping intentional here, or would a dedicated `BackendPolicyStopped` be worth adding for parity?

cmd/thv/app/list.go
```diff
@@ -165,10 +165,17 @@ func printTextOutput(workloadList []core.Workload) {
...
+		// Highlight unauthenticated and policy-stopped workloads with indicators
 		status := string(c.Status)
-		if c.Status == rt.WorkloadStatusUnauthenticated {
+		switch c.Status {
+		case rt.WorkloadStatusUnauthenticated:
 			status = "⚠️  " + status
+		case rt.WorkloadStatusPolicyStopped:
+			status = "🚫 " + status
+		case rt.WorkloadStatusRunning, rt.WorkloadStatusStopped, rt.WorkloadStatusError,
+			rt.WorkloadStatusStarting, rt.WorkloadStatusStopping, rt.WorkloadStatusUnhealthy,
```
jhrozek: Nit: this switch block is duplicated verbatim in `status.go`. Might be worth extracting a small helper in `common.go` so future status indicators only need updating in one place.

pkg/workloads/manager.go
reyortiz3: I noticed this, but it's the same pattern as above, for example:

```
case rt.WorkloadStatusStopped, rt.WorkloadStatusError, rt.WorkloadStatusStopping, rt.WorkloadStatusRemoving:
		return vmcp.BackendUnhealthy
```

docs/server/docs.go
```diff
@@ -942,7 +944,8 @@ const docTemplate = `{
                     "WorkloadStatusUnhealthy",
                     "WorkloadStatusRemoving",
                     "WorkloadStatusUnknown",
-                    "WorkloadStatusUnauthenticated"
+                    "WorkloadStatusUnauthenticated",
+                    "WorkloadStatusPolicyStopped"
                 ]
             },
             "github_com_stacklok_toolhive_pkg_container_templates.RuntimeConfig": {
```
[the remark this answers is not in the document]
reyortiz3: Fixed — `policy_stopped` is already in the `enums` tag at [pkg/api/v1/workload_types.go:35](pkg/api/v1/workload_types.go#L35). The swagger docs were regenerated as part of the same commit.

pkg/workloads/manager.go
```diff
@@ -1086,6 +1088,12 @@ func (d *DefaultManager) restartSingleWorkload(ctx context.Context, name string,
...
 
+	// Check policy gates before restarting — the loaded RunConfig carries the same
+	// fields (RegistryAPIURL, RegistryURL, RemoteURL) that the gate evaluates on create.
+	if err := runner.EagerCheckCreateServer(ctx, runConfig); err != nil {
+		return fmt.Errorf("server restart blocked by policy: %w", err)
+	}
+
 	// Check if this is a remote workload
 	if runConfig.RemoteURL != "" {
 		return d.restartRemoteWorkload(ctx, name, runConfig, foreground)
```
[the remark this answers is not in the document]
reyortiz3: Fixed — `EagerCheckCreateServer` is now called inside `maybeSetupContainerWorkload` (after `loadRunnerFromState`) at line 1306-1311, so it runs regardless of whether the outer `LoadState` succeeded. The canonical workload name is resolved from container labels before the check, so it always evaluates the actual config that will be used to restart.

reyortiz3:
@claude can you review what comments are pending address?

jhrozek (approved):
Changes look good. The status-setting and deduplication fixes address the review comments well.

pkg/workloads/manager.go
reyortiz3: solved

PR #4872, opened 2026-04-15
Add registry-aware workload creation

### LINKED ISSUE #4870: As a Studio user, I want to run a registry server without extracting fields myself
## User Story

As a ToolHive Studio user, I want to create a workload by referencing a registry server name so that I don't have to extract image, transport, env vars, and permissions client-side.

### What Changed

Add optional `registry` and `server` fields to `POST /api/v1beta/workloads`:

```json
{"registry": "default", "server": "io.github.stacklok/fetch", "env_vars": {"API_KEY": "..."}}
```

thv resolves the server from the registry, extracts metadata, and builds the RunConfig server-side. User-provided fields always override registry defaults.

### Acceptance Criteria

- [ ] Both `registry` and `server` must be provided together (400 if only one)
- [ ] Server resolved from registry, metadata fills missing request fields
- [ ] User-provided fields (env_vars, permissions, etc.) override registry defaults
- [ ] Image servers: fills image, transport, target_port, args, permissions, env var defaults
- [ ] Remote servers: fills URL, headers
- [ ] Studio can create workloads with a single POST (no GET+extract+POST)
- [ ] `task build` + `task test` + `task lint` pass

rdimitrov:
## Summary

- Studio currently does a multi-step flow to run a registry server: GET server metadata → extract fields client-side → POST /api/v1beta/workloads with all fields. This is fragile and requires the FE to understand the full metadata structure.
- Add optional `registry` and `server` fields to the workload creation API so clients can create workloads with a single POST. thv resolves the server from the registry and fills in image, transport, env vars, permissions, etc. server-side. User-provided fields always override registry defaults.

Fixes #4870

## Type of change

- [x] New feature

## Test plan

## Changes

| File | Change |
|------|--------|
| `pkg/api/v1/workload_types.go` | Add `registry` and `server` fields to `createRequest` |
| `pkg/api/v1/workload_service.go` | Registry resolution + defaults merging helpers |

Yes — `POST /api/v1beta/workloads` now accepts optional `registry` and `server` fields:
```json
{"registry": "default", "server": "io.github.stacklok/fetch", "env_vars": {"API_KEY": "..."}}
```

### DIFF
--- a/pkg/api/v1/workload_service.go
+++ b/pkg/api/v1/workload_service.go
@@ -154,7 +154,7 @@ func (s *WorkloadService) BuildFullRunConfig(
 	var registryResolvedMetadata regtypes.ServerMetadata
 	if req.Registry != "" && req.Server != "" {
 		var err error
-		registryResolvedMetadata, err = resolveRegistryServer(req)
+		registryResolvedMetadata, err = s.resolveRegistryServer(req)
 		if err != nil {
 			return nil, fmt.Errorf("failed to resolve server from registry: %w", err)
 		}
@@ -214,9 +214,36 @@ func (s *WorkloadService) BuildFullRunConfig(
 	// remote auth config, proxy port) picks it up automatically.
 	if registryResolvedMetadata != nil {
 		serverMetadata = registryResolvedMetadata
-		if img, ok := registryResolvedMetadata.(*regtypes.ImageMetadata); ok {
-			imageMetadata = img
-			imageURL = img.Image
+		switch md := registryResolvedMetadata.(type) {
+		case *regtypes.ImageMetadata:
+			imageMetadata = md
+			imageURL = md.Image
+		case *regtypes.RemoteServerMetadata:
+			if req.ProxyPort == 0 && md.ProxyPort > 0 {
+				registryProxyPort = md.ProxyPort
+			}
+			if md.OAuthConfig != nil {
+				resource := req.OAuthConfig.Resource
+				if resource == "" {
+					resource = md.OAuthConfig.Resource
+				}
+				if resource == "" && md.URL != "" {
+					resource = remote.DefaultResourceIndicator(md.URL)
+				}
+				remoteAuthConfig = &remote.Config{
+					ClientID:     req.OAuthConfig.ClientID,
+					Scopes:       md.OAuthConfig.Scopes,
+					CallbackPort: md.OAuthConfig.CallbackPort,
+					Issuer:       md.OAuthConfig.Issuer,
+					AuthorizeURL: md.OAuthConfig.AuthorizeURL,
+					TokenURL:     md.OAuthConfig.TokenURL,
+					UsePKCE:      md.OAuthConfig.UsePKCE,
+					Resource:     resource,
+					OAuthParams:  md.OAuthConfig.OAuthParams,
+					Headers:      md.Headers,
+					EnvVars:      md.EnvVars,
+				}
+			}
 		}
 	}
 
@@ -591,14 +618,14 @@ func (s *WorkloadService) GetWorkloadNamesFromRequest(ctx context.Context, req b
 
 // resolveRegistryServer resolves a server from the registry and fills in
 // default values on the request. User-provided fields are not overwritten.
-func resolveRegistryServer(req *createRequest) (regtypes.ServerMetadata, error) {
+func (s *WorkloadService) resolveRegistryServer(req *createRequest) (regtypes.ServerMetadata, error) {
 	// Only "default" registry is currently supported.
 	if req.Registry != "default" {
 		return nil, fmt.Errorf("unknown registry %q; only \"default\" is currently supported", req.Registry)
 	}
 
 	provider, err := registry.GetDefaultProviderWithConfig(
-		config.NewProvider(),
+		s.configProvider,
 		registry.WithInteractive(false),
 	)
 	if err != nil {
--- a/pkg/api/v1/workload_service.go
+++ b/pkg/api/v1/workload_service.go
@@ -147,9 +147,15 @@ func (s *WorkloadService) BuildFullRunConfig(
 	if (req.Registry != "" && req.Server == "") || (req.Registry == "" && req.Server != "") {
 		return nil, fmt.Errorf("both registry and server must be specified together")
 	}
-	// If registry+server specified, resolve from registry and fill defaults
+	// If registry+server specified, resolve from registry and fill defaults.
+	// The returned metadata is assigned to the local variables so the rest of
+	// BuildFullRunConfig (registry info, tool validation, remote auth, etc.)
+	// has access to it without re-looking up the server.
+	var registryResolvedMetadata regtypes.ServerMetadata
 	if req.Registry != "" && req.Server != "" {
-		if err := resolveRegistryServer(req); err != nil {
+		var err error
+		registryResolvedMetadata, err = resolveRegistryServer(req)
+		if err != nil {
 			return nil, fmt.Errorf("failed to resolve server from registry: %w", err)
 		}
 	}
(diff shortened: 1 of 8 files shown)


JAORMX (changes requested):
The registry metadata from `resolveRegistryServer` gets discarded after populating `req` fields, so the `RunConfig` registry info (`RegistryAPIURL`, `RegistryURL`, `RegistryServerName`) ends up empty in both the image and remote paths.

For image servers, `resolveRegistryServer` looks up by server name (e.g. `io.github.stacklok/fetch`), fills `req.Image` with the bare image ref (e.g. `ghcr.io/stacklok/fetch:latest`), then throws the metadata away. `BuildFullRunConfig` then calls `s.imageRetriever` with that image ref, which internally does `provider.GetServer("ghcr.io/stacklok/fetch:latest")`. That won't match by server name, so the lookup fails, `serverMetadata` stays nil, and `ResolveRegistrySourceURLs(nil, cfg)` returns empty strings.

For remote servers, `applyRemoteDefaults` fills `req.URL`, so we enter the `if req.URL != ""` branch and `serverMetadata` is never assigned at all. OAuth config, env vars, and proxy port from the registry metadata are also silently dropped since `applyRemoteDefaults` only copies URL and Headers.

On top of that, `imageMetadata` ends up nil for the image path (since the re-lookup by image ref fails), so the builder's `validateConfig` won't have metadata for tool validation, transport fallback, etc.

I think `resolveRegistryServer` should return the `ServerMetadata` it resolved, and `BuildFullRunConfig` should assign it to the existing `serverMetadata` variable directly. That way the existing code that populates registry info on the RunConfig, builds `remoteAuthConfig`, etc. all just works without duplicating logic.

pkg/api/v1/workload_service.go
```diff
@@ -557,3 +588,79 @@ func (s *WorkloadService) GetWorkloadNamesFromRequest(ctx context.Context, req b
...
+}
+
+func applyRemoteDefaults(req *createRequest, md *regtypes.RemoteServerMetadata) {
+	if req.URL == "" {
+		req.URL = md.URL
+	}
+	if len(req.Headers) == 0 && len(md.Headers) > 0 {
+		req.Headers = md.Headers
+	}
+}
```
JAORMX, line 735: So, this is the same class of issue I flagged before, but on the remote server path. `applyRemoteDefaults` only copies `URL` and `Headers`, but `RemoteServerMetadata` also carries `OAuthConfig`, `ProxyPort`, and `EnvVars`. Those get silently dropped.

Compare with lines 260-299 (the existing `imageRetriever` path for remote servers) which explicitly extracts `ProxyPort`, builds a full `remoteAuthConfig` from the metadata's `OAuthConfig` (Scopes, Issuer, AuthorizeURL, TokenURL, UsePKCE, etc.), and carries over `Headers` and `EnvVars`.

If someone does `{"registry": "default", "server": "some-oauth-remote-server"}`, they'd get the URL... but the OAuth config that makes it actually work would be gone. Silent failure, which is the worst kind.

pkg/api/v1/workload_service.go
```diff
@@ -190,6 +208,18 @@ func (s *WorkloadService) BuildFullRunConfig(
...
+	// If we resolved metadata from a registry reference, assign it to the
+	// local variables so downstream code (registry info, tool validation,
+	// remote auth config, proxy port) picks it up automatically.
+	if registryResolvedMetadata != nil {
+		serverMetadata = registryResolvedMetadata
+		if img, ok := registryResolvedMetadata.(*regtypes.ImageMetadata); ok {
+			imageMetadata = img
+			imageURL = img.Image
+		}
+	}
```
JAORMX, line 260: This block correctly handles `ImageMetadata` (sets `serverMetadata`, `imageMetadata`, and `imageURL`). That addresses the issue I raised before for image servers.

But for `RemoteServerMetadata`, it only sets `serverMetadata`. The `registryProxyPort` stays 0 and `remoteAuthConfig` is never populated from the registry metadata.

What happens next: `applyRemoteDefaults` set `req.URL`, so the flow enters `if req.URL != ""` at line 229, which calls `createRequestToRemoteAuthConfig(ctx, req)`. That function builds auth config purely from the user-provided `req.OAuthConfig` fields... so the registry's OAuth config, EnvVars, and ProxyPort are all lost.

I think you need either:
- A `*regtypes.RemoteServerMetadata` branch here that extracts `registryProxyPort` and builds `remoteAuthConfig` the same way lines 260-299 do, or
- Have `applyRemoteDefaults` merge everything (OAuthConfig fields, ProxyPort, EnvVars) into `req` so that `createRequestToRemoteAuthConfig` picks it all up

pkg/api/v1/workload_service.go
```diff
@@ -557,3 +588,79 @@ func (s *WorkloadService) GetWorkloadNamesFromRequest(ctx context.Context, req b
...
+func resolveRegistryServer(req *createRequest) (regtypes.ServerMetadata, error) {
+	// Only "default" registry is currently supported.
+	if req.Registry != "default" {
+		return nil, fmt.Errorf("unknown registry %q; only \"default\" is currently supported", req.Registry)
+	}
+
+	provider, err := registry.GetDefaultProviderWithConfig(
+		config.NewProvider(),
+		registry.WithInteractive(false),
+	)
```
JAORMX, line 660: `WorkloadService` already holds `s.configProvider` (injected at construction, used at line 321 for the config snapshot). This creates a separate provider via `config.NewProvider()`, which could return different config if something changes between the two calls. It also makes this function harder to test since you can't mock the config.

I'd make `resolveRegistryServer` a method on `WorkloadService` (or accept the config provider as a parameter) and use `s.configProvider` instead. That way it's consistent with the "snapshot config once per request" approach in `BuildFullRunConfig`.

pkg/api/v1/workloads.go
```diff
@@ -287,10 +287,10 @@ func (s *WorkloadRoutes) createWorkload(w http.ResponseWriter, r *http.Request)
 		)
 	}
 
-	// Validate that image or URL is provided
-	if req.Image == "" && req.URL == "" {
+	// Validate that image, URL, or registry+server is provided
+	if req.Image == "" && req.URL == "" && (req.Registry == "" || req.Server == "") {
```
JAORMX: Nit: when someone sends `registry: "default"` but forgets `server` (and no `image`/`url`), this returns "either 'image', 'url', or 'registry'+'server' fields are required". The actual problem is more specific: "you set registry but forgot server." The guard clause in `BuildFullRunConfig` (line 147) does catch this with a better message, but we'd hit this generic error first.

Not a blocker, but it'd be a nicer DX to check the partial registry+server case here and return a specific message before falling through.

JAORMX (changes requested):
Hey Rado! So, the core issue I flagged before (metadata getting discarded) is now fixed for the image server path. `resolveRegistryServer` returning `ServerMetadata` and assigning it to the local `serverMetadata`/`imageMetadata` variables is exactly the shape I had in mind. That part looks clean.

But... the remote server path still has the same class of problem. When `resolveRegistryServer` returns a `RemoteServerMetadata`, `applyRemoteDefaults` only copies `URL` and `Headers`. Then `BuildFullRunConfig` enters `if req.URL != ""` and calls `createRequestToRemoteAuthConfig(ctx, req)`, which builds auth config purely from user-provided fields. The registry's `OAuthConfig`, `EnvVars`, and `ProxyPort` are silently dropped. See the inline comments for details and suggestions.

A couple more things worth looking at:

- **Image provenance verification is skipped.** The normal `ResolveMCPServer` path calls `verifyImage` for Sigstore attestation checks. When registry+server bypasses `imageRetriever` entirely, that verification never runs. Worth at least documenting if intentional, or adding the check.

- **Tests.** The new functions (`resolveRegistryServer`, `applyRegistryDefaults`, `applyImageDefaults`, `applyRemoteDefaults`) don't have unit tests. At minimum the env var merge logic and the remote defaults path should be covered.

The overall approach is solid. I think the main fix is making sure remote servers get the same treatment as image servers (either extract `registryProxyPort` + `remoteAuthConfig` from the metadata in the assignment block, or have `applyRemoteDefaults` merge everything into `req`). Once that's in, this should be good to go.

JAORMX (approved):
Thanks Rado, all my previous feedback has been addressed:

- RemoteServerMetadata branch in `BuildFullRunConfig` now extracts `registryProxyPort` and builds `remoteAuthConfig` — OAuth config, headers, env vars, and proxy port no longer silently dropped on the remote path.
- `resolveRegistryServer` is now a method on `WorkloadService` using the injected `s.configProvider` — consistent with the snapshot-config approach.
- Partial `registry`/`server` now returns a specific 400 with a better message.
- Image provenance verification is called via the exported `retriever.VerifyImage` for registry-resolved image servers.
- Good test coverage added for `applyImageDefaults`, `applyRemoteDefaults`, `applyRegistryDefaults`, `buildRemoteAuthConfigFromMetadata`, and the unknown-registry path.

One minor nit (non-blocking): the registry-resolved path hardcodes `retriever.VerifyImageWarn`, while the `imageRetriever` path threads through a configured verification mode. Worth a follow-up to keep behavior consistent, but fine for this PR.

Also — could you mark the three stale review threads as resolved? They still show as open on my end even though the code is in good shape.

LGTM!

PR #4873, opened 2026-04-15
Drop registry-based group support

### LINKED ISSUE #4867: Drop registry-based group support from registry data
## Chore

Remove the ability to deploy server groups from registry data (`thv group run`). Registry-based groups were a ToolHive-specific concept embedded in registry JSON files that adds complexity without active use.

### What

- `groupRunCmdFunc` returns clear error with migration guidance
- `handleGroupLookup` in retriever returns "no longer supported" error
- Delete ~400 lines of dead group deployment/validation code

### Completion Checklist

- [ ] `thv group run` returns informative error pointing to `thv group create` + `thv run --group`
- [ ] Runtime groups (`thv group create/list/rm`) and groups API unaffected
- [ ] MCP optimizer continues to work (uses runtime groups, not registry groups)
- [ ] `task build` + `task test` + `task lint` pass

rdimitrov:
## Summary

- Registry-based groups (server collections defined in registry JSON files, deployed via `thv group run`) are a ToolHive-specific concept not in the upstream MCP spec. They add complexity without active use.
- Remove the registry group deployment path. Runtime groups (`thv group create/list/rm`) and the groups API continue to work unchanged. The MCP optimizer's use of runtime groups is not affected.

Fixes #4867

## Type of change

## Test plan

## Changes

| File | Change |
|------|--------|
| `cmd/thv/app/group.go` | `groupRunCmd` returns clear error; delete ~400 lines of dead deployment/validation code |
| `pkg/runner/retriever/retriever.go` | `handleGroupLookup` returns "no longer supported" error |

Yes — `thv group run` now returns an error directing users to use `thv group create` and `thv run --group` instead.

Runtime groups (workload grouping via `thv group create/list/rm` and the groups API) are completely unaffected. The MCP optimizer uses runtime groups, not registry groups, so it continues to work.


pkg/runner/retriever/retriever.go
```diff
@@ -218,62 +218,18 @@ func handleProtocolScheme(
...
-		// Server not found in group - fail explicitly
-		return "", nil, nil, fmt.Errorf("server '%s' not found in group '%s'", serverOrImage, groupName)
-	}
-
-	return imageToUse, imageMetadata, nil, nil
+	return "", nil, nil, fmt.Errorf(
+		"registry-based group '%s' is no longer supported; use 'thv group' commands to manage workload groups",
+		groupName,
+	)
 }
```
jhrozek: Nit (non-blocking): since `handleGroupLookup` now always returns an error, the caller at lines 98-108 has unreachable dead code after the call (the remote server early-return check can never execute). You could inline the error at the call site and delete this stub entirely:

```go
if groupName != "" {
    return "", nil, fmt.Errorf(
        "registry-based group '%s' is no longer supported; use 'thv group' commands to manage workload groups",
        groupName,
    )
}
```

Fine to defer to a follow-up that also removes the `groupName` parameter from the signature.

cmd/thv/app/group.go
```diff
@@ -65,12 +60,13 @@ var groupRmCmd = &cobra.Command{
 
 var groupRunCmd = &cobra.Command{
 	Use:   "run [group-name]",
-	Short: "Deploy all MCP servers from a registry group",
-	Long: `Deploy all MCP servers defined in a registry group.
-		 This creates a new runtime group and starts all MCP servers within it.`,
-	Args:    cobra.ExactArgs(1),
-	PreRunE: validateGroupArg(),
-	RunE:    groupRunCmdFunc,
+	Short: "Deploy all MCP servers from a registry group (removed)",
```
jhrozek: Nit (non-blocking): Cobra has a built-in `Deprecated` field that auto-hides the command from `--help` and prints a standard deprecation warning:

```go
Deprecated: "registry-based groups are no longer supported; use 'thv group create' and 'thv run --group' instead",
```

This would be slightly cleaner than showing `(removed)` in the help text. Either way works.

jhrozek (dismissed):
Clean removal. The deletion is thorough with no dangling references to removed functions (~500 lines gone, all callers verified).

Two non-blocking items for follow-up:

1. **Architecture docs** (`docs/arch/07-groups.md:62-76`, `docs/arch/06-registry-system.md:830-835`) still describe registry groups as a live feature and show `thv group run data-pipeline` as a valid command. Worth updating so readers don't try to use them.

2. **Dead `groupName` plumbing**: After this PR, no caller passes a non-empty `groupName` to `ResolveMCPServer` (`run.go:209`, `workload_service.go:217`, `run_server.go:48` all pass `""`). The parameter and the `if groupName != ""` branch are vestigial. Good follow-up cleanup candidate.

Inline comments below on two small things.

cmd/thv/app/group.go
rdimitrov: Thanks, addressed!

pkg/runner/retriever/retriever.go
rdimitrov: Thanks, addressed!

jhrozek (approved):

PR #4875, opened 2026-04-16
Default upstream redirectUri to {resourceUrl}/oauth/callback

### LINKED ISSUE #4874: Operator does not default redirectUri for OIDC upstream providers as documented
## Bug

The `MCPExternalAuthConfig` CRD documents that `redirectUri` on upstream providers defaults to `{resourceUrl}/oauth/callback` when not specified:

```go
// RedirectURI is the callback URL where the upstream IDP will redirect after authentication.
// When not specified, defaults to `{resourceUrl}/oauth/callback` where `resourceUrl` is the
// URL associated with the resource (e.g., MCPServer or vMCP) using this config.
// +optional
RedirectURI string `json:"redirectUri,omitempty"`
```

However, the operator passes the empty string through to the proxy runner without resolving the default. The runtime then rejects it:

```
Error: failed to create embedded auth server: failed to create auth server: invalid config: upstream "okta": redirect_uri is required
```

## Where

In `cmd/thv-operator/pkg/controllerutil/authserver.go` around line 618-621, the `RedirectURI` is copied verbatim from the CRD spec without applying the documented default:

```go
config.OIDCConfig = &authserver.OIDCUpstreamRunConfig{
    IssuerURL:   provider.OIDCConfig.IssuerURL,
    ClientID:    provider.OIDCConfig.ClientID,
    RedirectURI: provider.OIDCConfig.RedirectURI, // empty if not set — no defaulting
    Scopes:      provider.OIDCConfig.Scopes,
}
```

This affects both OIDC and OAuth2 upstream provider types.

## Expected behavior

When `redirectUri` is omitted from the upstream provider config, the operator should resolve it to `{resourceUrl}/oauth/callback` (using the MCPServer's `oidcConfig.resourceUrl`) before generating the runconfig.

## Workaround

Explicitly set `redirectUri` on the upstream provider in the `MCPExternalAuthConfig`.

mvanhorn:
## Summary

The `MCPExternalAuthConfig` CRD documents that `redirectUri` on upstream providers defaults to `{resourceUrl}/oauth/callback` when unset, but the operator was passing the empty string straight through to the proxy runner, which rejected it:

```
Error: failed to create embedded auth server: failed to create auth server:
invalid config: upstream "okta": redirect_uri is required
```

This means users have to set `redirectUri` explicitly, even though the CRD comment promises a default. The fix resolves the default inside `BuildAuthServerRunConfig` so the runtime config matches what the CRD documents.

- Resolve `{resourceUrl}/oauth/callback` as the fallback redirect URI inside `buildUpstreamRunConfig`, using the first allowed audience (which is the resource URL for both `MCPServer` and `VirtualMCPServer`) as the base.
- Apply the default to both OIDC and OAuth2 upstream provider blocks.
- New `defaultRedirectURI` helper keeps the null-check logic readable and testable.
- Explicit `redirectUri` values continue to pass through untouched; no `resourceURL` context means no defaulting (preserving existing validation behavior).

Fixes #4874

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [ ] Lint (`task lint-fix`)
- [x] Manual verification

Verified with `go build ./cmd/thv-operator/pkg/controllerutil/` and `go vet ./cmd/thv-operator/pkg/controllerutil/` locally. The existing `TestBuildAuthServerRunConfig` suite still compiles against the updated `buildUpstreamRunConfig` signature via the package-internal call at line 477. A follow-up commit could add dedicated coverage asserting both the default path (empty `redirectUri` + resourceURL -> `{resourceURL}/oauth/callback`) and the pass-through path (explicit `redirectUri` preserved), but I kept this PR focused on the reported behavioral gap to stay in scope.

Yes. Users who previously had to set `redirectUri` on every `UpstreamProviderConfig` no longer need to when `oidcConfig.resourceUrl` (or the virtual equivalent) is set — the operator now honors the documented default. Existing configs with explicit `redirectUri` values are unchanged.

### DIFF
--- a/cmd/thv-operator/pkg/controllerutil/authserver.go
+++ b/cmd/thv-operator/pkg/controllerutil/authserver.go
@@ -470,11 +470,18 @@ func BuildAuthServerRunConfig(
 		}
 	}
 
-	// Build upstream provider configs using shared bindings
+	// Build upstream provider configs using shared bindings.
+	// Pass the resource URL (first allowed audience) so upstream providers can
+	// default redirectUri to `{resourceUrl}/oauth/callback` when unset, as
+	// documented on the MCPExternalAuthConfig CRD.
+	resourceURL := ""
+	if len(allowedAudiences) > 0 {
+		resourceURL = allowedAudiences[0]
+	}
 	bindings := buildUpstreamSecretBindings(authConfig.UpstreamProviders)
 	config.Upstreams = make([]authserver.UpstreamRunConfig, 0, len(bindings))
 	for _, b := range bindings {
-		config.Upstreams = append(config.Upstreams, *buildUpstreamRunConfig(b.Provider, b.EnvVarName))
+		config.Upstreams = append(config.Upstreams, *buildUpstreamRunConfig(b.Provider, b.EnvVarName, resourceURL))
 	}
 
 	// Build storage configuration
@@ -603,9 +610,13 @@ func resolveSentinelAddrs(
 // buildUpstreamRunConfig converts CRD UpstreamProviderConfig to authserver.UpstreamRunConfig.
 // The envVarName is computed by buildUpstreamSecretBindings to keep Pod env
 // and runtime config in sync.
+// resourceURL is used to default redirectUri to `{resourceURL}/oauth/callback`
+// when the CRD leaves it empty, matching the default documented on
+// MCPExternalAuthConfig.
 func buildUpstreamRunConfig(
 	provider *mcpv1alpha1.UpstreamProviderConfig,
 	envVarName string,
+	resourceURL string,
 ) *authserver.UpstreamRunConfig {
 	config := &authserver.UpstreamRunConfig{
 		Name: provider.Name,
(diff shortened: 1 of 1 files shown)


JAORMX:
Thanks for taking a look at this! However, there are [contribution guidelines](https://github.com/stacklok/toolhive/blob/main/CONTRIBUTING.md) in the project that must be followed. Please follow them and we'll happily take contributions and help you in Discord in case you have questions.

Thanks again!

mvanhorn:
Apologies for jumping ahead of the claim-first step in CONTRIBUTING.md - I opened #4874 and went straight to the PR without waiting for assignment. Happy to close this and reopen after an assignment if that's the preferred path.

For CI: the failing `embeddingserver-cache` operator e2e scenario looks unrelated to this change (statefulset readiness timeout), but I'll defer to the team on re-running.

JAORMX:
@mvanhorn it's all good, just ask for an assignment in the issue, I'll assign you, and that's all. It's just a way for us to actually differentiate between users willing to engage and that follow the rules and folks that just do drive-by PRs with agents. Usage of agents is fine, and we don't want to discourage it, so it's mostly a way for us to have some control and engagement. I'll look into the flaky CI.

PR #4878, opened 2026-04-16
Account for embedded auth env in deployment drift checks

### LINKED ISSUE #4877: Proxy runner triggers StatefulSet rolling update on every restart, causing crash loop
## Bug

MCPServers with `externalAuthConfigRef` (embedded auth server) enter a crash loop where the proxy Deployment and backend StatefulSet are continuously rolled.

## Symptoms

1. The StatefulSet `.metadata.generation` increments on every proxy restart (observed going from 2 → 51+)
2. The proxy runner logs show the StatefulSet going from `ready_replicas: 1` → `ready_replicas: 0` each time it starts
3. After the 5-minute initialize timeout, the proxy reports "workload started successfully" but the transport immediately dies ("transport is no longer running")
4. The proxy enters CrashLoopBackOff

```
waiting for statefulset to be ready... ready_replicas:1, observed_generation:7, desired_generation:8
waiting for statefulset to be ready... ready_replicas:0
Waiting for MCP server to be ready, endpoint: http://localhost:8080/mcp
initialize not successful, but continuing
workload started successfully, press Ctrl+C to stop
transport is no longer running, attempting automatic restart
```

MCPServers **without** `externalAuthConfigRef` do not exhibit this behavior — their StatefulSets stabilize normally.

## To reproduce

1. Create an MCPServer with `externalAuthConfigRef` pointing to an `MCPExternalAuthConfig` of type `embeddedAuthServer` with an OIDC upstream provider
2. Ensure the required ExternalSecrets (ECDSA signing key, HMAC secret) are synced
3. Observe that the proxy Deployment pod enters CrashLoopBackOff
4. Check the StatefulSet generation — it increments on each proxy restart

## Environment

- ToolHive operator v0.20.0
- Kubernetes (EKS)

JAORMX:
## Summary
- include embedded auth env vars when comparing expected proxy deployment env for `MCPServer`
- do the same for `MCPRemoteProxy` so self-generated deployments are not treated as drifted
- add regression tests for embedded auth and token exchange cases, plus associative-list reordering checks in the Kubernetes client tests

This is the concrete operator-side fix identified while debugging #4877.

## Testing
- `go test ./cmd/thv-operator/controllers ./pkg/container/kubernetes`

Closes #4877.

### DIFF
--- a/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
+++ b/cmd/thv-operator/controllers/mcpremoteproxy_controller.go
@@ -1274,6 +1274,15 @@ func (r *MCPRemoteProxyReconciler) containerNeedsUpdate(
 
 	// Check if environment variables have changed
 	expectedEnv := r.buildEnvVarsForProxy(ctx, proxy)
+	if configName := ctrlutil.EmbeddedAuthServerConfigName(proxy.Spec.ExternalAuthConfigRef, proxy.Spec.AuthServerRef); configName != "" {
+		_, _, authServerEnvVars, err := ctrlutil.GenerateAuthServerConfigByName(
+			ctx, r.Client, proxy.Namespace, configName,
+		)
+		if err != nil {
+			return true
+		}
+		expectedEnv = append(expectedEnv, authServerEnvVars...)
+	}
 	if !reflect.DeepEqual(container.Env, expectedEnv) {
 		return true
 	}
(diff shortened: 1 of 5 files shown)


jhrozek (dismissed):
Looks good — the core fix is correct, env var ordering between creation and drift check paths is consistent, and test coverage directly validates the bug fix.

One nit: the two SSA reordering tests in `pkg/container/kubernetes/client_test.go` (`TestStatefulSetApplyAssociativeListReorderingDoesNotChangeGeneration`, `TestServiceApplyAssociativeListReorderingDoesNotChangeGeneration`) test the behavior of `fake.NewClientset()`, not any production code in this package. The fake client doesn't implement real SSA merge semantics, so these document an assumption about SSA associative-list handling rather than validating it. If real validation is wanted, these would need to run against envtest. They're also unrelated to the drift check fix itself — might be cleaner as a separate commit.

jhrozek (dismissed):

jhrozek (approved):

PR #4894, opened 2026-04-16
fix(thv-operator): default upstream redirectUri to {resourceUrl}/oauth/callback

### LINKED ISSUE #4874: Operator does not default redirectUri for OIDC upstream providers as documented

tmchow:
## Summary

Per the `MCPExternalAuthConfig.UpstreamProviderConfig.RedirectURI` godoc, the upstream provider `redirectUri` defaults to `{resourceUrl}/oauth/callback` when unset. The operator was copying the empty string from the CRD straight into the runconfig, so the embedded auth server rejected registration with `upstream "okta": redirect_uri is required`.

## Fix

- Thread `resourceUrl` from `AddEmbeddedAuthServerConfigOptions` (both MCPServer and VirtualMCPServer paths) and from `vmcpconfig/converter.go` into `BuildAuthServerRunConfig`.
- New helper `deriveDefaultRedirectURI(resourceURL)` returns `{resourceUrl}/oauth/callback` (with a trailing-slash trim so `"https://host/"` and `"https://host"` produce the same value), or `""` when resourceUrl is empty.
- `buildUpstreamRunConfig` takes a `defaultRedirectURI` argument and resolves it when the CRD's `RedirectURI` is empty. Explicit upstream `RedirectURI` values are preserved.
- Applies to both OIDC and OAuth2 upstream provider types.

## Note on path-prefixed resourceURLs

Codex review flagged an initial implementation that stripped the path from `resourceUrl` before appending `/oauth/callback`. After investigating the ingress/gateway deployment pattern (e.g. `https://mcp-gateway.example.com/mcp`), those deployments mount `/oauth/*` under the same path prefix, so literal `{resourceUrl}/oauth/callback` concatenation is correct in both host-only and path-prefixed cases. Kept the implementation aligned with the CRD godoc contract.

## Changes

- `cmd/thv-operator/pkg/controllerutil/authserver.go`: new `deriveDefaultRedirectURI` helper; extended `BuildAuthServerRunConfig` signature with `resourceURL`; threads the default through `buildUpstreamRunConfig`; both in-file callers now pass `oidcConfig.ResourceURL`.
- `cmd/thv-operator/pkg/vmcpconfig/converter.go`: new `deriveResourceURL` helper mirroring `deriveAllowedAudiences`; `convertAuthServerConfig` passes it through.
- `cmd/thv-operator/pkg/controllerutil/authserver_test.go`: existing `TestBuildAuthServerRunConfig` call sites updated for the new signature. Added cases for:
  - empty upstream `RedirectURI` defaulting to `{resourceUrl}/oauth/callback`
  - explicit upstream `RedirectURI` preserved over the default
  - path-prefixed resourceUrl preserves the path in the default
  - trailing-slash resourceUrl normalized before appending

## Test plan

```
go build ./...
go test ./cmd/thv-operator/pkg/controllerutil/... ./cmd/thv-operator/pkg/vmcpconfig/... -count=1
```

Both pass (323 tests total, +2 new cases for the defaulting behavior).

Fixes #4874
[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/cmd/thv-operator/pkg/controllerutil/authserver.go
+++ b/cmd/thv-operator/pkg/controllerutil/authserver.go
@@ -407,6 +407,7 @@ func AddEmbeddedAuthServerConfigOptions(
 	embeddedConfig, err := BuildAuthServerRunConfig(
 		namespace, mcpServerName, authServerConfig,
 		[]string{oidcConfig.ResourceURL}, oidcConfig.Scopes,
+		oidcConfig.ResourceURL,
 	)
 	if err != nil {
 		return fmt.Errorf("failed to build embedded auth server config: %w", err)
@@ -430,6 +431,7 @@ func BuildAuthServerRunConfig(
 	authConfig *mcpv1alpha1.EmbeddedAuthServerConfig,
 	allowedAudiences []string,
 	scopesSupported []string,
+	resourceURL string,
 ) (*authserver.RunConfig, error) {
 	config := &authserver.RunConfig{
 		SchemaVersion:                authserver.CurrentSchemaVersion,
@@ -473,8 +475,13 @@ func BuildAuthServerRunConfig(
 	// Build upstream provider configs using shared bindings
 	bindings := buildUpstreamSecretBindings(authConfig.UpstreamProviders)
 	config.Upstreams = make([]authserver.UpstreamRunConfig, 0, len(bindings))
+	// The embedded auth server mounts /oauth/* at the transport root, so the
+	// default upstream callback must hang off the resource's scheme+host, not
+	// its full path. A resourceUrl like "https://gw.example.com/mcp" still
+	// needs the callback at "https://gw.example.com/oauth/callback".
+	defaultRedirectURI := deriveDefaultRedirectURI(resourceURL)
 	for _, b := range bindings {
-		config.Upstreams = append(config.Upstreams, *buildUpstreamRunConfig(b.Provider, b.EnvVarName))
+		config.Upstreams = append(config.Upstreams, *buildUpstreamRunConfig(b.Provider, b.EnvVarName, defaultRedirectURI))
 	}
 
 	// Build storage configuration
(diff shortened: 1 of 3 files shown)


JAORMX:
Thanks for taking a look at this! However, there are [contribution guidelines](https://github.com/stacklok/toolhive/blob/main/CONTRIBUTING.md) in the project that must be followed. Please follow them and we'll happily take contributions and help you in Discord in case you have questions.

Thanks again!

PR #4897, opened 2026-04-16
Add OverrideTokenSource hook to registry factory

amirejaz:
## Summary

- The registry factory's `resolveTokenSource` only supports the built-in OAuth config path. External builds that manage authentication separately (e.g. via a session manager) have no way to inject a token source without forking OSS code.
- Adds an `OverrideTokenSource` package-level hook. When set, it is called before the OAuth logic; returning a non-nil `TokenSource` uses it, returning `nil` falls through to existing behaviour unchanged.
- In OSS builds the variable is `nil` and the code path is identical to before.

## Type of change

## Test plan

- [x] Existing unit tests pass (`task test`)
- [x] Manual review: OSS default path is a pure no-op when `OverrideTokenSource` is nil — the added guard is a nil check before the existing logic

### DIFF
--- a/pkg/registry/factory.go
+++ b/pkg/registry/factory.go
@@ -25,6 +25,12 @@ var (
 	// The mutex is NOT needed for GetDefaultProviderWithConfig since
 	// sync.Once already provides thread-safety for initialization.
 	defaultProviderMu sync.Mutex
+
+	// OverrideTokenSource, if non-nil, is called before the built-in OAuth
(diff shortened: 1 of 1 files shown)


amirejaz:
Closing — the OverrideTokenSource hook is no longer needed. The enterprise build handles registry authentication via AutoLoginRegistry, which copies the session refresh token to the registry secrets store at login time using the existing OAuth token source path. No OSS hook is required.

PR #4904, opened 2026-04-16
Validate audience matches resourceUrl for embedded auth server

### LINKED ISSUE #4860: Audience mismatch between embedded auth server and token validator when using OIDCConfigRef
## Summary

When an MCPServer or MCPRemoteProxy is configured with both an `OIDCConfigRef` and an embedded auth server (`AuthServerRef` or `ExternalAuthConfigRef`), the `audience` value used for token minting and the `audience` value used for token validation come from **different sources** and can silently diverge, causing token validation failures that are difficult to debug.

### Token minting path (embedded auth server)

1. The OIDC resolver auto-computes `ResourceURL` from the server's name, namespace, and port: `http://{name}.{namespace}.svc.cluster.local:{port}`
2. `ResourceURL` is passed as the sole entry in `AllowedAudiences` to `BuildAuthServerRunConfig()`
3. The RFC 9728 discovery endpoint (`/.well-known/oauth-protected-resource`) advertises this `ResourceURL` as the `resource` value
4. The OAuth client reads the discovery response and sends `resource=<ResourceURL>` in the token request
5. The embedded auth server validates the `resource` parameter against `AllowedAudiences` and mints a token with `aud = <ResourceURL>`

### Token validation path (OIDC middleware)

1. The token validator's `audience` field is set from `OIDCConfigRef.Audience` — a **user-specified, required, per-server string** in the Kubernetes manifest
2. On every request, the validator checks that the token's `aud` claim contains this value

### The mismatch

| Component | Audience source | Example value |
|---|---|---|
| Discovery endpoint (`resource`) | Auto-computed `ResourceURL` | `http://my-server.my-ns.svc.cluster.local:8080` |
| Auth server `AllowedAudiences` | Auto-computed `ResourceURL` | `http://my-server.my-ns.svc.cluster.local:8080` |
| Token `aud` claim | Derived from client's `resource` param | `http://my-server.my-ns.svc.cluster.local:8080` |
| Token validator `audience` check | User-specified `OIDCConfigRef.Audience` | `https://my-server.example.com` |

Unless the operator manually sets `OIDCConfigRef.Audience` to match the auto-computed internal service URL, token validation will reject every token the embedded auth server mints.

## Why it matters

- The failure mode is silent: tokens are minted successfully, but every authenticated request is rejected with an audience mismatch error.
- The operator has no way to know what `ResourceURL` will be computed to without reading the controller source code.
- The `OIDCConfigRef.Audience` field is documented as required and per-server, giving operators the impression it controls the audience — but in the embedded auth server flow, the entire OAuth chain is driven by `ResourceURL` instead.
- Inline OIDC config (now deprecated across MCPServer, MCPRemoteProxy, and vMCP) did not have this problem because the operator controlled all three values (`Audience`, `Resource`, `AllowedAudiences`) and a startup validator caught mismatches. As users migrate to `OIDCConfigRef`, this gap becomes increasingly relevant.

### Option A: Default `Audience` to `ResourceURL` when using the embedded auth server

Override the token validator's `audience` to use `ResourceURL` instead of the user-specified `Audience` when an embedded auth server is active. Single source of truth, impossible to misconfigure.

### Option B: Align both sides to accept both values

Include both `ResourceURL` and `OIDCConfigRef.Audience` in `AllowedAudiences`, and make the token validator accept both values. Backwards-compatible but more complex.

### Option C: Use `Audience` as the canonical value everywhere

Replace `ResourceURL` with `OIDCConfigRef.Audience` everywhere. Operator has full control but loses automatic per-server uniqueness.

### Option D: Validate consistency at reconciliation time

Add a reconciler check that errors when `OIDCConfigRef.Audience != ResourceURL` and an embedded auth server is active. Surfaces the problem early but forces operators to use the internal service URL.

## Affected code paths

- OIDC resolver: `cmd/thv-operator/pkg/oidc/resolver.go`
- Auth server config builder: `cmd/thv-operator/pkg/controllerutil/authserver.go` (`BuildAuthServerRunConfig`, `AddAuthServerRefOptions`, `AddEmbeddedAuthServerConfigOptions`)
- RunConfig builder: `pkg/runner/config_builder.go` (`WithOIDCConfig`)
- Token validator: `pkg/auth/token.go` (audience check at line 959)
- Discovery endpoint: `pkg/auth/token.go` (`NewAuthInfoHandler` at line 1261)
- MCPServer controller: `cmd/thv-operator/controllers/mcpserver_runconfig.go`
- MCPRemoteProxy controller: `cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go`

tgrunnagle:
## Summary

When an MCPServer uses both an `OIDCConfigRef` and an embedded auth server, the token `aud` claim is derived from the `ResourceURL` (which is auto-computed when not specified), but the token validator checks against the user-specified `OIDCConfigRef.Audience`. If these values differ, every authenticated request silently fails with an audience mismatch -- a frustrating failure mode with no clear error message pointing to the root cause.

This PR adds reconciliation-time validation that `Audience` and `ResourceURL` match when an embedded auth server is active, surfacing the misconfiguration early with actionable error messages instead of letting it fail silently at runtime.

Closes #4860

## Type of change

- [x] Bug fix

## Test plan

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/pkg/controllerutil/authserver.go` | Extract duplicated OIDC validation into `validateOIDCConfigForEmbeddedAuthServer` helper; add audience-vs-resourceUrl consistency check with descriptive error messages |
| `cmd/thv-operator/pkg/controllerutil/authserver_test.go` | Add test cases for audience mismatch, empty audience, and matching audience scenarios in both `AddEmbeddedAuthServerConfigOptions` and `AddAuthServerRefOptions` |
| `cmd/thv-operator/controllers/mcpserver_externalauth_runconfig_test.go` | Update test fixtures to use matching audience/resourceUrl values |

Yes. Operators who configure an `OIDCConfigRef` with an `Audience` that does not match the auto-computed `ResourceURL` will now see a clear reconciliation error instead of silently broken authentication. The error message tells them exactly what value to set.

This implements Option D from the issue (validate consistency at reconciliation time). This approach was chosen over silently overriding `Audience` because it keeps both values explicit and gives operators full visibility into what the system expects -- consistent with how the existing `ValidateAuthServerIntegration` works for inline vMCP config.

### DIFF
--- a/cmd/thv-operator/pkg/controllerutil/authserver.go
+++ b/cmd/thv-operator/pkg/controllerutil/authserver.go
@@ -395,24 +395,8 @@ func AddEmbeddedAuthServerConfigOptions(
 		return fmt.Errorf("embedded auth server configuration is nil for type embeddedAuthServer")
 	}
 
-	// Validate OIDC config is provided with ResourceURL (required for embedded auth server)
-	if oidcConfig == nil {
-		return fmt.Errorf("OIDC config is required for embedded auth server: OIDCConfigRef must be set on the MCPServer")
-	}
-	if oidcConfig.ResourceURL == "" {
-		return fmt.Errorf("OIDC config resourceUrl is required for embedded auth server: set resourceUrl in OIDCConfigRef")
-	}
-
-	// The embedded auth server mints tokens with aud = ResourceURL (the value
-	// clients send as the RFC 8707 resource parameter). The token validator
-	// checks aud against Audience. These must match or every authenticated
-	// request will fail with an audience mismatch (#4860).
-	if oidcConfig.Audience != oidcConfig.ResourceURL {
-		return fmt.Errorf(
-			"oidcConfigRef.audience %q must match resourceUrl %q when an embedded auth server is active; "+
-				"set audience to %q or set resourceUrl to match audience",
-			oidcConfig.Audience, oidcConfig.ResourceURL, oidcConfig.ResourceURL,
-		)
+	if err := validateOIDCConfigForEmbeddedAuthServer(oidcConfig); err != nil {
+		return err
 	}
 
 	// Build the embedded auth server config for runner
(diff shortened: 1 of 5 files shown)


jerm-dro (approved):

cmd/thv-operator/pkg/controllerutil/authserver.go
```diff
@@ -395,12 +395,8 @@ func AddEmbeddedAuthServerConfigOptions(
...
 	}
 
-	// Validate OIDC config is provided with ResourceURL (required for embedded auth server)
-	if oidcConfig == nil {
-		return fmt.Errorf("OIDC config is required for embedded auth server: OIDCConfigRef must be set on the MCPServer")
-	}
-	if oidcConfig.ResourceURL == "" {
-		return fmt.Errorf("OIDC config resourceUrl is required for embedded auth server: set resourceUrl in OIDCConfigRef")
+	if err := validateOIDCConfigForEmbeddedAuthServer(oidcConfig); err != nil {
+		return err
```
jhrozek, line 399: nit: should we set a failed phase and condition `ConditionTypeAuthConfigured`/`/`ConditionReasonAuthInvalid` in this case?

jhrozek (approved):

cmd/thv-operator/pkg/controllerutil/authserver.go
tgrunnagle: I will add in a follow up PR

PR #4910, opened 2026-04-17
Add --system flag to secret list and delete commands

amirejaz:
## Summary

- System-managed tokens (`__thv_*` keys: registry OAuth, workload auth, enterprise login) are intentionally hidden from normal `thv secret` commands via `UserProvider`. There was no way to inspect or clean them up in cases like stale tokens after a re-install or provider switch — the last item from the RFC (toolhive-rfcs#56) that was not yet implemented.
- Adds `--system` to `thv secret list` (shows all `__thv_*` keys with scope label) and `thv secret delete` (allows deleting a named `__thv_*` key; rejects non-system keys).
- Uses a raw unscoped provider that bypasses `UserProvider`, keeping all normal command behaviour unchanged.

Closes stacklok/stacklok-enterprise-platform#103

## Type of change

## Test plan

- [x] Existing unit tests pass (`task test`)
- [x] Build passes (`go build ./cmd/thv/app/`)
- [x] Manual review: non-`--system` paths are untouched; `--system` delete rejects keys not starting with `__thv_`


cmd/thv/app/secret.go
```diff
@@ -282,6 +303,15 @@ If descriptions exist for the secrets, the command displays them alongside the n
 		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			ctx := cmd.Context()
+
+			if systemFlag {
+				provider, err := getSystemSecretsProvider()
+				if err != nil {
+					return fmt.Errorf("failed to create secrets provider: %w", err)
+				}
+				return runSystemSecretList(ctx, provider, os.Stdout)
```
reyortiz3, line 329: The `--system` list path skips the capability check that the normal path performs at line 321 (`manager.Capabilities().CanList`). The same gap exists for delete: line 256 calls `getSystemSecretsProvider()` without checking `CanDelete`, while the normal path checks at line 269.

On a read-only provider (1Password, env), this surfaces a raw backend error instead of the user-friendly message the non-`--system` path produces.

🤖 Generated with [Claude Code](https://claude.ai/code)

cmd/thv/app/secret.go
```diff
@@ -431,6 +486,67 @@ For more information, visit: https://developer.1password.com/docs/service-accoun
...
+	return nil
+}
+
+// runSystemSecretDelete validates that name is a system key and deletes it
+// using the given provider.
+func runSystemSecretDelete(ctx context.Context, provider secrets.Provider, name string) error {
+	if err := validateSystemKeyName(name); err != nil {
+		return err
+	}
+	if err := provider.DeleteSecret(ctx, name); err != nil {
```
reyortiz3, line 521: The normal delete path calls `warnWorkloadsUsingSecret(ctx, name)` at line 277 before deleting. System secrets (`__thv_*`) are auth/registry tokens used directly by running workloads — omitting the warning here is more impactful than for user secrets.

🤖 Generated with [Claude Code](https://claude.ai/code)

cmd/thv/app/secret.go
```diff
@@ -354,6 +388,27 @@ func getSecretsManager() (secrets.Provider, error) {
 	return authsecrets.GetUserSecretsProvider()
 }
 
+// getSystemSecretsProvider returns a raw (unscoped) secrets provider for
+// advanced/emergency operations on system-managed keys. Unlike getSecretsManager,
+// this does not apply UserProvider filtering, allowing callers to read and
+// delete __thv_* prefixed keys.
+func getSystemSecretsProvider() (secrets.Provider, error) {
```
reyortiz3: Per [`.claude/rules/cli-commands.md`](https://github.com/stacklok/toolhive/blob/fe8cdeb260599f9afb9658e7e818ba74b6b955da/.claude/rules/cli-commands.md) (Thin Wrapper Principle): "Business logic MUST live in `pkg/` packages." `getSystemSecretsProvider()`, `runSystemSecretList()`, and `runSystemSecretDelete()` duplicate the provider-factory pattern that was previously extracted into `pkg/auth/secrets/secrets.go` (`GetUserSecretsProvider`). These functions belong in `pkg/`.

🤖 Generated with [Claude Code](https://claude.ai/code)

reyortiz3 (commented):
Found 4 issues during review. See inline comments.

🤖 Generated with [Claude Code](https://claude.ai/code)

cmd/thv/app/secret.go
```diff
@@ -431,6 +486,67 @@ For more information, visit: https://developer.1password.com/docs/service-accoun
...
+		}
+	}
+
+	return nil
+}
+
+// runSystemSecretDelete validates that name is a system key and deletes it
+// using the given provider.
+func runSystemSecretDelete(ctx context.Context, provider secrets.Provider, name string) error {
+	if err := validateSystemKeyName(name); err != nil {
```
reyortiz3: `validateSystemKeyName` is already called in the `RunE` closure at line 253 before `runSystemSecretDelete` is invoked, making this call redundant. The two call sites have conflicting ownership: the `RunE` comment says "validate before touching the provider", while `runSystemSecretDelete`'s doc comment says it "validates that name is a system key" — implying it owns the responsibility.

🤖 Generated with [Claude Code](https://claude.ai/code)

reyortiz3 (approved):

cmd/thv/app/secret.go
```diff
@@ -431,6 +482,65 @@ For more information, visit: https://developer.1password.com/docs/service-accoun
...
+	if err := provider.DeleteSecret(ctx, name); err != nil {
+		return fmt.Errorf("failed to delete secret %s: %w", name, err)
+	}
+	return nil
+}
+
+// formatSystemSecretEntry formats a system-managed secret key for display.
+// Key format: __thv_<scope>_<name>
+// The full key is shown so it can be passed directly to "thv secret delete --system".
+func formatSystemSecretEntry(key string) string {
```
jhrozek, line 530: Minor: the `__thv_<scope>_<name>` format is now parsed by hand in two places — here in `formatSystemSecretEntry`, and earlier in the delete RunE (around line 268) where the prefix is stripped before calling `warnWorkloadsUsingSecret`. Both use the same `TrimPrefix + strings.Cut(_, "_")` pattern, and that format is really owned by `pkg/secrets` (see `ScopedProvider.getScopedKey` and the invariants documented in `scoped.go`).

Would it make sense to add `secrets.ParseSystemKey(key) (scope, name, ok)` (and maybe `secrets.IsSystemKey(name)`) so both call sites use that? One less copy to keep in sync if the format ever changes.

jhrozek (approved):

Issue #4633, opened 2026-04-07
Switch status writes from Update to Patch across all controllers


ChrisJBurns:
## Summary

Every controller uses `r.Status().Update(ctx, obj)` which replaces the entire status subresource. Concurrent writers overwrite each other's changes. This is especially dangerous for VirtualMCPServer where the runtime process and controller write to different status fields simultaneously.

**Severity**: MUST FIX
**Area**: Controller Logic
**Breaking**: No

## Location

- Every controller file under `cmd/thv-operator/controllers/`
- 35 `Status().Update` call sites in `mcpserver_controller.go` alone
- Zero `Status().Patch` calls found anywhere in the codebase

## Problem

`r.Status().Update(ctx, obj)` replaces the entire status subresource, meaning concurrent writers overwrite each other's changes. This is especially dangerous for VirtualMCPServer where the runtime process writes `status.discoveredBackends`, `status.backendCount`, and `BackendsDiscovered` condition while the controller writes infrastructure conditions.

## Impact

- Concurrent status updates cause data loss (last writer wins)
- VirtualMCPServer runtime and controller can overwrite each other
- Conflict errors under load leading to unnecessary requeues

## Recommended Fix

1. Switch all status writes to use `client.MergeFrom` patch:
   ```go
   patch := client.MergeFrom(obj.DeepCopy())
   // ... mutate status ...
   r.Status().Patch(ctx, obj, patch)
   ```
2. Start with VirtualMCPServer (highest risk due to concurrent runtime writes)
3. Then MCPServer (highest volume — 35 call sites)
4. Then all other controllers

## Related Issues

- Batch MCPServer status updates to reduce API calls per reconciliation (#4630)

jhrozek:
Closing as duplicate of #4767, which is linked to a user story and has been updated to preserve the broader `r.Status().Update` → `r.Status().Patch` migration concern from this issue as Related Future Work. #4767 handles the spec-side fix needed immediately for the enterprise authz controller; the status-subresource migration is tracked as follow-up from there.

jhrozek:
Reopening — this broader migration (Status().Update → Status().Patch across all controllers) should not be folded into #4767, which only covers the 3 spec-side r.Update sites on MCPServer. Keeping this as the canonical tracker so the broader work doesn't get forgotten when #4767 closes. #4767 is a narrow precursor, not a replacement.

jhrozek:
[...]

PR #4914, opened 2026-04-17
Patch MCPServer spec instead of Update

### LINKED ISSUE #4767: Fix r.Update to r.Patch plus regression guard in MCPServer controller
Part of stacklok/stacklok-enterprise-platform#stacklok/stacklok-enterprise-platform#378

## Description

Replace the three `r.Update(ctx, mcpServer)` call sites in the MCPServer controller with `r.Patch(ctx, mcpServer, client.MergeFrom(original))` to prevent the full-PUT from clobbering fields written by external controllers. Add a `.claude/rules/operator.md` rule and a `lint-mcpserver-update` CI target to guard against regression. This fix is a prerequisite for deploying the enterprise authorization controller, which writes `spec.authzConfig` on MCPServer resources via SSA -- without this patch, every OSS reconcile cycle silently clears that field.

## Context

The MCPServer controller's `Reconcile` method currently uses `r.Update(ctx, mcpServer)` at three points: removing the finalizer on deletion (line 283), adding the finalizer on creation (line 294), and updating the restart annotation (line 751). `r.Update` performs a full PUT of the entire resource, which means every field not present in the local copy is implicitly cleared on the server. When an external controller (the enterprise authorization controller) writes `spec.authzConfig` via server-side apply, the MCPServer controller's next `r.Update` wipes it because its in-memory copy never had that field.

The `r.Patch(ctx, obj, client.MergeFrom(original))` pattern sends only the diff between the original (deep-copied before mutation) and the mutated object. This is the standard controller-runtime approach for controllers that share field ownership with other actors.

Note: there are 4 additional `r.Update` calls in the controller (lines 398, 460, 479, 805) targeting `deployment` and `service` resources. Those are out of scope -- they do not affect MCPServer field ownership. A full SSA migration of the entire controller is tracked separately.

**Dependencies**: None (standalone fix)
**Blocks**: stacklok/stacklok-enterprise-platform#384 (enterprise authorization controller depends on this fix being merged first)

## Acceptance Criteria

- [ ] `r.Update(ctx, mcpServer)` at line 283 (finalizer removal) is replaced with `r.Patch(ctx, mcpServer, client.MergeFrom(original))` using a deep copy taken before `controllerutil.RemoveFinalizer`
- [ ] `r.Update(ctx, mcpServer)` at line 294 (finalizer addition) is replaced with `r.Patch(ctx, mcpServer, client.MergeFrom(original))` using a deep copy taken before `controllerutil.AddFinalizer`
- [ ] `r.Update(ctx, mcpServer)` at line 751 (restart annotation) is replaced with `r.Patch(ctx, mcpServer, client.MergeFrom(original))` using a deep copy taken before annotation mutation
- [ ] A `.claude/rules/operator.md` file is added to the toolhive repo root documenting the "no `r.Update` on MCPServer" rule with correct pattern examples
- [ ] A `lint-mcpserver-update` task is added to the operator Taskfile (`cmd/thv-operator/Taskfile.yml`) that fails if `r.Update(ctx, mcpServer)` or `r.Update(ctx, mcpServer,` appears in `cmd/thv-operator/controllers/*.go` (but allows `r.Status().Update` and `r.Update` on other resources)
- [ ] The `lint-mcpserver-update` target is wired into the `operator-ci.yml` workflow
- [ ] Existing operator unit and integration tests pass without modification
- [ ] All tests pass

For each of the 3 call sites, the pattern is identical:

1. Deep-copy the MCPServer object before mutation
2. Perform the mutation (add/remove finalizer, set annotation)
3. Patch using `client.MergeFrom(original)`

```go
// Before: full PUT clears unowned fields
controllerutil.RemoveFinalizer(mcpServer, "mcpserver.toolhive.stacklok.dev/finalizer")
err := r.Update(ctx, mcpServer)

// After: patch sends only the diff
original := mcpServer.DeepCopy()
controllerutil.RemoveFinalizer(mcpServer, "mcpserver.toolhive.stacklok.dev/finalizer")
err := r.Patch(ctx, mcpServer, client.MergeFrom(original))
```

The same pattern applies to the finalizer-add site (line 294) and the restart-annotation site (line 751).

For the lint target, use `grep` to detect the forbidden pattern while excluding `r.Status().Update` (which is safe -- status is a subresource with separate field ownership):

```yaml
lint-mcpserver-update:
  desc: Ensure r.Update is not used on MCPServer objects in controller code
  cmds:
    - |
      if grep -n 'r\.Update(ctx, mcpServer' cmd/thv-operator/controllers/*.go; then
        echo "ERROR: r.Update(ctx, mcpServer) found. Use r.Patch(ctx, mcpServer, client.MergeFrom(original)) instead."
        exit 1
      fi
```

### Patterns & Frameworks

- **controller-runtime `client.MergeFrom`**: The standard strategic-merge-patch helper. Produces a JSON merge patch from the diff between the original and mutated objects. Used by virtually all production controllers that share field ownership.
- **`DeepCopy()` before mutation**: Generated by controller-gen as `DeepCopy()` on all CRD types. Required to capture the pre-mutation state for the patch base.
- **Taskfile lint targets**: ToolHive uses `Taskfile.yml` for all automation. The lint target follows the existing pattern of simple `grep`-based checks (similar to how `generate-crds` checks for uncommitted diff).

### Code Pointers

- `cmd/thv-operator/controllers/mcpserver_controller.go` -- The primary file to modify. Contains the 3 `r.Update(ctx, mcpServer)` call sites at lines 283, 294, and 751.
- `cmd/thv-operator/api/v1alpha1/mcpserver_types.go` -- MCPServer CRD types. The `DeepCopy()` method is available via `zz_generated.deepcopy.go`.
- `cmd/thv-operator/Taskfile.yml` -- Add the `lint-mcpserver-update` task here alongside existing operator tasks.
- `.github/workflows/operator-ci.yml` -- Wire the new lint task into CI. Currently runs `operator-test`, `operator-test-integration`, `build-operator`, `generate-crds`, and `generate-crd-docs` jobs.
- `cmd/thv-operator/test-integration/mcp-server/mcpserver_controller_integration_test.go` -- Existing envtest integration tests for the MCPServer controller. These should pass without modification after the change (patch semantics are transparent to the test assertions).

### Component Interfaces

No new interfaces or types are introduced. The change is purely mechanical -- replacing `r.Update` calls with `r.Patch` calls using the existing `client.MergeFrom` helper from controller-runtime.

```go
// Existing controller-runtime API used:
// client.MergeFrom(obj) returns a client.Patch that computes a
// strategic merge patch between obj (original) and the object
// passed to r.Patch().

// r.Patch signature (already available via client.Client embed):
func (c Client) Patch(ctx context.Context, obj Object, patch Patch, opts ...PatchOption) error
```

### Testing Strategy

**Unit Tests**
- [ ] No new unit tests required -- the behavior is transparent to callers. The existing test suite validates that finalizers are added/removed and annotations are set correctly; these assertions are agnostic to whether the persistence is a PUT or a PATCH.

**Integration Tests**
- [ ] Verify existing envtest integration tests in `cmd/thv-operator/test-integration/mcp-server/` pass without modification. These tests exercise the full reconcile loop including finalizer and restart flows.

**Manual Verification**
- [ ] Deploy the operator to a Kind cluster (`task operator-deploy-local`), create an MCPServer, manually set `spec.authzConfig` via `kubectl patch`, trigger a reconcile, and confirm `spec.authzConfig` is preserved.

**CI Lint**
- [ ] Verify the `lint-mcpserver-update` task exits 0 on the patched code and exits non-zero if an `r.Update(ctx, mcpServer)` is reintroduced.

## Out of Scope

- Replacing `r.Update` calls targeting `deployment` or `service` resources (lines 398, 460, 479, 805) -- these are separate resources with no shared field ownership concern
- Full SSA migration of the MCPServer controller (converting all updates to server-side apply) -- tracked as separate future work
- Replacing `r.Status().Update` calls -- status is a subresource with separate ownership; `r.Status().Update` does not clobber spec fields
- Changes to the MCPServer CRD types or spec

## Related Future Work

Consolidated from #4633 (closed as duplicate of this issue for the spec-side scope):

- **Broader `r.Status().Update` → `r.Status().Patch` migration** across all controllers (every controller file under `cmd/thv-operator/controllers/`; 35 `Status().Update` call sites in `mcpserver_controller.go` alone; zero `Status().Patch` calls anywhere). `r.Status().Update` replaces the entire status subresource, so concurrent writers overwrite each other. This is especially dangerous for VirtualMCPServer where the runtime process writes `status.discoveredBackends`, `status.backendCount`, and `BackendsDiscovered` while the controller writes infrastructure conditions.
  - Recommended ordering: VirtualMCPServer first (highest risk due to concurrent runtime writes), then MCPServer (highest volume), then remaining controllers.
  - Out of scope for this issue — track as a separate follow-up.
- Related: #4630 (batch MCPServer status updates to reduce API calls per reconciliation).

## References

- controller-runtime `client.MergeFrom` documentation: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#MergeFrom
- Kubernetes API conventions on strategic merge patch: https://kubernetes.io/docs/reference/using-api/server-side-apply/

jhrozek:
## Summary

- The controller writes finalizers, finalizer removal, and the restart-processed annotation via `r.Update`. `Update` is a full PUT, so any spec field the operator does not track — most importantly `spec.authzConfig`, which an external operator will soon own via server-side apply — is zeroed on every reconcile.
- Replace the three `Update` call sites with an optimistic-lock merge patch (`client.MergeFromWithOptions(orig, client.MergeFromWithOptimisticLock{})`). The merge-patch body carries only fields the caller changed, so untouched fields never hit the wire and cannot be clobbered. `MergeFromWithOptimisticLock` sends `resourceVersion` as a precondition, giving 409-on-collision semantics for concurrent writers and defending `metadata.finalizers` (which has no array-merge semantics under merge-patch) against wholesale replacement when another controller is mid-flight adding its own entry.
- First in a 12-PR migration track. PR 2+ will migrate status writes under a separate helper (#4633).

Fixes #4767.

## Type of change

- [x] Bug fix

## Test plan

**Unit tests**: `cmd/thv-operator/controllers/mcpserver_spec_patch_test.go` uses a patch-recording client wrapper to assert that each of the three migrated call sites (AddFinalizer, RemoveFinalizer, RestartAnnotation) emits a merge-patch body carrying the `resourceVersion` precondition — a deterministic wire-level signal that `MergeFromWithOptimisticLock` is in effect. A regression to plain `MergeFrom` would drop the precondition and fail the assertion independent of the higher-level survival test.

**Envtest integration**: `cmd/thv-operator/test-integration/mcp-server/mcpserver_spec_patch_integration_test.go` creates an MCPServer, writes `spec.authzConfig` out-of-band from a second client, and asserts the field survives both the finalizer-add reconcile and the restart-annotation reconcile. Run via `task operator-test-integration`.

**Manual**: `kubectl apply` an MCPServer, `kubectl patch --type=merge` to set `spec.authzConfig`, wait for reconcile, `kubectl get mcpserver -o yaml` — `spec.authzConfig` persists across reconciles.

No.

## Implementation plan

<details>

First PR of the `r.Update` → `r.Patch` migration tracked in #4767 and #4633. This PR covers Track A (#4767): the three MCPServer spec writes that must move to an optimistic-lock merge patch before an external operator starts writing `spec.authzConfig` via SSA. Status-subresource migration (#4633) is the separate Track B, starting in PR 2.

Key design decisions:

- Use `MergeFromWithOptimisticLock{}` for MCPServer spec patches — preserves conflict-detection parity with `Update`, forces requeue on concurrent SSA instead of silent clobber, defends against `metadata.finalizers` array replacement.
- Keep spec-patch migration inline (3 sites) — a helper will land for status (89 sites) in PR 2 of Track B.
- Test strategy: envtest (real apiserver) for field-survival assertions; patch-recording fake client for wire-level optimistic-lock assertions. Two independent Kubernetes-go reviewers concurred that in-cluster (chainsaw) tests add no unique signal for patch semantics (100% apiserver-side) over envtest.

</details>

- The three call sites are at `cmd/thv-operator/controllers/mcpserver_controller.go:196` (RemoveFinalizer), `:212` (AddFinalizer), and `:768` (restart annotation). Each follows the same `DeepCopy → mutate → Patch(MergeFromWithOptions)` idiom.
- `mcpserver_restart_test.go` renames the mock-client flag `failOnMCPServerUpdate` → `failOnMCPServerWrite` because it now intercepts both `Update` and `Patch` on MCPServer.
- A short "Spec / metadata patching" section was added to `.claude/rules/operator.md` documenting the pattern for future CR writes.
- Expect 409 Conflict reconciles to appear as routine log noise once external SSA writers land in a cluster — the optimistic-lock guard doing its job, not a bug.


rdimitrov (dismissed):

.claude/rules/operator.md
```diff
@@ -55,3 +55,36 @@ task operator-test            # Run unit tests
 task operator-e2e-test        # Run e2e tests
 task crdref-gen              # Generate CRD API docs (run from cmd/thv-operator/)
 ```
+
+## Spec / metadata patching
```
ChrisJBurns, line 71: ❤️

ChrisJBurns (commented):
**Agents consulted**: kubernetes-expert, go-expert-developer, code-reviewer, toolhive-expert

### Consensus Summary
| 2 | Other MCPServer-writing reconcilers still Update spec | 8/10 | HIGH | Discuss scope |
| 3 | Verbatim 4-line rationale comment duplicated 3× → helper | 7/10 | MEDIUM | Fix |
| 4 | Hardcoded finalizer string; other CRDs have exported constants | 7/10 | MEDIUM | Fix |
| 5 | Rule text misstates *why* optimistic lock defends finalizers | 7/10 | MEDIUM | Fix |
| 6 | envtest finalizer-add case can degenerate into a no-op | 7/10 | LOW | Fix |
| 7 | `patchRecordingClient` silently discards `patch.Data()` errors | 7/10 | LOW | Fix |
| 8 | envtest `cleanupServer` swallows errors; uses non-lock MergeFrom | 7/10 | LOW | Fix |

### Overall

The controller migration itself is correct: `DeepCopy → mutate → Patch(MergeFromWithOptions(orig, MergeFromWithOptimisticLock{}))` is the right tool for the "external controller owns a disjoint spec field" problem, the finalizer-defense reasoning is sound, and `MergeFromWithOptimisticLock` strictly improves on `Update` for this use case.

The PR scope closes the clobber hazard in `mcpserver_controller.go`, but `toolconfig_controller.go:167` and `mcpexternalauthconfig_controller.go:186` still call `r.Update(ctx, &server)` on MCPServer — so an external SSA writer of `spec.authzConfig` can still lose its write on any MCPToolConfig or MCPExternalAuthConfig reconcile.

The medium findings are polish: a DRY helper for the 3-site copy-paste pattern, an exported `MCPServerFinalizerName` constant to match every other CRD, and a correction to how the rule file explains the optimistic lock's role in defending finalizers.

### Documentation

The rule file addition in `.claude/rules/operator.md` is load-bearing for the convention being introduced. See finding \#5 — the "merge-patch has no array-merge semantics" paragraph conflates two independent mechanisms. The optimistic lock does not change merge-patch's array-replacement behavior; it forces a requeue on staleness, which then triggers a fresh Get that observes the concurrently-added finalizer.

---
*Generated with Claude Code*

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -187,9 +187,14 @@ func (r *MCPServerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 
+			// Use an optimistic-lock merge patch rather than Update so we do not
+			// silently overwrite spec fields (e.g. spec.authzConfig) that are owned
+			// by another controller. The resourceVersion is sent, so concurrent
+			// writers cause a 409 Conflict and a clean requeue.
+			original := mcpServer.DeepCopy()
 			controllerutil.RemoveFinalizer(mcpServer, "mcpserver.toolhive.stacklok.dev/finalizer")
-			err := r.Update(ctx, mcpServer)
-			if err != nil {
+			if err := r.Patch(ctx, mcpServer,
```
ChrisJBurns, line 199: **[HIGH] Scope: other reconcilers still `Update` MCPServer spec (8/10)**

`toolconfig_controller.go:167` and `mcpexternalauthconfig_controller.go:186` both call `r.Update(ctx, &server)` on the MCPServer to stamp an annotation when their referenced config changes. An external SSA writer of `spec.authzConfig` can still lose its write on any MCPToolConfig or MCPExternalAuthConfig reconcile. The PR title says "Patch MCPServer spec instead of Update" but only closes the hazard in this file.

Either include those two sites in this PR (same `DeepCopy → set annotation → Patch` pattern) or acknowledge the remaining gap in the PR description / tracking issue \#4767.

*Raised by: toolhive-expert*

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -745,13 +755,18 @@ func (r *MCPServerReconciler) handleRestartAnnotation(ctx context.Context, mcpSe
...
+	// by another controller. The resourceVersion is sent, so concurrent
+	// writers cause a 409 Conflict and a clean requeue.
+	original := mcpServer.DeepCopy()
 	if mcpServer.Annotations == nil {
 		mcpServer.Annotations = make(map[string]string)
 	}
 	mcpServer.Annotations[LastProcessedRestartAnnotationKey] = currentRestartedAt
-	err = r.Update(ctx, mcpServer)
-	if err != nil {
+	if err := r.Patch(ctx, mcpServer,
```
ChrisJBurns, line 771: **[MEDIUM] DRY: 4-line rationale comment + idiom repeated 3× → helper (7/10)**

The identical 4-line comment and `DeepCopy → mutate → Patch(MergeFromWithOptions(orig, OptimisticLock{}))` idiom appear at L190, L206, and here at L768. Per `.claude/rules/go-style.md` ("Keep Comments Synchronized With Code"), the rationale now has three copies that can drift. The rule file defers a "shared helper" to \#4633 based on scale (89 status sites) — but three is already the canonical DRY threshold, and a closure-form helper makes the "don't put unrelated work between DeepCopy and Patch" admonition a structural guarantee instead of a review-time check.

```suggestion
	if err := r.patchWithOptimisticLock(ctx, mcpServer, func(s *mcpv1beta1.MCPServer) {
		if s.Annotations == nil {
			s.Annotations = make(map[string]string)
		}
		s.Annotations[LastProcessedRestartAnnotationKey] = currentRestartedAt
	}); err != nil {
```

With the helper defined once near the top of the file:
```go
func (r *MCPServerReconciler) patchWithOptimisticLock(
    ctx context.Context, obj *mcpv1beta1.MCPServer, mutate func(*mcpv1beta1.MCPServer),
) error {
    original := obj.DeepCopy()
    mutate(obj)
    return r.Patch(ctx, obj, client.MergeFromWithOptions(original, client.MergeFromWithOptimisticLock{}))
}
```

*Raised by: code-reviewer, kubernetes-expert*

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -187,9 +187,14 @@ func (r *MCPServerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 				return ctrl.Result{}, err
 			}
 
+			// Use an optimistic-lock merge patch rather than Update so we do not
+			// silently overwrite spec fields (e.g. spec.authzConfig) that are owned
+			// by another controller. The resourceVersion is sent, so concurrent
+			// writers cause a 409 Conflict and a clean requeue.
+			original := mcpServer.DeepCopy()
 			controllerutil.RemoveFinalizer(mcpServer, "mcpserver.toolhive.stacklok.dev/finalizer")
```
ChrisJBurns: **[MEDIUM] Magic finalizer string — no exported constant (7/10)**

`"mcpserver.toolhive.stacklok.dev/finalizer"` appears 4× in this controller (L185, L190/L195, L200, L201) and is redeclared as an inline `const finalizerName` in both new test files. Every other CRD has an exported constant (`MCPGroupFinalizerName`, `ToolConfigFinalizerName`, `OIDCConfigFinalizerName`, `TelemetryConfigFinalizerName`, `ExternalAuthConfigFinalizerName`, `embeddingFinalizerName`). MCPServer is the outlier and this PR re-entrenches the pattern; the rule-file example also uses the magic string literally.

Add `const MCPServerFinalizerName = "mcpserver.toolhive.stacklok.dev/finalizer"` near the top of this file, replace the four call sites, and reference it from the tests and the rule example.

*Raised by: code-reviewer*

.claude/rules/operator.md
```diff
@@ -55,3 +55,36 @@ task operator-test            # Run unit tests
...
+Never use `r.Update` on a CR spec or metadata: `Update` is a full PUT,
+so any field our local copy does not track (e.g. `spec.authzConfig`
+written by a separate authorization controller) gets zeroed on every
+reconcile.
+
+Use an **optimistic-lock merge patch** instead. The merge-patch body
+only contains fields the caller changed, and `MergeFromWithOptimisticLock`
+sends `resourceVersion` as a precondition so concurrent writers get a
+409 and a clean requeue. This also defends `metadata.finalizers`:
+merge-patch has no array-merge semantics, so without the lock a write
```
ChrisJBurns: **[MEDIUM] Misstates *why* optimistic lock defends finalizers (7/10)**

"Merge-patch has no array-merge semantics, so without the lock a write including the finalizers array would replace it wholesale." — This conflates two facts. Merge-patch replaces arrays wholesale **regardless** of whether optimistic lock is set; the lock doesn't change that. What the lock actually does is detect staleness on `resourceVersion`, forcing a requeue → fresh Get → recompute-diff cycle, and the fresh Get is what observes any concurrently-added finalizer.

Consider rewriting to separate the two concerns: (1) merge-patch replaces arrays wholesale, so the patch body must be computed from a fresh Get; (2) `MergeFromWithOptimisticLock` fails the patch with a 409 if server state moved between Get and Patch, triggering requeue → fresh Get. Optional: note SSA with a distinct field-manager is the long-term answer.

*Raised by: kubernetes-expert, code-reviewer*

cmd/thv-operator/test-integration/mcp-server/mcpserver_spec_patch_integration_test.go
```diff
@@ -0,0 +1,200 @@
...
+			original := fresh.DeepCopy()
+			fresh.Finalizers = nil
+			_ = k8sClient.Patch(ctx, fresh, client.MergeFrom(original))
+		}
+		_ = k8sClient.Delete(ctx, fresh)
+	}
+
+	Context("When a second actor writes spec.authzConfig out-of-band", func() {
+		It("Should preserve spec.authzConfig across the finalizer-add reconcile", func() {
+			// Step 1: create the MCPServer.
```
ChrisJBurns: **[LOW] envtest finalizer-add case can degenerate into a no-op (7/10)**

The comment acknowledges a race: the test tries to either beat the controller's first reconcile or observe survival across "the next reconcile that mutates metadata." But in the happy path, after finalizer-add, no subsequent reconciles mutate metadata (only Status / owned resources). If the controller wins the race, this spec degenerates into "Eventually wrote authzConfig, still there" and does not exercise the optimistic-lock patch at all.

Either downgrade/remove this case (the unit test `TestMCPServerSpecPatchesAreOptimisticLock` already covers it deterministically), or block GC with a second finalizer and drive a deterministic RemoveFinalizer path.

*Raised by: kubernetes-expert*

cmd/thv-operator/controllers/mcpserver_spec_patch_test.go
```diff
@@ -0,0 +1,183 @@
...
+type recordedPatch struct {
+	obj  client.Object
+	body string
+}
+
+func (c *patchRecordingClient) Patch(
+	ctx context.Context, obj client.Object, patch client.Patch, opts ...client.PatchOption,
+) error {
+	if data, err := patch.Data(obj); err == nil {
+		c.mu.Lock()
```
ChrisJBurns, line 53: **[LOW] `patchRecordingClient.Patch` silently drops `patch.Data()` errors (7/10)**

`if data, err := patch.Data(obj); err == nil { ... }` drops the error. A marshal failure produces empty `patches` → `lastMCPServerPatchBody() == ""` → `require.NotEmpty` fires with a misleading "no MCPServer Patch was recorded" message that masks a real marshal bug. Violates `.claude/rules/go-style.md`'s "Comment ignored errors" rule.

Capture the error on the recorder (e.g., `c.dataErr = err`) and surface it from `lastMCPServerPatchBody`, or `t.Fatalf` via a test handle, so a marshal failure produces a precise test failure rather than an empty-body one.

*Raised by: kubernetes-expert, go-expert-developer*

cmd/thv-operator/test-integration/mcp-server/mcpserver_spec_patch_integration_test.go
```diff
@@ -0,0 +1,200 @@
...
+		return &mcpv1beta1.AuthzConfigRef{
+			Type: mcpv1beta1.AuthzConfigTypeConfigMap,
+			ConfigMap: &mcpv1beta1.ConfigMapAuthzRef{
+				Name: cmName,
+				Key:  "authz.json",
+			},
+		}
+	}
+
+	// newMCPServer returns a minimal stdio MCPServer used as a starting point
```
ChrisJBurns, line 59: **[LOW] `cleanupServer` swallows errors; uses non-lock MergeFrom (7/10)**

`_ = k8sClient.Patch(...)` and `_ = k8sClient.Delete(...)` silently drop teardown errors. With fixed fixture names in the shared `default` namespace, a leaked object collides with the next run as `AlreadyExists` — confusing trace. Separately, the cleanup uses `client.MergeFrom(original)` (no optimistic lock) — test-safe but a copy-paste hazard given this PR is codifying the "always use optimistic-lock merge patches on CRs" convention.

Log teardown errors via `GinkgoWriter.Printf`, or use per-spec namespaces. Either use `MergeFromWithOptions(..., OptimisticLock{})` for consistency with the rule being introduced, or add a comment like `// Test-only teardown: no concurrent writers; optimistic lock unnecessary.`

*Raised by: kubernetes-expert, go-expert-developer*

cmd/thv-operator/controllers/mcpserver_controller.go
jhrozek: Addressed. Migrated both sites to the optimistic-lock merge-patch pattern: `toolconfig_controller.go` (annotation-stamp loop tagging referencing MCPServers with `toolconfig-hash`) and `mcpexternalauthconfig_controller.go` (same pattern for `externalauthconfig-hash`). Full sweep confirmed these were the only two remaining `r.Update` sites on MCPServer outside `mcpserver_controller.go`. Will also update #4767 to drop the "only `mcpserver_controller.go`" framing.

cmd/thv-operator/controllers/mcpserver_controller.go
jhrozek: Agree the 5× duplication (now 5 after expanding to the two cross-controller sites from your scope finding) is a real drift risk. Deferring the helper extraction to a follow-up so this PR stays focused on unblocking the external writer for `spec.authzConfig`.

The helper design has cross-cutting implications worth doing deliberately: after the scope expansion, the sites now span three reconciler types (`MCPServerReconciler`, `ToolConfigReconciler`, `MCPExternalAuthConfigReconciler`), so a receiver-based helper on `MCPServerReconciler` doesn't fit. A free function with a `client.Object` constraint would, but that's the same shape #4633 needs for the 89 status sites — designing it once there avoids a near-term rewrite.

Tracked as a follow-up on my side.

cmd/thv-operator/controllers/mcpserver_controller.go
jhrozek: Addressed. Added `const MCPServerFinalizerName = "mcpserver.toolhive.stacklok.dev/finalizer"` at `mcpserver_controller.go:109`, replaced all 4 controller call sites plus the inline `finalizerName` consts in both new test files. The rule file example in `operator.md` also uses the constant now. MCPServer is no longer the outlier vs. the other CRDs' finalizer conventions.

.claude/rules/operator.md
jhrozek: Agreed and rewritten. The paragraph now separates the two mechanisms cleanly: (1) `MergeFromWithOptimisticLock` stamps `resourceVersion` as a precondition → 409 + requeue with a fresh Get; (2) merge-patch always replaces arrays wholesale (independent of the lock), so when the diff includes `finalizers` it must have been computed from an up-to-date snapshot — which is what the 409+requeue guarantees. Skipped the SSA mention to keep the rule tight.

cmd/thv-operator/test-integration/mcp-server/mcpserver_spec_patch_integration_test.go
jhrozek: Agreed. Removed the finalizer-add envtest case — the race made it degenerate into "wrote authzConfig, still there" whenever the controller won, which passes regardless of patch semantics. The unit test's `AddFinalizer` table row covers that migration site deterministically via the patch-recording client. Retained the restart-annotation case (deterministic: waits for finalizer, writes `authzConfig`, triggers `handleRestartAnnotation`, waits for `LastProcessedRestartAnnotationKey`). Lifted the finalizer-path explanation into the file header, pointing at `TestMCPServerSpecPatchesAreOptimisticLock` for the deterministic coverage.

cmd/thv-operator/controllers/mcpserver_spec_patch_test.go
jhrozek: Kept the minimal form: added a 4-line justification comment at the ignored error rather than plumbing a `dataErr` field through the recorder. `patch.Data` on a typed `*MCPServer` is `json.Marshal` of a generated struct — unreachable in practice, and a real failure would break the production controller's own patch and fire other assertions first. The comment satisfies the "Comment ignored errors" style rule without adding a new test-handle concept. Happy to switch to the recorder-plumbing form if you feel strongly.

cmd/thv-operator/test-integration/mcp-server/mcpserver_spec_patch_integration_test.go
jhrozek: Both halves addressed: teardown errors now surface via `GinkgoWriter.Printf` (both the finalizer-strip patch and the delete) so failures are visible in test output without flunking the spec; added a comment on the non-lock `MergeFrom` explaining it's deliberately plain (no concurrent writers in teardown) and pointing at `operator.md` to prevent copy-paste. Skipped switching teardown to the optimistic lock — it would "match the rule" but semantically adds nothing with no concurrent writers, and would obscure the fact that the pattern doesn't always apply. Skipped per-spec namespaces — envtest gives a fresh API server per `go test` invocation, so cross-run leakage isn't a real risk.

ChrisJBurns (approved):

rdimitrov (approved):

PR #4933, opened 2026-04-20
vmcp: allow embedders to inject custom session DataStorage

### LINKED ISSUE #4928: vmcp: allow embedders to plug a custom session DataStorage into server.Server
## Problem

`pkg/vmcp/server.Server` accepts a `SessionStorage *vmcpconfig.SessionStorageConfig` whose `Provider` field is enum-validated at `memory` or `redis`. The internal helper `buildSessionDataStorage` (`pkg/vmcp/server/server.go:263-291`) hardcodes `transportsession.NewLocalSessionDataStorage` or `transportsession.NewRedisSessionDataStorage` based on that enum. There is no way for an embedder to inject a `transportsession.DataStorage` implementation of its own — the interface is public but the server has no seam to accept one.

## Use case

Deployments that already run an operational datastore other than Redis (Postgres, DynamoDB, Spanner, MongoDB, etc.) can satisfy `transportsession.DataStorage` against that datastore to avoid adding Redis purely for MCP session persistence. Today they have to either fork `server.go` or fall back to in-memory sessions (no persistence across pod restarts).

## Proposed API

Add a `DataStorage transportsession.DataStorage` field to `server.Config` (or a functional-option equivalent). When set, `buildSessionDataStorage` short-circuits and returns the caller-supplied implementation; when nil, existing behaviour is preserved. Backwards-compatible.

Alternative: add `server.WithDataStorage(ds transportsession.DataStorage) Option` and thread it through `New(...)`.

## Scope

- Purely additive.
- No change to `transportsession.DataStorage` itself — the contract is already public and stable.
- No change to the `memory` / `redis` fast paths.

JAORMX:
## Summary

- **Why:** `pkg/vmcp/server.Server` currently forces embedders to pick between the two providers encoded in `SessionStorageConfig` — `memory` or `redis`. `buildSessionDataStorage` hardcodes `NewLocalSessionDataStorage` / `NewRedisSessionDataStorage` off that enum. Embedders that already operate other datastores (Postgres, DynamoDB, Spanner, MongoDB, etc.) either have to fork `server.go` or fall back to in-memory sessions and lose cross-pod persistence.
- **What:** Add an optional `Config.DataStorage transportsession.DataStorage` field. When non-nil, the server uses the caller-supplied store directly and `SessionStorage` is ignored in its entirety. Setting both is rejected at `New()` so misconfiguration surfaces loudly instead of silently favouring one.
- **Lifecycle:** the caller owns the injected store — the server never calls `Close()` on it. This matches the convention every other caller-supplied dependency on `Config` already follows (`TelemetryProvider`, `StatusReporter`, `Watcher`, `OptimizerFactory`). `buildSessionDataStorage` now returns an ownership-aware closer so the server-owned path still closes cleanly while the caller-owned path is left untouched on every exit (New error, Start error, Stop).
- **Interface:** no change to `transportsession.DataStorage`. No change to the `memory` / `redis` fast paths. Purely additive.

Closes #4928

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — full `pkg/vmcp/server/...` suite green (pre-existing flakes in `pkg/auth/oauth` and `pkg/transport/proxy/transparent` are unrelated to this change)
- [x] Manual testing (see below)

Three new unit tests in `pkg/vmcp/server/datastorage_injection_test.go`:

1. `TestNew_CallerOwnedDataStorageNotClosedOnStop` — constructs `Server` with a counting spy wrapping `LocalSessionDataStorage`, runs `Stop()`, asserts `closeCalls == 0`. This is the load-bearing contract assertion.
2. `TestNew_BothSessionStorageAndDataStorageErrors` — asserts `New()` returns an error mentioning both fields when `SessionStorage.Provider="redis"` and `DataStorage` are set simultaneously; also asserts the spy is still untouched on the rejection path.
3. `TestNew_ServerBuiltDataStorageStopSucceeds` — smoke-tests the unchanged server-owned lifecycle with `memory` provider, guarding against accidental regression when the close call moved from an inline block in `Stop()` onto `sessionDataStorageCloser`.

Manual: existing CLI flow (`thv vmcp serve -c <memory-session-storage-config>`) is unchanged; `pkg/vmcp/cli/serve.go` never sets `Config.DataStorage`, so every existing deployment path is bit-identical.

Yes — additive API on `pkg/vmcp/server.Config`. Embedders integrating vMCP as a library can now supply their own `transportsession.DataStorage` implementation. No CLI flag or config-file change; `thv vmcp serve` behaviour is unchanged.

## Implementation plan

<details>

### Context

`pkg/vmcp/server.Server` only supports `memory` or `redis` for session data storage via `SessionStorageConfig.Provider`. The `transportsession.DataStorage` interface is already public and stable, so the right fix is a narrow, purely additive seam: let the embedder pass their own `DataStorage` at server construction.

### Design Decisions (validated with an independent Plan agent)

1. **`Config` struct field, not a functional option.** Every other caller-supplied dependency on `server.Config` is a field. `.claude/rules/go-style.md` explicitly blesses the field pattern as compile-time safe.
2. **Caller owns the lifecycle.** Matches the convention for every other caller-supplied dependency on `Config`: server never calls `Close()`/`Shutdown()` on them.
3. **Closer return from `buildSessionDataStorage`** rather than a `bool` on `Server`, mirroring the existing `optimizerCleanup` style. Caller-supplied store → nil closer. Server-built store → real closer.
4. **Fail loudly on ambiguous config.** If both `Config.DataStorage` and a non-empty `Config.SessionStorage.Provider` are set, `New()` returns an error. Allowing silent precedence would leave Redis config looking live but ignored — a diagnosis nightmare.
5. **No changes to the `DataStorage` interface itself.** No changes to `SessionStorageConfig` or its validator. Purely additive.

### Changes

**`pkg/vmcp/server/server.go`**

- Add `DataStorage transportsession.DataStorage` to `Config` with a doc block covering mutual exclusivity with `SessionStorage`, caller-owned lifecycle (including Close ordering relative to `Server.Stop()`), that `cfg.SessionTTL` does not flow into a caller-supplied store, and that the store holds HMAC-hashed token material.
- Change `buildSessionDataStorage` signature to `(DataStorage, func(context.Context) error, error)`. Ambiguous config → error. Caller-supplied → `(store, nil, nil)`. Server-built → `(store, closerFor(store), nil)`.
- Add `Server.sessionDataStorageCloser func(context.Context) error`. Set once in `New()`; invoked once in `Stop()`; nil for caller-owned stores.
- Rework the close-on-error `defer` inside `New()` to gate on `sessionDataStorageCloser != nil` so caller-supplied stores are untouched even on construction failure.
- Rewire the end of `Stop()` to use the closer instead of calling `Close()` on the raw store, preserving the "close data storage last" ordering (after session manager and discovery manager have stopped).

**`pkg/vmcp/server/datastorage_injection_test.go`** (new)

- `countingDataStorage` — 10-line spy wrapping `LocalSessionDataStorage` with an atomic Close counter.
- Three tests (see Test plan above).

### Files NOT modified

- `pkg/transport/session/session_data_storage.go` — interface is stable.
- `pkg/vmcp/config/config.go` / `validator.go` — no config surface change (validator gap for `SessionStorageConfig.Provider` is a separate concern, tracked for a follow-up PR per the one-logical-change rule).
- `pkg/vmcp/cli/serve.go` — the CLI never sets `DataStorage`; behaviour preserved.
- `test/integration/vmcp/helpers/vmcp_server.go` — unchanged.
- `pkg/vmcp/server/sessionmanager/factory.go` — treats `DataStorage` as an opaque dependency.

### Review agents run before pushing

- **code-reviewer**: no blockers; suggested doc tweak around `SessionTTL`-not-applying and test-name clarity — both applied.
- **toolhive-expert**: verified architectural consistency with every other caller-supplied dependency on `Config`; verified `Stop()` ordering; no regression in CLI or operator paths.
- **security-advisor**: confirmed the change adds no data-leakage path, no auth-middleware bypass, and no new credential surface. Recommended doc hardening warning embedders that the store holds HMAC token material and must not leak credentials through `Close()` errors — applied.
- **vmcp-review skill**: clean against all 10 vMCP anti-patterns; noted the Server god-object and Config bloat as pre-existing conditions not expanded by this change.

</details>

### DIFF
--- a/pkg/vmcp/server/server.go
+++ b/pkg/vmcp/server/server.go
@@ -181,7 +181,38 @@ type Config struct {
 	// When provider is "redis", a Redis-backed store is created for cross-pod
 	// session persistence; the Redis password is read from the
 	// THV_SESSION_REDIS_PASSWORD environment variable.
+	//
+	// Mutually exclusive with DataStorage: setting both is rejected at New().
 	SessionStorage *vmcpconfig.SessionStorageConfig
+
+	// DataStorage optionally injects a caller-supplied session metadata store,
+	// bypassing the built-in memory/redis providers. When non-nil, the server
+	// uses this store as-is and SessionStorage is ignored in its entirety (no
+	// field of SessionStorage is consulted). Setting both DataStorage and a
+	// non-empty SessionStorage.Provider is rejected at New() as ambiguous
+	// configuration.
+	//
+	// Lifecycle: the caller owns it. The server does NOT call Close() on a
+	// caller-supplied DataStorage, even on error paths in New() or during
+	// Stop(). The caller is responsible for invoking Close() exactly once
+	// after Server.Stop() returns (not before — the session manager may issue
+	// final Update calls during Stop). The caller is likewise responsible for
+	// configuring the store's TTL; cfg.SessionTTL applies only to the
+	// transport-level session manager, not to the caller-supplied DataStorage.
+	//
+	// Sensitive material: the store holds HMAC-hashed token material and
+	// other session metadata. Embedders should treat the backing datastore as
+	// sensitive (dedicated credentials, encryption at rest, restricted read
+	// access). Implementations must not include credentials in Close() error
+	// messages — those errors are surfaced through Server.Stop().
+	//
+	// This seam lets embedders satisfy transportsession.DataStorage against
+	// datastores other than the built-in providers (e.g. Postgres, DynamoDB)
+	// without forking the server. It enables cross-replica session metadata
+	// sharing when backed by a shared store; it does NOT solve cross-replica
+	// message delivery — callers still need session affinity at the load
+	// balancer for streaming responses.
+	DataStorage transportsession.DataStorage
 }
 
 // Server is the Virtual MCP Server that aggregates multiple backends.
@@ -224,9 +255,15 @@ type Server struct {
 	sessionManager *transportsession.Manager
 
 	// sessionDataStorage is the pluggable key-value backend for session metadata.
-	// Currently always LocalSessionDataStorage (in-memory, single-process).
-	// Redis-backed storage for multi-pod deployments is not yet wired.
+	// It may be LocalSessionDataStorage (in-memory, single-process), a Redis-backed
+	// store, or a caller-supplied implementation injected via Config.DataStorage.
 	sessionDataStorage transportsession.DataStorage
+
+	// sessionDataStorageCloser closes sessionDataStorage on shutdown. It is
+	// set only when the server built the store itself (memory or redis
+	// providers). When Config.DataStorage was supplied by the caller, this is
+	// nil and the caller is responsible for closing the store.
+	sessionDataStorageCloser func(context.Context) error
 
 	// Capability adapter for converting aggregator types to SDK types
 	capabilityAdapter *adapter.CapabilityAdapter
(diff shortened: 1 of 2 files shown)


pkg/vmcp/server/server.go
```diff
@@ -180,7 +180,38 @@ type Config struct {
 	// When provider is "redis", a Redis-backed store is created for cross-pod
 	// session persistence; the Redis password is read from the
 	// THV_SESSION_REDIS_PASSWORD environment variable.
+	//
+	// Mutually exclusive with DataStorage: setting both is rejected at New().
```
yrobla, line 185: **nit / doc inconsistency (non-blocking):** The comment says "setting both is rejected" but the actual check in `buildSessionDataStorage` only rejects when `SessionStorage.Provider != ""`. An embedder who passes `&vmcpconfig.SessionStorageConfig{}` (struct with empty `Provider`) alongside `DataStorage` will *not* get an error — the server silently uses `DataStorage` and ignores the `SessionStorage` struct entirely.

This contradicts the `DataStorage` field's own comment a few lines down, which is more precise: *"Setting both DataStorage and a non-empty SessionStorage.Provider is rejected"*.

Suggested fix:
```suggestion
	// Mutually exclusive with DataStorage when Provider is non-empty: New()
	// returns an error if both Config.DataStorage and a non-empty
	// SessionStorage.Provider are set simultaneously.
```

Evidence: `buildSessionDataStorage` condition at `pkg/vmcp/server/server.go` line ~302:
```go
if cfg.DataStorage != nil {
    if cfg.SessionStorage != nil && cfg.SessionStorage.Provider != "" {
        return nil, nil, fmt.Errorf(...)
    }
    return cfg.DataStorage, nil, nil
}
```

yrobla (commented):
This is a clean, well-scoped addition. The lifecycle ownership model is consistent with every other caller-supplied dependency on `Config`, the error message on ambiguous configuration surfaces loudly, and the three new tests cover the load-bearing contracts well.

One doc-comment inconsistency worth tightening before merge.

yrobla (dismissed):

JAORMX:
@yrobla had to rebase

rdimitrov (dismissed):

JAORMX:
Closing this for now — deprioritizing this work. Thanks @yrobla and @rdimitrov for reviewing. Happy to reopen if the need for embedder-injected session DataStorage comes back up.

PR #4866, opened 2026-04-15
Treat unauthenticated backends as routable in vMCP health

lorr1:
## Summary

- vMCPs with OAuth-protected backends (e.g., `upstreamInject` auth) permanently reported `PhaseFailed` because health probes lack user tokens and get 401s, which `determinePhase()` counted as zero healthy backends. These backends are reachable and running — they just need per-request user auth.
- Introduce `Summary.Routable()` (healthy + unauthenticated) and use it for phase determination, `BackendCount`, and status messages across the health monitor, operator controller, and status collector. Give `BackendUnauthenticated` its own CRD status `"unauthenticated"` instead of mapping to `"unavailable"`.

Fixes #4824
Follow-up: #4920 (health probes cannot authenticate to OAuth-protected backends — tracked for when auth-server token support lands)

## Type of change

- [x] Bug fix

## Test plan

- [x] Unit tests (`task test`)
- [x] Linting (`task lint-fix`)

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/health/monitor.go` | Add `Summary.Routable()` method; update `determinePhase()`, `formatStatusMessage()`, `buildConditions()`, `BackendCount` to use routable count; add `pluralBackend()`, `pluralRequire()`, `quantifyBackends()` helpers; rename condition reasons `AllBackendsHealthy` → `AllBackendsRoutable`, `NoHealthyBackends` → `NoRoutableBackends` |
| `pkg/vmcp/types.go` | `ToCRDStatus()` maps `BackendUnauthenticated` to `"unauthenticated"` instead of `"unavailable"`; update `DiscoveredBackend.Status` doc |
| `cmd/thv-operator/api/v1alpha1/virtualmcpserver_types.go` | Add `BackendStatusUnauthenticated` constant; update `BackendCount` doc |
| `cmd/thv-operator/controllers/virtualmcpserver_controller.go` | `countBackendHealth` treats unauthenticated as routable |
| `cmd/thv-operator/pkg/virtualmcpserverstatus/collector.go` | Count unauthenticated backends in `BackendCount` |
| `pkg/vmcp/discovery/middleware.go` | TODO comment noting unauthenticated exclusion from discovery is intentional (references #4920) |
| `pkg/vmcp/server/status.go` | `/status` endpoint treats unauthenticated backends as routable; update Health field doc |
| `pkg/vmcp/health/status_builder_test.go` | Add 4 test cases for unauthenticated phase logic |
[...]

PR #4935, opened 2026-04-20
Treat 401/403 from auth-configured backends as healthy

### LINKED ISSUE #4920: Health probes cannot authenticate to OAuth-protected backends
## Problem

Health probes run as system-level background checks with no user context. When a backend requires per-request OAuth (e.g., `upstreamInject` auth), the probe gets a 401 and the backend is marked `BackendUnauthenticated`. This means:

- The health monitor cannot distinguish "backend is down" from "backend requires user auth"
- Discovery excludes unauthenticated backends from capability aggregation (it can't probe their tools/prompts/resources)
- The `/status` endpoint reports `healthy: false` for groups where all backends require OAuth

PR #4866 introduced `Routable()` (healthy + unauthenticated) as a workaround so these backends don't cause `PhaseFailed`, but the underlying limitation remains.

## Desired outcome

Health probes can authenticate to OAuth-protected backends, eliminating the `BackendUnauthenticated` state entirely. Every backend becomes truly healthy or unhealthy, and discovery can include all reachable backends in capability aggregation.

## Possible approach

Once the auth server is integrated with vMCP (stacklok-epics#198), the health monitor could request short-lived service-account tokens to probe OAuth backends. This would need design around: what identity the probe uses, what scopes it needs, and whether backends accept service-account tokens vs only user tokens.

## References

- #4824 — original bug (fixed by #4866)
- stacklok-epics#198 — auth server vMCP integration
- `pkg/vmcp/discovery/middleware.go:138` — TODO tracking this gap

JAORMX:
## Summary

- Health probes deliberately run with no user context, so per-user upstream tokens (`upstream_inject`) are not available at probe time. For OAuth-protected backends, a 401/403 challenge is the *expected* response and proves reachability, a working auth layer, and a working network/TLS path — so treat it as `BackendHealthy`.
- Narrow `BackendUnauthenticated` to a misconfiguration signal (backend demands auth but no outgoing auth strategy is configured on the backend target) and remove discovery's exclusion TODO — the state now accurately reflects "should not expose capabilities."
- Drop the `|| BackendUnauthenticated` clause from `/status`'s healthy computation since the state no longer indicates routability.

Closes #4920

## Type of change

- [x] Bug fix

## Changes

| File | Change |
|---|---|
| `pkg/vmcp/health/checker.go` | `categorizeError` now takes `target`; new `authErrorStatus` helper. `CheckHealth` returns `(BackendHealthy, nil)` on expected auth challenge so the monitor records success and the circuit breaker stays closed. |
| `pkg/vmcp/health/checker_test.go` | Full matrix: auth errors × {upstream_inject, token_exchange, header_injection, nil AuthConfig, StrategyTypeUnauthenticated, nil target} + non-auth errors unchanged. |
| `pkg/vmcp/types.go` | `BackendUnauthenticated` doc narrows to misconfig; `ToCRDStatus` comment updated. |
| `pkg/vmcp/discovery/middleware.go` | Removed `TODO(#4920)`; updated `filterHealthyBackends` comment — excluding `BackendUnauthenticated` from capability aggregation is now the correct behavior. |
| `pkg/vmcp/server/status.go` | `hasHealthyBackend` no longer counts `BackendUnauthenticated`. |
| `pkg/vmcp/health/monitor.go` | Stale "Authentication required" message updated to reflect new misconfig meaning. Doc-comment TODOs added on `Summary.Routable()`, `BuildStatus`, and `determinePhase` noting that PR #4866's routability logic should be reverted in a follow-up (needs operator controller + status collector changes — bigger blast radius). |

## Test plan

- [x] `task lint` — 0 issues
- [x] `task test` — full unit + race suite passes
- [x] New test matrix in `checker_test.go` covers all strategy-type × auth-error combinations plus non-auth errors
- [x] Code review pass (anti-patterns, go-style, security) — no blocking findings

Yes. vMCP deployments where every backend uses per-user OAuth (e.g., `upstream_inject` with Google Drive, Atlassian) will now report those backends as \`healthy\` in \`/status\` and include them in capability aggregation once an authenticated user session runs discovery. Before this fix (even with the PR #4866 workaround), such backends were excluded from capability aggregation and their tools did not appear in aggregated \`tools/list\`. `BackendUnauthenticated` is now reserved for a narrower misconfig case and no longer counts as "healthy-enough" in \`/status\`.

Intentionally narrow scope. Explicitly deferred to follow-up PRs:

1. **Revert PR #4866's `Routable() = Healthy + Unauthenticated` semantics** in \`monitor.go\` and the mirrored logic in \`cmd/thv-operator/controllers/virtualmcpserver_controller.go\` and \`cmd/thv-operator/pkg/virtualmcpserverstatus/collector.go\`. Flagged with TODO comments.
2. **`WWW-Authenticate` header validation** before treating 401 as healthy — would require plumbing HTTP response headers through the error chain.
3. **User-opportunistic capability cache** so `/status`-style readers can see tools before the first user session runs discovery.
4. **Opt-in `probe_credentials` strategy** for deployments that *do* have a service-account credential (Workspace, GitHub App) and want full-path probe validation.

## Implementation plan

<details>
<summary>Planned with Claude Code</summary>

The fix was designed after a multi-agent discussion:

1. **Health probe research** confirmed probes go through the same outgoing auth strategies as real requests, with each strategy short-circuiting on \`healthcontext.IsHealthCheck(ctx)\`. \`upstream_inject\` returns no header (no user identity); \`token_exchange\` has a partial client-credentials fallback only when credentials are configured.
2. **Key insight** (from the user): for \`upstream_inject\`, vMCP holds *no* credentials at startup — upstream tokens are per-user, populated only after each user completes their own OAuth (\`pkg/auth/identity.go:47-66\`, \`pkg/auth/token.go:1185-1201\`). \`BackendUnauthenticated\` at startup is the *correct* system state; the gateway has nothing to authenticate with.
3. **Design conclusion**: drop probe-auth, treat 401/403 as healthy if the backend has an outgoing auth strategy configured. RFC 6749 §5.2 / RFC 6750 §3.1 make 401 invalid_token the correct response from a properly-configured resource server — treating it as "reachable and RFC-compliant" is semantically honest.
4. **Code review** flagged the stale "Authentication required" status message and the \`server/status.go\` inconsistency as should-fix-in-this-PR; both addressed.

</details>

### DIFF
--- a/pkg/vmcp/discovery/middleware.go
+++ b/pkg/vmcp/discovery/middleware.go
@@ -135,9 +135,13 @@ func Middleware(
 // or degraded. Backends that are unhealthy, unknown, or unauthenticated are excluded
 // from capability aggregation to prevent exposing tools from unavailable backends.
 //
-// TODO(#4920): Unauthenticated backends are treated as routable for phase determination
-// but are excluded here because discovery probes cannot carry user tokens. If health
-// probes could authenticate, these backends would be fully healthy and included here.
+// A note on BackendUnauthenticated: a 401/403 from a backend that has an outgoing
+// auth strategy configured is treated as BackendHealthy by the health checker
+// (health probes deliberately do not carry user credentials — the challenge proves
+// reachability). BackendUnauthenticated therefore indicates a misconfiguration:
+// the backend requires authentication but no outgoing auth strategy is configured
+// on the backend target. Excluding such backends from capability aggregation is
+// the correct behavior — their capabilities cannot be safely exposed.
 //
 // Health status filtering:
 //   - healthy: included (fully operational)
@@ -145,7 +149,7 @@ func Middleware(
 //   - empty/zero-value: included (assume healthy when health monitoring is disabled)
 //   - unhealthy: excluded (not responding, circuit breaker may be open)
 //   - unknown: excluded (status not yet determined)
-//   - unauthenticated: excluded (authentication failed)
+//   - unauthenticated: excluded (misconfiguration: backend requires auth but none configured)
 //
 // When healthStatusProvider is provided, the current health status from the health
 // monitor is used (respects circuit breaker state). When nil, falls back to the
--- a/pkg/vmcp/health/checker.go
+++ b/pkg/vmcp/health/checker.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"github.com/stacklok/toolhive/pkg/vmcp"
+	authtypes "github.com/stacklok/toolhive/pkg/vmcp/auth/types"
 )
 
 // healthChecker implements vmcp.HealthChecker using ListCapabilities as the health check.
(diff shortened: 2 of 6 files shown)


amirejaz (approved):

JAORMX:
Had to rebase

jhrozek (approved):

PR #4942, opened 2026-04-20
Introduce MutateAndPatchStatus helper for status writes

jhrozek:
## Summary

- The operator's status writes use `r.Status().Update`, which sends a full
  PUT of the status stanza and zeros any field a disjoint writer owns
  (e.g. a runtime reporter writing status on the same CR). #4633 tracks
  migrating to merge-patch; this PR introduces the helper the migration
  will call at every site so the DeepCopy-before-mutate discipline is
  captured in one place instead of re-implemented per call.
- `MutateAndPatchStatus[T client.Object](ctx, c, obj, mutate)` collapses
  the `DeepCopy → mutate → Status().Patch(MergeFrom(original))` idiom to
  a single call. Plain merge patch, not optimistic-lock: status-subresource
  writers with disjoint fields must coexist without forcing a 409 on every
  overlap. Spec and metadata writes still require optimistic locking —
  see #4767 (tracking) / #4914 (MCPServer migration).
- The helper does not make every multi-writer pattern safe. The doc
  comment's Caller contract spells out what it cannot defend against:
  Conditions-array writers must be the sole owner of the entire array
  (merge-patch replaces arrays wholesale for CRDs; `+listType=map` is
  only honored by strategic-merge-patch), and a scalar re-computed from
  a stale snapshot that differs from the live value will clobber a
  concurrent writer.
- Operational safeguards: no-op mutations skip the wire call to avoid
  steady-state PATCH noise on fast-requeue reconcilers, and a nil obj
  returns a descriptive error instead of panicking in the `.(T)` assertion.
- Codified checklist for new call sites added to `.claude/rules/operator.md`.
- Pure addition — no existing call-site touched in this PR. Migration of
  the existing `r.Status().Update` call sites will follow.

Relates to #4633

## Type of change

## Test plan

Tests in `cmd/thv-operator/pkg/controllerutil/status_test.go` cover:
- Happy path: mutation applied in-memory AND reflected in the patch body.
- DeepCopy isolation: regression guard for snapshot aliasing.
- No-op mutate: helper skips the wire call when the diff is `{}`.
- Disjoint-writer preservation: stale snapshot + second writer on a different
  scalar field; fresh `Get` confirms both the mutated field and the second
  writer's fields survive.
- Stale snapshot clobbers conditions from another writer: guards the
  documented Caller contract (distinct condition types do not survive a
  stale writer's merge patch, so callers must own the entire array).
- Stale scalar computation: re-assigning the read value is a wire-level
  no-op (concurrent writer preserved); assigning a differing value
  overwrites live state.
- Nil obj rejected with a descriptive error, no PATCH issued.
- Error propagation: apiserver failure returned unchanged for requeue.

No.

### DIFF
--- a/cmd/thv-operator/pkg/controllerutil/status.go
+++ b/cmd/thv-operator/pkg/controllerutil/status.go
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package controllerutil
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"reflect"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// MutateAndPatchStatus captures the current state of obj, applies mutate,
+// and patches the status subresource using a plain JSON merge patch.
+//
+// This is the canonical idiom for every status write in the operator. See
+// #4633: a full status PUT (r.Status().Update) clobbers fields the operator
+// does not track (e.g. runtime-reporter-owned fields on VirtualMCPServer
+// status). A merge-patch body only carries fields the caller actually
+// changed, so disjoint status writers coexist.
+//
+// The patch is NOT optimistic-locked: status-subresource writes are scoped
+// to the status stanza, and forcing a 409 on every disjoint-field overlap
+// would produce permanent churn with nothing gained.
+//
+// Caller contract (important — the patch body is the diff between the
+// pre-mutate snapshot and the post-mutate object; it does NOT reflect
+// what is live on the apiserver):
+//
+//   - Conditions-array writes require the caller to be the sole owner
+//     of the entire Status.Conditions array on that CRD. Per-condition-
+//     type ownership is NOT sufficient: client.MergeFrom produces an
+//     RFC 7396 merge patch, which replaces the array wholesale for CRDs
+//     (the +listType=map marker is only honored by strategic-merge-patch).
+//     Any concurrent writer whose Patch lands between this caller's Get
+//     and this caller's Patch — on any condition type, including ones
+//     this caller does not touch — will be erased. A fresh Get narrows
+//     the TOCTOU window but cannot eliminate it. If two code paths must
+//     write conditions on the same CRD, consolidate them into a single
+//     owner or move one writer to a dedicated status field outside the
+//     array.
+//
+//   - Scalar fields land in the patch body only when the post-mutate
+//     value differs from the pre-mutate snapshot. Re-assigning a scalar
+//     to the same value it was read as is a no-op at the wire level —
+//     the field is absent from the patch and a concurrent writer's
+//     value on the live object is preserved. BUT if mutate assigns a
+//     value that differs from the snapshot (e.g., a stale derivation
+//     from pod state), that value will overwrite whatever a concurrent
+//     writer wrote to the live object. There is no defense against
+//     this at the helper level: a stale computation wins. For scalars
+//     co-owned by multiple writers, use a single-owner design or
+//     refresh the object via a fresh Get before calling this helper.
+//
+// Do NOT use for metadata or spec writes. Those need optimistic locking
+// and should call r.Patch directly with
+// client.MergeFromWithOptions(orig, client.MergeFromWithOptimisticLock{}).
+// Rationale and MCPServer spec migration: #4767 (tracking), #4914 (implementation).
+//
+// Typical usage:
+//
+//	err := controllerutil.MutateAndPatchStatus(ctx, r.Client, mcpServer,
+//	    func(s *mcpv1alpha1.MCPServer) {
+//	        meta.SetStatusCondition(&s.Status.Conditions, metav1.Condition{
+//	            Type:   mcpv1alpha1.ConditionReady,
+//	            Status: metav1.ConditionTrue,
+//	            Reason: mcpv1alpha1.ConditionReasonReady,
+//	        })
+//	    })
+func MutateAndPatchStatus[T client.Object](
+	ctx context.Context, c client.Client, obj T, mutate func(T),
+) error {
+	// Reject both a true-nil interface and a typed-nil pointer. T is
+	// constrained to client.Object; every real implementer is a pointer
+	// to a struct, so a nil obj is always a programmer error. Returning
+	// an explicit error is nicer than the raw panic that the subsequent
+	// .(T) type assertion would produce.
+	v := reflect.ValueOf(obj)
+	if !v.IsValid() || (v.Kind() == reflect.Pointer && v.IsNil()) {
+		return fmt.Errorf("MutateAndPatchStatus: obj must be non-nil")
+	}
+	original := obj.DeepCopyObject().(T)
+	mutate(obj)
+	data, err := client.MergeFrom(original).Data(obj)
+	if err != nil {
+		return err
+	}
+	// Skip the wire call for a no-op mutate. The apiserver runs the full
+	// admission and audit pipeline for every PATCH regardless of body
+	// content, so sending {} costs watch-cascade and audit log noise for
+	// no benefit. Controllers like EmbeddingServerReconciler that requeue
+	// at 1s would otherwise generate steady-state no-op PATCH traffic.
+	if bytes.Equal(data, []byte("{}")) {
+		return nil
+	}
+	return c.Status().Patch(ctx, obj, client.RawPatch(types.MergePatchType, data))
+}
--- a/cmd/thv-operator/pkg/controllerutil/status_test.go
+++ b/cmd/thv-operator/pkg/controllerutil/status_test.go
@@ -1,0 +1,493 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package controllerutil
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	mcpv1beta1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1beta1"
+)
+
+// statusPatchRecordingClient wraps a client.Client and intercepts
+// .Status().Patch calls so tests can assert the wire-level patch body.
+type statusPatchRecordingClient struct {
+	client.Client
+	mu       sync.Mutex
+	bodies   []string
+	forceErr error
+}
+
+func (c *statusPatchRecordingClient) Status() client.SubResourceWriter {
+	return &statusSubResourceRecorder{parent: c, inner: c.Client.Status()}
+}
+
+type statusSubResourceRecorder struct {
+	parent *statusPatchRecordingClient
+	inner  client.SubResourceWriter
+}
+
+func (r *statusSubResourceRecorder) Create(
+	ctx context.Context, obj client.Object, subResource client.Object, opts ...client.SubResourceCreateOption,
+) error {
+	return r.inner.Create(ctx, obj, subResource, opts...)
+}
+
+func (r *statusSubResourceRecorder) Update(
+	ctx context.Context, obj client.Object, opts ...client.SubResourceUpdateOption,
+) error {
+	return r.inner.Update(ctx, obj, opts...)
+}
+
+func (r *statusSubResourceRecorder) Patch(
+	ctx context.Context, obj client.Object, patch client.Patch, opts ...client.SubResourcePatchOption,
+) error {
+	if data, err := patch.Data(obj); err == nil {
+		r.parent.mu.Lock()
+		r.parent.bodies = append(r.parent.bodies, string(data))
+		r.parent.mu.Unlock()
+	}
+	if r.parent.forceErr != nil {
+		return r.parent.forceErr
+	}
+	return r.inner.Patch(ctx, obj, patch, opts...)
+}
+
+func (r *statusSubResourceRecorder) Apply(
+	ctx context.Context, obj runtime.ApplyConfiguration, opts ...client.SubResourceApplyOption,
+) error {
+	return r.inner.Apply(ctx, obj, opts...)
+}
+
+func (c *statusPatchRecordingClient) lastBody() string {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if len(c.bodies) == 0 {
+		return ""
+	}
+	return c.bodies[len(c.bodies)-1]
+}
+
+func buildStatusTestClient(t *testing.T, seed *mcpv1beta1.MCPServer) (*statusPatchRecordingClient, client.Client) {
+	t.Helper()
+	scheme := runtime.NewScheme()
+	require.NoError(t, mcpv1beta1.AddToScheme(scheme))
+	inner := fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithObjects(seed).
+		WithStatusSubresource(&mcpv1beta1.MCPServer{}).
+		Build()
+	recorder := &statusPatchRecordingClient{Client: inner}
+	return recorder, inner
+}
+
+func newSeedMCPServer(name string) *mcpv1beta1.MCPServer {
+	return &mcpv1beta1.MCPServer{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: "default",
+		},
+		Spec: mcpv1beta1.MCPServerSpec{
+			Image:     "example/mcp:latest",
+			Transport: "stdio",
+			ProxyMode: "sse",
+			ProxyPort: 8080,
+			MCPPort:   8080,
+		},
+	}
+}
+
+// TestMutateAndPatchStatus_AppliesMutation asserts the happy path:
+// the mutation is applied to the object in place AND persisted via a
+// status-subresource merge patch whose body carries the mutated fields.
+func TestMutateAndPatchStatus_AppliesMutation(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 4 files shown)


cmd/thv-operator/pkg/controllerutil/status.go
```diff
@@ -0,0 +1,48 @@
...
+// #4633: a full status PUT (r.Status().Update) clobbers fields the operator
+// does not track (e.g. runtime-reporter-owned fields on VirtualMCPServer
+// status). A merge-patch body only carries fields the caller actually
+// changed, so disjoint status writers coexist.
+//
+// The patch is NOT optimistic-locked: status-subresource writes are scoped
+// to the status stanza, and forcing a 409 on every disjoint-field overlap
+// would produce permanent churn with nothing gained. Conditions-array
+// overlap stays safe via the per-condition idiom (meta.SetStatusCondition
+// with each writer owning distinct condition types).
```
ChrisJBurns, line 25: **Conditions-array "per-condition idiom" claim is load-bearing on an unenforced invariant (8/10, HIGH)**

`client.MergeFrom` produces an RFC 7396 merge patch (verified in `controller-runtime@v0.23.3/pkg/client/patch.go` → `jsonpatch.CreateMergePatch`) which REPLACES arrays wholesale; the `+listType=map` marker on `Conditions` is only honored by strategic-merge-patch, which CRDs don't support. So the "per-condition idiom" is safe ONLY when each writer `Get`s the live object immediately before mutating — otherwise a stale snapshot's `Conditions` array will erase entries owned by another writer on the merge.

Suggestion: either rephrase the doc to state the precondition explicitly ("callers MUST hold a freshly-`Get`ted object when Conditions may be concurrently written"), or have the helper `Get` internally before `mutate`. The latter is safer; the former is cheaper.

— *kubernetes-expert*

cmd/thv-operator/pkg/controllerutil/status_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	mcpv1alpha1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1alpha1"
+)
+
+// statusPatchRecordingClient wraps a client.Client and intercepts
+// .Status().Patch calls so tests can assert the wire-level patch body.
+// This is the status-subresource analogue of the patch-recording client
+// used for spec patches; see cmd/thv-operator/controllers/mcpserver_spec_patch_test.go.
```
ChrisJBurns, line 26: **Dangling reference to non-existent test file (10/10, MEDIUM)**

`cmd/thv-operator/controllers/mcpserver_spec_patch_test.go` does not exist in the repo (verified with `ls`). The comment says this recorder is "the status-subresource analogue of the patch-recording client used for spec patches" — but the spec-patch analogue hasn't landed yet. Either drop the cross-reference or rephrase as "the spec-patch analogue lands in \#NNNN" once that PR exists.

— *toolhive-expert*

```suggestion
// statusPatchRecordingClient wraps a client.Client and intercepts
// .Status().Patch calls so tests can assert the wire-level patch body.
```

cmd/thv-operator/pkg/controllerutil/status.go
```diff
@@ -0,0 +1,48 @@
...
+//
+// Do NOT use for metadata or spec writes. Those need optimistic locking
+// and should call r.Patch directly with
+// client.MergeFromWithOptions(orig, client.MergeFromWithOptimisticLock{}).
+// See #4767 for the spec-side rationale.
```
ChrisJBurns, line 30: **`\#4767` reads as if the spec-side helper exists (8/10, MEDIUM)**

`\#4767` is an open tracking issue ("Fix r.Update to r.Patch plus regression guard in MCPServer controller"), not a landed pattern. Grepping for `MergeFromWithOptimisticLock` returns zero matches across the repo — no spec-patch helper exists yet. The comment currently reads as if the pattern is established; it should read as "spec writes should use optimistic locking (tracked in \#4767)" to make the forward-coupling explicit.

— *toolhive-expert*

```suggestion
// client.MergeFromWithOptions(orig, client.MergeFromWithOptimisticLock{}).
// The parallel spec-side helper is tracked in #4767.
```

cmd/thv-operator/pkg/controllerutil/status.go
```diff
@@ -0,0 +1,48 @@
...
+//
```
ChrisJBurns, line 31: **Stale-snapshot risk extends to scalar fields — worth stating (7/10, MEDIUM)**

The disjoint-preservation test covers the case where `mutate` only touches fields the other writer does not. The foot-gun: if `mutate` re-assigns `Status.Phase = "Pending"` (a value it read from a stale snapshot), the patch body WILL include `Phase` and clobber the concurrent writer's "Ready". Worth one sentence in the doc: "`mutate` must only write fields the caller actually intends to change. Re-assigning a field to its previously-read value will include it in the patch body and can clobber concurrent writes."

— *kubernetes-expert*

cmd/thv-operator/pkg/controllerutil/status.go
```diff
@@ -0,0 +1,48 @@
...
+//	    func(s *mcpv1alpha1.MCPServer) {
+//	        meta.SetStatusCondition(&s.Status.Conditions, metav1.Condition{
+//	            Type:   mcpv1alpha1.ConditionReady,
+//	            Status: metav1.ConditionTrue,
+//	            Reason: mcpv1alpha1.ConditionReasonReady,
+//	        })
+//	    })
+func MutateAndPatchStatus[T client.Object](
+	ctx context.Context, c client.Client, obj T, mutate func(T),
+) error {
```
ChrisJBurns, line 44: **No-op `mutate` still issues a PATCH (9/10, LOW — raised independently by 2 agents)**

When `mutate` makes no changes, `client.MergeFrom` produces `{}` and `Status().Patch` still issues the HTTP PATCH. Harmless but generates apiserver load in tight reconcile loops. One sentence in the doc would let callers decide whether to gate on a dirty check.

— *kubernetes-expert, code-reviewer*

cmd/thv-operator/pkg/controllerutil/status.go
```diff
@@ -0,0 +1,48 @@
...
+	original := obj.DeepCopyObject().(T)
```
ChrisJBurns, line 45: **Typed-nil `obj` triggers raw panic in `DeepCopyObject().(T)` (7/10, MEDIUM)**

Per `.claude/rules/go-style.md` ("Constructor Validation: Fail Loudly on Invalid Input"), required parameters should fail loudly with a descriptive error rather than a raw panic. `obj.DeepCopyObject()` panics inside the generated deepcopy methods when `obj` is a typed-nil. Either document the non-nil precondition on the function doc comment (cheapest) or add an explicit nil guard returning `fmt.Errorf("MutateAndPatchStatus: obj must be non-nil")`.

— *code-reviewer*

cmd/thv-operator/pkg/controllerutil/status_test.go
```diff
@@ -0,0 +1,263 @@
...
+// caller writes status from a stale snapshot, fields owned by a
+// different writer must survive. A full Status().Update would clobber
+// them (PUT semantics replace the whole status stanza); a merge patch
+// computed from the stale snapshot only carries the fields this caller
+// changed, so disjoint fields on the live object are left alone.
+//
+// Test shape: seed an object, snapshot it, let a "second writer" mutate
+// a disjoint field on the live object, then call the helper on the
+// stale snapshot and mutate a different field. Assert both fields are
+// present on a fresh Get of the live object.
```
ChrisJBurns, line 201: **Consider an envtest-backed variant of this test (7/10, LOW)**

The controller-runtime fake client's merge-patch implementation has historically diverged from a real apiserver, especially around status-subresource isolation and `+listType=map`. This test is exactly the kind of semantics-sensitive assertion that benefits from envtest. Per team convention (prefer envtest over fake/chainsaw for operator integration tests), a follow-up envtest variant would meaningfully strengthen the regression guard the disjoint-preservation test provides. Happy path + error propagation can stay on the fake client.

— *kubernetes-expert*

ChrisJBurns (commented):
**Agents consulted**: kubernetes-expert, go-expert-developer, code-reviewer, toolhive-expert

### Consensus Summary
| 1 | Conditions-array doc claim is only true when writers `Get` before mutating | 8/10 | HIGH | Clarify doc |
| 2 | Comment references non-existent `mcpserver_spec_patch_test.go` | 10/10 | MEDIUM | Fix / remove |
| 3 | `\#4767` referenced as if spec-side helper exists | 8/10 | MEDIUM | Clarify doc |
| 4 | Typed-nil `obj` triggers raw panic | 7/10 | MEDIUM | Document or guard |
| 5 | Stale-snapshot scalar-field clobber — doc gap | 7/10 | MEDIUM | Document contract |
| 6 | No-op `mutate` still PATCHes | 9/10 | LOW | One-line doc note |
| 7 | Envtest variant preferred over fake-client for merge-patch semantics | 7/10 | LOW | Follow-up test |

### Overall

The helper itself is correct, small (48 lines), well-tested, and solves a real problem from \#4633. The `DeepCopyObject().(T)` pattern, the decision to skip optimistic locking for status, the test structure (parallel, wire-body assertions, disjoint-writer regression), and the package placement are all sound. Four specialized agents reviewed the change and none disputed the fundamental design.

The findings cluster around documentation precision rather than the code behavior. The HIGH finding is the most important: the doc claim that "Conditions-array overlap stays safe via the per-condition idiom" is load-bearing on an invariant the helper does not enforce. `client.MergeFrom` produces RFC 7396 merge patches (confirmed in `controller-runtime@v0.23.3/pkg/client/patch.go` → `jsonpatch.CreateMergePatch`) which REPLACE arrays wholesale — the `+listType=map` marker on `Conditions` is only honored by strategic-merge-patch, which CRDs don't support. So the claim is true in practice only if every writer `Get`s the live object immediately before calling the helper. Worth stating that precondition explicitly so future call-site migrations don't take the doc at face value.

The two cross-references (`mcpserver_spec_patch_test.go` and `\#4767` "spec-side rationale") point at code that does not yet exist in the repo. `\#4767` is an open tracking issue, not a landed pattern, and there is no spec-patch-recording test file anywhere. These are fixable with a one-line edit each — either remove the forward references or rephrase to make clear they describe planned follow-up work.

None of these block merge in my view; a follow-up or amended commit covers them. The migration PRs that land on top of this helper will benefit from tighter documentation here.

### Documentation

Per CLAUDE.md, architecture doc sync was considered. This is an internal helper with no user-facing or architecture-doc surface — `docs/arch/` does not need updates. The in-package `doc.go` was correctly updated.

---
*Generated with Claude Code*

cmd/thv-operator/pkg/controllerutil/status_test.go
jhrozek: Agree this is worth doing, and agree with the reviewer's own "follow-up" framing. Scoping notes for posterity / whoever picks it up:

- The specific assertion that benefits from envtest is `TestMutateAndPatchStatus_PreservesDisjointStatusFields` — it's the only one here that depends on merge-patch wire semantics. Happy-path, deep-copy, no-op-skip, nil-guard, error-propagation, and the stale-snapshot clobber tests are all Go-level invariants that don't need apiserver fidelity.
- The `+listType=map` concern the reviewer raised is explicitly out of scope for envtest coverage — CRDs don't get strategic-merge-patch from the apiserver either, and that limitation is now documented in the helper's Caller contract and guarded by the new `TestMutateAndPatchStatus_StaleSnapshotClobbersConditionsFromAnotherWriter` regression.
- Reusable harness: `cmd/thv-operator/test-integration/mcp-server/suite_test.go` is the existing envtest bootstrap; the new spec can slot in there or under a new `controllerutil/` suite dir depending on preference.

Happy to file the follow-up if you agree the scope above is right.

jhrozek: Agree this is worth doing, and agree with the reviewer's own "follow-up" framing. Scoping notes for posterity / whoever picks it up:

- The specific assertion that benefits from envtest is `TestMutateAndPatchStatus_PreservesDisjointStatusFields` — it's the only one here that depends on merge-patch wire semantics. Happy-path, deep-copy, no-op-skip, nil-guard, error-propagation, and the stale-snapshot clobber tests are all Go-level invariants that don't need apiserver fidelity.
- The `+listType=map` concern the reviewer raised is explicitly out of scope for envtest coverage — CRDs don't get strategic-merge-patch from the apiserver either, and that limitation is now documented in the helper's Caller contract and guarded by the new `TestMutateAndPatchStatus_StaleSnapshotClobbersConditionsFromAnotherWriter` regression.
- Reusable harness: `cmd/thv-operator/test-integration/mcp-server/suite_test.go` is the existing envtest bootstrap; the new spec can slot in there or under a new `controllerutil/` suite dir depending on preference.

Happy to file the follow-up if you agree the scope above is right.

cmd/thv-operator/pkg/controllerutil/status.go
jhrozek: Addressed in `f6fab6aee`. The "per-condition idiom stays safe" framing has been removed from the doc comment and the commit message, and the Caller contract now explicitly states what the helper cannot defend against:

- Conditions-array writes require the caller to be the **sole owner of the entire `Status.Conditions` array** — per-condition-type ownership is NOT sufficient.
- The array is replaced wholesale because `client.MergeFrom` produces an RFC 7396 merge patch and `+listType=map` is only honored by strategic-merge-patch.
- A fresh `Get` narrows but does not eliminate the TOCTOU window between Get and Patch.

Added a regression test (`TestMutateAndPatchStatus_StaleSnapshotClobbersConditionsFromAnotherWriter`) that exercises the different-condition-type clobber case so the documented contract stays load-bearing against silent behavioural changes.

The same tightening landed in the codified checklist at `.claude/rules/operator.md`.

cmd/thv-operator/pkg/controllerutil/status_test.go
jhrozek: Addressed in `f6fab6aee`. The dangling reference to `cmd/thv-operator/controllers/mcpserver_spec_patch_test.go` has been removed from the `statusPatchRecordingClient` doc comment.

cmd/thv-operator/pkg/controllerutil/status.go
jhrozek: Addressed in `f6fab6aee`. The helper's doc comment now reads:

> Rationale and MCPServer spec migration: #4767 (tracking), #4914 (implementation).

This makes the tracking-vs-implementation distinction explicit so a reader doesn't assume the spec-side pattern is already landed.

cmd/thv-operator/pkg/controllerutil/status.go
jhrozek: Addressed in `f6fab6aee`. Two changes:

**Caller contract updated** — the doc comment now spells out the scalar semantics explicitly, including the slightly-counter-intuitive detail that re-assigning the same read value is a wire-level no-op (so a concurrent writer's value on the live object is preserved), but a differing value from a stale computation will clobber live state.

**Regression test added** — `TestMutateAndPatchStatus_StaleScalarComputationClobbersConcurrentWrite` has two sub-cases:

1. `reassigning_read_value_is_noop` — confirms the helper's merge-patch body omits unchanged scalars and the concurrent writer's value survives.
2. `stale_computation_clobbers_concurrent_write` — confirms a new value differing from the snapshot lands in the patch body and overwrites the live value.

Note: while writing this, I verified the reviewer's initial framing empirically — "re-assigning the stale-read value includes it in the patch body" is incorrect for `client.MergeFrom`. The diff is snapshot-vs-mutated, not snapshot-vs-live. The doc and tests now reflect the actual semantics.

The codified checklist in `.claude/rules/operator.md` includes scalar co-ownership as one of the three items to check when adding a new call site.

cmd/thv-operator/pkg/controllerutil/status.go
jhrozek: Addressed in `f6fab6aee`. The helper now short-circuits before the wire call when the merge-patch body is `{}`:

```go
data, err := client.MergeFrom(original).Data(obj)
if err != nil {
    return err
}
if bytes.Equal(data, []byte("{}")) {
    return nil
}
return c.Status().Patch(ctx, obj, client.RawPatch(types.MergePatchType, data))
```

Using `client.RawPatch(types.MergePatchType, data)` avoids a second serialization — we only marshal once via `MergeFrom(original).Data(obj)` and pass the raw bytes through.

Added `TestMutateAndPatchStatus_NoOpMutateSkipsWireCall` which seeds a condition and then calls `meta.SetStatusCondition` with the same Status/Reason/Message; the idempotent mutation produces a `{}` diff and the test asserts no PATCH is recorded by the wire-body recorder.

The rationale (apiserver admission + audit fires regardless of body content, and EmbeddingServerReconciler's 1s requeue would otherwise generate steady-state no-op PATCHes) is in an inline comment on the short-circuit.

cmd/thv-operator/pkg/controllerutil/status.go
jhrozek: Addressed in `f6fab6aee`. Added a nil guard at the top of the helper:

```go
v := reflect.ValueOf(obj)
if !v.IsValid() || (v.Kind() == reflect.Pointer && v.IsNil()) {
    return fmt.Errorf("MutateAndPatchStatus: obj must be non-nil")
}
```

This handles both a true-nil interface (`v.IsValid()` → false) and a typed-nil pointer (`v.IsNil()`), which is the case the `.(T)` type assertion would have panicked on. Returning a descriptive error lets the reconciler's requeue/log path handle the programmer error cleanly instead of crashing the worker.

Added `TestMutateAndPatchStatus_RejectsNilObj` which asserts the error message names the parameter for debugging (`"obj must be non-nil"`), that `mutate` is not called, and that no PATCH is issued.

cmd/thv-operator/pkg/controllerutil/status_test.go
ChrisJBurns: Yep thats sounds good. You want to file and issue or you think its quicker to just get CC to do a PR?

ChrisJBurns (approved):

cmd/thv-operator/pkg/controllerutil/status_test.go
jhrozek: I think letting CC do it has lower chance of being deferred/forgotten :-) I have a "master session" that tracks the implementation of #4633, I've added it to its plan.

ChrisJBurns (approved):

PR #4943, opened 2026-04-20
Allow container socket paths to be configured via config file

### LINKED ISSUE #4612: Support configuring container engine socket path via config file
## Current Behavior

Socket paths can only be overridden via environment variables:

- `TOOLHIVE_DOCKER_SOCKET`
- `TOOLHIVE_PODMAN_SOCKET`
- `TOOLHIVE_COLIMA_SOCKET`

These are checked in `pkg/container/docker/sdk/client_unix.go` before falling back to auto-detection of standard socket locations. There is no way to persist this configuration in the ToolHive config file (`~/.config/toolhive/config.yaml`), meaning users with non-standard socket paths must set environment variables in every shell session or wrapper script.

## Proposed Change

Add socket path fields to the `Config` struct in `pkg/config/config.go`, allowing users to configure them once via `thv config` or by editing the YAML file directly:

```yaml
container_runtime:
  docker_socket: /run/user/1000/docker.sock
  podman_socket: /run/user/1000/podman/podman.sock
  colima_socket: /home/user/.colima/custom/docker.sock
```

Socket resolution in `findPlatformContainerSocket` would check sources in this order, preserving existing precedence:

1. **Environment variable** (existing)
2. **Config file** (new)
3. **Auto-detection of standard paths** (existing)

## Motivation

Configuring environment variables across GUI and CLI apps can be a bad experience, because there are various ways env variables can be set up, and not all of them are accessible by both GUI and CLI processes - it can actually even depend on how the process is launched.

Misconfiguration is hard to spot, but can lead to confusing, unexpected experiences. This is not user friendly, because users should not need to have a deep understanding of how their operating system handles environment variables across different ways of launching apps just to be able to use ToolHive. 

See [[this Discord thread](https://discord.com/channels/1184987096302239844/1340110387453886515/1490713636682076340) for a concrete user report that illustrates the problem.

kantord:
## Summary

Users with non-standard container socket paths had to set environment variables
(`TOOLHIVE_DOCKER_SOCKET`, `TOOLHIVE_PODMAN_SOCKET`, `TOOLHIVE_COLIMA_SOCKET`) in every
shell session or wrapper script. This is fragile because GUI and CLI processes can inherit
environment variables differently depending on how they are launched, making misconfiguration
hard to spot and debug.

- Add `container_runtime` block to `~/.config/toolhive/config.yaml` so socket paths can be
  set once and persist across all launch methods
- Socket resolution now follows: env var → config file → auto-detection (existing precedence preserved)
- Add `SocketConfig` to `pkg/container/runtime` as the shared type used by both the config
  package and the socket detection code (avoids an import cycle)

Fixes #4612

## Type of change

- [x] New feature

## Test plan

New unit tests in `pkg/container/docker/sdk/client_unix_test.go`:
- `TestFindContainerSocket_ConfigPath` — table-driven: valid config path accepted, nonexistent path returns error
- `TestFindContainerSocket_EnvVarPrecedence` — env var overrides a config-supplied path

Yes. Users can now configure socket paths in `~/.config/toolhive/config.yaml`:

```yaml
container_runtime:
  docker_socket: /run/user/1000/docker.sock
  podman_socket: /run/user/1000/podman/podman.sock
  colima_socket: /home/user/.colima/custom/docker.sock
```

This is lower-friction than setting environment variables, and works consistently regardless of how ToolHive is launched.

The `SocketConfig` struct lives in `pkg/container/runtime` rather than `pkg/config` to avoid
an import cycle: `pkg/container/docker/sdk` → `pkg/config` → `pkg/secrets` → ... → `pkg/container/docker/sdk`.
The `pkg/container/runtime` package is already imported by both sides and has precedent for
this pattern (see `ScalingConfig`).


kantord:
@claude deep review of branch please

pkg/container/docker/sdk/client_unix.go
```diff
@@ -76,6 +76,11 @@ func findPlatformContainerSocket(rt runtime.Type) (string, runtime.Type, error)
 		return customSocketPath, runtime.TypeDocker, nil
 	}
 
+	// Check config file overrides (after env vars, before auto-detection)
```
jhrozek, line 79: I think there's a correctness issue here. `checkSocketConfigOverrides` doesn't consider the `rt` argument, so whichever override field is non-empty wins for every iteration of the outer loop in `NewDockerClient`.

Concrete scenario: a user has a stale `podman_socket:` in config and Docker is reachable via auto-detect. Iter 1 (Podman) fails because the Podman path doesn't exist. Iter 2 should auto-detect Docker — but `checkSocketConfigOverrides` returns the same Podman error again, Docker auto-detect never runs, and `NewDockerClient` returns `"no supported container runtime available: invalid Podman socket path from config"` on a machine where Docker works fine.

Windows has the same shape at `client_windows.go:92-99`.

Fix I'd suggest: make the override check match the current `rt` — only look at `overrides.PodmanSocket` when `rt == TypePodman`, etc. Then a broken podman override only disables podman, and the outer loop's independent-per-runtime semantics work as intended. Side benefit: the "priority ordering" question (which field wins when multiple are set) disappears, since each runtime consults only its own field.

pkg/container/docker/sdk/factory.go
```diff
@@ -49,15 +53,44 @@ const (
 
 var supportedSocketPaths = []runtime.Type{runtime.TypePodman, runtime.TypeDocker, runtime.TypeColima}
 
+// loadSocketOverrides reads socket path overrides from the ToolHive config file.
+// Best-effort: returns empty overrides on any error so auto-detection takes over.
+func loadSocketOverrides() runtime.SocketConfig {
```
jhrozek, line 58: Would it be worth routing this through the existing config Provider abstraction? `pkg/config.NewProvider()` dispatches to `DefaultProvider` / `PathProvider` / `KubernetesProvider` and honours `RegisterProviderFactory` for injected providers (see `pkg/config/interface.go:225`, `:396`, `:566`). Reading XDG directly means `PathProvider` consumers (custom config paths, tests) and any code using `RegisterProviderFactory` won't see this feature.

The reason to not just use `NewProvider().GetConfig()` is the "first-run creates the file + runs migrations" side effect — but in practice the CLI has almost always touched config by the time `NewDockerClient` runs, so the singleton is populated and no extra file creation happens. A read-only `LoadIfExists()` on the Provider interface would make the concern go away entirely.

pkg/container/docker/sdk/client_unix.go
```diff
@@ -254,3 +259,27 @@ func findColimaSocket() (string, error) {
...
+// Returns ("", "", nil) if no override applies.
+func checkSocketConfigOverrides(overrides runtime.SocketConfig) (string, runtime.Type, error) {
+	if overrides.PodmanSocket != "" {
+		return resolveConfigSocket(overrides.PodmanSocket, runtime.TypePodman, "Podman")
+	}
+	if overrides.DockerSocket != "" {
+		return resolveConfigSocket(overrides.DockerSocket, runtime.TypeDocker, "Docker")
+	}
+	if overrides.ColimaSocket != "" {
+		return resolveConfigSocket(overrides.ColimaSocket, runtime.TypeDocker, "Colima")
```
jhrozek, line 273: Small inconsistency: the config override for Colima returns `runtime.TypeDocker` here, but the auto-detect branch at `client_unix.go:98-103` returns `runtime.TypeColima`. The env-var branch at `client_unix.go:74-76` also returns `TypeDocker`, so this PR is copying the existing pattern — but the disagreement with auto-detect means any caller that branches on `runtime.TypeColima` behaves differently depending on how the user configured Colima.

Either align all three branches (env + config to return `TypeColima`, or auto-detect to `TypeDocker`), or leave a comment explaining why the three disagree.

pkg/container/docker/sdk/factory.go
```diff
@@ -49,15 +53,44 @@ const (
...
+	data, err := os.ReadFile(filepath.Clean(configPath))
+	if err != nil {
+		slog.Debug("failed to read config file for socket overrides", "error", err)
+		return runtime.SocketConfig{}
+	}
+
+	var cfg struct {
+		ContainerRuntime runtime.SocketConfig `yaml:"container_runtime,omitempty"`
+	}
+	if err := yaml.Unmarshal(data, &cfg); err != nil {
```
jhrozek, line 75: `loadSocketOverrides` has no tests, and it's the integration point that turns YAML into `SocketConfig`. It also has a deliberate "best-effort, return zero value on any error" contract across three distinct failure modes (xdg lookup, file read, yaml parse) — that contract should be pinned.

Worth a table-driven test that points `XDG_CONFIG_HOME` at `t.TempDir()` and covers: missing file, empty file, malformed YAML, valid file with each subset of fields populated, valid file with no `container_runtime` key, valid file with only unrelated keys. `adrg/xdg` provides `xdg.Reload()` if you hit caching issues on macOS. Non-parallel because of `t.Setenv`.

jhrozek (commented):
A few observations from reviewing the socket-override threading.

eleftherias:
@kantord I think this is still worth merging, could you take a look at the conflicts and comments?

kantord:
> @kantord I think this is still worth merging, could you take a look at the conflicts and comments?

absolutely, sorry for the delay

kantord:
Hi @jhrozek  sorry for the delayed response! I kept delaying it because your suggestions seemed to make a lot of sense, but also a good understanding of Go and the codebase :smile: But I never got the time to actually sit down and do it properly.

Would you mind actually pushing your suggestions directly to the branch?

pkg/container/docker/sdk/client_unix.go
stantheman0128: Same on the Windows file. `if overrides.PodmanSocket != ""` at `client_windows.go` does not look at `rt` either, so a leftover `podman_socket` fails `NewDockerClient`'s Docker pass the same way.

Current main already does this for `TOOLHIVE_PODMAN_SOCKET` before the `rt == TypeDocker` branch. Config should not copy that.

After rebase this helper also has to return `[]string` to match `findPlatformContainerSocket` on main. The Dial itself (`resolveConfigPipe` / `winio.DialPipeContext`) is the right probe for named pipes; I would not switch that to `os.Stat`.

stantheman0128:
Support. I originally treated this as Linux-only docker.sock config. That was wrong: `client_windows.go` already Dials env overrides with `winio.DialPipeContext`, and this PR adds `resolveConfigPipe` for config paths. Named pipes cannot be validated with `os.Stat`; Dial is the right probe.

The April Claude review that said Windows would not compile is stale. The current head already updates the Windows signature.

What still looks open from a Windows box:

1. Runtime gating, same as @jhrozek already wrote on the Unix helper. `if overrides.PodmanSocket != ""` does not look at `rt`, so a leftover `podman_socket` fails the Docker iteration too. Env vars on current main already have that shape (`TOOLHIVE_PODMAN_SOCKET` is checked before `rt == TypeDocker`). Config should not copy it.

2. Tests. `client_unix_test.go` is `//go:build !windows`. The DialPipe path never runs in Windows CI. `pkg/server/discovery/health_windows_test.go` already has a live named-pipe fixture.

Rebase is more than conflict markers. On current main, `findPlatformContainerSocket` returns `[]string`, not a single path. The PR still has the old signature.

Follow-up, not a reason to NACK: if someone puts a `unix://C:\...` URL in `docker_socket` (Docker Desktop WSL / AF_UNIX on Win11), neither Stat nor DialPipe is the right probe. `discovery.UnixSocketURL` from #6416 is the sibling for that shape.

I will not open a second PR on #4612 while this one is open. If a Windows DialPipe test plus an `rt` gate would help after the rebase, say so.

PR #4941, opened 2026-04-20
Add e2e tests for thv vmcp CLI

yrobla:
## Summary

Adds e2e test coverage for the \`thv vmcp\` CLI, closing #4888.

**\`test/e2e/vmcp_cli_test.go\`** — basic serve mode coverage:
- Quick mode (\`--group\`) and config-file mode (\`init → validate → serve --config\`), each with MCP client connectivity and \`tools/list\` assertions
- Binding test confirming the server listens only on \`127.0.0.1\` (verified via \`ss\`/\`lsof\`, not just dial)
- Negative tests for missing flags and a nonexistent config path

**\`test/e2e/vmcp_cli_features_test.go\`** — feature parity with the Kubernetes operator test suite:
- **Conflict resolution** — two yardstick backends, prefix strategy, asserts both \`backendA_echo\` and \`backendB_echo\` appear
- **Tool filtering** — per-workload \`Filter\` allow-list and \`ExcludeAll\`, asserts only the allowed tool is visible
- **Global ExcludeAll** — \`AggregationConfig.ExcludeAllTools: true\`, asserts \`tools/list\` returns an empty slice
- **Composite sequential tool** — inline \`echo_twice\` workflow, asserts it appears in listing and executes successfully
- **Tier-1 optimizer** — \`--optimizer\` quick-mode flag, asserts only \`find_tool\` + \`call_tool\` are exposed and \`find_tool\` returns results

Infra-heavy follow-ups (JWT/OIDC, circuit breaker, Redis sessions, Tier-2 TEI embedding) tracked in #4944.
Two small optimizer bug fixes found during review are tracked separately in #4945.

Fixes #4888

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring (no behavior change)
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [ ] Unit tests (\`task test\`)
- [x] E2E tests (\`task test-e2e\`)
- [x] Linting (\`task lint-fix\`)
- [ ] Manual testing (describe below)

## Changes

| File | Change |
|------|--------|
| \`test/e2e/vmcp_cli_test.go\` | New — basic serve mode e2e tests (quick mode, config-file mode, binding, error cases) |
| \`test/e2e/vmcp_cli_features_test.go\` | New — feature parity e2e tests (conflict resolution, filtering, ExcludeAll, composite, optimizer) |

## Does this introduce a user-facing change?

No — tests only.

## Special notes for reviewers

[...]

PR #4948, opened 2026-04-20
vmcp: stop the TEI manager on Start failure and apply image default in Serve

### LINKED ISSUE #4945: vmcp: injectOptimizerConfig leaks TEI container on Start failure; EmbeddingImage not defaulted in Serve()
Two small bugs in \`pkg/vmcp/cli/serve.go\` found during review of #4941:

### 1. TEI container leak on partial Start failure

In \`injectOptimizerConfig\`, if \`mgr.Start(ctx)\` returns an error after the container has been partially deployed (e.g. container created but health poll timed out), \`mgr.Stop\` is never called and the container is left running.

Fix: call \`mgr.Stop\` best-effort before returning the Start error.

\`\`\`go
teiURL, err := mgr.Start(ctx)
if err != nil {
    _ = mgr.Stop(context.Background()) // best-effort cleanup
    return nil, fmt.Errorf("failed to start TEI embedding service: %w", err)
}
\`\`\`

### 2. \`EmbeddingImage\` not defaulted in \`Serve()\`

\`ServeConfig.EmbeddingModel\` is defaulted to \`DefaultEmbeddingModel\` in \`Serve()\` when empty, but \`EmbeddingImage\` is passed directly to \`NewEmbeddingServiceManager\` without the same treatment. The manager applies its own internal default (\`defaultTEIImage\`), but:

- The two fields are inconsistent — one is defaulted at the \`Serve()\` layer, the other deep inside the manager.
- There is no exported \`DefaultEmbeddingImage\` constant for callers to reference.

Fix: export \`DefaultEmbeddingImage\` and apply it in \`Serve()\` alongside \`DefaultEmbeddingModel\`.

Both issues were raised by Copilot review on #4941 but dropped when that PR was scoped down to e2e tests only.

SAY-5:
## Summary

Two small fixes in `pkg/vmcp/cli/` raised by #4945 (both originally noticed in the #4941 review but dropped when that PR scoped down to e2e tests).

Fixes #4945.

## 1. TEI container leak on partial Start failure

`injectOptimizerConfig` in `pkg/vmcp/cli/serve.go` calls `mgr.Start(ctx)` and propagates the error straight up. If Start has already deployed the container and then e.g. the health poll times out, the container is left running and `mgr.Stop` is never called.

This PR adds a best-effort cleanup before returning the wrapped error:

```go
teiURL, err := mgr.Start(ctx)
if err != nil {
    _ = mgr.Stop(context.Background())
    return nil, fmt.Errorf("failed to start TEI embedding service: %w", err)
}
```

`Stop` is best-effort because the container may legitimately not exist at this point; either way we're no worse off than before.

## 2. `EmbeddingImage` defaulted at `Serve()` alongside `EmbeddingModel`

`Serve()` already defaults `ServeConfig.EmbeddingModel` to `DefaultEmbeddingModel` when empty. The sibling `EmbeddingImage` was passed straight to `NewEmbeddingServiceManager`, which applied its own private `defaultTEIImage` default. That's inconsistent, and there was no exported constant for external callers.

This PR:

- Renames `defaultTEIImage` → `DefaultEmbeddingImage` (exported) in `pkg/vmcp/cli/embedding_manager.go`.
- Applies the default in `Serve()` the same way `DefaultEmbeddingModel` is applied.
- Keeps the manager's empty-`Image` fallback intact (still points at the same constant) so direct callers of `NewEmbeddingServiceManager` are unaffected.

## Testing

- `go build ./...` and `go test ./pkg/vmcp/...` pass.
- No behaviour change on the happy path; image defaulting is semantically unchanged, just moved one layer up.

### DIFF
--- a/pkg/vmcp/cli/embedding_manager.go
+++ b/pkg/vmcp/cli/embedding_manager.go
@@ -28,8 +28,9 @@ import (
 )
 
 const (
-	// defaultTEIImage is the default HuggingFace Text Embeddings Inference image.
-	defaultTEIImage = "ghcr.io/huggingface/text-embeddings-inference:cpu-latest"
+	// DefaultEmbeddingImage is the default HuggingFace Text Embeddings
+	// Inference image used when ServeConfig.EmbeddingImage is empty.
+	DefaultEmbeddingImage = "ghcr.io/huggingface/text-embeddings-inference:cpu-latest"
 
 	// DefaultEmbeddingModel is the HuggingFace model used when EmbeddingModel is empty.
 	DefaultEmbeddingModel = "BAAI/bge-small-en-v1.5"
@@ -89,7 +90,7 @@ type EmbeddingServiceManagerConfig struct {
 	Model string
 
 	// Image is the TEI container image to run.
-	// Defaults to defaultTEIImage when empty.
+	// Defaults to DefaultEmbeddingImage when empty.
 	Image string
 }
 
@@ -142,7 +143,7 @@ func NewEmbeddingServiceManager(factory ContainerFactory, cfg EmbeddingServiceMa
 		return nil, fmt.Errorf("model must not be empty")
 	}
 	if cfg.Image == "" {
-		cfg.Image = defaultTEIImage
+		cfg.Image = DefaultEmbeddingImage
 	}
 
 	containerName := containerNameForModel(cfg.Model)
--- a/pkg/vmcp/cli/serve.go
+++ b/pkg/vmcp/cli/serve.go
@@ -302,9 +302,13 @@ func Serve(ctx context.Context, cfg ServeConfig) error {
 		if model == "" {
 			model = DefaultEmbeddingModel
 		}
+		image := cfg.EmbeddingImage
+		if image == "" {
+			image = DefaultEmbeddingImage
+		}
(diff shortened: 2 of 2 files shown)


yrobla (changes requested):
Both fixes look correct and well-scoped.

**Fix 1 (container leak):** The `_ = mgr.Stop(context.Background())` on the error path is the right approach — best-effort with a detached context so a cancelled caller context doesn't prevent cleanup. Code and comment are clear.

**Fix 2 (image default):** Clean consistency fix. The pattern mirrors how `DefaultEmbeddingModel` is already handled, and the belt-and-suspenders fallback in `NewEmbeddingServiceManager` is kept intact for direct callers.

One gap: `TestInjectOptimizerConfig_Tier2_StartError` (optimizer_wiring_test.go) tests the error path but never asserts `stub.stopSeen == true`, so the new cleanup behaviour isn't actually verified. The `stubEmbeddingManager` already tracks `stopSeen`, so it would be a one-liner:

```go
assert.True(t, stub.stopSeen, "Stop must be called on Start failure for best-effort cleanup")
```

Without that assertion, the test would still pass if the `mgr.Stop` call were accidentally removed in a future refactor.

yrobla (approved):
approving, non blocking comments

Issue #4950, opened 2026-04-21
thv registry list serves stale cache when OAuth flow fails, masking auth failure


lorr1:
## Bug description

When `thv registry list` (or anything that routes through `CachedAPIRegistryProvider.GetRegistry`) triggers an OAuth/OIDC auth flow that fails — the user never completes the browser flow, the flow times out, or the registry rejects the token — the CLI **silently returns stale cached registry data** instead of surfacing the auth failure.

The fallback in `refreshCache()` (`pkg/registry/provider_cached.go:116-143`) is indiscriminate: *any* error from the upstream API causes previously-cached data to be returned with `nil` error. It does not distinguish transient failures (network blip, upstream 5xx) from authorization failures. The persistent disk cache lives at `~/.toolhive/cache/` with `maxCacheAge = 7 * 24 * time.Hour` (`pkg/registry/provider_cached.go:26`), so stale-serving can persist for up to a week after a user loses access.

## Steps to reproduce

1. Configure `thv` against a registry behind OAuth:
   `thv config set-registry <url> --issuer <issuer> --client-id <id>`
2. Run `thv registry list` once and complete the OAuth flow so the cache gets populated on disk.
3. Wait past the in-memory cache TTL (1 hour) — or just run in a fresh process so the refresh path runs.
4. Run `thv registry list` again and **do nothing** when the OAuth URL is displayed. Wait for the flow to time out. (Reporters have seen a second OAuth flow kick off and also time out.)
5. Observe: after the OAuth flow(s) fail, the registry listing is printed anyway, and the process exits 0.

## Expected behavior

The CLI surfaces the auth error and exits non-zero, e.g.:

```
Error: failed to list servers: registry authentication required:
Run `thv registry login` to authenticate.
```

Authorization state may have changed since the cache was populated (group membership revoked, token rotated, session terminated), so silently returning stale data is a correctness and security hazard.

## Actual behavior

`thv registry list` prints the previous (cached) server list, no error, exit 0 — giving the user the false impression they are still authenticated. Follow-up commands like `thv run <server>` that rely on live auth fail later with confusing errors.

## Evidence

**Bug site**: `pkg/registry/provider_cached.go:116-143`

```go
func (p *CachedAPIRegistryProvider) refreshCache() (*types.Registry, error) {
    p.cacheMu.Lock()
    defer p.cacheMu.Unlock()

    registry, err := p.APIRegistryProvider.GetRegistry()
    if err != nil {
[...]

PR #4951, opened 2026-04-21
Reproduce registry auth stale-cache bug

lorr1:
## Summary

- **Why**: `CachedAPIRegistryProvider.refreshCache()` silently falls back to stale cached registry data when the upstream API call fails — including when the failure is an auth error. As a result, `thv registry list` against a registry behind OAuth prints previously-cached servers with no error even after the user's OAuth flow times out or their credentials are revoked. This masks the auth failure completely and is the bug reported in #4950.
- **What**: adds two failing unit tests in `pkg/registry/provider_cached_authbug_test.go` that exercise the two distinct error-propagation paths (server-side 401/403 and OAuth-browser-flow timeout) that both hit the same buggy fallback in `refreshCache()`.

**This PR is intentionally a draft and does not include a fix** — it's a reproducer for #4950. The fix direction is under discussion in the issue because a naive sentinel-check fix is insufficient for the browser-flow path (the error carries neither `auth.ErrRegistryAuthRequired` nor `api.ErrRegistryUnauthorized`).

Refs #4950

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue) — *reproducer only; fix to follow*

## Test plan

- [x] New unit tests added in `pkg/registry/provider_cached_authbug_test.go`.
- [x] Verified both tests fail on current `main`:
  ```
  --- FAIL: TestCachedProvider_AuthErrorNotMaskedByStaleCache (0.00s)
        An error is expected but got nil.
        expected auth error to propagate on refresh; got nil —
        stale cache is masking the auth failure (bug)

  --- FAIL: TestCachedProvider_OAuthFlowFailureNotMaskedByStaleCache (0.00s)
        An error is expected but got nil.
        expected OAuth flow failure to propagate on refresh; got nil —
        stale cache is masking the OAuth failure (bug)
  ```
- [x] `task lint-fix` passes (0 issues).

- The second test uses a stubbed `auth.TokenSource` (`failingTokenSource`) that mirrors the exact error wrapping `oauthTokenSource.Token()` applies when its browser flow fails — see `pkg/registry/auth/oauth_token_source.go:61-67` and `:110-113`. This is important because a sentinel-check-only fix in `refreshCache` would catch test 1 but miss test 2, which is the user-reported scenario.
- The first test uses `tokenSource=nil` so the construction-time validation probe (`pkg/registry/provider_api.go:57-74`) runs against the healthy server; the failure is introduced later by flipping the server's response.
- Scope beyond the CLI: `CachedAPIRegistryProvider` is also used by `pkg/api/v1/registry*.go` and `pkg/mcp/server/search_registry.go`, so the behavior change implied by a fix will affect headless contexts too. Flagging for reviewer awareness.

### DIFF
--- a/pkg/registry/provider_cached_authbug_test.go
+++ b/pkg/registry/provider_cached_authbug_test.go
@@ -1,0 +1,172 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package registry
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/stacklok/toolhive/pkg/registry/api"
+	"github.com/stacklok/toolhive/pkg/registry/auth"
+)
+
+// TestCachedProvider_AuthErrorNotMaskedByStaleCache reproduces a bug in
+// refreshCache() at provider_cached.go:116.
+//
+// Scenario: a user has an existing cache populated from a previous successful
+// fetch. Later, their registry returns 401 (token revoked, credentials
+// rejected server-side). The CLI should surface the auth error, because the
+// user's authorization state may have changed since the cache was populated.
+//
+// Current behavior: refreshCache() treats ANY error from the upstream fetch
+// as a signal to fall back to stale cached data, including authentication
+// errors. The CLI silently prints stale registry contents with no error.
+//
+// This test covers the 401/403 branch (server-side rejection).
+func TestCachedProvider_AuthErrorNotMaskedByStaleCache(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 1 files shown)


aponcedeleonch:
Thanks for putting this together. Heads up for future contributions: it's worth asking to have the issue assigned before opening a PR, per our [contributor guidelines](https://github.com/stacklok/toolhive/blob/main/CONTRIBUTING.md#claiming-an-issue), though in this case you filed the bug yourself so that's a minor point.

The good news is your reproducer did its job. The stale-cache auth bug in #4950 was fixed on main by #4981, and that fix actually merged this exact test file verbatim, plus it classifies token-acquisition failures at the source so the OAuth-flow path you predicted a sentinel-only fix would miss is covered too. I ran your two tests: they fail on the pre-fix base you wrote them against, and both pass on current main. So there's nothing left to merge here.

Going to close this as superseded, with thanks. The work effectively already shipped as part of the fix, and #4950 is now closed too.

PR #4953, opened 2026-04-21
Upgrade default uvx builder image to python:3.14-slim

JAORMX:
## Summary

- **Why**: Python 3.14 was released in Oct 2025 and is the current stable release. Bumping the default `uvx://` builder image picks up the latest runtime and security updates for new uvx-scheme workloads, and keeps the built-in default aligned with the current supported line.
- **What**: Change the default `RuntimeConfig.BuilderImage` for `TransportTypeUVX` from `python:3.13-slim` to `python:3.14-slim`, and update test expectations and docs accordingly. OpenAPI output regenerated via `task docs` picks up the doc-comment example change.

Users who need an older interpreter can still override via `--runtime-image python:3.13-slim` (or earlier) or a config file, exactly as before.

## Type of change

- [x] Other (default upgrade)

## Test plan

- [x] Cross-checked `python:3.14-slim` manifest digest against the Docker Hub API and the Docker Registry v2 API — both return `sha256:bc389f7dfcb21413e72a28f491985326994795e34d2b86c8ae2f417b4e7818aa`

Yes — new `uvx://`-scheme workloads built with default settings now use a Python 3.14 builder. Workloads with an explicit `--runtime-image` or a `runtime_configs.python.builder_image` entry are unaffected.

### DIFF
--- a/docs/arch/05-runconfig-and-permissions.md
+++ b/docs/arch/05-runconfig-and-permissions.md
@@ -92,7 +92,7 @@ The complete `RunConfig` struct is defined in `pkg/runner/config.go`.
 - `builder_image`: Override the default base image for the builder stage
   - Go: Default `golang:1.26-alpine`
   - Node: Default `node:22-alpine`
-  - Python: Default `python:3.13-slim`
+  - Python: Default `python:3.14-slim`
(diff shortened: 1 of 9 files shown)


samuv (dismissed):
:shipit:

rdimitrov (approved):

PR #4954, opened 2026-04-21
Upgrade default npx builder image to node:24-alpine

JAORMX:
## Summary

- **Why**: Node.js 24 is the current active LTS release (since Oct 2025); Node.js 22 has moved to maintenance LTS. Bumping the default `npx://` builder image keeps new npx-scheme workloads on the active LTS line with the latest runtime and security updates.
- **What**: Change the default `RuntimeConfig.BuilderImage` for `TransportTypeNPX` from `node:22-alpine` to `node:24-alpine`, and update test expectations and docs accordingly. OpenAPI output regenerated via `task docs` picks up the doc-comment example change.

Users who need an older runtime can still override via `--runtime-image node:22-alpine` (or earlier) or a config file, exactly as before.

## Type of change

- [x] Other (default upgrade)

## Test plan

- [x] Cross-checked `node:24-alpine` manifest digest against the Docker Hub API and the Docker Registry v2 API — both return `sha256:d1b3b4da11eefd5941e7f0b9cf17783fc99d9c6fc34884a665f40a06dbdfc94f` (last pushed 2026-04-16)

Yes — new `npx://`-scheme workloads built with default settings now use a Node 24 builder. Workloads with an explicit `--runtime-image` or a `runtime_configs.node.builder_image` entry are unaffected.

### DIFF
--- a/docs/arch/05-runconfig-and-permissions.md
+++ b/docs/arch/05-runconfig-and-permissions.md
@@ -91,7 +91,7 @@ The complete `RunConfig` struct is defined in `pkg/runner/config.go`.
 **Fields:**
 - `builder_image`: Override the default base image for the builder stage
   - Go: Default `golang:1.26-alpine`
-  - Node: Default `node:22-alpine`
+  - Node: Default `node:24-alpine`
   - Python: Default `python:3.14-slim`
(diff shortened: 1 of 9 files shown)


samuv (dismissed):
🚀

jhrozek (approved):

stacklokbot (approved):

PR #4955, opened 2026-04-21
Upgrade default go builder image to golang:1.26-alpine

JAORMX:
## Summary

- **Why**: Go 1.26 was released in Feb 2026 and is the current stable release. Bumping the default `go://` builder image keeps new go-scheme workloads on the current toolchain line with the latest runtime, stdlib, and security updates.
- **What**: Change the default `RuntimeConfig.BuilderImage` for `TransportTypeGO` from `golang:1.25-alpine` to `golang:1.26-alpine`, and update test expectations and docs accordingly. OpenAPI output regenerated via `task docs` picks up the doc-comment example change.
- **Drive-by doc fix**: Two references in `docs/runtime-version-customization.md` and `docs/arch/05-runconfig-and-permissions.md` claimed the Go runtime stage uses `alpine:3.22`, but `go.tmpl` already pins `alpine:3.23@sha256:...`. Updated both to match the code.

Users who need an older toolchain can still override via `--runtime-image golang:1.25-alpine` (or earlier) or a config file, exactly as before.

## Type of change

- [x] Other (default upgrade + minor doc correction)

## Test plan

- [x] Cross-checked `golang:1.26-alpine` manifest digest against the Docker Hub API and the Docker Registry v2 API — both return `sha256:f85330846cde1e57ca9ec309382da3b8e6ae3ab943d2739500e08c86393a21b1`

Yes — new `go://`-scheme workloads built with default settings now use a Go 1.26 builder. Workloads with an explicit `--runtime-image` or a `runtime_configs.go.builder_image` entry are unaffected.

The `alpine:3.23` digest currently pinned in `go.tmpl` is from an earlier snapshot of the 3.23 tag and is still a valid, working image — refreshing that digest is intentionally left for a separate PR to keep this change narrow.

### DIFF
--- a/docs/arch/05-runconfig-and-permissions.md
+++ b/docs/arch/05-runconfig-and-permissions.md
@@ -90,7 +90,7 @@ The complete `RunConfig` struct is defined in `pkg/runner/config.go`.
 
 **Fields:**
 - `builder_image`: Override the default base image for the builder stage
-  - Go: Default `golang:1.25-alpine`
+  - Go: Default `golang:1.26-alpine`
   - Node: Default `node:22-alpine`
   - Python: Default `python:3.13-slim`
 - `additional_packages`: Extra packages to install during the build and runtime stages (e.g., build tools, libraries)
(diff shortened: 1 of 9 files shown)


samuv (approved):
:shipit:

PR #4810, opened 2026-04-14
Add GetContent API for on-the-fly SKILL.md retrieval from OCI and git sources

samuv:
## Summary

- Add `GET /api/v1beta/skills/content?ref=<reference>` endpoint that retrieves SKILL.md content and file listings without installing the skill
- Support OCI references (`ghcr.io/org/skill:v1`), local build tags, git references (`git://github.com/org/repo#path`), HTTPS URLs, and registry catalog names (`io.github.stacklok/skill-creator`)
- Fix `resolveRegistryPackages` to include `Subfolder` as a `#path` fragment in git URLs — previously the subfolder field from the registry was silently dropped
- Extend `resolveFromRegistry` to accept qualified `namespace/name` identifiers for unambiguous catalog lookups

## Type of change

- [x] New feature
- [x] Bug fix

## Test plan

- [x] Unit tests (`task test`)
- [x] Linting (`task lint-fix`)

## Changes

| File | Change |
|------|--------|
| `pkg/skills/options.go` | Add `ContentOptions`, `SkillContent`, `SkillFileEntry` types |
| `pkg/skills/service.go` | Add `GetContent` to `SkillService` interface |
| `pkg/skills/mocks/mock_service.go` | Regenerated mock |
| `pkg/skills/skillsvc/skillsvc.go` | Implement `GetContent` with OCI, git, and registry fallback; fix subfolder handling; add `splitQualifiedName` |
| `pkg/api/v1/skills.go` | Add `GET /content` route and handler |
| `pkg/skills/client/client.go` | Add `GetContent` to HTTP client |
| `pkg/skills/client/client_test.go` | Tests for client GetContent |
| `pkg/skills/skillsvc/skillsvc_test.go` | Tests for git resolution, registry fallback, subfolder, splitQualifiedName |
| `docs/server/swagger.json` | New endpoint and schema definitions |

## Does this introduce a user-facing change?

Yes — new API endpoint allows the UI to display skill details (SKILL.md content and file list) directly from the registry without requiring installation first.

PR #4956, opened 2026-04-21
Return 502 Bad Gateway when OCI skill pull fails

samuv:
## Summary

- Follow-up on review feedback from #4810 ([discussion](https://github.com/stacklok/toolhive/pull/4810#discussion_r3082033217)): `registry.Pull` failures in `getContentFromOCI` and `installFromOCI` were mapped to `400 Bad Request`, but auth errors, unreachable registries, missing manifests, and network timeouts are upstream failures the caller cannot fix by changing their input.
- Return `502 Bad Gateway` instead, matching the git-resolve path in `getContentFromGit` which already uses `502` for the equivalent upstream failure.
- Parse / validation / "reference not found locally and registry not configured" paths continue to return `400` — those are genuine caller errors.

## Type of change

- [x] Bug fix (HTTP status semantics)

## Test plan

- [x] Unit tests (`go test ./pkg/skills/skillsvc/...`)
- [x] Existing tests updated to assert `502` where appropriate; `400` assertions for parse / missing-registry / invalid-reference cases left intact

## Changes

| File | Change |
|------|--------|
| `pkg/skills/skillsvc/content.go` | `getContentFromOCI`: `StatusBadRequest` → `StatusBadGateway` on `registry.Pull` failure |
| `pkg/skills/skillsvc/install_oci.go` | `installFromOCI`: `StatusBadRequest` → `StatusBadGateway` on `registry.Pull` failure |
| `pkg/skills/skillsvc/content_test.go` | Rename `pull failure propagates as 400` → `...as 502`; assert `StatusBadGateway` |
| `pkg/skills/skillsvc/install_oci_test.go` | Update `wantCode` to `StatusBadGateway` in four pull-failure fall-back tests (explicit tag, qualified name, digest ref, multi-segment ref) |

Yes — API clients that previously received `400 Bad Request` on a remote OCI pull failure now receive `502 Bad Gateway`. This is a more accurate signal that the failure originated upstream (registry/network) rather than from a malformed request, and matches the behaviour already in place for git pulls (and already documented in the swagger annotation introduced in #4810).

### DIFF
--- a/pkg/skills/skillsvc/content.go
+++ b/pkg/skills/skillsvc/content.go
@@ -170,7 +170,7 @@ func (s *service) getContentFromOCI(ctx context.Context, ref string) (*skills.Sk
 		if pullErr != nil {
 			return nil, httperr.WithCode(
 				fmt.Errorf("pulling OCI artifact %q: %w", qualifiedRef, pullErr),
-				http.StatusBadRequest,
+				http.StatusBadGateway,
 			)
 		}
 	}
--- a/pkg/skills/skillsvc/content_test.go
+++ b/pkg/skills/skillsvc/content_test.go
@@ -125,7 +125,7 @@ func TestGetContent(t *testing.T) {
 		assert.Equal(t, http.StatusBadRequest, httperr.Code(err))
 	})
 
-	t.Run("pull failure propagates as 400", func(t *testing.T) {
+	t.Run("pull failure propagates as 502", func(t *testing.T) {
 		t.Parallel()
 		ctrl := gomock.NewController(t)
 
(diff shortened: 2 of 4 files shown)


JAORMX (approved):

jhrozek:
One small follow-up on the swagger side (non-blocking, feel free to punt to a separate PR):

`installFromGit` already returns 502, and with this change `installFromOCI` does too — but the `@Failure` list on `installSkill` in `pkg/api/v1/skills.go` still only advertises 400/409/500. Might be worth adding `@Failure 502 {string} string "Bad Gateway"` there and regenerating the swagger docs so clients see the real contract. Pre-existing gap, not introduced here.

Issue #4972, opened 2026-04-21
Operator: add CI job that blocks breaking changes to the v1beta1 API


ChrisJBurns:
## Status: shipped

PRs that delivered this:

- **#4980** — initial advisory workflow that ran `crd-schema-checker` against the last-published CRD chart from OCI.
- **#4987** — simplified the approach: drop helm/OCI rendering entirely, compare the on-disk `files/crds/*.yaml` against `git show <last-release-tag>:<path>`. Also fixed a silent bug where the checker was only inspecting the first CRD in a multi-doc YAML.
- **#4991** — flipped the check from advisory to enforcing: `continue-on-error` removed, `api-break-allowed` label escape hatch added, `CONTRIBUTING.md` gained an "API Stability" section, PR template gained an "API Compatibility" section.
- **#4993** — UX + safety refinement: `labeled`/`unlabeled` added to the workflow's `pull_request` trigger so the escape hatch works on a single click and removing the label re-validates.

## What's running today

`.github/workflows/api-compat.yml` runs on any PR touching `cmd/thv-operator/api/**`, `deploy/charts/operator-crds/files/crds/**`, or the workflow itself. For each CRD file on HEAD, it:

1. Resolves the most recent release tag via `gh release list --limit 1`.
2. Shallow-fetches the tag.
3. For every `deploy/charts/operator-crds/files/crds/*.yaml`, runs `crd-schema-checker check-manifests` comparing `git show $TAG:<path>` (baseline) against the HEAD file.
4. Detects CRD removals via filename-set difference (the checker can't report removal).
5. Writes findings to the step log and the GitHub step summary.

The job is skipped when the `api-break-allowed` label is applied (coordinated with maintainers, per `CONTRIBUTING.md`).

## Scope — what "breaking" means

The check covers the **CRD schema surface** via `crd-schema-checker`:

- Field removal (`NoFieldRemoval`)
- Type change (`NoFieldTypeChange`)
- Required-field addition (`NoNewRequiredFields`)
- Narrowing validation (`NoMaxLengthDecrease`, `NoEnumRemoval`, etc.)
- Served/storage version handling (partial — the checker inspects schemas, not K8s-level version transitions)
- CRD removal (covered by the workflow itself via filename-set diff)

## Acceptance criteria

- [x] `.github/workflows/api-compat.yml` created and running on PRs that touch operator API surface. (#4980)
[...]

PR #4980, opened 2026-04-21
Add advisory CRD schema compatibility CI check

ChrisJBurns:
## Summary

The v1beta1 graduation in #4849 made a stability commitment that is enforced only by convention and code review. A reviewer can miss a field removal, tightened validation, or renamed finalizer and ship a breaking change to users of the released CRDs.

This PR adds the **Phase 1 (advisory)** part of the API compatibility CI plan: a new `.github/workflows/api-compat.yml` workflow that runs `crd-schema-checker` against the last published CRD chart on every PR touching operator API surface. Findings appear in the job's step summary. The job uses `continue-on-error: true` so breaking-change findings do not block merges yet — that flips in Phase 2 after ~2 weeks of observation.

Part of #4972.

<details>

- New top-level workflow `.github/workflows/api-compat.yml`, triggered on PRs that modify `cmd/thv-operator/api/**`, `deploy/charts/operator-crds/files/crds/**`, or `deploy/charts/operator-crds/templates/**`.
- Single job `crd-schema-check` that: resolves a baseline (published chart from `oci://ghcr.io/stacklok/toolhive/toolhive-operator-crds` for the latest tag, falling back to `origin/main` when the chart for the tag is not yet published); renders baseline and HEAD CRDs via `helm template`; runs `crd-schema-checker` with `NoBools,NoMaps` disabled; writes a markdown summary (baseline source + Compatible/Incompatible + raw output) to `\$GITHUB_STEP_SUMMARY`.
- `continue-on-error: true` at the job level: the step summary is visible and the check-run reports its real status, but the workflow's overall result doesn't block the PR. Phase 2 removes this flag.
- Phases 2 (enforcement + `api-break-allowed` label + CONTRIBUTING/template updates) and 3 (`apidiff` + toolhive invariants) ship as separate PRs after Phase 1 has been observed.

</details>

<details>

| File | Change |
|------|--------|
| `.github/workflows/api-compat.yml` | New workflow. Pinned action SHAs (checkout v6, setup-go v6, setup-helm v4.3.1) match operator-ci.yml. Installs `crd-schema-checker` via \`go install @latest\` (acceptable for Phase 1; pin after stability confirmed). Uses \`\${PIPESTATUS[0]}\` to capture the checker's exit code through the \`tee\` pipeline. |

</details>

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Dependency update
- [ ] Documentation
- [ ] Other

## Test plan

- [x] Workflow YAML parses cleanly (\`python3 -c 'import yaml; yaml.safe_load(...)'\`)
- [x] \`crd-schema-checker\` main-vs-main: exit 0, no output (clean baseline, no existing noise)
- [x] \`crd-schema-checker\` v0.22.0-vs-main with \`--disabled-validators=NoBools,NoMaps\`: exit 0, no output
- [x] Simulated field-removal (\`yq -i 'del(.. | select(has(\"replicas\")).replicas)'\`): correctly flagged as \`NoFieldRemoval\`, exit 1
- [x] \`helm pull oci://.../toolhive-operator-crds --version 0.22.0\`: succeeds — OCI pull path works
- [x] \`helm pull\` for the latest tag (v0.23.0): fails — confirms the \`origin/main\` fallback is exercised on this very PR since the v0.23.0 chart isn't yet published
- [ ] Unit tests (\`task test\`) — **N/A**, no Go changes
- [ ] E2E tests (\`task test-e2e\`) — **N/A**, no runtime changes
- [ ] Linting (\`task lint-fix\`) — **N/A**, no Go changes
- [x] Manual end-to-end verification will happen on this PR's own Actions run, which is the first real exercise of the workflow

No. Internal CI only. The \`api-break-allowed\` label and \`CONTRIBUTING.md\` "API Stability" section land in the Phase 2 PR.

## Implementation plan

<details>

Planned with Claude Code as a three-phase rollout (advisory → enforcement + docs → expanded invariants). This PR is Stage 1 only; Stages 2 and 3 will ship as separate PRs.

**Stage 1 scope** (this PR):
- Create \`.github/workflows/api-compat.yml\` as a top-level workflow with \`on: pull_request: paths:\` filter.
- Baseline resolution: OCI chart for last release tag, with \`origin/main\` fallback when \`helm pull\` fails.
- Single \`crd-schema-check\` job running with \`continue-on-error: true\`.
- Step summary output format: heading + baseline source + Compatible/Incompatible/Unknown status + fenced block with raw checker output.

**Key design decisions surfaced in this PR**:
- Top-level workflow (not reusable) to get \`paths:\` filter support; trade-off: won't appear under \`PR Checks\` umbrella, each job is its own check run.
- Style validators \`NoBools\` and \`NoMaps\` disabled — they are OpenShift API conventions, not compat-breaking rules. Verified locally that \`NoFieldRemoval\` and other real compat rules still fire with them disabled.
- \`go install ...@latest\` for \`crd-schema-checker\` accepted for Phase 1; pin once observation confirms stability.
- Label-skip guard (\`if: !contains(...'api-break-allowed'...)\`) is absent here — the label doesn't exist yet and is created alongside the Phase 2 enforcement flip to prevent early misuse.

**Follow-up PRs** (separate \`/implement-code\` runs):
- Stage 2: remove \`continue-on-error\`; add \`if: !contains(...)\` guard; create \`api-break-allowed\` label via \`gh label create\` runbook; update \`CONTRIBUTING.md\` with an "API Stability" section; add "API Compatibility" section to \`.github/pull_request_template.md\`; maintainer action: add \`crd-schema-check\` to required status checks.
- Stage 3: add \`go-api-check\` job running \`apidiff\` on \`./cmd/thv-operator/api/v1beta1\`; add \`toolhive-invariants\` job checking \`spec.names\` stability and finalizer-string stability (regex narrowed to \`/finalizer\` suffix to avoid false-positives on \`restart-*\` annotations).

</details>

- **First-run behavior**: this PR's own Actions run will hit the fallback path (v0.23.0's CRD chart isn't on OCI yet). The step summary should report \`Baseline source: main\` and \`Status: Compatible\`. If it reports anything else, treat as a real finding rather than workflow-setup noise.
- **Disabled validators**: \`NoBools,NoMaps\` are intentionally disabled. A \`main\` vs last-release run without them produces 9 spurious errors on \`embeddingservers\` fields added since v0.22.0. These validators are style rules from OpenShift, not compat-breaking semantics. Local verification confirmed real compat rules (\`NoFieldRemoval\`, etc.) still fire.
- **Check run will show a red X when findings appear**: \`continue-on-error: true\` prevents the workflow from blocking the PR, but GitHub still reports the job's real status on the check run. Reviewers should treat red X on this job as "read the step summary" rather than "this PR is broken" during Phase 1.
- **Tool pin deferred**: \`crd-schema-checker\` installs from \`@latest\`. If upstream breaks semantics in a release, Phase 1's advisory mode absorbs the noise. Pin after Phase 1 stabilizes.

### DIFF
--- a/.github/workflows/api-compat.yml
+++ b/.github/workflows/api-compat.yml
@@ -14,8 +14,16 @@ on:
   pull_request:
     paths:
       - 'cmd/thv-operator/api/**'
+      # files/crds is the source of truth — controller-gen emits here, and
+      # crd-helm-wrapper copies from here into templates/. Any drift in
+      # templates/ is caught by operator-ci.yml's generate-crds job, so
+      # watching templates/ would be redundant. values.yaml and the
+      # crd-helm-wrapper only affect Helm conditionals and annotations the
+      # checker ignores; the workflow's explicit --set flags force every
+      # CRD to render regardless of those files.
       - 'deploy/charts/operator-crds/files/crds/**'
-      - 'deploy/charts/operator-crds/templates/**'
+      # Self-exercise the workflow when the workflow itself changes.
+      - '.github/workflows/api-compat.yml'
 
 permissions:
   contents: read
@@ -26,6 +34,11 @@ jobs:
     runs-on: ubuntu-latest
     # Phase 1: advisory only. Remove in Phase 2 to enforce.
     continue-on-error: true
+    # Expected runtime is ~3 minutes (checkout + go/helm setup + helm pull +
+    # go install + two renders + checker). 10 minutes is a cheap upper bound
+    # that protects against a hung helm pull or go install without being
+    # disruptive to legitimate slow runs.
+    timeout-minutes: 10
     steps:
       - name: Checkout PR HEAD
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
@@ -55,9 +68,13 @@ jobs:
           if [ -n "$LATEST_VERSION" ]; then
             echo "Attempting to pull published chart for $LATEST_TAG..."
             mkdir -p /tmp/baseline-release
+            # Don't redirect stderr — helm pull's own error line (one line,
+            # e.g. "Error: ... not found") is useful evidence in the log
+            # when we fall back. Redirecting hides transient ghcr.io outages
+            # behind the expected "tag not yet published" path.
             if helm pull "oci://ghcr.io/stacklok/toolhive/toolhive-operator-crds" \
                  --version "$LATEST_VERSION" \
-                 --untar --untardir /tmp/baseline-release 2>/dev/null; then
+                 --untar --untardir /tmp/baseline-release; then
               BASELINE_SOURCE="release $LATEST_TAG"
               BASELINE_CHART_DIR="/tmp/baseline-release/toolhive-operator-crds"
               echo "Using published chart at $LATEST_TAG"
@@ -79,6 +96,13 @@ jobs:
           echo "chart-dir=$BASELINE_CHART_DIR" >> "$GITHUB_OUTPUT"
 
       - name: Render baseline CRDs
+        env:
+          # Route step outputs through env vars so bash quotes them instead
+          # of the runner substituting them directly into the script body.
+          # Current values are hardcoded literals, so not injectable today —
+          # this is defense-in-depth against a future edit that routes a
+          # PR-controlled string through these outputs.
+          CHART_DIR: ${{ steps.baseline.outputs.chart-dir }}
         run: |
           set -euo pipefail
           mkdir -p /tmp/baseline-crds
@@ -88,7 +112,7 @@ jobs:
           # The yq filter is defensive — the chart currently only templates
           # CRDs, but this keeps the invariant explicit if non-CRD resources
           # are ever added.
-          helm template toolhive-operator-crds "${{ steps.baseline.outputs.chart-dir }}" \
+          helm template toolhive-operator-crds "$CHART_DIR" \
             --set crds.install.server=true \
             --set crds.install.registry=true \
             --set crds.install.virtualMcp=true \
@@ -109,10 +133,19 @@ jobs:
           echo "Rendered $(grep -c '^kind: CustomResourceDefinition' /tmp/head-crds/all.yaml || echo 0) HEAD CRDs"
 
       - name: Install crd-schema-checker
-        run: go install github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@latest
+        # SHA-pinned: openshift/crd-schema-checker has no release tags at the
+        # time of writing, so @latest is the only other option. Pinning makes
+        # CI deterministic and mitigates supply-chain risk (upstream compromise
+        # would otherwise execute attacker code on the runner with GITHUB_TOKEN
+        # in env). Bump via a deliberate PR after verifying the new output
+        # locally. SHA pinned on 2026-04-21.
+        run: go install github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@3fee146022bfe6f4adf84998de35d7267b864bef
 
       - name: Check CRD schema compatibility
         id: checker
+        env:
+          # See Render baseline CRDs step for rationale on env-based interpolation.
+          BASELINE_SOURCE: ${{ steps.baseline.outputs.source }}
         run: |
           # NoBools and NoMaps are OpenShift API-style conventions, not
           # compat-breaking rules. They fire on fields we legitimately use
@@ -137,7 +170,7 @@ jobs:
           {
             echo "## API Compatibility — CRD Schema Check (Phase 1: advisory)"
             echo ""
-            echo "**Baseline source**: ${{ steps.baseline.outputs.source }}"
+            echo "**Baseline source**: $BASELINE_SOURCE"
             echo "**Status**: $STATUS"
             echo ""
             echo "<details><summary>crd-schema-checker output</summary>"
--- a/.github/workflows/api-compat.yml
+++ b/.github/workflows/api-compat.yml
@@ -59,41 +59,28 @@ jobs:
         run: |
           set -euo pipefail
 
-          BASELINE_SOURCE=""
-          BASELINE_CHART_DIR=""
-
-          LATEST_TAG="$(gh release list --repo "$GITHUB_REPOSITORY" --limit 1 --json tagName --jq '.[0].tagName' 2>/dev/null || true)"
+          # Baseline is always the most recent published release chart from
+          # OCI. If the pull fails (no releases, OCI outage, or the chart
+          # for the most recent tag is not yet published), the job fails —
+          # an API-touching PR must be checked against a real released
+          # baseline or not at all. Falling back to origin/main was
+          # rejected in review: it can silently compare against an
+          # already-broken baseline once a break lands on main.
+          LATEST_TAG="$(gh release list --repo "$GITHUB_REPOSITORY" --limit 1 --json tagName --jq '.[0].tagName')"
+          if [ -z "$LATEST_TAG" ]; then
+            echo "::error::No releases found for $GITHUB_REPOSITORY; cannot establish an API compatibility baseline."
+            exit 1
+          fi
           LATEST_VERSION="${LATEST_TAG#v}"
 
-          if [ -n "$LATEST_VERSION" ]; then
-            echo "Attempting to pull published chart for $LATEST_TAG..."
-            mkdir -p /tmp/baseline-release
-            # Don't redirect stderr — helm pull's own error line (one line,
-            # e.g. "Error: ... not found") is useful evidence in the log
-            # when we fall back. Redirecting hides transient ghcr.io outages
-            # behind the expected "tag not yet published" path.
-            if helm pull "oci://ghcr.io/stacklok/toolhive/toolhive-operator-crds" \
-                 --version "$LATEST_VERSION" \
-                 --untar --untardir /tmp/baseline-release; then
-              BASELINE_SOURCE="release $LATEST_TAG"
-              BASELINE_CHART_DIR="/tmp/baseline-release/toolhive-operator-crds"
-              echo "Using published chart at $LATEST_TAG"
-            else
-              echo "Chart pull failed (not yet published?); falling back to origin/main"
-            fi
-          else
-            echo "No releases found; falling back to origin/main"
-          fi
+          echo "Pulling published chart for $LATEST_TAG..."
+          mkdir -p /tmp/baseline-release
+          helm pull "oci://ghcr.io/stacklok/toolhive/toolhive-operator-crds" \
+            --version "$LATEST_VERSION" \
+            --untar --untardir /tmp/baseline-release
 
-          if [ -z "$BASELINE_SOURCE" ]; then
-            git fetch origin main --depth=1
-            git worktree add --detach /tmp/baseline-main FETCH_HEAD
-            BASELINE_SOURCE="main"
-            BASELINE_CHART_DIR="/tmp/baseline-main/deploy/charts/operator-crds"
-          fi
-
-          echo "source=$BASELINE_SOURCE" >> "$GITHUB_OUTPUT"
-          echo "chart-dir=$BASELINE_CHART_DIR" >> "$GITHUB_OUTPUT"
+          echo "source=release $LATEST_TAG" >> "$GITHUB_OUTPUT"
+          echo "chart-dir=/tmp/baseline-release/toolhive-operator-crds" >> "$GITHUB_OUTPUT"
 
       - name: Render baseline CRDs
         env:
--- a/.github/workflows/api-compat.yml
+++ b/.github/workflows/api-compat.yml
@@ -6,9 +6,6 @@ name: API Compatibility
 # continue-on-error: true so breaking-change findings are surfaced via the
 # job's step summary but do not block merges. Remove continue-on-error in
 # Phase 2 to start enforcing.
-#
-# See design-api-compat-ci.md and CONTRIBUTING.md ("API Stability") for the
-# full rollout plan and escape-hatch process.
 
 on:
   pull_request:
--- a/.github/workflows/api-compat.yml
+++ b/.github/workflows/api-compat.yml
@@ -1,0 +1,152 @@
+name: API Compatibility
+
+# This workflow guards the stability of the v1beta1 operator API surface.
+#
+# Phase 1 (current): advisory only. `crd-schema-check` runs with
+# continue-on-error: true so breaking-change findings are surfaced via the
+# job's step summary but do not block merges. Remove continue-on-error in
+# Phase 2 to start enforcing.
+#
+# See design-api-compat-ci.md and CONTRIBUTING.md ("API Stability") for the
+# full rollout plan and escape-hatch process.
+
+on:
+  pull_request:
+    paths:
+      - 'cmd/thv-operator/api/**'
+      - 'deploy/charts/operator-crds/files/crds/**'
+      - 'deploy/charts/operator-crds/templates/**'
+
+permissions:
+  contents: read
+
+jobs:
+  crd-schema-check:
+    name: CRD Schema Compatibility
+    runs-on: ubuntu-latest
+    # Phase 1: advisory only. Remove in Phase 2 to enforce.
+    continue-on-error: true
+    steps:
+      - name: Checkout PR HEAD
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Set up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
+        with:
+          go-version: 'stable'
+          cache: true
+
+      - name: Set up Helm
+        uses: azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4 # v4.3.1
+
+      - name: Resolve baseline source
+        id: baseline
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+
+          BASELINE_SOURCE=""
+          BASELINE_CHART_DIR=""
+
+          LATEST_TAG="$(gh release list --repo "$GITHUB_REPOSITORY" --limit 1 --json tagName --jq '.[0].tagName' 2>/dev/null || true)"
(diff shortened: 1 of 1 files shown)


ChrisJBurns (commented):
**Agents consulted**: security-advisor, kubernetes-expert, code-reviewer

### Consensus Summary
| 1 | Unpinned `go install @latest` for crd-schema-checker | 10/10 | MEDIUM | Fix before Phase 2 |
| 2 | Missing job-level `timeout-minutes` | 9/10 | MEDIUM | Fix |
| 3 | `helm pull ... 2>/dev/null` swallows transient failures | 8/10 | MEDIUM | Fix |
| 4 | Path filter misses `values.yaml`/`Chart.yaml` | 8/10 | MEDIUM | Fix |
| 5 | `origin/main` fallback masks already-merged breaks | 8/10 | MEDIUM | Discuss |
| 6 | crd-schema-checker misses scope/served/storage/conversion | 8/10 | MEDIUM | Phase 3 follow-up |
| 7 | `${{ outputs }}` interpolated into shell | 8/10 | LOW | Fix (defense-in-depth) |
| 8 | `yq` multi-doc filter has no count floor | 7/10 | MEDIUM | Fix |

### Overall
Phase 1 advisory rollout is sound and the PR description already acknowledges the biggest tradeoffs (unpinned checker, disabled NoBools/NoMaps validators). The workflow itself is well-commented, correctly pins action SHAs, uses minimal `contents: read` permissions, and handles `pipefail` + `PIPESTATUS` correctly around the `tee` pipeline.

Findings cluster into three buckets: robustness (no job timeout, stderr-swallowing `helm pull`), correctness (path filter missing `values.yaml`/`Chart.yaml`, origin/main fallback can compare against already-broken baselines), and defense-in-depth (`${{ outputs }}` interpolated directly into shell, no floor check on rendered CRD count). None block Phase 1 advisory mode. The `@latest` pin (F1) is the one item that must close before Phase 2 enforcement flips on — `continue-on-error` does not mitigate supply-chain exposure, only check-run visibility.

F5 (tag-based fallback) and F6 (crd-schema-checker coverage gaps) are worth discussing as design questions; the rest are small, contained edits.

---
*Generated with Claude Code*

.github/workflows/api-compat.yml
```diff
@@ -0,0 +1,152 @@
...
+          helm template toolhive-operator-crds deploy/charts/operator-crds \
+            --set crds.install.server=true \
+            --set crds.install.registry=true \
+            --set crds.install.virtualMcp=true \
+            | yq 'select(.kind == "CustomResourceDefinition")' \
+            > /tmp/head-crds/all.yaml
+          echo "Rendered $(grep -c '^kind: CustomResourceDefinition' /tmp/head-crds/all.yaml || echo 0) HEAD CRDs"
+
+      - name: Install crd-schema-checker
+        run: go install github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@latest
```
ChrisJBurns: **Unpinned `go install @latest`** (consensus 10/10, unanimous across security-advisor, kubernetes-expert, code-reviewer)

`@latest` resolves tip-of-tree at run time — two runs of the same PR can return different verdicts, and a compromised upstream would execute attacker-controlled code on the runner with access to `GITHUB_TOKEN` and the module cache. `continue-on-error: true` only suppresses the red X; the step still runs to completion. The "advisory" flag is an enforcement toggle, not a supply-chain mitigation.

Must close before Phase 2.

```suggestion
      - name: Install crd-schema-checker
        # TODO(phase-2): pin before removing continue-on-error
        run: go install github.com/openshift/crd-schema-checker/cmd/crd-schema-checker@latest
```

.github/workflows/api-compat.yml
```diff
@@ -0,0 +1,152 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  crd-schema-check:
+    name: CRD Schema Compatibility
+    runs-on: ubuntu-latest
+    # Phase 1: advisory only. Remove in Phase 2 to enforce.
+    continue-on-error: true
```
ChrisJBurns, line 33: **Missing job-level `timeout-minutes`** (consensus 9/10, code-reviewer)

No timeout declared → inherits GitHub's 360-minute default. Sibling workflows in the repo set explicit bounds (operator-ci.yml → 30, test-e2e-lifecycle.yml → 30). A hung `helm pull` or stuck `go install` could burn 6h per PR.

```suggestion
    # Phase 1: advisory only. Remove in Phase 2 to enforce.
    continue-on-error: true
    timeout-minutes: 15
```

.github/workflows/api-compat.yml
```diff
@@ -0,0 +1,152 @@
...
+
+          LATEST_TAG="$(gh release list --repo "$GITHUB_REPOSITORY" --limit 1 --json tagName --jq '.[0].tagName' 2>/dev/null || true)"
+          LATEST_VERSION="${LATEST_TAG#v}"
+
+          if [ -n "$LATEST_VERSION" ]; then
+            echo "Attempting to pull published chart for $LATEST_TAG..."
+            mkdir -p /tmp/baseline-release
+            if helm pull "oci://ghcr.io/stacklok/toolhive/toolhive-operator-crds" \
+                 --version "$LATEST_VERSION" \
+                 --untar --untardir /tmp/baseline-release 2>/dev/null; then
```
ChrisJBurns: **`helm pull ... 2>/dev/null` swallows transient failures** (consensus 8/10, code-reviewer)

`2>/dev/null` collapses "not yet published" (expected) and "ghcr.io transient outage" (unexpected) into the same fallback path. A brief registry outage silently shifts the baseline from the last release to `origin/main` with no indication to the reviewer that the shift was unintended.

Suggested fix: capture stderr to a temp file and print it on the fallback branch so the step summary explains the shift, e.g. `if helm pull ... 2>/tmp/helm-err; then ... else echo "helm pull failed:"; cat /tmp/helm-err; fi`.

.github/workflows/api-compat.yml
```diff
@@ -0,0 +1,152 @@
...
+#
+# See design-api-compat-ci.md and CONTRIBUTING.md ("API Stability") for the
+# full rollout plan and escape-hatch process.
+
+on:
+  pull_request:
+    paths:
+      - 'cmd/thv-operator/api/**'
+      - 'deploy/charts/operator-crds/files/crds/**'
+      - 'deploy/charts/operator-crds/templates/**'
```
ChrisJBurns: **Path filter misses `values.yaml` and `Chart.yaml`** (consensus 8/10, kubernetes-expert specialist)

Rendered CRDs also depend on `deploy/charts/operator-crds/values.yaml` (feature-flag defaults, `crds.keep` annotation) and `Chart.yaml` (chart name/version). A PR renaming a feature flag in `values.yaml` (e.g. `virtualMcp` → `virtualMCP`) without editing `templates/` would bypass the workflow.

```suggestion
      - 'cmd/thv-operator/api/**'
      - 'deploy/charts/operator-crds/files/crds/**'
      - 'deploy/charts/operator-crds/templates/**'
      - 'deploy/charts/operator-crds/values.yaml'
      - 'deploy/charts/operator-crds/Chart.yaml'
```

.github/workflows/api-compat.yml
```diff
@@ -0,0 +1,152 @@
...
+          # NoBools and NoMaps are OpenShift API-style conventions, not
+          # compat-breaking rules. They fire on fields we legitimately use
+          # (e.g. embeddingservers.spec.modelCache.enabled) and drown out
+          # real findings. Re-enable only if upstream clarifies breaking-
+          # change semantics for them.
+          set +e
+          crd-schema-checker check-manifests \
+            --existing-crd-filename /tmp/baseline-crds/all.yaml \
+            --new-crd-filename /tmp/head-crds/all.yaml \
+            --disabled-validators=NoBools,NoMaps \
```
ChrisJBurns, line 143: **crd-schema-checker doesn't cover scope, served/storage, conversion** (consensus 8/10, kubernetes-expert specialist)

The checker focuses on OpenAPI schema compat. It does NOT validate `spec.scope` flips (Namespaced↔Cluster), `spec.versions[*].served/storage` changes (requires conversion migration), `spec.conversion` webhook changes, or `spec.names.{plural,singular,shortNames,categories}` renames. The `mcpservers` CRD already has two versions with served/storage flags — exactly the shape that can silently slip through.

Not blocking for this PR — the PR description's Phase 3 `toolhive-invariants` job is the right home. Confirm those invariants cover these fields in the Phase 3 design.

.github/workflows/api-compat.yml
```diff
@@ -0,0 +1,152 @@
...
+        run: |
+          set -euo pipefail
+          mkdir -p /tmp/baseline-crds
+          # All three feature flags must be set so the Helm wrapper emits every
+          # CRD (see deploy/charts/operator-crds/crd-helm-wrapper). Missing a
+          # flag silently drops CRDs from the comparison.
+          # The yq filter is defensive — the chart currently only templates
+          # CRDs, but this keeps the invariant explicit if non-CRD resources
+          # are ever added.
+          helm template toolhive-operator-crds "${{ steps.baseline.outputs.chart-dir }}" \
```
ChrisJBurns: **`${{ outputs }}` interpolated directly into shell — defense-in-depth** (consensus 8/10, security-advisor specialist)

`${{ steps.baseline.outputs.chart-dir }}` and `${{ ... source }}` are expanded by the runner *before* bash parses the script — the classic Actions script-injection pattern. Current values aren't attacker-controllable (hardcoded strings + `gh release list` on the base repo), so this is defense-in-depth rather than an active bug. A future edit routing any PR-controlled string through `steps.baseline.outputs.*` becomes injectable silently.

```suggestion
      - name: Render baseline CRDs
        env:
          CHART_DIR: ${{ steps.baseline.outputs.chart-dir }}
        run: |
          set -euo pipefail
          mkdir -p /tmp/baseline-crds
          # All three feature flags must be set so the Helm wrapper emits every
          # CRD (see deploy/charts/operator-crds/crd-helm-wrapper). Missing a
          # flag silently drops CRDs from the comparison.
          # The yq filter is defensive — the chart currently only templates
          # CRDs, but this keeps the invariant explicit if non-CRD resources
          # are ever added.
          helm template toolhive-operator-crds "$CHART_DIR" \
```

Apply the same treatment to `${{ steps.baseline.outputs.source }}` on line 140.

.github/workflows/api-compat.yml
```diff
@@ -0,0 +1,152 @@
...
+            --set crds.install.server=true \
+            --set crds.install.registry=true \
+            --set crds.install.virtualMcp=true \
+            | yq 'select(.kind == "CustomResourceDefinition")' \
+            > /tmp/baseline-crds/all.yaml
+          echo "Rendered $(grep -c '^kind: CustomResourceDefinition' /tmp/baseline-crds/all.yaml || echo 0) baseline CRDs"
```
ChrisJBurns, line 105: **`yq` filter has no CRD count floor** (consensus 7/10, kubernetes-expert specialist)

If the `yq 'select(.kind == "CustomResourceDefinition")'` filter ever silently drops docs (yq upgrade, operator-precedence change), the checker passes on a reduced set and reports "Compatible". The `grep -c ... || echo 0` logs the count but does not assert a floor. The chart currently renders 12 CRDs.

Suggestion: `COUNT=$(grep -c '^kind: CustomResourceDefinition' /tmp/baseline-crds/all.yaml || echo 0); [ "$COUNT" -ge 12 ] || { echo "Expected >=12 CRDs, got $COUNT"; exit 1; }`. Apply to both the baseline and HEAD render steps.

.github/workflows/api-compat.yml
ChrisJBurns: Fixed in 009e8346, but went simpler than the suggested temp-file dance: dropped `2>/dev/null` entirely.

`helm pull`'s own stderr is one clean line on failure (`Error: ... not found` or network error) — that's useful evidence in the log, not noise. The only reason to redirect was cosmetic. Second-opinion reviewer made the same call.

Subsequent commit 5ba27a6d removed the fallback path altogether (see F5 reply), so this stderr is now printed once as the job's actual failure reason before exit.

.github/workflows/api-compat.yml
ChrisJBurns: Fixed in 009e8346, but not exactly as proposed — final path filter is narrower:

**Final filter:**
```yaml
- 'cmd/thv-operator/api/**'
- 'deploy/charts/operator-crds/files/crds/**'
- '.github/workflows/api-compat.yml'
```

**Rationale** (per second-opinion analysis):
- `templates/**` dropped — it's a generated duplicate of `files/crds/`; any drift is caught by `operator-ci.yml`'s `generate-crds` job.
- `values.yaml` dropped — the workflow sets `--set crds.install.server=true --set crds.install.registry=true --set crds.install.virtualMcp=true` explicitly, overriding any default. `crds.keep` only affects an annotation the checker ignores.
- `crd-helm-wrapper/**` dropped — it only adds Helm conditionals and the `helm.sh/resource-policy: keep` annotation; the checker compares schemas, not metadata annotations.
- `api-compat.yml` added — self-exercise the workflow on its own changes. This is why the second commit triggered a real checker run on the PR itself.
- `Chart.yaml` not added — chart version/name doesn't affect API semantics.

.github/workflows/api-compat.yml
ChrisJBurns: Fixed in 009e8346. SHA-pinned to `3fee146022bfe6f4adf84998de35d7267b864bef` with a dated comment.

Verified `openshift/crd-schema-checker` has **no release tags** (`git ls-remote --tags` returns empty), so SHA pin is the only option. Bump is a deliberate PR.

Second-opinion agent confirmed that `continue-on-error: true` does **not** sandbox the step — the `go install` and checker both execute fully, only the red X is suppressed. Severity rerated to 7/10 (advisory job, `contents: read` only, no secrets wired in), but the "must close before Phase 2 enforcement flip" timing stands.

.github/workflows/api-compat.yml
ChrisJBurns: Fixed in 009e8346. Set `timeout-minutes: 10`, not 15.

Second-opinion pushback on severity: the "sibling workflows set explicit bounds" framing is misleading. In `operator-ci.yml`, only the heavy `e2e-tests-operator` matrix job sets `timeout-minutes: 30` — the lighter jobs (`operator-tests`, `build-operator`, `generate-crds`) have no timeout and inherit the 360-minute default. So the convention is "bound on expensive kind/matrix jobs," not "bound everywhere."

Expected runtime is ~3 min (checkout + setup + helm pull + go install + 2× template + checker). 10 is a cheap upper bound; 15 was overcautious by comparison to sibling lighter jobs.

.github/workflows/api-compat.yml
```diff
@@ -0,0 +1,152 @@
...
+            fi
+          else
+            echo "No releases found; falling back to origin/main"
+          fi
+
+          if [ -z "$BASELINE_SOURCE" ]; then
+            git fetch origin main --depth=1
+            git worktree add --detach /tmp/baseline-main FETCH_HEAD
+            BASELINE_SOURCE="main"
+            BASELINE_CHART_DIR="/tmp/baseline-main/deploy/charts/operator-crds"
```
ChrisJBurns: Fixed in 5ba27a6d. Removed the `origin/main` fallback entirely and fail-closed on `helm pull` failure.

Matches your preference: a release chart always exists; if the pull fails, the job should fail rather than emit a false green.

Phase 1's `continue-on-error: true` absorbs the release-to-OCI publish-window gap — the PR is not blocked even though the job reports a failure during those minutes. Phase 2 will cover that window via the `api-break-allowed` escape hatch if needed.

.github/workflows/api-compat.yml
ChrisJBurns: Correctly deferred to Phase 3 per the consensus note.

The Phase 3 `toolhive-invariants` job will expand from 2 sub-checks to 5:

1. **`spec.names` equality** — plural, singular, kind, listKind, shortNames, categories.
2. **`spec.scope` equality** — catches Namespaced↔Cluster flips. All 12 CRDs are Namespaced today, so the check has a clean baseline.
3. **Served/storage transition rules** — forbids un-serving a served version without a deprecation cycle, forbids storage-version regression, asserts exactly one `storage: true` on HEAD.
4. **`spec.conversion.strategy` check** — forbids removing a conversion webhook when multiple stored versions exist. All 12 CRDs currently have no conversion webhook; transitions to `Webhook` will be advisory-only.
5. **Finalizer-string check** — narrow `/finalizer` suffix regex to avoid matching annotation strings.

Will implement when Stage 3 ships as its own PR, along with draft-PR test cases for each sub-check.

.github/workflows/api-compat.yml
ChrisJBurns: Fixed in 009e8346. Routed both outputs through `env:` blocks:

- `CHART_DIR: ${{ steps.baseline.outputs.chart-dir }}` on the `Render baseline CRDs` step, used as `"$CHART_DIR"` in the `helm template` call.
- `BASELINE_SOURCE: ${{ steps.baseline.outputs.source }}` on the `Check CRD schema compatibility` step, used as `"$BASELINE_SOURCE"` in the step summary.

Second-opinion agent agreed this is defense-in-depth only — no current injection path since both values are hardcoded literals in the baseline step. Severity rerated 4/10 (not 8/10), but worth doing now while the diff is tiny to prevent silent regression if a future edit routes PR-controlled data through these outputs.

.github/workflows/api-compat.yml
ChrisJBurns: **Rejecting** based on second-opinion pushback.

A hardcoded `>=12` floor guarantees a false positive the next time a CRD is legitimately added or removed. Compare to 3dbbf5ae (v1alpha1 graduation cleanup) — a similar-shape change that would have tripped a fixed floor.

The failure mode it guards against requires three coincident events: (1) `yq` silently changes filter semantics in a future upgrade, (2) the change drops CRDs without erroring, AND (3) the reviewer misses the `Rendered N baseline CRDs` log line that's already printed. That's a lower-probability failure than the false-positive it introduces.

Additionally, `crd-schema-checker` already reports a removed CRD as a breaking change — adding a floor would duplicate the guard at lower fidelity.

If paranoia persists after Phase 1 observation, a defensible alternative is a zero-floor (`[ $COUNT -gt 0 ]`) — that survives schema evolution and still catches total-drop scenarios. Happy to revisit if the Phase 1 observation surfaces a real case.

rdimitrov (approved):

PR #4988, opened 2026-04-21
TEST: rename mcpgroups.spec.description → summary (do not merge)

ChrisJBurns:
## ⚠️ DO NOT MERGE

Deliberate breaking change to validate the api-compat workflow merged in #4987 end-to-end on real CI. **Close this PR** once the check has been demonstrated.

## What changed

In `cmd/thv-operator/api/v1beta1/mcpgroup_types.go`, the JSON tag on `MCPGroupSpec.Description` is renamed from `"description"` to `"summary"`. The Go field name is kept (so the 20+ tests that reference `.Spec.Description` still compile), but the CRD schema emitted by `controller-gen` now exposes `spec.summary` instead of `spec.description`.

`task operator-generate` and `task operator-manifests` regenerated the three affected files:

- `cmd/thv-operator/api/v1beta1/mcpgroup_types.go`
- `deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpgroups.yaml`
- `deploy/charts/operator-crds/templates/toolhive.stacklok.dev_mcpgroups.yaml`

## Why this is a breaking change

Dropping `description` from the CRD schema means: existing `MCPGroup` resources in user clusters that have `spec.description` set will no longer survive a round-trip (the field gets stripped on read-back), and any client using `.spec.description` instead of `.spec.summary` fails. `crd-schema-checker`'s `NoFieldRemoval` rule catches exactly this.

## Expected CI behavior

The `CRD Schema Compatibility` job (from #4987) should:

1. Resolve the baseline tag via `gh release list --limit 1` → `v0.23.1`.
2. Shallow-fetch `refs/tags/v0.23.1:refs/tags/v0.23.1`.
3. Install `crd-schema-checker` from the pinned SHA.
4. Iterate `deploy/charts/operator-crds/files/crds/*.yaml`. When it gets to `toolhive.stacklok.dev_mcpgroups.yaml`:
   - `git show v0.23.1:.../toolhive.stacklok.dev_mcpgroups.yaml` returns a version with `spec.description`.
   - The on-disk version has `spec.summary`.
   - `crd-schema-checker check-manifests` reports two `NoFieldRemoval` errors — one for `v1alpha1`, one for `v1beta1`.
5. Exit 1. `continue-on-error: true` keeps the workflow overall status passing, so the PR merge button stays unblocked.
6. Step summary shows `Status: Incompatible or Unknown` with the two errors inside the collapsible.

The other 11 CRDs should report clean.

## What to look for on the PR

[...]

PR #4992, opened 2026-04-21
TEST: validate api-break-allowed escape hatch (do not merge)

ChrisJBurns:
## ⚠️ DO NOT MERGE

Phase 2 enforcement validation. Pairs with #4988 (pre-enforcement) to close the loop: the same breaking change that previously only turned the check red should now **block the PR**, and applying `api-break-allowed` should unblock it.

**Close this PR** once the escape hatch has been demonstrated.

## What changed

Same surgical change as #4988: the JSON tag on `MCPGroupSpec.Description` is renamed from `"description"` to `"summary"`. Go field name stays, only the schema-visible key moves. `task operator-generate` + `task operator-manifests` regenerated the CRD YAML and Helm template.

## Test procedure

1. **Observe enforcement kicks in.**
   - `CRD Schema Compatibility` runs, reports 2× `NoFieldRemoval` on `mcpgroups` (`v1alpha1` and `v1beta1`).
   - Check run conclusion: `failure`.
   - Workflow conclusion: also `failure` (unlike #4988, there's no `continue-on-error` softening it).
   - `mergeStateStatus: BLOCKED` — the PR cannot be merged.

2. **Apply the escape hatch.**
   - Add the `api-break-allowed` label to the PR.
   - **Important caveat**: the workflow's `pull_request` trigger uses default types (`opened`, `synchronize`, `reopened`) — label changes don't trigger a re-run on their own. After labelling, do one of:
     - Push a new commit (e.g. `git commit --allow-empty -m trigger && git push`).
     - Close + reopen the PR.
     - Manually re-run the failed job from the Actions UI.
   - The workflow re-evaluates its `if:` guard: `!contains(labels.*.name, 'api-break-allowed')` is now false, so the `crd-schema-check` job is **skipped**.
   - GitHub treats a skipped required check as passing. `mergeStateStatus` flips to an unblocked state.

3. **Close the PR** without merging.

## Expected signals

| Stage | check conclusion | workflow conclusion | mergeStateStatus |
|---|---|---|---|
| Initial (no label) | `failure` | `failure` | `BLOCKED` |
| After label + re-trigger | `skipped` | `skipped` (or whatever the other jobs produce) | `CLEAN` or similar |

## Recommended follow-up (not this PR)

Add `types: [opened, synchronize, reopened, labeled, unlabeled]` to the `pull_request` trigger in `.github/workflows/api-compat.yml`. That would let the escape hatch work without the push/reopen/manual-rerun dance — applying the label alone would re-run the workflow and skip the job. Small, self-contained follow-up PR.

## Type of change

- [x] Other (test-only / CI validation)

## Test plan

- [x] `task operator-generate` / `task operator-manifests` regenerated cleanly
- [x] Both `v1alpha1` and `v1beta1` CRD versions now expose `spec.summary` instead of `spec.description`
- [x] Phase 1: CRD Schema Compatibility check fails, workflow fails, PR is BLOCKED
- [ ] Phase 2: after applying `api-break-allowed` and re-triggering, the job is skipped and the PR is no longer blocked by this check
- [ ] PR closed without merging

## API Compatibility

(This is a test PR. The "break" is intentional and will be reverted by closing the PR without merging. The checkbox above is for contributor-facing PRs; this PR deliberately trips the check to validate enforcement.)

**No.** This PR will never be merged.

### DIFF
--- a/cmd/thv-operator/api/v1beta1/mcpgroup_types.go
+++ b/cmd/thv-operator/api/v1beta1/mcpgroup_types.go
@@ -9,9 +9,13 @@ import (
 
 // MCPGroupSpec defines the desired state of MCPGroup
 type MCPGroupSpec struct {
-	// Description provides human-readable context
+	// Description provides human-readable context.
+	// TEST: JSON tag deliberately renamed from "description" to "summary" to
+	// validate that enforcement blocks the PR and that the api-break-allowed
+	// escape hatch unblocks it. Go field name is kept so existing callers
+	// still compile. Do NOT merge this PR.
 	// +optional
-	Description string `json:"description,omitempty"`
+	Description string `json:"summary,omitempty"`
 }
 
 // MCPGroupStatus defines observed state
--- a/deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpgroups.yaml
+++ b/deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpgroups.yaml
@@ -59,8 +59,13 @@ spec:
           spec:
             description: MCPGroupSpec defines the desired state of MCPGroup
             properties:
-              description:
-                description: Description provides human-readable context
+              summary:
+                description: |-
+                  Description provides human-readable context.
+                  TEST: JSON tag deliberately renamed from "description" to "summary" to
+                  validate that enforcement blocks the PR and that the api-break-allowed
+                  escape hatch unblocks it. Go field name is kept so existing callers
+                  still compile. Do NOT merge this PR.
                 type: string
             type: object
           status:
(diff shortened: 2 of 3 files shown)


ChrisJBurns:
Closing — test served its purpose and uncovered a real gap.

**What we validated:**
- Phase 2 enforcement works: the initial run failed the PR with `failure` / `BLOCKED`, unlike the advisory-mode counterpart #4988 which softened to `success` via `continue-on-error`.
- The `api-break-allowed` label's `if:` guard works: after an empty-commit retrigger, the job conclusion flipped to `skipped` and the api-compat check stopped blocking the PR.

**What we uncovered:**
- Applying the label alone does not re-run the workflow. GitHub's default `pull_request` trigger types (`opened`, `synchronize`, `reopened`) don't fire on label changes, and re-running the existing job from the Actions UI replays the original event payload — so the label addition is invisible to the `if:` guard in that context. Users had to push an empty commit (or close+reopen) to get the skip.

**Follow-up opened:** #4993 adds `labeled` and `unlabeled` to the trigger types, closing both the UX gap (single click to activate escape hatch) and the security gap (unlabelling re-validates, so apply-skip-unlabel-merge can't sneak past).

PR #4994, opened 2026-04-21
Allow standalone Redis in auth server storage

### LINKED ISSUE #4990: Support standalone Redis in auth server storage (no Sentinel required)
## Background

The auth server storage (`pkg/authserver/storage`) currently requires Redis Sentinel — `validateConfig` hard-requires `SentinelConfig`, `MasterName`, and `SentinelAddrs`, and `NewRedisStorage` always creates a `redis.NewFailoverClient`.

This blocks deployments on managed Redis services such as **GCP Memorystore**, **AWS ElastiCache** (non-cluster), and **Azure Cache for Redis**, all of which present a single `host:port` endpoint and manage HA internally — they do not expose Sentinel endpoints to clients.

The pattern for supporting both modes already exists in this codebase: `pkg/transport/session` (`RedisConfig` with mutually exclusive `Addr` vs `SentinelConfig` fields, `buildRedisClient` branching between `redis.NewClient` and `redis.NewFailoverClient`).

## What needs to be done

- Add an `Addr string` field to `RedisConfig` in `pkg/authserver/storage/redis.go` for standalone mode.
- Relax `validateConfig` to accept either `Addr` (standalone) or `SentinelConfig` (Sentinel) — mutually exclusive, at least one required.
- Branch in `NewRedisStorage`: use `redis.NewClient(&redis.Options{...})` for standalone, existing `redis.NewFailoverClient` for Sentinel.
- Simplify TLS handling for standalone mode: no per-address dialer is needed, a single `TLSConfig` can be passed directly on `redis.Options`.
- Update the operator's `RedisRunConfig` CRD type and converter (`cmd/thv-operator/pkg/vmcpconfig/converter.go`, `cmd/thv-operator/pkg/controllerutil/authserver.go`) to expose and propagate the standalone addr field.
- Update unit and integration tests to cover the standalone path.

## Acceptance criteria

- A ToolHive auth server can connect to a standalone Redis instance (e.g., GCP Memorystore, AWS ElastiCache) using only `Addr`.
- Existing Sentinel deployments are unaffected.
- `validateConfig` returns a clear error if neither `Addr` nor `SentinelConfig` is set, or if both are set.
- The standalone path has test coverage equivalent to the Sentinel path.

## Notes

- `pkg/transport/session` already supports both modes — align `pkg/authserver/storage` with that design.
- `pkg/ratelimit/middleware.go` already uses standalone-only (`redis.NewClient`) and has a TODO to share a client builder with session storage. This issue could be a stepping stone toward that consolidation.
- Managed Redis services handle HA internally (failover, replica promotion) without exposing Sentinel endpoints. Requiring Sentinel actively blocks these deployments.

reyortiz3:
## Summary

- The auth server storage previously required Redis Sentinel, which is not available on managed Redis services such as GCP Memorystore, AWS ElastiCache, and Azure Cache for Redis — these services expose a single endpoint and manage HA internally.
- Adds a mutually exclusive `Addr` field (standalone mode) alongside the existing `SentinelConfig` (Sentinel mode). Exactly one must be set. Mirrors the pattern already used in `pkg/transport/session/redis_config.go`.
- Propagates standalone support through all layers: `RedisConfig`/`RedisRunConfig` in storage, `convertRedisRunConfig` in the embedded auth server runner, `RedisStorageConfig` CRD type with CEL XValidation, and `buildStorageRunConfig` in the operator controller.
- Makes `usernameSecretRef` optional in the CRD and removes the empty-username validation in `validateConfig`. When username is omitted, go-redis v9 sends `HELLO` with `"default"` as the username and falls back to legacy `AUTH <password>` for servers that predate HELLO. This is required for managed Redis tiers that do not expose ACL users (GCP Memorystore Basic/Standard HA, Azure Cache for Redis).
- Updates `docs/redis-storage.md` and `docs/arch/11-auth-server-storage.md` to document both connection modes, the optional username, TLS CA cert guidance for managed services (e.g. `gcloud redis instances get-server-ca-certs`), and the known limitation that Redis Cluster mode is not supported (tracked in #5010).

Fixes #4990

## Type of change

- [x] New feature

## Test plan

New tests added in the standalone addr PR:
- `TestRedisConfig_Validation` extended with standalone/conflict/neither cases
- `TestNewRedisStorage_Standalone_ConnectionFailure` — unreachable addr returns error
- `TestNewRedisStorage_Standalone_WithMiniredis` — successful standalone connection via miniredis
- `TestConvertRedisRunConfig` extended with standalone happy path, both-set, neither-set cases
- `TestBuildStorageRunConfig` extended with standalone addr case and both-set error case

New tests added in the optional-username commit:
- `TestConvertRedisRunConfig_WithEnvVars`: "empty UsernameEnvVar uses legacy password-only auth" — verifies empty `UsernameEnvVar` resolves to empty username without error
- `TestBuildStorageRunConfig`: "Redis standalone with password-only auth omits UsernameEnvVar" — verifies operator omits `UsernameEnvVar` from RunConfig when `usernameSecretRef` is absent

## API Compatibility

This PR adds an optional `addr` field to `RedisStorageConfig`, makes `sentinelConfig` optional (it was previously implicitly required), and makes `usernameSecretRef` optional in `RedisACLUserConfig`. All changes are additive and backward compatible — existing Sentinel-based `EmbeddedAuthServerConfig` resources with a username are unchanged.

## Changes

| File | Change |
|------|--------|
| `pkg/authserver/storage/redis.go` | Add `Addr` to `RedisConfig`; update `validateConfig` (XOR, drop empty-username check); branch `NewRedisStorage` on standalone vs Sentinel |
| `pkg/authserver/storage/config.go` | Add `Addr` to `RedisRunConfig` |
| `pkg/authserver/runner/embeddedauthserver.go` | Update `convertRedisRunConfig` to handle `Addr` XOR `SentinelConfig`; make `convertRedisACLConfig` skip username resolution when `UsernameEnvVar` is empty |
| `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go` | Add `Addr` to `RedisStorageConfig`; make `SentinelConfig` optional; add CEL XValidation; make `usernameSecretRef` optional |
| `cmd/thv-operator/pkg/controllerutil/authserver.go` | Update `buildStorageRunConfig` to handle standalone path; only populate `UsernameEnvVar` when `usernameSecretRef` is present |
| `docs/redis-storage.md` | Rewritten to document standalone and Sentinel modes, optional username, TLS CA cert setup, and cluster mode limitation |
| `docs/arch/11-auth-server-storage.md` | Updated to reflect standalone support and optional username |
| `docs/server/swagger.json/yaml` | Regenerated to include `addr` field |
| CRD YAMLs, `crd-api.md` | Regenerated |

Yes. Two user-facing changes:

1. Operators can now configure auth server Redis storage with a single `addr: "host:port"` field instead of requiring a Sentinel deployment. Existing Sentinel configurations are unchanged.
2. `usernameSecretRef` is now optional. Operators targeting managed Redis tiers without ACL support (GCP Memorystore Basic/Standard HA, Azure Cache for Redis) can omit it and rely on password-only AUTH.

## Implementation plan

<details>

See `docs/superpowers/plans/2026-04-21-authserver-standalone-redis.md`.

</details>

- The CEL XValidation rule uses `self.addr.size() > 0` rather than `self.addr != ''` because the `gci` formatter in golangci-lint converts single-quoted CEL string literals to Unicode curly quotes, which corrupts the generated CRD YAML. The `size() > 0` form is semantically equivalent and linter-safe.
- `SentinelTLS` is only populated when `SentinelConfig` is non-nil, matching the standalone path that has no separate dialer TLS config.
- go-redis v9 `Hello()` substitutes `"default"` for an empty username (see `commands.go:363`), so `HELLO 2 AUTH default <password>` is sent for password-only connections. Servers that reject HELLO fall back to legacy `AUTH <password>` automatically.
- Redis Cluster mode (GCP Memorystore Cluster, ElastiCache Serverless) is out of scope and tracked in #5010.


jhrozek (commented):
Directionally I think this is right — single-endpoint connection mode is how any managed Redis wants to be talked to, and the XOR between `addr` and `sentinelConfig` (CEL + validator + converter belt-and-braces) is clean.

Before merging I'd like to pin down which managed-Redis target we're actually going after, because the answer changes whether this PR is complete. The CRD docstring on `Addr` says "use for managed Redis services (GCP Memorystore, AWS ElastiCache)", but in practice the current configuration only works on a subset of those. Memorystore Basic and Standard HA tiers don't expose Redis ACL users — those tiers only support legacy AUTH-password (https://cloud.google.com/memorystore/docs/redis/about-redis-auth). ACL users only exist on Memorystore for Redis Cluster (https://cloud.google.com/memorystore/docs/cluster/about-rbac). Azure Cache for Redis is access-key only, no ACL subsystem. ElastiCache with RBAC on Redis 6+ works fine.

The knot is in `validateConfig` at `pkg/authserver/storage/redis.go:313`, which rejects an empty `Username`. That forces go-redis to always send `AUTH <username> <password>`, which Memorystore Basic/Standard HA and Azure will reject. If we allow empty `Username`, go-redis sends the legacy `AUTH <password>` form and those tiers work.

So — is the target Memorystore Cluster / ElastiCache with RBAC (in which case the PR is already correct and we should just tighten the CRD doc so we don't imply Basic Memorystore is covered), or Basic/Standard HA / Azure (in which case I'd rather do the `Username`-optional fix in this PR — it's a few lines and saves a same-day bug report)?

Two smaller flags while we're here. `docs/arch/11-auth-server-storage.md` and `docs/redis-storage.md` still describe Redis as Sentinel-only — the latter literally says "Only Redis Sentinel deployments are supported (not standalone or cluster mode)" on line 11, which will confuse anyone reading the user guide after this lands. And Memorystore TLS uses a private Google-managed CA (https://cloud.google.com/memorystore/docs/redis/about-in-transit-encryption), not a public root — the code already supports loading a custom CA via `RedisTLSConfig.CACert`, but the user docs should walk operators through the `gcloud redis instances get-server-ca-certs` step, otherwise the failure mode is an opaque TLS verification error.

reyortiz3:
Thanks for the review @jhrozek, good catch! It makes sense to allow the empty `Username` as part of this PR, will make the changes

reyortiz3:
@jhrozek comment addressed

jhrozek:
I think you forgot to push the new code @reyortiz3 ? I only see main merges.

reyortiz3:
@jhrozek ya, sorry the changes were not pushed, should be updated now/

jhrozek (commented):
One small item worth a follow-up: the default timeouts (3s read/write, 5s dial) are likely too short for managed-service failover windows — Memorystore and ElastiCache typically take 15–30s to complete a failover, so in-flight ops will error during that window. Worth either raising the defaults or adding a note in the user guide recommending higher values for managed deployments.

jhrozek (approved):

PR #5002, opened 2026-04-22
Fix Cedar upstream-claim evaluation on VirtualMCPServer

### LINKED ISSUE #4997: Cedar policies cannot access upstream IDP claims on VirtualMCPServer with embedded auth server
## Summary

When a VirtualMCPServer uses an embedded auth server with an upstream OIDC provider (e.g., Okta), Cedar policies cannot access claims from the upstream token. The `thv run` path handles this correctly, but the Kubernetes/vMCP path does not.

## Reproduction

1. Deploy a VirtualMCPServer with:
   - An embedded auth server (`authServerConfig`) chaining to an upstream OIDC provider (e.g., Okta)
   - Inline Cedar policies referencing upstream claims (e.g., `principal.claim_department`)
2. The upstream Okta token contains `"department": "Engineering"` (verified by decoding the JWT)
3. Cedar evaluation fails:
   ```
   authorization error: [while evaluating policy `policy1`: `Client::"ba900a8b-..."` does not have the attribute `claim_department`]
   ```

## Expected behavior

When an embedded auth server is configured with an upstream provider, Cedar policies should evaluate claims from the upstream IDP token — not the embedded auth server's issued token. A policy like:

```cedar
permit(principal, action == Action::"call_tool", resource)
when { principal.claim_department == "Engineering" };
```

should succeed when the upstream Okta token contains `"department": "Engineering"`.

This already works correctly with `thv run`. The Kubernetes/vMCP path should behave the same way.

## Affected versions

Current `main`

jhrozek:
## Summary

- **Why**: When a VirtualMCPServer uses an embedded auth server with upstream OIDC providers (e.g. Okta), Cedar policies referencing upstream claims like `principal.claim_department` fail at runtime with `does not have the attribute claim_department`. The operator's converter never populated `AuthzConfig.PrimaryUpstreamProvider`, so Cedar evaluated against the ToolHive-issued AS token instead of the upstream IDP token. The same setup works on the `thv run` path, which already has the wiring.
- **What**: Three commits close the gap — one fixes the converter, one adds a safety check so misconfigurations don't silently authorize against the wrong identity, and one pins the Cedar runtime contract below the fix with a behavioral regression test.
  - Commit 1 derives `PrimaryUpstreamProvider` from `Spec.AuthServerConfig.UpstreamProviders[0].Name` in `convertIncomingAuth`, mirroring `injectSubjectProviderIfNeeded` (outgoing auth) and `injectUpstreamProviderIfNeeded` (thv run path). Leaves the field empty when no embedded AS or no upstreams are configured so Cedar correctly falls back to ToolHive-issued claims in those modes.
  - Commit 2 adds `validateAuthzUpstreamAvailable` to the reconciler chain. When `AuthzConfig` is set but no upstream IDP is configured, the VirtualMCPServer goes to `Failed` with `AuthServerConfigValidated=False, Reason=AuthzRequiresUpstream`. The user-facing message points at `spec.authServerConfig.upstreamProviders`.
  - Commit 3 adds an integration test that flips `PrimaryUpstreamProvider` between `"okta"` and `""` with the same identity and policy, asserting the expected permit/deny outcome through the middleware stack. Pins the contract the operator fix relies on.

Fixes #4997

## Type of change

- [x] Bug fix

## Test plan

- [x] Integration tests — new `TestIntegrationPrimaryUpstreamProviderClaimAttributeAccess` in `pkg/authz/integration_test.go`
- [x] Manual end-to-end verification on a kind cluster (see Verified section below)

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/pkg/vmcpconfig/converter.go` | Populate `AuthzConfig.PrimaryUpstreamProvider` from the first upstream provider when an embedded auth server is configured |
| `cmd/thv-operator/pkg/vmcpconfig/converter_test.go` | Table-driven coverage: no auth server, empty upstream list, single named upstream, empty-name normalization, multi-upstream first-wins, no authz config |
| `cmd/thv-operator/api/v1beta1/virtualmcpserver_types.go` | New condition reason `ConditionReasonAuthzRequiresUpstream` |
| `cmd/thv-operator/controllers/virtualmcpserver_controller.go` | Extract `runAuthValidations` and add `validateAuthzUpstreamAvailable` |
| `cmd/thv-operator/controllers/virtualmcpserver_controller_test.go` | Seven subtests covering the validator, including the `anonymous + authzConfig + no upstream` scenario that motivated the safety check |
| `pkg/authz/integration_test.go` | Behavioral regression test for the Cedar runtime contract: permit/deny flips on `PrimaryUpstreamProvider` with identical identity and policy |

Yes. VirtualMCPServer deployments that combine an embedded auth server, upstream OIDC providers, and Cedar `authzConfig` policies referencing upstream claims now work correctly. Previously, those policies failed with `does not have the attribute claim_*`. Separately, VirtualMCPServers that declare `authzConfig` without any upstream provider are now rejected with a `Failed` phase rather than silently authorizing against the ToolHive-issued token.

## Verified

Manual end-to-end verification against a real Okta tenant on a kind cluster.

**VirtualMCPServer shape used:**

```yaml
kind: VirtualMCPServer
spec:
  authServerConfig:
    issuer: https://<as-host>/cedar-claims-test
    upstreamProviders:
      - name: okta
        type: oidc
        oidcConfig:
          issuerUrl: https://<okta-tenant>/oauth2/<auth-server-id>
          clientId: <client-id>
          clientSecretRef:
            name: okta-client-secret
            key: client-secret
          redirectUri: https://<as-host>/cedar-claims-test/oauth/callback
          scopes: [openid, profile, email, offline_access]
  incomingAuth:
    type: oidc
    oidcConfigRef:
      name: embedded-as
      audience: https://<mcp-host>/cedar-claims-test/mcp
      resourceUrl: https://<mcp-host>/cedar-claims-test/mcp
    authzConfig:
      type: inline
      inline:
        policies:
          - |
            permit(principal, action == Action::"call_tool", resource)
            when { principal.claim_groups.contains("engineering") };
          - |
            permit(principal, action == Action::"list_tools", resource);
          - |
            permit(principal, action == Action::"initialize", resource);
```

The Cedar policy deliberately references `principal.claim_groups` — a claim that lives on the Okta access token but not on the ToolHive-issued AS token. Without the fix, Cedar would read from the AS token (no `groups`) and deny. With the fix, Cedar reads from the upstream Okta token (`groups: [Everyone, engineering]`) and permits.

**Evidence from the pod logs after completing an OAuth flow through Okta and issuing `tools/list`:**

```
Resolved JWT claim keys for Cedar evaluation
  source=upstream
  keys=[auth_time, sub, iss, aud, scp, groups, ...aws/Groups, ver, jti, iat, exp, cid, uid]
cedar authorization check
  principal=Client::"jakub@stacklok.com"
  action=Action::"call_tool"
  resource=Tool::"backend-yardstick_echo"
cedar decision
  decision=allow
  diagnostic.reasons=[policy0]
```

`source=upstream` (from `pkg/authz/authorizers/cedar/core.go:452`) proves the runtime took the upstream branch at `core.go:421` — the exact code path enabled by Commit 1. The claim key set matches Okta's access token shape (`uid`, `cid`, `aws/.../Groups`), distinct from the AS-issued token shape which would include `tsid` and `client_id`. Pre-fix this log would have read `source=token` and Cedar would have evaluated against the AS-issued token.

**PR 2 validator observed** by applying a VirtualMCPServer with `authzConfig` set but no upstream providers:

```
status:
  phase: Failed
  conditions:
    - type: AuthServerConfigValidated
      status: "False"
      reason: AuthzRequiresUpstream
      message: >-
        spec.incomingAuth.authzConfig is set but no upstream IDP is
        configured: Cedar policies referencing principal.claim_* would
        evaluate against the ToolHive-issued token instead of the upstream
        IDP token. Configure spec.authServerConfig.upstreamProviders with
        at least one upstream IDP.
```

## Implementation plan

<details>

Scope **B + D** from the planning discussion:

- **PR 1 (B1)** — Operator converter auto-sets `AuthzConfig.PrimaryUpstreamProvider` from `Spec.AuthServerConfig.UpstreamProviders[0].Name`. Mirrors the existing `injectUpstreamProviderIfNeeded` helper on the `thv run` path. Leaves the field empty when no embedded AS or no upstreams, so Cedar falls back to ToolHive claims in those modes.
- **PR 2 (B2)** — Reconciler validator rejects the degenerate case where `AuthzConfig` is set but no upstream is available, surfacing `ConfigurationValid=False, Reason=AuthzRequiresUpstream`. Protects against silent authorization-on-wrong-identity when the AS and upstream claim namespaces overlap (`sub`, `aud`, `tsid`).
- **Option X** — Integration test pinning the runtime contract below the operator fix. Lands as Commit 3 on this PR.
- **PR 4 (D, follow-up)** — Separate RFC in `toolhive-rfcs` proposing explicit `authn: true` discriminator or per-upstream claim mappers to resolve the modeling gap the oauth-expert flagged (`UpstreamProviders[]` conflates authN IdP with resource-access OAuth upstreams). Not part of this fix.

Selection for this PR is **first-upstream-wins**, consistent with all four existing call sites (`pkg/runner/middleware.go:391`, `pkg/runner/middleware.go:355`, `pkg/vmcp/config/defaults.go:110`, `cmd/thv-operator/controllers/virtualmcpserver_controller.go:2099`). The broader modeling question is tracked for PR 4.

</details>

- The converter fix picks the first upstream provider. This matches every other call site in the codebase today, but as the oauth-expert flagged in the planning discussion, it's not semantically precise for a multi-upstream deployment where the first declared upstream isn't the authN IdP. A follow-up RFC in `toolhive-rfcs` will propose an explicit discriminator; this PR is not the right place to introduce that CRD surface.
- The validator reuses the existing `AuthServerConfigValidated` condition type rather than introducing a new `AuthzConfigValidated`. Rationale: the fix the user must make is on `spec.authServerConfig.upstreamProviders`, so surfacing under that condition points at the right field. A new `ConditionReasonAuthzRequiresUpstream` disambiguates.
- `runAuthValidations` is extracted from `runValidations` because `gocyclo` tripped on the added branch. The moved block is byte-for-byte identical to what was there before; no behavior change.
- During kind-cluster verification we also hit an unrelated streamable-http bug in the vMCP transport: when the MCP client POSTs a JSON-RPC *response* to a server-initiated request (e.g., a ping reply), the server returns 400 instead of the spec-mandated 202. This is being filed as a separate issue and is not in scope for #4997.


cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -468,6 +498,45 @@ func (*VirtualMCPServerReconciler) validateAuthServerConfig(
...
+	statusManager virtualmcpserverstatus.StatusManager,
+) error {
+	if vmcp.Spec.IncomingAuth == nil || vmcp.Spec.IncomingAuth.AuthzConfig == nil {
+		return nil
+	}
+	if vmcp.Spec.AuthServerConfig != nil && len(vmcp.Spec.AuthServerConfig.UpstreamProviders) > 0 {
+		return nil
+	}
+
+	message := "spec.incomingAuth.authzConfig is set but no upstream IDP is configured: " +
```
tgrunnagle: Question: isn't configuring incoming authz with no embedded auth server (`vmcp.Spec.AuthServerConfig == nil`) a valid use case? e.g. the client acquires an okta token itself. I think we want to make sure here that there is at least one upstream if the embedded authserver is configured.

tgrunnagle (commented):
Also trying the `/dev:pr-review` skill on this, so there will be some AI-generated comments

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -468,6 +498,45 @@ func (*VirtualMCPServerReconciler) validateAuthServerConfig(
...
+		"Cedar policies referencing principal.claim_* would evaluate against the " +
+		"ToolHive-issued token instead of the upstream IDP token. Configure " +
+		"spec.authServerConfig.upstreamProviders with at least one upstream IDP."
+
+	statusManager.SetPhase(mcpv1beta1.VirtualMCPServerPhaseFailed)
+	statusManager.SetMessage(message)
+	statusManager.SetAuthServerConfigValidatedCondition(
```
tgrunnagle: **[F1, MEDIUM, 10/10] ConditionTypeAuthServerConfigValidated reused for AuthzConfig misconfiguration**

`SetAuthServerConfigValidatedCondition` is called here to report an authz policy misconfiguration that fires even when `AuthServerConfig` is `nil`. An operator who sees `AuthServerConfigValidated=False` will look at `spec.authServerConfig`, not `spec.incomingAuth.authzConfig`. Consider introducing a dedicated `ConditionTypeAuthzConfigValidated` condition type for the `AuthzRequiresUpstream` reason — the constant is already new in this PR, so the addition would be clean.

*Raised by security-reviewer, k8s-operator-reviewer*

cmd/thv-operator/pkg/vmcpconfig/converter.go
```diff
@@ -190,6 +190,19 @@ func (c *Converter) convertIncomingAuth(
...
 		// TODO: Load policies from ConfigMap if Type is "configMap"
+
+		// When an embedded auth server with upstream providers is configured, Cedar
+		// policies must evaluate claims from the upstream IDP token rather than the
+		// ToolHive-issued AS token. Mirrors injectSubjectProviderIfNeeded in
+		// virtualmcpserver_controller.go (outgoing auth) and
+		// injectUpstreamProviderIfNeeded in pkg/runner/middleware.go (thv run path).
+		// Leaving PrimaryUpstreamProvider empty (no embedded AS or no upstreams) lets
+		// Cedar fall back to claims from the ToolHive-issued token.
+		if vmcp.Spec.AuthServerConfig != nil && len(vmcp.Spec.AuthServerConfig.UpstreamProviders) > 0 {
```
tgrunnagle, line 201: **[F2, MEDIUM, 10/10] No signal emitted when first-upstream-wins selects in a multi-upstream deployment**

When `len(UpstreamProviders) > 1`, Cedar is silently wired to `UpstreamProviders[0]`. A policy author targeting a second upstream's claims gets no warning. The PR notes a follow-up RFC for an explicit discriminator — in the meantime, consider emitting an advisory `Warning` status condition when `authzConfig` is set and multiple upstreams are configured, telling users which upstream was selected.

*Raised by security-reviewer, k8s-operator-reviewer*

cmd/thv-operator/controllers/virtualmcpserver_controller_test.go
```diff
@@ -3476,3 +3476,150 @@ func TestDiscoveredRBACRulesIncludeMCPServerEntries(t *testing.T) {
...
+					GroupRef:         &mcpv1beta1.MCPGroupRef{Name: testGroupName},
+					IncomingAuth:     tt.incomingAuth,
+					AuthServerConfig: tt.authServerConfig,
+				},
+			}
+
+			r := &VirtualMCPServerReconciler{}
+			statusManager := virtualmcpserverstatus.NewStatusManager(vmcp)
+			err := r.validateAuthzUpstreamAvailable(vmcp, statusManager)
+			_ = statusManager.UpdateStatus(context.Background(), &vmcp.Status)
```
tgrunnagle: **[F9, MEDIUM, 7/10] `UpdateStatus` return value discarded silently**

`_ = statusManager.UpdateStatus(...)` discards the `bool` return. If a regression caused `hasChanges` to stop being set, the error-path assertions (`assert.Equal(t, VirtualMCPServerPhaseFailed, vmcp.Status.Phase)`) would pass trivially against zero values. For error-path subtests, assert the return is `true`:
```go
assert.True(t, statusManager.UpdateStatus(t.Context(), &vmcp.Status))
```

*Raised by test-coverage-reviewer*

tgrunnagle (commented):
**Agents consulted**: security-reviewer, k8s-operator-reviewer, test-coverage-reviewer, general-quality-reviewer

### Consensus Summary
| F1 | ConditionTypeAuthServerConfigValidated reused for AuthzConfig misconfiguration | 10/10 | MEDIUM | Fix |
| F2 | First-upstream-wins: no signal in multi-upstream deployments | 10/10 | MEDIUM | Discuss |
| F3 | `fmt.Errorf("%s", msg)` → `errors.New` | 10/10 | LOW | Fix |
| F4 | Unreachable guard branch — ordering dependency undocumented | 7/10 | LOW | Document |
| F5 | No log emitted on AuthzRequiresUpstream failure | 7/10 | LOW | Fix |
| F6 | No positive condition after successful validation (undocumented) | 7/10 | INFO | Document |
| F7 | `runValidations` doc comment stale after refactor | 10/10 | LOW | Fix |
| F9 | `UpdateStatus` return value discarded in test | 7/10 | MEDIUM | Fix |
| F10 | Integration test 403 vs 500 Cedar diagnostic ambiguity | 7/10 | LOW | Clarify |
| F11 | No test exercises both validators in sequence via `runAuthValidations` | 10/10 | LOW | Fix |
| F12 | No double-nil (authServerConfig=nil + authzConfig=nil) converter test | 7/10 | LOW | Fix |
| F13 | No end-to-end test for runAuthValidations through runValidations | 7/10 | LOW | Fix |
| F14 | `context.Background()` → `t.Context()` in new tests | 5/10 | LOW | Fix |

### Overall
The core fix is correct and well-motivated: populating `PrimaryUpstreamProvider` in `convertIncomingAuth` mirrors the existing `injectUpstreamProviderIfNeeded` pattern on the `thv run` path precisely. The converter change is minimal and the validator adds genuine security value by surfacing a previously silent misconfiguration. Testing is thorough at the unit level with comprehensive table-driven cases.

The two MEDIUM findings are not blocking but worth addressing. F1 (condition type semantic mismatch) is the most actionable — `ConditionTypeAuthServerConfigValidated=False` fires when `AuthServerConfig` is absent, which points users at the wrong field. Since `ConditionReasonAuthzRequiresUpstream` is already a new constant in this PR, introducing a corresponding `ConditionTypeAuthzConfigValidated` would be a clean addition. F2 (first-upstream-wins with no signal) is acknowledged in the special notes as accepted debt with a follow-up RFC planned. The test suite has a structural gap: all new tests call private methods directly, so if `runAuthValidations` were accidentally dropped from `runValidations`, no test would catch it (F11/F13). Inline comments are posted only for MEDIUM findings; see the table above for LOW-level items.

### Documentation
No documentation files are changed by this PR. The new `ConditionReasonAuthzRequiresUpstream` failure mode should be noted in operator troubleshooting docs and release notes — existing VirtualMCPServers with `authzConfig` set but no upstream IDP will transition to `Failed` phase on their first reconcile after the operator upgrade.

---

cmd/thv-operator/controllers/virtualmcpserver_controller.go
jhrozek: Good catch, thanks. You're right — direct-IdP deployments (client presents an already-validated OIDC token, no embedded AS) are a valid configuration, and the validator was false-positiving them. Fixed in 1b042372f: the validator now early-returns `nil` when `AuthServerConfig == nil`, and only rejects when the embedded AS is configured but has no upstream providers. The error message also now explicitly points at removing `authServerConfig` as an alternative to adding upstreams. New positive test case covers the direct-IdP branch.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
jhrozek: With the narrowing from the other thread, this condition now only fires when `AuthServerConfig` is set but has no upstreams — i.e. the misconfiguration genuinely is in the auth server spec. Operators who see `AuthServerConfigValidated=False` will look at `spec.authServerConfig.upstreamProviders`, which is where the fix belongs. I think F1 resolves itself as a side effect of the narrowing, so I left the condition type as-is. Happy to split it into a dedicated `AuthzConfigValidated` condition if you'd rather.

cmd/thv-operator/pkg/vmcpconfig/converter.go
jhrozek: Added in 1b042372f: the validator now emits an advisory `AuthzUpstreamSelectionWarning=True` with reason `AuthzUpstreamAutoSelected` when authz is configured alongside 2+ upstreams. Message names the selected upstream explicitly so operators can reorder or prune the list if the auto-selection is wrong. The condition is cleared via `RemoveConditionsWithPrefix` on all other paths so normal VMCPs don't carry a False/NotApplicable row — that came out of self-review, wanted to avoid the `SessionStorageWarning` precedent of writing a False row on every VMCP.

New test case covers the multi-upstream warning, and a separate test (`_ClearsStaleWarning`) exercises the transition from multi-upstream back to single-upstream so the stale True condition is actually removed.

cmd/thv-operator/controllers/virtualmcpserver_controller_test.go
jhrozek: Fixed. Error-path subtests now assert `assert.True(t, statusManager.UpdateStatus(...))` so a regression that stopped setting `hasChanges` would fail the assertion instead of passing trivially.

tgrunnagle (approved):

PR #5004, opened 2026-04-22
Introduce MutateAndPatchSpec and adopt across spec-patch sites

jhrozek:
## Summary

- PR #4914 migrated three MCPServer `r.Update` call sites, plus two additional annotation-stamp sites in `toolconfig_controller.go` and `mcpexternalauthconfig_controller.go`, to an inline `DeepCopy + r.Patch(MergeFromWithOptimisticLock{})` pattern. Each site carries the same four-line rationale comment, and the reviewer on PR #4914 explicitly suggested extracting a helper. This PR implements that suggestion.
- Introduces `controllerutil.MutateAndPatchSpec[T]` in `cmd/thv-operator/pkg/controllerutil/patch.go`, siblinged with the existing `MutateAndPatchStatus`. Same reflection-based nil-guard, same DeepCopy-then-mutate-then-Patch flow — but using `MergeFromWithOptimisticLock` for 409-on-stale-resourceVersion semantics, which is what defends `spec.authzConfig` (owned by a forthcoming authorization controller) from being zeroed on every reconcile. Intentionally omits the status helper's no-op short-circuit because the OL header always embeds `metadata.resourceVersion` into the body; docstring explains why.
- Adopts the helper at all five call sites. Pure refactor — no behavior change. The four-line rationale comment is removed from each site and lives in the helper's doc comment.
- Updates `.claude/rules/operator.md` "Spec / metadata patching" to reference `MutateAndPatchSpec` as the canonical pattern, replacing the inline snippet.

## Type of change

## Test plan

- [x] Unit tests (`task test`) — the new helper has five tests (happy path + OL wire signal, DeepCopy isolation, 409 Conflict propagation, nil-obj rejection, disjoint-spec-field preservation); the existing `TestMCPServerSpecPatchesAreOptimisticLock` remains green; all controller unit tests pass.
- [x] Linting (`task lint-fix`) — clean.
- [x] Manual testing (describe below) — ran the mcp-server, mcp-toolconfig, and mcp-external-auth envtest integration suites via `ginkgo -p` against `setup-envtest` 1.31.0. All green, including `mcpserver_spec_patch_integration_test.go` which writes `spec.authzConfig` out-of-band and asserts it survives both the finalizer-add reconcile and the restart-annotation reconcile.

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/pkg/controllerutil/patch.go` | New: `MutateAndPatchSpec[T]` helper. |
| `cmd/thv-operator/pkg/controllerutil/patch_test.go` | New: five unit tests mirroring the status helper's shape. |
| `cmd/thv-operator/pkg/controllerutil/doc.go` | Index entry for the new file. |
| `cmd/thv-operator/controllers/mcpserver_controller.go` | Adopt at three sites (RemoveFinalizer, AddFinalizer, restart-annotation stamp). |
| `cmd/thv-operator/controllers/toolconfig_controller.go` | Adopt at the config-hash annotation stamp in the fan-out loop. |
| `cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go` | Adopt at the config-hash annotation stamp in the fan-out loop. |
| `.claude/rules/operator.md` | Point the "Spec / metadata patching" section at the helper. |

No.

## Implementation plan

<details>

### Approach

Mirror `MutateAndPatchStatus` exactly — same generic signature, same reflection-based nil-guard, same DeepCopy-then-mutate-then-Patch flow — with two deliberate differences:

1. Use `client.MergeFromWithOptions(original, client.MergeFromWithOptimisticLock{})` instead of plain `client.MergeFrom`. Attaches the `resourceVersion` precondition so concurrent writers get 409-and-requeue instead of silent clobber.
2. **No no-op short-circuit.** The OL header always emits `metadata.resourceVersion` into the body, so a naive `{}` check never fires; and none of the five call sites execute a "maybe no-op" mutate. Docstring explains the decision.

Each call site becomes a single `MutateAndPatchSpec(ctx, r.Client, obj, func(o *T) { ... })` invocation with the existing mutation lifted verbatim into the closure; the stale four-line rationale comment is deleted from each site (the rationale now lives once in the helper's doc comment).

### Reference implementations mirrored

- `cmd/thv-operator/pkg/controllerutil/status.go` — nil-guard pattern, doc-comment shape, SPDX header.
- `cmd/thv-operator/pkg/controllerutil/status_test.go` — `statusPatchRecordingClient` is the template for the spec helper's recording client (adapted to intercept top-level `Patch` instead of `Status().Patch`); `TestMutateAndPatchStatus_RejectsNilObj` is the exact template for the nil-guard test; error text parity: `"MutateAndPatchSpec: obj must be non-nil"`.

### MoE review panel (completed before push)

Three agents reviewed in parallel:

- **go-architect** — flagged a missing disjoint-spec-field preservation test (the regression guard that would fire if the helper were ever swapped back to `r.Update`). Addressed: `TestMutateAndPatchSpec_PreservesDisjointSpecFields` in `patch_test.go`.
- **code-reviewer** — LGTM. Confirmed all five sites converted faithfully, no behavior drift, imports clean, `ctrlutil` alias consistent with existing usage.
- **kubernetes-go-expert** — confirmed OL wire-level semantics preserved, DeepCopy ordering correct, 409 propagation sound, no-op short-circuit decision justified. Raised a freshness concern on the two loop-iteration sites that on closer reading conflates two objects (our `r.Status().Update(ctx, toolConfig)` advances `toolConfig.resourceVersion`, not the server list-copy); the underlying "should fan-out sites fresh-Get before Patch" question is a pre-existing design discussion that this refactor does not change.

</details>

- Stacks cleanly on merged `main` — PR #4914 is in `74a11845b`, this PR is the single commit `2873a373f` on top.
- The helper file is named `patch.go` rather than `spec.go` (my first cut) because the sibling `status.go` is already scoped to the status subresource, and `spec.go` as a generic name in a 20-file package was weak signal. Did **not** rename `status.go` to keep this PR's scope focused — that's a trivial rename a future PR can do if we want full symmetry.


cmd/thv-operator/pkg/controllerutil/patch.go
```diff
@@ -0,0 +1,71 @@
...
+// and every current call site carries a real mutation (finalizer add/remove,
+// annotation stamp), so there is no no-op caller to optimize for.
+//
+// Do NOT use for status writes. Status-subresource writes are scoped to the
+// status stanza, and forcing a 409 on every disjoint-field overlap would
+// produce permanent churn with nothing gained — use MutateAndPatchStatus.
+//
+// Typical usage:
+//
+//	err := controllerutil.MutateAndPatchSpec(ctx, r.Client, mcpServer,
```
ChrisJBurns: **[LOW] Doc example uses `controllerutil.` instead of `ctrlutil.` alias (consensus 7/10)**

Every caller in this PR and on `main` uses the `ctrlutil` alias to avoid colliding with `sigs.k8s.io/controller-runtime/pkg/controller/controllerutil`. The `.claude/rules/operator.md` snippet already uses `ctrlutil.` — the godoc should match so developers copy-pasting don't hit a collision. Only the outer call needs to change; the inner `controllerutil.AddFinalizer` correctly references the ctrl-runtime package.

```suggestion
//	err := ctrlutil.MutateAndPatchSpec(ctx, r.Client, mcpServer,
```

*Raised by: code-reviewer*

cmd/thv-operator/pkg/controllerutil/patch.go
```diff
@@ -0,0 +1,71 @@
...
+//	    func(m *mcpv1beta1.MCPServer) {
+//	        controllerutil.AddFinalizer(m, MCPServerFinalizerName)
+//	    })
+//	if err != nil {
+//	    return ctrl.Result{}, err
+//	}
+//
+// Expect 409s as routine log noise once external writers land — the guard
+// doing its job, not a bug.
```
ChrisJBurns, line 59: **[LOW] Doc should note `obj` is mutated even when Patch errors (consensus 8/10)**

`mutate(obj)` runs before `c.Patch`; on a 409 or transport error, the caller's `obj` sits in a post-mutation, uncommitted state. All current callers return immediately and re-Get on the next reconcile — safe — but the contract isn't stated anywhere. A future caller that retries in-place without re-fetching would silently corrupt the next patch body.

Suggest adding one sentence to the doc comment, e.g. just above the "Typical usage" block:

> *If Patch returns an error, obj has already been mutated; callers must re-fetch obj before retrying rather than reusing the modified in-memory copy.*

*Raised by: kubernetes-go-expert*

cmd/thv-operator/pkg/controllerutil/patch.go
```diff
@@ -0,0 +1,71 @@
...
+	// constrained to client.Object; every real implementer is a pointer
+	// to a struct, so a nil obj is always a programmer error. Returning
+	// an explicit error is nicer than the raw panic that the subsequent
+	// .(T) type assertion would produce.
+	v := reflect.ValueOf(obj)
+	if !v.IsValid() || (v.Kind() == reflect.Pointer && v.IsNil()) {
+		return fmt.Errorf("MutateAndPatchSpec: obj must be non-nil")
+	}
+	original := obj.DeepCopyObject().(T)
+	mutate(obj)
```
ChrisJBurns, line 73: **[LOW] Nil `mutate` is unguarded, asymmetric with nil-`obj` guard (consensus 7/10)**

The helper explicitly guards nil `obj` with a reflect-based check and a descriptive error, but `mutate(obj)` on this line has no nil check — a nil `mutate` panics. All real callers pass a literal closure so this is unreachable in practice, but the asymmetry with the obj guard above is surprising.

Decide the contract: either add a symmetric guard (`if mutate == nil { return fmt.Errorf("MutateAndPatchSpec: mutate must be non-nil") }`) with a mirror test, or add one line to the docstring stating nil mutate is caller's responsibility. Matching the existing nil-obj stance is simpler and keeps the helper's error surface uniform.

*Raised by: unit-test-writer*

cmd/thv-operator/pkg/controllerutil/patch_test.go
```diff
@@ -0,0 +1,281 @@
...
+	})
+	require.NoError(t, err)
+
+	body := recorder.lastBody()
+	require.NotEmpty(t, body)
+	// If DeepCopy had aliased obj, original and current would both carry
+	// "mutated":"after" by the time MergeFrom computes the diff, and the
+	// body would lack the new annotation. Its presence proves the snapshot
+	// captured the pre-mutation state.
+	assert.Contains(t, body, "mutated",
```
ChrisJBurns, line 159: **[LOW] DeepCopy isolation test could assert pre-existing annotation is absent from patch body (consensus 7/10)**

The test seeds `Annotations["existing"] = "before"` and mutates to add `Annotations["mutated"] = "after"`. Asserting the mutation is in the body proves it lands; asserting `"existing"` is *absent* is the cleaner signal that the snapshot captured pre-mutation state (an alias bug could still produce the mutated value otherwise).

Suggest adding after the existing assertions:

```go
assert.NotContains(t, body, "existing",
    "pre-mutation fields must not leak into the merge-patch diff; body=%s", body)
```

*Raised by: kubernetes-go-expert*

cmd/thv-operator/pkg/controllerutil/patch_test.go
```diff
@@ -0,0 +1,281 @@
...
+	require.NoError(t, inner.Get(context.TODO(), client.ObjectKeyFromObject(seed), live))
+	assert.Contains(t, live.Finalizers, finalizerName,
+		"mutated field must be persisted by the patch")
+	require.NotNil(t, live.Spec.AuthzConfig,
+		"disjoint spec field owned by another controller must survive; "+
+			"this is the #4767 regression guard")
+	assert.Equal(t, mcpv1beta1.AuthzConfigTypeConfigMap, live.Spec.AuthzConfig.Type)
+	require.NotNil(t, live.Spec.AuthzConfig.ConfigMap)
+	assert.Equal(t, "external-authz-policy", live.Spec.AuthzConfig.ConfigMap.Name)
+}
```
ChrisJBurns, line 281: **[MEDIUM] No test pins the documented "no short-circuit" divergence from `MutateAndPatchStatus` (consensus 8/10)**

The helper's docstring explicitly states it does NOT short-circuit on empty diffs because `MergeFromWithOptimisticLock` always emits `metadata.resourceVersion` into the body. `MutateAndPatchStatus` has a `NoOpMutateSkipsWireCall` test that pins its short-circuit. The spec helper has no mirror-image test, so a future refactor that copy-pasted the status short-circuit into this helper would silently pass every existing test while breaking OL-on-every-reconcile semantics.

Suggest adding `TestMutateAndPatchSpec_NoOpMutateStillPatches`: seed a server, call the helper with an empty closure (`func(*mcpv1beta1.MCPServer) {}`), then assert exactly one patch was issued AND the recorded body contains `"resourceVersion"`. This is the direct wire-level pin of the documented divergence — the one test that would catch a regression to a no-op short-circuit.

*Raised by: unit-test-writer*

ChrisJBurns (commented):
**Agents consulted**: code-reviewer, kubernetes-go-expert, unit-test-writer

### Consensus Summary
| 1 | No test pins documented "no short-circuit" divergence from status helper | 8/10 | MEDIUM | Add test |
| 2 | Doc example uses `controllerutil.` prefix, not the `ctrlutil` alias used at every call site | 7/10 | LOW | Fix example |
| 3 | Doc should note `obj` is mutated even when Patch errors | 8/10 | LOW | One-sentence doc addition |
| 4 | DeepCopy isolation test could strengthen with `NotContains "existing"` | 7/10 | LOW | One assertion |
| 5 | Nil `mutate` not guarded, asymmetric with nil-`obj` guard | 7/10 | LOW | Decide the contract |

### Overall

Clean, well-scoped refactor. The helper mirrors `MutateAndPatchStatus` exactly with two deliberate divergences (`MergeFromWithOptimisticLock`, no no-op short-circuit), both justified in the docstring and the PR body. The five adoption sites are behaviorally equivalent to the pre-refactor inline code, the `ctrlutil` alias is consistent with existing usage, and the 4-line rationale comment is cleanly consolidated into the helper's godoc. Wire-level OL semantics and 409 propagation are preserved.

The findings are all polish-level — no blockers. The single MEDIUM is the absence of a no-op-mutate test that would pin the documented "no short-circuit" contract: a future refactor that copy-pasted the status helper's short-circuit into the spec helper would silently pass every existing test while breaking OL-on-every-reconcile semantics. The four LOWs are mechanical one-line fixes (doc example alias, doc sentence on post-error obj state, one test assertion, nil-mutate symmetry).

The \#4767 regression guard (`TestMutateAndPatchSpec_PreservesDisjointSpecFields`) is the strongest test in the file and directly exercises the reason the helper exists.

### Documentation

`.claude/rules/operator.md` update lands correctly — new example uses the `ctrlutil.` alias, the removed "don't mutate between DeepCopy and Patch" warning is now implicit in the helper's closure-based API, and the forward-reference to \#4633 is resolved. `doc.go` index entry added in the expected position.

---
*Generated with Claude Code*

ChrisJBurns (dismissed):
Some minor low/mediums, can address if you want or leave

cmd/thv-operator/pkg/controllerutil/patch_test.go
jhrozek: Agreed — added `TestMutateAndPatchSpec_NoOpMutateStillPatches` in bf4214014. Sharpened the assertion per the panel's suggestion: `require.Len(bodies, 1)` + `assert.NotEqual("{}", body)`, which is directly inverse to the status helper's `bytes.Equal(data, []byte("{}"))` short-circuit rather than relying on `resourceVersion` substring presence as a proxy.

cmd/thv-operator/pkg/controllerutil/patch.go
jhrozek: Agreed — fixed in bf4214014. Brought `status.go`'s godoc example along for symmetry since `operator.md` already canonicalizes the `ctrlutil.` alias.

cmd/thv-operator/pkg/controllerutil/patch.go
jhrozek: Agreed — sentence added to both `patch.go` and `status.go` in bf4214014. Added an extra clause naming the reconciler return-and-requeue pattern as the correct retry path so a future reader doesn't invent an in-place retry loop.

cmd/thv-operator/pkg/controllerutil/patch_test.go
jhrozek: Taking a pass on this one, open to revisiting if you disagree. Walking through the alias-bug case: if `DeepCopy` aliased `original` to `obj`, then at `MergeFrom(original).Data(post)` we'd have `original == post`, and an empty `{}` body — which already fails the existing `assert.Contains(body, "mutated")`. The proposed `NotContains "existing"` asserts a property of `client.MergeFrom` (unchanged fields are omitted from diffs), not of this helper — `.claude/rules/testing.md`'s "Test Scope" rule steers against dependency testing. Happy to add it anyway if you'd rather have the redundancy.

cmd/thv-operator/pkg/controllerutil/patch.go
jhrozek: Taking a pass on this too. The nil-`obj` guard exists because `obj.DeepCopyObject().(T)` on a typed-nil panics deep inside generated code — the reflect guard converts an opaque crash into a named error at the API boundary. Nil `mutate` panics at the call line with a self-diagnosing trace, so the diagnostic payoff isn't there. Go stdlib convention is not to guard nil funcs: `sync.Once.Do`, `http.HandlerFunc`, `filepath.WalkFunc`, `errors.As` callbacks all panic on nil. Adding a symmetric guard here would diverge from ecosystem convention rather than align with it. The asymmetry is intentional — the two nils are different failure modes.

jhrozek:
@ChrisJBurns addressed a subset :-)

ChrisJBurns (approved):

PR #5011, opened 2026-04-22
Add StorageVersionMigrator controller to clean up storedVersions

### LINKED ISSUE #4969: Operator: add a StorageVersionMigrator controller to clean up storedVersions
## Problem statement

PR [#4849](https://github.com/stacklok/toolhive/pull/4849) graduated our CRDs from `v1alpha1` to `v1beta1` using a **multi-version serving** strategy:

- Both `v1alpha1` and `v1beta1` are served; `v1beta1` is the storage version.
- `conversion.strategy: None` — safe because the schemas are identical.
- Users can upgrade without deleting resources and migrate their manifests at their own pace.

As users re-apply their manifests at `v1beta1`, each CRD's `status.storedVersions` ends up looking like `["v1alpha1", "v1beta1"]`. That state is correct and working, but it **blocks a future release from dropping `v1alpha1` entirely**: the API server refuses to remove a version from `spec.versions` while it's still listed in `status.storedVersions`, since doing so would orphan objects in etcd that nothing can deserialise.

Kubernetes does **not** garbage-collect `storedVersions` automatically. Someone has to re-store every object at the current storage version and then explicitly patch `storedVersions` to drop the old entry.

## Motivation

Before the release that graduates CRDs to `v1beta2` (or any future storage version), every cluster's `status.storedVersions` on the ToolHive CRDs must contain only the current storage version. Without automation, users would have to either:

1. Re-apply every CR manually (error-prone at scale), or
2. Deploy [kube-storage-version-migrator](https://github.com/kubernetes-sigs/kube-storage-version-migrator) as a separate controller and run a migration per CRD.

Neither is a good experience for users who just want to upgrade ToolHive. Since the operator already watches every ToolHive CRD, it can do this itself — and ship the migrator one release before any version removal, so the `storedVersions` cleanup happens quietly in the background during the deprecation window.

## Goal

Add a `StorageVersionMigrator` controller to the ToolHive operator that, for each opted-in `toolhive.stacklok.dev` CRD:

1. Detects when `status.storedVersions` contains versions other than the current storage version.
2. Re-stores every CR of that kind so the API server re-encodes it at the storage version.
3. Once all CRs are re-stored, patches `status.storedVersions` down to `[<currentStorageVersion>]`.

The controller is generic (works for any current-and-future storage-version transition) and scoped per-CRD via an opt-in label so future CRDs that shouldn't be migrated stay out of scope automatically.

**Initial rollout policy**: the controller ships **disabled by default** (`ENABLE_STORAGE_VERSION_MIGRATOR=false`) for one release. This gives the feature a deprecation window where it can be opted into by admins who want it, before we flip the default to on alongside the user-facing docs in a later release.

## Proposed solution

A single controller (`cmd/thv-operator/controllers/storageversionmigrator_controller.go`) that watches `CustomResourceDefinition` objects, filters to a specific scope, and performs the migration per CRD.

### Scoping: opt-in label on the CRD

The controller watches CRDs matching:

- **Group filter**: `spec.group == "toolhive.stacklok.dev"`.
- **Opt-in label**: `metadata.labels["toolhive.stacklok.dev/auto-migrate-storage-version"] == "true"`.

The label is set at CRD-generation time via a one-line kubebuilder marker on each type:

```go
// +kubebuilder:metadata:labels=toolhive.stacklok.dev/auto-migrate-storage-version=true
type MCPServer struct { ... }
```

`task operator-manifests` bakes the label into the generated CRD YAML. All 12 current `v1beta1` root types get the marker. New CRDs opt in by adding the same marker; CRDs that shouldn't be migrated omit it.

To prevent the "silent forgot-the-marker" failure mode, a CI test walks `cmd/thv-operator/api/v1beta1/` and fails if any `+kubebuilder:object:root=true` type lacks either the migrate marker or an explicit `// +thv:storage-version-migrator:exclude` sibling marker.

### Write path: plain Get+Update on the main resource

To re-encode a CR at the current storage version, the controller issues a **plain `Get` + `Update` against the main resource** (no SSA, no annotation bump, no `/status` write).

Why this works:

- When an `Update` arrives, the API server runs the request body through the **storage encoder** to produce the etcd-shaped bytes, then compares those bytes against what's currently in etcd. If they differ, the Update is committed and the resourceVersion bumps.
- For a CR stored at an older `apiVersion` (e.g. `v1alpha1`), the freshly-encoded bytes from our Update use the current storage version, which differs from the etcd bytes — so the API server commits even though our request body is byte-for-byte the same as what we just Got. That's the storage-version migration.
- For a CR already stored at the current version, the encoded bytes match what's in etcd and the Update is elided server-side. That's the noop case.

This is the same approach upstream [`kube-storage-version-migrator`](https://github.com/kubernetes-sigs/kube-storage-version-migrator) uses today. We verified the mechanics empirically in envtest via `TestReconcile_ReencodesCRsStoredAtPriorVersion` — RV bumps when the prior storage was `v1alpha1`, doesn't bump when the prior storage was already `v1beta1`.

Webhook surface: ToolHive's existing validating webhooks (MCPServer, MCPGroup, VirtualMCPServer) run on the Update path, but they only reject changes that break spec invariants. A round-trip Update with the same spec passes by definition because the spec hasn't changed.

### Concurrent-update safety

The reconciler treats `IsConflict` and `IsNotFound` separately:

- **`IsNotFound`** during the per-CR `Get`: harmless — the CR was deleted between list and re-store. Skip; nothing to migrate.
- **`IsConflict`** on the `Update`: another writer raced us. We do **not** retry inline (that would risk lost-update semantics in a loop). Instead we count conflicts and, if any occurred, return a sentinel error so the **outer reconcile does not patch `status.storedVersions` down**. The CRD stays "dirty" and the next reconcile pass retries cleanly.

`status.storedVersions` is only trimmed after every CR in the pass succeeded with zero conflicts.

### Migration cache

A short-lived in-memory cache keyed on `(CRD name, CR UID, resourceVersion)` prevents re-issuing Updates against objects already re-stored within the last hour. A `manager.RunnableFunc` GC runs every 10 minutes to drop entries for CRs that have been deleted.

### Feature flag (operator-wide off switch)

The controller is gated by an environment variable `ENABLE_STORAGE_VERSION_MIGRATOR`:

- **This release (PR-A)**: defaults to `false`. The chart does **not** surface the flag; admins who want to opt in must inject the env var via `operator.env`. Keeps casual enablement out of the standard `helm install --set` paths during the deprecation window.
- **Follow-up release (PR-C)**: a `operator.features.storageVersionMigrator` value lands in the chart and the operator-code default flips to `true`. From that point onward the feature is on by default; admins running `kube-storage-version-migrator` externally opt out via the chart value.

## How the controller decides what to migrate

The label is just the opt-in gate — it does **not** drive version decisions. Every version decision is read live from each CRD's own `spec.versions`:

| Field | Role |
|---|---|
| `spec.versions[i].storage: true` | The one version per CRD that the API server currently encodes objects as. This is the **target**. |
| `status.storedVersions` | History of every version ever used for storage. If this isn't `[target]`, there's work to do. |

Reconcile flow per CRD:

1. Label present + group matches? Proceed; else return.
2. Walk `spec.versions`, find the entry with `storage: true` → target version `V`.
3. If `status.storedVersions == [V]`: noop, return.
4. List all CRs of this kind via paginated unstructured list.
5. For each CR: `Get` (live, via `APIReader`) + `Update`. Skip on `IsNotFound`. Count `IsConflict`.
6. If step 5 had zero conflicts and zero permanent errors: patch `CRD.status.storedVersions` to `[V]` using `MergeFrom` + optimistic lock. Otherwise return the sentinel error and let controller-runtime requeue.

Worked example:

| CRD | `spec.versions` | `storage: true` | `status.storedVersions` | Controller action |
|---|---|---|---|---|
| `mcpservers.toolhive.stacklok.dev` | v1alpha1, v1beta1 | v1beta1 | `[v1alpha1, v1beta1]` | Get+Update each CR, trim to `[v1beta1]` |
| `mcpgroups.toolhive.stacklok.dev` | v1alpha1, v1beta1 | v1beta1 | `[v1beta1]` | noop |
| `newshinything.toolhive.stacklok.dev` | v1beta1, v1beta2 | v1beta2 | `[v1beta1, v1beta2]` | Get+Update each CR, trim to `[v1beta2]` |
| `experimental.toolhive.stacklok.dev` (no label) | v1alpha1 | v1alpha1 | `[v1alpha1]` | ignored — label missing |

### Scoping: label vs. hard-coded allowlist vs. per-CRD config

| Option | Why rejected |
|---|---|
| Hard-coded Go allowlist | Adding a migratable CRD requires a code change + operator release. Worse cadence than the label. |
| CAPI-style per-CRD config map (`ByObjectConfig`) | Overkill — ToolHive CRDs are structurally uniform. Easy to layer on later. |
| Label on the CRD *(chosen)* | Declarative, Git-controlled, forward-compatible. New migratable CRD = one marker line. |

### Write path: Get+Update vs. SSA on /status vs. annotation bump

The original design proposed metadata-only SSA on `/status` to dodge admission webhooks. Empirical investigation in envtest showed:

- An empty-body SSA Apply is elided by the API server **before** reaching the storage encoder — no RV bump, no storage migration.
- An annotation-bump strategy works but leaves a permanent annotation on every CR (cruft, or requires a clean-up pass later).
- A plain `Get` + `Update` against the main resource gives identical re-encode semantics (the storage encoder runs at the same layer regardless of subresource) while having zero permanent state, no field-manager surface, and matching what `kube-storage-version-migrator` upstream actually does.

Get+Update was chosen for those reasons. Webhook surface is acceptable: ToolHive's validating webhooks only reject spec-invariant violations, which a same-spec round-trip Update cannot trigger.

### Deployment model: in-operator vs. external kube-storage-version-migrator

- It's a separate deployment with its own RBAC, lifecycle, and observability. More surface for users to understand.
- It's driven by `StorageVersionMigration` CRs that someone has to create per migration.
- For a targeted, single-product scope (our ~12 CRDs), having the operator self-migrate is simpler end-to-end.

Admins who already run `kube-storage-version-migrator` can still opt out via the feature flag (once it's surfaced in the chart in PR-C).

## Precedent and prior art

**Label-scoped CRD watches** — widely used: Knative Duck typing (`duck.knative.dev/<type>=true`), OLM (`operators.coreos.com/<operator>.<namespace>`), Crossplane (`pkg.crossplane.io/package`), Gatekeeper (`admission.gatekeeper.sh/ignore`).

**Storage-version migration via Get+Update**: the upstream pattern used by [`kube-storage-version-migrator`](https://github.com/kubernetes-sigs/kube-storage-version-migrator). The CAPI [`controllers/crdmigrator`](https://github.com/kubernetes-sigs/cluster-api/tree/main/controllers/crdmigrator) does the same with its own retry/conflict handling.

**The specific combination — in-operator `storedVersions` migrator scoped by a label on the CRD — is our synthesis** for the ergonomics (one-line marker to add a new migratable CRD, admin escape hatch, no compile-time coupling).

## Known tradeoffs and caveats

1. **Silent "forgot the marker" failure** — adding a new CRD without the marker produces no build error; the migrator just skips it. Mitigated by the PR-B CI test that requires every `v1beta1` root type to carry either the migrate marker or an explicit exclude marker.
2. **Label is user-modifiable in-cluster** — admin escape hatch, but can fight with GitOps reconciliation. Documented as emergency only; long-term opt-out is the feature flag.
3. **Third-party CRDs in the group** — wildcard RBAC `toolhive.stacklok.dev/*` means any third-party CRD that adopted our group + our label would be picked up. The group is Stacklok-owned by convention; risk is theoretical.
4. **Label contract is public** — once shipped and users have written Kyverno rules or GitOps filters against it, renaming becomes a breaking change. Treat the label key as a stable API surface.
5. **Feature-flag footgun (post-PR-C)** — once the default flips to on, an admin who disables the flag and forgets to re-enable it before the `v1alpha1`-removal release will have objects orphaned in etcd. The chart-value docs will call this out.

### Interface contracts

```go
package controllers

// StorageVersionMigratorReconciler reconciles CustomResourceDefinition objects
// in the toolhive.stacklok.dev group that carry the opt-in label
// toolhive.stacklok.dev/auto-migrate-storage-version=true. For each such CRD
// it re-stores every CR at the current storage version via a plain Get+Update
// and cleans status.storedVersions so a future release can drop deprecated
// versions from spec.versions without orphaning etcd objects.
//
// Disabled by default in the initial release; admins opt in via the
// ENABLE_STORAGE_VERSION_MIGRATOR environment variable. A follow-up release
// flips the default to enabled and surfaces a chart value for opt-out.
type StorageVersionMigratorReconciler struct {
    client.Client                  // cached reads/writes for CRs
    APIReader     client.Reader    // live reads (bypasses informer)
    Scheme        *runtime.Scheme
    Recorder      record.EventRecorder

    // cache prevents re-Update of objects whose (UID, resourceVersion) was
    // already successfully migrated within the TTL window. Periodically
    // GC'd for deleted CRs by a manager.RunnableFunc.
    cache *migrationCache
}

func (r *StorageVersionMigratorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error)
func (r *StorageVersionMigratorReconciler) SetupWithManager(mgr ctrl.Manager) error

const (
    AutoMigrateLabel = "toolhive.stacklok.dev/auto-migrate-storage-version"
    AutoMigrateValue = "true"
    ToolhiveGroup    = "toolhive.stacklok.dev"
)
```

In `cmd/thv-operator/app/app.go`:

```go
const envEnableStorageVersionMigrator = "ENABLE_STORAGE_VERSION_MIGRATOR"

func isStorageVersionMigratorEnabled() bool { /* defaults to false */ }
func setupStorageVersionMigrator(mgr ctrl.Manager) error
```

RBAC markers (regenerated into `deploy/charts/operator/templates/clusterrole/role.yaml`):

```go
// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions/status,verbs=patch;update
// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=*,verbs=get;list;update
// +kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=*/status,verbs=update
```

### Tests (envtest)

| Test | Verifies |
|---|---|
| `TestReconcile_NoopWhenClean` | storedVersions already `[v1beta1]`: no writes issued |
| `TestReconcile_HappyPathTrimsStoredVersions` | storedVersions trimmed after Get+Update on every CR |
| `TestReconcile_ReencodesCRsStoredAtPriorVersion` | empirical proof RV bumps when etcd holds bytes at an older apiVersion |
| `TestReconcile_PartialFailureLeavesStoredVersionsUntouched` | permanent error leaves storedVersions intact for retry |
| `TestReconcile_ConflictDuringRestoreLeavesStoredVersionsUntouched` | IsConflict leaves storedVersions intact, retries cleanly |
| `TestReconcile_SkipsForeignCRDs` | non-toolhive CRDs are ignored |
| `TestReconcile_SkipsUnlabelledToolhiveCRD` | toolhive CRDs without the opt-in label are ignored |
| `TestReconcile_Pagination` | the continue-token list loop works |

## PR strategy

Split into three sequential PRs to keep each one independently reviewable and to maintain default-off safety throughout the rollout.

### PR-A — Controller behind default-off flag (open: [#5362](https://github.com/stacklok/toolhive/pull/5362))

- `cmd/thv-operator/controllers/storageversionmigrator_controller.go` (reconciler + cache + RBAC markers)
- `cmd/thv-operator/app/app.go` (scheme registration, setup, env-var gating, **defaults to false**)
- `cmd/thv-operator/test-integration/storageversionmigrator/` (8 envtest scenarios — CRDs created at runtime, independent of opt-in labels)
- `deploy/charts/operator/templates/clusterrole/role.yaml` (regenerated)
- **Not in PR-A**: chart `values.yaml` flag entry, `deployment.yaml` env var, opt-in labels, user docs

### PR-B — Opt-in labels + marker-coverage CI

- `+kubebuilder:metadata:labels=toolhive.stacklok.dev/auto-migrate-storage-version=true` marker on all 12 `v1beta1` root types
- Regenerated CRD YAML (`deploy/charts/operator-crds/files/crds/*.yaml`, `deploy/charts/operator-crds/templates/*.yaml`)
- `cmd/thv-operator/api/v1beta1/marker_coverage_test.go` CI test enforcing every root type carries either the migrate marker or an explicit `+thv:storage-version-migrator:exclude` marker

### PR-C — Flip default on + chart surface + docs

- Operator-code default flipped to `true`
- `operator.features.storageVersionMigrator: true` value in `values.yaml` with helm-docs comment
- `ENABLE_STORAGE_VERSION_MIGRATOR` env var wired in `deployment.yaml`
- `docs/operator/storage-version-migration.md` — user-facing reference (auto-migration behaviour, label contract, opt-out via feature flag, interaction with future `v1alpha1` removal)
- `docs/operator/upgrade-guide/` — kind-cluster walkthrough + sample CR fixtures verified end-to-end

## Acceptance criteria

Mapped across the three PRs:

| AC | PR |
|---|---|
| New `StorageVersionMigrator` controller reconciling on `CustomResourceDefinition` changes, scoped to `toolhive.stacklok.dev` + opt-in label | PR-A |
| Env-var feature flag (`ENABLE_STORAGE_VERSION_MIGRATOR`) gates registration; defaults to false | PR-A |
| 8 envtest scenarios pass | PR-A |
| RBAC for the controller's verbs ships with the controller | PR-A |
| All 12 current `v1beta1` root types carry the opt-in label via kubebuilder markers | PR-B |
| CI marker-coverage test fails when a `v1beta1` root type lacks either the migrate marker or an explicit exclude marker | PR-B |
| Operator-code default flips to `true`; chart surfaces `operator.features.storageVersionMigrator` | PR-C |
| User-facing docs explain auto-migration, the label contract, the feature-flag opt-out, and interaction with future `v1alpha1` removal | PR-C |
| End-to-end kind walkthrough exercises the full upgrade path with sample CR fixtures | PR-C |

## Risks

| Risk | Severity | Mitigation |
|---|---|---|
| Large cluster: many CRs per kind, reconcile timeout | LOW | Paginated list + per-CR Update is O(n); migration cache prevents re-work |
| Webhook added later on the main resource breaks migrator | LOW | A same-spec round-trip Update cannot trigger spec-invariant rejections |
| Label forgotten on a newly-added CRD that *should* be migrated | LOW | CI marker-coverage test (PR-B) catches this at PR time |
| Admin disables feature flag (post-PR-C) and forgets to re-enable before `v1alpha1` removal | MED | Chart-value docs explicitly warn; the flag is intended for admins running `kube-storage-version-migrator` externally |
| GitOps reconciliation fights with admin escape-hatch label removal | LOW | Docs: label is emergency only, feature flag is the supported long-term opt-out |
| Cache entries leak for deleted CRs | LOW | Periodic GC every 10 minutes via a `manager.RunnableFunc` |

## When this is needed

Before the release that drops `v1alpha1` from the ToolHive CRDs. PR-A ships the controller (default off) so admins can opt in during the deprecation window. PR-B lands the opt-in labels. PR-C flips the default on and surfaces the chart value. By the time the `v1alpha1`-removal release lands, every cluster's `storedVersions` is already clean and the version removal is a zero-risk CRD update.

## References

- PR [#4849](https://github.com/stacklok/toolhive/pull/4849) — CRD graduation from `v1alpha1` to `v1beta1` (the reason this is needed)
- PR [#5011](https://github.com/stacklok/toolhive/pull/5011) — original combined-PR reference (kept open during the split)
- PR [#5362](https://github.com/stacklok/toolhive/pull/5362) — PR-A (controller, default-off)
- Upstream reference: [`kube-storage-version-migrator`](https://github.com/kubernetes-sigs/kube-storage-version-migrator)
- CAPI reference: [`cluster-api/controllers/crdmigrator`](https://github.com/kubernetes-sigs/cluster-api/tree/main/controllers/crdmigrator)
- [Kubernetes: serve multiple versions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/)
- [Kubernetes: Migrate Kubernetes Objects Using Storage Version Migration](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/storage-version-migration/)

ChrisJBurns:
## Summary

Adds a new `StorageVersionMigrator` controller to the ToolHive operator that watches opted-in `toolhive.stacklok.dev` CRDs and reconciles each one's `status.storedVersions` down to the current storage version, so a future release can drop deprecated API versions (e.g. `v1alpha1`) from `spec.versions` without orphaning etcd objects.

Closes #4969

<details>

- **Controller** — `cmd/thv-operator/controllers/storageversionmigrator_controller.go`. Watches `apiextensions.k8s.io/v1/CustomResourceDefinition` (metadata-only), filtered by opt-in label + group. Per-CRD: reads `spec.versions` to find the storage version, lists every CR via paginated unstructured list, then for each CR does a plain `Get` + `Update` of the unmodified live object. The apiserver re-encodes the request body at the current storage version and byte-compares against etcd; when the CR was stored at a different `apiVersion` (the migration scenario) the comparison fails and the write proceeds, re-encoding the object. When the CR is already at the current storage version, the comparison matches and the apiserver harmlessly elides the write — there was nothing to migrate. This matches upstream [`kube-storage-version-migrator`'s mechanism](https://github.com/kubernetes-sigs/kube-storage-version-migrator/issues/65). Once every CR has been processed, patches `status.storedVersions` via optimistic-lock merge. 1h migration cache (with periodic background GC) dedupes Get+Update RPCs across reconciles.
- **Opt-in label** — every v1beta1 root type carries a `+kubebuilder:metadata:labels=toolhive.stacklok.dev/auto-migrate-storage-version=true` marker, baked into the generated CRD YAML by `task operator-manifests`. New CRDs opt in by adding the same marker; CRDs that shouldn't be migrated omit it. A CI test (`TestV1beta1TypesMarkerCoverage`) fails the build if a root type is added without either this marker or an explicit `+thv:storage-version-migrator:exclude` sibling.
- **Feature flag** — `operator.features.storageVersionMigrator` in the helm chart (default `true`) wires to `ENABLE_STORAGE_VERSION_MIGRATOR` following the existing `ENABLE_SERVER` / `ENABLE_REGISTRY` / `ENABLE_VMCP` pattern. Admins who prefer to run `kube-storage-version-migrator` externally can disable it.
- **Conflict-safe trim** — if any per-CR re-store hits an `IsConflict` (concurrent writer), the controller refuses to trim `storedVersions` on that pass and returns a sentinel error so the next reconcile retries cleanly. The post-conflict state of the affected object is unverified, so trimming would be reasoning from an unproven premise.
- **Tests** — envtest suite at `cmd/thv-operator/test-integration/storageversionmigrator/` with 8 scenarios: noop-when-clean, happy-path-trims-storedVersions, **re-encodes CRs that are stored at a prior storage version** (the load-bearing test that proves the migration mechanism actually rewrites etcd — installs a CRD with `v1alpha1` storage, creates a CR, flips storage to `v1beta1`, asserts the CR's `resourceVersion` bumped after reconcile), partial-failure leaves storedVersions untouched, conflict-during-restore leaves storedVersions untouched, skips-foreign-CRDs, skips-unlabelled-CRDs, pagination (continue-token loop verified via a counting APIReader wrapper). Each test installs its own CRD fixture and calls Reconcile directly — no background manager, no timing races.
- **Docs** — `docs/operator/storage-version-migration.md` covers the actual write mechanism, webhook interaction, the opt-in label contract, feature-flag opt-out, the admin per-CRD escape hatch, and coordination with the future `v1alpha1`-removal release.

</details>

<details>

| File | Change |
|------|--------|
| `cmd/thv-operator/controllers/storageversionmigrator_controller.go` | Reconciler, SetupWithManager, migration cache + periodic GC, conflict-safe trim, RBAC markers |
| `cmd/thv-operator/controllers/marker_coverage_test.go` | Scans `api/v1beta1/*_types.go` and fails if a root type lacks either the migrate marker or an explicit exclude marker |
| `cmd/thv-operator/test-integration/storageversionmigrator/suite_test.go` | envtest bootstrap (no manager; direct Reconcile calls) |
| `cmd/thv-operator/test-integration/storageversionmigrator/controller_test.go` | 8 ginkgo scenarios incl. cross-version re-encode empirical proof |
| `cmd/thv-operator/main.go` | Register `apiextensions/v1` scheme, add `featureStorageVersionMigrator` const, `setupStorageVersionMigrator()` wiring |
| `cmd/thv-operator/api/v1beta1/*_types.go` (12 files) | Add `+kubebuilder:metadata:labels=...` marker to each root type |
| `deploy/charts/operator-crds/files/crds/*.yaml` + `templates/*.yaml` (12 × 2) | Regenerated CRDs — label now present in each CRD's `metadata.labels` |
| `deploy/charts/operator/templates/clusterrole/role.yaml` | Regenerated RBAC — adds `customresourcedefinitions` get/list/watch + `/status` update;patch + `toolhive.stacklok.dev/*` get/list/update + `*/status` update |
| `deploy/charts/operator/values.yaml` | New `operator.features.storageVersionMigrator: true` |
| `deploy/charts/operator/templates/deployment.yaml` | New `ENABLE_STORAGE_VERSION_MIGRATOR` env var |
| `deploy/charts/operator/README.md` | Regenerated via `task helm-docs` |
| `docs/operator/storage-version-migration.md` | New user-facing docs |

</details>

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation only

## Test plan

- [x] `task build` — clean
- [x] Operator unit tests (`go test $(go list ./cmd/thv-operator/... | grep -v /test-integration)`) — all green
- [x] Envtest suite — 8/8 pass (`task operator-test-integration` path, or directly with `KUBEBUILDER_ASSETS=$(setup-envtest use 1.31.0 -p path) go test ./cmd/thv-operator/test-integration/storageversionmigrator/`)
- [x] Marker coverage test — passes against all 12 current v1beta1 root types
- [x] `task operator-manifests` + `task helm-docs` produce no drift against committed state
- [x] `helm template deploy/charts/operator` renders the new env var correctly

**The write mechanism in this PR has been through three iterations during review** — final state described below; the journey is captured in the inline comment threads.

The current implementation is plain `Get` + `Update` of the unmodified CR — exactly what upstream [`kube-storage-version-migrator`](https://github.com/kubernetes-sigs/kube-storage-version-migrator) does. The Kubernetes apiserver's storage layer re-encodes the request body at the current storage version, then byte-compares against etcd. When the CR was originally stored at an older `apiVersion` (the actual migration scenario), the encoded apiVersion stamp differs from etcd's, the comparison fails, and the write proceeds — re-encoding the object. When the CR is already at the current storage version, the bytes match and the apiserver elides the write harmlessly. This is the upstream maintainers' explicit reasoning, captured at [kube-storage-version-migrator#65](https://github.com/kubernetes-sigs/kube-storage-version-migrator/issues/65).

A new envtest case (`re-encodes CRs that are stored at a prior storage version`) proves the migration mechanism empirically: install a CRD with `v1alpha1` as storage, create a CR (etcd writes `apiVersion: v1alpha1` bytes), flip storage to `v1beta1`, run the reconciler, and assert the CR's `resourceVersion` bumped. Observed bump 202→204; the elision sanity check on a same-storage-version CR shows RV stays at 204 as expected.

**PR size**: ~450 production lines across 15 files, slightly over the 400-line / 10-file convention. 12 of those files are 1-line kubebuilder marker additions in `api/v1beta1/*_types.go`; the substantive Go work is 3 files (controller, main.go wiring, helm values/deployment). I did not split because the controller + opt-in labels + marker-coverage CI test are one cohesive feature — splitting would mean landing unused labels or an unused controller mid-stack.

**Why this PR must ship at least one release before `v1alpha1` is removed**: the migrator needs time to walk every cluster's `storedVersions` during the deprecation window. Shipping it in the same release as the version removal provides zero benefit because clusters jumping straight to that release won't have run the migrator yet. Documented in `docs/operator/storage-version-migration.md`.

**Follow-ups not in scope for this PR**:
- Optional chainsaw smoke test against a real Kind cluster (envtest covers the same behaviour at the apiserver level; chainsaw would add coverage for the helm chart wiring and admission-webhook compatibility specifically).
- Audit existing operator webhooks (MCPServer, MCPGroup, VirtualMCPServer) for compatibility with migration-induced Updates. Migration writes go through normal admission flow on objects that need re-encoding; webhooks should already tolerate metadata-stable Updates but worth empirical validation in a kind cluster.

### DIFF
--- a/docs/operator/storage-version-migration.md
+++ b/docs/operator/storage-version-migration.md
@@ -1,0 +1,114 @@
+# Storage Version Migration
+
+The ToolHive operator ships a `StorageVersionMigrator` controller that keeps every ToolHive CRD's `status.storedVersions` list clean, so a future operator release can drop deprecated API versions (e.g. `v1alpha1`) without orphaning objects in etcd.
+
+## Why this exists
+
+When a CRD graduates from, say, `v1alpha1` to `v1beta1` with both versions served and `v1beta1` as the storage version, existing objects continue to work — they are transparently converted on read/write. But the API server records every version that has ever been used for storage in `CustomResourceDefinition.status.storedVersions`. Until that list is trimmed, the Kubernetes API server refuses to let you remove a version from `spec.versions`, because doing so would orphan any etcd-stored objects encoded at that version.
+
+The cleanup is not automatic. Someone has to re-store every existing object at the current storage version, then explicitly patch `status.storedVersions` to drop the old entry. The `StorageVersionMigrator` controller does this for you, on every opted-in ToolHive CRD, continuously. See [upstream Kubernetes documentation](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/storage-version-migration/) for the mechanism.
+
+## What the controller does
+
+For each opted-in CRD:
+
+1. Reads `spec.versions` to find the entry with `storage: true`.
+2. If `status.storedVersions` already equals `[<currentStorageVersion>]` and only one version is served, nothing to do.
+3. Otherwise, lists every Custom Resource of that kind and issues a metadata-only Server-Side Apply against the `/status` subresource with field manager `thv-storage-version-migrator`. This forces the API server to re-encode each object at the current storage version without triggering admission webhooks (SSA on `/status` typically bypasses webhooks registered on the main resource, and the empty apply owns no fields so it doesn't fight other controllers).
+4. Once every object has been re-stored, patches `CRD.status.storedVersions` to `[<currentStorageVersion>]` using an optimistic-lock merge — so concurrent API-server writes cause a clean retry rather than a silent overwrite.
+
+CRDs without a `/status` subresource fall back to main-resource SSA.
+
+## The opt-in label
+
+A CRD participates in migration only if it carries:
+
+```yaml
+metadata:
+  labels:
+    toolhive.stacklok.dev/auto-migrate-storage-version: "true"
+```
+
+The label is set at CRD-generation time via a kubebuilder marker on each Go type in `cmd/thv-operator/api/v1beta1/`:
+
+```go
+// +kubebuilder:metadata:labels=toolhive.stacklok.dev/auto-migrate-storage-version=true
+type MCPServer struct { ... }
+```
+
+`task operator-manifests` bakes the label into the generated CRD YAML. All current ToolHive root types ship with the marker. A CI test (`TestV1beta1TypesMarkerCoverage`) fails the build if a root type is added without either this marker or an explicit `// +thv:storage-version-migrator:exclude` sibling marker — so the migrator cannot silently forget a new CRD.
+
+Adding a new CRD that should be migrated:
+
+```go
+// +kubebuilder:metadata:labels=toolhive.stacklok.dev/auto-migrate-storage-version=true
+type NewShinyThing struct { ... }
+```
+
+Adding a new CRD that deliberately should NOT be migrated (e.g. an experimental kind that is still stabilising its schema):
+
+```go
+// +thv:storage-version-migrator:exclude
+type ExperimentalThing struct { ... }
+```
+
+## Disabling the controller
+
+Set the Helm feature flag:
+
+```yaml
+operator:
+  features:
+    storageVersionMigrator: false   # default: true
+```
+
+This sets `ENABLE_STORAGE_VERSION_MIGRATOR=false` on the operator Deployment, and the reconciler is not registered with the manager.
+
+Disable only if you are running an external migrator such as [kube-storage-version-migrator](https://github.com/kubernetes-sigs/kube-storage-version-migrator). Disabling without a replacement is a footgun: the next ToolHive release that removes a deprecated API version will refuse to apply its CRD update until `storedVersions` is cleaned, and you will have to clean it yourself.
+
+## Per-CRD emergency escape hatch
+
+Removing the label on a live cluster excludes that single CRD from migration immediately:
+
+```bash
+kubectl label crd/mcpservers.toolhive.stacklok.dev \
+  toolhive.stacklok.dev/auto-migrate-storage-version-
+```
+
+Intended for incident response only. If you deploy the operator via GitOps (Argo CD, Flux) or `helm upgrade`, the chart will re-apply the chart-set label within seconds. Use the `storageVersionMigrator` feature flag for long-term opt-out.
+
+## Interaction with version removal releases
+
+The `StorageVersionMigrator` must have had time to run against your cluster *before* an operator release that drops a deprecated CRD version ships. The typical sequence is:
+
+1. **Release N**: both versions served, newer version is storage, `StorageVersionMigrator` enabled. The controller quietly re-stores all objects and trims `storedVersions` on every cluster during this deprecation window.
+2. **Release N+1+**: the deprecated version is removed from `spec.versions`. Because every cluster's `storedVersions` was already cleaned in the previous release, the CRD update applies cleanly.
+
+If your cluster upgraded directly from a pre-migrator release to the version-removal release without ever running release N, you must clean `storedVersions` manually (or deploy `kube-storage-version-migrator` once) before the upgrade can succeed.
+
+## Verification
+
+For any ToolHive CRD in a cluster where the controller has run:
+
+```bash
+kubectl get crd mcpservers.toolhive.stacklok.dev \
+  -o jsonpath='{.status.storedVersions}'
+# ["v1beta1"]
+```
+
+If the list contains more than one entry, the controller has not yet finished migrating — check operator logs for reconcile errors and the `StorageVersionMigrationFailed` event on the CRD.
+
+## RBAC
+
+The controller requires (generated from kubebuilder markers, applied by the operator Helm chart):
+
+- `customresourcedefinitions.apiextensions.k8s.io`: `get`, `list`, `watch`
+- `customresourcedefinitions/status.apiextensions.k8s.io`: `update`, `patch`
+- `*.toolhive.stacklok.dev`: `get`, `list`, `patch`
+- `*/status.toolhive.stacklok.dev`: `patch`
+
+## Related
+
+- Issue: [stacklok/toolhive#4969](https://github.com/stacklok/toolhive/issues/4969)
+- Kubernetes CRD versioning: [official docs](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/)
+- Reference implementation: [kubernetes-sigs/cluster-api `crdmigrator`](https://github.com/kubernetes-sigs/cluster-api/tree/main/controllers/crdmigrator)
--- a/docs/operator/upgrade-guide/README.md
+++ b/docs/operator/upgrade-guide/README.md
@@ -1,0 +1,323 @@
+# Upgrade walkthrough — v1alpha1 → v1beta1 with StorageVersionMigrator
+
+End-to-end manual walkthrough that replays a real user upgrade against a local kind cluster: install the previous v0.21.0 release, create a CR of each of the 12 toolhive CRD kinds at `v1alpha1`, upgrade to the new multi-version chart, deploy this branch's operator with the migrator **disabled**, re-apply the CRs at `v1beta1`, and confirm `status.storedVersions` is stuck at `[v1alpha1, v1beta1]` on every CRD. Then enable the `StorageVersionMigrator` and confirm it converges every CRD to `[v1beta1]`.
+
+Total run time: ~30 minutes. The slow part is the first `ko build` of the operator + proxyrunner + vmcp images (~3 min); subsequent runs use the build cache and finish in ~30s.
+
+This guide is the canonical reproducible verification for the migrator. Companion reading:
+
+- [`docs/operator/storage-version-migration.md`](../storage-version-migration.md) — reference docs for the controller itself (label contract, opt-out, mechanism).
+- [Issue #4969](https://github.com/stacklok/toolhive/issues/4969) — the motivating problem.
+
+## Prerequisites
+
+- `kind`, `kubectl`, `helm`, `ko`, `task` on PATH (`go install github.com/google/ko@latest` for ko)
+- Working directory: the repo root (`task operator-deploy-local` and the relative chart paths assume this).
+- Cluster name is `toolhive`. If you already have a cluster with that name, delete it first or run from a different worktree.
+
+The CR fixtures used below ship alongside this doc:
+
+- [`crs-v1alpha1.yaml`](./crs-v1alpha1.yaml) — one CR of each of the 12 kinds at `v1alpha1`
+- [`crs-v1beta1.yaml`](./crs-v1beta1.yaml) — byte-identical to the v1alpha1 file except for `apiVersion`, simulating what `sed -i 's/v1alpha1/v1beta1/g'` would produce
+
+---
+
+## 0 · Set up the cluster
+
+```bash
+# If you already have a "toolhive" kind cluster from a previous run, delete it
+kind delete cluster --name toolhive 2>/dev/null
+
+kind create cluster --name toolhive --wait 60s
+kind get kubeconfig --name toolhive > kconfig.yaml
+export KUBECONFIG=$(pwd)/kconfig.yaml
+```
+
+## 1 · Install v0.21.0 (the last v1alpha1-only release)
+
+```bash
+helm install toolhive-operator-crds \
+  oci://ghcr.io/stacklok/toolhive/toolhive-operator-crds \
+  --version 0.21.0 --wait
+
+helm install toolhive-operator \
+  oci://ghcr.io/stacklok/toolhive/toolhive-operator \
+  --version 0.21.0 \
+  --namespace toolhive-system --create-namespace --wait
+
+kubectl get crd mcpservers.toolhive.stacklok.dev -o jsonpath='{.spec.versions[*].name}'
+# expected: v1alpha1   (only one version)
+
+kubectl wait --for=condition=Available deployment -n toolhive-system --all --timeout=180s
+```
+
+## 2 · Create one CR of each of the 12 kinds at v1alpha1
+
+```bash
+kubectl create namespace upgrade-test
+kubectl apply -f docs/operator/upgrade-guide/crs-v1alpha1.yaml
+
+# Confirm all 12 landed
+kubectl get \
+  mcpservers,mcpremoteproxies,mcptoolconfigs,mcpgroups,embeddingservers,mcpregistries,virtualmcpservers,virtualmcpcompositetooldefinitions,mcpoidcconfigs,mcptelemetryconfigs,mcpexternalauthconfigs,mcpserverentries \
+  -n upgrade-test --no-headers | wc -l
+# expected: 12
+```
+
+## 3 · Let the old operator reconcile + capture the baseline
+
+```bash
+sleep 60
+kubectl get deployments -n upgrade-test
+# expected: up to 5 Deployments — test-server, test-remote-proxy, test-virtual-server,
+# test-registry-api, and (sometimes) test-embedding shows as a StatefulSet not a Deployment
+
+# Snapshot the UIDs for later comparison
+kubectl get deployments -n upgrade-test \
+  -o jsonpath='{range .items[*]}{.metadata.name}={.metadata.uid}{"\n"}{end}' \
+  | sort > /tmp/before-upgrade.txt
+cat /tmp/before-upgrade.txt
+```
+
+These UIDs are the canary. If they change after the operator upgrade, a workload was recreated → downtime.
+
+## 4 · Upgrade the CRDs chart to multi-version
+
+```bash
+helm upgrade toolhive-operator-crds deploy/charts/operator-crds --wait --timeout 120s
+
+kubectl get crd mcpservers.toolhive.stacklok.dev -o jsonpath='{.spec.versions[*].name}'
+# expected: v1alpha1 v1beta1
+
+kubectl get crd mcpservers.toolhive.stacklok.dev -o jsonpath='{.spec.versions[?(@.storage==true)].name}'
+# expected: v1beta1
+```
+
+## 5 · Build the new operator + deploy with the migrator DISABLED
+
+This is the key deviation from `task operator-deploy-local` — we want to control the feature flag, so we run ko + helm manually rather than using the task.
+
+```bash
+# ~3 minutes on first run, ~30s with build cache
+OP=$(KO_DOCKER_REPO=kind.local ko build --local -B ./cmd/thv-operator | tail -1)
+PR=$(KO_DOCKER_REPO=kind.local ko build --local -B ./cmd/thv-proxyrunner | tail -1)
+VM=$(KO_DOCKER_REPO=kind.local ko build --local -B ./cmd/vmcp | tail -1)
+
+# Load all three into the kind cluster
+kind load docker-image --name toolhive "$OP"
+kind load docker-image --name toolhive "$PR"
+kind load docker-image --name toolhive "$VM"
+
+# Persist for later steps
+echo "$OP" > /tmp/img-op
+echo "$PR" > /tmp/img-pr
+echo "$VM" > /tmp/img-vm
+
+# Helm upgrade with migrator EXPLICITLY DISABLED
+helm upgrade toolhive-operator deploy/charts/operator \
+  --set operator.image="$OP" \
+  --set operator.toolhiveRunnerImage="$PR" \
+  --set operator.vmcpImage="$VM" \
+  --set operator.features.storageVersionMigrator=false \
+  --namespace toolhive-system
+
+kubectl rollout status deployment -n toolhive-system --timeout=180s
+
+# Confirm flag took effect
+NEW_POD=$(kubectl get pods -n toolhive-system --no-headers | grep "toolhive-operator-" | awk '{print $1}' | head -1)
+kubectl get pod "$NEW_POD" -n toolhive-system -o yaml | grep -A1 ENABLE_STORAGE_VERSION_MIGRATOR
+# expected: value: "false"
+
+kubectl logs "$NEW_POD" -n toolhive-system | grep "ENABLE_STORAGE_VERSION_MIGRATOR is disabled"
+# expected: one line — "ENABLE_STORAGE_VERSION_MIGRATOR is disabled, skipping StorageVersionMigrator controller"
+```
+
+## 6 · Verify zero downtime — Deployment UIDs unchanged
+
+```bash
+kubectl get deployments -n upgrade-test \
+  -o jsonpath='{range .items[*]}{.metadata.name}={.metadata.uid}{"\n"}{end}' \
+  | sort > /tmp/after-upgrade.txt
+
+diff /tmp/before-upgrade.txt /tmp/after-upgrade.txt && echo "All Deployment UIDs unchanged"
+```
+
+## 7 · Re-apply all 12 CRs at v1beta1 (the user migration step)
+
+```bash
+kubectl apply -f docs/operator/upgrade-guide/crs-v1beta1.yaml
+# expected: 12 "configured" lines (not "created")
+
+# Wait for the operator to observe each update
+sleep 10
+```
+
+## 8 · Confirm storedVersions is stuck at `[v1alpha1, v1beta1]` on all 12 CRDs (migrator is OFF)
+
+```bash
+echo "=== storedVersions on ALL 12 CRDs (migrator OFF) ==="
+for crd in $(kubectl get crd -o name | grep toolhive.stacklok.dev | sort); do
+  name=$(echo $crd | sed 's|.*/||')
+  stored=$(kubectl get $crd -o jsonpath='{.status.storedVersions}')
+  printf "  %-55s %s\n" "$name" "$stored"
+done
+```
+
+Expected: every line ends with `["v1alpha1","v1beta1"]`. This is the "post-graduation, pre-migration" state every cluster lands in after the v0.21.0 → multi-version upgrade. **Without the migrator, this state is permanent** — any future operator release that drops `v1alpha1` from `spec.versions` would fail with:
+
+```
+status.storedVersions[0]: Invalid value: "v1alpha1": must appear in spec.versions
+```
+
+## 9 · Enable the migrator + watch it converge
+
+Helm upgrade to flip the feature flag:
+
+```bash
+OP=$(cat /tmp/img-op)
+PR=$(cat /tmp/img-pr)
+VM=$(cat /tmp/img-vm)
+
+helm upgrade toolhive-operator deploy/charts/operator \
+  --set operator.image="$OP" \
+  --set operator.toolhiveRunnerImage="$PR" \
+  --set operator.vmcpImage="$VM" \
+  --namespace toolhive-system
+
+kubectl rollout status deployment -n toolhive-system --timeout=180s
+
+# Confirm flag is now true
+NEW_POD=$(kubectl get pods -n toolhive-system --no-headers | grep "toolhive-operator-" | awk '{print $1}' | head -1)
+kubectl get pod "$NEW_POD" -n toolhive-system -o yaml | grep -A1 ENABLE_STORAGE_VERSION_MIGRATOR
+# expected: value: "true"
+```
+
+Wait for convergence:
+
+```bash
+echo "=== waiting up to 60s for all 12 CRDs to converge ==="
+for i in $(seq 1 60); do
+  count=$(for c in $(kubectl get crd -o name | grep toolhive.stacklok.dev); do
+    kubectl get $c -o jsonpath='{.status.storedVersions}'
+    echo
+  done | grep -c '^\["v1beta1"\]$')
+  if [ "$count" = "12" ]; then
+    echo "All 12 CRDs converged after ${i}s"
+    break
+  fi
+  sleep 1
+done
+```
+
+In practice this completes in ~1-2 seconds once the new pod is ready.
+
+Verify:
+
+```bash
+echo "=== storedVersions on ALL 12 CRDs (migrator ON, post-converge) ==="
+for crd in $(kubectl get crd -o name | grep toolhive.stacklok.dev | sort); do
+  name=$(echo $crd | sed 's|.*/||')
+  stored=$(kubectl get $crd -o jsonpath='{.status.storedVersions}')
+  printf "  %-55s %s\n" "$name" "$stored"
+done
+# expected: every line ends with ["v1beta1"]
+
+echo "=== StorageVersionMigrationSucceeded events ==="
+kubectl get events -A --field-selector reason=StorageVersionMigrationSucceeded --no-headers | wc -l
+# expected: 12 — one event per CRD
+
+echo "=== StorageVersionMigrationFailed events (should be 0) ==="
+kubectl get events -A --field-selector reason=StorageVersionMigrationFailed --no-headers | wc -l
+# expected: 0
+```
+
+Confirm CRs still healthy (admission webhooks on MCPServer / MCPGroup / VirtualMCPServer all accepted the per-CR Updates):
+
+```bash
+kubectl get \
+  mcpservers,mcpremoteproxies,mcptoolconfigs,mcpgroups,embeddingservers,mcpregistries,virtualmcpservers,virtualmcpcompositetooldefinitions,mcpoidcconfigs,mcptelemetryconfigs,mcpexternalauthconfigs,mcpserverentries \
+  -n upgrade-test --no-headers | wc -l
+# expected: 12
+```
+
+## 10 · (Optional) Inspect migrator logs
+
+```bash
+NEW_POD=$(kubectl get pods -n toolhive-system --no-headers | grep "toolhive-operator-" | awk '{print $1}' | head -1)
+kubectl logs "$NEW_POD" -n toolhive-system | grep "storage version migration complete" | wc -l
+# expected: 12 lines — one per CRD
+```
+
+**If this prints 0**, the migration may have happened in a previous container instance (the operator pod can restart for unrelated reasons in a kind cluster — leases, OOM, etc.). Try the previous container's logs:
+
+```bash
+kubectl logs "$NEW_POD" -n toolhive-system --previous | grep "storage version migration complete" | wc -l
+```
+
+The migration is complete in either case — the events on the CRDs in step 9 are the authoritative signal.
+
+## 11 · (Optional) Simulate the next release that drops v1alpha1
+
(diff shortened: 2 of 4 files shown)


cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,422 @@
...
+	hasStatusSubresource bool,
+) error {
+	patch := &unstructured.Unstructured{}
+	patch.SetGroupVersionKind(gvk)
+	patch.SetName(original.GetName())
+	patch.SetNamespace(original.GetNamespace())
+	patch.SetUID(original.GetUID())                         // UID mismatch → typed conflict on races
+	patch.SetResourceVersion(original.GetResourceVersion()) // RV mismatch → typed conflict on races
+
+	applyConfig := client.ApplyConfigurationFromUnstructured(patch)
```
jhrozek: The SSA body here only carries name/namespace/uid/resourceVersion, no fields inside spec or status. On `/status` SSA that's effectively an empty apply, which the apiserver typically short-circuits without rewriting etcd. If that's what's happening then we trim storedVersions while the underlying bytes haven't actually been re-encoded, which is the exact footgun this controller is meant to prevent.

For reference, kube-storage-version-migrator does an `Update` with the full re-decoded object precisely to force the storage write. Another option is to include a single managed field here (e.g. an annotation like `toolhive.stacklok.dev/last-restored-at`) so the apply is non-empty.

Worth checking: is the metadata-only SSA actually forcing a re-encode in practice? The existing envtest asserts storedVersions converges but can't distinguish "bytes rewritten" from "storedVersions lied".

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,422 @@
...
+				errs = append(errs, fmt.Errorf("unexpected list item type %T", obj))
+				return nil
+			}
+			if r.cache.has(crd.Name, u.GetUID(), u.GetResourceVersion()) {
+				return nil
+			}
+			if err := r.restoreOne(ctx, gvk, u, hasStatusSubresource); err != nil {
+				if apierrors.IsNotFound(err) || apierrors.IsConflict(err) {
+					logger.V(1).Info("skip CR — deleted or updated elsewhere",
+						"object", client.ObjectKeyFromObject(u), "err", err)
```
jhrozek, line 399: IsNotFound is fine to swallow, but I'm not sure IsConflict is. The comment says "storage is already fresh", but a concurrent writer using the deprecated API version (still served during the deprecation window) produces a write encoded at the old version. If our SSA races with that writer and comes back Conflict, the CR is still stored at the old version, we silently skip it, and then patchStoredVersions trims storedVersions to [storageVersion]. That leaves storedVersions claiming we're clean while there's still an old-version-encoded object in etcd.

Options: retry the apply on Conflict (bounded), or track a "swallowed-conflict" flag and refuse to trim storedVersions on the pass where any occurred. Either way I don't think we should treat Conflict the same as NotFound here.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,422 @@
...
+
+func newMigrationCache(ttl time.Duration) *migrationCache {
+	return &migrationCache{
+		entries: make(map[string]cacheEntry),
+		ttl:     ttl,
+		now:     time.Now,
+	}
+}
+
+func (c *migrationCache) has(crdName string, uid apitypes.UID, resourceVersion string) bool {
```
jhrozek, line 579: The cache only evicts expired entries when `has()` is called with the same key. Once a CR is deleted from the cluster its UID never comes up in a list again, so its entry stays resident until the operator restarts. For an operator that runs for weeks with CR churn (ephemeral MCPServers, GitOps recreations) the map grows unbounded, each entry is small but never released.

Is this an accepted tradeoff given the 1h TTL and typical restart cadence? If so, happy to drop this. Otherwise a background sweep via `mgr.Add(RunnableFunc(...))` or an expirable LRU would bound it.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,422 @@
...
+	if len(stored) != 1 || stored[0] != storageVersion {
+		return true
+	}
+	servedCount := 0
+	for _, v := range crd.Spec.Versions {
+		if v.Served {
+			servedCount++
+		}
+	}
+	return servedCount > 1
```
jhrozek: I think this branch produces the opposite of what the doc comment describes. When storedVersions is already `[storageVersion]` and more than one version is still served (the common long-lived state during a multi-release deprecation window), this returns true, which forces a full list pass on every CRD event. The migrationCache suppresses the per-CR SSAs but the list call itself still fires, and a restart clears the cache.

If we want to re-scan periodically during deprecation as a defensive measure, worth stating so and acknowledging the cost. If not, `return len(stored) != 1 || stored[0] != storageVersion` on its own should be sufficient, new writes at the old version would append to storedVersions and re-trigger migration via the first branch.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,422 @@
...
+)
+
+const (
+	defaultMigrationCacheTTL = 1 * time.Hour
+	defaultListPageSize      = 500
+)
+
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions/status,verbs=update;patch
+//+kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=*,verbs=get;list;patch
```
jhrozek: Minor, just wanted to surface this for sign-off. The wildcard grants get/list/patch across every current and future kind in the toolhive group, not just the opted-in ones. I get that the opted-in set isn't known at codegen time, so the alternative (explicit per-kind rules) is awkward. Probably fine, but worth a brief comment on the marker noting the intentional wildcard plus the isManagedCRD gate as defense-in-depth.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,422 @@
...
+// release can drop deprecated versions from spec.versions without orphaning
+// etcd objects.
+//
+// Enabled by default. Opt out operator-wide via
+// operator.features.storageVersionMigrator (ENABLE_STORAGE_VERSION_MIGRATOR=false)
+// for admins who prefer to run kube-storage-version-migrator externally.
+// Per-kind escape hatch: remove the label from the CRD (emergency only — will
+// be re-applied by GitOps / helm upgrade).
+type StorageVersionMigratorReconciler struct {
+	client.Client                      // cached reads for CRs (unused in this reconciler, kept for kubebuilder convention)
```
jhrozek: Nit: the comment says "unused in this reconciler, kept for kubebuilder convention", but the field is used on lines 280 (`r.Client.Status().Apply`), 285 (`r.Apply`), and 301 (`r.Client.Status().Patch`). A maintainer reading the comment might assume they can drop it. Maybe change to something like "used for SSA writes and status patch; reads go through APIReader".

jhrozek (commented):
A few things I'd like to flag. The two bigger ones are about the SSA body being empty and about swallowed conflicts before we trim storedVersions, curious whether those are intentional. The rest are smaller.

cmd/thv-operator/test-integration/storageversionmigrator/suite_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(k8sClient).NotTo(BeNil())
+})
+
+var _ = AfterSuite(func() {
+	By("tearing down envtest")
+	cancel()
+	time.Sleep(100 * time.Millisecond)
```
jhrozek: The suite doesn't start a controller manager per the docstring, so I'm not sure what this sleep is waiting for. `cancel()` doesn't block and there are no background goroutines that need to drain. If it turns out envtest.Stop() flakes without it, there's usually a proper fix (draining a specific channel) rather than a fixed delay. Otherwise it can probably just go.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Empirically confirmed — wrote a probe that runs the exact metadata-only `/status` SSA against envtest and snapshots resourceVersion before/after. Result: `rvBefore=201 rvAfter=201 changed=false`. The apiserver short-circuits the empty apply, etcd is never re-encoded, and the controller's correctness claim was hollow.

Fixed in dad30c90 by switching to `Get` + `Update` on the main resource, mutating the `toolhive.stacklok.dev/storage-version-migrated-at` annotation to force a real diff. This matches the upstream `kube-storage-version-migrator` pattern. Plain `Status().Update(ctx, live)` was tried first and is also insufficient — Update with no diff is elided too, so the annotation toggle is load-bearing.

Trade-off: the Update goes through the main resource so admission webhooks see the write. Acknowledged in the controller docstring and `docs/operator/storage-version-migration.md`. Webhooks that need to ignore migration-only writes can branch on the annotation in the diff. The original `/status` webhook-bypass benefit was always notional since the SSA wasn't writing etcd at all.

The HappyPath envtest now snapshots each CR's resourceVersion before reconcile and asserts it has bumped — empirical proof the bug is fixed.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Agreed — fixed in dad30c90. `restoreCRs` now tracks per-pass conflict count in a function-local counter. `IsConflict` is logged at V(1) and skipped per-CR (still treated as transient), but if any conflicts occurred and no other errors did, the function returns a sentinel `errMigrationRetriedDueToConflicts`. `Reconcile` propagates it without calling `patchStoredVersions`, so storedVersions stays at `[v1alpha1, v1beta1]` until a conflict-free pass succeeds.

Added a new test `ConflictDuringRestoreLeavesStoredVersionsUntouched` that injects an `apierrors.NewConflict(...)` for one CR via a wrapping client, asserts the sentinel error is returned, asserts storedVersions is untouched, then drops the injection and verifies the next reconcile completes cleanly.

`IsNotFound` is still a clean swallow with no error — the object is gone, it can't be at the old version anymore.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Right, the lazy eviction never frees entries for deleted CRs. Fixed in dad30c90 by adding a periodic background sweep: a `manager.RunnableFunc` registered in `SetupWithManager` ticks every 10m (configurable via the new `CacheGCInterval` field) and calls `cache.gc()` to evict expired entries unconditionally. Ticker is `Stop()`'ed via defer; the runnable returns on `ctx.Done()` so it shuts down cleanly with the manager.

Went with hand-rolled GC over `k8s.io/utils/lru` to keep the public API surface small. Easy to swap later if we want capacity-bounding too.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Good catch — you're right that under None-strategy with identical schemas, normal writers cannot reintroduce stale versions to storedVersions, so the defensive re-scan has no scenario to defend against. Simplified in dad30c90 to your suggested form: `return len(stored) != 1 || stored[0] != storageVersion`. Doc comment updated to be honest about the (lack of) defensive value.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Added the comment in dad30c90 — explicit note above the kubebuilder marker block explaining the wildcard is intentional (opt-in set isn't known at codegen time), with `isManagedCRD` documented as the runtime gate forming defence in depth.

While I was there: the verbs themselves changed too — `patch` → `update` for both the main resource and `/status` to match the new write path (Update + annotation toggle, see your other comment). RBAC regenerated via `task operator-manifests`.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Yep, embarrassing comment drift — the field is used at three call sites. Fixed in dad30c90:

```go
client.Client                      // used for /status writes (CR re-stores, CRD storedVersions patch); reads go through APIReader
```

(Slightly tweaked from your suggestion to mention CRD storedVersions, since the reconciler also uses `Client.Status().Patch` on the CRD itself.)

cmd/thv-operator/test-integration/storageversionmigrator/suite_test.go
ChrisJBurns: Right, dead code copy-pasted from `virtualmcp/suite_test.go` (which actually has a manager goroutine). Removed in dad30c90 along with the now-unused `time` import.

ChrisJBurns:
## Design correction — withdrawing the "upstream is broken" claim

I owe a correction on the design pivot from `dad30c90`. That commit message and the inline replies on this PR both claimed `kube-storage-version-migrator` upstream and Cluster-API's `crdmigrator` were silently broken, based on probes that showed plain `Update` of an unmodified CR was elided by the apiserver. **That conclusion was wrong** — I was testing the wrong scenario.

The corrected understanding (verified against [ksvm#65](https://github.com/kubernetes-sigs/kube-storage-version-migrator/issues/65) and confirmed empirically against our envtest in `ab03d5e1`):

- The apiserver's storage layer elision happens **after** the storage encoder runs, comparing freshly-encoded bytes against etcd bytes.
- The encoded form includes the storage version's `apiVersion` stamp.
- When etcd holds bytes stamped at an older version (the actual migration scenario) and the controller writes via the current storage version, the encoded bytes differ → write proceeds → RV bumps → etcd re-encoded.
- When etcd is already at the current storage version (no migration needed), the bytes match → write elides — and that's correct, not a bug.

My earlier probes only tested the second case (CR already at storage version), saw the elision, and overgeneralised. The probe in `ab03d5e1`'s new envtest case exercises the actual cross-version migration scenario and confirms plain `Update` bumps RV correctly.

**Practical consequence for this PR**: the annotation toggle (`toolhive.stacklok.dev/storage-version-migrated-at`) added in `dad30c90` was unnecessary. `ab03d5e1` removes it — the controller now does plain `Get` + `Update`, which matches upstream `kube-storage-version-migrator` exactly. CRs no longer carry the persistent migration annotation that earlier review surfaced as a concern.

**Apologies to the upstream maintainers** of `kube-storage-version-migrator` and `cluster-api/crdmigrator` for the mischaracterisation. Both projects' mechanisms are correct for the migration scenario they target.

ChrisJBurns:
Superseded by:

- PR #5362 — controller + envtest (merged)
- PR #5391 — opt-in labels + marker-coverage CI (merged)
- PR #5418 — chart-side feature flag (merged)
- PR #5451 — reference docs + upgrade-guide walkthrough (open)

Closing as the work has been split into the four PRs above. The doc content here was drafted against the pre-review controller design; #5451 carries the corrected, against-shipped-code version.

PR #5013, opened 2026-04-22
Forward MCPServerEntry headerForward to vMCP outbound requests

### LINKED ISSUE #4996: MCPServerEntry.spec.headerForward is accepted by CRD but never sent with requests by VirtualMCPServer
### Summary

`MCPServerEntry.spec.headerForward.addPlaintextHeaders` (and `addHeadersFromSecret`) is defined in the CRD and accepted at admission, but the values are never actually sent on outbound requests to `remoteUrl` when the entry is consumed as a static backend of a `VirtualMCPServer`. Only `MCPRemoteProxy` currently wires this field through.

### Reproducer

```yaml
kind: MCPServerEntry
metadata:
  name: github-copilot-projects
  namespace: stacklok-tunnel
spec:
  remoteUrl: https://api.githubcopilot.com/mcp/
  transport: streamable-http
  groupRef: { name: stacklok-group }
  externalAuthConfigRef: { name: github-oauth-upstream-inject }
  headerForward:
    addPlaintextHeaders:
      X-MCP-Toolsets: "projects,issues,pull_requests,users,context"
```

GitHub's remote MCP server uses `X-MCP-Toolsets` to combine multiple toolsets (the `?toolsets=` query string is ignored and the URL path only accepts one toolset). Requests from the vMCP arrive at `api.githubcopilot.com` without this header, so only the default toolset's tools surface.

### Evidence

`HeaderForward` is consumed only by `MCPRemoteProxy`:
- `cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go`
- `cmd/thv-operator/controllers/mcpremoteproxy_deployment.go`
- `cmd/thv-operator/pkg/controllerutil/externalauth.go`

For `MCPServerEntry`, the only reference is the type definition at `cmd/thv-operator/api/v1beta1/mcpserverentry_types.go`. No controller/runtime path feeds `MCPServerEntry.spec.headerForward` into the vMCP's outbound HTTP client for static backends.

### Expected

When a `VirtualMCPServer` loads a backend from an `MCPServerEntry`, the runtime HTTP client used to reach `remoteUrl` should inject the `addPlaintextHeaders` and `addHeadersFromSecret` values on every outbound request (health checks, listTools, tool invocations).

### Workaround

Use the URL path (`/mcp/x/<toolset>` or `/mcp/x/all`) since the GitHub upstream is path-addressable. Trades granularity for functionality.

### Context

Discovered wiring `github-copilot-projects` behind the vMCP for the Stacklok Anthropic tunnel — see stacklok/infra#4485 and stacklok/infra branch `fix/github-copilot-toolsets-header`.

ChrisJBurns:
## Summary

Closes #4996

`MCPServerEntry.spec.headerForward` was accepted by the CRD but silently ignored when the entry was consumed as a static backend of `VirtualMCPServer`. Only `MCPRemoteProxy` honored the field, so vMCP requests to `remoteUrl` arrived without the configured headers — breaking use cases like GitHub Copilot's `X-MCP-Toolsets` multi-toolset selection.

<details>

- **Runtime model**: `pkg/vmcp.Backend` and `BackendTarget` gain a `HeaderForward` field; `vmcpconfig.StaticBackendConfig` gains the same so headers ride alongside auth/CA-bundle through the whole pipeline.
- **Operator ingestion**: `mcpServerEntryToBackend` extracts `headerForward`. A new `buildHeaderForwardMap` mirrors `buildCABundlePathMap` and threads per-backend header config into `convertBackendsToStaticBackends`. Plaintext headers are copied verbatim; secret refs become only env-var **identifiers** — secret values never enter the ConfigMap.
- **Operator deployment**: `buildHeaderForwardEnvVarsForEntries` emits one `valueFrom.secretKeyRef` env var per (entry, header), scoped by entry name to prevent collisions across entries in the same group.
- **vMCP runtime**: a new `headerForwardRoundTripper` in `pkg/vmcp/client` sits between `identityPropagatingRoundTripper` (outer) and `authRoundTripper` (inner). Headers are resolved once at client creation via `secrets.EnvironmentProvider` (`TOOLHIVE_SECRET_<ident>`). Restricted headers (Host, hop-by-hop, X-Forwarded-*, etc.) are rejected via the existing `middleware.RestrictedHeaders` set — same list MCPRemoteProxy uses.
- **Health parity**: the health monitor's `BackendTarget` construction now carries `HeaderForward`, `CABundlePath`, and `CABundleData` so health checks hit backends with the same TLS trust and header injection as list/call traffic.
- **Reconciler validation**: `MCPServerEntry` gains a `HeaderSecretRefsValidated` condition (reusing MCPRemoteProxy's `HeaderSecretNotFound` reason) that flips the entry to `Failed` when a referenced Secret is missing.
- **Helper generalization**: `GenerateHeaderForwardSecretEnvVarName(proxyName, …)` renamed to `(ownerName, …)` so both `MCPRemoteProxy` and `MCPServerEntry` share one source of truth.

</details>

<details>

| File | Change |
|------|--------|
| `pkg/vmcp/types.go` | Add `HeaderForwardConfig` type (+gendoc, +kubebuilder:object:generate); add `HeaderForward` field to `Backend` and `BackendTarget`. |
| `pkg/vmcp/doc.go` | Add `+groupName` marker so `HeaderForwardConfig` renders in `docs/operator/crd-api.md`. |
| `pkg/vmcp/registry.go` | `BackendToTarget` now copies `HeaderForward`. |
| `pkg/vmcp/config/config.go` | `StaticBackendConfig.HeaderForward` field (`*vmcp.HeaderForwardConfig`). |
| `pkg/vmcp/aggregator/discoverer.go` | `discoverFromStaticConfig` propagates `HeaderForward`. |
| `pkg/vmcp/workloads/k8s.go` | `mcpServerEntryToBackend` populates `backend.HeaderForward`; new `buildHeaderForwardFromEntry` helper turns secret refs into identifiers via the shared helper. |
| `pkg/vmcp/health/monitor.go` | Carry `HeaderForward`, `CABundlePath`, `CABundleData` into the health-check `BackendTarget`. |
| `pkg/vmcp/client/header_forward.go` | New `headerForwardRoundTripper`; `buildHeaderForwardTripper` and `resolveHeaderForward` resolve headers once at factory time with restricted-header rejection. |
| `pkg/vmcp/client/client.go` | Add `secretsProvider` field on `httpBackendClient`; insert tripper in chain between identity (outer) and auth (inner). |
| `cmd/thv-operator/api/v1beta1/mcpserverentry_types.go` | Add `ConditionTypeMCPServerEntryHeaderSecretRefsValidated` and reason constants. |
| `cmd/thv-operator/controllers/mcpserverentry_controller.go` | New `validateHeaderForwardSecretRefs`; wired into the reconcile fan-out. |
| `cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go` | New `buildHeaderForwardMap`; threads through to config assembly. |
| `cmd/thv-operator/controllers/virtualmcpserver_controller.go` | `convertBackendsToStaticBackends` accepts `headerForwardMap` and populates `StaticBackendConfig.HeaderForward`. |
| `cmd/thv-operator/controllers/virtualmcpserver_deployment.go` | New `buildHeaderForwardEnvVarsForEntries`; called from `buildEnvVarsForVmcp`. |
| `cmd/thv-operator/pkg/controllerutil/externalauth.go` | Rename `proxyName` parameter to `ownerName`; no behavioral change. |
| `cmd/thv-operator/Taskfile.yml` | Add `pkg/vmcp` to controller-gen paths so `HeaderForwardConfig` gets a generated DeepCopy. |
| `docs/operator/crd-api.md`, `deploy/charts/operator-crds/**`, `pkg/vmcp/zz_generated.deepcopy.go`, `pkg/vmcp/config/zz_generated.deepcopy.go` | Regenerated artefacts. |

</details>

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Test plan

- [x] `task lint-fix` passes (0 issues)
- [x] `task build` passes
- [x] `task test` passes for all touched packages (48 packages across `pkg/vmcp/...` and `cmd/thv-operator/...`)
- [x] New unit tests:
  - `TestBuildHeaderForwardFromEntry` (+ secret-leak sentinel) in `pkg/vmcp/workloads/`
  - `TestBuildHeaderForwardMap` (+ per-entry identifier scoping, secret-leak sentinel) in `cmd/thv-operator/controllers/`
  - `TestBuildHeaderForwardEnvVarsForEntries` in `cmd/thv-operator/controllers/`
  - `TestHeaderForwardRoundTripper_*`, `TestResolveHeaderForward_*`, `TestBuildHeaderForwardTripper_*`, and an end-to-end `httptest.Server` test in `pkg/vmcp/client/`
  - New `MCPServerEntry` reconciler cases covering `HeaderSecretsValid` / `HeaderSecretNotFound`
- [x] `task operator-generate`, `task operator-manifests`, `task crdref-gen` run cleanly with no residual diff after commit

- **Size**: ~15 prod files, ~560 prod lines — slightly over the 400-line / 10-file soft limit but this is one logical change (close the bug end-to-end); splitting would leave PR 1 without a user-visible fix for the reproducer.
- **Secret rotation**: Matches MCPRemoteProxy parity — no Secret watch is added. Rotation requires a pod restart. A cross-CRD follow-up to add Secret watches for both resources is out of scope here.
- **RoundTripper ordering**: Chain is `trace → identity → headerForward → auth → http`. The new tripper refuses to clobber an already-set header, so auth/identity/trace always win. Restricted-header rejection uses the existing `pkg/transport/middleware.RestrictedHeaders` set to keep parity.
- **Generated code**: `zz_generated.deepcopy.go` updates and CRD YAML regeneration are bundled in the same commit. `cmd/thv-operator/Taskfile.yml` adds `pkg/vmcp` to the controller-gen paths (needed because `HeaderForwardConfig` lives at the `pkg/vmcp` root to keep `pkg/vmcp/config` from cycling back onto itself).
- **Implementation plan**: an approved design doc was produced before this PR (`design-mcpserverentry-headerforward.md`). Not committed as it's a planning artefact, but happy to share if useful.


cmd/thv-operator/controllers/virtualmcpserver_deployment.go
```diff
@@ -482,6 +492,71 @@ func (r *VirtualMCPServerReconciler) buildOutgoingAuthEnvVars(
...
+							Name: ref.ValueSecretRef.Name,
+						},
+						Key: ref.ValueSecretRef.Key,
+					},
+				},
+			})
+		}
+	}
+
+	return envVars, nil
```
jhrozek, line 557: `buildHeaderForwardEnvVarsForEntries` returns `envVars` without sorting, but the two sibling functions in this same file (`discoverExternalAuthConfigSecrets` at line 622, `discoverInlineExternalAuthConfigSecrets` at line 668) both sort by name and include a comment explaining why: `containerNeedsUpdate` uses `reflect.DeepEqual` on the env list, and unsorted env vars produce a continuous deployment update loop. PR #4783 fixed exactly this bug for the sibling functions.

With ≥2 entries declaring `headerForward.addHeadersFromSecret`, the order that `listMCPServerEntriesAsMap` + inner slice iteration produces isn't guaranteed stable across reconciles, so this function can trigger the same rollout loop.

Suggest adding the same sort before `return envVars, nil`, mirroring the existing comment:

```go
sort.Slice(envVars, func(i, j int) bool {
    return envVars[i].Name < envVars[j].Name
})
```

pkg/vmcp/client/header_forward.go
```diff
@@ -0,0 +1,157 @@
...
+
+	for name, identifier := range cfg.AddHeadersFromSecret {
+		canonical := http.CanonicalHeaderKey(name)
+		if middleware.RestrictedHeaders[canonical] {
+			return nil, fmt.Errorf(
+				"backend %q: header %q is restricted and cannot be forwarded",
+				backendName, canonical,
+			)
+		}
+		value, err := provider.GetSecret(nil, identifier) //nolint:staticcheck // provider ignores ctx
```
jhrozek, line 135: `provider.GetSecret(nil, identifier)` passes a literal `nil` `context.Context`. The current `EnvironmentProvider` ignores ctx so this works today, but the `secretsProvider` field is typed `secrets.Provider` to allow substitution (Vault, 1Password, cloud KMS, etc.). Any provider that calls `ctx.Err()` or passes ctx into an outgoing HTTP/gRPC client will nil-deref on every request.

Suggest passing `context.Background()` here (resolution happens at client-factory construction time, so the caller's ctx isn't needed) and removing the `//nolint:staticcheck` suppression.

As a future improvement, it would be cleaner to thread a `ctx context.Context` parameter through `buildHeaderForwardTripper` → `resolveHeaderForward` so any future cancellation-aware provider gets the caller's context.

cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
```diff
@@ -299,6 +300,96 @@ func (r *VirtualMCPServerReconciler) buildTransportMap(
...
+		}
+	}
+
+	return result, nil
+}
+
+// staticHeaderForwardFromEntry builds the ConfigMap-ready HeaderForwardConfig for
+// a single MCPServerEntry. Returns nil when the entry declares no headers (e.g.,
+// all AddHeadersFromSecret entries have nil ValueSecretRef).
+func staticHeaderForwardFromEntry(entry *mcpv1beta1.MCPServerEntry) *vmcptypes.HeaderForwardConfig {
```
jhrozek, line 358: `staticHeaderForwardFromEntry` here and `buildHeaderForwardFromEntry` in `pkg/vmcp/workloads/k8s.go:598` are near-duplicates — same return type (`*vmcp.HeaderForwardConfig`), same logic (plaintext copy + secret refs → identifiers via `GenerateHeaderForwardSecretEnvVarName`), minor cosmetic differences (`maps.Copy` vs manual loop, `plaintext == nil` vs `len(plaintext) == 0`).

Both paths must produce byte-identical identifiers because the operator-side function writes to the ConfigMap and the runtime-side function writes to the `vmcp.Backend` registry that the runtime reads — divergence would produce a silent "header not forwarded" mismatch between what the ConfigMap declares and what the runtime looks up.

Consider extracting a single helper in `cmd/thv-operator/pkg/controllerutil/` (next to `GenerateHeaderForwardSecretEnvVarName`, which both callers already import) so drift is impossible. Not blocking for this PR, but worth doing as a follow-up.

pkg/vmcp/doc.go
```diff
@@ -155,4 +155,6 @@
 // - MCP Specification: https://modelcontextprotocol.io/specification
 //
 // See individual subpackage documentation for detailed usage and examples.
+//
+// +groupName=toolhive.stacklok.dev
```
jhrozek, line 159: Adding `+groupName=toolhive.stacklok.dev` here (necessary for `controller-gen` to emit DeepCopy for the new `HeaderForwardConfig`) creates a new `## toolhive.stacklok.dev/vmcp` section in `docs/operator/crd-api.md` with ~30 blank lines before the single `HeaderForwardConfig` entry.

The sibling packages that set `+groupName` (`pkg/audit`, `pkg/telemetry`, `pkg/vmcp/config`, `pkg/vmcp/auth/types`) also set `+versionName=<subgroup>` and have short 4-line package comments. `pkg/vmcp/doc.go` has a 157-line package comment, which `crd-ref-docs` appears to render as group description with each blank-ish line becoming a blank markdown line.

Minimal fix: add `// +versionName=vmcp` to match the sibling convention and regenerate. If that doesn't clean up the whitespace, filtering the group in `docs/operator/crd-ref-config.yaml` is an alternative.

jhrozek (commented):
Four inline comments from a review pass. All non-blocking suggestions. See individual comments for details.

PR #5019, opened 2026-04-22
Add awsSts auth type support to vMCP

### LINKED ISSUE #5018: vMCP fails to deploy MCPServerEntry with awsSts auth type
## Summary

Deploying an `MCPServerEntry` that references an `MCPExternalAuthConfig` with `type: awsSts` fails with:

```
Failed to convert discovered auth config: failed to convert MCPExternalAuthConfig: unsupported auth type: awsSts
```

## Root Cause

`ExternalAuthTypeAWSSts` (`"awsSts"`) is a valid, documented auth type in the `MCPExternalAuthConfig` CRD, with a full `AWSStsConfig` struct and kubebuilder validation. However, the vMCP auth converter registry (`pkg/vmcp/auth/converters`) never registered a converter for this type. When `DiscoverAndResolveAuth` calls `GetConverter("awsSts")`, the registry returns an "unsupported auth type" error.

The following auth types have converters registered: `tokenExchange`, `headerInjection`, `unauthenticated`, `upstreamInject`. The `awsSts` type was missing.

## Impact

Any `MCPServerEntry` using AWS STS authentication cannot be deployed through vMCP. The error occurs at auth discovery time, preventing the backend from starting.

## Fix

Add an `AwsStsConverter` to `pkg/vmcp/auth/converters/` that maps `AWSStsConfig` fields to an internal `AwsStsConfig` type, and register it in `NewRegistry()`. `ResolveSecrets` is a no-op since AWS STS credentials are obtained at runtime via IRSA or the pod's instance profile — no Kubernetes secrets are needed.

tgrunnagle:
## Summary

`MCPServerEntry` resources using `type: awsSts` failed to deploy through vMCP with "unsupported auth type: awsSts" because the vMCP auth converter registry had no converter registered for this type, even though `AWSStsConfig` was fully specified in the CRD. This PR closes that gap by implementing the full vMCP-side AWS STS auth path: a converter, an outgoing auth strategy, and the plumbing to wire them into the existing registry and factory.

- The `AwsStsConverter` in `pkg/vmcp/auth/converters/` maps CRD `AWSStsConfig` fields to the internal `AwsStsConfig` type and is registered in `NewRegistry()`. `ResolveSecrets` is a no-op because AWS STS credentials are obtained at runtime via IRSA or the pod's instance profile.
- The `AwsStsStrategy` in `pkg/vmcp/auth/strategies/` handles outgoing request authentication: it extracts the bearer token (or the upstream provider's token when `TokenProviderName` is set), selects the IAM role via CEL-based claim matching, calls `AssumeRoleWithWebIdentity`, and signs the request with SigV4. A per-config cache (keyed by SHA-256 over all config fields) avoids redundant STS client construction across requests.
- `TokenProviderName` is a new optional field on `AWSStsConfig` that enables upstream token selection when an embedded auth server is present on a vMCP. The proxy runner and remote proxy are unaffected — their upstream swap middleware already replaces the bearer token before the STS middleware runs. The vMCP controller auto-populates `TokenProviderName` with the first configured upstream provider when left empty, matching the existing `SubjectProviderName` convention for token exchange.
- `pkg/auth/awssts.RequestSigner` is exported (previously unexported) so the vMCP strategy can sign requests outside the HTTP middleware flow.
- `ExtractSessionName` is exported from `pkg/auth/awssts` to eliminate duplication between the middleware and the strategy.
- Config validation in `pkg/vmcp/config/validator.go` and the YAML loader now explicitly handle `aws_sts`, and `pkg/vmcp/auth/types/types.go` gains the `AwsStsConfig` and `RoleMapping` types with deep-copy generated code.

Closes #5018

## Type of change

- [x] Bug fix

## Test plan

Unit tests cover:

- `AwsStsConverter`: nil config, valid conversion with and without role mappings, TokenProviderName mapping, `ResolveSecrets` no-op behavior.
- `AwsStsStrategy`: `Validate` (missing config, missing region, valid), `Authenticate` (health check skip, nil config, missing identity, upstream token selection via TokenProviderName), cache key stability and distinctness across different configs.
- `registry_test.go`: verifies `awsSts` is now a registered converter type.
- `buildAwsStsCacheKey`: configs sharing role ARNs but differing in Claim/Matcher/TokenProviderName produce distinct keys.

## API Compatibility

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go` | Add `TokenProviderName` field to `AWSStsConfig` |
| `cmd/thv-operator/controllers/virtualmcpserver_controller.go` | Extend `injectSubjectProviderIfNeeded` to auto-populate `TokenProviderName` for `aws_sts` strategies |
| `pkg/vmcp/auth/converters/aws_sts.go` | New: `AwsStsConverter` — maps CRD `AWSStsConfig` to `BackendAuthStrategy` |
| `pkg/vmcp/auth/converters/aws_sts_test.go` | New: unit tests for `AwsStsConverter` |
| `pkg/vmcp/auth/converters/interface.go` | Register `AwsStsConverter` in `NewRegistry()` |
| `pkg/vmcp/auth/converters/registry_test.go` | Verify `awsSts` converter is registered |
| `pkg/vmcp/auth/strategies/aws_sts.go` | New: `AwsStsStrategy` — STS token exchange, SigV4 signing, upstream token selection, per-config cache |
| `pkg/vmcp/auth/strategies/aws_sts_test.go` | New: unit tests for `AwsStsStrategy` and `buildAwsStsCacheKey` |
| `pkg/vmcp/auth/factory/outgoing.go` | Register `AwsStsStrategy` in `NewOutgoingAuthRegistry()` |
| `pkg/vmcp/auth/types/types.go` | Add `AwsStsConfig`, `RoleMapping` types with `TokenProviderName` field |
| `pkg/vmcp/auth/types/zz_generated.deepcopy.go` | Generated deep-copy for new types |
| `pkg/vmcp/config/validator.go` | Validate `aws_sts` strategy fields (region required) |
| `pkg/vmcp/config/yaml_loader.go` | Handle `aws_sts` in `processBackendAuthStrategy` |
| `pkg/auth/awssts/config.go` | Add `TokenProviderName` to internal `Config` |
| `pkg/auth/awssts/signer.go` | Export `RequestSigner` and add `NewRequestSigner` constructor |
| `pkg/auth/awssts/middleware.go` | Export `ExtractSessionName`, update `newRequestSigner` call sites |
| `pkg/auth/awssts/middleware_test.go` | Update test references for exported functions |
| `docs/operator/crd-api.md` | Updated CRD API docs |
| `deploy/charts/operator-crds/` | Regenerated CRD manifests |

Yes. `MCPServerEntry` resources with `type: awsSts` in their `MCPExternalAuthConfig` can now be deployed through vMCP. Previously they failed immediately at startup with an "unsupported auth type" error. A new optional `tokenProviderName` field on `awsSts` config allows selecting which upstream provider's token to use for STS exchange when an embedded auth server is configured.

The `AwsStsStrategy` maintains a per-config cache of `roleMapper` and `exchanger` instances using double-checked locking (matching `TokenExchangeStrategy`). The cache key is a SHA-256 hash over all config fields including `TokenProviderName`, role mapping Claim/Matcher/Priority, and session config to avoid collisions between backends that share role ARNs but differ in matching logic.

`injectSubjectProviderIfNeeded` in the vMCP controller was refactored from a `token_exchange`-only function to a switch-based dispatcher that also handles `aws_sts`. The shared provider resolution logic is extracted into `resolveFirstUpstreamProvider`.

## Large PR Justification

- Mostly generated, docs, and test code. Of the 572 product line changes, most are in one file (`strategies/aws_sts.go` - 356 lines) that is difficult to split.


pkg/vmcp/auth/strategies/aws_sts.go
```diff
@@ -0,0 +1,327 @@
...
+// authenticateWithCached performs the STS token exchange and SigV4 signing
+// for an outgoing request using pre-built roleMapper and exchanger instances.
+func authenticateWithCached(
+	ctx context.Context,
+	req *http.Request,
+	cfg *awssts.Config,
+	roleMapper *awssts.RoleMapper,
+	exchanger *awssts.Exchanger,
+) error {
+	identity, ok := auth.IdentityFromContext(ctx)
```
tgrunnagle, line 120: This needs to use one of the UpstreamTokens - ideally ProviderName optionally specified in the CRD with a default to the first upstream provider

pkg/vmcp/auth/strategies/aws_sts.go
```diff
@@ -0,0 +1,346 @@
...
+	sb.WriteString(cfg.Service)
+	sb.WriteByte(0)
+	sb.WriteString(cfg.FallbackRoleArn)
+	sb.WriteByte(0)
+	sb.WriteString(cfg.RoleClaim)
+	sb.WriteByte(0)
+	sb.WriteString(cfg.SessionNameClaim)
+	sb.WriteByte(0)
+	sb.WriteString(cfg.TokenProviderName)
+	sb.WriteByte(0)
```
jhrozek, line 307: `SessionDuration` is stored in the cached `awsStsContext` (line 262) but is never written into the hash. Two backends sharing all other fields but differing only in `SessionDuration` will compute the same key and share a cache entry — the second backend silently uses the wrong session duration for every STS call.

Short-term fix: add `fmt.Fprintf(&sb, "%d", cfg.GetSessionDuration())` after `TokenProviderName` here, and update the doc comment to include `SessionDuration` in the list of hashed fields.

Longer-term: worth a follow-up ticket to move this cache into the `awssts` package itself (a shared `ContextFactory`), so both the middleware and vMCP strategy share one construction path and avoid independent cache drift.

pkg/vmcp/auth/strategies/aws_sts.go
```diff
@@ -0,0 +1,346 @@
...
+// backend configurations: Region, Service, FallbackRoleArn, every RoleMapping's
+// Claim/Matcher/RoleArn/Priority (sorted by RoleArn for stability), RoleClaim,
+// and SessionNameClaim. Using a hash avoids structural ambiguity from colons
+// embedded in ARN strings and ensures configs that share role ARNs but differ in
+// matching logic (Claim, Matcher) produce distinct keys.
+func buildAwsStsCacheKey(cfg *awssts.Config) string {
+	// Sort role mappings by RoleArn for a stable ordering.
+	mappings := make([]awssts.RoleMapping, len(cfg.RoleMappings))
+	copy(mappings, cfg.RoleMappings)
+	sort.Slice(mappings, func(i, j int) bool {
```
jhrozek: `sort.Slice` is not stable. If two mappings share the same `RoleArn` but differ in `Claim` or `Matcher`, their relative order is non-deterministic across calls, producing different hash inputs for logically identical configs (spurious cache misses). `role_mapper.go` already uses `slices.SortStableFunc` — suggest matching that here.

pkg/vmcp/auth/strategies/aws_sts_test.go
```diff
@@ -0,0 +1,246 @@
...
+		},
+		{
+			name: "returns error when neither fallback role nor mappings are configured",
+			strategy: &authtypes.BackendAuthStrategy{
+				Type: authtypes.StrategyTypeAwsSts,
+				AwsSts: &authtypes.AwsStsConfig{
+					Region: "us-east-1",
+				},
+			},
+			wantErr: true,
```
jhrozek, line 144: All four test cases in `TestAwsStsStrategy_Validate` have `wantErr: true` — no success path is tested. `NewExchanger` uses `aws.AnonymousCredentials{}` at construction, so it makes no network calls and doesn't need real AWS credentials. A success case with a valid region and `FallbackRoleArn` should work in CI without any mocking.

jhrozek (commented):
Also: `pkg/vmcp/auth/factory/outgoing_test.go` (not changed in this PR, so can't be an inline comment) — `NewOutgoingAuthRegistry` now registers five strategies including `aws_sts`, but `strategyTypes` in `TestNewOutgoingAuthRegistry` still lists four. Add `authtypes.StrategyTypeAwsSts` to the slice so a regression in registration would be caught.

pkg/vmcp/auth/types/types.go
```diff
@@ -130,3 +139,55 @@ type UpstreamInjectConfig struct {
...
+	Claim string `json:"claim,omitempty" yaml:"claim,omitempty"`
+
+	// Matcher is a CEL expression for complex matching against JWT claims.
+	Matcher string `json:"matcher,omitempty" yaml:"matcher,omitempty"`
+
+	// RoleArn is the IAM role ARN to assume when this mapping matches.
+	RoleArn string `json:"roleArn" yaml:"roleArn"`
+
+	// Priority determines evaluation order (lower values = higher priority).
+	Priority *int32 `json:"priority,omitempty" yaml:"priority,omitempty"`
```
jhrozek: `authtypes.RoleMapping.Priority` is `*int32`, but `awssts.RoleMapping.Priority` is `*int`. The `toAwsStsConfig` comment says the types "mirror each other" — they don't: there's an explicit `int(*rm.Priority)` conversion at the call site. On 64-bit platforms this is safe, but the mismatch will compound if either side evolves. `awssts.RoleMapping.Priority` is used with `math.MaxInt` semantics in `effectivePriority`, so `*int` is the natural type for both.

pkg/vmcp/auth/strategies/aws_sts.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil {
+		return err
+	}
+
+	var bearerToken string
+	if cfg.TokenProviderName != "" {
+		bearerToken = identity.UpstreamTokens[cfg.TokenProviderName] // nil map safe in Go
+		if bearerToken == "" {
+			return fmt.Errorf("provider %q: %w", cfg.TokenProviderName, authtypes.ErrUpstreamTokenNotFound)
+		}
```
jhrozek, line 139: `token_exchange` uses `SubjectProviderName` for the identical concept — "which upstream provider's token to pull from `identity.UpstreamTokens`" — and `upstream_inject` uses `ProviderName`. That gives three different names for the same idea across three adjacent strategies. Was `TokenProviderName` chosen deliberately, or would `SubjectProviderName` have worked here? Aligning on one name (or an embedded shared field) would make the CRD surface more learnable and let the controller's `injectSubjectProviderIfNeeded` collapse from a switch with per-strategy arms into a single generic case.

pkg/vmcp/auth/strategies/aws_sts.go
```diff
@@ -0,0 +1,346 @@
...
+	} else {
+		var err error
+		bearerToken, err = auth.ExtractBearerToken(req)
+		if err != nil {
+			return fmt.Errorf("bearer token required for aws_sts: %w", err)
+		}
+	}
```
jhrozek, line 149: When `TokenProviderName` is empty, this falls back to `auth.ExtractBearerToken(req)` where `req` is the *outgoing* backend request being constructed. `token_exchange` handles the same empty-provider case by reading `identity.Token` instead. If the proxy strips or does not forward the `Authorization` header onto the backend request, this call fails even though the identity was already validated from the incoming request. Should this be `identity.Token` for consistency?

jhrozek: Apologies for this AI slop comment. I was testing CC in auto mode during review and it didn't have me check the comments before submitting :-/

I guess the question here  is - is aws_sts with an empty TokenProviderName a supported configuration at all ? If yes shouldn't the fallback be identity.Token and not ExtractBearerToken?

Or maybe I misread the code?

jhrozek (approved):
LGTM

Issue #4484, opened 2026-04-01
[BUG] Improve MCPServer Scaling Coordination


olamide226:
# Scale-to-Zero Leaves Orphaned Backend StatefulSet

## Description
When the `MCPServer` Proxy Deployment is scaled to 0, the backend StatefulSet continues running, resulting in a zombie state with no proxy managing or routing traffic to it.

This happens because the Proxy (proxyrunner) is currently responsible for provisioning and managing the backend StatefulSet. When the Proxy is terminated, no component remains to reconcile or clean up the backend unless the entire `MCPServer` CR is deleted.

Additionally, the current model implicitly couples Proxy and Backend scaling. In practice, this is suboptimal because the Proxy behaves like a control-plane component, while the backend StatefulSet behaves like a data-plane component. These two should not scale identically.

## Environment
- Operator Version: v0.14.1  
- CRD Version: v0.14.1  

## Current Behavior
- Scaling Proxy Deployment to 0 terminates Proxy pods
- Backend StatefulSet remains running
- Backend is only cleaned up when the MCPServer CR is deleted
- Scaling behavior implicitly ties Proxy and Backend replica counts

## Expected Behavior
- Scaling to 0 should result in both:
  - Proxy Deployment scaled to 0
  - Backend StatefulSet scaled to 0
- Scaling above 1 replica should primarily affect the backend StatefulSet, not the Proxy Deployment
- Proxy should remain singleton by default (control-plane role)

## Current Workaround
- Scale Proxy to 0 via MCPServer CR
- Manually scale StatefulSet to 0 using kubectl

Note: This is brittle because proxyrunner uses server-side apply and may overwrite manual scaling on restart.

### 1. Enable Scale-to-Zero in CRD
Update CRD validation:
- Change replicas minimum from 1 to 0

### 2. Decouple Proxy and Backend Scaling
Introduce clear scaling semantics:

- `replicas = 0`
  - Proxy = 0
  - Backend = 0

- `replicas = 1`
  - Proxy = 1
  - Backend = 1

- `replicas > 1`
  - Proxy = 1 (default)
  - Backend = N

Optional:
- Allow explicit configuration for Proxy replicas for advanced use cases

### 3. Operator-Coordinated Scaling
Enhance MCPServerReconciler:
- Handle scaling logic instead of relying solely on proxyrunner
- Ensure StatefulSet is scaled down when replicas = 0
- Prevent drift between desired state and actual backend replicas

[...]

Issue #4635, opened 2026-04-07
Add generation-change predicates to controller watches


ChrisJBurns:
## Summary

Zero predicate usage exists across all controller watch configurations. Every metadata-only change (label additions, annotation updates, resource version bumps) triggers a full reconciliation cycle. This is particularly wasteful for MCPServer reconciliation which performs expensive operations like image validation and StatefulSet comparison.

**Severity**: SHOULD FIX
**Area**: Controller Efficiency
**Breaking**: No

## Location

- All controller `SetupWithManager` methods across `cmd/thv-operator/controllers/`

## Problem

Example from a typical controller:

```go
func (r *MCPServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
    return ctrl.NewControllerManagedBy(mgr).
        For(&v1alpha1.MCPServer{}).
        Owns(&appsv1.StatefulSet{}).
        Complete(r)
}
```

No `builder.WithPredicates(...)` is applied to any watch.

## Impact

- Metadata-only changes (e.g., adding a label) trigger full reconciliation
- MCPServer reconciliation is particularly expensive due to image validation, StatefulSet comparison, and Service reconciliation
- Higher API server load from unnecessary reconciliation cycles
- Increased log noise from no-op reconciliations
- In clusters with frequent metadata updates (e.g., from GitOps tools adding annotations), this causes continuous unnecessary work

## Recommended Fix

Add generation-change predicates on primary resource watches:

```go
import "sigs.k8s.io/controller-runtime/pkg/predicate"

func (r *MCPServerReconciler) SetupWithManager(mgr ctrl.Manager) error {
    return ctrl.NewControllerManagedBy(mgr).
        For(&v1alpha1.MCPServer{},
            builder.WithPredicates(predicate.GenerationChangedPredicate{})).
        Owns(&appsv1.StatefulSet{}).
        Complete(r)
}
```

Apply to all controller primary resource watches. Do NOT apply to `Owns()` watches since owned resource status changes need to trigger reconciliation.

## Related Issues

- Remove redundant annotation-based reconcile triggers in config controllers (#4623)

Sanskarzz:
@ChrisJBurns  I'd like to work on this issue. Please assign me.

ChrisJBurns:
[...]

PR #5024, opened 2026-04-23
Gate proxyrunner StatefulSet apply by MCPServer generation

### LINKED ISSUE #5023: Stale proxy pod clobbers backing StatefulSet during rolling update after spec.image change
## Summary

During a rolling update of the proxy Deployment (triggered by a change to `MCPServer.spec.image` or any other field that rolls the pod template), two proxyrunner pods are alive simultaneously: the old-RS pod with the old args, and the new-RS pod with the new args. Each runs `DeployWorkload` → server-side apply on the backing `StatefulSet` using the same field manager (`toolhive-container-manager`) with `Force: true`. Last-writer-wins, so the **old-RS pod's apply can land after the new-RS pod's and overwrite the image back to the stale digest**. The StatefulSet is then re-rolled and its pod pulls the old image.

The symptom: `MCPServer.spec.image` has the new digest, the proxy Deployment rolled to the new image in its args, but `StatefulSet.spec.template.spec.containers[0].image` is still the old digest — and stays that way indefinitely. No operator log lines reference the MCPServer because the operator-side reconcile completed correctly; the bug is entirely in the proxy pods racing each other.

## Affected versions

Confirmed in `v0.24.0`. The same mechanism exists in all prior versions that use server-side apply from proxyrunner to manage the backing StatefulSet.

## Reproduction (observed in production)

MCPServer `stacklok-force-mcp` in namespace `stacklok-tunnel` on EKS. `spec.image` was bumped from `@sha256:fa7b95c9…` → `@sha256:772be49f…` via Flux.

Timeline (UTC):

| Time | Event |
|---|---|
| 06:09:17 | Operator scales up new RS `7c5db697bb` (new args with new digest) |
| 06:09:18 | New-RS proxy pod `…-s2jq6` scheduled |
| 06:09:19 | STS pod `stacklok-force-mcp-0` killed — first SSA from new-RS pod landed |
| 06:09:21-22 | STS pod pulls **new** digest, starts (correct state, briefly) |
| 06:09:25 | Old-RS proxy pod `f7bc47454-rtkr7` (old args) runs DeployWorkload — SSA overwrites STS |
| 06:09:34 | STS re-rolls because spec image reverted to old digest |
| 06:09:36-38 | STS pod pulls **old** digest, starts — and stays here indefinitely |
| 06:09:49 | Old-RS pod scaled down (too late) |

`StatefulSet.metadata.managedFields` confirms: the most recent `Apply` from `toolhive-container-manager` was at `06:09:25Z` — the write that regressed the image.

## Root cause

- The operator does not `Owns(&appsv1.StatefulSet{})` — only `Owns(&appsv1.Deployment{})` and `Owns(&corev1.Service{})`.
- The StatefulSet is authored by `thv-proxyrunner` inside each proxy pod (`pkg/container/kubernetes/client.go` → `DeployWorkload`).
- Both pods SSA the same `StatefulSet` as the same field manager with `Force: true`. Apply ordering is arbitrary; the stale writer can win.

## Workaround

Use the documented `restarted-at` annotation with `restart-strategy: immediate` (`docs/operator/restart-annotation.md`) to delete pods directly rather than rolling. This avoids the overlap window and so avoids the race.

```
kubectl -n <ns> annotate mcpserver <name> \
  mcpserver.toolhive.stacklok.dev/restarted-at="$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
  mcpserver.toolhive.stacklok.dev/restart-strategy=immediate --overwrite
```

## Proposed fix (Phase 1 — narrow, targeted)

Add a monotonic version to the RunConfig and gate the SSA on it:

1. Operator stamps `MCPServer.metadata.generation` into the RunConfig when rendering.
2. Proxyrunner threads it into `DeployWorkloadOptions`.
3. Before `Apply`, proxyrunner `Get`s the existing StatefulSet and reads a new annotation `toolhive.stacklok.dev/mcpserver-generation` on the pod template. If the stamped generation is strictly greater than ours, the apply is skipped.
4. Otherwise, our generation is stamped and apply proceeds.

Zero generation (old operator → new proxyrunner, or non-operator callers) bypasses both the gate and the stamp for backward compatibility.

This closes the image/pod-spec dimension of the race but does not cover scaling (`spec.replicas` dimension). That is intentional — see the related issues below.

## Proposed follow-up (Phase 2 — architectural)

Move StatefulSet ownership into the operator. The operator becomes the sole SSA writer; proxyrunner stops authoring the StatefulSet and instead waits on the Service endpoint. This retires the class of race entirely and is aligned with #4484's acceptance criteria.

## Related issues

- **#4484** — Same root cause family (proxyrunner SSA clobbering operator-adjacent state), different dimension (scale-to-zero and manual scaling). Phase 1 does not close #4484. Phase 2 would.
- **#4635** — Proposes `GenerationChangedPredicate` on controller watches. Tangential: it uses the same `.metadata.generation` signal but to skip reconciles, not to version writes. The Phase 1 pivot to deterministic Generation-based RunConfig rendering also reduces the no-op-reconcile churn that #4635 targets. Heads-up: #4635's example pseudocode shows `Owns(&appsv1.StatefulSet{})` which is not the current state of the operator.
- **#4877** (closed) — "Proxy runner triggers StatefulSet rolling update on every restart, causing crash loop." Same architectural smell (proxyrunner re-applying the STS on every startup); Phase 2 closes that pattern for good.

JAORMX:
## Summary

During a rolling update of the proxy Deployment, two proxyrunner pods are alive at the same time: the old-RS pod with the old args, and the new-RS pod with the new args. Both call `DeployWorkload` → server-side apply on the backing `StatefulSet` as the same field manager with `Force: true`. The stale writer can land after the fresh writer and clobber the image back to the old digest — the StatefulSet then re-rolls its pod onto the old image and stays there indefinitely. Confirmed in production on `v0.24.0`; details and timeline in #5023.

- Operator stamps `MCPServer.metadata.generation` into the rendered RunConfig as a monotonic version.
- Proxyrunner threads it through to `DeployWorkloadOptions` and into the K8s runtime.
- Before `Apply`, the runtime reads the existing StatefulSet's `toolhive.stacklok.dev/mcpserver-generation` pod-template annotation; if it is strictly greater than ours, the apply is skipped. Otherwise our generation is stamped on the pod template and apply proceeds as today.
- Zero generation is the backward-compat signal: bypasses both the gate and the stamp, preserving existing behavior for callers that don't pass one (old operator → new proxyrunner, Docker runtime, tests).

Fixes #5023

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

New table-driven test `TestDeployWorkload_RunConfigMCPServerGenerationGate` covers the seven gate cases: absent STS, existing STS with (missing / older / equal / strictly newer / unparseable) annotation, zero options generation (backward-compat). Round-trip JSON test for `MCPServerGeneration` with `omitempty` behavior. Operator-side test that `createRunConfigFromMCPServer` sets the generation from the CR. Three pre-existing operator determinism tests continue to pass; they were the reason this PR pivoted from a `time.Time` RenderedAt field to `int64 MCPServerGeneration`.

## API Compatibility

No CRD or REST API changes. The RunConfig JSON schema gains an `omitempty` integer field — old consumers ignore unknown fields; new consumers see zero when reading old ConfigMaps.

No behavior change in the common case. Users will see a new pod-template annotation `toolhive.stacklok.dev/mcpserver-generation` on StatefulSets backing MCPServers created by upgraded operators. Users debugging a skipped apply will find a `DEBUG` log line on the proxyrunner pod: `skipping StatefulSet apply; newer MCPServer generation already applied`.

## Implementation plan

<details>

### Design

Monotonic version carried end-to-end from the CR to the StatefulSet annotation.

1. `RunConfig` gains `MCPServerGeneration int64` with JSON tag `mcpserver_generation,omitempty`. Zero is the unversioned / backward-compat signal.
2. `WithMCPServerGeneration(gen int64)` builder option.
3. Operator sets `runner.WithMCPServerGeneration(m.Generation)` in `createRunConfigFromMCPServer`.
4. `DeployWorkloadOptions.RunConfigMCPServerGeneration int64` carries it through the runtime seam.
5. `runner.Run` → `runtime.Setup` → k8s `DeployWorkload` plumb the value.
6. K8s runtime adds three helpers:
   - `runConfigGeneration(options)` — safe extract with nil-options handling.
   - `shouldSkipStatefulSetApply(ctx, ns, name, ourGen)` — `Get` the STS, parse the annotation, return `true` when `theirs > ours`. Parse errors warn and fall through. Not-found returns false. Zero ourGen returns false (no gate).
   - `applyStatefulSet(...)` — stamps the annotation when ourGen > 0, builds the apply-config, performs SSA.
7. New annotation constant `RunConfigMCPServerGenerationAnnotation = \"toolhive.stacklok.dev/mcpserver-generation\"` alongside `serviceFieldManager`. Doc comment calls out that the gate only becomes effective once proxyrunner is upgraded.

### Invariants

- Ties (equal generation) → apply (idempotent).
- No annotation on existing STS → apply + stamp.
- Strictly newer annotation → skip, return `(0, nil)`, log DEBUG.
- Unparseable annotation → log WARN, apply + re-stamp with ours.
- Zero ourGen → apply without gate, without stamp.

### Pivot during implementation

The first draft used `RenderedAt time.Time = time.Now().UTC()`. Two problems surfaced during test validation:

1. `time.Now()` made `createRunConfigFromMCPServer` non-deterministic. The operator's ConfigMap content-checksum would have changed every reconcile, which would have flipped the proxy Deployment's pod-template annotation and caused the proxy to roll on every reconcile — defeating the entire system. Three pre-existing determinism tests regressed to prove it.
2. `omitempty` does not actually drop zero `time.Time` (Go JSON quirk) — backward-compat output would have contained `\"0001-01-01T00:00:00Z\"`.

Pivoted to `int64 MCPServerGeneration` sourced from `MCPServer.metadata.generation`. `omitempty` works on int64. Generation is monotonic (K8s-enforced) and only bumps on spec changes, so RunConfig rendering is deterministic across no-op reconciles. The three determinism tests pass unchanged.

### Scope boundary

This PR fixes the image / pod-spec dimension of the race. It does not cover `spec.replicas` clobbering (the scaling dimension tracked in #4484). Phase 2 — moving StatefulSet ownership into the operator — is the complete architectural fix and would close both. Out of scope here.

</details>

- Related: **#4484** (same SSA-clobber root cause, scaling dimension — not closed by this PR). **#4635** (`GenerationChangedPredicate` efficiency work — same `.metadata.generation` signal, different use; Phase 1's deterministic Generation-based rendering also reduces the no-op-reconcile churn #4635 targets). **#4877** (closed — same family, different symptom).
- The gate is only effective once proxyrunner is upgraded. During an operator-only upgrade window the race window remains. The doc comment on the annotation const calls this out. Rolling operator + proxyrunner together is the intended upgrade path.
- Field manager is still shared (`toolhive-container-manager`) and apply still uses `Force: true`. The TOCTOU window between `Get` and `Apply` is acceptable because the worst case is losing to a strictly newer writer — which is the desired outcome.


pkg/container/kubernetes/client.go
```diff
@@ -397,22 +405,21 @@ func (c *Client) DeployWorkload(ctx context.Context,
...
-			Force:        true,
-		})
+	ourGen := runConfigGeneration(options)
+	skip, err := c.shouldSkipStatefulSetApply(ctx, namespace, containerName, ourGen)
 	if err != nil {
-		return 0, fmt.Errorf("failed to apply statefulset: %w", err)
+		return 0, err
+	}
+	if skip {
+		return 0, nil
```
yrobla, line 420: **[WARNING] `ensureBackendServices` is skipped when the gate fires**

When `shouldSkipStatefulSetApply` returns `true`, the early return at this line also skips `ensureBackendServices`. That means if backend services (headless + ClusterIP) were deleted or corrupted *after* the newer generation's apply and *before* this gated call, they won't be reconciled — leaving the workload unreachable until the next non-gated reconcile.

The happy-path assumption is that the newer gen already left services healthy, but partial-failure windows (e.g. service creation failed in the newer pod, then this pod is gated) can violate that.

Consider either:
- Still running `ensureBackendServices` in the skip path (requires a separate `Get` of the existing STS for owner-reference), or
- Adding an explicit comment documenting that service reconciliation is intentionally omitted when gated, and why the gap is acceptable.

pkg/container/kubernetes/client.go
```diff
@@ -435,6 +442,84 @@ func (c *Client) DeployWorkload(ctx context.Context,
...
+		return nil, fmt.Errorf("failed to apply statefulset: %w", err)
+	}
+	slog.Debug("applied statefulset", "name", createdStatefulSet.Name)
+	return createdStatefulSet, nil
+}
+
+// shouldSkipStatefulSetApply returns true when the existing StatefulSet is already
+// stamped with a strictly greater MCPServer generation than ours, meaning a newer
+// proxyrunner pod has already reconciled the workload and ours would be a regression.
+// Returns false (apply as normal) when ourGen is zero, when the StatefulSet does not
```
yrobla: **[NIT] Doc comment says "zero" but guard uses `<= 0`**

The comment on line 487 says:
> Returns false (apply as normal) when **ourGen is zero** …

but the guard at line 492 is `ourGen <= 0`. K8s generation is always ≥ 0 so the two are functionally identical, but the comment should match the code to avoid confusion. Either update the comment to say "zero or negative" or change the guard to `ourGen == 0`.

pkg/container/kubernetes/client.go
```diff
@@ -49,6 +49,14 @@ const (
...
 	// serviceFieldManager is the field manager name for server-side apply operations
 	serviceFieldManager = "toolhive-container-manager"
+
+	// RunConfigMCPServerGenerationAnnotation carries the MCPServer .metadata.generation that
+	// produced the RunConfig applied to this StatefulSet. Used as a monotonic version stamp
+	// to prevent stale proxyrunner pods (from an old Deployment ReplicaSet) from clobbering
+	// a newer RunConfig's apply. The gate only becomes effective once proxyrunner is upgraded
+	// to a version that reads this annotation; operator-only upgrades leave the race window
+	// in place until proxyrunner is also rolled.
+	RunConfigMCPServerGenerationAnnotation = "toolhive.stacklok.dev/mcpserver-generation"
```
yrobla, line 60: **[NIT] Exported constant while sibling `serviceFieldManager` is unexported**

`RunConfigMCPServerGenerationAnnotation` is exported (capital R) but the adjacent `serviceFieldManager` constant is unexported. Looking at the diff, this constant is only referenced within the `kubernetes` package (including the test file). If nothing outside the package needs it, making it unexported (`runConfigMCPServerGenerationAnnotation`) would be more consistent with the existing style.

If it is intentionally exported for operator or diagnostic tooling to reference, a short comment explaining that would help.

pkg/container/kubernetes/client_test.go
```diff
@@ -1367,3 +1368,185 @@ func TestDeployWorkloadBackendReplicas(t *testing.T) {
...
+				assert.Equal(t, tc.seedAnnotation,
+					sts.Spec.Template.Annotations[RunConfigMCPServerGenerationAnnotation],
+					"seeded annotation should be preserved when apply is gated")
+				return
+			}
+
+			// Apply should have occurred.
+			require.NoError(t, getErr)
+			switch tc.wantAnnotationIs {
+			case "missing":
```
yrobla, line 1542: **[NIT] Dead `"missing"` branch — corresponding test case never added**

The `case "missing":` branch in this switch is never reached: all seven test cases use `wantAnnotationIs: "equal"`. The natural case that would exercise it is **zero gen + no existing STS** (old-operator path creates a brand-new StatefulSet — apply should succeed, no annotation should be stamped). That case is currently untested.

Either:
- Add a test case like:
  ```go
  {
      name:             "zero_options_generation_no_existing_sts",
      seedSTS:          false,
      optionsGen:       int64(0),
      expectApply:      true,
      wantAnnotationIs: "missing",
  },
  ```
- Or remove the `"missing"` branch if this scenario is intentionally out of scope (and note why).

yrobla (commented):
Good implementation overall — the monotonic generation gate is well-reasoned, backward-compat signal is clean, and tests cover the seven core gate cases. Four findings below (warning + nits; no blockers).

yrobla (dismissed):

pkg/container/kubernetes/client.go
JAORMX: Thanks for flagging this. Spent time on it and I think leaving the skip in place is actually the correct call, but the missing comment is a fair ask.

The gate only fires when a peer pod has already stamped a *strictly greater* generation on the STS. That peer has just run `ensureBackendServices` moments before; if it succeeded, services are healthy; if it failed, the peer returned an error and will retry on its own.

Reconciling services from the gated (stale) pod would reintroduce the exact race this PR closes: `createHeadlessService`/`createMCPService` SSA under the same field manager `toolhive-container-manager` with `Force: true`, and the stale pod's RunConfig carries old `TargetPort`/`ProxyPort` values. The owner-ref is fine (we can derive it from the existing STS), but the port/spec fields would regress.

Added a comment at the skip site spelling this out so future contributors don't "helpfully" add the call back.

pkg/container/kubernetes/client_test.go
JAORMX: Good catch — added the `zero_options_generation_no_existing_sts` row. Verified the test now exercises the `"missing"` branch and asserts the zero-gen-never-stamps invariant.

pkg/container/kubernetes/client.go
JAORMX: Keeping it exported — it's a wire-format contract that future Phase 2 (operator-side STS ownership) and ops/diagnostic tooling will want to consume as a typed symbol rather than duplicate as a string literal. Added a godoc sentence explaining that.

pkg/container/kubernetes/client.go
JAORMX: Fair. Updated the godoc to say "zero or negative" to match the `<= 0` guard. Keeping the defensive `<= 0` — idiomatic for an int64 from an external struct.

yrobla (approved):

PR #5033, opened 2026-04-23
pkg/llm: implement OIDC token source and thv llm token command

### LINKED ISSUE #5028: thv llm: OIDC token lifecycle and thv llm token command
## Context

Part of #5016 (RFC: stacklok/toolhive-rfcs#70). Depends on #5027 (foundation).

## Scope

- Token source implementing a three-tier strategy:
  1. In-memory cache
  2. Secrets-provider refresh token
  3. Browser OIDC+PKCE flow
- Preemptive refresh 30s before expiry
- Access tokens held in memory only — never written to disk or logged
- Refresh tokens stored via ToolHive's existing secrets provider (OS keyring / encrypted file fallback), using `ScopeLLM`
- `thv llm token` command — prints a fresh JWT to stdout (all other output on stderr), suitable for use as `apiKeyHelper` or `auth.command`

## Acceptance Criteria

- [ ] AT: `thv llm token` prints a fresh JWT to stdout with all other output on stderr
- [ ] Unit: `thv llm token` exits with an error in non-interactive mode when no cached or refreshable token exists — it never launches a browser flow
- [ ] Unit: Token source uses the three-tier strategy with preemptive refresh 30s before expiry (tested with mocked OIDC provider and secrets provider)
- [ ] Access tokens are held in memory only and never written to disk or logged
- [ ] Refresh tokens are stored via ToolHive's existing secrets provider

## Dependencies

- #5027 (foundation: package skeleton, config types)

## References

- Parent issue: #5016

yrobla:
## Summary

Adds a three-tier TokenSource (in-memory → cached refresh token → browser OIDC+PKCE flow) for the LLM gateway. Access tokens are held in memory only; refresh tokens are persisted via ScopeLLM secrets. A 30 s preemptive refresh window avoids gateway rejections on expiry.

thv llm token is now fully implemented: non-interactive, prints a fresh JWT to stdout with all other output on stderr, suitable for use as apiKeyHelper or auth.command in Claude Code / Cursor.

Fixes #5028 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## API Compatibility

## Changes

| File | Change |
|------|--------|
|      |        |

## Implementation plan

<details>

</details>

## Large PR Justification

This is an isolated PR that covers a single functionality. Cannot be split.

### DIFF
--- a/pkg/llm/tokensource.go
+++ b/pkg/llm/tokensource.go
@@ -1,0 +1,257 @@
+// SPDX-FileCopyrightText: Copyright 2026 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package llm
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"log/slog"
+	"sync"
+	"time"
+
+	"golang.org/x/oauth2"
+
+	"github.com/stacklok/toolhive/pkg/auth/oauth"
+	"github.com/stacklok/toolhive/pkg/auth/remote"
+	"github.com/stacklok/toolhive/pkg/secrets"
+)
+
+// ErrTokenRequired is returned when a fresh token is needed but no cached or
+// refreshable token exists and the caller is non-interactive (browser flow
+// disabled). The caller should run "thv llm setup" to perform initial login.
+var ErrTokenRequired = errors.New(
+	"LLM gateway authentication required: run \"thv llm setup\" to log in",
+)
+
+// preemptiveRefreshWindow is how far before actual expiry a token is treated as
+// expired, triggering a proactive refresh before the gateway rejects it.
+const preemptiveRefreshWindow = 30 * time.Second
+
+// TokenRefUpdater is a callback invoked after a successful browser flow or token
+// refresh to persist the refresh token key and expiry into the application config.
+// The callback is called with the secret key and token expiry time.
+// Callers typically wire this to config.UpdateConfig.
+type TokenRefUpdater func(refreshTokenKey string, expiry time.Time)
+
+// TokenSource provides fresh LLM gateway access tokens using a three-tier strategy:
+//
+//  1. In-memory cached oauth2.TokenSource (auto-refreshes transparently)
+//  2. Refresh token stored in the secrets provider (restores across CLI invocations)
+//  3. Browser-based OIDC+PKCE flow (only when interactive is true)
+//
+// Access tokens are held in memory only and are never written to disk or logged.
+type TokenSource struct {
+	cfg             *Config
+	secretsProvider secrets.Provider
+	interactive     bool
+	tokenRefUpdater TokenRefUpdater
+	mu              sync.Mutex
+	tokenSource     oauth2.TokenSource
+}
+
+// NewTokenSource creates a TokenSource for the LLM gateway.
+// secretsProvider may be nil if the secrets store is unavailable (tier 2 is skipped).
+// tokenRefUpdater is called after login/refresh to persist the token reference into
+// config — pass nil to skip config persistence (useful in tests).
+// Set interactive to false for non-interactive callers such as thv llm token.
+func NewTokenSource(
+	cfg *Config, secretsProvider secrets.Provider, interactive bool, tokenRefUpdater TokenRefUpdater,
+) *TokenSource {
+	return &TokenSource{
+		cfg:             cfg,
+		secretsProvider: secretsProvider,
+		interactive:     interactive,
+		tokenRefUpdater: tokenRefUpdater,
+	}
+}
+
+// Token returns a valid LLM gateway access token.
+// It is safe for concurrent use.
+func (t *TokenSource) Token(ctx context.Context) (string, error) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	// Tier 1: in-memory token source handles auto-refresh transparently.
+	if t.tokenSource != nil {
+		tok, err := t.tokenSource.Token()
+		if err == nil && tok.Valid() {
+			return tok.AccessToken, nil
+		}
+		t.tokenSource = nil
+	}
+
+	// Tier 2: restore from a cached refresh token in the secrets provider.
+	if err := t.tryRestoreFromCache(ctx); err == nil && t.tokenSource != nil {
+		tok, err := t.tokenSource.Token()
+		if err == nil && tok.Valid() {
+			return tok.AccessToken, nil
+		}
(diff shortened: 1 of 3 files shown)


pkg/llm/tokensource.go
```diff
@@ -0,0 +1,349 @@
...
+// Token returns a valid LLM gateway access token.
+// It is safe for concurrent use.
+func (t *TokenSource) Token(ctx context.Context) (string, error) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	// lastErr tracks the most recent actionable error from tiers 1/2. In
+	// non-interactive mode we return it instead of the generic ErrTokenRequired
+	// so the caller sees the real cause (e.g. invalid_grant, keyring locked).
+	var lastErr error
```
jerm-dro, line 99: **suggestion:** Each `thv llm token` invocation is a separate process — AI tools spawn it fresh each time they need a token. The in-memory tier provides no benefit across processes, so every invocation hits tier 2 (load refresh token → send to IdP → get new access token).

OIDC providers commonly rotate refresh tokens on use. If two `thv llm token` processes overlap (e.g., the AI tool fires two requests in quick succession), both read the same refresh token, both send it to the IdP, and one gets `invalid_grant` — which may invalidate the session entirely depending on the provider. Worth considering how to serialize or deduplicate concurrent token refreshes across processes.

jerm-dro (dismissed):

pkg/llm/tokensource.go
```diff
@@ -0,0 +1,403 @@
...
+//     on the refresh token when multiple short-lived processes run concurrently)
+//  3. Refresh token stored in the secrets provider (restores across CLI invocations)
+//  4. Browser-based OIDC+PKCE flow (only when interactive is true)
+//
+// Both the refresh token and the short-lived access token are stored in the
+// secrets provider (system keychain). The access token is cached alongside the
+// refresh token to prevent concurrent processes from racing to exchange the same
+// refresh token — a race that causes invalid_grant on providers that rotate
+// refresh tokens on use. Access tokens are never written to log output.
+type TokenSource struct {
```
aponcedeleonch, line 63: This file duplicates `pkg/registry/auth/oauth_token_source.go` almost verbatim. `ensureOfflineAccess` is byte-for-byte identical. `DeriveSecretKey` differs only in the prefix (`LLM_OAUTH_` vs `REGISTRY_OAUTH_`). `Token()`, `tryRestoreFromCache`, `performBrowserFlow`, `buildFlowConfig`, `buildOAuth2Config`, `makeTokenPersister`, and `refreshTokenKey` all have near-identical implementations. The comment at line 436 even says "matching the pattern used by the registry auth package."

I think the right move is to extract a shared `OAuthTokenSource` into `pkg/auth/oauth/` parameterized by an OIDC config getter, a `secretKeyDerivation` function (so registry gets `REGISTRY_OAUTH_` and LLM gets `LLM_OAUTH_`), and a `configPersister` callback. The `TokenRefUpdater` pattern introduced here is already cleaner than the registry's direct `config.UpdateConfig` call, so the LLM side is a nice starting point for the abstraction. That also lets the genuinely new bits (preemptive refresh, access-token cache) apply to both consumers.

If the extraction feels too big for this PR, I'd be fine landing the duplication behind a tracking issue that pulls the registry side into the new abstraction as a follow-up, but I don't want to drift further without a plan.

pkg/llm/tokensource.go
```diff
@@ -0,0 +1,403 @@
...
+		return &shifted, nil
+	}
+	return tok, nil
+}
+
+// withPreemptiveRefresh wraps src so tokens appear expired preemptiveRefreshWindow
+// before they actually expire, then re-wraps with ReuseTokenSource so the refresh
+// is only triggered once per window.
+func withPreemptiveRefresh(src oauth2.TokenSource) oauth2.TokenSource {
+	return oauth2.ReuseTokenSource(nil, &preemptiveTokenSource{inner: src})
```
aponcedeleonch, line 391: I think this composition doesn't behave as advertised. Both callers pass in `src` values that are already `oauth2.ReuseTokenSource` instances (`flow.TokenSource()` at line 301, and `remote.CreateTokenSourceFromCached` at line 270). When the outer wrapper's cached token hits the shifted expiry:

1. Outer `ReuseTokenSource` says expired, calls preemptive.
2. Preemptive calls inner, which sees the real expiry and returns the same cached token unchanged.
3. Preemptive shifts the expiry again, outer caches it.
4. Next call: outer says expired again (same shifted expiry).

So every `Token()` call inside the 30-second preemptive window triggers a full inner round-trip rather than one refresh per window. The fix is to wrap the refresher before `oauth2.ReuseTokenSource`, not after. A regression test with a fake refresher that counts calls would catch this cleanly.

pkg/llm/tokensource.go
```diff
@@ -0,0 +1,403 @@
...
+
+// makeTokenPersister returns a remote.TokenPersister that stores the refresh
+// token in the secrets provider and updates the config expiry reference.
+func (t *TokenSource) makeTokenPersister(key string) remote.TokenPersister {
+	return func(refreshToken string, expiry time.Time) error {
+		ctx := context.Background()
+		if err := t.secretsProvider.SetSecret(ctx, key, refreshToken); err != nil {
+			return fmt.Errorf("persisting LLM gateway refresh token: %w", err)
+		}
+		t.updateConfigTokenRef(key, expiry)
```
aponcedeleonch, line 287: When a rotated refresh token is written here, the associated `_AT` cache entry is left as-is. Between this write and the next `Token()` call, tier 1.5 could serve an access token minted against the now-rotated refresh. The preemptive window masks this in practice, but overwriting or deleting the `_AT` entry inside the persister callback is cheap insurance.

aponcedeleonch (commented):
Three inline comments on the new token source: one architectural (the bulk of this file duplicates `pkg/registry/auth/oauth_token_source.go`), and two concrete correctness issues (`withPreemptiveRefresh` composition, and the `_AT` cache not being invalidated on refresh-token rotation). Happy to pair on the shared extraction if that's the direction.

pkg/llm/tokensource.go
yrobla: https://github.com/stacklok/toolhive/issues/5050

aponcedeleonch (approved):

pkg/llm/tokensource.go
yrobla: https://github.com/stacklok/toolhive/issues/5051

PR #5035, opened 2026-04-23
Add localhost reverse proxy for LLM gateway

### LINKED ISSUE #5029: thv llm: localhost reverse proxy for proxy-mode tools
## Context

Part of #5016 (RFC: stacklok/toolhive-rfcs#70). Depends on #5028 (token lifecycle). Can be implemented in parallel with #5028 once #5027 is done.

## Scope

- Localhost reverse proxy listening on `http://localhost:<port>/v1`
- Strips incoming `Authorization` header and injects a fresh OIDC bearer token from the token source
- Forwards requests to upstream gateway preserving original path, query string, headers, and body
- Passes through upstream responses faithfully: SSE streaming (chat completions), non-streaming JSON (embeddings, completions), error responses (4xx/5xx) — no buffering or body modification
- Loopback-only guard: reject startup if listen address is not a loopback interface
- Background process lifecycle management (start, stop, PID tracking)
- `thv llm proxy start` — foreground variant with full log output for debugging

## Acceptance Criteria

- [ ] AT: Proxy accepts requests on `http://localhost:<port>/v1`, strips `Authorization`, injects fresh OIDC bearer token, forwards to upstream gateway — preserving path, query string, headers, and body
- [ ] AT: Proxy passes through SSE streaming, non-streaming JSON, and error responses (4xx/5xx) without buffering or modification
- [ ] Unit: Proxy rejects startup if listen address is not a loopback interface

## Dependencies

- #5027 (foundation: package skeleton, config types)
- #5028 (token source, for token injection)

## References

- Parent issue: #5016

yrobla:
## Summary

Implements a localhost reverse proxy that bridges AI tools using static API keys (e.g. Cursor) to an OIDC-protected LLM gateway.

**New \`pkg/llm/proxy\` package:**
- \`proxy.go\`: \`httputil.ReverseProxy\` wrapper that strips incoming \`Authorization\` headers, injects fresh OIDC tokens, and forwards to the configured gateway. Binds listener in \`New()\` so \`Addr()\` is valid before \`Start()\`. Loopback guard rejects non-localhost listen addresses. SSE streaming works without buffering (\`FlushInterval: -1\`). Graceful 5-second shutdown on context cancellation.

**CLI wiring (\`cmd/thv/app/llm.go\`):**
- \`thv llm proxy start\`: runs in the foreground (blocking), consistent with all other \`thv proxy\` commands. Users who want background execution can use shell backgrounding (\`thv llm proxy start &\`) or a process manager.

No daemon/PID-file machinery — the proxy lifecycle follows the same foreground pattern as \`thv proxy stdio\` and \`thv proxy tunnel\`.

Fixes #5029

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [ ] E2E tests (\`task test-e2e\`)

## API Compatibility

- [x] This PR does not break the \`v1beta1\` API, OR the \`api-break-allowed\` label is applied and the migration guidance is described above.

## Changes

| File | Change |
|------|--------|
| \`pkg/llm/proxy/proxy.go\` | New: reverse proxy with token injection, listener binding, graceful shutdown |
| \`pkg/llm/proxy/proxy_test.go\` | New: unit tests for loopback guard, token injection, streaming, error paths, ErrTokenRequired |
| \`cmd/thv/app/llm.go\` | New \`thv llm proxy start\` command (foreground, blocking) |

Yes. \`thv llm proxy start\` starts a localhost reverse proxy on \`127.0.0.1:14000\` (configurable) that transparently injects OIDC tokens for AI tools configured to use a static API key endpoint.

The proxy intentionally has no \`stop\` subcommand or PID-file daemon — this matches how all other ToolHive proxy commands work (\`thv proxy stdio\`, \`thv proxy tunnel\`). Users signal termination with Ctrl+C or their process manager.


cmd/thv/app/llm.go
```diff
@@ -289,23 +292,112 @@ func newLLMProxyCommand() *cobra.Command {
...
 	}
 }
 
+// runLLMProxyForeground builds a TokenSource and starts the proxy in this process.
+func runLLMProxyForeground(ctx context.Context, llmCfg *llm.Config) error {
+	secretsProvider, err := secrets.GetSystemSecretsProvider()
+	if err != nil {
+		return fmt.Errorf("failed to get secrets provider: %w", err)
+	}
+	scoped := pkgsecrets.NewScopedProvider(secretsProvider, pkgsecrets.ScopeLLM)
```
jerm-dro: **suggestion:** The token source is created with `interactive: false`, so if the refresh token expires or is revoked, `Token()` returns `ErrTokenRequired` and every proxied request gets a 502. For a long-running background daemon, this could happen hours or days after start — the user sees opaque failures from their AI tool with no indication that re-authentication is needed.

Consider logging a clear message to the log file with instructions (e.g., "run `thv llm setup` to re-authenticate"), or including an actionable hint in the 502 response body since some tools surface error text.

pkg/llm/proxy/daemon.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return err
+	}
+	if !running {
+		return errors.New("LLM proxy is not running")
+	}
+
+	if err := process.KillProcess(pid); err != nil {
+		return fmt.Errorf("stopping process %d: %w", pid, err)
+	}
```
jerm-dro: **suggestion:** `Stop()` kills by PID without verifying the process is actually the LLM proxy. If the proxy crashes (no graceful shutdown), the PID file goes stale. If the OS reuses that PID for an unrelated process, `Stop()` will kill it.

`isRunningAt` checks liveness via signal 0 but not process identity. A lightweight mitigation: verify the process command line contains `thv` before killing (the `process` package already has `hasToolHiveDetachedEnv` which checks for the `TOOLHIVE_DETACHED` env var on a process).

jerm-dro (dismissed):

pkg/llm/proxy/proxy.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Addr returns the actual bound listen address (e.g. "127.0.0.1:14000").
+func (p *Proxy) Addr() string {
+	return p.listener.Addr().String()
+}
+
+// checkLoopback returns an error if addr is not a loopback interface.
+// addr must be a host:port string with a raw IP host (not a hostname).
+func checkLoopback(addr string) error {
```
JAORMX: **Blocker (reuse).** So, `pkg/networking/utilities.go` already ships `IsPrivateIP`, `AddressReferencesPrivateIp`, and `IsLocalhost`. RFC-0070 §Validation explicitly calls out `ValidateLoopbackAddress()` as the shared helper, and it's named that way for a reason... vMCP and whatever grows next will want the same check.

What this PR does instead: a private, narrower `checkLoopback` that lives inside the LLM proxy package and doesn't help anyone else.

The fix: add `ValidateLoopbackAddress(addr string) error` to `pkg/networking/utilities.go` next to `AddressReferencesPrivateIp`, and call it from `New()`. Delete this private function.

Also: `New()` now hard-codes `"127.0.0.1:%d"` at line 55, so the hostname branch this doc comment implies never runs. The comment at line 77 still says "must be a host:port string with a raw IP host (not a hostname)" which is technically true but redundant with the hard-coding. Per `.claude/rules/go-style.md` "Keep Comments Synchronized" ... if we're moving this into `pkg/networking` and hardening it properly (accept hostnames that resolve only to loopback), the comment goes with it.

Blocker because the whole point of building this inside ToolHive instead of as a standalone binary was reuse (RFC-0070 §Alternatives Considered). If we copy/paste every validator we need, that argument falls over.

cmd/thv/app/llm.go
```diff
@@ -297,23 +298,67 @@ func newLLMProxyCommand() *cobra.Command {
...
+			}
+
+			llmCfgCopy := llmCfg
+			return runLLMProxyForeground(cmd.Context(), &llmCfgCopy)
 		},
 	}
 }
 
+// runLLMProxyForeground builds a TokenSource and starts the proxy in this process.
+func runLLMProxyForeground(ctx context.Context, llmCfg *llm.Config) error {
```
JAORMX, line 275: **Blocker (reuse).** This function and `runLLMToken` at line 211 build the exact same pipeline:

```
GetSystemSecretsProvider → NewScopedProvider(ScopeLLM) → identical updater closure → NewTokenSource
```

The `updater` closure here (lines 342-350) is copy-pasted verbatim from `runLLMToken` (lines 225-233). `thv llm setup` will be the third caller when it lands, and it'll end up with its own copy too.

Extract a single `llm.NewDefaultTokenSource(cfg, interactive) (*llm.TokenSource, error)` in the `llm` package so:

- The updater lives in one place.
- `ScopeLLM` is locked in at the library boundary, not at every caller.
- When we change how LLM refresh-token persistence works (say, we stop using `CachedRefreshTokenRef`), we change it once.

Per `.claude/rules/go-style.md` "Prefer Existing Code and Packages Over From-Scratch Implementations" and `.claude/rules/cli-commands.md` thin-wrapper rule. Blocker alongside the loopback helper because we committed to building this inside ToolHive specifically *because* we have these building blocks. If we copy them, we lose the leverage.

pkg/llm/proxy/proxy.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		tokenCtx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
+		defer cancel()
+		token, err := p.tokenSource.Token(tokenCtx)
+		if err != nil {
+			if errors.Is(err, llm.ErrTokenRequired) {
+				slog.Error("LLM proxy: re-authentication required", "error", err)
+				http.Error(w, err.Error(), http.StatusBadGateway)
```
JAORMX: **Blocker (RFC compliance + response leak).** RFC-0070 §Proxy is explicit:

> **OpenAI-compatible errors**. Token acquisition failures return HTTP 401 with an error body matching the OpenAI API schema, so tools display meaningful messages.

What this PR does: detects `ErrTokenRequired` (good), but still returns `502` instead of `401`, and writes `err.Error()` straight into the response body. Two problems:

1. `502` is wrong. The client-side failure mode that matters for a background proxy is "refresh token was rotated or revoked while I slept". When that happens days later, Cursor shows... "Bad Gateway". The user has no idea to run `thv llm setup` again.
2. `err.Error()` on the wire leaks whatever the TokenSource error happens to contain. If the underlying error wraps `*oauth2.RetrieveError`, that includes the IdP's response body. Some IdPs echo submitted parameters back.

Minimum change:

```go
if errors.Is(err, llm.ErrTokenRequired) {
    slog.Error("LLM proxy: re-authentication required")
    w.Header().Set("Content-Type", "application/json")
    w.WriteHeader(http.StatusUnauthorized)
    _, _ = io.WriteString(w, `{"error":{"message":"LLM gateway authentication required: run 'thv llm setup' to log in","type":"authentication_error","code":"token_required"}}`)
    return
}
slog.Error("LLM proxy: failed to obtain token")
w.Header().Set("Content-Type", "application/json")
w.WriteHeader(http.StatusBadGateway)
_, _ = io.WriteString(w, `{"error":{"message":"LLM gateway token fetch failed","type":"server_error"}}`)
return
```

Note: I dropped the `"error", err` structured field too. See the separate comment on line 113 about token material leaking through log.

Blocker because this is the single failure mode the user is most likely to hit in production, and today it gives them nothing to work with.

pkg/llm/proxy/proxy.go
```diff
@@ -0,0 +1,156 @@
...
+			pr.SetURL(p.gatewayURL)
+		},
+		FlushInterval: -1,
+		ErrorHandler: func(w http.ResponseWriter, r *http.Request, err error) {
+			slog.Error("LLM proxy upstream error", "error", err, "path", r.URL.Path)
+			http.Error(w, "upstream error", http.StatusBadGateway)
+		},
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
JAORMX, line 124: **Blocker (security + RFC §Threat Model).** So... "trusts localhost" is doing a lot of work here. A browser loading attacker-controlled JS can rebind `evil.com` to `127.0.0.1` and send credentialed requests through this proxy. We'd happily mint a full OIDC token and forward the request to the upstream gateway, based purely on the fact that the TCP connection landed on a loopback socket.

Loopback-only bind guards *where* the socket listens. It doesn't guard *what `Host` the client claims to be talking to*.

Four lines in the handler close it:

```go
host := r.Host
if i := strings.IndexByte(host, ':'); i >= 0 {
    host = host[:i]
}
if host != "127.0.0.1" && host != "localhost" && host != "[::1]" && host != "::1" {
    http.Error(w, "forbidden", http.StatusForbidden)
    return
}
```

(Or reuse `networking.IsLocalhost(r.Host)` directly... another reuse opportunity since that helper already exists.)

Explicit in RFC-0070 §Threat Model. This is the cheap half of the "local auth oracle" concern. The harder half (UNIX socket, SO_PEERCRED, or a shared-secret header for cross-process caller identity) can be a follow-up. The Host-header check should not.

pkg/llm/proxy/proxy.go
```diff
@@ -0,0 +1,156 @@
...
+		tokenCtx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
+		defer cancel()
+		token, err := p.tokenSource.Token(tokenCtx)
+		if err != nil {
+			if errors.Is(err, llm.ErrTokenRequired) {
+				slog.Error("LLM proxy: re-authentication required", "error", err)
+				http.Error(w, err.Error(), http.StatusBadGateway)
+			} else {
+				slog.Error("LLM proxy: failed to obtain token", "error", err)
```
JAORMX: **Major (log hygiene).** Both branches here (`slog.Error` at line 110 and 113) include `"error", err` in the structured log. That's fine in spirit, but `err` may wrap `*oauth2.RetrieveError` coming out of the token source, whose `Error()` serializes the full IdP response body. Some IdPs echo submitted parameters back in errors (the refresh token, client assertion), and then we've quietly written bearer material into whatever slog handler is active.

`.claude/rules/go-style.md`: "Never log credentials, tokens, API keys, or passwords." Getting close to the bone here.

A small sanitizing wrapper in `pkg/llm/tokensource.go` that extracts only `RetrieveError.ErrorCode` and `ErrorDescription` (never `Body`) would fix it. Log the sanitized form from here; let callers that don't log still see the full error. Or at minimum, drop the `"error", err` field from both slog lines now and let the classification (`ErrTokenRequired` vs other) carry the signal.

pkg/llm/proxy/proxy.go
```diff
@@ -0,0 +1,156 @@
...
+	p.server = &http.Server{
+		Handler:           p.handler(),
+		ReadHeaderTimeout: 30 * time.Second,
+	}
+
+	slog.Info("LLM proxy started", "addr", p.Addr(), "gateway", p.cfg.GatewayURL)
+
+	serveErr := make(chan error, 1)
+	go func() {
+		if err := p.server.Serve(p.listener); err != nil && err != http.ErrServerClosed {
```
JAORMX: **Minor.** `.claude/rules/go-style.md` error-handling section says `errors.Is` for error inspection:

```suggestion
		if err := p.server.Serve(p.listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
```

`errors` is already imported (line 9), so it's a one-line change.

pkg/llm/proxy/proxy.go
```diff
@@ -0,0 +1,156 @@
...
+		return nil, fmt.Errorf("cfg must not be nil")
+	}
+	if ts == nil {
+		return nil, fmt.Errorf("ts must not be nil")
+	}
+	gatewayURL, err := url.Parse(cfg.GatewayURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid gateway URL %q: %w", cfg.GatewayURL, err)
+	}
+	if gatewayURL.Scheme == "" || gatewayURL.Host == "" {
```
JAORMX: **Minor (defence in depth).** `config.Validate()` at the config boundary already checks `gateway_url` is HTTPS, but the proxy itself doesn't re-verify at startup. A user editing `config.yaml` by hand (no validation on disk reads) can bypass this and ship plaintext bearer tokens to an `http://` upstream.

Cheap:

```go
if gatewayURL.Scheme != "https" {
    return nil, fmt.Errorf("gateway URL must use HTTPS, got scheme %q", gatewayURL.Scheme)
}
```

Enforces the invariant at the boundary where it actually matters (plaintext token on the wire). The current check at this line only asserts scheme+host are non-empty, which `http://whatever.com` satisfies.

pkg/llm/proxy/proxy.go
```diff
@@ -0,0 +1,156 @@
...
+				slog.Error("LLM proxy: re-authentication required", "error", err)
+				http.Error(w, err.Error(), http.StatusBadGateway)
+			} else {
+				slog.Error("LLM proxy: failed to obtain token", "error", err)
+				http.Error(w, "failed to obtain authentication token", http.StatusBadGateway)
+			}
+			return
+		}
+		r = r.Clone(r.Context())
+		r.Header.Del("Authorization")
```
JAORMX: **Nit.** Two small things:

1. `r.Header.Del("Authorization")` before `r.Header.Set("Authorization", ...)` is redundant. `Set` replaces any existing value. Drop the `Del`.
2. The `httputil.ReverseProxy` at line 93 is allocated on every call to `handler()`. `handler()` is only called once today (from `Start` at line 129), so this is benign. Still, stashing `rp` on the struct in `New()` is the cleaner shape, and if you do that you can also move the Authorization set into the `Rewrite` hook on `pr.Out.Header`. That eliminates the `r.Clone(...)` at line 118 entirely, since ReverseProxy already clones internally.

JAORMX (dismissed):
Solid work on the scope trim since the last push. Dropping the daemon lifecycle and running the proxy in the foreground only was the right call. Requesting changes on two axes.

**Reuse.** The whole reason we're building this inside ToolHive and not as a standalone binary is that we already have the pieces. RFC-0070 §Alternatives Considered is explicit about this. So when we ship a second copy of something that already exists, the RFC's own argument for why this lives here starts to wobble. Two spots in the current diff:

- `checkLoopback` in `proxy.go:78` reinvents a narrow version of helpers already in `pkg/networking/utilities.go` (`IsPrivateIP`, `AddressReferencesPrivateIp`, `IsLocalhost`). RFC-0070 §Validation explicitly named `ValidateLoopbackAddress` as the shared helper. It never got added.
- TokenSource construction is built twice in `cmd/thv/app/llm.go`: `runLLMToken` (lines 211-243) and `runLLMProxyForeground` (lines 335-360) have identical `GetSystemSecretsProvider → NewScopedProvider(ScopeLLM) → updater → NewTokenSource` pipelines, with a copy-pasted `updater` closure. `thv llm setup` will be the third caller when it lands.

Easier to consolidate now. Two callers today, three inevitable.

**RFC compliance.** Two gaps that matter:

- Token-fetch failure now distinguishes `ErrTokenRequired` (good), but still returns `502` and writes `err.Error()` into the response body. RFC §Proxy is explicit: `401` + OpenAI-compatible JSON schema. Without this, the daemon silently stops working days later (rotated refresh token) and Cursor users see "Bad Gateway" with zero path to recovery. Writing `err.Error()` into the body also leaks internal error detail to whatever is on the other end.
- No `Host`-header allowlist. DNS rebinding turns the proxy into an auth oracle for browser JS. Four-line fix, directly in RFC §Threat Model.
- Minor: `/healthz` endpoint from RFC §Proxy isn't implemented. RFC also specified `pkg/llm/internal/proxy.go`; PR uses non-internal `pkg/llm/proxy/`. Either fix the layout or update the RFC.

**Housekeeping.** A few meta items:

- The PR is behind main (`mergeStateStatus: BEHIND`, merge-base is `854e0ab7`, main is at `68f4c2f1`). Please rebase onto current main before merge so the diff view stops showing phantom deletions of `pkg/llm/manage.go`, `pkg/oidc/*`, and `test/e2e/cli_llm_config_test.go` from #5049.
- Please squash the two "changes from review" fixups into the original commit.
- Per `.claude/rules/pr-creation.md`, optional PR sections should be filled in or removed. The empty `<details>` block, the empty `Changes` table row, and the unused `Does this introduce a user-facing change?` section all want deletion.

Happy to re-review once the blockers land.

JAORMX (approved):

Issue #4978, opened 2026-04-21
Authserver DCR integration (Phase 2, Steps 2a-2g)


jhrozek:
## Description

Phase 2 of the DCR story (#4976). Wire RFC 7591 Dynamic Client Registration into `pkg/authserver/` upstream initialization so an OAuth2 upstream can be configured without a pre-issued `client_id`.

This issue has been broken into four sub-issues, each targeting a ≤1000 line PR, with CRD changes deferred to the last PR so the operator API evolves only after the authserver plumbing is fully merged and stable.

## Sub-Issues

- **#5037** — Discovery function and config structs (Steps 2a/2b/2e): `FetchAuthorizationServerMetadata`, `DCRUpstreamConfig`, `OAuth2Config.Validate()` XOR check. Purely additive, no behavior change, no CRD changes.
- **#5038** — Credential store and resolver (Steps 2f/2c): `DCRCredentialStore` interface + in-memory implementation, `resolveDCRCredentials` with full RFC 7591/7592/8414 flow.
- **#5039** — Wiring, observability, and integration tests (Steps 2d/2g): thread DCR resolver into `buildUpstreamConfigs`, three structured-log upgrades, end-to-end integration tests.
- **#5040** — Operator CRD surface (CRD changes last): expose `DCRConfig` on the `OAuth2UpstreamConfig` CRD type, conversion from CRD → authserver config, regenerate manifests.

## Dependency Chain

```
#4977 (Phase 1 rename, pkg/oauthproto exists)
  └─► #5037 (discovery fn + authserver config structs)
        └─► #5038 (credential store + resolver)
              └─► #5039 (wiring + logs + integration tests)
                    └─► #5040 (operator CRD surface — CRD changes last)
                          └─► #4979 (Phase 3 persistent backends)
```

## Out of Scope

- Persistent `DCRCredentialStore` backends (memory + Redis) — Phase 3, #4979.
- Authserver OTEL metrics + tracing.
- Automatic re-register on `invalid_client` at the upstream token endpoint.
- CLI persistence of `registration_client_uri` in `pkg/auth/remote/config.go`.
- RFC 8707 resource-indicator confirmation for the existing upstream flow.
- RFC 7592 `client_secret` rotation via PUT.
- Provider-specific quirks (`login.microsoftonline.com` `resource` param, `mcp.zoho.com` `access_type`).
- Initial access token rotation / refresh.
- Any change to `CommonOAuthConfig.Validate()`.
[...]

Issue #4977, opened 2026-04-21
Rename pkg/oauth to pkg/oauthproto and relocate DCR primitives


jhrozek:
## Description

Phase 1 of the DCR story (#4976). Rename the existing `pkg/oauth` leaf package to `pkg/oauthproto`, move the RFC 7591 Dynamic Client Registration primitives out of `pkg/auth/oauth/dynamic_registration.go` into `pkg/oauthproto/dcr.go`, and relocate the handful of helpers (`IsLocalhost`, `UserAgent`) they depend on so that `pkg/oauthproto/` stays a true leaf package (no `pkg/networking` import). This unblocks Phase 2: once the rename lands, `pkg/authserver/` can import the DCR client without creating an import cycle.

## Context

Today, DCR primitives live in `pkg/auth/oauth/dynamic_registration.go`, which transitively imports `pkg/networking`. The authserver cannot import `pkg/auth/oauth` without creating a cycle, so there is no path for `pkg/authserver/` to reuse the DCR client. This task solves that by preserving the `pkg/oauth` leaf-package invariant under a new name (`pkg/oauthproto`) and moving only the pure protocol primitives into it. `FetchAuthorizationServerMetadata`, `DCRUpstreamConfig`, and the authserver wiring are deliberately deferred to Phase 2 (#4978).

Key design points carried into this task:

- `pkg/oauthproto/` must remain a leaf package. `IsLocalhost` is therefore relocated **into** `pkg/oauthproto/` (as a private `isLoopbackHost`), and `pkg/networking` imports it back out rather than the other way around. This keeps the `pkg/oauthproto` dep graph pointing at `net/url` / `errors` / `fosite` only.
- `RegisterClientDynamically` takes `*http.Client` directly (nil means "build a default one"). The private `registerClientDynamicallyWithClient` helper and the `networking.HTTPClient` interface indirection are dropped in this move.
- The CLI-specific error string in `handleHTTPResponse` (referencing `--remote-auth-client-id` / `--remote-auth-client-secret`) is generalized *before* the move, because the protocol package must not leak CLI assumptions.
- The CLI-side `NewDynamicClientRegistrationRequest(scopes, callbackPort)` helper — which hardcodes `http://localhost:<port>/callback` — stays out of the protocol package and is hosted in `pkg/auth/discovery/`, its sole caller.
[...]

PR #5036, opened 2026-04-23
Rename pkg/oauth to pkg/oauthproto and move DCR primitives

tgrunnagle:
## Summary

- The authserver could not import `pkg/auth/oauth` to reuse the Dynamic Client Registration (DCR) client because doing so would create an import cycle. This blocked Phase 2 of the DCR story (#4978), which wires the authserver to upstream OAuth providers.
- This PR renames `pkg/oauth` to `pkg/oauthproto`, moves the RFC 7591 DCR primitives from `pkg/auth/oauth/dynamic_registration.go` into `pkg/oauthproto/dcr.go`, and relocates the loopback-host helper into `pkg/oauthproto` so the package remains a true leaf with no `pkg/networking` dependency.

Implements (most of) #4977 (the authserver still defines its own DCRRequest/DCRResponse in pkg/authserver/server/registration/dcr.go alongside the new types in pkg/oauthproto. Reconciling Scope string vs ScopeList is the blocking work, so it doesn't belong in this PR).

## Type of change

## Test plan

Unit tests for all moved DCR logic now live in `pkg/oauthproto/dcr_test.go` (package `oauthproto_test`), covering `RegisterClientDynamically` happy paths, nil-client default, endpoint validation, `validateAndSetDefaults`, `handleHTTPResponse` error mapping (including regression guard that the CLI-flag hint is absent), `ScopeList` marshal/unmarshal, `isLoopbackHost`, and the 1 MB body-cap via `io.LimitReader`.

## Changes

| File | Change |
|------|--------|
| `pkg/oauth/` (all files) | Renamed directory to `pkg/oauthproto/`; `package oauth` → `package oauthproto` |
| `pkg/oauthproto/dcr.go` | DCR types, `RegisterClientDynamically`, and helpers moved from `pkg/auth/oauth/dynamic_registration.go`; `getHTTPClient`/`registerClientDynamicallyWithClient` removed; `*http.Client` parameter replaces the `networking.HTTPClient` interface indirection |
| `pkg/oauthproto/dcr_test.go` | New file — DCR unit tests relocated from `pkg/auth/oauth/dynamic_registration_test.go`; mock-client tests rewritten using `httptest.NewServer` |
| `pkg/oauthproto/locality.go` | New file — `IsLoopbackHost` (exported) hosts the canonical loopback check; keeps `pkg/oauthproto` free of any `pkg/networking` import |
| `pkg/oauthproto/constants.go` | `UserAgent = "ToolHive/1.0"` moved here from `pkg/auth/oauth/oidc.go` |
| `pkg/oauthproto/doc.go` | Updated to document the expanded (still leaf-package) surface including DCR and the leaf-package invariant |
| `pkg/networking/utilities.go` | `IsLocalhost` made a thin wrapper over `oauthproto.IsLoopbackHost` — all existing callers unchanged |
| `pkg/auth/discovery/dcr_request.go` | New file — `NewDynamicClientRegistrationRequest` factory (CLI-specific loopback redirect URI assumption) extracted here from `pkg/auth/oauth/dynamic_registration.go` |
| `pkg/auth/discovery/discovery.go` | `registerDynamicClient` updated to call the local factory and `oauthproto.RegisterClientDynamically` |
| `pkg/auth/oauth/dynamic_registration.go` | Deleted |
| `pkg/auth/oauth/dynamic_registration_test.go` | Deleted (tests moved to `pkg/oauthproto/dcr_test.go`) |
| 17 import sites | `pkg/oauth` → `pkg/oauthproto`; the 8 sites that already aliased as `oauthproto` drop the alias |

## Implementation Details

- `pkg/oauthproto` is a leaf package: it depends only on the standard library and `fosite` (already pulled in by `redirect.go`). `pkg/networking` now imports `pkg/oauthproto` for the loopback check, not the other way around. This is the structural change that unblocks `pkg/authserver` from importing the DCR client in Phase 2.
- `RegisterClientDynamically` now takes `*http.Client` directly (nil means "use the 30 s / 10 s / 10 s default"). The former `networking.HTTPClient` interface indirection and the private `registerClientDynamicallyWithClient` entry point are removed, eliminating a layer of unnecessary abstraction.
- The CLI-specific error hint ("Please configure OAuth client credentials using `--remote-auth-client-id`…") in `handleHTTPResponse` is replaced with a protocol-neutral message before the function moves into the protocol package.
- `NewDynamicClientRegistrationRequest` remains in `pkg/auth/discovery/` because it hardcodes `http://localhost:<port>/callback` — a CLI assumption that must not leak into a protocol-level package.

No.

- Import cycle elimination is the core goal. `go list -deps ./pkg/oauthproto/...` should show no `pkg/networking` path. A `grep -r "pkg/networking" pkg/oauthproto` returns no hits.
- The `pkg/auth/oauth/dynamic_registration_test.go` → `pkg/oauthproto/dcr_test.go` rename is a large hunk in the diff but is mostly a structural move; the test logic is equivalent with mock-based tests converted to `httptest.NewServer`.
- `pkg/oauth/doc.go` is deleted; `pkg/oauthproto/doc.go` replaces it with updated content reflecting the expanded surface area.
- Phase 2 (#4978) is blocked on this PR landing first.

## Large PR Justification

- Refactor has to be done in one PR


jhrozek (dismissed):
Looks good. One follow-up worth tracking under #4977 before we close it out: the authserver still defines its own DCRRequest/DCRResponse in pkg/authserver/server/registration/dcr.go alongside the new types in pkg/oauthproto. Reconciling Scope string vs ScopeList is the blocking work, so it doesn't belong in this PR, but it's worth a line in the #4977 plan so the two don't drift.

Minor aside, not blocking: the doc.go surface-area bullet says "RFC 7591 Dynamic Client Registration: request/response types" which overclaims slightly while the authserver still has its own. Tightening the wording to make it clear these are the client-side types would avoid future confusion.

jhrozek (dismissed):

jhrozek (approved):

PR #5041, opened 2026-04-23
Add authserver DCR discovery and config surface

### LINKED ISSUE #5037: Authserver DCR: discovery function and config structs (Phase 2, Steps 2a/2b/2e)
## Context

Part of #4978 (Phase 2 of the DCR story, #4976). This sub-issue covers the purely additive, leaf-package and config-surface work — no behavior change, no CRD changes.

**Depends on**: #4977 (pkg/oauthproto rename must land first).
**Followed by**: Sub-issue B (credential store + resolver).

---

### Step 2a — `FetchAuthorizationServerMetadata` (`pkg/oauthproto/discovery.go`)

Add to the existing `pkg/oauthproto/discovery.go`:
- `FetchAuthorizationServerMetadata(ctx context.Context, issuer string, client *http.Client) (*AuthorizationServerMetadata, error)` implementing a three-path RFC 8414 / OIDC Discovery fallback. Mirror the plumbing of `pkg/auth/discovery/discovery.go:745` (`FetchResourceMetadata`): bounded HTTP client (5–10 s), `Accept: application/json`, `io.LimitReader`, content-type check, issuer equality check per RFC 8414 §3.3.
- `ErrRegistrationEndpointMissing` sentinel returned when the winning response omits `registration_endpoint`.
- No `pkg/networking` import (leaf-package invariant).

New file `pkg/oauthproto/discovery_test.go`:
- Table-driven: one `httptest.NewServer` per discovery path, issuer mismatch error, missing `registration_endpoint` case.

### Step 2b — `DCRUpstreamConfig` struct (`pkg/authserver/config.go`)

Add adjacent to `OAuth2UpstreamRunConfig`:
```go
type DCRUpstreamConfig struct {
    DiscoveryURL             string `json:"discovery_url,omitempty"              yaml:"discovery_url,omitempty"`
    RegistrationEndpoint     string `json:"registration_endpoint,omitempty"      yaml:"registration_endpoint,omitempty"`
    InitialAccessTokenFile   string `json:"initial_access_token_file,omitempty"   yaml:"initial_access_token_file,omitempty"`
    InitialAccessTokenEnvVar string `json:"initial_access_token_env_var,omitempty" yaml:"initial_access_token_env_var,omitempty"`
    SoftwareID               string `json:"software_id,omitempty"               yaml:"software_id,omitempty"`
    SoftwareStatement        string `json:"software_statement,omitempty"        yaml:"software_statement,omitempty"`
}
```

Add `DCRConfig *DCRUpstreamConfig` field to `OAuth2UpstreamRunConfig` with doc comment.

### Step 2e — `OAuth2Config.Validate()` (`pkg/authserver/config.go`)

Extend the existing validator:
- Reject empty `ClientID` AND nil `DCRConfig` (both missing).
- Reject non-empty `ClientID` AND non-nil `DCRConfig` (both set).
- Reject `DCRConfig` with both `DiscoveryURL` and `RegistrationEndpoint` set.
- Reject `DCRConfig` with neither `DiscoveryURL` nor `RegistrationEndpoint` set.
- Do NOT touch `CommonOAuthConfig.Validate()`.

Extend `pkg/authserver/config_test.go` with table-driven cases for all four ClientID × DCRConfig combinations plus the exactly-one-of rule.

---

### Step 2a
- [ ] `pkg/oauthproto/discovery.go` exports `FetchAuthorizationServerMetadata(ctx context.Context, issuer string, client *http.Client) (*AuthorizationServerMetadata, error)` (`client == nil` builds a default bounded client).
- [ ] Three-path fallback tried in order, first 200-with-clean-JSON wins: (1) RFC 8414 path-insertion `{origin}/.well-known/oauth-authorization-server{path}`; (2) OIDC Discovery `{issuer}/.well-known/openid-configuration`; (3) bare RFC 8414 `{origin}/.well-known/oauth-authorization-server`.
- [ ] After decode, `metadata.issuer == issuer` enforced (RFC 8414 §3.3); mismatch returns a descriptive error.
- [ ] Bounded per-call timeout (5–10 s) regardless of caller's context deadline.
- [ ] `ErrRegistrationEndpointMissing` returned when `registration_endpoint` is absent from winning response.
- [ ] `go list -deps ./pkg/oauthproto/...` does not list `github.com/stacklok/toolhive/pkg/networking`.
- [ ] Unit test: each of the three discovery paths populates the returned metadata (one `httptest.NewServer` per path).
- [ ] Unit test: issuer mismatch returns the RFC 8414 §3.3 error.
- [ ] Unit test: missing `registration_endpoint` surfaces `ErrRegistrationEndpointMissing`.

### Step 2b
- [ ] `DCRUpstreamConfig` struct added to `pkg/authserver/config.go` with `DiscoveryURL`, `RegistrationEndpoint`, `InitialAccessTokenFile`, `InitialAccessTokenEnvVar`, `SoftwareID`, `SoftwareStatement` fields (JSON/YAML tags, with doc comments).
- [ ] `OAuth2UpstreamRunConfig` gains `DCRConfig *DCRUpstreamConfig` field with doc comment.

### Step 2e
- [ ] `OAuth2Config.Validate()` rejects empty `ClientID` AND nil `DCRConfig` (both missing).
- [ ] `OAuth2Config.Validate()` rejects non-empty `ClientID` AND non-nil `DCRConfig` (both set).
- [ ] `OAuth2Config.Validate()` rejects `DCRConfig` with both `DiscoveryURL` and `RegistrationEndpoint` set.
- [ ] `OAuth2Config.Validate()` rejects `DCRConfig` with neither `DiscoveryURL` nor `RegistrationEndpoint` set.
- [ ] `CommonOAuthConfig.Validate()` is unchanged (verified by diff).
- [ ] Unit tests: table-driven covering all four ClientID × DCRConfig combinations and the exactly-one-of rule.

### Cross-cutting
- [ ] All new `.go` files have SPDX 2-line header (`task license-fix` catches omissions).

---

## Patterns & References

- Template for `FetchAuthorizationServerMetadata`: `pkg/auth/discovery/discovery.go:745` (`FetchResourceMetadata`) — same timeout, LimitReader, content-type check, required-field validation pattern.
- RFC 8414 §3.1 path-insertion precedent: `pkg/auth/oauth/oidc.go:253` (`buildWellKnownURLs`) — the OAuth well-known URL is `{base}/.well-known/oauth-authorization-server/{tenant}`, not `{base}/{tenant}/.well-known/…`.
- Unit test HTTP mock style: `pkg/auth/oauth/dynamic_registration_test.go:1-80` (minimal per-request httptest pattern). Register teardown with `t.Cleanup(server.Close)`.
- Plain `testing` + `testify` (not Ginkgo) — matches existing unit tests in `pkg/oauthproto/` and `pkg/authserver/`.
- SPDX header required on all new files.

tgrunnagle:
## Summary

- The authserver currently requires every OAuth2 upstream to be pre-provisioned with a `ClientID`, which blocks operators who would prefer to register the client at runtime against the IdP. Phase 2 of the DCR story (#4978) introduces RFC 7591 Dynamic Client Registration; this sub-issue lays the leaf-package and config foundation needed by every later step.
- Adds `oauthproto.FetchAuthorizationServerMetadata` — a bounded, leaf-package RFC 8414 / OIDC Discovery client implementing the three-path fallback required to resolve `registration_endpoint` from a bare issuer URL.
- Adds `DCRUpstreamConfig` and a `DCRConfig *DCRUpstreamConfig` field on `OAuth2UpstreamRunConfig`, plus exactly-one-of validation rules (`ClientID` xor `DCRConfig`, and `DiscoveryURL` xor `RegistrationEndpoint`) so misconfiguration fails at startup rather than silently degrading at runtime.
- Purely additive — no behavior change, no CRD changes, no consumer wired up yet. Sub-issue B (credential store + resolver) consumes this surface area.

Closes #5037

## Type of change

- [x] New feature

## Test plan

New table-driven tests cover:

- All three RFC 8414 / OIDC Discovery fallback paths (one `httptest.NewServer` per path), with the bare-path test using a non-empty tenant so the third fallback is actually exercised.
- RFC 8414 §3.3 issuer-equality enforcement (mismatch surfaces a descriptive error).
- `ErrRegistrationEndpointMissing` returned with a non-nil partial metadata document when the winning response omits `registration_endpoint`.
- Issuer URL validation: empty, malformed, missing scheme/host, non-loopback `http://`, and `ftp://` schemes are rejected; loopback `http://` is allowed.
- Internal 10 s timeout overrides a `context.Background` caller (uses a package-private `discoveryTimeout` var so the test does not have to wait the full 10 s).
- All four `ClientID` × `DCRConfig` combinations on `OAuth2UpstreamRunConfig.Validate`, plus the `DiscoveryURL` xor `RegistrationEndpoint` rule on `DCRUpstreamConfig.Validate`.

## API Compatibility

## Changes

| File | Change |
|------|--------|
| `pkg/oauthproto/discovery.go` | Add `FetchAuthorizationServerMetadata` with three-path fallback, bounded timeout, issuer-equality check, and `ErrRegistrationEndpointMissing` partial-success contract. |
| `pkg/oauthproto/discovery_test.go` | Table-driven tests for all three discovery paths, issuer mismatch, missing registration endpoint, issuer validation, loopback carve-out, and timeout override. |
| `pkg/oauthproto/errors.go` | Add `ErrRegistrationEndpointMissing` sentinel. |
| `pkg/authserver/config.go` | Add `DCRUpstreamConfig` struct, `DCRConfig` field on `OAuth2UpstreamRunConfig`, and `Validate()` methods enforcing exactly-one-of rules. |
| `pkg/authserver/config_test.go` | Table-driven tests for the new validation rules. |

No — the new config fields are not yet consumed by any code path, and the discovery function is not yet called from the authserver. User-visible DCR behavior arrives in later sub-issues of #4978.

- **Rebase note**: this branch contains commit `3cb8de58` which is the `pkg/oauth` -> `pkg/oauthproto` rename from #4977 (PR #5036). #5036 is still open as of this PR's creation; once it merges to `main`, this branch should be rebased so only the three commits scoped to #5037 remain (`b408d6b2`, `a7de63d6`, `73985137`). The 5037-only diff touches just five files.
- **Partial-success error contract on `FetchAuthorizationServerMetadata`**: returns `(metadata, ErrRegistrationEndpointMissing)` rather than `(nil, err)` when the document is otherwise valid but lacks `registration_endpoint`, so callers that only need `TokenEndpoint` / `JWKSURI` can still use the result. This deviates from the usual Go "err != nil ⇒ value invalid" idiom; the doc comment includes a worked `errors.Is` example and the test asserts the partial document fields are populated.
- **Leaf-package invariant**: `pkg/oauthproto` must not import `pkg/networking`, so the per-call timeout, body cap (1 MB `io.LimitReader`), `Accept` header, and content-type check are implemented directly inside `discovery.go` rather than reused from `pkg/networking/utilities.go`. This duplication is intentional and is documented in the package doc.
- **`discoveryTimeout` is a `var`, not a `const`**: required so the timeout-override test can shorten it to keep the test suite fast. Production code MUST NOT mutate it; the comment on the declaration says so explicitly.
- **`buildDiscoveryURLs` deduplication**: when the issuer has no path component, the path-insertion URL and the bare URL are identical, so the function emits only the deduplicated set to avoid a redundant round-trip and a misleading "tried 3 URLs" error message.
- `CommonOAuthConfig.Validate()` is intentionally not modified — the new `ClientID`/`DCR` mutual-exclusion rules are specific to `OAuth2UpstreamRunConfig` and do not belong in the shared validator.

## Implementation plan

<details>

This PR implements Steps 2a, 2b, and 2e of the Phase 2 plan tracked in #4978:

- **Step 2a** — `FetchAuthorizationServerMetadata` in `pkg/oauthproto/discovery.go`. Three-path RFC 8414 / OIDC Discovery fallback (path-insertion, then OIDC Discovery, then bare RFC 8414), bounded 10 s per-call timeout, RFC 8414 §3.3 issuer equality check, `ErrRegistrationEndpointMissing` sentinel for partial success. Mirrors the plumbing of `pkg/auth/discovery/discovery.go:FetchResourceMetadata` (bounded HTTP client, `Accept: application/json`, `io.LimitReader`, content-type check) but stays in the leaf package — no `pkg/networking` import.
- **Step 2b** — `DCRUpstreamConfig` struct in `pkg/authserver/config.go` with `DiscoveryURL`, `RegistrationEndpoint`, `InitialAccessTokenFile`, `InitialAccessTokenEnvVar`, `SoftwareID`, `SoftwareStatement` fields (JSON + YAML tags). Adds `DCRConfig *DCRUpstreamConfig` field to `OAuth2UpstreamRunConfig`.
- **Step 2e** — Extends `OAuth2UpstreamRunConfig.Validate()` and adds `DCRUpstreamConfig.Validate()` to reject all three invalid combinations: both missing, both set, and `DiscoveryURL`/`RegistrationEndpoint` both set or both missing. `CommonOAuthConfig.Validate()` is left alone.

Steps 2c and 2d (credential store + resolver) are deferred to sub-issue B per #4978.

</details>

## Large PR Justification

- Majority lines are generated docs and tests

### DIFF
--- a/pkg/authserver/config.go
+++ b/pkg/authserver/config.go
@@ -211,6 +211,8 @@ type OAuth2UpstreamRunConfig struct {
 	TokenEndpoint string `json:"token_endpoint" yaml:"token_endpoint"`
 
 	// ClientID is the OAuth 2.0 client identifier registered with the upstream IDP.
+	// Mutually exclusive with DCRConfig: when DCRConfig is set, ClientID is obtained
+	// at runtime via RFC 7591 Dynamic Client Registration and must be left empty.
 	ClientID string `json:"client_id" yaml:"client_id"`
 
 	// ClientSecretFile is the path to a file containing the OAuth 2.0 client secret.
@@ -246,6 +248,56 @@ type OAuth2UpstreamRunConfig struct {
 	// Google's access_type=offline.
 	//nolint:lll // field tags require full JSON+YAML names
 	AdditionalAuthorizationParams map[string]string `json:"additional_authorization_params,omitempty" yaml:"additional_authorization_params,omitempty"`
+
+	// DCRConfig enables RFC 7591 Dynamic Client Registration against the
+	// upstream authorization server. When set, the client credentials are
+	// obtained at runtime rather than being pre-provisioned via ClientID /
+	// ClientSecretFile / ClientSecretEnvVar, and ClientID must be left empty.
+	// Mutually exclusive with ClientID.
+	DCRConfig *DCRUpstreamConfig `json:"dcr_config,omitempty" yaml:"dcr_config,omitempty"`
+}
+
+// DCRUpstreamConfig configures RFC 7591 Dynamic Client Registration for an
+// upstream authorization server. When present on an OAuth2 upstream, the
+// authserver performs registration at runtime to obtain client credentials,
+// replacing the need to pre-provision a ClientID.
+//
+// Exactly one of DiscoveryURL or RegistrationEndpoint must be set. DiscoveryURL
+// points at RFC 8414 / OIDC Discovery metadata from which the registration
+// endpoint is resolved; RegistrationEndpoint is used directly when the upstream
+// does not publish discovery metadata.
+type DCRUpstreamConfig struct {
+	// DiscoveryURL is the RFC 8414 / OIDC Discovery URL from which the
+	// registration_endpoint is resolved at runtime. Mutually exclusive with
+	// RegistrationEndpoint.
+	DiscoveryURL string `json:"discovery_url,omitempty" yaml:"discovery_url,omitempty"`
+
+	// RegistrationEndpoint is the RFC 7591 registration endpoint URL used
+	// directly, bypassing discovery. Mutually exclusive with DiscoveryURL.
+	RegistrationEndpoint string `json:"registration_endpoint,omitempty" yaml:"registration_endpoint,omitempty"`
+
+	// InitialAccessTokenFile is the path to a file containing the RFC 7591
+	// initial access token presented to the registration endpoint. Mutually
+	// exclusive with InitialAccessTokenEnvVar. Both may be omitted for open
+	// registration endpoints.
+	//nolint:lll // field tags require full JSON+YAML names
+	InitialAccessTokenFile string `json:"initial_access_token_file,omitempty" yaml:"initial_access_token_file,omitempty"`
+
+	// InitialAccessTokenEnvVar is the name of an environment variable
+	// containing the RFC 7591 initial access token. Mutually exclusive with
+	// InitialAccessTokenFile.
+	//nolint:lll // field tags require full JSON+YAML names
+	InitialAccessTokenEnvVar string `json:"initial_access_token_env_var,omitempty" yaml:"initial_access_token_env_var,omitempty"`
+
+	// SoftwareID is the RFC 7591 "software_id" registration metadata value,
+	// identifying the client software independent of any particular
+	// registration instance.
+	SoftwareID string `json:"software_id,omitempty" yaml:"software_id,omitempty"`
+
+	// SoftwareStatement is the RFC 7591 "software_statement" JWT asserting
+	// metadata about the client software, signed by a party the authorization
+	// server trusts.
+	SoftwareStatement string `json:"software_statement,omitempty" yaml:"software_statement,omitempty"`
 }
 
 // TokenResponseMappingRunConfig maps non-standard token response fields to standard fields.
@@ -596,5 +648,53 @@ func validateIssuerURL(issuer string) error {
 		return fmt.Errorf("must not have trailing slash")
 	}
 
+	return nil
+}
+
+// Validate checks that the OAuth2UpstreamRunConfig is internally consistent.
+// It enforces the mutual exclusivity of ClientID and DCRConfig: exactly one must
+// be set. A ClientID is required for pre-provisioned clients; a DCRConfig is
+// required when client credentials are obtained at runtime via RFC 7591
+// Dynamic Client Registration. When DCRConfig is present, its own validity is
+// also checked via DCRUpstreamConfig.Validate.
+//
+// Validate intentionally does not verify fields handled by the shared
+// CommonOAuthConfig or upstream.OAuth2Config validators — it only covers the
+// run-config surface area unique to OAuth2UpstreamRunConfig.
+func (c *OAuth2UpstreamRunConfig) Validate() error {
+	hasClientID := c.ClientID != ""
+	hasDCR := c.DCRConfig != nil
+	switch {
+	case !hasClientID && !hasDCR:
+		return fmt.Errorf("oauth2 upstream: either client_id or dcr_config is required")
+	case hasClientID && hasDCR:
+		return fmt.Errorf("oauth2 upstream: client_id and dcr_config are mutually exclusive")
+	}
+
+	if hasDCR {
+		if err := c.DCRConfig.Validate(); err != nil {
+			return fmt.Errorf("oauth2 upstream: invalid dcr_config: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// Validate checks that the DCRUpstreamConfig specifies exactly one of
+// DiscoveryURL or RegistrationEndpoint.
+//
+// DiscoveryURL triggers runtime resolution of the registration endpoint via
+// RFC 8414 / OIDC Discovery; RegistrationEndpoint bypasses discovery for
+// providers that do not publish metadata. Requiring exactly one prevents
+// ambiguity about which URL the authserver should contact for registration.
+func (c *DCRUpstreamConfig) Validate() error {
+	hasDiscovery := c.DiscoveryURL != ""
+	hasRegistration := c.RegistrationEndpoint != ""
+	switch {
+	case !hasDiscovery && !hasRegistration:
+		return fmt.Errorf("dcr_config: either discovery_url or registration_endpoint is required")
+	case hasDiscovery && hasRegistration:
+		return fmt.Errorf("dcr_config: discovery_url and registration_endpoint are mutually exclusive")
+	}
 	return nil
 }
--- a/pkg/authserver/config.go
+++ b/pkg/authserver/config.go
@@ -479,6 +479,87 @@ func (c *Config) Validate() error {
 	return nil
 }
 
+// Validate checks that the OAuth2UpstreamRunConfig is internally consistent.
+// It enforces the mutual exclusivity of ClientID and DCRConfig: exactly one must
+// be set. A ClientID is required for pre-provisioned clients; a DCRConfig is
+// required when client credentials are obtained at runtime via RFC 7591
+// Dynamic Client Registration. When DCRConfig is present, its own validity is
+// also checked via DCRUpstreamConfig.Validate.
+//
+// Validate intentionally does not verify fields handled by the shared
+// CommonOAuthConfig or upstream.OAuth2Config validators — it only covers the
+// run-config surface area unique to OAuth2UpstreamRunConfig.
+//
+// Called from buildPureOAuth2Config at the RunConfig → upstream.OAuth2Config
+// conversion boundary so that DCR-specific fields are validated before they
+// are dropped during conversion.
+func (c *OAuth2UpstreamRunConfig) Validate() error {
+	hasClientID := c.ClientID != ""
+	hasDCR := c.DCRConfig != nil
+	switch {
+	case !hasClientID && !hasDCR:
+		return fmt.Errorf("oauth2 upstream: either client_id or dcr_config is required")
+	case hasClientID && hasDCR:
+		return fmt.Errorf("oauth2 upstream: client_id and dcr_config are mutually exclusive")
+	}
+
+	if hasDCR {
+		if err := c.DCRConfig.Validate(); err != nil {
+			return fmt.Errorf("oauth2 upstream: invalid dcr_config: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// Validate checks that the DCRUpstreamConfig specifies exactly one of
+// DiscoveryURL or RegistrationEndpoint, that the configured URL is well-formed
+// and uses HTTPS (or http on a loopback host for local development), and that
+// the two initial-access-token sources (InitialAccessTokenFile and
+// InitialAccessTokenEnvVar) are not both set.
+//
+// DiscoveryURL triggers runtime resolution of the registration endpoint via
+// RFC 8414 / OIDC Discovery; RegistrationEndpoint bypasses discovery for
+// providers that do not publish metadata. Requiring exactly one prevents
+// ambiguity about which URL the authserver should contact for registration.
+//
+// URL well-formedness and HTTPS are enforced here at the schema-validation
+// boundary so misconfiguration fails fast at startup rather than at first DCR
+// attempt; the runtime callers (pkg/oauthproto/discovery.go and
+// pkg/oauthproto/dcr.go) defend in depth, but this is the natural fail-fast
+// point.
+//
+// Rejecting a config that supplies both an InitialAccessTokenFile and an
+// InitialAccessTokenEnvVar prevents a credential-rotation footgun: if both
+// were accepted, an operator updating the env-var value would not realize
+// the file source still wins (or vice versa) and would silently keep
+// presenting a stale token at registration.
+func (c *DCRUpstreamConfig) Validate() error {
+	hasDiscovery := c.DiscoveryURL != ""
+	hasRegistration := c.RegistrationEndpoint != ""
+	switch {
+	case !hasDiscovery && !hasRegistration:
+		return fmt.Errorf("dcr_config: either discovery_url or registration_endpoint is required")
+	case hasDiscovery && hasRegistration:
+		return fmt.Errorf("dcr_config: discovery_url and registration_endpoint are mutually exclusive")
+	case hasDiscovery:
+		if err := networking.ValidateEndpointURL(c.DiscoveryURL); err != nil {
+			return fmt.Errorf("dcr_config: invalid discovery_url: %w", err)
+		}
+	case hasRegistration:
+		if err := networking.ValidateEndpointURL(c.RegistrationEndpoint); err != nil {
+			return fmt.Errorf("dcr_config: invalid registration_endpoint: %w", err)
+		}
+	}
+
+	if c.InitialAccessTokenFile != "" && c.InitialAccessTokenEnvVar != "" {
+		return fmt.Errorf(
+			"dcr_config: initial_access_token_file and initial_access_token_env_var are mutually exclusive")
+	}
+
+	return nil
+}
+
 // validateUpstreams validates the upstream configurations.
 func (c *Config) validateUpstreams() error {
 	if len(c.Upstreams) == 0 {
@@ -648,53 +729,5 @@ func validateIssuerURL(issuer string) error {
 		return fmt.Errorf("must not have trailing slash")
 	}
 
-	return nil
-}
-
-// Validate checks that the OAuth2UpstreamRunConfig is internally consistent.
-// It enforces the mutual exclusivity of ClientID and DCRConfig: exactly one must
-// be set. A ClientID is required for pre-provisioned clients; a DCRConfig is
-// required when client credentials are obtained at runtime via RFC 7591
-// Dynamic Client Registration. When DCRConfig is present, its own validity is
-// also checked via DCRUpstreamConfig.Validate.
-//
-// Validate intentionally does not verify fields handled by the shared
-// CommonOAuthConfig or upstream.OAuth2Config validators — it only covers the
-// run-config surface area unique to OAuth2UpstreamRunConfig.
-func (c *OAuth2UpstreamRunConfig) Validate() error {
-	hasClientID := c.ClientID != ""
-	hasDCR := c.DCRConfig != nil
-	switch {
-	case !hasClientID && !hasDCR:
-		return fmt.Errorf("oauth2 upstream: either client_id or dcr_config is required")
-	case hasClientID && hasDCR:
-		return fmt.Errorf("oauth2 upstream: client_id and dcr_config are mutually exclusive")
-	}
-
-	if hasDCR {
-		if err := c.DCRConfig.Validate(); err != nil {
-			return fmt.Errorf("oauth2 upstream: invalid dcr_config: %w", err)
-		}
-	}
-
-	return nil
-}
-
-// Validate checks that the DCRUpstreamConfig specifies exactly one of
-// DiscoveryURL or RegistrationEndpoint.
-//
-// DiscoveryURL triggers runtime resolution of the registration endpoint via
-// RFC 8414 / OIDC Discovery; RegistrationEndpoint bypasses discovery for
-// providers that do not publish metadata. Requiring exactly one prevents
-// ambiguity about which URL the authserver should contact for registration.
-func (c *DCRUpstreamConfig) Validate() error {
-	hasDiscovery := c.DiscoveryURL != ""
-	hasRegistration := c.RegistrationEndpoint != ""
-	switch {
-	case !hasDiscovery && !hasRegistration:
-		return fmt.Errorf("dcr_config: either discovery_url or registration_endpoint is required")
-	case hasDiscovery && hasRegistration:
-		return fmt.Errorf("dcr_config: discovery_url and registration_endpoint are mutually exclusive")
-	}
 	return nil
 }
--- a/pkg/oauthproto/discovery.go
+++ b/pkg/oauthproto/discovery.go
@@ -6,6 +6,7 @@ package oauthproto
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -32,11 +33,18 @@ const discoveryMaxResponseSize = 1024 * 1024 // 1MB
 // metadata for the given issuer, using a three-path fallback that mirrors
 // the guidance in RFC 8414 §3.1 and OpenID Connect Discovery 1.0.
 //
-// The paths are tried in this order; the first 200 OK with a clean
-// application/json response that passes the RFC 8414 §3.3 issuer check wins:
+// The paths are tried in this order:
 //  1. RFC 8414 path-insertion: {origin}/.well-known/oauth-authorization-server{path}
 //  2. OIDC Discovery:          {issuer}/.well-known/openid-configuration
 //  3. Bare RFC 8414:           {origin}/.well-known/oauth-authorization-server
+//
+// A 200 OK response with a clean application/json body that passes the
+// RFC 8414 §3.3 issuer check is a candidate. The first candidate with a
+// non-empty registration_endpoint wins immediately. If no candidate has a
+// registration_endpoint but at least one was otherwise valid, the first such
+// partial document is returned with ErrRegistrationEndpointMissing — never
+// short-circuiting on a partial doc, since a tenant-aware IdP may publish
+// path-insertion metadata without DCR while the OIDC document advertises it.
 //
 // If client is nil, a default *http.Client with a bounded 10 s timeout is
 // used. When client is non-nil, the same bounded 10 s per-call timeout is
@@ -48,12 +56,13 @@ const discoveryMaxResponseSize = 1024 * 1024 // 1MB
 //   - On full success, returns (metadata, nil) with a non-empty
 //     RegistrationEndpoint.
 //
-//   - When the winning document is otherwise valid but omits
-//     registration_endpoint, returns (metadata, ErrRegistrationEndpointMissing).
-//     The metadata is non-nil so callers can reuse the other fields
-//     (TokenEndpoint, JWKSURI, etc.). Note this deviates from the usual Go
-//     "err != nil ⇒ value is invalid" idiom; callers must check with
-//     errors.Is and explicitly decide whether to use the partial doc:
+//   - When at least one candidate document was issuer-validated but every such
+//     document omits registration_endpoint, returns the first partial document
+//     paired with ErrRegistrationEndpointMissing. The metadata is non-nil so
+//     callers can reuse the other fields (TokenEndpoint, JWKSURI, etc.). Note
+//     this deviates from the usual Go "err != nil ⇒ value is invalid" idiom;
+//     callers must check with errors.Is and explicitly decide whether to use
+//     the partial doc:
 //
 //     md, err := FetchAuthorizationServerMetadata(ctx, issuer, nil)
 //     switch {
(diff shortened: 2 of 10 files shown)


pkg/authserver/config.go
```diff
@@ -598,3 +650,51 @@ func validateIssuerURL(issuer string) error {
...
+// It enforces the mutual exclusivity of ClientID and DCRConfig: exactly one must
+// be set. A ClientID is required for pre-provisioned clients; a DCRConfig is
+// required when client credentials are obtained at runtime via RFC 7591
+// Dynamic Client Registration. When DCRConfig is present, its own validity is
+// also checked via DCRUpstreamConfig.Validate.
+//
+// Validate intentionally does not verify fields handled by the shared
+// CommonOAuthConfig or upstream.OAuth2Config validators — it only covers the
+// run-config surface area unique to OAuth2UpstreamRunConfig.
+func (c *OAuth2UpstreamRunConfig) Validate() error {
```
tgrunnagle: **[HIGH] OAuth2UpstreamRunConfig.Validate() has no caller in the tree (consensus: 9/10)**

Verified by grepping `pkg/authserver`: this method is never invoked. The runtime validation path is `Config.Validate() -> validateUpstreams() -> validateUpstreamType()` (line 542), which calls `up.OAuth2Config.Validate()` on the *parallel* `pkg/authserver/upstream.OAuth2Config` struct — not on `OAuth2UpstreamRunConfig`. So today, every rule introduced here (`ClientID` xor `DCRConfig`, propagation to `DCRUpstreamConfig.Validate`) is dead code.

The PR description says mutual exclusion is "enforced". That is technically false until either (a) `OAuth2UpstreamRunConfig.Validate()` is wired into a `RunConfig.Validate()` method called at the load boundary, or (b) the conversion in `runner/embeddedauthserver.go:buildPureOAuth2Config` calls it before the existing validator. Adding a one-line caller in this PR would make the description accurate. If wiring is genuinely deferred to sub-issue B, link the issue from this method's doc comment so future readers don't assume it runs.

*Raised by: config-and-api-design.*

pkg/authserver/config.go
```diff
@@ -598,3 +650,51 @@ func validateIssuerURL(issuer string) error {
...
+	}
+
+	if hasDCR {
+		if err := c.DCRConfig.Validate(); err != nil {
+			return fmt.Errorf("oauth2 upstream: invalid dcr_config: %w", err)
+		}
+	}
+
+	return nil
+}
```
tgrunnagle: **[LOW] Public Validate methods placed below private helpers (consensus: 7/10)**

`.claude/rules/go-style.md` mandates "Public methods in the top half of files, private methods in the bottom half." `OAuth2UpstreamRunConfig.Validate` and `DCRUpstreamConfig.Validate` (this method ends here at line 681; the second runs through line 700) are placed after the private `validateUpstreams`, `validateUpstreamName`, `validateUpstreamType`, `applyDefaults`, and `validateIssuerURL`. The pre-existing public `Config.Validate` is correctly placed at line 442.

Move these two methods adjacent to `Config.Validate` so the three public validators appear together near the top of the file.

*Raised by: go-style.*

pkg/authserver/config.go
```diff
@@ -598,3 +650,51 @@ func validateIssuerURL(issuer string) error {
...
+
+// Validate checks that the DCRUpstreamConfig specifies exactly one of
+// DiscoveryURL or RegistrationEndpoint.
+//
+// DiscoveryURL triggers runtime resolution of the registration endpoint via
+// RFC 8414 / OIDC Discovery; RegistrationEndpoint bypasses discovery for
+// providers that do not publish metadata. Requiring exactly one prevents
+// ambiguity about which URL the authserver should contact for registration.
+func (c *DCRUpstreamConfig) Validate() error {
```
tgrunnagle: **[MEDIUM] InitialAccessToken{File,EnvVar} doc says "Mutually exclusive" but Validate does not enforce (consensus: 9/10 — three independent agents)**

The doc comments at lines 281 and 288 both state the two fields are mutually exclusive. `DCRUpstreamConfig.Validate()` only enforces the `DiscoveryURL`/`RegistrationEndpoint` xor and the test case at `config_test.go:354` ("optional metadata is ignored by validate") explicitly sets BOTH and asserts no error. So the docs and the validator (and the tests) disagree.

At runtime this is a footgun: whichever resolver wins in sub-issue B will silently win, and an operator rotating the token through the env var will see stale file-based tokens used instead, with no error at startup. This violates `.claude/rules/go-style.md` "Constructor Validation: Fail Loudly on Invalid Input."

Fix one of:
1. Add the xor check to `Validate()` and update the test to expect an error when both are set.
2. Remove the "Mutually exclusive" wording and document precedence (`InitialAccessTokenFile` wins, say) on both fields.

Option 1 is preferred — credential ambiguity is a security smell.

*Raised by: security, test-quality, config-and-api-design.*

pkg/authserver/config.go
```diff
@@ -598,3 +650,51 @@ func validateIssuerURL(issuer string) error {
...
+	hasDiscovery := c.DiscoveryURL != ""
+	hasRegistration := c.RegistrationEndpoint != ""
+	switch {
+	case !hasDiscovery && !hasRegistration:
+		return fmt.Errorf("dcr_config: either discovery_url or registration_endpoint is required")
+	case hasDiscovery && hasRegistration:
+		return fmt.Errorf("dcr_config: discovery_url and registration_endpoint are mutually exclusive")
```
tgrunnagle: **[MEDIUM] DCRUpstreamConfig.Validate does not enforce HTTPS or URL well-formedness (consensus: 7/10)**

Validate only checks the xor between `DiscoveryURL` and `RegistrationEndpoint`. There is no `url.Parse` check, no scheme check, and no loopback-aware HTTPS enforcement, while the surrounding code consistently fails-fast on these:
- `validateIssuerURL` (config.go:610-ish) enforces https-or-loopback
- `buildDiscoveryURLs` (this PR, `pkg/oauthproto/discovery.go:120-127`) rejects empty, malformed, missing scheme/host, and non-https-non-loopback
- `validateRegistrationEndpoint` in `pkg/oauthproto/dcr.go:194` enforces HTTPS

Defense-at-use-site exists at runtime, but the schema-level Validate is the natural fail-fast point. A config like `DiscoveryURL: "://broken"` or `DiscoveryURL: "http://idp.example.com"` will pass Validate (once it actually runs — see finding \#1) and fail much later with a less obvious error.

A five-line addition matching the patterns of `validateIssuerURL` would make Validate authoritative.

*Raised by: config-and-api-design.*

pkg/authserver/config.go
```diff
@@ -246,6 +248,56 @@ type OAuth2UpstreamRunConfig struct {
 	// Google's access_type=offline.
 	//nolint:lll // field tags require full JSON+YAML names
 	AdditionalAuthorizationParams map[string]string `json:"additional_authorization_params,omitempty" yaml:"additional_authorization_params,omitempty"`
+
+	// DCRConfig enables RFC 7591 Dynamic Client Registration against the
+	// upstream authorization server. When set, the client credentials are
+	// obtained at runtime rather than being pre-provisioned via ClientID /
+	// ClientSecretFile / ClientSecretEnvVar, and ClientID must be left empty.
+	// Mutually exclusive with ClientID.
+	DCRConfig *DCRUpstreamConfig `json:"dcr_config,omitempty" yaml:"dcr_config,omitempty"`
```
tgrunnagle, line 257: **[MEDIUM] DCRConfig field widens drift between two parallel OAuth2 config structs (consensus: 7/10)**

`OAuth2UpstreamRunConfig` (this struct, run-config) and `pkg/authserver/upstream.OAuth2Config` (runtime) are now confirmed parallel: both hold ClientID, ClientSecret*, RedirectURI, Scopes, AuthorizationEndpoint, TokenEndpoint, UserInfo, TokenResponseMapping, AdditionalAuthorizationParams. The Swagger description for `OAuth2UpstreamRunConfig` literally says "OAuth2Config contains OAuth 2.0-specific configuration" — naming evidence of the same confusion. Adding `DCRConfig` only on the run-config side widens the surface gap.

`.claude/rules/go-style.md` ("Avoid parallel types that drift") explicitly warns against this pattern. Out of scope to consolidate in this PR, but the divergence is growing — please file a follow-up issue (linked from this struct's doc comment) noting that `DCRConfig` is run-config-only and that future work should propagate it to `upstream.OAuth2Config` (or remove the parallel altogether).

*Raised by: config-and-api-design.*

pkg/authserver/config.go
```diff
@@ -246,6 +248,56 @@ type OAuth2UpstreamRunConfig struct {
...
+
+// DCRUpstreamConfig configures RFC 7591 Dynamic Client Registration for an
+// upstream authorization server. When present on an OAuth2 upstream, the
+// authserver performs registration at runtime to obtain client credentials,
+// replacing the need to pre-provision a ClientID.
+//
+// Exactly one of DiscoveryURL or RegistrationEndpoint must be set. DiscoveryURL
+// points at RFC 8414 / OIDC Discovery metadata from which the registration
+// endpoint is resolved; RegistrationEndpoint is used directly when the upstream
+// does not publish discovery metadata.
```
tgrunnagle, line 268: **[HIGH/note] DCRConfig is silently dropped by buildPureOAuth2Config conversion (consensus: 7/10)**

`runner/embeddedauthserver.go:buildPureOAuth2Config` converts `OAuth2UpstreamRunConfig` to `upstream.OAuth2Config` field-by-field but never reads `rc.OAuth2Config.DCRConfig`. So a config that sets `DCRConfig` and leaves `ClientID` empty will (a) pass any future `RunConfig.Validate()` once wired, then (b) be converted to `upstream.OAuth2Config{ClientID: ""}`, then (c) fail at `upstream.OAuth2Config.Validate()` with `"client_id is required"` — the very error the new mutual-exclusion rule promises to replace.

The PR is documented as purely additive groundwork, so deferring the conversion update is reasonable, but "DCR-only configs parse but cannot run" deserves an explicit note on this struct's doc comment (or on `OAuth2UpstreamRunConfig`) referencing the sub-issue that wires it up. Otherwise a reader who finds this struct via godoc will assume DCR works end-to-end.

*Raised by: config-and-api-design.*

pkg/oauthproto/discovery.go
```diff
@@ -1,18 +1,234 @@
...
+	defer cancel()
+
+	var attemptErrs []string
+	for _, u := range urls {
+		metadata, err := fetchDiscoveryDocument(fetchCtx, httpClient, u, issuer)
+		if err != nil {
+			attemptErrs = append(attemptErrs, fmt.Sprintf("%s: %v", u, err))
+			continue
+		}
+		if metadata.RegistrationEndpoint == "" {
```
tgrunnagle, line 105: **[HIGH/discuss] Early return on missing registration_endpoint short-circuits remaining fallback URLs (consensus: 7/10)**

When URL #1 (path-insertion) returns 200 OK + valid issuer + missing `registration_endpoint`, this branch returns `(metadata, ErrRegistrationEndpointMissing)` without trying URL #2 (OIDC) or URL #3 (bare RFC 8414). For real-world IdPs (Keycloak, Auth0, Okta) where a tenant-aware deployment's path-insertion document might not advertise DCR but the OIDC Discovery document does, this silently loses DCR support.

The doc-comment says "the first 200 OK with a clean application/json response that passes the RFC 8414 §3.3 issuer check wins" — which is faithful to the current code, but RFC 8414 §3 implicitly contemplates trying alternate locations until one succeeds. "Succeeded for issuer match but missing the field DCR callers actually want" is not the same as "winner found."

Consider: continue iterating when a candidate is otherwise valid but missing `registration_endpoint`; track the first such partial-success metadata as a fallback; only return `(partial, ErrRegistrationEndpointMissing)` if every URL fails to produce a registration endpoint. Update the doc comment to match. Add a regression test where path-insertion serves partial doc and OIDC serves the full doc — current code returns partial, ideal code returns full.

Flagging for human adjudication; not a mechanical fix.

*Raised by: security/oauth-compliance.*

pkg/oauthproto/discovery.go
```diff
@@ -1,18 +1,234 @@
...
+
+	// Bound per-call timeout regardless of caller context.
+	fetchCtx, cancel := context.WithTimeout(ctx, discoveryTimeout)
+	defer cancel()
+
+	var attemptErrs []string
+	for _, u := range urls {
+		metadata, err := fetchDiscoveryDocument(fetchCtx, httpClient, u, issuer)
+		if err != nil {
+			attemptErrs = append(attemptErrs, fmt.Sprintf("%s: %v", u, err))
```
tgrunnagle: **[MEDIUM] Aggregated attempt errors lose wrap chain (consensus: 8/10)**

Per-attempt errors are flattened with `fmt.Sprintf("%s: %v", u, err)` and joined into a single string at line 101. The inner errors lose their wrap chain — callers cannot use `errors.Is` / `errors.As` on the returned error to discover, e.g., `context.DeadlineExceeded` or transport-specific sentinels. `.claude/rules/go-style.md` says "Use `fmt.Errorf` with `%w` to preserve error chains."

For multi-attempt failures, the canonical Go ≥1.20 form is `errors.Join`:

```go
var attemptErrs []error
for _, u := range urls {
    metadata, err := fetchDiscoveryDocument(...)
    if err != nil {
        attemptErrs = append(attemptErrs, fmt.Errorf("%s: %w", u, err))
        continue
    }
    ...
}
return nil, fmt.Errorf(
    "failed to discover authorization server metadata for issuer %q: %w",
    issuer, errors.Join(attemptErrs...),
)
```

The `errors` package is not currently imported by this file — add it.

*Raised by: go-style, general-code-quality.*

pkg/oauthproto/discovery.go
```diff
@@ -1,18 +1,234 @@
...
+	}
+	if parsed.Scheme == "" || parsed.Host == "" {
+		return nil, fmt.Errorf("invalid issuer URL: scheme and host are required")
+	}
+	if parsed.Scheme != "https" && !IsLoopbackHost(parsed.Host) {
+		return nil, fmt.Errorf("issuer must use https (got %q)", parsed.Scheme)
+	}
+
+	// Strip trailing slash from issuer path so URL joins stay predictable.
+	tenant := strings.Trim(parsed.EscapedPath(), "/")
```
tgrunnagle: **[MEDIUM] path.Join on EscapedPath risks double-encoding tenants with special chars (consensus: 8/10)**

`tenant := strings.Trim(parsed.EscapedPath(), "/")` returns the *escaped* path. Below, `pathInsertion.Path = path.Join(WellKnownOAuthServerPath, tenant)` and `oidc.Path = path.Join("/", tenant, WellKnownOIDCPath)` assign these into `url.URL.Path`, which is the *unescaped* form. When `pathInsertion.String()` runs, it re-escapes `Path`, so percent-encoded characters in the tenant get double-encoded.

Concrete: an issuer `https://idp/tenants/a%2Fb` (where `a%2Fb` is a literal tenant containing `/`) yields `tenant = "tenants/a%2Fb"`, and the final URL becomes `.../oauth-authorization-server/tenants/a%252Fb` — `%` is re-encoded to `%25`.

Fix: use `parsed.Path` (unescaped) instead of `parsed.EscapedPath()`. Then `String()` performs a single round of escaping correctly. Add a regression test with a tenant containing characters `url.PathEscape` actually escapes (spaces, unicode); current tests only cover `/tenants/acme` which is escape-identity.

```suggestion
	tenant := strings.Trim(parsed.Path, "/")
```

*Raised by: security/oauth-compliance, go-style.*

tgrunnagle (commented):
**Agents consulted**: security/oauth-compliance, go-style, test-quality, config-and-api-design, general-code-quality

### Consensus Summary
| 1 | `OAuth2UpstreamRunConfig.Validate()` is orphaned (no caller) | 9/10 | HIGH | Wire it in or document |
| 2 | `InitialAccessToken{File,EnvVar}` doc says "mutually exclusive" but `Validate()` does not enforce | 9/10 | MEDIUM | Fix |
| 3 | `path.Join(WellKnownOAuthServerPath, parsed.EscapedPath())` risks double-encoding tenants with special chars | 8/10 | MEDIUM | Fix |
| 4 | Aggregated attempt errors flatten to string and lose `errors.Is`/`errors.As` reachability | 8/10 | MEDIUM | Fix (use `errors.Join`) |
| 5 | Early return `(metadata, ErrRegistrationEndpointMissing)` from URL #1 short-circuits URLs #2/#3 | 7/10 | HIGH | Discuss |
| 6 | No test asserts `buildDiscoveryURLs` deduplication for tenant-less issuers | 7/10 | MEDIUM | Add test |
| 7 | `DCRUpstreamConfig.Validate()` does not enforce HTTPS or URL well-formedness | 7/10 | MEDIUM | Fix |
| 8 | Two parallel OAuth2 config structs (`OAuth2UpstreamRunConfig` vs `upstream.OAuth2Config`) widening drift | 7/10 | MEDIUM | Note |
| 9 | `DCRConfig` is silently dropped by `buildPureOAuth2Config` conversion (DCR-only configs cannot reach runtime) | 7/10 | HIGH | Document or wire |
| 10 | Public `Validate` methods placed at bottom of `config.go`, after private helpers | 7/10 | LOW | Move |

### Overall

This is a well-executed groundwork PR. Commits are tight, doc comments are thorough, the SPDX header migration is correct, the leaf-package invariant for `pkg/oauthproto` is preserved, and the partial-success error contract on `FetchAuthorizationServerMetadata` is conscientiously documented. Test coverage exercises all three RFC 8414 / OIDC fallback paths, the loopback HTTP carve-out, the issuer-mismatch enforcement, and the four-quadrant config validation matrix.

The findings cluster around three themes. **Validation completeness**: the new `Validate()` methods are honest about their scope but two real gaps slipped through — the `InitialAccessToken{File,EnvVar}` xor that the doc comments promise but the validator never checks (with a test that codifies the contradiction), and the `OAuth2UpstreamRunConfig.Validate()` having no caller in the tree today (the runtime path is `validateUpstreamType -> upstream.OAuth2Config.Validate`, a parallel struct that doesn't know about DCR). Both are addressable in this PR or with a clearly-scoped follow-up commit; neither is a blocker on its own. **URL handling subtleties** in `discovery.go`: `path.Join` on `EscapedPath()` will double-escape tenants containing `%` characters, and the aggregated multi-attempt error drops the wrap chain rather than using `errors.Join`. Both are fixable in a few lines. **Design questions**: the early-return on missing `registration_endpoint` is documented but arguably wrong — RFC 8414 fallback semantics suggest continuing through remaining URLs and only returning partial success if all attempts fail. This is the one item that warrants explicit discussion rather than a mechanical fix.

The scope is honest about deferrals (sub-issues B onward) but the PR description's claim that mutual-exclusion is "enforced" is technically false today since nothing calls the new method. Worth either tightening the wording or adding a one-line wiring change so the validator actually runs.

Nothing here blocks the foundational direction; the surface area being added is the right shape for what sub-issue B will need.

---

pkg/oauthproto/discovery_test.go
```diff
@@ -125,3 +123,259 @@ func TestOIDCDiscoveryDocument_SupportsGrantType(t *testing.T) {
...
+
+// writeJSONMetadata serves an AuthorizationServerMetadata document as JSON.
+// Silently swallows encoding errors: the caller is an httptest server handler,
+// which has no way to surface an error back to the test.
+func writeJSONMetadata(w http.ResponseWriter, md AuthorizationServerMetadata) {
+	w.Header().Set("Content-Type", "application/json")
+	_ = json.NewEncoder(w).Encode(md)
+}
+
+func TestFetchAuthorizationServerMetadata(t *testing.T) {
```
tgrunnagle, line 136: **[MEDIUM] No test asserts buildDiscoveryURLs deduplication for tenant-less issuer (consensus: 7/10)**

The PR notes call out that `buildDiscoveryURLs` deduplicates the path-insertion and bare RFC 8414 URLs when the issuer has no tenant path, dropping the tried-URL count from 3 to 2. This is a behavior change but no test asserts it. `TestFetchAuthorizationServerMetadata_AllowsLoopbackHTTP` (line 255) uses a tenant-less issuer but its handler answers every path, so a regression that re-introduced the duplicate (or, conversely, removed the dedup) is silently invisible.

Add a focused test (e.g., `TestFetchAuthorizationServerMetadata_DedupesPathInsertionAndBare`) that:
1. Uses a tenant-less issuer
2. Records `r.URL.Path` on every request
3. Returns 404 for all paths
4. Asserts the recorded paths contain exactly two distinct values: `/.well-known/oauth-authorization-server` and `/.well-known/openid-configuration`

Bonus: this also locks in the documented priority order (path-insertion before OIDC before bare).

*Raised by: test-quality.*

pkg/oauthproto/discovery.go
```diff
@@ -1,18 +1,234 @@
...
+
+	var attemptErrs []string
+	for _, u := range urls {
+		metadata, err := fetchDiscoveryDocument(fetchCtx, httpClient, u, issuer)
+		if err != nil {
+			attemptErrs = append(attemptErrs, fmt.Sprintf("%s: %v", u, err))
+			continue
+		}
+		if metadata.RegistrationEndpoint == "" {
+			return metadata, ErrRegistrationEndpointMissing
```
jhrozek: Follow-up thought, not blocking: returning `ErrRegistrationEndpointMissing` on the first 200 OK means we don't try paths 2 or 3 if path 1 happened to omit `registration_endpoint`. Real upstreams like Keycloak can publish the field on one well-known but not the other, so the short-circuit risks blocking DCR for them.

There's prior art in `pkg/auth/oauth/oidc.go:114-134` — when one well-known doc lacks `registration_endpoint`, it tries the other and merges the field after re-checking issuer equality. Worth mirroring here when sub-issue B starts consuming this.

Sketch (preserve the partial across iterations, only return the sentinel after URLs exhaust):

```go
var partial *AuthorizationServerMetadata
for _, u := range urls {
    metadata, err := fetchDiscoveryDocument(fetchCtx, httpClient, u, issuer)
    if err != nil {
        attemptErrs = append(attemptErrs, fmt.Errorf("%s: %w", u, err))
        continue
    }
    if metadata.RegistrationEndpoint != "" {
        return metadata, nil
    }
    if partial == nil {
        partial = metadata
    }
}
if partial != nil {
    return partial, ErrRegistrationEndpointMissing
}
```

Tangent for a separate cleanup PR: most of `buildDiscoveryURLs` / `fetchDiscoveryDocument` / the default client here overlaps with `discoverOIDCEndpointsWithClientAndValidation` in `pkg/auth/oauth/oidc.go`. Pulling `pkg/auth/oauth` wholesale into `oauthproto` is blocked by the leaf-package invariant, but exporting the URL builder and per-URL fetch helper from `oauthproto` would let `pkg/auth/oauth` call them while keeping its loose-issuer / `insecureAllowHTTP` / per-endpoint-revalidate / merge-on-missing logic on top. Worth scoping after this lands.

pkg/oauthproto/discovery.go
```diff
@@ -1,18 +1,234 @@
...
+		}
+		return metadata, nil
+	}
+
+	return nil, fmt.Errorf(
+		"failed to discover authorization server metadata for issuer %q: %s",
+		issuer, strings.Join(attemptErrs, "; "),
+	)
```
jhrozek, line 124: Optional follow-up: stringifying each attempt error into `attemptErrs` and joining with `strings.Join` breaks `errors.Is` — a future caller that wants to retry on `context.DeadlineExceeded` but not on issuer mismatch will end up string-matching.

`errors.Join` (Go 1.20+) keeps the same UX (newline-separated `Error()`) and preserves the chain:

```go
var attemptErrs []error
for _, u := range urls {
    ...
    attemptErrs = append(attemptErrs, fmt.Errorf("%s: %w", u, err))
    ...
}
return nil, fmt.Errorf(
    "failed to discover authorization server metadata for issuer %q: %w",
    issuer, errors.Join(attemptErrs...),
)
```

Pairs naturally with the other comment since both touch the same loop.

jhrozek (dismissed):
Approving — purely additive surface, tests are thorough, lint clean. Two non-blocking suggestions inline that can land as follow-up PRs after sub-issue B starts wiring up consumers.

pkg/authserver/config.go
tgrunnagle: Structurally handled later in the stack

jhrozek (approved):

PR #5044, opened 2026-04-23
Wire authserver DCR resolver and add structured logs

### LINKED ISSUE #5039: Authserver DCR: wire resolver into authserver and add structured logs (Phase 2, Steps 2d/2g)
## Context

Part of #4978 (Phase 2 of the DCR story, #4976). This sub-issue wires the DCR resolver from #5038 into the embedded authserver startup path and adds the three structured-log upgrades.

**Depends on**: #5038 (Sub-issue B — `DCRCredentialStore`, `resolveDCRCredentials`, `needsDCR`, `applyResolution` must land first).
**Followed by**: Sub-issue D (operator CRD surface).

---

### Step 2d — Wiring (`pkg/authserver/runner/embeddedauthserver.go`)

- Add `dcrStore DCRCredentialStore` field to `EmbeddedAuthServer` struct.
- Initialize it in `NewEmbeddedAuthServer` via `NewInMemoryDCRCredentialStore()`.
- In `buildUpstreamConfigs` (~line 274), use the already-present `ctx` parameter (currently underscore-ignored — just remove the `_`).
- For each upstream run-config element: `rcCopy := rc` (copy-before-mutate, per `.claude/rules/go-style.md`). When `needsDCR(&rcCopy)` is true, call `resolveDCRCredentials(ctx, &rcCopy, s.issuer, s.dcrStore)` and `applyResolution(&rcCopy, resolution)` on the copy before passing to `buildPureOAuth2Config(rcCopy)`.
- `buildPureOAuth2Config` signature and body are **unchanged**: no `ctx` parameter added, no I/O introduced.

### Step 2g — Structured logs (three touch-points)

1. `pkg/authserver/server/handlers/dcr.go:98` — upgrade the successful-registration log from `slog.Debug` to `slog.Info` with structured fields `upstream`, `issuer`, `client_id`, `software_id`, `token_endpoint_auth_method`, `scopes`.
2. `pkg/authserver/runner/dcr.go` (from #5038) — add inside `resolveDCRCredentials`:
   - Cache-hit: `slog.Debug` with `dcr_age_days` derived from `res.CreatedAt`.
   - Cache-hit with `dcr_age_days > dcrStaleAgeThreshold`: additionally `slog.Warn` with a remediation hint.
   - Any failure: `slog.Error` with wrapped cause plus a `step` attribute identifying which step failed (metadata discovery, DCR call, store read, store write).
3. `pkg/auth/monitored_token_source.go:361` — inside the existing `*oauth2.RetrieveError` permanent-4xx branch of `isTransientNetworkError`, add `slog.Warn` with fields `workload`, `upstream`, `client_id` and remediation text: `"cached DCR client is no longer recognized by the authorization server; delete the cached credentials and restart to re-register."` The `return false` boolean is unchanged.

### Integration tests (`pkg/authserver/integration_test.go`)

Extend the existing file (do NOT create a new file). Add two scenarios:

1. **DCR boot**: authserver boots with `OAuth2UpstreamRunConfig{ClientID: "", DCRConfig: &DCRUpstreamConfig{DiscoveryURL: mockASURL}}`; mock AS serves metadata + `/register`; after boot, the upstream config output has DCR-resolved `ClientID` and `ClientSecret`; in-memory store contains the corresponding `DCRKey`.
2. **Cache-hit**: second call to `buildUpstreamConfigs` against the same in-memory store makes zero additional HTTP requests to the mock AS.
3. **Copy-before-mutate verification**: `s.cfg.Upstreams[i]` (the original slice element) is unchanged after both calls.

---

### Step 2d
- [ ] `EmbeddedAuthServer` struct has a `dcrStore DCRCredentialStore` field.
- [ ] `NewEmbeddedAuthServer` initializes `dcrStore` with `NewInMemoryDCRCredentialStore()`.
- [ ] `buildUpstreamConfigs` uses the `ctx` parameter (no longer underscore-ignored).
- [ ] `buildUpstreamConfigs` performs copy-before-mutate: `rcCopy := rc` before any mutation.
- [ ] When `needsDCR(&rcCopy)` is true, `resolveDCRCredentials(ctx, &rcCopy, issuer, s.dcrStore)` and `applyResolution(&rcCopy, resolution)` are called on the copy before `buildPureOAuth2Config`.
- [ ] `buildPureOAuth2Config` has no `context.Context` parameter — verified by: `grep -n 'context.Context' pkg/authserver/runner/embeddedauthserver.go` does not include `buildPureOAuth2Config`.
- [ ] `s.cfg.Upstreams[i]` (the original slice element) is unchanged after `buildUpstreamConfigs` returns.

### Step 2g
- [ ] `pkg/authserver/server/handlers/dcr.go` line 98: `slog.Info` (not `Debug`) on successful registration with fields `upstream`, `issuer`, `client_id`, `software_id`, `token_endpoint_auth_method`, `scopes`.
- [ ] `resolveDCRCredentials` emits `slog.Debug` on cache hit with `dcr_age_days` field.
- [ ] `resolveDCRCredentials` additionally emits `slog.Warn` when `dcr_age_days` exceeds `dcrStaleAgeThreshold`.
- [ ] `resolveDCRCredentials` emits `slog.Error` with `step` field on any failure.
- [ ] `isTransientNetworkError` permanent-4xx branch emits `slog.Warn` with fields `workload`, `upstream`, `client_id` and DCR remediation text. `return false` unchanged.
- [ ] Secret grep across all modified files returns no `slog.*` arguments containing `client_secret`, `registration_access_token`, `initial_access_token`, or `refresh_token`: `grep -nE '(client_secret|registration_access_token|initial_access_token|refresh_token)' $(git diff --name-only main -- 'pkg/authserver' 'pkg/oauthproto' 'pkg/auth/monitored_token_source.go')` returns only struct-field / JSON-tag hits.

### Integration tests
- [ ] Scenario 1: authserver boots with empty `ClientID` + `DCRConfig.DiscoveryURL`; mock AS serves metadata + `/register`; DCR-resolved `ClientID` and `ClientSecret` present in upstream config output; in-memory store contains the corresponding `DCRKey`.
- [ ] Scenario 2: second `buildUpstreamConfigs` call against same in-memory store makes zero additional HTTP requests to mock AS.
- [ ] `s.cfg.Upstreams[i]` unchanged after both calls.

### Cross-cutting

---

## Patterns & References

- Copy-before-mutate: `.claude/rules/go-style.md` §"Copy Before Mutating Caller Input".
- Purity gate: `buildPureOAuth2Config` at `pkg/authserver/runner/embeddedauthserver.go:~369` must keep zero network I/O and no `context.Context` parameter.
- `resolveSecret` at `pkg/authserver/runner/embeddedauthserver.go:409` — already used for `ClientSecretFile`/`ClientSecretEnvVar`; used by the resolver for initial access token.
- Integration test home: `pkg/authserver/integration_test.go` — extend this file, do not create a new one.
- Plain `testing` + `testify` (not Ginkgo). `t.Cleanup(server.Close)`.
- Never log secrets (`.claude/rules/go-style.md` §"Never log credentials").

tgrunnagle:
## Summary

- **Why**: `resolveDCRCredentials` was wired up and unit-tested in the prior sub-issue, but `EmbeddedAuthServer` never called it, so any upstream configured with `DCRConfig` still booted without a client. This PR makes the resolver load-bearing and adds the operator-visible structured logs the Phase 2 spec calls for (cache-hit age, stale-registration warning, per-step errors, and a remediation hint when an upstream rejects a cached client as 4xx).
- **What**:
  - `EmbeddedAuthServer` now owns an in-memory `DCRCredentialStore` and calls `resolveDCRCredentials` + `consumeResolution` for every OAuth2 upstream whose config requests DCR. `buildPureOAuth2Config` deliberately keeps its pure, network-free signature; the DCR-resolved `ClientSecret` is layered on afterwards via a new `applyResolutionToOAuth2Config` helper so the two application sites stay side-by-side.
  - Each upstream `RunConfig` element is shallow-copied and its OAuth2 sub-config is deep-copied before mutation, preserving the caller's `RunConfig.Upstreams` slice.
  - `resolveDCRCredentials` returns a new `dcrStepError` wrapping the originating step (metadata discovery, DCR call, store read, store write) instead of logging-and-returning at each branch. The boundary caller (`buildUpstreamConfigs`) emits one `slog.Error` via `logDCRStepError`, so there is a single error line per failure. A panic inside the singleflight closure is converted to the same `dcrStepError` carrying the captured `debug.Stack()`, so the boundary log surfaces it without an in-defer duplicate.
  - Cache-hit path emits `slog.Debug` with `dcr_age_days` and additionally `slog.Warn` with remediation text when the cached registration exceeds the 90-day `dcrStaleAgeThreshold`.
  - The `/oauth/register` handler success log emits at `Debug` with `issuer`, `client_id`, `software_id`, `token_endpoint_auth_method`, and `scopes` — silent at INFO+ per the project's logging convention. `SoftwareID` is threaded through `ValidateDCRRequest`, which now also caps it to 256 printable-ASCII characters.
  - `MonitoredTokenSource` accepts `upstream` and `client_id` at construction (replacing the earlier `SetUpstreamContext` setter, which had an unsynchronized writer). On the first transition to `Unauthenticated` driven by a permanent token-endpoint 4xx, it emits a single `slog.Warn` with a DCR/CIMD remediation hint (gated by `stopOnce` so a tight `Token()` loop cannot spam it, and suppressed when no `client_id` context is available). The classifier itself (`isTransientNetworkError`) is side-effect free.
  - `NewHandler` now fails loudly when `AuthorizationServerConfig` or its embedded `*fosite.Config` is nil; `issuer()` no longer carries a runtime nil-guard.
  - Error strings are run through a sanitizer that strips URL **userinfo, query, and fragment** components before logging. The regex matches `http(s)://` case-insensitively per RFC 3986 §3.1, and trailing sentence punctuation around a URL is preserved so prose is not mangled.
  - `pkg/oauthproto/dcr.go` `handleHTTPResponse` caps the upstream `/register` error body at 8 KiB via `io.LimitReader` so a non-conformant or malicious upstream cannot inflate operator log volume.

Closes #5039

## Type of change

- [x] New feature

## Test plan

New coverage added:

- `pkg/authserver/runner/embeddedauthserver_test.go` — full DCR boot against a mock AS (metadata + `/register`), cache-hit short-circuit (zero additional HTTP requests on the second `buildUpstreamConfigs`), copy-before-mutate assertion on the caller's `RunConfig.Upstreams` slice, and `TestNewEmbeddedAuthServer_DCRBoot` driving the real constructor to assert `dcrStore` is populated.
- `pkg/authserver/runner/dcr_test.go` — cache-hit Debug log with `dcr_age_days`, stale-age Warn above threshold, one `dcrStepError`-returning case per failure step, and the panic-recovery test now asserts the wrapped error is a `*dcrStepError(dcrStepRegister, …, Stack: <captured>)`.
- `pkg/authserver/server/handlers/handler_chain_test.go` — `TestHandler_issuer` covers both populated and zero-value `AccessTokenIssuer`. `TestNewHandler_ErrorsOnNilConfig` (in `authorize_test.go`) pins the new constructor invariant.
- `pkg/authserver/server/registration/dcr_test.go` — `software_id` length cap and non-printable-ASCII rejection.
- `pkg/auth/monitored_token_source_test.go` — permanent-4xx branch emits the remediation Warn with `workload`/`upstream`/`client_id` populated via the new constructor parameters.
- `pkg/runner/runner_test.go` — `TestRemoteAuthLogContext` covers the full `cached CIMD > cached DCR > static` precedence.
- Sanitizer unit tests cover commas, periods, closing parens, mixed runs, a Go `http.Client`-style quoted URL, embedded `user:pass@host` userinfo (with and without query), implicit-flow `#access_token=…` fragments, and uppercase `HTTPS://` schemes.

## Changes

| File | Change |
|------|--------|
| `pkg/authserver/runner/embeddedauthserver.go` | Add `dcrStore` field and wire resolver into `buildUpstreamConfigs`; introduce `applyResolutionToOAuth2Config`; document the per-instance store vs. process-wide singleflight asymmetry. |
| `pkg/authserver/runner/dcr.go` | Introduce `dcrStepError` / `logDCRStepError`; emit cache-hit Debug, stale-age Warn, and per-step error context; carry panic stack on the wrapped error so a single boundary record surfaces it; URL-secret-stripping sanitizer (userinfo + query + fragment, case-insensitive scheme match). |
| `pkg/authserver/runner/dcr_store.go` | Inline doc rationale (no path citations) for resource-leak, copy-before-mutate, and constructor-validation reasoning. |
| `pkg/authserver/server/handlers/dcr.go` | DCR success log emits at `Debug`, drop redundant `upstream` attribute, thread `software_id` through. |
| `pkg/authserver/server/handlers/handler.go` | `NewHandler` validates `config` and `config.Config`; remove the runtime nil-guard from `issuer()`. |
| `pkg/authserver/server/registration/dcr.go` | Cap `software_id` to 256 printable-ASCII characters; export `MaxSoftwareIDLength`. |
| `pkg/auth/monitored_token_source.go` | Accept `upstream`/`client_id` as constructor parameters; emit DCR/CIMD remediation Warn from `markAsUnauthenticated` (once per state transition, gated by `stopOnce`, suppressed when `clientID == ""`). The classifier is side-effect free. |
| `pkg/runner/runner.go` | Populate new `MonitoredTokenSource` fields from `RemoteAuthConfig` with `cached CIMD > cached DCR > static` precedence (mirrors `remote.Handler.resolveClientCredentials`). |
| `pkg/oauthproto/dcr.go` | Cap the upstream `/register` error body at 8 KiB via `io.LimitReader` before embedding it in the returned error string. |
| `pkg/authserver/runner/embeddedauthserver_test.go` | DCR boot, cache-hit, and copy-before-mutate integration tests. |
| `pkg/authserver/runner/dcr_test.go` | Log and per-step error tests, sanitizer userinfo/fragment/uppercase cases, augmented panic-recovery assertions. |
| `pkg/authserver/server/handlers/handler_chain_test.go`, `authorize_test.go` | `TestHandler_issuer` and `TestNewHandler_ErrorsOnNilConfig`. |
| `pkg/authserver/server/handlers/dcr_test.go`, `pkg/authserver/server/registration/dcr_test.go` | Assertions for new log fields and `software_id` validation. |
| `pkg/auth/monitored_token_source_test.go`, `pkg/runner/runner_test.go` | New constructor plumbing, remediation Warn, and the CIMD-wins precedence case (closes review finding F26). |

## Large PR Justification

The PR exceeds the 400-LOC / 10-file guideline (1342+/104- across 18 files), but the change cannot be split without leaving the codebase in an intermediate non-functional state:

- ~660 lines are tests (`*_test.go`) — table-driven coverage for the resolver wiring, the new error taxonomy, the staleness + remediation logs, and the sanitizer.
- The remaining ~510 production lines across 9 files implement one logical change (wire the resolver into the auth-server boot path) plus four tightly-coupled sub-concerns: resolver wiring, `dcrStepError` taxonomy + boundary logging, monitored-token-source remediation log, and `software_id` validation. Splitting the resolver wiring without the matching error taxonomy would land a regression in observability; splitting the error taxonomy without the wiring would orphan the new types.

Operators gain two new diagnostic signals on DCR-enabled upstreams:

1. A `Warn` when a cached DCR registration is older than 90 days, hinting that the operator may want to rotate.
2. A `Warn` on the first transition to `Unauthenticated` driven by a permanent token-endpoint 4xx, hinting that the workload's cached DCR or CIMD credentials may be stale and that deleting them and restarting will trigger re-registration. The Warn fires at most once per state transition and is suppressed entirely for workloads with no `client_id` context.

The `MonitoredTokenSource` constructor signature has changed to require `upstream` and `client_id`, but it is an internal API with no public surface, and the prior `SetUpstreamContext` setter is removed.

Successful DCR registrations log at `Debug`, not `Info`, in keeping with the project's silent-success-at-INFO+ convention.

- The purity gate on `buildPureOAuth2Config` is load-bearing: it takes no `context.Context` and performs no network I/O. DCR resolution is deliberately sequenced *around* it, not inside it, so the pure config builder stays trivially testable. Please flag any future change that threads `ctx` into it.
- `dcrStepError` moves the "one error line per failure" invariant from the resolver to the boundary caller. If a future caller forgets to invoke `logDCRStepError`, the error will still propagate normally but the structured log will be missing. Worth keeping in mind if we add more resolver call sites.
- `consumeResolution` (formerly `applyResolution`) is a one-shot state transition, not an idempotent default-fill: a second call after the first is a no-op only because the first cleared `DCRConfig`. The companion `applyResolutionToOAuth2Config` writes the inline-only `ClientSecret` onto the built `*upstream.OAuth2Config`; any future output path from `OAuth2UpstreamRunConfig` to `upstream.OAuth2Config` must call **both** helpers to get a fully-resolved DCR client.
- The package-level `dcrFlight` singleflight is intentionally process-wide while `EmbeddedAuthServer.dcrStore` is per-instance — see the doc comments on each. The asymmetry is load-bearing for concurrent `EmbeddedAuthServer` instances in the same process.

### DIFF
--- a/pkg/auth/monitored_token_source.go
+++ b/pkg/auth/monitored_token_source.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net"
+	"net/http"
 	"os"
 	"strconv"
 	"strings"
@@ -444,18 +445,33 @@ func isTransientNetworkError(err error) bool {
 }
 
 // isPermanentTokenEndpointError reports whether err is an *oauth2.RetrieveError
-// with a non-5xx HTTP status (i.e. a permanent token-endpoint response such as
-// 400, 401, 403). Used at state-transition boundaries to decide whether to
-// emit a DCR/CIMD remediation hint alongside the unauthentication.
+// whose status implies the cached client credentials are themselves the
+// problem — specifically 400 (invalid_grant / invalid_client), 401, or
+// 403. Used at state-transition boundaries to decide whether to emit a
+// DCR/CIMD remediation hint alongside the unauthentication.
+//
+// Other 4xx codes are intentionally NOT treated as permanent here even
+// though isTransientNetworkError classifies the whole RetrieveError
+// branch as non-transient. 408 (Request Timeout) and 429 (Too Many
+// Requests) are typically transient back-pressure that the operator
+// cannot remediate by deleting cached credentials; firing the
+// "delete the cached credentials and restart" Warn on those would
+// mislead operators chasing a transient hiccup. The narrower allowlist
+// keeps the remediation hint truthful.
 func isPermanentTokenEndpointError(err error) bool {
 	retrieveErr, ok := errors.AsType[*oauth2.RetrieveError](err)
 	if !ok {
 		return false
 	}
 	if retrieveErr.Response == nil {
-		return true
+		return false
 	}
-	return retrieveErr.Response.StatusCode < 500
+	switch retrieveErr.Response.StatusCode {
+	case http.StatusBadRequest, http.StatusUnauthorized, http.StatusForbidden:
+		return true
+	default:
+		return false
+	}
 }
 
 // isOAuthParseError detects errors from the oauth2 library that indicate the
--- a/pkg/auth/monitored_token_source.go
+++ b/pkg/auth/monitored_token_source.go
@@ -132,9 +132,9 @@ type transientRefresher struct {
 	// upstream identifies the upstream authorization server that issued the
 	// token, and clientID is the OAuth 2.0 client_id used by this workload.
 	// Both are optional and are only surfaced in structured logs (notably the
-	// DCR remediation warning emitted from isTransientNetworkError when a
-	// permanent 4xx indicates a stale cached DCR client). Empty strings are
-	// acceptable and are omitted from log output by the slog handler.
+	// DCR/CIMD remediation warning emitted from MonitoredTokenSource on the
+	// transition to Unauthenticated). Empty strings are acceptable and are
+	// omitted from log output by the slog handler.
 	upstream string
 	clientID string
 
@@ -181,7 +181,7 @@ func (r *transientRefresher) retry(ctx context.Context, origErr error) (*oauth2.
 		if tokenErr == nil {
 			return t, nil
 		}
-		if !isTransientNetworkError(tokenErr, r.workload, r.upstream, r.clientID) {
+		if !isTransientNetworkError(tokenErr) {
 			return nil, backoff.Permanent(tokenErr)
 		}
 		return nil, tokenErr
@@ -239,11 +239,12 @@ type MonitoredTokenSource struct {
 // oauth2.TokenSource and monitors it for authentication failures.
 //
 // upstream and clientID annotate structured logs emitted by the token source,
-// most importantly the DCR remediation warning fired from
-// isTransientNetworkError when the token endpoint returns a permanent 4xx
-// (which frequently indicates a stale cached RFC 7591 registration). Pass
-// empty strings when the workload does not use DCR or the upstream issuer
-// is unknown; the slog handler will render the attributes as empty.
+// most importantly the DCR/CIMD remediation warning fired on the transition
+// to Unauthenticated when the token endpoint returns a permanent 4xx (which
+// frequently indicates stale cached credentials). Pass empty strings when
+// the workload does not use DCR/CIMD or the upstream issuer is unknown; the
+// remediation log will be suppressed when clientID is empty since its
+// operator-correlation field would be blank.
 //
 // The fields are fixed at construction time rather than exposed via a setter
 // so there is no data race between a late writer and the readers in Token()
@@ -295,8 +296,11 @@ func (mts *MonitoredTokenSource) Token() (*oauth2.Token, error) {
 		return tok, nil
 	}
 
-	if !isTransientNetworkError(err, mts.workloadName, mts.upstream, mts.clientID) {
-		mts.markAsUnauthenticated(fmt.Sprintf("Token retrieval failed: %v", err))
+	if !isTransientNetworkError(err) {
+		mts.markAsUnauthenticated(
+			fmt.Sprintf("Token retrieval failed: %v", err),
+			isPermanentTokenEndpointError(err),
+		)
 		return nil, err
 	}
 
@@ -305,7 +309,10 @@ func (mts *MonitoredTokenSource) Token() (*oauth2.Token, error) {
 	tok, err = mts.refresher.Refresh(mts.monitoringCtx, err)
 	if err != nil {
 		if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
-			mts.markAsUnauthenticated(fmt.Sprintf("Token refresh failed after retries: %v", err))
+			mts.markAsUnauthenticated(
+				fmt.Sprintf("Token refresh failed after retries: %v", err),
+				isPermanentTokenEndpointError(err),
+			)
 		}
 		return nil, err
 	}
@@ -382,12 +389,11 @@ func (mts *MonitoredTokenSource) onTick() (bool, time.Duration) {
 // (certificate verification, handshake failure) are NOT considered transient and
 // return false so the workload is marked unauthenticated immediately.
 //
-// workload, upstream, and clientID are context strings used only in
-// structured logs — notably the DCR remediation warning emitted in the
-// permanent-4xx branch, which suggests the cached RFC 7591 registration is
-// no longer recognised by the authorization server. All three are optional;
-// pass empty strings when the context is unknown.
-func isTransientNetworkError(err error, workload, upstream, clientID string) bool {
+// The function is side-effect free; callers that want to emit a DCR
+// remediation hint on a permanent 4xx must do so themselves at the
+// state-transition boundary using isPermanentTokenEndpointError to
+// classify, so a tight Token() loop does not spam the same record.
+func isTransientNetworkError(err error) bool {
 	if err == nil ||
 		errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 		return false
@@ -403,20 +409,6 @@ func isTransientNetworkError(err error, workload, upstream, clientID string) boo
 			)
 			return true
 		}
-		// Permanent 4xx from the token endpoint frequently indicates that the
-		// cached Dynamic Client Registration has been revoked, rotated, or is
-		// no longer recognised by the authorization server. Emit a remediation
-		// hint before returning false so operators can correlate the
-		// unauthentication with a stale DCR rather than a user-action issue.
-		// The returned boolean is unchanged.
-		//nolint:gosec // G706: client_id is public metadata per RFC 7591.
-		slog.Warn(
-			"cached DCR client is no longer recognized by the authorization server; "+
-				"delete the cached credentials and restart to re-register.",
-			"workload", workload,
-			"upstream", upstream,
-			"client_id", clientID,
-		)
 		return false
 	}
 
@@ -451,6 +443,21 @@ func isTransientNetworkError(err error, workload, upstream, clientID string) boo
 	return false
 }
 
+// isPermanentTokenEndpointError reports whether err is an *oauth2.RetrieveError
+// with a non-5xx HTTP status (i.e. a permanent token-endpoint response such as
+// 400, 401, 403). Used at state-transition boundaries to decide whether to
+// emit a DCR/CIMD remediation hint alongside the unauthentication.
+func isPermanentTokenEndpointError(err error) bool {
+	retrieveErr, ok := errors.AsType[*oauth2.RetrieveError](err)
+	if !ok {
+		return false
+	}
+	if retrieveErr.Response == nil {
+		return true
+	}
+	return retrieveErr.Response.StatusCode < 500
+}
+
 // isOAuthParseError detects errors from the oauth2 library that indicate the
 // token endpoint returned an unparsable response body on a 2xx status. This
 // typically happens when a load balancer, CDN, or reverse proxy intercepts the
@@ -466,13 +473,39 @@ func isOAuthParseError(err error) bool {
 		strings.Contains(msg, "oauth2: cannot parse response")
 }
 
-// markAsUnauthenticated marks the workload as unauthenticated and stops background monitoring.
-func (mts *MonitoredTokenSource) markAsUnauthenticated(reason string) {
+// markAsUnauthenticated marks the workload as unauthenticated and stops
+// background monitoring. If permanent4xx is true and the workload was
+// constructed with a non-empty client_id, a one-shot DCR/CIMD remediation
+// hint is emitted alongside the stop transition. The hint and the close
+// of stopMonitoring share stopOnce, so a caller (e.g. a tight Token()
+// loop) cannot spam the record on every call after the workload has
+// already transitioned to Unauthenticated.
+func (mts *MonitoredTokenSource) markAsUnauthenticated(reason string, permanent4xx bool) {
 	_ = mts.statusUpdater.SetWorkloadStatus(
 		context.Background(),
 		mts.workloadName,
 		runtime.WorkloadStatusUnauthenticated,
 		reason,
 	)
-	mts.stopOnce.Do(func() { close(mts.stopMonitoring) })
+	mts.stopOnce.Do(func() {
+		// A permanent 4xx from the token endpoint commonly indicates the
+		// cached client (DCR or CIMD) is no longer recognised — but the
+		// same branch fires for revoked consent, disabled accounts, and
+		// statically configured clients, so the message has to be honest
+		// about the variability. Gating on clientID != "" suppresses the
+		// log entirely for workloads where no client_id context is
+		// available; the operator-correlation it provides would be empty.
+		if permanent4xx && mts.clientID != "" {
+			//nolint:gosec // G706: client_id is public metadata per RFC 7591.
+			slog.Warn(
+				"token endpoint returned a permanent error; if this workload uses "+
+					"cached DCR or CIMD credentials they may be stale — delete the "+
+					"cached credentials and restart to re-register.",
+				"workload", mts.workloadName,
+				"upstream", mts.upstream,
+				"client_id", mts.clientID,
+			)
+		}
+		close(mts.stopMonitoring)
+	})
 }
--- a/pkg/auth/monitored_token_source.go
+++ b/pkg/auth/monitored_token_source.go
@@ -129,6 +129,15 @@ type transientRefresher struct {
 	source   oauth2.TokenSource
 	workload string
 
+	// upstream identifies the upstream authorization server that issued the
+	// token, and clientID is the OAuth 2.0 client_id used by this workload.
+	// Both are optional and are only surfaced in structured logs (notably the
+	// DCR remediation warning emitted from isTransientNetworkError when a
+	// permanent 4xx indicates a stale cached DCR client). Empty strings are
+	// acceptable and are omitted from log output by the slog handler.
+	upstream string
+	clientID string
+
 	// newBackOff is a factory for the backoff used during retries.
 	// Nil in production; overridable in tests for fast execution.
 	newBackOff func() backoff.BackOff
@@ -172,7 +181,7 @@ func (r *transientRefresher) retry(ctx context.Context, origErr error) (*oauth2.
 		if tokenErr == nil {
 			return t, nil
 		}
-		if !isTransientNetworkError(tokenErr) {
+		if !isTransientNetworkError(tokenErr, r.workload, r.upstream, r.clientID) {
 			return nil, backoff.Permanent(tokenErr)
 		}
 		return nil, tokenErr
@@ -212,6 +221,8 @@ func (r *transientRefresher) getBackOff() backoff.BackOff {
 type MonitoredTokenSource struct {
 	tokenSource    oauth2.TokenSource
 	workloadName   string
+	upstream       string
+	clientID       string
 	statusUpdater  StatusUpdater
 	monitoringCtx  context.Context
 	stopMonitoring chan struct{}
@@ -226,20 +237,41 @@ type MonitoredTokenSource struct {
 
 // NewMonitoredTokenSource creates a new MonitoredTokenSource that wraps the provided
 // oauth2.TokenSource and monitors it for authentication failures.
+//
+// upstream and clientID annotate structured logs emitted by the token source,
+// most importantly the DCR remediation warning fired from
+// isTransientNetworkError when the token endpoint returns a permanent 4xx
+// (which frequently indicates a stale cached RFC 7591 registration). Pass
+// empty strings when the workload does not use DCR or the upstream issuer
+// is unknown; the slog handler will render the attributes as empty.
+//
+// The fields are fixed at construction time rather than exposed via a setter
+// so there is no data race between a late writer and the readers in Token()
+// / transientRefresher.retry() — both of which may run on the background
+// monitor goroutine started by StartBackgroundMonitoring.
 func NewMonitoredTokenSource(
 	ctx context.Context,
 	tokenSource oauth2.TokenSource,
 	workloadName string,
+	upstream string,
+	clientID string,
 	statusUpdater StatusUpdater,
 ) *MonitoredTokenSource {
 	return &MonitoredTokenSource{
 		tokenSource:    tokenSource,
 		workloadName:   workloadName,
+		upstream:       upstream,
+		clientID:       clientID,
 		statusUpdater:  statusUpdater,
 		monitoringCtx:  ctx,
 		stopMonitoring: make(chan struct{}),
 		stopped:        make(chan struct{}),
-		refresher:      &transientRefresher{source: tokenSource, workload: workloadName},
+		refresher: &transientRefresher{
+			source:   tokenSource,
+			workload: workloadName,
+			upstream: upstream,
+			clientID: clientID,
+		},
 	}
 }
 
@@ -263,7 +295,7 @@ func (mts *MonitoredTokenSource) Token() (*oauth2.Token, error) {
 		return tok, nil
 	}
 
-	if !isTransientNetworkError(err) {
+	if !isTransientNetworkError(err, mts.workloadName, mts.upstream, mts.clientID) {
 		mts.markAsUnauthenticated(fmt.Sprintf("Token retrieval failed: %v", err))
 		return nil, err
 	}
@@ -349,7 +381,13 @@ func (mts *MonitoredTokenSource) onTick() (bool, time.Duration) {
 // OAuth2 client-level auth failures (invalid_grant, 401, 400) and TLS errors
 // (certificate verification, handshake failure) are NOT considered transient and
 // return false so the workload is marked unauthenticated immediately.
-func isTransientNetworkError(err error) bool {
+//
+// workload, upstream, and clientID are context strings used only in
+// structured logs — notably the DCR remediation warning emitted in the
+// permanent-4xx branch, which suggests the cached RFC 7591 registration is
+// no longer recognised by the authorization server. All three are optional;
+// pass empty strings when the context is unknown.
+func isTransientNetworkError(err error, workload, upstream, clientID string) bool {
 	if err == nil ||
 		errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 		return false
@@ -365,6 +403,20 @@ func isTransientNetworkError(err error) bool {
 			)
 			return true
 		}
+		// Permanent 4xx from the token endpoint frequently indicates that the
+		// cached Dynamic Client Registration has been revoked, rotated, or is
+		// no longer recognised by the authorization server. Emit a remediation
+		// hint before returning false so operators can correlate the
+		// unauthentication with a stale DCR rather than a user-action issue.
+		// The returned boolean is unchanged.
+		//nolint:gosec // G706: client_id is public metadata per RFC 7591.
+		slog.Warn(
+			"cached DCR client is no longer recognized by the authorization server; "+
+				"delete the cached credentials and restart to re-register.",
+			"workload", workload,
+			"upstream", upstream,
+			"client_id", clientID,
+		)
 		return false
 	}
 
--- a/pkg/authserver/runner/dcr.go
+++ b/pkg/authserver/runner/dcr.go
@@ -12,6 +12,7 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"regexp"
 	"runtime/debug"
 	"slices"
 	"sort"
@@ -21,6 +22,7 @@ import (
 	"golang.org/x/sync/singleflight"
 
 	"github.com/stacklok/toolhive/pkg/authserver"
+	"github.com/stacklok/toolhive/pkg/authserver/upstream"
 	"github.com/stacklok/toolhive/pkg/networking"
 	"github.com/stacklok/toolhive/pkg/oauthproto"
 )
@@ -96,6 +98,15 @@ type DCRResolution struct {
 	// token endpoint for this client.
 	TokenEndpointAuthMethod string
 
+	// RedirectURI is the redirect URI presented to the authorization server
+	// during registration. When the caller's run-config did not specify one,
+	// this holds the defaulted value derived from the issuer + /oauth/callback
+	// (via resolveUpstreamRedirectURI). Persisting it on the resolution lets
+	// applyResolution write it back onto the run-config COPY so that
+	// downstream consumers (buildPureOAuth2Config, upstream.OAuth2Config
+	// validation) see a non-empty RedirectURI.
+	RedirectURI string
+
 	// ClientIDIssuedAt is the RFC 7591 §3.2.1 "client_id_issued_at" value
 	// converted to a Go time.Time. Zero when the upstream omitted the field
 	// (the field is OPTIONAL per RFC 7591). Informational; not used to
@@ -132,23 +143,41 @@ func needsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
 	return rc.ClientID == "" && rc.DCRConfig != nil
 }
 
-// applyResolution copies resolved credentials and endpoints from res into rc.
+// applyResolution copies resolved credentials and endpoints from res into rc
+// and consumes rc.DCRConfig (sets it to nil), transitioning the run-config
+// copy from "DCR-pending" (ClientID == "" && DCRConfig != nil) to "DCR-
+// resolved" (ClientID populated && DCRConfig == nil).
+//
 // Callers must pass a COPY of the upstream run-config (per the
 // copy-before-mutate rule in .claude/rules/go-style.md); applyResolution does
 // not clone rc internally.
 //
-// All three fields (ClientID, AuthorizationEndpoint, TokenEndpoint) are
-// written only when rc leaves them empty — explicit caller configuration
-// always wins. resolveDCRCredentials enforces ClientID == "" up front via
-// validateResolveInputs, so the conditional write here is defence-in-depth
-// against future call sites that bypass the resolver and invoke
-// applyResolution directly: an unconditional overwrite would silently
+// Why DCRConfig is cleared: OAuth2UpstreamRunConfig.Validate enforces
+// ClientID xor DCRConfig — a resolved copy that left DCRConfig set would
+// fail the validator that runs downstream in buildPureOAuth2Config. The
+// caller's *original* OAuth2Config is unaffected because
+// buildUpstreamConfigs deep-copies before resolution; only the post-
+// resolution copy is mutated here.
+//
+// ClientID, the endpoints, and RedirectURI are written only when rc leaves
+// them empty — explicit caller configuration always wins. The conditional
+// ClientID write is defence-in-depth against future call sites that bypass
+// the resolver's validateResolveInputs precondition (which enforces
+// ClientID == "" up front); an unconditional overwrite would silently
 // clobber a pre-provisioned ClientID with no error.
 //
-// Note: the resolved ClientSecret is NOT copied onto rc because
-// OAuth2UpstreamRunConfig models secrets as file-or-env references, not
-// inline values. Callers that need the resolved secret must read it from
-// the DCRResolution directly.
+// The defaulted RedirectURI write closes the loop on resolver-side defaulting:
+// when the caller's run-config left RedirectURI empty, resolveUpstreamRedirectURI
+// derived issuer + /oauth/callback and persisted it on the resolution; copying
+// it back here means the downstream upstream.OAuth2Config has a non-empty
+// RedirectURI, which authserver.Config validation requires.
+//
+// Note on ClientSecret: applyResolution does NOT write the resolved secret
+// to rc because OAuth2UpstreamRunConfig models secrets as file-or-env
+// references only. To propagate the DCR-resolved secret into the final
+// upstream.OAuth2Config, callers must pair this call with
+// applyResolutionToOAuth2Config once the config has been built. Keeping the
+// two helpers side-by-side localises the DCR-specific application logic.
 func applyResolution(rc *authserver.OAuth2UpstreamRunConfig, res *DCRResolution) {
 	if rc == nil || res == nil {
 		return
@@ -156,12 +185,35 @@ func applyResolution(rc *authserver.OAuth2UpstreamRunConfig, res *DCRResolution)
 	if rc.ClientID == "" {
 		rc.ClientID = res.ClientID
 	}
+	rc.DCRConfig = nil
 	if rc.AuthorizationEndpoint == "" {
 		rc.AuthorizationEndpoint = res.AuthorizationEndpoint
 	}
 	if rc.TokenEndpoint == "" {
 		rc.TokenEndpoint = res.TokenEndpoint
 	}
+	if rc.RedirectURI == "" {
+		rc.RedirectURI = res.RedirectURI
+	}
+}
+
+// applyResolutionToOAuth2Config overlays the DCR-resolved ClientSecret onto
+// a built *upstream.OAuth2Config. This is the companion to applyResolution:
+// where that function writes fields representable in the file-or-env run-
+// config model, this one writes the inline-only ClientSecret directly on
+// the runtime config.
+//
+// The split exists because buildPureOAuth2Config intentionally retains a
+// narrow file-or-env contract (no DCR awareness) and because OAuth2's
+// ClientSecret on the run-config is modelled as a reference rather than an
+// inline string. Any future output path from OAuth2UpstreamRunConfig to
+// upstream.OAuth2Config must call both helpers to get a fully-resolved
+// DCR client — exercised by TestBuildUpstreamConfigs_DCR.
+func applyResolutionToOAuth2Config(cfg *upstream.OAuth2Config, res *DCRResolution) {
+	if cfg == nil || res == nil {
+		return
+	}
+	cfg.ClientSecret = res.ClientSecret
 }
 
 // scopesHash returns the SHA-256 hex digest of the canonical scope set.
@@ -193,6 +245,58 @@ func scopesHash(scopes []string) string {
 	return hex.EncodeToString(h.Sum(nil))
 }
 
+// Step identifiers for structured error logs emitted by the caller of
+// resolveDCRCredentials. These values flow through the "step" attribute so
+// operators can narrow failures to a specific phase without parsing error
+// messages. They are reported only at the boundary log — see
+// DCRStepError — so a single failure produces a single slog.Error record.
+const (
+	dcrStepValidate         = "validate"
+	dcrStepResolveRedirect  = "resolve_redirect_uri"
+	dcrStepCacheRead        = "cache_read"
+	dcrStepMetadata         = "metadata_discovery"
+	dcrStepSelectAuthMethod = "select_auth_method"
+	dcrStepRegister         = "dcr_call"
+	dcrStepCacheWrite       = "cache_write"
+)
+
+// DCRStepError annotates a resolver error with the phase it was produced
+// in. The boundary caller (buildUpstreamConfigs) emits the single
+// slog.Error record for the failure; individual error branches inside
+// resolveDCRCredentials do not log so that each failure surfaces exactly
+// once in the combined log stream.
+//
+// RedirectURI is included when known so that operators can correlate the
+// failure with a specific upstream registration without parsing the
+// wrapped error string. A zero-value DCRStepError is invalid; construct
+// via newDCRStepError or the resolver's internal helpers.
+type DCRStepError struct {
+	Step        string
+	Issuer      string
+	RedirectURI string
+	Err         error
+}
+
+// Error implements error. The "step" tag mirrors the structured-log
+// attribute so command-line log scrapers see the same phase identifier.
+func (e *DCRStepError) Error() string {
+	return fmt.Sprintf("dcr: %s: %s", e.Step, e.Err.Error())
+}
+
+// Unwrap lets errors.Is / errors.As reach the wrapped cause.
+func (e *DCRStepError) Unwrap() error { return e.Err }
+
+// newDCRStepError builds a DCRStepError. It never returns nil for a
+// non-nil cause.
+func newDCRStepError(step, issuer, redirectURI string, err error) *DCRStepError {
+	return &DCRStepError{
+		Step:        step,
+		Issuer:      issuer,
+		RedirectURI: redirectURI,
+		Err:         err,
+	}
+}
+
 // resolveDCRCredentials performs Dynamic Client Registration for rc against
 // the upstream authorization server identified by rc.DCRConfig, caching the
 // resulting credentials in cache. On cache hit the resolver returns
@@ -212,6 +316,16 @@ func scopesHash(scopes []string) string {
 // The caller is responsible for applying the returned resolution onto a COPY
 // of rc via applyResolution (per the copy-before-mutate rule). This function
 // neither mutates rc nor the cache on failure.
+//
+// Observability: this function never calls slog.Error directly — all
+// failures are annotated with a *DCRStepError and returned to the caller,
+// which is expected to emit the boundary Error record. This avoids the
+// double-logging pattern where both the resolver and the outer frame
+// report the same failure. Cache-hit Debug / stale-Warn logs and the
+// successful-registration Debug log are emitted here because they have no
+// outer-frame equivalent. No secret values (client_secret,
+// registration_access_token, initial_access_token) are ever logged — only
+// public metadata such as client_id and redirect_uri.
 func resolveDCRCredentials(
 	ctx context.Context,
 	rc *authserver.OAuth2UpstreamRunConfig,
@@ -219,12 +333,13 @@ func resolveDCRCredentials(
 	cache DCRCredentialStore,
 ) (*DCRResolution, error) {
 	if err := validateResolveInputs(rc, localIssuer, cache); err != nil {
-		return nil, err
+		return nil, newDCRStepError(dcrStepValidate, localIssuer, "", err)
 	}
 
 	redirectURI, err := resolveUpstreamRedirectURI(rc.RedirectURI, localIssuer)
 	if err != nil {
-		return nil, fmt.Errorf("dcr: resolve redirect uri: %w", err)
+		return nil, newDCRStepError(dcrStepResolveRedirect, localIssuer, "",
+			fmt.Errorf("resolve redirect uri: %w", err))
 	}
 
 	scopes := slices.Clone(rc.Scopes)
@@ -236,7 +351,7 @@ func resolveDCRCredentials(
 
 	// Cache lookup short-circuits before any network I/O.
 	if cached, hit, err := lookupCachedResolution(ctx, cache, key, localIssuer, redirectURI); err != nil {
-		return nil, err
+		return nil, newDCRStepError(dcrStepCacheRead, localIssuer, redirectURI, err)
 	} else if hit {
 		return cached, nil
 	}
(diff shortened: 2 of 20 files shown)


pkg/runner/runner.go
```diff
@@ -783,6 +801,25 @@ func (r *Runner) handleRemoteAuthentication(ctx context.Context) (oauth2.TokenSo
...
+// MonitoredTokenSource. Returns ("", "") when cfg is nil so callers do not
+// need to guard the call. DCR-cached credentials win over statically
+// configured ones, mirroring the precedence used by
+// remote.Handler.resolveClientCredentials at runtime.
+func remoteAuthLogContext(cfg *remote.Config) (upstream, clientID string) {
+	if cfg == nil {
+		return "", ""
+	}
+	upstream = cfg.Issuer
+	if cfg.CachedClientID != "" {
```
tgrunnagle: **F3 — Comment lies; precedence is wrong (consensus 10/10, HIGH)**

The doc above (line 804–809) says "DCR-cached credentials win over statically configured ones, mirroring the precedence used by `remote.Handler.resolveClientCredentials` at runtime." But `resolveClientCredentials` (`pkg/auth/remote/handler.go:235`) actually checks **`CachedCIMDClientID` first**, then `CachedClientID` (DCR), then static `ClientID` — see `pkg/auth/remote/handler_test.go:942` ("CachedCIMDClientID > CachedClientID (DCR) >"). This helper only handles DCR > static. For CIMD-authenticated workloads the new DCR remediation Warn will report a stale or empty client_id rather than the CIMD client_id actually being sent on token refresh, defeating the operator-correlation the field exists for.

```suggestion
	if cfg.HasCachedCIMDClientID() {
		return cfg.Issuer, cfg.CachedCIMDClientID
	}
	if cfg.CachedClientID != "" {
		return cfg.Issuer, cfg.CachedClientID
	}
	return cfg.Issuer, cfg.ClientID
```

And add a `cached CIMD client_id wins over DCR/static` case to `TestRemoteAuthLogContext` (closes F26).

*Raised by: error-handling, general-quality, tests*

pkg/authserver/runner/dcr.go
```diff
@@ -343,6 +462,113 @@ func registerAndCache(
...
+		// the URL's raw query and dropped along with the rest of the
+		// query component, mangling the error text. The trimmed
+		// punctuation is re-appended to the replacement so the
+		// surrounding prose is preserved verbatim.
+		core, tail := trimURLTrailingPunctuation(match)
+		u, parseErr := url.Parse(core)
+		if parseErr != nil || u.RawQuery == "" {
+			return match
+		}
+		u.RawQuery = ""
```
tgrunnagle, line 565: **F1 + F19 — Sanitizer leaves userinfo and fragments in the URL (consensus 9/10 + 7/10, HIGH+MEDIUM)**

`u.RawQuery = ""` clears only the query. `u.User` (which holds `user:password` for embedded-credentials URLs) is preserved, and `u.Fragment` (which OAuth implicit-flow callbacks use to carry access tokens — `#access_token=...`) is preserved too. The function's own doc comment positions it as defense-in-depth "regardless" of future flow changes; the gap is precisely what the function exists to prevent.

```suggestion
		u.User = nil
		u.RawQuery = ""
		u.Fragment = ""
```

Also add test cases to `TestSanitizeErrorForLog`: `https://user:pass@host/path?q=1` → `https://host/path` and `https://app/cb#access_token=abc` → `https://app/cb`.

*Raised by: security*

pkg/authserver/runner/dcr.go
```diff
@@ -343,6 +462,113 @@ func registerAndCache(
...
+}
+
+// queryStrippingPattern matches URL-shaped substrings inside an error
+// message — sufficient to reach the url.Parse path in sanitizeErrorForLog
+// and let it decide whether a query component exists to strip. The regexp
+// is intentionally narrow (http/https schemes only) to avoid false
+// positives. Trailing sentence punctuation that the character class
+// happens to include (e.g. '.', ',', ')') is stripped by
+// trimURLTrailingPunctuation before the match is parsed.
+var queryStrippingPattern = regexp.MustCompile(`https?://[^\s"']+`)
```
tgrunnagle: **F18 — Regex is case-sensitive; `HTTPS://example.com/?token=…` would escape sanitization (consensus 7/10, LOW)**

URL schemes are case-insensitive per RFC 3986 §3.1. Go's `*url.Error` lowercases scheme on `String()`, but log lines often originate from upstream metadata or user input that may carry mixed-case schemes. One-line fix:

```suggestion
var queryStrippingPattern = regexp.MustCompile(`(?i)https?://[^\s"']+`)
```

*Raised by: security*

pkg/authserver/runner/dcr.go
```diff
@@ -343,6 +462,113 @@ func registerAndCache(
...
+// generic "unknown" step — resolveDCRCredentials always wraps its errors,
+// so this branch indicates a programming error in a future caller rather
+// than a runtime condition.
+//
+// Every wrapped error is passed through sanitizeErrorForLog to strip URL
+// query parameters that could plausibly contain sensitive tokens (defense
+// in depth — the current DCR flow sends the initial access token as an
+// Authorization header, not a query parameter, but nothing in the type
+// system prevents a future refactor from doing otherwise).
+func LogDCRStepError(upstreamName string, err error) {
```
tgrunnagle: **F20 — `LogDCRStepError(nil)` emits a misleading "resolve failed" record (consensus 7/10, LOW)**

`errors.As(nil, &stepErr)` returns false, so the `unknown`-step branch fires and emits a fake-failure record. The current sole caller guards with `err != nil`, so this is unreachable today — but the function is package-API and the trap will outlive the guard.

```suggestion
func LogDCRStepError(upstreamName string, err error) {
	if err == nil {
		return
	}
```

*Raised by: error-handling*

pkg/authserver/runner/dcr.go
```diff
@@ -343,6 +462,113 @@ func registerAndCache(
...
+//
+// Note that ')' and ']' are stripped unconditionally — a URL legitimately
+// containing a percent-encoded closing bracket will have it as "%29" or
+// "%5D", not as a literal, so this cannot truncate a real URL path or
+// query. The reverse case (an unescaped ')' inside a path) is
+// non-conforming per RFC 3986 and out of scope for a log sanitiser.
+func trimURLTrailingPunctuation(s string) (core, tail string) {
+	const terminators = ".,;:!?)]}>"
+	i := len(s)
+	for i > 0 && strings.ContainsRune(terminators, rune(s[i-1])) {
```
tgrunnagle: **F23 — `ContainsRune` on byte-indexed input (consensus 8/10, LOW)**

The terminator set is pure ASCII (`.,;:!?)]}>`), so `strings.IndexByte(terminators, s[i-1]) >= 0` does the same thing without the rune-machinery overhead. Functionally correct as-is, but reads as if it supports Unicode terminators when it does not — a future contributor adding `」` or similar would silently mis-trim continuation bytes mid-grapheme.

```suggestion
	const terminators = ".,;:!?)]}>"
	i := len(s)
	for i > 0 && strings.IndexByte(terminators, s[i-1]) >= 0 {
		i--
	}
	return s[:i], s[i:]
```

Or keep `ContainsRune` and rename `terminators` → `asciiTerminators` with a comment that Unicode terminators are out of scope.

*Raised by: go-conventions, concurrency*

pkg/authserver/runner/dcr.go
```diff
@@ -193,6 +245,58 @@ func scopesHash(scopes []string) string {
...
+// in. The boundary caller (buildUpstreamConfigs) emits the single
+// slog.Error record for the failure; individual error branches inside
+// resolveDCRCredentials do not log so that each failure surfaces exactly
+// once in the combined log stream.
+//
+// RedirectURI is included when known so that operators can correlate the
+// failure with a specific upstream registration without parsing the
+// wrapped error string. A zero-value DCRStepError is invalid; construct
+// via newDCRStepError or the resolver's internal helpers.
+type DCRStepError struct {
```
tgrunnagle: **F13 — Exported but no external consumer (consensus 8/10, MEDIUM)**

`DCRStepError`, `LogDCRStepError`, and `newDCRStepError` are all exported, but every caller lives in `pkg/authserver/runner` itself — `LogDCRStepError` is invoked once from `embeddedauthserver.go` (same package), `newDCRStepError` only inside `dcr.go`, and `DCRStepError` is referenced via package-internal `errors.As`. Per `.claude/rules/go-style.md` "Package API Surface" — keep the export minimal until a real external consumer exists.

Lowercase to `dcrStepError` / `logDCRStepError` / `newDCRStepError` (the last is already lowercase). The constructor and helper stay package-private; the type does too.

*Raised by: architecture*

pkg/authserver/runner/dcr.go
```diff
@@ -132,36 +143,77 @@ func needsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
...
+// applyResolutionToOAuth2Config once the config has been built. Keeping the
+// two helpers side-by-side localises the DCR-specific application logic.
 func applyResolution(rc *authserver.OAuth2UpstreamRunConfig, res *DCRResolution) {
 	if rc == nil || res == nil {
 		return
 	}
 	if rc.ClientID == "" {
 		rc.ClientID = res.ClientID
 	}
+	rc.DCRConfig = nil
```
tgrunnagle, line 198: **F11 — `applyResolution` is non-idempotent state transition; the name doesn't communicate that (consensus 8/10, LOW)**

Before this PR, `applyResolution` was "fill missing fields" (idempotent). The new `rc.DCRConfig = nil` is unconditional and turns this into a state-machine transition: a second call after the first is a no-op only because the first cleared `DCRConfig`. The pin test is named `TestApplyResolution_ConsumesDCRConfig` — the function's responsibility shifted but the name didn't.

Rename to `consumeResolution`, or split into `applyResolution` (idempotent defaulting) + `markDCRResolved` (state transition that clears DCRConfig). Aligns the function name with what `TestApplyResolution_ConsumesDCRConfig` already documents.

*Raised by: architecture*

pkg/authserver/runner/dcr.go
```diff
@@ -132,36 +143,77 @@ func needsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
...
+// config model, this one writes the inline-only ClientSecret directly on
+// the runtime config.
+//
+// The split exists because buildPureOAuth2Config intentionally retains a
+// narrow file-or-env contract (no DCR awareness) and because OAuth2's
+// ClientSecret on the run-config is modelled as a reference rather than an
+// inline string. Any future output path from OAuth2UpstreamRunConfig to
+// upstream.OAuth2Config must call both helpers to get a fully-resolved
+// DCR client — exercised by TestBuildUpstreamConfigs_DCR.
+func applyResolutionToOAuth2Config(cfg *upstream.OAuth2Config, res *DCRResolution) {
```
tgrunnagle, line 225: **F12 + F25 — Paired-call invariant unenforced; doc comment names a specific test (consensus 8/10 + 7/10, MEDIUM + LOW)**

**F12**: This file now requires that every code path producing an `upstream.OAuth2Config` from a DCR resolution call BOTH `applyResolution` (run-config copy) AND `applyResolutionToOAuth2Config` (built `*OAuth2Config`). Today there's exactly one call site (`buildUpstreamConfigs`), and the doc comment names this requirement at line 210 — but the type system does not enforce it. A future second build path (e.g., live re-registration triggered by the staleness Warn) can silently forget the second call and leave `ClientSecret` empty.

Options: (a) collapse both into a single `applyDCRResolution(rcCopy, cfg, res)` that fans out internally, (b) extend `OAuth2UpstreamRunConfig` with an internal-only inline `ClientSecret` field consumed by `buildPureOAuth2Config` (eliminates the file-or-env vs inline asymmetry), or (c) add a serialization round-trip test asserting any `*upstream.OAuth2Config` produced from a `DCRResolution` carries a non-empty `ClientSecret` whenever the resolution had one.

**F25**: The trailing clause `— exercised by TestBuildUpstreamConfigs_DCR.` in the doc comment names a specific test from production code. Per `CLAUDE.md` `.claude/rules`, comments shouldn't reference specific test fixtures; if the test is renamed/split, the production comment goes stale silently.

```suggestion
// applyResolutionToOAuth2Config overlays the DCR-resolved ClientSecret onto
// a built *upstream.OAuth2Config. This is the companion to applyResolution:
// where that function writes fields representable in the file-or-env run-
// config model, this one writes the inline-only ClientSecret directly on
// the runtime config.
//
// The split exists because buildPureOAuth2Config intentionally retains a
// narrow file-or-env contract (no DCR awareness) and because OAuth2's
// ClientSecret on the run-config is modelled as a reference rather than an
// inline string. Any future output path from OAuth2UpstreamRunConfig to
// upstream.OAuth2Config must call both helpers to get a fully-resolved
// DCR client.
func applyResolutionToOAuth2Config(cfg *upstream.OAuth2Config, res *DCRResolution) {
```

*Raised by: architecture, general-quality*

pkg/authserver/runner/dcr.go
```diff
@@ -262,7 +379,8 @@ func resolveDCRCredentials(
 					"panic", fmt.Sprintf("%v", r),
 					"stack", string(debug.Stack()),
 				)
-				err = fmt.Errorf("dcr: registration panicked: %v", r)
+				err = newDCRStepError(dcrStepRegister, localIssuer, redirectURI,
```
tgrunnagle: **F4 — Singleflight panic recovery double-logs (consensus 8/10, MEDIUM)**

The PR's stated invariant is "single error log per failure." This branch breaks it: the deferred recover (line 376–386) emits `slog.Error("dcr: registration panicked", "panic", ..., "stack", ...)`, then returns a `*DCRStepError(dcrStepRegister, ...)`. The boundary caller in `buildUpstreamConfigs` then runs `LogDCRStepError` on that wrapped error and emits a second `slog.Error("dcr: resolve failed", "step", "dcr_call", ...)`. The same panic is logged twice with two different messages — exactly what this PR is meant to eliminate.

Options: (a) demote the in-defer log to `slog.Debug` for the stack trace and attach the stack as a `*DCRStepError` field so `LogDCRStepError` can carry the single Error record; (b) keep the in-defer Error and have `LogDCRStepError` skip on a known panic-step sentinel. Option (a) is closer to the existing structure.

*Raised by: error-handling*

tgrunnagle (commented):

**Agents consulted**: security, concurrency, error-handling, tests, architecture, go-conventions, general-quality, review-friendliness. Codex cross-review skipped (codex CLI not installed).

**Reviewer note**: Posting as `COMMENT` because the authenticated reviewer is the PR author and the PR is a draft. The recommendation underneath is **REQUEST CHANGES** — there are three correctness/security issues (F3, F1, F2 below) that should land before this PR is review-ready.

### Consensus Summary (29 retained findings)

| # | Finding | Score | Severity | Action |
|---|---|---|---|---|
| F3 | `remoteAuthLogContext` misses CIMD precedence; doc claim is wrong | 10/10 | HIGH | Fix |
| F1 | `sanitizeErrorForLog` does not strip URL userinfo | 9/10 | HIGH | Fix |
| F2 | Upstream `/register` error body is `io.ReadAll`-ed unbounded then logged | 8/10 | HIGH | Fix or follow-up |
| F6 | `Handler.issuer()` will panic on nil config — validation belongs in `NewHandler` | 9/10 | LOW | Fix |
| F7 | DCR success log promoted Debug→Info contradicts `silent-success-at-INFO+` rule | 9/10 | LOW | Fix or amend rule |
| F4 | Singleflight panic recovery double-logs (in-defer Error + boundary Error) | 8/10 | MEDIUM | Fix |
| F8 | Permanent-4xx Warn has no test asserting fire/contents | 8/10 | MEDIUM | Add test |
| F9 | Cache-hit `dcr_age_days` Debug + stale-threshold Warn unexercised | 8/10 | MEDIUM | Add test |
| F10 | `TestBuildUpstreamConfigs_DCR` lacks failure-path subtest | 8/10 | MEDIUM | Add test |
| F11 | `applyResolution` is non-idempotent state transition; rename it | 8/10 | LOW | Rename |
| F12 | `applyResolution` + `applyResolutionToOAuth2Config` paired-call invariant unenforced | 8/10 | MEDIUM | Collapse or document |
| F13 | `DCRStepError`, `LogDCRStepError`, `newDCRStepError` exported but no external consumer | 8/10 | MEDIUM | Lowercase |
| F23 | `trimURLTrailingPunctuation` uses `ContainsRune` on byte input | 8/10 | LOW | `IndexByte` instead |
| F19 | Sanitizer doesn't strip URL fragments | 7/10 | MEDIUM | One-line fix |
| F5 | Permanent-4xx Warn risk of log-spam from `Token()` entry | 7/10 | MEDIUM | Hoist out of classifier |
| F14 | `dcr.go` is 1,186 lines; sanitizer is independent of DCR | 7/10 | MEDIUM | Split file |
| F15 | New private helpers placed above the `// Private helpers` divider | 7/10 | MEDIUM | Reorder |
| F16 | PR exceeds 400-LOC / 10-file limit; lacks `Large PR Justification` | 7/10 | MEDIUM | Add section |
| F18 | Sanitizer regex case-sensitive (`HTTPS://` escapes) | 7/10 | LOW | `(?i)` flag |
| F20 | `LogDCRStepError(nil)` emits misleading record | 7/10 | LOW | Nil guard |
| F21 | `dcrStore` lifetime / interface lacks `Close` | 7/10 | LOW | Document |
| F22 | Process-wide `dcrFlight` vs per-instance `dcrStore` undocumented | 7/10 | LOW | Document |
| F24 | Production comments cite `.claude/rules/...` paths | 7/10 | LOW | Inline principle |
| F25 | Production comment names a specific test | 7/10 | LOW | Drop ref |
| F26 | `TestRemoteAuthLogContext` doesn't cover CIMD path (ties to F3) | 8/10 | MEDIUM | Add case w/ F3 |
| F27 | `TestHandler_issuer` doesn't assert nil-config invariant | 7/10 | LOW | Add assertion |
| F28 | Mock-AS request count uses `Greater(0)` not exact | 7/10 | LOW | Tighten |
| F29 | Panic-recovery test doesn't `errors.As(*DCRStepError)` | 7/10 | LOW | Augment |
| F30 | Step-wrap tests use substring instead of `errors.As` | 7/10 | LOW | Augment |

### Overall

The PR's structural shape is sound. The wiring of `resolveDCRCredentials` into `EmbeddedAuthServer.NewEmbeddedAuthServer` correctly observes the copy-before-mutate rule (deep-copying both the outer `UpstreamRunConfig` and the inner `*OAuth2UpstreamRunConfig` before passing to the resolver), and `TestBuildUpstreamConfigs_DCR` + `TestNewEmbeddedAuthServer_DCRBoot` pin those guarantees explicitly via `assert.Same` on the caller's pointer. The `DCRStepError` step taxonomy + boundary `LogDCRStepError` is a genuinely good observability pattern for a multi-step resolver. The new sanitizer's trailing-punctuation handling is well-tested with regression cases.

However, three correctness/security items should land before merge:

1. **F3 (HIGH, score 10/10)** — `remoteAuthLogContext` claims to mirror the precedence of `pkg/auth/remote/handler.go::resolveClientCredentials`, but the actual precedence in that function is **CIMD > DCR > static**, while the new helper handles only **DCR > static**. Workloads using cached CIMD client_ids will see a stale or empty `client_id` in the new DCR remediation Warn — defeating the operator-correlation the field exists for. Either fix the precedence or rewrite the comment.
2. **F1 (HIGH, score 9/10)** — `sanitizeErrorForLog` only clears `u.RawQuery`. URLs with embedded credentials (`https://user:pass@host`) have their userinfo preserved; URLs with fragments (e.g., implicit-flow callbacks `#access_token=...`) keep them too (F19). The function is documented as defense-in-depth, so the gap is precisely what the function exists to prevent. Two-line fix in the same block: `u.User = nil; u.Fragment = ""`.
3. **F2 (HIGH, score 8/10)** — `pkg/oauthproto/dcr.go:307` reads the upstream's error body via `io.ReadAll(resp.Body)` with **no** `LimitReader` (the success path correctly caps at 1 MiB), then embeds the body verbatim in `fmt.Errorf("...status %d: %s", code, body)`. That error flows into `LogDCRStepError`, which only sanitizes URL query strings. A misbehaving or compromised upstream can echo arbitrary content into operator logs unbounded. The bug is technically in `oauthproto/dcr.go` (outside this PR's diff), but this PR is what surfaces the risk operationally — worth either fixing in this PR or filing a follow-up that blocks merge.

Beyond the HIGH items, the test-coverage cluster (F8, F9, F10, F26, F27–F30) is the next-most-important bucket: the PR's central observability contributions (the 4xx Warn, the staleness Warn, the boundary Error log) are **not asserted** by any test. A regression that drops or changes those records would not fail CI. Adding a slog-handler-tee fixture and exercising the three branches would close the gap.

The architecture cluster (F11–F13) is polish but worth considering before merge: the API surface widening (exported `DCRStepError`, etc.) and the paired-helpers contract (`applyResolution` + `applyResolutionToOAuth2Config`) are both maintenance hazards if left as comment-enforced invariants.

F14–F16 are review-experience findings; the github-actions bot has already left a `CHANGES_REQUESTED` for the >1000-line size, and adding a `Large PR Justification` section addresses that mechanically.

### Notes for items that don't have an inline comment

- **F2** (unbounded upstream error body) — the affected file `pkg/oauthproto/dcr.go` is not in this PR's diff. Fix at `pkg/oauthproto/dcr.go:307`: replace `io.ReadAll(resp.Body)` with `io.ReadAll(io.LimitReader(resp.Body, 8*1024))` and consider parsing as RFC 7591 JSON instead of `%s`-interpolating the raw body.
- **F8/F9/F10/F26–F30** (test gaps) — see the per-finding lines in the consensus table; rather than 7 tiny inline comments, group these into one PR-level test plan: (1) capture slog records via a tee handler in monitored_token_source_test.go, (2) drive the cache-hit + stale-threshold path in dcr_test.go with a pre-populated old `CreatedAt`, (3) add a `failure-path` subtest to `TestBuildUpstreamConfigs_DCR` with a 500 mock, (4) add a CIMD case to `TestRemoteAuthLogContext`, (5) augment the existing wrapped-error tests with `errors.As(*DCRStepError)` assertions.
- **F14** (dcr.go file size) — split `sanitizeErrorForLog`, `trimURLTrailingPunctuation`, `queryStrippingPattern`, and `TestSanitizeErrorForLog` into a new `pkg/authserver/runner/logsanitize.go` and `logsanitize_test.go`. Optionally also split `DCRStepError` + `LogDCRStepError` + the `dcrStep*` constants into `pkg/authserver/runner/dcr_errors.go`.
- **F15** (file org) — move `newDCRStepError`, `sanitizeErrorForLog`, `trimURLTrailingPunctuation`, and `queryStrippingPattern` below the `// ----- Private helpers -----` divider in `dcr.go` (line ~573).
- **F16** (PR meta) — add a `## Large PR Justification` section to the PR description: ~620 of the 1133 additions are tests, the remaining ~510 across 7 production files implement one logical change, and the four sub-concerns (resolver wiring, step taxonomy, monitored-token-source remediation log, software_id validation) cannot be split without orphaned-code intermediate states. Alternatively, run `/split-pr` to evaluate carving out the URL sanitizer + tests as a standalone PR.
- **F22** (`dcrFlight` lifetime) — the var declaration is package-level pre-existing code (not in any hunk). Add a doc comment explaining the intentional process-wide lifetime, contrasting with the per-instance `dcrStore`.
- **F26** is structurally identical to F3; fixing F3 closes both.

---

pkg/auth/monitored_token_source.go
```diff
@@ -365,6 +403,20 @@ func isTransientNetworkError(err error) bool {
...
 			return true
 		}
+		// Permanent 4xx from the token endpoint frequently indicates that the
+		// cached Dynamic Client Registration has been revoked, rotated, or is
+		// no longer recognised by the authorization server. Emit a remediation
+		// hint before returning false so operators can correlate the
+		// unauthentication with a stale DCR rather than a user-action issue.
+		// The returned boolean is unchanged.
+		//nolint:gosec // G706: client_id is public metadata per RFC 7591.
+		slog.Warn(
```
tgrunnagle: **F5 — Warn fires from `Token()` entry path on every call (consensus 7/10, MEDIUM)**

`transientRefresher.retry` calls `isTransientNetworkError` only at the end of the backoff loop returning false (which surfaces as `backoff.Permanent`), so retry is well-behaved — Warns at most once per refresh. But `MonitoredTokenSource.Token()` (line 298 of this file) also calls `isTransientNetworkError` on the very first try before falling into the unauth-marking path. If a workload's caller retries `Token()` in a tight loop (e.g., a request-driven path that sees the workload as still authenticated for a brief window before the status update propagates), this Warn fires on every single Token() call that hits a permanent 4xx.

The shape itself is also a smell: a classifier function (`is X transient?`) is doing a side-effect (`emit a remediation Warn`). Hoist the Warn out of `isTransientNetworkError` into the call sites that have visibility into retry context — emit once from `MonitoredTokenSource.Token()` right before `markAsUnauthenticated`, and once from `transientRefresher.retry` at the same boundary where it returns `backoff.Permanent`. Or add a per-(workload, client_id) once-per-state-transition guard.

Also: the message asserts "cached DCR client is no longer recognized" but the same branch fires for any non-transient 4xx (revoked consent, disabled account, statically configured client). Workloads not using DCR will still see this remediation telling operators to "delete cached credentials" — wrong remediation. Either gate on `clientID != ""` or reword conditionally.

*Raised by: error-handling, security, review-friendliness*

pkg/authserver/server/handlers/dcr.go
```diff
@@ -95,10 +95,30 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
...
-	)
+		"software_id", validated.SoftwareID,
+		"token_endpoint_auth_method", validated.TokenEndpointAuthMethod,
+		"scopes", scopes,
+	}
+	if issuer := h.issuer(); issuer != "" {
+		logAttrs = append(logAttrs, "issuer", issuer)
+	}
+	//nolint:gosec // G706: client_id is public metadata per RFC 7591.
+	slog.Info("registered new DCR client", logAttrs...)
```
tgrunnagle: **F7 — Promotion to `slog.Info` contradicts `silent-success-at-INFO+` rule (consensus 9/10, LOW)**

`.claude/rules/go-style.md` §Logging is explicit: "Silent success — no output at INFO or above for successful operations" and "INFO sparingly — only for long-running operations like image pulls." A successful DCR registration is neither long-running nor exceptional. The inline comment justifies as audit-trail signal, which is reasonable, but it unilaterally re-defines the rule for one call site.

Either (a) keep at `Debug` and move the audit concern to a dedicated audit-log emission path, or (b) update `.claude/rules/go-style.md` first to carve out an explicit security-event exception. Documenting the decision in the rules file is the more durable fix; otherwise this gets reverted in a future cleanup pass.

*Raised by: go-conventions, error-handling, review-friendliness*

pkg/authserver/server/handlers/handler.go
```diff
@@ -136,3 +136,17 @@ func (h *Handler) upstreamByName(name string) (upstream.OAuth2Provider, bool) {
...
+// not exercise this path (no separate constructor validation for the
+// embedded Config), but every handler that calls issuer() reaches it only
+// after the handler has been fully wired via a working
+// AuthorizationServerConfig. Tests that exercise issuer()-emitting
+// handlers must therefore supply a valid *fosite.Config (even if its
+// fields are zero-valued). Returning a silent default on a nil config
+// would hide real wiring bugs — see .claude/rules/go-style.md
+// §"Constructor Validation: Fail Loudly on Invalid Input".
+func (h *Handler) issuer() string {
+	return h.config.AccessTokenIssuer
```
tgrunnagle, line 151: **F6 — Validation belongs in the constructor per the cited rule (consensus 9/10, LOW)**

The doc comment correctly cites `.claude/rules/go-style.md §"Constructor Validation: Fail Loudly on Invalid Input"`, but per that rule the right place to fail loudly is **in the constructor**, not as a panic at request time. `NewHandler` (`handler.go:51`) does not validate `config != nil` or `config.Config != nil`, so a misconfigured caller panics inside an HTTP handler. The PR's own test diff is evidence the trap is reachable: every `AuthorizationServerConfig` literal in `dcr_test.go` had to add `Config: &fosite.Config{...}` to avoid the panic.

```suggestion
func NewHandler(
	provider fosite.OAuth2Provider,
	config *server.AuthorizationServerConfig,
	stor storage.Storage,
	upstreams []NamedUpstream,
) (*Handler, error) {
	if config == nil || config.Config == nil {
		return nil, fmt.Errorf("handlers: AuthorizationServerConfig with embedded *fosite.Config must be non-nil")
	}
	if len(upstreams) == 0 {
```

(Then add this guard on top of `NewHandler` rather than mutating `issuer()`.) Per the cited rule: "Misconfiguration that fails at startup is far easier to diagnose than misconfiguration that silently degrades behavior at runtime."

*Raised by: security, error-handling, go-conventions*

pkg/authserver/server/handlers/handler.go
```diff
@@ -136,3 +136,17 @@ func (h *Handler) upstreamByName(name string) (upstream.OAuth2Provider, bool) {
...
+
+// issuer returns the authorization-server issuer URL. Both h.config and
+// the embedded *fosite.Config are required to be non-nil — NewHandler does
+// not exercise this path (no separate constructor validation for the
+// embedded Config), but every handler that calls issuer() reaches it only
+// after the handler has been fully wired via a working
+// AuthorizationServerConfig. Tests that exercise issuer()-emitting
+// handlers must therefore supply a valid *fosite.Config (even if its
+// fields are zero-valued). Returning a silent default on a nil config
+// would hide real wiring bugs — see .claude/rules/go-style.md
```
tgrunnagle: **F24 — Production comment cites `.claude/rules/...` path (consensus 7/10, LOW)**

Production comments citing the rules tree by path create a refactor hazard: if the rule section is renamed, every cite goes stale silently. They also leak working-environment vocabulary into the codebase. The rationale itself (`silent default would hide real wiring bugs`) is the value — keep it inline, drop the file-path citation. Same pattern at `embeddedauthserver.go` `buildUpstreamConfigs` doc and `dcr.go` `applyResolution` doc.

*Raised by: general-quality*

pkg/authserver/runner/embeddedauthserver.go
```diff
@@ -39,8 +39,12 @@ const (
 type EmbeddedAuthServer struct {
 	server      authserver.Server
 	keyProvider keys.KeyProvider
-	closeOnce   sync.Once
-	closeErr    error
+	// dcrStore caches RFC 7591 Dynamic Client Registration resolutions across
+	// calls to buildUpstreamConfigs so that re-entrant boot/reload paths reuse
+	// previously-registered upstream clients instead of re-registering.
+	dcrStore  DCRCredentialStore
```
tgrunnagle, line 60: **F21 + F22 — `dcrStore` lifecycle / `dcrFlight` lifetime documentation gaps (consensus 7/10 each, LOW)**

**F21**: For the in-memory store this is benign — the map is GC'd when `EmbeddedAuthServer` is unreferenced. But `DCRCredentialStore` is an interface with no `Close` method, so a future Redis- or sqlite-backed implementation (Phase 3) cannot release handles cleanly without a breaking interface change. Add a comment to `EmbeddedAuthServer.Close` noting `dcrStore` intentionally does not require release for the in-memory implementation, and that Phase 3 must extend `Close` when a backend with handles is introduced.

**F22**: The package-level `dcrFlight` (`dcr.go`) is **process-wide** while `dcrStore` is **per-instance**. The cache-recheck-after-singleflight-entry pattern depends on this asymmetry but the rationale is undocumented. Worth a one-line comment on `dcrFlight` itself (or here on `dcrStore`) explaining the intentional lifetime mismatch — the singleflight only deduplicates the network call, not the cache state.

*Raised by: concurrency*

pkg/runner/runner.go
tgrunnagle: Addressed in f60b6574 — remoteAuthLogContext now mirrors the documented CIMD > DCR > static precedence, and TestRemoteAuthLogContext gained a CIMD-wins case (closes F26).

pkg/authserver/runner/dcr.go
tgrunnagle: Addressed in 5f1d5ffb — sanitizeErrorForLog now clears `u.User` and `u.Fragment` in addition to `u.RawQuery`, with the early-return condition updated to fall through whenever any secret-bearing component is present. New TestSanitizeErrorForLog cases cover embedded userinfo (with and without query) and implicit-flow fragments.

pkg/authserver/runner/dcr.go
tgrunnagle: Addressed in 5f1d5ffb — queryStrippingPattern now uses the `(?i)` flag, and a TestSanitizeErrorForLog case pins that an uppercase `HTTPS://` URL is sanitised. (url.Parse normalises the scheme to lowercase on the way out, so the post-sanitisation form matches the canonical casing.)

pkg/authserver/runner/dcr.go
tgrunnagle: Addressed in 5f1d5ffb — logDCRStepError (now lowercased per F13) returns early on `err == nil` so the unknown-step branch can no longer emit a fake-failure record.

pkg/authserver/runner/dcr.go
tgrunnagle: Addressed in 5f1d5ffb — trimURLTrailingPunctuation switched to `strings.IndexByte`, with a comment noting that the terminator set is intentionally ASCII-only and Unicode terminators are out of scope.

pkg/authserver/runner/dcr.go
tgrunnagle: Addressed in 5f1d5ffb — `DCRStepError`, `LogDCRStepError`, and `newDCRStepError` are now lowercased to `dcrStepError`, `logDCRStepError`, and `newDCRStepError` (the constructor was already lowercase). All callers and tests updated; tests now use `var got *dcrStepError`.

pkg/authserver/runner/dcr.go
tgrunnagle: Addressed in 5f1d5ffb — renamed to `consumeResolution`, with a doc comment stating that the unconditional `DCRConfig = nil` makes this a one-shot state transition rather than an idempotent default-fill. Tests renamed in step (`TestConsumeResolution_*`).

pkg/authserver/runner/dcr.go
tgrunnagle: Addressed in 5f1d5ffb — applyResolutionToOAuth2Config's doc comment now states the paired-call invariant explicitly ("any future output path … must call BOTH consumeResolution AND applyResolutionToOAuth2Config") without referencing TestBuildUpstreamConfigs_DCR by name. Went with option (a)/(b) hybrid for the doc; the type-system enforcement option (b) is a larger refactor I left for a follow-up.

pkg/authserver/runner/dcr.go
tgrunnagle: Addressed in 5f1d5ffb — went with option (a). The deferred recover no longer emits a slog.Error; instead the captured `debug.Stack()` is attached to a new `Stack` field on `dcrStepError`, and `logDCRStepError` includes it as a `stack` attribute on the single boundary record. The panic-recovery test now also asserts `errors.As(*dcrStepError)` and that `Stack` is non-empty.

pkg/auth/monitored_token_source.go
tgrunnagle: Addressed in 1488e9ff — `isTransientNetworkError` is now side-effect free. The DCR/CIMD remediation Warn moved into `markAsUnauthenticated`, gated by the existing `stopOnce` so a tight Token() loop cannot spam it. Workloads with empty `clientID` get the log suppressed entirely (the operator-correlation field would be blank), and the message is reworded to be honest about the variability ("if this workload uses cached DCR or CIMD credentials they may be stale").

pkg/authserver/server/handlers/dcr.go
tgrunnagle: Addressed in ca3bb82e — went with option (a). The DCR success log is back at `slog.Debug`. The doc comment now notes that an audit-log emission path is the right home for the audit signal if that becomes a requirement, rather than promoting one site to INFO.

pkg/authserver/server/handlers/handler.go
tgrunnagle: Addressed in ca3bb82e — `NewHandler` now returns an error when `config == nil || config.Config == nil`, and `issuer()` was simplified to rely on the constructor invariant. New `TestNewHandler_ErrorsOnNilConfig` pins both the nil-config and nil-embedded-Config cases.

pkg/authserver/server/handlers/handler.go
tgrunnagle: Addressed in 5f1d5ffb (and ca3bb82e for handler.go) — the `.claude/rules/...` path citations were dropped from the production comments in `dcr.go`, `dcr_store.go`, `embeddedauthserver.go`, and `handler.go`. Each rationale is now inlined (e.g. "failing loudly at the boundary makes such bugs visible at the first call").

pkg/authserver/runner/embeddedauthserver.go
tgrunnagle: Addressed in 5f1d5ffb — the `dcrStore` field comment in `EmbeddedAuthServer` now explains the per-instance vs process-wide split versus `dcrFlight`, and notes that `DCRCredentialStore` lacks a `Close` method only because the in-memory implementation needs no release; a future Phase-3 backend will have to extend the interface and `EmbeddedAuthServer.Close`. The `dcrFlight` doc was updated symmetrically with the load-bearing rationale for the asymmetry.

tgrunnagle:
Body-only review items follow-up:

**F2 (HIGH)** — Addressed in 4d504fe9. `pkg/oauthproto/dcr.go` was outside this PR's diff; the unbounded `io.ReadAll` on the upstream's `/register` error body is now wrapped in `io.LimitReader(resp.Body, 8*1024)`, so a non-conformant upstream cannot inflate operator log volume. I kept the existing `%s`-into-`fmt.Errorf` shape rather than parsing as RFC 7591 JSON; the cap is the load-bearing fix here.

**F8 / F9 / F10 / F26-F30 (test-gap cluster)** — Deferred. F26 was closed mechanically by the new CIMD case in `TestRemoteAuthLogContext` that landed with the F3 fix. The remaining items (slog-handler tee, failure-path subtest, panic-recovery `errors.As(*dcrStepError)` augmentation, etc.) I'd like to land as a follow-up so this PR stays focused on the correctness/security fixes. Happy to open a tracking issue if useful.

pkg/auth/monitored_token_source.go
```diff
@@ -399,6 +443,21 @@ func isTransientNetworkError(err error) bool {
...
+// emit a DCR/CIMD remediation hint alongside the unauthentication.
+func isPermanentTokenEndpointError(err error) bool {
+	retrieveErr, ok := errors.AsType[*oauth2.RetrieveError](err)
+	if !ok {
+		return false
+	}
+	if retrieveErr.Response == nil {
+		return true
+	}
+	return retrieveErr.Response.StatusCode < 500
```
jhrozek: I traced this and it only affects the DCR remediation Warn — retry behavior is gated entirely by `isTransientNetworkError` above, so flipping the verdict here doesn't change retries. But 408 (Request Timeout) and 429 (Too Many Requests) end up classified as permanent here, which means the "delete the cached credentials and restart" Warn fires on what's typically transient back-pressure. Worth narrowing to an explicit allowlist (400/401/403) so the hint doesn't mislead operators chasing a transient hiccup.

pkg/oauthproto/dcr.go
```diff
@@ -303,8 +303,13 @@ func handleHTTPResponse(resp *http.Response) (*DynamicClientRegistrationResponse
...
 	if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusOK {
-		// Try to read error response
-		errorBody, _ := io.ReadAll(resp.Body)
+		// Cap the upstream's error body so a misbehaving or malicious AS
+		// cannot inflate operator logs by echoing arbitrary content. 8 KiB
+		// is far larger than any conformant RFC 7591 error response and
+		// still small enough that the surrounding error chain stays
+		// readable.
+		const maxErrorBodySize = 8 * 1024
+		errorBody, _ := io.ReadAll(io.LimitReader(resp.Body, maxErrorBodySize))
```
jhrozek, line 312: This `io.ReadAll(io.LimitReader(...))` leaves the body undrained when the upstream returns more than 8 KiB, and the deferred `Close` then prevents the connection from being reused. The success branch a few lines down (the non-JSON content-type case) already does `io.Copy(io.Discard, resp.Body)` for exactly this reason, and `.claude/rules/go-style.md` spells out the rule. Easiest fix is one extra line right after the `LimitReader` read.

pkg/runner/runner.go
```diff
@@ -783,6 +801,28 @@ func (r *Runner) handleRemoteAuthentication(ctx context.Context) (oauth2.TokenSo
 	return tokenSource, nil
 }
 
+// remoteAuthLogContext extracts the upstream issuer and resolved client_id
+// from a remote auth config for use as log-correlation fields on the
+// MonitoredTokenSource. Returns ("", "") when cfg is nil so callers do not
+// need to guard the call. Mirrors the precedence applied at runtime when
+// sending the client_id on token refresh: cached CIMD URL > cached DCR
+// client_id > statically configured client_id.
+func remoteAuthLogContext(cfg *remote.Config) (upstream, clientID string) {
```
jhrozek: Quick question rather than a fix request — does this want to live in `pkg/auth/remote` (e.g. as `Config.LogContext()`) instead? It only reads `remote.Config` fields, and the precedence (cached CIMD > cached DCR > static) is the same one `resolveClientCredentials` uses. Two copies of the precedence ordering in different packages will drift if someone ever adds a fourth cached field. Not blocking, just a packaging nit.

jhrozek (commented):
A few notes — none blocking.

pkg/auth/monitored_token_source.go
tgrunnagle: Addressed in 0073ad77 — `isPermanentTokenEndpointError` now uses an explicit 400 / 401 / 403 allowlist. 408 and 429 (and any other non-5xx that doesn't suggest stale credentials) no longer fire the remediation Warn. Retry behaviour is unchanged because `isTransientNetworkError` was already returning false for the whole RetrieveError branch.

pkg/oauthproto/dcr.go
tgrunnagle: Addressed in 0073ad77 — added `io.Copy(io.Discard, resp.Body)` immediately after the LimitReader-bounded ReadAll, with a comment pointing at the connection-reuse rationale. Mirrors the existing pattern in the non-JSON content-type branch.

pkg/runner/runner.go
tgrunnagle: Good call — addressed in 0073ad77. Moved to `pkg/auth/remote/Config.LogContext()` so the precedence has a single home next to `resolveClientCredentials`. Tests moved to `pkg/auth/remote/config_test.go`; the `pkg/runner` caller is now `r.Config.RemoteAuthConfig.LogContext()`.

jhrozek (approved):
Verified all three follow-ups landed in 0073ad77 and look right — the classifier allowlist, the body drain, and the LogContext relocation. Thanks for the quick turnaround.

PR #5052, opened 2026-04-24
pkg/llm: fix withPreemptiveRefresh composition and stale _AT on rotation

### LINKED ISSUE #5051: pkg/llm: fix withPreemptiveRefresh composition and stale _AT cache on refresh token rotation
Two issues flagged in PR review by @aponcedeleonch on the `thv-llm-token` branch.

---

## 1. `withPreemptiveRefresh` composition is broken

**File:** `pkg/llm/tokensource.go`

Both callers of `withPreemptiveRefresh` pass an `src` that is already an `oauth2.ReuseTokenSource`:
- `flow.TokenSource()` returns `oauth2.ReuseTokenSource(token, base)` (flow.go:511)
- `remote.CreateTokenSourceFromCached` returns `oauth2.ReuseTokenSource(token, base)` (persisting_token_source.go:104), then wrapped in `PersistingTokenSource`

The current stacking is:
```
outer ReuseTokenSource(nil, preemptiveTokenSource{inner: PersistingTokenSource{source: inner ReuseTokenSource{source: rawRefresher}}})
```

The bug: when the outer `ReuseTokenSource`'s cached token hits the shifted expiry, it calls `preemptiveTokenSource`, which calls the inner chain. The inner `ReuseTokenSource` sees the real expiry is still valid and returns the same cached token unchanged. `preemptiveTokenSource` shifts the expiry again and returns the same shifted value. The outer caches it — but it immediately appears expired on the next call because `time.Now()` has already crossed `real_expiry - 40s`. Every `Token()` call inside the 30-second preemptive window triggers a full inner round-trip instead of one refresh per window.

**Fix:** wrap the raw refresher with `preemptiveTokenSource` before `oauth2.ReuseTokenSource`, not after. Target stacking:
```
ReuseTokenSource(nil, preemptiveTokenSource{inner: PersistingTokenSource{source: rawRefresher}})
```

Requires:
- In `tryRestoreFromCache`: build the raw token source directly instead of using `CreateTokenSourceFromCached` (which adds an inner `ReuseTokenSource`)
- In `performBrowserFlow`: build raw token source from `flowCfg` fields directly instead of calling `flow.TokenSource()` (which is a `ReuseTokenSource`)
- Note: `oauth2.Config.TokenSource` also returns a `reuseTokenSource` internally, so `NewResourceTokenSource` (which does its own validity check + stores `token` field) is the correct stateless refresher for the resource path
- Add a regression test with a fake refresher that counts calls, verifying that multiple `Token()` calls within the preemptive window produce exactly one inner `Token()` call

---

## 2. Stale `_AT` cache entry after refresh token rotation

**File:** `pkg/llm/tokensource.go` — `makeTokenPersister`

When the persister callback fires (a rotated refresh token is written to the secrets store), the associated `_AT` access token cache entry is left unchanged. Between that write and the next `Token()` call, tier 1.5 could serve an access token that was minted against the now-invalidated refresh token. The preemptive window reduces the window in practice, but it is not zero.

**Fix:** inside `makeTokenPersister`, after `SetSecret` succeeds, overwrite or delete the `_AT` entry so the next `Token()` call falls through to tier 2 and gets a fresh access token:

```go
// Invalidate the access-token cache so the next call does not serve a token
// minted against the now-rotated refresh token.
_ = t.secretsProvider.SetSecret(ctx, t.accessTokenCacheKey(), "")
```

Or delete it if the provider supports no-op deletes for missing keys.

---

Both issues are on `pkg/llm/tokensource.go` introduced in the `thv-llm-token` branch / PR.

yrobla:
## Summary

Two bugs in `pkg/llm/tokensource.go` flagged in PR review on the `thv-llm-token` branch (#5051).

### Bug 1: `withPreemptiveRefresh` composition is broken

Both callers passed a source already wrapped in `oauth2.ReuseTokenSource` (from `oauth.NewResourceTokenSource` and `oauth2Cfg.TokenSource`), creating a double-caching chain. In the preemptive window the inner `ReuseTokenSource` returned the same cached real-expiry token, `preemptiveTokenSource` shifted it to already-expired, and every `Token()` call re-entered the inner chain indefinitely — no actual IdP refresh occurred.

**Fix:** replace both inner sources with `oauth.NonCachingRefresher` (in `pkg/auth/oauth`), a type that always performs a network round-trip when `Token()` is called, on **both** the resource path (`Audience != ""`) and the standard path (`Audience == ""`). The outer `ReuseTokenSource` (via `withPreemptiveRefresh`) is now the sole caching layer. Target stacking:

```
ReuseTokenSource(nil, preemptiveTokenSource{inner: PersistingTokenSource{source: NonCachingRefresher}})
```

`NonCachingRefresher` consolidates the RFC 8707 resource-indicator logic and standard OAuth 2.0 refresh into one type (branching on `resource == ""`), is safe for concurrent use via an internal `sync.Mutex`, and preserves the previous refresh token when the IdP does not rotate one.

`resourceTokenSource` in `pkg/auth/oauth` is refactored to delegate to `NonCachingRefresher` instead of duplicating the `cfg.Exchange` logic — the two shared the same RFC 8707 resource handling and explicit scope forwarding.

`withPreemptiveRefreshFrom` is added to pre-seed the outer `ReuseTokenSource` with the shifted initial token after the browser flow, so the just-obtained access token is served without an immediate network call.

Also extracts an `oauth2ConfigFrom` helper to reuse the already-discovered `flowCfg` in `performBrowserFlow` without a second OIDC round-trip.

### Bug 2: Stale `_AT` cache after refresh token rotation

`makeTokenPersister` wrote the new refresh token but left the `_AT` access-token cache entry unchanged. A concurrent process (e.g. `thv llm token`) could read the stale `_AT` entry and serve a token minted against the now-invalidated refresh token.

**Fix:** after `SetSecret` for the refresh token succeeds, overwrite the `_AT` entry with an empty string so the next `Token()` call falls through to a fresh access token. Errors from this invalidation are now logged at `Warn` instead of silently discarded.

Closes #5051

## Follow-up issues

- #5066 — Lift `ensureOfflineAccess` and `DeriveSecretKey` to `pkg/auth/oauth` (byte-identical with counterparts in `pkg/registry/auth`) and consider retry/singleflight for `pkg/llm` token refresh (parity with `MonitoredTokenSource`)

## Type of change

- [x] Bug fix

## Test plan

Added:
- `TestWithPreemptiveRefresh_ExactlyOneRefreshPerWindow`: verifies exactly 2 inner calls (initial + one preemptive refresh), all subsequent calls from outer cache — exercises the resource path (`Audience != ""`)
- `TestWithPreemptiveRefresh_NoResource_ExactlyOneRefreshPerWindow`: same assertion via a real HTTP test server and `NonCachingRefresher` with `resource = ""` — exercises the standard (non-resource) path and catches a revert to `oauth2Cfg.TokenSource`
- `TestWithPreemptiveRefreshFrom_PreSeededToken`: verifies the initial token is served without hitting the inner source
- `TestWithPreemptiveRefreshFrom_NilInitial_BehavesLikeWithPreemptiveRefresh`
- `TestNonCachingRefresher_EmptyRefreshToken_ReturnsError`
- `TestNonCachingRefresher_StandardPath_NoResourceParam`: exercises the standard path against a mock HTTP server, verifying no `resource=` param is sent and the previous RT is preserved when the IdP omits one
- `TestWithPreemptiveRefresh_CachingInnerSource_ThrashesAcrossOuterCalls`: regression test documenting the failure mode (N outer `Token()` calls → N inner calls; contrast with the fixed behaviour above)

Updated `TestTokenSource_MakeTokenPersister_StoresTokenAndCallsUpdater` to expect the AT cache invalidation write in order after the refresh token write, and to align the refresh-token key with the AT key via `CachedRefreshTokenRef`.

No — internal token caching logic. Users see fewer redundant network refreshes in the preemptive window and no stale access tokens after refresh token rotation.

- The `_AT` invalidation uses `SetSecret(..., "")` rather than `DeleteSecret` for provider compatibility (some providers do not support delete; an empty value is treated as a cache miss by `tryAccessTokenCache`).
- `NonCachingRefresher` holds a `sync.Mutex` so it is self-safe for concurrent use. It lives in `pkg/auth/oauth` and is now used by both `pkg/llm` and `resourceTokenSource`, eliminating the duplicate `cfg.Exchange` logic.
- The non-resource path (`Audience == ""`) is fully fixed by this PR. The stale hedge in earlier description drafts ("non-resource path still partially broken") was incorrect — `NonCachingRefresher` handles both paths unconditionally.

### DIFF
--- a/pkg/llm/tokensource.go
+++ b/pkg/llm/tokensource.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"net/http"
 	"strings"
 	"sync"
 	"time"
@@ -191,24 +192,24 @@ func (t *TokenSource) tryRestoreFromCache(ctx context.Context) error {
 		return fmt.Errorf("building oauth2 config for cache restore: %w", err)
 	}
 
-	// Build the raw refresher directly without an inner ReuseTokenSource.
-	// CreateTokenSourceFromCached wraps in ReuseTokenSource, which causes the
-	// preemptive window to loop: the inner source returns the cached real token,
-	// preemptiveTokenSource shifts it back, the outer ReuseTokenSource caches an
-	// already-expired shifted token, and every subsequent Token() call re-enters
-	// the inner chain instead of one refresh per window.
-	// Target stacking: ReuseTokenSource(nil, preemptive{PersistingTokenSource{rawRefresher}})
-	rawToken := &oauth2.Token{
-		RefreshToken: refreshToken,
-		Expiry:       t.cfg.OIDC.CachedTokenExpiry,
-		TokenType:    "Bearer",
-	}
-	var rawRefresher oauth2.TokenSource
-	if t.cfg.OIDC.Audience != "" {
-		rawRefresher = oauth.NewResourceTokenSource(oauth2Cfg, rawToken, t.cfg.OIDC.Audience)
-	} else {
-		rawRefresher = oauth2Cfg.TokenSource(ctx, rawToken)
-	}
+	// Use a non-caching refresher as the innermost source.
+	//
+	// oauth.NewResourceTokenSource and oauth2Cfg.TokenSource both cache the token
+	// internally and only refresh when the real expiry passes. When the outer
+	// ReuseTokenSource enters the preemptive window (30 s before real expiry) it
+	// calls preemptiveTokenSource, which calls the inner source; the inner source
+	// sees the real token as still valid and returns it unchanged;
+	// preemptiveTokenSource shifts the expiry back by 30 s, producing an
+	// already-expired token; the outer ReuseTokenSource then re-enters the chain
+	// on every subsequent call — an infinite non-refreshing loop.
+	//
+	// A non-caching refresher always performs a network round-trip when called, so
+	// the first call inside the preemptive window obtains a fresh token with a
+	// real-future expiry. The outer ReuseTokenSource caches the shifted result and
+	// serves it until the next window — exactly one refresh per window.
+	//
+	// Target stacking: ReuseTokenSource(nil, preemptive{PersistingTokenSource{nonCachingRefresher}})
+	rawRefresher := newNonCachingRefresher(oauth2Cfg, refreshToken, t.cfg.OIDC.Audience)
 
 	// Persist rotated refresh tokens back to the secrets provider so future CLI
 	// invocations can still restore the session if the provider invalidates the
@@ -239,10 +240,9 @@ func (t *TokenSource) performBrowserFlow(ctx context.Context) error {
 		return fmt.Errorf("OAuth flow start failed: %w", err)
 	}
 
-	// Build raw token source from flowCfg fields directly instead of calling
-	// flow.TokenSource(), which returns a ReuseTokenSource and causes the same
-	// double-caching composition issue as CreateTokenSourceFromCached (see
-	// tryRestoreFromCache for a detailed explanation).
+	// Build a non-caching refresher as the innermost source (see tryRestoreFromCache
+	// for a detailed explanation of why caching inner sources cause an infinite loop
+	// inside the preemptive window).
 	// Reuse the already-discovered flowCfg to avoid a second OIDC round-trip.
 	oauth2Cfg := oauth2ConfigFrom(flowCfg)
 	initialToken := &oauth2.Token{
@@ -251,12 +251,7 @@ func (t *TokenSource) performBrowserFlow(ctx context.Context) error {
 		Expiry:       tokenResult.Expiry,
 		TokenType:    tokenResult.TokenType,
 	}
-	var base oauth2.TokenSource
-	if flowCfg.Resource != "" {
-		base = oauth.NewResourceTokenSource(oauth2Cfg, initialToken, flowCfg.Resource)
-	} else {
-		base = oauth2Cfg.TokenSource(ctx, initialToken)
-	}
+	var base oauth2.TokenSource = newNonCachingRefresher(oauth2Cfg, initialToken.RefreshToken, flowCfg.Resource)
 	key := t.refreshTokenKey()
 
 	if t.secretsProvider != nil {
@@ -272,8 +267,9 @@ func (t *TokenSource) performBrowserFlow(ctx context.Context) error {
 		}
 	}
 
-	// Wrap with preemptive refresh so tokens are renewed 30 s before real expiry.
-	t.tokenSource = withPreemptiveRefresh(base)
+	// Pre-seed the outer ReuseTokenSource with the shifted initial token so the
+	// just-obtained access token is served without an immediate network round-trip.
+	t.tokenSource = withPreemptiveRefreshFrom(initialToken, base)
 	return nil
 }
 
@@ -332,7 +328,10 @@ func (t *TokenSource) makeTokenPersister(key string) remote.TokenPersister {
 		t.updateConfigTokenRef(key, expiry)
 		// Invalidate the access-token cache so the next call does not serve a
 		// token minted against the now-rotated refresh token.
-		_ = t.secretsProvider.SetSecret(ctx, t.accessTokenCacheKey(), "")
+		if err := t.secretsProvider.SetSecret(ctx, t.accessTokenCacheKey(), ""); err != nil {
+			slog.Warn("failed to invalidate access-token cache after refresh token rotation",
+				"error", err)
+		}
 		return nil
 	}
 }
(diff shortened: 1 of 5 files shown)


pkg/llm/tokensource.go
```diff
@@ -403,7 +450,77 @@ func (p *preemptiveTokenSource) Token() (*oauth2.Token, error) {
...
+// nonCachingRefresher is an oauth2.TokenSource that always performs a network
+// refresh when Token() is called — it holds no internal token cache. This makes it
+// the correct innermost source for withPreemptiveRefresh: the outer ReuseTokenSource
+// provides caching; the inner source must always refresh when asked so that
+// preemptiveTokenSource triggers exactly one network round-trip per window.
+//
+// It handles both standard OAuth 2.0 refresh (resource == "") and RFC 8707
+// resource-indicator refresh (resource != "") in a single type, eliminating the
+// branching that previously introduced caching inner sources on the resource path.
+type nonCachingRefresher struct {
```
JAORMX: Two things on the doc comment, please:

First, this isn't safe for concurrent use as written. `n.refreshToken` is mutated on every `Token()` call. Right now we're fine because the outer `ReuseTokenSource` serializes everything into us, but anyone who picks this up and uses it outside that chain (e.g. the follow-up where we share it with `resourceTokenSource`) will race. I'd lean toward adding a `sync.Mutex` and making it self-safe rather than just documenting the requirement... that kind of invariant gets forgotten in a refactor.

Second, worth noting that we preserve the previous RT when the IdP doesn't rotate one. That mirrors `pkg/auth/oauth/resource_token_source.go:94-96` and it's the correct behavior, but it's not obvious from a glance.

pkg/llm/tokensource.go
```diff
@@ -403,7 +450,77 @@ func (p *preemptiveTokenSource) Token() (*oauth2.Token, error) {
...
+		resource:     resource,
+		refreshToken: refreshToken,
+		httpClient:   &http.Client{Timeout: 30 * time.Second},
+	}
+}
+
+// Token always performs a token-endpoint refresh. It updates the stored refresh
+// token when the IdP rotates it, so PersistingTokenSource above can detect the
+// change and persist it to the secrets provider.
+func (n *nonCachingRefresher) Token() (*oauth2.Token, error) {
```
JAORMX: This is doing the same `cfg.Exchange(ctx, "", opts...)` dance as `pkg/auth/oauth/resource_token_source.go:88`. The only meaningful difference is that `resourceTokenSource` short-circuits on `r.token.Valid()` (line 45), which is exactly what `withPreemptiveRefresh` needs to *not* happen.

Could we factor a shared non-caching refresher in `pkg/auth/oauth` rather than growing a second copy here? Happy to take it as a follow-up if you'd rather land the fix first, but please file the issue and link it. I don't want this to be the one we forget about and rediscover six months later when we add a third caller.

pkg/llm/tokensource.go
```diff
@@ -273,18 +310,28 @@ func (t *TokenSource) buildOAuth2Config(ctx context.Context) (*oauth2.Config, er
...
 func (t *TokenSource) makeTokenPersister(key string) remote.TokenPersister {
 	return func(refreshToken string, expiry time.Time) error {
 		ctx := context.Background()
 		if err := t.secretsProvider.SetSecret(ctx, key, refreshToken); err != nil {
 			return fmt.Errorf("persisting LLM gateway refresh token: %w", err)
 		}
 		t.updateConfigTokenRef(key, expiry)
+		// Invalidate the access-token cache so the next call does not serve a
+		// token minted against the now-rotated refresh token.
+		if err := t.secretsProvider.SetSecret(ctx, t.accessTokenCacheKey(), ""); err != nil {
```
JAORMX, line 337: Could we add a one-liner explaining why we write empty rather than `DeleteSecret`? The PR description has the answer (provider compatibility, plus `tryAccessTokenCache` reads `""` as a miss) but the next person reading this won't have the PR description in front of them.

Also worth noting in the same comment that the order is intentional: RT first, then AT invalidation. That follows the project's "write durable before in-memory" guidance, even though both writes are durable here. The race window is narrowed but not closed... a reader interleaved between the two `SetSecret` calls still sees the old AT, just for a much smaller window than before.

pkg/llm/tokensource_test.go
```diff
@@ -597,6 +609,206 @@ func TestTokenSource_CacheAccessToken_WriteFailure_DoesNotPanic(t *testing.T) {
...
+//
+// A caching source always returns the same token regardless of how many times
+// Token() is called. When preemptiveTokenSource shifts its expiry back by 30 s,
+// the outer ReuseTokenSource sees an already-expired token on the very next call
+// and re-enters the chain — forever, with no actual IdP refresh.
+//
+// A non-caching inner source (nonCachingRefresher) fixes this: the first call
+// inside the window returns a fresh token with a real-future expiry, which the
+// outer cache can hold until the next window.
+func TestWithPreemptiveRefresh_CachingInnerSource_LoopRegression(t *testing.T) {
```
JAORMX: The name says "loop regression" / "infinite loop within one `Token()` call" but the assertion (`cachingInner.calls == iterations`) just checks one inner call per outer call. The actual old pathology was thrashing across outer calls, not recursion inside one.

Could we either rename to something like `TestWithPreemptiveRefresh_DoesNotThrashAcrossOuterCalls`, or strengthen the assertion to actually pin down the "loop" wording? E.g. "100 outer calls inside the preemptive window cause exactly 1 inner call when the inner returns a long-lived token." Right now the test name promises more than the body delivers.

pkg/llm/tokensource_test.go
```diff
@@ -597,6 +609,206 @@ func TestTokenSource_CacheAccessToken_WriteFailure_DoesNotPanic(t *testing.T) {
...
+// the composition bug where an inner ReuseTokenSource inside the preemptive
+// chain would return the same stale cached token on every call inside the
+// preemptive window, causing the outer ReuseTokenSource to see an expired token
+// on every successive call and re-enter the inner chain indefinitely.
+//
+// Correct behaviour: the first call inside the preemptive window triggers
+// exactly one inner Token() call that returns a fresh long-lived token. The
+// outer ReuseTokenSource caches the shifted expiry (which is now in the future)
+// and serves all subsequent calls from cache — no further inner calls.
+func TestWithPreemptiveRefresh_ExactlyOneRefreshPerWindow(t *testing.T) {
```
JAORMX, line 644: We need a sibling test with `Audience == ""` before this lands.

The PR description hedged that the non-resource path is "still partially broken." That hedge turned out to be stale, but the only reason the hedge survived two reviewers is that there's no test on that path. Same setup as this test, just exercise the cache-restore stack with empty `Audience`. If the hedge ever becomes true again (someone reverts to using `oauth2Cfg.TokenSource`), this test catches it.

JAORMX (commented):
Hey Yolanda! Thanks for chasing this one down... the trace through the broken composition is clear and the fix is correct as far as I can tell. Two inner `Token()` calls per window instead of N is exactly what we want.

So, a few things came out of reviewing this that I'd like us to talk through before stamping it.

## The PR description doesn't match the code

The summary says "the non-resource path still wraps in `reuseTokenSource` internally (oauth2 library limitation). The fix fully applies on the resource path which is the expected LLM gateway configuration."

But the actual diff replaces `oauth2Cfg.TokenSource(ctx, token)` with `newNonCachingRefresher(...)` unconditionally on both call sites. `nonCachingRefresher` handles both paths in one type by branching on `if n.resource != ""`. So Bug 1 is fixed on both paths, not just the resource one.

Could you update the description? It threw a couple of reviewers off and I had to cross-check the diff to be sure.

## The bigger concern... are we duplicating things we already have?

This is what I really wanted to figure out before stamping. I had a hunch we were reinventing pieces of `thv proxy`'s OAuth path, so I went and inventoried it.

`nonCachingRefresher` does almost exactly the same thing as `pkg/auth/oauth/resourceTokenSource` (same `cfg.Exchange(ctx, "", opts...)` trick, same RFC 8707 `resource` handling, same explicit `scope=` forwarding). The one real difference is `resourceTokenSource.Token()` short-circuits on `r.token.Valid()` (line 45), which is exactly the inner cache that breaks `withPreemptiveRefresh`.

I don't want to block this PR on a refactor, but the right shape here is a shared non-caching refresher in `pkg/auth/oauth` that both `pkg/llm` and `resourceTokenSource` use. Otherwise we'll have two implementations of the same RFC 6749 §6 trick and they'll drift. Please file a follow-up and link it here.

While we're at it: `ensureOfflineAccess` (`pkg/llm/tokensource.go:411`) is byte-identical to `pkg/registry/auth/oauth_token_source.go:170`, and `DeriveSecretKey` is the same formula in both packages with a different prefix. None of that's introduced by this PR... it predates it. But registry-OAuth + LLM-OAuth is two callers of the same shape already, and a third is likely down the road. Worth a follow-up to lift the shared bits to `pkg/auth/oauth`.

What's genuinely new and worth keeping in `pkg/llm`: the cross-process `_AT` cache (justified by `thv llm token` being short-lived per invocation, which `thv proxy` isn't), and the preemptive-refresh wrapper. `MonitoredTokenSource` (which `thv proxy` actually uses) only refreshes at expiry, not before, so the preemptive layer is a real new pattern, not a duplicate.

Note on the `thv proxy` comparison since it's relevant to where we want this codebase to land: `thv proxy` has retry, singleflight, and exponential backoff via `MonitoredTokenSource`, none of which `pkg/llm` has today. A transient blip mid-refresh is currently a hard 502 to the LLM client. Out of scope here, but file it.

## In-PR (please address before merge)

- [ ] Update PR description (non-resource path is fixed too)
- [ ] Doc comment on `nonCachingRefresher` covering concurrency assumptions and RT preservation when the IdP returns none
- [ ] Add a test for `Audience == ""`. The current regression test only exercises the resource path... which is precisely the path the stale description claims is "fully fixed".
- [ ] Either rename `TestWithPreemptiveRefresh_CachingInnerSource_LoopRegression` or strengthen the assertion to match the "loop" wording

## Follow-up issues (link them here)

- [ ] Factor a shared non-caching refresher into `pkg/auth/oauth` and migrate `resourceTokenSource` to it
- [ ] Lift `ensureOfflineAccess` and `DeriveSecretKey` to `pkg/auth/oauth` (or wherever makes sense). The `pkg/registry/auth` and `pkg/llm` duplication is silently growing.
- [ ] Consider giving `pkg/llm` retry/singleflight similar to `MonitoredTokenSource`

Otherwise, the fix is correct and the tests are a meaningful step forward. Thanks again!

yrobla:
Filed #5066 to track extracting the shared non-caching refresher into `pkg/auth/oauth` and lifting the duplicated helpers — linked there so it doesn't get lost.

pkg/llm/tokensource.go
yrobla: stacklok/toolhive#5066

JAORMX (approved):

PR #5055, opened 2026-04-25
feat(authserver): add OnUpstreamAuthFailure policy to MCPExternalAuthConfig

BreezeAviation:
## Summary

When an embedded auth server forwards a request to a backend MCP and the backend rejects it with HTTP 401 because the cached upstream access token is no longer usable (upstream provider rotated session, role claims drifted, user lost access, etc.), Toolhive currently passes the 401 through to the client unchanged. The local Toolhive session remains valid, so the next scheduled refresh produces the same stale-claim token and every subsequent request fails identically until the user manually re-authenticates.

We hit this in production with **Microsoft Entra ID upstream + Snowflake as the backend MCP target**. Manual OAuth replay against Entra confirmed the protocol layer is fine — refresh tokens preserve the `session:role-any` scope, claims look identical to the initial auth, and the resulting token connects to Snowflake successfully when used directly. The gap is purely in Toolhive's reaction to a backend rejection.

This PR adds an opt-in policy on `EmbeddedAuthServerConfig`:

```yaml
spec:
  type: embeddedAuthServer
  embeddedAuthServer:
    onUpstreamAuthFailure:
      policy: invalidate-session  # or silent-refresh (default, current behavior)
      triggerOn:                  # optional, defaults to [{status:[401], error:"invalid_token"}]
        - status: [401]
          wwwAuthenticateError: "invalid_token"
```

`silent-refresh` preserves current behavior. `invalidate-session` is the new mode that revokes the user's Toolhive-issued access + refresh tokens and surfaces a proper RFC 6750 401 to the client, prompting MCP-spec-compliant clients to restart the OAuth flow.

Fixes #

## Type of change

- [x] New feature

## Scope of this PR — Design + helper, no wiring yet

This PR establishes the spec and a small, well-tested matching helper. It deliberately does **not** yet wire the policy into the proxy hot path or implement session invalidation — those need maintainer guidance on the right integration points.

### Included

- `EmbeddedAuthServerConfig.OnUpstreamAuthFailure` field on the v1beta1 CRD
- `UpstreamAuthFailurePolicy` (`silent-refresh` | `invalidate-session`) with optional `triggerOn` matchers
- `UpstreamAuthFailureMatcher` (status + `WWW-Authenticate` `error` code)
- Generated DeepCopy implementations
- `pkg/authserver.ResponseMatchesAuthFailure` helper with primitive types (avoids import cycle with the operator API package)
- Robust `WWW-Authenticate: Bearer` parser distinguishing `error=` from `error_description=` and `error_uri=`
- Unit tests covering the default matcher, custom matchers, status-only matchers, header parsing edge cases, and silent-refresh fallback

### NOT included (deliberately) — looking for guidance

- Wiring the helper into ``pkg/transport/proxy/transparent`` or the embedded auth server's session manager
- Token revocation flow (Toolhive access + refresh token eviction + cached upstream token clearing) when the policy fires
- Response rewriting to ensure the client always sees a proper RFC 6750 401 even if the upstream MCP didn't emit one

## Open questions for maintainers

1. **Hook location.** The transparent proxy already has `onUnauthorizedResponse` (`pkg/transport/proxy/transparent/transparent_proxy.go:619`) that fires on backend 401. Should the policy evaluation live there, or in the embedded auth server itself? My initial read is the proxy is the right place since it has both the request context (which user/session) and the response in hand.

2. **Invalidation API surface.** What's the right abstraction over the session/storage layer to invalidate a single user's tokens on demand? I see `pkg/authserver/refresher.go:RefreshAndStore` and the OAuth2 storage interfaces but didn't want to make assumptions about the right composition.

3. **Retry interaction.** Some 401s are transient. Should we add a (small, configurable) backoff/cooldown so a single transient 401 doesn't kick the user out? Or trust that the upstream refresh path already handles transient cases and only invalidate on persistent failure?

4. **Status code for the client response.** Pass the original 401 through (with possibly-rewritten `WWW-Authenticate` to be RFC 6750-compliant), or always emit a synthesized 401? Current sketch passes through; happy to change.

Happy to follow up with the wiring once direction is clear.

## How was this tested?

- `go test ./pkg/authserver/ -run "ExtractBearer|ResponseMatchesAuthFailure" -v` — all pass
- `go test ./cmd/thv-operator/api/v1beta1/` — passes (no regressions to existing CRD validation)
- `go build ./cmd/thv-operator/... ./pkg/authserver/...` — builds clean

Header parser test cases include:
- empty header
- `Bearer error="invalid_token"` (quoted)
- `Bearer error=invalid_token` (unquoted)
- `BEARER error="invalid_token"` (case-insensitive scheme)
- multi-param header with `error_description` and `error_uri` (won't be confused with `error=`)
- `error_uri=...` alone (won't false-match)
- whitespace around `=`

## Checklist

- [x] My code follows the project style guide
- [x] I have added tests covering the new code (helper + matchers)
- [x] I have signed off on my commits via the DCO trailer
- [x] I have updated CRD generated code (`zz_generated.deepcopy.go`)
- [ ] I have updated user-facing docs (deferred until the wiring direction is settled)


BreezeAviation:
Closing — re-opening under personal account, see new PR.

PR #5056, opened 2026-04-25
feat(authserver): add OnUpstreamAuthFailure policy to MCPExternalAuthConfig

dallinstevens:
## Summary

When the embedded auth server forwards a request to a backend MCP and the backend returns HTTP 401 with `WWW-Authenticate: Bearer error="invalid_token"` (per RFC 6750), Toolhive currently passes the response through to the MCP client unchanged. The locally-issued Toolhive JWT remains valid, the cached upstream tokens stay in place, and every subsequent request from that client uses the same upstream token and fails identically — until the user manually re-authenticates.

This PR adds an opt-in policy on `EmbeddedAuthServerConfig`:

```yaml
spec:
  type: embeddedAuthServer
  embeddedAuthServer:
    onUpstreamAuthFailure:
      policy: invalidate-session  # or silent-refresh (default, current behavior)
      triggerOn:                  # optional, defaults to [{status:[401], error:"invalid_token"}]
        - status: [401]
          wwwAuthenticateError: "invalid_token"
```

`silent-refresh` preserves current behavior. `invalidate-session` revokes the user's Toolhive-issued access + refresh tokens (and the cached upstream token set), and surfaces a proper RFC 6750 401 to the client. Spec-compliant MCP clients interpret that as "re-auth required" and restart their OAuth flow.

Fixes #

## Type of change

- [x] New feature

## Why this is needed

We hit a recurring production failure with **Microsoft Entra ID upstream + Snowflake as the backend MCP target**. Pattern:

1. User signs in via OAuth (authorization code flow) — works
2. Queries succeed for ~1h while the original Entra access token is valid
3. Toolhive silently refreshes the upstream Entra token via the cached refresh token (proxy logs confirm this happens)
4. Next request — Snowflake rejects with: *"role 'X' is not listed in the Access Token or was filtered"*
5. The Toolhive-issued JWT is still valid → the client keeps retrying with the same cached upstream token → every request fails
6. User manually re-authenticates → fresh token → works again until the next refresh

We invested significant time pinning down whether this is an Entra-side issue (claim drift on refresh, conditional access, resource binding mismatch) or a Toolhive-side issue. **Manual OAuth replay** against Entra produces tokens that are claim-identical to the failing production scenario at the protocol layer and connect to Snowflake successfully when used directly. We **could not reproduce the failure outside Toolhive's session handling**.

The exact root cause is still under investigation (we suspect resource binding mismatch or CAE-style scope filtering on the refreshed token), but it doesn't matter much for the architectural argument: **Toolhive currently has no way to react when a backend reports the upstream token is unusable**. Whatever causes the rejection — claim drift, expired session, revoked grant, transient Entra issue — the right behavior is the same: invalidate the user's session, force a fresh OAuth flow, and let the upstream provider re-issue with current state.

## What this PR includes

- `EmbeddedAuthServerConfig.OnUpstreamAuthFailure` field on the v1beta1 CRD
- `UpstreamAuthFailurePolicy` (`silent-refresh` | `invalidate-session`) with optional `triggerOn` matchers
- `UpstreamAuthFailureMatcher` (HTTP status + `WWW-Authenticate` `error` code)
- Generated DeepCopy implementations
- `pkg/authserver.ResponseMatchesAuthFailure` helper with primitive types (avoids import cycle with the operator API package)
- Robust `WWW-Authenticate: Bearer` parser distinguishing `error=` from `error_description=` and `error_uri=`
- Unit tests covering the default matcher, custom matchers, status-only matchers, header parsing edge cases, and the silent-refresh fallback

## What this PR deliberately does NOT include — looking for direction

The wiring from "policy matched" to "session invalidated" depends on internal abstractions I didn't want to assume:

- **Where to hook the matching call.** The transparent proxy already has `onUnauthorizedResponse` (`pkg/transport/proxy/transparent/transparent_proxy.go:619`) firing on backend 401. Should the policy evaluation live there, or in the embedded auth server?
- **Invalidation API surface.** What's the right way to revoke a single user's `(access, refresh)` token pair plus their cached upstream token set on demand? `pkg/authserver/refresher.go:RefreshAndStore` exists but I'd rather have maintainer guidance on the right composition.
- **Retry interaction.** Some 401s are transient — should there be a small backoff/cooldown before invalidating, or trust the upstream refresh path to handle transients and only invalidate on persistent failure?
- **Status code on the rewritten client response.** Pass through with possibly-rewritten `WWW-Authenticate`, or always synthesize a 401? Current sketch is pass-through.

Happy to follow up with the wiring as a second commit on this branch once direction is clear.

## Why "tie JWT expiry to upstream expiry" doesn't fit here

A natural-sounding alternative — make the Toolhive JWT expire when the cached upstream token expires — wouldn't fix this class of issue. The upstream Entra access token is being refreshed successfully and on time; it's the *contents* of the refreshed token (or the backend's reaction to those contents) that's at issue. Tying JWT lifetime to upstream lifetime would just mean we re-auth on a timer regardless of whether anything is actually wrong, while still failing in the same way during the windows where the timer hadn't fired yet.

The right trigger is **"backend says this token doesn't work"**, which is exactly what the 401 + `WWW-Authenticate` signal carries.

## How was this tested?

- `go test ./pkg/authserver/ -run "ExtractBearer|ResponseMatchesAuthFailure" -v` — all pass
- `go test ./cmd/thv-operator/api/v1beta1/` — passes (no regressions to existing CRD validation)
- `go build ./cmd/thv-operator/... ./pkg/authserver/...` — builds clean

`WWW-Authenticate: Bearer` parser test cases:
- empty header
- `Bearer error="invalid_token"` (quoted)
- `Bearer error=invalid_token` (unquoted)
- `BEARER error="invalid_token"` (case-insensitive scheme)
- multi-param header with `error_description` and `error_uri` (won't be confused with bare `error=`)
- `error_uri=...` alone (won't false-match)
- whitespace around `=`

## Checklist

- [x] My code follows the project style guide
- [x] I have added tests covering the new code (helper + matchers)
- [x] I have signed off on my commits via the DCO trailer
- [x] I have updated CRD generated code (`zz_generated.deepcopy.go`)
- [ ] I have updated user-facing docs (deferred until the wiring direction is settled)


jhrozek:
@dallinstevens I apologise for the late reply. For some reason this PR slipped in the traffic - I remember seeing a notification but for some reason I thought it was closed.

Having reviewed PR #5096 I wonder if that would address your concern fully or would you prefer us to review this PR as well?

dallinstevens:
@jhrozek yeah this PR is still in draft so wasn't expecting a review on it. It's wasn't worked through all the way yet. I built and deployed 5096 and it resolved my issues, so I don't think there's a need for this at the moment.

PR #5064, opened 2026-04-26
fix(#5063): deterministically order env vars in DeployWorkload

### LINKED ISSUE #5063: MCPServer StatefulSet template flaps due to non-deterministic env var ordering
## Summary

`pkg/container/kubernetes/client.go` builds the StatefulSet pod's env list by iterating a Go `map[string]string` directly. Because Go map iteration order is randomized by the runtime spec, the resulting `[]corev1.EnvVar` ordering changes between reconciles. Kubernetes hashes the pod template (including env list order) to compute the controller revision, so the StatefulSet sees a "changed" template every time iteration order shifts and **rolls the pod**.

This produces:
- repeated StatefulSet pod recreations with no actual spec change
- a 15–30 s window per recreation where the MCP server is unavailable
- secondary effects (proxy `/health` ping races kubelet liveness probe, proxy gets killed, operator reconciles, env-order shifts again → loop)

I observed 7 controllerrevisions over 36 days on a stable `MCPServer`, with 3 in a 4-hour window after the operator pod was rescheduled. Diffing two consecutive revisions showed the only difference was the position of `FASTMCP_PORT` in the env list (`[ULI_MCP_TOKEN, MCP_PORT, MCP_TRANSPORT, MCP_HOST, FASTMCP_PORT]` vs `[ULI_MCP_TOKEN, FASTMCP_PORT, MCP_PORT, MCP_TRANSPORT, MCP_HOST]`).

## Code location

[`pkg/container/kubernetes/client.go` (v0.24.1, lines 360–364)](https://github.com/stacklok/toolhive/blob/v0.24.1/pkg/container/kubernetes/client.go#L360-L364):

```go
// Convert environment variables to Kubernetes format
var envVarList []*corev1apply.EnvVarApplyConfiguration
for k, v := range envVars {
    envVarList = append(envVarList, corev1apply.EnvVar().WithName(k).WithValue(v))
}
```

These lines are byte-identical between v0.12.1 and v0.24.1, so the bug is current.

## Reproduction

1. Deploy any `MCPServer` with transport `streamable-http` (e.g. an image that uses both `MCP_PORT` and `FASTMCP_PORT` envs).
2. Watch `kubectl get controllerrevision -l app=<server-name>` over time, particularly across operator pod restarts.
3. Diff two recent revisions — the env list ordering will differ even though the values match.

## Why it bites in practice

The flapping pod template intersects with a tight liveness/health probe budget:

- The proxy runner's `/health` handler internally calls `Pinger.Ping(ctx)` with a hardcoded 5 s timeout (`pkg/healthcheck/healthcheck.go` `checkMCPStatus`).
- The pinger's HTTP client also has a 5 s default timeout (`pkg/transport/proxy/transparent/pinger.go` `DefaultPingerTimeout`).
- The kubelet liveness probe is configured `timeoutSeconds: 5` (`cmd/thv-operator/controllers/mcpserver_controller.go`).

When the StatefulSet rolls because of env reorder, the bridge pod is briefly unreachable; the proxy's `/health` blocks waiting for the pinger to time out, and the kubelet probe expires at the same 5 s mark. After 3 consecutive expirations the proxy container is killed, which causes another operator reconcile, which re-iterates the map, which produces yet another env order, etc.

The trigger is benign on its own (a no-op spec rewrite) but cascades into proxy crash-looping when probe budgets are tight.

## Suggested fix

Sort the keys before iterating, so the env list is deterministic regardless of map iteration order:

```go
// Convert environment variables to Kubernetes format (deterministic order).
keys := make([]string, 0, len(envVars))
for k := range envVars {
    keys = append(keys, k)
}
sort.Strings(keys)

var envVarList []*corev1apply.EnvVarApplyConfiguration
for _, k := range keys {
    envVarList = append(envVarList, corev1apply.EnvVar().WithName(k).WithValue(envVars[k]))
}
```

This is the same class of fix you applied to `VirtualMCPServer` discovery in #4783 and to vMCP server ordering in #3450 — just on the `MCPServer` code path.

## Environment

- toolhive `v0.12.1` (originally observed); also confirmed bug present in `v0.24.1` source
- Kubernetes (k3s) `v1.34.4`
- Single `MCPServer` with `transport: streamable-http`, no special config

I have a fix ready and will open a PR shortly.

nalditopr:
Fixes #5063.

## Summary

`DeployWorkload` built the StatefulSet pod's env list by iterating a Go `map[string]string` directly. Go map iteration order is randomized per spec, so the resulting `[]EnvVar` order shifted between calls even when contents were identical. Kubernetes folds env list order into the pod template hash, so each reorder produced a "different" template and triggered a no-op StatefulSet rollout.

In a real deployment this surfaced as repeated MCPServer pod recreations after the operator pod was rescheduled, with a nasty secondary effect: during the bridge recreation window, the proxy runner's `/health` pinger blocks waiting for its 5 s HTTP timeout, racing the kubelet liveness probe (also 5 s by default) — the kubelet wins, kills the proxy, the operator reconciles, the env map is iterated again with yet another order, and the cycle repeats.

## Fix

Sort the keys before iterating so the env list is stable across reconciles, regardless of map iteration order:

```go
envKeys := make([]string, 0, len(envVars))
for k := range envVars {
    envKeys = append(envKeys, k)
}
sort.Strings(envKeys)
envVarList := make([]*corev1apply.EnvVarApplyConfiguration, 0, len(envKeys))
for _, k := range envKeys {
    envVarList = append(envVarList, corev1apply.EnvVar().WithName(k).WithValue(envVars[k]))
}
```

Same class of fix as #4783 (vMCP discovery) and #3450 (vMCP server ordering), on the `MCPServer` code path that goes through `pkg/container/kubernetes/client.go`.

## Regression test

Added `TestDeployWorkloadEnvVarOrderingDeterministic` which:
- Calls `DeployWorkload` 11 times with the same env map
- Reads back the StatefulSet's mcp container env from the fake clientset
- Asserts the env list matches a fixed sorted slice on every call

I confirmed the test **fails reliably within a few iterations against the unfixed code** (Go's map iteration is randomized aggressively enough that ordering shifts well within 10 calls) and **passes on every run** with the fix.

## Notes / risk

- The other map iteration in this file (`containerLabels` at line 1310) writes into a target map, so order is irrelevant — left as is.
- No behavior change other than ordering; `corev1.EnvVar` semantics don't depend on position.


ChrisJBurns (commented):
**Agents consulted**: orchestrator only (single-agent review — 13-line fix didn't warrant parallel specialist review)

### Consensus Summary
| 1 | PR title uses conventional-commit prefix | n/a | INFO | Optional |

### Overall
Tight, correct fix. The root-cause analysis is precise: Go map iteration is randomized, env list ordering feeds the K8s pod template hash, and any reorder triggers a StatefulSet rollout. Sorting keys before iteration is the right remedy and matches prior fixes (\#4783, \#3450).

I verified the other map iteration in this file (`containerLabels` at client.go:1301) writes into a destination map and is correctly left alone. The regression test asserts against an explicit sorted slice — documenting the contract rather than just stability — and runs 11 iterations to cover Go's map randomization variance.

The only nit is the title: `.claude/rules/pr-creation.md` says PR titles should not use conventional-commit prefixes (`fix(...)`). Recent history is mixed and this isn't strictly enforced, so non-blocking.

---
*Generated with Claude Code*

ChrisJBurns (dismissed):

ChrisJBurns:
@nalditopr Thanks for this! I approved but the linter is having issues because of the complexity of the method, you able to put the env var sorting into a separate function?

ChrisJBurns (approved):

ChrisJBurns (approved):

samuv (approved):

PR #5065, opened 2026-04-27
Add tool adapter registry and per-tool LLM gateway implementations

### LINKED ISSUE #5030: thv llm: tool adapter registry and per-tool implementations
## Context

Part of #5016 (RFC: stacklok/toolhive-rfcs#70). Depends on #5028 (token lifecycle) and #5029 (proxy).

## Scope

**Registry interface:**
- `Detect() bool`, `Apply(config) error`, `Revert() error` interface for tool adapters
- Auto-discovery registry — adding a new tool requires only implementing the interface and registering; no changes to setup/teardown orchestration

**Token helper (direct) mode tools:**
- **Claude Code** — patches `apiKeyHelper` and base URL in `~/.claude/settings.json`
- **Gemini CLI** — patches token command and base URL in Gemini CLI's config

**Proxy mode tools:**
- **Cursor** — patches base URL and placeholder API key in Cursor's config
- **VS Code (GitHub Copilot)** — patches base URL and placeholder API key in VS Code's settings
- **Apple Xcode** — patches base URL and placeholder API key in Xcode's config

## Acceptance Criteria

- [ ] AT: Claude Code is configured as a token helper tool — setup patches `apiKeyHelper` and base URL in `~/.claude/settings.json`
- [ ] AT: Gemini CLI is configured as a token helper tool — setup patches token command and base URL in Gemini CLI's config
- [ ] AT: Cursor is configured as a proxy-mode tool — setup patches base URL and placeholder API key in Cursor's config
- [ ] AT: VS Code (GitHub Copilot) is configured as a proxy-mode tool — setup patches base URL and placeholder API key in VS Code's settings
- [ ] AT: Apple Xcode is configured as a proxy-mode tool — setup patches base URL and placeholder API key in Xcode's config
- [ ] Adding a new tool requires only implementing `Detect/Apply/Revert` and registering — no changes to orchestration, proxy, or token source

## Dependencies

- #5027 (foundation: config types, package structure)
- #5028 (token source, for token helper tools)
- #5029 (proxy, for proxy-mode tools)

## References

- Parent issue: #5016

yrobla:
## Summary

Implements \`thv llm setup\` / \`thv llm teardown\` by extending the existing \`pkg/client\` framework rather than introducing a parallel adapter registry.

**Why this approach:**
The reviewer (JAORMX) correctly identified that the original \`pkg/llm/tools\` design duplicated path-resolution, file-locking, JSONC handling, and atomic writes that \`pkg/client\` already provides for 28 MCP client integrations. This PR adopts the suggested alternative: extend \`clientAppConfig\` with LLM gateway fields and add \`ConfigureLLMGateway\` / \`RevertLLMGateway\` methods to \`ClientManager\`.

**What changed:**
- Extended \`clientAppConfig\` with \`LLMGatewayMode\`, \`LLMGatewayOnly\`, \`LLMSettingsFile\`, \`LLMSettingsRelPath\`, \`LLMSettingsPlatformPrefix\`, and \`LLMGatewayKeys\` fields.
- Populated LLM gateway settings for Claude Code (direct/token-helper), Gemini CLI (direct), VS Code (proxy), Cursor (proxy), and Xcode (proxy, macOS-only, LLM-gateway-only — not an MCP client).
- Added \`ConfigureLLMGateway\` / \`RevertLLMGateway\` / \`DetectedLLMGatewayClients\` to \`ClientManager\`, picking up file locking (\`fileutils.WithFileLock\`), JSONC preservation (\`hujson\`), and atomic writes (\`fileutils.AtomicWriteFile\`) for free.
- Replaced the "not implemented" stubs in \`thv llm setup\` / \`thv llm teardown\` with real orchestration: setup detects installed tools, patches each settings file, and rolls back on config-persist failure; teardown reverts patches and optionally purges cached OIDC tokens.
- \`GetAllClients()\`, \`IsValidClient()\`, and \`GetClientListFormatted()\` now exclude \`LLMGatewayOnly\` entries so Xcode does not appear in MCP client listings.
- \`runLLMSetup\` / \`runLLMTeardown\` accept an injected \`config.Provider\` for full unit testability. Tests cover: no detected tools, partial apply failure, rollback on config-persist failure, teardown of all/one tool, unknown-tool error.

Fixes #5030

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [ ] E2E tests (\`task test-e2e\`)

## API Compatibility

- [ ] This PR does not break the \`v1beta1\` API, OR the \`api-break-allowed\` label is applied and the migration guidance is described above.

- JSON Patch operations are now constructed via \`json.Marshal([]llmPatchOp{...})\` rather than \`fmt.Sprintf\` to prevent CodeQL "unsafe quoting" findings.
- \`RevertLLMGateway\` guards with \`os.Stat\` before acquiring the lock so a missing/deleted settings directory is a clean no-op rather than a lock-file-creation error.
- \`parentJSONPointer\` uses \`strings.LastIndex\` to return the true direct parent for arbitrarily nested pointers; \`ensurePathExists\` (already segment-walking) handles multi-level creation.
- Follow-up issue for binary-presence detection (\`exec.LookPath\`): #5070.

## Large PR Justification

Integrated refactor: the original \`pkg/llm/tools\` package and its tests are deleted; equivalent functionality is added inside \`pkg/client\` as an extension of the existing client framework. The change cannot be split because the deletion and the extension are a single coherent replacement.


pkg/llm/tools/registry.go
```diff
@@ -0,0 +1,66 @@
...
+var defaultRegistry = &Registry{} //nolint:gochecknoglobals
+
+// Register adds an adapter to the default registry.
+// It is intended to be called from init() functions in tool files.
+func Register(a Adapter) {
+	defaultRegistry.Register(a)
+}
+
+// Default returns the default adapter registry.
+func Default() *Registry { return defaultRegistry }
```
JAORMX: So, the global + `init()`-driven registration pattern doesn't appear in `pkg/client` (which uses an explicit slice literal at `config.go:195`), and the only consumer of this registry is `cmd/thv/app/llm.go`, which knows the full adapter list at compile time. Two costs we're paying for no benefit:

1. The package-level `sync.RWMutex` is dead weight. Nothing concurrently mutates the registry, and `init()` runs single-threaded.
2. Tests can't isolate from the global. `TestDefaultRegistry_ContainsExpectedTools` will quietly start passing for any future adapter, and orchestration tests can't substitute fakes without touching package state.

Easiest move: drop the global. Expose `tools.Builtin() *Registry` that constructs and returns a fresh registry each call, and pass it as a parameter into `runLLMSetup`/`runLLMTeardown`. That removes the `nolint:gochecknoglobals`, removes the mutex, and lets us inject fakes for the orchestration tests we're missing.

That said... if we end up reusing `pkg/client` (see top-level), this whole file goes away.

pkg/llm/tools/adapter.go
```diff
@@ -0,0 +1,57 @@
...
+func applyPatch(pathFn func() (string, error), patchFn func(map[string]any)) (string, error) {
+	path, err := pathFn()
+	if err != nil {
+		return "", err
+	}
+	if err := patchJSONFile(path, patchFn); err != nil {
+		return "", err
+	}
+	return path, nil
+}
```
JAORMX: Quick simplify nit. This helper saves three lines per adapter at the cost of a function-pointer indirection that makes traces harder to read. The name `applyPatch` also shadows the package-internal `patchJSONFile` next door, so the call site reads ambiguously.

If we keep this package, I'd inline the body into each adapter's `Apply()`. The five-line idiom is more obvious than the helper.

(If we land on `JSONConfigUpdater` from `pkg/client` per the top-level comment, this is moot.)

pkg/llm/tools/jsonpatch.go
```diff
@@ -0,0 +1,166 @@
...
+		return err
+	}
+
+	patchFn(m)
+
+	if err := os.MkdirAll(filepath.Dir(path), 0o700); err != nil {
+		return fmt.Errorf("creating directory for %s: %w", path, err)
+	}
+	return writeJSONAtomic(path, m)
+}
```
JAORMX: Two things on this hunk that I want fixed before we touch user files, both already solved in `pkg/client/config_editor.go`:

1. **No file locking.** Two concurrent `thv llm setup` runs, or a setup racing with a running VS Code that's also writing settings.json, can torn-write the file. `pkg/client` wraps every read-modify-write in `fileutils.WithFileLock(path, fn)` for exactly this reason. We need the same here.

2. **JSONC comments are silently dropped.** The doc comment on this function literally says "comments are not preserved". For VS Code and Cursor users that's a real regression... `settings.json` is routinely commented by humans, and we'll rip those comments out on every patch. `pkg/client`'s path uses `hujson.Parse(content)` followed by `v.Patch([]byte(...))` followed by `hujson.Format(v.Pack())`, which keeps comments and trailing commas intact. Same dependency we're already importing, less destructive use of it.

Both fixes come for free if we route LLM gateway edits through `JSONConfigUpdater` (or a sibling that targets `settings.json` instead of `mcp.json`). Even if we keep this package as-is in the short term, those two need to land before this hits users.

pkg/llm/tools/jsonpatch.go
```diff
@@ -0,0 +1,166 @@
...
+	sub, ok := m[head].(map[string]any)
+	if !ok {
+		return
+	}
+	deleteNestedKey(sub, tail)
+	// Remove empty intermediate maps left behind after deletion.
+	if len(sub) == 0 {
+		delete(m, head)
+	}
+}
```
JAORMX: This pruning is a correctness blocker for me. `deleteNestedKey` removes a parent map whenever the deletion left it empty, but it can't tell whether the parent existed before `Apply` touched it.

Concretely: a user has `env: { ANTHROPIC_BASE_URL: "https://...", OTHER_VAR: "..." }` before setup, then between setup and teardown they manually drop `OTHER_VAR`. On teardown we delete `ANTHROPIC_BASE_URL`, see `env` is empty, and helpfully delete the user's `env` block too. They created it... we shouldn't be removing it.

The accompanying test `TestDeleteNestedKey_RemovesEmptyIntermediateMap` asserts the buggy behavior, so it'd need to flip with the fix.

Easiest call: just don't prune empty parents. An empty `{}` is harmless, and we don't have enough information at revert time to know whether the parent was ours.

(If we want to clean up "ours" specifically, we'd need to record in `ToolConfig` exactly which dot-paths `Apply` created and only prune those. Probably not worth it... but worth a thought.)

pkg/llm/tools/jsonpatch.go
```diff
@@ -0,0 +1,166 @@
...
+	if err := tmp.Close(); err != nil {
+		_ = os.Remove(tmpName)
+		return fmt.Errorf("closing temp file for %s: %w", path, err)
+	}
+	if err := os.Rename(tmpName, path); err != nil {
+		_ = os.Remove(tmpName)
+		return fmt.Errorf("renaming temp file to %s: %w", path, err)
+	}
+	return nil
+}
```
JAORMX: This is a from-scratch reimplementation of `pkg/fileutils.AtomicWriteFile`, which we already use in `pkg/client/config_editor.go` for exactly this kind of edit. Three things this version misses that `AtomicWriteFile` gets right:

1. It hardcodes the temp file's mode to whatever `os.CreateTemp` gives us (0o600). If the user's existing settings.json was 0o644 (default on most systems), we silently downgrade. `AtomicWriteFile` accepts a `perm` parameter; we'd stat the existing file first and pass that mode.
2. No `tmp.Sync()` before close. On power loss the rename can land while the content is zero-length on ext4. `AtomicWriteFile` syncs.
3. `os.Rename` returns `EXDEV` on cross-filesystem renames. That's realistic for users with `XDG_CONFIG_HOME` symlinked across volumes (encrypted home, separate `/home` partition). We don't fall back.

Easiest call: replace the body with `fileutils.AtomicWriteFile(path, out, perm)`. Better still, route the whole patch through `JSONConfigUpdater` per the top-level comment.

pkg/llm/tools/platform.go
```diff
@@ -0,0 +1,34 @@
...
+const osDarwin = "darwin"
+
+// xdgConfigHome returns the XDG_CONFIG_HOME directory, falling back to
+// ~/.config when the env var is not set.
+func xdgConfigHome(home string) string {
+	if dir := os.Getenv("XDG_CONFIG_HOME"); dir != "" {
+		return dir
+	}
+	return filepath.Join(home, ".config")
+}
```
JAORMX: Two things here:

1. The XDG spec says `XDG_CONFIG_HOME` must be an absolute path; non-absolute values are required to be ignored. `os.Getenv != ""` doesn't enforce that. If `XDG_CONFIG_HOME=foo`, this silently returns a relative path and we end up writing to the cwd. Wrap with `filepath.IsAbs(dir)`.

2. More broadly: `pkg/client/config.go:143-165` already encodes per-OS prefix logic via `PlatformPrefix map[Platform][]string` for 28 client integrations. Reusing that gets us VSCode Insiders, VSCodium, Code-OSS, Code-Server, etc. for free, and gives us one place to fix XDG handling for everyone. That's the path I'd push for here... `pkg/llm/tools/platform.go` should not exist.

pkg/llm/tools/claudecode.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Detect reports whether ~/.claude/ exists (proxy for Claude Code being installed).
+func (a *claudeCodeAdapter) Detect() bool {
+	home, err := a.homeDirFn()
+	if err != nil {
+		return false
+	}
+	_, err = os.Stat(filepath.Join(home, ".claude"))
+	return err == nil
+}
```
JAORMX: `os.Stat(~/.claude)` is a false-positive prone signal. The directory hangs around from prior installs, MCP-only setups (`thv client register claude-code`), or unrelated tooling that decided to use the same prefix. If a user has the dir but no claude binary on `$PATH`, we'll happily go ahead and create `~/.claude/settings.json` with our patches.

Same goes for `~/.gemini` in the geminicli adapter.

Fix: combine the directory check with `exec.LookPath("claude")` (and `"gemini"` respectively) and require both before we say it's installed.

Note that `pkg/client.IsClientInstalled(ClaudeCode)` already does this style of detection for the existing client list... so if we move LLM gateway support into `pkg/client` per the top-level comment, this consideration moves with it.

pkg/llm/tools/cursor.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		return filepath.Join(appData, "Cursor", "User", "settings.json"), nil
+	default:
+		home, err := a.homeDirFn()
+		if err != nil {
+			return "", err
+		}
+		return filepath.Join(xdgConfigHome(home), "Cursor", "User", "settings.json"), nil
+	}
+}
```
JAORMX: This whole switch is duplicated almost verbatim in `vscode.go:58-79`. Only the literal string ("Cursor" vs "Code") differs. The new abstraction was supposed to remove this kind of duplication, and we ended up reproducing it.

More importantly: Cursor is already in `pkg/client` as `Cursor ClientApp = "cursor"` (`config.go:313`), with `RelPath: []string{".cursor"}`. The PR targets a different settings file (`Library/Application Support/Cursor/User/settings.json` instead of `~/.cursor/mcp.json`), but the path-resolution machinery (`PlatformPrefix` + `RelPath` + `SettingsFile`) handles exactly this kind of split for VS Code already, and the constants the PR uses (`"cursor"`) line up with what's already there.

Two paths forward, both kill this duplication:

1. Add an optional `LLMGatewaySettingsFile` (and friends) to `clientAppConfig`, plus a second SettingsFile variant on the existing Cursor entry. This file disappears.
2. If we keep `pkg/llm/tools/`, factor out `vscodeStyleSettingsPath(appDir, homeDirFn, appDataFn)` into a helper so cursor and vscode share the body.

Same comment applies to `vscode.go`.

pkg/llm/tools/vscode.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		return filepath.Join(appData, "Code", "User", "settings.json"), nil
+	default:
+		home, err := a.homeDirFn()
+		if err != nil {
+			return "", err
+		}
+		return filepath.Join(xdgConfigHome(home), "Code", "User", "settings.json"), nil
+	}
+}
```
JAORMX: Same point as `cursor.go`. VS Code is already in `pkg/client` (`VSCode ClientApp = "vscode"`, `config.go:284`) with the exact `PlatformPrefix` this code reinvents:

```go
PlatformPrefix: map[Platform][]string{
    PlatformLinux:   {".config"},
    PlatformDarwin:  {"Library", "Application Support"},
    PlatformWindows: {"AppData", "Roaming"},
},
RelPath: []string{"Code", "User"},
```

That produces the exact path this function builds. The same table also covers `VSCodeInsider`, `VSCodeServer`, AmpVSCode, AmpVSCodeInsider... none of which we pick up here. Reusing the table keeps Insiders working for free.

If we keep this package short term, please at least factor out the OS switch into a shared `vscodeStyleSettingsPath(appDir, ...)` helper so cursor and vscode share it.

pkg/llm/tools/geminicli.go
```diff
@@ -0,0 +1,57 @@
...
+	return revertJSONFile(configPath, "auth.tokenCommand", "baseUrl")
+}
+
+func (a *geminiCLIAdapter) settingsPath() (string, error) {
+	home, err := a.homeDirFn()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(home, ".gemini", "settings.json"), nil
+}
```
JAORMX: Worth flagging: this resolves to `~/.gemini/settings.json`, which is the **identical** path `pkg/client.GeminiCli` (`config.go:708-727`) builds via `RelPath: []string{".gemini"}` and `SettingsFile: "settings.json"`.

So we're patching the same file with a parallel code path, and along the way we lose the file lock and JSONC preservation that `JSONConfigUpdater` would give us for free. If a future change to Gemini's config layout lands in `pkg/client`, this adapter silently stops matching.

If LLM gateway capability becomes a flag on `clientAppConfig`, this whole adapter collapses to one extra map of keys-to-templates on the existing `GeminiCli` entry. Cleanest case for reuse in the whole PR.

cmd/thv/app/llm.go
```diff
@@ -203,26 +204,138 @@ func buildLLMTokenSource(cfg *llm.Config, interactive bool) (*llm.TokenSource, e
...
+			}
+			if revertErr := a.Revert(tc.ConfigPath); revertErr != nil {
+				_, _ = fmt.Fprintf(cmd.ErrOrStderr(),
+					"Warning: rollback of %s failed: %v\n", tc.Tool, revertErr)
+			}
+		}
+		return fmt.Errorf("persisting tool configuration: %w", err)
+	}
+	return nil
+}
```
JAORMX: `runLLMSetup` is the most user-visible code in the PR (it's the function that touches user settings files), and the rollback branch on lines 275-294 is the riskiest path in the function... it only runs when `UpdateConfig` fails *after* we've already mutated user files. There are no unit tests for any of this, which is what's driving the codecov 62.6% complaint.

The blocker for testing today is `llmtools.Default()` being a process global. If we expose `tools.Builtin()` and pass `*Registry` (or `[]Adapter`) into `runLLMSetup`/`runLLMTeardown` as a parameter, we can drive the function end-to-end with fake adapters and exercise:

- `Detected()` returns empty (the silent no-op path).
- One adapter's `Apply` fails (partial-failure, warning emitted, others continue).
- `UpdateConfig` fails after success (the rollback path... the one we most need to know works).
- `runLLMTeardown` with a single-tool target vs. all-tools.

`mergeToolConfigs` and `isTarget` should also pick up direct unit tests... they're cheap and they fail loud if the merge logic ever drifts.

cmd/thv/app/llm.go
```diff
@@ -203,26 +204,138 @@ func buildLLMTokenSource(cfg *llm.Config, interactive bool) (*llm.TokenSource, e
...
-		RunE: func(_ *cobra.Command, _ []string) error {
-			return fmt.Errorf("not implemented: coming in a future release")
+		Short: "Remove LLM gateway configuration from all (or one) configured tools",
+		Long: `Revert the configuration changes made by "thv llm setup" for all configured
+tools, or for a single tool when tool-name is provided.
+
+Use --purge-tokens to also remove cached OIDC tokens from the secrets provider.`,
+		Args: cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			purge, _ := cmd.Flags().GetBool("purge-tokens")
```
JAORMX: Style nit. The rest of this file binds flags via `BoolVar`/`StringVar` directly to a struct or local variable (see `newConfigSetCommand` for the pattern). This is the only place we read via `cmd.Flags().GetBool` and discard the error with `_`.

Two costs:

1. `.claude/rules/go-style.md` says ignored errors must be commented.
2. The pattern is inconsistent with the rest of the file, which is a small but real readability tax.

```go
var purgeTokens bool
cmd := &cobra.Command{
    ...,
    RunE: func(cmd *cobra.Command, args []string) error {
        return runLLMTeardown(cmd, args, purgeTokens)
    },
}
cmd.Flags().BoolVar(&purgeTokens, "purge-tokens", false, "...")
```

JAORMX (changes requested):
Hey Yolanda, thanks for putting this together. There's real thought in here, the `Detect/Apply/Revert` shape is clean, and the atomic-rename instinct is right. I want to push back on the overall direction before we go further though, because I think we're reinventing a fair chunk of what `pkg/client` already does.

## The headline: this overlaps a lot with `pkg/client`

`pkg/client` is the data-driven framework that already handles "find this developer tool's config dir per OS, edit its JSON file safely". It supports 28 client integrations today via the `clientAppConfig` table at `pkg/client/config.go:195`, and the way new capabilities (like skills) get added to it is by extending the existing struct, not by forking a new package next to it.

Of the 5 tools this PR adds adapters for, 4 are already entries in that table, under the **exact same identifier strings** the PR uses:

| PR adapter | `pkg/client` constant | value |
|---|---|---|
| `claude-code` | `ClaudeCode` | `"claude-code"` |
| `gemini-cli` | `GeminiCli` | `"gemini-cli"` |
| `cursor` | `Cursor` | `"cursor"` |
| `vscode` | `VSCode` | `"vscode"` (also `VSCodeInsider`, `VSCodeServer`) |
| `xcode` | (genuinely new) | |

For VS Code, the `PlatformPrefix` map already produces the **exact path** this PR's `vscode.go` reinvents (`Library/Application Support/Code/User/settings.json` on darwin, `AppData/Roaming/Code/User/settings.json` on windows, `.config/Code/User/settings.json` on linux). For Gemini CLI, the path is already byte-for-byte identical (`~/.gemini/settings.json`). The PR is patching the same files that `pkg/client` already knows how to find.

## What `pkg/client` does that the new package gets wrong

While we're paralleling all that path-resolution logic, we drop three correctness properties `pkg/client/config_editor.go` already gives us, and the JSONC one is a real user-data concern:

- **No file locking.** `pkg/client` wraps every read-modify-write in `fileutils.WithFileLock(path, fn)`. Two concurrent `thv llm setup` runs (or a setup racing with a running editor that's also writing settings.json) torn-write the file. `pkg/llm/tools/jsonpatch.go` has no locking.
- **JSONC comments silently dropped.** The doc comment on `patchJSONFile` literally says so. `settings.json` is routinely commented by humans for VS Code and Cursor users, and we'll rip those out on every patch. `pkg/client` uses `hujson.Parse` then `v.Patch(...)` then `hujson.Format(v.Pack())`, which keeps comments and trailing commas intact. Same dependency we already import, less destructive use of it.
- **`writeJSONAtomic` is a from-scratch reimplementation of `pkg/fileutils.AtomicWriteFile`** that misses `Sync()`, hardcodes mode 0o600 (downgrades a 0o644 file silently), and won't survive an `EXDEV` cross-filesystem rename.

There's a project rule that's directly on point here. From `.claude/rules/go-style.md`:

> Search the toolhive repo first... check if an existing method, utility, or package already provides the functionality or something close enough to extend.

That should have been the first move, and `pkg/client` is the obvious match.

## What I'd like to see instead

1. Extend `clientAppConfig` with optional `LLMGatewayMode` and a small map of "JSON pointer path -> value template" plus a second `SettingsFile` (Claude Code uses `.claude/settings.json` rather than `.claude.json`, that's fine, just a second settings file on the existing entry). The pattern is the same one used to add `SupportsSkills` + `SkillsGlobalPath` to those entries. Extend, don't fork.
2. Add `ClientManager.ConfigureLLMGateway` / `RevertLLMGateway` methods that route through `JSONConfigUpdater` so they pick up file locking, JSONC preservation, and atomic writes for free.
3. Drop `pkg/llm/tools/{registry,platform,jsonpatch}.go` entirely. The five "adapters" become entries in `supportedClientIntegrations`.
4. Xcode (genuinely not in `pkg/client` today) gets added as a new entry with `PlatformPrefix` set only for darwin.

That keeps the orchestration in `cmd/thv/app/llm.go` more or less intact... it just iterates over `pkg/client` clients filtered by `LLMGatewayMode != ""` instead of over a separate registry.

## Other things worth fixing in addition to the reuse refactor

A few that came up while reading and that I left inline:

- **`revertJSONFile` deletes empty parent maps.** That can clobber a user's `env` block on the second teardown if other env keys were removed in between. Correctness blocker as written. The accompanying test `TestDeleteNestedKey_RemovesEmptyIntermediateMap` asserts the buggy behavior.
- **The init()-driven global registry is a new pattern in the codebase** (`pkg/client` itself uses an explicit slice literal). The only consumer is `cmd/thv/app/llm.go`, which knows the full set at compile time. Drop the global, expose `tools.Builtin()`, and pass it as a parameter so we can unit test the orchestration with fakes.
- **Codecov flags 62.6% patch coverage.** The orchestration in `cmd/thv/app/llm.go`, including the rollback path inside `runLLMSetup`, has zero unit tests. That's the riskiest code in the PR (it runs only after we've already mutated user files). Injecting the registry fixes the testability blocker.
- **`Detect()` based on directory existence is unreliable.** `~/.claude` and `~/.gemini` hang around from prior installs or unrelated tooling. Combine with `exec.LookPath("claude")` / `exec.LookPath("gemini")`. (`pkg/client.IsClientInstalled` already does this kind of detection.)
- **`cursor.go` and `vscode.go` `settingsPath()` switches are line-for-line duplicates.** The new abstraction was supposed to remove this kind of duplication.
- **"Cannot be split" doesn't hold up.** There are at least three natural seams (shared helpers + `Adapter` interface + tests, per-tool adapters, `cmd/thv/app/llm.go` orchestration). The project's `pr-creation.md` rule asks for ~400 LOC, and this is ~1500 across 15 files.

## Bottom line

The work is good direction, I just want LLM gateway support to live as another capability on the existing client framework rather than as a parallel one. The current shape forks the supported-tools list into two places that will drift, ships behind weaker file-handling guarantees than `pkg/client` already provides, and silently destroys user comments in `settings.json`.

Happy to pair on the refactor if useful, or to talk through the design before you go too far. If you'd rather keep the current shape, I want to understand the reasoning before we go further... I might be missing something.

Thanks again for the careful work on the structure. Let me know what makes sense.

pkg/llm/tools/jsonpatch.go
yrobla: removed

pkg/llm/tools/jsonpatch.go
yrobla: removed

pkg/llm/tools/jsonpatch.go
yrobla: removed

pkg/llm/tools/vscode.go
yrobla: removed

pkg/llm/tools/geminicli.go
yrobla: removed

yrobla:
## Follow-up tracking

The major architecture concerns raised in this review have been addressed in the latest push (the `pkg/llm/tools` package has been removed; LLM gateway support now lives as an extension of `clientAppConfig` in `pkg/client`, picking up `fileutils.WithFileLock`, `hujson` JSONC preservation, and `fileutils.AtomicWriteFile` for free).

One actionable item from the review is being deferred to a follow-up issue rather than expanded in this PR:

- **Tool detection via `exec.LookPath`** (JAORMX's comment on `claudecode.go` / `Detect()` false positives): tracked in #5070 — *"llm: strengthen tool detection with binary presence check"*.

All other inline comments are either resolved or were about the now-deleted `pkg/llm/tools` package.

cmd/thv/app/llm.go
```diff
@@ -203,34 +204,208 @@ func buildLLMTokenSource(cfg *llm.Config, interactive bool) (*llm.TokenSource, e
...
+		GatewayURL:         llmCfg.GatewayURL,
+		ProxyBaseURL:       proxyBaseURL,
+		TokenHelperCommand: fmt.Sprintf(`"%s" llm token`, self),
+	}
+
+	detected := cm.DetectedLLMGatewayClients()
+	if len(detected) == 0 {
+		_, _ = fmt.Fprintln(cmd.OutOrStdout(), "No supported AI tools detected.")
+		return nil
+	}
```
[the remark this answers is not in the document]
yrobla: Stale comment — the PR has been fully refactored. `pkg/llm/tools` was deleted and its functionality is now implemented by extending `pkg/client` (see `pkg/client/llm_gateway.go`). The PR description has been updated to reflect this.

pkg/client/llm_gateway.go
```diff
@@ -0,0 +1,228 @@
...
+			value := llmValueForSpec(spec.ValueField, cfg)
+			valueJSON, err := json.Marshal(value)
+			if err != nil {
+				return fmt.Errorf("marshaling value for %s: %w", spec.JSONPointer, err)
+			}
+			pathJSON, err := json.Marshal(spec.JSONPointer)
+			if err != nil {
+				return fmt.Errorf("marshaling path %s: %w", spec.JSONPointer, err)
+			}
+			patch := fmt.Sprintf(`[{"op":"add","path":%s,"value":%s}]`, pathJSON, valueJSON)
```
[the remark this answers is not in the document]
yrobla: fixed

yrobla:
## Follow-up: xcode in swagger ClientApp enum

The Copilot review correctly flagged that `xcode` appears in the generated `ClientApp` swagger enum despite being LLM-gateway-only.

**Why it can't be fixed here:** `swag init` derives the enum from all Go constants of a type. Since `Xcode ClientApp = "xcode"` is a valid constant, it will always appear in the generated docs — manually removing it is overwritten on every `task docs` run.

**Current protection:** Code-level guards prevent misuse — `CreateClientConfig`, `GetClientStatus`, and `IsClientInstalled` all return `ErrUnsupportedClientType`/skip for `LLMGatewayOnly` entries.

**Tracked for proper fix:** #5073 — *"llm: use separate type for LLM-gateway-only clients to remove them from MCP API swagger enum"*.

JAORMX (changes requested):
Hey Yolanda, this is a much tighter shape — `pkg/llm/tools` is gone, the file-handling guarantees are the same ones `pkg/client` already gives us, and the orchestration tests cover the rollback path I was worried about. Big improvement, thanks for the rework.

Round 2 is mostly nits. One real doc/code mismatch I'd want resolved before merge, plus a quoting hardening and a rebase.

## What's resolved from round 1

For my own confidence I walked through every prior item against the current head:

| Item | Status | Where |
|---|---|---|
| `pkg/llm/tools` parallel registry | ✅ deleted; 4/5 clients extend existing `clientAppConfig` entries | `pkg/client/config.go` |
| File locking | ✅ `fileutils.WithFileLock` | `pkg/client/llm_gateway.go:55, 123` |
| JSONC preservation | ✅ `hujson.Parse → Patch → Format` | `:63, 90, 95` |
| Atomic writes | ✅ `fileutils.AtomicWriteFile` | `:99, 158` |
| Empty-parent pruning bug | ✅ gone (RFC 6902 `remove` only, no recursive cleanup) | — |
| `init()` global registry | ✅ gone | — |
| Injected `config.Provider` + rollback path tested | ✅ via `errOnUpdateProvider` stub | `cmd/thv/app/llm_test.go:202, 309` |
| Cursor/VSCode `settingsPath` duplication | ✅ all paths route through `buildConfigFilePath` + `LLMSettingsPlatformPrefix` | — |
| `BoolVar` style | ✅ | `cmd/thv/app/llm.go:347` |
| CodeQL "unsafe quoting" | ✅ all patches via `json.Marshal([]llmPatchOp{…})` | — |
| `RevertLLMGateway` lock-on-missing-dir | ✅ `os.Stat` before `WithFileLock` | `:119-121` |
| `parentJSONPointer` true direct parent | ✅ `strings.LastIndex` | `:223-229` |
| `LLMGatewayOnly` filtering | ✅ all six entry points (`GetAllClients`, `IsValidClient`, `GetClientListFormatted`, `IsClientInstalled`, `GetClientStatus`, `CreateClientConfig`) | — |
| Binary-presence detection | ⏭ deferred to #5070, fine by me | — |

## What I'd want before merge

**1. The PR description and `ConfigureLLMGateway` disagree with the code on multi-level parent creation.**

Description says *"`ensurePathExists` (already segment-walking) handles multi-level creation"*. The actual `ensureLLMParent` at `pkg/client/llm_gateway.go:236-252` only ensures the **single direct parent**. Trace `/a/b/c` where `/a` is missing:

1. `parentJSONPointer("/a/b/c")` → `/a/b`
2. `jsonPointerExists` → false
3. `add /a/b → {}` patch fails because RFC 6902 requires `/a` to exist

Today no `LLMGatewayKeySpec` goes deeper than 2 (`/env/ANTHROPIC_BASE_URL`, `/auth/tokenCommand`), so this is dormant... but the next config that needs `/a/b/c` will silently fail at apply time. Two ways out:

- Walk the segments and ensure each ancestor exists (~8 LOC).
- My preference: route through the existing `ensurePathExists` in `pkg/client/config_editor.go`. It already segment-walks AND parses with hujson. That deletes `jsonPointerExists` + `ensureLLMParent` + the standardize dance, and the multi-level guarantee comes for free. Closes the loop on round-1's "reuse `pkg/client` primitives" theme.

Either fix the code or trim the description to say "single direct parent only" and document the constraint on `LLMGatewayKeySpec.JSONPointer`.

**2. Token-helper command quoting.** `cmd/thv/app/llm.go:251` builds `fmt.Sprintf("\"%s\" llm token", self)` from `os.Executable()`. The string is JSON-marshalled correctly — that's fine — but it's then re-parsed as a **shell command line** by Claude Code's `apiKeyHelper` and Gemini's `auth.tokenCommand`. JSON escaping doesn't help against shell metacharacters. If the executable path contains a literal `"`, `\`, or `;` (legal on macOS/Linux, occasional on Windows installers), the helper command becomes garbage and the user's tool can't authenticate.

Cheapest defense: reject `os.Executable()` paths containing `"\\;\n\r` with a clear error from `runLLMSetup`. Or POSIX-quote on Unix. Either way — this gets written into long-lived user config, so getting it wrong is very visible.

**3. Rebase.** PR is `behind` main. #5052 (`68d29fbc`, withPreemptiveRefresh fix) touches `pkg/llm/tokensource.go` and added `pkg/auth/oauth/non_caching_refresher.go` after this branch was cut. Not a revert risk — the merge would pull them in — but please rebase so CI runs against current main and any semantic conflicts in the LLM-token area surface now rather than after merge.

## Smaller stuff (would-be-nice, not blocking)

- **Symlink TOCTOU in `RevertLLMGateway`** — Stat→Lock→ReadFile follows symlinks; a same-UID process could swap the file between checks. `os.Lstat` reject of symlinks hardens this. Low likelihood given the threat model.
- **Brittle string match at `RevertLLMGateway:147-148`** — `"path not found"` is dead code (hujson only emits `"value not found"`). Cleaner: pre-check with `jsonPointerExists` and skip the patch when absent.
- **`pkg/config/config.go`'s new `Marshal()`** exists solely for one test helper. Inline `yaml.Marshal(cfg)` in `tempProvider` and don't grow the public surface.
- **`runLLMSetup` / `runLLMTeardown`** is real business logic with 390 LOC of unit tests in `cmd/thv/app/`. Per `.claude/rules/cli-commands.md`, that's a smell — orchestration belongs in `pkg/`. Future-cleanup direction, not for this PR.
- **Tests** — `TestConfigureLLMGateway_JSONCPreservesExistingParent` only substring-asserts `EXISTING_KEY`; the headline JSONC promise (comments + trailing commas survive) deserves a literal assertion. The rollback test in `runLLMSetup` configures one tool; the contract is "all patched tools roll back" — a 2-tool variant would lock that. Both cheap to add.

Re-iterating: the core is in great shape. Once 1 and 2 are addressed (and rebased), this is good to go.

Thanks again for taking the round-1 feedback seriously and doing the actual refactor — the result is markedly cleaner.

JAORMX (approved):

Issue #5016, opened 2026-04-22
thv llm subcommand for LLM gateway authentication


jerm-dro:
## User Story

**As a** developer using AI coding tools in an enterprise environment,
**I want** ToolHive to automatically configure my tools to authenticate with our OIDC-protected LLM gateway,
**So that** I don't have to manually manage tokens or deal with silent failures when they expire.

## Context

RFC-0070 (stacklok/toolhive-rfcs#70) proposes a `thv llm` command group with two authentication modes:

- **Proxy mode** — localhost reverse proxy for static-key-only tools (Cursor, VS Code, Xcode) that injects fresh OIDC tokens
- **Token helper mode** — `thv llm token` prints a fresh JWT for OIDC-capable tools (Claude Code, Gemini CLI)

A single `thv llm setup` command detects installed tools, configures them, starts the proxy, and handles OIDC login.

### Setup & teardown lifecycle
- [ ] AT: `thv llm setup --gateway-url <url> --issuer <issuer> --client-id <id>` detects installed AI tools, configures each for the gateway, starts the background proxy if any proxy-mode tool is detected, and triggers the OIDC browser login
- [ ] AT: `thv llm teardown` removes gateway configuration from all tool config files, stops the background proxy, and optionally purges cached OIDC tokens with `--purge-tokens`
- [ ] AT: `thv llm teardown <tool-name>` targets a single tool without affecting others

### Proxy mode (static-key tools)
- [ ] AT: The localhost proxy accepts requests on `http://localhost:<port>/v1`, strips the incoming `Authorization` header, injects a fresh OIDC bearer token, and forwards to the upstream gateway — preserving the original request path, query string, headers, and body
- [ ] AT: The proxy passes through upstream responses faithfully — including SSE streaming (chat completions), non-streaming JSON (embeddings, completions), and error responses (4xx/5xx) — without buffering or modifying response bodies
- [ ] Unit: Proxy rejects startup if the listen address is not a loopback interface. No AT because loopback validation is a startup guard with no observable e2e behavior beyond a failed start.

### Token helper mode (OIDC-capable tools)
- [ ] AT: `thv llm token` prints a fresh JWT to stdout (all other output on stderr) suitable for use as a tool's `apiKeyHelper` or `auth.command`
- [ ] Unit: `thv llm token` exits with an error in non-interactive mode when no cached or refreshable token exists — it never launches a browser flow. No AT because this is a flag-gated code path tested without external dependencies.

### Token lifecycle
[...]

PR #5068, opened 2026-04-27
thv llm: complete setup/teardown orchestration

### LINKED ISSUE #5031: thv llm: setup and teardown orchestration
## Context

Part of #5016 (RFC: stacklok/toolhive-rfcs#70). Final subtask — depends on all previous subtasks.

## Scope

**`thv llm setup`:**
- Detects installed AI tools via the tool adapter registry
- Configures each detected tool (Apply)
- Starts the background proxy if any proxy-mode tool is detected
- Triggers the OIDC browser login flow

**`thv llm teardown`:**
- Removes gateway configuration from all tool config files (Revert)
- Stops the background proxy
- `--purge-tokens` flag to also delete cached OIDC tokens from the secrets provider
- `thv llm teardown <tool-name>` — targets a single tool without affecting others

## Acceptance Criteria

- [ ] AT: `thv llm setup --gateway-url <url> --issuer <issuer> --client-id <id>` detects installed tools, configures each, starts the background proxy if any proxy-mode tool is present, and triggers OIDC browser login
- [ ] AT: `thv llm teardown` removes gateway config from all tool config files, stops the proxy, and optionally purges cached tokens with `--purge-tokens`
- [ ] AT: `thv llm teardown <tool-name>` targets a single tool without affecting others

## Dependencies

- #5027 (foundation)
- #5028 (token lifecycle)
- #5029 (proxy)
- #5030 (tool adapters)

## References

- Parent issue: #5016

yrobla:
## Summary

Implements `thv llm setup` and `thv llm teardown`, the final piece of the LLM gateway authentication epic (#5016).

**Why:** The tool adapter scaffolding was already in place (detection, config patching, revert), but setup and teardown were stubs. Users had no end-to-end command to configure their AI tools to authenticate through the LLM gateway.

**What changed:**

- `thv llm setup` accepts inline `--gateway-url` / `--issuer` / `--client-id` (and the other config flags), merging them into the persisted config before running. This lets you combine `config set` and `setup` into one invocation.

- OIDC login runs **before** tool config files are patched. If login fails, no tool config files or persisted state are modified. For proxy-mode tools (Cursor, VS Code, Xcode), setup prints a hint to run `thv llm proxy start` — the proxy is a normal foreground command, not a background daemon.

- `thv llm teardown` reverts all (or a single named) tool's config patches. `--purge-tokens` deletes cached OIDC tokens from the secrets provider and clears the cached token reference/expiry from config.

- `purgeCachedTokens` and `hasProxyMode` extracted as helpers. `loginFunc` is injected into `runLLMSetup` so unit tests can skip the keyring.

- E2E tests in `cli_llm_setup_test.go` cover: inline-flag config + tool patching, full teardown revert, single-tool targeted teardown, no-config error, and `--purge-tokens` cleanup.

Fixes #5031

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## API Compatibility

Yes — `thv llm setup` configures all detected AI tools and triggers OIDC login in a single command. `thv llm teardown` reverts those changes. Proxy-mode users run `thv llm proxy start` separately as a foreground process.

The proxy is intentionally not started automatically by `setup` — no daemon, no PID tracking, no platform-specific detach logic. Users run `thv llm proxy start` themselves.

## Follow-up

stacklok/toolhive#5091 — move setup/teardown orchestration to `pkg/llm/` so HTTP API and operator callers can reuse it without importing cobra.


yrobla:
**Note on proxy lifecycle (scope deviation from issue #5031)**

The original acceptance criteria requires setup to start the background proxy and teardown to stop it. This PR intentionally does not implement that: instead, setup prints a hint to run `thv llm proxy start` manually, and teardown does not touch the proxy.

**Reasoning:** Background daemon management (PID persistence, cross-platform signal handling, process detachment) adds significant complexity and fragility. The proxy already works as a foreground command; users can run it in a terminal, under `systemd`, `launchd`, or any process supervisor of their choice.

If the team prefers we implement the daemon approach anyway, I can add it back — but wanted to flag this deviation explicitly rather than leave reviewers to discover it in the diff.

cmd/thv/app/llm.go
```diff
@@ -421,17 +494,34 @@ func runLLMTeardown(
...
+			_, _ = fmt.Fprintf(errOut, "Warning: could not clear cached token metadata: %v\n", updateErr)
 		}
 	}
 
 	return nil
 }
 
+// purgeCachedTokens deletes all cached OIDC tokens from the system secrets
+// provider. Errors are logged as warnings rather than returned.
+func purgeCachedTokens(ctx context.Context, errOut io.Writer) {
```
aponcedeleonch, line 500: Nice extraction. One thing though: `newConfigResetCommand` (around lines 144-151 in this file) still has the same secrets-provider-get + `DeleteCachedTokens` + warn-on-error logic inlined. I think it should call this helper too, otherwise the two paths will drift the next time someone tweaks the warning text or adds a step.

cmd/thv/app/llm.go
```diff
@@ -221,21 +223,73 @@ Token-helper tools (Claude Code, Gemini CLI) are configured to call
...
+				}
+			}
 			cm, err := client.NewClientManager()
 			if err != nil {
 				return fmt.Errorf("initializing client manager: %w", err)
 			}
-			return runLLMSetup(cmd.OutOrStdout(), cmd.ErrOrStderr(), cm, config.NewDefaultProvider())
+			return runLLMSetup(
+				cmd.Context(), cmd.OutOrStdout(), cmd.ErrOrStderr(),
+				cm, config.NewDefaultProvider(), oidcLogin, inlineOpts,
```
aponcedeleonch: I think this loop can go. `SetFields` already filters zero values field-by-field, so passing `opts` straight through gives the same result. The comment also says it protects against `--proxy-port 0` being mistaken for unset, but `SetFields` ignores `ProxyPort == 0` either way, so the loop doesn't actually do that.

```suggestion
			cm, err := client.NewClientManager()
			if err != nil {
				return fmt.Errorf("initializing client manager: %w", err)
			}
			return runLLMSetup(
				cmd.Context(), cmd.OutOrStdout(), cmd.ErrOrStderr(),
				cm, config.NewDefaultProvider(), oidcLogin, opts,
```

cmd/thv/app/llm.go
```diff
@@ -289,6 +351,11 @@ func runLLMSetup(out, errOut io.Writer, cm *client.ClientManager, provider confi
 	}
 
 	if err := provider.UpdateConfig(func(c *config.Config) error {
+		// Persist inline flag overrides and ConfiguredTools atomically so that
+		// if this write fails, no partial state is left on disk.
+		if err := c.LLM.SetFields(inlineOpts); err != nil {
```
aponcedeleonch, line 344: `SetFields` already runs once on the local copy at line 289, so this second call validates the same merged opts again. You could pass the merged `llmCfg` straight into the closure and skip this:

```go
if err := provider.UpdateConfig(func(c *config.Config) error {
    c.LLM = llmCfg
    c.LLM.ConfiguredTools = mergeToolConfigs(c.LLM.ConfiguredTools, configured)
    return nil
}); err != nil { ... }
```

That said, replacing `c.LLM` wholesale loses any concurrent on-disk changes that `UpdateConfig` would otherwise pick up. If that matters here, fine to leave as-is.

cmd/thv/app/llm.go
```diff
@@ -421,17 +494,34 @@ func runLLMTeardown(
...
 	if purgeTokens {
-		secretsProvider, err := secrets.GetSystemSecretsProvider()
-		if err != nil {
-			_, _ = fmt.Fprintf(errOut, "Warning: could not get secrets provider: %v\n", err)
-		} else if err := llm.DeleteCachedTokens(context.Background(), secretsProvider); err != nil {
-			_, _ = fmt.Fprintf(errOut, "Warning: could not remove cached LLM tokens: %v\n", err)
+		purgeCachedTokens(ctx, errOut)
+		// Clear cached token metadata from config so stale references do not
+		// remain after the secrets have been deleted.
+		if updateErr := provider.UpdateConfig(func(c *config.Config) error {
```
aponcedeleonch: This second `UpdateConfig` could fold into the one earlier in the function. With `purgeTokens` true, you'd clear `CachedRefreshTokenRef` and `CachedTokenExpiry` in the first write and call `purgeCachedTokens` after. One disk write, and no window where the secrets are gone but the cached refs still point at them.

aponcedeleonch (commented):
Looks solid overall. Nice job on the e2e suite, the fake-browser script and the cancellable auth wait are well thought out. A few inline comments below.

One higher-level thing I wanted to flag: `runLLMSetup` and `runLLMTeardown` are now around 100 lines each of orchestration, with login injection, rollback, and atomic persistence. Per `.claude/rules/cli-commands.md` the CLI layer should be a thin wrapper with business logic in `pkg/`. Not a blocker for this PR, but worth thinking about before more callers (HTTP API, operator) need the same flow.

A couple of testing gaps I noticed too: I think it's worth adding a unit test that injects a `loginFunc` returning an error and asserts no tool config files were touched and no config was persisted. The PR description calls this out as a key behavior change but I don't see it covered. Same for the explicit-zero case on `--proxy-port 0` once you decide what to do with the flag-loop comment below.

aponcedeleonch (approved):
All four fixes addressed cleanly. The intentional decision to keep the two `SetFields` calls with the updated comment about preserving concurrent writes is the right call. The login-failure test is exactly the gap I had in mind. Nice extra touch reordering detection before login so an empty detection skips the browser flow entirely.

LGTM.

Issue #4979, opened 2026-04-21
Persistent DCRCredentialStore backends (memory + Redis)


jhrozek:
## Description

Phase 3 of the DCR story (#4976). Ship persistent `DCRCredentialStore` backends (in-process memory + Redis) that drop in behind the interface defined by #4978, and wire the persistent store into `pkg/authserver/runner/embeddedauthserver.go` in place of the Phase 2 in-memory stub. This is the last missing piece for the Datadog-style upstream demo to survive authserver restarts and scale across replicas: restarting (or scaling out to) an authserver backed by Redis reuses already-registered DCR credentials instead of re-registering at the upstream.

This umbrella tracks three sub-issues:

1. #5183 — types, `DCRCredentialStore` interface, and in-memory implementation
2. #5184 — Redis backend
3. #5185 — wire the persistent store into `EmbeddedAuthServer`

## Context

#4978 (Phase 2) defines the narrow `DCRCredentialStore` interface — `Get(ctx, DCRKey) (*DCRResolution, bool, error)` and `Put(ctx, DCRKey, *DCRResolution) error` — together with the `DCRKey{Issuer, RedirectURI, ScopesHash}` canonical cache key, and ships an in-process in-memory stub keyed on `(issuer, redirect_uri, scopes_hash)`. Full RFC 7592 fields (`RegistrationAccessToken`, `RegistrationClientURI`) are already captured by Phase 2 so Phase 3 has everything it needs to round-trip a credential through durable storage.

The persistent backends must:

- Match the existing `pkg/authserver/storage/` shape — one in-memory implementation, one Redis Sentinel implementation — and follow the prior-art Redis key scheme `thv:auth:{ns:name}:<type>:<id>` (see `pkg/authserver/storage/redis_keys.go:63-83`) so DCR keys share the `{ns:name}` Redis Cluster hash tag with the rest of a server's state.
- Honor `client_secret_expires_at` (RFC 7591 §3.2.1, unix seconds; `0` means "never") as the Redis TTL when the authorization server provides it. When absent or `0`, no TTL is applied and entries are long-lived.
- Obey `.claude/rules/go-style.md` §"Write to Durable Storage Before Updating In-Memory State" on every write path — durable first, in-memory after.
[...]

PR #5069, opened 2026-04-27
Expose DCR config in operator CRD for OAuth2 upstreams

### LINKED ISSUE #5040: Authserver DCR: expose DCR config in operator CRD (Phase 2, CRD surface)
## Context

Part of #4978 (Phase 2 of the DCR story, #4976). This is the final sub-issue: it exposes the DCR configuration surface in the operator CRD so that Kubernetes users can configure Dynamic Client Registration on `VirtualMCPServer` upstreams. CRD changes are intentionally deferred to this last PR so the operator API evolves only after the authserver plumbing is fully merged and stable.

**Depends on**: #5039 (Sub-issue C — authserver wiring, observability, and integration tests must land first).
**Blocks**: #4979 (Phase 3 persistent backends — the `DCRCredentialStore` interface defined in earlier sub-issues is the drop-in point).

---

### Operator API types (`cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go`)

- Add `DCRUpstreamConfig` CRD type mirroring `pkg/authserver/config.go`'s `DCRUpstreamConfig`. Use a secret ref for the initial access token (following the `ClientSecretRef` pattern already used in `OAuth2UpstreamConfig`) rather than a file/env-var path:
  ```go
  // DCRUpstreamConfig configures Dynamic Client Registration for an OAuth2 upstream.
  // +kubebuilder:object:generate=true
  type DCRUpstreamConfig struct {
      // DiscoveryURL is the base URL for RFC 8414 AS metadata discovery.
      // Mutually exclusive with RegistrationEndpoint.
      DiscoveryURL string `json:"discoveryUrl,omitempty"`
      // RegistrationEndpoint is an explicit override for the DCR endpoint.
      // Mutually exclusive with DiscoveryURL.
      RegistrationEndpoint string `json:"registrationEndpoint,omitempty"`
      // InitialAccessTokenRef is an optional reference to a Secret carrying
      // an RFC 7591 §3 initial access token. Mirrors the ClientSecretRef pattern.
      InitialAccessTokenRef *corev1.SecretKeySelector `json:"initialAccessTokenRef,omitempty"`
      // SoftwareID is a stable identifier for this ToolHive deployment.
      SoftwareID string `json:"softwareId,omitempty"`
      // SoftwareStatement is an optional RFC 7591 §2.3 signed JWT.
      SoftwareStatement string `json:"softwareStatement,omitempty"`
  }
  ```
- Add `DCRConfig *DCRUpstreamConfig` field to `OAuth2UpstreamConfig`.
- Add kubebuilder CEL validation enforcing `DiscoveryURL` and `RegistrationEndpoint` are mutually exclusive and at least one is set when `DCRConfig` is present.

### Generated deepcopy (`cmd/thv-operator/api/v1beta1/zz_generated.deepcopy.go`)

Regenerated by `task gen` — no manual edits.

### Conversion (`cmd/thv-operator/pkg/controllerutil/authserver.go`)

In the `UpstreamProviderTypeOAuth2` conversion block (~line 683), after the existing field mappings, map `DCRConfig`:
```go
if provider.OAuth2Config.DCRConfig != nil {
    dcrCfg := provider.OAuth2Config.DCRConfig
    rc := &authserver.DCRUpstreamConfig{
        DiscoveryURL:         dcrCfg.DiscoveryURL,
        RegistrationEndpoint: dcrCfg.RegistrationEndpoint,
        SoftwareID:           dcrCfg.SoftwareID,
        SoftwareStatement:    dcrCfg.SoftwareStatement,
    }
    if dcrCfg.InitialAccessTokenRef != nil {
        // Resolve secret ref to env-var name, following the same pattern as ClientSecretRef.
        rc.InitialAccessTokenEnvVar = envVarName // computed from ref
    }
    config.OAuth2Config.DCRConfig = rc
}
```

### CRD YAML manifests

Regenerated by `task crdref-gen` + `task operator-manifests` — no manual edits.

### Validation tests (`cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go`)

Extend with table-driven cases:
- `DCRConfig` set with non-empty `ClientID` → validation error.
- `DCRConfig` with both `DiscoveryURL` and `RegistrationEndpoint` set → validation error.
- `DCRConfig` with neither set → validation error.
- `DCRConfig` with only `DiscoveryURL` → valid.
- `DCRConfig` with only `RegistrationEndpoint` → valid.

---

## Acceptance Criteria

- [ ] `DCRUpstreamConfig` CRD type defined in `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go` with fields `DiscoveryURL`, `RegistrationEndpoint`, `InitialAccessTokenRef` (secret ref, mirrors `ClientSecretRef` pattern), `SoftwareID`, `SoftwareStatement`.
- [ ] `OAuth2UpstreamConfig` gains `DCRConfig *DCRUpstreamConfig` field.
- [ ] Kubebuilder CEL validation enforces: `DiscoveryURL` and `RegistrationEndpoint` are mutually exclusive; at least one must be set when `DCRConfig` is non-nil.
- [ ] Conversion in `cmd/thv-operator/pkg/controllerutil/authserver.go` maps all `DCRConfig` fields from CRD type → `authserver.DCRUpstreamConfig`, including resolving `InitialAccessTokenRef` to an env-var name following the same pattern as `ClientSecretRef`.
- [ ] `task gen` regenerates `zz_generated.deepcopy.go` cleanly with no unexpected diff beyond the new types.
- [ ] `task crdref-gen` and `task operator-manifests` regenerate CRD YAML with the new `dcrConfig` field present in the OpenAPI schema.
- [ ] Validation tests: table-driven covering DCRConfig+ClientID conflict, both endpoints set, neither endpoint set, and valid single-endpoint cases.

---

## Patterns & References

- `ClientSecretRef` handling in `cmd/thv-operator/pkg/controllerutil/authserver.go:~293` and `~683` — follow the same secret-ref → env-var-name resolution pattern for `InitialAccessTokenRef`.
- Existing kubebuilder CEL validation examples in `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go` — follow the same `+kubebuilder:validation:XValidation` annotation style.
- `cmd/thv-operator/api/v1beta1/zz_generated.deepcopy.go` is fully generated — run `task gen`, do not hand-edit.
- CRD YAML manifests are fully generated — run `task crdref-gen` + `task operator-manifests`, do not hand-edit.
- `task license-fix` adds SPDX headers to any new files if needed.

tgrunnagle:
# Expose DCR config in operator CRD for OAuth2 upstreams

Closes #5040

## Summary

Phase 2 of the DCR story (#4978) finishes by surfacing Dynamic Client Registration (RFC 7591) in the operator API so Kubernetes users can configure DCR on `VirtualMCPServer` OAuth2 upstreams. The CRD changes were intentionally held back until the authserver plumbing landed, and this PR threads the new `dcrConfig` field from the CRD through conversion to the existing `authserver.DCRUpstreamConfig` runtime type. Validation lives in CEL on the CRD plus a defense-in-depth check at reconcile time, and the initial access token follows the same Secret-ref → env-var pattern already used by `ClientSecretRef`.

### Operator API (`cmd/thv-operator/api/v1beta1/`)

- Add `DCRUpstreamConfig` type with `discoveryUrl`, `registrationEndpoint`, `initialAccessTokenRef` (a `corev1.SecretKeySelector`), `softwareId`, and `softwareStatement` fields.
- Add optional `dcrConfig *DCRUpstreamConfig` to `OAuth2UpstreamConfig` and make `clientId` optional.
- CEL validation enforces: exactly one of `clientId` or `dcrConfig` on `OAuth2UpstreamConfig`; exactly one of `discoveryUrl` or `registrationEndpoint` inside `dcrConfig`; and a fail-closed discriminator on `UpstreamProviderConfig` so unknown future provider types are rejected at admission instead of silently accepted.
- `MaxLength=255` on `softwareId` and `MaxLength=4096` on `softwareStatement` to bound CR size against etcd object limits.

### Conversion (`cmd/thv-operator/pkg/controllerutil/authserver.go`)

- Map `DCRConfig` from the CRD onto `authserver.DCRUpstreamConfig` in the OAuth2 upstream branch, including resolving `InitialAccessTokenRef` to a `TOOLHIVE_UPSTREAM_DCR_INITIAL_ACCESS_TOKEN_<PROVIDER>` env var sourced from the referenced Secret — same pattern as `ClientSecretRef`.
- Call the exported `ValidateOAuth2DCRConfig` from `BuildAuthServerRunConfig` so malformed `ClientID`/`DCRConfig` pairs that bypass admission still fail at reconcile time rather than at authserver startup. OIDC and OAuth2 branches were split into helpers to keep gocyclo within limits.

### Authserver runtime (`pkg/authserver/`, `pkg/oauthproto/`, `pkg/auth/`)

- DCR credential store and resolver: `pkg/authserver/runner/dcr.go`, `dcr_store.go`. In-memory store keyed by `(Issuer, RedirectURI, ScopesHash)`, no TTL (RFC 7591 registrations are long-lived). The key shape is designed to compose a Redis segment in Phase 3 without redefining the canonical form.
- `EmbeddedAuthServer` now owns a `DCRCredentialStore` and resolves DCR for any OAuth2 upstream with `DCRConfig` before building the upstream config. The resolved `ClientSecret` is overlaid on the built `upstream.OAuth2Config` via a new `applyResolutionToOAuth2Config` helper.
- Each `UpstreamRunConfig` element is shallow-copied and its OAuth2 sub-config deep-copied before DCR resolution to honor the "Copy Before Mutating Caller Input" rule.
- The `/oauth/register` handler emits a structured Info log on success with `issuer`, `client_id`, `software_id`, `token_endpoint_auth_method`, and `scopes`. `software_id` is threaded through `ValidateDCRRequest` and capped at 256 printable-ASCII characters.
- `MonitoredTokenSource` gains required `upstream` and `clientID` constructor parameters so DCR remediation warnings carry correlation fields, and the runner prefers the DCR-cached `client_id` over the statically configured one.
- `pkg/oauth` is renamed to `pkg/oauthproto` and the DCR/discovery primitives from `pkg/auth/oauth/dynamic_registration.go` and `pkg/auth/discovery/` are consolidated there. Callers updated.

### Generated artifacts

- `zz_generated.deepcopy.go`, `deploy/charts/operator-crds/**/*.yaml`, `docs/operator/crd-api.md`, and `docs/server/{docs.go,swagger.json,swagger.yaml}` regenerated via `task gen` / `task crdref-gen` / `task operator-manifests`. No hand edits.

## Implementation Details

- **Fail-closed CEL discriminator on `UpstreamProviderConfig`.** Restructured the previous binary check into a chain ending in explicit `false`, with a contributor-facing doc comment reminding future authors to extend both the CEL rule and `validateUpstreamProvider` when adding a new `UpstreamProviderType`.
- **No Unicode smart quotes in kubebuilder markers.** `gofmt`'s doc-comment formatter was reintroducing U+201D into `!= ''` markers on every `lint-fix`. Switched to CEL's `size(...) > 0` idiom so regeneration is idempotent.
- **Single error log at the boundary.** The DCR resolver now wraps step failures in `DCRStepError` and the boundary caller (`buildUpstreamConfigs`) emits a single `slog.Error` via `LogDCRStepError`, instead of every error branch logging on its own.
- **URL-trailing punctuation preserved.** `sanitizeErrorForLog`'s URL-stripping regex was greedily consuming sentence-ending punctuation. Trailing terminal punctuation is now split off the match before parsing and re-appended after the query is stripped, with regression coverage for commas, periods, parens, and quoted URLs.

## Testing

- `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go`: table-driven CEL validation covering `dcrConfig`+`clientId` conflict, both endpoints set, neither endpoint set, valid single-endpoint cases, OIDC-typed provider rejecting an OAuth2 `dcrConfig`, and the non-DCR `clientId`-only path.
- `cmd/thv-operator/pkg/controllerutil/authserver_test.go`: new `TestBuildAuthServerRunConfig_InvalidDCR` suite covering all four invalid `DCRConfig`/`ClientID` pairings at the conversion layer; conversion tests for `discoveryUrl`/`registrationEndpoint` paths and `initialAccessTokenRef` env-var wiring; assertion that the upstream name appears exactly once in scoped error strings.
- `pkg/authserver/runner/dcr_test.go`, `dcr_store_test.go`, `embeddedauthserver_test.go`: full DCR boot path against a mock authorization server, cache-hit short-circuit asserting zero additional HTTP requests, caller's original `RunConfig.Upstreams` slice element unchanged across calls, `TestNewEmbeddedAuthServer_DCRBoot` driving the full constructor and asserting `dcrStore` is populated after boot.
- `pkg/oauthproto/dcr_test.go`, `discovery_test.go`: relocated and extended coverage for DCR primitives and AS metadata discovery in the consolidated package.
- `pkg/authserver/server/handlers/handler_chain_test.go`: `TestHandler_issuer` so a real wiring bug fails loudly instead of logging `issuer=""`.
- `task lint`, `task lint-fix`, `task license-check`, and the affected unit tests all pass; CRD/deepcopy regeneration is idempotent.

## API Compatibility

- This PR does not break the `v1beta1` API. `dcrConfig` is a new optional field, `clientId` becomes optional but the existing CEL rule requires exactly one of `clientId` or `dcrConfig`, so previously-valid CRs that only set `clientId` continue to validate.

## Large PR Justification

The diff is ~2k LoC, split roughly in half between hand-written code and autogenerated artifacts. A clean split would leave the feature in an unreviewable intermediate state.

- **Generated artifacts (~1090 LoC, 5 files).** Adding `DCRUpstreamConfig` and reshaping `OAuth2UpstreamConfig` regenerates four CRD YAMLs (the `MCPExternalAuthConfig` and `VirtualMCPServer` CRDs each appear in `deploy/charts/operator-crds/files/crds/` and the `templates/` mirror), `zz_generated.deepcopy.go`, and `docs/operator/crd-api.md`. These are produced by `task operator-manifests` / `task crdref-gen` / `task gen` and are not hand-edited.
- **Tightly coupled CRD + conversion + tests slice (~1080 LoC, 4 files).** `mcpexternalauthconfig_types.go` adds the API surface (`DCRUpstreamConfig`, optional `clientId`, CEL XOR rules, the unified discriminator check, length caps, the `ValidateOAuth2DCRConfig` reconcile-time backstop). `pkg/controllerutil/authserver.go` does the conversion onto the existing `authserver.DCRUpstreamConfig` runtime type, including the `TOOLHIVE_UPSTREAM_DCR_INITIAL_ACCESS_TOKEN_<PROVIDER>` env-var binding. The matching test files exercise both layers. Splitting would either ship a CRD that admits DCR but is dropped at conversion time, or conversion code with no API to invoke it — the user-visible feature is the union of the two.
- **Review-feedback hardening already factored into separate commits.** The last three commits (`Tighten and align upstream OAuth2/DCR validation`, `Refactor ValidateOAuth2DCRConfig and OAuth2 conversion helper`, `Strengthen DCR env-var assertions and consolidate redundant tests`) split the response to the [16-finding consensus review](https://github.com/stacklok/toolhive/pull/5069#pullrequestreview-4212486535) along clear themes — validation tightening, validator API refactor, test consolidation — so each is reviewable on its own diff on top of the original feature commit.

Yes. Cluster admins can now configure RFC 7591 Dynamic Client Registration on `VirtualMCPServer` OAuth2 upstreams via the new `dcrConfig` block on `OAuth2UpstreamConfig`, including a Secret-backed initial access token, a `softwareId`/`softwareStatement` pair, and either an RFC 8414 `discoveryUrl` or an explicit `registrationEndpoint`. ToolHive registers a client with the upstream on first use and caches the resulting credentials in memory.

- This unblocks #4979 (Phase 3 persistent backends). The `DCRCredentialStore` interface is the drop-in point — no further interface churn is expected when the Redis-backed implementation lands.
- The `pkg/oauth` → `pkg/oauthproto` rename is in this PR because the DCR resolver imports the consolidated package; it touches a number of import statements but is otherwise a pure rename with no behavior change.
- `MonitoredTokenSource`'s `upstream`/`clientID` constructor parameters are now required (replacing the previous `SetUpstreamContext` setter). Call sites were updated; the change forces compile-time visibility of the correlation fields rather than relying on a post-construction setter.


cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -410,6 +435,81 @@ type OAuth2UpstreamConfig struct {
...
+	// +optional
+	// +kubebuilder:validation:MaxLength=255
+	SoftwareID string `json:"softwareId,omitempty"`
+
+	// SoftwareStatement is the RFC 7591 "software_statement" JWT asserting
+	// metadata about the client software, signed by a party the authorization
+	// server trusts. Bounded to 4096 characters to prevent unbounded
+	// user input from inflating the CR beyond etcd object limits.
+	// +optional
+	// +kubebuilder:validation:MaxLength=4096
```
tgrunnagle: **F1 / MEDIUM / consensus 10/10 — softwareStatement MaxLength=4096 may reject realistic signed JWTs** _(raised by CRD-specialist + security-specialist)_

Real-world RFC 7591 software_statements with embedded `x5c` certificate chains, JWKS keys, or OIDC-Federation trust-framework metadata routinely exceed 4 KB. A two-cert RSA-2048 chain alone is ~3 KB base64url-encoded; with payload + signature you cross 4 KB easily. The cap will reject otherwise-valid statements at admission with a generic "too long" error.

Either raise the limit (e.g. to 16384 — still well under etcd object limits) or document the cap as a deliberate constraint with operator-facing guidance for what to do when their IdP exceeds it.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -410,6 +435,81 @@ type OAuth2UpstreamConfig struct {
...
+	// DiscoveryURL is the RFC 8414 / OIDC Discovery document URL. The resolver
+	// issues a single GET against this URL (no well-known-path fallback) and
+	// reads registration_endpoint, authorization_endpoint, token_endpoint,
+	// token_endpoint_auth_methods_supported, and scopes_supported from the
+	// response.
+	// Mutually exclusive with RegistrationEndpoint.
+	// MaxLength bounds CEL cost estimation on the parent struct's
+	// XValidation rule and matches the conventional URL length cap.
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https?://.*$`
```
tgrunnagle: **F2 / MEDIUM / consensus 9/10 — DCR URL pattern allows `http://` and lacks whitespace constraints** _(raised by security-specialist)_

The `^https?://.*$` pattern on `discoveryUrl` (here) and `registrationEndpoint` (line 488) accepts plaintext `http://` to arbitrary hosts. Per RFC 7591 §3 and RFC 8414 §3, the registration endpoint MUST be HTTPS — it carries the initial access token (a bearer credential) and the issued client_secret in the response. The runtime `DCRUpstreamConfig.Validate()` enforces HTTPS-or-loopback, but admission silently accepts the misconfiguration today, shifting the failure from `kubectl apply` to authserver `CrashLoopBackOff`. The `OIDCUpstreamConfig.IssuerURL` precedent in this same file (line 323) already uses `^https://.*$`.

Separately, the existing `UserInfo` patterns (lines 181, 191) use the more restrictive `^https?://[^\s?#]+[^/\s?#]$` which rejects whitespace, fragments, and trailing slashes — the DCR fields use the looser variant.

```suggestion
	// +kubebuilder:validation:Pattern=`^https://[^\s?#]+[^/\s?#]$`
```

Apply the same change at line 488 for `RegistrationEndpoint`. If `http://` on loopback must be supported for parity with the runtime layer's local-dev allowance, encode that explicitly rather than allowing arbitrary `http://`.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -956,10 +1056,62 @@ func (*MCPExternalAuthConfig) validateUpstreamProvider(index int, provider *Upst
...
+// stored objects (e.g., objects stored before CEL rules were added or
+// validated through code paths that bypass admission).
+//
+// The prefix is embedded in error messages to identify the offending upstream
+// (e.g., "upstreamProviders[i]"). Pass an empty string when the caller already
+// wraps the error with the upstream identifier, to avoid duplicating it.
+// Exported so the controllerutil conversion layer can reuse the same
+// invariants when building runtime configs, rejecting malformed objects at
+// reconcile time rather than waiting until the authserver process parses them.
+func ValidateOAuth2DCRConfig(prefix string, cfg *OAuth2UpstreamConfig) error {
```
tgrunnagle: **F3 / MEDIUM / consensus 8/10 — `ClientSecretRef` + `DCRConfig` together is silently accepted** _(raised by security-specialist)_

`ValidateOAuth2DCRConfig` (and the matching CEL on `OAuth2UpstreamConfig`) only enforces XOR between `ClientID` and `DCRConfig`. A user can still set `ClientSecretRef` alongside `DCRConfig`, but in DCR mode the `client_secret` comes from the registration response (RFC 7591 §3.2.1 makes the response authoritative). The static `ClientSecretRef` is then either dead config or a competing source of truth — a future-debugging trap either way.

Add the rejection here and a matching CEL counterpart on `OAuth2UpstreamConfig` (e.g. `!has(self.dcrConfig) || !has(self.clientSecretRef)`), plus a new test case in both `TestMCPExternalAuthConfig_validateUpstreamProvider` and `TestBuildAuthServerRunConfig_InvalidDCR`.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -956,10 +1056,62 @@ func (*MCPExternalAuthConfig) validateUpstreamProvider(index int, provider *Upst
...
+		return fmt.Errorf("%s.dcrConfig: exactly one of discoveryUrl or registrationEndpoint must be set", scoped)
+	}
+	return nil
+}
+
+// scopedFieldPath joins a non-empty prefix to a field name with a dot. When
+// the prefix is empty, it returns just the field name so callers that already
+// supply their own scope (e.g., a wrapping `fmt.Errorf("upstream %q: %w", ...)`)
+// don't end up with a leading dot.
+func scopedFieldPath(prefix, field string) string {
```
tgrunnagle: **F4 / MEDIUM / consensus 8/10 — `prefix` parameter is an unenforced convention; `scopedFieldPath` is single-use** _(raised by Go-architecture-specialist)_

The `prefix` parameter on `ValidateOAuth2DCRConfig` (above, line 1083) has two implicit modes: pass `upstreamProviders[i]` from admission, pass `""` from the conversion layer (so `BuildAuthServerRunConfig`'s outer `upstream %q: %w` wrap doesn't duplicate the name). The convention lives in two doc comments — there is no compile-time enforcement. A third caller passing the upstream name would silently produce a doubly-prefixed message, and the `strings.Count(err.Error(), "acme-idp") == 1` test guards only the path that exists today.

Two cleaner shapes:
- Split into `validateOAuth2DCRConfigScoped(prefix, cfg) error` (admission) and `ValidateOAuth2DCRConfig(cfg) error` (always unscoped, callers wrap).
- Drop the prefix parameter entirely; always emit `oauth2Config: ...` and let every caller wrap with their own scope.

Either change makes `scopedFieldPath` (this line, single caller) disappear, which is the right outcome — it doesn't pull its weight as a separate helper.

tgrunnagle (commented):

**Agents consulted**: pr-reviewer × 4 specialists (CRD/CEL design, OAuth/security, Go architecture, test quality). Codex cross-review skipped (codex CLI not installed). Posted as a `COMMENT` event because this PR is a draft authored by the reviewer's GitHub account.

### Consensus Summary

| 1 | softwareStatement 4096-char cap may reject realistic JWTs (x5c cert chains, OIDC Federation) | 10/10 | MEDIUM | Discuss/Fix |
| 2 | DCR URL patterns allow `http://` and lack whitespace constraints | 9/10 | MEDIUM | Fix |
| 3 | `ClientSecretRef` + `DCRConfig` together silently accepted | 8/10 | MEDIUM | Fix |
| 4 | `ValidateOAuth2DCRConfig` prefix parameter is unenforced convention | 8/10 | MEDIUM | Fix |
| 5 | DCR env-var test does not assert `SecretKeyRef.Key` | 8/10 | MEDIUM | Fix |
| 6 | `MaxLength=2048` doc-comment rationale is incorrect | 8/10 | LOW | Fix |
| 7 | "OIDC+DCRConfig" test misleadingly named — fires OIDC discriminator, not DCR check | 8/10 | LOW | Fix |
| 8 | `buildOAuth2UpstreamRunConfig` redundant params + signature asymmetry with OIDC sibling | 8/10 | LOW | Fix |
| 9 | CEL discriminator vs Go validator messages diverge | 7/10 | LOW | Polish |
| 10 | `dcrConfig: {}` produces less-specific outer error | 7/10 | LOW | Polish |
| 11 | softwareStatement plaintext in CR spec (no Secret-ref alternative) | 7/10 | LOW | Discuss |
| 12 | Dead defensive nil-check on `provider.OAuth2Config` | 7/10 | LOW | Polish |
| 13 | `extractUpstreamSecretRefs` named-return convention inconsistent | 7/10 | LOW | Polish |
| 14 | Multi-upstream + long-name env-var coverage missing | 7/10 | LOW | Polish |
| 15 | Boundary length tests missing | 7/10 | LOW | Polish |
| 16 | `TestBuildAuthServerRunConfig_InvalidDCR` has 3 redundant cases | 7/10 | LOW | Polish |

### Overall

This is a well-thought-through operator-side slice of the larger DCR feature. The CRD additions follow established kubebuilder patterns, the conversion-layer refactor is clean, and the defense-in-depth between CEL and Go validators is the right architecture. The PR description is unusually thorough — it walks reviewers through structural choices (XOR rules, fail-closed discriminator, the `gofmt` smart-quote trap, the URL-trailing-punctuation regression) and explains why each one exists. That level of write-up paid off: there are no architectural blockers in the diff.

The four MEDIUMs are worth resolving before this exits draft. Three of them (F1, F2, F3) are incomplete-validation cases — the XOR contracts in this PR are correct as far as they go, but they have edges that admission silently accepts: a too-tight softwareStatement cap that some real-world IdPs will cross, plaintext `http://` on credential-bearing endpoints, and the missing `clientSecretRef` ↔ `dcrConfig` mutual-exclusivity check. F4 is the architectural one — the prefix-string parameter on the validator is a documented convention with no compile-time enforcement, and a future caller passing the upstream name would silently produce duplicated identifiers.

The cross-cutting theme on the LOWs is **CEL ↔ Go validator alignment** — F6, F9, F2 all touch the contract that the two layers should say the same things at the same scope. Worth doing once now while the contract is fresh, rather than ratcheting it after the next XValidation rule lands. Tests are solid in shape but have two specific blind spots: missing `SecretKeyRef.Key` assertions (F5 — the test currently can't tell if the DCR initial-access-token Key gets crossed with the client-secret Key) and missing multi-upstream / boundary scenarios (F14, F15).

### Documentation

Two doc fixes are propagated artifacts — the `MaxLength=2048` rationale (F6) lives in the source field comments and is auto-replicated into both CRD YAMLs and `docs/operator/crd-api.md`. After correcting the source comment, re-run `task operator-manifests` and `task crdref-gen` so the propagated text picks up the correction.

---

cmd/thv-operator/pkg/controllerutil/authserver_test.go
```diff
@@ -497,6 +498,86 @@ func TestGenerateAuthServerEnvVars(t *testing.T) {
...
+									Name: "acme-dcr-token",
+									Key:  "token",
+								},
+							},
+						},
+					},
+				},
+			},
+			wantEnvNames:    []string{UpstreamDCRInitialAccessTokenEnvVarPrefix + "_ACME_IDP"},
+			wantSecretNames: []string{"acme-dcr-token"},
```
tgrunnagle, line 531: **F5 / MEDIUM / consensus 8/10 — DCR env-var test does not assert `SecretKeyRef.Key`** _(raised by test-specialist)_

`wantSecretNames` only checks `SecretKeyRef.Name`. The next test row at line 558 uses two different Keys (`"client-secret"` for `ClientSecretRef` and `"token"` for the DCR `InitialAccessTokenRef`) — the perfect case to also pin Key, but the assertion at line 600 ignores it. A refactor that swapped the Keys would still pass the test.

Add a parallel `wantSecretKeys []string` field to the test struct and assert `envVars[i].ValueFrom.SecretKeyRef.Key` alongside `Name`.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -410,6 +435,81 @@ type OAuth2UpstreamConfig struct {
...
+	// issues a single GET against this URL (no well-known-path fallback) and
+	// reads registration_endpoint, authorization_endpoint, token_endpoint,
+	// token_endpoint_auth_methods_supported, and scopes_supported from the
+	// response.
+	// Mutually exclusive with RegistrationEndpoint.
+	// MaxLength bounds CEL cost estimation on the parent struct's
+	// XValidation rule and matches the conventional URL length cap.
+	// +optional
+	// +kubebuilder:validation:Pattern=`^https?://.*$`
+	// +kubebuilder:validation:MaxLength=2048
```
tgrunnagle, line 498: **F6 / LOW / consensus 8/10 — `MaxLength=2048` doc-comment rationale is technically incorrect** _(raised by CRD-specialist)_

The comment above ("MaxLength bounds CEL cost estimation on the parent struct's XValidation rule") implies the cap is needed because of CEL cost. The parent rule is `has(self.discoveryUrl) != has(self.registrationEndpoint)` — `has()` is O(1) and the apiserver's CEL cost estimator does not scale by string length. The cap is justifiable as a defensive size cap (DoS, etcd budget, regex Pattern evaluation cost) but the stated reason is wrong. The same incorrect rationale is propagated verbatim into both CRD YAMLs and `docs/operator/crd-api.md`.

Fix the source comment, then re-run `task operator-manifests` and `task crdref-gen` so the propagated text picks up the correction. Same fix needed at line 489 for `RegistrationEndpoint`.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
```diff
@@ -571,6 +571,125 @@ func TestMCPExternalAuthConfig_validateUpstreamProvider(t *testing.T) {
...
+			},
+			expectErr: true,
+			// Pin the scoped prefix so a rename of the oauth2Config label surfaces as a test failure.
+			errMsg: "oauth2Config: exactly one of clientId or dcrConfig must be set",
+		},
+		{
+			// Regression guard: conversion only maps DCRConfig in the OAuth2
+			// branch, so an OIDC-typed provider carrying OAuth2Config/DCRConfig
+			// must be rejected at validate time rather than silently dropped.
+			name: "OIDC provider with OAuth2Config carrying DCRConfig is rejected",
```
tgrunnagle: **F7 / LOW / consensus 8/10 — Test name claims to assert DCR-specific rejection but actually fires the OIDC discriminator** _(raised by test-specialist)_

The case has `Type: oidc`, `OIDCConfig: nil`, `OAuth2Config: ...`. The first check in `validateUpstreamProvider` (types.go:1049) — `(provider.OIDCConfig == nil) == (provider.Type == UpstreamProviderTypeOIDC)` — fires immediately and returns `"oidcConfig must be set when type is 'oidc'..."`. The DCR XOR check at line 1060 is never reached. Removing `DCRConfig` from the test input produces the identical error.

The case adds no DCR-specific coverage beyond the existing OIDC-discriminator test at line 192. Either rename it to reflect what it actually checks, or reshape it to a true DCR-leakage guard by also setting a valid `OIDCConfig` so the OIDC discriminator passes — then the OAuth2 discriminator at line 1052 would catch the spurious OAuth2 payload.

cmd/thv-operator/pkg/controllerutil/authserver.go
```diff
@@ -661,59 +726,121 @@ func buildUpstreamRunConfig(
...
+}
+
+// buildOAuth2UpstreamRunConfig converts a CRD OAuth2UpstreamConfig to the
+// runtime representation. It rejects malformed ClientID/DCRConfig pairs
+// before producing a RunConfig — mirroring the CEL XValidation rules on
+// OAuth2UpstreamConfig / DCRUpstreamConfig — so objects that reached etcd
+// without passing admission (stored-before-CEL, apiserver patches, test
+// fixtures bypassing validation) fail at reconcile time rather than at
+// authserver startup.
+func buildOAuth2UpstreamRunConfig(
```
tgrunnagle, line 799: **F8 / LOW / consensus 8/10 — Redundant params and asymmetric signatures with `buildOIDCUpstreamRunConfig`** _(raised by Go-architecture-specialist)_

`buildOAuth2UpstreamRunConfig(provider *UpstreamProviderConfig, b *upstreamSecretBinding, resourceURL string)` — but `b.Provider` is the same `provider`. The caller (line 716, `buildUpstreamRunConfig`) sets `provider := b.Provider` and passes both. Inside the body, `provider` is only read once for `provider.OAuth2Config`, which is reachable via `b.Provider.OAuth2Config`. The aliasing invariant has no enforcement; a future refactor pulling `b` from a different source could pass a non-matching pair.

Separately, `buildOIDCUpstreamRunConfig` (line 747) takes `(cfg *OIDCUpstreamConfig, clientSecretEnvVar string, resourceURL string)` — config + scalar env-var name. Make the OAuth2 helper parallel: drop `provider`, pass `cfg *OAuth2UpstreamConfig` plus two scalar env-var names (`clientSecretEnvVar` and `initialAccessTokenEnvVar`). Adjust the call site at line 733 to pass `b.EnvVarName` and `b.DCRInitialAccessTokenEnvVar` directly.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -273,6 +273,21 @@ const (
...
+// set OAuth2Config. The CEL rule below enforces the pairing at admission; the
+// matching Go-level check in validateUpstreamProvider provides defense-in-depth
+// for stored objects.
+//
+// The rule is structured as a chain of equality checks ending in an explicit
+// `false`, so adding a new UpstreamProviderType value without extending this
+// rule fails admission instead of silently demanding the OAuth2 shape. When
+// adding a new type, extend both this rule and validateUpstreamProvider.
+//
+// +kubebuilder:validation:XValidation:rule="self.type == 'oidc' ? (has(self.oidcConfig) && !has(self.oauth2Config)) : self.type == 'oauth2' ? (has(self.oauth2Config) && !has(self.oidcConfig)) : false",message="type must be 'oidc' or 'oauth2'; oidcConfig must be set when type is 'oidc' and oauth2Config must be set when type is 'oauth2' (and the other must not be set)"
```
tgrunnagle, line 288: **F9 / LOW / consensus 7/10 — CEL discriminator and Go validator emit different messages for the same root cause** _(raised by CRD-specialist)_

CEL (this line): `"type must be 'oidc' or 'oauth2'; oidcConfig must be set when type is 'oidc' and oauth2Config must be set when type is 'oauth2' (and the other must not be set)"`.
Go (line 1050): `"oidcConfig must be set when type is 'oidc' and must not be set otherwise"`.

Same root cause, two different diagnostics depending on which gate fires first. Pick one wording (the CEL one is more informative) and align both. The accompanying test at line 326 of types_test.go pins the Go-validator wording, so updating either side requires syncing the test.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -354,6 +369,14 @@ type OIDCUpstreamConfig struct {
 
 // OAuth2UpstreamConfig contains configuration for pure OAuth 2.0 providers.
 // OAuth 2.0 providers require explicit endpoint configuration.
+//
+// Exactly one of ClientID or DCRConfig must be set: ClientID is used when the
+// client is pre-provisioned out of band, DCRConfig enables RFC 7591 Dynamic
+// Client Registration at runtime.
+//
+// +kubebuilder:validation:XValidation:rule="(has(self.clientId) && size(self.clientId) > 0) ? !has(self.dcrConfig) : has(self.dcrConfig)",message="exactly one of clientId or dcrConfig must be set"
```
tgrunnagle, line 394: **F10 / LOW / consensus 7/10 — `dcrConfig: {}` traverses to the inner-rule message** _(raised by CRD-specialist)_

The OAuth2 XOR rule treats `clientId: ""` as absent (`size(self.clientId) > 0`) but treats `dcrConfig: {}` as present (`has(self.dcrConfig)` returns true for an empty object). For input `{ clientId: "", dcrConfig: {} }` the outer rule passes and the inner DCR XOR fires with `"exactly one of discoveryUrl or registrationEndpoint must be set"` — actionable, but not as informative as the outer-level `"exactly one of clientId or dcrConfig must be set"`.

Polish only. If you want the outer error to fire first, add a non-empty subfield check (e.g. `has(self.dcrConfig.discoveryUrl) || has(self.dcrConfig.registrationEndpoint)`). Be aware that inflates CEL cost; document the layered behavior either way.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -410,6 +435,81 @@ type OAuth2UpstreamConfig struct {
...
+	// +kubebuilder:validation:MaxLength=255
+	SoftwareID string `json:"softwareId,omitempty"`
+
+	// SoftwareStatement is the RFC 7591 "software_statement" JWT asserting
+	// metadata about the client software, signed by a party the authorization
+	// server trusts. Bounded to 4096 characters to prevent unbounded
+	// user input from inflating the CR beyond etcd object limits.
+	// +optional
+	// +kubebuilder:validation:MaxLength=4096
+	SoftwareStatement string `json:"softwareStatement,omitempty"`
```
tgrunnagle, line 547: **F11 / LOW / consensus 7/10 — softwareStatement is plaintext in CR spec; no Secret-ref alternative** _(raised by security-specialist)_

RFC 7591 §2.3 software_statements are signed JWTs that authenticate the client software to the AS. Some deployments treat them as bearer-style attestations and rotate them like secrets. The current placement makes them visible to anyone with `get/list/watch` on `MCPExternalAuthConfig`, surfaces them in `kubectl get -o yaml`, and stores them in etcd backups in plaintext. The existing pattern in this file uses SecretKeyRef for credentials.

Consider an alternate `softwareStatementRef *SecretKeyRef` field, mutually exclusive with the inline `softwareStatement`. At minimum, document explicitly that the inline value is treated as non-sensitive (signed but world-readable).

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -956,10 +1056,62 @@ func (*MCPExternalAuthConfig) validateUpstreamProvider(index int, provider *Upst
 		return fmt.Errorf("%s: unsupported provider type: %s", prefix, provider.Type)
 	}
 
+	// Validate OAuth2-specific DCR / ClientID constraints (defense-in-depth with CEL).
+	if provider.Type == UpstreamProviderTypeOAuth2 && provider.OAuth2Config != nil {
```
tgrunnagle: **F12 / LOW / consensus 7/10 — Dead defensive nil-check on `provider.OAuth2Config`** _(raised by Go-architecture-specialist)_

The pairing check at line 1052 already guarantees: if Type is OAuth2 and we're past that check, `OAuth2Config != nil`. The second nil-check here is dead.

```suggestion
	if provider.Type == UpstreamProviderTypeOAuth2 {
```

cmd/thv-operator/pkg/controllerutil/authserver.go
```diff
@@ -55,41 +55,122 @@ const (
...
+//   - OIDC providers: only `clientSecretRef` is ever non-nil.
+//     `initialAccessTokenRef` is always nil because DCR is OAuth2-only and
+//     OAuth2Config must be nil for OIDC-typed providers.
+//   - OAuth2 providers: `clientSecretRef` and `initialAccessTokenRef` are
+//     independent — either, both, or neither may be non-nil.
+//   - Any other (currently unreachable) Type value: both are nil.
+//
+// Callers must not rely on the third bullet to mask an admission-bypassing
+// object — `BuildAuthServerRunConfig` is the reconcile-time backstop for that.
+func extractUpstreamSecretRefs(
```
tgrunnagle, line 139: **F13 / LOW / consensus 7/10 — Named-return convention is inconsistent with sibling helpers** _(raised by security + Go specialists)_

`extractUpstreamSecretRefs` declares named returns (`clientSecretRef, initialAccessTokenRef *mcpv1beta1.SecretKeyRef`) but ends with an explicit `return clientSecretRef, initialAccessTokenRef`, not a naked return. The named returns only document the result shape; siblings (`buildOIDCUpstreamRunConfig`, `buildOAuth2UpstreamRunConfig`, `buildDCRUpstreamRunConfig`) all use unnamed returns. Pick one convention.

cmd/thv-operator/pkg/controllerutil/authserver_test.go
```diff
@@ -497,6 +498,86 @@ func TestGenerateAuthServerEnvVars(t *testing.T) {
 			},
 			wantSecretNames: []string{"okta-secret", "github-secret"},
 		},
+		{
+			name: "OAuth2 provider with DCR initial access token ref emits separate env var",
```
tgrunnagle, line 508: **F14 / LOW / consensus 7/10 — Multi-upstream + long-name env-var coverage missing** _(raised by test-specialist)_

All new DCR rows test single-provider scenarios. The realistic case for `VirtualMCPServer` is multiple OAuth2 upstreams each with their own DCR. A regression that hashed names instead of sanitize-and-uppercase would not be caught by the current table.

Add a row with two OAuth2 providers, each with `InitialAccessTokenRef` set, asserting two distinct env vars with distinct Names and distinct per-upstream secret refs. Optionally add a row exercising a name near the 63-char DNS-label boundary.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
```diff
@@ -571,6 +571,125 @@ func TestMCPExternalAuthConfig_validateUpstreamProvider(t *testing.T) {
 			},
```
tgrunnagle, line 593: **F15 / LOW / consensus 7/10 — Boundary length tests missing for softwareStatement and URLs** _(raised by test-specialist)_

The CRD imposes `MaxLength=4096` on `softwareStatement` and `MaxLength=2048` on URLs. Test cases use only short fixed strings. If the Go validator (`ValidateOAuth2DCRConfig`) is the documented reconcile-time backstop, length should be enforced there too — and tested at the boundary (cap, cap+1, cap-1). If only CEL enforces lengths, the asymmetry is itself a defense-in-depth gap worth calling out.

cmd/thv-operator/pkg/controllerutil/authserver_test.go
```diff
@@ -1190,6 +1407,109 @@ func TestBuildAuthServerRunConfig(t *testing.T) {
 	}
 }
 
+// TestBuildAuthServerRunConfig_InvalidDCR verifies that BuildAuthServerRunConfig
+// rejects OAuth2 upstreams whose ClientID / DCRConfig pairing violates the
+// mutual-exclusivity invariants (matching the CEL XValidation rules on
+// OAuth2UpstreamConfig and DCRUpstreamConfig). This is the reconcile-time
+// defense-in-depth against stored objects that bypassed admission.
+func TestBuildAuthServerRunConfig_InvalidDCR(t *testing.T) {
```
tgrunnagle, line 1451: **F16 / LOW / consensus 7/10 — `TestBuildAuthServerRunConfig_InvalidDCR` has 3 redundant cases** _(raised by test-specialist)_

All four cases here mirror cases in the `validateUpstreamProvider` table (types_test.go lines 242, 260, 278, 293). Both call paths route through `ValidateOAuth2DCRConfig` — same validator. The unique value this function adds is (a) the outer `upstream %q:` wrap and (b) the upstream-name-appears-once invariant — neither requires four cases.

Reduce to a single representative case (e.g. "both ClientID and DCRConfig set") that asserts the outer wrap, the substring count, and the error content. Or document explicitly per-case what each one uniquely contributes that the types-level table doesn't.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
tgrunnagle: Addressed in 1fb61314.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in 01ae2328.

cmd/thv-operator/pkg/controllerutil/authserver.go
tgrunnagle: Addressed in 01ae2328.

cmd/thv-operator/pkg/controllerutil/authserver.go
tgrunnagle: Addressed in 01ae2328.

cmd/thv-operator/pkg/controllerutil/authserver_test.go
tgrunnagle: Addressed in 1b0e7817.

cmd/thv-operator/pkg/controllerutil/authserver_test.go
tgrunnagle: Addressed in 1b0e7817.

cmd/thv-operator/pkg/controllerutil/authserver_test.go
tgrunnagle: Addressed in 1b0e7817.

jhrozek (approved):
LGTM. Defense-in-depth between admission CEL and the reconcile-time validator looks solid, conversion is cleanly factored, and tests cover the validator surface well.

A handful of non-blocking carry-overs worth picking up in a follow-up — none stand in the way of merging this:

- URL regex `^https://[^\s?#]+[^/\s?#]$` on `discoveryUrl`/`registrationEndpoint` permits `@`, so a URL like `https://user:pass@idp.example.com/...` is accepted. Once stored on the CR it's visible to anyone with `get`, and Go's `net/http` will route to the host correctly so it's primarily a credential-in-logs concern. Tightening to exclude `@` in the authority (or a `url.Parse + u.User != nil` check in `ValidateOAuth2DCRConfig`) would close it.
- `SoftwareID` has `MaxLength=255` on the kubebuilder marker but `ValidateOAuth2DCRConfig` doesn't enforce it; the doc comment says it mirrors all caps. Either add the check + a `MaxSoftwareIDLength` constant or narrow the doc claim.
- `ValidateOAuth2DCRConfig` is exported now and dereferences `cfg` on the first line; both current callers nil-check, but a one-line `if cfg == nil { return nil }` would harden the contract.
- `VirtualMCPServerReconciler.validateAuthServerConfig` doesn't call `ValidateOAuth2DCRConfig` even though `MCPExternalAuthConfig.Validate()` does. CEL still catches it at admission so it's strictly a defense-in-depth parity gap.
- In the vMCP converter, DCR errors propagate as a plain wrapped `error` rather than `*SpecValidationError`, so a malformed-DCR vMCP that bypassed admission backs off forever without an `AuthServerConfigValidated=False` condition. Wrapping the converter error makes it consistent with the sibling `ValidateAuthServerIntegration` path.

Tests-wise: missing `registrationEndpoint` boundary cases, no `softwareStatement` at-cap accept case, and no direct unit test for `ValidateOAuth2DCRConfig` (everything reaches it via `validateUpstreamProvider`). Worth picking up next time someone is in this file.

Nice work on the iterative tightening — the doc comments around the layered XOR semantics and the softwareStatement plaintext caveat are exactly what operators need to read at the API surface.

jhrozek:
please ignore the nits, I told claude to skip them during the assisted code walkthrough and there it is posting them on its own, sigh

PR #5079, opened 2026-04-27
Configure rate limits on VirtualMCPServer PR A

### LINKED ISSUE #4552: Configure rate limits on VirtualMCPServer
## User Story

**As a** cluster admin,
**I want to** configure per-user and global rate limits on VirtualMCPServer,
**so that** aggregated backends are protected from abuse.

## Context

See [THV-0057: Rate Limiting for MCP Servers](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0057-rate-limiting.md) for full design details.

## Acceptance Criteria

- [ ] `VirtualMCPServerSpec` includes `rateLimiting` at `spec.rateLimiting` (not `spec.config`)
- [ ] CRD admission validation: `perUser` requires auth enabled
- [ ] Unit: Per-user limits shared across all backends (single bucket per user for the vMCP, not per-backend)
- [ ] Unit: Per-tool limits use post-aggregation names (underscore separator by default)
- [ ] Unit: With optimizer enabled, per-tool limit extracts inner `tool_name` from `call_tool` arguments
- [ ] Unit: Per-user and global limits both enforced on `tools/call`
- [ ] E2E: Deploy VirtualMCPServer with per-user rate limit, send traffic, verify rejection after limit exceeded

## Dependencies

- STORY-001 and STORY-002 (core rate limit middleware and CRD types)

## Out of Scope

- Observability (STORY-004)
- `prompts/get` and `resources/read` (deferred, STORY-005)

Sanskarzz:
PR A — shared→global rename + CRD schema for vMCP
## Summary

Add the `VirtualMCPServer` rate-limiting API surface in line with THV-0057.

This change introduces `spec.config.rateLimiting` support on `VirtualMCPServer`. The field lives on the vMCP runtime config type so the operator can serialize it into the generated `config.yaml` and the vMCP process can read it at startup.

The CRD validation rejects invalid combinations early, including per-user limits without OIDC incoming auth and any rate limiting without Redis session storage. Generated CRDs and operator docs were updated accordingly.

Runtime vMCP enforcement is intentionally split out of this PR and will be handled in a follow-up PR.

Fixes #4552 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [ ] Manual testing (describe below) I will describe full manual testing in follow-up PR

## API Compatibility

 This PR adds optional `VirtualMCPServer.spec.rateLimiting` configuration to the `v1beta1` API. This is an additive, backward-compatible API change.

  Existing `VirtualMCPServer` resources do not need migration. Cluster admins can opt in by adding `spec.rateLimiting` and must configure Redis-backed `spec.sessionStorage` when rate
  limiting is enabled. Per-user limits additionally require OIDC incoming auth.
  
## Changes

- Add `VirtualMCPServer.spec.config.rateLimiting` on the vMCP config type
- Add CEL validation for Redis session storage and OIDC-required per-user limits
- Preserve rate-limit config through CRD-to-vMCP config conversion
- Keep `pkg/ratelimit` using the CRD `RateLimitConfig` type directly
- Add unit coverage for vMCP config serialization and converter passthrough

## Large PR Justification


jerm-dro (changes requested):
Hey Sanskar, thanks for putting this together — the end-to-end shape is right: top-level `spec.rateLimiting` with CEL validation, converter wiring, and the vMCP middleware integration with optimizer-aware tool name resolution. The manual testing scenario in the description is thorough. Requesting changes on two axes before we go further.

## 1. Drop `pkg/ratelimit/config` — use the CRD type directly

The proxyrunner path already uses `*v1beta1.RateLimitConfig` directly throughout `pkg/ratelimit` on main today. vMCP can do the same — just pass the CRD type through. The new package duplicates the CRD types 1:1, still imports `metav1.Duration`, and adds conversion boilerplate (`ToInternal()`, hand-written `DeepCopy`, unused `EffectiveGlobal()` methods) without solving a real problem.

## 2. Split the PR

Even after removing `pkg/ratelimit/config`, this PR bundles several separable changes. At minimum I'd split into:

**PR A — `shared`→`global` rename + CRD schema for vMCP** (pure schema change, no runtime behavior):
- Add `Global` field to `RateLimitConfig` / `ToolRateLimitConfig` with `shared` as deprecated alias
- Add `spec.rateLimiting` to `VirtualMCPServerSpec` with CEL validation
- Converter wiring
- Generated CRDs, docs

**PR B — vMCP rate-limit middleware wiring** (depends on A):
- `buildRateLimitMiddleware` + Redis client lifecycle
- Middleware chain refactoring in `server.go` (this changes execution order — MCP parsing moves before audit — and deserves focused review)
- Optimizer `call_tool` tool-name unwrapping
- Unit tests + E2E test

Sanskarzz:
@jerm-dro  The PR is ready for re-review. I will create a separate PR for `vMCP rate-limit middleware wiring` after the merge of this PR.

pkg/vmcp/config/config.go
```diff
@@ -173,6 +173,14 @@ type Config struct {
...
 	// +optional
 	SessionStorage *SessionStorageConfig `json:"sessionStorage,omitempty" yaml:"sessionStorage,omitempty"`
+
+	// RateLimiting configures request rate limits for tools/call.
+	// Kubernetes users should set VirtualMCPServer.spec.rateLimiting; the operator
+	// copies that top-level field here for the vMCP runtime.
+	// This remains YAML-only so it can be written to the runtime config ConfigMap
+	// without exposing spec.config.rateLimiting in the Kubernetes API.
+	// +optional
+	RateLimiting *thvjson.Map `json:"-" yaml:"rateLimiting,omitempty"`
```
jerm-dro: **`blocker:` Match the proxyrunner pattern — use `*v1beta1.RateLimitConfig` directly.**

`pkg/runner.RunConfig` (`pkg/runner/config.go:169`) already solves this with a typed pointer:

```go
RateLimitConfig *v1beta1.RateLimitConfig `json:"rate_limit_config,omitempty" yaml:"rate_limit_config,omitempty"`
```

The operator wires it in one line at `mcpserver_runconfig.go:263` — no converter, no marshal/unmarshal, no map wrapper.

Let's do the same here.

cmd/thv-operator/api/v1beta1/mcpserver_types.go
```diff
@@ -505,28 +505,32 @@ type SessionStorageConfig struct {
...
-// +kubebuilder:validation:XValidation:rule="has(self.shared) || has(self.perUser) || (has(self.tools) && size(self.tools) > 0)",message="at least one of shared, perUser, or tools must be configured"
+// +kubebuilder:validation:XValidation:rule="has(self.global) || has(self.shared) || has(self.perUser) || (has(self.tools) && size(self.tools) > 0)",message="at least one of global, shared, perUser, or tools must be configured"
 //
 //nolint:lll // CEL validation rules exceed line length limit
 type RateLimitConfig struct {
-	// Shared is a token bucket shared across all users for the entire server.
+	// Global is a token bucket shared across all users for the entire server.
 	// +optional
-	Shared *RateLimitBucket `json:"shared,omitempty"`
+	Global *RateLimitBucket `json:"global,omitempty" yaml:"global,omitempty"`
```
jerm-dro: **`blocker:` Drop the `shared` → `global` rename.**

It's not necessary for the functional changes we're after — vMCP can use `shared` as-is. Removing the rename drops the deprecated alias, the resolution helpers, the dual Redis keyspaces, and the doubled CRD/deepcopy churn.

jerm-dro (changes requested):

pkg/vmcp/config/config.go
Sanskarzz: oops i missed it. I updated this to remove the map wrapper/conversion path and use the typed rate-limit config directly, matching the proxyrunner approach.

cmd/thv-operator/api/v1beta1/mcpserver_types.go
Sanskarzz: Done now the config now keeps using shared as-is for both MCPServer and VirtualMCPServer.

cmd/thv-operator/api/v1beta1/mcpserver_types.go
```diff
@@ -503,69 +505,16 @@ type SessionStorageConfig struct {
...
-	// Shared token bucket for this specific tool.
-	// +optional
-	Shared *RateLimitBucket `json:"shared,omitempty"`
-
-	// PerUser token bucket configuration for this tool.
-	// +optional
-	PerUser *RateLimitBucket `json:"perUser,omitempty"`
-}
+// +gendoc
+type ToolRateLimitConfig = vmcpconfig.ToolRateLimitConfig
```
jerm-dro: can you avoid moving these structs and creating the type aliases?

Sanskarzz: I have moved the rate limit structs back into v1beta1 as real CRD types and removed the aliases instead of trying to patch the generated docs.

Sanskarzz: @jerm-dro  Getting error in [CI](https://github.com/stacklok/toolhive/actions/runs/25136567420/job/73676295578?pr=5079) but it looks unrelated to the scope of this PR. Could you please re-run the workflow?

jerm-dro: Done!

Sanskarzz: Again, this new CI [error](https://github.com/stacklok/toolhive/actions/runs/25150311435/job/73719908982?pr=5079) looks unrelated.

Sanskarzz: Now it's fixed

cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig.go
```diff
@@ -130,6 +135,12 @@ func (r *VirtualMCPServerReconciler) ensureVmcpConfigConfigMap(
 	return nil
 }
 
+type vmcpConfigMapConfig struct {
+	vmcpconfig.Config `yaml:",inline"`
+
+	RateLimiting *mcpv1beta1.RateLimitConfig `yaml:"rateLimiting,omitempty"`
+}
```
jerm-dro: Why introduce this new type? Why not put ratelimiting on `vmcpconfig.Config`? 

As implemented, I don't think the RateLimiting config ever makes it into vmcp's server.go. The CRD field  doesn't survive the conversion step, since there's nowhere to write the RateLimiting config to.

Sanskarzz: You’re right. I had added an operator-side ConfigMap wrapper, but that was incomplete because vMCP loads the file into `vmcpconfig.Config`, so `rateLimiting` would be ignored before reaching `server.go`.
I removed the wrapper and reverted this path to marshal `vmcpconfig.Config` directly. Since this PR is now scoped to the CRD/schema validation side, I’ll leave the actual runtime propagation/enforcement for PR B, where we can wire the config into vMCP and the middleware properly.

jerm-dro: Almost there. You'll want to put `RateLimiting *RateLimitConfig` on the config type, [here](https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/config/config.go#L95), rather than on the CRD type, `VirtualMCPServerSpec`. The config type is already within the `VirtualMCPServerSpec`. Putting the `RateLimiting` field on the config struct means that configuration information will be available to vMCP when it's running. vMCP and most other toolhive services don't read CRDs at runtime, they read config types. If the field is not on the config, then it cannot enable rate limiting.

Sanskarzz: Thanks, that makes sense. I was treating `rateLimiting` as a VMCP CRD-level field, but I missed that vMCP only sees the serialized runtime config, not the CRD object directly.

 I moved it onto `pkg/vmcp/config.Config` as `spec.config.rateLimiting`, removed the top-level `VirtualMCPServerSpec.rateLimiting` field, updated the VMCP CEL validation to reference
`config.rateLimiting`, and regenerated CRDs/docs/deepcopy.

 I also added converter coverage to verify the field survives the `VirtualMCPServer` -> vMCP config path.

jerm-dro: Sorry, I should've been more exact. I didn't exactly mean `RateLimiting *RateLimitConfig` on the `pkg/vmcp/config.Config`. I meant you should reuse the existing `v1beta1.RateLimitConfig`. You shouldn't have to introduce the types: `RateLimitConfig`, `RateLimitBucket`, and `ToolRateLimitConfig` in pkg/vmcp/config/config.go. Those can be deleted.

Sanskarzz: I looked at this and agree we should not duplicate `RateLimitConfig`, `RateLimitBucket`, and `ToolRateLimitConfig` in `pkg/vmcp/config`.

One implementation detail: `pkg/vmcp/config` cannot directly use `v1beta1.RateLimitConfig`, because `v1beta1.VirtualMCPServerSpec` already embeds `pkg/vmcp/config.Config`, so importing `v1beta1` from `pkg/vmcp/config` creates an import cycle:

  `pkg/vmcp/config -> cmd/thv-operator/api/v1beta1 -> pkg/vmcp/config`

To avoid that while still keeping a single concrete type, I’m thinking of moving the rate-limit structs into a lower-level shared package, e.g. `pkg/ratelimit/types`, and then aliasing them from `v1beta1`:

 `type RateLimitConfig = ratelimittypes.RateLimitConfig`

That lets existing code keep using `v1beta1.RateLimitConfig`, while `pkg/vmcp/config.Config.RateLimiting` uses the same underlying type without duplicating structs.

Does that package split sound acceptable, or would you prefer a different home for the shared rate-limit config types?

jerm-dro: That sounds reasonable. Thank you for explaining.

jerm-dro (approved):
Thanks for volunteering and iterating on this 😄

Sanskarzz:
@jerm-dro Welcome! Once this PR is merged, I’ll start working on the vMCP rate-limiting middleware wiring.

PR #5082, opened 2026-04-28
Return *oauth2.RetrieveError from tokenexchange

jhrozek:
## Summary

- **Why**: Token exchange currently returns a private `oAuthError` whose only output is a formatted string. Callers cannot inspect the RFC 6749 §5.2 fields without parsing the message. This PR aligns the error shape with the upcoming JWT Bearer grant, both of which will share helpers in `pkg/oauth` — `*oauth2.RetrieveError` from `golang.org/x/oauth2` is the library-standard surface for non-2xx token endpoint responses.
- **What**: Replace `oAuthError` with `*oauth2.RetrieveError`. `validateResponseStatus` now takes `*http.Response` so it can attach the full response and best-effort-parse the body. On non-conformant bodies (no `error` field, e.g. proxy HTML 5xx), the raw body is logged at debug level and cleared from the returned error so `RetrieveError.Error()` cannot interpolate arbitrary upstream content (HTML, hostnames, stack traces) into wrapped log strings — same two-tier pattern already used by `formatOAuth2Error` in `pkg/authserver`. `parseTokenExchangeResponse` wraps `json.Unmarshal` failures with `%w`.

## Type of change

(Note: the body-redaction defense above is a deliberate hardening of an information-disclosure path the previous code already had via the implicit fmt error string; user-visible structured behavior is unchanged.)

## Test plan

- [x] Unit tests (`task test`) — `pkg/auth/tokenexchange` passes; two pre-existing failures on `main` in unrelated packages (`pkg/skills/client` hardcoded port, `pkg/workloads` empty-group filter) are not introduced by this PR.

`TestExchangeToken_HTTPErrorResponses` was migrated from substring matching on `err.Error()` to `errors.As(err, &*oauth2.RetrieveError{})` and now asserts on `Response.StatusCode`, `ErrorCode`, `ErrorDescription`, and `Body` (preserved on conformant responses, cleared on the 503 non-JSON case).

## API Compatibility

No. Internal error type change. Operators who scrape the existing `slog.Debug` log fields (`oauth_error_code`, `description`, `status`, `body_length`) will see identical field names; the body-content debug field is new and additive.

- The commit is intentionally narrow and isolated from the `pkg/oauth` shared-helpers migration (subsequent commits 5b/5c) so a future bisect can distinguish "error shape regressed" from "plumbing regressed."
- Upstream caller `pkg/vmcp/auth/strategies/tokenexchange.go:142` wraps with `%w`, so `errors.As` continues to work and the leading `"token exchange failed: "` prefix is preserved.
- The body-redaction behavior was added in response to review feedback on the original commit — see commit message for rationale and the `formatOAuth2Error` precedent.

### DIFF
--- a/pkg/auth/tokenexchange/exchange.go
+++ b/pkg/auth/tokenexchange/exchange.go
@@ -7,7 +7,6 @@ package tokenexchange
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -65,34 +64,6 @@ func NormalizeTokenType(tokenType string) (string, error) {
 	default:
 		return "", fmt.Errorf("invalid token type %q: must be one of: access_token, id_token, jwt", tokenType)
 	}
-}
-
-// oAuthError represents an OAuth 2.0 error response as defined in RFC 6749 Section 5.2.
-type oAuthError struct {
-	Error            string `json:"error"`
-	ErrorDescription string `json:"error_description,omitempty"`
-	ErrorURI         string `json:"error_uri,omitempty"`
-	StatusCode       int    `json:"-"`
-}
-
-func (e *oAuthError) String() string {
-	if e.ErrorURI != "" {
-		return fmt.Sprintf("OAuth error %q (status %d): see %s", e.Error, e.StatusCode, e.ErrorURI)
-	}
-	return fmt.Sprintf("OAuth error %q (status %d)", e.Error, e.StatusCode)
-}
-
-// parseOAuthError attempts to parse an OAuth error response from the given response body.
-func parseOAuthError(statusCode int, body []byte) *oAuthError {
-	var oauthErr oAuthError
-	if err := json.Unmarshal(body, &oauthErr); err != nil {
-		return nil
-	}
-	if oauthErr.Error == "" {
-		return nil
-	}
-	oauthErr.StatusCode = statusCode
-	return &oauthErr
 }
 
 // defaultHTTPClient is the default HTTP client used for token exchange requests.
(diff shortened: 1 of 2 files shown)


pkg/auth/tokenexchange/exchange.go
```diff
@@ -502,37 +473,58 @@ func executeTokenExchangeRequest(client *http.Client, req *http.Request) ([]byte
...
+		retrieveErr.ErrorURI = oauthErr.ErrorURI
+	}
+
+	if retrieveErr.ErrorCode != "" {
+		slog.Debug("Token exchange OAuth error",
+			"oauth_error_code", retrieveErr.ErrorCode,
+			"description", retrieveErr.ErrorDescription)
+	} else {
+		slog.Debug("Token exchange failed", "status", resp.StatusCode, "body_length", len(body), "body", string(body))
+		retrieveErr.Body = nil
```
tgrunnagle: Question: should we clear the `retrieveErr.Body` in either case?

tgrunnagle (dismissed):

pkg/auth/tokenexchange/exchange.go
jhrozek: Good catch. Agreed — clearing it in only one branch was a half-measure. Pushed a follow-up commit that drops the conditional, so `Body = nil` runs unconditionally. The RFC 6749 §5.2 fields are already extracted onto `RetrieveError`, so callers using `errors.As` lose nothing, and the full body content is still preserved in `slog.Debug` for ops. Also matches Ory Hydra's pattern of never surfacing raw upstream error bodies through its public error type.

ChrisJBurns (approved):

tgrunnagle (approved):

PR #5083, opened 2026-04-28
Strengthen tool detection with binary presence check

### LINKED ISSUE #5070: llm: strengthen tool detection with binary presence check
## Background

`DetectedLLMGatewayClients()` in `pkg/client/llm_gateway.go` currently detects an installed AI tool by checking whether the tool's settings *directory* exists on disk. This was flagged in PR #5065 review (JAORMX) as false-positive prone:

> \`~/.claude\` and \`~/.gemini\` hang around from prior installs or unrelated tooling. Combine with \`exec.LookPath("claude")\` / \`exec.LookPath("gemini")\`.

## Problem

A settings directory like `~/.claude` or `~/.gemini` can exist even when the tool is no longer installed (leftover from a prior install, created by unrelated tooling, or seeded by `thv client register`). In that case `thv llm setup` will attempt to patch `settings.json` for a tool the user doesn't actually have, producing a confusing no-op or stale config file.

## Proposed fix

Extend `clientAppConfig` with an optional `BinaryName string` field. `DetectedLLMGatewayClients()` should require **both** conditions:
1. Settings directory exists (`os.Stat` of `filepath.Dir(settingsPath)` succeeds) — current check.
2. Binary is on `$PATH` (`exec.LookPath(cfg.BinaryName)` succeeds) — new check, skipped when `BinaryName` is empty.

Populate `BinaryName` for tools where a canonical binary name is well-known:

| Client | `BinaryName` |
|---|---|
| `claude-code` | `claude` |
| `gemini-cli` | `gemini` |
| `cursor` | `cursor` |
| `vscode` | `code` |
| `xcode` | *(skip — macOS app, not on PATH)* |

Note: `IsClientInstalled()` (MCP detection) has the same limitation. Fixing `DetectedLLMGatewayClients` first is the right scope for this issue; a follow-up can extend `IsClientInstalled` consistently.

## References

- PR #5065 review comment by JAORMX

yrobla:
## Summary

DetectedLLMGatewayClients() previously detected an installed AI tool by checking whether its settings directory exists on disk. This produced false positives when directories like ~/.claude or ~/.gemini were left over from a prior uninstall or created by unrelated tooling — causing `thv llm setup` to attempt patching config files for tools the user no longer has installed.

- Add an optional `LLMBinaryName` field to `clientAppConfig`. When set, `DetectedLLMGatewayClients()` now requires **both** the settings directory and the binary on `$PATH` (via `exec.LookPath`) before considering a tool installed.
- Tools not on `$PATH` (e.g. Xcode, a macOS GUI app) leave `LLMBinaryName` empty and fall back to the directory-only check.
- Populate `LLMBinaryName` for `claude-code` (`claude`), `gemini-cli` (`gemini`), `cursor` (`cursor`), `vscode` (`code`), and `vscode-insider` (`code`).
- Inject `lookPath` as a field on `ClientManager` (defaulting to `exec.LookPath`) to keep tests parallel and self-contained.

Fixes #5070

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## API Compatibility

Yes. Users who have leftover `~/.claude`, `~/.gemini`, or `~/.cursor` directories from a prior uninstall will no longer see those tools incorrectly reported as detected by `thv llm setup`. Only tools with both a settings directory **and** their binary on `$PATH` are treated as installed.

The binary check is intentionally skipped when `LLMBinaryName` is empty (e.g. Xcode), preserving the existing directory-only detection for GUI tools that are not on `$PATH`.


pkg/client/config.go
```diff
@@ -317,6 +322,7 @@ var supportedClientIntegrations = []clientAppConfig{
 		SkillsProjectPath: []string{".github", "skills"},
 		// LLM gateway: patches settings.json (same dir as mcp.json, different file)
 		LLMGatewayMode:     "proxy",
+		LLMBinaryName:      "code",
```
aponcedeleonch, line 337: Quick question: VS Code Insiders ships its CLI shim as `code-insiders` rather than `code`. If a user has only Insiders installed (no stable VS Code), `lookPath("code")` will miss it and the integration won't be detected. Should this be `"code-insiders"`? Worth confirming on a machine that has only Insiders installed.

aponcedeleonch (approved):
LGTM. The `lookPath` injection point is clean and the doc comment matches the new behavior. Left one question inline about the VS Code Insiders binary name, doesn't block approval.

yrobla:
Good catch @aponcedeleonch — opened #5089 to track the `code-insiders` fix. I'll address it in a follow-up PR once verified on a machine with only Insiders installed.

Issue #2728, opened 2025-11-25
Implement Client ID Metadata Documents (CIMD) for OAuth Client Registration


amirejaz:
## Background

The MCP Authorization Specification (2025-11-25) introduces **Client ID Metadata Documents (CIMD)** as the **default** method for OAuth client registration, replacing Dynamic Client Registration (DCR) as the primary approach. This change addresses several security and operational concerns with DCR:

- **Security**: DCR can be vulnerable to Denial of Service (DoS) attacks and requires complex security logic
- **Operational**: Most Authorization Servers (AS) don't support DCR, forcing developers to implement workarounds
- **Policy**: CIMD enables AS to make policy decisions based on client metadata URLs (e.g., domain-based allow/deny lists)

**Reference**: [What's New In The 2025-11-25 MCP Authorization Spec](https://den.dev/blog/mcp-november-authorization-spec/)

## Current State

ToolHive currently uses **Dynamic Client Registration (RFC 7591)** as implemented in:
- `pkg/auth/oauth/dynamic_registration.go` - DCR request/response handling
- `pkg/auth/discovery/discovery.go` - Integration with OAuth flow

The current flow:
1. When no `client_id` is provided, ToolHive attempts DCR
2. POSTs registration request to `registration_endpoint` from discovery document
3. Receives `client_id` and `client_secret` from AS
4. Uses these credentials for OAuth flow

## Specification Changes

According to the new spec:

1. **CIMD is now the default** - Every client has its own metadata available at a pre-defined URL
2. **DCR is now MAY (not SHOULD)** - Maintained for backwards compatibility only
3. **Priority order for client registration**:
   1. Pre-registration (if available)
   2. CIMD-based approach
   3. DCR-based approach (fallback)
   4. User-provided client details (manual entry)

### 1. CIMD Support

Implement Client ID Metadata Document support:

- **CIMD Document Structure**: Each client must host a metadata document at a well-known URL
- **Document Fetching**: Fetch and parse CIMD from client's metadata URL
- **Validation**: Validate CIMD document structure and required fields
- **Security**: Implement security checks for SSRF risks, `localhost` redirect URI risks, and trust policies

**Key Components**:
- CIMD document parser/validator
- HTTP client for fetching CIMD documents
- Security validation (SSRF protection, domain-based policies)
- Integration with existing OAuth flow

### 2. Update Registration Priority

Modify `pkg/auth/discovery/discovery.go` to implement the priority order:

```go
// Priority order:
[...]

PR #5085, opened 2026-04-28
Support CIMD as preferred OAuth client registration for thv run

### LINKED ISSUE #4826: Prefer CIMD over DCR when the proxy acts as OAuth client to remote MCP servers
As a user running `thv proxy` to connect to a remote MCP server that requires authentication, I want the proxy to use CIMD when the remote authorization server supports it, so the proxy follows the MCP spec's preferred client registration mechanism and doesn't depend on the remote AS supporting DCR.

## Background

When `thv proxy` connects to a remote MCP server, the **proxy itself is the OAuth client**. At startup, it discovers auth requirements from the remote server (RFC 9728), performs DCR with the remote AS, obtains tokens, and injects `Authorization: Bearer` headers on all proxied requests. VS Code / Claude Code never interacts with the remote AS -- it only talks to the local proxy.

Today the proxy always uses DCR (RFC 7591). The MCP spec prefers CIMD over DCR ([draft-ietf-oauth-client-id-metadata-document](https://datatracker.ietf.org/doc/html/draft-parecki-oauth-client-id-metadata-document-03)). Some future AS implementations may support CIMD but not DCR.

## How CIMD works for the proxy

The proxy uses a **static, publicly-hosted metadata document URL** as its `client_id`. The remote AS fetches the document from that URL to learn about the proxy. The proxy does not serve the document itself -- it's a static JSON file hosted by the ToolHive project, following the same pattern VS Code uses (`productService.authClientIdMetadataUrl` in `product.json`).

The `redirect_uris` in the document point to loopback addresses for the local browser-based auth flow. This sidesteps the localhost reachability problem entirely: the AS only needs to reach the static metadata URL (public), not the proxy's callback server (localhost).

## Acceptance criteria

- [ ] After discovering a remote AS, the proxy checks for `client_id_metadata_document_supported: true` in the AS metadata
- [ ] When CIMD is supported, the proxy uses the built-in metadata document URL as `client_id` in authorize and token requests (no `/oauth/register` call)
- [ ] When CIMD is not supported, falls back to DCR (current behavior, no regression)
- [ ] If the remote AS rejects the CIMD `client_id`, falls back to DCR gracefully
- [ ] Registration priority: stored credentials > CIMD > DCR

### Detection

- Parse `client_id_metadata_document_supported` from the remote AS metadata after RFC 8414/OIDC discovery
- Same `ClientIDMetadataDocumentSupported` field on `AuthorizationServerMetadata` added in #4825

### Client metadata document (static, project-hosted)

- Hosted by the ToolHive project at a stable public HTTPS URL (e.g., `https://toolhive.dev/.well-known/oauth-client/thv-proxy`)
- Built-in as a default in the proxy binary -- no user configuration required (same pattern as VS Code's `authClientIdMetadataUrl` in `product.json`)
- Document contents:
  ```json
  {
    "client_id": "https://toolhive.dev/.well-known/oauth-client/thv-proxy",
    "client_name": "ToolHive Proxy",
    "redirect_uris": [
      "http://127.0.0.1/callback",
      "http://127.0.0.1:33419/callback"
    ],
    "grant_types": ["authorization_code", "refresh_token"],
    "response_types": ["code"],
    "token_endpoint_auth_method": "none"
  }
  ```
- Two loopback redirect URIs following VS Code's pattern (`src/vs/base/common/oauth.ts:869-878`): one without port (RFC 8252 Section 7.3 port-agnostic matching) and one with a fixed default port (for servers that do exact matching despite the spec)

### Loopback callback server

- Try binding to a fixed default port first (e.g., 33419)
- Fall back to OS-assigned port 0 if the default is busy (same approach as VS Code at `loopbackServer.ts:152`)
- This keeps the redirect URI stable across runs when possible

### Auth flow changes in `pkg/auth/remote/handler.go`

- After AS discovery, before attempting DCR, check CIMD viability (AS supports it + metadata URL available)
- When using CIMD: pass the metadata document URL as `client_id` in authorize and token requests. Skip `/oauth/register`.
- When falling back: call `/oauth/register` as today
- Matches the priority chain VS Code uses (`mainThreadAuthentication.ts:160-170`): stored > CIMD > DCR > manual

### Configuration

- Default metadata document URL baked into the binary (overridable via config for forks or custom deployments)
- No user configuration needed for the common case

## Key files

| Area | Files |
|------|-------|
| Metadata type (shared with #4825) | `pkg/oauth/discovery.go` |
| Discovery / AS detection | `pkg/auth/discovery/discovery.go` |
| Remote auth flow | `pkg/auth/remote/handler.go` |
| OAuth flow orchestration | `pkg/auth/discovery/discovery.go` (`PerformOAuthFlow`, `handleDynamicRegistration`) |

## Prior art

VS Code's implementation (`src/vs/workbench/api/browser/mainThreadAuthentication.ts:160-170`, `src/vs/base/common/oauth.ts:869-878`) confirms the pattern:
- Static metadata document URL baked into the product config
- Dual loopback redirect URIs (port-less + fixed-port) for compatibility
- Fixed default callback port with dynamic fallback
- CIMD preferred over DCR when the AS advertises support

amirejaz:
## Summary

Implements Phase 1 of [RFC-0071](https://github.com/stacklok/toolhive-rfcs/pull/71) — CIMD (Client ID Metadata Document) support for \`thv run\` as an OAuth client.

When a remote authorization server advertises \`client_id_metadata_document_supported: true\` in its discovery document, \`thv run\` now presents \`https://toolhive.dev/oauth/client-metadata.json\` as its \`client_id\` instead of performing a Dynamic Client Registration (DCR) round-trip. Falls back to DCR gracefully if the AS rejects the CIMD \`client_id\`.

> **Draft until \`https://toolhive.dev/oauth/client-metadata.json\` is live.** Tracked in [stacklok/infra#4549](https://github.com/stacklok/infra/pull/4549). The code is complete and reviewed; the PR can merge once the URL is reachable.

## Changes

| File | Change |
|------|--------|
| \`pkg/oauthproto/cimd.go\` | New: \`ToolHiveClientMetadataDocumentURL\` constant and \`IsClientIDMetadataDocumentURL\` helper |
| \`pkg/oauthproto/discovery.go\` | Added \`ClientIDMetadataDocumentSupported\` to \`AuthorizationServerMetadata\` |
| \`pkg/auth/discovery/discovery.go\` | Added \`ClientIDMetadataDocumentSupported\` to \`AuthServerInfo\`; \`applyDiscoveryPreCheck\` helper caches discovered endpoints and applies CIMD URL before the DCR gate; \`SkipCIMD\` field on \`OAuthFlowConfig\` prevents pre-check re-applying CIMD on fallback retry |
| \`pkg/auth/remote/config.go\` | Added \`CachedCIMDClientID\`; added clear to \`ClearCachedClientCredentials\` |
| \`pkg/auth/remote/handler.go\` | CIMD \`client_id\` set in \`buildOAuthFlowConfig\`; DCR fallback in \`performOAuthFlow\` via \`isCIMDRejectionError\` (triggers on \`invalid_client\`/\`unauthorized_client\` only; \`SkipCIMD=true\` prevents retry loop) |
| \`test/e2e/oidc_mock.go\` | Added \`client_id_metadata_document_supported: true\` to mock OIDC discovery document for future E2E tests |

## PKCE

PKCE S256 is already enforced unconditionally for all flows in \`createOAuthConfig\` (discovery.go lines 659/675) — CIMD flows inherit it without additional changes.

## Test plan

- [x] Unit tests: \`pkg/oauthproto\`, \`pkg/auth/remote\`, \`pkg/auth/discovery\` — all pass
- [x] \`TestIsCIMDRejectionError\` covers all RFC-specified retry/no-retry cases
- [ ] E2E: pending \`https://toolhive.dev/oauth/client-metadata.json\` going live (infra PR above)

Closes #4826


pkg/oauthproto/cimd.go
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package oauthproto
+
+import "strings"
+
+// ToolHiveClientMetadataDocumentURL is the stable HTTPS URL where ToolHive's
+// client metadata document is hosted. This URL is the client_id ToolHive
+// presents to remote authorization servers that support CIMD.
+// ToolHiveClientMetadataDocumentURL is the stable HTTPS URL where ToolHive's
```
jhrozek: Looks like the lead sentence got pasted twice — both paragraphs start with "ToolHiveClientMetadataDocumentURL is the stable HTTPS URL where ToolHive's client metadata document is hosted." godoc renders this as one block, so the duplicated header shows up in the rendered output. Suggest collapsing into one paragraph:

```go
// ToolHiveClientMetadataDocumentURL is the stable HTTPS URL where ToolHive's
// client metadata document is hosted. ToolHive presents this URL as its
// client_id to remote authorization servers that support CIMD. The URL must
// be live and serving toolhive-client-metadata.json before the feature can
// be used in production.
```

pkg/oauthproto/cimd.go
```diff
@@ -0,0 +1,20 @@
...
+// client metadata document is hosted. This URL must be live and serving
+// toolhive-client-metadata.json before this feature can be used in production.
+const ToolHiveClientMetadataDocumentURL = "https://toolhive.dev/oauth/client-metadata.json"
+
+// IsClientIDMetadataDocumentURL returns true if clientID is an HTTPS URL,
+// indicating it should be treated as a CIMD client_id rather than a DCR-issued UUID.
+func IsClientIDMetadataDocumentURL(clientID string) bool {
```
jhrozek, line 24: Non-blocking. The CIMD draft §3 is stricter than `HasPrefix("https://")` — it requires a non-empty host, a non-empty path, no fragment, no userinfo, and no `..`/`.` path segments. For Phase 1 this is fine, since the predicate only gates the local persistence flag and we always present the hard-coded constant. For Phase 2, when this same predicate decides whether the embedded AS fetches a document submitted by a third party, the loose check will admit malformed inputs the spec forbids.

Two small asks:

1. Expand the doc comment to make the heuristic intentional — something like "any HTTPS URL is treated as a CIMD client_id; DCR-issued IDs are opaque and never match. Do not tighten to an exact match against `ToolHiveClientMetadataDocumentURL`."
2. Add a `// TODO(phase2): tighten per draft-ietf-oauth-client-id-metadata-document §3 (require host+path, reject fragment/userinfo) before wiring into the AS GetClient decorator.`

pkg/auth/discovery/discovery.go
```diff
@@ -539,6 +540,19 @@ func PerformOAuthFlow(ctx context.Context, issuer string, config *OAuthFlowConfi
 		return nil, fmt.Errorf("OAuth flow config cannot be nil")
 	}
 
+	// Before resolving ports or attempting DCR, check whether the AS advertises CIMD
+	// support. This handles issuer discovery paths (configured issuer, realm-derived)
+	// that return without fetching the AS discovery document, so the CIMD flag would
+	// otherwise never be seen.
+	if shouldDynamicallyRegisterClient(config) {
+		if doc, err := getDiscoveryDocument(ctx, issuer, config); err == nil &&
```
jhrozek: Two things on this block:

**(a) Discovery doc gets fetched twice on the AS-without-CIMD path.** When the AS doesn't advertise CIMD, this pre-check still calls `getDiscoveryDocument`, drops the result on the floor, and then `handleDynamicRegistration` calls `getDiscoveryDocument` again. The cache short-circuit at the top of `getDiscoveryDocument` only triggers when `RegistrationEndpoint`, `AuthorizeURL`, *and* `TokenURL` are pre-populated — none of which we set here.

Suggest populating those endpoints from `doc` whether or not CIMD is supported, so the second fetch short-circuits:

```go
if doc, err := getDiscoveryDocument(ctx, issuer, config); err == nil && doc != nil {
    if doc.RegistrationEndpoint != "" && config.RegistrationEndpoint == "" {
        config.RegistrationEndpoint = doc.RegistrationEndpoint
    }
    // …same for AuthorizeURL/TokenURL…
    if doc.ClientIDMetadataDocumentSupported {
        config.ClientID = oauthproto.ToolHiveClientMetadataDocumentURL
    }
}
```

**(b) The error is silently swallowed.** RFC-0071 explicitly mandates `WARN — CIMD fetch failed (URL, error); execution continues to DCR fallback`. Right now a transient discovery failure (DNS, 5xx, TLS hiccup) silently skips CIMD — the operator sees a DCR registration with no breadcrumb. At minimum:

```go
if err != nil {
    slog.Warn("CIMD pre-check: discovery fetch failed, falling back to DCR path", "issuer", issuer, "err", err)
}
```

Separately on the success line — the RFC asks for `INFO`, but `.claude/rules/go-style.md` says "silent success — no output at INFO or above for successful operations." I'd keep this at `Debug` and flag the conflict to the RFC author rather than land an INFO log per `thv run` startup. Worth a back-channel with @amirejaz.

pkg/auth/discovery/discovery.go
```diff
@@ -539,6 +540,19 @@ func PerformOAuthFlow(ctx context.Context, issuer string, config *OAuthFlowConfi
 		return nil, fmt.Errorf("OAuth flow config cannot be nil")
 	}
 
+	// Before resolving ports or attempting DCR, check whether the AS advertises CIMD
+	// support. This handles issuer discovery paths (configured issuer, realm-derived)
+	// that return without fetching the AS discovery document, so the CIMD flag would
+	// otherwise never be seen.
+	if shouldDynamicallyRegisterClient(config) {
```
jhrozek: Question on the CIMD path: RFC-0071 calls out twice that "all CIMD flows must use PKCE with `code_challenge_method=S256`" and lists "enforce PKCE S256 on CIMD flows" as a Phase 1 deliverable. I don't see anything PKCE-specific in this PR. Two possibilities:

1. The existing `PerformOAuthFlow` already enforces PKCE for *all* paths (e.g. via `oauth2.S256ChallengeOption`), and CIMD inherits it. In that case, can you point at the line where it happens, and consider adding a regression test that fails if PKCE is dropped from a CIMD authorize request?
2. PKCE is not universally enforced today, in which case CIMD is exactly the path that needs explicit S256 enforcement now.

Also: `hack/mock-cimd-server/main.go` doesn't validate `code_challenge` on `/oauth/authorize` or `code_verifier` on `/oauth/token` — making the mock strict on this would turn manual E2E into a useful regression guard.

pkg/auth/remote/config.go
```diff
@@ -71,6 +71,11 @@ type Config struct {
 	// RegistrationAccessToken is used to update/delete the client registration.
 	// Stored as a secret reference since it's sensitive.
 	CachedRegTokenRef string `json:"cached_reg_token_ref,omitempty" yaml:"cached_reg_token_ref,omitempty"`
+
+	// CachedCIMDClientID stores the CIMD URL used as client_id when CIMD was used
+	// for authentication. Non-sensitive — it is a public URL. Set to distinguish
+	// a CIMD-sourced client_id from a DCR-issued one across restarts.
+	CachedCIMDClientID string `json:"cached_cimd_client_id,omitempty" yaml:"cached_cimd_client_id,omitempty"`
```
jhrozek, line 80: Question — is the reader for `CachedCIMDClientID` deliberately deferred to a follow-up PR? Right now the field is written in `handler.go` `wrapWithPersistence` but no production code reads it (verified via grep — only `_test.go` and the field's own definition reference it). The doc comment says "set to distinguish a CIMD-sourced client_id from a DCR-issued one across restarts" but nothing currently consumes that distinction.

RFC-0071 says about this field: *"set when CIMD was used; **avoids re-detection on restart**."* That short-circuit isn't here yet, so the file persists a bit of state that no path consults.

Two follow-ups regardless of which path you pick:

1. `ClearCachedClientCredentials` doesn't clear `CachedCIMDClientID`. If the reader lands later, this becomes a stale-cache bug on credential rotation. Easy to add now alongside the existing clears.
2. The writer at `handler.go:218` uses `IsClientIDMetadataDocumentURL(result.ClientID)` to decide whether to cache. Per the doc-comment fix on `cimd.go`, that's fine for Phase 1 — but worth a short comment at the writer site explaining that any HTTPS-shaped client_id will be cached as CIMD, since DCR-issued IDs are always opaque.

Happy to land the reader + the missing clear in this PR if you'd like. Otherwise, would you mind dropping a follow-up issue link in the field's doc comment so future readers know it's intentional?

pkg/auth/remote/handler.go
```diff
@@ -131,15 +132,19 @@ func (h *Handler) performOAuthFlow(
 ) (oauth2.TokenSource, error) {
 	slog.Debug("Starting OAuth authentication flow", "issuer", issuer)
 
-	// Create OAuth flow config
 	flowConfig := h.buildOAuthFlowConfig(scopes, authServerInfo)
+	cimdUsed := flowConfig.ClientID == oauthproto.ToolHiveClientMetadataDocumentURL
 
 	result, err := discovery.PerformOAuthFlow(ctx, issuer, flowConfig)
+	if err != nil && cimdUsed {
```
jhrozek: Two issues here, one of them blocking — sorry to land on this one heavy.

**(1) The retry doesn't actually fall back to DCR.** Trace:

- `flowConfig.ClientID = ""` is set on retry.
- We re-call `discovery.PerformOAuthFlow`.
- Inside `PerformOAuthFlow` (discovery.go:547), `shouldDynamicallyRegisterClient(config)` returns true (ClientID empty), the pre-check fetches discovery, sees the AS still advertises `client_id_metadata_document_supported: true` (it does — it's the same AS), and re-applies `config.ClientID = ToolHiveClientMetadataDocumentURL`.
- `shouldDynamicallyRegisterClient` is now false, DCR is skipped, OAuth flow runs again with the same CIMD URL, and fails the same way.

Net effect: against an AS that advertises CIMD support but rejects the URL (mid-rollout, allowlists, policy gating), `thv run` hard-fails despite the code reading like it has a fallback.

Cleanest fix I can see: add `OAuthFlowConfig.SkipCIMD bool`, set it on retry, check it at the top of the discovery.go pre-check. Open to alternatives.

**(2) The trigger condition is broader than RFC-0071 specifies.** The RFC says:

> "The retry triggers on `invalid_client` or `unauthorized_client` OAuth error codes, or a raw HTTP 400/401 with no parseable RFC 6749 error body. `invalid_request` is not a trigger... The fallback does not trigger at the token exchange stage."

Current `if err != nil && cimdUsed` retries on every error class — `invalid_request`, network errors, transport failures, and token-exchange failures all silently downgrade to DCR. That hides real bugs.

Suggest using `errors.As` against `*oauth2.RetrieveError` (or whichever wrapped type carries the RFC 6749 `error` field) to inspect the code, and only retry on the RFC-allowed set. Bail on `invalid_request` and on any token-exchange-stage error.

Marking this as the only thing I'd consider blocking in the review — the rest is alignment and cleanup.

pkg/auth/remote/handler_test.go
```diff
@@ -797,6 +798,89 @@ func TestAuthenticate_BearerTokenPriority(t *testing.T) {
...
+
+			if tt.wantClientID != "" {
+				assert.Equal(t, tt.wantClientID, flowConfig.ClientID)
+			}
+			if tt.wantNotClientID != "" {
+				assert.NotEqual(t, tt.wantNotClientID, flowConfig.ClientID)
+			}
+		})
+	}
+}
```
jhrozek, line 882: RFC-0071's testing strategy enumerates these by name:

> "graceful fallback trigger on `invalid_client` and `unauthorized_client`, `invalid_request` surfaced as error not fallback, `CachedCIMDClientID` persistence."

What's here covers the construction matrix nicely, but the behavioral cases the RFC commits to are missing. Specifically would be good to add:

- AS responds `invalid_client` on authorize → DCR fallback → success.
- AS responds `unauthorized_client` on authorize → DCR fallback → success.
- AS responds `invalid_request` on authorize → error returned, **no** DCR fallback.
- Token-exchange-stage failure → error returned, **no** DCR fallback.
- Warm start: `CachedCIMDClientID` populated → discovery pre-check is skipped.

Existing tests in this file already use `httptest.Server` so the infrastructure is in place. The first batch would also have caught the retry-loop issue I flagged on `handler.go:137`.

hack/mock-cimd-server/main.go
```diff
@@ -0,0 +1,210 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// mock-cimd-server runs a minimal OAuth AS and MCP server for manual CIMD testing.
```
jhrozek: Two small follow-ups now that this is staying as part of the PR:

1. `go runAS(); go runMCPServer(); select{}` with `os.Exit(1)` from inside the goroutines — if one server fails to start, the other goroutine's logs may be lost before the process exits. An `errgroup.WithContext` plus signal-handled `srv.Shutdown` would be roughly the same line count, and Ctrl-C would shut down cleanly.

2. Tied to the PKCE question on `discovery.go` — if the authorize handler required `code_challenge_method=S256` and the token handler verified `code_verifier`, this mock would catch a PKCE regression on its own. Manual E2E currently accepts anything.

jhrozek (commented):
Mostly RFC-0071 alignment feedback. One concern I'd flag as a blocker (the DCR fallback retry doesn't actually fall back), several spec-conformance gaps around the trigger conditions, logging, and tests called out by name in the RFC. The rest is cleanup. Happy to pair on any of this.

jhrozek (commented):
Two follow-ups after the rework — both readability/robustness, neither blocking.

pkg/auth/remote/handler.go
```diff
@@ -131,15 +133,32 @@ func (h *Handler) performOAuthFlow(
...
-	// Create OAuth flow config
+	// Client registration priority (MCP spec: stored credentials → CIMD → DCR):
+	// Priority 1: Pre-configured credentials — set by buildOAuthFlowConfig from h.config.ClientID/ClientSecret.
+	// Priority 2: CIMD — AS advertises support and no credentials are set; use metadata URL as client_id.
+	// Priority 3: DCR — PerformOAuthFlow handles this when ClientID is still empty after the above.
 	flowConfig := h.buildOAuthFlowConfig(scopes, authServerInfo)
+	if authServerInfo != nil &&
+		authServerInfo.ClientIDMetadataDocumentSupported &&
+		flowConfig.ClientID == "" &&
+		flowConfig.ClientSecret == "" {
```
jhrozek: Now that the rework is in, this is the only multi-clause `if` in the new code, and it's doing three different things in one expression: a nil-guard, a capability check, and a precondition. Reader has to parse all four sub-conditions just to recognize the predicate "should we use CIMD?".

There's already a precedent for naming this kind of question — `pkg/auth/discovery/discovery.go:531` defines `shouldDynamicallyRegisterClient(config)` for exactly the same reason. Mirroring it at the handler level keeps the call site one line and gives the predicate a name:

```go
// shouldUseCIMD reports whether the CIMD client_id should be presented to the AS:
// the AS must advertise CIMD support and no pre-configured credentials may be set.
func shouldUseCIMD(authServerInfo *discovery.AuthServerInfo, flowConfig *discovery.OAuthFlowConfig) bool {
    if authServerInfo == nil || !authServerInfo.ClientIDMetadataDocumentSupported {
        return false
    }
    return flowConfig.ClientID == "" && flowConfig.ClientSecret == ""
}
```

Call site collapses to:

```go
flowConfig := h.buildOAuthFlowConfig(scopes, authServerInfo)
if shouldUseCIMD(authServerInfo, flowConfig) {
    flowConfig.ClientID = oauthproto.ToolHiveClientMetadataDocumentURL
    slog.Debug("Using CIMD client_id", "url", oauthproto.ToolHiveClientMetadataDocumentURL)
}
```

The sibling check at line 152 (`flowConfig.ClientID == oauthproto.ToolHiveClientMetadataDocumentURL && isCIMDRejectionError(err)`) reads cleanly enough as-is — two clauses, both named — so I'd leave it alone.

Non-blocking; pure readability.

pkg/auth/remote/handler.go
```diff
@@ -457,3 +488,15 @@ func (h *Handler) tryDiscoverFromWellKnown(
...
+// isCIMDRejectionError returns true if err indicates the AS rejected the CIMD
+// client_id. Only the RFC 6749 error codes invalid_client and unauthorized_client
+// trigger a DCR retry; all other errors — including invalid_request and
+// token-exchange failures — surface as-is.
+func isCIMDRejectionError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "invalid_client") || strings.Contains(msg, "unauthorized_client")
```
jhrozek: Non-blocking follow-up to the original retry comment.

This works in practice today — `golang.org/x/oauth2` does include the RFC 6749 `error` field as a substring in the formatted error message — but `strings.Contains` on an unstructured error message has two soft risks:

1. **Substring collision.** Any error whose message happens to contain the literal `"invalid_client"` would trigger a fallback. RFC 6749 itself only defines `invalid_client`, but downstream systems sometimes synthesize codes like `invalid_client_metadata` (RFC 7591 §3.2.2) — if such a string appears anywhere in the error chain, we'd silently downgrade to DCR when we shouldn't. Theoretical right now, but the cost of guarding against it is one line.
2. **Format coupling.** If `golang.org/x/oauth2` ever changes how it serializes the response body into the error message (e.g. switches to a structured representation, escapes the JSON differently), this predicate silently stops triggering and the retry path goes dark — and the `TestIsCIMDRejectionError` cases keep passing because they synthesize errors of the same shape.

Safer pattern is to inspect the structured error directly:

```go
func isCIMDRejectionError(err error) bool {
    if err == nil {
        return false
    }
    var rerr *oauth2.RetrieveError
    if !errors.As(err, &rerr) {
        return false
    }
    switch rerr.ErrorCode {
    case "invalid_client", "unauthorized_client":
        return true
    }
    return false
}
```

This also positions you to handle the RFC-0071 "raw HTTP 400/401 with no parseable error body" case cleanly later (just inspect `rerr.Response.StatusCode` when `rerr.ErrorCode == ""`).

If the structured form turns out not to be reachable from the call site (some wrapping in between), happy to look at that with you.

pkg/auth/remote/config.go
jhrozek: Thanks for the cleared-on-rotation addition and the issue #2728 reference in the doc — that's a clean trade-off. Resolving.

pkg/auth/remote/handler.go
jhrozek: Retry mechanism is structurally fixed by the rework — confirmed by tracing the new path (CIMD pre-check is gone from `PerformOAuthFlow`, so the retry's `flowConfig.ClientID = ""` actually leads to DCR). Trigger condition is also narrowed correctly to `invalid_client`/`unauthorized_client`.

Posted a separate non-blocking comment about `isCIMDRejectionError` using `strings.Contains` rather than `errors.As(*oauth2.RetrieveError)` — that's a robustness follow-up, not a regression. Resolving this thread.

pkg/auth/remote/handler_test.go
jhrozek: `TestIsCIMDRejectionError` plus the new E2E suite cover the predicate and both happy paths (CIMD-supported and DCR-when-not-advertised) nicely. The PKCE `code_challenge` assertion in the E2E is a particularly nice addition.

Two cases I'd flag for a follow-up issue rather than this PR:

- E2E for AS-advertises-CIMD-then-rejects-with-`invalid_client` → DCR fallback succeeds. The unit test covers the predicate; nothing exercises the full retry against a mock AS.
- Warm-start cached CIMD test (tied to the deferred reader in #2728).

Resolving as adequately addressed for the scope of this PR.

pkg/auth/discovery/discovery.go
amirejaz: PKCE is already enforced unconditionally for all flows — including CIMD. In `pkg/auth/discovery/discovery.go`, `createOAuthConfig` passes `true` for PKCE at both the manual-endpoints path (line 659) and the OIDC-discovery path (line 675). CIMD flows inherit this without any additional change needed. The PKCE `code_challenge` assertion in the E2E test (`Expect(authReq.CodeChallenge).ToNot(BeEmpty())`) provides the regression guard.

hack/mock-cimd-server/main.go
amirejaz: The `hack/mock-cimd-server` binary has been removed from the PR entirely — it was added for a manual test session and had no E2E backing. CIMD testing is now covered by the `cimd_auth_test.go` E2E suite using a proper `httptest`-based mock AS. Both points are moot.

pkg/auth/remote/handler_test.go
amirejaz: The E2E suite in `test/e2e/cimd_auth_test.go` now covers both happy paths (CIMD-supported and DCR-when-not-advertised) with the `code_challenge` assertion for PKCE. The two follow-up cases you flagged — AS-advertises-CIMD-then-rejects and warm-start — will be tracked as follow-up issues against #2728 and the Phase 2 work.

hack/mock-cimd-server/main.go
```diff
@@ -0,0 +1,204 @@
...
+		callback, _ := url.Parse(redirectURI)
+		q := callback.Query()
+		q.Set("code", code)
+		q.Set("state", state)
+		callback.RawQuery = q.Encode()
+
+		// Deliver the code directly to thv's callback server (skip browser)
+		go func() {
+			time.Sleep(300 * time.Millisecond)
+			resp, err := http.Get(callback.String()) //nolint:noctx
```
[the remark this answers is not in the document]
amirejaz: The `hack/mock-cimd-server` binary has been removed from this PR entirely — it was added only for a manual test session. The CodeQL alert is no longer applicable.

hack/mock-cimd-server/main.go
```diff
@@ -0,0 +1,210 @@
...
+		q.Set("code", code)
+		q.Set("state", state)
+		callback.RawQuery = q.Encode()
+		callbackURL := callback.String()
+
+		// Deliver the code directly to thv's callback server (skip browser).
+		// Safe: redirect_uri is validated to be localhost above.
+		go func() {
+			time.Sleep(300 * time.Millisecond)
+			resp, err := http.Get(callbackURL) //nolint:noctx,gosec // G107: URL validated to localhost above
```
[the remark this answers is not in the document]
amirejaz: Same as above — the file has been removed from the PR.

pkg/oauthproto/cimd.go
amirejaz: Fixed — collapsed into a single paragraph.

pkg/oauthproto/cimd.go
amirejaz: Done. Expanded the doc comment to make the intentional heuristic explicit and added the Phase 2 TODO pointing at draft-ietf-oauth-client-id-metadata-document §3.

pkg/auth/discovery/discovery.go
amirejaz: Both addressed: (a) Priority 1 and Priority 2 in `discoverIssuerAndScopes` now call `ValidateAndDiscoverAuthServer` so the discovered endpoints are populated on `authServerInfo` before reaching `buildOAuthFlowConfig` — `handleDynamicRegistration` short-circuits on the second call because `RegistrationEndpoint`/`AuthorizeURL`/`TokenURL` are already set. (b) Added the WARN log when AS discovery fetch fails on the Priority 1/2 paths.

pkg/auth/remote/config.go
amirejaz: The reader is now implemented in `resolveClientCredentials` — it checks `HasCachedCIMDClientID()` first and returns the CIMD URL as `client_id` for token refresh on restart, before falling through to DCR credentials. `ClearCachedClientCredentials` deliberately does not clear `CachedCIMDClientID` (documented in its comment) since the CIMD URL is a stable constant that does not need rotating alongside DCR secrets.

pkg/auth/remote/handler.go
amirejaz: Both fixed. (1) The CIMD pre-check was moved out of `PerformOAuthFlow` entirely — it now lives in `performOAuthFlow` in handler.go. On retry, `flowConfig.ClientID = ""` and `PerformOAuthFlow` goes straight to DCR with no re-application of the CIMD URL. (2) `isCIMDRejectionError` now uses `errors.As(*oauth2.RetrieveError)` for token-endpoint errors and `strings.HasPrefix` for auth-endpoint callback errors (format: `"OAuth error: <code> - <desc>"` from flow.go), covering both stages precisely.

pkg/auth/remote/handler.go
amirejaz: Done. Extracted into `shouldUseCIMD(authServerInfo, flowConfig)` mirroring the `shouldDynamicallyRegisterClient` pattern.

pkg/auth/remote/handler.go
amirejaz: Implemented — and discovered that the structured check alone was not sufficient. CIMD rejection from the authorization endpoint arrives as a plain error from flow.go's `handleCallback` (format: `"OAuth error: \<code\> - \<desc\>"`), not as `*oauth2.RetrieveError`. The function now checks `errors.As` first for token-endpoint errors, then falls back to `strings.HasPrefix("OAuth error: invalid_client")` for auth-endpoint errors, covering both stages without the substring-collision risk you flagged.

pkg/auth/remote/handler.go
jhrozek: Looks great — `shouldUseCIMD` extracted exactly as suggested at `handler.go:504`, mirroring `shouldDynamicallyRegisterClient`, and the call site at line 142 collapses to one clear line. Resolving.

pkg/auth/remote/handler.go
jhrozek: Nice catch on the auth-endpoint case in commit e8f6c2282 — my pure-`errors.As` rewrite would have broken the authorize-stage retry entirely, since `flow.go:314` formats those callback errors as plain `fmt.Errorf("OAuth error: %s - %s", ...)` rather than `*oauth2.RetrieveError`. The hybrid (structured `errors.As` for the token endpoint, `strings.HasPrefix` for the auth callback) is the right answer; the doc comment at `handler.go:516-519` explaining the two stages is also a nice touch.

Test coverage now covers both stages (`auth callback invalid_client/unauthorized_client/invalid_request/access_denied`). Resolving.

jhrozek (approved):
LGTM — thanks for the careful iteration on the rejection-error handling. Good catch that the auth-callback path needed the HasPrefix branch alongside errors.As; the structured-first, string-fallback hybrid is the right shape.

Issue #5066, opened 2026-04-27
pkg/auth/oauth: factor shared non-caching refresher and lift duplicated helpers


yrobla:
## Status: Absorbed into #5050

All scope from this issue has been addressed as part of #5050 (Extract shared OAuthTokenSource from pkg/registry/auth and pkg/llm).

## Original context

Flagged in #5052 (review by @JAORMX). `pkg/llm/tokensource.go` introduced `nonCachingRefresher`, which does the same `cfg.Exchange(ctx, "", opts...)` dance as `pkg/auth/oauth/resourceTokenSource` (resource_token_source.go:88). The only meaningful difference is that `resourceTokenSource.Token()` short-circuits on `r.token.Valid()` (line 45), which is exactly what `withPreemptiveRefresh` needs to not happen.

## What was done (in #5050)

**Scope 1 — Non-caching refresher:** `NonCachingRefresher` was already extracted to `pkg/auth/oauth` in #5052. `resourceTokenSource` delegates to it. `pkg/llm.nonCachingRefresher` was removed as part of the #5050 extraction — the shared type is used via `pkg/auth/tokensource`.

**Scope 2 — Lifted helpers:**
- `ensureOfflineAccess` → `tokensource.EnsureOfflineAccess` in `pkg/auth/tokensource`
- `DeriveSecretKey` → `tokensource.DeriveSecretKey(prefix, url, issuer)` (parameterized prefix) in `pkg/auth/tokensource`

Both helpers are now shared across `pkg/llm` and `pkg/registry/auth`; the duplicates have been deleted.

yrobla:
Closing in favour of #5050 which covers the same ground as a superset. The `NonCachingRefresher` extraction and the `ensureOfflineAccess`/`DeriveSecretKey` helper lifts are part of the shared `OAuthTokenSource` refactor scoped there.

PR #5090, opened 2026-04-28
Extract shared OAuthTokenSource into pkg/auth/tokensource

### LINKED ISSUE #5050: Extract shared OAuthTokenSource from pkg/registry/auth and pkg/llm
## Background

`pkg/llm/tokensource.go` (introduced in the `thv-llm-token` branch) and `pkg/registry/auth/oauth_token_source.go` share substantial duplication: `ensureOfflineAccess` is byte-for-byte identical; `DeriveSecretKey` differs only in prefix; `Token`, `tryRestoreFromCache`, `performBrowserFlow`, `buildFlowConfig`, `buildOAuth2Config`, `makeTokenPersister`, and `refreshTokenKey` are near-identical in structure.

Flagged in PR review by @aponcedeleonch.

Also absorbs #5066 (factor shared non-caching refresher and lift duplicated helpers).

## Goal

Extract a shared `OAuthTokenSource` into `pkg/auth/tokensource/` parameterized by:

- An OIDC config getter (issuer, clientID, scopes, audience, callbackPort)
- A secret-key derivation function (so registry gets `REGISTRY_OAUTH_` prefix, LLM gets `LLM_OAUTH_`)
- A `ConfigPersister` callback (replaces both the hardcoded `config.UpdateConfig` in the registry side and the injected `TokenRefUpdater` in the LLM side)

The LLM implementation is the better starting point — it has:
- Principled `lastErr` propagation (vs silent error swallow in registry)
- `errCacheMiss` sentinel + `IsNotFoundError` classification
- `PersistingTokenSource` wrapping on `tryRestoreFromCache` (registry side is missing this — latent bug: rotated refresh tokens aren't re-persisted on cache restore)
- Injected callback for config persistence (more testable than direct `config.UpdateConfig`)

The genuinely new capabilities — preemptive refresh window and access-token cross-process cache — should apply to both consumers after extraction.

### Shared `OAuthTokenSource` (pkg/auth/tokensource)

- Extract into `pkg/auth/tokensource` parameterized as above
- Both `pkg/llm/tokensource.go` and `pkg/registry/auth/oauth_token_source.go` become thin wrappers
- Registry `tryRestoreFromCache` wraps with `PersistingTokenSource` (parity with LLM fix)

### Lifted helpers

- `ensureOfflineAccess` → `tokensource.EnsureOfflineAccess` (was byte-identical in both packages)
- `DeriveSecretKey` → `tokensource.DeriveSecretKey(prefix, url, issuer)` (parameterized prefix)

### Non-caching refresher (from #5066)

- `NonCachingRefresher` already extracted into `pkg/auth/oauth` (landed in #5052)
- `resourceTokenSource` already delegates to it
- Duplication in `pkg/llm.nonCachingRefresher` removed as part of this extraction

## Acceptance criteria

- [x] Shared `OAuthTokenSource` in `pkg/auth/tokensource` parameterized as above
- [x] `pkg/llm/tokensource.go` uses the shared type
- [x] `pkg/registry/auth/oauth_token_source.go` migrated to shared type (deleted)
- [x] Registry `tryRestoreFromCache` wraps with `PersistingTokenSource` (parity with LLM fix)
- [x] `EnsureOfflineAccess` and `DeriveSecretKey` lifted to shared package (closes #5066 scope 2)
- [x] `pkg/llm.nonCachingRefresher` removed — uses `oauth.NonCachingRefresher` (closes #5066 scope 1)
- [x] No behaviour change in either consumer (intentional improvement: backend errors now surface specifically in registry path)

yrobla:
## Summary

Both pkg/llm and pkg/registry/auth implemented the same four-tier OAuth token strategy (in-memory cache → secrets-cached access token → refresh token → browser OIDC/PKCE). This extracts that logic into a single shared package, pkg/auth/tokensource, and rewires both consumers as thin wrappers.

**Behavior changes (intentional):**

1. **Registry path now has cross-process access-token caching.** The shared `OAuthTokenSource` writes a `<key>_AT` secret after each successful token fetch (expiry encoded in the value). Concurrent short-lived `thv` processes can reuse this cached AT instead of all racing to exchange the same refresh token — preventing `invalid_grant` errors from OIDC providers that rotate refresh tokens on each exchange. This is new for the registry path (the old `oauth_token_source.go` did not have this tier).

2. **Registry path now wraps its restored token source with `PersistingTokenSource`.** Rotated refresh tokens are re-persisted after a cache restore, matching the LLM path and fixing a gap where a provider-rotated refresh token would be lost on the next invocation.

Fixes #5050
Incorporates #5066

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Other (describe): Refactoring with intentional behavior improvements — see above

## Test plan

## API Compatibility

## Changes

| File | Change |
|------|--------|
| `pkg/auth/tokensource/tokensource.go` | New shared package: four-tier `OAuthTokenSource`, `DeriveSecretKey`, `EnsureOfflineAccess` |
| `pkg/auth/tokensource/preemptive.go` | Preemptive refresh chain (`withPreemptiveRefresh`, `withPreemptiveRefreshFrom`) |
| `pkg/auth/oauth/noncaching.go` | `NonCachingRefresher` — always performs a network round-trip, prevents preemptive-window thrashing |
| `pkg/llm/tokensource.go` | Reduced to type aliases + thin constructor over shared package |
| `pkg/registry/auth/auth.go` | Reduced to thin constructor over shared package; `updateRegistryTokenRef` simplified |
| `pkg/registry/auth/oauth_token_source.go` | Deleted — superseded by shared package |

The registry path gains cross-process AT caching (extra keyring/secrets writes). Users with very restrictive keyring write policies may need to allow writes to `<key>_AT` keys. In practice this is transparent and improves reliability for concurrent `thv` invocations.

- The `_AT` cache key is `KeyProvider() + "_AT"`. For registry, `KeyProvider` returns either `CachedRefreshTokenRef` or `REGISTRY_OAUTH_<8hex>`, so the AT key is e.g. `REGISTRY_OAUTH_ab12cd34_AT`.
- `cacheAccessToken` suppresses write errors (debug log only) — a keyring write failure degrades gracefully to a full refresh on the next call.
- `makeTokenPersister` writes `""` to `key+"_AT"` to invalidate the AT cache after a refresh token rotation (rather than `DeleteSecret`, for provider compatibility).

## Large PR Justification

This is a refactor to unify OAuthTokenSource used in several places. It reduces code duplication and provides better integration with cross-process token caching as a side effect.


pkg/auth/tokensource/tokensource.go
```diff
@@ -0,0 +1,466 @@
...
+// expired, triggering a proactive refresh before the upstream rejects it.
+const preemptiveRefreshWindow = 30 * time.Second
+
+// OIDCParams holds the OIDC connection parameters for a token source.
+type OIDCParams struct {
+	Issuer       string
+	ClientID     string
+	Scopes       []string // "offline_access" is appended automatically if absent
+	Audience     string
+	CallbackPort int
```
aponcedeleonch, line 53: Suggest pushing the default into this package so both LLM and registry callers get consistent behavior. Right now `pkg/registry/auth/auth.go:45-48` applies `remote.DefaultCallbackPort` when zero, but the LLM path doesn't. Defaulting here removes the asymmetry, callers can still validate the configured value at config-load time and treat `0` as 'use default'.

pkg/registry/auth/auth_test.go
```diff
@@ -197,396 +146,242 @@ func TestNewTokenSource(t *testing.T) {
...
-				oauthCfg: &config.RegistryOAuthConfig{
-					Issuer:   "https://auth.example.com",
-					ClientID: "test-client",
-				},
-				registryURL:     "https://registry.example.com",
-				secretsProvider: mockProvider,
-			}
+// newTokenServer builds a server that handles OIDC discovery AND a token
+// endpoint that returns the given access token and refresh token.
+func newTokenServer(t *testing.T, at, rt string) *httptest.Server {
```
aponcedeleonch: `newTokenServer` here and `newOIDCTestServer` in `helpers_test.go` are nearly-identical OIDC discovery handlers in two test packages (this file is `package auth_test`, helpers_test.go is `package auth`). Consider consolidating into a single helper that takes an optional `/token` payload, either move both into helpers_test.go and dot-import here, or revert this file to `package auth`.

pkg/auth/tokensource/tokensource_test.go
```diff
@@ -0,0 +1,601 @@
...
+	ts := tokensource.New(opts)
+
+	_, err := ts.Token(context.Background())
+	require.Error(t, err)
+	// OIDC failure must propagate as lastErr, not the generic FallbackErr.
+	assert.False(t, errors.Is(err, errTestFallback), "OIDC discovery failure must not collapse to FallbackErr")
+}
+
+// Ensure the package-level oauth helpers exist and are importable.
+var _ = oauth.NewFlow
```
aponcedeleonch: This compile-time assertion is redundant, the `oauth` package is already used elsewhere in this file via `oauth.NewNonCachingRefresher`. Safe to drop.

aponcedeleonch (commented):
Three small follow-ups inline. Nice refactor overall.

aponcedeleonch (approved):
Thanks for the clean refactor. All three of my comments are addressed. The regression surface looks tight: LLM is a pure extraction, registry's behavior changes are intentional and well-tested. LGTM.

PR #5092, opened 2026-04-28
Fix non-expiring upstream token handling and storage TTL bugs

### LINKED ISSUE #5046: Non-expiring upstream tokens incorrectly marked as 1h-expiring
## Summary

Non-expiring upstream tokens (Slack user tokens, GitHub user tokens) are incorrectly marked as expiring in 1 hour, causing the refresh path to fire unnecessarily and — when the provider returns the same non-expiring token — creating churn or failures on subsequent lookups.

## Root cause

`pkg/authserver/upstream/oauth2.go:224-228` fabricates a 1-hour expiry whenever the upstream token response has no `expires_in`:

```go
expiresAt := token.Expiry
if expiresAt.IsZero() {
    // Default to 1 hour if not specified
    expiresAt = time.Now().Add(defaultTokenExpiration)
}
```

`convertOAuth2Token` is called from both the initial code exchange (`ExchangeCode`, line ~455) and subsequent refreshes (`RefreshTokens`, line ~495), so the fabricated expiry taints every leg.

Downstream code at `pkg/auth/upstreamtoken/service.go:104-108` explicitly special-cases a zero `ExpiresAt` as "never expires, don't refresh":

```go
if tokens.ExpiresAt.IsZero() || !tokens.IsExpired(time.Now()) {
    result[providerName] = tokens.AccessToken
    continue
}
```

This correct branch is defeated by the fabricated 1h value.

## Affected providers

Any upstream that does not return `expires_in`:
- Slack user tokens (non-rotating, do not expire)
- GitHub user access tokens without expiration
- Any OAuth 2.0 provider configured to issue long-lived tokens

## Proposed fix

Two changes are required — removing only the `oauth2.go` fabrication is insufficient because the storage layer also applies a default TTL.

1. **`pkg/authserver/upstream/oauth2.go:224-228`** — delete the `if expiresAt.IsZero()` block and let `ExpiresAt` remain zero. Also remove the unused `defaultTokenExpiration` constant at line 93-94.

2. **Storage layer default TTL** — `pkg/authserver/storage/memory.go:705-712` and `pkg/authserver/storage/redis.go:888-896` currently apply a default TTL to upstream token records when `ExpiresAt` is zero. For records explicitly marked as non-expiring, they should be stored with no TTL (memory) or a configured long TTL (Redis). Alternatively, carry a separate `NeverExpires bool` field on `UpstreamTokens` to disambiguate.

## Related

Compounds with #5047 to produce the observed symptom of "all upstream providers require re-authorization every ~24 hours." This issue (fake expiry) triggers the re-authorization cascade; the linked issue (session-keyed storage + callback token clobber) ensures the cascade is fatal.

## Reproduction

1. Configure ToolHive with a Slack OAuth upstream.
2. Complete initial authorization. Observe Slack token is stored.
3. Wait 1 hour.
4. Trigger a downstream request that reads the Slack token.
5. Observe the refresher is invoked even though Slack issued a non-expiring token.

### LINKED ISSUE #5047: Upstream refresh tokens orphaned on re-authorization
## Summary

Upstream refresh tokens are orphaned when a user re-authorizes. Two independent mechanisms cause this: (1) storage is keyed by an ephemeral session ID that is freshly generated on every `/authorize` call, and (2) the callback stores the new IDP response verbatim, overwriting a previously-stored refresh token with the empty string when the provider does not re-issue one.

The second mechanism is most visible with providers that only return `refresh_token` on first consent or when `prompt=consent` is explicitly passed (Google OAuth 2.0 is a prominent example). After any re-authorization, the stored refresh token is effectively lost, and subsequent refresh attempts fail with `ErrNoRefreshToken`.

## Root cause A — session-keyed storage

`pkg/authserver/server/handlers/authorize.go:91` generates a fresh session ID on every `/authorize`:

```go
pending := &storage.PendingAuthorization{
    ...
    SessionID: rand.Text(),   // ← fresh random on every /authorize
    CreatedAt: time.Now(),
}
```

`pkg/authserver/server/handlers/callback.go:142` stores upstream tokens keyed by that session. The session ID is then embedded as the `tsid` claim in the issued JWT (`pkg/authserver/server/session/session.go:120-123`), and downstream read paths look up by `tsid`/session ID (`pkg/auth/token.go:1189-1196`, `pkg/authserver/storage/types.go:222-240`).

`UserID` (the internal ToolHive user subject) is stored as a field on `UpstreamTokens` (`callback.go:138`) but no read path uses it as a lookup key. It is only used for cleanup/metadata.

**Consequence:** any re-authorization mints a new session; the old session's tokens become unreachable through the normal read path.

## Root cause B — callback clobbers prior refresh token

`pkg/authserver/server/handlers/callback.go:131-142` stores `idpTokens.RefreshToken` verbatim, with no preservation of a previously-stored refresh token:

```go
storageTokens := &storage.UpstreamTokens{
    ProviderID:   providerID,
    AccessToken:  idpTokens.AccessToken,
    RefreshToken: idpTokens.RefreshToken,   // empty if provider didn't re-issue
    ...
}
if err := h.storage.StoreUpstreamTokens(ctx, sessionID, providerID, storageTokens); err != nil {
```

`pkg/authserver/refresher.go:73-75` already has the correct pattern for the refresh path (preserve prior refresh token when provider omits a new one). The callback path does not mirror this.

Even if root cause A is fixed (e.g., switching to `UserID`-keyed storage), root cause B will still overwrite a good refresh token with the empty string on every re-auth where the provider does not re-issue a refresh token.

`pkg/auth/upstreamtoken/service.go:135-136` then returns `ErrNoRefreshToken`:

```go
if expired.RefreshToken == "" {
    return nil, ErrNoRefreshToken
}
```

## Failure mode

1. Initial consent completes. Upstream tokens, including a `refresh_token`, are stored under session `S1` with key `(S1, providerID)`.
2. Some time later, a new `/authorize` call is made (any reason — including the compound effect of #5046 causing spurious refresh/reauth cycles).
3. Authorize handler mints a new session `S2`. The new JWT's `tsid` claim points at `S2`.
4. Callback completes. If the provider does not re-issue `refresh_token` (e.g., Google without `prompt=consent` — `pkg/authserver/upstream/oidc.go:367-369` shows `prompt=consent` is optional, not default), the callback stores `{AccessToken: new, RefreshToken: ""}` at `(S2, providerID)`.
5. When the new access token expires, the refresher reads from `(S2, providerID)`, finds `RefreshToken == ""`, and returns `ErrNoRefreshToken`. The provider is then omitted from downstream responses with the log line `omitting provider with unrefreshable expired token` (`pkg/auth/upstreamtoken/service.go:114`).
6. The user is forced to re-authorize. The cycle repeats.

## Proposed fix

Two independent changes — neither alone is sufficient.

1. **Key upstream tokens by `UserID`, not session ID.**
   - Add a `(user_id, provider_id) → tokens` index in both memory and Redis storage backends.
   - Update the downstream read path (`pkg/auth/token.go:1189-1196` and friends) to resolve `tsid → user_id` via the session store, then read upstream tokens by `user_id`.
   - This is net-new index work; `UserID` is currently only metadata.

2. **Preserve prior refresh token in the callback.** Before calling `StoreUpstreamTokens`, load any existing record for `(user_id, provider_id)` and, if `idpTokens.RefreshToken == ""`, carry forward the existing `RefreshToken`. Mirror the pattern already used at `pkg/authserver/refresher.go:73-75`.

## Related

Compounds with #5046. The fake 1h expiry described there triggers the re-authorization cascade that exposes this bug.

## Reproduction

Assumes a Google OAuth client (OIDC) is configured as a ToolHive upstream with the default config (no `prompt=consent` set).

1. Start ToolHive with Google configured as an upstream OIDC provider.
2. As a user, complete the initial `/authorize` flow. Accept Google's consent screen.
3. Verify the stored upstream tokens include a non-empty `RefreshToken`. For the in-memory backend this can be inspected via the authserver's storage; for Redis, read the `upstream_tokens:<session_id>:google` key.
4. Trigger a second `/authorize` for the same Google account, without the initial session being invalidated. This can be done by:
   - Calling the OAuth endpoint again from a fresh client registration, or
   - Signing out of the ToolHive client and signing back in, or
   - Any path that causes the client to restart the authorization flow.
5. Complete the flow. Google returns a new access token but **no** `refresh_token` (this is Google's documented behavior without `prompt=consent`).
6. Inspect the stored tokens under the new session's key: `RefreshToken` is now `""`. The previous session's tokens (with the good refresh token) still exist under the old session key but are unreachable via the new JWT's `tsid`.
7. Wait for the new access token to expire (~1 hour with Google's default).
8. Trigger a downstream request that causes `InProcessService.GetAccessTokens` to run. Observe the log line:

   ```
   omitting provider with unrefreshable expired token provider=google
   ```

9. The Google upstream is now omitted from downstream responses until the user re-authorizes — which will reproduce the same failure.

## Caveats

The claim "Google does not re-issue refresh tokens without `prompt=consent`" reflects Google's documented OAuth 2.0 behavior. The repo only shows `prompt=consent` is optional (`pkg/authserver/upstream/oidc.go:367-369`).

jhrozek:
## Summary

Users authenticated with non-expiring upstream providers (GitHub OAuth Apps, Slack user tokens, Vault, on-prem providers without `expires_in`) hit a forced re-authentication every 60 minutes. The auth server fabricated a 1h `ExpiresAt` for any token that lacked `expires_in`, then treated those tokens as expired, attempted to refresh them with refresh tokens that didn't exist, hit `ErrNoRefreshToken`, and dropped the provider from downstream responses with `omitting provider with unrefreshable expired token`.

A second silent failure compounded the first: in multi-upstream chains that combined a non-expiring provider with an expiring one (e.g., GitHub PAT + Google), the Redis session-index set could be evicted while the non-expiring per-provider key was still alive — `GetAllUpstreamTokens` would silently drop the non-expiring provider exactly an hour after the session started.

This PR:

- Removes the fabricated 1h `ExpiresAt` in `convertOAuth2Token`. Tokens without `expires_in` flow through with zero `ExpiresAt`.
- Fixes `IsExpired`/`IsExpiredAt` to treat zero `ExpiresAt` as non-expiring (was: returned true → trapped tokens in the cascade).
- Stops storage backends from evicting zero-`ExpiresAt` tokens.
- Adds `SessionExpiresAt` to `UpstreamTokens` as a Fosite-session-bound storage TTL fallback so non-expiring tokens have a finite storage lifetime even when the upstream asserts none. Set unconditionally on every callback so the refresh path is safe even if a provider stops asserting `expires_in` on a later rotation.
- Fixes a Lua TTL inversion in `storeUpstreamTokensScript` that re-applied a TTL to a `PERSIST`'d session index, silently orphaning non-expiring tokens after a subsequent expiring write to the same session.
- Guards the `expires_at` log fields against zero-time formatting (was: emitted `0001-01-01T00:00:00Z`). Now uses a `slog.LogValuer` that lazy-renders to `"none"` for zero time, RFC 3339 otherwise.

Closes #5046

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other

## Test plan

- [x] Race detector (`go test -race ./pkg/authserver/...`) — clean
- [ ] E2E tests (`task test-e2e`) — not run; no change at the CLI/operator boundary

New regression coverage:

- `TestRedisStorage_UpstreamTokens/mixed-expiry_session:_expiring_write_after_non-expiring_keeps_index_persistent` — the Lua bug
- `TestRedisStorage_UpstreamTokens/fresh_expiring_write_applies_index_TTL` — guards the discrimination between fresh-set and PERSIST'd-set
- `TestRedisStorage_UpstreamTokens/longer_expiring_write_after_shorter_extends_index_TTL` and `/shorter_expiring_write_after_longer_leaves_index_TTL_unchanged` — lock down the comparator
- `TestUpstreamTokenRefresher_RefreshAndStore/preserves_SessionExpiresAt_when_provider_omits_expires_in` — refresh-path bound
- `TestIntegration_UpstreamTokenService_NonExpiringToken` — end-to-end regression for the original cascade

## API Compatibility

Yes. Suggested release note:

> Fixes a regression where MCP sessions using non-expiring upstream providers (GitHub OAuth Apps, Slack user tokens, Vault) were forced into hourly re-authentication. Also fixes a silent partial-token-loss in multi-upstream chains where the session index could be evicted before its non-expiring members.

- **Scope**: this branch fixes #5046 and a related index-eviction bug. It does **not** fix #5047 (re-authorization orphans Google refresh tokens). #5047 is a structural bug requiring a storage rekey from `(SessionID, providerName)` to `(UserID, providerName)`, which is being designed in a separate RFC and will land as a follow-on PR sequence. This branch is a strict improvement on its own and is forward-compatible with the rekey.
- **The Lua change** in `pkg/authserver/storage/redis.go` (commit `fec89b040`) is the most subtle. The control flow is structured as a decision keyed on `(ttlMs, idxExisted, idxTTL)`. The trick is reading `EXISTS` *before* `SADD` — after `SADD` the set always exists, so a fresh-set and a previously-`PERSIST`'d set are indistinguishable from `PTTL` alone. Inline comments flag the bug-prone branch (`idxTTL == -1` after pre-existing set).
- **Commit-by-commit review** is intentional: each commit is self-contained and the messages explain motivation. The history was rebased to fold MoE-review fixups into their parents.

## Large PR Justification

- there's a lot of tests, mostly

### DIFF
--- a/pkg/authserver/integration_test.go
+++ b/pkg/authserver/integration_test.go
@@ -19,10 +19,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/alicebob/miniredis/v2"
 	"github.com/go-jose/go-jose/v4"
 	"github.com/go-jose/go-jose/v4/jwt"
 	"github.com/oauth2-proxy/mockoidc"
 	"github.com/ory/fosite"
+	"github.com/redis/go-redis/v9"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -46,11 +48,29 @@ const (
 )
 
 // testServer bundles all test server components together.
+//
+// The mr field is non-nil only when the test server was constructed
+// with withRedisBackedStorage(); otherwise it is nil and the in-memory
+// backend is used. Tests that need to advance Redis time call Miniredis(t)
+// rather than dereferencing this field directly so a misconfigured test
+// fails loudly instead of nil-panicking.
 type testServer struct {
 	Server     *httptest.Server
 	PrivateKey *rsa.PrivateKey
 	authServer Server
 	storage    storage.UpstreamTokenStorage
+	mr         *miniredis.Miniredis
+}
+
+// Miniredis returns the miniredis instance backing this test server. Fails
+// the test loudly with a useful message if the server was not constructed
+// with withRedisBackedStorage(). Use this rather than dereferencing the
+// field directly.
+func (ts *testServer) Miniredis(t *testing.T) *miniredis.Miniredis {
+	t.Helper()
+	require.NotNil(t, ts.mr,
+		"testServer was not constructed with withRedisBackedStorage(); call setupTestServer*(..., withRedisBackedStorage()) to enable miniredis access")
+	return ts.mr
 }
 
 // testServerOptions configures the test server setup.
@@ -58,6 +78,13 @@ type testServerOptions struct {
 	upstream            upstream.OAuth2Provider
 	scopes              []string
 	accessTokenLifespan time.Duration
+	// storageFactory, when non-nil, supplies the storage backend instead of
+	// the default in-memory implementation. The factory may also return a
+	// *miniredis.Miniredis instance; when it does, the setup helper stashes
+	// it on testServer.mr (accessed via testServer.Miniredis()) so tests can drive its clock. A nil
+	// miniredis return value is valid (e.g., for non-Redis alternative
+	// backends).
+	storageFactory func(t *testing.T) (storage.Storage, *miniredis.Miniredis)
 }
 
 // testServerOption is a functional option for test server setup.
@@ -81,6 +108,31 @@ func withScopes(scopes []string) testServerOption {
 func withAccessTokenLifespan(d time.Duration) testServerOption {
 	return func(opts *testServerOptions) {
 		opts.accessTokenLifespan = d
+	}
+}
+
+// withRedisBackedStorage swaps the default in-memory storage for a
+// miniredis-backed *RedisStorage. This exercises the same Lua scripts and
+// Redis-shape key layout used in production, while remaining hermetic and
+// fast: each test gets its own miniredis instance with no external
+// dependencies. The {ns:test} hash tag in the key prefix matches the
+// production-shape cluster routing so multi-key Lua operations target the
+// same hash slot.
+//
+// The setup helper stashes the *miniredis.Miniredis on testServer.mr (accessed via testServer.Miniredis())
+// so tests can call FastForward(d) to advance Redis-side TTLs without
+// real-world sleeping.
+func withRedisBackedStorage() testServerOption {
+	return func(opts *testServerOptions) {
+		opts.storageFactory = func(t *testing.T) (storage.Storage, *miniredis.Miniredis) {
+			t.Helper()
+			mr := miniredis.RunT(t)
+			client := redis.NewClient(&redis.Options{Addr: mr.Addr()})
+			t.Cleanup(func() {
+				_ = client.Close()
+			})
+			return storage.NewRedisStorageWithClient(client, "test:auth:{ns:test}:"), mr
+		}
 	}
 }
 
@@ -130,8 +182,17 @@ func setupTestServer(t *testing.T, opts ...testServerOption) *testServer {
 	_, err = rand.Read(secret)
 	require.NoError(t, err)
 
-	// 3. Create storage
-	stor := storage.NewMemoryStorage()
+	// 3. Create storage. Tests opting into withRedisBackedStorage() get a
+	// miniredis-backed *RedisStorage; everyone else keeps the default
+	// in-memory backend. The mr return value is preserved so tests can
+	// FastForward Redis-side TTLs.
+	var (
+		stor storage.Storage = storage.NewMemoryStorage()
+		mr   *miniredis.Miniredis
+	)
+	if options.storageFactory != nil {
+		stor, mr = options.storageFactory(t)
+	}
 
 	// 4. Register test client (public client for PKCE)
 	err = stor.RegisterClient(ctx, &fosite.DefaultClient{
@@ -198,6 +259,7 @@ func setupTestServer(t *testing.T, opts ...testServerOption) *testServer {
 		PrivateKey: privateKey,
 		authServer: srv,
 		storage:    srv.IDPTokenStorage(),
+		mr:         mr,
 	}
 }
 
@@ -842,9 +904,18 @@ func TestIntegration_FullPKCEFlow_DefaultAudience(t *testing.T) {
 //   - UpstreamConfig{Type: OIDC, OIDCConfig: ...}
 //   - defaultUpstreamFactory dispatching to NewOIDCProvider
 //   - OIDCProviderImpl with OIDC discovery, ID token validation, and nonce support
-func setupTestServerWithOIDCProvider(t *testing.T, m *mockoidc.MockOIDC) *testServerWithUpstream {
+//
+// Variadic opts allow swapping the storage backend (e.g. withRedisBackedStorage);
+// the upstream is fixed because this helper exists specifically to exercise the
+// real OIDC factory path. Other testServerOptions are silently ignored.
+func setupTestServerWithOIDCProvider(t *testing.T, m *mockoidc.MockOIDC, opts ...testServerOption) *testServerWithUpstream {
 	t.Helper()
 	ctx := context.Background()
+
+	options := &testServerOptions{}
+	for _, opt := range opts {
+		opt(options)
+	}
 
 	cfg := m.Config()
 
@@ -857,8 +928,14 @@ func setupTestServerWithOIDCProvider(t *testing.T, m *mockoidc.MockOIDC) *testSe
 	_, err = rand.Read(secret)
 	require.NoError(t, err)
 
-	// 3. Create storage
-	stor := storage.NewMemoryStorage()
+	// 3. Create storage. See setupTestServer for the factory contract.
+	var (
+		stor storage.Storage = storage.NewMemoryStorage()
+		mr   *miniredis.Miniredis
+	)
+	if options.storageFactory != nil {
+		stor, mr = options.storageFactory(t)
+	}
 
 	// 4. Register test client (public client for PKCE)
 	err = stor.RegisterClient(ctx, &fosite.DefaultClient{
@@ -917,6 +994,7 @@ func setupTestServerWithOIDCProvider(t *testing.T, m *mockoidc.MockOIDC) *testSe
 			PrivateKey: privateKey,
 			authServer: srv,
 			storage:    srv.IDPTokenStorage(),
+			mr:         mr,
 		},
 		mockOIDC: m,
 	}
@@ -1634,9 +1712,18 @@ func TestIntegration_RefreshPreservesUpstreamTokenBinding(t *testing.T) {
 // configured as sequential upstream providers. This exercises the multi-upstream
 // authorization chain where the callback handler redirects to the next upstream
 // after each successful code exchange.
-func setupTestServerWithTwoUpstreams(t *testing.T, m1, m2 *mockoidc.MockOIDC) *testServer {
+//
+// Variadic opts allow swapping the storage backend (e.g. withRedisBackedStorage)
+// without affecting the rest of the chain wiring. Upstream-related options are
+// silently ignored because the helper hard-wires the two providers itself.
+func setupTestServerWithTwoUpstreams(t *testing.T, m1, m2 *mockoidc.MockOIDC, opts ...testServerOption) *testServer {
 	t.Helper()
 	ctx := context.Background()
+
+	options := &testServerOptions{}
+	for _, opt := range opts {
+		opt(options)
+	}
 
 	// 1. Generate RSA key for signing
 	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
@@ -1647,8 +1734,14 @@ func setupTestServerWithTwoUpstreams(t *testing.T, m1, m2 *mockoidc.MockOIDC) *t
 	_, err = rand.Read(secret)
 	require.NoError(t, err)
 
-	// 3. Create storage
-	stor := storage.NewMemoryStorage()
+	// 3. Create storage. See setupTestServer for the factory contract.
+	var (
+		stor storage.Storage = storage.NewMemoryStorage()
+		mr   *miniredis.Miniredis
+	)
+	if options.storageFactory != nil {
+		stor, mr = options.storageFactory(t)
+	}
 
 	// 4. Register test client (public client for PKCE)
 	err = stor.RegisterClient(ctx, &fosite.DefaultClient{
@@ -1752,6 +1845,7 @@ func setupTestServerWithTwoUpstreams(t *testing.T, m1, m2 *mockoidc.MockOIDC) *t
 		PrivateKey: privateKey,
 		authServer: srv,
 		storage:    srv.IDPTokenStorage(),
+		mr:         mr,
 	}
 }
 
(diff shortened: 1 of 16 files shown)


pkg/authserver/refresher.go
```diff
@@ -59,14 +59,15 @@ func (r *upstreamTokenRefresher) RefreshAndStore(
...
-		ExpiresAt:       newTokens.ExpiresAt,
-		UserID:          expired.UserID,
-		UpstreamSubject: expired.UpstreamSubject,
-		ClientID:        expired.ClientID,
+		ProviderID:       expired.ProviderID,
+		AccessToken:      newTokens.AccessToken,
+		RefreshToken:     newTokens.RefreshToken,
+		IDToken:          newTokens.IDToken,
+		ExpiresAt:        newTokens.ExpiresAt,
+		SessionExpiresAt: expired.SessionExpiresAt,
```
tgrunnagle: **F1 — Refresh path doesn't re-anchor `SessionExpiresAt` for rows missing it (consensus 9/10, MEDIUM)**

The PR's stated invariant is that `SessionExpiresAt` is set unconditionally on every callback so the refresh path is safe. That holds for fresh writes after this PR ships, but pre-PR Redis rows decode `SessionExpiresAt` as zero (the JSON field is absent). If such a legacy row is refreshed and the upstream rotation drops `expires_in`, this carries `expired.SessionExpiresAt=0` forward → both bounds are zero → the row gets stored without a Redis TTL with a `PERSIST`'d index, and Memory keeps it forever via `cleanupExpired`'s `IsZero()` guard. The exact "leak forever" outcome the design is meant to prevent.

Consider re-anchoring to a configured fallback when the carried-forward value is zero, e.g.:

```go
sessionBound := expired.SessionExpiresAt
if sessionBound.IsZero() {
    sessionBound = time.Now().Add(r.refreshTokenLifespan) // injected via constructor
}
```

Alternatively, log a `Warn` when this fallback fires so operators can track legacy rows. — *raised by storage-reviewer, oauth-reviewer*

pkg/authserver/storage/redis.go
```diff
@@ -832,14 +836,42 @@ else
...
+    redis.call('PERSIST', KEYS[2])
+elseif idxExisted == 0 then
+    -- Fresh set; our SADD created it. Apply our TTL.
+    redis.call('PEXPIRE', KEYS[2], ttlMs)
+else
+    -- Existing set; its TTL summarises prior members.
+    local idxTTL = redis.call('PTTL', KEYS[2])
+    if idxTTL == -1 then
+        -- A previous non-expiring write PERSIST'd it. Leave it alone —
+        -- applying a TTL here is the bug we are fixing.
```
tgrunnagle, line 879: **F2 — Same-provider rewrite from non-expiring to expiring leaves index permanently `PERSIST`'d (consensus 9/10, LOW)**

The Lua-script doc block is excellent for the cross-provider mixed-expiry case. The same-provider rewrite trade-off (the only non-expiring member is rewritten as expiring) is pinned by `same provider rewrite from non-expiring to expiring keeps PERSIST'd until rewrite` in `redis_test.go`, but the production-code comment doesn't mention it. A future maintainer who tightens the rule will see a cryptic test failure with `acceptable limitation` and have to reverse-engineer why.

```suggestion
        -- applying a TTL here is the bug we are fixing.
        -- Limitation: if the only non-expiring member of a session is later rewritten
        -- as expiring (same provider), this branch keeps the index PERSIST'd. Cleanup
        -- relies on DeleteUpstreamTokens for the whole session. We accept this rather
        -- than tracking per-member TTL state in Lua.
```

— *raised by lua-reviewer, storage-reviewer, oauth-reviewer, compat-reviewer, test-reviewer*

pkg/authserver/storage/redis.go
```diff
@@ -887,11 +925,20 @@ func marshalUpstreamTokensWithTTL(tokens *UpstreamTokens) ([]byte, time.Duration
...
+	// Zero ExpiresAt means the token never expires; return ttl=0 so the Lua script
+	// stores the key without a Redis TTL.
+	var ttl time.Duration // zero means no Redis TTL (non-expiring token with no known session bound)
 	if !tokens.ExpiresAt.IsZero() {
 		ttl = time.Until(tokens.ExpiresAt) + DefaultRefreshTokenTTL
-		if ttl < 0 {
-			ttl = DefaultRefreshTokenTTL
+		if ttl <= 0 {
+			// Access token and its refresh grace period have both passed — evict promptly.
+			ttl = time.Second
```
tgrunnagle, line 945: **F3 — Silent semantic change: stale-token TTL fallback dropped from 30d to 1s (consensus 8/10, MEDIUM)**

Pre-PR, when `time.Until(tokens.ExpiresAt) + DefaultRefreshTokenTTL` was negative (access token + its 30-day refresh grace had both passed), the marshal returned `ttl = DefaultRefreshTokenTTL` — the row got a 30-day grace. Now it gets 1 second. Same change applies to the new `SessionExpiresAt` branch (line 941). The cold-path callers that hit this (refresher races, admin tools, legacy-row rewrites) silently observe a different lifetime, and there's no regression test for either negative-edge branch.

If the intent is "evict promptly because the row is no longer useful," worth either preserving the prior fallback (`ttl = DefaultRefreshTokenTTL`) for the access-token branch, or adding a regression test that pins the new contract:

```go
t.Run("rewriting a deeply-stale token evicts in ~1s", func(t *testing.T) {
    // hand-roll a write whose ExpiresAt is 60d in the past, assert mr.TTL ≈ 1s
})
```

— *raised by compat-reviewer, storage-reviewer*

pkg/authserver/storage/redis_test.go
```diff
@@ -658,6 +658,208 @@ func TestRedisStorage_UpstreamTokens(t *testing.T) {
 			assert.Equal(t, "no-expiry-token", retrieved.AccessToken)
 			assert.Equal(t, "test-provider", retrieved.ProviderID)
 			assert.True(t, retrieved.ExpiresAt.IsZero())
+
+			// Verify the key has no Redis TTL (miniredis returns 0 for keys without expiry).
+			key := redisUpstreamKey(s.keyPrefix, "no-expiry", "provider-a")
+			assert.Equal(t, time.Duration(0), mr.TTL(key), "non-expiring token must have no Redis TTL")
```
tgrunnagle, line 664: **F4 — Missing regression test for legacy Redis JSON decode (no `session_expires_at` field) (consensus 9/10, LOW)**

The new `storedUpstreamTokens` struct adds `session_expires_at` without `omitempty`, and the read path uses a `!= 0` sentinel — this is correctly forward/backward compatible, but no test exercises it. A future change that, e.g., switches to `json.NewDecoder().DisallowUnknownFields()` or renames the JSON tag would silently break legacy data without failing this suite.

A small subtest that hand-rolls a legacy JSON string directly into miniredis would pin the wire-shape compat:

```go
t.Run("legacy JSON without session_expires_at decodes safely", func(t *testing.T) {
    withRedisStorage(t, func(ctx context.Context, s *RedisStorage, mr *miniredis.Miniredis) {
        legacyJSON := `{"provider_id":"github","access_token":"legacy","expires_at":1700000000}`
        key := redisUpstreamKey(s.keyPrefix, "legacy-sess", "github")
        require.NoError(t, mr.Set(key, legacyJSON))
        got, err := s.GetUpstreamTokens(ctx, "legacy-sess", "github")
        require.NoError(t, err)
        require.NotNil(t, got)
        assert.True(t, got.SessionExpiresAt.IsZero())
    })
})
```

— *raised by compat-reviewer, storage-reviewer, test-reviewer*

pkg/authserver/integration_test.go
```diff
@@ -1329,6 +1331,72 @@ func TestIntegration_UpstreamTokenService_RefreshExpiredTokens(t *testing.T) {
 	_ = originalAccessToken // used only to confirm the flow completed
 }
 
+// TestIntegration_UpstreamTokenService_NonExpiringToken verifies that a token with
+// a zero ExpiresAt is treated as non-expiring: GetValidTokens must return the
+// stored access token unchanged and must not attempt a refresh. If a refresh were
+// triggered, mockoidc would return an error because no user is queued — that
+// outcome is the failure signal for this test.
```
tgrunnagle, line 1416: **F5 — Misleading comment: this is not the failure mode (consensus 8/10, LOW)**

mockoidc only consumes `UserQueue` at `/authorize` (`handlers.go: m.UserQueue.Pop()`). The `/token` endpoint's refresh-grant validates against `SessionStore` and would return 200 with new tokens regardless of queued users. The actual regression signal in this test is the access-token equality and `IsZero()` assertions — those are doing the work.

```suggestion
// triggered, the access token returned by GetValidTokens would differ from the
// stored value and ExpiresAt would gain a non-zero value (mockoidc's refresh
// response carries expires_in). The assertions below pin both invariants.
```

The same misleading phrasing also appears around line 1965 in `TestIntegration_FullFlow_NonExpiringUpstreamToken`. — *raised by test-reviewer*

pkg/authserver/refresher_test.go
```diff
@@ -113,6 +115,39 @@ func TestUpstreamTokenRefresher_RefreshAndStore(t *testing.T) {
...
+						assert.True(t, tokens.ExpiresAt.IsZero(),
+							"new ExpiresAt should be zero (provider omitted expires_in)")
+						return nil
+					})
+			},
+			checkResult: func(t *testing.T, result *storage.UpstreamTokens) {
+				t.Helper()
+				assert.Equal(t, sessionBound, result.SessionExpiresAt,
+					"returned tokens must also carry SessionExpiresAt forward")
+				assert.True(t, result.ExpiresAt.IsZero())
```
tgrunnagle, line 148: **F6 — Missing test for refresh rotation across expiring↔non-expiring (consensus 7/10, LOW)**

The new case covers refresh-TO-non-expiring (provider drops `expires_in` on rotation). The inverse — `expired` has zero `ExpiresAt` + non-zero `SessionExpiresAt`, and the provider returns a token with non-zero `ExpiresAt` — isn't covered. This rotation transitions the storage row from "no Redis TTL" to a TTL'd state, which is exactly the kind of asymmetric edge that's easy to break in a future refactor.

Suggested table case (sketch):

```go
{
    name:      "refresh that gains expires_in transitions row to TTL'd",
    sessionID: "transition",
    expired: &storage.UpstreamTokens{
        // ExpiresAt zero, SessionExpiresAt set
        SessionExpiresAt: sessionBound,
        // ...
    },
    setupProvider: /* returns Tokens with non-zero ExpiresAt */,
    setupStorage: /* assert tokens.ExpiresAt != zero, SessionExpiresAt preserved */,
}
```

— *raised by test-reviewer*

pkg/authserver/storage/types.go
```diff
@@ -81,9 +90,18 @@ type UpstreamTokens struct {
...
+//
+// This method is intended for storage eviction decisions only — it checks exact
+// expiry with no preemptive buffer. For "should I refresh before using?" decisions,
+// use upstream.Tokens.IsExpiredAt which includes a preemptive buffer window.
 func (t *UpstreamTokens) IsExpired(now time.Time) bool {
-	return now.After(t.ExpiresAt)
+	if t == nil {
+		return true
+	}
+	return !t.ExpiresAt.IsZero() && now.After(t.ExpiresAt)
```
tgrunnagle, line 107: **F7 — `IsExpired` contract change should be in release notes (consensus 8/10, LOW)**

Pre-PR `IsExpired` returned `true` for any zero `ExpiresAt` (because `now.After(time.Time{})` is true). Post-PR it returns `false`. All current callers in this repo are internal and are updated correctly, but `IsExpired` is exported on `*storage.UpstreamTokens` and could be called by downstream consumers (operator, vmcp, future packages). The PR description has a release-note suggestion — please extend it with one line on the IsExpired/IsExpiredAt semantic flip so anyone calling these methods directly knows to audit their assumptions.

Suggested addition to the release note:

> Behavior change: `storage.UpstreamTokens.IsExpired` and `upstream.Tokens.IsExpiredAt` now treat a zero `ExpiresAt` as non-expiring (previously: expired). Callers that depended on the old semantics for non-expiring tokens should audit their behavior.

— *raised by compat-reviewer*

pkg/authserver/upstream/oauth2.go
```diff
@@ -457,9 +447,9 @@ func (p *BaseOAuth2Provider) exchangeCodeForTokens(ctx context.Context, code, co
 		return nil, err
 	}
 
-	slog.Info("authorization code exchange successful",
+	slog.Debug("authorization code exchange successful",
```
tgrunnagle, line 450: **F8 — INFO → DEBUG demotion without compensating metric (consensus 8/10, LOW)**

The demotion (here and on line 487, plus `oidc.go:262, 408`) matches the project's `silent success at INFO+` rule. Operationally, however, these were the only positive signals that the OAuth code-exchange and refresh paths were healthy in production — without them, ops can only observe failures and has no baseline. If a counter like `authserver_upstream_token_exchange_total{outcome=...}` and `authserver_upstream_token_refresh_total{outcome=...}` already exists in your telemetry, no action needed. Otherwise consider adding one near these call sites so SREs retain a per-provider success rate.

— *raised by compat-reviewer, oauth-reviewer*

pkg/authserver/storage/types.go
```diff
@@ -61,7 +61,16 @@ type UpstreamTokens struct {
...
-	ExpiresAt    time.Time
+	// ExpiresAt is when the access token expires. Zero value means the provider
+	// did not assert an expiry; callers must treat it as non-expiring.
+	ExpiresAt time.Time
+	// SessionExpiresAt is the Fosite session expiry time, set on every write by
+	// the callback handler. Storage backends use it as a fallback storage lifetime
+	// when ExpiresAt is zero (non-expiring upstream token), bounding the row to the
+	// Fosite session. Set unconditionally — including for tokens with their own
+	// ExpiresAt — so the refresh path is safe even if the upstream provider stops
+	// asserting expires_in on a later rotation.
```
tgrunnagle: **F9 — `SessionExpiresAt` godoc misleading about who writes it (consensus 8/10, LOW)**

The doc reads as if only the callback handler sets this field, but `refresher.go:67` also writes it (carries it forward unchanged on every refresh). The new test `preserves SessionExpiresAt when provider omits expires_in` pins exactly that.

```suggestion
	// SessionExpiresAt is the Fosite session expiry time. The callback handler
	// sets it from RefreshTokenLifespan on every initial write; the refresher
	// carries it forward unchanged on each subsequent refresh. Storage backends
	// use it as a fallback storage lifetime when ExpiresAt is zero (non-expiring
	// upstream token), bounding the row to the Fosite session. Set unconditionally
	// — including for tokens with their own ExpiresAt — so the refresh path is
	// safe even if the upstream provider stops asserting expires_in on a later
	// rotation.
```

— *raised by quality-reviewer*

pkg/authserver/storage/memory.go
```diff
@@ -222,7 +222,11 @@ func (s *MemoryStorage) cleanupExpired() {
 
 	var expiredUpstreamTokens []upstreamKey
 	for k, v := range s.upstreamTokens {
-		if now.After(v.expiresAt) {
+		// Zero expiresAt is the sentinel for "no TTL" (non-expiring token with no session
+		// bound). Other entry types never use zero expiresAt, so only upstream tokens need
```
tgrunnagle, line 226: **F10 — Comment asserts an unenforced invariant about other entry types (consensus 7/10, LOW)**

The claim "Other entry types never use zero `expiresAt`" is true today but unenforced — a future contributor adding a new entry type with optional expiry would silently have every entry swept on the next cleanup tick, because `time.Time{}` compares as before any real time. The comment makes the assumption explicit (which is good) but doesn't lock it down.

Two low-cost options: (a) extend the same `!v.expiresAt.IsZero() &&` guard to the other cleanup loops in this function for uniform sentinel handling, or (b) reject zero `expiresAt` at the corresponding `Store*` entry points so the invariant fails loudly at write time rather than at the next cleanup. — *raised by quality-reviewer, storage-reviewer*

tgrunnagle (commented):
**Agents consulted**: storage-reviewer, lua-reviewer, oauth-reviewer, test-reviewer, compat-reviewer, quality-reviewer

### Consensus Summary
| 1 | Refresh path doesn't re-anchor `SessionExpiresAt` for legacy/edge-case rows | 9/10 | MEDIUM | Discuss / fix |
| 2 | Same-provider non-expiring → expiring rewrite leaves index permanently `PERSIST`'d | 9/10 | LOW | Document in code |
| 3 | Stale-token TTL fallback narrowed from `DefaultRefreshTokenTTL` (30d) to `time.Second` | 8/10 | MEDIUM | Discuss |
| 4 | Missing regression test for legacy Redis JSON decoding (no `session_expires_at` field) | 9/10 | LOW | Add test |
| 5 | Misleading comment about mockoidc refresh-failure mode in non-expiring integration tests | 8/10 | LOW | Fix comment |
| 6 | Missing test for refresh rotation across expiring↔non-expiring | 7/10 | LOW | Add test |
| 7 | `IsExpired`/`IsExpiredAt` contract change should be documented in release notes | 8/10 | LOW | Note in release |
| 8 | INFO → DEBUG log demotion without compensating metric | 8/10 | LOW | Discuss observability |
| 9 | `SessionExpiresAt` godoc misleading about who writes it | 8/10 | LOW | Fix doc |
| 10 | `cleanupExpired` comment asserts unenforced invariant | 7/10 | LOW | Strengthen |
| 11 | `stripExpiresInMiddleware` test helper has fragility issues | 7/10 | LOW | Harden |

### Overall

This is a well-engineered bug fix that correctly diagnoses two distinct issues (the fabricated 1h `ExpiresAt` and an independent Lua TTL inversion) and fixes them with comprehensive regression coverage. The Lua-script comment block explaining the `EXISTS`-pre-`SADD` discrimination is particularly good — it makes the bug-prone branch self-documenting. Wire compatibility is preserved (Go's default `json.Unmarshal` ignores unknown fields, so v_old↔v_new co-existence works during a rolling deploy), and the pluggable storage-backend semantics for zero `ExpiresAt` are now consistent between Memory and Redis.

The consensus findings are mostly polish — several documentation/comment inaccuracies, two test gaps, and a couple of observability concerns. The two MEDIUM findings are worth a closer look before merge:

- **F1** identifies a residual gap in the PR's stated invariant (`SessionExpiresAt` is set unconditionally → refresh path is safe). For pre-PR Redis rows, `SessionExpiresAt` decodes to zero. If such a row's upstream rotates to no `expires_in` on the first refresh, the refresher carries `SessionExpiresAt=0` forward and the row becomes indefinitely retained. Blast radius is bounded by the legacy 30-day TTL, but the invariant doesn't hold for in-flight legacy data. Worth either a defensive re-anchor in `RefreshAndStore` when `expired.SessionExpiresAt.IsZero()`, or a documented operational mitigation.

- **F3** is a silent semantic change in `marshalUpstreamTokensWithTTL` — when both access token and the 30-day refresh grace have passed, the row's TTL drops from `DefaultRefreshTokenTTL` (30d) to `time.Second`. This is a cold path (writing a deeply-stale token), but it's a real behavior change with no regression test.

No blocking issues. The same-provider rewrite limitation (F2) is acknowledged in tests as acceptable; surfacing it in the production Lua-script comment block would help future maintainers.

### Documentation

- `pkg/authserver/storage/types.go:67` — `SessionExpiresAt` godoc claims "set on every write by the callback handler," but the refresher also writes it. Fix to reflect both writers.
- `pkg/authserver/storage/types.go:104` — `IsExpired` semantic change should appear in the release notes alongside the existing suggested note.
- `pkg/authserver/storage/redis.go:869` — Lua-script comment block is excellent; consider one extra line documenting the same-provider rewrite trade-off for symmetry with the test that pins it.
- `pkg/authserver/integration_test.go:1337` and ~1965 — "if a refresh were triggered, mockoidc would return an error" is incorrect; mockoidc's `UserQueue` is consumed only at `/authorize`, not at `/token` refresh-grant. Tests still pin the regression via access-token comparison, so the comment is the only thing wrong.

### Test gaps to consider

- A small `redis_test.go` subtest that hand-rolls a legacy JSON byte string (no `session_expires_at` field) via `mr.Set(...)` and asserts round-trip stability — pins the wire-shape contract.
- A `refresher_test.go` case where `expired` has zero `ExpiresAt` + non-zero `SessionExpiresAt` and the provider returns a token with non-zero `ExpiresAt` (refresh-rotation across expiring↔non-expiring).

---
*Generated with Claude Code*

pkg/authserver/integration_test.go
```diff
@@ -1866,3 +1858,354 @@ func TestIntegration_MultiUpstreamSequentialChain(t *testing.T) {
...
+			capture := newCaptureResponseWriter()
+			next.ServeHTTP(capture, req)
+
+			// Only rewrite successful JSON token responses. Errors and
+			// non-JSON bodies are passed through unchanged.
+			body := capture.body.Bytes()
+			contentType := capture.header.Get("Content-Type")
+			if capture.status == http.StatusOK && strings.Contains(contentType, "json") {
+				var payload map[string]interface{}
+				if err := json.Unmarshal(body, &payload); err == nil {
```
tgrunnagle, line 2005: **F11 — `stripExpiresInMiddleware` test helper has minor fragility (consensus 7/10, LOW)**

Three small concerns about this helper:

1. `payload` could be `nil` after a successful `Unmarshal` of a `null` body — `delete` no-ops safely but the subsequent `Marshal` emits `null`, which is silently incorrect for a token response. A `if payload == nil { return /* pass through */ }` guard documents the contract.
2. `map[string]interface{}` round-trip risks float-precision drift if mockoidc ever adds a numeric top-level field. Currently mockoidc's token response only has string fields plus the integer `expires_in` we delete, so this is safe today — worth a comment noting the assumption.
3. `captureResponseWriter` (around line 1875) implements only `Header`/`WriteHeader`/`Write`. Adequate for mockoidc's buffered handlers; a doc comment noting the limitation would help if this is ever reused for streaming responses.

All three are tiny — bundling them here so the helper is robust before another test reaches for it. — *raised by test-reviewer, quality-reviewer*

pkg/authserver/refresher.go
jhrozek: Good catch — fixed in 6b2ef9b8b. The refresher now defensively re-anchors `SessionExpiresAt` to `now + RefreshTokenLifespan` when both `expired.SessionExpiresAt` and `newTokens.ExpiresAt` are zero. The lifespan value is threaded through the `upstreamTokenRefresher` from the same `cfg.RefreshTokenLifespan` the callback uses, so the re-anchor matches what a fresh callback would have set. Pinned by the new test `re-anchors SessionExpiresAt when legacy row and provider both omit expiry`.

pkg/authserver/storage/redis.go
jhrozek: Fixed in 346823ec8 — added a "Known trade-off" paragraph to the Lua doc block explaining the same-provider rewrite case, with a reference to the pinning test by name.

pkg/authserver/storage/redis.go
jhrozek: Agreed the new `time.Second` clamp is correct (a 30-day grace on a deeply-stale token would be a memory leak). Added regression tests in b8f52faf6 that pin both branches: `deeply stale ExpiresAt branch clamps TTL to one second` and `deeply stale SessionExpiresAt branch clamps TTL to one second`. Both assert `mr.TTL(key) <= 2*time.Second`, so a future revert to `DefaultRefreshTokenTTL` would fail loudly.

pkg/authserver/storage/redis_test.go
jhrozek: Fixed in 1b05c7b4c — added `legacy JSON without session_expires_at decodes with zero SessionExpiresAt`. Hand-rolls a pre-PR JSON via `mr.Set(...)`, asserts round-trip stability and `SessionExpiresAt.IsZero()`. A future `DisallowUnknownFields()` flip would fail this test on `require.NoError`.

pkg/authserver/integration_test.go
jhrozek: You're right about mockoidc's UserQueue mechanics — the comment is misleading even though the assertions are correct. Filing this as a follow-up since it's comment-only and the regression signal still works.

pkg/authserver/refresher_test.go
jhrozek: Considered but holding off in this PR. The refresher's code path is symmetric — the same three lines (`ExpiresAt` write, `SessionExpiresAt` carry-forward, store) run in both directions. There's no separate branch that could regress one direction without the other. Filing a follow-up issue for the inverse case as a hygiene addition.

pkg/authserver/storage/types.go
jhrozek: Will extend the release note in the PR description to call out the `IsExpired`/`IsExpiredAt` semantic flip. Suggested wording:

> Behavior change for direct API consumers: `storage.UpstreamTokens.IsExpired(now)` and `IsExpiredAt(now)` now return `false` when `ExpiresAt` is the zero value (previously returned `true`). Zero `ExpiresAt` is the documented sentinel for "upstream did not assert an expiry." Callers that relied on the old behavior to detect non-expiring tokens should switch to checking `tokens.ExpiresAt.IsZero()` explicitly.

No in-repo callers are affected — verified.

pkg/authserver/upstream/oauth2.go
jhrozek: Fair concern — confirmed `pkg/authserver/` has zero metrics infrastructure today, so the demotion strictly reduces observability. I'd rather file this as a follow-up issue than add a new metrics surface in a focused fix PR — picking meter names, label cardinality, and `pkg/telemetry/providers` wiring is its own design. Suggested follow-up scope: `authserver_upstream_token_{exchange,refresh,userinfo}_total{provider, outcome}` counters + duration histograms. Keeping the DEBUG demotion for now.

pkg/authserver/storage/types.go
jhrozek: Fixed in 80a7f9f02 — godoc now lists all three writers: callback handler (initial), refresher carry-forward, and refresher re-anchor for legacy rows.

pkg/authserver/storage/memory.go
jhrozek: The two options you suggested cut against each other (extending the guard would mask a real bug — a tokens type that should have an expiry but accidentally produced zero would silently never evict). The existing comment is the contract, and panicking from a background cleanup goroutine has its own DoS risk. I think the comment alone is right for now; if a new entry type is added that legitimately uses optional expiry, that'd be the moment to revisit. Leaving as-is.

pkg/authserver/integration_test.go
jhrozek: Agreed all three are speculative for current call sites. Filing as a follow-up for the scope comment + any of the three sub-concerns that prove material when the helper is reused.

tgrunnagle (approved):

PR #10, opened 2025-03-26
Implement secrets management

dmjb:
Implements a set of secrets management commands, along with a flag for
`run` which allows secrets to be injected as environment variables.

Note that this implementation stores the secrets to an unencrypted file
on disk. This will be changed in an upcoming PR.

Example of use:

```
$ bin/vt secret list
Available secrets:
  - fizz
  - foo
$ bin/vt run --secret=foo,target=FOO --secret=fizz,target=FIZZ alpine -- env
Logging to: /tmp/vibetool-alpine.log
MCP server alpine-1742982443 is running in the background (PID: 18425)
Use 'vibetool stop alpine-1742982443' to stop the server
$ docker ps -a
CONTAINER ID   IMAGE                           COMMAND                  CREATED              STATUS                          PORTS     NAMES
ed1bda951b03   alpine                          "env"                    About a minute ago   Exited (0) About a minute ago             alpine-1742982443
$ docker logs ed1bda951b03
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=ed1bda951b03
FOO=bar
FIZZ=buzz
MCP_TRANSPORT=stdio
HOME=/root
```

PR #5094, opened 2026-04-28
Allow OAuth2 upstreams to omit userInfo config

tgrunnagle:
## Summary

- **Why**: Some OAuth 2.0 authorization servers — notably MCP authorization servers per the MCP authorization spec — expose no userinfo endpoint and issue opaque (non-JWT) access tokens. Atlassian's MCP server (`cf.mcp.atlassian.com`) is a concrete example: its discovery doc advertises only `authorization_endpoint`, `token_endpoint`, `registration_endpoint`, and `revocation_endpoint`; there is no userinfo or introspection surface, and the access token (93-char opaque) carries no decodable claims. With `OAuth2UpstreamConfig.userInfo` previously required, the embedded auth server could not integrate any such upstream — token exchange would succeed and the userinfo step would then fail (e.g. 401 against an unrelated identity API), surfacing as `ErrIdentityResolutionFailed` and aborting the auth flow.
- **What**:
  - Make `OAuth2UpstreamConfig.userInfo` optional in the v1beta1 CRD (drop `+kubebuilder:validation:Required`, add `+optional` and `omitempty`). Mirror the change on `authserver.OAuth2UpstreamRunConfig.UserInfo` (JSON/YAML tag now `omitempty`).
  - In `BaseOAuth2Provider.ExchangeCodeForIdentity`, when `UserInfo` is nil, skip the userinfo HTTP call and synthesize a deterministic, non-PII `Subject` from a SHA-256 prefix of the access token via the new `synthesizeSubjectFromAccessToken`. Subject format: `tk-` prefix + 32 hex chars (35 total) so synthesized subjects are recognizable in logs and JWTs. `Name` and `Email` remain empty in this mode.
  - Scoped to OAuth2: OIDC providers always have an ID-token-derived subject and never reach this path.
  - Synthesized subject is stable for the lifetime of the access token and rotates on re-authentication — sufficient for the proxy's session-key role.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Manual testing: built and deployed the operator + proxyrunner against a kind cluster with an `MCPRemoteProxy` configured to proxy Atlassian's hosted MCP server (`https://mcp.atlassian.com/v1/mcp`) using DCR + an embedded auth server with no `userInfo` block. Walked the full OAuth flow in a browser — DCR registration succeeded against `cf.mcp.atlassian.com/v1/register`, authorization-code exchange succeeded, and the embedded auth server reached a successful `Identity` with a synthesized `tk-…` subject (where it previously failed at the userinfo step with HTTP 401).

New automated coverage:

- `TestBaseOAuth2Provider_ExchangeCodeForIdentity / synthesizes identity when UserInfo is nil` — drives the full exchange path with `UserInfo = nil`, asserts the returned identity carries the prefix and the deterministic hash, asserts empty `Name`/`Email`, and uses a tripwire userinfo server that fails the test if any HTTP request reaches a userinfo endpoint.
- `TestSynthesizeSubjectFromAccessToken` — table-driven: deterministic for the same input, distinct for different inputs, output shape (`tk-` prefix + 32 valid hex chars), and stable on empty input.

## API Compatibility

This change is **additive**: it relaxes a required field to optional. CRs that already set `userInfo` continue to work unchanged; new CRs may omit it.

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go` | Drop `+kubebuilder:validation:Required` on `OAuth2UpstreamConfig.UserInfo`; add `+optional` and `omitempty`; expand the doc comment to describe the synthesized-subject mode. |
| `pkg/authserver/config.go` | `OAuth2UpstreamRunConfig.UserInfo` JSON/YAML tags become `omitempty`; doc comment cross-references the synthesis rule. |
| `pkg/authserver/upstream/oauth2.go` | Add `crypto/sha256` and `encoding/hex` imports. In `ExchangeCodeForIdentity`, early-return with a synthesized `Identity` when `p.config.UserInfo == nil`. New helper `synthesizeSubjectFromAccessToken` and exported `synthesizedSubjectPrefix` constant. |
| `pkg/authserver/upstream/oauth2_test.go` | New `synthesizes identity when UserInfo is nil` subtest with userinfo-tripwire server; new `TestSynthesizeSubjectFromAccessToken` table-driven test. |
| `deploy/charts/operator-crds/...` | Regenerated CRD manifests (both `mcpexternalauthconfigs` and `virtualmcpservers`, both `files/crds/` and `templates/`). |
| `docs/operator/crd-api.md`, `docs/server/...` | Regenerated CRD API reference and swagger docs. |

Yes — operators can now configure OAuth2 upstream providers in `MCPExternalAuthConfig` without a `userInfo` block. The embedded auth server will synthesize a non-PII subject (prefixed with `tk-`) from the access token. This unblocks integrating with OAuth 2.0 authorization servers that expose no userinfo endpoint, including MCP authorization servers per the MCP authorization spec.

When `userInfo` is configured, behavior is unchanged.

## Implementation plan

<details>

The change was decided after discovering, while wiring up an Atlassian MCP integration, that:

1. The MCP authorization server has no userinfo endpoint (verified by probing every reasonable path variant under both `mcp.atlassian.com` and `cf.mcp.atlassian.com` — all 404).
2. There is no OIDC discovery alternative (`/.well-known/openid-configuration` 404).
3. The access token is opaque (length 93, 0 dots — not a JWT, no claims to decode locally).
4. The token cannot be used against general Atlassian Cloud APIs like `api.atlassian.com/me` (401 — different OAuth provider scope).

So there is genuinely no way for a downstream client to resolve user identity from the token. Three options were considered:

- **A. Probe for an undocumented userinfo endpoint** — eliminated; nothing exists at any reasonable path.
- **B. Decode the access token if it's a JWT** — eliminated; token is opaque.
- **C. Synthesize a non-PII subject from the access token** — selected; the only viable path that lets the embedded auth server reach a usable `Identity` for upstreams with no identity surface.

Implementation:

1. CRD: relax `OAuth2UpstreamConfig.userInfo` from required to optional (additive — does not break the API).
2. Runtime: in `BaseOAuth2Provider.ExchangeCodeForIdentity`, when `UserInfo` is nil, skip the userinfo call and return an `Identity` whose `Subject` is a SHA-256 prefix of the access token, with empty `Name`/`Email`.
3. Scope to OAuth2 — OIDC providers always have an ID-token-derived subject and are unaffected.
4. Add unit tests covering the synthesis path (including a userinfo-tripwire server to assert no HTTP request is issued) and the helper's properties (deterministic, prefix-tagged, non-empty for empty input, valid hex).

</details>

- **Subject rotates on re-authentication.** A user who re-authenticates (e.g., after token expiry) gets a different `sub` for the new session. That is fine for stateless proxying but breaks any per-user state stored under the subject. As far as I can tell, the embedded auth server uses `Subject` only as a session key, not as a user-facing identity, so this is acceptable. Worth a second pair of eyes on that assumption.
- **Should there be an explicit opt-in instead of implicit-when-nil?** I opted for implicit (omitting `userInfo` switches to synthesis) because the synthesis mode is the only sensible behavior for an OAuth2 upstream with no userinfo surface — there is no other useful interpretation of "OAuth2 upstream without userinfo configured". An explicit field (e.g., `subjectStrategy: opaqueTokenHash`) would be more discoverable but at the cost of an extra knob. Happy to reshape if reviewers prefer the explicit form.
- **Keep `userInfo` configured when the upstream supports it.** This relaxation only kicks in when `userInfo` is omitted; CRs that set it continue through the existing userinfo path and get real `Subject`/`Name`/`Email` claims.
- **OIDC providers are unaffected.** OIDC always has an ID token; the synthesis path is unreachable from `OIDCProviderImpl`.

## Large PR Justification

- Most lines are documentation, comments, or tests
- Remaining product code lines should all be done in one PR


pkg/authserver/upstream/oauth2.go
```diff
@@ -400,14 +402,28 @@ func (p *BaseOAuth2Provider) buildAuthorizationURL(
...
 		return nil, err
 	}
 
+	// No userinfo configured: synthesize a deterministic, non-PII subject from
+	// the access token. Stable for the lifetime of the token; rotates when the
+	// user re-authenticates.
+	if p.config.UserInfo == nil {
+		return &Identity{
+			Tokens:  tokens,
+			Subject: synthesizeSubjectFromAccessToken(tokens.AccessToken),
```
tgrunnagle: **[HIGH — Architecture/Security] #1 Synthesized subject rotates per access token — unbounded internal user-record growth (consensus 9/10)**

Access tokens rotate on every authorization code flow, so this synthesized `Subject` rotates per re-auth. `pkg/authserver/server/handlers/callback.go:112` passes it to `UserResolver.ResolveUser`, which in `pkg/authserver/server/handlers/user.go:55` calls `createUserWithIdentity` whenever the `(providerID, providerSubject)` pair is unknown — i.e., every re-authentication of the same human creates a new `storage.User` row + a new `ProviderIdentity` row. For a long-running auth server fronting a synthesis-mode upstream, the `users` table grows monotonically and audit/per-user state fragments across many internal IDs.

**Recommendation**: bypass `UserResolver` for synthesis-mode upstreams (use the synthesized value directly as an ephemeral session key without persisting a User row), or acknowledge the growth in `docs/arch/` and add a janitor for orphaned synthesized identities. Refresh-token rotation is already safe (`refresher.go:55-68` preserves `expired.UpstreamSubject`), so this is strictly about new auth flows.

*Raised by: Security, OAuth/OIDC, Architecture*

pkg/authserver/upstream/oauth2.go
```diff
@@ -400,14 +402,28 @@ func (p *BaseOAuth2Provider) buildAuthorizationURL(
...
 func (p *BaseOAuth2Provider) ExchangeCodeForIdentity(ctx context.Context, code, codeVerifier, _ string) (*Identity, error) {
 	tokens, err := p.exchangeCodeForTokens(ctx, code, codeVerifier)
 	if err != nil {
 		return nil, err
 	}
 
+	// No userinfo configured: synthesize a deterministic, non-PII subject from
+	// the access token. Stable for the lifetime of the token; rotates when the
+	// user re-authenticates.
+	if p.config.UserInfo == nil {
```
tgrunnagle, line 429: **[HIGH — Architecture/Security] #2 Implicit-when-nil activates silently on misconfiguration (consensus 9/10)**

This branch fires for any reason `UserInfo` ends up nil — a YAML typo (`userinfos:`, wrong indentation under `oauth2Config`), a missing block due to copy-paste, an outdated CRD bundle, or an operator who didn't realize the field is now optional. There is no `slog.Warn` at provider construction, no status condition, and no error. With `+kubebuilder:validation:Required` removed, an operator with a working configuration today could regress into synthesis mode by accident on a future apply and only notice when downstream code expecting `name`/`email` claims silently degrades. Per `.claude/rules/go-style.md` "Constructor Validation: Fail Loudly on Invalid Input".

**Recommendation** (in increasing strictness):
1. **Minimum**: emit `slog.Warn` once at provider construction when `UserInfo == nil` with `provider_id` and a sentence calling out that no real identity is being resolved. Add a CR status condition `UpstreamUserInfoMode=Synthesized` (parallel to `ConditionTypeAuthzUpstreamSelectionWarning` at `virtualmcpserver_types.go:283`).
2. **Defensive**: add an explicit `synthesizeSubject *bool` (or `subjectStrategy: opaqueTokenHash | userInfo` enum) on `OAuth2UpstreamConfig`. Reject configs where `userInfo == nil && synthesizeSubject != true` via CEL/webhook.

The PR description acknowledges this tradeoff and asks reviewers — the consensus is that the explicit form (or at minimum the warning) is worth the extra knob.

*Raised by: Security, OAuth/OIDC, Architecture, Operator*

pkg/authserver/upstream/oauth2_test.go
```diff
@@ -1198,6 +1201,103 @@ func TestBaseOAuth2Provider_ExchangeCodeForIdentity(t *testing.T) {
...
+	t.Run("synthesizes identity when UserInfo is nil", func(t *testing.T) {
+		t.Parallel()
+
+		mock := newMockOAuth2Server()
+		t.Cleanup(mock.Close)
+
+		// Tripwire: any HTTP request to a userinfo endpoint when UserInfo is
+		// nil is a regression. The provider should not contact one.
+		var userinfoHits int32
+		userInfoTripwire := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
```
tgrunnagle: **[MEDIUM — Test Coverage] #3 Tripwire is decorative (consensus 8/10)**

`userInfoTripwire`'s URL is never wired into the `OAuth2Config` (the comment says `// UserInfo intentionally nil`). The production code at `oauth2.go:420` short-circuits on `p.config.UserInfo == nil` before any HTTP path could touch the tripwire, so `userinfoHits == 0` (line 1256) is structurally guaranteed — vacuously true. The tripwire only catches a nonexistent regression class.

**Recommendation**: either drop the tripwire and rely on the existing `Name`/`Email`-empty assertions (which actually prove the synthesis branch was taken), or split into a paired test that wires the tripwire URL into a configured `UserInfoConfig` and asserts userinfo IS called when configured — that's the regression class worth catching.

*Raised by: Test Coverage, General code quality*

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -365,9 +365,14 @@ type OAuth2UpstreamConfig struct {
...
 	// UserInfo contains configuration for fetching user information from the upstream provider.
-	// Required for OAuth2 providers to resolve user identity.
-	// +kubebuilder:validation:Required
-	UserInfo *UserInfoConfig `json:"userInfo"`
+	// When omitted, the embedded auth server cannot resolve a real user identity for this
+	// upstream and instead synthesizes a stable subject from the access token. This is the
+	// correct shape for upstreams that expose no userinfo surface (e.g., MCP authorization
+	// servers per the MCP authorization spec, or any pure OAuth 2.0 server without a
+	// userinfo or introspection endpoint). In that mode, downstream JWTs receive a
+	// non-PII synthesized subject and no display name or email.
```
tgrunnagle: **[MEDIUM — Operator/CRD] #4 No status condition surfaces synthesis-mode to operators (consensus 7/10)**

When `userInfo` is omitted, neither `MCPExternalAuthConfigStatus` nor `VirtualMCPServerStatus` publishes any condition telling the operator they are running in fallback mode. Per `.claude/rules/operator.md` "Status Condition Parity", advisory conditions about misconfiguration or unsupported states should be surfaced consistently. There's precedent: `ConditionTypeAuthzUpstreamSelectionWarning` (`virtualmcpserver_types.go:283`) is described as exactly this kind of advisory.

**Recommendation**: add an advisory `ConditionTypeIdentitySynthesized` (or similar) on both `MCPExternalAuthConfigStatus` and `VirtualMCPServerStatus`, set when any OAuth2 upstream has nil `UserInfo`. Use `controllerutil.MutateAndPatchStatus` per operator rules.

*Raised by: Operator/CRD, Security*

pkg/authserver/upstream/oauth2.go
```diff
@@ -424,6 +440,29 @@ func (p *BaseOAuth2Provider) ExchangeCodeForIdentity(ctx context.Context, code,
...
+const synthesizedSubjectPrefix = "tk-"
+
+// synthesizeSubjectFromAccessToken returns a stable, opaque, non-secret
+// identifier derived from an opaque access token, for OAuth2 upstreams that
+// expose no userinfo endpoint.
+//
+// The output is the synthesizedSubjectPrefix concatenated with the lowercase
+// hex of the first 16 bytes of the SHA-256 digest of the access token (35
+// characters total, e.g. "tk-89abcdef0123456789abcdef01234567"). It is
+// deterministic for a given token, contains no PII, and reveals nothing about
```
tgrunnagle: **[LOW — Docs/Security] #5 "Reveals nothing about the token contents" claim is stronger than the algorithm guarantees (consensus 7/10)**

Two issues with the wording: (1) preimage-resistance does not prevent guessing-style attacks against tokens with low entropy — an attacker who observes the synthesized subject in a log/audit record and has any candidate token can verify equality offline. That is not "no information" — it confirms which token a session was bound to. (2) The helper accepts any string; for JWT access tokens, the source token still carries PII even though the digest does not. The helper's contract does not constrain inputs to opaque tokens.

```suggestion
// hex of the first 16 bytes of the SHA-256 digest of the access token (35
// characters total, e.g. "tk-89abcdef0123456789abcdef01234567"). It is
// deterministic for a given token. The output is non-PII assuming the
// upstream issues opaque (non-JWT) bearer tokens; the digest itself reveals
// nothing about the input beyond what an attacker who already has a candidate
// token could confirm by re-hashing. 16 bytes (128 bits) is sufficient
// collision resistance for a session-key role.
```

*Raised by: Security, Architecture, General code quality*

pkg/authserver/upstream/oauth2.go
```diff
@@ -400,14 +402,28 @@ func (p *BaseOAuth2Provider) buildAuthorizationURL(
...
 // ExchangeCodeForIdentity exchanges an authorization code for tokens and resolves
 // the user's identity in a single atomic operation.
-// For pure OAuth2 providers, identity is resolved via the UserInfo endpoint.
+// For pure OAuth2 providers, identity is resolved via the UserInfo endpoint when
+// configured. When UserInfo is nil — typically because the upstream authorization
+// server exposes no userinfo surface (e.g., MCP authorization servers per the MCP
+// authorization spec) — the subject is synthesized from a hash of the access token
+// via synthesizeSubjectFromAccessToken. Name and Email remain empty in that mode.
 // The nonce parameter is ignored for pure OAuth2 providers (no ID token validation).
 func (p *BaseOAuth2Provider) ExchangeCodeForIdentity(ctx context.Context, code, codeVerifier, _ string) (*Identity, error) {
```
tgrunnagle, line 419: **[LOW — Docs] #6 `Identity.Subject` doc in `types.go` now contradicts the synthesis path (consensus 7/10)**

Note for `pkg/authserver/upstream/types.go:31` (outside this diff but reachable from this PR): `Identity.Subject` is documented as "the canonical user identifier from the upstream IDP (the 'sub' claim)". After this PR, the synthesis branch produces a value that is neither from the upstream nor the `sub` claim — it's a SHA-256-derived local construct prefixed with `tk-`.

**Recommendation**: update the doc comment in `types.go` to acknowledge both sources: "For OIDC providers it is the `sub` claim from the ID token. For OAuth2 providers it is either the `sub` returned by the userinfo endpoint or — when no userinfo is configured — a synthesized opaque value with the `tk-` prefix derived from the access token." Surface the stability contract there too (rotates on re-auth, stable across token refresh).

*Raised by: Architecture*

pkg/authserver/upstream/oauth2_test.go
```diff
@@ -1198,6 +1201,103 @@ func TestBaseOAuth2Provider_ExchangeCodeForIdentity(t *testing.T) {
...
+		assert.Len(t, hexPart, 32, "first 16 bytes of SHA-256 in hex is 32 chars")
+		// Must be valid hex.
+		_, err := hex.DecodeString(hexPart)
+		assert.NoError(t, err)
+	})
+
+	t.Run("empty token still produces a stable subject", func(t *testing.T) {
+		t.Parallel()
+		// Defensive: caller should never pass empty, but if it does, the
+		// function must not panic and must remain deterministic.
```
tgrunnagle: **[LOW — Test Coverage] #7 "empty token" subtest doesn't verify the function's contract (consensus 8/10)**

The subtest is named "empty token still produces a stable subject" and the comment claims the function "must not panic and must remain deterministic". The test only asserts two calls with `""` are equal — trivially true for any deterministic pure function. A regression that short-circuits empty input to `""` (or to a non-prefixed value) would still pass.

```suggestion
		t.Run("empty token still produces a stable subject", func(t *testing.T) {
			t.Parallel()
			// Defensive: caller should never pass empty, but if it does, the
			// function must not panic and must remain deterministic.
			got1 := synthesizeSubjectFromAccessToken("")
			got2 := synthesizeSubjectFromAccessToken("")
			assert.Equal(t, got1, got2)
			assert.True(t, strings.HasPrefix(got1, synthesizedSubjectPrefix))
			assert.Len(t, strings.TrimPrefix(got1, synthesizedSubjectPrefix), 32)
		})
```

*Raised by: Test Coverage, General code quality*

pkg/authserver/upstream/oauth2.go
```diff
@@ -424,6 +440,29 @@ func (p *BaseOAuth2Provider) ExchangeCodeForIdentity(ctx context.Context, code,
...
+// The output is the synthesizedSubjectPrefix concatenated with the lowercase
+// hex of the first 16 bytes of the SHA-256 digest of the access token (35
+// characters total, e.g. "tk-89abcdef0123456789abcdef01234567"). It is
+// deterministic for a given token, contains no PII, and reveals nothing about
+// the token contents — SHA-256 is preimage-resistant, and 16 bytes (128 bits)
+// is sufficient collision resistance for a session-key role.
+//
+// This is only used for OAuth2 upstreams whose UserInfo is nil. OIDC providers
+// always have an ID token-derived subject and never reach this path.
+func synthesizeSubjectFromAccessToken(accessToken string) string {
```
tgrunnagle, line 484: **[LOW — Code Quality] #8 Empty access token produces a stable synthesized subject — no defensive guard (consensus 7/10)**

`synthesizeSubjectFromAccessToken("")` returns a fixed value (SHA-256 of empty string, prefixed). The caller path makes empty input unlikely (`exchangeCodeForTokens` rejects empty access tokens), but if a future bug or upstream misbehavior produced an empty token, every synthesis call would return the same constant — and `UserResolver.ResolveUser` would happily collapse all those sessions into one shared internal user. Per `.claude/rules/go-style.md` "Constructor Validation: Fail Loudly on Invalid Input".

**Recommendation**: either panic on empty input (it's a programming error from the only caller) or return an error and propagate. Treat reachability of the empty path as a bug, not as input the helper must absorb. Update the test in finding 7 accordingly.

*Raised by: Security, Architecture*

pkg/authserver/config.go
```diff
@@ -228,8 +228,11 @@ type OAuth2UpstreamRunConfig struct {
 	// Scopes are the OAuth scopes to request from the upstream IDP.
 	Scopes []string `json:"scopes,omitempty" yaml:"scopes,omitempty"`
 
-	// UserInfo contains configuration for fetching user information (required for OAuth2).
-	UserInfo *UserInfoRunConfig `json:"userinfo" yaml:"userinfo"`
+	// UserInfo contains configuration for fetching user information.
+	// Optional: when nil, the upstream provider synthesizes a stable subject
+	// from the access token instead of calling a userinfo endpoint. See
+	// BaseOAuth2Provider.ExchangeCodeForIdentity for the synthesis rule.
```
tgrunnagle: **[LOW — Docs] #9 Cross-package reference will silently rot (consensus 7/10)**

`OAuth2UpstreamRunConfig` lives in `pkg/authserver`, but `BaseOAuth2Provider.ExchangeCodeForIdentity` lives in the sibling `pkg/authserver/upstream` package. The reference is unqualified, and Go has no compile-time check that prose names exist — if the method is renamed or moved, this comment silently drifts (per `.claude/rules/go-style.md` "Keep Comments Synchronized With Code").

```suggestion
	// UserInfo contains configuration for fetching user information.
	// Optional: when nil, the upstream OAuth2 provider derives a deterministic
	// subject by SHA-256-hashing the access token (with a "tk-" prefix) instead
	// of calling a userinfo endpoint. OIDC providers always derive Subject from
	// the ID token and are unaffected.
	UserInfo *UserInfoRunConfig `json:"userinfo,omitempty" yaml:"userinfo,omitempty"`
```

*Raised by: General code quality*

tgrunnagle (commented):
**Agents consulted**: Security, OAuth/OIDC protocol, Operator/CRD, Test coverage, Architecture & Design, General code quality

### Consensus Summary
| 1 | Synthesis-mode re-auth creates a new internal user record per re-auth | 9/10 | HIGH | Discuss / fix |
| 2 | Implicit-when-nil synthesis activates silently on YAML typos / drift | 9/10 | HIGH | Discuss / fix |
| 3 | Tripwire test is decorative — `UserInfo` nil so URL never wired up | 8/10 | MEDIUM | Fix or remove |
| 4 | No CR status condition surfaces synthesis-mode to operators | 7/10 | MEDIUM | Add condition |
| 5 | Helper doc "reveals nothing about the token contents" too strong | 7/10 | LOW | Soften wording |
| 6 | `Identity.Subject` doc (in `types.go`) now contradicts synthesis path | 7/10 | LOW | Update doc |
| 7 | "empty token" subtest only asserts equality — no shape verification | 8/10 | LOW | Strengthen |
| 8 | `synthesizeSubjectFromAccessToken("")` returns deterministic value | 7/10 | LOW | Reject empty |
| 9 | `config.go` cross-package reference `BaseOAuth2Provider.ExchangeCodeForIdentity` will rot | 7/10 | LOW | Inline or qualify |
| 10 | `synthesizedSubjectPrefix` package-private but visible in JWTs/logs | 7/10 | LOW | Export |

### Overall

The PR is well-scoped, well-tested for the happy path, and the synthesis approach is the right call given the constraint (Atlassian's MCP authorization server has no userinfo, no introspection, and issues opaque tokens). SHA-256 + 16-byte truncation + a `tk-` prefix is a sound construction for a session-key role; OIDC providers correctly remain unaffected via `OIDCProviderImpl.ExchangeCodeForIdentity` overriding the base method.

The two HIGH findings are architectural decisions worth deciding consciously before this lands, not blocking-correctness defects. (1) The synthesized subject rotates on re-auth, which means `UserResolver.ResolveUser` (`pkg/authserver/server/handlers/user.go:55`) creates a brand-new internal `User` row plus a new `ProviderIdentity` row on every re-authentication of the same human — the storage growth is monotonic and per-user state fragments across N internal IDs. (2) The synthesis path activates on any nil `UserInfo`, including a YAML typo or drift that an operator did not intend; there is no startup warning, no status condition, and no explicit opt-in, so silent regression into synthesis mode is possible. Refresh-token rotation is already handled correctly by `refresher.go` preserving the original `UpstreamSubject`, so the issue is strictly about new auth flows.

The MEDIUM finding on the tripwire test is a real test-quality issue worth fixing in this PR — with `UserInfo` nil, the production code returns at line 420 before any HTTP path could touch the tripwire, so the `userinfoHits == 0` assertion is structurally guaranteed. Either drop the tripwire and rely on the existing `Name`/`Email` empty assertions, or split into a paired test that wires the tripwire URL into a configured `UserInfoConfig` and asserts userinfo IS called when configured.

The LOW findings are polish: doc comments that need to track the new branch (`Identity.Subject` in `types.go`, the helper's PII claim, the cross-package reference in `config.go`), an empty-token subtest that doesn't actually verify shape, a defensive guard for the empty-token edge case, and exporting `synthesizedSubjectPrefix` so downstream tooling can recognize synthesized subjects without hardcoding a magic string.

### Documentation
Doc updates that would land outside this diff but are part of the consensus:
- `pkg/authserver/upstream/types.go:31` — `Identity.Subject` doc no longer matches the synthesis path (finding 6).
- `docs/arch/` — surface synthesis-mode storage growth as an architectural choice (finding 1).

---

pkg/authserver/upstream/oauth2.go
```diff
@@ -424,6 +440,29 @@ func (p *BaseOAuth2Provider) ExchangeCodeForIdentity(ctx context.Context, code,
 	}, nil
 }
 
+// synthesizedSubjectPrefix tags subjects produced by
+// synthesizeSubjectFromAccessToken so they are visually distinguishable from
+// real upstream-provided subjects in logs, audit records, and JWTs.
+const synthesizedSubjectPrefix = "tk-"
```
tgrunnagle, line 454: **[LOW — Architecture] #10 `synthesizedSubjectPrefix` is package-private but part of the externally observable contract (consensus 7/10)**

The `tk-` prefix is intended to be visible in JWTs, logs, and audit records (per the doc comment), so it's part of the externally observable contract of the auth server. Downstream code (audit pipelines, the proxy runner doing per-request logging, debug tooling, and the status condition recommended in finding 4) cannot recognize a synthesized subject without hardcoding the magic string `"tk-"`.

**Recommendation**: export as `SynthesizedSubjectPrefix`, or expose a small predicate `IsSynthesizedSubject(string) bool`. This documents the prefix as part of the package's stable surface and makes it easier to reason about backwards compatibility if the format ever needs to change. Note that the test in `oauth2_test.go` would need its references updated to the new name.

*Raised by: Operator/CRD, Architecture*

pkg/authserver/upstream/oauth2.go
```diff
@@ -400,14 +410,27 @@ func (p *BaseOAuth2Provider) buildAuthorizationURL(
...
 	tokens, err := p.exchangeCodeForTokens(ctx, code, codeVerifier)
 	if err != nil {
 		return nil, err
 	}
 
+	// No userinfo: synthesize a non-PII subject from the access token.
+	// Synthetic=true tells the callback handler to bypass UserResolver — the
+	// synthesized subject rotates per access token, so persisting it would
+	// create a new `users` row on every re-authentication.
+	if p.config.UserInfo == nil {
```
jhrozek, line 429: **[HIGH — Security/Correctness] Empty access token produces a constant, cross-tenant synthesized subject**

Follow-on to resolved thread #3156683898 with a stronger angle than the original LOW framing.

`synthesizeSubjectFromAccessToken("")` is deterministic — `sha256.Sum256([]byte(""))` is the well-known constant `e3b0c44298fc1c14...`, so an empty access token always yields the same synthesized subject (`tk-e3b0c44298fc1c14...`) regardless of provider, tenant, or session.

The cross-tenant angle that the original comment did not surface: with the `UserResolver` bypass introduced in this PR, the synthesized value flows directly into `tok.UserID` for the upstream-token storage row (callback.go around `subject = result.Subject` for synthetic identities). If an OAuth2 upstream ever returns 200 with an empty `access_token` field (some misconfigured ASes do this on certain grant flows; custom token-response mappings can drop the field), every affected session **collapses into one storage bucket** — cross-tenant cross-user state mixing on a single Redis key.

The current test at `pkg/authserver/upstream/oauth2_test.go` (`empty token still produces a stable subject`) explicitly accepts this behavior. That conflicts with `.claude/rules/go-style.md` "Constructor Validation: Fail Loudly on Invalid Input".

Suggested fix at this branch:

```go
if p.config.UserInfo == nil {
    if tokens.AccessToken == "" {
        return nil, fmt.Errorf("%w: empty access token, cannot synthesize subject", ErrIdentityResolutionFailed)
    }
    return &Identity{
        Tokens:    tokens,
        Subject:   synthesizeSubjectFromAccessToken(tokens.AccessToken),
        Synthetic: true,
    }, nil
}
```

And invert the empty-token subtest to assert the error path.

pkg/authserver/upstream/oauth2.go
```diff
@@ -331,6 +333,14 @@ func NewOAuth2Provider(config *OAuth2Config, opts ...OAuth2ProviderOption) (*Bas
 		"token_endpoint", config.TokenEndpoint,
 	)
 
+	if config.UserInfo == nil {
+		// Surface synthesis mode at construction so operators see it once
+		// per provider rather than only inferring from missing claims later.
+		slog.Info("oauth2 upstream has no userinfo configured; using synthesis mode " +
```
jhrozek: **[Logging — Convention] `slog.Info` should be `slog.Warn` per `.claude/rules/go-style.md`**

`.claude/rules/go-style.md` Logging section:

> - INFO sparingly — only for long-running operations like image pulls
> - WARN for non-fatal issues (deprecations, fallback behavior, cleanup failures)

Synthesis mode is exactly "fallback behavior" — the auth server is degrading from real-identity resolution to a hashed-token surrogate because the operator did not configure a userinfo endpoint. The commit message describes this as "WARN at provider construction" but the code uses `slog.Info`.

Two improvements together:

1. Switch to `slog.Warn`.
2. Add `token_endpoint` as a structured field — operators running a multi-upstream auth server need to identify which provider the warning is about, and the unstructured prose string today provides no way to filter or correlate.

Verified call frequency: `NewOAuth2Provider` is invoked from `defaultUpstreamFactory` (`pkg/authserver/server_impl.go:49`) at server construction. Once per provider per pod lifetime — safe at WARN, no log spam.

```go
slog.Warn("oauth2 upstream has no userinfo configured; using synthesis mode "+
    "(synthesized non-PII subject from access token; no Name/Email claims). "+
    "If a userinfo endpoint exists for this upstream, configure it.",
    "token_endpoint", config.TokenEndpoint,
)
```

cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
```diff
@@ -109,6 +109,13 @@ func (r *MCPExternalAuthConfigReconciler) Reconcile(ctx context.Context, req ctr
 		ObservedGeneration: externalAuthConfig.Generation,
 	})
 
+	// Set or clear the IdentitySynthesized advisory. Either-direction
+	// transitions count as a condition change so the existing Status().Update
+	// path picks them up.
+	if r.applyIdentitySynthesizedCondition(externalAuthConfig) {
```
jhrozek: **[Operator — Status Lifecycle] Stale `IdentitySynthesized` lingers when validation fails**

The new advisory is set here on the validation-success path, but multiple early-return branches skip this block:

- This file at the early-return for `if err := externalAuthConfig.Validate(); err != nil`: returns `ctrl.Result{}, nil` after setting `Valid=False`, never reaching `applyIdentitySynthesizedCondition`.
- Parallel surface in `cmd/thv-operator/controllers/virtualmcpserver_controller.go`: `validateAuthServerConfig` early-returns on `Issuer == ""`, `len(UpstreamProviders) == 0`, and `ValidateAdditionalAuthorizationParams` failure, all before the new advisory block.

Concrete failure mode:

1. Operator deploys with a synthesizing upstream → `IdentitySynthesized=True`.
2. Operator edits the spec and breaks it (e.g. removes `tokenEndpoint`).
3. Reconciler sets `Valid=False` and returns from the validation-failure branch, skipping the advisory block.
4. `kubectl describe` now shows `Valid=False (ValidationFailed)` **AND** `IdentitySynthesized=True (OAuth2UpstreamWithoutUserInfo)` referencing an upstream that the broken spec no longer mentions.

Two fixes:

- **Option A (preferred)**: Move the advisory computation **above** the validation guards in both reconcilers. The advisory is a pure function of `cfg.Spec.EmbeddedAuthServer.UpstreamProviders` field shape — it does not depend on issuer URL validity or other `Validate()` concerns.
- **Option B**: Explicitly clear the condition on every early-return path. More mechanical and easy to forget in a future early-exit branch.

A transition test pins the fix: reconcile #1 with synthesis upstream → assert True; mutate spec to break validation; reconcile #2 → assert condition cleared (or matches the new spec).

pkg/authserver/upstream/types.go
```diff
@@ -28,14 +28,31 @@ type Identity struct {
...
 
 	// Email is the user's email address from the upstream IDP (optional).
 	Email string
+
+	// Synthetic is true when Subject was generated locally because the
+	// upstream has no userinfo or ID-token-derived identity. Synthetic
+	// subjects rotate per re-auth; callers MUST NOT persist them as stable
+	// per-user keys (doing so grows `users` without bound). Use the
+	// synthesized Subject as an ephemeral session key only.
+	Synthetic bool
```
jhrozek, line 55: **[Documentation] Synthesis-mode behavior should be reflected in `docs/arch/`**

Per `CLAUDE.md`: "When making changes that affect architecture, update relevant docs in `docs/arch/`. See `docs/arch/README.md` for structure."

This PR introduces a new identity-resolution mode and changes downstream contracts that operators encounter directly:

1. `tok.UserID` in upstream-token storage may now be a `tk-…` synthesized value rather than a stable UUID.
2. `UserResolver` is bypassed for synthetic identities (no `users` row, no `provider_identities` row, no `UpdateLastAuthenticated` call).
3. The Redis `KeyTypeUserUpstream` reverse-index assumption (one set per stable user) is inverted under synthesis (one set per session per re-auth) — see `pkg/authserver/storage/redis.go:43-45` existing TODO.

Two minimal additions:

- **`docs/arch/11-auth-server-storage.md`** — add a "Synthesis-mode subjects" subsection in the `UserStorage` / `UpstreamTokenStorage` section. Cover when the path triggers, the `tk-` userID format, the `UserResolver` bypass, and the reverse-index implication.
- **`docs/arch/10-virtual-mcp-architecture.md`** — one paragraph in "Incoming Authentication" cross-referencing the storage doc, so an operator searching the architecture index for the new behavior finds it.

Comparison context worth including in the docs: ToolHive's synthesis-from-access-token fallback is **more permissive** than peer projects. Ory Hydra requires the login provider to supply a `subject` (no synthesis at all). `identity-api` derives subject from `iss + sub` claims (assumes parseable structured tokens). oauth2-proxy fails closed when neither ID token nor userinfo provides identity. ToolHive accepting an opaque token is a deliberate choice, but operators arriving from any of those projects will not expect this fallback to exist unless docs call it out.

jhrozek (commented):
Follow-on review after the second pass of commits. Three net-new findings (B, C, D) plus one escalation of resolved thread #3156683898 (A) where the original LOW framing did not surface a cross-tenant collision angle that became material once the `UserResolver` bypass landed.

Nothing here is a hard blocker — A is the most concrete correctness concern, B and C are convention/lifecycle gaps with simple fixes, and D is a docs sync.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -365,9 +365,14 @@ type OAuth2UpstreamConfig struct {
...
+	// +optional
+	UserInfo *UserInfoConfig `json:"userInfo,omitempty"`
```
jhrozek, line 375: just a note - we should probably have a blob in the docs repo - maybe worth a follow up?

jhrozek (approved):
Looks good!

The only action item is to file a follow-up in the docs repo so this is discoverable and known - the admins should know the implications of not setting UserInfo.

PR #5096, opened 2026-04-29
fix(authserver): send scope explicitly on upstream token refresh

dallinstevens:
## Summary

The upstream refresh path uses `golang.org/x/oauth2`'s `TokenSource`, which omits `scope` on refresh requests. RFC 6749 §6 says the param is optional and the AS SHOULD preserve original scopes — but Entra ID's v1 token issuer (and likely others) silently narrows refreshed tokens to the user's default consent set (e.g. `User.Read`), dropping any custom resource scope the original token had.

This sends `scope` explicitly on every refresh.

## Repro

Embedded auth + Entra upstream + a custom resource scope (we hit it with `api://<app-id>/session:role-any` for Snowflake's external OAuth integration):

1. Initial code → token: `scp` includes the custom scope ✅ — works for ~1h
2. Token refresh: `scp` reduced to `User.Read` — bearer-valid but missing the scope downstream resources enforce
3. Snowflake rejects every subsequent query with *"role 'X' is not listed in the Access Token or was filtered"*
4. User must manually re-auth to recover

We confirmed via decoded token claims: `aud`, `iss`, `appid` all stay correct on refresh — only `scp` is wrong. Sending `scope=` explicitly on refresh produces tokens with the original scope set preserved.

## What changed

`pkg/authserver/upstream/oauth2.go::BaseOAuth2Provider.RefreshTokens` — replaces `TokenSource` with a direct POST to the token endpoint so `scope` can be set. Standard library doesn't expose a way to inject scope on `TokenSource`-driven refreshes.

Preserves all existing behavior: client_secret semantics, error shape, response rewriting via `wrapHTTPClientWithMapping`, OIDC subject validation in the layer above.

Two extras worth noting:
- Caps `expires_in` at `math.MaxInt32` to match `golang.org/x/oauth2` (avoids time arithmetic overflow when an emitter returns nanoseconds — `oauth2-proxy/mockoidc` does this in tests).
- Preserves the original refresh token if the AS doesn't issue a new one (RFC 6749 §6).

## Tests

New:
- refresh request includes the configured scope set
- original refresh token preserved when AS omits one

Existing 4 RefreshTokens tests pass unchanged. Integration test against `mockoidc` passes.

`go test ./pkg/authserver/...`, `go build ./...`, `go vet` all clean.

## Verified in production

Forked image deployed to a live Entra-backed embedded-auth flow against Snowflake. Before fix: refreshed tokens carried `scp: User.Read`, queries failed every ~1h. After fix: refreshed tokens carry `scp: session:role-any`, queries succeed indefinitely across refresh boundaries.

### DIFF
--- a/pkg/authserver/upstream/oauth2.go
+++ b/pkg/authserver/upstream/oauth2.go
@@ -28,7 +28,6 @@ import (
 	"net/url"
 	"os"
 	"strings"
-	"time"
 
 	"golang.org/x/oauth2"
 
@@ -456,6 +455,16 @@ func (p *BaseOAuth2Provider) exchangeCodeForTokens(ctx context.Context, code, co
 }
 
 // RefreshTokens refreshes the upstream IDP tokens.
+//
+// Sends `scope` explicitly. RFC 6749 §6 makes the param optional and says the
+// AS SHOULD preserve original scopes on omission, but some ASes (notably
+// Entra ID v1) silently narrow refreshed tokens to the user's default consent
+// set when `scope` is omitted, dropping custom resource scopes.
+//
+// Uses oauth2.Config.Exchange with SetAuthURLParam overrides (mirroring the
+// pattern in pkg/auth/oauth/non_caching_refresher.go) because the standard
+// library's TokenSource refresh path doesn't expose a way to inject scope.
+// The empty code= side-effect is tolerated — ASes dispatch on grant_type first.
 func (p *BaseOAuth2Provider) RefreshTokens(ctx context.Context, refreshToken, _ string) (*Tokens, error) {
 	if refreshToken == "" {
 		return nil, errors.New("refresh token is required")
@@ -463,21 +472,22 @@ func (p *BaseOAuth2Provider) RefreshTokens(ctx context.Context, refreshToken, _ 
 
 	slog.Info("refreshing tokens",
 		"token_endpoint", p.config.TokenEndpoint,
+		"scope_count", len(p.oauth2Config.Scopes),
 	)
 
 	// Wrap HTTP client with token response rewriter if mapping is configured.
 	httpClient := wrapHTTPClientWithMapping(p.httpClient, p.config.TokenResponseMapping, p.config.TokenEndpoint)
 	ctx = context.WithValue(ctx, oauth2.HTTPClient, httpClient)
 
-	// Create an expired token with the refresh token to trigger refresh
-	expiredToken := &oauth2.Token{
-		RefreshToken: refreshToken,
-		Expiry:       time.Now().Add(-time.Hour), // Expired token forces refresh
+	opts := []oauth2.AuthCodeOption{
+		oauth2.SetAuthURLParam("grant_type", "refresh_token"),
+		oauth2.SetAuthURLParam("refresh_token", refreshToken),
+	}
+	if len(p.oauth2Config.Scopes) > 0 {
+		opts = append(opts, oauth2.SetAuthURLParam("scope", strings.Join(p.oauth2Config.Scopes, " ")))
 	}
 
-	// Use TokenSource to get a new token via refresh
-	tokenSource := p.oauth2Config.TokenSource(ctx, expiredToken)
-	token, err := tokenSource.Token()
+	token, err := p.oauth2Config.Exchange(ctx, "", opts...)
 	if err != nil {
 		return nil, formatOAuth2Error(err, "token request failed")
 	}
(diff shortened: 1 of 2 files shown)


pkg/authserver/upstream/oauth2.go
```diff
@@ -466,39 +467,108 @@ func (p *BaseOAuth2Provider) exchangeCodeForTokens(ctx context.Context, code, co
...
+		"client_id":     {p.oauth2Config.ClientID},
+	}
+	if p.oauth2Config.ClientSecret != "" {
+		form.Set("client_secret", p.oauth2Config.ClientSecret)
+	}
+	if len(p.oauth2Config.Scopes) > 0 {
+		form.Set("scope", strings.Join(p.oauth2Config.Scopes, " "))
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, p.config.TokenEndpoint, strings.NewReader(form.Encode()))
```
jhrozek: Have you tried `oauth2.Config.Exchange(ctx, "", opts...)` with `SetAuthURLParam` overrides instead of a direct POST? `pkg/auth/oauth/non_caching_refresher.go:117-131` already uses this pattern for the resource-indicator refresh path — `SetAuthURLParam` lets you override `grant_type`, `refresh_token`, and `scope` on the form body while reusing the lib's POST plumbing.

Sketch:

```go
opts := []oauth2.AuthCodeOption{
    oauth2.SetAuthURLParam("grant_type", "refresh_token"),
    oauth2.SetAuthURLParam("refresh_token", refreshToken),
}
if len(p.oauth2Config.Scopes) > 0 {
    opts = append(opts, oauth2.SetAuthURLParam("scope", strings.Join(p.oauth2Config.Scopes, " ")))
}
token, err := p.oauth2Config.Exchange(ctx, "", opts...)
if err != nil {
    return nil, formatOAuth2Error(err, "token request failed")
}
tokens, err := convertOAuth2Token(token)
// §6 fallback + return
```

Upsides: `convertOAuth2Token` stays the single token-shape parser, the `expires_in` MaxInt32 cap drops out (the lib already does it), `wrapHTTPClientWithMapping` still applies via the RoundTripper, and `formatOAuth2Error` already produces the `token request failed: <code>: <description>` shape from `oauth2.RetrieveError`.

The empty `code=""` is sent — same caveat as `non_caching_refresher` — and is tolerated because ASes dispatch on `grant_type` first.

Both new tests should still pass: the scope test inspects the form body (Exchange still posts a form) and the refresh-token-preservation test inspects the returned `Tokens.RefreshToken` (your §6 fallback line stays).

Not a blocker — curious if you considered this and chose direct POST for a reason I'm missing.

jhrozek (commented):
One non-blocking suggestion on the new refresh path — see inline comment.

pkg/authserver/upstream/oauth2.go
jhrozek: @dallinstevens to ground the suggestion - it is not blocking and I'll be happy to merge this PR and refactor myself, seeing that you needed the patch to unblock an issue for you. Just let me know your preference.

jhrozek:
Either way looks like the linter is unhappy - I believe going the `oauth2.Config.Exchange(ctx, "", opts...)` route would fix the lint issues by not introducing that code in the first place.

dallinstevens:
@jhrozek thanks — that pattern is much cleaner. Refactored to use `oauth2.Config.Exchange(ctx, "", opts...)` with `SetAuthURLParam` overrides, mirroring `non_caching_refresher.go::refreshWithResource`.

jhrozek (dismissed):
LGTM. Sending scope explicitly is RFC 6749 §6 compliant ("MUST NOT include any scope not originally granted") and the Entra v1 narrowing behavior is a known interop wart. The pattern mirrors what's already in production at pkg/auth/oauth/non_caching_refresher.go::refreshWithResource, so this is consistent with prior art.

Verified the error-shape contract still holds — Exchange and TokenSource both route through the same retrieveToken helper, so *oauth2.RetrieveError surfaces identically and formatOAuth2Error works unchanged. The OIDC override at oidc.go:387 is also unaffected; ID token validation still runs on top.

jhrozek:
@dallinstevens oh I'm sorry we managed to push another commit and there is a conflict now, would you mind rebasing? Sorry!

dallinstevens:
@jhrozek No worries! Rebased and resolved conflicts. Lmk if that still looks good

jhrozek (approved):

jhrozek:
Merged!

Thank you very much for the contribution @dallinstevens 🚀

PR #5104, opened 2026-04-29
Preserve user fields in EmbeddingServer podTemplateSpec merge

### LINKED ISSUE #5100: EmbeddingServer drops imagePullSecrets and other fields from podTemplateSpec
## Bug description

`EmbeddingServer.spec.podTemplateSpec` is documented as the customization mechanism for the spawned StatefulSet pod (`cmd/thv-operator/api/v1beta1/embeddingserver_types.go:91-98`). Validation passes for any well-formed PodTemplateSpec.

However, the controller's merge function only copies a hardcoded subset of fields:

```go
// cmd/thv-operator/controllers/embeddingserver_controller.go:662-700
func (r *EmbeddingServerReconciler) mergePodTemplateSpec(...) {
    ...
    if userTemplate.Spec.NodeSelector != nil { ... }
    if userTemplate.Spec.Affinity != nil { ... }
    if len(userTemplate.Spec.Tolerations) > 0 { ... }
    if userTemplate.Spec.SecurityContext != nil { ... }
    if userTemplate.Spec.ServiceAccountName != "" { ... }
    // + container SecurityContext only
}
```

Anything else the user puts in `podTemplateSpec.spec` — including `imagePullSecrets`, `volumes`, `volumeMounts` on user-added containers, `priorityClassName`, `topologySpreadConstraints`, `runtimeClassName`, etc. — is silently dropped. Validation succeeds and the user has no signal that their config was ignored.

This is comparable to the closed issue #3436 (MCPServer's equivalent gap), but for EmbeddingServer.

For reference, MCPRegistry handles this correctly by starting from `userPTS.DeepCopy()` and merging defaults in (`cmd/thv-operator/pkg/registryapi/podtemplatespec.go:404`), and VirtualMCPServer handles it via strategic merge patch on raw JSON (`cmd/thv-operator/controllers/virtualmcpserver_deployment.go:929`).

## Steps to reproduce

1. Create an `EmbeddingServer` whose image lives in a private registry.
2. Set:
   ```yaml
   spec:
     podTemplateSpec:
       spec:
         imagePullSecrets:
           - name: regcred
   ```
3. The CR is accepted; the StatefulSet's PodSpec contains no `imagePullSecrets`.
4. The pod fails with `ImagePullBackOff`.

## Expected behavior

`spec.podTemplateSpec` should preserve user-provided fields the same way MCPRegistry and VirtualMCPServer do.

## Suggested fix

Replace the field-by-field copy in `mergePodTemplateSpec` with either:
- a deep-copy-and-merge approach (as in `MergePodTemplateSpecs` in `pkg/registryapi/podtemplatespec.go`), or
- a strategic merge patch on the raw user JSON (as in `applyPodTemplateSpecToDeployment` for VirtualMCPServer).

The fix should preserve *all* user-provided pod-spec fields, not just the currently enumerated five. Add a regression test that verifies `imagePullSecrets` set via `podTemplateSpec` reach the resulting StatefulSet pods.

JAORMX:
## Summary

The previous merge function in the EmbeddingServer controller copied
only an enumerated subset of `PodSpec` fields — `NodeSelector`,
`Affinity`, `Tolerations`, `SecurityContext`, `ServiceAccountName`,
and the embedding container's `SecurityContext`. Anything else the
user supplied on `spec.podTemplateSpec.spec` (`imagePullSecrets`,
additional `volumes`, `priorityClassName`, `topologySpreadConstraints`,
`runtimeClassName`, init containers, sidecars, …) was silently dropped.
Validation still passed, so users had no signal that the operator was
ignoring half of their pod template.

This PR replaces the enumerated copy with a strategic merge patch on
the raw user JSON, mirroring the approach already in use in
`VirtualMCPServer`'s deployment builder. The patch preserves every
user-supplied `PodSpec` field while controller defaults remain where
the user did not set a value. Using the raw user JSON (rather than a
re-marshaled struct) avoids the nil-slice-becomes-empty-array problem
that strategic merge patch interprets as "replace with empty".

The old `mergePodTemplateSpec` and `mergeContainerSecurityContext`
methods are removed so there is one merge code path, not two.

Per review feedback, the marshal → `StrategicMergePatch` → unmarshal
sequence has been extracted into a new helper
`ctrlutil.ApplyPodTemplateSpecPatch` in
`cmd/thv-operator/pkg/controllerutil/podtemplatespec_patch.go` and is
now consumed by both the EmbeddingServer (this PR's call site) and
the VirtualMCPServer controller. The helper is policy-neutral: it
returns `(corev1.PodTemplateSpec, error)` and lets each caller pick
its error-handling policy. The two callers keep their existing
divergent behavior — vMCP hard-fails (blocks Deployment creation),
EmbeddingServer soft-fails (logs and falls back to controller
defaults). Both decisions are documented at their call sites; the
helper's godoc explains the trade-off and points at both consumers.

Fixes #5100

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Operator integration tests (`ginkgo ./cmd/thv-operator/test-integration/embedding-server/...` — 80/80 pass with `KUBEBUILDER_ASSETS` from `setup-envtest`)

Added regression tests:

- `TestEmbeddingServer_PodTemplateSpec_PreservesUserFields` (table-driven) — reconciles an EmbeddingServer with each previously-dropped field and asserts the resulting StatefulSet's pod template carries it through. Cases covered: `imagePullSecrets`, `priorityClassName`, extra `volumes`, `runtimeClassName`, `topologySpreadConstraints`, and a user-provided sidecar container alongside the controller-generated embedding container.
- `TestEmbeddingServer_PodTemplateSpec_SoftFailFallback` — uses a payload that the validation pass accepts but `StrategicMergePatch` rejects (an unknown `$patch` directive nested inside a container). Verifies the StatefulSet is still created with controller defaults — the soft-fail contract.
- `TestEmbeddingServer_PodTemplateSpec_EmptyObjectIsNoOp` — verifies a `{}` patch is a no-op: every controller-generated field on the embedding container (image, probes, ports, args) survives unchanged.
- `TestApplyPodTemplateSpecPatch` (new helper unit tests) — covers nil/empty/`{}` inputs, a user-fields merge, malformed JSON, and a type-mismatched patch.

## API Compatibility

Yes. Users of `EmbeddingServer` whose `spec.podTemplateSpec` includes
fields outside the previously-enumerated subset (most notably
`imagePullSecrets`) will now see those fields applied to the pod that
runs the embedding model. Existing manifests that only used the
enumerated fields are unaffected.

- The new helper lives at `cmd/thv-operator/pkg/controllerutil/podtemplatespec_patch.go` next to the existing `podtemplatespec_builder.go`.
- The `EmbeddingServer` `applyPodTemplateSpecToStatefulSet` no longer returns an error (the linter caught that the soft-fail path made the return value dead). Updated the call site accordingly.
- Reviewer flagged a pre-existing gap in `PodTemplateSpecBuilder.isEmpty()` that affects vMCP (now invisible on the EmbeddingServer path because the early-exit on `builder.Build() == nil` was already removed in the previous commit on this branch). Filed as #5110 — separate from this PR's scope.

### DIFF
--- a/cmd/thv-operator/controllers/embeddingserver_controller.go
+++ b/cmd/thv-operator/controllers/embeddingserver_controller.go
@@ -7,6 +7,7 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"maps"
 	"reflect"
@@ -21,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/tools/events"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -396,7 +398,7 @@ func (r *EmbeddingServerReconciler) validateAndUpdatePodTemplateStatus(
 
 // statefulSetForEmbedding creates a StatefulSet for the embedding server
 func (r *EmbeddingServerReconciler) statefulSetForEmbedding(
-	_ context.Context,
+	ctx context.Context,
 	embedding *mcpv1beta1.EmbeddingServer,
 ) *appsv1.StatefulSet {
 	replicas := embedding.GetReplicas()
@@ -436,6 +438,14 @@ func (r *EmbeddingServerReconciler) statefulSetForEmbedding(
 	// Add volumeClaimTemplates if model caching is enabled
 	if embedding.IsModelCacheEnabled() {
 		statefulSet.Spec.VolumeClaimTemplates = r.buildVolumeClaimTemplates(embedding)
+	}
+
+	// Apply user-provided PodTemplateSpec customizations via strategic merge patch.
+	// This must happen after the controller-generated template is fully populated so
+	// that user fields override controller defaults rather than the other way around.
+	if err := r.applyPodTemplateSpecToStatefulSet(ctx, embedding, statefulSet); err != nil {
+		log.FromContext(ctx).Error(err, "Failed to apply PodTemplateSpec to StatefulSet")
+		return nil
 	}
 
 	if err := ctrl.SetControllerReference(embedding, statefulSet, r.Scheme); err != nil {
@@ -634,13 +644,17 @@ func (*EmbeddingServerReconciler) applyResourceRequirements(embedding *mcpv1beta
 	}
 }
 
-// buildPodTemplate builds the pod template for the statefulset
-func (r *EmbeddingServerReconciler) buildPodTemplate(
-	embedding *mcpv1beta1.EmbeddingServer,
+// buildPodTemplate builds the pod template for the statefulset.
+// User-provided PodTemplateSpec customizations are applied later in
+// statefulSetForEmbedding via strategic merge patch.
+func (*EmbeddingServerReconciler) buildPodTemplate(
+	_ *mcpv1beta1.EmbeddingServer,
 	labels map[string]string,
 	container corev1.Container,
 ) corev1.PodTemplateSpec {
-	podTemplate := corev1.PodTemplateSpec{
+	// Note: Volumes for model cache are managed by StatefulSet volumeClaimTemplates
+	// and will be automatically mounted with the name "model-cache"
+	return corev1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels: labels,
 		},
@@ -648,73 +662,60 @@ func (r *EmbeddingServerReconciler) buildPodTemplate(
 			Containers: []corev1.Container{container},
 		},
 	}
-
-	// Note: Volumes for model cache are managed by StatefulSet volumeClaimTemplates
-	// and will be automatically mounted with the name "model-cache"
-
-	// Merge with user-provided PodTemplateSpec if specified
-	r.mergePodTemplateSpec(embedding, &podTemplate)
-
-	return podTemplate
 }
 
-// mergePodTemplateSpec merges user-provided PodTemplateSpec customizations
-func (r *EmbeddingServerReconciler) mergePodTemplateSpec(
+// applyPodTemplateSpecToStatefulSet applies user-provided PodTemplateSpec customizations
+// to the StatefulSet's pod template using strategic merge patch. This preserves every
+// user-supplied PodSpec field (imagePullSecrets, additional volumes, priorityClassName,
+// topologySpreadConstraints, init containers, sidecars, etc.) while keeping controller
+// defaults for fields the user did not set.
+//
+// The raw user JSON is used as the patch input (rather than re-marshaling a parsed
+// PodTemplateSpec). This avoids Go's `json.Marshal` turning nil slices into `[]`, which
+// strategic merge patch interprets as "replace with empty" and would clobber controller
+// defaults.
+func (*EmbeddingServerReconciler) applyPodTemplateSpecToStatefulSet(
+	ctx context.Context,
 	embedding *mcpv1beta1.EmbeddingServer,
-	podTemplate *corev1.PodTemplateSpec,
-) {
-	if embedding.Spec.PodTemplateSpec == nil {
-		return
+	statefulSet *appsv1.StatefulSet,
+) error {
+	if embedding.Spec.PodTemplateSpec == nil || len(embedding.Spec.PodTemplateSpec.Raw) == 0 {
+		return nil
 	}
 
-	builder, err := ctrlutil.NewPodTemplateSpecBuilder(embedding.Spec.PodTemplateSpec, embeddingContainerName)
+	// Validate the user-provided PodTemplateSpec is well-formed JSON.
+	// We don't check builder.Build() == nil for "empty" customizations: that helper
+	// only enumerates a subset of PodSpec fields and would skip the patch for
+	// fields like runtimeClassName or topologySpreadConstraints. Strategic merge
+	// patch is a no-op for `{}` anyway, so always running it is safe.
+	if _, err := ctrlutil.NewPodTemplateSpecBuilder(embedding.Spec.PodTemplateSpec, embeddingContainerName); err != nil {
+		return fmt.Errorf("failed to build PodTemplateSpec: %w", err)
+	}
+
+	userJSON := embedding.Spec.PodTemplateSpec.Raw
+
+	originalJSON, err := json.Marshal(statefulSet.Spec.Template)
 	if err != nil {
-		return
+		return fmt.Errorf("failed to marshal original PodTemplateSpec: %w", err)
 	}
 
-	userTemplate := builder.Build()
-	if userTemplate == nil {
-		return
+	patchedJSON, err := strategicpatch.StrategicMergePatch(originalJSON, userJSON, corev1.PodTemplateSpec{})
+	if err != nil {
+		return fmt.Errorf("failed to apply strategic merge patch: %w", err)
 	}
 
-	// Merge user customizations into base pod template
-	if userTemplate.Spec.NodeSelector != nil {
-		podTemplate.Spec.NodeSelector = userTemplate.Spec.NodeSelector
-	}
-	if userTemplate.Spec.Affinity != nil {
-		podTemplate.Spec.Affinity = userTemplate.Spec.Affinity
-	}
-	if len(userTemplate.Spec.Tolerations) > 0 {
-		podTemplate.Spec.Tolerations = userTemplate.Spec.Tolerations
-	}
-	if userTemplate.Spec.SecurityContext != nil {
-		podTemplate.Spec.SecurityContext = userTemplate.Spec.SecurityContext
-	}
-	if userTemplate.Spec.ServiceAccountName != "" {
-		podTemplate.Spec.ServiceAccountName = userTemplate.Spec.ServiceAccountName
+	var patchedPodTemplateSpec corev1.PodTemplateSpec
+	if err := json.Unmarshal(patchedJSON, &patchedPodTemplateSpec); err != nil {
+		return fmt.Errorf("failed to unmarshal patched PodTemplateSpec: %w", err)
 	}
 
-	// Merge container-level customizations
-	r.mergeContainerSecurityContext(podTemplate, userTemplate)
-}
+	statefulSet.Spec.Template = patchedPodTemplateSpec
 
-// mergeContainerSecurityContext merges container-level security context
-func (*EmbeddingServerReconciler) mergeContainerSecurityContext(
-	podTemplate *corev1.PodTemplateSpec,
-	userTemplate *corev1.PodTemplateSpec,
-) {
-	for i := range podTemplate.Spec.Containers {
-		if podTemplate.Spec.Containers[i].Name != embeddingContainerName {
-			continue
-		}
-		for _, userContainer := range userTemplate.Spec.Containers {
-			if userContainer.Name == embeddingContainerName && userContainer.SecurityContext != nil {
-				podTemplate.Spec.Containers[i].SecurityContext = userContainer.SecurityContext
-				break
-			}
-		}
-		break
-	}
+	log.FromContext(ctx).V(1).Info("Applied PodTemplateSpec customizations to StatefulSet",
+		"embeddingserver", embedding.Name,
+		"namespace", embedding.Namespace)
+
+	return nil
 }
 
 // applyStatefulSetOverrides applies statefulset-level overrides and returns annotations and labels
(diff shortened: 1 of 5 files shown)


cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -634,87 +644,100 @@ func (*EmbeddingServerReconciler) applyResourceRequirements(embedding *mcpv1beta
...
+// defaults.
+//
+// Invalid user input is treated as a soft failure: the merge is skipped and the
+// StatefulSet is built from controller defaults. The user-facing signal lives on the
+// EmbeddingServer status (set by validateAndUpdatePodTemplateStatus): Phase=Failed
+// and ConditionPodTemplateValid=False. This mirrors the pre-existing tolerant
+// behavior — refusing to create the StatefulSet would leave the resource stuck with
+// no pod and no observable controller-side state, while the validation condition
+// already tells the user exactly why their input was rejected.
+func (*EmbeddingServerReconciler) applyPodTemplateSpecToStatefulSet(
```
yrobla, line 682: **[warning] Code duplication — extractable to `pkg/controllerutil`**

The marshal → `StrategicMergePatch` → unmarshal sequence here is a near-identical copy of `applyPodTemplateSpecToDeployment` in `virtualmcpserver_deployment.go:929`. The only structural difference is the target type (`StatefulSet` vs `Deployment`) and the field path (`.Spec.Template` in both cases).

This logic could live in `pkg/controllerutil` as a generic helper, e.g.:

```go
// ApplyPodTemplateSpecPatch applies a raw strategic merge patch to a base PodTemplateSpec.
func ApplyPodTemplateSpecPatch(base corev1.PodTemplateSpec, patch []byte) (corev1.PodTemplateSpec, error)
```

Both controllers then call the helper and handle errors according to their own policy. As-is, the next controller that needs this (MCPServer?) will copy-paste again.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -634,87 +644,100 @@ func (*EmbeddingServerReconciler) applyResourceRequirements(embedding *mcpv1beta
...
+	originalJSON, err := json.Marshal(statefulSet.Spec.Template)
+	if err != nil {
+		return fmt.Errorf("failed to marshal original PodTemplateSpec: %w", err)
 	}
-	if userTemplate.Spec.SecurityContext != nil {
-		podTemplate.Spec.SecurityContext = userTemplate.Spec.SecurityContext
+
+	patchedJSON, err := strategicpatch.StrategicMergePatch(originalJSON, userJSON, corev1.PodTemplateSpec{})
+	if err != nil {
+		logger.Info("Skipping PodTemplateSpec merge: strategic merge patch failed; StatefulSet will use controller defaults",
```
yrobla, line 710: **[warning] Divergent error handling from the vMCP controller**

In `virtualmcpserver_deployment.go:964`, a `StrategicMergePatch` failure is returned as a hard error that blocks Deployment creation (the caller returns `nil`, leaving the resource unschedulable). Here the same failure is logged at `Info` level and swallowed (`return nil`), so the StatefulSet is silently created with controller defaults.

The soft-fail rationale in the doc comment is sound (validation status already signals the error to the user), but the inconsistency across controllers is surprising: the same invalid `podTemplateSpec` will block a `VirtualMCPServer` from deploying but allow an `EmbeddingServer` to deploy with silently dropped customizations.

Suggestion: either align vMCP to soft-fail, or align EmbeddingServer to hard-fail, and document the chosen policy in `pkg/controllerutil` so future controllers inherit it consistently.

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -634,87 +644,100 @@ func (*EmbeddingServerReconciler) applyResourceRequirements(embedding *mcpv1beta
...
 }
 
-// buildPodTemplate builds the pod template for the statefulset
-func (r *EmbeddingServerReconciler) buildPodTemplate(
-	embedding *mcpv1beta1.EmbeddingServer,
+// buildPodTemplate builds the pod template for the statefulset.
+// User-provided PodTemplateSpec customizations are applied later in
+// statefulSetForEmbedding via strategic merge patch.
+func (*EmbeddingServerReconciler) buildPodTemplate(
+	_ *mcpv1beta1.EmbeddingServer,
```
yrobla: **[nit] Unused parameter**

The `EmbeddingServer` parameter is now `_` since `mergePodTemplateSpec` was removed. The function could simply drop it:

```suggestion
func (*EmbeddingServerReconciler) buildPodTemplate(
	labels map[string]string,
	container corev1.Container,
) corev1.PodTemplateSpec {
```

(Requires updating the single callsite.)

cmd/thv-operator/controllers/embeddingserver_controller.go
```diff
@@ -634,87 +644,100 @@ func (*EmbeddingServerReconciler) applyResourceRequirements(embedding *mcpv1beta
...
 
-	userTemplate := builder.Build()
-	if userTemplate == nil {
-		return
+	// Validate the user-provided PodTemplateSpec is well-formed.
+	// We don't check builder.Build() == nil for "empty" customizations: that helper
+	// only enumerates a subset of PodSpec fields and would skip the patch for
+	// fields like runtimeClassName or topologySpreadConstraints. Strategic merge
+	// patch is a no-op for `{}` anyway, so always running it is safe.
+	if _, err := ctrlutil.NewPodTemplateSpecBuilder(embedding.Spec.PodTemplateSpec, embeddingContainerName); err != nil {
```
yrobla, line 698: **[nit] `PodTemplateSpecBuilder.isEmpty()` gap — pre-existing but newly visible**

The comment here correctly explains why `builder.Build() == nil` is intentionally not used: `isEmpty()` in `pkg/controllerutil/podtemplatespec_builder.go` doesn't enumerate `TopologySpreadConstraints`, `RuntimeClassName`, `HostNetwork`, `DNSConfig`, `ReadinessGates`, and others. So if a user supplies _only_ one of those fields, `Build()` returns `nil` and the patch is skipped.

This is a pre-existing bug (it also affects the vMCP controller on those same fields), but it's now clearly visible in the codebase. Worth tracking as a follow-up: either expand `isEmpty()` to cover the full PodSpec surface, or remove `isEmpty()` entirely (strategic merge patch on `{}` is always a safe no-op as this comment notes).

cmd/thv-operator/controllers/embeddingserver_controller_test.go
```diff
@@ -801,3 +801,200 @@ func TestUpdateEmbeddingServerStatus(t *testing.T) {
...
+			sts := &appsv1.StatefulSet{}
+			err = fakeClient.Get(ctx, types.NamespacedName{
+				Name:      embedding.Name,
+				Namespace: embedding.Namespace,
+			}, sts)
+			require.NoError(t, err, "StatefulSet should be created")
+
+			tt.assertPodSpec(t, sts.Spec.Template.Spec)
+		})
+	}
```
yrobla, line 1060: **[nit] Missing test for soft-failure paths**

The test suite covers the happy path thoroughly. Missing coverage for:
1. Invalid JSON that passes `NewPodTemplateSpecBuilder` validation but causes `StrategicMergePatch` to fail — should fall back to controller defaults (no error returned).
2. A `podTemplateSpec` that is entirely empty `{}` — should be a no-op and not clobber controller-generated fields.

These would pin down the soft-failure contract and prevent regressions if the error handling is later changed.

yrobla (commented):
## Review: Preserve user fields in EmbeddingServer podTemplateSpec merge

Solid approach — strategic merge patch is clearly the right fix. The core logic is correct and well-tested. Main concern is the duplicate implementation and divergent error-handling vs the vMCP controller.

---

**Summary of findings:**
- `[warning]` Core strategic-merge-patch logic is duplicated from `virtualmcpserver_deployment.go` — should live in `pkg/controllerutil`
- `[warning]` Error handling diverges from the vMCP pattern (soft-fail here vs hard-fail there) — inconsistent operator behavior
- `[nit]` `buildPodTemplate` has an unused `_ *mcpv1beta1.EmbeddingServer` parameter
- `[nit]` `PodTemplateSpecBuilder.isEmpty()` doesn't cover the full PodSpec surface — this PR correctly avoids using it, but the gap remains in the vMCP path
- `[nit]` No test for the soft-failure path (invalid JSON that passes builder validation but fails strategic merge patch)

cmd/thv-operator/controllers/embeddingserver_controller_test.go
jhrozek: +1, and one more worth pinning down in the same vein: a user-supplied container named `embedding` with extra env vars or volumeMounts. The sidecar test already covers the additive case (different name → new container), but env and volumeMounts merge by `name` *into* the controller's container, which is a separate code path. Easy for a future change to break that without anyone noticing.

cmd/thv-operator/controllers/embeddingserver_controller.go
JAORMX: Extracted to `ctrlutil.ApplyPodTemplateSpecPatch` in `cmd/thv-operator/pkg/controllerutil/podtemplatespec_patch.go`. Both this controller and the vMCP controller (`virtualmcpserver_deployment.go`) now consume it, so the marshal → strategic merge patch → unmarshal sequence lives in one place.

The helper is policy-neutral — it returns `(corev1.PodTemplateSpec, error)` and lets each caller decide what to do with the error. See the next reply (`:718`) for the rationale on why the two callers continue to make different choices.

cmd/thv-operator/controllers/embeddingserver_controller.go
JAORMX: Kept the existing divergent policies but documented the choice clearly:

- `ctrlutil.ApplyPodTemplateSpecPatch` is policy-neutral (returns `(corev1.PodTemplateSpec, error)`).
- vMCP keeps hard-fail (a patch error blocks Deployment creation, surfaced as a controller error).
- EmbeddingServer keeps soft-fail (logs and falls back to controller defaults; user-facing signal is the existing `PodTemplateValid=False` status condition set by the validation pass).

Both call sites now have a one-line comment naming the policy and pointing at the godoc; the helper's godoc names both consumers and explains the trade-off. Aligning the two would be a behavior change that's out of scope for the bug fix in #5100 — and going the other direction (hard-failing the EmbeddingServer) would regress what this PR fixed.

If we want to align them later (probably toward soft-fail across the board, since that's what the validation condition is for), it should be its own PR with its own user-facing change note.

cmd/thv-operator/controllers/embeddingserver_controller.go
JAORMX: Dropped. `buildPodTemplate` now takes only `labels` and `container`, and the single call site in `statefulSetForEmbedding` is updated.

cmd/thv-operator/controllers/embeddingserver_controller.go
JAORMX: Filed as #5110 — kept out of this PR's scope so the bug fix stays narrow. The issue captures the missing fields (TopologySpreadConstraints, RuntimeClassName, HostNetwork, DNSConfig, ReadinessGates), notes that this PR makes the gap invisible on the EmbeddingServer path but leaves the vMCP early-exit (`builder.Build() == nil`) still affected, and proposes either expanding `isEmpty()` or removing it (strategic merge patch on `{}` is a safe no-op).

cmd/thv-operator/controllers/embeddingserver_controller_test.go
JAORMX: Added two new tests in `cmd/thv-operator/controllers/embeddingserver_controller_test.go`:

1. `TestEmbeddingServer_PodTemplateSpec_SoftFailFallback` — uses a payload that the validation pass accepts but `StrategicMergePatch` rejects: `{"spec":{"containers":[{"name":"embedding","$patch":"invalid"}]}}`. `$patch` is dropped silently by `json.Unmarshal` into `corev1.Container`, so the builder is happy; strategic merge patch then rejects the unknown directive at apply time. Test verifies the StatefulSet is still created and that controller defaults (image, probes, ports) survive.
2. `TestEmbeddingServer_PodTemplateSpec_EmptyObjectIsNoOp` — `{}` patch is a no-op; every controller-generated field on the embedding container survives unchanged.

Also added `TestApplyPodTemplateSpecPatch` in `cmd/thv-operator/pkg/controllerutil/podtemplatespec_patch_test.go` covering the helper directly: nil/empty/`{}` inputs, a user-fields merge, malformed JSON, and a type-mismatched patch.

JAORMX: Good catch — added as a sixth case to `TestEmbeddingServer_PodTemplateSpec_PreservesUserFields`: a user-supplied container named `embedding` with extra `env` and `volumeMounts`. Asserts only one container exists in the result (no append), the controller-set image survives the by-name merge, and both the user env var and volumeMount are present on the merged container.

yrobla (approved):

PR #5105, opened 2026-04-29
Add operator-level defaultImagePullSecrets across all controllers

### LINKED ISSUE #5102: Operator chart: cluster-wide default imagePullSecrets for all spawned workloads
## Summary

The operator Helm chart already exposes `operator.imagePullSecrets` (`deploy/charts/operator/values.yaml:27`), but that value is consumed only by the operator's own Deployment template (`deploy/charts/operator/templates/deployment.yaml:27-28`). It does *not* cascade to the workloads the operator reconciles (MCPServer proxy-runner pods, MCPRegistry pods, VirtualMCPServer pods, EmbeddingServer pods, MCPRemoteProxy pods).

Currently, every CR that pulls from a private registry must specify its own image-pull-secret reference (or rely on the cluster-default ServiceAccount). For deployments where every CR pulls from the same private registry — which is common in an enterprise install of the operator — this is per-CR toil with no DRY mechanism at the operator/chart level.

## Proposal

Add a chart value that the operator picks up at startup and applies as a default to every workload it spawns when the corresponding CR doesn't set its own. Suggested shape:

```yaml
operator:
  # Default image pull secrets applied to all workloads spawned by the operator
  # (proxy-runner, MCPRegistry, VirtualMCPServer, EmbeddingServer, MCPRemoteProxy).
  # Per-CR settings, when present, take precedence.
  defaultImagePullSecrets: []
```

The chart would render this into the operator Deployment as either an env var (e.g. `THV_DEFAULT_IMAGE_PULL_SECRETS=regcred,otherscred`) or a mounted ConfigMap, and the operator's reconcilers would consult it as a fallback when constructing each workload's PodSpec and ServiceAccount.

## Precedence

When both a chart-level default and a CR-level value are set, the CR-level value should win. Suggested merge: chart-level provides defaults, CR-level appends/overrides (matching how `podTemplateSpec` already merges in MCPRegistry/VMCP).

## Why not just use the cluster-default ServiceAccount?

For MCPServer (and ideally MCPRemoteProxy/EmbeddingServer once their gaps are fixed), the operator creates a *dedicated* ServiceAccount per workload. Pre-populating the namespace's `default` ServiceAccount with imagePullSecrets does not help, because the spawned pods don't use it.

## Related issues

- #5099 — MCPRemoteProxy controller silently ignores its existing `spec.resourceOverrides.proxyDeployment.imagePullSecrets` field (bug)
- #5100 — EmbeddingServer's `podTemplateSpec` merge silently drops `imagePullSecrets` (bug)
- #5101 — Add explicit `imagePullSecrets` field on MCPRegistry and VirtualMCPServer (enhancement)

This chart-level default complements the CR-level fields above; it does not replace them.

JAORMX:
## Summary

**Why**: Cluster operators frequently need a registry pull secret (for example, a private mirror of `ghcr.io/stacklok/toolhive/proxyrunner`) applied to every workload the operator spawns. Today the chart only exposes `operator.imagePullSecrets` for the operator's *own* pod (`deploy/charts/operator/templates/deployment.yaml`), forcing users to set the secret on every MCPServer/MCPRemoteProxy/VirtualMCPServer/MCPRegistry/EmbeddingServer CR or to mutate the namespace-default ServiceAccount. This is brittle and easy to miss.

**What**: Adds a chart value `operator.defaultImagePullSecrets`, a small shared `imagepullsecrets` package the operator's reconcilers consume, and wires the value through **every** workload-spawning controller in one shot:

- MCPServer
- MCPRemoteProxy
- MCPRegistry (via `registryapi.manager`)
- VirtualMCPServer
- EmbeddingServer (chart defaults set on base PodSpec; user PodTemplateSpec strategic-merges additively, since `corev1.PodSpec.ImagePullSecrets` carries `patchStrategy:"merge",patchMergeKey:"name"`)

The earlier "foundational PR + follow-ups" plan is superseded — there is no follow-up work needed for the other controllers.

### Precedence rule

When both a chart-level default and a per-CR `imagePullSecrets` are set:

- The CR-level list comes first in the resulting slice (preserving its order).
- Each chart-level default is appended only if its `Name` does not already appear in the CR-level list.
- Deduplication is by `Name`; the CR-level entry wins on name collision.
- The CR's slice is never mutated; callers receive a fresh slice.

This rule is documented on the `Defaults.Merge` Go doc and in the chart `values.yaml` comment.

### Drift-detection bug fixed in MCPServer

While extending scope, the existing `deploymentNeedsUpdate` in MCPServer was found to compare the live `Spec.Template.Spec.ImagePullSecrets` against the CR-only slice while the construction site applied the merged `defaults+CR` slice. With any chart default configured this would have looped forever. Both sites now route through `imagePullSecretsForMCPServer`. Regression test: `TestDeploymentNeedsUpdate_DefaultImagePullSecrets`.

Closes #5102

## Type of change

## Test plan

- [x] `task test` (operator unit and controller tests pass; operator integration tests fail only because `envtest` requires `/usr/local/kubebuilder/bin/etcd` which is not installed in this environment).
- [x] `task lint-fix` (zero issues).
- [x] `helm template test deploy/charts/operator/ --set 'operator.defaultImagePullSecrets={regcred,otherscred}'` renders `TOOLHIVE_DEFAULT_IMAGE_PULL_SECRETS=regcred,otherscred` on the operator container, with `operator.env` rendered before chart-managed env vars so chart-managed values win on duplicate names.
- [x] `helm template test deploy/charts/operator/` (default values) does not render the env var at all.
- [x] `helm lint deploy/charts/operator/` passes.
- [x] `task helm-docs` regenerates `deploy/charts/operator/README.md`.
- [x] New unit tests cover env-var parsing, the `Merge` precedence rule, and the wiring on each newly-integrated controller (`*_default_imagepullsecrets_test.go`).
- [x] New regression test `TestDeploymentNeedsUpdate_DefaultImagePullSecrets` confirms the MCPServer drift fix.

## Changes

| File | Why |
|------|-----|
| `cmd/thv-operator/pkg/imagepullsecrets/defaults.go` (new) | Shared package: `Defaults` value type, `LoadDefaultsFromEnv`, `NewDefaults`, `Merge`, `List`. |
| `cmd/thv-operator/pkg/imagepullsecrets/defaults_test.go` (new) | Unit tests for the package. |
| `cmd/thv-operator/main.go` | Parse env var once at startup; thread `Defaults` through `setupServerControllers`, `setupRegistryController`, `setupAggregationControllers`. Log on typo'd env var. |
| `cmd/thv-operator/controllers/mcpserver_controller.go` | Add `ImagePullSecretsDefaults` field; merge defaults at construction *and* drift-detection sites via `imagePullSecretsForMCPServer`. |
| `cmd/thv-operator/controllers/mcpremoteproxy_controller.go` + `mcpremoteproxy_deployment.go` | Same wiring via `imagePullSecretsForRemoteProxy`. |
| `cmd/thv-operator/controllers/virtualmcpserver_controller.go` + `virtualmcpserver_deployment.go` | Same wiring via `imagePullSecretsForVMCP`; hash annotation now over the merged list. |
| `cmd/thv-operator/controllers/embeddingserver_controller.go` | Chart defaults placed on base PodSpec; strategic-merge with user PTS unions additively. |
| `cmd/thv-operator/pkg/registryapi/manager.go` + `rbac.go` + `deployment.go` | `NewManager` takes `Defaults`; `manager` merges with `mcpRegistry.Spec.ImagePullSecrets` for both SA and Deployment construction. |
| `cmd/thv-operator/controllers/{mcpserver,mcpremoteproxy,virtualmcpserver,embeddingserver}_default_imagepullsecrets_test.go` | Wiring tests + drift regression. |
| `deploy/charts/operator/values.yaml` | New `defaultImagePullSecrets` value with documentation. |
| `deploy/charts/operator/templates/deployment.yaml` | Render `operator.env` before chart-managed env vars (chart-managed wins on collision); render `TOOLHIVE_DEFAULT_IMAGE_PULL_SECRETS` env var on the operator container. |
| `deploy/charts/operator/README.md` | Regenerated by `helm-docs`. |

Yes. Cluster operators can now set `operator.defaultImagePullSecrets` in the chart values to apply pull secrets to every workload the operator spawns, instead of editing each CR. Existing per-CR `imagePullSecrets` continue to work unchanged and take precedence on name collision.

- Replies to the prior round of inline comments are inline; the substantive surface (drift bug + scope expansion) is in a top-level comment so it's not buried.
- All five workload-spawning reconcilers now consume `Defaults`. Pure-config reconcilers (MCPGroup, ToolConfig, MCPExternalAuthConfig, MCPOIDCConfig, MCPTelemetryConfig, MCPServerEntry) do not need it.
- `Defaults` is a value type rather than a pointer to keep the zero value usable: a reconciler that omits the field still gets correct merge behavior (CR-level value passes through unchanged).

## Large PR Justification

The diff exceeds the 1000-line threshold because the only useful end state for this feature is **all five workload-spawning controllers consuming the chart-level defaults**. A staged rollout was the original plan — the v1 of this PR landed only `MCPServer` as a "foundational" change — but reviewer feedback (and a perpetual-reconcile-loop bug found in MCPServer's drift detection while extending scope) made it clear the partial state is worse than the complete one:

1. **The chart `values.yaml` claim is only true if every controller is wired**. Splitting meant shipping `defaultImagePullSecrets: [...]` documented as "applied to every workload" while four out of five controllers silently ignored it. Reviewers correctly flagged this on the v1 of the PR.
2. **Each controller's wiring is mechanically identical** (`ImagePullSecretsDefaults` field on the reconciler + helper method that calls `Defaults.Merge` + setup-path threading). Splitting these into four follow-up PRs would have produced four near-identical reviews.
3. **The shared `imagepullsecrets` package and the `main.go` threading must land together** with at least one consumer; otherwise the package is unused and `setupControllersAndWebhooks` has a parameter nothing reads.
4. **Drift detection must mirror construction in the same PR** that introduces the merge. The MCPServer drift bug fixed here would not exist if both sites had been touched together — splitting that fix into its own PR would have been pure churn.

What's NOT in the diff that could have been: the integration tests for embedding-server, mcp-external-auth, mcp-registry, and virtualmcp packages all fail locally because envtest binaries aren't installed; those are pre-existing test-infrastructure issues unrelated to this change. The wiring tests for the four newly-integrated controllers are all unit tests that exercise the exact code path this PR touches, kept deliberately minimal (two cases per helper) with cross-references to `defaults_test.go::TestDefaultsMerge` for the exhaustive merge-precedence cases.


deploy/charts/operator/values.yaml
```diff
@@ -25,6 +25,15 @@ operator:
...
   imagePullSecrets: []
+
+  # -- List of image pull secrets that the operator applies as defaults to every
+  # workload it spawns (proxy runners, vMCP servers, registry API, etc.).
+  # Per-CR `imagePullSecrets` take precedence on name collisions; chart-level
+  # entries are appended additively. Each entry is a string (the Secret name);
+  # the operator parses these once at startup from the
+  # THV_DEFAULT_IMAGE_PULL_SECRETS environment variable. The Secrets must exist
+  # in the namespace where each workload is created.
+  defaultImagePullSecrets: []
```
jhrozek, line 43: This says the defaults apply to every workload (proxy runners, vMCP servers, registry API, etc.), but the diff only wires `MCPServerReconciler`. MCPRemoteProxy / MCPRegistry / VirtualMCPServer / EmbeddingServer / registry API will silently ignore the env var until follow-up PRs land. Are those follow-ups planned? If yes, worth tightening this comment to reflect what's actually wired today (with a link to a tracking issue) — and the same wording lives in the package doc-comment in `cmd/thv-operator/pkg/imagepullsecrets/defaults.go` and the README row, both of which would need the same edit. Right now a chart user who sets `defaultImagePullSecrets: [regcred]` will be confused why their MCPRemoteProxy / EmbeddingServer pods still ImagePullBackOff.

deploy/charts/operator/templates/deployment.yaml
```diff
@@ -76,6 +76,10 @@ spec:
             value: "{{ .Values.operator.proxyHost }}"
           - name: TOOLHIVE_REGISTRY_API_IMAGE
             value: "{{ .Values.registryAPI.image }}"
+          {{- with .Values.operator.defaultImagePullSecrets }}
+          - name: THV_DEFAULT_IMAGE_PULL_SECRETS
```
jhrozek: `operator.env` is rendered right after this block (line 84). If a user happens to put `THV_DEFAULT_IMAGE_PULL_SECRETS` into `operator.env`, two env entries with the same name end up on the container and Kubernetes silently keeps the last one — meaning the user-supplied entry overrides the chart-managed value with no warning. Probably worth either rendering `operator.env` first (so chart-managed values win) or `fail`ing at template time if the user declares a reserved env name.

cmd/thv-operator/pkg/imagepullsecrets/defaults.go
```diff
@@ -0,0 +1,119 @@
...
+// environment variable.
+//
+// The variable is a comma-separated list of secret names. An empty or unset
+// variable yields an empty Defaults whose Merge is a no-op.
+func LoadDefaultsFromEnv() Defaults {
+	return NewDefaults(strings.Split(os.Getenv(EnvVar), ","))
+}
+
+// List returns a copy of the configured default imagePullSecrets. Callers
+// must not mutate the returned slice's element values.
```
jhrozek: Doc says callers `must not mutate the returned slice's element values`, but the implementation already does a fresh `make + copy` on every call (and `TestDefaultsListReturnsCopy` proves mutation doesn't leak back). `LocalObjectReference` only holds a string anyway, so element mutation can't reach internal state regardless. The restriction in the doc is misleading — it tells callers to be careful about something the implementation already protects against.

```suggestion
// List returns a freshly allocated copy of the configured default
// imagePullSecrets. The caller may freely mutate the returned slice.
```

cmd/thv-operator/pkg/imagepullsecrets/defaults_test.go
```diff
@@ -0,0 +1,282 @@
...
+				prev, hadPrev := os.LookupEnv(EnvVar)
+				t.Cleanup(func() {
+					if hadPrev {
+						_ = os.Setenv(EnvVar, prev)
+					} else {
+						_ = os.Unsetenv(EnvVar)
+					}
+				})
+				require.NoError(t, os.Unsetenv(EnvVar))
+			} else {
```
jhrozek: The `unset` branch is functionally redundant with the `empty env var yields empty defaults` case — `strings.Split("", ",")` produces `[""]`, which `NewDefaults` filters out, so both cases exercise the same parsing path. Dropping the `unset` field plus this whole branch lets every sub-test go through `t.Setenv` (matching `.claude/rules/testing.md`) and removes the silent error swallowing in the cleanup (`_ = os.Setenv(...)` / `_ = os.Unsetenv(...)`).

cmd/thv-operator/main.go
```diff
@@ -98,7 +99,18 @@ func main() {
...
-	if err := setupControllersAndWebhooks(mgr); err != nil {
+	// Parse cluster-wide default imagePullSecrets once at startup. The Defaults
+	// value is shared (by copy) with every reconciler that constructs workloads.
+	imagePullSecretsDefaults := imagepullsecrets.LoadDefaultsFromEnv()
+	if defaults := imagePullSecretsDefaults.List(); len(defaults) > 0 {
+		names := make([]string, 0, len(defaults))
+		for _, ref := range defaults {
+			names = append(names, ref.Name)
+		}
+		setupLog.Info("loaded cluster-wide default imagePullSecrets", "secrets", names)
```
jhrozek: If an admin typos the env var (e.g. `THV_DEFAULT_IMAGE_PULL_SECRETS=' , '`), the parser silently yields an empty Defaults and there's no log entry — this branch only fires when non-empty, so the misconfiguration is invisible. An `else if os.Getenv(imagepullsecrets.EnvVar) != ""` log at debug level would make typos diagnosable without adding INFO noise.

cmd/thv-operator/controllers/mcpserver_default_imagepullsecrets_test.go
```diff
@@ -0,0 +1,172 @@
...
+	t.Parallel()
+
+	tests := []struct {
+		name        string
+		defaults    []string
+		crSecrets   []corev1.LocalObjectReference
+		wantSecrets []corev1.LocalObjectReference
+	}{
+		{
+			name:        "chart default applied when CR has none",
```
jhrozek: These four cases (`chart default applied`, `CR value used`, `no overlap`, `name collision`) are already exhaustively covered in `defaults_test.go::TestDefaultsMerge`. The unique value of this controller test is proving that the merged slice actually reaches `ServiceAccount.ImagePullSecrets` and `Deployment.Spec.Template.Spec.ImagePullSecrets` — re-proving Merge precedence here is noise that doubles the diff. Could collapse to two cases per helper (one merged-with-overlap to show the wiring works, one nil to show empty stays empty) without losing coverage.

cmd/thv-operator/controllers/mcpserver_default_imagepullsecrets_test.go
```diff
@@ -0,0 +1,172 @@
...
+						ImagePullSecrets: tt.crSecrets,
+					},
+				}
+			}
+
+			require.NoError(t, tc.ensureRBACResources())
+
+			// Proxy-runner ServiceAccount.
+			sa := &corev1.ServiceAccount{}
+			require.NoError(t, tc.client.Get(t.Context(), types.NamespacedName{
```
jhrozek, line 76: Tiny nit: sibling `mcpserver_rbac_test.go` (where `setupTest` lives) uses `context.TODO()` everywhere; this file uses `t.Context()`. Both work, but mixing them within the same package is distracting. If you want to standardize on `t.Context()` going forward, worth migrating `mcpserver_rbac_test.go` in the same PR.

cmd/thv-operator/main.go
jhrozek: Naming nit: the structured key `"secrets"` won't match a log search for `imagePullSecrets`, which is the user-facing terminology everywhere else (Helm key `defaultImagePullSecrets`, env var `THV_DEFAULT_IMAGE_PULL_SECRETS`, Go field `ImagePullSecretsDefaults`). Consider `"imagePullSecrets"`.

jhrozek (commented):
Reviewed with a few specialist agents. Design shape looks right (matches the cert-manager / OpenTelemetry operator pattern of operator-flag/env-var defaults), so most comments are details. The two I'd want a response on before merge are the values.yaml scope claim and the operator.env ordering question.

deploy/charts/operator/values.yaml
JAORMX: Took the broader fix here: rather than tighten the comment, this revision wires every workload-spawning controller (MCPRemoteProxy, MCPRegistry via registryapi.manager, VirtualMCPServer, EmbeddingServer) to consume the chart-level Defaults, so the values.yaml claim now matches what's actually applied. The 'foundational PR' framing in the original description no longer applies — this lands the full plumbing in one shot.

deploy/charts/operator/templates/deployment.yaml
JAORMX: Fixed by rendering `operator.env` first, before all chart-managed env vars. Kubernetes keeps the last entry on a duplicate-named env, so chart-managed values now win on collision. Inline comment in the template documents the rationale so a future edit doesn't silently flip the precedence.

cmd/thv-operator/pkg/imagepullsecrets/defaults.go
JAORMX: Applied your suggested wording. Also swapped the manual `make+copy` for `slices.Clone` since the early-return-nil is the only thing keeping it from being a one-liner.

cmd/thv-operator/pkg/imagepullsecrets/defaults_test.go
JAORMX: Dropped the `unset` field and the manual `os.LookupEnv`+`os.Setenv` cleanup branch; every sub-test now goes through `t.Setenv` per `.claude/rules/testing.md`. The empty-string case already exercises `strings.Split("", ",")` so coverage is preserved.

cmd/thv-operator/main.go
JAORMX: Added an `else if` branch that logs at INFO when the env var is set but parses to no valid names (e.g., `' , '` or `',,,'`). The diagnostic includes the raw value so the misconfiguration is visible without needing DEBUG. controller-runtime's logr handle has no native WARN level, so INFO with a clear message is the closest we get.

cmd/thv-operator/controllers/mcpserver_default_imagepullsecrets_test.go
JAORMX: Trimmed to two cases per helper (overlap-with-collision + empty) and added a comment pointing readers at `defaults_test.go::TestDefaultsMerge` for the exhaustive precedence cases. Also added a `TestDeploymentNeedsUpdate_DefaultImagePullSecrets` regression test for a related drift bug I found while extending scope — see the top-level PR comment for the bug write-up.

cmd/thv-operator/controllers/mcpserver_default_imagepullsecrets_test.go
JAORMX: Left `mcpserver_rbac_test.go` alone for this PR — migrating the ~17 `context.TODO()` callers in that file is unrelated to imagePullSecrets and would balloon the diff for a stylistic consistency win. Worth its own follow-up if we want to standardize the package on `t.Context()`.

cmd/thv-operator/main.go
JAORMX: Renamed the structured key from `secrets` to `imagePullSecrets` in both the success-load and the typo-detection branches, matching the Helm key, the env var, and the Go field.

JAORMX:
Force-pushed a substantially expanded revision (b36ed41f5). Two things worth flagging beyond the inline replies:

**Scope expanded — this is no longer a "foundational" PR.** All five workload-spawning reconcilers now consume `imagepullsecrets.Defaults`: MCPServer, MCPRemoteProxy, MCPRegistry (via `registryapi.manager`), VirtualMCPServer, and EmbeddingServer. The follow-up issues for the remaining controllers can be closed once this lands. Also added focused tests for each newly-wired controller (`*_default_imagepullsecrets_test.go`).

**Drift-detection bug found and fixed in MCPServer.** While extending the wiring I noticed `deploymentNeedsUpdate` compared the live `Deployment.Spec.Template.Spec.ImagePullSecrets` against `mcpServer.Spec.ResourceOverrides.ProxyDeployment.ImagePullSecrets` (CR-only), while the construction site at `deploymentForMCPServer` was setting the merged `defaults+CR` slice. With any chart default configured, the comparison was always unequal and `deploymentNeedsUpdate` returned `true` on every reconcile — perpetual loop. Both sites now route through `imagePullSecretsForMCPServer`, the same helper-method pattern the other reconcilers use. Regression test: `TestDeploymentNeedsUpdate_DefaultImagePullSecrets` in `mcpserver_default_imagepullsecrets_test.go`. The same risk exists in spirit for VirtualMCPServer's `imagePullSecretsNeedsUpdate` and MCPRemoteProxy's `podSpecNeedsUpdate`; both already route through their `imagePullSecretsForX` helpers, so they were correct from the start of this revision.

The PR description still calls this "foundational" — I'll fix that in a separate edit before merge.

jhrozek (commented):
Two small things from a re-read of the chart wiring — both are pre-merge ergonomics, neither is a blocker.

deploy/charts/operator/templates/deployment.yaml
```diff
@@ -76,8 +86,9 @@ spec:
             value: "{{ .Values.operator.proxyHost }}"
           - name: TOOLHIVE_REGISTRY_API_IMAGE
             value: "{{ .Values.registryAPI.image }}"
-          {{- if .Values.operator.env }}
-            {{- toYaml .Values.operator.env | nindent 10 }}
+          {{- with .Values.operator.defaultImagePullSecrets }}
+          - name: THV_DEFAULT_IMAGE_PULL_SECRETS
+            value: "{{ join "," . }}"
```
jhrozek: The values.yaml comment says each entry is a string, but `operator.imagePullSecrets` a few lines above takes the standard `[{name: foo}]` shape and users will copy that pattern. If they do, `join` ends up calling `fmt.Sprint` on each map and the env var renders as `THV_DEFAULT_IMAGE_PULL_SECRETS=map[name:foo]` — the operator then logs that it loaded a Secret literally named `map[name:foo]` and pods quietly hit ImagePullBackOff. Worth either accepting both shapes (`range` + `default . .name`), adding a `values.schema.json` constraint, or at least making the doc loud about it ("plain strings only — not `{name: ...}` objects").

cmd/thv-operator/pkg/imagepullsecrets/defaults.go
```diff
@@ -0,0 +1,120 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// EnvVar is the environment variable name that the operator parses at startup
+// to populate cluster-wide default imagePullSecrets.
+//
+// The value is a comma-separated list of secret names, e.g. "regcred,otherscred".
+// Whitespace around entries is tolerated; empty entries are skipped.
+const EnvVar = "THV_DEFAULT_IMAGE_PULL_SECRETS"
```
jhrozek: Every other env var on the operator container in the Helm template uses the `TOOLHIVE_*` prefix (`TOOLHIVE_RUNNER_IMAGE`, `TOOLHIVE_PROXY_HOST`, `TOOLHIVE_REGISTRY_API_IMAGE`, `TOOLHIVE_USE_CONFIGMAP`). `THV_*` is the CLI-side convention. Worth aligning before this ships — once released, renaming costs a deprecation cycle. Matching changes if you take this: this const, the log string in `main.go`, and the comment + rendered name in `templates/deployment.yaml`.

```suggestion
const EnvVar = "TOOLHIVE_DEFAULT_IMAGE_PULL_SECRETS"
```

JAORMX: Renamed to `TOOLHIVE_DEFAULT_IMAGE_PULL_SECRETS` everywhere — `EnvVar` const + package doc, the `main.go` typo-branch log string, the chart `values.yaml` doc, the `templates/deployment.yaml` env name + comment, and the regenerated `README.md`. No `THV_*` references remain. Took the suggestion as-is; agreed it's worth aligning before this ships.

deploy/charts/operator/templates/deployment.yaml
JAORMX: Took the "accept both shapes" path so users copying `[{name: foo}]` from `operator.imagePullSecrets` above don't silently break. The template now normalizes:

- plain strings: `["regcred", "otherscred"]`
- objects with a `name` field: `[{name: regcred}, {name: otherscred}]`
- mixed entries also work

Anything else (numbers, nested lists, objects missing `name`) fails the helm template render with `operator.defaultImagePullSecrets[i]: entry must be a string or an object with a 'name' field, got <kind>` instead of producing `TOOLHIVE_DEFAULT_IMAGE_PULL_SECRETS=map[name:foo]` and a runtime ImagePullBackOff. The `values.yaml` comment now spells out both shapes with an example, and the regenerated `README.md` carries the same wording.

Verified locally:
- `defaultImagePullSecrets={regcred,otherscred}` → `value: "regcred,otherscred"`
- `[{name: regcred}, {name: otherscred}]` → `value: "regcred,otherscred"`
- `[regcred, {name: otherscred}]` → `value: "regcred,otherscred"`
- `[123]` → `Error: ... entry must be a string or an object with a 'name' field, got float64`
- `[{foo: bar}]` → `Error: ... object entry must have a non-empty 'name' field`
- default (empty list) → no env var rendered

jhrozek (approved):
Verified the b4d4248d7 follow-up locally — env var rename is complete (no THV_* references left, tests use the EnvVar const so they track automatically), and the Helm template now accepts both shapes correctly. Walked through all the cases: plain strings, [{name: foo}] objects, mixed, empty default (no env var rendered), and the bad-input paths (object without name, object with empty name, number, nested list) all fail the render with clear messages. helm lint passes, operator unit tests pass.

One nit not worth blocking on: an empty-string entry [""] in the strings path renders as value: ",regcred" (leading comma). The operator's parser filters empties after the split so it's functionally correct, just slightly ugly. A {{- if not $entry -}}{{- continue -}}{{- end -}} skip would symmetrize with the empty-name rejection if you want it.

Issue #5101, opened 2026-04-29
Add explicit imagePullSecrets field to MCPRegistry and VirtualMCPServer


JAORMX:
## Summary

MCPRegistry and VirtualMCPServer both spawn workloads from container images, but neither has an explicit `imagePullSecrets` field. Users who need to pull from a private registry must know to use the raw-JSON `spec.podTemplateSpec` escape hatch:

```yaml
spec:
  podTemplateSpec:
    spec:
      imagePullSecrets:
        - name: regcred
```

This *does* work today (verified by the integration test at `cmd/thv-operator/test-integration/mcp-registry/deployment_update_test.go:63` for MCPRegistry, and by the strategic-merge implementation for VirtualMCPServer), but it has two drawbacks:

1. **Discoverability.** A user reading the CRD schema sees no field for image pull secrets and won't know the workaround exists.
2. **Inconsistency.** MCPServer has a first-class field at `spec.resourceOverrides.proxyDeployment.imagePullSecrets`. Anyone managing both kinds of resources has to remember the API differs.

## Proposal

Add an explicit field to both CRDs that mirrors the MCPServer shape, e.g.:

- `MCPRegistry.spec.resourceOverrides.deployment.imagePullSecrets`
- `VirtualMCPServer.spec.resourceOverrides.deployment.imagePullSecrets`

The field should be of type `[]corev1.LocalObjectReference` (matching MCPServer's `ProxyDeploymentOverrides.ImagePullSecrets`) and propagate to:

- the spawned Deployment's `PodSpec.ImagePullSecrets`, and
- the ServiceAccount the operator creates for that workload.

The existing `podTemplateSpec` workaround should continue to work (so existing manifests don't break); when both are set, an explicit precedence rule should be documented (suggested: explicit field acts as default, podTemplateSpec wins on overlap, matching the existing podTemplateSpec merge semantics).

## Out of scope

This issue is *not* about fixing broken propagation — both CRDs already work via `podTemplateSpec`. It's purely about adding a discoverable, consistent API surface.

For genuinely broken propagation in the same area, see #5099 (MCPRemoteProxy) and #5100 (EmbeddingServer).

PR #5106, opened 2026-04-29
Add explicit imagePullSecrets field to MCPRegistry

JAORMX:
## Summary

MCPRegistry workloads previously had no first-class image pull secrets field — users had to use the raw-JSON `spec.podTemplateSpec` escape hatch. That works (the existing integration test at `cmd/thv-operator/test-integration/mcp-registry/deployment_update_test.go:63` proves it) but it has two drawbacks:

1. **Discoverability** — a user reading the CRD schema sees no field for image pull secrets and won't know the workaround exists.
2. **Inconsistency** — MCPServer has a first-class field at `spec.resourceOverrides.proxyDeployment.imagePullSecrets`. Anyone managing both kinds of resources has to remember the API differs.

This PR adds `spec.imagePullSecrets` to `MCPRegistry`. The field is `[]corev1.LocalObjectReference` and propagates to:

- the registry-api Deployment's `PodSpec.ImagePullSecrets`
- the operator-managed ServiceAccount the registry API runs as

**Placement choice.** I chose `spec.imagePullSecrets` (flat, top-level on the spec) rather than introducing a `spec.resourceOverrides.deployment.imagePullSecrets` struct. MCPRegistry has no `ResourceOverrides` today, and adding one with a single field would be speculative scaffolding. If/when other override fields land, `imagePullSecrets` can be moved or kept as a top-level alias — but adding a struct now without a second field to justify it is unnecessary.

**Precedence with podTemplateSpec.** When both `spec.imagePullSecrets` and `spec.podTemplateSpec.spec.imagePullSecrets` are set:
- `spec.imagePullSecrets` is applied first as a controller-generated default.
- `spec.podTemplateSpec.spec.imagePullSecrets` is the user override and wins on overlap (the list is treated atomically — user's list replaces the default entirely on the Deployment).
- The ServiceAccount is always populated from `spec.imagePullSecrets`; `podTemplateSpec` has no notion of a ServiceAccount.

This matches the existing `podTemplateSpec` merge semantics in `MergePodTemplateSpecs` and is documented in the field's Go doc comment. The legacy `podTemplateSpec` workaround keeps working unchanged.

Part of #5101

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — operator unit tests pass; the new `TestBuildRegistryAPIDeployment_ImagePullSecrets` covers explicit field, empty, podTemplateSpec override, podTemplateSpec without imagePullSecrets, and legacy podTemplateSpec-only behaviour. `TestEnsureRBACResources_ImagePullSecrets` verifies the ServiceAccount picks up the explicit field.
- [x] Manual testing — generated CRD YAML inspected (`deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpregistries.yaml`) shows the new `imagePullSecrets` array property and the legacy `podTemplateSpec` field is unchanged.

## API Compatibility

This adds a new optional field. Existing CRs continue to work; no migration required.

Yes. Users can now set `spec.imagePullSecrets` directly on `MCPRegistry`:

```yaml
kind: MCPRegistry
metadata:
  name: my-registry
spec:
  configYAML: |
    sources:
      - name: default
        format: toolhive
  imagePullSecrets:
    - name: regcred
```

The previous `spec.podTemplateSpec` workaround keeps working.

VirtualMCPServer has the same gap (tracked by #5101) but is being landed as a separate follow-up PR to keep each diff small and reviewable. The two CRDs do not share the deployment-build code path so the changes don't combine cleanly.


cmd/thv-operator/pkg/registryapi/podtemplatespec.go
```diff
@@ -106,6 +106,26 @@ func WithContainer(container corev1.Container) PodTemplateSpecOption {
...
 }
 
+// WithImagePullSecrets sets the image pull secrets on the PodSpec.
+// Existing secrets are preserved; new secrets (by name) are appended for idempotency.
+func WithImagePullSecrets(secrets []corev1.LocalObjectReference) PodTemplateSpecOption {
+	return func(pts *corev1.PodTemplateSpec) {
+		if len(secrets) == 0 {
+			return
+		}
+		existing := make(map[string]bool, len(pts.Spec.ImagePullSecrets))
```
yrobla: **[warning] Deduplication logic is dead code**

The `WithImagePullSecrets` option is applied to `b.defaultSpec` (via `Apply(...)`), and `defaultSpec` is always initialized as `&corev1.PodTemplateSpec{}` — so `pts.Spec.ImagePullSecrets` is always empty when this runs. The `existing` map is always empty and the dedup loop never filters anything out.

This complexity is misleading: a reader might assume there are pre-existing entries that need deduplication, but there never are. A simple assignment is both clearer and correct:

```suggestion
		pts.Spec.ImagePullSecrets = secrets
```

(The `len(secrets) == 0` early-return above already handles the empty case.)

yrobla (commented):
## Review: Add `spec.imagePullSecrets` to MCPRegistrySpec

Clean, well-structured feature — the CRD field, RBAC wiring, and deployment propagation are all correct. Main concerns are a dead dedup path in `WithImagePullSecrets`, a subtle SA/Deployment behavioral split that users may not expect, and missing integration test coverage for the new field.

---

**Summary of findings:**
- `[warning]` `WithImagePullSecrets` deduplication logic is dead code — `defaultSpec` always starts empty, so the `existing` map never has anything to deduplicate
- `[warning]` No integration test exercises the new `spec.imagePullSecrets` field — existing integration tests only cover the `podTemplateSpec` path
- `[warning]` SA/Deployment split-brain: using only `spec.podTemplateSpec.spec.imagePullSecrets` (the old way) populates the Deployment but leaves the SA without credentials; the docs mention this but the risk is subtle
- `[nit]` `MergePodTemplateSpecs` docstring says "All other PodSpec fields: User values preserved as-is" — now stale since imagePullSecrets has explicit merge logic
- `[nit]` `WithImagePullSecrets` function comment is misleading about what it does

cmd/thv-operator/pkg/registryapi/podtemplatespec.go
```diff
@@ -106,6 +106,26 @@ func WithContainer(container corev1.Container) PodTemplateSpecOption {
 	}
 }
 
+// WithImagePullSecrets sets the image pull secrets on the PodSpec.
+// Existing secrets are preserved; new secrets (by name) are appended for idempotency.
```
yrobla: **[nit] Comment is inaccurate**

"Existing secrets are preserved; new secrets (by name) are appended for idempotency" describes merge behaviour that never actually occurs (see sibling comment on the `existing` map). The real behaviour is: the field is set from `spec.imagePullSecrets` when the defaultSpec is built; `MergePodTemplateSpecs` later decides whether the user's PodTemplateSpec overrides it.

Suggested:
```suggestion
// WithImagePullSecrets sets the image pull secrets on the pod spec from
// spec.imagePullSecrets. These are treated as defaults; a user-provided
// PodTemplateSpec can override them via MergePodTemplateSpecs.
```

cmd/thv-operator/pkg/registryapi/podtemplatespec.go
```diff
@@ -436,6 +456,14 @@ func MergePodTemplateSpecs(defaultPTS, userPTS *corev1.PodTemplateSpec) corev1.P
 	// Merge volumes: add default volumes that user hasn't specified
 	result.Spec.Volumes = mergeVolumesUserFirst(defaultPTS.Spec.Volumes, result.Spec.Volumes)
 
+	// Merge image pull secrets: user values win on overlap; otherwise inherit defaults.
+	// The list is treated atomically — if the user specifies any imagePullSecrets in
+	// PodTemplateSpec, theirs replace the defaults entirely. This matches the +listType=atomic
+	// semantics on MCPRegistrySpec.ImagePullSecrets.
+	if len(result.Spec.ImagePullSecrets) == 0 {
```
yrobla, line 456: **[nit] `MergePodTemplateSpecs` docstring is now stale**

The existing comment block says:
> All other PodSpec fields: User values preserved as-is

That's no longer true — `imagePullSecrets` now has explicit "inherit default if user didn't set it" logic. The "Merge behavior" list should include:
```
// - ImagePullSecrets: User list wins atomically if non-empty; otherwise inherits defaults
```

cmd/thv-operator/api/v1beta1/mcpregistry_types.go
```diff
@@ -109,6 +109,23 @@ type MCPRegistrySpec struct {
...
+	// These are applied to both the registry-api Deployment's PodSpec.ImagePullSecrets
+	// and to the operator-managed ServiceAccount the registry API runs as, so private
+	// images are pullable through either path.
+	//
+	// Precedence with PodTemplateSpec:
+	//   - This field is applied first as the controller-generated default.
+	//   - Values set under spec.podTemplateSpec.spec.imagePullSecrets are user overrides
+	//     and win on overlap. If the user supplies imagePullSecrets via PodTemplateSpec,
+	//     those replace the default list on the Deployment (the list is treated atomically).
+	//   - The ServiceAccount is always populated from this field — PodTemplateSpec does not
```
yrobla, line 135: **[warning] SA/Deployment split-brain is a latent footgun**

The docs correctly state: "The ServiceAccount is always populated from this field — PodTemplateSpec does not affect the ServiceAccount."

This means:
- User sets only `spec.podTemplateSpec.spec.imagePullSecrets: [creds]` (legacy path) → Deployment gets `creds`, SA does NOT → pod may fail to pull if the node relies on SA-level credential injection (common on managed K8s, e.g. GKE Workload Identity or OpenShift).
- User sets only `spec.imagePullSecrets: [creds]` (new path) → both Deployment and SA get `creds` → correct.
- User sets both → Deployment uses PodTemplateSpec override, SA uses the spec field → they can diverge.

The documentation is accurate but passive. Consider adding an explicit warning here and/or in the CRD doc so operators know they should prefer `spec.imagePullSecrets` over the PodTemplateSpec path for private image auth.

cmd/thv-operator/pkg/registryapi/deployment_test.go
```diff
@@ -604,3 +604,88 @@ func TestBuildRegistryAPIDeployment_PodTemplateSpecHash(t *testing.T) {
...
+					Name:      "test-registry",
+					Namespace: "test-namespace",
+				},
+				Spec: tt.spec,
+			}
+			deployment, err := mgr.buildRegistryAPIDeployment(t.Context(), mcpRegistry, "test-registry-server-config")
+			require.NoError(t, err)
+			require.NotNil(t, deployment)
+
+			assert.Equal(t, tt.expected, deployment.Spec.Template.Spec.ImagePullSecrets)
```
yrobla, line 688: **[warning] No integration test for the new `spec.imagePullSecrets` field**

The existing integration tests in `test-integration/mcp-registry/deployment_update_test.go` only test imagePullSecrets via the `spec.podTemplateSpec` raw JSON path. None of them exercise the new dedicated `spec.imagePullSecrets` field.

Missing integration coverage:
1. Set `spec.imagePullSecrets` → verify Deployment pod spec gets it.
2. Set `spec.imagePullSecrets` → verify the ServiceAccount gets it (the SA path is not tested at all in integration).
3. Update `spec.imagePullSecrets` (change secret name) → verify both Deployment and SA are updated.
4. Set both `spec.imagePullSecrets` and `spec.podTemplateSpec.spec.imagePullSecrets` → verify Deployment uses PodTemplateSpec override, SA uses spec field.

The unit tests in this file cover the Deployment side adequately. The SA reconcile on field change has no integration-level verification.

cmd/thv-operator/pkg/registryapi/podtemplatespec.go
```diff
@@ -106,6 +106,26 @@ func WithContainer(container corev1.Container) PodTemplateSpecOption {
 	}
 }
 
+// WithImagePullSecrets sets the image pull secrets on the PodSpec.
+// Existing secrets are preserved; new secrets (by name) are appended for idempotency.
+func WithImagePullSecrets(secrets []corev1.LocalObjectReference) PodTemplateSpecOption {
+	return func(pts *corev1.PodTemplateSpec) {
+		if len(secrets) == 0 {
```
jhrozek, line 114: The empty-slice case treats the Deployment and the ServiceAccount differently:

- `WithImagePullSecrets` short-circuits on `len(secrets) == 0`, so an explicit `spec.imagePullSecrets: []` is a no-op on the Deployment (just like omitting the field).
- The RBAC path (`pkg/kubernetes/rbac/rbac.go:121`) only preserves prior values when `desiredImagePullSecrets == nil`. An empty (non-nil) slice fails that check and **wipes the SA's existing `ImagePullSecrets`**, which is also the only way to clear OpenShift's auto-managed `dockercfg` entries from the SA.

So `imagePullSecrets: []` becomes the de-facto "force-clear-the-SA" affordance, while `imagePullSecrets:` (omitted) is "leave the SA alone." Operators have no way to discover that distinction from the field doc.

Two reasonable fixes:
1. Normalize empty slice → nil before passing to RBAC, so empty/omitted behave identically (omitting the field becomes the only path; clearing requires deleting and recreating, matching CRD-merge intuition).
2. Document the distinction explicitly in the field comment and call out that `imagePullSecrets: []` is the supported way to force-clear SA-level pull secrets.

Either is fine; the current behaviour is just undocumented and surprising.

cmd/thv-operator/pkg/registryapi/podtemplatespec.go
JAORMX: Done in 1c014c10. Replaced the `existing` map and dedup loop with a direct `pts.Spec.ImagePullSecrets = secrets` (the `len(secrets) == 0` early-return is preserved). The user-vs-default merge already happens later in `MergePodTemplateSpecs`.

cmd/thv-operator/pkg/registryapi/podtemplatespec.go
JAORMX: Done in 1c014c10. Rewrote the godoc to your suggested wording: it now explains that `WithImagePullSecrets` sets defaults from `spec.imagePullSecrets` and that a user PodTemplateSpec can override them via `MergePodTemplateSpecs`.

cmd/thv-operator/pkg/registryapi/podtemplatespec.go
JAORMX: Done in 1c014c10. Added a new bullet to the `MergePodTemplateSpecs` godoc list: `ImagePullSecrets: User list wins atomically if non-empty; otherwise inherits defaults`, sitting next to the existing per-field bullets.

cmd/thv-operator/api/v1beta1/mcpregistry_types.go
JAORMX: Done in bbdfcb32. Strengthened the godoc on `MCPRegistrySpec.ImagePullSecrets` to call out that this field is the ONLY path that reaches the operator-managed ServiceAccount, that `spec.podTemplateSpec.spec.imagePullSecrets` populates the Deployment pod spec only, and that platforms relying on SA-level credential injection (GKE Workload Identity, OpenShift dockercfg, EKS IRSA) can fail when only the legacy path is used. Also added a one-line "Use this field for new manifests" recommendation. Regenerated CRD manifests and `docs/operator/crd-api.md` so the warning lands in the user-facing CRD reference.

cmd/thv-operator/pkg/registryapi/deployment_test.go
JAORMX: Done in 16f54da8. Added four Ginkgo specs in `cmd/thv-operator/test-integration/mcp-registry/deployment_update_test.go` under a new `Context("spec.imagePullSecrets is the SA-aware path for image pull credentials")`:

1. `spec.imagePullSecrets` only -> Deployment pod spec carries the value.
2. `spec.imagePullSecrets` only -> ServiceAccount carries the value (this SA path was previously untested in integration).
3. `spec.imagePullSecrets` updated on an existing CR -> both Deployment and SA rotate.
4. `spec.imagePullSecrets` + `spec.podTemplateSpec.spec.imagePullSecrets` together -> Deployment uses the PodTemplateSpec override (atomic replacement), SA still tracks `spec.imagePullSecrets`.

I ran the suite locally with `setup-envtest` (1.31.0): all 25 specs in `mcp-registry` pass.

cmd/thv-operator/pkg/registryapi/podtemplatespec.go
JAORMX: Picked option 1 — normalize the empty slice to nil at the RBAC client. After auditing the four `EnsureRBACResources` callers (MCPServer, MCPRemoteProxy, VirtualMCPServer, MCPRegistry), none of them passes a non-nil empty slice with intent to wipe; all derive `ImagePullSecrets` from a CRD field whose empty/omitted distinction has the same problem. So the central fix in `cmd/thv-operator/pkg/kubernetes/rbac/rbac.go` (commit a3772031) gives every CRD the same surprise-free behavior in one place.

Empty and nil are now treated identically in `upsertServiceAccount` (preserves both `Secrets` and `ImagePullSecrets`). Updated the godoc on `MCPRegistrySpec.ImagePullSecrets` to spell out the equivalence and that recreating the resource is the supported way to clear SA-level pull secrets.

Guarded by two new unit tests:
- `cmd/thv-operator/pkg/kubernetes/rbac/rbac_test.go` `TestUpsertServiceAccountWithOwnerReference` subtest "preserves existing ImagePullSecrets when desired list is empty (not nil)" — covers the contract for all callers.
- `cmd/thv-operator/pkg/registryapi/rbac_test.go` `TestEnsureRBACResources_EmptyImagePullSecretsPreservesSAPullSecrets` — exercises the full MCPRegistry path with `spec.imagePullSecrets: []` and a pre-populated SA simulating OpenShift dockercfg.

yrobla (approved):

PR #5107, opened 2026-04-29
Add explicit imagePullSecrets field to VirtualMCPServer

JAORMX:
## Summary

VirtualMCPServer workloads previously had no first-class image pull secrets field — users had to use the raw-JSON `spec.podTemplateSpec` escape hatch. That works (the strategic-merge code at `cmd/thv-operator/controllers/virtualmcpserver_deployment.go:921` preserves it) but it has two drawbacks:

1. **Discoverability** — a user reading the CRD schema sees no field for image pull secrets and won't know the workaround exists.
2. **Inconsistency** — MCPServer has a first-class field at `spec.resourceOverrides.proxyDeployment.imagePullSecrets`. Anyone managing both kinds of resources has to remember the API differs.

This PR adds `spec.imagePullSecrets` to `VirtualMCPServer`. The field is `[]corev1.LocalObjectReference` and propagates to:

- the vMCP Deployment's `PodSpec.ImagePullSecrets`
- the operator-managed ServiceAccount the vMCP server runs as

**Placement choice.** I chose `spec.imagePullSecrets` (flat, top-level on the spec) rather than introducing a `spec.resourceOverrides.deployment.imagePullSecrets` struct. VirtualMCPServer has no `ResourceOverrides` today, and adding one with a single field would be speculative scaffolding. If/when other override fields land, `imagePullSecrets` can be moved or kept as a top-level alias.

**Precedence with podTemplateSpec.** When both `spec.imagePullSecrets` and `spec.podTemplateSpec.spec.imagePullSecrets` are set:

- `spec.imagePullSecrets` is applied first as a controller-generated default on the Deployment's `PodSpec.ImagePullSecrets`.
- The existing strategic merge patch in `applyPodTemplateSpecToDeployment` handles overlap natively: `corev1.PodSpec.ImagePullSecrets` has `patchStrategy:"merge"` with `patchMergeKey:"name"`. Distinct secret names from the explicit field and PodTemplateSpec are **unioned**; if the same name appears in both, the PodTemplateSpec entry wins on overlap.
- The ServiceAccount is always populated from `spec.imagePullSecrets`; `podTemplateSpec` has no notion of a ServiceAccount.

This is documented in the field's Go doc comment. Note that VirtualMCPServer's overlap behaviour (union by name via strategic merge) differs slightly from MCPRegistry's (atomic replace, sister PR #5106) — that's because vMCP already uses Kubernetes-native strategic merge for podTemplateSpec, while MCPRegistry uses a custom merge function with atomic-list semantics on its CRD field. Each PR follows its CRD's existing merge convention.

The legacy `podTemplateSpec` workaround keeps working unchanged.

Part of #5101

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — operator unit tests pass; the new `TestDeploymentForVirtualMCPServer_ImagePullSecrets` covers explicit field, empty, podTemplateSpec union/override, podTemplateSpec without imagePullSecrets, and legacy podTemplateSpec-only behaviour. `TestVirtualMCPServerEnsureRBACResources_ImagePullSecrets` verifies the ServiceAccount picks up the explicit field.
- [x] Manual testing — generated CRD YAML inspected (`deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_virtualmcpservers.yaml`) shows the new `imagePullSecrets` array property and existing fields are unchanged.

## API Compatibility

This adds a new optional field. Existing CRs continue to work; no migration required.

Yes. Users can now set `spec.imagePullSecrets` directly on `VirtualMCPServer`:

```yaml
kind: VirtualMCPServer
metadata:
  name: my-vmcp
spec:
  groupRef:
    name: my-group
  incomingAuth:
    type: anonymous
  imagePullSecrets:
    - name: regcred
```

The previous `spec.podTemplateSpec` workaround keeps working.

This is the second of two PRs for #5101. The MCPRegistry counterpart is #5106. The two CRDs do not share the deployment-build code path so the changes don't combine cleanly — they are landed as independent PRs to keep each diff small.


yrobla (commented):
## Review: Add explicit `imagePullSecrets` field to VirtualMCPServer

The field is wired correctly on initial create — both the SA and the Deployment PodSpec are populated. However there is a blocker: once deployed, changing `spec.imagePullSecrets` will never propagate to the Deployment. Also flagging a cross-PR API inconsistency vs MCPRegistry.

---

**Summary of findings:**
- `[blocker]` `deploymentNeedsUpdate` has no check for `spec.imagePullSecrets` — post-create changes are silently ignored by the reconciler
- `[warning]` Merge semantics differ from MCPRegistry (PR #5106): vMCP unions by name (strategic merge), MCPRegistry replaces atomically — inconsistent API across sibling resources
- `[warning]` No integration test for the new field — no e2e coverage of SA population, Deployment population, or update propagation
- `[nit]` `builder.Build() == nil` guard in `applyPodTemplateSpecToDeployment` can still swallow valid PodTemplateSpec patches (pre-existing gap in `isEmpty()`)

cmd/thv-operator/controllers/virtualmcpserver_deployment.go
```diff
@@ -195,6 +195,7 @@ func (r *VirtualMCPServerReconciler) deploymentForVirtualMCPServer(
 				Spec: corev1.PodSpec{
 					TerminationGracePeriodSeconds: int64Ptr(vmcpTerminationGracePeriodSeconds),
 					ServiceAccountName:            serviceAccountName,
+					ImagePullSecrets:              vmcp.Spec.ImagePullSecrets,
```
yrobla, line 206: **[blocker] `spec.imagePullSecrets` changes are never propagated to an existing Deployment**

This line correctly sets `ImagePullSecrets` when the Deployment is first created. But `deploymentNeedsUpdate` (in `virtualmcpserver_controller.go`) has no check for this field:

- `containerNeedsUpdate` checks image, args, env, ports, resources, security context, service account — not `PodSpec.ImagePullSecrets`.
- `podTemplateSpecNeedsUpdate` only hashes `vmcp.Spec.PodTemplateSpec.Raw` — `spec.imagePullSecrets` is a separate field and not included.
- `podTemplateMetadataNeedsUpdate` checks labels/annotations only.

Result: a user who adds or changes `spec.imagePullSecrets` on an existing resource will see no Deployment rollout. The pod continues pulling with the old (or missing) credentials until the Deployment is updated by some other change.

Fix: add a check in `deploymentNeedsUpdate` (or `podTemplateSpecNeedsUpdate`) comparing the live Deployment's `PodSpec.ImagePullSecrets` against what `deploymentForVirtualMCPServer` would produce. The simplest form:

```go
if !reflect.DeepEqual(deployment.Spec.Template.Spec.ImagePullSecrets, vmcp.Spec.ImagePullSecrets) {
    return true
}
```

(Note: when `spec.podTemplateSpec` also sets imagePullSecrets the deployed list is a union, not a direct copy of `vmcp.Spec.ImagePullSecrets`. The comparison needs to account for that, or fall back to rebuilding the full desired deployment and comparing.)

cmd/thv-operator/controllers/virtualmcpserver_deployment_test.go
```diff
@@ -868,3 +868,106 @@ func TestBuildCABundleVolumesForEntries(t *testing.T) {
...
+			require.NoError(t, mcpv1beta1.AddToScheme(scheme))
+			require.NoError(t, corev1.AddToScheme(scheme))
+
+			vmcp := &mcpv1beta1.VirtualMCPServer{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-vmcp",
+					Namespace: "default",
+				},
+				Spec: tt.spec,
+			}
```
yrobla, line 960: **[blocker] No test covering the update path — the above gap is untestable with this suite**

All test cases in `TestDeploymentForVirtualMCPServer_ImagePullSecrets` call `deploymentForVirtualMCPServer` directly and assert on initial creation. There is no test that:
1. Creates a Deployment with `spec.imagePullSecrets: [A]`
2. Calls `deploymentNeedsUpdate` with a mutated spec `[B]`
3. Asserts that an update is triggered and the Deployment gets `[B]`

This is the test that would have caught the `deploymentNeedsUpdate` gap above.

cmd/thv-operator/api/v1beta1/virtualmcpserver_types.go
```diff
@@ -107,6 +108,25 @@ type VirtualMCPServerSpec struct {
...
+	//   - This field is applied first as the controller-generated default.
+	//   - Values set under spec.podTemplateSpec.spec.imagePullSecrets are user
+	//     overrides and merged on top using Kubernetes strategic merge patch
+	//     (patchStrategy:"merge", patchMergeKey:"name"). Distinct secret names
+	//     from the explicit field and PodTemplateSpec are unioned; if the same
+	//     name appears in both, the PodTemplateSpec entry wins on overlap.
+	//   - The ServiceAccount is always populated from this field — PodTemplateSpec
+	//     does not affect the ServiceAccount.
+	//
+	// +listType=atomic
```
yrobla, line 143: **[warning] Merge semantics inconsistent with MCPRegistry (PR #5106)**

The doc here says: _"Distinct secret names from the explicit field and PodTemplateSpec are unioned"_ — i.e. strategic merge by name.

MCPRegistry's sibling field (PR #5106) says: _"If the user supplies imagePullSecrets via PodTemplateSpec, those replace the default list on the Deployment (the list is treated atomically)"_ — i.e. atomic replace.

This difference comes from vMCP using `strategicpatch.StrategicMergePatch` on the whole deployment (which merges `imagePullSecrets` by `name` per the `+patchMergeKey` tag) while MCPRegistry uses `MergePodTemplateSpecs` with an `if len == 0` fallback (atomic).

Neither is wrong on its own, but a user managing both resource types will get different semantics for what looks like the same field. Worth aligning — either change MCPRegistry to use strategic merge too, or document the difference prominently in both CRDs.

cmd/thv-operator/controllers/virtualmcpserver_controller_test.go
```diff
@@ -288,6 +288,55 @@ func TestVirtualMCPServerEnsureRBACResources(t *testing.T) {
...
+
+	sa := &corev1.ServiceAccount{}
+	require.NoError(t, fakeClient.Get(t.Context(), types.NamespacedName{
+		Name:      vmcpServiceAccountName(vmcp.Name),
+		Namespace: vmcp.Namespace,
+	}, sa))
+
+	expected := []corev1.LocalObjectReference{
+		{Name: "vmcp-creds"},
+		{Name: "extra-creds"},
```
yrobla, line 335: **[warning] No integration test for the new field**

The `test-integration/virtualmcp/` suite has no coverage for `spec.imagePullSecrets`. Missing scenarios:
1. Create a VirtualMCPServer with `spec.imagePullSecrets` → verify the SA gets them and the Deployment pod spec gets them.
2. Update `spec.imagePullSecrets` → verify Deployment rolls out with new value (this will fail today given the blocker above).
3. Set both `spec.imagePullSecrets` and `spec.podTemplateSpec.spec.imagePullSecrets` → verify they are unioned on the Deployment but the SA only reflects `spec.imagePullSecrets`.

cmd/thv-operator/controllers/virtualmcpserver_deployment.go
yrobla, line 206: **[nit] `builder.Build() == nil` in `applyPodTemplateSpecToDeployment` can silently skip valid patches (pre-existing)**

If the user's `spec.podTemplateSpec` only sets fields not covered by `isEmpty()` (e.g. `topologySpreadConstraints`, `runtimeClassName`), `Build()` returns nil and the patch is skipped. This is a pre-existing gap flagged in the #5104 review, but it's worth noting here too: a user who sets ONLY `spec.podTemplateSpec.spec.topologySpreadConstraints` (with no `spec.imagePullSecrets`) will have their constraints silently dropped. The interaction with the new `spec.imagePullSecrets` field makes it more likely users combine these approaches.

JAORMX: Fixed in 6c9e1b14. Added a new `imagePullRefsHashAnnotation` (mirroring the existing `podTemplateSpecHashAnnotation` pattern the reviewer suggested) plus a corresponding `imagePullSecretsNeedsUpdate` step in the drift-detection chain — see `cmd/thv-operator/controllers/virtualmcpserver_controller.go` (the new helper after `podTemplateSpecNeedsUpdate`) and the new constant + `imagePullSecretsHash` helper in `cmd/thv-operator/controllers/virtualmcpserver_deployment.go`.

Direct comparison of `deployment.Spec.Template.Spec.ImagePullSecrets` against `vmcp.Spec.ImagePullSecrets` was not safe because the live list is the strategic-merge union with `spec.podTemplateSpec.spec.imagePullSecrets`. The hash-annotation approach matches the convention already used for PodTemplateSpec drift; PodTemplateSpec changes (including imagePullSecrets a user adds there) remain covered by `podTemplateSpecNeedsUpdate`, so the two annotations together cover every input that influences `PodSpec.ImagePullSecrets`.

The hash normalizes the slice by sorting on `Name` before hashing, so a reorder is not treated as drift (covered by the `reorder is a no-op` test case).

cmd/thv-operator/controllers/virtualmcpserver_deployment_test.go
JAORMX: Added `TestDeploymentForVirtualMCPServer_ImagePullSecrets_UpdatePath` and `TestImagePullSecretsHash` in `cmd/thv-operator/controllers/virtualmcpserver_deployment_test.go` (commit 6c9e1b14).

The update-path test is table-driven and covers every case the comment requested:
- pure add (`nil` → `[secret-a]`)
- pure remove (`[secret-a]` → `nil`)
- replace (`[secret-a]` → `[secret-b]`)
- extend (`[secret-a]` → `[secret-a, secret-b]`)
- replace combined with PodTemplateSpec union (asserts the deployed list is the strategic-merge union of the new explicit field and the PodTemplateSpec entry)
- reorder is a no-op (regression guard against spurious update churn from non-deterministic ordering)

Each subtest:
1. Builds the initial Deployment via `deploymentForVirtualMCPServer`.
2. Mutates `vmcp.Spec.ImagePullSecrets` to the new value.
3. Asserts both `imagePullSecretsNeedsUpdate` and the parent `deploymentNeedsUpdate` flag the change.
4. Rebuilds the Deployment and asserts `Spec.Template.Spec.ImagePullSecrets` matches the expected union.
5. Re-runs the drift check against the freshly-built Deployment and asserts it returns `false` (steady-state — no reconcile loop).

`TestImagePullSecretsHash` separately covers the hash helper's order-insensitivity, empty-list semantics, and discriminating between different sets.

cmd/thv-operator/api/v1beta1/virtualmcpserver_types.go
JAORMX: Strengthened the godoc on `VirtualMCPServerSpec.ImagePullSecrets` in commit ef169409. The field comment now explicitly documents:
- The Kubernetes-native strategic-merge-by-name semantics (with the `corev1.PodSpec` patch tags called out by name).
- That order in the merged result is not guaranteed (strategic merge by name is order-insensitive).
- That the operator-managed ServiceAccount's imagePullSecrets list is populated ONLY from this field — `spec.podTemplateSpec.spec.imagePullSecrets` does NOT reach the SA, with explicit guidance to list a secret here rather than in PodTemplateSpec when SA-side pull is needed.
- A "Note on cross-CRD consistency" paragraph that names the divergence with MCPRegistry's atomic-replace strategy and points at the follow-up.

Filed the alignment follow-up as #5111 — labeled `enhancement,kubernetes,operator,vmcp,tech-debt`. The issue body lays out both possible resolutions (align both to K8s-native strategic merge OR align both to atomic replace) without picking a side, links #5106 and #5107, and lists acceptance criteria including removing the cross-CRD note from this godoc once the decision lands.

Regenerated `crd-api.md` and the operator CRD manifests to pick up the longer description.

cmd/thv-operator/controllers/virtualmcpserver_controller_test.go
JAORMX: Added `cmd/thv-operator/test-integration/virtualmcp/virtualmcpserver_imagepullsecrets_integration_test.go` in commit 4e11c25c. The new file is labeled `Label("k8s", "imagepullsecrets")` and adds three Ordered Contexts:

1. **`When spec.imagePullSecrets is set`** — creates a vMCP with two secrets and asserts (a) the Deployment's `PodSpec.ImagePullSecrets` carries both names, and (b) the operator-managed ServiceAccount's `ImagePullSecrets` carries both names.

2. **`When spec.imagePullSecrets is updated on an existing CR`** — regression test for the drift-detection blocker. Creates with `[secret-a]`, waits for the Deployment to converge, then patches the spec to `[secret-b]` and asserts both the Deployment and the SA roll over to the new value. This is the exact path the unit tests in the controllers package can't reach — they only run the Deployment builder, not the full reconcile.

3. **`When both spec.imagePullSecrets and spec.podTemplateSpec carry imagePullSecrets`** — covers the combined-source case the comment asked for. Sets `spec.imagePullSecrets: [shared, explicit-only]` and `spec.podTemplateSpec.spec.imagePullSecrets: [shared, podtemplate-only]` (overlap on `shared`, disjoint otherwise). Asserts:
   - The Deployment carries the strategic-merge union: `[shared, explicit-only, podtemplate-only]`.
   - The ServiceAccount reflects ONLY `spec.imagePullSecrets`: `[shared, explicit-only]`.

Ran locally with envtest 1.31.0:
```
KUBEBUILDER_ASSETS="$(setup-envtest use 1.31.0 -p path)" \
  ginkgo --label-filter="imagepullsecrets" -v ./cmd/thv-operator/test-integration/virtualmcp/...
# Test Suite Passed (5/5 specs in this file)
```
And the full `cmd/thv-operator/test-integration/virtualmcp/...` suite (37/37) passes alongside it.

cmd/thv-operator/controllers/virtualmcpserver_deployment.go
JAORMX: Acknowledged — pre-existing, scoped to the #5104 follow-up rather than this PR. There's no dedicated tracking issue filed yet for the `isEmpty()` / `builder.Build() == nil` gap, so leaving it tracked separately for the #5104 follow-up. Happy to file one explicitly if a maintainer prefers; just flag the preference and I'll open it.

yrobla (approved):

PR #5117, opened 2026-04-29
Add --session-ttl flag and fix three session timeout bugs

JAORMX:
## Summary

MCP sessions were being evicted by the background TTL cleanup goroutine even while actively
in use, and there was no way to configure how long sessions should live. This PR fixes both.

**Configurable TTL (`--session-ttl` flag):**
- Added `SessionTTL time.Duration` to `types.Config`, `RunConfig`, and `vmcp.ServeConfig`
- Added `WithSessionTTL` builder option in `pkg/runner`
- Added `--session-ttl` flag to `thv run`, `thv vmcp serve`, and `vmcp serve`
- Fixed option-ordering trap: all proxy constructors defer session manager creation until
  after all options are applied (previously `WithSessionTTL` after `WithSessionStorage`
  would silently drop Redis storage)

**Three session timeout bug fixes:**
1. SSE keep-alive ticks now call `sessionManager.Get` to refresh the TTL while the socket
   is open, preventing eviction of idle-SSE clients
2. Notifications forwarded via the streamable proxy now refresh the session TTL (the session
   header was not being read in that code path)
3. Removed unbounded ephemeral session creation for requests without `Mcp-Session-Id` —
   sessionless requests now use `sessID=""` for in-process waiter-channel correlation
   without creating persistent session objects (memory leak fix)

## Type of change

- [x] Bug fix
- [x] New feature

## Test plan

All tests in the changed packages pass. Pre-existing failures in `pkg/mcp/server` are
unrelated (missing container runtime in the sandbox environment).

## Changes

| File | Change |
|------|--------|
| `pkg/transport/types/transport.go` | Add `SessionTTL` to `Config` |
| `pkg/transport/proxy/httpsse/http_proxy.go` | Defer manager construction; fix keep-alive TTL refresh |
| `pkg/transport/proxy/streamable/streamable_proxy.go` | Defer manager construction; fix notification TTL refresh; remove ephemeral sessions |
| `pkg/transport/proxy/transparent/transparent_proxy.go` | Defer manager construction; wire `SessionTTL` |
| `pkg/transport/http.go`, `pkg/transport/stdio.go` | Add `sessionTTL` field; wire from factory |
| `pkg/transport/factory.go` | Pass `SessionTTL` from config to transports |
| `pkg/runner/config.go` | Add `SessionTTL` to `RunConfig` |
| `pkg/runner/config_builder.go` | Add `WithSessionTTL` option |
| `pkg/runner/runner.go` | Resolve effective TTL; pass to transport config and Redis |
| `cmd/thv/app/run_flags.go`, `cmd/thv/app/vmcp.go` | Add `--session-ttl` flag |
| `cmd/vmcp/app/commands.go` | Add `--session-ttl` flag |
| `pkg/vmcp/cli/serve.go` | Add `SessionTTL` to `ServeConfig`; validation |
| `pkg/runner/config_builder_test.go` | Tests for `WithSessionTTL` |

Yes. New optional `--session-ttl` flag on `thv run`, `thv vmcp serve`, and `vmcp serve`.
Accepts a Go duration string (e.g. `--session-ttl=4h`). Defaults to the existing hardcoded
values (2h for proxies, 30m for vMCP) so existing behaviour is unchanged when omitted.

### DIFF
--- a/pkg/runner/config.go
+++ b/pkg/runner/config.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"time"
 
 	"github.com/stacklok/toolhive-core/permissions"
 	v1beta1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1beta1"
@@ -202,6 +203,10 @@ type RunConfig struct {
 	// POST-based health check instead of the default GET probe.
 	// Applies to both remote URLs and local container workloads.
 	Stateless bool `json:"stateless,omitempty" yaml:"stateless,omitempty"`
+
+	// SessionTTL is the inactivity timeout for proxy sessions.
+	// Zero uses the transport default (2h). Negative values are rejected by the builder.
+	SessionTTL time.Duration `json:"session_ttl,omitempty" yaml:"session_ttl,omitempty" swaggertype:"primitive,integer"`
 
 	// ProxyMode is the effective HTTP protocol the proxy uses.
 	// For stdio transports, this is the configured mode (sse or streamable-http).
--- a/pkg/runner/config.go
+++ b/pkg/runner/config.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
-	"time"
 
 	"github.com/stacklok/toolhive-core/permissions"
 	v1beta1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1beta1"
@@ -204,9 +203,13 @@ type RunConfig struct {
 	// Applies to both remote URLs and local container workloads.
 	Stateless bool `json:"stateless,omitempty" yaml:"stateless,omitempty"`
 
-	// SessionTTL is the inactivity timeout for proxy sessions.
-	// Zero uses the transport default (2h). Negative values are rejected by the builder.
-	SessionTTL time.Duration `json:"session_ttl,omitempty" yaml:"session_ttl,omitempty" swaggertype:"primitive,integer"`
+	// SessionTTL is the inactivity timeout for proxy sessions, expressed as a Go
+	// duration string (e.g. "30m", "2h", "168h"). Empty uses the transport
+	// default (2h). Negative durations and values that fail time.ParseDuration
+	// are rejected at runtime.
+	// String (not time.Duration) keeps the wire format unit-explicit: a
+	// time.Duration field serializes as nanoseconds in JSON.
+	SessionTTL string `json:"session_ttl,omitempty" yaml:"session_ttl,omitempty" example:"2h"`
 
 	// ProxyMode is the effective HTTP protocol the proxy uses.
 	// For stdio transports, this is the configured mode (sse or streamable-http).
--- a/pkg/transport/proxy/streamable/streamable_proxy.go
+++ b/pkg/transport/proxy/streamable/streamable_proxy.go
@@ -56,6 +56,14 @@ type HTTPProxy struct {
 	// Session manager for streamable HTTP sessions
 	sessionManager *session.Manager
 
+	// sessionTTL is the resolved inactivity timeout for the session manager.
+	// Defaults to session.DefaultSessionTTL; overridable via WithSessionTTL.
+	sessionTTL time.Duration
+
+	// sessionStorage is the optional custom storage backend for the session manager.
+	// When nil, in-memory LocalStorage is used. Set via WithSessionStorage.
+	sessionStorage session.Storage
+
 	// Waiters keyed by JSON-encoded request ID -> one-shot channel for response delivery
 	waiters sync.Map // map[string]chan jsonrpc2.Message
 	// Map of compositeKey(sessID|idKey) -> original client JSON-RPC ID to restore before replying
@@ -78,11 +86,18 @@ func WithSessionStorage(storage session.Storage) Option {
 		if storage == nil {
 			return
 		}
-		if p.sessionManager != nil {
-			_ = p.sessionManager.Stop()
+		p.sessionStorage = storage
+	}
+}
+
+// WithSessionTTL overrides the session inactivity timeout used by this proxy.
+// Zero or negative values are ignored so the constructor's default is preserved.
+func WithSessionTTL(ttl time.Duration) Option {
+	return func(p *HTTPProxy) {
+		if ttl <= 0 {
+			return
 		}
-		sFactory := func(id string) session.Session { return session.NewStreamableSession(id) }
-		p.sessionManager = session.NewManagerWithStorage(session.DefaultSessionTTL, sFactory, storage)
+		p.sessionTTL = ttl
 	}
 }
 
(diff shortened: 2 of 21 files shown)


pkg/transport/proxy/streamable/streamable_proxy.go
```diff
@@ -662,17 +682,13 @@ func (p *HTTPProxy) resolveSessionForRequest(
...
+	// Non-initialize requests without a session ID are allowed in sessionless mode
+	// (no session object created, composite key uses empty sessID for in-process routing).
 	if sessID == "" {
-		sessID = uuid.New().String()
-		if err := p.ensureSession(sessID); err != nil {
-			writeHTTPError(w, http.StatusInternalServerError, fmt.Sprintf("Failed to create session: %v", err))
-			return "", false, err
-		}
-		return sessID, false, nil
+		return "", false, nil
```
jerm-dro: **`blocker:`** Removing `ensureSession` here also removed the per-request unique routing token. `sessID=""` collapses every concurrent sessionless request onto the same `compositeKey("", idKey)`, and `p.waiters.Store(ck, ch)` (`sync.Map`) silently overwrites.

Concrete trace with two clients both sending `id=1`:

1. Client A stores waiter `ch_A` at `"|n:1"`. `idRestore["|n:1"] = id_A`.
2. Client B stores waiter `ch_B` at `"|n:1"` — overwrites `ch_A`. `idRestore["|n:1"] = id_B`.
3. Both requests forwarded to the container with id rewritten to `"|n:1"`.
4. Container returns two responses with id `"|n:1"`. The dispatcher delivers the first one to the surviving waiter (`ch_B`) and drops the second with `"waiter channel full"`.
5. `restoreResponseID` rewrites the id back to `id_B`. Client B receives a response with B's expected JSON-RPC id but a **payload from whichever request the container processed first** — possibly client A's tool result. Client A's goroutine sits on the orphaned `ch_A` until `requestTimeout` (60s) and times out.

JSON-RPC clients pick IDs locally and typically start at `1`, so collision is the expected state under any concurrent sessionless traffic, not the unlucky one. Auth middleware ran per request, but the response *payload* still crosses client boundaries — that's a data-leakage path within the proxy's auth boundary.

The "memory leak" framing in the PR description is real but bounded by TTL cleanup; this swaps a bounded leak for a silent correctness regression. The minimal fix decouples "don't create a session object" from "have a unique routing token":

```go
// Non-initialize sessionless: per-request routing token only.
// Not registered with sessionManager — no session object, no Redis write,
// no leak — but composite-key uniqueness is preserved.
return uuid.New().String(), false, nil
```

Same fix needed in `resolveSessionForBatch` for the sessionless-batch path. Worth a regression test that fires two concurrent sessionless requests with the same JSON-RPC id and asserts each gets its own response.

jerm-dro (dismissed):

pkg/transport/proxy/streamable/streamable_proxy.go
JAORMX: You're right -- thanks for the detailed trace. Confirmed each step: the in-process `waiters` and `idRestore` `sync.Map`s silently overwrite under `compositeKey("", idKey)`, and the symptom is exactly what you described (cross-client payload delivery with the id rewritten back to the receiver's, plus a 60s timeout for the loser). Confidentiality issue, not just bounded leak vs. correctness — agreed.

Pushed [72521d09e](https://github.com/stacklok/toolhive/pull/5117/commits/72521d09e) with your suggested minimal fix:

- `resolveSessionForRequest` and `resolveSessionForBatch` now return `uuid.New().String()` for sessionless POSTs. The token is used solely as an in-process routing key and is **not** registered with `sessionManager` (no session object, no storage write), so the memory-leak fix the PR was originally addressing is preserved.
- `setSessionHeader` stays `false` on the sessionless path, so the UUID never leaks to the client and subsequent requests remain sessionless.
- Doc comments on both functions plus the `waiters`/`idRestore` field comments now spell out the uniqueness invariant and frame it as a confidentiality bug, so a future contributor doesn't downgrade the token back to `""`.

Regression test: `TestSessionlessConcurrentRequestsAreNotMixed` in `streamable_proxy_spec_test.go`. It fires two concurrent sessionless POSTs with the same JSON-RPC id but different methods, with a sync barrier in the fake backend (collect both incoming requests before sending any response) so both proxy waiters are registered first -- that's the precondition under which the bug overwrites. The backend echoes `req.Method` back so each client can prove it received its own payload. I verified the test reliably fails against the pre-fix code (504 + cross-talk) and passes with the fix.

pkg/runner/config.go
```diff
@@ -203,6 +204,10 @@ type RunConfig struct {
 	// Applies to both remote URLs and local container workloads.
 	Stateless bool `json:"stateless,omitempty" yaml:"stateless,omitempty"`
 
+	// SessionTTL is the inactivity timeout for proxy sessions.
+	// Zero uses the transport default (2h). Negative values are rejected by the builder.
+	SessionTTL time.Duration `json:"session_ttl,omitempty" yaml:"session_ttl,omitempty" swaggertype:"primitive,integer"`
```
jhrozek: Nit (non-blocking): `time.Duration` serializes as nanoseconds in JSON, which puts this field out of step with every other duration in the API surface — `AccessTokenLifespan`, `DialTimeout`, `ReadTimeout`, `WriteTimeout` are all `string` parsed via `time.ParseDuration` ("5s", "3s", …).

The Swagger description ends with "default (2h)" but doesn't tell the caller the wire unit is ns, so anyone sending `7200` thinking "seconds" gets a 7.2µs TTL with no error. Cleanest fix is probably `string` + `time.ParseDuration` to match the convention; alternative is a custom JSON codec. At minimum it'd be good to mention the unit in the Swagger description so it isn't a silent footgun.

pkg/transport/proxy/streamable/streamable_proxy.go
```diff
@@ -662,17 +702,14 @@ func (p *HTTPProxy) resolveSessionForRequest(
...
+	// setSessionHeader stays false so the client never sees this UUID and the
+	// next request remains sessionless.
 	if sessID == "" {
-		sessID = uuid.New().String()
-		if err := p.ensureSession(sessID); err != nil {
-			writeHTTPError(w, http.StatusInternalServerError, fmt.Sprintf("Failed to create session: %v", err))
-			return "", false, err
-		}
-		return sessID, false, nil
+		return uuid.New().String(), false, nil
```
jhrozek: Nit (non-blocking): `uuid.New()` is `Must(NewRandom())` under the hood, so it panics if `crypto/rand.Read` ever fails. Pre-PR that was only a concern on initialize; with this change it runs on every sessionless POST, which is a hot path. Restricted environments (FIPS, tight seccomp) where rand reads can fail would turn into per-request panics.

Might be worth switching this (and the two siblings at 662 and 695) to `uuid.NewRandom()` and writing a 500 on err — matches the pattern already used after `ensureSession` on line 698.

jhrozek (commented):
Two non-blocking nits inline, plus one scope question:

The `--session-ttl` knob is wired through `thv run`, `thv vmcp serve`, and `vmcp serve`, but I don't see a matching field on `MCPServerSpec` or `VirtualMCPServerSpec`. Intentional follow-up (and is there an issue tracking it), or did it just not land here? The arch doc already calls out the gap at `docs/arch/13-vmcp-scalability.md:58`, so it'd feel natural to close it in the same PR — but I get it if you want to keep this one tight.

Nothing blocking. The sessionless cross-talk fix and the deferred manager-construction pattern are clean.

JAORMX:
@jhrozek I can do it as a follow up for this one. I wanted to keep it small so it's easy to review.

amirejaz (approved):

amirejaz:
LGTM

jhrozek (approved):
Both nits addressed cleanly — string wire format with the unit spelled out in the swagger description, and uuid.NewRandom + 500 on err in all three sites. Tests and lint pass locally on the touched packages. LGTM, thanks for the quick turnaround.

PR #5130, opened 2026-04-29
Add follow-up CIMD E2E and unit tests

amirejaz:
## Summary

Follow-up to #5085 — adds the two test cases Jakub flagged as out-of-scope for the main PR.

## Changes

**E2E tests (`test/e2e/`)**
- Extended `cimdMockAuthServer` with a `rejectCIMD` option: on the first authorize request with a CIMD URL as `client_id`, the mock AS redirects with `error=invalid_client`. Subsequent requests (with a DCR-issued `client_id`) succeed normally.
- New test: *"falls back to DCR when AS rejects the CIMD client_id"* — exercises the full retry path end-to-end: CIMD attempted → rejected → DCR fallback fires → server becomes reachable.

**Unit test (`pkg/auth/remote/handler_test.go`)**
- `TestResolveClientCredentials`: 5 table-driven cases covering `CachedCIMDClientID` precedence over DCR and static credentials, empty-secret guarantee for CIMD, and DCR/static fallback when `CachedCIMDClientID` is empty.

## Dependency

Depends on #5085. Targeting that branch as base; will be rebased onto main once #5085 merges.


jhrozek (approved):
LGTM. Tests-only follow-up to the CIMD work — useful additions:

- TestResolveClientCredentials pins down the CIMD > DCR > static priority chain, which is the kind of precedence logic that's easy to break in a refactor.
- The "AS rejects CIMD client_id" e2e scenario is the right next test after #5085 — it exercises the runtime fallback that the discovery-based test doesn't reach.

A few non-blocking nits if you fancy a follow-up:

- `isCIMDClientID` in the helper duplicates `oauthproto.IsClientIDMetadataDocumentURL`. The other test file in this package already imports `oauthproto`, so we could just call the production function directly. There's a TODO in `cimd.go` to tighten that check per the IETF draft, and if that lands the mock will silently accept URLs production rejects.
- The "CachedClientID used when CachedCIMDClientID is empty" case asserts that a DCR client_id pairs with a *static* client_secret when no `CachedClientSecretRef` is stored. Looking at handler.go:236-264 that's what the code does today, but it feels off — a public DCR client and a configured secret being combined. Worth a follow-up issue to either fix it or comment why it's intentional, but not for this PR.
- `TestResolveClientCredentials` doesn't cover the secretProvider success or error paths (the warn-and-fall-through branch). A `mockSecretProvider` already lives in this package — easy to add.

Separately: the `*bytes.Buffer` shared between exec's stream-copy goroutine and the test's `Eventually` polling is a latent data race that pre-dates this PR (came in with #5085). The new test adds more polling against it, which makes flakes a bit more likely. Worth wrapping in a tiny mutex-guarded snapshot helper at some point — `go test -race` should fault on it.

PR #5132, opened 2026-04-29
Carry forward upstream refresh token on re-authorization

### LINKED ISSUE #5047: Upstream refresh tokens orphaned on re-authorization

jhrozek:
## Summary

This PR addresses **Root Cause B** of [#5047](https://github.com/stacklok/toolhive/issues/5047): when a user re-authorizes through the auth server, the OAuth callback writes the IdP-supplied `RefreshToken` verbatim into a new `storage.UpstreamTokens` row. Real IdPs commonly **omit** `refresh_token` on a silent re-auth (Google without `prompt=consent`, Microsoft Entra when consent was already granted, GitHub/Okta/Auth0/Keycloak similarly). In every case the previously-issued RT remains valid at the IdP — but today the callback overwrites it with `""`, orphaning the prior good RT and making the next refresh attempt fail. That forces another re-auth, perpetuating the loop.

The fix mirrors the preservation pattern already in place in `upstreamTokenRefresher.RefreshAndStore`: if the IdP returns no RT, look up the most recent prior row for `(userID, providerID)` and carry the prior RT into the new row. A defensive `prior.UpstreamSubject == providerSubject` guard prevents carrying credentials across distinct upstream identities. Storage failures during the lookup are non-fatal — they log a warning and proceed with the empty RT, which is no worse than the current status quo.

- New storage method `UpstreamTokenStorage.GetLatestUpstreamTokensForUser(ctx, userID, providerID)` keyed on user+provider rather than the primary `(sessionID, providerName)` key. Implemented on both Memory and Redis backends; both backends pick the highest `ExpiresAt` so the carry-forward decision is consistent across deployment shapes. Expired rows are intentionally returned (callers want the RT regardless of access-token expiry).
- Callback wiring in `pkg/authserver/server/handlers/callback.go` extracted into `maybeCarryForwardRefreshToken` (gocyclo budget) and called between `storageTokens` construction and `StoreUpstreamTokens`.
- The Redis backend uses the existing `KeyTypeUserUpstream:{userID}` reverse-index set (already maintained atomically by `storeUpstreamTokensScript`), no new Lua script.
- End-to-end integration test against `mockoidc` + a thin RT-stripping reverse proxy (mockoidc unconditionally issues RTs for `authorization_code` and offers no API to suppress them).

**Out of scope, deferred to a follow-up PR**: Root Cause A — the storage rekey from `(sessionID, providerID)` to `(userID, providerID)` plus a one-time Redis migration. Trey's [comment on #5047](https://github.com/stacklok/toolhive/issues/5047#issuecomment-4336470125) describes the split; this PR is the smaller, surgical fix and is valuable on its own because it ends the orphaning loop even before Root Cause A lands.

Closes #5047 partially (Root Cause B; Root Cause A tracked separately).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Specifically verified:
- `go test -race -count=1 ./pkg/authserver/...` — all 13 packages pass.
- `go test -race -count=1 -v -run TestIntegration_ ./pkg/authserver/` — all 23 integration tests pass, including the new `TestIntegration_Callback_PreservesRefreshTokenOnReauth` and the existing `TestIntegration_FullFlow_NonExpiringUpstreamToken[_Redis]` and `TestIntegration_MultiUpstreamChain_MixedExpiryOrderings[_Redis]` from #5092.
- `task test` — pre-existing failure in `pkg/workloads/TestDefaultManager_ListWorkloadsInGroup/workloads_with_empty_group_names` is unrelated to this branch.

## API Compatibility

The new method on `storage.UpstreamTokenStorage` is an internal Go interface, not a CRD field. No CRD schema changes.

## Changes

| File | Change |
|------|--------|
| `pkg/authserver/storage/types.go` | Adds `GetLatestUpstreamTokensForUser` to the `UpstreamTokenStorage` interface with explicit godoc on liveness contract (expired rows are returned) and cross-identity caller responsibility. |
| `pkg/authserver/storage/memory.go` | Implements the new method on `MemoryStorage` — RLock'd iteration, deep-copy on hit, wrapped `ErrNotFound` on miss. |
| `pkg/authserver/storage/redis.go` | Implements the new method on `RedisStorage` using `SMembers` over the existing user-upstream reverse index, single `MGet`, deserialized-`ProviderID` filtering. Tolerates dangling SMEMBERs and the `null` deletion tombstone. |
| `pkg/authserver/storage/mocks/mock_storage.go` | Regenerated. |
| `pkg/authserver/server/handlers/callback.go` | New `maybeCarryForwardRefreshToken` helper invoked between struct construction and `StoreUpstreamTokens`. `UpstreamSubject` guard, non-fatal on storage error. |
| `pkg/authserver/storage/memory_test.go` | Table-driven test for the Memory backend (9 subtests including deep-copy assertion, latest-by-`ExpiresAt`, empty-input validation). |
| `pkg/authserver/storage/redis_test.go` | Table-driven test for the Redis backend (9 subtests including the dangling-SMEMBER stale-index case via miniredis `SAdd`, two-zero-`ExpiresAt` corner case). |
| `pkg/authserver/server/handlers/callback_test.go` | Table-driven `TestCallbackHandler_RefreshTokenCarryForward` consolidating the 5 callback scenarios (preserve, subject mismatch, fresh wins, no prior row, storage error). |
| `pkg/authserver/server/handlers/helpers_test.go` | New `withGetLatestUpstreamTokensError` option mirroring the existing `withStorePendingError` pattern; mock checks it first. |
| `pkg/authserver/integration_test.go` | `TestIntegration_Callback_PreservesRefreshTokenOnReauth` — full callback flow against mockoidc with an RT-stripping reverse proxy. |

Yes. Users who previously hit a refresh-token-orphaning loop on re-authorization (most commonly with Google, Entra, Okta, GitHub, Auth0, Keycloak — all of which omit `refresh_token` on silent re-auth) will see refreshes succeed instead of failing into another forced re-auth. The fix is silent: no configuration changes are required.

## Implementation plan

This PR was planned with Claude Code via a per-step `go-expert-developer` + MoE (`code-reviewer` + `go-security-reviewer` / `oauth-expert` / `redis-valkey-advisor`) workflow. Each step landed as a separate commit with independent review.

<details>
<summary>Approved plan (5 steps)</summary>

1. **Add `GetLatestUpstreamTokensForUser` to storage interface + stub implementations + mock regen.** Keeps the diff tiny so the interface surface is reviewed independently.
2. **Implement Memory backend + 9-case table-driven test.** Filters by `(UserID, ProviderID)`, picks highest `ExpiresAt`, returns deep copy.
3. **Implement Redis backend + 9-case table-driven test.** Uses the existing `KeyTypeUserUpstream` reverse set; SMembers → MGet → unmarshal → filter; tolerates stale index entries and `null` tombstones.
4. **Wire the OAuth callback handler + 5-case table-driven test.** Helper `maybeCarryForwardRefreshToken` between `storageTokens` construction and `StoreUpstreamTokens`. Non-fatal on storage error. `UpstreamSubject` guard.
5. **End-to-end integration test against mockoidc + RT-stripping reverse proxy.** Three legs: initial RT issued → strip flag on, second auth carries forward → third auth as different user gets `ErrNotFound` and stays empty.

Cross-project research (via MCP-mounted source): Hydra delegates upstream-token storage to its consent app and avoids the bug class. oauth2-proxy has the same blind-overwrite shape but sidesteps it via opaque-ticket session keys (orphaned rows TTL out instead of being clobbered) — their `IDToken` preservation pattern at `pkg/providers/oidc.go:140-180` is the explicit code-level analog we mirror for `RefreshToken`. infratographer/identity-api is RFC-8693 only, no broker.

</details>

- **This is Root Cause B only.** Root Cause A — the full storage rekey from `(sessionID, providerID)` to `(userID, providerID)` with a one-time Redis migration — is intentionally deferred to a follow-up PR. See Trey's plan in #5047. This fix is valuable on its own: even with Root Cause A unresolved, carrying the RT forward to the new sessionID lets the refresher succeed when the new access token expires, ending the re-auth cycle.
- The `UpstreamSubject == providerSubject` guard is **defense-in-depth**. Given the current identity model (`ResolveUser` keys identity by `(providerID, providerSubject)`), the same internal `UserID` cannot legitimately have two different upstream subjects on the same provider — same-provider account-linking is a future TODO in `storage/types.go`. The guard exists so that future code can't accidentally violate that.
- The integration test does **not** request `offline_access` because mockoidc's `ScopesSupported` rejects it as `invalid_scope`. mockoidc unconditionally issues RTs regardless of scope, so the regression assertion (leg 2 RT matches leg 1 RT) is unaffected — but reviewers should know the test does not exercise the realistic precondition that real IdPs gate RT issuance on `offline_access`.

## Large PR Justification

- the new integration test is the culprit

### DIFF
--- a/pkg/authserver/server/handlers/callback_test.go
+++ b/pkg/authserver/server/handlers/callback_test.go
@@ -884,6 +884,13 @@ func TestCallbackHandler_RefreshTokenCarryForward(t *testing.T) {
 			lookupErr:        errors.New("simulated storage failure"),
 			expectedStoredRT: "",
 		},
+		{
+			name:             "does not carry prior RT when prior RT is empty",
+			priorRow:         &priorRow{sessionID: "old-session", upstreamSubject: "user-123", refreshToken: ""},
+			idpRefreshToken:  "",
+			idpSubject:       "user-123",
+			expectedStoredRT: "",
+		},
 	}
 
 	for _, tc := range cases {
@@ -964,6 +971,12 @@ func TestCallbackHandler_RefreshTokenCarryForward(t *testing.T) {
 			newRow, ok := storState.upstreamTokens[newSessionID+":"+providerName]
 			require.True(t, ok, "token row for new session should be stored")
 			assert.Equal(t, tc.expectedStoredRT, newRow.RefreshToken)
+			// Sanity-check the rest of the row was written by the callback path so a
+			// regression that early-returns before StoreUpstreamTokens cannot pass.
+			assert.Equal(t, "new-access-token", newRow.AccessToken)
+			assert.Equal(t, "new-id-token", newRow.IDToken)
+			assert.Equal(t, tc.idpSubject, newRow.UpstreamSubject)
+			assert.False(t, newRow.ExpiresAt.IsZero(), "ExpiresAt must be populated")
 		})
 	}
 }
--- a/pkg/authserver/server/handlers/callback_test.go
+++ b/pkg/authserver/server/handlers/callback_test.go
@@ -826,6 +826,148 @@ func TestCallbackHandler_TwoUpstreams_StorePendingError_CleansUp(t *testing.T) {
 	}
 }
 
+// TestCallbackHandler_RefreshTokenCarryForward verifies the OAuth callback's
+// behavior when the upstream IdP omits refresh_token on re-authorization. The
+// handler looks up a prior (user, provider) row and carries the prior RT
+// forward, with a defensive UpstreamSubject guard against account-linking
+// edge cases. Storage errors during the lookup are non-fatal.
+func TestCallbackHandler_RefreshTokenCarryForward(t *testing.T) {
+	t.Parallel()
+
+	type priorRow struct {
+		sessionID       string
+		upstreamSubject string
+		refreshToken    string
+	}
+
+	cases := []struct {
+		name             string
+		priorRow         *priorRow // nil = no prior row
+		idpRefreshToken  string    // RT returned by upstream exchange
+		idpSubject       string    // subject claim returned by upstream
+		lookupErr        error     // if non-nil, storage lookup returns this error
+		expectedStoredRT string    // expected RefreshToken on the new row
+	}{
+		{
+			name:             "preserves prior RT when IdP omits it",
+			priorRow:         &priorRow{sessionID: "old-session", upstreamSubject: "user-123", refreshToken: "rt-prior"},
+			idpRefreshToken:  "",
+			idpSubject:       "user-123",
+			expectedStoredRT: "rt-prior",
+		},
+		{
+			name:             "no carry across different upstream subjects",
+			priorRow:         &priorRow{sessionID: "alice-session", upstreamSubject: "alice@idp", refreshToken: "rt-prior"},
+			idpRefreshToken:  "",
+			idpSubject:       "bob@idp",
+			expectedStoredRT: "",
+		},
+		{
+			name:             "fresh IdP RT wins",
+			priorRow:         &priorRow{sessionID: "old-session", upstreamSubject: "user-123", refreshToken: "rt-prior"},
+			idpRefreshToken:  "rt-fresh",
+			idpSubject:       "user-123",
+			expectedStoredRT: "rt-fresh",
+		},
+		{
+			name:             "no prior row accepts empty RT",
+			priorRow:         nil,
+			idpRefreshToken:  "",
+			idpSubject:       "user-123",
+			expectedStoredRT: "",
+		},
+		{
+			name:             "storage error during lookup is non-fatal",
+			priorRow:         nil,
+			idpRefreshToken:  "",
+			idpSubject:       "user-123",
+			lookupErr:        errors.New("simulated storage failure"),
+			expectedStoredRT: "",
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			const (
+				providerName   = "test-upstream"
+				existingUserID = "user-id"
+				newSessionID   = "new-session"
+			)
+
+			var opts []baseTestSetupOption
+			if tc.lookupErr != nil {
+				opts = append(opts, withGetLatestUpstreamTokensError(tc.lookupErr))
+			}
+			handler, storState, mockUpstream := handlerTestSetup(t, opts...)
+
+			mockUpstream.exchangeResult = &upstream.Identity{
+				Tokens: &upstream.Tokens{
+					AccessToken:  "new-access-token",
+					RefreshToken: tc.idpRefreshToken,
+					IDToken:      "new-id-token",
+					ExpiresAt:    time.Now().Add(time.Hour),
+				},
+				Subject: tc.idpSubject,
+			}
+
+			// Pre-populate user + identity so ResolveUser is deterministic.
+			storState.users[existingUserID] = &storage.User{
+				ID:        existingUserID,
+				CreatedAt: time.Now(),
+				UpdatedAt: time.Now(),
+			}
+			storState.providerIdentities[providerName+":"+tc.idpSubject] = &storage.ProviderIdentity{
+				UserID:          existingUserID,
+				ProviderID:      providerName,
+				ProviderSubject: tc.idpSubject,
+				LinkedAt:        time.Now(),
+				LastUsedAt:      time.Now(),
+			}
+
+			if tc.priorRow != nil {
+				storState.upstreamTokens[tc.priorRow.sessionID+":"+providerName] = &storage.UpstreamTokens{
+					ProviderID:      providerName,
+					AccessToken:     "old-access",
+					RefreshToken:    tc.priorRow.refreshToken,
+					ExpiresAt:       time.Now().Add(30 * time.Minute),
+					ClientID:        testAuthClientID,
+					UserID:          existingUserID,
+					UpstreamSubject: tc.priorRow.upstreamSubject,
+				}
+			}
+
+			internalState := testInternalState
+			storState.pendingAuths[internalState] = &storage.PendingAuthorization{
+				ClientID:             testAuthClientID,
+				RedirectURI:          testAuthRedirectURI,
+				State:                "client-state",
+				PKCEChallenge:        "challenge123",
+				PKCEMethod:           "S256",
+				Scopes:               []string{"openid"},
+				InternalState:        internalState,
+				UpstreamPKCEVerifier: "verifier-1234567890123456789012345678",
+				SessionID:            newSessionID,
+				UpstreamProviderName: providerName,
+				CreatedAt:            time.Now(),
+			}
+
+			req := httptest.NewRequest(http.MethodGet, "/oauth/callback?code=test-code&state="+internalState, nil)
+			rec := httptest.NewRecorder()
+			handler.CallbackHandler(rec, req)
+
+			require.Equal(t, http.StatusSeeOther, rec.Code)
+			assert.NotContains(t, rec.Header().Get("Location"), "error=")
+
+			newRow, ok := storState.upstreamTokens[newSessionID+":"+providerName]
+			require.True(t, ok, "token row for new session should be stored")
+			assert.Equal(t, tc.expectedStoredRT, newRow.RefreshToken)
+		})
+	}
+}
+
 func TestRoutesIncludeAuthorizeAndCallback(t *testing.T) {
 	t.Parallel()
 	handler, _, _ := handlerTestSetup(t)
--- a/pkg/authserver/storage/memory.go
+++ b/pkg/authserver/storage/memory.go
@@ -838,6 +838,23 @@ func (s *MemoryStorage) DeleteUpstreamTokens(_ context.Context, sessionID string
 	return nil
 }
 
+// compareExpiry orders ExpiresAt values for the GetLatestUpstreamTokensForUser
+// tie-breaker. Non-expiring rows (zero ExpiresAt — "alive forever") rank latest;
+// among finite expiries, later ranks latest. Mirrors time.Compare but with the
+// zero sentinel reinterpreted. Returns -1/0/+1.
+func compareExpiry(a, b time.Time) int {
+	aZero, bZero := a.IsZero(), b.IsZero()
+	switch {
+	case aZero && bZero:
+		return 0
+	case aZero:
+		return 1
+	case bZero:
+		return -1
+	}
+	return a.Compare(b)
+}
+
 // GetLatestUpstreamTokensForUser implements UpstreamTokenStorage.
 //
 // Expired tokens (past ExpiresAt) are returned so callers can use the refresh
@@ -859,7 +876,7 @@ func (s *MemoryStorage) GetLatestUpstreamTokensForUser(_ context.Context, userID
 		if entry.value == nil || entry.value.UserID != userID || entry.value.ProviderID != providerID {
 			continue
 		}
-		if winner == nil || entry.value.ExpiresAt.After(winner.ExpiresAt) {
+		if winner == nil || compareExpiry(entry.value.ExpiresAt, winner.ExpiresAt) > 0 {
 			winner = entry.value
 		}
 	}
--- a/pkg/authserver/storage/memory.go
+++ b/pkg/authserver/storage/memory.go
@@ -746,6 +746,24 @@ func (s *MemoryStorage) StoreUpstreamTokens(_ context.Context, sessionID, provid
 	return nil
 }
 
+// cloneUpstreamTokens returns a field-by-field copy of t, or nil if t is nil.
+func cloneUpstreamTokens(t *UpstreamTokens) *UpstreamTokens {
+	if t == nil {
+		return nil
+	}
+	return &UpstreamTokens{
+		ProviderID:       t.ProviderID,
+		AccessToken:      t.AccessToken,
+		RefreshToken:     t.RefreshToken,
+		IDToken:          t.IDToken,
+		ExpiresAt:        t.ExpiresAt,
+		SessionExpiresAt: t.SessionExpiresAt,
+		UserID:           t.UserID,
+		UpstreamSubject:  t.UpstreamSubject,
+		ClientID:         t.ClientID,
+	}
+}
+
 // GetUpstreamTokens retrieves the upstream IDP tokens for a session and provider.
 // Returns a defensive copy to prevent aliasing issues.
 func (s *MemoryStorage) GetUpstreamTokens(_ context.Context, sessionID, providerName string) (*UpstreamTokens, error) {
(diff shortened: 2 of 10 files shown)


pkg/authserver/storage/redis.go
```diff
@@ -1124,6 +1124,111 @@ func (s *RedisStorage) DeleteUpstreamTokens(ctx context.Context, sessionID strin
...
+	for i, val := range values {
+		stored, ok := parseUserUpstreamEntry(val, providerID, members[i])
+		if !ok {
+			continue
+		}
+
+		// Pick the row with the highest ExpiresAt. Zero ExpiresAt (the no-expiry
+		// sentinel) is represented as 0 — the smallest int64 — so it loses to any
+		// real timestamp unless it is the only candidate.
+		if winner == nil || stored.ExpiresAt > winner.ExpiresAt {
```
tgrunnagle: Question: should we prefer non-expiring tokens?

pkg/authserver/storage/redis.go
```diff
@@ -1124,6 +1124,111 @@ func (s *RedisStorage) DeleteUpstreamTokens(ctx context.Context, sessionID strin
...
+			winnerTokens = &UpstreamTokens{
+				ProviderID:      stored.ProviderID,
+				AccessToken:     stored.AccessToken,
+				RefreshToken:    stored.RefreshToken,
+				IDToken:         stored.IDToken,
+				ExpiresAt:       expiresAt,
+				UserID:          stored.UserID,
+				UpstreamSubject: stored.UpstreamSubject,
+				ClientID:        stored.ClientID,
+			}
```
tgrunnagle: nit: can this happen outside of the loop, referring to `winner` instead of `stored`?

tgrunnagle (commented):
## Multi-Agent Consensus Review (MEDIUM findings only)
**Agents consulted**: oauth-flow, security, redis-storage, test-coverage, go-style, general-quality (6 specialists, parallel review)

A full consensus review surfaced 18 actionable findings. Posting only the MEDIUM-severity items here; the LOW findings are polish and didn't feel worth the inline noise on a focused fix PR.

### MEDIUM Findings
| # | Finding | Consensus | Action |
|---|---------|-----------|--------|
| F1 | `SessionExpiresAt` dropped from returned record (Memory + Redis) | 10/10 | Fix — see inline suggestions |
| F7 | Missing handler test — prior row with empty `RefreshToken` | 9/10 | Add table case |

### Overall
The carry-forward design is sound and the implementation is correct for the documented use case. Mirroring `RefreshAndStore`'s preservation pattern at the cross-row level (with the `UpstreamSubject` defense-in-depth guard) is a sensible surgical fix for Root Cause B. Scope is one logical change, the integration-test harness (`rtStrippingProxy`) is genuine novel infrastructure, and the explicit deferral of Root Cause A is appropriate.

The latent bug worth fixing in this PR is **F1**: both backends silently omit `SessionExpiresAt` from the returned struct in `GetLatestUpstreamTokensForUser`. No production caller reads it today, but it's a contract-shaped trap for any future caller that uses this method as a substitute for `GetUpstreamTokens` — exactly the kind of expansion the deferred Root Cause A rekey will encourage. One-line fix per backend; suggestion blocks attached.

**F7** is a test gap: the production guard `prior.RefreshToken != ""` (callback.go:187) is currently untested. The case matters in real chains because the carry-forward chain itself can leave rows with empty RTs, so the guard is load-bearing.

---
*Generated with Claude Code*

pkg/authserver/storage/memory.go
```diff
@@ -846,6 +846,48 @@ func (s *MemoryStorage) DeleteUpstreamTokens(_ context.Context, sessionID string
...
+	if winner == nil {
+		return nil, fmt.Errorf("%w: %w", ErrNotFound, fosite.ErrNotFound.WithHint("Upstream tokens not found"))
+	}
+
+	return &UpstreamTokens{
+		ProviderID:      winner.ProviderID,
+		AccessToken:     winner.AccessToken,
+		RefreshToken:    winner.RefreshToken,
+		IDToken:         winner.IDToken,
+		ExpiresAt:       winner.ExpiresAt,
```
tgrunnagle: **F1 — `SessionExpiresAt` dropped from returned record** (10/10, MEDIUM)

The sibling `GetUpstreamTokens` (memory.go:773-783) populates `SessionExpiresAt` but this new method omits it. Today's only caller (`maybeCarryForwardRefreshToken`) reads only `RefreshToken` and `UpstreamSubject`, so the bug is latent — but the interface returns `*UpstreamTokens` with no "partial copy" caveat in the godoc, so any future caller (e.g. the deferred Root Cause A rekey) will get a silently-zero `SessionExpiresAt` and break the storage TTL invariant the refresher relies on (refresher.go:68-85).

Raised by: redis-storage, go-style, general-quality

```suggestion
		ExpiresAt:        winner.ExpiresAt,
		SessionExpiresAt: winner.SessionExpiresAt,
```

pkg/authserver/storage/redis.go
```diff
@@ -1124,6 +1124,111 @@ func (s *RedisStorage) DeleteUpstreamTokens(ctx context.Context, sessionID strin
...
+			var expiresAt time.Time
+			if stored.ExpiresAt != 0 {
+				expiresAt = time.Unix(stored.ExpiresAt, 0)
+			}
+			winnerTokens = &UpstreamTokens{
+				ProviderID:      stored.ProviderID,
+				AccessToken:     stored.AccessToken,
+				RefreshToken:    stored.RefreshToken,
+				IDToken:         stored.IDToken,
+				ExpiresAt:       expiresAt,
```
tgrunnagle: **F1 — `SessionExpiresAt` dropped from returned record** (10/10, MEDIUM)

Same issue as the Memory backend: `unmarshalUpstreamTokens` (redis.go:1163-1166) populates `SessionExpiresAt`, this method does not. Field-list duplication here means there are now four copies of the same struct construction in this package — extracting a `(s *storedUpstreamTokens).toUpstreamTokens()` helper would prevent further drift.

```suggestion
				ExpiresAt:       expiresAt,
				SessionExpiresAt: func() time.Time {
					if stored.SessionExpiresAt != 0 {
						return time.Unix(stored.SessionExpiresAt, 0)
					}
					return time.Time{}
				}(),
```

pkg/authserver/server/handlers/callback_test.go
```diff
@@ -767,6 +767,148 @@ func TestCallbackHandler_TwoUpstreams_StorePendingError_CleansUp(t *testing.T) {
...
+func TestCallbackHandler_RefreshTokenCarryForward(t *testing.T) {
+	t.Parallel()
+
+	type priorRow struct {
+		sessionID       string
+		upstreamSubject string
+		refreshToken    string
+	}
+
+	cases := []struct {
```
tgrunnagle, line 843: **F7 — Missing test case: prior row with empty `RefreshToken`** (9/10, MEDIUM)

The production guard `prior.RefreshToken != ""` (callback.go:187) is currently untested. The case matters in real chains: the carry-forward chain itself can leave rows with empty RTs (when leg N had no prior to carry from), so this guard is load-bearing. Suggest adding a table case:

```
{
    name:             "prior row with empty RT does not carry",
    priorRow:         &priorRow{sessionID: "old-session", upstreamSubject: "user-123", refreshToken: ""},
    idpRefreshToken:  "",
    idpSubject:       "user-123",
    expectedStoredRT: "",
},
```

Raised by: test-coverage, security

pkg/authserver/storage/memory.go
jhrozek: Fixed in c162013d. Also extracted a `cloneUpstreamTokens` helper in 6d1d2678 since this was the third site doing the same field-by-field copy and the next field addition would hit the same trap. Whole-struct round-trip test in 15391fd9 — a future drop fails loudly.

pkg/authserver/storage/redis.go
jhrozek: Fixed in c162013d. Extracted `(*storedUpstreamTokens).toUpstreamTokens()` in 6d1d2678 — kept it as a method on the stored type since Redis converts int64 epoch sentinels to time.Time, while Memory's `cloneUpstreamTokens` is a free function (different semantics). Round-trip test added there too.

pkg/authserver/server/handlers/callback_test.go
jhrozek: Added in 678e415a. While there, broadened the shared assertion block (AccessToken, IDToken, UpstreamSubject, non-zero ExpiresAt across all cases) — without those, a regression that early-returns from the callback before StoreUpstreamTokens would still pass an RT-only assertion.

pkg/authserver/storage/redis.go
jhrozek: Good catch — flipped in f564b0f8. After re-reading #5092, the old rule was inconsistent with the rest of the package (IsExpired, cleanupExpired, marshalUpstreamTokensWithTTL all treat zero ExpiresAt as "alive forever" for Slack and GitHub OAuth Apps). Carry-forward caller is largely unaffected today since those rows have empty RefreshToken and the guard rejects them anyway, but the flipped rule is the more sensible default for any future caller. Note that #5047 Scope A's rekey to (userID, providerID) collapses this to one row per pair and makes the whole tie-breaker vestigial — so this is also a stopgap on that road.

pkg/authserver/storage/redis.go
jhrozek: Done in 6d1d2678. Loop now tracks `winner *storedUpstreamTokens` and calls `winner.toUpstreamTokens()` once after the loop.

jhrozek:
Thanks for the review. Both MEDIUM findings and both nits addressed — replied on each thread with specifics.

tgrunnagle (approved):

PR #5133, opened 2026-04-29
Add core registry types: Entry, Registry, Store

rdimitrov:
## Summary

Why: `pkg/registry` has accumulated four near-duplicate `Provider` implementations, a factory singleton wrapping them, and several parallel CRUD shims. Replacing this incrementally requires a new, smaller surface for consumers to migrate onto without a flag day. This PR introduces that surface only — no behavior changes.

What:

- `Entry` — generic catalog item carrying a `Kind` and one optional payload pointer per kind. `Server` uses upstream `v0.ServerJSON`; `Skill` uses `toolhive-core` `types.Skill` (no upstream equivalent exists). Adding a new kind = one enum value + one field, no other surgery.
- `Registry` — minimal interface (`Name`, `Get`, `List`, `Search`) with a `Filter` for kind narrowing. Implementations must be safe for concurrent use.
- `Store` — in-memory `Registry` implementation. Will back the file/URL sources in the next PR.
- `ErrEntryNotFound` paired with the new types.

The new code lives alongside the existing `Provider` hierarchy and has zero callers; subsequent PRs migrate consumers and delete the old code.

Part of #4199.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring (no behavior change)
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Unit tests (\`task test\`)
- [x] Linting (\`task lint-fix\`)

## Does this introduce a user-facing change?

No.

## Special notes for reviewers

- The new \`Registry\` interface and the existing \`Provider\` interface coexist for now; they diverge in scope across the next PRs as consumers migrate to \`Registry\` and \`Provider\` is deleted.
- I deliberately scoped this PR to additive changes only — the single modification to \`errors.go\` is one new \`ErrEntryNotFound\` variable. Easy to revert if needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #4199, opened 2026-03-18
Add upstream v0.1 registry API endpoints to thv


rdimitrov:
Update the thv API to expose read-only endpoints following the upstream MCP registry v0.1 spec, aligning the Desktop experience with Cloud UI.

Scope:
- Add new route group (e.g. /v0.1/servers) parallel to existing /api/v1beta/registry
- Return upstream ServerJSON/ServerResponse types instead of ToolHive-specific types
- Support cursor-based pagination (limit, cursor)
- Support query parameters: search, version, updated_since, include_deleted
- Single registry only (no multi-registry or extensions for now)
- Read-only (no publish/auth endpoints)
- Same goes for the skills endpoints available under `v0.1/x`

Strategy:
- New endpoints live alongside existing ones — no breaking changes
- Once complete and stable, deprecate the old /api/v1beta/registry endpoints
- Reference implementation: toolhive-registry-server which already implements the spec

Out of scope (for now):
- Write endpoints (publish, edit, status)
- Auth endpoints
- Multiple registries
- Server versioning beyond "latest"

PR #5134, opened 2026-04-29
Add file, URL, and embedded registry sources

rdimitrov:
## Summary

Why: PR 1 introduces the `Registry` interface and `Store` but no implementations. This PR adds the bulk-load sources — file, URL, and embedded — that the manager in PR 3 will own. These are the simple cases: load once at construction, serve from memory. The API source (proxy semantics, per-registry expansion) is split off into a follow-up PR for a focused review.

What:

- `parse.go` — shared parser. Decodes `toolhivetypes.UpstreamRegistry` JSON into `[]*Entry`. Detects the legacy ToolHive format up front via `legacyhint` and returns `ErrLegacyFormat` carrying the migration message — Go's JSON decoder silently produces an empty registry from legacy input, so we have to short-circuit.
- `file_registry.go` — `NewFileRegistry(name, path)` reads the file and hands bytes to the parser.
- `url_registry.go` — `NewURLRegistry(ctx, name, url, client)` does an HTTP GET, caps the body at 32 MiB, parses. Network errors map to `UnavailableError` so handlers can translate to 503.
- `embedded_registry.go` — `NewEmbeddedRegistry(name)` reads the bytes baked into the binary via `toolhive-catalog`.

Each loads at construction; runtime config changes still require restart in v1 (matches the broader simplification).

Stacked on PR #5133 — base set accordingly.

Part of #4199.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

No.

- Bulk-load (read-once-at-construction) is a deliberate departure from today's \`LocalRegistryProvider\` (which re-reads on every \`GetRegistry\` call). With multi-active fan-out, repeating disk I/O per query is wasteful, and config-mutability is restart-only in v1.
- API source (\`api_registry.go\`) is the obvious missing piece. It's in its own follow-up PR because it has different semantics (proxy, expansion to N peer registries via \`GET /v1/registries\`) and should not be reviewed in the same diff as bulk-load.
- The 32 MiB body cap in url_registry.go is a defensive default. Adjustable in future via an option if anyone hits it.
- The \`#nosec G304\` annotation on \`os.ReadFile(path)\` is intentional — reading a user-supplied path is the entire purpose of a file source.

### DIFF
--- a/pkg/registry/embedded_registry.go
+++ b/pkg/registry/embedded_registry.go
@@ -1,0 +1,33 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package registry
+
+import (
+	"fmt"
+
+	catalog "github.com/stacklok/toolhive-catalog/pkg/catalog/toolhive"
+)
+
+// EmbeddedRegistryName is the canonical name for the registry backed by the
+// asset baked into the binary. It is what `thv registry list` displays for
+// the default-shipped entry.
+const EmbeddedRegistryName = "embedded"
+
+// NewEmbeddedRegistry constructs a Registry from the registry data baked
+// into the binary at build time. Used as the default registry in fresh
+// installs.
+//
+// The optional name override lets callers override the default
+// EmbeddedRegistryName — useful only for tests; production uses the
+// constant.
+func NewEmbeddedRegistry(name string) (Registry, error) {
+	if name == "" {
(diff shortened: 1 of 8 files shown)


rdimitrov:
Closing temporarily — moving to a local-only workflow until the full registry simplification is verified end-to-end. Branch and commits preserved; will reopen as part of the final stack.

PR #5142, opened 2026-04-30
Correct Gemini CLI LLM gateway config to proxy mode

### LINKED ISSUE #5141: Add Gemini CLI support to `thv llm setup`
## Bug description

`thv llm setup` does not support Gemini CLI. When trying to manually configure Gemini CLI to work with the LLM gateway, the natural guess at the settings format is wrong in two ways:

**Incorrect manual config (what users try):**
```json
{
    "auth": {"tokenCommand": "\"/path/to/thv\" llm token"},
    "baseUrl": "https://gateway:30443"
}
```

Both keys are invalid:
- `auth.tokenCommand` — not a recognised Gemini CLI `settings.json` key (this is Claude Code's `apiKeyHelper` concept; Gemini CLI has no equivalent)
- `baseUrl` — not a valid top-level key; silently ignored by Gemini CLI

## Expected behavior

`thv llm setup` detects Gemini CLI and writes the correct `~/.gemini/settings.json` using the proxy path (same as Cursor), since Gemini CLI has no dynamic token command support:

```json
{
    "security": {
        "auth": {
            "selectedType": "gemini-api-key"
        }
    },
    "env": {
        "GEMINI_API_KEY": "sk-toolhive-proxy",
        "GOOGLE_GEMINI_BASE_URL": "http://127.0.0.1:14000",
        "NODE_TLS_REJECT_UNAUTHORIZED": "0"
    }
}
```

Key points:
- `security.auth.selectedType: "gemini-api-key"` — required for `GOOGLE_GEMINI_BASE_URL` to be honoured (the fix in [PR #25357](https://github.com/google-gemini/gemini-cli/pull/25357), shipped in v0.40.0, only applies to API key auth, not OAuth)
- `GEMINI_API_KEY` — dummy value; the local proxy does not validate it but Gemini CLI requires a non-empty key
- `GOOGLE_GEMINI_BASE_URL` — points to the local `thv llm proxy` (default `http://127.0.0.1:14000`), which injects the real OIDC token before forwarding to the upstream gateway

## Actual behavior

No Gemini CLI entry in `Registry()` in `pkg/llm/setup.go`; users fall back to manual config with incorrect keys.

## Additional context

- Gemini CLI detectable via `~/.gemini/` directory
- Config written to `~/.gemini/settings.json`
- Implementation would follow the `ToolKindProxy` pattern already used for Cursor (see `cursorSpec()` in `pkg/llm/setup.go`)
- `NODE_TLS_REJECT_UNAUTHORIZED` should only be written when `TLSSkipVerify` is set, same as the Claude Code spec
- **Gateway format note:** the Envoy AI Gateway currently only exposes an OpenAI-compatible endpoint (`/v1/chat/completions`); there is no Gemini API format backend configured. Gemini CLI sends requests to `/v1beta/models/{model}:generateContent`, which the gateway rejects. A separate issue should track adding a Google/Gemini backend to the gateway so Gemini CLI requests can actually be served end-to-end.

yrobla:
## Summary

Gemini CLI has no dynamic token-command equivalent (no apiKeyHelper analogue), so it must use proxy mode like Cursor rather than direct mode.

The previous config used two invalid settings.json keys:
- /auth/tokenCommand  — not a recognised Gemini CLI field
- /baseUrl            — silently ignored as a top-level key

Replace with the correct proxy-mode keys:
- /security/auth/selectedType = "gemini-api-key" (required so that GOOGLE_GEMINI_BASE_URL is honoured; OAuth auth ignores the override, fixed in gemini-cli v0.40.0 PR #25357)
- /env/GEMINI_API_KEY = placeholder key (proxy does not validate it)
- /env/GOOGLE_GEMINI_BASE_URL = local proxy address

Fixes #5141

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## API Compatibility

## Changes

| File | Change |
|------|--------|
|      |        |

## Implementation plan

<details>

</details>

### DIFF
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -161,14 +161,20 @@ const (
 // literals by hujson.Patch.
 // ValueField names which LLMApplyConfig field to write: "GatewayURL",
 // "ProxyBaseURL", "TokenHelperCommand", "PlaceholderAPIKey" (constant "thv-proxy"),
-// or "NodeTLSRejectUnauthorized" (writes "0" when TLSSkipVerify is true).
+// "NodeTLSRejectUnauthorized" (writes "0" when TLSSkipVerify is true), or
+// "ProxyOrigin" (like ProxyBaseURL but with the path stripped — use for tools
+// such as Gemini CLI that append their own API path to avoid doubled segments).
+// Any other string is written as a literal value; use this to hard-code a
+// constant string into a settings key (e.g. forcing an auth type to a fixed value).
 // ClearWhenEmpty: when true and the resolved value is empty, the key is removed
 // from the settings file rather than skipped. Use for conditional keys like
 // NODE_TLS_REJECT_UNAUTHORIZED that must be cleaned up when the flag is cleared.
 type LLMGatewayKeySpec struct {
-	JSONPointer    string // RFC 6901 path
-	ValueField     string // "GatewayURL" | "ProxyBaseURL" | "TokenHelperCommand" | "PlaceholderAPIKey" | "NodeTLSRejectUnauthorized"
-	ClearWhenEmpty bool   // remove the key when the resolved value is empty
+	JSONPointer string // RFC 6901 path
+	// ValueField: "GatewayURL" | "ProxyBaseURL" | "ProxyOrigin" |
+	// "TokenHelperCommand" | "PlaceholderAPIKey" | "NodeTLSRejectUnauthorized"
+	ValueField     string
+	ClearWhenEmpty bool // remove the key when the resolved value is empty
 }
 
 // clientAppConfig represents a configuration path for a supported MCP client.
@@ -839,17 +845,28 @@ var supportedClientIntegrations = []clientAppConfig{
 		SupportsSkills:    true,
 		SkillsGlobalPath:  []string{".agents", skillsDirName},
 		SkillsProjectPath: []string{".agents", skillsDirName},
-		// LLM gateway: patches the same settings.json used for MCP
-		LLMGatewayMode:     "direct",
+		// LLM gateway: patches the same settings.json used for MCP.
+		// Gemini CLI has no dynamic token-command equivalent, so it uses the
+		// proxy path. GOOGLE_GEMINI_BASE_URL is only honoured when
+		// security.auth.selectedType is "gemini-api-key" (fixed in
+		// gemini-cli v0.40.0, PR #25357); OAuth auth ignores the override.
+		//
+		// NODE_TLS_REJECT_UNAUTHORIZED is intentionally omitted: in proxy mode
+		// the tool connects to the local proxy over plain HTTP, so there is no
+		// TLS handshake on that leg. Setting the env var would globally disable
+		// TLS verification for all other HTTPS requests the Gemini CLI process
+		// makes, which is an unacceptable side-effect.
+		LLMGatewayMode:     "proxy",
 		LLMBinaryName:      "gemini",
 		LLMSettingsFile:    "settings.json",
 		LLMSettingsRelPath: []string{".gemini"},
 		LLMGatewayKeys: []LLMGatewayKeySpec{
-			{JSONPointer: "/auth/tokenCommand", ValueField: "TokenHelperCommand"},
-			{JSONPointer: "/baseUrl", ValueField: "GatewayURL"},
-			// NODE_TLS_REJECT_UNAUTHORIZED is only written when --tls-skip-verify is set.
-			// ClearWhenEmpty ensures it is removed when the flag is later cleared.
-			{JSONPointer: "/env/NODE_TLS_REJECT_UNAUTHORIZED", ValueField: "NodeTLSRejectUnauthorized", ClearWhenEmpty: true},
+			// Force API-key auth so GOOGLE_GEMINI_BASE_URL is respected.
+			// "gemini-api-key" is an unrecognised ValueField, so llmValueForSpec
+			// writes it as a literal string (see LLMGatewayKeySpec doc comment).
+			{JSONPointer: "/security/auth/selectedType", ValueField: "gemini-api-key"},
+			{JSONPointer: "/env/GEMINI_API_KEY", ValueField: "PlaceholderAPIKey"},
+			{JSONPointer: "/env/GOOGLE_GEMINI_BASE_URL", ValueField: "ProxyOrigin"},
 		},
 	},
 	{
--- a/pkg/client/config.go
+++ b/pkg/client/config.go
@@ -159,22 +159,29 @@ const (
 // path (e.g. "/apiKeyHelper" or "/env/ANTHROPIC_BASE_URL"). Dots in flat
 // top-level key names (e.g. "/cursor.general.openAIBaseURL") are treated as
 // literals by hujson.Patch.
-// ValueField names which LLMApplyConfig field to write: "GatewayURL",
-// "ProxyBaseURL", "TokenHelperCommand", "PlaceholderAPIKey" (constant "thv-proxy"),
-// "NodeTLSRejectUnauthorized" (writes "0" when TLSSkipVerify is true), or
-// "ProxyOrigin" (like ProxyBaseURL but with the path stripped — use for tools
-// such as Gemini CLI that append their own API path to avoid doubled segments).
-// Any other string is written as a literal value; use this to hard-code a
-// constant string into a settings key (e.g. forcing an auth type to a fixed value).
-// ClearWhenEmpty: when true and the resolved value is empty, the key is removed
-// from the settings file rather than skipped. Use for conditional keys like
-// NODE_TLS_REJECT_UNAUTHORIZED that must be cleaned up when the flag is cleared.
+//
+// Exactly one of ValueField or Literal must be set:
+//   - ValueField names which ApplyConfig field to write. Valid values:
+//     "GatewayURL", "ProxyBaseURL", "ProxyOrigin", "TokenHelperCommand",
+//     "PlaceholderAPIKey", "NodeTLSRejectUnauthorized".
+//     An unrecognised ValueField is a programming error and causes
+//     ConfigureLLMGateway to return an error.
+//   - Literal is written verbatim into the settings key (e.g. a fixed auth
+//     type string). Use Literal instead of ValueField for constant values so
+//     that typos in ValueField are caught as errors rather than silently
+//     written as unexpected strings.
+//
+// ClearWhenEmpty: when true and the resolved value is empty, the key is
+// removed from the settings file rather than skipped. Use for conditional
+// keys like NODE_TLS_REJECT_UNAUTHORIZED that must be cleaned up when the
+// flag is cleared. Ignored when Literal is set (literals are never empty).
 type LLMGatewayKeySpec struct {
 	JSONPointer string // RFC 6901 path
 	// ValueField: "GatewayURL" | "ProxyBaseURL" | "ProxyOrigin" |
 	// "TokenHelperCommand" | "PlaceholderAPIKey" | "NodeTLSRejectUnauthorized"
 	ValueField     string
-	ClearWhenEmpty bool // remove the key when the resolved value is empty
+	Literal        string // constant value written verbatim; mutually exclusive with ValueField
+	ClearWhenEmpty bool   // remove the key when the resolved value is empty (ignored for Literal)
 }
 
 // clientAppConfig represents a configuration path for a supported MCP client.
@@ -229,6 +236,10 @@ type clientAppConfig struct {
 	// LLMGatewayKeys lists the JSON Pointer paths and value-field mappings to
 	// apply when setting up (or reverting) LLM gateway access.
 	LLMGatewayKeys []LLMGatewayKeySpec
+	// LLMSetupNote is an optional message printed to stdout after a successful
+	// "thv llm setup" for this client. Use it to surface manual steps that
+	// toolhive cannot automate (e.g. env vars the tool reads from a .env file).
+	LLMSetupNote string
 }
 
 // extractServersKeyFromConfig extracts the servers key from MCPServersPathPrefix
@@ -862,10 +873,11 @@ var supportedClientIntegrations = []clientAppConfig{
 		LLMSettingsRelPath: []string{".gemini"},
 		LLMGatewayKeys: []LLMGatewayKeySpec{
 			// Force API-key auth so GOOGLE_GEMINI_BASE_URL is respected.
-			// "gemini-api-key" is an unrecognised ValueField, so llmValueForSpec
-			// writes it as a literal string (see LLMGatewayKeySpec doc comment).
-			{JSONPointer: "/security/auth/selectedType", ValueField: "gemini-api-key"},
-			{JSONPointer: "/env/GEMINI_API_KEY", ValueField: "PlaceholderAPIKey"},
+			{JSONPointer: "/security/auth/selectedType", Literal: "gemini-api-key"},
+			// Placeholder API key (proxy does not validate it).
+			{JSONPointer: "/env/GEMINI_API_KEY", Literal: llmPlaceholderAPIKey},
+			// Proxy origin for Gemini CLI; path is stripped because Gemini
+			// CLI appends its own /v1beta/... path to the base URL.
 			{JSONPointer: "/env/GOOGLE_GEMINI_BASE_URL", ValueField: "ProxyOrigin"},
 		},
 	},
--- a/pkg/client/llm_gateway.go
+++ b/pkg/client/llm_gateway.go
@@ -261,6 +261,10 @@ func (cm *ClientManager) buildLLMSettingsPath(cfg *clientAppConfig) string {
 // llmValueForSpec returns the config value corresponding to the ValueField name.
 // For "NodeTLSRejectUnauthorized", returns "0" when TLSSkipVerify is true, or ""
 // when false (which triggers removal when ClearWhenEmpty is set on the spec).
+// For "ProxyOrigin", returns ProxyBaseURL with the path stripped (origin only),
+// for tools like Gemini CLI that append their own API path to avoid doubled segments.
+// Any unrecognised ValueField string is returned as-is, allowing a spec entry to
+// hard-code a constant value into a settings key (e.g. forcing an auth type).
 func llmValueForSpec(valueField string, cfg llmgateway.ApplyConfig) string {
 	switch valueField {
 	case "GatewayURL":
@@ -276,6 +280,9 @@ func llmValueForSpec(valueField string, cfg llmgateway.ApplyConfig) string {
 			return "0"
 		}
 		return ""
+	case "ProxyOrigin":
+		// Like ProxyBaseURL but with the path stripped; see llmgateway.ProxyOriginOf.
+		return llmgateway.ProxyOriginOf(cfg.ProxyBaseURL)
 	default:
 		return valueField // treat unknown field names as literal values
 	}
--- a/pkg/client/llm_gateway.go
+++ b/pkg/client/llm_gateway.go
@@ -80,9 +80,20 @@ func (cm *ClientManager) ConfigureLLMGateway(clientType ClientApp, cfg llmgatewa
 // allowing conditional keys (e.g. NODE_TLS_REJECT_UNAUTHORIZED) to be cleaned
 // up when the associated flag is cleared.
 func applyLLMGatewayKeys(v *hujson.Value, specs []LLMGatewayKeySpec, cfg llmgateway.ApplyConfig, filePath string) error {
+	// Resolve all values up front so an unknown ValueField fails fast before
+	// any file is modified.
+	values := make([]string, len(specs))
+	for i, spec := range specs {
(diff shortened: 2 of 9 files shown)


pkg/client/config.go
```diff
@@ -839,17 +845,28 @@ var supportedClientIntegrations = []clientAppConfig{
...
-			{JSONPointer: "/baseUrl", ValueField: "GatewayURL"},
-			// NODE_TLS_REJECT_UNAUTHORIZED is only written when --tls-skip-verify is set.
-			// ClearWhenEmpty ensures it is removed when the flag is later cleared.
-			{JSONPointer: "/env/NODE_TLS_REJECT_UNAUTHORIZED", ValueField: "NodeTLSRejectUnauthorized", ClearWhenEmpty: true},
+			// Force API-key auth so GOOGLE_GEMINI_BASE_URL is respected.
+			// "gemini-api-key" is an unrecognised ValueField, so llmValueForSpec
+			// writes it as a literal string (see LLMGatewayKeySpec doc comment).
+			{JSONPointer: "/security/auth/selectedType", ValueField: "gemini-api-key"},
+			{JSONPointer: "/env/GEMINI_API_KEY", ValueField: "PlaceholderAPIKey"},
+			{JSONPointer: "/env/GOOGLE_GEMINI_BASE_URL", ValueField: "ProxyOrigin"},
```
JAORMX: These `/env/*` keys won't actually inject env vars into the Gemini CLI process.

- Gemini CLI's [settings.schema.json](https://raw.githubusercontent.com/google-gemini/gemini-cli/main/schemas/settings.schema.json) declares `additionalProperties: false` at the root and has no top-level `env` field. The only `env` field is at `$defs.MCPServerConfig.properties.env`, used to inject env vars into MCP server child processes Gemini CLI spawns... not into Gemini CLI itself.
- In [`packages/core/src/core/contentGenerator.ts`](https://github.com/google-gemini/gemini-cli/blob/main/packages/core/src/core/contentGenerator.ts), both `GEMINI_API_KEY` and `GOOGLE_GEMINI_BASE_URL` are read straight off `process.env`. There's no settings.json to process.env wiring.
- The v0.40.0 release note for PR #25357 says `fix(core): honor GOOGLE_GEMINI_BASE_URL`... lines up with the env-var read path, not a settings field.

Net effect: `thv llm setup --client gemini-cli` produces a settings file Gemini CLI silently ignores, so it keeps hitting the real Google API. The unit tests pass because they only check what we write to disk, not what Gemini CLI reads back.

The right place for these is `~/.gemini/.env` (Gemini CLI loads `$GEMINI_DIR/.env` and cwd `.env`), or OS-level env, or a wrapper script. None fit the `LLMGatewayKeys` JSON-pointer-patching mechanism.

Two paths forward:

1. Extend `clientAppConfig` with a `.env`-file mechanism aimed at `~/.gemini/.env`. Pair with the gateway-side `/v1beta/...` issue you flagged in #5141 (the Envoy AI Gateway only exposes `/v1/chat/completions`, so even with envs working we'd still get rejected upstream).
2. Drop the `/env/*` keys here, keep only `/security/auth/selectedType = "gemini-api-key"`, and surface a "manual step required: set `GEMINI_API_KEY` and `GOOGLE_GEMINI_BASE_URL` in `~/.gemini/.env`" message. Honest partial support, follow-up issue for the rest.

pkg/llmgateway/config.go
```diff
@@ -5,6 +5,27 @@
...
+	if err != nil || u == nil {
+		return rawURL
+	}
+	u.Path = ""
+	u.RawPath = ""
+	u.RawQuery = ""
+	u.ForceQuery = false
+	u.Fragment = ""
+	u.RawFragment = ""
+	return u.String()
```
JAORMX: `ProxyOriginOf` mishandles scheme-less and opaque inputs.

`url.Parse("localhost:14000/v1")` parses as `Scheme=localhost, Opaque=14000/v1`. The function clears `Path` but never `Opaque`, so the result still has the path. The doc comment claims "only the scheme and host remain"... not what the code delivers. The `"::invalid"` test case only catches the parse-error branch.

Suggestion: validate `u.Host != ""` and either fall back to `rawURL` or build the output as `u.Scheme + "://" + u.Host`. While we're at it, set `u.User = nil` so userinfo doesn't get persisted to a user-readable settings file if `ProxyBaseURL` ever has it (matches the function's stated intent).

Worth adding a dedicated `pkg/llmgateway/config_test.go` with cases for: empty input, opaque URL, scheme-less `host:port`, userinfo, IPv6 host, fragment-only.

pkg/client/llm_gateway.go
```diff
@@ -276,6 +280,9 @@ func llmValueForSpec(valueField string, cfg llmgateway.ApplyConfig) string {
 			return "0"
 		}
 		return ""
+	case "ProxyOrigin":
+		// Like ProxyBaseURL but with the path stripped; see llmgateway.ProxyOriginOf.
+		return llmgateway.ProxyOriginOf(cfg.ProxyBaseURL)
 	default:
 		return valueField // treat unknown field names as literal values
```
JAORMX: The `default:` literal-string fallback is a foot-gun.

A typo in any known `ValueField` (say `"GatwayURL"` for `GatewayURL`) silently writes the typo as a literal string into the user's settings file. No error signal, no test catches it... config corruption with no signal. The new `gemini-api-key` use relies on this same path, so it codifies the behavior.

Suggestion: add an explicit `Literal string` field on `LLMGatewayKeySpec` and validate `ValueField` against the closed set of names. The Gemini entry becomes `{JSONPointer: "/security/auth/selectedType", Literal: "gemini-api-key"}`, unknown `ValueField` values panic/error at startup, and typos can no longer slip through.

If you keep the current design, the doc comment needs to be much louder about the foot-gun (something like "WARNING: typos in known ValueField names will silently be written as literals") so a future maintainer doesn't "fix" the missing case to a typed enum and break the only intentional use.

pkg/client/llm_gateway_test.go
```diff
@@ -57,9 +57,12 @@ func TestRealClientConfigs_ConfigureAndRevert(t *testing.T) {
 		},
 		{
 			// ~/.gemini/settings.json
+			// NODE_TLS_REJECT_UNAUTHORIZED must NOT appear when TLSSkipVerify is false.
 			clientType: GeminiCli,
 			expectedKeys: []string{
-				"tokenCommand", "baseUrl", "https://gw.example.com",
+				"selectedType", "gemini-api-key",
+				"GEMINI_API_KEY", "thv-proxy",
+				"GOOGLE_GEMINI_BASE_URL", "http://localhost:14000",
```
JAORMX: `TestRealClientConfigs_ConfigureAndRevert` uses `assert.Contains` on the file body, so it would pass even if `gemini-api-key` landed at a wrong pointer (or as a key name elsewhere) and `GOOGLE_GEMINI_BASE_URL` had the wrong value.

Worth parsing the resulting JSON and reading values via `jsonPointerGet` (or equivalent), e.g. asserting `result./security/auth/selectedType == "gemini-api-key"` and `result./env/GOOGLE_GEMINI_BASE_URL == "http://localhost:14000"`. That way a future refactor moving these values gets caught.

(Less critical given the blocker on the `/env/*` approach, but still worth fixing once the env mechanism is sorted.)

JAORMX (commented):
Did some digging on the gemini-cli side and ran into a problem. The `/env/*` JSON pointers being written to `~/.gemini/settings.json` don't actually inject env vars (Gemini CLI's settings schema declares `additionalProperties: false` at the root and has no top-level `env` key... only `MCPServerConfig.env` for child processes). So this PR won't redirect Gemini CLI to the local proxy end to end. Details and two suggested paths forward in the inline comment on the `LLMGatewayKeys` block.

A few smaller things on `ProxyOriginOf` correctness, the literal-string fallback in `llmValueForSpec`, and a test-coverage gap, also inline.

pkg/client/config.go
yrobla: ok may be because i tested just locally, but i could access gemini endpoint properly... anyway going with option 2 for now

yrobla:
@copilot review

pkg/client/config.go
JAORMX: Yeah, easy false positive there. The two likely confounders if you want to verify cleanly: `GOOGLE_GEMINI_BASE_URL` already in shell env from prior testing, or a stale `~/.gemini/.env` (gemini-cli's `findEnvFile` walks up to home and uses the first match). `unset GOOGLE_GEMINI_BASE_URL && mv ~/.gemini/.env ~/.gemini/.env.bak 2>/dev/null` and retest if you're curious. Option 2 is the right call either way.

yrobla: ok i will implement in a follow up https://github.com/stacklok/toolhive/issues/5173

test/e2e/cli_llm_all_clients_test.go
```diff
@@ -448,11 +449,11 @@ var _ = Describe("thv llm — all-client matrix", Label("cli", "llm", "clients",
 			Expect(err).ToNot(HaveOccurred())
 			var geminiAfter map[string]any
 			Expect(json.Unmarshal(data, &geminiAfter)).To(Succeed())
-			tokenCmd, found := jsonPointerGet(geminiAfter, "/auth/tokenCommand")
+			baseURL, found := jsonPointerGet(geminiAfter, "/env/GOOGLE_GEMINI_BASE_URL")
```
JAORMX, line 458: This assertion is now stale and will fail when the e2e suite runs. We no longer write `/env/GOOGLE_GEMINI_BASE_URL` for gemini-cli, so the pointer won't be found and `Expect(found).To(BeTrue())` will fail.

The block's intent is to verify gemini-cli settings survive a claude-code revert. The right replacement is to check the only key we still patch:

```go
selectedType, found := jsonPointerGet(geminiAfter, "/security/auth/selectedType")
Expect(found).To(BeTrue(),
    "gemini-cli /security/auth/selectedType should still be present after claude-code teardown")
Expect(selectedType).To(Equal("gemini-api-key"),
    "gemini-cli selectedType should still be 'gemini-api-key' after claude-code teardown")
```

pkg/client/config.go
```diff
@@ -839,18 +856,36 @@ var supportedClientIntegrations = []clientAppConfig{
...
-		},
+			// Force API-key auth so GOOGLE_GEMINI_BASE_URL is respected.
+			{JSONPointer: "/security/auth/selectedType", Literal: "gemini-api-key"},
+		},
+		LLMSetupNote: "Note (gemini-cli): settings.json cannot inject env vars into the Gemini CLI process.\n" +
+			"To complete proxy setup, add the following to ~/.gemini/.env:\n" +
+			"  GEMINI_API_KEY=thv-proxy\n" +
+			"  GOOGLE_GEMINI_BASE_URL=<proxy-origin>  # e.g. http://localhost:14000\n" +
+			"Full end-to-end proxy routing requires gateway support for the /v1beta/* path; " +
+			"track progress in the follow-up issue.",
```
JAORMX: Two small things on the note:

1. "track progress in the follow-up issue" doesn't link the issue. You already have #5173 for the .env mechanism... worth embedding `https://github.com/stacklok/toolhive/issues/5173` here so users can find it from the CLI output.
2. The example port `http://localhost:14000` is hardcoded but `EffectiveProxyPort()` is configurable. If a user runs the proxy on a non-default port, the note tells them the wrong value. Either drop the `e.g.` example to keep it generic, or thread the actual `proxyBaseURL` through `LLMSetupNote` (probably overkill for this PR... up to you).

JAORMX (commented):
Round 2: all four threads from round 1 addressed cleanly. One stale e2e assertion that'll break the suite, plus two small nits on the new code.

pkg/client/config.go
```diff
@@ -159,16 +159,29 @@ const (
...
 // top-level key names (e.g. "/cursor.general.openAIBaseURL") are treated as
 // literals by hujson.Patch.
-// ValueField names which LLMApplyConfig field to write: "GatewayURL",
-// "ProxyBaseURL", "TokenHelperCommand", "PlaceholderAPIKey" (constant "thv-proxy"),
-// or "NodeTLSRejectUnauthorized" (writes "0" when TLSSkipVerify is true).
-// ClearWhenEmpty: when true and the resolved value is empty, the key is removed
-// from the settings file rather than skipped. Use for conditional keys like
-// NODE_TLS_REJECT_UNAUTHORIZED that must be cleaned up when the flag is cleared.
+//
+// Exactly one of ValueField or Literal must be set:
```
JAORMX, line 163: Nit: the doc comment says `Exactly one of ValueField or Literal must be set`, but `llmValueForSpec` only enforces "at least one" (Literal wins if both are set, no validation). Either:

- Tighten the code: error when both `ValueField != ""` and `Literal != ""`, so the contract matches the doc, or
- Loosen the doc: `If Literal is set it takes precedence; otherwise ValueField is resolved`.

Not a blocker, but as written a future maintainer reading the doc might assume mutual exclusion is enforced and rely on it.

JAORMX (dismissed):
Approving on the basis that the round-1 findings landed cleanly. The three round-2 items (stale `/env/GOOGLE_GEMINI_BASE_URL` assertion in the e2e teardown test, missing `#5173` link in `LLMSetupNote`, and the doc/code mutual-exclusion mismatch on `ValueField` / `Literal`) are non-blocking and will be picked up as follow-up.

Worth getting the e2e fix in soon so the Ginkgo suite stays green when it runs.

pkg/llmgateway/config.go
yrobla: Tracked in #5176.

pkg/client/llm_gateway.go
```diff
@@ -276,6 +281,20 @@ func llmValueForSpec(valueField string, cfg llmgateway.ApplyConfig) string {
...
+		// Gemini CLI that append their own API path (e.g. /v1beta/...).
+		// ProxyBaseURL is typically "http://localhost:<port>/v1", so using it
+		// directly would produce doubled paths like "/v1/v1beta/...".
+		u, err := url.Parse(cfg.ProxyBaseURL)
+		if err != nil || u == nil {
+			return cfg.ProxyBaseURL
+		}
+		u.Path = ""
+		u.RawPath = ""
+		u.RawQuery = ""
```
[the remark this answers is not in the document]
yrobla: Tracked in #5176.

pkg/client/llm_gateway.go
```diff
@@ -283,6 +284,19 @@ func llmValueForSpec(valueField string, cfg LLMApplyConfig) string {
...
```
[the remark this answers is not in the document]
yrobla: Tracked in #5176.

pkg/client/llm_gateway.go
yrobla: Tracked in #5177.

pkg/client/config.go
```diff
@@ -836,17 +836,20 @@ var supportedClientIntegrations = []clientAppConfig{
...
 		LLMGatewayKeys: []LLMGatewayKeySpec{
-			{JSONPointer: "/auth/tokenCommand", ValueField: "TokenHelperCommand"},
-			{JSONPointer: "/baseUrl", ValueField: "GatewayURL"},
-			// NODE_TLS_REJECT_UNAUTHORIZED is only written when --tls-skip-verify is set.
-			// ClearWhenEmpty ensures it is removed when the flag is later cleared.
-			{JSONPointer: "/env/NODE_TLS_REJECT_UNAUTHORIZED", ValueField: "NodeTLSRejectUnauthorized", ClearWhenEmpty: true},
+			// Force API-key auth so GOOGLE_GEMINI_BASE_URL is respected.
+			{JSONPointer: "/security/auth/selectedType", ValueField: "gemini-api-key"},
+			{JSONPointer: "/env/GEMINI_API_KEY", ValueField: "PlaceholderAPIKey"},
+			{JSONPointer: "/env/GOOGLE_GEMINI_BASE_URL", ValueField: "ProxyOrigin"},
```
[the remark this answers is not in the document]
yrobla: Tracked in #5177.

pkg/client/config.go
yrobla: Tracked in #5177 (adding an explicit Literal field will also enforce the mutual-exclusion contract).

test/e2e/cli_llm_all_clients_test.go
yrobla: Tracked in #5178.

test/e2e/cli_llm_all_clients_test.go
```diff
@@ -89,11 +89,15 @@ func allClientTestCases() []llmClientTestCase {
...
-			mode: "direct",
+			mode: "proxy",
 			expectedKeys: map[string]func(string, string) string{
-				"/auth/tokenCommand": func(_, _ string) string { return "llm token" },
-				"/baseUrl": func(gatewayURL, _ string) string {
-					return gatewayURL
+				// Force API-key auth so GOOGLE_GEMINI_BASE_URL is respected.
+				"/security/auth/selectedType": func(_, _ string) string { return "gemini-api-key" },
+				"/env/GEMINI_API_KEY":         func(_, _ string) string { return clientThvProxy },
+				// ProxyOrigin strips the /v1 path from the proxy base URL so Gemini
```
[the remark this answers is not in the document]
yrobla: Tracked in #5178.

test/e2e/cli_llm_all_clients_test.go
```diff
@@ -89,11 +89,16 @@ func allClientTestCases() []llmClientTestCase {
...
-				"/baseUrl": func(gatewayURL, _ string) string {
-					return gatewayURL
+				// Force API-key auth so GOOGLE_GEMINI_BASE_URL is respected.
+				"/security/auth/selectedType": func(_, _ string) string { return "gemini-api-key" },
+				"/env/GEMINI_API_KEY":         func(_, _ string) string { return clientThvProxy },
+				// ProxyOrigin strips the entire path (and any query/fragment) from
+				// the proxy base URL so Gemini CLI can append its own /v1beta/...
+				// path without doubling segments.
+				"/env/GOOGLE_GEMINI_BASE_URL": func(_, proxyURL string) string {
+					return strings.TrimSuffix(proxyURL, "/v1")
```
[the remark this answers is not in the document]
yrobla: Tracked in #5178.

pkg/client/llm_gateway_test.go
```diff
@@ -624,6 +627,147 @@ func TestConfigureLLMGateway_TLSSkipVerify_ClearRemovesKey(t *testing.T) {
...
+		{"some-literal", cfg, "some-literal"},
+	}
+
+	for _, tc := range cases {
+		name := tc.valueField
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			got := llmValueForSpec(tc.valueField, tc.cfg)
+			assert.Equal(t, tc.want, got)
+		})
```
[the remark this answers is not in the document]
yrobla: Tracked in #5179.

pkg/client/llm_gateway_test.go
```diff
@@ -624,6 +627,77 @@ func TestConfigureLLMGateway_TLSSkipVerify_ClearRemovesKey(t *testing.T) {
...
+		// ProxyOrigin falls back to the raw value when URL parsing fails
+		{"ProxyOrigin", llmgateway.ApplyConfig{ProxyBaseURL: "::invalid"}, "::invalid"},
+		// Unrecognised ValueField is written as a literal constant
+		{"gemini-api-key", cfg, "gemini-api-key"},
+		{"some-literal", cfg, "some-literal"},
+	}
+
+	for _, tc := range cases {
+		name := tc.valueField
+		t.Run(name, func(t *testing.T) {
```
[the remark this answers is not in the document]
yrobla: Tracked in #5179.

pkg/client/llm_gateway_test.go
yrobla: Tracked in #5179.

aponcedeleonch (approved):
Nice fix. The bug analysis is spot on, and the side-improvements (the `Literal` field, returning an error on unknown `ValueField`, and switching the test assertions to JSON-pointer lookups) catch a class of bugs that would otherwise have stayed hidden.

Two things worth following up on, neither of which I'd block on:

1. `LLMSetupNote` is plumbed end to end (struct field, getter, `GatewayManager` interface method, adapter, stub, output formatting in `configureDetectedTools`) but never set on any client. If the intent is to surface the gemini-cli >= v0.40.0 requirement (or the TLS note from point 2), I'd populate it for Gemini in this PR. Otherwise I'd drop the plumbing and reintroduce when there's an actual note to write. As-is it reads as added abstraction beyond what the task needs.

2. `--tls-skip-verify` is now silently a no-op for Gemini CLI. The reasoning for omitting `NODE_TLS_REJECT_UNAUTHORIZED` is sound (it's process-global, so flipping it would weaken every other HTTPS call the Gemini process makes), and I think omitting it is the right call. From a user's perspective though, running `thv llm setup --tls-skip-verify` against Gemini will succeed and quietly do nothing on that flag. Either a printed warning when the flag is set for a Gemini target, or a note surfaced via `LLMSetupNote`, would close the loop. This also ties point 1 back to a concrete use case.

PR #5147, opened 2026-04-30
Fall back to request-token claims for opaque upstream tokens

### LINKED ISSUE #5146: VirtualMCPServer Cedar authorization rejects all requests when upstream OIDC provider issues opaque access tokens
## Summary

When a `VirtualMCPServer` is configured with an embedded auth server using an upstream OIDC provider whose access tokens are opaque rather than JWTs (Google issues `ya29.*` opaque tokens), every Cedar authorization check fails. The vMCP gateway logs `Authorization check failed for tool, skipping` for every tool and returns an empty `tools/list` to authenticated clients.

#5002 introduced the `PrimaryUpstreamProvider` auto-binding for VirtualMCPServer that mirrors the working `thv run` path (verified in that PR against Okta, whose access tokens are JWTs). The same auto-binding causes the JWT parse step to fail unconditionally for OIDC providers that issue opaque access tokens.

## Reproduction

Minimal `VirtualMCPServer` with Google upstream + a permit-everything Cedar policy:

```yaml
kind: VirtualMCPServer
metadata:
  name: gateway
spec:
  groupRef:
    name: my-mcps
  authServerConfig:
    issuer: "https://gateway.example.com"
    hmacSecretRefs:
      - name: hmac-secret
        key: hmac-key
    upstreamProviders:
      - name: google
        type: oidc
        oidcConfig:
          issuerUrl: "https://accounts.google.com"
          clientId: "..."
          clientSecretRef: { name: google-oauth, key: clientSecret }
          redirectUri: "https://gateway.example.com/oauth/callback"
          additionalAuthorizationParams:
            access_type: offline
          scopes: ["openid", "email", "profile"]
  config:
    groupRef: my-mcps
    aggregation:
      conflictResolution: prefix
      conflictResolutionConfig: { prefixFormat: "{workload}_" }
  incomingAuth:
    type: oidc
    oidcConfigRef:
      name: gateway-incoming
      audience: "https://gateway.example.com/mcp"
    authzConfig:
      type: inline
      inline:
        policies:
          - 'permit(principal, action, resource);'
  outgoingAuth:
    source: discovered
```

1. Apply the manifest above plus the corresponding `MCPOIDCConfig` and Secret resources, on v0.21.0 or later (any release containing #5002).
2. Authenticate via the embedded auth server's Google OAuth flow.
3. Connect an MCP client (Claude Desktop, claude.ai, etc.) and call `tools/list`.

## Expected behavior

The empty Cedar policy `permit(principal, action, resource);` permits all tools, and the client receives the aggregated tool list.

## Actual behavior

Every tool is skipped with this gateway log line:

```
Authorization check failed for tool, skipping
  tool: <tool-name>
  error: failed to parse upstream token for provider "google":
    upstream token is not a parseable JWT: token is malformed:
    token contains an invalid number of segments
```

The client receives an empty tool list and cannot use the gateway.

## Root cause

Verified against `f9489a4b88cfdf6f984d4f665e00a01e90f13dfc` (current `main` plus PR #4994):

1. **Operator auto-binds without an opt-out.** `cmd/thv-operator/pkg/vmcpconfig/converter.go:201-205` unconditionally sets `incoming.Authz.PrimaryUpstreamProvider` to the first upstream provider's name whenever `authServerConfig.upstreamProviders` is non-empty. There is no CRD field exposed to override this.

2. **Storage holds the ID token but the reader returns only the access token.** `pkg/authserver/storage/types.go` defines `UpstreamTokens` with both `AccessToken` (line 61) and `IDToken` (line 63) fields. The OAuth callback at `pkg/authserver/server/handlers/callback.go:131-140` populates both from `idpTokens`. `pkg/auth/upstreamtoken/service.go` (lines 81, 107, 122, 174) returns `tokens.AccessToken` only. `IDToken` is never plumbed into `UpstreamCredential` or `identity.UpstreamTokens`.

3. **Cedar parse has no fallback.** `pkg/authz/authorizers/cedar/core.go:421-443` calls `parseUpstreamJWTClaims(upstreamToken)` whenever `primaryUpstreamProvider != ""`, regardless of whether the policy text references upstream-prefixed claims. The function comment at line 461 explicitly notes "Returns an error if the token is not a parseable JWT (e.g. opaque token)" — but the caller returns the error directly with no fallback path.

The `pkg/authserver/server/session/session.go` JWT session builder already maps the upstream `email`/`name`/`sub` into the AS-issued JWT's claims (around lines 135-138), so falling back to `identity.Claims` produces the same authorization decisions for the common case where policies reference the standard OIDC claims.

## Proposed fix

Smallest defensible change: in `resolveClaims`, when `parseUpstreamJWTClaims` returns the "not a parseable JWT" error, log a WARN and fall back to `jwt.MapClaims(identity.Claims)`. Preserves Okta and other JWT-access-token providers (no behavior change) while letting opaque-access-token providers like Google reuse the AS-issued token's claims.

A complementary operator change — only auto-set `PrimaryUpstreamProvider` when the user hasn't asked for the explicit fallback — would expose the documented escape hatch via a CRD field. Not required for the immediate bug fix but worth pairing.

## Workarounds today

- `incomingAuth.authzConfig.type: "none"` disables Cedar entirely.
- Removing `authServerConfig.upstreamProviders` and pointing `incomingAuth.oidcConfigRef` directly at Google as the OIDC issuer keeps Cedar functional but loses the embedded-AS features (token persistence, server-side token exchange to backends).

## Environment

- ToolHive operator + vmcp built from `f9489a4b88cfdf6f984d4f665e00a01e90f13dfc` (current `main` + PR #4994).
- Confirmed against v0.26.1 stock images carrying the same converter and Cedar code.

## Intent to contribute

Planning to submit a PR with the Cedar fallback plus a unit test covering the opaque-upstream-token path. Happy to also pair the operator opt-out change if that's of interest.

cjohnhanson:
## Summary

- VirtualMCPServer's incoming Cedar authorizer denied every request when the embedded auth server's upstream provider issues opaque OAuth 2.0 access tokens (Google's `ya29.*`, GitHub's `gho_*`). `resolveClaims` JWT-parsed the upstream token unconditionally and returned the parse error verbatim, so every authorization check failed and the gateway skipped every tool.
- Discriminate by token shape: if the upstream token is not three dot-separated segments it cannot be a JWT (per JOSE compact serialization), so fall back to `identity.Claims` (the request-token claims). The embedded auth server already mirrors the upstream OIDC `sub`/`email`/`name` into its issued AS token (see `pkg/authserver/server/session/session.go`), so policies referencing standard OIDC claims continue to evaluate correctly.
- JWT-shaped tokens (three segments) that fail to parse still return the error: a tampered or corrupted upstream JWT must not silently degrade to fallback claims. The added `looksLikeJWT(tokenStr) = strings.Count(tokenStr, ".") == 2` discriminator keeps that boundary explicit.
- The auto-binding of `PrimaryUpstreamProvider` added in #5002 (operator side, `cmd/thv-operator/pkg/vmcpconfig/converter.go`) doesn't expose an opt-out CRD field, so users with opaque-token providers had no in-config workaround. This change unblocks them without changing operator behavior.

Closes #5146

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Unit tests (`go test ./pkg/authz/...`)
- [x] `go vet ./pkg/authz/...`
- [x] `gofmt -l pkg/authz/authorizers/cedar/` (clean)
- [x] End-to-end verified on a GKE staging cluster: `VirtualMCPServer` with `accounts.google.com` upstream, Cedar policy referencing `claim_email`. Pre-fix: `tools/list` returns empty with `Authorization check failed for tool, skipping` per tool. Post-fix: full aggregated tool list returned, single WARN line `upstream token is not a JWT; falling back to request-token claims for Cedar evaluation provider=google`.

The existing `TestParseUpstreamJWTClaims/opaque_token_returns_error` row stays unchanged — the helper still returns the same error; only the caller's recovery behavior changes.

## API Compatibility

- [x] This PR does not break the `v1beta1` API. No CRD or `ConfigOptions` schema change.

## Changes

| File | Change |
|------|--------|
| `pkg/authz/authorizers/cedar/core.go` | `resolveClaims` discriminates `looksLikeJWT(tokenStr)` before falling back. Opaque (≠3 segments) → fall back to `identity.Claims` with a WARN. JWT-shaped but unparseable → return the original parse error (deny). Adds the `looksLikeJWT` helper. |
| `pkg/authz/authorizers/cedar/core_test.go` | Replaces the `upstream_token_opaque_not_parseable` row in `TestAuthorizeWithJWTClaims_UpstreamProvider` with three rows: opaque-fallback-denied (policy mismatch via fallback claims), opaque-fallback-permitted (policy matches via fallback claims), and JWT-shaped-but-malformed-still-errors (security-regression guard for tampered JWTs). |

Yes — bug-fix only. Operators with Google (or any opaque-access-token) upstream OIDC provider now receive the aggregated tools list instead of an empty one. No CRD or configuration change required. A WARN log surfaces each fallback so operators can observe the path their identity claims take.

Tampered or corrupted upstream JWTs still hard-deny — there is no behavior change for JWT-access-token providers (Okta, Azure AD with JWT access tokens, etc.).

A complementary operator-side change — only auto-set `PrimaryUpstreamProvider` when the user hasn't asked for the explicit fallback — would expose the documented escape hatch via a CRD field, since the docstring on `PrimaryUpstreamProvider` already says *"When empty, claims from the ToolHive-issued token are used."* That route is more invasive (CRD + types + converter + tests). Happy to PR it separately if you'd prefer the opt-in surface over the implicit fallback in this PR. Either approach unblocks the failing case.

The fallback could also be argued to belong behind a `ConfigOptions.AllowOpaqueTokenFallback` knob (default false). I left it implicit here because (a) the discriminator preserves the deny for tampered JWTs, which is the security-relevant case, and (b) for opaque-token providers the previous behavior was always-deny — there's no operator who relied on it doing anything useful. Open to adding the gate if you'd prefer the more conservative surface.

### DIFF
--- a/pkg/authz/authorizers/cedar/core.go
+++ b/pkg/authz/authorizers/cedar/core.go
@@ -492,6 +492,30 @@ func (a *Authorizer) resolveClaims(identity *auth.Identity) (jwt.MapClaims, erro
 		}
 		parsedClaims, err := parseUpstreamJWTClaims(upstreamToken)
 		if err != nil {
+			// Distinguish "not JWT-shaped" (opaque OAuth 2.0 access token —
+			// Google's ya29.*, GitHub's gho_*, etc.) from "JWT-shaped but
+			// malformed/tampered" (a JWT with three segments that fails to
+			// parse). Only fall back for the former; preserve the deny for
+			// the latter so a tampered upstream JWT cannot bypass policy.
+			//
+			// The embedded auth server already mirrors the upstream OIDC
+			// sub/email/name claims into its issued AS token (see
+			// pkg/authserver/server/session/session.go). For opaque-token
+			// providers, falling back to identity.Claims preserves identity
+			// for policies referencing standard OIDC claims; policies that
+			// reference upstream-only claims (groups, hd, custom namespaced
+			// claims) will see those attributes as absent and must be
+			// authored defensively (`has(claim_groups) && ...`).
+			if !looksLikeJWT(upstreamToken) {
+				// The Warn shares claimKeyLog with logClaimKeys below so a busy
+				// Google/GitHub deployment does not emit one line per tool call.
+				a.claimKeyLog.Do(func() {
+					slog.Warn("upstream token is not a JWT; falling back to request-token claims for Cedar evaluation",
+						"provider", a.primaryUpstreamProvider)
+				})
+				a.logClaimKeys("token-fallback", jwt.MapClaims(identity.Claims))
+				return jwt.MapClaims(identity.Claims), nil
+			}
 			return nil, fmt.Errorf("failed to parse upstream token for provider %q: %w",
 				a.primaryUpstreamProvider, err)
 		}
@@ -502,6 +526,13 @@ func (a *Authorizer) resolveClaims(identity *auth.Identity) (jwt.MapClaims, erro
 	claims := jwt.MapClaims(identity.Claims)
 	a.logClaimKeys("token", claims)
 	return claims, nil
+}
+
+// looksLikeJWT returns true when the token has the three-segment shape of a
+// JOSE-compact-serialized JWT (`header.payload.signature`). It does not
+// validate the contents; the parser handles that.
+func looksLikeJWT(tokenStr string) bool {
+	return strings.Count(tokenStr, ".") == 2
 }
 
 // logClaimKeys emits a rate-limited DEBUG log listing the JWT claim keys
--- a/pkg/authz/authorizers/cedar/core_test.go
+++ b/pkg/authz/authorizers/cedar/core_test.go
@@ -1296,7 +1296,7 @@ func TestAuthorizeWithJWTClaims_UpstreamProvider(t *testing.T) {
 			errContains: "upstream token for provider",
 		},
 		{
-			name: "upstream_token_opaque_not_parseable",
+			name: "upstream_token_opaque_falls_back_to_request_claims_denied",
 			identity: &auth.Identity{
 				PrincipalInfo: auth.PrincipalInfo{
 					Subject: "thv-user",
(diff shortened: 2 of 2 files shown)


pkg/authz/authorizers/cedar/core.go
```diff
@@ -492,6 +492,26 @@ func (a *Authorizer) resolveClaims(identity *auth.Identity) (jwt.MapClaims, erro
...
+			//
+			// The embedded auth server already mirrors the upstream OIDC
+			// sub/email/name claims into its issued AS token (see
+			// pkg/authserver/server/session/session.go). For opaque-token
+			// providers, falling back to identity.Claims preserves identity
+			// for policies referencing standard OIDC claims; policies that
+			// reference upstream-only claims (groups, hd, custom namespaced
+			// claims) will see those attributes as absent and must be
+			// authored defensively (`has(claim_groups) && ...`).
+			if !looksLikeJWT(upstreamToken) {
```
jhrozek, line 509: logClaimKeys right below this is already rate-limited via a.claimKeyLog.Do, but this Warn isn't, so any Google/GitHub deployment it'll emit one line per tool call

jhrozek (dismissed):

jhrozek:
hey @cjohnhanson first of all sorry it took so long to review your PR, it somehow slipped through the review queue. That shouldn't happen!

I've included just one nit inline, let me know if you want to address it if not, I'm fine sending a follow up (feeling a bit bad to ask you to do more work after letting you wait for almost 2 weeks..)

jhrozek:
oh looks like the linter is unhappy as well, this might require another push

cjohnhanson:
@jhrozek no worries, sorry for the delay on coming back to this. Force-pushed with the nit addressed.

jhrozek (approved):

PR #5160, opened 2026-05-01
Append Anthropic path prefix when writing ANTHROPIC_BASE_URL

### LINKED ISSUE #5158: `thv llm setup` doesn't append provider path prefix for Envoy AI Gateway
## Problem

`thv llm setup` writes `ANTHROPIC_BASE_URL=<gateway-url>` verbatim. When the gateway is an Envoy AI Gateway, native-Anthropic traffic is served at `/anthropic/v1/messages`, not `/v1/messages`. Without the prefix, Claude Code's requests fall through to a catchall route and the user sees:

> There's an issue with the selected model (claude-sonnet-4-6). It may not exist or you may not have access to it.

Gateway logs show `404`, `model: null`, with auth headers populated — i.e. auth works, the path doesn't.

## Workaround

After running `thv llm setup`, manually append `/anthropic` to the gateway URL in the patched tool config (e.g. set `ANTHROPIC_BASE_URL=https://<gateway>/anthropic` in Claude Code's settings).

## What needs to change

`thv llm setup` should know that some gateways prefix provider APIs. Two pieces:

1. **`--anthropic-path-prefix` flag** (default empty). Users on Envoy AI Gateway pass `--anthropic-path-prefix=/anthropic`; others leave it.
2. **Optional auto-probe.** HEAD/OPTIONS `<gateway>/anthropic/v1/messages` vs `/v1/messages` and pick the one that returns `401` (auth-reachable) over `404`.

Scope is direct-mode tools only (Claude Code, Codex). Proxy-mode tools (Cursor, VS Code) route through `thv llm proxy` and don't need the prefix in their config.

## Acceptance

- [ ] `--anthropic-path-prefix` flag exists and is honored when patching `ANTHROPIC_BASE_URL`.
- [ ] Setup against an Envoy AI Gateway with the flag set produces a working Claude Code config.
- [ ] Setup against a non-prefixed gateway (LiteLLM, direct Anthropic) still works with the flag unset.

jerm-dro:
## Summary

- Envoy AI Gateway serves native-Anthropic traffic at `/anthropic/v1/messages`, but `thv llm setup` writes `ANTHROPIC_BASE_URL=<gateway-url>` verbatim. Claude Code's requests fall through to Envoy's catchall and the user sees "model may not exist or you may not have access" while gateway logs show 404s with auth headers populated — auth works, the path is wrong.
- Add a persisted `llm.anthropic_path_prefix` setting (typed as a nullable string so `nil` and `""` are distinguishable) and an `--anthropic-path-prefix` flag on both `thv llm setup` and `thv llm config set`.
- **Default the prefix to `/anthropic`** (Envoy AI Gateway is the primary target). Users on LiteLLM or direct Anthropic opt out with `--anthropic-path-prefix=""`. The CLI uses `cmd.Flags().Changed()` so an explicit empty override is persisted instead of being discarded as a zero value, mirroring the existing `--tls-skip-verify` pattern.
- Proxy-mode tools (Cursor, VS Code) are unaffected because they forward OpenAI-shaped traffic through `thv llm proxy` on a separate route.

Closes #5158

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Manual: ran `go run ./cmd/thv llm setup --help` and `go run ./cmd/thv llm config set --help` to confirm the flag advertises the `/anthropic` default and the opt-out instruction on both commands.

New unit tests cover (red→green TDD):
- `pkg/client/llm_gateway_test.go::TestConfigureLLMGateway_AnthropicPathPrefix` — empty `ApplyConfig.AnthropicPathPrefix` produces a bare `ANTHROPIC_BASE_URL` (regression guard); `/anthropic` produces `<gateway>/anthropic`.
- `pkg/llm/config_test.go::TestConfig_EffectiveAnthropicPathPrefix` — nil resolves to `/anthropic` (default), explicit empty stays empty, explicit value passes through.
- `pkg/llm/manage_test.go::TestConfig_SetFields` — pointer cases assert that `&"/anthropic"` is accepted, `&""` opts out, and bare values / query strings / shell metacharacters are rejected.

## API Compatibility

Yes. After upgrade, a fresh `thv llm setup` writes `ANTHROPIC_BASE_URL=<gateway>/anthropic` by default — the right value for Envoy AI Gateway, which is the primary deployment target. Users on LiteLLM or direct Anthropic must now pass `--anthropic-path-prefix=""` (or set `llm.anthropic_path_prefix: ""` in `~/.toolhive/config.yaml`). `thv llm config show` reports the resolved value with a `(default)` or `(none — direct Anthropic / LiteLLM)` annotation so the active mode is visible at a glance.

## Implementation plan

<details>

### Approach

Persist the prefix in LLM config (it's a property of the deployed gateway, not a per-tool setting), expose it as a flag on `thv llm setup` and `thv llm config set`, and append it only when writing `ANTHROPIC_BASE_URL` for direct-mode tools. Default to `/anthropic` to optimise for the most common deployment (Envoy AI Gateway). The optional auto-probe from the issue is intentionally skipped — the issue marks it optional, and an HTTP probe over an OIDC-protected endpoint adds meaningful complexity (timeouts, TLS skip handling, false negatives) for marginal benefit when the flag is explicit.

### Changes

1. `pkg/llm/config.go` — add nullable `AnthropicPathPrefix *string` to `Config`, `DefaultAnthropicPathPrefix = "/anthropic"`, `EffectiveAnthropicPathPrefix()`; validate the prefix in `ValidatePartial` (must start with `/`, no query/fragment, length cap, no shell metacharacters).
2. `pkg/llm/manage.go` — `SetOptions.AnthropicPathPrefix *string`; `Show` annotates `(default)` and `(none)` modes.
3. `pkg/llmgateway/config.go` — add `AnthropicPathPrefix string` to `ApplyConfig` (resolved value).
4. `pkg/llm/setup.go` — pipe `EffectiveAnthropicPathPrefix()` into `configureDetectedTools`.
5. `pkg/client/llm_gateway.go` — new `AnthropicBaseURL` ValueField returns `cfg.GatewayURL + cfg.AnthropicPathPrefix`.
6. `pkg/client/config.go` — Claude Code's `/env/ANTHROPIC_BASE_URL` spec uses the new ValueField.
7. `cmd/thv/app/llm.go` — register `--anthropic-path-prefix` on both `setup` and `config set`; flag default is `/anthropic`; use `Flags().Changed()` to persist explicit empty overrides.
8. Tests as described in Test plan.

### Out of scope (intentional)

- Auto-probe — marked optional in the issue.
- Codex direct-mode — not currently implemented in the codebase.
- Proxy-mode upstream — issue confirms proxy-mode tools route through `thv llm proxy` and don't need the prefix in tool config; the proxy forwards OpenAI-shaped traffic on a separate Envoy route.

</details>

- This PR contains two commits: the first introduces the flag with no default; the second flips the default to `/anthropic` and switches the field to `*string` so the empty-override case can be persisted. Both commits keep tests green.
- The shell-metacharacter check on the prefix is defensive — the value flows into a Node-process env var, not a shell command — but it keeps surprises out of any future caller that does invoke a shell.
- LLM commands are `Hidden: true` in the CLI tree, so `task docs` does not need to regenerate anything for this PR.

### DIFF
--- a/cmd/thv/app/llm.go
+++ b/cmd/thv/app/llm.go
@@ -107,6 +107,9 @@ Example:
 	cmd.Flags().IntVar(&opts.CallbackPort, "callback-port", 0, "OIDC callback port (omit to keep current; default: ephemeral)")
 	cmd.Flags().BoolVar(&tlsSkipVerify, "tls-skip-verify", false,
 		"Skip TLS certificate verification for the upstream gateway (local dev only; use --tls-skip-verify=false to clear)")
+	cmd.Flags().StringVar(&opts.AnthropicPathPrefix, "anthropic-path-prefix", "",
+		"Path prefix appended to the gateway URL when writing ANTHROPIC_BASE_URL "+
+			"(e.g. \"/anthropic\" for Envoy AI Gateway). Leave empty for LiteLLM or direct Anthropic.")
 
 	return cmd
 }
@@ -266,6 +269,9 @@ Run "thv llm teardown" to revert all changes.`,
 			"For direct-mode tools (Claude Code, Gemini CLI) this sets NODE_TLS_REJECT_UNAUTHORIZED=0, "+
 			"disabling TLS for ALL of that tool's outbound connections. "+
 			"For proxy-mode tools only the proxy-to-gateway connection is affected.")
+	cmd.Flags().StringVar(&opts.AnthropicPathPrefix, "anthropic-path-prefix", "",
+		"Path prefix appended to the gateway URL when writing ANTHROPIC_BASE_URL "+
+			"(e.g. \"/anthropic\" for Envoy AI Gateway). Leave empty for LiteLLM or direct Anthropic.")
 	cmd.Flags().StringVar(&targetClient, "client", "",
 		"Configure only this AI tool by name (e.g. claude-code, cursor). Omit to configure all detected tools.")
 
--- a/cmd/thv/app/llm.go
+++ b/cmd/thv/app/llm.go
@@ -74,8 +74,9 @@ func newConfigCommand() *cobra.Command {
 
 func newConfigSetCommand() *cobra.Command {
 	var (
-		opts          llm.SetOptions
-		tlsSkipVerify bool
+		opts                llm.SetOptions
+		tlsSkipVerify       bool
+		anthropicPathPrefix string
 	)
 
 	cmd := &cobra.Command{
@@ -93,6 +94,9 @@ Example:
 			if cmd.Flags().Changed("tls-skip-verify") {
 				opts.TLSSkipVerify = &tlsSkipVerify
 			}
+			if cmd.Flags().Changed("anthropic-path-prefix") {
+				opts.AnthropicPathPrefix = &anthropicPathPrefix
+			}
 			return config.UpdateConfig(func(c *config.Config) error {
 				return c.LLM.SetFields(opts)
 			})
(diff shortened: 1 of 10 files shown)


jerm-dro:
Closing in favor of https://github.com/stacklok/toolhive/pull/5174/changes

PR #5168, opened 2026-05-02
Set User-Agent on OAuth token refresh requests

### LINKED ISSUE #5167: Token refresh requests do not advertise the ToolHive User-Agent
## Bug description

OAuth token refresh requests sent by ToolHive advertise `User-Agent: Go-http-client/1.1` (or `2.0`) instead of `User-Agent: ToolHive/1.0`. Server operators investigating WAF blocks, rate limits, or IP allowlists cannot identify the traffic as ToolHive, and `Go-http-client/*` is itself a common bot-detection signal that may trigger false positives.

ToolHive already sets `User-Agent: ToolHive/1.0` (the `oauthproto.UserAgent` constant) on the OAuth/OIDC requests it constructs directly — DCR registration (`pkg/oauthproto/dcr.go`), discovery (`pkg/oauthproto/discovery.go`), token introspection (`pkg/auth/token.go`), the GitHub provider (`pkg/auth/github_provider.go`). Token refresh is the gap: those requests are constructed by `golang.org/x/oauth2`, which uses `http.DefaultClient` (and therefore the default `Go-http-client/<version>` User-Agent) unless a custom `*http.Client` is injected via the `oauth2.HTTPClient` context value. None of the three call sites that create a refreshing `oauth2.TokenSource` do that today.

## Steps to reproduce

The snippet below uses the actual `pkg/auth/remote.CreateTokenSourceFromCached` function and captures the User-Agent on the wire for both the standard and RFC 8707 resource-aware refresh paths. Drop it into a checkout of the repo (e.g. `_repro/main.go`) and `go run` it:

```go
package main

import (
	"encoding/json"
	"fmt"
	"net/http"
	"net/http/httptest"
	"time"

	"golang.org/x/oauth2"

	"github.com/stacklok/toolhive/pkg/auth/remote"
)

func main() {
	for _, resource := range []string{"", "https://api.example.com"} {
		var got string
		srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			got = r.Header.Get("User-Agent")
			w.Header().Set("Content-Type", "application/json")
			_ = json.NewEncoder(w).Encode(map[string]any{
				"access_token": "x", "refresh_token": "y", "token_type": "Bearer", "expires_in": 3600,
			})
		}))
		cfg := &oauth2.Config{ClientID: "c", ClientSecret: "s", Endpoint: oauth2.Endpoint{TokenURL: srv.URL}}
		ts := remote.CreateTokenSourceFromCached(cfg, "old", time.Now().Add(-time.Hour), resource)
		_, _ = ts.Token()
		srv.Close()
		fmt.Printf("resource=%q  User-Agent: %q\n", resource, got)
	}
}
```

On `main` (8c90184f) this prints:

```
resource=""                          User-Agent: "Go-http-client/1.1"
resource="https://api.example.com"   User-Agent: "Go-http-client/1.1"
```

In production the same gap manifests as `User-Agent: Go-http-client/<n>` reaching the IdP's token endpoint, where `<n>` is `1.1` or `2.0` depending on the negotiated HTTP version.

## Expected behavior

OAuth token refresh requests should advertise `User-Agent: ToolHive/1.0`, matching every other outbound OAuth/OIDC request originated by ToolHive.

## Actual behavior

Refresh requests advertise Go's default `User-Agent: Go-http-client/1.1` (HTTP/1.1) or `Go-http-client/2.0` (HTTP/2). The `oauthproto.UserAgent` constant is defined and used by every other outbound OAuth/OIDC code path, but is never sent on refresh.

## Environment

- OS/version: any (the gap is in pure Go code paths, not OS-dependent).
- ToolHive version: current `main` (8c90184f); the latest release is `v0.26.1`.

## Additional context

Three call sites construct a refreshing `oauth2.TokenSource` without injecting a User-Agent-aware `*http.Client`:

- `pkg/auth/oauth/non_caching_refresher.go:50` — the `*http.Client` stored on `NonCachingRefresher` has no User-Agent transport. Both the standard and RFC 8707 resource-aware refresh paths flow through this client (via `oauth2.HTTPClient` context injection in `Token()`).
- `pkg/auth/oauth/flow.go:507` — `processToken` returns a long-lived `TokenSource` from `f.oauth2Config.TokenSource(context.Background(), token)` with no client injected.
- `pkg/auth/remote/persisting_token_source.go:101` — `CreateTokenSourceFromCached` calls `config.TokenSource(context.TODO(), token)` for the non-resource branch.

Each falls back to `http.DefaultClient` when the oauth2 library performs the refresh; `http.DefaultClient` uses `http.DefaultTransport`, which has no User-Agent header set, so net/http synthesizes the default `Go-http-client/<version>` value.

gkatz2:
## Summary

- ToolHive sets `User-Agent: ToolHive/1.0` on the OAuth/OIDC requests it constructs directly (DCR, discovery, token introspection, GitHub provider), but token refresh requests go through `golang.org/x/oauth2` and fall back to Go's default `Go-http-client/<version>` User-Agent. Server operators investigating WAF blocks, rate limits, or IP allowlists cannot identify the traffic as ToolHive, and `Go-http-client/*` is itself a common bot-detection signal that may trigger false positives.
- Adds `oauthproto.UserAgentTransport` (an `http.RoundTripper` that sets the ToolHive User-Agent when one is not already set) and `oauthproto.NewHTTPClient()` (a small constructor that bundles the transport with a 30s timeout).
- Wires the helper into the three call sites that construct a refreshing `oauth2.TokenSource`: `pkg/auth/oauth/non_caching_refresher.go`, `pkg/auth/oauth/flow.go`, and `pkg/auth/remote/persisting_token_source.go`.

Fixes #5167

## Type of change

- [x] Bug fix

## Test plan

- [x] Unit tests (`task test`) — including a new `TestUserAgentTransport_RoundTrip` table test, a new `TestUserAgentTransport_NilBase` round-trip test, an extra `User-Agent` assertion on the existing `TestResourceTokenSource_Token_ExpiredToken`, and a new `TestCreateTokenSourceFromCached_SetsUserAgent` covering both refresh paths.
- [x] Manual testing — built `thv` from this branch and ran a standalone harness that imports `pkg/auth/remote.CreateTokenSourceFromCached` against an `httptest.Server` and captures the User-Agent on the wire. On `main` (8c90184f) both refresh paths emit `Go-http-client/1.1`; on this branch both emit `ToolHive/1.0`.

## API Compatibility

OAuth token refresh requests now advertise `User-Agent: ToolHive/1.0` instead of Go's default `Go-http-client/<version>`. Operators of upstream IdPs can identify ToolHive traffic in logs, WAF rules, and rate-limit dashboards.

## Implementation plan

<details>

**Helper** — `pkg/oauthproto/useragent.go`:
- `UserAgentTransport` struct with a public `Base http.RoundTripper` field (mirrors stdlib `http.Transport` and `oauth2.Transport` shape). Nil `Base` falls back to `http.DefaultTransport`. `RoundTrip` clones the request before mutating headers (per the RoundTripper contract) and sets `User-Agent` only when not already set, so layered transports can override.
- `NewHTTPClient()` returns an `*http.Client` whose transport is `&UserAgentTransport{}` and whose `Timeout` is `30 * time.Second`. Used at all three call sites to lock in a single timeout invariant.

**Wire-up** — three call sites:
- `pkg/auth/oauth/non_caching_refresher.go` — `httpClient: oauthproto.NewHTTPClient()` on the `NonCachingRefresher` struct. Covers both standard and RFC 8707 resource-aware refresh (the refresher injects this client via `oauth2.HTTPClient` context value before each refresh). This also covers `pkg/auth/oauth/resource_token_source.go`, which delegates to `NonCachingRefresher`.
- `pkg/auth/oauth/flow.go` (`processToken` non-resource branch) — build a `*http.Client` via `oauthproto.NewHTTPClient()` and inject it via `oauth2.HTTPClient` into the context passed to `oauth2Config.TokenSource`. Preserves the existing `context.Background()` reasoning (the OAuth callback ctx is cancelled when the callback server shuts down).
- `pkg/auth/remote/persisting_token_source.go` (`CreateTokenSourceFromCached` non-resource branch) — same pattern; replaces the previous `context.TODO()`.

**Why the helper lives in `pkg/oauthproto`:** co-located with the existing `oauthproto.UserAgent` constant; the package already owns "set the ToolHive UA on outbound OAuth requests" via `pkg/oauthproto/dcr.go` and `pkg/oauthproto/discovery.go`. A future PR that fixes non-OAuth gaps (e.g. webhook delivery, transparent proxy health checks, etc.) can import the transport from here or move it to `pkg/networking` if that ends up being the better long-term home.

</details>

- The doc comment on `UserAgentTransport` calls out the WAF/attribution motivation so the rationale is durable in the source.
- `TestUserAgentTransport_NilBase` exercises the `nil → http.DefaultTransport` fallback through a real `httptest.Server` round-trip rather than relying on a recorder, since a recorder would mask the fallback.
- The standalone refresh paths are covered, but I deliberately did not add a dedicated unit test for `flow.go`'s `processToken` change — that path requires a full OAuth callback flow to exercise. The `non_caching_refresher` and `persisting_token_source` tests verify the transport behavior; the `flow.go` wire-up is the same three-line pattern.
- The fix is scoped to the OAuth token refresh path. ToolHive has other outbound HTTP code paths that are also missing the User-Agent (e.g. `pkg/auth/discovery`, `pkg/authserver/upstream`, `pkg/webhook`).

### DIFF
--- a/pkg/auth/oauth/flow.go
+++ b/pkg/auth/oauth/flow.go
@@ -503,8 +503,12 @@ func (f *Flow) processToken(_ context.Context, token *oauth2.Token) *TokenResult
 		// Use resourceTokenSource wrapper to add resource parameter to refresh requests (RFC 8707)
 		base = NewResourceTokenSource(f.oauth2Config, token, f.config.Resource)
 	} else {
-		// No resource parameter needed, use standard token source
-		base = f.oauth2Config.TokenSource(context.Background(), token)
+		// No resource parameter needed, use standard token source. Inject an
+		// HTTP client whose transport sets the ToolHive User-Agent so the
+		// oauth2 library does not fall back to Go-http-client/2.0 on token
+		// refresh requests.
+		ctx := context.WithValue(context.Background(), oauth2.HTTPClient, oauthproto.NewHTTPClient())
+		base = f.oauth2Config.TokenSource(ctx, token)
 	}
 
 	// ReuseTokenSource ensures that refresh happens only when needed
--- a/pkg/auth/oauth/non_caching_refresher.go
+++ b/pkg/auth/oauth/non_caching_refresher.go
@@ -13,6 +13,8 @@ import (
 	"time"
 
 	"golang.org/x/oauth2"
+
+	"github.com/stacklok/toolhive/pkg/oauthproto"
 )
 
 // NonCachingRefresher is an oauth2.TokenSource that always performs a network
@@ -47,7 +49,7 @@ func NewNonCachingRefresher(cfg *oauth2.Config, refreshToken, resource string) *
 		cfg:          cfg,
 		resource:     resource,
 		refreshToken: refreshToken,
-		httpClient:   &http.Client{Timeout: 30 * time.Second},
+		httpClient:   oauthproto.NewHTTPClient(),
 	}
 }
 
--- a/pkg/auth/oauth/resource_token_source_test.go
+++ b/pkg/auth/oauth/resource_token_source_test.go
@@ -14,6 +14,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/oauth2"
+
+	"github.com/stacklok/toolhive/pkg/oauthproto"
 )
 
 func clientCredentialsFromRequest(r *http.Request) (clientID, clientSecret string) {
@@ -164,6 +166,7 @@ func TestResourceTokenSource_Token_ExpiredToken(t *testing.T) {
 		require.NotNil(t, capturedRequest)
 		assert.Equal(t, "POST", capturedRequest.Method)
 		assert.Equal(t, "application/x-www-form-urlencoded", capturedRequest.Header.Get("Content-Type"))
+		assert.Equal(t, oauthproto.UserAgent, capturedRequest.Header.Get("User-Agent"))
 	})
 
 	t.Run("includes client credentials in refresh request", func(t *testing.T) {
--- a/pkg/auth/remote/persisting_token_source.go
+++ b/pkg/auth/remote/persisting_token_source.go
@@ -12,6 +12,7 @@ import (
 	"golang.org/x/oauth2"
 
 	"github.com/stacklok/toolhive/pkg/auth/oauth"
+	"github.com/stacklok/toolhive/pkg/oauthproto"
 )
 
 // TokenPersister is a callback function that persists OAuth refresh tokens.
@@ -98,7 +99,11 @@ func CreateTokenSourceFromCached(
 	if resource != "" {
 		base = oauth.NewResourceTokenSource(config, token, resource)
 	} else {
-		base = config.TokenSource(context.TODO(), token)
+		// Inject an HTTP client whose transport sets the ToolHive User-Agent
+		// so the oauth2 library does not fall back to Go-http-client/2.0 on
+		// token refresh requests.
+		ctx := context.WithValue(context.Background(), oauth2.HTTPClient, oauthproto.NewHTTPClient())
+		base = config.TokenSource(ctx, token)
 	}
 
 	return oauth2.ReuseTokenSource(token, base)
--- a/pkg/auth/remote/persisting_token_source_test.go
+++ b/pkg/auth/remote/persisting_token_source_test.go
@@ -4,7 +4,10 @@
 package remote
 
 import (
+	"encoding/json"
 	"errors"
+	"net/http"
+	"net/http/httptest"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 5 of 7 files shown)


jhrozek (approved):
LGTM. The fix is well-scoped: the `oauth2.HTTPClient` context key is the documented extension point for injecting a custom `*http.Client` into `golang.org/x/oauth2`'s refresh path, and a wrapping `http.RoundTripper` is the canonical way to add a header to requests the library builds internally. Test coverage looks good — the table-driven tests cover both the "UA set when missing" and "caller-supplied UA preserved" cases, plus the nil-Base fallback through a real `httptest` server, and the persisting-token-source test exercises both the standard and resource-aware refresh paths end-to-end.

PR #5170, opened 2026-05-03
Retry OAuth token refresh on infrastructure 4xx

### LINKED ISSUE #5169: Token refresh treats transient 4xx as permanent auth failures
## Bug description

When the OAuth token refresh endpoint returns any 4xx response, `MonitoredTokenSource` immediately marks the remote MCP workload as `unauthenticated` and stops monitoring. Recovery requires manual re-authentication. No further refresh attempts are made.

This is too aggressive. Many real-world 4xx responses are transient infrastructure-level errors that resolve on their own:

- **WAF / firewall blocks**: a Cloudflare or similar WAF returns 403 with an HTML body when the refresh request comes from a non-allowlisted IP. A momentary VPN flap that routes the request through the user's home IP triggers a block; the next refresh would succeed once the tunnel reconnects.
- **Rate limiting**: some OAuth servers return 429 (or 403) for rate limiting. The condition clears after a short cooldown.
- **Transient bad-config deploys**: an OAuth server briefly returns 400 for a few minutes before rolling back. ToolHive marks the workload permanently dead for a blip.

The retry infrastructure already exists in `MonitoredTokenSource` (added in #4513 for 5xx errors). The root cause is that `isTransientNetworkError()` in `pkg/auth/monitored_token_source.go` treats every non-5xx `*oauth2.RetrieveError` as permanent, regardless of whether the response actually contained a structured RFC 6749 error code.

## Steps to reproduce

1. Run a remote MCP server with OAuth authentication: `thv run <remote-url> --remote-auth ...`.
2. Arrange for the next token refresh to hit a 4xx response without a parseable RFC 6749 `error` field. The simplest controlled reproduction is a mock token endpoint returning 403 with `Content-Type: text/html` and an HTML body. In practice this happens organically when a VPN drops momentarily and the request egresses through a WAF.
3. Inspect the workload status file (`~/Library/Application Support/toolhive/statuses/<server>.json` on macOS). Observe `status: unauthenticated` and `status_context` containing the raw 4xx body, with no preceding retry attempts in the logs.

The bug manifests during real-world transient outages and infrastructure events. The code path can be verified by reading `isTransientNetworkError()` and `Token()` in `pkg/auth/monitored_token_source.go`.

## Expected behavior

A 4xx response without a populated RFC 6749 `error` field (e.g. an HTML page from a WAF) should be classified as transient and enter the existing retry loop. Only 4xx responses where the OAuth server returned a structured error code (`invalid_grant`, `invalid_client`, etc.) should be treated as permanent — those are the cases where the protocol explicitly says the credentials are bad.

## Actual behavior

All 4xx responses are treated as permanent. The workload is immediately marked `unauthenticated` and the monitor stops.

Example status produced by a Cloudflare WAF block:

```json
{
  "status": "unauthenticated",
  "status_context": "Token retrieval failed: oauth2: cannot fetch token: 403 Forbidden\nResponse: <html>...Cloudflare Firewall Block...</html>"
}
```

Subsequent requests through the proxy fail until the workload is manually re-authenticated, even after the underlying issue has cleared.

## Environment (if relevant)

- ToolHive v0.26.1 (the affected code path is unchanged on `main` at the time of filing)

### Classification rule that fixes the misclassification

The `golang.org/x/oauth2` library populates `RetrieveError.ErrorCode` only when the response body is parseable JSON containing an RFC 6749 `error` field. An empty `ErrorCode` therefore signals an infrastructure-level response (HTML page from a WAF, CDN, or reverse proxy), not an OAuth protocol failure. This gives a clean structural signal for the classification:

- 5xx: transient (existing behavior, from #4513)
- 429: transient regardless of body (HTTP standard)
- 4xx with an empty `ErrorCode`: transient (infrastructure error)
- 4xx with a populated `ErrorCode`: permanent (OAuth told us specifically what's wrong)

### Scope

This issue is specifically about *misclassification* of transient infrastructure errors as permanent. A more architectural change — keeping the monitor alive across token expiry intervals so workloads recover from outages that span longer than a single refresh attempt — would require a new "transiently failing" workload state, status file representation, and monitor lifecycle changes. That's a separable conversation. The classification fix is independently valuable for the common short-blip cases.

### Relevant prior work

- #4512 / #4513 added retry for 5xx and HTML-on-200 token-endpoint responses, but kept all 4xx classified as permanent.
- #5044 (in flight) emits a DCR remediation hint on permanent 4xx. Sharpening the classification removes false triggers — the hint should fire when DCR credentials are actually stale, not when a WAF blocks the request.

gkatz2:
## Summary

- The OAuth token refresh endpoint returns 4xx for infrastructure events (WAF/firewall blocks, rate limits, transient bad-config deploys) as well as for OAuth protocol failures. Today every 4xx is treated as a permanent auth failure, killing the workload until manual re-authentication. A momentary VPN flap that routes a refresh request through a non-allowlisted IP is enough to leave the workload dead.
- Treat 4xx responses that lack a structured RFC 6749 error code as transient, so they enter the existing retry loop added in #4513. Only 4xx with a populated error code (`invalid_grant`, `invalid_client`, etc.) remains permanent. 429 is always transient per HTTP standard.

Fixes #5169

## Type of change

- [x] Bug fix

## Test plan

Added `TestIsTransientNetworkError_AgainstRealOAuth2Library`, which exercises the classification function end-to-end through `golang.org/x/oauth2.Config.TokenSource` against an `httptest.Server` returning each response shape (HTML 403, HTML 401, JSON `invalid_grant`, JSON `invalid_client`, 429, 503). This pins the assumption that `RetrieveError.ErrorCode` is populated only for parseable RFC 6749 error responses, so a future oauth2 upgrade that changes that behavior would surface here, not in production.

## API Compatibility

Remote MCP servers with OAuth authentication now survive transient 4xx token-endpoint events (WAF blocks, rate limits, brief upstream config errors) instead of permanently going dark on the first occurrence.

### Classification rule

`golang.org/x/oauth2` populates `RetrieveError.ErrorCode` only when the response body is parseable JSON containing an RFC 6749 `error` field. An empty `ErrorCode` therefore signals an infrastructure-level response (HTML page from a WAF, CDN, or reverse proxy), not an OAuth protocol failure. The new `classifyOAuthRetrieveError` helper applies this rule:

- 5xx: transient (existing behavior, from #4513)
- 429: transient regardless of body (HTTP standard)
- 4xx with empty `ErrorCode`: transient (infrastructure error)
- 4xx with populated `ErrorCode`: permanent (OAuth protocol failure)

### Scope

Classification only — no changes to retry parameters, monitor lifecycle, or workload state machine. Outages that span longer than a single refresh attempt (e.g. a multi-hour VPN outage that crosses a token expiry boundary) still result in the workload being marked unauthenticated; that's a separable architectural change requiring a new "transiently failing" state and is not part of this fix.

### Interaction with #5044

PR #5044 (in flight) emits a DCR remediation hint when classification returns permanent 4xx. This change reduces false triggers of that hint: the hint will fire for `invalid_grant`/`invalid_client` (where DCR creds genuinely may be stale) but no longer for WAF blocks (where the hint would have been misleading). Pure mechanical merge conflict on the same file, no semantic conflict.

### DIFF
--- a/pkg/auth/monitored_token_source.go
+++ b/pkg/auth/monitored_token_source.go
@@ -283,11 +283,11 @@ func (mts *MonitoredTokenSource) Stopped() <-chan struct{} {
 	return mts.stopped
 }
 
-// Token retrieves a token, retrying with exponential backoff on transient errors
-// (see isTransientNetworkError for the full list). On non-transient errors
-// (OAuth 4xx, TLS failures) it marks the workload as unauthenticated and returns
-// immediately. Context cancellation (workload removal) stops the retry without
-// marking the workload as unauthenticated.
+// Token retrieves a token, retrying with exponential backoff on transient
+// errors and marking the workload as unauthenticated on non-transient errors.
+// See isTransientNetworkError for the classification rule. Context
+// cancellation (workload removal) stops the retry without marking the
+// workload as unauthenticated.
 //
 // Concurrent callers are deduplicated via singleflight so that only one retry
 // loop runs at a time during transient failures.
@@ -382,13 +382,20 @@ func (mts *MonitoredTokenSource) onTick() (bool, time.Duration) {
 	return false, wait
 }
 
-// isTransientNetworkError reports whether err represents a transient condition
-// (DNS failure, TCP transport error, timeout, OAuth server 5xx, unparsable
-// token response) that is likely to resolve on its own.
+// isTransientNetworkError reports whether err represents a transient
+// condition that is likely to resolve on its own. The categories are:
 //
-// OAuth2 client-level auth failures (invalid_grant, 401, 400) and TLS errors
-// (certificate verification, handshake failure) are NOT considered transient and
-// return false so the workload is marked unauthenticated immediately.
+//   - Network-level failures: DNS lookup errors, TCP transport errors,
+//     timeouts.
+//   - OAuth token-endpoint responses classified as transient by
+//     isTransientRetrieveError.
+//   - Unparsable token responses on a 2xx status (typically an HTML page
+//     from a load balancer or CDN).
+//
+// All other errors return false, causing the workload to be marked
+// unauthenticated. TLS failures (certificate verification, handshake
+// failure) are intentionally non-transient even though they surface
+// through net.OpError like transport-level errors.
 //
 // The function is side-effect free; callers that want to emit a DCR
 // remediation hint on a permanent 4xx must do so themselves at the
@@ -400,17 +407,8 @@ func isTransientNetworkError(err error) bool {
 		return false
 	}
 
-	// OAuth HTTP-level errors: 5xx (Bad Gateway, Service Unavailable, Gateway
-	// Timeout) are transient server-side issues that typically resolve on their
-	// own. 4xx errors (invalid_grant, invalid_client) are permanent auth failures.
 	if retrieveErr, ok := errors.AsType[*oauth2.RetrieveError](err); ok {
-		if retrieveErr.Response != nil && retrieveErr.Response.StatusCode >= 500 {
-			slog.Debug("treating OAuth server error as transient",
-				"status_code", retrieveErr.Response.StatusCode,
-			)
-			return true
-		}
-		return false
+		return isTransientRetrieveError(retrieveErr)
 	}
 
 	// Non-JSON responses from the OAuth server (e.g. load balancer HTML pages).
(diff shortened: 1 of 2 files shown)


pkg/auth/monitored_token_source.go
```diff
@@ -399,6 +398,54 @@ func isTransientNetworkError(err error) bool {
...
+	statusCode := retrieveErr.Response.StatusCode
+
+	if statusCode >= 500 {
+		slog.Debug("treating OAuth server error as transient",
+			"status_code", statusCode,
+		)
+		return true
+	}
+
+	if statusCode == http.StatusTooManyRequests {
```
jhrozek, line 500: 429 is now transient, but `Retry-After` is ignored — the existing 10s → 2m exponential backoff applies regardless of what the server suggests. Real providers (GitHub, Google) routinely set the header, so a `Retry-After: 600` will get hit ~5–30 times before the 5-minute cap stops things, instead of waiting the requested cooldown. RFC 6585 §4 says it's `MAY`, so ignoring it is technically compliant.

Honoring it cleanly would mean threading a "min wait" hint from the classifier into the retry loop, which is a bigger change than this PR's classification-only scope. Would you want to file a follow-up issue (or follow-up PR) for it?

pkg/auth/monitored_token_source_test.go
```diff
@@ -657,6 +684,109 @@ func TestMonitoredTokenSource_BackgroundMonitor_ErrorClassification(t *testing.T
...
 
+// TestIsTransientNetworkError_AgainstRealOAuth2Library exercises the
+// classification function against errors actually produced by
+// golang.org/x/oauth2 when refreshing a token against a real HTTP server.
+//
+// The classification rule depends on whether the oauth2 library populates
+// RetrieveError.ErrorCode for a given response shape. This test pins that
+// assumption so a future oauth2 upgrade that changes ErrorCode population
+// would surface here, not in production.
+func TestIsTransientNetworkError_AgainstRealOAuth2Library(t *testing.T) {
```
jhrozek, line 747: All six subtests here mirror cases already covered by the synthetic table in `TestMonitoredTokenSource_BackgroundMonitor_ErrorClassification` (`400 invalid_grant`, `429 empty body`, `503`, HTML 401/403, etc.).

The doc comment frames the real value well — pinning the `golang.org/x/oauth2` `ErrorCode`-population contract so a future library upgrade that changes that behavior fails here, not in production. That's worth keeping. But as written it reads like overlapping unit coverage, which `.claude/rules/testing.md` flags ("Avoid overlapping test cases exercising the same code path" / "Tests must only test code in the package under test").

Two ways to make the intent visible:

1. Trim to just the cases where the synthetic helpers could diverge from real library output — i.e. the HTML/WAF shapes (where the empty-`ErrorCode` premise is non-obvious), plus an edge case like a form-encoded error body. Drop the JSON `invalid_grant`/`invalid_client` cases that the synthetic table already pins.
2. Move to a `contract_test.go` with a header comment noting its role is dependency pinning, not unit coverage of the package's own logic.

Either works — preference?

jhrozek (dismissed):
LGTM. Classification logic is RFC 6749 / 6585 compliant, the real-library contract test is the right way to defend against future `golang.org/x/oauth2` upgrades changing `ErrorCode` semantics, and singleflight already prevents the cross-replica refresh-rotation race from getting worse. PR scope is tight (classification only, ~206 LOC).

Two non-blocking comments below — both are suggestions for follow-up scope and a small test-design refinement, not anything that should hold this up.

pkg/auth/monitored_token_source_test.go
gkatz2: I went with option (1) in the latest amend. I trimmed to 3 cases – kept HTML 401 and HTML 403 (WAF), dropped the JSON and 5xx/429 cases (synthetic table covers them), and added a 400-with-form-encoded-`invalid_grant` case where the synthetic helper would lie (`golang.org/x/oauth2` populates ErrorCode from form-encoded bodies – `internal/token.go:288` in v0.34.0). I tightened the doc comment to frame the test as dependency-pinning, not unit coverage.

gkatz2:
I force-pushed an amended version with the rebase plus a few additional refinements. Headline:

1. **Rebased onto current main** (clean conflict resolution against #5044 and #5168).

2. **Tightened `isPermanentTokenEndpointError`** to be the strict inverse of `isTransientRetrieveError` on the `*oauth2.RetrieveError` branch. The DCR remediation Warn now fires only when the response carries a populated RFC 6749 `error` code — not on 4xx-with-empty-ErrorCode (HTML pages from a WAF, CDN, or reverse proxy). The reasoning: a non-spec-compliant response means the OAuth server didn't render a verdict, so recommending "delete cached credentials" based on it would frequently mislead operators whose real problem is upstream of the OAuth server. This is a behavior tweak to code originally added in #5044 — flagging this for @tgrunnagle in case you want to weigh in.

3. **Renamed `classifyOAuthRetrieveError` → `isTransientRetrieveError`** for naming symmetry with `isPermanentTokenEndpointError` and to match the file's `is<Adjective><Noun>Error` convention.

4. **Added `TestIsPermanentTokenEndpointError`** to assert the new inverse behavior directly. The function had 100% line coverage before, but no test asserted on its return value, so the semantic change wouldn't have failed any existing test.

5. **Trimmed `TestIsTransientNetworkError_AgainstRealOAuth2Library`** per your [comment](https://github.com/stacklok/toolhive/pull/5170#discussion_r3188613394).

Please re-review when you have a chance, @jhrozek. The diff between the previous approval state and current HEAD is centered on `pkg/auth/monitored_token_source.go` around `isPermanentTokenEndpointError` / `isTransientRetrieveError`.

jhrozek (approved):
Both review items from the previous pass are resolved. The Retry-After point was non-blocking and explicitly a follow-up.

Also a nice bonus: tightening isPermanentTokenEndpointError to require a populated ErrorCode means the DCR remediation Warn no longer fires on WAF/CDN HTML pages, which would have misled operators. Worth flagging @tgrunnagle since this touches the Warn from #5044.

jhrozek:
I apologize for the delay @gkatz2 this somehow slipped in the review queue madness

gkatz2:
> I apologize for the delay @gkatz2 this somehow slipped in the review queue madness

All good, @jhrozek! Thanks for the review. I can't imagine what it is like to handle the PR load on an open source project these days. 🫡

PR #5172, opened 2026-05-03
Reset RunWorkload retry counter after stable run

### LINKED ISSUE #5171: Stdio MCP server gives up after sustained Docker daemon disruption
## Bug description

`RunWorkload`'s retry loop (in `pkg/workloads/manager.go`) can exhaust its
10-attempt retry budget over the course of 10–25 minutes when the Docker
daemon is repeatedly unresponsive, and permanently stop trying to keep an
stdio workload alive — even though every individual restart cycle ToolHive
attempted actually succeeded in starting the container. The user-facing
symptom: a previously-healthy MCP server goes offline and stays offline
until manually restarted.

The retry counter does not reset between successful runs. Every cycle in
which the container starts cleanly, runs for tens of seconds, then is killed
by the next disruption, counts equally toward the budget alongside genuine
failures-to-start. After ~10 such cycles, the manager logs `failed to
restart after max attempts, giving up` and leaves the workload in `error`
state.

I have observed this in my environment with OrbStack on macOS, where the
Docker socket periodically becomes unresponsive for several minutes at a
time. The container runtime emits errors like
`Error response from daemon: handle request: read response: unexpected EOF`
and
`read unix ->/.../orbstack/run/docker.sock: use of closed network connection`
across the affected window. Each of those connection drops kills the stdio
attach session, the container's PID 1 exits on stdin EOF, and ToolHive's
manager records a retry. After 10 such cycles, the manager gives up.

The bug is in the retry loop's accounting, not in any environment-specific
code path — the same loop runs for every workload regardless of runtime, so
any container runtime that exhibits sustained connection instability would
trigger the same outcome.

## Steps to reproduce

This requires sustained disruption — a single container kill is not enough,
because the retry budget is sized to absorb that. The bug needs ~10 cycles
where the container starts successfully, runs briefly, and is killed before
the disturbance ends.

```bash
# Start a stdio MCP server
thv run --transport stdio --name test-recovery mcp/time:latest

# Wait for it to be running
until docker ps --filter name=test-recovery --format '{{.Names}}' | grep -q test-recovery; do
  sleep 1
done

# Simulate a sustained Docker daemon disruption: kill the container shortly
# after every restart. After ~10 cycles spanning roughly 8 minutes, the
# manager exhausts its retry budget and gives up.
while true; do
  while ! docker ps --filter name=test-recovery --format '{{.Names}}' 2>/dev/null | grep -q test-recovery; do
    sleep 1
  done
  sleep 5
  docker kill test-recovery 2>/dev/null
done

# After ~8 minutes, in another terminal:
grep -E "running as detached|workload exited unexpectedly|failed to restart" \
  ~/Library/Application\ Support/toolhive/logs/test-recovery.log | tail -25
```

The log will show the retry counter climbing from `attempt 1/10` through
`attempt 10/10`, followed by `failed to restart after max attempts, giving up`.

## Expected behavior

A series of *successful* short-lived restarts caused by an external
disturbance should not be treated equivalently to a tight crash loop where
the workload cannot get started at all. Once the disturbance ends, the
manager should recover the workload.

## Actual behavior

After 10 attempts (each successful at starting the container, but each
interrupted by the next disruption), the manager logs `failed to restart
after max attempts, giving up`, sets the workload status to `error`, and
the workload stays offline until manually restarted via `thv restart` or
equivalent.

## Environment (if relevant)

- OS/version: macOS Darwin 25.3.0; container runtime: OrbStack 2.1.0 / 2.1.1
- ToolHive version: observed across multiple versions including v0.14.1 and v0.26.1.

## Additional context

Production log excerpt from one occurrence (slack server, 2026-03-30,
disruption lasting ~24 minutes):

```
23:32:18 INFO  running as detached process pid=23599
23:32:48 WARN  workload exited unexpectedly, restarting attempt=2 maxRetries=10
23:32:59 INFO  running as detached process pid=23599
... 7 more cycles, each running for 10–30s before next disruption ...
23:55:00 INFO  running as detached process pid=23599
23:55:15 WARN  workload exited unexpectedly, restarting attempt=9 maxRetries=10
23:55:16 INFO  restart attempt attempt=10 delay=60s
23:56:17 INFO  running as detached process pid=23599
23:56:27 WARN  workload exited unexpectedly, restarting attempt=10 maxRetries=10
23:56:28 ERROR failed to restart after max attempts, giving up
```

Each `running as detached process` line indicates a successful start. The
container ran for tens of seconds before being killed by the next
disruption. The retry counter incremented across these *successful* starts
and exhausted at attempt 10.

I have observed seven such failures in two months of ToolHive logs across
five stdio servers (atlassian, gitlab, google-workspace, pagerduty, slack).
Each ended with `failed to restart after max attempts, giving up`.

### Root cause

In `RunWorkload`'s retry loop (`pkg/workloads/manager.go`), the attempt
counter is incremented on every `ErrContainerExitedRestartNeeded` regardless
of how long the workload ran successfully before failing. There is no signal
distinguishing a failure inside a tight crash loop from a failure that came
after the workload had been running healthily.

gkatz2:
## Summary

When the Docker daemon is repeatedly unresponsive, `RunWorkload`'s 10-attempt retry budget can be exhausted by short-lived but successful starts: each container start succeeds, runs for tens of seconds, gets killed by the next disruption, and counts as one of the 10 attempts. After ~10 cycles the manager logs `failed to restart after max attempts, giving up` and the workload stays offline until manually restarted, even though the original cause was transient.

This PR tracks the duration of each `runner.Run` call and resets the retry counter when the workload ran past a stable threshold (30s) before failing. The retry budget continues to bound tight crash loops where the workload can never get healthy.

I have observed this on five stdio MCP servers (atlassian, gitlab, google-workspace, pagerduty, slack) in my local environment over two months. The bug report has the production log evidence.

Fixes #5171

## Type of change

## Changes

- `pkg/workloads/manager.go`:
  - Track `runDuration`; if it ≥ stable threshold on an `ErrContainerExitedRestartNeeded`, reset `attempt` counter and backoff.
  - Bundle retry parameters into a `retryConfig` struct; production callers leave the field nil and get defaults matching the previous hard-coded values (`maxRetries=10`, `initialDelay=5s`, `maxBackoff=60s`, `stableRunThreshold=30s`).
  - Add an `mcpRunner` interface + factory field on `DefaultManager` so the retry loop can be exercised against a fake runner that returns `ErrContainerExitedRestartNeeded` or success after a controlled duration, without needing a real container runtime.
  - Differentiate the `WorkloadStatusStarting` reason string between stable-run failures and tight-crash-loop failures.
- `pkg/workloads/manager_test.go` — 7 table-driven cases covering: all-short-runs exhaust budget; single stable run extends budget; mid-sequence stable resets; boundary `==threshold`; stable-then-success; non-restart error bails immediately; immediate success.
- `pkg/workloads/mocks/mock_manager.go` — regenerated. Mockgen directive switched from source mode to reflect mode so only the public `Manager` interface is mocked; the internal `mcpRunner` test seam stays private. Auto-generated header line updated accordingly.

## Test plan

- [x] Unit tests pass (`task test`).
- [x] `task lint-fix` reports 0 issues.
- [x] Manual end-to-end reproduction. To keep iteration fast I rebuilt the binary with shortened retry constants (`maxRetries=5`, `initialDelay=2s`, `maxBackoff=5s`, `stableRunThreshold=15s`) and restored production values before committing.
  - Bug repro without the fix: ran `docker kill <container>; sleep 7` in a 10-iteration loop against a stdio test server. Each container lifetime was 5–10 seconds — under the 15s shortened threshold. The manager incremented the attempt counter on every cycle and gave up with `failed to restart after max attempts, giving up` at attempt 5/5.
  - Fix verified for stable runs: ran `sleep 18; for i in 1..10; do docker kill <container>; sleep 20; done`. Each container ran 20–25 seconds (over the 15s shortened threshold). The attempt counter pinned at 1 across all 10 cycles.
  - Safety check: re-ran the 7s kill loop with the fix applied. The manager still hit attempt 5/5 and gave up — the threshold-based reset does not enable runaway loops on workloads that never get healthy.

The `attempt = 0` assignment in the reset path relies on the for-clause's `attempt++` running on `continue`, landing the next iteration at `attempt = 1`. There is an inline comment explaining this; it's a small wart but the alternatives (a `skipSleep` flag, restructuring the loop) seemed worse. Open to feedback.


pkg/workloads/manager.go
```diff
@@ -97,6 +97,62 @@ type DefaultManager struct {
...
+
+	// retryConfig holds tunable parameters for the RunWorkload retry loop.
+	// Production callers leave this nil so RunWorkload uses defaultRetryConfig();
+	// tests inject smaller values for fast execution.
+	retryConfig *retryConfig
+
+	// newRunner is the factory used by RunWorkload to construct the per-attempt
+	// runner. Production callers leave this nil so RunWorkload falls back to
+	// runner.NewRunner; tests inject a mock factory.
+	newRunner mcpRunnerFactory
```
amirejaz: Both `retryConfig` and `newRunner` are nil-checked fields whose sole purpose is test injection. A common Go pattern to avoid advertising test seams on the production struct is to expose these through **functional constructor options** instead:

```go
type ManagerOption func(*DefaultManager)

func WithRetryConfig(cfg retryConfig) ManagerOption {
    return func(d *DefaultManager) { d.retryConfig = &cfg }
}

func WithRunnerFactory(f mcpRunnerFactory) ManagerOption {
    return func(d *DefaultManager) { d.newRunner = f }
}
```

The `retryConfigOrDefault()` / `newRunnerOrDefault()` helpers can stay as-is. If `DefaultManager` already has a constructor, it just needs a `...ManagerOption` parameter; tests pass the options, production callers pass nothing. This keeps the struct definition free of "nil in production" fields.

pkg/workloads/manager.go
```diff
@@ -449,17 +510,44 @@ func (d *DefaultManager) RunWorkload(ctx context.Context, runConfig *runner.RunC
...
+				// would have kept recovering once the disturbance ended.
+				if stable {
+					slog.Debug("resetting retry counter after stable run",
+						"workload", runConfig.BaseName, "duration", runDuration)
+					// Setting attempt to 0 means the for-clause's attempt++ on `continue`
+					// lands at 1, so the `if attempt > 1` sleep block is skipped and the
+					// next run starts immediately. retryDelay is reset so that any later
+					// failures within this fresh sequence use the initial backoff.
+					attempt = 0
+					retryDelay = cfg.initialDelay
```
amirejaz: The `attempt = 0` trick is correct but relies on a non-obvious interaction with the for-clause increment. Even with the comment, future readers have to hold "the increment runs on `continue`" in their head to follow the reset semantics. A manual loop removes the implicit coupling entirely:

```go
attempt := 1
for attempt <= maxRetries {
    if attempt > 1 {
        // sleep / backoff
    }

    // ... run ...

    if stable {
        attempt = 1          // explicit: restart the sequence
        retryDelay = cfg.initialDelay
        continue             // no hidden increment
    }

    attempt++
}
```

With this shape `attempt = 1` clearly means "start over" and there's nothing to trip over.

amirejaz (commented):
The fix is correct and well-motivated. Test coverage is thorough — 7 table-driven cases, `fakeRunner` handles context cancellation properly, and the mockgen source→reflect switch is the right call. Two things worth addressing before merge.

gkatz2:
Thanks for the careful review. Pushed a fixup commit addressing both points:

**#1 — functional options:** added unexported `managerOption` type and `withRetryConfig` / `withRunnerFactory` helpers. The struct field comments that advertised test-only intent are gone — the options carry that intent now.

One deviation from your literal suggestion: I didn't add `...managerOption` to the four public constructors. Since `managerOption` and the option helpers are unexported, no external caller can construct one — adding a variadic the caller can never use felt like dead API surface. Tests apply options directly: `withRetryConfig(testCfg)(manager)`. Happy to switch to the constructor-variadic shape if you'd prefer.

**#2 — manual loop:** rewrote with explicit `attempt = 1` (reset) and `attempt++` (advance). The `attempt = 0` trick and its explanatory comment are both gone.

Smoke-tested the rebuilt binary end-to-end.

amirejaz (approved):

Issue #4976, opened 2026-04-21
Authserver-driven DCR for upstream OAuth 2.1 MCP servers


jhrozek:
## User Story

**As a** ToolHive operator,
**I want to** use the authserver as a DCR client,
**so that** vMCP can proxy to upstream MCP servers (e.g. Datadog) that require OAuth 2.0 / RFC 7591 without hand-registering a client.

## Context

Today the ToolHive authserver cannot configure an OAuth2 upstream without a pre-issued `client_id`. RFC 7591 Dynamic Client Registration primitives already exist under `pkg/auth/oauth/` but are not reachable from `pkg/authserver/` without introducing an import cycle. This story closes that gap end-to-end across three phases: the primitives are relocated into a new leaf package `pkg/oauthproto/` (alongside a rename of the existing `pkg/oauth/`), the authserver performs RFC 8414 metadata discovery and RFC 7591 registration at startup, credentials are cached (first in-memory, then persistently with memory and Redis backends) keyed on `(issuer, redirect_uri, scopes_hash)`, and the standard upstream OAuth2 flow drives the session after that.

The end-to-end flow this unlocks:

```
MCP Client -> [Corp OIDC] -> vMCP embedded AS -> [Datadog OAuth] -> Datadog MCP Server
```

Datadog's MCP server at `https://mcp.datadoghq.com` is the reference OAuth 2.1 / RFC 7591 target.

**Dependencies**: None

## Acceptance Criteria

Capability-level outcomes a ToolHive operator (or an e2e check) can observe once the work under this story has landed:

- [ ] An operator can configure an OAuth2 upstream in the authserver using only a `discovery_url` (no pre-issued `client_id` / `client_secret`), and the authserver starts successfully.
- [ ] On startup, the authserver performs RFC 8414 authorization server metadata discovery (with RFC 8414 §3.3 issuer-equality validation) and RFC 7591 Dynamic Client Registration against the upstream, then drives the standard upstream OAuth2 flow with the obtained credentials.
- [ ] Configuration validation rejects invalid combinations at startup (e.g., both `ClientID` and `DCRConfig` set, or both `DiscoveryURL` and `RegistrationEndpoint` set inside `DCRConfig`) with a clear error.
- [ ] After a successful registration, restarting the authserver does not re-register at the upstream — the cached credentials (in-memory in the first delivery, persistent afterwards) short-circuit the discovery + registration network calls.
[...]

PR #5186, opened 2026-05-04
Add persistent DCRCredentialStore types and memory backend

### LINKED ISSUE #5183: Persistent DCRCredentialStore: types, interface, and in-memory backend
## Description

Sub-task 1 of 3 for Phase 3 of the DCR story (#4976, parent #4979). Define the persisted `DCRCredentials` value type and the storage-level `DCRCredentialStore` interface in `pkg/authserver/storage/`, and ship the in-process memory implementation that both unit tests and single-replica deployments use. This sub-issue is the foundation the Redis backend (sub-issue 2) and the wiring change (sub-issue 3) build on.

## Context

#4978 (Phase 2) defined the narrow runner-side `DCRCredentialStore` interface — `Get(ctx, DCRKey) (*DCRResolution, bool, error)` and `Put(ctx, DCRKey, *DCRResolution) error` — together with the `DCRKey{Issuer, RedirectURI, ScopesHash}` canonical cache key, and shipped an in-process in-memory stub. Phase 3 promotes the interface and value type into `pkg/authserver/storage/` so the same interface is implemented by both `MemoryStorage` and `RedisStorage`. Full RFC 7592 fields (`RegistrationAccessToken`, `RegistrationClientURI`) are already captured by Phase 2 and round-trip through this PR's value type.

This sub-issue introduces no new protocol behavior. It is a structural refactor of the Phase 2 abstraction onto the existing `pkg/authserver/storage/` shape, plus the in-memory implementation that the Redis backend (sub-issue 2) parallels.

**Dependencies**: ##4978
**Blocks**: sub-issue 2 (Redis), sub-issue 3 (wiring)

### `DCRCredentials` value type — `pkg/authserver/storage/types.go`

- [ ] New exported type `DCRCredentials` is added with the following fields (order may follow the local file convention, but all fields must be present):
  - `Key DCRKey` — canonical cache key (mirrors the `DCRKey` type from `pkg/authserver/runner/dcr_store.go` defined in #4978; Phase 3 consumes the same canonical form — do NOT redefine the hashing).
  - `ProviderName string` — debug / audit only (the upstream `UpstreamRunConfig.Name` that triggered the registration; never used as a primary key).
  - `ClientID string`
  - `ClientSecret string` — `//nolint:gosec // G117` comment consistent with `UpstreamTokens.AccessToken`.
  - `TokenEndpointAuthMethod string`
  - `RegistrationAccessToken string` (RFC 7592 — `//nolint:gosec` as above)
  - `RegistrationClientURI string` (RFC 7592)
  - `AuthorizationEndpoint string`
  - `TokenEndpoint string`
  - `CreatedAt time.Time`
- [ ] Doc comment on `DCRCredentials` states that consumers receive a defensive copy (the type is effectively immutable from the caller's perspective).
- [ ] The existing `DCRKey` type is re-exported from (or imported into) `pkg/authserver/storage/` in a way that keeps the canonical hashing helper in a single place. Acceptable options: (a) move `DCRKey` + its `ScopesHash` constructor from `pkg/authserver/runner/` into `pkg/authserver/storage/` and update the runner to depend on it; (b) leave `DCRKey` in the runner and reference it via type alias in storage. Pick one and document the choice in the commit message. Do NOT duplicate the hashing logic.

### `DCRCredentialStore` interface — `pkg/authserver/storage/types.go`

- [ ] `DCRCredentialStore` interface is declared in `pkg/authserver/storage/types.go` with:
  - `GetDCRCredentials(ctx context.Context, key DCRKey) (*DCRCredentials, error)` returning `ErrNotFound` (already defined in `types.go:36`) when the key is absent.
  - `StoreDCRCredentials(ctx context.Context, creds *DCRCredentials) error`.
  - Optionally `DeleteDCRCredentials(ctx context.Context, key DCRKey) error` if needed for tests; otherwise omit.
- [ ] The interface is added to the `//go:generate mockgen` directive at the top of `types.go` so `task gen` produces a mock in `pkg/authserver/storage/mocks/`.
- [ ] The runner-side `DCRCredentialStore` interface from #4978 is either (a) replaced by this storage-level interface everywhere in `pkg/authserver/runner/`, or (b) kept as a thinner adapter that delegates to the storage-level one. Pick the cleaner of the two — no duplicated abstraction. (Note: full removal of the Phase 2 standalone in-memory stub is sub-issue 3's job; this sub-issue lands the new interface and the new implementation, leaving the wire-up swap to the wiring PR.)

### In-memory backend — `pkg/authserver/storage/memory.go`

- [ ] `MemoryStorage` gains a `dcrCredentials map[DCRKey]*DCRCredentials` field guarded by the existing `sync.RWMutex`.
- [ ] `NewMemoryStorage` initializes the map.
- [ ] `StoreDCRCredentials` performs a defensive copy of the input before storing (matching the `StoreUpstreamTokens` pattern at `memory.go:693-734`).
- [ ] `GetDCRCredentials` returns a defensive copy; returns `ErrNotFound` (wrapped per the existing convention: `fmt.Errorf("%w: ...", ErrNotFound, ...)`) on miss.
- [ ] Compile-time interface-compliance assertion is added at the bottom of the file: `_ DCRCredentialStore = (*MemoryStorage)(nil)`.
- [ ] DCR credentials are intentionally excluded from the periodic `cleanupExpired` loop (they are long-lived; the Redis backend applies TTL via `SetEX` instead of in-process cleanup).

### Regenerate mocks

- [ ] `task gen` is rerun; the new mock for `DCRCredentialStore` appears in `pkg/authserver/storage/mocks/mock_storage.go` and compiles.

### Cross-cutting

1. **`DCRCredentials` value type** — add to `pkg/authserver/storage/types.go` adjacent to `UpstreamTokens`. Mirror the defensive-copy semantics documented for `UpstreamTokens`. Capture `RegistrationAccessToken` / `RegistrationClientURI` verbatim from Phase 2 — they're already captured by `resolveDCRCredentials` and live on `DCRResolution`.

2. **Consolidate `DCRKey`** — decide in this PR whether `DCRKey` lives in `pkg/authserver/storage/` or `pkg/authserver/runner/`. Recommended: move it to `storage/` and have the runner import it, since the storage backends need to hash keys identically and the runner already depends on `storage/` for `createStorage`. If you move it, rewrite the runner's `dcr_store.go` to import the storage-level version and delete the local copy.

3. **`DCRCredentialStore` interface** — add to `pkg/authserver/storage/types.go`. Keep the surface minimal: `Get` + `Store`, nothing else. Wire it into the `mockgen` directive at the top of the file.

4. **Memory backend** — extend `MemoryStorage` with a `dcrCredentials` map guarded by the existing `sync.RWMutex`. Copy-before-mutate on both read and write (see `StoreUpstreamTokens` at `memory.go:693-734` and `GetUpstreamTokens` at `memory.go:739-781` as the reference). DO NOT add the map to `cleanupExpired` — DCR entries are long-lived; the Redis backend uses TTL for the rare case the provider expires the secret.

5. **Mocks** — `task gen` regenerates `pkg/authserver/storage/mocks/mock_storage.go`. Integration consumers in `pkg/authserver/runner/` use the regenerated mock directly — no custom fakes.

6. **Verify** — `task build`, `task test`, `task lint-fix`, `task license-check`. Run the secret-grep assertion from #4978.

### Patterns & Frameworks

- Plain `testing` + `github.com/stretchr/testify` per `.claude/rules/testing.md` — NOT Ginkgo. `require.NoError` over `t.Fatal`. Table-driven tests for the memory backend round-trip.
- Copy-before-mutate per `.claude/rules/go-style.md` §"Copy Before Mutating Caller Input". Apply on both read and write paths, just like the existing `StoreUpstreamTokens` / `GetUpstreamTokens` pair in `memory.go`.
- Write durable first per `.claude/rules/go-style.md` §"Write to Durable Storage Before Updating In-Memory State". The memory backend has a single in-process map; the rule is trivially satisfied here but the invariant must be preserved if a write-through cache is added later.
- SPDX 2-line header on any newly-created Go file. `task license-fix` catches omissions.

### Code Pointers

- `pkg/authserver/storage/types.go:54-81` — `UpstreamTokens` is the closest precedent for a storage value type with sensitive fields (`//nolint:gosec // G117` pattern). Mirror that shape for `DCRCredentials`.
- `pkg/authserver/storage/types.go:19` — `//go:generate mockgen` directive; add `DCRCredentialStore` to the `-source` list (or extend the comma-separated interface list) so `task gen` produces a mock.
- `pkg/authserver/storage/types.go:339-383` — `Storage` superset. DO NOT add DCR methods to `Storage`; keep `DCRCredentialStore` as a parallel, segregated interface (follows the same interface-segregation pattern already used for `UpstreamTokenStorage` / `ClientRegistry` / `UserStorage`).
- `pkg/authserver/storage/memory.go:54-105` — `MemoryStorage` struct + constructor; extend with the `dcrCredentials` map.
- `pkg/authserver/storage/memory.go:693-781` — `StoreUpstreamTokens` / `GetUpstreamTokens` are the definitive in-tree reference for defensive copy + `ErrNotFound` wrapping + `ErrExpired` semantics. Mirror them for the DCR methods (minus the expiry branch — DCR entries don't track access-token-style expiry in-memory).
- `pkg/authserver/storage/memory.go:1174-1181` — compile-time interface assertion block; add `_ DCRCredentialStore = (*MemoryStorage)(nil)`.
- `pkg/authserver/runner/dcr_store.go` — defined by #4978. In this PR, either consolidate into `pkg/authserver/storage/` or leave as a thin alias. Don't keep two competing implementations.
- `pkg/auth/oauth/dynamic_registration.go:142` — `ClientSecretExpiresAt int64` field on the existing DCR response struct; this field will eventually drive the Redis-TTL plumbing (sub-issue 2), so propagate it through `DCRCredentials` (already allowed by #4978's capture of full RFC 7591 fields).

### Component Interfaces

```go
// pkg/authserver/storage/types.go (additions)

// DCRCredentials is the persisted form of a DCR registration result.
// All fields are populated from the RFC 7591 DCR response. The RFC 7592
// management fields (RegistrationAccessToken, RegistrationClientURI) are
// preserved so future rotation / management flows can use them.
//
// Callers receive a defensive copy from the store; mutations do not
// affect persisted state.
type DCRCredentials struct {
    // Key is the canonical cache key: (Issuer, RedirectURI, ScopesHash).
    Key DCRKey

    // ProviderName is the upstream's UpstreamRunConfig.Name. Debug / audit
    // only — never used as a primary key. Two upstreams with different
    // ProviderName but identical Key share one credential record.
    ProviderName string

    ClientID                string
    ClientSecret            string //nolint:gosec // G117: field legitimately holds sensitive data
    TokenEndpointAuthMethod string

    // RegistrationAccessToken and RegistrationClientURI are RFC 7592 fields
    // captured for future management operations (rotation, deletion).
    RegistrationAccessToken string //nolint:gosec // G117: field legitimately holds sensitive data
    RegistrationClientURI   string

    AuthorizationEndpoint string
    TokenEndpoint         string

    // CreatedAt supports the cache-hit stale-age log from #4978.
    CreatedAt time.Time
}

// DCRCredentialStore is a narrow, segregated interface for dynamic-client-
// registration credential persistence. Both MemoryStorage and RedisStorage
// implement it; an authserver backed by Redis shares DCR credentials
// across replicas and restarts.
//
// Cross-replica limitation: sharing DCR credentials does NOT imply cross-
// replica session / token delivery. Callers that need that must still route
// through the proxy runner and (if applicable) pin sessions to a replica.
type DCRCredentialStore interface {
    // GetDCRCredentials returns the credentials for the given key.
    // Returns ErrNotFound if no entry exists for the key.
    // The returned value is a defensive copy.
    GetDCRCredentials(ctx context.Context, key DCRKey) (*DCRCredentials, error)

    // StoreDCRCredentials persists the credentials. Implementations MUST
    // honor DCR client_secret_expires_at as a TTL when the source DCR
    // response provided it (propagated through the DCRResolution /
    // DCRCredentials value). When absent or zero, no TTL is applied.
    StoreDCRCredentials(ctx context.Context, creds *DCRCredentials) error
}
```

## Testing Strategy

**Unit Tests**

- [ ] `pkg/authserver/storage/memory_test.go` (new or extend existing): `StoreDCRCredentials` + `GetDCRCredentials` round-trip (all fields preserved); two distinct `DCRKey` values do not collide; overwrite semantics — a second `Store` with the same `Key` replaces the first; `Get` on an absent key returns `ErrNotFound` via `errors.Is`; returned value is a defensive copy (mutating the returned pointer does not affect subsequent `Get` calls).

**Edge Cases**

- [ ] `ErrNotFound` plumbed through correctly (`errors.Is(err, storage.ErrNotFound)` holds).
- [ ] Mocks for the new interface compile (`task gen` then `task build`) — catches signature-drift between the runner and storage packages.

## Out of Scope

- Redis backend — see sub-issue 2.
- Wiring `EmbeddedAuthServer` to use the storage-level interface and removing the Phase 2 standalone in-memory stub — see sub-issue 3.

## References

- Parent issue: #4979
- Upstream dependency: ##4978 (Phase 2 defines `DCRCredentialStore` + `DCRKey` + `DCRResolution` and captures RFC 7592 fields required here).
- RFC 7591 §3.2.1, RFC 7592.

tgrunnagle:
## Summary

- **Why**: Phase 3 of the authserver-driven DCR story (#4979 / parent #4976) needs the persisted credential store moved into `pkg/authserver/storage/` so both `MemoryStorage` and the future Redis backend (sub-issue 2) can implement the same interface, and so the wiring change (sub-issue 3) has a single canonical type to swap to. Until that lands, every authserver replica re-registers its own DCR client on boot, no `client_secret_expires_at` TTL is honored at the storage layer, and the runner-side cache from #4978 has no production-shaped sibling.
- **What** — adds the storage-layer `DCRCredentials` value type, the segregated `DCRCredentialStore` interface, and a `MemoryStorage` implementation. Production deployments will use the Redis backend introduced in sub-issue 2; the in-memory implementation lands here for tests and single-replica development. The wiring swap is intentionally out of scope.
  - New `DCRCredentials` value type in `pkg/authserver/storage/types.go` carrying the canonical `DCRKey`, `ProviderName`, the RFC 7591 client fields (`ClientID`, `ClientSecret`, `TokenEndpointAuthMethod`), the RFC 7592 management fields (`RegistrationAccessToken`, `RegistrationClientURI`), the upstream-resolved endpoints (`AuthorizationEndpoint`, `TokenEndpoint`), and `CreatedAt` / `ClientSecretExpiresAt` for the staleness and TTL signals.
  - New `DCRCredentialStore` interface (`GetDCRCredentials`, `StoreDCRCredentials`) added to the package's `mockgen` directive; mocks regenerated. The interface doc states the defensive-copy contract, the SHOULD-honor-TTL contract for backends with native TTL support, and the rationale for embedding `DCRKey` in the value type (self-describing blobs for Redis SCAN / admin tooling).
  - `DCRKey` and its `ScopesHash` constructor moved from `pkg/authserver/runner/` into `pkg/authserver/storage/` so any future backend hashes keys identically. The runner keeps a package-local `type DCRKey = storage.DCRKey` alias and call sites use `storage.ScopesHash` directly.
  - `MemoryStorage` gains a `dcrCredentials map[DCRKey]*DCRCredentials` guarded by the existing `sync.RWMutex`. `StoreDCRCredentials` defensively copies on write and rejects nil creds, an unpopulated `Key` (empty `Issuer`, `RedirectURI`, or `ScopesHash`), and missing RFC 7591 mandatory response fields (`ClientID`, `AuthorizationEndpoint`, `TokenEndpoint`); `GetDCRCredentials` defensively copies on read and returns wrapped `ErrNotFound` on miss. DCR entries are intentionally excluded from `cleanupExpired` (registrations are long-lived; Redis will TTL via `SetEX`). `Stats` exposes the new map's count for parity with the other in-memory maps.
  - Renamed the runner-side `DCRCredentialStore` (a runtime resolution cache for `*DCRResolution`) to lowercase `dcrResolutionCache` so it does not collide with the new `storage.DCRCredentialStore` during the Phase 3 migration. The runner-side cache is the thin adapter sub-issue 3 will swap.

Closes #5183

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Specifically:

- `pkg/authserver/storage/memory_test.go` — round-trip on every `DCRCredentials` field including `ClientSecretExpiresAt`; distinct `DCRKey` values do not collide along the Issuer / RedirectURI / Scopes axes; overwrite semantics replace the prior entry; `Get` on an absent key returns wrapped `ErrNotFound` via `errors.Is`; defensive-copy assertions on both read and write paths; `StoreDCRCredentials` rejects nil creds and the seven invalid-input cases (empty Issuer / RedirectURI / ScopesHash / ClientID / AuthorizationEndpoint / TokenEndpoint, plus nil); `Stats` reports the `DCRCredentials` count; DCR entries survive `cleanupExpired`; concurrent fan-out (16 workers × 200 ops, `-race`) over overlapping and disjoint key spaces with a fail-fast deadline.
- `pkg/authserver/storage/memory_test.go` (`TestScopesHash_*`) — canonical scopes-hash form exercised here as the single source of truth (permutation, duplicates, boundary-join collision, distinct-scopes inequality).
- `task gen` rerun; `pkg/authserver/storage/mocks/mock_storage.go` regenerated for the new interface.

The branch is one commit on top of `main`. The runner-side `dcrResolutionCache` is renamed but functionally unchanged — it still holds `*DCRResolution` values and is consumed only by `resolveDCRCredentials`. Sub-issue 3 will swap the resolver to read/write `storage.DCRCredentialStore` and delete the runner-side cache; this PR keeps both side by side under non-colliding names so the intermediate state is unambiguous to reviewers and tooling.

The interface contract documents `ClientSecretExpiresAt` as SHOULD-honor (not MUST), with the in-memory backend retaining the field verbatim and the resolver re-checking on read. The Redis backend (sub-issue 2) plumbs the field into `SetEX` so an entry evicts before the upstream rejects the secret at the token endpoint.


tgrunnagle (commented):
## Multi-agent review summary

6 specialist agents reviewed the diff (Go style/API design, concurrency/locking, OAuth/RFC compliance & security, test coverage, review-friendliness, general quality). 8 findings reached consensus (score ≥ 7); 13 dropped to the appendix below.

**Author note**: I'm the PR author, so this review is posted as `COMMENT` (GitHub rejects APPROVE / REQUEST_CHANGES on your own PR). If a co-reviewer agrees with the HIGH/MEDIUM findings, a formal `REQUEST_CHANGES` is warranted.

### Consensus summary

| ID | Finding | Severity | Score |
|----|---------|----------|-------|
| F1 | PR body advertises scope not present in the diff | HIGH (process) | 7 |
| F2 | Two `DCRCredentialStore` interfaces with the same name in sibling packages | HIGH | 9 |
| F3 | `var scopesHash = storage.ScopesHash` is a mutable runtime binding, not an alias | MEDIUM | 8 |
| F4 | `StoreDCRCredentials` validation is narrower than the doc contract claims | MEDIUM | 9 |
| F5 | "Bounded by upstream count" claim is incorrect when scopes change | MEDIUM | 8 |
| F6 | `StoreDCRCredentials` signature/naming inconsistent with sibling Storage methods | MEDIUM | 7 |
| F7 | Missing concurrent-access test for new DCR Store/Get methods | LOW-MEDIUM | 7 |
| F8 | PR is marked DRAFT but a review was requested | MEDIUM (process) | 7 |

F2–F6 are inline below. F1, F7, F8 are PR-level and addressed here:

---

### F1 — PR body / diff mismatch (HIGH, process)

The Summary enumerates changes in `cmd/thv-operator/api/v1beta1/`, `pkg/auth/monitored_token_source.go`, `pkg/oauthproto/dcr.go`, `pkg/runner/runner.go`, and the v1beta1 CRD/operator paths. None of those files appear in the diff: the actual change is 7 files confined to `pkg/authserver/storage/` and `pkg/authserver/runner/`. Reviewers will read the body, expect operator/runner-side wiring, and either think the diff is incomplete or burn time hunting for code that isn't here.

**Recommendation**: Rewrite the Summary so it describes only what is in the diff (the storage-package DCR types + memory backend + runner-side type aliases). Move the "follow-up stack" narrative to a forward-looking note in `Special notes for reviewers`, not to the Summary.

### F7 — Missing concurrent-access test for new DCR Store/Get methods (LOW-MEDIUM)

`TestMemoryStorage_ConcurrentAccess` doesn't include the new methods, and all the new DCR tests run single-goroutine. A regression that, e.g., dropped the lock or returned an internal pointer instead of a clone would not be caught by the current suite under `-race`. The runner-side `TestInMemoryDCRCredentialStore_ConcurrentAccess` covers a different implementation.

**Recommendation**: Add a fan-out test in `pkg/authserver/storage/memory_test.go` — N goroutines doing `StoreDCRCredentials` with distinct keys, M doing `GetDCRCredentials`, with a timed `wg.Wait()` barrier per `.claude/rules/testing.md`. Pattern after the existing concurrent tests in this package or the runner-side `TestInMemoryDCRCredentialStore_ConcurrentAccess`.

### F8 — PR is marked DRAFT (process)

Body opens with "# DRAFT - not ready for review". Combined with F1, this suggests the body was written before commits were dropped or rebased. The 555-line / 7-file diff is well within the 400-line / 10-file PR-size limit when generated/test files are excluded, so size isn't a blocker.

**Recommendation**: Once F1–F6 are addressed, take the PR out of draft and request reviewers.

---

### Holistic assessment

Architecturally the change is sound: the storage interface is narrowly segregated (`Get`/`Store` only), the defensive-copy contract mirrors the existing `UpstreamTokens` pattern, the cross-replica/TTL contract is documented honestly (the in-memory backend's TTL ignorance is offset by the resolver re-checking `ClientSecretExpiresAt` on read at `dcr.go:672`), and the compile-time interface assertion (`memory.go:1316`) catches drift. The PR is well-scoped against the 400-line/10-file limit when generated and test files are excluded (~250 LOC of production code across 4 files).

The main holistic concern is the **transitional state**: after this PR lands there are two `DCRCredentialStore` interfaces with the same name in sibling packages (`runner.DCRCredentialStore` and `storage.DCRCredentialStore`). The PR body acknowledges this is intentional during sub-issue 2/3, but the chosen name collision is the most reviewer-hostile option — it makes every subsequent review of intermediate PRs require mental disambiguation. F2 below proposes the cheapest fix.

### Dropped findings appendix (score < 7)

| Title | Severity | Reason dropped |
|-------|----------|----------------|
| `cloneDCRCredentials` shallow copy brittle to future fields | LOW | Correct today; doc warns; reflection-test mitigation is optional |
| Silent overwrite when two upstreams collide on Key but differ in ProviderName | LOW | Improbable; a slog.Warn on collision would suffice |
| `ExcludedFromCleanupExpired` test uses misleading stale `CreatedAt` | LOW | Cosmetic: assertion is correct, just the comment misleads |
| `Stats().DCRCredentials` lacks positive-count assertion | LOW | One-line test addition; nice-to-have |
| `ScopesHash` empty-input non-emptiness not asserted | LOW | One `assert.NotEmpty`; nice-to-have |
| Inconsistent "DCR"/"dcr" casing in error hints | LOW | Cosmetic |
| Comment-only placeholder where deleted tests used to live | LOW | Style preference |
| Doc comment redundancy between canonical type and runner-side alias | LOW | Style preference |
| `DCRCredentials.ClientSecret` no `String()`/`Marshal` redaction | LOW | Defense-in-depth; today's code doesn't need it |
| `DCRCredentials.ProviderName` stored but never read | LOW | YAGNI; would be consumed by a future PR |
| `Stats` struct addition compat | INFO | Project uses keyed struct literals — non-issue |
| G117 nolint cites wrong upstream rule | INFO | Resolved — follows established repo convention per `.golangci.yml` |
| `slog.Debug` call inside RLock on miss path | LOW | Resolved — matches existing pattern (memory.go:345/490/545/...) |


---

## File-by-file findings (would-be inline comments)

GitHub's inline-comment API rejected the batched and per-comment posts with a `pull_request_review_thread.base` internal-error, so the inline findings are folded into the main body below. Each block names the exact file and line.

### `pkg/authserver/storage/types.go:263`

**F2 — Naming collision with `runner.DCRCredentialStore` (HIGH, score 9)**

After this PR lands there are two interfaces sharing this name in sibling packages:
- `runner.DCRCredentialStore` (`pkg/authserver/runner/dcr_store.go:57`) — `Get`/`Put` returning `*DCRResolution`, the in-process resolver cache.
- `storage.DCRCredentialStore` (here) — `GetDCRCredentials`/`StoreDCRCredentials` returning `*DCRCredentials`, the persistent contract.

The runner package imports `pkg/authserver/storage`, so unqualified references inside `runner/` are ambiguous to readers and grep tooling. The PR body documents this is intentional during sub-issue 2/3, but the same-name collision is the worst possible outcome for review-friendliness — it prevents the compiler/reader from telling them apart at a glance, and the two will inevitably drift before sub-issue 3 swaps them.

This is the cross-cutting `.claude/rules/go-style.md` "Avoid parallel types that drift" failure mode.

**Recommendation**: Rename the runner-side interface (the cheap path) to `dcrResolutionCache` (lowercase — only the resolver consumes it). Reserve the canonical `DCRCredentialStore` name for the storage package the future Redis backend will implement. Doing this in *this* PR costs one rename and removes ambiguity from every subsequent review.


### `pkg/authserver/storage/types.go:272`

**F6 — Method signature & naming inconsistent with sibling Storage methods (MEDIUM, score 7)**

Two related drifts from the package's local conventions:

1. **Signature shape**: Sibling methods take `(ctx, key, value)`:
   - `StorePendingAuthorization(ctx, state, pending)`
   - `StoreUpstreamTokens(ctx, sessionID, providerName, tokens)`
   - `StoreAccessTokenSession(ctx, signature, request)`

   This method takes `(ctx, creds)` with the key embedded as `creds.Key`. The embedded `Key` doubles storage and creates a structural opportunity for caller/key inconsistency — a caller could plausibly construct `&DCRCredentials{Key: keyA, ...}` and expect Store to use a separate key parameter; the implementation insists on `creds.Key`, but only because the parameter list doesn't expose the alternative.

2. **Method names**: Sibling methods omit the type suffix because the verb implies the subject — `StoreAuthCode`/`GetAuthCode`, `StoreUpstreamTokens`/`GetUpstreamTokens`, `GetUser`, `GetClient`. `StoreDCRCredentials`/`GetDCRCredentials` over-specifies.

**Recommendation**:
```go
// Suggested
type DCRCredentialStore interface {
    GetDCR(ctx context.Context, key DCRKey) (*DCRCredentials, error)
    StoreDCR(ctx context.Context, key DCRKey, creds *DCRCredentials) error
}
```
...and drop `DCRCredentials.Key` accordingly. Worth fixing before the interface is exported into the wiring PR (sub-issue 3) since renaming after merge churns more files. If retaining the embedded `Key` is intentional for self-describing Redis serialisation, document the rationale on the interface so future readers see the reason for the inconsistency.


### `pkg/authserver/storage/memory.go:97`

**F5 — "Bounded by upstream count" claim is incorrect when scopes change (MEDIUM, score 8)**

The comment asserts: *"The map is bounded by the operator-configured upstream count, so unbounded growth is not a concern."*

This is only true if `(Issuer, RedirectURI, ScopesHash)` is invariant per upstream over the process lifetime. `ScopesHash` changes whenever:
- the operator widens/narrows the scope set on `MCPServer.Spec.OAuthConfig.Scopes`
- the upstream's RFC 8414 `scopes_supported` rotates and the resolver re-derives the scope set

Each distinct `ScopesHash` produces a fresh entry; nothing evicts the prior entry (`cleanupExpired` deliberately skips this map per the comment just below). In a long-lived process where scope sets rotate, growth is `upstream count × distinct-scope-sets-ever-registered`, not `upstream count`. Each retained entry holds a `ClientSecret`, so this is also a (slow) sensitive-data accumulation.

Same false claim is also at `pkg/authserver/runner/dcr_store.go:71`.

**Recommendation**: Tighten the comment:

```suggestion
	// dcrCredentials maps DCRKey -> DCRCredentials for RFC 7591 Dynamic Client
	// Registration credentials. Entries are intentionally excluded from the
	// periodic cleanupExpired loop: DCR registrations are long-lived and the
	// authoritative expiry signal is RFC 7591 client_secret_expires_at, which
	// is honored at read time by callers (and by the future Redis backend's
	// SetEX TTL). Growth is bounded by the operator-configured upstream count
	// times the number of distinct scope sets ever registered against each
	// upstream during the process lifetime; for a stable configuration this
	// is the upstream count, but rotating scope sets accumulate stale entries
	// that survive until process restart. The Redis backend's SetEX TTL
	// mitigates this in production deployments.
```


### `pkg/authserver/storage/memory.go:1231`

**F4 — Validation is narrower than the documented contract (MEDIUM, score 9)**

The doc says *"callers must populate Key"* but the implementation only validates `Key.Issuer != ""` and `Key.RedirectURI != ""`.

- **`Key.ScopesHash`** is silently accepted as empty even though the field doc on `DCRKey` (in `types.go`) says *"Use ScopesHash() to compute this value — do NOT hash scopes by hand at call sites."* A caller that forgets to populate `ScopesHash` gets a successful Store under the empty-string slot; a sibling that *did* compute `ScopesHash(nil)` (a non-empty digest) lands on a different slot. The two records refer to the same logical scope set but live in different cache cells.
- **`ClientID`, `AuthorizationEndpoint`, `TokenEndpoint`** are RFC 7591 mandatory response fields. The struct doc says *"All fields are populated from the upstream's DCR response."* An empty value is a caller bug, not a real registration — but Store accepts it. A subsequent Get returns a working-looking record that fails downstream at the upstream's token endpoint, with no audit trail tying the failure back to the bad Store.

Violates `.claude/rules/go-style.md` *"Validate Parsed Results"* / *"Constructor Validation: Fail Loudly on Invalid Input"*.

**Recommendation**: Tighten validation — reject empty `Key.ScopesHash`, `ClientID`, `AuthorizationEndpoint`, and `TokenEndpoint` with `fosite.ErrInvalidRequest.WithHint(...)` matching the existing pattern at lines 1235–1243. Leave `ClientSecret` permissive (RFC 7591 §2 public clients with `none` auth method legitimately have no secret).


### `pkg/authserver/runner/dcr.go:235`

**F3 — `var scopesHash = storage.ScopesHash` is a mutable runtime binding, not an alias (MEDIUM, score 8)**

Unlike `type DCRKey = storage.DCRKey` in `dcr_store.go:27` (a true compile-time type alias — sound, identity-preserving), this is a package-level *variable* whose value happens to be a function. Three concrete drawbacks:

1. **Mutable**: any code in the runner package can reassign it at runtime. No compile-time guard.
2. **Indirection**: extra hop on every call; tempts future contributors to add behaviour by reassigning "the local scopesHash".
3. **Tooling**: `gopls` find-references, refactoring tools, and `go doc` all see the variable, not `storage.ScopesHash`. The canonical-source-of-truth motivation argues *for* visible qualification, not against it.

The alias semantics are also asymmetric with the type alias in the same package — readers see two different aliasing strategies for two related symbols.

**Recommendation** (pick one):

*Option A* — qualify call sites:
```go
// at every call site in runner/dcr.go and runner/dcr_store_test.go
storage.ScopesHash(scopes)
```
Mechanical edit, removes the indirection, makes the storage dependency visible.

*Option B* — forwarding func (immutable, gets inlined):
```go
func scopesHash(s []string) string { return storage.ScopesHash(s) }
```

Option A preferred — the entire reason `ScopesHash` moved to `storage/` was to make it the single source of truth, so the qualifier is informational, not noise.

tgrunnagle:
Addressed all 8 findings from the multi-agent review (`pullrequestreview-4237822982`).

| ID | Finding | Resolution |
|----|---------|-----------|
| F1 (HIGH, process) | PR body / diff mismatch | Body rewritten to describe only what's in this PR; forward-looking notes moved to "Special notes for reviewers". |
| F2 (HIGH, score 9) | Naming collision with `runner.DCRCredentialStore` | Renamed runner-side interface to lowercase `dcrResolutionCache` (and `inMemoryDCRResolutionCache` / `newInMemoryDCRResolutionCache`). Storage-side `DCRCredentialStore` retains the canonical name for the future Redis backend. Commit `96eac7b4`. |
| F3 (MEDIUM, score 8) | `var scopesHash` is a mutable runtime binding | Removed the var; runner call sites now invoke `storage.ScopesHash` directly. Commit `29d1c917`. |
| F4 (MEDIUM, score 9) | `StoreDCRCredentials` validation narrower than the doc contract | Now also rejects empty `Key.ScopesHash`, `ClientID`, `AuthorizationEndpoint`, `TokenEndpoint`. `ClientSecret` left permissive (RFC 7591 §2 public clients). Table-driven test extended to cover all seven invalid-input cases. Commit `012e2b30`. |
| F5 (MEDIUM, score 8) | "Bounded by upstream count" comment incorrect when scopes change | Both parallel comments (`memory.go:97` and `runner/dcr_store.go:61`) now state growth is upstream count × distinct-scope-sets-ever-registered, with the Redis SetEX TTL mitigation called out. Commit `22ab153e`. |
| F6 (MEDIUM, score 7) | Signature/naming inconsistent with sibling Storage methods | Documented the rationale on the interface — embedding `DCRKey` in `DCRCredentials` makes the persisted blob self-describing for Redis SCAN / admin tooling, which is load-bearing for sub-issue 2. Commit `72337f47`. |
| F7 (LOW-MEDIUM, score 7) | Missing concurrent-access test for new methods | Added `TestMemoryStorage_DCRCredentials_ConcurrentAccess` — 16 workers × 200 ops over overlapping and disjoint key spaces, fail-fast deadline, passes under `-race`. Commit `409a66e0`. |
| F8 (MEDIUM, process) | PR marked DRAFT but review requested | PR is now out of draft. |

`task lint-fix` clean (0 issues); `go test -count=1 -race ./pkg/authserver/...` green across all packages.

jhrozek (approved):
LGTM. Spot-checked the three risk areas locally — ScopesHash/DCRKey move is byte-identical, all dcrCredentials accesses on MemoryStorage are properly locked (and cleanupExpired doesn't touch it), and the runner-side rename is text-only.

Few small things for a follow-up if you want, none blocking:
- the slog.Debug in GetDCRCredentials on miss is noisy and out of step with the other Get* methods in memory.go
- sort.Strings → slices.Sort in ScopesHash would let you drop the sort import
- a positive test pinning "empty ClientSecret is allowed" would prevent a future tightening from silently breaking public clients
- a LogValuer on DCRCredentials would guard against accidental secret leaks via %+v or default slog

PR #5195, opened 2026-05-05
Add Redis backend for DCRCredentialStore

### LINKED ISSUE #5184: Persistent DCRCredentialStore: Redis backend
## Description

Sub-task 2 of 3 for Phase 3 of the DCR story (#4976, parent #4979). Implement the Redis backend for `DCRCredentialStore` (defined in sub-issue 1), so an authserver backed by Redis Sentinel shares DCR credentials across replicas and survives restarts. This is the persistent half of the parent issue's deliverable; sub-issue 3 wires it up to `EmbeddedAuthServer`.

## Context

#4978 (Phase 2) shipped an in-process in-memory `DCRCredentialStore` stub. Sub-issue 1 of this Phase 3 promotes the interface and value type into `pkg/authserver/storage/` and lands the in-memory implementation. This sub-issue adds the Redis Sentinel implementation behind the same interface, so a single `storage_type: redis` config toggles DCR persistence alongside the rest of authserver state.

The Redis backend must:

- Match the existing `pkg/authserver/storage/` shape and follow the prior-art Redis key scheme `thv:auth:{ns:name}:<type>:<id>` (see `pkg/authserver/storage/redis_keys.go:63-83`) so DCR keys share the `{ns:name}` Redis Cluster hash tag with the rest of a server's state.
- Honor `client_secret_expires_at` (RFC 7591 §3.2.1, unix seconds; `0` means "never") as the Redis TTL when the authorization server provides it. When absent or `0`, no TTL is applied and entries are long-lived.
- Obey `.claude/rules/go-style.md` §"Write to Durable Storage Before Updating In-Memory State" — durable first.

This sub-issue introduces no new protocol behavior. It is a drop-in persistence layer behind the interface from sub-issue 1.

**Dependencies**: sub-issue 1 (types, interface, in-memory implementation)
**Blocks**: sub-issue 3 (wiring)

### Redis backend — `pkg/authserver/storage/redis.go` + `redis_keys.go`

- [ ] Add `KeyTypeDCR = "dcr"` to `pkg/authserver/storage/redis_keys.go` adjacent to the existing `KeyType*` consts.
- [ ] Add a DCR key helper (e.g. `redisDCRKey(prefix string, key DCRKey) string`) that produces `{prefix}dcr:<id>` where `<id>` is a deterministic, collision-free encoding of `(issuer, redirect_uri, scopes_hash)`. Two safe encodings: (a) a length-prefixed format mirroring `redisProviderKey` at `redis_keys.go:87-89`; (b) a single SHA-256 hex over the tuple. Either is acceptable — pick one and document it adjacent to the helper. DO NOT use plain `fmt.Sprintf("%s:%s:%s", ...)` because `redirect_uri` contains colons.
- [ ] `RedisStorage.StoreDCRCredentials` JSON-marshals the `DCRCredentials` value and calls `SetEX` with a TTL derived from the upstream's `client_secret_expires_at`:
  - When the AS advertised `client_secret_expires_at > 0` (captured at registration time — plumb it through `DCRCredentials` or compute `ttl = expiresAt - time.Now()` at the call site and document the source), use `ttl` as the Redis TTL.
  - When `client_secret_expires_at == 0` or not provided, use `Set` with no expiration (long-lived).
- [ ] `RedisStorage.GetDCRCredentials` JSON-unmarshals the value; returns `ErrNotFound` on `redis.Nil`; returns the unwrapped error otherwise.
- [ ] Compile-time interface-compliance assertion is added at the bottom of the file: `_ DCRCredentialStore = (*RedisStorage)(nil)`.
- [ ] Connection-error handling follows the existing `drainCloseErr`/`warnOnCleanupErr` patterns already used in `redis.go`.

### Cross-cutting

1. **Redis backend** — add `KeyTypeDCR` to `redis_keys.go`. Add a DCR key helper that handles colons in `redirect_uri` safely (length-prefix form or SHA-256 hex; pick one). `StoreDCRCredentials` JSON-marshals the value and calls `SetEX` with a TTL when `client_secret_expires_at > 0`, or `Set` with no expiration otherwise. `GetDCRCredentials` handles `redis.Nil` as `ErrNotFound`.

2. **Verify** — `task build`, `task test`, `task lint-fix`, `task license-check`. Run the secret-grep assertion from #4978.

### Patterns & Frameworks

- Plain `testing` + `github.com/stretchr/testify` per `.claude/rules/testing.md` — NOT Ginkgo. `require.NoError` over `t.Fatal`. Table-driven tests for the Redis key-encoding helper.
- Redis integration tests use **testcontainers** (consistent with `pkg/authserver/storage/redis_integration_test.go`), NOT `miniredis`. The existing helper `newRedisSentinelCluster` + `withIntegrationStorage` is the shape to follow; add DCR coverage there rather than spinning up a separate harness. `miniredis` is not in `go.mod` and should not be added.
- Integration tests live behind the `//go:build integration` tag already used by `redis_integration_test.go`. Extend that file for the Redis DCR coverage.
- Write durable first per `.claude/rules/go-style.md` §"Write to Durable Storage Before Updating In-Memory State". This backend is a single durable write; no in-process cache sits in front of it, so the rule is trivially satisfied — but the invariant must be preserved if a write-through cache is added later.
- SPDX 2-line header on any newly-created Go file. `task license-fix` catches omissions.

### Code Pointers

- `pkg/authserver/storage/redis_keys.go:10-61` — `KeyType*` consts; add `KeyTypeDCR = "dcr"` here.
- `pkg/authserver/storage/redis_keys.go:85-89` — `redisProviderKey` length-prefix pattern for colon-containing inputs; follow the same shape if picking option (a) for the DCR key helper.
- `pkg/authserver/storage/redis.go:35-51` — `warnOnCleanupErr` + the best-effort-index-cleanup convention; DCR has no secondary index, so this is a one-pass `Set`/`Get`, but preserve the error-handling idioms.
- `pkg/authserver/storage/redis.go:103-110` — `RedisStorage` struct; add DCR methods as a new section after the upstream-token methods (follow the existing `// ---` section-header comment style).
- `pkg/authserver/storage/redis_integration_test.go:29-35` (`testcontainers` imports) and the `newRedisSentinelCluster` / `withIntegrationStorage` helpers in the same file — extend these rather than adding a second integration-test harness.
- `pkg/auth/oauth/dynamic_registration.go:142` — `ClientSecretExpiresAt int64` field on the existing DCR response struct; this is the source of the Redis-TTL plumbing. Propagate it through `DCRResolution` / `DCRCredentials` (already allowed by #4978's capture of full RFC 7591 fields) or compute `ttl` at the call site when storing.

### Component Interfaces

```go
// pkg/authserver/storage/redis_keys.go (addition)

// KeyTypeDCR is the key type for DCR credentials.
const KeyTypeDCR = "dcr"

// redisDCRKey generates a Redis key for a DCR credential entry.
// Uses a length-prefixed encoding to handle colons in RedirectURI safely,
// mirroring redisProviderKey's approach.
//
// Format: "{prefix}dcr:<len(issuer)>:<issuer>:<len(redirect)>:<redirect>:<scopes_hash>"
// The trailing scopes_hash is already SHA-256 hex and contains no colons.
func redisDCRKey(prefix string, key DCRKey) string {
    return fmt.Sprintf("%s%s:%d:%s:%d:%s:%s",
        prefix, KeyTypeDCR,
        len(key.Issuer), key.Issuer,
        len(key.RedirectURI), key.RedirectURI,
        key.ScopesHash)
}
```

## Testing Strategy

**Unit Tests**

- [ ] `pkg/authserver/storage/redis_keys_test.go` (new or extend existing): `redisDCRKey` is deterministic; distinct `DCRKey` tuples produce distinct Redis keys; keys containing colons in `RedirectURI` do not collide.

**Integration Tests**

- [ ] `pkg/authserver/storage/redis_integration_test.go` (extend): Round-trip `DCRCredentials` through the real Redis Sentinel cluster; two distinct keys coexist; overwrite on matching key updates the value; `StoreDCRCredentials` with a non-zero `client_secret_expires_at` plumbed in results in an observable Redis TTL on the key (use `TTL` command through the test client to assert); `StoreDCRCredentials` without a TTL results in no expiration (`TTL` returns `-1`); concurrent `Put` / `Get` from multiple goroutines is race-free under `go test -race`.

**Edge Cases**

- [ ] `ErrNotFound` plumbed through correctly (`errors.Is(err, storage.ErrNotFound)` holds).
- [ ] `client_secret_expires_at = 0` (RFC 7591 "never") means no Redis TTL, not a zero TTL (which would be immediate expiry).
- [ ] A DCR key colliding with an unrelated Redis key type (e.g. `upstream:...`) is impossible because the `KeyTypeDCR = "dcr"` segment is distinct from all other `KeyType*` values in `redis_keys.go`.

## Out of Scope

- In-memory backend, types, and `DCRCredentialStore` interface — see sub-issue 1.
- Wiring `EmbeddedAuthServer` to use the storage-level interface and removing the Phase 2 standalone in-memory stub — see sub-issue 3.

## References

- Parent issue: #4979
- Upstream dependency: ##4978 (Phase 2).
- RFC 7591 §3.2.1 (`client_secret_expires_at` semantics: `0` means "never", unix seconds otherwise).
- RFC 7592 (DCR Management Protocol — fields persisted, management endpoints deferred).

tgrunnagle:
## Summary

An authserver replica that registers itself as a DCR client against an upstream
authorization server currently keeps the resulting `(client_id, client_secret)`
in the in-process `MemoryStorage` from sub-issue 1. Restarts and horizontal
scale-outs lose the registration, forcing every replica to re-register on cold
start and breaking RFC 7592 management URLs. This PR adds the persistent half
of `DCRCredentialStore` so a Redis-Sentinel-backed authserver shares DCR
credentials across replicas and survives restarts.

- Add `KeyTypeDCR` and a length-prefixed `redisDCRKey(prefix, DCRKey)` helper
  that handles colons in `RedirectURI` and `Issuer` without ambiguity, mirroring
  the existing `redisProviderKey` shape.
- Add `RedisStorage.StoreDCRCredentials` / `GetDCRCredentials` with JSON
  serialisation (acting as a defensive copy on read) and TTL semantics derived
  from `client_secret_expires_at`: zero means no Redis TTL (RFC 7591 "never"),
  a future expiry uses `time.Until(expiry)`, and an already-past expiry is
  bounded to a 1s eviction window so already-dead secrets do not linger.
- Wire a compile-time `_ DCRCredentialStore = (*RedisStorage)(nil)` assertion
  alongside the existing interface checks at the bottom of `redis.go`.
- Add miniredis-backed unit tests covering key distinctness/determinism, full
  round-trip, overwrite, validation, defensive copy via decode, all three TTL
  branches, `ErrNotFound` semantics, and concurrent `Store` / `Get`.
- Add Redis Sentinel integration tests (testcontainers, `//go:build integration`)
  pinning the wire-level TTL contract against real Redis (`-1` for never-expires)
  plus round-trip, distinct-keys, overwrite, and concurrent access — extending
  `redis_integration_test.go` rather than introducing a second harness.

Closes #5184

## Type of change

- [x] New feature

## Test plan

Ran the integration suite locally against a Docker-backed Redis Sentinel
cluster: `go test -tags=integration ./pkg/authserver/storage/...` (TTL,
round-trip, distinct-keys, overwrite, and concurrent-access cases all pass,
including the wire-level `TTL == -1` assertion for never-expires rows).

## API Compatibility

## Changes

| File | Change |
|------|--------|
| `pkg/authserver/storage/redis_keys.go` | Add `KeyTypeDCR` const and `redisDCRKey` length-prefixed key helper. |
| `pkg/authserver/storage/redis.go` | Add `storedDCRCredentials` wire type, `StoreDCRCredentials` / `GetDCRCredentials`, `pastExpiryDCRTTL` bound, and `DCRCredentialStore` interface assertion. |
| `pkg/authserver/storage/redis_test.go` | Add miniredis unit coverage: key encoding, round-trip, overwrite, validation, defensive copy, TTL (never / future / past), `ErrNotFound`, concurrent access. |
| `pkg/authserver/storage/redis_integration_test.go` | Add Redis Sentinel integration coverage extending `withIntegrationStorage`: round-trip, distinct keys, overwrite, real-Redis TTL, concurrent access. |

No. This is internal storage plumbing behind the `DCRCredentialStore` interface
introduced in sub-issue 1. Sub-issue 3 will wire `EmbeddedAuthServer` to select
the Redis backend via the existing `storage_type: redis` config toggle.

- This branch went through two review iterations. The final state has zero
  CRITICAL / HIGH / MEDIUM findings. One LOW finding remains: a stale docstring
  on a unit-test helper in `redis_test.go` that still references the
  pre-`pastExpiryDCRTTL` behaviour. Happy to fix in this PR or as a follow-up
  if reviewers prefer to keep this PR focused on the storage primitive.
- Past-expiry TTL behaviour is deliberately `1s` rather than rejecting the
  write or storing long-lived. Rationale is in the `pastExpiryDCRTTL` constant
  doc comment and the `StoreDCRCredentials` docstring: caller's expiry
  timestamp still round-trips so a downstream reader can observe it and trigger
  re-registration, while the row self-evicts almost immediately. Worth a look
  during review to confirm the policy matches how sub-issue 3's resolver will
  consume it.
- Unit tests use `miniredis` (already in `go.mod` from the surrounding
  `redis_test.go` suite). Integration tests use the existing testcontainers
  Redis Sentinel harness — both layers are required: the unit layer pins
  in-process semantics for `task test`, and the integration layer pins the
  wire contract (`TTL` returns `-1` for "no TTL") that miniredis cannot
  faithfully reproduce.
- No new dependencies were introduced.

### DIFF
--- a/pkg/authserver/storage/redis.go
+++ b/pkg/authserver/storage/redis.go
@@ -1447,17 +1447,11 @@ func (s *storedDCRCredentials) toDCRCredentials() *DCRCredentials {
 // fail-loud-but-tolerant posture: the caller's expiry timestamp round-trips so
 // a downstream reader can still observe it and trigger re-registration.
 //
-// Returns fosite.ErrInvalidRequest for nil creds, empty Issuer, or empty
-// RedirectURI — the same fail-loud contract as MemoryStorage.StoreDCRCredentials.
+// Validation is delegated to validateDCRCredentialsForStore so the rejection
+// set stays in sync with MemoryStorage and any future backend.
 func (s *RedisStorage) StoreDCRCredentials(ctx context.Context, creds *DCRCredentials) error {
-	if creds == nil {
-		return fosite.ErrInvalidRequest.WithHint("dcr credentials cannot be nil")
-	}
-	if creds.Key.Issuer == "" {
-		return fosite.ErrInvalidRequest.WithHint("dcr credentials key issuer cannot be empty")
-	}
-	if creds.Key.RedirectURI == "" {
-		return fosite.ErrInvalidRequest.WithHint("dcr credentials key redirect_uri cannot be empty")
+	if err := validateDCRCredentialsForStore(creds); err != nil {
+		return err
 	}
 
 	key := redisDCRKey(s.keyPrefix, creds.Key)
@@ -1511,10 +1505,10 @@ func (s *RedisStorage) StoreDCRCredentials(ctx context.Context, creds *DCRCreden
 // Returns ErrNotFound (wrapped) when no entry exists. The returned value is a
 // fresh struct decoded from JSON, which acts as a defensive copy.
 //
-// An unpopulated key (empty Issuer or empty RedirectURI) cannot match any
-// stored row because StoreDCRCredentials rejects such keys, so a Get against
-// one is a normal miss — ErrNotFound — matching MemoryStorage.GetDCRCredentials
-// and the DCRCredentialStore interface contract.
+// An unpopulated key (empty Issuer, RedirectURI, or ScopesHash) cannot match
+// any stored row because StoreDCRCredentials rejects such keys, so a Get
+// against one is a normal miss — ErrNotFound — matching
+// MemoryStorage.GetDCRCredentials and the DCRCredentialStore interface contract.
 func (s *RedisStorage) GetDCRCredentials(ctx context.Context, key DCRKey) (*DCRCredentials, error) {
 	redisKey := redisDCRKey(s.keyPrefix, key)
 	data, err := s.client.Get(ctx, redisKey).Bytes()
--- a/pkg/authserver/storage/redis.go
+++ b/pkg/authserver/storage/redis.go
@@ -32,6 +32,17 @@ const (
 // nullMarker is used to store nil upstream tokens in Redis.
 const nullMarker = "null"
 
+// pastExpiryDCRTTL is the bounded TTL applied when a caller writes DCR
+// credentials whose ClientSecretExpiresAt is already in the past. The row is
+// still accepted (the resolver decides when to re-register) but it self-evicts
+// almost immediately so the store does not hold an already-expired secret
+// forever. One second is small enough that no caller can usefully read the
+// row, large enough that real Redis applies the TTL reliably (sub-second TTLs
+// are technically supported via PEXPIRE but the second-grain TTL command is
+// the broadly-tested path), and short enough that operational metrics do not
+// confuse this row with a healthy long-lived registration.
+const pastExpiryDCRTTL = time.Second
+
 // warnOnCleanupErr logs a warning when a best-effort cleanup operation fails.
 //
 // Secondary index cleanup in Redis (SRem from reverse-lookup sets, Del of orphaned
@@ -1353,6 +1364,176 @@ func unmarshalUpstreamTokens(data []byte) (*UpstreamTokens, error) {
 }
 
 // -----------------------
+// DCR Credentials Storage
+// -----------------------
+
+// storedDCRCredentials is the on-the-wire JSON representation of DCRCredentials.
+// Time fields use int64 Unix epoch; 0 is the sentinel meaning "not set", matching
+// the storedUpstreamTokens convention. ClientSecretExpiresAt == 0 specifically
+// encodes the RFC 7591 §3.2.1 "client_secret does not expire" semantics, in
+// which case StoreDCRCredentials persists the entry without a Redis TTL.
+type storedDCRCredentials struct {
+	// Embed the canonical key so a row recovered without its lookup key
+	// (e.g. via SCAN during diagnostics) still self-identifies.
+	KeyIssuer      string `json:"key_issuer"`
+	KeyRedirectURI string `json:"key_redirect_uri"`
+	KeyScopesHash  string `json:"key_scopes_hash"`
+
+	ProviderName string `json:"provider_name,omitempty"`
+
+	ClientID                string `json:"client_id"`
+	ClientSecret            string `json:"client_secret,omitempty"` //nolint:gosec // G117: field legitimately holds sensitive data
+	TokenEndpointAuthMethod string `json:"token_endpoint_auth_method,omitempty"`
+
+	// Bearer token for the RFC 7592 management endpoint.
+	//nolint:gosec // G117: field legitimately holds sensitive data
+	RegistrationAccessToken string `json:"registration_access_token,omitempty"`
+	RegistrationClientURI   string `json:"registration_client_uri,omitempty"`
+
+	AuthorizationEndpoint string `json:"authorization_endpoint,omitempty"`
+	TokenEndpoint         string `json:"token_endpoint,omitempty"`
+
+	CreatedAt             int64 `json:"created_at"`
+	ClientSecretExpiresAt int64 `json:"client_secret_expires_at"`
+}
+
+// toDCRCredentials decodes the stored form back into the public type, mirroring
+// storedUpstreamTokens.toUpstreamTokens. Zero epoch values become the zero
+// time.Time, preserving the "not set" sentinel.
+func (s *storedDCRCredentials) toDCRCredentials() *DCRCredentials {
+	var createdAt time.Time
+	if s.CreatedAt != 0 {
+		createdAt = time.Unix(s.CreatedAt, 0)
+	}
+	var clientSecretExpiresAt time.Time
+	if s.ClientSecretExpiresAt != 0 {
+		clientSecretExpiresAt = time.Unix(s.ClientSecretExpiresAt, 0)
+	}
+	return &DCRCredentials{
+		Key: DCRKey{
+			Issuer:      s.KeyIssuer,
+			RedirectURI: s.KeyRedirectURI,
+			ScopesHash:  s.KeyScopesHash,
+		},
+		ProviderName:            s.ProviderName,
+		ClientID:                s.ClientID,
+		ClientSecret:            s.ClientSecret,
+		TokenEndpointAuthMethod: s.TokenEndpointAuthMethod,
+		RegistrationAccessToken: s.RegistrationAccessToken,
+		RegistrationClientURI:   s.RegistrationClientURI,
+		AuthorizationEndpoint:   s.AuthorizationEndpoint,
+		TokenEndpoint:           s.TokenEndpoint,
+		CreatedAt:               createdAt,
+		ClientSecretExpiresAt:   clientSecretExpiresAt,
+	}
+}
+
+// StoreDCRCredentials persists DCR credentials, overwriting any existing entry
+// for the same Key. Defensive copy is provided implicitly by JSON serialisation —
+// caller mutations after the call cannot reach the persisted bytes.
+//
+// # TTL
+//
+// When creds.ClientSecretExpiresAt is non-zero (the upstream advertised an
+// RFC 7591 §3.2.1 client_secret_expires_at), the entry is stored with a Redis
+// TTL derived from time.Until(ClientSecretExpiresAt) so the row evicts before
+// the upstream rejects the secret at the token endpoint. When zero (RFC 7591
+// "never"), Set with TTL=0 is used and the entry is long-lived.
+//
+// If ClientSecretExpiresAt is already in the past at call time, the entry is
+// written with the bounded TTL pastExpiryDCRTTL (1 second) rather than rejected
+// or stored long-lived. This keeps the store's "already-expired secret" window
+// narrow even if the resolver never re-reads the row, and matches the
+// fail-loud-but-tolerant posture: the caller's expiry timestamp round-trips so
+// a downstream reader can still observe it and trigger re-registration.
+//
+// Returns fosite.ErrInvalidRequest for nil creds, empty Issuer, or empty
+// RedirectURI — the same fail-loud contract as MemoryStorage.StoreDCRCredentials.
+func (s *RedisStorage) StoreDCRCredentials(ctx context.Context, creds *DCRCredentials) error {
+	if creds == nil {
+		return fosite.ErrInvalidRequest.WithHint("dcr credentials cannot be nil")
+	}
+	if creds.Key.Issuer == "" {
+		return fosite.ErrInvalidRequest.WithHint("dcr credentials key issuer cannot be empty")
+	}
+	if creds.Key.RedirectURI == "" {
+		return fosite.ErrInvalidRequest.WithHint("dcr credentials key redirect_uri cannot be empty")
+	}
+
+	key := redisDCRKey(s.keyPrefix, creds.Key)
+
+	stored := storedDCRCredentials{
+		KeyIssuer:               creds.Key.Issuer,
+		KeyRedirectURI:          creds.Key.RedirectURI,
+		KeyScopesHash:           creds.Key.ScopesHash,
+		ProviderName:            creds.ProviderName,
+		ClientID:                creds.ClientID,
+		ClientSecret:            creds.ClientSecret,
+		TokenEndpointAuthMethod: creds.TokenEndpointAuthMethod,
+		RegistrationAccessToken: creds.RegistrationAccessToken,
+		RegistrationClientURI:   creds.RegistrationClientURI,
+		AuthorizationEndpoint:   creds.AuthorizationEndpoint,
+		TokenEndpoint:           creds.TokenEndpoint,
+	}
+	if !creds.CreatedAt.IsZero() {
+		stored.CreatedAt = creds.CreatedAt.Unix()
+	}
+	if !creds.ClientSecretExpiresAt.IsZero() {
+		stored.ClientSecretExpiresAt = creds.ClientSecretExpiresAt.Unix()
+	}
+
+	data, err := json.Marshal(stored) //nolint:gosec // G117 - internal Redis storage serialization, not exposed to users
+	if err != nil {
+		return fmt.Errorf("failed to marshal dcr credentials: %w", err)
+	}
+
+	// Derive Redis TTL from ClientSecretExpiresAt:
+	//   * Zero (unset)   -> TTL=0 (no expiration) per RFC 7591 §3.2.1 "never".
+	//   * Future expiry  -> TTL = time.Until(expiry).
+	//   * Past expiry    -> TTL = pastExpiryDCRTTL (bounded eviction window).
+	// See the function docstring for the past-expiry rationale.
+	ttl := time.Duration(0)
+	if !creds.ClientSecretExpiresAt.IsZero() {
+		if until := time.Until(creds.ClientSecretExpiresAt); until > 0 {
+			ttl = until
+		} else {
+			ttl = pastExpiryDCRTTL
+		}
+	}
+
+	if err := s.client.Set(ctx, key, data, ttl).Err(); err != nil {
+		return fmt.Errorf("failed to store dcr credentials: %w", err)
+	}
+	return nil
+}
+
+// GetDCRCredentials retrieves the credentials previously persisted under key.
+// Returns ErrNotFound (wrapped) when no entry exists. The returned value is a
+// fresh struct decoded from JSON, which acts as a defensive copy.
+//
+// An unpopulated key (empty Issuer or empty RedirectURI) cannot match any
+// stored row because StoreDCRCredentials rejects such keys, so a Get against
+// one is a normal miss — ErrNotFound — matching MemoryStorage.GetDCRCredentials
+// and the DCRCredentialStore interface contract.
+func (s *RedisStorage) GetDCRCredentials(ctx context.Context, key DCRKey) (*DCRCredentials, error) {
+	redisKey := redisDCRKey(s.keyPrefix, key)
+	data, err := s.client.Get(ctx, redisKey).Bytes()
+	if err != nil {
+		if errors.Is(err, redis.Nil) {
+			return nil, fmt.Errorf("%w: %w", ErrNotFound, fosite.ErrNotFound.WithHint("DCR credentials not found"))
+		}
+		return nil, fmt.Errorf("failed to get dcr credentials: %w", err)
+	}
+
+	var stored storedDCRCredentials
+	if err := json.Unmarshal(data, &stored); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal dcr credentials: %w", err)
+	}
+
+	return stored.toDCRCredentials(), nil
+}
+
+// -----------------------
 // Pending Authorization Storage
 // -----------------------
 
@@ -1862,4 +2043,5 @@ var (
 	_ ClientRegistry              = (*RedisStorage)(nil)
 	_ UpstreamTokenStorage        = (*RedisStorage)(nil)
 	_ UserStorage                 = (*RedisStorage)(nil)
+	_ DCRCredentialStore          = (*RedisStorage)(nil)
 )
--- a/pkg/authserver/storage/redis_integration_test.go
+++ b/pkg/authserver/storage/redis_integration_test.go
@@ -1461,8 +1461,6 @@ func TestIntegration_DCRCredentials_RoundTrip(t *testing.T) {
 }
 
 func TestIntegration_DCRCredentials_DistinctKeysCoexist(t *testing.T) {
-	t.Parallel()
-
 	withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
 		mkKey := func(issuer, redirect string, scopes []string) DCRKey {
 			return DCRKey{Issuer: issuer, RedirectURI: redirect, ScopesHash: ScopesHash(scopes)}
@@ -1495,8 +1493,6 @@ func TestIntegration_DCRCredentials_DistinctKeysCoexist(t *testing.T) {
 }
 
 func TestIntegration_DCRCredentials_OverwriteSemantics(t *testing.T) {
-	t.Parallel()
-
 	withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
 		key := dcrFixtureKey()
 		mk := func(clientID string) *DCRCredentials {
@@ -1578,14 +1574,12 @@ func TestIntegration_DCRCredentials_ConcurrentAccess(t *testing.T) {
 	t.Parallel()
 
 	t.Run("overlapping_key", func(t *testing.T) {
-		t.Parallel()
 		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
 			runDCRConcurrentAccess(ctx, t, s, dcrConcurrentOverlappingKey, 30*time.Second)
 		})
 	})
 
 	t.Run("disjoint_keys", func(t *testing.T) {
-		t.Parallel()
 		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
 			runDCRConcurrentAccess(ctx, t, s, dcrConcurrentDisjointKeys, 30*time.Second)
 		})
--- a/pkg/authserver/storage/redis_integration_test.go
+++ b/pkg/authserver/storage/redis_integration_test.go
@@ -1467,11 +1467,19 @@ func TestIntegration_DCRCredentials_DistinctKeysCoexist(t *testing.T) {
 		mkKey := func(issuer, redirect string, scopes []string) DCRKey {
 			return DCRKey{Issuer: issuer, RedirectURI: redirect, ScopesHash: ScopesHash(scopes)}
 		}
+		mk := func(key DCRKey, clientID string) *DCRCredentials {
+			return &DCRCredentials{
+				Key:                   key,
+				ClientID:              clientID,
+				AuthorizationEndpoint: "https://idp.example.com/auth",
+				TokenEndpoint:         "https://idp.example.com/token",
+			}
+		}
 		entries := []*DCRCredentials{
-			{Key: mkKey("https://idp-a.example.com", "https://x/cb", []string{"openid"}), ClientID: "a"},
-			{Key: mkKey("https://idp-b.example.com", "https://x/cb", []string{"openid"}), ClientID: "b"},
-			{Key: mkKey("https://idp-a.example.com", "https://y/cb", []string{"openid"}), ClientID: "c"},
-			{Key: mkKey("https://idp-a.example.com", "https://x/cb", []string{"openid", "email"}), ClientID: "d"},
+			mk(mkKey("https://idp-a.example.com", "https://x/cb", []string{"openid"}), "a"),
+			mk(mkKey("https://idp-b.example.com", "https://x/cb", []string{"openid"}), "b"),
+			mk(mkKey("https://idp-a.example.com", "https://y/cb", []string{"openid"}), "c"),
+			mk(mkKey("https://idp-a.example.com", "https://x/cb", []string{"openid", "email"}), "d"),
 		}
 		for _, e := range entries {
 			require.NoError(t, s.StoreDCRCredentials(ctx, e))
@@ -1491,9 +1499,17 @@ func TestIntegration_DCRCredentials_OverwriteSemantics(t *testing.T) {
 
 	withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
 		key := dcrFixtureKey()
+		mk := func(clientID string) *DCRCredentials {
+			return &DCRCredentials{
+				Key:                   key,
+				ClientID:              clientID,
+				AuthorizationEndpoint: "https://idp.example.com/auth",
+				TokenEndpoint:         "https://idp.example.com/token",
+			}
+		}
 
-		require.NoError(t, s.StoreDCRCredentials(ctx, &DCRCredentials{Key: key, ClientID: "first"}))
-		require.NoError(t, s.StoreDCRCredentials(ctx, &DCRCredentials{Key: key, ClientID: "second"}))
+		require.NoError(t, s.StoreDCRCredentials(ctx, mk("first")))
+		require.NoError(t, s.StoreDCRCredentials(ctx, mk("second")))
 
 		got, err := s.GetDCRCredentials(ctx, key)
 		require.NoError(t, err)
@@ -1517,6 +1533,8 @@ func TestIntegration_DCRCredentials_TTL(t *testing.T) {
 			require.NoError(t, s.StoreDCRCredentials(ctx, &DCRCredentials{
 				Key:                   key,
 				ClientID:              "client-with-expiry",
+				AuthorizationEndpoint: "https://idp.example.com/auth",
+				TokenEndpoint:         "https://idp.example.com/token",
 				ClientSecretExpiresAt: expires,
 			}))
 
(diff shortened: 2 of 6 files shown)


pkg/authserver/storage/redis.go
```diff
@@ -1322,6 +1333,176 @@ func unmarshalUpstreamTokens(data []byte) (*UpstreamTokens, error) {
...
+	if creds == nil {
+		return fosite.ErrInvalidRequest.WithHint("dcr credentials cannot be nil")
+	}
+	if creds.Key.Issuer == "" {
+		return fosite.ErrInvalidRequest.WithHint("dcr credentials key issuer cannot be empty")
+	}
+	if creds.Key.RedirectURI == "" {
+		return fosite.ErrInvalidRequest.WithHint("dcr credentials key redirect_uri cannot be empty")
+	}
+
```
tgrunnagle, line 1456: **[HIGH]** Validation contract divergence with MemoryStorage.

`MemoryStorage.StoreDCRCredentials` (parent PR's `memory.go`) rejects **seven** invalid inputs: nil, empty `Issuer`, empty `RedirectURI`, empty `ScopesHash`, empty `ClientID`, empty `AuthorizationEndpoint`, empty `TokenEndpoint`.

This function rejects only the first three. The `DCRCredentialStore` interface docstring in `types.go` explicitly defers to MemoryStorage's rejection list as canonical, so this is an interface-contract violation — the two backends are not observationally equivalent for the same inputs. A caller bug producing an empty `ClientID` will fail loud against Memory and silently persist a working-looking-but-broken row in Redis.

Flagged independently by 4 of 5 specialist agents (backend-equivalence, security, test-quality, general-go).

Suggested approach: extract the shared validation into a free function and call it from both backends so the rejection set can't drift again.

```go
// validateDCRCredentialsForStore checks the contract enforced by the
// DCRCredentialStore interface. Called by both MemoryStorage and
// RedisStorage so the rejection set stays in sync.
func validateDCRCredentialsForStore(creds *DCRCredentials) error {
    if creds == nil {
        return fosite.ErrInvalidRequest.WithHint("dcr credentials cannot be nil")
    }
    if creds.Key.Issuer == "" {
        return fosite.ErrInvalidRequest.WithHint("dcr credentials key issuer cannot be empty")
    }
    if creds.Key.RedirectURI == "" {
        return fosite.ErrInvalidRequest.WithHint("dcr credentials key redirect_uri cannot be empty")
    }
    if creds.Key.ScopesHash == "" {
        return fosite.ErrInvalidRequest.WithHint("dcr credentials key scopes_hash cannot be empty")
    }
    if creds.ClientID == "" {
        return fosite.ErrInvalidRequest.WithHint("dcr credentials client_id cannot be empty")
    }
    if creds.AuthorizationEndpoint == "" {
        return fosite.ErrInvalidRequest.WithHint("dcr credentials authorization_endpoint cannot be empty")
    }
    if creds.TokenEndpoint == "" {
        return fosite.ErrInvalidRequest.WithHint("dcr credentials token_endpoint cannot be empty")
    }
    return nil
}
```

After adding the helper, also extend `TestRedisStorage_DCRCredentials_StoreInvalidInputRejected` to cover all 7 cases (matching `TestMemoryStorage_DCRCredentials_StoreInvalidInputRejected`).

pkg/authserver/storage/redis.go
```diff
@@ -1322,6 +1333,176 @@ func unmarshalUpstreamTokens(data []byte) (*UpstreamTokens, error) {
...
+//
+// If ClientSecretExpiresAt is already in the past at call time, the entry is
+// written with the bounded TTL pastExpiryDCRTTL (1 second) rather than rejected
+// or stored long-lived. This keeps the store's "already-expired secret" window
+// narrow even if the resolver never re-reads the row, and matches the
+// fail-loud-but-tolerant posture: the caller's expiry timestamp round-trips so
+// a downstream reader can still observe it and trigger re-registration.
+//
+// Returns fosite.ErrInvalidRequest for nil creds, empty Issuer, or empty
+// RedirectURI — the same fail-loud contract as MemoryStorage.StoreDCRCredentials.
```
tgrunnagle: **[HIGH]** This docstring claims "the same fail-loud contract as `MemoryStorage.StoreDCRCredentials`". As written, that's false — see the inline at the function body for the four missing validation checks.

Once the implementation is brought into parity, this comment becomes accurate again. Otherwise it must be rewritten to enumerate the actual narrower set and explicitly call out the asymmetry, and the interface docstring in `types.go` would also need updating to permit the divergence.

pkg/authserver/storage/redis_test.go
```diff
@@ -2140,3 +2229,287 @@ func TestRedisStorage_GetLatestUpstreamTokensForUser(t *testing.T) {
...
+
+// TestRedisStorage_DCRCredentials_TTL pins the RFC 7591 §3.2.1
+// client_secret_expires_at semantics:
+//   - When ClientSecretExpiresAt is non-zero, the Redis row carries a TTL
+//     so it evicts before the upstream rejects the secret.
+//   - When ClientSecretExpiresAt is zero ("never"), the row is persistent
+//     (Redis TTL of -1).
+//   - When ClientSecretExpiresAt is in the past at write time, the row is
+//     written without a TTL (resolver re-checks expiry on read; see
+//     StoreDCRCredentials docstring).
```
tgrunnagle: **[MEDIUM]** Stale docstring — contradicts both the implementation and the subtest below.

The production code at `redis.go` line 1461 sets `ttl = pastExpiryDCRTTL` for past-expiry writes. The `past expiry uses bounded TTL` subtest at line 2452 asserts `assert.Equal(t, pastExpiryDCRTTL, mr.TTL(...), "past-expiry write must use the bounded pastExpiryDCRTTL, not TTL=0")`. This bullet says the opposite.

Acknowledged in the PR description as a known stale docstring.

```suggestion
//   - When ClientSecretExpiresAt is in the past at write time, the row
//     is written with the bounded `pastExpiryDCRTTL` (1 second) so an
//     already-expired secret self-evicts almost immediately rather than
//     persisting forever (see StoreDCRCredentials docstring).
```

pkg/authserver/storage/redis_test.go
```diff
@@ -2140,3 +2229,287 @@ func TestRedisStorage_GetLatestUpstreamTokensForUser(t *testing.T) {
...
+
+		var wg sync.WaitGroup
+		wg.Add(goroutines)
+		for g := 0; g < goroutines; g++ {
+			gid := g
+			go func() {
+				defer wg.Done()
+				for i := 0; i < iterations; i++ {
+					key := DCRKey{
+						Issuer:      fmt.Sprintf("https://idp-%d.example.com", gid),
```
tgrunnagle: **[MEDIUM]** This test names itself `_ConcurrentAccess`, but the key is built from `gid` — every goroutine writes/reads its own private key. Concurrent contention on the *same* key is never exercised.

The Memory baseline (`TestMemoryStorage_DCRCredentials_ConcurrentAccess`) deliberately runs both an `overlappingKey()` variant (shared keyspace, exercises last-write-wins) and a `disjointKey()` variant (per-goroutine, race-detector smoke test) and documents the rationale.

Suggested fix: either rename to `_DisjointKeysConcurrent` to match what's actually pinned, or add a same-key variant where every goroutine writes/reads `dcrFixtureKey()` and asserts every Get returns one of the values written by some goroutine.

Also flagged on the integration test in `redis_integration_test.go` line 1569.

pkg/authserver/storage/redis_integration_test.go
```diff
@@ -1413,3 +1413,179 @@ func TestIntegration_MigrateLegacyUpstreamData(t *testing.T) {
...
+
+		var wg sync.WaitGroup
+		wg.Add(goroutines)
+		for g := 0; g < goroutines; g++ {
+			gid := g
+			go func() {
+				defer wg.Done()
+				for i := 0; i < iterations; i++ {
+					key := DCRKey{
+						Issuer:      fmt.Sprintf("https://idp-%d.example.com", gid),
```
tgrunnagle: **[MEDIUM]** Same disjoint-keys-per-goroutine issue as the unit test (`redis_test.go` line 2493). With `Issuer: fmt.Sprintf("https://idp-%d.example.com", gid)` every worker writes its own key — concurrent same-key contention is never tested. Consider mirroring the Memory baseline's overlapping/disjoint pair.

pkg/authserver/storage/redis_test.go
```diff
@@ -2140,3 +2229,287 @@ func TestRedisStorage_GetLatestUpstreamTokensForUser(t *testing.T) {
...
+						RedirectURI: "https://x/cb",
+						ScopesHash:  ScopesHash([]string{"openid"}),
+					}
+					require.NoError(t, s.StoreDCRCredentials(ctx, &DCRCredentials{
```
tgrunnagle: **[MEDIUM]** `require.NoError` in a spawned goroutine.

Per the Go [testing docs](https://pkg.go.dev/testing#T.FailNow): "FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines." `require.NoError` invokes `FailNow` on failure.

The Memory baseline (`TestMemoryStorage_DCRCredentials_ConcurrentAccess`) avoids this with an `atomic.AddInt32(&errCount, 1)` pattern checked from the test goroutine after `wg.Wait()`. Apply the same pattern here and at line 2502 (the Get call), and to the integration test at lines 1573 / 1578.

pkg/authserver/storage/redis_integration_test.go
```diff
@@ -1413,3 +1413,179 @@ func TestIntegration_MigrateLegacyUpstreamData(t *testing.T) {
...
+						RedirectURI: "https://x/cb",
+						ScopesHash:  ScopesHash([]string{"openid"}),
+					}
+					require.NoError(t, s.StoreDCRCredentials(ctx, &DCRCredentials{
```
tgrunnagle: **[MEDIUM]** Same `require.NoError`-in-goroutine issue as the unit test at `redis_test.go` line 2497. See that comment for the recommended atomic-counter pattern.

pkg/authserver/storage/redis_keys.go
```diff
@@ -88,6 +94,24 @@ func redisProviderKey(prefix, providerID, providerSubject string) string {
...
+// (Issuer, RedirectURI, ScopesHash) tuple that DCRKey canonicalises.
+//
+// Format: "{prefix}dcr:<len(issuer)>:<issuer>:<len(redirect_uri)>:<redirect_uri>:<scopes_hash>"
+//
+// The first two segments are length-prefixed to handle colons in RedirectURI
+// (and, for symmetry, Issuer) without ambiguity, mirroring redisProviderKey.
+// ScopesHash is a SHA-256 hex digest produced by storage.ScopesHash; it
+// contains only [0-9a-f] and never contains a colon, so it can be appended
+// without a length prefix.
+func redisDCRKey(prefix string, key DCRKey) string {
```
tgrunnagle, line 111: **[LOW]** The docstring at lines 104-106 asserts that `ScopesHash` contains only `[0-9a-f]` and never contains a colon, justifying its appending without a length prefix. The assertion is never enforced — neither here nor in `RedisStorage.StoreDCRCredentials` (which doesn't currently validate empty `ScopesHash`, see the HIGH finding on `redis.go`).

Length-prefix collision-safety is preserved (ScopesHash is the trailing segment), so the broader correctness story holds, but a caller bypassing `storage.ScopesHash()` could pass arbitrary strings — empty, colon-bearing, whitespace.

Mostly subsumed by the validation fix on `StoreDCRCredentials`. Optional follow-up: after the validation tightens, also reject non-hex content (cheap `len(creds.Key.ScopesHash)==64 && hex.DecodeString(...) succeeds`) so the assumption documented here is fully enforced.

pkg/authserver/storage/redis.go
```diff
@@ -1322,6 +1333,176 @@ func unmarshalUpstreamTokens(data []byte) (*UpstreamTokens, error) {
...
+	if err := s.client.Set(ctx, key, data, ttl).Err(); err != nil {
+		return fmt.Errorf("failed to store dcr credentials: %w", err)
+	}
+	return nil
+}
+
+// GetDCRCredentials retrieves the credentials previously persisted under key.
+// Returns ErrNotFound (wrapped) when no entry exists. The returned value is a
+// fresh struct decoded from JSON, which acts as a defensive copy.
+//
```
tgrunnagle, line 1507: **[LOW]** This docstring's reasoning omits `ScopesHash`: it says an unpopulated key (empty Issuer or RedirectURI) cannot match any stored row "because StoreDCRCredentials rejects such keys". With the current Redis Store accepting empty `ScopesHash`, the symmetry is incomplete.

After tightening Store-side validation (HIGH finding on the function above), update this Get docstring to enumerate all rejected components: `"empty Issuer, RedirectURI, or ScopesHash"`. Keeps Get-side and Store-side reasoning observably consistent.

Also consider adding an `empty scopes_hash` subtest to `TestRedisStorage_DCRCredentials_NotFound` to pin the symmetry.

tgrunnagle (commented):
## Multi-agent review summary

Reviewed by 5 specialist agents (backend equivalence, concurrency/TTL, security, test quality, general Go quality). Findings below survived consensus scoring (threshold ≥ 7/10).

**Review event**: COMMENT (PR is by the posting user and marked DRAFT — GitHub disallows self-APPROVE/REQUEST_CHANGES anyway).

### Consensus summary

| # | Finding | Severity | Anchor | Agent agreement |
|---|---------|----------|--------|-----------------|
| 1 | RedisStorage.StoreDCRCredentials validation set diverges from MemoryStorage; docstring's "same fail-loud contract" claim is false. Empty `ScopesHash` / `ClientID` / `AuthorizationEndpoint` / `TokenEndpoint` are rejected by Memory, accepted by Redis. | HIGH | `redis.go` 1421-1432 | 4 / 5 |
| 2 | TTL test docstring ("row is written without a TTL") contradicts the `past expiry uses bounded TTL` subtest body and the production code's `pastExpiryDCRTTL` branch. Acknowledged in PR description. | MEDIUM | `redis_test.go` 2413-2415 | 2 / 5 |
| 3 | `*_ConcurrentAccess` tests partition the keyspace by goroutine ID — every worker writes its own key, so concurrent same-key contention is not exercised. The Memory baseline has both a disjoint and an overlapping variant. | MEDIUM | `redis_test.go` 2493, `redis_integration_test.go` 1569 | 2 / 5 |
| 4 | `require.NoError(t, ...)` is called from spawned goroutines in both `*_ConcurrentAccess` tests. Per Go testing docs, `FailNow` (which `require.NoError` invokes on failure) must be called only from the goroutine running the test function — behaviour from another goroutine is undefined. The Memory baseline uses an `atomic.AddInt32(&errCount, 1)` pattern instead. | MEDIUM | `redis_test.go` 2497, `redis_integration_test.go` 1573 | 1 / 5 |
| 5 | `redisDCRKey` documents that `ScopesHash` is `[0-9a-f]` and never contains a colon, then appends it without a length prefix. The assumption isn't enforced anywhere — once finding #1 is fixed (validate empty `ScopesHash` on Store) most of the gap closes; full enforcement would also reject non-hex content or length-prefix the segment. | LOW | `redis_keys.go` 104-107 | 1 / 5 |
| 6 | `GetDCRCredentials` docstring claims that an unpopulated key (empty Issuer / RedirectURI) cannot match any stored row "because StoreDCRCredentials rejects such keys". The reasoning is incomplete: it omits `ScopesHash`, and once #1 is fixed should be updated to enumerate all rejected components for symmetry. | LOW | `redis.go` 1483-1486 | 1 / 5 |

### Holistic assessment

**Backend interchangeability is the load-bearing contract here.** The `DCRCredentialStore` interface (parent PR) explicitly defers to MemoryStorage's rejected-input list as canonical. The Redis backend's narrower validation breaks substitutability — a record with an empty `ClientID` or `ScopesHash` will fail loud against Memory and silently persist in Redis. The cleanest fix is extracting a free function `validateDCRCredentialsForStore(*DCRCredentials) error` and calling it from both backends so the validation set cannot drift again.

The TTL docstring drift (#2) and the `*_ConcurrentAccess` naming-vs-behaviour gap (#3) don't affect correctness today but degrade the test set's value as a regression guard. Both are cheap fixes.

The `require.NoError`-in-goroutine issue (#4) is a Go testing.T contract violation — not a current bug, but undefined-behaviour territory on test failure. The Memory baseline already does the right thing.

Forward-looking: when sub-issue 3 wires the embedded auth server to use this backend, the bounded-TTL past-expiry behaviour will be exercised under real workloads. A `slog.Warn` on that branch (currently silent) would give operators a signal they can correlate with upstream clock skew. Captured in the dropped-findings appendix as a follow-up.

### Architecture notes (no action required)

- Pattern parity with `storedUpstreamTokens` (Unix-epoch fields, length-prefixed keys, JSON wire format) is excellent.
- No migration concern — `KeyTypeDCR` is a new key type with no overlap.
- Security posture is sound: no secrets in errors, no logging of credentials, multi-tenant prefix correctly applied to every new key.
- Codex cross-review: skipped (codex CLI not installed locally).

### Dropped findings (consensus < 7/10)

- `,omitempty` on validated key fields divergence from `storedUpstreamTokens`
- Past-expiry write should emit a `slog.Warn`
- "Second-grain TTL is the broadly-tested path" rationale is technically misleading for the future-expiry branch (which uses PX)
- `GetDCRCredentials` does not validate the unmarshaled stored struct (defensive-in-depth only)
- `NotFound` test "fully empty key" subtest is redundant
- Defensive-copy test is shape-fragile against future struct evolution
- Past-expiry exposure window (1s) deserves explicit rationale-vs-reject doc
- 10 s unit-test concurrent timeout could be tight under `-race` on slow CI
- `pastExpiryDCRTTL` exact-equality assertion fragile against real-Redis clock drift (miniredis pin only)
- `dcrFixtureKey` shared between unit/integration via build-tag visibility

🤖 Generated by `/dev:pr-review`

pkg/authserver/storage/redis.go
tgrunnagle: Addressed in df0ad8fa.

pkg/authserver/storage/redis.go
tgrunnagle: Addressed in df0ad8fa.

pkg/authserver/storage/redis_test.go
tgrunnagle: Addressed in df0ad8fa.

pkg/authserver/storage/redis_test.go
tgrunnagle: Addressed in df0ad8fa.

pkg/authserver/storage/redis_integration_test.go
tgrunnagle: Addressed in df0ad8fa.

pkg/authserver/storage/redis_test.go
tgrunnagle: Addressed in df0ad8fa.

pkg/authserver/storage/redis_integration_test.go
tgrunnagle: Addressed in df0ad8fa.

pkg/authserver/storage/redis_keys.go
tgrunnagle: Addressed in df0ad8fa.

pkg/authserver/storage/redis.go
tgrunnagle: Addressed in df0ad8fa.

pkg/authserver/storage/redis_integration_test.go
```diff
@@ -1413,3 +1413,181 @@ func TestIntegration_MigrateLegacyUpstreamData(t *testing.T) {
...
+	t.Run("get non-existent", func(t *testing.T) {
+		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
+			_, err := s.GetDCRCredentials(ctx, dcrFixtureKey())
+			requireRedisNotFoundError(t, err)
+		})
+	})
+}
+
+func TestIntegration_DCRCredentials_DistinctKeysCoexist(t *testing.T) {
+	t.Parallel()
```
jhrozek: **HIGH — runtime panic on integration suite.**

This `t.Parallel()` followed by `withIntegrationStorage(t, ...)` panics: the helper at line 343 already calls `t.Parallel()` on the same `*testing.T`. Reproduced:

```
--- FAIL: TestIntegration_DCRCredentials_DistinctKeysCoexist (0.00s)
panic: testing: t.Parallel called multiple times [recovered, repanicked]
```

Same bug at lines 1498, 1581, and 1588 (separate inline comments). Fix: drop this line — the helper handles parallelism.

```suggestion
```

More context (and a note on why CI didn't catch this) in the review summary.

pkg/authserver/storage/redis_integration_test.go
```diff
@@ -1413,3 +1413,181 @@ func TestIntegration_MigrateLegacyUpstreamData(t *testing.T) {
...
+			got, err := s.GetDCRCredentials(ctx, want.Key)
+			require.NoError(t, err)
+			require.NotNil(t, got)
+			assert.Equal(t, want.ClientID, got.ClientID)
+		}
+	})
+}
+
+func TestIntegration_DCRCredentials_OverwriteSemantics(t *testing.T) {
+	t.Parallel()
```
jhrozek: Same `t.Parallel()`-then-`withIntegrationStorage(t, ...)` panic as line 1464. Drop this line.

```suggestion
```

pkg/authserver/storage/redis_integration_test.go
```diff
@@ -1413,3 +1413,181 @@ func TestIntegration_MigrateLegacyUpstreamData(t *testing.T) {
...
+// real Redis. Mirrors the unit-test sibling (overlapping + disjoint
+// keyspaces) using the shared runDCRConcurrentAccess helper from
+// redis_test.go (visible across build tags), with a longer timeout to
+// absorb real-Redis network latency. Run with -race to validate the
+// data-race detector is clean.
+func TestIntegration_DCRCredentials_ConcurrentAccess(t *testing.T) {
+	t.Parallel()
+
+	t.Run("overlapping_key", func(t *testing.T) {
+		t.Parallel()
```
jhrozek: Same `t.Parallel()`-then-`withIntegrationStorage(t, ...)` panic as line 1464. Drop this line.

```suggestion
```

pkg/authserver/storage/redis_integration_test.go
```diff
@@ -1413,3 +1413,181 @@ func TestIntegration_MigrateLegacyUpstreamData(t *testing.T) {
...
+		withIntegrationStorage(t, func(ctx context.Context, s *RedisStorage) {
+			runDCRConcurrentAccess(ctx, t, s, dcrConcurrentOverlappingKey, 30*time.Second)
+		})
+	})
+
+	t.Run("disjoint_keys", func(t *testing.T) {
+		t.Parallel()
```
jhrozek: Same `t.Parallel()`-then-`withIntegrationStorage(t, ...)` panic as line 1464. Drop this line.

```suggestion
```

jhrozek (commented):
Two findings from a fresh review pass.

## Finding 1 (HIGH) — runtime panic on integration suite

Four DCR integration test functions panic at runtime because they call `t.Parallel()` and then `withIntegrationStorage(t, ...)` on the same `*testing.T`; the helper itself calls `t.Parallel()` (line 343), so the second invocation panics with `testing: t.Parallel called multiple times`. Reproduced locally with:

```
go test -tags=integration -run TestIntegration_DCRCredentials_DistinctKeysCoexist ./pkg/authserver/storage/
--- FAIL: TestIntegration_DCRCredentials_DistinctKeysCoexist (0.00s)
panic: testing: t.Parallel called multiple times [recovered, repanicked]
```

The other DCR integration tests (`_RoundTrip`, `_TTL`) get this right: they call `t.Parallel()` only on the outer test and let `withIntegrationStorage` handle parallelism inside `t.Run` subtests where `t` is a different instance. Inline suggestions below.

**FYI on the CI side**: this passed CI green because the Redis Sentinel suite is gated behind `//go:build integration`, and no GitHub Actions workflow builds with that tag — `task test-coverage` (which runs in `.github/workflows/test.yml`) excludes integration-tagged tests, and the existing `task test-integration` target isn't wired up. So the suite has zero CI coverage today; that's worth a separate fix but is orthogonal to this PR.

## Finding 2 (MEDIUM) — stale forward references in `types.go`

Three pre-existing docstrings on the `DCRCredentialStore` interface still describe the Redis backend as future work. After this PR merges they will contradict the implementation, which violates the project's `Keep Comments Synchronized With Code` rule. (These lines aren't in the PR's diff hunks, so I can't attach inline suggestions — flagging here.)

**`pkg/authserver/storage/types.go:230-231`** (`DCRCredentials` doc, `# Lifetime` section)
```diff
-// the periodic cleanup loop. The Redis backend (future sub-issue) applies
-// TTL via SetEX when ClientSecretExpiresAt is non-zero.
+// the periodic cleanup loop. The Redis backend applies TTL via SET with a
+// duration when ClientSecretExpiresAt is non-zero.
```

**`pkg/authserver/storage/types.go:266-267`** (`ClientSecretExpiresAt` field doc)
```diff
-	// the persisted entry: the Redis backend (sub-issue 2) plumbs it through
-	// SetEX so the row evicts before the upstream rejects the secret at the
+	// the persisted entry: the Redis backend plumbs it through SET with a
+	// duration so the row evicts before the upstream rejects the secret at the
```

**`pkg/authserver/storage/types.go:275-276`** (`DCRCredentialStore` interface doc)
```diff
-// dynamic-client-registration credentials. Both MemoryStorage and a future
-// Redis-backed store implement it; an authserver backed by Redis shares DCR
+// dynamic-client-registration credentials. Both MemoryStorage and
+// RedisStorage implement it; an authserver backed by Redis shares DCR
```

(Side note on `SetEX`: the implementation uses `Set` with a `time.Duration`, which go-redis emits as `SET ... EX/PX` — slightly inaccurate as written, but the bigger drift is the "future" framing.)

---

Other findings from the review (past-expiry backend divergence, Store-side defensive-copy parity test, `t.Fatalf` in goroutine, vacuous `nolint:gosec`, TTL-slack inconsistency) were considered and intentionally not surfaced — happy to discuss any of them if useful.

pkg/authserver/storage/redis_integration_test.go
tgrunnagle: Addressed in 670ea89d.

pkg/authserver/storage/redis_integration_test.go
tgrunnagle: Addressed in 670ea89d.

pkg/authserver/storage/redis_integration_test.go
tgrunnagle: Addressed in 670ea89d.

pkg/authserver/storage/redis_integration_test.go
tgrunnagle: Addressed in 670ea89d.

tgrunnagle:
Finding 2 (stale 'future Redis backend' framing in `types.go`) addressed in 670ea89d. Applied your three suggested diffs at lines 230-231, 266-267, and 275-276, plus the matching SetEX→SET-with-duration correction at line 295 (interface `# TTL handling` section) for consistency — same drift pattern.

On the orthogonal CI gap you flagged (no workflow builds with `-tags=integration`): noted as a follow-up, will track separately.

jhrozek (approved):

PR #5196, opened 2026-05-05
Wire persistent DCRCredentialStore into EmbeddedAuthServer

### LINKED ISSUE #5185: Persistent DCRCredentialStore: wire into EmbeddedAuthServer
## Description

Sub-task 3 of 3 for Phase 3 of the DCR story (#4976, parent #4979). Wire the persistent `DCRCredentialStore` (in-memory from sub-issue 1, Redis from sub-issue 2) into `pkg/authserver/runner/embeddedauthserver.go` in place of the Phase 2 in-memory stub, and consolidate the runner-side abstraction so there is exactly one `DCRCredentialStore` implementation per backend. This is the last missing piece for the Datadog-style upstream demo to survive authserver restarts and scale across replicas: restarting (or scaling out to) an authserver backed by Redis reuses already-registered DCR credentials instead of re-registering at the upstream.

## Context

Sub-issue 1 lands the storage-level `DCRCredentialStore` interface and the in-memory implementation. Sub-issue 2 lands the Redis implementation. This sub-issue swaps the runner's wiring so a single `storage_type: redis` config toggles DCR persistence alongside the rest of authserver state, and removes the Phase 2 standalone in-memory stub now that `MemoryStorage` itself implements the interface.

Wiring site: `createStorage(ctx, cfg.Storage)` at `pkg/authserver/runner/embeddedauthserver.go:455` selects memory vs. Redis for the main `Storage` backend; the `DCRCredentialStore` reuses that same branch via a type assertion so no second selection branch is needed.

This sub-issue introduces no new protocol behavior. It is a wiring change.

**Dependencies**: sub-issue 1 (types, interface, in-memory implementation), sub-issue 2 (Redis implementation)
**Blocks**: (none — terminal task in the DAG)

### Wiring — `pkg/authserver/runner/embeddedauthserver.go`

- [ ] The `EmbeddedAuthServer` struct field that holds the DCR store (introduced in #4978) is now typed against the storage-level `DCRCredentialStore` interface.
- [ ] `NewEmbeddedAuthServer` derives the DCR store from the same `Storage` value returned by `createStorage(ctx, cfg.Storage)` (see `embeddedauthserver.go:97` and `:455`). Since both `MemoryStorage` and `RedisStorage` now implement the storage-level `DCRCredentialStore` interface, a plain `stor.(DCRCredentialStore)` type assertion is sufficient. Do NOT add a second storage-selection branch.
- [ ] The Phase 2 standalone in-memory stub from `pkg/authserver/runner/dcr_store.go` is removed (or reduced to a re-export/alias) so there is exactly one `DCRCredentialStore` implementation per backend. Document this in the commit message.
- [ ] `buildPureOAuth2Config` remains pure (unchanged signature, no `ctx`, no I/O) — the architectural gate established in #4978 is preserved.

### Cross-cutting

The wiring site at `embeddedauthserver.go:455` (`createStorage`) already returns `storage.Storage`. Since both backends now implement `DCRCredentialStore`, a single type assertion is all the wiring change needs:

```go
dcrStore, ok := stor.(storage.DCRCredentialStore)
if !ok {
    return nil, fmt.Errorf("storage backend %T does not implement DCRCredentialStore", stor)
}
```

Pass `dcrStore` into the `EmbeddedAuthServer` struct field introduced by #4978; delete the Phase 2 `NewInMemoryDCRCredentialStore()` fallback construction (or demote it to the no-op default for unit tests only).

`buildPureOAuth2Config` MUST remain pure — no `ctx`, no I/O. The DCR store is consumed by the resolver added in #4978; the wiring change only swaps the implementation passed in, not the call shape.

Verify with `task build`, `task test`, `task lint-fix`, `task license-check`. Run the secret-grep assertion from #4978. Build the whole repo to confirm the runner and storage packages compile together.

### Patterns & Frameworks

- Plain `testing` + `github.com/stretchr/testify` per `.claude/rules/testing.md` — NOT Ginkgo. `require.NoError` over `t.Fatal`.
- Authserver restart integration — extend `pkg/authserver/integration_test.go` (do NOT create a new integration-test file) to cover the durable-restart case: boot an `EmbeddedAuthServer` with a mock AS, register, tear down, rebuild against the same `MemoryStorage` instance, and assert the second boot makes zero AS requests. This is the direct analog of the Phase 2 cache-hit integration assertion from #4978, just crossing an `EmbeddedAuthServer` lifecycle boundary.

### Code Pointers

- `pkg/authserver/runner/embeddedauthserver.go:97` — call site for `createStorage`; the storage value returned here is the single source of truth for the DCR backend after this PR.
- `pkg/authserver/runner/embeddedauthserver.go:455-467` — `createStorage` backend-selection switch; NO change required here. The DCR store is derived via type assertion on the returned `storage.Storage`.
- `pkg/authserver/runner/dcr_store.go` — Phase 2 standalone stub. Remove or demote to a thin alias.
- `pkg/authserver/integration_test.go` — extend with the authserver-restart case; do NOT create a new integration-test file.

### Component Interfaces

```go
// pkg/authserver/runner/embeddedauthserver.go (wiring change; sketch)

stor, err := createStorage(ctx, cfg.Storage)
if err != nil {
    return nil, fmt.Errorf("failed to create storage: %w", err)
}

dcrStore, ok := stor.(storage.DCRCredentialStore)
if !ok {
    return nil, fmt.Errorf("storage backend %T does not implement DCRCredentialStore", stor)
}

// existing authserver.New call, now passing dcrStore into the DCR
// resolver as well (field name per #4978, e.g. e.dcrStore = dcrStore).
```

## Testing Strategy

**Integration Tests**

- [ ] `pkg/authserver/integration_test.go` (extend, do NOT create a new file): boot an `EmbeddedAuthServer` with a mock AS serving metadata + `/register`; trigger DCR during `NewEmbeddedAuthServer`; close the server; rebuild a second `EmbeddedAuthServer` against the same `MemoryStorage` instance (the Redis variant is covered by the Redis integration test in sub-issue 2); assert the second boot makes ZERO HTTP requests to the mock AS — the persisted credential short-circuits the resolver. This is the direct analog of the #4978 cache-hit integration case, crossing an `EmbeddedAuthServer` lifecycle boundary.

**Edge Cases**

- [ ] Type assertion failure surfaces as a clear `NewEmbeddedAuthServer` error rather than a nil-pointer panic at first DCR resolve. (If sub-issue 1 chose to keep `DCRCredentialStore` as a method on every `Storage` impl, the assertion is unconditional and this case becomes a compile-time guarantee — preferred.)

## Out of Scope

- Types, in-memory backend, and `DCRCredentialStore` interface — sub-issue 1.
- Redis backend — sub-issue 2.

## References

- Parent issue: #4979
- Upstream dependency: ##4978 (Phase 2).
- RFC 7591 §3.2.1, RFC 7592.

tgrunnagle:
## Summary

- **Why:** Phase 2 of the DCR story shipped an in-memory stub for the `DCRCredentialStore` that lived only in the runner package. Restarting (or scaling out to) an authserver dropped every RFC 7591 client registration on the floor and re-registered against the upstream on every boot, which is unworkable for the Datadog-style upstream demo and for any multi-replica deployment. This PR wires the persistent `DCRCredentialStore` introduced in earlier sub-issues (in-memory + Redis backends) into `EmbeddedAuthServer` so a Redis-backed authserver reuses already-registered clients across replicas and restarts.
- **What:**
  - `EmbeddedAuthServer.dcrStore` is now typed against `storage.DCRCredentialStore` and is derived from the same `storage.Storage` value returned by `createStorage`, so a single `storage_type: redis` config toggles DCR persistence alongside the rest of authserver state. The `storage.Storage` interface embeds `storage.DCRCredentialStore`, promoting the previously-needed runtime type assertion to a compile-time guarantee.
  - The Phase 2 standalone in-memory `DCRCredentialStore` in `pkg/authserver/runner/dcr_store.go` is collapsed into a thin `storageBackedStore` adapter that delegates to `storage.DCRCredentialStore` and translates `DCRResolution` <-> `DCRCredentials` at the boundary. There is now exactly one persistence implementation per backend.
  - The constructor is split into the public `NewEmbeddedAuthServer` (creates storage) and an unexported `newEmbeddedAuthServerWithStorage` (owns the cleanup contract). Any error after entry closes the storage backend via a deferred cleanup gated on a named return error, so a crash-looping caller no longer leaks the Redis client connection pool / `MemoryStorage` cleanup goroutine on every restart.
  - `buildPureOAuth2Config` remains pure (unchanged signature, no `ctx`, no I/O); `buildUpstreamConfigs` is the boundary that consumes the resolver and overlays DCR-resolved credentials onto each upstream.
  - Adds `DCRStore()` accessor on `EmbeddedAuthServer` mirroring `IDPTokenStorage` / `UpstreamTokenRefresher`, used by integration tests to verify the resolver and the authserver write through the same backend.

This PR also lands the dependency stack that #5185 builds on (the persistent `DCRCredentialStore` types + memory backend, the Redis backend, the operator CRD surface for DCR, and the runner-side DCR resolver wiring). Each layer was developed and reviewed as a separate commit on this branch; commits are sequenced so each one builds and tests cleanly.

Closes #5185

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Notable test coverage added by this PR:

- `pkg/authserver/integration_dcr_restart_test.go` (new) — `TestEmbeddedAuthServer_DCRSurvivesRestart` boots an `EmbeddedAuthServer` against a mock AS, captures the DCR store via the new `DCRStore()` accessor, closes the server, and asserts the persisted DCR row survives the first server's `Close`. Lives in `package authserver_test` to avoid the runner -> authserver import cycle. The full "boot, close, boot again, observe zero `/register`" scenario across a fresh constructor is documented as a gap (the production Redis path requires Sentinel, which miniredis does not speak); test docstring records the conditions under which it can be closed.
- `pkg/authserver/runner/embeddedauthserver_test.go` — `TestBuildUpstreamConfigs_DCR` exercises first-call registration + cache-hit on the second call (zero additional HTTP requests) and asserts the caller's `RunConfig.Upstreams` slice is never mutated. `TestNewEmbeddedAuthServer_ClosesStorageOnError` uses a `closeTrackingStorage` wrapper to verify the deferred-cleanup contract.
- `pkg/authserver/storage/memory_test.go`, `redis_test.go`, `redis_integration_test.go` — coverage for the persistent `DCRCredentialStore` operations on both backends, including `ScopesHash` canonicalisation (sort + dedupe + newline join).

## API Compatibility

The CRD changes are additive: `OAuth2UpstreamConfig.clientId` becomes optional with a CEL constraint requiring exactly one of `clientId` or `dcrConfig`, and a new `dcrConfig` field is added. Existing `MCPExternalAuthConfig` / `VirtualMCPServer` resources that set `clientId` continue to validate unchanged.

Yes. Operators of OAuth2 upstreams can now configure RFC 7591 Dynamic Client Registration in the operator CRD via `dcrConfig` (with `discoveryUrl` or `registrationEndpoint`, plus optional `initialAccessTokenRef`, `softwareId`, `softwareStatement`) instead of statically configuring `clientId` + `clientSecret`. When the authserver is configured with `storage_type: redis`, DCR registrations persist across restarts and are shared across replicas; in single-replica memory mode, registrations live for the process lifetime as before.

- This PR is the terminal task in the Phase 3 DCR DAG and pulls along the dependency stack from sub-issues 1 and 2 (persistent `DCRCredentialStore` types + memory + Redis backends), the operator CRD surface, and the Phase 2 resolver wiring. The size is above the usual 400-line / 10-file limit; each commit is self-contained and the stack reads top-to-bottom in commit order. Reviewers may prefer to walk the per-commit diffs.
- The full "boot, close, boot again, zero `/register`" cross-constructor restart scenario is not exercised; closing it requires either miniredis-Sentinel emulation or a Docker-based Redis Sentinel cluster in the test harness. The wiring that the second boot would consume — the type of `dcrStore` being the same `storage.DCRCredentialStore` that `authserver.New` writes through — is verified at compile time by `storage.Storage` embedding `storage.DCRCredentialStore` and by `TestEmbeddedAuthServer_DCRSurvivesRestart` asserting the persistence boundary.
- `buildPureOAuth2Config` was kept intentionally pure (no `ctx`, no I/O) to preserve the architectural gate established in #4978; the wiring change swaps the implementation passed into the resolver, not the call shape.
- No secrets (`client_secret`, `registration_access_token`, `initial_access_token`, refresh tokens) appear as arguments to `slog.*` calls; the grep assertion from #4978 still applies.


pkg/authserver/runner/dcr_store.go
```diff
@@ -54,68 +56,45 @@ type dcrResolutionCache interface {
...
-//     local maps; whichever registration the upstream accepts last wins on
-//     that side, the loser becomes orphaned. The
-//     resolveDCRCredentials-level singleflight in dcr.go only deduplicates
-//     within one process.
+// deployments. It is a thin adapter over storage.NewMemoryStorage so the
+// runner-side cache and the authserver's main storage backend share a
+// single in-memory implementation; production deployments configure a
+// Redis-backed storage.DCRCredentialStore via storage.NewRedisStorage and
+// reach this same adapter through newStorageBackedStore.
 func newInMemoryDCRResolutionCache() dcrResolutionCache {
```
tgrunnagle: **F1 — Tests leak MemoryStorage cleanupLoop goroutine** (consensus 9/10, MEDIUM)

Pre-PR `newInMemoryDCRResolutionCache` returned a plain map+RWMutex. Post-PR it returns `newStorageBackedStore(storage.NewMemoryStorage())`, and `storage.NewMemoryStorage()` unconditionally launches `go s.cleanupLoop()` (memory.go:157). Every test that calls this constructor — and `dcr_test.go` calls it 25+ times, plus the 9 sites in `dcr_store_test.go` and 2 in `embeddedauthserver_test.go` — leaks a cleanupLoop goroutine for the run. None of the call sites register `t.Cleanup` to close the storage. Violates the `.claude/rules/go-style.md` resource-pairing rule and the testing rule about `t.Cleanup` in parallel tests.

**Recommendation**: Either change this constructor's signature so callers register cleanup (e.g., `func newTestDCRResolutionCache(t *testing.T) dcrResolutionCache` that registers `t.Cleanup`), or add a `Close()` method to the `dcrResolutionCache` interface and have callers register it. Consider running `goleak.VerifyNone(t)` in `TestMain` of the runner package to make future regressions self-detecting.

*Raised by: concurrency, test-coverage*

pkg/authserver/runner/dcr_store_test.go
```diff
@@ -92,10 +92,22 @@ func TestInMemoryDCRResolutionCache_DistinctKeysDoNotCollide(t *testing.T) {
 		ScopesHash:  storage.ScopesHash([]string{"openid", "email"}),
```
tgrunnagle, line 92: **F3 — Test names reference a deleted type** (consensus 7/10, MEDIUM)

This file's tests are named `TestInMemoryDCRResolutionCache_*`, but the struct `inMemoryDCRResolutionCache` was deleted in this PR — the tests now exercise `storageBackedStore` (reached via the test-only `newInMemoryDCRResolutionCache` wrapper). Future contributors greppin for the type behind the suite will find nothing. Violates `.claude/rules/testing.md` "test names must match what they actually assert."

**Recommendation**: Rename to `TestStorageBackedStore_*` (or `TestDCRResolutionCache_*`). Optionally also rename the constructor `newInMemoryDCRResolutionCache` → `newMemoryDCRStore` to reflect its actual semantics. Same applies to `dcr_test.go` if its tests share the naming.

*Raised by: concurrency, test-coverage*

pkg/authserver/runner/dcr_store.go
tgrunnagle: **F4 — `newInMemoryDCRResolutionCache` doc misleadingly markets it for production** (consensus 7/10, MEDIUM)

The docstring says this constructor is "intended for tests and single-replica development deployments," but production code in `NewEmbeddedAuthServer` no longer reaches it — production calls `newStorageBackedStore(stor)` directly with the shared authserver Storage. This function is now test-only (verified by `git grep newInMemoryDCRResolutionCache` — only `*_test.go` callers).

**Recommendation**: Tighten the doc to a single line: "test-only convenience constructor wrapping `storage.NewMemoryStorage` in the runner-side adapter; production deployments call `newStorageBackedStore` on the shared authserver Storage." Drop the cross-replica/durability paragraphs — they belong on the storage backend, not the runner adapter. Combined with F3, consider also renaming the function and the tests it powers.

*Raised by: concurrency, general-quality*

pkg/authserver/integration_dcr_restart_test.go
```diff
@@ -0,0 +1,217 @@
...
+// is verified at compile time by the storage.Storage interface
+// embedding storage.DCRCredentialStore and by the existing
+// TestNewEmbeddedAuthServer_DCRBoot in pkg/authserver/runner.
+func TestEmbeddedAuthServer_DCRSurvivesRestart(t *testing.T) {
+	t.Parallel()
+
+	server, requestCount := newMockUpstreamAS(t)
+
+	cfg := &authserver.RunConfig{
+		SchemaVersion: authserver.CurrentSchemaVersion,
```
tgrunnagle: **F5 — Restart test name overstates coverage; depends on undocumented MemoryStorage post-Close behavior** (consensus 9/10, MEDIUM)

The test name promises restart-survival, but no second `EmbeddedAuthServer` is ever constructed. The test boots once, captures the storage handle via `DCRStore()`, calls `Close()`, then reads from the same captured `MemoryStorage` instance. This passes only because `MemoryStorage.Close()` (memory.go:169) merely stops the cleanup goroutine and leaves the `dcrCredentials` map intact — an undocumented contract this test silently depends on. A future MemoryStorage that nils maps on Close (or the Redis equivalent of this test, which would error on a closed connection pool) would silently break this assertion. The PR's user-visible "DCR survives restart with Redis storage" promise remains untested in the suite — the 44-line gap-narrating docstring honestly acknowledges this but does not substitute for coverage.

**Recommendation** (three-part):
1. Rename to `TestEmbeddedAuthServer_DCRStorePersistsAcrossClose` (matches what is asserted).
2. Either (a) refactor the test to read credentials BEFORE `Close()`, dropping the post-Close dependency; OR (b) add an explicit comment to `storage.MemoryStorage.Close` codifying that read methods continue to function post-Close.
3. File a follow-up issue to land the real cross-boot reuse test once a Sentinel-capable miniredis or Docker-Sentinel harness lands, and link it from the test docstring.

*Raised by: concurrency, test-coverage, security*

pkg/authserver/storage/types.go
```diff
@@ -613,11 +613,21 @@ type UserStorage interface {
...
+	// pkg/authserver/runner.NewEmbeddedAuthServer relies on when wiring the
+	// authserver and its DCR resolver to a single shared backend; without it,
+	// a future Storage implementation that omitted the DCR methods would
+	// silently boot and fail at first DCR resolve. With it, the type system
+	// rejects any such implementation at compile time.
 	UpstreamTokenStorage
 	PendingAuthorizationStorage
 	ClientRegistry
 	UserStorage
+	DCRCredentialStore
```
tgrunnagle: **F6 — Storage interface widening couples impls to DCR; broadens secret reach** (consensus 7/10, MEDIUM)

Embedding `DCRCredentialStore` into the top-level `Storage` interface makes every future storage implementation provide DCR persistence even when not needed (e.g., a hypothetical SQLite backend that defers DCR to a sidecar). It also implicitly grants `GetDCRCredentials` access to every existing consumer of `storage.Storage` — `server.go:47`, `server_impl.go:131`, `handler.go:54+` — broadening the surface that can read raw `client_secret` and `registration_access_token`. Per `.claude/rules/security.md` ("route through security-enforcing components"), interface widening is the type-system analog of a privilege uplift: function signatures taking `storage.Storage` cannot tell the reader that DCR secrets are now reachable.

**Recommendation**: Keep `DCRCredentialStore` as a sibling interface, NOT embedded into `Storage`. Have `NewEmbeddedAuthServer` accept the DCR store as a separate explicit parameter, OR pull it off a narrower extension interface that only the constructor sees. The compile-time-guarantee motivation is already satisfied by the per-backend `var _ DCRCredentialStore = (*MemoryStorage)(nil)` (memory.go:1339) and `_ DCRCredentialStore = (*RedisStorage)(nil)` (redis.go:2040) assertions. If the embedding stays, add a `// SECURITY:` doc note on `Storage` stating that callers obtain raw DCR secrets through this seam.

*Raised by: architecture, security*

pkg/authserver/runner/dcr_store.go
```diff
@@ -124,16 +103,70 @@ func (s *inMemoryDCRResolutionCache) Get(_ context.Context, key DCRKey) (*DCRRes
...
+//   - ClientIDIssuedAt: informational only per RFC 7591 §3.2.1; the resolver
+//     does not consult it for cache invalidation, so it does not need to
+//     survive a process restart.
+//   - RedirectURI: already encoded into key.RedirectURI; storing it twice
+//     would risk drift between the canonical key and the persisted value.
+//
+// CreatedAt and ClientSecretExpiresAt are preserved so cache observers
+// (e.g. lookupCachedResolution's staleness Warn) and TTL-aware backends
+// (Redis) keep their existing behaviour after a restart.
+func resolutionToCredentials(key DCRKey, res *DCRResolution) *storage.DCRCredentials {
```
tgrunnagle, line 119: **F7 — `resolutionToCredentials` / `credentialsToResolution` drop fields with no round-trip test** (consensus 7/10, MEDIUM)

The new conversion seam between the runner's `DCRResolution` and the persisted `storage.DCRCredentials` deliberately drops `ClientIDIssuedAt` (zeroed on read), drops/restores `RedirectURI` via `Key`, and never populates `DCRCredentials.ProviderName` (which the storage docstring describes as "debug/audit only" but the converter leaves empty). No test asserts which fields round-trip and which do not. A future field added to `DCRResolution` would silently fail to persist unless someone remembers to update both converters — the exact "parallel types drift" failure mode `.claude/rules/go-style.md` warns about.

**Recommendation**:
1. Add a focused table-driven `TestResolutionCredentialsRoundTrip` at the conversion-function level that exercises every field of both types (preserved, dropped, key-recovered).
2. Add `// MUST update both converters when adding fields` comments on the `DCRResolution` and `DCRCredentials` struct definitions (currently only the converters carry this warning, where readers adding a struct field won't see it).
3. Decide on `ProviderName`: either thread it through `cache.Put` (DCRResolution would gain a ProviderName field), or remove it from `DCRCredentials` if no consumer actually reads it.

*Raised by: architecture, test-coverage*

pkg/authserver/runner/embeddedauthserver.go
```diff
@@ -74,6 +73,50 @@ func NewEmbeddedAuthServer(ctx context.Context, cfg *authserver.RunConfig) (*Emb
...
+// on every restart. The named return retErr is the gate.
+func newEmbeddedAuthServerWithStorage(
+	ctx context.Context,
+	cfg *authserver.RunConfig,
+	stor storage.Storage,
+) (retEAS *EmbeddedAuthServer, retErr error) {
+	// From here on, any error must close stor before returning.
+	defer func() {
+		if retErr != nil {
+			if closeErr := stor.Close(); closeErr != nil {
```
tgrunnagle, line 110: **F8 — Deferred-cleanup `slog.Warn` logs unredacted `retErr`; can leak upstream response body** (consensus 7/10, MEDIUM)

This warning logs `"error", closeErr, "original_error", retErr`. The most common `retErr` reaching this gate is a wrapped DCR failure whose error chain inlines several KiB of the upstream's raw `/register` response body (per `oauthproto/dcr.go`). Unlike the existing `logDCRStepError` boundary log, this path does not pass `retErr` through `sanitizeErrorForLog`, so it bypasses the URL-stripping defense in depth. A hostile or misconfigured upstream AS can echo back attacker-controlled markup — potentially including any RFC 7591 initial-access-token in the response, or strings designed to confuse log scrapers. The PR's grep-only "no secret literals as `slog.*` arguments" assertion does not catch secrets smuggled into `error.Error()` strings.

**Recommendation**: Either (a) wrap `retErr` with `sanitizeErrorForLog` before logging, OR (b) log only a structured summary (e.g., `"original_error_step"`, `"original_error_msg"` with a length-bounded short message) instead of the full chain. Add a unit test that constructs an error chain containing a fake `client_secret=` substring and asserts the secret does not appear in captured slog output. Optional naming nit: rename `"original_error"` to match the package-wide vocabulary, e.g. `"trigger_error"` or `"cause"`.

*Raised by: security (lead), concurrency (style note)*

pkg/authserver/integration_dcr_restart_test.go
```diff
@@ -0,0 +1,217 @@
...
+// either exporting the helper — bloating runner's public surface — or
+// sharing a test-helpers package). Keeping a small, self-contained
+// duplicate is the lower-cost option for this single use site.
+func newMockUpstreamAS(t *testing.T) (*httptest.Server, *int32) {
+	t.Helper()
+
+	var total int32
+	var server *httptest.Server
+
+	mux := http.NewServeMux()
```
tgrunnagle: **F9 — `newMockUpstreamAS` duplicates `pkg/authserver/runner.newMockAuthorizationServer`** (consensus 7/10, LOW)

The new helper is line-for-line identical to the runner-package helper (same metadata response, same `/register` handler returning the same hardcoded values, same catch-all counter). The PR docstring acknowledges this and argues "lower-cost option for this single use site" — which is a defensible one-off, but the duplication will silently drift between the two copies as the spec or assertions evolve.

**Recommendation**: Acceptable for this PR. File a follow-up to extract a shared helper in `pkg/authserver/internal/testhelpers` (or `internal/mockas`); when a third caller appears the rewrite will be small and bounded. As a tripwire, add a `// DO NOT COPY THIS A THIRD TIME — extract to testhelpers` comment on both copies.

*Raised by: architecture, test-coverage*

tgrunnagle (commented):

**Agents consulted**: concurrency, architecture, test-coverage, security, general-quality. Codex unavailable (skipped).

### Consensus Summary

| F1 | Tests leak MemoryStorage cleanupLoop goroutine via newInMemoryDCRResolutionCache | 9/10 | MEDIUM | Fix |
| F3 | Test names `TestInMemoryDCRResolutionCache_*` reference deleted type | 7/10 | MEDIUM | Fix |
| F4 | newInMemoryDCRResolutionCache doc still markets it for production deployments | 7/10 | MEDIUM | Fix |
| F5 | Restart test name overstates coverage; depends on undocumented MemoryStorage post-Close behavior | 9/10 | MEDIUM | Fix |
| F6 | Storage interface widening (DCRCredentialStore embed) couples impls to DCR; broadens secret reach | 7/10 | MEDIUM | Discuss |
| F7 | resolutionToCredentials/credentialsToResolution drop fields with no round-trip test | 7/10 | MEDIUM | Fix |
| F8 | Deferred-cleanup `slog.Warn` logs unredacted retErr; can leak upstream response body | 7/10 | MEDIUM | Fix |
| F9 | newMockUpstreamAS duplicates pkg/authserver/runner.newMockAuthorizationServer | 7/10 | LOW | Fix |
| F10 | DCRStore() accessor: secret reach + delegation pattern + lifecycle docs | 8/10 | MEDIUM | Discuss |
| F11 | PR is DRAFT; size over budget; e2e gap not documented | n/a | INFO | Comment |

### Overall

The wiring change is structurally correct: deferred-cleanup contract via named return `retErr` works, no double-close path between the deferred close and `EmbeddedAuthServer.Close → server.Close → storage.Close`, and `closeOnce` continues to make the caller path idempotent. The architectural gate from #4978 (`buildPureOAuth2Config` remains pure) is preserved. The `closeTrackingStorage` test seam is the right shape for verifying the contract — interface seam, not a hook field.

The dominant cluster of findings is comment drift and test-naming drift around the `dcrResolutionCache` / `inMemoryDCRResolutionCache` rename (F3, F4) plus a real test-infra goroutine leak that this PR introduces by routing test fixtures through `storage.NewMemoryStorage` (F1). These are fast to fix.

F6 and F10 are the design-discussion items. Embedding `DCRCredentialStore` into `Storage` was a deliberate choice to land a compile-time guarantee, but it widens the secret-bearing surface across every consumer of `storage.Storage` and every future backend. The new public `DCRStore()` accessor returns the raw secret-bearing interface and breaks the `e.server.*` delegation pattern used by sibling getters; if its only consumer is the integration test, a test-only export would be cleaner. Both are worth a deliberate decision rather than a side effect of compile-time-safety convenience.

F5 is the integration test's coverage story: the `TestEmbeddedAuthServer_DCRSurvivesRestart` name promises a restart, but the assertion is "reading from the same MemoryStorage instance after Close still works," which is an undocumented contract of MemoryStorage. The test docstring honestly acknowledges the cross-boot scenario is deferred — but the test name and the PR's user-visible "survives restart with Redis" promise are not yet in the suite.

### Documentation

- `dcrResolutionCache` interface doc and `newInMemoryDCRResolutionCache` doc need updates per F4.
- `EmbeddedAuthServer.dcrStore` field comment + `DCRStore()` doc need a security note per F10.
- The `Storage` interface doc should call out the DCR-credential reach if F6's design lands as embedded.

### Verification notes

Three findings raised by individual agents were verified against the post-PR file state and dropped as false positives — they cited the deleted (`-`) lines in the diff as if they were still in the new file. Specifically: (a) one HIGH security finding claimed `RedisStorage` does not implement `DCRCredentialStore` — verified false at `pkg/authserver/storage/redis.go:2040` on the base branch `dcr-3b_issue_5184`; (b) two findings about the `dcrResolutionCache` interface doc retaining "in-memory" / "never expired" claims — verified false; the post-PR file at lines 33-46 contains only the new "thin adapter" framing.

### Process notes

- Self-review caveat: the authenticated user is the PR author, so the review event is necessarily `COMMENT` (GitHub rejects REQUEST_CHANGES on own PRs). The findings would otherwise lean toward `REQUEST_CHANGES` due to the cluster of MEDIUM consensus items.
- Per `.claude/rules/pr-creation.md`, this PR exceeds the 400-line / 10-file budget — the author already acknowledged this in the body. Splitting the integration test (217 LOC) into a follow-up would put the wiring change inside budget.

---

pkg/authserver/runner/embeddedauthserver.go
```diff
@@ -173,6 +210,22 @@ func (e *EmbeddedAuthServer) KeyProvider() keys.KeyProvider {
...
+// callers can use it to inspect persisted RFC 7591 client registrations
+// without bypassing the storage backend the authserver itself reads from.
+//
+// This accessor mirrors IDPTokenStorage / UpstreamTokenRefresher on this type
+// and is intended for admin / diagnostic code paths and integration tests
+// that need to verify that the DCR resolver is wired into the same backend
+// the authserver writes to. It does NOT expose any I/O the
+// runner-DCRCredentialStore adapter does not already provide; the returned
+// value is the same storage.DCRCredentialStore the constructor surfaced.
+func (e *EmbeddedAuthServer) DCRStore() storage.DCRCredentialStore {
```
tgrunnagle, line 229: **F10 — `DCRStore()` accessor: secret reach + delegation pattern + lifecycle docs** (consensus 8/10, MEDIUM)

Three concerns about this new public accessor cluster together:

1. **Secret reach** (security): the accessor returns the raw `storage.DCRCredentialStore`, so any caller of `*EmbeddedAuthServer` can read cleartext `ClientSecret` and `RegistrationAccessToken` for every registered upstream. The doc justifies this as "admin/diagnostic + integration tests" but Go has no scope mechanism to enforce that. The integration test added in this PR is currently the only consumer.
2. **Delegation asymmetry** (architecture): `IDPTokenStorage()` and `UpstreamTokenRefresher()` delegate through `e.server.*` (the authserver owns the seam). `DCRStore()` reads `e.dcrStore` directly. This stores the same value twice and creates a drift window if `authserver.Server` ever swaps storage.
3. **Lifecycle docs** (concurrency): the doc does not state that the returned handle's lifetime is bound to `EmbeddedAuthServer.Close()` — calling methods after Close has backend-specific behavior (works on MemoryStorage by accident, errors on RedisStorage's closed connection pool).

**Recommendation** (layered):
- (a) If only tests need this, move it to a test-only seam — `runner_export_test.go` with `var DCRStoreForTest = ...` — and remove the public method.
- (b) If diagnostic access is genuinely required, define a narrower interface that exposes existence/metadata (issuer, client_id, age, has_secret bool) without surfacing raw secrets.
- (c) Add `DCRStore()` to the `authserver.Server` interface and have `EmbeddedAuthServer` delegate through `e.server.DCRStore()` to match the sibling pattern, eliminating the redundant `e.dcrStore` field.
- (d) At minimum, add a `// SECURITY:` doc note: "returns the raw secret-bearing interface; callers MUST NOT log or render the returned values; lifetime is bound to `EmbeddedAuthServer.Close` — calling methods afterwards is backend-specific."

*Raised by: concurrency, architecture, security*

pkg/authserver/runner/dcr_store.go
tgrunnagle: Addressed in 39dfc62b: replaced `newInMemoryDCRResolutionCache()` with a test-only `newMemoryDCRStore(t *testing.T)` in a new `dcr_testhelpers_test.go` that calls `t.Cleanup(stor.Close)`, plugging the cleanupLoop goroutine leak across all ~32 call sites.

pkg/authserver/runner/dcr_store_test.go
tgrunnagle: Addressed in 39dfc62b: renamed `TestInMemoryDCRResolutionCache_*` → `TestStorageBackedStore_*` so the suite names match the type they actually exercise. Updated the stale `sync.RWMutex` reference in the ConcurrentAccess docstring.

pkg/authserver/runner/dcr_store.go
tgrunnagle: Addressed in 39dfc62b: dropped the production-marketing paragraphs. The new `newMemoryDCRStore` docstring states the test-only purpose plus the cleanup contract directly; cross-replica/durability prose now lives only on the storage backends where it belongs.

pkg/authserver/integration_dcr_restart_test.go
tgrunnagle: Addressed in 39dfc62b: renamed to `TestEmbeddedAuthServer_DCRStorePersistsAcrossClose` and refactored to issue the Get BEFORE `Close`, dropping the silent dependency on `MemoryStorage`'s post-Close readability. The 44-line gap-narrating docstring is now scoped tightly to what is exercisable today plus a pointer to the deferred Redis-Sentinel follow-up.

pkg/authserver/storage/types.go
tgrunnagle: Addressed in 2e59bec2: `Storage` no longer embeds `DCRCredentialStore`. The runner and the authserver constructors now obtain the DCR-capable handle via an explicit `stor.(DCRCredentialStore)` assertion at the boundary; the per-backend `var _ DCRCredentialStore = (*MemoryStorage)(nil)` and `(*RedisStorage)(nil)` checks preserve the compile-time guarantee, so the runtime assertion is provably safe while keeping the wider `Storage` surface narrow. `MockStorage` was regenerated and no longer carries DCR methods either.

pkg/authserver/runner/dcr_store.go
tgrunnagle: Addressed in cbff536f: added `TestResolutionCredentialsRoundTrip` (table-driven) covering the preserved / dropped / key-recovered field groups plus the nil-shortcircuit. Added MUST-update-both-converters comments on `DCRResolution` (dcr.go) and `storage.DCRCredentials` (storage/types.go) so a future field addition shows the converter obligation at the struct definition. `ProviderName` is the one DCRCredentials field with no DCRResolution counterpart; the storage docstring already classifies it as "debug/audit only" and no consumer reads it, so I left it intentionally unset and pinned that decision with an assertion in the test.

pkg/authserver/runner/embeddedauthserver.go
tgrunnagle: Addressed in cbff536f: both `closeErr` and `retErr` now route through `sanitizeErrorForLog` before reaching the Warn record (so a wrapped DCR failure that inlines an upstream `/register` response body cannot leak userinfo, query, or fragment components). Renamed `"original_error"` → `"cause"` to match the package-wide vocabulary. Added `TestNewEmbeddedAuthServer_DeferredCleanupSanitizesLog` which captures the Warn record by swapping `slog.Default()` and asserts both that literal secret markers are scrubbed and that host components survive for operator correlation. (The test cannot run `t.Parallel()` because of the global slog swap; the trade-off is documented at the test with a `//nolint:paralleltest` rationale.)

pkg/authserver/integration_dcr_restart_test.go
tgrunnagle: Addressed in 39dfc62b: added a "DO NOT COPY THIS A THIRD TIME — extract to testhelpers" tripwire on both `newMockUpstreamAS` (authserver_test) and `newMockAuthorizationServer` (runner). The next caller is now directed at a shared `pkg/authserver/internal/testhelpers` package; the full extraction is deferred per the recommendation and will be a small, bounded rewrite when the third caller arrives.

pkg/authserver/runner/embeddedauthserver.go
tgrunnagle: Addressed in 2e59bec2 (option c — delegate through `authserver.Server`): added `DCRStore() storage.DCRCredentialStore` to the `authserver.Server` interface, implemented on the impl by capturing the DCR-asserted handle at `newServer`, and `EmbeddedAuthServer.DCRStore()` now delegates via `e.server.DCRStore()`. The redundant `e.dcrStore` field is removed, eliminating the drift window. The accessor's docstring carries an explicit SECURITY block plus a lifecycle paragraph ('returned handle's lifetime is bound to Server.Close — methods invoked after Close have backend-specific behavior'). Combined with F6 the secret-bearing surface is no longer reachable from `storage.Storage`-typed consumers.

tgrunnagle:
Re: F11 (PR state / size / e2e gap from the multi-agent review body): noted, no commit. The size-over-budget acknowledgement and the deferred Sentinel-restart e2e are already called out in the PR body, and the PR remains in draft pending the size discussion. The Sentinel-restart gap is now explicitly recorded in the docstring on `TestEmbeddedAuthServer_DCRStorePersistsAcrossClose` (39dfc62b) so a future contributor adding the harness has a single grep target.

jhrozek (approved):
well-designed and executed. The only issue I see (and it must be down to Claude updates) is that there are comments that won't make sense in the future (e.g. referring to "Phase 3").

PR #5198, opened 2026-05-05
Extract DCR resolver into pkg/auth/dcr

### LINKED ISSUE #5145: Consolidate the two RFC 7591 DCR client implementations into pkg/auth/dcr
## Context

Surfaced during review of #5042 ([comment](https://github.com/stacklok/toolhive/pull/5042#issuecomment-4352821640)).

We are about to ship two RFC 7591 Dynamic Client Registration client implementations side by side:

1. **`pkg/authserver/runner/dcr.go`** (this PR, sub-issue B of #4978) — the embedded authserver's resolver: stateful, in-process cache (`DCRCredentialStore`), `singleflight.Group` deduplication, scope canonicalisation, RFC 7591 §3.2.1 expiry-driven refetch, S256 PKCE gating, panic recovery, etc. Public-PKCE / confidential-client capable depending on what the upstream advertises.
2. **`pkg/auth/discovery/discovery.go::PerformOAuthFlow`** — the CLI flow. Already calls `oauthproto.RegisterClientDynamically` and persists credentials as `OAuthFlowResult.ClientID` / `ClientSecret`, which the remote handler later reads via `Config.CachedClientID` / `CachedClientSecretRef` / `CachedRegTokenRef` (see `pkg/auth/remote/handler.go`).

Same RFC role, different client profile, different concurrency model, different cache shape. Without consolidation each side will diverge: one will gain features (S256 gate, expiry refetch, singleflight) the other won't, and a third consumer landing later will inherit whichever one the author happened to import.

## Proposal

Lift the shared logic into a new `pkg/auth/dcr` package consumed by both call sites:

- **Stateful concerns** belong in `pkg/auth/dcr`:
  - The credential cache (`DCRCredentialStore` interface + in-memory backend, eventually a Redis-backed one — see issue #5135 for the SSRF hardening that lives downstream of this).
  - `singleflight.Group` keyed on the canonical key tuple.
  - Scope set canonicalisation (currently `scopesHash` in the resolver).
  - Token-endpoint auth-method selection with the RFC 7636 / OAuth 2.1 S256 gate.
  - RFC 7591 §3.2.1 expiry-driven cache invalidation.
  - The bearer-token transport with redirect refusal.
  - Panic recovery around the registration body.
  - Endpoint resolution (discovery vs direct, with the path-preserving synthesis fallback).

- **Stateless RFC 7591 primitives** belong in `pkg/oauthproto`:
  - `DynamicClientRegistrationRequest` / `DynamicClientRegistrationResponse` (already there).
  - `RegisterClientDynamically` (already there).
  - `FetchAuthorizationServerMetadata*` (already there).
  - Anything else that's pure "RFC 7591 wire shape, no state".

- **Profile differences** stay at the call sites — the resolver caller chooses its own client profile (public PKCE vs confidential), redirect URI policy, and credential persistence strategy. The shared package is profile-agnostic.

## Migration sketch

1. Land sub-issues C (#5039) and D of the current epic so the resolver has its consumers wired up.
2. Extract `pkg/auth/dcr` from `pkg/authserver/runner/dcr.go` + `dcr_store.go`. The embedded authserver becomes the first consumer; its existing tests carry over with minimal surface change (mostly import paths).
3. Migrate `pkg/auth/discovery/discovery.go::PerformOAuthFlow` to consume `pkg/auth/dcr` instead of calling `oauthproto.RegisterClientDynamically` directly. The CLI will then inherit S256 gating, expiry refetch, redirect refusal, and so on automatically.
4. Audit `pkg/auth/remote/handler.go::CachedClientID` / `CachedClientSecretRef` / `CachedRegTokenRef` usage. Either replace with the shared cache interface, or treat those fields as the CLI profile's persistence backend implementing the shared interface.

## Why now

> Easier to migrate now than after another consumer lands.

The third consumer almost certainly arrives in the next 1–2 quarters (every new OAuth2 upstream that supports DCR is a candidate). Migrating two sites is a straightforward refactor; migrating three or four with subtly diverged behaviour is a multi-week archaeology project.

## Acceptance

- `pkg/auth/dcr` exists and exports a stateful resolver consumed by both the embedded authserver and the CLI flow.
- Stateless RFC 7591 wire-shape helpers live in `pkg/oauthproto`.
- No direct calls to `oauthproto.RegisterClientDynamically` outside `pkg/auth/dcr` (search-grep enforces this).
- Behaviour properties added during #5042 review (S256 gate, expiry-driven refetch, redirect refusal, panic recovery, singleflight) apply to the CLI flow without per-call-site re-implementation.

## References

- PR #5042 (sub-issue B of #4978)
- Comment that prompted this: https://github.com/stacklok/toolhive/pull/5042#issuecomment-4352821640
- Related: #5135 (SSRF hardening on operator-supplied DCR URLs — lives downstream of the consolidated package).

### LINKED ISSUE #5220: Extract DCR resolver into pkg/auth/dcr (package + EmbeddedAuthServer migration)
## Description

Sub-issue 4a of #5145. Extract the stateful RFC 7591 DCR resolver and credential store from `pkg/authserver/runner` into a new shared `pkg/auth/dcr` package, and migrate `EmbeddedAuthServer` to consume it. The CLI flow at `pkg/auth/discovery::PerformOAuthFlow` is migrated separately in sub-issue 4b (#5219); this slice intentionally stops there to keep the diff under the project's 400-LOC / 10-file PR-size cap.

This sub-issue is the **package-creation** half of the parent consolidation: it stands up the destination package and wires the first consumer. AC#2 of the parent (`pkg/oauthproto` already holds the stateless RFC 7591 helpers) is satisfied trivially because that move predates this work. AC#1, AC#3, and AC#4 of the parent are met *partially* and close fully when 4b lands.

## Context

The two RFC 7591 DCR client implementations identified in #5145 — `pkg/authserver/runner/dcr.go` (embedded authserver) and `pkg/auth/discovery::PerformOAuthFlow` (CLI flow) — were on a path to diverging. The embedded-authserver side had accumulated review-derived behaviours (S256 PKCE gating, RFC 7591 §3.2.1 expiry refetch, bearer-token transport with redirect refusal, panic recovery, singleflight deduplication) that the CLI side did not inherit. Lifting the shared logic into a single package is the prerequisite for the CLI to inherit those behaviours via consumer migration.

This sub-issue does the **lift only**, not the second migration. The package's public API still takes embedded-authserver-shaped types (`*authserver.OAuth2UpstreamRunConfig`, `*upstream.OAuth2Config`); designing a profile-neutral input type with one consumer in hand would be speculative. That refactor is the first acceptance criterion of #5219.

**Dependencies**: none.
**Blocks**: sub-issue 4b (#5219) — the CLI migration consumes the package landed here.

### Package extraction — `pkg/auth/dcr`

- [x] New package `pkg/auth/dcr` exists.
- [x] Files moved with `git mv` so history is preserved and the diff renders as renames:
  - [x] `pkg/authserver/runner/dcr.go` → `pkg/auth/dcr/resolver.go`
  - [x] `pkg/authserver/runner/dcr_store.go` → `pkg/auth/dcr/store.go`
  - [x] `pkg/authserver/runner/dcr_test.go` → `pkg/auth/dcr/resolver_test.go`
  - [x] `pkg/authserver/runner/dcr_store_test.go` → `pkg/auth/dcr/store_test.go`
- [x] Public symbols are renamed to drop the redundant `DCR` prefix on packagisation (e.g. `DCRResolution` → `dcr.Resolution`, `DCRCredentialStore` → `dcr.CredentialStore`, `resolveDCRCredentials` → `dcr.ResolveCredentials`, `consumeResolution` → `dcr.ConsumeResolution`, `applyResolutionToOAuth2Config` → `dcr.ApplyResolutionToOAuth2Config`, `logDCRStepError` → `dcr.LogStepError`).
- [x] Package doc comment in `pkg/auth/dcr/resolver.go` documents the process-global `singleflight.Group` (it is package-level state shared by all consumers, not per-instance) and explicitly labels the profile-agnostic framing as a *target state* for sub-issue 4b rather than as a description of the current API shape.

### EmbeddedAuthServer migration — `pkg/authserver/runner/embeddedauthserver.go`

- [x] `EmbeddedAuthServer` consumes the new package — imports `pkg/auth/dcr` and calls the renamed exports.
- [x] Doc comments on `EmbeddedAuthServer.DCRStore()` and `buildUpstreamConfigs` use the new public names (`dcr.CredentialStore`, `dcr.ConsumeResolution`, `dcr.ApplyResolutionToOAuth2Config`, `dcr.ResolveCredentials`, `dcr.LogStepError`) — no stale identifiers.
- [x] Externally-observable DCR behaviour of the embedded authserver is unchanged: registration shape, caching, error reporting, structured logging, expiry refetch, S256 gating.

### Drift guard for duplicated `resolveSecret`

- [x] `resolveSecret` is duplicated identically in `pkg/auth/dcr` and `pkg/authserver/runner` because the new package must remain reachable-from-runner but not back. Promotion to a shared helper is deferred to 4b (where a third call site emerges).
- [x] A parallel `TestResolveSecret` / `TestResolveSecretWithEnvVar` suite in `pkg/auth/dcr/secret_test.go` mirrors the runner-package twin's observable contract, so any future drift between the two copies fails CI on at least one side.

### Defensive hardening

- [x] `endpointsFromMetadata` nil-guards its `metadata` parameter; the function's doc comment names the `oauthproto` contract it relies on.
- [x] A `flightKeyOf(Key) string` helper extracts the canonical key shape used by the singleflight group, so future tests and consumers can inspect it without re-deriving the format.

### Cross-cutting

- [x] `task license-check` — clean.
- [x] `go test -count=1 -race ./pkg/auth/dcr/... ./pkg/authserver/...` — all pass.
- [x] No new direct callers of `oauthproto.RegisterClientDynamically` introduced. (The pre-existing CLI caller at `pkg/auth/discovery/discovery.go:747` is intentionally left in place for 4b to migrate; the global ban in parent AC#3 lands with 4b's CI grep guard.)

## Out of Scope

- CLI flow migration — sub-issue 4b (#5219).
- Designing a profile-neutral resolver input — sub-issue 4b (#5219).
- Lifting `resolveSecret` into a shared helper — deferred to 4b once a third call site exists.
- The CI grep guard that bans `oauthproto.RegisterClientDynamically` outside `pkg/auth/dcr` — sub-issue 4b (#5219).
- All other items in the parent issue's "Out of Scope" section.

## References

- Parent issue: #5145.
- Sibling sub-issue: #5219 — CLI flow migration; closes #5145.
- Pull request: #5198 — implementation of this sub-issue.
- Originating PR: #5042 — defined the review-property behaviours that motivated consolidation.
- `.claude/rules/go-style.md`, `.claude/rules/testing.md`, `.claude/rules/pr-creation.md`, `CLAUDE.md`.
- RFC 7591 §3.2.1.

tgrunnagle:
## Summary

- **Why**: Issue #5145 calls for consolidating two parallel RFC 7591 Dynamic Client Registration (DCR) implementations (`pkg/authserver/runner/dcr.go` and `pkg/auth/discovery::PerformOAuthFlow`) into a single shared `pkg/auth/dcr` package, so review-derived behaviours (S256 PKCE gate, RFC 7591 §3.2.1 expiry refetch, redirect refusal, panic recovery, singleflight) cannot diverge between the embedded authserver and the CLI flow as future consumers land. The full migration is too large for a single PR, so this slice (sub-issue 4a) does the package extraction and migrates the embedded authserver only.
- **What**: Extracts the DCR resolver and credential store from `pkg/authserver/runner` into a new `pkg/auth/dcr` package and migrates `EmbeddedAuthServer` to consume it. File moves use `git mv` so the diff renders as renames rather than delete+add. Public symbols are renamed to drop the redundant `DCR` prefix on packagisation (e.g., `DCRResolution` → `dcr.Resolution`, `DCRCredentialStore` → `dcr.CredentialStore`, `resolveDCRCredentials` → `dcr.ResolveCredentials`). Adds a parallel `TestResolveSecret` suite mirroring the runner-package twin so any future drift between the two `resolveSecret` copies fails CI.

Closes #5220 (sub-issue 4a). Part of #5145 — does **not** close the parent. Sibling sub-issue #5219 (4b — CLI flow migration in `pkg/auth/discovery::PerformOAuthFlow`) closes #5145 when it lands; AC#1, AC#3, and AC#4 of the parent depend on it.

## Stacking

This branch is stacked on PR #5196 (`dcr-3c_issue_5185` — "Wire persistent DCRCredentialStore into EmbeddedAuthServer"), which is itself stacked on PRs #5186 and #5195. The earlier commits on this branch (`c0fed523` through `781a0b97`) implement sub-issues #5183 / #5184 / #5185 (the persistent DCR store and Redis backend) and are shown for context only — review against PR #5196 once it merges, or review the two commits unique to this branch in isolation:

- `df84256f` Extract DCR resolver into `pkg/auth/dcr`
- `633ed9ed` Address code review feedback

## Type of change

## Test plan

- [x] Unit tests (`task test`) — `go test -count=1 -race ./pkg/auth/dcr/... ./pkg/authserver/...` all pass
- [x] License check (`task license-check`) — clean

## Changes

| File | Change |
|------|--------|
| `pkg/auth/dcr/resolver.go` | Renamed from `pkg/authserver/runner/dcr.go`; public symbols renamed to drop the `DCR` prefix; package doc updated with a `# Concurrency` section documenting the process-global singleflight; new `flightKeyOf(Key)` helper makes the canonical key shape inspectable; `endpointsFromMetadata` gains a defensive nil check and an `oauthproto` contract note. |
| `pkg/auth/dcr/store.go` | Renamed from `pkg/authserver/runner/dcr_store.go`; constructors renamed (`NewInMemoryStore`, `NewStorageBackedStore`); type renamed to `CredentialStore`. |
| `pkg/auth/dcr/resolver_test.go` | Renamed from `pkg/authserver/runner/dcr_test.go`; updated for new package boundary and renamed symbols. |
| `pkg/auth/dcr/store_test.go` | Renamed from `pkg/authserver/runner/dcr_store_test.go`; updated for new package boundary. |
| `pkg/auth/dcr/secret_test.go` | New parallel `resolveSecret` suite that mirrors the runner-package twin's observable contract so future drift between the two duplicated helpers fails CI on at least one side. |
| `pkg/authserver/runner/embeddedauthserver.go` | Migrated to consume `pkg/auth/dcr`; doc comments on `buildUpstreamConfigs` and `DCRStore()` updated to use the new public names. |
| `pkg/authserver/runner/embeddedauthserver_test.go` | Updated import paths and constructor names for the moved types. |

No. This is an internal refactor. The embedded authserver's externally-observable DCR behaviour (registration, caching, error reporting, structured logging, expiry refetch, S256 gating) is unchanged.

## Implementation plan

<details>

Sub-issue 4a (#5220) of #5145 — extract the resolver into `pkg/auth/dcr` and migrate the embedded authserver only. Defer the `pkg/auth/discovery::PerformOAuthFlow` migration to sub-issue 4b (#5219) so this slice stays under the 400-line / 10-file PR-size limit.

Profile-neutral input shape: the package's API still takes embedded-authserver types directly. Designing a profile-neutral input type with only one consumer in hand would be speculative; the right design moment is when 4b lands the CLI flow as the second consumer. The package doc comment in `pkg/auth/dcr/resolver.go` calls this coupling out explicitly and labels the profile-agnostic framing as a target state for 4b rather than the current API shape.

Duplication of `resolveSecret`: deliberately duplicated across `pkg/authserver/runner` and `pkg/auth/dcr` because the DCR package must stay reachable-from-runner-but-not-back. Promoting to a shared helper is deferred until 4b lands a third call site. Drift between the two copies is guarded by parallel test suites with identical observable contracts.

</details>

- **Why renames, not delete+add**: file moves use `git mv` so the diff renders as renames. The percentage similarity in `git log --stat --follow` reflects that the bulk of the move is mechanical and the substantive deltas are confined to (a) the package doc rewrite at the top of `resolver.go`, (b) the `flightKeyOf` extraction and singleflight comment, (c) the `endpointsFromMetadata` nil guard, and (d) the import/symbol rename sweep in `embeddedauthserver.go`.
- **Singleflight is process-global, not per-instance**: documented at the package surface and on the `dcrFlight` var. The cross-consumer caveat ("third consumer with colliding redirect URI" failure mode) is named explicitly so the next migration cannot silently re-introduce a collision.
- **Acceptance criteria status against #5145** (as per the commit message of `df84256f`):
  - AC#1 (`pkg/auth/dcr` exists, exports a stateful resolver, consumed by both call sites): partially met — package exists and is consumed by `EmbeddedAuthServer`. CLI flow migration is sub-issue 4b (#5219).
  - AC#2 (stateless RFC 7591 helpers in `pkg/oauthproto`): met (already satisfied before this PR).
  - AC#3 (no direct `oauthproto.RegisterClientDynamically` calls outside `pkg/auth/dcr`): not yet met — `pkg/auth/discovery` still calls it. Resolved by sub-issue 4b (#5219).
  - AC#4 (review-property behaviours apply to CLI flow): not yet met — same dependency on 4b (#5219).

### DIFF
--- a/pkg/auth/dcr/resolver.go
+++ b/pkg/auth/dcr/resolver.go
@@ -229,23 +229,26 @@ func NeedsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
 	return rc.ClientID == "" && rc.DCRConfig != nil
 }
 
-// ConsumeResolution copies resolved credentials and endpoints from res into
-// rc and consumes rc.DCRConfig (sets it to nil), transitioning the run-
-// config copy from "DCR-pending" (ClientID == "" && DCRConfig != nil) to
-// "DCR-resolved" (ClientID populated && DCRConfig == nil). The "consume"
-// name is deliberate: a second call after the first is a no-op only
-// because the first cleared DCRConfig — this is a one-shot state
-// transition, not an idempotent default-fill.
+// ConsumeResolution returns a copy of rc with the resolved credentials and
+// endpoints from res copied in and DCRConfig consumed (set to nil),
+// transitioning the run-config from "DCR-pending" (ClientID == "" &&
+// DCRConfig != nil) to "DCR-resolved" (ClientID populated && DCRConfig
+// == nil). The "consume" name is deliberate: a second call on the
+// returned value is a no-op only because the first cleared DCRConfig —
+// this is a one-shot state transition, not an idempotent default-fill.
 //
-// Callers must pass a COPY of the upstream run-config so the caller's
-// original is unaffected; ConsumeResolution does not clone rc internally.
+// rc is taken by value and the modified copy is returned. The caller's
+// original is never observably mutated; the value-in / value-out shape
+// makes the no-mutation contract compile-time enforced rather than a
+// prose discipline the caller is required to remember. Pointer-typed
+// fields (DCRConfig) share storage with the caller's copy via the struct
+// shallow-copy, but the only mutation here is to assign nil to the
+// copy's DCRConfig — nil-assignment to the local field does not reach
+// back through the original pointer.
 //
 // Why DCRConfig is cleared: OAuth2UpstreamRunConfig.Validate enforces
 // ClientID xor DCRConfig — a resolved copy that left DCRConfig set would
-// fail the validator that runs downstream in buildPureOAuth2Config. The
-// caller's *original* OAuth2Config is unaffected because
-// buildUpstreamConfigs deep-copies before resolution; only the post-
-// resolution copy is mutated here.
+// fail the validator that runs downstream in buildPureOAuth2Config.
 //
 // ClientID, the endpoints, and RedirectURI are written only when rc leaves
 // them empty — explicit caller configuration always wins. The conditional
@@ -261,15 +264,15 @@ func NeedsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
 // RedirectURI, which authserver.Config validation requires.
 //
 // Note on ClientSecret: ConsumeResolution does NOT write the resolved
-// secret to rc because OAuth2UpstreamRunConfig models secrets as file-or-
-// env references only. To propagate the DCR-resolved secret into the
-// final upstream.OAuth2Config, callers must pair this call with
+// secret because OAuth2UpstreamRunConfig models secrets as file-or-env
+// references only. To propagate the DCR-resolved secret into the final
+// upstream.OAuth2Config, callers must pair this call with
 // ApplyResolutionToOAuth2Config once the config has been built. Keeping
 // the two helpers side-by-side localises the DCR-specific application
 // logic.
-func ConsumeResolution(rc *authserver.OAuth2UpstreamRunConfig, res *Resolution) {
-	if rc == nil || res == nil {
-		return
+func ConsumeResolution(rc authserver.OAuth2UpstreamRunConfig, res *Resolution) authserver.OAuth2UpstreamRunConfig {
+	if res == nil {
+		return rc
 	}
 	if rc.ClientID == "" {
 		rc.ClientID = res.ClientID
@@ -284,28 +287,35 @@ func ConsumeResolution(rc *authserver.OAuth2UpstreamRunConfig, res *Resolution) 
 	if rc.RedirectURI == "" {
 		rc.RedirectURI = res.RedirectURI
 	}
+	return rc
 }
 
-// ApplyResolutionToOAuth2Config overlays the DCR-resolved ClientSecret onto
-// a built *upstream.OAuth2Config. This is the companion to
+// ApplyResolutionToOAuth2Config returns a copy of cfg with the DCR-
+// resolved ClientSecret overlaid onto it. This is the companion to
 // ConsumeResolution: where that function writes fields representable in
 // the file-or-env run-config model, this one writes the inline-only
 // ClientSecret directly on the runtime config.
 //
-// The split exists because buildPureOAuth2Config intentionally retains a
-// narrow file-or-env contract (no DCR awareness) and because OAuth2's
-// ClientSecret on the run-config is modelled as a reference rather than
-// an inline string. Any future output path from OAuth2UpstreamRunConfig
-// to upstream.OAuth2Config must call BOTH ConsumeResolution (run-config
-// side) AND ApplyResolutionToOAuth2Config (built-config side) to get a
-// fully-resolved DCR client. Forgetting the second call leaves
-// ClientSecret empty and produces silent auth failures at request time —
-// the type system does not enforce the pair, so the invariant lives here.
-func ApplyResolutionToOAuth2Config(cfg *upstream.OAuth2Config, res *Resolution) {
-	if cfg == nil || res == nil {
-		return
+// cfg is taken by value and the modified copy is returned, mirroring
+// ConsumeResolution. The no-mutation contract is compile-time enforced
+// by the signature rather than a prose discipline.
+//
+// The split between these two helpers exists because buildPureOAuth2Config
+// intentionally retains a narrow file-or-env contract (no DCR awareness)
+// and because OAuth2's ClientSecret on the run-config is modelled as a
+// reference rather than an inline string. Any future output path from
+// OAuth2UpstreamRunConfig to upstream.OAuth2Config must call BOTH
+// ConsumeResolution (run-config side) AND ApplyResolutionToOAuth2Config
+// (built-config side) to get a fully-resolved DCR client. Forgetting the
+// second call leaves ClientSecret empty and produces silent auth failures
+// at request time — the type system does not enforce the pair, so the
+// invariant lives here.
+func ApplyResolutionToOAuth2Config(cfg upstream.OAuth2Config, res *Resolution) upstream.OAuth2Config {
+	if res == nil {
+		return cfg
 	}
 	cfg.ClientSecret = res.ClientSecret
+	return cfg
 }
 
 // Step identifiers for structured error logs emitted by the caller of
--- a/pkg/auth/dcr/resolver.go
+++ b/pkg/auth/dcr/resolver.go
@@ -1,15 +1,57 @@
 // SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-package runner
+// Package dcr is the shared RFC 7591 Dynamic Client Registration client used
+// by every consumer in the codebase that needs to register a downstream
+// OAuth 2.x client at runtime. The package owns the stateful concerns of the
+// flow — credential cache, in-process singleflight deduplication, scope-set
+// canonicalisation, token-endpoint auth-method selection (with the RFC 7636 /
+// OAuth 2.1 S256 PKCE gate), RFC 7591 §3.2.1 expiry-driven cache invalidation,
+// the bearer-token transport with redirect refusal, and panic recovery around
+// the registration body. Stateless RFC 7591 wire-shape primitives live in
+// pkg/oauthproto.
+//
+// # Concurrency
+//
+// The package maintains a process-global singleflight keyed on the tuple
+// (issuer, redirectURI, scopesHash) so concurrent ResolveCredentials calls
+// across all consumers in a single process coalesce when their cache keys
+// match. Consumers that share any of those three values will share a flight
+// — the deduplication is a feature for the embedded authserver but means
+// callers cannot assume per-call-site flight isolation. See the dcrFlight
+// doc comment below for the rationale.
+//
+// # Current API coupling — sub-issue 4a only
+//
+// As of issue #5145 sub-issue 4a (the slice that lifted this code out of
+// pkg/authserver/runner), the public functions on this package take
+// embedded-authserver types — *authserver.OAuth2UpstreamRunConfig,
+// *authserver.DCRUpstreamConfig, *upstream.OAuth2Config — directly on
+// their signatures. This matches the embedded authserver's existing
+// internal shapes verbatim and was the cheapest move-only change.
+//
+// The CLI flow migration in sub-issue 4b will introduce the second
+// consumer (pkg/auth/discovery::PerformOAuthFlow) and is the right
+// trigger for replacing those parameters with a profile-neutral input
+// type — designing the neutral shape now, with only one consumer in
+// hand, would be speculative. Until 4b lands, callers outside the
+// embedded authserver MUST adapt their inputs to the authserver types
+// at the call site, and the "profile-agnostic" framing in this package's
+// charter is a target state, not the current state of the API.
+//
+// See issue #5145 for the design discussion that motivated lifting this out
+// of pkg/authserver/runner.
+package dcr
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"log/slog"
 	"net/http"
 	"net/url"
+	"os"
 	"regexp"
 	"runtime/debug"
 	"slices"
@@ -25,14 +67,14 @@ import (
 	"github.com/stacklok/toolhive/pkg/oauthproto"
 )
 
-// dcrFlight coalesces concurrent resolveDCRCredentials calls that share the
-// same DCRKey. Two goroutines hitting the resolver for the same upstream and
+// dcrFlight coalesces concurrent ResolveCredentials calls that share the
+// same Key. Two goroutines hitting the resolver for the same upstream and
 // scope set will both miss the cache, so without coalescing they would both
 // call RegisterClientDynamically and the loser's registration would become
 // orphaned at the upstream IdP — an operator-visible cleanup task and
 // possibly a transient startup failure if the upstream rate-limits
 // concurrent registrations. Followers wait for the leader's result and
-// observe the same DCRResolution.
+// observe the same Resolution.
 //
 // Lifetime: process-wide. This intentionally contrasts with
 // EmbeddedAuthServer.dcrStore, which is per-instance. The asymmetry is
@@ -44,7 +86,29 @@ import (
 // decides whether the resolution is fresh enough to reuse. A future
 // Redis-backed store would still want this in-process gate so a single
 // replica does not double-register against itself.
+//
+// Cross-consumer caveat (matters once issue #5145 sub-issue 4b lands the
+// CLI flow as the second consumer): because dcrFlight is package-global,
+// two consumers that happen to construct identical Keys (same issuer, same
+// redirect URI, same scopes hash) will share a single in-flight
+// registration even if they semantically want different client profiles.
+// The current call sites do not collide — the embedded authserver's
+// redirect URI lives on the AS origin, the CLI flow's lives on a
+// loopback — but a future consumer that defaults its redirect URI into
+// either of those spaces would silently coalesce. Keep this in mind when
+// adding a third consumer.
 var dcrFlight singleflight.Group
+
+// flightKeyOf canonicalises a Key into the singleflight string used by
+// dcrFlight. The "\n" separator is safe because newline is not a valid byte
+// in any of the three components: OAuth scope tokens (RFC 6749 §3.3), URI
+// reference characters (RFC 3986 §2), or hex digests (the form
+// storage.ScopesHash always emits). Exposed as a function so tests and
+// future inspection helpers can compute the exact key the resolver would
+// route through dcrFlight without re-implementing the concatenation.
+func flightKeyOf(key Key) string {
+	return key.Issuer + "\n" + key.RedirectURI + "\n" + key.ScopesHash
+}
 
 // defaultUpstreamRedirectPath is the redirect path derived from the issuer
 // origin when the caller's run-config does not supply an explicit RedirectURI.
@@ -67,23 +131,23 @@ var authMethodPreference = []string{
 	"none",
 }
 
-// DCRResolution captures the full RFC 7591 + RFC 7592 response for a
+// Resolution captures the full RFC 7591 + RFC 7592 response for a
 // successful Dynamic Client Registration, together with the endpoints the
 // upstream advertises so the caller need not re-discover them.
 //
-// The struct is the unit of storage in dcrResolutionCache and the unit of
-// application via consumeResolution.
+// The struct is the unit of storage in CredentialStore and the unit of
+// application via ConsumeResolution.
 //
 // MUST update both converters (resolutionToCredentials and
-// credentialsToResolution in dcr_store.go) when adding, renaming, or
+// credentialsToResolution in store.go) when adding, renaming, or
 // removing a field here. The two converters are the seam between this
-// runner-side type and the persisted *storage.DCRCredentials shape; a
+// dcr-package type and the persisted *storage.DCRCredentials shape; a
 // field added here without a paired converter update will silently fail
 // to round-trip across an authserver restart, the exact "parallel types
 // drift" failure mode .claude/rules/go-style.md warns about. The
 // round-trip behaviour is pinned by TestResolutionCredentialsRoundTrip
-// in dcr_store_test.go.
-type DCRResolution struct {
+// in store_test.go.
+type Resolution struct {
 	// ClientID is the RFC 7591 "client_id" returned by the authorization
 	// server.
 	ClientID string
@@ -117,7 +181,7 @@ type DCRResolution struct {
 	// during registration. When the caller's run-config did not specify one,
 	// this holds the defaulted value derived from the issuer + /oauth/callback
 	// (via resolveUpstreamRedirectURI). Persisting it on the resolution lets
-	// consumeResolution write it back onto the run-config COPY so that
+	// ConsumeResolution write it back onto the run-config COPY so that
 	// downstream consumers (buildPureOAuth2Config, upstream.OAuth2Config
 	// validation) see a non-empty RedirectURI.
 	RedirectURI string
@@ -147,18 +211,18 @@ type DCRResolution struct {
 	CreatedAt time.Time
 }
 
-// needsDCR reports whether rc requires runtime Dynamic Client Registration.
+// NeedsDCR reports whether rc requires runtime Dynamic Client Registration.
 // A run-config needs DCR exactly when ClientID is empty and DCRConfig is
 // non-nil (the mutually-exclusive constraint is enforced by
 // OAuth2UpstreamRunConfig.Validate; this helper is a convenience check).
-func needsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
+func NeedsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
 	if rc == nil {
 		return false
 	}
 	return rc.ClientID == "" && rc.DCRConfig != nil
 }
 
-// consumeResolution copies resolved credentials and endpoints from res into
+// ConsumeResolution copies resolved credentials and endpoints from res into
 // rc and consumes rc.DCRConfig (sets it to nil), transitioning the run-
 // config copy from "DCR-pending" (ClientID == "" && DCRConfig != nil) to
 // "DCR-resolved" (ClientID populated && DCRConfig == nil). The "consume"
@@ -167,7 +231,7 @@ func needsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
 // transition, not an idempotent default-fill.
 //
 // Callers must pass a COPY of the upstream run-config so the caller's
-// original is unaffected; consumeResolution does not clone rc internally.
+// original is unaffected; ConsumeResolution does not clone rc internally.
 //
 // Why DCRConfig is cleared: OAuth2UpstreamRunConfig.Validate enforces
 // ClientID xor DCRConfig — a resolved copy that left DCRConfig set would
@@ -189,14 +253,14 @@ func needsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
 // it back here means the downstream upstream.OAuth2Config has a non-empty
 // RedirectURI, which authserver.Config validation requires.
 //
-// Note on ClientSecret: consumeResolution does NOT write the resolved
+// Note on ClientSecret: ConsumeResolution does NOT write the resolved
 // secret to rc because OAuth2UpstreamRunConfig models secrets as file-or-
 // env references only. To propagate the DCR-resolved secret into the
 // final upstream.OAuth2Config, callers must pair this call with
-// applyResolutionToOAuth2Config once the config has been built. Keeping
+// ApplyResolutionToOAuth2Config once the config has been built. Keeping
 // the two helpers side-by-side localises the DCR-specific application
 // logic.
-func consumeResolution(rc *authserver.OAuth2UpstreamRunConfig, res *DCRResolution) {
+func ConsumeResolution(rc *authserver.OAuth2UpstreamRunConfig, res *Resolution) {
 	if rc == nil || res == nil {
 		return
 	}
@@ -215,9 +279,9 @@ func consumeResolution(rc *authserver.OAuth2UpstreamRunConfig, res *DCRResolutio
 	}
 }
 
-// applyResolutionToOAuth2Config overlays the DCR-resolved ClientSecret onto
+// ApplyResolutionToOAuth2Config overlays the DCR-resolved ClientSecret onto
 // a built *upstream.OAuth2Config. This is the companion to
-// consumeResolution: where that function writes fields representable in
+// ConsumeResolution: where that function writes fields representable in
 // the file-or-env run-config model, this one writes the inline-only
 // ClientSecret directly on the runtime config.
 //
@@ -225,12 +289,12 @@ func consumeResolution(rc *authserver.OAuth2UpstreamRunConfig, res *DCRResolutio
 // narrow file-or-env contract (no DCR awareness) and because OAuth2's
 // ClientSecret on the run-config is modelled as a reference rather than
 // an inline string. Any future output path from OAuth2UpstreamRunConfig
-// to upstream.OAuth2Config must call BOTH consumeResolution (run-config
-// side) AND applyResolutionToOAuth2Config (built-config side) to get a
+// to upstream.OAuth2Config must call BOTH ConsumeResolution (run-config
+// side) AND ApplyResolutionToOAuth2Config (built-config side) to get a
 // fully-resolved DCR client. Forgetting the second call leaves
 // ClientSecret empty and produces silent auth failures at request time —
 // the type system does not enforce the pair, so the invariant lives here.
-func applyResolutionToOAuth2Config(cfg *upstream.OAuth2Config, res *DCRResolution) {
+func ApplyResolutionToOAuth2Config(cfg *upstream.OAuth2Config, res *Resolution) {
 	if cfg == nil || res == nil {
 		return
 	}
@@ -238,7 +302,7 @@ func applyResolutionToOAuth2Config(cfg *upstream.OAuth2Config, res *DCRResolutio
 }
 
 // Step identifiers for structured error logs emitted by the caller of
-// resolveDCRCredentials. These values flow through the "step" attribute so
+// ResolveCredentials. These values flow through the "step" attribute so
 // operators can narrow failures to a specific phase without parsing error
 // messages. They are reported only at the boundary log — see
 // dcrStepError — so a single failure produces a single slog.Error record.
@@ -255,13 +319,13 @@ const (
 // dcrStepError annotates a resolver error with the phase it was produced
 // in. The boundary caller (buildUpstreamConfigs) emits the single
 // slog.Error record for the failure; individual error branches inside
-// resolveDCRCredentials do not log so that each failure surfaces exactly
+// ResolveCredentials do not log so that each failure surfaces exactly
 // once in the combined log stream.
 //
 // RedirectURI is included when known so that operators can correlate the
 // failure with a specific upstream registration without parsing the
 // wrapped error string. Stack carries a captured stack trace for the
-// dcrStepRegister panic-recovery branch so logDCRStepError can include
+// dcrStepRegister panic-recovery branch so LogStepError can include
 // it in the single boundary record without the in-defer site emitting
 // its own duplicate slog.Error. A zero-value dcrStepError is invalid;
 // construct via newDCRStepError or the resolver's internal helpers.
@@ -293,7 +357,7 @@ func newDCRStepError(step, issuer, redirectURI string, err error) *dcrStepError 
 	}
 }
 
-// resolveDCRCredentials performs Dynamic Client Registration for rc against
+// ResolveCredentials performs Dynamic Client Registration for rc against
 // the upstream authorization server identified by rc.DCRConfig, caching the
 // resulting credentials in cache. On cache hit the resolver returns
 // immediately without any network I/O.
@@ -310,7 +374,7 @@ func newDCRStepError(step, issuer, redirectURI string, err error) *dcrStepError 
 // redirect and a cache key that does not identify the auth-server context.
 //
 // The caller is responsible for applying the returned resolution onto a COPY
-// of rc via consumeResolution (per the copy-before-mutate rule). This function
+// of rc via ConsumeResolution (per the copy-before-mutate rule). This function
 // neither mutates rc nor the cache on failure.
 //
 // Observability: this function never calls slog.Error directly — all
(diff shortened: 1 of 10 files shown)


pkg/auth/dcr/store.go
```diff
@@ -18,64 +18,78 @@ import (
...
+// NewInMemoryStore returns a thread-safe in-memory CredentialStore intended
+// for tests and single-replica development deployments. It is a thin adapter
+// over storage.NewMemoryStorage so the resolver-side cache and the
+// authserver's main storage backend share a single in-memory implementation.
+//
+// Production deployments should use a Redis-backed
+// storage.DCRCredentialStore (instantiated via storage.NewRedisStorage and
+// passed through this package's storage-backed adapter), which addresses
+// cross-replica sharing, durability, and cross-process coordination.
+func NewInMemoryStore() CredentialStore {
```
tgrunnagle: **`NewInMemoryStore` returns store whose cleanup goroutine cannot be stopped** (consensus 9/10, MEDIUM)

`storage.NewMemoryStorage()` spawns a background `cleanupLoop` goroutine on construction. The returned `CredentialStore` interface has no `Close()` method, so callers of `NewInMemoryStore()` have no way to stop the goroutine — it leaks for the process lifetime. The package's own test helper `newMemoryDCRStore` works around this by holding the underlying `*MemoryStorage` directly and registering `t.Cleanup(stor.Close)`, but this public constructor cannot replicate that wiring.

Pick one:
1. Remove `NewInMemoryStore()` from this PR until a real consumer with lifecycle management appears in 4b (lightest fix, also resolves the "public API ahead of caller" smell).
2. Return both the `CredentialStore` and an explicit shutdown closure: `func NewInMemoryStore() (CredentialStore, func() error)`.
3. Add `Close() error` to the `CredentialStore` interface so callers can release the goroutine.

Raised by: api-design, concurrency, test-quality, cross-cutting

pkg/auth/dcr/store.go
```diff
@@ -18,64 +18,78 @@ import (
...
+}
+
+// NewStorageBackedStore returns a CredentialStore that delegates to a
 // storage.DCRCredentialStore for durable persistence and translates
-// DCRResolution values into DCRCredentials at the boundary. The returned
+// Resolution values into DCRCredentials at the boundary. The returned
 // store is safe for concurrent use because the underlying
 // storage.DCRCredentialStore must be (per its interface contract).
-func newStorageBackedStore(backend storage.DCRCredentialStore) dcrResolutionCache {
+func NewStorageBackedStore(backend storage.DCRCredentialStore) CredentialStore {
```
tgrunnagle, line 67: **`NewStorageBackedStore` accepts `nil` backend silently** (consensus 7/10, MEDIUM)

No nil check on `backend`. A `nil` argument propagates to `&storageBackedStore{backend: nil}`, which looks valid until the first `Get`/`Put` call dereferences `s.backend` and panics with a stack trace pointing at `s.backend.GetDCRCredentials` rather than the caller that passed `nil`. Per `.claude/rules/go-style.md` “Constructor Validation: Fail Loudly on Invalid Input”.

`validateResolveInputs` already guards `cache == nil` defence-in-depth at the resolver call site, but the constructor itself should fail at the point of misuse so the diagnostic is actionable.

```suggestion
// NewStorageBackedStore returns a CredentialStore that delegates to a
// storage.DCRCredentialStore for durable persistence and translates
// Resolution values into DCRCredentials at the boundary. The returned
// store is safe for concurrent use because the underlying
// storage.DCRCredentialStore must be (per its interface contract).
//
// Panics if backend is nil — a nil backend is unambiguously a programming
// error and silent acceptance would only delay the eventual nil-pointer
// dereference to the first Get/Put call, far from the constructor site.
func NewStorageBackedStore(backend storage.DCRCredentialStore) CredentialStore {
	if backend == nil {
		panic("dcr: NewStorageBackedStore: backend must not be nil")
	}
	return &storageBackedStore{backend: backend}
}
```

Raised by: api-design

pkg/auth/dcr/resolver.go
```diff
@@ -25,14 +67,14 @@ import (
...
-// same DCRKey. Two goroutines hitting the resolver for the same upstream and
+// dcrFlight coalesces concurrent ResolveCredentials calls that share the
+// same Key. Two goroutines hitting the resolver for the same upstream and
 // scope set will both miss the cache, so without coalescing they would both
 // call RegisterClientDynamically and the loser's registration would become
 // orphaned at the upstream IdP — an operator-visible cleanup task and
 // possibly a transient startup failure if the upstream rate-limits
 // concurrent registrations. Followers wait for the leader's result and
-// observe the same DCRResolution.
+// observe the same Resolution.
```
tgrunnagle, line 78: **`dcrFlight` doc names `EmbeddedAuthServer.dcrStore` field that no longer exists post-\#5196** (consensus 8/10, MEDIUM)

A few lines below this hunk (in the unchanged middle of the `dcrFlight` doc comment), the second paragraph reads:

> Lifetime: process-wide. This intentionally contrasts with `EmbeddedAuthServer.dcrStore`, which is per-instance.

Verified against `pkg/authserver/runner/embeddedauthserver.go:44-49`: the `EmbeddedAuthServer` struct has only `server`, `keyProvider`, `closeOnce`, `closeErr`. \#5196 removed the `dcrStore` field. The runtime topology now is "the `dcr.CredentialStore` wrapping the authserver's shared `storage.DCRCredentialStore`" — per-instance for the memory backend, shared across replicas for Redis.

Reframe the lifetime contrast in terms of the public injection point (the `CredentialStore` instance the embedded authserver constructs) rather than naming a private field across package boundaries.

Raised by: documentation

tgrunnagle (commented):

**Agents consulted**: api-design, concurrency/refactor, security, test-quality, documentation, cross-cutting (codex skipped — CLI not installed)

### Consensus Summary

| # | Finding | Score | Severity | Action |
|---|---------|-------|----------|--------|
| 1 | `NewInMemoryStore` returns store whose cleanup goroutine cannot be stopped | 9/10 | MEDIUM | Fix |
| 2 | `pkg/authserver/storage/types.go` "Converter contract" doc points at deleted files/types | 9/10 | MEDIUM | Fix (out-of-diff) |
| 3 | `dcrFlight` doc names `EmbeddedAuthServer.dcrStore` field that no longer exists post-\#5196 | 8/10 | MEDIUM | Fix |
| 4 | Drift-guard cross-reference is one-way — runner-package twin lacks pointer back | 8/10 | MEDIUM | Fix |
| 5 | `NewStorageBackedStore` accepts `nil` backend silently; panics on first call | 7/10 | MEDIUM | Fix |
| 6 | `SanitizeErrorForLog` exported as a generic-looking URL sanitiser, only handles http(s) | 8/10 | LOW | Discuss |
| 7 | Stale path `pkg/authserver/runner/dcr.go` in `handlers/dcr.go:109` comment | 7/10 | LOW | Fix (out-of-diff) |
| 8 | Package doc lists `*authserver.DCRUpstreamConfig` as parameter type; no exported function takes it | 7/10 | LOW | Fix |
| 9 | `flightKeyOf` doc cites RFC 6749 §3.3 (scope tokens) for the `ScopesHash` hex-digest field | 7/10 | LOW | Fix |
| 10 | Stale "DCRKey" mentions in `embeddedauthserver_test.go` comments and assertion messages | 7/10 | LOW | Fix |
| 11 | Singleflight global state will become a cross-consumer hazard once 4b lands | 7/10 | LOW | Discuss |
| 12 | `dcr.resolveSecret` has zero production callers; `#nosec G304` staged ahead of caller | 7/10 | LOW | Discuss |

### Overall

This is a clean refactor that extracts the DCR resolver from `pkg/authserver/runner` into a new `pkg/auth/dcr` package, with public-symbol renames on packagisation. All reviewable behaviours — S256 PKCE gate, redirect refusal, RFC 7591 §3.2.1 expiry refetch, panic recovery, package-global singleflight semantics, URL log sanitisation — are preserved across the move and verified case-by-case by the specialist agents. The only genuinely new code in the diff is `flightKeyOf` (extraction of an inlined string concatenation, byte-identical output) and a defensive nil-guard in `endpointsFromMetadata` (catches a hypothetical `oauthproto` contract regression). Six specialist agents reviewed in parallel; no specialist dismissed another's concern in a way that triggered conflict resolution.

The 12 surviving consensus findings split between three buckets: (a) **out-of-diff documentation debt** — the rename sweep missed `pkg/authserver/storage/types.go` and `pkg/authserver/server/handlers/dcr.go`, both of which still name pre-rename paths or types; (b) **constructor-validation hardening** for `NewInMemoryStore` (resource leak — the cleanup goroutine cannot be stopped through the returned interface) and `NewStorageBackedStore` (silent `nil` acceptance); (c) **forward-looking concerns for sub-issue 4b** — the package-global singleflight, the duplicated `resolveSecret` with no production caller in this PR, and the `SanitizeErrorForLog` export of a function that reads as generic but only handles http(s).

None of the findings are blocking. The PR is in DRAFT status which is consistent with this state. Recommendation is **COMMENT** — the bucket-(a) and bucket-(b) items are easy to address in this PR (small, mechanical); bucket-(c) items are reasonable to defer to 4b with explicit tracking links.

### Out-of-diff documentation debt

Three items live outside the PR's diff and need to be pulled in (or filed as a follow-up explicitly linked from this PR):

- `pkg/authserver/storage/types.go:235-243` — `DCRCredentials` "Converter contract" doc names `pkg/authserver/runner/dcr_store.go`, `*DCRResolution`, and `pkg/authserver/runner/dcr_store_test.go`. After this PR, the converters live in `pkg/auth/dcr/store.go` and the test in `pkg/auth/dcr/store_test.go`. Update those three references to the new paths and to `*dcr.Resolution`. **(Finding \#2, MEDIUM)**
- `pkg/authserver/server/handlers/dcr.go:109` — comment names `pkg/authserver/runner/dcr.go`. The resolver now lives at `pkg/auth/dcr/resolver.go`. **(Finding \#7, LOW)**
- `pkg/authserver/runner/embeddedauthserver_test.go:289-374` — `TestResolveSecret` and `TestResolveSecretWithEnvVar` should carry a doc comment naming the new `pkg/auth/dcr/secret_test.go` twin so a future bug fix on the runner side has a proximate hint that the dcr-side parallel test must be updated too. The dcr-side comment already names the runner-side twin; the back-pointer is missing. (Inline comment on the dcr-side at `secret_test.go:17` discusses this further.) **(Finding \#4, MEDIUM)**

### Items to keep in mind for sub-issue 4b

Three of the LOW findings are forward-looking and become more expensive once 4b lands the CLI flow:

- The package-global `singleflight.Group` will be shared across both consumers. Consider including a consumer identifier in the flight key now (cheap) or filing a tracking issue.
- `dcr.resolveSecret` has no production caller in this PR — either defer to 4b or extract into a shared helper (`pkg/auth/secretref` or similar) in this PR with both consumers migrated.
- `SanitizeErrorForLog` is exported as a generic-looking URL sanitiser but only handles http(s); a future caller wrapping a `redis://user:pass@…` error would silently leak credentials. Tighten the doc comment, move the helper to a small shared package, or keep it unexported behind a narrower `SanitizeDCRError` wrapper.

### What's good

All specialists confirmed the refactor preserves observable behaviour. The rename sweep is exhaustive within the 9 diffed files (the misses are in three external files). `flightKeyOf` produces byte-identical output to the inlined concatenation. Singleflight semantics, panic recovery, and `t.Cleanup` placements are correct. The new defensive nil-guard in `endpointsFromMetadata` is well-documented and turns a future contract regression into a clean error rather than a nil-deref panic. The drift-guard tests for `resolveSecret` are observably equivalent to the runner-package twin (case-by-case audit). 25+ NO-ISSUES verifications across the specialists are recorded in the local review.

---
*Generated with Claude Code*

pkg/auth/dcr/resolver.go
```diff
@@ -1,15 +1,57 @@
...
+// — the deduplication is a feature for the embedded authserver but means
+// callers cannot assume per-call-site flight isolation. See the dcrFlight
+// doc comment below for the rationale.
+//
+// # Current API coupling — sub-issue 4a only
+//
+// As of issue #5145 sub-issue 4a (the slice that lifted this code out of
+// pkg/authserver/runner), the public functions on this package take
+// embedded-authserver types — *authserver.OAuth2UpstreamRunConfig,
+// *authserver.DCRUpstreamConfig, *upstream.OAuth2Config — directly on
```
tgrunnagle: **Package doc lists `*authserver.DCRUpstreamConfig` as parameter type, but no exported function takes it** (consensus 7/10, LOW)

Auditing every exported function (`NeedsDCR`, `ConsumeResolution`, `ApplyResolutionToOAuth2Config`, `ResolveCredentials`, `LogStepError`, `SanitizeErrorForLog`, `NewStorageBackedStore`, `NewInMemoryStore`), none takes `*authserver.DCRUpstreamConfig` as a parameter. It is reached only transitively via `OAuth2UpstreamRunConfig.DCRConfig`. The section's purpose is to give 4b an accurate snapshot of "what shapes the neutral input type must replace", so the imprecision matters.

```suggestion
// embedded-authserver types — *authserver.OAuth2UpstreamRunConfig and
// *upstream.OAuth2Config (with *authserver.DCRUpstreamConfig reached
// transitively via OAuth2UpstreamRunConfig.DCRConfig) — directly on
```

Raised by: documentation

pkg/auth/dcr/resolver.go
```diff
@@ -44,8 +86,30 @@ import (
...
+// The current call sites do not collide — the embedded authserver's
+// redirect URI lives on the AS origin, the CLI flow's lives on a
+// loopback — but a future consumer that defaults its redirect URI into
+// either of those spaces would silently coalesce. Keep this in mind when
+// adding a third consumer.
 var dcrFlight singleflight.Group
 
+// flightKeyOf canonicalises a Key into the singleflight string used by
+// dcrFlight. The "\n" separator is safe because newline is not a valid byte
+// in any of the three components: OAuth scope tokens (RFC 6749 §3.3), URI
```
tgrunnagle: **`flightKeyOf` doc cites RFC 6749 §3.3 (scope tokens) for the `ScopesHash` field** (consensus 7/10, LOW)

`Key.ScopesHash` is a hex digest produced by `storage.ScopesHash()` (per the type doc and `hex.EncodeToString` body). It is not a raw scope token. The hex-digest argument alone proves separator safety for that field; the RFC 6749 §3.3 citation is misaligned here — it belongs on `storage.ScopesHash` itself, not on this helper. Three components, three character classes is enough; four citations confuses the chain.

```suggestion
// in any of the three components: URI reference characters in Issuer and
// RedirectURI (RFC 3986 §2), and hex digits in ScopesHash (the form
```

Raised by: concurrency/refactor

pkg/auth/dcr/resolver.go
```diff
@@ -44,8 +86,30 @@ import (
...
+// Cross-consumer caveat (matters once issue #5145 sub-issue 4b lands the
+// CLI flow as the second consumer): because dcrFlight is package-global,
+// two consumers that happen to construct identical Keys (same issuer, same
+// redirect URI, same scopes hash) will share a single in-flight
+// registration even if they semantically want different client profiles.
+// The current call sites do not collide — the embedded authserver's
+// redirect URI lives on the AS origin, the CLI flow's lives on a
+// loopback — but a future consumer that defaults its redirect URI into
+// either of those spaces would silently coalesce. Keep this in mind when
+// adding a third consumer.
```
tgrunnagle: **Singleflight global state becomes a cross-consumer hazard once 4b lands** (consensus 7/10, LOW — forward-looking)

The caveat above honestly describes the failure mode: two consumers that compute identical `Key`s share a single in-flight registration even when they semantically want different client profiles. Today's two call sites do not collide; sub-issue 4b is the inflection point. The cost of fixing now is one new field on the flight key (e.g., a consumer identifier); the cost of fixing after 4b ships is much higher because client profiles are already wired through.

Consider one of:
1. Including a consumer-identifier string in the flight key shape so cross-profile collisions cannot occur.
2. Making the singleflight injectable via a constructor option (defaulting to the package-global for the embedded authserver).
3. Filing a tracking issue so this is not silently inherited by 4b.

Raised by: api-design

pkg/auth/dcr/resolver.go
```diff
@@ -519,7 +583,7 @@ func logDCRStepError(upstreamName string, err error) {
 // host:port) are not sanitised; the current DCR flow never embeds
 // those in errors, and broadening the match risks false positives on
 // unrelated text.
-func sanitizeErrorForLog(err error) string {
+func SanitizeErrorForLog(err error) string {
```
tgrunnagle, line 614: **`SanitizeErrorForLog` exported as a generic-looking URL sanitiser, only handles http(s)** (consensus 8/10, LOW)

The function is exported because the embedded authserver's deferred-cleanup gate at `embeddedauthserver.go:110-111` calls it on potentially-non-DCR `closeErr`/`retErr`. The body's `queryStrippingPattern` regex `(?i)https?://[^\s"']+` only matches http(s); a future caller wrapping an error containing `redis://user:pass@host/0` or `postgres://user:pass@host/db` would silently leak credentials in logs.

Three options, in order of safety:
1. Move `SanitizeErrorForLog` to a small shared helper package (`pkg/oauthproto/internal/sanitize` or `pkg/secretref`); both `dcr` and `authserver/runner` consume it. The helper has nothing DCR-specific in it.
2. Keep it unexported and expose a narrower `SanitizeDCRError(err) string` wrapper that the embedded authserver calls.
3. At minimum, tighten the doc comment to a usage warning: "Callers MUST verify that any non-http(s) URL schemes their errors might contain are not credential-bearing — this function only strips http/https URLs."

Raised by: api-design, security

pkg/auth/dcr/resolver.go
```diff
@@ -1197,3 +1275,33 @@ func newDCRHTTPClient(initialAccessToken string) *http.Client {
...
+// precedence over env var. Returns an error if file is specified but
+// unreadable, or if envVar is specified but not set. Returns empty string with
+// no error if neither file nor envVar is specified.
+//
+// This duplicates the logic in pkg/authserver/runner/embeddedauthserver.go
+// because the DCR package is profile-agnostic and must not reach back into
+// the runner — but the secret-resolution shape (file-or-env) is the same one
+// every consumer needs. Future consolidation can move this into a shared
+// pkg/secrets-style helper if a third caller appears.
+func resolveSecret(file, envVar string) (string, error) {
```
tgrunnagle, line 1323: **`dcr.resolveSecret` has zero production callers in this PR** (consensus 7/10, LOW)

This helper is byte-identical to `pkg/authserver/runner/embeddedauthserver.go::resolveSecret` (verified line-by-line). The only call site is `secret_test.go`. No production code in `pkg/auth/dcr` calls it — the embedded authserver still uses its own runner-side copy. The `#nosec G304` annotation on `os.ReadFile(file)` is approving a code path that has no production caller, so the file-path-trust justification cannot be validated against an actual call site.

Pick one:
1. Defer the duplication to sub-issue 4b, where the consumer (CLI flow) actually calls into it.
2. Promote `resolveSecret` to a small shared package (`pkg/auth/secretref` or similar) now and migrate both call sites in this PR — single copy, single test set, no drift hazard.

The doc comment proposes option 2 "if a third caller appears". With the second caller already lined up for 4b and the helper genuinely consumer-agnostic, the threshold is arguably already met.

Raised by: api-design, concurrency, security, test-quality

pkg/auth/dcr/secret_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestResolveSecret pins the dcr-package copy of resolveSecret to the same
+// observable contract as the parallel runner-package copy
+// (pkg/authserver/runner/embeddedauthserver_test.go::TestResolveSecret*).
```
tgrunnagle, line 17: **Drift-guard cross-reference is one-way** (consensus 8/10, MEDIUM)

This comment correctly names the runner-package twin, so a contributor editing the dcr-package copy has a proximate hint that a parallel test exists. The runner-package twin (`pkg/authserver/runner/embeddedauthserver_test.go::TestResolveSecret`, `:TestResolveSecretWithEnvVar`) has no such comment back. A bug fix on the runner-side `resolveSecret` could land without anyone noticing the dcr-side parallel test must also be updated; the dcr-side test would still pass against the unchanged dcr-side helper, and silent drift takes hold.

Add a reciprocal doc comment on the runner-side `TestResolveSecret` and `TestResolveSecretWithEnvVar` naming this dcr-side test path explicitly. The drift-guard pattern only works if both sides know about each other.

(This file is not in the diff hunks; the comment cannot be applied as a `suggestion` block from here — the back-pointer needs to be added at `pkg/authserver/runner/embeddedauthserver_test.go:289` and `:338`.)

Raised by: test-quality

pkg/authserver/runner/embeddedauthserver_test.go
```diff
@@ -1582,7 +1582,7 @@ func TestBuildUpstreamConfigs_DCR(t *testing.T) {
 
 		// Store now contains the resolution under the canonical DCRKey.
```
tgrunnagle: **Stale "DCRKey" mentions in comments and assertion messages** (consensus 7/10, LOW)

This comment and three nearby siblings (the assertion message at `:1592`, comments at `:1668`, `:1718`) read "the canonical DCRKey" without qualification. After this PR, `DCRKey` is no longer a runner-package symbol — the type literal at `:1585` was correctly migrated to `storage.DCRKey{...}`, but the surrounding prose still names "DCRKey" unqualified. Cosmetic only — the unqualified "DCRKey" still has a referent (`storage.DCRKey`) — but reads awkwardly post-rename and is the kind of comment-vs-code drift `.claude/rules/go-style.md` cautions against.

Either rephrase to `storage.DCRKey` for clarity, or leave a one-line note at the top of the test file that "DCRKey" in comments refers to `storage.DCRKey`.

Raised by: concurrency/refactor, cross-cutting

pkg/auth/dcr/store.go
tgrunnagle: Addressed in 23a231f5 by removing `NewInMemoryStore` (option 1). The constructor had no production caller in this PR and the leaked goroutine was unfixable through the `CredentialStore` interface alone. When a real consumer with lifecycle management appears in 4b, the right shape can be designed against an actual use case.

pkg/auth/dcr/store.go
tgrunnagle: Addressed in 23a231f5. Applied the suggested panic guard verbatim and updated the doc comment to explain the rationale.

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in 29c751a4. Reframed the lifetime contrast in terms of the `CredentialStore` the embedded authserver constructs and injects into `ResolveCredentials`, dropping the cross-package reference to the deleted `EmbeddedAuthServer.dcrStore` field.

pkg/auth/dcr/secret_test.go
tgrunnagle: Addressed in 29c751a4. Added reciprocal doc comments on `TestResolveSecret` and `TestResolveSecretWithEnvVar` in `pkg/authserver/runner/embeddedauthserver_test.go` naming the dcr-side twin. The drift-guard is now two-way.

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in 29c751a4 by tightening the doc (option 3). Added an explicit "IMPORTANT — caller responsibility" section warning that callers receiving non-http(s) URLs with credential-bearing components (`redis://user:pass@host`, `postgres://`, `smtp://`) MUST verify those separately. Long-term consolidation into a shared sanitiser is appropriate as a follow-up; doing it here would force naming the shared package without a second non-http(s) caller in hand.

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in 29c751a4. Applied the suggested wording verbatim.

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in 29c751a4. Applied the suggested wording verbatim.

pkg/authserver/runner/embeddedauthserver_test.go
tgrunnagle: Addressed in 29c751a4. Qualified "DCRKey" as "storage.DCRKey" in the four flagged sites (`:1583`, `:1592`, `:1668`, `:1718`). Also fixed a sibling staleness in the same test-doc block at `:1666-1669` where the comment named the deleted `dcrStore` field — same class of post-rename drift, surfaced while editing the surrounding lines.

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in 29c751a4 by option 3 — added a `TODO(#5219)` immediately on `dcrFlight` pointing at the sub-issue 4b PR as the resolution moment. The cross-consumer caveat above already describes the failure mode; this TODO ties it to where the design call will be made. A consumer-identifier component (option 1) is the likely landing place but is cheaper to design against the actual second consumer's profile shape.

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in 29c751a4 by tightening the doc to defer to 4b (option 1). The duplication-doc now states that the helper is staged for sub-issue 4b (#5219) to wire as part of the CLI flow migration, and that the threshold for consolidation into a shared helper is "second caller wires up" (i.e., when 4b lands), not "third caller appears". Promoting to a shared package now would force naming and locating the helper without the 4b CLI flow's actual call shape in hand.

tgrunnagle:
Addressing the two out-of-diff documentation-debt findings from the review body:

- Finding #2 — `pkg/authserver/storage/types.go` `DCRCredentials` Converter contract doc named the deleted `pkg/authserver/runner/dcr_store.go` path and `*DCRResolution` type. Fixed in 29c751a43cdbbb32d6530eda63e76d55ccbcee44 (paths updated to `pkg/auth/dcr/store.go` and `pkg/auth/dcr/store_test.go`; type updated to `*dcr.Resolution`).
- Finding #7 — `pkg/authserver/server/handlers/dcr.go:109` named the deleted `pkg/authserver/runner/dcr.go` path. Fixed in the same commit (updated to `pkg/auth/dcr/resolver.go`).

Both files were outside the original PR diff; pulling them in here rather than filing a follow-up since they are mechanical rename-sweep misses and small enough to live alongside the doc-sweep commit.

pkg/auth/dcr/resolver.go
```diff
@@ -189,14 +260,14 @@ func needsDCR(rc *authserver.OAuth2UpstreamRunConfig) bool {
...
+// Note on ClientSecret: ConsumeResolution does NOT write the resolved
 // secret to rc because OAuth2UpstreamRunConfig models secrets as file-or-
 // env references only. To propagate the DCR-resolved secret into the
 // final upstream.OAuth2Config, callers must pair this call with
-// applyResolutionToOAuth2Config once the config has been built. Keeping
+// ApplyResolutionToOAuth2Config once the config has been built. Keeping
 // the two helpers side-by-side localises the DCR-specific application
 // logic.
-func consumeResolution(rc *authserver.OAuth2UpstreamRunConfig, res *DCRResolution) {
+func ConsumeResolution(rc *authserver.OAuth2UpstreamRunConfig, res *Resolution) {
```
jhrozek: `ConsumeResolution` and `ApplyResolutionToOAuth2Config` mutate caller-supplied pointers and document the "pass a copy" invariant only in prose. The current single caller in `buildUpstreamConfigs` deep-copies correctly, so this PR is safe. But when 4b (#5219) adds `pkg/auth/discovery::PerformOAuthFlow` as the second consumer, it has to remember the same discipline, and the type system won't catch a mistake.

Cheapest moment to fix is now, while there's one call site. A value-in / value-out signature makes the contract compile-time enforced:

```go
func ConsumeResolution(rc authserver.OAuth2UpstreamRunConfig, res *Resolution) authserver.OAuth2UpstreamRunConfig
func ApplyResolutionToOAuth2Config(cfg upstream.OAuth2Config, res *Resolution) upstream.OAuth2Config
```

Fine to do in a follow-up; flagging here so it doesn't get lost when 4b lands.

jhrozek (dismissed):
Approving. Solid refactor; the move preserves the security defences and the tests pin them.

One non-blocking suggestion inline on `ConsumeResolution` about the caller-mutation pattern. Fine to address in a follow-up before 4b lands.

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in cf6e17c4 — fixing now rather than as a follow-up, matching your "cheapest moment is now" framing. Both `ConsumeResolution` and `ApplyResolutionToOAuth2Config` are value-in / value-out; the prose-discipline paragraph that required callers to pass a copy is gone. `ResolveCredentials` is intentionally left pointer-typed because its rc input is read-only and the pointer is just to avoid a hot-path struct copy. Satisfies the `.claude/rules/go-style.md` "Copy Before Mutating Caller Input" rule.

jhrozek (approved):

PR #5200, opened 2026-05-05
Add identity extractor for OAuth2 token responses

### LINKED ISSUE #5152: Add identity extractor for OAuth2 token responses
# Add identity extractor for OAuth2 token responses

## Description

Phase 1 of the Snowflake / `identityFromToken` story (#5150). Add a
pure helper that takes a raw OAuth2 token-endpoint response body and a
config struct and returns the extracted user identity (subject, name,
email), using gjson dot-notation paths into the JSON envelope. This is
the foundational building block consumed by the OAuth2 upstream
provider in a later phase — no operator surface, no runtime wiring,
no I/O.

## Context

`MCPExternalAuthConfig`'s OAuth2 upstream type currently requires
`userInfo`. Some providers (Snowflake, Slack v2, Shopify online tokens)
have no usable userinfo endpoint but include user identity directly in
the token-endpoint response body. The `identityFromToken` story (#5150)
adds a sibling extraction mechanism that pulls identity from configured
gjson paths into that response body.

Splitting the pure helper out keeps the surface deterministic and the
test matrix tractable. Key design points:

- Only scalar string and number values are accepted for the subject
  path. Objects, arrays, null, and missing paths are rejected — a
  subject that quietly resolves to a stringified JSON blob would be a
  silent footgun.
- Numeric subjects preserve integer precision beyond 2^53 (some
  providers issue 64-bit numeric IDs that float64 formatting would
  truncate).
- Optional name / email paths degrade silently to empty string on
  type mismatch, with a `slog.Warn` that names the path but never
  the value.
- Errors must never include any portion of the response body — the
  body may contain access or refresh tokens.

**Dependencies**: PR #5094 (already merged) — provides the
`ErrIdentityResolutionFailed` sentinel the helper wraps.
**Blocks**: the OAuth2 upstream provider integration cannot consume
identity from the token response until this helper lands.

## Acceptance Criteria

- [ ] A pure `extractIdentityFromTokenResponse(body, cfg)` helper
  exists in the upstream package, returning `(subject, name, email,
  err)` (concrete shape internal to the package).
- [ ] Subject path enforces scalar-only resolution; objects, arrays,
  null, booleans, and missing paths produce an error wrapping
  `ErrIdentityResolutionFailed`.
- [ ] Numeric subjects retain full integer precision (no float64
  round-tripping).
- [ ] Optional name and email paths handle missing or wrong-typed
  values gracefully, with WARN-level logs that name the path but
  never include the value or any body content.
- [ ] No part of the response body appears in any returned error or
  any logged record at any level — asserted by a dedicated test that
  embeds a unique marker in the body and checks it never surfaces.
- [ ] No I/O in the helper; pure function.
- [ ] Unit tests cover realistic Snowflake-flat, Slack-nested, and
  Shopify-nested response shapes plus the type-guard edge cases above.
- [ ] `task lint-fix` and `task test` pass.

## Out of Scope

- The CRD type and its admission validation.
- Wiring into the OAuth2 upstream provider's identity-resolution chain.
- Operator-side translation from CRD to runtime config.
- YAML examples.
- Audit-log changes — the helper just produces an identity value;
  what the auditor does with it is unchanged.

jhrozek:
## Summary

Some OAuth2 upstream providers do not expose a usable userinfo endpoint and instead place user identity directly in the token endpoint response. Two shapes appear in practice:

- **Identity as side-attributes alongside the tokens** — Snowflake's `username`, Slack's `authed_user.id`, Shopify's `associated_user.{id,email,first_name}`.
- **Identity claims embedded inside a JWT-shaped access token** — Auth0, Azure AD, Keycloak, Cognito.

This PR adds a pure helper `extractIdentityFromTokenResponse(body, cfg)` that reads operator-supplied gjson dot-notation paths from the raw token-response body to extract subject, name, and email. It also registers a gjson modifier `@upstreamjwt` so paths can pipe through a JWT payload decode step (e.g. `access_token|@upstreamjwt|sub`). The modifier does not verify the JWT signature — trust comes from the TLS-authenticated channel to the AS, the same trust model as the existing userinfo path. Signed-token flows remain handled by the existing OIDC provider type.

The helper is consumed by the embedded auth server's OAuth2 upstream provider in a later commit; nothing in this PR calls it yet.

Closes #5152

## Type of change

- [x] New feature / enhancement

## Test plan

- [x] Unit tests for happy paths (Snowflake-flat, Slack-nested, Shopify-nested)
- [x] Unit tests for subject type-guard rejection (object, null, missing path, empty body, malformed JSON, nil cfg, empty subject value)
- [x] Unit tests for optional name/email handling (path not configured, missing in body, warn-and-skip on object value)
- [x] Unit tests for JWT-embedded identity (happy path + malformed JWT)
- [x] Numeric subject precision test (`9007199254740993`, beyond 2^53)
- [x] Property-style test asserting error messages never contain any portion of the response body
- [x] `task lint-fix` reports 0 issues

## Implementation plan

<details>
<summary>Approved plan from the design session</summary>

- Pure helper `extractIdentityFromTokenResponse(body, cfg)` with type-guard subject validation (string or number only; objects, arrays, null, booleans, missing paths rejected with `ErrIdentityResolutionFailed`).
- Optional name/email paths warn-and-skip on type mismatch via `slog.Warn` that names the path but never the value or any body content.
- Numeric precision preserved via `result.Raw` rather than `gjson.Result.String()` (which formats via float64 and truncates >2^53).
- Errors carry the configured path name but never any portion of the body — pinned by a dedicated property test embedding a unique secret marker.
- gjson custom modifier `@upstreamjwt` for JWT-embedded identity claims. No signature verification (trust via TLS-to-AS, same as userinfo path). Returns `""` on any failure mode (non-string input, wrong dot-count, base64 error), letting downstream `validateIdentityField` produce a uniform `ErrIdentityResolutionFailed`.
- Modifier registration via explicit `RegisterModifiers()` rather than `init()` — caller-controlled, testable, no `gochecknoinits` suppression. Idempotent (gjson.AddModifier overwrites).
- Test surface kept minimal: 17 table cases covering distinct code branches plus one body-leakage property test.

</details>

No. This is an internal helper with no consumer in this PR. The OAuth2 upstream provider integration that exposes the feature to operators — and the corresponding CRD type — lands in follow-up PRs.

- **Beyond the linked issue's stated criteria**: issue #5152's acceptance criteria enumerate only the side-attribute shapes (Snowflake/Slack/Shopify). The `@upstreamjwt` modifier is an additive extension for the JWT-embedded shape (Auth0/Azure/Keycloak/Cognito). All eight original acceptance criteria from the issue are met; the JWT modifier is incremental on top.
- **Wire-up obligation for the integrating PR**: `RegisterModifiers()` must be called once during application or test wire-up before any consumer issues a path containing `@upstreamjwt`. The integrating commit will add this call to the OAuth2 upstream provider's constructor.
- **Downstream JWT serialization**: numeric-source subjects are returned as Go strings (`scalarToString` uses `result.Raw`). Any downstream JWT issuer must serialize them as JSON strings per RFC 7519 §4.1.2 (`StringOrURI`). A test for that contract belongs in the integrating PR.
- **Trust-model docstring**: the trust caveats (no signature verification, TLS-to-AS as integrity boundary) live on `upstreamJWTModifier`'s godoc. The corresponding CRD type (out of scope here) should also carry operator-facing warnings about which fields are appropriate for `subjectPath` (don't use `access_token`, `token_type`, etc.).

### DIFF
--- a/pkg/authserver/upstream/identity_from_token.go
+++ b/pkg/authserver/upstream/identity_from_token.go
@@ -1,0 +1,176 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package upstream
+
+import (
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"log/slog"
+	"strings"
+
+	"github.com/tidwall/gjson"
+)
+
+// partialIdentity holds identity fields extracted from a token response body.
+// It is used internally to pass the extracted subject, name, and email
+// between extractIdentityFromTokenResponse and the provider layer.
+type partialIdentity struct {
+	Subject string
+	Name    string
+	Email   string
+}
+
+// IdentityFromTokenConfig is the runtime configuration for extracting user
+// identity directly from an OAuth2 token endpoint response body.
+//
+// Each path is a gjson dot-notation path (e.g. "username" or
+// "associated_user.id") into the raw JSON body returned by the token
+// endpoint. Path semantics, trust-model warnings, and uniqueness notes are
+// documented on the corresponding CRD type
+// (cmd/thv-operator/api/v1alpha1.IdentityFromTokenConfig).
+type IdentityFromTokenConfig struct {
+	// SubjectPath is the gjson path to the unique user identifier (required).
+	SubjectPath string
+
+	// NamePath is the gjson path to the user's display name (optional).
+	// Leave empty to skip name extraction.
+	NamePath string
+
+	// EmailPath is the gjson path to the user's email address (optional).
+	// Leave empty to skip email extraction.
+	EmailPath string
+}
+
+// RegisterModifiers registers the gjson custom modifiers used by this
+// package's path-based identity extractors. Call once during application
+// or test wire-up before invoking any extractor that consumes a
+// modifier-bearing path. Repeated calls are safe — gjson.AddModifier
+// overwrites the existing entry.
+//
+// Modifiers registered:
+//   - @upstreamjwt: see upstreamJWTModifier.
+func RegisterModifiers() {
+	gjson.AddModifier("upstreamjwt", upstreamJWTModifier)
+}
+
+// extractIdentityFromTokenResponse extracts user identity fields from a raw
+// OAuth2 token endpoint response body using the paths in cfg.
+//
+// SubjectPath must resolve to a string or number value; objects, arrays, null,
+// and missing paths are rejected with ErrIdentityResolutionFailed. NamePath
+// and EmailPath are optional: type mismatches or missing paths produce a
+// slog.Warn and leave the respective field empty. Empty NamePath/EmailPath in
+// cfg means "do not extract" and are skipped silently.
+func extractIdentityFromTokenResponse(body []byte, cfg *IdentityFromTokenConfig) (partialIdentity, error) {
+	if cfg == nil {
+		return partialIdentity{}, errors.New("identity extraction config is required")
+	}
+
+	subjectResult := gjson.GetBytes(body, cfg.SubjectPath)
+	if err := validateIdentityField(subjectResult); err != nil {
+		return partialIdentity{}, fmt.Errorf("%w: subjectPath %q %s", ErrIdentityResolutionFailed, cfg.SubjectPath, err.Error())
+	}
+
+	name := extractOptionalField(body, cfg.NamePath, "namePath")
+	email := extractOptionalField(body, cfg.EmailPath, "emailPath")
+
+	return partialIdentity{
+		Subject: scalarToString(subjectResult),
+		Name:    name,
+		Email:   email,
+	}, nil
+}
+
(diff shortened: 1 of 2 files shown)


pkg/authserver/upstream/identity_from_token.go
```diff
@@ -0,0 +1,176 @@
...
+		slog.Warn("optional identity field has unexpected type, skipping", "field", fieldName, "path", path)
+		return ""
+	}
+	// Unreachable: all gjson.Type cases are handled above.
+	return ""
+}
+
+// upstreamJWTModifier is a gjson modifier that decodes the payload of a
+// JWT-shaped string value and returns it as JSON, enabling further gjson
+// path drilling (e.g. "access_token|@upstreamjwt|sub").
```
tgrunnagle, line 147: This is neat! Probably need to make sure it's documented clearly.

tgrunnagle (approved):
LGTM

PR #5201, opened 2026-05-06
Add Windows named-pipe support to the API listener

samuv:
## Summary

- toolhive-studio on Windows wants to talk to `thv serve` over a Windows named pipe (`\\.\pipe\thv-api`), the same way it uses a UNIX socket on macOS/Linux. The current `--socket` plumbing in [pkg/api/server.go](pkg/api/server.go) only handles AF_UNIX paths: `setupUnixSocket` runs `os.Stat`/`os.Remove`/`os.MkdirAll`/`os.Chmod` around `net.Listen("unix", ...)`. None of that applies to named pipes — they are a different kernel object — so passing a pipe path made `thv serve` fail before the listener even opened.
- A second silent break would have hit Windows: `Server.ListenURL()` writes the discovery-file URL, and [pkg/server/discovery/health.go](pkg/server/discovery/health.go) `HTTPClientForURL` only understood `http://` and `unix://`. Every `thv serve` startup runs `discovery.Discover` to detect a previous instance via that URL; without an `npipe://` scheme, two concurrent `thv` processes on Windows would not see each other and would silently overwrite the discovery file.
- Split the listener and the discovery dialer per platform with build tags, mirroring `pkg/container/docker/sdk/client_{unix,windows}.go`. On Windows, a `\\.\pipe\` prefix routes to `winio.ListenPipe` (skipping stat/remove/mkdir/chmod) on the listener side and to `winio.DialPipeContext` on the health-check side. AF_UNIX fallback paths still work but drop `os.Chmod` because POSIX modes are meaningless on Windows. `ListenURL` emits `npipe://<name>` and `discovery.HTTPClientForURL` parses it via a new `ParseNamedPipeURL` validator. Non-Windows builds reject pipe addresses up front rather than creating a literal-backslash file.
- `github.com/Microsoft/go-winio v0.6.2` is already a direct dependency (used by `pkg/container/docker/sdk/client_windows.go`), so no `go.mod` change is needed. The existing `--socket` flag works for both transports — the studio passes the right address per platform.
- Four independent commits — discovery production, listener production, discovery tests, listener tests — so each can be reviewed / reverted in isolation.

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — full repo green; new `TestParseNamedPipeURL`, `TestSocketURL_Unix`, `TestIsNamedPipeAddress`, and `TestCheckHealth_NamedPipe_Unsupported_OnNonWindows` all pass on macOS.
- [x] Linting (`task lint-fix` — 0 issues).
- [x] Cross-compile (`GOOS=windows go vet ./pkg/api/... ./pkg/server/discovery/...` and `GOOS=windows go test -c -o /dev/null` for both packages) — both clean, confirming `socket_windows.go`, `socket_windows_test.go`, `pipe_windows.go`, and `health_windows_test.go` build correctly.
- [x] Manual testing on Windows — pending; happy to defer to a reviewer with a Windows host. The Windows-only e2e in [pkg/server/discovery/health_windows_test.go](pkg/server/discovery/health_windows_test.go) does the full production round trip: `winio.ListenPipe` → `http.Server.Serve` → `CheckHealth("npipe://<name>", nonce)` asserts 204 with matching nonce.

## Changes

| File | Change |
|------|--------|
| `pkg/api/server.go` | Remove `setupUnixSocket` / `cleanupUnixSocket`; add cross-platform `namedPipePrefix` and `isNamedPipeAddress`; route `ListenURL` through `socketURL`; tighten `createListener` (label `"Windows named pipe"`; reject pipe addresses on non-Windows). Stdlib `runtime` aliased to `stdruntime` to avoid collision with the existing `pkg/container/runtime` import. |
| `pkg/api/socket_other.go` | New, `//go:build !windows`. Original POSIX `setupUnixSocket` / `cleanupUnixSocket`; `socketURL` always emits `unix://`. |
| `pkg/api/socket_windows.go` | New, `//go:build windows`. Pipe-aware `setupUnixSocket` (`winio.ListenPipe`, 64 KiB buffers, byte-stream mode), `cleanupUnixSocket` no-op for pipes, AF_UNIX fallback without chmod, `socketURL` emits `npipe://<name>`. |
| `pkg/server/discovery/health.go` | Add `npipe://` arm in `HTTPClientForURL` (transport `DialContext` calls `dialNamedPipe`); add `ParseNamedPipeURL` validator (rejects empty, wrong scheme, path separators, `..`); update doc comments. |
| `pkg/server/discovery/pipe_windows.go` | New, `//go:build windows`. `dialNamedPipe` via `winio.DialPipeContext`. |
| `pkg/server/discovery/pipe_other.go` | New, `//go:build !windows`. Stub returning `"named pipes are only supported on Windows"`. |
| `pkg/api/socket_other_test.go` | New, `//go:build !windows`. `TestSocketURL_Unix`, `TestIsNamedPipeAddress`. |
| `pkg/api/socket_windows_test.go` | New, `//go:build windows`. `TestSocketURL_Windows`, `TestSetupUnixSocket_NamedPipe` (real `winio.ListenPipe` + dial round trip), `TestCleanupUnixSocket_NamedPipe_NoOp`. |
| `pkg/server/discovery/health_test.go` | Add `TestParseNamedPipeURL` (valid + 6 rejection branches) and `TestCheckHealth_NamedPipe_Unsupported_OnNonWindows`. |
| `pkg/server/discovery/health_windows_test.go` | New, `//go:build windows`. `TestCheckHealth_NamedPipe_Success` (end-to-end listener + serve + check) and `TestCheckHealth_NamedPipe_NotFound`. |

Yes — on Windows, `thv serve --socket \\.\pipe\thv-api` now works, which unblocks toolhive-studio talking to `thv` over a named pipe. macOS/Linux behaviour is unchanged: passing a pipe-prefixed address there fails fast with `named pipe addresses are only supported on Windows`. The `--socket` flag itself is unchanged and accepts both AF_UNIX paths and `\\.\pipe\…` addresses depending on platform.

## Implementation plan

<details>

(paste the contents of `.cursor/plans/windows-named-pipe-api-listener_87ed9c7e.plan.md` here)

</details>

- **URL scheme**: `npipe://<name>` rather than the Docker-style `npipe:////./pipe/<name>` because we control both writer (`socketURL`) and reader (`HTTPClientForURL`), so the simpler form parses cleanly with `net/url` — no embedded backslashes. The Docker SDK code in [pkg/container/docker/sdk/client_windows.go](pkg/container/docker/sdk/client_windows.go) uses Docker's convention because Docker's client demands it; that constraint does not apply here.
- **MessageMode** is left at `false` (byte stream) explicitly in the `winio.PipeConfig` literal as a documentation cue — HTTP requires byte framing. The zero value is already `false`, so the assignment is redundant but informative.
- **Buffer sizes** (64 KiB in/out) match what Docker, containerd, and Podman use for similar same-host control-plane pipes; well above any single HTTP header chunk so we never block on the kernel-side ring buffer.
- **Commit ordering**: the discovery-side dialer lands first so the API listener's `npipe://` emission in commit 2 is immediately consumable by `discovery.Discover`. No intermediate commit leaves the codebase emitting URLs it cannot dial.

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	stdruntime "runtime"
 	"strings"
 	"time"
 
@@ -366,41 +367,20 @@ func (b *ServerBuilder) setupDefaultRoutes(r *chi.Mux) {
 	}
 }
 
+// namedPipePrefix is the Windows named-pipe namespace prefix. Both per-platform
+// socket files use it, and ListenURL/createListener inspect the address prefix
+// to choose between AF_UNIX and named-pipe semantics.
+const namedPipePrefix = `\\.\pipe\`
+
+// isNamedPipeAddress reports whether address is a Windows named-pipe path.
+// The check is platform-agnostic so callers on non-Windows can fail fast with
+// a clear error before reaching the listener code.
+func isNamedPipeAddress(address string) bool {
+	return strings.HasPrefix(address, namedPipePrefix)
+}
+
 func setupTCPListener(address string) (net.Listener, error) {
 	return net.Listen("tcp", address)
-}
-
-func setupUnixSocket(address string) (net.Listener, error) {
-	// Remove the socket file if it already exists
-	if _, err := os.Stat(address); err == nil {
-		if err := os.Remove(address); err != nil {
-			return nil, fmt.Errorf("failed to remove existing socket: %w", err)
-		}
-	}
-
-	// Create the directory for the socket file if it doesn't exist
-	if err := os.MkdirAll(filepath.Dir(address), 0750); err != nil {
-		return nil, fmt.Errorf("failed to create socket directory: %w", err)
-	}
-
-	// Create UNIX socket listener
-	listener, err := net.Listen("unix", address)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create UNIX socket listener: %w", err)
-	}
-
-	// Set file permissions on the socket to allow other local processes to connect
-	if err := os.Chmod(address, socketPermissions); err != nil {
-		return nil, fmt.Errorf("failed to set socket permissions: %w", err)
-	}
-
-	return listener, nil
-}
-
-func cleanupUnixSocket(address string) {
-	if err := os.Remove(address); err != nil && !os.IsNotExist(err) {
-		slog.Warn("failed to remove socket file", "error", err)
-	}
 }
 
 func headersMiddleware(next http.Handler) http.Handler {
@@ -592,7 +572,7 @@ func NewServer(ctx context.Context, builder *ServerBuilder) (*Server, error) {
 // bound address from the listener (important when binding to port 0).
 func (s *Server) ListenURL() string {
 	if s.isUnixSocket {
-		return fmt.Sprintf("unix://%s", s.address)
+		return socketURL(s.address)
 	}
 	return fmt.Sprintf("http://%s", s.listener.Addr().String())
 }
@@ -715,24 +695,30 @@ func (s *Server) cleanup() {
 	}
 }
 
-// createListener creates the appropriate listener based on the configuration
+// createListener creates the appropriate listener based on the configuration.
+// Named-pipe addresses are only supported on Windows; other platforms reject
+// them up front rather than creating a literal-backslash file via AF_UNIX.
 func createListener(address string, isUnixSocket bool) (net.Listener, string, error) {
-	var listener net.Listener
-	var addrType string
-	var err error
-
-	if isUnixSocket {
-		listener, err = setupUnixSocket(address)
-		addrType = "UNIX socket"
-	} else {
-		listener, err = setupTCPListener(address)
-		addrType = "HTTP"
+	if !isUnixSocket {
+		listener, err := setupTCPListener(address)
+		if err != nil {
+			return nil, "", err
+		}
+		return listener, "HTTP", nil
 	}
 
+	addrType := "UNIX socket"
+	if isNamedPipeAddress(address) {
+		if stdruntime.GOOS != "windows" {
+			return nil, "", fmt.Errorf("named pipe addresses are only supported on Windows: %s", address)
+		}
+		addrType = "Windows named pipe"
+	}
+
+	listener, err := setupUnixSocket(address)
 	if err != nil {
 		return nil, "", err
 	}
-
 	return listener, addrType, nil
 }
 
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -27,7 +27,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	stdruntime "runtime"
 	"strings"
 	"time"
 
@@ -367,16 +366,21 @@ func (b *ServerBuilder) setupDefaultRoutes(r *chi.Mux) {
 	}
 }
 
-// namedPipePrefix is the Windows named-pipe namespace prefix. Both per-platform
-// socket files use it, and ListenURL/createListener inspect the address prefix
-// to choose between AF_UNIX and named-pipe semantics.
-const namedPipePrefix = `\\.\pipe\`
+// namedPipePrefix is the Windows named-pipe namespace prefix. The canonical
+// definition lives in pkg/server/discovery so the listener and dialer cannot
+// drift; pkg/api re-aliases it here so per-platform socket files do not need
+// to import discovery directly.
+const namedPipePrefix = discovery.NamedPipePrefix
 
 // isNamedPipeAddress reports whether address is a Windows named-pipe path.
 // The check is platform-agnostic so callers on non-Windows can fail fast with
-// a clear error before reaching the listener code.
+// a clear error before reaching the listener code. The comparison is
+// case-insensitive because the Windows pipe namespace is case-insensitive at
+// the kernel layer; without EqualFold an address like \\.\Pipe\foo would
+// silently fall through to AF_UNIX and then fail to bind.
 func isNamedPipeAddress(address string) bool {
-	return strings.HasPrefix(address, namedPipePrefix)
+	return len(address) >= len(namedPipePrefix) &&
+		strings.EqualFold(address[:len(namedPipePrefix)], namedPipePrefix)
 }
 
 func setupTCPListener(address string) (net.Listener, error) {
@@ -709,7 +713,7 @@ func createListener(address string, isUnixSocket bool) (net.Listener, string, er
 
 	addrType := "UNIX socket"
 	if isNamedPipeAddress(address) {
-		if stdruntime.GOOS != "windows" {
+		if !supportsNamedPipe() {
 			return nil, "", fmt.Errorf("named pipe addresses are only supported on Windows: %s", address)
 		}
 		addrType = "Windows named pipe"
--- a/pkg/api/socket_other.go
+++ b/pkg/api/socket_other.go
@@ -1,0 +1,54 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !windows
+
+package api
+
+import (
+	"fmt"
+	"log/slog"
+	"net"
+	"os"
+	"path/filepath"
+)
+
+// setupUnixSocket creates a UNIX domain socket listener at the given path.
+// On non-Windows platforms named-pipe addresses are not supported; callers
+// guard against that in createListener.
+func setupUnixSocket(address string) (net.Listener, error) {
+	if _, err := os.Stat(address); err == nil {
+		if err := os.Remove(address); err != nil {
+			return nil, fmt.Errorf("failed to remove existing socket: %w", err)
+		}
+	}
+
+	if err := os.MkdirAll(filepath.Dir(address), 0750); err != nil {
+		return nil, fmt.Errorf("failed to create socket directory: %w", err)
+	}
+
+	listener, err := net.Listen("unix", address)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create UNIX socket listener: %w", err)
+	}
+
+	if err := os.Chmod(address, socketPermissions); err != nil {
+		return nil, fmt.Errorf("failed to set socket permissions: %w", err)
+	}
+
+	return listener, nil
+}
+
+// cleanupUnixSocket removes the socket file at address. Missing files are not
+// an error since cleanup may run after a partial startup.
+func cleanupUnixSocket(address string) {
+	if err := os.Remove(address); err != nil && !os.IsNotExist(err) {
+		slog.Warn("failed to remove socket file", "error", err)
+	}
+}
+
+// socketURL returns the URL form of a Unix-socket address for the discovery
+// file. Non-Windows platforms only ever produce unix:// URLs.
+func socketURL(address string) string {
+	return "unix://" + address
+}
--- a/pkg/api/socket_unix.go
+++ b/pkg/api/socket_unix.go
@@ -6,21 +6,26 @@
 package api
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"log/slog"
 	"net"
 	"os"
 	"path/filepath"
 )
 
+// supportsNamedPipe reports whether the current build target can host a
+// Windows named-pipe listener. Used by createListener to reject pipe addresses
+// before reaching the per-platform setupUnixSocket implementation.
+func supportsNamedPipe() bool { return false }
+
 // setupUnixSocket creates a UNIX domain socket listener at the given path.
 // On non-Windows platforms named-pipe addresses are not supported; callers
 // guard against that in createListener.
 func setupUnixSocket(address string) (net.Listener, error) {
-	if _, err := os.Stat(address); err == nil {
-		if err := os.Remove(address); err != nil {
-			return nil, fmt.Errorf("failed to remove existing socket: %w", err)
-		}
+	if err := os.Remove(address); err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return nil, fmt.Errorf("failed to remove existing socket: %w", err)
 	}
 
 	if err := os.MkdirAll(filepath.Dir(address), 0750); err != nil {
@@ -33,6 +38,11 @@ func setupUnixSocket(address string) (net.Listener, error) {
 	}
 
 	if err := os.Chmod(address, socketPermissions); err != nil {
+		// Roll back the bound listener and the socket file rather than leaking
+		// either: the listener owns the AF_UNIX file and net.Listen will not
+		// rebind the same path while the file exists.
+		_ = listener.Close()
+		_ = os.Remove(address)
 		return nil, fmt.Errorf("failed to set socket permissions: %w", err)
 	}
 
@@ -42,7 +52,7 @@ func setupUnixSocket(address string) (net.Listener, error) {
 // cleanupUnixSocket removes the socket file at address. Missing files are not
 // an error since cleanup may run after a partial startup.
 func cleanupUnixSocket(address string) {
-	if err := os.Remove(address); err != nil && !os.IsNotExist(err) {
+	if err := os.Remove(address); err != nil && !errors.Is(err, fs.ErrNotExist) {
 		slog.Warn("failed to remove socket file", "error", err)
 	}
 }
--- a/pkg/api/socket_unix_test.go
+++ b/pkg/api/socket_unix_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestSocketURL_Unix(t *testing.T) {
@@ -25,6 +26,7 @@ func TestIsNamedPipeAddress(t *testing.T) {
 	}{
 		{"plain socket", "/tmp/thv.sock", false},
 		{"named pipe", `\\.\pipe\thv-api`, true},
+		{"named pipe mixed case", `\\.\Pipe\thv-api`, true},
 		{"empty", "", false},
 	}
 	for _, tt := range tests {
@@ -33,4 +35,14 @@ func TestIsNamedPipeAddress(t *testing.T) {
 			assert.Equal(t, tt.want, isNamedPipeAddress(tt.address))
 		})
 	}
+}
+
+// TestCreateListener_NamedPipe_Unsupported asserts that createListener rejects
+// pipe addresses on non-Windows up front, mirroring the dialer-side guard
+// covered by TestCheckHealth_NamedPipe_Unsupported_OnNonWindows.
+func TestCreateListener_NamedPipe_Unsupported(t *testing.T) {
+	t.Parallel()
+	_, _, err := createListener(`\\.\pipe\thv-api`, true)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "only supported on Windows")
 }
--- a/pkg/api/socket_windows.go
+++ b/pkg/api/socket_windows.go
@@ -6,12 +6,13 @@
 package api
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"log/slog"
 	"net"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/Microsoft/go-winio"
 )
@@ -20,6 +21,11 @@ import (
 // per pipe instance. 64 KiB matches what go-winio uses in similar consumers
 // (Docker, containerd, Podman) and is well above any single HTTP header chunk.
 const namedPipeBufferSize = 64 * 1024
+
+// supportsNamedPipe reports whether the current build target can host a
+// Windows named-pipe listener. Used by createListener to choose between the
+// pipe and AF_UNIX paths without dragging the runtime package into server.go.
+func supportsNamedPipe() bool { return true }
 
 // setupUnixSocket creates either a Windows named-pipe listener (when address
 // has the \\.\pipe\ prefix) or an AF_UNIX listener at a filesystem path.
@@ -33,7 +39,7 @@ const namedPipeBufferSize = 64 * 1024
 // AF_UNIX is supported on Windows 10 1803+. The chmod step is dropped on this
 // path because POSIX file modes do not apply on Windows.
 func setupUnixSocket(address string) (net.Listener, error) {
-	if strings.HasPrefix(address, namedPipePrefix) {
+	if isNamedPipeAddress(address) {
 		// MessageMode is left at false (byte stream) explicitly because HTTP
 		// requires byte-oriented framing.
 		listener, err := winio.ListenPipe(address, &winio.PipeConfig{
@@ -47,10 +53,8 @@ func setupUnixSocket(address string) (net.Listener, error) {
 		return listener, nil
 	}
 
-	if _, err := os.Stat(address); err == nil {
-		if err := os.Remove(address); err != nil {
-			return nil, fmt.Errorf("failed to remove existing socket: %w", err)
-		}
+	if err := os.Remove(address); err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return nil, fmt.Errorf("failed to remove existing socket: %w", err)
 	}
 
 	if err := os.MkdirAll(filepath.Dir(address), 0750); err != nil {
@@ -68,10 +72,10 @@ func setupUnixSocket(address string) (net.Listener, error) {
 // cleanupUnixSocket removes the AF_UNIX socket file at address, or no-ops for
 // named pipes (the pipe is destroyed when the listener closes).
 func cleanupUnixSocket(address string) {
-	if strings.HasPrefix(address, namedPipePrefix) {
+	if isNamedPipeAddress(address) {
 		return
 	}
-	if err := os.Remove(address); err != nil && !os.IsNotExist(err) {
+	if err := os.Remove(address); err != nil && !errors.Is(err, fs.ErrNotExist) {
 		slog.Warn("failed to remove socket file", "error", err)
 	}
 }
@@ -80,8 +84,8 @@ func cleanupUnixSocket(address string) {
 // the discovery file. Named pipes are emitted as npipe://<name> where <name>
 // is everything after the \\.\pipe\ prefix.
 func socketURL(address string) string {
-	if strings.HasPrefix(address, namedPipePrefix) {
-		return "npipe://" + strings.TrimPrefix(address, namedPipePrefix)
+	if isNamedPipeAddress(address) {
+		return "npipe://" + address[len(namedPipePrefix):]
 	}
 	return "unix://" + address
 }
--- a/pkg/api/socket_windows.go
+++ b/pkg/api/socket_windows.go
@@ -1,0 +1,87 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build windows
+
+package api
+
+import (
+	"fmt"
+	"log/slog"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/Microsoft/go-winio"
+)
+
+// namedPipeBufferSize is the size of the input/output buffers winio allocates
+// per pipe instance. 64 KiB matches what go-winio uses in similar consumers
+// (Docker, containerd, Podman) and is well above any single HTTP header chunk.
+const namedPipeBufferSize = 64 * 1024
+
+// setupUnixSocket creates either a Windows named-pipe listener (when address
+// has the \\.\pipe\ prefix) or an AF_UNIX listener at a filesystem path.
+//
+// Named pipes are kernel objects rather than files, so the os.Stat / os.Remove
+// precheck, os.MkdirAll, and os.Chmod steps are skipped: the pipe namespace
+// has no parent directory, and access control is governed by the security
+// descriptor on the listener (winio's default restricts access to the
+// creating user, which matches the toolhive-studio same-user use case).
+//
+// AF_UNIX is supported on Windows 10 1803+. The chmod step is dropped on this
+// path because POSIX file modes do not apply on Windows.
+func setupUnixSocket(address string) (net.Listener, error) {
+	if strings.HasPrefix(address, namedPipePrefix) {
+		// MessageMode is left at false (byte stream) explicitly because HTTP
+		// requires byte-oriented framing.
+		listener, err := winio.ListenPipe(address, &winio.PipeConfig{
+			MessageMode:      false,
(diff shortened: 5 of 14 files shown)


pkg/api/socket_windows.go
```diff
@@ -0,0 +1,87 @@
...
+// descriptor on the listener (winio's default restricts access to the
+// creating user, which matches the toolhive-studio same-user use case).
+//
+// AF_UNIX is supported on Windows 10 1803+. The chmod step is dropped on this
+// path because POSIX file modes do not apply on Windows.
+func setupUnixSocket(address string) (net.Listener, error) {
+	if strings.HasPrefix(address, namedPipePrefix) {
+		// MessageMode is left at false (byte stream) explicitly because HTTP
+		// requires byte-oriented framing.
+		listener, err := winio.ListenPipe(address, &winio.PipeConfig{
```
aponcedeleonch, line 45: `PipeConfig.SecurityDescriptor` is empty here, so the pipe inherits the `CreateNamedPipeW` default DACL. The doc comment above (lines 28-31) claims winio "restricts access to the creating user," but that's not a hard guarantee. On a shared / RDP / Citrix host or any box with a second interactive account, another local user can `CreateFile(\\.\pipe\thv-api)` and exercise the entire REST API (OIDC is opt-in via `WithOIDCConfig`).

Set an explicit SDDL granting the current user only (e.g. `D:P(A;;GA;;;OW)` plus SYSTEM) and soften the comment to match what's actually enforced.

pkg/api/socket_windows.go
```diff
@@ -0,0 +1,87 @@
...
+		if err := os.Remove(address); err != nil {
+			return nil, fmt.Errorf("failed to remove existing socket: %w", err)
+		}
+	}
+
+	if err := os.MkdirAll(filepath.Dir(address), 0750); err != nil {
+		return nil, fmt.Errorf("failed to create socket directory: %w", err)
+	}
+
+	listener, err := net.Listen("unix", address)
```
aponcedeleonch, line 64: The AF_UNIX fallback drops the previous `os.Chmod(0660)` (correctly, since POSIX modes are meaningless on Windows), but nothing replaces it. The `.sock` file inherits the parent dir's ACL, and the `0750` mode passed to `MkdirAll` on line 56 is also advisory on NTFS.

Either document this path is best-effort and the named-pipe path is preferred, or set an explicit DACL via `windows.SetNamedSecurityInfo` after `net.Listen`.

pkg/server/discovery/health.go
```diff
@@ -144,3 +164,25 @@ func ParseUnixSocketPath(rawURL string) (string, error) {
 
 	return path, nil
 }
+
+// ParseNamedPipeURL extracts and validates the pipe name from an npipe:// URL
+// and returns the full Windows pipe path (e.g. \\.\pipe\thv-api). The name
+// portion must be a single segment with no path separators or traversal
+// components, since the toolhive listener only ever publishes local pipes
+// under the \\.\pipe\ namespace.
+func ParseNamedPipeURL(rawURL string) (string, error) {
```
aponcedeleonch, line 175: Replace this prefix-and-contains pipeline with `net/url.Parse`. The current approach has two problems: it misses structural cases that `url.Parse` rejects for free (backslashes, whitespace, NUL and control chars, query/fragment, userinfo, port suffixes), and `strings.Contains(name, "..")` rejects legitimate names like `npipe://my..api`.

Shape:
```go
u, err := url.Parse(rawURL)
if err != nil { ... }
if u.Scheme != "npipe" { ... }
if u.Path != "" || u.Opaque != "" || u.RawQuery != "" ||
    u.Fragment != "" || u.User != nil || u.Port() != "" { ... }
name := strings.ToLower(u.Hostname())
```

Then add explicit checks on top of the parser for: length cap (Windows: 247 chars after `\\.\pipe\`), reserved Windows names (`CON`, `NUL`, `PRN`, `AUX`, `COM1-9`, `LPT1-9`), and a positive charset like `^[A-Za-z0-9._-]+$`. Case normalization via `strings.ToLower(u.Hostname())` matters because pipes are case-insensitive at the kernel layer.

pkg/server/discovery/health.go
```diff
@@ -95,6 +101,20 @@ func HTTPClientForURL(serverURL string) (*http.Client, string, error) {
 		}
 		return client, "http://localhost", nil
 
+	case strings.HasPrefix(serverURL, "npipe://"):
```
aponcedeleonch, line 106: Move the scheme dispatcher from prefix matching to `url.Parse` plus `switch u.Scheme`. Go's `url.Parse` lowercases the scheme during parsing, so this fixes case-insensitive scheme matching as a side effect. The `http://` arm at line 118 already uses `url.Parse` via `ValidateLoopbackURL`, so the change tightens consistency in the same file:

```go
u, err := url.Parse(serverURL)
if err != nil { ... }
switch u.Scheme {
case "unix":
    // ...
case "npipe":
    // ...
case "http":
    // ...
default:
    return nil, "", fmt.Errorf("unsupported URL scheme: %s", u.Scheme)
}
```

pkg/api/socket_windows.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	if err := os.Remove(address); err != nil && !os.IsNotExist(err) {
+		slog.Warn("failed to remove socket file", "error", err)
+	}
+}
+
+// socketURL returns the URL form of a Unix-socket or named-pipe address for
+// the discovery file. Named pipes are emitted as npipe://<name> where <name>
+// is everything after the \\.\pipe\ prefix.
+func socketURL(address string) string {
```
aponcedeleonch, line 86: Migrate `socketURL` and `ParseUnixSocketPath` to `net/url` for both `unix://` and `npipe://`. The `npipe://` half is already implied by the parser and dispatcher comments. Doing the `unix://` half now is a small additional change in the same files, and it kills a real bug rather than leaving a comment behind.

Today, this function emits `unix://C:\path\thv.sock` for Windows AF_UNIX paths, which `net/url.Parse` rejects with `invalid port ":\\path\\thv.sock"`. The dispatcher only works because it uses `strings.TrimPrefix`. `TestSocketURL_Windows` even pins this malformed form. If the dispatcher moves to `url.Parse`, the Windows AF_UNIX path silently breaks unless `socketURL` also produces a parseable URL.

```go
// socket_unix.go (after rename)
func socketURL(address string) string {
    return (&url.URL{Scheme: "unix", Path: address}).String()
}

// socket_windows.go
func socketURL(address string) string {
    if isNamedPipeAddress(address) {
        name := address[len(namedPipePrefix):]
        return (&url.URL{Scheme: "npipe", Host: name}).String()
    }
    return (&url.URL{Scheme: "unix", Path: address}).String()
}
```

And `ParseUnixSocketPath` becomes the symmetric reader using `url.Parse` plus the existing `..` / absolute / `Clean` checks. Update `TestSocketURL_Windows` to assert the encoded form (`unix:///C:%5Cpath%5Cthv.sock`) so the test pins correct behavior rather than the current malformed string.

No on-disk migration concern: the discovery file is local per-user and overwritten on every `thv serve` startup. For typical POSIX paths like `/tmp/thv.sock`, `url.URL{Scheme: "unix", Path: address}.String()` produces `unix:///tmp/thv.sock`, identical to current output.

pkg/api/server.go
```diff
@@ -366,41 +367,20 @@ func (b *ServerBuilder) setupDefaultRoutes(r *chi.Mux) {
...
-	// Set file permissions on the socket to allow other local processes to connect
-	if err := os.Chmod(address, socketPermissions); err != nil {
-		return nil, fmt.Errorf("failed to set socket permissions: %w", err)
-	}
-
-	return listener, nil
+// isNamedPipeAddress reports whether address is a Windows named-pipe path.
+// The check is platform-agnostic so callers on non-Windows can fail fast with
+// a clear error before reaching the listener code.
+func isNamedPipeAddress(address string) bool {
```
aponcedeleonch, line 381: `strings.HasPrefix` here is byte-exact. `\\.\Pipe\foo` silently falls through to AF_UNIX, which then fails to bind. The standard library doesn't have a clean cross-platform "is named-pipe path" predicate (`filepath.VolumeName` is platform-conditional and returns empty on non-Windows), so the practical fix is `strings.EqualFold` on a prefix slice:

```suggestion
func isNamedPipeAddress(address string) bool {
	return len(address) >= len(namedPipePrefix) &&
		strings.EqualFold(address[:len(namedPipePrefix)], namedPipePrefix)
}
```

Then route the inline `strings.HasPrefix(address, namedPipePrefix)` checks in `pkg/api/socket_windows.go` (lines 36, 71, 83) through `isNamedPipeAddress` instead of duplicating the byte-exact check.

aponcedeleonch (commented):
Review of the Windows named-pipe support. Most comments target `net/url` adoption (parser, dispatcher, and producer side) so the npipe scheme is parsed and emitted via the same standard-library tool. Two items below don't have a target line in this PR's diff and live here in the summary instead:

- The `--socket` flag help text in `cmd/thv/app/server.go` still says "UNIX socket path." On Windows it now also accepts `\\.\pipe\<name>`. Update and run `task docs`.
- Discovery file trust boundary on Windows. `xdg.StateHome` lands under `%LOCALAPPDATA%`, which is per-user by default, but the POSIX modes passed to `MkdirAll` in `pkg/server/discovery/discovery.go` are advisory on Windows. If a local attacker can write `npipe://attacker-pipe` into `server.json`, health checks dial the attacker's pipe over plaintext HTTP. Worth setting an explicit DACL on the discovery dir at create time via `windows.SetNamedSecurityInfo`.

Nothing blocking on functionality. The security and `net/url` items are the most important; the rest are tightening.

pkg/api/socket_unix.go
```diff
@@ -0,0 +1,54 @@
...
+	if err := os.MkdirAll(filepath.Dir(address), 0750); err != nil {
+		return nil, fmt.Errorf("failed to create socket directory: %w", err)
+	}
+
+	listener, err := net.Listen("unix", address)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create UNIX socket listener: %w", err)
+	}
+
+	if err := os.Chmod(address, socketPermissions); err != nil {
```
aponcedeleonch, line 40: AF_UNIX listener leak on `Chmod` failure. If `net.Listen("unix", ...)` succeeds but this `os.Chmod` fails, the function returns the error without closing the listener or removing the socket file. The listener is bound, the file exists, no one can close it.

Pre-existing on the POSIX side, but the refactor touched this code so it's fair to flag. Fix:

```go
if err := os.Chmod(address, socketPermissions); err != nil {
    _ = listener.Close()
    _ = os.Remove(address)
    return nil, fmt.Errorf("failed to set socket permissions: %w", err)
}
```

Same defer-on-error pattern should land if a Windows DACL step gets added.

pkg/api/socket_other.go
```diff
@@ -0,0 +1,54 @@
...
+	"net"
+	"os"
+	"path/filepath"
+)
+
+// setupUnixSocket creates a UNIX domain socket listener at the given path.
+// On non-Windows platforms named-pipe addresses are not supported; callers
+// guard against that in createListener.
+func setupUnixSocket(address string) (net.Listener, error) {
+	if _, err := os.Stat(address); err == nil {
```
aponcedeleonch: `os.Stat` then `os.Remove` widens the TOCTOU window for no benefit. Replace with a single `os.Remove(address)` that ignores `fs.ErrNotExist`, matching what `cleanupUnixSocket` already does:

```suggestion
	if err := os.Remove(address); err != nil && !errors.Is(err, fs.ErrNotExist) {
		return nil, fmt.Errorf("failed to remove existing socket: %w", err)
	}
```

Same change applies to `pkg/api/socket_windows.go:50-54`. (Remember to add `errors` and `io/fs` imports.)

pkg/api/server.go
```diff
@@ -366,41 +367,20 @@ func (b *ServerBuilder) setupDefaultRoutes(r *chi.Mux) {
...
-
-	// Create UNIX socket listener
-	listener, err := net.Listen("unix", address)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create UNIX socket listener: %w", err)
-	}
+// namedPipePrefix is the Windows named-pipe namespace prefix. Both per-platform
+// socket files use it, and ListenURL/createListener inspect the address prefix
+// to choose between AF_UNIX and named-pipe semantics.
+const namedPipePrefix = `\\.\pipe\`
```
aponcedeleonch: Duplicate constant: `namedPipePrefix` is also defined in `pkg/server/discovery/health.go:28`. Same value, two sources of truth. If only one updates, listener and dialer drift.

Promote to a shared leaf-package constant. One option: export from `discovery` (the lower-level dependency) and import into `pkg/api`.

pkg/api/server.go
```diff
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	stdruntime "runtime"
```
aponcedeleonch: `stdruntime` is used exactly once in this file (line 712). A small `supportsNamedPipe() bool` helper in the per-platform files lets `server.go` drop the runtime import entirely and sidesteps the package-name collision with `pkg/container/runtime`:

```go
// socket_unix.go
func supportsNamedPipe() bool { return false }

// socket_windows.go
func supportsNamedPipe() bool { return true }
```

Then `createListener` becomes `if !supportsNamedPipe() { return ... "only supported on Windows" ... }`.

pkg/api/socket_unix.go
```diff
@@ -0,0 +1,54 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !windows
+
+package api
```
aponcedeleonch, line 6: Convention nit: rename this file (and `socket_other_test.go`) to `socket_unix.go` / `socket_unix_test.go` to match the existing `client_unix.go` / `client_windows.go` convention in `pkg/container/docker/sdk/`. Functionally identical, but the project already picked a convention and the rest of the PR mirrors that pattern.

pkg/api/socket_windows.go
```diff
@@ -0,0 +1,87 @@
...
+//
+// AF_UNIX is supported on Windows 10 1803+. The chmod step is dropped on this
+// path because POSIX file modes do not apply on Windows.
+func setupUnixSocket(address string) (net.Listener, error) {
+	if strings.HasPrefix(address, namedPipePrefix) {
+		// MessageMode is left at false (byte stream) explicitly because HTTP
+		// requires byte-oriented framing.
+		listener, err := winio.ListenPipe(address, &winio.PipeConfig{
+			MessageMode:      false,
+			InputBufferSize:  namedPipeBufferSize,
```
aponcedeleonch, line 47: winio's default `MaxInstances` is 255, which is a hard concurrency ceiling unlike POSIX. Probably fine for desktop traffic, but bursty parallel fetches could surface as `ERROR_PIPE_LISTENING` / `ERROR_NO_DATA`.

The `http.Server` also has `ReadHeaderTimeout` set but no `IdleTimeout` (see `pkg/api/server.go`). Add `Server.IdleTimeout` and a connection cap as defense in depth.

pkg/api/socket_unix_test.go
```diff
@@ -0,0 +1,36 @@
...
+		{"named pipe", `\\.\pipe\thv-api`, true},
+		{"empty", "", false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tt.want, isNamedPipeAddress(tt.address))
+		})
+	}
+}
```
aponcedeleonch, line 38: Missing test: `createListener` on a non-Windows build should reject named-pipe addresses up front. The existing `TestCheckHealth_NamedPipe_Unsupported_OnNonWindows` covers the dialer side; this rounds out the listener side.

```go
func TestCreateListener_NamedPipe_Unsupported(t *testing.T) {
    t.Parallel()
    _, _, err := createListener(`\\.\pipe\thv-api`, true)
    require.Error(t, err)
    assert.Contains(t, err.Error(), "only supported on Windows")
}
```

pkg/server/discovery/health_test.go
```diff
@@ -44,6 +45,49 @@ func TestParseUnixSocketPath_Empty(t *testing.T) {
...
+}
+
+func TestCheckHealth_NamedPipe_Unsupported_OnNonWindows(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("non-Windows guard test")
+	}
+	t.Parallel()
+	err := CheckHealth(context.Background(), "npipe://thv-api", "")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "Windows")
```
aponcedeleonch, line 88: Missing round-trip test: `WriteServerInfo` followed by `ReadServerInfo` should preserve `npipe://<name>` URLs through the discovery file. The pieces are individually tested; the seam isn't.

Worth adding alongside the existing tests so the producer/consumer contract for named-pipe URLs is pinned end to end.

pkg/api/socket_windows_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func TestCleanupUnixSocket_NamedPipe_NoOp(t *testing.T) {
+	t.Parallel()
+	// Passing a pipe address to cleanup must not error or panic. There is no
+	// file to remove; the assertion here is simply that the call returns
+	// cleanly.
+	cleanupUnixSocket(`\\.\pipe\thv-api-cleanup-noop`)
+}
```
aponcedeleonch, line 89: Missing test: collision behavior when two `winio.ListenPipe` calls target the same name. winio sets `FILE_FLAG_FIRST_PIPE_INSTANCE`, so first-wins is the expected semantic and the second call should fail with `ERROR_ACCESS_DENIED`. Worth pinning the behavior down so a future winio bump can't silently change it.

pkg/server/discovery/health_windows_test.go
```diff
@@ -0,0 +1,53 @@
...
+	err = CheckHealth(context.Background(), "npipe://"+pipeName, expectedNonce)
+	require.NoError(t, err)
+}
+
+func TestCheckHealth_NamedPipe_NotFound(t *testing.T) {
+	t.Parallel()
+	err := CheckHealth(context.Background(), "npipe://nonexistent-pipe-thv-test", "")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "health check failed")
+}
```
aponcedeleonch, line 53: Missing test: "pipe exists, server hung, context cancels." The current tests cover success and not-found, but not the case where `winio.DialPipeContext` blocks on a dead/orphaned listener and the caller's context expires. Relevant to the discovery `StateUnhealthy` path: a hung peer shouldn't wedge `CheckHealth`.

samuv:
Thanks @aponcedeleonch for the thorough review. Disposition below; the fixes that fit the size envelope landed here, the rest are split into focused follow-ups (stacked on this branch so they'll auto-rebase onto `main` on merge).

### Landed in this PR (commit `785ddc34`)

- Updated `--socket` flag help text to mention named pipes; ran `task docs`.
- Promoted `namedPipePrefix` to `discovery.NamedPipePrefix` as the canonical definition; `pkg/api` re-aliases it locally so listener and dialer cannot drift. ([#3201085425](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085425))
- Made `isNamedPipeAddress` case-insensitive via `strings.EqualFold`. ([#3201085392](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085392))
- Collapsed `os.Stat` -> `os.Remove` into a single `os.Remove` that tolerates `fs.ErrNotExist` on both POSIX and the Windows AF_UNIX fallback. ([#3201085418](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085418))
- Closed the AF_UNIX listener and removed the socket file when `os.Chmod` fails. ([#3201085412](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085412))
- Replaced the one-off `stdruntime.GOOS` check with a per-platform `supportsNamedPipe()` helper, dropping the runtime-package alias and its collision with `pkg/container/runtime`. ([#3201085428](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085428))
- Renamed `socket_other.{go,_test.go}` -> `socket_unix.{go,_test.go}` to match the `client_unix`/`client_windows` convention. ([#3201085430](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085430))
- Added `TestCreateListener_NamedPipe_Unsupported` to round out the listener-side reject coverage on non-Windows. ([#3201085438](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085438))

### Split into stacked follow-ups

- **#5214 - Restrict Windows named-pipe DACL to creating user.** Sets explicit SDDL `D:P(A;;GA;;;OW)(A;;GA;;;SY)` on the pipe (current user + SYSTEM only, no inheritance), softens the related comments, marks the AF_UNIX fallback path as best-effort since NTFS inheritance still applies, and adds `http.Server.IdleTimeout` cross-platform as defense in depth around `winio.MaxInstances`. Addresses [#3201085355](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085355), [#3201085359](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085359), and [#3201085433](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085433).
- **#5215 - Move pipe and socket URL handling to net/url.** Migrates `socketURL`, `ParseUnixSocketPath`, `ParseNamedPipeURL`, and the `HTTPClientForURL` dispatcher to `net/url.Parse`. Includes a real bug fix for the unparseable Windows AF_UNIX URL form (`unix://C:\path\thv.sock` -> `unix:///C:%5Cpath%5Cthv.sock`), plus the length cap, reserved-name rejection, positive charset, and case normalisation you suggested. Addresses [#3201085366](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085366), [#3201085375](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085375), and [#3201085383](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085383).
- **#5216 - Pin npipe round-trip and pipe lifecycle invariants.** Adds the `WriteServerInfo` -> `ReadServerInfo` round-trip test for `npipe://` URLs, the `FILE_FLAG_FIRST_PIPE_INSTANCE` first-wins test, and the hung-peer-context-cancels test for `CheckHealth`. Addresses [#3201085442](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085442), [#3201085446](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085446), and [#3201085449](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085449).

### Filed as separate issue

- **#5217 - Set explicit DACL on Windows discovery directory under `%LOCALAPPDATA%`.** The pipe-side trust-boundary concern from your PR-summary item is being addressed in #5214; the discovery-directory side is pre-existing in `pkg/server/discovery/discovery.go` and worth tracking on its own since the fix needs Windows-only `windows.SetNamedSecurityInfo` plumbing.

Happy to fold any of the stacked PRs back in here if you'd prefer one batch over four staged reviews.

pkg/api/socket_windows.go
samuv: Addressed in #5214. The pipe is now created with `SecurityDescriptor: "D:P(A;;GA;;;OW)(A;;GA;;;SY)"` — Protected DACL granting `GenericAll` to the creating user (`OW`) and `SYSTEM` (`SY`) only, no inheritance from any container. Softened the doc comment block above to describe what the listener actually enforces rather than relying on winio defaults.

Went with current-user-only by default; happy to add `BUILTIN\Administrators` (`BA`) if we want admins to be able to attach for diagnostics.

pkg/api/socket_windows.go
samuv: Took the documentation path in #5214 — the doc comment on `setupUnixSocket` now flags this fallback as best-effort and points callers at the named-pipe path when listener access control matters. Skipped the `windows.SetNamedSecurityInfo` plumbing for now because the AF_UNIX path on Windows is mostly theoretical (toolhive-studio dials `\\.\pipe\` directly). Happy to upgrade to an explicit DACL here too if we want belt-and-braces.

pkg/server/discovery/health.go
samuv: Done in #5215. The dispatcher now does `u, err := url.Parse(serverURL)` then `switch u.Scheme` exactly as you proposed, so `NPIPE://x` and `UNIX:///x` route through the same arms as their lowercase forms. Pinned by `TestHTTPClientForURL_SchemeDispatchCaseInsensitive`.

pkg/server/discovery/health.go
samuv: Done in #5215. Migrated to `url.Parse` exactly along the shape you sketched, plus the additional checks you called out:

- 247-char length cap (`maxNamedPipeNameLen`)
- Reserved-name set (`CON`, `NUL`, `PRN`, `AUX`, `COM1`-`COM9`, `LPT1`-`LPT9`)
- Positive charset `^[A-Za-z0-9._-]+$`
- `strings.ToLower(u.Hostname())` for case normalization
- Strict URL shape: rejects non-empty path, opaque, query, fragment, userinfo, or port

Dropped the over-eager `strings.Contains(name, "..")` guard you flagged — `npipe://my..api` is accepted now, pinned by the `valid_with_dot_dot_in_name` test case.

pkg/api/socket_windows.go
samuv: Done in #5215, including the bug fix you spotted. `socketURL` on both platforms now goes through `(&url.URL{Scheme: ..., Path/Host: ...}).String()`, so the Windows AF_UNIX path emits `unix:///C:%5Cpath%5Cthv.sock` and round-trips cleanly. `ParseUnixSocketPath` is the symmetric reader: rejects non-empty authority/query/fragment/userinfo, then strips the synthetic leading slash that `url.URL` inserts in front of a drive letter so `C:\path\thv.sock` comes back out unchanged.

`TestSocketURL_Windows` now pins the encoded form, and I added `TestSocketURL_RoundTrip_Unix` / `TestSocketURL_RoundTrip_NamedPipe` so `socketURL` -> `Parse...` is checked end to end.

pkg/api/socket_unix.go
samuv: Done in 785ddc3 on the renamed `socket_unix.go`. Chmod failure now closes the listener and removes the socket file before returning the error, so we don't leak a bound AF_UNIX listener that nothing owns.

pkg/api/socket_other.go
samuv: Done in 785ddc3. Both `socket_unix.go` and `socket_windows.go` now use a single `os.Remove(address)` that tolerates `fs.ErrNotExist` via `errors.Is`, matching what `cleanupUnixSocket` already does. Added the `errors` and `io/fs` imports as you noted.

pkg/api/server.go
samuv: Done in 785ddc3. `isNamedPipeAddress` now does `len(address) >= len(namedPipePrefix) && strings.EqualFold(address[:len(namedPipePrefix)], namedPipePrefix)`. The inline `strings.HasPrefix(address, namedPipePrefix)` calls in `socket_windows.go` (lines 36, 71, 83) are now routed through `isNamedPipeAddress` so the byte-exact comparison can't sneak back in there, and `socketURL` slices off the prefix by length rather than `TrimPrefix`. New `TestIsNamedPipeAddress/named_pipe_mixed_case` covers `\\.\Pipe\thv-api`.

pkg/api/server.go
samuv: Done in 785ddc3. Replaced with a per-platform `supportsNamedPipe() bool` helper (returns `true` in `socket_windows.go`, `false` in `socket_unix.go`). `createListener` now reads `if !supportsNamedPipe() { return ... }`, the `stdruntime` alias is gone, and the `runtime` / `pkg/container/runtime` package-name collision goes away with it.

pkg/api/server.go
samuv: Done in 785ddc3. Promoted to `discovery.NamedPipePrefix` (exported from the lower-level `pkg/server/discovery`) as the canonical definition. `pkg/api/server.go` keeps a local `const namedPipePrefix = discovery.NamedPipePrefix` alias so per-platform socket files don't need a discovery import; the listener and dialer can no longer drift.

pkg/api/socket_unix.go
samuv: Done in 785ddc3. Renamed via `git mv` so history is preserved: `socket_other.go` -> `socket_unix.go` and `socket_other_test.go` -> `socket_unix_test.go`. Now matches the `client_unix.go` / `client_windows.go` convention from `pkg/container/docker/sdk/`.

pkg/api/socket_windows.go
samuv: `Server.IdleTimeout` added in #5214 (120s, cross-platform), with a comment in the field block tying it back to `winio.MaxInstances=255` so the rationale isn't lost. Deferred the explicit connection cap until we have evidence of bursty traffic — happy to add `Server.MaxConcurrentConnections` (or a custom limiter middleware) in a follow-up if you'd like a hard ceiling now.

pkg/api/socket_unix_test.go
samuv: Done in 785ddc3. `TestCreateListener_NamedPipe_Unsupported` is now in the renamed `socket_unix_test.go` with the exact body you suggested — it calls `createListener(\`\\\\.\\pipe\\thv-api\`, true)` and asserts the error contains `"only supported on Windows"`.

pkg/server/discovery/health_test.go
samuv: Done in #5216. `TestWriteReadServerInfo_NamedPipe` lives in `pkg/server/discovery/discovery_test.go` next to the existing `_TCP` / `_UnixSocket` round-trip tests (as you suggested), so the producer/consumer contract for `npipe://` is now pinned end to end through the discovery file's JSON serialization.

pkg/api/socket_windows_test.go
samuv: Done in #5216. `TestSetupUnixSocket_NamedPipe_FirstInstanceWins` calls `setupUnixSocket` twice on the same pipe path and asserts the second call returns an error wrapping `"failed to create named pipe listener"`, which is what winio's `FILE_FLAG_FIRST_PIPE_INSTANCE` produces. A future winio bump that silently relaxed first-wins would now fail this test instead of corrupting prod.

pkg/server/discovery/health_windows_test.go
samuv: Done in #5216. `TestCheckHealth_NamedPipe_HungServerCancelsOnContext` spins up a `winio.ListenPipe`, drains accepts in a goroutine but never writes a response, then calls `CheckHealth` with a 250ms context. Asserts the call returns within 2s (well below the 5s `healthTimeout`), so a hung peer can no longer wedge the discovery `StateUnhealthy` probe.

aponcedeleonch (approved):

Issue #3125, opened 2025-12-20
Simplify VMCP Configuration


jerm-dro:
## Summary

VirtualMCPServer configuration is defined at least thrice: once in the CRD spec (`cmd/thv-operator/api/v1alpha1/`), again in the server config (`pkg/vmcp/config/`), and once more in `pkg/vmcp/config/yaml_loader.go`. Adding or modifying any configurable feature requires touching multiple layers, with bugs at any step going undetected until production.

### Developer Velocity
Every new configurable feature must traverse the entire configuration pipeline: CRD type → Converter → Server Config → Execution. This multiplies the effort required for even simple additions and creates friction for contributors.

### Quality / Bug Prevention  
Missing or incorrect transformations at any step result in silent configuration bugs. Real examples:
- **Telemetry configuration**: Worked in integration tests but failed with CRD due to broken rawConfig conversion
- **PR #3118**: `on_error.action: continue` was silently ignored due to broken config-to-execution plumbing

Integration tests pass because they configure execution-layer data structures directly, bypassing transformation layers entirely.

### Confusing Documentation
The current architecture produces two different configuration formats—YAML tags for CLI (`snake_case`) vs CRD spec (`camelCase`). Documenting the same configuration twice with subtle differences creates user confusion and maintenance burden. See PR #3070 which fixed documentation that incorrectly mixed these conventions.

### Higher Value, Lower Cost Testing
Issue #2852 proposes moving e2e tests to integration tests. This is blocked by the current architecture—integration tests that configure data structures directly miss config plumbing bugs. Simplifying the types makes this transition safe.

### API Extensibility
Fewer, more powerful data structures describing vmcp configuration make it easier to eventually support mutating and querying this data via API.

## Current Architecture (Problem)

Configuration flows through multiple transformations:

```
mcpv1alpha1.VirtualMCPServer → vmcpconfig.Config -> rawConfig during unmarshall -> vmcpconfig.Config  again  → Data Structures Used in Execution
```

Key problematic areas:
1. **Duplicative type definitions**: `VirtualMCPServerSpec` (CRD) and `Config` (server) define the same concepts with different types
2. **Manual conversion**: `cmd/thv-operator/pkg/vmcpconfig/converter.go` (900+ lines) manually maps CRD types to config types
3. **YAML loader duplication**: `pkg/vmcp/config/yaml_loader.go` has its own `rawConfig` types that mirror the CRD types
[...]

PR #5209, opened 2026-05-06
Add CRD-runtime drift detection test framework

ChrisJBurns:
## Summary

- Silent drift between CRD types and their runtime config counterparts has caused user-facing bugs (e.g. #3118, #3125) where new fields appeared to work in unit tests but were not wired through the CRD conversion path. The only existing safety nets — code review and end-to-end tests — do not scale.
- Adds a reusable reflection-based field walker and a bidirectional drift detection pattern. The pattern requires every leaf JSON field on either side of a CRD↔runtime boundary to be either explicitly mapped to the other side, or explicitly declared as intentionally unmapped with a justification string. Drift is allowed; *unannounced* drift is not.
- Applies the pattern to `MCPTelemetryConfig` ↔ `telemetry.Config` as the first real-world exercise. The drift test passes against the current codebase.

## Type of change

- [x] New feature

## Test plan

Tests added (all passing): walker unit tests in `cmd/thv-operator/internal/testutil/reflect_test.go`; drift tests in `cmd/thv-operator/pkg/spectoconfig/telemetry_drift_test.go` (`TestTelemetryConfigDrift_CRDFieldsCovered`, `TestTelemetryConfigDrift_RuntimeFieldsCovered`, `TestTelemetryConfigDrift_MappingTableSanity`).

`task lint-fix` could not be run locally — the installed `golangci-lint` v2.12.1 is built with go1.25 but the project targets go1.26. `go vet` and `gofmt` are clean on the new files. CI will run the project's pinned linter.

## API Compatibility

Test-only change. No production code, CRD schemas, or APIs are modified.

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/internal/testutil/reflect.go` | New `FlattenJSONLeafFields` helper that returns sorted dot-delimited JSON leaf paths for a struct type. Recurses into nested structs, deref'd pointers, and slice/map element types. Stops at primitives and a small allowlist (`time.Duration`, `metav1.Duration`, `json.RawMessage`, `thvjson.Map/Any`, `vmcpconfig.Duration`). Skips `metav1.TypeMeta`/`ObjectMeta`/`ListMeta` and unexported / `json:\"-\"` fields. Honors `,inline` and Go's anonymous-field promotion. Cycle-protected. |
| `cmd/thv-operator/internal/testutil/reflect_test.go` | Table-driven walker tests: pointers, slice/array/map elements, embedded structs (with and without `,inline`), `json:\"-\"`, missing tags, omitempty, unexported fields, leaf allowlist, sorted output, nil/non-struct input. |
| `cmd/thv-operator/pkg/spectoconfig/telemetry_drift_test.go` | Bidirectional drift test for `MCPTelemetryConfigSpec` ↔ `telemetry.Config`. Two coverage tests (one per direction) sharing a single `telemetryFieldMappings` source-of-truth table, plus per-side `IgnoredOnCRDOnly` / `IgnoredOnRuntimeOnly` maps with justification strings. A third test asserts mapping-table sanity (no duplicates, no overlap with ignore lists, non-empty justifications). |

No.

The pattern is bidirectional by design — it does **not** mandate parity between CRD and runtime types. Either side may have fields the other lacks; the test simply forces every divergence to be explicitly classified. New entries in `IgnoredOnCRDOnly` / `IgnoredOnRuntimeOnly` require a justification string, which makes intentional decoupling visible in code review.

The walker uncovered one CRD leaf that prior manual analysis missed: `openTelemetry.caBundleRef.configMapRef.optional`, a `*bool` promoted from `corev1.ConfigMapKeySelector` via `LocalObjectReference`. It is now explicitly declared in `telemetryIgnoredOnCRDOnly` with a justification — exactly the pattern working as intended.

This is the first PR in a series. Follow-ups will (1) introduce a CRD-owned `v1beta1.VirtualMCPConfig` mirror to decouple `VirtualMCPServerSpec.Config` from `pkg/vmcp/config.Config` without changing the user-facing CRD schema, and (2) extend the drift pattern to other converter boundaries (OIDC, external auth strategies, embedded auth server config).

## Implementation plan

<details>
<summary>Approved drift detection plan</summary>

PR 1 (this PR): build the reusable reflection walker + bidirectional drift test machinery, exercise it against an already-decoupled boundary (telemetry) to validate the pattern.

PR 2 (next): mirror `vmcpconfig.Config` top-level fields into `v1beta1.VirtualMCPConfig`, with nested fields still pointing at runtime types. CRD JSON schema unchanged. Replace `vmcp.Spec.Config.DeepCopy()` in the converter with explicit field-by-field copy at the top level.

PR 3+: add bidirectional drift tests for `VirtualMCPConfig` and incrementally mirror nested types (`AggregationConfig`, `OperationalConfig`, …), one per PR, with the drift test guiding each mirror.

</details>


cmd/thv-operator/pkg/spectoconfig/telemetry_drift_test.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// telemetryIgnoredOnRuntimeOnly lists runtime leaf fields that intentionally
+// have no CRD counterpart. Each entry MUST include a justification.
+var telemetryIgnoredOnRuntimeOnly = map[string]string{
+	"serviceName": "per-server, set from MCPTelemetryConfigReference.ServiceName and defaulted at runtime by " +
+		"telemetry.ResolveServiceName; intentionally absent from the shared MCPTelemetryConfig",
+	"serviceVersion":       "resolved at runtime from binary version (issue #2296)",
+	"environmentVariables": "CLI-only, not applicable to CRD-managed telemetry",
+	"caCertPath": "filesystem path injected by runconfig.AppendTelemetryRunnerOption after the operator computes the volume mount " +
```
jhrozek: The justification names `runconfig.AppendTelemetryRunnerOption`, which doesn't exist anywhere in the tree. Did you mean `runconfig.AddMCPTelemetryConfigRefOptions` (in `cmd/thv-operator/pkg/runconfig/telemetry.go`), where `config.CACertPath = caBundleFilePath` is actually assigned?

cmd/thv-operator/internal/testutil/reflect.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+
+		// Determine the path segment for this field.
+		segment := func() string {
+			// Anonymous fields are inlined by encoding/json when they have no
+			// json name (or an explicit `,inline` tag). Treat them as inline.
+			if field.Anonymous && (name == "" || inline) {
+				return ""
+			}
+			if inline {
```
jhrozek: The standalone `if inline { return "" }` collapses the segment for any `,inline`-tagged field, including non-anonymous named ones. `encoding/json` does not recognize `,inline` — collapse only happens via Go's anonymous-field promotion (already handled by the `field.Anonymous && (name == "" || inline)` branch above). A future named field like `Foo SomeStruct` tagged `json:"foo,inline"` would be flattened by the walker but rendered as `"foo":{...}` by `encoding/json`, silently producing wrong drift results. Recommend dropping this branch and relying on the anonymous-field check.

jhrozek (dismissed):
Framework is well-conceived, walker semantics are sound, and the drift pattern is a good safety net for future CRD↔runtime work. Two non-blocking notes inline — approving.

cmd/thv-operator/pkg/spectoconfig/telemetry_drift_test.go
jerm-dro: **suggestion**: can `telemetryIgnoredOnRuntimeOnly` and `telemetryIgnoredOnCRDOnly` be specified as annotations/tags on the actual struct itself? That keeps the type's documentation in one spot. Annotations are accessible via the `reflect` package.

cmd/thv-operator/pkg/spectoconfig/telemetry_drift_test.go
```diff
@@ -0,0 +1,232 @@
...
+		if _, live := runtimeLeaves[m.Runtime]; !live {
+			t.Errorf("telemetryFieldMappings entry %q is not a live leaf on telemetry.Config — stale entry?", m.Runtime)
+		}
+	}
+	for path := range telemetryIgnoredOnRuntimeOnly {
+		if _, live := runtimeLeaves[path]; !live {
+			t.Errorf("telemetryIgnoredOnRuntimeOnly entry %q is not a live leaf on telemetry.Config — stale entry?", path)
+		}
+	}
+}
```
jerm-dro: What would it look like to have a generic helper here, something like:

```
func TestTelemetryConfigDrift(t testing.T) {
   drift.Test[telemetry.Config, v1beta1.MCPTelemetryConfigSpec](mappings)
}
```

We want this dead-simple so agents know exactly how to follow this pattern and there's no room for slight deviations.

jerm-dro (commented):
I like this direction. A few suggestions to make this more general purpose and require less boilerplate.

cmd/thv-operator/pkg/spectoconfig/telemetry_drift_test.go
ChrisJBurns: Went back and forth with CC and it suggests keeping the maps. An ignore entry is really a cross-type fact ("no peer on the other side"), not an intrinsic property of one field. Putting it on the type also pushes edits into stable v1beta1 API code or into `pkg/telemetry`, which is shared with the standalone `thv vmcp serve` CLI and shouldn't carry operator-test metadata. Justification strings with parens, commas, and concatenation flatten awkwardly into tag values too. Since the Mappings table has to stay relational regardless, the savings would only be ~30 lines of strings. Happy to revisit if I'm missing something.

cmd/thv-operator/internal/testutil/drift.go
```diff
@@ -0,0 +1,248 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package testutil
```
jerm-dro, line 4: suggestion: I don't think this should be entirely a `testutil`. 

Ideally, the mapping defined in `telemetry_drift_test.go` is what is used in the production code conversion. Then, in a test, we assert on the mappings and types that there has been no drift. In other words, the only test part of this code should be:

```
func TestTelemetryConfigDrift(t *testing.T) {
	t.Parallel()
	testutil.AssertNoDrift[telemetry.Config, v1beta1.MCPTelemetryConfigSpec](t, testutil.DriftSpec{
		Domain:               "telemetry",
        // The following values are defined in the production code,
        // since they are used to drive the conversion.
		Mappings:             telemetryFieldMappings,
		IgnoredOnCRDOnly:     telemetryIgnoredOnCRDOnly,
		IgnoredOnRuntimeOnly: telemetryIgnoredOnRuntimeOnly,
	})
}
```

WDYT?

jerm-dro (approved):
Approving since this is a net-positive but I have one suggestion on how to make this more useful beyond testing.

PR #5215, opened 2026-05-07
Move pipe and socket URL handling to net/url

samuv:
## Summary

Follow-up to #5201 ([review thread](https://github.com/stacklok/toolhive/pull/5201#pullrequestreview-4243582372)). Stacked on `socker-windows`; will rebase onto `main` once #5201 merges. Sibling to #5214.

Replaces prefix-and-trim string handling with `net/url.Parse` on both producer (`socketURL`) and consumer (`HTTPClientForURL`, `ParseNamedPipeURL`, `ParseUnixSocketPath`) sides so pipe and socket addresses round-trip through one library and one set of rules.

The change includes one real bug fix: `socketURL` previously emitted `unix://C:\path\thv.sock` for Windows AF_UNIX paths, which `url.Parse` rejects with `invalid port ":\\path\\thv.sock"`. The dispatcher only worked because it used `strings.TrimPrefix`. New behavior emits `unix:///C:%5Cpath%5Cthv.sock`, which round-trips cleanly. Round-trip is now exercised by `TestSocketURL_RoundTrip_Unix` and `TestSocketURL_RoundTrip_NamedPipe`.

Other improvements that fall out of the migration:

- **Case-insensitive scheme dispatch** — `NPIPE://x` routes the same arm as `npipe://x` because `url.Parse` lowercases the scheme during parsing. Pinned by `TestHTTPClientForURL_SchemeDispatchCaseInsensitive`.
- **Case-insensitive pipe-name normalization** — `npipe://Thv-API` resolves to `\\.\pipe\thv-api` via `strings.ToLower(u.Hostname())`; matches the kernel's case-insensitive pipe namespace.
- **Length cap** on pipe names (247 chars, the `CreateNamedPipeW` `lpName` limit after the `\\.\pipe\` prefix).
- **Reserved-name rejection** for legacy Windows device names (`CON`, `NUL`, `PRN`, `AUX`, `COM1-9`, `LPT1-9`).
- **Positive charset** of `[A-Za-z0-9._-]+` for pipe names — replaces the over-eager `strings.Contains(name, "..")` guard which rejected legitimate names like `my..api`.
- **Stricter URL shape**: `ParseNamedPipeURL` rejects pipe URLs with non-empty path / query / fragment / userinfo / port; `ParseUnixSocketPath` rejects unix URLs with non-empty authority.

Addresses inline review comments [3201085366](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085366) (`ParseNamedPipeURL` migration), [3201085375](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085375) (dispatcher migration), and [3201085383](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085383) (`socketURL` / `ParseUnixSocketPath` migration + Windows AF_UNIX bug).

## Type of change

- Bug fix (Windows AF_UNIX URL round-trip)
- Code quality / hardening

## Test plan

- [x] Unit tests (`go test ./pkg/api/... ./pkg/server/discovery/...`) — green on macOS; new round-trip tests, scheme-case test, reserved-name tests, length-cap test, and shape-rejection cases all pass.
- [x] `GOOS=windows go vet` and `GOOS=windows go test -c -o /dev/null` for both packages — both clean.

Yes (Windows-only). On Windows, `socketURL` for AF_UNIX paths now emits a percent-encoded URL form. The discovery file is local per-user and overwritten on every `thv serve` startup, so there is no on-disk migration concern. POSIX paths like `/tmp/thv.sock` produce identical output before and after.

### DIFF
--- a/pkg/api/socket_windows.go
+++ b/pkg/api/socket_windows.go
@@ -11,6 +11,7 @@ import (
 	"io/fs"
 	"log/slog"
 	"net"
+	"net/url"
 	"os"
 	"path/filepath"
 
@@ -95,10 +96,14 @@ func cleanupUnixSocket(address string) {
 
 // socketURL returns the URL form of a Unix-socket or named-pipe address for
 // the discovery file. Named pipes are emitted as npipe://<name> where <name>
-// is everything after the \\.\pipe\ prefix.
+// is everything after the \\.\pipe\ prefix. AF_UNIX paths are emitted via
+// (&url.URL{}).String() so a Windows drive-letter path round-trips through
+// net/url cleanly (the previous concatenation form produced unix://C:\... ,
+// which url.Parse rejects with "invalid port").
 func socketURL(address string) string {
 	if isNamedPipeAddress(address) {
-		return "npipe://" + address[len(namedPipePrefix):]
+		name := address[len(namedPipePrefix):]
+		return (&url.URL{Scheme: "npipe", Host: name}).String()
 	}
-	return "unix://" + address
+	return (&url.URL{Scheme: "unix", Path: address}).String()
 }
--- a/pkg/api/socket_windows.go
+++ b/pkg/api/socket_windows.go
@@ -96,14 +96,21 @@ func cleanupUnixSocket(address string) {
 
 // socketURL returns the URL form of a Unix-socket or named-pipe address for
 // the discovery file. Named pipes are emitted as npipe://<name> where <name>
-// is everything after the \\.\pipe\ prefix. AF_UNIX paths are emitted via
-// (&url.URL{}).String() so a Windows drive-letter path round-trips through
-// net/url cleanly (the previous concatenation form produced unix://C:\... ,
-// which url.Parse rejects with "invalid port").
+// is everything after the \\.\pipe\ prefix. AF_UNIX paths are emitted as
+// unix:///<path> so a Windows drive-letter path round-trips through net/url
+// cleanly (the previous concatenation form produced unix://C:\... , which
+// url.Parse rejects with "invalid port").
+//
+// The synthetic leading slash is required: with Host == "" and a Path that
+// does not start with /, url.URL.String() emits only scheme://+path (two
+// slashes), and url.Parse then mis-reads the drive letter as host:port.
+// Forcing the leading slash yields the three-slash form, and the consumer
+// (ParseUnixSocketPath) strips that synthetic slash before the drive letter
+// so the round-trip closes.
 func socketURL(address string) string {
 	if isNamedPipeAddress(address) {
 		name := address[len(namedPipePrefix):]
 		return (&url.URL{Scheme: "npipe", Host: name}).String()
 	}
-	return (&url.URL{Scheme: "unix", Path: address}).String()
+	return (&url.URL{Scheme: "unix", Path: "/" + address}).String()
 }
--- a/pkg/api/socket_windows_test.go
+++ b/pkg/api/socket_windows_test.go
@@ -15,6 +15,8 @@ import (
 	"github.com/Microsoft/go-winio"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/stacklok/toolhive/pkg/server/discovery"
 )
 
 // pipeNameSeq disambiguates concurrent test pipes so parallel runs don't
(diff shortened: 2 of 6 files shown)


pkg/api/socket_windows.go
```diff
@@ -82,10 +83,14 @@ func cleanupUnixSocket(address string) {
...
+// net/url cleanly (the previous concatenation form produced unix://C:\... ,
+// which url.Parse rejects with "invalid port").
 func socketURL(address string) string {
 	if isNamedPipeAddress(address) {
-		return "npipe://" + address[len(namedPipePrefix):]
+		name := address[len(namedPipePrefix):]
+		return (&url.URL{Scheme: "npipe", Host: name}).String()
 	}
-	return "unix://" + address
+	return (&url.URL{Scheme: "unix", Path: address}).String()
```
jhrozek: I think this branch doesn't do what the doc comment claims. With `Host: ""` and a path that doesn't start with `/`, `url.URL.String()` only writes `scheme://` + path — there's no slash inserted between the empty authority and the path because Go only does that when `Host != ""`.

So for `address = "C:\path\thv.sock"` we get:

```
(&url.URL{Scheme:"unix", Path: address}).String()
  → "unix://C:%5Cpath%5Cthv.sock"   // two slashes, not three

url.Parse("unix://C:%5Cpath%5Cthv.sock")
  → invalid port ":%5Cpath%5Cthv.sock" after host
```

That's the same `invalid port` error the commit message says we're fixing, so afaict the consumer side still can't parse what the producer emits on Windows AF_UNIX. The Windows test row at `socket_windows_test.go:42` asserts the three-slash form (`unix:///C:%5Cpath%5Cthv.sock`), which would fail when CI actually runs on Windows.

Forcing a leading slash on `Path` makes `String()` emit the three-slash form, and `ParseUnixSocketPath` already strips that synthetic leading slash before the drive letter, so the round-trip closes:

```suggestion
	return (&url.URL{Scheme: "unix", Path: "/" + address}).String()
```

Verified locally:

```
(&url.URL{Scheme:"unix", Path:"/" + `C:\path\thv.sock`}).String() = "unix:///C:%5Cpath%5Cthv.sock"
url.Parse(...) → Path="/C:\\path\\thv.sock", err=nil
```

jhrozek (commented):
Two related issues on the Windows side. The af_unix branch of socketURL doesn't actually produce the URL the test fixture asserts, and there's no round-trip test for that branch — which is why the mismatch slipped through. Both should be addressable with a one-char producer change plus the matching round-trip test. Verified locally with Go 1.26.

Rest of the change reads fine.

pkg/api/socket_windows_test.go
```diff
@@ -43,6 +49,16 @@ func TestSocketURL_Windows(t *testing.T) {
...
 
+// TestSocketURL_RoundTrip_NamedPipe pins that socketURL output is always
+// parseable by ParseNamedPipeURL, closing the producer/consumer loop.
+func TestSocketURL_RoundTrip_NamedPipe(t *testing.T) {
+	t.Parallel()
+	addr := `\\.\pipe\thv-api`
+	got, err := discovery.ParseNamedPipeURL(socketURL(addr))
+	require.NoError(t, err)
+	assert.Equal(t, addr, got)
+}
```
jhrozek, line 60: There's a round-trip test for named pipes but not for the AF_UNIX Windows path, which is exactly the case where the producer/consumer contract is most fragile (the drive-letter percent-encoding dance). Adding a parallel test would have caught the mismatch on the other comment immediately, and it pins the contract going forward.

```suggestion
}

// TestSocketURL_RoundTrip_AFUnix pins that socketURL output for an AF_UNIX
// drive-letter path is always parseable by ParseUnixSocketPath, closing the
// producer/consumer loop on the case where percent-encoding matters most.
func TestSocketURL_RoundTrip_AFUnix(t *testing.T) {
	t.Parallel()
	addr := `C:\path\thv.sock`
	got, err := discovery.ParseUnixSocketPath(socketURL(addr))
	require.NoError(t, err)
	assert.Equal(t, addr, got)
}
```

pkg/api/socket_windows.go
samuv: Good catch — diagnosis is exact. Reproduced locally:

```
(&url.URL{Scheme:"unix", Path:`C:\path\thv.sock`}).String()
  → "unix://C:%5Cpath%5Cthv.sock"   (two slashes)
url.Parse(...) → invalid port ":%5Cpath%5Cthv.sock" after host

(&url.URL{Scheme:"unix", Path:"/" + `C:\path\thv.sock`}).String()
  → "unix:///C:%5Cpath%5Cthv.sock"  (three slashes)
url.Parse(...) → Path="/C:\\path\\thv.sock", err=nil
```

Applied the suggested one-char fix on the AF_UNIX branch only (the npipe branch sets `Host`, so it already gets the right shape; `socket_unix.go` is also unchanged because POSIX paths already start with `/`, so prepending another would yield `unix:////tmp/thv.sock` and break that round-trip).

Also expanded the doc comment on `socketURL` to call out *why* the synthetic leading slash is required so the next person reading the code doesn't trip the same edge.

Fixed in df44c257.

pkg/api/socket_windows_test.go
samuv: Agreed — that's exactly the test that would have made the producer/consumer mismatch fail closed instead of relying on the fixture row. Added `TestSocketURL_RoundTrip_AFUnix` as a sibling to `TestSocketURL_RoundTrip_NamedPipe` so both Windows transports are pinned the same way.

Fixed in df44c257.

aponcedeleonch (approved):
Approving since Jakub's feedback is also taken care of

jhrozek (approved):
Both Windows af_unix findings addressed cleanly in b0ffadfe7. The producer now emits the three-slash form, the new round-trip test closes the loop, and the expanded doc comment explains why the leading slash is load-bearing. LGTM.

PR #5216, opened 2026-05-07
Pin npipe round-trip and pipe lifecycle invariants

samuv:
## Summary

Follow-up to #5201 ([review thread](https://github.com/stacklok/toolhive/pull/5201#pullrequestreview-4243582372)). Stacked on `socker-windows`; will rebase onto `main` once #5201 merges. Sibling to #5214 and #5215.

Adds three regression tests around the named-pipe transport that the original PR did not cover. None reproduce a live bug today; they pin invariants the named-pipe path implicitly relies on so a future change can't silently break them.

- **`TestWriteReadServerInfo_NamedPipe`** (cross-platform) — closes the producer/consumer loop for the discovery file: an `npipe://` URL written by `socketURL` must survive `WriteServerInfo` + `ReadServerInfo` without being mangled. The individual emit/parse pieces have their own tests; this one pins the seam.
- **`TestSetupUnixSocket_NamedPipe_FirstInstanceWins`** (Windows-only) — asserts that two `winio.ListenPipe` calls on the same name fail the second time. `winio` sets `FILE_FLAG_FIRST_PIPE_INSTANCE` so two `thv` processes cannot bind the same pipe and race for traffic; if a future `winio` bump silently relaxed that, this test catches it before the discovery layer does in production.
- **`TestCheckHealth_NamedPipe_HungServerCancelsOnContext`** (Windows-only) — covers the `StateUnhealthy` path: a peer that accepts connections but never responds must not wedge `CheckHealth` past the caller's context deadline. The existing success and not-found tests don't exercise this case.

Addresses inline review comments [3201085442](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085442) (round-trip), [3201085446](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085446) (first-instance-wins), and [3201085449](https://github.com/stacklok/toolhive/pull/5201#discussion_r3201085449) (hung-peer).

## Type of change

- Test coverage / regression hardening

## Test plan

- [x] `go test ./pkg/api/... ./pkg/server/discovery/...` — green on macOS (round-trip and Unix tests).
- [x] `GOOS=windows go vet` and `GOOS=windows go test -c -o /dev/null` for both packages — both clean.
- [x] Manual `go test -tags windows -run NamedPipe` on a Windows host — pending; deferred to a reviewer with a Windows host. The Windows-only tests are guarded by build tags and do not run on macOS / Linux.

No.

### DIFF
--- a/pkg/api/socket_windows_test.go
+++ b/pkg/api/socket_windows_test.go
@@ -131,10 +131,9 @@ func TestSetupUnixSocket_NamedPipe_FirstInstanceWins(t *testing.T) {
 	require.NoError(t, err)
 	t.Cleanup(func() { _ = first.Close() })
 
-	second, err := setupUnixSocket(pipePath)
+	// setupUnixSocket returns (nil, err) on the named-pipe failure path, so
+	// no defensive Close is needed for the second listener.
+	_, err = setupUnixSocket(pipePath)
 	require.Error(t, err, "second ListenPipe on the same name must fail")
-	if second != nil {
-		_ = second.Close()
-	}
 	assert.Contains(t, err.Error(), "failed to create named pipe listener")
 }
--- a/pkg/api/socket_windows_test.go
+++ b/pkg/api/socket_windows_test.go
@@ -116,3 +116,25 @@ func TestCleanupUnixSocket_NamedPipe_NoOp(t *testing.T) {
 	// cleanly.
 	cleanupUnixSocket(`\\.\pipe\thv-api-cleanup-noop`)
 }
+
+// TestSetupUnixSocket_NamedPipe_FirstInstanceWins pins winio's first-wins
+// semantics: ListenPipe sets FILE_FLAG_FIRST_PIPE_INSTANCE, so a second
+// listener targeting the same name must fail. This is the safety net the
+// discovery layer relies on to detect a stale-PID conflict; if a future winio
+// bump silently relaxed it, two thv processes could bind the same pipe and
+// quietly race for traffic.
+func TestSetupUnixSocket_NamedPipe_FirstInstanceWins(t *testing.T) {
+	t.Parallel()
+	pipePath := uniqueTestPipe()
+
+	first, err := setupUnixSocket(pipePath)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = first.Close() })
+
+	second, err := setupUnixSocket(pipePath)
+	require.Error(t, err, "second ListenPipe on the same name must fail")
+	if second != nil {
+		_ = second.Close()
+	}
+	assert.Contains(t, err.Error(), "failed to create named pipe listener")
+}
--- a/pkg/server/discovery/health_windows_test.go
+++ b/pkg/server/discovery/health_windows_test.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/Microsoft/go-winio"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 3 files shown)


pkg/server/discovery/health_windows_test.go
```diff
@@ -51,3 +52,46 @@ func TestCheckHealth_NamedPipe_NotFound(t *testing.T) {
...
+		for {
+			conn, acceptErr := listener.Accept()
+			if acceptErr != nil {
+				return
+			}
+			// Hold the connection open without responding so CheckHealth's
+			// HTTP read blocks until the context deadline fires.
+			t.Cleanup(func() { _ = conn.Close() })
+		}
+	}()
```
aponcedeleonch, line 104: `t.Cleanup` is meant to be called from the test goroutine. Calling it from a background goroutine races with test teardown: registrations after the test body returns can panic with "Log in goroutine after Test has completed" or silently no-op. In this specific test the dial that `CheckHealth` makes accepts before the test body returns, so it works in the happy path, but correctness depends on accept timing.

Cleaner pattern: collect conns into a slice guarded by a mutex and close them all in a single `t.Cleanup` registered from the test body before launching the goroutine.

```go
var (
    mu    sync.Mutex
    conns []net.Conn
)
t.Cleanup(func() {
    mu.Lock()
    defer mu.Unlock()
    for _, c := range conns {
        _ = c.Close()
    }
})
go func() {
    for {
        conn, acceptErr := listener.Accept()
        if acceptErr != nil {
            return
        }
        mu.Lock()
        conns = append(conns, conn)
        mu.Unlock()
    }
}()
```

pkg/api/socket_windows_test.go
```diff
@@ -116,3 +116,25 @@ func TestCleanupUnixSocket_NamedPipe_NoOp(t *testing.T) {
...
+
+	first, err := setupUnixSocket(pipePath)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = first.Close() })
+
+	second, err := setupUnixSocket(pipePath)
+	require.Error(t, err, "second ListenPipe on the same name must fail")
+	if second != nil {
+		_ = second.Close()
+	}
```
aponcedeleonch: Dead code in practice since winio returns a nil listener on error. Defensible as defense-in-depth, fine to drop for noise reduction. Take it or leave it.

aponcedeleonch (commented):
Two comments on the new tests.

pkg/server/discovery/health_windows_test.go
samuv: Good catch — you're right that calling `t.Cleanup` from the accept goroutine is load-bearing on the dial timing. Adopted your suggested pattern in f2d36929: pre-registered a single `t.Cleanup` from the test body that drains a `[]net.Conn` under a mutex, with the accept loop just appending under the same mutex.

One small wrinkle worth noting: the listener `t.Cleanup(listener.Close)` is registered before the conn-cleanup, so by Go's LIFO cleanup ordering the conn-close runs first and the listener-close second. That's actually the order we want here: by the time the conn-close iterates the slice, no new dials are arriving (the test body has long since returned, and `CheckHealth` was the only dialer), so the slice is stable. The listener-close then unblocks the accept goroutine, which exits cleanly.

Fixed in f2d36929.

pkg/api/socket_windows_test.go
samuv: Took it — dropped the guard and replaced it with a one-line comment so the next reader doesn't re-introduce it. `setupUnixSocket` returns `(nil, err)` on the named-pipe failure path, so the guard could never have been entered. Fixed in f2d36929.

aponcedeleonch (approved):
Both review comments addressed cleanly. The mutex-guarded slice + pre-registered t.Cleanup matches the suggested rewrite, and the LIFO ordering note in the commit message is a nice touch. LGTM.

Issue #5159, opened 2026-05-01
Recognise identityFromToken in synthesis-mode detection


jhrozek:
# Recognise `identityFromToken` in synthesis-mode detection

## Description

Final phase of the Snowflake / `identityFromToken` story (#5150).
PR #5094 added two synthesis-mode-detection sites — an operator status
condition (`IdentitySynthesized`) and a runtime WARN at provider
construction — both of which check only `UserInfo == nil`. After
`identityFromToken` lands, configs that use it instead of `userInfo`
have a real-identity resolution path but are falsely flagged as
synthesis-mode by both surfaces. Update both predicates and their
associated text so the operator-visible state matches reality.

## Context

`EmbeddedAuthServerConfig.SyntheticIdentityUpstreams` drives the
`IdentitySynthesized` advisory condition published on both
`MCPExternalAuthConfig` and `VirtualMCPServer` status. The runtime
`NewOAuth2Provider` WARN fires once at construction. Both currently
treat "no userInfo" as the synthesis-mode signal.

When `identityFromToken` is configured, the runtime priority chain
honours it BEFORE considering synthesis, so the upstream is NOT in
synthesis mode. The detection predicates need to also exclude
upstreams that have `identityFromToken` set.

The condition messages and the runtime WARN text both need updating
to recommend either `userInfo` OR `identityFromToken` as the
identity-resolution alternatives — operators encountering the
condition should see the new option.

The reason-constant string values change to reflect the broader
criterion:

- `OAuth2UpstreamWithoutUserInfo` → `OAuth2UpstreamWithoutIdentityResolution`
- `AllUpstreamsHaveUserInfo` → `AllUpstreamsHaveIdentityResolution`

This is a breaking change in the observable Kubernetes API surface of
the `IdentitySynthesized` condition. PR #5094 just landed with no
released consumers, so the rename is safe now and would not have been
if deferred. Anyone matching the old strings in alerting rules,
dashboards, or GitOps tooling will need to update them.

Per `.claude/rules/operator.md`'s status-condition-parity rule, both
`MCPExternalAuthConfig` and `VirtualMCPServer` controllers must use
identical condition messages — they share the helper here, so the
fix is single-source by construction.

**Dependencies**: #5155 (CRD type), #5156 (provider integration so
the runtime priority chain actually does what the predicate now
claims).
**Blocks**: nothing — this is the final piece.

## Acceptance Criteria

[...]

Issue #5157, opened 2026-05-01
Translate identityFromToken from CRD to runtime config


jhrozek:
# Translate `identityFromToken` from CRD to runtime config

## Description

Phase 4 of the Snowflake / `identityFromToken` story (#5150). Wire the
new CRD field through the operator-side run-config so the embedded
auth server actually receives it. Mirrors the existing
`tokenResponseMapping` plumbing exactly. Without this, the CRD field
exists and the provider integration honours it, but no real cluster
deployment can connect the two ends.

## Context

The CRD field added in #5155 lives on the v1beta1
`OAuth2UpstreamConfig`. The runtime field consumed by the OAuth2
upstream provider in #5156 lives on `pkg/authserver/upstream`'s
`OAuth2Config`. Two translation points carry it through:

- `cmd/thv-operator/pkg/controllerutil/`'s auth-server config builder
  takes the v1beta1 type and produces a `pkg/authserver` run-config.
- `pkg/authserver/runner/`'s embedded-auth-server bootstrapper takes
  that run-config and produces the runtime `pkg/authserver/upstream`
  config the provider consumes.

Both translations are mechanical mirrors of the existing
`tokenResponseMapping` translation — no validation logic in the
runner; that is owned by the CRD admission validator and the
`OAuth2Config.Validate()` method.

**Dependencies**: #5155 (CRD type), #5156 (runtime config mirror).
**Blocks**: nothing in the implementation chain — but a real cluster
deployment cannot honour `identityFromToken` until both translation
points are in place.

## Acceptance Criteria

- [ ] The v1beta1 → run-config translation copies the
  `identityFromToken` block (with all three paths) into the
  authserver run-config.
- [ ] The run-config → runtime translation does the same for the
  `pkg/authserver/upstream` provider config.
- [ ] Nil-config does not allocate a runtime block (no spurious
  pointers).
- [ ] Round-trip tests on both translation points cover the all-three-
  paths case and the nil case.
- [ ] `task lint-fix` and `task test` pass.

## Out of Scope

- Validation logic — already enforced by the CRD admission validator
  (#5155) and the runtime `Validate()` (#5156).
- Synthesis-mode-detection helpers — separate task.
- YAML examples — separate task.

jhrozek:
This was done

Issue #5155, opened 2026-05-01
Add identityFromToken to MCPExternalAuthConfig CRD


jhrozek:
# Add `identityFromToken` to MCPExternalAuthConfig CRD

## Description

Phase 2 of the Snowflake / `identityFromToken` story (#5150). Add a
new optional `identityFromToken` block on the `OAuth2UpstreamConfig`
CRD type, with admission validation and regenerated CRD manifests.
This makes the field observable to operators and reachable through
`kubectl apply`. The runtime still ignores the field at this point —
the wiring lands in the next phase.

## Context

The runtime helper from #5152 needs an operator-facing surface. A new
`IdentityFromTokenConfig` type with `subjectPath`, `namePath`, and
`emailPath` (gjson dot-notation paths) lives next to the existing
`tokenResponseMapping` and `userInfoConfig` types. The doc comment is
the authoritative place for the trust-model warning (claims are
trusted via TLS only, not cryptographically verified — prefer OIDC +
ID-token validation when verifiable identity matters) and the
subject-uniqueness guidance (scope is per-upstream-provider entry; use
distinct entries for distinct trust domains).

Cross-references in the existing `tokenResponseMapping` and
`userInfoConfig` doc comments help operators discover which block
applies to their provider.

Admission validation requires a non-empty `subjectPath` when the
block is set; `namePath` and `emailPath` are optional.

**Dependencies**: #5152 (runtime helper exists). The field is
structurally additive — existing manifests are unaffected.
**Blocks**: provider integration (next phase) cannot honour the field
until it exists in the CRD and runtime config.

## Acceptance Criteria

- [ ] `OAuth2UpstreamConfig` gains an optional `identityFromToken`
  field of a new `IdentityFromTokenConfig` type carrying the three
  paths.
- [ ] The new type's doc comment includes:
  - The trust-model warning (TLS-only, prefer OIDC for verifiable
    claims).
  - The subject-uniqueness note (per upstream provider entry; use
    distinct entries for distinct trust domains).
  - Brief gjson dot-notation guidance with examples.
- [ ] Doc-comment cross-references between `tokenResponseMapping`,
  `userInfoConfig`, and the new `identityFromToken` are coherent.
- [ ] Admission validation rejects an `identityFromToken` block with
  an empty `subjectPath`.
- [ ] CRD manifests and reference docs are regenerated; `task
  operator-generate`, `task operator-manifests`, and `task
  crdref-gen` are idempotent.
[...]

PR #5222, opened 2026-05-07
Wire identityFromToken into the OAuth2 upstream provider

### LINKED ISSUE #5156: Wire identityFromToken into the OAuth2 upstream provider
# Wire `identityFromToken` into the OAuth2 upstream provider

## Description

Phase 3 of the Snowflake / `identityFromToken` story (#5150). Connect
the pure helper (#5152) and the CRD field (#5155) into the embedded
auth server's runtime: extract identity from the raw token-endpoint
response body and use it as the resolved upstream identity, sibling
to the existing `userInfo` and PR #5094 synthesis paths.

## Context

The existing token-response rewriter already reads and parses every
successful token-endpoint response in order to normalise non-standard
envelopes (e.g. GovSlack's nested fields). Extend it to also extract
user identity claims from the same body when an
`IdentityFromTokenConfig` is configured on `OAuth2Config`.

Identity extraction runs on the RAW pre-rewrite body, so identity
paths are resolved against the original provider response even when
`tokenResponseMapping` is also configured (Slack v2 needs both: the
user-level access token nested under `authed_user.access_token` and
the user identifier under `authed_user.id`).

The new priority chain in `BaseOAuth2Provider.ExchangeCodeForIdentity`:

1. `IdentityFromToken` — when configured, return the extracted
   identity. If extraction failed (path didn't resolve), return
   `ErrIdentityResolutionFailed` without falling back to `userInfo`
   or synthesis — "identityFromToken set" is an explicit operator
   declaration that the token response is the identity source, and
   silently falling through would mask misconfiguration.
2. `UserInfo` — existing `fetchUserInfo` path, unchanged.
3. Synthesis — existing `synthesizeIdentity` path (PR #5094),
   unchanged.

The refresh path passes nil for the identity config, because providers
like Snowflake omit `username` on refresh; identity is captured once
at auth-code time and persisted into session storage by the callback
handler.

OIDC providers always have an ID-token-derived subject and discard
the rewriter's identityFromToken return value. A defence-in-depth
WARN fires if a future config-loader bug ever sets `IdentityFromToken`
on an OIDC base config (the field is structurally absent on the OIDC
CRD type today).

A tripwire test asserts the userinfo HTTP endpoint is never contacted
when `identityFromToken` is configured — including on extraction
failure. This is the single most important security-relevant test in
the chain.

**Dependencies**: #5152 (helper), #5155 (CRD type and runtime config
mirror).
**Blocks**: nothing in the implementation chain — the operator
translation phase consumes the runtime-config field added here, but
the controller-side translation can be built in parallel and merged
in either order.

## Acceptance Criteria

- [ ] `OAuth2Config` has the runtime mirror of `IdentityFromToken`,
  with `Validate()` requiring a non-empty subject path when the block
  is set.
- [ ] The token-response rewriter extracts identity from the raw
  pre-rewrite body when `IdentityFromTokenConfig` is configured;
  extraction runs before any wire-format rewrite.
- [ ] `ExchangeCodeForIdentity` honours the priority chain:
  `identityFromToken` first, then `userInfo`, then synthesis.
- [ ] When `identityFromToken` is configured but extraction fails
  (e.g. wrong path), the call returns an error wrapping
  `ErrIdentityResolutionFailed`. The userinfo HTTP endpoint is NOT
  contacted in this case (asserted by a tripwire test).
- [ ] When `identityFromToken` is configured and extraction succeeds,
  the returned `Identity` has `Synthetic == false` so the callback
  handler runs UserResolver normally.
- [ ] The refresh path does not perform identity extraction.
- [ ] OIDC providers ignore `IdentityFromToken` if it is somehow set
  on their base config, and emit a defence-in-depth WARN.
- [ ] No part of the response body appears in any returned error or
  any logged record at any level.
- [ ] `task lint-fix` and `task test` pass (including a `go test
  -race` run) with no regressions in the existing
  `pkg/authserver/upstream/...` test suite.

## Out of Scope

- Operator-side translation from CRD field to runtime config —
  separate task.
- Updating synthesis-mode-detection helpers (the operator status
  condition + provider-construction WARN) to recognise
  `identityFromToken` as a real-identity path — separate task.
- YAML examples — separate task.
- Audit-log changes — the JWT `name` claim populated from the
  extracted `Identity.Name` flows through the existing audit reader
  unchanged.

jhrozek:
## Summary

- The embedded auth server's pure-OAuth2 path resolved identity in two ways: a UserInfo HTTP fetch when `userInfo` was configured, otherwise a synthesized `tk-...` subject (PR #5094). Upstreams that expose identity directly in the token-endpoint response (Snowflake's `username`, Slack v2's `authed_user.id`) had no good answer — they would either go through synthesis (losing real subject identity for `users` row creation) or require a fake `userInfo` URL.
- This PR wires `IdentityFromToken` into `BaseOAuth2Provider.ExchangeCodeForIdentity` as a third resolution mode that runs ahead of `userInfo`. The existing `tokenResponseRewriter` already reads and parses every successful token-endpoint response to normalise non-standard envelopes; it now extracts identity claims from the same body using gjson dot-notation paths when an operator configures it.
- New priority chain: (1) `IdentityFromToken` extracts identity from the raw pre-rewrite body, (2) `UserInfo` fetches via HTTP (existing), (3) `synthesizeIdentity` falls back (existing). When extraction is configured but fails, the operator-actionable extractor error (path name + type description, never body content) is surfaced through `ErrIdentityResolutionFailed` rather than silently falling through to userinfo or synthesis — "identityFromToken set" is an explicit operator claim.
- `RefreshTokens` deliberately passes `nil` for the identity config: providers like Snowflake omit username on refresh; identity is cached at auth-code time and read from session storage on subsequent requests.
- The OIDC provider discards the rewriter's identity return value with a defensive WARN if a future config-loader bug ever sets `IdentityFromToken` on an OIDC base config (structurally absent on the OIDC CRD type today).
- Two `slog.Info` calls in `exchangeCodeForTokens` were downgraded to `slog.Debug` for silent-success-at-INFO compliance.

Closes #5156

## Type of change

- [x] New feature

## Test plan

New tests cover the happy path, the userinfo-bypass tripwire (asserts no userinfo HTTP call when identityFromToken is configured), the userInfo-only regression, the synthesis fallback, the refresh path (no identity extraction), the `@upstreamjwt` modifier path through the provider, and an information-disclosure assertion that the raw token body never appears in error messages.

## API Compatibility

This PR adds an additive optional field to the runtime `OAuth2Config` only; the matching CRD type lands in #5155.

No direct CRD-level user-facing change in this PR. End-to-end functionality is gated on the runtime translation in #5157 (the next PR in the stack), which wires the v1beta1 CRD field into this runtime config.

## Implementation plan

<details>

This is phase 3 of the Snowflake / `identityFromToken` story (#5150):

- Phase 1 (#5152, merged via PR #5200): pure helper `extractIdentityFromTokenResponse` and the `@upstreamjwt` gjson modifier.
- Phase 2 (#5155): `IdentityFromTokenConfig` CRD type on `MCPExternalAuthConfig` v1beta1.
- **Phase 3 (this PR, #5156)**: Wire the helper into `BaseOAuth2Provider.ExchangeCodeForIdentity` via the existing `tokenResponseRewriter`. New priority chain. Identity-source set on the rewriter, consumed by the caller after `Exchange` returns.
- Phase 4 (#5157): Operator-side translation from CRD to runtime config; bootstrap call to `RegisterModifiers()` so `@upstreamjwt` paths actually work in production.
- Phases 5-6 (#5158, #5159): Examples + synthesis-detection refinement so the `IdentitySynthesized` controller condition no longer fires when `IdentityFromToken` is configured.

Design constraints surfaced and addressed in this PR:

- Identity extraction runs on the raw pre-rewrite body. `rewriteTokenResponse` only relocates standard OAuth fields today and never touches identity-shaped fields, but extracting first makes the ordering invariant independent of that assumption.
- The rewriter is single-use by construction (`wrapHTTPClientForTokenExchange` creates a fresh one per call); no state shared across goroutines.
- Extraction failure does NOT fall through to userinfo. Operators who configure `IdentityFromToken` are claiming identity is in the token response; we surface the failure rather than mask it.
- The extractor's error format is `subjectPath %q %s` where `%s` is a static validator description ("path not found in token response", etc.). Never body bytes. The ExchangeCodeForIdentity caller wraps this directly so operators see the misconfigured path name in the returned error.

</details>

- **Stack ordering**: The doc comments at `pkg/authserver/upstream/oauth2.go:160-164` and `pkg/authserver/upstream/identity_from_token.go:31-32` reference `cmd/thv-operator/api/v1alpha1.IdentityFromTokenConfig` (the v1beta1 CRD type). That type lands in #5155. Until #5155 merges, those references dangle on this branch in isolation. End-to-end functionality requires #5157 too, which adds the operator-to-runner translation and the `RegisterModifiers()` bootstrap call.
- The `docs/arch/11-auth-server-storage.md` update reframes the previous "Synthesis-mode subjects" section as a 3-priority chain. It also flags a known gap: the controller predicate `SyntheticIdentityUpstreams()` checks only `userInfo == nil` and does not yet account for `IdentityFromToken`, so an upstream with only `IdentityFromToken` configured will still trigger `IdentitySynthesizedActive` until #5159 lands.
- The `synthesize fallback` subtest in `oauth2_test.go` was rewritten to use `NewOAuth2Provider` instead of hand-building a struct literal, aligning it with the rest of the suite.


pkg/authserver/upstream/token_exchange.go
```diff
@@ -7,28 +7,44 @@ import (
...
 // dot-notation paths, and rewrites the response body with standard top-level
 // field names so the oauth2 library can parse them normally.
+//
+// When identityCfg is set, identity extraction runs on the RAW pre-rewrite
+// body before any field relocation occurs. The extracted identity is stored
+// in extractedIdentity and consumed by the caller after Exchange returns.
+// If extraction fails, extractionErr holds the cause so callers can surface
+// operator-actionable diagnostics (path names, type descriptions) without
+// re-reading the response body.
 type tokenResponseRewriter struct {
```
tgrunnagle, line 38: **F1 — Rewriter mutable state lacks single-use / no-concurrent-use doc** (consensus 9/10, MEDIUM)

The struct gained `extractedIdentity` and `extractionErr` fields written inside `RoundTrip` and read by the caller after `Exchange` returns. The current `wrapHTTPClientForTokenExchange` pattern enforces single-use correctness (fresh rewriter per call, read after the synchronous `Exchange` returns), but the contract is implicit — a future change that pools or shares rewriters would silently introduce a data race. Per `.claude/rules/go-style.md` "Concurrency Comments" and "Document Architectural Constraints," add a doc note on the struct stating: each instance handles exactly one token exchange, must not be shared across goroutines or reused, and the read-after-Exchange pattern is the synchronization edge.

*Raised by: concurrency, security, code-quality (3 agents)*

pkg/authserver/upstream/oauth2.go
```diff
@@ -152,7 +152,16 @@ type OAuth2Config struct {
...
 	// golang.org/x/oauth2) and extracts fields using gjson dot-notation paths.
 	// When nil, standard OAuth 2.0 token response parsing is used.
+	// See also: IdentityFromToken for extracting user identity from the same response.
 	TokenResponseMapping *TokenResponseMapping `json:"token_response_mapping,omitempty" yaml:"token_response_mapping,omitempty"`
+
+	// IdentityFromToken extracts user identity from the token-endpoint response
+	// body when the upstream provider includes identity claims there (e.g.,
+	// Snowflake's `username`, Slack's `authed_user.id`). When set, the embedded
+	// auth server skips the userinfo HTTP call entirely. See the CRD type
+	// (cmd/thv-operator/api/v1alpha1.IdentityFromTokenConfig) for the
```
tgrunnagle: **F2 — Wrong API version in CRD reference** (consensus 9/10, MEDIUM)

The sibling CRD types in this feature — `EmbeddedAuthServerConfig`, `TokenResponseMapping`, `UserInfoConfig` — all live in `cmd/thv-operator/api/v1beta1`. `IdentityFromTokenConfig` landing in #5155 will be in `v1beta1` too. The `v1alpha1` package is the deprecated thin wrapper. The same wrong reference appears in `docs/arch/11-auth-server-storage.md` line 77.

*Raised by: docs-and-api*

```suggestion
	// (cmd/thv-operator/api/v1beta1.IdentityFromTokenConfig) for the
```

docs/arch/11-auth-server-storage.md
```diff
@@ -64,19 +64,46 @@ The storage layer implements multiple interfaces from the [fosite](https://githu
...
+
+**Subject format.** Real, stable subject string extracted from the token response body via a gjson dot-notation path (e.g. `username`, `authed_user.id`). For token responses that embed a JWT, the `@upstreamjwt` modifier decodes the payload for further drilling (e.g. `access_token|@upstreamjwt|sub`). The `@upstreamjwt` modifier performs no signature verification — it is intended only for JWTs received directly from the upstream token endpoint over a TLS-authenticated channel. The returned `*Identity` carries `Synthetic = false`. Path semantics and trust-model notes are documented on the runtime config struct `IdentityFromTokenConfig` in `pkg/authserver/upstream/identity_from_token.go`. The corresponding CRD type (`cmd/thv-operator/api/v1alpha1.IdentityFromTokenConfig`) is defined in a sibling PR; operator-to-runner translation of this config lands separately.
```
tgrunnagle: **F2 — Wrong API version in CRD reference** (consensus 9/10, MEDIUM)

Same factual error as `oauth2.go:162`: the sibling CRD types live in `v1beta1`, not `v1alpha1`. Update `cmd/thv-operator/api/v1alpha1.IdentityFromTokenConfig` to `cmd/thv-operator/api/v1beta1.IdentityFromTokenConfig` on this line.

*Raised by: docs-and-api*

pkg/authserver/upstream/oauth2_test.go
```diff
@@ -2137,3 +2139,353 @@ func TestAuthorizationURL_AdditionalAuthorizationParams(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// RefreshTokens must not blow up and must return valid tokens.
+		tokens, err := provider.RefreshTokens(ctx, "old-refresh-token", "")
+		require.NoError(t, err)
+		assert.Equal(t, "refreshed", tokens.AccessToken)
+
+		// Verify that the wrapped client on the refresh path uses nil identityCfg
+		// by inspecting the transport after constructing it the same way RefreshTokens does.
+		_, rewriter := wrapHTTPClientForTokenExchange(provider.httpClient, provider.config.TokenResponseMapping, nil, provider.config.TokenEndpoint)
```
tgrunnagle: **F3 — Refresh-path test verifies the helper, not `RefreshTokens`** (consensus 8/10, MEDIUM)

This assertion calls `wrapHTTPClientForTokenExchange` from the test (with the args the test author *believes* `RefreshTokens` uses) and asserts the returned rewriter is nil. That checks a property of the helper, not `RefreshTokens` — if a future change in `RefreshTokens` passed `provider.config.IdentityFromToken` non-nil to the helper, this test would still pass and the assertion would be silently meaningless. The `assert.Equal(t, "refreshed", tokens.AccessToken)` above also can't distinguish "extraction was skipped" from "extraction ran and we ignored the result."

Consider one of: (a) a tripwire token server that records whether the `username` field was queried against the refresh response, (b) exposing the rewriter `RefreshTokens` actually used for direct inspection, or (c) configure `IdentityFromToken` with a path that would fail extraction on the refresh body and assert `RefreshTokens` did NOT surface that error (proving extraction did not run).

*Raised by: test-coverage, code-quality (2 agents)*

pkg/authserver/upstream/token_exchange_test.go
```diff
@@ -280,3 +289,193 @@ func TestOAuth2Config_Validate_TokenResponseMapping(t *testing.T) {
...
+// mapping that lifts access_token to the top level; the assertion confirms
+// extraction reads the original nested location, not a post-rewrite shape.
+func TestTokenResponseRewriter_IdentityFromRawBody(t *testing.T) {
+	t.Parallel()
+
+	rawBody := `{"authed_user":{"access_token":"x","id":"U1234"},"username":""}`
+	server := httptest.NewServer(tokenEndpointHandler(rawBody))
+	t.Cleanup(server.Close)
+
+	mapping := &TokenResponseMapping{AccessTokenPath: "authed_user.access_token"}
```
tgrunnagle: **F4 — Test doesn't exercise the ordering invariant it claims to verify** (consensus 8/10, MEDIUM)

This fixture configures the mapping to lift `authed_user.access_token` to the top level while extracting `authed_user.id` — but `rewriteTokenResponse` only mutates top-level `access_token`/`token_type`/`refresh_token`/`expires_in`/`scope`, never `authed_user.id`. The test would pass identically whether extraction ran before *or* after the rewrite, so it doesn't actually pin the documented "extract from RAW pre-rewrite body" invariant. A regression that reversed the order would not be caught.

Construct a fixture where the rewrite would clobber the identity field. For example, set `SubjectPath: "token_type"` with `tokenBody` containing `"token_type":"raw-marker"` — `rewriteTokenResponse` unconditionally forces `token_type` to `"Bearer"`, so extraction-pre-rewrite reads `"raw-marker"` while extraction-post-rewrite would read `"Bearer"`.

*Raised by: test-coverage*

pkg/authserver/upstream/oauth2.go
```diff
@@ -152,7 +152,16 @@ type OAuth2Config struct {
...
+	// authoritative trust-model and uniqueness documentation.
+	IdentityFromToken *IdentityFromTokenConfig `json:"identity_from_token,omitempty" yaml:"identity_from_token,omitempty"`
```
tgrunnagle, line 164: **F5 — Constructor WARN "using synthesis mode" now lies when `IdentityFromToken` is configured** (consensus 8/10, MEDIUM)

Adding this field means the existing `NewOAuth2Provider` WARN at `oauth2.go:325` (`"oauth2 upstream has no userinfo configured; using synthesis mode"`, fires whenever `UserInfo == nil`) is now incorrect: operators configuring `IdentityFromToken` will see a startup warning that contradicts the runtime priority chain this PR establishes. The arch-doc update correctly describes the new three-way resolution, but the constructor's operator-facing signal does not. Guard the WARN with `config.UserInfo == nil && config.IdentityFromToken == nil`, or split into a positive log when `IdentityFromToken` is configured.

*Raised by: error-handling*

pkg/authserver/upstream/oidc.go
```diff
@@ -242,27 +242,36 @@ func (p *OIDCProviderImpl) ExchangeCodeForIdentity(
...
 
-	tokens, err := p.exchangeCodeForTokens(ctx, code, codeVerifier)
+	// OIDC resolves identity from the validated ID token, not the token response
+	// body, so the extracted identity return value is intentionally discarded.
+	// Defense-in-depth: warn if the OAuth2 base config carries a non-nil
+	// IdentityFromToken — the field is structurally absent on the OIDC CRD type
+	// today, but a future config-loader bug or hand-built runtime config could
+	// silently drop the operator's intent without this signal.
+	if p.config.IdentityFromToken != nil {
+		slog.Warn("OIDC provider ignoring IdentityFromToken; identity is resolved from the validated ID token")
```
tgrunnagle: **F6 — OIDC defensive WARN is unreachable today and would be misleading if it fired** (consensus 8/10, MEDIUM)

`OIDCProviderImpl`'s constructor (oidc.go:191-195) builds a fresh `OAuth2Config{}` with only `CommonOAuthConfig`/`AuthorizationEndpoint`/`TokenEndpoint` set — it never copies `IdentityFromToken` from any source, so this WARN cannot fire on current code. More importantly, if a future config-loader bug DID propagate `IdentityFromToken` onto the embedded config, the rewriter inside `exchangeCodeForTokens` would still **run extraction** (it gates on `t.identityCfg != nil`), populate `extractedIdentity`, then OIDC's caller would silently discard it — the WARN message implies extraction was skipped, but extraction work happened.

Consider one of: (a) demote to DEBUG since the field is structurally absent on the OIDC CRD path today, (b) per `.claude/rules/go-style.md` "fail loudly on invalid input," move the check into a constructor-time validation that returns an error, or (c) explicitly set `p.config.IdentityFromToken = nil` after the WARN so the embedded rewriter cannot run extraction (matches what the message currently claims).

*Raised by: security, error-handling, code-quality, docs-and-api (4 agents)*

pkg/authserver/upstream/oauth2_test.go
```diff
@@ -2137,3 +2139,353 @@ func TestAuthorizationURL_AdditionalAuthorizationParams(t *testing.T) {
...
+
+		identity, err := provider.ExchangeCodeForIdentity(ctx, "test-code", "", "")
+		require.NoError(t, err)
+		assert.Equal(t, "u1", identity.Subject)
+		assert.Equal(t, "User One", identity.Name)
+		assert.Equal(t, "u1@example.com", identity.Email)
+		assert.NotEmpty(t, identity.Tokens.AccessToken)
+	})
+
+	t.Run("@upstreamjwt modifier resolves identity from JWT-shaped access token", func(t *testing.T) {
```
tgrunnagle, line 2206: **F7 — Add a trust-boundary unit test for `@upstreamjwt`** (consensus 7/10, LOW)

The `@upstreamjwt` modifier intentionally performs no signature verification — this is correct per the trust model (only safe for JWTs received over a TLS-authenticated channel directly from the upstream AS). The trust assumption is documented in `identity_from_token.go:149-153` and the arch doc, but no test pins the *no-verification* behavior as intentional rather than accidental.

Consider adding a sibling subtest that feeds a hand-forged unsigned JWT (any base64-decodable payload, even with a garbage signature segment) and asserts the modifier extracts whatever `sub`/`name` claims are present. A future contributor who tries to "harden" this modifier by adding signature checks would then hit a test failure that forces the trust-model conversation, rather than silently changing semantics.

*Raised by: security*

pkg/authserver/upstream/oauth2.go
```diff
@@ -400,39 +414,71 @@ func (p *BaseOAuth2Provider) buildAuthorizationURL(
...
+	// Priority 1: identityFromToken (configured by operator).
+	if p.config.IdentityFromToken != nil {
+		if exchanged.identity == nil {
+			// The rewriter logged the extraction failure at WARN with the
+			// operator-supplied path. Surface the same error here so callers
+			// can report it without requiring WARN-level log access.
+			if exchanged.extractionErr != nil {
+				return nil, exchanged.extractionErr
+			}
+			// Unreachable in practice: when identity is nil the rewriter always
```
tgrunnagle: **F8 — "Unreachable in practice" overstates an oauth2-library-dependent invariant** (consensus 7/10, LOW)

This branch is unreachable only as long as (a) the oauth2 library makes exactly one token-endpoint request per `Exchange`, (b) the rewriter's `RoundTrip` runs extraction inside the `resp.StatusCode == 200` guard, and (c) extraction always sets exactly one of `extractedIdentity`/`extractionErr`. Today all three hold, but "in practice" reads as absolute. A non-200 success the oauth2 library somehow accepts, or a future library that retries via the same transport, would reach this branch.

Suggest softening to something like: "Defense-in-depth: the rewriter guarantees one of `extractedIdentity` or `extractionErr` is set when the token endpoint returns 200; this branch fires only if that invariant is ever relaxed (e.g., a non-200-success status the oauth2 library accepts)."

*Raised by: concurrency, error-handling, code-quality (3 agents)*

pkg/authserver/upstream/oauth2.go
```diff
@@ -496,21 +542,45 @@ func synthesizeIdentity(tokens *Tokens) (*Identity, error) {
...
+// exchange-level error (network, HTTP, token parsing).
+func (p *BaseOAuth2Provider) exchangeCodeForTokens(
+	ctx context.Context, code, codeVerifier string,
+) (*tokenExchangeResult, error) {
 	if code == "" {
 		return nil, errors.New("authorization code is required")
 	}
 
-	slog.Info("exchanging authorization code for tokens",
+	slog.Debug("exchanging authorization code for tokens",
```
tgrunnagle, line 574: **F9 — Log-level downgrade is asymmetric** (consensus 7/10, LOW)

This INFO→DEBUG fix aligns with `.claude/rules/go-style.md` "silent success — no output at INFO or above for successful operations" — good. But `RefreshTokens` at `oauth2.go:556` on main still uses `slog.Info("refreshing tokens", ...)`, and refresh is invoked far more often than auth-code exchange in normal operation. Either downgrade `RefreshTokens` in this PR for consistency, or call out in the PR description why only this site was changed.

*Raised by: security, error-handling, code-quality (3 agents)*

pkg/authserver/upstream/oauth2_test.go
```diff
@@ -2137,3 +2139,353 @@ func TestAuthorizationURL_AdditionalAuthorizationParams(t *testing.T) {
...
+
+		provider, err := NewOAuth2Provider(config)
+		require.NoError(t, err)
+
+		identity, err := provider.ExchangeCodeForIdentity(ctx, "test-code", "", "")
+		require.NoError(t, err)
+		assert.Equal(t, "u1", identity.Subject)
+		assert.Equal(t, "User One", identity.Name)
+		assert.Equal(t, "u1@example.com", identity.Email)
+		assert.NotEmpty(t, identity.Tokens.AccessToken)
```
tgrunnagle, line 2202: **F10 — Assert `identity.Synthetic == false` in this happy-path test** (consensus 7/10, LOW)

The `Synthetic` field is load-bearing — it gates `UserResolver.ResolveUser` bypass in `callback.go:118/137`. The priority-1 contract is specifically `Synthetic = false` (route through user resolution, write a stable users row). A future refactor that incorrectly set `Synthetic = true` for the `IdentityFromToken` path would not be caught by this test — only the `@upstreamjwt` subtest asserts it.

```suggestion
		assert.Equal(t, "u1@example.com", identity.Email)
		assert.NotEmpty(t, identity.Tokens.AccessToken)
		assert.False(t, identity.Synthetic, "IdentityFromToken path must not produce synthetic identities")
```

*Raised by: test-coverage*

pkg/authserver/upstream/token_exchange_test.go
```diff
@@ -211,8 +219,9 @@ func TestWrapHTTPClientWithMapping_NilMapping(t *testing.T) {
 	t.Parallel()
 
 	original := &http.Client{}
-	result := wrapHTTPClientWithMapping(original, nil, "https://example.com/token")
+	result, rewriter := wrapHTTPClientForTokenExchange(original, nil, nil, "https://example.com/token")
```
tgrunnagle, line 222: **F11 — Stale test function name after helper rename** (consensus 7/10, LOW)

The enclosing function is still `TestWrapHTTPClientWithMapping_NilMapping` (declared just above this hunk), but the body now calls `wrapHTTPClientForTokenExchange`. New tests added in this PR use the `TestWrapHTTPClientForTokenExchange_*` prefix; this one diverges. Per `.claude/rules/go-style.md` "Keep Comments Synchronized With Code" applied to test function names, rename to `TestWrapHTTPClientForTokenExchange_NilMapping` — or consolidate with `TestWrapHTTPClientForTokenExchange_BothNil` further down since they assert essentially the same property.

*Raised by: docs-and-api*

tgrunnagle (commented):
## Multi-agent review — `/dev:pr-review`

Reviewed by 6 specialist agents (security, concurrency, error-handling, test-coverage, docs-and-api, code-quality). Codex cross-review skipped (CLI not installed).

**Recommendation: COMMENT** — no blocking issues. The PR's design is sound, the priority chain is correct and well-tested, and the architecture doc is updated alongside the code. Findings below are quality improvements (doc accuracy, test rigor, log consistency) rather than correctness blockers.

### Consensus findings (11 surviving from ~80 raw)

| # | Finding | Severity | Score |
|---|---------|----------|-------|
| F1 | Rewriter mutable state lacks single-use / no-concurrent-use doc | MEDIUM | 9 |
| F2 | CRD type forward-reference points to `v1alpha1` but sibling types live in `v1beta1` | MEDIUM | 9 |
| F3 | Refresh-path test reconstructs the helper call instead of observing `RefreshTokens` behavior | MEDIUM | 8 |
| F4 | `TestTokenResponseRewriter_IdentityFromRawBody` doesn't exercise the ordering invariant it claims | MEDIUM | 8 |
| F5 | Constructor WARN "using synthesis mode" still fires when `IdentityFromToken` is configured | MEDIUM | 8 |
| F6 | OIDC defensive WARN is currently unreachable and would be misleading if it fired | MEDIUM | 8 |
| F7 | `@upstreamjwt` trust-boundary deserves an explicit unit test pinning unsigned-payload behavior | LOW | 7 |
| F8 | "Unreachable in practice" comment overstates an oauth2-library-dependent invariant | LOW | 7 |
| F9 | INFO→DEBUG downgrade is asymmetric: `RefreshTokens` still uses `slog.Info` | LOW | 7 |
| F10 | Happy-path test does not assert `identity.Synthetic == false` | LOW | 7 |
| F11 | `TestWrapHTTPClientWithMapping_NilMapping` retains stale name after helper rename | LOW | 7 |

### Holistic assessment

**Strengths verified across multiple agents:**
- Pre-rewrite extraction ordering means `SubjectPath` always resolves against the original provider response, immune to future `rewriteTokenResponse` changes.
- Information-disclosure guard is real and tested: extraction errors carry path name + static type description only, never body bytes (regression test in place).
- `sha256("")` collision protection preserved through the new priority chain.
- No downgrade attack on the refresh path: identity is anchored at auth-code time and `RefreshTokens` consumes the cached `UserID` from storage.
- Error wrapping chain holds through `errors.Is(err, ErrIdentityResolutionFailed)` for both extraction-failure and userInfo-failure branches.
- `Synthetic = false` defaults correctly for the priority-1 identity, routing through `UserResolver.ResolveUser`.
- The userinfo-tripwire test pins the early-return invariant.

**Architecture.** This is phase 3 of a 6-PR stack (#5156). Forward references to the CRD type (lands in #5155) and operator translation (#5157) are appropriately gated. The `SyntheticIdentityUpstreams()` predicate gap is correctly flagged as a known forward-looking limitation. The additive `OAuth2Config.IdentityFromToken` field is non-breaking.

**Cross-cutting clusters.** The MEDIUM findings cluster around two themes:
1. *Doc-code synchronization* (F2): a one-line factual error in two locations that future readers will be confused by.
2. *Concurrency-invariant documentation* (F1): the rewriter has grown mutable per-call state without a doc comment stating the single-use lifecycle. The current caller pattern is correct, but the contract is implicit.

The operator-visible log inconsistencies (F5, F6, F9) form a smaller cluster — none affect runtime correctness, but they undermine the architecture doc's claim that operator visibility is a designed property of the new path.

### Dropped findings (< 7 consensus, no action expected)

Validate() doesn't check NamePath/EmailPath gjson syntax; missing OIDC IdentityFromToken WARN test (branch unreachable); Validate() table missing field combinations; missing edge-case test for misconfigured EmailPath at integration layer (covered at unit layer); no concurrent-use test for rewriter (subsumed by F1); missing rewriter-level negative test (subsumed by F4); `tokenExchangeResult` doc missing mutual-exclusion invariant (related to F1); extraction failure may double-log; `RefreshTokens` silently drops identity fields (acceptable trade-off per PR description); `fmt.Errorf %w + %s` loses validator chain depth (errors.Is still works); test helpers duplicated; `identity` vs `extractedIdentity` naming; awkward refresh-path comment; SPDX header format (pre-existing); rewriter mapping-time error not surfaced on refresh; rewriter retained longer than needed.

---
*Generated by `/dev:pr-review` (multi-agent consensus review). Inline comments follow.*

tgrunnagle (dismissed):
The implementation LGTM. The agent-raised issues could be addressed in a separate PR, so approving in case you want to do that.

jhrozek:
> The implementation LGTM. The agent-raised issues could be addressed in a separate PR, so approving in case you want to do that.

Thank you, since there will be more PRs in this stacked series, I am going to include the fixes in the next PR.

pkg/authserver/upstream/token_exchange.go
jhrozek: Documented in eb6ee4ba6 — added a single-use contract paragraph on `tokenResponseRewriter` covering: one exchange per instance, no goroutine sharing, and the read-after-Exchange synchronization edge.

pkg/authserver/upstream/oauth2.go
jhrozek: Corrected to `v1beta1` in eb6ee4ba6.

docs/arch/11-auth-server-storage.md
jhrozek: Corrected to `v1beta1` in eb6ee4ba6.

pkg/authserver/upstream/oauth2_test.go
jhrozek: Reworked using option (c) from your suggestions in eb6ee4ba6. The subtest now configures `SubjectPath: "username"` against a refresh response body that omits `username` — if extraction ran on the refresh path it would fail, and `RefreshTokens` would error. `require.NoError` proves extraction was skipped. Renamed to `"IdentityFromToken configured does not cause RefreshTokens to fail when identity field is absent"`.

pkg/authserver/upstream/token_exchange_test.go
jhrozek: Rewrote the fixture using your suggested approach in eb6ee4ba6 — `SubjectPath: "token_type"` against a body containing `"token_type":"raw-marker"`. Because `rewriteTokenResponse` unconditionally overwrites `token_type` to `"Bearer"`, the assertion `Subject == "raw-marker"` can only hold if extraction ran on the pre-rewrite body.

pkg/authserver/upstream/oauth2.go
jhrozek: Guarded the WARN in eb6ee4ba6 — the `NewOAuth2Provider` condition is now `if config.UserInfo == nil && config.IdentityFromToken == nil`, so it only fires when neither identity surface is configured.

pkg/authserver/upstream/oidc.go
jhrozek: Demoted to `slog.Debug` in eb6ee4ba6 (option (a) from your suggestions). Picked DEBUG over the stricter options to keep this purely a defense-in-depth tripwire — the OIDC CRD path can't reach this code today, and turning a non-existent failure mode into a startup error felt heavy-handed. Happy to revisit toward (b) or (c) if you'd prefer.

pkg/authserver/upstream/oauth2_test.go
jhrozek: Added in eb6ee4ba6 — new subtest `"@upstreamjwt modifier accepts hand-forged unsigned JWT"` feeds a hand-built JWT with `"INVALIDSIG"` as the signature segment and asserts the modifier extracts the `sub` claim regardless. A future PR that adds signature verification will hit this test and force the trust-model conversation.

pkg/authserver/upstream/oauth2.go
jhrozek: Rewrote the comment in eb6ee4ba6 using your suggested wording — now reads as defense-in-depth and names the specific invariants (single token-endpoint request per `Exchange`, extraction inside the 200 guard, exactly one of `extractedIdentity`/`extractionErr` set) that this branch would catch if relaxed.

pkg/authserver/upstream/oauth2.go
jhrozek: Downgraded `RefreshTokens` `"refreshing tokens"` log to `slog.Debug` in eb6ee4ba6 for consistency with the success-side log and the silent-success-at-INFO rule.

pkg/authserver/upstream/oauth2_test.go
jhrozek: Applied your suggested assertion in eb6ee4ba6 — added `assert.False(t, identity.Synthetic, "IdentityFromToken path must not produce synthetic identities")` to the happy-path subtest.

pkg/authserver/upstream/token_exchange_test.go
jhrozek: Renamed to `TestWrapHTTPClientForTokenExchange_NilMapping` in eb6ee4ba6. Kept it as a standalone test rather than consolidating with `_BothNil` because the two assert different paths (this one passes nil mapping only; `_BothNil` passes both nils).

tgrunnagle (approved):

PR #5232, opened 2026-05-08
Tolerate spec-violating list methods on backend init

### LINKED ISSUE #5231: vMCP backend init fails fatally when server advertises resources capability without resources/list
## Bug description

vMCP's per-session backend bootstrap (`initAndQueryCapabilities` in [`pkg/vmcp/session/internal/backend/mcp_session.go`](https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/session/internal/backend/mcp_session.go#L346-L400)) treats any error from `resources/list` (and by symmetry `prompts/list`) as fatal. If a backend advertises the `resources` capability in its `initialize` response but does not actually implement `resources/list`, vMCP refuses to initialise the backend at all and silently drops every tool from that backend out of `tools/list`.

The ergonomic outcome: a user goes through the entire OAuth chain successfully, then has zero visibility that anything failed — the backend's tools just don't appear.

We've hit this against Atlassian's hosted Rovo MCP server (`https://mcp.atlassian.com/v1/mcp`), which advertises `capabilities.resources` in its `initialize` response but responds to `resources/list` with JSON-RPC `-32601 Method not found`. The same pattern is plausible for any third-party MCP server we don't control — we shouldn't require server-side spec compliance to use a backend at all.

## Steps to reproduce

1. Configure a `VirtualMCPServer` with at least one backend that advertises `resources` in its `initialize` response but doesn't implement `resources/list`. (Atlassian Rovo via `MCPServerEntry` + `upstreamInject` is one real-world example.)
2. Complete the OAuth flow for the upstream so the user has a valid token for that backend.
3. Open an MCP session against vMCP.

## Expected behavior

The backend initialises successfully and its tools appear in `tools/list`. Resources/prompts are absent (because the backend doesn't actually offer any), and a WARN log records that the backend advertised the capability but didn't honour the list method.

## Actual behavior

vMCP logs:

```
{\"level\":\"WARN\",\"msg\":\"Failed to initialise backend for session; continuing without it\",\"backendID\":\"atlassian-rovo\",\"backendName\":\"atlassian-rovo\",\"error\":\"failed to initialise backend atlassian-rovo: list resources failed: method not found: Method not found\"}
```

The backend's tools never appear in the session. The user has no UI signal that anything went wrong.

The fatal error originates here:

- [`pkg/vmcp/session/internal/backend/mcp_session.go:387-389`](https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/session/internal/backend/mcp_session.go#L387-L389)

  ```go
  resResult, listErr := c.ListResources(ctx, mcp.ListResourcesRequest{})
  if listErr != nil {
      return nil, fmt.Errorf(\"list resources failed: %w\", listErr)
  }
  ```

  Same shape at [`prompts/list`](https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/session/internal/backend/mcp_session.go#L402-L406) (lines 402-406) and [`tools/list`](https://github.com/stacklok/toolhive/blob/main/pkg/vmcp/session/internal/backend/mcp_session.go#L369-L373) (lines 369-373).

The doubled \"method not found: Method not found\" string is from [`mcp-go`'s `JSONRPCErrorDetails.AsError()`](https://github.com/mark3labs/mcp-go/blob/v0.49.0/mcp/errors.go#L86-L132) wrapping the sentinel `mcp.ErrMethodNotFound` (\"method not found\") with the upstream message (\"Method not found\"). Practical implication: callers can detect this case via `errors.Is(err, mcp.ErrMethodNotFound)`.

## Suggested fix

In `initAndQueryCapabilities`, treat `mcp.ErrMethodNotFound` from `resources/list` and `prompts/list` as \"the backend has no resources/prompts\" rather than as a fatal init error. Tools is intentionally left strict: a backend whose `tools/list` is unavailable is not useful to surface at all.

Sketch:

```go
if serverCaps.Resources != nil {
    resResult, listErr := c.ListResources(ctx, mcp.ListResourcesRequest{})
    switch {
    case errors.Is(listErr, mcp.ErrMethodNotFound):
        slog.Warn(\"backend advertised resources capability but does not implement resources/list\",
            \"backendID\", target.WorkloadID)
    case listErr != nil:
        return nil, fmt.Errorf(\"list resources failed: %w\", listErr)
    default:
        for _, r := range resResult.Resources { /* existing append */ }
    }
}
// same shape for prompts/list
```

## Acceptance criteria

- [ ] When a backend advertises `capabilities.resources` in `initialize` but `resources/list` returns JSON-RPC error `-32601` (`mcp.ErrMethodNotFound`), the backend initialises successfully with an empty resource set; its tools are reachable via `tools/list` on the vMCP session.
- [ ] Same behaviour for `prompts/list`: a `-32601` from a backend that advertised `prompts` does not abort backend init.
- [ ] `tools/list` failures remain fatal (a backend with no working tool surface is not useful to expose).
- [ ] Any non-`-32601` error from `resources/list` or `prompts/list` (timeout, transport error, other JSON-RPC error codes) still aborts init — we are not silencing arbitrary failures, only the specific spec-violation case where the server contradicts its own advertised capabilities.
- [ ] A WARN-level log line records the spec violation, including the backend ID and which list method was missing, so operators can surface a bug to the upstream server author. Do not log at ERROR — this is a recoverable, expected-in-the-wild condition.
- [ ] Unit tests in `pkg/vmcp/session/internal/backend/` cover:
  - resources/list returns `-32601` after server advertised `resources` → init succeeds, no resources, WARN logged
  - prompts/list returns `-32601` after server advertised `prompts` → init succeeds, no prompts, WARN logged
  - tools/list returns `-32601` → init still fails (regression guard)
  - resources/list returns a non-`-32601` JSON-RPC error → init still fails (regression guard)
- [ ] Manual verification with the Atlassian Rovo MCP server (`https://mcp.atlassian.com/v1/mcp`) wired up via `MCPServerEntry` + `upstreamInject`: after OAuth completes, `tools/list` on the vMCP session returns the workload-prefixed Atlassian tools (e.g. `atlassian-rovo_<tool>`).

## Environment

- ToolHive version: `main` (reproduced on the `gateway-5fdf997cdc-*` image running in `arn:aws:eks:us-east-1:781189302813:cluster/rovo-demo3-toolhive-dev`)
- mcp-go: v0.49.0

## Additional context

- The same fragility class is worth a quick audit elsewhere in the session bootstrap: any place that derives \"the server supports method X\" from the `initialize` response and then unconditionally calls X. `prompts/list` is the obvious sibling and is included above. Subscriptions / completions paths are worth a glance.
- This is purely a vMCP-side fix; we cannot wait for upstream MCP servers to become spec-compliant before users can call their tools through vMCP.

tgrunnagle:
## Summary

A vMCP backend that advertises `capabilities.resources` (or `capabilities.prompts`) in its `initialize` response but does not actually implement `resources/list` (or `prompts/list`) currently fails per-session backend init outright. Every tool from that backend silently disappears from `tools/list`, leaving users with no signal that anything went wrong — a real problem against third-party servers like Atlassian's Rovo MCP that we don't control. This PR makes the per-session bootstrap tolerate that specific spec violation while keeping all other failure modes fatal.

In `initAndQueryCapabilities`, a JSON-RPC `-32601 Method not found` from `resources/list` or `prompts/list` is now treated as "the backend has no resources/prompts" rather than a fatal init error, with a WARN log recording the spec violation. `tools/list` errors and any non-`-32601` error from the resources/prompts list calls remain fatal.

Closes #5231

### `pkg/vmcp/session/internal/backend/mcp_session.go`
- Detect `mcp.ErrMethodNotFound` from `resources/list` and `prompts/list` via `errors.Is` and recover with an empty result set instead of returning an error.
- Emit a WARN log on recovery, including `backendID` and the offending list method, so operators can flag the upstream bug without ERROR-level noise.
- Leave `tools/list` strict (a backend with no tool surface is not useful to expose) and leave non-`-32601` list errors fatal (we are not silencing arbitrary failures).

### `pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go` (new)
- Table-driven tests covering each acceptance-criteria scenario via a fake streamable-http MCP backend.

## Implementation Details

- The `-32601` detection relies on `mcp-go`'s `JSONRPCErrorDetails.AsError()` wrapping the sentinel `mcp.ErrMethodNotFound`, so `errors.Is(listErr, mcp.ErrMethodNotFound)` is the correct discriminator across upstream message variations (e.g., the doubled `"method not found: Method not found"` string seen in production).
- The recovery is scoped narrowly to the spec-violation case. Transport errors, timeouts, and other JSON-RPC error codes still abort init, preserving visibility into genuine problems.

## Testing

- New unit tests in `pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go` cover the acceptance criteria:
  - `resources/list` returns `-32601` after the server advertised `resources` -> init succeeds with no resources, tools remain reachable.
  - `prompts/list` returns `-32601` after the server advertised `prompts` -> init succeeds with no prompts, tools remain reachable.
  - Both list methods return `-32601` simultaneously -> init still succeeds, tools remain reachable.
  - `tools/list` returns `-32601` -> init still fails (regression guard).
  - `resources/list` returns `-32601` `INTERNAL_ERROR` (non-`-32601`) -> init still fails (regression guard).
  - `prompts/list` returns `INVALID_PARAMS` (non-`-32601`) -> init still fails (regression guard).
- `task test` (unit tests) and `task lint-fix` (linting) run clean on the touched package.

## Additional Notes

- Manual verification against the Atlassian Rovo MCP server (`https://mcp.atlassian.com/v1/mcp`) called out in the issue's acceptance criteria has not been performed in this PR; the unit tests exercise the same code path with a fake backend that produces the equivalent `-32601` response. Reviewers who have the Rovo wiring available should confirm `tools/list` returns the workload-prefixed Atlassian tools after OAuth completes.
- The same fragility class is worth a follow-up audit elsewhere in the session bootstrap (subscriptions, completions) — anywhere we derive "the server supports method X" from the `initialize` response and then unconditionally call X. Out of scope for this PR.

### DIFF
--- a/pkg/vmcp/session/internal/backend/mcp_session.go
+++ b/pkg/vmcp/session/internal/backend/mcp_session.go
@@ -5,6 +5,7 @@ package backend
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
@@ -385,40 +386,62 @@ func initAndQueryCapabilities(
 
 	if serverCaps.Resources != nil {
 		resResult, listErr := c.ListResources(ctx, mcp.ListResourcesRequest{})
-		if listErr != nil {
+		switch {
+		case errors.Is(listErr, mcp.ErrMethodNotFound):
+			// Spec violation: backend advertised the resources capability in its
+			// initialize response but does not implement resources/list. Recover
+			// with an empty resource set so the backend's tools remain usable
+			// instead of failing init outright. See issue #5231.
+			slog.Warn("backend advertised resources capability but does not implement resources/list",
+				"backendID", target.WorkloadID,
+				"method", "resources/list",
+			)
+		case listErr != nil:
 			return nil, fmt.Errorf("list resources failed: %w", listErr)
-		}
-		for _, r := range resResult.Resources {
-			caps.Resources = append(caps.Resources, vmcp.Resource{
-				URI:         r.URI,
-				Name:        r.Name,
-				Description: r.Description,
-				MimeType:    r.MIMEType,
-				BackendID:   target.WorkloadID,
-			})
+		default:
+			for _, r := range resResult.Resources {
+				caps.Resources = append(caps.Resources, vmcp.Resource{
+					URI:         r.URI,
+					Name:        r.Name,
+					Description: r.Description,
+					MimeType:    r.MIMEType,
+					BackendID:   target.WorkloadID,
+				})
+			}
 		}
 	}
 
 	if serverCaps.Prompts != nil {
 		promptsResult, listErr := c.ListPrompts(ctx, mcp.ListPromptsRequest{})
-		if listErr != nil {
+		switch {
+		case errors.Is(listErr, mcp.ErrMethodNotFound):
+			// Spec violation: backend advertised the prompts capability in its
+			// initialize response but does not implement prompts/list. Recover
+			// with an empty prompt set so the backend's tools remain usable
+			// instead of failing init outright. See issue #5231.
+			slog.Warn("backend advertised prompts capability but does not implement prompts/list",
+				"backendID", target.WorkloadID,
+				"method", "prompts/list",
+			)
+		case listErr != nil:
 			return nil, fmt.Errorf("list prompts failed: %w", listErr)
-		}
-		for _, p := range promptsResult.Prompts {
-			args := make([]vmcp.PromptArgument, len(p.Arguments))
-			for j, a := range p.Arguments {
-				args[j] = vmcp.PromptArgument{
-					Name:        a.Name,
-					Description: a.Description,
-					Required:    a.Required,
+		default:
+			for _, p := range promptsResult.Prompts {
+				args := make([]vmcp.PromptArgument, len(p.Arguments))
+				for j, a := range p.Arguments {
+					args[j] = vmcp.PromptArgument{
+						Name:        a.Name,
+						Description: a.Description,
+						Required:    a.Required,
+					}
 				}
+				caps.Prompts = append(caps.Prompts, vmcp.Prompt{
+					Name:        p.Name,
+					Description: p.Description,
+					Arguments:   args,
+					BackendID:   target.WorkloadID,
+				})
 			}
-			caps.Prompts = append(caps.Prompts, vmcp.Prompt{
-				Name:        p.Name,
-				Description: p.Description,
-				Arguments:   args,
-				BackendID:   target.WorkloadID,
-			})
 		}
 	}
 
--- a/pkg/vmcp/session/internal/backend/mcp_session.go
+++ b/pkg/vmcp/session/internal/backend/mcp_session.go
@@ -388,12 +388,14 @@ func initAndQueryCapabilities(
 		resResult, listErr := c.ListResources(ctx, mcp.ListResourcesRequest{})
 		switch {
 		case errors.Is(listErr, mcp.ErrMethodNotFound):
-			// Spec violation: backend advertised the resources capability in its
-			// initialize response but does not implement resources/list. Recover
-			// with an empty resource set so the backend's tools remain usable
-			// instead of failing init outright. See issue #5231.
+			// Tolerate JSON-RPC -32601 here so a backend that advertises the
+			// resources capability but does not implement resources/list (e.g.
+			// Atlassian Rovo, see #5231) still contributes its tools instead of
+			// being dropped. HTTP-level method absence is intentionally fatal.
 			slog.Warn("backend advertised resources capability but does not implement resources/list",
 				"backendID", target.WorkloadID,
+				"name", target.WorkloadName,
+				"baseURL", target.BaseURL,
 				"method", "resources/list",
 			)
 		case listErr != nil:
@@ -415,12 +417,14 @@ func initAndQueryCapabilities(
 		promptsResult, listErr := c.ListPrompts(ctx, mcp.ListPromptsRequest{})
 		switch {
 		case errors.Is(listErr, mcp.ErrMethodNotFound):
-			// Spec violation: backend advertised the prompts capability in its
-			// initialize response but does not implement prompts/list. Recover
-			// with an empty prompt set so the backend's tools remain usable
-			// instead of failing init outright. See issue #5231.
+			// Tolerate JSON-RPC -32601 here so a backend that advertises the
+			// prompts capability but does not implement prompts/list (e.g.
+			// Atlassian Rovo, see #5231) still contributes its tools instead of
+			// being dropped. HTTP-level method absence is intentionally fatal.
 			slog.Warn("backend advertised prompts capability but does not implement prompts/list",
 				"backendID", target.WorkloadID,
+				"name", target.WorkloadName,
+				"baseURL", target.BaseURL,
 				"method", "prompts/list",
 			)
 		case listErr != nil:
--- a/pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go
+++ b/pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go
@@ -1,0 +1,380 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package backend
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"testing"
+
+	mcpclient "github.com/mark3labs/mcp-go/client"
+	mcptransport "github.com/mark3labs/mcp-go/client/transport"
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/stacklok/toolhive/pkg/vmcp"
+)
+
+// fakeBackend is a minimal JSON-RPC over streamable-HTTP fake server. It lets
+// each test declare exactly which capabilities to advertise in the initialize
+// response and how to respond to each list method, including JSON-RPC error
+// codes. This is the surface relied on by initAndQueryCapabilities, and
+// constructing it directly avoids the mcp-go server's automatic capability
+// inference (which would not let us simulate the spec-violation case where a
+// capability is advertised but the corresponding list method returns -32601).
+type fakeBackend struct {
+	t *testing.T
+
+	// Capabilities to advertise in the initialize response.
+	advertiseTools     bool
+	advertiseResources bool
+	advertisePrompts   bool
+
+	// Optional overrides for list responses. If nil, the server returns an
+	// empty list of the corresponding type. To inject a JSON-RPC error,
+	// populate listResourcesErr / listPromptsErr / listToolsErr.
+	listResourcesErr *jsonRPCError
+	listPromptsErr   *jsonRPCError
+	listToolsErr     *jsonRPCError
+
+	// Tools/resources/prompts to return when the corresponding list method
+	// is not configured to error.
+	tools     []mcp.Tool
+	resources []mcp.Resource
+	prompts   []mcp.Prompt
+
+	// methodCalls counts how many times each method was invoked, so tests can
+	// assert that (e.g.) tools/list was reached after a recoverable
+	// resources/list failure.
+	mu          sync.Mutex
+	methodCalls map[string]int
+}
+
+type jsonRPCError struct {
+	code    int
+	message string
+}
+
+// newFakeBackend wires up an httptest.Server speaking just enough JSON-RPC for
+// the streamable-HTTP transport to complete Initialize and the three list
+// methods. It returns the server URL to point a client at.
+func newFakeBackend(t *testing.T, fb *fakeBackend) string {
+	t.Helper()
+	fb.t = t
+	fb.methodCalls = make(map[string]int)
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/mcp", fb.handle)
+
+	ts := httptest.NewServer(mux)
+	t.Cleanup(ts.Close)
+	return ts.URL + "/mcp"
+}
+
+func (f *fakeBackend) callCount(method string) int {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return f.methodCalls[method]
+}
+
+// handle implements the JSON-RPC subset needed for backend init. The
+// streamable-HTTP transport sends POST requests with Accept:
+// application/json, text/event-stream — we always reply with
+// application/json since we only need single-shot responses.
+func (f *fakeBackend) handle(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		// The streamable-HTTP transport opens a GET to listen for server-pushed
+		// notifications. Returning 405 cleanly tells it the server doesn't
+		// support push, which is fine for our tests.
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	body, err := io.ReadAll(r.Body)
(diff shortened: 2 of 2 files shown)


pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go
```diff
@@ -0,0 +1,380 @@
...
+	t.Cleanup(func() { _ = c.Close() })
+	return c
+}
+
+// TestInitAndQueryCapabilities_RecoversFromMethodNotFound verifies the issue
+// #5231 fix: when a backend advertises resources or prompts capability but
+// returns JSON-RPC -32601 to the corresponding list method, init must succeed
+// with an empty capability set rather than aborting and dropping the backend's
+// tools.
+func TestInitAndQueryCapabilities_RecoversFromMethodNotFound(t *testing.T) {
```
lorr1, line 218: **[MEDIUM] Success-path (default switch arm) is unverified.**

The production refactor moves the populate loop into a new `default` arm of the switch. None of the cases here exercise that arm with a non-empty payload — every recovery test injects an error, and every fatal-error test aborts before reaching it. A future regression that inverts a condition or drops a field assignment in the populate loop would not be caught.

Consider adding at least one positive subtable row that returns a non-empty `resources` (and one for `prompts`, ideally with arguments) and asserts field-by-field mapping including `BackendID == target.WorkloadID`. The existing `fakeBackend` already supports this — set `advertiseResources: true` and populate `resources: []mcp.Resource{{...}}`.

pkg/vmcp/session/internal/backend/mcp_session.go
```diff
@@ -385,40 +386,62 @@ func initAndQueryCapabilities(
 
 	if serverCaps.Resources != nil {
 		resResult, listErr := c.ListResources(ctx, mcp.ListResourcesRequest{})
-		if listErr != nil {
+		switch {
+		case errors.Is(listErr, mcp.ErrMethodNotFound):
+			// Spec violation: backend advertised the resources capability in its
+			// initialize response but does not implement resources/list. Recover
+			// with an empty resource set so the backend's tools remain usable
+			// instead of failing init outright. See issue #5231.
```
lorr1: **[LOW] Comment phrasing.** Three concerns converge on this comment block (mirrored in the prompts case at line 418):

1. "Recover with an empty resource set" reads as if a slice is allocated. Actually `caps.Resources` simply remains at its zero value (`nil`) — range over nil and `len` are both zero, so the behavior is correct, but the wording mildly misleads about state.
2. The MCP spec doesn't contain a normative MUST for "if you advertise capability X you must implement X/list." The label "spec violation" is rhetorically reasonable but not directly citable from the spec.
3. The "why" (keep tools usable) is the third sentence in; the first two restate what the surrounding `errors.Is(...)` and `slog.Warn` already convey.

Suggest tightening to one rationale-focused sentence, e.g.:

```suggestion
			// Tolerate -32601 here so a backend that advertises the resources
			// capability but does not implement resources/list (e.g. Atlassian Rovo,
			// see #5231) still contributes its tools instead of being dropped.
```

Apply symmetrically to the prompts block at line 418.

pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go
```diff
@@ -0,0 +1,380 @@
...
+
+	helloTool := mcp.Tool{Name: "hello"}
+
+	tests := []struct {
+		name             string
+		fb               *fakeBackend
+		wantTools        int
+		wantResources    int
+		wantPrompts      int
+		wantToolsCalled  bool
```
lorr1: **[LOW] `wantToolsCalled bool` is redundant and asymmetric.**

Every populated row sets `wantToolsCalled: true` and the assertion at line 305 is hard-coded to 1, so the bool encodes no information beyond "do this assert." Meanwhile resources and prompts use explicit `int` counters (`wantResListCalls`, `wantPromptCalls`).

Suggest replacing with `wantToolsCalls int`, set to 1 in each row, and applying the same pattern as the other counters at the assertion site:

```go
if tc.wantToolsCalls > 0 {
    assert.Equal(t, tc.wantToolsCalls, tc.fb.callCount(string(mcp.MethodToolsList)))
}
```

Makes future cases that assert `tools/list` was *not* called (e.g., when neither `resources` nor `prompts` is advertised) trivial to add.

pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go
```diff
@@ -0,0 +1,380 @@
...
+			target := &vmcp.BackendTarget{
+				WorkloadID:    "fake-backend",
+				WorkloadName:  "fake-backend",
+				BaseURL:       url,
+				TransportType: "streamable-http",
+			}
+
+			_, err := initAndQueryCapabilities(context.Background(), c, target)
+			require.Error(t, err)
+			assert.True(t, strings.Contains(err.Error(), tc.errSubstr),
```
lorr1: **[LOW] Use `assert.Contains` (or `require.ErrorContains`) instead of reinventing it.**

Gives a clearer failure message (testify shows both haystack and needle automatically) and lets you drop the `strings` import.

```suggestion
			assert.ErrorContains(t, err, tc.errSubstr)
```

pkg/vmcp/session/internal/backend/mcp_session.go
```diff
@@ -385,40 +386,62 @@ func initAndQueryCapabilities(
...
+			slog.Warn("backend advertised resources capability but does not implement resources/list",
+				"backendID", target.WorkloadID,
+				"method", "resources/list",
+			)
```
lorr1, line 400: **[LOW] WARN log lacks workload name and base URL.**

When this WARN fires in production it's the operator's primary breadcrumb that a backend silently produced an empty resource/prompt set. `backendID` (= `target.WorkloadID`) requires an extra ID-to-name lookup to correlate to a specific MCP server. Adding `name` and `baseURL` makes the log directly greppable. Same applies to the prompts WARN at line 422.

```suggestion
			slog.Warn("backend advertised resources capability but does not implement resources/list",
				"backendID", target.WorkloadID,
				"name", target.WorkloadName,
				"baseURL", target.BaseURL,
				"method", "resources/list",
			)
```

pkg/vmcp/session/internal/backend/mcp_session.go
```diff
@@ -385,40 +386,62 @@ func initAndQueryCapabilities(
 
 	if serverCaps.Resources != nil {
 		resResult, listErr := c.ListResources(ctx, mcp.ListResourcesRequest{})
-		if listErr != nil {
+		switch {
+		case errors.Is(listErr, mcp.ErrMethodNotFound):
```
lorr1, line 390: **[LOW] `errors.Is` will not match HTTP-level method-missing variants.**

Confirmed `mcp-go@v0.49.0`'s `JSONRPCErrorDetails.AsError()` wraps `mcp.ErrMethodNotFound` for the `-32601` JSON-RPC envelope, so this discriminator is correct for the Atlassian Rovo case (which returns a proper JSON-RPC error response).

Caveat worth noting: some real backends behind reverse proxies or API gateways surface "this method isn't here" as HTTP 404 / 405 / 501 rather than a JSON-RPC `-32601` envelope. Those return non-sentinel errors from the transport layer, so they will fall through to the fatal `case listErr != nil` arm under this PR. The narrow scope is defensible (the issue specifically calls out -32601), but the next operator hitting an HTTP-404 backend will wonder why their case wasn't recovered.

No change required for this PR. Worth either (a) a follow-up issue to enumerate which transport-level signals should also be treated as method-missing, or (b) a half-line in the recovery comment noting that only JSON-RPC -32601 envelopes are recovered (HTTP-level method absence is intentionally fatal).

lorr1 (commented):
## Multi-agent review summary

**Recommendation: COMMENT** (non-blocking).

The fix is correct, well-scoped, and end-to-end safe. `errors.Is(listErr, mcp.ErrMethodNotFound)` is the right discriminator against the `mcp-go` v0.49.0 sentinel (`JSONRPCErrorDetails.AsError()` wraps it via `fmt.Errorf("%w: ...")`, so `errors.Is` matches both bare and wrapped variants). Switch ordering is nil-safe. `tools/list` strictness is preserved. Empty `caps.Resources`/`caps.Prompts` flow correctly through `pkg/vmcp/session/factory.go` and the aggregator merge — a downstream client `resources/list` against vMCP simply returns an empty merged set.

Four specialist reviewers (Go correctness, MCP protocol, test coverage, general code quality) consulted. Codex cross-review was attempted but timed out; not blocking.

### Summary of inline findings

| # | Severity | Theme |
|---|----------|-------|
| 1 | MEDIUM | Switch `default` arm (the success-path populate loop) is unverified |
| 2 | LOW | Comment phrasing ("spec violation" / "empty resource set") |
| 3 | LOW | `wantToolsCalled bool` redundant with int counters |
| 4 | LOW | `assert.True(strings.Contains(...))` should be `assert.Contains` |
| 5 | LOW | WARN log lacks workload name and base URL |
| 6 | LOW | `errors.Is` won't match HTTP-level method-missing variants (404/501) |

### Out-of-scope / informational

- Pre-existing: list calls do not loop on `nextCursor` — file a follow-up issue if pagination matters.
- PR description references `mcp-go v0.43.2`; `go.mod` actually pins `v0.49.0`. Sentinel and wrapping semantics are identical, so the code is correct — doc nit only.
- WARN may be noisy under reconnect churn; rate-limit per `(backendID, method)` only if it shows up in practice.
- vMCP does not currently forward `notifications/resources/list_changed`, so the empty-recovery path does not create phantom subscriptions today.

pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go
tgrunnagle: Addressed in a5648e9f32a8e0373a9d0eb537dd1c9a4d3b55e0.

pkg/vmcp/session/internal/backend/mcp_session.go
tgrunnagle: Addressed in a5648e9f32a8e0373a9d0eb537dd1c9a4d3b55e0.

pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go
tgrunnagle: Addressed in a5648e9f32a8e0373a9d0eb537dd1c9a4d3b55e0.

pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go
tgrunnagle: Addressed in a5648e9f32a8e0373a9d0eb537dd1c9a4d3b55e0.

pkg/vmcp/session/internal/backend/mcp_session.go
tgrunnagle: Addressed in a5648e9f32a8e0373a9d0eb537dd1c9a4d3b55e0.

pkg/vmcp/session/internal/backend/mcp_session.go
tgrunnagle: Partially addressed in a5648e9f32a8e0373a9d0eb537dd1c9a4d3b55e0: the recovery comment now reads "Tolerate JSON-RPC -32601 here…HTTP-level method absence is intentionally fatal" (option (b) from your suggestion). A follow-up issue to enumerate which transport-level signals (HTTP 404/405/501) should also be treated as method-missing is left out of this PR's scope.

tgrunnagle: Created https://github.com/stacklok/toolhive/issues/5242 to track

jhrozek (approved):

Issue #5224, opened 2026-05-07
Add `requiredClientScopes` to embedded auth server for DCR scope baseline


jhrozek:
## Background

ToolHive's embedded OAuth/OIDC authorization server implements RFC 7591 Dynamic Client Registration (DCR). When a client registers, ToolHive validates each requested scope is in `scopes_supported` and registers the client with exactly that set (`pkg/authserver/server/registration/dcr.go:264-305`). The registered scope set is then used by fosite to validate later `/oauth/authorize` requests.

This is RFC-correct, but brittle against buggy DCR clients. Notably, [anthropics/claude-code#4540](https://github.com/anthropics/claude-code/issues/4540) (open since ~Aug 2025): Claude Code's MCP OAuth client narrows the DCR `scope` field to `openid`, then later requests `offline_access` at `/authorize`. ToolHive correctly responds:

```
invalid_scope: The OAuth 2.0 Client is not allowed to request scope 'offline_access'
```

The asymmetry between well-behaved DCR clients and clients that narrow the scope field can be illustrated as:

| Client behavior | DCR-registered scope | `/authorize` requests | Result |
|---|---|---|---|
| DCR sends full scope set | `openid profile email offline_access` | `openid profile email offline_access` | OK |
| DCR narrows to `openid` only | `openid` | `openid offline_access` | FAIL |
| DCR narrows, but client uses pre-registered `client_id` | `openid profile email offline_access` | `openid offline_access` | OK |

The current workaround — pre-register a public OAuth client at `POST /oauth/register` and ask every affected user to add the resulting `client_id` to their MCP client config — is per-user friction that doesn't scale across a multi-user deployment, and it hands out a shared OAuth client identity (audit attribution still works because the user-facing identity comes from the upstream provider's session, but the shape is leaky).

## Proposal

Add an operator-configurable field `requiredClientScopes` on `EmbeddedAuthServerConfig` that declares a baseline scope set unioned into every DCR registration:

```yaml
authServerConfig:
  issuer: https://auth.example.com
  requiredClientScopes:
    - openid
    - profile
    - email
    - offline_access
  upstreamProviders: [...]
```

**Semantics:**

- DCR request with empty `scope` → register with `requiredClientScopes` (or hardcoded `DefaultScopes` if unset, preserving current behavior).
- DCR request with non-empty `scope` → register with `union(requested, requiredClientScopes)`, deduped, validated as ⊆ `scopesSupported`.
[...]

PR #5233, opened 2026-05-08
Allow operators to inject baseline scopes into DCR registrations

jhrozek:
## Summary

Some MCP clients (notably Claude Code) DCR-register with a narrowed `scope` field at `/oauth/register` but later request a wider set at `/oauth/authorize`. fosite rejects those requests with `invalid_scope` because the registered client's scope set does not include the requested scopes. RFC 7591 §3.2.1 explicitly permits the AS to replace requested client metadata; the embedded auth server now uses that to expand the registered scope set.

- Add `EmbeddedAuthServerConfig.baselineClientScopes` to the CRD — operators declare scopes that must be in every registered client's scope set.
- Plumb the field through `RunConfig` → runtime `Config` → `AuthorizationServerConfig` so the DCR handler can read it at request time.
- DCR handler unions the baseline into each new registration; logs a WARN when expansion actually happens (no-op short-circuited via `slices.Equal`).
- Validate `baseline ⊆ scopesSupported` at three layers (CRD-loaded `RunConfig`, `applyDefaults`, `validateParams`) so misconfiguration fails loudly at startup.
- Integration regression test reproduces the Claude Code repro pattern (anthropics/claude-code#4540) end-to-end.

Refs #5224

## Type of change

- [x] New feature

## Test plan

- [x] E2E tests (`task test-e2e`) — covered by the new integration test in `test/integration/authserver/`

Manual: end-to-end deploy still pending — opening as draft for that.

## API Compatibility

`baselineClientScopes` is a new optional field with `omitempty` and a nil-safe deepcopy. Existing CRs without the field decode as `nil`; the validator early-returns and behavior is unchanged.

Yes — operators of \`MCPExternalAuthConfig\` and \`VirtualMCPServer\` resources gain a new optional \`spec.embedded.baselineClientScopes\` field. When set, every DCR-registered client's scope set is expanded to include these scopes regardless of what the client requested at \`/oauth/register\`.

- Defense-in-depth validation runs at three layers; the wire-format-boundary check (\`RunConfig.Validate\`) is the earliest, the \`Config.applyDefaults\` guard catches direct callers, and \`validateParams\` is a final safety net.
- The \`unionScopes\` helper preserves requested-scope order; the \`slices.Equal\` no-op detection in the handler depends on that invariant.
- \`MaxItems=10\` on the CRD field bounds operator misuse; we considered an item-level \`Pattern\` marker and decided runtime validation suffices for v1.

## Large PR Justification

- the CRDs got regenerated


tgrunnagle (commented):
## Multi-agent review summary

Five specialist reviewers (OAuth/security, Kubernetes operator/CRD, Go correctness, test coverage, general code quality) plus a consensus-scoring pass. Codex cross-review was skipped because the `codex` CLI isn't installed locally.

**Overall: COMMENT — no blockers, nice piece of work.** The defense-in-depth design, the three-layer validation, and the explicit regression test for anthropics/claude-code#4540 are all strong. The notes below cluster into trust-boundary documentation, observability, defense-in-depth completeness, CRD validation hardening, and operator UX — each addressable without rework.

### Consensus summary

| # | Finding | Severity | Score |
|---|---------|----------|-------|
| F1 | Operator-configured baseline silently widens every DCR client's registered scope set — trust boundary only weakly documented | MEDIUM | 8/10 |
| F2 | WARN log includes attacker-controlled `client_name` (log-injection risk) | LOW | 7/10 |
| F3 | WARN log fires per registration when expansion is the documented happy path | LOW | 7/10 |
| F4 | `BaselineClientScopes` slice flows by reference through all layers — copy-before-mutate rule | LOW | 7/10 |
| F5 | No direct unit test for the third defense-in-depth gate (`validateParams`) | MEDIUM | 8/10 |
| F6 | Negative integration subtest's `400` branch doesn't assert `invalid_scope` | MEDIUM | 7/10 |
| F7 | No operator status condition for baseline misconfig — surfaces only as CrashLoopBackOff | MEDIUM | 7/10 |
| F8 | CRD field has no `items:MinLength=1` — empty/whitespace entries pass admission | LOW | 8/10 |
| F9 | No user-facing docs / Helm examples for the new operator knob | MEDIUM | 7/10 |
| F10 | `ValidateScopeSubset` doc enumerates 2 of 3 defense-in-depth gates | LOW | 7/10 |
| F11 | `RegisterClientHandler` response comment can be misread as implying re-validation of unioned set | LOW | 7/10 |
| F12 | `Config.Validate()` (sibling helper at line 472) doesn't run the subset check — three-layer claim slightly overstated | LOW | 7/10 |

Detailed findings are inline. The dropped findings (consensus < 7, mostly stylistic / single-agent informational) are listed in the run log and not surfaced here to keep the review focused.

### Holistic assessment

- **Backwards compatibility**: safe. Empty/nil `BaselineClientScopes` is bit-exact no-op via `len(...) > 0` guard, and the field is `+optional` with `omitempty`.
- **Plumbing symmetry**: the CRD → `controllerutil` → `RunConfig` → `Config` → `AuthorizationServerParams` → `AuthorizationServerConfig` flow exactly mirrors `ScopesSupported` and `AllowedAudiences`. Easy to follow.
- **CRD generation**: all four chart copies (`files/×2`, `templates/×2`) and `docs/operator/crd-api.md` are consistent; deepcopy is regenerated.
- **Integration regression test**: excellent — explicit reference to `anthropics/claude-code#4540` and end-to-end coverage of the bug-trigger pattern. The negative subtest catches the obvious privilege-escalation regression; tightening per F6 closes the last gap.
- **Test coverage**: thorough for `unionScopes`, `ValidateScopeSubset`, `RunConfig.Validate`, `Config.applyDefaults`, and the DCR handler path. One gate (`validateParams`) lacks a direct test — see F5.
- **Open design question (non-blocking)**: this baseline mechanism applies globally to every DCR registration on the AS instance. CRD docs warn operators to keep it narrow but the implementation has no per-client filter (e.g. by `software_id`). F1 flags this as documentation, not code.

Review consulted: OAuth/DCR/Security, K8s operator/CRD, Go correctness, test coverage, and general code quality reviewers.

pkg/authserver/server/handlers/dcr.go
```diff
@@ -63,6 +64,28 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
...
+	// Union with the operator-configured scope baseline. RFC 7591 §3.2.1 permits
+	// the AS to replace requested client metadata values during registration; we
+	// use that to expand the registered scope set so a client whose DCR request
+	// narrowed the scope field can still request the baseline at /oauth/authorize.
+	// h.config.BaselineClientScopes is validated at startup to be a subset of
+	// ScopesSupported, so the union is guaranteed to be a subset of advertised
+	// scopes. Operators should keep the baseline narrow (e.g. openid,
+	// offline_access) — every DCR-registered client gains the ability to request
+	// these scopes at /oauth/authorize regardless of what they registered with.
+	if len(h.config.BaselineClientScopes) > 0 {
```
tgrunnagle, line 76: **MEDIUM — F1: Trust boundary documentation**

This block silently widens every DCR-registered client's scope set. If an operator (intentionally or by mistake) places a privileged scope like `admin:read` in `baselineClientScopes` *and* in `scopesSupported`, every public DCR client — Claude Code, Cursor, VS Code, anyone calling /oauth/register — gains the ability to request `admin:read` at /oauth/authorize. The trust boundary is correct (operators own the AS) but only weakly defended in docs ("operators should keep the baseline narrow").

Suggestions, in order of cost:
- Tighten the CRD field doc with an explicit privilege-escalation example so the trust boundary is obvious to operators writing YAML.
- At server startup, log the effective baseline once: `slog.Info("DCR registrations will be auto-granted baseline scopes", "scopes", c.BaselineClientScopes)` so operators see what they enabled.
- Consider a hard allowlist of "safe baseline scopes" (e.g. OIDC standard scopes only) gating what may be configured.

pkg/authserver/server/handlers/dcr.go
```diff
@@ -63,6 +64,28 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
...
+		effective := unionScopes(scopes, h.config.BaselineClientScopes)
+		if !slices.Equal(effective, scopes) {
+			slog.Warn("DCR registered scope set expanded by baseline_client_scopes",
+				"client_name", validated.ClientName,
```
tgrunnagle, line 85: **LOW — F2: Log injection via `client_name`**

`validated.ClientName` is length-capped (256) but not restricted to printable ASCII — unlike `software_id`, which goes through that filter in `validateDCRRequest`. A malicious DCR caller can supply a `client_name` containing newlines, ANSI escapes, or control characters that pollute this now-elevated WARN line and confuse log readers / SIEM correlation. This was tolerable at DEBUG; promoting to WARN amplifies the impact.

Fix: extend `validateDCRRequest` to apply the same printable-ASCII restriction to `client_name` as it does to `software_id`. That fixes it everywhere (DEBUG log path too).

pkg/authserver/server/handlers/dcr.go
```diff
@@ -63,6 +64,28 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
...
+	// narrowed the scope field can still request the baseline at /oauth/authorize.
+	// h.config.BaselineClientScopes is validated at startup to be a subset of
+	// ScopesSupported, so the union is guaranteed to be a subset of advertised
+	// scopes. Operators should keep the baseline narrow (e.g. openid,
+	// offline_access) — every DCR-registered client gains the ability to request
+	// these scopes at /oauth/authorize regardless of what they registered with.
+	if len(h.config.BaselineClientScopes) > 0 {
+		effective := unionScopes(scopes, h.config.BaselineClientScopes)
+		if !slices.Equal(effective, scopes) {
+			slog.Warn("DCR registered scope set expanded by baseline_client_scopes",
```
tgrunnagle: **LOW — F3: WARN level for the documented happy path**

Per `.claude/rules/go-style.md`, WARN is reserved for "non-fatal issues (deprecations, fallback behavior, cleanup failures)." When `baselineClientScopes` is configured, expansion is the *intended* behavior — every Claude-Code-style narrow-DCR registration will fire this WARN, which is noisy and reads like something is wrong when it isn't.

Demote to `slog.Debug`. If you want operator-visible signal that the baseline is in effect, log it once at server startup instead (see F1).

pkg/authserver/runner/embeddedauthserver.go
```diff
@@ -163,6 +171,7 @@ func newEmbeddedAuthServerWithStorage(
 		AuthCodeLifespan:             authCodeLifespan,
 		Upstreams:                    upstreams,
 		ScopesSupported:              cfg.ScopesSupported,
+		BaselineClientScopes:         cfg.BaselineClientScopes,
```
tgrunnagle: **LOW — F4: Pass `slices.Clone` for caller isolation**

`cfg.BaselineClientScopes` flows by reference into `RunConfig` → `Config` → `AuthorizationServerParams` → `AuthorizationServerConfig`, and the DCR handler reads `h.config.BaselineClientScopes` on every request. If any caller (a test, a future hot-reload path) mutates the original slice while a request is in flight, the handler races on a read of changing memory.

`.claude/rules/go-style.md` "Copy Before Mutating Caller Input" recommends a clone; with ≤10 entries it's trivial. The same hardening applies symmetrically to `ScopesSupported` and `AllowedAudiences` in this constructor — consider tightening all three together.

pkg/authserver/server/provider.go
```diff
@@ -189,7 +198,14 @@ func validateParams(cfg *AuthorizationServerParams) error {
...
 	}
-	return validateAllowedAudiences(cfg.AllowedAudiences)
+	if err := validateAllowedAudiences(cfg.AllowedAudiences); err != nil {
+		return err
+	}
+	// Defense-in-depth: re-check the baseline-⊆-scopes_supported invariant.
+	// RunConfig.Validate performs the same check at the operator-supplied
+	// wire-format boundary; this gate covers callers that construct
+	// AuthorizationServerParams programmatically and bypass that path.
+	return registration.ValidateScopeSubset(cfg.BaselineClientScopes, cfg.ScopesSupported, "baseline_client_scopes")
```
tgrunnagle, line 209: **MEDIUM — F5: Add a direct unit test for this gate**

The PR designs three defense-in-depth gates and tests two (`RunConfig.Validate`, `Config.applyDefaults`'s narrow check). This third gate's entire purpose is to catch programmatic callers that bypass YAML loading — exactly the path that would silently regress in a refactor with no coverage.

Add a test in `pkg/authserver/server/provider_test.go` that constructs `AuthorizationServerParams` with a baseline scope absent from `ScopesSupported`, calls `NewAuthorizationServerConfig`, and asserts an error containing `"baseline_client_scopes contains"`. Without it, only the operator YAML path is regression-protected.

test/integration/authserver/authserver_integration_test.go
```diff
@@ -466,6 +467,137 @@ func TestEmbeddedAuthServer_ResourceCleanup(t *testing.T) {
...
+			location := resp.Header.Get("Location")
+			require.NotEmpty(t, location)
+			assert.NotContains(t, location, upstream.URL(),
+				"rejected request must NOT redirect to the upstream IDP")
+			redirectURL, err := url.Parse(location)
+			require.NoError(t, err)
+			assert.Contains(t, redirectURL.RawQuery, "invalid_scope",
+				"rejection error must be invalid_scope")
+		} else {
+			assert.Equal(t, http.StatusBadRequest, resp.StatusCode,
```
tgrunnagle: **MEDIUM — F6: Tighten the 400 branch to assert `invalid_scope`**

The 400 branch only checks the status code; it doesn't parse the body to assert `error == invalid_scope`. A future fosite upgrade that returns 400 with a different error code (`server_error`, `unauthorized_client`, etc.) would silently pass this test even while the privilege-escalation regression slipped through. The redirect branch already asserts `invalid_scope` in the URL fragment; the 400 branch should mirror that against the JSON body.

Note: the outer `defer io.Copy(io.Discard, resp.Body)` drains the body before this branch can read it. You'll need to read the body up front (and tee for the discard) or restructure the drain. Either approach is fine — the contract to enforce is "NOT-upstream redirect OR 400-with-invalid_scope-body".

cmd/thv-operator/pkg/controllerutil/authserver.go
```diff
@@ -530,6 +530,7 @@ func BuildAuthServerRunConfig(
 		AuthorizationEndpointBaseURL: authConfig.AuthorizationEndpointBaseURL,
 		AllowedAudiences:             allowedAudiences,
 		ScopesSupported:              scopesSupported,
+		BaselineClientScopes:         authConfig.BaselineClientScopes,
```
tgrunnagle, line 533: **MEDIUM — F7: Surface baseline misconfig via a status condition**

When `baselineClientScopes` references a scope not in the upstream-derived `scopesSupported`, `RunConfig.Validate()` fails inside `NewEmbeddedAuthServer` and the auth-server pod CrashLoopBackOffs. Operators have to read pod logs to discover the misconfiguration — no signal on the CR itself.

Per `.claude/rules/operator.md` "Status condition parity": configuration failures that gate startup should surface symmetrically on both `MCPExternalAuthConfig` and `VirtualMCPServer`. Two options:

1. Pre-validate the baseline against discovered upstream scopes in the reconciler and emit a `BaselineScopesInvalid` (or similar) condition via `controllerutil.MutateAndPatchStatus` before letting the workload roll. This is the operator-native UX.
2. Acknowledge the gap in the PR description and open a follow-up issue.

Low-priority since the failure mode is loud (pod won't start), but the operator-facing UX of "check pod logs" is friction worth removing.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -230,11 +230,17 @@ type EmbeddedAuthServerConfig struct {
...
-	// ScopesSupported is the list of OAuth 2.0 scopes that this authorization server supports.
-	// For an embedded auth server, this can be derived from the server's (MCP or vMCP) OIDC configuration.
+	// BaselineClientScopes is a baseline set of OAuth 2.0 scopes guaranteed to be
+	// included in every client registration. The embedded auth server unions these
+	// scopes into the registered set returned by RFC 7591 Dynamic Client
+	// Registration, so a client that narrows the `scope` field at /oauth/register
+	// can still request the baseline scopes at /oauth/authorize. All values must
+	// be present in the upstream-derived scopesSupported set; the auth server
+	// fails to start if any value is missing.
+	// +kubebuilder:validation:MaxItems=10
```
tgrunnagle, line 247: **LOW — F8: Reject empty/whitespace scope entries at admission**

The current markers admit `baselineClientScopes: [""]` or `["foo bar"]` — the runtime subset check would reject these only because they don't appear in `scopesSupported`, leading to a confusing error rather than a clean admission rejection. Three reviewers (CRD, Go correctness, test coverage) converged on the same gap.

Add `items:MinLength=1`. Optionally also `items:Pattern` for the RFC 6749 scope-token character class (`^[\x21\x23-\x5B\x5D-\x7E]+$`).

```suggestion
	// +kubebuilder:validation:MaxItems=10
	// +kubebuilder:validation:items:MinLength=1
```

Don't forget to regenerate the CRD YAMLs with `task operator-manifests` (and `task crdref-gen`) after editing.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -230,11 +230,17 @@ type EmbeddedAuthServerConfig struct {
...
+	// +listType=atomic
+	// +optional
+	BaselineClientScopes []string `json:"baselineClientScopes,omitempty"`
```
tgrunnagle, line 252: **MEDIUM — F9: User-facing docs / Helm examples for the new knob**

Only the auto-generated `docs/operator/crd-api.md` is updated. There's no example in the operator-crds Helm `values.yaml`, no manifest example under `examples/`, and no operator README mention of the canonical Claude Code DCR scope-narrowing scenario. Operators writing YAML by hand will not discover this knob.

Add at least one of:
- A commented-out example to the operator-crds chart values file.
- A sample `MCPExternalAuthConfig` manifest under `examples/` showing `baselineClientScopes: ["offline_access"]` and a one-line comment explaining when to set it.
- A short paragraph in the operator README explaining when this is needed (with the Claude Code narrow-DCR pattern as the motivating example).

pkg/authserver/server/registration/dcr.go
```diff
@@ -18,12 +18,50 @@
...
 
 	"github.com/stacklok/toolhive/pkg/oauthproto"
 )
 
+// ValidateScopeSubset checks that every scope in subset is also present in
+// superset, returning an error that names fieldName and the offending scope.
+//
+// This is the shared implementation for two defense-in-depth validation gates:
+//
+//  1. RunConfig.Validate (pkg/authserver/config.go) — the earliest gate, run
```
tgrunnagle: **LOW — F10: Doc enumerates 2 of 3 defense-in-depth gates**

The PR establishes three gates that call this helper (or its narrow sibling): `RunConfig.Validate` here, `Config.applyDefaults`'s empty-`ScopesSupported`-with-non-empty-baseline rejection (config.go:767-771), and `NewAuthorizationServerConfig` (provider.go:208). The doc lists only #1 and #3.

Either enumerate all three or drop the per-call-site list and just describe what the function does. Per `.claude/rules/go-style.md` "Keep Comments Synchronized With Code" — comments that name specific call sites rot when call sites move.

pkg/authserver/server/handlers/dcr.go
```diff
@@ -123,9 +146,10 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
 	slog.Debug("registered new DCR client", logAttrs...)
 
 	// Build response per RFC 7591 Section 3.2.1.
-	// Scope reflects the scopes actually granted to this client (from
-	// ValidateScopes above), not all server-supported scopes. This lets
-	// the client know exactly which scopes it can request.
+	// Scope reflects the scopes actually granted to this client (validated
```
tgrunnagle: **LOW — F11: Comment can be misread as implying re-validation of the unioned set**

"validated against ScopesSupported and, if configured, unioned with BaselineClientScopes" reads as if the unioned set is re-validated against `ScopesSupported`. It isn't — the union is trusted-by-construction via the startup-time subset invariant, and `ValidateScopes` runs *before* the union.

Suggested rephrase to make the ordering explicit:

```suggestion
	// Build response per RFC 7591 Section 3.2.1.
	// Scope reflects the scopes actually granted to this client: the client-supplied
	// scope set is validated against ScopesSupported by ValidateScopes, then (if
	// configured) unioned with BaselineClientScopes — which is itself guaranteed by
	// startup-time validation to be a subset of ScopesSupported. The unioned set
	// is NOT re-validated here.
```

pkg/authserver/config.go
```diff
@@ -456,6 +484,17 @@ type Config struct {
...
+	// BaselineClientScopes is a baseline set of OAuth 2.0 scopes the embedded
+	// DCR handler unions into every newly registered client's scope set. Empty
+	// means current behavior is preserved (DCR registers exactly what the client
+	// requested, or registration.DefaultScopes if the client requested none).
+	// All entries must also be present in ScopesSupported.
+	// When BaselineClientScopes is non-empty, ScopesSupported must be set
+	// explicitly: applyDefaults rejects an empty ScopesSupported in that case
+	// rather than silently substituting DefaultScopes, which would otherwise
+	// mask the operator's intent.
+	BaselineClientScopes []string
```
tgrunnagle, line 504: **LOW — F12: `Config.Validate()` at line 472 doesn't include this subset check**

A caller that bypasses YAML loading and constructs `Config` directly with mismatched `BaselineClientScopes` / `ScopesSupported` (both populated) is only caught at the inner `validateParams` boundary in provider.go. Adding one line to `Config.Validate()` would close the gap and produce a friendlier error site (clearer call stack than the inner constructor):

```go
if err := registration.ValidateScopeSubset(c.BaselineClientScopes, c.ScopesSupported, "BaselineClientScopes"); err != nil {
    return err
}
```

This also makes the "three-layer defense in depth" claim in the PR description fully accurate — currently the middle layer (`Config.Validate`) doesn't participate in the subset invariant.

pkg/authserver/server/handlers/dcr.go
jhrozek: Thanks for catching this. I looked at our log call and slog handles it — the text and JSON handlers both escape control chars in structured key/value pairs, so `"foo\nbar"` ends up as a literal `\n` rather than a real newline. So the injection isn't reachable through this call site as written.

I almost tightened anyway, but printable-ASCII (like software_id) felt too strict for client_name — it's the display name per RFC 7591 §2, so "Café App" or non-Latin names should be allowed. The middle ground of rejecting control chars only seemed like inventing a one-off convention.

Happy to revisit if a future log call drops structured logging, or if you'd rather have the stricter validation anyway.

pkg/authserver/server/handlers/dcr.go
jhrozek: Good call. Demoted to Debug in 2289650a7 — F1's startup INFO already gives the operator-visible signal that the baseline is in effect, so per-registration audit at Debug is the right tier.

pkg/authserver/runner/embeddedauthserver.go
jhrozek: Done in b494a0f58. Cloned all three slices (BaselineClientScopes, ScopesSupported, AllowedAudiences) at the runner boundary as you suggested — once at the operator-input-to-runtime transition, so downstream stages can share by reference safely.

pkg/authserver/server/provider.go
jhrozek: Added in 313a014f2 — extra case in the existing TestNewAuthorizationServerConfig_InvalidConfig table that constructs Params with a baseline scope outside ScopesSupported and asserts the `baseline_client_scopes contains "…"` error. Folded it into the existing table per testing.md's consolidation rule rather than adding a separate test function.

test/integration/authserver/authserver_integration_test.go
jhrozek: Done in 799d30951. The 400 branch now parses the JSON body and asserts `error == "invalid_scope"` — symmetric with the redirect branch's query-string check. Switched to reading the body up front (io.ReadAll → close) instead of teeing around the deferred drain.

cmd/thv-operator/pkg/controllerutil/authserver.go
jhrozek: Agreed on the gap. Going with your option 2 to keep this PR scope-disciplined — opened #5252 to cover pre-validation in the reconciler with a BaselineScopesValid (or similar) status condition across MCPExternalAuthConfig and VirtualMCPServer (per the operator parity rule), using MutateAndPatchStatus. The CrashLoop + pod-log error stays as the backstop in the meantime.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
jhrozek: Added both markers in 02acf266a: items:MinLength=1 plus items:Pattern for the RFC 6749 §3.3 scope-token charset (`^[\x21\x23-\x5B\x5D-\x7E]+$` — printable ASCII excluding space, ", and \). That moves rejection from runtime CrashLoopBackOff to apiserver admission. CRD YAMLs and crd-api.md regenerated.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
jhrozek: Good point — user-facing prose belongs in the docs-website repo rather than alongside the auto-generated crd-api.md. Opened stacklok/docs-website#866 to add a worked example of MCPExternalAuthConfig with type: embeddedAuthServer and the Claude Code scenario as the motivating case. Keeping this PR scoped to the code + auto-generated reference docs.

pkg/authserver/server/registration/dcr.go
jhrozek: Dropped the call-site list in 2049a6188 — went with your second option ("just describe what the function does") for resilience. Worth noting: the third check you mentioned in Config.applyDefaults doesn't actually call this helper (it checks a different invariant — "baseline non-empty + supported empty → reject" rather than subset membership), but the doc was misleading either way.

pkg/authserver/server/handlers/dcr.go
jhrozek: Done in f4201d379 — picked up your suggested rephrase almost verbatim, made the ordering (validate-then-union) and the "not re-validated here" intent explicit.

pkg/authserver/config.go
jhrozek: Wired up in 790f9e710. Config.Validate now calls registration.ValidateScopeSubset for BaselineClientScopes — closes the gap for direct-constructed Configs and makes the three-gate defense in depth (RunConfig.Validate → Config.Validate → validateParams) fully symmetric on the subset invariant. Added a matching case in TestConfigValidate.

rdimitrov:
I think this needs a small fix before merging if we plan to release `baselineClientScopes` as supported behavior.

The new `RunConfig.Validate()` checks `baseline_client_scopes` against the raw `ScopesSupported` slice before defaults are applied. However, the operator paths still intentionally pass nil/empty scopes to mean “use the auth server defaults” (for example from `deriveScopesSupported()` / `oidcConfig.Scopes`). That makes a natural config like:

```yaml
baselineClientScopes:
  - offline_access
```

fail startup with `baseline_client_scopes contains "offline_access" which is not in scopes_supported` when OIDC scopes are omitted, even though `offline_access` is included in `registration.DefaultScopes`.

Suggested fix: validate against the effective/defaulted scope set, or have the operator populate default `ScopesSupported` before building the RunConfig when `baselineClientScopes` is configured. Otherwise the first-use path for the new field is likely to fail unless users also explicitly configure all default scopes.

jhrozek:
Good catch, fixed in b1d7ca33e. Both validate gates now treat an empty `scopesSupported` as the default scope set for the subset check (the same set `applyDefaults` would substitute), so the natural config you described just works. The redundant guard in `applyDefaults` is gone too. Updated the doc comments and tests to match.

rdimitrov (approved):

PR #4367, opened 2026-03-25
wire spec-driven replicas and Redis password for VirtualMCPServer

yrobla:
## Summary

- Nil-passthrough for Deployment.Spec.Replicas (hands-off when spec.replicas is nil, spec-driven when set) for HPA compatibility
- Add terminationGracePeriodSeconds (30s) to vMCP pod spec
- Add buildRedisPasswordEnvVar: injects THV_SESSION_REDIS_PASSWORD from SessionStorageConfig.PasswordRef when provider is redis
- Sync Spec.Replicas on update when spec.replicas is non-nil
- deploymentNeedsUpdate detects replica drift for spec-driven scaling

Fixes #4215 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring (no behavior change)
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [ ] Unit tests (`task test`)
- [ ] E2E tests (`task test-e2e`)
- [x] Linting (`task lint-fix`)
- [ ] Manual testing (describe below)

## Changes

| File | Change |
|------|--------|
|      |        |

PR #5237, opened 2026-05-09
Add replicas and sessionStorage to MCPRemoteProxy for HA

aron-muon:
## Summary

When MCPRemoteProxy runs with multiple replicas behind a load balancer that doesn't preserve client-IP affinity (e.g. AWS ALB across multiple AZs), every non-initialize request fails with `Session not found`. The transparent proxy validates `Mcp-Session-Id` against pod-local in-memory state on every hop — `transparent_proxy.go` even calls this out:

```go
// Guard: reject non-initialize requests with unknown session IDs.
// When multiple proxyrunner replicas share a Redis session store,
// a valid session will always be found.
```

On top of that, MCPRemoteProxy had no way to *run* multiple replicas in the first place: the operator hardcoded `replicas: 1` on Deployment creation, so operators had to pin counts externally (e.g. a min=max HPA workaround that exploits the reconciler preserving the live count).

This PR completes the HA story by mirroring the two pieces MCPServer (#4368) and VirtualMCPServer (#4367) already have: a `spec.replicas` field and Redis-backed shared session storage.

**What changed:**

- **Add `MCPRemoteProxySpec.Replicas`** — same `*int32` semantics as `VirtualMCPServer.spec.replicas`: `nil` leaves `Deployment.Spec.Replicas` unset (apiserver defaults to 1; an HPA or other external controller owns scaling and is never fought by the reconciler), non-nil is operator-authoritative — set on create, drift-checked in `deploymentNeedsUpdate`, synced on update.
- **Add `MCPRemoteProxySpec.SessionStorage`** — same `SessionStorageConfig` shape used by `MCPServer` and `VirtualMCPServer`. `populateScalingConfigForRemoteProxy` writes the non-sensitive Redis parameters into `runner.ScalingConfig.SessionRedis` (before `PopulateMiddlewareConfigs`, since rate limiting reads `SessionRedis`); `buildRedisPasswordEnvVarForRemoteProxy` injects `THV_SESSION_REDIS_PASSWORD` via `SecretKeyRef` so the password never lands in the RunConfig ConfigMap.
- **Surface the shared `SessionStorageWarning` condition** when `replicas > 1` without Redis session storage — status-condition parity with the MCPServer and VirtualMCPServer validators.
- **Cross-reference `sessionAffinity` and `sessionStorage` in the field docs** — Redis-backed storage delivers HA only with `sessionAffinity: None`, and `None` without Redis is the exact failure mode this PR fixes.

The change is intentionally a near-verbatim mirror of the MCPServer / VirtualMCPServer implementations to keep review easy — it's the same pattern reviewers have already accepted twice.

## Type of change

- [x] Bug fix
- [x] New feature

## Test plan

- [x] Unit tests — full suite via `task test`. New coverage: `TestPopulateScalingConfigForRemoteProxy` (4 cases incl. password-never-serialized), `TestBuildRedisPasswordEnvVarForRemoteProxy` (4 cases), `TestDeploymentForMCPRemoteProxyReplicas` (nil/0/3), `TestMCPRemoteProxyDeploymentNeedsUpdateReplicas` (drift matrix), `TestMCPRemoteProxyEnsureDeploymentReplicaSync` (fake-cluster preserve-vs-sync), `TestValidateSessionStorageForReplicasRemoteProxy` (condition matrix)
- [x] Build verification — `task build` and `task build-operator`
- [x] CRDs/docs regenerated — `task operator-generate`, `task operator-manifests`, `task crdref-gen` (controller-gen v0.17.3)

## API Compatibility

- [x] This PR does not break the `v1beta1` API. Both added fields (`spec.replicas`, `spec.sessionStorage`) are optional; when omitted, behavior is identical to today (the only delta: a freshly created Deployment carries no explicit `replicas: 1` and relies on the apiserver default of 1).

## Changes

| File | Lines | Change |
|------|------:|--------|
| `cmd/thv-operator/api/v1beta1/mcpremoteproxy_types.go` | +39 | Add `Replicas *int32` and `SessionStorage *SessionStorageConfig` with cross-referenced docs |
| `cmd/thv-operator/api/v1beta1/zz_generated.deepcopy.go` | +10 | Generated DeepCopy for both fields (controller-gen v0.17.3) |
| `cmd/thv-operator/controllers/mcpremoteproxy_controller.go` | +53 | Replica drift check in `deploymentNeedsUpdate`; non-nil sync / nil preserve on update; `validateSessionStorageForReplicas` advisory condition |
| `cmd/thv-operator/controllers/mcpremoteproxy_deployment.go` | +36 | `Deployment.Spec.Replicas` from spec (was hardcoded 1); Redis password env var via `SecretKeyRef` |
| `cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go` | +29 | `populateScalingConfigForRemoteProxy`, called before `PopulateMiddlewareConfigs` |
| `cmd/thv-operator/controllers/mcpremoteproxy_replicas_test.go` | +255 | Replica create/drift/sync + condition tests |
| `cmd/thv-operator/controllers/mcpremoteproxy_deployment_test.go` | +66 | Redis password env var tests; nil-replicas assertion |
| `cmd/thv-operator/controllers/mcpremoteproxy_runconfig_test.go` | +86 | `TestPopulateScalingConfigForRemoteProxy` |
| `deploy/charts/operator-crds/{files,templates}/...mcpremoteproxies.yaml` | +162 ea | Generated `replicas` + `sessionStorage` schema |
| `docs/operator/crd-api.md` | +4/−1 | Generated CRD API reference |

Yes. Two new optional fields on `MCPRemoteProxy`:

- **`spec.replicas`** — desired proxy replica count. Leave unset to keep scaling external (HPA-compatible: the operator never overwrites the live count); set it to make the operator authoritative.
- **`spec.sessionStorage`** — Redis-backed shared session state across replicas, same shape as `MCPServer.spec.sessionStorage`. For HA set `replicas >= 2`, `sessionStorage.provider: redis`, and `sessionAffinity: None`; a `SessionStorageWarning` status condition flags `replicas > 1` without Redis.

## Large PR Justification

The diff is ~1,150 added lines, but the hand-written production change is only **~154 lines** (the `Replicas`/`SessionStorage` fields, the controller drift/sync + `SessionStorageWarning` logic, and the runconfig/deployment wiring). The remainder cannot be meaningfully split out:

- **Generated code (~338 lines)** — CRD schemas (`files/` + `templates/`), `zz_generated.deepcopy.go`, and `crd-api.md`/swagger, all produced by `controller-gen`/`crd-ref-docs`. These must land in the same commit as the type change: a CRD-runtime drift-detection test fails if they drift, so an intermediate split would not pass CI.
- **Tests (~655 lines)** — the unit coverage plus the cross-pod e2e (`mcpremoteproxy_scaling_test.go`) that a reviewer explicitly requested to prove cross-replica session routing. Dropping them to shrink the PR would defeat the review requirement.

Net: this is one logical change (MCPRemoteProxy HA), the production surface is small, and the bulk is generated artifacts and requested test coverage that must ship atomically with it.


cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go
```diff
@@ -173,9 +173,37 @@ func (r *MCPRemoteProxyReconciler) createRunConfigFromMCPRemoteProxy(
 		return nil, fmt.Errorf("failed to populate middleware configs: %w", err)
 	}
 
+	// Populate ScalingConfig.SessionRedis from spec.sessionStorage so the
+	// proxy runner has the address/db/keyPrefix needed to construct a
+	// shared Redis-backed session store. The Redis password is intentionally
+	// excluded here — it is injected as the THV_SESSION_REDIS_PASSWORD env
+	// var by buildRedisPasswordEnvVar in mcpremoteproxy_deployment.go.
+	populateScalingConfigForRemoteProxy(runConfig, proxy)
```
yrobla, line 177: **Ordering bug (latent):** `populateScalingConfigForRemoteProxy` must be called _before_ `PopulateMiddlewareConfigs`, not after.

`PopulateMiddlewareConfigs` → `addRateLimitMiddleware` reads `config.ScalingConfig.SessionRedis` and returns an error when it is nil and a rate-limit config is present. The equivalent call in `mcpserver_runconfig.go` explicitly documents this:

```go
// Must run before PopulateMiddlewareConfigs because rate limiting reads SessionRedis.
populateScalingConfig(runConfig, m)
if err := runner.PopulateMiddlewareConfigs(runConfig); err != nil { ... }
```

`MCPRemoteProxy` has no `rateLimiting` field today so this is harmless right now, but the moment rate limiting is added (a natural follow-on given MCPServer already has it) this ordering will cause a confusing runtime error. Please swap the two blocks:

```suggestion
	// Populate ScalingConfig.SessionRedis from spec.sessionStorage so the
	// proxy runner has the address/db/keyPrefix needed to construct a
	// shared Redis-backed session store. The Redis password is intentionally
	// excluded here — it is injected as the THV_SESSION_REDIS_PASSWORD env
	// var by buildRedisPasswordEnvVar in mcpremoteproxy_deployment.go.
	// Must run before PopulateMiddlewareConfigs because rate limiting reads SessionRedis.
	populateScalingConfigForRemoteProxy(runConfig, proxy)

	// Populate middleware configs from the configuration fields
	// This ensures that middleware_configs is properly set for serialization
	if err := runner.PopulateMiddlewareConfigs(runConfig); err != nil {
		return nil, fmt.Errorf("failed to populate middleware configs: %w", err)
	}

	return runConfig, nil
```

cmd/thv-operator/api/v1beta1/mcpremoteproxy_types.go
```diff
@@ -136,6 +136,22 @@ type MCPRemoteProxySpec struct {
...
+	// (e.g. AWS ALB across multiple AZs).
+	//
+	// The transparent proxy validates `Mcp-Session-Id` against this store on
+	// every non-initialize request (see pkg/transport/proxy/transparent/
+	// transparent_proxy.go) and rewrites client-facing session IDs to backend
+	// session IDs using session metadata. Both lookups require shared state
+	// across replicas.
+	//
+	// Mirrors MCPServer.spec.sessionStorage and VirtualMCPServer.spec.sessionStorage.
+	// +optional
```
yrobla, line 202: **Consider documenting the `sessionAffinity` + `sessionStorage` interaction.**

These two fields now coexist without any cross-reference, which creates a footgun. When `sessionStorage.provider == redis` the correct HA setup also requires `sessionAffinity: None` — otherwise the k8s Service's ClientIP stickiness makes the Redis store redundant. Conversely, `sessionAffinity: None` without Redis is the broken configuration this PR was written to fix.

At minimum, the `SessionStorage` field comment should say: _"When using the Redis provider, also set `sessionAffinity: None` so the Service routes requests round-robin and all replicas rely on the shared session store rather than pod-local state."_

A stronger option is a kubebuilder `XValidation` that requires `sessionAffinity == "None"` when `sessionStorage.provider == "redis"`, catching the misconfiguration at admission time.

yrobla (commented):
## Functional Review — Wire SessionStorage into MCPRemoteProxy

### Context

The fix addresses a real architectural gap. When `MCPRemoteProxy` runs with >1 replica behind a load balancer that lacks client-IP affinity (AWS ALB across AZs, GCP NEGs, etc.), every non-`initialize` request fails with `Session not found` because the transparent proxy validates `Mcp-Session-Id` against pod-local in-memory state. `MCPServer` and `VirtualMCPServer` already had the Redis-backed escape hatch; `MCPRemoteProxy` was the only component missing it.

### What changed and does it make sense?

The PR adds three wiring pieces:
1. **`MCPRemoteProxySpec.SessionStorage`** — same optional field shape used by MCPServer/VirtualMCPServer. Additive and backward-compatible. ✅
2. **`populateScalingConfigForRemoteProxy`** — writes non-sensitive Redis params into `runner.ScalingConfig.SessionRedis` in the RunConfig ConfigMap. ✅
3. **`buildRedisPasswordEnvVarForRemoteProxy`** — injects `THV_SESSION_REDIS_PASSWORD` as a `SecretKeyRef` pod env var, so the password never lands in the ConfigMap. ✅

The security split (connection params in ConfigMap, password as pod secret) matches the established pattern and is correct.

### Bug: `populateScalingConfigForRemoteProxy` is called **after** `PopulateMiddlewareConfigs`

In `mcpserver_runconfig.go` the analogous call order is:
```go
// Must run before PopulateMiddlewareConfigs because rate limiting reads SessionRedis.
populateScalingConfig(runConfig, m)
if err := runner.PopulateMiddlewareConfigs(runConfig); err != nil { ... }
```
`PopulateMiddlewareConfigs` → `addRateLimitMiddleware` reads `config.ScalingConfig.SessionRedis` and errors when nil and rate limiting is configured. The PR reverses this in `mcpremoteproxy_runconfig.go`. MCPRemoteProxy has no `rateLimiting` field today so this is benign now, but it diverges from the documented invariant and will break the moment rate limiting is added to MCPRemoteProxy. See inline comment for suggested fix.

### Design note: `sessionAffinity` / `sessionStorage` interaction needs documentation

`MCPRemoteProxySpec.SessionAffinity` defaults to `ClientIP`. The two fields now coexist without guidance:

| `sessionAffinity` | `sessionStorage` | Result |
|---|---|---|
| `ClientIP` (default) | nil / memory | ✅ works (pod-sticky routing) |
| `ClientIP` | redis | Redis wired but Service stickiness makes it a no-op |
| `None` | redis | ✅ intended HA configuration |
| `None` | nil / memory | ❌ `Session not found` on any cross-pod request |

The last row is the exact failure mode this PR fixes — but users can still hit it if they set `sessionAffinity: None` without configuring Redis. The `SessionStorage` field comment should mention setting `sessionAffinity: None` for the HA case, or a kubebuilder `XValidation` could enforce it at admission time (similar to how MCPServer validates that rate limiting requires Redis).

### Alternative approaches

- **Consistent hashing at the ingress** (hash `Mcp-Session-Id` to a specific pod): avoids Redis but requires custom LB config outside ToolHive's control.
- **Session migration between pods**: more complex with no clear benefit over Redis here.

Redis is the right call — same decision already made for MCPServer and VirtualMCPServer, and the operational overhead is low (one Redis shared across all three CRD types).

### Summary

| Finding | Severity | Action |
|---------|----------|--------|
| `populateScalingConfigForRemoteProxy` called after `PopulateMiddlewareConfigs` | Medium — latent bug | Reorder above `PopulateMiddlewareConfigs` |
| No guidance on `sessionAffinity` + `sessionStorage` interaction | Low — docs/UX | Add cross-reference in field comments |
| Implementation correctness, security split, API compatibility | ✅ | No action needed |

cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go
aron-muon: Fixed in 4c065e7d3acac51d88d8d8bf2641f419c53dcfce — applied the suggestion as-is; `populateScalingConfigForRemoteProxy` now runs before `PopulateMiddlewareConfigs`, mirroring the MCPServer ordering.

cmd/thv-operator/api/v1beta1/mcpremoteproxy_types.go
aron-muon: Documented in 9364db044f587c7437a810eda2be1589c0bf3800 — both field comments now cross-reference each other (regenerated into the CRD schema and crd-api.md). Went with docs-only rather than the CEL rule: ClientIP + redis is still a valid combination (the shared store preserves sessions when a pod dies and affinity re-pins the client), so rejecting it at admission felt too strict.

cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go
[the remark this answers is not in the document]
aron-muon: Fixed in 4c065e7d3acac51d88d8d8bf2641f419c53dcfce.

aron-muon:
Pushed d6264234fe: added `spec.replicas` alongside `spec.sessionStorage`, mirroring the `VirtualMCPServer` semantics — `nil` leaves scaling to an HPA/external controller (the reconciler never overwrites the live count), non-nil is operator-authoritative with drift reconciliation, and the shared `SessionStorageWarning` condition flags `replicas > 1` without Redis. Previously the operator hardcoded `replicas: 1`, so multi-replica deployments required pinning the count with an external min=max HPA workaround. Also merged latest main. PR title/description updated to reflect the fuller HA scope.

jerm-dro (commented):
Thanks for your patience on this one, and for keeping it such a faithful mirror of the MCPServer/VirtualMCPServer pattern — the wiring, the password-via-`SecretKeyRef` split, and the drift/sync semantics all line up, and the unit coverage for those is solid.

The one thing I'd like before approving is an end-to-end test proving the cross-pod behavior, matching what the two sibling types already ship:

- `test/e2e/thv-operator/virtualmcp/mcpserver_scaling_test.go` — *"MCPServer Cross-Replica Session Routing with Redis"*: deploys Redis, creates the resource with `replicas=2` + `sessionAffinity: None`, initializes a session on pod A, then drives requests with the **same session ID against pod B** and asserts they resolve via shared Redis state (plus `SessionStorageWarning=False`).
- `test/e2e/thv-operator/virtualmcp/virtualmcp_redis_session_test.go` — same cross-pod shape, plus pod-restart session survival.

The current unit tests prove the config is *populated* correctly, but not that a session created on one replica actually works on another. That end-to-end path is the most important thing to lock down here, because MCPRemoteProxy's transparent proxy does something the unit tests don't exercise: it validates `Mcp-Session-Id` against the store **and rewrites client-facing session IDs to backend session IDs** on every hop. That rewrite working across pods via shared Redis is exactly the failure mode this PR fixes, so it's the behavior most worth a real e2e.

An `mcpremoteproxy` test mirroring `mcpserver_scaling_test.go` would do it. The one difference vs. the siblings: MCPRemoteProxy targets a remote MCP server by URL rather than a deployed container — but you could likely just point it at another MCP server already deployed in the kind cluster (reachable by its in-cluster Service URL), so the fixture stays close to the existing pattern. I'll leave the exact approach to you.

aron-muon:
@jerm-dro thanks — added the cross-pod e2e in 98017e40a0def9826d370b623b5e9702a8e3b7b5: `test/e2e/thv-operator/virtualmcp/mcpremoteproxy_scaling_test.go`, mirroring `mcpserver_scaling_test.go`.

It deploys Redis + a backend MCPServer, creates an `MCPRemoteProxy` with `replicas=2`, `sessionAffinity: None`, and Redis `sessionStorage`, then:
- asserts `SessionStorageWarning=False` once Redis is configured;
- initializes a session on proxy pod **A** and drives 5 requests with the same `Mcp-Session-Id` against proxy pod **B**, asserting each resolves via shared Redis and returns the same tools as A — i.e. the transparent proxy's validate-against-store + client→backend session-ID rewrite working across pods, which is the exact failure mode this PR fixes.

Per your note about MCPRemoteProxy targeting a remote URL rather than a deployed container: the one fixture difference is that the proxy points at the backend MCPServer's in-cluster Service URL (`http://mcp-<backend>-proxy.<ns>.svc.cluster.local:8080/mcp`), so everything else stays close to the sibling pattern and reuses the shared `deployRedis`/`portForwardToPod`/`CreateInitializedMCPClient` helpers.

One caveat for transparency: I could not run the operator e2e suite locally (no kind cluster on this box), so I validated with `go vet` (compiles the suite), `gofmt`, and `golangci-lint` (0 issues) and a careful trace against the controller — service name/port/path, proxy pod labels, the `Ready` phase gate, and the `SessionStorageWarning` wiring. It runs in the same Ginkgo suite as the siblings, so the operator-e2e CI job will exercise it here.

jerm-dro (approved):

jerm-dro:
The flaky test here doesn't appear to be caused by you. I'll fix this in a follow-up PR. Thanks @aron-muon for the contribution 🚀

PR #5243, opened 2026-05-11
Fix operator RBAC for event recording

pl4nty:
## Summary

Fix operator RBAC for event recording, after controller-runtime breaking change in #4183. Noticed in errors like

```
E0511 02:03:05.615321 1 event_broadcaster.go:270] "Server rejected event (will not retry!)"
err="events.events.k8s.io is forbidden:
User "system:serviceaccount:toolhive-system:toolhive-operator" cannot create resource "events" in API group "events.k8s.io" in the namespace "toolhive-system""
event="&Event{
  ObjectMeta:{toolhive.18ae607b50e05fea toolhive-system 0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] [] []
},
EventTime:2026-05-11 02:03:05.612044577 +0000 UTC m=+798.614640250,Series:nil,
ReportingController:virtualmcpserver-controller,ReportingInstance:virtualmcpserver-controller-toolhive-operator-855bf4f6f8-hcgn7,
Action:UpdateDeployment,Reason:DeploymentUpdated,Regarding:{
  VirtualMCPServer toolhive-system toolhive 002de0e1-8d34-46a7-a339-89ce77f15140 toolhive.stacklok.dev/v1beta1 661778437
},
Related:nil,Note:Deployment updated, rolling out new configuration,Type:Normal,DeprecatedSource:{ },
DeprecatedFirstTimestamp:0001-01-01 00:00:00 +0000 UTC,DeprecatedLastTimestamp:0001-01-01 00:00:00 +0000 UTC,DeprecatedCount:0,}"
```

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## API Compatibility

No

## Implementation plan

Agent-Logs-Url: https://github.com/pl4nty/toolhive/sessions/91c560a3-ef9f-42db-82be-d6fceeadeff7


ChrisJBurns (dismissed):

ChrisJBurns:
@pl4nty Thanks for the PR, you'll need to run the [helm-docs / pre-commit](https://github.com/stacklok/toolhive/blob/main/deploy/charts/operator/CONTRIBUTING.md) commands locally to make sure that the docs are all updated with the updated vaues.yaml file

pl4nty:
Thanks, fixed! Also noticed I didn't run all the e2e tests, so I've fixed those up too

jhrozek:
I approved the CI workflows so that the PR runs through tests but you'll want to rebase @pl4nty

amirejaz (commented):
Thanks for this fix, @pl4nty! The RBAC correction is exactly right — `controller-runtime` v0.19+ switched its event recorder to use the `events.k8s.io` API group, so the operator service account needs the corresponding permission.

The change is thorough: all five `//+kubebuilder:rbac` markers are updated (so `make generate`/`controller-gen` will emit the right RBAC going forward), the Helm chart ClusterRole template and values are updated, and the Chainsaw E2E assertion files are in sync. The README docs accurately reflect the new defaults.

One minor note on the README files inline below — otherwise LGTM.

deploy/charts/operator/README.md
```diff
@@ -43,10 +43,10 @@ The command removes all the Kubernetes components associated with the chart and
 ## Values
 
 | Key | Type | Default | Description |
-|-----|------|---------|-------------|
+|-----|-------------|------|---------|
```
amirejaz: The separator row here appears to be a side-effect of whatever doc-generation tool was used to regenerate the README. The header row still reads `| Key | Type | Default | Description |` (four columns), but the separator widths now suggest a different column ordering. Markdown renderers only use separator rows for alignment hints, not for re-ordering, so the table renders correctly — but it is a minor visual inconsistency.

This README is likely auto-generated (e.g. via `helm-docs`), so it may be simplest to re-run the doc generator rather than editing the separator manually. If there is a `task docs` or equivalent target, running it should restore the canonical output.

deploy/charts/operator-crds/README.md
```diff
@@ -80,7 +80,7 @@ This is a one-time operation. After patching, future upgrades will work as long
 ## Values
 
 | Key | Type | Default | Description |
-|-----|------|---------|-------------|
+|-----|-------------|------|---------|
```
amirejaz: Same separator-row inconsistency as `deploy/charts/operator/README.md` — same suggestion applies: re-running the doc generator should normalize this.

deploy/charts/operator/README.md
pl4nty: Yeah, the separator change was generated with helm-docs

amirejaz (dismissed):

amirejaz:
The Helm charts CI is failing because CI's `helm-docs` generates a slightly different separator row than what was committed. Just run:
```
task helm-docs
```
and commit the updated READMEs — that's the same command CI uses and will produce the exact output it expects.

amirejaz (approved):

ChrisJBurns (approved):

PR #5245, opened 2026-05-11
Namespace operator.* Helm helpers to prevent umbrella chart collisions

### LINKED ISSUE #5097: helm: namespace operator.* helpers in deploy/charts/operator to prevent umbrella collisions
## Summary

`deploy/charts/operator/templates/_helpers.tpl` defines its helper family under the generic `operator.*` prefix:

- `operator.name`
- `operator.fullname`
- `operator.chart`
- `operator.labels`
- `operator.selectorLabels`
- `operator.serviceAccountName`

In standalone installs this works. In umbrella installs where this chart is pulled alongside any other chart that also defines `operator.*` helpers, Helm's **global** template-helper namespace causes one chart's definitions to silently shadow the other's. Whichever `_helpers.tpl` Helm parses last wins for every callsite in the dependency tree, regardless of which chart issued the `include`.

The shadowing is independent of `condition:` gating: `_helpers.tpl` files are loaded at chart-load time, not render time, so even a disabled subchart's helpers leak into the global namespace and can hijack callsites in other (enabled) subcharts.

The standard Helm convention is to prefix helpers with the chart name. The same `_helpers.tpl` already isolates its sibling family correctly (`toolhive.labels`), so the operator family is internally inconsistent.

## Failure mode

When two charts in the same dependency tree both define `operator.serviceAccountName` (each reading the value of `.Values.<their-own-path>.name`), and one chart's deployment template invokes `{{ include "operator.serviceAccountName" . }}`, Helm dispatches to whichever chart's definition was parsed last. If that definition reads a value path that exists in *its* value tree but not in the calling chart's, render fails with:

```
nil pointer evaluating interface {}.name
```

The error trace is misleading because the file-and-line reported is the *helper's* source, not the chart that called it — debugging this is non-trivial for anyone who hasn't seen it before.

## Proposed change

Rename the helper family in `deploy/charts/operator/templates/_helpers.tpl`:

| Before | After |
|---|---|
| `operator.name` | `toolhive-operator.name` |
| `operator.fullname` | `toolhive-operator.fullname` |
| `operator.chart` | `toolhive-operator.chart` |
| `operator.labels` | `toolhive-operator.labels` |
| `operator.selectorLabels` | `toolhive-operator.selectorLabels` |
| `operator.serviceAccountName` | `toolhive-operator.serviceAccountName` |

Update the call-site files (`deployment.yaml`, `serviceaccount.yaml`, `hpa.yaml`, `clusterrole/role.yaml`, `clusterrole/rolebinding.yaml`, `leader-election-role.yaml`) accordingly. `toolhive.labels` stays unchanged — already correctly namespaced.

This is a pure rename — output is byte-identical under standalone install. No `values.yaml` change required. The only externally visible effect is that anything outside this chart that calls `{{ include "operator.fullname" . }}` would need to update — which would only happen in an umbrella, which is exactly the scenario this fixes.

`deploy/charts/operator-crds` is also worth checking, though I haven't.

## Severity

Low for standalone consumers (no current failure). A footgun for any umbrella that pulls this chart alongside another with `operator.*` helpers. Filing as housekeeping rather than urgent breakage.

wucm667:
## Summary

The operator chart's `_helpers.tpl` defines helpers under the generic `operator.*` prefix. In umbrella chart installs where multiple charts are used together, this causes template namespace collisions — whichever chart's `_helpers.tpl` Helm parses last wins, silently shadowing other charts' definitions.

This renames the helper family from `operator.*` to `toolhive-operator.*`, following the standard Helm convention of prefixing helpers with the chart name. The `toolhive.labels` helper was already correctly namespaced and is unchanged.

This is a pure rename — output for standalone installs is byte-identical. No `values.yaml` changes required.

Fixes #5097

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Manual testing

Verified:
- All 6 helper definitions in `_helpers.tpl` renamed with correct internal references updated
- All call sites in `deployment.yaml`, `hpa.yaml`, `serviceaccount.yaml`, `leader-election-role.yaml` updated
- No remaining `include "operator.` references in any template files
- 5 files changed, 22 insertions, 22 deletions — pure text rename, zero logic changes

## API Compatibility

No API surface change. This only affects internal Helm template helper names within the operator chart.


wucm667:
The "Test Go Code" failure is in `TestSessionManager_CreateSession/returns_error_when_session_is_terminated_during_backend_initialization` (`pkg/vmcp/server/sessionmanager/session_manager_test.go:481`) — a race condition in the test timing that predates this PR. This PR only renames Helm template helpers in `deploy/charts/operator/templates/` and doesn't touch any Go code. The failure is unrelated to these changes.

ChrisJBurns (approved):

PR #5246, opened 2026-05-11
Fall back across Docker sockets on connect failure

### LINKED ISSUE #5120: Linux: `thv` fails with "no container runtime available" when only Docker Desktop is installed
## Bug description

On Linux, when Docker Desktop is the only Docker installation (i.e. no system Docker Engine running on `/var/run/docker.sock`), `thv` cannot find a container runtime and fails with:

```
no container runtime available
```

Docker Desktop on Linux runs the Docker Engine inside a VM and exposes its socket at `~/.docker/desktop/docker.sock`. It also registers a `desktop-linux` Docker CLI context pointing at that socket, which is why `docker` / `docker ps` work fine from the same shell — the Docker CLI honors contexts.

`thv`'s runtime auto-detection only probes `/var/run/docker.sock` and a handful of macOS-specific paths (Docker Desktop on macOS, Rancher Desktop, OrbStack), so the Docker Desktop on Linux socket is never discovered. As a result `IsAvailable()` returns false and `CheckRuntimeAvailable` errors out, even though Docker is actually running and reachable.

This affects any Linux user who installs Docker Desktop only, with no separate system Docker Engine.

## Steps to reproduce

On a Linux host with **only** Docker Desktop installed (no system Docker Engine, no Podman, no Colima):

1. Confirm Docker Desktop is running and the CLI works:
   ```
   docker context ls       # should show desktop-linux as current
   docker ps               # succeeds
   ls -l ~/.docker/desktop/docker.sock
   ```
2. Run any `thv` command that requires a container runtime, e.g.:
   ```
   thv version
   thv list
   ```

## Expected behavior

`thv` discovers the Docker Desktop on Linux socket at `~/.docker/desktop/docker.sock` (matching what the active `desktop-linux` Docker context points to) and proceeds normally, the same way `docker` does.

## Actual behavior

`thv` fails with:

```
no container runtime available
```

even though `docker ps` in the same shell works against Docker Desktop.

## Environment (if relevant)

- OS/version: Linux (any distro running Docker Desktop for Linux)
- Docker: Docker Desktop for Linux only — no `/var/run/docker.sock` from a system Docker Engine
- ToolHive version: current `main` (`pkg/container/docker/sdk/client_unix.go` runtime auto-detection)

## Additional context

- The Docker CLI works because it honors `~/.docker/config.json` / `~/.docker/contexts/meta` and resolves the `desktop-linux` context. ToolHive's Go runtime detection does not consult Docker contexts; it just stats a fixed list of socket paths.
- Workaround until a fix lands — point ToolHive at the socket explicitly via the existing env override:
  ```
  export TOOLHIVE_DOCKER_SOCKET="$HOME/.docker/desktop/docker.sock"
  ```
  This is honored by `findPlatformContainerSocket` and lets `thv` connect immediately.

samuv:
## Summary

- On a Linux box where `/var/run/docker.sock` exists but is not connectable (typically because the user is not in the `docker` group, returning `EACCES`), socket discovery short-circuited on the first stat-OK path and never tried alternative Docker sockets such as Docker Desktop's `~/.docker/desktop/docker.sock`. The outer loop in `NewDockerClient` only iterated runtime *types* (Podman → Docker → Colima), not multiple sockets within a runtime, so users with a Docker-Desktop-on-Linux setup that also exposes a host-side wrapper at `/var/run/docker.sock` were told "no container runtime available" even though a usable Docker daemon was reachable via a different socket.
- `findDockerSocket` / `findPodmanSocket` / `findColimaSocket` now return every candidate path that exists on disk, in priority order. `NewDockerClient` tries each one and only moves on to the next runtime after every candidate for the current runtime has failed.
- Added `dockerPermissionHint`: when the final error is a `permission denied` on a socket connect, the user now gets an actionable hint (`sudo usermod -aG docker $USER`, log out / back in, link to the Docker post-install docs) appended to the existing error instead of being left to decode `permission denied` themselves.
- Behavior of the `TOOLHIVE_*_SOCKET` env-var overrides is unchanged — the override is explicit so we still short-circuit candidate collection to a single path.

Fixes #5120

## Type of change

- [x] Bug fix

## Test plan

- [x] Unit tests (`task test`) — focused package: `go test -count=1 -race ./pkg/container/docker/sdk/...` passes; full `task test` fails on an unrelated pre-existing test in `pkg/skills/client` that picks up a stale server-discovery file on hosts where another `thv` is running. The diff is confined to `pkg/container/docker/sdk/`.
- [x] Manual testing — verified the new `TestFindDockerSocket_ReturnsAllCandidates` (two stat-OK Docker sockets, both returned in priority order) and `TestDockerPermissionHint` (covers nil, non-permission, `fs.ErrPermission`-wrapped, and the literal `"permission denied"` substring shape that the Docker SDK renders).

## API Compatibility

- [x] This PR does not break the `v1beta1` API. No operator surface touched; all changes are in `pkg/container/docker/sdk/`.

## Changes

| File | Change |
|------|--------|
| `pkg/container/docker/sdk/client_unix.go` | `findDockerSocket` / `findPodmanSocket` / `findColimaSocket` now return `[]string` of every stat-OK candidate; `findPlatformContainerSocket` returns `([]string, runtime.Type, error)`. |
| `pkg/container/docker/sdk/client_windows.go` | `findPlatformContainerSocket` mirrored to return `[]string` — single-element on success (Windows has one named pipe per runtime). |
| `pkg/container/docker/sdk/factory.go` | `NewDockerClient` iterates over every socket candidate per runtime before moving on; adds `dockerPermissionHint` for actionable EACCES errors. |
| `pkg/container/docker/sdk/client_unix_test.go` | Updates existing tests for the new return types; adds `TestFindDockerSocket_ReturnsAllCandidates` and `TestDockerPermissionHint`. |

Yes. Users with a Docker-Desktop-on-Linux setup, where `/var/run/docker.sock` is present but not connectable by the calling user, will now have `thv` automatically fall back to `~/.docker/desktop/docker.sock` and start cleanly. When all attempts fail with permission errors, the error message now includes a docker-group hint and a link to the Docker post-install docs.

- The env-var override path (`TOOLHIVE_DOCKER_SOCKET` etc.) intentionally still returns a single-element slice: when the user explicitly points us at a socket, we should not silently fall through to a different one.
- `dockerPermissionHint` uses both `errors.Is(err, fs.ErrPermission)` and a substring match on `"permission denied"`. The Docker client SDK wraps the underlying `*os.SyscallError` in its own formatted message, so the `errors.Is` chain doesn't always reach `fs.ErrPermission`; the substring fallback keeps the hint reliable across that wrapping.

### DIFF
--- a/pkg/container/docker/sdk/client_unix.go
+++ b/pkg/container/docker/sdk/client_unix.go
@@ -43,127 +43,122 @@ func newPlatformClient(socketPath string) (*http.Client, []client.Opt) {
 	return httpClient, opts
 }
 
-// findPlatformContainerSocket finds a container socket path on Unix systems
-func findPlatformContainerSocket(rt runtime.Type) (string, runtime.Type, error) {
-	// First check for custom socket paths via environment variables
+// findPlatformContainerSocket finds container socket paths on Unix systems.
+// Returns every candidate path that exists on disk, in priority order, so the
+// caller can fall back to the next candidate when an earlier one is found but
+// cannot be connected to (e.g. /var/run/docker.sock present but EACCES, while
+// Docker Desktop's per-user socket would have worked).
+func findPlatformContainerSocket(rt runtime.Type) ([]string, runtime.Type, error) {
+	// First check for custom socket paths via environment variables. The
+	// override is explicit, so we don't add other fallbacks alongside it.
 	if customSocketPath := os.Getenv(PodmanSocketEnv); customSocketPath != "" {
 		//nolint:gosec // G706: socket path from trusted environment variable
 		slog.Debug("using Podman socket from env", "path", customSocketPath)
-		// validate the socket path
 		if _, err := os.Stat(customSocketPath); err != nil { //nolint:gosec // G703: socket path from trusted environment variable
-			return "", runtime.TypePodman, fmt.Errorf("invalid Podman socket path: %w", err)
+			return nil, runtime.TypePodman, fmt.Errorf("invalid Podman socket path: %w", err)
 		}
-		return customSocketPath, runtime.TypePodman, nil
+		return []string{customSocketPath}, runtime.TypePodman, nil
 	}
 
 	if customSocketPath := os.Getenv(DockerSocketEnv); customSocketPath != "" {
 		//nolint:gosec // G706: socket path from trusted environment variable
 		slog.Debug("using Docker socket from env", "path", customSocketPath)
-		// validate the socket path
 		if _, err := os.Stat(customSocketPath); err != nil { //nolint:gosec // G703: socket path from trusted environment variable
-			return "", runtime.TypeDocker, fmt.Errorf("invalid Docker socket path: %w", err)
+			return nil, runtime.TypeDocker, fmt.Errorf("invalid Docker socket path: %w", err)
 		}
-		return customSocketPath, runtime.TypeDocker, nil
+		return []string{customSocketPath}, runtime.TypeDocker, nil
 	}
 
 	if customSocketPath := os.Getenv(ColimaSocketEnv); customSocketPath != "" {
 		//nolint:gosec // G706: socket path from trusted environment variable
 		slog.Debug("using Colima socket from env", "path", customSocketPath)
-		// validate the socket path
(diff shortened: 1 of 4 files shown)


samuv (commented):
TEST **Redacted**

rdimitrov (approved):

PR #5042, opened 2026-04-23
Add authserver DCR credential store and resolver

tgrunnagle:
## Summary

Upstream Identity Providers that support RFC 7591 Dynamic Client Registration can now be configured without a pre-provisioned `client_id`. At startup the authserver discovers the registration endpoint, registers itself once per `(issuer, redirect_uri, scopes)` tuple, and caches the resulting credentials for the process lifetime. This unblocks the remaining wiring + observability work in sub-issue C and removes the last manual provisioning step for DCR-capable upstreams (e.g., nanobot, Hydra).

- WHY: Operators running the embedded authserver against a DCR-capable upstream were forced to pre-register a client manually and hand-roll the `client_id`/`client_secret` into the run config. Phase 2 of the DCR story (#4978) lets the authserver do this at startup, keyed canonically so a future Redis backend (Phase 3) can share the cache across replicas without redefining the key shape.
- WHAT:
  - `DCRCredentialStore` interface + in-memory implementation keyed by `(Issuer, RedirectURI, ScopesHash)`. `ScopesHash` is SHA-256 of the deduplicated + sorted scope list, so permuted or duplicated scope inputs share a digest. No TTL, no eviction — RFC 7591 registrations are long-lived and are only revoked by explicit RFC 7592 deregistration.
  - `resolveDCRCredentials` performs (1) cache lookup (short-circuits before any network I/O on hit), (2) authorization-server metadata discovery via `FetchAuthorizationServerMetadata` when `DiscoveryURL` is set, or the operator-supplied `RegistrationEndpoint` directly, (3) explicit `rc.AuthorizationEndpoint` / `rc.TokenEndpoint` override of discovered values, (4) redirect-URI derivation from the issuer origin + `/oauth/callback` with HTTPS enforcement outside loopback, (5) auth-method intersection against the preference order `private_key_jwt > client_secret_basic > client_secret_post > none`, (6) synthesis of `{origin}/register` when metadata omits `registration_endpoint` (nanobot/Hydra convention), (7) a single call to `oauthproto.RegisterClientDynamically` with the initial access token injected as `Authorization: Bearer` via a wrapping `RoundTripper`, and (8) `DCRCredentialStore.Put` before returning — per the write-durable-first rule in `.claude/rules/go-style.md`.
[...]

PR #5250, opened 2026-05-11
Migrate CLI OAuth flow to pkg/auth/dcr resolver

### LINKED ISSUE #5219: Migrate CLI OAuth flow to pkg/auth/dcr resolver
## Description

Sub-issue 4b of #5145. With sub-issue 4a (PR #5198) the new `pkg/auth/dcr` package is in place and `EmbeddedAuthServer` consumes it. The CLI OAuth flow at `pkg/auth/discovery/discovery.go::PerformOAuthFlow` still calls `oauthproto.RegisterClientDynamically` directly, so it does not yet inherit the review-property behaviours added during #5042 (S256 PKCE gating, RFC 7591 §3.2.1 expiry-driven refetch, bearer-token transport with redirect refusal, panic recovery, singleflight deduplication). This sub-issue migrates that call site to the shared resolver and locks in the invariant with a CI grep guard.

After 4b lands, all four acceptance criteria of the parent issue are met and #5145 closes.

## Context

`pkg/auth/dcr.ResolveCredentials` today takes embedded-authserver-shaped types — `*authserver.OAuth2UpstreamRunConfig` and `*upstream.OAuth2Config` — because in 4a we deliberately kept the move-only scope rather than designing a profile-neutral input with only one consumer in hand. 4b is when the second consumer arrives and the right shape becomes obvious.

The CLI side speaks `OAuthFlowConfig` (`pkg/auth/discovery/discovery.go:501`) and persists DCR results into `OAuthFlowResult.ClientID` / `ClientSecret` (`pkg/auth/discovery/discovery.go:517`). The downstream remote handler reads those back via `pkg/auth/remote/handler.go`'s `CachedClientID` / `CachedClientSecretRef` / `CachedRegTokenRef` fields. That is the CLI's existing persistence model and the migration must preserve it.

**Dependencies**: sub-issue 4a (PR #5198) — provides the `pkg/auth/dcr` package and the `EmbeddedAuthServer` consumer.
**Blocks**: closure of #5145.

### Profile-neutral resolver input — `pkg/auth/dcr`

- [ ] `pkg/auth/dcr` exposes a profile-neutral input type (or a small adapter pair) so both `EmbeddedAuthServer` and the CLI flow can call `ResolveCredentials` without coupling the package to either consumer's domain types. The current `*authserver.OAuth2UpstreamRunConfig` / `*upstream.OAuth2Config` parameters are either replaced or wrapped.
- [ ] The package doc comment in `pkg/auth/dcr/resolver.go` is updated to drop the "target state" caveat added in 4a — the package is now genuinely profile-agnostic.
- [ ] Existing `EmbeddedAuthServer` call sites continue to work (either unchanged via an adapter, or migrated to the new shape in the same PR).

### CLI flow migration — `pkg/auth/discovery`

- [ ] `pkg/auth/discovery/discovery.go::PerformOAuthFlow` no longer calls `oauthproto.RegisterClientDynamically` directly. The DCR step routes through `dcr.ResolveCredentials`.
- [ ] `handleDynamicRegistration` (or its replacement) populates `OAuthFlowResult.ClientID` / `ClientSecret` from the `dcr.Resolution` returned by the resolver.
- [ ] CLI-specific concerns stay at the call site: loopback redirect URI per RFC 8252 §7.3 (the existing `NewDynamicClientRegistrationRequest(config.Scopes, config.CallbackPort)` shape), and the existing CLI-flag fallback error message ("configure with `--remote-auth-client-id` / `--remote-auth-client-secret`") when the provider does not advertise a registration endpoint.

### Credential persistence — `pkg/auth/remote/handler.go`

- [ ] `CachedClientID` / `CachedClientSecretRef` / `CachedRegTokenRef` continue to work as the CLI's persistent credential store across runs. Either: (a) wrap them as a `dcr.CredentialStore` adapter so the resolver short-circuits on cache hit, or (b) treat them as a read-only fallback after `dcr.ResolveCredentials` returns, with the resolver running against `dcr.NewInMemoryStore()` for the duration of a single CLI invocation.
- [ ] The chosen approach is documented in the commit message and in a comment at the wiring site.

### Inherited behaviours — verified by tests

- [ ] CLI flow inherits **S256 PKCE gating**: when the upstream advertises only `plain` PKCE, the CLI surfaces a clear error from the resolver rather than silently downgrading.
- [ ] CLI flow inherits **RFC 7591 §3.2.1 expiry-driven refetch**: when a cached registration is expired, the next `PerformOAuthFlow` call re-registers.
- [ ] CLI flow inherits **bearer-token transport redirect refusal**: registration POSTs do not follow redirects.
- [ ] CLI flow inherits **panic recovery** around the registration body.
- [ ] CLI flow inherits **singleflight deduplication**: two concurrent `PerformOAuthFlow` invocations with the same `(issuer, scopes, redirectURI)` tuple result in exactly one upstream `/register` call.

### Invariant guard

- [ ] CI fails if `oauthproto.RegisterClientDynamically` is referenced outside `pkg/auth/dcr/` or `pkg/oauthproto/` itself. Implementation: a grep step in the existing lint/check pipeline (or a Taskfile target invoked from CI), not a new linter. The parent issue calls this out: *"No direct calls to `oauthproto.RegisterClientDynamically` outside `pkg/auth/dcr` (search-grep enforces this)."*

### Cross-cutting

- [ ] The PR closes #5145.

### Recommended implementation

1. **Design the neutral input.** Walk both call sites (embedded authserver and CLI) and identify the minimal field set the resolver actually reads: registration endpoint (or discovery URL), requested scopes, redirect URI, optional initial access token, optional explicit endpoint overrides, and an HTTP client. Promote that into a `dcr.Request` (or similar) struct. Convert the existing `OAuth2UpstreamRunConfig` / `upstream.OAuth2Config` consumers to construct a `dcr.Request` at the call site; the resolver no longer imports either domain type.

2. **Write the CLI adapter.** `pkg/auth/discovery/discovery.go` constructs a `dcr.Request` from `OAuthFlowConfig` (issuer, scopes, callback port → loopback redirect URI per RFC 8252 §7.3) and either:
   - **(a)** passes a `dcr.CredentialStore` backed by `OAuthFlowResult` / `pkg/auth/remote/handler.go`'s `Cached*` fields. This is the cleaner long-term shape because cache hits stay inside the resolver. It does require designing a small `Get` / `Put` adapter over the `secretProvider` interface used by the remote handler.
   - **(b)** passes `dcr.NewInMemoryStore()` and continues to populate `OAuthFlowResult` from the returned `dcr.Resolution`. Simpler, ships faster, but forfeits cross-invocation cache hits at the resolver layer (the remote handler still reads `CachedClientID` directly, so cross-invocation reuse still works in practice — just outside the resolver).

   Recommend **(a)** if the secret-provider plumbing is straightforward; otherwise **(b)** is acceptable provided the commit message documents the trade-off.

3. **Replace the discovery.go DCR call.** Delete the direct `oauthproto.RegisterClientDynamically` call at `pkg/auth/discovery/discovery.go:747`. Translate the existing CLI-flag fallback error message (`"this provider does not support Dynamic Client Registration (DCR). Please configure ..."`) into a check on the resolver's returned error, since the resolver also synthesises a registration endpoint per RFC 8414 in some cases.

4. **Add the CI grep guard.** Either a new `task check-dcr-isolation` Taskfile target invoked from the existing CI lint/check job, or a one-line grep in the existing job. The check fails if `RegisterClientDynamically` matches anywhere outside `pkg/auth/dcr/` and `pkg/oauthproto/`.

5. **Drop the 4a doc caveat.** With a real second consumer in hand, edit `pkg/auth/dcr/resolver.go`'s package doc comment to remove the "target state" framing introduced in 4a's review-feedback commit.

### Code pointers

- `pkg/auth/discovery/discovery.go:500-535` — `OAuthFlowConfig`, `OAuthFlowResult`, `shouldDynamicallyRegisterClient`. The migration site is centred here.
- `pkg/auth/discovery/discovery.go:589-613` — `handleDynamicRegistration`. This is the function that currently calls `oauthproto.RegisterClientDynamically` (line 747 is inside its helper).
- `pkg/auth/discovery/discovery.go:744-754` — the direct `oauthproto.RegisterClientDynamically` call to remove.
- `pkg/auth/dcr/resolver.go:386` — `ResolveCredentials` signature. Refactor target.
- `pkg/auth/dcr/resolver.go:140` — `Resolution` struct. CLI consumer reads `ClientID` / `ClientSecret` off this.
- `pkg/auth/dcr/resolver.go:208,253,287` — `NeedsDCR` / `ConsumeResolution` / `ApplyResolutionToOAuth2Config`. Profile-coupled API to either neutralise or split.
- `pkg/auth/remote/handler.go:252-258` — where `CachedClientID` / `CachedClientSecretRef` are read back. The persistence contract this PR must preserve.

### Patterns & frameworks

- Plain `testing` + `github.com/stretchr/testify` per `.claude/rules/testing.md` — NOT Ginkgo.
- Always drive builds/tests/lint through `task` commands.
- Commit-message style: imperative mood, capitalized subject, no trailing period, 50-char limit, no Conventional Commits prefixes.
- PR-size cap: 400 LOC / 10 files (excluding tests/docs/generated). If 4b exceeds the cap after the input refactor, split into a (4b-i: neutralise resolver inputs) + (4b-ii: migrate CLI + add guard) pair.

### Unit tests

- [ ] `pkg/auth/dcr` — resolver behaviour tests under the new neutral input type. Existing test cases carry over with minimal surface change.
- [ ] `pkg/auth/discovery` — `PerformOAuthFlow` against a mock authorization server that advertises `/register`. Assert that one `OAuthFlowResult.ClientID` is populated and that the resolver was the code path that produced it (e.g. via a fake `dcr.CredentialStore` whose `Put` call is observed).

### Integration tests

- [ ] **S256 gating**: mock AS advertises only `plain` PKCE; `PerformOAuthFlow` returns a resolver-shaped error.
- [ ] **Expiry refetch**: cache an expired registration; next `PerformOAuthFlow` invocation re-registers.
- [ ] **Redirect refusal**: mock AS responds to `/register` with 302 to an evil host; resolver refuses and the CLI surfaces the error.
- [ ] **Singleflight**: two goroutines call `PerformOAuthFlow` concurrently with the same (issuer, scopes, redirectURI); the mock AS records exactly one `/register` request.
- [ ] **Persistence preserved**: register once, then construct a fresh `PerformOAuthFlow` against the same `Cached*` fields; assert the second call short-circuits (zero `/register` requests).

### Edge cases

- [ ] Provider does not advertise a `registration_endpoint` and synthesis is unavailable: CLI surfaces the existing flag-fallback message verbatim.
- [ ] Provider advertises a registration endpoint but returns 404/405/501: CLI surfaces a clear error (the protocol-neutral message lives in `pkg/oauthproto.handleHTTPResponse`; the resolver surfaces it through).

## Out of Scope

- RFC 7592 client management / rotation (#3631).
- SSRF hardening on operator-supplied DCR URLs (#5135).
- The `Notify When DCR Is Unsupported for Remote MCP Server` UX work (#2680).
- Any changes to `pkg/oauthproto` wire-shape types.
- Refactoring `pkg/auth/remote/handler.go`'s caching beyond what is needed to satisfy the persistence-preservation acceptance criterion.

## References

- Parent issue: #5145.
- Sub-issue 4a: PR #5198 — extracted `pkg/auth/dcr` and migrated `EmbeddedAuthServer`.
- Originating PR: #5042 — defined the review-property behaviours that this PR propagates to the CLI flow.
- `.claude/rules/go-style.md`, `.claude/rules/testing.md`, `.claude/rules/pr-creation.md`, `CLAUDE.md`.
- RFC 7591 §3.2.1 (registration expiry), RFC 7636 / OAuth 2.1 (S256 PKCE), RFC 8252 §7.3 (loopback redirect for native CLI clients).

tgrunnagle:
## Summary

Sub-issue 4b of #5145. The CLI OAuth flow in `pkg/auth/discovery::PerformOAuthFlow` used to call `oauthproto.RegisterClientDynamically` directly, so it bypassed the review-property behaviours added during #5042 (S256 PKCE gating, RFC 7591 §3.2.1 expiry-driven refetch, bearer-token transport with redirect refusal, panic recovery, singleflight deduplication). This PR routes that call site through the shared `pkg/auth/dcr` resolver introduced in sub-issue 4a (PR #5198) so both consumers — the embedded authserver and the CLI — share the same hardened DCR path.

To make the resolver genuinely profile-agnostic, `pkg/auth/dcr` now exposes a `Request` struct carrying exactly the fields the resolver reads. The embedded-authserver adapter helpers move into `pkg/authserver/runner/dcr_adapter.go` where they belong by ownership, and `ResolveCredentials` no longer imports authserver or upstream domain types.

Closes #5219.

## Large PR Justification

The diff is 1903 insertions / 845 deletions across 18 files (872 / 414 excluding tests). Production-code changes that cannot be split:

- **Neutralising `ResolveCredentials`'s API is a single atomic step.** Replacing `*authserver.OAuth2UpstreamRunConfig` with the profile-neutral `dcr.Request` envelope and lifting `NeedsDCR` / `ConsumeResolution` / `ApplyResolutionToOAuth2Config` into `pkg/authserver/runner/dcr_adapter.go` must land together — splitting them would leave the resolver importing authserver types between PRs, which is the architectural drift this sub-issue exists to close.
- **The CLI migration is the second consumer that makes the neutralised API load-bearing.** Migrating `pkg/auth/discovery::PerformOAuthFlow` onto the shared resolver in the same PR is what proves the neutralisation is correct; a follow-up CLI migration would leave the API change unjustified and unexercised at merge time.

Test churn accounts for the remainder. `pkg/auth/dcr/resolver_test.go` was rewritten to the new `Request` API (the old test file constructed `OAuth2UpstreamRunConfig` fixtures throughout); the consumer-side tests for the lifted helpers moved to `pkg/authserver/runner/dcr_adapter_test.go` to follow the production code. New `pkg/auth/discovery/dcr_resolver_test.go` pins the CLI's inherited review properties (S256 PKCE gating, redirect refusal, singleflight dedup, fallback error verbatim) — these are the load-bearing acceptance criteria for sub-issue 4b and cannot be deferred to a follow-up.

## Type of change

## Test plan

New tests in `pkg/auth/discovery/dcr_resolver_test.go` cover the CLI's inherited properties:
- S256 PKCE gating — CLI surfaces a clear resolver error when upstream advertises only `plain`.
- Bearer-token transport redirect refusal — registration POSTs do not follow redirects.
- Singleflight deduplication — concurrent invocations with the same `(issuer, scopes, redirectURI)` produce one `/register` call.
- Fallback error preservation — `TestHandleDynamicRegistration_MissingRegistrationEndpoint` pins the verbatim "configure with `--remote-auth-client-id` / `--remote-auth-client-secret`" message.

## API Compatibility

No operator API surface is touched. The `pkg/auth/dcr` package was added in 4a (#5198) and has no external consumers; the `ResolveCredentials` signature change is internal.

## Changes

| File | Change |
|------|--------|
| `pkg/auth/dcr/request.go` | New profile-neutral `Request` input type for `ResolveCredentials`. |
| `pkg/auth/dcr/resolver.go` | `ResolveCredentials` now takes `*dcr.Request`; no longer imports authserver / upstream. |
| `pkg/auth/dcr/store.go` | New `CloseableCredentialStore` interface; `NewInMemoryStore` returns it so `defer Close()` is compile-time safe. |
| `pkg/auth/discovery/discovery.go` | `handleDynamicRegistration` builds a `dcr.Request` and calls `dcr.ResolveCredentials` instead of `oauthproto.RegisterClientDynamically`. Loopback redirect URI (RFC 8252 §7.3) and CLI-flag fallback error preserved. |
| `pkg/auth/discovery/dcr_request.go` | Deleted — superseded by `pkg/auth/dcr.Request`. |
| `pkg/auth/discovery/dcr_resolver_test.go` | New tests pinning the CLI's inherited review properties. |
| `pkg/authserver/runner/dcr_adapter.go` | Embedded-authserver-specific adapter helpers (`needsDCR`, `newDCRRequest`, `consumeResolution`, `applyResolutionToOAuth2Config`) moved here from `pkg/auth/dcr`. |
| `pkg/authserver/runner/embeddedauthserver.go` | Call sites updated to use the adapter and pass `*dcr.Request` into `dcr.ResolveCredentials`. |
| `pkg/authserver/runner/dcr_store.go` | Deleted — folded into the shared `pkg/auth/dcr/store.go` (4a follow-through). |

No. The CLI OAuth flow's observable behaviour is unchanged on the happy path. The new error paths (S256 gating, redirect refusal) are inherited security improvements that surface as clearer error messages when an upstream provider misbehaves.

**Persistence model — option (b) from the issue.** The resolver runs against an in-memory `dcr.CredentialStore` scoped to one `PerformOAuthFlow` invocation. Cross-invocation persistence is handled outside the resolver by `pkg/auth/remote/handler.go`'s existing `CachedClientID` / `CachedClientSecretRef` fields, which already preserved cross-invocation reuse and continue to do so unchanged. Option (a) — wrapping the secret provider as a `CredentialStore` adapter — was rejected as out-of-scope churn; the trade-off is documented at the wiring site in `handleDynamicRegistration`. A consequence worth flagging: cross-invocation **expiry refetch** lives in the remote handler, not in the resolver, on the CLI path. Closing that loop is a natural follow-up once option (a) is wired.

**PublicClient flag.** A new bool on `dcr.Request` tells the resolver to register as a public PKCE client (`token_endpoint_auth_method=none`). The S256 gate still fires — the CLI surfaces a clear resolver error rather than silently downgrading when the upstream advertises only `plain` PKCE.

**Resolver invariant — code review only.** The issue's acceptance criterion for a CI grep guard against direct `oauthproto.RegisterClientDynamically` calls outside `pkg/auth/dcr` is intentionally **out of scope for this PR**. A `task check-dcr-isolation` Taskfile target and matching workflow step were prototyped and then removed (commit fd66a1fe) per reviewer preference. The resolver boundary is enforced by code review alone going forward.


pkg/auth/dcr/resolver.go
```diff
@@ -89,21 +86,17 @@ import (
...
+// redirect URI, same scopes hash) will share a single in-flight
+// registration even if they semantically want different client profiles.
+// The two current call sites do not collide by construction — the embedded
+// authserver's redirect URI lives on the AS origin, and the CLI flow's
+// redirect URI lives on a loopback (http://localhost:{port}/callback per
+// RFC 8252 §7.3). A future consumer that defaults its redirect URI into
+// either of those spaces would silently coalesce; if a third profile is
+// added the flight key MUST gain a consumer-identifier component so a
+// collision is impossible rather than improbable.
 var dcrFlight singleflight.Group
```
tgrunnagle, line 103: **[MEDIUM] `PublicClient` absent from singleflight and cache keys** (Consensus: 9/10)

The flight key (and `storage.DCRKey`) is `(Issuer, RedirectURI, ScopesHash)`. `req.PublicClient` is not part of either. The singleflight closure captures the **leader's** `req` and returns the leader's `*Resolution` to every follower — so two callers with different `PublicClient` values whose keys happen to match would receive the same registration. Today's two consumers don't collide by construction (CLI loopback redirect vs AS-origin redirect), but the comment two lines down already calls this out: *"if a third profile is added the flight key MUST gain a consumer-identifier component so a collision is impossible rather than improbable."*

The same gap also affects the persisted Redis-backed cache for the embedded authserver: if option-(a) (wrap the CLI's secret provider as a CredentialStore) ever lands, a CLI invocation could read a confidential-client resolution the AS persisted.

**Recommendation**: include `PublicClient` (or a typed `ConsumerKind` enum tag on `dcr.Request`) in both `storage.DCRKey` and `flightKeyOf`. Cost: one extra field + one extra concat. Benefit: the safety invariant becomes structural rather than review-only.

*Raised by: oauth-security (HIGH), concurrency (MEDIUM), architecture (MEDIUM), go-correctness (LOW), code-quality (INFO)*

tgrunnagle (commented):
**Agents consulted**: oauth-security, go-correctness, architecture, concurrency, test-coverage, general-code-quality (codex cross-review skipped — CLI not installed)

### Consensus Summary

| 1 | `PublicClient` absent from singleflight + cache keys | 9/10 | MEDIUM | Discuss |
| 2 | `applyExplicitEndpointOverrides` bypasses URL validation | 7/10 | MEDIUM | Fix |
| 3 | Two-call `consumeResolution`/`applyResolutionToOAuth2Config` invariant not type-enforced | 8/10 | MEDIUM | Discuss |
| 4 | `SanitizeErrorForLog` is http(s)-only — redis-go error chain | 7/10 | MEDIUM | Discuss |
| 5 | `DCRKey.Issuer` doc contradicts new CLI usage | 8/10 | MEDIUM | Fix |
| 6 | `_ = store.Close()` swallows error without comment/log | 7/10 | LOW | Fix |
| 7 | `TestNewDCRRequest` missing env-var IAT branch | 7/10 | LOW | Fix |
| 8 | `TestConsumeResolution` does not pin nil-input no-op | 6/10 | LOW | Optional |
| 9 | `LogStepError` "upstream" attribute shape varies per consumer | 7/10 | LOW | Discuss |
| 10 | Direct-registration + `PublicClient=true` error msg is misleading | 6/10 | LOW | Optional |
| 11 | Wall-clock `ClientSecretExpiresAt` comparison lacks skew tolerance | 6/10 | LOW | Optional |
| 12 | PR body still reads "DRAFT - not ready for review" | 7/10 | LOW | Fix |
| 13 | Option-(b) rationale duplicated across two function docs | 6/10 | LOW | Optional |

### Overall

This is a well-executed migration of the CLI OAuth flow onto the shared `pkg/auth/dcr` resolver. The package boundary is structurally clean (resolver no longer imports any authserver/discovery types — verified), the new `dcr.Request` envelope is genuinely profile-neutral, and the new inherited-property tests (S256 gate, redirect refusal, singleflight dedup) pin exactly the security-relevant behaviours that motivated the migration. The PR description, doc comments, and orientation prose are unusually good — a cold reviewer can pick up the architectural picture in a single pass.

The consensus findings cluster around one theme: **correctness invariants encoded in comments rather than types**. The singleflight + cache keys do not include `PublicClient`, the two-call resolution invariant relies on review, the sanitiser is scheme-specific despite the generic name, and `DCRKey.Issuer`'s field doc still claims "local issuer ... NOT the upstream's" even though the CLI now puts the upstream issuer there. Each one is a small refactor away from being structural. Current call sites do not exhibit the failure modes by construction; the risk is forward-looking, not active.

None of the findings rise to a hard blocker. Recommended priorities are #1 (PublicClient in keys), #3 (two-call consolidation), and #5 (DCRKey doc fix) — all cheap, all close defense-in-depth gaps the author already calls out in inline comments.

For consistency, please note: the authenticated reviewer and PR author are the same user, so this review uses `COMMENT` rather than `REQUEST_CHANGES` — this is not a draft-blocker assessment, it's an artefact of GitHub's policy.

### Additional findings (file not in this PR's diff, or line outside diff hunks)

- **[MEDIUM] `DCRKey.Issuer` doc drift** — `pkg/authserver/storage/types.go:122-130` documents `Issuer` as "the local issuer of the embedded authorization server that performed the registration, NOT the upstream's". The new CLI call site at `pkg/auth/discovery/discovery.go:712-713` puts the **upstream's** issuer URL there. The new `dcr.Request.Issuer` doc accommodates both consumers, but the underlying storage type's field doc still claims the old single-consumer semantic. Update `DCRKey.Issuer` to reflect the dual semantic, e.g. *"For the embedded authserver this is its own (local) issuer; for the CLI flow this is the upstream's issuer because the CLI has no separate logical issuer of its own. The cache key (Issuer, RedirectURI, ScopesHash) keeps the two consumers' entries apart via the RedirectURI component."*
- **[MEDIUM] `SanitizeErrorForLog` is http(s)-only** — `pkg/auth/dcr/resolver.go:481-538` (outside this PR's diff hunks). The function name reads generic, but the implementation only strips userinfo/query/fragment from `http(s)://` URLs. On the embedded-authserver path, `cache.Get`/`Put` errors from `storage/redis.go` flow into `LogStepError`. If a future operator (or redis-go release) puts credentials in a sentinel/cluster URL, those credentials hit the slog.Error attribute. Either rename to `sanitizeHTTPURLs` so the scope is unambiguous, or extend the regex to also match `redis(s)?://`. Add a small test asserting a redis-style error string with credentials is sanitised.
- **[LOW] `LogStepError` "upstream" attribute is inconsistent per consumer** — `pkg/auth/dcr/resolver.go:452-479` (outside this PR's diff hunks). Embedded-authserver path passes `rc.Name` (operator-defined short name); CLI path passes `issuer` (upstream issuer URL). Operators reading mixed logs see two different shapes under the same key. Either rename the parameter to `upstreamID` and document it as opaque, or accept `slog.Attr` pairs so callers can label (`"upstream_name"` vs `"upstream_issuer"`).
- **[LOW] `ClientSecretExpiresAt` wall-clock comparison lacks skew tolerance** — `pkg/auth/dcr/resolver.go:641` (outside this PR's diff hunks). `time.Now().After(cached.ClientSecretExpiresAt)` is exact. NTP step on resume or AS-clock-ahead skew leads to either serving an expired secret (already broken at the AS) or refetching too early. Consider `time.Now().Add(30 * time.Second).After(cached.ClientSecretExpiresAt)` as defense-in-depth.
- **[LOW] PR body still reads "DRAFT - not ready for review"** — The PR is in draft state and the body header says "DRAFT - not ready for review". Since you are explicitly requesting a review now, either flip out of draft or remove the header so other reviewers landing on this PR aren't confused about its status.

### Dropped findings appendix (consensus < 7)

The following observations were raised by individual agents but did not meet the consensus threshold. Listed for auditability only — no action requested:

- Regex `(?i)https?://[^\s"']+` can match across a comma (security LOW, score 5). Failure mode is "did not sanitise", not corrupted output.
- `selectTokenEndpointAuthMethod` with empty `serverSupported` + `publicClient=true` trusts upstream to accept "none" (security LOW, score 5).
- `CloseableCredentialStore` loses Close capability when widened to base interface (architecture LOW, score 6). No occurrence in this PR.
- Hardcoded `CallbackPort: 8765` in CLI tests (test INFO, score 5). Tests don't bind the port — acceptable today.
- `dcr.Key` is generically named (quality INFO, score 5).
- `flightKeyOf(key)` is a free function where a method would read more idiomatically (quality LOW, score 5).
- Long doc comments on `Request`/`Resolution` partially restate code (quality LOW, score 5).
- `handleDynamicRegistration` doc preamble is 26 lines for a 45-line body (quality LOW, score 5).
- `ResolveCredentials` takes `*Request` by pointer; could take by value (architecture LOW, score 5).
- `PublicClient bool` tri-state suggestion to fail-loud on unset (security INFO, score 5).
- CLI cross-invocation expiry refetch not in resolver's loop (architecture INFO, score 6) — explicitly documented as option-(b) trade-off.

---

pkg/auth/dcr/resolver.go
```diff
@@ -771,14 +679,14 @@ func lookupCachedResolution(
...
 // applyExplicitEndpointOverrides overwrites the discovered
 // authorizationEndpoint / tokenEndpoint in endpoints with explicit values
-// from rc when rc specifies them. Explicit caller configuration always wins
-// over discovery.
-func applyExplicitEndpointOverrides(endpoints *dcrEndpoints, rc *authserver.OAuth2UpstreamRunConfig) {
-	if rc.AuthorizationEndpoint != "" {
-		endpoints.authorizationEndpoint = rc.AuthorizationEndpoint
+// from req when req specifies them. Explicit caller configuration always
+// wins over discovery.
+func applyExplicitEndpointOverrides(endpoints *dcrEndpoints, req *Request) {
```
tgrunnagle: **[MEDIUM] `applyExplicitEndpointOverrides` bypasses `validateUpstreamEndpointURL`** (Consensus: 7/10)

`endpointsFromMetadata` runs `validateUpstreamEndpointURL` against discovered `authorization_endpoint` / `token_endpoint`. This function then overwrites them with `req.AuthorizationEndpoint` / `req.TokenEndpoint` **without re-validating**. The validation contract documented on `validateUpstreamEndpointURL` claims "every point where an endpoint URL enters the resolver from outside" — the override path violates that promise.

Today's two callers (`newDCRRequest`, `resolveDCRCredentials`) populate these from already-validated sources, so this is defense-in-depth, not an active bug. But the cost to close the gap is trivial.

```suggestion
func applyExplicitEndpointOverrides(endpoints *dcrEndpoints, req *Request) error {
	if req.AuthorizationEndpoint != "" {
		if err := validateUpstreamEndpointURL(req.AuthorizationEndpoint, "authorization_endpoint"); err != nil {
			return fmt.Errorf("dcr: explicit %w", err)
		}
		endpoints.authorizationEndpoint = req.AuthorizationEndpoint
	}
	if req.TokenEndpoint != "" {
		if err := validateUpstreamEndpointURL(req.TokenEndpoint, "token_endpoint"); err != nil {
			return fmt.Errorf("dcr: explicit %w", err)
		}
		endpoints.tokenEndpoint = req.TokenEndpoint
	}
	return nil
}
```

(Signature change requires updating the one caller at `registerAndCache` to propagate the error via `newDCRStepError(dcrStepMetadata, …)`.)

*Raised by: oauth-security*

pkg/auth/dcr/resolver.go
```diff
@@ -1194,7 +1104,36 @@ func validateUpstreamEndpointURL(rawURL, label string) error {
...
+// publicClient: when true (CLI flow), the function returns "none" if and
+// only if S256 PKCE is advertised. Any other outcome is an error — the
+// caller has declared intent and the resolver must not silently switch to
+// a confidential-client method.
+func selectTokenEndpointAuthMethod(serverSupported, codeChallengeMethodsSupported []string, publicClient bool) (string, error) {
+	pkceS256Advertised := slices.Contains(codeChallengeMethodsSupported, oauthproto.PKCEMethodS256)
+
+	if publicClient {
+		if !pkceS256Advertised {
+			return "", fmt.Errorf(
```
tgrunnagle, line 1149: **[LOW] Misleading error for direct-registration + `PublicClient=true`** (Consensus: 6/10)

When a caller passes `RegistrationEndpoint` (no `DiscoveryURL`) and `PublicClient=true`, the resolver correctly refuses — but via this branch with the message "public client requested but upstream does not advertise S256 in code_challenge_methods_supported (got [])". Operators who never asked for discovery will look for a metadata document that doesn't exist in their config.

**Recommendation**: detect the `PublicClient=true && RegistrationEndpoint != "" && DiscoveryURL == ""` case earlier (in `validateResolveInputs`) and return: *"PublicClient=true requires DiscoveryURL so the S256 PKCE gate can be evaluated; got RegistrationEndpoint only"*. The behaviour is correct today; only the diagnostic needs sharpening.

*Raised by: oauth-security*

pkg/authserver/runner/dcr_adapter.go
```diff
@@ -0,0 +1,158 @@
...
+// The split exists because buildPureOAuth2Config intentionally retains a
+// narrow file-or-env contract (no DCR awareness) and because OAuth2's
+// ClientSecret on the run-config is modelled as a reference rather than
+// an inline string. Any future output path from OAuth2UpstreamRunConfig
+// to upstream.OAuth2Config must call BOTH consumeResolution (run-config
+// side) AND applyResolutionToOAuth2Config (built-config side) to get a
+// fully-resolved DCR client. Forgetting the second call leaves
+// ClientSecret empty and produces silent auth failures at request time —
+// the type system does not enforce the pair, so the invariant lives here.
+func applyResolutionToOAuth2Config(cfg *upstream.OAuth2Config, res *dcr.Resolution) {
```
tgrunnagle: **[MEDIUM] Two-call invariant not type-enforced** (Consensus: 8/10)

The doc on this function acknowledges the gap: *"Forgetting the second call leaves ClientSecret empty and produces silent auth failures at request time — the type system does not enforce the pair, so the invariant lives here."* The only call site (`buildUpstreamConfigs`) is correct, but a future call site that copies the `consumeResolution` pattern would silently produce a no-secret config — exactly the "silent no-op that breaks callers" pattern `.claude/rules/go-style.md` warns against.

**Recommendations** (pick one):
1. Collapse into a single `applyDCRResolution(rc *OAuth2UpstreamRunConfig, cfg *upstream.OAuth2Config, res *dcr.Resolution)` that writes everything atomically.
2. Keep the split but add a regression test that exercises only `consumeResolution` and asserts `cfg.ClientSecret == ""` — a tripwire so any future code-shaped refactor has an explicit failure mode.

*Raised by: architecture (MEDIUM), go-correctness (LOW)*

pkg/auth/discovery/discovery.go
```diff
@@ -585,31 +586,183 @@ func PerformOAuthFlow(ctx context.Context, issuer string, config *OAuthFlowConfi
...
+	}
+	if resolution.TokenEndpoint != "" {
+		config.TokenURL = resolution.TokenEndpoint
 	}
 
 	return nil
 }
 
-// getDiscoveryDocument retrieves the OIDC discovery document
+// resolveDCRCredentials routes the CLI-flow DCR registration through the
```
tgrunnagle, line 666: **[LOW] Option-(b) rationale duplicated across `handleDynamicRegistration` and `resolveDCRCredentials`** (Consensus: 6/10)

Both function docs carry multi-paragraph explanations of the option-(b) persistence trade-off in overlapping vocabulary. They are not duplicates (one is about persistence, the other about the redundant discovery fetch), but they share substantial conceptual ground and reference the same sub-issue numbering. A future maintainer modifying option-(b) behaviour must update both.

**Recommendation**: pick this function (where the trade-off actually manifests) as the canonical site. In `handleDynamicRegistration`, summarise in 2-3 lines and link here.

*Raised by: code-quality*

pkg/auth/discovery/discovery.go
```diff
@@ -585,31 +586,183 @@ func PerformOAuthFlow(ctx context.Context, issuer string, config *OAuthFlowConfi
...
+	// resolver would have nothing to fetch, surfacing a clean validation
+	// error rather than dereferencing an empty URL.
+	if discoveredDoc.Issuer != "" {
+		req.DiscoveryURL = strings.TrimRight(discoveredDoc.Issuer, "/") + "/.well-known/openid-configuration"
+	} else {
+		req.RegistrationEndpoint = discoveredDoc.RegistrationEndpoint
+	}
+
+	store := dcr.NewInMemoryStore()
+	defer func() { _ = store.Close() }()
```
tgrunnagle: **[LOW] `_ = store.Close()` silently drops the error** (Consensus: 7/10)

`.claude/rules/go-style.md`: *"Comment ignored errors — explain why and typically log them."* `inMemoryStore.Close` returns nil today, but a future change to `storage.MemoryStorage.Close` (e.g. timeout on cleanup goroutine teardown) would surface a real error and the CLI would drop it without a trace.

```suggestion
	defer func() {
		if err := store.Close(); err != nil {
			slog.Debug("dcr: in-memory store close failed", "error", err)
		}
	}()
```

*Raised by: concurrency*

pkg/authserver/runner/dcr_adapter_test.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tc.expected, needsDCR(tc.rc))
+		})
+	}
+}
+
+func TestConsumeResolution_RespectsExplicitEndpoints(t *testing.T) {
```
tgrunnagle, line 55: **[LOW] `TestConsumeResolution` does not pin the nil-input no-op contract** (Consensus: 6/10)

`consumeResolution` returns early on `rc == nil || res == nil`, but no test exercises that early-return. `TestApplyResolutionToOAuth2Config` does cover nil cfg and nil res (lines 145-147), so the asymmetry is conspicuous. A regression that dereferenced `rc` before the nil check would crash the embedded-authserver boot in a hard-to-diagnose way.

**Recommendation**: add a one-line subtest that calls `consumeResolution(nil, &dcr.Resolution{})` and `consumeResolution(&authserver.OAuth2UpstreamRunConfig{}, nil)` and asserts no panic. Cheap, high-signal.

*Raised by: test-coverage*

pkg/authserver/runner/dcr_adapter_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// nil-safe: nil cfg or nil res is a no-op rather than a crash.
+	applyResolutionToOAuth2Config(nil, res)
+	applyResolutionToOAuth2Config(cfg, nil)
+}
+
+// TestNewDCRRequest covers the OAuth2UpstreamRunConfig → dcr.Request
+// translation, including the file-based InitialAccessToken resolution that
+// previously lived inside the resolver.
+func TestNewDCRRequest(t *testing.T) {
```
tgrunnagle, line 235: **[LOW] `TestNewDCRRequest` missing env-var `InitialAccessToken` branch** (Consensus: 7/10)

The four cases cover file-based IAT, no IAT, nil run-config, and missing dcr_config. The env-var sibling (`InitialAccessTokenEnvVar`) is not exercised end-to-end through `newDCRRequest` → `resolveSecret`. The underlying `resolveSecret` is independently covered in `embeddedauthserver_test.go::TestResolveSecretWithEnvVar`, but the adapter-level wiring boundary is not pinned — a refactor that swapped the argument order or accidentally read the wrong field on DCRConfig would not be caught.

**Recommendation**: add a fifth table case that sets `DCRConfig.InitialAccessTokenEnvVar` via `t.Setenv` on a unique env-var name (drop `t.Parallel()` for that subcase, matching the `TestResolveSecretWithEnvVar` pattern) and asserts the resulting `req.InitialAccessToken`.

*Raised by: test-coverage*

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in c805aa37.

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in c805aa37.

pkg/authserver/runner/dcr_adapter.go
tgrunnagle: Addressed in c805aa37 via option 2 (tripwire test). `TestConsumeResolution_AloneLeavesClientSecretEmpty` pins the failure mode so any future refactor that drops the paired `applyResolutionToOAuth2Config` call fails at test time rather than at upstream auth time on a deployed binary.

pkg/auth/discovery/discovery.go
tgrunnagle: Addressed in c805aa37.

pkg/authserver/runner/dcr_adapter_test.go
tgrunnagle: Addressed in c805aa37 via `TestNewDCRRequest_EnvVarInitialAccessToken`. Kept as a standalone test (not a table case) because `t.Setenv` is incompatible with `t.Parallel`, matching the pattern in `embeddedauthserver_test.go::TestResolveSecretWithEnvVar`.

pkg/auth/dcr/resolver.go
tgrunnagle: Deferring — the diagnostic is suboptimal but the behaviour is correct, and threading `code_challenge_methods_supported` through `AuthServerInfo` (the natural way to surface this earlier) is the follow-up already noted in `resolveDCRCredentials`'s doc comment. Better to fix the diagnostic alongside that refactor than to harden the error string twice.

pkg/auth/discovery/discovery.go
tgrunnagle: Deferring — the two doc blocks cover overlapping but non-identical ground (`handleDynamicRegistration` documents the persistence model, `resolveDCRCredentials` documents the redundant-fetch trade-off), and consolidating risks losing reader-orientation context at each call site. Re-evaluate when threading `code_challenge_methods_supported` through `AuthServerInfo` lets us drop the redundant-fetch trade-off entirely.

pkg/authserver/runner/dcr_adapter_test.go
tgrunnagle: Partially obsolete post-rebase onto `dcr-4a_issue_5145`. The 4a base flipped `consumeResolution` to value-in / value-out, so nil-rc is structurally impossible — the function takes `authserver.OAuth2UpstreamRunConfig` by value. The nil-resolution branch is already pinned by `TestConsumeResolution_NilResolutionIsNoop`.

tgrunnagle:
Responding to the review-body findings (no inline anchor):

- **MEDIUM — \`DCRKey.Issuer\` doc drift** (`pkg/authserver/storage/types.go:122-130`): addressed in c805aa37. The field doc now describes both consumer profiles (embedded authserver's own issuer vs CLI's upstream issuer) and the cache-key non-collision invariant via the RedirectURI component.

- **MEDIUM — \`SanitizeErrorForLog\` http-only** (`pkg/auth/dcr/resolver.go:481-538`): addressed in c805aa37. `queryStrippingPattern` now matches `(?i)(?:https?|rediss?)://` so redis-go error chains on the embedded-authserver `Get`/`Put` critical path get their userinfo / query / fragment stripped. Added `TestSanitizeErrorForLog` cases pinning `redis://` and `rediss://` URL userinfo redaction.

- **LOW — \`LogStepError\` "upstream" attribute shape varies**: deferring. Two consumers populate the field differently today (`rc.Name` for embedded authserver, upstream issuer URL for the CLI). Renaming to `upstreamID` is correct but cosmetic — operators reading mixed logs already have to disambiguate by `step` and the other structured attributes. A typed-slog.Attr API change would be invasive enough to warrant its own PR; the current shape is documented and the observability cost is low.

- **LOW — \`ClientSecretExpiresAt\` lacks skew tolerance** (`pkg/auth/dcr/resolver.go:641`): deferring. The 30-second early-refetch buffer is a real ergonomics improvement but defense-in-depth on top of an already-defensive refetch path (RFC 7591 §3.2.1 expiry handling). Best paired with the broader review of clock-sensitive operations across the auth path rather than landed in isolation.

- **LOW — PR body "DRAFT" header**: removed via `gh pr edit`.

pkg/auth/discovery/discovery.go
```diff
@@ -585,31 +586,193 @@ func PerformOAuthFlow(ctx context.Context, issuer string, config *OAuthFlowConfi
...
+// CachedClientID / CachedClientSecretRef before this code path runs and
+// short-circuits to a refresh-token flow when a usable cached client
+// exists.
+//
+// One consequence of option (b) is that the resolver's RFC 7591 §3.2.1
+// expiry-driven refetch does NOT participate in the CLI's cross-
+// invocation persistence loop: each PerformOAuthFlow call builds a fresh
+// in-memory store, so a "cached but expired" entry from the previous
+// invocation never reaches the resolver. Cross-invocation expiry is the
+// remote handler's responsibility (its HasCachedClientCredentials gate
```
jhrozek: nit: this comment says cross-invocation expiry is handled by the remote handler's `HasCachedClientCredentials` gate, but that gate only checks `CachedClientID != ""` — it doesn't consult `CachedSecretExpiry`. So a cached-but-expired client gets reused on the next invocation and surfaces as a token-endpoint failure rather than a clean DCR re-registration. Either tighten the gate to also check `CachedSecretExpiry`, or update the comment to acknowledge cross-invocation expiry is not currently enforced (and is open follow-up work).

pkg/authserver/storage/redis_keys.go
```diff
@@ -97,23 +97,29 @@ func redisProviderKey(prefix, providerID, providerSubject string) string {
...
+// Store path) already rejects an empty ScopesHash, and callers are required
+// to compute the hash via storage.ScopesHash. Length-prefix collision-safety
+// is preserved on the leading segments either way.
 func redisDCRKey(prefix string, key DCRKey) string {
-	return fmt.Sprintf("%s%s:%d:%s:%d:%s:%s",
+	publicFlag := "0"
+	if key.PublicClient {
+		publicFlag = "1"
+	}
+	return fmt.Sprintf("%s%s:%d:%s:%d:%s:%s:%s",
```
jhrozek: nit/question: appending the `:0`/`:1` flag changes the Redis key format, so any pre-existing DCR entries become unreachable after upgrade. For entries with `client_secret_expires_at = 0` the row has no TTL and would just sit orphaned in Redis while the authserver re-registers on startup. Has the Redis-backed `DCRCredentialStore` actually shipped to anyone relying on persisted entries? If not, no migration needed; if it has, worth either a release note or a one-time read-side fallback to the legacy key format.

pkg/auth/dcr/resolver.go
```diff
@@ -837,7 +767,7 @@ func performRegistration(
 
 	// Call exactly once — no retry loop. Step 2g will add retry/backoff at a
```
jhrozek: nit: "Step 2g" doesn't map to anything searchable in the repo — grep returns just this line and a sibling at resolver.go:223. Could you link the tracking issue or rephrase to plain English so the next reader has something to act on?

jhrozek (dismissed):
Looks good — a few nits inline, all non-blocking.

pkg/auth/discovery/discovery.go
tgrunnagle: Addressed in 2621269c — opted for the comment-update path. The handleDynamicRegistration option-(b) doc now states that HasCachedClientCredentials only checks CachedClientID and does NOT consult CachedSecretExpiry, so an expired-but-still-cached client gets reused on the next invocation; tightening the gate is open follow-up rather than this PR's scope.

pkg/authserver/storage/redis_keys.go
tgrunnagle: Addressed in 182188a1 — dropped PublicClient from DCRKey and redisDCRKey entirely. Today's two consumers (embedded authserver = AS-origin redirect URI, CLI = RFC 8252 loopback) use disjoint RedirectURI address spaces, so the RedirectURI component alone separates public-client and confidential-client profiles at both the persistent cache and singleflight layers. The Redis key format is unchanged from main, so no migration story is needed for existing entries.

pkg/auth/dcr/resolver.go
tgrunnagle: Addressed in 2621269c. Both "Step 2g" references were stale forward-pointers from an earlier internal numbering scheme; rephrased to describe today's reality — staleness observability already lives in lookupCachedResolution (warns when age exceeds dcrStaleAgeThreshold), and the no-retry comment now spells out that any future retry/backoff belongs above ResolveCredentials so the singleflight + cache semantics stay simple.

pkg/auth/discovery/discovery.go
tgrunnagle: Correction on the hash above — the original commit was amended with a spell-check fix before push and is now f01ac635.

pkg/auth/dcr/resolver.go
tgrunnagle: Correction on the hash above — the original commit was amended with a spell-check fix before push and is now f01ac635.

jhrozek (approved):

PR #5255, opened 2026-05-12
Add registry AuthDefaulter extension point for OIDC discovery

reyortiz3:
## Summary

- **Why**: `thv config set-registry <url>` and `PUT /api/v1beta/registry/{name}` currently require the caller to pass `--issuer`/`--client-id` (or an `auth` body) explicitly. Downstream consumers (enterprise builds, GUIs, future managed deployments) want to fill these in automatically from an OIDC discovery document, but there is no extension point to plug that resolver into either entry point.
- **What**: Introduces a pluggable `registry.AuthDefaulter` in `pkg/registry/`. When the user/API caller omits auth, the CLI and the API handler consult the registered defaulter (if any) and use what it returns. OSS builds register no defaulter and behave exactly as before.

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`go test ./pkg/registry/... ./pkg/api/v1/... ./cmd/thv/app/...`)
  - New: `TestResolveAuthDefaults`, `TestRegisterAuthDefaulter` (pkg/registry)
  - New: `TestUpdateRegistry_AuthDefaulterFillsMissingAuth` (pkg/api/v1, uses a mocked OIDC discovery server)
- [ ] E2E tests (`task test-e2e`) — not run; behavior unchanged for OSS builds
- [ ] Manual testing — not applicable

No (for OSS). The new extension point is dormant unless a build registers a defaulter via `registry.RegisterAuthDefaulter`. Existing CLI/API contracts are unchanged: explicit `--issuer`/`--client-id` or `auth` body still wins, and omitting them with no defaulter registered still produces a registry with no OAuth (the prior behavior).

- Errors from the defaulter fall back to "no auth" so the call site preserves the legacy lenient behavior — we don't want a flaky discovery server to break unrelated CLI commands.
- `RegisterAuthDefaulter(nil)` clears the active defaulter (needed by tests).
- The defaulter sees `(issuer, clientID)` only — audience and scopes still come from the caller's request or the defaults already in `pkg/registry/auth`.

### DIFF
--- a/cmd/thv/app/config.go
+++ b/cmd/thv/app/config.go
@@ -16,7 +16,6 @@ import (
 	"github.com/stacklok/toolhive/pkg/registry/auth"
 )
 
-
 var configCmd = &cobra.Command{
 	Use:   "config",
 	Short: "Manage application configuration",
--- a/cmd/thv/app/config.go
+++ b/cmd/thv/app/config.go
@@ -16,6 +16,7 @@ import (
 	"github.com/stacklok/toolhive/pkg/registry/auth"
 )
 
+
 var configCmd = &cobra.Command{
 	Use:   "config",
 	Short: "Manage application configuration",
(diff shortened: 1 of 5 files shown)


blkt (approved):

jhrozek:
@reyortiz3 the api-workloads test is failing consistently. There were timeout issues that I tried to address in PR #5275 so it might be worth merging that one and rebasing, but worth taking a look in case the tests uncover a real issue

PR #5256, opened 2026-05-12
Fix audit events logged as INFO+2 instead of AUDIT

### LINKED ISSUE #4296: Audit events logged at non-standard level INFO+2, breaking level detection in log pipelines
## Summary

vmcp (and potentially other components) log audit events at \`"level":"INFO+2"\` — a Go \`log/slog\` artifact from using \`slog.LevelInfo + 2\` as the verbosity level rather than a standard named level. This breaks level detection in all common log aggregation systems.

## Observed behaviour

Audit event log lines contain \`"level":"INFO+2"\`:

\`\`\`json
{"time":"2026-03-20T16:33:03.481912959Z","level":"INFO+2","msg":"audit_event","audit_id":"...","type":"mcp_tools_list",...}
\`\`\`

Standard log levels recognised by log aggregators (Loki \`detected_level\`, Elasticsearch, Splunk, etc.) are: \`trace\`, \`debug\`, \`info\`, \`warn\`, \`warning\`, \`error\`, \`fatal\`, \`critical\`. The value \`INFO+2\` matches none of these, so audit events appear as \`unknown\` level in Grafana's Explore Logs view.

This is especially problematic because audit events are the primary log type users want to filter and alert on.

## Impact

- Audit events show as \`detected_level=unknown\` in Loki/Grafana
- Users cannot write level-based LogQL filters like \`{service="vmcp"} | level = "audit"\` or use level facets in Explore Logs
- Any log pipeline with a level-based filter (e.g. "only ship WARN and above") will silently drop audit events

## Suggested fix

Use a standard level or a properly named custom level for audit events. Options:

1. **Log at \`INFO\`** — simplest, audit events are informational by nature
2. **Define a named custom level** — e.g. \`AUDIT\` at a specific numeric value, which log/slog supports via \`slog.Level\` with a custom \`String()\` method

Related: the plain-text startup line issue is tracked in #4295.

kimjune01:
## Summary

- Add `ReplaceAttr` to the audit logger's JSON handler so the custom audit level renders as `"AUDIT"` instead of `"INFO+2"`
- Fixes compatibility with log aggregation systems (Loki, Elasticsearch) that expect standard level names for filtering

Fixes #4296

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Added two tests: one verifying audit events render as `"AUDIT"`, another verifying standard levels (INFO, WARN) are preserved. `go test ./pkg/audit/... -count=1`: 28 tests passed.

## API Compatibility

No. Audit log output changes from `"INFO+2"` to `"AUDIT"` in JSON logs, but this is a bug fix restoring intended behavior.


pkg/audit/auditor.go
```diff
@@ -56,6 +56,16 @@ func NewAuditLogger(w io.Writer) *slog.Logger {
 
 	handler := slog.NewJSONHandler(w, &slog.HandlerOptions{
 		Level: LevelAudit,
+		ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
+			// Replace the custom audit level with "AUDIT" string for better
+			// compatibility with log aggregation systems (Loki, Elasticsearch, etc.)
+			// that expect standard level names. This prevents audit events from
+			// appearing as "INFO+2" which breaks level-based filtering.
+			if a.Key == slog.LevelKey && a.Value.Any() == LevelAudit {
```
amirejaz: The `a.Value.Any() == LevelAudit` comparison works correctly today because `slog` stores level attributes as `slog.Level` values, but the equality check on the bare `any` value is implicit about that type contract. A type assertion makes the expectation explicit and is slightly more defensive:

```go
if a.Key == slog.LevelKey {
    if level, ok := a.Value.Any().(slog.Level); ok && level == LevelAudit {
        a.Value = slog.StringValue("AUDIT")
    }
}
```

This is a minor style point rather than a correctness issue — the current code is fine.

pkg/audit/auditor_test.go
```diff
@@ -1002,3 +1003,53 @@ func TestMiddlewareDetectsJSONRPCErrors(t *testing.T) {
...
+		assert.Contains(t, logOutput, `"level":"AUDIT"`)
+		assert.NotContains(t, logOutput, "INFO+2")
+		assert.NotContains(t, logOutput, "INFO+")
+	})
+
+	t.Run("preserves standard log levels", func(t *testing.T) {
+		t.Parallel()
+		var logBuf bytes.Buffer
+		// Create a logger with DEBUG level to capture all levels
+		handler := slog.NewJSONHandler(&logBuf, &slog.HandlerOptions{
```
amirejaz: The `preserves standard log levels` subtest re-implements the same `ReplaceAttr` closure from `NewAuditLogger` rather than using `NewAuditLogger` itself. If the production implementation had a bug that accidentally modified INFO or WARN labels, this test wouldn't catch it because it's exercising a hand-rolled copy.

It's worth noting that `NewAuditLogger` sets `Level: LevelAudit` (= `slog.Level(2)`), which means `INFO` (`slog.Level(0)`) is filtered out and never reaches `ReplaceAttr`. So you can't test INFO-preservation through `NewAuditLogger` directly. But WARN (`slog.Level(4)`) does pass through and can be tested:

```go
t.Run("preserves standard log levels", func(t *testing.T) {
    t.Parallel()
    var logBuf bytes.Buffer
    logger := NewAuditLogger(&logBuf)

    // WARN (level 4) is above LevelAudit (level 2) and passes through the audit logger.
    logger.Warn("warn message")
    logOutput := logBuf.String()
    assert.Contains(t, logOutput, `"level":"WARN"`)
    assert.NotContains(t, logOutput, `"level":"AUDIT"`)
})
```

This exercises the actual production `ReplaceAttr` path and confirms it doesn't mis-label WARN events.

amirejaz (commented):
Nice fix, and a clear PR description — the problem ("INFO+2" leaking into log aggregators) is well-motivated. The implementation is small and correct. I have two suggestions below: one about making the level comparison more explicit, and one about the second test case re-implementing rather than exercising `NewAuditLogger`.

kimjune01:
Thanks for the review @amirejaz — both points addressed in 9bd90a0.

1. **Explicit level comparison** ([auditor.go:64-68](https://github.com/stacklok/toolhive/blob/9bd90a0b00471cac10862e84241c7148f17d9ad0/pkg/audit/auditor.go#L64-L68)): replaced `a.Value.Any() == LevelAudit` with a `slog.Level` type assertion. Makes the contract explicit and defensive.

2. **Test exercises NewAuditLogger** ([auditor_test.go:1027-1041](https://github.com/stacklok/toolhive/blob/9bd90a0b00471cac10862e84241c7148f17d9ad0/pkg/audit/auditor_test.go#L1027-L1041)): the `preserves standard log levels` subtest now calls `NewAuditLogger` directly and asserts on WARN (above the LevelAudit threshold, so it survives the level filter). A regression in the production `ReplaceAttr` would now be caught. Comment in the test explains why INFO can't be tested through the production path.

`go test ./pkg/audit/...` passes locally.

amirejaz:
Linter is flagging two things in the fixup commit:

1. `auditor.go:59` — `groups` in the `ReplaceAttr` signature is unused. Rename it to `_`.
2. `auditor_test.go:1015` — `logger.Log(nil, ...)` passes a nil context. Replace `nil` with `context.TODO()`.

Otherwise looks good — happy to approve once those are fixed.

kimjune01:
Lint findings addressed in 081101c5.

- `groups` → `_` in `auditor.go:59`
- `nil` → `context.TODO()` in `auditor_test.go:1015`

Tests pass.

ChrisJBurns (approved):

PR #5260, opened 2026-05-12
fix(registry): surface legacy registry format as a structured API error

### LINKED ISSUE #5259: Surface legacy registry format as a structured API error
## Problem

When a user has a custom registry URL configured (`config.RegistryUrl`) that points to an endpoint serving the legacy ToolHive registry format, the backend correctly detects the legacy format and logs a helpful migration hint:

```
custom registry at <url> is not reachable: registry validation failed:
registry at <url>: registry file appears to be in the legacy ToolHive format;
run `thv registry convert --in <path> --in-place` to migrate to the upstream MCP format
```

However, this hint is **not surfaced to API consumers**. At `thv serve` startup, `NewRemoteRegistryProvider.validateConnectivity()` returns the legacy hint as a plain `fmt.Errorf` (see `pkg/registry/provider_remote.go:84-86`), which bubbles up through `factory.go:80` and is then handled by `getCurrentProvider` in `pkg/api/v1/registry.go:259`. Because the error matches neither `UnavailableError` nor `auth.ErrRegistryAuthRequired`, it falls into the generic branch and the API responds with:

```
HTTP 500 Internal Server Error
Failed to get registry provider
```

The migration hint ends up only in `main.log`. Desktop clients (Studio) cannot distinguish this case from a real internal error, so users see the generic error boundary screen instead of a targeted recovery flow.

## Real-world impact

User report (Discord, 12 May 2026): after upgrading Studio to 0.34, the app failed to start because the previously-saved custom registry URL was returning legacy JSON. Recovery required debugging plus running `thv config unset-registry` from the CLI. The actionable hint was logged but invisible from the UI.

## Proposal

Add a typed `LegacyFormatError` (sister of `UnavailableError`) and surface it via a structured **HTTP 502 Bad Gateway** response with `code: "registry_legacy_format"`. 502 is correct per [RFC 9110 §15.6.3](https://datatracker.ietf.org/doc/html/rfc9110#section-15.6.3): thv serve acts as a gateway to the upstream registry, and the upstream returned a response we cannot process. It also aligns with the existing PUT registry handler (`pkg/api/v1/registry.go:545`) which already returns 502 for the same legacy-format detection via `ErrRegistryValidationFailed`.

Concretely:

1. Add a typed `LegacyFormatError` in `pkg/registry/errors.go` (sister of `UnavailableError`). Fields: `URL` (optional source). `Error()` embeds `legacyhint.MigrationMessage`. Implements `Is(target error) bool` so existing `errors.Is(err, errLegacyFormat)` keeps working.

2. Make the package-private `errLegacyFormat` in `pkg/registry/upstream_parser.go` an instance of `LegacyFormatError`, so both the local-file path (`parseRegistryData`) and the remote-URL path return the same typed error.

3. In `pkg/registry/provider_remote.go:85`, replace the plain `fmt.Errorf` with `&LegacyFormatError{URL: p.registryURL}`.

4. In `pkg/api/v1/registry.go`:
   - Add `RegistryLegacyFormatCode = "registry_legacy_format"`.
   - Add `writeRegistryLegacyFormatError` returning **HTTP 502 Bad Gateway** with structured JSON `{ code, message }` matching the existing `registryErrorResponse` shape.
   - Add an `errors.As(err, &legacyErr)` branch in `getCurrentProvider`, `listRegistries`, `getRegistry`, `listServers`, and `registry_v01.go`'s `getRegistryProvider`.

5. Unit tests:
   - `pkg/registry/errors_test.go`: cover `Error()` with/without URL and `errors.Is` matching.
   - `pkg/api/v1/registry_test.go`: end-to-end test where the upstream returns legacy JSON and the API returns 502 + `registry_legacy_format` + migration hint in the message.

## Out of scope

- Re-classifying the existing `registry_unavailable` 503 path (which also covers upstream 404 — arguably 502 territory by RFC). Cleanup in a separate PR.
- API-protocol (`RegistryApiUrl`) registries — they speak the MCP Registry API and cannot serve legacy ToolHive JSON.

## Why 502 (and not 503 or 500)

| Code | RFC 9110 wording | Fit |
|------|------------------|-----|
| 500 | "unexpected condition" | ❌ current state we want to fix |
| **502** | "while acting as a gateway or proxy, received an **invalid response** from an inbound server" | ✅ exact match |
| 503 | "temporary overload or scheduled maintenance" | ❌ not temporary, not overload |
| 422 | media type OK but "couldn't process semantic content" | ✗ pensato per request body del client, non per upstream |

Internal consistency: PUT registry already returns 502 for this exact detection via `ErrRegistryValidationFailed`. Returning 503 from GET would split the same condition across two status codes inside the same handler file.

peppescg:
## Summary

When a configured custom registry URL serves data in the legacy ToolHive format, the API now returns **HTTP 502 Bad Gateway** with a structured `registry_legacy_format` code, instead of a generic 500 that hides the actionable migration hint in the log.

Before, `NewRemoteRegistryProvider.validateConnectivity()` returned a plain `fmt.Errorf` that the API handler did not recognise (`pkg/api/v1/registry.go:259`), so the response was `HTTP 500 Failed to get registry provider` and clients had no way to branch on the legacy-format case. The hint (`run thv registry convert --in <path> --in-place`) only showed up in `main.log`.

This PR adds a typed `*LegacyFormatError` in `pkg/registry/errors.go` (sister of `*UnavailableError`) that carries the offending source URL, and wires `errors.As` detection plus a new `writeRegistryLegacyFormatError` writer through all four `GET` handlers and the `v0.1` router.

The package-private `errLegacyFormat` sentinel in `pkg/registry/upstream_parser.go` is now an instance of `*LegacyFormatError`; its `Is()` method preserves `errors.Is(err, errLegacyFormat)` for existing callers while enabling `errors.As` extraction of the URL upstream.

Closes #5259

## Why HTTP 502 Bad Gateway

Per [RFC 9110 §15.6.3](https://datatracker.ietf.org/doc/html/rfc9110#section-15.6.3): 502 = "while acting as a gateway or proxy, received an **invalid response** from an inbound server." `thv serve` is acting as a gateway to the configured upstream registry, and that upstream is returning data we cannot process. Exact fit.

502 also matches what the **existing PUT registry handler** already does for the same legacy-format detection (`pkg/api/v1/registry.go:545`, via `ErrRegistryValidationFailed`). Returning 503 from `GET` would split the same condition across two status codes inside the same file.

503 ("temporary overload or scheduled maintenance") does not fit a permanently misconfigured registry source.

The `registry_unavailable` path also handles upstream 404s on a 503 today — arguably also 502 territory — but reclassifying that is out of scope for this PR.

## Changes

- `pkg/registry/errors.go` — new `LegacyFormatError` type with `URL`, `Error()`, `Is()`.
- `pkg/registry/provider_remote.go` — `validateConnectivity()` returns `&LegacyFormatError{URL: p.registryURL}` instead of `fmt.Errorf`.
- `pkg/registry/upstream_parser.go` — `errLegacyFormat` switches to `&LegacyFormatError{}` (same `Error()` output, now typed for `errors.As`).
- `pkg/api/v1/registry.go` — new `RegistryLegacyFormatCode` + `writeRegistryLegacyFormatError` returning **502**; new `errors.As(err, &legacyErr)` branch in `getCurrentProvider`, `listRegistries`, `getRegistry`, `listServers`.
- `pkg/api/v1/registry_v01.go` — same branch in `getRegistryProvider`.
- `pkg/registry/errors_test.go` — covers `Error()` with/without URL, `Is` matching across populated/empty URLs and via `Unwrap`, and `As` extraction from a wrapped error.
- `pkg/api/v1/registry_test.go` — end-to-end test: configure a remote-URL registry pointing at an httptest server that serves legacy JSON, hit each `GET` handler, assert **502** + `code: registry_legacy_format` + migration hint in the message + source URL in the message.

## Test plan

- [x] `go test ./pkg/registry/ -run "TestLegacyFormatError" -count=1` — new error type tests pass
- [x] `go test ./pkg/api/v1/ -run "TestRegistryAPI_GetEndpoint_LegacyFormat" -count=1` — all three GET handlers return 502 with the new code
- [x] `go test ./pkg/registry/ -run "TestParseRegistryData_LegacyFormatDetection|TestLocalRegistryProvider_LegacyFileReturnsMigrationHint|TestRemoteRegistryProvider_ValidateConnectivity" -count=1` — existing legacy-detection coverage still green
- [x] `go vet ./pkg/registry/... ./pkg/api/v1/...` clean
- [x] `gofmt -l` clean

Four pre-existing failures on `main` (`TestGetDefaultProvider_NoFactoryRegistered`, `TestGetDefaultProvider_FactoryReturnsNil_FallsThrough`, `TestRegistryV01Router_GetServer_NotFound`, `TestRegistryV01Router_ListSkills_PaginationBeyondResults`) reproduce on `git stash` of this branch — unrelated to this change.

### DIFF
--- a/pkg/api/v1/registry.go
+++ b/pkg/api/v1/registry.go
@@ -57,7 +57,7 @@ func writeRegistryAuthRequiredError(w http.ResponseWriter) {
 const RegistryUnavailableCode = "registry_unavailable"
 
 // RegistryLegacyFormatCode is the machine-readable error code returned in the
-// structured JSON 503 response when the configured registry source serves data
+// structured JSON 502 response when the configured registry source serves data
 // in the legacy ToolHive format instead of the upstream MCP registry format.
 // Desktop clients (Studio) match on this value to display a targeted recovery
 // flow (reset to default registry, point to a `?format=upstream` endpoint, or
@@ -76,17 +76,21 @@ func writeRegistryUnavailableError(w http.ResponseWriter, unavailableErr *regpkg
 	_ = json.NewEncoder(w).Encode(body)
 }
 
-// writeRegistryLegacyFormatError writes a structured JSON 503 response when the
-// configured registry source returns data in the legacy ToolHive format. The
-// message embeds legacyhint.MigrationMessage so CLI users still see the
-// actionable migration step, while desktop clients can branch on Code.
+// writeRegistryLegacyFormatError writes a structured JSON 502 Bad Gateway
+// response when the configured registry source returns data in the legacy
+// ToolHive format. HTTP 502 is correct per RFC 9110 §15.6.3: thv serve is
+// acting as a gateway to the upstream registry, and the upstream returned a
+// response we cannot process. It also aligns with the existing PUT handler
+// which already returns 502 for the same legacy-format detection. The message
+// embeds legacyhint.MigrationMessage so CLI users still see the actionable
+// migration step, while desktop clients can branch on Code.
 func writeRegistryLegacyFormatError(w http.ResponseWriter, legacyErr *regpkg.LegacyFormatError) {
 	body := registryErrorResponse{
 		Code:    RegistryLegacyFormatCode,
 		Message: legacyErr.Error(),
 	}
 	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusServiceUnavailable)
+	w.WriteHeader(http.StatusBadGateway)
 	_ = json.NewEncoder(w).Encode(body)
 }
 
--- a/pkg/api/v1/registry.go
+++ b/pkg/api/v1/registry.go
@@ -56,12 +56,34 @@ func writeRegistryAuthRequiredError(w http.ResponseWriter) {
 // structured JSON 503 response when the upstream registry is unreachable.
 const RegistryUnavailableCode = "registry_unavailable"
 
+// RegistryLegacyFormatCode is the machine-readable error code returned in the
+// structured JSON 503 response when the configured registry source serves data
+// in the legacy ToolHive format instead of the upstream MCP registry format.
+// Desktop clients (Studio) match on this value to display a targeted recovery
+// flow (reset to default registry, point to a `?format=upstream` endpoint, or
+// run `thv registry convert`).
+const RegistryLegacyFormatCode = "registry_legacy_format"
+
 // writeRegistryUnavailableError writes a structured JSON 503 response when the
 // upstream registry cannot be reached or returns an unexpected error (e.g. 404).
 func writeRegistryUnavailableError(w http.ResponseWriter, unavailableErr *regpkg.UnavailableError) {
 	body := registryErrorResponse{
 		Code:    RegistryUnavailableCode,
 		Message: unavailableErr.Error(),
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusServiceUnavailable)
+	_ = json.NewEncoder(w).Encode(body)
+}
+
+// writeRegistryLegacyFormatError writes a structured JSON 503 response when the
+// configured registry source returns data in the legacy ToolHive format. The
+// message embeds legacyhint.MigrationMessage so CLI users still see the
+// actionable migration step, while desktop clients can branch on Code.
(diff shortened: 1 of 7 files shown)


peppescg:
@claude succinct review

rdimitrov (approved):

PR #5267, opened 2026-05-12
References printcolumn shows raw JSON instead of useful summary

### LINKED ISSUE #4618: References printcolumn shows raw JSON instead of useful summary
## Summary

Three CRD types define a printer column for `referencingWorkloads` that renders as raw JSON in `kubectl get` output. The `referencingWorkloads` field is an array of `WorkloadReference` objects, so `type="string"` causes kubectl to display the full JSON serialization, which is not user-friendly and breaks table formatting.

**Severity**: SHOULD FIX
**Area**: CRD PrintColumns
**Breaking**: No

## Location

- `cmd/thv-operator/api/v1alpha1/mcpoidcconfig_types.go:193`
- `cmd/thv-operator/api/v1alpha1/mcpexternalauthconfig_types.go:730`
- `cmd/thv-operator/api/v1alpha1/toolconfig_types.go:109`

## Problem

```go
// +kubebuilder:printcolumn:name="References",type="string",
//   JSONPath=".status.referencingWorkloads"
```

kubectl output looks like:

```
REFERENCES
[{"kind":"MCPServer","name":"my-server","namespace":"default"}]
```

## Impact

- Poor CLI user experience when listing these resources
- Wide columns break table formatting in narrow terminals
- Users cannot quickly see the reference count at a glance

## Recommended Fix

1. Add a `referenceCount` integer field to the status of each affected type:
   ```go
   // ReferenceCount is the number of workloads referencing this config.
   ReferenceCount int32 `json:"referenceCount,omitempty"`
   ```

2. Update controllers to populate `referenceCount` alongside `referencingWorkloads`.

3. Update the printer column to use the count:
   ```go
   // +kubebuilder:printcolumn:name="References",type="integer",
   //   JSONPath=".status.referenceCount"
   ```

Alternatively, remove the References column entirely and let users inspect references via `kubectl get <resource> -o yaml`.

## Related Issues

- MCPGroup Ready column misleadingly shows MCPServersChecked condition (#4617)
- CompositeToolDefinition broken columns — another PrintColumn issue

Sanskarzz:
## Summary

- `kubectl get` output for MCPExternalAuthConfig, MCPOIDCConfig, and MCPToolConfig rendered `References` as raw JSON because the printer column pointed at
`.status.referencingWorkloads`, which is an array of workload reference objects.
- Added a `status.referenceCount` field to those resources and updated the `References` printer column to show the integer count instead.
- Updated controllers to keep `referenceCount` in sync whenever `referencingWorkloads` changes.
- Regenerated CRD manifests and Helm CRD templates.

Fixes #4618 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

## API Compatibility

This adds an optional status field, referenceCount, and keeps the existing referencingWorkloads field unchanged.

## Changes

| File | Change |
|------|--------|
| cmd/thv-operator/api/v1alpha1/types.go | Updated affected v1alpha1 print columns to use .status.referenceCount |
| cmd/thv-operator/api/v1beta1/*config_types.go | Added referenceCount status fields and updated print columns |
| cmd/thv-operator/controllers/* | Populated referenceCount alongside referencingWorkloads |
| cmd/thv-operator/controllers/workload_reference_count.go | Added guarded helper for converting workload reference length to int32 |
| deploy/charts/operator-crds/** | Regenerated CRD manifests and Helm templates |

Yes. kubectl get now shows the References column as a count instead of raw JSON for MCPExternalAuthConfig, MCPOIDCConfig, and MCPToolConfig.

The detailed workload references are still available in .status.referencingWorkloads; this only changes the default table output to use .status.referenceCount.

### DIFF
--- a/cmd/thv-operator/api/v1alpha1/types.go
+++ b/cmd/thv-operator/api/v1alpha1/types.go
@@ -47,7 +47,7 @@ type EmbeddingServerList struct {
 //+kubebuilder:resource:shortName=extauth;mcpextauth,categories=toolhive
 //+kubebuilder:printcolumn:name="Type",type=string,JSONPath=`.spec.type`
 //+kubebuilder:printcolumn:name="Valid",type=string,JSONPath=`.status.conditions[?(@.type=='Valid')].status`
-//+kubebuilder:printcolumn:name="References",type=string,JSONPath=`.status.referencingWorkloads`
+//+kubebuilder:printcolumn:name="References",type=integer,JSONPath=`.status.referenceCount`
 //+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
 // MCPExternalAuthConfig is the deprecated v1alpha1 version of the MCPExternalAuthConfig resource.
@@ -105,7 +105,7 @@ type MCPGroupList struct {
 //+kubebuilder:resource:shortName=mcpoidc,categories=toolhive
 //+kubebuilder:printcolumn:name="Source",type=string,JSONPath=`.spec.type`
 //+kubebuilder:printcolumn:name="Valid",type=string,JSONPath=`.status.conditions[?(@.type=='Valid')].status`
-//+kubebuilder:printcolumn:name="References",type=string,JSONPath=`.status.referencingWorkloads`
+//+kubebuilder:printcolumn:name="References",type=integer,JSONPath=`.status.referenceCount`
 //+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
 // MCPOIDCConfig is the deprecated v1alpha1 version of the MCPOIDCConfig resource.
@@ -311,7 +311,7 @@ type MCPWebhookConfigList struct {
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:shortName=tc;toolconfig,categories=toolhive
 //+kubebuilder:printcolumn:name="Valid",type=string,JSONPath=`.status.conditions[?(@.type=='Valid')].status`
-//+kubebuilder:printcolumn:name="References",type=string,JSONPath=`.status.referencingWorkloads`
+//+kubebuilder:printcolumn:name="References",type=integer,JSONPath=`.status.referenceCount`
 //+kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
 
 // MCPToolConfig is the deprecated v1alpha1 version of the MCPToolConfig resource.
--- a/cmd/thv-operator/controllers/mcpserver_controller.go
+++ b/cmd/thv-operator/controllers/mcpserver_controller.go
@@ -2298,12 +2298,19 @@ func (r *MCPServerReconciler) updateOIDCConfigReferencingWorkloads(
 	// Check if already listed
 	for _, entry := range oidcConfig.Status.ReferencingWorkloads {
 		if entry.Kind == ref.Kind && entry.Name == ref.Name {
+			if oidcConfig.Status.ReferenceCount != workloadReferenceCount(oidcConfig.Status.ReferencingWorkloads) {
+				oidcConfig.Status.ReferenceCount = workloadReferenceCount(oidcConfig.Status.ReferencingWorkloads)
+				if err := r.Status().Update(ctx, oidcConfig); err != nil {
+					return fmt.Errorf("failed to update MCPOIDCConfig ReferenceCount: %w", err)
+				}
+			}
 			return nil
 		}
 	}
 
 	// Add the workload reference
 	oidcConfig.Status.ReferencingWorkloads = append(oidcConfig.Status.ReferencingWorkloads, ref)
+	oidcConfig.Status.ReferenceCount = workloadReferenceCount(oidcConfig.Status.ReferencingWorkloads)
 	if err := r.Status().Update(ctx, oidcConfig); err != nil {
 		return fmt.Errorf("failed to update MCPOIDCConfig ReferencingWorkloads: %w", err)
 	}
--- a/cmd/thv-operator/controllers/mcpserver_controller.go
+++ b/cmd/thv-operator/controllers/mcpserver_controller.go
@@ -2298,12 +2298,6 @@ func (r *MCPServerReconciler) updateOIDCConfigReferencingWorkloads(
 	// Check if already listed
 	for _, entry := range oidcConfig.Status.ReferencingWorkloads {
 		if entry.Kind == ref.Kind && entry.Name == ref.Name {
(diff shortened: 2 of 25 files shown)


cmd/thv-operator/api/v1alpha1/types.go
```diff
@@ -47,7 +47,7 @@ type EmbeddingServerList struct {
 //+kubebuilder:resource:shortName=extauth;mcpextauth,categories=toolhive
 //+kubebuilder:printcolumn:name="Type",type=string,JSONPath=`.spec.type`
 //+kubebuilder:printcolumn:name="Valid",type=string,JSONPath=`.status.conditions[?(@.type=='Valid')].status`
-//+kubebuilder:printcolumn:name="References",type=string,JSONPath=`.status.referencingWorkloads`
+//+kubebuilder:printcolumn:name="References",type=integer,JSONPath=`.status.referenceCount`
```
ChrisJBurns, line 50: Thanks for the PR @Sanskarzz but I think we should be adding it alongside the `referencingWorkloads` on the status object, not just replacing it.

Sanskarzz: Thanks @ChrisJBurns. I agree, and the current implementation keeps `referencingWorkloads` on the status object.

For the affected status structs, `referenceCount` is added alongside the existing `referencingWorkloads` field. The only replacement is in the printer column JSONPath, so `kubectl get` displays the count instead of rendering the full workload reference array as raw JSON.

The detailed references remain available via `.status.referencingWorkloads`.

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -2298,12 +2298,19 @@ func (r *MCPServerReconciler) updateOIDCConfigReferencingWorkloads(
 	// Check if already listed
 	for _, entry := range oidcConfig.Status.ReferencingWorkloads {
 		if entry.Kind == ref.Kind && entry.Name == ref.Name {
+			if oidcConfig.Status.ReferenceCount != workloadReferenceCount(oidcConfig.Status.ReferencingWorkloads) {
```
amirejaz: The stale-count repair path here (and its duplicates in mcpremoteproxy_controller.go and virtualmcpserver_controller.go) is a bit surprising to find inside a function whose stated purpose is "ensure this workload is listed".

This branch fires when the workload is **already** in ReferencingWorkloads but ReferenceCount does not match the slice length. That divergence can only arise if the count was written incorrectly (or not written) by a previous reconcile -- exactly the kind of bug this PR is fixing by adding ReferenceCount everywhere. Going forward, every code path that writes ReferencingWorkloads also writes ReferenceCount, so the divergence should not recur.

Consider whether this repair block is still needed. If the OIDC/ExternalAuth/ToolConfig reconcile loops (which already call Status().Update on every relevant change) reliably keep the count in sync, this "already listed" branch only needs to return nil -- no extra status write required. Removing the repair would also eliminate an extra Status().Update call on reconcile hot-paths where the workload is already tracked.

If you do want to keep it as a safety net, a short comment explaining when the divergence could occur would help future readers.

cmd/thv-operator/controllers/workload_reference_count.go
```diff
@@ -0,0 +1,22 @@
...
+	"math"
+
+	mcpv1beta1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1beta1"
+)
+
+func workloadReferenceCount(refs []mcpv1beta1.WorkloadReference) int32 {
+	return workloadReferenceCountFromLen(len(refs))
+}
+
+func workloadReferenceCountFromLen(length int) int32 {
```
amirejaz, line 16: The two-function split (workloadReferenceCount -> workloadReferenceCountFromLen) is a reasonable way to keep the overflow guard testable in isolation, and the nolint:gosec comment correctly explains the suppression.

One small nit: the overflow guard (length > math.MaxInt32) is technically unreachable in practice -- the Kubernetes API server enforces etcd object-count limits that are orders of magnitude below math.MaxInt32, and a []WorkloadReference slice that large would exhaust memory long before reaching this code. If you prefer, you can simplify to a direct int32(len(refs)) and drop the guard entirely (the gosec lint warning can be addressed with the same nolint comment). That said, keeping the guard as a belt-and-suspenders measure is also fine -- just noting it so the decision is intentional.

amirejaz (commented):
Thanks for the contribution! The overall approach is solid — storing a derived referenceCount integer on the status and pointing the printer column at it is exactly the right fix for the raw-JSON display problem. The code is well-structured, tests are thorough, and the CRD manifests and docs are all regenerated consistently.

I have a couple of observations worth discussing, mostly around the stale-count repair path that was added to the three updateOIDCConfigReferencingWorkloads functions.

cmd/thv-operator/controllers/workload_reference_count_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	mcpv1beta1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1beta1"
+)
+
+func TestWorkloadReferenceCount(t *testing.T) {
```
amirejaz, line 15: The existing tests cover the happy path (two items) and the overflow boundary. It might be worth adding a case for an empty/nil slice to make it explicit that workloadReferenceCount(nil) returns 0 -- this is guaranteed by Go's len(nil) == 0 rule, but a quick assertion makes the contract self-documenting and guards against any future refactor.

For example:
assert.EqualValues(t, 0, workloadReferenceCount(nil))
assert.EqualValues(t, 0, workloadReferenceCount([]mcpv1beta1.WorkloadReference{}))

cmd/thv-operator/controllers/mcpserver_controller.go
Sanskarzz: Thanks, agreed. I removed the stale-count repair branch from all three `updateOIDCConfigReferencingWorkloads` helpers. The already-listed path now just returns without issuing another status update, and `ReferenceCount` is still updated whenever a new reference is added.

cmd/thv-operator/controllers/workload_reference_count.go
Sanskarzz: I kept the overflow guard intentionally. It is not expected to be reachable in practice, but it keeps the `int -> int32` conversion defensively bounded and makes the `gosec` suppression explicit.

cmd/thv-operator/controllers/workload_reference_count_test.go
Sanskarzz: Added explicit nil and empty slice assertions to `TestWorkloadReferenceCount` so the zero-reference behavior is documented and covered.

ChrisJBurns (approved):

PR #5269, opened 2026-05-13
Add identityFromToken to MCPExternalAuthConfig CRD

### LINKED ISSUE #5155: Add identityFromToken to MCPExternalAuthConfig CRD

jhrozek:
## Summary

Some OAuth2 upstream providers don't expose a usable userinfo endpoint but include user identity in the token response itself — Slack returns `authed_user.id`, Snowflake embeds `username` inside the access-token JWT. Without an explicit identity-extraction surface, the embedded auth server falls back to synthesis mode and produces non-PII subjects with no Name/Email, which means real operators can't reliably distinguish users across these providers.

This PR adds an `identityFromToken` block on `OAuth2UpstreamConfig` that maps `subject`, `name`, and `email` to gjson dot-notation paths into the token-endpoint response body. The new type lives alongside `TokenResponseMapping` and `UserInfoConfig`, with cross-referenced doc comments so operators can pick the right one.

- New `IdentityFromTokenConfig` type with `subjectPath` (required), `namePath`, `emailPath` (optional). All three: `MinLength=1`, `MaxLength=256`. Subject is required by both kubebuilder markers and a reconcile-time helper for defense-in-depth.
- Trust-model warning lives both on the type doc and inline on `SubjectPath` so it surfaces in `kubectl explain` and the generated API reference: claims are trusted via TLS only, not cryptographically verified — prefer OIDC ID tokens when verifiable claims matter. The same caveat applies to `@upstreamjwt`, which performs no signature verification.
- Coexistence with `UserInfo` and `TokenResponseMapping` is allowed at admission; runtime priority is decided by the embedded auth server in a later commit.
- `SyntheticIdentityUpstreams()` — the predicate driving `ConditionTypeIdentitySynthesized` — now also skips upstreams configured with `IdentityFromToken`. This is a partial step toward #5159; the full work (runtime WARN, controller condition messages, reason-constant rename, parity tests) lands in a follow-up.
- `VirtualMCPServer` CRD regenerates because both resources embed `EmbeddedAuthServerConfig`.

Runtime parsing of `IdentityFromToken` from the token response lands separately.

Closes #5155

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Manual checks performed during review iteration:

- Verified the regenerated CRD YAML (Helm `files/crds/` + `templates/` copies) carries `maxLength: 256` / `minLength: 1` for all three path fields and the trust-model warning text on `subjectPath`.
- Verified `docs/operator/crd-api.md` renders the warning inline in the `subjectPath` row and shows the validation column.
- Ran the focused test for `SyntheticIdentityUpstreams` — two new cases lock in that `IdentityFromToken`-only upstreams are excluded.
- Ran the full operator unit suite (`./cmd/thv-operator/...`) to confirm controller tests still pass with the predicate change.

## API Compatibility

Purely additive: a new optional `identityFromToken` block on `OAuth2UpstreamConfig`. Existing manifests are unaffected.

Yes. Operators configuring an `MCPExternalAuthConfig` or `VirtualMCPServer` can now add an `identityFromToken` block to an OAuth2 upstream:

```yaml
oauth2Config:
  identityFromToken:
    subjectPath: "authed_user.id"      # Slack
    # subjectPath: "access_token|@upstreamjwt|sub"  # Snowflake (JWT-embedded)
    namePath: "authed_user.name"
    emailPath: "authed_user.email"
```

Until the runtime wiring lands, the field is admitted but ignored by the embedded auth server.

The synthesis-predicate change (`SyntheticIdentityUpstreams` skipping `IdentityFromToken` upstreams) overlaps with #5159's scope. #5155 explicitly said synthesis-mode awareness was out of scope, but leaving the predicate stale here would have produced misleading status conditions for any cluster that adopts the new field before the full #5159 work lands. The follow-up still needs to rename the reason constants, update the runtime WARN in `NewOAuth2Provider`, update the controller condition messages, and add parity tests.


amirejaz (approved):
Non-blocking, but worth a follow-up: the constant value `\"AllUpstreamsHaveUserInfo\"` is now technically stale, it surfaces in `kubectl describe` via both controllers, so an operator running purely with `identityFromToken` will see a reason string that doesn't match their config. The comment above was updated correctly, just the value wasn't. Changing it would be a v1beta1 break, so fine to defer, but worth tracking

Issue #5208, opened 2026-05-06
vMCP: Load authz config from ConfigMap (policies + primaryUpstreamProvider)


tgrunnagle:
## Background

`pkg/vmcpconfig/converter.go` has two TODOs covering the configMap-sourced authz path on `VirtualMCPServer`:

- `converter.go:192` — `TODO: Load policies from ConfigMap if Type is "configMap"`
- `converter.go:210` — `TODO: load primaryUpstreamProvider from configMap`

Today, both TODOs mean a `VirtualMCPServer` with `spec.incomingAuth.authzConfig.type: configMap` silently produces a vMCP runtime that has no Cedar middleware installed (the factory at `pkg/vmcp/auth/factory/incoming.go:88` requires `len(cfg.Authz.Policies) > 0`). And there is no way for configMap users to override the auto-selected primary upstream provider, because that field lives on `InlineAuthzConfig` only.

## Tasks

- [ ] Extract a shared loader (e.g. in `cmd/thv-operator/pkg/controllerutil/`) that reads + parses the policy ConfigMap. The MCPServer/runner path (`AddAuthzConfigOptions`) and the vMCP converter should both call it. The loader should return policies + any primary-upstream-provider configuration.
- [ ] Wire the converter: when `Type == \"configMap\"`, call the loader, populate `incoming.Authz.Policies` (and `PrimaryUpstreamProvider` if applicable). Fail closed on missing/malformed/empty configMap.
- [ ] Decide where `primaryUpstreamProvider` lives for configMap users:
  - **Option A** — parse it from the policy ConfigMap body (mirrors inline shape, but validator must read the configMap to validate).
  - **Option B** — lift `PrimaryUpstreamProvider` from `InlineAuthzConfig` to `AuthzConfigRef` so it lives on the spec and is source-agnostic. Validator stays simple. Recommended.
- [ ] Status condition parity: add `ConditionReasonAuthzConfigMapNotFound` to `virtualmcpserver_types.go` (already exists on `MCPRemoteProxy` at `mcpremoteproxy_types.go:336`).
- [ ] Update the validator: drop the inline-only assumption in `AuthzConfigRef.ExplicitPrimaryUpstreamProvider()` once the schema decision lands.
- [ ] Tests: converter cases (success, missing CM, missing key, empty value, malformed YAML); validator case for configMap with explicit provider.

## Discovered via

- PR #5199 review (consensus comment, finding F5).
[...]

Issue #4919, opened 2026-04-17
VirtualMCPServer doesn't resolve authzConfig.configMap — vMCP rejects 'type: configMap' at startup


danbarr:
## Summary

A `VirtualMCPServer` with `spec.incomingAuth.authzConfig.type: configMap` produces a rendered vMCP config of:

```yaml
incomingAuth:
  type: oidc
  oidc: { ... }
  authz:
    type: configMap   # ← passed through raw from the CRD
```

The vmcp binary then fails validation at startup:

```
Configuration validation failed: invalid configuration:
  - incomingAuth.authz: type must be one of: cedar, none
```

The binary's `AuthzConfig` struct only accepts `type: cedar` (with inline `policies`) or `type: none` — see the [CRD API doc's `vmcp.config.AuthzConfig`](https://docs.stacklok.com/toolhive/reference/operator-crd-api#vmcpconfigauthzconfig). The operator's converter is expected to **read the referenced ConfigMap and inline its `cedar.policies` + `cedar.entities_json` into the rendered config**, as the same doc's `vmcp.config.IncomingAuthConfig` explicitly notes:

> the operator's converter will resolve the CRD's IncomingAuth (which supports Kubernetes-native references like SecretKeyRef, ConfigMapRef, etc.) and populate this IncomingAuthConfig with the resolved values.

Reproduced on operator `v0.21.0`.

## Expected

When `authzConfig.type: configMap`, the operator reads the ConfigMap at `<name>/<key>`, parses its `{version, type: cedarv1, cedar: {policies, entities_json}}` payload, and emits:

```yaml
authz:
  type: cedar
  policies: [ ... ]         # from cedar.policies
  # entities_json similarly flattened if supported
```

(Equivalent to what's produced when a user chooses `authzConfig.type: inline` directly.)

## Actual

The operator passes `type: configMap` through without resolving the ConfigMap. No volume is mounted for the authz ConfigMap (unlike `caBundleRef`, which at least mounts its ConfigMap — see #4918 for the related gap). The vMCP binary then crashes with the validation error above, deployment goes into CrashLoopBackOff.

## Minimal reproduction

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: example-authz
  namespace: toolhive-system
data:
  authz-config.json: |
    {
      "version": "1.0",
      "type": "cedarv1",
      "cedar": {
        "policies": [
[...]

PR #5271, opened 2026-05-13
Watch authz ConfigMaps from VirtualMCPServer

### LINKED ISSUE #5270: VirtualMCPServer controller should re-reconcile when a referenced authz ConfigMap changes
### Background

When #5208 lands, a `VirtualMCPServer` with `spec.incomingAuth.authzConfig.type: configMap` will have its policies resolved by the operator at reconcile time: the converter reads the referenced ConfigMap, extracts the Cedar policies, and bakes them into the vmcp `config.yaml` ConfigMap. A content-checksum annotation on the pod template then triggers a rollout when config.yaml changes.

This architecture differs from `MCPServer`, where the authz ConfigMap is mounted as a live volume — kubelet keeps it current and no pod restart is needed when policies change.

### Problem

The `VirtualMCPServer` controller has no watch on ConfigMaps referenced by `spec.incomingAuth.authzConfig.configMap`. The controller is purely event-driven with no polling fallback. So when the referenced ConfigMap's content changes after the initial reconcile:

1. The controller is never notified
2. The converter is never re-run
3. The vmcp config.yaml retains the old policies
4. The running vmcp pod enforces stale Cedar policies **indefinitely**

This matters especially in the enterprise use case, where an external controller rewrites the authz ConfigMap whenever Cedar policies change.

### Expected behaviour

When a ConfigMap referenced by `spec.incomingAuth.authzConfig.configMap.name` has its `.data` changed, the owning `VirtualMCPServer` should be enqueued for reconciliation. The reconcile re-runs the converter, regenerates config.yaml, and — if the content changed — the checksum annotation triggers an automatic pod rollout.

A metadata-only update to the ConfigMap (labels, annotations) should not trigger a reconcile.

### Implementation sketch

In `SetupWithManager`, add a `Watches` entry for `corev1.ConfigMap` with:

- A mapper that lists all `VirtualMCPServer` objects and enqueues any that reference the changed ConfigMap via `spec.incomingAuth.authzConfig.configMap.name`
- A predicate that admits only `Update` events where `.data` actually changed

This follows the existing pattern used for `MCPOIDCConfig` and `MCPTelemetryConfig` references (`mapOIDCConfigToVirtualMCPServer`, `mapTelemetryConfigToVirtualMCPServer`).

### Dependency

Requires #5208 — without ConfigMap resolution in the converter, a re-reconcile produces a config.yaml with no policies regardless.

### Acceptance criteria

- Updating an authz ConfigMap referenced by a `VirtualMCPServer` triggers an immediate reconcile (no polling delay)
- The vmcp pod is rolled out with updated policies after the reconcile completes
- A metadata-only ConfigMap update (label/annotation change) does not trigger a reconcile

blkt:
The VirtualMCPServer controller did not watch ConfigMaps referenced by spec.incomingAuth.authzConfig.configMap, so changes to a referenced authz ConfigMap were never observed: the converter was not re-run, the rendered vmcp config.yaml retained stale policies, and the running pod enforced them indefinitely. This matters in particular for the enterprise flow where an external controller rewrites the authz ConfigMap when Cedar policies change.

Add a Watches entry on corev1.ConfigMap with:

  * a mapper that lists VirtualMCPServers in the ConfigMap's namespace and enqueues those whose AuthzConfig.Type is "configMap" and whose ConfigMap.Name matches the changed object
  * a predicate that admits Create and Delete events unconditionally and Update events only when .Data or .BinaryData actually changed, so metadata-only updates (labels, annotations, resourceVersion bumps) do not trigger reconciliation

The watch is functionally meaningful only once #5208 lands and the converter actually resolves ConfigMap-sourced policies; until then the re-reconcile still produces an empty-policy config. Wiring it now unblocks that follow-up without an extra round trip.

Closes #5270

### DIFF
--- a/cmd/thv-operator/controllers/virtualmcpserver_authz_configmap.go
+++ b/cmd/thv-operator/controllers/virtualmcpserver_authz_configmap.go
@@ -1,0 +1,94 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package controllers
+
+import (
+	"context"
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	mcpv1beta1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1beta1"
+)
+
+// mapAuthzConfigMapToVirtualMCPServer maps ConfigMap changes to VirtualMCPServer reconciliation
+// requests. Used by SetupWithManager to trigger reconciliation when a ConfigMap referenced via
+// spec.incomingAuth.authzConfig.configMap is updated, so the converter can re-resolve policies
+// and roll out a pod with the new config.
+//
+// The mapper lists VirtualMCPServers in the ConfigMap's namespace and enqueues any that
+// reference this ConfigMap. ConfigMaps are cluster-wide objects but authz references are
+// namespace-scoped, so the lookup is bounded to a single namespace.
+func (r *VirtualMCPServerReconciler) mapAuthzConfigMapToVirtualMCPServer(
+	ctx context.Context, obj client.Object,
+) []reconcile.Request {
+	cm, ok := obj.(*corev1.ConfigMap)
+	if !ok {
+		return nil
+	}
+
+	vmcpList := &mcpv1beta1.VirtualMCPServerList{}
+	if err := r.List(ctx, vmcpList, client.InNamespace(cm.Namespace)); err != nil {
+		log.FromContext(ctx).Error(err, "Failed to list VirtualMCPServers for authz ConfigMap watch")
+		return nil
+	}
+
+	var requests []reconcile.Request
+	for _, vmcp := range vmcpList.Items {
+		if !vmcpReferencesAuthzConfigMap(&vmcp, cm.Name) {
+			continue
+		}
+		requests = append(requests, reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Name:      vmcp.Name,
+				Namespace: vmcp.Namespace,
+			},
+		})
+	}
+
+	return requests
+}
+
+// vmcpReferencesAuthzConfigMap reports whether the VirtualMCPServer references the named
+// ConfigMap via spec.incomingAuth.authzConfig.
+func vmcpReferencesAuthzConfigMap(vmcp *mcpv1beta1.VirtualMCPServer, configMapName string) bool {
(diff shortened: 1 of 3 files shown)


cmd/thv-operator/controllers/virtualmcpserver_authz_configmap.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+			newCM, ok := e.ObjectNew.(*corev1.ConfigMap)
+			if !ok {
+				return false
+			}
+			return !reflect.DeepEqual(oldCM.Data, newCM.Data) ||
+				!reflect.DeepEqual(oldCM.BinaryData, newCM.BinaryData)
+		},
+		CreateFunc:  func(_ event.CreateEvent) bool { return true },
+		DeleteFunc:  func(_ event.DeleteEvent) bool { return true },
```
jhrozek, line 91: When a referenced ConfigMap is deleted, the predicate correctly admits the event and reconciliation is triggered. However, because the authz ConfigMap is mounted as a pod volume rather than `Get`-ted during reconciliation, the controller will complete successfully — but the pod will be stuck `Pending` because Kubernetes can't mount the missing volume.

From a user's perspective, the VirtualMCPServer shows no diagnostic status condition explaining the problem; they'd have to inspect pod events to find the cause.

Worth a follow-up to add an explicit `Get` on the referenced ConfigMap in the reconcile path and surface an `AuthzConfigReady=False` condition when the ConfigMap is `NotFound`. That would let the watch event on ConfigMap creation automatically recover the VirtualMCPServer without any other intervention.

jhrozek (approved):
that comment was meant to be a follow-up suggestion, not a blocker

cmd/thv-operator/controllers/virtualmcpserver_authz_configmap.go
blkt: Thanks for catching this. One clarification on the premise: for `VirtualMCPServer` the authz ConfigMap isn't mounted as a pod volume. `buildVolumesForVmcp` at `cmd/thv-operator/controllers/virtualmcpserver_deployment.go:285-330` only mounts the operator-generated `vmcp-config` (the rendered `config.yaml`). `GenerateAuthzVolumeConfig` is called only from `mcpserver_controller.go:1187` and `mcpremoteproxy_deployment.go:140` — so the Pending-pod failure mode you describe applies to `MCPServer`/`MCPRemoteProxy` but not vMCP. (#4919 calls this out as "No volume is mounted for the authz ConfigMap".)

The intended vMCP flow is "operator reads the CM at reconcile time and bakes policies into `config.yaml`" — so a missing CM never reaches the pod. Today it silently goes nowhere because the converter doesn't `Get` it yet; that's the TODO at `cmd/thv-operator/pkg/vmcpconfig/converter.go:192` tracked by #5208.

Your underlying diagnostic concern is still right, though, and it's already on #5208's task list:
- "Wire the converter: when `Type == \"configMap\"`, call the loader … Fail closed on missing/malformed/empty configMap."
- "Status condition parity: add `ConditionReasonAuthzConfigMapNotFound` to `virtualmcpserver_types.go` (already exists on `MCPRemoteProxy`)."

So once #5208 lands, a deleted/missing authz ConfigMap will produce a `ConfigurationInvalid` condition with `AuthzConfigMapNotFound`, and the watch wired here will recover the VMS automatically when the CM is recreated — which is exactly the behavior you're asking for. No separate follow-up needed.

PR #3093, opened 2025-12-18
add feature flags to enable/disable controller groups

amirejaz:
## Summary

Adds operator-side feature flags to complement the CRD installation flags from #2809, allowing users to conditionally enable/disable controller groups at runtime.

## Changes

- Add `ENABLE_SERVER`, `ENABLE_REGISTRY`, `ENABLE_VMCP` environment variables in `main.go`
- Refactor `setupControllersAndWebhooks` to conditionally start controllers based on flags
- Add dependency checking (VMCP requires Server to be enabled)
- Wire feature flags through Helm values (`operator.features.server/registry/virtualMCP`)
- Update deployment template to pass environment variables

## Usage

Skip Virtual MCP controllers (matches `crds.install.virtualMcp=false`):

helm install operator ... --set operator.features.virtualMCP=false

## Related

Complements #2809 (CRD installation feature flags)

PR #5281, opened 2026-05-14
Drop per-component CRD and controller gating from operator install

ChrisJBurns:
## Summary

Two related changes that move toolhive away from per-component install/runtime gating for operator CRDs and controllers:

1. **Adopt the external [`stacklok/helm-crd-wrapper`](https://github.com/stacklok/helm-crd-wrapper) binary** in place of the in-repo `crd-helm-wrapper/` Go program. Sharing one tool across charts (toolhive, stacklok-llm-gateway, …) keeps wrap behavior consistent and removes a copy of the code from this repo. The external tool intentionally has no per-CRD configuration — its README says "either every CRD survives `helm uninstall` or none do; mixed install groups inside one chart … usually means there should be two charts." We agree.
2. **Remove the operator-side feature-flag gating** added in #3093 (`ENABLE_SERVER` / `ENABLE_REGISTRY` / `ENABLE_VMCP`). We never use the ability to run with a partial controller set, and the runtime gating only adds a way to land in a broken half-installed state (e.g. `virtualMCP=true` with `server=false`, guarded by an ad-hoc dependency map). All controllers now register unconditionally.

The two changes pair naturally: the chart no longer offers granular install knobs and the operator no longer offers granular runtime knobs, so the install/start path is uniform.

The external `helm-crd-wrapper` is pinned to commit `a736420b` because it has no tagged release yet — swap for a tag when one is cut.

## Type of change

(Behavior change is limited to the chart values schema — see "user-facing change" below. Default behavior is unchanged for both pieces.)

## Test plan

- [x] Unit tests (`task operator-test`)
- [x] Manual testing (described below)

Manual verification:

1. `task operator-manifests` (from `cmd/thv-operator/`) regenerates all 13 wrapped CRD templates via the external binary.
2. `helm template test deploy/charts/operator-crds` renders 13 CRDs with defaults; `--set crds.install=false` renders 0.
3. `helm template test deploy/charts/operator` renders the operator Deployment with only `ENABLE_EXPERIMENTAL_FEATURES` remaining among the `ENABLE_*` env vars.
4. `helm-docs --chart-search-root=deploy/charts` regenerates both chart READMEs cleanly.
5. `ct lint --charts deploy/charts/operator-crds` passes chart validation (only `yamllint`-not-installed warning, unrelated).
6. `go build ./cmd/thv-operator/...` succeeds; `task operator-test` passes.

### Chart-level wrapper swap

| File | Change |
| --- | --- |
| `cmd/thv-operator/Taskfile.yml` | `operator-manifests` now `go install`s the external binary and invokes it with `-source`/`-target` only. |
| `deploy/charts/operator-crds/values.yaml` | Collapse `crds.install.{server,registry,virtualMcp}` into a single `crds.install: true` boolean. |
| `deploy/charts/operator-crds/templates/*.yaml` (13 files) | Regenerated — first-line conditional becomes `{{- if .Values.crds.install }}`. |
| `deploy/charts/operator-crds/crd-helm-wrapper/` | Deleted (replaced by the external binary). |
| `deploy/charts/operator-crds/.helmignore`, `.gitignore` | Drop entries that referenced the deleted directory/binary. |
| `deploy/charts/operator-crds/README.md` | Regenerated by `helm-docs`. |

### Operator controller gating removal

| File | Change |
| --- | --- |
| `cmd/thv-operator/main.go` | Remove feature-flag constants, `controllerDependencies` map, dispatcher branching, and `isFeatureEnabled`. `setupControllersAndWebhooks` now calls each setup helper unconditionally. |
| `cmd/thv-operator/main_test.go` | Deleted — only tested `isFeatureEnabled` and the dependency map. |
| `deploy/charts/operator/values.yaml` | Drop `operator.features.{server,registry,virtualMCP}`. |
| `deploy/charts/operator/templates/deployment.yaml` | Drop `ENABLE_SERVER` / `ENABLE_REGISTRY` / `ENABLE_VMCP` env vars. |
| `deploy/charts/operator/README.md` | Regenerated by `helm-docs`. |

## ⚠️ Breaking change

This is a breaking change for any consumer who set one of the removed flags to `false`. Default behavior is unchanged — anyone running on defaults (or who only set them to `true`) is unaffected — but explicit opt-outs are silently dropped:

- **`crds.install.{server,registry,virtualMcp}=false`** (operator-crds chart): override is ignored; all CRDs install. Most likely affected: consumers who set `virtualMcp=false` to skip vMCP CRDs.
- **`operator.features.{server,registry,virtualMCP}=false`** (operator chart): override is ignored; all controllers register. Pairs with the above — vMCP opt-outs are the most likely case.
- **`ENABLE_SERVER` / `ENABLE_REGISTRY` / `ENABLE_VMCP`** env vars on the operator binary: silently ignored. Operators run outside the chart with these set will start every controller anyway.

No CRD schema or v1beta1 API break. The `api-compat.yml` check is unaffected.

Migration: remove the old keys from your values files (left in place they are silently ignored). If a consumer was depending on skipping vMCP CRDs/controllers for operational reasons, raise it before merge so we can decide whether to put gating back behind a single coarser knob.

Yes — chart consumers who set any of the removed values must update their values files:

```yaml
# operator-crds chart — before
crds:
  install:
    server: true
    registry: true
    virtualMcp: true

# operator-crds chart — after
crds:
  install: true
```

```yaml
# operator chart — before
operator:
  features:
    server: true
    registry: true
    virtualMCP: true

# operator chart — after (these keys removed entirely)
```

The previous keys are silently ignored after this change. Default behavior (all CRDs installed, all controllers started) is unchanged.

- `stacklok/helm-crd-wrapper` has no tagged release yet, so the Taskfile pins `@a736420b58d68304fac3934cce57c67d0a989fe7`. Replace with a semver tag once one is cut.
- The `api-compat.yml` workflow's comments still reference `crd-helm-wrapper` generically; they remain accurate (the source-of-truth path `files/crds/` is unchanged) so I left them alone.


amirejaz (approved):
Two comments in `.github/workflows/api-compat.yml` (lines 24 and 27) still reference `crd-helm-wrapper` — should be `helm-crd-wrapper` now that the Taskfile switched to the external package.

Issue #4217, opened 2026-03-18
Extend MCPServer reconciler and deployment builder with spec-driven replicas, stdio cap, session storage validation, and terminationGracePeriodSeconds


yrobla:
## Description

Extend `deploymentForMCPServer` in `mcpserver_controller.go` to set `Deployment.Spec.Replicas` from `spec.replicas` with nil-passthrough (preserving HPA compatibility) and set `terminationGracePeriodSeconds` on the pod spec. Extend the reconciler to enforce the stdio transport hard cap (force to 1 and set a status condition when `spec.transport == stdio && spec.replicas > 1`) and emit a `Warning` status condition when `spec.replicas > 1` but session storage is absent or set to `memory`. These changes implement RC-3 and RC-14 of epic THV-0047 for the MCPServer resource.

## Context

Epic THV-0047 adds horizontal scaling support to the ToolHive operator. TASK-001 (#267) and TASK-002 (#269) added the `Replicas`, `BackendReplicas`, and `SessionStorage` fields to the CRD specs; TASK-003 (#275) regenerated the deepcopy functions and CRD YAML manifests, making the new fields available at compile time. This task wires those fields into the MCPServer reconciler and deployment builder so they are actually respected at runtime.

The existing `deploymentForMCPServer` in `mcpserver_controller.go` hardcodes `replicas := int32(1)` and never sets `terminationGracePeriodSeconds`. The existing reconciler at lines 392–407 handles the stdio cap for externally-driven scaling (HPA or `kubectl scale`), but has no path for spec-driven replica values. This task adds both the spec-driven replica path and the session storage validation warning.

**Dependencies**: #275 (TASK-003 — Regenerate deepcopy and CRD manifests after scaling type changes)
**Blocks**: TASK-008 (Unit Tests)

## Acceptance Criteria

- [ ] `deploymentForMCPServer` sets `Deployment.Spec.Replicas` to `spec.replicas` when `spec.replicas` is non-nil; when `spec.replicas` is nil, `Deployment.Spec.Replicas` is left nil (not set to any default), allowing external scaling tools (HPA, KEDA, `kubectl scale`) to manage replica counts freely
- [ ] `deploymentForMCPServer` sets `TerminationGracePeriodSeconds` on `PodSpec` using the value from a named constant or a documented default (e.g., 30 seconds); the field must appear in the generated `Deployment` even when `spec.replicas` is nil
[...]

Issue #4215, opened 2026-03-18
Extend deploymentForVirtualMCPServer with spec-driven replicas, terminationGracePeriodSeconds, and Redis password env var


yrobla:
## Description

Extend `deploymentForVirtualMCPServer` in `virtualmcpserver_deployment.go` to support horizontal scaling for VirtualMCPServer pods. This task sets `Deployment.Spec.Replicas` from `spec.replicas` with nil-passthrough (for HPA compatibility), adds `terminationGracePeriodSeconds` to the pod spec for graceful shutdown, and injects the Redis password as an environment variable (`THV_SESSION_REDIS_PASSWORD`) from `SessionStorageConfig.PasswordRef` using the established `SecretKeyRef` env var pattern.

## Context

Epic THV-0047 introduces horizontal scaling support for the ToolHive operator. This task covers the VirtualMCPServer deployment builder changes needed for that support. Once TASK-003 (#275 — regenerate deepcopy and CRD manifests) has made the new `Replicas` and `SessionStorage` fields available in the compiled API types, this task consumes them in the deployment builder to make the vMCP Deployment reflect the desired replica count and inject session storage credentials into pods.

The nil-passthrough pattern for replicas is critical: when `spec.replicas` is nil, `Deployment.Spec.Replicas` must be omitted (not set to 1), preserving compatibility with HPAs and manual scaling. Redis password injection follows the existing `buildOIDCEnvVars` and `buildHMACSecretEnvVar` patterns already in the same file, which source secrets via `ValueFrom.SecretKeyRef`.

**Dependencies**: #275 (TASK-003 — Regenerate deepcopy and CRD manifests)
**Blocks**: TASK-008 (Unit Tests)

## Acceptance Criteria

- [ ] When `spec.replicas` is set to a non-nil value (e.g., 3), `deploymentForVirtualMCPServer` sets `Deployment.Spec.Replicas` to that value
- [ ] When `spec.replicas` is nil, `Deployment.Spec.Replicas` is nil (not set to `int32(1)` or any other default), allowing HPAs and external scaling tools to manage replica counts
- [ ] `terminationGracePeriodSeconds` is set on the pod spec (`Deployment.Spec.Template.Spec.TerminationGracePeriodSeconds`) in all cases, regardless of replica count
[...]

PR #5286, opened 2026-05-14
fix(operator): inject THV_SESSION_REDIS_PASSWORD for MCPServer

dallinstevens:
## Problem

\`MCPServer.spec.sessionStorage.passwordRef\` is accepted by the CRD schema, and the runner reads \`THV_SESSION_REDIS_PASSWORD\` (see \`pkg/transport/session/redis_config.go\`), but the operator never wires the two together. \`populateScalingConfig\` in \`cmd/thv-operator/controllers/mcpserver_runconfig.go\` copies only \`Address\`/\`DB\`/\`KeyPrefix\` into RunConfig, and no env var is appended to the proxy runner container.

Result: any \`MCPServer\` pointed at a password-protected Redis fails on startup with:

\`\`\`
{"level":"ERROR","msg":"error executing command","error":"failed to create Redis session storage: failed to connect to redis: NOAUTH Authentication required."}
\`\`\`

reproducible at v0.27.2 and on \`main\` as of this PR's base.

## Why this was missing

[PR #4368](https://github.com/stacklok/toolhive/pull/4368) (which introduced MCPServer runconfig wiring) explicitly noted:

> the Redis password is intentionally excluded and injected as a pod env var instead

The env-var injection then landed for \`VirtualMCPServer\` in [#4215](https://github.com/stacklok/toolhive/issues/4215) via \`buildRedisPasswordEnvVar\` — but the parallel MCPServer change wasn't made. From the user's side, the field looks supported (CRD validation passes, runner accepts the value) but silently no-ops at runtime.

## Fix

Mirror the VirtualMCPServer pattern:

1. **\`mcpserver_controller.go\`** — new helper \`buildMCPServerRedisPasswordEnvVar\` that returns \`[]corev1.EnvVar\` with a single \`SecretKeyRef\`-backed \`THV_SESSION_REDIS_PASSWORD\` entry when \`sessionStorage.provider == \"redis\"\` and \`passwordRef\` is set; returns \`nil\` otherwise.
2. Append its output to the proxy container's env slice in the MCPServer reconciler, immediately **before** \`resourceOverrides.proxyDeployment.env\` so user-supplied overrides still win on name collisions.
3. **\`mcpserver_replicas_test.go\`** — add \`TestBuildMCPServerRedisPasswordEnvVar\` mirroring \`TestBuildRedisPasswordEnvVar\` for VirtualMCPServer (nil storage, memory provider, redis without passwordRef, redis with passwordRef).

## Scope

- No CRD changes — the schema already supports \`passwordRef\`.
- No behavior change when \`passwordRef\` is unset — the helper returns \`nil\` and no env var is added.
- Identical secret-mounting semantics to the VirtualMCPServer path, so security-review notes from #4215 carry over.

## Test plan

- [x] \`go build ./cmd/thv-operator/...\`
- [x] \`go test -count=1 ./cmd/thv-operator/controllers/\` — all green (3.7s)
- [x] New \`TestBuildMCPServerRedisPasswordEnvVar\` covers four cases (nil, memory, redis-no-passwordRef, redis-with-passwordRef) and asserts the env var uses \`SecretKeyRef\` (no plaintext \`Value\`).
- [ ] Integration verification: deployed locally against an MCPServer running with \`replicas: 2\` against a password-protected Redis; proxy connects cleanly.

## Related

- #4368 — MCPServer runconfig wiring (deferred password env var)
- #4215 — VirtualMCPServer Redis password env var (precedent for this PR's pattern)
- #4217 — MCPServer reconciler/builder scaling support (out-of-scope sister)

### DIFF
--- a/cmd/thv-operator/controllers/mcpserver_controller.go
+++ b/cmd/thv-operator/controllers/mcpserver_controller.go
@@ -42,6 +42,7 @@ import (
 	"github.com/stacklok/toolhive/cmd/thv-operator/pkg/validation"
 	"github.com/stacklok/toolhive/pkg/container/kubernetes"
 	"github.com/stacklok/toolhive/pkg/transport"
+	"github.com/stacklok/toolhive/pkg/transport/session"
 )
 
 // MCPServerReconciler reconciles a MCPServer object
@@ -1144,6 +1145,11 @@ func (r *MCPServerReconciler) deploymentForMCPServer(
 			})
 		}
 	}
+
+	// Mount Redis password secret when session storage provider is Redis.
+	// Appended after user overrides so the secretRef-backed env wins on
+	// name collision (ResourceOverrides.Env only accepts plain strings).
+	env = append(env, r.buildRedisPasswordEnvVar(m)...)
 
 	// Add volume mounts for user-defined volumes
 	for _, v := range m.Spec.Volumes {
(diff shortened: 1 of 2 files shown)


jhrozek (approved):

jhrozek:
thanks for the PR @dallinstevens and I apologise for the delay in having this reviewed

PR #5288, opened 2026-05-14
Validate upstream JSON-RPC responses in transparent proxy

### LINKED ISSUE #5247: security: proxy forwards malformed upstream JSON-RPC frames to clients without validation
## Summary

The transparent proxy (`thv run`) forwards upstream MCP server responses to clients without validating JSON-RPC structure. When an upstream server returns a malformed frame — missing `jsonrpc` field, invalid `id` type, non-object body — the proxy passes it through with HTTP 200 rather than returning a structured error.

## Why the proxy should validate

The proxy sits between untrusted upstream MCP servers and trusted clients. An attacker who controls an upstream server (supply chain compromise, misconfigured registry, prompt injection redirecting to a malicious server) can send malformed frames that:

- Crash MCP clients that validate JSON-RPC strictly
- Cause clients to misinterpret responses and act on garbage data
- Trigger undefined behaviour in client-side JSON parsers

The MCP spec (JSON-RPC 2.0) requires responses to contain `"jsonrpc": "2.0"` and a valid `id`. A proxy that silently forwards protocol violations provides weaker security guarantees than one that rejects them at the boundary.

## Observed behaviour

The transparent proxy reads the upstream HTTP response and writes it to the client unchanged — including HTTP status and body — with no JSON-RPC structural validation.

**Reproduction:** Start `thv` proxying a server that returns:
```json
{"id": "not-an-integer", "result": {"broken": true}}
```
(missing `"jsonrpc"` field; `id` is a string, not an integer or null)

The proxy returns HTTP 200 with that body to the client.

## Expected behaviour

The proxy should detect structurally invalid JSON-RPC responses from upstream and return a structured error (HTTP 400 or a JSON-RPC error object) to the client rather than forwarding the malformed frame.

## Suggested fix

Add upstream response validation in the transparent proxy's response handling path. For JSON responses, verify the presence of `"jsonrpc": "2.0"` and that `id` is a string, number, or null before forwarding. Return a structured JSON-RPC error to the client when validation fails.

Discovered via DAST scanning in stacklok/stacklok-enterprise-platform.

bishnubista:
## Summary

The transparent proxy (`thv run`) currently forwards upstream MCP server responses to clients with HTTP 200 even when the JSON-RPC frame is structurally invalid — missing `jsonrpc:"2.0"`, invalid `id` type, non-object body, etc. The proxy sits between potentially untrusted upstream MCP servers and trusted clients, so a compromised or misconfigured upstream can push malformed frames straight through, where they may crash strict JSON-RPC parsers, be misinterpreted as legitimate responses, or trigger undefined behaviour in client-side parsers.

This PR validates upstream JSON-RPC frames at the proxy boundary and rewrites invalid responses into a structured JSON-RPC error so the proxy stops being a silent amplifier for malformed (or adversarial) upstream servers.

- Validation runs in `NoOpResponseProcessor.ProcessResponse` (`streamable-http` / default transport). SSE is unaffected.
- Structural rules: single JSON value (trailing JSON rejected); object or non-empty array of objects; per object `jsonrpc=="2.0"`, `id` ∈ {string, number, null}, exactly one of `result`/`error`; `error.code` must be an integer, `error.message` must be a string.
- Gating: `POST` + `200`, `application/json`(`-rpc`) media type, non-identity `Content-Encoding` passed through untouched, request must carry an MCP streamable-HTTP signal (`MCP-Protocol-Version` or `Mcp-Session-Id`) so non-MCP JSON traffic flowing through the catch-all is not rewritten.
- Invalid frame → HTTP 502 with body `{"jsonrpc":"2.0","id":null,"error":{"code":-32000,"message":"Invalid upstream JSON-RPC response","data":"<detail>"}}`. `-32000` is the JSON-RPC implementation-defined server-error code; `-32603` is reserved for internal JSON-RPC errors. Response headers are replaced wholesale so upstream `Mcp-Session-Id`, `Set-Cookie`, `ETag`, `Cache-Control`, etc. are not smuggled into the proxy-generated error.
- Body reads bounded by `io.LimitReader` at `maxJSONRPCResponseBytes = 100 << 20` (100 MiB), matching streamable-HTTP precedent in `pkg/vmcp/client` and `pkg/vmcp/session/internal/backend`. Oversized responses are rejected with the same 502 shape so the proxy cannot be amplified into a memory DoS.

Closes #5247

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Local verification:

- `go test ./pkg/transport/proxy/transparent/... -count=1` — pass.
- `golangci-lint run --timeout=5m ./pkg/transport/proxy/transparent/...` — 0 issues.

New test cases in `pkg/transport/proxy/transparent/response_processor_test.go`:

- Valid pass-through: result, error, batch, `result: null`, `application/json; charset=utf-8`.
- Invalid → 502: missing `jsonrpc`, invalid `id` type, non-object body, `result`+`error` both present, trailing JSON, fractional `error.code`.
- Non-validated pass-through: non-POST, non-200, non-JSON, `text/event-stream`, `application/jsonsomethingelse`.
- Compressed responses (`Content-Encoding: gzip`) passed through unchanged for both valid and malformed payloads; explicit `Content-Encoding: identity` still validates.
- MCP-signal gate: no MCP request headers → pass through even with malformed body; `MCP-Protocol-Version` or `Mcp-Session-Id` → validate.
- Rewritten 502 strips `Mcp-Session-Id`, `Set-Cookie`, `Etag`, `Cache-Control` from the response.
- Oversized response (> `maxJSONRPCResponseBytes`) is rejected with a size-limit error.

## API Compatibility

No CRD or operator API surface is touched.

Yes. Clients of the transparent proxy that previously received malformed upstream responses verbatim with HTTP 200 will now receive HTTP 502 with a synthetic JSON-RPC error body. Conformant upstream MCP servers are unaffected — only structurally invalid responses are rewritten.

Two scope decisions called out explicitly so reviewers can push back if either is wrong for the project:

1. **Validation runs in `ModifyResponse`, after `tracingTransport.RoundTrip` may have observed an upstream `Mcp-Session-Id` and registered proxy-side session state.** This PR strips upstream session headers from the proxy-generated 502 so clients never receive a session id derived from a malformed initialize response, but it does not roll back server-side proxy session state created before validation. Moving validation earlier (or adding a rollback path for invalid initialize responses) touches `tracingTransport` and felt outside the scope of "structurally validate the upstream frame." Happy to open a follow-up — or fold it into this PR if maintainers prefer.

2. **Backward-compat clients that POST without `MCP-Protocol-Version` on the very first initialize will not trigger validation** — same as today's behaviour, so no regression. If broader coverage is preferred, `tracingTransport.RoundTrip` already buffers and parses the request body to detect `initialize`; a small context-propagated marker would let `ProcessResponse` validate those frames too. Easy follow-up.

Open to feedback on:
- The 100 MiB body cap (matched `pkg/vmcp` precedent; can tighten to 10 MiB if maintainers prefer an explicit security default).
- Whether the MCP-signal narrowing should be widened to also accept the buffered-request-body sniff above.


pkg/transport/proxy/transparent/response_processor.go
```diff
@@ -36,6 +80,154 @@ func (*NoOpResponseProcessor) ShouldProcess(_ *http.Response) bool {
...
+func validateJSONRPCResponse(body []byte) error {
+	var payload any
+	dec := json.NewDecoder(bytes.NewReader(body))
+	if err := dec.Decode(&payload); err != nil {
+		return fmt.Errorf("invalid JSON body: %w", err)
+	}
+	if dec.More() {
+		return fmt.Errorf("JSON-RPC response must contain a single JSON value")
+	}
+	if err := dec.Decode(&struct{}{}); err != io.EOF {
```
amirejaz: The second `Decode` here is unreachable. When `dec.More()` returns false there is nothing left in the stream, so the subsequent `Decode` will always return `io.EOF` and the condition never fires. The `More()` check above already covers the trailing-value case completely. Nit, but worth removing the dead branch.

pkg/transport/proxy/transparent/response_processor.go
```diff
@@ -22,12 +40,38 @@ type ResponseProcessor interface {
 	ShouldProcess(resp *http.Response) bool
 }
 
-// NoOpResponseProcessor is a processor that does nothing.
-// Used for transports that don't require response processing (e.g., streamable-http).
+// NoOpResponseProcessor is the default processor for non-SSE transports.
```
amirejaz, line 43: Scoping to non-SSE makes sense for now — SSE validation would need a fundamentally different approach. By the time `ProcessResponse` runs on an SSE response, the `200 OK` and `Content-Type: text/event-stream` headers are already committed downstream, so there is no way to rewrite to 502 for a malformed mid-stream event. A proper fix would need a per-event streaming interceptor inside the SSE processor that can either synthesize an error event or close the stream — a meaningfully larger scope than this PR.

Worth opening a follow-up issue to track it explicitly, so it is clear the gap is deferred rather than forgotten.

JAORMX:
We have middleware that does the validation and parsing. This happens outside of the transparent proxy, but still in the path. Why not use the existing middleware instead?

bishnubista:
> We have middleware that does the validation and parsing. This happens outside of the transparent proxy, but still in the path. Why not use the existing middleware instead?

Good point. I agree the JSON-RPC validation logic should not live as bespoke protocol code in `pkg/transport/proxy/transparent`.

I traced the path again: the existing MCP parser middleware runs before the transparent proxy and parses inbound requests into context, but I didn’t find an existing generic response validator that can inspect upstream response bodies and rewrite them before they reach the client. The response-side parsing I found is authz-specific or audit-only/lenient.

I’ll adjust this by moving the response validation helper into `pkg/mcp`, using `mcp.GetParsedMCPRequest(resp.Request.Context())` as the primary MCP signal, and keeping the current header fallback for batch/backcompat cases. I think `ModifyResponse` should remain the enforcement point because it can still rewrite the upstream response to 502 before the reverse proxy commits it.

pkg/transport/proxy/transparent/response_processor.go
```diff
@@ -22,12 +40,38 @@ type ResponseProcessor interface {
...
+// It validates JSON-RPC responses for streamable HTTP and otherwise leaves responses unchanged.
 type NoOpResponseProcessor struct{}
```
aponcedeleonch, line 45: Non-blocking nit: this type is no longer a no-op (the doc comment even says it validates now), and ShouldProcess still returns false while ProcessResponse does real work. That's safe today only because modifyResponse calls ProcessResponse directly, but a future caller that gates on ShouldProcess would silently skip validation. Worth renaming (e.g. StreamableHTTPResponseProcessor) and making ShouldProcess reflect reality.

pkg/transport/proxy/transparent/response_processor.go
```diff
@@ -22,12 +40,38 @@ type ResponseProcessor interface {
...
-// ProcessResponse is a no-op implementation.
-func (*NoOpResponseProcessor) ProcessResponse(_ *http.Response) error {
+// ProcessResponse validates JSON-RPC responses when applicable.
+func (*NoOpResponseProcessor) ProcessResponse(resp *http.Response) error {
+	if !shouldValidateJSONRPCResponse(resp) {
+		return nil
+	}
+
+	// Read one byte past the cap so we can detect oversize without allocating beyond it.
+	body, err := io.ReadAll(io.LimitReader(resp.Body, maxJSONRPCResponseBytes+1))
```
aponcedeleonch, line 54: Non-blocking: this buffers the entire response body (up to 100 MiB) and parses it on every streamable-http response, replacing the previously streamed path. Fine for typical MCP payloads and the cap bounds memory, but large tool outputs now incur a full buffer plus parse. Worth a comment here or a follow-up if latency on big results ever matters.

aponcedeleonch (approved):
Thanks, this is a well-scoped security fix. Process note for next time: please ask to have the issue assigned to you before opening a PR, per our [contributor guidelines](https://github.com/stacklok/toolhive/blob/main/CONTRIBUTING.md#claiming-an-issue). #5247 is the linked issue.

On the change: I confirmed both halves. On main the transparent proxy's response hook is a literal no-op invoked unconditionally, so malformed upstream frames reach clients as 200s. This PR adds strict JSON-RPC 2.0 validation at the streamable-http boundary and rewrites invalid frames into a bounded, header-scrubbed 502, gated conservatively so it won't false-reject SSE, compressed bodies, or non-MCP JSON. Bodies are forwarded byte-for-byte, so no int-precision or re-serialization risk. Builds, passes the touched packages plus full-repo lint, and the test matrix is thorough.

Reconciling with the earlier review: JAORMX's point about reusing existing validation instead of bespoke protocol code is addressed. The helper now lives in pkg/mcp (ValidateJSONRPCResponseBody) and reuses the parsed-request signal. amirejaz's dead-branch nit is also gone: the old dec.More() construct was replaced with a reachable single-value guard that's now tested.

Approving. Left a couple of non-blocking nits inline. One follow-up ask that isn't code: amirejaz suggested opening an issue to track the deferred SSE-response validation gap (an SSE stream can't be rewritten to 502 mid-response), worth opening so the gap stays explicit.

pkg/transport/proxy/transparent/response_processor.go
aponcedeleonch: Addressed: the dec.More() construct was replaced with a reachable single-value guard (Decode then expect io.EOF), now covered by trailing-value tests.

pkg/transport/proxy/transparent/response_processor.go
aponcedeleonch: Accepted scope: non-SSE only for now, and I flagged the deferred SSE-response validation gap in my review to be tracked as a follow-up issue.

aponcedeleonch (changes requested):
Upon further review the CI is failing, we should fix it before merging

PR #5290, opened 2026-05-15
Resolve authz `ConfigMap` for `VirtualMCPServer`

### LINKED ISSUE #4919: VirtualMCPServer doesn't resolve authzConfig.configMap — vMCP rejects 'type: configMap' at startup

### LINKED ISSUE #5208: vMCP: Load authz config from ConfigMap (policies + primaryUpstreamProvider)

### LINKED ISSUE #5277: vMCP AuthzConfig struct missing GroupEntityType, GroupClaimName, RoleClaimName — Cedar group/role policies fail silently
### Background

PR #5121 added `GroupEntityType` to `cedar.ConfigOptions` so deployments can configure the entity type the Cedar authorizer uses when building principal parent UIDs from JWT group/role claims. `GroupClaimName` and `RoleClaimName` were already present. All three fields are read from `authz.json` at runtime by MCPServer and MCPRemoteProxy, which load the file directly via `LoadConfig()` → `ConfigOptions`.

The enterprise controller writes all three into `authz.json`:

```json
{
  "cedar": {
    "group_entity_type": "ClaimGroup",
    "group_claim_name":  "groups",
    "role_claim_name":   "roles"
  }
}
```

### Problem

The vmcp internal `AuthzConfig` struct (`pkg/vmcp/config/config.go`) only carries:

```go
type AuthzConfig struct {
    Type                    string   `json:"type"`
    Policies                []string `json:"policies,omitempty"`
    PrimaryUpstreamProvider string   `json:"primaryUpstreamProvider,omitempty"`
}
```

`GroupEntityType`, `GroupClaimName`, and `RoleClaimName` are absent. The operator converter (`pkg/vmcpconfig/converter.go`) builds this struct manually for both the inline and ConfigMap paths, and has nowhere to put these values. They are silently dropped.

As a result, the Cedar authorizer inside the vmcp binary is always constructed with:
- `GroupEntityType` defaulting to `"THVGroup"` (the upstream hardcoded legacy value)
- `GroupClaimName` defaulting to the well-known fallback list (`groups`, `roles`, `cognito:groups`)
- `RoleClaimName` empty

For deployments using the enterprise controller, the static entity store uses `ClaimGroup` entities. Because `GroupEntityType` is never forwarded, the runtime Cedar authorizer builds `THVGroup::` parents for the principal while the static entity store has `ClaimGroup::` entities — severing the `Client → ClaimGroup → PlatformRole` hierarchy. Cedar `in` checks silently evaluate to false and all group/role-scoped policies deny.

This affects both the inline path today and the ConfigMap path once #5208 lands.

### Fix

1. Add `GroupEntityType`, `GroupClaimName`, and `RoleClaimName` to `pkg/vmcp/config/config.go`'s `AuthzConfig` struct.
2. Wire them through the Cedar authorizer factory so the vmcp binary passes them to `NewCedarAuthorizer` via `ConfigOptions`.
3. For the **ConfigMap path** (#5208): extract all three from the parsed `authz.json` and populate them into `AuthzConfig`.
4. For the **inline path**: either add these fields to `InlineAuthzConfig` (CRD change), or lift them to `AuthzConfigRef` as part of #5208's Option B schema decision — the approach should be coordinated with that issue.

### Related

- #5208 — converter ConfigMap loading (the ConfigMap path won't forward these without this fix)
- #5121 — added `GroupEntityType` to `ConfigOptions` (the receiving end this issue completes for vMCP)
- #5270 — authz ConfigMap watch (companion to #5208)

### Acceptance criteria

- A VirtualMCPServer with Cedar group/role policies enforces them correctly at runtime
- `group_entity_type`, `group_claim_name`, and `role_claim_name` from an enterprise `authz.json` ConfigMap are forwarded into the vmcp `config.yaml` and reach the Cedar authorizer

blkt:
A `VirtualMCPServer` with `spec.incomingAuth.authzConfig.type: configMap` silently produced a vmcp `config.yaml` that referenced the unresolved `configMap` type token. The vmcp binary's `AuthzConfig` validator only accepts `cedar` or `none`, so the pod crashed in `CrashLoopBackOff` at startup. Inline authz also silently dropped `GroupClaimName`, `RoleClaimName`, `GroupEntityType`, and `EntitiesJSON`, so any enterprise Cedar policy that walked a `Client → ClaimGroup → PlatformRole` hierarchy denied every request because the runtime Cedar authorizer built `THVGroup::` parents while the entity store contained `ClaimGroup::` entities.

Wire the configMap path end-to-end, plumb the four missing fields through both source paths, and move `PrimaryUpstreamProvider` onto the auth server config where it belongs:

* Extract `LoadAuthzConfigFromConfigMap` as the shared fetch/parse/ validate helper in `controllerutil`; `AddAuthzConfigOptions` now delegates to it. The vMCP converter calls the same helper so the failure modes match the `MCPServer`/`MCPRemoteProxy` runner path.

* Extend `pkg/vmcp/config.AuthzConfig` with `EntitiesJSON`, `GroupClaimName`, `RoleClaimName`, `GroupEntityType`, and forward all four into `cedar.ConfigOptions` in the Cedar middleware factory. `EntitiesJSON` defaults to `"[]"` when unset to preserve the historical Cedar contract.

* Lift the source-agnostic Cedar JWT-claim mapping fields (`GroupClaimName`, `RoleClaimName`, `GroupEntityType`) onto `AuthzConfigRef` so they work identically for inline and configMap users. For configMap users the parsed payload provides the default and the spec-level field overrides when set.

* Move `PrimaryUpstreamProvider` onto `EmbeddedAuthServerConfig` (`spec.authServerConfig.primaryUpstreamProvider` on `VirtualMCPServer`). The field describes which upstream IDP token Cedar reads claims from, which is a property of the embedded auth server, not of the authz policies. Placing it next to `upstreamProviders` co-locates the choice with the set it selects from. Reachable from `MCPExternalAuthConfig` of type `embeddedAuthServer` as well; on `MCPServer`/`MCPRemoteProxy` (single-upstream consumers) the only validated values are empty (auto-select) or the name of that single upstream. The legacy `spec.incomingAuth.authzConfig.inline.primaryUpstreamProvider` location is read as a backward-compatibility fallback for one release with a `Warning` event of reason `AuthzPrimaryUpstreamProviderDeprecated`.

* Pre-validate the referenced authz `ConfigMap` in the controller and distinguish `NotFound` from other parse/validation failures via two condition reasons on `AuthConfigured`: the existing shared `AuthzConfigMapNotFound` and a new `AuthzConfigMapInvalid`. This mirrors the diagnostic `MCPRemoteProxy` emits today and gives users a status-level error before the converter fails opaquely.

* Validate the resolved `PrimaryUpstreamProvider` (from either source) against the declared embedded auth server upstreams so an unresolvable provider is rejected at convert time.

CRD compatibility: not a breaking change. The legacy `spec.incomingAuth.authzConfig.inline.primaryUpstreamProvider` location continues to work, with a Warning event emitted whenever it is read; planned removal one release after the deprecation cycle. The new fields on `AuthzConfigRef` (`GroupClaimName`, `RoleClaimName`, `GroupEntityType`) and on `EmbeddedAuthServerConfig` (`PrimaryUpstreamProvider`) are additive.

Closes #4919
Closes #5208
Closes #5277

## Large PR Justification

~950 lines of code are relevant, everything else is auto-generated.


blkt:
Heads up on dropping `InlineAuthzConfig.PrimaryUpstreamProvider`. Flagging because it's a breaking change.

The four Cedar fields touched here split into two buckets:

- **Policy content** (`Policies, `EntitiesJSON`): lives wherever the policies live.
- **JWT-claim mapping** (`PrimaryUpstreamProvider`, `GroupClaimName`, `RoleClaimName`, `GroupEntityType`): describes how Cedar reads the token, a separate concern from the policies themselves.

So the second bucket went onto `AuthzConfigRef` directly. Same field, same semantics, regardless of whether you're inline or configMap.

The non-breaking alternative was just leaving `PrimaryUpstreamProvider` on `InlineAuthzConfig`. Two reasons I didn't:

- `configMap` users would still have no spec-level knob; they'd have to mutate the `ConfigMap`, and in our enterprise flow that's owned by an external controller.
- Two YAML paths for the same value, plus `ExplicitPrimaryUpstreamProvider()` keeps an inline-only branch that's easy to miss.

The cost is the schema break: anyone setting `spec.incomingAuth.authzConfig.inline.primaryUpstreamProvider` today needs to move it up one level. If that feels too aggressive, happy to do a deprecation window instead (keep the inline field for a release, fall through in the helper, remove next minor). Small isolated change.

ChrisJBurns (commented):
**Agents consulted**: kubernetes-expert, go-security-reviewer, code-reviewer, toolhive-expert, go-expert-developer, documentation-writer

### Consensus Summary
| 1 | New AuthzConfigRef claim-mapping fields dropped on MCPServer/MCPRemoteProxy | 10/10 | HIGH | Fix |
| 2 | virtualmcpserver-api.md still pins primaryUpstreamProvider at deprecated location | 10/10 | HIGH | Fix |
| 3 | virtualmcpserver-kubernetes-guide.md troubleshooting uses deprecated location | 10/10 | HIGH | Fix |
| 4 | EmbeddedAuthServerConfig.PrimaryUpstreamProvider neither validated nor wired on MCPServer/MCPRemoteProxy | 10/10 | MEDIUM | Fix or soften docstring |
| 5 | Comment drift on InlineAuthzConfig.PrimaryUpstreamProvider deprecation | 8/10 | MEDIUM | Fix |
| 6 | AuthzConfigMapInvalid not mirrored on MCPRemoteProxy validateK8sRefs | 8/10 | MEDIUM | Follow-up |
| 7 | No migration guidance for primaryUpstreamProvider move | 8/10 | MEDIUM | Note |
| 8 | Naming collision: ExplicitPrimaryUpstreamProvider on two types, different signatures | 7/10 | MEDIUM | Rename |
| 9 | convertAuthzConfig silently accepts unknown authz types | 7/10 | LOW | Guard |
| 10 | LoadAuthzConfigFromConfigMap dedicated tests missing | 8/10 | LOW | Tests |
| 11 | validateAuthzConfigMapRef doesn't run cedar.ExtractConfig | 9/10 | LOW | Add check |
| 12 | Architecture docs not updated | 8/10 | LOW | Optional |
| 13 | AuthzConfigMapInvalid undocumented in troubleshooting | 7/10 | LOW | Note |
| 14 | Deprecation Warning event fires every reconcile | 7/10 | LOW | Optional |
| 15 | Deprecation Warning event not emitted in no-auth-server branch | 7/10 | LOW | Emit |

### Overall
The PR fixes two real bugs — the vmcp `CrashLoopBackOff` on `type: configMap`, and the `EntitiesJSON="[]"` silent-deny on Cedar policies that walk `Client → ClaimGroup → PlatformRole`. The vMCP-side wiring is clean: a shared `LoadAuthzConfigFromConfigMap`, distinct condition reasons (`AuthzConfigMapNotFound` vs new `AuthzConfigMapInvalid`), defense-in-depth at convert time, and thorough table-driven tests that lock the deprecation precedence and the event emission contract.

The findings cluster around one structural concern: the new fields are added to *shared* types (`AuthzConfigRef` for the JWT-claim mapping; `EmbeddedAuthServerConfig` for `PrimaryUpstreamProvider`) but the runtime wiring and the user-facing docs only land on the `VirtualMCPServer` consumer. `MCPServer` and `MCPRemoteProxy` users who follow the PR's own deprecation notes will hit silent no-ops. None of this blocks the vMCP fix shipping, but the field additions to `AuthzConfigRef` are particularly worth resolving before merge — either by threading them through `addAuthzInlineConfigOptions` and the configMap branch of `AddAuthzConfigOptions`, or by tightening the docstrings to say "VirtualMCPServer-only" with an explicit advisory condition on the other CRDs.

The remaining items are smaller: a few low-severity test gaps, a `default:` arm missing on `convertAuthzConfig`, the docs sync for the user-facing guides, and a couple of minor consistency nits in the validator.

### Documentation
- `docs/operator/virtualmcpserver-api.md:90` still lists `primaryUpstreamProvider` only at `authzConfig.inline.primaryUpstreamProvider`. The canonical location (`spec.authServerConfig.primaryUpstreamProvider`) and the new `groupClaimName` / `roleClaimName` / `groupEntityType` fields on `AuthzConfigRef` are absent. **[F2]**
- `docs/operator/virtualmcpserver-kubernetes-guide.md:631-637` troubleshooting block pins primary upstream at the deprecated inline location; following the guide will produce `AuthzPrimaryUpstreamProviderDeprecated` events on every reconcile. **[F3]**
- No CHANGELOG / RELEASE-NOTES / migration doc captures the deprecation. Consider a short "Migration: primaryUpstreamProvider location" subsection in the kubernetes guide. The Warning event message text is currently the only guidance and it's buried in `kubectl get events`. **[F7]**
- `docs/arch/10-virtual-mcp-architecture.md` (or `09-operator-architecture.md`) doesn't describe the new shared `LoadAuthzConfigFromConfigMap` resolver, the spec-vs-ConfigMap precedence for claim-mapping fields, or the `PrimaryUpstreamProvider` relocation. Optional follow-up. **[F12]**
- The `AuthConfigured=False` reasons (`AuthzConfigMapNotFound`, `AuthzConfigMapInvalid`) aren't called out in the "Authorization Policy Errors" troubleshooting block of the kubernetes guide. **[F13]**

---
*Generated with Claude Code*

cmd/thv-operator/pkg/controllerutil/authz.go
```diff
@@ -213,59 +282,14 @@ func AddAuthzConfigOptions(
 		return addAuthzInlineConfigOptions(authzRef, options)
 
 	case mcpv1beta1.AuthzConfigTypeConfigMap:
```
ChrisJBurns, line 348: **\[F1 HIGH — consensus 10/10]** New `AuthzConfigRef` claim-mapping fields silently dropped on MCPServer/MCPRemoteProxy runner path.

The PR adds `GroupClaimName`, `RoleClaimName`, and `GroupEntityType` to the shared `AuthzConfigRef` type (used by all three CRDs) and the CRD docstring explicitly promises spec-over-ConfigMap override semantics. The vMCP converter (`convertAuthzConfig`) threads them through to `cedar.ConfigOptions`. But neither `addAuthzInlineConfigOptions` (line 169, the inline runner path) nor this configMap branch applies them — the loaded `*authz.Config` is passed straight to `runner.WithAuthzConfig`. Setting any of these fields on an MCPServer or MCPRemoteProxy is a silent no-op at runtime, which contradicts both the field documentation and the "no silent no-ops" rule in `.claude/rules/go-style.md`.

**Fix**: thread the three fields into the `cedar.ConfigOptions` built by `addAuthzInlineConfigOptions`; in this configMap branch, after `LoadAuthzConfigFromConfigMap`, apply the spec-wins-over-ConfigMap override the vMCP converter performs (mutating a copy of `cfg`). Add a unit test asserting the values flow into the built `runner.WithAuthzConfig` argument — there is currently zero coverage of the new fields on the MCPServer/MCPRemoteProxy code paths.

*Raised by code-reviewer, toolhive-expert, go-security-reviewer.*

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -225,6 +225,19 @@ type EmbeddedAuthServerConfig struct {
...
+	// of one of the entries in UpstreamProviders. When empty, the controller
+	// auto-selects the first entry of UpstreamProviders. Only meaningful when
+	// at least one upstream is configured; on MCPServer and MCPRemoteProxy
+	// (single-upstream consumers) the only validated values are empty or the
+	// name of the sole upstream.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`
+	PrimaryUpstreamProvider string `json:"primaryUpstreamProvider,omitempty"`
```
ChrisJBurns, line 253: **\[F4 MEDIUM — consensus 10/10]** This field is unreachable on the MCPServer/MCPRemoteProxy path.

The docstring claims "on MCPServer and MCPRemoteProxy (single-upstream consumers) the only validated values are empty or the name of the sole upstream" — but no validator in `mcpserver_controller.go`, `mcpremoteproxy_controller.go`, or `pkg/controllerutil/` references `EmbeddedAuthServerConfig.PrimaryUpstreamProvider`, and the field is never threaded into the proxyrunner configuration. The structural advisory `AuthzPrimaryUpstreamProviderIgnored` only inspects the legacy `InlineAuthzConfig.PrimaryUpstreamProvider`. Result: users following the deprecation guidance "move the value to embeddedAuthServer.primaryUpstreamProvider" get no error, no warning, and no effect.

**Fix**: either (a) wire the new field through the MCPServer/MCPRemoteProxy → MCPExternalAuthConfig consumer and emit `AuthzPrimaryUpstreamProviderIgnored` when set on those CRDs, or (b) soften this docstring to say the field is structurally meaningful only on VirtualMCPServer until further notice.

*Raised by kubernetes-expert, toolhive-expert.*

cmd/thv-operator/api/v1beta1/mcpserver_types.go
```diff
@@ -736,30 +766,38 @@ func (r *MCPGroupRef) GetName() string {
...
-	// spec.authServerConfig.upstreamProviders; otherwise the VirtualMCPServer is
-	// rejected with AuthServerConfigValidated=False. MCPServer and MCPRemoteProxy
-	// have no embedded auth server; setting this field on those CRs surfaces an
-	// AuthzPrimaryUpstreamProviderIgnored advisory condition on the resource.
+	// PrimaryUpstreamProvider names the upstream IDP whose access token's
+	// claims Cedar should evaluate.
+	//
+	// Deprecated: this field has moved to
+	// spec.authServerConfig.primaryUpstreamProvider on VirtualMCPServer (or to
+	// the embeddedAuthServer.primaryUpstreamProvider field of a referenced
```
ChrisJBurns: **\[F5 MEDIUM — consensus 8/10]** The deprecation block self-contradicts.

The first half points users to "the embeddedAuthServer.primaryUpstreamProvider field of a referenced MCPExternalAuthConfig for MCPServer and MCPRemoteProxy". The second half then says "the field remains structurally meaningless (no embedded auth server runtime) and continues to surface the AuthzPrimaryUpstreamProviderIgnored advisory". Both can't be true — either there's a canonical replacement on MCPServer/MCPRemoteProxy or there isn't. Per `.claude/rules/go-style.md` "Keep Comments Synchronized With Code".

**Fix**: pick one stance. If MCPServer/MCPRemoteProxy have no canonical home today (matches the actual wiring — see F4), reword to: "this field has always been a structural no-op on MCPServer and MCPRemoteProxy (no embedded auth server). The deprecation only applies to VirtualMCPServer, where it has moved to `spec.authServerConfig.primaryUpstreamProvider`."

*Raised by code-reviewer.*

cmd/thv-operator/api/v1beta1/virtualmcpserver_types.go
```diff
@@ -501,6 +508,28 @@ func (r *VirtualMCPServer) ResolveGroupName() string {
...
+// provider name and a flag indicating whether the value came from the
+// deprecated spec.incomingAuth.authzConfig.inline.primaryUpstreamProvider
+// location (fromDeprecated=true) or the canonical
+// spec.authServerConfig.primaryUpstreamProvider location (fromDeprecated=false).
+// Returns ("", false) when neither location is set.
+//
+// Precedence: the canonical location wins if set; the deprecated location is
+// read only as a backward-compatibility fallback. Callers should emit a
+// Warning event when fromDeprecated is true.
+func (r *VirtualMCPServer) ExplicitPrimaryUpstreamProvider() (name string, fromDeprecated bool) {
```
ChrisJBurns, line 521: **\[F8 MEDIUM — consensus 7/10]** `ExplicitPrimaryUpstreamProvider` now exists on two related types with different signatures.

`AuthzConfigRef.ExplicitPrimaryUpstreamProvider()` was renamed to `DeprecatedInlinePrimaryUpstreamProvider()`, and this new method takes its name with a different return shape (`(name string, fromDeprecated bool)`). A reader grepping for `ExplicitPrimaryUpstreamProvider` will hit two definitions and can't tell at a glance which receiver/return is in play — exactly the kind of API confusion that bites navigation.

**Fix**: either rename this to `ResolveExplicitPrimaryUpstreamProvider()` or `PrimaryUpstreamProviderWithSource()` to signal the precedence-resolution semantic, or fold the deprecated-fallback logic into a single method `AuthzConfigRef.PrimaryUpstreamProvider() (name string, fromDeprecated bool)` and have `VirtualMCPServer` compose the two locations.

*Raised by code-reviewer.*

cmd/thv-operator/controllers/mcpremoteproxy_controller.go
```diff
@@ -1091,7 +1091,7 @@ func (r *MCPRemoteProxyReconciler) validateGroupRef(ctx context.Context, proxy *
 // Mirrors the validateGroupRef convention: this only sets/removes the
 // condition; the caller is responsible for persisting status.
 func (*MCPRemoteProxyReconciler) validateAuthzPrimaryUpstreamProviderIgnored(proxy *mcpv1beta1.MCPRemoteProxy) {
-	provider := proxy.Spec.AuthzConfig.ExplicitPrimaryUpstreamProvider()
+	provider := proxy.Spec.AuthzConfig.DeprecatedInlinePrimaryUpstreamProvider()
```
ChrisJBurns, line 1094: **\[F6 MEDIUM (parity) — consensus 8/10]** `MCPRemoteProxy` lacks the new `AuthzConfigMapInvalid` reason.

`MCPRemoteProxyReconciler.validateK8sRefs` still routes every ConfigMap failure (NotFound, missing key, malformed payload, Validate-fail) through `ConditionReasonAuthzConfigMapNotFound`. After this PR, `VirtualMCPServer` distinguishes `NotFound` from `Invalid` by calling the shared `LoadAuthzConfigFromConfigMap`. A malformed but present authz ConfigMap on MCPRemoteProxy still leads to pod `CrashLoopBackOff` with no actionable status — the exact failure mode this PR is fixing on VMCP. Per `.claude/rules/operator.md` "Status Condition Parity".

**Fix**: in a follow-up PR (acceptable as a tracking issue linked from here), switch `validateK8sRefs` to use `LoadAuthzConfigFromConfigMap` and add `ConditionReasonAuthzConfigMapInvalid` in `mcpremoteproxy_types.go`. The shared loader makes this mechanical.

*Raised by kubernetes-expert.*

cmd/thv-operator/pkg/vmcpconfig/converter.go
```diff
@@ -174,71 +175,146 @@ func (c *Converter) convertIncomingAuth(
...
+	vmcp *mcpv1beta1.VirtualMCPServer,
+) (*vmcpconfig.AuthzConfig, error) {
+	authzRef := vmcp.Spec.IncomingAuth.AuthzConfig
+
+	// Map Kubernetes API type to vmcp config type. API "inline" maps to vmcp "cedar";
+	// other types pass through (e.g. "configMap" → "cedar" once we populate policies).
+	authzType := authzRef.Type
+	if authzType == authzLabelValueInline || authzType == mcpv1beta1.AuthzConfigTypeConfigMap {
+		authzType = "cedar"
+	}
```
ChrisJBurns: **\[F9 LOW — consensus 7/10]** No `default:` arm on the switch — unknown `authzRef.Type` silently produces a Cedar config with zero policies.

The CRD enum (`Enum: [configMap inline]`) blocks unknown types at admission today, but this converter alone has no defense. A CLI dry-run, a webhook running against a stale schema, or a future authz type would walk past the inline/configMap cases with `authz.Type = "cedar"` and empty `Policies`. Cedar then denies every request silently.

**Fix**: add a `default:` arm that returns `fmt.Errorf("unsupported authz config type: %q", authzRef.Type)` — mirrors the equivalent guard in `AddAuthzConfigOptions` (`controllerutil/authz.go` line 295). The path-mapping comment two lines up ("other types pass through (e.g. 'configMap' → 'cedar' once we populate policies)") also no longer matches the code and should be tightened or dropped.

*Raised by go-expert-developer.*

cmd/thv-operator/pkg/controllerutil/authz.go
```diff
@@ -196,6 +196,75 @@ func addAuthzInlineConfigOptions(
...
+// Failure modes (all returned as errors, never silently succeed):
+//   - authzRef nil or not of type "configMap"
+//   - ConfigMap reference missing name
+//   - kubernetes client not configured
+//   - ConfigMap not found, missing key, empty value, or malformed payload
+//   - authz.Config fails validation
+//
+// The returned *authz.Config is safe to embed directly into RunConfig (via
+// runner.WithAuthzConfig) or to read field-by-field for the vMCP converter.
+func LoadAuthzConfigFromConfigMap(
```
ChrisJBurns, line 224: **\[F10 LOW — consensus 8/10]** No dedicated unit tests for `LoadAuthzConfigFromConfigMap`.

The new helper has six distinct failure modes (nil ref, wrong type, missing name, nil client, NotFound, missing key, empty value, malformed payload, `Validate()` failure). The converter test (`converter_authz_configmap_test.go`) covers the apiserver-side errors (NotFound, missing key, empty, malformed) but not (a) `nil`/wrong-type `authzRef`, (b) `nil` client, or (c) **the highest-value case** — a payload that parses successfully but fails `cfg.Validate()` (e.g. zero policies). The latter is the only thing protecting Cedar from a permit-nothing config.

**Fix**: add a focused `TestLoadAuthzConfigFromConfigMap` in `cmd/thv-operator/pkg/controllerutil/authz_test.go` with table entries for each failure mode and at least one "valid JSON, fails Validate" case.

*Raised by go-expert-developer.*

cmd/thv-operator/controllers/virtualmcpserver_deployment.go
```diff
@@ -1089,6 +1089,28 @@ func (*VirtualMCPServerReconciler) validateBackendAuthSecrets(
...
 }
 
+// validateAuthzConfigMapRef pre-validates the referenced authz ConfigMap when
+// spec.incomingAuth.authzConfig.type is "configMap". It uses the same shared loader
+// as the converter so the diagnostic surfaces the same parse/validation errors the
+// converter would later produce, but earlier in the reconcile and as a status condition
+// rather than as a generic conversion failure. Inline and absent authzConfig are no-ops.
+//
+// Mirrors the pattern in mcpremoteproxy_controller.go's validateK8sRefs.
+func (r *VirtualMCPServerReconciler) validateAuthzConfigMapRef(
```
ChrisJBurns, line 1121: **\[F11 LOW — consensus 9/10]** `validateAuthzConfigMapRef` discards the loaded config and bypasses `cedar.ExtractConfig`.

The pre-validator calls `LoadAuthzConfigFromConfigMap` and ignores the return value. The converter (`convertAuthzConfig`) then re-loads the same ConfigMap and additionally runs `cedar.ExtractConfig` on it. A payload that parses as a valid `authz.Config` but isn't Cedar-flavored (wrong type field, or a future HTTP authorizer) will pass this pre-validator → set `AuthConfigured=True` → fail at convert time with a different error. The two should agree on what "valid" means.

**Fix**: either (a) add a `cedar.ExtractConfig(loaded)` smoke check here, or (b) refactor `LoadAuthzConfigFromConfigMap` to assert Cedar-ness so both sites use the same definition. Add a test case for "valid `authz.Config` but type != cedarv1" returning `AuthzConfigMapInvalid`.

*Raised by code-reviewer, kubernetes-expert.*

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -623,14 +623,13 @@ func (*VirtualMCPServerReconciler) validateAuthzUpstreamAvailable(
 	// admission for the same "fail loudly instead of denying every request"
 	// reason as the configured-AS mismatch path below.
 	if vmcp.Spec.AuthServerConfig == nil {
```
ChrisJBurns, line 644: **\[F15 LOW — consensus 7/10]** Deprecation Warning event not emitted in the no-auth-server branch.

When `vmcp.Spec.AuthServerConfig == nil`, the validator calls `vmcp.ExplicitPrimaryUpstreamProvider()` and discards the `fromDeprecated` flag (`explicitProvider, _ := ...`). Lower at line 674, when `AuthServerConfig != nil`, the same call emits `AuthzPrimaryUpstreamProviderDeprecated` whenever `fromDeprecated` is true. So a user mid-migration (deprecated location set, auth server not yet configured) gets the hard reject but never sees the deprecation hint about the new field location.

**Fix**: emit the same `AuthzPrimaryUpstreamProviderDeprecated` Warning event in this branch before calling `rejectAuthzAdmission`, or extract a small helper that both branches call.

*Raised by toolhive-expert.*

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -665,11 +664,19 @@ func (*VirtualMCPServerReconciler) validateAuthzUpstreamAvailable(
...
-	// request at runtime — fail loudly at admission instead.
-	explicitProvider := vmcp.Spec.IncomingAuth.AuthzConfig.ExplicitPrimaryUpstreamProvider()
+	// If the user has set primaryUpstreamProvider explicitly (either on the
+	// canonical spec.authServerConfig location or on the deprecated
+	// spec.incomingAuth.authzConfig.inline location), the name must resolve to
+	// one of the declared upstreams after normalization on both sides. A
+	// mismatch would cause Cedar to deny every request at runtime — fail loudly
+	// at admission instead.
+	explicitProvider, fromDeprecated := vmcp.ExplicitPrimaryUpstreamProvider()
+	if explicitProvider != "" && fromDeprecated && r.Recorder != nil {
```
ChrisJBurns: **\[F14 LOW — consensus 7/10]** Deprecation Warning event fires on every reconcile cycle while the deprecated field is set.

Kubernetes event aggregation will dedupe within 10 minutes, so the cardinality is bounded — but VMCP reconciles can be frequent (watch resync, dependent resource changes), so this will fire often during the deprecation window. The advisory condition `AuthzPrimaryUpstreamProviderIgnored` already uses the better pattern (`metav1.Condition` with `LastTransitionTime` → emit event on transition only).

**Fix (optional)**: gate this `Eventf` on an `AuthzPrimaryUpstreamProviderDeprecated` status condition so the event fires only on first observation per generation. Acceptable to defer if the team is comfortable with event-aggregation dedup as the only throttle.

*Raised by kubernetes-expert.*

cmd/thv-operator/api/v1beta1/mcpserver_types.go
jhrozek: I read the two sentences as describing different fields rather than contradicting each other:

- Sentence 1 (*"this field has moved to … `embeddedAuthServer.primaryUpstreamProvider` … for MCPServer and MCPRemoteProxy"*) names the **new canonical location**.
- Sentence 2 (*"On MCPServer and MCPRemoteProxy the field remains structurally meaningless"*) describes **this very field** — the deprecated `InlineAuthzConfig.PrimaryUpstreamProvider`.

Both can be true simultaneously: the new location is the canonical replacement, *and* the deprecated form on this type was always a no-op on MCPServer/MCPRemoteProxy.

The real issue underneath this is your F4: the docstring promises a canonical replacement on those CR types that isn't wired. If F4 is resolved by wiring the field, this docstring is accurate as-written; if F4 is resolved by softening, the same softening probably needs to apply here. I'd treat this as subsumed by F4.

cmd/thv-operator/api/v1beta1/virtualmcpserver_types.go
jhrozek: I don't think the rename creates a collision — it removes one:

- `(*VirtualMCPServer).ExplicitPrimaryUpstreamProvider() (string, bool)` is the only definition of `ExplicitPrimaryUpstreamProvider`.
- `(*AuthzConfigRef).DeprecatedInlinePrimaryUpstreamProvider() string` is the only definition of `DeprecatedInlinePrimaryUpstreamProvider`.

A grep for either name lands on exactly one definition; the receiver disambiguates at every call site. The pre-rename world was ambiguous (`ExplicitPrimaryUpstreamProvider` on two receivers), but the rename is precisely the fix. I'd leave this as-is.

cmd/thv-operator/pkg/controllerutil/authz.go
jhrozek: The failure modes are covered, just indirectly via `TestAddAuthzConfigOptions` (which is `LoadAuthzConfigFromConfigMap`'s sole public consumer):

- nil / wrong-type `authzRef` → `authz_test.go:329` ("Nil authz ref returns nil"), `:367` ("Inline type with nil inline config returns error")
- nil client → `:431` ("ConfigMap type with nil client returns error")
- valid JSON that fails `cfg.Validate()` → `:542` ("ConfigMap type with invalid config returns error")
- plus NotFound, missing key, empty value, custom key (lines 453, 476, 509, 616)

Every failure mode is exercised — just not under a `TestLoadAuthzConfigFromConfigMap` name. A direct test would be nice for explicit coverage but doesn't add new behavioral coverage. I'd treat this as nice-to-have rather than required.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
jhrozek: I think this is OK as-is for now. Chris's own comment acknowledges that Event aggregation dedupes within ~10 minutes, so practical cardinality is bounded.

The cleaner fix — gate emission on the `False→True` transition of an `AuthzPrimaryUpstreamProviderDeprecated` *condition* — requires introducing that condition, which is a design decision (do we want a Deprecated condition mirroring the existing Ignored one?). If the team adds that condition later, this throttling falls out naturally.

I'd dismiss this as a nit until/unless we decide on the broader Deprecated-condition design.

cmd/thv-operator/pkg/vmcpconfig/converter.go
```diff
@@ -174,71 +175,146 @@ func (c *Converter) convertIncomingAuth(
...
-				if authserver.ResolveUpstreamName(up.Name) == resolved {
-					matched = true
-					break
-				}
-			}
-			if !matched {
-				return nil, fmt.Errorf(
-					"authz primaryUpstreamProvider %q does not match any upstream declared "+
-						"on spec.authServerConfig.upstreamProviders", explicit)
+		cedarCfg, err := cedar.ExtractConfig(loaded)
```
jhrozek: Nit / follow-up suggestion — not blocking.

The converter imports `pkg/authz/authorizers/cedar` solely to call `ExtractConfig` here. The runner-path consumer of the same shared loader (`AddAuthzConfigOptions` in `controllerutil/authz.go`) stays authorizer-agnostic and forwards `*authz.Config` opaquely. Adding a future authorizer means editing the converter; the runner path doesn't have that pressure.

Consider extracting a sibling helper in `controllerutil` — e.g. `ExtractCedarAuthzOptions(*authz.Config) (*cedar.ConfigOptions, error)` — so the converter imports only `controllerutil`, not `cedar`. Eliminates the layering smell and keeps Cedar-specific glue in one package.

Also relates to F1: when the spec-over-ConfigMap precedence logic is added to `controllerutil/authz.go`, it'd be a natural fit to land alongside the extracted helper so both code paths share precedence resolution.

jhrozek (commented):
One non-blocking nit from cross-checking the review pass.

jhrozek:
Nit / drift-prevention observation — not blocking, doesn't anchor on a changed line so filing as a PR comment instead of inline.

`ConditionTypeAuthzPrimaryUpstreamProviderIgnored` at `cmd/thv-operator/api/v1beta1/mcpserver_types.go:99` correctly describes its current behavior in the doc comment, but doesn't carry the deprecation lineage. This condition fires only for the *deprecated* `spec.authzConfig.inline.primaryUpstreamProvider` field, which per this PR is slated for removal one release after the deprecation cycle. When that field is removed, this condition (and the constant) should be removed in the same change — but a future maintainer reading just the comment block has no signal to do that cleanup.

Suggested options:

1. Add a single sentence to the doc comment: *"This condition fires only for the deprecated `spec.authzConfig.inline.primaryUpstreamProvider`; it should be removed together with that field at end of the deprecation cycle."*
2. Or rename the constant to carry the deprecation lineage in its name (e.g., `ConditionTypeAuthzDeprecatedInlinePrimaryUpstreamProvider`), which survives even if the comment drifts again.

Filing as a regular PR comment because line 99 is outside the diff hunks of this PR.

cmd/thv-operator/controllers/virtualmcpserver_controller.go
```diff
@@ -600,7 +600,26 @@ func rejectAuthzAdmission(
...
 // embedded AS) because the field is meaningless without an AS, and (b) any
 // name that does not resolve to one of spec.authServerConfig.upstreamProviders.
-func (*VirtualMCPServerReconciler) validateAuthzUpstreamAvailable(
+// emitPrimaryUpstreamProviderDeprecatedEvent emits a Warning event with reason
+// AuthzPrimaryUpstreamProviderDeprecated when the resolved primary upstream
+// provider value came from the deprecated
+// spec.incomingAuth.authzConfig.inline.primaryUpstreamProvider location.
+// Called from every validation branch that observes the explicit provider so
+// the kubectl-visible hint is consistent regardless of the validation outcome.
+func (r *VirtualMCPServerReconciler) emitPrimaryUpstreamProviderDeprecatedEvent(
```
jhrozek, line 609: nit / non-blocking: this `Recorder.Eventf` fires on every reconcile when the deprecated field is set, no `Generation` guard. MCPServer's parallel `ConditionTypeAuthzPrimaryUpstreamProviderIgnored` uses a Condition with `ObservedGeneration` which dedups naturally. Could gate this on `vmcp.Generation != vmcp.Status.ObservedGeneration`, or convert to a Condition for parity with MCPServer. Fine as a follow-up.

cmd/thv-operator/pkg/controllerutil/authz.go
```diff
@@ -166,36 +166,169 @@ func EnsureAuthzConfigMap(
...
+		return nil, fmt.Errorf("inline authz config type specified but inline config is nil")
 	}
-
-	policies := authzRef.Inline.Policies
-	entitiesJSON := authzRef.Inline.EntitiesJSON
-
-	// Create authorization config using the full config structure
-	// This maintains backwards compatibility with the v1.0 schema
 	authzCfg, err := authz.NewConfig(cedar.Config{
 		Version: "v1",
```
jhrozek, line 182: nit / non-blocking: `Version: "v1"` here doesn't match the `"version": "1.0"` written by `EnsureAuthzConfigMap` at L131, and the rest of `pkg/authz/` consistently uses `"1.0"` (e.g. `middleware_test.go:482`, `config_test.go:39`, `authorizers/config_test.go:66`). `authz.Config.Validate()` only checks non-empty today so neither breaks, but if a version enum ever lands the two paths will diverge silently. Worth extracting a constant and aligning on `"1.0"` in a follow-up.

cmd/thv-operator/pkg/controllerutil/authz.go
```diff
@@ -166,36 +166,169 @@ func EnsureAuthzConfigMap(
...
+
+	return &cfg, nil
+}
+
+// ExtractCedarAuthzOptions unwraps the Cedar-specific options embedded in an
+// authz.Config. Returns an error when cfg is nil or is not a Cedar config
+// (e.g. a future HTTP authorizer); callers that need to handle non-Cedar
+// configs gracefully should treat the error as "not Cedar, pass through".
+//
+// This wrapper exists so callers outside pkg/authz can avoid importing
```
jhrozek, line 284: nit / non-blocking: this comment claims the wrapper localises the cedar dependency for "callers outside `pkg/authz`", but the return type is `*cedar.ConfigOptions` so any caller that uses the value still has to import `pkg/authz/authorizers/cedar`. The reason `vmcpconfig` stays free of the import is that it copies fields into its own struct, not anything this wrapper does. Either narrow the comment (e.g., "so `vmcpconfig` can read fields without holding a typed value") or change the return type to a neutral struct.

jhrozek (approved):
LGTM. Left three non-blocking nits inline — happy for them to be addressed in a follow-up commit/PR.

Issue #4929, opened 2026-04-18
vmcp/session: expose a pluggable backendConnector on NewSessionFactory


JAORMX:
## Problem

`session.NewSessionFactory(registry, opts...)` at `pkg/vmcp/session/factory.go:187` is the only public constructor for `MultiSessionFactory`. It hardcodes `backend.NewHTTPConnector(registry)` as the connector.

The internal plumbing for a custom connector exists:
- `backendConnector` type at `factory.go:117` (unexported)
- `newSessionFactoryWithConnector` at `factory.go:194` (unexported, appears test-only)
- `backend.Session` + `backend.NewHTTPConnector` live under `pkg/vmcp/session/internal/backend/` — Go's `internal` path rule prevents external use.

## Use case

Embedders whose gateway already resolves per-backend capability lists (for conflict resolution, ACL filtering, static catalog composition) need the session factory to present those pre-resolved capabilities instead of re-fetching them over HTTP at every session `initialize`. Similarly, embedders with stateless dispatch models do not want the factory to open N live MCP client connections per session.

## Proposed API

Export the `backendConnector` interface (or lift it to `pkg/vmcp/session/types` next to `MultiSession`), and add a functional option:

```go
func WithBackendConnector(conn BackendConnector) MultiSessionFactoryOption
```

When supplied, `NewSessionFactory` uses it instead of `backend.NewHTTPConnector`. The default stays unchanged.

Separately, expose `backend.Session` (or a narrower interface specifically for connector implementations) via `pkg/vmcp/session/types` so external implementations have a target type to return.

## Scope

- Additive. No change to the existing HTTP-connector default.
- May require widening `backend.Session` to live outside `internal/` — suggest moving just the interface, not the HTTP implementation.

PR #5301, opened 2026-05-16
Wire HeaderForward into vMCP per-session HTTP client

### LINKED ISSUE #5289: MCPServerEntry.headerForward.addPlaintextHeaders is silently ignored — never reaches upstream
## Summary

`MCPServerEntry.spec.headerForward.addPlaintextHeaders` is silently ignored. Headers configured on a remote-URL `MCPServerEntry` (proxied through a `VirtualMCPServer`) never reach the upstream MCP server. The field is defined in the CRD and accepted by the operator, but is dropped during the conversion from CR to runtime `vmcp.Backend` and has no representation anywhere in the vMCP outgoing HTTP client.

Verified in `v0.27.1`.

## Reproduction

`MCPServerEntry` configured against GitHub's Copilot MCP at `api.githubcopilot.com/mcp/`, with the `X-MCP-Toolsets` header configured to select a specific set of toolsets:

```yaml
kind: MCPServerEntry
metadata:
  name: github-copilot-projects
spec:
  remoteUrl: "https://api.githubcopilot.com/mcp/"
  transport: streamable-http
  externalAuthConfigRef:
    name: github-oauth-upstream-inject
  headerForward:
    addPlaintextHeaders:
      X-MCP-Toolsets: "projects,issues,pull_requests,users,repos"
```

**Direct call** to `https://api.githubcopilot.com/mcp/` with the same token + same `X-MCP-Toolsets` header returns 39 tools strictly filtered to the 5 requested toolsets, including `projects_get`, `projects_list`, `projects_write`.

**Through the vMCP**, the upstream returns 43 tools matching GitHub's *default* toolset for the Copilot endpoint — which includes `copilot`, `orgs`, `secret_protection` tools (NOT requested) and is missing `projects_*` (which IS requested). The fingerprint matches "header not sent → GitHub returns defaults."

## Root cause

Source trace in `v0.27.1`:

- **CRD field exists**: `cmd/thv-operator/api/v1beta1/mcpserverentry_types.go:38-41` defines `HeaderForward *HeaderForwardConfig`; `cmd/thv-operator/api/v1beta1/mcpremoteproxy_types.go:12-16` defines `HeaderForwardConfig.AddPlaintextHeaders`.

- **Reconciler doesn't read it**: `pkg/vmcp/workloads/k8s.go:503-586` (`mcpServerEntryToBackend`) reads `RemoteURL`, `Transport`, `CABundleRef`, `ExternalAuthConfigRef` — but **never reads `entry.Spec.HeaderForward`**.

- **No field to carry it**: `pkg/vmcp/types.go:290-333` `vmcp.Backend` has no field for static headers.

- **No middleware to apply it**: `pkg/vmcp/client/client.go:265-311` shows the vMCP outgoing HTTP client's transport chain is `size limiter → trace propagation → identity propagation → auth → HTTP`. There is no static-header layer.

- **`MCPRemoteProxy` works** via a separate code path: `cmd/thv-operator/controllers/mcpremoteproxy_runconfig.go:270-297` calls `runner.WithHeaderForward(proxy.Spec.HeaderForward.AddPlaintextHeaders)`, which flows through `pkg/runner/middleware.go:301-310` into `pkg/transport/middleware/header_forward.go:100-137`. `MCPServerEntry` bypasses RunConfig entirely and goes straight to the vMCP `Backend` struct, so it never picks up this middleware.

## Suggested fix

1. Add a `HeaderForward map[string]string` field (or equivalent) to `vmcp.Backend` in `pkg/vmcp/types.go`.
2. Populate it from `entry.Spec.HeaderForward.AddPlaintextHeaders` in `mcpServerEntryToBackend` (`pkg/vmcp/workloads/k8s.go`).
3. Add a static-header middleware layer in the vMCP HTTP client transport chain (`pkg/vmcp/client/client.go`) that reads the per-backend headers and injects them into outgoing requests.
4. Consider doing the same for `addHeadersFromSecret` so secret-sourced headers are also forwarded.

## Impact

Any vMCP backend that depends on header-based configuration of the upstream (e.g. GitHub's `X-MCP-Toolsets`, API keys passed via custom headers, correlation IDs) silently does nothing. The field accepting valid input without applying it is the worst form — there is no validation error, no warning in logs, nothing in `kubectl describe` to indicate the headers aren't being sent.

lorr1:
## Summary

- **Why:** PR #5239 wired `HeaderForward` into the vMCP startup capability-discovery HTTP client at `pkg/vmcp/client/client.go:309`, but the **per-session MCP HTTP client** at `pkg/vmcp/session/internal/backend/mcp_session.go` builds a parallel transport chain (`DefaultTransport → auth → identity`) and never reads `target.HeaderForward`. Every post-initialize MCP call (`tools/list`, `tools/call`, …) reaches the upstream without user-configured headers. Reported in #5289 and reproduced on staging in v0.27.2 — GitHub Copilot's `X-MCP-Toolsets` filter is silently ignored, so the upstream returns the default 43-tool set instead of the configured subset.
- **What:** Export the existing `BuildHeaderForwardTripper` helper from `pkg/vmcp/client`, construct a single `secrets.EnvironmentProvider` at `NewHTTPConnector` build time, plumb it through `createMCPClient`, and wrap the shared transport chain with the helper as the outermost stage. The ordering preserves vMCP's authoritative auth/identity headers because `headerForwardRoundTripper.RoundTrip` skips header names already set by inner stages.

Fixes #5289

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Unit tests (`task test`) — new `TestHTTPSession_AppliesHeaderForwardToPostInitializeRequests` in `pkg/vmcp/session/internal/backend/`: fakes an MCP server, drives `initialize` + `tools/call`, asserts the post-initialize request carries `X-MCP-Toolsets`. Confirmed fails on the pre-fix code (header is empty on the wire) and passes after the fix. All other tests in `pkg/vmcp/...` still pass. Second commit adds overlap-precedence and `AddHeadersFromSecret` subtests using a shared `fakeBackend`, plus a restricted-name rejection subtest. All pass.
- [ ] Manual testing — original symptom reproduced on staging (vmcp `0.1.36-main.46.c905a7cc`, toolhive submodule `97b0cc3f` = v0.27.2): `MCPServerEntry/github-copilot-projects` with `X-MCP-Toolsets: projects,issues,pull_requests,users,repos` got the default 43-tool set instead of the filtered ~39-tool set. Fix verified locally via the regression test; staging re-run pending image cut.

## API Compatibility

No CRD changes. The only exported-API delta is renaming `pkg/vmcp/client.buildHeaderForwardTripper` → `BuildHeaderForwardTripper` — there was no prior exported name to break.

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/client/header_forward.go` | Export `BuildHeaderForwardTripper`; doc comment notes both callers. Comment on `headerForwardRoundTripper.RoundTrip` corrected to describe actual mechanism (outermost runs first; auth/identity win via unconditional `Set()`, not via the "skip if already set" guard). |
| `pkg/vmcp/client/client.go` | Update the one in-package call site to the exported name. |
| `pkg/vmcp/client/header_forward_test.go` | Rename-only test fixup so the package compiles. |
| `pkg/vmcp/session/internal/backend/mcp_session.go` | Construct `secrets.NewEnvironmentProvider()` at connector build time, plumb `ctx` + `provider` into `createMCPClient`, wrap chain with `BuildHeaderForwardTripper` as outermost stage. Chain-ordering comment rewritten to describe actual mechanism (outermost runs first; auth/identity win via unconditional `Set()`, not via the "skip if already set" guard). |
| `pkg/vmcp/session/internal/backend/mcp_session_test.go` | Update one call site for the new `createMCPClient` signature. |
| `pkg/vmcp/session/internal/backend/mcp_session_header_forward_test.go` | New end-to-end regression test for post-initialize header forwarding. Second commit: replace standalone `headerCapturingBackend` with shared `fakeBackend`; add overlap-precedence, restricted-name, and secret-source subtests; trim diff-narrating comments. |
| `pkg/vmcp/session/internal/backend/mcp_session_capabilities_test.go` | Extend `fakeBackend` with per-method header recording + `tools/call` handler. Backward-compatible. |

Yes. `MCPServerEntry.spec.headerForward.addPlaintextHeaders` (and `addHeadersFromSecret`) now reach the upstream MCP server on every per-session call, not just on startup capability discovery. Users relying on header-based upstream config (e.g. GitHub Copilot's `X-MCP-Toolsets`, custom API keys, correlation IDs) will see them applied to live MCP traffic for the first time.

## Implementation plan

<details>
<summary>Approved implementation plan (Option A — minimal change)</summary>

Three options were considered:

- **A — Construct `secrets.Provider` inline in `NewHTTPConnector`** (chosen). ~10 LOC, mirrors `httpBackendClient`, no public-API growth on `NewSessionFactory`/`NewHTTPConnector`. Needed a small bump in scope to *export* `BuildHeaderForwardTripper` (originally package-private) so the session backend can call it — ~3 extra trivial lines.
- B — Thread `secrets.Provider` through `NewSessionFactory` and `NewHTTPConnector` signatures (~25 LOC, signature growth on two public funcs, touches `serve.go`).
- C — `WithSecretsProvider` functional option (~25 LOC, single-consumer config knob — anti-pattern #8).

**Selected A** for the smallest diff and the cleanest match to the existing `httpBackendClient` pattern. Test surface defined first (red-phase test in `pkg/vmcp/session/internal/backend/`), then production fix landed against it. Transport-chain ordering: header-forward wraps *outside* `identityRoundTripper`, matching `client.go:309` and preserving the "inner stages win on overlapping header names" invariant from `headerForwardRoundTripper.RoundTrip`.

</details>

## Follow-ups

- **Move helpers to `pkg/vmcp/headerforward/`** — opened as parallel building-block PR #5302 (draft). Mechanical relocation of `BuildHeaderForwardTripper` and its support types out of `pkg/vmcp/client/` (one of the two consumers) into a neutral home. Independent of this PR; whichever merges first, the other rebases with a one-line import update.
- **SSE backend test fake + table-driven regression** — filed as #5303. The header-forward wrap is shared between `streamable-http` and `sse` branches in `createMCPClient`, so SSE is wired correctly today, but a future refactor could silently break it. The test fake work to lock that down is substantial enough to defer.
- **Provider DI at the composition root** — overlaps with the design surface in #4929 (pluggable backendConnector). Deferred until #4929's plumbing settles, since any DI shape we adopt now would have to be reshaped to fit.
- **`createMCPClient` extraction (`buildSharedBackendTransport`)** — marker only. Worth doing if a 3rd shared-chain dependency (e.g. trace propagation) lands; not needed today.
- **Anti-pattern #8 consolidation** — `pkg/vmcp/client/client.go::defaultClientFactory` and `pkg/vmcp/session/internal/backend/mcp_session.go::createMCPClient` are still two parallel HTTP-client builders. This PR + #5302 reduce the duplication; full consolidation is its own piece of work.

- **Anti-pattern #8 surface.** Two parallel HTTP-client builders in vMCP (`pkg/vmcp/client/client.go` for startup discovery, `pkg/vmcp/session/internal/backend/mcp_session.go` for per-session traffic) caused this bug — the first fix only landed in one. Unifying them is out of scope here; PR #5302 is the next mechanical step (helper relocation), and full consolidation is tracked as a follow-up above.
- **Provider DI deferred to the #4929 surface.** A "single `secrets.Provider` constructed at the composition root and injected through `NewSessionFactory`/`NewHTTPConnector`" cleanup overlaps directly with #4929's pluggable `backendConnector` work. Doing provider DI here would force a reshape once #4929 lands, so it is intentionally deferred.
- **Two `secrets.EnvironmentProvider` instances now exist** (one per builder). They're stateless — no shared state, no contention — but if we ever want a single shared provider, that refactor is orthogonal to this fix and folds into the #4929 follow-up above.
- **Test boundary.** The original regression test exercises `AddPlaintextHeaders` only. The second commit adds an explicit `AddHeadersFromSecret` subtest (env-resolved secret value reaches the wire), so both header sources are now covered.

### DIFF
--- a/pkg/vmcp/headerforward/transport.go
+++ b/pkg/vmcp/headerforward/transport.go
@@ -35,9 +35,13 @@ type headerForwardRoundTripper struct {
 }
 
 // RoundTrip injects the pre-resolved headers onto a clone of the request and
-// delegates to the wrapped transport. Headers already present on the request
-// are left untouched so inner transports (auth, identity, trace) can always
-// override user-supplied values for the same name.
+// delegates to the wrapped transport. Names already present on the inbound
+// request are skipped — this preserves any header the *caller* set on the
+// request before invoking the round-tripper chain. Inner (closer to the wire)
+// transports run AFTER this one and use Set() unconditionally, so on any
+// overlapping name those stages still win on the wire. Restricted names are
+// blocked at resolve time, so user-supplied config cannot reach this point
+// for Host, hop-by-hop, or X-Forwarded-* anyway.
 func (h *headerForwardRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	if len(h.headers) == 0 {
 		return h.base.RoundTrip(req)
--- a/pkg/vmcp/headerforward/transport.go
+++ b/pkg/vmcp/headerforward/transport.go
@@ -64,6 +64,11 @@ func (h *headerForwardRoundTripper) RoundTrip(req *http.Request) (*http.Response
 // backend's pre-resolved HeaderForwardConfig. Returns base unchanged when no
 // header injection is configured or the effective header set is empty.
 //
+// Used by both the vMCP backend client (startup capability discovery) and the
+// per-session backend connector (long-lived MCP traffic). Exported so the
+// session backend in pkg/vmcp/session/internal/backend can share the same
+// transport-chain wiring.
+//
 // Fails loudly (constructor validation, per go-style.md) when a secret identifier
 // cannot be resolved through the provider, so a misconfigured backend surfaces
 // at pod startup — not as a silent missing-header on every request.
--- a/pkg/vmcp/session/internal/backend/mcp_session.go
+++ b/pkg/vmcp/session/internal/backend/mcp_session.go
@@ -17,11 +17,13 @@ import (
 	"github.com/mark3labs/mcp-go/mcp"
 
 	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/secrets"
 	"github.com/stacklok/toolhive/pkg/versions"
 	"github.com/stacklok/toolhive/pkg/vmcp"
 	vmcpauth "github.com/stacklok/toolhive/pkg/vmcp/auth"
 	authtypes "github.com/stacklok/toolhive/pkg/vmcp/auth/types"
 	"github.com/stacklok/toolhive/pkg/vmcp/conversion"
+	"github.com/stacklok/toolhive/pkg/vmcp/headerforward"
 )
 
 const (
@@ -196,19 +198,25 @@ func (c *mcpSession) GetPrompt(
 //
 // registry provides the authentication strategy for outgoing backend requests.
 // Pass a registry configured with the "unauthenticated" strategy to disable auth.
+//
+// A single secrets.EnvironmentProvider is constructed once per connector and
+// shared across every session it creates; its lifetime matches the connector's.
+// It is consumed by BuildHeaderForwardTripper to resolve secret-backed entries
+// in target.HeaderForward.
 func NewHTTPConnector(registry vmcpauth.OutgoingAuthRegistry) func(
 	ctx context.Context,
 	target *vmcp.BackendTarget,
 	identity *auth.Identity,
 	sessionHint string,
 ) (Session, *vmcp.CapabilityList, error) {
+	provider := secrets.NewEnvironmentProvider()
 	return func(
 		ctx context.Context,
 		target *vmcp.BackendTarget,
 		identity *auth.Identity,
 		sessionHint string,
 	) (Session, *vmcp.CapabilityList, error) {
-		c, err := createMCPClient(target, identity, registry, sessionHint)
+		c, err := createMCPClient(ctx, target, identity, registry, sessionHint, provider)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to create MCP client for backend %s: %w", target.WorkloadID, err)
 		}
@@ -238,11 +246,17 @@ func NewHTTPConnector(registry vmcpauth.OutgoingAuthRegistry) func(
 // to client.Close(), not to any caller-supplied init context.
 // sessionHint, when non-empty, is passed as the initial Mcp-Session-Id for
 // streamable-HTTP transports so the backend can resume an existing session.
+//
+// ctx is used only to resolve secret-backed entries in target.HeaderForward at
+// client-creation time; the transport itself is started with context.Background()
+// as described above. provider supplies values for those secret-backed headers.
 func createMCPClient(
+	ctx context.Context,
 	target *vmcp.BackendTarget,
 	identity *auth.Identity,
 	registry vmcpauth.OutgoingAuthRegistry,
 	sessionHint string,
+	provider secrets.Provider,
 ) (*mcpclient.Client, error) {
 	// Resolve and validate the auth strategy once at client creation time.
 	strategyName := authtypes.StrategyTypeUnauthenticated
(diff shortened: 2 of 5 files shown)


tgrunnagle (dismissed):
LGTM!

tgrunnagle (approved):

Issue #5292, opened 2026-05-15
feat(proxy): validate JSON-RPC frames in SSE event stream


amirejaz:
## Background

PR #5288 adds upstream JSON-RPC response validation for the streamable-HTTP transport. SSE is explicitly out of scope because the fix architecture (buffer full body → validate → rewrite or pass through) doesn't work for a streaming event format.

## The gap

When an SSE-based MCP upstream sends a structurally invalid JSON-RPC frame (`data:` line without `jsonrpc:"2.0"`, wrong `id` type, etc.), the transparent proxy forwards it unchanged. Clients receive the malformed event and may crash, misparse it, or treat it as a legitimate response.

## Why it's harder than streamable-HTTP

For streamable-HTTP, `ProcessResponse` buffers the entire HTTP response body before any bytes reach the client, so it can atomically reject and rewrite to 502.

For SSE:
- The `200 OK` and `Content-Type: text/event-stream` headers are committed before any events arrive. There is no way to rewrite the status code for a malformed event mid-stream.
- A fix needs a per-event streaming interceptor that wraps the SSE body reader, parses each `data:` line as it arrives, validates the JSON-RPC content, and either synthesizes an error event or closes the stream cleanly on the first invalid frame.
- "Close the stream" is the strongest guarantee available — unlike streamable-HTTP, the proxy cannot return a structured 502 for an individual bad event.

## Proposed scope

A follow-up to #5288 scoped to the SSE response processor:
- Wrap the SSE body reader with a per-event interceptor
- On an invalid `data:` JSON-RPC frame: write a synthetic `data: {"jsonrpc":"2.0","id":null,"error":{"code":-32000,"message":"Invalid upstream JSON-RPC frame"}}\n\n` event and close the stream
- Leave non-`data:` SSE lines (comments, `event:`, `retry:`) untouched
- The guarantee is weaker than streamable-HTTP (client receives the error event, not a 502) — document this explicitly

## Context

Found via DAST adversarial testing (`checkMalformedFraming` in the enterprise DAST suite). The test currently documents this as a known finding for the SSE transport.

PR #5304, opened 2026-05-17
Stop SSE filter from leaking tools/list on undecodable lines

### LINKED ISSUE #5257: authz: SSE response filter leaks unfiltered tools/list on undecodable / non-Response data lines (bypasses cedar)
## Summary

`ResponseFilteringWriter.processSSEResponse` in `pkg/authz/response_filter.go` writes the **entire raw upstream response unfiltered** whenever any `data:` line in the SSE stream either fails `jsonrpc2.DecodeMessage` or decodes to a message that is not a `*jsonrpc2.Response` (for example, an MCP notification interleaved on the stream).

When this happens on a `tools/list` reply, the client receives the full unfiltered tool catalog — bypassing the cedar authorization filter that should have removed tools the caller can't invoke.

The same code path emits a visible side-effect: `http: superfluous response.WriteHeader call from … response_filter.go:191`. That warning is how we noticed the bug.

Confirmed against `v0.27.0` (commit `182a025`).

## Impact

**Authorization bypass on `tools/list`.** A caller whose cedar policy denies every action on a backend can still receive that backend's full tool list, as long as the backend's SSE reply contains a `data:` line that doesn't parse as `*jsonrpc2.Response`. The bypass is silent — no `cedar decision` log lines are produced for that response, because the per-tool evaluation never runs.

Observed on `proxyrunner:v0.27.0` with two `MCPServer`s configured identically — same `authzConfig` (cedar policy denies all actions for the caller), same middleware chain (`auth → tokenexchange → mcp-parser → usagemetrics → authorization → audit → recovery`), same incoming `transport_type: streamable-http`. They differ only in the upstream backend image:

- **`ghcr.io/stackloklabs/mkp/server:0.2.4`** (native `streamable-http`, JSON responses): pod logs `cedar decision: deny` per tool, response correctly filtered to an empty `tools` array. Expected.
- **`ghcr.io/gorizond/mcp-rancher-multi:0.1.16`** (fronted by the npm `mcp-proxy` stdio↔SSE adapter, `text/event-stream` responses): **zero** `cedar decision` lines in pod logs, ~700 `INFO http: superfluous response.WriteHeader call from … response_filter.go:191` warnings (one per `tools/list`), client receives the full unfiltered tool list. Bug.

The differentiating factor is the **upstream MCP server's response Content-Type**, which selects between `processJSONResponse` (works) and `processSSEResponse` (broken). The bug lives entirely on the SSE side.

## Reproduction

1. Deploy two `MCPServer`s with identical `authzConfig` containing a cedar policy that denies all actions for a given caller, e.g.

   ```cedar
   permit(principal, action, resource) when { context has claim_groups && context.claim_groups.contains(\"/never-matches\") };
   ```

2. Pick backends that produce different response framings:
   - **Backend A** (filter works): `ghcr.io/stackloklabs/mkp/server:0.2.4`. Responds with `application/json`.
   - **Backend B** (filter bypassed): `ghcr.io/gorizond/mcp-rancher-multi:0.1.16`. Fronted by the npm `mcp-proxy` package as a stdio→SSE bridge; responds with `text/event-stream`.

3. Call `tools/list` on both with a token that cedar denies.
4. Backend A: filtered response (empty `tools`), pod logs `cedar decision: deny` per tool. ✅
5. Backend B: full unfiltered `tools` list returned, pod logs one `http: superfluous response.WriteHeader call from … response_filter.go:191` per call. ❌

Both pods run `proxyrunner:v0.27.0` with `transport_type: streamable-http` on the toolhive side.

## Root cause

`pkg/authz/response_filter.go`, `processSSEResponse`, lines ~188–201 (v0.27.0):

```go
if data, ok := bytes.CutPrefix(line, []byte(\"data:\")); ok {
    message, err := jsonrpc2.DecodeMessage(data)
    if err != nil {
        rfw.ResponseWriter.WriteHeader(rfw.statusCode)            // line 191
        _, err := rfw.ResponseWriter.Write(rawResponse)            // writes ENTIRE raw payload
        return err
    }

    response, ok := message.(*jsonrpc2.Response)
    if !ok {
        rfw.ResponseWriter.WriteHeader(rfw.statusCode)            // line 198
        _, err := rfw.ResponseWriter.Write(rawResponse)            // writes ENTIRE raw payload
        return err
    }
    ...
}
```

Two problems:

1. **Per-line failure aborts filtering of the whole stream.** When any single `data:` line fails to decode as `*jsonrpc2.Response` — either because of a decode error (line 191) or because the decoded message is a non-Response (e.g. a notification, line 198) — the function writes `rawResponse` (the **entire** buffered SSE payload) and returns. The apparent intent is \"fall back to pass-through for this line,\" but the implementation passes through the whole response, including any subsequent `data:` lines containing the real `tools/list` payload. A single unparseable or non-Response line defeats filtering for the entire reply.

   Notifications interleaved on the response stream are explicitly permitted by the MCP spec (progress updates, log messages, `notifications/message`, etc.). So even a fully spec-compliant upstream MCP server using SSE would trip the line-198 branch and silently bypass cedar.

2. **Double `WriteHeader`.** The fallback explicitly calls `WriteHeader(rfw.statusCode)`. The wrapped writer may already have emitted headers via `Flush()` (lines 125–130) before `FlushAndFilter` runs, particularly when `httputil.ReverseProxy` uses `FlushInterval: -1`. The net effect is the `superfluous WriteHeader` warning and undefined behavior around when/if the filtered output actually reaches the wire.

## Suggested fix direction

- In `processSSEResponse`, treat undecodable or non-`Response` `data:` lines as pass-through **for that line only**: write the original line and the line separator, then `continue` the loop. Do not write `rawResponse` and do not `return`.
- Audit the same pattern in `processJSONResponse` (lines 132–145) — there it's the whole response by definition, so probably already correct, but worth a second look.
- Consider WARN-logging when filtering is skipped on a `tools/list` response so future bypasses surface in audit logs rather than going silent.

A regression test covering an SSE stream that mixes a non-Response `data:` line (e.g. a `notifications/message`) with a real `tools/list` response would catch this. The current `pkg/authz/response_filter_test.go` does not appear to exercise that path.

## Versions

- toolhive `proxyrunner`: **v0.27.0**
- Toolhive incoming transport: `streamable-http`
- Backend images used in repro:
  - `ghcr.io/stackloklabs/mkp/server:0.2.4` (filter works — `application/json` responses)
  - `ghcr.io/gorizond/mcp-rancher-multi:0.1.16` (filter bypassed — `text/event-stream` responses via the npm `mcp-proxy` bridge)

saivedant169:
## Summary

`ResponseFilteringWriter.processSSEResponse` used to abort filtering of the entire SSE stream whenever a single `data:` line failed `jsonrpc2.DecodeMessage` or decoded to a non-`*jsonrpc2.Response` message. The fallback wrote the **entire** buffered upstream payload and returned, so any subsequent `data:` line containing the real `tools/list` reply reached the client unfiltered, bypassing the cedar authorization filter. The fallback also re-called `WriteHeader` after the reverse proxy's first `Flush()` had already emitted headers, producing the `http: superfluous response.WriteHeader call from … response_filter.go:191` warning that surfaced the bug.

This PR:

- Treats an undecodable or non-`Response` `data:` line as pass-through **for that line only**: fall through to the existing line writer (the loop's `if !written` branch) rather than dumping `rawResponse` and returning.
- Drops the explicit `WriteHeader(rfw.statusCode)` calls in the fallback branches, which removes the double-header warning at line 191.
- Emits a WARN log when a `tools/list` reply contains a `data:` line that bypasses the filter, so future bypasses are visible in audit logs instead of silent.

Notifications interleaved on a response stream are explicitly allowed by the MCP spec, so this case can be hit by fully spec-compliant upstreams (`notifications/message`, progress updates, etc.) as well as by upstreams fronted by an SSE bridge (the npm `mcp-proxy` case from the issue reproduction).

Closes #5257

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Added `TestResponseFilteringWriter_SSE_PerLineFallthrough` as a table-driven regression test covering both branches (decode error and non-`Response` decode). It builds an SSE stream that interleaves a notification (or an undecodable garbage line) with a real `tools/list` response, then asserts:

1. The preceding line is preserved verbatim on the wire.
2. The `tools/list` response is filtered to only authorized tools.
3. The literal string `"admin_tool"` (the unauthorized tool from the unfiltered upstream payload) does not appear anywhere in the output.

The test fails on the prior code with `admin_tool` reaching the recorder, and passes on the patched code.

Verified on `pkg/authz/...` with `-race`:
- `go test -ldflags=-extldflags=-Wl,-w -race ./pkg/authz/...` passes (including the existing JSON-path tests and the new SSE test).
- `golangci-lint run ./pkg/authz/...` reports 0 issues.

Yes. On SSE upstreams (`Content-Type: text/event-stream`), `tools/list`, `prompts/list`, and `resources/list` responses now respect the configured cedar authorization filter even when the upstream interleaves notifications or sends an undecodable `data:` line. Previously such streams returned the unfiltered tool/prompt/resource catalog to the caller. Filtered behavior is what operators already get on `application/json` upstreams; this brings SSE in line.

The change is intentionally minimal. I considered also auditing `processJSONResponse` for the same shape (`return rawResponse on per-line failure`) but it processes a single message, not a stream, so the same code shape there is already correct. Happy to extend if you want a defensive log added on that side too.


jhrozek:
This PR looks like it's targetting issue #5292 ?

jhrozek:
cc @amirejaz

amirejaz:
Not quite — #5304 closes #5257, which is an authorization bypass in the Cedar SSE response filter (`pkg/authz/response_filter.go`): when a `data:` line fails to decode as a `*jsonrpc2.Response` (e.g. an interleaved notification), the filter was dumping the entire raw unfiltered upstream payload, bypassing Cedar entirely.

Our issue #5292 is a separate concern in a different layer: per-event JSON-RPC frame validation in the transparent proxy (`pkg/transport/proxy/transparent/`), analogous to what #5288 did for streamable-HTTP. #5292 is about the proxy rejecting/closing-stream on structurally malformed frames, not about the Cedar filter.

The two are complementary — #5304 fixes a more serious auth bypass, #5292 is about frame validation in the non-auth transparent proxy path.

pkg/authz/response_filter.go
```diff
@@ -187,35 +187,42 @@ func (rfw *ResponseFilteringWriter) processSSEResponse(rawResponse []byte) error
...
-
-			filteredData, err := jsonrpc2.EncodeMessage(filteredResponse)
-			if err != nil {
-				return rfw.writeErrorResponse(response.ID, err)
+			switch {
+			case err != nil:
+				// Pass this line through unfiltered. Earlier revisions wrote
+				// rawResponse and returned here, which leaked every subsequent
+				// data line on the stream past the filter (issue #5257).
+				if rfw.method == string(mcp.MethodToolsList) {
```
yrobla: **Warning: WARN logs are only emitted for `tools/list`, but the bypass applies to all four filtered methods.**

Both log conditions — `err != nil` branch here, and the `else if` for non-`*jsonrpc2.Response` in the `default:` branch below — are gated on `rfw.method == string(mcp.MethodToolsList)`. However, `requiresResponseFiltering` covers four methods: `tools/list`, `prompts/list`, `resources/list`, and `find_tool`. An SSE backend serving `prompts/list` or `resources/list` with interleaved MCP notifications would silently hit the same passthrough path — no log line at all.

The security fix itself is correct for all methods (the `written` flag controls passthrough regardless of method), but the observability goal from the issue — _"future bypasses surface in audit logs rather than going silent"_ — is only fulfilled for `tools/list`.

Consider replacing both method checks with `requiresResponseFiltering(rfw.method)`, which already encodes the right set.

pkg/authz/response_filter_test.go
```diff
@@ -863,3 +863,126 @@ func TestOptimizerPassThroughToolsInResponseFilter(t *testing.T) {
 			"admin_tool has no permit policy and is not a pass-through tool")
 	})
 }
+
+// TestResponseFilteringWriter_SSE_PerLineFallthrough is a regression test for
+// issue #5257: when an SSE upstream interleaves a non-Response data line (e.g.
+// an MCP notification) or an undecodable data line with a real tools/list
+// response, the filter previously wrote the entire raw upstream payload and
+// returned, leaking the unfiltered tools/list past Cedar. It must instead pass
+// only the offending line through and continue filtering the rest of the
```
yrobla: **Warning: Regression test only covers `MethodToolsList`; the same bypass applies to `prompts/list` and `resources/list` over SSE.**

The new test hardcodes `string(mcp.MethodToolsList)`. The same code path in `processSSEResponse` runs for any method where `requiresResponseFiltering` is true — including `MethodPromptsList` and `MethodResourcesList`. A notification-interleaved SSE response on those paths is not currently exercised.

A brief additional sub-test for `MethodPromptsList` or `MethodResourcesList` with an interleaved notification would confirm parity and guard against a future regression on those paths.

yrobla (commented):
The direction is correct and the core fix is sound — removing the whole-payload dump and letting the existing `if !written` branch handle per-line passthrough is exactly right. Two warnings below about logging coverage and test coverage.

saivedant169:
Both points addressed in 3984262.

The WARN logs no longer gate on tools/list, so any method that hits processSSEResponse via requiresResponseFiltering (tools/list, prompts/list, resources/list, find_tool) now WARNs when a data line is undecodable or non-Response.

The regression test is now a method-by-preceding-line table over all three list methods. Each method has its own cedar policy and a one-authorized + one-unauthorized fixture; both the notification and undecodable-line cases run against tools/list, prompts/list, and resources/list. Confirmed the unauthorized entry never reaches the recorder for any of the six combinations.

@yrobla ready for another look when you have a moment.

pkg/authz/response_filter.go
```diff
@@ -187,35 +187,44 @@ func (rfw *ResponseFilteringWriter) processSSEResponse(rawResponse []byte) error
...
+						return rfw.writeErrorResponse(response.ID, err)
+					}
+
+					_, err = rfw.ResponseWriter.Write([]byte("data: " + string(filteredData) + "\n"))
+					if err != nil {
+						return fmt.Errorf("%w: %w", errBug, err)
+					}
+
+					written = true
+				} else {
```
JAORMX: The pass-through gate here is still keyed on the JSON-RPC message *type*, which the upstream controls. `DecodeMessage` only classifies a frame as `*Response` when it has no `method` field. So a frame carrying both `method` (e.g. `notifications/initialized`) and `result` (with the real tools list) decodes to a `*Request`, fails the `*Response` assertion, and passes through unfiltered via the `if !written` branch below.

Same bypass class as #5257, just narrowed from "any non-Response line" to "a deliberately framed one." Since the proxy's threat model treats upstream list responses as untrusted (that's the whole reason `requiresResponseFiltering` exists), I think this residual is worth closing.

Cheapest fix: for filtered methods, if the raw `data` contains a `result` field, route it through `filterListResponse` regardless of message type — or drop the line as a protocol violation. A regression test with a frame like `{"jsonrpc":"2.0","method":"notifications/initialized","id":1,"result":{"tools":[...]}}` would lock it.

Happy to file this as a follow-up if you'd rather ship the accidental-bypass fix now and track it separately — your call.

pkg/authz/response_filter.go
```diff
@@ -187,35 +187,44 @@ func (rfw *ResponseFilteringWriter) processSSEResponse(rawResponse []byte) error
...
+					if err != nil {
+						return rfw.writeErrorResponse(response.ID, err)
+					}
+
+					filteredData, err := jsonrpc2.EncodeMessage(filteredResponse)
+					if err != nil {
+						return rfw.writeErrorResponse(response.ID, err)
+					}
+
+					_, err = rfw.ResponseWriter.Write([]byte("data: " + string(filteredData) + "\n"))
```
JAORMX: This hardcodes `"\n"`, but the loop already writes the detected `linesep` at line 228. On a `\r\n` stream that produces `data: <json>\n\r\n` — a stray `\n` a strict SSE parser might read as an empty event. The passthrough branch writes `line + linesep` (no extra `\n`), so the two paths produce structurally different output frames.

Drop the `"\n"` here and let `linesep` terminate. One-line change.

pkg/authz/response_filter.go
```diff
@@ -187,35 +187,44 @@ func (rfw *ResponseFilteringWriter) processSSEResponse(rawResponse []byte) error
...
-
-			filteredResponse, err := rfw.filterListResponse(response)
-			if err != nil {
-				return rfw.writeErrorResponse(response.ID, err)
-			}
-
-			filteredData, err := jsonrpc2.EncodeMessage(filteredResponse)
-			if err != nil {
-				return rfw.writeErrorResponse(response.ID, err)
+			switch {
```
JAORMX: The `switch { default: if/else }` is harder to read than a plain `if / else if / else` chain here. There's exactly one condition (`err != nil`) plus a type assertion — the `switch` adds no dispatch value, and the happy path (decode ok, it's a Response, filter it) is buried in a `default` with a nested `if`, three levels deep.

An `if err != nil { warn } else if response, ok := ...; ok { filter } else { warn }` chain keeps the happy path at one indentation level and makes the three outcomes symmetric. No behavior change, just clarity.

JAORMX (commented):
Core fix is correct — per-line fallthrough closes the accidental bypass from #5257, and the regression test locking all three list methods is solid. Three points below before we merge.

saivedant169:
Pushed `eaa6c0d`.

@JAORMX:
- Disguised result frame: added `carriesResult`, so a non-Response frame that still carries a `result` is dropped as a protocol violation instead of passing through. Regression test with the `notifications/initialized` + `result` frame, confirmed it leaks without the drop. Fixed here rather than as a follow-up.
- Line termination: dropped the hardcoded `\n` on the filtered write; the loop's `linesep` terminates now, so `\r\n` streams stay consistent.
- Readability: flattened the decode `switch` into the `if / else if / else` chain.

@yrobla: your two points are already in on the current branch, the WARN logs are unconditional (so every filtered method logs), and the SSE fallthrough test exercises tools/list, prompts/list, and resources/list.

pkg/authz/response_filter.go
JAORMX: Confirmed FIXED — all three branches in `processSSEResponse` (decode-error, carriesResult-drop, genuine-non-Response) now `slog.Warn(..., "method", rfw.method)` unconditionally. Since this function only runs for the four `requiresResponseFiltering` methods, bypasses on all of them now log.

pkg/authz/response_filter_test.go
JAORMX: Confirmed FIXED — `TestResponseFilteringWriter_SSE_PerLineFallthrough` is a genuine 3 methods × 2 scenarios table (tools/list, prompts/list, resources/list × notification-interleaved, undecodable-line), each with its own Cedar policy and authorized/unauthorized fixture. Not a shallow method-name parameterization.

pkg/authz/response_filter.go
JAORMX: Partially fixed. The exact PoC shape you gave (method+result, valid jsonrpc/id) is now caught — `carriesResult` (response_filter.go:278-283) does a proper `json.Unmarshal` into a struct, not substring matching, and `TestResponseFilteringWriter_SSE_DisguisedResponseFrame` genuinely exercises it.

But the fix is incomplete: `carriesResult` is only consulted in the `err == nil && !isResponse` branch (line 217). `jsonrpc2.DecodeMessage` returns an *error* — not a benign non-Response — for several upstream-controlled shapes (missing/invalid `jsonrpc` version tag, a response-shaped frame with no `id`, a non-scalar `id`, or a batch/array payload), and all of those land in the `if err != nil` branch at line 191, which still passes the line through unfiltered without ever checking for a smuggled `result`. That's the same bypass class, just relocated to an easier-to-reach branch. Left a detailed comment with concrete exploit frames on line 191.

Also: `processJSONResponse` (line 132) has the identical unpatched shape for the `application/json` transport — the PR's 'Special notes for reviewers' says that path is 'already correct', but the disguised-frame issue is transport-independent, not stream-specific, so that note isn't right.

pkg/authz/response_filter.go
JAORMX: Confirmed FIXED — the filtered-write path no longer appends a hardcoded terminator; the loop's shared `linesep` write at the bottom handles termination for every branch. No more \r\n desync.

pkg/authz/response_filter.go
JAORMX: Confirmed FIXED — this is now a flat `if err != nil {...} else if isResponse {...} else if carriesResult(...) {...} else {...}` chain at one indentation level. Happy path is no longer buried in a default case.

pkg/authz/response_filter.go
```diff
@@ -187,35 +187,51 @@ func (rfw *ResponseFilteringWriter) processSSEResponse(rawResponse []byte) error
 		var written bool
 		if data, ok := bytes.CutPrefix(line, []byte("data:")); ok {
 			message, err := jsonrpc2.DecodeMessage(data)
+			response, isResponse := message.(*jsonrpc2.Response)
 			if err != nil {
```
JAORMX: This is the actual hole. `jsonrpc2.DecodeMessage` returns a non-nil `err` — landing here, not in the `carriesResult` branch below — for several frame shapes a malicious/compromised upstream fully controls:

- missing or invalid `jsonrpc` version tag
- a response-shaped frame (no `method`) with no valid `id`
- an `id` of a non-scalar type (object/array/bool)
- a JSON **array** (batch) payload

All of these hit this branch and get written through unfiltered, e.g.:

```
data: {"id":1,"result":{"tools":[{"name":"admin_tool"}]}}
```
(missing `jsonrpc` tag → decode error → unfiltered passthrough), or

```
data: [{"jsonrpc":"2.0","id":1,"result":{"tools":[{"name":"admin_tool"}]}}]
```
(batch array → decode error; note `carriesResult`'s struct-based probe would also return false on this shape even if it ran here, since it doesn't handle arrays).

Any client lenient about the `jsonrpc` tag, a missing `id`, or JSON-RPC batching (which MCP supported prior to 2025-06-18) will read the smuggled `result` as the real response — exactly the Cedar bypass this PR sets out to close, just relocated here. Recommend restructuring so the smuggled-result check runs independent of whether `DecodeMessage` succeeded, and extending it to detect a `result` key inside array-shaped payloads too. This is the same class as #5257 and should fail closed, not open, on malformed/adversarial input.

pkg/authz/response_filter.go
```diff
@@ -187,35 +187,51 @@ func (rfw *ResponseFilteringWriter) processSSEResponse(rawResponse []byte) error
...
+					"method", rfw.method)
+				written = true
+			} else {
+				// Genuine non-Response frame (e.g. an interleaved notifications/*
+				// message) with no result payload. Pass through unfiltered for
+				// this line only; a later data line may still carry the real
+				// response and must reach the filter.
+				slog.Warn("SSE data line was not a JSON-RPC Response; passing through unfiltered",
+					"method", rfw.method)
 			}
```
JAORMX, line 195: Minor: this branch (genuine non-Response frame, e.g. an interleaved `notifications/*` message) is spec-legitimate, routine SSE traffic, but it logs at `Warn` — the same level as the two actually-suspicious branches (undecodable line, smuggled-result drop). Warning on every routine notification will train operators to filter the WARN level, which then buries the two branches that actually indicate something adversarial. Consider demoting this one to `Debug` so severity tracks suspiciousness.

JAORMX (dismissed):
Re-reviewed against head commit eaa6c0d. Items 1, 2, and 4 from the two review rounds are genuinely fixed with real regression tests that fail without the fix — good work there. Item 3 (the method+result frame-smuggling bypass) is only partially closed: `carriesResult` correctly catches the exact PoC shape via a proper JSON unmarshal, but it's only consulted in the `err == nil && !isResponse` branch. `jsonrpc2.DecodeMessage` returns an *error* — not a benign non-Response — for several upstream-controlled cases (missing/invalid `jsonrpc` version tag, a response-shaped frame with no `id`, a non-scalar `id`, or a batch/array payload), and all of those land in the `err != nil` branch, which still passes the line through unfiltered without ever checking for a smuggled `result`. That's the exact #5257 bypass this PR exists to close, just relocated to an easier-to-reach branch — and it fails open on a component whose entire job is enforcing an authz boundary against an untrusted upstream.

The sibling `processJSONResponse` has the identical unpatched shape for the `application/json` transport. The PR's 'Special notes for reviewers' dismisses this as 'already correct', but that's not right — the disguised-frame issue is transport-independent, not stream-specific.

Blocking:
- Make the result-smuggling check run regardless of `DecodeMessage`'s error outcome — fail closed on any `data:` line under a filtered method that either can't be classified as a benign non-response or carries a `result` — and handle array/batch payloads too.
- Apply the same guard to `processJSONResponse`, or explicitly track it as a fast-follow and correct the PR notes (don't leave 'already correct' standing).
- Extend `TestResponseFilteringWriter_SSE_DisguisedResponseFrame` with the additional smuggled shapes (missing `jsonrpc` tag, missing `id`, batch array) — all three currently leak `admin_tool` per this review.

Non-blocking:
- `carriesResult` has no standalone unit test independent of the SSE integration test.
- `find_tool` isn't covered by the new fallthrough test table (only the three list methods are).

Replied inline on the specific threads, plus two new comments below.

saivedant169:
Pushed 644506a. You were right on both counts.

The smuggled-result check now runs independent of DecodeMessage's outcome. I reordered the SSE branch so `carriesResult` sits between the clean-Response filter and the decode-error passthrough, so a result on any frame that is not a clean Response gets dropped, whether DecodeMessage returned a non-Response or errored. That covers the four shapes you listed: missing/invalid jsonrpc tag, no id, non-scalar id, and batch arrays.

`carriesResult` handles arrays now too: it probes the object form, and for a `[`-prefixed payload it unmarshals into a slice and drops the frame if any element carries a result.

`processJSONResponse` had the same shape, so I applied the same fix there: if the frame is not a clean Response but carries a result, it fails closed (writes `{}`) instead of passing the raw body through. My "already correct" note was wrong, the bug is transport-independent like you said.

Demoted the genuine-notification SSE branch to Debug so the two branches that actually indicate something adversarial (undecodable line, smuggled-result drop) keep Warn to themselves.

Tests: `TestResponseFilteringWriter_SSE_DisguisedResponseFrame` is now a table over all five frame shapes, and there is a new `TestResponseFilteringWriter_JSON_DisguisedResponseFrame` for the application/json path. I checked them against the pre-fix source first: batch, missing-jsonrpc, no-id, and the JSON transport all leaked `admin_tool` there and pass after the fix.

One judgment call worth flagging: on the JSON path I fail closed by writing `{}` rather than synthesizing a JSON-RPC error, since a rejected or batch frame may not carry a usable id. Can switch it to an error response if you would prefer that shape.

saivedant169:
@JAORMX pushed the fix for the decode-error and batch bypass last week (644506a) with regression tests for all the frame shapes you listed. Ready for another look when you have a moment, and the workflows need an approval to run.

pkg/authz/response_filter.go
```diff
@@ -131,17 +131,22 @@ func (rfw *ResponseFilteringWriter) Flush() {
...
+	if err != nil || !ok {
+		// Not a clean Response. The disguised-result bypass (#5257) is
+		// transport-independent, so apply the same check here as on the SSE
+		// path: if a non-Response, undecodable, or batch frame still carries a
+		// result, fail closed rather than passing the smuggled list through.
+		if carriesResult(rawResponse) {
+			slog.Warn("JSON response carried a result outside a clean Response frame; dropping as a protocol violation",
+				"method", rfw.method)
+			rfw.ResponseWriter.WriteHeader(rfw.statusCode)
+			_, werr := rfw.ResponseWriter.Write([]byte("{}"))
```
aponcedeleonch: Non-blocking nit: this fail-closed path writes a bare `{}`, which isn't a valid JSON-RPC response. The clean-but-filtered path already returns a proper error object, so returning a JSON-RPC error here too would be more consistent for clients. Cosmetic.

pkg/authz/response_filter.go
```diff
@@ -187,35 +192,49 @@ func (rfw *ResponseFilteringWriter) processSSEResponse(rawResponse []byte) error
...
+				}
+
+				// The loop writes linesep after this line, so do not append a
+				// terminator here. A hardcoded "\n" desyncs \r\n streams.
+				if _, err := rfw.ResponseWriter.Write([]byte("data: " + string(filteredData))); err != nil {
+					return fmt.Errorf("%w: %w", errBug, err)
+				}
+
+				written = true
+			} else if carriesResult(data) {
```
aponcedeleonch: Non-blocking nit: the `//nolint:gocyclo` on processSSEResponse (line 167) could be dropped by extracting this classify-and-handle chain (decode-error / carriesResult / genuine non-Response) into a small helper. Purely cosmetic.

aponcedeleonch (dismissed):
Great catch, and thanks for the thorough tests. One process note for next time: please ask to have the issue assigned to you before opening a PR, per our [contributor guidelines](https://github.com/stacklok/toolhive/blob/main/CONTRIBUTING.md#claiming-an-issue). #5257 is the issue this closes.

On the fix: this is a confirmed P0 authz bypass (the cedar filter leaking the full tool/prompt/resource catalog on SSE upstreams), and it addresses root cause. I built the head and verified it: it reproduces on unfixed main and is fully resolved here, with a table-driven suite covering all three list methods, both fallthrough triggers, and five disguised-result frame shapes across both transports.

Reconciling with the earlier review: JAORMX's blocking item from July 3 (the result-smuggling check only ran in the err == nil branch, so undecodable / no-id / non-scalar-id / batch frames still leaked) is now closed. carriesResult is consulted for all of those shapes and they fail closed, with regression coverage. The routine interleaved-notification log is also already demoted to Debug. @JAORMX you may want to dismiss your changes-requested when you get a chance.

Approving. Left two cosmetic nits inline, neither blocking.

pkg/authz/response_filter.go
aponcedeleonch: Addressed: all branches now log with the method, so bypasses on every filtered method surface, not just tools/list.

pkg/authz/response_filter_test.go
aponcedeleonch: Addressed: the regression table now covers tools/list, prompts/list, and resources/list, not just tools/list.

pkg/authz/response_filter.go
aponcedeleonch: Addressed: frames are now classified by carriesResult and fail closed regardless of the decoded message type, so a method+result frame no longer passes through.

pkg/authz/response_filter.go
aponcedeleonch: Addressed (you confirmed this inline): the filtered-write path no longer appends a terminator and lets the shared linesep write handle it.

pkg/authz/response_filter.go
aponcedeleonch: Addressed (you confirmed this inline): it is now a flat if/else-if chain with the happy path at one indentation level.

pkg/authz/response_filter.go
aponcedeleonch: Addressed: the July 3 follow-up commit routes decode-error, no-id, non-scalar-id, and batch frames through the fail-closed carriesResult check, with regression tests for each shape.

pkg/authz/response_filter.go
aponcedeleonch: Addressed: the genuine non-Response branch now logs at Debug, leaving Warn for the two suspicious branches.

aponcedeleonch (approved):
Re-reviewed against c7ae85da3. Both cosmetic nits from my last review are addressed:

- The fail-closed path in `processJSONResponse` now returns a proper JSON-RPC error via `writeErrorResponse` instead of a bare `{}`.
- The `//nolint:gocyclo` on `processSSEResponse` is gone, with the classify-and-handle chain pulled into the new `writeSSEDataLine` helper. Reads cleanly.

Verified the authz package builds and tests pass with race detection. Approving.

pkg/authz/response_filter.go
aponcedeleonch: addressed

pkg/authz/response_filter.go
aponcedeleonch: addressed

PR #5319, opened 2026-05-19
fix: validate k8s export volume format

immanuwell:
Small papercut fix for k8s export.

`host-path:container-path[:ro]` was documented, but the exporter accepted stuff like `/host:/container:rw` or `/host:/container:ro:extra` and just kept going. bit yikes, because the generated MCPServer can hide a bad saved RunConfig instead of failing fast.

Repro:
```bash
go test ./pkg/export -run TestParseVolumeString -count=1
```

Before this patch, the new malformed cases fail because no error is returned. After it, unknown modes and extra fields are rejected.

Checked:
```bash
go test ./pkg/export -count=1
go test ./cmd/thv/app -run Test -count=1
go test ./pkg/...
```

No external quota or k8s limit makes this unreachable; it is plain RunConfig string data loaded by `thv export --format k8s`.


jhrozek (approved):

jhrozek:
thanks for the contribution @immanuwell !

Issue #4825, opened 2026-04-14
Support Client ID Metadata Document (CIMD) in the embedded authorization server


jhrozek:
As an MCP client (VS Code, Claude Code, or any spec-compliant client) connecting to a ToolHive-managed MCP server locally, I want to use an HTTPS URL as my `client_id` so the authorization server fetches my client metadata automatically, without requiring a separate DCR registration step.

## Background

The MCP 2025-11-25 spec defines client registration priority as: pre-registered credentials > **CIMD (preferred)** > DCR > user prompt. ToolHive's embedded AS currently supports only DCR (RFC 7591). CIMD ([draft-ietf-oauth-client-id-metadata-document](https://datatracker.ietf.org/doc/html/draft-parecki-oauth-client-id-metadata-document-03)) is the spec-preferred mechanism but neither ToolHive, Ory Hydra, nor Obot implement it today.

In the local flow, the MCP client (VS Code, Claude Code) is the OAuth client. The thv proxy is the resource server and hosts the embedded AS. The proxy itself does not need CIMD awareness for this flow.

VS Code already implements CIMD client-side support (`mainThreadAuthentication.ts:160-170`): when the AS advertises `client_id_metadata_document_supported: true`, VS Code uses its product-configured metadata URL as `client_id` instead of performing DCR.

## Acceptance criteria

- [ ] MCP clients can pass an HTTPS URL as `client_id` in `/oauth/authorize` and `/oauth/token` and complete the full authorization code + PKCE flow without calling `/oauth/register`
- [ ] Discovery documents (`/.well-known/oauth-authorization-server`, `/.well-known/openid-configuration`) advertise `"client_id_metadata_document_supported": true` when CIMD is enabled
- [ ] Existing DCR continues to work alongside CIMD (no regression)
- [ ] CIMD is opt-in via configuration (disabled by default)

### URL detection and coexistence with DCR

- A `client_id` is treated as CIMD when it starts with `https://`
- Inner storage is checked first: if `GetClient` finds an existing client (DCR or static), use it. CIMD fetch only on `ErrNotFound`. This prevents collisions with existing client IDs.

### Storage decorator pattern

Fosite resolves clients via `Storage.GetClient(ctx, id string)` in both `NewAuthorizeRequest` (authorize endpoint) and `NewAccessRequest` (token endpoint). A decorator wrapping `storage.Storage` that intercepts `GetClient` handles both flows transparently.

[...]

PR #5320, opened 2026-05-19
Add CIMD document fetch/validate and extend SSRF protections

amirejaz:
## Summary

Phase 2 PR 1 of CIMD embedded AS support — closes part of #4825.

The embedded authorization server (Phase 2) will need to fetch and validate Client ID Metadata Documents when MCP clients present HTTPS URLs as `client_id`. This PR adds that foundation as a pure library layer with no server wiring — wiring comes in PR 2 (storage decorator) and PR 3 (config + discovery).

- **`FetchClientMetadataDocument`** fetches a CIMD document with the security controls required by the spec: HTTPS-only, 5 s timeout, 10 KB body cap, one-hop redirect limit, per-dial SSRF protection, Content-Type validation, and strict self-referential binding (`client_id` must exactly equal the serving URL).
- **`ValidateClientMetadataDocument`** checks required fields and allowed redirect URI schemes.
- **`ClientMetadataDocument`** struct matches draft-ietf-oauth-client-id-metadata-document fields.
- **SSRF range extensions** in `pkg/networking`: RFC6598 CGN (`100.64.0.0/10`) and RFC5737 documentation ranges added to the private IP block list.
- **`WithDisableKeepAlives`** on `HttpClientBuilder` so callers can prevent connection reuse from bypassing per-dial SSRF checks.

The SSRF check in `cimd.go` is implemented inline rather than via `pkg/networking` to preserve the `oauthproto` leaf-package invariant (documented in `doc.go`).

## Type of change

- [x] New feature (non-breaking)

## Test plan

- [x] Unit tests: `go test ./pkg/oauthproto/... ./pkg/networking/...` — all pass
- [x] Fetch/validate tests cover: valid document, non-HTTPS rejection, non-200 status, 10 KB cap, Content-Type validation, `application/*+json` subtype, `client_id` mismatch, missing `redirect_uris`, SSRF private IP rejection
- [x] New IP range tests: CGN and documentation ranges verified as private

The duplicate SSRF block list in `cimd.go` is intentional — see the comment referencing `doc.go`. If the leaf-package invariant is ever relaxed, `cimdPrivateBlocks` can be replaced with a call to `networking.IsPrivateIP`.


pkg/oauthproto/cimd.go
```diff
@@ -12,15 +23,295 @@ import "strings"
...
+}
+
+// isLoopbackHTTP reports whether rawURL is an HTTP URL targeting a loopback
+// address (localhost, 127.x.x.x, or [::1]). Allowing loopback HTTP makes
+// local development and testing practical without weakening production security.
+func isLoopbackHTTP(rawURL string) bool {
+	return strings.HasPrefix(rawURL, "http://localhost") ||
+		strings.HasPrefix(rawURL, "http://127.") ||
+		strings.HasPrefix(rawURL, "http://[::1]")
+}
```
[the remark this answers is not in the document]
amirejaz: Fixed. isLoopbackHTTP now parses the URL and compares the hostname exactly (using parsed.Hostname() + net.ParseIP/IsLoopback) rather than a string prefix. http://localhost.evil.com/ is now correctly rejected.

pkg/oauthproto/cimd.go
```diff
@@ -12,15 +23,295 @@ import "strings"
...
+	client := &http.Client{
+		Timeout:   5 * time.Second,
+		Transport: transport,
+		CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
+			redirectCount++
+			if redirectCount > 1 {
+				return http.ErrUseLastResponse
+			}
+			return nil
+		},
```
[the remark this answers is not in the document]
amirejaz: Fixed. CheckRedirect now calls validateCIMDClientURL on every redirect target before following it. HTTP redirects and redirects to private/forbidden URLs are rejected at hop time, and the 1-hop limit is enforced via len(via).

pkg/oauthproto/cimd.go
```diff
@@ -12,15 +23,295 @@ import "strings"
...
+		DisableKeepAlives:     true,
+		DialContext: func(ctx context.Context, network, address string) (net.Conn, error) {
+			host, port, err := net.SplitHostPort(address)
+			if err != nil {
+				return nil, fmt.Errorf("cimd: invalid dial address %q: %w", address, err)
+			}
+			ips, err := net.DefaultResolver.LookupHost(ctx, host)
+			if err != nil {
+				return nil, fmt.Errorf("cimd: failed to resolve %q: %w", host, err)
+			}
```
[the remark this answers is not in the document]
amirejaz: Fixed. After resolving DNS, the DialContext now dials by the validated IP literal (net.JoinHostPort(ipStr, port)) instead of by the original hostname. This eliminates the TOCTOU window between resolution and dial.

pkg/oauthproto/cimd.go
```diff
@@ -12,15 +23,295 @@ import "strings"
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch client metadata document: %w", err)
+	}
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
```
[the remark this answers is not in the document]
amirejaz: Fixed. Removed the io.Copy(io.Discard) drain from the defer. Keep-alive connections are disabled (DisableKeepAlives: true) so connection reuse is not a concern, and draining would defeat the 10 KB body cap by reading past the LimitReader.

pkg/oauthproto/cimd.go
```diff
@@ -12,15 +23,295 @@ import "strings"
...
+	}
+
+	if doc.TokenEndpointAuthMethod != "" {
+		for _, forbidden := range forbiddenAuthMethods {
+			if doc.TokenEndpointAuthMethod == forbidden {
+				return fmt.Errorf("token_endpoint_auth_method %q is not permitted in a CIMD document: "+
+					"symmetric shared-secret methods cannot be used without pre-registration", forbidden)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
amirejaz: Clarified with a comment. Only symmetric shared-secret methods are forbidden per the CIMD spec (client_secret_post, client_secret_basic, client_secret_jwt). Asymmetric methods such as private_key_jwt, tls_client_auth, and none are valid and remain accepted. The PR description has been updated to match.

pkg/oauthproto/cimd_test.go
```diff
@@ -42,3 +51,311 @@ func TestIsClientIDMetadataDocumentURL(t *testing.T) {
...
+		_, err := FetchClientMetadataDocument(context.Background(), serverURL)
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "redirect_uris")
+	})
+
+	t.Run("SSRF: private non-loopback IP is blocked", func(t *testing.T) {
+		t.Parallel()
+
+		// 10.0.0.1 is a private RFC1918 address; the SSRF dial guard must
+		// reject it before a connection is established.
```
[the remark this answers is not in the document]
amirejaz: Fixed. The old test is renamed to 'SSRF: URL scheme check rejects non-loopback HTTP' to accurately describe what it exercises. A new test 'SSRF: dial-guard blocks private IP served via HTTPS hostname' uses https://192.0.2.1/metadata.json (TEST-NET-1, RFC5737) — it passes validateCIMDClientURL but is caught by the DialContext SSRF guard, which returns 'refusing connection to private address'.

pkg/oauthproto/cimd.go
```diff
@@ -12,15 +23,307 @@ import "strings"
...
 }
+
+// validateCIMDClientURL enforces the client_id URL requirements from
+// draft-ietf-oauth-client-id-metadata-document §3:
+//   - MUST use https scheme (http://localhost permitted for development only)
+//   - MUST contain a non-empty path component (not just "/")
+//   - MUST NOT contain a fragment component
+//   - MUST NOT contain userinfo (username or password)
+//   - MUST NOT contain single-dot or double-dot path segments
+func validateCIMDClientURL(rawURL string) error {
```
jhrozek: why not use `pkg/oauthproto.ValidateRedirectURI` ?

pkg/oauthproto/cimd.go
```diff
@@ -12,15 +23,307 @@ import "strings"
...
+			return fmt.Errorf("client_id URL must not contain dot-segment path components")
+		}
+	}
+	return nil
+}
+
+// isLoopbackHTTP reports whether rawURL is an HTTP URL targeting a loopback
+// address. The host is parsed from the URL (not matched as a prefix) to
+// prevent bypasses via hosts like "http://localhost.evil.com/".
+func isLoopbackHTTP(rawURL string) bool {
```
jhrozek: why not use `pkg/oauthproto.IsLoopbackHost` ?

pkg/oauthproto/cimd.go
```diff
@@ -12,15 +23,307 @@ import "strings"
...
+	}
+
+	return &doc, nil
+}
+
+// newCIMDHTTPClient builds an *http.Client with the SSRF and redirect guards
+// required for fetching client metadata documents. Keep-alive connections are
+// disabled so the per-dial SSRF check fires on every request rather than being
+// bypassed by a reused connection.
+func newCIMDHTTPClient() *http.Client {
```
jhrozek: why not use `networking.NewHttpClientBuilder().WithPrivateIPs(false).WithDisableKeepAlives(true).Build()` ?

pkg/oauthproto/cimd.go
```diff
@@ -12,15 +23,307 @@ import "strings"
...
+	}
+
+	ct := resp.Header.Get("Content-Type")
+	mediaType, _, parseErr := mime.ParseMediaType(ct)
+	if parseErr != nil || (mediaType != "application/json" && !isJSONSubtype(mediaType)) {
+		return nil, fmt.Errorf("client metadata document has unexpected Content-Type: %q", ct)
+	}
+
+	const maxBodyBytes = 10 * 1024
+	limited := io.LimitReader(resp.Body, maxBodyBytes)
```
jhrozek: Why not use `networking.FetchJSON[T]` ?

jhrozek (commented):
I think the code itself is good, but I think we should reduce the code duplication

pkg/oauthproto/cimd.go
amirejaz: Fixed. validateRedirectURI is removed and replaced with oauthproto.ValidateRedirectURI(uri, oauthproto.RedirectURIPolicyStrict), which was already in the same package — we just missed it.

pkg/oauthproto/cimd.go
amirejaz: Fixed. isLoopbackHTTP now uses oauthproto.IsLoopbackHost(parsed.Hostname()) after url.Parse, which also closes the localhost.evil.com prefix-match bypass that Copilot flagged.

pkg/oauthproto/cimd.go
amirejaz: The duplication of the private IP block list is resolved: newCIMDHTTPClient now uses networking.IsPrivateIP instead of the old cimdPrivateBlocks init block. HttpClientBuilder.WithPrivateIPs(false) cannot be used directly because it also blocks 127.0.0.0/8, which breaks the intentional http://localhost development exception. The custom DialContext is kept for this nuance but is documented explicitly.

pkg/oauthproto/cimd.go
amirejaz: FetchJSON[T] uses a 1MB hardcoded size limit (maxResponseSize) while the CIMD spec recommends 5KB and our consumer-side guard is 10KB. Using FetchJSON would silently bypass our body cap. The move to the sub-package does allow importing pkg/networking and reusing IsPrivateIP, which addresses the block-list duplication.

jhrozek: I'd rather we fix `FetchJSON` than work around it here.

The 1 MB cap at `fetch.go:29` is hardcoded today, but adding `WithMaxResponseSize(int64) FetchOption` is a couple of lines in the existing options struct. We've got several other places that hand-roll `io.LimitReader` with their own constants (`pkg/auth/token.go`, `pkg/auth/discovery/discovery.go`, `pkg/oauthproto/discovery.go`, `pkg/oauthproto/dcr.go`) which would also benefit.

The other thing: the content-type check at `fetch.go:155` is actually wrong. `strings.Contains("application/ld+json", "application/json")` returns false. Per RFC 6839 `+json` subtypes are JSON, so that's a latent bug in `FetchJSON` independent of CIMD. The `mime.ParseMediaType` + `+json` check you wrote in this PR is the correct version and we should move it into `FetchJSON`.

If we do both, the fetch loop here boils down to:

```go
return networking.FetchJSON[ClientMetadataDocument](ctx, httpClient, rawURL,
    networking.WithMaxResponseSize(10*1024),
)
```

About 30 lines gone and the `+json` bug squashed for any future caller. The custom HTTP client stays, that's CIMD-specific.

Fine with either doing it in this PR or as a follow-up, your call.

amirejaz: Done. Added WithMaxResponseSize(int64) to FetchJSON and fixed the Content-Type bug (strings.Contains was rejecting valid +json subtypes like ld+json). FetchClientMetadataDocument now calls networking.FetchJSON[ClientMetadataDocument](ctx, newCIMDHTTPClient(), rawURL, networking.WithMaxResponseSize(10*1024)) — the custom HTTP client stays for the loopback exception and IP-literal dialing, but the fetch loop is gone.

pkg/oauthproto/cimd/fetch.go
```diff
@@ -0,0 +1,217 @@
...
+			}
+			return nil, fmt.Errorf("cimd: no valid address found for %q", host)
+		},
+	}
+
+	client := &http.Client{
+		Timeout:   5 * time.Second,
+		Transport: transport,
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if len(via) >= 1 {
```
jhrozek: nit: this branch is unreachable. `CheckRedirect` is invoked with `via` already containing the original request, so `len(via) >= 1` is true on the first redirect and the `validateCIMDClientURL` call below never runs. Net effect today: zero redirects are followed (a 3xx surfaces to the caller as an `HTTP 30x` error from the status check). Per CIMD §3.2 the AS MUST NOT auto-follow redirects, so the *behaviour* is fine — but the validator block is dead and the comment on `newCIMDHTTPClient` ("enforces the 1-hop limit") doesn't match. Either drop the dead branch and simplify to `return http.ErrUseLastResponse`, or change the guard to `len(via) >= 2` to actually permit one hop.

pkg/oauthproto/cimd/fetch.go
```diff
@@ -0,0 +1,217 @@
...
+//   - MUST NOT contain a fragment component
+//   - MUST NOT contain userinfo (username or password)
+//   - MUST NOT contain single-dot or double-dot path segments
+func validateCIMDClientURL(rawURL string) error {
+	parsed, err := url.Parse(rawURL)
+	if err != nil {
+		return fmt.Errorf("client_id URL is not a valid URL: %w", err)
+	}
+	isLoopback := parsed.Scheme == "http" && oauthproto.IsLoopbackHost(parsed.Hostname())
+	if !strings.HasPrefix(rawURL, "https://") && !isLoopback {
```
jhrozek: nit: the loopback branch one line up uses `parsed.Scheme`; this line falls back to `HasPrefix` on the raw string. `HTTPS://example.com/metadata.json` parses fine (`parsed.Scheme == "https"`) but fails this prefix check and is rejected with a misleading "must use the https scheme" error. URI schemes are case-insensitive per RFC 3986 §3.1.

```suggestion
	if parsed.Scheme != "https" && !isLoopback {
```

pkg/oauthproto/cimd/fetch.go
```diff
@@ -0,0 +1,217 @@
...
+	if parsed.Fragment != "" {
+		return fmt.Errorf("client_id URL must not contain a fragment component")
+	}
+	if parsed.User != nil {
+		return fmt.Errorf("client_id URL must not contain userinfo (username or password)")
+	}
+	if parsed.Path == "" || parsed.Path == "/" {
+		return fmt.Errorf("client_id URL must contain a non-empty path component")
+	}
+	for _, segment := range strings.Split(parsed.Path, "/") {
```
jhrozek, line 171: nit / follow-up: CIMD §3 also says "Client identifier URLs SHOULD NOT include a query string component" — `parsed.RawQuery` isn't checked anywhere. Happy to defer to a follow-up if you'd rather keep this PR tight.

jhrozek (dismissed):
The FetchJSON refactor in ab8a5b80a cleaned up the duplication I was about to ask about — nice. Three small nits below, all non-blocking. Approve.

pkg/oauthproto/cimd/fetch.go
amirejaz: Fixed. The CIMD spec (§4) says the AS MUST NOT automatically follow HTTP redirects, so the correct behaviour is no redirects at all. Simplified to unconditional `return http.ErrUseLastResponse` with the spec citation in the comment. `FetchJSON` then surfaces the 3xx as a non-200 error as expected.

pkg/oauthproto/cimd/fetch.go
amirejaz: Fixed — changed to `parsed.Scheme != "https"` to be consistent with the loopback check above and correct per RFC 3986 §3.1.

pkg/oauthproto/cimd/fetch.go
amirejaz: Deferred to a follow-up issue — it's a SHOULD, and keeping this PR tight feels right. Will track it alongside the other Phase 2 follow-ups.

jhrozek (approved):

Issue #5334, opened 2026-05-19
[vMCP] Extend identity-refresh integration test to wire TokenValidator.Middleware + fake UpstreamTokenRefresher


tgrunnagle:
## Context

The fix for #5323 added a higher-level integration test (`pkg/vmcp/session/internal/backend/mcp_session_identity_refresh_test.go`) that exercises the full transport chain — `headerForwardRoundTripper → identityRoundTripper → authRoundTripper (UpstreamInjectStrategy) → http.DefaultTransport` — and verifies that a fresh identity placed on `req.Context()` drives the upstream `Authorization` header.

That test attaches `freshIdentity` to the request context directly via `auth.WithIdentity(...)`, which is the same call `TokenValidator.Middleware` makes at `pkg/auth/token.go:1245`. The iteration 2 review of #5323 explicitly allowed this higher-level alternative, but noted that it does not verify the link from `TokenValidator.Middleware` → `loadUpstreamTokens` → `UpstreamTokenRefresher` → fresh map on `identity.UpstreamTokens`.

## Gap

The user-visible "forced re-auth every ~24h" symptom can recur via multiple distinct failure modes:

1. The transport overriding fresh identity (the bug fixed in #5323, now covered).
2. `loadUpstreamTokens` returning a stale map because `UpstreamTokenRefresher` is broken.
3. `TokenValidator.Middleware` failing to call `WithIdentity` on the refresh path.

The new test only catches (1). Modes (2) and (3) would still produce the same user-visible symptom.

## Proposal

Extend `mcp_session_identity_refresh_test.go` (or create a sibling test in a higher-level package such as `pkg/vmcp/server/...`) with an end-to-end case that:

1. Builds a minimal `TokenValidator` wired to a fake `UpstreamTokenRefresher` whose returned tokens differ between the first and second call (simulating transparent refresh).
2. Constructs a signed JWT carrying a `tsid` claim that the validator can resolve via the fake `UpstreamTokenRefresher`.
3. Issues two requests through the validator's middleware into the backend session.
4. Asserts the upstream observes the rotated token on the second call.

This is the AC5 acceptance criterion as literally written on #5323. Closing this gap is non-trivial test setup (signed JWT, fake refresher, real middleware wiring), which is why the iteration 2 review allowed deferring it.

## Acceptance criteria

- [ ] A test exists that exercises `TokenValidator.Middleware → loadUpstreamTokens → UpstreamTokenRefresher` end-to-end against a vMCP backend session.
[...]

Issue #5333, opened 2026-05-19
[vMCP] Consolidate duplicate identity-propagation round-trippers into shared internal transport package


tgrunnagle:
## Context

After the fix in #5323 (commits `2298d349` + `109a8a15`), the identity-propagation round-tripper exists in two parallel implementations:

- `pkg/vmcp/client/client.go` — `identityPropagatingRoundTripper` (with `isHealthCheck` field; used by the per-call `httpBackendClient`).
- `pkg/vmcp/session/internal/backend/mcp_session.go` — `identityRoundTripper` (without `isHealthCheck`; used by the persistent-session connector).

The two types now have:
- Near-identical struct definitions (`base`, `fallbackIdentity`, with one carrying an extra `isHealthCheck` bool).
- Near-identical `RoundTrip` implementations, both gating fallback injection on `!hasIdentity`.
- Near-identical doc comments (each citing #5323).
- Near-identical unit test suites (both following the `*_PerRequestIdentity_*` / `*_FallbackIdentity_*` grouping).

The asymmetric health-check handling is documented in the `identityRoundTripper` doc comment, but the structural duplication remains a long-term maintenance hazard: any future change to the identity invariant (e.g. logging on fallback injection, metrics counters, additional checks) must be repeated in both files with no compile-time link between them.

## Proposal

Extract both round-trippers into a small internal helper package — e.g. `pkg/vmcp/internal/transport` — and have both `pkg/vmcp/client/client.go` and `pkg/vmcp/session/internal/backend/mcp_session.go` consume the same type. Parameterise the health-check marker so the same implementation serves both call sites:

- Constructor option pattern (`transport.WithHealthCheckMarker()`) or
- Two thin wrapper types in the same package sharing a common implementation.

The consolidation should also unify the unit test suites in one package, removing the lock-step maintenance requirement.

## Acceptance criteria

- [ ] Single round-tripper type (or family) lives in `pkg/vmcp/internal/transport` (or similar) and is consumed by both `pkg/vmcp/client/client.go` and `pkg/vmcp/session/internal/backend/mcp_session.go`.
- [ ] Health-check marker propagation is opt-in via a constructor option, not a separate type that the session-backend variant lacks.
- [ ] Unit test suite for the round-tripper(s) lives alongside the shared implementation; the consumer packages keep only integration-level tests that exercise the full transport chain.
[...]

Issue #3877, opened 2026-02-18
[vMCP] Proactive auth credential refresh via per-request identity resolution (Phase 2)


yrobla:
Depends on #3869

This is the Phase 2 follow-up to #3869 (auth credential expiration with retry). Phase 1 handles stale credentials reactively (detect 401/403, recreate client). This issue eliminates the problem proactively by resolving credentials on every operation rather than at client creation time.

**Context**: With session-scoped clients, outgoing credentials are resolved once during `SessionFactory.MakeSession()` and held for the session lifetime. Short-lived credentials (OAuth tokens, expiring API keys) can become stale mid-session, triggering the retry path in #3869. This issue removes that failure mode entirely for credentials with a known expiry.

**Implementation**:

**Part 1 — Per-request identity resolution**:
- Modify `identityPropagatingRoundTripper` in `pkg/vmcp/client/client.go` to read identity from the request context on each operation, instead of capturing it at client creation time
- This ensures every backend call automatically picks up the latest credentials without any expiry tracking logic
- Eliminates stale credential issues entirely for backends that return fresh credentials on each resolution

**Part 2 — Proactive client recreation for known-expiry credentials**:
- For credentials that carry an explicit expiry (JWT `exp` claim, OAuth `expires_in` response field), schedule proactive client recreation shortly before the expiry time
- This avoids the reactive retry path in #3869 for these credential types — the client is already fresh before the first 401 occurs
- Proactive recreation should use the same `singleflight` deduplication as #3869 to handle concurrent requests during the recreation window

### Acceptance Criteria

- [ ] `identityPropagatingRoundTripper` resolves identity from the request context on each operation rather than capturing it at construction time
- [ ] Backends that return fresh credentials on each resolution never trigger the reactive retry path from #3869
- [ ] For credentials with a known expiry (`exp` claim or `expires_in`), proactive recreation is scheduled before the expiry time
- [ ] Proactive recreation uses `singleflight` to deduplicate concurrent requests during the recreation window
- [ ] Proactive recreation does not block in-flight requests — existing clients remain usable until replaced
- [ ] No credentials are logged or stored beyond what is required for the current request
- [ ] Unit tests cover: per-request identity resolution picks up updated credentials, proactive recreation fires before expiry, concurrent requests during recreation window are deduplicated
- [ ] Integration tests cover: full request lifecycle with short-lived credentials, no 401s observed when proactive refresh is working correctly

---
[...]

Issue #3869, opened 2026-02-18
[vMCP] Handle auth credential expiration with retry and proactive refresh


yrobla:
With session-scoped clients, short-lived outgoing credentials (e.g., OAuth tokens) resolved at client creation can become stale mid-session. Implement automatic client recreation on auth failure with proper safeguards.

**Phase 1 (this issue)**:
- Detect 401/403 from backend; recreate client with fresh credentials from `OutgoingAuthRegistry`
- Retry with exponential backoff, max 3 attempts
- Circuit-breaker: after N consecutive auth failures per backend, disable retries and log a warning
- Use `singleflight` to deduplicate concurrent recreation attempts for the same backend
- Auth-failure retries add non-trivial latency — surface this overhead in distributed traces

**Phase 2 (follow-up, separate issue)**:
- Modify `identityPropagatingRoundTripper` to read identity from request context per-operation, eliminating the stale credential problem entirely
- For credentials with a known expiry (JWT `exp`, OAuth `expires_in`): schedule proactive client recreation before expiry

### Acceptance Criteria

- [ ] A 401 or 403 from a backend triggers client recreation with fresh credentials
- [ ] Recreation is retried with exponential backoff up to a maximum of 3 attempts
- [ ] After N consecutive auth failures for a backend, the circuit breaker opens and further retries are skipped
- [ ] Concurrent auth failures for the same backend result in only one recreation attempt (via `singleflight`)
- [ ] Auth-failure retry latency is visible in distributed traces
- [ ] The raw credentials are never logged
- [ ] Unit tests cover: successful recreation, max retries exhausted, circuit breaker opens, concurrent failures deduplicated

PR #5335, opened 2026-05-19
Preserve fresh per-request identity in vMCP backend transports

### LINKED ISSUE #5323: [vMCP] identityPropagatingRoundTripper captures stale identity, bypassing per-request upstream token refresh
## Summary

`identityPropagatingRoundTripper` in `pkg/vmcp/client/client.go` captures the request `*auth.Identity` once at backend client creation time and re-injects that snapshot on every outgoing backend HTTP request, **overriding** the fresh identity placed onto the request context by `TokenValidator.Middleware` moments earlier. This bypasses the transparent upstream-token refresh performed in `GetAllValidTokens` and results in stale upstream access tokens being injected by `UpstreamInjectStrategy` / `TokenExchangeStrategy` for the entire lifetime of the backend client.

Users are observing a forced re-auth roughly every 24 hours. This matches the access-token lifespan of common upstream providers and is consistent with the captured-identity bug rather than with any embedded-auth-server refresh-token expiry (which is configurable up to 30 days and validated to be ≥1h in `pkg/authserver/server/provider.go`).

## Root cause

The bug is a two-path identity flow that doesn't reconnect:

**Path A — per-request refresh, works correctly**

`pkg/auth/token.go:1203` `TokenValidator.Middleware` runs on every incoming request, including tool calls. It calls `loadUpstreamTokens` → `InProcessService.GetAllValidTokens` (`pkg/auth/upstreamtoken/service.go:86`) which transparently refreshes any expired upstream access tokens via the configured `UpstreamTokenRefresher` and writes the fresh map into `identity.UpstreamTokens`. The fresh identity is then attached to the request context via `auth.WithIdentity` at `pkg/auth/token.go:1245`.

**Path B — captured at session-init, becomes stale**

`defaultClientFactory` in `pkg/vmcp/client/client.go:271` constructs the backend client's transport chain at `SessionFactory.MakeSession()` time. At `pkg/vmcp/client/client.go:319-324`, the identity is snapshotted **once**:

```go
identity, _ := auth.IdentityFromContext(ctx)
baseTransport = &identityPropagatingRoundTripper{
    base:          baseTransport,
    identity:      identity,        // captured here, held for session lifetime
    isHealthCheck: healthcontext.IsHealthCheck(ctx),
}
```

On every subsequent outgoing request, `identityPropagatingRoundTripper.RoundTrip` at `pkg/vmcp/client/client.go:195-207` **overrides** whatever identity is on the request context with the captured one:

```go
func (i *identityPropagatingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
    ctx := req.Context()
    if i.identity != nil {
        ctx = auth.WithIdentity(ctx, i.identity)   // discards fresh identity from Path A
    }
    ...
}
```

The downstream auth strategies read identity from the (now overridden) context and inject the captured stale token:

- `pkg/vmcp/auth/strategies/upstream_inject.go:80` — `token := identity.UpstreamTokens[providerName]`
- `pkg/vmcp/auth/strategies/tokenexchange.go:124` — `subjectToken = identity.UpstreamTokens[config.SubjectProviderName]`
- `pkg/vmcp/auth/strategies/aws_sts.go:136` — `bearerToken = identity.UpstreamTokens[cfg.SubjectProviderName]`

So tokens **are** refreshed at the vMCP edge on every tool call (Path A), then immediately discarded by the backend roundtripper (Path B) before they can be used.

## Observed user-facing behavior

- vMCP session opens; tool calls succeed.
- Tool calls continue to succeed for ~the upstream provider's access-token lifespan, even though the captured token is never refreshed at the backend layer.
- Once the captured access token passes its upstream-side expiry, every tool call to that backend starts returning 401 from the upstream API.
- Users perceive this as needing to re-authenticate. The cadence (~24h) tracks the shortest-lived upstream AT in the session's backend set, not anything ToolHive-configurable.

## Why the existing safety nets don't catch this today

- **#3869** ([vMCP] Handle auth credential expiration with retry and proactive refresh) — reactive 401/403 client recreation. **Open, not implemented.**
- **#3877** ([vMCP] Proactive auth credential refresh via per-request identity resolution) — Phase 2, whose Part 1 acceptance criterion is exactly the fix below. **Open, not implemented.**
- Reference: THV-0038 — Session-scoped client lifecycle.

This issue is filed separately from #3877 to flag that the Part 1 fix has a concrete user-visible impact (daily forced re-auth) and is worth landing on its own ahead of the rest of #3877's scope.

## Suggested fix

Make `identityPropagatingRoundTripper` source identity from the live request context instead of from a captured field. Sketch:

```go
type identityPropagatingRoundTripper struct {
    base          http.RoundTripper
    isHealthCheck bool
    // identity field removed
}

func (i *identityPropagatingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
    ctx := req.Context()
    // If mcp-go reconstructed this request with context.Background() (as it does
    // for streamable-HTTP Close()), there will be no identity on ctx. In that
    // case we can't recover one — health-check requests don't need it; for
    // session-bound requests the caller is responsible for propagating context.
    if i.isHealthCheck {
        ctx = healthcontext.WithHealthCheckMarker(ctx)
        req = req.Clone(ctx)
    }
    return i.base.RoundTrip(req)
}
```

Constructor at `pkg/vmcp/client/client.go:319-324` drops the `identity` capture:

```go
baseTransport = &identityPropagatingRoundTripper{
    base:          baseTransport,
    isHealthCheck: healthcontext.IsHealthCheck(ctx),
}
```

Open question for implementation: the existing comment at `pkg/vmcp/client/client.go:314-318` documents that mcp-go's `Close()` issues a DELETE built from `context.Background()` which loses identity. The current capture-at-construction behavior was intentional for that path. With the fix, that DELETE will arrive without identity. Two options:

1. Skip identity injection for the streamable-HTTP DELETE (it's a session-teardown request; upstream auth on a DELETE may not be needed).
2. Capture only what is strictly needed for teardown (e.g., a token specifically for the DELETE) and inject it only on requests where `req.Context()` carries no identity. This preserves teardown while letting normal requests use the fresh per-request identity.

Whichever path is chosen, the invariant must be: **a non-nil identity in `req.Context()` is never silently overridden by a captured one**.

## Reproduction

1. Configure a vMCP with one or more `MCPServerEntry` backends using `upstreamInject` outgoing auth, against an upstream IDP whose access tokens expire in less than 24 hours (most providers default to 1h–24h).
2. Authenticate through vMCP from an MCP client and exercise a tool against the upstream backend (succeeds).
3. Keep the session open and idle past the upstream's access-token lifespan. Do not restart the client or close the session.
4. Make another tool call to the same backend. **Expected**: vMCP refreshes the upstream token transparently and the call succeeds. **Actual**: the captured AT is used; the upstream returns 401; the call fails.

A targeted unit test on `identityPropagatingRoundTripper.RoundTrip` should assert that an identity placed on `req.Context()` is preserved into the outgoing request and is **not** replaced by a different identity captured at construction time.

## Acceptance criteria

- [ ] `identityPropagatingRoundTripper` no longer holds a captured `*auth.Identity` field, OR if it does, that field is never used to override an identity already present on `req.Context()`.
- [ ] A fresh identity placed on `req.Context()` by `TokenValidator.Middleware` reaches `UpstreamInjectStrategy.Authenticate` / `TokenExchangeStrategy.Authenticate` / `AWSSTSStrategy.Authenticate` unchanged.
- [ ] The streamable-HTTP `Close()`/DELETE path (where mcp-go uses `context.Background()`) continues to function without panicking or erroring on missing identity.
- [ ] Unit tests cover: (a) identity present in `req.Context()` is propagated, (b) identity absent in `req.Context()` is handled gracefully for the DELETE/teardown path, (c) `isHealthCheck` marker behavior is preserved.
- [ ] Integration test that idles a vMCP session past the upstream AT lifespan and asserts the next tool call succeeds via transparent refresh.

## Related

- #3877 — Phase 2 follow-up (this is essentially Part 1 of that issue, filed separately for prioritization).
- #3869 — Phase 1 reactive retry safety net.
- THV-0038 — Session-scoped client lifecycle RFC.

tgrunnagle:
## Summary

- vMCP backend HTTP clients captured the request `*auth.Identity` once at session/client-creation time and re-injected that snapshot on every outgoing backend request, overriding the fresh identity placed on `req.Context()` by `auth.TokenValidator.Middleware`. Because that fresh identity carries upstream tokens transparently refreshed by `upstreamtoken.InProcessService.GetAllValidTokens`, the override silently re-injected stale upstream access tokens for the entire session lifetime — users saw a forced re-auth roughly every 24h (the upstream provider's access-token lifespan).
- The captured identity is now treated as a fallback: it is injected only when `req.Context()` carries no identity. This preserves the streamable-HTTP `Close()`/DELETE teardown path — mcp-go constructs that request from `context.Background()`, which loses the per-request identity — while letting normal backend calls use the freshly refreshed identity from the request context.
- The fix is applied to both transport chains that exhibited the bug: `identityPropagatingRoundTripper` in `pkg/vmcp/client/client.go` (per-call client) and the new `identityRoundTripper` in `pkg/vmcp/session/internal/backend/mcp_session.go` (persistent session-backed client). Existing health-check marker propagation in the per-call client is preserved.

Closes #5323

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Unit-test coverage added/updated:

- `pkg/vmcp/client/client_test.go` — round-tripper tests reorganized into Per-request / Fallback / Health-check sections; new cases assert that a fresh identity on `req.Context()` is preserved unchanged, that the fallback is injected only when the request context has no identity, and that the health-check marker behavior is unchanged.
- `pkg/vmcp/session/internal/backend/roundtripper_test.go` — mirrors the above for the session-backed `identityRoundTripper`.
- `pkg/vmcp/session/internal/backend/mcp_session_identity_refresh_test.go` (new) — higher-level regression test that exercises the full transport chain (`headerForwardRoundTripper` → `identityRoundTripper` → `authRoundTripper` with `UpstreamInjectStrategy` → `http.DefaultTransport`) against a fake backend. Asserts that the upstream `Authorization` header on each `tools/call` reflects the identity placed on the per-request context, NOT the identity captured at session-init time. Verified to fail under the pre-fix always-override behavior, so it is a true #5323 regression guard.

Manual verification: traced through `auth.TokenValidator.Middleware` → `InProcessService.GetAllValidTokens` → vMCP outgoing-auth strategies (`upstream_inject`, `tokenexchange`, `aws_sts`) to confirm that all three read identity from `req.Context()` and therefore benefit from the fix.

## API Compatibility

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/client/client.go` | `identityPropagatingRoundTripper` no longer overrides identity on `req.Context()`; the captured identity is now a fallback used only when the request context carries none. Doc comments rewritten to document the invariant. |
| `pkg/vmcp/session/internal/backend/mcp_session.go` | New `identityRoundTripper` with the same fallback-only semantics, inserted into the session-backed transport chain. Documents asymmetric health-check handling (not needed in this chain). |
| `pkg/vmcp/client/client_test.go` | Reorganized round-tripper tests; new cases for per-request identity preservation, fallback-only injection, and health-check marker behavior. |
| `pkg/vmcp/session/internal/backend/roundtripper_test.go` | Same coverage for the session-backed round-tripper. |
| `pkg/vmcp/session/internal/backend/mcp_session_identity_refresh_test.go` | New higher-level regression test exercising the full transport chain end-to-end. |

Yes. Users of vMCP backends configured with `upstreamInject`, `tokenExchange`, or `awsSTS` outgoing auth no longer see forced re-auth when the upstream provider's access token expires mid-session. Upstream tokens are transparently refreshed on every tool call as long as the vMCP-issued refresh token remains valid.

- The fix is intentionally scoped to the round-tripper layer. The broader Phase 2 work tracked in #3877 (proactive refresh, session-scoped client lifecycle per THV-0038) and the reactive 401/403 retry safety net in #3869 remain open and unchanged by this PR. This change is essentially Part 1 of #3877, filed separately because it has a concrete user-visible impact worth landing on its own.
- Two duplicate round-tripper implementations now exist (one in `pkg/vmcp/client`, one in `pkg/vmcp/session/internal/backend`). They share the same invariant and were left duplicated to keep this PR focused on the bug fix. Consolidation is tracked in follow-up issue #5333.
- The higher-level regression test stops short of wiring `auth.TokenValidator.Middleware` end-to-end (which would pull in `pkg/auth/upstreamtoken` and a JWT signer) and instead injects the fresh identity directly onto the per-request context — which is exactly what the middleware does on every authenticated request. Extending the test to literally wire the middleware + a fake `UpstreamTokenRefresher` (AC5 as written on the issue) is tracked in follow-up issue #5334.
- Invariant to preserve in future changes: **a non-nil identity in `req.Context()` must never be silently overridden by a captured one**. The doc comments on both round-trippers state this explicitly.

### DIFF
--- a/pkg/vmcp/client/client.go
+++ b/pkg/vmcp/client/client.go
@@ -177,31 +177,68 @@ func (f roundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error) {
 	return f(req)
 }
 
-// identityPropagatingRoundTripper propagates identity and health-check markers to backend HTTP requests.
-// This ensures that identity information from the vMCP handler is available for authentication
-// strategies that need it (e.g., token exchange).
+// identityPropagatingRoundTripper propagates the health-check marker and a
+// fallback identity to backend HTTP requests.
 //
-// The health-check marker is stored at transport creation time and re-injected into every
-// outgoing request, including the DELETE that mcp-go sends when closing a streamable-HTTP
-// session. Without this, mcp-go's Close() creates a fresh context.Background()-based request
-// that loses the health-check marker, causing auth strategies (UpstreamInjectStrategy,
-// TokenExchangeStrategy) to fail with "no identity found in context".
+// Identity invariant: an identity already present on req.Context() is NEVER
+// overridden. The per-request identity placed on the request context by
+// auth.TokenValidator.Middleware carries the freshest upstream tokens (the
+// middleware refreshes them transparently on every incoming request via
+// upstreamtoken.InProcessService.GetAllValidTokens). Overriding it with a
+// snapshot captured at session-init time would silently re-inject stale
+// upstream access tokens on every backend call, forcing users to re-auth
+// once the captured access token expired (see issue #5323).
+//
+// The fallback identity is used only when req.Context() carries no identity
+// at all. This covers mcp-go's streamable-HTTP Close() path, which constructs
+// its DELETE request from context.Background() and therefore loses any
+// identity attached to the original tool-call context. Without a fallback,
+// auth strategies (UpstreamInjectStrategy, TokenExchangeStrategy, AWSSTSStrategy)
+// would reject the teardown DELETE with "no identity found in context" — the
+// teardown would still complete locally, but the upstream backend would see
+// an unauthenticated DELETE. The fallback may itself be stale, which is
+// acceptable for a session-close: the request is best-effort and any 401
+// from the upstream does not affect functional behavior.
+//
+// The health-check marker is similarly captured at transport creation time and
+// re-injected into every outgoing request — including the Close() DELETE — so
+// that auth strategies correctly skip authentication for health probes even
+// when the request context has been replaced with context.Background().
 type identityPropagatingRoundTripper struct {
-	base          http.RoundTripper
-	identity      *auth.Identity
-	isHealthCheck bool
+	base http.RoundTripper
+	// fallbackIdentity is injected only when req.Context() carries no identity.
+	// It must never override a non-nil identity present on the request context.
+	fallbackIdentity *auth.Identity
+	isHealthCheck    bool
 }
 
-// RoundTrip implements http.RoundTripper by adding identity and health-check marker to the request context.
+// RoundTrip implements http.RoundTripper.
+//
+// It preserves any identity already present on req.Context() (the fresh,
+// per-request identity placed there by TokenValidator.Middleware). When the
+// request context carries no identity — for example, mcp-go's Close() DELETE
+// is constructed from context.Background() — the captured fallback identity
+// is injected so session-teardown requests can still authenticate. The
+// health-check marker is always re-injected when configured.
 func (i *identityPropagatingRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	ctx := req.Context()
-	if i.identity != nil {
-		ctx = auth.WithIdentity(ctx, i.identity)
+	mutated := false
+
+	// Inject the fallback identity ONLY when the request context carries no
+	// identity. We must never overwrite a non-nil identity already on ctx —
+	// doing so would clobber the freshly refreshed upstream tokens that the
+	// auth middleware places on every incoming request (see #5323).
+	if _, hasIdentity := auth.IdentityFromContext(ctx); !hasIdentity && i.fallbackIdentity != nil {
+		ctx = auth.WithIdentity(ctx, i.fallbackIdentity)
+		mutated = true
 	}
+
 	if i.isHealthCheck {
 		ctx = healthcontext.WithHealthCheckMarker(ctx)
+		mutated = true
 	}
-	if i.identity != nil || i.isHealthCheck {
+
+	if mutated {
 		req = req.Clone(ctx)
 	}
 	return i.base.RoundTrip(req)
@@ -314,16 +351,22 @@ func (h *httpBackendClient) defaultClientFactory(ctx context.Context, target *vm
 		return nil, fmt.Errorf("failed to build header-forward transport: %w", err)
 	}
 
-	// Extract identity and health-check marker from context and propagate them to backend
-	// requests. The health-check marker must be carried through to the DELETE request that
-	// mcp-go emits when closing a streamable-HTTP session: mcp-go creates that request with
-	// context.Background(), which loses both the identity and the health-check marker that
-	// were present on the original ListCapabilities call context.
-	identity, _ := auth.IdentityFromContext(ctx)
+	// Capture a fallback identity and the health-check marker from the construction
+	// context. These are used ONLY when the per-request context lacks them:
+	//
+	//   - Normal backend requests inherit the fresh, per-request identity placed on
+	//     the request context by auth.TokenValidator.Middleware. The transport
+	//     preserves it untouched so upstream-token refresh is not bypassed (#5323).
+	//
+	//   - mcp-go's streamable-HTTP Close() builds its DELETE from context.Background(),
+	//     which loses both the identity and the health-check marker. Re-injecting them
+	//     here keeps the teardown DELETE authenticated and tagged as a health-check
+	//     probe when the original session was a probe (#4613).
+	fallbackIdentity, _ := auth.IdentityFromContext(ctx)
 	baseTransport = &identityPropagatingRoundTripper{
-		base:          baseTransport,
-		identity:      identity,
-		isHealthCheck: healthcontext.IsHealthCheck(ctx),
+		base:             baseTransport,
+		fallbackIdentity: fallbackIdentity,
+		isHealthCheck:    healthcontext.IsHealthCheck(ctx),
 	}
 
 	// Inject W3C Trace Context headers (traceparent/tracestate) into outgoing requests.
--- a/pkg/vmcp/client/client.go
+++ b/pkg/vmcp/client/client.go
@@ -204,10 +204,21 @@ func (f roundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error) {
 // re-injected into every outgoing request — including the Close() DELETE — so
 // that auth strategies correctly skip authentication for health probes even
 // when the request context has been replaced with context.Background().
+//
+// This is the canonical location for the #5323 fallback-only identity
+// invariant. A near-clone exists at identityRoundTripper in
+// pkg/vmcp/session/internal/backend/mcp_session.go (without isHealthCheck
+// support, since health probes do not flow through the session-backed
+// connector). Keep the invariant in sync across both implementations until
+// #5333 lands a shared transport.
 type identityPropagatingRoundTripper struct {
 	base http.RoundTripper
 	// fallbackIdentity is injected only when req.Context() carries no identity.
 	// It must never override a non-nil identity present on the request context.
+	//
+	// Shared across all concurrent RoundTrip invocations on this transport; the
+	// pointed-to *auth.Identity (including UpstreamTokens) MUST be treated as
+	// immutable — see pkg/auth/identity.go.
 	fallbackIdentity *auth.Identity
 	isHealthCheck    bool
 }
--- a/pkg/vmcp/client/client.go
+++ b/pkg/vmcp/client/client.go
@@ -368,13 +368,20 @@ func (h *httpBackendClient) defaultClientFactory(ctx context.Context, target *vm
 	// Scope note: httpBackendClient calls this factory on every CallTool /
 	// ReadResource / GetPrompt / ListCapabilities with the per-call context, so
 	// the captured "fallback" is per-call, not per-session, in this path. The
-	// fallback is therefore functionally only required to keep mcp-go's
-	// Close() DELETE — which is built from context.Background() and loses any
-	// identity attached to the original call context — authenticated. The
-	// always-stale capture-at-session-init behavior that caused #5323 lives
-	// in the persistent-session path (pkg/vmcp/session/internal/backend); this
+	// always-stale capture-at-session-init behavior that caused #5323 lives in
+	// the persistent-session path (pkg/vmcp/session/internal/backend); this
 	// transport uses the same RoundTripper out of consistency, not because it
 	// suffered the same bug.
+	//
+	// Fallback injection scope: the round-tripper injects the captured fallback
+	// identity on ANY outgoing request whose context lacks an identity — not
+	// only the teardown DELETE. In the current code paths the only known caller
+	// that drops the per-request identity is mcp-go's streamable-HTTP Close()
+	// (which builds its DELETE from context.Background()), so in practice the
+	// fallback only fires for teardown. If a future code path wraps the client
+	// with a context that doesn't carry identity (audit, telemetry, background
+	// reconciliation), it will be silently authenticated using the captured
+	// snapshot. Add an explicit method/URL gate if that becomes undesirable.
 	//
 	//   - Normal backend requests inherit the fresh, per-request identity placed on
 	//     the request context by auth.TokenValidator.Middleware. The transport
--- a/pkg/vmcp/client/client.go
+++ b/pkg/vmcp/client/client.go
@@ -352,7 +352,18 @@ func (h *httpBackendClient) defaultClientFactory(ctx context.Context, target *vm
 	}
 
 	// Capture a fallback identity and the health-check marker from the construction
-	// context. These are used ONLY when the per-request context lacks them:
+	// context. These are used ONLY when the per-request context lacks them.
+	//
+	// Scope note: httpBackendClient calls this factory on every CallTool /
+	// ReadResource / GetPrompt / ListCapabilities with the per-call context, so
+	// the captured "fallback" is per-call, not per-session, in this path. The
+	// fallback is therefore functionally only required to keep mcp-go's
+	// Close() DELETE — which is built from context.Background() and loses any
+	// identity attached to the original call context — authenticated. The
+	// always-stale capture-at-session-init behavior that caused #5323 lives
+	// in the persistent-session path (pkg/vmcp/session/internal/backend); this
+	// transport uses the same RoundTripper out of consistency, not because it
+	// suffered the same bug.
 	//
 	//   - Normal backend requests inherit the fresh, per-request identity placed on
 	//     the request context by auth.TokenValidator.Middleware. The transport
--- a/pkg/vmcp/client/client_test.go
+++ b/pkg/vmcp/client/client_test.go
@@ -1055,15 +1055,8 @@ func TestWrapBackendError(t *testing.T) {
 // identityPropagatingRoundTripper
 // ---------------------------------------------------------------------------
 //
-// Design invariant (see issue #5323): the per-request identity placed on
-// req.Context() by auth.TokenValidator.Middleware is THE canonical source of
-// identity for outgoing backend requests. It carries upstream tokens that the
-// middleware transparently refreshes per request. The captured fallback
-// identity is a teardown-DELETE escape hatch only — it covers the case where
-// mcp-go's Close() builds its DELETE from context.Background() and loses the
-// per-request identity. The fallback MUST NEVER override a non-nil identity
-// already present on req.Context(); doing so silently re-injects stale tokens
-// and is the bug fixed by #5323.
+// See identityPropagatingRoundTripper in pkg/vmcp/client/client.go for the
+// canonical description of the #5323 fallback-only identity invariant.
 //
 // Tests below are grouped to reflect this hierarchy:
 //   1. Per-request identity (normal path)            — *_PerRequestIdentity_*
--- a/pkg/vmcp/client/client_test.go
+++ b/pkg/vmcp/client/client_test.go
@@ -1055,12 +1055,15 @@ func TestWrapBackendError(t *testing.T) {
 // identityPropagatingRoundTripper
 // ---------------------------------------------------------------------------
 
-func TestIdentityPropagatingRoundTripper_WithIdentity_PropagatesIdentityInContext(t *testing.T) {
+// TestIdentityPropagatingRoundTripper_FallbackIdentity_InjectedWhenContextLacksIdentity
+// verifies that the fallback identity is injected when req.Context() carries no
+// identity (e.g. mcp-go's Close() DELETE built from context.Background()).
+func TestIdentityPropagatingRoundTripper_FallbackIdentity_InjectedWhenContextLacksIdentity(t *testing.T) {
 	t.Parallel()
 
 	base := &mockRoundTripper{response: &http.Response{StatusCode: http.StatusOK}}
-	identity := &pkgauth.Identity{PrincipalInfo: pkgauth.PrincipalInfo{Subject: "user-1"}}
-	rt := &identityPropagatingRoundTripper{base: base, identity: identity}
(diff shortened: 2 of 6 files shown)


tgrunnagle (commented):
## Multi-agent review summary

Reviewed against the repo's `.claude/rules/` (go-style, vmcp-anti-patterns, security, testing, pr-creation) using 5 specialist agents (auth/security, concurrency, Go code quality, vMCP architecture, general). Codex cross-review skipped (CLI not installed).

**Bottom line:** the fix is correct and well-tested at the layer it targets. No HIGH findings. Six pre-merge polish items improve clarity, robustness, and drift-resistance.

### Consensus findings (score ≥ 7)

| # | Severity | Where | Title |
|---|----------|-------|-------|
| 1 | MEDIUM | `mcp_session_identity_refresh_test.go` | "Maintenance note" + Initialize assertion is a lint-driven test smell |
| 2 | MEDIUM | both round-trippers | Duplicate implementation risks drift; add cross-reference comments now |
| 3 | MEDIUM | `client.go` factory comment | Doc claim understates fallback scope (injects on any context lacking identity) |
| 4 | LOW | `pkg/auth/context.go` (not in diff) | `IdentityFromContext` returns `ok=true` for typed-nil `*Identity` |
| 5 | LOW | 4 files | Identity-invariant prose duplicated across 4 files (drift risk) |
| 6 | LOW | both round-tripper structs | Document immutability invariant on the captured `*Identity` |

Details on #1, #2, #3, #5, #6 are inline. **#4** is independent of this PR's diff so noted here:

> `pkg/auth/context.go:51` — `IdentityFromContext` does `ctx.Value(...).(*Identity)` and returns `(nil, true)` if a caller stored a typed-nil `*Identity` via raw `context.WithValue`. The new round-tripper guards `if !hasIdentity && fallbackIdentity != nil` so a typed-nil would suppress the fallback **and** pass `nil` to `UpstreamInjectStrategy`, which would nil-deref on `identity.UpstreamTokens[providerName]`. `WithIdentity` is nil-safe so this is latent rather than active. **Recommendation:** tighten to `return identity, ok && identity != nil`. One-line, defensive, no caller impact. Out of scope for this PR — file a follow-up if you agree.

### What's solid

- Identity invariant correctly inverted and consistently applied in both transport chains
- Regression test (`TestHTTPSession_PerRequestIdentity_DrivesUpstreamAuthHeader`) verified to fail under pre-fix behavior — true #5323 regression guard
- No new log statements, no credential leakage risk introduced
- No public API change; all renames on unexported types
- Doc comments accurately describe behavior; field rename `identity` → `fallbackIdentity` makes the invariant self-documenting
- Inner round-trippers (`authRoundTripper`, `headerForwardRoundTripper`) still clone before mutating, so the new "skip-clone-when-no-mutation" fast path is safe
- Test files correctly colocated with existing `fakeBackend` helpers
- The doc comment correctly notes WHY health-check propagation is absent in the session-backed connector (forward-pointer for future maintainers)

### Dropped findings (consensus < 7, listed for auditability)

- Mutable `mutated` flag violates immutable-assignment style (LOW, 6) — stylistic taste
- `sess.Close()` in `t.Cleanup` lacks a bounded path (LOW, 6) — in-process httptest, low practical risk
- Upstream mcp-go fix not tracked anywhere (LOW, 6) — fixing `Close()` to reuse session ctx upstream would obviate the fallback entirely
- Over-long "Scope note" comment on `defaultClientFactory` (LOW, 5) — specialists disagreed; appropriate for the bug's subtlety
- E2E test bypasses `TokenValidator.Middleware` (LOW, 6) — acknowledged in PR description and tracked as #5334
- Behavior change is a soft backwards-compat break (LOW, 5) — OOT callers relying on the buggy override
- Several INFO-level items (test naming, magic strings, 5s timeouts, line count vs CLAUDE.md cap)

### Process notes

- PR is in **draft** state. Per `.claude/rules/pr-creation.md` (["Reviewer trust signals"](https://github.com/stacklok/toolhive/blob/main/.claude/rules/pr-creation.md)), mark ready-for-review before requesting human review, or note in the description that draft is intentional pending follow-up.
- Net production code is small (~10 lines of real logic + comments + renames). The PR's line count is dominated by tests and doc comments, well under the 400-line cap for non-test/non-doc changes.

pkg/vmcp/session/internal/backend/mcp_session_identity_refresh_test.go
```diff
@@ -0,0 +1,200 @@
...
+	// This documents the captured fallback's role at session-init time.
+	//
+	// Maintenance note: do not delete this block without also varying the
+	// other headersFor call site below. fakeBackend.headersFor takes a
+	// `method` parameter; if both call sites end up passing the same
+	// constant (e.g. only tools/call), the `unparam` linter will fail
+	// because the parameter becomes effectively dead. This Initialize
+	// assertion is therefore load-bearing for lint as well as for test
+	// documentation — keep at least one non-tools/call invocation.
+	initHeaders := fb.headersFor(string(mcp.MethodInitialize))
```
tgrunnagle: **[MEDIUM] Lint-driven test smell.**

This Initialize-headers assertion and the 7-line "Maintenance note" above exist solely to keep `unparam` from flagging `fakeBackend.headersFor`'s `method` parameter as effectively constant. A test that requires a paragraph explaining *why it exists* is doing two unrelated jobs (proving the #5323 regression AND propping up a linter rule). The lint constraint has leaked into the test's design — a future contributor who deletes this block to focus the test will trip a lint failure that has nothing to do with the test's purpose.

**Recommendation:** drop the Initialize assertion and the maintenance note. If `unparam` does flag `headersFor`, address it at the helper:
- (a) add `//nolint:unparam` on `headersFor` with a one-line justification (it's a test helper that legitimately accepts a method name even if currently only one method is inspected at most call sites), or
- (b) rename to `lastHeaders()` / `headersForToolsCall()` and add a parameterized variant only when a second method's headers actually need inspecting.

Test code should not be shaped around a linter.

pkg/vmcp/client/client.go
```diff
@@ -177,31 +177,68 @@ func (f roundTripperFunc) RoundTrip(req *http.Request) (*http.Response, error) {
...
+// teardown would still complete locally, but the upstream backend would see
+// an unauthenticated DELETE. The fallback may itself be stale, which is
+// acceptable for a session-close: the request is best-effort and any 401
+// from the upstream does not affect functional behavior.
+//
+// The health-check marker is similarly captured at transport creation time and
+// re-injected into every outgoing request — including the Close() DELETE — so
+// that auth strategies correctly skip authentication for health probes even
+// when the request context has been replaced with context.Background().
 type identityPropagatingRoundTripper struct {
```
tgrunnagle, line 214: **[MEDIUM] Add cross-reference to the twin in `pkg/vmcp/session/internal/backend/mcp_session.go`.**

This struct and `identityRoundTripper` in `mcp_session.go` are now near-clones of the same fallback-only invariant, with subtly different shapes (this one has `isHealthCheck`; the twin doesn't). Consolidation is deferred to #5333 — fine — but landing two copies of the invariant at the moment the invariant is being tightened is the highest-risk window for drift.

The twin in `mcp_session.go` already cross-references this file via its health-check-absent note. Adding the symmetric reverse-reference here makes the relationship discoverable from either side and reduces the chance that a future change updates only one site.

**Recommendation (small):** add one line to this doc comment, e.g.:
> `// Mirror of identityRoundTripper in pkg/vmcp/session/internal/backend/mcp_session.go — keep the fallback-only invariant in sync until #5333 lands.`

Also applicable to the immutability concern: the captured `*auth.Identity` is shared across all concurrent `RoundTrip` calls on this transport. Concurrent reads are race-free *only* because `Identity` is treated as immutable after construction (per `pkg/auth/identity.go`). Consider adding: `// fallbackIdentity is shared across all concurrent RoundTrips; the pointed-to Identity (including UpstreamTokens) MUST be treated as immutable.`

pkg/vmcp/client/client.go
```diff
@@ -314,16 +351,33 @@ func (h *httpBackendClient) defaultClientFactory(ctx context.Context, target *vm
 		return nil, fmt.Errorf("failed to build header-forward transport: %w", err)
 	}
 
-	// Extract identity and health-check marker from context and propagate them to backend
-	// requests. The health-check marker must be carried through to the DELETE request that
-	// mcp-go emits when closing a streamable-HTTP session: mcp-go creates that request with
-	// context.Background(), which loses both the identity and the health-check marker that
-	// were present on the original ListCapabilities call context.
-	identity, _ := auth.IdentityFromContext(ctx)
+	// Capture a fallback identity and the health-check marker from the construction
```
tgrunnagle, line 365: **[MEDIUM] Doc comment understates the actual fallback scope.**

The comment says the fallback is "functionally only required to keep mcp-go's `Close()` DELETE authenticated." The **code** does not enforce that scope: the fallback is injected on ANY outgoing request whose `req.Context()` lacks an identity, not just teardown DELETEs. Pre-fix the same always-injection happened (so this is not a regression), but the new doc now makes a precise-sounding claim about scope that is not actually checked.

Future readers will trust the comment and may rely on the scoping promise. If a new call path (audit, telemetry, background reconciliation) ever wraps the client with a context that doesn't carry identity, it will silently be authenticated using the captured snapshot instead of failing with a clean `"no identity found in context"`.

**Recommendation (pick one):**
- **(a) Enforce the scope in code:** gate the fallback injection on the teardown shape, e.g. `req.Method == http.MethodDelete` (and optionally a URL pattern). The doc's promise becomes load-bearing.
- **(b) Update the comment to match the code:** "The fallback is injected on ANY request whose context lacks an identity, not only teardown DELETEs. This is safe in current call paths only because mcp-go's Close() is the only known caller that drops identity."

(a) is more defensive and is recommended.

pkg/vmcp/client/client_test.go
```diff
@@ -1054,13 +1054,105 @@ func TestWrapBackendError(t *testing.T) {
 // ---------------------------------------------------------------------------
 // identityPropagatingRoundTripper
 // ---------------------------------------------------------------------------
+//
+// Design invariant (see issue #5323): the per-request identity placed on
```
tgrunnagle: **[LOW] Duplicated invariant prose across 4 files.**

This ~20-line "Design invariant" block is also present in `roundtripper_test.go` (header comment) and the doc comments on both `identityPropagatingRoundTripper` (this file's client.go) and `identityRoundTripper` (mcp_session.go). That's four near-copies that have to be kept in sync. `.claude/rules/go-style.md` ("Keep Comments Synchronized With Code") warns about exactly this: when the invariant evolves — e.g. when #5333 lands a shared helper, or when health-check propagation is unified — one of the copies will be missed and start contradicting the code.

**Recommendation:** pick one canonical location (e.g. `identityPropagatingRoundTripper`'s struct doc since it's the older and most thorough copy), and have the other three reference it by symbol name:

```
// See identityPropagatingRoundTripper in pkg/vmcp/client/client.go
// for the canonical invariant description (#5323).
```

This keeps the rationale discoverable without four-way drift.

pkg/vmcp/session/internal/backend/mcp_session.go
```diff
@@ -59,15 +59,42 @@ func (a *authRoundTripper) RoundTrip(req *http.Request) (*http.Response, error)
...
+// is acceptable for the teardown DELETE since the request is best-effort.
+//
+// No health-check marker is re-injected here. Health probes do not flow through
+// session-backed clients — they go through the per-call client in
+// pkg/vmcp/client (whose identityPropagatingRoundTripper IS health-check aware).
+// If a future change routes health probes through this connector, the
+// isHealthCheck propagation pattern from pkg/vmcp/client/client.go must be
+// mirrored here so the Close() DELETE built from context.Background() retains
+// the marker.
 type identityRoundTripper struct {
```
tgrunnagle, line 77: **[LOW] Document the immutability invariant on the captured pointer.**

`fallbackIdentity` is held for the lifetime of this round-tripper and is shared across all concurrent `RoundTrip` invocations on the transport. Concurrent reads of `UpstreamTokens` are race-free *only* because `*auth.Identity` is treated as immutable after construction — this is documented in `pkg/auth/identity.go` but not re-stated where the captured pointer actually lives.

A future maintainer who decides to "refresh the captured tokens in place" (which would seem natural given the #5323 bug being fixed) would silently introduce a data race that the test suite would not reliably catch.

**Recommendation:** add one sentence to this struct's doc comment, e.g.:

```
// fallbackIdentity is shared across all concurrent RoundTrip invocations on
// this transport; the pointed-to *auth.Identity (including UpstreamTokens)
// MUST be treated as immutable — see pkg/auth/identity.go.
```

Apply the symmetric note to `identityPropagatingRoundTripper` in `pkg/vmcp/client/client.go`.

pkg/vmcp/client/client.go
tgrunnagle: Addressed in 4bec90a8 — added a reverse cross-reference to the twin in `pkg/vmcp/session/internal/backend/mcp_session.go` and a concurrency/immutability note on `fallbackIdentity`. The same canonical-source pattern landed across the four prose copies (see #4 reply) so the canonical block lives only in `identityPropagatingRoundTripper`'s struct doc.

pkg/vmcp/session/internal/backend/mcp_session.go
tgrunnagle: Addressed in 4bec90a8 — added the symmetric immutability sentence on `fallbackIdentity` here and reduced the surrounding doc block to a reference to the canonical description in `identityPropagatingRoundTripper`.

pkg/vmcp/client/client_test.go
tgrunnagle: Addressed in 4bec90a8 — `identityPropagatingRoundTripper` in `pkg/vmcp/client/client.go` is now the canonical location for the #5323 invariant prose. The three duplicates (this file's header block, `roundtripper_test.go`'s header block, and `mcp_session.go`'s struct doc) now reference it by symbol, keeping per-location specifics (e.g. session-backed twin omits health-check propagation) inline.

pkg/vmcp/session/internal/backend/mcp_session_identity_refresh_test.go
tgrunnagle: Addressed in e6f45edf — dropped the Initialize-headers assertion and the 7-line maintenance note that propped up unparam. The shield now sits on `fakeBackend.headersFor` itself with a justification that the helper is method-agnostic by design. The test focuses purely on the #5323 regression invariant.

pkg/vmcp/client/client.go
tgrunnagle: Addressed in 76ad5271 with option (b) — updated the `Scope note` to honestly describe the broader actual behavior (fallback fires for any context-less request; mcp-go's `Close()` is the only known caller exercising that path today). Added an explicit warning that future code paths dropping identity would be silently authenticated by the captured snapshot unless a method/URL gate is added. Chose (b) over (a) to keep this PR focused on the #5323 regression fix — enforcing DELETE-only would be a behavior change and may break paths we don't know about.

tgrunnagle:
Re: body finding #4 (`IdentityFromContext` returns `(nil, true)` for typed-nil `*Identity`) — filed as #5337 for follow-up. Keeping it out of this PR to stay focused on the #5323 transport-layer regression; the hazard is latent (no in-tree caller currently stores a typed-nil) and the fix is a 1-line defensive tighten at the read site, well-scoped to its own PR.

jhrozek (approved):

PR #5343, opened 2026-05-20
Add CIMD storage decorator for embedded AS

amirejaz:
## Summary

Phase 2 PR 2 of CIMD embedded AS support — depends on PR #5320 (merged).

The embedded authorization server currently requires clients to call `/oauth/register` (DCR) before the authorization flow. With this PR, clients can present an HTTPS URL as `client_id` and the server fetches and validates the CIMD document on demand, with no prior registration.

- **`CIMDStorageDecorator`** wraps `storage.Storage`, overrides only `GetClient`, and delegates all other methods transparently. When fosite calls `GetClient("https://vscode.dev/oauth/client-metadata.json")`, the decorator fetches the document via `pkg/oauthproto/cimd.FetchClientMetadataDocument`, builds a `fosite.Client`, caches it with a configurable TTL via `hashicorp/golang-lru/v2`, and deduplicates concurrent fetches for the same URL via `singleflight`.
- **`Unwrap()`** exposes the underlying storage so the `DCRCredentialStore` and `RedisStorage` type assertions in `server_impl.go` still reach the concrete backend through the decorator layer.
- **`runLegacyMigration`** helper extracted from `newServer` to keep cyclomatic complexity under the lint limit after the Unwrap additions.

## Type of change

- [x] New feature (non-breaking)

## Test plan

- [x] Unit tests: `go test ./pkg/authserver/storage/...` — all pass including cache hit/miss, singleflight deduplication, TTL expiry, `Unwrap`, disabled path, DCR client unaffected
- [x] Full authserver suite: `go test ./pkg/authserver/...` — 13 packages pass

- `NewCIMDStorageDecorator` returns `base` unchanged when `enabled=false` — no allocation, no behavior change for existing deployments
- `LoopbackClient` wraps clients whose `redirect_uris` contain loopback addresses, enabling RFC 8252 §7.3 dynamic port matching for native app clients


pkg/authserver/storage/cimd_decorator.go
```diff
@@ -0,0 +1,210 @@
...
+
+	responseTypes := doc.ResponseTypes
+	if len(responseTypes) == 0 {
+		responseTypes = defaultCIMDResponseTypes
+	}
+
+	// The embedded AS advertises only "none" in token_endpoint_auth_methods_supported.
+	// Override any other value from the document to avoid an inconsistent client
+	// configuration — CIMD clients are always public and have no pre-shared secret.
+	tokenEndpointAuthMethod := defaultCIMDTokenEndpointAuthMethod
```
jhrozek: In `buildFositeClient`, `TokenEndpointAuthMethod` is set to `"none"` unconditionally, but the upstream validator in `pkg/oauthproto/cimd/fetch.go` only rejects the symmetric methods — `private_key_jwt`, `tls_client_auth`, and similar asymmetric methods pass validation and then get silently rewritten here.

Is the silent rewrite the intended behavior? Rather than accepting a doc that says "I'll use `private_key_jwt`" and then running as if it had said `"none"`, would it be cleaner to reject the registration when the method isn't one the AS actually supports? Happy to be wrong on this — just want to make sure it's a conscious call.

Also: the doc comment on `defaultCIMDTokenEndpointAuthMethod` (line 139) says the constant is "applied when the CIMD document omits `token_endpoint_auth_method`", but in practice it's applied unconditionally. If you keep the override behavior, that comment needs to match what the code does; if you switch to rejecting non-`none`, the comment becomes accurate.

jhrozek (commented):
One question on the token_endpoint_auth_method handling.

pkg/authserver/storage/cimd_decorator.go
amirejaz: Good catch — went with rejection. `fetch()` now returns an error if `token_endpoint_auth_method` is anything other than `"none"`, so a doc claiming `private_key_jwt` fails fast instead of being silently accepted as public. The constant comment is updated to match.

jhrozek (approved):

Issue #5329, opened 2026-05-19
Admit obo in MCPExternalAuthConfig CRD enum and regenerate manifests


tgrunnagle:
## Description

Light up apiserver-level admission of `obo`-typed `MCPExternalAuthConfig` resources. Append `;obo` to the kubebuilder enum marker, declare an opaque `OBOConfig` placeholder struct + `OBO *OBOConfig` spec field, add a CEL rule asserting the field is set iff the type is `obo`, add the matching defense-in-depth check inside `validateTypeConfigConsistency`, add a structural-only no-op arm to the `Validate()` switch, and regenerate both the operator CRD chart YAML and the public CRD reference docs.

This is the moment the feature becomes user-visible. Until this task merges, the apiserver enum on `MCPExternalAuthConfig.spec.type` does not list `obo`, so `kubectl apply` of an `obo`-typed CR is rejected by admission before any of the previously-landed dispatch wiring runs. After this task merges, the apiserver admits the CR, the reconciler routes it through the function-pointer hook, and an upstream-only build surfaces `status.conditions[Valid] = False` / `Reason: EnterpriseRequired` end-to-end.

## Context

This is the fourth and final flat technical task under the parent OSS story. The parent establishes the function-pointer override hook design for a new `obo` external auth type. The earlier tasks landed dark infrastructure: the Go constant `ExternalAuthTypeOBO`, the default `oboHandler` returning `ErrEnterpriseRequired`, the new `pkg/auth/obo/` package with `RegisterFactory`, the `OBOConverterStub` registered in `NewRegistry()`, the two operator-side switch arms, and the reconcile-loop branch that maps `ErrEnterpriseRequired` to `Reason: EnterpriseRequired`. None of that activated user-facing behavior because the CRD enum did not admit `obo`.

This task closes the loop with a contained, surgical edit to one Go source file plus regenerated manifests:

1. **Append `;obo`** to the `+kubebuilder:validation:Enum` marker on `MCPExternalAuthConfigSpec.Type` (existing marker lists seven values).
2. **Declare an opaque `OBOConfig` placeholder struct** at the top level of the file (near the other `*Config` structs). The inner schema is intentionally empty in this revision; sub-fields land in a follow-up.
3. **Add `OBO *OBOConfig`** to `MCPExternalAuthConfigSpec` with the standard `json:"obo,omitempty"` tag (alongside the existing seven optional config fields).
[...]

Issue #5327, opened 2026-05-19
Add default OBO handler hooks and vMCP/proxy converter stubs


tgrunnagle:
## Description

Stand up the default stub infrastructure for a new `obo` external auth type at the three dispatch layers — operator, vMCP, and proxy runtime — plus the Go constant the other layers reference. After this task, an out-of-tree build can register a real OBO implementation by calling three setter functions; an upstream-only build hits the defaults and surfaces a sentinel error at every dispatch point.

This task does **not** make the CRD admit `obo` typed configs. The kubebuilder enum marker, CEL rule, `OBOConfig` struct, and regenerated YAML land in a later task. Until that task merges, the constant, handler hooks, middleware-factory entry, and converter stub live as dark infrastructure that activates the moment the CRD enum is widened.

## Context

This is the second of four flat technical tasks under the parent OSS story. The parent establishes the function-pointer override hook design: a small, single-slot, last-write-wins setter at each dispatch point that defaults to producing a sentinel error and can be replaced from an `init()` in an out-of-tree package.

What lands here:

1. **The Go constant** — `ExternalAuthTypeOBO ExternalAuthType = "obo"` in the CRD types file, alongside the existing seven type constants. The constant is needed before the dispatch wiring task so the switch arms compile.
2. **The operator-time handler hook** — `OBOHandler` struct, package-level default value returning `ErrEnterpriseRequired`, `RegisterOBOHandler` setter, and exported `OBOValidate` / `OBOSecretEnvVars` wrappers in `cmd/thv-operator/pkg/controllerutil/tokenexchange.go` (filename unchanged — the file picks up these new exports alongside its existing token-exchange helpers).
3. **The new `pkg/auth/obo/` package** — `MiddlewareType = "obo"` constant, a `CreateMiddleware` factory variable matching the canonical `pkg/auth/awssts/middleware.go:56` signature, and a `RegisterFactory` setter. The default factory produces a `Middleware` whose `Handler()` responds HTTP 503 to every request with a body indicating the OBO middleware factory must be registered.
4. **The middleware-factory map entry** — one new line in the literal map at `pkg/runner/middleware.go:32-47` (16 entries today → 17 after this task) wiring `obo.MiddlewareType` to `obo.CreateMiddleware`.
[...]

Issue #5325, opened 2026-05-19
[Entra] Add `obo` external auth type and function-pointer override hooks


tgrunnagle:
## User Story

**As a** ToolHive operator maintainer,
**I want to** add an `obo` value to the `MCPExternalAuthConfig.spec.type` enum and stand up function-pointer override hooks at each dispatch point (operator handler, vMCP strategy converter, proxy middleware factory),
**so that** out-of-tree builds can register a real on-behalf-of (OBO) implementation by replacing the default stubs, while an upstream-only build admits the new CRD value, reports a clear `EnterpriseRequired` status, and never falls through to a generic "unsupported external auth type" or "unknown middleware type" error.

## Context

This story is the upstream foundation for a new `ExternalAuthType` value, `obo`, in the `MCPExternalAuthConfig` CRD. The name refers to the OAuth 2.0 on-behalf-of pattern (RFC 8693 token exchange and IdP-specific variants such as Microsoft Entra's OBO flow): a proxy that receives a user's access token exchanges it for a downstream token scoped to the same user identity, so per-user identity survives backend egress.

The actual OBO protocol implementation (IdP-specific token-exchange grants, user-token cache, refresh-token rotation, audit-event schema) is **not** part of this story — it is deferred to a follow-up RFC. What lands here is the surface plus default stubs:

- A new `ExternalAuthType` constant `ExternalAuthTypeOBO = "obo"`, admitted by the CRD enum.
- An opaque `OBOConfig` placeholder struct on `MCPExternalAuthConfigSpec` (no inner fields yet — they ship in the follow-up).
- A single-slot function-pointer override hook at each of three dispatch points:
  1. **Operator-time** — `controllerutil.OBOHandler` (a struct of `Validate` / `ApplyRunConfig` / `SecretEnvVars` function pointers) plus a `RegisterOBOHandler` setter. The default `oboHandler` returns a new exported `ErrEnterpriseRequired` sentinel from all three methods.
  2. **vMCP** — `OBOConverterStub` registered in `(converters.Registry).NewRegistry()` as a built-in. The stub's `Convert` returns an error wrapping `ErrEnterpriseRequired`.
[...]

PR #5345, opened 2026-05-20
Wire OBO dispatch arms and reconciler branch

### LINKED ISSUE #5328: Wire OBO dispatch arms and reconciler branch; add integration tests
## Description

Wire the new `obo` external auth type into the three operator dispatch sites and prove the wiring end-to-end with envtest integration tests. After this task, applying an `obo`-typed `MCPExternalAuthConfig` from any of the three consumer CRs (`MCPServer`, `MCPRemoteProxy`, `VirtualMCPServer`) surfaces a uniform `status.conditions[Valid] = False` / `Reason: EnterpriseRequired` outcome instead of the generic "unsupported external auth type" or "unknown middleware type" error that would otherwise leak from the `default:` arms of the existing switches.

This is purely the dispatch + test layer. The default stubs and exported symbols this task consumes (`OBOHandler`, `RegisterOBOHandler`, `OBOValidate`, `OBOSecretEnvVars`, `ErrEnterpriseRequired`, `MiddlewareType`, the registered `OBOConverterStub`) all land in the dependency task and exist at compile time before this task starts.

## Context

This is the third of four flat technical tasks under the parent OSS story. The parent story establishes the function-pointer override hook design for `obo`. The first task factors the operator and proxyrunner mains. The second task ("default handler infrastructure") creates the default-stub package surface this task wires into.

What lands here:

1. **Two operator-side switch arms** — one new `case mcpv1beta1.ExternalAuthTypeOBO` arm placed before the `default:` in each of the two existing `switch externalAuthConfig.Spec.Type` blocks. Both arms route through the function-pointer hook from the dependency task; neither falls through to the `default:`'s "unsupported external auth type" path.
2. **One new reconcile-loop branch** — a branch in `cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go` that invokes `controllerutil.OBOValidate` and maps `ErrEnterpriseRequired` to `status.conditions[Valid] = False` / `Reason: EnterpriseRequired` via `controllerutil.MutateAndPatchStatus`. The exact reason string `EnterpriseRequired` is part of the user-facing contract — external consumers pattern-match on it.
3. **Integration tests** — applying an `obo`-typed `MCPExternalAuthConfig` to an envtest cluster, referenced from each of the three consumer CR types, and asserting the `Valid=False` / `EnterpriseRequired` condition propagates correctly. A separate guard asserts neither the reconciler logs nor the resource status surface a generic "unsupported external auth type" or "unknown middleware type" string.

**Dependencies**: #5325 (parent story), #5327 (consumes `OBOHandler`, `RegisterOBOHandler`, `OBOValidate`, `OBOSecretEnvVars`, `ErrEnterpriseRequired`, `MiddlewareType`, and the registered `OBOConverterStub` — none of these symbols exist before #5327 merges).
**Blocks**: #5329 (CRD enum admission — the LAST OSS task, which lights up apiserver-level admission of `obo` typed configs).

## Acceptance Criteria

Code surface:

- [ ] In `cmd/thv-operator/pkg/controllerutil/tokenexchange.go` `AddExternalAuthConfigOptions`, the switch at line 101 admits `case mcpv1beta1.ExternalAuthTypeOBO` as a new arm placed before the `default:`. The seven existing arms (`tokenExchange`, `headerInjection`, `bearerToken`, `unauthenticated`, `embeddedAuthServer`, `awsSts`, `upstreamInject`) and the `default:` arm are byte-identical to before.
- [ ] In `cmd/thv-operator/controllers/virtualmcpserver_deployment.go` `(*VirtualMCPServerReconciler).getExternalAuthConfigSecretEnvVar`, the switch at line 789 admits `case mcpv1beta1.ExternalAuthTypeOBO` as a new arm placed before the `default:`. The seven existing arms and the `default:` arm are byte-identical to before.
- [ ] In `cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go`, a new branch checks `cfg.Spec.Type == mcpv1beta1.ExternalAuthTypeOBO`, calls `controllerutil.OBOValidate(cfg)`, and on error sets `status.conditions[Valid] = False` via `controllerutil.MutateAndPatchStatus` (NOT `r.Status().Update`). The reason string is exactly `EnterpriseRequired` when `errors.Is(err, controllerutil.ErrEnterpriseRequired)` is true, and exactly `InvalidConfig` for any other error.
- [ ] The existing `r.Status().Update(ctx, externalAuthConfig)` call sites at lines 108 and 137 of the reconciler are unchanged. (Refactoring them is explicitly out of scope — see "Out of Scope" below.)

Behavior:

- [ ] Applying an `obo`-typed `MCPExternalAuthConfig` referenced from an `MCPServer` causes the `MCPExternalAuthConfig` resource to transition to `status.conditions[Valid] = False` with `Reason: EnterpriseRequired` and a fixed, human-readable message describing the registration requirement.
- [ ] Applying an `obo`-typed `MCPExternalAuthConfig` referenced from an `MCPRemoteProxy` produces the same transition on the `MCPExternalAuthConfig` resource.
- [ ] Applying an `obo`-typed `MCPExternalAuthConfig` referenced from a `VirtualMCPServer` produces the same transition on the `MCPExternalAuthConfig` resource. The vMCP converter path wraps the sentinel through `pkg/vmcp/auth/converters/interface.go:191` (`"failed to convert to strategy: %w"`); the reconciler recognizes the wrapped sentinel via `errors.Is(err, externalauth.ErrEnterpriseRequired)`, not by matching the wrap prefix.
- [ ] In all three consumer CR cases, the consumer (`MCPServer` / `MCPRemoteProxy` / `VirtualMCPServer`) surfaces a parallel `Valid=False` / `EnterpriseRequired` condition. It does NOT surface a generic "unsupported external auth type" error.
- [ ] Generic-error guard: applying any `obo`-typed CR — across all three consumer paths — does not cause any reconciler log line or any status message to contain the substrings "unsupported external auth type" or "unknown middleware type".

Regression:

- [ ] Every existing integration test under `cmd/thv-operator/test-integration/mcp-external-auth/` passes unchanged.

Hygiene:

- [ ] No new lint findings; `task lint-fix` succeeds; `gofmt` and `goimports` are clean.
- [ ] All tests pass via `task test`.

#### (a) Switch arm in `cmd/thv-operator/pkg/controllerutil/tokenexchange.go`

Add one new arm to the existing 7-arm switch at line 101. Place it immediately before the `default:` arm. The dependency task introduced `oboHandler` as a package-level `OBOHandler` value; this arm calls through its `ApplyRunConfig` method so an out-of-tree-registered handler dispatches first and the default surfaces `ErrEnterpriseRequired`:

```go
case mcpv1beta1.ExternalAuthTypeOBO:
    return oboHandler.ApplyRunConfig(ctx, c, namespace, externalAuthConfig, options)
```

The seven existing arms (`tokenExchange`, `headerInjection`, `bearerToken`, `unauthenticated`, `embeddedAuthServer`, `awsSts`, `upstreamInject`) and the `default:` are unchanged. The new arm short-circuits before the `default:`'s `fmt.Errorf("unsupported external auth type: %s", externalAuthConfig.Spec.Type)` so the user-visible error path stays inside the `ErrEnterpriseRequired` chain.

#### (b) Switch arm in `cmd/thv-operator/controllers/virtualmcpserver_deployment.go`

Add one new arm to the existing 7-arm switch at line 789, placed before the `default:`. This switch returns `*corev1.EnvVar` (or `nil, nil` if the type doesn't need a secret env var). The dependency task introduced the `controllerutil.OBOSecretEnvVars` wrapper that calls through `oboHandler.SecretEnvVars`:

```go
case mcpv1beta1.ExternalAuthTypeOBO:
    envVars, err := ctrlutil.OBOSecretEnvVars(externalAuthConfig)
    if err != nil {
        return nil, err
    }
    if len(envVars) == 0 {
        return nil, nil
    }
    return &envVars[0], nil
```

The shape matches the other arms in the function: each existing arm either returns `(nil, nil)` (no secret to mount), or builds an `EnvVar` from the type-specific secret ref and returns it. Routing OBO through the function-pointer hook preserves the same shape while letting the default stub surface `ErrEnterpriseRequired`.

The seven existing arms and the `default:` are unchanged.

#### (c) Reconcile-loop branch in `cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go`

Add a new branch that runs the OBO function-pointer validation after the existing `Validate()` call succeeds, then maps the result to a `Valid=False` condition with the right reason string:

```go
if cfg.Spec.Type == mcpv1beta1.ExternalAuthTypeOBO {
    if err := controllerutil.OBOValidate(cfg); err != nil {
        reason := "InvalidConfig"
        if errors.Is(err, controllerutil.ErrEnterpriseRequired) {
            reason = "EnterpriseRequired"
        }
        return setInvalid(cfg, err, reason)
    }
}
```

Where `setInvalid` is a small helper that writes the condition through `controllerutil.MutateAndPatchStatus` per `.claude/rules/k8s-controller.md`. If a helper of this shape is not already in the file, introduce it in this task. Sketch:

```go
func (r *MCPExternalAuthConfigReconciler) setInvalid(
    ctx context.Context,
    cfg *mcpv1beta1.MCPExternalAuthConfig,
    err error,
    reason string,
) (ctrl.Result, error) {
    if patchErr := controllerutil.MutateAndPatchStatus(ctx, r.Client, cfg, func(c *mcpv1beta1.MCPExternalAuthConfig) {
        meta.SetStatusCondition(&c.Status.Conditions, metav1.Condition{
            Type:               mcpv1beta1.ConditionTypeValid,
            Status:             metav1.ConditionFalse,
            Reason:             reason,
            Message:            err.Error(),
            ObservedGeneration: c.Generation,
        })
    }); patchErr != nil {
        return ctrl.Result{}, patchErr
    }
    return ctrl.Result{}, nil // Don't requeue — the user must fix the spec or register a handler.
}
```

The `Reason` string MUST be the exact literal `EnterpriseRequired` (capital E, capital R) when `errors.Is(err, controllerutil.ErrEnterpriseRequired)` is true. External consumers and downstream tooling pattern-match on this exact string. For any other error from `OBOValidate`, use the literal `InvalidConfig`.

### Drift caveat — `r.Status().Update` vs `MutateAndPatchStatus` in this reconciler

The existing reconciler at `cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go` currently writes status via `r.Status().Update(ctx, externalAuthConfig)` at lines 108 and 137. This pre-existing pattern is inconsistent with the operator-rules convention that all new status writes go through `controllerutil.MutateAndPatchStatus`.

**The new OBO branch added in this task MUST use `controllerutil.MutateAndPatchStatus`.**

**Refactoring the existing `r.Status().Update` call sites at lines 108 and 137 is OUT OF SCOPE.** That cleanup is a separate concern, tracked outside this work. Reviewers should not request it as part of this task — coupling the cleanup to this task would inflate the diff and introduce unrelated regression risk into a task whose central change is the new OBO dispatch.

### CRD-enum sequencing caveat for integration tests

This task does NOT depend on the CRD admitting `obo` typed configs at the apiserver level — that change is the next task in the OSS sequence. Until that task lands, the dispatch arms added here are reachable only via paths that bypass the apiserver enum validation.

For the integration tests in this task, the practical options are:

- Use envtest's looser CRD enum validation (envtest can accept CRs the production apiserver would reject because envtest does not run all the same admission webhooks by default), OR
- Temporarily mutate the CRD enum schema in test setup to admit `obo` for the duration of the test suite, OR
- Drive the reconciler directly at the Go level (instantiate the reconciler and call `Reconcile()` with a hand-built `*MCPExternalAuthConfig`), bypassing the apiserver entirely.

Pick whichever option fits the existing `cmd/thv-operator/test-integration/mcp-external-auth/suite_test.go` shape best. The integration-test goal is to prove the dispatch wiring routes `obo` correctly across all three consumer paths; the apiserver-level admission test belongs to the CRD enum admission task.

### Patterns & Frameworks

- **Switch arm placement** — every new arm goes immediately before the `default:` so the generic "unsupported" error never fires for `obo`. The existing seven arms are not reordered.
- **Status write — `MutateAndPatchStatus`** — required by `.claude/rules/k8s-controller.md`. Pass a mutate closure that uses `meta.SetStatusCondition` so condition ordering and `ObservedGeneration` stay consistent with the existing call sites in this reconciler.
- **Sentinel error matching — `errors.Is`** — required by the project's `go-style.md`. The sentinel passes through `fmt.Errorf("...: %w", ...)` wraps unchanged; `errors.Is` walks the wrap chain.
- **Integration-test pattern** — match the existing `suite_test.go` envtest setup. Apply a CR, wait for the controller to reconcile, fetch the updated resource, assert on its `Status.Conditions`. Use `assert.Eventually(...)` (or `gomega.Eventually(...)`, whichever the existing tests use) to ride out the reconcile latency.

### Code Pointers

(All verified against upstream toolhive `v0.27.2` / `97b0cc3f`.)

- `cmd/thv-operator/pkg/controllerutil/tokenexchange.go:101` — `AddExternalAuthConfigOptions` switch site. Existing seven cases + `default:`. New `case ExternalAuthTypeOBO` arm placed before `default:`.
- `cmd/thv-operator/controllers/virtualmcpserver_deployment.go:789` — `(*VirtualMCPServerReconciler).getExternalAuthConfigSecretEnvVar` switch site. Existing seven cases + `default:`. New `case ExternalAuthTypeOBO` arm placed before `default:`.
- `cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go` — the reconciler whose `Reconcile()` method gains the new OBO branch. Existing `r.Status().Update(ctx, externalAuthConfig)` call sites are at lines 108 and 137; both stay unchanged.
- `cmd/thv-operator/pkg/controllerutil/status.go:77` — `MutateAndPatchStatus` signature. The helper is generic over `T client.Object` and takes a `mutate func(T)` closure.
- `pkg/vmcp/auth/converters/interface.go:191` — `"failed to convert to strategy: %w"` wrap site inside `DiscoverAndResolveAuth`. The vMCP integration test exercises this wrap; the reconciler must use `errors.Is(err, externalauth.ErrEnterpriseRequired)` to recognize the sentinel through it (matching by string would be brittle and is forbidden).
- `cmd/thv-operator/test-integration/mcp-external-auth/suite_test.go` — existing envtest harness for this directory; new tests slot in alongside the existing `mcpexternalauthconfig_controller_integration_test.go`.

### Component Interfaces

This task does not add any new exported types. It consumes the symbols defined in the dependency task:

```go
// In cmd/thv-operator/pkg/controllerutil (from the dependency task):

var ErrEnterpriseRequired error
func OBOValidate(cfg *mcpv1beta1.MCPExternalAuthConfig) error
func OBOSecretEnvVars(cfg *mcpv1beta1.MCPExternalAuthConfig) ([]corev1.EnvVar, error)
// oboHandler.ApplyRunConfig is called directly from the switch arm.

// In cmd/thv-operator/api/v1beta1 (from the dependency task):

const ExternalAuthTypeOBO ExternalAuthType = "obo"
```

The only new private helper introduced here is `setInvalid` (or equivalent) on the reconciler if one of that shape is not already present.

### Testing Strategy

**Unit tests** — none net new for the switch arms themselves; their behavior is covered by the integration tests below. If the new `setInvalid` helper has non-trivial logic (e.g., picks reason string from error wrap), add a small table-driven test next to the reconciler asserting:

- [ ] Given `controllerutil.ErrEnterpriseRequired`, the helper writes `Reason: EnterpriseRequired`.
- [ ] Given a wrapped `controllerutil.ErrEnterpriseRequired` (`fmt.Errorf("...: %w", ErrEnterpriseRequired)`), the helper still writes `Reason: EnterpriseRequired` (i.e., it uses `errors.Is`, not equality).
- [ ] Given any other error, the helper writes `Reason: InvalidConfig`.

**Integration tests** — in `cmd/thv-operator/test-integration/mcp-external-auth/` (build tag `//go:build integration`). Add three test cases (one per consumer CR type) plus a guard. Each follows the same shape: apply an `obo`-typed `MCPExternalAuthConfig`, apply a consumer CR referencing it, wait for the reconciler to settle, assert on the `MCPExternalAuthConfig`'s `Status.Conditions` and on the consumer CR's surfaced condition.

- [ ] **`MCPServer` consumer path**: Apply an `obo` `MCPExternalAuthConfig`, then an `MCPServer` with `spec.externalAuth` referencing it. Assert the `MCPExternalAuthConfig` has `status.conditions[Valid] = False`, `Reason: EnterpriseRequired`, with a fixed message describing the registration requirement. Assert the `MCPServer` surfaces a parallel `Valid=False` / `EnterpriseRequired` condition.
- [ ] **`MCPRemoteProxy` consumer path**: Same as above but with the consumer being an `MCPRemoteProxy` referencing the `obo` config.
- [ ] **`VirtualMCPServer` consumer path**: Same as above but with the consumer being a `VirtualMCPServer`. The vMCP runtime's `converters.DiscoverAndResolveAuth` at `pkg/vmcp/auth/converters/interface.go:191` wraps the converter result with `"failed to convert to strategy: %w"`; the test confirms the reconciler recognizes the sentinel through the wrap via `errors.Is(err, externalauth.ErrEnterpriseRequired)` (not by string match) and still surfaces `Reason: EnterpriseRequired`.
- [ ] **Generic-error guard** (a single test, or assertions woven into the three above): collect the reconciler logs and the consumer-CR / `MCPExternalAuthConfig` status messages across the three test cases above. Assert NONE of these substrings appear in any of those strings: `"unsupported external auth type"`, `"unknown middleware type"`. (This guards against the `default:` arm or the runner's middleware-map lookup leaking through for `obo`.)

**Edge cases**

- [ ] CRD-enum bypass for tests: ensure the envtest setup admits `obo` typed configs. Document the chosen bypass strategy (envtest's looser validation, in-test CRD mutation, or direct-reconciler invocation) in the test file's package comment so the next maintainer doesn't get blindsided.

**Regression**

- [ ] All existing tests under `cmd/thv-operator/test-integration/mcp-external-auth/` pass unchanged.

## Out of Scope

- Kubebuilder `+kubebuilder:validation:Enum=...;obo` marker update on `MCPExternalAuthConfigSpec.Type`. (Next OSS task — CRD enum admission.)
- `OBOConfig` placeholder struct definition + `OBO *OBOConfig` field on `MCPExternalAuthConfigSpec`. (Next OSS task.)
- CEL rule `self.type == 'obo' ? has(self.obo) : !has(self.obo)`. (Next OSS task.)
- New row in `validateTypeConfigConsistency` for the `OBOConfig` field. (Next OSS task.)
- No-op `case mcpv1beta1.ExternalAuthTypeOBO` arm in the `Validate()` switch at `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go:1019`. (Next OSS task.)
- Regenerated `deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpexternalauthconfigs.yaml` and `task crdref-gen` output. (Next OSS task.)
- **Refactoring the existing `r.Status().Update(ctx, externalAuthConfig)` call sites at lines 108 and 137 of `cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go` to use `controllerutil.MutateAndPatchStatus`.** Only the new OBO branch must use `MutateAndPatchStatus`. The cleanup of the pre-existing call sites is a separate concern, tracked outside this work.
- The stub middleware factory itself, the default `OBOHandler` global, the `ErrEnterpriseRequired` sentinel, and the `OBOConverterStub` — those are the dependency task's scope.
- Refactoring either operator-side switch into a registry. The two-switch shape (with one new arm per switch) is intentional for the minimal-variant rollout.
- Refactoring the literal middleware-factory map at `pkg/runner/middleware.go:30-49` into a registry. Same rationale.

## References

- `cmd/thv-operator/pkg/controllerutil/tokenexchange.go:101` — first dispatch site.
- `cmd/thv-operator/controllers/virtualmcpserver_deployment.go:789` — second dispatch site.
- `cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go` — reconciler whose new OBO branch is added by this task. Pre-existing `r.Status().Update` call sites at lines 108 and 137 are explicitly out of scope.
- `cmd/thv-operator/pkg/controllerutil/status.go:77` — `MutateAndPatchStatus` signature.
- `pkg/vmcp/auth/converters/interface.go:191` — error-wrap site for the vMCP integration test path.
- [RFC 8693 — OAuth 2.0 Token Exchange](https://datatracker.ietf.org/doc/html/rfc8693) — standards basis for the broader on-behalf-of pattern.
- Upstream code pointers cited above, all anchored to `stacklok/toolhive` `v0.27.2` (`97b0cc3f`).

tgrunnagle:
## Summary

Wires the new `obo` external auth type into the three operator dispatch sites and the `MCPExternalAuthConfig` reconciler, then proves the wiring with unit-level tests against the registered default handler.

- **Why**: Before this change, the `obo` arms added in #5327 existed only to satisfy the `exhaustive` linter — they returned `nil` (silent no-op) or a stub error that did not flow through the reconciler's status path. As a result, applying an `obo`-typed `MCPExternalAuthConfig` either disappeared silently or surfaced a generic `unsupported external auth type` / `unknown middleware type` error in a consumer reconciler downstream. This PR routes `obo` through the registered handler at every dispatch site so the upstream-only default handler can return `obo.ErrEnterpriseRequired` and the reconciler can translate that sentinel into a uniform `status.conditions[Valid] = False` / `Reason: EnterpriseRequired` outcome — the contract external consumers and downstream tooling pattern-match on.
- **What**:
  - `AddExternalAuthConfigOptions` (`cmd/thv-operator/pkg/controllerutil/tokenexchange.go`): replaced the placeholder `obo` arm with a call into the existing `OBOApplyRunConfig` wrapper.
  - `(*VirtualMCPServerReconciler).getExternalAuthConfigSecretEnvVar` (`cmd/thv-operator/controllers/virtualmcpserver_deployment.go`): the `obo` arm now dispatches through `controllerutil.OBOSecretEnvVars` via a small extracted helper (`firstOBOSecretEnvVar`) so the function stays under the cyclomatic-complexity lint threshold.
  - `MCPExternalAuthConfigReconciler.Reconcile` (`cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go`): added an `ExternalAuthTypeOBO` branch that calls `controllerutil.OBOValidate`, maps `obo.ErrEnterpriseRequired` to `Reason: EnterpriseRequired` (and any other validation error to `Reason: InvalidConfig`), and writes the condition through a new `setInvalid` helper that uses `MutateAndPatchStatus` per `.claude/rules/k8s-controller.md`.
  - Added exported reason constants `ConditionReasonEnterpriseRequired` and `ConditionReasonInvalidConfig` in `mcpexternalauthconfig_types.go` so the user-facing literals live in one place.
  - Added unit-level test coverage for each of the three dispatch sites against the registered default handler, plus reconciler-level coverage for the new `setInvalid` path (including the wrapped-sentinel case to confirm `errors.Is` matching).

Closes #5328

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

New tests cover each new code path:

- `TestAddExternalAuthConfigOptions_OBO` — proves the `obo` arm of `AddExternalAuthConfigOptions` returns an error wrapping `obo.ErrEnterpriseRequired` and never reaches the `default:` arm's `unsupported external auth type` formatter. Also asserts neither `"unsupported external auth type"` nor `"unknown middleware type"` appears in the error message.
- `TestGetExternalAuthConfigSecretEnvVar_OBO` — proves the `obo` arm of the vMCP deployment's secret-env-var switch dispatches through the registered handler instead of silently returning `nil, nil`. Asserts both generic-error guard substrings as well.
- `TestMCPExternalAuthConfigReconciler_OBO_DefaultHandler_SetsEnterpriseRequired` — drives the reconciler with a fake client against an `obo`-typed `MCPExternalAuthConfig`, asserts `Valid=False` / `Reason=EnterpriseRequired` lands on the resource, asserts the message equals `obo.ErrEnterpriseRequired.Error()`, and asserts neither generic substring leaks into the message.
- `TestMCPExternalAuthConfigReconciler_OBO_ClearsStaleIdentitySynthesized` — **regression guard**: when a user switches an existing `MCPExternalAuthConfig` from `embeddedAuthServer` (which set `IdentitySynthesized=True`) to `obo`, the stale `IdentitySynthesized` condition must be removed alongside the new `Valid=False`/`EnterpriseRequired` write. Defends against the `MutateAndPatchStatus` snapshot/diff trap where the pre-mutate snapshot already contained the in-memory `applyIdentitySynthesizedCondition` write — which would silently disappear from the merge patch.
- `TestMCPExternalAuthConfigReconciler_OBO_ReasonSelectionInReconcile` — table-driven test that drives the production reason-selection branch through `Reconcile` by registering a stub OBO handler whose `Validate` returns each test's error. Covers (a) bare sentinel → `EnterpriseRequired`, (b) wrapped sentinel through `fmt.Errorf("...: %w", ...)` → `EnterpriseRequired` (proves the `errors.Is` matching that lets the vMCP `pkg/vmcp/auth/converters/interface.go` wrap survive), (c) any other error → `InvalidConfig`.
- `TestDiscoverAndResolveAuth_OBO_SentinelSurvivesWrap` (`pkg/vmcp/auth/converters/`) — proves `errors.Is` matches `obo.ErrEnterpriseRequired` through the vMCP `"failed to convert to strategy: %w"` wrap site at `interface.go:191`.

## API Compatibility

Only additive changes to the `v1beta1` API: two new exported `ConditionReasonXxx` string constants in `mcpexternalauthconfig_types.go`. No CRD schema change in this PR (the CRD enum still rejects `"obo"` at the apiserver level until #5329 lands).

Yes. Applying an `obo`-typed `MCPExternalAuthConfig` to a cluster running an upstream-only build now produces a deterministic, machine-readable status:

```
status.conditions[Valid].status  = False
status.conditions[Valid].reason  = "EnterpriseRequired"
status.conditions[Valid].message = "on-behalf-of (OBO) external auth type requires an enterprise build"
```

External consumers and downstream tooling can pattern-match on the literal `EnterpriseRequired` reason string — it is part of the user-facing contract (codified by the new `ConditionReasonEnterpriseRequired` constant). The message text comes from `obo.ErrEnterpriseRequired.Error()` at `pkg/auth/obo/errors.go`. Out-of-tree builds that register a real handler via `controllerutil.RegisterOBOHandler` bypass this path entirely.

(Note: the CRD enum currently rejects `"obo"` at the apiserver layer, so this status surface is only reachable today through paths that bypass apiserver enum validation — see "Special notes for reviewers" below. The CRD admission change is the next OSS task, #5329.)

## Implementation plan

<details>

The implementation followed the recommended approach in issue #5328 verbatim — the issue body already contains the full implementation plan (code pointers, exact switch-arm placements, the `setInvalid` helper sketch, the reason-string contract, and the integration-test strategy). No separate plan document was generated for this task; the issue is the design doc.

One bounded deviation from the issue's integration-test plan: the issue offered three options for the CRD-enum bypass and recommended picking whichever fit the existing envtest harness best. We chose **option 3 — drive the reconciler directly at the Go level with a fake client**, because the CRD enum does not admit `"obo"` until #5329 lands and direct-reconciler invocation kept the test surface contained to this PR. The integration tests under `cmd/thv-operator/test-integration/mcp-external-auth/` are unchanged; the per-site coverage lives in standard `*_test.go` files instead.

</details>

- **Stacked PR — target main, but review against the post-#5325/#5327 tree.** This is task 3 of 4 in the OSS-OBO stack. The branch contains commits from #5325 (operator/proxyrunner `Run` factor) and #5327 (default OBO handler hooks + vMCP/proxy converter stubs) because of the local stack workflow. Per `.claude/rules/pr-creation.md`, this PR should be reviewed and merged **after** the two upstream PRs land on `main`; once they do, the diff against `main` will reduce to just the #5328-scoped changes (the three commits prefixed `Wire OBO dispatch arms`, `Address code review feedback on OBO dispatch wiring`, `Consolidate OBO test coverage`). A scoped diff at `git diff 30953187..HEAD` previews exactly that surface (~467 lines, 8 files).
- **CRD-enum sequencing — chose option 3 from the issue.** The CRD enum does not admit `"obo"` until #5329 lands, so apiserver-routed integration tests would require either an in-test CRD mutation or relying on envtest's looser admission. We picked the issue's option 3 — **drive the reconciler directly at the Go level with a fake client** — because it kept the test surface contained to this PR and avoided coupling test infrastructure to the next task. All three dispatch sites have direct unit-level coverage as a result; the existing integration suite under `cmd/thv-operator/test-integration/mcp-external-auth/` is unchanged.
- **Consumer-CR Status Condition Parity gap — tracked in #5347.** Issue #5328 AC4 calls for the consumer CRDs (`MCPServer` / `MCPRemoteProxy` / `VirtualMCPServer`) to surface a parallel `Valid=False` / `EnterpriseRequired` condition. The codebase does **not** mirror `MCPExternalAuthConfig` conditions onto consumer CR statuses today — there is no existing condition-mirror machinery to extend. This PR satisfies the AC via the dispatch error path (each consumer reconciler now receives `obo.ErrEnterpriseRequired` rather than the generic "unsupported" error and surfaces it through its existing failure path), but the literal condition-mirror is deferred to **#5347** so it can land before #5329 makes `obo` reachable in production. Per `.claude/rules/operator.md`'s Status Condition Parity rule, the mirror is genuine parity work and must merge before #5329.
- **Pre-existing `r.Status().Update` call sites are intentionally untouched.** The reconciler still calls `r.Status().Update(ctx, externalAuthConfig)` at the ValidationFailed return and the conditionChanged write. The issue explicitly lists refactoring those to `MutateAndPatchStatus` as out of scope; the new `setInvalid` helper uses `MutateAndPatchStatus` as required for new status writes.
- **`setInvalid` re-fetches before patching.** This is deliberate. `MutateAndPatchStatus` snapshots the object at the start of the call and diffs the post-mutate state against that snapshot; any mutation already present on `cfg.Status.Conditions` (notably the in-memory `applyIdentitySynthesizedCondition` write earlier in `Reconcile`) would land in both halves of the diff and silently disappear from the merge patch. Re-fetching and re-applying the advisory inside the patch closure folds both the advisory transition and the `Valid=False` write into the same patch. `applyIdentitySynthesizedCondition` is idempotent on the same spec, so the double computation is benign. The `TestMCPExternalAuthConfigReconciler_OBO_ClearsStaleIdentitySynthesized` regression test guards this invariant — temporarily remove the `r.applyIdentitySynthesizedCondition(c)` call inside `setInvalid`'s closure and the test should fail.


cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -1026,6 +1026,23 @@ const (
...
 
+// Condition reasons for ConditionTypeValid on OBO-typed configs. These
+// literals are part of the user-facing contract — external consumers and
+// downstream tooling pattern-match on them.
+const (
+	// ConditionReasonEnterpriseRequired: an obo-typed MCPExternalAuthConfig
+	// requires an enterprise build that has registered an OBO handler via
+	// controllerutil.RegisterOBOHandler. Upstream-only builds surface this
+	// reason for every obo-typed config.
+	ConditionReasonEnterpriseRequired = "EnterpriseRequired"
```
tgrunnagle, line 1037: **F1 — MEDIUM (consensus 8/10): Consumer-CR Status Condition Parity gap.**

This new constant `ConditionReasonEnterpriseRequired` is the user-facing contract for the `MCPExternalAuthConfig` Valid condition, but issue #5328 AC4 says the *consumer* CRDs (`MCPServer` / `MCPRemoteProxy` / `VirtualMCPServer`) must each surface a parallel `Valid=False` / `EnterpriseRequired` condition. A grep of the consumer reconcilers confirms none of them references `ConditionReasonEnterpriseRequired`; each consumer has its own reason taxonomy (e.g., `ConditionReasonMCPRemoteProxyExternalAuthConfigFetchError`).

The PR body acknowledges this as "interpretive" because no condition-mirror machinery exists today, but per `.claude/rules/operator.md` Status Condition Parity ("a gap means one type silently accepts invalid config that the other rejects") it remains a genuine parity gap. Since #5329 (CRD enum admission) is what makes obo reachable in production, the gap MUST be closed before #5329 lands.

**Action**: open a follow-up task explicitly tracking the consumer-CR condition mirroring before #5329, and call out the deferral under the PR body's "Special notes" so cold reviewers don't have to walk the AC list to find the gap.

cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
```diff
@@ -90,6 +92,14 @@ func (r *MCPExternalAuthConfigReconciler) Reconcile(ctx context.Context, req ctr
 	// it before validation ensures the advisory tracks the current spec on
 	// every reconcile — including the validation-failure path — so a broken
 	// edit cannot leave a stale True/upstream-name dangling.
+	//
+	// Note: the OBO failure path routes through setInvalid, which discards
+	// this in-memory mutation (its MutateAndPatchStatus call re-fetches and
+	// re-applies the advisory inside its patch closure). The in-memory
+	// mutation here is therefore load-bearing only on the validation-failure
+	// path (line 110) and the Valid=True path (line 139). The function is
```
tgrunnagle: **F3 — LOW (consensus 8/10): Stale line-number references.**

This comment cites "the validation-failure path (line 110)" and "the Valid=True path (line 139)", but in the post-PR file the actual `r.Status().Update` for the validation-failure path is at **line 118** and for the Valid=True path at **line 161**. Lines 110 and 139 are just neighboring comment / `meta.SetStatusCondition` lines, not the call sites the prose describes.

Per `.claude/rules/go-style.md` (Keep Comments Synchronized With Code), line-number references in comments are brittle — they drift on every re-format or insertion.

```suggestion
	// Note: the OBO failure path routes through setInvalid, which discards
	// this in-memory mutation (its MutateAndPatchStatus call re-fetches and
	// re-applies the advisory inside its patch closure). The in-memory
	// mutation here is therefore load-bearing only on the validation-failure
	// path (the r.Status().Update on the ValidationFailed return) and the
	// Valid=True path (the r.Status().Update on the conditionChanged write).
	// The function is idempotent on the same spec, so the double computation
	// on the OBO path is benign.
```

tgrunnagle (commented):
## Multi-agent review summary

No HIGH-severity findings; 2 MEDIUM (F1, F2) and 8 LOW. Recommended action: **COMMENT** — the MEDIUM findings are worth addressing before merge but neither is a hard blocker. (`event=COMMENT` because the reviewer is also the PR author; nothing else to read into.)

### Consensus-scored findings

| # | Title | Severity | Score | Raised by |
|---|-------|----------|-------|-----------|
| F1 | Consumer-CR Status Condition Parity gap (issue #5328 AC4 unimplemented) | MEDIUM | 8 | operator, intent |
| F2 | `setInvalid_ReasonSelection` test duplicates production logic — tautological | MEDIUM | 9 | error-handling, test-coverage, general |
| F3 | Stale line-number references in the new `Reconcile` comment block | LOW | 8 | operator, style |
| F4 | PR body's enumerated test list is stale vs the diff (5 vs 6) | LOW | 7 | test-coverage, intent |
| F5 | `TODO(#5328)` marker left behind at `mcpexternalauthconfig_types.go:1118` | LOW | 7 | style (verified) |
| F6 | `OBO_DefaultHandler...EnterpriseRequired` lacks positive message assertion | LOW | 7 | test-coverage, operator |
| F7 | Generic-error guard missing "unknown middleware type" in 2 of 4 tests | LOW | 7 | error-handling |
| F8 | PR body's promised user-facing message ≠ actual `obo.ErrEnterpriseRequired` text | LOW | 7 | general (verified) |
| F9 | `OBOApplyRunConfig` doc comment references caller (style rule violation) | LOW | 6 | style |
| F10 | `ClearsStaleIdentitySynthesized` guards an implicit invariant without naming it | LOW | 7 | error-handling, test-coverage |

Findings below threshold (score < 7) are listed under §11 of the local review at `.pr-review/pr-5345/findings.md`. They're not included here.

### Executive summary

This PR wires the OBO external-auth type through the three operator dispatch sites (`AddExternalAuthConfigOptions`, `getExternalAuthConfigSecretEnvVar`, the `MCPExternalAuthConfig` reconcile loop) and adds a `setInvalid` helper that uses `MutateAndPatchStatus` to write `Valid=False` / `Reason=EnterpriseRequired` when the default OBO handler returns `obo.ErrEnterpriseRequired`. The approach is structurally sound: it follows #5328's recommended implementation almost verbatim, removes the two pre-existing `TODO(#5328)` no-op stubs at the dispatch sites, and uses `errors.Is` so the sentinel survives the vMCP `"failed to convert to strategy: %w"` wrap. Roughly half the diff is targeted unit tests, including a load-bearing regression guard (`OBO_ClearsStaleIdentitySynthesized`) for the `setInvalid` re-`Get` + re-apply pattern that defends against `MutateAndPatchStatus`'s snapshot/diff trap.

### Pros / Cons

**Pros**
- Tight implementation that mirrors #5328's recommended approach; dispatch arms placed correctly before `default:` so the "unsupported external auth type" path can never fire for `obo`.
- `setInvalid`'s re-`Get` + re-apply pattern is genuinely subtle and correct — the regression test makes the invariant testable.
- Reason strings (`EnterpriseRequired` / `InvalidConfig`) extracted to typed constants with doc comments declaring them part of the user-facing contract.
- Generic-error guard assertions in three of the four new tests directly enforce the no-leak-of-"unsupported"/"unknown-middleware" criterion.
- `stderrors "errors"` alias is consistent with the existing convention in sibling controller files.

**Cons**
- AC4 not implemented: consumer CRDs don't surface a parallel `EnterpriseRequired` condition (see F1).
- The `setInvalid_ReasonSelection` table-test asserts against its own re-implementation of the production reason-selection logic, not the production code (see F2).
- Two comments cite absolute line numbers (110, 139) that don't quite match the post-PR file (see F3).
- A third `TODO(#5328)` marker is left behind at `mcpexternalauthconfig_types.go:1118` after the issue this PR closes (see F5).
- The PR body promises a user-facing message string that doesn't match the actual `obo.ErrEnterpriseRequired` text (see F8).

### How to test

```
git fetch origin pull/5345/head:pr-5345 && git checkout pr-5345
task lint-fix
task test
```

Load-bearing-invariant check: temporarily remove the `r.applyIdentitySynthesizedCondition(c)` call inside `setInvalid`'s `MutateAndPatchStatus` closure (controller.go:~225). `TestMCPExternalAuthConfigReconciler_OBO_ClearsStaleIdentitySynthesized` should fail; if it doesn't, the regression guard is not load-bearing.

### Out-of-diff findings (not posted inline)

**F4 — PR body test enumeration is stale.** The PR description lists five tests, but the diff actually adds six. The two `setInvalid_WrappedSentinel`/`NonSentinelError` entries are consolidated into one table-driven `setInvalid_ReasonSelection`, and the regression-guard `OBO_ClearsStaleIdentitySynthesized` is not mentioned at all. Update the body's enumeration to match the diff and add a one-line note for `ClearsStaleIdentitySynthesized` explaining what bug it guards.

**F5 — Leftover `TODO(#5328)` at `mcpexternalauthconfig_types.go:1118`.** The PR removes the `TODO(#5328)` markers from `tokenexchange.go:228` and `virtualmcpserver_deployment.go:835`, but a third marker at `mcpexternalauthconfig_types.go:1118` (in the `Validate` switch's OBO arm — not touched by this diff) is left behind. The comment body is load-bearing WHY content and should stay; just strip the `TODO(#5328):` prefix so the marker doesn't outlive the issue it references.

**F8 — User-facing message in PR body doesn't match the actual sentinel.** The PR body's "Does this introduce a user-facing change?" section quotes `condition.Message` as `"OBO external auth requires an enterprise build with a registered OBO handler"`. The actual `obo.ErrEnterpriseRequired.Error()` is `"on-behalf-of (OBO) external auth type requires an enterprise build"` (`pkg/auth/obo/errors.go:19-20`) — the phrase *"with a registered OBO handler"* does not appear. External consumers should pattern-match the reason string (not the message), but downstream tooling docs built against the PR body will be wrong. Update either the PR body or the sentinel text.

### Agents consulted

6 specialist pr-reviewers: operator/k8s, error-handling, Go-style, test-coverage, intent-alignment, general. Codex cross-review skipped (CLI not installed locally).

---

cmd/thv-operator/controllers/mcpexternalauthconfig_controller_test.go
```diff
@@ -1476,3 +1479,206 @@ func findCondition(conditions []metav1.Condition, t string) *metav1.Condition {
...
+	validCond := findCondition(updated.Status.Conditions, mcpv1beta1.ConditionTypeValid)
+	require.NotNil(t, validCond, "Valid condition must be set for OBO-typed config")
+	assert.Equal(t, metav1.ConditionFalse, validCond.Status)
+	assert.Equal(t, mcpv1beta1.ConditionReasonEnterpriseRequired, validCond.Reason,
+		"the exact reason string is part of the user-facing contract — external consumers pattern-match on it")
+
+	// Generic-error guard: the dispatch path must NOT leak a generic
+	// "unsupported external auth type" or "unknown middleware type" message.
+	assert.NotContains(t, validCond.Message, "unsupported external auth type")
+	assert.NotContains(t, validCond.Message, "unknown middleware type")
```
tgrunnagle, line 1541: **F6 — LOW (consensus 7/10): Missing positive assertion on condition message.**

The test asserts the message *doesn't* contain two generic strings, but never asserts what it *does* contain. Issue #5328 describes the message as "a fixed, human-readable message describing the registration requirement"; the actual sentinel text is `"on-behalf-of (OBO) external auth type requires an enterprise build"` (`pkg/auth/obo/errors.go:19-20`). If a future refactor swaps the sentinel message to empty or wrong text, only the reason-string check would catch it — and that check would still pass for an empty message.

```suggestion
	// Generic-error guard: the dispatch path must NOT leak a generic
	// "unsupported external auth type" or "unknown middleware type" message.
	assert.NotContains(t, validCond.Message, "unsupported external auth type")
	assert.NotContains(t, validCond.Message, "unknown middleware type")

	// Positive assertion: the message is the sentinel's user-facing text.
	assert.Equal(t, obo.ErrEnterpriseRequired.Error(), validCond.Message,
		"condition.Message must surface the sentinel's user-facing text")
```

cmd/thv-operator/controllers/mcpexternalauthconfig_controller_test.go
```diff
@@ -1476,3 +1479,206 @@ func findCondition(conditions []metav1.Condition, t string) *metav1.Condition {
...
+
+	// IdentitySynthesized must be removed (the spec is no longer embeddedAuthServer).
+	assert.Nil(t, findCondition(updated.Status.Conditions, mcpv1beta1.ConditionTypeIdentitySynthesized),
+		"stale IdentitySynthesized condition must be removed once the spec leaves embeddedAuthServer")
+
+	// Valid=False/EnterpriseRequired must be set in the same reconcile pass.
+	validCond := findCondition(updated.Status.Conditions, mcpv1beta1.ConditionTypeValid)
+	require.NotNil(t, validCond)
+	assert.Equal(t, metav1.ConditionFalse, validCond.Status)
+	assert.Equal(t, mcpv1beta1.ConditionReasonEnterpriseRequired, validCond.Reason)
```
tgrunnagle, line 1618: **F10 — LOW (consensus 7/10): Implicit reconciler-ordering invariant not documented.**

This test catches a regression where `setInvalid`'s `MutateAndPatchStatus` could silently drop the stale `IdentitySynthesized` removal. The defense is BOTH the re-`Get` of the fresh object AND the re-invocation of `applyIdentitySynthesizedCondition` inside the patch closure. A future maintainer "DRY-ing" the patch closure to remove the duplicate `applyIdentitySynthesizedCondition` call would break the regression guarantee silently.

The test asserts the final state but doesn't name the invariant it guards. Consider adding a one-line comment near `controller.go:225` (where `r.applyIdentitySynthesizedCondition(c)` is re-called inside the closure) that says: *"This re-invocation is load-bearing: it folds the advisory transition into the same `MutateAndPatchStatus` patch as the `Valid=False` write. See `TestMCPExternalAuthConfigReconciler_OBO_ClearsStaleIdentitySynthesized`."* That makes the load-bearing nature visible at the modification site.

cmd/thv-operator/controllers/mcpexternalauthconfig_controller_test.go
```diff
@@ -1476,3 +1479,206 @@ func findCondition(conditions []metav1.Condition, t string) *metav1.Condition {
...
+
+			fakeClient := fake.NewClientBuilder().
+				WithScheme(scheme).
+				WithObjects(cfg).
+				WithStatusSubresource(&mcpv1beta1.MCPExternalAuthConfig{}).
+				Build()
+
+			r := &MCPExternalAuthConfigReconciler{Client: fakeClient, Scheme: scheme}
+
+			reason := mcpv1beta1.ConditionReasonInvalidConfig
```
tgrunnagle: **F2 — MEDIUM (consensus 9/10): Test duplicates production reason-selection logic.**

This block reproduces the reason-selection from controller.go:130-133 (`reason := InvalidConfig; if errors.Is(...) { reason = EnterpriseRequired }`), then passes the computed `reason` into `setInvalid`. `setInvalid` itself never makes the `errors.Is` decision — it just writes whatever reason it's handed. So the test asserts "setInvalid writes the reason I gave it" — it does NOT exercise the production reason-selection at controller.go:130-133.

If a refactor regresses the production `errors.Is` check (e.g., swaps `stderrors.Is` for `==`), this test still passes because the assertion runs against the test's *own copy* of the logic. The wrapped-sentinel subcase — arguably the most important behavior to guard, since the issue's `pkg/vmcp/auth/converters/interface.go:191` wrap is the whole reason for using `errors.Is` — is not actually covered by any test that exercises the production decision branch.

Three options:
1. **Drive `Reconcile`**: register a stub OBO handler via `RegisterOBOHandler` that returns each `tt.inputErr` from `Validate`, then call `Reconcile` and assert the resulting condition. This exercises the production selection.
2. **Extract the helper**: pull the selection into a tiny pure helper (e.g., `func oboReasonFor(err error) string`) and table-test that helper directly.
3. **Rename**: at minimum rename to `setInvalid_WritesGivenReason` so the assertion matches what it actually verifies, and add a complementary test for the selection branch elsewhere.

Option 1 or 2 is preferred; option 3 keeps the regression-coverage gap but at least removes the misleading test name.

cmd/thv-operator/controllers/virtualmcpserver_externalauth_test.go
```diff
@@ -1649,3 +1650,43 @@ func TestBuildOutgoingAuthConfig_InlineBackendSubjectProviderInjection(t *testin
...
+		Client:           fakeClient,
+		Scheme:           scheme,
+		PlatformDetector: ctrlutil.NewSharedPlatformDetector(),
+	}
+
+	envVar, err := r.getExternalAuthConfigSecretEnvVar(t.Context(), "default", authCfg.Name)
+	require.Error(t, err)
+	assert.ErrorIs(t, err, obo.ErrEnterpriseRequired,
+		"the default OBO handler returns obo.ErrEnterpriseRequired; the dispatch arm must propagate it")
+	assert.Nil(t, envVar, "no env var should be returned on the error path")
```
tgrunnagle, line 1691: **F7b — LOW (consensus 7/10): Generic-error guard missing on this dispatch site.**

Issue #5328's Generic-error guard AC requires that NEITHER `"unsupported external auth type"` NOR `"unknown middleware type"` appears in any error surfaced for an obo-typed config — across all three consumer paths. This test asserts `ErrorIs(sentinel)` and `Nil(envVar)` but does not explicitly check either substring. If a future wrapper re-introduces one of the generic strings (e.g., a `fmt.Errorf("unknown middleware type for %s: %w", ...)` wrap), `ErrorIs` would still match and this test would silently pass.

The other tests (`mcpexternalauthconfig_controller_test.go::OBO_DefaultHandler...`, `obo_test.go::DiscoverAndResolveAuth_OBO_SentinelSurvivesWrap`) do assert both substrings. Bring this test into uniformity:

```suggestion
	envVar, err := r.getExternalAuthConfigSecretEnvVar(t.Context(), "default", authCfg.Name)
	require.Error(t, err)
	assert.ErrorIs(t, err, obo.ErrEnterpriseRequired,
		"the default OBO handler returns obo.ErrEnterpriseRequired; the dispatch arm must propagate it")
	assert.Nil(t, envVar, "no env var should be returned on the error path")

	// Generic-error guard: per issue #5328 AC, neither generic substring may leak.
	assert.NotContains(t, err.Error(), "unsupported external auth type")
	assert.NotContains(t, err.Error(), "unknown middleware type")
```

cmd/thv-operator/pkg/controllerutil/tokenexchange.go
```diff
@@ -115,9 +115,9 @@ func OBOSecretEnvVars(cfg *mcpv1beta1.MCPExternalAuthConfig) ([]corev1.EnvVar, e
 
 // OBOApplyRunConfig runs the registered OBO handler's ApplyRunConfig function.
 // With the default handler it returns obo.ErrEnterpriseRequired without
-// mutating the supplied options slice. The follow-up dispatch-wiring task
-// (#5328) is the first caller of this wrapper; exporting it now keeps the
-// three-method surface symmetric with OBOValidate / OBOSecretEnvVars.
+// mutating the supplied options slice. Called from the ExternalAuthTypeOBO
```
tgrunnagle: **F9 — LOW (consensus 6/10): Doc comment references its caller (style rule).**

The "Called from the `ExternalAuthTypeOBO` arm of `AddExternalAuthConfigOptions`" clause names a specific caller — exactly the pattern `.claude/rules/go-style.md` warns against ("Don't reference the current task/fix/callers ('used by X', 'added for the Y flow')"). The symmetry rationale is genuinely useful and worth keeping; the caller reference will rot the moment the caller moves.

```suggestion
// OBOApplyRunConfig runs the registered OBO handler's ApplyRunConfig function.
// With the default handler it returns obo.ErrEnterpriseRequired without
// mutating the supplied options slice. Exported so that the three-method
// surface stays symmetric with OBOValidate / OBOSecretEnvVars; routes through
// the package-level OBO handler registered via RegisterOBOHandler.
```

cmd/thv-operator/pkg/controllerutil/tokenexchange_test.go
```diff
@@ -234,3 +237,52 @@ func TestRegisterOBOHandler_PanicsOnNilField(t *testing.T) {
...
+		nil, // oidcConfig — not required for OBO
+		&options,
+	)
+	require.Error(t, err)
+	assert.ErrorIs(t, err, obo.ErrEnterpriseRequired,
+		"the default OBO handler returns obo.ErrEnterpriseRequired; the dispatch arm must propagate it")
+
+	// Generic-error guard: the dispatch arm must short-circuit the default
+	// arm's "unsupported external auth type: ..." path.
+	assert.NotContains(t, err.Error(), "unsupported external auth type")
```
tgrunnagle, line 287: **F7a — LOW (consensus 7/10): Generic-error guard missing "unknown middleware type".**

Issue #5328's Generic-error guard AC requires that NEITHER `"unsupported external auth type"` NOR `"unknown middleware type"` appears. This test checks only the first substring; the other tests in this PR (e.g., `obo_test.go`, `mcpexternalauthconfig_controller_test.go`) check both. Bring this test into uniformity with the AC and the sibling tests:

```suggestion
	// Generic-error guard: the dispatch arm must short-circuit the default
	// arm's "unsupported external auth type: ..." path and must not leak the
	// middleware-map lookup's "unknown middleware type" path either.
	assert.NotContains(t, err.Error(), "unsupported external auth type")
	assert.NotContains(t, err.Error(), "unknown middleware type")
	assert.Empty(t, options, "default OBO handler must not append to the options slice")
```

cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
tgrunnagle: Addressed in edea978b (F3): replaced the line-number prose with behavior-named references ("the r.Status().Update on the ValidationFailed return" / "the r.Status().Update on the conditionChanged write") per the suggestion.

cmd/thv-operator/controllers/mcpexternalauthconfig_controller_test.go
tgrunnagle: Addressed in edea978b (F6): added the positive `assert.Equal(t, obo.ErrEnterpriseRequired.Error(), validCond.Message, ...)` assertion.

cmd/thv-operator/controllers/mcpexternalauthconfig_controller_test.go
tgrunnagle: Addressed in edea978b (F10): added a comment at the re-invocation site naming `TestMCPExternalAuthConfigReconciler_OBO_ClearsStaleIdentitySynthesized` as the regression guard and explaining that removing the call would cause MutateAndPatchStatus to silently drop the IdentitySynthesized transition.

cmd/thv-operator/controllers/mcpexternalauthconfig_controller_test.go
tgrunnagle: Addressed in edea978b (F2): went with option 1 — replaced the tautological table-test with `TestMCPExternalAuthConfigReconciler_OBO_ReasonSelectionInReconcile`, which uses `RegisterOBOHandler` to install a stub whose `Validate` returns each `tt.validateErr`, then calls `Reconcile` and asserts on the resulting condition. The production `stderrors.Is` decision branch is now exercised, including the wrapped-sentinel case.

cmd/thv-operator/controllers/virtualmcpserver_externalauth_test.go
tgrunnagle: Addressed in edea978b (F7b): added both `assert.NotContains` guards (for "unsupported external auth type" and "unknown middleware type") matching the sibling tests' pattern.

cmd/thv-operator/pkg/controllerutil/tokenexchange.go
tgrunnagle: Addressed in edea978b (F9): applied the suggested rewrite — dropped the "Called from the `ExternalAuthTypeOBO` arm of `AddExternalAuthConfigOptions`" clause; kept the symmetry rationale and routed-through-RegisterOBOHandler note.

cmd/thv-operator/pkg/controllerutil/tokenexchange_test.go
tgrunnagle: Addressed in edea978b (F7a): added `assert.NotContains(t, err.Error(), "unknown middleware type")` so both generic substrings are guarded.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Tracked in #5347 — opened a follow-up issue specifically for mirroring the `MCPExternalAuthConfig` `Valid=False`/`EnterpriseRequired` condition onto the three consumer CRs (`MCPServer`/`MCPRemoteProxy`/`VirtualMCPServer`), with an explicit acceptance criterion that it must merge before #5329 lands. Also updated the PR body's "Special notes for reviewers" to call out the deferral so cold reviewers see it without walking the AC list.

tgrunnagle:
Addressed out-of-diff findings from the review summary:

- **F4** (PR body test enumeration stale): updated the body to list all 6 new tests, with the consolidated table-driven `OBO_ReasonSelectionInReconcile` and the previously-unmentioned `OBO_ClearsStaleIdentitySynthesized` regression guard each described.
- **F8** (PR body user-facing message string mismatched the sentinel): the body now quotes the actual `obo.ErrEnterpriseRequired.Error()` text (`"on-behalf-of (OBO) external auth type requires an enterprise build"`) and points at `pkg/auth/obo/errors.go` as the source.
- **F1** (consumer-CR condition parity gap): tracked in #5347, called out under "Special notes for reviewers" with the must-merge-before-#5329 constraint.

cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
```diff
@@ -111,6 +122,20 @@ func (r *MCPExternalAuthConfigReconciler) Reconcile(ctx context.Context, req ctr
...
+	// handler returns obo.ErrEnterpriseRequired so upstream-only builds surface
+	// Valid=False / Reason=EnterpriseRequired here rather than failing later
+	// inside a consumer reconciler with a generic "unsupported" error.
+	if externalAuthConfig.Spec.Type == mcpv1beta1.ExternalAuthTypeOBO {
+		if err := ctrlutil.OBOValidate(externalAuthConfig); err != nil {
+			reason := mcpv1beta1.ConditionReasonInvalidConfig
+			if stderrors.Is(err, obo.ErrEnterpriseRequired) {
+				reason = mcpv1beta1.ConditionReasonEnterpriseRequired
+			}
+			return r.setInvalid(ctx, externalAuthConfig, err, reason)
```
jhrozek: Non-blocking thought on the future enterprise handler — `setInvalid` always returns `(ctrl.Result{}, nil)`, so any non-sentinel error from `OBOValidate` lands `Reason: InvalidConfig` permanently until the user touches the spec. For the upstream stub that's fine (it only ever returns `ErrEnterpriseRequired`), but a future out-of-tree `Validate` doing I/O (Secret read, JWKS, webhook) would get the same treatment on a transient blip — no self-healing.

One way to nail the contract down before the enterprise handler PR — typed error matching the existing `registryapi.Error` / `SpecValidationError` precedent:

```go
// pkg/auth/obo/errors.go
type OBOValidationError struct {
    Message string // verbatim into condition.Message; must be safe to expose
}
func (e *OBOValidationError) Error() string { return e.Message }
```

```go
// the OBO branch
switch {
case stderrors.Is(err, obo.ErrEnterpriseRequired):
    return r.setInvalid(ctx, externalAuthConfig, err, mcpv1beta1.ConditionReasonEnterpriseRequired)
default:
    var valErr *obo.OBOValidationError
    if stderrors.As(err, &valErr) {
        return r.setInvalid(ctx, externalAuthConfig, valErr, mcpv1beta1.ConditionReasonInvalidConfig)
    }
    return ctrl.Result{}, fmt.Errorf("OBO handler validation failed: %w", err)
}
```

Three-way contract: `ErrEnterpriseRequired` → not licensed, `*OBOValidationError` → permanent (user-fix), anything else → transient (requeue). The `Message` field also doubles as a 'this is what shows in `kubectl describe`' hint, so handler authors don't accidentally leak Secret names or internal addressing via `fmt.Errorf` chains.

Same contract should apply to `ApplyRunConfig` and `SecretEnvVars`. No migration concern since no enterprise handler exists yet. Happy to defer to a follow-up issue if you'd rather keep #5328 scoped tight.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -1115,14 +1132,14 @@ func (r *MCPExternalAuthConfig) Validate() error {
...
-		// CRD-level Validate() stays a no-op for OBO and exists only to keep
-		// the `exhaustive` linter happy now that ExternalAuthTypeOBO is
-		// defined. The CRD enum currently rejects "obo" at the apiserver
-		// layer, so this arm is unreachable in upstream-only builds.
+		// OBO validation is delegated to the registered OBO handler at the
+		// controllerutil layer (via controllerutil.OBOValidate ->
+		// obo.ErrEnterpriseRequired in upstream-only builds), invoked from
+		// the reconcile loop. The CRD-level Validate() stays a no-op for OBO
+		// and exists only to keep the `exhaustive` linter happy now that
+		// ExternalAuthTypeOBO is defined. The CRD enum currently rejects
```
jhrozek, line 1140: Just double-checking #5329 (the CRD enum update) is on the same release train — wanted to make sure it didn't slip through. As written, this whole dispatch surface is unreachable in production until `obo` lands in the enum at line 69 (the apiserver rejects every `obo`-typed config first), so an admin trying to apply one between this PR and #5329 would get `spec.type: Unsupported value` with no CRD status to explain that an enterprise build would have served them better.

jhrozek (dismissed):
LGTM. Two non-blocking comments inline — both about the future enterprise handler contract / release sequencing, neither needs to land in this PR.

cmd/thv-operator/controllers/mcpexternalauthconfig_controller.go
tgrunnagle: Addressed in 3504141e. Implemented the three-bucket contract you proposed:

- `pkg/auth/obo.ValidationError` (named to avoid the package-prefix stutter — mirrors the operator's `SpecValidationError` precedent) with a `Message string` field exposed verbatim via `Error()`.
- The `OBOHandler` type doc comment documents the contract: `ErrEnterpriseRequired` → permanent EnterpriseRequired; `*obo.ValidationError` → permanent InvalidConfig with the typed Message exposed verbatim; anything else → transient, reconciler returns the error so controller-runtime requeues with backoff.
- The reconciler branch uses `errors.Is` / `errors.As` for the triage (so a wrapped `*ValidationError` is also matched). Extracted into a `triageOBOValidation` helper so `Reconcile` stayed under the gocyclo threshold.
- New table-driven test `OBO_ErrorTriageInReconcile` covers all five paths: bare sentinel, wrapped sentinel, bare `*ValidationError`, wrapped `*ValidationError`, and the transient case (asserts `Reconcile` returns the error AND writes no Valid condition).
- The doc note on safe-to-expose `Message` content (no Secret names, no internal addressing) is on both `pkg/auth/obo/errors.go` and the `OBOHandler` doc.

Same contract applies to `ApplyRunConfig` and `SecretEnvVars` per the OBOHandler doc; their default implementations already return `ErrEnterpriseRequired`, and there are no other consumers downstream of those dispatch sites that need to triage today — the contract is published so the enterprise handler PR can rely on it from day one.

jhrozek (approved):

PR #5348, opened 2026-05-20
Wire CIMD config through embedded AS and enable storage decorator

amirejaz:
## Summary

Phase 2 PR 3 of CIMD embedded AS support — depends on #5343.

This is the wiring PR. All CIMD logic lives in PRs 1 and 2; this PR threads the configuration through the existing authserver config chain and activates the decorator.

**Config chain:** `RunConfig.CIMD` → `Config.CIMD*` → `AuthorizationServerParams` → `AuthorizationServerConfig` → discovery handler

With this PR, operators can enable CIMD by setting:
```yaml
authServer:
  cimd:
    enabled: true
    cache_max_size: 256       # optional, default 256
    cache_fallback_ttl: 5m    # optional, default 5m
```

When enabled:
- The embedded AS discovery documents advertise `client_id_metadata_document_supported: true` on both `/.well-known/oauth-authorization-server` and `/.well-known/openid-configuration`
- `CIMDStorageDecorator` wraps the storage so `GetClient` intercepts HTTPS `client_id` values, fetches the metadata document, and caches the result

CIMD is **opt-in** (disabled by default) to avoid introducing outbound HTTPS fetching in existing deployments without explicit operator action.

## Key implementation note

The decorator is applied **after** `runLegacyMigration` (which needs the unwrapped `*RedisStorage` for type assertion) and **before** `createProvider` (so fosite holds the decorated storage and HTTPS `client_id` values are intercepted during `GetClient`).

## Known limitations

- **Operator CRD not updated**: `MCPExternalAuthConfig` has no `authServer.cimd` field yet. Kubernetes operators must configure CIMD via `RunConfig` YAML directly. CRD support is deferred to a follow-up (tracked with `// TODO(cimd)` comments in config.go).

## Type of change

- [x] New feature (non-breaking, opt-in only)

## Test plan

- [x] `go test ./pkg/authserver/...` — all 13 packages pass
- [x] Config validation: `CIMDCacheMaxSize < 1` rejected at boot when enabled
- [x] Defaults applied: 256 entries, 5 min fallback TTL when not specified
- [x] Discovery tests: `client_id_metadata_document_supported` present/absent based on `CIMDEnabled`
- [x] `resolveCIMDConfig` nil-guard tested

### DIFF
--- a/pkg/authserver/config.go
+++ b/pkg/authserver/config.go
@@ -90,6 +90,11 @@ type RunConfig struct {
 	// Storage configures the storage backend for the auth server.
 	// If nil, defaults to in-memory storage.
 	Storage *storage.RunConfig `json:"storage,omitempty" yaml:"storage,omitempty"`
+
+	// CIMD controls client_id metadata document support. When enabled, the
+	// embedded authorization server accepts HTTPS URLs as client_id values
+	// and resolves them via the CIMD protocol instead of requiring DCR.
+	CIMD *CIMDRunConfig `json:"cimd,omitempty" yaml:"cimd,omitempty"`
 }
 
 // Validate checks that the on-disk RunConfig is internally consistent. Called
@@ -116,6 +121,22 @@ func (c *RunConfig) validateBaselineClientScopes() error {
 		effective = registration.DefaultScopes
 	}
 	return registration.ValidateScopeSubset(c.BaselineClientScopes, effective, "baseline_client_scopes")
+}
+
+// CIMDRunConfig controls client_id metadata document (CIMD) support.
+type CIMDRunConfig struct {
+	// Enabled activates CIMD client lookup when true.
+	Enabled bool `json:"enabled" yaml:"enabled"`
+
+	// CacheMaxSize is the maximum number of CIMD documents held in the LRU cache.
+	// Defaults to 256 when Enabled is true and this field is zero.
+	CacheMaxSize int `json:"cache_max_size,omitempty" yaml:"cache_max_size,omitempty"`
+
+	// CacheFallbackTTL is how long a cached CIMD document is considered valid when
+	// the fetched document carries no Cache-Control header.
+	// Defaults to 5 minutes when Enabled is true and this field is zero.
+	//nolint:lll // field tags require full JSON+YAML names
+	CacheFallbackTTL time.Duration `json:"cache_fallback_ttl,omitempty" yaml:"cache_fallback_ttl,omitempty" swaggertype:"string" example:"5m"`
 }
 
 // SigningKeyRunConfig configures where to load signing keys from.
@@ -537,6 +558,20 @@ type Config struct {
 	// When empty, any request with a "resource" parameter will be rejected with
 	// "invalid_target". Configure this for proper MCP specification compliance.
 	AllowedAudiences []string
+
+	// CIMDEnabled enables the CIMD storage decorator so the authorization server
+	// accepts HTTPS URLs as client_id values without prior DCR registration.
+	CIMDEnabled bool
+
+	// CIMDCacheMaxSize is the maximum number of CIMD documents held in the LRU
+	// cache. Zero is replaced by a default (256) in applyDefaults when CIMDEnabled
+	// is true.
+	CIMDCacheMaxSize int
+
+	// CIMDCacheFallbackTTL is the TTL applied to cached CIMD documents that carry
+	// no Cache-Control header. Zero is replaced by a default (5 minutes) in
+	// applyDefaults when CIMDEnabled is true.
+	CIMDCacheFallbackTTL time.Duration
 }
 
 // Validate checks that the Config is valid.
@@ -587,6 +622,10 @@ func (c *Config) Validate() error {
 		if err := registration.ValidateScopeSubset(c.BaselineClientScopes, effective, "baseline_client_scopes"); err != nil {
 			return err
 		}
+	}
+
+	if c.CIMDEnabled && c.CIMDCacheMaxSize < 1 {
+		return fmt.Errorf("cimd.cache_max_size must be >= 1 when CIMD is enabled")
 	}
 
 	slog.Debug("authserver config validation passed",
@@ -818,6 +857,12 @@ func (c *Config) applyDefaults() error {
 	if len(c.ScopesSupported) == 0 {
 		c.ScopesSupported = registration.DefaultScopes
 		slog.Debug("applied default scopes_supported", "scopes", c.ScopesSupported)
+	}
+	if c.CIMDEnabled && c.CIMDCacheMaxSize == 0 {
+		c.CIMDCacheMaxSize = 256
+	}
+	if c.CIMDEnabled && c.CIMDCacheFallbackTTL == 0 {
+		c.CIMDCacheFallbackTTL = 5 * time.Minute
 	}
 	return nil
 }
--- a/pkg/authserver/config.go
+++ b/pkg/authserver/config.go
@@ -124,6 +124,10 @@ func (c *RunConfig) validateBaselineClientScopes() error {
 }
 
 // CIMDRunConfig controls client_id metadata document (CIMD) support.
+//
+// TODO(cimd): expose these fields in the MCPExternalAuthConfig CRD so Kubernetes
+// operators can configure CIMD through the normal CRD workflow instead of
+// writing RunConfig YAML directly.
 type CIMDRunConfig struct {
 	// Enabled activates CIMD client lookup when true.
 	Enabled bool `json:"enabled" yaml:"enabled"`
@@ -132,8 +136,8 @@ type CIMDRunConfig struct {
 	// Defaults to 256 when Enabled is true and this field is zero.
 	CacheMaxSize int `json:"cache_max_size,omitempty" yaml:"cache_max_size,omitempty"`
 
-	// CacheFallbackTTL is how long a cached CIMD document is considered valid when
-	// the fetched document carries no Cache-Control header.
+	// CacheFallbackTTL is the fixed TTL applied to every cached CIMD document.
+	// Cache-Control header parsing is not yet implemented; all entries use this value.
 	// Defaults to 5 minutes when Enabled is true and this field is zero.
 	//nolint:lll // field tags require full JSON+YAML names
 	CacheFallbackTTL time.Duration `json:"cache_fallback_ttl,omitempty" yaml:"cache_fallback_ttl,omitempty" swaggertype:"string" example:"5m"`
--- a/pkg/authserver/config.go
+++ b/pkg/authserver/config.go
@@ -572,9 +572,9 @@ type Config struct {
 	// is true.
 	CIMDCacheMaxSize int
 
-	// CIMDCacheFallbackTTL is the TTL applied to cached CIMD documents that carry
-	// no Cache-Control header. Zero is replaced by a default (5 minutes) in
-	// applyDefaults when CIMDEnabled is true.
+	// CIMDCacheFallbackTTL is the fixed TTL applied to all cached CIMD documents
+	// (Cache-Control header parsing is not yet implemented). Zero is replaced by
+	// a default (5 minutes) in applyDefaults when CIMDEnabled is true.
 	CIMDCacheFallbackTTL time.Duration
 }
 
@@ -630,6 +630,9 @@ func (c *Config) Validate() error {
 
 	if c.CIMDEnabled && c.CIMDCacheMaxSize < 1 {
 		return fmt.Errorf("cimd.cache_max_size must be >= 1 when CIMD is enabled")
+	}
+	if c.CIMDEnabled && c.CIMDCacheFallbackTTL < 0 {
+		return fmt.Errorf("cimd.cache_fallback_ttl must be non-negative when CIMD is enabled")
 	}
 
 	slog.Debug("authserver config validation passed",
--- a/pkg/authserver/config.go
+++ b/pkg/authserver/config.go
@@ -102,6 +102,11 @@ type RunConfig struct {
 // catches operator-supplied misconfiguration early so server startup fails
 // loudly instead of degrading silently at runtime.
 func (c *RunConfig) Validate() error {
+	if c.CIMD != nil {
+		if err := c.CIMD.Validate(); err != nil {
+			return fmt.Errorf("cimd: %w", err)
+		}
+	}
 	return c.validateBaselineClientScopes()
 }
 
@@ -138,9 +143,29 @@ type CIMDRunConfig struct {
 
 	// CacheFallbackTTL is the fixed TTL applied to every cached CIMD document.
 	// Cache-Control header parsing is not yet implemented; all entries use this value.
-	// Defaults to 5 minutes when Enabled is true and this field is zero.
-	//nolint:lll // field tags require full JSON+YAML names
-	CacheFallbackTTL time.Duration `json:"cache_fallback_ttl,omitempty" yaml:"cache_fallback_ttl,omitempty" swaggertype:"string" example:"5m"`
+	// Format: Go duration string (e.g. "5m", "10m", "1h").
+	// Defaults to 5 minutes when Enabled is true and this field is omitted.
+	CacheFallbackTTL string `json:"cache_fallback_ttl,omitempty" yaml:"cache_fallback_ttl,omitempty" example:"5m"`
+}
+
+// Validate checks that the CIMDRunConfig fields are internally consistent.
+func (c *CIMDRunConfig) Validate() error {
+	if !c.Enabled {
+		return nil
+	}
+	if c.CacheMaxSize < 0 {
+		return fmt.Errorf("cache_max_size must be non-negative when CIMD is enabled, got %d", c.CacheMaxSize)
+	}
+	if c.CacheFallbackTTL != "" {
+		d, err := time.ParseDuration(c.CacheFallbackTTL)
+		if err != nil {
+			return fmt.Errorf("cache_fallback_ttl: %w", err)
+		}
+		if d < 0 {
+			return fmt.Errorf("cache_fallback_ttl must be non-negative when CIMD is enabled, got %s", c.CacheFallbackTTL)
+		}
+	}
+	return nil
 }
 
 // SigningKeyRunConfig configures where to load signing keys from.
@@ -867,9 +892,11 @@ func (c *Config) applyDefaults() error {
 	}
 	if c.CIMDEnabled && c.CIMDCacheMaxSize == 0 {
 		c.CIMDCacheMaxSize = 256
+		slog.Debug("applied default cimd cache_max_size", "size", c.CIMDCacheMaxSize)
 	}
 	if c.CIMDEnabled && c.CIMDCacheFallbackTTL == 0 {
 		c.CIMDCacheFallbackTTL = 5 * time.Minute
+		slog.Debug("applied default cimd cache_fallback_ttl", "ttl", c.CIMDCacheFallbackTTL)
 	}
 	return nil
 }
--- a/pkg/authserver/config_test.go
+++ b/pkg/authserver/config_test.go
@@ -114,6 +114,13 @@ func TestConfigValidate(t *testing.T) {
 		{name: "baseline scope not in scopes_supported", config: Config{Issuer: "https://example.com", KeyProvider: validKeyProvider, HMACSecrets: validHMAC, Upstreams: validUpstreams, AllowedAudiences: []string{"https://mcp.example.com"}, ScopesSupported: []string{"openid"}, BaselineClientScopes: []string{"offline_access"}}, wantErr: true, errMsg: `baseline_client_scopes contains "offline_access"`},
 		{name: "nil supported with baseline in DefaultScopes passes", config: Config{Issuer: "https://example.com", KeyProvider: validKeyProvider, HMACSecrets: validHMAC, Upstreams: validUpstreams, AllowedAudiences: []string{"https://mcp.example.com"}, ScopesSupported: nil, BaselineClientScopes: []string{"offline_access"}}},
 
+		// CIMD validation
+		{name: "CIMD enabled zero cache_max_size rejected", config: Config{Issuer: "https://example.com", KeyProvider: validKeyProvider, HMACSecrets: validHMAC, Upstreams: validUpstreams, AllowedAudiences: []string{"https://mcp.example.com"}, CIMDEnabled: true, CIMDCacheMaxSize: 0}, wantErr: true, errMsg: "cache_max_size must be >= 1"},
+		{name: "CIMD enabled negative cache_max_size rejected", config: Config{Issuer: "https://example.com", KeyProvider: validKeyProvider, HMACSecrets: validHMAC, Upstreams: validUpstreams, AllowedAudiences: []string{"https://mcp.example.com"}, CIMDEnabled: true, CIMDCacheMaxSize: -1}, wantErr: true, errMsg: "cache_max_size must be >= 1"},
+		{name: "CIMD enabled negative cache_fallback_ttl rejected", config: Config{Issuer: "https://example.com", KeyProvider: validKeyProvider, HMACSecrets: validHMAC, Upstreams: validUpstreams, AllowedAudiences: []string{"https://mcp.example.com"}, CIMDEnabled: true, CIMDCacheMaxSize: 256, CIMDCacheFallbackTTL: -time.Second}, wantErr: true, errMsg: "cache_fallback_ttl must be non-negative"},
+		{name: "CIMD disabled ignores invalid cache fields", config: Config{Issuer: "https://example.com", KeyProvider: validKeyProvider, HMACSecrets: validHMAC, Upstreams: validUpstreams, AllowedAudiences: []string{"https://mcp.example.com"}, CIMDEnabled: false, CIMDCacheMaxSize: -1, CIMDCacheFallbackTTL: -time.Second}},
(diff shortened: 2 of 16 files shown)


pkg/authserver/config.go
```diff
@@ -118,6 +123,26 @@ func (c *RunConfig) validateBaselineClientScopes() error {
...
+
+	// CacheMaxSize is the maximum number of CIMD documents held in the LRU cache.
+	// Defaults to 256 when Enabled is true and this field is zero.
+	CacheMaxSize int `json:"cache_max_size,omitempty" yaml:"cache_max_size,omitempty"`
+
+	// CacheFallbackTTL is the fixed TTL applied to every cached CIMD document.
+	// Cache-Control header parsing is not yet implemented; all entries use this value.
+	// Defaults to 5 minutes when Enabled is true and this field is zero.
+	//nolint:lll // field tags require full JSON+YAML names
+	CacheFallbackTTL time.Duration `json:"cache_fallback_ttl,omitempty" yaml:"cache_fallback_ttl,omitempty" swaggertype:"string" example:"5m"`
```
tgrunnagle: **[HIGH] `time.Duration` field cannot parse `"5m"` from YAML/JSON**

Declaring `CacheFallbackTTL time.Duration` with `swaggertype:"string" example:"5m"` advertises a string wire format that Go's standard JSON/YAML decoders cannot actually consume — `time.Duration` is an `int64` alias with no custom unmarshaller. Operators submitting `cache_fallback_ttl: "5m"` will get a decode error; integer nanoseconds is what actually parses.

The rest of this repo solves this correctly. Compare:
- `TokenLifespanRunConfig.AccessTokenLifespan string` (config.go:144) + `parseTokenLifespans` in `runner/embeddedauthserver.go:345`
- `RedisRunConfig.DialTimeout/ReadTimeout/WriteTimeout string` + `time.ParseDuration` in `runner/embeddedauthserver.go:739/746/753`

This is the only `time.Duration` field declared directly on a `RunConfig` type in the project.

```suggestion
	//nolint:lll // field tags require full JSON+YAML names
	CacheFallbackTTL string `json:"cache_fallback_ttl,omitempty" yaml:"cache_fallback_ttl,omitempty" example:"5m"`
```

Then parse via `time.ParseDuration` in a sibling of `parseTokenLifespans` and feed the resulting `time.Duration` into `Config.CIMDCacheFallbackTTL`. Add a round-trip test that loads `cache_fallback_ttl: "5m"` from YAML.

tgrunnagle (commented):
## Multi-agent code review

This review was produced by 5 specialist agents (security, Go correctness, architecture, tests, code quality) plus consensus scoring. Codex cross-review was skipped (codex CLI not installed).

### Executive summary

This is the wiring stage of a multi-PR CIMD sequence: `RunConfig.CIMD` → `Config.CIMD*` → `AuthorizationServerParams.CIMDEnabled` → `AuthorizationServerConfig.CIMDEnabled` → discovery handler. The decorator wrap is performed in `newServer` after the legacy Redis migration and before fosite provider creation. The approach is internally coherent and opt-in (default disabled), but the configuration plumbing diverges from the codebase's established `string`→`time.ParseDuration` pattern for durations, so the swagger-documented `cache_fallback_ttl: "5m"` YAML cannot actually be unmarshalled. The PR also surfaces several pre-existing CIMD security gaps that landing this opt-in wiring will expose publicly via the discovery flag.

### Consensus summary

| 1 | HIGH | `pkg/authserver/config.go` | `time.Duration` field cannot parse `"5m"` from YAML/JSON |
| 2 | MEDIUM | `pkg/oauthproto/cimd.go` | Loopback HTTP CIMD fetch — SSRF/port-scan oracle |
| 3 | MEDIUM | `pkg/authserver/storage/cimd_decorator.go` | CIMD scope handling lacks `ScopesSupported`/`BaselineClientScopes` validation |
| 4 | MEDIUM | `pkg/authserver/storage/cimd_decorator.go` | CIMD doc `grant_types`/`response_types` not validated against AS allowlist |
| 5 | MEDIUM | `pkg/authserver/config.go` | Fixed cache TTL ignores `Cache-Control` — revocation window |
| 6 | MEDIUM | `pkg/authserver/config.go` | `applyDefaults` silently rewrites explicit zero values |
| 7 | MEDIUM | `pkg/authserver/config.go` | `RunConfig.Validate` does not validate CIMD |
| 8 | MEDIUM | `pkg/authserver/config.go` | Four parallel CIMD config representations will drift |
| 9 | MEDIUM | `pkg/authserver/server_impl.go` | Wrap decision and discovery flag derived independently |
| 10 | MEDIUM | `pkg/authserver/config_test.go` | Missing `Validate()` tests for new CIMD validation rules |
| 11 | MEDIUM | `pkg/authserver/server_impl.go` | No test verifies decorator is installed when `CIMDEnabled=true` |
| 12 | LOW | `pkg/authserver/server/handlers/discovery.go` | OIDC discovery advertises `client_id_metadata_document_supported` (off-spec) |
| 13 | LOW | `pkg/authserver/server_impl.go` | Decorated storage masks underlying `*RedisStorage` |
| 14 | LOW | `pkg/oauthproto/cimd.go` | `IsClientIDMetadataDocumentURL` widening changes pkg/auth/remote behavior |

### Pros

- Opt-in by default — existing deployments see no behavior change.
- `omitempty` on the discovery flag preserves backward compatibility, verified by a raw-map decode test.
- Decorator wrap order is correct: legacy Redis migration and `dcrStore` extraction run *before* the wrap, preserving type-assertion paths.
- Defensive testing of loopback subdomain bypasses (`localhost.evil.com`, `127.0.0.1.evil.com`).
- Follows repo conventions: SPDX headers, `t.Parallel()`, table-driven tests, doc-commented intent.

### Cons

- `CIMDRunConfig.CacheFallbackTTL time.Duration` breaks the repo's established `string`→`time.ParseDuration` pattern (compare `TokenLifespanRunConfig`/`RedisRunConfig`); the advertised `"5m"` YAML will fail to decode.
- CIMD lacks the `ScopesSupported`/`BaselineClientScopes`/`grant_types`/`response_types` allowlist validation that DCR enforces — privilege divergence between two registration paths reaching the same AS.
- `applyDefaults` silently rewrites explicit zero values; a typo'd `cache_max_size: 0` becomes 256 with no log.
- Four parallel representations of CIMD config violate `.claude/rules/go-style.md` "Avoid parallel types that drift".
- Wrap decision (`Config.CIMDEnabled` in `newServer`) and discovery flag (`AuthorizationServerConfig.CIMDEnabled` in handler) are derived independently — invariant is implicit, not enforced.
- `MCPExternalAuthConfig` CRD has no CIMD field — feature inaccessible through the Kubernetes operator workflow.
- No integration test confirms the decorator is actually installed when `CIMDEnabled=true`.

### How to verify the HIGH

```go
// In a *_test.go file inside pkg/authserver:
func TestCIMDFallbackTTL_FromYAMLString(t *testing.T) {
    var rc authserver.RunConfig
    src := []byte("cimd:\n  enabled: true\n  cache_fallback_ttl: \"5m\"\n")
    if err := yaml.Unmarshal(src, &rc); err != nil {
        t.Fatalf("yaml.Unmarshal: %v", err)  // expected to fail on this PR
    }
    if rc.CIMD == nil || rc.CIMD.CacheFallbackTTL != 5*time.Minute {
        t.Fatalf("want 5m, got %v", rc.CIMD.CacheFallbackTTL)
    }
}
```

### Recommendation

**REQUEST CHANGES** at minimum for finding #1 (HIGH). Several MEDIUMs (#3/#4 scope/grant validation parity, #6 silent zero rewrite, #11/#12 missing wiring tests) are worth resolving before merge. Findings #2/#5 (loopback SSRF, fixed TTL) are CIMD-substrate concerns that could be acknowledged in this PR and tracked separately.

---

*Below: 14 inline comments mapped to the diff. A separate dropped-findings appendix (~20 LOW items) is retained in the local review and can be shared if helpful.*

pkg/oauthproto/cimd.go
```diff
@@ -12,11 +15,27 @@ import "strings"
...
+	if strings.HasPrefix(clientID, "https://") {
+		return true
+	}
+	// Allow loopback HTTP URLs (localhost, 127.0.0.1, [::1]) for local
+	// development and integration testing. These are the only HTTP URLs that
+	// FetchClientMetadataDocument / validateCIMDClientURL also accept.
+	if strings.HasPrefix(clientID, "http://") {
+		parsed, err := url.Parse(clientID)
+		if err != nil {
+			return false
```
tgrunnagle, line 41: **[MEDIUM] Loopback HTTP acceptance creates an SSRF/port-scan oracle**

The custom dialer in `pkg/oauthproto/cimd/fetch.go` explicitly bypasses `networking.IsPrivateIP` for `IsLoopback` addresses. With this widening, any fosite `GetClient(clientID)` where `clientID` is `http://127.0.0.1:6379/x` or `http://localhost:8500/y` provokes an outbound dial to that host:port from the AS process.

If the embedded AS is reachable beyond loopback (kube Service, ingress), an unauthenticated attacker can use `GET /oauth/authorize?client_id=http://127.0.0.1:N/path` as a confused-deputy probe to fingerprint internal services on the AS host. The 5s timeout, 10KB cap, redirect-blocking, and document-shape validation bound exfil but the timing/error oracles are sufficient for slow scanning.

The scheme-based check (URL form) is the wrong control point. Either:
- gate loopback HTTP acceptance behind an explicit `CIMDAllowLoopbackHTTP bool` defaulting to `false` in production, or
- only enable loopback acceptance when running under an explicit dev/test flag.

Also tighten the doc comment from "local development and integration tests" to state the production risk explicitly.

pkg/authserver/storage/cimd_decorator.go
```diff
@@ -172,7 +173,11 @@ func buildFositeClient(doc *cimd.ClientMetadataDocument) fosite.Client {
 		tokenEndpointAuthMethod = defaultCIMDTokenEndpointAuthMethod
 	}
 
-	var scopes []string
+	// When the document omits the scope field, apply the same defaults as DCR
+	// registration so CIMD clients can request openid/profile/email/offline_access
+	// without needing to enumerate them explicitly in the metadata document.
+	// Clone to avoid aliasing the package-level DefaultScopes slice.
+	scopes := slices.Clone(registration.DefaultScopes)
```
tgrunnagle, line 180: **[MEDIUM] CIMD scope handling diverges from DCR — privilege escalation risk**

When the document omits `scope`, this builds the client with `registration.DefaultScopes`. The DCR equivalent runs each scope through `registration.ValidateScopes` against the operator-configured `ScopesSupported` and unions in `BaselineClientScopes`. The CIMD path has no access to those (the decorator was constructed without them).

Concretely: operator sets `ScopesSupported=["openid"]` to disable profile-data exposure; DCR will reject a `DefaultScopes` request; CIMD will still grant `profile`/`email`.

Thread `ScopesSupported`/`BaselineClientScopes` into `NewCIMDStorageDecorator` parameters and validate the document's scope set (or `DefaultScopes` when empty) through `registration.ValidateScopes` before constructing the fosite client. Add a test mirroring the DCR scope-validation cases.

tgrunnagle, line 180: **[MEDIUM] CIMD documents' `grant_types`/`response_types` not validated against the AS allowlist**

`buildFositeClient` accepts the doc's `GrantTypes` and `ResponseTypes` verbatim (only defaulting when absent). DCR's `validateGrantTypes`/`validateResponseTypes` restrict public clients to `authorization_code+refresh_token` and `code`, returning `invalid_client_metadata` for anything else. A CIMD document declaring `grant_types=["client_credentials"]` or `response_types=["token"]` loads here without rejection.

The `token_endpoint_auth_method` check (forbiddenAuthMethods, in `fetch()`) is the right shape — extend the same pattern to `grant_types` and `response_types` so misconfigured CIMD docs fail fast at fetch time, not at `/oauth/token` time.

pkg/authserver/config.go
```diff
@@ -118,6 +123,26 @@ func (c *RunConfig) validateBaselineClientScopes() error {
...
+	// Enabled activates CIMD client lookup when true.
+	Enabled bool `json:"enabled" yaml:"enabled"`
+
+	// CacheMaxSize is the maximum number of CIMD documents held in the LRU cache.
+	// Defaults to 256 when Enabled is true and this field is zero.
+	CacheMaxSize int `json:"cache_max_size,omitempty" yaml:"cache_max_size,omitempty"`
+
+	// CacheFallbackTTL is the fixed TTL applied to every cached CIMD document.
+	// Cache-Control header parsing is not yet implemented; all entries use this value.
+	// Defaults to 5 minutes when Enabled is true and this field is zero.
```
tgrunnagle: **[MEDIUM] Fixed cache TTL ignores `Cache-Control` — revocation window**

The field name `CacheFallbackTTL` implies header parsing was intended, but `cimdCacheEntry.expires` is always `time.Now().Add(d.ttl)`. The CIMD document author cannot signal "I rotated my redirect_uris; please refresh now".

If a CIMD client revokes a redirect URI, the AS continues to honor the old `redirect_uris` for up to 5 minutes (default), widening the stolen-authcode replay window. The CIMD draft contemplates `Cache-Control` directing TTL.

Either parse `Cache-Control: max-age` and use `min(header, fallbackTTL)`, or at minimum recommend a low default (e.g. 60s) in the field comment for environments where metadata churn matters. Add a tracked `TODO` referencing the future work.

pkg/authserver/config.go
```diff
@@ -819,6 +865,12 @@ func (c *Config) applyDefaults() error {
 		c.ScopesSupported = registration.DefaultScopes
 		slog.Debug("applied default scopes_supported", "scopes", c.ScopesSupported)
 	}
+	if c.CIMDEnabled && c.CIMDCacheMaxSize == 0 {
```
tgrunnagle, line 893: **[MEDIUM] `applyDefaults` silently rewrites explicit zero values**

When the operator writes `cache_max_size: 0` in YAML (typo, or intending "disable caching"), this silently sets it to 256; the `Validate()` check (`< 1`) is bypassed because the rewrite happens first. The sibling defaults block at lines 794/798/802 logs each substitution via `slog.Debug` — this block does not.

Per `.claude/rules/go-style.md` "Constructor Validation: Fail Loudly on Invalid Input — Don't let zero silently mean 'unlimited' or 'disabled'". JSON/YAML unmarshalling cannot distinguish "field absent" from "field explicitly zero" with `int` fields.

Options:
1. Switch `CacheMaxSize`/`CacheFallbackTTL` to `*int`/`*time.Duration` so the runner can distinguish absent vs explicit zero.
2. Add `slog.Debug("applied default CIMD cache_max_size", "size", c.CIMDCacheMaxSize)` matching the sibling defaults pattern so substitutions are auditable.

_Raised by 3 agents (security, go-correctness, architecture)._

pkg/authserver/config.go
```diff
@@ -118,6 +123,26 @@ func (c *RunConfig) validateBaselineClientScopes() error {
 	return registration.ValidateScopeSubset(c.BaselineClientScopes, effective, "baseline_client_scopes")
 }
 
+// CIMDRunConfig controls client_id metadata document (CIMD) support.
+//
+// TODO(cimd): expose these fields in the MCPExternalAuthConfig CRD so Kubernetes
+// operators can configure CIMD through the normal CRD workflow instead of
+// writing RunConfig YAML directly.
+type CIMDRunConfig struct {
```
tgrunnagle, line 136: **[MEDIUM] `RunConfig.Validate` does not validate CIMD**

`RunConfig.Validate()` only calls `validateBaselineClientScopes()` and ignores the new `CIMD` field. The function's doc claims it "catches operator-supplied misconfiguration early so server startup fails loudly". A negative `cache_fallback_ttl` or invalid combination in operator YAML therefore propagates through to runtime `Config.Validate()` rather than failing at the natural wire-format boundary.

Any future CRD admission webhook that uses `RunConfig.Validate()` as the authoritative check will accept invalid CIMD configs.

Add a `CIMDRunConfig.Validate()` method and call it from `RunConfig.Validate()` alongside `validateBaselineClientScopes()`.

pkg/authserver/config.go
```diff
@@ -537,6 +562,20 @@ type Config struct {
 	// When empty, any request with a "resource" parameter will be rejected with
 	// "invalid_target". Configure this for proper MCP specification compliance.
 	AllowedAudiences []string
+
+	// CIMDEnabled enables the CIMD storage decorator so the authorization server
+	// accepts HTTPS URLs as client_id values without prior DCR registration.
+	CIMDEnabled bool
```
tgrunnagle, line 593: **[MEDIUM] Four parallel CIMD config representations will drift**

A single CIMD configuration concern is now expressed as four distinct types:
1. `CIMDRunConfig` (nested, YAML-facing, 3 fields)
2. `Config.CIMDEnabled/CIMDCacheMaxSize/CIMDCacheFallbackTTL` (flat, in-memory)
3. `AuthorizationServerParams.CIMDEnabled` (1 field)
4. `AuthorizationServerConfig.CIMDEnabled` (1 field)

The sibling `RunConfig.Storage *storage.RunConfig` is the project's accepted pattern: passed through as a pointer rather than flattened and re-aggregated.

Per `.claude/rules/go-style.md`: "Avoid parallel types that drift — don't define a separate config/data type that mirrors an existing one. Embed or reuse the original."

Hold the nested struct on `Config` as `CIMD *CIMDConfig` (or reuse `CIMDRunConfig` directly). Then `resolveCIMDConfig` can return a single struct pointer instead of a 3-tuple, and each future CIMD field requires touching one type definition rather than four.

_Raised independently by go-correctness and architecture specialists._

pkg/authserver/server_impl.go
```diff
@@ -173,6 +170,20 @@ func newServer(ctx context.Context, cfg Config, stor storage.Storage, opts ...se
 		return nil, err
 	}
 
+	// Wrap storage with the CIMD decorator before constructing the fosite provider
+	// so that GetClient calls for HTTPS client_id values are intercepted at the
+	// fosite level (not just the handler level).
+	if cfg.CIMDEnabled {
```
tgrunnagle, line 179: **[MEDIUM] Wrap decision and discovery flag derived independently**

`cfg.CIMDEnabled` is the source of truth used in two places: (1) here, to wrap the storage; (2) in `oauthParams` above, to populate `AuthorizationServerConfig.CIMDEnabled` which the discovery handler reads to set `client_id_metadata_document_supported`. A future change with a different condition (e.g., "CIMD enabled AND a CIMD-aware upstream is configured") could leave the discovery flag advertising support that isn't installed, or vice versa.

Consider moving the wrap step inside `NewAuthorizationServerConfig` (or a sibling helper that returns both the config and the decorated storage), so the same code path that decides "advertise CIMD in discovery" also decides "decorate storage with CIMD". The invariant "discovery advertises iff decorator wraps" should be enforced syntactically rather than relying on two distant derivations of the same bit.

pkg/authserver/config_test.go
```diff
@@ -589,3 +589,49 @@ func TestConfigApplyDefaults_BaselineClientScopes(t *testing.T) {
 		})
 	}
 }
+
+func TestConfigApplyDefaults_CIMD(t *testing.T) {
```
tgrunnagle, line 608: **[MEDIUM] Missing `Validate()` tests for new CIMD validation rules**

`Config.Validate()` adds two new error paths (`CIMDEnabled && CIMDCacheMaxSize<1` and `CIMDEnabled && CIMDCacheFallbackTTL<0`) but `TestConfigValidate` has no cases exercising them. The validators are dead-untested code — a regression that loosens the bound or removes the check would not be caught.

Add four cases to the existing `TestConfigValidate` table:
1. `CIMDEnabled=true, CIMDCacheMaxSize=0` → error containing `"cache_max_size must be >= 1"`
2. `CIMDCacheMaxSize=-1` → same error
3. `CIMDCacheFallbackTTL=-1s` → error containing `"cache_fallback_ttl must be non-negative"`
4. Valid bounds → passes

Also consider a "CIMDEnabled=false with negative cache fields passes" case to document that the gate truly depends on `CIMDEnabled`.

pkg/authserver/server_impl.go
```diff
@@ -173,6 +170,20 @@ func newServer(ctx context.Context, cfg Config, stor storage.Storage, opts ...se
...
+		stor, err = storage.NewCIMDStorageDecorator(stor, true, cfg.CIMDCacheMaxSize, cfg.CIMDCacheFallbackTTL)
```
tgrunnagle, line 180: **[MEDIUM] No test verifies the decorator is actually installed when `CIMDEnabled=true`**

The decorator has thorough tests in PR #5343, and `resolveCIMDConfig` has unit tests here, but no test exercises the full wiring: build an authserver via `newServer` with `CIMDEnabled=true` and confirm `GetClient("https://...")` is intercepted (or, more simply, that `s.storage.(*CIMDStorageDecorator)` succeeds).

A regression swapping `cfg.CIMDEnabled` for `false` here would not be caught. I see `cimd-phase2-pr4-integration-tests` exists on origin, but a seam-level test in *this* PR would catch wiring regressions immediately without waiting.

Minimal approach: add a test that builds a minimal `Config` with `CIMDEnabled=true`, calls `newServer`, and asserts the returned `s.storage` chain contains a `*CIMDStorageDecorator` (use `Unwrap()` to walk it).

pkg/authserver/server/handlers/discovery.go
```diff
@@ -114,6 +114,8 @@ func (h *Handler) buildOAuthMetadata() sharedobauth.AuthorizationServerMetadata
 		},
 		CodeChallengeMethodsSupported:     []string{crypto.PKCEChallengeMethodS256},
 		TokenEndpointAuthMethodsSupported: []string{sharedobauth.TokenEndpointAuthMethodNone},
+
+		ClientIDMetadataDocumentSupported: h.config.CIMDEnabled,
```
tgrunnagle, line 123: **[LOW] `client_id_metadata_document_supported` advertised on OIDC discovery (off-spec)**

`buildOAuthMetadata` is shared by both `/.well-known/oauth-authorization-server` (RFC 8414, where the CIMD extension belongs) and `/.well-known/openid-configuration` (OIDC Discovery 1.0). CIMD is an OAuth WG draft. OIDC Discovery has no `client_id_metadata_document_supported` concept in its spec.

This is not actively harmful (OIDC consumers ignore unknown fields), but it bloats the OIDC document and may confuse strict validators.

Either consciously accept and add a one-line comment in `buildOAuthMetadata` explaining the deliberate co-location, or zero out the field for the OIDC handler before serializing.

_Raised by 2 specialists (security, architecture)._

pkg/authserver/server_impl.go
tgrunnagle, line 180: **[LOW] Decorated storage masks underlying `*RedisStorage`**

After this wrap, `s.storage` is `*CIMDStorageDecorator`. The legacy-migration assertion at line 174 runs BEFORE the wrap (correct, and well-positioned). `dcrStore` is captured at line 110 before the wrap (also correct).

But any *future* code that adds `if rs, ok := h.storage.(*storage.RedisStorage); ok { ... }` will silently take the "not Redis" branch when CIMD is enabled. The decorator does provide `Unwrap() Storage`, but nothing in the current code path uses it.

Mitigation options:
1. Add a doc comment on the `server.storage` field noting that wrappers may hide concrete types and that type assertions must walk `Unwrap()`.
2. Add a regression test that constructs a CIMD-decorated `*RedisStorage` and verifies that downstream code can still reach the Redis-specific methods via the unwrap chain.

_Raised by 2 specialists (security, architecture)._

pkg/oauthproto/cimd.go
tgrunnagle, line 41: **[LOW] Widening this predicate changes `pkg/auth/remote` callers' behavior**

`IsClientIDMetadataDocumentURL` is called from production code in `pkg/auth/remote/handler.go` (two sites, ~lines 209 and 219) that decide whether to persist a `client_id` as DCR credentials vs. as a CIMD URL. Before this PR, an HTTP-loopback URL would have been treated as a DCR client_id and persisted via `clientCredentialsPersister`; after this PR, the same value is now treated as a CIMD URL and assigned to `h.config.CachedCIMDClientID`.

In practice no current deployment uses `http://localhost...` as a DCR client_id (real DCR servers issue opaque IDs), so the change is almost certainly safe. But the doc comment frames the widening only as "local development and integration testing" — a reviewer of `pkg/auth/remote/handler.go` would not realize this predicate now decides persistence semantics differently for loopback HTTP.

Either extend the doc comment to list the known callers and explicitly confirm safety for each, or introduce a narrower predicate (`IsCIMDFetchURL`) used only by the decorator and keep `IsClientIDMetadataDocumentURL` HTTPS-only for the persistence-decision callers.

amirejaz: Fair point. The loopback exception is test-only infrastructure — in production the embedded AS is pod-local. Tracking the opt-in flag as a follow-up to keep this PR scope-bounded.

pkg/authserver/storage/cimd_decorator.go
amirejaz: Agreed — ScopesSupported needs to thread into the decorator for full parity. Deferring to a follow-up that refactors NewCIMDStorageDecorator's signature; would be a broader change than fits here.

amirejaz: Same refactor dependency as the scope issue above. Adding it to the same follow-up.

pkg/authserver/config.go
amirejaz: Already acknowledged in the field comment. The TODO is there; full Cache-Control parsing is a separate piece of work.

pkg/authserver/config.go
amirejaz: Agreed on the direction. Collapsing the flat Config fields into a *CIMDConfig pointer is a refactor that touches multiple packages — keeping it as a tracked follow-up rather than expanding this PR.

pkg/authserver/server_impl.go
amirejaz: Good invariant to enforce. The coupling point is newEmbeddedAuthServerWithStorage — worth consolidating in a follow-up, but not a regression risk with the current single-flag gating.

tgrunnagle (approved):

PR #5350, opened 2026-05-20
Recover OAuth workloads from transient refresh failures

### LINKED ISSUE #5349: OAuth monitor gives up on transient failures, leaving workloads dead
# OAuth monitor gives up on transient failures, leaving workloads dead

## Bug description

When an OAuth token-refresh attempt returns an error that
`isTransientNetworkError` classifies as transient — 5xx, 429, or 4xx
without an RFC 6749 `error` code, per the rule established by #5170 —
`pkg/auth/monitored_token_source.go` runs an in-loop short retry (5
attempts with exponential backoff, ~1–2 minutes at defaults, bounded
by `TOOLHIVE_TOKEN_REFRESH_MAX_TRIES` and
`TOOLHIVE_TOKEN_REFRESH_MAX_ELAPSED_TIME`). If the error persists past
that window, the monitor marks the workload `unauthenticated` and
exits its goroutine. **No further refresh is ever attempted**, even
after the underlying condition clears. The workload stays
`unauthenticated` until manual intervention (`thv restart`, `thv rm`
+ `thv run`, or similar).

The transient classification is correct: the response shape doesn't
carry a definitive "denied" verdict from the OAuth server (no RFC 6749
error code), so ToolHive can't conclude the credentials are bad. The
gap is at the next layer up — the in-loop retry window is too short
to cover realistic recovery time scales for the conditions that
produce these errors (see Additional context). The monitor should
keep trying on a longer cadence, not give up after ~2 minutes.

## Steps to reproduce

Reproduction requires either (a) a real OAuth endpoint behind a network
control point that can be selectively dropped, or (b) the naturally-
occurring real-world trigger described in Additional context (e.g.,
client-side VPN disconnect routing requests through an IP-allowlisted
WAF or CDN).

1. Run an OAuth-backed remote workload with `thv run --remote-url ...`
   and let it complete the initial OAuth flow.
2. Wait for the cached access token to expire (typically 1 hour) so the
   monitor will attempt a refresh.
3. Just before refresh time, block traffic to the token endpoint
   (`pfctl` on macOS / `iptables` on Linux) for several minutes — long
   enough for the short retry to exhaust all 5 attempts (see Additional
   context for the default backoff schedule; ~3 minutes is comfortable).
4. Observe the workload transitions to `unauthenticated`. Restore the
   network — observe that ToolHive does NOT attempt to recover, even
   after waiting an arbitrarily long time.

Note that all conditions for the bug must be met: (a) the failure
classifies as transient (5xx, 429, or 4xx without an RFC 6749 error
code — see #5170), (b) the failure persists past the short-retry
window. Permanent OAuth failures (`invalid_grant`, `invalid_client`)
correctly stop the monitor and are not affected.

## Expected behavior

When a transient token-refresh failure exceeds the short-retry window,
the background monitor should keep attempting refresh on a longer
cadence until either the upstream recovers (→ `running`) or a
configurable ceiling is reached, at which point the workload is finally
marked `unauthenticated`. Workloads should not be permanently broken by
transient failures that resolve within a reasonable ceiling.

## Actual behavior

After the short-retry window exhausts on a still-transient error:

- The retry exhaustion branch in `Token()` (in the `if err != nil`
  block following `refresher.Refresh(...)` in
  `pkg/auth/monitored_token_source.go`) calls `markAsUnauthenticated`.
- `markAsUnauthenticated` writes `WorkloadStatusUnauthenticated` and
  closes the `stopMonitoring` channel.
- The monitor goroutine exits via the `monitorLoop`'s select on
  `stopMonitoring`.
- No further refresh is ever attempted by this workload's monitor.
- The workload remains `unauthenticated` indefinitely.

## Environment (if relevant)

- OS: macOS (also affects Linux — the relevant code path is platform-
  independent).
- ToolHive: current `main` (post-#5170 and post-#5044). The affected
  code paths in `pkg/auth/monitored_token_source.go::Token` and
  `::onTick` have had this shape since the short-retry layer was
  introduced in #4281.

## Additional context

**Real-world trigger:** the canonical scenario is a client-side
network-context change — disconnecting from a corporate VPN, putting
the laptop to sleep on one network and resuming on another, etc.
Token-refresh requests that previously traversed an IP-allowlisted
path now reach the OAuth server from a residential IP, where a WAF or
CDN consistently returns 403+HTML until the trusted path is restored.
The block isn't intermittent from the WAF's perspective — it's a
stable response to a different network origin — but from the
workload's perspective it's a transient failure window that resolves
on its own when the user reconnects.

In one such environment the bug surfaced every 1–3 days; each time the
underlying network state reverted on its own (e.g., morning VPN
reconnect), but the workload stayed `unauthenticated` until manual
recovery.

**Production error shape** (real recurrence):

```
oauth2: cannot fetch token: 403 Forbidden
Response: <!DOCTYPE html>...
```

This is a 4xx-without-RFC-6749-error-code response — correctly
classified as transient by `isTransientNetworkError` after #5170. The
short retry exhausts via the 5-try cap (typically within ~1–2 minutes
given the default backoff: 10s initial interval, 1.5 multiplier, ±50%
randomization, 5 max tries), well before the 5-minute
`MAX_ELAPSED_TIME` cap. The monitor exits at that point.

**Affected code:**

- `pkg/auth/monitored_token_source.go::Token` — the `refresher.Refresh`
  exhaustion branch (the `if err != nil` block right after the
  `refresher.Refresh` call).
- `pkg/auth/monitored_token_source.go::onTick` — calls `Token()`, so
  inherits the same exit-on-exhaustion behavior. The monitor's only
  response to a transient refresh error that outlasts the short retry
  is to mark the workload `unauthenticated` and stop.
- `pkg/auth/monitored_token_source.go::markAsUnauthenticated` is the
  single-shot exit point both call into.

**Related PRs:**

- #5170 — 4xx-without-error-code classification (merged). Necessary
  precondition for this bug to manifest predictably; the short retry
  now correctly retries WAF-shaped responses, then exhausts via the
  5-try cap (~1–2 minutes default).
- #4513 — retry transient errors in the background monitor (merged).
  Refined the short-retry layer.
- #4281 — introduced the short-retry layer in the background monitor.
- #5044 — DCR Warn + `upstream`/`clientID` constructor context
  (merged). A fix here must continue to gate the DCR remediation Warn
  correctly (only on permanent errors, not on transient-ceiling
  give-ups).

gkatz2:
## Summary

- When an OAuth token refresh fails as a transient error for longer
  than the in-loop short retry (~1–2 minutes at defaults), the
  monitor today marks the workload `unauthenticated` and exits its
  goroutine. The workload stays permanently dead even after the
  underlying condition clears on its own. The canonical trigger is a
  brief client-side network-context change—a VPN disconnect, a
  laptop sleeping on one network and resuming on another, etc.—that
  causes token-refresh requests to reach the OAuth server from a
  different network origin. See #5349 for the full failure mode.
- Add a `WorkloadStatusAuthRetrying` workload status. After the short
  retry exhausts on a still-transient error, the monitor stays alive
  and re-attempts refresh on a longer cadence (default 10 min,
  configurable via `TOOLHIVE_TOKEN_AUTH_RETRYING_TICK_INTERVAL`) until
  either success (→ `Running`) or a configurable ceiling (default 24
  h, configurable via `TOOLHIVE_TOKEN_AUTH_RETRYING_MAX_ELAPSED`)
  (→ `Unauthenticated`).
- Hot callers (request-path `Token()` calls, e.g. from the token
  injection middleware) fast-fail with the cached error during
  AuthRetrying, returning 503+Retry-After immediately rather than
  blocking on another short-retry duration against the broken
  endpoint.
- New `pkg/oauthproto/oauthtest/server.go` provides a scriptable
  OAuth server fixture used by two end-to-end integration tests that
  drive the state machine against the real `golang.org/x/oauth2`
  library + real HTTP responses.
- The `vmcp` backend health mapping treats `AuthRetrying` as
  `BackendDegraded` so tool discovery still aggregates the workload
  while invocation surfaces the underlying 503.

Fixes #5349

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Unit tests (`task test`)—11 new tests across the PR:
      7 covering the AuthRetrying state machine (entry on short-retry
      exhaustion, recovery to `Running`, ceiling transition to
      `Unauthenticated`, hot-caller fast-fail, permanent-error mid-
      AuthRetrying, DCR Warn silence on ceiling, post-stop gate);
      2 integration tests that drive the state machine against a
      real `httptest.NewServer` through the real `golang.org/x/oauth2`
      library; 2 covering the new switch arms in
      `mapWorkloadStatusToVMCPHealth` and `workloadStatusIndicator`.
- [x] Manual testing—built a local override binary, replaced the
      deployed `~/.local/bin/thv`, restarted two real OAuth-backed
      remote workloads (Sourcegraph and Sourcegraph Deep Search) on
      the new binary. Verified MCP `initialize` round-trip works
      end-to-end with token injection through the proxy. Verified
      `thv list` (default, no flag) and `thv list --all` both render
      the new `auth_retrying` status via a manual status-file
      simulation.

**Note on confidence.** The bug's natural trigger (a real WAF block
during a real token refresh) recurs every 1–3 days in the affected
environment and needs sudo-level network controls (`pfctl` /
`iptables`) to drive on demand—neither fits CI or routine
verification. The two `TestIntegration_*` cases in
`pkg/auth/monitored_token_source_test.go` are where regressions in
the long-tail behavior would surface: they drive the full state
machine through the real `golang.org/x/oauth2` library against a
scriptable OAuth server fixture (`pkg/oauthproto/oauthtest/`)
returning the actual `<!DOCTYPE html>` 4xx-without-RFC-6749-error-
code response shape observed in production. Two scenarios:
`Running → AuthRetrying → Running` (recovery on next tick) and
`Running → AuthRetrying → Unauthenticated` (ceiling exceeded). The
mock-based unit tests in the same file pin individual edge cases at
higher speed; together with the integration tests they cover both
the state machine logic and the real-library response-parsing path.

## API Compatibility

The `v1beta1` operator API surface is not touched by this PR. The
new `auth_retrying` workload status is added to the runtime enum and
to the OpenAPI swagger for the workloads HTTP API, both of which are
additive—existing clients that don't know about the new value will
see it as an unrecognised string and route through their default
handlers.

Yes:

- New `auth_retrying` workload status, surfaced in default
  `thv list` alongside `running` (the workload is still self-recovering
  without user intervention – every other non-Running status is terminal
  until external intervention, and stays hidden behind `--all`).
  Also in the OpenAPI swagger and the architecture state diagram.
- Operators can tune the AuthRetrying cadence (default `10m`) and
  ceiling (default `24h`) via `TOOLHIVE_TOKEN_AUTH_RETRYING_TICK_INTERVAL`
  and `TOOLHIVE_TOKEN_AUTH_RETRYING_MAX_ELAPSED`.
- In environments where the bug surfaces (described in #5349),
  operators no longer need to manually re-auth workloads after a
  transient network-context change resolves on its own.

## Large PR Justification

The 1825-line additions break down as:

- **Tests**: ~1213 lines across 5 `*_test.go` files. The bulk is in
  `pkg/auth/monitored_token_source_test.go` (1029 new lines), covering
  the new state machine, the concurrent stress test of the gate-under-mu
  invariant (1000 iterations), 2 integration tests that drive the full
  flow against a real `golang.org/x/oauth2` library, and regression
  tests for the bug fixes from review iteration.
- **Test infrastructure**: 136 lines in
  `pkg/oauthproto/oauthtest/server.go`, a new scriptable OAuth server
  fixture used by the integration tests.
- **Generated swagger**: 12 lines, regenerated from the new
  `auth_retrying` enum value in `pkg/api/v1/workload_types.go`.
- **Architecture docs**: 15 lines documenting the new state.
- **Production code**: ~450 lines. The change is necessarily atomic –
  state transitions, ceiling logic, hot-caller fast-fail, and the
  `MonitoredTokenSource` plumbing reference each other and would not
  function if split.

Generated with [Claude Code](https://claude.com/claude-code).


gkatz2:
Gentle nudge — I'm still seeing this issue across multiple OAuth-backed MCP servers in our environment. Happy to address review feedback whenever a maintainer can take a look. Thanks!

jhrozek (commented):
Test comment

docs/arch/02-core-concepts.md
```diff
@@ -38,6 +38,7 @@ A **workload** is the fundamental deployment unit in ToolHive. It represents eve
 - `error` - Workload encountered an error
 - `unhealthy` - Workload is running but unhealthy
 - `unauthenticated` - Remote workload cannot authenticate (expired tokens)
+- `auth_retrying` - Remote workload's token refresh is failing transiently; monitor is still retrying until success (→ `running`) or the configured ceiling (→ `unauthenticated`)
 - `unknown` - Workload status cannot be determined
```
jhrozek, line 42: Test

jhrozek (commented):
test

pkg/auth/monitored_token_source.go
```diff
@@ -368,22 +478,90 @@ func (mts *MonitoredTokenSource) resetTimer(d time.Duration) {
...
+	// Call the raw token source directly — NOT Token() — so the monitor
+	// bypasses the AuthRetrying fast-fail and the singleflight retry. The
+	// monitor owns those state transitions itself.
+	tok, err := mts.tokenSource.Token()
+
+	if err == nil {
+		mts.exitAuthRetrying()
+		if tok == nil || tok.Expiry.IsZero() {
+			return true, 0
+		}
```
jhrozek, line 523: test

pkg/auth/monitored_token_source.go
```diff
@@ -368,22 +478,90 @@ func (mts *MonitoredTokenSource) resetTimer(d time.Duration) {
...
+
+	// First transient failure on this tick: run the existing short retry once
+	// so we benefit from singleflight + exponential backoff for brief blips.
+	tok, err = mts.refresher.Refresh(mts.monitoringCtx, err)
+	if err == nil {
+		if tok == nil || tok.Expiry.IsZero() {
+			return true, 0
+		}
+		return false, waitUntilExpiry(tok.Expiry)
+	}
```
jhrozek, line 564: test

jhrozek (commented):
test

pkg/auth/monitored_token_source.go
```diff
@@ -309,19 +400,38 @@ func (mts *MonitoredTokenSource) Token() (*oauth2.Token, error) {
...
+	select {
+	case <-mts.stopMonitoring:
+		return nil, err
+	default:
+	}
+
 	// Transient network error — funnel all concurrent callers through a
 	// single retry loop so we don't hammer the token endpoint.
 	tok, err = mts.refresher.Refresh(mts.monitoringCtx, err)
-	if err != nil {
```
jhrozek, line 315: test

jhrozek (commented):
test

pkg/auth/monitored_token_source.go
```diff
@@ -541,32 +719,169 @@ func isOAuthParseError(err error) bool {
 // of stopMonitoring share stopOnce, so a caller (e.g. a tight Token()
 // loop) cannot spam the record on every call after the workload has
 // already transitioned to Unauthenticated.
+//
+// Ordering matters: stopMonitoring is closed first so any concurrent
+// enterAuthRetrying call sees the gate closed before it acquires the
+// field mutex, eliminating the race where a hot caller could write
+// AuthRetrying to the status file just after we've written
```
jhrozek: test

jhrozek (commented):
test

pkg/workloads/manager.go
```diff
@@ -331,6 +331,14 @@ func mapWorkloadStatusToVMCPHealth(status rt.WorkloadStatus) vmcp.BackendHealthS
...
 		return vmcp.BackendUnauthenticated
+	case rt.WorkloadStatusAuthRetrying:
+		// Token refresh is in transient retry; the workload may yet recover
+		// without operator intervention. Map to BackendDegraded so vmcp
+		// distinguishes this from the terminal Unauthenticated case and
+		// keeps the backend in tool-discovery aggregation — tool invocation
+		// will still fail with 503 until the token refresh recovers, but
+		// discovery callers see the backend's capabilities throughout.
+		return vmcp.BackendDegraded
 	case rt.WorkloadStatusPolicyStopped:
```
jhrozek, line 342: test

jhrozek (commented):
test

pkg/auth/monitored_token_source.go
```diff
@@ -368,22 +478,90 @@ func (mts *MonitoredTokenSource) resetTimer(d time.Duration) {
...
-	if err != nil {
+	// Call the raw token source directly — NOT Token() — so the monitor
+	// bypasses the AuthRetrying fast-fail and the singleflight retry. The
+	// monitor owns those state transitions itself.
+	tok, err := mts.tokenSource.Token()
+
+	if err == nil {
+		mts.exitAuthRetrying()
+		if tok == nil || tok.Expiry.IsZero() {
+			return true, 0
```
jhrozek, line 522: This `return true, 0` exits `monitorLoop` via the `defer close(mts.stopped)` path, but `stopMonitoring` is never closed — that only happens through `markAsUnauthenticated` / `stopOnce.Do`.

After the monitor exits here, the gate in `Token()` stays open. If tokens later start failing, a hot caller falls past the `stopMonitoring` select, runs the full `refresher.Refresh` (up to 5min), then calls `enterAuthRetrying` — which correctly aborts because its own gate checks the channel under the lock. But the workload has now entered AuthRetrying with no monitor alive to drive the ceiling or recovery ticks.

Zero-expiry tokens do appear in practice (some sources omit `expires_in`). Worth closing `stopMonitoring` via `stopOnce.Do` on this exit path, or routing through a small helper that does so without emitting `Unauthenticated`.

jhrozek (commented):
test

pkg/auth/monitored_token_source.go
jhrozek, line 523: This `return true, 0` exits `monitorLoop` via `defer close(mts.stopped)`, but `stopMonitoring` is never closed here — that only happens through `markAsUnauthenticated` / `stopOnce.Do`.

After the monitor exits this way, the gate in `Token()` stays open forever. If tokens later start failing, a hot caller falls past the `stopMonitoring` select, runs the full `refresher.Refresh` (up to 5min), then calls `enterAuthRetrying` — which correctly aborts because its own gate checks the channel under the lock. But the workload has now entered AuthRetrying with no monitor alive to drive the ceiling or recovery ticks.

Zero-expiry tokens do appear in practice (some sources omit `expires_in`). Worth closing `stopMonitoring` via `stopOnce.Do` on this exit path, or routing through a small helper that does so without emitting `Unauthenticated`.

pkg/auth/monitored_token_source.go
jhrozek, line 564: The first success path (raw `tokenSource.Token()` succeeds) calls `exitAuthRetrying()` before returning — but this path (short retry succeeded) doesn't.

Narrow race: a hot caller enters AuthRetrying between the monitor's `inAuthRetrying()` check and `Refresh` returning. The monitor then succeeds here and returns to normal scheduling, but `fastFailIfAuthRetrying` keeps rejecting hot callers until the next tick reaches the `inAuthRetrying` branch.

`exitAuthRetrying` is a no-op when not in AuthRetrying, so adding it before this `return` is safe and defensive.

pkg/auth/monitored_token_source.go
jhrozek, line 315: TOCTOU between the `stopMonitoring` select a few lines above and this `Refresh` call: a hot caller passes the gate, gets preempted, `markAsUnauthenticated` fires on another goroutine and closes the channel, then the hot caller resumes here and runs the full 5-min short-retry against a dead endpoint.

The downstream `enterAuthRetrying` call is safe — its gate checks `stopMonitoring` under the lock — but the singleflight slot gets held for up to 5min unnecessarily.

A second non-blocking `select` on `stopMonitoring` immediately before this line would close the window.

pkg/auth/monitored_token_source.go
jhrozek: The in-memory invariant is well-protected — the concurrent stress test validates this — but "eliminating the race" reads as a stronger guarantee than what's actually provided. The `SetWorkloadStatus(AuthRetrying)` emit happens after `mu` is released, so `markAsUnauthenticated` can still write `Unauthenticated` to disk and then `enterAuthRetrying` overwrites it with `AuthRetrying`. The comment a few lines down in the function body already acknowledges this: "a narrower disk-write inversion is still possible".

Suggested wording: "eliminates the in-memory field race" — keeps both comments consistent.

pkg/workloads/manager.go
jhrozek, line 342: Intentional design question: should `auth_retrying` workloads appear in `thv list` without `--all`?

The PR adds a retrying indicator and a styled pill for this status, but both are unreachable — `pkg/workloads/statuses/file_status.go` line 310 filters to `WorkloadStatusRunning` only before the display code is reached. `WorkloadStatusUnauthenticated` and `WorkloadStatusUnhealthy` are also filtered there, so this may be deliberate, but the new visual work suggests the intent was to surface this state prominently by default. `pkg/workloads/statuses/status.go` (via `IsRunning()`) and this listing path have similar guards.

docs/arch/02-core-concepts.md
jhrozek, line 42: There is a third exit path missing from this description: if a monitor tick during the `auth_retrying` window observes a permanent OAuth error (e.g. `invalid_grant`), `markAsUnauthenticated` fires immediately without waiting for the ceiling. The `08-workloads-lifecycle.md` state diagram already captures this correctly as "Ceiling Exceeded or Permanent Error".

```suggestion
- `auth_retrying` - Remote workload's token refresh is failing transiently; monitor is still retrying until success (-> `running`), a permanent error is observed (-> `unauthenticated`), or the configured ceiling is exceeded (-> `unauthenticated`)
```

jhrozek (commented):
Overall the state-machine design is solid and the test suite is thorough — especially the concurrent stress test and the real-HTTP integration tests. A few things worth discussing below.

jhrozek:
sorry about the stray test messages, I was driving the review from a CC session and it couldn't figure out how to post an inline comment :-/

pkg/auth/monitored_token_source.go
gkatz2: Good catch – fixed. The zero-expiry exit now closes `stopMonitoring` via a small `stopMonitoringOnce()` helper without emitting any status transition. New test `TestMonitor_ZeroExpiryTokenClosesStopMonitoring` pins it.

pkg/auth/monitored_token_source.go
gkatz2: Fixed – `exitAuthRetrying()` now runs on the short-retry success path too. New test `TestMonitor_AuthRetryingClearedOnRefreshRecovery` covers it.

pkg/auth/monitored_token_source.go
gkatz2: Looking at this again carefully – between the existing gate at line 415 and the `Refresh` call at line 423 there's no executable code, just comment lines. A second `select` there would sit back-to-back with the existing one. The race window that actually matters is while `Refresh` is running (potentially for several minutes through the short-retry loop), where `markAsUnauthenticated` can close the channel at any point – which a select before `Refresh` wouldn't catch.

Catching that mid-`Refresh` close would need cancellation threaded into the context `Refresh` uses, which felt out of scope here. Acceptable? Happy to be talked into making a change if I'm missing something.

pkg/auth/monitored_token_source.go
gkatz2: Reworded as suggested.

pkg/workloads/manager.go
gkatz2: Good catch. This was an oversight on my part.

The rationale: auth_retrying is the only status that's self-recovering without user intervention. Every other non-Running status is terminal until external action. Surfacing it in default `thv list` means a user whose tool call returned 503+Retry-After can see the 🔄 indicator instead of a missing workload.

Updated the filter in both `pkg/workloads/statuses/file_status.go` and `pkg/workloads/manager.go:getRemoteWorkloadsFromState` — the latter being the path auth_retrying workloads actually take, since they're OAuth-backed remote MCP servers. Help text on `--all` reworded accordingly. New test `TestGetRemoteWorkloadsFromState_AuthRetryingVisibleWithoutAll` and a companion entry in `file_status_test.go` cover both filters.

Thanks for the careful read.

docs/arch/02-core-concepts.md
gkatz2: Applied verbatim.

jhrozek (approved):

PR #5338, opened 2026-05-19
Add default OBO handler hooks and vMCP/proxy converter stubs

tgrunnagle:
## Summary

This is the second of four flat technical tasks for the OSS on-behalf-of (OBO) work. It stands up the default stub infrastructure for a new `obo` external auth type at the three dispatch layers (operator, vMCP converter, proxy runtime), plus the Go constant the other layers reference. After this change, an out-of-tree build can register a real OBO implementation by calling three setter functions; an upstream-only build hits the defaults and surfaces a sentinel `obo.ErrEnterpriseRequired` error at every dispatch point.

The CRD enum is **not** widened in this PR — `spec.type: obo` is still rejected at the apiserver layer. The constant, handler hooks, middleware-factory entry, and converter stub live as dark infrastructure that activates the moment the CRD admission task (#5329) merges.

- **Why**: enable out-of-tree builds to register a real OBO implementation without forking, while keeping a clean, function-pointer override hook surface in upstream. Defining the hooks now lets the dispatch-wiring task (#5328) land in isolation against a stable, reviewed surface.
- **What**:
  - Adds `ExternalAuthTypeOBO` constant in `mcpexternalauthconfig_types.go` and minimal no-op `exhaustive` linter arms.
  - Adds the new `pkg/auth/obo/` package — leaf home for `ErrEnterpriseRequired`, `MiddlewareType`, the default `CreateMiddleware` redirector (503 stub), `DefaultFactory`, and `RegisterFactory`. Wires one new entry into `pkg/runner/middleware.go`.
  - Adds the `OBOHandler` struct, `RegisterOBOHandler` setter (panics on partial registration), and `OBOValidate` / `OBOSecretEnvVars` / `OBOApplyRunConfig` wrappers in `controllerutil/tokenexchange.go`. Reads/writes guarded by a `sync.RWMutex`.
  - Adds `OBOConverter` in `pkg/vmcp/auth/converters/obo.go` and registers it as a built-in in `NewRegistry()`.
  - Adds `StrategyTypeOBO` constant in `pkg/vmcp/auth/types/types.go`.
  - Unit tests cover every default-stub behavior, the last-write-wins setter semantics, and the panic-on-partial-registration contract.

Closes #5327

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`)
- [x] Linting (`task lint-fix`)
[...]

PR #5332, opened 2026-05-19
Factor thv-operator main into app.Run; add proxyrunner Run helper

tgrunnagle:
## Summary

Out-of-tree builds need to wrap `thv-operator` and `thv-proxyrunner` with a
literal two-line `func main() { app.Run() }` so that side-effecting
integration can arrive through blank-import `init()` registration without
the wrapper re-implementing logger setup, signal handling, scheme
registration, or manager wire-up.

- Move all 446 lines of pre-`Run` setup from `cmd/thv-operator/main.go`
  (scheme registration, flag parsing, controller-runtime / `slog` logger
  bridge, leader election + cache options, controller and webhook setup,
  health/ready checks, telemetry runnable, `mgr.Start`) into a new
  `cmd/thv-operator/app` package exposing a single `Run()` entry point.
- Add an additive `Run()` helper to the existing `cmd/thv-proxyrunner/app`
  package that absorbs the 46-line `func main()` body
  (`viper.BindEnv("TOOLHIVE_DEBUG")`, `logging.New` + `slog.SetDefault`,
  signal-aware context honoring `SIGINT`/`SIGTERM`/`SIGQUIT`, and
  `NewRootCmd().ExecuteContext`). The existing `NewRootCmd` export is
  preserved unchanged.
- Shrink both `cmd/thv-operator/main.go` and `cmd/thv-proxyrunner/main.go`
  to two-line wrappers that call `app.Run()`.

Closes #5326

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring (no behavior change)
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Unit tests (`task test`)
- [x] Linting (`task lint-fix`)
- [x] Manual testing (describe below)

Manual: built both binaries and confirmed startup logs, flag handling
(`--metrics-bind-address`, `--health-probe-bind-address`, `--leader-elect`),
`TOOLHIVE_DEBUG` env handling, and `SIGINT`/`SIGTERM`/`SIGQUIT` graceful
shutdown are identical to pre-refactor behavior. No test files were
modified.

## API Compatibility

- [x] This PR does not break the `v1beta1` API, OR the `api-break-allowed` label is applied and the migration guidance is described above.

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/main.go` | Reduced to two-line wrapper that calls `app.Run()`. |
[...]

PR #5354, opened 2026-05-20
Mirror MCPExternalAuthConfig Valid=False onto consumer CR conditions

### LINKED ISSUE #5347: Mirror MCPExternalAuthConfig.Valid=False/EnterpriseRequired to consumer CRs
## Background

#5328 (PR #5345) wired the OBO external-auth type through the operator dispatch sites and the `MCPExternalAuthConfig` reconciler, surfacing `Valid=False` / `Reason=EnterpriseRequired` on the `MCPExternalAuthConfig` itself in upstream-only builds.

The original AC for #5328 also called for the **consumer CRDs** (`MCPServer`, `MCPRemoteProxy`, `VirtualMCPServer`) to surface a parallel `Valid=False` / `EnterpriseRequired` condition. That parity does not exist in the codebase today — a grep of `cmd/thv-operator/controllers/` confirms only one site sets the `ConditionTypeExternalAuthConfigValidated` condition (for the multi-upstream rejection), not as a general propagation from `MCPExternalAuthConfig.Status`. PR #5345's tests verify the dispatch error path, not the parallel-condition propagation.

Per `.claude/rules/operator.md`'s Status Condition Parity rule (\"a gap means one type silently accepts invalid config that the other rejects\"), this remains a real parity gap. It was deferred from #5345 as out-of-scope cross-cutting work.

## Why this matters now

#5329 (CRD enum admission) is what makes `obo` reachable in production. Before #5329 lands, an OBO-typed `MCPExternalAuthConfig` referenced from a consumer CR fails the consumer's reconcile silently (the dispatch error propagates up the runconfig-build error path but no `Valid=False`/`EnterpriseRequired` condition surfaces on the consumer's own `Status.Conditions`).

## Acceptance criteria

- [ ] When an `MCPServer` references an `MCPExternalAuthConfig` whose `Valid` condition is `False`/`EnterpriseRequired`, the `MCPServer.Status.Conditions` surfaces a parallel `Valid=False`/`Reason=EnterpriseRequired` condition (or equivalent — confirm the reason taxonomy with the operator-rules owner).
- [ ] Same for `MCPRemoteProxy`.
- [ ] Same for `VirtualMCPServer`.
- [ ] Integration tests under `cmd/thv-operator/test-integration/mcp-external-auth/` verify the propagation end-to-end against envtest, with CRD-enum bypass per #5328's documented strategy.
- [ ] Must merge before #5329, so production users see the failure on the consumer CR they applied (not buried in the referenced `MCPExternalAuthConfig`).

## References

- PR #5345 — the dispatch-wiring work that exposed this gap (reviewer F1)
- #5328 — the issue PR #5345 closed; AC4 is the parallel-condition criterion
- #5329 — CRD enum admission, downstream of this
- `.claude/rules/operator.md` Status Condition Parity section

tgrunnagle:
## Summary

Closes #5347.

Today, when an `MCPExternalAuthConfig` surfaces `Valid=False` (for example,
an `obo`-typed config that the default OBO handler rejects with
`Reason=EnterpriseRequired` in upstream-only builds), the failure is visible
only on the referenced `MCPExternalAuthConfig` resource. The consumer CR
(`MCPServer`, `MCPRemoteProxy`, `VirtualMCPServer`) reports only a generic
dispatch error, so users must inspect a different object to learn why their
workload will not deploy. This violates the Status Condition Parity rule in
`.claude/rules/operator.md` and was deferred from #5345 as cross-cutting work.

This PR closes that parity gap:

- Adds a shared `external_auth_mirror.go` primitive
  (`mirroredExternalAuthConfigInvalid`, `mirroredInvalidExternalAuthConfigError`,
  `mirroredReasonFromError`) so all three consumer reconcilers probe the source
  the same way and emit the same typed error.
- `MCPServer` mirrors the source's reason and message onto
  `ExternalAuthConfigValidated=False`, and clears any stale mirror once the
  source heals (it has no downstream `True` setter, so without the heal path
  the `False` would persist after the user fixed the spec).
- `MCPRemoteProxy` mirrors the same condition; it has an existing downstream
  `ExternalAuthConfigValidated=True` writer that overwrites any stale `False`
  naturally, so no separate heal branch is needed.
- `VirtualMCPServer` mirrors the source reason onto each per-backend
  `DiscoveredAuthConfig-<name>` / `BackendAuthConfig-<name>` condition via a
  new `AuthConfigError.Reason` field. `setAuthConfigConditions` falls back to
  `ConversionFailed` when no source reason is present, preserving existing
  behavior for non-mirror failures. vMCP heals naturally because
  `setAuthConfigConditions` rebuilds per-backend conditions on every reconcile.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Added or extended unit coverage for every site touched:

- `mcpserver_externalauth_test.go` — table-driven mirror behavior on the
  `MCPServer` reconciler covering `Valid=False/EnterpriseRequired`,
  `Valid=False/InvalidConfig`, `Valid=True` (no mirror), no `Valid` condition
  (no mirror), and the heal-clearing regression guard for a stale mirror left
  by a prior reconcile.
- `mcpremoteproxy_controller_test.go` — adds the `Valid=False` mirror case to
  the existing `TestHandleExternalAuthConfig` table on the `MCPRemoteProxy`
  reconciler.
- `virtualmcpserver_externalauth_test.go` — exercises the discovered and
  inline backend paths, verifying the per-backend condition surfaces the
  mirrored reason and that the `AuthConfigError.Reason` field round-trips
  through `convertBackendAuthConfigToVMCP` -> `buildOutgoingAuthConfig` ->
  `setAuthConfigConditions`.
- `virtualmcpserver_vmcpconfig_test.go` — direct coverage of
  `setAuthConfigConditions` reason-selection: mirrored reason when present,
  `ConversionFailed` fallback otherwise, across the default, discovered, and
  inline backend code paths.

## API Compatibility

- [x] This PR does not break the `v1beta1` API, OR the `api-break-allowed`
  label is applied and the migration guidance is described above.

This PR adds an unexported `Reason` field to the in-process `AuthConfigError`
struct and reads existing `Status.Conditions` on consumer CRs. No CRD schema
changes.

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/controllers/external_auth_mirror.go` | New shared probe, typed error, and reason-extraction helper used by all three consumer reconcilers. |
| `cmd/thv-operator/controllers/mcpserver_controller.go` | Mirror `Valid=False` onto `ExternalAuthConfigValidated`; clear stale mirror when the source heals. |
| `cmd/thv-operator/controllers/mcpremoteproxy_controller.go` | Mirror `Valid=False` onto `ExternalAuthConfigValidated`; downstream `True` setter handles healing. |
| `cmd/thv-operator/controllers/virtualmcpserver_controller.go` | Add `AuthConfigError.Reason`; mirror in discover/convert paths; propagate via `buildOutgoingAuthConfig`; surface in `setAuthConfigConditions`. |
| `cmd/thv-operator/controllers/mcpserver_externalauth_test.go` | Table-driven mirror tests including heal-clearing regression guard. |
| `cmd/thv-operator/controllers/mcpremoteproxy_controller_test.go` | Added `Valid=False` mirror case. |
| `cmd/thv-operator/controllers/virtualmcpserver_externalauth_test.go` | New mirror cases across discovered + inline backend paths. |
| `cmd/thv-operator/controllers/virtualmcpserver_vmcpconfig_test.go` | Direct reason-selection coverage for `setAuthConfigConditions`. |

Yes. When a consumer CR references an `MCPExternalAuthConfig` whose `Valid`
condition is `False`, the consumer's own `Status.Conditions` now carries a
parallel `False` condition with the same reason (for example
`EnterpriseRequired`) and message. Users no longer need to inspect a
different object to diagnose the failure on the workload they applied.

**Stacked PR.** This PR is the next in a stacked series. The upstream PRs in
the stack are #5332, #5338, and #5345 — PR #5345 is the immediate predecessor
and shares branch history. Until #5345 merges to `main`, the diff against
`main` will include those earlier commits. The in-scope commits for this PR
are the three HEAD-most:

1. `Mirror MCPExternalAuthConfig Valid=False onto consumer CR conditions`
2. `Address code review feedback on consumer-CR mirror`
3. `Tighten mirror probe API; trim duplicated error prefix`

Please review after #5345 merges so the diff narrows to just this PR's scope.

**Deferred AC4 (envtest integration tests).** Issue #5347's AC4 calls for
integration tests under `cmd/thv-operator/test-integration/mcp-external-auth/`
that verify propagation end-to-end against envtest with a CRD-enum bypass.
Per in-session direction we deferred this to #5329 to avoid adding a
`setup-envtest` dependency here. Once #5329 admits `obo` at the CRD layer,
the integration test no longer needs a CRD-enum bypass and is substantially
simpler to write. Unit-level coverage in this PR exercises every consumer
reconciler's mirror behavior, the heal path on `MCPServer`, the shared
probe, the reason-propagation helpers, and the per-backend condition wiring
in `setAuthConfigConditions`.

**Known follow-up: `MCPServerEntry`.** `MCPServerEntry` (the per-backend
entry inside `VirtualMCPServer`) has the same parity gap, but it is
out-of-scope for #5347's explicit AC list
(`MCPServer`/`MCPRemoteProxy`/`VirtualMCPServer`). Suggest filing a separate
issue to track the entry-level mirror; not filing from this PR.

**Merge ordering.** This PR must merge before #5329 so production users see
the `EnterpriseRequired` failure on the resource they applied (the consumer
CR) the moment apiserver-level admission of `obo`-typed configs is lit up.

### DIFF
--- a/cmd/thv-operator/controllers/external_auth_mirror.go
+++ b/cmd/thv-operator/controllers/external_auth_mirror.go
@@ -29,11 +29,17 @@ func (e *mirroredInvalidExternalAuthConfigError) Error() string {
 }
 
 // mirroredExternalAuthConfigInvalid inspects a fetched MCPExternalAuthConfig
-// and returns (reason, error) when its Valid condition is False, or ("", nil)
-// when Valid is True/absent. The returned error is a
-// *mirroredInvalidExternalAuthConfigError that carries the source's reason and
-// message so callers can surface them on the consumer's status without
-// re-fetching the object.
+// and returns a non-nil *mirroredInvalidExternalAuthConfigError when its Valid
+// condition is False, or nil when Valid is True/absent. The returned value
+// implements error and carries the source's reason and message so callers can
+// surface them on the consumer's status without re-fetching the object.
+//
+// Returning the concrete pointer (rather than an (error, string) tuple) keeps
+// callers from needing errors.As to recover the reason/message fields when
+// they have the value in hand. The error path still satisfies the standard
+// error interface, so callers that propagate it through layers that only
+// carry error (notably convertBackendAuthConfigToVMCP -> buildOutgoingAuthConfig)
+// can recover the same pointer via mirroredReasonFromError below.
 //
 // Consumer reconcilers (MCPServer, MCPRemoteProxy, VirtualMCPServer) use this
 // probe to mirror the source's Valid=False condition onto the consumer CR,
@@ -43,12 +49,12 @@ func (e *mirroredInvalidExternalAuthConfigError) Error() string {
 // consumer's status to report only the generic dispatch failure.
 func mirroredExternalAuthConfigInvalid(
 	externalAuthConfig *mcpv1beta1.MCPExternalAuthConfig,
-) (string, error) {
+) *mirroredInvalidExternalAuthConfigError {
 	validCond := meta.FindStatusCondition(externalAuthConfig.Status.Conditions, mcpv1beta1.ConditionTypeValid)
 	if validCond == nil || validCond.Status != metav1.ConditionFalse {
-		return "", nil
+		return nil
 	}
-	return validCond.Reason, &mirroredInvalidExternalAuthConfigError{
+	return &mirroredInvalidExternalAuthConfigError{
 		Reason:  validCond.Reason,
 		Message: validCond.Message,
 	}
--- a/cmd/thv-operator/controllers/external_auth_mirror.go
+++ b/cmd/thv-operator/controllers/external_auth_mirror.go
@@ -13,40 +13,37 @@ import (
 	mcpv1beta1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1beta1"
 )
 
+// Status Condition Parity machinery for #5347. The probe reads a referenced
+// MCPExternalAuthConfig's Valid condition; the per-consumer mirror functions
+// transcribe its reason+message onto the consumer CR's parallel condition.
+// Without this, an obo-typed MCPExternalAuthConfig in an upstream-only build
+// surfaces EnterpriseRequired only on the referenced resource and the consumer
+// status reports only the generic dispatch failure.
+
+// fallbackInvalidReason is substituted when a source surfaces Valid=False with
+// an empty Reason. metav1.Condition requires Reason to be non-empty and the
+// apiserver rejects empty-Reason patches, so the mirror cannot copy an empty
+// Reason verbatim without trapping the consumer in a noisy reconcile loop.
+const fallbackInvalidReason = "InvalidExternalAuthConfig"
+
 // mirroredInvalidExternalAuthConfigError signals that a referenced
-// MCPExternalAuthConfig had Status.Conditions[Valid]=False and the source's
-// reason+message should be mirrored onto the consumer's status. Callers may
-// extract the reason via mirroredReasonFromError when they need to propagate
-// it through layers that only carry a generic error (notably the
-// VirtualMCPServer AuthConfigError aggregator).
+// MCPExternalAuthConfig had Status.Conditions[Valid]=False. Carries the
+// source's reason+message so callers can surface them on the consumer's
+// status without re-fetching the object, and satisfies the error interface
+// so it can travel through error-returning APIs (notably
+// convertBackendAuthConfigToVMCP -> buildOutgoingAuthConfig).
 type mirroredInvalidExternalAuthConfigError struct {
 	Reason  string
 	Message string
 }
 
 func (e *mirroredInvalidExternalAuthConfigError) Error() string {
-	return fmt.Sprintf("referenced MCPExternalAuthConfig is invalid (%s): %s", e.Reason, e.Message)
+	return fmt.Sprintf("invalid (%s): %s", e.Reason, e.Message)
 }
 
-// mirroredExternalAuthConfigInvalid inspects a fetched MCPExternalAuthConfig
-// and returns a non-nil *mirroredInvalidExternalAuthConfigError when its Valid
-// condition is False, or nil when Valid is True/absent. The returned value
-// implements error and carries the source's reason and message so callers can
-// surface them on the consumer's status without re-fetching the object.
-//
-// Returning the concrete pointer (rather than an (error, string) tuple) keeps
-// callers from needing errors.As to recover the reason/message fields when
-// they have the value in hand. The error path still satisfies the standard
-// error interface, so callers that propagate it through layers that only
-// carry error (notably convertBackendAuthConfigToVMCP -> buildOutgoingAuthConfig)
-// can recover the same pointer via mirroredReasonFromError below.
-//
-// Consumer reconcilers (MCPServer, MCPRemoteProxy, VirtualMCPServer) use this
-// probe to mirror the source's Valid=False condition onto the consumer CR,
-// closing the Status Condition Parity gap described in #5347. Without the
-// mirror, an obo-typed MCPExternalAuthConfig in an upstream-only build would
-// surface EnterpriseRequired only on the referenced resource, leaving the
-// consumer's status to report only the generic dispatch failure.
+// mirroredExternalAuthConfigInvalid returns a non-nil typed error when the
+// source's Valid condition is False, or nil otherwise. Use
+// [mirroredReasonFromError] to recover the reason from a wrapped chain.
 func mirroredExternalAuthConfigInvalid(
 	externalAuthConfig *mcpv1beta1.MCPExternalAuthConfig,
 ) *mirroredInvalidExternalAuthConfigError {
@@ -54,21 +51,75 @@ func mirroredExternalAuthConfigInvalid(
 	if validCond == nil || validCond.Status != metav1.ConditionFalse {
 		return nil
 	}
-	return &mirroredInvalidExternalAuthConfigError{
-		Reason:  validCond.Reason,
-		Message: validCond.Message,
+	reason := validCond.Reason
+	if reason == "" {
+		reason = fallbackInvalidReason
 	}
+	return &mirroredInvalidExternalAuthConfigError{Reason: reason, Message: validCond.Message}
 }
 
 // mirroredReasonFromError returns the mirrored source reason embedded in err
 // (via *mirroredInvalidExternalAuthConfigError) or "" if err does not carry
-// one. Used by the VirtualMCPServer buildOutgoingAuthConfig pipeline to
-// propagate the source's reason onto the per-backend AuthConfigError without
-// re-fetching the MCPExternalAuthConfig.
+// one. Walks the wrap chain via errors.As so it remains correct when callers
+// wrap the typed error with fmt.Errorf("...: %w", err) before passing it on
+// (notably buildOutgoingAuthConfig in the VirtualMCPServer pipeline).
 func mirroredReasonFromError(err error) string {
 	var mirrored *mirroredInvalidExternalAuthConfigError
 	if stderrors.As(err, &mirrored) {
 		return mirrored.Reason
 	}
 	return ""
+}
+
+// mirrorInvalidOnMCPServer mirrors the source's Valid=False condition onto the
+// MCPServer's ExternalAuthConfigValidated condition. When the source is healed
+// (Valid=True or absent), it clears any stale mirror so the condition does not
+// outlive its cause. Returns (true, err) when a False mirror was written so
+// the caller can mark Phase=Failed; (false, nil) otherwise.
+//
+// See package-level Status Condition Parity comment for the #5347 motivation.
+func mirrorInvalidOnMCPServer(
+	m *mcpv1beta1.MCPServer,
+	externalAuthConfig *mcpv1beta1.MCPExternalAuthConfig,
+) (bool, error) {
+	mirrored := mirroredExternalAuthConfigInvalid(externalAuthConfig)
+	if mirrored == nil {
+		meta.RemoveStatusCondition(&m.Status.Conditions, mcpv1beta1.ConditionTypeExternalAuthConfigValidated)
+		return false, nil
+	}
+	meta.SetStatusCondition(&m.Status.Conditions, metav1.Condition{
+		Type:               mcpv1beta1.ConditionTypeExternalAuthConfigValidated,
+		Status:             metav1.ConditionFalse,
+		Reason:             mirrored.Reason,
+		Message:            mirrored.Message,
+		ObservedGeneration: m.Generation,
+	})
+	return true, fmt.Errorf("MCPExternalAuthConfig %s/%s: %w", m.Namespace, externalAuthConfig.Name, mirrored)
+}
+
+// mirrorInvalidOnRemoteProxy mirrors the source's Valid=False condition onto
+// the MCPRemoteProxy's ExternalAuthConfigValidated condition. When the source
+// is healed, it clears any stale mirror defensively — the downstream True
+// writer in handleExternalAuthConfig also sets the success reason, but a
+// future early return between this site and that writer would otherwise leak
+// a stale False. Returns (true, err) when a False mirror was written so the
+// caller can short-circuit; (false, nil) otherwise.
+func mirrorInvalidOnRemoteProxy(
+	proxy *mcpv1beta1.MCPRemoteProxy,
+	externalAuthConfig *mcpv1beta1.MCPExternalAuthConfig,
+) (bool, error) {
+	mirrored := mirroredExternalAuthConfigInvalid(externalAuthConfig)
+	if mirrored == nil {
+		meta.RemoveStatusCondition(
+			&proxy.Status.Conditions, mcpv1beta1.ConditionTypeMCPRemoteProxyExternalAuthConfigValidated)
+		return false, nil
+	}
+	meta.SetStatusCondition(&proxy.Status.Conditions, metav1.Condition{
+		Type:               mcpv1beta1.ConditionTypeMCPRemoteProxyExternalAuthConfigValidated,
+		Status:             metav1.ConditionFalse,
+		Reason:             mirrored.Reason,
+		Message:            mirrored.Message,
+		ObservedGeneration: proxy.Generation,
+	})
+	return true, fmt.Errorf("MCPExternalAuthConfig %s/%s: %w", proxy.Namespace, externalAuthConfig.Name, mirrored)
 }
--- a/cmd/thv-operator/controllers/external_auth_mirror.go
+++ b/cmd/thv-operator/controllers/external_auth_mirror.go
@@ -1,0 +1,68 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package controllers
+
+import (
+	stderrors "errors"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	mcpv1beta1 "github.com/stacklok/toolhive/cmd/thv-operator/api/v1beta1"
+)
+
+// mirroredInvalidExternalAuthConfigError signals that a referenced
+// MCPExternalAuthConfig had Status.Conditions[Valid]=False and the source's
+// reason+message should be mirrored onto the consumer's status. Callers may
+// extract the reason via mirroredReasonFromError when they need to propagate
+// it through layers that only carry a generic error (notably the
+// VirtualMCPServer AuthConfigError aggregator).
+type mirroredInvalidExternalAuthConfigError struct {
+	Reason  string
+	Message string
+}
+
+func (e *mirroredInvalidExternalAuthConfigError) Error() string {
+	return fmt.Sprintf("referenced MCPExternalAuthConfig is invalid (%s): %s", e.Reason, e.Message)
+}
+
+// mirroredExternalAuthConfigInvalid inspects a fetched MCPExternalAuthConfig
+// and returns (reason, error) when its Valid condition is False, or ("", nil)
+// when Valid is True/absent. The returned error is a
+// *mirroredInvalidExternalAuthConfigError that carries the source's reason and
+// message so callers can surface them on the consumer's status without
+// re-fetching the object.
+//
+// Consumer reconcilers (MCPServer, MCPRemoteProxy, VirtualMCPServer) use this
+// probe to mirror the source's Valid=False condition onto the consumer CR,
+// closing the Status Condition Parity gap described in #5347. Without the
+// mirror, an obo-typed MCPExternalAuthConfig in an upstream-only build would
+// surface EnterpriseRequired only on the referenced resource, leaving the
+// consumer's status to report only the generic dispatch failure.
+func mirroredExternalAuthConfigInvalid(
+	externalAuthConfig *mcpv1beta1.MCPExternalAuthConfig,
+) (string, error) {
+	validCond := meta.FindStatusCondition(externalAuthConfig.Status.Conditions, mcpv1beta1.ConditionTypeValid)
+	if validCond == nil || validCond.Status != metav1.ConditionFalse {
+		return "", nil
+	}
+	return validCond.Reason, &mirroredInvalidExternalAuthConfigError{
+		Reason:  validCond.Reason,
+		Message: validCond.Message,
+	}
+}
+
+// mirroredReasonFromError returns the mirrored source reason embedded in err
+// (via *mirroredInvalidExternalAuthConfigError) or "" if err does not carry
+// one. Used by the VirtualMCPServer buildOutgoingAuthConfig pipeline to
+// propagate the source's reason onto the per-backend AuthConfigError without
+// re-fetching the MCPExternalAuthConfig.
+func mirroredReasonFromError(err error) string {
+	var mirrored *mirroredInvalidExternalAuthConfigError
+	if stderrors.As(err, &mirrored) {
+		return mirrored.Reason
+	}
+	return ""
+}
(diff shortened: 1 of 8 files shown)


cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -2076,6 +2084,41 @@ func (r *MCPServerReconciler) handleExternalAuthConfig(ctx context.Context, m *m
...
+	m *mcpv1beta1.MCPServer,
+	externalAuthConfig *mcpv1beta1.MCPExternalAuthConfig,
+) (bool, error) {
+	mirrored := mirroredExternalAuthConfigInvalid(externalAuthConfig)
+	if mirrored == nil {
+		// Source healed (Valid=True or absent). Clear any stale mirror so it
+		// doesn't persist after the cause is gone. Downstream branches in
+		// handleExternalAuthConfig (e.g. multi-upstream) re-set False if they
+		// still apply on this reconcile.
+		meta.RemoveStatusCondition(&m.Status.Conditions, mcpv1beta1.ConditionTypeExternalAuthConfigValidated)
```
tgrunnagle: **F1 (HIGH)** — The heal-clearing path you add here only fires when the helper is reached. `handleExternalAuthConfig` at line 2018 returns early when `m.Spec.ExternalAuthConfigRef == nil` and never invokes this helper, so the `meta.RemoveStatusCondition` you put here will not fire when a user removes the ref from spec to fix a stale `False/EnterpriseRequired` mirror. Result: the mirrored condition outlives its cause indefinitely.

`MCPRemoteProxy.handleExternalAuthConfig` at `mcpremoteproxy_controller.go:704` already clears the parallel condition in its no-ref branch — the inconsistency between the two reconcilers is exactly the parity drift `.claude/rules/operator.md` §Status Condition Parity is meant to prevent.

```suggestion
		meta.RemoveStatusCondition(&m.Status.Conditions, mcpv1beta1.ConditionTypeExternalAuthConfigValidated)
		return false, nil
	}
```

_(The diff above is unchanged at this site; the real fix is to add the same `RemoveStatusCondition` call inside `handleExternalAuthConfig`'s no-ref early-return branch, plus a regression test in `TestMCPServerReconciler_handleExternalAuthConfig_MirrorsInvalidCondition`.)_

tgrunnagle (commented):
## Multi-agent review summary

Four specialist agents (k8s-conditions, go-style, tests, general-quality) reviewed the diff. Codex cross-review skipped — `codex` CLI not installed on the reviewer host.

### Consensus summary

| F1 | HIGH | `mcpserver_controller.go` | MCPServer leaks mirrored False when `ExternalAuthConfigRef` is removed from spec |
| F2 | MEDIUM | `mcpremoteproxy_controller.go` | MCPRemoteProxy mirror lacks defensive clear-on-heal |
| F3 | MEDIUM | `virtualmcpserver_externalauth_test.go` | Wrapped-error round-trip through `buildOutgoingAuthConfig` not exercised |
| F4 | MEDIUM | `mcpserver_controller.go` | Asymmetric receiver vs. free-function design between parallel mirror helpers |
| F5 | LOW | `mcpserver_controller.go` | Doubled `MCPExternalAuthConfig` prefix in error message |
| F6 | LOW | `external_auth_mirror.go` | Empty source `Reason` copied verbatim onto consumer condition |
| F7 | LOW | `external_auth_mirror.go` | Probe doc comment is verbose; rationale duplicated across file |
| F8 | LOW | `mcpremoteproxy_controller_test.go` | MCPRemoteProxy mirror test does not validate message propagation |
| F9 | LOW | `mcpserver_externalauth_test.go` | `ObservedGeneration` is set but not asserted in tests |
| F10 | LOW | `mcpserver_externalauth_test.go` | `Remove ref` and `source NotFound` heal edges not tested (related to F1) |

### Holistic assessment

**What the PR does well**

- Closes a real Status Condition Parity gap targeted by `.claude/rules/operator.md`.
- Shared probe + typed error in `external_auth_mirror.go` keeps the per-reconciler call sites small and uniform.
- Good unit coverage on the three consumer reconcilers (AC1–AC3), including a regression guard for MCPServer heal-via-source-flip.
- `errors.As` round-trip through `mirroredReasonFromError` is the right idiom for threading the source reason across the vMCP `AuthConfigError` aggregator.
- `ObservedGeneration` is set correctly on every new condition write.

**Risks / trade-offs to weigh**

- MCPServer.handleExternalAuthConfig returns early when the ref is removed *without* clearing the mirrored False, so the condition outlives its cause in that healing scenario. This is the one issue that should be addressed before merge — it directly contradicts the PR's stated heal-path intent.
- MCPRemoteProxy heal correctness depends on a downstream True-writer being reachable on every code path, vs. MCPServer's defensive clear-on-heal. Pick one model and apply it to both.
- Two near-identical helpers have asymmetric shapes — method on `*MCPServerReconciler` with unused receiver vs. free function for `MCPRemoteProxy`.
- The `mirroredReasonFromError` test does not validate the `errors.As` walk through the `fmt.Errorf("... %w", err)` wrap that production code actually applies in `buildOutgoingAuthConfig` — a future `%w` → `%v` regression would not be caught.

### Recommendation

Fix F1 (one-line `meta.RemoveStatusCondition` in the no-ref branch of `MCPServerReconciler.handleExternalAuthConfig` plus a regression test). The MEDIUM findings (F2–F4) are worth addressing in the same iteration; the LOW findings are quality follow-ups the author can accept or defer.

### AC traceability vs. #5347

- AC1 (MCPServer) — addressed, but heal-path-on-ref-removal is incomplete (F1).
- AC2 (MCPRemoteProxy) — addressed; heal contract is fragile (F2).
- AC3 (VirtualMCPServer) — addressed; reason-propagation correctness depends on wrapped-error contract that lacks a test (F3).
- AC4 (envtest integration) — explicitly deferred to #5329 per PR body; out of scope.
- AC5 (merge ordering) — process item; out of code-review scope.

_Generated with [Claude Code](https://claude.com/claude-code)._

cmd/thv-operator/controllers/mcpremoteproxy_controller.go
```diff
@@ -777,6 +785,33 @@ func (r *MCPRemoteProxyReconciler) handleExternalAuthConfig(ctx context.Context,
...
+// downstream, which overwrites any stale False this helper previously wrote.
+// See the equivalent helper on MCPServerReconciler for the propagation
+// rationale.
+func mirrorExternalAuthConfigInvalidForRemoteProxy(
+	proxy *mcpv1beta1.MCPRemoteProxy,
+	externalAuthConfig *mcpv1beta1.MCPExternalAuthConfig,
+) (bool, error) {
+	mirrored := mirroredExternalAuthConfigInvalid(externalAuthConfig)
+	if mirrored == nil {
+		return false, nil
```
tgrunnagle: **F2 (MEDIUM)** — The doc comment above asserts the heal path is handled "because handleExternalAuthConfig sets ExternalAuthConfigValidated=True downstream", but that's only true if execution reaches the final `meta.SetStatusCondition(... ConditionTrue ...)` at line 762. The multi-upstream rejection at line 746, or any future early return added between this helper and the True-write, would leave a stale `False/EnterpriseRequired` on the proxy.

`MCPServerReconciler.mirrorExternalAuthConfigInvalid` handles this defensively with an explicit `meta.RemoveStatusCondition` on the `mirrored == nil` path. Consider doing the same here for symmetry — the downstream True/multi-upstream writes will still set the correct status, but the heal path becomes robust to future control-flow changes.

```suggestion
	if mirrored == nil {
		meta.RemoveStatusCondition(&proxy.Status.Conditions, mcpv1beta1.ConditionTypeMCPRemoteProxyExternalAuthConfigValidated)
		return false, nil
	}
```

cmd/thv-operator/controllers/virtualmcpserver_externalauth_test.go
```diff
@@ -850,6 +850,153 @@ func TestConvertBackendAuthConfigToVMCP(t *testing.T) {
...
+	}
+
+	strategy, err := r.convertBackendAuthConfigToVMCP(context.Background(), "default", &mcpv1beta1.BackendAuthConfig{
+		Type:                  mcpv1beta1.BackendAuthTypeExternalAuthConfigRef,
+		ExternalAuthConfigRef: &mcpv1beta1.ExternalAuthConfigRef{Name: "obo-source"},
+	})
+
+	require.Error(t, err, "must short-circuit when source Valid=False")
+	require.Nil(t, strategy)
+	assert.Equal(t, mcpv1beta1.ConditionReasonEnterpriseRequired, mirroredReasonFromError(err),
```
tgrunnagle, line 900: **F3 (MEDIUM)** — This assertion exercises `mirroredReasonFromError(err)` on the **unwrapped** error returned by `convertBackendAuthConfigToVMCP`. In production, `buildOutgoingAuthConfig` wraps that error with `fmt.Errorf("failed to convert ... auth config: %w", err)` *before* invoking `mirroredReasonFromError`. The test never validates that `errors.As` actually walks the `%w` chain — a future refactor that drops `%w` (e.g. switches to `%v` or `errors.New(fmt.Sprintf(...))`) would silently break reason propagation on every per-backend condition.

Suggest adding a wrapped-round-trip assertion:

```go
wrapped := fmt.Errorf("failed to convert backend auth config: %w", err)
assert.Equal(t, mcpv1beta1.ConditionReasonEnterpriseRequired, mirroredReasonFromError(wrapped),
    "buildOutgoingAuthConfig wraps the err once before extracting the reason — the contract must survive that wrap")
```

Or, better, drive an assertion through `buildOutgoingAuthConfig` end-to-end and inspect the resulting `AuthConfigError.Reason`.

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -2076,6 +2084,41 @@ func (r *MCPServerReconciler) handleExternalAuthConfig(ctx context.Context, m *m
...
+// Valid=True (i.e. has healed since a previous False), the helper clears any
+// stale mirror it previously wrote so the condition doesn't outlive the cause.
+// Returns (true, err) when a mirror was written so callers can short-circuit;
+// (false, nil) otherwise.
+//
+// This propagates failures like Reason=EnterpriseRequired (set by the
+// MCPExternalAuthConfig reconciler for obo-typed configs in upstream-only
+// builds) onto the consumer CR so users see the failure on the resource they
+// applied, not buried in the referenced config's status.
+func (*MCPServerReconciler) mirrorExternalAuthConfigInvalid(
```
tgrunnagle: **F4 (MEDIUM)** — This is declared as a method with an explicitly unused receiver `(*MCPServerReconciler)`, while the parallel proxy helper `mirrorExternalAuthConfigInvalidForRemoteProxy` (`mcpremoteproxy_controller.go:797`) is a free function. Neither helper uses receiver state. The mixed shapes force readers to mentally re-correlate two helpers that do the same job.

Pick one model and apply to both:
- Free functions in `external_auth_mirror.go` with type-disambiguating names (`mirrorInvalidOnMCPServer`, `mirrorInvalidOnRemoteProxy`), or
- Methods on both reconcilers with matching short names (`mirrorExternalAuthConfigInvalid`).

The free-function approach is the simpler change since one of the two is already a free function.

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -2076,6 +2084,41 @@ func (r *MCPServerReconciler) handleExternalAuthConfig(ctx context.Context, m *m
...
+		return false, nil
+	}
+	meta.SetStatusCondition(&m.Status.Conditions, metav1.Condition{
+		Type:               mcpv1beta1.ConditionTypeExternalAuthConfigValidated,
+		Status:             metav1.ConditionFalse,
+		Reason:             mirrored.Reason,
+		Message:            mirrored.Message,
+		ObservedGeneration: m.Generation,
+	})
+	return true, fmt.Errorf("MCPExternalAuthConfig %s/%s: %w", m.Namespace, externalAuthConfig.Name, mirrored)
```
tgrunnagle: **F5 (LOW)** — The PR title claims to "trim duplicated error prefix," but the resulting message still says `MCPExternalAuthConfig` twice: the outer wrap here adds `MCPExternalAuthConfig %s/%s:` and the inner `mirroredInvalidExternalAuthConfigError.Error()` already returns `referenced MCPExternalAuthConfig is invalid (%s): %s`. The namespace in the outer wrap is also redundant since the consumer's namespace is already known to the caller.

Either drop the prefix from the inner `Error()` (return just `"invalid (%s): %s"`) and keep the outer wrap, or drop the outer wrap from both consumer helpers and let `Error()` carry the noun. Same change applies to `mcpremoteproxy_controller.go:812`.

cmd/thv-operator/controllers/external_auth_mirror.go
```diff
@@ -0,0 +1,74 @@
...
+// consumer's status to report only the generic dispatch failure.
+func mirroredExternalAuthConfigInvalid(
+	externalAuthConfig *mcpv1beta1.MCPExternalAuthConfig,
+) *mirroredInvalidExternalAuthConfigError {
+	validCond := meta.FindStatusCondition(externalAuthConfig.Status.Conditions, mcpv1beta1.ConditionTypeValid)
+	if validCond == nil || validCond.Status != metav1.ConditionFalse {
+		return nil
+	}
+	return &mirroredInvalidExternalAuthConfigError{
+		Reason:  validCond.Reason,
```
tgrunnagle: **F6 (LOW)** — `validCond.Reason` is copied verbatim with no minimum-length guard. `metav1.Condition` requires `Reason` non-empty (CRD validation rejects empty Reason on patch), so if a source ever has `Valid=False` with empty `Reason` (corrupt/partial status, or a future bug in the source reconciler), the mirror would copy that empty Reason onto the consumer's condition and trigger an apiserver validation error on the next Status patch — a noisy reconcile loop with an opaque error.

The source reconciler always sets a non-empty Reason today, but defense-in-depth is cheap here: treat `validCond.Reason == ""` as either "absent" (return nil) or substitute a fallback like `"InvalidExternalAuthConfig"`. Add a `TestMirroredExternalAuthConfigInvalid` subtest for the empty-Reason input.

cmd/thv-operator/controllers/external_auth_mirror.go
```diff
@@ -0,0 +1,74 @@
...
+// carry error (notably convertBackendAuthConfigToVMCP -> buildOutgoingAuthConfig)
+// can recover the same pointer via mirroredReasonFromError below.
+//
+// Consumer reconcilers (MCPServer, MCPRemoteProxy, VirtualMCPServer) use this
+// probe to mirror the source's Valid=False condition onto the consumer CR,
+// closing the Status Condition Parity gap described in #5347. Without the
+// mirror, an obo-typed MCPExternalAuthConfig in an upstream-only build would
+// surface EnterpriseRequired only on the referenced resource, leaving the
+// consumer's status to report only the generic dispatch failure.
+func mirroredExternalAuthConfigInvalid(
```
tgrunnagle, line 47: **F7 (LOW)** — The ~14-line doc comment above this function (lines 31–49) re-explains rationale that already appears on the type doc (lines 16–21) and is restated again on each consumer helper's doc. The "why concrete pointer vs. (error, string) tuple" rationale is interesting design context but it's the third place it appears in this file.

Suggest trimming to ~5 lines: contract + a single line on the concrete-pointer choice + a pointer to `mirroredReasonFromError`. Keep one canonical home for the #5347 motivation (likely on the consumer helpers, which is where future readers debugging a mirror failure will land).

cmd/thv-operator/controllers/mcpremoteproxy_controller_test.go
```diff
@@ -671,6 +671,49 @@ func TestHandleExternalAuthConfig(t *testing.T) {
...
+						Reason:  mcpv1beta1.ConditionReasonEnterpriseRequired,
+						Message: "on-behalf-of (OBO) external auth type requires an enterprise build",
+					}},
+				},
+			},
+			expectError:        true,
+			errContains:        "EnterpriseRequired",
+			expectCondition:    true,
+			expectedCondStatus: metav1.ConditionFalse,
+			expectedCondReason: mcpv1beta1.ConditionReasonEnterpriseRequired,
```
tgrunnagle: **F8 (LOW)** — This row asserts `expectedCondStatus` and `expectedCondReason` but not the propagated `Message`. The parallel MCPServer test (`mcpserver_externalauth_test.go:410`) does verify message verbatim — the test-table struct here has no `expectedCondMessage` field. A future refactor that prefixes/truncates the mirrored message would silently break the contract on the proxy side without test failure.

Consider extending the table struct with an optional `expectedCondMessage` and asserting it on this row.

cmd/thv-operator/controllers/mcpserver_externalauth_test.go
```diff
@@ -463,3 +464,152 @@ func TestMCPServerReconciler_handleExternalAuthConfig_NoHashInConfig(t *testing.
...
+
+// TestMCPServerReconciler_handleExternalAuthConfig_MirrorsInvalidCondition verifies
+// the Status Condition Parity mirror added for #5347: when the referenced
+// MCPExternalAuthConfig has Status.Conditions[Valid]=False (e.g. an obo-typed
+// config that the default OBO handler rejected with Reason=EnterpriseRequired
+// in upstream-only builds), the MCPServer reconciler must surface a parallel
+// ExternalAuthConfigValidated=False condition that carries the same reason and
+// message — instead of silently propagating the dispatch error through the
+// generic phase-Failed path.
+func TestMCPServerReconciler_handleExternalAuthConfig_MirrorsInvalidCondition(t *testing.T) {
```
tgrunnagle, line 476: **F9 (LOW)** — The mirror code at `mcpserver_controller.go:2117` correctly sets `ObservedGeneration: m.Generation` on the mirrored condition (and the parallel proxy at `mcpremoteproxy_controller.go:810`). This test verifies `Status`, `Reason`, and `Message` but never asserts `ObservedGeneration`. A regression that drops it (or uses `externalAuthConfig.Generation` by mistake) would not be caught.

Set `mcpServer.Generation` to a non-zero value in the test fixture and add `assert.Equal(t, int64(<n>), cond.ObservedGeneration)`. Same gap applies to the new MCPRemoteProxy mirror row.

cmd/thv-operator/controllers/mcpserver_externalauth_test.go
```diff
@@ -463,3 +464,152 @@ func TestMCPServerReconciler_handleExternalAuthConfig_NoHashInConfig(t *testing.
...
+	t.Cleanup(cancel)
+
+	const (
+		authName  = "obo-config"
+		namespace = "default"
+		reason    = mcpv1beta1.ConditionReasonEnterpriseRequired
+		message   = "on-behalf-of (OBO) external auth type requires an enterprise build"
+	)
+
+	tests := []struct {
```
tgrunnagle, line 489: **F10 (LOW)** — This test covers heal-via-source-flip (the "stale mirror is cleared when source has healed" subtest) but not the two other heal paths the helper's doc comment implies are handled:

1. **User removes `m.Spec.ExternalAuthConfigRef`** while a stale mirror exists — `handleExternalAuthConfig`'s no-ref early-return at line 2018 never reaches the helper. (See F1.)
2. **Source is deleted (NotFound)** between reconciles — `GetExternalAuthConfigForMCPServer` returns the not-found path before reaching the helper.

Once F1 is fixed, please add both subtests as regression guards.

cmd/thv-operator/controllers/mcpserver_controller.go
tgrunnagle: Addressed in e1f0179fd0083ec4a059016347915683fc19147b.

cmd/thv-operator/controllers/mcpremoteproxy_controller.go
tgrunnagle: Addressed in e1f0179fd0083ec4a059016347915683fc19147b.

cmd/thv-operator/controllers/virtualmcpserver_externalauth_test.go
tgrunnagle: Addressed in 01aae22a62d26624db12dba57e3c4c2fdf0a836e.

cmd/thv-operator/controllers/mcpserver_controller.go
tgrunnagle: Addressed in 01aae22a62d26624db12dba57e3c4c2fdf0a836e.

cmd/thv-operator/controllers/mcpserver_controller.go
tgrunnagle: Addressed in 01aae22a62d26624db12dba57e3c4c2fdf0a836e.

cmd/thv-operator/controllers/external_auth_mirror.go
tgrunnagle: Addressed in 01aae22a62d26624db12dba57e3c4c2fdf0a836e.

cmd/thv-operator/controllers/external_auth_mirror.go
tgrunnagle: Addressed in 01aae22a62d26624db12dba57e3c4c2fdf0a836e.

cmd/thv-operator/controllers/mcpremoteproxy_controller_test.go
tgrunnagle: Addressed in 01aae22a62d26624db12dba57e3c4c2fdf0a836e.

cmd/thv-operator/controllers/mcpserver_externalauth_test.go
tgrunnagle: Addressed in 01aae22a62d26624db12dba57e3c4c2fdf0a836e.

cmd/thv-operator/controllers/mcpserver_externalauth_test.go
tgrunnagle: Addressed in e1f0179fd0083ec4a059016347915683fc19147b.

jhrozek (approved):

ChrisJBurns (approved):

PR #5357, opened 2026-05-21
Fix DCR failure for authorization servers with non-root issuer paths

### LINKED ISSUE #5356: v0.28.0 regression: DCR fails for issuers with non-root paths
## Bug description

After upgrading to v0.28.0, `thv run` cannot complete fresh OAuth client registration for remote MCP servers whose authorization server has a non-root issuer path (e.g., issuer `https://example.com/oauth` rather than `https://example.com`). The browser never opens and the workload never enters running state.

The regression fires only on fresh DCR: first-time setup, after `thv rm` + `thv run`, or when cached client credentials are invalidated. Workloads whose runconfigs still have a cached `client_id` are unaffected because DCR is skipped on that path, so users who upgrade without removing existing workloads may not notice the bug until something forces re-registration.

In practice this affects any RFC 8414 §3 compliant authorization server with a non-root issuer path, not just unusual multi-tenant IdPs. [PR #5250](https://github.com/stacklok/toolhive/pull/5250) ("Migrate CLI OAuth flow to pkg/auth/dcr resolver") removed a fallback that the pre-existing discovery utility provided, and its docstring flagged this as a "known limitation of the option (b) migration."

## Steps to reproduce

Use a remote MCP server whose authorization server has a non-root issuer path and serves its metadata at the RFC 8414 §3 host-root path (`https://example.com/.well-known/oauth-authorization-server/oauth`) but not at `https://example.com/oauth/.well-known/openid-configuration`. With no cached OAuth client credentials for the workload:

```
thv rm <server>
thv run <server> [...flags...]
```

## Expected behavior

DCR completes, the OAuth browser flow opens, and the workload enters running state. This was the v0.27.x behavior.

## Actual behavior

`thv run` exits before launching the browser:

```
Error: failed to restart workload <server>: failed to authenticate to remote server: remote authentication failed: dynamic client registration failed: dcr: metadata_discovery: discover authorization server metadata: fetch discovery document from "https://example.com/oauth/.well-known/openid-configuration": unexpected content-type "text/html"
```

The workload does not appear in `thv ls`. The corresponding workload log file contains the structured failure event:

```
{"level":"ERROR","msg":"dcr: resolve failed","step":"metadata_discovery","issuer":"https://example.com/oauth","error":"discover authorization server metadata: fetch discovery document from \"https://example.com/oauth/.well-known/openid-configuration\": unexpected content-type \"text/html\""}
```

The HTTP response from the OIDC issuer-suffix path varies by server (404, redirect, HTML, etc.); the resolver always names `metadata_discovery` as the failing step.

## Environment (if relevant)

- OS/version: macOS Darwin 25.3.0 / arm64
- ToolHive version: v0.28.0 (commit `2ce3ef76ac17d53fa45564b847025d9518a45c34`, official release binary)

## Additional context

The new `resolveDCRCredentials` function at `pkg/auth/discovery/discovery.go:741` builds a single discovery URL:

```go
req.DiscoveryURL = strings.TrimRight(discoveredDoc.Issuer, "/") + "/.well-known/openid-configuration"
```

The pre-existing `discoverOIDCEndpointsWithClientAndValidation` at `pkg/auth/oauth/oidc.go:114-149` tried both well-known variants (the OIDC issuer-suffix path AND the RFC 8414 §3 host-root path) and fell back from the first to the second. The new resolver only tries the first.

The redundant fetch exists because the resolver needs `code_challenge_methods_supported` to evaluate the S256 PKCE gate, and `AuthServerInfo` doesn't carry that field. The [PR #5250](https://github.com/stacklok/toolhive/pull/5250) docstring identifies the architectural follow-up: "Threading `code_challenge_methods_supported` through `AuthServerInfo` (and updating every caller of `ValidateAndDiscoverAuthServer`) would close this loop and is the natural follow-up; it is out of scope for sub-issue 4b."

Workaround for users on v0.28.0: downgrade to v0.27.2.

amirejaz:
## Summary

Authorization servers whose issuer has a non-root path (e.g. `https://example.com/oauth`) may serve their RFC 8414 metadata exclusively at the path-insertion URL (`scheme://host/.well-known/oauth-authorization-server/path`), not at the OIDC issuer-suffix URL (`{issuer}/.well-known/openid-configuration`). After upgrading to v0.28.0, `thv run` would fail for these servers before opening the browser — the workload never entered running state.

- PR #5250 introduced a re-fetch of AS metadata inside `resolveDCRCredentials` to get `code_challenge_methods_supported` for the S256 PKCE gate. It constructed a single OIDC issuer-suffix URL for that fetch and passed it to `FetchAuthorizationServerMetadataFromURL`, which fetches exactly one URL with no fallback. Servers that only serve the RFC 8414 path-insertion URL returned HTML or a 404, producing the `unexpected content-type "text/html"` error from the issue.
- The pre-existing v0.27.x code (`discoverOIDCEndpointsWithClientAndValidation`) tried both URL forms and fell back gracefully; that fallback was lost in the migration.
- Fix routes the re-fetch through `oauthproto.FetchAuthorizationServerMetadata`, which tries all three well-known URL forms in priority order (RFC 8414 path-insertion → OIDC issuer-suffix → bare RFC 8414). The fetched `code_challenge_methods_supported` is forwarded to the resolver via a new `dcr.Request.CodeChallengeMethodsSupported` field so the S256 PKCE gate fires without a second round-trip inside the resolver.

Fixes #5356.

## Type of change

- [x] Bug fix

## Test plan

New and updated tests:

- `TestHandleDynamicRegistration_NonRootIssuerRFC8414PathInsertion` — regression test: mounts metadata only at `/.well-known/oauth-authorization-server/oauth`, verifies DCR succeeds end-to-end (would have failed before the fix).
- `TestHandleDynamicRegistration_PreDiscoveredPathNonRootIssuer` — same scenario on the pre-discovered path (endpoints already in `OAuthFlowConfig`), which is the more common production case.
- `TestHandleDynamicRegistration_SynthesisesEndpointWhenMetadataOmitsIt` — verifies the `ErrRegistrationEndpointMissing` synthesis branch in `resolveDCRCredentials` routes to `{issuer}/register`.
- `TestResolveDCRCredentials_CodeChallengeMethodsSupportedFieldEnablesPublicClient` — pins the new `dcr.Request.CodeChallengeMethodsSupported` field: S256 present allows public client, `plain`-only or absent rejects it, absent is fine for confidential clients.

## API Compatibility

## Changes

| File | Change |
|------|--------|
| `pkg/auth/dcr/request.go` | Add `CodeChallengeMethodsSupported []string` to `Request`; active on the `RegistrationEndpoint` branch so callers can supply already-discovered capability fields. |
| `pkg/auth/dcr/resolver.go` | Thread `req.CodeChallengeMethodsSupported` into `dcrEndpoints` on the `RegistrationEndpoint` branch so the S256 gate fires from the caller-supplied field. |
| `pkg/auth/discovery/discovery.go` | Replace the single OIDC URL construction in `resolveDCRCredentials` with `oauthproto.FetchAuthorizationServerMetadata` (multi-URL fallback); forward `CodeChallengeMethodsSupported` to the resolver. Update function doc. |
| `pkg/auth/discovery/dcr_resolver_test.go` | Add three tests (regression, pre-discovered path, endpoint synthesis). |
| `pkg/auth/dcr/resolver_test.go` | Add table test for `CodeChallengeMethodsSupported` field on `RegistrationEndpoint` branch. |

DCR now succeeds for authorization servers with non-root issuer paths (e.g. `https://example.com/oauth`) that serve RFC 8414 metadata at the path-insertion URL. These servers were broken in v0.28.0; this restores v0.27.x behaviour.

The architectural root cause (#5250 doc comment, "known limitation of the option (b) migration") is that `code_challenge_methods_supported` is not threaded through `AuthServerInfo` to the CLI call site. That follow-up (threading the field through `ValidateAndDiscoverAuthServer` and all its callers) would eliminate the re-fetch entirely on the pre-discovered path and remains open. This PR takes the minimal targeted fix: use the multi-URL fallback for the re-fetch and forward the result via the new field, without touching `AuthServerInfo` or its callers.

### DIFF
--- a/pkg/auth/discovery/discovery.go
+++ b/pkg/auth/discovery/discovery.go
@@ -673,30 +673,22 @@ func handleDynamicRegistration(ctx context.Context, issuer string, config *OAuth
 // to register with token_endpoint_auth_method=none and to refuse when
 // S256 PKCE is not advertised (rather than silently downgrading).
 //
-// Redundant discovery fetch — acknowledged trade-off. The CLI's
-// pre-discovery layer (pkg/auth/remote/handler.go via
-// ValidateAndDiscoverAuthServer) populates OAuthFlowConfig.AuthorizeURL /
-// TokenURL / RegistrationEndpoint but does NOT carry the upstream's
-// code_challenge_methods_supported through to here — that field is the
-// S256 PKCE gate's only input, and AuthServerInfo does not model it. So
-// when the CLI already has endpoints in hand, the resolver MUST still
-// re-fetch the upstream's discovery document to evaluate the gate; the
-// alternative ("skip the gate on the pre-discovered path") would either
-// regress the inherited-S256 acceptance criterion or silently downgrade
-// to a non-public-PKCE registration. Threading
-// code_challenge_methods_supported through AuthServerInfo (and updating
-// every caller of ValidateAndDiscoverAuthServer) would close this loop
-// and is the natural follow-up; it is out of scope for sub-issue 4b.
+// Discovery fetch — multi-URL fallback. oauthproto.FetchAuthorizationServerMetadata
+// tries the three well-known URL forms defined by RFC 8414 §3.1 and OIDC
+// Discovery in priority order: RFC 8414 path-insertion
+// (scheme://host/.well-known/oauth-authorization-server/{path}), OIDC
+// issuer-suffix ({issuer}/.well-known/openid-configuration), and bare
+// RFC 8414 (scheme://host/.well-known/oauth-authorization-server). Using
+// this function rather than constructing a single OIDC URL ensures that
+// authorization servers with non-root issuer paths whose metadata lives
+// at the path-insertion URL are correctly reached (fixes #5356).
+// The fetched code_challenge_methods_supported is forwarded to the
+// resolver via dcr.Request so the S256 PKCE gate fires without a
+// second discovery round-trip inside the resolver.
 //
-// Threat-model correctness in the multi-tenant case: a few upstreams
-// (multi-tenant IdPs) serve their OIDC discovery document at a path
-// the issuer-derived well-known URL would not reach. For those
-// providers the remote handler's pre-discovery already accepted the
-// non-well-known URL via DiscoverActualIssuer; the resolver's second
-// fetch against {issuer}/.well-known/openid-configuration may then 404.
-// The CLI surfaces a clean resolver error in that case rather than
-// silently registering. This is a known limitation of the option (b)
-// migration and is documented for follow-up.
+// Threading code_challenge_methods_supported through AuthServerInfo (and
+// updating every caller of ValidateAndDiscoverAuthServer) would eliminate
+// this fetch on the pre-discovered path and is a natural follow-up.
 func resolveDCRCredentials(
 	ctx context.Context,
 	issuer string,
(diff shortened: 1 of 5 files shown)


pkg/auth/discovery/discovery.go
```diff
@@ -723,22 +715,37 @@ func resolveDCRCredentials(
...
+			u, parseErr := url.Parse(discoveredDoc.Issuer)
+			if parseErr != nil {
+				return nil, fmt.Errorf("dynamic client registration failed: parse issuer for endpoint synthesis: %w", parseErr)
+			}
+			regEndpoint = (&url.URL{
+				Scheme: u.Scheme,
+				Host:   u.Host,
+				Path:   strings.TrimRight(u.Path, "/") + "/register",
+			}).String()
+		}
```
tgrunnagle, line 746: This synthesis logic duplicates `synthesiseRegistrationEndpoint` in `pkg/auth/dcr/resolver.go` (the doc comment already calls this out). Not a blocker for this PR, but worth a follow-on to export the helper from `pkg/auth/dcr` and reuse it here — keeps the two sites from drifting (e.g. if the nanobot/Hydra convention ever changes, or someone adds scheme/host validation to one and not the other).

tgrunnagle (approved):
Fix correctly addresses #5356 — the multi-URL fallback restores v0.27.x behavior and the regression test mounts metadata only at the RFC 8414 path-insertion URL, which is exactly the failure scenario from the issue. Doc comments are updated cleanly and the deferred architectural follow-up (threading capability fields through `AuthServerInfo`) is appropriately scoped out. One non-blocking note inline about synthesis duplication.

PR #5358, opened 2026-05-21
Validate LocalStore paths stay in state directory

### LINKED ISSUE #4736: LocalStore.getFilePath lacks path traversal validation
## Description

`LocalStore.getFilePath` in `pkg/state/local.go` constructs a file path by joining `basePath` with a user-provided `name` via `filepath.Join`, but does not validate that the resolved path stays within `basePath`. A crafted name containing `..` components (e.g. `../../../etc/passwd`) can escape the designated state directory.

The `#nosec G304` comments on every caller claim "filePath is controlled by getFilePath which ensures it's within our designated directory" — but `getFilePath` does not actually enforce this.

## Impact

All `LocalStore` operations are affected: `GetReader`, `GetWriter`, `CreateExclusive`, `Delete`, and `Exists`. A malicious or accidentally crafted state name could read, write, or delete files outside the state directory.

Note: the workload-facing code paths already validate names via `fileutils.ValidateWorkloadNameForPath`, which rejects path separators and `..` components. This issue affects the lower-level `LocalStore` directly, which is also used for RunConfig and group persistence where the validation may not be applied upstream.

## Suggested fix

Add containment validation in `getFilePath` using the `strings.HasPrefix(resolved, basePath + separator)` pattern already established in `pkg/fileutils/contained.go` (`WriteContainedFile`).

I'd like to work on this and will submit a PR.

resolvicomai:
## Summary

- `LocalStore.getFilePath` accepted names that could resolve outside the state directory when they contained path traversal segments.
- Added path containment validation before returning file paths and propagated invalid-name errors through read, write, create, delete, and exists operations.
- Corrected the `#nosec G304` justifications and added regression coverage for traversal attempts across all affected operations.

Fixes #4736

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Unit tests (`go test ./pkg/state`)
- [x] Unit tests (`go test -count=1 ./pkg/state ./pkg/groups ./pkg/workloads/types ./pkg/migration`)
- [x] Whitespace check (`git diff --check`)

## API Compatibility

No.

I also attempted `go test ./...` locally. It reached unrelated environment-bound suites and failed because this machine is missing `/usr/local/kubebuilder/bin/etcd`, a `thv` binary in `PATH`, a kubeconfig, and an available container runtime; one runner test also found port 8001 unavailable.

### DIFF
--- a/pkg/state/local.go
+++ b/pkg/state/local.go
@@ -52,20 +52,30 @@ func NewLocalStore(appName string, storeName string) (*LocalStore, error) {
 	}, nil
 }
 
-// getFilePath returns the full file path for a configuration
-func (s *LocalStore) getFilePath(name string) string {
+// getFilePath returns the full file path for a configuration.
+func (s *LocalStore) getFilePath(name string) (string, error) {
 	// Ensure the name has the correct extension
 	if !strings.HasSuffix(name, FileExtension) {
 		name = name + FileExtension
 	}
-	return filepath.Join(s.basePath, name)
+
+	basePath := filepath.Clean(s.basePath)
+	filePath := filepath.Clean(filepath.Join(basePath, name))
+	if !strings.HasPrefix(filePath, basePath+string(os.PathSeparator)) {
+		return "", fmt.Errorf("state path %q escapes state directory", name)
+	}
+
+	return filePath, nil
 }
 
 // GetReader returns a reader for the state data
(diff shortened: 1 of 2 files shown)


jhrozek:
openclaw slop

PR #5361, opened 2026-05-21
Admit obo in MCPExternalAuthConfig CRD enum

### LINKED ISSUE #5329: Admit obo in MCPExternalAuthConfig CRD enum and regenerate manifests

tgrunnagle:
## Summary

Until this PR, the apiserver enum on `MCPExternalAuthConfig.spec.type` did
not list `obo`, so `kubectl apply` of an `obo`-typed CR was rejected by
admission before any of the previously-landed dispatch wiring could run.
This is the fourth and final flat task under the parent OSS story
(#5325) — the moment the OBO feature becomes user-visible end-to-end on an
upstream-only build.

What changed:

- Append `;obo` to the `+kubebuilder:validation:Enum` marker on
  `MCPExternalAuthConfigSpec.Type` and add the matching `XValidation` CEL
  rule (`self.type == 'obo' ? has(self.obo) : !has(self.obo)`).
- Declare an opaque `OBOConfig struct{}` placeholder and an
  `OBO *OBOConfig` field on `MCPExternalAuthConfigSpec`. The inner schema
  is intentionally empty in this revision; sub-fields land in a follow-up
  RFC.
- Extend the `unauthenticated` CEL rule and the corresponding Go-level
  `validateTypeConfigConsistency` check to also reject `self.obo` /
  `Spec.OBO`, so the "no configuration when unauthenticated" invariant
  holds for the new type.
- Add a no-op `case ExternalAuthTypeOBO` arm to the `Validate()` switch
  and document the two-tier validation pattern: structural validation
  here, semantic validation behind the `controllerutil.OBOValidate`
  function-pointer hook at reconcile time.
- Regenerate the operator-crds chart YAML, the v1beta1 `DeepCopy`
  artifacts, and the public CRD reference docs.
- Update existing OBO-typed test fixtures across operator and vMCP
  packages to populate `Spec.OBO: &OBOConfig{}` so they continue to pass
  the new biconditional check.

Closes #5329

## Type of change

- [x] New feature

## Test plan

Manual testing performed:

- Applied the regenerated CRD YAML to a local Kind cluster and confirmed
  `kubectl explain mcpexternalauthconfig.spec.type` lists `obo`.
- `kubectl apply` of an `obo`-typed `MCPExternalAuthConfig` with
  `spec.obo: {}` succeeds for the first time. The resource transitions
  to `status.conditions[Valid] = False` / `Reason: EnterpriseRequired`
  via the dispatch wiring from the prior task (#5328).
- `kubectl apply` of an `obo`-typed config that omits `spec.obo` is
  rejected at admission with the new CEL message.
- `kubectl apply` of a non-`obo`-typed config that sets `spec.obo: {}`
  is rejected at admission with the new CEL message.
- Existing integration tests in
  `cmd/thv-operator/test-integration/mcp-external-auth/` pass
  unchanged.

## API Compatibility

- [x] This PR does not break the `v1beta1` API. The change is additive:
  one new enum value, one new optional spec field, one new CEL rule, one
  new Go-level validation row. Existing CRs and existing CEL rules are
  unaffected.

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go` | Extend enum, add CEL rule, declare `OBOConfig`, add `OBO` field, add `Validate()` no-op arm, extend `validateTypeConfigConsistency` (incl. `unauthenticated` guard), document two-tier validation and `gocyclo` suppression. |
| `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go` | Add four `Validate()` cases covering the OBO biconditional and the `unauthenticated` guard. |
| `cmd/thv-operator/api/v1beta1/zz_generated.deepcopy.go` | Regenerated `DeepCopyInto` for `OBOConfig` and updated `MCPExternalAuthConfigSpec` copier. |
| `deploy/charts/operator-crds/files/crds/toolhive.stacklok.dev_mcpexternalauthconfigs.yaml` (and `cmd/thv-operator/...` copy) | Regenerated via `task operator-manifests`: new enum value, new CEL rule, new `obo` object property, extended `unauthenticated` rule. |
| `docs/operator/crd-api.md` | Regenerated via `task crdref-gen`. |
| `cmd/thv-operator/controllers/*_test.go`, `cmd/thv-operator/pkg/controllerutil/tokenexchange_test.go`, `pkg/vmcp/auth/converters/obo_test.go` | Populate `Spec.OBO: &OBOConfig{}` on existing OBO-typed fixtures so they satisfy the new biconditional check. |

Yes. `MCPExternalAuthConfig.spec.type` now admits `obo` as a valid enum
value at the apiserver layer. On an upstream-only build, an
`obo`-typed CR is admitted, reconciled, and transitions to
`status.conditions[Valid] = False` with `Reason: EnterpriseRequired`
(via the dispatch wiring landed in #5328). Out-of-tree builds that
register a real OBO handler via `controllerutil.RegisterOBOHandler`
short-circuit the sentinel and run their own protocol-level checks.

The `OBOConfig` struct itself is an intentionally empty placeholder in
this revision; sub-fields land in a follow-up RFC.

- **Stacked branch caveat.** GitHub will show
  this branch's diff against `main` as 7 commits / ~12 files. Only
  `ef00b85c` and `d1c3c817` are in scope here; the other 5 are the
  consumer-CR mirror work under PR for #5347. Easiest review path: wait
  for the mirror PR to merge, then re-render this diff. Alternative:
  scope your review to `cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go`,
  the regenerated YAML/docs, and the small `&OBOConfig{}` test-fixture
  additions.
- **Two-tier validation is the load-bearing design choice.** The
  `Validate()` arm for OBO is deliberately a no-op. Structural validation
  (OBO field set iff type is `obo`) runs in `validateTypeConfigConsistency`
  and mirrors the new CEL rule. Semantic validation (is an OBO handler
  registered for this build?) runs at reconcile time via the
  `controllerutil.OBOValidate` function-pointer hook. Splitting the tiers
  this way is what keeps the upstream CRD schema stable across out-of-tree
  builds — please push back if you'd prefer a different split, but note
  that the prior tasks in this story (#5327, #5328) are already wired to
  this contract.
- **`gocyclo` suppression on `validateTypeConfigConsistency`.** The
  per-type `if` shape mirrors the CEL rules on `MCPExternalAuthConfigSpec`
  one-to-one. A table-driven loop would lower the score but obscure that
  correspondence, which is the property a reviewer relies on to audit the
  structural-validation contract. The suppression is documented in the
  function's doc comment per the "confirmed false positive" bar in
  `.claude/rules/go-style.md`.
- **The `unauthenticated` guard is shape-parity, not load-bearing.** The
  per-type biconditionals always intercept a populated config field
  first. The guard exists so a future contributor adding a new type
  cannot forget to extend the "no configuration must be set" invariant.
  Documented in the function so it doesn't read as dead code.


cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -1176,15 +1204,22 @@ func (r *MCPExternalAuthConfig) validateTypeConfigConsistency() error {
...
 	}
+	if (r.Spec.OBO == nil) == (r.Spec.Type == ExternalAuthTypeOBO) {
+		return fmt.Errorf("obo configuration must be set if and only if type is 'obo'")
+	}
 
-	// Check that unauthenticated has no config
+	// Belt-and-braces guard for `unauthenticated`: shape-parity with the per-type
+	// biconditionals above, which always intercept a populated config field first.
+	// Listed here so a future contributor adding a new type cannot forget to extend
+	// the "no configuration must be set" invariant.
```
tgrunnagle: **F1 · Unauthenticated-guard OBO disjunct is unreachable; comment should say so** _(10/10 consensus, LOW)_

The new `r.Spec.OBO != nil` disjunct below cannot fire today: for `Type=Unauthenticated` with any non-nil config field, the per-type biconditional row above (e.g. line 1207 for OBO, line 1204 for `UpstreamInject`, etc.) evaluates to `false == false = true` and returns its own error first. Test #4 (line 353) confirms this — it asserts the OBO biconditional message, not the unauthenticated one. The current comment leaves ambiguous whether the guard catches a live case.

Make the redundancy explicit so a future contributor reading the diff understands the guard is forward-compat-only:

```suggestion
	// Redundant with the per-type biconditionals above — each fires first for
	// Type=Unauthenticated with any non-nil field — but retained as a single
	// readable invariant so a contributor adding a new ExternalAuthType extends
	// the "no configuration must be set" check here too.
```

_Raised by: api-design specialist + general specialist (independent)._

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -1132,14 +1149,18 @@ func (r *MCPExternalAuthConfig) Validate() error {
 		// No complex validation needed for these types
 		return nil
 	case ExternalAuthTypeOBO:
```
tgrunnagle, line 1177: **F2 · Stale `lines 44-49` reference in `Validate()` doc comment** _(10/10 consensus, LOW)_

This comment is attached to the nearest diff line; the actual issue is at **line 1125** (the `Validate()` doc comment) which sits ~26 lines above this hunk. That doc comment reads:

> _"Note: These validations provide defense-in-depth alongside CEL validation rules (lines 44-49)."_

The CEL rules currently live at lines 58-65 (not 44-49), and they shift further down each time a new rule is added. The reference has already drifted and is on a path to drift again. Per `.claude/rules/go-style.md` ("Keep Comments Synchronized With Code"), and since you're already editing the surrounding `Validate()` body, this is a natural drive-by fix.

Replace `(lines 44-49)` with a structural reference, e.g. `(the +kubebuilder:validation:XValidation markers on MCPExternalAuthConfigSpec)`.

_Raised by: api-design specialist._

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -99,8 +100,24 @@ type MCPExternalAuthConfigSpec struct {
...
+	// +optional
+	OBO *OBOConfig `json:"obo,omitempty"`
 }
 
+// OBOConfig is a placeholder for On-Behalf-Of (OBO) external auth configuration.
+// The inner schema is intentionally empty in this revision; sub-fields land in a
+// follow-up. The struct exists today so that the CRD schema admits `spec.obo: {}`
+// (matching the CEL rule "obo field must be set iff type is obo") and so that
+// downstream tools that introspect the API surface can see the placeholder
+// before the protocol-level fields land.
```
tgrunnagle: **F3 · `OBOConfig` doc justifies the struct with a speculative downstream consumer** _(10/10 consensus, LOW)_

The phrase _"downstream tools that introspect the API surface"_ invokes a consumer that isn't named in the diff or issue. Per `.claude/rules/go-style.md` ("Document Architectural Constraints on Exported Functions"), doc comments should describe what the type does and doesn't do in concrete, verifiable terms.

The real justification is verifiable: the type must exist so `OBO *OBOConfig` compiles and so the CRD schema admits `spec.obo: {}` (because `has(self.obo)` is the biconditional's RHS). The empty-struct shape also keeps the follow-up RFC additive.

```suggestion
// OBOConfig is a placeholder for On-Behalf-Of (OBO) external auth configuration.
// The inner schema is intentionally empty in this revision; sub-fields land in a
// follow-up RFC. The struct exists so OBO *OBOConfig compiles and the CRD
// schema admits `spec.obo: {}` — the CEL rule "obo field must be set iff type
// is obo" requires has(self.obo), which evaluates true for an empty object.
// Stored objects with `obo: {}` will round-trip cleanly when sub-fields land,
// because Go zero values fill in.
```

_Raised by: api-design specialist._

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -1132,14 +1149,18 @@ func (r *MCPExternalAuthConfig) Validate() error {
...
 	case ExternalAuthTypeOBO:
-		// OBO validation is delegated to the registered OBO handler at the
-		// controllerutil layer (via controllerutil.OBOValidate ->
-		// obo.ErrEnterpriseRequired in upstream-only builds), invoked from
-		// the reconcile loop. The CRD-level Validate() stays a no-op for OBO
-		// and exists only to keep the `exhaustive` linter happy now that
-		// ExternalAuthTypeOBO is defined. The CRD enum currently rejects
-		// "obo" at the apiserver layer, so this arm is unreachable in
-		// upstream-only builds.
+		// OBO uses a two-tier validation pattern. Structural validation
```
tgrunnagle: **F4 · OBO `Validate()` arm comment doesn't note structural validation already ran** _(10/10 consensus, LOW)_

The rewritten arm comment accurately describes the two tiers but doesn't tell the reader that `r.validateTypeConfigConsistency()` is invoked unconditionally at line 1130 (top of `Validate()`), so by the time control reaches this arm the structural invariant already holds. A reader landing on this arm in isolation may wonder whether the arm should be running the structural check itself.

Consider adding one sentence near the top of the comment block:

> "Structural validation has already run via `r.validateTypeConfigConsistency()` at the top of this method, so this arm is reached only when the structural invariant holds."

_Raised by: api-design specialist._

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
```diff
@@ -304,6 +304,66 @@ func TestMCPExternalAuthConfig_Validate(t *testing.T) {
...
+					Type:          ExternalAuthTypeTokenExchange,
+					TokenExchange: &TokenExchangeConfig{TokenURL: "https://example.com/token"},
+					OBO:           &OBOConfig{},
+				},
+			},
+			expectErr: true,
+			errMsg:    "obo configuration must be set if and only if type is 'obo'",
+		},
+		{
+			name: "invalid obo config set on unauthenticated type",
```
tgrunnagle: **F5 · Test #4 name implies coverage of the unauthenticated guard; the OBO biconditional intercepts first** _(10/10 consensus, LOW)_

Resolve together with F1. This case sets `Type=Unauthenticated, OBO=&OBOConfig{}` and asserts the OBO biconditional message — which is exactly what fires, because `validateTypeConfigConsistency` iterates top-to-bottom and the OBO row (line 1207 in the source) intercepts before the unauthenticated guard. So the new `r.Spec.OBO != nil` disjunct in the unauthenticated guard is NOT exercised by any test. A future contributor who deletes it will see all tests still pass.

Rename to reflect what the assertion actually proves:

```suggestion
			name: "invalid obo config on unauthenticated type (obo biconditional intercepts first)",
```

Optionally add a one-line code comment under `expectErr` noting that case is also intentional shape-parity coverage for the unauthenticated guard's `OBO != nil` disjunct.

_Raised by: test-coverage specialist._

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
```diff
@@ -304,6 +304,66 @@ func TestMCPExternalAuthConfig_Validate(t *testing.T) {
 			expectErr: true,
 			errMsg:    "oidcConfig must be set when type is 'oidc'",
 		},
+		{
```
tgrunnagle: **F6 · New OBO test cases appended at end of table instead of grouped with per-type cases** _(10/10 consensus, LOW)_

The existing table groups cases by `ExternalAuthType` (tokenExchange, headerInjection, ..., upstreamInject, then OIDC/OAuth2 provider cases). The four new OBO cases sit after the OIDC group at the very end, breaking per-type grouping.

Move them to immediately after the `upstreamInject` group (around line 277 in the file) so the biconditional matrix reads in declaration order. Non-blocking readability nit — feel free to skip if a future reader's `git blame` would obscure who added which case.

_Raised by: test-coverage specialist._

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -1149,13 +1170,20 @@ func (r *MCPExternalAuthConfig) Validate() error {
...
+// the rows are syntactically uniform. Collapsing the rows into a
+// table-driven loop would reduce the score but obscure the one-to-one
+// correspondence with the CEL rules on MCPExternalAuthConfigSpec, which is
+// the property reviewers rely on to audit the structural-validation
+// contract. See issue #5329 for the broader discussion.
 //
-//	(r.Spec.OBO == nil) == (r.Spec.Type == ExternalAuthTypeOBO)
-//
-// and update the unauthenticated check below to also assert !has(self.obo).
+//nolint:gocyclo // one if per ExternalAuthType parallels the CEL rules on the spec; see doc comment
```
tgrunnagle: **F7 · `//nolint:gocyclo` rationale and directive are separated by ~5 lines of doc** _(10/10 consensus, LOW · optional polish)_

The rationale lives in the function doc (correct per `.claude/rules/go-style.md`, which requires `//nolint` to be a documented confirmed false positive). The trailing comment _"see doc comment"_ already points at it, so this is fine as-is. Two optional improvements:

1. Copy a one-line summary onto the directive itself so a reviewer skimming the diff doesn't have to scroll up — e.g. `//nolint:gocyclo // one if per ExternalAuthType mirrors CEL rules; collapsing would obscure parity — see doc comment`.
2. The rationale could state the actual gocyclo score / threshold so a future maintainer knows whether removing the unreachable unauthenticated disjunct (F1) would drop the function under the limit and let the suppression be deleted.

Neither is blocking — the current suppression is already well-justified.

_Raised by: general + api-design specialists._

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
```diff
@@ -61,12 +61,13 @@ type ExternalAuthType string
...
-// +kubebuilder:validation:XValidation:rule="self.type == 'unauthenticated' ? (!has(self.tokenExchange) && !has(self.headerInjection) && !has(self.bearerToken) && !has(self.embeddedAuthServer) && !has(self.awsSts) && !has(self.upstreamInject)) : true",message="no configuration must be set when type is 'unauthenticated'"
+// +kubebuilder:validation:XValidation:rule="self.type == 'obo' ? has(self.obo) : !has(self.obo)",message="obo configuration must be set if and only if type is 'obo'"
+// +kubebuilder:validation:XValidation:rule="self.type == 'unauthenticated' ? (!has(self.tokenExchange) && !has(self.headerInjection) && !has(self.bearerToken) && !has(self.embeddedAuthServer) && !has(self.awsSts) && !has(self.upstreamInject) && !has(self.obo)) : true",message="no configuration must be set when type is 'unauthenticated'"
 //
 //nolint:lll // CEL validation rules exceed line length limit
 type MCPExternalAuthConfigSpec struct {
 	// Type is the type of external authentication to configure
```
tgrunnagle: **F8 · `kubectl explain spec.type` won't surface the build-requires-handler note for `obo`** _(10/10 consensus, LOW · optional)_

Issue \#5329's acceptance criterion: `kubectl explain mcpexternalauthconfig.spec.type` should list `obo` _"with a description noting it requires a build with a registered OBO handler"_. The build-requires-handler note in this PR lives on the `OBO` field doc (lines 104-108), but `kubectl explain spec.type` surfaces the `Type` field doc (this line), which only says _"Type is the type of external authentication to configure"_.

Optional: extend the `Type` field doc with an obo-specific note, e.g.:

> "When `obo`, the cluster must run a build with a registered OBO handler — upstream-only builds surface `status.conditions[Valid] = False` with `Reason: EnterpriseRequired`."

If you prefer to leave the note on the `OBO` field only, consider calling that out in the PR description so the acceptance-criteria reviewer doesn't trip over the letter of the bullet.

_Raised by: general specialist._

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
```diff
@@ -304,6 +304,66 @@ func TestMCPExternalAuthConfig_Validate(t *testing.T) {
...
+			name: "valid obo type with empty config",
```
tgrunnagle: **F9 · "Empty config" wording reads as zero-valued rather than placeholder** _(10/10 consensus, INFO · optional polish)_

The phrase _"with empty config"_ could be read as "the OBO sub-fields are at zero values". Since `OBOConfig` is intentionally a placeholder with no sub-fields, a future contributor adding fields could read the name as inviting a `"populated config"` companion case for symmetry.

Optional rename:

```suggestion
			name: "valid obo type with placeholder OBOConfig",
```

_Raised by: test-coverage specialist._

tgrunnagle (commented):
**Agents consulted**: pr-reviewer (CRD/CEL specialist), pr-reviewer (Go API/validation), pr-reviewer (test coverage), pr-reviewer (generated artifacts), pr-reviewer (general quality + intent vs issue)
**Codex cross-review**: skipped (codex CLI not installed)

### Consensus Summary

| F1 | Unauthenticated-guard OBO disjunct is unreachable; comment should say so | 10/10 | LOW | Fix |
| F2 | Stale `lines 44-49` reference in `Validate()` doc comment | 10/10 | LOW | Fix |
| F3 | `OBOConfig` doc justifies the struct with a speculative downstream consumer | 10/10 | LOW | Fix |
| F4 | OBO `Validate()` arm comment doesn't note structural validation already ran | 10/10 | LOW | Fix |
| F5 | Test #4 name implies coverage of unauthenticated guard; OBO biconditional intercepts first | 10/10 | LOW | Fix |
| F6 | New OBO test cases appended at end of table instead of grouped per-type | 10/10 | LOW | Optional |
| F7 | `//nolint:gocyclo` rationale and directive are separated by ~5 lines of doc | 10/10 | LOW | Optional |
| F8 | `kubectl explain spec.type` won't surface build-requires-handler note | 10/10 | LOW | Optional |
| F9 | "Empty config" wording in test #1 reads as zero-valued rather than placeholder | 10/10 | INFO | Optional |
| F10 | Stale `type: obo` fixtures elsewhere | dropped (verified vacuous) | INFO | None — `grep` returned zero matches |

### Overall

This is a textbook execution of a stacked task — surgical edit to one Go source file plus complete regeneration of the generated artifacts and a thorough fixture sweep. Every acceptance criterion from \#5329 is met: enum extended, `OBOConfig` placeholder struct declared, `OBO *OBOConfig` field added, new CEL biconditional rule paired with the matching Go-level row in `validateTypeConfigConsistency`, `unauthenticated` CEL rule and Go guard extended, generated YAML / deepcopy / docs all updated consistently across the four insertion points. The eight-type symmetry is preserved row-for-row across kubebuilder markers, CEL rules, Go bicond, error messages, and deepcopy.

There are no HIGH-severity findings and no specialist disputes. All ten findings are LOW or INFO and focus on making implicit invariants explicit — primarily around the new unauthenticated-guard disjunct (F1/F5: it's unreachable today and the test that appears to cover it actually exercises the per-type biconditional that intercepts first) and a few comment-accuracy items (F2/F3/F4) that the project's `.claude/rules/go-style.md` "Keep Comments Synchronized With Code" rule highlights.

The most impactful change to bundle is **F1 + F5 together**: clarify the unauthenticated-guard comment so it explicitly says the disjunct is shape-parity-for-future-types (not catching a live case today), and rename test #4 to reflect that the OBO biconditional row is what fires. The remaining items are nice-to-have polish.

Event type is `COMMENT` because (a) the authenticated `gh` user matches the PR author — GitHub would 422 on `APPROVE` or `REQUEST_CHANGES`, and (b) the PR is in draft status.

---

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in ac9cec59.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in ac9cec59.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in ac9cec59.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in ac9cec59.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
tgrunnagle: Addressed in ac9cec59.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
tgrunnagle: Addressed in ac9cec59.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in ac9cec59.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types.go
tgrunnagle: Addressed in ac9cec59.

cmd/thv-operator/api/v1beta1/mcpexternalauthconfig_types_test.go
tgrunnagle: Addressed in ac9cec59.

jhrozek (dismissed):

rdimitrov (approved):

PR #5362, opened 2026-05-21
Add StorageVersionMigrator controller (opt-in, default off)

ChrisJBurns:
## Summary

Adds a `StorageVersionMigrator` controller to the operator behind an opt-in feature flag (`TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR`). The flag defaults to **false** — this PR has **no functional impact** until an admin explicitly opts in. Required before a future release can drop a deprecated `apiVersion` (e.g. `v1alpha1`) from `spec.versions` without orphaning rows in etcd.

This is the first of three sequential PRs that together deliver the migrator described in #4969. The full work lives in #5011 as a reference; this PR extracts just the controller code.

Part of #4969.

<details>

- New `StorageVersionMigratorReconciler` at `cmd/thv-operator/controllers/storageversionmigrator_controller.go` — issues a Get+Update against each CR on opted-in CRDs, which re-encodes etcd bytes at the current storage version, then trims `status.storedVersions`
- App wiring in `cmd/thv-operator/app/app.go`: `apiextensions/v1` scheme registration, `setupStorageVersionMigrator()` setup, `isStorageVersionMigratorEnabled()` helper defaulting to `false` and failing loudly on an unparsable env-var value
- RBAC regenerated to include the controller's verbs (so the flag can be flipped at any time without permission errors)
- envtest suite under `cmd/thv-operator/test-integration/storageversionmigrator/` with 8 scenarios including a cross-version re-encode test
- **No helm chart flag entry and no `deployment.yaml` env-var wiring** — deferred to PR-C alongside the default-on flip. Reduces the chance of accidental enablement during the deprecation window. Early adopters who know about the feature can opt in via `operator.env`.
- **No documentation** — also lands in PR-C

</details>

<details>

| File | Change |
|------|--------|
| `cmd/thv-operator/controllers/storageversionmigrator_controller.go` | New controller (reconciler, migration cache, RBAC markers, SetupWithManager) |
| `cmd/thv-operator/app/app.go` | Register `apiextensions/v1` in scheme; add `setupStorageVersionMigrator()` + `isStorageVersionMigratorEnabled()` helper (defaults `false`, fails loudly on unparsable values); gate registration on the flag |
| `cmd/thv-operator/test-integration/storageversionmigrator/suite_test.go` | envtest harness |
| `cmd/thv-operator/test-integration/storageversionmigrator/controller_test.go` | 8 ginkgo scenarios |
| `deploy/charts/operator/templates/clusterrole/role.yaml` | Regenerated by `task operator-manifests` — picks up the controller's `+kubebuilder:rbac:` markers |
| `test/e2e/chainsaw/operator/multi-tenancy/setup/assert-rbac-clusterrole.yaml` | Updated to match the regenerated ClusterRole |

</details>

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

**Yes — operator ServiceAccount permissions widen on this PR.** The regenerated `toolhive-operator-manager-role` ClusterRole now grants:

- `get/list/watch` on `apiextensions.k8s.io/customresourcedefinitions`
- `patch/update` on `customresourcedefinitions/status`
- `get/list/update` on `toolhive.stacklok.dev/*` (wildcard within the operator's own API group)

**Specifically affected resources** — two CRDs that previously had **read-only** RBAC (`get;list;watch` only) are now SA-writable via the wildcard `update`:

- `mcpserverentries.toolhive.stacklok.dev`
- `virtualmcpcompositetooldefinitions.toolhive.stacklok.dev`

This widens the SA bound but does **not** widen the controller's behavior: the runtime gate (`isManagedCRD` requiring both the `toolhive.stacklok.dev` group AND the opt-in label) prevents the migrator from writing to anything that hasn't explicitly opted in. The wildcard ships unconditionally so admins can flip the env var on at any time without hitting permission-denied. Templating the RBAC behind a chart value is deferred to PR-C alongside the chart surface for the feature flag.

No CR schema changes. No behavior change for the operator until an admin sets `TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR=true`.

## Test plan

- [x] `task build` passes
- [x] Operator unit tests pass: `go test ./cmd/thv-operator/... | grep -v /test-integration`
- [x] envtest suite passes 8/8: `go test ./cmd/thv-operator/test-integration/storageversionmigrator/...`
- [x] Default-OFF verified: with the flag unset, the operator suppresses the disabled-state log at INFO and the controller is not registered
- [x] Fails loudly on an unparsable env-var value: `TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR=ture` returns a startup error rather than silently disabling
- [ ] `task lint-fix` — blocked by a pre-existing tooling issue (`golangci-lint` built with go1.25 vs. go1.26 target); not introduced by this PR

## Large PR Justification

This PR exceeds 1000 lines, but the bulk is **non-production code** that cannot reasonably be split from the production code it covers:

- **~712 lines** of envtest scenarios (`controller_test.go` + `suite_test.go`) — 8 ginkgo specs including a cross-version re-encode probe that empirically verifies the apiserver's storage-encoder behavior. Shipping the controller without its integration tests would be poor practice.
- **~481 lines** of controller production code — the reconciler is one cohesive unit (migration cache, list pagination, conflict counting, optimistic-locked status patch). Splitting it further would create no independently reviewable boundary.
- **~52 lines** of regenerated RBAC YAML (`role.yaml`) — driven by `+kubebuilder:rbac:` markers in the controller; cannot be authored separately.
- **~76 lines** of app wiring + chainsaw fixture update.

The PR ships behind a default-off feature flag, so the production code is dormant unless explicitly opted in. PR-B (opt-in labels + marker-coverage CI) and PR-C (default-on flip + chart surface + docs) follow as separate, smaller PRs.

- **Why no chart surface**: per design discussion, we want to keep this feature deliberately obscure until docs and the default-on flip ship together in PR-C. Reduces the chance of someone toggling it via casual `helm install --set` without understanding the semantics.
- **Why ship RBAC now**: if RBAC landed only with the default-on flip, anyone manually opting in between PR-A and PR-C would hit permission-denied. Pre-shipping RBAC is harmless when the controller isn't registered.
- **Env-var name**: `TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR` (with prefix) matches sibling operator env vars like `TOOLHIVE_DEFAULT_IMAGE_PULL_SECRETS` and avoids collisions with cluster-wide variables set by other operators or CI tooling.
- **Follow-ups**:
  1. PR-B: opt-in label markers on the 12 `v1beta1` root types + marker-coverage CI test
  2. PR-C: flip default to ON + add helm chart surface (`values.yaml` flag + `deployment.yaml` env var) + user docs + upgrade-guide walkthrough


ChrisJBurns (commented):
DELETED - Review submitted pre-maturely

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,481 @@
...
+// match and the apiserver harmlessly elides the write — there was nothing to
+// migrate. After all CRs have been processed it patches
+// CRD.status.storedVersions down to [<currentStorageVersion>] so a future
+// release can drop deprecated versions from spec.versions without orphaning
+// etcd objects. See https://github.com/kubernetes-sigs/kube-storage-version-migrator/issues/65
+// for the upstream maintainers' explanation of this mechanism.
+//
+// Enabled by default. Opt out operator-wide via
+// operator.features.storageVersionMigrator (ENABLE_STORAGE_VERSION_MIGRATOR=false)
+// for admins who prefer to run kube-storage-version-migrator externally.
```
ChrisJBurns: **[F1 — HIGH | 10/10]** This doc block contradicts the actual default in this PR and points at a helm value path that doesn't exist yet.

The PR description and `app.go`'s `isStorageVersionMigratorEnabled()` (which returns `false` when the env var is unset) both make this controller **default-off, opt-in via `ENABLE_STORAGE_VERSION_MIGRATOR=true`**. There is no `operator.features.storageVersionMigrator` helm value in this PR — that surface is deferred to PR-C per the PR description.

Violates `.claude/rules/go-style.md` → "Keep Comments Synchronized With Code": a doc comment misrepresenting the activation mechanism will send readers hunting for a helm key that doesn't exist.

```suggestion
// Disabled by default in this release. Admins opt in operator-wide via
// ENABLE_STORAGE_VERSION_MIGRATOR=true. The helm chart surface and the
// default-on flip land together in a follow-up PR; until then, early
// adopters can set the env var directly through `operator.env`.
// Per-kind escape hatch: remove the label from the CRD (emergency only —
// will be re-applied by GitOps / helm upgrade).
```

*Raised by code-reviewer + go-architect (both HIGH).*

cmd/thv-operator/app/app.go
```diff
@@ -150,10 +159,53 @@ func setupControllersAndWebhooks(mgr ctrl.Manager, imagePullSecretsDefaults imag
...
+	if !found {
+		return false
+	}
+	enabled, err := strconv.ParseBool(value)
+	if err != nil {
+		setupLog.Info(
+			"invalid boolean value for "+envEnableStorageVersionMigrator+", using default (disabled)",
+			"value", value,
+		)
+		return false
```
ChrisJBurns: **[F2 — MEDIUM | 9/10]** Silent fallback on an unparseable env-var value violates `.claude/rules/go-style.md` → "Constructor Validation: Fail Loudly on Invalid Input":

> Misconfiguration that fails at startup is far easier to diagnose than misconfiguration that silently degrades behavior at runtime.

`ENABLE_STORAGE_VERSION_MIGRATOR=ture` (typo) gets one INFO line buried in startup logs and the feature stays disabled — opposite of what was asked, with no runtime signal the admin can match against.

Preferred fix — return the error so `setupControllersAndWebhooks` fails startup:

```go
func isStorageVersionMigratorEnabled() (bool, error) {
	value, found := os.LookupEnv(envEnableStorageVersionMigrator)
	if !found {
		return false, nil
	}
	enabled, err := strconv.ParseBool(value)
	if err != nil {
		return false, fmt.Errorf(
			"invalid value for %s: %q (expected true/false): %w",
			envEnableStorageVersionMigrator, value, err)
	}
	return enabled, nil
}
```

Fallback option: at minimum log at WARN level — INFO is the wrong severity for "your config is wrong and we're silently ignoring it."

*Raised by code-reviewer (MEDIUM) + go-architect (HIGH).*

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,481 @@
...
+// kubebuilder marker can't enumerate kinds. The runtime gate is the
+// isManagedCRD check inside Reconcile, which requires both the toolhive
+// API group AND the opt-in label. Wildcard RBAC plus isManagedCRD form the
+// defence in depth: RBAC bounds the controller to a single API group, and
+// the label gate further restricts it to opted-in CRDs.
+
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions/status,verbs=update;patch
+//+kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=*,verbs=get;list;update
+//+kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=*/status,verbs=update
```
ChrisJBurns: **[F3 — MEDIUM | 7/10]** These kubebuilder markers regenerate `role.yaml` such that the operator SA gains `get/list/update` on **every current and future** `toolhive.stacklok.dev` resource (and `update` on every `/status` subresource) regardless of whether `ENABLE_STORAGE_VERSION_MIGRATOR` is set.

The documented runtime gate (`isManagedCRD` requiring the opt-in label) defends against the wildcard at execution time, but the chart ships the broader permissions to every operator install unconditionally. That's a real change to the operator SA's permission footprint.

No concrete change request — the design rationale ("defence in depth: RBAC bounds the controller to a single API group, and the label gate further restricts it") is sound. Two asks:

1. **Flag this in the PR's "Does this introduce a user-facing change?" section** so chart consumers see the permission widening on review.
2. **Optional**: template the migrator RBAC behind `if .Values.operator.features.storageVersionMigrator` in `role.yaml` so the default install footprint is unchanged until the chart surface ships in PR-C.

*Raised by code-reviewer, kubernetes-go-expert; toolhive-expert accepts as-is with documented rationale.*

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,481 @@
...
+func (r *StorageVersionMigratorReconciler) patchStoredVersions(
+	ctx context.Context,
+	crd *apiextensionsv1.CustomResourceDefinition,
+	storageVersion string,
+) error {
+	original := crd.DeepCopy()
+	crd.Status.StoredVersions = []string{storageVersion}
+	return r.Client.Status().Patch(ctx, crd,
+		client.MergeFromWithOptions(original, client.MergeFromWithOptimisticLock{}))
+}
```
ChrisJBurns, line 506: **[F6 — INFO | 7/10]** `.claude/rules/operator.md` mandates `controllerutil.MutateAndPatchStatus` for status writes. Here, using the helper would actually be **wrong** — the target is `apiextensions.k8s.io` CRD whose `Status.StoredVersions` is co-owned by kube-apiserver (which appends on first write at each version), and the helper performs a plain merge-patch without the optimistic lock you correctly use here.

A one-line comment will pre-empt future reviewers (human or AI) re-raising this against the operator rule:

```suggestion
// patchStoredVersions overwrites CRD.status.storedVersions to exactly
// [storageVersion], using an optimistic lock on the CRD's resourceVersion so
// a concurrent API-server write rejects the patch and triggers a requeue.
//
// Does NOT use controllerutil.MutateAndPatchStatus: the target CRD is an
// apiextensions.k8s.io type co-owned by kube-apiserver (which appends to
// storedVersions on first write at each version), so the optimistic lock is
// load-bearing here — the helper's plain MergeFrom would race with the
// apiserver's append.
func (r *StorageVersionMigratorReconciler) patchStoredVersions(
	ctx context.Context,
	crd *apiextensionsv1.CustomResourceDefinition,
	storageVersion string,
) error {
	original := crd.DeepCopy()
	crd.Status.StoredVersions = []string{storageVersion}
	return r.Client.Status().Patch(ctx, crd,
		client.MergeFromWithOptions(original, client.MergeFromWithOptimisticLock{}))
}
```

*Raised by kubernetes-go-expert + toolhive-expert.*

cmd/thv-operator/app/app.go
```diff
@@ -150,10 +159,53 @@ func setupControllersAndWebhooks(mgr ctrl.Manager, imagePullSecretsDefaults imag
 	if err := setupAggregationControllers(mgr, imagePullSecretsDefaults); err != nil {
 		return err
 	}
+	if isStorageVersionMigratorEnabled() {
+		if err := setupStorageVersionMigrator(mgr); err != nil {
+			return err
+		}
+	} else {
+		setupLog.Info("StorageVersionMigrator disabled", "envVar", envEnableStorageVersionMigrator)
```
ChrisJBurns: **[F7 — LOW | 5/10]** Per `.claude/rules/go-style.md` → Logging: *"Silent success — no output at INFO or above for successful operations."* Disabled-by-default is the steady-state for this release, not a noteworthy event. Every operator restart will emit this line.

Drop to DEBUG, or remove the log entirely. The enabled path is the one worth surfacing at INFO.

*Raised by toolhive-expert.*

cmd/thv-operator/app/app.go
```diff
@@ -42,10 +44,17 @@ var (
 	setupLog = log.Log.WithName("setup")
 )
 
+// envEnableStorageVersionMigrator is the opt-in for the StorageVersionMigrator
+// controller. The controller defaults to OFF in this release so the change can
+// ship safely without functional impact. Set to "true" (or "1", "t") to enable.
+// A follow-up release will flip the chart default to true.
+const envEnableStorageVersionMigrator = "ENABLE_STORAGE_VERSION_MIGRATOR"
```
ChrisJBurns: **[F8 — LOW | 5/10]** Sibling toolhive-specific operator env vars use the `TOOLHIVE_` prefix (e.g. `TOOLHIVE_DEFAULT_IMAGE_PULL_SECRETS` in this same file). Generic-named env vars risk collisions with cluster-wide variables set by other operators or CI tooling.

```suggestion
const envEnableStorageVersionMigrator = "TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR"
```

*Raised by toolhive-expert.*

ChrisJBurns (commented):
**Agents consulted**: kubernetes-go-expert, code-reviewer, go-architect, toolhive-expert (go-security-reviewer's sandbox couldn't read the diff — its threat-model checklist was independently covered by the others)

### Consensus Summary
| F1 | Doc comment "Enabled by default" contradicts default-OFF | 10/10 | HIGH | Fix |
| F2 | Silent env-var parse fallback (fail-loudly rule) | 9/10 | MEDIUM | Fix |
| F3 | Wildcard RBAC widens group-wide permissions | 7/10 | MEDIUM | Discuss |
| F6 | `Status().Patch` exception to `MutateAndPatchStatus` needs in-code note | 7/10 | INFO | Polish |
| F7 | Disabled-state INFO log should be DEBUG | 5/10 | LOW | Optional |
| F8 | Env var name lacks `TOOLHIVE_` prefix | 5/10 | LOW | Optional |

### Overall

The mechanism is sound and the test design is genuinely strong. The cross-version RV-bump test ("re-encodes CRs that are stored at a prior storage version") is the load-bearing empirical proof that the apiserver actually re-encodes etcd bytes on cross-version Update — that's the only thing reviewers really need to see to trust the design, and it's there with explicit pre/post resourceVersion assertions. The pagination test wraps APIReader with a list-call counter for direct continue-token proof; the partial-failure and conflict-retry tests pin the storedVersions invariant; the use of `APIReader` (informer bypass) plus runtime `isManagedCRD` re-verification correctly handles the TOCTOU window between watch-time and reconcile-time. The defense-in-depth around the wildcard RBAC (RBAC bounds the API group; the label gate narrows within it) is well-reasoned and the `errMigrationRetriedDueToConflicts` sentinel correctly prevents trimming storedVersions when any per-CR write is unverified.

The findings here are mostly polish, not architectural. One HIGH and two MEDIUMs:

- **F1 (HIGH)** is straightforward doc/code drift — the type doc says one thing, the env-var gate does another, and the referenced `operator.features.storageVersionMigrator` helm path doesn't exist yet. Fixable in a single comment block.
- **F2 (MEDIUM)** is the only operability concern: `ENABLE_STORAGE_VERSION_MIGRATOR=ture` silently disables the feature with an INFO log. Per `.claude/rules/go-style.md` "Constructor Validation: Fail Loudly", that misconfiguration should fail startup or at minimum surface at WARN.
- **F3 (MEDIUM)** is the wildcard RBAC. The runtime gate (`isManagedCRD`) defends against the broad RBAC at execution time, but the regenerated chart now ships `groups=toolhive.stacklok.dev,resources=*` on every install regardless of whether the migrator is on. Worth surfacing in the PR's user-facing-change section even if the design is intentional.

Nothing here blocks the merge in a strict sense; all findings have unambiguous, contained fixes. The HIGH is HIGH only because doc/code drift on the activation mechanism is the kind of thing that causes operator escalations during a deprecation window.

### Documentation

`cmd/thv-operator/controllers/storageversionmigrator_controller.go:97-101` carries the contradicting doc comment (F1). No other documentation files in this PR need updating, but I'd suggest noting the operator SA permission widening (F3) in the PR's "Does this introduce a user-facing change?" section so chart consumers see it on review.

---
*Generated with Claude Code*

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Added a "Does this introduce a user-facing change?" section to the PR description explicitly listing the operator-SA permission widening (`apiextensions.k8s.io/customresourcedefinitions` get/list/watch + status patch/update; `toolhive.stacklok.dev/*` get/list/update + `*/status` update). Templating behind a chart value is deferred to PR-C alongside the chart surface for the feature flag — shipping the RBAC unconditionally in PR-A means admins can flip the env var on at any time without permission-denied. Closing this thread on the documented rationale.

cmd/thv-operator/app/app.go
ChrisJBurns: Fixed in d6ca39c19. Demoted to `V(1)` — disabled-by-default is the steady-state for this release and shouldn't surface at INFO on every operator restart.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Fixed in d6ca39c19. Doc block now reads "Disabled by default in this release. Admins opt in operator-wide via TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR=true. The helm chart surface and the default-on flip land together in a follow-up PR…" — matches the actual default-off behavior and points at the env var (the helm path doesn't exist yet).

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Applied in d6ca39c19. `patchStoredVersions` now carries a note explaining that the operator-wide `MutateAndPatchStatus` helper would race with kube-apiserver's append on first write at each version, so the optimistic lock is load-bearing here — pre-empts the rule-driven flag in future reviews.

cmd/thv-operator/app/app.go
ChrisJBurns: Fixed in d6ca39c19. Renamed to `TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR` — aligns with sibling operator env vars and avoids cross-operator collisions. PR description and design doc updated to match.

cmd/thv-operator/app/app.go
ChrisJBurns: Fixed in d6ca39c19. `isStorageVersionMigratorEnabled` now returns `(bool, error)` and `setupControllersAndWebhooks` propagates the error so an unparsable value (e.g. `TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR=ture`) fails startup with `invalid value for TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR: "ture" (expected true/false): …` instead of a silent INFO line.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,519 @@
...
+// EventReasonMigrationSucceeded and EventReasonMigrationFailed are the event
+// reasons emitted on the owning CRD when a migration completes or fails.
+const (
+	EventReasonMigrationSucceeded = "StorageVersionMigrationSucceeded"
+	EventReasonMigrationFailed    = "StorageVersionMigrationFailed"
+)
+
+const (
+	defaultMigrationCacheTTL = 1 * time.Hour
+	defaultListPageSize      = 500
```
JAORMX: A 500-item `UnstructuredList` page is fully decoded into RAM before iteration. For an MCPServer with PodTemplateSpec, envFiles, and secrets (~30–100 KB per object), one page is 15–50 MB. Combined with `restoreOne`'s second `live` copy per CR and concurrent reconciles across 12 CRDs, the operator's memory ceiling is unbounded by typical chart-side `resources.limits.memory`.

```suggestion
	defaultListPageSize      = 100
```

Or document the per-CR memory ceiling so chart-side limits can be sized accordingly.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,519 @@
...
+//
+// Chart-consumer note: these markers regenerate role.yaml, so every chart
+// install gains get/list/update on toolhive.stacklok.dev/* regardless of
+// whether the migrator is opted in via TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR.
+// Templating this rule behind a helm conditional is deferred to PR-C alongside
+// the rest of the chart surface for the feature flag.
+
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions/status,verbs=update;patch
+//+kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=*,verbs=get;list;update
```
JAORMX, line 104: Wildcard grants `update` on every toolhive resource, but `mcpserverentries` and `virtualmcpcompositetooldefinitions` had read-only RBAC before this PR (`get;list;watch` only, intentionally). Wildcard silently widens both to writable at the ServiceAccount level. The runtime `isManagedCRD` gate prevents the controller from exercising this, but RBAC bound is now wider than baseline.

Fix: enumerate the resources the migrator actually targets, or document the widening in the PR body so admins running `helm diff` know what they're signing up for.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,519 @@
...
+//+kubebuilder:rbac:groups=toolhive.stacklok.dev,resources=*/status,verbs=update
```
JAORMX: Unused. `restoreOne` updates the CR main resource (line 377); `patchStoredVersions` patches the *CRD's* status (line 403), not any toolhive CR status. grep confirms zero callers writing CR statuses.

Remove this marker — RBAC tightens, no functional change.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,519 @@
...
+	// (SetupWithManager, Reconcile, and any future entrypoint). Without it,
+	// two concurrent callers seeing zero values could race on the field writes.
+	initOnce sync.Once
+}
+
+// Reconcile runs for each opted-in toolhive.stacklok.dev CRD event. See the
+// package-level docs on StorageVersionMigratorReconciler for the full flow.
+// Returns a non-nil error to trigger exponential backoff; the CRD watch
+// re-enqueues on any status change, so explicit requeue intervals are not used.
+func (r *StorageVersionMigratorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
JAORMX, line 162: No observability for migration progress or stuck states. A loop returning `errMigrationRetriedDueToConflicts` (see comment on line 349) emits only `V(1)` logs — invisible at default verbosity. An SRE has no way to tell whether the migrator finished, progressed, or is stuck.

Add metrics:
- `storageversionmigrator_crs_processed_total{crd, outcome="success|cache_hit|conflict|not_found"}` (counter)
- `storageversionmigrator_reconcile_duration_seconds{crd}` (histogram)
- `storageversionmigrator_cache_size{crd}` (gauge)
- `storageversionmigrator_consecutive_conflict_passes{crd}` (gauge)

At minimum, INFO-log or emit a Normal event when conflict cycles exceed N consecutive passes — the failure mode in the line 349 comment needs to be visible at default verbosity.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,519 @@
...
+		}
+
+		continueToken = list.GetContinue()
+		if continueToken == "" {
+			break
+		}
+	}
+
+	if len(errs) == 0 && conflicts > 0 {
+		return errMigrationRetriedDueToConflicts
```
JAORMX, line 419: Current logic treats any single 409 as "post-conflict state unverified" and aborts the whole-pass trim. But: a 409 on Update means another writer just succeeded against the object, and every successful write through the apiserver re-encodes at the storage version. The 409 is *evidence* the object is at the storage version — not a reason to doubt it.

Upstream `kube-storage-version-migrator` (`pkg/migrator/core.go`) treats 409 as a per-object transient: re-Get + retry per item. Current behavior + controller-runtime's exponential backoff (5ms → ~16min) means under steady-state churn — e.g., MCPServerReconciler writing `.status` on a 10s cadence while the migrator runs — the migrator can repeatedly hit one conflict per pass and never converge in any useful time horizon.

Two changes that compose:
1. In `restoreOne`, on `IsConflict` re-Get + retry the Update with bounded N attempts (3–5) before counting as conflict at the outer level.
2. In `Reconcile`, return `ctrl.Result{RequeueAfter: 30 * time.Second}, nil` for the `errMigrationRetriedDueToConflicts` path rather than a non-nil error — avoids exponential backoff applying to a normal-steady-state condition.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,519 @@
...
+// restoreOne issues a plain Get + Update on the live CR. The apiserver
+// re-encodes the request body at the current storage version and compares
+// it to etcd's record; when the CR was originally stored at a different
+// apiVersion the bytes differ, the write proceeds, and etcd is re-encoded
+// at the current storage version. When the CR is already at the current
+// storage version the bytes match and the apiserver harmlessly elides the
+// write — there was nothing to migrate. The Update goes through the main
+// resource, so validating/mutating admission webhooks on the kind see this
+// request as part of normal admission flow; only requests that actually
+// persist produce downstream state changes. Returns the live object after
```
JAORMX: The "only requests that actually persist produce downstream state changes" clause is inaccurate. Admission webhooks (mutating + validating) fire on **every** Update before the storage layer's bytes-equality check runs. Only the etcd write is elided in the no-op case; admission has already executed, including all side-effects (audit logs, webhook calls, accounting).

For toolhive today this is moot (no in-tree admission webhooks on these kinds), but the comment will mislead future readers reasoning about webhook load — and is materially wrong if a third-party admission webhook (Kyverno, Gatekeeper) is in the cluster.

Suggested rewording: "only requests that actually persist bump resourceVersion and trigger watch notifications; admission webhooks fire on every Update regardless of whether bytes change."

cmd/thv-operator/controllers/storageversionmigrator_controller.go
```diff
@@ -0,0 +1,519 @@
...
+// the update so the caller can record its post-update resourceVersion in
+// the cache.
+func (r *StorageVersionMigratorReconciler) restoreOne(
+	ctx context.Context,
+	gvk schema.GroupVersionKind,
+	original *unstructured.Unstructured,
+) (*unstructured.Unstructured, error) {
+	live := &unstructured.Unstructured{}
+	live.SetGroupVersionKind(gvk)
+	if err := r.APIReader.Get(ctx, client.ObjectKeyFromObject(original), live); err != nil {
```
JAORMX: `restoreCRs` already has the full object from the List page (line 307 returns full objects, not OnlyMetadata). This Get is a second apiserver read of the same object — doubles read load and widens the conflict window.

Pass `original` to `r.Update` directly. The apiserver still re-encodes; storage-version-migration property is preserved.

cmd/thv-operator/test-integration/storageversionmigrator/controller_test.go
```diff
@@ -0,0 +1,801 @@
...
+}
+
+// ------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------
+
+var _ = Describe("StorageVersionMigrator", func() {
+	Describe("Reconcile", func() {
+
+		It("is a noop when storedVersions is already [storageVersion] and only one version is served", func() {
```
JAORMX, line 224: The `migrationCache` has an injectable `now func()` hook (controller.go:462) but no test uses it. Untested:

- Cache hit prevents redundant Update (controller.go:317-336).
- TTL expiration in `has()` (controller.go:486-489).
- `gc()` removes expired entries (controller.go:501-512).
- The `manager.RunnableFunc` GC ticker (controller.go:234-245).

The `now func()` hook exists for exactly this purpose. An undetected GC bug means unbounded cache growth — `.claude/rules/go-style.md` flags this under Resource Leaks. Suggest 3–4 table-driven tests at the cache level.

cmd/thv-operator/test-integration/storageversionmigrator/controller_test.go
```diff
@@ -0,0 +1,801 @@
...
+			Expect(err).To(HaveOccurred(),
+				"reconcile must return an error when a Conflict was swallowed")
+			Expect(getStoredVersions(spec.Name)).To(Equal([]string{"v1alpha1", "v1beta1"}),
+				"storedVersions must not be trimmed on a pass with any swallowed Conflict")
+
+			// Drop the injection and retry. The cache may have absorbed the
+			// non-conflicting CR's RV from the first pass — that's fine, the
+			// conflicting one was never recorded in the cache so it'll be
+			// re-attempted, succeed, and let the storedVersions patch fire.
+			injectConflict = false
```
JAORMX, line 676: This toggles `injectConflict = false` after pass 1 — verifies *recovery from a one-shot conflict*, not *resistance to sustained conflict pressure*. The docstring at controller.go:172-177 claims conflicts "self-heal on the next reconcile", but nothing validates behavior when every pass hits ≥1 conflict.

Add a test with persistent conflict injection across N=5 sequential reconciles asserting:
- `storedVersions` stays untrimmed throughout.
- Zero `MigrationSucceeded` events fire.
- Bounded log/event noise.

This is the test that would have caught the convergence concern raised in controller.go:349.

cmd/thv-operator/app/app.go
```diff
@@ -150,10 +160,56 @@ func setupControllersAndWebhooks(mgr ctrl.Manager, imagePullSecretsDefaults imag
...
+	}
+	return nil
+}
+
+// isStorageVersionMigratorEnabled reports whether the StorageVersionMigrator
+// controller should be registered. Defaults to false in this release — admins
+// must explicitly opt in via TOOLHIVE_ENABLE_STORAGE_VERSION_MIGRATOR=true.
+// An unparsable value returns an error so startup fails loudly rather than
+// silently disabling the feature an admin asked to turn on.
+func isStorageVersionMigratorEnabled() (bool, error) {
```
JAORMX, line 199: PR body claims "Default-OFF verified", but verification was manual. `isStorageVersionMigratorEnabled` has no unit test, and nothing asserts `setupStorageVersionMigrator` is not called when the env var is unset/false — exactly the regression that ships a "disabled" feature as silently enabled.

Suggest a 4-row table-driven test against `isStorageVersionMigratorEnabled` covering: unset, `"true"`, `"false"`, unparseable.

JAORMX (commented):
Reviewed with a deep-dive plus 7 specialist agent reviews (concurrency, RBAC, scale, webhook surface, conventions, tests, edge cases). Core mechanism is correct: Get+Update through the storage encoder, optimistic-locked storedVersions trim, schema-identity by construction (validated against #4849's commit message and the v1alpha1 wrapper types).

10 must-fix items below — 2 RBAC, 1 doc accuracy, 4 operational/scale, 3 test coverage. The most architecturally significant is the conflict-poisoning convergence behavior (the comment on line 349): a deliberate tightening over upstream `kube-storage-version-migrator` semantics that, combined with controller-runtime's exponential backoff, can fail to converge in clusters with status-update churn. Worth a thread before merge.

Should-fix items and nits to follow in a separate review.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Applied in 0582b752b (controller logic) + b7c4116fd (envtest spec validating it). The previous commit b7c4116fd accidentally dropped the controller-file changes during a rebase conflict resolution — 0582b752b recovers them. `restoreOne` now retries the Update up to `restoreOneMaxRetries=3` attempts when IsConflict surfaces, re-Getting the live object between attempts to refresh the resourceVersion. `Reconcile` returns `ctrl.Result{RequeueAfter: 30s}` with **nil** error when `restoreCRs` returns the sentinel — avoiding controller-runtime's exponential backoff for what is normal steady-state behaviour under concurrent `.status` churn. The envtest `under sustained conflict pressure...` validates the contract across 6 sequential passes (no error, RequeueAfter=30s, storedVersions stays untrimmed, zero MigrationSucceeded events).

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Applied in 0582b752b (the actual controller changes — b7c4116fd dropped them during a rebase conflict). The list-page object is now passed directly to `r.Update` on attempt 0; the redundant `Get` only fires on retry attempts (N>0) where refreshing the RV is the whole point of the retry. Pairs with the #5 retry fix — without per-CR retry, dropping the Get would worsen conflict frequency.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Marker removal applied in 0582b752b (controller .go) + b7c4116fd (regenerated role.yaml + chainsaw fixtures). The */status kubebuilder marker is gone from the controller; regenerated `role.yaml` via `task operator-manifests`; both chainsaw RBAC fixtures (`single-tenancy` and `multi-tenancy`) updated to match. Confirmed: `restoreOne` updates the main resource and `patchStoredVersions` patches the CRD's status (under apiextensions.k8s.io), so the marker was indeed dead. (Split across two commits because of an intervening rebase mishap on the controller file.)

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Comment rewritten in 0582b752b (b7c4116fd dropped this controller change during a rebase conflict). The new wording: 'apiserver harmlessly elides the etcd write and watch fanout — but validating/mutating admission webhooks still fire on every Update, before the bytes-equality elision check, so callers must not assume an already-migrated CR is webhook-free.' Moot for toolhive today (no admission webhooks on these kinds) but accurate for clusters running Kyverno/Gatekeeper.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Light version applied in 0582b752b (controller changes were dropped from b7c4116fd by a rebase conflict; recovered here). The reconciler now tracks per-CRD `conflictPasses` (mutex-guarded map). After `sentinelConflictLogThreshold=5` consecutive sentinel passes the controller emits an INFO log surfacing operator-visible signal that migration is not converging. Counter resets to zero on the next successful `patchStoredVersions`. Full metrics pipeline (`crs_processed_total`, `reconcile_duration_seconds`, `cache_size`, `consecutive_conflict_passes`) deferred to follow-up — significant scope for PR-A.

cmd/thv-operator/app/app.go
ChrisJBurns: Unit test added in b7c4116fd at `cmd/thv-operator/app/app_test.go`. Table-driven `TestIsStorageVersionMigratorEnabled` covers 7 rows: unset (default-off), `true`/`false`, ParseBool truthy/falsy (`1`/`0`), unparseable typo, and empty string. Asserts the error path's message contains both the env-var name and the bad value. (Subtests aren't `t.Parallel`-marked because `t.Setenv` panics if any ancestor is parallel — runtime constraint, not stylistic.)

cmd/thv-operator/test-integration/storageversionmigrator/controller_test.go
ChrisJBurns: Added in b7c4116fd: envtest spec `under sustained conflict pressure, storedVersions never trims and no success event is emitted`. Injects IsConflict on one CR on **every** pass (never toggled off), runs 6 sequential reconciles (> `sentinelConflictLogThreshold=5`), and asserts: nil error each pass, `RequeueAfter=30s` each pass, storedVersions stays `[v1alpha1, v1beta1]` throughout, and `Consistently` no `MigrationSucceeded` event ever fires. This is the test that validates the #5 architectural fix's convergence contract.

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: Lowered to `defaultListPageSize = 100` — landed via Chris's suggestion-block accept (8846ce6ee) and confirmed present at HEAD (0582b752b). Brings peak list-page envelope to ~5 MB. Worth noting on framing: controller-runtime's `MaxConcurrentReconciles` defaults to **1** for this controller (the migrator has one controller), so reconciles are sequential — not 'concurrent across 12 CRDs'. Peak memory is bounded by one list page + one per-CR live + cache entries, comfortably inside the default 128Mi operator memory limit. The 500→100 change is defensive headroom, not a leak fix.

cmd/thv-operator/test-integration/storageversionmigrator/controller_test.go
ChrisJBurns: Respectfully pushing back — these are already covered. The `migrationCache`'s `now func()` hook **is** used, in the **unit-test** file `cmd/thv-operator/controllers/storageversionmigrator_controller_test.go`:

- `TestMigrationCache_TTLExpiryLazilyEvictsInHas` — frozen clock advanced past TTL; asserts has() returns false AND entry is removed from the map.
- `TestMigrationCache_GCEvictsOnlyExpiredEntries` — frozen clock; asserts gc() removes only entries whose expiresAt passed.
- `TestMigrationCache_AddOverwritesExistingEntry` — verifies add() refreshes the expiry timer.
- `TestMigrationCache_ConcurrentAccess` — 16 goroutines × 200 iters × {has, add, gc} under -race.
- `TestRestoreCRs_CacheSkipsAlreadyMigratedCRs` — verifies cache hit prevents redundant Update via interceptor counting.

The only thing not unit-tested is the `manager.RunnableFunc` ticker itself, which would require a running manager — deliberate skip (the ticker just calls `gc()` which is fully covered).

cmd/thv-operator/controllers/storageversionmigrator_controller.go
ChrisJBurns: The general wildcard widening is already documented in the PR body's 'Does this introduce a user-facing change?' section, but you're right that `mcpserverentries` and `virtualmcpcompositetooldefinitions` deserve specific call-out as previously read-only resources that are now SA-writable. Will refine the PR body wording. Keeping the wildcard itself per the staging plan (PR-C revisits alongside chart-conditional gating). The runtime `isManagedCRD` gate keeps the controller from actually exercising the widened verbs on those previously-RO CRDs in this release.

JAORMX (approved):

PR #5364, opened 2026-05-21
Freeze MCPServer generation per pod via downward API

### LINKED ISSUE #5360: shouldSkipStatefulSetApply generation gate defeated by live-updating RunConfig ConfigMap during proxy runner     rolling update (regression in v0.28.1)
**Summary**

  After upgrading to v0.28.1, MCP workload pods consistently end up running stale images/config after a helm
  upgrade. The generation gate introduced in #5024 does not prevent this because the RunConfig ConfigMap is a
  live-updating volume mount — all proxy runner pods (old and new) converge on the same generation value,
  neutralising the gate.

  **Environment**

  - Toolhive operator: ghcr.io/stacklok/toolhive/operator:v0.28.1
  - Toolhive proxyrunner: ghcr.io/stacklok/toolhive/proxyrunner:v0.28.1
  - Kubernetes: v1.32.2 (k3s)
  - Helm chart: toolhive-operator 0.28.1

  **Steps to Reproduce**

  1. Deploy an MCPServer via helm. Confirm everything is healthy (MCPServer generation=1, StatefulSet
  mcpserver-generation annotation=1, workload pod running).
  2. Perform a helm upgrade that changes the MCPServer spec.image to a new tag.
  3. Observe the proxy runner Deployment begin a rolling update (old RS pod and new RS pod coexist briefly).
  4. Observe the workload StatefulSet cycling between ControllerRevisions — alternating between the new image
  and the old image — until the workload pod lands in ImagePullBackOff or ErrImagePull on whichever revision
  lost the final race.

  **Root Cause**

  In v0.28.1, mcpserver_generation is read from the RunConfig ConfigMap mounted at
  /etc/runconfig/runconfig.json. Kubernetes ConfigMap volume mounts are live-updated — within ~30–60 seconds all
   pods mounting the same ConfigMap see the updated contents.

  shouldSkipStatefulSetApply skips if theirsGen > ourGen. Since both pods now read the same generation N from
  the live-updated ConfigMap, N > N is false — neither pod is gated. They race:

  - New proxy runner applies the StatefulSet with new-image-tag + generation annotation N
  - Old proxy runner (still reading gen=N, still holding old-image-tag in its CLI arg) overwrites it back to
  old-image-tag + generation annotation N

  The result is rapid alternating ControllerRevisions. The workload pod gets stuck on whichever revision it was
  last assigned.

  This was confirmed live by inspecting two consecutive ControllerRevisions immediately after a spec.image
  patch:

  ControllerRevision 7 → image: new-image-tag    (new proxy runner won)
  ControllerRevision 8 → image: old-image-tag    (old proxy runner overwrote it)

  The mcpserver-generation annotation on both revisions was identical (3), confirming neither pod was gated
  against the other.

  **Why the Fix in #5024 Does Not Help Here**

  PR #5024 was designed to prevent the old proxy runner from clobbering the new runner's apply when the old pod
  has a lower mcpserver_generation. That design was sound for the previous architecture where generation was a
  fixed CLI arg. In v0.28.1, generation comes from a shared live-updating ConfigMap, so the old proxy runner's
  generation is updated out from under it, making it an equal peer rather than a lower-generation stale actor.

  **Expected Behaviour**

  A proxy runner pod that was created before a helm upgrade should not be able to overwrite a StatefulSet spec
  that a newer proxy runner pod has applied with updated configuration.

  **Suggested Fix**

  The underlying issue is that the proxy runner uses the CLI positional arg as the authoritative source for the
  container image, but the RunConfig ConfigMap as the authoritative source for mcpserver_generation. These are
  now decoupled — old pods can get a new generation without getting the new image.

  The fix is to make the RunConfig ConfigMap the single authoritative source for all fields including the
  container image. If old proxy runner pods apply the image from RunConfig (not from their stale CLI arg), both
  old and new pods would apply the same correct image during a rolling update, and the race becomes harmless.

  Alternatively: snapshot mcpserver_generation into a source that does not live-update during the pod's lifetime
   (e.g. a pod annotation set via the downward API at creation time), so old pods cannot inadvertently acquire a
   generation value they were not started with.

ChrisJBurns:
## Summary

The apply-gate at `pkg/container/kubernetes/client.go:530` (PR #5024) was silently defeated as soon as `MCPServerGeneration` started flowing through the `/etc/runconfig/runconfig.json` ConfigMap volume — that volume is mounted without `subPath`, so Kubernetes live-updates it inside every running pod. A restarted old-RS proxyrunner pod (OOM, drain, attach-failure `c.exitFunc(1)`) re-reads the file after a helm upgrade and picks up the new generation `N`, while still holding the old image in its CLI args. Both old and new pods then call `DeployWorkload` with the same `ourGen=N`, the strict-greater-than gate cannot tell them apart, and the stale-image apply clobbers the fresh one — manifesting as alternating `ControllerRevision`s and `ImagePullBackOff` in the bug report.

Freeze `MCPServerGeneration` per pod via the Kubernetes downward API. The operator stamps `toolhive.stacklok.dev/mcpserver-generation: <m.Generation>` onto the proxy Deployment's pod template and projects that annotation into the proxyrunner container as the `THV_MCPSERVER_GENERATION` env var. The proxyrunner uses the env var as an override on the file value, so two coexisting proxyrunner pods carry distinct generations again — symmetric with how the image is already frozen via the CLI positional arg.

Closes #5360.

<details>

- **Shared contract**: `pkg/container/kubernetes/client.go` — new exported constant `EnvVarMCPServerGeneration = "THV_MCPSERVER_GENERATION"` alongside the existing `RunConfigMCPServerGenerationAnnotation`. Doc comment on the annotation updated to describe the downward-API projection.
- **Operator**: `cmd/thv-operator/controllers/mcpserver_controller.go` — stamps the annotation onto `deploymentTemplateAnnotations` (next to the existing runconfig-checksum) and injects the env var into the proxyrunner container via `valueFrom.fieldRef`. `deploymentNeedsUpdate` mirrors both fields so drift detection stays stable.
- **Proxyrunner**: `cmd/thv-proxyrunner/app/run.go` — `tryLoadConfigFromFile` applies the env-var override on the loaded `RunConfig.MCPServerGeneration` (warn + fall through on parse error).
- **Tests**:
  - Lock-in regression test (`pkg/container/kubernetes/client_test.go`) — two `DeployWorkload` calls with equal gen + different images; asserts the second clobbers the first. Documents what the gate cannot defend against alone.
  - TDD red→green test (`cmd/thv-proxyrunner/app/run_test.go`) — confirms env value (3) wins over file value (5) when both are set. Failed before the fix.
  - Downward-API shape test (`cmd/thv-operator/controllers/mcpserver_resource_overrides_test.go`) — asserts the proxyrunner container declares the env var with `ValueFrom.FieldRef.FieldPath` pointing at the pod's mcpserver-generation annotation.

</details>

<details>

| File | Change |
|------|--------|
| `pkg/container/kubernetes/client.go` | Add `EnvVarMCPServerGeneration` constant; expand doc comment on `RunConfigMCPServerGenerationAnnotation` to cover the downward-API projection. |
| `cmd/thv-proxyrunner/app/run.go` | Add `strconv`/`kubernetes` imports; call new `applyMCPServerGenerationOverride` after `runner.ReadJSON`; helper reads the env var, parses as int64, overrides `config.MCPServerGeneration`. |
| `cmd/thv-operator/controllers/mcpserver_controller.go` | Add `strconv` import. In `deploymentForMCPServer`: stamp `RunConfigMCPServerGenerationAnnotation` on `deploymentTemplateAnnotations`; append the downward-API `THV_MCPSERVER_GENERATION` env var. Mirror both in `deploymentNeedsUpdate` so drift comparison stays equal. |
| `cmd/thv-proxyrunner/app/run_test.go` | New file. Test `TestTryLoadConfigFromFile_MCPServerGenerationEnvOverride`. |
| `pkg/container/kubernetes/client_test.go` | New test `TestDeployWorkload_EqualGenerationDifferentImageClobbers`. |
| `cmd/thv-operator/controllers/mcpserver_resource_overrides_test.go` | New test `TestDeploymentForMCPServer_MCPServerGenerationDownwardAPI`. Update existing annotation-count and env-var assertions to account for the new annotation + env var (added at the same relative position in both build and drift paths). |

</details>

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Dependency update
- [ ] Documentation
- [ ] Other

## Test plan

- [x] `task test` passes (exit 0)
- [x] `task build` passes
- [x] New unit tests: `TestTryLoadConfigFromFile_MCPServerGenerationEnvOverride`, `TestDeploymentForMCPServer_MCPServerGenerationDownwardAPI`, `TestDeployWorkload_EqualGenerationDifferentImageClobbers`
- [x] Existing drift-detection tests (`TestMCPServerDeploymentNeedsUpdate_*`) updated and pass — confirms `deploymentNeedsUpdate` stays in sync with `deploymentForMCPServer` so reconcile doesn't churn.
- [ ] Manual live-cluster reproduction recommended before merge (kind cluster + helm upgrade with `maxSurge=1, maxUnavailable=0` + induced container restart on old-RS pod; see issue #5360 for the exact procedure).

- **Upgrade ordering**: the operator and proxyrunner changes are mutually independent during a rolling upgrade. Operator-only or proxyrunner-only upgrades leave the race window open until both sides are at this version — same wire-contract caveat as PR #5024. Worth calling out in release notes.
- **Annotation value `"0"`**: when reading existing deployments built by older operators, the pod template won't carry the annotation and the downward-API env var resolves to an empty string. The proxyrunner's `applyMCPServerGenerationOverride` returns without modifying the config in that case, so the file value wins (existing behavior). After a single reconcile, the operator backfills the annotation and the override kicks in.
- **Position of the env var in the container `env` slice matters** — `deploymentNeedsUpdate` uses `equality.Semantic.DeepEqual` on the slice. I positioned the new entry in both build sites at the same relative location (after Redis password / user overrides, before embedded auth env vars).
- **Out of scope for this PR**: the `spec.replicas` clobber dimension (tracked in #4484) — same root cause family, different field, separate fix.
- The fix path was: investigation → lock-in test for the broken invariant → TDD red test for the env-var override → operator/proxyrunner wiring → drift-detection updates. Each commit compiles and `task test` passes independently.


cmd/thv-proxyrunner/app/run.go
```diff
@@ -159,6 +163,28 @@ func tryLoadConfigFromFile() (*runner.RunConfig, error) {
...
+	if raw == "" {
+		return
+	}
+	gen, err := strconv.ParseInt(raw, 10, 64)
+	if err != nil {
+		slog.Warn("ignoring unparseable env var; falling back to runconfig value",
+			"env", kubernetes.EnvVarMCPServerGeneration, "value", raw, "err", err)
+		return
+	}
+	runConfig.MCPServerGeneration = gen
```
ChrisJBurns, line 198: Per `.claude/rules/go-style.md` ("Validate Parsed Results"), a successful `strconv.ParseInt` only proves the input is a valid int64 — it does not prove the value is meaningful for `MCPServerGeneration`. `metadata.generation` is a monotonically-increasing positive integer per Kubernetes API conventions; a negative value here cannot have come from a legitimate downward-API projection of a real annotation, and accepting it would let a bogus env var (typo, manual override, malicious mutation) downgrade the gate's monotonic version.

The `0` case is intentional backward-compat (older operators that haven't backfilled the annotation), so it should keep falling through to the file value. Suggest guarding the negative case explicitly:

```suggestion
	gen, err := strconv.ParseInt(raw, 10, 64)
	if err != nil {
		slog.Warn("ignoring unparseable env var; falling back to runconfig value",
			"env", kubernetes.EnvVarMCPServerGeneration, "value", raw, "err", err)
		return
	}
	if gen < 0 {
		slog.Warn("ignoring negative env var; falling back to runconfig value",
			"env", kubernetes.EnvVarMCPServerGeneration, "value", raw)
		return
	}
	runConfig.MCPServerGeneration = gen
```

Minor nit while you're here: a single `slog.Debug` log on the success path would be useful when diagnosing future occurrences of #5360-class issues — right now there's no signal in pod logs that the override actually applied vs. fell through.

cmd/thv-operator/controllers/mcpserver_controller.go
```diff
@@ -1759,6 +1780,21 @@ func (r *MCPServerReconciler) deploymentNeedsUpdate(
 			}
 		}
 
+		// Project the MCPServer generation pod-template annotation into the
+		// proxyrunner container via the downward API. Position mirrors the
+		// construction site in deploymentForMCPServer (after Redis password,
+		// before embedded auth) so DeepEqual against container.Env succeeds.
+		// See #5360.
+		expectedProxyEnv = append(expectedProxyEnv, corev1.EnvVar{
```
ChrisJBurns, line 1797: Heads-up that the comment claim "Position mirrors the construction site in deploymentForMCPServer (after Redis password, before embedded auth)" is slightly inaccurate: `deploymentNeedsUpdate`'s `expectedProxyEnv` does **not** include the Redis password env var at all (no call to `r.buildRedisPasswordEnvVar(m)` in this function — search confirms the only call site is line 1153 in `deploymentForMCPServer`).

This is a **pre-existing latent bug**, not something you introduced — but it means: when `sessionStorage.provider == "redis"` and a `passwordRef` is set, the actual container env carries the `THV_SESSION_REDIS_PASSWORD` entry while the expected env doesn't, so `equality.Semantic.DeepEqual` flags perpetual drift and reconcile keeps rewriting the deployment.

For this PR, your positioning is correct for the common (non-Redis) case — both paths produce the same sequence and DeepEqual passes. So no action needed here. But:

1. Consider softening the comment to say "before embedded auth" only (since the Redis claim doesn't hold for this code path), and
2. Worth filing a follow-up issue for the Redis-password mirror so the drift detector picks it up — happy to file it if you want.

ChrisJBurns (commented):
Nice surgical fix to #5360 — the diagnosis (live-mounted ConfigMap defeating the strict-greater-than gate) is sharp, and the symmetry with how the image is already frozen via the CLI positional arg is the right framing.

The testing pyramid is excellent: lock-in regression (`TestDeployWorkload_EqualGenerationDifferentImageClobbers`) pinning what the gate cannot defend against, a TDD red→green for the override (`TestTryLoadConfigFromFile_MCPServerGenerationEnvOverride`), the downward-API shape unit test, and the envtest integration test covering both initial reconcile and spec-bump drift. The `defaultAnnotations`-takes-precedence merge guarantees user overrides can't accidentally clobber the new annotation.

A couple of small points inline — one defensive-validation nit on the proxyrunner override, and a heads-up that the position-mirror comment in `deploymentNeedsUpdate` papers over a pre-existing Redis-password drift bug that's worth tracking separately.

cmd/thv-operator/controllers/mcpserver_resource_overrides_test.go
```diff
@@ -252,7 +252,8 @@ func TestResourceOverrides(t *testing.T) {
 			},
 			expectedDeploymentAnns: map[string]string{},
 			expectedPodTemplateAnns: map[string]string{
-				"toolhive.stacklok.dev/runconfig-checksum": "test-checksum",
+				"toolhive.stacklok.dev/runconfig-checksum":   "test-checksum",
+				"toolhive.stacklok.dev/mcpserver-generation": "0",
```
ChrisJBurns, line 263: Minor: the `"0"` value here comes from the fake client not auto-incrementing `metadata.generation` on Create (the real API server starts at 1). The integration test handles this correctly by reading `fresh.Generation` after Create. For these unit-level table cases it's fine to keep `"0"`, but consider adding a one-line comment so a future reader doesn't read it as "the operator stamps zero" — e.g. `// fake client leaves Generation at 0; in production this mirrors metadata.generation`. Non-blocking.

cmd/thv-proxyrunner/app/run.go
ChrisJBurns: Good catch on the `gen < 0` guard — applied in commit d2121daf6. The branch added:
- Reject negative values with a Warn log and fall through to file (defensive validation per the cited rule).
- Debug log on the success path showing both the file value and the env-sourced override, so the override is observable in pod logs.
- New table test `TestApplyMCPServerGenerationOverride` covering empty / valid / zero / unparseable / negative branches in isolation.

The `0` backward-compat case is unchanged: an unstamped pod annotation downward-API-projects to an empty string, which hits the `if raw == ""` early return and leaves the file value intact.

cmd/thv-operator/controllers/mcpserver_controller.go
ChrisJBurns: Fair point. Softened the comment in d2121daf6 to drop the Redis reference — now just says "must come before the embedded-auth env vars below". The Redis-mirror gap is real; filed as #5365.

cmd/thv-operator/controllers/mcpserver_resource_overrides_test.go
ChrisJBurns: Added a header comment to `TestResourceOverrides` in d2121daf6 explaining the fake-client Generation behavior and pointing future readers at the envtest for realistic generation tracking.

JAORMX (approved):

PR #5372, opened 2026-05-22
Consolidate authserver DCR types onto pkg/oauthproto

### LINKED ISSUE #5371: Consolidate authserver DCR types onto pkg/oauthproto
## Description

Consolidate the parallel DCR (RFC 7591) request/response type definitions in the authserver package onto the shared types in `pkg/oauthproto`. Today `pkg/authserver/server/registration/dcr.go` defines its own `DCRRequest` / `DCRResponse` with `Scope string`, while `pkg/oauthproto/dcr.go` defines `DynamicClientRegistrationRequest` / `DynamicClientRegistrationResponse` with `Scope ScopeList`. The duplication is a leftover from #5036 and was explicitly deferred there because reconciling the `Scope` wire representation is non-trivial.

This is the cleanup that #5036 left behind.

## Context

[#5036](https://github.com/stacklok/toolhive/issues/5036) introduced `pkg/oauthproto` as the canonical RFC 7591 / RFC 8414 protocol package and moved the upstream DCR client into it. The authserver-side DCR handler (`/oauth/register`) was not migrated at the time. The follow-up [comment on #4977](https://github.com/stacklok/toolhive/issues/4977#issuecomment-4327137053) called out the remaining work:

> the authserver still defines its own DCRRequest/DCRResponse in pkg/authserver/server/registration/dcr.go alongside the new types in pkg/oauthproto. Reconciling Scope string vs ScopeList is the blocking work, so it doesn't belong in #5036.

That status still holds on `main` as of 2026-05-22.

The duplication has two costs:

1. **Drift risk.** Two parallel structs that model the same RFC 7591 surface will diverge over time — exactly the anti-pattern called out in `.claude/rules/go-style.md` ("Avoid parallel types that drift").
2. **Format mismatch on the wire.** `ScopeList.UnmarshalJSON` already handles the RFC 7591 ambiguity (scope can arrive as a space-delimited string *or* a JSON array). The authserver's `Scope string` only handles the space-delimited form, so well-formed RFC 7591 clients that send `"scope": ["a", "b"]` get rejected by the authserver but accepted by the upstream-facing DCR client. Two implementations in one repo behave differently against the same spec.

## Current state

Authserver types — [pkg/authserver/server/registration/dcr.go:83-146](pkg/authserver/server/registration/dcr.go#L83-L146):

```go
type DCRRequest struct {
    RedirectURIs            []string `json:"redirect_uris"`
    ClientName              string   `json:"client_name,omitempty"`
    TokenEndpointAuthMethod string   `json:"token_endpoint_auth_method,omitempty"`
    GrantTypes              []string `json:"grant_types,omitempty"`
    ResponseTypes           []string `json:"response_types,omitempty"`
    Scope                   string   `json:"scope,omitempty"`        // space-delimited only
    SoftwareID              string   `json:"software_id,omitempty"`  // not on oauthproto type
}

type DCRResponse struct {
    ClientID                string   `json:"client_id"`
    ClientIDIssuedAt        int64    `json:"client_id_issued_at,omitempty"`
    RedirectURIs            []string `json:"redirect_uris"`
    ClientName              string   `json:"client_name,omitempty"`
    TokenEndpointAuthMethod string   `json:"token_endpoint_auth_method"`
    GrantTypes              []string `json:"grant_types"`
    ResponseTypes           []string `json:"response_types"`
    Scope                   string   `json:"scope,omitempty"`        // space-delimited only
}
```

Shared types — [pkg/oauthproto/dcr.go:22-149](pkg/oauthproto/dcr.go#L22-L149):

```go
type DynamicClientRegistrationRequest struct {
    RedirectURIs            []string  `json:"redirect_uris"`
    ClientName              string    `json:"client_name,omitempty"`
    TokenEndpointAuthMethod string    `json:"token_endpoint_auth_method,omitempty"`
    GrantTypes              []string  `json:"grant_types,omitempty"`
    ResponseTypes           []string  `json:"response_types,omitempty"`
    Scopes                  ScopeList `json:"scope,omitempty"`        // []string + dual-format codec
    // (no SoftwareID)
}
```

`ScopeList` ([pkg/oauthproto/dcr.go:49-110](pkg/oauthproto/dcr.go#L49-L110)) already implements the dual-format `UnmarshalJSON` (string or array) and a space-delimited `MarshalJSON` for wire compatibility.

Consumers of the authserver types that need to be updated:

- [pkg/authserver/server/handlers/dcr.go:44,160](pkg/authserver/server/handlers/dcr.go#L44) — decodes the request and builds the response.
- [pkg/authserver/server/handlers/dcr_test.go](pkg/authserver/server/handlers/dcr_test.go) — eight test sites referencing `registration.DCRRequest` / `registration.DCRResponse`.

The validation surface that operates on the local types — `ValidateDCRRequest`, `ValidateScopes`, `FormatScopes` — also needs to flip from `string` to `[]string` (or to take `ScopeList` directly).

## Proposed fix

1. **Add `SoftwareID` to `oauthproto.DynamicClientRegistrationRequest`.** It is an RFC 7591 §2 field and the only structural gap between the two request types. This keeps `pkg/oauthproto` as the single source of truth for the RFC 7591 wire shape.
2. **Delete `registration.DCRRequest` and `registration.DCRResponse`.** Replace every reference with the `oauthproto` equivalents.
3. **Change `registration.ValidateDCRRequest`** to accept `*oauthproto.DynamicClientRegistrationRequest` and return the same type. The validator stays in the authserver package — it encodes server-side policy (loopback-only public clients, `software_id` length cap, `token_endpoint_auth_method == "none"`) that does not belong in the protocol package.
4. **Change `registration.ValidateScopes`** signature from `(requestedScope string, allowedScopes []string)` to `(requestedScopes []string, allowedScopes []string)`. Caller in [pkg/authserver/server/handlers/dcr.go:61](pkg/authserver/server/handlers/dcr.go#L61) passes `dcrReq.Scopes` (a `ScopeList`, which is `[]string`) directly.
5. **Delete `registration.FormatScopes`.** The handler currently calls it at [pkg/authserver/server/handlers/dcr.go:168](pkg/authserver/server/handlers/dcr.go#L168) to produce the space-delimited response scope. That formatting now lives in `ScopeList.MarshalJSON`, so the handler just assigns `Scopes: oauthproto.ScopeList(scopes)`.
6. **Keep `registration.DCRError`** where it is. It is a server-side response shape, not a request/response duplicated in `pkg/oauthproto`.
7. **Update tests in `pkg/authserver/server/handlers/dcr_test.go`** to construct `oauthproto.DynamicClientRegistrationRequest` literals and decode into `oauthproto.DynamicClientRegistrationResponse`. Add at least one positive-path test that submits scope as a JSON array (`"scope": ["a", "b"]`) to lock in the array-form wire compatibility that this change unlocks.

## Acceptance Criteria

- [ ] `oauthproto.DynamicClientRegistrationRequest` gains a `SoftwareID string` field with the `software_id,omitempty` JSON tag.
- [ ] `registration.DCRRequest` and `registration.DCRResponse` are removed from `pkg/authserver/server/registration/dcr.go`.
- [ ] `grep -rn "registration.DCRRequest\|registration.DCRResponse"` returns no hits across the repository.
- [ ] `registration.ValidateDCRRequest` takes `*oauthproto.DynamicClientRegistrationRequest` and returns `(*oauthproto.DynamicClientRegistrationRequest, *DCRError)`.
- [ ] `registration.ValidateScopes` takes `(requestedScopes []string, allowedScopes []string)`.
- [ ] `registration.FormatScopes` is deleted; the handler builds the response scope as `oauthproto.ScopeList(scopes)`.
- [ ] A new positive-path test in `pkg/authserver/server/handlers/dcr_test.go` submits `"scope": ["openid", "offline_access"]` (JSON array form) and observes a 201 response with the registered scope set. This proves the format-mismatch bug described above is closed.
- [ ] All existing space-delimited scope tests continue to pass unchanged (i.e., `ScopeList.UnmarshalJSON`'s string-form branch is exercised by the legacy test bodies).
- [ ] `task build`, `task test`, `task lint-fix`, and `task license-check` are clean.

## Out of Scope

- Promoting `registration.DCRError` into `pkg/oauthproto`. It only models the server-emitted error response and has no upstream-client counterpart today.
- Restructuring `pkg/authserver/server/registration/` more broadly — validation helpers, the fosite client factory, and storage glue stay where they are.
- Any changes to the resolver / upstream DCR flow in `pkg/auth/dcr/`.

## References

- Parent: #4977
- Predecessor that left this gap: #5036
- `.claude/rules/go-style.md` — "Avoid parallel types that drift" (Interface Design section).
- RFC 7591 §2 (`software_id`), §3.1 (request format — scope ambiguity), §3.2.1 (response format).

tgrunnagle:
## Summary

- The authserver carried its own `DCRRequest` / `DCRResponse` types alongside the canonical `pkg/oauthproto` ones, a leftover from #5036 that was explicitly deferred there. Two parallel structs modelling the same RFC 7591 surface drift over time — the exact anti-pattern flagged in `.claude/rules/go-style.md` — and they had already diverged in behavior.
- Concretely, the authserver's `Scope string` only accepted the space-delimited wire form, while `oauthproto.ScopeList` accepts both the string and JSON-array forms permitted by RFC 7591 §2. Well-formed clients sending `"scope": ["a", "b"]` to `/oauth/register` were being rejected by the authserver but accepted by the upstream-facing DCR client in the same binary.
- This PR removes the duplication: `registration.DCRRequest`, `registration.DCRResponse`, and `registration.FormatScopes` are deleted; the handler and validators now operate on `oauthproto.DynamicClientRegistrationRequest` / `Response`; and `oauthproto.DynamicClientRegistrationRequest` gains the `SoftwareID` field that was the only structural gap between the two request types.

Closes #5371

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

Note: this is a behavior-preserving refactor for existing clients (space-delimited scope still works), but it does fix a latent format-mismatch where the array form of `scope` was incorrectly rejected. New test coverage locks the array-form path in.

## Test plan

All existing space-delimited scope tests in `pkg/authserver/server/handlers/dcr_test.go` and `pkg/authserver/server/registration/dcr_test.go` continue to pass unchanged, exercising `ScopeList.UnmarshalJSON`'s string-form branch. A new positive-path test, `TestRegisterClientHandler_ScopeAsJSONArray`, submits `"scope": ["openid", "offline_access"]` and asserts a 201 with the granted scopes echoed in the response and propagated to storage.

## Changes

| File | Change |
|------|--------|
| `pkg/oauthproto/dcr.go` | Add `SoftwareID string` field with `software_id,omitempty` JSON tag to `DynamicClientRegistrationRequest` (RFC 7591 §2). |
| `pkg/authserver/server/registration/dcr.go` | Delete `DCRRequest`, `DCRResponse`, and `FormatScopes`. Re-type `ValidateDCRRequest` to take/return `*oauthproto.DynamicClientRegistrationRequest`. Re-type `ValidateScopes` from `(requestedScope string, allowedScopes []string)` to `(requestedScopes, allowedScopes []string)` — `ScopeList` already does the space-delimited / array decode upstream. `DCRError` and the server-side validation policy stay put. |
| `pkg/authserver/server/handlers/dcr.go` | Decode into `oauthproto.DynamicClientRegistrationRequest`, build an `oauthproto.DynamicClientRegistrationResponse`, and assign `Scopes: oauthproto.ScopeList(scopes)` instead of calling the removed `FormatScopes` (the space-delimited wire form is now produced by `ScopeList.MarshalJSON`). |
| `pkg/authserver/server/handlers/dcr_test.go` | Migrate all eight call sites to the `oauthproto` types and add `TestRegisterClientHandler_ScopeAsJSONArray` covering the array-form scope path. |
| `pkg/authserver/server/registration/dcr_test.go` | Migrate validator tests to the `oauthproto` types and the new `[]string` scope signature. |

Yes. `/oauth/register` now accepts the RFC 7591 array form of `scope` (e.g. `"scope": ["openid", "offline_access"]`) in addition to the space-delimited string form it already accepted. No client that was working before needs to change. `software_id` is now formally part of the shared `oauthproto` request type, so it is available to any future consumer of that type, not just the authserver handler.

## Implementation plan

<details>

From issue #5371 — "Proposed fix":

1. Add `SoftwareID` to `oauthproto.DynamicClientRegistrationRequest` (RFC 7591 §2 field; only structural gap between the two request types).
2. Delete `registration.DCRRequest` and `registration.DCRResponse`. Replace every reference with the `oauthproto` equivalents.
3. Change `registration.ValidateDCRRequest` to accept `*oauthproto.DynamicClientRegistrationRequest` and return the same type. The validator stays in the authserver package — it encodes server-side policy (loopback-only public clients, `software_id` length cap, `token_endpoint_auth_method == "none"`) that does not belong in the protocol package.
4. Change `registration.ValidateScopes` signature from `(requestedScope string, allowedScopes []string)` to `(requestedScopes []string, allowedScopes []string)`. Caller in `pkg/authserver/server/handlers/dcr.go` passes `dcrReq.Scopes` (a `ScopeList`, which is `[]string`) directly.
5. Delete `registration.FormatScopes`. The handler currently calls it to produce the space-delimited response scope. That formatting now lives in `ScopeList.MarshalJSON`, so the handler just assigns `Scopes: oauthproto.ScopeList(scopes)`.
6. Keep `registration.DCRError` where it is. It is a server-side response shape, not a request/response duplicated in `pkg/oauthproto`.
7. Update tests in `pkg/authserver/server/handlers/dcr_test.go` to construct `oauthproto.DynamicClientRegistrationRequest` literals and decode into `oauthproto.DynamicClientRegistrationResponse`. Add at least one positive-path test that submits scope as a JSON array (`"scope": ["a", "b"]`) to lock in the array-form wire compatibility that this change unlocks.

</details>

- `registration.DCRError` is intentionally kept in the authserver package. It only models the server-emitted error response and has no upstream-client counterpart in `pkg/oauthproto` today; promoting it is called out as out-of-scope in the issue.
- The validator's returned `DynamicClientRegistrationRequest` deliberately does not carry the requested scopes — scope validation is the separate `ValidateScopes` step, which takes the caller's policy inputs (`ScopesSupported`, `BaselineClientScopes`). The new doc comment on `ValidateDCRRequest` makes this explicit.
- `ScopeList` preserves duplicates from the wire; `ValidateScopes` deduplicates per RFC 6749 §3.3 set semantics. This matches the previous behavior of the space-delimited path.
- No changes to the resolver / upstream DCR flow in `pkg/auth/dcr/` and no changes outside `pkg/authserver/server/{handlers,registration}` and `pkg/oauthproto/dcr.go`.

### DIFF
--- a/pkg/authserver/server/handlers/dcr.go
+++ b/pkg/authserver/server/handlers/dcr.go
@@ -14,6 +14,7 @@ import (
 	"github.com/google/uuid"
 
 	"github.com/stacklok/toolhive/pkg/authserver/server/registration"
+	"github.com/stacklok/toolhive/pkg/oauthproto"
 )
 
 // maxDCRBodySize is the maximum allowed size for DCR request bodies (64KB).
@@ -40,8 +41,10 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
 		return
 	}
 
-	// Parse request body
-	var dcrReq registration.DCRRequest
+	// Parse request body. oauthproto.ScopeList.UnmarshalJSON handles both
+	// RFC 7591 wire formats for "scope" (space-delimited string or JSON
+	// array) so we accept either shape transparently here.
+	var dcrReq oauthproto.DynamicClientRegistrationRequest
 	if err := json.NewDecoder(req.Body).Decode(&dcrReq); err != nil {
 		writeDCRError(w, http.StatusBadRequest, &registration.DCRError{
 			Error:            registration.DCRErrorInvalidClientMetadata,
@@ -58,7 +61,7 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
 	}
 
 	// Validate requested scopes against server's supported scopes
-	scopes, dcrErr := registration.ValidateScopes(dcrReq.Scope, h.config.ScopesSupported)
+	scopes, dcrErr := registration.ValidateScopes(dcrReq.Scopes, h.config.ScopesSupported)
 	if dcrErr != nil {
 		writeDCRError(w, http.StatusBadRequest, dcrErr)
 		return
@@ -151,13 +154,14 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
 	slog.Debug("registered new DCR client", logAttrs...)
 
 	// Build response per RFC 7591 Section 3.2.1.
-	// Scope reflects the scopes actually granted to this client: the
+	// Scopes reflects the scopes actually granted to this client: the
 	// client-supplied scope set was validated against ScopesSupported by
 	// ValidateScopes above, then (if configured) unioned with
 	// BaselineClientScopes — which is itself guaranteed by startup-time
 	// validation to be a subset of ScopesSupported. The unioned set is NOT
-	// re-validated here.
-	response := registration.DCRResponse{
+	// re-validated here. ScopeList.MarshalJSON emits the RFC 7591 §2
+	// space-delimited wire form on the way out.
+	response := oauthproto.DynamicClientRegistrationResponse{
 		ClientID:                clientID,
 		ClientIDIssuedAt:        time.Now().Unix(),
 		RedirectURIs:            validated.RedirectURIs,
@@ -165,7 +169,7 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
 		TokenEndpointAuthMethod: validated.TokenEndpointAuthMethod,
 		GrantTypes:              validated.GrantTypes,
 		ResponseTypes:           validated.ResponseTypes,
-		Scope:                   registration.FormatScopes(scopes),
+		Scopes:                  oauthproto.ScopeList(scopes),
 	}
 
 	w.Header().Set("Content-Type", "application/json")
--- a/pkg/authserver/server/handlers/dcr_test.go
+++ b/pkg/authserver/server/handlers/dcr_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/stacklok/toolhive/pkg/authserver/server"
 	"github.com/stacklok/toolhive/pkg/authserver/server/registration"
 	"github.com/stacklok/toolhive/pkg/authserver/storage/mocks"
+	"github.com/stacklok/toolhive/pkg/oauthproto"
 )
 
 func TestRegisterClientHandler(t *testing.T) {
@@ -36,7 +37,7 @@ func TestRegisterClientHandler(t *testing.T) {
 	}{
 		{
 			name: "success",
-			requestBody: registration.DCRRequest{
+			requestBody: oauthproto.DynamicClientRegistrationRequest{
 				RedirectURIs: []string{"http://127.0.0.1:8080/callback"},
 				ClientName:   "Test Client",
 			},
@@ -50,7 +51,7 @@ func TestRegisterClientHandler(t *testing.T) {
 		},
 		{
 			name: "validation error propagated",
-			requestBody: registration.DCRRequest{
+			requestBody: oauthproto.DynamicClientRegistrationRequest{
 				RedirectURIs: []string{"http://example.com/callback"},
 			},
 			expectedStatus: http.StatusBadRequest,
@@ -58,7 +59,7 @@ func TestRegisterClientHandler(t *testing.T) {
 		},
 		{
 			name: "storage failure returns 500",
-			requestBody: registration.DCRRequest{
+			requestBody: oauthproto.DynamicClientRegistrationRequest{
 				RedirectURIs: []string{"http://127.0.0.1:8080/callback"},
 			},
 			storageErr:      errors.New("disk full"),
@@ -111,7 +112,7 @@ func TestRegisterClientHandler(t *testing.T) {
 					assert.Contains(t, errResp.ErrorDescription, tc.expectedErrDesc)
 				}
 			} else {
-				var resp registration.DCRResponse
+				var resp oauthproto.DynamicClientRegistrationResponse
 				require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
 				assert.NotEmpty(t, resp.ClientID)
 				assert.NotZero(t, resp.ClientIDIssuedAt)
@@ -137,7 +138,7 @@ func TestRegisterClientHandler_ScopeInResponse(t *testing.T) {
 		},
 	}
 
-	reqBody, err := json.Marshal(registration.DCRRequest{
+	reqBody, err := json.Marshal(oauthproto.DynamicClientRegistrationRequest{
 		RedirectURIs: []string{"http://127.0.0.1:8080/callback"},
 	})
 	require.NoError(t, err)
@@ -149,9 +150,9 @@ func TestRegisterClientHandler_ScopeInResponse(t *testing.T) {
 	handler.RegisterClientHandler(w, req)
 	require.Equal(t, http.StatusCreated, w.Code)
 
-	var resp registration.DCRResponse
+	var resp oauthproto.DynamicClientRegistrationResponse
 	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
-	assert.Equal(t, registration.FormatScopes(registration.DefaultScopes), resp.Scope,
+	assert.Equal(t, []string(registration.DefaultScopes), []string(resp.Scopes),
 		"DCR response should include granted scopes per RFC 7591 Section 3.2.1")
 }
 
@@ -239,9 +240,9 @@ func TestRegisterClientHandler_BaselineClientScopes(t *testing.T) {
 			}
 			handler := &Handler{storage: stor, config: cfg}
 
-			reqBody, err := json.Marshal(registration.DCRRequest{
+			reqBody, err := json.Marshal(oauthproto.DynamicClientRegistrationRequest{
 				RedirectURIs: []string{"http://127.0.0.1:8080/callback"},
-				Scope:        tc.requestScope,
+				Scopes:       oauthproto.ScopeList(strings.Fields(tc.requestScope)),
 			})
 			require.NoError(t, err)
 
@@ -253,9 +254,9 @@ func TestRegisterClientHandler_BaselineClientScopes(t *testing.T) {
 
 			require.Equal(t, http.StatusCreated, w.Code)
 
-			var resp registration.DCRResponse
+			var resp oauthproto.DynamicClientRegistrationResponse
 			require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
-			assert.Equal(t, registration.FormatScopes(tc.expectedScopes), resp.Scope,
+			assert.Equal(t, tc.expectedScopes, []string(resp.Scopes),
 				"DCR response scope must equal the union of requested and baseline scopes")
 
 			require.NotNil(t, capturedClient, "storage was not called")
@@ -285,7 +286,7 @@ func TestRegisterClientHandler_ClientIsStored(t *testing.T) {
 	}
 	handler := &Handler{storage: stor, config: cfg}
 
-	reqBody, err := json.Marshal(registration.DCRRequest{
+	reqBody, err := json.Marshal(oauthproto.DynamicClientRegistrationRequest{
 		RedirectURIs: []string{"http://127.0.0.1:8080/callback"},
 		ClientName:   "Stored Client",
 	})
@@ -298,7 +299,7 @@ func TestRegisterClientHandler_ClientIsStored(t *testing.T) {
 	handler.RegisterClientHandler(w, req)
 	require.Equal(t, http.StatusCreated, w.Code)
 
-	var resp registration.DCRResponse
+	var resp oauthproto.DynamicClientRegistrationResponse
 	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
 
 	require.NotNil(t, storedClient)
@@ -310,4 +311,51 @@ func TestRegisterClientHandler_ClientIsStored(t *testing.T) {
 	assert.Equal(t, []string{"http://127.0.0.1:8080/callback"}, loopbackClient.GetRedirectURIs())
 	assert.Equal(t, fosite.Arguments(allowedAudiences), loopbackClient.GetAudience(),
 		"DCR client must inherit server's AllowedAudiences so refresh token requests with resource= succeed")
+}
+
+// TestRegisterClientHandler_ScopeAsJSONArray verifies that the /oauth/register
+// endpoint accepts the RFC 7591 array form of "scope". Prior to consolidating
+// onto oauthproto.ScopeList, the handler only accepted space-delimited strings
+// and would reject this body as a JSON decode error.
+func TestRegisterClientHandler_ScopeAsJSONArray(t *testing.T) {
+	t.Parallel()
+
+	ctrl := gomock.NewController(t)
+	stor := mocks.NewMockStorage(ctrl)
+	var capturedClient fosite.Client
+	stor.EXPECT().RegisterClient(gomock.Any(), gomock.Any()).DoAndReturn(
+		func(_ context.Context, c fosite.Client) error {
+			capturedClient = c
+			return nil
+		})
+
+	handler := &Handler{
+		storage: stor,
+		config: &server.AuthorizationServerConfig{
+			Config:          &fosite.Config{AccessTokenIssuer: "https://test-authserver"},
+			ScopesSupported: registration.DefaultScopes,
+		},
+	}
+
+	// Raw JSON body with scope as a JSON array — the form that
+	// well-formed RFC 7591 clients can send and that the previous
+	// space-delimited-only authserver implementation rejected.
+	body := []byte(`{"redirect_uris":["http://127.0.0.1:8080/callback"],"scope":["openid","offline_access"]}`)
+	req := httptest.NewRequest(http.MethodPost, "/oauth/register", bytes.NewReader(body))
+	req.Header.Set("Content-Type", "application/json")
+	w := httptest.NewRecorder()
+
+	handler.RegisterClientHandler(w, req)
+
+	require.Equal(t, http.StatusCreated, w.Code,
+		"array-form scope must be accepted per RFC 7591 §2 ambiguity")
+
+	var resp oauthproto.DynamicClientRegistrationResponse
+	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
+	assert.Equal(t, []string{"openid", "offline_access"}, []string(resp.Scopes),
+		"granted scopes must reflect the array-form request")
+
+	require.NotNil(t, capturedClient, "storage was not called")
+	assert.Equal(t, fosite.Arguments([]string{"openid", "offline_access"}), capturedClient.GetScopes(),
+		"the array-form scopes must reach storage")
 }
--- a/pkg/authserver/server/handlers/dcr_test.go
+++ b/pkg/authserver/server/handlers/dcr_test.go
@@ -165,7 +165,7 @@ func TestRegisterClientHandler_BaselineClientScopes(t *testing.T) {
 
 	tests := []struct {
 		name                 string
-		requestScope         string
+		requestScopes        []string
 		baselineClientScopes []string
 		extraScopesSupported []string // appended to DefaultScopes in ScopesSupported
 		expectedScopes       []string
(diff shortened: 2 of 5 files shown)


tgrunnagle (commented):
## Multi-agent code review — PR #5372

This is a 4-agent specialist review (OAuth/RFC 7591, Go style/refactor, Test coverage, General quality) of the consolidation of authserver DCR types onto `pkg/oauthproto`. Codex cross-review was skipped (CLI not installed).

The refactor is correct and well-scoped — every acceptance criterion in #5371 is satisfied and the parallel-types-that-drift anti-pattern called out in `.claude/rules/go-style.md` is fully resolved. Observable behavior on the happy path (string-form scope in, space-delimited string out) is preserved.

Two cross-reviewer-confirmed concerns (both MEDIUM, both surgical fixes) warrant attention before merge.

### Findings summary

| # | Severity | Title |
|---|---|---|
| 1 | MEDIUM | Response fields `redirect_uris`/`token_endpoint_auth_method`/`grant_types`/`response_types` silently gained `omitempty` |
| 2 | MEDIUM | Migrated scope assertions no longer pin the RFC 7591 §3.2.1 wire form |
| 3 | LOW | `ValidateScopes` doc comment couples to one caller and undersells the function's contract |
| 4 | LOW | No array-form negative test; empty-array (`"scope": []`) wire-behavior change is untested |
| 5 | INFO | Test helper retains `requestScope string` behind a `strings.Fields` wrapper |

### What this PR does well

- Fully closes the deferred work from #5036; `grep` against the post-PR tree is clean.
- New comments on `ValidateDCRRequest` and `ValidateScopes` make the validator/scope-validator split explicit — load-bearing context that was previously implicit.
- The new `TestRegisterClientHandler_ScopeAsJSONArray` posts a literal `[]byte` body (not a Go literal that goes through marshal), so it actually exercises the dual-format `UnmarshalJSON` path end-to-end.
- Loopback-only redirect policy, `software_id` length/charset validation, and audit log fields are byte-identical to before.

### Suggested PR description tweak

The "Does this introduce a user-facing change?" section documents the new array-form acceptance well. Worth adding one line noting (a) the empty-array case (`"scope": []`) also now succeeds with defaults rather than returning HTTP 400, and (b) the response shape for `redirect_uris`/`token_endpoint_auth_method`/`grant_types`/`response_types` changed at the type-system level from "always present" to "present iff non-zero" (no observable change in the happy path, but the safety net is weaker).

Recommended event: `COMMENT`.

---

pkg/authserver/server/handlers/dcr.go
```diff
@@ -151,21 +154,22 @@ func (h *Handler) RegisterClientHandler(w http.ResponseWriter, req *http.Request
...
+	// Scopes reflects the scopes actually granted to this client: the
 	// client-supplied scope set was validated against ScopesSupported by
 	// ValidateScopes above, then (if configured) unioned with
 	// BaselineClientScopes — which is itself guaranteed by startup-time
 	// validation to be a subset of ScopesSupported. The unioned set is NOT
-	// re-validated here.
-	response := registration.DCRResponse{
+	// re-validated here. ScopeList.MarshalJSON emits the RFC 7591 §2
+	// space-delimited wire form on the way out.
+	response := oauthproto.DynamicClientRegistrationResponse{
```
tgrunnagle, line 164: **MEDIUM — 4 response fields silently gained `omitempty` after consolidation**

The old `registration.DCRResponse` declared `redirect_uris`, `token_endpoint_auth_method`, `grant_types`, and `response_types` *without* `omitempty`. The new `oauthproto.DynamicClientRegistrationResponse` (see `pkg/oauthproto/dcr.go:127-130`) declares all four *with* `omitempty`. RFC 7591 §3.2.1 effectively requires these to be echoed.

In current code the handler always populates them (validator fills defaults), so the wire shape is unchanged. But the struct-level safety net is gone — a future code path that constructs the response without going through `ValidateDCRRequest` (e.g., a partial response in a new error path, or a new endpoint reusing the type) could silently drop a required key.

**Suggestion**: either drop `omitempty` on those four tags in `pkg/oauthproto/dcr.go:127-130` (recommended — keeps the AS-emitter safe by construction), or extend the response type's doc comment to state that callers MUST populate these fields per RFC 7591 §3.2.1 and point to `ValidateDCRRequest` as the canonical populator.

_(Raised independently by 3 of the 4 specialist reviewers.)_

pkg/authserver/server/handlers/dcr_test.go
```diff
@@ -311,3 +312,50 @@ func TestRegisterClientHandler_ClientIsStored(t *testing.T) {
...
+	req := httptest.NewRequest(http.MethodPost, "/oauth/register", bytes.NewReader(body))
+	req.Header.Set("Content-Type", "application/json")
+	w := httptest.NewRecorder()
+
+	handler.RegisterClientHandler(w, req)
+
+	require.Equal(t, http.StatusCreated, w.Code,
+		"array-form scope must be accepted per RFC 7591 §2 ambiguity")
+
+	var resp oauthproto.DynamicClientRegistrationResponse
```
tgrunnagle: **MEDIUM — this assertion can no longer distinguish a space-delimited response from a JSON-array response**

`oauthproto.DynamicClientRegistrationResponse.Scopes` is a `ScopeList`, whose `UnmarshalJSON` accepts both string and array forms (`pkg/oauthproto/dcr.go:74-110`). So `assert.Equal(t, []string{"openid", "offline_access"}, []string(resp.Scopes), ...)` passes identically whether the server emitted `"scope":"openid offline_access"` (correct, per RFC 7591 §3.2.1) or `"scope":["openid","offline_access"]` (a regression).

The same fidelity loss affects `TestRegisterClientHandler_ScopeInResponse` and `TestRegisterClientHandler_BaselineClientScopes` — both were byte-level string comparisons before the migration and now decode through the dual-format codec. But this is the test that exists specifically to lock in the request-array → response-string asymmetry, so the wire pin matters most here.

**Suggestion**: add a raw-body assertion next to the existing slice check:

```go
assert.Contains(t, w.Body.String(), `"scope":"openid offline_access"`,
    "RFC 7591 §3.2.1: response scope must be space-delimited string, not JSON array")
```

_(Raised independently by 3 of the 4 specialist reviewers.)_

pkg/authserver/server/registration/dcr.go
```diff
@@ -342,20 +280,25 @@ func ValidateRedirectURI(uri string) *DCRError {
 // ValidateScopes validates that all requested scopes are in the allowed set.
 // Returns the validated scopes (or defaults if empty) and any error.
 // This enforces server-side scope restrictions per RFC 7591 Section 2.
-func ValidateScopes(requestedScope string, allowedScopes []string) ([]string, *DCRError) {
+//
+// Callers pass already-parsed scope slices: oauthproto.ScopeList handles the
+// RFC 7591 dual-format (string or array) decode on the wire, so by the time
+// scopes reach this function they are a plain []string. Duplicates in the
+// input are preserved by ScopeList — this function deduplicates per RFC 6749
```
tgrunnagle: **LOW — doc comment couples `ValidateScopes` to one specific caller and undersells the function's contract**

Two issues:

1. The comment anchors the contract to `oauthproto.ScopeList`. The function's signature accepts any `[]string`; a future caller that builds the slice differently shouldn't have to read about `ScopeList` to understand the contract.
2. "Duplicates in the input are preserved by `ScopeList`" slightly oversells `ScopeList` — neither branch of `ScopeList.UnmarshalJSON` deduplicates (the string branch uses `strings.Fields` which retains tokens; the array branch only trim-and-filters empty entries). Saying "the input is treated as-is; this function deduplicates" is more accurate. Also, the function's other load-bearing behaviors (rejecting unknown scopes, defaulting to `DefaultScopes` on empty input) only appear in the one-line header — worth enumerating in the body too.

**Suggested rewording**:

```go
// ValidateScopes validates a slice of already-parsed scope tokens against
// the server's allowed set per RFC 7591 §2.
//
//   - Empty/nil input falls back to DefaultScopes (which must itself be a
//     subset of allowedScopes; otherwise the call returns an error).
//   - Each requested scope must appear in allowedScopes; otherwise returns
//     invalid_client_metadata.
//   - Duplicates in the input are tolerated and deduplicated per RFC 6749
//     §3.3 (scope is a set of case-sensitive strings).
```

_(Raised independently by 3 of the 4 specialist reviewers.)_

pkg/authserver/server/handlers/dcr_test.go
```diff
@@ -311,3 +312,50 @@ func TestRegisterClientHandler_ClientIsStored(t *testing.T) {
 	assert.Equal(t, fosite.Arguments(allowedAudiences), loopbackClient.GetAudience(),
 		"DCR client must inherit server's AllowedAudiences so refresh token requests with resource= succeed")
 }
+
+// TestRegisterClientHandler_ScopeAsJSONArray verifies that the /oauth/register
+// endpoint accepts the RFC 7591 array form of "scope". Prior to consolidating
+// onto oauthproto.ScopeList, the handler only accepted space-delimited strings
```
tgrunnagle, line 318: **LOW — array-form negative path and empty-array case are untested at the handler boundary**

The new positive-path test is great. Two adjacent paths through the same array-form decoder are untested at the handler:

1. **Bad scope in array form**: `"scope":["openid","sneaky_admin"]` should hit `ValidateScopes` and return 400 `invalid_client_metadata`. Today only string-form bad scopes are tested through the handler.
2. **Empty array**: `"scope": []` previously returned HTTP 400 (the `Scope string` field couldn't decode an array). It now silently succeeds with `DefaultScopes`. This is an intentional behavior change per the RFC 7591 ambiguity fix but has no regression test.

**Suggestion**: convert this function into a small table covering happy path (already here), bad-scope-in-array (expect 400), and empty-array (expect 201 with defaults). Three rows; ~30 lines.

_(Raised independently by the OAuth/RFC 7591 and Test-coverage specialists.)_

pkg/authserver/server/handlers/dcr_test.go
```diff
@@ -239,9 +240,9 @@ func TestRegisterClientHandler_BaselineClientScopes(t *testing.T) {
 			}
 			handler := &Handler{storage: stor, config: cfg}
 
-			reqBody, err := json.Marshal(registration.DCRRequest{
+			reqBody, err := json.Marshal(oauthproto.DynamicClientRegistrationRequest{
 				RedirectURIs: []string{"http://127.0.0.1:8080/callback"},
-				Scope:        tc.requestScope,
+				Scopes:       oauthproto.ScopeList(strings.Fields(tc.requestScope)),
```
tgrunnagle: **INFO — test helper still threads a string through `strings.Fields` even though the canonical type is `[]string`**

Now that the production type is `[]string`, the table column could be `requestScopes []string` directly and the call site could drop `strings.Fields`. Currently the test exclusively exercises the string-form marshaler path (since `oauthproto.ScopeList` marshals to a space-delimited string regardless), but the natural shape would let the test express input intent directly.

Pure code-clarity nit; defer to a follow-up if you'd rather keep this PR mechanically tight.

_(Raised independently by the Test-coverage and General-quality specialists.)_

pkg/authserver/server/handlers/dcr.go
tgrunnagle: Addressed in 0a60a737. Dropped `omitempty` on `redirect_uris`, `token_endpoint_auth_method`, `grant_types`, and `response_types` per the recommendation. The struct now enforces the RFC 7591 §3.2.1 "always echo" requirement at the type level, so any future code path that builds a response without going through `ValidateDCRRequest` will still emit those keys.

pkg/authserver/server/registration/dcr.go
tgrunnagle: Addressed in 0a60a737. Adopted the suggested doc wording — the contract is now stated independent of `ScopeList` (since the function accepts any `[]string`) and enumerates the load-bearing behaviors (`DefaultScopes` fallback, unknown-scope rejection, RFC 6749 §3.3 dedup) in the body rather than just the header.

pkg/authserver/server/handlers/dcr_test.go
tgrunnagle: Addressed in cc184260. Added the suggested `assert.Contains(t, w.Body.String(), `"scope":"openid offline_access"`, ...)` and applied the same wire-form pin to the new empty-array case. The test now catches a hypothetical regression that emits the response scope as a JSON array — `ScopeList.UnmarshalJSON` would otherwise transparently re-decode it.

pkg/authserver/server/handlers/dcr_test.go
tgrunnagle: Addressed in cc184260. Converted `TestRegisterClientHandler_ScopeAsJSONArray` into a three-row table:

1. Happy path (existing) — `["openid","offline_access"]` → 201 + space-delimited response.
2. Array-form unsupported scope — `["openid","sneaky_admin"]` → 400 `invalid_client_metadata`.
3. Empty array — `[]` → 201 with `DefaultScopes`, documented as the intentional behavior change vs. the pre-consolidation space-delimited-only decoder.

Each row sends a raw JSON body so the dual-format `UnmarshalJSON` path is exercised end-to-end at the handler boundary.

pkg/authserver/server/handlers/dcr_test.go
tgrunnagle: Addressed in cc184260. Renamed the column to `requestScopes []string` and dropped the `strings.Fields` wrapper. The test now expresses scope intent directly as a slice — call site is just `oauthproto.ScopeList(tc.requestScopes)`.

ChrisJBurns (approved):

PR #5380, opened 2026-05-25
Fixed VirtualMCPCompositeToolDefinition printer columns output

### LINKED ISSUE #4619: VirtualMCPCompositeToolDefinition printer columns show broken output
## Summary

The VirtualMCPCompositeToolDefinition CRD defines "Steps" and "Refs" printer columns using `[*]` JSONPath wildcards with `type=integer`. The `[*]` wildcard returns all array elements, not a count. When combined with `type=integer`, kubectl cannot coerce array elements (which are objects) into integers, resulting in empty or malformed output.

**Severity**: SHOULD FIX
**Area**: CRD PrintColumns
**Breaking**: No

## Location

- `cmd/thv-operator/api/v1alpha1/virtualmcpcompositetooldefinition_types.go:100-102`

## Problem

```go
// +kubebuilder:printcolumn:name="Steps",type="integer",
//   JSONPath=".spec.steps[*]"
// +kubebuilder:printcolumn:name="Refs",type="integer",
//   JSONPath=".spec.toolRefs[*]"
```

- **Expected**: Show the count of steps and refs
- **Actual**: Empty or malformed output because `[*]` expands array elements (objects), not a count

## Impact

- `kubectl get virtualmcpcompositetooldefinitions` shows broken columns
- Users cannot see step/ref counts at a glance
- Suggests the columns were never manually tested after definition

## Recommended Fix

1. Add status fields for the counts:
   ```go
   type VirtualMCPCompositeToolDefinitionStatus struct {
       // StepCount is the number of steps in the composite tool.
       StepCount int32 `json:"stepCount,omitempty"`
       // RefCount is the number of tool references.
       RefCount  int32 `json:"refCount,omitempty"`
   }
   ```

2. Update the controller to populate these during reconciliation.

3. Update printer columns to reference the status fields:
   ```go
   // +kubebuilder:printcolumn:name="Steps",type="integer",
   //   JSONPath=".status.stepCount"
   // +kubebuilder:printcolumn:name="Refs",type="integer",
   //   JSONPath=".status.refCount"
   ```

## Related Issues

- MCPGroup Ready column misleadingly shows MCPServersChecked condition (#4617)
- References printcolumn shows raw JSON instead of useful summary (#4618)

Sanskarzz:
## Summary

- Removes the malformed `Steps` and `Refs` printer columns from `VirtualMCPCompositeToolDefinition`.
- Avoids adding display-only status fields or a new reconciler just to maintain printer column counts.
- Regenerates the CRD manifests and API documentation so `kubectl get virtualmcpcompositetooldefinitions` no longer renders broken integer columns.

Fixes #4619

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

- [x] Generated artifacts (`task operator-generate`, `task operator-manifests`, `task crdref-gen`)
- [x] Targeted compile tests (`go test ./cmd/thv-operator/api/v1alpha1 ./cmd/thv-operator/api/v1beta1 ./cmd/thv-operator/app -run '^$' -count=1`)
- [x] Lint (`task lint`)

## API Compatibility

- [x] This PR does not break the `v1beta1` API, OR the `api-break-allowed` label is applied, and the migration guidance is described above.

## Changes

| File | Change |
|------|--------|
| `cmd/thv-operator/api/v1beta1/virtualmcpcompositetooldefinition_types.go` | Removes the broken `Steps` and `Refs` printer column markers. |
| `cmd/thv-operator/api/v1alpha1/types.go` | Removes the same deprecated-version printer column markers. |
| Generated CRD/docs output | Removes the malformed additional printer columns from published manifests and documentation. |

Yes. `kubectl get virtualmcpcompositetooldefinitions` no longer shows the broken `Steps` and `Refs` columns. The remaining columns continue to render normally.

Kubernetes CRD printer column JSONPath cannot compute array lengths. A count-based implementation would require scalar status fields plus reconciliation logic to keep them current. Per review feedback, this PR uses the lower-risk fix and removes the malformed display columns instead.


ChrisJBurns:
@Sanskarzz  Looking at this, do we need another controller just for a printer column? It seems a bit overkill?

Sanskarzz:
@ChrisJBurns  The reason I added the small reconciler is that the Kubernetes printer column JSONPath cannot compute array lengths. The existing `.spec.steps[*]` / `.status.referencingVirtualServers[*]` paths expand array contents, so to display integer counts reliably, we need scalar fields such as `status.stepCount` and `status.refCount`.

Because `VirtualMCPCompositeToolDefinition` did not already have a controller maintaining its status, this reconciler is what keeps those fields accurate when the definition changes or when VirtualMCPServers add/remove `compositeToolRefs`.

That said, I agree it is more machinery than ideal for a display-only issue, but the approach was recommended in the issue description. If you prefer the lower-scope fix, I can simplify this PR to remove the broken `Steps`/`Refs` columns instead of maintaining count fields.

ChrisJBurns:
@Sanskarzz  Yeah I think removing them is probably better for us. I thought the change would have been possible without another Reconciler, but if not, I'm happy for you to remove the malformed columns instead

Sanskarzz:
Sounds good, thanks. I’ve removed the malformed `Steps` and `Refs` printer columns instead of adding status count fields and a reconciler. Also updated PR description.

ChrisJBurns:
Thanks @Sanskarzz , I'll get @jerm-dro to take a look at this internally as vMCP is his wheelhouse, hopefully we should be good to go.

ChrisJBurns (approved):

PR #5385, opened 2026-05-26
Validate CIMD scope, grant_types and response_types against AS policy

amirejaz:
## Summary

- Stacked on #5348 (wire CIMD config) — merge that first.
- Addresses two deferred review comments from tgrunnagle on PR #5348.

**C3 — CIMD scope handling now consistent with DCR:**
`ScopesSupported` is threaded into `NewCIMDStorageDecorator`. Scope validation in `fetch()` now uses `registration.ValidateScopes` — the same function as the DCR handler — so both paths enforce the same scope policy. A CIMD document declaring scopes outside `ScopesSupported` is rejected at `GetClient` time. The fallback when `scope` is omitted also respects `ScopesSupported`.

**C4 — grant_types / response_types validated at fetch time:**
`fetch()` now rejects documents declaring `grant_types` outside `[authorization_code, refresh_token]` or `response_types` outside `[code]`. Consistent with DCR which returns `invalid_client_metadata` for the same cases, and matches the existing `token_endpoint_auth_method` validation pattern already in the decorator.

## Type of change

- [x] Bug fix / security improvement

## Test plan

- [x] `go test ./pkg/authserver/...` passes
- [x] `task lint-fix` clean
- New tests: scope validation with/without `ScopesSupported`, `grant_type` rejection for unsupported values, `response_type` rejection for `token`


pkg/authserver/server/handlers/scopes_test.go
```diff
@@ -108,7 +110,7 @@ func TestUnionScopes(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			got := unionScopes(tt.req, tt.baseline)
+			got := registration.UnionScopes(tt.req, tt.baseline)
```
tgrunnagle: **F1 (MEDIUM) — `TestUnionScopes` lives in the wrong package after the move.**

This test now calls `registration.UnionScopes` across the package boundary, but the file itself stayed in `package handlers`. Meanwhile `handlers/scopes.go` was reduced to just the SPDX header and `package handlers` (no contents). This violates `.claude/rules/testing.md` ("tests must only test code in the package under test") and leaves an orphaned empty source-file pair in the wrong location.

Anyone running `go test ./pkg/authserver/server/registration/...` after editing `UnionScopes` will not exercise these table-driven cases.

**Fix:** move this test (and the file, or fold into `registration/dcr_test.go`) to `pkg/authserver/server/registration/`. Delete the now-empty `pkg/authserver/server/handlers/scopes.go` and this file. Raised by 3 of 4 specialist agents (gocorr, tests, general).

tgrunnagle (commented):
## Multi-agent review

Four specialist agents (oauth/security, Go correctness/style, test quality, general/API) reviewed this PR. Codex cross-review was skipped (codex CLI not installed locally). 0 HIGH, 5 MEDIUM, 7 LOW findings survived consensus scoring. None are blocking; this is posted as `COMMENT`.

### Executive summary

A focused follow-up to stacked PR #5348 closing deferred review comments C3/C4. It threads `ScopesSupported` and `BaselineClientScopes` into the CIMD storage decorator so CIMD documents are validated against the same policy as DCR (via `registration.ValidateScopes`), and rejects unsupported `grant_types` / `response_types` at fetch time. As supporting refactor, `unionScopes` is promoted to `registration.UnionScopes`. Approach is correct in shape: reuse a single validator, reject earlier in the pipeline, surface a startup warning for the CIMD+baseline interaction. Most surviving findings are correctness-leaning details and test-quality nits.

### Pros

- Reuses `registration.ValidateScopes` instead of duplicating policy.
- Rejects invalid CIMD documents at fetch time (not at `/authorize`), giving operators a clear, early error.
- Error class change `ErrNotFound` → `ErrInvalidClient` for "fetched but policy-violating" is semantically correct per RFC 6749 §5.2.
- Public/PKCE invariant preserved (`Public: true` unconditional; allowlists still exclude `client_credentials`).
- `slog.Warn` at startup surfaces the non-obvious CIMD+baseline interaction.
- Doc comments explain the design intent and the test-only nil escape hatch.
- PR title, body, and size comply with `.claude/rules/pr-creation.md`.

### Cons

- The "same scope policy as DCR" claim has one silent divergence: when `doc.Scope == ""` and `ScopesSupported` ⊃ `DefaultScopes`, the CIMD client receives the full `ScopesSupported` while a DCR client in the same position receives only `DefaultScopes`. See F5.
- `NewCIMDStorageDecorator` now takes six positional args with two adjacent same-typed `[]string` slices — easy to swap silently.
- `TestUnionScopes` was not relocated alongside the moved function and still lives in `pkg/authserver/server/handlers/`, where `scopes.go` is now empty.
- New rejection tests assert `require.Error(...)` only — they would still pass if the code regressed to `fosite.ErrNotFound`, defeating the PR's main intent.

### Consensus findings table

| F1 | MEDIUM | handlers/scopes_test.go | `TestUnionScopes` lives in wrong package after move; pair with deleting now-empty `scopes.go` |
| F2 | MEDIUM | storage/cimd_decorator_test.go | Rejection tests don't assert `errors.Is(err, fosite.ErrInvalidClient)` |
| F3 | MEDIUM | storage/cimd_decorator_test.go | Consolidate 11 new top-level tests into ~3 table-driven tests |
| F4 | MEDIUM | storage/cimd_decorator.go | `NewCIMDStorageDecorator` signature: 6 positional args, two adjacent same-typed slices |
| F5 | MEDIUM | storage/cimd_decorator.go | Omitted-scope branch on CIMD diverges from DCR when `ScopesSupported` ⊃ `DefaultScopes` |
| F6 | LOW | storage/cimd_decorator.go | `WithHint(dcrErr.Error)` passes OAuth error code where descriptive hint belongs |
| F7 | LOW | storage/cimd_decorator.go | Constructor stores caller-supplied scope slices without defensive `slices.Clone` |
| F8 | LOW | storage/cimd_decorator.go | Doc comments document cross-file invariants not enforced in code |
| F9 | LOW | storage/cimd_decorator.go | Missing DCR-parity defense-in-depth: `response_types must include 'code'`; grant-type validation order reversed vs DCR |
| F10 | LOW | storage/cimd_decorator_test.go | Test names don't match assertions |
| F11 | LOW | storage/cimd_decorator_test.go | Redundant unit-level test; missing baseline-exceeds-supported coverage |
| F12 | LOW | authserver/server_impl.go | `slog.Warn` level/clarity for CIMD+baseline interaction |

Detailed comments are inline on the relevant changed lines.

### How to test

```
git fetch origin pull/5385/head:pr-5385 && git checkout pr-5385
task lint-fix
task test
```

Plus targeted checks against an embedded AS:
1. **C3 — declared scope outside `ScopesSupported`**: serve a CIMD doc declaring `scope: "openid profile"` with `scopes_supported: ["openid"]`. Expect `fosite.ErrInvalidClient`.
2. **C3 — omitted scope**: same setup but doc omits `scope`. Confirm the client carries exactly `["openid"]`, not `DefaultScopes` (and consider what should happen when `ScopesSupported` ⊃ `DefaultScopes` — see F5).
3. **C4 — unsupported grant_type**: serve a doc with `grant_types: ["client_credentials"]`. Expect rejection.
4. **C4 — missing `authorization_code`**: serve a doc with `grant_types: ["refresh_token"]` only. Expect rejection.
5. **C4 — unsupported response_type**: serve a doc with `response_types: ["token"]`. Expect rejection.
6. **slog.Warn**: start the AS with `cimd.enabled: true` and `baseline_client_scopes: ["offline_access"]`. Expect one WARN line at startup.
7. **Load-bearing-test check**: temporarily revert `if !allowedCIMDGrantTypes[gt]` to always-false, re-run `go test ./pkg/authserver/storage/...`. `TestFetch_RejectsUnsupportedGrantType` should fail; revert when done.

Generated with [Claude Code](https://claude.com/claude-code) (multi-agent /dev:pr-review).

pkg/authserver/storage/cimd_decorator_test.go
```diff
@@ -442,3 +445,206 @@ func TestFetch_RejectsUnsupportedTokenEndpointAuthMethod(t *testing.T) {
 	_, err := dec.fetchOrCached(context.Background(), srv.URL+"/meta.json")
 	require.Error(t, err, "fetch must fail when token_endpoint_auth_method is not \"none\"")
 }
+
+// --- grant_types / response_types validation ---
+
+func TestFetch_RejectsUnsupportedGrantType(t *testing.T) {
```
tgrunnagle: **F3 (MEDIUM) — Consolidate the new tests into a few table-driven tests.**

11 new top-level `Test*` functions added below share an almost-identical shape: `httptest.NewServer` with one customized doc field, `dec.fetchOrCached`, assert. `.claude/rules/testing.md` prefers table-driven declarative tests and explicitly calls out consolidation when small funcs exercise the same function.

Suggested shape — three tables:
- `TestFetch_GrantTypeValidation` rows: omitted (accept), `[authorization_code, refresh_token]` (accept), `[refresh_token]` only (reject), each unsupported grant (reject).
- `TestFetch_ResponseTypeValidation` rows for accepted/rejected response_types.
- `TestFetch_ScopeResolution` rows for: outside `ScopesSupported` (reject), within (accept exact set), omitted with `ScopesSupported` set, omitted with no `ScopesSupported`, with baseline unioned in, with no constraints.

Each row carries `(scopesSupported, baseline, docFields, wantErr, wantScopes)`. Reduces ~217 added LOC of repeated fixtures to one helper + a table. Optionally extend `serveCIMDDoc` to take a doc-mutator callback so each row is a one-liner.

pkg/authserver/storage/cimd_decorator_test.go
```diff
@@ -442,3 +445,206 @@ func TestFetch_RejectsUnsupportedTokenEndpointAuthMethod(t *testing.T) {
...
+					RedirectURIs: []string{"https://example.com/callback"},
+					GrantTypes:   []string{unsupported},
+				}
+				w.Header().Set("Content-Type", "application/json")
+				_ = json.NewEncoder(w).Encode(doc)
+			}))
+			t.Cleanup(srv.Close)
+			dec := newEnabledDecorator(t, newTestBase(t), 10, time.Minute)
+			_, err := dec.fetchOrCached(context.Background(), srv.URL+"/meta.json")
+			require.Error(t, err, "unsupported grant_type %q must be rejected", unsupported)
```
tgrunnagle: **F2 (MEDIUM) — Rejection tests don't assert `errors.Is(err, fosite.ErrInvalidClient)`.**

This PR's whole point is to switch CIMD rejections from `fosite.ErrNotFound` to `fosite.ErrInvalidClient` (the OAuth-spec-aligned error). But every new rejection test (here and at lines 498, 534, 580+) asserts only `require.Error(...)`. The existing `TestCIMDStorageDecorator_FetchOrCached_FetchFailureReturnsNotFound` (line 219) already demonstrates the right pattern with `assert.ErrorIs(t, err, fosite.ErrNotFound)` — the file has the idiom in place.

A regression that reverts to `ErrNotFound` (or wraps a different fosite error) would slip through every new test. The behavior this PR is changing is not actually pinned by a test.

**Suggested addition** in each new rejection test:
```go
assert.ErrorIs(t, err, fosite.ErrInvalidClient)
assert.NotErrorIs(t, err, fosite.ErrNotFound)
```

Raised by both the oauth and tests specialists.

pkg/authserver/storage/cimd_decorator_test.go
```diff
@@ -442,3 +445,206 @@ func TestFetch_RejectsUnsupportedTokenEndpointAuthMethod(t *testing.T) {
...
+		})
+	}
+}
+
+func TestFetch_AcceptsSupportedGrantTypes(t *testing.T) {
```
tgrunnagle: **F10 (LOW) — Test names don't match assertions.**

- `TestFetch_AcceptsSupportedGrantTypes` (here): the name reads as "explicit supported values are accepted," but the body uses `serveCIMDDoc(t, "/meta.json", nil)` which produces a document with `GrantTypes` **omitted**. This actually tests the default/omitted-grants case. The explicit-values case is covered by the separate `TestFetch_AcceptsExplicitSupportedGrantTypes` at line 501.
- `TestFetch_RejectsRefreshTokenOnlyGrantTypes` (line 483): exercises a distinct branch — `refresh_token` IS in `allowedCIMDGrantTypes`; the doc is rejected because `authorization_code` is **required**. The name suggests the value itself is unsupported, hiding the real assertion.

**Suggested renames** (`.claude/rules/testing.md` says "test names must match what they actually assert"):
- `TestFetch_AcceptsOmittedGrantTypes`
- `TestFetch_RejectsGrantTypesMissingAuthorizationCode`

Or absorb both as rows in the consolidated table from F3.

pkg/authserver/storage/cimd_decorator_test.go
```diff
@@ -442,3 +445,206 @@ func TestFetch_RejectsUnsupportedTokenEndpointAuthMethod(t *testing.T) {
...
+	}))
+	t.Cleanup(srv.Close)
+	dec := newEnabledDecorator(t, newTestBase(t), 10, time.Minute)
+	_, err := dec.fetchOrCached(context.Background(), srv.URL+"/meta.json")
+	require.Error(t, err, "unsupported response_type \"token\" must be rejected")
+}
+
+// --- scope validation against ScopesSupported ---
+
+func TestBuildFositeClient_ScopeDefaultsToScopesSupported(t *testing.T) {
```
tgrunnagle: **F11 (LOW) — Redundant unit-level test; also missing one coverage case.**

This test calls `buildFositeClient(doc, scopesSupported)` and asserts the returned client carries `scopesSupported` — but `buildFositeClient` doesn't *make* a decision here; it just receives the already-resolved scope list from `fetch()`. The realistic end-to-end check `TestFetch_EmptyScopeDefaultsToScopesSupported` (line 612 or thereabouts) already covers the real decision path. Per the "avoid overlapping test cases" rule, this adds no unique coverage.

The sibling test on line 552 (`TestBuildFositeClient_ScopeDefaultsToDefaultScopesWhenNoScopesSupported`) **should** be kept — it exercises the actual `if len(scopes) == 0 { scopes = slices.Clone(registration.DefaultScopes) }` fallback branch in `buildFositeClient`.

**Separately — missing coverage**: there is no test for the misconfiguration `scopesSupported=["openid"]`, `baseline=["offline_access"]`, `doc.Scope=""`. Startup config validation likely catches it earlier, but the decorator's behavior in that case is not asserted anywhere. Either add a row documenting expected behavior, or note in the doc comment that the constructor relies on startup validation.

pkg/authserver/storage/cimd_decorator.go
```diff
@@ -43,11 +45,20 @@ type cimdCacheEntry struct {
 // it returns base unchanged (no allocation). cacheMaxSize must be >= 1;
 // fallbackTTL is the fixed TTL applied to every cache entry (Cache-Control
 // header parsing is not yet implemented; all entries use this value).
+// scopesSupported is the AS-configured scope allowlist; documents that declare
+// scopes outside this set are rejected at fetch time. In production this is
+// always non-nil because applyDefaults populates ScopesSupported before the
```
tgrunnagle: **F8 (LOW) — Doc comments document cross-file invariants not enforced in code.**

Two related drift hazards:

1. **Here (line 50)**: "In production this is always non-nil because `applyDefaults` populates `ScopesSupported` before the decorator is constructed." This is true today (verified against `pkg/authserver/config.go::applyDefaults`), but it depends on call ordering in a *different file*. A future refactor that reorders or adds an alternate construction path could silently break this without breaking any test — and CIMD would quietly fall back to "unconstrained scope passthrough" in production.

2. **buildFositeClient doc (lines 240–241)**: "when nil, DefaultScopes is used (unconstrained AS)" frames the fallback as a decorator-state decision, but the actual code is `if len(scopes) == 0` — also triggered by, e.g., a whitespace-only `doc.Scope` on a constrained-AS path. The repo's `go-style.md` "Keep Comments Synchronized With Code" rule applies.

**Suggestions:**
- For (1): either `// see authserver.applyDefaults` cross-reference, **or** move the invariant assertion to the production caller in `server_impl.go` (e.g. `if cfg.CIMDEnabled && len(cfg.ScopesSupported) == 0 { return nil, errors.New("...") }`).
- For (2): "Fall back to DefaultScopes when no scopes were resolved by fetch()."

pkg/authserver/storage/cimd_decorator.go
```diff
@@ -43,11 +45,20 @@ type cimdCacheEntry struct {
...
+// passthrough.
+// baselineClientScopes mirrors the AS-level baseline: it is unioned into every
+// CIMD client's scope set after validation, matching DCR handler behaviour.
 func NewCIMDStorageDecorator(
 	base Storage,
 	enabled bool,
 	cacheMaxSize int,
 	fallbackTTL time.Duration,
+	scopesSupported []string,
+	baselineClientScopes []string,
```
tgrunnagle: **F4 (MEDIUM) — `NewCIMDStorageDecorator` now has 6 positional args with two adjacent same-typed slices.**

The constructor signature is `(base, enabled, cacheMaxSize, fallbackTTL, scopesSupported, baselineClientScopes)`. The last two args are both `[]string` and semantically distinct (allowlist vs. union-set) — silently swappable at every call site with no compile-time protection. The go-style rule on "Package API Surface" recommends introducing a config struct or functional options when runtime shape meaningfully differs from input or there are multiple config sources.

A future addition (e.g., Cache-Control TTL parsing, per-document size cap) would push the signature further past readability limits.

**Suggested refactor** — introduce a config struct:
```go
type CIMDDecoratorConfig struct {
    Enabled              bool
    CacheMaxSize         int
    FallbackTTL          time.Duration
    ScopesSupported      []string
    BaselineClientScopes []string
}

func NewCIMDStorageDecorator(base Storage, cfg CIMDDecoratorConfig) (Storage, error)
```

The single production caller in `server_impl.go` and the test helpers update mechanically. The struct also makes the `enabled=false → returns base` pattern more obviously a configuration concern.

pkg/authserver/storage/cimd_decorator.go
```diff
@@ -63,9 +74,11 @@ func NewCIMDStorageDecorator(
 	}
 
 	return &CIMDStorageDecorator{
-		Storage: base,
-		cache:   c,
-		ttl:     fallbackTTL,
+		Storage:              base,
+		cache:                c,
+		ttl:                  fallbackTTL,
+		scopesSupported:      scopesSupported,
```
tgrunnagle: **F7 (LOW) — Constructor stores caller-supplied scope slices without defensive `slices.Clone`.**

`scopesSupported` and `baselineClientScopes` are stored by slice-header copy. In `server_impl.go:187` the same `cfg.ScopesSupported` slice is *also* passed to the auth-server config constructor, so multiple components hold views over the same backing array. The repo's `go-style.md` "Copy Before Mutating Caller Input" rule promotes defensive cloning at constructor boundaries — a one-liner that prevents a class of future bugs.

```go
return &CIMDStorageDecorator{
    Storage:              base,
    cache:                c,
    ttl:                  fallbackTTL,
    scopesSupported:      slices.Clone(scopesSupported),
    baselineClientScopes: slices.Clone(baselineClientScopes),
}, nil
```

If a future config-reload path mutates these slices in place (e.g., appends a new scope), the decorator's view will not silently drift.

pkg/authserver/storage/cimd_decorator.go
```diff
@@ -119,17 +132,75 @@ func (d *CIMDStorageDecorator) fetch(ctx context.Context, id string) (fosite.Cli
...
+
+	// Reject documents that declare response_types the embedded AS does not support.
+	for _, rt := range doc.ResponseTypes {
+		if !allowedCIMDResponseTypes[rt] {
+			return nil, fmt.Errorf("%w: CIMD document at %s claims response_type %q "+
+				"but this server only supports %v",
+				fosite.ErrInvalidClient.WithHint("unsupported response_type"),
+				id, rt, defaultCIMDResponseTypes)
+		}
+	}
```
tgrunnagle, line 156: **F9 (LOW) — DCR-parity gaps around `response_types` and grant-type validation order.**

Two related findings around the parity claim:

1. **Response-type defense-in-depth missing.** DCR's `validateResponseTypes` (registration/dcr.go) enforces `slices.Contains(responseTypes, "code")` *in addition to* the allowlist walk. The CIMD code (this loop) only walks the allowlist. Today the allowlist contains only `"code"`, so the inputs are functionally equivalent — but a future PR adding `"code id_token"` to either allowlist would silently accept docs where `"code"` is missing on the CIMD path while DCR still rejects.

2. **Grant-type validation order reversed.** DCR checks `slices.Contains(grantTypes, "authorization_code")` *first*, then walks the allowlist. CIMD (lines 148–160 above) walks the allowlist *first*, then checks `authorization_code` presence. For input `["client_credentials"]`, DCR rejects with `"grant_types must include 'authorization_code'"`, CIMD rejects with `"unsupported grant_type"`. Different operator-visible error for the same input — minor friction in a mixed-mode debugging session.

**Cleanest fix:** extract `ValidatePublicGrantTypes` and `ValidatePublicResponseTypes` helpers in the `registration` package and call them from both DCR's handler and the CIMD decorator. The PR already does this for scope (`registration.ValidateScopes`); doing the same for grant_types/response_types would close the parity claim completely and eliminate the duplicated `allowedCIMDGrantTypes` / `allowedCIMDResponseTypes` maps.

pkg/authserver/storage/cimd_decorator.go
```diff
@@ -119,17 +132,75 @@ func (d *CIMDStorageDecorator) fetch(ctx context.Context, id string) (fosite.Cli
...
+	// scopes are used directly, or nil to let buildFositeClient apply DefaultScopes.
+	// In both cases BaselineClientScopes is unioned in after validation,
+	// matching the DCR handler's behaviour.
+	var resolvedScopes []string
+	if len(d.scopesSupported) > 0 {
+		if doc.Scope != "" {
+			computed, dcrErr := registration.ValidateScopes(strings.Fields(doc.Scope), d.scopesSupported)
+			if dcrErr != nil {
+				return nil, fmt.Errorf("%w: CIMD document at %s: %s",
+					fosite.ErrInvalidClient.WithHint(dcrErr.Error), id, dcrErr.ErrorDescription)
```
tgrunnagle: **F6 (LOW) — `WithHint(dcrErr.Error)` passes the OAuth error code where the descriptive hint belongs.**

`DCRError` has two struct fields: `Error string` (the OAuth error code, e.g. `"invalid_client_metadata"`) and `ErrorDescription string` (the human description, e.g. `"unsupported scope: foo"`). There is no `Error()` method — `dcrErr.Error` reads the field, not a stringification.

The current code passes the **error code** as the fosite hint. The descriptive text ends up in the wrapped error chain via `%s` but **not** in the fosite hint that ultimately surfaces in `WWW-Authenticate` / `error_description`. Operators / clients see `"hint":"invalid_client_metadata"` instead of `"hint":"unsupported scope: foo"`.

The other rejection branches in this file (lines 141, 152, 158, 167) pass static descriptive hints like `"unsupported grant_type"`. This branch is inconsistent.

```suggestion
				return nil, fmt.Errorf("%w: CIMD document at %s: %s",
					fosite.ErrInvalidClient.WithHint(dcrErr.ErrorDescription), id, dcrErr.ErrorDescription)
```

(Or drop the redundant `%s` at the end since the hint now carries the description.)

pkg/authserver/storage/cimd_decorator.go
```diff
@@ -119,17 +132,75 @@ func (d *CIMDStorageDecorator) fetch(ctx context.Context, id string) (fosite.Cli
...
+			}
+			resolvedScopes = computed
+		} else {
+			resolvedScopes = slices.Clone(d.scopesSupported)
```
tgrunnagle: **F5 (MEDIUM) — Omitted-scope branch silently diverges from DCR when `ScopesSupported` ⊃ `DefaultScopes`.**

When `doc.Scope == ""` and `ScopesSupported` is set, this branch gives the CIMD client the **full** `ScopesSupported`. The DCR equivalent (`registration.ValidateScopes(nil, allowed)`) returns `registration.DefaultScopes` after subset-checking it against `allowed`.

In the default config these are equal because `applyDefaults` populates `ScopesSupported = DefaultScopes`, so the divergence is invisible. But an operator who configures `scopes_supported: ["openid", "profile", "email", "offline_access", "groups"]` will see CIMD clients with no declared scope silently get `groups`, while a DCR client in the same position gets only `DefaultScopes`. The PR description says "both paths enforce the same scope policy" — this branch contradicts that.

In the worst case the extra scopes include elevated permissions (`admin:*`, custom resource scopes) that the operator added to `ScopesSupported` for internal-DCR use only, never intending to extend to dynamically-resolved third-party CIMD clients.

**Suggested fix** — call the same validator DCR uses, which returns `DefaultScopes` when the input is empty:

```suggestion
		} else {
			computed, _ := registration.ValidateScopes(nil, d.scopesSupported)
			resolvedScopes = computed
		}
```

(The error return can't trigger here in production because `applyDefaults` ensures `DefaultScopes ⊆ ScopesSupported`; if you'd rather be explicit, propagate the error.)

If the current behavior is intentional, remove the "same scope policy as DCR" claim from the PR description and add a note to the doc comment explaining the intentional divergence.

pkg/authserver/server_impl.go
```diff
@@ -177,7 +177,14 @@ func newServer(ctx context.Context, cfg Config, stor storage.Storage, opts ...se
 	// so that GetClient calls for HTTPS client_id values are intercepted at the
 	// fosite level (not just the handler level).
 	if cfg.CIMDEnabled {
-		stor, err = storage.NewCIMDStorageDecorator(stor, true, cfg.CIMDCacheMaxSize, cfg.CIMDCacheFallbackTTL)
+		if len(cfg.BaselineClientScopes) > 0 {
+			slog.Warn("CIMD is enabled with baseline_client_scopes configured; "+
```
tgrunnagle, line 181: **F12 (LOW) — `slog.Warn` level/clarity for CIMD+baseline interaction.**

The warning fires every server start when both `CIMDEnabled` and `BaselineClientScopes` are set. Two angles, both raised:

- **General/level**: per `.claude/rules/go-style.md` — "WARN for non-fatal issues (deprecations, fallback behavior, cleanup failures)". This combination isn't a fallback or deprecation; it's a legitimate operator config mirroring DCR behavior. Warning on a deliberate config means operators get a persistent WARN line they can't silence without changing behavior.
- **OAuth/clarity**: the message is true but doesn't surface the security-relevant consequence — every dynamically-resolved third-party client URL (which the operator has no a-priori knowledge of) will receive the baseline scopes, potentially including `offline_access`. For an operator who set `BaselineClientScopes` mainly for internal DCR clients, this is the moment to learn that the same expansion applies to unknown CIMD clients.

**Pick one:**
- (a) Tighten the message to name the consequence: `"CIMD is enabled with baseline_client_scopes configured; third-party clients resolved via CIMD will receive these scopes — ensure they are scopes you would grant by default to any unknown client"`. Keep at Warn.
- (b) Downgrade to `slog.Info` and document the interaction on the `BaselineClientScopes` config-field godoc.

The current shape splits the difference.

tgrunnagle (approved):
Re-review LGTM — all 12 findings from the prior multi-agent review are resolved. F3 went from 11 separate tests to 3 table-driven tests (plus a `serveCIMDDocWithFields` helper), F5 chose the fail-closed option (require explicit `scope` when `DefaultScopes ⊄ ScopesSupported`), and F9 extracted `ValidatePublicGrantTypes` / `ValidatePublicResponseTypes` so DCR and CIMD now share a single validation source. The `errors.Is(err, fosite.ErrInvalidClient)` assertions pin the error-type change so a regression to `ErrNotFound` would fail tests.

### One nice-to-have follow-up (not blocking)

`defaultCIMDGrantTypes` and `defaultCIMDResponseTypes` in `cimd_decorator.go` are still parallel to `defaultGrantTypes` / `defaultResponseTypes` in the `registration` package. After F9 the validation is shared, but the default application is still duplicated — if someone later widens `registration.defaultGrantTypes` for DCR (e.g., adds a new public-client grant), they'd need to remember to update the CIMD copy too, and `task lint` won't catch the mismatch.

Since `ValidatePublicGrantTypes` / `ValidatePublicResponseTypes` already return the defaulted slice, `buildFositeClient` could use those return values and the CIMD-side constants could be deleted. Closes the last DCR/CIMD parity gap. Not blocking for this PR — could be a small follow-up.

Approving.

PR #5394, opened 2026-05-28
Free orphaned proxy port on stop and rm

### LINKED ISSUE #5393: thv stop/rm leave the port held when the status file is missing
## Bug description

`thv stop` and `thv rm` report success but do not actually stop or clean up the workload when its status file is missing. The workload's proxy process is left running and keeps holding the workload's port:

- After `thv stop`, the surviving proxy's supervisor restarts the container, so the workload returns to `running` on its own.
- After `thv rm`, the container is removed but the orphaned proxy keeps holding the port.

Either way there is no `thv`-only recovery: a later `thv run`/`thv start` on the same port fails with "address already in use", and the user has to find and kill the proxy manually (e.g. `lsof -ti :<port> | xargs kill`). The documented stop behavior is "Stops proxy process → stops container → preserves state" (`docs/arch/08-workloads-lifecycle.md`), so the surviving proxy is a contract violation.

## Steps to reproduce

The precondition is a missing status file. The steps below remove it deliberately to reproduce on demand; we have also seen workloads reach this state on their own (listed in `thv ls` but with no status file). This report is about the broken stop/rm behavior, not the cause of the missing file.

1. Start a container workload with a pinned proxy port:
   `thv run --name demo --proxy-port 51999 fetch`
2. Confirm the proxy is listening and note its PID:
   `lsof -i :51999 -sTCP:LISTEN`
3. Remove the workload's status file:
   - macOS: `~/Library/Application Support/toolhive/statuses/demo.json`
   - Linux: `~/.local/share/toolhive/statuses/demo.json`
4. `thv ls` still lists `demo` as `running` (this entry comes from the runtime, not the status file).
5. Run `thv stop demo` (or `thv rm demo`). It exits successfully.
6. Observe that the proxy was not terminated (the PID from step 2 is still alive, e.g. `ps -p <pid>`, and `:51999` is still held):
   - After `thv stop`: within ~10s `thv ls` shows `demo` back to `running`, because the surviving supervisor restarted the container.
   - After `thv rm`: `demo` is gone from `thv ls`, but the proxy still holds `:51999`, so `thv run --proxy-port 51999 ...` fails with "address already in use".

## Expected behavior

`thv stop` and `thv rm` terminate the workload's proxy and free its port, even when the status file is missing — matching the documented "stops proxy process" behavior.

## Actual behavior

The proxy is not terminated. After `thv stop`, the supervisor restarts the container and the workload returns to `running`. After `thv rm`, the container is removed but the orphaned proxy keeps holding the port. Both commands report success regardless.

## Environment (if relevant)

- OS/version: macOS 15 (Darwin 25.5.0), OrbStack runtime.
- ToolHive version: v0.28.3 (also reproduced against `main` at `8932f938`).

## Additional context

Root cause: the proxy-stop path (`stopProcess` in `pkg/workloads/manager.go`) terminates the proxy using only the PID recorded in the status file (`GetWorkloadPID`). When the status file is missing, `GetWorkloadPID` returns `(0, nil)`, `KillProcess(0)` fails, and no proxy is killed. The container is stopped/removed through the runtime, but the detached proxy (`thv start <name> --foreground`) keeps running; its supervisor (the `RunWorkload` retry loop) then restarts the container on `stop`. This path reads the PID from the status file on every platform, so the issue is not macOS-specific.

gkatz2:
## Summary

When a workload's status file is missing, `thv stop` and `thv rm` report success but leave the workload's proxy process running and holding its port. The proxy-stop path kills the proxy by the PID recorded in the status file, so with the file gone nothing is killed:

- After `thv stop`, the surviving supervisor restarts the container, so the workload returns to `running` on its own.
- After `thv rm`, the container is removed but the orphaned proxy keeps holding the port, so it cannot be reused without killing the process by hand.

This makes stop and rm fall back to the existing port-based cleanup when the PID-based stop finds no proxy to kill, so the proxy is terminated and the port freed even when the status file is missing. The fallback reuses `freePortHolderIfNeeded` (already used on the restart path), which only kills a process verified to be this workload's own proxy.

- Make `stopProcess` / `stopProxyIfNeeded` report whether a tracked proxy was actually killed.
- Thread the already-loaded `runConfig` into the container stop/delete paths so the fallback knows the proxy port.
- When the PID-based stop fails for a non-auxiliary workload, run the port-based cleanup as a backstop.

Fixes #5393

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Manual testing on macOS + OrbStack with a real container workload (`fetch`):

- Reproduced the bug: with the status file moved aside, `thv stop` left the supervisor process alive (verified by PID) and it recreated the container — a new container ID and new `StartedAt` — returning the workload to `running`; `thv rm` left the orphaned proxy holding the port.
- With the fix: both `thv stop` and `thv rm` terminate the proxy (PID gone) and free the port.
- Confirmed the normal path (status file present) is unchanged: the proxy is stopped by PID and the port-based fallback does not run.

The added unit tests fail without the fix and pass with it.

Yes. `thv stop` and `thv rm` now reliably stop the workload's proxy and free its port even when the workload's status file is missing, instead of leaving an orphaned proxy that holds the port (and, for `stop`, restarts the container).

- The fallback is **gated**: it runs only when the PID-based stop returns false (no tracked PID, or the kill failed). The normal stop/rm path is unchanged — no added latency, no behavior change.
- The kill is **identity-verified**: `freePortHolderIfNeeded` → `process.IsToolHiveProxyForWorkload` confirms the process on the port is this workload's `thv start <name>` proxy before killing it, so it cannot touch an unrelated process or another workload's proxy.
- **Limitation:** if `runner.LoadState` itself fails (the run config is gone, not just the status file), the proxy port cannot be recovered. In the reported scenario only the status file is missing, so `LoadState` succeeds and the port is recoverable.
- The deeper question of *why* a status file goes missing is out of scope here; this fixes the resulting inability to stop/remove the workload.

### DIFF
--- a/docs/arch/08-workloads-lifecycle.md
+++ b/docs/arch/08-workloads-lifecycle.md
@@ -134,6 +134,12 @@ The same `Applier` backs both the CLI (`thv upgrade apply`) and the API (`POST /
 
 **Implementation**: `pkg/workloads/upgrade/`, `cmd/thv/app/upgrade.go`
 
+### Proxy termination
+
+Stop and delete terminate the proxy using its recorded PID. When PID-based termination is unavailable or fails (for example, the status file is missing, records no PID, or the process is already gone), they fall back to port-based cleanup: the process holding the proxy port is terminated only after it is confirmed to be this workload's proxy. This prevents an orphaned proxy from continuing to hold the port after the container has been stopped or removed.
+
+**Implementation**: `pkg/workloads/manager.go`
+
 ### List
 
 Listing combines container workloads from the runtime with remote workloads from persisted state. The manager can filter workloads by label or group, and can optionally include stopped workloads.
--- a/pkg/workloads/manager.go
+++ b/pkg/workloads/manager.go
@@ -101,6 +101,7 @@ type DefaultManager struct {
 	retryConfig     *retryConfig
 	newRunner       mcpRunnerFactory
 	detachedSpawner detachedProcessSpawner
+	portFreer       portFreer
 }
 
 // mcpRunner is the subset of *runner.Runner that RunWorkload's retry loop
@@ -122,6 +123,11 @@ type mcpRunnerFactory func(*runner.RunConfig, statuses.StatusManager) mcpRunner
 // Tests override this with a no-op so the test binary does not re-exec
 // itself.
 type detachedProcessSpawner func(ctx context.Context, runConfig *runner.RunConfig) (pid int, err error)
+
+// portFreer kills an orphaned ToolHive proxy still holding a workload's proxy
+// port. In production it is freePortHolderIfNeeded; tests override it to observe
+// the stop/delete fallback without making real OS calls.
+type portFreer func(ctx context.Context, runConfig *runner.RunConfig)
 
 // retryConfig bundles the RunWorkload retry loop's tunable parameters.
 type retryConfig struct {
@@ -160,6 +166,12 @@ func withDetachedSpawner(s detachedProcessSpawner) managerOption {
 	return func(d *DefaultManager) { d.detachedSpawner = s }
 }
 
+// withPortFreer overrides the function used to free an orphaned proxy port
+// during stop/delete. Intended for tests so they do not make real OS calls.
+func withPortFreer(f portFreer) managerOption {
+	return func(d *DefaultManager) { d.portFreer = f }
+}
+
 // retryConfigOrDefault returns the manager's retryConfig if set, otherwise defaults.
 func (d *DefaultManager) retryConfigOrDefault() retryConfig {
 	if d.retryConfig == nil {
@@ -185,6 +197,15 @@ func (d *DefaultManager) detachedSpawnerOrDefault() detachedProcessSpawner {
 		return d.detachedSpawner
 	}
 	return d.spawnDetached
+}
+
+// portFreerOrDefault returns the manager's port freer if set, otherwise the
+// production default that frees the proxy port via freePortHolderIfNeeded.
+func (d *DefaultManager) portFreerOrDefault() portFreer {
+	if d.portFreer != nil {
+		return d.portFreer
+	}
+	return d.freePortHolderIfNeeded
 }
 
 // ErrWorkloadNotRunning is returned when a container cannot be found by name.
@@ -415,9 +436,10 @@ func (d *DefaultManager) stopSingleWorkload(ctx context.Context, name string) er
 	// First, try to load the run configuration to check if it's a remote workload
 	runConfig, err := runner.LoadState(childCtx, name)
 	if err != nil {
-		// If we can't load the state, it might be a container workload or the workload doesn't exist
-		// Try to stop it as a container workload
-		return d.stopContainerWorkload(childCtx, name)
+		// If we can't load the state, it might be a container workload or the workload doesn't exist.
+		// Try to stop it as a container workload. Without a run config we cannot recover the proxy
+		// port for port-based cleanup, so pass nil.
+		return d.stopContainerWorkload(childCtx, name, nil)
 	}
 
 	// Check if this is a remote workload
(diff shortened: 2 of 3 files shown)


pkg/workloads/manager.go
```diff
@@ -976,40 +1007,51 @@ func (d *DefaultManager) isSupervisorProcessAlive(ctx context.Context, name stri
...
+// reports whether a tracked proxy process was stopped (see stopProcess); false
+// when there is no base name or no proxy was killed.
+func (d *DefaultManager) stopProxyIfNeeded(ctx context.Context, name, baseName string) bool {
 	slog.Debug("removing proxy process", "workload", name)
-	if baseName != "" {
-		d.stopProcess(ctx, baseName)
+	if baseName == "" {
+		return false
 	}
+	return d.stopProcess(ctx, baseName)
```
jhrozek, line 1062: Now that stopProxyIfNeeded reports whether a proxy was actually stopped, the container stop/delete paths use it to trigger the port-based fallback. stopRemoteWorkload (around line 472) still calls this and discards the result, so a remote workload with a missing status file wouldn't get the orphan-port cleanup. Remote workloads run a local proxy that holds a port too, so the same gap seems to apply there. Is the remote path intentionally out of scope for this PR, or worth adding the same `if !... { portFreerOrDefault()(ctx, runConfig) }` fallback?

pkg/workloads/manager.go
```diff
@@ -976,40 +1007,51 @@ func (d *DefaultManager) isSupervisorProcessAlive(ctx context.Context, name stri
...
 }
 
 // freePortHolderIfNeeded kills the process holding the proxy port if it is in use.
```
jhrozek: This doc comment only describes the restart re-bind case, but after this change freePortHolderIfNeeded is also the fallback used during stop/delete, where there is no child re-binding. Could you broaden it to cover both uses so it doesn't read as restart-only?

jhrozek (commented):
A couple of non-blocking questions on the orphan-port cleanup. The fix itself looks correct for the container stop/rm paths. Nothing here blocks merge.

jhrozek (commented):
**[HIGH] Remote-workload paths don't check `stopProxyIfNeeded` return — no port-cleanup fallback**

The PR correctly checks the boolean return of `stopProxyIfNeeded` in both container-workload paths (`stopSingleContainerWorkload` line 1620, `deleteContainerWorkload` line 951) and falls back to `portFreerOrDefault()` when it returns `false`. But the two remote-workload equivalents still discard the return:

- `pkg/workloads/manager.go` line 473 (`stopRemoteWorkload`): `d.stopProxyIfNeeded(ctx, name, runConfig.BaseName)` — return ignored
- `pkg/workloads/manager.go` line 896 (`deleteRemoteWorkload`): `d.stopProxyIfNeeded(ctx, name, runConfig.BaseName)` — return ignored

Remote workloads can also have orphaned proxies after a status-file loss. The same pattern should apply to both:

```go
if runConfig.BaseName != "" {
    if !d.stopProxyIfNeeded(ctx, name, runConfig.BaseName) {
        d.portFreerOrDefault()(ctx, runConfig)
    }
}
```

Note: the third discarded call at `maybeSetupRemoteWorkload` line 1347 is intentionally left as-is — there's an unconditional `portFreerOrDefault()` call a few lines later in that function.

docs/arch/08-workloads-lifecycle.md
```diff
@@ -96,6 +96,12 @@ thv rm my-server
 
 **Implementation**: `pkg/workloads/manager.go`
 
+### Proxy termination
+
+Stop and delete terminate the proxy using its recorded PID. When that PID is unavailable (for example, the status file is missing or records no PID), they fall back to port-based cleanup: the process holding the proxy port is terminated only after it is confirmed to be this workload's proxy. This prevents an orphaned proxy from continuing to hold the port after the container has been stopped or removed.
```
jhrozek: The fallback also fires when PID-based termination *fails* (e.g. `KillProcess` returns an error for a stale PID whose process is already gone), not only when the PID is unavailable. The current wording implies the fallback is only for the no-PID case.

```suggestion
Stop and delete terminate the proxy using its recorded PID. When PID-based termination is unavailable or fails (for example, the status file is missing, records no PID, or the process is already gone), they fall back to port-based cleanup: the process holding the proxy port is terminated only after it is confirmed to be this workload's proxy. This prevents an orphaned proxy from continuing to hold the port after the container has been stopped or removed.
```

jhrozek (dismissed):

jhrozek:
Still looking good would you mind resolving the conflicts so I can reapprove?

gkatz2:
Rebased onto current main and resolved the conflict — it was only in `manager_test.go`, against the AuthRetrying tests from #5350 (kept both test sets). Ready for re-review whenever you have a moment.

pkg/workloads/manager.go
gkatz2: Addressed, with a correction. Verifying end-to-end, I found that the remote fallback was unreachable in the real orphan case: with the status file gone, `stopRemoteWorkload` returns at the `GetWorkload`→`ErrWorkloadNotFound` guard *before* the fallback (my unit test hid it by mocking a running workload). Restructured so the proxy/port cleanup runs regardless of the status check, and verified against a real remote orphan this time — the proxy is now actually killed. (`rm` was already fine — no such guard.)

jhrozek (approved):

PR #5407, opened 2026-06-01
Add workload upgrade detection package

JAORMX:
## Summary

CLI and API users have no way to discover when a registry-sourced MCP server has a newer version available. Today only ToolHive Studio detects image-tag drift, and it does so in its frontend — so CLI users are blind to updates and every new client would have to reimplement the same logic. This is the first slice of [RFC THV-0068 (Workload Upgrade)](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0068-workload-upgrade.md), moving detection into the backend where all clients can consume it. Local experience only (Kubernetes is out of scope for this RFC).

- Add a new `pkg/workloads/upgrade` package with a `Checker` that compares a running workload's image tag against its current registry entry.
- Tag comparison is semver-aware (via `golang.org/x/mod/semver`) with a string-equality fallback; it degrades safely to `unknown` for `:latest`, digest references, repository changes (host-normalized), and non-registry-sourced workloads, so only a strictly-newer tag on the same repository yields `upgrade-available`.
- Reports environment-variable drift (new/removed registry-declared vars) and configuration drift (permission profile, transport, network isolation), with secret env-var defaults cleared from results.

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — table-driven coverage of tag comparison (incl. double-digit-minor and prerelease ordering), env/config drift, and status gating; registry provider mocked.

## Changes

| File | Change |
|------|--------|
| `pkg/workloads/upgrade/doc.go` | Package doc, cycle guard, no-rollback note |
| `pkg/workloads/upgrade/types.go` | `CheckResult`, `EnvVarDrift`, `ConfigDrift`, `UpgradeStatus`, `ApplyOptions` |
| `pkg/workloads/upgrade/version.go` | Semver-aware tag comparison helper |
| `pkg/workloads/upgrade/checker.go` | `Checker` + `Check`/`CheckAll` |
| `pkg/workloads/upgrade/*_test.go` | Unit tests incl. a drift-guard for the env-var boundary type |

No — this PR adds an internal package only. The `thv upgrade` commands and REST endpoints that consume it land in later PRs of this stack.

## Large PR Justification

This PR is ~1,158 lines, but **627 of those are unit tests** (`*_test.go`) and only **531 are source** across one cohesive new package (`doc.go` + `types.go` + `version.go` + `checker.go`).

- **The source cannot be split smaller without leaving code uncompilable or untested.** The `Checker` depends on its result types (`types.go`) and the tag-comparison helper (`version.go`); separating them yields a package that does not build. This is one self-contained unit, the smallest reviewable slice of the feature.
- **Tests must ship with the code, not separately.** Splitting the 627 lines of table-driven tests into a follow-up PR would merge untested detection logic — exactly what review is meant to prevent.
- The feature as a whole was deliberately split into 6 stacked PRs (this is #1) to keep each phase focused; this phase is the irreducible detection core.

This is **PR 1 of 6** in a stack implementing RFC THV-0068 (split to stay within the repo's PR-size limit). Each PR is one logical phase; later PRs are based on this branch:

1. **Detection package** ← this PR
2. Check REST API (`GET .../upgrade-check`)
3. CLI `thv upgrade check` + `thv list --check-upgrades`
4. `upgrade.Applier`
5. Apply CLI + API (`thv upgrade apply`, `POST .../upgrade`)
6. e2e + lifecycle docs


pkg/workloads/upgrade/version.go
```diff
@@ -0,0 +1,119 @@
...
+// version (e.g. "1.2.3" -> "v1.2.3") so it can be validated and compared by
+// golang.org/x/mod/semver, which requires the "v" prefix. Tags that already
+// start with "v", or that are not semver-shaped, are returned unchanged.
+func normalizeSemver(tag string) string {
+	if tag == "" {
+		return tag
+	}
+	if tag[0] == 'v' || tag[0] == 'V' {
+		return tag
+	}
```
jhrozek: Uppercase `V`-prefixed tags silently suppress upgrade detection. `golang.org/x/mod/semver` requires a lowercase `v` prefix — `semver.IsValid("V1.2.3")` returns `false` (verified). So a `V`-prefixed tag is returned unchanged here, then fails the semver path and falls through to the non-semver exact-match fallback. For `V1.2.0` vs `V1.3.0` that yields `comparisonUndecidable` → `StatusUnknown`, hiding a real upgrade. Since the code already special-cases `'V'`, the intent was clearly to support it.

```suggestion
	if tag[0] == 'V' {
		return "v" + tag[1:] // lowercase the prefix so semver.IsValid accepts it
	}
	if tag[0] == 'v' {
		return tag
	}
```

Worth adding a `version_test.go` case asserting `V1.0.0` vs `V2.0.0` → `comparisonNewer`.

pkg/workloads/upgrade/checker.go
```diff
@@ -0,0 +1,218 @@
...
+	currentTransport := cfg.Transport.String()
+	candidateTransport := imgMeta.GetTransport()
+	if candidateTransport != "" && currentTransport != "" && currentTransport != candidateTransport {
+		drift.Transport = &StringChange{From: currentTransport, To: candidateTransport}
+	}
+
+	// Network isolation is a plain boolean on the config; the registry entry has
+	// no explicit network-isolation field, so the candidate posture is the
+	// default (false). Report drift only when the workload currently isolates.
+	if cfg.IsolateNetwork {
```
jhrozek: `NetworkIsolation` drift fires unconditionally for any isolated workload. Because the registry has no network-isolation field, the candidate posture is hard-coded to `false`, so every upgrade-available workload launched with `--isolate-network` always reports `{From: true, To: false}` — drift that's unrelated to the candidate version and would appear for *any* version of the image.

The comment explains the mechanism but not that it always fires. Two concerns: (1) it's noise that frames a stable local choice as a registry-driven change; (2) if a future apply path reconciles toward the candidate, honoring this would silently drop network isolation and weaken the security posture. Consider omitting it (it can't originate from the registry) or documenting/renaming it as a workload-posture reminder rather than registry drift.

pkg/workloads/upgrade/checker.go
```diff
@@ -0,0 +1,218 @@
...
+	case comparisonNewer:
+		result.Status = StatusUpgradeAvailable
+		result.EnvVarDrift = computeEnvDrift(cfg, imgMeta)
+		result.ConfigDrift = computeConfigDrift(cfg, imgMeta)
+	case comparisonSameOrOlder:
+		result.Status = StatusUpToDate
+	case comparisonUndecidable:
+		result.Status = StatusUnknown
+		result.Reason = reason
+	}
```
jhrozek, line 100: Defensive nit (non-blocking): this switch is exhaustive today, but it has no `default`, and `comparisonSameOrOlder` is the `iota` zero value of `tagComparison`. If a future `tagComparison` value is added and a path returns it (or returns an unset zero value), the result silently becomes `StatusUpToDate`/`""` — the least-safe outcome for an upgrade checker. A `default: result.Status = StatusUnknown` here (and/or reordering the enum so `comparisonUndecidable` is the zero value) would make the safe outcome the default.

jhrozek (commented):
Detection logic looks solid and the tests are thorough. A few comments from a focused review of the comparison/drift logic — one correctness item on semver normalization, plus two defensive/semantic notes. Nothing blocking.

pkg/workloads/upgrade/checker.go
```diff
@@ -0,0 +1,218 @@
...
+
+	server, err := c.provider.GetServer(cfg.RegistryServerName)
+	if err != nil {
+		if errors.Is(err, registry.ErrServerNotFound) {
+			result.Status = StatusServerNotFound
+			return result, nil
+		}
+		result.Status = StatusUnknown
+		result.Reason = fmt.Sprintf("registry lookup failed: %v", err)
```
jhrozek: Heads up for when the API endpoints land on top of this (PR #5408): `Reason` is serialized straight into the HTTP response, and here it wraps the raw provider error. For an unreachable/misconfigured registry that can surface internal detail to callers — `registry lookup failed: dial tcp 10.x.x.x:443: ...` and the like. The `check failed: %v` reason in `CheckAll` has the same property.

Might be worth logging the detail at DEBUG and returning a fixed string instead:

```go
slog.Debug("registry lookup failed", "server", cfg.RegistryServerName, "error", err)
result.Reason = "registry lookup failed"
```

The comparison-derived reasons are built from the user's own image refs, so those are fine. Non-blocking.

pkg/workloads/upgrade/version.go
JAORMX: Good catch — fixed. `normalizeSemver` now lowercases an uppercase `V` prefix (`V1.2.3` → `v1.2.3`) so semver comparison applies, and I added a `V1.2.0` vs `V1.3.0` → `comparisonNewer` case to `version_test.go` (plus updated the old `TestNormalizeSemver` expectation). Pushed in `Address review feedback on upgrade detection`.

pkg/workloads/upgrade/checker.go
JAORMX: Agreed — dropped it entirely. The registry has no network-isolation field, so this could never be a registry-driven change; it only added noise (and risked a future apply weakening the posture). Removed the detection, the `ConfigDrift.NetworkIsolation` field, and the now-unused `BoolChange` type, and flipped the test to assert isolation no longer drifts. The CLI print and the lifecycle doc were updated to match further up the stack.

pkg/workloads/upgrade/checker.go
JAORMX: Done — added `default: result.Status = StatusUnknown` so an unexpected/zero comparison value lands on the safe outcome rather than `StatusUpToDate`.

pkg/workloads/upgrade/checker.go
JAORMX: Fixed. Both the `Check` registry-lookup path and the `CheckAll` path now log the raw error at DEBUG and set a fixed `Reason` (`"registry lookup failed"` / `"check failed"`), so the provider error (and any internal addressing it carries) never reaches the serialized response. The comparison-derived reasons are left as-is since they're built from the user's own image refs.

jhrozek (approved):

jhrozek (approved):

PR #5408, opened 2026-06-01
Add upgrade-check REST endpoints for workloads

JAORMX:
## Summary

The upgrade checker (PR #5407) is only useful if clients can reach it. CLI, Studio, and automation should share one backend source of truth for upgrade availability instead of each reimplementing registry drift detection. This exposes the checker over the existing workloads API. Part of [RFC THV-0068](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0068-workload-upgrade.md), local scope.

- Add `GET /api/v1beta/workloads/upgrade-check` (batch) and `GET /api/v1beta/workloads/{name}/upgrade-check` (single).
- The batch handler reuses the **exact** group/all authorization scoping of the list endpoint and intersects it with the enumerated run configs, so it can never report a workload outside the caller's scope.
- Responses carry only non-sensitive `CheckResult` metadata; secret env-var defaults are cleared. Both routes are read-only and skip image pulls, so they use the standard request timeout.
- Regenerate the OpenAPI spec.

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — single (up-to-date / available / 404 / 400), batch (mixed results, group filter, stale-on-disk config excluded, unloadable config skipped-not-fatal), a no-secret-leak assertion, and a chi route-ordering test (`/upgrade-check` does not resolve to `getWorkload`).

## Changes

| File | Change |
|------|--------|
| `pkg/api/v1/workloads_upgrade.go` | Batch + single check handlers, run-config enumeration |
| `pkg/api/v1/workloads.go` | Route registration + injected deps |
| `pkg/api/v1/workload_types.go` | Response types |
| `docs/server/*` | Regenerated OpenAPI |

Yes — two new read-only REST endpoints for querying upgrade availability. No existing endpoints change.

## Large PR Justification

This PR is ~1,434 lines, but **only 231 are hand-written source.** The rest is mechanically generated or test code that cannot be meaningfully split out:

- **764 lines are auto-generated OpenAPI** (`docs/server/docs.go`, `swagger.json`, `swagger.yaml`), produced by `task docs` from the swag annotations on the two new handlers. These must be regenerated and committed in the same PR (a `Verify Swagger Documentation` CI check fails otherwise), and they cannot be split.
- **439 lines are integration tests** for the new endpoints, which must ship with the handlers they cover.
- The 231 source lines are the two read-only handlers plus their response types — already the smallest coherent API slice (check endpoints only; the apply endpoint is deferred to PR #5411).

**PR 2 of 6** in the RFC THV-0068 stack; based on #5407. The dedicated apply endpoint (`POST .../upgrade`) lands in PR 5, reusing the response types added here.


pkg/api/v1/workloads_upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+	if groupFilter != "" {
+		if err := groupval.ValidateName(groupFilter); err != nil {
+			return httperr.WithCode(
+				fmt.Errorf("invalid group name: %w", err),
+				http.StatusBadRequest,
+			)
+		}
+		workloadList, err = workloads.FilterByGroup(workloadList, groupFilter)
+		if err != nil {
+			return err // groups.ErrGroupNotFound already has 404 status code
```
jhrozek: `FilterByGroup` never returns `ErrGroupNotFound` — `FilterByGroups` just returns an empty slice with a nil error when nothing matches (pkg/workloads/filter.go). So this `err != nil` branch can't actually fire on an unknown group, and a typo'd `?group=` returns `200 {"results":[]}` rather than the 404 the swagger annotation advertises. That silently hides the fact the group name was wrong.

Since `WorkloadRoutes` already holds `groupManager`, an existence check before filtering would make the 404 real:

```go
if _, err := s.groupManager.Get(ctx, groupFilter); err != nil {
    return err // ErrGroupNotFound (404)
}
workloadList, err = workloads.FilterByGroup(workloadList, groupFilter)
```

`listWorkloads` has the same pattern, so worth fixing both (or dropping the 404 from the docs). Not blocking.

JAORMX: Fixed for real rather than dropping the 404 from the docs. Both handlers now call `groupManager.Exists` before filtering and return `groups.ErrGroupNotFound` (404) for an unknown group. I fixed `listWorkloads` here too (it already advertised the 404 in its swagger annotation, so this makes the docs honest) rather than spinning up a 7th stacked PR. Added a bulk upgrade-check test covering the unknown-group → 404 path.

jhrozek (dismissed):

jhrozek (approved):

PR #5409, opened 2026-06-01
Add thv upgrade check and list --check-upgrades

JAORMX:
## Summary

CLI users still can't see whether their registry-sourced MCP servers have newer versions. This surfaces the upgrade checker on the command line. Part of [RFC THV-0068](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0068-workload-upgrade.md), local scope.

- Add a `thv upgrade` command group with a `check [name]` subcommand: with a name it prints a verbose report (candidate image, new env vars, permission/transport/network posture drift); with no name it prints a table for all workloads. Supports `--format json|text`.
- Add an opt-in `--check-upgrades` flag to `thv list` that appends an upgrade column.
- Both reuse the `pkg/workloads/upgrade` checker and only format results. The default `thv list` path is unchanged and performs no registry lookup, so it stays offline-friendly. Bulk output is sorted by name to match `thv list`.

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — output-formatting helpers only (per the repo's CLI testing convention; command behavior is covered by the e2e in PR 6).

## Changes

| File | Change |
|------|--------|
| `cmd/thv/app/upgrade.go` | `upgrade` group + `check` subcommand |
| `cmd/thv/app/list.go` | `--check-upgrades` flag + column |
| `cmd/thv/app/commands.go` | Register the command |
| `docs/cli/*` | Regenerated CLI reference |

Yes — new `thv upgrade check` command and an opt-in `thv list --check-upgrades` column.

**PR 3 of 6** in the RFC THV-0068 stack; based on #5408. The CLI stays thin — all comparison logic lives in `pkg/workloads/upgrade`.


cmd/thv/app/list.go
```diff
@@ -81,10 +87,20 @@ func listCmdFunc(cmd *cobra.Command, _ []string) error {
 		}
 	}
 
+	// Optionally compute upgrade status for each workload. This is the only path
+	// that performs a registry lookup; the default list stays offline-friendly.
+	var upgrades map[string]*upgrade.CheckResult
+	if listCheckUpgrades {
+		upgrades, err = checkUpgradesForWorkloads(ctx, workloadList)
```
jhrozek, line 108: When `--check-upgrades` is combined with `--format mcpservers`, this still runs the full upgrade check (a registry provider build plus a `LoadState` + registry lookup per workload), but the `mcpservers` branch below calls `printMCPServersOutput(workloadList)` and never uses the result. So the user pays for the network round-trips and gets no difference in output.

Could we either skip the computation when the format is `mcpservers`, or reject that flag combo in `PreRunE`? Moving the call into the JSON/text branches that actually consume `upgrades` would also do it.

cmd/thv/app/upgrade.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create upgrade checker: %w", err)
+	}
+	return checker, nil
+}
+
+// loadWorkloadRunConfigs enumerates all workloads (mirroring the list command's
+// path) and loads each workload's saved RunConfig. Configs that fail to load are
+// skipped so a single corrupt entry does not abort the whole check.
+func loadWorkloadRunConfigs(ctx context.Context) ([]*runner.RunConfig, error) {
```
jhrozek, line 125: This load loop is near-identical to `checkUpgradesForWorkloads` in list.go (same `runner.LoadState` per workload, same skip-on-error debug log). Not a runtime concern since each path is used by one command, but the two copies will drift over time - a future tweak to the skip behavior or filtering would likely only land in one. Might be worth extracting a shared `loadRunConfigsForWorkloads(ctx, []core.Workload)` helper that both call. Not blocking.

cmd/thv/app/upgrade.go
```diff
@@ -0,0 +1,265 @@
...
+		return
+	}
+
+	w := tabwriter.NewWriter(os.Stdout, 0, 0, 3, ' ', 0)
+	if _, err := fmt.Fprintln(w, "NAME\tSTATUS\tCURRENT\tCANDIDATE\tNEW-ENV\tPOSTURE"); err != nil {
+		slog.Warn(fmt.Sprintf("Failed to write output header: %v", err))
+		return
+	}
+
+	for _, r := range results {
```
jhrozek, line 177: Minor: this loop dereferences `r` without a nil check. `CheckAll` never returns nil entries today, so it's safe, but the guarantee only holds by reading that implementation. A cheap `if r == nil { continue }` here would keep this robust if `CheckAll` ever changes.

jhrozek (commented):
A few notes from reviewing the CLI surface. Nothing blocking - the change is clean and follows the existing command patterns well. Left one usability point and two minor maintainability/hardening suggestions inline.

cmd/thv/app/list.go
JAORMX: Fixed by rejecting the combination in `PreRunE`: `--check-upgrades` with `--format mcpservers` now errors with an actionable message instead of doing per-workload registry lookups and discarding them. Went with failing loudly over silently skipping, since the combo is meaningless (the mcpservers format has no upgrade column).

cmd/thv/app/upgrade.go
JAORMX: Added `if r == nil { continue }` to the bulk loop.

cmd/thv/app/upgrade.go
JAORMX: Acknowledged — leaving the shared `loadRunConfigsForWorkloads` extraction out of this round to keep the diff focused on the behavioral fixes. Agree the two copies will drift; I'll do the extraction as a small follow-up so it isn't lost.

jhrozek (dismissed):

jhrozek (approved):

PR #5410, opened 2026-06-01
Add Applier for upgrading workloads in place

JAORMX:
## Summary

Detecting an available upgrade is only useful if users can apply it while keeping their configuration. This adds the apply core that the CLI and API drive in the next PR. Part of [RFC THV-0068](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0068-workload-upgrade.md), local scope.

- Add `upgrade.Applier`: it reloads the workload's saved config, **re-runs the check on fresh state** (so a stale result can never drive an apply — closes the TOCTOU window), resolves the candidate from the registry, and rebuilds the run config **preserving the full user configuration** (auth, authz, audit, telemetry, tools filters, volumes, secrets, ports, permission profile, …), changing only the image, merged env/secrets, and re-resolved registry URLs. Newly required env vars surface through the injected validator.
- **Security-critical ordering:** the candidate image is verified and pulled (and the policy gate runs) **before** the destructive stop/delete/start, so a missing or unverifiable image leaves the running workload untouched — there is no rollback once `UpdateWorkload` begins. Verification uses the same path as `thv run`.
- Apply awaits `UpdateWorkload`'s completion so the upgraded config is durably persisted before returning (both CLI and API are synchronous).

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — no-op when up-to-date; resolve/verify/pull failure leaves the workload untouched (`UpdateWorkload` never called); ordering assertion (verify+pull before update, completion awaited); the exact gated+pulled config is the deployed one; full-config preservation; copy-before-mutate; completion-failure surfaces as an error.

## Changes

| File | Change |
|------|--------|
| `pkg/workloads/upgrade/applier.go` | `Applier`, `Apply`, config merge/preservation |
| `pkg/workloads/upgrade/applier_test.go` | Unit tests (manager + retriever seams mocked) |

No — internal apply path. The `thv upgrade apply` command and `POST .../upgrade` endpoint that drive it land in PR 5.

## Large PR Justification

This PR is ~1,015 lines, but **576 are unit tests** and **439 are a single source file** (`applier.go`).

- **The apply path must be atomic to be reviewable.** `Apply` is one sequence — load → re-check → resolve+verify → rebuild (preserving every user-owned field) → pull → recreate → await completion. Splitting it would create half-built intermediate states that neither compile meaningfully nor convey the security-critical ordering that is the whole point of the change.
- **This is the highest-risk code in the feature**, so its tests are intentionally exhaustive (the 576 lines cover every failure point, the verify-before-destroy ordering, full-config preservation, and copy-before-mutate). They must ship with the code under review.
- No generated code; the size is genuinely source+tests for one cohesive unit.

**PR 4 of 6** in the RFC THV-0068 stack; based on #5409. This is the highest-risk change — the destructive recreate path. The verify-then-pull-before-destroy ordering and full-config preservation were the focus of the security/Go review (an earlier draft silently dropped auth/audit/tools-filter config on upgrade; now every user-owned field is preserved). On Kubernetes the byte-level pull is delegated to the kubelet, but verification + policy gate still precede destruction; local runtime is the RFC's scope.

### DIFF
--- a/pkg/workloads/upgrade/applier.go
+++ b/pkg/workloads/upgrade/applier.go
@@ -1,0 +1,472 @@
+// SPDX-FileCopyrightText: Copyright 2026 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package upgrade
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"maps"
+	"slices"
+	"time"
+
+	regtypes "github.com/stacklok/toolhive-core/registry/types"
+	"github.com/stacklok/toolhive/pkg/config"
+	"github.com/stacklok/toolhive/pkg/container/templates"
+	"github.com/stacklok/toolhive/pkg/runner"
+	"github.com/stacklok/toolhive/pkg/runner/retriever"
+	"github.com/stacklok/toolhive/pkg/secrets"
+	"github.com/stacklok/toolhive/pkg/workloads"
+)
+
+// resolveFunc resolves an MCP server definition from the registry and verifies
+// its image provenance. It mirrors retriever.ResolveMCPServer so the real
+// implementation can be swapped for a stub in tests.
+type resolveFunc func(
+	ctx context.Context,
+	serverOrImage string,
+	rawCACertPath string,
+	verificationType string,
+	groupName string,
+	runtimeOverride *templates.RuntimeConfig,
+) (string, regtypes.ServerMetadata, error)
+
+// enforcePullFunc runs the policy gate and performs a verified image pull. It
+// mirrors retriever.EnforcePolicyAndPullImage so the real implementation can be
+// swapped for a stub in tests.
+type enforcePullFunc func(
+	ctx context.Context,
+	runConfig *runner.RunConfig,
+	serverMetadata regtypes.ServerMetadata,
+	imageURL string,
+	puller retriever.ImagePuller,
+	pullTimeout time.Duration,
+	locallyBuilt bool,
+) error
+
+// loadStateFunc loads a workload's persisted RunConfig. It mirrors
+// runner.LoadState so the real implementation can be swapped in tests.
+type loadStateFunc func(ctx context.Context, name string) (*runner.RunConfig, error)
+
+// ErrApplyAfterDestroy marks an Apply failure that occurred AFTER the
+// destructive recreate began (the manager has already started to stop/delete
+// the workload). Unlike preparation failures — which leave the running workload
+// untouched — a failure tagged with this sentinel means the workload may be
+// stopped, deleted, or only partially recreated: its state is uncertain and the
+// failure is NOT safely retryable against an intact workload. Callers use
+// errors.Is(err, ErrApplyAfterDestroy) to distinguish this 5xx-class condition
+// from the 4xx-class "candidate could not be prepared" failures.
+var ErrApplyAfterDestroy = errors.New("upgrade failed after workload recreate began; workload state is uncertain")
+
+// Applier is the canonical, security-critical path for applying an available
+// registry-sourced upgrade to a running workload. It re-resolves and verifies
+// the candidate image, pulls it, and only then asks the workload manager to
+// recreate the workload with the new image.
+//
+// Applier is the single place where an upgrade is materialized; both the CLI
+// (Phase D2) and the API (Phase D3) delegate here so that the verify-then-pull
+// ordering and TOCTOU guard live in exactly one place.
+type Applier struct {
+	manager   workloads.Manager
+	checker   *Checker
+	appConfig config.Provider
+
+	// resolveFn, enforcePullFn, and loadStateFn wrap the corresponding
+	// package-level functions in pkg/runner and pkg/runner/retriever. They are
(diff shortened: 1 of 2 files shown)


pkg/workloads/upgrade/applier.go
```diff
@@ -0,0 +1,439 @@
...
+	// Permission profile: if the workload pinned a named/path profile, preserve
+	// that name/path (do NOT re-pin to the resolved object or to the candidate's
+	// profile) so it stays consistent with what the checker reports as drift
+	// (the checker compares PermissionProfileNameOrPath). Otherwise preserve the
+	// resolved profile object.
+	if old.PermissionProfileNameOrPath != "" {
+		options = append(options, runner.WithPermissionProfileNameOrPath(old.PermissionProfileNameOrPath))
+	} else {
+		options = append(options, runner.WithPermissionProfile(old.PermissionProfile))
+	}
```
jhrozek, line 337: The doc comment on `buildUpgradedConfig` promises it "never mutates old or any of its maps/slices", but this branch hands `old.PermissionProfile` to the builder by pointer. When the workload has volumes, `processVolumeMounts` appends to `PermissionProfile.Read`/`.Write`, which mutates `old`'s profile through that shared pointer.

It's contained today because `old` is loaded internally and then discarded, but it breaks the stated contract, and `TestApplier_Apply_DoesNotMutateOldConfig` doesn't check the profile so it would go unnoticed. Cloning before handing it over keeps the guarantee:

```suggestion
	} else if old.PermissionProfile != nil {
		// Clone first: processVolumeMounts appends to Read/Write, which would
		// otherwise mutate old's profile through the shared pointer.
		cloned := *old.PermissionProfile
		cloned.Read = slices.Clone(old.PermissionProfile.Read)
		cloned.Write = slices.Clone(old.PermissionProfile.Write)
		options = append(options, runner.WithPermissionProfile(&cloned))
	} else {
		options = append(options, runner.WithPermissionProfile(old.PermissionProfile))
	}
```

`ToolsOverride` and `HeaderForward` are shared by reference in `preserveUserConfigFields` too — lower risk, but a `maps.Clone` there would let the "never mutates old" contract hold strictly.

pkg/workloads/upgrade/applier.go
```diff
@@ -0,0 +1,439 @@
...
+	// would double-gate the same config against the same policy. A failure here
+	// still leaves the running workload untouched.
+	if err := a.enforcePullFn(ctx, newConfig, serverMeta, imageURL, retriever.PullMCPServerImage, 0, false); err != nil {
+		return nil, fmt.Errorf("failed to prepare candidate image for workload %q: %w", name, err)
+	}
+
+	// 7. Destructive apply. The manager stops, deletes, and recreates the
+	// workload with the new config. We use a background context so an upgrade in
+	// progress is not aborted by request-scoped cancellation mid-recreate.
+	completion, err := a.manager.UpdateWorkload(context.Background(), name, newConfig)
```
jhrozek: `context.Background()` drops more than the cancellation signal — it also discards the OTel trace span and any request-scoped values, so the stop/delete/save/start goroutine runs with no trace lineage. `context.WithoutCancel(ctx)` keeps the values while still severing cancellation/deadline, which is what the comment is actually after:

```suggestion
	completion, err := a.manager.UpdateWorkload(context.WithoutCancel(ctx), name, newConfig)
```

The comment on the lines just above says "background context" — worth updating to match.

jhrozek (commented):
Two notes from re-reviewing the Applier (the rest of my findings I'm tracking elsewhere, or they belong to the stacked check PRs). Both are non-blocking — the verify-then-pull ordering and TOCTOU guard look solid.

pkg/workloads/upgrade/applier.go
JAORMX: Fixed in `01fbc6f3`. The `else if old.PermissionProfile != nil` branch now clones the profile before handing it to the builder (`cloned := *old.PermissionProfile` + `slices.Clone` of `Read`/`Write`), so `processVolumeMounts`'s appends no longer leak back into `old`. I also cloned `ToolsOverride` with `maps.Clone` in `preserveUserConfigFields` so the contract holds strictly.

`TestApplier_Apply_DoesNotMutateOldConfig` now sets a non-named profile plus a volume and asserts `Read`/`Write`/`ToolsOverride` are untouched — I verified it fails without the clone (the volume mount leaks into `old.PermissionProfile.Write`).

On `HeaderForward`: it's a `*HeaderForwardConfig` pointer that nothing in the build path mutates, so I left it shared rather than add a false-confidence shallow copy that would still share its inner maps. I added a comment noting the profile clone is intentionally shallow-except-`Read`/`Write` and that `Network` would need cloning too if a future change ever calls `WithNetworkMode` here.

pkg/workloads/upgrade/applier.go
JAORMX: Fixed in `01fbc6f3` — switched to `context.WithoutCancel(ctx)`, which keeps the trace span and request-scoped values while still severing cancellation/deadline. Updated the comment to describe the value-preservation rationale rather than "background context".

jhrozek (approved):

PR #5411, opened 2026-06-01
Add upgrade apply for the CLI and API

JAORMX:
## Summary

With the `Applier` in place (PR #5410), expose it to users so they can apply an upgrade while preserving configuration, instead of manually re-running a workload with a new image. Part of [RFC THV-0068](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0068-workload-upgrade.md), local scope.

- Add `thv upgrade apply <name>`: runs the check, shows the candidate image, new env vars, and any permission/transport/network posture drift, then prompts for confirmation. `--dry-run` reports the plan without applying; `--env`/`--secret` supply values for newly required variables; `--yes` (or a non-interactive shell) skips the prompt and fails loudly on missing required values; `--image-verification` mirrors `thv run`.
- Add `POST /api/v1beta/workloads/{name}/upgrade`, delegating to the **same** `Applier` so all clients share one apply path. The API path is always non-interactive (detached validator) and sources image verification from server config; the request body can only supply env/secret values — it cannot redirect the image or weaken verification. Apply failures return a sanitized 422 with the detailed cause logged server-side, so secret references in an error chain are never echoed to the caller.

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests (`task test`) — API: happy path, no-op, 404/400, no-secret-leak (asserts a secret reference in the request never appears in success or 422 responses), route ordering. CLI: output-formatting helper.

## Changes

| File | Change |
|------|--------|
| `cmd/thv/app/upgrade.go` | `apply` subcommand + flags + confirmation |
| `pkg/api/v1/workloads_upgrade.go` | `POST .../upgrade` handler |
| `pkg/api/v1/workloads.go`, `workload_types.go` | Route + request type |
| `docs/cli/*`, `docs/server/*` | Regenerated reference |

Yes — `thv upgrade apply` and `POST /api/v1beta/workloads/{name}/upgrade`. The command stops and replaces the existing workload (verifying and pulling the candidate image first); there is no automatic rollback, which the help text states.

## Large PR Justification

This PR is ~1,007 lines, but **only 330 are hand-written source**; the remainder is generated reference and tests:

- **374 lines are auto-generated** OpenAPI (`docs/server/*`) and CLI reference (`docs/cli/*`), produced by `task docs`. They must be regenerated and committed together (CI verifies them) and cannot be split.
- **303 lines are tests** for the new handler and CLI helper.
- The 330 source lines are the CLI `apply` subcommand and the API `POST` handler — **two thin callers of the same `Applier`** that the user explicitly chose to ship together so all clients share one apply path. Splitting CLI from API here would duplicate the wiring and the "all clients benefit" goal of the RFC, and both are well under the threshold individually.

**PR 5 of 6** in the RFC THV-0068 stack; based on #5410. The API path hardcodes the detached env-var validator and sources the verify mode from server config — a client cannot weaken verification or redirect the image via the request body.


pkg/api/v1/workloads_upgrade.go
```diff
@@ -88,6 +99,98 @@ func (s *WorkloadRoutes) upgradeCheckSingle(w http.ResponseWriter, r *http.Reque
...
+		// response: it may reference the request's secret parameters (e.g. an
+		// env/secret validation failure). Log the detailed cause server-side and
+		// return a sanitized, secret-free message to the caller. The log line
+		// carries only the error chain (which itself references secrets by name,
+		// never resolved values).
+		slog.Error("failed to apply workload upgrade", "workload", name, "error", err)
+		return httperr.WithCode(
+			fmt.Errorf("failed to apply upgrade for workload %q", name),
+			http.StatusUnprocessableEntity,
+		)
```
jhrozek, line 196: This maps every `Apply` error to 422, but the "the running workload is untouched on any of these" comment only holds for the preparation phase. Once `Apply` reaches the destructive recreate, two of its error paths fire after the workload has already been torn down:

- `UpdateWorkload` failing (stop/delete already initiated)
- `completion()` failing (old workload deleted, replacement didn't start)

In those cases a 422 is misleading: it tells the client the request couldn't be processed and nothing changed, when the workload may actually be gone and not running. A client seeing 422 will reasonably assume it's safe to retry against an intact workload, when it's really a 5xx-class state.

Could we distinguish the two? A typed/sentinel error out of `Apply` (e.g. preparation vs recreate) lets the handler return 422 for prep failures and 500 for recreate failures. At minimum it'd be worth fixing the comment so it doesn't claim "untouched" for the post-`UpdateWorkload` paths.

JAORMX: Good catch — fixed by distinguishing the two failure classes with a sentinel.

`Apply` now tags every error raised after the destructive recreate begins (the `UpdateWorkload` error path and the `completion()` error path) with `upgrade.ErrApplyAfterDestroy` via `errors.Join` (in #5410, `01fbc6f3`). The handler checks `errors.Is(err, upgrade.ErrApplyAfterDestroy)` and returns **500** for those — the workload state is uncertain — while preparation failures (resolve/verify/pull/build/validate) stay **422** since the running workload is untouched. The 422 comment no longer claims "untouched" for the post-`UpdateWorkload` paths.

The 5xx body is sanitized by `httperr` to a generic message, so the "state is uncertain" detail stays server-side in the log. Added `TestUpgradeWorkload/post-destruction failure maps to 500` here, plus applier-side `assert.ErrorIs` coverage on both post-destroy paths in #5410 so the sentinel tag can't be silently dropped.

jhrozek (approved):

Issue #5437, opened 2026-06-03
P1.4 New(cfg) -> VMCP core constructor


tgrunnagle:
## Description

Implement the concrete `*coreVMCP` and `New(cfg *Config) (VMCP, error)` constructor
in the root `pkg/vmcp` package by **relocating** (not rewriting) the domain wiring that
lives in `server.New` today. This is the structural heart of RFC Phase 1: it turns the
`VMCP` contract from #5434 into a working, identity-parameterized core that aggregates
capabilities, routes calls, and drives composite workflows — without any mcp-go types
crossing the boundary and without touching `server.New`.

## Context

The core constructor relocates the collaborator wiring currently buried in `server.New`'s
god-object body (`server.go:330-465`): telemetry backend-client decoration, the workflow
auditor, the in-memory state store + `NewWorkflowEngine`, the per-session
`sessionComposerFactory`/`NewSessionRouter` pattern, and `validateWorkflows` fail-fast.
Per `architecture.md` ("New (core) wiring relocated", lines ~278-284 and
"Statelessness, Session Consistency & Health Filtering (R5/R7)"), the **core filters,
Serve caches**: the core is stateless w.r.t. sessions, takes an injected
`health.StatusProvider`, and runs `filterHealthyBackends` internally over the backend
registry before aggregating — preserving today's exact include/exclude rules. The
`StatusProvider` is **built at the composition root** (the `server.New` wrapper / test
harness, per A2) and the same `*health.Monitor` is lifecycled by `Serve` (#5443); the
core only reads it, and a nil `StatusProvider` means no health filtering (all backends
included), matching today's no-monitor behavior. `ListTools`
drives `aggregator.AggregateCapabilities` on demand; `CallTool`/`ReadResource`/`GetPrompt`
route via `router.Router` + composer; `Lookup*` resolve advertised names (incl.
`BackendID`) without invoking. The admission seam is **NOT** wired here — it lands in
#5438. Identity is an explicit `*auth.Identity` on every method, never read from
context (anti-pattern #1).

**Parent Story**: #[#5430]
**Dependencies**: #[#5434] (VMCP interface + core Config), #[#5436] (domain-typed `ElicitationRequester`)
**Blocks**: #5438, #5439

> **Split note (pre-planned axis):** This PR is likely to exceed the **400 LOC / 10 file**
> limit — it relocates the whole `server.go:330-465` wiring *and* implements 10 interface
> methods *and* unit tests — so plan to split it along this axis:
[...]

Issue #5436, opened 2026-06-03
P1.3 Domain-typed ElicitationRequester (bounded rewrite)


tgrunnagle:
## Description

Replace `composer.SDKElicitationRequester` — which leaks mcp-go's `mcp.ElicitationRequest`/`mcp.ElicitationResult` across what is about to become the domain boundary — with a **domain-typed `ElicitationRequester`** interface plus `ElicitationRequest`/`ElicitationResult` value types that carry **no mcp-go types**. This is bounded rewrite **R4**: the core (added in #5437) drives composite-tool elicitation steps through this domain seam, and all mcp-go translation is confined to the adapter in `pkg/vmcp/server`. Without this, anti-pattern #5 (SDK coupling leaking through abstractions) would cross the new `VMCP` boundary the moment `New(cfg) -> VMCP` is introduced.

## Context

The `composer.DefaultElicitationHandler` already isolates elicitation behind an injected `SDKElicitationRequester`, but that interface's method signature uses mcp-go types directly (`elicitation_handler.go:80-84`). The architecture phase identified this as risk **R4**: it must become a domain-typed interface, translated to/from mcp-go **only** inside the existing `sdkElicitationAdapter` (`server/sdk_elicitation_adapter.go:46`). This is one of the two **bounded rewrites** in Phase 1 (the other is the admission seam, R1/#5438).

This is a **one-to-one field mirror** of the mcp-go types the composer already uses — **not** a redesign of elicitation semantics. URL-mode fields (`ElicitationID`/`URL`/`Mode`) are deliberately omitted because the composer does not construct them today (`elicitation_handler.go:168-173` builds form-mode requests only). The composer's `DefaultElicitationHandler` (`elicitation_handler.go:111`) keeps **all** of its security validation (timeout cap, schema size/depth, content size — `elicitation_handler.go:142-197`) unchanged; only the underlying requester type changes from mcp-go to the domain type.

See architecture.md, section "`ElicitationRequester` boundary (R4 — domain-typed, bounded rewrite)" (lines ~200-241), including the one-to-one mcp-go field-mapping table.

**Parent Story**: #<#5430>
**Dependencies**: None (root task; can start immediately)
**Blocks**: #5437 (`New(cfg) -> VMCP` core constructor — the core depends on the domain `ElicitationRequester`)

## Acceptance Criteria

[...]

Issue #5430, opened 2026-06-03
Phase 1: VMCP interface, core constructor, admission + elicitation seams


tgrunnagle:
## Description

Deliver the foundation of the vMCP domain/transport split: a public,
identity-parameterized `VMCP` domain interface and core `Config` in the root
`pkg/vmcp` package, a `New(cfg) -> VMCP` constructor that **relocates** (does not
rewrite) `server.New`'s domain wiring, and the two deliberate **bounded rewrites**
— the Cedar-backed admission seam (R1) and the domain-typed `ElicitationRequester`
(R4). This story maps to **RFC Phase 1** and moves the domain seam (aggregation,
routing, admission, elicitation) out from behind the transport god-object — while
`server.New` stays completely untouched.

## Context

See [RFC THV-0076: vMCP Core Interface](/Users/trey/Documents/GitHub/stacklok/toolhive-rfcs/rfcs/THV-0076-vmcp-core-interface.md)
for full design details. Part of the vMCP interface refactor (epic #5419).

This is the **first** story and the entry point for the whole refactor: it
establishes the `VMCP` contract that `Serve` (Phase 2) and the `server.New`
wrapper reduction (Phase 3) build on. Per `architecture.md` ("PR-Sized
Decomposition Guidance → Phase 1"), the new core is constructed by relocating the
collaborator wiring that lives in `server.New` today — aggregator, router, backend
registry, backend client, composer, the per-session
`sessionComposerFactory`/`NewSessionRouter` pattern, and workflow validation —
plus an injected `health.StatusProvider` so the core runs `filterHealthyBackends`
internally (core filters, Serve caches). Identity is an explicit `*auth.Identity`
on every method and is never read from context (anti-pattern #1); no mcp-go types
cross the `VMCP` boundary (anti-pattern #5).

Crucially, this story carries the **two highest-risk items in the entire epic** —
the two bounded rewrites that re-platform existing behavior rather than merely
relocating it:

- **Admission seam (#5438, R1) — HIGHEST regression risk.** Moves the
  authorization decision from the HTTP `AuthzMiddleware` into a core `Admission`
  seam that wraps the **existing** Cedar `authorizers.Authorizer` (no new policy
  model). It must produce the **same** decision in `List*` (filter) and
  `Call/Read/Get` (deny), re-inject the explicit identity into ctx for Cedar's
  `AuthorizeWithJWTClaims`, source `Tool.Annotations` directly from the core
[...]

Issue #5438, opened 2026-06-03
P1.5 Core admission seam (bounded rewrite)


tgrunnagle:
## Description

Add the `Admission` seam (bounded rewrite **R1**, the highest-regression-risk
change in this refactor): an `Admission` interface plus a Cedar-backed
implementation that wraps the **existing** `authorizers.Authorizer` (no new
policy model). Wire it into the `VMCP` core so `List*` (filter) and
`Call/Read/Get` (deny) enforce the **same** authorization decision — closing the
"list says yes / call says no" gap that today is only achievable because authz
runs as HTTP middleware on a single path. This re-platforms enforcement from two
middleware (`AuthzMiddleware` + `AnnotationEnrichmentMiddleware`) into the core
while preserving byte-for-byte authorization behavior.

## Context

Today authorization runs **only as HTTP middleware** (`Config.AuthzMiddleware`,
built by `factory.NewIncomingAuthMiddleware`, Cedar-based) and is applied
conditionally in `Handler` (`server.go:606`), and it **depends on the
annotation-enrichment middleware** running first to inject `Tool.Annotations`
into context so Cedar `when`-clauses (e.g. `resource.readOnlyHint`) evaluate.
The RFC requires that `ListTools` and `CallTool` enforce the same decision from
one source. Per architecture.md "Admission seam (R1)" (lines 140-198) and the
resolved decision in research.md (R1 → core admission seam), the chosen approach
is a core-level admission seam fed by the existing Cedar authorizer — reusing
the policy model and factory, not introducing a new one. The core already holds
the aggregated `Tool.Annotations`, so it sources them directly and the
annotation-enrichment middleware is retired on the domain path.

This task adds the seam and wires it into the core methods implemented in
#5437. The authz/annotation HTTP middleware stays in the live `server.New`
path until #5441 removes it (Phase 2) — so there is temporary
double-enforcement **only inside tests that exercise the core directly**, never
in the shipped `server.New` path.

**Parent Story**: ##5430
**Dependencies**: ##5437 (the `New(cfg) -> VMCP` core + method bodies the seam wires into)
**Blocks**: #5441 (removes the authz + annotation-enrichment HTTP middleware once this core seam exists)

## Acceptance Criteria

[...]

PR #5450, opened 2026-06-03
Define VMCP core interface and Config contract

### LINKED ISSUE #5434: P1.1 Define VMCP interface + core Config
## Description

Add the public, identity-parameterized `VMCP` domain interface and the core `Config`
struct to the root `pkg/vmcp` package (new file `pkg/vmcp/vmcp.go`). This is the
**contract-only** first step of the vMCP New/Serve split: it establishes the domain
boundary that `New(cfg) -> VMCP` (#5437) implements and that `Serve` and decorators
consume — with **no implementation yet**. Placing the interface in the root package,
alongside the existing `Tool`/`Resource`/`Prompt`/result types in `types.go`, lets it
reference those domain types directly without introducing an import cycle.

## Context

This is the very first PR of Phase 1 of RFC THV-0076 (the vMCP core-interface refactor),
which extracts a small, identity-parameterized domain object from today's god-object
`server.New`. The interface is the **primary contract** (architecture.md "API Contracts
→ VMCP interface", lines 84-138): every method takes an explicit `*auth.Identity` and
never reads identity from context (anti-pattern #1), and no mcp-go types cross the
boundary (anti-pattern #5). The method set is the non-session-scoped generalization of
the existing `Caller` interface (`pkg/vmcp/session/types/session.go:28`), which is the
prior art for the shape and the nil-identity/anonymous semantics. Per decision R2, `meta`
is **retained** on `CallTool`/`GetPrompt` (diverging from the RFC's illustrative
signature) to match `Caller` and preserve `_meta` forwarding through
`vmcp.BackendClient.CallTool` (`types.go:646`; the `BackendClient` interface is declared
at `types.go:642`). Per R6, `ResourceReadResult.Meta` may be
nil due to an SDK handler limitation. See research.md (organizational standards: stdlib
testing, SPDX headers, go-style) and architecture.md "Core Principles" for the full
rationale.

**Parent Story**: #5430
**Dependencies**: None (root; can start immediately)
**Blocks**: #5437

## Acceptance Criteria

- [ ] `VMCP` interface declared in `pkg/vmcp/vmcp.go` with `ListTools`, `CallTool`,
  `ListResources`, `ReadResource`, `ListPrompts`, `GetPrompt`, `LookupTool`,
  `LookupResource`, `LookupPrompt`, and `Close`, with identity as an explicit
  `*auth.Identity` parameter on every data method.
- [ ] `meta` (`map[string]any`) is retained on `CallTool` and `GetPrompt` (R2), with a
  doc note on the `_meta`-forwarding rationale.
- [ ] No mcp-go types appear anywhere in the interface signature (anti-pattern #5);
  the interface references only root-package domain types + `*auth.Identity`.
- [ ] The interface doc comment encodes the full behavioral contract: identity explicit
  and never read from context; nil identity = anonymous; decorators may only **subtract**
  reachability; `args`/`meta` maps are read-only (copy-before-mutate); and
  `ResourceReadResult.Meta` may be nil per the SDK handler limitation (R6).
- [ ] Core `Config` struct declared with its **typed fields** (collaborators,
  `workflowDefs`, the domain `ElicitationRequester`, `Authz`, cross-cutting
  `TelemetryProvider`/`AuditConfig`, and the injected `health.StatusProvider`) — field
  declarations only, no wiring/construction logic (#5437 fills the bodies). Typed (not
  comment-only) so the first PR doesn't draw an empty-struct lint nit (C4).
- [ ] Package compiles with no import cycle (`task build` green); SPDX header on the new
  `.go` file.

Create `pkg/vmcp/vmcp.go` in the **root** `pkg/vmcp` package and declare the `VMCP`
interface exactly as specified in architecture.md (lines 105-127), generalizing the
existing `Caller` method set to be non-session-scoped. Add the core `Config` struct as a
field-only declaration documenting the collaborators it will hold (aggregator, router,
backend registry, backend client, composer, workflow defs, the `Authz` config for the
later admission seam, plus cross-cutting `TelemetryProvider`/`AuditConfig`) — but **no
constructor body**: `New(cfg *Config) (VMCP, error)` is defined and implemented in
#5437, not here.

Declare the core `Config` with its **typed fields** (aggregator, router, backend registry,
backend client, the domain `ElicitationRequester`, workflow defs, the `Authz` config for
the later admission seam, the cross-cutting `TelemetryProvider`/`AuditConfig`, and the
injected `health.StatusProvider`) — field declarations only, **no constructor body**: `New`
is implemented in #5437. Typed fields (not a comment-only stub) mean #5437 fills in
*bodies* rather than the struct shape, and the first PR doesn't draw an empty-struct lint
nit (C4).

This PR is the **contract only**. Per the task scope, declare the `func New(cfg *Config)
(VMCP, error)` (and `Serve`) signatures as forward declarations **only if needed to make
the package compile**; otherwise omit them entirely and leave their bodies to #5437 /
#5439. Prefer not to add stubs that the linter will flag as unused — keep the surface
to the interface, its doc contract, and the `Config` type. The full doc comment on the
interface is a first-class deliverable here: it is the durable encoding of the contract
that downstream PRs must honor.

### Patterns & Frameworks

- **Domain-only boundary**: no mcp-go imports in `vmcp.go`. The interface speaks only in
  root-package domain types (`Tool`, `Resource`, `Prompt`, `ToolCallResult`,
  `ResourceReadResult`, `PromptGetResult`) and `*auth.Identity` (anti-pattern #5).
- **Identity explicit, never from context** on the public path (anti-pattern #1) —
  `ctx` is for cancellation/trace only.
- **Decorator extension model**: the doc contract states decorators hold only an
  `inner VMCP` and may only subtract reachability; they cannot widen access. Mirrors the
  list-filter/call-deny pairing in `pkg/authz/tool_filter.go`.
- **Conventions**: `.claude/rules/go-style.md` (SPDX header, doc comments, copy-before-
  mutating-caller-input), `.claude/rules/vmcp-anti-patterns.md`. Per R8, tests in this
  area use stdlib `testing.T` + testify — though this contract-only PR may need no new
  tests beyond a compile check.

### Code Pointers

- `pkg/vmcp/vmcp.go` *(new)* — the `VMCP` interface + core `Config`. Authoritative shape:
  architecture.md "API Contracts → VMCP interface" (lines 89-134).
- `pkg/vmcp/types.go:367,406,424` — `Tool`, `Resource`, `Prompt` structs the interface
  returns; `Tool.BackendID` (386), `Resource.BackendID` (420), `Prompt.BackendID` (435)
  — logical ids safe to expose; populated by the advertising filter (pinned in #5435).
- `pkg/vmcp/types.go:512,552,576` — `ToolCallResult`, `ResourceReadResult`,
  `PromptGetResult` returned by the call/read/get methods.
- `pkg/vmcp/types.go:556-561` — `ResourceReadResult.Meta` nil note (R6): the SDK
  resources/read handler cannot forward `_meta`; the interface doc must document this.
- `pkg/vmcp/types.go:646` — `vmcp.BackendClient.CallTool` (the `_meta`-forwarding path
  that motivates keeping `meta` on `CallTool`, R2); the `BackendClient` interface is
  declared at `types.go:642`.
- `pkg/vmcp/session/types/session.go:28` — `Caller` interface (prior art). `CallTool`
  with `meta` (41-48), `ReadResource` (58), `GetPrompt` (71), `Close` (80); the `VMCP`
  method set generalizes this to be non-session-scoped.
- `pkg/vmcp/session/types/session.go:182,224,228` — `ShouldAllowAnonymous`,
  `ErrUnauthorizedCaller`, `ErrNilCaller`: the nil-identity/anonymous semantics the doc
  contract references (the concrete behavior is reproduced by `New` in #5437).
- `pkg/vmcp/server/server.go:301` — `server.New` (7-param signature); **stays stable** and
  is **not touched** by this PR.
- `github.com/stacklok/toolhive/pkg/auth` — `auth.Identity` (the explicit identity param;
  redacts `Token`/`UpstreamTokens`, so it is never logged — Core Principle #5).

### Component Interfaces

The exact contract to declare (root `pkg/vmcp` package). `meta` is **retained** on
`CallTool`/`GetPrompt` per R2. The doc comment shown is the load-bearing deliverable.

```go
// VMCP is the core Virtual MCP domain object.
//
// Contract:
//   - Identity is an explicit parameter on every method and is NEVER read from
//     context (anti-pattern #1). A nil identity is anonymous; bound-identity
//     mismatch is the caller's concern only at the session layer (in Serve), not
//     here — the core takes an already-authenticated *auth.Identity.
//   - Implementations MUST be safe for concurrent use.
//   - Decorators may only SUBTRACT reachability: filter list output or refuse a
//     call before delegating to inner. They have no path to backends except
//     through inner, so they cannot widen access.
//   - args/meta maps are treated as read-only; the core copies before mutating.
//   - ReadResource results: Meta may be nil — the mcp-go resources/read handler
//     cannot forward _meta (see types.go:556-561). Do not rely on it (R6).
type VMCP interface {
	ListTools(ctx context.Context, identity *auth.Identity) ([]Tool, error)
	CallTool(ctx context.Context, identity *auth.Identity, name string,
		args map[string]any, meta map[string]any) (*ToolCallResult, error)

	ListResources(ctx context.Context, identity *auth.Identity) ([]Resource, error)
	ReadResource(ctx context.Context, identity *auth.Identity, uri string) (*ResourceReadResult, error)

	ListPrompts(ctx context.Context, identity *auth.Identity) ([]Prompt, error)
	GetPrompt(ctx context.Context, identity *auth.Identity, name string,
		args map[string]any) (*PromptGetResult, error)

	// Lookup* resolve an advertised name/URI to the capability (incl. BackendID)
	// WITHOUT invoking it. Returns an error for an unknown/unadvertised name —
	// the validation seam for the call path. Lookups apply the same admission
	// filter as List*, so they never resolve a denied capability.
	LookupTool(ctx context.Context, identity *auth.Identity, name string) (*Tool, error)
	LookupResource(ctx context.Context, identity *auth.Identity, uri string) (*Resource, error)
	LookupPrompt(ctx context.Context, identity *auth.Identity, name string) (*Prompt, error)

	// Close releases core-held resources (backend connections, etc.). Idempotent.
	Close() error
}

// Config holds the collaborators New assembles into the core (typed fields declared
// here; New's body lands in #5437). Cross-cutting TelemetryProvider/AuditConfig are
// consumed by both New and Serve (not a clean partition, R3); HealthStatusProvider is
// the read-only health view built at the composition root (A2; nil => no filtering).
type Config struct {
	Aggregator           aggregator.Aggregator
	Router               router.Router
	BackendRegistry      vmcp.BackendRegistry
	BackendClient        vmcp.BackendClient
	WorkflowDefs         map[string]*composer.WorkflowDefinition
	Elicitation          ElicitationRequester  // domain-typed (#5436)
	Authz                *authz.Config          // feeds the admission seam (#5438)
	TelemetryProvider    *telemetry.Provider    // cross-cutting (also on ServerConfig, R3)
	AuditConfig          *audit.Config          // cross-cutting (also on ServerConfig, R3)
	HealthStatusProvider health.StatusProvider  // injected; built at the composition root (A2)
	// Exact field set is finalized by New (#5437) and deriveCoreConfig (#5444);
	// this PR declares the shape (typed fields), not the construction logic.
}

// Implemented in #5437; declared here only if required to compile.
// func New(cfg *Config) (VMCP, error)
```

### Testing Strategy

**Unit Tests**
- [ ] Compile-time conformance is the primary guarantee for a contract-only PR
  (`task build` / `task test` green). No behavior to exercise yet — `New` is unimplemented.
- [ ] (Optional) A trivial compile-only assertion that a `nil`/placeholder value can be
  typed as `VMCP`, if it adds clarity without introducing a dead stub.

- [ ] None required for this PR — `server.New` is untouched and the core is not yet wired
  into the live path; the parity suite remains green unchanged.

**Edge Cases**
- [ ] Verify no import cycle is introduced between `pkg/vmcp` (root) and `types.go`/
  `pkg/auth` (the reason the interface lives in the root package).
- [ ] Confirm zero mcp-go imports in `vmcp.go` (grep guard) — enforces anti-pattern #5 at
  the boundary from day one.

## Out of Scope

- **No `New` implementation** — the `*coreVMCP` concrete type, collaborator wiring, health
  filtering, and method bodies land in #5437.
- **No `Serve`/`ServerConfig` implementation** — #5439.
- **No admission seam** (`Admission` interface / Cedar wiring) — #5438.
- **No domain `ElicitationRequester`** — #5436.
- No changes to `server.New`, `cli/serve.go`, `vmcpconfig.Config`, the CRD/YAML model, or
  any wire/storage format.
- No BackendID population changes (the population contract is pinned by tests in #5435).

## References

- `architecture.md` "API Contracts → VMCP interface" (lines 84-138), "Core Principles"
- `research.md` (organizational standards, codebase patterns)
- Parent story: #5430
- Epic: #5419

tgrunnagle:
## Summary

RFC THV-0076 extracts a small, identity-parameterized domain object from today's vMCP transport god-object (`server.New`), so that aggregation, routing, and composite workflows live behind a single explicit contract instead of being entangled with HTTP/session concerns. Before any of that machinery can be built, the domain boundary itself has to exist and be agreed on — that is what this PR delivers.

This is the **contract-only** first PR of Phase 1. It declares:

- The identity-parameterized `VMCP` domain interface (`ListTools`/`CallTool`, `ListResources`/`ReadResource`, `ListPrompts`/`GetPrompt`, the `Lookup*` resolution methods, and `Close`), with its full behavioral contract encoded in the doc comment.
- The core `Config` struct of collaborators as typed field declarations.

There is **no implementation**: `New(cfg) (VMCP, error)` and the concrete `coreVMCP` bodies land in #5437. The deliverable here is the durable contract that #5437 implements and that `Serve` and the decorators downstream consume.

Closes #5434

## Type of change

## Test plan

This is a contract-only change, so compile-time conformance is the guarantee per the issue's testing strategy — there is no behavior to exercise until `New` is implemented (#5437). Verification performed:

- `task build` green — proves the package compiles with no import cycle.
- `task test` green (exit 0, no races) — full suite unchanged; `server.New` is untouched.
- The new package lints clean. The one `task lint` finding is a pre-existing gosec G115 in `cmd/thv/app/upgrade.go`, unrelated to this PR and not in this diff.

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/core/core.go` | New file (156 lines): the `VMCP` interface with its full contract doc comment, and the typed `Config` collaborator struct. |

No.

**Package location differs from the issue text (user-approved).** The issue specified placing the interface and `Config` in the root `pkg/vmcp` package (`pkg/vmcp/vmcp.go`). However, `Config` references the aggregator, router, composer, health, and authz collaborators, and all of those transitively import the root `pkg/vmcp` package — verified empirically with `go list -deps`. A root-package `Config` would therefore create an import cycle. After confirming this, the interface **and** `Config` were placed in a new `pkg/vmcp/core` subpackage instead. This also resolves the same latent cycle that #5437's `New`/`coreVMCP` would otherwise hit. The issue text and #5437/#5430 likely need a follow-up note to reflect this package location.

**Two other user-approved deviations from the issue's literal spec:**

- The `Config.Elicitation ElicitationRequester` field is **omitted** here because its domain type is owned by #5436 and is out of scope for this PR. A doc comment marks where it will be added.
- `meta` is on `CallTool` only, not `GetPrompt`. This matches the authoritative interface signature, the `Caller` prior art, and `vmcp.BackendClient.GetPrompt` (which has no `meta` parameter — there is no client `_meta` to forward on the prompts/get path). AC#2's prose mentioning `GetPrompt` is a wording discrepancy in the issue.

**Context:** Parent story #5430; this PR blocks #5437.

### DIFF
--- a/pkg/vmcp/core/core.go
+++ b/pkg/vmcp/core/core.go
@@ -39,9 +39,8 @@ import (
 //     context (anti-pattern #1). A nil identity is anonymous; bound-identity
 //     mismatch is the caller's concern only at the session layer (in Serve), not
 //     here — the core takes an already-authenticated *auth.Identity. The
-//     nil-identity/anonymous semantics mirror
-//     [pkg/vmcp/session/types.ShouldAllowAnonymous]; the concrete behavior is
-//     reproduced by New.
+//     nil-identity/anonymous semantics mirror session/types.ShouldAllowAnonymous;
+//     the concrete behavior is reproduced by New.
 //   - Implementations MUST be safe for concurrent use.
 //   - Decorators may only SUBTRACT reachability: filter list output or refuse a
 //     call before delegating to inner. They hold only an inner VMCP and have no
@@ -50,7 +49,7 @@ import (
 //   - args/meta maps are treated as read-only; the core copies before mutating
 //     (go-style: copy before mutating caller input).
 //   - ReadResource results: Meta may be nil — the mcp-go resources/read handler
-//     cannot forward _meta (see [vmcp.ResourceReadResult] and types.go:556-561).
+//     cannot forward _meta (see the NOTE on [vmcp.ResourceReadResult].Meta).
 //     Do not rely on it.
 //
 // ctx is used only for cancellation, deadlines, and trace propagation — never to
--- a/pkg/vmcp/core/core.go
+++ b/pkg/vmcp/core/core.go
@@ -1,0 +1,156 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package core defines the Virtual MCP domain object: the identity-parameterized
+// [VMCP] interface and the [Config] of collaborators that New assembles into a
+// working core.
+//
+// The core is the domain boundary extracted from today's transport god-object
+// (server.New): it aggregates backend capabilities, routes calls, and drives
+// composite workflows behind a single, identity-explicit contract. It lives in
+// its own package — rather than the root pkg/vmcp package — because Config
+// references the aggregator, router, composer, health, and authz collaborators,
+// all of which import pkg/vmcp; placing the core here lets it depend on both the
+// root domain types and those collaborators without an import cycle.
+//
+// No mcp-go types cross this boundary (vmcp anti-pattern #5): the interface speaks
+// only in root pkg/vmcp domain types ([vmcp.Tool], [vmcp.Resource], [vmcp.Prompt],
+// and their result wrappers) plus an explicit [*auth.Identity].
+package core
+
+import (
+	"context"
+
+	"github.com/stacklok/toolhive/pkg/audit"
+	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/authz"
+	"github.com/stacklok/toolhive/pkg/telemetry"
+	"github.com/stacklok/toolhive/pkg/vmcp"
+	"github.com/stacklok/toolhive/pkg/vmcp/aggregator"
+	"github.com/stacklok/toolhive/pkg/vmcp/composer"
+	"github.com/stacklok/toolhive/pkg/vmcp/health"
+	"github.com/stacklok/toolhive/pkg/vmcp/router"
+)
+
+// VMCP is the core Virtual MCP domain object.
+//
+// Contract:
+//   - Identity is an explicit parameter on every method and is NEVER read from
+//     context (anti-pattern #1). A nil identity is anonymous; bound-identity
+//     mismatch is the caller's concern only at the session layer (in Serve), not
+//     here — the core takes an already-authenticated *auth.Identity. The
+//     nil-identity/anonymous semantics mirror
+//     [pkg/vmcp/session/types.ShouldAllowAnonymous]; the concrete behavior is
+//     reproduced by New.
+//   - Implementations MUST be safe for concurrent use.
+//   - Decorators may only SUBTRACT reachability: filter list output or refuse a
+//     call before delegating to inner. They hold only an inner VMCP and have no
+//     path to backends except through it, so they cannot widen access. This
+//     mirrors the list-filter/call-deny pairing in pkg/authz/tool_filter.go.
+//   - args/meta maps are treated as read-only; the core copies before mutating
+//     (go-style: copy before mutating caller input).
+//   - ReadResource results: Meta may be nil — the mcp-go resources/read handler
+//     cannot forward _meta (see [vmcp.ResourceReadResult] and types.go:556-561).
+//     Do not rely on it.
+//
+// ctx is used only for cancellation, deadlines, and trace propagation — never to
+// carry identity or other domain data.
+type VMCP interface {
+	// ListTools returns the tools advertised to identity. The returned
+	// [vmcp.Tool] values carry their logical BackendID.
+	ListTools(ctx context.Context, identity *auth.Identity) ([]vmcp.Tool, error)
+
+	// CallTool invokes the named tool on behalf of identity.
+	//
+	// args contains the tool input parameters. meta carries protocol-level
+	// metadata (_meta) forwarded from the client; it is retained on this method
+	// (rather than dropped) so the core can forward it through
+	// [vmcp.BackendClient.CallTool] to the backend MCP server. Both maps are
+	// treated as read-only.
+	CallTool(ctx context.Context, identity *auth.Identity, name string,
+		args map[string]any, meta map[string]any) (*vmcp.ToolCallResult, error)
+
+	// ListResources returns the resources advertised to identity.
+	ListResources(ctx context.Context, identity *auth.Identity) ([]vmcp.Resource, error)
+
+	// ReadResource reads the resource at uri on behalf of identity. The
+	// returned result's Meta may be nil (see the interface contract).
+	ReadResource(ctx context.Context, identity *auth.Identity, uri string) (*vmcp.ResourceReadResult, error)
+
+	// ListPrompts returns the prompts advertised to identity.
+	ListPrompts(ctx context.Context, identity *auth.Identity) ([]vmcp.Prompt, error)
+
+	// GetPrompt retrieves the named prompt on behalf of identity.
+	//
+	// args contains the prompt input parameters and is treated as read-only.
+	// Unlike CallTool there is no meta parameter: prompts/get carries no client
+	// _meta to forward.
+	GetPrompt(ctx context.Context, identity *auth.Identity, name string,
+		args map[string]any) (*vmcp.PromptGetResult, error)
+
+	// LookupTool resolves an advertised tool name to its capability (incl.
+	// BackendID) WITHOUT invoking it. Returns an error for an unknown or
+	// unadvertised name — the validation seam for the call path. Lookups apply
+	// the same admission filter as ListTools, so they never resolve a denied
+	// capability.
+	LookupTool(ctx context.Context, identity *auth.Identity, name string) (*vmcp.Tool, error)
+
+	// LookupResource resolves an advertised resource URI to its capability
+	// (incl. BackendID) WITHOUT reading it. Returns an error for an unknown or
+	// unadvertised URI. Applies the same admission filter as ListResources.
+	LookupResource(ctx context.Context, identity *auth.Identity, uri string) (*vmcp.Resource, error)
+
+	// LookupPrompt resolves an advertised prompt name to its capability (incl.
+	// BackendID) WITHOUT retrieving it. Returns an error for an unknown or
+	// unadvertised name. Applies the same admission filter as ListPrompts.
+	LookupPrompt(ctx context.Context, identity *auth.Identity, name string) (*vmcp.Prompt, error)
+
+	// Close releases core-held resources (backend connections, etc.).
+	// Implementations must be idempotent: calling Close multiple times returns nil.
+	Close() error
+}
+
+// Config holds the collaborators New assembles into the core. The fields are
+// declared here as the contract; New's body (which wires them into a concrete
+// *coreVMCP) lands in a later change.
+//
+// Cross-cutting TelemetryProvider/AuditConfig are consumed by both New and Serve
+// (not a clean partition). HealthStatusProvider is the read-only health view
+// built at the composition root and injected here; a nil provider means no health
+// filtering (all backends included), matching today's no-monitor behavior.
+type Config struct {
+	// Aggregator discovers and merges backend capabilities into the advertised set.
+	Aggregator aggregator.Aggregator
+
+	// Router resolves an advertised capability to its backend target.
+	Router router.Router
+
+	// BackendRegistry enumerates the configured backends.
+	BackendRegistry vmcp.BackendRegistry
+
+	// BackendClient performs MCP protocol calls against backend servers.
+	BackendClient vmcp.BackendClient
+
+	// WorkflowDefs holds the composite-tool workflow definitions, keyed by name.
+	WorkflowDefs map[string]*composer.WorkflowDefinition
+
+	// Authz feeds the admission seam (wired in a later change). A nil Authz
+	// means authorization is unconfigured (allow-all).
+	Authz *authz.Config
+
+	// TelemetryProvider is the cross-cutting telemetry provider (also consumed by Serve).
+	TelemetryProvider *telemetry.Provider
+
+	// AuditConfig is the cross-cutting audit configuration (also consumed by Serve).
+	AuditConfig *audit.Config
+
+	// HealthStatusProvider is the read-only backend health view built at the
+	// composition root. Nil means no health filtering (all backends included).
+	HealthStatusProvider health.StatusProvider
+
+	// Elicitation (the domain-typed ElicitationRequester) is added once that
+	// domain type is introduced; it is intentionally omitted here.
+
+	// The exact field set is finalized when New is implemented; this change
+	// declares the shape (typed fields), not the construction logic.
+}


pkg/vmcp/core/core.go
```diff
@@ -0,0 +1,156 @@
...
+//     mismatch is the caller's concern only at the session layer (in Serve), not
+//     here — the core takes an already-authenticated *auth.Identity. The
+//     nil-identity/anonymous semantics mirror
+//     [pkg/vmcp/session/types.ShouldAllowAnonymous]; the concrete behavior is
+//     reproduced by New.
+//   - Implementations MUST be safe for concurrent use.
+//   - Decorators may only SUBTRACT reachability: filter list output or refuse a
+//     call before delegating to inner. They hold only an inner VMCP and have no
+//     path to backends except through it, so they cannot widen access. This
+//     mirrors the list-filter/call-deny pairing in pkg/authz/tool_filter.go.
```
tgrunnagle, line 48: **[LOW] godoc link won't render.** `[pkg/vmcp/session/types.ShouldAllowAnonymous]` uses a full import path qualified with a symbol, and `pkg/vmcp/session/types` is not imported by this file — so godoc cannot resolve it and it renders as literal bracketed text rather than a hyperlink. (The other links here — `[vmcp.Tool]`, `[*auth.Identity]`, `[VMCP]` — are fine because those packages are imported or same-package.)

Since the package isn't imported, plain prose is the cleanest fix:

```suggestion
//     here — the core takes an already-authenticated *auth.Identity. The
//     nil-identity/anonymous semantics mirror session/types.ShouldAllowAnonymous;
//     the concrete behavior is reproduced by New.
```

pkg/vmcp/core/core.go
```diff
@@ -0,0 +1,156 @@
...
+//   - args/meta maps are treated as read-only; the core copies before mutating
+//     (go-style: copy before mutating caller input).
+//   - ReadResource results: Meta may be nil — the mcp-go resources/read handler
+//     cannot forward _meta (see [vmcp.ResourceReadResult] and types.go:556-561).
+//     Do not rely on it.
+//
+// ctx is used only for cancellation, deadlines, and trace propagation — never to
+// carry identity or other domain data.
+type VMCP interface {
+	// ListTools returns the tools advertised to identity. The returned
```
tgrunnagle, line 58: **[LOW] Hard-coded line numbers will rot.** `types.go:556-561` is accurate today, but the citation drifts silently the moment anything is inserted above it in `types.go`, after which it points a reader at the wrong place (go-style: "Keep Comments Synchronized With Code"). The `[vmcp.ResourceReadResult]` doc-link on the same line is already self-maintaining and sufficient.

```suggestion
//   - ReadResource results: Meta may be nil — the mcp-go resources/read handler
//     cannot forward _meta (see the NOTE on [vmcp.ResourceReadResult].Meta).
//     Do not rely on it.
```

*Raised independently by all three reviewers.*

tgrunnagle (commented):
## Review summary

Clean, well-scoped **contract-only** PR: a single 156-line new file (`pkg/vmcp/core/core.go`) declaring the identity-parameterized `VMCP` interface and a typed `Config` collaborator struct, with no implementation. All ten acceptance-criteria methods are present, identity is an explicit `*auth.Identity` on every data method (anti-pattern #1), and zero mcp-go types cross the boundary (anti-pattern #5). The interface is a faithful non-session-scoped generalization of the prior-art `Caller` interface.

**Verified against the codebase:**
- The package-location deviation (`pkg/vmcp/core` instead of the issue's root `pkg/vmcp`) is **justified** — `aggregator`, `router`, `composer`, and `health` all import root `pkg/vmcp`, so a root-package `Config` referencing them would create an import cycle. Confirmed empirically.
- All referenced types resolve (`vmcp.Tool/Resource/Prompt`, the result wrappers, `BackendRegistry`, `BackendClient`, and every `Config` collaborator type).
- `meta`-on-`CallTool`-only (not `GetPrompt`) matches both the `Caller` prior art and `BackendClient.GetPrompt` (which has no `meta` param). The approved deviation from AC#2's wording is sound.
- Omitting the `Elicitation` field is correct — no domain `ElicitationRequester` type exists yet, and the field is purely additive on a struct with zero consumers.

**Three minor doc nits** (all non-blocking, see inline comments).

| # | Severity | Finding | Raised by |
|---|----------|---------|-----------|
| 1 | LOW | Doc comment hard-codes `types.go:556-561` line numbers that will rot | 3 reviewers |
| 2 | LOW | godoc link `[pkg/vmcp/session/types.ShouldAllowAnonymous]` won't render (package not imported) | 1 reviewer |

**Doc reconciliation (no code change):** Issue #5434's acceptance-criteria checkboxes reference `pkg/vmcp/vmcp.go`; since the canonical location is now `core.VMCP`, the issue and downstream PRs #5437/#5439 should be annotated so the `core` package is recognized as the single front door.

**Notes for downstream PRs** (contract claims this PR cannot enforce — flag during implementation review):
- #5437 must actually copy `args`/`meta` before mutating (the read-only guarantee is doc-only here), and `List*`/`Lookup*` must share a single admission filter so a denied capability is invisible to both paths.
- #5438's admission decorator constructor should take only an inner `VMCP` (no backend-reaching collaborators) to keep the "decorators may only subtract reachability" invariant structurally true.
- Confirm the `composer` collaborator is constructed internally by `New` from `WorkflowDefs` rather than injected; if injected, add the field now to avoid a later `Config` shape change.

**Recommendation: APPROVE** — no blocking issues; the inline items are optional polish.

pkg/vmcp/core/core.go
tgrunnagle: Addressed in b9638f9aea85a1c0d803efb955eb343a6620b843.

pkg/vmcp/core/core.go
tgrunnagle: Addressed in b9638f9aea85a1c0d803efb955eb343a6620b843.

jerm-dro (approved):

Issue #5442, opened 2026-06-03
P2.4 Replace discovery-into-context with direct VMCP calls


tgrunnagle:
## Description

Wire the transport to drive capability aggregation through the core's explicit-identity
`VMCP` methods instead of the discovery-middleware-into-`context` path. On the `Serve`
path, `Serve` calls `core.ListTools(ctx, identity)` **once per MCP session** at
`OnRegisterSession` and **feeds that result into the session factory** as its
pre-queried capability set — so `core.ListTools` is the single source of truth, the
session factory stops aggregating independently, and `Serve` owns the "fixed at
initialize" cache while the core stays stateless. The discovery middleware is **guarded
to the legacy `server.New` path** (not deleted) so the still-live path is unaffected;
its physical removal — along with the `handleSubsequentRequest` routing-table injection
and the `DiscoveredCapabilities` context seam — is deferred to Phase 3 (#5445).

## Context

This is step P2.4 of Phase 2 and the phase's join point: it consumes the `Serve`
SDK-hook wiring (#5440) and the relocated HTTP middleware chain (#5441).

**Why guard, not delete (the A1 sequencing constraint).** `(*Server).Handler` is shared
by both `server.New`'s `*Server` and a `Serve`-built `*Server`, and the discovery
middleware is applied **unconditionally** today (`server.go:633-637`). In Phase 2,
`server.New` is **not yet** routed through `Serve` (Phase 3 / #5445), so deleting the
discovery middleware from the shared `Handler` would break capability aggregation on the
live `server.New` path. Instead, this task adds a guard so discovery is applied **only on
the legacy path** (`s.core == nil`); on the `Serve` path (`s.core != nil`), discovery is
skipped and the transport calls the core directly. The discovery middleware,
`handleSubsequentRequest`'s routing-table injection, and the
`WithDiscoveredCapabilities`/`DiscoveredCapabilitiesFromContext` seam are physically
removed in Phase 3 once the legacy path is gone.

**Source-of-truth (A3 — explicit wiring).** The advertised per-session set is built by
the session-factory path (`MultiSession` → `aggregator.ProcessPreQueriedCapabilities`,
`session/factory.go`). To avoid double-aggregation **and** drift, `core.ListTools` is the
authoritative aggregation and **feeds** the factory: at `OnRegisterSession` on the
`Serve` path, `Serve` calls `core.ListTools(ctx, identity)` once and passes that slice as
[...]

Issue #5439, opened 2026-06-03
P2.1 Serve skeleton + ServerConfig


tgrunnagle:
## Description

Introduce the transport-side entry point of the New/Serve split: add
`Serve(ctx, v VMCP, cfg *ServerConfig) (*Server, error)` and the transport-only
`ServerConfig` struct in a new file `pkg/vmcp/server/serve.go`. `Serve` returns a
`*Server` wrapping the **existing** struct and builds the mcp-go server, the HTTP
route mux, and the HTTP lifecycle skeleton. This is the foundation the rest of
Phase 2 builds on — `server.New` is **not** yet routed through `Serve`, so observable
behavior is unchanged.

## Context

Part of the vMCP interface refactor (epic #5419), which extracts an
identity-parameterized `VMCP` domain interface from the `server.New` god-object and
splits it into a core (`New(cfg) -> VMCP`) and a transport (`Serve(ctx, VMCP,
serverCfg) -> *Server`). This task lands the `Serve` skeleton + `ServerConfig`; the
remaining transport concerns (SDK hooks, two-phase session creation, the middleware
chain, AS runner, status reporter, optimizer, health monitor) are relocated under
`Serve` by the subsequent Phase 2 tasks (#5440, #5441, #5443, #5442).

The guiding principle is **relocate, don't rewrite**: `Serve` returns the *same*
`*Server` struct, and all existing `(*Server)` accessors are carried forward
unchanged. See architecture.md "ServerConfig (R3)" (lines 243-274) and "Serve
(transport) wiring relocated" (lines 285-290); research.md "`Config` struct"
field-placement table (lines 144-167) for which fields land in `ServerConfig`.
`ServerConfig` is justified per go-style (which cautions against premature config
types) because these fields configure the HTTP/SDK runtime and are meaningless to
the core (R3).

**Parent Story**: #5431
**Dependencies**: #5437 (cross-story; needs `VMCP`/`New` from `pkg/vmcp`)
**Blocks**: #5440, #5441, #5443

## Acceptance Criteria

- [ ] `Serve(ctx context.Context, v VMCP, cfg *ServerConfig) (*Server, error)` and the
  `ServerConfig` struct (transport-only fields) exist in `pkg/vmcp/server/serve.go`.
- [ ] `Serve` builds the mcp-go server (`server.NewMCPServer`), the HTTP route mux, and
  the HTTP lifecycle skeleton, and returns a `*Server` wrapping the existing struct.
[...]

Issue #5419, opened 2026-06-02
vMCP interface refactor


tgrunnagle:
Parent issue tracking the implementation of the vMCP interface refactor described in https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0076-vmcp-core-interface.md.

PR #5457, opened 2026-06-04
Implement stateless core VMCP constructor

### LINKED ISSUE #5437: P1.4 New(cfg) -> VMCP core constructor

tgrunnagle:
## Summary

RFC Phase 1 of the vMCP refactor (epic #5419, story #5430) extracts a stateless, identity-explicit domain core out of the `server.New` god-object. This turns the `VMCP` contract from #5434 into a working core that aggregates backend capabilities, routes calls, and drives composite-tool workflows — without mcp-go types crossing the boundary and without changing `server.New`'s signature or observable behavior.

- **Why**: The domain wiring (telemetry decoration, workflow engine, composer factory, health filtering, workflow validation) is currently buried in `server.New`'s body, coupling the domain logic to the transport god-object and to context-injected capabilities. Phase 1 relocates that wiring behind a clean, identity-parameterized contract so later phases can re-home the transport (`Serve`, #5439) and wire the Cedar admission seam (#5438) without untangling it again.
- **What**:
  - New `pkg/vmcp/core` implements `*coreVMCP` and `New(cfg *Config) (VMCP, error)` with all 10 interface methods, relocating (not rewriting) the domain wiring from `server.New`: telemetry backend-client decoration, the optional workflow auditor, the in-memory state store + workflow engine, the per-call composer factory, and fail-fast workflow validation.
  - The core is stateless with respect to sessions: `List`/`Lookup`/`Call` health-filter the registry and aggregate on demand, then route through a per-call `router.NewSessionRouter` bound to the freshly aggregated routing table. This removes composite tools' dependency on context-injected capabilities (vmcp anti-pattern #1). The core filters; `Serve` caches (a later decorator).
  - An injected `health.StatusProvider` drives `filterHealthyBackends` before aggregation, preserving today's exact include/exclude rules (healthy/degraded/empty included; unhealthy/unknown/unauthenticated excluded). A nil provider means no filtering (all backends included), matching today's no-monitor behavior.
  - Composite tools are advertised in `ListTools` (the core is the only layer that can add them, since decorators may only subtract reachability) and executed in `CallTool` through the relocated per-session composer pattern.
  - `Close()` stops the state-store cleanup goroutine and is idempotent (`sync.Once`).
  - The backend-client telemetry decorator moved from `pkg/vmcp/server/telemetry.go` to a new shared leaf package `pkg/vmcp/internal/backendtelemetry` (exported as `MonitorBackends`) so both `server` and `core` can use it without an import cycle.
  - Identity is an explicit `*auth.Identity` on every method, never read from context and never logged. The Cedar admission seam (#5438) is intentionally not wired here.

Closes #5437

## Type of change

## Test plan

Unit tests for `pkg/vmcp/core` cover: constructor validation (loud failure on nil required collaborators), on-demand aggregation, health-filtering parity with `discovery/middleware.go`, lookups (including renamed/prefixed capability names and unknown-name errors), copy-before-mutate of `args`/`meta`, composite-tool execution across all `executeComposite` branches, idempotent `Close`, and verification that identity is never logged. Existing `pkg/vmcp/server` and the relocated `pkg/vmcp/internal/backendtelemetry` tests pass unchanged. `golangci-lint` reports 0 issues for the changed packages.

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/core/core_vmcp.go` | New: `*coreVMCP`, `New` constructor, list/lookup paths, health filtering, workflow validation, idempotent `Close`. |
| `pkg/vmcp/core/core_calls.go` | New: `CallTool`/`ReadResource`/`GetPrompt` call paths and composite-tool execution. |
| `pkg/vmcp/core/core.go` | Wire `New` to the previously contract-only `Config`. |
| `pkg/vmcp/elicitation.go` | New: domain-typed `ElicitationRequester` seam (no mcp-go types). |
| `pkg/vmcp/internal/backendtelemetry/` | Relocated backend-client telemetry decorator from `server/telemetry.go`; exported as `MonitorBackends`. |
| `pkg/vmcp/server/server.go` | Call the relocated `backendtelemetry.MonitorBackends`; signature and behavior unchanged. |
| `pkg/vmcp/composer/`, `pkg/vmcp/server/sdk_elicitation_adapter*.go` | Consume the domain-typed `ElicitationRequester`; SDK translation stays confined to the adapter. |

No. The core is additive and not yet wired into the request path; `server.New`'s signature and observable behavior are unchanged.

- **Approved scope exception (exceeds the 400-LOC criterion).** This PR intentionally implements both pre-planned sub-tasks (TASK-104a list/lookup/constructor + TASK-104b call/read/get) in one change, totaling roughly 580 net-new production LOC across 4 production files — over the 400-LOC acceptance criterion but within the 10-file limit. This was an explicit, up-front decision; the alternative was a 104a PR plus a 104b follow-up.
- **Intentional temporary duplication (C2).** `filterHealthyBackends` is a deliberate copy of `discovery/middleware.go`'s implementation. The discovery middleware keeps its own copy on the legacy `server.New` path until that path is removed in Phase 3 (#5442/#5445). The include/exclude rules are kept identical across both copies — this is not accidental duplication.
- **Pre-existing, unrelated findings observed during verification (not introduced by this PR):**
  - The only repo-wide `task lint-fix` finding is a pre-existing gosec G115 in `cmd/thv/app/upgrade.go` (the `.golangci.yml` gosec exclude list omits G115).
  - The only `task test` failures are `pkg/mcp/server/TestBuildServerConfig`, which require a container runtime (Docker/Podman) not present locally.

## Large PR Justification

This PR is large for three reasons, none of which are net-new production logic:

  - **Atomic full-scope implementation (multiple related changes that would break if separated).**
    Issue #5437 pre-planned a split into TASK-104a (constructor + wiring + list/lookup +
    `Close`) and TASK-104b (the call/read/get paths + composite execution). Per an explicit
    up-front decision, both land here as one change. Splitting them would leave `*coreVMCP`
    not satisfying the `VMCP` interface between PRs (the interface requires all ten methods),
    so the halves are not independently mergeable as working code. Actual net-new production
    logic is ~580 LOC across 4 files (`core_vmcp.go`, `core_calls.go`, `core.go`, `server.go`).

  - **A file move inflates the diff (refactor that must be atomic).**
    The backend-client telemetry decorator was relocated from `pkg/vmcp/server/telemetry.go`
    to a shared `pkg/vmcp/internal/backendtelemetry/` package so both `server` and `core` can
    use it without an import cycle. Git counts this as ~500 added + deleted lines, but it is a
    near-verbatim move (≈94% similarity) with net-zero behavior change — `server.New`'s
    signature and observable behavior are unchanged.

  - **Comprehensive unit tests (test-only, low review risk).**
    The remaining bulk is table-driven tests for the new core package (constructor validation,
    on-demand aggregation, health-filtering parity, lookups, copy-before-mutate, composite
    execution, advertise-vs-execute conflict parity, idempotent `Close`, identity-not-logged)
    plus the relocated telemetry tests.

  Splitting further would either ship non-compiling intermediate states or separate tests from
  the code they cover, both of which reduce reviewability rather than improve it.

### DIFF
--- a/pkg/vmcp/core/core_calls.go
+++ b/pkg/vmcp/core/core_calls.go
@@ -85,7 +85,9 @@ func (c *coreVMCP) ReadResource(
 		}
 		return nil, fmt.Errorf("routing resource %q: %w", uri, err)
 	}
-	return c.backendClient.ReadResource(ctx, target, target.GetBackendCapabilityName(uri))
+	// Pass the advertised URI; the backend client owns the single translation to
+	// the backend's capability name (client.go:874), matching CallTool.
+	return c.backendClient.ReadResource(ctx, target, uri)
 }
 
 // GetPrompt retrieves the named prompt from its backend. args is treated as
@@ -109,7 +111,9 @@ func (c *coreVMCP) GetPrompt(
 		}
 		return nil, fmt.Errorf("routing prompt %q: %w", name, err)
 	}
-	return c.backendClient.GetPrompt(ctx, target, target.GetBackendCapabilityName(name), maps.Clone(args))
+	// Pass the advertised name; the backend client owns the single translation to
+	// the backend's capability name (client.go:927), matching CallTool.
+	return c.backendClient.GetPrompt(ctx, target, name, maps.Clone(args))
 }
 
 // executeComposite runs a composite-tool workflow and converts the result to a
--- a/pkg/vmcp/core/core_calls.go
+++ b/pkg/vmcp/core/core_calls.go
@@ -14,7 +14,6 @@ import (
 	"github.com/stacklok/toolhive/pkg/auth"
 	"github.com/stacklok/toolhive/pkg/vmcp"
 	"github.com/stacklok/toolhive/pkg/vmcp/composer"
-	"github.com/stacklok/toolhive/pkg/vmcp/internal/compositetools"
 	"github.com/stacklok/toolhive/pkg/vmcp/router"
 )
 
@@ -42,13 +41,12 @@ func (c *coreVMCP) CallTool(
 		return nil, err
 	}
 
-	// Composite tool: execute the workflow when the tool is defined AND reachable
-	// in the current view (same accessibility filter ListTools advertises with).
-	if def, ok := c.workflowDefs[name]; ok {
-		accessible := compositetools.FilterWorkflowDefsForSession(c.workflowDefs, agg.RoutingTable)
-		if _, ok := accessible[name]; !ok {
-			return nil, fmt.Errorf("%w: composite tool %q", vmcp.ErrNotFound, name)
-		}
+	// Composite tool: execute only when the workflow is actually advertised in the
+	// current view — accessible AND not shadowed by a conflicting backend tool. This
+	// uses the same gate as ListTools (accessibleComposites), so advertised equals
+	// executed. A name that collides with a backend tool is NOT in the set and falls
+	// through to backend routing, matching the legacy decorator.
+	if def, ok := c.accessibleComposites(agg)[name]; ok {
 		engine := c.composerFactory(agg.RoutingTable, agg.Tools)
 		return executeComposite(ctx, engine, def, argsCopy)
 	}
--- a/pkg/vmcp/core/core_calls.go
+++ b/pkg/vmcp/core/core_calls.go
@@ -1,0 +1,159 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package core
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+	"maps"
+
+	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/vmcp"
+	"github.com/stacklok/toolhive/pkg/vmcp/composer"
+	"github.com/stacklok/toolhive/pkg/vmcp/internal/compositetools"
+	"github.com/stacklok/toolhive/pkg/vmcp/router"
+)
+
+// CallTool invokes the named tool. Composite tools (those defined as workflows)
+// execute through a per-call composer bound to the freshly aggregated routing
+// table; all other names route to a single backend via a session router built
+// from the same table. Returns vmcp.ErrNotFound for an unadvertised name.
+//
+// args and meta are treated as read-only and copied before being forwarded
+// (go-style: copy before mutating caller input). identity is accepted for the
+// admission seam (#5438); the core performs no identity-based filtering yet and
+// never logs identity. See ListTools for the nil/anonymous semantics.
+func (c *coreVMCP) CallTool(
+	ctx context.Context,
+	_ *auth.Identity,
+	name string,
+	args map[string]any,
+	meta map[string]any,
+) (*vmcp.ToolCallResult, error) {
+	argsCopy := maps.Clone(args)
+	metaCopy := maps.Clone(meta)
+
+	agg, err := c.aggregatedView(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Composite tool: execute the workflow when the tool is defined AND reachable
+	// in the current view (same accessibility filter ListTools advertises with).
+	if def, ok := c.workflowDefs[name]; ok {
+		accessible := compositetools.FilterWorkflowDefsForSession(c.workflowDefs, agg.RoutingTable)
+		if _, ok := accessible[name]; !ok {
+			return nil, fmt.Errorf("%w: composite tool %q", vmcp.ErrNotFound, name)
+		}
+		engine := c.composerFactory(agg.RoutingTable, agg.Tools)
+		return executeComposite(ctx, engine, def, argsCopy)
+	}
+
+	// Backend tool: route through a session router bound to the fresh table. The
+	// backend client translates the advertised name to the backend's capability
+	// name internally (client.go:772), mirroring the legacy tool handler.
+	target, err := router.NewSessionRouter(agg.RoutingTable).RouteTool(ctx, name)
+	if err != nil {
+		if errors.Is(err, router.ErrToolNotFound) {
+			return nil, fmt.Errorf("%w: tool %q", vmcp.ErrNotFound, name)
+		}
+		return nil, fmt.Errorf("routing tool %q: %w", name, err)
+	}
+	return c.backendClient.CallTool(ctx, target, name, argsCopy, metaCopy)
+}
+
+// ReadResource reads the resource at uri from its backend. Returns
+// vmcp.ErrNotFound for an unadvertised URI. See ListTools for identity semantics.
+func (c *coreVMCP) ReadResource(
+	ctx context.Context,
+	_ *auth.Identity,
+	uri string,
+) (*vmcp.ResourceReadResult, error) {
+	agg, err := c.aggregatedView(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	target, err := router.NewSessionRouter(agg.RoutingTable).RouteResource(ctx, uri)
+	if err != nil {
+		if errors.Is(err, router.ErrResourceNotFound) {
+			return nil, fmt.Errorf("%w: resource %q", vmcp.ErrNotFound, uri)
+		}
+		return nil, fmt.Errorf("routing resource %q: %w", uri, err)
+	}
+	return c.backendClient.ReadResource(ctx, target, target.GetBackendCapabilityName(uri))
+}
+
+// GetPrompt retrieves the named prompt from its backend. args is treated as
+// read-only and copied before being forwarded. Returns vmcp.ErrNotFound for an
+// unadvertised name. See ListTools for identity semantics.
+func (c *coreVMCP) GetPrompt(
+	ctx context.Context,
+	_ *auth.Identity,
+	name string,
+	args map[string]any,
+) (*vmcp.PromptGetResult, error) {
+	agg, err := c.aggregatedView(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	target, err := router.NewSessionRouter(agg.RoutingTable).RoutePrompt(ctx, name)
+	if err != nil {
+		if errors.Is(err, router.ErrPromptNotFound) {
+			return nil, fmt.Errorf("%w: prompt %q", vmcp.ErrNotFound, name)
+		}
+		return nil, fmt.Errorf("routing prompt %q: %w", name, err)
+	}
+	return c.backendClient.GetPrompt(ctx, target, target.GetBackendCapabilityName(name), maps.Clone(args))
+}
+
+// executeComposite runs a composite-tool workflow and converts the result to a
+// ToolCallResult. Workflow failures are returned as an IsError result (not a
+// transport error), mirroring the legacy compositeToolsDecorator
+// (internal/compositetools/decorator.go:76-114).
+func executeComposite(
+	ctx context.Context,
+	engine composer.Composer,
+	def *composer.WorkflowDefinition,
+	params map[string]any,
+) (*vmcp.ToolCallResult, error) {
+	result, err := engine.ExecuteWorkflow(ctx, def, params)
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			slog.Warn("workflow execution timeout", "tool", def.Name, "error", err)
+			return compositeErrorResult("Workflow execution timeout exceeded"), nil
+		}
+		slog.Error("workflow execution failed", "tool", def.Name, "error", err)
+		return compositeErrorResult(fmt.Sprintf("Workflow execution failed: %v", err)), nil
+	}
+	if result == nil {
+		slog.Error("workflow executor returned nil result", "tool", def.Name)
+		return compositeErrorResult("Workflow executor returned nil result"), nil
+	}
+	if result.Error != nil {
+		slog.Error("workflow completed with error", "tool", def.Name, "error", result.Error)
+		return compositeErrorResult(fmt.Sprintf("Workflow error: %v", result.Error)), nil
+	}
+
+	jsonBytes, err := json.Marshal(result.Output)
+	if err != nil {
+		return compositeErrorResult(fmt.Sprintf("failed to marshal output: %v", err)), nil
+	}
+	return &vmcp.ToolCallResult{
+		Content:           []vmcp.Content{{Type: vmcp.ContentTypeText, Text: string(jsonBytes)}},
+		StructuredContent: result.Output,
+	}, nil
+}
+
+// compositeErrorResult builds a tool-level error result for a failed workflow.
+func compositeErrorResult(msg string) *vmcp.ToolCallResult {
+	return &vmcp.ToolCallResult{
+		Content: []vmcp.Content{{Type: vmcp.ContentTypeText, Text: msg}},
+		IsError: true,
+	}
+}
--- a/pkg/vmcp/core/core_calls_test.go
+++ b/pkg/vmcp/core/core_calls_test.go
@@ -196,6 +196,34 @@ func TestGetPrompt(t *testing.T) {
 	assert.Equal(t, want, got)
 }
 
+func TestGetPrompt_CopyBeforeMutate(t *testing.T) {
+	t.Parallel()
+	cfg, m := baseConfig(t)
+
+	target := backendTarget()
+	expectAggregation(m, &aggregator.AggregatedCapabilities{
+		RoutingTable: &vmcp.RoutingTable{Prompts: map[string]*vmcp.BackendTarget{"p1": target}},
+	})
+
+	// The backend client mutates the args it receives; the caller's original must be
+	// untouched because GetPrompt forwards a clone (parity with CallTool).
+	m.client.EXPECT().
+		GetPrompt(gomock.Any(), gomock.Any(), "p1", gomock.Any()).
+		DoAndReturn(func(_ context.Context, _ *vmcp.BackendTarget, _ string, args map[string]any) (*vmcp.PromptGetResult, error) {
+			args["injected"] = true
+			return &vmcp.PromptGetResult{}, nil
+		})
+
+	c, err := New(cfg)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = c.Close() })
+
+	args := map[string]any{"x": 1}
+	_, err = c.GetPrompt(context.Background(), nil, "p1", args)
+	require.NoError(t, err)
+	assert.Equal(t, map[string]any{"x": 1}, args, "caller args must not be mutated")
+}
+
 func TestGetPrompt_NotFound(t *testing.T) {
 	t.Parallel()
 	cfg, m := baseConfig(t)
@@ -344,6 +372,13 @@ func TestExecuteComposite(t *testing.T) {
 			composer:    stubComposer{result: &composer.WorkflowResult{Error: errors.New("step failed")}},
 			wantIsError: true,
 			wantMsg:     "Workflow error",
+		},
+		{
+			name: "marshal failure",
+			// A channel is not JSON-serializable, forcing json.Marshal(Output) to fail.
+			composer:    stubComposer{result: &composer.WorkflowResult{Output: map[string]any{"ch": make(chan int)}}},
+			wantIsError: true,
+			wantMsg:     "failed to marshal output",
 		},
 	}
 
--- a/pkg/vmcp/core/core_calls_test.go
+++ b/pkg/vmcp/core/core_calls_test.go
@@ -1,0 +1,209 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package core
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	"github.com/stacklok/toolhive/pkg/vmcp"
+	"github.com/stacklok/toolhive/pkg/vmcp/aggregator"
+	"github.com/stacklok/toolhive/pkg/vmcp/composer"
+)
+
+// expectAggregation wires reg.List + agg.AggregateCapabilities to return agg once.
+func expectAggregation(m *coreMocks, agg *aggregator.AggregatedCapabilities) {
+	m.reg.EXPECT().List(gomock.Any()).Return([]vmcp.Backend{{ID: "be1", HealthStatus: vmcp.BackendHealthy}})
+	m.agg.EXPECT().AggregateCapabilities(gomock.Any(), gomock.Any()).Return(agg, nil)
+}
+
+func TestCallTool_RoutesToBackend(t *testing.T) {
+	t.Parallel()
+	cfg, m := baseConfig(t)
+
+	target := backendTarget()
+	expectAggregation(m, &aggregator.AggregatedCapabilities{
+		Tools:        []vmcp.Tool{backendTool("tool_a")},
+		RoutingTable: &vmcp.RoutingTable{Tools: map[string]*vmcp.BackendTarget{"tool_a": target}},
+	})
+
+	want := &vmcp.ToolCallResult{StructuredContent: map[string]any{"result": "ok"}}
+	m.client.EXPECT().
+		CallTool(gomock.Any(), gomock.Any(), "tool_a", gomock.Any(), gomock.Any()).
+		Return(want, nil)
+
+	c, err := New(cfg)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = c.Close() })
+
+	got, err := c.CallTool(context.Background(), nil, "tool_a", map[string]any{"a": 1}, nil)
+	require.NoError(t, err)
+	assert.Equal(t, want, got)
+}
+
+func TestCallTool_NotFound(t *testing.T) {
+	t.Parallel()
+	cfg, m := baseConfig(t)
+
+	expectAggregation(m, &aggregator.AggregatedCapabilities{RoutingTable: &vmcp.RoutingTable{}})
+
+	c, err := New(cfg)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = c.Close() })
+
+	_, err = c.CallTool(context.Background(), nil, "missing", nil, nil)
+	assert.ErrorIs(t, err, vmcp.ErrNotFound)
+}
+
+func TestCallTool_CopyBeforeMutate(t *testing.T) {
+	t.Parallel()
+	cfg, m := baseConfig(t)
+
+	target := backendTarget()
+	expectAggregation(m, &aggregator.AggregatedCapabilities{
+		RoutingTable: &vmcp.RoutingTable{Tools: map[string]*vmcp.BackendTarget{"tool_a": target}},
+	})
+
+	// The backend client mutates the maps it receives; the caller's originals
+	// must be untouched because CallTool forwards clones.
+	m.client.EXPECT().
+		CallTool(gomock.Any(), gomock.Any(), "tool_a", gomock.Any(), gomock.Any()).
+		DoAndReturn(func(_ context.Context, _ *vmcp.BackendTarget, _ string, args, meta map[string]any) (*vmcp.ToolCallResult, error) {
+			args["injected"] = true
+			meta["injected"] = true
+			return &vmcp.ToolCallResult{}, nil
+		})
+
+	c, err := New(cfg)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = c.Close() })
+
+	args := map[string]any{"a": 1}
+	meta := map[string]any{"m": "n"}
+	_, err = c.CallTool(context.Background(), nil, "tool_a", args, meta)
+	require.NoError(t, err)
+
+	assert.Equal(t, map[string]any{"a": 1}, args, "caller args must not be mutated")
+	assert.Equal(t, map[string]any{"m": "n"}, meta, "caller meta must not be mutated")
+}
+
+func TestCallTool_CompositeWorkflow(t *testing.T) {
+	t.Parallel()
+	cfg, m := baseConfig(t)
+	cfg.WorkflowDefs = map[string]*composer.WorkflowDefinition{
+		"wf": {Name: "wf", Steps: []composer.WorkflowStep{{ID: "s1", Type: composer.StepTypeTool, Tool: "be1.echo"}}},
+	}
+
+	target := backendTarget()
+	expectAggregation(m, &aggregator.AggregatedCapabilities{
+		Tools:        []vmcp.Tool{backendTool("be1.echo")},
+		RoutingTable: &vmcp.RoutingTable{Tools: map[string]*vmcp.BackendTarget{"be1.echo": target}},
+	})
+
+	// The composite workflow's single tool step routes to the backend through the
+	// per-call composer built from the aggregated routing table.
+	m.client.EXPECT().
+		CallTool(gomock.Any(), gomock.Any(), "be1.echo", gomock.Any(), gomock.Any()).
+		Return(&vmcp.ToolCallResult{StructuredContent: map[string]any{"ok": true}}, nil)
+
+	c, err := New(cfg)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = c.Close() })
+
+	got, err := c.CallTool(context.Background(), nil, "wf", nil, nil)
+	require.NoError(t, err)
+	require.NotNil(t, got)
+	assert.False(t, got.IsError)
+	assert.Equal(t, true, got.StructuredContent["ok"])
+}
+
+func TestCallTool_CompositeNotAccessible(t *testing.T) {
+	t.Parallel()
+	cfg, m := baseConfig(t)
+	cfg.WorkflowDefs = map[string]*composer.WorkflowDefinition{
+		"wf": {Name: "wf", Steps: []composer.WorkflowStep{{ID: "s1", Type: composer.StepTypeTool, Tool: "be1.echo"}}},
+	}
+
+	// Routing table does not contain be1.echo, so the composite is not reachable.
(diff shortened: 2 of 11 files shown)


pkg/vmcp/core/core_calls.go
```diff
@@ -0,0 +1,163 @@
...
+	metaCopy := maps.Clone(meta)
+
+	agg, err := c.aggregatedView(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Composite tool: execute the workflow when the tool is defined AND reachable
+	// in the current view (same accessibility filter ListTools advertises with).
+	if def, ok := c.workflowDefs[name]; ok {
```
tgrunnagle: **F1 · HIGH (latent) — advertise-vs-execute divergence on tool-name conflict.**

`advertisedTools` (`core_vmcp.go:818`) runs `ValidateNoToolConflicts` and drops **all** composites on a name collision (the backend tool is advertised). But this branch checks `c.workflowDefs[name]` first and only applies the *accessibility* filter — not the conflict check — so on a conflict it **executes the composite**, shadowing the advertised backend tool.

Legacy is consistent: `factory.go:165-168` returns the undecorated session on conflict and `decorator.go:83-86` routes the name to the **backend** (advertised == executed). This is a behavior change under a "no behavior change" PR and a latent ambiguity bug that goes live when the core replaces the legacy path (Phase 3).

**Fix**: extract a shared helper used by both `advertisedTools` and `CallTool` (e.g. `accessibleComposites(agg) (defs, conflict bool)`); on conflict, fall through to backend routing. Add the parity test in F6 / a dedicated conflict test.

pkg/vmcp/core/core_vmcp.go
```diff
@@ -0,0 +1,415 @@
...
+		// Include healthy, degraded, and empty/zero-value (assume healthy) backends.
+		// Explicitly exclude unhealthy, unknown, and unauthenticated backends.
+		if healthStatus == "" ||
+			healthStatus == vmcp.BackendHealthy ||
+			healthStatus == vmcp.BackendDegraded {
+			healthy = append(healthy, *backend)
+		} else {
+			excluded++
+			//nolint:gosec // G706: backend fields are internal, not user-controlled
+			slog.Debug("excluding backend from capability aggregation due to health status",
```
tgrunnagle, line 449: **F2 · LOW — C2 copy drops the `"source"` log field.**

The legacy `discovery/middleware.go:196-201` emits `"source": "health_monitor"|"registry"` on this exclusion log; the relocated copy omits it. The include/exclude *rules* are identical (behavior preserved), but the function comment reads as a full-fidelity claim. Either add the field back to match byte-for-byte, or narrow the comment to "only the include/exclude rules are kept identical."

pkg/vmcp/core/core_vmcp.go
```diff
@@ -0,0 +1,415 @@
...
+	elicitationHandler := composer.NewDefaultElicitationHandler(cfg.Elicitation)
+
+	// State store + workflow engine (server.go:386-387). The state store starts a
+	// background cleanup goroutine immediately, so any error path after this point
+	// must stop it to avoid a leak.
+	stateStore := composer.NewInMemoryStateStore(stateStoreCleanupInterval, stateStoreMaxAge)
+
+	// composerFactory builds a composite-tool engine bound to a specific routing
+	// table, generalizing server.New's sessionComposerFactory (server.go:393-398).
+	composerFactory := func(sessionRT *vmcp.RoutingTable, sessionTools []vmcp.Tool) composer.Composer {
```
tgrunnagle, line 167: **F3 · LOW — workflow-level telemetry not wrapped here (confirm deferral).**

Legacy wraps composite executors with `telemetryWorkflowExecutor` (`sessionmanager/factory.go:174-176`) for `toolhive_vmcp_workflow_executions/_errors/_duration` + a span. This `composerFactory` doesn't — but it faithfully mirrors `server.go:393-398`, and the instrument wrapping lives in the **session/Serve layer**, so this is likely correct deferral to #5439. Please confirm and add a one-line comment so the omission doesn't read as a regression.

pkg/vmcp/core/core_vmcp.go
```diff
@@ -0,0 +1,415 @@
...
+	}
+	slog.Info("loaded valid composite tool workflows", "count", len(validDefs))
+	return validDefs, nil
+}
+
+// stopStateStore stops the state store's background cleanup goroutine when the
+// concrete implementation supports it. The WorkflowStateStore interface does not
+// expose Stop, so the in-memory implementation is detected via a capability
+// assertion; other stores (no goroutine to stop) are a no-op.
+func stopStateStore(store composer.WorkflowStateStore) {
```
tgrunnagle: **F4 · LOW — silent capability assertion (latent leak).**

`store.(interface{ Stop() })` no-ops on failure — the silent-no-op pattern `go-style.md` warns against. Latent today (the store is always `*inMemoryStateStore`), but `composer` also defines `InMemoryWorkflowStateStore`, which has no `Stop()`; a future constructor swap would leak the cleanup goroutine with no compile error. Prefer a typed reference or a narrow `Stopper` capability returned by `NewInMemoryStateStore`.

pkg/vmcp/core/core_vmcp_test.go
```diff
@@ -0,0 +1,417 @@
...
+		require.NoError(t, c.Close())
+		require.NoError(t, c.Close())
+	})
+}
+
+// Must run serially: it swaps the global slog default to capture output into a
+// non-thread-safe buffer, which is unsafe if other tests log concurrently.
+//
+//nolint:paralleltest // installs a global slog default + non-thread-safe buffer; must not run in parallel
+func TestIdentityNotLogged(t *testing.T) {
```
tgrunnagle, line 444: **F5 · LOW — identity-not-logged coverage is narrow.**

Only `ListTools` (success) + `CallTool` (ErrNotFound) are exercised; `ReadResource`/`GetPrompt`/`Lookup*`/`executeComposite` error branches/`New` INFO lines aren't. `auth.Identity` implements only `String()`/`MarshalJSON()` (not `slog.LogValuer`), so a future `slog.X("...", "identity", identity)` would leak via reflection and only these two paths would catch it. The `AnyTimes()` mocks also let the test pass if a path short-circuits before logging. Drive all methods, assert the buffer is non-empty, and consider adding `slog.LogValuer` to `auth.Identity`.

pkg/vmcp/core/core_vmcp_test.go
```diff
@@ -0,0 +1,417 @@
...
+				assert.Nil(t, c)
+				return
+			}
+			require.NoError(t, err)
+			require.NoError(t, c.Close())
+		})
+	}
+}
+
+func TestListTools_AggregatesOnDemand(t *testing.T) {
```
tgrunnagle, line 209: **F6 · MEDIUM — missing issue-mandated parity test.**

The issue's Testing Strategy requires asserting the core's `ListTools` capability set matches the session-factory `ProcessPreQueriedCapabilities` path (same aggregator/filter/registry, no drift / no double-aggregation). All `ListTools` tests here mock `AggregateCapabilities` and assert the mocked return, so they can't detect pipeline drift. Add a test driving a real/shared aggregator through both paths and `ElementsMatch` the advertised sets (name + `BackendID`), or document the deferral with a follow-up reference.

pkg/vmcp/core/core_calls_test.go
```diff
@@ -0,0 +1,320 @@
...
+}
+func (stubComposer) ValidateWorkflow(_ context.Context, _ *composer.WorkflowDefinition) error {
+	return nil
+}
+func (stubComposer) GetWorkflowStatus(_ context.Context, _ string) (*composer.WorkflowStatus, error) {
+	return nil, nil
+}
+func (stubComposer) CancelWorkflow(_ context.Context, _ string) error { return nil }
+
+func TestExecuteComposite(t *testing.T) {
```
tgrunnagle, line 334: **F7 · LOW — 6th `executeComposite` branch untested.**

Covers 5 of 6 branches; the `json.Marshal(result.Output)` failure path (`core_calls.go:147`) is untested. Add a row with an unmarshalable `Output` (e.g. `map[string]any{"c": make(chan int)}`) asserting `IsError` + content contains `"failed to marshal output"`.

pkg/vmcp/core/core_calls_test.go
```diff
@@ -0,0 +1,320 @@
...
+
+	c, err := New(cfg)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = c.Close() })
+
+	_, err = c.ReadResource(context.Background(), nil, "file://missing")
+	assert.ErrorIs(t, err, vmcp.ErrNotFound)
+}
+
+func TestGetPrompt(t *testing.T) {
```
tgrunnagle, line 178: **F8 · LOW — `GetPrompt` copy-before-mutate unverified.**

`GetPrompt` clones `args` (`core_calls.go:116`) but this test doesn't assert the caller's map is untouched, unlike `TestCallTool_CopyBeforeMutate`. Add a `DoAndReturn` that mutates the received `args` and assert the caller's original is unchanged (or fold both into one copy-before-mutate table).

pkg/vmcp/core/core_vmcp_test.go
```diff
@@ -0,0 +1,417 @@
...
+
+func TestNew_NilConfig(t *testing.T) {
+	t.Parallel()
+	c, err := New(nil)
+	require.Error(t, err)
+	assert.ErrorIs(t, err, vmcp.ErrInvalidConfig)
+	assert.Nil(t, c)
+}
+
+func TestNew_ValidatesWorkflows(t *testing.T) {
```
tgrunnagle, line 116: **F9 · LOW — no goroutine-leak guard.**

This test drives the `New` validation-error path (which calls `stopStateStore`) but can't confirm the cleanup goroutine exited; no goleak guard exists for the package. Add `goleak.VerifyTestMain(m)` in a `TestMain` for `pkg/vmcp/core` — cheap durable coverage for the "releases core-held resources" criterion and the error-path cleanup.

tgrunnagle (commented):
## 🔍 Multi-agent review — Phase 1 core VMCP constructor

**Verdict: COMMENT.** Faithful, well-executed "relocate, don't rewrite" refactor; safe to merge as additive code (core not yet wired into the request path). One **HIGH-severity *latent* parity defect** and two MEDIUM/LOW test-coverage gaps warrant attention before the core is wired in (Phase 3). Reviewed by 5 specialist agents (general quality, architecture/parity, concurrency, security, tests). Codex cross-review skipped (CLI not installed).

### Consensus findings
| # | Finding | Location | Severity | Score |
|---|---------|----------|----------|-------|
| F1 | `CallTool` composite branch lacks the conflict gate `ListTools` applies → executes a withheld composite on name conflict (legacy routes to backend) | `core_calls.go:47` | **HIGH** *(latent)* | 9 |
| F2 | C2 `filterHealthyBackends` copy drops the legacy `"source"` debug-log field | `core_vmcp.go:399` | LOW | 8 |
| F3 | Workflow-level telemetry instruments not carried into `composerFactory` — confirm Serve-layer deferral | `core_vmcp.go:142` | LOW | 7 |
| F4 | `stopStateStore` silent `interface{ Stop() }` assertion; a 2nd store impl lacks `Stop()` (latent leak) | `core_vmcp.go:337` | LOW | 7 |
| F5 | `TestIdentityNotLogged` covers 2 of ~8 paths; `auth.Identity` has no `slog.LogValuer` | `core_vmcp_test.go:393` | LOW | 7 |
| F6 | Missing issue-mandated capability-set parity test vs `ProcessPreQueriedCapabilities` | `core_vmcp_test.go:158` | MEDIUM | 7 |
| F7 | `executeComposite` `json.Marshal`-failure branch (6th) untested | `core_calls_test.go:260` | LOW | 6 |
| F8 | `GetPrompt` copy-before-mutate unverified (asymmetric with `CallTool`) | `core_calls_test.go:178` | LOW | 6 |
| F9 | No goroutine-leak guard for the constructor error-path | `core_vmcp_test.go:116` | LOW | 6 |

### Holistic notes
- **Strengths**: state-store constants, telemetry-before-engine ordering, `executeComposite` error semantics, and `filterHealthyBackends` include/exclude *rules* all match legacy; comment line-citations verified accurate; clean dependency direction (no cycle); no mcp-go type crosses the boundary; identity never logged. `Close()` actually **fixes a pre-existing goroutine leak** in the relocated `server.New` code.
- **F1 is the one to weigh**: it contradicts the "no behavior change" label and the issue's parity criteria. Not reachable in production from this PR (additive), so non-blocking — but fix it or add an explicit tracking note before Phase 3 wires the core in.
- **Docs**: no `task docs` needed. Minor: the PR "Changes" table lists `elicitation.go`/`composer/`/`sdk_elicitation_adapter*.go`, which are not in this diff (they're in the stacked base `#5436`).

*Inline comments below map 1:1 to the table. `REQUEST_CHANGES` was intentionally not auto-set — set it yourself if you'd prefer to block on F1.*

pkg/vmcp/core/core_calls.go
tgrunnagle: Addressed in f8bd1890. Extracted a shared `accessibleComposites(agg)` helper used by both `advertisedTools` and `CallTool`; on a name conflict all composites are dropped and `CallTool` falls through to backend routing, so advertised == executed (matching `factory.go:165-168`/`decorator.go:83-86`). Added `TestCompositeNameConflict_AdvertisedEqualsExecuted` as the parity guard.

pkg/vmcp/core/core_vmcp.go
tgrunnagle: Addressed in b581bb5c. Restored the `"source": "health_monitor"|"registry"` field on the exclusion debug log so the relocated filter now matches the legacy log byte-for-byte.

pkg/vmcp/core/core_vmcp.go
tgrunnagle: Confirmed deferral — addressed in b581bb5c. Added a comment that workflow-level telemetry (toolhive_vmcp_workflow_*) is intentionally NOT wired here; that instrumentation lives in the session/Serve layer (sessionmanager/factory.go:174-176) and lands when the core is wired into Serve (#5439). This mirrors the uninstrumented server.go:393-398.

pkg/vmcp/core/core_vmcp.go
tgrunnagle: Addressed in b581bb5c. The stopper is now captured at construction (the store is created internally, so the type is known there) and stored as a `stopStore func()` field; if a future store swap exposes no `Stop()`, New logs a WARN instead of silently no-op'ing, so a leaked goroutine is diagnosable. Close() just invokes it under sync.Once.

pkg/vmcp/core/core_vmcp_test.go
tgrunnagle: Addressed in 3d6bcd2d. TestIdentityNotLogged now drives all nine identity-taking methods (List*/Lookup*/CallTool/ReadResource/GetPrompt) and asserts the captured log buffer is non-empty (an excluded backend forces a DEBUG line every aggregatedView), so the no-secret assertion can't pass vacuously. Adding slog.LogValuer to auth.Identity is a good defensive follow-up but touches pkg/auth (outside this PR's scope).

pkg/vmcp/core/core_calls_test.go
tgrunnagle: Addressed in 3d6bcd2d. Added a 'marshal failure' row to TestExecuteComposite with an unserializable Output (map containing a channel), asserting IsError and content containing 'failed to marshal output' — all six branches are now covered.

pkg/vmcp/core/core_calls_test.go
tgrunnagle: Addressed in 3d6bcd2d. Added TestGetPrompt_CopyBeforeMutate: the mock client mutates the received args and the test asserts the caller's original map is unchanged, matching TestCallTool_CopyBeforeMutate.

pkg/vmcp/core/core_vmcp_test.go
tgrunnagle: Deferred (no commit). The core only ever calls `AggregateCapabilities`; the parity between that and `ProcessPreQueriedCapabilities` is an invariant of the *aggregator* package (the core can't drift from a method it doesn't call), so a true no-drift assertion belongs in aggregator tests rather than mocked core tests. Since this is additive code not yet on the request path, tracking it as a Phase-3 follow-up (when the core is wired into Serve and both paths run live) — happy to add an aggregator-level parity test in that work if you'd prefer it tracked there.

pkg/vmcp/core/core_vmcp_test.go
tgrunnagle: Deferred (no commit). goleak has no precedent anywhere in this repo and is a transitive-only dependency (adding a TestMain would require a new direct require in go.mod), and VerifyTestMain tends to flag unrelated background goroutines from imported subsystems (otel/telemetry/composer), risking flakiness. The 'releases core-held resources' criterion and the error-path cleanup are already exercised behaviorally (TestClose_Idempotent + the New validation-error path now routing through the captured stopStore). Happy to introduce a package-wide goleak guard as a separate, repo-wide initiative if the team wants that pattern.

pkg/vmcp/core/core_vmcp.go
```diff
@@ -0,0 +1,444 @@
...
+		workflowAuditor, err = audit.NewWorkflowAuditor(cfg.AuditConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create workflow auditor: %w", err)
+		}
+		slog.Info("workflow audit logging enabled")
+	}
+
+	// The elicitation handler depends only on the domain-typed ElicitationRequester
+	// (#5436); no mcp-go types cross this boundary (vmcp anti-pattern #5).
+	elicitationHandler := composer.NewDefaultElicitationHandler(cfg.Elicitation)
```
jerm-dro, line 142: **blocker:** `cfg.Elicitation` is allowed to be nil here, but the resulting `DefaultElicitationHandler` will nil-dereference the moment a workflow hits an elicitation step. The legacy `server.New` path can't exhibit this because it always wires a non-nil `sdkElicitationAdapter` — so this is a new failure mode introduced by the core.

Repro:

1. Construct the core with `Config{Elicitation: nil, WorkflowDefs: {"wf": <workflow with an elicitation step>}}`. Validation passes (it only checks structure).
2. `CallTool(ctx, _, "wf", ...)` builds an engine, finds the composite, calls `ExecuteWorkflow`.
3. The engine reaches the elicitation step. `workflow_engine.go:628` checks `e.elicitationHandler == nil` — false, the handler exists.
4. Falls through to `h.requester.RequestElicitation(reqCtx, req)` at `composer/elicitation_handler.go:168`. `h.requester` is the nil `vmcp.ElicitationRequester` you passed → panic.

Preferred fix (in `composer/elicitation_handler.go`, not here) — protects every caller, not just core:

```go
var ErrElicitationNotConfigured = errors.New("elicitation requester not configured")

func (h *DefaultElicitationHandler) RequestElicitation(
    ctx context.Context, workflowID, stepID string, config *ElicitationConfig,
) (*ElicitationResponse, error) {
    if h.requester == nil {
        return nil, fmt.Errorf("%w: step %s", ErrElicitationNotConfigured, stepID)
    }
    // ...existing body...
}
```

Alternative (here in `core.New`) — fail fast at construction so the contract is "if any workflow needs elicitation, `Elicitation` must be non-nil":

```go
if cfg.Elicitation == nil && workflowsNeedElicitation(cfg.WorkflowDefs) {
    return nil, fmt.Errorf("%w: Elicitation is required when workflows contain elicitation steps", vmcp.ErrInvalidConfig)
}
```

jerm-dro (changes requested):

pkg/vmcp/core/core_vmcp.go
tgrunnagle: Addressed in a7e1b7f2 — went with both fixes you outlined, since they're complementary:

- **Composer guard (your preferred):** `DefaultElicitationHandler.RequestElicitation` now returns a new `ErrElicitationNotConfigured` when `h.requester == nil` instead of dereferencing it, so every caller fails cleanly rather than panicking.
- **core.New fail-fast:** `New` now rejects construction with `vmcp.ErrInvalidConfig` when `Elicitation` is nil and any workflow contains an elicitation step (directly or as a forEach inner step), so the misconfiguration surfaces at startup per the documented contract (go-style fail-loudly).

Tests added: `TestDefaultElicitationHandler_NilRequester` (composer) and `TestNew_ElicitationRequiredWhenWorkflowElicits` (core, table-driven over elicitation / forEach-inner-elicitation / tool-only). Updated the `Config.Elicitation` doc to state the enforced contract.

jerm-dro (approved):

Issue #5441, opened 2026-06-03
P2.3 Move middleware chain under Serve; remove authz + annotation mw


tgrunnagle:
## Description

Relocate the full HTTP middleware chain so the new `Serve`-built `*Server` uses it —
recovery, header-validation, write-timeout (SSE), auth, audit (+ backend enrichment),
telemetry, MCP-parsing — preserving the existing wrapping/execution order. On the
`Serve` path, **authz and annotation-enrichment are omitted via the existing
`cfg.AuthzMiddleware != nil` nil-guard**: `Serve` leaves `AuthzMiddleware` nil, so the
guards at `server.go:606`/`:614` skip both blocks, and authorization is enforced instead
by the core admission seam from #5438. The authz/annotation blocks are **NOT deleted
in Phase 2** — they remain in the shared `(*Server).Handler` so the still-live
`server.New` path keeps enforcing authz with no regression; their physical removal is
deferred to Phase 3 (#5445), after `server.New` is routed through `Serve`.

## Context

Part of the vMCP interface refactor (epic #5419), which splits the
`server.New` god-object into a core (`New(cfg) -> VMCP`) and a transport
(`Serve(ctx, VMCP, serverCfg) -> *Server`). Today the entire MCP-endpoint
middleware chain is assembled in `(*Server).Handler` (`server.go:574-675`),
wrapped outermost-last so execution runs recovery → header-val → write-timeout →
auth+parser → audit → discovery → annotation-enrichment → authz →
backend-enrichment → MCP-parsing → telemetry → handler. Two of those layers —
`AuthzMiddleware` (applied only when `cfg.AuthzMiddleware != nil`,
`server.go:606`) and `AnnotationEnrichmentMiddleware` (`server.go:615`, gated on
the same condition, block `611-617`) — exist solely to make the Cedar authorization
decision on the HTTP path, with annotation-enrichment injecting `Tool.Annotations`
into context so Cedar `when`-clauses evaluate.

**Sequencing constraint (why nothing is deleted here).** `(*Server).Handler` is a
**single shared method** read from `s.config`; both `server.New`'s `*Server` and a
`Serve`-built `*Server` use it. In Phase 2, `server.New` is **not yet** routed through
`Serve` (that is Phase 3 / #5445), and the core admission seam is not in its call
path. If this task deleted the authz/annotation blocks from the shared `Handler`,
`server.New` would lose authorization enforcement entirely until Phase 3 — a real
[...]

PR #5459, opened 2026-06-05
Wire Cedar admission seam into vMCP core

### LINKED ISSUE #5438: P1.5 Core admission seam (bounded rewrite)

tgrunnagle:
## Summary

Today vMCP authorization runs **only as HTTP middleware** (`AuthzMiddleware` +
`AnnotationEnrichmentMiddleware`), applied on a single request path. Because the
list path and the call path are enforced by separate mechanisms, the policy
model permits a "list says yes / call says no" gap and depends on
annotation-enrichment middleware running first to inject `Tool.Annotations` for
Cedar `when`-clauses. RFC THV-0076 requires `ListTools` and `CallTool` to enforce
the **same** decision from one source.

This PR adds the `Admission` seam — bounded rewrite **R1**, the
highest-regression-risk change in the core refactor — and wires it into the
`VMCP` core implemented in #5437. It re-platforms the existing Cedar decision into
the core without changing the policy model: it wraps the existing
`authorizers.Authorizer` (no new policy language), so `List*` filter and
`Call`/`Read`/`Get` deny on one shared decision, and `Lookup*` never resolve a
denied capability.

- **Why:** close the list/call enforcement gap and give the core a single
  authorization source, unblocking #5441 (which removes the HTTP authz +
  annotation-enrichment middleware in Phase 2).
- **What:** new `Admission` interface + Cedar-backed adapter in
  `pkg/vmcp/core/admission.go`; `New` builds the seam from `cfg.Authz`
  (allow-all when unconfigured) and the core methods consult it.

Closes #5438

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

The shipped `server.New` path is unchanged: the HTTP authz middleware still runs
in Phase 1, so observable behavior is identical. The new core seam is exercised
directly by unit tests and goes live when #5441 retires the middleware.

## Test plan

Verification performed:

- `go test -race ./pkg/vmcp/core/...` passes (new `admission_test.go` covers all
  R1 acceptance criteria — see Changes below).
- `golangci-lint run ./pkg/vmcp/core/...` reports 0 issues.
- `task build` succeeds.
- Full `task test` passes. The only repo-wide lint failure (gosec G115 in
  `cmd/thv/app/upgrade.go`) is pre-existing and unrelated to this change.

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/core/admission.go` (new) | `Admission` interface (`FilterTools`/`AllowToolCall`/`FilterResources`/`AllowResourceRead`/`FilterPrompts`/`AllowPromptGet`, domain-typed); `cedarAdmission` adapter wrapping `authorizers.Authorizer`; `allowAllAdmission` no-op; `newAdmission` factory; annotation converter and advertised-tool lookup helpers. |
| `pkg/vmcp/core/core.go` | `Config` gains `ServerName` and `PassThroughTools`; clarified `Authz` doc to describe the allow-all-when-nil parity. |
| `pkg/vmcp/core/core_vmcp.go` | `coreVMCP` gains an `admission` field; `New` builds the seam (before acquiring resources, so a bad policy fails fast without leaking the state-store goroutine); `List*` now filter through the seam and `Lookup*` inherit that view. |
| `pkg/vmcp/core/core_calls.go` | `CallTool`/`ReadResource`/`GetPrompt` consult the seam to deny, returning `vmcp.ErrAuthorizationFailed`; `CallTool` sources the tool's annotations from the advertised set so annotation-gated policies evaluate. |
| `pkg/vmcp/core/admission_test.go` (new) | R1 parity tests: list/call/lookup agree on the same decision; annotation-gated `when`-clauses (`readOnlyHint`); allow-all when unconfigured; `passThroughTools` exempt; identity never logged; correct feature/operation per method; per-tool error skips (log-and-continue); resource/prompt parity; `newAdmission` config validation. |

## Implementation details

- **No new policy model.** The adapter wraps the existing
  `authorizers.Authorizer`, built via the same registry path the HTTP middleware
  uses (`authorizers.GetFactory` + `CreateAuthorizer`). Cedar's `when`-clause and
  feature/operation semantics are unchanged.
- **Identity is explicit, never read from ctx on the public path** (vMCP
  anti-pattern #1). The adapter re-injects identity into the ctx the wrapped
  authorizer reads — an internal adapter detail only.
- **Annotations sourced from the core.** `CallTool` looks up the tool in the
  recomputed advertised set and the adapter injects its annotations via
  `authorizers.WithToolAnnotations`, replacing the annotation-enrichment
  middleware on the domain path. The converter only writes annotation ctx when a
  hint is set (matching the existing `hasAnyHint` gate).
- **Allow-all parity.** A nil `cfg.Authz` yields an `allowAllAdmission`: `Filter*`
  return input unchanged and `Allow*` return `true`, preserving today's
  `AuthzMiddleware != nil` guard.

No. `server.New`'s signature and observable behavior are unchanged; the seam is
not yet on the shipped request path.

1. **Deferred to #5441 (security-relevant).** This seam intentionally exempts the
   optimizer meta-tools (`find_tool`/`call_tool`) from its decision, per AC R1(4).
   It does **not** replicate the HTTP middleware's `call_tool` inner-tool
   authorization (`handleToolsCall` re-authorizes the wrapped inner tool). This is
   harmless today because the optimizer decorator calls the backend handler
   directly and never routes an unwrapped inner name through `core.CallTool`.
   **#5441 (which removes the HTTP authz middleware) MUST carry an explicit
   acceptance criterion** that the inner `call_tool` target is authorized through
   `Admission.AllowToolCall` with the real tool name, with a test for a
   `call_tool`-wrapped denied tool. This is documented in `admission.go`'s
   `AllowToolCall` comment.

2. **Follow-up for #5441.** Two authorizer-construction paths (core `newAdmission`
   vs the live `newCedarAuthzMiddleware`) now both carry the empty-`ServerName`
   check; consolidate them into one shared constructor when #5441 collapses the
   middleware.

3. **PR sizing.** Production (non-test) change is ~436 net lines (489 added / 53
   removed) across 4 files (within the ≤10-file cap) — modestly over the 400-line
   guideline. The overage is reviewer-requested documentation comments and a small
   `validateConfig` extraction added to keep `New` under the gocyclo limit after
   rebasing onto the merged #5437. Tests (~790 lines, the majority of the diff) are
   excluded from the cap. Kept as one PR because the interface + adapter + wiring are
   one cohesive change; see the Large PR Justification section below.

4. **File location.** The issue text referenced `pkg/vmcp/admission.go`, but the
   core moved to its own `pkg/vmcp/core` package in #5437; the seam lives at
   `pkg/vmcp/core/admission.go` because the Cedar adapter imports `pkg/authz`,
   which would create an import cycle with the root `pkg/vmcp` package.

No separate AI-generated plan document exists for this change — the issue's own
"Technical Approach" served as the plan, so the optional implementation-plan
section is omitted.

## Large PR Justification

The total diff (~1280 insertions) exceeds the 1000-line threshold, but this is a
single, cohesive, bounded change — the R1 core admission seam (#5438): the `Admission`
interface, its Cedar-backed adapter, and the core wire-in (`List*` filter,
`Call`/`Read`/`Get` deny, `Lookup*`). The issue explicitly scoped it as one PR and
flagged the size as "borderline".

- **The majority (~790 lines) is the new test file** `pkg/vmcp/core/admission_test.go`,
  covering the R1.1–R1.5 security-parity matrix (list/call enforce the same decision,
  annotation-gated policies, allow-all parity, pass-through exemption, identity never
  logged) plus construction, error-classification, and args-flow cases. Tests are
  excluded from the project's 400-line production cap.
- **Production code is ~436 net lines across 4 files** (within the ≤10-file cap),
  modestly over the 400-line guideline; the overage is reviewer-requested doc comments
  and a small `validateConfig` extraction (to keep `New` under the gocyclo limit after
  rebasing onto the merged #5437).
- **Splitting is not warranted:** the interface, adapter, and wire-in are one bounded
  contract that reviews more clearly together than in fragments, and the PR has already
  completed two human/multi-agent review rounds.

### DIFF
--- a/pkg/vmcp/core/admission.go
+++ b/pkg/vmcp/core/admission.go
@@ -53,11 +53,27 @@ type Admission interface {
 // HTTP middleware uses (authorizers.GetFactory + CreateAuthorizer, the
 // construction inside newCedarAuthzMiddleware). The Cedar factory is registered by
 // pkg/authz's blank import, which the core package already pulls in.
+//
+// authzCfg is an already-built, authorizer-agnostic config: the cedar-specific
+// normalization the live path performs at construction time (e.g. defaulting an
+// empty EntitiesJSON to "[]", incoming.go:128) belongs to whoever builds
+// authzCfg, not to this generic factory path, which consumes RawConfig as-is.
 func newAdmission(
 	authzCfg *authorizers.Config, serverName string, passThroughTools map[string]struct{},
 ) (Admission, error) {
 	if authzCfg == nil {
 		return allowAllAdmission{}, nil
+	}
+
+	// Fail loudly on an empty server name, mirroring the live factory
+	// (incoming.go:120). Cedar attaches the MCP::"<serverName>" parent UID to
+	// resource entities only when serverName is non-empty
+	// (cedar/entity.go:170-174), so an empty name silently changes resource-scoped
+	// policy semantics rather than erroring (go-style: fail loudly on invalid input).
+	if serverName == "" {
+		return nil, fmt.Errorf(
+			"%w: ServerName must not be empty when Authz is set (Cedar resource-scoped policies require it)",
+			vmcp.ErrInvalidConfig)
 	}
 
 	factory := authorizers.GetFactory(string(authzCfg.Type))
@@ -122,8 +138,16 @@ func (a *cedarAdmission) FilterTools(
 }
 
 // AllowToolCall mirrors pkg/authz authorizeToolCall (tool_filter.go:64). Pass-through
-// meta-tools are exempt: their inner backend tool is authorized by the optimizer
-// decorator, matching their exemption through the HTTP middleware today.
+// meta-tools are exempt from the seam's decision, matching how they bypass the HTTP
+// authz path's tool checks today (#5438 AC: "passThroughTools remain exempt").
+//
+// call_tool parity note: the HTTP middleware's handleToolsCall (middleware.go:276)
+// additionally RE-TARGETS authorization at the inner backend tool wrapped inside a
+// call_tool request (reading args["tool_name"]). This seam deliberately does NOT
+// replicate that inner-tool authorization — it only exempts the meta-tool name, per
+// this task's scope. Ensuring the unwrapped inner call is authorized is a call-path
+// wiring requirement for #5441 (which removes the HTTP middleware): that wiring must
+// route the inner call through a path that consults this seam with the real tool name.
 func (a *cedarAdmission) AllowToolCall(
 	ctx context.Context, identity *auth.Identity, tool *vmcp.Tool, args map[string]any,
 ) (bool, error) {
@@ -240,11 +264,16 @@ func (allowAllAdmission) AllowPromptGet(_ context.Context, _ *auth.Identity, _ *
 
 // convertAnnotations maps vmcp.ToolAnnotations to authorizers.ToolAnnotations,
 // returning nil when no hint field is set so the adapter only writes annotation
-// ctx when a hint exists (matching the existing hasAnyHint gate). It mirrors
-// server.convertAnnotations (annotation_enrichment.go:92): an intentional,
-// temporary duplication — the server-side middleware copy is retired on the
-// domain path in #5441. Only authorization-relevant hints are mapped;
-// informational fields like Title are not used in policy evaluation.
+// ctx when a hint exists (matching the existing hasAnyHint gate). Only
+// authorization-relevant hints are mapped; informational fields like Title are not
+// used in policy evaluation.
+//
+// This duplicates server.convertAnnotations (annotation_enrichment.go:92) rather
+// than reusing it: that copy lives in package server, which imports this core
+// package, so importing it here would create a cycle (server -> core). Like the
+// filterHealthyBackends C2 duplication in this package, it is intentional and
+// temporary — #5441 retires the server-side middleware (and its copy) on the domain
+// path. Keep the two in sync until then.
 func convertAnnotations(ann *vmcp.ToolAnnotations) *authorizers.ToolAnnotations {
 	if ann == nil {
 		return nil
--- a/pkg/vmcp/core/admission.go
+++ b/pkg/vmcp/core/admission.go
@@ -1,0 +1,276 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package core
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+
+	"github.com/stacklok/toolhive/pkg/auth"
+	"github.com/stacklok/toolhive/pkg/authz/authorizers"
+	"github.com/stacklok/toolhive/pkg/vmcp"
+)
+
+// Admission decides whether an identity may see or use a capability. It wraps the
+// existing [authorizers.Authorizer] (Cedar in the vMCP path); it does NOT define a
+// new policy model.
+//
+// The seam re-platforms the authorization decision that runs as HTTP middleware
+// today (AuthzMiddleware + AnnotationEnrichmentMiddleware) into the core, so that
+// List* (filter) and Call/Read/Get (deny) enforce the SAME decision from one
+// source — closing the "list says yes / call says no" gap. Identity is an explicit
+// parameter (never read from ctx, vmcp anti-pattern #1); the adapter re-injects it
+// into the ctx the wrapped authorizer reads. A seam can only SUBTRACT reachability
+// (filter list output / refuse a call), never widen it ([VMCP] contract).
+type Admission interface {
+	// FilterTools returns the subset of tools the identity may call. Mirrors
+	// pkg/authz filterToolsByPolicy: a per-tool AuthorizeWithJWTClaims(call) using
+	// the tool's annotations; a per-tool authorizer error skips that tool
+	// (log-and-continue), it is not a hard failure.
+	FilterTools(ctx context.Context, identity *auth.Identity, tools []vmcp.Tool) ([]vmcp.Tool, error)
+	// AllowToolCall mirrors pkg/authz authorizeToolCall (MCPFeatureTool/Call),
+	// sourcing the tool's annotations for when-clause evaluation.
+	AllowToolCall(ctx context.Context, identity *auth.Identity, tool *vmcp.Tool, args map[string]any) (bool, error)
+	// FilterResources / AllowResourceRead use MCPFeatureResource/Read.
+	FilterResources(ctx context.Context, identity *auth.Identity, resources []vmcp.Resource) ([]vmcp.Resource, error)
+	AllowResourceRead(ctx context.Context, identity *auth.Identity, resource *vmcp.Resource) (bool, error)
+	// FilterPrompts / AllowPromptGet use MCPFeaturePrompt/Get.
+	FilterPrompts(ctx context.Context, identity *auth.Identity, prompts []vmcp.Prompt) ([]vmcp.Prompt, error)
+	AllowPromptGet(ctx context.Context, identity *auth.Identity, prompt *vmcp.Prompt) (bool, error)
+}
+
+// newAdmission builds the admission seam New wires into the core.
+//
+// A nil authzCfg means authorization is unconfigured and the seam is allow-all,
+// preserving today's `cfg.AuthzMiddleware != nil` guard (server.go:606): the
+// composition root only populates Authz when Cedar policies exist (mirroring
+// newCedarAuthzMiddleware's `len(Policies) > 0` check), so "no policies" reaches
+// the core as a nil Authz, not an empty authorizer.
+//
+// When authzCfg is set, the authorizer is built via the same registry path the
+// HTTP middleware uses (authorizers.GetFactory + CreateAuthorizer, the
+// construction inside newCedarAuthzMiddleware). The Cedar factory is registered by
+// pkg/authz's blank import, which the core package already pulls in.
+func newAdmission(
+	authzCfg *authorizers.Config, serverName string, passThroughTools map[string]struct{},
+) (Admission, error) {
+	if authzCfg == nil {
+		return allowAllAdmission{}, nil
+	}
+
+	factory := authorizers.GetFactory(string(authzCfg.Type))
+	if factory == nil {
+		return nil, fmt.Errorf("%w: unsupported authz type %q", vmcp.ErrInvalidConfig, authzCfg.Type)
+	}
+	authorizer, err := factory.CreateAuthorizer(authzCfg.RawConfig(), serverName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build admission authorizer: %w", err)
+	}
+	return newCedarAdmission(authorizer, passThroughTools), nil
+}
+
+// cedarAdmission enforces the wrapped [authorizers.Authorizer]'s decision. It is
+// named for the Cedar authorizer it backs in the vMCP path, but wraps the
+// single-method Authorizer interface generically. passThroughTools (optimizer
+// meta-tools) bypass the decision exactly as they do in the HTTP response filter.
+type cedarAdmission struct {
+	authorizer       authorizers.Authorizer
+	passThroughTools map[string]struct{}
+}
+
+// newCedarAdmission wraps an already-built authorizer. Separated from newAdmission
+// so tests can inject a stub authorizer (or a real cedar.NewCedarAuthorizer)
+// without round-tripping through the config/factory.
+func newCedarAdmission(a authorizers.Authorizer, passThroughTools map[string]struct{}) *cedarAdmission {
+	return &cedarAdmission{authorizer: a, passThroughTools: passThroughTools}
+}
+
+// FilterTools mirrors pkg/authz filterToolsByPolicy (tool_filter.go:19) and the
+// response filter's tool split (response_filter.go:306): pass-through tools are
+// kept unconditionally; every other tool is authorized for call with its
+// annotations injected, and a per-tool authorizer error skips that tool.
+func (a *cedarAdmission) FilterTools(
+	ctx context.Context, identity *auth.Identity, tools []vmcp.Tool,
+) ([]vmcp.Tool, error) {
+	ctx = auth.WithIdentity(ctx, identity)
+	// Instantiate so an empty result marshals as [] rather than null on the wire.
+	filtered := make([]vmcp.Tool, 0, len(tools))
+	for i := range tools {
+		tool := &tools[i]
+		if _, passThrough := a.passThroughTools[tool.Name]; passThrough {
+			filtered = append(filtered, *tool)
+			continue
+		}
+
+		toolCtx := ctx
+		if ann := convertAnnotations(tool.Annotations); ann != nil {
+			toolCtx = authorizers.WithToolAnnotations(toolCtx, ann)
+		}
+		allowed, err := a.authorizer.AuthorizeWithJWTClaims(
+			toolCtx, authorizers.MCPFeatureTool, authorizers.MCPOperationCall, tool.Name, nil)
+		if err != nil {
+			slog.Warn("admission: tool authorization check failed, skipping", "tool", tool.Name, "error", err)
+			continue
+		}
+		if allowed {
+			filtered = append(filtered, *tool)
+		}
+	}
+	return filtered, nil
+}
+
+// AllowToolCall mirrors pkg/authz authorizeToolCall (tool_filter.go:64). Pass-through
+// meta-tools are exempt: their inner backend tool is authorized by the optimizer
+// decorator, matching their exemption through the HTTP middleware today.
+func (a *cedarAdmission) AllowToolCall(
+	ctx context.Context, identity *auth.Identity, tool *vmcp.Tool, args map[string]any,
+) (bool, error) {
+	if _, passThrough := a.passThroughTools[tool.Name]; passThrough {
+		return true, nil
+	}
+	ctx = auth.WithIdentity(ctx, identity)
+	if ann := convertAnnotations(tool.Annotations); ann != nil {
+		ctx = authorizers.WithToolAnnotations(ctx, ann)
+	}
+	return a.authorizer.AuthorizeWithJWTClaims(
+		ctx, authorizers.MCPFeatureTool, authorizers.MCPOperationCall, tool.Name, args)
+}
+
+// FilterResources mirrors the response filter's filterResourcesResponse
+// (response_filter.go:407): per-resource read authorization, skipping on error.
+func (a *cedarAdmission) FilterResources(
+	ctx context.Context, identity *auth.Identity, resources []vmcp.Resource,
+) ([]vmcp.Resource, error) {
+	ctx = auth.WithIdentity(ctx, identity)
+	filtered := make([]vmcp.Resource, 0, len(resources))
+	for i := range resources {
+		allowed, err := a.authorizer.AuthorizeWithJWTClaims(
+			ctx, authorizers.MCPFeatureResource, authorizers.MCPOperationRead, resources[i].URI, nil)
+		if err != nil {
+			slog.Warn("admission: resource authorization check failed, skipping",
+				"resource", resources[i].URI, "error", err)
+			continue
+		}
+		if allowed {
+			filtered = append(filtered, resources[i])
+		}
+	}
+	return filtered, nil
+}
+
+// AllowResourceRead authorizes a single resource read (MCPFeatureResource/Read).
+func (a *cedarAdmission) AllowResourceRead(
+	ctx context.Context, identity *auth.Identity, resource *vmcp.Resource,
+) (bool, error) {
+	ctx = auth.WithIdentity(ctx, identity)
+	return a.authorizer.AuthorizeWithJWTClaims(
+		ctx, authorizers.MCPFeatureResource, authorizers.MCPOperationRead, resource.URI, nil)
+}
+
+// FilterPrompts mirrors the response filter's filterPromptsResponse
+// (response_filter.go:346): per-prompt get authorization, skipping on error.
+func (a *cedarAdmission) FilterPrompts(
+	ctx context.Context, identity *auth.Identity, prompts []vmcp.Prompt,
+) ([]vmcp.Prompt, error) {
+	ctx = auth.WithIdentity(ctx, identity)
+	filtered := make([]vmcp.Prompt, 0, len(prompts))
+	for i := range prompts {
+		allowed, err := a.authorizer.AuthorizeWithJWTClaims(
+			ctx, authorizers.MCPFeaturePrompt, authorizers.MCPOperationGet, prompts[i].Name, nil)
+		if err != nil {
+			slog.Warn("admission: prompt authorization check failed, skipping",
+				"prompt", prompts[i].Name, "error", err)
+			continue
+		}
+		if allowed {
+			filtered = append(filtered, prompts[i])
+		}
+	}
+	return filtered, nil
+}
+
+// AllowPromptGet authorizes a single prompt get (MCPFeaturePrompt/Get).
+func (a *cedarAdmission) AllowPromptGet(
+	ctx context.Context, identity *auth.Identity, prompt *vmcp.Prompt,
+) (bool, error) {
+	ctx = auth.WithIdentity(ctx, identity)
+	return a.authorizer.AuthorizeWithJWTClaims(
+		ctx, authorizers.MCPFeaturePrompt, authorizers.MCPOperationGet, prompt.Name, nil)
+}
+
+// allowAllAdmission is the no-op seam used when authorization is unconfigured. It
+// preserves today's behavior when AuthzMiddleware is nil: Filter* return their
+// input unchanged and Allow* return true (the nil-authorizer no-ops at
+// tool_filter.go:20,67).
+type allowAllAdmission struct{}
+
+func (allowAllAdmission) FilterTools(
+	_ context.Context, _ *auth.Identity, tools []vmcp.Tool,
+) ([]vmcp.Tool, error) {
+	return tools, nil
+}
+
+func (allowAllAdmission) AllowToolCall(
+	_ context.Context, _ *auth.Identity, _ *vmcp.Tool, _ map[string]any,
+) (bool, error) {
+	return true, nil
+}
+
+func (allowAllAdmission) FilterResources(
+	_ context.Context, _ *auth.Identity, resources []vmcp.Resource,
+) ([]vmcp.Resource, error) {
+	return resources, nil
+}
+
+func (allowAllAdmission) AllowResourceRead(_ context.Context, _ *auth.Identity, _ *vmcp.Resource) (bool, error) {
+	return true, nil
+}
+
+func (allowAllAdmission) FilterPrompts(
+	_ context.Context, _ *auth.Identity, prompts []vmcp.Prompt,
+) ([]vmcp.Prompt, error) {
+	return prompts, nil
+}
+
+func (allowAllAdmission) AllowPromptGet(_ context.Context, _ *auth.Identity, _ *vmcp.Prompt) (bool, error) {
+	return true, nil
+}
+
+// convertAnnotations maps vmcp.ToolAnnotations to authorizers.ToolAnnotations,
+// returning nil when no hint field is set so the adapter only writes annotation
+// ctx when a hint exists (matching the existing hasAnyHint gate). It mirrors
+// server.convertAnnotations (annotation_enrichment.go:92): an intentional,
+// temporary duplication — the server-side middleware copy is retired on the
+// domain path in #5441. Only authorization-relevant hints are mapped;
+// informational fields like Title are not used in policy evaluation.
+func convertAnnotations(ann *vmcp.ToolAnnotations) *authorizers.ToolAnnotations {
+	if ann == nil {
+		return nil
+	}
+	if ann.ReadOnlyHint == nil && ann.DestructiveHint == nil &&
+		ann.IdempotentHint == nil && ann.OpenWorldHint == nil {
+		return nil
+	}
+	return &authorizers.ToolAnnotations{
+		ReadOnlyHint:    ann.ReadOnlyHint,
+		DestructiveHint: ann.DestructiveHint,
+		IdempotentHint:  ann.IdempotentHint,
+		OpenWorldHint:   ann.OpenWorldHint,
+	}
+}
+
+// findAdvertisedTool returns a pointer to the advertised tool with the given name,
+// or nil. CallTool uses it to source the tool's annotations for the admission
+// decision — mirroring the annotation cache the HTTP middleware populates from
+// tools/list. A name absent from the advertised set carries no annotations
+// (nil), and routing remains the authority on whether the call resolves.
+func findAdvertisedTool(tools []vmcp.Tool, name string) *vmcp.Tool {
+	for i := range tools {
+		if tools[i].Name == name {
+			return &tools[i]
+		}
+	}
+	return nil
+}
--- a/pkg/vmcp/core/admission.go
+++ b/pkg/vmcp/core/admission.go
@@ -58,6 +58,9 @@ type Admission interface {
 // normalization the live path performs at construction time (e.g. defaulting an
 // empty EntitiesJSON to "[]", incoming.go:128) belongs to whoever builds
 // authzCfg, not to this generic factory path, which consumes RawConfig as-is.
+// newCedarAuthzMiddleware (incoming.go:114) is the sibling construction path; the
+// two must stay in lockstep (notably the empty-serverName check below) until #5441
+// collapses them into a single shared constructor.
 func newAdmission(
 	authzCfg *authorizers.Config, serverName string, passThroughTools map[string]struct{},
 ) (Admission, error) {
@@ -111,7 +114,10 @@ func (a *cedarAdmission) FilterTools(
 	ctx context.Context, identity *auth.Identity, tools []vmcp.Tool,
 ) ([]vmcp.Tool, error) {
 	ctx = auth.WithIdentity(ctx, identity)
-	// Instantiate so an empty result marshals as [] rather than null on the wire.
+	// A filter returns a subset, so build a fresh non-nil slice — mirroring
+	// pkg/authz filterToolsByPolicy (its configured-authorizer path is also
+	// non-nil; its nil-authorizer no-op, like the allow-all seam here, returns the
+	// input as-is). nil-vs-[] wire normalization is the Serve layer's concern.
 	filtered := make([]vmcp.Tool, 0, len(tools))
 	for i := range tools {
 		tool := &tools[i]
--- a/pkg/vmcp/core/admission.go
+++ b/pkg/vmcp/core/admission.go
@@ -44,23 +44,27 @@ type Admission interface {
 // newAdmission builds the admission seam New wires into the core.
 //
 // A nil authzCfg means authorization is unconfigured and the seam is allow-all,
-// preserving today's `cfg.AuthzMiddleware != nil` guard (server.go:606): the
-// composition root only populates Authz when Cedar policies exist (mirroring
-// newCedarAuthzMiddleware's `len(Policies) > 0` check), so "no policies" reaches
-// the core as a nil Authz, not an empty authorizer.
+// preserving today's `AuthzMiddleware != nil` guard: the composition root only
+// populates Authz when Cedar policies exist (mirroring newCedarAuthzMiddleware's
+// `len(Policies) > 0` check), so "no policies" reaches the core as a nil Authz.
+//
+// A non-nil authzCfg with zero policies is NOT treated as allow-all here: it falls
+// through to CreateAuthorizer, which returns ErrNoPolicies, so the core fails
+// CLOSED. This is a deliberate divergence from the live HTTP path, which allows-all
+// for the same input. Fail-closed is the safer default; #5441 reconciles the two
+// when it collapses the construction paths into one shared constructor.
 //
 // When authzCfg is set, the authorizer is built via the same registry path the
-// HTTP middleware uses (authorizers.GetFactory + CreateAuthorizer, the
-// construction inside newCedarAuthzMiddleware). The Cedar factory is registered by
-// pkg/authz's blank import, which the core package already pulls in.
+// HTTP middleware uses (authorizers.GetFactory + CreateAuthorizer, the construction
+// inside newCedarAuthzMiddleware). The Cedar factory is registered by pkg/authz's
+// blank import, which the core package already pulls in.
 //
 // authzCfg is an already-built, authorizer-agnostic config: the cedar-specific
-// normalization the live path performs at construction time (e.g. defaulting an
-// empty EntitiesJSON to "[]", incoming.go:128) belongs to whoever builds
-// authzCfg, not to this generic factory path, which consumes RawConfig as-is.
-// newCedarAuthzMiddleware (incoming.go:114) is the sibling construction path; the
-// two must stay in lockstep (notably the empty-serverName check below) until #5441
-// collapses them into a single shared constructor.
+// normalization newCedarAuthzMiddleware performs at construction time (e.g.
+// defaulting an empty EntitiesJSON to "[]") belongs to whoever builds authzCfg, not
+// to this generic factory path, which consumes RawConfig as-is. newCedarAuthzMiddleware
+// is the sibling construction path; the two must stay in lockstep (notably the
+// empty-serverName check below) until #5441 collapses them.
 func newAdmission(
 	authzCfg *authorizers.Config, serverName string, passThroughTools map[string]struct{},
 ) (Admission, error) {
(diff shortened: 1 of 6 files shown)


pkg/vmcp/core/admission.go
```diff
@@ -0,0 +1,311 @@
...
+// authz path's tool checks today (#5438 AC: "passThroughTools remain exempt").
+//
+// call_tool parity note: the HTTP middleware's handleToolsCall (middleware.go:276)
+// additionally RE-TARGETS authorization at the inner backend tool wrapped inside a
+// call_tool request (reading args["tool_name"]). This seam deliberately does NOT
+// replicate that inner-tool authorization — it only exempts the meta-tool name, per
+// this task's scope. Ensuring the unwrapped inner call is authorized is a call-path
+// wiring requirement for #5441 (which removes the HTTP middleware): that wiring must
+// route the inner call through a path that consults this seam with the real tool name.
+func (a *cedarAdmission) AllowToolCall(
```
tgrunnagle, line 177: **HIGH (non-blocking for this PR).** The HTTP middleware's `handleToolsCall` re-targets authorization at the inner backend tool named in `args["tool_name"]`; this seam exempts the `call_tool` meta-tool name and never inspects the inner target. That is correct for this task's scope (AC R1.4) and **not exploitable today** because `core.New` has no non-test callers — the risk is purely latent.

The mitigation rests entirely on follow-up discipline. **Recommendation:** make #5441's acceptance criteria *explicitly* require routing the inner `call_tool` target through `Admission.AllowToolCall` (with the real tool name), with a test for a `call_tool`-wrapped denied tool. A tracked failing/skipped guard test would prevent the gap from shipping silently. The doc comment here is good; the follow-up just needs to live in the issue, not only in code.

pkg/vmcp/core/core_calls.go
```diff
@@ -0,0 +1,197 @@
...
+	// HTTP middleware populates from tools/list) so annotation-gated policies
+	// evaluate. advertisedTools includes composites, so their annotations are sourced
+	// too. A name absent from the recomputed advertised set carries no annotations:
+	// an annotation-gated decision then evaluates with no hints (and may deny) — but
+	// such a name is also unroutable below, so it would not resolve regardless.
+	tool := findAdvertisedTool(c.advertisedTools(agg), name)
+	if tool == nil {
+		tool = &vmcp.Tool{Name: name}
+	}
+	if allowed, err := c.admission.AllowToolCall(ctx, identity, tool, argsCopy); err != nil {
```
tgrunnagle, line 63: **MEDIUM.** On the `err != nil` branch this returns the wrapped authorizer's error verbatim (e.g. Cedar `ErrMissingPrincipal` for a nil identity); only the `!allowed` branch wraps `vmcp.ErrAuthorizationFailed`. The future Serve adapter classifies core errors via `errors.Is`, so a raw Cedar error would fall through to the generic "call failed" path and be indistinguishable from a transport failure (and would leak the internal error type). Test-only today, but #5441 inherits it. Consider classifying the error branch as `ErrAuthorizationFailed` here, and decide deliberately whether the raw Cedar string should reach the client. Same pattern in `ReadResource` and `GetPrompt`.

pkg/vmcp/core/admission_test.go
```diff
@@ -0,0 +1,657 @@
...
+
+	id := &auth.Identity{
+		PrincipalInfo:  auth.PrincipalInfo{Subject: "user", Claims: map[string]interface{}{"secret": secret}},
+		Token:          secret,
+		UpstreamTokens: map[string]string{"github": secret},
+	}
+
+	var buf bytes.Buffer
+	prev := slog.Default()
+	slog.SetDefault(slog.New(slog.NewTextHandler(&buf, &slog.HandlerOptions{Level: slog.LevelDebug})))
```
tgrunnagle: **MEDIUM.** This swaps the process-global `slog` default over a non-thread-safe `bytes.Buffer`. Sibling `t.Parallel()` tests in this package (e.g. `TestCedarAdmission_FilterTools` skip case, `TestCedarAdmission_WithCedarAuthorizer` nil-identity skip) emit `slog.Warn` to the default logger and can run concurrently with this serial test, making the buffer write/read a `-race`-detectable hazard. The assertion itself is fine and non-vacuous. **Recommendation:** have the seam log through an injectable `*slog.Logger` and pass a buffer-backed logger into the test, eliminating the global swap. At minimum, verify under `go test -race -count=5 ./pkg/vmcp/core/...`.

pkg/vmcp/core/admission_test.go
```diff
@@ -0,0 +1,657 @@
...
+		assert.Equal(t, "user123", c.identitySubject)
+	}
+	// Annotations are injected only for the tool that carries a hint.
+	byName := map[string]mockCall{mock.calls[0].resourceID: mock.calls[0], mock.calls[1].resourceID: mock.calls[1]}
+	require.NotNil(t, byName["hinted"].annotations)
+	assert.Equal(t, boolPtr(true), byName["hinted"].annotations.ReadOnlyHint)
+	assert.Nil(t, byName["plain"].annotations, "no annotation ctx written when the tool has no hints")
+}
+
+func TestCedarAdmission_AllowToolCall(t *testing.T) {
```
tgrunnagle, line 207: **MEDIUM (test gap).** `mockAuthorizer` discards the `args` parameter (`_ map[string]interface{}`), so no test verifies that call arguments actually flow from `CallTool` → `AllowToolCall` → the authorizer. The HTTP path uses these args for argument-gated Cedar policies, and the issue's required list includes mirroring `TestAuthorizeToolCall_WithArguments`. Add an `args` field to `mockCall`, capture the 5th parameter, and assert the args map reaches the authorizer (optionally with a real Cedar arg-gated policy).

pkg/vmcp/core/admission.go
```diff
@@ -0,0 +1,311 @@
...
+// authorization-relevant hints are mapped; informational fields like Title are not
+// used in policy evaluation.
+//
+// This duplicates server.convertAnnotations (annotation_enrichment.go:92) rather
+// than reusing it: that copy lives in package server, which imports this core
+// package, so importing it here would create a cycle (server -> core). Like the
+// filterHealthyBackends C2 duplication in this package, it is intentional and
+// temporary — #5441 retires the server-side middleware (and its copy) on the domain
+// path. Keep the two in sync until then.
+func convertAnnotations(ann *vmcp.ToolAnnotations) *authorizers.ToolAnnotations {
```
tgrunnagle, line 300: **LOW (drift).** This copy adds `if ann == nil { return nil }`; the server copy at `pkg/vmcp/server/annotation_enrichment.go` lacks it and would panic on a nil `ann`. The two share a name and a "keep the two in sync" comment but are *already* divergent — exactly the parallel-functions-that-drift hazard from `go-style.md`. Cheapest fix: add the same nil-guard to the server copy now so the comment is truthful, and confirm #5441's scope explicitly deletes the server copy. (The core version is the safer of the two — keep its guard in the eventual consolidated converter.)

pkg/vmcp/core/admission.go
```diff
@@ -0,0 +1,311 @@
...
+// normalization the live path performs at construction time (e.g. defaulting an
+// empty EntitiesJSON to "[]", incoming.go:128) belongs to whoever builds
+// authzCfg, not to this generic factory path, which consumes RawConfig as-is.
+// newCedarAuthzMiddleware (incoming.go:114) is the sibling construction path; the
+// two must stay in lockstep (notably the empty-serverName check below) until #5441
+// collapses them into a single shared constructor.
+func newAdmission(
+	authzCfg *authorizers.Config, serverName string, passThroughTools map[string]struct{},
+) (Admission, error) {
+	if authzCfg == nil {
```
tgrunnagle, line 78: **LOW.** Only `authzCfg == nil` is treated as allow-all. The live sibling `newCedarAuthzMiddleware` guards on `len(Policies) > 0` and treats a non-nil-but-empty-policies config as "no middleware" (allow-all). Here, a non-nil cedar config with zero policies falls through to `CreateAuthorizer`, which returns `ErrNoPolicies` — so the core **fails closed** while the HTTP path **allows all** for the same input. Fail-closed is the safer default, but it's an undocumented divergence in the lockstep pair. Either mirror the live guard or add a one-line note so #5441's collapse-into-one-constructor reconciles it deliberately.

pkg/vmcp/core/core_calls.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	// Admission deny: enforce the same decision ListTools filters on, sourcing the
+	// tool's annotations from the advertised set (mirroring the annotation cache the
+	// HTTP middleware populates from tools/list) so annotation-gated policies
+	// evaluate. advertisedTools includes composites, so their annotations are sourced
+	// too. A name absent from the recomputed advertised set carries no annotations:
+	// an annotation-gated decision then evaluates with no hints (and may deny) — but
+	// such a name is also unroutable below, so it would not resolve regardless.
+	tool := findAdvertisedTool(c.advertisedTools(agg), name)
```
tgrunnagle, line 53: **LOW (test gap).** The `findAdvertisedTool` nil branch — called name absent from the advertised set, synthesizing `&vmcp.Tool{Name: name}` with no annotations — is never exercised end-to-end through `CallTool`, nor is `findAdvertisedTool` unit-tested directly. The inline comment reasons such a name is "unroutable below anyway," but that's unverified. Add a case calling `CallTool` with a name absent from aggregated `Tools` but present in the routing table under an annotation-gated policy (assert the no-hint deny), or a test proving routing rejects it first so the branch is provably dead. A small table test on `findAdvertisedTool` (present / absent / empty) would also help.

pkg/vmcp/core/admission.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+	}
+	return filtered, nil
+}
+
+// AllowToolCall mirrors pkg/authz authorizeToolCall (tool_filter.go:64). Pass-through
+// meta-tools are exempt from the seam's decision, matching how they bypass the HTTP
+// authz path's tool checks today (#5438 AC: "passThroughTools remain exempt").
+//
+// call_tool parity note: the HTTP middleware's handleToolsCall (middleware.go:276)
```
tgrunnagle: **LOW (citation drift).** `handleToolsCall` is declared at `middleware.go:266`; line 276 falls mid-body. Consider citing by symbol name rather than `file:line` (these break silently on any edit and can't be caught by the compiler). Suggested literal fix:
```suggestion
// call_tool parity note: the HTTP middleware's handleToolsCall (middleware.go:266)
```

pkg/vmcp/core/admission.go
```diff
@@ -0,0 +1,311 @@
...
+// the core as a nil Authz, not an empty authorizer.
+//
+// When authzCfg is set, the authorizer is built via the same registry path the
+// HTTP middleware uses (authorizers.GetFactory + CreateAuthorizer, the
+// construction inside newCedarAuthzMiddleware). The Cedar factory is registered by
+// pkg/authz's blank import, which the core package already pulls in.
+//
+// authzCfg is an already-built, authorizer-agnostic config: the cedar-specific
+// normalization the live path performs at construction time (e.g. defaulting an
+// empty EntitiesJSON to "[]", incoming.go:128) belongs to whoever builds
```
tgrunnagle: **LOW (citation drift).** The default-to-`"[]"` assignment is at `incoming.go:130`; line 128 is the read of `EntitiesJSON`. Suggested literal fix:
```suggestion
// empty EntitiesJSON to "[]", incoming.go:130) belongs to whoever builds
```

tgrunnagle (commented):
## Multi-agent review — Cedar admission seam (R1)

Reviewed by 5 specialist agents (security/authz parity, Go correctness/concurrency, test coverage, vMCP architecture, general quality). Codex cross-review skipped (CLI not installed).

**Verdict: COMMENT — no blocking issues.** The approach is sound and the implementation is high quality. I independently confirmed the central *no-behavior-change* claim: `core.New` has **zero non-test callers**, so the seam is genuinely test-only in Phase 1 and the shipped `server.New` path (still HTTP-middleware-enforced) is untouched — no double enforcement. Faithful mirror of the legacy `pkg/authz` filter/call logic; identity handling is exemplary (explicit param, re-injected only as an adapter detail, never logged, non-vacuous redaction test); subtract-only and domain-typing contracts hold; `New` builds admission before acquiring the state-store goroutine so a bad policy fails fast.

### Findings (consensus-scored)

| Sev | Finding |
|-----|---------|
| HIGH | `call_tool` inner-tool authz deliberately dropped — latent bypass **if** #5441 wires the core live without honoring the documented follow-up (not exploitable today; core is test-only) |
| MEDIUM | Authorizer error returned raw on the `err` branch, unclassified as `ErrAuthorizationFailed` |
| MEDIUM | `TestAdmission_IdentityNeverLogged` swaps the global `slog` default → `-race` hazard vs `t.Parallel()` siblings |
| MEDIUM | No test asserts call args flow through the seam into the authorizer (missing `WithArguments` mirror) |
| MEDIUM | Issue-required `server.New` behavioral-parity test absent and not flagged as deferred |
| MEDIUM→LOW | `convertAnnotations` core copy already drifted from the server copy (added nil-guard); "keep in sync" comment is no longer true |
| LOW | `newAdmission` fails closed on a non-nil-but-empty-policies config while the live path allows-all (undocumented divergence) |
| LOW | `findAdvertisedTool` nil branch untested end-to-end and directly |
| LOW | Drifted doc-comment line citations (`middleware.go:276`→266, `incoming.go:128`→130) |
| LOW | Nonexistent tool under a restrictive policy returns `ErrAuthorizationFailed`, not `ErrNotFound` (Call vs Lookup error divergence — note: this is an info-disclosure *improvement*; just document the deliberate divergence) |
| LOW | `CreateAuthorizer` failure wrapped without the `ErrInvalidConfig` sentinel its sibling paths use |
| LOW | `ServerName` is a hard precondition when `Authz` is set, but documented only as "consulted when" |
| INFO | Line-number citations are a maintainability smell (prefer package-qualified symbol names); `Admission` exported though used only within `core`; `cedarAdmission` name is authorizer-agnostic |

### Most valuable concrete actions before merge
1. **Fix the `slog` global-mutation test race** (inject a logger instead of swapping the default).
2. **Align the two `convertAnnotations` copies** (add the nil-guard to the server copy) so the "keep in sync" comment is truthful.
3. **Carry the `call_tool` inner-authz follow-up into #5441's acceptance criteria**, not just a code comment.

Inline comments below have the detail. Nice work — the test suite covers R1.1–R1.5 thoroughly and non-vacuously.

🤖 Reviewed with multi-agent [Claude Code](https://claude.com/claude-code) pr-review

pkg/vmcp/core/admission.go
```diff
@@ -0,0 +1,311 @@
...
+			vmcp.ErrInvalidConfig)
+	}
+
+	factory := authorizers.GetFactory(string(authzCfg.Type))
+	if factory == nil {
+		return nil, fmt.Errorf("%w: unsupported authz type %q", vmcp.ErrInvalidConfig, authzCfg.Type)
+	}
+	authorizer, err := factory.CreateAuthorizer(authzCfg.RawConfig(), serverName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build admission authorizer: %w", err)
```
tgrunnagle: **LOW (consistency).** The two sibling construction failures above classify with `vmcp.ErrInvalidConfig` (unsupported type, empty server name), but a `CreateAuthorizer` failure (e.g. a malformed `RawConfig`/bad policy) wraps without the sentinel. A caller doing `errors.Is(err, vmcp.ErrInvalidConfig)` would catch the first two but not a bad-policy parse failure — also fundamentally an invalid-config condition. Consider `fmt.Errorf("%w: failed to build admission authorizer: %w", vmcp.ErrInvalidConfig, err)` for uniform classification.

pkg/vmcp/core/core.go
```diff
@@ -133,10 +133,24 @@ type Config struct {
...
 
-	// Authz feeds the admission seam (wired in a later change). A nil Authz
-	// means authorization is unconfigured (allow-all).
+	// Authz feeds the admission seam New builds. A nil Authz means authorization
+	// is unconfigured (allow-all), matching today's `AuthzMiddleware != nil` guard:
+	// the composition root only populates this when Cedar policies exist (mirroring
+	// newCedarAuthzMiddleware's `len(Policies) > 0` check).
 	Authz *authz.Config
 
+	// ServerName is the VirtualMCPServer name used as the Cedar resource entity
```
tgrunnagle: **LOW (doc).** `newAdmission` returns `ErrInvalidConfig` when `Authz` is set but `ServerName == ""` — so `ServerName` is a hard **precondition** when `Authz` is non-nil, not merely "consulted." When #5441 wires the core into `server.New`, any composition root that sets `Authz` without `ServerName` gets a hard construction failure. Recommend strengthening the `Config.Authz` doc to state `ServerName` is *required* (not optional) when `Authz` is non-nil, and noting `ServerName`/`PassThroughTools` are silently ignored when `Authz` is nil.

pkg/vmcp/core/core_calls.go
tgrunnagle: Addressed in d048950e — the `err != nil` branch in `CallTool`/`ReadResource`/`GetPrompt` now wraps `vmcp.ErrAuthorizationFailed` (fail closed, matching the live `authorizeAndServe`), preserving the underlying error in the chain for diagnostics while letting the Serve adapter decide what reaches the client.

pkg/vmcp/core/admission_test.go
tgrunnagle: Addressed in d048950e — the seam now logs through an injected `*slog.Logger` (`withLogger` option); the test injects a buffer-backed logger instead of swapping the global default, so it's parallel-safe. Verified clean under `go test -race -count=5 ./pkg/vmcp/core/...`.

pkg/vmcp/core/admission_test.go
tgrunnagle: Addressed in d048950e — `mockCall` now captures the args with an assertion that they reach the authorizer, plus an end-to-end Cedar arg-gated test through `CallTool` (`TestAdmission_CallToolForwardsArgsToAuthorizer`) mirroring `TestAuthorizeToolCall_WithArguments`.

pkg/vmcp/core/admission.go
tgrunnagle: Addressed in d048950e — added the same nil-guard to the server `convertAnnotations` copy so the two are identical and the "keep in sync" comment is truthful; both are retired together in #5441.

pkg/vmcp/core/admission.go
tgrunnagle: Addressed in d048950e — documented the deliberate fail-closed divergence in `newAdmission` (a non-nil/empty-policies config returns `ErrNoPolicies` here vs allow-all on the live path; #5441 reconciles when collapsing the constructors). Kept fail-closed as the safer default rather than mirroring the live allow-all.

pkg/vmcp/core/core_calls.go
tgrunnagle: Addressed in d048950e — added a direct `TestFindAdvertisedTool` table test and an end-to-end nil-branch test (`TestAdmission_CallToolNilAdvertisedToolDeniesUnderAnnotationPolicy`: a routable name absent from the advertised set is evaluated with no hints and denied). Also corrected the inaccurate "unroutable below" comment — such a name IS routable; it's just denied by the no-hint annotation eval.

pkg/vmcp/core/admission.go
tgrunnagle: Addressed in d048950e — rather than bump the line number (which drifts again), I dropped it and cite `handleToolsCall` by symbol, per your own suggestion and the INFO note.

pkg/vmcp/core/admission.go
tgrunnagle: Addressed in d048950e — same approach as the other citation: dropped the line number and cite the `newCedarAuthzMiddleware` EntitiesJSON default by symbol.

pkg/vmcp/core/admission.go
tgrunnagle: Addressed in d048950e — `CreateAuthorizer` failures now wrap `vmcp.ErrInvalidConfig` for uniform classification with the unsupported-type and empty-serverName errors.

pkg/vmcp/core/core.go
tgrunnagle: Addressed in d048950e — strengthened the `Config.Authz`/`ServerName` docs to state `ServerName` is required when `Authz` is non-nil (New fails with `ErrInvalidConfig` otherwise) and ignored when `Authz` is nil.

pkg/vmcp/core/admission.go
tgrunnagle: Tracked: added an explicit acceptance criterion to #5441 requiring the inner `call_tool` target be authorized through `Admission.AllowToolCall` (with a guard test for a `call_tool`-wrapped denied tool), since the seam intentionally only exempts the meta-tool name (AC R1.4). No code change in this PR — the `AllowToolCall` doc already documents the deferral. (Noted there that if the direct-`VMCP.CallTool` switch lands in #5442 instead, the criterion carries there.)

tgrunnagle:
On the **INFO** row (no inline thread):
- **Line-number citation smell** — partially addressed in d048950e: the two flagged citations (`handleToolsCall`, the `EntitiesJSON` default) and the `newAdmission` doc block now cite by symbol instead of `file:line`. The remaining mirror-citations to `pkg/authz` functions (`filterToolsByPolicy`, `authorizeToolCall`, the response-filter helpers) are kept as precise anchors to the legacy logic the seam mirrors; happy to strip those too if you'd prefer fully symbol-only.
- **`Admission` exported though used only within `core`** — intentional: it's part of the core's published contract (the `VMCP` interface doc references "the same admission filter") and leaves room for a future Serve-layer decorator. The implementations (`cedarAdmission`, `allowAllAdmission`) stay unexported.
- **`cedarAdmission` name is authorizer-agnostic** — kept to match the issue's spec naming; the struct doc notes it wraps the generic single-method `Authorizer` (Cedar in the vMCP path).

These are non-changes with rationale rather than fixes; let me know if you'd like any of them adjusted.

pkg/vmcp/core/core.go
```diff
@@ -133,10 +133,26 @@ type Config struct {
...
+	// the HTTP authz middleware. It is REQUIRED when Authz is non-nil (New returns
+	// vmcp.ErrInvalidConfig on an empty value, since Cedar resource-scoped policies
+	// depend on it) and ignored when Authz is nil.
+	ServerName string
+
+	// PassThroughTools names the optimizer meta-tools (find_tool/call_tool) that are
+	// exempt from admission filtering and denial, mirroring how they bypass the HTTP
+	// authz response filter today. Nil when the optimizer is disabled; ignored when
+	// Authz is nil (the allow-all seam exempts everything).
+	PassThroughTools map[string]struct{}
```
jerm-dro: `blocker:` Drop `PassThroughTools` from this PR and explicitly defer optimizer support.

As it stands, the field carves a hole in admission for the optimizer's meta-tools but covers only the list-time visibility — not the `call_tool` inner-target authorization, the `find_tool` response filtering, or inner-target annotation sourcing that the HTTP middleware does today. #5441's ACs now track the `call_tool` inner-authz piece (good), but `find_tool` response filtering and annotation sourcing aren't tracked anywhere I can find. Shipping the partial commitment locks in "the seam knows about meta-tools" as the implicit answer without scoping the full work, and creates a deferral surface that can land incomplete.

**Concrete ask for this PR:**
1. Remove `PassThroughTools` from `Config`, `cedarAdmission`, `newAdmission`, and the corresponding test (`TestCedarAdmission_PassThroughExempt`).
2. At the composition root that will wire `core.New` in #5441, fail-fast with `ErrInvalidConfig` when `Authz != nil` and the optimizer is enabled. Until the optimizer-admission integration lands as its own focused PR, the HTTP middleware stays on the optimizer path and the new seam is unambiguously unsupported with the optimizer. (`core.New` itself doesn't need to know about the optimizer — the check belongs at the composition root.)

**Worth considering for the eventual optimizer-admission PR (informational, not for this PR):** `passThroughTools` has always been tech debt — a quick fix that patches exactly the gaps above by carving optimizer-specific exemptions into a generic authorization layer. The long-term ideal is for *fewer* things to know the optimizer exists, not more. One promising direction: model the optimizer as a `VMCP` **decorator** that wraps whatever `VMCP` `core.New` returns (or composes at the root), sitting above admission and any user decorators. Inside, the optimizer translates meta-tools into real ones — `CallTool("call_tool", {tool_name, arguments})` becomes `inner.CallTool(toolName, arguments, meta)`; `CallTool("find_tool", ...)` post-filters its search through `inner.ListTools`; `ListTools` replaces the inner set with `find_tool + call_tool`. The inner `VMCP` (with admission) sees nothing but real tool names and applies the policy uniformly. Auth becomes optimizer-unaware, optimizer becomes auth-unaware, and `PassThroughTools` disappears from every layer it currently touches (the admission seam, `cli/serve.go`, the HTTP middleware). Worth weighing against the explicit-exemption alternative when the optimizer-admission PR is designed.

jerm-dro (changes requested):
Asking to defer optimizer support cleanly — details inline.

pkg/vmcp/core/core.go
tgrunnagle: Addressed in 1602e4ff. Removed `PassThroughTools` entirely — from `Config`, `cedarAdmission`, `newAdmission`, and deleted `TestCedarAdmission_PassThroughExempt`. The seam is now optimizer-agnostic: it authorizes every capability uniformly by name, no meta-tool carve-out.

Per ask #2, #5441's acceptance criteria are updated:
- **Replaced** the prior call_tool-inner-authz AC with a **composition-root fail-fast** — when `Authz` is set and the optimizer is enabled, construction fails with `vmcp.ErrInvalidConfig`, so the combination never silently routes through the seam; the optimizer path keeps the HTTP middleware. `core.New` stays optimizer-unaware (check is at the composition root), with a test asserting the failure.
- **Recorded** the full optimizer-admission integration (call_tool inner-target authz, find_tool response filtering, inner-target annotation sourcing) as its own focused PR in #5441's Out of Scope — explicitly noting the `VMCP`-decorator direction you proposed (optimizer wraps the admission-bearing core; auth and optimizer stay mutually unaware; `PassThroughTools` disappears from every layer). That's a cleaner end state; logged so the integration PR weighs it against explicit exemptions.

Side benefit: production diff drops to ~419 net lines.

pkg/vmcp/core/admission.go
tgrunnagle: Superseded by the resolution to r3377688855: rather than have this seam exempt the meta-tool names and rely on a later phase to authorize the inner target, `PassThroughTools` was removed entirely in 1602e4ff — the seam no longer special-cases the optimizer at all. #5441 now **fails fast** on (Authz + optimizer), and the inner-target authorization folds into the deferred optimizer-admission PR (tracked in #5441's Out of Scope).

jerm-dro (approved):

PR #5466, opened 2026-06-08
Add passthroughHeaders to VirtualMCPServer for header forwarding

### LINKED ISSUE #3958: vMCP: add header_passthrough outgoing auth strategy to forward dynamic incoming headers to backends
## Problem

When a client calls vMCP with dynamic request headers (e.g. `x-mcp-api-key`, `x-mcp-auth`), those headers are silently dropped. The vMCP creates a new HTTP client per backend call and only forwards pre-configured static auth headers — the original incoming headers never reach the backend.

This breaks use cases where the backend MCP server requires per-user/per-session headers for authentication (e.g. an external API that is API-key filtered and the key varies per client).

**Observed**: Backend receives requests with only vMCP-generated headers.  
**Expected**: Backend receives selected incoming headers forwarded transparently.

## Current workarounds

- `header_injection`: Only supports static pre-configured values — unusable for dynamic per-client keys.
- `token_exchange`: Requires an OAuth IdP; not applicable for simple API key forwarding.

## Proposed solution

Add a new outgoing auth strategy: **`header_passthrough`**.

**Configuration example:**
```yaml
outgoingAuth:
  type: header_passthrough
  headerPassthrough:
    headers:
      - x-mcp-api-key
      - x-mcp-auth
```

**Implementation touch points:**
1. `pkg/vmcp/auth/types/types.go` — add `StrategyTypeHeaderPassthrough` + `HeaderPassthroughConfig` struct
2. `pkg/vmcp/auth/strategies/header_passthrough.go` — new strategy reads specified headers from context
3. Incoming middleware / server — store raw incoming request headers in context (currently only `Identity` is stored)
4. `pkg/vmcp/client/client.go` — `identityPropagatingRoundTripper` pattern can be extended to carry selected headers

## Security considerations

- Only explicitly listed headers should be forwarded (allowlist, not passthrough-all)
- Header names should be validated (CRLF injection prevention, same as `header_injection`)
- Sensitive headers (e.g. `authorization`) should be allowed but documented as requiring care

juancarlosm:
## Summary

When a client calls vMCP with a header its backend needs — e.g. `x-mcp-api-key` that the backend resolves to a specific user — vMCP silently drops it. vMCP opens a fresh request to each backend and sends only its own headers, so the caller's header never arrives. Backends that authenticate per-user from a header therefore can't be used through vMCP today — only by hitting the backend directly.

This adds **`spec.passthroughHeaders`**: an allowlist of incoming header names that vMCP forwards, unchanged, to every backend it calls.

```yaml
spec:
  passthroughHeaders:
    - x-mcp-api-key
```

A listed header is read from the client request and attached to that session's backend requests. Headers not listed are ignored. Restricted names (`Host`, `Authorization`, `X-Forwarded-*`, hop-by-hop) are rejected.

Approach discussed and agreed with @jerm-dro on #3958 (assigned to me there).

Closes #3958

## Type of change

- [x] New feature

## Test plan

- [x] Unit tests across all layers (identity redaction, header merge, validation, capture middleware, converter precedence)
- [x] Integration test (`test/integration/vmcp/passthrough_headers_test.go`): real client → capture → backend; asserts allowlisted header forwarded, non-allowlisted dropped
- [x] `golangci-lint` clean; CRD/deepcopy/docs regenerated (no drift)

Yes:

```yaml
spec:
  passthroughHeaders: [x-litellm-api-key]
```

- Implementation: the header is captured onto the per-request `*auth.Identity` at the incoming-auth boundary (cloned, not mutated) and injected by the per-session backend client — it flows as an explicit parameter, not a context value, so no `vmcp-anti-patterns` §1/§7 coupling and the round-tripper is unchanged. This is the no-context-plumbing path discussed in #3958 once the session refactor landed; the `vmcp-review` audit was clean.
- **Limitation:** on session restore (HA), only the `(iss, sub)` tuple is persisted — forwarded headers aren't, same as bearer tokens. Low impact with `sessionAffinity: ClientIP`.

## Large PR Justification

Most of the diff is **required tests** (~700 lines: unit + an in-process e2e) and **generated code that can't be split** (CRD manifests in two API versions × two chart copies, deepcopy, API docs). Hand-written source is ~240 lines. Already trimmed from ~1157 to ~1030 added lines by de-duplicating tests (coverage unchanged); the remainder is cohesive single-feature code that wouldn't be safe to split.


pkg/auth/identity.go
```diff
@@ -63,6 +63,12 @@ type Identity struct {
 	// Redacted in MarshalJSON() to prevent token leakage.
 	// MUST NOT be mutated after the Identity is placed in the request context.
 	UpstreamTokens map[string]string
+
+	// ForwardedHeaders holds incoming request headers explicitly allowlisted for
+	// pass-through to backends. Resolved once at the auth boundary; the session
+	// is identity-bound so these are stable for the session's lifetime. Values
+	// may be sensitive (API keys) and MUST be redacted in MarshalJSON.
+	ForwardedHeaders map[string]string
```
jerm-dro: **blocker (architectural):** Carry forwarded headers on a typed context value in `pkg/vmcp/headerforward`, not on `*auth.Identity`. The prior anti-context-coupling concern from #3958 was about *business logic* depending on context-stuffed values — this is plumbing, which is exactly what context is for. Tying these to `Identity` makes `pkg/auth` carry a field that isn't an authentication concept and forces `factory.NewIncomingAuthMiddleware` to grow a parameter unrelated to its job.

Roughly:

```go
// pkg/vmcp/headerforward/capture.go
package headerforward

type forwardedHeadersKey struct{}

func WithForwardedHeaders(ctx context.Context, h map[string]string) context.Context {
    return context.WithValue(ctx, forwardedHeadersKey{}, h)
}

func ForwardedHeadersFromContext(ctx context.Context) map[string]string {
    h, _ := ctx.Value(forwardedHeadersKey{}).(map[string]string)
    return h
}

// CaptureMiddleware copies allowlisted incoming headers into ctx.
func CaptureMiddleware(allow []string) func(http.Handler) http.Handler { ... }
```

This lets `pkg/auth` and `factory.NewIncomingAuthMiddleware` revert to their pre-PR state, and `mergeForwardedHeaders` reads its input from context as a plain `map[string]string`. The CRD field, validator, converter, generated code, and e2e test are unaffected.

Happy to discuss if there's a reason Identity-as-carrier is load-bearing that I'm missing.

jerm-dro (changes requested):

pkg/auth/identity.go
juancarlosm: Makes sense — agreed. Moving the carrier to a typed context value in `pkg/vmcp/headerforward` and reverting `pkg/auth` + `NewIncomingAuthMiddleware` to their pre-PR shape. You're right that this is plumbing, not an authentication concept.

One wiring detail your sketch leaves open: where `CaptureMiddleware` gets mounted now that it's out of the auth factory. Plan is to wire it in the vMCP server's own middleware chain (`pkg/vmcp/server`) via a new `Config.PassthroughHeaders`, so the factory stays untouched. The captured map rides the same request context that already carries `Identity` down to session creation (`session_manager` → `MakeSessionWithID` → `initOneBackend`'s `WithTimeout(ctx)` → backend client), so `mergeForwardedHeaders` reads it from context with no new plumbing.

CRD field, validator, converter, generated code and the e2e assertions are unaffected; the only test touch is the integration helper swapping the factory param for a `CaptureMiddleware` wrap.

Optional, your call: instead of `mergeForwardedHeaders` reading the map from a deep context, we could read it at the session boundary (next to `IdentityFromContext`) and thread it explicitly — keeps business logic out of context entirely. Your sketch's context read is less churn and you already blessed it as plumbing, so I'll go that way unless you prefer the explicit thread.

FWIW validated end-to-end on a live cluster: with `passthroughHeaders: [x-litellm-api-key]` the backend MCP receives the header and resolves identity from it; without the patch it's dropped. Will push the refactor shortly.

pkg/vmcp/config/config.go
```diff
@@ -179,6 +179,14 @@ type Config struct {
 	// Requires Redis session storage to be configured for distributed rate limiting.
 	// +optional
 	RateLimiting *ratelimittypes.RateLimitConfig `json:"rateLimiting,omitempty" yaml:"rateLimiting,omitempty"`
+
+	// PassthroughHeaders is an allowlist of incoming client request header names
+	// forwarded verbatim to all backends. Resolved per-request at the auth
+	// boundary and baked into the per-session backend client. Names must not be
+	// in the restricted set (Host, hop-by-hop, X-Forwarded-*, etc.).
```
jerm-dro: `suggestion:` Doc comment is stale after the refactor. "Resolved per-request at the auth boundary" doesn't match the code anymore — `headerforward.CaptureMiddleware` is mounted outside the auth middleware (it runs *before* auth in execution order), and the forwarded map is consumed once at *session creation* in `createMCPClient`, not per-request. Suggested rewording:

```go
// PassthroughHeaders is an allowlist of incoming client request header names
// forwarded verbatim to all backends. Captured at the vMCP incoming edge by
// headerforward.CaptureMiddleware and consumed once at session creation
// when the per-session backend client's HeaderForwardConfig is built. Names
// must not be in the restricted set (Host, hop-by-hop, X-Forwarded-*, etc.).
```

pkg/vmcp/session/internal/backend/mcp_session.go
```diff
@@ -511,3 +518,54 @@ func initAndQueryCapabilities(
...
+//  3. Forwarded header names are canonicalized via http.CanonicalHeaderKey and
+//     checked against middleware.RestrictedHeaders; restricted names are silently
+//     dropped (defense-in-depth — they were already filtered upstream, but we
+//     guard here too).
+//  4. A forwarded header does NOT overwrite a name already present as an explicit
+//     static value in base.AddPlaintextHeaders (static config wins).
+func mergeForwardedHeaders(base *vmcp.HeaderForwardConfig, forwarded map[string]string) *vmcp.HeaderForwardConfig {
+	if len(forwarded) == 0 {
+		return base
+	}
```
jerm-dro, line 544: `suggestion:` Drop the "static config wins" precedence and fail loudly on collision instead. A header name appearing in both `cfg.PassthroughHeaders` and a backend's static `HeaderForward` (plaintext or secret) is an admin misconfiguration, not a precedence question — silently dropping one side hides the conflict, and the current wins-by-overwrite-ordering for secret-backed collisions is a latent bug waiting on someone trusting the comment.

Suggested shape: have `mergeForwardedHeaders` return `(*HeaderForwardConfig, error)` and error when a forwarded canonical name appears in `base.AddPlaintextHeaders` *or* `base.AddHeadersFromSecret`. Propagate through `createMCPClient` — the first session creation per misconfigured backend fails with a clear "forwarded header X collides with backend Y's static header-forward config" message, and the operator fixes the config.

pkg/vmcp/config/validator.go
```diff
@@ -487,6 +495,25 @@ func (*DefaultValidator) validateCompositeToolRefs(refs []CompositeToolRef) erro
...
 
+func (*DefaultValidator) validatePassthroughHeaders(cfg *Config) error {
+	for i, name := range cfg.PassthroughHeaders {
+		if name == "" {
+			return fmt.Errorf("passthroughHeaders[%d]: header name must not be empty", i)
+		}
+
+		canonical := http.CanonicalHeaderKey(name)
+
+		if middleware.RestrictedHeaders[canonical] {
```
jerm-dro, line 506: `suggestion:` Add `Authorization` to `pkg/transport/middleware/header_forward.go`'s `RestrictedHeaders` map. Backend bearer tokens have a dedicated `header_injection` outgoing-auth strategy; allowing `Authorization` in `HeaderForward` / `passthroughHeaders` is a footgun (it silently survives to backends with the `unauthenticated` strategy) with no documented use case.

test/integration/vmcp/passthrough_headers_test.go
```diff
@@ -0,0 +1,137 @@
...
+	// Assertion 2: non-allowlisted header was dropped by the capture middleware.
+	assert.NotContains(t, text, nonAllowlistedValue,
+		"non-allowlisted header %q must not reach the backend",
+		nonAllowlistedHeader)
+
+	// Confirm the non-allowlisted slot shows the absent sentinel to rule out
+	// the handler silently returning wrong data.
+	assert.Contains(t, text, absentSentinel,
+		"non-allowlisted header slot should contain the absent sentinel")
+}
```
jerm-dro, line 158: `suggestion:` Add a second `CallTool` on the same client to lock in the documented "captured once at session creation, stable for session lifetime" behavior. Right now a regression that re-reads forwarded headers on every request (or, worse, drops them after the first call) would pass this test. A second call with a *changed* `X-Test-Api-Key` value asserting the backend still sees the original would lock that contract.

pkg/vmcp/auth/factory/incoming.go
```diff
@@ -98,12 +98,12 @@ func NewIncomingAuthMiddleware(
...
-	// parsed MCP data from context.
+	// Auth middleware composes auth + MCP parser.
+	// Order (outermost → innermost): authMiddleware → parser. The parser is
+	// innermost so downstream middleware (audit, authz) sees both the identity
+	// and the parsed MCP request.
 	composedAuth := func(next http.Handler) http.Handler {
-		withParser := mcp.ParsingMiddleware(next)
-		return authMiddleware(withParser)
+		return authMiddleware(mcp.ParsingMiddleware(next))
 	}
```
jerm-dro: `nitpick:` Refactor was supposed to leave this file at its pre-PR shape, but the comment got rewritten and the two-line form collapsed. Behavior is identical; consider reverting verbatim to keep this PR's diff scoped to the passthrough feature.

jerm-dro (commented):
New direction LGTM — just a few minor notes inline.

pkg/vmcp/config/validator.go
juancarlosm: Agreed the footgun is real, but I'd push back on blocking `Authorization` outright — I have a concrete use case that depends on it. I run a vMCP fronting gmail/calendar/drive MCPs behind LiteLLM's Zero Trust MCP setup ([docs](https://docs.litellm.ai/docs/mcp_zero_trust)).

LiteLLM Zero Trust MCP works like this: the LiteLLM gateway signs a short-lived JWT on every MCP call so the downstream servers can verify the request genuinely came from LiteLLM (and which user). Concretely, its `mcp_jwt_signer` guardrail sends that JWT as `Authorization: Bearer <jwt>` (issuer=litellm, aud=mcp, 300s TTL). The flow:

```
LiteLLM (signs JWT) → vMCP (incomingAuth: anonymous — transparent aggregator)
   → passthrough authorization verbatim → backends validate the JWT (issuer/aud/JWKS)
```

vMCP isn't the enforcement point here by design — the backends are. They run the `unauthenticated` outgoing strategy because the *caller* supplies the token. Blocking `Authorization` in passthrough removes the only mechanism that makes this work, and this is in production for me, not hypothetical.

I think the real problem is the *silence*, not the header:
- authed backend → forwarded `Authorization` is silently clobbered by `authRoundTripper`
- unauthenticated backend → it silently survives (intended here, surprising elsewhere)

So rather than blocking, what if we **fail loud on an actual collision**? It lines up with your `mergeForwardedHeaders` comment: error when a passthrough header is also owned by the backend — its static `HeaderForward`, or its outgoing-auth strategy. The misconfig becomes a clear error, while the "anonymous vMCP → unauthenticated backend" passthrough stays valid since there's nothing backend-side to collide with.

Happy to keep the static-`HeaderForward` collision check in this PR and move the strategy-side check to a follow-up if that's cleaner. No objection to keeping `Host`/hop-by-hop/`X-Forwarded-*` restricted — it's specifically `Authorization` I'd like to keep allowed. What do you think?

jerm-dro: > So rather than blocking, what if we fail loud on an actual collision?

This sounds reasonable to me. Thank you for explaining the use case.

pkg/auth/identity.go
juancarlosm: Resolved — forwarded headers now ride a typed unexported context key in `pkg/vmcp/headerforward`, off `auth.Identity`. It's request-scoped transport plumbing (read only when building the outbound client, never by business logic), so it stays clear of anti-pattern #1 — and passing it explicitly would mean adding a param to the mocked `MakeSessionWithID` interface for a single feature (#8). Thanks for the push here.

juancarlosm:
Thanks for the review — pushed a round addressing all the notes:

- **Headers on `Identity` (blocker):** moved to a typed unexported context key in `pkg/vmcp/headerforward`; added an inline note on why this is request-scoped transport plumbing, not the context-coupling anti-pattern.
- **`mergeForwardedHeaders` precedence:** now fails loud — a forwarded header that collides with a backend's static `HeaderForward` (plaintext or secret) returns an error at session creation instead of silently picking a winner. Propagated through `createMCPClient`.
- **`Authorization`:** kept allowable (the LiteLLM Zero Trust case needs it) and folded into the fail-loud-on-collision approach we agreed on. The strategy-side check (passthrough vs `upstream_inject`/`token_exchange`) is tracked as a follow-up.
- **Session stability:** added a second `CallTool` with a changed caller header, asserting the backend still sees the value captured at session creation.
- **Doc + scope:** reworded the `PassthroughHeaders` doc to match the capture/consume flow; reverted the unrelated `incoming.go` auth-middleware comment to its pre-PR form.

Also merged latest `main` (resolved `serve.go`/`server.go` — `authfactory` alias + both `PassthroughHeaders` and `RateLimitMiddleware`) and fixed the resulting `Handler` gocyclo + CRD codegen regen. CI is green.

Follow-up (separate PR): fail loud when a passthrough header collides with a backend's outgoing-auth strategy that manages the same header.

juancarlosm:
@jerm-dro do you need anything else?

jerm-dro (approved):
Approving. One question before merge: is the follow-up for failing loud on passthrough-vs-**strategy** header collisions (a `token_exchange`/`header_injection` backend managing the same name, e.g. `Authorization`) filed yet? That's the other half of the original `Authorization` footgun — the static-config collision is handled, but a strategy-managed collision is still silently overwritten.

juancarlosm:
Filed it as #5535. Thanks for the thorough review on this one!

jerm-dro:
@juancarlosm Anytime! Thanks for the contribution 😃

Issue #5444, opened 2026-06-03
P3.1 deriveCoreConfig/deriveServerConfig config split


tgrunnagle:
## Description

Introduce `deriveCoreConfig` and `deriveServerConfig` to split the in-memory-only
`server.Config` into the core `Config` (collaborators + `workflowDefs` + `Authz` for the
admission seam + the cross-cutting `TelemetryProvider`/`AuditConfig`) and a transport-only
`ServerConfig`. This is the first half of RFC Phase 3 (the config decomposition); it sets up
the data shapes that #5445 then feeds into the `Serve(ctx, New(deriveCoreConfig(cfg, …)),
deriveServerConfig(cfg))` wrapper. `ServerConfig` carries **no** `AuthzMiddleware` field
(authz now lives in the core admission seam from Phase 1). The `AuthzMiddleware` field on
`server.Config` itself is **kept** (not removed in this epic): `cli/serve.go` sets it at
`serve.go:~374`, and the "cli/serve.go unchanged" constraint forbids touching that
assignment — so the field stays as a vestigial input the new path ignores.
`deriveServerConfig` simply omits it; the core admission seam derives its authorizer from
the `Authz` config instead. (The now-redundant HTTP authz/annotation **blocks** are deleted
as cleanup in #5445 once `server.New` routes through `Serve` — but the field itself
remains, so `cli/serve.go` needs no change.)

## Context

By the time this task runs, Phase 1 has produced the `VMCP` core (`New(cfg) -> VMCP` with the
admission and elicitation seams) and Phase 2 has re-homed every transport concern under
`Serve(ctx, VMCP, *ServerConfig) -> *Server` — but `server.New` does not yet call `Serve`.
This task adds the two derivation helpers that map the existing single `server.Config` onto
the core/transport pair, without changing `server.New`'s body or observable behavior yet
(that is #5445/P3.2).

The split is **not a clean partition** (Risk R3, `architecture.md` lines 483-487): the
cross-cutting fields `TelemetryProvider`, `AuditConfig`, and the health view are consumed on
**both** sides — the core decorates the backend client with telemetry and runs the workflow
auditor; the transport adds telemetry/audit middleware and owns the health monitor — so they
are passed to **both** `deriveCoreConfig` and `deriveServerConfig`. The field-by-field
placement is given in the `research.md` Config table (lines 144-167) and `architecture.md`
"ServerConfig" (lines 243-274). This is an in-memory-only change: `vmcpconfig.Config`, the
[...]

Issue #5443, opened 2026-06-03
P2.5 Move AS runner, status reporter, optimizer, health monitor under Serve


tgrunnagle:
## Description

Relocate the remaining transport lifecycle concerns under `Serve`'s `Start`/`Stop`
orchestration: the embedded authorization-server (AS) runner, the status reporter +
`periodicStatusReporting` goroutine, the optimizer cleanup, and the health monitor's
**lifecycle**. This is the final Phase 2 relocation — after it, `Serve` owns every
transport subsystem's start/stop. Crucially, `Serve` does **not construct** the
`health.Monitor`: because the Phase-3 wrapper is
`Serve(ctx, New(deriveCoreConfig(cfg)), deriveServerConfig(cfg))`, **`New` is evaluated
before `Serve`**, so the monitor must already exist when the core is constructed. The
monitor is therefore built at the **composition root**, injected into the core as a
`health.StatusProvider` (#5437), and handed to `Serve` as a built `*health.Monitor`
via `ServerConfig` so `Serve` owns only its `Start`/`Stop` lifecycle. Teardown order and
`shutdownFuncs` are preserved byte-for-byte; `server.New`'s signature and observable
behavior are unchanged.

## Context

Part of the vMCP interface refactor (epic #5419), which splits the `server.New`
god-object into a core (`New(cfg) -> VMCP`) and a transport (`Serve(ctx, VMCP,
serverCfg) -> *Server`). #5439 landed the `Serve` skeleton + `ServerConfig`;
#5440/203/204 relocated the SDK hooks, the middleware chain, and the
discovery-into-context path. This task moves the last lifecycle pieces — AS runner,
status reporter goroutine, optimizer cleanup, and the health-monitor lifecycle — into
`Serve`'s orchestration.

**Health-monitor ownership & construction order (A2).** Today a single `server.New`
builds the monitor (`server.go:436-450`) and uses it at `Handler`-time (`server.go:625-634`)
— there is no separate "core" object needing it at construction, so no ordering tension
exists. The New/Serve split introduces one: the core's `filterHealthyBackends` needs a
`health.StatusProvider` **at construction**, but `New` runs before `Serve`. Resolution
(per architecture.md R5/R7 and the resolved review decision): build the `health.Monitor`
**once at the composition root** (the `server.New` wrapper in #5445; the test harness
in standalone Phase-2 tests), inject it as a `StatusProvider` into `New` (#5437), and
[...]

Issue #5440, opened 2026-06-03
P2.2 Move SDK hooks + two-phase session creation under Serve


tgrunnagle:
## Description

Relocate the three mcp-go SDK hooks and the two-phase session-creation wiring out of
`server.New` and into the new `Serve` transport helper. The `OnRegisterSession` hook
(capability registration) and the `OnBeforeListTools`/`OnBeforeCallTool` hooks (cross-pod
Redis re-hydration) — together with the transport session manager, the session data
storage backend, and the vMCP session manager — are SDK-lifecycle concerns that belong in
the transport layer of the New/Serve split, not in the core `VMCP` domain object.

## Context

This is step P2.2 of Phase 2. #5439 introduced `Serve(ctx, VMCP, *ServerConfig) ->
*Server` plus the mcp-go server, route mux, and HTTP lifecycle skeleton, but the SDK hooks
and session creation still live in `server.New`. This task moves them under `Serve`.

mcp-go's two-phase session creation is the canonical SDK-coupling anti-pattern (#5): the
session ID is sent to the client during `initialize`, and the `OnRegisterSession` hook then
fires *after* registration to attach the per-session capability set. Because the SDK only
fires `OnRegisterSession` during `initialize`, a session re-hydrated from Redis on a
different pod has an empty per-session tool store; the `OnBeforeListTools`/
`OnBeforeCallTool` hooks lazily re-inject the tools before the SDK handler reads them. All
of this — hooks, two-phase dance, heartbeat, session-ID manager, and the Redis path — must
stay isolated inside `Serve`/the adapter and never cross the `VMCP` boundary (architecture.md
Core Principle #1; research.md lines 221-226).

The session layer's "fixed at initialize" capability set, identity binding
(`MetadataKeyIdentityBinding`), and the cross-pod Redis re-hydration path stay in
Serve/session. `server.New` remains unchanged in observable behavior throughout Phase 2.

**Parent Story**: #5431
**Dependencies**: #5439 (needs the `Serve` skeleton + `ServerConfig` to register hooks into)
**Blocks**: #5442

## Acceptance Criteria

- [ ] The three SDK hooks (`AddOnRegisterSession` -> `handleSessionRegistration`,
  `AddBeforeListTools` / `AddBeforeCallTool` -> `lazyInjectSessionTools`) are registered
  from within `Serve` (on the `server.Hooks{}` wired into the mcp-go server).
- [ ] Two-phase session creation wiring lives under `Serve`: the transport session manager
[...]

PR #5467, opened 2026-06-08
Add vMCP Serve skeleton and ServerConfig

### LINKED ISSUE #5438: P1.5 Core admission seam (bounded rewrite)

### LINKED ISSUE #5439: P2.1 Serve skeleton + ServerConfig

tgrunnagle:
## Summary

This is the transport-side entry point of the New/Serve split that the rest of
the vMCP Phase 2 refactor (epic #5419) builds on. Today `server.New` is a
god-object that constructs both the identity-parameterized core domain logic and
the HTTP/SDK transport in one call. This PR lands the additive `Serve` skeleton so
later tasks can relocate the remaining transport concerns onto it without churning
`server.New`.

- Add `Serve(ctx, core.VMCP, *ServerConfig) (*Server, error)` in the new
  `pkg/vmcp/server/serve.go`. `Serve` wraps an already-constructed core `VMCP` and
  returns the **existing** `*Server` struct, so all carried-forward accessors
  (`Handler`/`Start`/`Stop`/`Address`/`MCPServer`/`Ready`) keep working unchanged.
- Add the transport-only `ServerConfig` struct — the subset of the legacy
  `server.Config` that configures the HTTP/SDK runtime, plus the cross-cutting
  `TelemetryProvider`/`AuditConfig` consumed by both the core and the transport.
- `Serve` applies the same transport defaults `server.New` applies today
  (`Host` → `127.0.0.1`, `EndpointPath` → `/mcp`, `Name`/`Version`/`SessionTTL`
  defaults; `Port == 0` stays "OS-assigned"), builds the mcp-go server, and wires
  the core's `Close` into the server's shutdown sequence.
- Guard the `/status` handler against a nil backend registry, since the transport
  does not yet own the registry at this phase.

`server.New` is intentionally **not** routed through `Serve` in this PR, so the
change is purely additive and observable behavior is unchanged. This keeps the PR
small and the parity surface flat; Phase 3 (#5444/#5445) reduces `server.New` to a
wrapper over `Serve`.

> **Stacked PR.** This branch is stacked on #5459 (admission seam, `Closes #5438`),
> which is itself stacked on the `New`/`VMCP` work (#5437). It should target the
> parent branch `vmcp-core-p1-5_issue_5438` for review, not `main`. The
> `pkg/vmcp/core/*` changes visible in a diff against `main` are inherited from
> those upstream PRs and are not part of this change. The only files this PR
> introduces are `pkg/vmcp/server/serve.go`, `pkg/vmcp/server/serve_test.go`, and
> the `pkg/vmcp/server/status.go` nil-registry guard.

Closes #5439

## Type of change

## Test plan

Added `pkg/vmcp/server/serve_test.go` covering:
- `Serve` applies transport defaults identically to `server.New` and leaves the
  caller's `ServerConfig` unmutated and `Port == 0` OS-assigned.
- `Serve` preserves explicitly set config values.
- The returned `*Server` exposes a usable `MCPServer()`, and `Handler(ctx)`
  registers the unauthenticated routes (`/health`, `/ping`, `/readyz`, `/status`,
  `/api/backends/health`); `/metrics` is absent without a telemetry provider.
- `Stop` runs the shutdown funcs and closes the injected core.
- Validation returns a `vmcp.ErrInvalidConfig`-wrapped error for nil `cfg`, nil
  core, and nil `SessionFactory`.
- A reflection-based drift guard asserts every `Config` field is populated by
  `buildServeConfig` except the documented intentionally-unmapped set
  (`AuthzMiddleware`, `HealthMonitorConfig`).

The existing `server.New`-driven behavioral-parity suite stays green; `server.New`
is untouched and not routed through `Serve`. Repo-wide lint is clean apart from a
pre-existing, unrelated gosec G115 finding in `cmd/thv/app/upgrade.go`.

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/server/serve.go` | New `Serve` entry point and transport-only `ServerConfig`; `buildServeConfig` maps config and applies defaults |
| `pkg/vmcp/server/serve_test.go` | Unit tests for defaults, config preservation, route registration, shutdown wiring, validation, and mapping drift |
| `pkg/vmcp/server/status.go` | Nil-registry guard in `buildStatusResponse` so the Serve skeleton returns an empty backend view instead of panicking |

No.

- **Relocate, don't rewrite.** `Serve` returns the same `*Server` struct rather
  than a new type, and mirrors the defaults and mcp-go construction that
  `server.New` performs today. No mcp-go types cross the `VMCP` boundary.
- **Why `ServerConfig` exists.** These fields configure the HTTP/SDK runtime and
  are meaningless to the core (R3). The stutter `server.ServerConfig` is
  intentional and `nolint`-annotated: the split mandates the name and the package's
  existing `Config` is the legacy type being decomposed.
- **Deferred follow-ups.** SDK session hooks and two-phase creation (#5440), the
  authenticated middleware chain and direct `VMCP` request path (#5441), and the AS
  runner / status reporter / optimizer / health monitor lifecycle (#5443) are moved
  under `Serve` by later Phase 2 tasks. The `buildServeConfig` drift-guard test
  documents the two fields deliberately unmapped until then.

### DIFF
--- a/pkg/vmcp/server/serve.go
+++ b/pkg/vmcp/server/serve.go
@@ -112,16 +112,16 @@ type ServerConfig struct {
 	AuditConfig *audit.Config
 }
 
-// Serve builds the mcp-go server and HTTP transport skeleton around an
-// already-constructed core [core.VMCP], returning the existing *Server.
+// Serve is the transport-side entry point of the New/Serve split: it wraps an
+// already-constructed core [core.VMCP] and returns the existing *Server.
 //
-// This is the transport-side entry point of the New/Serve split. At this phase it
-// performs only the self-contained transport construction: applying the transport
-// defaults (mirroring server.New), building the mcp-go server, and wiring the core's
-// Close into the server's shutdown sequence. The route mux and HTTP lifecycle are
-// provided by the carried-forward (*Server).Handler/Start/Stop, which register the
-// unauthenticated routes (/health, /ping, /readyz, /status, /api/backends/health,
-// the metrics and .well-known endpoints, and any embedded auth-server routes).
+// At this phase Serve constructs only the self-contained transport pieces: it
+// applies the transport defaults (mirroring server.New), builds the mcp-go server,
+// and wires the core's Close into the server's shutdown sequence. It does NOT build
+// the route mux or the HTTP lifecycle here — those remain in the carried-forward
+// (*Server).Handler/Start/Stop, which register the unauthenticated routes (/health,
+// /ping, /readyz, /status, /api/backends/health, the metrics and .well-known
+// endpoints, and any embedded auth-server routes) when Serve's *Server is served.
 //
 // The remaining transport concerns — the SDK session hooks and two-phase session
 // creation, the authenticated middleware chain, the direct VMCP request path, and
@@ -181,6 +181,14 @@ func Serve(_ context.Context, v core.VMCP, cfg *ServerConfig) (*Server, error) {
 // defaults server.New applies today. Defaults are applied here rather than by
 // mutating the caller's ServerConfig (go-style: copy before mutating caller input).
 // Port 0 is left untouched to mean "OS-assigned".
+//
+// Two Config fields are deliberately NOT mapped at this phase (see
+// TestBuildServeConfigMapsSharedFields, which guards this list against drift):
+//   - AuthzMiddleware: the authenticated/authz middleware chain is relocated under
+//     Serve by #5441, after which authorization moves to the core admission seam.
+//   - HealthMonitorConfig: Serve receives the already-built *health.Monitor via
+//     ServerConfig.HealthMonitor (A2) and assigns it to the Server directly, so it
+//     never needs the monitor's construction config.
 func buildServeConfig(cfg *ServerConfig) *Config {
 	return &Config{
 		Name:                    cmp.Or(cfg.Name, "toolhive-vmcp"),
--- a/pkg/vmcp/server/serve.go
+++ b/pkg/vmcp/server/serve.go
@@ -133,6 +133,12 @@ type ServerConfig struct {
 // Serve returns a vmcp.ErrInvalidConfig-wrapped error for a nil cfg, a nil core, or
 // a nil required collaborator (SessionFactory). The ctx parameter is reserved for
 // the transport wiring relocated by later phases.
+//
+// Contract: the returned *Server has nil session manager, discovery manager, and
+// backend registry at this phase. It must NOT be Start()ed or served on the "/" MCP
+// route until #5440/#5441/#5442 wire those fields — the carried-forward Handler
+// passes them to WithSessionIdManager and discovery.Middleware, which would nil-deref.
+// The unauthenticated routes (registered as direct mux entries) are safe to serve.
 func Serve(_ context.Context, v core.VMCP, cfg *ServerConfig) (*Server, error) {
 	if cfg == nil {
 		return nil, fmt.Errorf("%w: nil server config", vmcp.ErrInvalidConfig)
@@ -182,21 +188,24 @@ func Serve(_ context.Context, v core.VMCP, cfg *ServerConfig) (*Server, error) {
 // mutating the caller's ServerConfig (go-style: copy before mutating caller input).
 // Port 0 is left untouched to mean "OS-assigned".
 //
-// Two Config fields are deliberately NOT mapped at this phase (see
+// Three Config fields are deliberately NOT mapped at this phase (see
 // TestBuildServeConfigMapsSharedFields, which guards this list against drift):
 //   - AuthzMiddleware: the authenticated/authz middleware chain is relocated under
 //     Serve by #5441, after which authorization moves to the core admission seam.
 //   - HealthMonitorConfig: Serve receives the already-built *health.Monitor via
 //     ServerConfig.HealthMonitor (A2) and assigns it to the Server directly, so it
 //     never needs the monitor's construction config.
+//   - StatusReporter: Serve assigns ServerConfig.StatusReporter to the Server field
+//     directly (like HealthMonitor); nothing reads Config.StatusReporter on the Serve
+//     path, so mapping it here would be dead.
 func buildServeConfig(cfg *ServerConfig) *Config {
 	return &Config{
-		Name:                    cmp.Or(cfg.Name, "toolhive-vmcp"),
-		Version:                 cmp.Or(cfg.Version, "0.1.0"),
+		Name:                    cmp.Or(cfg.Name, defaultServerName),
+		Version:                 cmp.Or(cfg.Version, defaultServerVersion),
 		GroupRef:                cfg.GroupRef,
-		Host:                    cmp.Or(cfg.Host, "127.0.0.1"),
+		Host:                    cmp.Or(cfg.Host, defaultHost),
 		Port:                    cfg.Port,
-		EndpointPath:            cmp.Or(cfg.EndpointPath, "/mcp"),
+		EndpointPath:            cmp.Or(cfg.EndpointPath, defaultEndpointPath),
 		SessionTTL:              cmp.Or(cfg.SessionTTL, defaultSessionTTL),
 		AuthMiddleware:          cfg.AuthMiddleware,
 		AuthInfoHandler:         cfg.AuthInfoHandler,
@@ -207,7 +216,6 @@ func buildServeConfig(cfg *ServerConfig) *Config {
 		Watcher:                 cfg.Watcher,
 		OptimizerFactory:        cfg.OptimizerFactory,
 		OptimizerConfig:         cfg.OptimizerConfig,
-		StatusReporter:          cfg.StatusReporter,
 		SessionFactory:          cfg.SessionFactory,
 		SessionStorage:          cfg.SessionStorage,
 	}
--- a/pkg/vmcp/server/serve.go
+++ b/pkg/vmcp/server/serve.go
@@ -1,0 +1,206 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package server
+
+import (
+	"cmp"
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/mark3labs/mcp-go/server"
+
+	"github.com/stacklok/toolhive/pkg/audit"
+	asrunner "github.com/stacklok/toolhive/pkg/authserver/runner"
+	"github.com/stacklok/toolhive/pkg/telemetry"
+	"github.com/stacklok/toolhive/pkg/vmcp"
+	vmcpconfig "github.com/stacklok/toolhive/pkg/vmcp/config"
+	"github.com/stacklok/toolhive/pkg/vmcp/core"
+	"github.com/stacklok/toolhive/pkg/vmcp/health"
+	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
+	vmcpsession "github.com/stacklok/toolhive/pkg/vmcp/session"
+	vmcpstatus "github.com/stacklok/toolhive/pkg/vmcp/status"
+)
+
+// ServerConfig holds the transport-only runtime configuration for Serve.
+//
+// It carries the subset of the legacy server.Config that configures the HTTP/SDK
+// runtime, plus the cross-cutting TelemetryProvider/AuditConfig that are consumed
+// by both the core (core.New) and the transport (Serve) — not a clean partition.
+// The fields the core owns (Aggregator, Router, BackendRegistry, BackendClient,
+// WorkflowDefs, Authz) live on core.Config instead and are absent here.
+//
+// The name intentionally stutters as server.ServerConfig: it is mandated by the
+// New/Serve split, and the package's existing Config is the legacy transport
+// config being decomposed, so it cannot be reused for this transport-only type.
+//
+//nolint:revive // exported-stutter is intentional; see the doc comment above.
+type ServerConfig struct {
+	// Name and Version are exposed in the MCP protocol via the mcp-go server.
+	Name    string
+	Version string
+
+	// GroupRef is the name of the MCPGroup containing backend workloads, used for
+	// operational visibility in the status endpoint and logging.
+	GroupRef string
+
+	// Host is the bind address (default: "127.0.0.1").
+	Host string
+
+	// Port is the bind port. A zero value means "let the OS assign a random port".
+	Port int
+
+	// EndpointPath is the MCP endpoint path (default: "/mcp").
+	EndpointPath string
+
+	// SessionTTL is the session time-to-live duration (default: 30 minutes).
+	SessionTTL time.Duration
+
+	// AuthMiddleware is the optional authentication middleware applied to MCP routes.
+	// If nil, no authentication is required.
+	AuthMiddleware func(http.Handler) http.Handler
+
+	// AuthInfoHandler is the optional handler for the
+	// /.well-known/oauth-protected-resource endpoint.
+	AuthInfoHandler http.Handler
+
+	// AuthServer is the optional embedded authorization server. When non-nil, its
+	// routes are registered on the mux alongside the protected resource metadata.
+	AuthServer *asrunner.EmbeddedAuthServer
+
+	// HealthMonitor is the already-built backend health monitor (constructed at the
+	// composition root so the same instance's StatusProvider can be injected into the
+	// core). Serve owns only its Start/Stop lifecycle. Nil disables health monitoring.
+	HealthMonitor *health.Monitor
+
+	// StatusReportingInterval is the interval for reporting status updates.
+	// If zero, the default reporting interval is used.
+	StatusReportingInterval time.Duration
+
+	// StatusReporter enables the vMCP runtime to report operational status.
+	// If nil, status reporting is disabled.
+	StatusReporter vmcpstatus.Reporter
+
+	// Watcher is the optional Kubernetes backend watcher for dynamic mode, used by
+	// the /readyz endpoint to gate readiness on cache sync.
+	Watcher Watcher
+
+	// SessionFactory creates MultiSessions for session management. Required; Serve
+	// returns an error when it is nil.
+	SessionFactory vmcpsession.MultiSessionFactory
+
+	// SessionStorage configures the session storage backend. When nil or provider is
+	// "memory", local in-process storage is used.
+	SessionStorage *vmcpconfig.SessionStorageConfig
+
+	// OptimizerFactory builds an optimizer from a list of tools. If nil, the
+	// optimizer is disabled.
+	OptimizerFactory func(context.Context, []server.ServerTool) (optimizer.Optimizer, error)
+
+	// OptimizerConfig holds the parsed optimizer search parameters. A nil value
+	// disables the optimizer.
+	OptimizerConfig *optimizer.Config
+
+	// TelemetryProvider is the cross-cutting telemetry provider (also consumed by
+	// core.New). If nil, no telemetry is recorded.
+	TelemetryProvider *telemetry.Provider
+
+	// AuditConfig is the cross-cutting audit configuration (also consumed by
+	// core.New). If nil, no audit logging is performed.
+	AuditConfig *audit.Config
+}
+
+// Serve builds the mcp-go server and HTTP transport skeleton around an
+// already-constructed core [core.VMCP], returning the existing *Server.
+//
+// This is the transport-side entry point of the New/Serve split. At this phase it
+// performs only the self-contained transport construction: applying the transport
+// defaults (mirroring server.New), building the mcp-go server, and wiring the core's
+// Close into the server's shutdown sequence. The route mux and HTTP lifecycle are
+// provided by the carried-forward (*Server).Handler/Start/Stop, which register the
+// unauthenticated routes (/health, /ping, /readyz, /status, /api/backends/health,
+// the metrics and .well-known endpoints, and any embedded auth-server routes).
+//
+// The remaining transport concerns — the SDK session hooks and two-phase session
+// creation, the authenticated middleware chain, the direct VMCP request path, and
+// the AS runner / status reporter / optimizer / health monitor lifecycle — are
+// relocated under Serve by the subsequent Phase 2 tasks. server.New is not yet
+// routed through Serve, so this is purely additive and observable behavior is
+// unchanged.
+//
+// Serve returns a vmcp.ErrInvalidConfig-wrapped error for a nil cfg, a nil core, or
+// a nil required collaborator (SessionFactory). The ctx parameter is reserved for
+// the transport wiring relocated by later phases.
+func Serve(_ context.Context, v core.VMCP, cfg *ServerConfig) (*Server, error) {
+	if cfg == nil {
+		return nil, fmt.Errorf("%w: nil server config", vmcp.ErrInvalidConfig)
+	}
+	if v == nil {
+		return nil, fmt.Errorf("%w: VMCP is required", vmcp.ErrInvalidConfig)
+	}
+	if cfg.SessionFactory == nil {
+		return nil, fmt.Errorf("%w: SessionFactory is required", vmcp.ErrInvalidConfig)
+	}
+
+	serveCfg := buildServeConfig(cfg)
+
+	// Build the mcp-go server (mirrors server.New): tools and resources are
+	// registered dynamically, so capabilities start disabled. Hooks are empty for
+	// now; the session hooks are relocated here by a later Phase 2 task.
+	mcpServer := server.NewMCPServer(
+		serveCfg.Name,
+		serveCfg.Version,
+		server.WithToolCapabilities(false),
+		server.WithResourceCapabilities(false, false),
+		server.WithLogging(),
+		server.WithHooks(&server.Hooks{}),
+	)
+
+	srv := &Server{
+		config:         serveCfg,
+		mcpServer:      mcpServer,
+		ready:          make(chan struct{}),
+		healthMonitor:  cfg.HealthMonitor,
+		statusReporter: cfg.StatusReporter,
+	}
+
+	// Serve owns the injected core's lifecycle: release its backend connections
+	// when the server stops. core.VMCP.Close is idempotent, so this is safe even
+	// if Stop runs before Start.
+	srv.shutdownFuncs = append(srv.shutdownFuncs, func(context.Context) error {
+		return v.Close()
+	})
+
+	return srv, nil
+}
+
+// buildServeConfig maps the transport-only ServerConfig onto the existing *Config
+// the carried-forward (*Server) methods read from, applying the same transport
+// defaults server.New applies today. Defaults are applied here rather than by
+// mutating the caller's ServerConfig (go-style: copy before mutating caller input).
+// Port 0 is left untouched to mean "OS-assigned".
+func buildServeConfig(cfg *ServerConfig) *Config {
+	return &Config{
+		Name:                    cmp.Or(cfg.Name, "toolhive-vmcp"),
+		Version:                 cmp.Or(cfg.Version, "0.1.0"),
+		GroupRef:                cfg.GroupRef,
+		Host:                    cmp.Or(cfg.Host, "127.0.0.1"),
+		Port:                    cfg.Port,
+		EndpointPath:            cmp.Or(cfg.EndpointPath, "/mcp"),
+		SessionTTL:              cmp.Or(cfg.SessionTTL, defaultSessionTTL),
+		AuthMiddleware:          cfg.AuthMiddleware,
+		AuthInfoHandler:         cfg.AuthInfoHandler,
+		AuthServer:              cfg.AuthServer,
+		TelemetryProvider:       cfg.TelemetryProvider,
+		AuditConfig:             cfg.AuditConfig,
+		StatusReportingInterval: cfg.StatusReportingInterval,
+		Watcher:                 cfg.Watcher,
+		OptimizerFactory:        cfg.OptimizerFactory,
+		OptimizerConfig:         cfg.OptimizerConfig,
+		StatusReporter:          cfg.StatusReporter,
+		SessionFactory:          cfg.SessionFactory,
+		SessionStorage:          cfg.SessionStorage,
+	}
+}
--- a/pkg/vmcp/server/serve_test.go
+++ b/pkg/vmcp/server/serve_test.go
@@ -7,14 +7,23 @@ import (
 	"context"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 	"testing"
+	"time"
 
+	"github.com/mark3labs/mcp-go/server"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/stacklok/toolhive/pkg/audit"
 	"github.com/stacklok/toolhive/pkg/auth"
+	asrunner "github.com/stacklok/toolhive/pkg/authserver/runner"
+	"github.com/stacklok/toolhive/pkg/telemetry"
 	"github.com/stacklok/toolhive/pkg/vmcp"
+	vmcpconfig "github.com/stacklok/toolhive/pkg/vmcp/config"
 	"github.com/stacklok/toolhive/pkg/vmcp/core"
+	"github.com/stacklok/toolhive/pkg/vmcp/health"
+	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
 )
 
 // stubVMCP is a no-op core.VMCP for Serve tests. The Serve skeleton never invokes
@@ -54,6 +63,20 @@ func (*stubVMCP) LookupPrompt(context.Context, *auth.Identity, string) (*vmcp.Pr
 	return nil, nil
 }
 func (s *stubVMCP) Close() error { s.closed = true; return nil }
+
+// stubWatcher is a non-nil Watcher for the drift-guard test; its behavior is not
+// exercised there.
+type stubWatcher struct{}
+
+func (stubWatcher) WaitForCacheSync(context.Context) bool { return true }
+
+// stubServeReporter is a non-nil vmcpstatus.Reporter for the drift-guard test.
+type stubServeReporter struct{}
+
+func (stubServeReporter) ReportStatus(context.Context, *vmcp.Status) error { return nil }
+func (stubServeReporter) Start(context.Context) (func(context.Context) error, error) {
+	return func(context.Context) error { return nil }, nil
+}
 
 func TestServeAppliesTransportDefaults(t *testing.T) {
 	t.Parallel()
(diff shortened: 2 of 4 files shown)


pkg/vmcp/server/serve.go
```diff
@@ -0,0 +1,214 @@
...
+// Two Config fields are deliberately NOT mapped at this phase (see
+// TestBuildServeConfigMapsSharedFields, which guards this list against drift):
+//   - AuthzMiddleware: the authenticated/authz middleware chain is relocated under
+//     Serve by #5441, after which authorization moves to the core admission seam.
+//   - HealthMonitorConfig: Serve receives the already-built *health.Monitor via
+//     ServerConfig.HealthMonitor (A2) and assigns it to the Server directly, so it
+//     never needs the monitor's construction config.
+func buildServeConfig(cfg *ServerConfig) *Config {
+	return &Config{
+		Name:                    cmp.Or(cfg.Name, "toolhive-vmcp"),
```
tgrunnagle: **[MEDIUM] Default literals are now triplicated.** `"toolhive-vmcp"`, `"0.1.0"`, `"127.0.0.1"`, `"/mcp"` are hard-coded here, in `New` (`server.go:312-328`), and asserted as literals in `TestServeAppliesTransportDefaults` — three independent copies that can drift (e.g. `New` bumps `"0.1.0"`→`"0.2.0"` and `Serve` silently disagrees until Phase 3 unifies them). Only `SessionTTL` uses a shared const. Suggest extracting `defaultServerName`/`defaultServerVersion`/`defaultHost`/`defaultEndpointPath` package consts referenced by both `New` and `buildServeConfig`.

pkg/vmcp/server/serve.go
```diff
@@ -0,0 +1,214 @@
...
+		AuthMiddleware:          cfg.AuthMiddleware,
+		AuthInfoHandler:         cfg.AuthInfoHandler,
+		AuthServer:              cfg.AuthServer,
+		TelemetryProvider:       cfg.TelemetryProvider,
+		AuditConfig:             cfg.AuditConfig,
+		StatusReportingInterval: cfg.StatusReportingInterval,
+		Watcher:                 cfg.Watcher,
+		OptimizerFactory:        cfg.OptimizerFactory,
+		OptimizerConfig:         cfg.OptimizerConfig,
+		StatusReporter:          cfg.StatusReporter,
```
tgrunnagle: **[MEDIUM] Dead mapping on the Serve path.** `Serve` sets `srv.statusReporter` directly (line 172) and `Start` reads that struct field, never `config.StatusReporter` — so this assignment writes a `Config` field nothing on the `Serve` path reads. It exists only because the drift guard requires every non-exempt `Config` field be non-zero. This is also inconsistent with `HealthMonitor`, which is set directly and *not* mapped. Suggest either adding `StatusReporter` to `intentionallyUnmapped` (reason: "set directly on Server; `Config.StatusReporter` only read by `New`") and dropping this line, or a comment that it's carried for `New`-path / Phase-3 parity and is currently dead here.

pkg/vmcp/server/serve.go
```diff
@@ -0,0 +1,214 @@
...
+	mcpServer := server.NewMCPServer(
+		serveCfg.Name,
+		serveCfg.Version,
+		server.WithToolCapabilities(false),
+		server.WithResourceCapabilities(false, false),
+		server.WithLogging(),
+		server.WithHooks(&server.Hooks{}),
+	)
+
+	srv := &Server{
```
tgrunnagle, line 167: **[LOW-MEDIUM] Document the nil-field contract.** This `*Server` is left with nil `vmcpSessionMgr`/`discoveryMgr`/`backendRegistry`. The carried-forward `Handler` passes these to `WithSessionIdManager` and `discovery.Middleware` for the `/` route, so `Start`-ing or serving `/` on a `Serve`-built server would nil-deref. Only `/status` was hardened (status.go). Not reachable today (`New` isn't routed through `Serve`), but the trap is silent — suggest a doc-comment line stating the returned `*Server` must not be `Start`ed/served on `/` until #5440/#5441/#5442 wire those fields, or a fail-loud nil check in `Start`.

pkg/vmcp/server/serve_test.go
```diff
@@ -0,0 +1,259 @@
...
+	cfg := &ServerConfig{SessionFactory: testMinimalFactory()}
+	srv, err := Serve(context.Background(), &stubVMCP{}, cfg)
+	require.NoError(t, err)
+
+	handler, err := srv.Handler(context.Background())
+	require.NoError(t, err)
+
+	// The unauthenticated routes are direct mux entries and respond without the
+	// not-yet-relocated authenticated MCP chain or its dependencies.
+	for _, path := range []string{"/health", "/ping", "/readyz", "/status", "/api/backends/health"} {
```
tgrunnagle, line 152: **[MEDIUM] Test/comment gaps vs. the acceptance criteria.** (1) `.well-known/` is listed among the required unauthenticated routes in the issue's Testing Strategy but isn't probed here — add a `/.well-known/oauth-protected-resource` request asserting the clean 404 JSON (not the catch-all's 406). (2) The comment at lines 367-368 ("respond without the not-yet-relocated authenticated MCP chain") overstates reality: `Handler` still builds and mounts the full middleware chain at `/` in this PR; these health routes merely bypass it as direct mux entries. Suggest softening the comment to match, or adding an explicit assertion of the intended "no auth gate" edge case.

pkg/vmcp/server/serve_test.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+}
+
+// TestBuildServeConfigMapsSharedFields guards the ServerConfig -> Config mapping
+// against silent drift: every Config field except the documented intentionally-
+// unmapped set must be populated by buildServeConfig. If a future field is added to
+// Config and forgotten in buildServeConfig, this test fails (the field is zero).
+// When a field is deliberately not mapped, add it to intentionallyUnmapped with a
+// reason, mirroring the buildServeConfig doc comment.
+func TestBuildServeConfigMapsSharedFields(t *testing.T) {
```
tgrunnagle, line 271: **[MEDIUM, part of the parity-drift finding] Drift guard checks presence, not correctness.** Because it only asserts `!IsZero()` with every source field set non-zero, it can't catch (a) a wrong-source mapping (`Host: cfg.GroupRef` still passes), (b) default-value drift (the `cmp.Or` defaults are never exercised since all sources are non-zero), or (c) the inverse direction (a `ServerConfig` field that's never threaded into `Config`). Suggest per-field sentinel-value equality assertions for the simple pass-through fields, and a note in the comment that this guards presence while `TestServeAppliesTransportDefaults`/`TestServePreservesExplicitConfig` carry the value correctness.

pkg/vmcp/server/status.go
```diff
@@ -60,8 +60,14 @@ func (s *Server) handleStatus(w http.ResponseWriter, r *http.Request) {
...
 // Uses the provided context for request cancellation and tracing propagation.
 func (s *Server) buildStatusResponse(ctx context.Context) StatusResponse {
-	// Get current backends from registry (supports dynamic backend changes)
-	backends := s.backendRegistry.List(ctx)
+	// Get current backends from registry (supports dynamic backend changes).
+	// The registry is owned by the core and may not be wired into the transport
+	// yet (the Serve skeleton, before later Phase 2 tasks relocate it); a nil
+	// registry yields an empty backend view rather than a panic.
+	var backends []vmcp.Backend
+	if s.backendRegistry != nil {
```
tgrunnagle, line 71: **[LOW] Guard could mask a real misconfig on the `New` path.** Correct and necessary for the skeleton's unauthenticated `/status`. But on the `server.New` path `backendRegistry` is a required constructor arg and should never be nil — here a nil would silently degrade to an empty backend view (`Healthy:false`, HTTP 200) and hide a wiring bug. Suggest a `slog.Debug`/`Warn` in the nil branch so an unexpectedly-nil registry post-#5443 is observable rather than silent. Non-blocking.

tgrunnagle (commented):
## Multi-agent review — Serve skeleton + ServerConfig

Sound, behavior-preserving additive skeleton. The deferral of the route mux/lifecycle to the carried-forward `Handler`/`Start`/`Stop` (rather than rebuilding it in `Serve`) is the **correct** design call and faithful to the epic's "relocate, don't rewrite" principle — not a scope gap. Tests pass; clean boundary (no mcp-go types cross `core.VMCP`); `buildServeConfig` copies-before-mutating (better than `New`).

**No HIGH findings — nothing blocks merge.** The MEDIUM items are about tightening the *parity-drift protection* this staged refactor leans on, plus a couple of test/comment-precision gaps.

| Severity | Finding |
|----------|---------|
| MEDIUM | Default literals triplicated (`New` / `buildServeConfig` / test) + drift guard checks presence, not value/inverse |
| MEDIUM | Dead `Config.StatusReporter` mapping — set directly on the struct, mapped only to satisfy the drift test |
| MEDIUM | `.well-known/` route in the AC but untested; "auth-chain absent" comment overstates reality (`Handler` still builds the chain) |
| LOW-MED | `Serve`-built `*Server` is not startable (nil session/discovery/registry); document the nil-field contract |
| LOW | nil-registry guard may silently mask a real misconfig on the `server.New` path |

Minor (non-blocking, no inline comment): no positive `/metrics`-registered assertion; `Close` idempotency and a combined nil-cfg+nil-vmcp ordering case aren't asserted.

_Reviewed by 4 specialist agents (correctness, architecture, test, security). Codex cross-review skipped (CLI not installed)._

pkg/vmcp/server/serve.go
tgrunnagle: Addressed in 3d5e3316. Extracted `defaultServerName`/`defaultServerVersion`/`defaultHost`/`defaultEndpointPath` package consts in server.go, now referenced by both `New` (the defaulting block) and `buildServeConfig`, and asserted against in `TestServeAppliesTransportDefaults` — no more triplicated literals.

pkg/vmcp/server/serve.go
tgrunnagle: Addressed in 3d5e3316. Went with the drop-and-exempt option: removed the dead `Config.StatusReporter` line from `buildServeConfig` (Serve sets `srv.statusReporter` directly, like `HealthMonitor`; nothing on the Serve path reads `Config.StatusReporter`) and added `StatusReporter` to the drift guard's `intentionallyUnmapped` set with that reason. Now consistent with `HealthMonitor`.

pkg/vmcp/server/serve.go
tgrunnagle: Addressed in 3d5e3316. Added a doc-comment contract to `Serve` stating the returned `*Server` has nil session/discovery/registry and must not be `Start()`ed or served on `/` until #5440/#5441/#5442 wire those fields. Went with the doc approach rather than a nil-check in `Start` since AC #3 keeps `Start` unchanged on the `New` path.

pkg/vmcp/server/serve_test.go
tgrunnagle: Addressed in 3d5e3316. (1) Added a `/.well-known/oauth-protected-resource` request asserting the clean JSON 404 (status + `application/json`), distinct from the catch-all's 406. (2) Softened the comment to say these routes are direct mux entries that bypass the chain `Handler` still builds and mounts at `/`.

pkg/vmcp/server/serve_test.go
tgrunnagle: Addressed in 3d5e3316. Added a note to the drift guard's doc comment that it is a presence-only check, and that value-correctness lives in `TestServePreservesExplicitConfig` (pass-through scalars) and `TestServeAppliesTransportDefaults` (defaults). Extended `TestServePreservesExplicitConfig` with distinct `SessionTTL`/`StatusReportingInterval` values so wrong-source mapping of those scalars is caught.

pkg/vmcp/server/status.go
tgrunnagle: Addressed in 3d5e3316. Added `slog.Debug` in the nil-registry branch noting the registry isn't wired. Used Debug rather than Warn: nil is expected on the Serve skeleton path and `/status` may be polled, so Warn would spam — Debug keeps it observable when investigating a `New`-path wiring bug without routine noise.

tgrunnagle:
Re: the non-blocking minor items in the review body (addressed in 3d5e3316):

- **Positive `/metrics` assertion** — added `TestServeHandlerRegistersMetricsWhenTelemetryEnabled`, which builds a real `telemetry.Provider` with `EnablePrometheusMetricsPath: true` and asserts `/metrics` returns 200.
- **Combined nil-cfg + nil-vmcp ordering** — added a `nil config and nil vmcp` case to `TestServeValidation`, confirming it fails cleanly (cfg is checked first; no nil-deref).
- **`Close` idempotency** — left as-is: idempotency is a contract of real `core.VMCP` implementations, not of `Serve`'s wiring. `Serve` appends exactly one `Close` to `shutdownFuncs`, which `TestServeStopClosesCore` already verifies. Happy to add a count-based assertion if you'd prefer it pinned.

jerm-dro (approved):
Approving to unblock the stack; will look harder once `New` is actually routed through `Serve` and `v` is load-bearing.

PR #5469, opened 2026-06-09
Renew public client TTL on successful token exchange

mani-muon:
## Summary

Public clients registered through Dynamic Client Registration (DCR) get `DefaultPublicClientTTL` (30 days) stamped on their Redis registration at registration time, and nothing refreshes it. An actively-used public client — e.g. a long-lived MCP desktop/IDE client that caches its `client_id` indefinitely — is therefore evicted 30 days after it registered, regardless of use. Its next token request then fails at the token endpoint with `invalid_client`, and the only recovery is to fully re-register the client (which silently recurs ~monthly).

- Add `RenewClientTTL` to the `ClientRegistry` storage interface. The Redis backend `EXPIRE`s the public client's key back to `DefaultPublicClientTTL`; the in-memory backend is a documented no-op (its clients have no TTL).
- Call it best-effort from the token endpoint after a successful exchange/refresh, so registration survival tracks *proven use*. Confidential clients (which have no TTL) and already-evicted keys are left untouched.
- Renewal deliberately runs on the token path, **not** on the client read. `GetClient` is reached from the unauthenticated front-channel `GET /oauth/authorize` before any auth, and public clients have no secret, so renewing there would let any caller who knows a public `client_id` keep its registration alive indefinitely and defeat the anti-bloat TTL.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Dependency update
- [ ] Documentation
- [ ] Other (describe):

## Test plan

Ran the `pkg/authserver/storage` and `pkg/authserver/server/handlers` unit tests with `-race` (the Taskfile's `go test` invocation, scoped to those packages) — all pass. The storage test exercises `RenewClientTTL` directly (public renews, confidential stays TTL-less, unknown key is a safe no-op); a handler test asserts a successful token exchange renews the public client's TTL, and I verified it fails without the wiring. Ran `go vet` and `go build ./...`. golangci-lint was not run locally; leaving the full lint to CI.

Yes. Long-lived public OAuth/MCP clients (e.g. Claude Desktop, Claude Code, Cursor, VS Code) backed by the Redis storage no longer stop working ~30 days after they were first authorized; an in-use client's registration is refreshed on each successful token exchange. No configuration change is required. The in-memory backend is unaffected.

- Renewal lives on the token-endpoint success path rather than in `GetClient` specifically so an unauthenticated `/oauth/authorize` probe cannot keep a public client alive — raised in review on the first revision.
- The in-memory backend stores clients in a bare map with no TTL, so public clients there never expire; `RenewClientTTL` is a no-op for it. The two backends already diverge on this and this PR does not change in-memory behavior.

### DIFF
--- a/pkg/authserver/storage/redis.go
+++ b/pkg/authserver/storage/redis.go
@@ -181,7 +181,8 @@ func (s *RedisStorage) RegisterClient(ctx context.Context, client fosite.Client)
 		return fmt.Errorf("failed to marshal client: %w", err)
 	}
 
-	// Public clients (from DCR) expire to prevent unbounded growth.
+	// Public clients (from DCR) expire to prevent unbounded growth; GetClient
+	// renews this TTL on use so actively-used clients are not evicted.
 	// Confidential clients don't expire.
 	ttl := time.Duration(0)
 	if client.IsPublic() {
@@ -192,6 +193,16 @@ func (s *RedisStorage) RegisterClient(ctx context.Context, client fosite.Client)
 }
 
 // GetClient loads the client by its ID.
+//
+// For public (DCR) clients, a successful load renews the registration TTL to
+// DefaultPublicClientTTL, making expiry a sliding window of inactivity rather than
+// a fixed window from registration time. Without this, an actively-used public
+// client is evicted DefaultPublicClientTTL after it registered regardless of use,
+// which surfaces to the client as invalid_client on its next token request and
+// forces a full re-registration. Abandoned clients still expire after the
+// inactivity window, preserving the anti-bloat intent of RegisterClient.
+// Confidential clients are stored without a TTL and are left untouched. TTL
+// renewal is best-effort: a failure is logged and does not fail the lookup.
 func (s *RedisStorage) GetClient(ctx context.Context, id string) (fosite.Client, error) {
 	key := redisKey(s.keyPrefix, KeyTypeClient, id)
 
@@ -206,6 +217,15 @@ func (s *RedisStorage) GetClient(ctx context.Context, id string) (fosite.Client,
 	var stored storedClient
 	if err := json.Unmarshal(data, &stored); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal client: %w", err)
+	}
+
+	// Renew the registration TTL on use so actively-used public (DCR) clients are
+	// not evicted mid-lifecycle. Mirrors the TTL set in RegisterClient; confidential
+	// clients have no TTL and are left untouched.
+	if stored.Public {
+		if err := s.client.Expire(ctx, key, DefaultPublicClientTTL).Err(); err != nil {
+			slog.Warn("failed to renew public client TTL", "client_id", id, "error", err)
+		}
 	}
 
 	return &redisClient{storedClient: stored}, nil
--- a/pkg/authserver/storage/redis.go
+++ b/pkg/authserver/storage/redis.go
@@ -181,8 +181,8 @@ func (s *RedisStorage) RegisterClient(ctx context.Context, client fosite.Client)
 		return fmt.Errorf("failed to marshal client: %w", err)
 	}
 
-	// Public clients (from DCR) expire to prevent unbounded growth; GetClient
-	// renews this TTL on use so actively-used clients are not evicted.
+	// Public clients (from DCR) expire to prevent unbounded growth; RenewClientTTL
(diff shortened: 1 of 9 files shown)


ChrisJBurns (commented):
## Review summary — Renew public client TTL on read

Small, well-scoped bug fix: `GetClient` now renews `DefaultPublicClientTTL` for public (DCR) clients on read, turning a fixed 30-day window into a sliding inactivity window so actively-used MCP clients aren't evicted and forced to re-register. The fix is **correct and well-implemented** — renewal is gated on `stored.Public`, best-effort (WARN + continue), and the doc comments were updated in lockstep. The new test's positive assertions are load-bearing and the confidential negative case is correct.

**No blocking issues** (0 HIGH). One MEDIUM design trade-off is worth the author's consideration:

| # | Severity | Finding |
|---|----------|---------|
| F1 | MEDIUM | Renewal fires from the **unauthenticated** authorize path, so anyone who knows a public `client_id` can keep its row alive indefinitely — softening the anti-bloat guarantee the TTL was added for. |

*Verified: `handlers/authorize.go` → `NewAuthorizeRequest` reaches `GetClient` with the caller-supplied `client_id` before any client authentication; public clients have no secret.*

Reviewed by 3 specialist agents (Redis/correctness, security/OAuth, test-coverage). 🤖 Generated with [Claude Code](https://claude.com/claude-code)

pkg/authserver/storage/redis.go
```diff
@@ -208,6 +219,15 @@ func (s *RedisStorage) GetClient(ctx context.Context, id string) (fosite.Client,
 		return nil, fmt.Errorf("failed to unmarshal client: %w", err)
 	}
 
+	// Renew the registration TTL on use so actively-used public (DCR) clients are
+	// not evicted mid-lifecycle. Mirrors the TTL set in RegisterClient; confidential
+	// clients have no TTL and are left untouched.
+	if stored.Public {
+		if err := s.client.Expire(ctx, key, DefaultPublicClientTTL).Err(); err != nil {
```
ChrisJBurns: **[MEDIUM · Security] Renewal fires on an unauthenticated read, weakening the anti-bloat guarantee.**

`GetClient` is invoked by fosite's `NewAuthorizeRequest` from the unauthenticated `GET /oauth/authorize` handler with the raw caller-supplied `client_id`, **before** any PKCE/secret/user auth. Public clients have no secret, so a successful `GetClient` requires only a known (or guessed) `client_id`. This `EXPIRE` therefore turns every such read into a full 30-day reset, letting any unauthenticated caller pin an arbitrary set of public-client rows alive forever simply by polling more often than every 30 days.

The guarantee degrades from "abandoned ⇒ evicted in ≤30 days" to "evicted only if nobody — legitimate or not — ever looks the id up again." Note the renewal conflates *"someone looked this client up"* with *"this client is actively in use"*: a scanner, health-checker, or stale cached `client_id` hitting `/oauth/authorize` without ever completing a flow now keeps the row alive indefinitely. This also makes the added doc line "Abandoned clients still expire after the inactivity window" overstate the actual guarantee.

Not an active compromise — DCR registration (the growth *vector*) is unchanged, and `client_id`s are random/unguessable — but it does undercut the resource-exhaustion defense this very TTL exists to provide.

**Mitigation options (most robust first):**
1. **Renew on proven use, not on read** — move the `Expire` to the token-endpoint success path (after the code is exchanged / PKCE validates), so a front-channel probe no longer counts as "active." Best matches the PR's stated intent.
2. **Absolute max lifetime** — keep read-renewal but store `CreatedAt` and refuse to extend past `CreatedAt + maxLifetime`, bounding the worst case regardless of probing.
3. **Accept and document** — if indefinite survival of recently-read public clients is acceptable, state it explicitly and fix the "abandoned clients still expire" doc line so it doesn't claim a guarantee the code no longer provides.

mani-muon:
Good catch on F1 — you're right, `GetClient` runs on the unauthenticated `/oauth/authorize` path, so renewing there let anyone who knows a public `client_id` keep it alive forever.

Switched to renewing on proven use instead: a new `RenewClientTTL` that the token handler calls after a successful exchange/refresh, so a front-channel probe no longer counts. `GetClient` is a plain read again, and the docs/tests were updated to match.

Addressed in 2c3676f9.

ChrisJBurns (dismissed):

reyortiz3 (approved):

ChrisJBurns (approved):

PR #5471, opened 2026-06-09
Move SDK hooks and two-phase session creation under Serve

### LINKED ISSUE #5440: P2.2 Move SDK hooks + two-phase session creation under Serve

tgrunnagle:
## Summary

This is step **P2.2** of the vMCP New/Serve split (Phase 2). The mcp-go SDK hooks and the
two-phase session-creation wiring are SDK-lifecycle concerns that belong in the transport
layer of the New/Serve split, not in the core `VMCP` domain object — but they still lived in
`server.New`. This PR relocates them under `Serve` so they execute identically when `Serve` is
exercised directly, keeping the mcp-go two-phase dance (anti-pattern #5) isolated behind the
transport boundary and out of the core interface.

- **Why**: Finish moving the SDK session layer behind `Serve` so the core `VMCP` object stays
  free of mcp-go session concepts (hooks, two-phase creation, cross-pod Redis re-hydration).
- **What**: Move the three SDK hooks (`OnRegisterSession`, `OnBeforeListTools`,
  `OnBeforeCallTool`) and the session-creation wiring (transport session manager, session data
  storage, vMCP session manager) from `server.New` into `Serve` in `pkg/vmcp/server/serve.go`.
- **Scope**: Purely additive. `server.New` keeps its own copy until Phase 3, so its signature
  and observable behavior are **unchanged**; the relocated callbacks and collaborators keep
  their existing shapes.

Closes #5440

## Type of change

## Test plan

- [x] Unit tests (`task test`) — `pkg/vmcp/server` passes with `-race`
- [x] Linting — `golangci-lint run ./pkg/vmcp/server/...` reports 0 issues

New `Serve`-level tests drive the relocated hooks and two-phase session creation through the
SDK lifecycle directly (mounting the Streamable HTTP server on the relocated `mcpServer` +
`vmcpSessionMgr`, bypassing the not-yet-relocated discovery middleware that #5441/#5442 own):

- `TestServeRegistersSessionHooks` — an MCP `initialize` fires `OnRegisterSession`, which runs
  two-phase creation (`MakeSessionWithID`) and injects the per-session tools so a subsequent
  `tools/list` advertises them; the `OnBeforeListTools` hook runs and no-ops on the pod-local
  path.
- `TestServeLazyInjectsToolsForRehydratedSession` — covers the cross-pod re-injection branch:
  with a session registered in the vMCP session manager, a fresh SDK `ClientSession` whose
  per-session tool store is empty (as on a second pod, where `OnRegisterSession` never fired)
  gets its tools re-injected by `lazyInjectSessionTools`; a populated store is left untouched.
- `TestServeReturnsErrorWhenSessionManagerConstructionFails` — asserts Serve surfaces a
  `sessionmanager.New` failure that occurs after the session data storage is built (the path
  the `closeStorageOnErr` guard protects). It confirms the guarded path is reached; it does not
  directly observe `Close()` (the storage is built internally).
- `TestBuildSessionDataStorage` / `TestBuildSessionDataStorageRedis` — provider selection:
  nil/empty/`memory` (case-insensitive) yields in-process storage; `redis` takes the Redis path
  reading `THV_SESSION_REDIS_PASSWORD`; unknown providers are rejected.

The ~1.2k-line HTTP session-management integration suite stays on `server.New` (unchanged) as
the behavioral-parity gate.

> Note on repo-wide checks: `task lint-fix` has a pre-existing, **unrelated** gosec G115 failure
> in `cmd/thv/app/upgrade.go`, and `task test` has a pre-existing, **unrelated**
> Docker-dependent failure in `pkg/mcp/server` (`TestBuildServerConfig` "no available runtime
> found"). Neither is caused by this change; the affected packages are untouched here.

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/server/serve.go` | Relocate the three SDK hooks and two-phase session-creation wiring (transport session manager, `buildSessionDataStorage` with the `closeStorageOnErr` guard, vMCP session manager) into `Serve`; rework `ServerConfig` (see below). |
| `pkg/vmcp/server/serve_session_test.go` | New: `Serve`-level coverage of the relocated hooks + two-phase creation driven through the SDK lifecycle, plus `buildSessionDataStorage` provider selection. |
| `pkg/vmcp/server/serve_test.go` | Update existing `Serve` tests for the new `ServerConfig` shape (`SessionManagerConfig` + `BackendRegistry`); add nil-`BackendRegistry` validation case; register `t.Cleanup` teardown. |

## Implementation details

- **`ServerConfig` reshaped**: added a pre-built `*sessionmanager.FactoryConfig` slot
  (`SessionManagerConfig`) and a shared `BackendRegistry`. The now-redundant standalone
  `SessionFactory` / `OptimizerFactory` / `OptimizerConfig` slots were folded into the
  `FactoryConfig`. Both new collaborators are validated and fail loudly with
  `vmcp.ErrInvalidConfig` when nil — `BackendRegistry` in particular, because the
  `OnRegisterSession` hook enumerates it inside a goroutine where the error would otherwise be
  swallowed.
- **Mechanical relocation**: the hook callbacks and the `*Server` receiver methods
  (`handleSessionRegistration`, `lazyInjectSessionTools`) are unchanged — they already operate
  against `s.vmcpSessionMgr` and the SDK `server.ClientSession`. The hooks object is built up
  front (so it can be passed to `NewMCPServer`) but its callbacks are registered after `srv` is
  assembled, since they close over `srv`.
- **Resource-leak guard preserved**: `buildSessionDataStorage` keeps the `closeStorageOnErr`
  defer so the storage's background cleanup goroutine is released on any error path out of
  `Serve` (pairs acquisition with release).
- **Boundary held**: no mcp-go types cross the `VMCP` interface; the "fixed at initialize"
  capability set, identity binding, and the cross-pod Redis re-hydration path stay in
  Serve/session.

No. This relocates internal transport wiring within `pkg/vmcp/server`; `server.New`'s signature
and observable behavior are unchanged.

- **Stacked PR**: this branch is stacked on its dependency #5439. The base branch is
  `vmcp-core-p2-1_issue_5439` (not `main`) so the diff contains only this issue's two commits.
- The returned `*Server` has a live session manager and backend registry but a still-nil
  discovery manager / router / backend client at this phase. It must not be `Start()`ed or
  served on the `/` MCP route until #5441/#5442 wire those fields; the unauthenticated routes
  remain safe to serve. The new tests mount the Streamable HTTP server directly to stay within
  this task's scope.
- Follow-up tasks relocate the remaining transport concerns: the authenticated middleware chain
  (#5441), the direct `VMCP` request path (#5442), and the AS runner / status reporter /
  optimizer / health monitor lifecycle (#5443). Phase 3 (#5444) reduces `server.New` to a
  `Serve(ctx, New(...), ...)` wrapper.

### DIFF
--- a/pkg/vmcp/server/serve.go
+++ b/pkg/vmcp/server/serve.go
@@ -10,17 +10,18 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
 
 	"github.com/stacklok/toolhive/pkg/audit"
 	asrunner "github.com/stacklok/toolhive/pkg/authserver/runner"
 	"github.com/stacklok/toolhive/pkg/telemetry"
+	transportsession "github.com/stacklok/toolhive/pkg/transport/session"
 	"github.com/stacklok/toolhive/pkg/vmcp"
 	vmcpconfig "github.com/stacklok/toolhive/pkg/vmcp/config"
 	"github.com/stacklok/toolhive/pkg/vmcp/core"
 	"github.com/stacklok/toolhive/pkg/vmcp/health"
-	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
-	vmcpsession "github.com/stacklok/toolhive/pkg/vmcp/session"
+	"github.com/stacklok/toolhive/pkg/vmcp/server/sessionmanager"
 	vmcpstatus "github.com/stacklok/toolhive/pkg/vmcp/status"
 )
 
@@ -29,8 +30,14 @@ import (
 // It carries the subset of the legacy server.Config that configures the HTTP/SDK
 // runtime, plus the cross-cutting TelemetryProvider/AuditConfig that are consumed
 // by both the core (core.New) and the transport (Serve) — not a clean partition.
-// The fields the core owns (Aggregator, Router, BackendRegistry, BackendClient,
-// WorkflowDefs, Authz) live on core.Config instead and are absent here.
+// The fields the core exclusively owns (Aggregator, Router, BackendClient, Authz)
+// live on core.Config instead and are absent here. Two collaborators are shared
+// rather than core-owned, so they appear on both configs: BackendRegistry (the
+// Serve session layer also enumerates backends when creating a session) and the
+// composite-tool WorkflowDefs (carried indirectly via SessionManagerConfig, whose
+// ComposerFactory closes over the core-owned router/backend client). The
+// composition root passes the same BackendRegistry instance to both core.New and
+// Serve and assembles SessionManagerConfig.
 //
 // The name intentionally stutters as server.ServerConfig: it is mandated by the
 // New/Serve split, and the package's existing Config is the legacy transport
@@ -87,21 +94,26 @@ type ServerConfig struct {
 	// the /readyz endpoint to gate readiness on cache sync.
 	Watcher Watcher
 
-	// SessionFactory creates MultiSessions for session management. Required; Serve
-	// returns an error when it is nil.
-	SessionFactory vmcpsession.MultiSessionFactory
+	// BackendRegistry enumerates the configured backends. It is a shared
+	// collaborator: the core (core.Config.BackendRegistry) consumes it for
+	// capability aggregation, and the Serve session layer consumes it here — when
+	// the OnRegisterSession hook fires, sessionmanager.CreateSession lists the
+	// registry's backends to open the session's backend connections. The
+	// composition root passes the same instance to both core.New and Serve.
+	BackendRegistry vmcp.BackendRegistry
 
 	// SessionStorage configures the session storage backend. When nil or provider is
 	// "memory", local in-process storage is used.
 	SessionStorage *vmcpconfig.SessionStorageConfig
 
-	// OptimizerFactory builds an optimizer from a list of tools. If nil, the
-	// optimizer is disabled.
-	OptimizerFactory func(context.Context, []server.ServerTool) (optimizer.Optimizer, error)
-
-	// OptimizerConfig holds the parsed optimizer search parameters. A nil value
-	// disables the optimizer.
-	OptimizerConfig *optimizer.Config
+	// SessionManagerConfig is the pre-built construction config for the vMCP session
+	// manager (sessionmanager.New). FactoryConfig.Base is the required underlying
+	// MultiSessionFactory; the config also carries the composite-tool WorkflowDefs and
+	// ComposerFactory, the optimizer factory/config, and the telemetry provider.
+	// Required; Serve returns an error when it is nil. The composition root assembles
+	// it because the ComposerFactory closes over the core-owned router and backend
+	// client.
+	SessionManagerConfig *sessionmanager.FactoryConfig
 
 	// TelemetryProvider is the cross-cutting telemetry provider (also consumed by
 	// core.New). If nil, no telemetry is recorded.
@@ -115,61 +127,104 @@ type ServerConfig struct {
 // Serve is the transport-side entry point of the New/Serve split: it wraps an
 // already-constructed core [core.VMCP] and returns the existing *Server.
 //
-// At this phase Serve constructs only the self-contained transport pieces: it
-// applies the transport defaults (mirroring server.New), builds the mcp-go server,
-// and wires the core's Close into the server's shutdown sequence. It does NOT build
-// the route mux or the HTTP lifecycle here — those remain in the carried-forward
-// (*Server).Handler/Start/Stop, which register the unauthenticated routes (/health,
-// /ping, /readyz, /status, /api/backends/health, the metrics and .well-known
-// endpoints, and any embedded auth-server routes) when Serve's *Server is served.
+// At this phase Serve constructs the self-contained transport pieces — it applies
+// the transport defaults (mirroring server.New), builds the mcp-go server, and wires
+// the core's Close into the server's shutdown sequence — plus the SDK session layer:
+// the three session hooks and the two-phase session-creation wiring (the transport
+// session manager, the session data storage backend, and the vMCP session manager).
+// It does NOT build the route mux or the HTTP lifecycle here — those remain in the
+// carried-forward (*Server).Handler/Start/Stop, which register the unauthenticated
+// routes (/health, /ping, /readyz, /status, /api/backends/health, the metrics and
+// .well-known endpoints, and any embedded auth-server routes) when Serve's *Server
+// is served.
 //
-// The remaining transport concerns — the SDK session hooks and two-phase session
-// creation, the authenticated middleware chain, the direct VMCP request path, and
-// the AS runner / status reporter / optimizer / health monitor lifecycle — are
-// relocated under Serve by the subsequent Phase 2 tasks. server.New is not yet
-// routed through Serve, so this is purely additive and observable behavior is
-// unchanged.
+// The remaining transport concerns — the authenticated middleware chain (#5441), the
+// direct VMCP request path (#5442), and the AS runner / status reporter / optimizer /
+// health monitor lifecycle (#5443) — are relocated under Serve by the subsequent
+// Phase 2 tasks. server.New is not yet routed through Serve and keeps its own copy of
+// the session wiring until Phase 3, so this is purely additive and observable behavior
+// is unchanged.
 //
-// Serve returns a vmcp.ErrInvalidConfig-wrapped error for a nil cfg, a nil core, or
-// a nil required collaborator (SessionFactory). The ctx parameter is reserved for
-// the transport wiring relocated by later phases.
+// Serve returns a vmcp.ErrInvalidConfig-wrapped error for a nil cfg, a nil core, or a
+// nil required collaborator (SessionManagerConfig). The session hooks close over the
+// constructed *Server, so they are registered after it is assembled.
 //
-// Contract: the returned *Server has nil session manager, discovery manager, and
-// backend registry at this phase. It must NOT be Start()ed or served on the "/" MCP
-// route until #5440/#5441/#5442 wire those fields — the carried-forward Handler
-// passes them to WithSessionIdManager and discovery.Middleware, which would nil-deref.
-// The unauthenticated routes (registered as direct mux entries) are safe to serve.
-func Serve(_ context.Context, v core.VMCP, cfg *ServerConfig) (*Server, error) {
+// Contract: the returned *Server now has a live session manager and backend registry,
+// but a nil discovery manager, router, and backend client at this phase. It must NOT
+// be Start()ed or served on the "/" MCP route until #5441/#5442 wire those fields —
+// the carried-forward Handler passes them to discovery.Middleware, which would
+// nil-deref. The unauthenticated routes (registered as direct mux entries) are safe
+// to serve.
+func Serve(ctx context.Context, v core.VMCP, cfg *ServerConfig) (*Server, error) {
 	if cfg == nil {
 		return nil, fmt.Errorf("%w: nil server config", vmcp.ErrInvalidConfig)
 	}
 	if v == nil {
 		return nil, fmt.Errorf("%w: VMCP is required", vmcp.ErrInvalidConfig)
 	}
-	if cfg.SessionFactory == nil {
-		return nil, fmt.Errorf("%w: SessionFactory is required", vmcp.ErrInvalidConfig)
+	if cfg.SessionManagerConfig == nil {
+		return nil, fmt.Errorf("%w: SessionManagerConfig is required", vmcp.ErrInvalidConfig)
 	}
 
 	serveCfg := buildServeConfig(cfg)
 
+	// Build the SDK hooks up front so they can be passed to NewMCPServer, but
+	// register their callbacks after srv is assembled below — the callbacks close
+	// over srv (relocated from server.New, where the hooks object is co-located with
+	// the mcp-go server for the same reason).
+	hooks := &server.Hooks{}
+
 	// Build the mcp-go server (mirrors server.New): tools and resources are
-	// registered dynamically, so capabilities start disabled. Hooks are empty for
-	// now; the session hooks are relocated here by a later Phase 2 task.
+	// registered dynamically, so capabilities start disabled.
 	mcpServer := server.NewMCPServer(
 		serveCfg.Name,
 		serveCfg.Version,
 		server.WithToolCapabilities(false),
 		server.WithResourceCapabilities(false, false),
 		server.WithLogging(),
-		server.WithHooks(&server.Hooks{}),
+		server.WithHooks(hooks),
 	)
 
+	// Two-phase session-creation wiring (relocated from server.New). The transport
+	// session manager holds the lightweight Streamable HTTP placeholders; the
+	// pluggable data storage holds serialisable session metadata (memory or Redis,
+	// reading THV_SESSION_REDIS_PASSWORD); the vMCP session manager owns the live,
+	// node-local MultiSessions.
+	sessionManager := transportsession.NewManager(serveCfg.SessionTTL, transportsession.NewStreamableSession)
+
+	sessionDataStorage, err := buildSessionDataStorage(ctx, serveCfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create session data storage: %w", err)
+	}
+	// Close sessionDataStorage if Serve returns an error after this point so the
+	// background cleanup goroutine does not leak (go-style: pair acquisition with
+	// release).
+	closeStorageOnErr := true
+	defer func() {
+		if closeStorageOnErr {
+			_ = sessionDataStorage.Close()
+		}
+	}()
+
+	vmcpSessMgr, optimizerCleanup, err := sessionmanager.New(sessionDataStorage, cfg.SessionManagerConfig, cfg.BackendRegistry)
+	if err != nil {
+		return nil, err
+	}
+
 	srv := &Server{
-		config:         serveCfg,
-		mcpServer:      mcpServer,
-		ready:          make(chan struct{}),
-		healthMonitor:  cfg.HealthMonitor,
-		statusReporter: cfg.StatusReporter,
+		config:             serveCfg,
+		mcpServer:          mcpServer,
+		backendRegistry:    cfg.BackendRegistry,
+		sessionManager:     sessionManager,
+		sessionDataStorage: sessionDataStorage,
+		vmcpSessionMgr:     vmcpSessMgr,
+		ready:              make(chan struct{}),
+		healthMonitor:      cfg.HealthMonitor,
+		statusReporter:     cfg.StatusReporter,
+	}
+
+	if optimizerCleanup != nil {
+		srv.shutdownFuncs = append(srv.shutdownFuncs, optimizerCleanup)
 	}
 
 	// Serve owns the injected core's lifecycle: release its backend connections
@@ -179,6 +234,22 @@ func Serve(_ context.Context, v core.VMCP, cfg *ServerConfig) (*Server, error) {
 		return v.Close()
 	})
 
+	// Register the three SDK hooks against the assembled *Server (relocated from
+	// server.New). They drive phase-2 of two-phase creation (OnRegisterSession) and
+	// the cross-pod Redis re-hydration of per-session tools (OnBeforeListTools /
+	// OnBeforeCallTool); the *Server receiver methods are unchanged.
+	hooks.AddOnRegisterSession(func(hookCtx context.Context, session server.ClientSession) {
+		srv.handleSessionRegistration(hookCtx, session)
+	})
+	hooks.AddBeforeListTools(func(hookCtx context.Context, _ any, _ *mcp.ListToolsRequest) {
+		srv.lazyInjectSessionTools(hookCtx)
+	})
+	hooks.AddBeforeCallTool(func(hookCtx context.Context, _ any, _ *mcp.CallToolRequest) {
+		srv.lazyInjectSessionTools(hookCtx)
+	})
+
+	// Disarm the close-on-error guard: the Server is fully constructed.
+	closeStorageOnErr = false
 	return srv, nil
 }
 
@@ -188,7 +259,7 @@ func Serve(_ context.Context, v core.VMCP, cfg *ServerConfig) (*Server, error) {
 // mutating the caller's ServerConfig (go-style: copy before mutating caller input).
 // Port 0 is left untouched to mean "OS-assigned".
 //
-// Three Config fields are deliberately NOT mapped at this phase (see
+// Several Config fields are deliberately NOT mapped at this phase (see
 // TestBuildServeConfigMapsSharedFields, which guards this list against drift):
 //   - AuthzMiddleware: the authenticated/authz middleware chain is relocated under
 //     Serve by #5441, after which authorization moves to the core admission seam.
@@ -198,6 +269,10 @@ func Serve(_ context.Context, v core.VMCP, cfg *ServerConfig) (*Server, error) {
 //   - StatusReporter: Serve assigns ServerConfig.StatusReporter to the Server field
 //     directly (like HealthMonitor); nothing reads Config.StatusReporter on the Serve
 //     path, so mapping it here would be dead.
+//   - SessionFactory, OptimizerFactory, OptimizerConfig: the vMCP session manager is
+//     built directly in Serve from ServerConfig.SessionManagerConfig (a pre-built
+//     *sessionmanager.FactoryConfig that carries the session factory and optimizer
+//     wiring), not via Config→New, so these Config fields are unused on the Serve path.
 func buildServeConfig(cfg *ServerConfig) *Config {
 	return &Config{
 		Name:                    cmp.Or(cfg.Name, defaultServerName),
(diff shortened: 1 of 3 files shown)


pkg/vmcp/server/serve.go
```diff
@@ -87,21 +94,32 @@ type ServerConfig struct {
...
-	OptimizerConfig *optimizer.Config
+	// SessionManagerConfig is the pre-built construction config for the vMCP session
+	// manager (sessionmanager.New). FactoryConfig.Base is the required underlying
+	// MultiSessionFactory; the config also carries the composite-tool WorkflowDefs and
+	// ComposerFactory, the optimizer factory/config, and the telemetry provider.
+	// Required; Serve returns an error when it is nil. The composition root assembles
+	// it because the ComposerFactory closes over the core-owned router and backend
+	// client.
+	//
+	// Caller responsibility: unlike server.New, Serve does NOT run validateWorkflows on
```
tgrunnagle, line 117: **[MEDIUM] `validateWorkflows` obligation moved without an enforcement seam**

`server.New` runs `validateWorkflows` (server.go:410 — cycle detection + tool-reference validation via `composer.ValidateWorkflow`) before building the session-manager config. `Serve` omits this; `sessionmanager.New` only checks the `WorkflowDefs`/`ComposerFactory` *pairing*, not workflow content. The divergence is honestly documented here as a caller responsibility, but it's pushed onto an unenforced contract.

**Impact:** when `server.New` is routed through `Serve` in Phase 3, if the composition root forgets this step, malformed composite-tool definitions — including cycles, which `validateWorkflows` flags as a resource-exhaustion concern — would be accepted silently rather than rejected at startup. Not a current bug (Serve is not yet on the live path).

**Recommendation:** acceptable for this phase since it's documented and off the live path. Track it explicitly — add a Phase 3 (#5444) acceptance criterion requiring the composition root to call `validateWorkflows` before assembling `SessionManagerConfig`, and consider moving cycle/reference validation into `sessionmanager.New` itself so the safety property travels with construction rather than relying on caller discipline.

tgrunnagle (commented):
## Multi-agent review — Move SDK hooks + two-phase session creation under Serve

_Reviewed by 4 specialist agents (Go correctness, vMCP architecture, test coverage, general quality). Codex cross-review skipped (CLI not installed)._

**This is a faithful, behavior-preserving relocation.** The block moved into `Serve` is a near-byte-for-byte copy of `server.New` (server.go:338–522), differing only by a safe `ctx`→`hookCtx` rename. The SDK boundary holds — zero mcp-go imports leak onto `core.VMCP`. `ServerConfig` reshape *reduces* config surface by reusing the existing `FactoryConfig`. All acceptance criteria in #5440 are met and scope is well under the 400 LOC / 10 file budget. On a normal PR this would be an **APPROVE**; left as a COMMENT because the PR is a draft (and GitHub disallows self-approval).

### Findings

| # | Severity | Finding | Location |
|---|----------|---------|----------|
| M1 | MEDIUM | `validateWorkflows` obligation moved to composition root with no enforcement seam | serve.go:117 |
| M2 | MEDIUM | Cross-pod re-injection path untested; PR description's "covered by integration suite" claim is inaccurate | serve_session_test.go:103 |
| L1 | LOW | `sessionManager` cleanup goroutine leaks on the Serve error path (asymmetric with the new storage guard) | serve.go:206 |
| L2 | LOW | `toolSessionState.lastSession` written under mutex but never read (dead state) | serve_session_test.go:45 |
| L3 | LOW | `TestServeClosesStorageOnSessionManagerError` name overstates — does not observe `Close()` | serve_session_test.go:169 |
| L4 | LOW | Redis test assertion not bound to the Redis branch | serve_session_test.go:245 |

**0 HIGH — no blocking issues.** The two MEDIUM items are worth resolving (or consciously deferring) before this leaves draft: M2 is the only one really worth acting on *in this PR* — either add the missing test or correct the description. M1 and L1 are reasonable to track for Phase 3 (#5444) when the two wiring copies collapse. L2–L4 are minor test-quality nits. Details are in the inline comments.

_Dropped below threshold (no action): a wording nit on the `discovery.Middleware` nil-deref comment (the named mechanism was independently verified accurate, just incomplete); a doc clarification that the telemetry provider is intentionally set in two places; and an optional `.Times(1)` hardening of the `MakeSessionWithID` mock._

pkg/vmcp/server/serve_session_test.go
```diff
@@ -0,0 +1,267 @@
...
+		}).AnyTimes()
+	return factory, state
+}
+
+// TestServeRegistersSessionHooks verifies that Serve registers the OnRegisterSession
+// hook and wires two-phase session creation: an MCP initialize triggers the hook,
+// which creates the session (MakeSessionWithID) and injects its tools, so a
+// subsequent tools/list advertises them. The OnBeforeListTools hook also runs (and
+// no-ops, since the tools are already present on this pod).
+func TestServeRegistersSessionHooks(t *testing.T) {
```
tgrunnagle, line 98: **[MEDIUM] Cross-pod re-injection path is untested — and the PR description's coverage claim doesn't hold**

The issue's Testing Strategy requires a test that `lazyInjectSessionTools` **injects from the vMCP session manager when the SDK tool store is empty** (the cross-pod case). This test only exercises the pod-local **no-op** branch: `initialize` runs `OnRegisterSession`, which populates the store, so the subsequent `OnBeforeListTools` hits the `len(GetSessionTools()) > 0` early return (server.go:1095) and never reaches the injection branch (server.go:1099–1106).

A grep of the whole `pkg/vmcp/server` test tree found no test driving `tools/list`/`tools/call` against a session ID without a preceding `initialize` on the same server — so the PR description's claim that this path is "covered by the existing integration suite on `server.New`" does not appear to hold.

**Impact:** low for *this* PR — the relocated receiver methods are byte-for-byte unchanged, so the relocation itself doesn't regress the behavior. But the documented edge case has no regression guard anywhere.

**Recommendation:** either (a) add a direct Serve-level test of the empty-store branch (unit-test `lazyInjectSessionTools` with a fake `SessionWithTools` whose `GetSessionTools()` returns empty, asserting `setSessionToolsDirect` runs), or (b) correct the PR description to stop claiming integration coverage that doesn't exist.

pkg/vmcp/server/serve.go
```diff
@@ -115,61 +133,111 @@ type ServerConfig struct {
...
-		server.WithHooks(&server.Hooks{}),
+		server.WithHooks(hooks),
 	)
 
+	// Two-phase session-creation wiring (relocated from server.New). The transport
+	// session manager holds the lightweight Streamable HTTP placeholders; the
+	// pluggable data storage holds serialisable session metadata (memory or Redis,
+	// reading THV_SESSION_REDIS_PASSWORD); the vMCP session manager owns the live,
+	// node-local MultiSessions.
+	sessionManager := transportsession.NewManager(serveCfg.SessionTTL, transportsession.NewStreamableSession)
```
tgrunnagle: **[LOW] `sessionManager` cleanup goroutine leaks on the Serve error path**

`transportsession.NewManager` starts `go m.cleanupRoutine()` immediately (manager.go:86), released only by `Manager.Stop()`. `sessionManager` is built here — *before* the fallible `buildSessionDataStorage` (line 208) and `sessionmanager.New` (line 222). On either error path, `Serve` returns an error; the new `closeStorageOnErr` defer closes `sessionDataStorage` but `sessionManager.Stop()` is never called, leaking its goroutine.

**Impact:** narrow — only on a construction error after partial wiring. This is a verbatim carry-over of the same gap in `server.New`, so not a regression. Notable only because this PR adds a leak guard for the *sibling* resource in the same block while leaving `sessionManager` unguarded.

**Recommendation:** either construct `sessionManager` *after* the fallible calls, or extend the deferred guard to also `sessionManager.Stop()` on the error path. Reasonable to defer to Phase 3 when the two copies collapse — but worth recording so it isn't lost.

pkg/vmcp/server/serve_session_test.go
```diff
@@ -0,0 +1,267 @@
...
+// the authenticated discovery middleware (relocated by #5441/#5442). That keeps the
+// test within this task's scope while proving the hooks fire and two-phase session
+// creation runs identically when Serve is exercised directly. The full HTTP suite
+// stays on server.New (its parity gate) in session_management_integration_test.go.
+
+// toolSessionState tracks observable behaviour of the mock session factory below.
+type toolSessionState struct {
+	makeWithIDCalled atomic.Bool
+	mu               sync.Mutex
+	lastSession      *sessionmocks.MockMultiSession
```
tgrunnagle: **[LOW] Dead test state: `lastSession` is written but never read**

`toolSessionState` carries `mu sync.Mutex` + `lastSession`, written under lock further down, but `lastSession` is never read in this file — a copy-from-template artifact from `session_management_integration_test.go` (whose analogue *does* read it). The `mu`/`lastSession` pair is dead state implying an assertion that never happens.

**Recommendation:** remove `mu` and `lastSession` (keep only `makeWithIDCalled`). The HTTP-level `tools/list` body assertion already proves the session was created and surfaced its tools.

pkg/vmcp/server/serve_session_test.go
```diff
@@ -0,0 +1,267 @@
...
+	require.Equal(t, http.StatusOK, listResp.StatusCode, "tools/list should succeed; body: %s", string(body))
+	assert.Contains(t, string(body), testTool.Name,
+		"tools/list should advertise the tool injected by the OnRegisterSession hook")
+}
+
+// TestServeClosesStorageOnSessionManagerError verifies the closeStorageOnErr guard:
+// when sessionmanager.New fails after the session data storage is built, Serve
+// returns the error (so the deferred guard closes the storage and its cleanup
+// goroutine does not leak). A negative CacheCapacity is the cheapest forced failure.
+func TestServeClosesStorageOnSessionManagerError(t *testing.T) {
```
tgrunnagle: **[LOW] Test name overstates what it verifies**

This test proves the error path is *reached* (failure occurs after `buildSessionDataStorage`, confirmed by `ErrorContains("CacheCapacity")` + `NotErrorIs(ErrInvalidConfig)`), but it cannot observe `sessionDataStorage.Close()` — the concrete storage is built internally and the test holds no handle to it. A regression that deleted the `defer` guard would still pass this test.

**Recommendation:** either rename to match the assertions (e.g., `TestServeReturnsErrorWhenSessionManagerConstructionFails`), or use `uber-go/goleak` (`goleak.VerifyNone`) so a missing `Close()` surfaces as a leaked cleanup-ticker goroutine — the only way to truly guard the documented no-leak invariant.

pkg/vmcp/server/serve_session_test.go
```diff
@@ -0,0 +1,267 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+	ds, err := buildSessionDataStorage(ctx, &Config{
+		SessionTTL: time.Minute,
+		SessionStorage: &vmcpconfig.SessionStorageConfig{
+			Provider: "redis",
+			Address:  "127.0.0.1:1",
+		},
+	})
+	require.Error(t, err)
```
tgrunnagle, line 329: **[LOW] Redis test assertion is not bound to the Redis branch**

This asserts only `require.Error` + `assert.Nil(ds)`. The "in-process would not error" discriminator is sound, but *any* error passes — a future unsupported-provider or config-validation error would satisfy it without exercising the Redis path.

**Recommendation:** bind the assertion to the Redis path, e.g. `assert.ErrorContains(t, err, "redis")` or match the dial/connection-refused substring — mirroring how the memory test asserts `ErrorContains("unsupported session storage provider")`.

pkg/vmcp/server/serve_session_test.go
tgrunnagle: Addressed in 5c5103d1 — added `TestServeLazyInjectsToolsForRehydratedSession`, a direct Serve-level test of the empty-store inject branch: a session is registered in the vMCP session manager, then a fresh SDK `ClientSession` with an empty per-session tool store (as on a second pod, where `OnRegisterSession` never fired) gets its tools re-injected by `lazyInjectSessionTools`; the populated-store no-op branch is covered too. Also corrected the PR description's test section to list this test and drop the inaccurate integration-coverage claim.

pkg/vmcp/server/serve.go
tgrunnagle: Addressed in 5c5103d1 — the transport `sessionManager` is now constructed *after* the fallible `buildSessionDataStorage`/`sessionmanager.New` calls, so an error return can no longer leak its cleanup goroutine. This restores symmetry with the `closeStorageOnErr` guard on the sibling storage (rather than extending the guard).

pkg/vmcp/server/serve_session_test.go
tgrunnagle: Addressed in 5c5103d1 — removed the dead `mu`/`lastSession` state; `toolSessionState` now carries only `makeWithIDCalled`. The HTTP-level `tools/list` assertion already proves the session was created and surfaced its tools.

pkg/vmcp/server/serve_session_test.go
tgrunnagle: Addressed in 5c5103d1 — renamed to `TestServeReturnsErrorWhenSessionManagerConstructionFails` and reworded the doc comment to state it verifies the guarded error path is *reached* (failure after storage construction, distinct from config validation), not that it observes `Close()`.

pkg/vmcp/server/serve_session_test.go
tgrunnagle: Addressed in 5c5103d1 — bound the assertion to the Redis connection path: `assert.ErrorContains(err, "redis: failed to connect")`. Used the connection-failure wrap rather than "redis" since the unsupported-provider error text also lists "redis", so that substring wouldn't discriminate.

pkg/vmcp/server/serve.go
tgrunnagle: Deferring per your recommendation — this is off the live path (Serve isn't routed yet) and the responsibility shift is already documented on `SessionManagerConfig`. I'll track it as a Phase 3 (#5444) acceptance criterion: the composition root must run `validateWorkflows` before assembling `SessionManagerConfig`. Your stronger suggestion — moving cycle/reference validation into `sessionmanager.New` so the safety property travels with construction rather than caller discipline — is the better long-term fix; I'll raise it on #5444 since it's an interface change out of scope for this relocation PR.

jerm-dro (approved):

PR #5479, opened 2026-06-10
Wire RFC 9728 discovery into HTTP and stdio transports

jhrozek:
## Summary

Before this change, `transport.Factory.Create` never passed `authInfoHandler` or `prefixHandlers` to `StdioTransport`. For any stdio-backed `MCPServer` with an embedded auth server (`spec.authServerRef`), every OAuth/OIDC discovery endpoint returned 404 — the RFC 9728 protected-resource document, the RFC 8414 authorization server metadata, the OIDC discovery document, and the JWKS endpoint were all unreachable. MCP clients that follow the OAuth discovery chain could not bootstrap auth against stdio-backed servers.

- Add `authInfoHandler` and `prefixHandlers` fields to `StdioTransport`; forward both to the underlying streamable/SSE proxy in `Start()`
- Add the same two fields and matching functional options to `HTTPSSEProxy` (`pkg/transport/proxy/httpsse/http_proxy.go`) and `HTTPProxy` (`pkg/transport/proxy/streamable/streamable_proxy.go`)
- Mount `/.well-known/` in both proxy backends so OAuth clients always get JSON rather than Go's plain-text 404 when auth is off
- Wire both fields through `factory.Create` for the stdio case (SSE and streamable-HTTP were already wired on main; this PR adds the 6-line stdio path)
- Add factory-level tests for SSE and streamable-HTTP confirming `authInfoHandler` and `prefixHandlers` survive the `Create` call

## Type of change

- [x] Bug fix

## Test plan

Deployed a stdio-backed `MCPServer` (`transport: stdio`, `proxyMode: streamable-http`) with `spec.authServerRef` pointing to an embedded auth server backed by Okta in a kind cluster. Port-forwarded the proxy service and verified all four endpoints return HTTP 200 with valid JSON:

| Endpoint | Result |
|---|---|
| `/.well-known/oauth-protected-resource` | ✅ 200 — `resource` + `authorization_servers` fields present |
| `/.well-known/oauth-authorization-server` | ✅ 200 — full RFC 8414 document |
| `/.well-known/openid-configuration` | ✅ 200 — full OIDC discovery document |
| `/.well-known/jwks.json` | ✅ 200 — EC P-256 signing key |

## Changes

| File | Change |
|------|--------|
| `pkg/transport/stdio.go` | Add `authInfoHandler`/`prefixHandlers` fields and setters; forward to proxy in `Start()` |
| `pkg/transport/factory.go` | Wire both fields from `Config` into `StdioTransport` (the stdio-only path) |
| `pkg/transport/proxy/httpsse/http_proxy.go` | Add `authInfoHandler`/`prefixHandlers` fields + `WithAuthInfoHandler`/`WithPrefixHandlers` options; mount `/.well-known/` in `Start()` |
| `pkg/transport/proxy/streamable/streamable_proxy.go` | Same as httpsse |
| `pkg/transport/stdio_test.go` | Factory tests for SSE and streamable-HTTP preserving both fields |
| `pkg/transport/proxy/httpsse/http_proxy_integration_test.go` | Add `TestHTTPSSEProxy_StartMountsAuthDiscoveryEndpoint` |
| `pkg/transport/proxy/streamable/streamable_proxy_integration_test.go` | Integration test for streamable proxy auth discovery endpoint |
| `pkg/transport/proxy/streamable/streamable_proxy_test.go` | Unit tests for streamable proxy options |

Yes. For `MCPServer` resources using `transport: stdio` with `spec.authServerRef`, the OAuth/OIDC discovery endpoints (`/.well-known/oauth-protected-resource`, `/.well-known/oauth-authorization-server`, `/.well-known/openid-configuration`, `/.well-known/jwks.json`) now return correct responses instead of 404. MCP clients that follow RFC 9728 discovery can now bootstrap OAuth against stdio-backed servers.

### DIFF
--- a/pkg/transport/proxy/httpsse/http_proxy_test.go
+++ b/pkg/transport/proxy/httpsse/http_proxy_test.go
@@ -181,7 +181,7 @@ func TestConcurrentClientRemoval(t *testing.T) {
 //nolint:paralleltest // Test modifies shared proxy state
 func TestForwardResponseToClients(t *testing.T) {
 	proxy := NewHTTPSSEProxy("localhost", 8080, false, nil, nil)
-	ctx := context.Background()
+	ctx := t.Context()
 
 	// Create a client session
 	clientID := testClientID
@@ -220,7 +220,7 @@ func TestForwardResponseToClients(t *testing.T) {
 //nolint:paralleltest // Test modifies shared proxy state
 func TestForwardResponseToClients_NoClients(t *testing.T) {
 	proxy := NewHTTPSSEProxy("localhost", 8080, false, nil, nil)
-	ctx := context.Background()
+	ctx := t.Context()
 
 	// Create a test response
 	response, err := jsonrpc2.NewResponse(jsonrpc2.StringID("test"), "test result", nil)
@@ -609,7 +609,7 @@ func TestNewHTTPSSEProxyWithSessionStorage(t *testing.T) {
 //nolint:paralleltest // Test starts/stops HTTP server
 func TestNewHTTPSSEProxy_AuthInfoHandlerNil(t *testing.T) {
 	proxy := NewHTTPSSEProxy("localhost", 0, false, nil, nil) // no WithAuthInfoHandler
-	ctx := context.Background()
+	ctx := t.Context()
 
 	require.NoError(t, proxy.Start(ctx))
 	t.Cleanup(func() {
@@ -643,7 +643,7 @@ func TestNewHTTPSSEProxy_PrefixHandlerMounted(t *testing.T) {
 	proxy := NewHTTPSSEProxy("localhost", 0, false, nil, nil,
 		WithPrefixHandlers(map[string]http.Handler{"/oauth/token": sentinel}),
 	)
-	ctx := context.Background()
+	ctx := t.Context()
 
 	require.NoError(t, proxy.Start(ctx))
 	t.Cleanup(func() {
@@ -676,7 +676,7 @@ func TestNewHTTPSSEProxy_PrefixHandlerDoesNotShadowSSE(t *testing.T) {
 	proxy := NewHTTPSSEProxy("localhost", 0, false, nil, nil,
 		WithPrefixHandlers(map[string]http.Handler{"/oauth/token": sentinel}),
 	)
-	ctx := context.Background()
+	ctx := t.Context()
 
 	require.NoError(t, proxy.Start(ctx))
 	t.Cleanup(func() {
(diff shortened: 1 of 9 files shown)


pkg/transport/proxy/httpsse/http_proxy_test.go
```diff
@@ -603,6 +603,101 @@ func TestNewHTTPSSEProxyWithSessionStorage(t *testing.T) {
 	require.NotNil(t, proxy.sessionManager)
 }
 
+// TestNewHTTPSSEProxy_AuthInfoHandlerNil verifies that when no authInfoHandler is set,
+// requests to /.well-known/oauth-protected-resource return a JSON 404.
+//
+//nolint:paralleltest // Test starts/stops HTTP server
+func TestNewHTTPSSEProxy_AuthInfoHandlerNil(t *testing.T) {
+	proxy := NewHTTPSSEProxy("localhost", 0, false, nil, nil) // no WithAuthInfoHandler
+	ctx := context.Background()
```
JAORMX: We use `t.Context()` instead of `context.Background()` in tests (the only exception is mock callback signatures). Applies to all nine new `Start(ctx)` call sites across the four test files. The `context.Background()` inside `t.Cleanup` is fine as-is, since `t.Context()` is already cancelled by the time cleanup runs.

pkg/transport/stdio_test.go
```diff
@@ -1036,3 +1038,196 @@ func TestStdioTransport_ShouldRestart(t *testing.T) {
...
+	tr := NewStdioTransport("localhost", 8080, nil, false, false, nil)
+	tr.SetPrefixHandlers(handlers)
+
+	require.Len(t, tr.prefixHandlers, 1)
+	require.Contains(t, tr.prefixHandlers, "/oauth/token")
+}
+
+// TestFactory_Create_StdioPreservesAuthInfoHandler verifies that the factory
+// forwards AuthInfoHandler from Config to the StdioTransport it creates.
+func TestFactory_Create_StdioPreservesAuthInfoHandler(t *testing.T) {
```
JAORMX: These six factory tests are the same test with a different `Config` field and type assertion. One table-driven test would do it. Same goes for the start-then-poll-then-cleanup block that's copy-pasted eight times across the test files... a small `t.Helper()` that starts the proxy and returns the base URL would trim a lot of this.

pkg/transport/proxy/streamable/streamable_proxy_test.go
```diff
@@ -214,3 +215,137 @@ func TestNewHTTPProxyWithSessionStorage(t *testing.T) {
...
+	// WellKnownHandler returns a JSON 404, not 200.
+	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+	assert.Contains(t, resp.Header.Get("Content-Type"), "application/json")
+}
+
+// TestNewHTTPProxy_PrefixHandlerMounted verifies that a handler registered via
+// WithPrefixHandlers is reachable at its declared path prefix.
+//
+//nolint:paralleltest // Test starts/stops HTTP server
+func TestNewHTTPProxy_PrefixHandlerMounted(t *testing.T) {
```
JAORMX, line 287: None of the new tests pin the actual crux: an exact `/.well-known/openid-configuration` prefix handler (which is what `Routes()` actually registers) coexisting with the `/.well-known/` PRM mount on the same mux. The whole fix relies on ServeMux specificity sending the exact pattern to the AS handler and `oauth-protected-resource` to the well-known handler. One test per proxy asserting both land in the right handler would lock that in, and would catch a regression if anyone ever "simplifies" the mounting order.

JAORMX (changes requested):
The core fix is right and I want it merged. I went down the rabbit hole on the part this whole PR hangs on: two handlers competing for the `/.well-known/` subtree. The embedded AS registers exact patterns (`openid-configuration`, `oauth-authorization-server`, `jwks.json`) while the proxies register the `/.well-known/` subtree for the PRM handler, and ServeMux resolves by specificity, not registration order. So the exact patterns win, `oauth-protected-resource` falls through to the PRM handler, and there's no duplicate-pattern panic. Also checked RFC 9728 section 3.1: the path-insertion form works since `NewWellKnownHandler` matches subpaths. And `buildWWWAuthenticate` does the same host/path insertion on the 401 side, so the discovery chain actually closes end to end for stdio now. Nice.

The PR description needs a pass though; three claims don't match the diff:

- `HTTPTransport` is untouched. It already had `authInfoHandler`/`prefixHandlers` on main; what this PR extends is the httpsse and streamable proxy structs that only stdio uses.
- The factory already wired both fields for SSE and streamable-HTTP on main. Only the stdio case is new (+6 lines in factory.go).
- The diff is +621/-0, so "fix a flaky integration test" and "remove a duplicate unit test" never happened relative to main. Those bullets describe intra-branch history; please drop them.

Worth fixing since the description is what release notes and future archaeology will read.

Follow-ups I'd like as issues rather than blockers here (all pre-existing patterns that this PR extends to stdio workloads):

- `POST /oauth/register` is open dynamic client registration with no rate limiting, and it's now reachable on the Service-exposed proxy port of stdio workloads. Any pod in the cluster can fill the DCR store unless a NetworkPolicy says otherwise.
- `/oauth/token` and `/oauth/register` bypass the audit middleware, so credential issuance leaves no audit trail. Same as the transparent proxy today, but token issuance going unaudited deserves its own issue.
- `Routes()` mounts the whole `/oauth/` subtree, so any future route added under it ships unauthenticated and middleware-free automatically. An exact-path enumeration, or at least a test pinning the route set, would make additions conscious. Related: if `Routes()` ever returns `/metrics` or `/.well-known/` as a key, `mux.Handle` panics at workload start... in three proxies now.

Inline comments for the test bits.

JAORMX (approved):

pkg/transport/proxy/httpsse/http_proxy_test.go
jhrozek: Done — replaced all Start(ctx) call sites across the four test files with t.Context(). The context.Background() calls inside t.Cleanup stay as-is.

pkg/transport/stdio_test.go
jhrozek: Consolidated the six into a single TestFactory_Create_PreservesAuthFields table-driven test. The start-then-poll helper duplication is a good follow-up — deferring it to keep this PR focused.

pkg/transport/proxy/streamable/streamable_proxy_test.go
jhrozek: Added TestNewHTTPSSEProxy_ExactWellKnownBeatsSubtree and TestNewHTTPProxy_ExactWellKnownBeatsSubtree. Each registers a 418 sentinel at /.well-known/openid-configuration alongside a 200 PRM handler and asserts both: exact path returns 418, oauth-protected-resource returns 200.

reyortiz3 (approved):

Issue #5445, opened 2026-06-03
P3.2 Reduce server.New body to the wrapper


tgrunnagle:
## Description

Reduce `server.New`'s **body** to the thin wrapper
`Serve(ctx, New(deriveCoreConfig(cfg, …)), deriveServerConfig(cfg, healthMon))`, keeping
the existing 7-param signature byte-for-byte unchanged. This is the moment the
god-object `server.New` (anti-pattern #3) is dismantled and the **single live path**
becomes the `New`/`Serve` split — so it is also where the A1-deferred cleanup lands:
build the health monitor at the composition root (A2), then **delete the now-dead
authz/annotation HTTP blocks and retire the discovery middleware/seam + its `s.core ==
nil` guard** that Phase 2 left in place behind guards, **and delete `session.WithAggregator`
+ the factory `aggregator` field** so the AC2 "factory must not aggregate" contract (#5442)
becomes structural rather than a remembered comment. Both `//nolint:gocyclo` directives
are removed (the complexity is gone, not suppressed). This is the **highest-integration-risk
PR in the epic**: until now the behavioral-parity suite exercised the legacy path; here
`server.New` is rerouted through `Serve` for real, so the full parity suite is the gate.

## Context

This is the final code change of Phase 3 (RFC THV-0076). #5444 introduced
`deriveCoreConfig`/`deriveServerConfig`, which split the in-memory `server.Config`
into the core `Config` and transport `ServerConfig` (and threaded the built health
monitor). This task wires those derivers into `server.New` so its body delegates entirely
to `New` + `Serve`. Per architecture.md "server.New compatibility wrapper" (14-17, 43-47,
368, 480-482, 501), the 7-param signature and observable MCP behavior stay stable
throughout — the `server.New(...)` call site (`cli/serve.go:421`) and the external
`brood-box` embedder require no changes. (The one `cli/serve.go` edit this task makes is
dropping the factory's `WithAggregator(agg)` wiring — see "Factory aggregator removal
(AC2 made structural)" below.)

**Why the A1 cleanup lands here.** Phase 2 deliberately did **not** delete the
authz/annotation middleware (#5441) or the discovery middleware (#5442) from the
shared `(*Server).Handler`, because `server.New` still used the legacy chain and deleting
shared middleware would have opened an authz/aggregation regression window. Instead, Phase
2 made those layers **inert on the `Serve` path** (authz/annotation via a nil
[...]

Issue #5431, opened 2026-06-03
Phase 2: Serve transport helper, re-home transport, replace discovery


tgrunnagle:
## Description

Introduce the transport half of the vMCP domain/transport split: a
`Serve(ctx, VMCP, *ServerConfig) -> *Server` helper plus a transport-only
`ServerConfig`, and relocate every transport concern that lives in the
`server.New` god-object today under `Serve`. This story also routes the new `Serve` path's
aggregation through direct `VMCP.ListTools`/`CallTool` calls (guarding the discovery
middleware to the legacy path) and makes the authz + annotation-enrichment middleware
**inert on the `Serve` path** (their decision now lives in the core admission seam built
in #5430). Crucially, **nothing is deleted in Phase 2** — those layers stay in the
shared `(*Server).Handler` so the still-live `server.New` path is unaffected; physical
removal is deferred to Phase 3 (#5445). This is the **largest phase (5-6 PRs)** and
maps to **RFC Phase 2**; it moves the transport seam (mcp-go server, SDK hooks,
two-phase session creation, full middleware chain, AS runner, status reporter,
optimizer, health monitor) out from behind `server.New`.

## Context

See [RFC THV-0076: vMCP Core Interface](/Users/trey/Documents/GitHub/stacklok/toolhive-rfcs/rfcs/THV-0076-vmcp-core-interface.md)
for full design details. Part of the vMCP interface refactor (epic #5419).

This story consumes the `VMCP` interface and `New(cfg) -> VMCP` core delivered
by #5430 (Phase 1) and builds the transport layer that drives it. Per
`architecture.md` ("PR-Sized Decomposition Guidance → Phase 2", and "Serve
(transport) wiring relocated from `server.New` + Handler + Start"), `Serve`
takes ownership of the mcp-go server (`NewMCPServer`, `NewStreamableHTTPServer`),
the route mux and HTTP lifecycle, the three SDK hooks and two-phase session
creation (`handleSessionRegistration*`, `lazyInjectSessionTools`,
`transportsession.NewManager`, `buildSessionDataStorage`, `sessionmanager.New`),
the full middleware chain, the embedded AS runner, the status reporter goroutine,
the optimizer cleanup, and the health monitor. The session-scoped concerns —
the "fixed at initialize" capability set, identity binding
(`MetadataKeyIdentityBinding`), and cross-pod Redis rehydration — stay in
Serve/the session layer and never cross the `VMCP` boundary (anti-pattern #5).

[...]

PR #5482, opened 2026-06-10
Cover Serve-path authz and annotation-enrichment omission

### LINKED ISSUE #5441: P2.3 Move middleware chain under Serve; remove authz + annotation mw

tgrunnagle:
## Summary

Part of the vMCP core refactor (epic #5419, parent story #5431; blocks #5442). This is
P2.3.

The `Serve`-built `*Server` must produce the MCP middleware chain **without** the HTTP
authz and annotation-enrichment layers — authorization is moving to the core admission
seam (#5438) — while the still-live `server.New` path must keep enforcing authz with no
regression. The nil-guard that makes this possible already exists from dependency #5439:
`Serve` leaves `cfg.AuthzMiddleware` nil (`buildServeConfig` does not map it), so the two
`s.config.AuthzMiddleware != nil` guards in the **shared** `(*Server).Handler` skip both
the authz block and the annotation-enrichment block on the `Serve` path, while the
`server.New` path (which still sets `AuthzMiddleware`) continues to run both.

Because the mechanism already exists, the production-code delta here is
comment-synchronization only; the substance of this PR is two new tests that lock in the
behavior and prove the single shared `Handler` correctly serves both modes.

- **Why:** Lock in and document that the `Serve` path omits HTTP authz/annotation via the
  existing nil-guard, without opening an authz-regression window on the still-live
  `server.New` path. Per the issue's "guard, don't delete (yet)" strategy, the blocks stay
  in the shared `Handler`; physical removal is deferred to Phase 3 (#5445).
- **What:** Added two tests covering both sides of the guard, and synchronized stale
  forward-reference comments left behind by earlier phases.

Closes #5441

## Type of change

## Test plan

`pkg/vmcp/server` and `pkg/vmcp/core` pass with `-race`. `task lint-fix` is clean (only a
pre-existing, unrelated gosec G115 finding in `cmd/thv/app/upgrade.go` remains, untouched
by this PR).

## Changes

| File | Change |
|------|--------|
| `pkg/vmcp/server/server_test.go` | New `TestHandlerAppliesAuthzAndAnnotationOnlyWhenConfigured`: drives the shared `Handler` with `AuthzMiddleware` set vs. nil and observes that both authz and annotation-enrichment run iff configured — proving the same `Handler` serves both modes and the blocks are not deleted. |
| `pkg/vmcp/server/serve_test.go` | New `TestServeOmitsAuthzAndAnnotation`: asserts a `Serve`-built server has `config.AuthzMiddleware == nil` and still builds its `Handler`. Updated the `intentionallyUnmapped` comment for `AuthzMiddleware` in `TestBuildServeConfigMapsSharedFields`. |
| `pkg/vmcp/server/server.go` | Added an explanatory comment in `Handler` documenting that the authz/annotation layers are guarded by `AuthzMiddleware != nil` (on for `server.New`, off for `Serve`) and remain in the shared `Handler` until Phase 3 (#5445). |
| `pkg/vmcp/server/serve.go` | Synced `Serve`/`buildServeConfig` comments to describe the now-current middleware-chain wiring and the deliberate nil `AuthzMiddleware`. |
| `pkg/vmcp/server/annotation_enrichment.go` | Corrected the `convertAnnotations` comment: #5441 makes the middleware inert via the nil-guard; physical removal of the duplicated converter is Phase 3 (#5445). |
| `pkg/vmcp/core/admission.go` | Corrected stale #5441 attributions: the (Authz + optimizer) fail-fast lands with the core-enforcement switch in #5442; the shared-constructor reconciliation lands in Phase 3 (#5445). |
| `pkg/vmcp/server/serve_session_test.go` | Corrected discovery-middleware relocation attribution from #5441/#5442 to #5442. |

No. `server.New`'s signature and observable behavior are unchanged, and the `Serve` path
is not yet wired into any production composition root.

Per the issue's "guard, don't delete (yet)" strategy, several items are deliberately
deferred — please review with these scope boundaries in mind:

- **Authz/annotation blocks are NOT deleted here.** They remain in the shared
  `(*Server).Handler` so the still-live `server.New` path keeps enforcing authz. Physical
  removal is Phase 3 (#5445), after `server.New` is routed through `Serve`.
  `annotation_enrichment.go` (including `convertAnnotations`) is intentionally left in
  place — the core admission seam reuses `convertAnnotations` per #5438.
- **AC #6 (optimizer + authz fail-fast at the `core.New` composition root) is deferred to
  #5442.** There is no production composition root wiring `core.New` yet — `cli/serve.go`
  still uses the legacy `server.New`, which retains HTTP authz, so the (Authz + optimizer)
  combination is handled correctly there. The issue's AC #6 parenthetical explicitly
  permits carrying this criterion to #5442 if the core-enforcement switch lands there.
- **AC #5's `Serve`-path admission-seam parity lands with the request path in #5442.**
  Guarding/retiring the discovery middleware is out of scope here. The `server.New`
  parity surface is unchanged and green.

### DIFF
--- a/pkg/vmcp/core/admission.go
+++ b/pkg/vmcp/core/admission.go
@@ -174,8 +174,8 @@ func (a *cedarAdmission) FilterTools(
 // the tool's annotations for when-clause evaluation. It authorizes the tool named in
 // `tool.Name` — there is no optimizer meta-tool special-casing here (see the
 // [Admission] doc): the optimizer's call_tool inner-target authorization is deferred
-// to a dedicated optimizer-admission PR, and the (Authz + optimizer) fail-fast (#5442)
-// keeps that combination from reaching this seam in the meantime.
+// to a dedicated optimizer-admission PR; the (Authz + optimizer) fail-fast that keeps that
+// combination from reaching this seam lands with the core-enforcement switch in #5442.
 func (a *cedarAdmission) AllowToolCall(
 	ctx context.Context, identity *auth.Identity, tool *vmcp.Tool, args map[string]any,
 ) (bool, error) {
--- a/pkg/vmcp/core/admission.go
+++ b/pkg/vmcp/core/admission.go
@@ -30,9 +30,11 @@ import (
 // The optimizer-admission integration the HTTP path performs today — call_tool
 // inner-target authorization, find_tool response filtering, and inner-target
 // annotation sourcing — is deliberately NOT in this seam; it is deferred to its own
-// focused PR. Until then the optimizer keeps the HTTP middleware, and #5441's
-// composition root fails fast (vmcp.ErrInvalidConfig) when Authz is set together
-// with the optimizer, so the combination can never silently route through this seam.
+// focused PR. Until then the optimizer keeps the HTTP middleware, and the composition
+// root that wires core.New must fail fast (vmcp.ErrInvalidConfig) when Authz is set
+// together with the optimizer, so the combination can never silently route through this
+// seam. That fail-fast lands with the core-enforcement switch in #5442 (#5441 keeps the
+// legacy server.New path, which retains the HTTP authz middleware).
 type Admission interface {
 	// FilterTools returns the subset of tools the identity may call. Mirrors
 	// pkg/authz filterToolsByPolicy: a per-tool AuthorizeWithJWTClaims(call) using
@@ -60,8 +62,8 @@ type Admission interface {
 // A non-nil authzCfg with zero policies is NOT treated as allow-all here: it falls
 // through to CreateAuthorizer, which returns ErrNoPolicies, so the core fails
 // CLOSED. This is a deliberate divergence from the live HTTP path, which allows-all
-// for the same input. Fail-closed is the safer default; #5441 reconciles the two
-// when it collapses the construction paths into one shared constructor.
+// for the same input. Fail-closed is the safer default; Phase 3 (#5445) reconciles the
+// two when it collapses the construction paths into one shared constructor.
 //
 // When authzCfg is set, the authorizer is built via the same registry path the
 // HTTP middleware uses (authorizers.GetFactory + CreateAuthorizer, the construction
@@ -73,7 +75,7 @@ type Admission interface {
 // defaulting an empty EntitiesJSON to "[]") belongs to whoever builds authzCfg, not
 // to this generic factory path, which consumes RawConfig as-is. newCedarAuthzMiddleware
 // is the sibling construction path; the two must stay in lockstep (notably the
-// empty-serverName check below) until #5441 collapses them.
+// empty-serverName check below) until Phase 3 (#5445) collapses them.
 func newAdmission(authzCfg *authorizers.Config, serverName string) (Admission, error) {
 	if authzCfg == nil {
 		return allowAllAdmission{}, nil
@@ -172,8 +174,8 @@ func (a *cedarAdmission) FilterTools(
 // the tool's annotations for when-clause evaluation. It authorizes the tool named in
 // `tool.Name` — there is no optimizer meta-tool special-casing here (see the
 // [Admission] doc): the optimizer's call_tool inner-target authorization is deferred
-// to a dedicated optimizer-admission PR, and #5441 fails fast on (Authz + optimizer)
-// so that combination never reaches this seam in the meantime.
+// to a dedicated optimizer-admission PR, and the (Authz + optimizer) fail-fast (#5442)
+// keeps that combination from reaching this seam in the meantime.
 func (a *cedarAdmission) AllowToolCall(
 	ctx context.Context, identity *auth.Identity, tool *vmcp.Tool, args map[string]any,
 ) (bool, error) {
@@ -295,8 +297,8 @@ func (allowAllAdmission) AllowPromptGet(_ context.Context, _ *auth.Identity, _ *
 // than reusing it: that copy lives in package server, which imports this core
 // package, so importing it here would create a cycle (server -> core). Like the
 // filterHealthyBackends C2 duplication in this package, it is intentional and
-// temporary — #5441 retires the server-side middleware (and its copy) on the domain
-// path. Keep the two in sync until then.
+// temporary — Phase 3 (#5445) retires the server-side middleware (and its copy) on the
+// domain path. Keep the two in sync until then.
 func convertAnnotations(ann *vmcp.ToolAnnotations) *authorizers.ToolAnnotations {
 	if ann == nil {
 		return nil
--- a/pkg/vmcp/server/annotation_enrichment.go
+++ b/pkg/vmcp/server/annotation_enrichment.go
@@ -92,7 +92,9 @@ func findToolAnnotations(toolName string, caps *aggregator.AggregatedCapabilitie
 //
 // Kept identical to core.convertAnnotations (pkg/vmcp/core/admission.go), including
 // the nil guard below — the two are intentional, temporary duplicates (the core
-// cannot import this package: server imports core) and #5441 deletes this copy.
+// cannot import this package: server imports core). #5441 makes this middleware inert
+// on the Serve path via the AuthzMiddleware nil-guard; physical removal of this copy
+// is deferred to Phase 3 (#5445).
 func convertAnnotations(ann *vmcp.ToolAnnotations) *authorizers.ToolAnnotations {
 	if ann == nil {
 		return nil
--- a/pkg/vmcp/server/annotation_enrichment.go
+++ b/pkg/vmcp/server/annotation_enrichment.go
@@ -94,7 +94,9 @@ func findToolAnnotations(toolName string, caps *aggregator.AggregatedCapabilitie
 // the nil guard below — the two are intentional, temporary duplicates (the core
 // cannot import this package: server imports core). #5441 makes this middleware inert
 // on the Serve path via the AuthzMiddleware nil-guard; physical removal of this copy
-// is deferred to Phase 3 (#5445).
+// is deferred to Phase 3 (#5445). Do NOT delete it early on the import-cycle reasoning
+// alone: the core admission seam (#5438) reuses the conversion logic, so this copy must
+// stay in lockstep with core.convertAnnotations until the middleware itself is retired.
 func convertAnnotations(ann *vmcp.ToolAnnotations) *authorizers.ToolAnnotations {
 	if ann == nil {
 		return nil
--- a/pkg/vmcp/server/serve_test.go
+++ b/pkg/vmcp/server/serve_test.go
@@ -194,6 +194,32 @@ func TestServeHandlerRegistersUnauthenticatedRoutes(t *testing.T) {
 	assert.NotEqual(t, http.StatusOK, rec.Code)
 }
 
+// TestServeOmitsAuthzAndAnnotation proves the Serve path produces the MCP middleware
+// chain WITHOUT the authz and annotation-enrichment layers. The mechanism is the shared
+// (*Server).Handler guard `s.config.AuthzMiddleware != nil`: Serve leaves AuthzMiddleware
+// nil (buildServeConfig does not map it — see TestBuildServeConfigMapsSharedFields), so
+// both the authz block (server.go:614) and the AnnotationEnrichmentMiddleware block
+// (:622) are skipped on the Serve path. Authorization instead runs through the core
+// admission seam (#5438). The blocks are NOT deleted here — they stay in the shared
+// Handler so the still-live server.New path keeps enforcing authz; physical removal is
+// Phase 3 (#5445). The companion TestHandlerAppliesAuthzAndAnnotationOnlyWhenConfigured
+// proves the same shared Handler DOES apply both layers when AuthzMiddleware is non-nil.
+func TestServeOmitsAuthzAndAnnotation(t *testing.T) {
+	t.Parallel()
+
+	srv, err := Serve(context.Background(), &stubVMCP{}, testMinimalServeConfig())
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = srv.Stop(context.Background()) })
+
+	assert.Nil(t, srv.config.AuthzMiddleware,
+		"Serve must leave AuthzMiddleware nil so the shared Handler omits authz + annotation-enrichment")
+
+	// The Serve path still produces the rest of the chain: Handler builds without error.
+	handler, err := srv.Handler(context.Background())
+	require.NoError(t, err)
+	require.NotNil(t, handler)
+}
+
 func TestServeHandlerRegistersMetricsWhenTelemetryEnabled(t *testing.T) {
 	t.Parallel()
 
@@ -300,7 +326,7 @@ func TestBuildServeConfigMapsSharedFields(t *testing.T) {
 	t.Parallel()
 
 	intentionallyUnmapped := map[string]struct{}{
-		"AuthzMiddleware":     {}, // authenticated/authz chain relocated by #5441
+		"AuthzMiddleware":     {}, // intentionally nil on Serve path; authz moves to core admission seam (#5438), shared Handler skips it
 		"HealthMonitorConfig": {}, // monitor injected pre-built via ServerConfig.HealthMonitor (A2)
 		"StatusReporter":      {}, // set directly on Server; Config.StatusReporter only read by New
 		"SessionFactory":      {}, // session manager built in Serve from ServerConfig.SessionManagerConfig
--- a/pkg/vmcp/server/serve_test.go
+++ b/pkg/vmcp/server/serve_test.go
@@ -198,8 +198,8 @@ func TestServeHandlerRegistersUnauthenticatedRoutes(t *testing.T) {
 // chain WITHOUT the authz and annotation-enrichment layers. The mechanism is the shared
 // (*Server).Handler guard `s.config.AuthzMiddleware != nil`: Serve leaves AuthzMiddleware
 // nil (buildServeConfig does not map it — see TestBuildServeConfigMapsSharedFields), so
-// both the authz block (server.go:614) and the AnnotationEnrichmentMiddleware block
-// (:622) are skipped on the Serve path. Authorization instead runs through the core
+// both the authz block and the AnnotationEnrichmentMiddleware block — each gated on that
+// guard in Handler — are skipped on the Serve path. Authorization instead runs through the core
 // admission seam (#5438). The blocks are NOT deleted here — they stay in the shared
 // Handler so the still-live server.New path keeps enforcing authz; physical removal is
 // Phase 3 (#5445). The companion TestHandlerAppliesAuthzAndAnnotationOnlyWhenConfigured
--- a/pkg/vmcp/server/server_test.go
+++ b/pkg/vmcp/server/server_test.go
@@ -17,7 +17,11 @@ import (
 	"go.uber.org/mock/gomock"
 
 	"github.com/stacklok/toolhive/pkg/audit"
+	"github.com/stacklok/toolhive/pkg/authz/authorizers"
+	mcpparser "github.com/stacklok/toolhive/pkg/mcp"
 	"github.com/stacklok/toolhive/pkg/vmcp"
+	"github.com/stacklok/toolhive/pkg/vmcp/aggregator"
+	"github.com/stacklok/toolhive/pkg/vmcp/discovery"
 	discoveryMocks "github.com/stacklok/toolhive/pkg/vmcp/discovery/mocks"
 	"github.com/stacklok/toolhive/pkg/vmcp/mocks"
 	"github.com/stacklok/toolhive/pkg/vmcp/optimizer"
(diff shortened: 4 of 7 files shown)


pkg/vmcp/core/admission.go
```diff
@@ -172,8 +174,8 @@ func (a *cedarAdmission) FilterTools(
 // the tool's annotations for when-clause evaluation. It authorizes the tool named in
 // `tool.Name` — there is no optimizer meta-tool special-casing here (see the
 // [Admission] doc): the optimizer's call_tool inner-target authorization is deferred
-// to a dedicated optimizer-admission PR, and #5441 fails fast on (Authz + optimizer)
-// so that combination never reaches this seam in the meantime.
+// to a dedicated optimizer-admission PR, and the (Authz + optimizer) fail-fast (#5442)
+// keeps that combination from reaching this seam in the meantime.
```
tgrunnagle: **[LOW] Present-tense fail-fast for a mechanism that doesn't exist yet.** This reads as an active safeguard, but the (Authz + optimizer) fail-fast lands in #5442 — the same file's `[Admission]` doc (lines 33-37) correctly frames it as future work ("must fail fast", "lands with…#5442"). The `(#5442)` tag makes intent recoverable, but the tense is inconsistent. Suggest aligning:

```suggestion
// to a dedicated optimizer-admission PR; the (Authz + optimizer) fail-fast that keeps that
// combination from reaching this seam lands with the core-enforcement switch in #5442.
```

pkg/vmcp/server/server_test.go
```diff
@@ -785,3 +789,122 @@ func TestAcceptHeaderValidation(t *testing.T) {
...
+			// ParsingMiddleware and the discovery middleware would). With no Mcp-Session-Id
+			// and session-scoped routing, the discovery layer passes through without using
+			// the manager, preserving the injected ctx for annotation-enrichment.
+			ctx := context.WithValue(t.Context(), mcpparser.MCPRequestContextKey,
+				&mcpparser.ParsedMCPRequest{Method: "tools/call", ResourceID: "my_tool"})
+			ctx = discovery.WithDiscoveredCapabilities(ctx, caps)
+			reqCtx, reqCancel := context.WithCancel(ctx)
+			t.Cleanup(reqCancel)
+
+			req := httptest.NewRequest(http.MethodPost, "/mcp", nil).WithContext(reqCtx)
```
tgrunnagle: **[LOW] Negative case 400s on content-type before reaching a representative chain point.** With no `Content-Type`, the nil-authz request falls through to the SDK `streamableServer`, which 400s — so the `assert.NotEqual(t, sentinelStatus, rec.Code, …)` below reduces to `NotEqual(418, 400)` and proves authz-absence only incidentally. Adding the header makes both cases exercise the chain identically (the positive case already pre-injects `ParsedMCPRequest`, so parsing is bypassed regardless):

```suggestion
			req := httptest.NewRequest(http.MethodPost, "/mcp", nil).WithContext(reqCtx)
			req.Header.Set("Content-Type", "application/json")
```

pkg/vmcp/server/server_test.go
```diff
@@ -785,3 +789,122 @@ func TestAcceptHeaderValidation(t *testing.T) {
...
+			require.NoError(t, err)
+
+			// Craft a tools/call request. This chain has no auth-parser, so inject the
+			// parsed request and discovered capabilities directly into ctx (as
+			// ParsingMiddleware and the discovery middleware would). With no Mcp-Session-Id
+			// and session-scoped routing, the discovery layer passes through without using
+			// the manager, preserving the injected ctx for annotation-enrichment.
+			ctx := context.WithValue(t.Context(), mcpparser.MCPRequestContextKey,
+				&mcpparser.ParsedMCPRequest{Method: "tools/call", ResourceID: "my_tool"})
+			ctx = discovery.WithDiscoveredCapabilities(ctx, caps)
```
tgrunnagle, line 881: **[LOW] Cross-package coupling to discovery's session-scoped pass-through.** This relies on `discovery.Middleware` (with `WithSessionScopedRouting()` + no `Mcp-Session-Id`) passing the request through unmodified (`discovery/middleware.go:110-116`), preserving these injected capabilities for annotation-enrichment. If that pass-through changes — e.g. discovery begins overwriting ctx — `annotationsSeen` silently becomes false, degrading the assertion rather than failing loudly. The comment documents the reliance; consider naming `discovery/middleware.go:110` as the load-bearing line so a future change there flags this test.

pkg/vmcp/server/server_test.go
```diff
@@ -785,3 +789,122 @@ func TestAcceptHeaderValidation(t *testing.T) {
...
+			// through to the inner SDK handler, so run in a goroutine and cancel after the
+			// short-circuit window. rec/markers are read only after the goroutine returns.
+			done := make(chan struct{})
+			go func() {
+				defer close(done)
+				handler.ServeHTTP(rec, req)
+			}()
+			select {
+			case <-done:
+			case <-time.After(200 * time.Millisecond):
```
tgrunnagle: **[LOW] Timeout machinery never fires.** Both subtests return synchronously (~0.00s): the positive case short-circuits at the sentinel, the negative 400s immediately. The 200ms `select`, `reqCancel`, and 2s secondary timeout are never exercised — ~25 lines of concurrency scaffolding (and the test's only data-race surface) for no exercised behavior. Either simplify to a direct `handler.ServeHTTP(rec, req)`, or keep the goroutine with a one-line note that it's purely defensive against a future hang regression. (No race today — the `done`-channel close establishes happens-before for the marker reads; confirmed clean under `-race -count=3`.)

tgrunnagle (commented):
## Multi-agent review summary

**Recommendation: COMMENT** — no blocking issues (0 HIGH, 0 MEDIUM). 3 specialist reviewers + independent verification; codex cross-review skipped (CLI not installed).

This PR's production delta is comment-synchronization only; the substance is two load-bearing tests proving the shared `(*Server).Handler` applies authz + annotation-enrichment iff `AuthzMiddleware != nil` (`server.New` on, `Serve` off). The "guard, don't delete (yet)" strategy from #5441 is faithfully implemented — confirmed by mutation testing that the new tests fail if the guards are removed, and that all production changes are comment-only.

### Findings (all non-blocking; details inline)

| # | Severity | File | Summary |
|---|----------|------|---------|
| 1 | LOW | core/admission.go | Fail-fast described in present tense for a #5442 mechanism that doesn't exist yet — inconsistent with same-file lines 33-37 |
| 2 | LOW | server/server_test.go | Negative case 400s on content-type before exercising the chain; `NotEqual(418)` is a weak absence proof |
| 3 | LOW | server/server_test.go | Annotation assertion silently weakens if discovery's session-scoped pass-through changes (cross-package coupling) |
| 4 | LOW | server/server_test.go | Goroutine + 200ms/2s timeout machinery never fires (both paths return synchronously) |
| 5 | INFO | server/annotation_enrichment.go | convertAnnotations comment omits the stronger "core seam #5438 reuses it" rationale |
| 6 | INFO | server/serve_test.go | "Serve omits" proof split across two files/packages; neither independently sufficient |
| 7 | INFO | server/server_test.go | Discover/List mock expectations never fire on the session-scoped path |

### Verified correct
Execution ordering (annotation-enrichment before authz), discovery session-scoped pass-through preserving injected capabilities, ParsingMiddleware no-op-when-parsed, no data race (`-race -count=3` clean), scope discipline (production = comment-only, `server.New` unchanged), SPDX headers, and cross-file issue-number attributions (#5438 / #5442 / #5445). AC #6's (Authz + optimizer) fail-fast is correctly deferred to #5442, which the issue explicitly permits.

pkg/vmcp/server/annotation_enrichment.go
```diff
@@ -92,7 +92,9 @@ func findToolAnnotations(toolName string, caps *aggregator.AggregatedCapabilitie
 //
 // Kept identical to core.convertAnnotations (pkg/vmcp/core/admission.go), including
 // the nil guard below — the two are intentional, temporary duplicates (the core
-// cannot import this package: server imports core) and #5441 deletes this copy.
+// cannot import this package: server imports core). #5441 makes this middleware inert
```
tgrunnagle, line 95: **[INFO] Comment is accurate but omits the stronger rationale.** Per the issue (AC + Technical Approach), this copy survives specifically because the core admission seam (#5438) *reuses* `convertAnnotations`. Worth a half-sentence so a future reader doesn't delete it prematurely on the "temporary duplicate / import cycle" reasoning alone.

pkg/vmcp/server/serve_test.go
```diff
@@ -194,6 +194,32 @@ func TestServeHandlerRegistersUnauthenticatedRoutes(t *testing.T) {
...
+// chain WITHOUT the authz and annotation-enrichment layers. The mechanism is the shared
+// (*Server).Handler guard `s.config.AuthzMiddleware != nil`: Serve leaves AuthzMiddleware
+// nil (buildServeConfig does not map it — see TestBuildServeConfigMapsSharedFields), so
+// both the authz block and the AnnotationEnrichmentMiddleware block — each gated on that
+// guard in Handler — are skipped on the Serve path. Authorization instead runs through the core
+// admission seam (#5438). The blocks are NOT deleted here — they stay in the shared
+// Handler so the still-live server.New path keeps enforcing authz; physical removal is
+// Phase 3 (#5445). The companion TestHandlerAppliesAuthzAndAnnotationOnlyWhenConfigured
+// proves the same shared Handler DOES apply both layers when AuthzMiddleware is non-nil.
+func TestServeOmitsAuthzAndAnnotation(t *testing.T) {
```
tgrunnagle, line 207: **[INFO] Config-contract + build-smoke test.** This proves `Serve` leaves `AuthzMiddleware` nil and the `Handler` builds; the *behavioral* omission proof lives in the companion `TestHandlerAppliesAuthzAndAnnotationOnlyWhenConfigured`. The split is reasonable (it respects the internal/external package boundary) and both tests cross-reference each other — flagging only so it's clear neither test is independently sufficient for AC bullet 2 ("on the Serve path, no authz/annotation layer is applied").

pkg/vmcp/server/server_test.go
```diff
@@ -785,3 +789,122 @@ func TestAcceptHeaderValidation(t *testing.T) {
...
+			t.Parallel()
+
+			ctrl := gomock.NewController(t)
+			t.Cleanup(ctrl.Finish)
+			mockRouter := routerMocks.NewMockRouter(ctrl)
+			mockBackendClient := mocks.NewMockBackendClient(ctrl)
+			mockDiscoveryMgr := discoveryMocks.NewMockManager(ctrl)
+			mockBackendRegistry := mocks.NewMockBackendRegistry(ctrl)
+			mockBackendRegistry.EXPECT().List(gomock.Any()).Return(nil).AnyTimes()
+			mockDiscoveryMgr.EXPECT().Discover(gomock.Any(), gomock.Any()).Return(nil, nil).AnyTimes()
```
tgrunnagle, line 839: **[INFO] These expectations never fire.** With `WithSessionScopedRouting()` and no `Mcp-Session-Id`, `discovery.Middleware` returns at `middleware.go:114` before calling `Discover`/`List`. `.AnyTimes()` is the right choice (don't switch to exact counts), but `Discover`/`List` are dead setup that could mislead a reader into thinking discovery aggregation is exercised — optionally drop them or add a one-line note that they keep the mock permissive on the pass-through path.

pkg/vmcp/core/admission.go
tgrunnagle: Addressed in 24e58361 — applied your suggestion. The note now frames the fail-fast as future work that lands with the core-enforcement switch in #5442, consistent with the [Admission] doc at lines 33-37.

pkg/vmcp/server/server_test.go
tgrunnagle: Addressed in 24e58361 — set `Content-Type: application/json` so the nil-authz request reaches the inner SDK handler representatively instead of a content-negotiation 400. Both cases now exercise the chain identically.

pkg/vmcp/server/server_test.go
tgrunnagle: Addressed in 24e58361. I documented the reliance by *mechanism* (`WithSessionScopedRouting()` + no `Mcp-Session-Id` ⇒ pass-through that preserves ctx) rather than a hard `discovery/middleware.go:110` citation — the first review round flagged absolute line numbers as drift-prone. The comment now names the behavior and points at the positive-case `assert.True(annotationsSeen)` as the guard if that pass-through ever changes.

pkg/vmcp/server/server_test.go
tgrunnagle: Addressed in 24e58361 — simplified to a direct `handler.ServeHTTP(rec, req)`. Confirmed both subtests return synchronously and the package stays clean under `-race -count=1`. Removed the goroutine, done-channel, and 200ms/2s timeouts.

pkg/vmcp/server/annotation_enrichment.go
tgrunnagle: Addressed in 24e58361 — added the half-sentence that the core admission seam (#5438) reuses the conversion logic, so this copy must stay in lockstep with `core.convertAnnotations` and should not be deleted early on the import-cycle reasoning alone.

pkg/vmcp/server/server_test.go
tgrunnagle: Addressed in 24e58361 — kept `AnyTimes()` (your recommendation) and added a one-line note that `List`/`Discover` stay permissive but do not fire on the session-scoped pass-through path; `Stop` fires via `srv.Stop` in cleanup.

pkg/vmcp/server/serve_test.go
tgrunnagle: Acknowledged — no code change. The split is intentional: it respects the internal/external package boundary, and the two tests cross-reference each other. `TestHandlerAppliesAuthzAndAnnotationOnlyWhenConfigured` carries the behavioral proof for AC bullet 2; this one carries the config contract (Serve leaves AuthzMiddleware nil) plus a build smoke test.

jerm-dro (approved):

